PR #4556, opened 2021-04-13
core: Add FFA_PARTITION_INFO

jellesels-arm:
Allow SPs to query all SPs loaded inside the system.
Change FFA_PARTITION_INFO to return all SPs loaded in the system.

PR #4620, opened 2021-05-18
FF-A add memory sharing for SPs

jellesels-arm:
FFA_MEM_SHARE was only supported to share for the Normal World to the endpoint. This PR add support to share a memory region from the Normal World with a SP or from a SP to another SP.

The following features are not supported and will be added later:
-Ownership is not tracked yet. The shared memory region is mapped in the SPs address space. The region is kept in the address space of the calling SP. Currently we still allow the same memory region to be shared again to any other endpoint both from the destination and the source endpoint. We don't keep track of who is the owner of the region or if the region is shared or exclusive.
-Memory region attributes usage are not supported. Currently all shares have the cache as TEE_MATTR_CACHE_CACHED
-When using FFA_MEM_RETRIEVE_REQ, the Address field of the constituent memory region descriptor is ignored. The regions will be mapped to virtual addresses assigned by the SPMC. 
-The memory management interfaces only support the use of the rxtx buffers when being called from a SP. The address parameter is not supported.

core: Add FFA_PARTITION_INFO will be removed when [#4556](https://github.com/OP-TEE/optee_os/pull/4556l) is merged.

Travis fails cause of 2 longer print statements on core/arch/arm/kernel/spmc_sp_handler.c:355 and on core/arch/arm/kernel/spmc_sp_handler.c:354

### DIFF
--- a/core/arch/arm/include/kernel/spmc_sp_handler.h
+++ b/core/arch/arm/include/kernel/spmc_sp_handler.h
@@ -17,6 +17,23 @@
 void spmc_sp_thread_entry(uint32_t a0, uint32_t a1, uint32_t a2, uint32_t a3);
 void spmc_sp_msg_handler(struct thread_smc_args *args,
 			 struct sp_session *caller_sp);
+TEE_Result spmc_sp_add_share(struct ffa_mem_transaction *input_descr,
+			     uint64_t global_handle, size_t blen);
+
+struct sp_shared_mem {
+	size_t counter;
+	struct shared_mem *s_mem;
+	struct ffa_mem_access *access_descr;
+	uint16_t endpoint_id;
+	SLIST_ENTRY(sp_shared_mem) link;
+};
+
+struct shared_mem {
+	TAILQ_ENTRY(shared_mem) link;
+	uint16_t owner_id;
+	struct ffa_mem_transaction *mem_descr;
+	SLIST_HEAD(mem_head_t, sp_shared_mem) sp_head;
+};
 
 #ifdef CFG_SECURE_PARTITION
 void spmc_sp_start_thread(struct thread_smc_args *args);
--- a/core/arch/arm/include/kernel/spmc_sp_handler.h
+++ b/core/arch/arm/include/kernel/spmc_sp_handler.h
@@ -26,6 +26,7 @@ struct sp_shared_mem {
 	struct ffa_mem_access *access_descr;
 	uint16_t endpoint_id;
 	SLIST_ENTRY(sp_shared_mem) link;
+	bool zero_flag;
 };
 
 struct shared_mem {
@@ -42,5 +43,4 @@ static inline void spmc_sp_start_thread(struct thread_smc_args *args __unused)
 {
 }
 #endif
-
 #endif /* __KERNEL_SPMC_SP_HANDLER_H */
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -56,6 +56,20 @@ static void set_sp_ctx_ops(struct ts_ctx *ctx)
 	ctx->ops = _sp_ops;
 }
 
+struct ts_session *sp_get_active(void)
+{
+	struct ts_session *ts = 0;
+
+	/*
+	 * Make sure that ts_get_current_session_may_fail() won't print an
+	 * error.
+	 */
+	if (thread_get_id_may_fail() >= 0)
+		ts = ts_get_current_session_may_fail();
+
+	return ts;
+}
+
 TEE_Result sp_find_session_id(const TEE_UUID *uuid, uint32_t *session_id)
 {
 	struct sp_session *s = NULL;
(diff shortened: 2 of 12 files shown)


core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -4,13 +4,20 @@
...
+#include <io.h>
 #include <kernel/panic.h>
 #include <kernel/secure_partition.h>
 #include <kernel/spinlock.h>
 #include <kernel/spmc_sp_handler.h>
 #include <optee_ffa.h>
+#include <string.h>
 #include "thread_private.h"
 
+#define FFA_NW_ID			0
```
jenswi-linaro, line 16: Please use thread_get_tsd()->rpc_target_info instead of this define which hard codes the ID of normal world.

core/arch/arm/include/kernel/spmc_sp_handler.h
```diff
@@ -17,6 +17,23 @@
 void spmc_sp_thread_entry(uint32_t a0, uint32_t a1, uint32_t a2, uint32_t a3);
 void spmc_sp_msg_handler(struct thread_smc_args *args,
 			 struct sp_session *caller_sp);
+TEE_Result spmc_sp_add_share(struct ffa_mem_transaction *input_descr,
+			     uint64_t global_handle, size_t blen);
+
+struct sp_shared_mem {
+	size_t counter;
+	struct shared_mem *s_mem;
+	struct ffa_mem_access *access_descr;
```
jenswi-linaro, line 26: Overall I'd rather avoid using the structs which work directly with the FF-A ABI unless they are a perfect match. In this case it doesn't seem like a perfect match

core/arch/arm/kernel/secure_partition.c
```diff
@@ -56,6 +56,20 @@ static void set_sp_ctx_ops(struct ts_ctx *ctx)
 	ctx->ops = _sp_ops;
 }
 
+struct ts_session *sp_get_active(void)
+{
+	struct ts_session *ts = 0;
+
+	/*
+	 * Make sure that ts_get_current_session_may_fail() won't print an
```
jenswi-linaro, line 64: This comment isn't accurate, or at least a serious understatement.

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -52,6 +59,175 @@ static TEE_Result ffa_get_dst(struct thread_smc_args *args,
...
+
+		if (!ssm)
+			return FFA_NO_MEMORY;
+
+		ssm->access_descr = mem_acc;
+
+		ssm->counter = 0;
+		ssm->endpoint_id = s->endpoint_id;
+
+		/* Only allow each endpoint ones */
```
jenswi-linaro, line 106: s/ones/once/

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -52,6 +59,175 @@ static TEE_Result ffa_get_dst(struct thread_smc_args *args,
...
+		 * struct sp_shared_mem
+		 */
+		ssm->s_mem = mem_share;
+		SLIST_INSERT_HEAD(&s->mem_head, ssm, link);
+		SLIST_INSERT_HEAD(&mem_share->sp_head, ssm,
+				  link);
+
+	} else {
+		/*
+		 * We don't register memory that is shared with
```
jenswi-linaro, line 122: Why?

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -52,6 +59,175 @@ static TEE_Result ffa_get_dst(struct thread_smc_args *args,
...
+		uctx = to_sp_ctx(ts->ctx);
+		calling_s = uctx->open_session;
+		caller_id = calling_s->endpoint_id;
+	}
+
+	num_mem_accs = READ_ONCE(input_descr->mem_access_count);
+
+	if (num_mem_accs) {
+		/*
+		 * We store the full incoming transaction buffer and have
```
jenswi-linaro, line 162: This seems to be wasting a bit of memory and also perhaps a greater risk of accessing unchecked fields.

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -52,6 +59,175 @@ static TEE_Result ffa_get_dst(struct thread_smc_args *args,
...
+			if (rc)
+				goto cleanup;
+		}
+		/* Return if we processed a valid message */
+		if (!rc)
+			return rc;
+	}
+
+cleanup:
+	if (num_mem_accs) {
```
jenswi-linaro, line 206: I'm not sure what's happening here. Are we undoing everything done above, regardless of outcome?

jenswi-linaro (commented):
Here are some comment from halfway through "core: FF-A: ADD FFA_MEM_SHARE for SPs".
This is a bit much to review in one go, especially since I'm not entire sure about what I'm reviewing.
What is this tested? What is not tested?

jforissier:
Gentle ping to @jellesels-arm, if you would like any of this to go into the next release, there is 1.5 week left before -rc1.

jellesels-arm:
> Gentle ping to @jellesels-arm, if you would like any of this to go into the next release, there is 1.5 week left before -rc1.

Sorry, I will close this one. I changed the implementation of the FF-A memory sharing. I will create a new PR later.

jforissier:
No problem.

PR #4623, opened 2021-05-18
core: pager: fix compiler warning with Clang

jforissier:
Function rwp_unpaged_iv_free() is reduced to a call to panic() when
CFG_WITH_PAGER=y and CFG_CORE_PAGE_TAG_AND_IV=y. In this case, Clang 12
suggests a noreturn attribute:

 $ make -s CFG_WITH_PAGER=y COMPILER=clang
 core/mm/fobj.c:322:1: warning: function 'rwp_unpaged_iv_free' could be
 declared with attribute 'noreturn' [-Wmissing-noreturn]
 {
 ^
 1 warning generated.

However the attribute cannot be applied since it would be inappropriate
when CFG_CORE_PAGE_TAG_AND_IV != y. Therefore, disable the warning for
the file core/mm/fobj.c when the problematic configuration is enabled.

### DIFF
--- a/core/mm/sub.mk
+++ b/core/mm/sub.mk
@@ -1,4 +1,5 @@
 srcs-y += mobj.c
 srcs-y += fobj.c
+cflags-fobj.c-$(CFG_CORE_PAGE_TAG_AND_IV) := -Wno-missing-noreturn
 srcs-y += file.c
 srcs-y += vm.c


etienne-lms:

jenswi-linaro:

jforissier:
IBART failure is a known issue, ignoring.

PR #4625, opened 2021-05-18
FF-A: secondary core init for FVP

balint-dobszay-arm:
Replace the non-standard secondary core init mechanism on FVP platform.
Depends on TF-A v2.5, related PR: https://github.com/OP-TEE/manifest/pull/183

### DIFF
--- a/core/arch/arm/plat-vexpress/fvp_spmc_pm.c
+++ b/core/arch/arm/plat-vexpress/fvp_spmc_pm.c
@@ -1,89 +1,27 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2019, Arm Limited
+ * Copyright (c) 2019-2021, Arm Limited
  */
 
-#include <arm.h>
 #include <ffa.h>
-#include <initcall.h>
-#include <keep.h>
 #include <kernel/boot.h>
-#include <kernel/interrupt.h>
-#include <kernel/misc.h>
 #include <kernel/panic.h>
-#include <mm/core_memprot.h>
-#include <mm/core_mmu.h>
-#include <platform_config.h>
-#include <sm/psci.h>
-#include <stdint.h>
-#include <string.h>
-#include <trace.h>
+#include <kernel/thread.h>
 
-/*
- * Lookup table of core and cluster affinities on the FVP. In the absence of a
- * DT that provides the same information, this table is used to initialise
- * OP-TEE on secondary cores.
- */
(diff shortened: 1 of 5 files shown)


jenswi-linaro:

gyuri-szing:
Could this be made configurable to keep backwards compatibility to TF-A v2.4?
I see this as a change to enable v2.5 compatibility, but as long as no 2.5 feature is needed, new op-tee features shall work with 2.4 too. That would make life of OEMs and platform owners easier. Any thoughts?

balint-dobszay-arm:
Makes sense, I don't see any obstacles to this. I'll update the PR.

core/arch/arm/plat-vexpress/fvp_spmc_pm.c
```diff
@@ -1,24 +1,15 @@
...
-#include <mm/core_memprot.h>
-#include <mm/core_mmu.h>
-#include <platform_config.h>
-#include <sm/psci.h>
-#include <stdint.h>
-#include <string.h>
-#include <trace.h>
+#include <kernel/thread.h>
 
+#ifdef CFG_ARM_TRUSTED_FW_2_4
```
jenswi-linaro, line 12: Does this work with the QEMU v8 setup we have too? Is it possible to keep the AArch32 compatibility?
I suppose this should be marked as deprecated too since we'd like to get rid of it at some time.

jenswi-linaro:
> That would make life of OEMs and platform owners easier

By the way do we know any of these OEMs and platform owners who uses FF-A with OP-TEE? Perhaps it isn't worth carrying this baggage?

gyuri-szing:
> 
> 
> > That would make life of OEMs and platform owners easier
> 
> By the way do we know any of these OEMs and platform owners who uses FF-A with OP-TEE? Perhaps it isn't worth carrying this baggage?

After further discussions you are right Jens, and for v2.4 compatibility there is no real demand. @balint-dobszay-arm can you please revert? Sorry for any inconveniences!

balint-dobszay-arm:
Sure, no problem.

core/arch/arm/plat-vexpress/fvp_spmc_pm.c
balint-dobszay-arm: Reverted to original version of the patch, no longer relevant.

balint-dobszay-arm:
Reverted to original version of the patch (bdef61d8).

jenswi-linaro:
Looks good, I see that my tag is already applied. OK to merge from my point of view.

PR #4626, opened 2021-05-18
mem_usage.py: consider all allocatable sections

jforissier:
All allocatable sections end up using memory when the TEE binary is
loaded. Therefore the 'A' (allocatable) flag in the readelf output is
all that matters when gathering memory usage data using mem_usage.py.
The hardcoded combinations that are currently hardcoded in the script
('AX', 'WA', 'A', 'AL') are fragile and need to be replaced. For
example, with COMPILER=clang many sections have the 'W' flag set.

### DIFF
--- a/scripts/mem_usage.py
+++ b/scripts/mem_usage.py
@@ -118,8 +118,7 @@ def main():
                  flags) = words[:8]
             except BaseException:
                 continue
-            if (flags == 'AX' or flags == 'WA' or flags == 'A' or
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

jenswi-linaro:

You may want to rephase _The hardcoded combinations that are currently hardcoded_ . I guess removing the first occurrence of hardcoded should be good enough.

PR #4629, opened 2021-05-19
tee_rpmb_fs: Adding offset to memref to ensure 32-bit aligment

litchipi:
### Problem
The supplicant from U-Boot cannot pass the RPMB data frame formatted by OP-TEE core to the RPMB driver on platform stm32mp157c-ed1, as the driver requires a 32 bits alignment for the memory address of the data.
Most of the other rpmb drivers in U-Boot expects a 32 bits aligned address as well.
The supplicant gets the data after a `struct rpmb_req` header of size 6 bytes, which breaks the 4 bytes (32 bits) alignment.
Because the problem is generalized to most of the rpmb drivers, the problem is addressed in optee_os directly.

### Changes
Modifying thread_param struct to add the correct offset.
The `struct rpmb_req` with the offset reaches a size modulo 32bits.
As the memory allocation size is already roundup to 32bits, no need to add the offset as it is calculated to reach a roundup of 32bits.
Adding offset to `struct rpmb_req` fixes this and doesn't break the ABI.

Signed-off-by: Timothée Cercueil <timothee.cercueil@st.com>
Signed-off-by: Timothée Cercueil <litchi.pi@protonmail.com>

### DIFF
--- a/core/tee/tee_rpmb_fs.c
+++ b/core/tee/tee_rpmb_fs.c
@@ -211,6 +211,9 @@ struct rpmb_req {
 	/* uint8_t data[]; REMOVED! */
 };
 
+#define RPMB_REQ_MEMREF_OFFSET \
+	(ROUNDUP(sizeof(struct rpmb_req), 4) - sizeof(struct rpmb_req))
+
 #define TEE_RPMB_REQ_DATA(req) \
 		((void *)((struct rpmb_req *)(req) + 1))
 
@@ -453,6 +456,8 @@ static TEE_Result tee_rpmb_alloc(size_t req_size, size_t resp_size,
 		goto out;
 	}
 
+	*req = (uint8_t *)*req + RPMB_REQ_MEMREF_OFFSET;
+
 	mem->req_size = req_size;
 	mem->resp_size = resp_size;
 
@@ -465,8 +470,8 @@ out:
 static TEE_Result tee_rpmb_invoke(struct tee_rpmb_mem *mem)
 {
 	struct thread_param params[2] = {
-		[0] = THREAD_PARAM_MEMREF(IN, mem->phreq_mobj, 0,
-					  mem->req_size),
+		[0] = THREAD_PARAM_MEMREF(IN, mem->phreq_mobj,
+				RPMB_REQ_MEMREF_OFFSET, mem->req_size),
 		[1] = THREAD_PARAM_MEMREF(OUT, mem->phresp_mobj, 0,
 					  mem->resp_size),
 	};


jenswi-linaro:
I don't understand why this is needed. The buffer is originally allocated by U-Boot with a requested minimum alignment of 8 bytes.

litchipi:
> I don't understand why this is needed. The buffer is originally allocated by U-Boot with a requested minimum alignment of 8 bytes.

I updated the PR message (should have done it clear in the first place) with elements of answer.
U-Boot allocates the buffer, but the RPMB data is passed after the `struct rpmb_req` which adds 6 bytes, and that is what breaks the alignment.
The initial solution was to allocate another buffer and copy the RPMB raw frame inside, but someone (I think it is Ilias but I am not sure, I will add it to the message after verification) gave Etienne and I a solution with no copy and yet without breaking the ABI

litchipi:
The CI has failed because the offset is not properly handled by Linux tee-supplicant, it requires this patch to work: https://github.com/OP-TEE/optee_client/pull/270

Erratum: Apparently it was you who proposed us this fix idea

jenswi-linaro:
If it breaks with Linux it breaks the ABI. We need to find a way that doesn't break the ABI.

litchipi:
Actually the PR for optee_client is a fix for a bug that didn't appeared before because nobody used offset (apparently).
Linux broke because of the presence of an offset, and that resulted in not the good size, but that was before he even touched to any rpmb-related stuff.
So I don't think it breaks the ABI, it's just that Linux tee-supplicant had a bug concerning offsets in memory references from optee-core.

jenswi-linaro:
The ABI breakage rule is about not breaking stuff when you make changes. If it's technically not an ABI change doesn't matter, the problem remains. With this change we'll break RPMB in tee-supplicant if I haven't misunderstood things.

That we have a fix for this in tee-supplicant is nice, but it doesn't help those that don't have that change.

litchipi:
A workaround would be to allocate more, and declare that the size is S+Offset in the memref (optee-core side), this way, the tee-supplicant when substracting the offset would get the proper size. It only costs some more bytes of memory.
But does that mean the fix in tee-supplicant has to be cancelled ? And the bug kept ?

etienne-lms:
as implementation is done currently, we reference to the buffer with a non-null offset.
True, this breaks tee-supplicant because of the bug.
I think the tee-supplicant shall be fixed. Your thoughts?

But, as Jens says, maybe we can't already leverage the fix, because breaking old supplicants.
So we can't use this P-R any any alternative using an offset.

Remains the possible change in u-boot @litchipi mentionned, to use a temporary aligned rpmb frame buffer when needed.
Is that the preferred way?

jenswi-linaro:
> as implementation is done currently, we reference to the buffer with a non-null offset.
> True, this breaks tee-supplicant because of the bug.
> I think the tee-supplicant shall be fixed. Your thoughts?

+1

> 
> But, as Jens says, maybe we can't already leverage the fix, because breaking old supplicants.
> So we can't use this P-R any any alternative using an offset.

Correct, not directly at least.

> 
> Remains the possible change in u-boot @litchipi mentionned, to use a temporary aligned rpmb frame buffer when needed.
> Is that the preferred way?

It is unfortunate. I'd prefer if we could avoid this.

litchipi:
@jenswi-linaro Do you still preconise to apply changes in optee_os or should I find a workaround in linux / u-boot ?

A way to still bring the changes would be to add a config switch to activate manually, saying the linux tee-supplicant has be fixed, but the problem remains.
Is there a way to influence the underlying pointer in a memobj to apply a dirty +OFFSET on the pointer to be passed as a memref ?
The allocation is done through `mem->phreq_mobj = thread_rpc_alloc_payload(size)`, is there a way to access the underlying memory address in optee_os context ? Is it even desirable to have such a hack ?

jenswi-linaro:
> @jenswi-linaro Do you still preconise to apply changes in optee_os

We can do changes, but they must not break existing setups. They should also not be too complicated, I imagine that we at some point in the future might want to remove legacy stuff or at least make it configurable. If that happens it should be nice to have a rather clean solution left.

 or should I find a workaround in linux / u-boot ?
> 
> A way to still bring the changes would be to add a config switch to activate manually, saying the linux tee-supplicant has be fixed, but the problem remains.
> Is there a way to influence the underlying pointer in a memobj to apply a dirty +OFFSET on the pointer to be passed as a memref ?

With it looks like tee-supplicant would accept a slightly larger RPMB buffer and still do the right thing. I'm not sure if it's worth building on that though.

> The allocation is done through `mem->phreq_mobj = thread_rpc_alloc_payload(size)`, is there a way to access the underlying memory address in optee_os context ? Is it even desirable to have such a hack ?

Another option could be add another parameter to `OPTEE_SMC_RPC_FUNC_ALLOC` and `OPTEE_RPC_CMD_SHM_ALLOC` specifying a header-size parameter at bit like the `hdr_size` parameter for `raw_memalign()`. Or perhaps rather code the header-size into the alignment field since it's currently ignored in both Linux and U-Boot. 

With this said, perhaps the easiest for now would still be to double buffer in U-Boot. It's not like it's going to be a performance problem, it's only a bit ugly. What do others think?

litchipi:
A patch is currently being discussed in the U-boot mailing list to deal with this issue in a more ABI-friendly way.
It seams that RPMB-related drivers needs to ensure memory alignment anyway, so the fix that will result from the discussion should address this issue inside U-boot.
[Patch proposed in U-boot mailing list archive](http://u-boot.10912.n7.nabble.com/PATCH-optee-rpmb-c-Fix-driver-routing-memory-alignment-using-a-temporary-buffer-tt451690.html)

litchipi:
Update: After a request to put the patch in a more generic part of the code, a new patch has been proposed to the U-boot mailing list to fix the issue: http://u-boot.10912.n7.nabble.com/PATCH-mmc-rpmb-Fix-driver-routing-memory-alignment-with-tmp-buffer-tt452144.html

PR #4631, opened 2021-05-19
core: a32: fix parameter passing for __thread_std_smc_entry()

jenswi-linaro:
With the commit referred below is __thread_std_smc_entry() changed to
take 6 arguments instead of 4. This means with the arm32 calling
convention [1] that the last two parameters are passed on the stack.
This is handled automatically by the C compiler, but has to be done by
hand when calling from assembly. __thread_std_smc_entry() is called from
assembly so fix the two places where the function is called.

Link [1]: https://developer.arm.com/documentation/ihi0042/latest/

Fixes: 4107d2f93e3e ("core: add a4 and a5 to thread_alloc_and_run()")

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc_a32.S
+++ b/core/arch/arm/kernel/thread_optee_smc_a32.S
@@ -161,7 +161,9 @@ DECLARE_KEEP_PAGER thread_vector_table
 
 FUNC thread_std_smc_entry , :
 UNWIND(	.cantunwind)
+	push	{r4, r5} /* Pass these following the arm32 calling convention */
 	bl	__thread_std_smc_entry
+	add	sp, sp, #8 /* There's nothing return, just restore the sp */
 	mov	r4, r0	/* Save return value for later */
 
 	/* Disable interrupts before switching to temporary stack */
--- a/core/arch/arm/kernel/thread_spmc_a32.S
+++ b/core/arch/arm/kernel/thread_spmc_a32.S
@@ -50,8 +50,10 @@ END_FUNC ffa_msg_send_direct_resp
 FUNC thread_std_smc_entry , :
 UNWIND(	.cantunwind)
 
+	push	{r4, r5} /* Pass these following the arm32 calling convention */
 	ror	r4, r0, #16 /* Save target info with src and dst swapped */
 	bl	__thread_std_smc_entry
+	add	sp, sp, #8 /* There's nothing return, just restore the sp */
 	mov	r5, r0	/* Save return value */
 
 	/* Mask all maskable exceptions before switching to temporary stack */


core/arch/arm/kernel/thread_optee_smc_a32.S
```diff
@@ -161,7 +161,9 @@ DECLARE_KEEP_PAGER thread_vector_table
 
 FUNC thread_std_smc_entry , :
 UNWIND(	.cantunwind)
+	push	{r4, r5}
```
jforissier, line 164: Maybe add `/* arm32 calling convention args 5 and 6 passed on the stack */`?

core/arch/arm/kernel/thread_spmc_a32.S
```diff
@@ -50,8 +50,10 @@ END_FUNC ffa_msg_send_direct_resp
 FUNC thread_std_smc_entry , :
 UNWIND(	.cantunwind)
 
+	push	{r4, r5} /* Pass these following the arm32 calling convention */
```
jforissier, line 53: OK, use the same comment as in the other file (either mine or yours it doesn't matter much)

core/arch/arm/kernel/thread_spmc_a32.S
```diff
@@ -50,8 +50,10 @@ END_FUNC ffa_msg_send_direct_resp
...
 	ror	r4, r0, #16 /* Save target info with src and dst swapped */
```
jforissier, line 54: Shouldn't this be done before pushing `r4`?

jforissier (commented):
I'm wondering why the issue appeared only with Clang. Behavior dependent on the stack content?

jenswi-linaro:
> I'm wondering why the issue appeared only with Clang. Behavior dependent on the stack content?

The arguments where used so gcc never loaded from the addresses. Clang however found it necessary to load and immediately throw away the content. Good in this case since we caught the bug. :-)

core/arch/arm/kernel/thread_optee_smc_a32.S
jenswi-linaro: Yes sorry, I had it in the changed buffer in my editor.

core/arch/arm/kernel/thread_spmc_a32.S
jenswi-linaro: No, it is the content of `r4` when `thread_std_smc_entry()` was called we'd like to pass on.
What we do with `r4` after the push has nothing to do with that, `r4` is just another preserved register.

jenswi-linaro:
Update

core/arch/arm/kernel/thread_spmc_a32.S
jforissier: Of course :facepalm:

jforissier (approved):

jenswi-linaro:
Squashed and tag applied.

PR #4614, opened 2021-05-17
Add support for OP-TEE TA signature stitching during the build process.

aleksandarn-ee:
Signed-off-by: Aleksandar Nikolic <aleksandar.nikolic@mixed-mode.de>

The current build process does not allow stitching of the signature to the OP-TEE TA (e.g. trusted_keys), but only signing. The patch allows the user to sign the OP-TEE TAs (or more accurately the .stripped.elf files) offline and stitch the generated signature during the OP-TEE build process.

**Usage**

The following environment variables are to be defined by the user:

```
CFG_STITCH_TA=y     # Perform the signature stitching instead of TA signing.
TA_PUBLIC_KEY       # Path to the public key used for signature verification. Defaults to TA_SIGN_KEY.
TA_SIG_DIR          # Path to the directory where the signatures are. Defaults to ./signatures. The stitching process expects the signature files to be named as: <ta-uuid>.sig
TA_SIGN_ALG         # Hash and signing algorithm. Defaults to TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA256 (same as in sign_encrypt.py).
```

**Example in Yocto**

```
EXTRA_OEMAKE = " \
    ...
    CFG_STITCH_TA=y \
    TA_PUBLIC_KEY=<path to the public key> \
    TA_SIG_DIR=<my-custom-sig-dir> \
    TA_SIGN_ALG=TEE_ALG_RSASSA_PKCS1_V1_5_SHA256 \
    ...
"
```

**Additional improvement**

With this patch the OPTEE build process expects the signature files to be named as: `<ta-uuid>.sig`. In my opinion this shouldn't be an issue for the user. However, if someone does not agree with this, I could adjust the patch so the user can give the list of signatures to be used during the stitching process (this makes things complicated imho)

PR #4632, opened 2021-05-20
TA offline signing

jforissier:


ta/arch/arm/link.mk
```diff
@@ -103,12 +103,21 @@ ifeq ($(CFG_ENCRYPT_TA),y)
 crypt-args$(user-ta-uuid) := --enc-key $(TA_ENC_KEY)
 cmd-echo$(user-ta-uuid) := SIGNENC
 endif
+ifeq ($(TA_SIGN_METHOD),stitch)
+# Offline signing
+TA_SIG_DIR ?= signatures
+crypt-args$(user-ta-uuid) := --sig $(TA_SIG_DIR)/$(user-ta-uuid).sig
+sign-method$(user-ta-uuid) := stitch
+cmd-echo$(user-ta-uuid) := SIGNST #
```
wamserma, line 111: stray `#`

jforissier:
> @jforissier Will you also add a variable for the signing scheme or I should have it locally? More info [here](https://github.com/OP-TEE/optee_os/pull/4614#discussion_r635880695).

Missed that sorry. Patch updated.

aleksandarn-ee:
> > @jforissier Will you also add a variable for the signing scheme or I should have it locally? More info [here](https://github.com/OP-TEE/optee_os/pull/4614#discussion_r635880695).
> 
> Missed that sorry. Patch updated.

No worries.

I tested the patch, the building process seems great now, however one thing is driving me crazy and I have no idea what is going on. After building the OPTEE core and TA elf files, I am stitching the signatures with the following command (we use the default `TA_SIG_DIR` directory):

```
TA_SIGN_METHOD=stitch TA_SIG_ALGO=TEE_ALG_RSASSA_PKCS1_V1_5_SHA256 make ta
```

However, the error is as following:

```
alenik@nb152:~/projects/github/optee/optee_os_jforissier$ TA_SIGN_METHOD=stitch TA_SIG_ALGO=TEE_ALG_RSASSA_PKCS1_V1_5_SHA256 make ta
  CHK     out/arm-plat-vexpress/conf.mk
  CHK     out/arm-plat-vexpress/include/generated/conf.h
  SIGNST  out/arm-plat-vexpress/ta/avb/023f8f1a-292a-432b-8fc4-de8471358067.ta
ERROR:sign_encrypt.py:No signature file found. Please sign
 023f8f1a-292a-432b-8fc4-de8471358067.dig
offline and place the signature at 
 /023f8f1a-292a-432b-8fc4-de8471358067.sig
or pass a different location using the --sig argument.

make: *** [ta/arch/arm/link.mk:139: out/arm-plat-vexpress/ta/avb/023f8f1a-292a-432b-8fc4-de8471358067.ta] Fehler 1
```

From the output above, the variable `TA_SIG_DIR` seems to not be even set (notice `signatures` is missing here ` /023f8f1a-292a-432b-8fc4-de8471358067.sig`). 

I debugged a little further and apparently all the variables in this if statement are **not** set. Interestingly enough when I compile with `make -j4` `TA_SIG_DIR` is set to `signatures` but only for the `trusted_keys` and `pkcs11` TAs, which are built after the `avb` TA. Finally, compiling the TAs with `TA_SIG_DIR=./signatures` works.

Am I missing something here? I know the problem is not in the OPTEE Makefile, since `ifeq` is used throughout the code and people would notice the problem, but rather how I call the `make` command.

jforissier:
Updated, @aleksandarn-ee please try again :)

core/arch/arm/kernel/link.mk
```diff
@@ -216,6 +216,9 @@ $(link-out-dir)/tee.symb_sizes: $(link-out-dir)/tee.elf
 	@$(cmd-echo-silent) '  GEN     $@'
 	$(q)$(NMcore) --print-size --reverse-sort --size-sort $< > $@
 
+tee_bin: $(link-out-dir)/tee-header_v2.bin $(link-out-dir)/tee-pager_v2.bin \
+	 $(link-out-dir)/tee-pageable_v2.bin
```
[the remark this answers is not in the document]
etienne-lms: +1   platform can also boot from `tee.bin`. I agree it should be built.

ta/arch/arm/link.mk
```diff
@@ -73,6 +73,8 @@ link-script-cppflags-$(sm) := \
 
 link-script-pp-makefiles$(sm) = $(filter-out %.d %.cmd,$(MAKEFILE_LIST))
 
+TA_SIG_DIR ?= signatures
```
etienne-lms, line 76: Should be documented, somehting like:
`Path of the directory that stores file <uuid>.sig files each storing the signature to be stitched to <uuid>.elf file into a loadable signed <uuid>.ta image.`

etienne-lms (commented):
"Add 'ta_elf' target"
"Add 'ta' target"
"Add 'tee_bin' target" (with @aleksandarn-ee's comment addressed)
Maybe prefix each header line with: `make: ` or `build: `

**(edited):** Or change header line to "Add 'xxx' make target".

ta/arch/arm/link.mk
```diff
@@ -103,12 +105,23 @@ ifeq ($(CFG_ENCRYPT_TA),y)
 crypt-args$(user-ta-uuid) := --enc-key $(TA_ENC_KEY)
 cmd-echo$(user-ta-uuid) := SIGNENC
 endif
+ifeq ($(TA_SIGN_METHOD),stitch)
```
etienne-lms, line 108: Should we set a default value to TA_SIGN_METHOD, in mk/config.mk (i.e. `TA_SIGN_METHOD=sign-enc`)?

wamserma: Not necessary. The script defaults to signing.

ta/arch/arm/link.mk
```diff
@@ -103,12 +105,23 @@ ifeq ($(CFG_ENCRYPT_TA),y)
...
+# Offline signing
+crypt-args$(user-ta-uuid) := --sig $(TA_SIG_DIR)/$(user-ta-uuid).sig
```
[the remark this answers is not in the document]
jforissier: Good catch, will fix.

core/arch/arm/kernel/link.mk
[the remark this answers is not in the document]
jforissier: OK

ta/arch/arm/link.mk
jforissier: I will add a comment. Yes, all paths are relative to the top level directory of the project.

ta/arch/arm/link.mk
jforissier: `sign_enc.py` assumes sign-enc is the default command, so I'd rather keep it as is for simplicity.

jforissier:
Update

ta/arch/arm/link.mk
etienne-lms: It makes the makefiles implementation not that obvious i think. TA_SIGN_METHOD referred only at 2 places, only 1 value found, alternative is its default non-existence empty value.
But i'm fine as-is if you think it's easier to maintain that way.

jforissier:
@jbech-linaro is IBART missing `sudo apt install libgmp-dev`? First time I see this error... :confused:

jbech-linaro:
> @jbech-linaro is IBART missing `sudo apt install libgmp-dev`? First time I see this error...

Yes, weird, but I've installed the package now. Thanks!

ta/arch/arm/link.mk
aleksandarn-ee: > Path of the directory that stores file <uuid>.sig files each storing the signature to be stitched to <uuid>.elf file into a loadable signed <uuid>.ta image.

Aren't signatures added to the `<uuid>.stripped.elf` files?

aleksandarn-ee:
@jforissier Does the OPTEE documentation needs to be updated?

wamserma:
> @jforissier Does the OPTEE documentation needs to be updated?

Yes, here https://optee.readthedocs.io/en/latest/building/trusted_applications.html#offline-signing-of-tas
I can take care of the doc once this PR is merged.

jforissier:
@wamserma that would be great, thanks

vesajaaskelainen:
Just a comment on what we do.

We use devel keys during building process for signing and then re-sign the elf image in our code signing environment. This way we know exactly what we signed and have a trace. These hash signing things do not leave the trace.

aleksandarn-ee:
> Just a comment on what we do.
> 
> We use devel keys during building process for signing and then re-sign the elf image in our code signing environment. This way we know exactly what we signed and have a trace. These hash signing things do not leave the trace.

@vesajaaskelainen Doesn't the devel public key get embedded into the OPTEE core image during building in this case?

vesajaaskelainen:
> > Just a comment on what we do.
> > We use devel keys during building process for signing and then re-sign the elf image in our code signing environment. This way we know exactly what we signed and have a trace. These hash signing things do not leave the trace.
> 
> @vesajaaskelainen Doesn't the devel public key get embedded into the OPTEE core image during building in this case?

Nope:
https://github.com/OP-TEE/optee_os/commit/3ecd96bbf821d170671cc19a9bb6dc3eba5b9136 - core: Add support to import external TA signing public key

So in devel builds for TEE OS the keys will be the same -- but when preparing for TEE OS release we inject real TA release signing public key in TEE OS -- then we can first test TA with devel TEE OS and once verified to be OK then we can re-sign the TA binary with release keys and then use it with TEE OS with TA release public key.

Only catch with this is that both TA signing keys must be of same size so that signature takes same amount of data. Eg. you cannot mix 2k and 4k keys.

aleksandarn-ee:
> Nope:
3ecd96b - core: Add support to import external TA signing public key

Didn't see this commit until now, it makes a lot of sense to go this way.

> Only catch with this is that both TA signing keys must be of same size so that signature takes same amount of data. Eg. you cannot mix 2k and 4k keys.

I dont understand this fully. Why is it a problem if the devel keys are e.g. RSA2048 and the release keys are RSA4096? The public key that goes into the OPTEE OS will be an RSA4096 public key and the TAs will be signed with the RSA4096 private key? Unless the OPTEE build fails if the release public key (TA_PUBLIC_KEY) and devel private key (TA_SIGN_KEY) do not match in size? 

@jforissier Considering the commit 3ecd96b, do we need this PR? Maybe only the new Make Targets? Either way I would update the OPTEE documentation so it nicely describes the commit 3ecd96b.

jforissier:
@aleksandarn-ee 

> @jforissier Considering the commit [3ecd96b](https://github.com/OP-TEE/optee_os/commit/3ecd96bbf821d170671cc19a9bb6dc3eba5b9136), do we need this PR? Maybe only the new Make Targets? Either way I would update the OPTEE documentation so it nicely describes the commit [3ecd96b](https://github.com/OP-TEE/optee_os/commit/3ecd96bbf821d170671cc19a9bb6dc3eba5b9136).

I suggest you propose a doc update and we'll see if any of the commits in this PR bring any value. @vesajaaskelainen thanks for chiming in!

aleksandarn-ee:
> I suggest you propose a doc update and we'll see if any of the commits in this PR bring any value. 

@jforissier  @wamserma Something like [this](https://github.com/OP-TEE/optee_docs/pull/125)? Feel free to comment and update by yourself.

jforissier:
> [...] we'll see if any of the commits in this PR bring any value [...]

Apparently not, let's close this.

PR #4633, opened 2021-05-20
core: pta: scmi: fix missing threaded state of the channel

etienne-lms:
Enable SMT channel threaded state when SCMI PTA gets a channel. Before
this fixup, Core panics when SCMI message is posted since the assertion
on channel threaded field value in scmi_smt_threaded_entry() when
in debug mode.

Fixes: b0a1c2504aaf ("core: pta: scmi: new interface to REE SCMI agent")


jenswi-linaro (commented):

etienne-lms:
tag applied.

PR #4354, opened 2021-01-25
plat: rcar: enable CORE ASLR

lorc:
This set of patches enables various features on Renesas RCAR platform to enable core ASLR.

During testing it appeared that `CFG_VIRTUALIZATION` conflicts with `CFG_CORE_ASLR` , so for time being they are made mutually exclusive. I'm looking at this issue right now.

PR #4416, opened 2021-02-23
core: drivers: export crypto_rng_read to REE

ldts:
This enables the REE to register the TEE's crypto_rng_read as a hwrng
provider.

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>

PR #4641, opened 2021-05-26
generic HW RNG pseudo TA

sa-kib:
Initial idea behind this patch series is to use Linux driver `optee_rng`  on platforms other than Developerbox.
`optee_rng` driver itself is pretty generic and platform-agnostic, on the other hand companion PTA is specific to Developerbox, as it uses specific sensors for entropy collection.
Also several platforms equipped with hardware-based RNG already allow for its usage by implementing `hw_get_random_byte()` routine, which so far seems to be used only internally by OP-TEE.
So aim of the patch series is to provide Linux with strong entropy source already available to TEE, via `optee_rng` and re-using its PTA interfaces.

Important notes:
1. merged ~Patches has been verified on rcar platform, and therefore depend on #4354 which implements `hw_get_random_byte()` for rcar.~
2. Entropy quality of 1024 bits is typical choise for HW RNGs, yet it needs clarification from vendor. So it's not final and is specified for RFC.


etienne-lms:
Thank for the proposal.
For info: P-R https://github.com/OP-TEE/optee_os/pull/4416 proposed the same feature.
RNG quality and rates where set from C source, here you propose specific CFG_* switches. What do you think about how https://github.com/OP-TEE/optee_os/pull/4416 defines these?

etienne-lms:
This change make RNG PTA to expose only so-called HW RNG. Why not allowing to expose optee PRNG the same way, using a lower "quality" level, as proposed by #4416?

jenswi-linaro:
> This change make RNG PTA to expose only so-called HW RNG. Why not allowing to expose optee PRNG the same way, using a lower "quality" level, as proposed by #4416?

I think it's the other way around. Why should we expose OP-TEE PRNG to normal world? Do we have a use-case where it could be justified?

etienne-lms:
Right, i took some shortcut. I agree, there is no real point exposing a SW PRNG to REE.
I have this HW seeded PRNG thing i did not finalized. So i was rather considering this hw seeded PRNG.
Discard my comment. HWRNG is fine.

sa-kib:
> Thank for the proposal.
> For info: P-R #4416 proposed the same feature.
> RNG quality and rate **~where~ were** set from C source **while** here you propose specific CFG_* switches. What do you think about how #4416 defines these?
> **(edited)**

I've missed that PR, otherwised I'd probably re-used it.

The idea of CFG_ switch here is that _vendors_ provide it along with `hw_get_random_byte()` implementation, because they know how their TRNG operates. As far as I know they don't share details, but might share TRNG's charachteristics, or at least with what standard or recomendation it complies.
Different platforms may implement TRNG differently, hence actually having different charachteristics. That's what those CFG_ 
 in platform's `conf.mk` reflect.

core/pta/hwrng.c
```diff
@@ -0,0 +1,82 @@
...
+ *
+ */
+
+#include <kernel/pseudo_ta.h>
+#include <rng_support.h>
+#include <crypto/rng_pta_client.h>
+
+#define PTA_NAME "rng.pta"
+#ifndef CFG_RNG_RATE
+	#define CFG_RNG_RATE 0 /* not rate-limited by default */
```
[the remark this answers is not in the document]
sa-kib: yes, sure

core/pta/hwrng.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_OUTPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		EMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	/* Output RNG rate (per second) */
```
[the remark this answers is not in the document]
sa-kib: Linux `optee_rng` driver expects _bytes per second_, `synquacer/rng_pta.c` also defines it in bytes per second (125 in their case).

core/arch/arm/plat-rcar/conf.mk
```diff
@@ -16,11 +16,13 @@ $(call force,CFG_CORE_ARM64_PA_BITS,36)
 # 2. OP-TEE crashes during boot with enabled CFG_CORE_ASLR.
 $(call force,CFG_CORE_ASLR,n)
 
+$(call force,CFG_WITH_SOFTWARE_PRNG,n)
```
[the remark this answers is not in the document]
sa-kib: I agree, it's not necessary. Will remove it.

core/pta/sub.mk
```diff
@@ -10,5 +10,10 @@ srcs-$(CFG_WITH_STATS) += stats.c
 srcs-$(CFG_SYSTEM_PTA) += system.c
 srcs-$(CFG_NXP_SE05X) += scp03.c
 srcs-$(CFG_SCMI_PTA) += scmi.c
+ifeq ($(CFG_WITH_SOFTWARE_PRNG),n)
```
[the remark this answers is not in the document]
sa-kib: agree, should someone enable `CFG_HWRNG_PTA` but mistakenly forget to set `CFG_WITH_SOFTWARE_PRNG=n` the build won't pass anyway.

core/arch/arm/plat-rcar/conf.mk
```diff
@@ -16,11 +16,13 @@ $(call force,CFG_CORE_ARM64_PA_BITS,36)
...
 $(call force,CFG_TEE_CORE_NB_CORE,8)
 
 CFG_TZDRAM_START ?= 0x44100000
 CFG_TZDRAM_SIZE	 ?= 0x03D00000
 CFG_TEE_RAM_VA_SIZE ?= 0x100000
+CFG_RNG_QUALITY ?= 1024
```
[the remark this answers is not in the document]
sa-kib: Ok

core/pta/hwrng.c
```diff
@@ -0,0 +1,82 @@
...
+		return rng_get_info(nParamTypes, pParams);
+	default:
+		break;
+	}
+
+	return TEE_ERROR_NOT_IMPLEMENTED;
+}
+
+pseudo_ta_register(.uuid = PTA_RNG_UUID, .name = PTA_NAME,
+		   .flags = PTA_DEFAULT_FLAGS | TA_FLAG_DEVICE_ENUM,
```
etienne-lms, line 81: I think this service should this be concurrent capable, because there is not reason it should not be. `| TA_FLAG_CONCURRENT`

core/pta/sub.mk
```diff
@@ -10,5 +10,10 @@ srcs-$(CFG_WITH_STATS) += stats.c
...
+ifdef CFG_RNG_QUALITY
```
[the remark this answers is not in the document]
sa-kib: how about using `CFG_HWRNG_QUALITY` instead, to overcome discrepancy?

core/pta/hwrng.c
```diff
@@ -0,0 +1,82 @@
...
+	params[0].value.b = CFG_RNG_QUALITY;
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result invoke_command(void *pSessionContext __unused,
+				 uint32_t nCommandID, uint32_t nParamTypes,
+				 TEE_Param pParams[TEE_NUM_PARAMS])
+{
+	FMSG("command entry point for pseudo-TA \"%s\"", PTA_NAME);
```
etienne-lms, line 66: maybe shorten the trace message :)

core/pta/hwrng.c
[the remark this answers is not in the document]
etienne-lms: The ABI is defined in lib/libutee/include/rng_pta_client.h.
Can copy/past, or remove the inline comments here.

etienne-lms (commented):
> I've missed that PR, otherwise I'd probably re-used it.

It's also nice to compare approaches.

core/pta/sub.mk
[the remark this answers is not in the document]
jforissier: Works for me. Then `CFG_HWRNG_RATE` obviously.

core/pta/hwrng.c
sa-kib: done

core/pta/hwrng.c
sa-kib: shortened, now message is similar to the one from SCMI PTA

mk/config.mk
```diff
@@ -661,3 +661,12 @@ CFG_COMPAT_GP10_DES ?= y
 
 # Defines a limit for many levels TAs may call each others.
 CFG_CORE_MAX_SYSCALL_RECURSION ?= 4
+
+ifeq ($(CFG_HWRNG_PTA),y)
```
jforissier, line 665: Minor: I suggest adding a short description such as:
```
# Pseudo-TA to export hardware RNG output to Normal World
# RNG characteristics are platform specific
```

jforissier (commented):
One very minor comment but other than that I think this loks fine.

@b49020 does this look OK to you? Should plat-synquacer do `$(call force,CFG_HWRNG_PTA,n)` since it has its own implementation?

core/pta/hwrng.c
```diff
@@ -0,0 +1,79 @@
...
+		EMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	e = (uint8_t *)params[0].memref.buffer;
+	if (!e)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	for (i = 0; i < params[0].memref.size; i++)
+		e[i] = hw_get_random_byte();
```
ldebieve, line 35: Hi,
Reviewing this commit, I'm surprised about the usage of the direct call to `hw_get_random_byte()` which, I think,  must not be exposed at that level. In the entire code, the interface that might centralized the call to rng is through the crypto interface `crypto_rng_read()`.

I understand that the PTA interface must only be available while using `RNG_HW`, in that case, why not defining the PTA interface linked to `RNG_HW` defined.
The access to the RNG can be done from TA or PTA using the same crypto interface rather than using `get_hw_random_byte()` that **_MUST_** be implemented in a low level driver as used in a `__weak__` framework function.

Using a single interface will allow to control access to the RNG whatever the implementation and whatever the source as the is only one way to get it.
BR,

jenswi-linaro: > Reviewing this commit, I'm surprised about the usage of the direct call to `hw_get_random_byte()` which, I think, must not be exposed at that level.

What is the problem with `hw_get_random_byte()`? Except that one could wish for a function allowing better efficiency, but that another issue.

ldebieve: Question is more why do we need to call `hw_get_random_byte()` as we have a `crypto_rng_read()` function that is the only one calling the `hw_get_random_byte()` from the hw_rng.c ?

Why not directly use the `crypto_rng_read()` here?

sa-kib: > I understand that the PTA interface must only be available while using RNG_HW, in that case, why not defining the PTA interface linked to RNG_HW defined.

The idea of using `hw_get_random_byte()` specifically is _not_ to let it work with PRNG in any circumstances _and_ keep it simple.
Surely `crypto_rng_read()` can be used here, but in that case we need to add several extra build checks to cover more configuration states (check for condition `CFG_WITH_SOFTWARE_PRNG=n`), which adds some complexity for no really good reason.
Also, should `CFG_WITH_SOFTWARE_PRNG` ever be modified we would have many different places that need to be remembered and taken care of (`core/crypto/sub.mk` and `mk/config.mk` + platform `conf.mk`), so that hwrng PTA does not ever accidentaly use PRNG.

Long story short, it is implemented that way to be more simpler and safer. However I don't know exact purpose of that interface, so if I misused it I'll gladly change it for `crypto_rng_read()`.

mk/config.mk
sa-kib: done

core/pta/hwrng.c
```diff
@@ -0,0 +1,79 @@
...
+				 TEE_Param pParams[TEE_NUM_PARAMS])
+{
+	FMSG(PTA_NAME" command %#"PRIx32" ptypes %#"PRIx32,
+	     nCommandID, nParamTypes);
+
+	switch (nCommandID) {
+	case PTA_CMD_GET_ENTROPY:
+		return rng_get_entropy(nParamTypes, pParams);
+	case PTA_CMD_GET_RNG_INFO:
+		return rng_get_info(nParamTypes, pParams);
```
etienne-lms, line 68: minor: prefer use snake_case labels:
```c
static TEE_Result invoke_command(void *p_session __unused,
				 uint32_t command_id, uint32_t param_types,
				 TEE_Param params[TEE_NUM_PARAMS])
{
	FMSG(PTA_NAME" command %#"PRIx32" ptypes %#"PRIx32,
	     command_id, param_types);

	switch (command_id) {
	case PTA_CMD_GET_ENTROPY:
		return rng_get_entropy(param_types, params);
	case PTA_CMD_GET_RNG_INFO:
		return rng_get_info(nParamTypes, pParams);
	(...)

File: core/pta/hwrng.c (hunk 10 above)
Comment (ldebieve) [reply core/pta/hwrng.c:35]: OK, maybe later discussion around the cryto_rng_interface.

I would say that a simple 
`ifeq ($(CFG_WITH_SOFTWARE_PRNG),n)`
`srcs-$(CFG_HWRNG_PTA) += hwrng.c`
`endif`
probably solve this issue and do not introduce too much complexity?
BR

Comment (sa-kib) [reply core/pta/hwrng.c:35]: > `ifeq ($(CFG_WITH_SOFTWARE_PRNG),n)`
> `srcs-$(CFG_HWRNG_PTA) += hwrng.c`
> `endif`

I've just removed such code by @jforissier request, see [this](https://github.com/OP-TEE/optee_os/pull/4641#discussion_r639561687) comment.
Probably `mk/config.mk` is the place.

File: core/pta/hwrng.c [hunk 12]
```diff
@@ -0,0 +1,79 @@
...
+ * Copyright (C) 2018, Linaro Limited
+ * Copyright (c) 2021, EPAM Systems. All rights reserved.
+ *
+ * Based on plat-synquacer/rng_pta.c
+ *
+ */
+
+#include <kernel/pseudo_ta.h>
+#include <rng_support.h>
+#include <rng_pta_client.h>
```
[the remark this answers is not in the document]
sa-kib: done, 
97c6298

core/pta/hwrng.c
```diff
@@ -0,0 +1,79 @@
...
+				  TEE_Param params[TEE_NUM_PARAMS])
+{
+	uint8_t *e = NULL;
+	uint32_t i = 0;
+
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INOUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		EMSG("bad parameters types: 0x%" PRIx32, types);
```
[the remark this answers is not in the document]
sa-kib: yes, done in  97c6298

core/pta/hwrng.c
sa-kib: sure, done in 97c6298

core/pta/hwrng.c
sa-kib: done, see eddb770 & d9bd599

core/arch/arm/plat-rcar/conf.mk
```diff
@@ -21,6 +21,9 @@ $(call force,CFG_TEE_CORE_NB_CORE,8)
 CFG_TZDRAM_START ?= 0x44100000
 CFG_TZDRAM_SIZE	 ?= 0x03D00000
 CFG_TEE_RAM_VA_SIZE ?= 0x100000
+CFG_HWRNG_QUALITY ?= 1024
+CFG_HWRNG_PTA ?= y
+CFG_WITH_SOFTWARE_PRNG = n
```
etienne-lms, line 26: this statement has no strong effect. use either `?=` or `$(call force,...)`.
Here, I would recommend `CFG_WITH_SOFTWARE_PRNG ?= n`.

core/pta/hwrng.c
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright (C) 2018, Linaro Limited
+ * Copyright (c) 2021, EPAM Systems. All rights reserved.
+ *
+ * Based on plat-synquacer/rng_pta.c
+ *
+ */
+
+#include <kernel/pseudo_ta.h>
+#include <rng_pta_client.h>
+#include <crypto/crypto.h>
```
etienne-lms, line 12: move up. we use alpha order, here:
```c
#include <crypto/crypto.h>
#include <kernel/pseudo_ta.h>
#include <rng_pta_client.h>

File: mk/config.mk [hunk 16]
```diff
@@ -661,3 +661,17 @@ CFG_COMPAT_GP10_DES ?= y
 
 # Defines a limit for many levels TAs may call each others.
 CFG_CORE_MAX_SYSCALL_RECURSION ?= 4
+
+# Pseudo-TA to export hardware RNG output to Normal World
+# RNG characteristics are platform specific
+ifeq ($(CFG_HWRNG_PTA),y)
```
etienne-lms, line 667: add a default value for CFG_HWRNG_PTA before this test:
```
CFG_HWRNG_PTA ?= n
```

core/pta/hwrng.c
```diff
@@ -0,0 +1,74 @@
...
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	e = (uint8_t *)params[0].memref.buffer;
+	if (!e)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	return crypto_rng_read(e, params[0].memref.size);
```
jenswi-linaro, line 33: NAK. This function is not suitable for this even if it under some circumstance might do the right thing.

mk/config.mk
jforissier: Maintainer question: should we adopt this as the preferred way? Undefined also means disabled. But `n` has the advantage of being explicitly saved in the generated build configuration file (`$(out-dir)/conf.mk`) which means that if the default ever changes to `y`, the old configuration could still be used with `make CFG_OPTEE_CONFIG=path/to/saved_conf.mk` and effectively disable the feature.

jforissier:
Azure CI did not run, I have no idea why :-/ but, `make PLATFORM=rcar` fails with:
```
arm-linux-gnueabihf-ld.bfd: out/arm-plat-rcar/core/crypto/rng_hw.o: in function `crypto_rng_read':
optee_os/core/crypto/rng_hw.c:32: undefined reference to `hw_get_random_byte'
make: *** [core/arch/arm/kernel/link.mk:45: out/arm-plat-rcar/core/all_objs.o] Error 1
```

lorc:
> Azure CI did not run, I have no idea why 

Because this PR needs to be rebased onto master. Right now it is based on older revision, where HW RNG for Rcar3 is not merged yet.

mk/config.mk
jenswi-linaro: I'm not sure how useful it is to be able to supply an old `conf.mk`, but I like the idea of assigning default disabled config variables. It makes the description of the variable more natural.

jforissier: OK let's do that.

core/pta/hwrng.c
ldebieve: @jenswi-linaro 
Sorry, I think this rework from @sa-kib is linked to my previous comment and discussion around that function call. Still not clear to me what kind of circumstance you highlight here that may not do the right thing?

jenswi-linaro: `crypto_rng_read()` isn't guaranteed to provide raw hwrng entropy. If you need raw hwrng use `hw_get_random_byte()`.

sa-kib: I'll drop two last fixups eddb770 & d9bd599 so that prev. use of `hw_get_random_byte()` restored.

core/arch/arm/plat-rcar/conf.mk
sa-kib: dropped this statement altogether

core/pta/hwrng.c
sa-kib: done

mk/config.mk
sa-kib: done

sa-kib:
> > Azure CI did not run, I have no idea why
> 
> Because this PR needs to be rebased onto master. Right now it is based on older revision, where HW RNG for Rcar3 is not merged yet.

yes, when review is through I'll rebase on master, then CI is expected to pass.

etienne-lms:
I think you can squash fixup commits and rebase.

I agree with @ldebieve is that `crypto_rng_read()` implementation is a bit strange. It's (almost) the only `crypto_xxx()` helper function being weaked. After looking back the code, i think we could clarify that by removing its `__weak` attribute in _rng_hw.c_ and change all related _core/drivers/..._ sources files to not implement `crypto_rng_read()` but only `hw_get_randmo_byte()`. Yet, `crypto_rng_add_event()` & `crypto_rng_init()` must be also considered. I'll see if I can propose something through a P-R.

For the contradiction, few words about exposing a PRNG through the RNG PTA.
It's not really useful, right, but would it be a real issue if exposed? PTA can explicitly describe a 0 quality entropy. If the PRNG is fair enough for TEE side, it should be also fine for the Linux side. There may already be REE applications that rely on random numbers generated from TAs. In the end, i think exposing `crypto_rng_read()` from PTA even when `CFG_WITH_SOFTWARE_PRNG=y` if something that could be implemented. That said i'm fine with this P-R, it can be discussed in a specific thread.

jenswi-linaro:

Sorry for hijacking the PR for this discussion, perhaps we can continue an issue or PR instead.

> I think you can squash fixup commits and rebase.
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`
> 
> I agree with @ldebieve is that `crypto_rng_read()` implementation is a bit strange. It's (almost) the only `crypto_xxx()` helper function being weaked. After looking back the code, i think we could clarify that by removing its `__weak` attribute in _rng_hw.c_ and change all related _core/drivers/..._ sources files to not implement `crypto_rng_read()` but only `hw_get_randmo_byte()`. Yet, `crypto_rng_add_event()` & `crypto_rng_init()` must be also considered. I'll see if I can propose something through a P-R.

Sure, propose something that we can discuss.

> 
> For the contradiction, few words about exposing a PRNG through the RNG PTA.
> It's not really useful, right, but would it be a real issue if exposed? PTA can explicitly describe a 0 quality entropy. If the PRNG is fair enough for TEE side, it should be also fine for the Linux side. There may already be REE applications that rely on random numbers generated from TAs. In the end, i think exposing `crypto_rng_read()` from PTA even when `CFG_WITH_SOFTWARE_PRNG=y` if something that could be implemented. That said i'm fine with this P-R, it can be discussed in a specific thread.

The Linux kernel has its own SW PRNG, the only thing needed is some seeding.
So the OP-TEE SW PRNG has no advantage over the Linux counterpart.
Giving direct access to the OP-TEE SW PRNG could open up for some attacks if combined with some other bug. So it's an increased risk. If there's no benefit then we should not take that risk.

sa-kib:
Patches squashed and rebased. Took me several turns to also make CI happy.
Thanks for review everybody!

jforissier:
@sa-kib thanks for your determination :wink:

PR #4642, opened 2021-05-26
drivers: caam: fix CMAC data input handling

clementfaure:
Depending on the memory buffer input configuration, the function
caam_dmaobj_sgtbuf_build() might modify the data size to be processed in
the loop.
This case happens sometimes on the imx8mp where the input buffer
physical address in above 32 bits.
This implies reporting the data size re-adjustment when data is saved in
the context buffer.


jforissier:

clementfaure:
Thanks @jforissier. Tag applied

PR #4644, opened 2021-05-26
Ffa updates

jenswi-linaro:

### DIFF
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -56,11 +56,21 @@
 	.endm
 
 FUNC _start , :
+#if defined(CFG_CORE_SEL1_SPMC)
+	/*
+	 * With OP-TEE as SPMC at S-EL1 the SPMD (SPD_spmd) in TF-A passes
+	 * the DTB in x0, pagaeble part in x1 and the rest of the registers
+	 * are unused
+	 */
+	mov	x19, x1		/* Save pagable part */
+	mov	x20, x0		/* Save DT address */
+#else
 	mov	x19, x0		/* Save pagable part address */
 #if defined(CFG_DT_ADDR)
 	ldr     x20, =CFG_DT_ADDR
 #else
 	mov	x20, x2		/* Save DT address */
+#endif
 #endif
 
 	adr	x0, reset_vect_table
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1083,7 +1083,7 @@ static void update_external_dt(void)
 	if (!dt->blob)
 		return;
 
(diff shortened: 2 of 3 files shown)


core/arch/arm/kernel/entry_a64.S
```diff
@@ -56,11 +56,20 @@
 	.endm
 
 FUNC _start , :
+#if defined(CFG_CORE_SEL1_SPMC)
+	/*
+	 * With OP-TEE as SPMC at S-EL1 the SPMD (SPD_spmd) in TF-A passes
+	 * the DTB in x0 and the rest of the registers are unused
+	 */
+	mov	x19, xzr	/* Pagable part not set */
```
jforissier, line 64: Does it mean CFG_CORE_SEL1_SPMC cannot be used with pager enabled?

jforissier (commented):
One question, in any case:

core/arch/arm/kernel/entry_a64.S
jenswi-linaro: You're right. I've changed my mind, I'll enable pager support too.

jenswi-linaro:
Tags applied. Enabled pager in "core: ffa: FF-A specific boot arguments".

core/arch/arm/kernel/entry_a64.S
```diff
@@ -56,11 +56,20 @@
 	.endm
 
 FUNC _start , :
+#if defined(CFG_CORE_SEL1_SPMC)
+	/*
+	 * With OP-TEE as SPMC at S-EL1 the SPMD (SPD_spmd) in TF-A passes
+	 * the DTB in x0 and the rest of the registers are unused
```
jforissier, line 62: `x1` is pageable part

What about the a32 implementation? Shouldn't `r1` play that role too?

jenswi-linaro: You're right. But AArch32 hasn't been tested with FF-A yet, I don't have a working prototype for that. I've just made the corresponding changes.

jenswi-linaro:
Updated arm32 boot args too. Note that this is untested. I tried booting arm32 and it panicked a bit later on something else, but the argument passing seems OK at at least.

core/arch/arm/kernel/entry_a64.S
jforissier: Sounds reasonable.
Back to the a64 code, this comment still needs to be updated "DTB in x0 and the rest of the registers are unused", wrong, x1 is used.

jenswi-linaro: Sorry, I'll fix.

jenswi-linaro:
Fixed up register usage for both arm64 and arm32.

PR #4651, opened 2021-05-28
drivers:caam: Update DRVCRYPT_OID_MB_US_RSADSI macro with correct value

sahilnxp:
This macro forms the HASH OID for MD5 algorithm,

It is defined as:
id-md5 OBJECT IDENTIFIER ::= {
  iso(1) member-body(2) us(840) rsadsi(113549) digestAlgorithm(2) 5
}

According to OpenSSL,
iso(1) member-body(2) us(840) rsadsi(113549) digestAlgorithm(2) 5
part is encoded as
0x2A,0x86,0x48,0x86,0xF7,0x0D,0x02,0x05

OpenSSL Link for reference: https://bit.ly/3hVZ7Is

But in this case it was being formed as
0x2A,0x86,0x48,0x86,0x48,0x02,0x05 which was wrong.

### DIFF
--- a/core/drivers/crypto/crypto_api/include/drvcrypt_asn1_oid.h
+++ b/core/drivers/crypto/crypto_api/include/drvcrypt_asn1_oid.h
@@ -42,7 +42,7 @@
  * us(840) rsadsi(113549)
  */
 #define DRVCRYPT_OID_MB_US	  "\x86\x48"
-#define DRVCRYPT_OID_MB_US_RSADSI DRVCRYPT_OID_MB_US "\x86\x48"
+#define DRVCRYPT_OID_MB_US_RSADSI DRVCRYPT_OID_MB_US "\x86\xF7\x0D"
 
 /*
  * ISO Identified organization


jenswi-linaro:
It's good to use OpenSSL as an example, but I'd prefer referring to a specific standard as the authority.

clementfaure:
Hello,

The RSADSI OID is the following : https://oidref.com/1.2.840.113549
For nodes > 128, simple way to do OID hexadecimal encoding is the following (not from me):

113549d = 0x1BB8D
0x1BB8D & 0x7F = 0xD
((0x1BB8D >> 7) & 0x7F) | 0x80 = 0xF7
((0x1BB8D >> 14) & 0x7F) | 0x80 = 0x86

-> x86xF7x0D

sahilnxp:
Updated tags of @clementfaure 
Thanks @clementfaure for OID to Hex conversion logic

@jenswi-linaro Added RFC link in commit message.

jenswi-linaro:

sahilnxp:
Tags updated.

Issue #4659, opened 2021-06-02
Storage benchmark xtest test cases are failing.


sahilnxp:
Getting following error when running  xtest -t benchmark 1001

```
root@localhost:~# xtest -t benchmark 1001
Test ID: 1001
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# benchmark
#
######################################################

* benchmark_1001 TEE Trusted Storage Performance Test (WRITE)
E/TC:?? 00
E/TC:?? 00 TA panicked with code 0xffff0006
E/LD:   Status of TA f157cda0-550c-11e5-a6fa-0002a5d5c51b
E/LD:    arch: aarch64
E/LD:   region  0: va 0x40004000 pa 0xfc26c000 size 0x002000 flags rw-s (ldelf)
E/LD:   region  1: va 0x40006000 pa 0xfc26e000 size 0x008000 flags r-xs (ldelf)
E/LD:   region  2: va 0x4000e000 pa 0xfc276000 size 0x001000 flags rw-s (ldelf)
E/LD:   region  3: va 0x4000f000 pa 0xfc277000 size 0x004000 flags rw-s (ldelf)
E/LD:   region  4: va 0x40013000 pa 0xfc27b000 size 0x001000 flags r--s
E/LD:   region  5: va 0x40014000 pa 0xfc299000 size 0x001000 flags rw-s (stack)
E/LD:   region  6: va 0x40016000 pa 0x00001000 size 0x011000 flags r-xs [0]
E/LD:   region  7: va 0x40027000 pa 0x00012000 size 0x00c000 flags rw-s [0]
E/LD:    [0] f157cda0-550c-11e5-a6fa-0002a5d5c51b @ 0x40016000
E/LD:   Call stack:
E/LD:    0x0000000040018800
E/LD:    0x00000000400162dc
E/LD:    0x0000000040019a54
E/LD:    0x00000000400166ec
optee-test/host/xtest/benchmark_1000.c:119: run_chunk_access_test(cmd, data_size_table[i], chunk_size, &                                  records[i]) has an unexpected value: 0xffff3024 = TEE_ERROR_TARGET_DEAD, expected 0x0 = TEEC_SUCCESS
E/TC:?? 00
E/TC:?? 00 TA panicked with code 0xffff0006
E/LD:   Status of TA f157cda0-550c-11e5-a6fa-0002a5d5c51b
E/LD:    arch: aarch64
[...]

PR #4660, opened 2021-06-02
core: syscall_storage_obj_open: Add TEE_DATA_FLAG_OVERWRITE to valid flags

sahilnxp:
During Storage benchmark testing.

Now here we try to Open persistent object with following flags:

TEE_DATA_FLAG_ACCESS_READ | TEE_DATA_FLAG_ACCESS_WRITE |
TEE_DATA_FLAG_ACCESS_WRITE_META | TEE_DATA_FLAG_OVERWRITE

which lands into syscall_storage_obj_open(), where it checks the flags
which were passed during opening the object and valid flags

TEE_DATA_FLAG_ACCESS_READ | TEE_DATA_FLAG_ACCESS_WRITE |
TEE_DATA_FLAG_ACCESS_WRITE_META | TEE_DATA_FLAG_SHARE_READ |
TEE_DATA_FLAG_SHARE_WRITE

TEE_DATA_FLAG_OVERWRITE flag was missing from valid_flags

Fixes: https://github.com/OP-TEE/optee_os/issues/4659

### DIFF
--- a/core/tee/tee_svc_storage.c
+++ b/core/tee/tee_svc_storage.c
@@ -239,7 +239,8 @@ TEE_Result syscall_storage_obj_open(unsigned long storage_id, void *object_id,
 					  TEE_DATA_FLAG_ACCESS_WRITE |
 					  TEE_DATA_FLAG_ACCESS_WRITE_META |
 					  TEE_DATA_FLAG_SHARE_READ |
-					  TEE_DATA_FLAG_SHARE_WRITE;
+					  TEE_DATA_FLAG_SHARE_WRITE |
+					  TEE_DATA_FLAG_OVERWRITE;
 	const struct tee_file_operations *fops =
 			tee_svc_storage_file_ops(storage_id);
 	struct ts_session *sess = ts_get_current_session();


jforissier:
No, `TEE_DATA_FLAG_OVERWRITE` is not a valid flag for `TEE_OpenPersistentObject()` (and therefore `syscall_storage_obj_open()`). See GP spec section 5.7.1.

sahilnxp:
Thanks @jforissier, My bad, I just checked GP spec and it is not there.
So we need to solve it in Storage benchmark TA then way [here](https://github.com/OP-TEE/optee_test/blob/master/ta/storage_benchmark/benchmark.c#L343).

We should not OpenPersistent Object with `TEE_DATA_FLAG_OVERWRITE` flag

Will that be the right approach ?

jforissier:
@sahilnxp it seems `TEE_CreatePersistentObject()` sould be called instead of `TEE_OpenPersistentObject()`.

sahilnxp:
> @sahilnxp it seems `TEE_CreatePersistentObject()` sould be called instead of `TEE_OpenPersistentObject()`.

@jforissier  Creation of object was done in `prepare_test_file()` [here ](https://github.com/OP-TEE/optee_test/blob/a4ab95742fd4718bb0b0e2178d9e1acd5197b42f/ta/storage_benchmark/benchmark.c#L336)just before calling  `TEE_OpenPersistentObject()`.

Do we still want to create the object again ?

jforissier:
@sahilnxp oh I missed that. No, probably not. Have you tried simply removing `TEE_DATA_FLAG_OVERWRITE`?

sahilnxp:
@jforissier yes, I will try with that and update you.

sahilnxp:
@jforissier , Yes, by removing flag `TEE_DATA_FLAG_OVERWRITE`, it works fine.
I will raise a pull request in optee_test for this.

sahilnxp:
Pull request has been raised in optee_test: https://github.com/OP-TEE/optee_test/pull/516

sahilnxp:
This is not the right approach to solve the issue, will be taken care by OP-TEE/optee_test#516 in OP-TEE test.

PR #4664, opened 2021-06-03
Small fixes

jenswi-linaro:

### DIFF
--- a/core/arch/arm/kernel/thread_a32.S
+++ b/core/arch/arm/kernel/thread_a32.S
@@ -206,7 +206,10 @@ END_FUNC thread_save_state
  *			    unsigned long a2, unsigned long a3)
  */
 FUNC thread_smc , :
+	push	{r4-r7}
+UNWIND(	.save	{r4-r7})
 	smc	#0
+	pop	{r4-r7}
 	bx	lr
 END_FUNC thread_smc
 
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -769,12 +769,10 @@ static void gen_malloc_add_pool(struct malloc_ctx *ctx, void *buf, size_t len)
 					(~(SizeQuant - 1))) +
 				sizeof(struct bhead) * 2;
 
-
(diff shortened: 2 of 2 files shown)


jforissier:
For "libutils: fix range check in gen_malloc_add_pool()":

jforissier:
For "core: arm32: update thread_smc() for SMCCC v1.2":

> but the normal function calling convention requires
> r4-r7 to be preserved so save a and restore them.

So this is more of a bug fix even with SMCCC < v1.2, isn't it? Why mention v1.2 at all?

jenswi-linaro:
> For "core: arm32: update thread_smc() for SMCCC v1.2":
> 
> > but the normal function calling convention requires
> > r4-r7 to be preserved so save a and restore them.
> 
> So this is more of a bug fix even with SMCCC < v1.2, isn't it? Why mention v1.2 at all?

No, it's with the v1.2 calling convention that r3-r7 could be used to return values. But even then only for specific functions, notably the FF-A family of functions.

jforissier:
I think I got it, it's the `smc #0` that would potentially (when SMCCC v1.2 is used) return a value in r4-r7 and thus change them which the caller of `thread_smc()` wouldn't expect so you're preserving them always.
Why push/pop r3-r7 and not r4-r7 though?

jenswi-linaro:
I looks like you answered your own question there. :-) Let me know if anything is unclear. ;-)

jenswi-linaro:
Updated the commit "core: arm32: update thread_smc() for SMCCC v1.2" with r4-r7 instead.

jforissier:
And what about r8, (r9), r10 and r11? Don't they need preserving, too?

jenswi-linaro:
> And what about r8, (r9), r10 and r11? Don't they need preserving, too?

No, they are required to be preserved together with `lr`.

jforissier:
For "core: arm32: update thread_smc() for SMCCC v1.2":

(please fix typo in commit description: "so save ~~a~~ and restore")

jenswi-linaro:
Tag applied. Typo fixed.

PR #4671, opened 2021-06-07
core_mmu: fix implicit behavior of core_mmu_add_mapping()

a-rybakov:
In core_mmu_add_mapping() requested physical address
rounded up/down to granule size (0x100000), which leads
to establishing of virtual mappings with overlapped
physical counterparts. If two virtual mappings overlaps
due to such roundings, then following phys_to_virt() can
implicitly return result of virtual address from
unexpected mapping. This patch fix such behavior by
returning virtual address of newly established mapping.

Signed-off-by: Anton Rybakov <a.rybakov@omp.ru>


jenswi-linaro:
Looks good.

etienne-lms (approved):
ditto

PR #4674, opened 2021-06-07
libtomcrypt: fix rsa key generation public exponent

cneveux:
Fix libtomcrypt crypto_acipher_gen_rsa_key() to call rsa_make_key_bn_e()
API to generate a RSA key with a public exponent upto 256 bits.

RSA standard specify that public exponent e can be between 65537 (included)
and 2^256 (excluded).


jforissier:

PR #4679, opened 2021-06-10
core: REE FS: introduce CFG_REE_FS_ALLOW_RESET

etienne-lms:
New boolean configuration switch CFG_REE_FS_ALLOW_RESET that, when
enabled, will make OP-TEE OS to allow REE FS content to be reset in
the Linux filesystem even when RPMB FS is enabled and already stores a
REE FS rollback protection hash. This switch is intended to test purpose
where REE FS can be wiped because the device flash memory was programmed
with brand new build artifacts.

### DIFF
--- a/core/tee/tee_ree_fs.c
+++ b/core/tee/tee_ree_fs.c
@@ -4,6 +4,7 @@
  */
 
 #include <assert.h>
+#include <config.h>
 #include <kernel/mutex.h>
 #include <kernel/panic.h>
 #include <kernel/thread.h>
@@ -497,9 +498,21 @@ static TEE_Result open_dirh(struct tee_fs_dirfile_dirh **dirh)
 		return res;
 
 	res = tee_fs_dirfile_open(false, hashp, &ree_dirf_ops, dirh);
-	if (res == TEE_ERROR_ITEM_NOT_FOUND)
-		res = tee_fs_dirfile_open(true, NULL, &ree_dirf_ops, dirh);
+	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
+		if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET) && hashp) {
+			DMSG("Clear REE FS hash in RPMB");
+			res = rpmb_fs_ops.truncate(ree_fs_rpmb_fh, 0);
+			if (res) {
+				DMSG("Clear REE FS hash failed: %#"PRIx32, res);
+				res = TEE_ERROR_SECURITY;
+				goto out;
+			}
+		}
 
+		res = tee_fs_dirfile_open(true, NULL, &ree_dirf_ops, dirh);
+	}
+
+out:
 	if (res)
 		rpmb_fs_ops.close(&ree_fs_rpmb_fh);
 
--- a/core/tee/tee_ree_fs.c
+++ b/core/tee/tee_ree_fs.c
@@ -498,12 +498,19 @@ static TEE_Result open_dirh(struct tee_fs_dirfile_dirh **dirh)
 		return res;
 
 	res = tee_fs_dirfile_open(false, hashp, &ree_dirf_ops, dirh);
+
 	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
-		if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET) && hashp) {
-			DMSG("Clear REE FS hash in RPMB");
-			res = rpmb_fs_ops.truncate(ree_fs_rpmb_fh, 0);
-			if (res) {
-				DMSG("Clear REE FS hash failed: %#"PRIx32, res);
+		if (hashp) {
+			if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET)) {
+				DMSG("dirf.db not found, clear hash in RPMB");
+				res = rpmb_fs_ops.truncate(ree_fs_rpmb_fh, 0);
+				if (res) {
+					DMSG("Can't clear hash: %#"PRIx32, res);
+					res = TEE_ERROR_SECURITY;
+					goto out;
+				}
+			} else {
+				DMSG("dirf.db file not found");
 				res = TEE_ERROR_SECURITY;
 				goto out;
 			}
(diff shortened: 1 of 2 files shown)


core/tee/tee_ree_fs.c
```diff
@@ -497,9 +498,19 @@ static TEE_Result open_dirh(struct tee_fs_dirfile_dirh **dirh)
 		return res;
 
 	res = tee_fs_dirfile_open(false, hashp, &ree_dirf_ops, dirh);
-	if (res == TEE_ERROR_ITEM_NOT_FOUND)
+	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
 		res = tee_fs_dirfile_open(true, NULL, &ree_dirf_ops, dirh);
 
+		if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET) && !res && hashp) {
+			DMSG("Clear REE FS hash in RPMB");
+			res = rpmb_fs_ops.truncate(ree_fs_rpmb_fh, 0);
```
jenswi-linaro, line 506: We have a race here. But perhaps we don't care since it's testing code only?

etienne-lms: i don't wan't to introduce bugs in the core. Sorry I don't see the reace.

Would you prefer we wipe the hash data in RPMB before we create the file in REE FS? Like:
```c
	res = tee_fs_dirfile_open(false, hashp, &ree_dirf_ops, dirh);
	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
		if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET) && hashp) {
			DMSG("Clear REE FS hash in RPMB");
			res = rpmb_fs_ops.truncate(ree_fs_rpmb_fh, 0);
			if (res) {
				DMSG("Clear REE FS hash failed: %#"PRIx32, res);
				res = TEE_ERROR_SECURITY;
				goto out;
			}
		}

		res = tee_fs_dirfile_open(true, NULL, &ree_dirf_ops, dirh);
	}
```

jenswi-linaro: Yes something like that.
The race I was after is if execution is stopped after the dirfile is created and we restart we wouldn't be able to reach this code path again. It could of course be fixed by the user removing the dirfile again, but I'd prefer doing it right as it's easier to understand the code then.

etienne-lms: Ok, agree. Let's clear the hash first.
Note that removing REE FS content here will work only if OP-TEE has not yet accessed it since it booted up.

etienne-lms: By the way, I have another change I was to propose, i think it simplifies the sequence. 
When CFG_REE_FS_ALLOW_RESET=n (default config),  there is a REE FS hash in the RPMB and dirf.db is not found, then we should report the error straight, not attempt to create back dirfs.db file and report the hash mismatch later.

I'll append the commit in this P-R.

jenswi-linaro: That's going to result in a chicken-egg problem, right?

etienne-lms: > That's going to result in a chicken-egg problem, right?

Talking about commit "core: REE FS: report earlier unexpected REE FS reset"?
The idea is to report the mismatch as early as possible. When looking into this part, it looks strange OP-TEE requests the creation of a file and later fails on its content.

jenswi-linaro: Sorry, I misunderstood what you meant.

jenswi-linaro:

core/tee/tee_ree_fs.c
```diff
@@ -498,12 +498,19 @@ static TEE_Result open_dirh(struct tee_fs_dirfile_dirh **dirh)
...
+			if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET)) {
+				DMSG("Clear REE FS hash in RPMB");
+				res = rpmb_fs_ops.truncate(ree_fs_rpmb_fh, 0);
+				if (res) {
+					DMSG("Can't clear hash: %#"PRIx32, res);
+					res = TEE_ERROR_SECURITY;
+					goto out;
+				}
+			} else {
+				DMSG("Secure storage corruption detected");
```
jforissier, line 513: - Should this one be `EMSG()` perhaps?
- I suggest mentioning `dirf.db` to help troubleshooting, because that file name is hidden quite far from here. How about "Secure storage corruption detected (no dirf.db)"? (ignoring the long line >80 characters)

etienne-lms: I used `DMSG()` as around line 427 (`DMSG("Secure storage corruption detected");`)

Maybe `DMSG("Expected dirf.db file not found");` ? (fits in 80chars)

jforissier: > I used `DMSG()` as around line 427 (`DMSG("Secure storage corruption detected");`)

OK, let's be consistent then.

> Maybe `DMSG("Expected dirf.db file not found");` ? (fits in 80chars)

Good, you can even drop "Expected".

etienne-lms:
Change trace messages and log level:
`IMSG()` when resetting REE FS content despite hash in RPMB
`DMSG()` when reporting REE FS corruption when _dirf.db_ is missing

core/tee/tee_ree_fs.c
```diff
@@ -502,15 +502,15 @@ static TEE_Result open_dirh(struct tee_fs_dirfile_dirh **dirh)
 	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
 		if (hashp) {
 			if (IS_ENABLED(CFG_REE_FS_ALLOW_RESET)) {
-				DMSG("Clear REE FS hash in RPMB");
+				IMSG("dirf.db not found, clear hash in RPMB");
```
jforissier, line 505: DMSG?

jforissier:
> Change trace messages and log level:
> `IMSG()` when resetting REE FS content despite hash in RPMB
> `DMSG()` when reporting REE FS corruption when _dirf.db_ is missing

I don't really understand the rationale for IMSG vs. DMSG?

etienne-lms:
maybe subjective.

When corruption is reported, TEE_ERROR_SECURITY is returned and client gets the info. The trace is more debug info since we have a return status but could me moved to error level, if preferred.

When corruption is ignored, there's no feedback but trace message. Using info level makes it a bit more likely to be seen by tester.

I have no strong opinion here. I made the change to trig the discussion (and maybe waste your time :sorry:).
If you prefer `DMSG()` everywhere, i'm fine.

jforissier:
> When corruption is ignored, there's no feedback but trace message. Using info level makes it a bit more likely to be seen by tester.

I agree that a trace is valuable in this case, but I'd rather keep it debug rather than info.

etienne-lms:
ok

PR #4698, opened 2021-06-22
libmbedtls: Support dynamic check CRT parameters

lubing521:
Keymasster VTS need enable mbedtls with CRT.
But some case of the Xtest 4006 no provide CRT
This will cause OP-tee OS check it fail. in order
to support this scenario work well,libmedtls need
frendly deal with RSA with CRT or not.

Signed-off-by: Lu Bing <lubing@eswin.com>
Tested-by: Lu Bing <lubing@eswin.com>

### DIFF
--- a/lib/libmbedtls/core/sm2-dsa.c
+++ b/lib/libmbedtls/core/sm2-dsa.c
@@ -139,6 +139,7 @@ out:
 	mbedtls_mpi_free(&r);
 	mbedtls_mpi_free(&s);
 	mbedtls_mpi_free(&tmp);
+	mbedtls_ecp_group_free(&grp);
 	return res;
 }
 
@@ -262,5 +263,6 @@ out:
 	mbedtls_mpi_free(&t);
 	mbedtls_mpi_free(&eprime);
 	mbedtls_mpi_free(&R);
+	mbedtls_ecp_group_free(&grp);
 	return res;
 }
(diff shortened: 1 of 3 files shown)


jenswi-linaro:
The approach to check if CRT parameters are present makes sense. But basic fixes are needed.
1. I believe only a single commit is needed
2. You should try to follow the MbedTLS coding style a bit closer
3. I would really help if this was just a cherry-pick from upstream MbedTLS repository, that is, the patch was already accepted upstream

Normally when making changes to MbedTLS or any other library kept in sync with upstream we upstream the change first and then bring it into OP-TEE once the solution has been accepted. This allows us to keep a minimal delta compared to upstream. Keeping a small delta is key to be able upgrade to a later release. There's also the benefit that the patches has been reviewed by experts on the subject.

lubing521:
> The approach to check if CRT parameters are present makes sense. But basic fixes are needed.
> 
> 1. I believe only a single commit is needed
> 2. You should try to follow the MbedTLS coding style a bit closer
> 3. I would really help if this was just a cherry-pick from upstream MbedTLS repository, that is, the patch was already accepted upstream
> 
> Normally when making changes to MbedTLS or any other library kept in sync with upstream we upstream the change first and then bring it into OP-TEE once the solution has been accepted. This allows us to keep a minimal delta compared to upstream. Keeping a small delta is key to be able upgrade to a later release. There's also the benefit that the patches has been reviewed by experts on the subject.

Thank you very much for your advice.
I will try do it on upstream.

jenswi-linaro:
> I will try do it on upstream.

Thank you, much appreciated.

lubing521:
> > I will try do it on upstream.
> 
> Thank you, much appreciated.

This is upstream status.

https://github.com/ARMmbed/mbedtls/pull/4721

PR #4709, opened 2021-06-28
ta: export CFG_VIRTUALIZATION

jforissier:
xtest 1013 needs to know when OP-TEE is built with CFG_VIRTUALIZATION=y
[1], therefore export this variable to TAs.

Link: [1] https://github.com/OP-TEE/optee_test/commit/444e15178dce

### DIFF
--- a/ta/ta.mk
+++ b/ta/ta.mk
@@ -40,6 +40,7 @@ ta-mk-file-export-vars-$(sm) += CFG_TA_MCOUNT
 ta-mk-file-export-vars-$(sm) += CFG_CORE_TPM_EVENT_LOG
 ta-mk-file-export-add-$(sm) += CFG_TEE_TA_LOG_LEVEL ?= $(CFG_TEE_TA_LOG_LEVEL)_nl_
 ta-mk-file-export-vars-$(sm) += CFG_TA_BGET_TEST
+ta-mk-file-export-vars-$(sm) += CFG_VIRTUALIZATION
 
 # Expand platform flags here as $(sm) will change if we have several TA
 # targets. Platform flags should not change after inclusion of ta/ta.mk.


lorc:
This worries me a bit. Virtualization support was designed in a way that it should now interfere with rest of the system. So, do we really need to tell TAs that they will be built against OP-TEE with virtualization support? Does this implies that they potentially could not work OP-TEE built in other way?

Tests 1013 and 1034 fail not because of virtualization, but because of limited heap. For instance, I rather not skip this two tests on my Rcar Gen3 boards. This platform provides 64 MBs of secure memory, and there never was any problems with the mentioned tests. 

I believe, the right solution for this problem is to either increase secure memory on QEMUv8 or to skip the tests if heap size is not big enough.

jforissier:
> This worries me a bit. Virtualization support was designed in a way that it should now interfere with rest of the system. So, do we really need to tell TAs that they will be built against OP-TEE with virtualization support? Does this implies that they potentially could not work OP-TEE built in other way?
> 
> Tests 1013 and 1034 fail not because of virtualization, but because of limited heap. For instance, I rather not skip this two tests on my Rcar Gen3 boards. This platform provides 64 MBs of secure memory, and there never was any problems with the mentioned tests.

I agree. My intention was to use this information only for testing (`xtest`), but that is definitely a bad idea, because once we export that information we can't control how it is used and people will abuse it. Moreover, it is a clear example of testing for something (virtualization) when we really mean to test for something else (the heap size) which is more difficult to probe... clearly not good :-/

> I believe, the right solution for this problem is to either increase secure memory on QEMUv8

Fundamentally, the issue is not platform-specific.

> or to skip the tests if heap size is not big enough.

That's the best option IMO. Perhaps we should just ignore out of memory errors in xtest 1013.2? The rationale being, 1013.1 should pass on all platforms, while 1013.2 might fail on memory constrained platforms but only with "out of memory" errors.
I will propose something in `optee_test`.

Thanks for the feedback!

jforissier:
Closing this, replaced with https://github.com/OP-TEE/optee_test/pull/527.

PR #4713, opened 2021-06-29
hikey, hikey960: disable CFG_SECURE_DATA_PATH by default

jforissier:
Since linaro-swg/linux.git branch optee [1] was rebased onto kernel
v5.12, Secure Data Path is broken in xtest [2] because the client side
is based on the ION allocator, which was removed from the kernel.

Therefore, disable SDP support by default.

Link: [1] https://github.com/linaro-swg/linux/tree/optee-v5.12-20210628
Link: [2] https://github.com/OP-TEE/optee_test/blob/3.13.0/host/xtest/regression_1000.c#L1220-L1263


jenswi-linaro:

PR #4720, opened 2021-07-01
Miscellaneous small fixes

clementleger:
This series contains minor fixes:
- Typo fix in scmi code
- Removal of 0x prefix to node address in unit name as per devicetree specification
- Add option CFG_EXTERNAL_DTB_OVERLAY_ERASE_AT_BOOT which erase the device tree overlay at boot time

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -653,6 +653,7 @@ static int add_dt_overlay_fragment(struct dt_descriptor *dt, int ioffs)
 
 static int init_dt_overlay(struct dt_descriptor *dt, int __maybe_unused dt_size)
 {
+#if !defined(CFG_EXTERNAL_DTB_OVERLAY_ERASE_AT_BOOT)
 	int fragment;
 	int ret;
 
@@ -662,6 +663,7 @@ static int init_dt_overlay(struct dt_descriptor *dt, int __maybe_unused dt_size)
 			dt->frag_id += 1;
 		return ret;
 	}
+#endif
 
 #ifdef CFG_DT_ADDR
 	return fdt_create_empty_tree(dt->blob, dt_size);
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -887,7 +887,7 @@ static int add_res_mem_dt_node(struct dt_descriptor *dt, const char *name,
 	}
 
 	ret = snprintf(subnode_name, sizeof(subnode_name),
-		       "%s@0x%" PRIxPA, name, pa);
+		       "%s@%" PRIxPA, name, pa);
 	if (ret < 0 || ret >= (int)sizeof(subnode_name))
 		DMSG("truncated node \"%s@0x%"PRIxPA"\"", name, pa);
 	offs = fdt_add_subnode(dt->blob, offs, subnode_name);
--- a/core/drivers/scmi-msg/clock.c
+++ b/core/drivers/scmi-msg/clock.c
@@ -319,7 +319,7 @@ static void scmi_clock_describe_rates(struct scmi_msg *msg)
 	} else if (status == SCMI_NOT_SUPPORTED) {
 		unsigned long triplet[3] = { 0, 0, 0 };
 
-		/* Platform may support min§max/step triplet description */
+		/* Platform may support min/max/step triplet description */
 		status =  plat_scmi_clock_rates_by_step(msg->channel_id,
 							clock_id, triplet);
 		if (status == SCMI_SUCCESS) {


core/arch/arm/kernel/boot.c
```diff
@@ -653,6 +653,7 @@ static int add_dt_overlay_fragment(struct dt_descriptor *dt, int ioffs)
 
 static int init_dt_overlay(struct dt_descriptor *dt, int __maybe_unused dt_size)
 {
+#if !defined(CFG_EXTERNAL_DTB_OVERLAY_ERASE_AT_BOOT)
```
jenswi-linaro, line 656: This flag doesn't seem to erase anything, it seems to affect if something is done or not though.
The config flag should have a name which describes what what it does, preferably without negative logic.
Please add some description of this flag in `mk/config.mk`.
If possible favor `if (IS_ENABLED(...))` over `#if ...`.

clementleger: Agreed, is CFG_SKIP_OVERLAY_REUSE more understandable to you ?

jenswi-linaro: What would the REUSE part stand for here? (From an OP-TEE point of view)

clementleger: Hum, this is indeed missleading. Currently, if an overlay is given in FDT_ADDR and CFG_CTB_OVERLAY is set, then the DTB overlay will be reused. I guess CFG_IGNORE_BOOT_OVERLAY might be more accurate in this case.

clementleger: But as you said, this is using negative logic so CFG_USE_BOOT_OVERLAY should be more appropriate.

jenswi-linaro: Wouldn't it work to just set `CFG_EXTERNAL_DTB_OVERLAY=n`? Or there's something that needs to be done still?

etienne-lms: Currently, with `CFG_EXTERNAL_DTB_OVERLAY=y`, OP-TEE uses `FDT_ADDR` as a DTB overleay memory area:
* if there is already an overlay: new overlays are added.
* if no valid DTB overlay header found: a DTB overlay format is created and overlays added.
`CFG_EXTERNAL_DTB_OVERLAY=n`, `FDT_ADDR` is not accessed as an DTB overlay area.

I  think @clementleger wants `CFG_EXTERNAL_DTB_OVERLAY=y` but always discarding previous content of `FDT_ADDR` area.

core/arch/arm/kernel/boot.c
```diff
@@ -887,7 +887,7 @@ static int add_res_mem_dt_node(struct dt_descriptor *dt, const char *name,
 	}
 
 	ret = snprintf(subnode_name, sizeof(subnode_name),
-		       "%s@0x%" PRIxPA, name, pa);
+		       "%s@%" PRIxPA, name, pa);
```
etienne-lms, line 890: fix also the trace at line 892.

etienne-lms (commented):
"drivers: scmi-msg: fix typo"
 "core: dt: remove 0x in reserved memory node unit name" (with comment addressed)

core/arch/arm/kernel/boot.c
clementleger: @etienne-lms: Yes exactly, currently the overlay always grows when resetting because the DRAM keeps its content.
I would like OP-TEE to recreate the dt overlay at each boot to avoid that.

core/arch/arm/kernel/boot.c
clementleger: Done

core/arch/arm/kernel/boot.c
jenswi-linaro: Thanks for the clarification. To me it seems that the DT configuration is getting messier and messier. I think we're lacking a list of the different use cases and which flags used to implement them in each case. As it looks now a small change may have unexpected consequences for a missed use case.

jforissier: @jenswi-linaro I agree, I think such a list would be welcome in `mk/config.mk` and also all combinations that are not explicitly useful should cause an error.

clementleger: After looking at config.mk, I read the following:

```
# Device Tree Overlay support.
# This define enables support for an OP-TEE provided DTB overlay.
# One of two modes is supported in this case:
# 1. Append OP-TEE nodes to an existing DTB overlay located at CFG_DT_ADDR or
#    passed in arg2
# 2. Generate a new DTB overlay at CFG_DT_ADDR
# A subsequent boot stage must then merge the generated overlay DTB into a main
# DTB using the standard fdt_overlay_apply() method.
CFG_EXTERNAL_DTB_OVERLAY ?= n
```

This seems to imply there is 2 modes available but there is not acutally since the logic only check if there is already something at CFG_DT_ADDR.  So I guess a new config variable is actually needed for my use-case (which should be covered by mode 2 described in the previous comment). IMHO, fixing the comment in config.mk and adding a new config variable is needed to really handle both modes.

clementleger: Should I continue adding this feature and adding a comment for it in `mk/config.mk` ?

etienne-lms: `CFG_EXTERNAL_DTB_OVERLAY` has some legacy in optee_os. Maybe we could keep it for what it describes and introduce a new alternative `CFG_GENERATE_DTB_OVERLAY` which could be described as:
```
# Device Tree Overlay support.
# CFG_EXTERNAL_DTB_OVERLAY allows to append a DTB overlay into an existing
# external DTB. The overlay is created when no valid DTB overlay is found.
# CFG_GENERATE_DTB_OVERLAY allows to create a DTB overlay at external
# DTB location.
# External DTB location (pysical address) is provided either by boot
# argument arg2 or from CFG_DT_ADDR if defined.
# A subsequent boot stage can then merge the generated overlay DTB into a main
# DTB using the standard fdt_overlay_apply() method.
CFG_EXTERNAL_DTB_OVERLAY ?= n
CFG_GENERATE_DTB_OVERLAY ?= n

ifeq (y-y,$(CFG_EXTERNAL_DTB_OVERLAY)-$(CFG_GENERATE_DTB_OVERLAY))
$(error CFG_EXTERNAL_DTB_OVERLAY and CFG_GENERATE_DTB_OVERLAY are exclusive)
endif

Comment (clementleger) [reply core/arch/arm/kernel/boot.c]: Looks pretty clear to me and clearly describe the goal of each options ! Thanks for your comment.

**Comment by clementleger** [comment]
New version:
- Add new CFG_GENERATE_DTB_OVERLAY config option
- Add intermediate variable CFG_USE_DTB_OVERLAY to avoid using if defined(CFG_EXTERNAL_DTB_OVERLAY) || if defined(CFG_GENERATE_DTB_OVERLAY) in boot.c
- Remove space after "," in config.mk

File: mk/config.mk [hunk 3]
```diff
@@ -423,14 +423,22 @@ CFG_DT ?= n
...
+# argument arg2 or from CFG_DT_ADDR if defined.
+# A subsequent boot stage can then merge the generated overlay DTB into a main
 # DTB using the standard fdt_overlay_apply() method.
 CFG_EXTERNAL_DTB_OVERLAY ?= n
+CFG_GENERATE_DTB_OVERLAY ?= n
+
+ifeq (yy,$(CFG_EXTERNAL_DTB_OVERLAY)$(CFG_GENERATE_DTB_OVERLAY))
+$(error CFG_EXTERNAL_DTB_OVERLAY and CFG_GENERATE_DTB_OVERLAY are exclusive)
+endif
+CFG_USE_DTB_OVERLAY ?= $(call cfg-one-enabled,CFG_EXTERNAL_DTB_OVERLAY \
```
jforissier, line 440: It is OK to define a derived variable like this, but it should be prefixed by an underscore (`_CFG_USE_DTB_OVERLAY`). The rationale being, it makes it more clear what is supposed to be set by the user or not; and also they are not exported to `$(out-dir)/conf.mk`.

clementleger: Ok thanks, I wondered if you had such thing. Should I move it to core.mk or let it in config.mk ?

jforissier: You can leave it here. Also better use `:=` rather than `?=` since the reason we're using `?=` everywhere is to consider environment variables (`CFG_FOO=x make` vs. `make CFG_FOO=x` should do the same). That is not applicable here.

clementleger:
New version:
- Switch to _CFG_USE_DTB_OVERLAY to make it a "hidden" config option and use :=

etienne-lms (commented):
Comments for commit "core: dt: add option to generate DTB overlay at boot".

mk/config.mk
```diff
@@ -423,14 +423,22 @@ CFG_DT ?= n
...
+# CFG_GENERATE_DTB_OVERLAY allows to create a DTB overlay at external
+# DTB location.
+# External DTB location (physical address) is provided either by boot
+# argument arg2 or from CFG_DT_ADDR if defined.
+# A subsequent boot stage can then merge the generated overlay DTB into a main
 # DTB using the standard fdt_overlay_apply() method.
 CFG_EXTERNAL_DTB_OVERLAY ?= n
+CFG_GENERATE_DTB_OVERLAY ?= n
+
+ifeq (yy,$(CFG_EXTERNAL_DTB_OVERLAY)$(CFG_GENERATE_DTB_OVERLAY))
```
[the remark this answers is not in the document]
clementleger: Ok, I wanted to keep it coherent with what I saw in config.mk. I'll modify that.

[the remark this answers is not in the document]
etienne-lms: _mk/config.mk_ is not very clean IMO. Fix proposal in https://github.com/OP-TEE/optee_os/pull/4832.

core/arch/arm/kernel/boot.c
```diff
@@ -656,11 +656,13 @@ static int init_dt_overlay(struct dt_descriptor *dt, int __maybe_unused dt_size)
...
+	if (!IS_ENABLED(CFG_GENERATE_DTB_OVERLAY)) {
+		ret = fdt_check_header(dt->blob);
+		if (!ret) {
+			fdt_for_each_subnode(fragment, dt->blob, 0)
+				dt->frag_id += 1;
+			return ret;
+		}
 	}
 
 #ifdef CFG_DT_ADDR
```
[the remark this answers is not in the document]
etienne-lms: I don't think it really makes sense the secure boot stage passes a DTB address in r2/x2 when the area is not filled with a DTB. But maybe the OP-TEE implementation can allow that.

mk/config.mk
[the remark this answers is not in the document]
jforissier: IMO, 'yy' is not a valid value for something that is enabled. Only 'y' means enabled. Anything else means disabled. I know it is not exactly mentioned like this in https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#boolean-configuration-variables but if we don't follow strictly this assumption, many 'is enabled' tests will fail, I mean `ifneq (y,$(CFG_XXX))`. Is there a test somewhere that allows 'yy' for enabled?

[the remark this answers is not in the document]
etienne-lms: I mean building with `CFG_EXTERNAL_DTB_OVERLAY=yy` and `CFG_GENERATE_DTB_OVERLAY` not defined will build with both disabled, whish is expected but would failon build error:
`$(error CFG_EXTERNAL_DTB_OVERLAY and CFG_GENERATE_DTB_OVERLAY are exclusive)`
which is not really relevant here, imho.

core/arch/arm/kernel/boot.c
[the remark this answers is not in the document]
clementleger: Actually, that would make sense to avoid hardcoding the external device tree location with CFG_DT_ADDR.

mk/config.mk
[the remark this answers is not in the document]
jforissier: > I mean building with `CFG_EXTERNAL_DTB_OVERLAY=yy` and `CFG_GENERATE_DTB_OVERLAY` not defined will build with both disabled, whish is expected but would failon build error:
`$(error CFG_EXTERNAL_DTB_OVERLAY and CFG_GENERATE_DTB_OVERLAY are exclusive)`
which is not really relevant here, imho.

+1

core/arch/arm/kernel/boot.c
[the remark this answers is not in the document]
etienne-lms: There is some lelgacy in imx for 2 paltforms enabling `CFG_EXTERNAL_DTB_OVERLAY` and setting `CFG_DT_ADDR`.
Do you always expect to use boot argument arg#2 (arm register r2/x2) to pass DTB location to use with `CFG_GENERATE_DTB_OVERLAY=y`?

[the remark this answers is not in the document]
clementleger: What I meant is to support both (either specified in r2 or CFG_DT_ADDR). Currently, only the `ifdef` forbid using this feature without CFG_DT_ADDR. Removing the ifdef would allow both r2/CFG_DT_ADDR to work with CFG_EXTERNAL_DTB_OVERLAY/CFG_GENERATE_DTB_OVERLAY I guess. Do you see any potential problem here ?

clementleger:
Looks like there is a false positive error triggered by checkdiff:

```
WARNING: IS_ENABLED(_CFG_USE_DTB_OVERLAY) is normally used as IS_ENABLED(CFG__CFG_USE_DTB_OVERLAY)
#87: FILE: core/arch/arm/kernel/boot.c:857:
+	if (IS_ENABLED(_CFG_USE_DTB_OVERLAY)) {
```

jforissier:
@clementleger yes, the Linux IS_ENABLED doesn't expect the full config variable name, only the part after the prefix (CONFIG_ by default, CFG_ in optee_os). We cannot easily fix our version of IS_ENABLED because as is the case here thz prefix is not always CFG_.

etienne-lms (commented):
> What I meant is to support both (either specified in r2 or CFG_DT_ADDR). Currently, only the ifdef forbid using this feature without CFG_DT_ADDR. Removing the ifdef would allow both r2/CFG_DT_ADDR to work with CFG_EXTERNAL_DTB_OVERLAY/CFG_GENERATE_DTB_OVERLAY I guess. Do you see any potential problem here ?

No, i don't see issues. I'm fine with that, code is simpler.
As for the legacy, only 2 flavors do enable `CFG_EXTERNAL_DTB_OVERLAY` and they both do configure `CFG_DT_ADDR` so no regression foreseen.

etienne-lms (commented):

jenswi-linaro:

PR #4732, opened 2021-07-07
FF-A: Add secure partition build support

imre-kis-arm:
A separate build environment called SP dev kit is being introduced by this pull request for enabling [FF-A secure partition](https://developer.arm.com/documentation/den0077/latest) builds. The system is based on the TA dev kit and it provides the entry point, the interface implementations for libutils and the necessary make environment for building secure partitions.


jenswi-linaro (commented):
There's quite a bit copied from TA dev kit which I'm not sure will ever be used for a SP. Everything which isn't needed should be removed since it's just confusing to have there.

I'd like to know the long term strategy for this. In OP-TEE we care about backwards compatibility so whatever we bring in we should be prepared to maintain as long as it's relevant. When things are changed we make sure there's a smooth transition that hopefully shouldn't even be noticed by most users. So some changes are quite hard to make. If an ABI is changed we must have a way of dealing with the old ABI too, that's one reason for many of the quirks with the TA format.

keys/default_sp.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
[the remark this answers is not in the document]
imre-kis-arm: Would it be a better solution to have `default_ta.pem` renamed to `default_ts.pem` and use it for both TAs and SPs?

core/arch/arm/arm.mk
```diff
@@ -210,12 +210,25 @@ supported-ta-targets += ta_arm64
 endif
 endif
 
+ifeq (,$(supported-sp-targets))
+supported-sp-targets = sp_arm32
```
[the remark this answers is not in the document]
imre-kis-arm: We are planning to add support for AArch32 platforms in CQ1'22 according to [our roadmap](https://developer.trustedfirmware.org/w/trusted-services/roadmap/). This feature mainly targets low-end platforms.
The SP dev kit now contains both AArch32 and AArch64 support despite we don't have the AArch32 SPMC in place.
As the SP dev kit is based on the TA dev kit where the AArch32 support is already available it was easier to keep this part of the system because we are planning to use it in the near future.

core/arch/arm/arm.mk
```diff
@@ -308,8 +321,96 @@ ta-mk-file-export-add-ta_arm64 += COMPILER_ta_arm64 ?= $$(COMPILER)_nl_
...
+
+sp-mk-file-export-add-sp_arm64 += CROSS_COMPILE64 ?= $$(CROSS_COMPILE)_nl_
+sp-mk-file-export-add-sp_arm64 += CROSS_COMPILE_sp_arm64 ?= $$(CROSS_COMPILE64)_nl_
+sp-mk-file-export-add-sp_arm64 += COMPILER ?= gcc_nl_
+sp-mk-file-export-add-sp_arm64 += COMPILER_sp_arm64 ?= $$(COMPILER)_nl_
+sp-mk-file-export-add-sp_arm64 += PYTHON3 ?= python3_nl_
+endif
+
+# Set cross compiler prefix for each submodule
+$(foreach sm, core $(ta-targets), $(eval CROSS_COMPILE_$(sm) ?= $(CROSS_COMPILE$(arch-bits-$(sm)))))
```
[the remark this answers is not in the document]
imre-kis-arm: Ack

sp/include/optee_sp_internal_api.h
```diff
@@ -0,0 +1,67 @@
...
+	uint16_t timeMid;
+	uint16_t timeHiAndVersion;
+	uint8_t clockSeqAndNode[8];
+};
+
+struct optee_sp_head {
+	struct optee_sp_uuid uuid;
+	uint32_t stack_size;
+	uint32_t flags;
+	uint64_t depr_entry;
```
etienne-lms, line 44: I don't think you need this deprecated field from `struct ta_head`.

sp/include/optee_sp_internal_api.h
```diff
@@ -0,0 +1,67 @@
...
+/*
+ * The file describes the API between the SP dev kit and the SP. It contains the
+ * function prototypes that the user code must define.
+ * The SP code base should also contain a header file named
+ * "optee_sp_user_defines.h" for passing the following definitions to the SP dev
+ * kit:
+ * * OPTEE_SP_HEAP_SIZE: Heap size in bytes
+ * * OPTEE_SP_UUID: UUID of the SP as an sp_uuid structure
+ * * OPTEE_SP_STACK_SIZE: Stack size in bytes
+ * * OPTEE_SP_FLAGS: SP attributes (currently none available, set to zero)
```
etienne-lms, line 23: Do SPs plan to have attributes?

keys/default_sp.pem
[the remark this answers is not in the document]
etienne-lms: If SPs authentication relies on the very same public use for TAs and imported in core rodata section, then there should not be 2 different PEM files. I that case renaming _default_ta.pem_ looks legitimate to me.

I think for now, SPs cannot be loaded/authenticated using a core build-in public key. I would suggest to use _default_ta.pem_ for now and post a specific commit/P-R to either introduce a specific SP pub key (or key pair) file  _deafult_sp.pem_ or rename _default_ta.pem_ to _deafult_ts.pem_, with all necessary updates in optee_os (embed the pub key, use it to verify SP binary blob, ...) when SPs will need them.

core/arch/arm/arm.mk
```diff
@@ -308,8 +321,96 @@ ta-mk-file-export-add-ta_arm64 += COMPILER_ta_arm64 ?= $$(COMPILER)_nl_
...
+sp_arm32-platform-cppflags += $(arm32-platform-cppflags)
+sp_arm32-platform-cflags += $(arm32-platform-cflags)
+sp_arm32-platform-cflags += $(platform-cflags-optimization)
+sp_arm32-platform-cflags += $(platform-cflags-debug-info)
+sp_arm32-platform-cflags += -fpic
+
+# Thumb mode doesn't support function graph tracing due to missing
+# frame pointer support required to trace function call chain. So
+# rather compile in ARM mode if function tracing is enabled.
+ifeq ($(CFG_FTRACE_SUPPORT),y)
```
etienne-lms, line 337: ftrace (from libutils) in TAs currently mandates some specific syscall support from libutee.
Maybe somewhere it should be stated that `CFG_FTRACE_SUPPORT` must be disabled when generating the SP dev kit.

core/arch/arm/arm.mk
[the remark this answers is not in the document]
jenswi-linaro: OK, thanks.

sp/include/optee_sp_internal_api.h
imre-kis-arm: Currently there's an implicit dependency between `struct ta_head` and `struct sp_head` as [ldelf always uses](https://github.com/OP-TEE/optee_os/blob/master/ldelf/ta_elf.c#L1073) `struct ta_head` regardless of if it's loading a TA or an SP. So essentially the two structs should be the same.
One solution would be to pass a parameter to ldelf so it can diverge according to the type (TA or SP) of the entity being loaded.
However the preferred way of loading SPs in SPMs is to avoid ELF loader and simply have the SPs as raw binaries. So in the near future we are planning to omit ldelf from the SP load process so I don't think it's worth the effort of modifying the ldelf. As far as I know @jenswi-linaro is aware of these discussions.
If this sounds reasonable I would leave it this way for now. I can also add a comment of this explanation to the code.

sp/include/optee_sp_internal_api.h
imre-kis-arm: The [same](https://github.com/OP-TEE/optee_os/pull/4732#discussion_r691247152) goes to the `flags` field but I agree that `OPTEE_SP_FLAGS` is useless as it can be replaced by setting the field to 0 directly in `sp_header.c`.

core/arch/arm/arm.mk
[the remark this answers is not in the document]
imre-kis-arm: Done

Makefile
```diff
@@ -97,6 +97,25 @@ endef
...
+endef
+$(foreach t, $(sp-targets), $(eval $(call build-sp-target, $(t))))
+
+# Build SPs included in this git
+define build-sp
+sp-mk-file := $(1)
+include sp/mk/build-sp.mk
+endef
+$(foreach t, $(wildcard sp/*/sp.mk), $(eval $(call build-sp,$(t))))
+endif
```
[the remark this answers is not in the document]
imre-kis-arm: Done

sp/include/optee_sp_internal_api.h
imre-kis-arm: > Then why not use struct ta_head for now?
`struct ta_head` is defined in `user_ta_header.h` of libutee and it depends on other files of libutee. I didn't want to involve libutee in the build process of SPs as they wouldn't use much of it.

> Funny thing, we went basically the opposite way for TAs... but I do understand the use cases are a bit different.
Yes, I guess it's mostly because of other SPM solutions use the raw binary format so it is better the have OP-TEE aligned with this method as well.

keys/default_sp.pem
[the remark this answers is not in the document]
imre-kis-arm: I've removed the signing and encrypting part of the SP build process as we don't have a way to verify/load signed/encrypted SPs and according to the discussions with my team we are not intending to have it in the near future.

core/arch/arm/arm.mk
imre-kis-arm: I've added a check and an error message for SP dev kit build & `CFG_FTRACE_SUPPORT=y` cases.

Makefile
```diff
@@ -97,6 +97,25 @@ endef
 $(foreach t, $(sort $(wildcard ta/*/user_ta.mk)), $(eval $(call build-user-ta,$(t))))
 endif
 
+# Platform/arch config is supposed to assign the targets
+sp-targets ?= invalid
+default-sp-target ?= $(firstword $(sp-targets))
```
etienne-lms, line 102: these 3 lines should go with the conditioned instructions below.

sp/arch/arm/sp.ld.S
```diff
@@ -0,0 +1,89 @@
...
+		*(.gnu.linkonce.r.*)
+		*(.rodata .rodata.*)
+	}
+	.ARM.extab : { *(.ARM.extab* .gnu.linkonce.armextab.*) }
+	/* .ARM.exidx is sorted, so has to go in its own output section.  */
+	PROVIDE_HIDDEN(__exidx_start = .);
+	.ARM.exidx : { *(.ARM.exidx* .gnu.linkonce.armexidx.*) }
+	PROVIDE_HIDDEN(__exidx_end = .);
+        .ctors : { *(.ctors) }
+        .dtors : { *(.dtors) }
```
etienne-lms, line 44: indentation

sp/sp.mk
```diff
@@ -0,0 +1,133 @@
+include mk/cleanvars.mk
+
+# Set current submodule (used for module specific flags compile result etc)
+sm := $(sp-target)
+sm-$(sm) := y
+
+# Setup compiler for this sub module
+COMPILER_$(sm)		?= $(COMPILER)
```
etienne-lms, line 8: prefer a space char than these tabs

sp/include/optee_sp_internal_api.h
```diff
@@ -0,0 +1,66 @@
...
+ * The SP code base should also contain a header file named
+ * "optee_sp_user_defines.h" for passing the following definitions to the SP dev
+ * kit:
+ * * OPTEE_SP_HEAP_SIZE: Heap size in bytes
+ * * OPTEE_SP_UUID: UUID of the SP as an sp_uuid structure
+ * * OPTEE_SP_STACK_SIZE: Stack size in bytes
+ */
+
+/*
+ * The file defines the functions and variables that other
```
etienne-lms, line 26: something is missing in the sentence

sp/include/optee_sp_internal_api.h
etienne-lms: minor: maybe rename `depr_entry` to `reserved`.

sp/mk/sp_dev_kit.mk
```diff
@@ -0,0 +1,115 @@
...
+ifneq (1, $(words $(BINARY)))
+$(error You must specify the BINARY variable)
+endif
+
+ifneq ($O,)
+out-dir := $O
+else
+out-dir := .
+endif
+link-out-dir := $(out-dir)	# backward compat
```
etienne-lms, line 19: i think you don't need this legacy $(link-out-dir).

etienne-lms (commented):
minor comments

sp/arch/arm/sp.ld.S
imre-kis-arm: Done

sp/sp.mk
imre-kis-arm: Done

sp/include/optee_sp_internal_api.h
imre-kis-arm: This is some leftover comment which I've now removed. Lines 14-23 explain the purpose of this file properly.

sp/include/optee_sp_internal_api.h
imre-kis-arm: Done

sp/mk/sp_dev_kit.mk
imre-kis-arm: Done

Makefile
imre-kis-arm: Done

etienne-lms (commented):
@jenswi-linaro, are you ok @imre-kis-arm squashes commits?

jenswi-linaro:
> @jenswi-linaro, are you ok @imre-kis-arm squashes commits?

OK to squash commits. However, at the beginning of this review I asked this:
> There's quite a bit copied from TA dev kit which I'm not sure will ever be used for a SP. Everything which isn't needed should be removed since it's just confusing to have there.
>
> I'd like to know the long term strategy for this. In OP-TEE we care about backwards compatibility so whatever we bring in we should be prepared to maintain as long as it's relevant. When things are changed we make sure there's a smooth transition that hopefully shouldn't even be noticed by most users. So some changes are quite hard to make. If an ABI is changed we must have a way of dealing with the old ABI too, that's one reason for many of the quirks with the TA format.

@imre-kis-arm what's the plan? Do you think you've removed everything which isn't needed?

core/arch/arm/arm.mk
```diff
@@ -210,12 +210,25 @@ supported-ta-targets += ta_arm64
 endif
 endif
 
+ifeq (,$(supported-sp-targets))
+supported-sp-targets = sp_arm32
+ifeq ($(CFG_ARM64_core),y)
+supported-sp-targets += sp_arm64
+endif
+endif
```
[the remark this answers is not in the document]
imre-kis-arm: Done

ruchi393:
> > @jenswi-linaro, are you ok @imre-kis-arm squashes commits?
> 
> OK to squash commits. However, at the beginning of this review I asked this:
> 
> > There's quite a bit copied from TA dev kit which I'm not sure will ever be used for a SP. Everything which isn't needed should be removed since it's just confusing to have there.
> > I'd like to know the long term strategy for this. In OP-TEE we care about backwards compatibility so whatever we bring in we should be prepared to maintain as long as it's relevant. When things are changed we make sure there's a smooth transition that hopefully shouldn't even be noticed by most users. So some changes are quite hard to make. If an ABI is changed we must have a way of dealing with the old ABI too, that's one reason for many of the quirks with the TA format.
> 
> @imre-kis-arm what's the plan? Do you think you've removed everything which isn't needed?

@imre-kis-arm , any updates on this ?

imre-kis-arm:
After considering different ways to progress we ended up with the conclusion of having the build system of FF-A SPs in [trusted-services](https://git.trustedfirmware.org/TS/trusted-services.git/) is better. As I've mentioned in an earlier comment there's an ongoing discussion of having a standardized binary format for FF-A SPs so all SPMCs can handle the same binaries. Because of this it is not worth merging this pull request to optee_os as it introduces an intermediate format which will be eventually replaced by the standard one.
Unfortunately the standardization process started when this pull request was already here and the plan was not clear until now, so sorry for the extra work.
If you find this explanation reasonable and agree with it I will close this pull request.

jforissier:
@imre-kis-arm sounds fine as far as I'm concerned.

jenswi-linaro:
Sounds good to me, thanks for the update.

Issue #4725, opened 2021-07-06
TEE_MACCompareFinal fail with NIST test vector


josebeo2016:
1. I have a test vector which is:
```
Count = 94
Klen = 32 
Mlen = 48
Tlen = 10
Key = 7cfc086d10659d7cb9247208358dd82c03b8dbd8233223231df218e2448f4a79
Msg = a3af8f99703a601086c2a1ffe55fde4c2c4153dbff8d6601ab68743c0d50d021b0b3099535ba6c40f866ca3ff0df7c19
Mac = 9d713a1944b8eb649584
```
from NIST CMACAESgen256.rsp
After input the test vector into TEE_MACCompareFinal, I got TEE_ERROR_MAC_INVALID

I try to investigate TEE_MACCompareFinal() source code, It's seem like there is something wrong with the following code:

```
if (operation->operationState != TEE_OPERATION_STATE_ACTIVE) {
		res = TEE_ERROR_BAD_PARAMETERS;
		goto out;
	}

	res = TEE_MACComputeFinal(operation, message, messageLen, computed_mac,
				  &computed_mac_size);
	if (res != TEE_SUCCESS)
		goto out;

	if (computed_mac_size != macLen) {
		res = TEE_ERROR_MAC_INVALID;
		goto out;
	}

	if (consttime_memcmp(mac, computed_mac, computed_mac_size) != 0) {
		res = TEE_ERROR_MAC_INVALID;
		goto out;
	}
```

The function compares the computed_mac_size to the macLen before comparing the memory of the two buffers. In our case, if we input the custom mac's length, e.g., Tlen = 10, then the function always return invalid.

jenswi-linaro:
The `TEE_MACComputeFinal()` expects a certain length of the computed MAC, not a user supplied one. This has been our interpretation of the GlobalPlatform specification and as far as I know this is also what the GlobalPlatform tests expect. The GP tests are more about the API than the algorithms so it has quite few test vectors for some algorithms and may perhaps have some blind spots.

It might be worth checking with GP if the behaviour you're looking for is expected to be supported by the TEE Internal Core API Specification.

josebeo2016:
[...]

PR #4735, opened 2021-07-07
TEE_MACCompareFinal() updates

jforissier:
Addressing a small non-compliance with the spec (first commit) and issue #4725 (second commit).

### DIFF
--- a/lib/libutee/tee_api_operations.c
+++ b/lib/libutee/tee_api_operations.c
@@ -1265,6 +1265,11 @@ TEE_Result TEE_MACCompareFinal(TEE_OperationHandle operation,
 	if (res != TEE_SUCCESS)
 		goto out;
 
+	if (macLen > computed_mac_size) {
+		res = TEE_ERROR_BAD_PARAMETERS;
+		goto out;
+	}
+
 	if (computed_mac_size != macLen) {
 		res = TEE_ERROR_MAC_INVALID;
 		goto out;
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

lib/libutee/tee_api_operations.c
```diff
@@ -1265,12 +1265,12 @@ TEE_Result TEE_MACCompareFinal(TEE_OperationHandle operation,
 	if (res != TEE_SUCCESS)
 		goto out;
 
-	if (computed_mac_size != macLen) {
-		res = TEE_ERROR_MAC_INVALID;
+	if (macLen > computed_mac_size) {
```
vchong, line 1268: Should we consider the case for `macLen = 0`? If some implementations expect an error for it while others don't, then perhaps just leave it to the TA to handle?

jforissier: Ideally this case needs to be clarified by GP too, but as you said since it is not useful in practice it can be handled by the TA. The current code would return success I suppose, which totally makes sense IMO.

etienne-lms:

jforissier:
I have some news from GP. I'm afraid we will have to discard these changes except for the first commit ("libutee: TEE_MACCompareFinal(): panic if input size is too large"). GP confirmed that TEE_ERROR_MAC_INVALID shall be returned when macLen is smaller than expected. The function shall return OK only when the provided MAC is **strictly identical** to the computed one, and that means same size, same content. The use cases with a smaller size (the Tlen parameter in the NIST stuff) are unsupported, and would need proper updates to the GP spec should sufficient interest be shown.

So I'm closing the PR, I will open a new one with the first commit only -- thanks for reviewing!

jenswi-linaro:
Thanks for looking into this.

Issue #4673, opened 2021-06-07
cannot find libgcc.a


aleksandarn-ee:
Getting the following error when trying to build an external TA in Yocto:

```
arm-XXX-linux-gnueabi-ld.bfd: cannot find libgcc.a: No such file or directory
```

The error occur when trying to compile the TA part. The `Makefile` and `sub.mk` are almost identical to the ones use in the in-tree TAs. The patch provided [here](http://git.yoctoproject.org/cgit/cgit.cgi/meta-freescale/tree/recipes-security/optee/optee-os-qoriq/0001-allow-setting-sysroot-for-libgcc-lookup.patch?h=master) has been applied and the in-tree TAs are built without any problems. Any clues where the problem lies and how to solve it?

aleksandarn-ee:
For anyone facing the same problem, patching OPTEE OS with the patch from [here](http://git.yoctoproject.org/cgit/cgit.cgi/meta-freescale/tree/recipes-security/optee/optee-os-qoriq/0001-allow-setting-sysroot-for-libgcc-lookup.patch?h=master) and adding the following variables to `EXTRA_OEMAKE` in the external TA's bitbake recipe solves the problem:

```
LIBGCC_LOCATE_CFLAGS="--sysroot=${STAGING_DIR_TARGET}"
comp-cflagscore="-sysroot=${STAGING_DIR_TARGET}"
```

Issue #4485, opened 2021-03-19
Xilinx ZCU102 building issues ( cannot find libgcc.a)


AsmX0r:
Hello!

I'm currently working on a board Xilinx ZCU102 and I'm trying to compile OP-TEE for it. I have thoroughly followed [the official documentation](https://optee.readthedocs.io/en/latest/building/devices/zynqmp.html) and I'm facing a similar issue raised by [this issue](https://github.com/OP-TEE/optee_os/issues/3950), I get the error `aarch64-xilinx-linux-ld.bfd: cannot find libgcc.a` while executing the command `make -f zynqmp.mk`. I have applied the recommendation of [the linked thread](https://forums.xilinx.com/t5/Embedded-Linux/libgcc-a-not-found-when-building-OPTEE-on-ZCU102/m-p/1121532#M43341) of the previously mentioned issue referenced as the solution, but it did not work, which the original author seems to also indicates in that external thread. I have contacted the author of the issue in the related issue, but since it has been raised a couple of months ago, this person does not have any track of the resolution of the compilation error.

Here is a breakdown of my steps to bring more lights on where the issue is encountered. Based on [the official documentation](https://optee.readthedocs.io/en/latest/building/devices/zynqmp.html), I have completed all the steps successfully before `make -f zynqmp.mk`. My current setup is Ubuntu 16.04, with Petalinux 2018.2, with the latest version of OP-TEE. I execute the following commands, which are part of `make -f zynqmp.mk`:

- `make -f zynqmp.mk petalinux-create`
- Execute the recommended procedure to include gcc into Petalinux, as mentioned in [the external link](https://forums.xilinx.com/t5/Embedded-Linux/libgcc-a-not-found-when-building-OPTEE-on-ZCU102/m-p/1121532#M43341) of the previously mentioned issue. Everything behaves as mentioned.
- `make -f zynqmp.mk petalinux-config`
- `make -f zynqmp.mk petalinux-build`; this is where the error is raised. I have attached a more comprehensive log at the end of the issue to have more insights.

I know this may be hard for you to diagnosis the issue, since it seems to be bound to Petalinux. This is why it would be great if @ricardosalveti (listed as maintainer of the board) could give me some tips on that issue. :)

Many thanks for your help!

```
$ make -f zynqmp.mk petalinux-build
[INFO] building project
[INFO] sourcing bitbake
INFO: bitbake petalinux-user-image
[...]

PR #2853, opened 2019-02-27
    Allow to set the sysroot for libgcc lookup

mimok:
OP-TEE compilation fails when OE/bb is used:
    libgcc.a cannot be find when gcc.mk is executed.
    This commit allows to set the sysroot to solve this problem.
    Sysroot can be defined using the variable LIBGCC_LOCATE_CFLAGS.
    e.g. LIBGCC_LOCATE_CFLAGS=--sysroot=<new_sysroot>

    This patch originally come from https://git.linaro.org/openembedded/meta-linaro.git/

Issue #4188, opened 2020-10-27
Not simple to use a CC that needs arguments passed


rossburton:
When building in Yocto/OpenEmbedded, the cross compiler is "poisoned" so you need to tell it where the sysroot is.  This is because the sysroot is *per package* for isolation purposes.

Typically, this is done by setting `CC = "arch-prefix-gcc --sysroot=/path/to/sysroot`.  Build systems use that compiler, and everything works.

However, `CC` is explicitly not used.  optee-os sets `CC$(sm)` to `$(CROSS_COMPILE)gcc`, and there are many variations of `sm`: `core`, `ldelf`, `ta_arm32`, `ta_arm64` in optee-os alone.  Also, the same applies to `CXX`.

Can there be a way to set a default `CC` for all builds that overrides `$(CROSS_COMPILE)gcc`?

rossburton:
Add `avb`, `pkcs11` and `trusted_keys` to the list of candidates.

Further context: the primary failure is any linking to `libgcc` or`libgcc_eh`, as those as `CC` where the files are and without the sysroot, it has no idea.

Further fun: optee-os can find `libgcc` as we can set `CFLAGS` to include the sysroot, but `libgcc_eh`is found using `CXX` and there is no easy flag for that, it's the `$(sm)` combinational explosion again.

Emantor:
FWIW, the meta-linaro layer contains a patch to the OP-TEE build system and explicitly passes the  libgcc location, see [here](https://git.linaro.org/openembedded/meta-linaro.git/tree/meta-optee/recipes-security/optee/optee-os/0001-allow-setting-sysroot-for-libgcc-lookup.patch).
I remember doing a similar operation for `libgcc_eh`, but would of couse prefer an upstream solution.

jenswi-linaro:
Setting `CROSS_COMPILE_core`, `CROSS_COMPILE_ta_arm32`, and `CROSS_COMPILE_ta_arm64` to the different cross compiler you need should be enough. `CROSS_COMPILE_ldelf` is assigned from `CROSS_COMPILE_core` and `CROSS_COMPILE_avb` and friends are assigned from `CROSS_COMPILE_ta_arm32` or `CROSS_COMPILE_ta_arm64`.

rossburton:
[...]

PR #4737, opened 2021-07-08
mk: LIBGCC_LOCATE_CFLAGS allows setting sysroot path for GCC

etienne-lms:
OP-TEE OS compilation fails with Yocto like build environment
as libgcc.a cannot be find when gcc.mk is executed [1]. A trick
used  was to set comp-cflagscore='--sysroot=/path/to/sysroot'
at OP-TEE OS build. This is no more sufficient as CXXFLAGS
(comp-cxxflagscore=...) may also be needed.

This change implements LIBGCC_LOCATE_CFLAGS as a better way to
path sysroot info for the expected purpose, as proposed in [2] and
[3], recently mentioned in [1] and used in several distributions.

Build environment are expected to set path when needed:
LIBGCC_LOCATE_CFLAGS='--sysroot=/path/to/sysroot'

Link: [1] https://github.com/OP-TEE/optee_os/issues/4673
Link: [2] https://github.com/OP-TEE/optee_os/pull/2853
Link: [3] https://github.com/OP-TEE/optee_os/issues/4485#issuecomment-803025139

### DIFF
--- a/mk/gcc.mk
+++ b/mk/gcc.mk
@@ -13,11 +13,11 @@ nostdinc$(sm)	:= -nostdinc -isystem $(shell $(CC$(sm)) \
 			-print-file-name=include 2> /dev/null)
 
 # Get location of libgcc from gcc
-libgcc$(sm)  	:= $(shell $(CC$(sm)) $(CFLAGS$(arch-bits-$(sm))) \
+libgcc$(sm)  	:= $(shell $(CC$(sm)) $(LIBGCC_LOCATE_CFLAGS) $(CFLAGS$(arch-bits-$(sm))) \
 			-print-libgcc-file-name 2> /dev/null)
-libstdc++$(sm)	:= $(shell $(CXX$(sm)) $(CXXFLAGS$(arch-bits-$(sm))) $(comp-cxxflags$(sm)) \
+libstdc++$(sm)	:= $(shell $(CXX$(sm)) $(LIBGCC_LOCATE_CFLAGS) $(CXXFLAGS$(arch-bits-$(sm))) $(comp-cxxflags$(sm)) \
 			-print-file-name=libstdc++.a 2> /dev/null)
-libgcc_eh$(sm)	:= $(shell $(CXX$(sm)) $(CXXFLAGS$(arch-bits-$(sm))) $(comp-cxxflags$(sm)) \
+libgcc_eh$(sm)	:= $(shell $(CXX$(sm)) $(LIBGCC_LOCATE_CFLAGS) $(CXXFLAGS$(arch-bits-$(sm))) $(comp-cxxflags$(sm)) \
 			-print-file-name=libgcc_eh.a 2> /dev/null)
 
 # Define these to something to discover accidental use


jforissier:
So this is supposed to be used as `make LIBGCC_LOCATE_CFLAGS=--sysroot=/path/to/sysroot`, right? I think we would be better off with `SYSROOT_FLAGS` which would be more clear, note LIBGCC_ is not so good since it is used for libstdc++ and libgcc_eh too. When using a specific sysroot, shouldn't `--sysroot=...` be given to all compilation commands?

jforissier:
And what about 32 vs 64-bits? We potentially have two toolchains involved in one build (`CROSS_COMPILE32`, `CROSS_COMPILE64`).

vesajaaskelainen:
We have been using this patch for a while and seems to work. We are building for both 32 bit and 64 bits.

Logic for that originated from:
https://git.linaro.org/openembedded/meta-linaro.git/tree/meta-optee/recipes-security/optee/optee-os/0001-allow-setting-sysroot-for-libgcc-lookup.patch

But in addition to this we need to define more stuff.

For optee-os:
```
OPTEE_ARCH ?= "arm32"
OPTEE_ARCH_armv7a = "arm32"
OPTEE_ARCH_aarch64 = "arm64"

# Define device tree blob maximum size as 128kB by default,
# which is twice the size limit set in optee-os source code.
OPTEE_DTB_MAX_SIZE ?= "0x20000"

EXTRA_OEMAKE = " \
    CFG_DTB_MAX_SIZE=${OPTEE_DTB_MAX_SIZE} \
...
    CFG_EXTERNAL_DTB_OVERLAY=y \
    CFG_NS_ENTRY_ADDR= \
    CROSS_COMPILE64=${HOST_PREFIX} \
    CROSS_COMPILE=${HOST_PREFIX} \
    LDFLAGS= \
    NOWERROR=1 \
    O=${B} \
    PLATFORM=${PLATFORM} \
    PLATFORM_FLAVOR=${PLATFORM_FLAVOR} \
...
"

EXTRA_OEMAKE_append_aarch64 = " \
    CFG_ARM64_core=y \
    ta-targets=ta_arm64 \
"

do_compile () {
    unset LDFLAGS
    export CFLAGS="${CFLAGS} --sysroot=${STAGING_DIR_HOST}"
    oe_runmake -C ${S} all CFG_TEE_TA_LOG_LEVEL=0 CFG_WITH_USER_TA=y
}
```

Where `OPTEE_ARCH` is used later to determine what to do for output binaries.

optee-test recipe also needs matching exports:
```
EXTRA_OEMAKE = "\
    V=1 \
...
    CFG_PKCS11_TA=y \
"

do_compile () {
    if [ ${DEFAULTTUNE} = "aarch64" ];then
        export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_arm64/
        export ARCH=arm64
    else
        export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_arm32/
        export ARCH=arm
    fi
    export OPTEE_CLIENT_EXPORT=${STAGING_DIR_HOST}/usr
    export CROSS_COMPILE_HOST=${HOST_PREFIX}
    export CROSS_COMPILE_TA=${HOST_PREFIX}
    export CROSS_COMPILE=${HOST_PREFIX}
    export OPTEE_OPENSSL_EXPORT=${STAGING_INCDIR}
    export OPTEE_PKCS11_TA_HEADERS=${STAGING_INCDIR}
    export LIBGCC_LOCATE_CFLAGS=--sysroot=${STAGING_DIR_HOST}

    oe_runmake
}
```

And finally a patch is needed for optee-client:
```patch
diff --git a/flags.mk b/flags.mk
index 71f3d18..af13495 100644
--- a/flags.mk
+++ b/flags.mk
@@ -6,7 +6,7 @@ CROSS_COMPILE   ?= arm-linux-gnueabihf-
 CC              ?= $(CROSS_COMPILE)gcc
 AR		?= $(CROSS_COMPILE)ar
 
-CFLAGS          := -Wall -Wbad-function-cast -Wcast-align \
+CFLAGS          += -Wall -Wbad-function-cast -Wcast-align \
 		   -Werror-implicit-function-declaration -Wextra \
 		   -Wfloat-equal -Wformat-nonliteral -Wformat-security \
 		   -Wformat=2 -Winit-self -Wmissing-declarations \
-- 
2.21.0
```
and in optee-client recipe:
```
EXTRA_OEMAKE_append = " \
    ARCH=${@bb.utils.contains('DEFAULTTUNE', 'aarch64', 'arm64', 'arm', d)} \
...
    SBINDIR=${sbindir} \
    LIBDIR=${libdir} \
    INCLUDEDIR=${includedir} \
...
    -C ${S} \
"

do_compile() {
    oe_runmake
}
```

vesajaaskelainen:
@jforissier and @etienne-lms -- shall we go forward with the proposed patch. We know that it works. Or do we have good alternatives?

jforissier:
@vesajaaskelainen I'm willing to take this as-is if it helps, because it is not so intrusive really, but I made comments that were not replied to.

etienne-lms:
Hi @jforissier, sorry not answered earlier. I'm was a bit puzzled. It looks like different yocto distribs may handle this differently.

From the OE layers i've looked at (_*.bb_ files in [Ledge RP](https://github.com/Linaro/meta-ledge/tree/hardknott/meta-ledge-sw/recipes-security/optee) and [ST's OSTL](https://github.com/STMicroelectronics/meta-st-openstlinux/tree/openstlinux-20-02-19/recipes-security/optee) oe layers), only packages building secure world SW (core, TAs, TA libs) do need to inject sysroot path info to the secure SW. They apply this `LIBGCC_LOCATE_CFLAGS` patch in optee_os's _gcc.mk_ (impacts core and TA devkit) and have in the bitbake recipe of the related packages (optee_os, optee_test, optee_examples, optee_ftpm, ...) something like: 
`EXTRA_OEMAKE += "LIBGCC_LOCATE_CFLAGS=--sysroot=${STAGING_DIR_TARGET}"`.
Looking at [zynqmp](https://github.com/OP-TEE/build/blob/3.14.0/zynqmp.mk#L49-L51) from OP-TEE release (petalinux/OE based), it seems to rely on yocto's meta-arm layer that does the same thing through an [optee.inc (tag 3.3)](https://git.yoctoproject.org/cgit/cgit.cgi/meta-arm/tree/meta-arm/recipes-security/optee/optee.inc?h=3.3&id=bcbd365cfea4ca1ac29d7755b465891da67927b0#n23) file.

Changes in the OE recipes for OP-TEE build maybe specific to those recipes and may change as OE/yocto evolves. It seems that what does not change is meant for build environments with package specific toolchain paths (as yocto is) do need package specific compilation directive, gcc using `--sysroot`. 

Maybe you're right and LIBGCC_LOCATE_CFLAGS is misnamed. Whatever the name, those OE layers will need to upgrade.

jforissier:
Could `--sysroot` be passed in `CFLAGS` (or `CFLAGS32` / `CFLAGS64`)? It seems it should work:
```
$ git grep -n CFLAGS mk/config.mk mk/gcc.mk
mk/config.mk:25:ifdef CFLAGS
mk/config.mk:26:CFLAGS32 ?= $(CFLAGS)
mk/config.mk:27:CFLAGS64 ?= $(CFLAGS)
mk/gcc.mk:16:libgcc$(sm)        := $(shell $(CC$(sm)) $(CFLAGS$(arch-bits-$(sm))) \
```

etienne-lms:
I've rather have input from an OE expert but it seems some OE recipes go your proposed way. From yocto's [meta-freescale (from currently recent master branch top)](https://git.yoctoproject.org/cgit/cgit.cgi/meta-freescale/tree/recipes-security/optee-imx/optee-os_3.10.0.imx.bb?id=f1eae1b9151ec12fd4e92edb29a04acf9f6c9ce4#n59):
> `CFLAGS += "--sysroot=${STAGING_DIR_HOST}"`
> `CXXFLAGS += "--sysroot=${STAGING_DIR_HOST}"`

@vesajaaskelainen recipes extracts [above](https://github.com/OP-TEE/optee_os/pull/4737#issuecomment-877657686) also show uses of such path, in optee_os's `do_compile` recipe:
`export CFLAGS="${CFLAGS} --sysroot=${STAGING_DIR_HOST}"`

I'll ask Linaro Ledge RP maintainers and ST's OSTL maintainers whether this path addresses their needs. I fear some corner case that makes this not fully suitable.

vesajaaskelainen:
We do have issue https://github.com/OP-TEE/optee_os/issues/4188 where this has been discussed a bit too -- no conclusions there thou.

My worry is that we start to have quite many different ways on how one builds OP-TEE in yocto.

Kinda feels that there should be one base recipe and then customizations would be in .bbappended files.

I don't think there is other common layers than "composite" `poky` and then `meta-openembedded` where such base recipe could reside. Before someone mentions about meta-security -- no I don't think it should be there. Haven't seen `meta-arm` used much. Perhaps this is something that should be discussed with Yocto community.

First step is of course getting rid of those patches ;).

vesajaaskelainen:
Btw in meta-arm there was this too: https://git.yoctoproject.org/cgit/cgit.cgi/meta-arm/tree/meta-arm/recipes-security/optee/optee-os/0007-allow-setting-sysroot-for-clang.patch

vesajaaskelainen:
One thing to remember is that in optee-tests it compiles both target binaries for OS and test TAs. They have a bit different requirements. So TA SDK need a bit different handling. Don't know if we can handle that with CFLAGS?

etienne-lms:
I fear the same about issues when CFLAGS gets --sysroot directive and is used also for the CAs compilation.

jforissier:
If `CFLAGS` doesn't work due to it also applying to the host, then you can use `CFLAGS32` and `CFLAGS64` which are OP-TEE things.

etienne-lms:
`CFLAGS32`/`CFLAGS64`, makes sense. I hope no future conflicts since the name is rather short.
I'll ask Ledge RP people view on this. Reply may be delayed by summer leaves.

etienne-lms:
I tested some configs (with a toolchain that mandates optee_os to provide sysroot path) and actually CFLAGS32 (or CFLAGS64) content is really used only when building `core` and `ta_armXX`, not when building other sub-parts (`$sm`) as ldelf and intree TAs. For example, building my 32bit platform with CFLAGS32=--sysroot=... fails when generating the ldelf application:

>   CC      out/arm-plat-stm32mp1/ldelf-lib/libutils/ext/arch/arm/auxval.o
  AR      out/arm-plat-stm32mp1/ldelf-lib/libutils/libutils.a
  CPP     out/arm-plat-stm32mp1/ldelf/ldelf.lds
  LD      out/arm-plat-stm32mp1/ldelf/ldelf.elf
arm-ostl-linux-gnueabi-ld.bfd: cannot find libgcc.a: No such file or directory
ldelf/link.mk:53: recipe for target 'out/arm-plat-stm32mp1/ldelf/ldelf.elf' failed
make: *** [out/arm-plat-stm32mp1/ldelf/ldelf.elf] Error 1

This issue is the `arch-bits-$(sm)` is set (to `32` or `64`) only for `core`, `ta_arm32` and `ta_arm64`. Should we fix that?  Maybe something like:
```patch
diff --git a/ldelf/ldelf.mk b/ldelf/ldelf.mk
index a8bf718a..784e7b8f 100644
--- a/ldelf/ldelf.mk
+++ b/ldelf/ldelf.mk
@@ -17,6 +17,7 @@ endif
 ifeq ($(CFG_ARM32_core),y)
 CFG_ARM32_$(sm) := y
 endif
+arch-bits-ldelf := $(arch-bits-core)
 
 cppflags$(sm)	+= -include $(conf-file)
 cppflags$(sm)	+= -DTRACE_LEVEL=$(CFG_TEE_CORE_LOG_LEVEL)
diff --git a/ta/mk/build-user-ta.mk b/ta/mk/build-user-ta.mk
index 9c26b6a6..9072265c 100644
--- a/ta/mk/build-user-ta.mk
+++ b/ta/mk/build-user-ta.mk
@@ -2,6 +2,13 @@ include mk/cleanvars.mk
 sm := $(lastword $(subst /, ,$(dir $(ta-mk-file))))
 sm-$(sm) := y
 
+ifeq ($(ta-target),ta_arm32)
+arch-bits-$(sm) := 32
+endif
+ifeq ($(ta-target),ta_arm64)
+arch-bits-$(sm) := 64
+endif
+
 # Select TA target (aka TA dev kit, when delivered)
 ta-target := $(strip $(if $(CFG_USER_TA_TARGET_$(sm)), \
 		$(filter $(CFG_USER_TA_TARGET_$(sm)), $(ta-targets)), \
```

jforissier:
> This issue is the `arch-bits-$(sm)` is set (to `32` or `64`) only for `core`, `ta_arm32` and `ta_arm64`. Should we fix that?

Yes, I think so, good catch! Your patch above looks good to me.

etienne-lms:
I have created https://github.com/OP-TEE/optee_os/pull/4829 for the changes.

Note i needed to patch also _ta_dev_kit.mk_ so that external packages (as optee_test, optee_client and others) can also use CFLAGS32/CFLAGS64 to pass TA specific build directives.

etienne-lms:
Addressed in #4829.

PR #4738, opened 2021-07-08
plat-stm32mp1: fix helper function and increase CFG_DTB_MAX_SIZE

etienne-lms:

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -60,6 +60,7 @@ CFG_TEE_CORE_NB_CORE ?= 2
 CFG_WITH_PAGER ?= y
 CFG_WITH_LPAE ?= y
 CFG_MMAP_REGIONS ?= 23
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -60,6 +60,7 @@ CFG_TEE_CORE_NB_CORE ?= 2
 CFG_WITH_PAGER ?= y
 CFG_WITH_LPAE ?= y
 CFG_MMAP_REGIONS ?= 23
+CFG_DTB_MAX_SIZE ?= 0x40000
```
jforissier, line 63: I personally prefer `262144` or even better `(256 * 1024)` should work -- we have this already:
```
core/arch/arm/arm.mk:CFG_RESERVED_VASPACE_SIZE ?= (1024 * 1024 * 10)
```

jforissier:
Regardless of how you prefer to write the 256 KB value...

etienne-lms:
tags applied.

PR #4741, opened 2021-07-08
Revert "libutee: TEE_MACCompareFinal(): panic if input size is too large"

jforissier:
This reverts commit dbb3274a60f0b258fe115ed1678fc569335c0c5d. It turns
out the panic reason cited in the commit ("If input data exceeds maximum
length for the algorithm") applies to the message only and has nothing
to do with macLen. The same sentence appears elsewhere in the spec where
there is no ambiguity.


jforissier:
Sorry for wasting reviewer's time on this. The original implementation was correct.

jenswi-linaro:

PR #4743, opened 2021-07-09
crypto: optimize speed of AES CBC MAC

jforissier:
The current AES CBC MAC implementation invokes the AES CBC algorithm via
crypto_cipher_update() for each 16-byte block of the input data. This
can be inefficient especially with hardware accelerated implementations
which may have a significant overhead (I am thinking of proprietary
implementations of MBed TLS for example).

This commit introduces a new config option:
CFG_CRYPTO_CBC_MAC_BUNDLE_BLOCKS (default 64) which allows to bundle
several 16-byte blocks of input data when calling the AES CBC function.
Therefore with the default value, data are processed 1 KB at a time
(assuming the caller provides enough data of course). There is a small
memory overhead (malloc) of the same size at most.

### DIFF
--- a/core/crypto/cbc-mac.c
+++ b/core/crypto/cbc-mac.c
@@ -57,6 +57,10 @@ static TEE_Result crypto_cbc_mac_init(struct crypto_mac_ctx *ctx,
 static TEE_Result crypto_cbc_mac_update(struct crypto_mac_ctx *ctx,
 					const uint8_t *data, size_t len)
 {
+	size_t nblocks = 0;
+	size_t out_len = 0;
+	uint8_t *out_tmp = NULL;
+	uint8_t *out = NULL;
 	TEE_Result res = TEE_SUCCESS;
 	struct crypto_cbc_mac_ctx *mc = to_cbc_mac_ctx(ctx);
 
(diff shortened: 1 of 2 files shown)


core/crypto/cbc-mac.c
```diff
@@ -77,14 +80,37 @@ static TEE_Result crypto_cbc_mac_update(struct crypto_mac_ctx *ctx,
...
 
 	while (len >= mc->block_len) {
+		size_t nblocks = MIN(len / mc->block_len,
+				     (size_t)CFG_CRYPTO_CBC_MAC_BUNDLE_BLOCKS);
+
+		if (nblocks > 1) {
+			out_len = nblocks * mc->block_len;
+			if (!out_tmp) {
+				out_tmp = malloc(out_len);
+				if (!out_tmp) {
```
jenswi-linaro, line 90: I'd rather fall back to process one block at a time than aborting here.

jforissier: I agree, updated.

core/crypto/cbc-mac.c
```diff
@@ -77,14 +80,36 @@ static TEE_Result crypto_cbc_mac_update(struct crypto_mac_ctx *ctx,
 	}
 
 	while (len >= mc->block_len) {
+		size_t nblocks = MIN(len / mc->block_len,
+				     (size_t)CFG_CRYPTO_CBC_MAC_BUNDLE_BLOCKS);
+
+		if (nblocks > 1) {
+			out_len = nblocks * mc->block_len;
+			if (!out_tmp) {
+				out_tmp = malloc(out_len);
```
jenswi-linaro, line 89: In case `malloc()` fails we'll try this again and again for each block.
How about moving the logic to provide the temporary buffer before this loop is entered?

jforissier: Well in theory it could eventually succeed but perhaps thee chances are slim in such a tight loop. I'll fix.

jforissier:
Update

jenswi-linaro (commented):

PR #4617, opened 2021-05-17
core_mmu: fix implicit behavior of core_mmu_add_mapping()

a-rybakov:
In core_mmu_add_mapping() requested physical address rounded up/down to granule size (0x100000), which leads
to establising of virtual mappings with overlapped physical counterparts. If two virtual mappings overlaps due to such roundings, then following phys_to_virt() can implicitly return result of virtual address from unexpected mapping. This patch fix such behavior by returning virtual address of newly established mapping.

PR #4744, opened 2021-07-10
mm: fix mobj split by adding core_mmu_find_mapping() helper

a-rybakov:
This fixes mobj splitting onto RX/RW parts. Now split can be done
incorrectly if RX and RW regions doesn`t mapped contiguosly.
Added helper core_mmu_find_mapping() allows to find mapping for
specified type and length independently of their order, so then
RX/RW regions for mobjects should be determined correctly.

Signed-off-by: Anton Rybakov <a.rybakov@omp.ru>

### DIFF
--- a/core/arch/arm/include/mm/core_mmu.h
+++ b/core/arch/arm/include/mm/core_mmu.h
@@ -631,6 +631,17 @@ void *core_mmu_add_mapping(enum teecore_memtypes type, paddr_t addr,
 			   size_t len);
 
 /*
+ * core_mmu_find_mapping_exclusive() - Find mapping of specified type and
+ *				       length. If more than one mapping of
+ *				       specified type is present, NULL will be
+ *				       returned.
+ * @type:	memory type
+ * @len:	length in bytes
+ */
+struct tee_mmap_region *
+core_mmu_find_mapping_exclusive(enum teecore_memtypes type, size_t len);
+
+/*
  * tlbi_mva_range() - Invalidate TLB for virtual address range
  * @va:		start virtual address, must be a multiple of @granule
  * @len:	length in bytes of range, must be a multiple of @granule
--- a/core/arch/arm/include/mm/core_mmu.h
+++ b/core/arch/arm/include/mm/core_mmu.h
@@ -89,6 +89,9 @@
  * MEM_AREA_TEE_RAM_RX:  core private read-only/executable memory (secure)
  * MEM_AREA_TEE_RAM_RO:  core private read-only/non-executable memory (secure)
  * MEM_AREA_TEE_RAM_RW:  core private read/write/non-executable memory (secure)
+ * MEM_AREA_INIT_RAM_RO: init private read-only/non-executable memory (secure)
+ * MEM_AREA_INIT_RAM_RX: init private read-only/executable memory (secure)
+ * MEM_AREA_NEX_RAM_RO: nexus private read-only/non-executable memory (secure)
  * MEM_AREA_NEX_RAM_RW: nexus private r/w/non-executable memory (secure)
  * MEM_AREA_TEE_COHERENT: teecore coherent RAM (secure, reserved to TEE)
  * MEM_AREA_TEE_ASAN: core address sanitizer RAM (secure, reserved to TEE)
@@ -113,6 +116,9 @@ enum teecore_memtypes {
 	MEM_AREA_TEE_RAM_RX,
 	MEM_AREA_TEE_RAM_RO,
 	MEM_AREA_TEE_RAM_RW,
+	MEM_AREA_INIT_RAM_RO,
+	MEM_AREA_INIT_RAM_RX,
+	MEM_AREA_NEX_RAM_RO,
 	MEM_AREA_NEX_RAM_RW,
 	MEM_AREA_TEE_COHERENT,
 	MEM_AREA_TEE_ASAN,
@@ -142,6 +148,9 @@ static inline const char *teecore_memtype_name(enum teecore_memtypes type)
 		[MEM_AREA_TEE_RAM_RX] = "TEE_RAM_RX",
 		[MEM_AREA_TEE_RAM_RO] = "TEE_RAM_RO",
 		[MEM_AREA_TEE_RAM_RW] = "TEE_RAM_RW",
+		[MEM_AREA_INIT_RAM_RO] = "INIT_RAM_RO",
+		[MEM_AREA_INIT_RAM_RX] = "INIT_RAM_RX",
+		[MEM_AREA_NEX_RAM_RO] = "NEX_RAM_RO",
 		[MEM_AREA_NEX_RAM_RW] = "NEX_RAM_RW",
 		[MEM_AREA_TEE_ASAN] = "TEE_ASAN",
 		[MEM_AREA_IDENTITY_MAP_RX] = "IDENTITY_MAP_RX",
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -1936,6 +1936,31 @@ TEE_Result core_mmu_remove_mapping(enum teecore_memtypes type, void *addr,
 	return TEE_SUCCESS;
 }
 
+struct tee_mmap_region *
+core_mmu_find_mapping_exclusive(enum teecore_memtypes type, size_t len)
+{
+	struct tee_mmap_region *map = NULL;
+	struct tee_mmap_region *map_found = NULL;
+
+	if (!len)
+		return NULL;
+
+	for (map = get_memory_map(); !core_mmap_is_end_of_table(map); map++) {
+		if (map->type != type)
+			continue;
+
+		if (map_found)
+			return NULL;
+
+		map_found = map;
+	}
+
+	if (!map_found || map_found->size < len)
+		return NULL;
+
+	return map_found;
+}
+
 void *core_mmu_add_mapping(enum teecore_memtypes type, paddr_t addr, size_t len)
 {
 	struct core_mmu_table_info tbl_info;
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -96,7 +96,7 @@ register_phys_mem_ul(MEM_AREA_TEE_RAM_RO, VCORE_UNPG_RO_PA,
 		     VCORE_UNPG_RO_SZ_UNSAFE);
 
 #ifdef CFG_VIRTUALIZATION
-register_phys_mem_ul(MEM_AREA_TEE_RAM_RO, VCORE_UNPG_RW_PA,
+register_phys_mem_ul(MEM_AREA_NEX_RAM_RO, VCORE_UNPG_RW_PA,
 		     VCORE_UNPG_RW_SZ_UNSAFE);
 register_phys_mem_ul(MEM_AREA_NEX_RAM_RW, VCORE_NEX_RW_PA,
 		     VCORE_NEX_RW_SZ_UNSAFE);
@@ -106,9 +106,9 @@ register_phys_mem_ul(MEM_AREA_TEE_RAM_RW, VCORE_UNPG_RW_PA,
 #endif
 
 #ifdef CFG_WITH_PAGER
-register_phys_mem_ul(MEM_AREA_TEE_RAM_RX, VCORE_INIT_RX_PA,
+register_phys_mem_ul(MEM_AREA_INIT_RAM_RX, VCORE_INIT_RX_PA,
 		     VCORE_INIT_RX_SZ_UNSAFE);
-register_phys_mem_ul(MEM_AREA_TEE_RAM_RO, VCORE_INIT_RO_PA,
+register_phys_mem_ul(MEM_AREA_INIT_RAM_RO, VCORE_INIT_RO_PA,
 		     VCORE_INIT_RO_SZ_UNSAFE);
 #endif /*CFG_WITH_PAGER*/
 #else /*!CFG_CORE_RWDATA_NOEXEC*/
@@ -633,11 +633,14 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_TEE_RAM:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PRWX | cached;
 	case MEM_AREA_TEE_RAM_RX:
+	case MEM_AREA_INIT_RAM_RX:
 	case MEM_AREA_IDENTITY_MAP_RX:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PRX | cached;
 	case MEM_AREA_TEE_RAM_RO:
+	case MEM_AREA_INIT_RAM_RO:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PR | cached;
 	case MEM_AREA_TEE_RAM_RW:
+	case MEM_AREA_NEX_RAM_RO: /* This has to be r/w during init runtime */
 	case MEM_AREA_NEX_RAM_RW:
 	case MEM_AREA_TEE_ASAN:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PRW | cached;
@@ -674,7 +677,10 @@ static bool __maybe_unused map_is_tee_ram(const struct tee_mmap_region *mm)
 	case MEM_AREA_TEE_RAM_RX:
 	case MEM_AREA_TEE_RAM_RO:
 	case MEM_AREA_TEE_RAM_RW:
+	case MEM_AREA_INIT_RAM_RX:
+	case MEM_AREA_INIT_RAM_RO:
 	case MEM_AREA_NEX_RAM_RW:
+	case MEM_AREA_NEX_RAM_RO:
 	case MEM_AREA_TEE_ASAN:
 		return true;
 	default:
@@ -1132,7 +1138,10 @@ static void check_mem_map(struct tee_mmap_region *map)
 		case MEM_AREA_TEE_RAM_RX:
 		case MEM_AREA_TEE_RAM_RO:
 		case MEM_AREA_TEE_RAM_RW:
+		case MEM_AREA_INIT_RAM_RX:
+		case MEM_AREA_INIT_RAM_RO:
 		case MEM_AREA_NEX_RAM_RW:
+		case MEM_AREA_NEX_RAM_RO:
 		case MEM_AREA_IDENTITY_MAP_RX:
 			if (!pbuf_is_inside(secure_only, m->pa, m->size))
 				panic("TEE_RAM can't fit in secure_only");
@@ -2221,12 +2230,17 @@ static void *phys_to_virt_tee_ram(paddr_t pa)
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_NEX_RAM_RW, pa);
 	if (!mmap)
+		mmap = find_map_by_type_and_pa(MEM_AREA_NEX_RAM_RO, pa);
+	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM_RW, pa);
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM_RO, pa);
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM_RX, pa);
-
+	/*
+	 * Note that MEM_AREA_INIT_RAM_RO and MEM_AREA_INIT_RAM_RX are only
+	 * used with pager and not needed here.
+	 */
 	return map_pa2va(mmap, pa);
 }
 #endif
@@ -2243,6 +2257,7 @@ void *phys_to_virt(paddr_t pa, enum teecore_memtypes m)
 	case MEM_AREA_TEE_RAM_RX:
 	case MEM_AREA_TEE_RAM_RO:
 	case MEM_AREA_TEE_RAM_RW:
+	case MEM_AREA_NEX_RAM_RO:
 	case MEM_AREA_NEX_RAM_RW:
 		va = phys_to_virt_tee_ram(pa);
 		break;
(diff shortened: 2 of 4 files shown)


jforissier:
Once this is reviewed, could you please also add `Fixes: ff01e2452169 ("mm: split mobj_tee_ram onto rw/rx parts")`? Although we don't use such a tag like in the linux kernel to backport to stable branches, it can still be helpful.

a-rybakov:
> Once this is reviewed, could you please also add `Fixes: ff01e2452169 ("mm: split mobj_tee_ram onto rw/rx parts")`? Although we don't use such a tag like in the linux kernel to backport to stable branches, it can still be helpful.

Of course, I will add this. Just one remark - previous patch by itself works well, but this is not enough because it still inherits previous imprecision. 

P.S. New phys_to_virt() with size parameter from PR #4617 detects this, so it seems useful.

jforissier:
> > Once this is reviewed, could you please also add `Fixes: ff01e2452169 ("mm: split mobj_tee_ram onto rw/rx parts")`? Although we don't use such a tag like in the linux kernel to backport to stable branches, it can still be helpful.
> 
> Of course, I will add this. Just one remark - previous patch by itself works well, but this is not enough because it still inherits previous imprecision.

OK, perhaps the `Fixes:` tag is not a good idea then (it really makes sense only if a bug is introduced or triggered by the cited patch). I let you judge.

core/mm/mobj.c
```diff
@@ -141,15 +141,26 @@ static struct mobj *mobj_phys_init(paddr_t pa, size_t size, uint32_t cattr,
...
 	if ((pa & CORE_MMU_USER_PARAM_MASK) ||
 	    (size & CORE_MMU_USER_PARAM_MASK)) {
 		DMSG("Expect %#x alignment", CORE_MMU_USER_PARAM_SIZE);
 		return NULL;
 	}
 
+	if (pa) {
+		va = phys_to_virt(pa, area_type);
+	} else {
+		map = core_mmu_find_mapping(area_type, size);
```
jenswi-linaro, line 155: What we need here is a function that can guarantee that we get the correct map. One way of doing that might be to require that there is only one `MEM_AREA_TEE_RAM_RW_DATA`.

core/mm/mobj.c
```diff
@@ -685,16 +696,15 @@ static TEE_Result mobj_init(void)
 		panic("Failed to register secure ta ram");
 
 	if (IS_ENABLED(CFG_CORE_RWDATA_NOEXEC)) {
-		mobj_tee_ram_rx = mobj_phys_init(TEE_RAM_START,
+		mobj_tee_ram_rx = mobj_phys_init(0,
```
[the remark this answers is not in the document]
a-rybakov: Setting pa to 0 tells to mobj_phys_init() that pa should be searched with core_mmu_find_mapping_exclusive() instead of phys_to_virt()

core/mm/mobj.c
a-rybakov: Sounds reasonable. I transformed core_mmu_find_mapping() to core_mmu_find_mapping_exclusive() to guarantee that found mapping is unique of that type.

core/arch/arm/mm/core_mmu.c
```diff
@@ -1927,6 +1927,27 @@ TEE_Result core_mmu_remove_mapping(enum teecore_memtypes type, void *addr,
 	return TEE_SUCCESS;
 }
 
+struct tee_mmap_region *core_mmu_find_mapping_exclusive(enum
+							teecore_memtypes type,
+							size_t len)
+{
+	unsigned int map_cnt = 0;
```
jenswi-linaro, line 1934: This function could be simplified as (skipping `map_cnt`, exit early and avoiding the ternary operator at the end):
``` C
        struct tee_mmap_region *map = NULL; 
        struct tee_mmap_region *map_found = NULL;
        
        if (!len)
                return NULL;
        
        for (map = get_memory_map(); !core_mmap_is_end_of_table(map); map++) {
                if (map->type != type)
                        continue;
                
                if (map_found) 
                        return NULL;
                map_found = map;
        }

        if (!map_found || map_found->size < len)
                return NULL;

        return map_found;
```

core/arch/arm/include/mm/core_mmu.h
```diff
@@ -621,6 +621,18 @@ TEE_Result core_mmu_remove_mapping(enum teecore_memtypes type, void *addr,
...
 
+/*
+ * core_mmu_find_mapping_exclusive() - Find mapping of specified type and
+ *				       length. If more than one mapping of
+ *				       specified type is present, NULL will be
+ *				       returned.
+ * @type:	memory type
+ * @len:	length in bytes
+ */
+struct tee_mmap_region *core_mmu_find_mapping_exclusive(enum
```
jenswi-linaro, line 632: Please break the line as:
``` C
struct tee_mmap_region *
core_mmu_find_mapping_exclusive(enum teecore_memtypes type, size_t len);
```
Please fix at the implementation too.

jenswi-linaro (commented):
With my final comments addressed:

a-rybakov:
@jforissier - could you please provide logs for running this patch on HiKey with TRACE_DEBUG level? Need to see memory map to find what is going wrong.

jforissier:
@a-rybakov I can probably try it on Monday. I will let you know.

a-rybakov:
> @a-rybakov I can probably try it on Monday. I will let you know.

Thank you! But I`ve added temporarily some debug into patch, so I think it isn`t necessary for now.

a-rybakov:
33:  NOTICE:  BL31: v2.5(release):v2.5
    34:  NOTICE:  BL31: Built : 17:30:38, JulE/TC:0   dump_mmap_table:711 type IO_NSEC      va 0x3ac00000..0x3adfffff pa 0xf8000000..0xf81fffff size 0x00200000 (pgdir)
    35:  E/TC:0   dump_mmap_table:711 type IO_NSEC      va 0x3ae00000..0x3affffff pa 0xf7000000..0xf71fffff size 0x00200000 (pgdir)
    36:  E/TC:0   dump_mmap_table:711 type TA_RAM       va 0x3b000000..0x3bdfffff pa 0x3f200000..0x3fffffff size 0x00e00000 (pgdir)
    37:  E/TC:0   dump_mmap_table:711 type NSEC_SHM     va 0x3c000000..0x3c1fffff pa 0x3ee00000..0x3effffff size 0x00200000 (pgdir)
    38:  E/TC:0   dump_mmap_table:711 type RES_VASPACE  va 0x3c200000..0x3cbfffff pa 0x00000000..0x009fffff size 0x00a00000 (pgdir)
    39:  E/TC:0   dump_mmap_table:711 type SHM_VASPACE  va 0x3ce00000..0x3edfffff pa 0x00000000..0x01ffffff size 0x02000000 (pgdir)
    40:  E/TC:0   dump_mmap_table:711 type TEE_RAM_RX   va 0x3f000000..0x3f010fff pa 0x3f000000..0x3f010fff size 0x00011000 (smallpg)
    41:  E/TC:0   dump_mmap_table:711 type TEE_RAM_RW   va 0x3f011000..0x3f045fff pa 0x3f011000..0x3f045fff size 0x00035000 (smallpg)
    42:  E/TC:0   dump_mmap_table:711 type TEE_RAM_RX   va 0x3f046000..0x3f06ffff pa 0x3f046000..0x3f06ffff size 0x0002a000 (smallpg)
    43:  E/TC:0   dump_mmap_table:711 type PAGER_VASPACE va 0x3f070000..0x3f1fffff pa 0x00000000..0x0018ffff size 0x00190000 (smallpg)
    
    Seems that patch works well - as for this platform defined two TEE_RAM_RX regions, it fails because of "exclusive" requirements. @jenswi-linaro @jforissier is it necessary to have two TEE_RAM_TX for this platform, what do you think?

jforissier:
> Seems that patch works well - as for this platform defined two TEE_RAM_RX regions, it fails because of "exclusive" requirements. @jenswi-linaro @jforissier is it necessary to have two TEE_RAM_TX for this platform, what do you think?

The two regions are because of `CFG_WITH_PAGER=y`, this is not specific to HiKey (just that CI happens to test with pager enabled on HiKey). You can reproduce the same issue on QEMUv8.

See https://github.com/OP-TEE/optee_os/blob/3.14.0/core/arch/arm/mm/core_mmu.c#L93-L94 and https://github.com/OP-TEE/optee_os/blob/3.14.0/core/arch/arm/mm/core_mmu.c#L109-L110.

I don't think it is realistic to require a single region for `VCORE_UNPG_RX_PA` and `VCORE_INIT_RX_PA`, but I will let @jenswi-linaro comment on this because it is definitely not my area of expertise.

jenswi-linaro:
> I don't think it is realistic to require a single region for `VCORE_UNPG_RX_PA` and `VCORE_INIT_RX_PA`, but I will let @jenswi-linaro comment on this because it is definitely not my area of expertise.

No, that's not possible. They must be separated or you'd cripple the pager.

a-rybakov:
> > I don't think it is realistic to require a single region for `VCORE_UNPG_RX_PA` and `VCORE_INIT_RX_PA`, but I will let @jenswi-linaro comment on this because it is definitely not my area of expertise.
> 
> No, that's not possible. They must be separated or you'd cripple the pager.

@jenswi-linaro So I suppose there is no possiblity to require for TEE_RAM_RX mapping to be exclusive. For this patch it seems enough to check TEE_RAM_RX mapping to be match with required size. Do you agree?

jenswi-linaro:
> > > I don't think it is realistic to require a single region for `VCORE_UNPG_RX_PA` and `VCORE_INIT_RX_PA`, but I will let @jenswi-linaro comment on this because it is definitely not my area of expertise.
> > 
> > 
> > No, that's not possible. They must be separated or you'd cripple the pager.
> 
> @jenswi-linaro So I suppose there is no possibility to require for TEE_RAM_RX mapping to be exclusive. For this patch it seems enough to check TEE_RAM_RX mapping to be match with required size. Do you agree?

No, that's not good enough. I believe you need something like this:
``` patch
From 659c873e5e273017a2be8bbfc03ec4ec995aedf5 Mon Sep 17 00:00:00 2001
From: Jens Wiklander <jens.wiklander@linaro.org>
Date: Mon, 26 Jul 2021 12:21:06 +0200
Subject: [PATCH] core: add new init and nexus memory types

Adds the new memory types MEM_AREA_INIT_RAM_RO, MEM_AREA_INIT_RAM_RX and
MEM_AREA_NEX_RAM_RO to make sure that the memory types
MEM_AREA_TEE_RAM_RX, MEM_AREA_TEE_RAM_RO and MEM_AREA_TEE_RAM_RW are
used only once. This is needed when to uniquely identify those memory
areas in mobj_init() and mobj_phys_init().

---
 core/arch/arm/include/mm/core_mmu.h   |  9 +++++++++
 core/arch/arm/kernel/virtualization.c |  1 +
 core/arch/arm/mm/core_mmu.c           | 23 +++++++++++++++++++----
 3 files changed, 29 insertions(+), 4 deletions(-)

diff --git a/core/arch/arm/include/mm/core_mmu.h b/core/arch/arm/include/mm/core_mmu.h
index 4005dda72c36..09a457a6fb08 100644
--- a/core/arch/arm/include/mm/core_mmu.h
+++ b/core/arch/arm/include/mm/core_mmu.h
@@ -89,6 +89,9 @@
  * MEM_AREA_TEE_RAM_RX:  core private read-only/executable memory (secure)
  * MEM_AREA_TEE_RAM_RO:  core private read-only/non-executable memory (secure)
  * MEM_AREA_TEE_RAM_RW:  core private read/write/non-executable memory (secure)
+ * MEM_AREA_INIT_RAM_RO: init private read-only/non-executable memory (secure)
+ * MEM_AREA_INIT_RAM_RX: init private read-only/executable memory (secure)
+ * MEM_AREA_NEX_RAM_RO: nexus private read-only/non-executable memory (secure)
  * MEM_AREA_NEX_RAM_RW: nexus private r/w/non-executable memory (secure)
  * MEM_AREA_TEE_COHERENT: teecore coherent RAM (secure, reserved to TEE)
  * MEM_AREA_TEE_ASAN: core address sanitizer RAM (secure, reserved to TEE)
@@ -113,6 +116,9 @@ enum teecore_memtypes {
 	MEM_AREA_TEE_RAM_RX,
 	MEM_AREA_TEE_RAM_RO,
 	MEM_AREA_TEE_RAM_RW,
+	MEM_AREA_INIT_RAM_RO,
+	MEM_AREA_INIT_RAM_RX,
+	MEM_AREA_NEX_RAM_RO,
 	MEM_AREA_NEX_RAM_RW,
 	MEM_AREA_TEE_COHERENT,
 	MEM_AREA_TEE_ASAN,
@@ -142,6 +148,9 @@ static inline const char *teecore_memtype_name(enum teecore_memtypes type)
 		[MEM_AREA_TEE_RAM_RX] = "TEE_RAM_RX",
 		[MEM_AREA_TEE_RAM_RO] = "TEE_RAM_RO",
 		[MEM_AREA_TEE_RAM_RW] = "TEE_RAM_RW",
+		[MEM_AREA_INIT_RAM_RO] = "INIT_RAM_RO",
+		[MEM_AREA_INIT_RAM_RX] = "INIT_RAM_RX",
+		[MEM_AREA_NEX_RAM_RO] = "NEX_RAM_RO",
 		[MEM_AREA_NEX_RAM_RW] = "NEX_RAM_RW",
 		[MEM_AREA_TEE_ASAN] = "TEE_ASAN",
 		[MEM_AREA_IDENTITY_MAP_RX] = "IDENTITY_MAP_RX",
diff --git a/core/arch/arm/kernel/virtualization.c b/core/arch/arm/kernel/virtualization.c
index fbf80d850f5e..c595832b31e9 100644
--- a/core/arch/arm/kernel/virtualization.c
+++ b/core/arch/arm/kernel/virtualization.c
@@ -152,6 +152,7 @@ void virt_init_memory(struct tee_mmap_region *memory_map)
 		switch (map->type) {
 		case MEM_AREA_TEE_RAM_RX:
 		case MEM_AREA_TEE_RAM_RO:
+		case MEM_AREA_NEX_RAM_RO:
 		case MEM_AREA_NEX_RAM_RW:
 			DMSG("Carving out area of type %d (0x%08lx-0x%08lx)",
 			     map->type, map->pa, map->pa + map->size);
diff --git a/core/arch/arm/mm/core_mmu.c b/core/arch/arm/mm/core_mmu.c
index cb972ce02e65..bfcd13760dc0 100644
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -96,7 +96,7 @@ register_phys_mem_ul(MEM_AREA_TEE_RAM_RO, VCORE_UNPG_RO_PA,
 		     VCORE_UNPG_RO_SZ_UNSAFE);
 
 #ifdef CFG_VIRTUALIZATION
-register_phys_mem_ul(MEM_AREA_TEE_RAM_RO, VCORE_UNPG_RW_PA,
+register_phys_mem_ul(MEM_AREA_NEX_RAM_RO, VCORE_UNPG_RW_PA,
 		     VCORE_UNPG_RW_SZ_UNSAFE);
 register_phys_mem_ul(MEM_AREA_NEX_RAM_RW, VCORE_NEX_RW_PA,
 		     VCORE_NEX_RW_SZ_UNSAFE);
@@ -106,9 +106,9 @@ register_phys_mem_ul(MEM_AREA_TEE_RAM_RW, VCORE_UNPG_RW_PA,
 #endif
 
 #ifdef CFG_WITH_PAGER
-register_phys_mem_ul(MEM_AREA_TEE_RAM_RX, VCORE_INIT_RX_PA,
+register_phys_mem_ul(MEM_AREA_INIT_RAM_RX, VCORE_INIT_RX_PA,
 		     VCORE_INIT_RX_SZ_UNSAFE);
-register_phys_mem_ul(MEM_AREA_TEE_RAM_RO, VCORE_INIT_RO_PA,
+register_phys_mem_ul(MEM_AREA_INIT_RAM_RO, VCORE_INIT_RO_PA,
 		     VCORE_INIT_RO_SZ_UNSAFE);
 #endif /*CFG_WITH_PAGER*/
 #else /*!CFG_CORE_RWDATA_NOEXEC*/
@@ -633,11 +633,14 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_TEE_RAM:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PRWX | cached;
 	case MEM_AREA_TEE_RAM_RX:
+	case MEM_AREA_INIT_RAM_RX:
 	case MEM_AREA_IDENTITY_MAP_RX:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PRX | cached;
 	case MEM_AREA_TEE_RAM_RO:
+	case MEM_AREA_INIT_RAM_RO:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PR | cached;
 	case MEM_AREA_TEE_RAM_RW:
+	case MEM_AREA_NEX_RAM_RO: /* This has to be r/w during init runtime */
 	case MEM_AREA_NEX_RAM_RW:
 	case MEM_AREA_TEE_ASAN:
 		return attr | TEE_MATTR_SECURE | TEE_MATTR_PRW | cached;
@@ -674,7 +677,10 @@ static bool __maybe_unused map_is_tee_ram(const struct tee_mmap_region *mm)
 	case MEM_AREA_TEE_RAM_RX:
 	case MEM_AREA_TEE_RAM_RO:
 	case MEM_AREA_TEE_RAM_RW:
+	case MEM_AREA_INIT_RAM_RX:
+	case MEM_AREA_INIT_RAM_RO:
 	case MEM_AREA_NEX_RAM_RW:
+	case MEM_AREA_NEX_RAM_RO:
 	case MEM_AREA_TEE_ASAN:
 		return true;
 	default:
@@ -1132,7 +1138,10 @@ static void check_mem_map(struct tee_mmap_region *map)
 		case MEM_AREA_TEE_RAM_RX:
 		case MEM_AREA_TEE_RAM_RO:
 		case MEM_AREA_TEE_RAM_RW:
+		case MEM_AREA_INIT_RAM_RX:
+		case MEM_AREA_INIT_RAM_RO:
 		case MEM_AREA_NEX_RAM_RW:
+		case MEM_AREA_NEX_RAM_RO:
 		case MEM_AREA_IDENTITY_MAP_RX:
 			if (!pbuf_is_inside(secure_only, m->pa, m->size))
 				panic("TEE_RAM can't fit in secure_only");
@@ -2245,13 +2254,18 @@ static void *phys_to_virt_tee_ram(paddr_t pa)
 	mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM, pa);
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_NEX_RAM_RW, pa);
+	if (!mmap)
+		mmap = find_map_by_type_and_pa(MEM_AREA_NEX_RAM_RO, pa);
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM_RW, pa);
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM_RO, pa);
 	if (!mmap)
 		mmap = find_map_by_type_and_pa(MEM_AREA_TEE_RAM_RX, pa);
-
+	/*
+	 * Note that MEM_AREA_INIT_RAM_RO and MEM_AREA_INIT_RAM_RX are only
+	 * used with pager and not needed here.
+	 */
 	return map_pa2va(mmap, pa);
 }
 #endif
@@ -2268,6 +2282,7 @@ void *phys_to_virt(paddr_t pa, enum teecore_memtypes m)
 	case MEM_AREA_TEE_RAM_RX:
 	case MEM_AREA_TEE_RAM_RO:
 	case MEM_AREA_TEE_RAM_RW:
+	case MEM_AREA_NEX_RAM_RO:
 	case MEM_AREA_NEX_RAM_RW:
 		va = phys_to_virt_tee_ram(pa);
 		break;
-- 
2.31.1
```

a-rybakov:
> > 
> > @jenswi-linaro So I suppose there is no possibility to require for TEE_RAM_RX mapping to be exclusive. For this patch it seems enough to check TEE_RAM_RX mapping to be match with required size. Do you agree?
> 
> No, that's not good enough. I believe you need something like this:

Yes, this looks better. So I suppose my patch have to wait until yours will be committed.

jenswi-linaro:
> > > @jenswi-linaro So I suppose there is no possibility to require for TEE_RAM_RX mapping to be exclusive. For this patch it seems enough to check TEE_RAM_RX mapping to be match with required size. Do you agree?
> > 
> > 
> > No, that's not good enough. I believe you need something like this:
> 
> Yes, this looks better. So I suppose my patch have to wait until yours will be committed.

Feel free to add it first in this pull request. I can do a separate PR if you'd prefer that, it up to you.

a-rybakov:
> > > 
> > > No, that's not good enough. I believe you need something like this:
> > 
> > 
> > Yes, this looks better. So I suppose my patch have to wait until yours will be committed.
> 
> Feel free to add it first in this pull request. I can do a separate PR if you'd prefer that, it up to you.

@jenswi-linaro - with your patch applied all seems working as it should. Thank you!

jenswi-linaro:
> @jenswi-linaro - with your patch applied all seems working as it should. Thank you!

That's great. Let's hope this was the last obstacle to overcome.

jforissier:
> > > > No, that's not good enough. I believe you need something like this:
> > > 
> > > 
> > > Yes, this looks better. So I suppose my patch have to wait until yours will be committed.
> > 
> > 
> > Feel free to add it first in this pull request. I can do a separate PR if you'd prefer that, it up to you.
> 
> @jenswi-linaro - with your patch applied all seems working as it should. Thank you!

Jens' patch needs to come first to preserve bisectability.

a-rybakov:
> Jens' patch needs to come first to preserve bisectability.

@jforissier Done.

jforissier:
This looks mostly ready for merging but I'd like a Reviewed-by: or Acked-by: on the first commit, @a-rybakov from you for instance. Thanks.

a-rybakov:
> This looks mostly ready for merging but I'd like a Reviewed-by: or Acked-by: on the first commit, @a-rybakov from you for instance. Thanks.

Ok, done.

PR #4750, opened 2021-07-14
core: mm: remove duplicate mobj_put() in vm_map_pad()

ZheTingLiu:
Whatever the return value of vm_map_pad, the mobj_put() is always called.
Therefore mobj_put() should not be called when mobj_get_cattr(mobj, &cattr) fails.

Signed-off-by: ZheTingLiu \<Gavin.Liu@mediatek.com>

### DIFF
--- a/core/mm/vm.c
+++ b/core/mm/vm.c
@@ -293,7 +293,7 @@ TEE_Result vm_map_pad(struct user_mode_ctx *uctx, vaddr_t *va, size_t len,
 
 	res = umap_add_region(&uctx->vm_info, reg, pad_begin, pad_end, align);
 	if (res)
-		goto err_free_reg;
+		goto err_put_mobj;
 
 	res = alloc_pgt(uctx);
 	if (res)
@@ -326,8 +326,9 @@ TEE_Result vm_map_pad(struct user_mode_ctx *uctx, vaddr_t *va, size_t len,
 
 err_rem_reg:
 	TAILQ_REMOVE(&uctx->vm_info.regions, reg, link);
-err_free_reg:
+err_put_mobj:
 	mobj_put(reg->mobj);
+err_free_reg:
 	free(reg);
 	return res;
 }


core/mm/vm.c
```diff
@@ -327,7 +327,6 @@ TEE_Result vm_map_pad(struct user_mode_ctx *uctx, vaddr_t *va, size_t len,
 err_rem_reg:
 	TAILQ_REMOVE(&uctx->vm_info.regions, reg, link);
 err_free_reg:
-	mobj_put(reg->mobj);
```
etienne-lms, line 330: sorry, we misunderstood ourselves. This `mobj_put()` is needed to balance `mobj_get()` (line 287) in case the sequecen fails (lines 296, 300, ...).

Want is needed her IMO is:
```patch
 	(...)
 	res = umap_add_region(&uctx->vm_info, reg, pad_begin, pad_end, align);
 	if (res)
-		goto err_free_reg;
+ 		goto err_put_mobj;
	res = alloc_pgt(uctx);
	if (res)
			goto err_free_reg;
 
	(...)
 
 	return TEE_SUCCESS;
 
 err_rem_reg:
 	TAILQ_REMOVE(&uctx->vm_info.regions, reg, link);
+err_put_mobj:
+	mobj_put(reg->mobj);
 err_free_reg:
-	mobj_put(reg->mobj);
 	free(reg);
 	return res;

**Review by etienne-lms — COMMENTED** [review]
Comment on commit log.
`Signed-off-by:` shall mention `Real Name <some-address@xxx.yyy>`. Please fix.
See https://optee.readthedocs.io/en/latest/general/contribute.html.

File: core/mm/vm.c (hunk 1 above)
Comment (ZheTingLiu) [reply core/mm/vm.c:330]: After vm_map_pad done, it will do `mobj_put()`. Does `mobj_put()`(line 157)  balance `mobj_get()` (line 287)?
https://github.com/OP-TEE/optee_os/blob/203ee23d005b2cec2f21b5de334c5a246be32599/core/pta/system.c#L151-L161

Comment (etienne-lms) [reply core/mm/vm.c:330]: The picture is:
* `system_map_zi()` gets a refcount on the mobj due to `mobj_with_fobj_alloc()`.
* `system_map_zi()` calls `vm_map_pad()`.
  * `vm_map_pad()` gets a refcount on mobj and maps the object.
  On failure to map, it must release its refcount before returning with an error status.
* `system_map_zi()` releases its refcount on mobj.
   If mobj was successfully mapped, the mapping holds a refcount.
   If mobj failed to be mapped, its recount now reaches 0 and it is destroyed/released.

When the mobj reference will be unmapped, there is a call to `mobj_put(reg->mobj);` to release to refcount held from `vm_map_pad()` and trigger mobj release:
https://github.com/OP-TEE/optee_os/blob/d5ad7ccf818dc8516f2abe9e3262955c91a03627/core/mm/vm.c#L750-L755

Comment (ZheTingLiu) [reply core/mm/vm.c:330]: Ok, I understand the flow. I think I miss some points, thanks

**Comment by etienne-lms** [comment]
Implementation looks good to me. Please  fix the commit log (see below)
and add my `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` tag in the commit log next to your-S-o-b tag.

Header line should use prefix `core: mm: remove ...`.
The description is not the exact one. This fixes the case where `mobj_get_cattr()` fails. Before your fix there was a unexpected  refcount decrement on the mobj.
Also don't refer to source files line numbers in the commit log. One applying your commit on an older optee_os git version would get inaccurate information from the commit log.
Last, fix your S-o-b tag using your real name as full name. Suggestion:
```
core: mm: remove duplicate mobj_put() in vm_map_pad()

When mobj_get_cattr() fails vm_map_pad() doesn't need to
call mobj_put() which is expected to balance mobj_get() called
only after mobj_get_cattr() succeeds. The issue was introduced
in release 3.8.0 with struct mobj reference counting.

Signed-off-by: Gavin Lui <Gavin.Liu@mediatek.com>

ZheTingLiu:
I just modified the commit message.
Thank you for proofreading.

jenswi-linaro:
If I'm understanding this correctly, this is not fixing problem in the code, it's rather a style issue or optimization. `reg->mobj` is always `NULL` until `mobj_get()` has been called so `mobj_put()` will do nothing before that.

However I agree that that call isn't needed until `mobj_get()` has been called. So we should perhaps change the title to
```
core: mm: remove redundant mobj_put() in vm_map_pad()
```
to make it more clear that it's not a bugfix that may need to be cherry-picked for older releases.

ZheTingLiu:
> If I'm understanding this correctly, this is not fixing problem in the code, it's rather a style issue or optimization. `reg->mobj` is always `NULL` until `mobj_get()` has been called so `mobj_put()` will do nothing before that.
> 
> However I agree that that call isn't needed until `mobj_get()` has been called. So we should perhaps change the title to
> 
> ```
> core: mm: remove redundant mobj_put() in vm_map_pad()
> ```
> 
> to make it more clear that it's not a bugfix that may need to be cherry-picked for older releases.

Ok, I just changed the commit message.

jenswi-linaro:

ZheTingLiu:

Thanks. Applied tag.

PR #4751, opened 2021-07-14
vexpress-qemu_v8a: Enable trusted_keys as early TA

b49020:
Demonstrate trusted keys usage on qemu_v8a via bundling trusted keys TA
as an early TA being part of OP-TEE OS binary.

Signed-off-by: Sumit Garg <sumit.garg@linaro.org>


jenswi-linaro:
I know that we have a similar thing for AVB on Hikey, but perhaps this is something that should rather belong to the build git since it's only needed when building for this demo.

b49020:
Apart from demo, my other intention was to include trusted keys testing using keyctl (expect it to be a simple shell script) in the CI loop. In another PR to build git repo, @jforissier suggested to use `make check` for this feature, is that part of CI loop? So would it be acceptable to have all corresponding changes to the build git?

jenswi-linaro:
Sure, no problem with making all the changes in the build git.

jenswi-linaro:
> Apart from demo, my other intention was to include trusted keys testing using keyctl (expect it to be a simple shell script) in the CI loop. In another PR to build git repo, @jforissier suggested to use `make check` for this feature, is that part of CI loop?

I missed the other question, yes that's part of the CI loop.

b49020:
Thanks @jenswi-linaro. Let me close this PR.

PR #4758, opened 2021-07-19
ci: azure: run 'git pull' to update repo tool before 'repo sync'

jforissier:
Update the repo tool to the latest version before running 'repo sync'.
Fixes the following error [1]:
```
 sudo -E bash -c "cd /root/optee_repo_qemu_v8 && repo sync -j 10"
 project .repo/manifests/
 Updating 7b69f4f..30fb9a8
 Fast-forward

 info: A new version of repo is available
 warning: repo is not tracking a remote branch, so it will not receive updates
 repo: Updating release signing keys to keyset ver 2.3
 repo reset: error: Entry 'docs/manifest-format.md' not uptodate. Cannot merge.
 fatal: Could not reset index file to revision 'v2.16.2^0'.
```
Link: [1] https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=855&view=logs&j=705748f3-7146-5e86-79af-1c0266d20a8c&t=d3458a7d-1a81-5639-8e63-8927d807e53a&l=23

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -189,6 +189,7 @@ jobs:
           set -e -v
           export LC_ALL=C
           WD=$(pwd)
+          sudo -E bash -c "cd /root/optee_repo_qemu_v8/.repo/repo && git pull"
           sudo -E bash -c "cd /root/optee_repo_qemu_v8 && repo sync -j 10"
           sudo mv /root/optee_repo_qemu_v8/optee_os /root/optee_repo_qemu_v8/optee_os_old
           sudo ln -s ${WD} /root/optee_repo_qemu_v8/optee_os


vesajaaskelainen:
As the people are on holidays and pipeline is stuck I can give my acks for this.

I assume you have tested it out that it actually works.

The change seems sane.

Just a note that you can think if you want to change now or in future.

Change itself does not tell what git repo branch (or git repository) is being tracked here. There is `stable` branch which I suppose should be used.

But even in `stable` branch the tip may not be release tag (as of now it is).

Wondering if it would be better to fetch latest released version as per instructions in:
https://source.android.com/setup/develop#installing-repo

`Acked-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

I suppose alternative for this change would be to update the docker image to have newer repo tool so that this CI pipeline would not require fetching of repo on every test run.

jenswi-linaro:

jforissier:
> As the people are on holidays and pipeline is stuck I can give my acks for this.

Thanks, appreciated (in addition to the huge work you are doing on PKCS#11).

> I assume you have tested it out that it actually works.

It does, see the CI status of this PR. I also tested manually in the Docker image (`docker run --rm -it jforissier/optee_os_ci:qemuv8_check`).

> The change seems sane.
> 
> Just a note that you can think if you want to change now or in future.
> 
> Change itself does not tell what git repo branch (or git repository) is being tracked here. There is `stable` branch which I suppose should be used.

Yes, the `default` local branch tracks the `stable` remote branch. I will add that to the commit description.

> But even in `stable` branch the tip may not be release tag (as of now it is).
> 
> Wondering if it would be better to fetch latest released version as per instructions in:
> https://source.android.com/setup/develop#installing-repo

I don't think it maters much since the `stable` branch is supposed to be... stable :)

> `Acked-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`
> 
> I suppose alternative for this change would be to update the docker image to have newer repo tool so that this CI pipeline would not require fetching of repo on every test run.

I thought about that too, and I will definitely update the image from time to time, but relying only on the image being up-to-date would be too fragile.

Thanks for reviewing!

PR #4759, opened 2021-07-20
plat-marvell: Add HUK support for OcteonTX2 Platforms

anilreddy3:
Added support for fetching Hardware Unique Key(HUK)
from the OcteonTX2 platform.

Signed-off-by: Anil Kumar Reddy <areddy3@marvell.com>
Reviewed-by: Sunil Kovvuri Goutham <sgoutham@marvell.com>
Reviewed-by: Bharat Bhushan <bbhushan2@marvell.com>


core/arch/arm/plat-marvell/main.c
```diff
@@ -108,3 +114,36 @@ void console_init(void)
 #endif
 	register_serial_console(&console_data.chip);
 }
+
+#ifdef CFG_HW_UNQ_KEY_SUPPORT
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	uint8_t hw_unq_key[sizeof(hwkey->data)] __aligned(64);
+	uint64_t huk1 = 0, huk2 = 0;
+	vaddr_t fuse_huk1 = 0, fuse_huk2 = 0;
```
jforissier, line 123: - One variable per line please.
- How about making `fuse_huk1` and `fuse_huk2` `void *`? It seems you could get rid of the casts.

core/arch/arm/plat-marvell/main.c
```diff
@@ -108,3 +114,36 @@ void console_init(void)
 #endif
 	register_serial_console(&console_data.chip);
 }
+
+#ifdef CFG_HW_UNQ_KEY_SUPPORT
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	uint8_t hw_unq_key[sizeof(hwkey->data)] __aligned(64);
```
jforissier, line 121: Why `__aligned(64)`?

core/arch/arm/plat-marvell/main.c
anilreddy3: > * One variable per line please.
Okay agree. Will make the change.
> * How about making `fuse_huk1` and `fuse_huk2` `void *`? It seems you could get rid of the casts.
correct I thought about it, but just for the consistency of type vaddr_t as per most of the places of optee os I forced to maintain the same.

core/arch/arm/plat-marvell/main.c
anilreddy3: This will insure the compiler to allocate the variable on a 64-byte boundary, this can make copy operations more efficient, so that compiler can use whatever instructions copy the biggest chunks of memory when performing copies to or from the variable.

core/arch/arm/plat-marvell/main.c
anilreddy3: > * One variable per line please.
> * How about making `fuse_huk1` and `fuse_huk2` `void *`? It seems you could get rid of the casts.

- Okay agree. Will make the change.
- Correct I thought about it, but just for the consistency of type vaddr_t as per most of the places of optee os I forced to maintain the same.

core/arch/arm/plat-marvell/main.c
anilreddy3: > Why `__aligned(64)`?

This will insure the compiler to allocate the variable on a 64-byte boundary, this can make copy operations more efficient, so that compiler can use whatever instructions copy the biggest chunks of memory when performing copies to or from the variable.

jenswi-linaro: I don't think we need this kind of micro optimization. By the way, which instruction needs a 64-byte aligned address?

anilreddy3: My understanding is, memcpy performance depends on many factors and various methods to optimize it. Address alignment is the one the base factor. For example, load and store addresses should be aligned to cache lines to allow more efficient memory accesses.
Reference links: 
https://developer.arm.com/documentation/102159/0400/Load-and-store---leftovers
https://www.nxp.com/docs/en/application-note/AN12628.pdf   (2.2.1 Address align)

core/arch/arm/plat-marvell/main.c
jforissier: >     * Correct I thought about it, but just for the consistency of type vaddr_t as per most of the places of optee os I forced to maintain the same.

Because very often `phys_to_virt()` is used to obtain a base address, which is later used with byte offsets (base + x). `void *` would not work in this case and `vaddr_t` is appropriate. Here it is not the case and `void *` is definitely the right choice.

core/arch/arm/plat-marvell/main.c
jforissier: This is undoubtedly true, but why optimize for speed here? And even so, would `__aligned(64)` make a measurable difference? Have you run benchmarks? :wink:
We are certainly not going to add alignment directives to all variables passed to `memcpy()`.

core/arch/arm/plat-marvell/main.c
jforissier: >     * Correct I thought about it, but just for the consistency of type vaddr_t as per most of the places of optee os I forced to maintain the same.

Because very often `phys_to_virt()` is used to obtain a base address, which is later used with byte offsets (base + x). `void *` would not work in this case and `vaddr_t` is appropriate. Here it is not the case and `void *` is definitely the right choice.

anilreddy3: > > ```
> > * Correct I thought about it, but just for the consistency of type vaddr_t as per most of the places of optee os I forced to maintain the same.
> > ```
> 
> Because very often `phys_to_virt()` is used to obtain a base address, which is later used with byte offsets (base + x). `void *` would not work in this case and `vaddr_t` is appropriate. Here it is not the case and `void *` is definitely the right choice.

Agree, will make the change.

core/arch/arm/plat-marvell/main.c
anilreddy3: > This is undoubtedly true, but why optimize for speed here? And even so, would `__aligned(64)` make a measurable difference? Have you run benchmarks? 😉
> We are certainly not going to add alignment directives to all variables passed to `memcpy()`.

Not really measured the performance difference. Okay, I will remove the directive.

core/arch/arm/plat-marvell/main.c
```diff
@@ -48,6 +48,8 @@
 #include <platform_config.h>
 #include <stdint.h>
 #include <string.h>
+#include <kernel/tee_common_otp.h>
+#include <io.h>
```
jforissier, line 52: Please order alphabetically

core/arch/arm/plat-marvell/main.c
```diff
@@ -108,3 +114,38 @@ void console_init(void)
...
+
+#ifdef CFG_HW_UNQ_KEY_SUPPORT
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	uint8_t hw_unq_key[sizeof(hwkey->data)];
+	void *fuse_huk1;
+	void *fuse_huk2;
+	uint64_t huk1 = 0;
+	uint64_t huk2 = 0;
+	uint8_t  byte;
```
jforissier, line 126: All variables have to be initialized: `= { };` `= NULL;` `= 0;`.

jforissier (commented):
Thanks for the update. With my last comments addressed:

core/arch/arm/plat-marvell/main.c
jforissier: `{ }` (brace, space, brace), not `{0}`. And `byte = 0`.

anilreddy3:
Thanks @jforissier for your review comments and the ACK, have updated the Tag

core/arch/arm/plat-marvell/main.c
```diff
@@ -108,3 +114,38 @@ void console_init(void)
...
+
+	huk1 = get_le64(fuse_huk1);
+	huk2 = get_le64(fuse_huk2);
+
+	for (byte = 0; byte <= 7; byte++) {
+		hw_unq_key[byte] = (uint8_t)(((huk1 >> (byte * 8))) & 0xFF);
+		hw_unq_key[8 + byte] = (uint8_t)(((huk2 >> (byte * 8))) & 0xFF);
+	}
+
+	memcpy(&hwkey->data[0], hw_unq_key, sizeof(hwkey->data));
```
jforissier, line 147: Sorry one more comment. Why not just do: 
```C
memcpy(hwkey->data, fuse_huk1, sizeof(hwkey->data));
```
?

ldts:
@anilreddy3 what makes this fuse a HUK? is the value protected somehow? could you explain a bit? I couldnt find any information on the web.
it being a fuse I wonder if it could be accessed from other exception levels or perhaps via JTAG and therefore causing a security threat.

I am raising it since I am having a similar issue with Zynqmp and the DNA eFUSE (unique id). In my case, the fuse is only readable in EL3 and via JTAG independently if the board has been put in secured state (issue 1). That problem aside what I am having to do is to have SPL read it and pass it to op-tee (S-EL1) via device tree (though I am not fully convinced about this strategy security wise since there is a small time gap where I suppose the memory could somehow be read before op-tee had the chance to delete the property from the device tree after reading it). Anyway, just sharing this since I am wondering about what security implications have you considered.

core/arch/arm/plat-marvell/main.c
anilreddy3: > Sorry one more comment. Why not just do:
> 
> ```c
> memcpy(hwkey->data, fuse_huk1, sizeof(hwkey->data));
> ```
> 
> ?

As huk1 and huk2 fuse addresses are contiguous yes we can. Thanks for the suggestion.

anilreddy3:
> @anilreddy3 what makes this fuse a HUK? is the value protected somehow? could you explain a bit? I couldnt find any information on the web.
> it being a fuse I wonder if it could be accessed from other exception levels or perhaps via JTAG and therefore causing a security threat.
> 
> I am raising it since I am having a similar issue with Zynqmp and the DNA eFUSE (unique id). In my case, the fuse is only readable in EL3 and via JTAG independently if the board has been put in secured state (issue 1). That problem aside what I am having to do is to have SPL read it and pass it to op-tee (S-EL1) via device tree (though I am not fully convinced about this strategy security wise since there is a small time gap where I suppose the memory could somehow be read before op-tee had the chance to delete the property from the device tree after reading it). Anyway, just sharing this since I am wondering about what security implications have you considered.

The eFUSE typically One Time Progromamble(OTP) memory array, which are used to store the cryptographic keys, one among them is HUK. These keys typically programmed at trust-fuse time on factory, and where HUK usually programmed with unique random number per device.
In my case these FUSE arrays are SECURE READ ONLY memory property set, so it can only accessible by Secure State of the processer, and so HUK can only accessible to trusted/secure software.

core/arch/arm/plat-marvell/main.c
jforissier: @anilreddy3 OK then please simplify the function as much as possible.

core/arch/arm/plat-marvell/main.c
```diff
@@ -108,3 +114,21 @@ void console_init(void)
...
 }
+
+#ifdef CFG_HW_UNQ_KEY_SUPPORT
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	void *fuse_huk = NULL;
+
+	fuse_huk = phys_to_virt(PLAT_MARVELL_FUSF_FUSE_BASE +
+							PLAT_MARVELL_FUSF_HUK_OFFSET,
+							MEM_AREA_IO_SEC);
```
jforissier, line 125: Please fix alignment and assign directly. Also how about simplifying the names to make the code easier to read?
```suggestion
	void *huk = phys_to_virt(FUSF_FUSE_BASE + FUSF_HUK_OFFSET,
				 MEM_AREA_IO_SEC);
```

jforissier (commented):
Final suggestions below, please fix at least the code alignment (I leave it to you to pick the shorter names or not). Then apply:

Thanks!

jforissier:
Checkpatch found another alignment issue: https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=897&view=logs&j=b2d025ba-e520-5b70-346d-474de3082414&t=69f02062-8a35-547d-8a2c-add7e74b55fc&l=16

anilreddy3:
> Final suggestions below, please fix at least the code alignment (I leave it to you to pick the shorter names or not). Then apply:
> `Reviewed-by: Jerome Forissier <jerome@forissier.org>`
> 
> Thanks!

Thanks @jforissier for your suggestions and review, and have updated the Tag

ldts:
@anilreddy3 what do you mean by the secure state of the processor? do you mean by S-EL1 or only after the board as being put in such a state that will only access signed images?

anilreddy3:
> @anilreddy3 what do you mean by the secure state of the processor? do you mean by S-EL1 or only after the board as being put in such a state that will only boot signed images? this distinction is relevant and important

Yes S-EL1.

ldts:
ok so during early development  I suppose someone could get hold of the board and print the board's HUK before bootloader authentication is enabled?  (I assume this board implements a hardware root of trust)

If that is the case, the behavior is pretty much the same we have with Xilinx ZynqMP when using the DNA eFUSE - and I dont think it is ideal (ideal being the way NXP does it, the user only given access to a _development_ HUK until the board is closed).

PR #4762, opened 2021-07-26
core_mmu_lpae: adding level 0 support for address space >= 40 bits

Izikd:
**Update:**

A new pull request  after the original one was stale and closed; https://github.com/OP-TEE/optee_os/pull/4675.

* Original commits were rebased over latest master branch.
* User mapping handling was fixed (new commit message below)
* Test with hello_world TA on Amazon dedicated platform

>     core: lpae: support user mapping when base level is 0
>     
>     User mapping (i.e. TAs) is expected to be at level 2,
>     so an level 1 entry points to this mapping.
>     If base level is 1, as it was supported, nothing changes.
>     If base level is 0 then an extra page is created at level 1,
>     so user mapping can be pointed from level 1 entry, as it's
>     supported by user mappings.

**Original message:**

A set of commits to add support for ARM MMU with CFG_LPAE_ADDR_SPACE_BITS>=40.
Till now OPTEE silently ignored that supporting >=40 bits of address space requires using level 0 page table.

Development stages:
1. References to l1_table were renamed to base_table, referencing from TF-A xlat_tables naming.
2. Importing few macros from TF-A xlat_tables_def.h for level 0 calculations.
3. Non-64bit "1 << x", causing bad calculations, were replaced with "BIT64(x)"
4. Build-time CORE_MMU_BASE_TABLE_LEVEL is determined to be 0 or 1 that affects few hard coded places that were referring "1" as the base level.
5. Most of the page table crunching functions (e.g. looping over levels) are magically working with the extra level 0.

References:
* ARMv8 page table granularity and required levels: https://developer.arm.com/documentation/den0024/a/The-Memory-Management-Unit/Translation-tables-in-ARMv8-A/Effect-of-granule-sizes-on-translation-tables


jforissier:
@Izikd thanks for the update.

As you can see in the CI log, there is an issue when NW virtualization is enabled. Also, I failed to boot with 44 bit address space (I just picked 44 as an example). Would you please check the following configurations with what we call the "QEMUv8 environment" (https://optee.readthedocs.io/en/latest/building/devices/qemu.html#qemu-v8) and make sure they work OK?
```
$ make -j10 run
$ make -j10 CFG_LPAE_ADDR_SPACE_BITS=44 run
$ make -j10 XEN_BOOT=y run
$ make -j10 CFG_LPAE_ADDR_SPACE_BITS=44 XEN_BOOT=y run
```
Note, you may need to run `make arm-tf-clean` when switching to/from `XEN_BOOT=y`.
For each configuration above, type `c` at the QEMU prompt then log in as user `test` (no password) and run `xtest`, it should end with a success message and no failures.

Izikd:
@jforissier thanks for the instructions!

Before your message I didn't notice that section at the documentation, so I started to reverse engineer the CI.
Currently I can reproduce the issue yes/no depending on the optee_os source code that I use.
Is this a good enough environment?

Something like:
```
docker -ti -v $PWD/optee_os:/root/optee_os jforissier/optee_os_ci:qemuv8_check

cd /root/optee_repo_qemu_v8/.repo/repo && git pull
cd /root/optee_repo_qemu_v8 && repo sync -j 10
mv /root/optee_repo_qemu_v8/optee_os /root/optee_repo_qemu_v8/optee_os_old
sudo ln -s /root/optee_os /root/optee_repo_qemu_v8/optee_os

touch /boot/vmlinuz-$(uname -r)
rm -rf /root/optee_repo_qemu_v8/out-br/build/optee_test*
make -C /root/optee_repo_qemu_v8/build arm-tf-clean

make -C /root/optee_repo_qemu_v8/build -j$(getconf _NPROCESSORS_ONLN) CFG_TEE_CORE_LOG_LEVEL=0 check XEN_BOOT=y
```

jforissier:
@Izikd yes that should work but will be harder to debug because `make run` is not directly supported in that environment. `make run` opens two terminals for the NW and SW consoles. Without `gnome-terminal` nor `xterm` you need to run `./soc_term/soc_term 54321` and `./soc_term/soc_term 54320` manually before QEMU is started plus a few things (install netcat and set `LAUNCH_TERMINAL=false`). Something like that (tested):
```
[First terminal]
$ docker run --rm -it -v $PWD/optee_os:/root/optee_os jforissier/optee_os_ci:qemuv8_check
# apt install netcat
# cd /root/optee_repo_qemu_v8/.repo/repo && git pull
# cd /root/optee_repo_qemu_v8 && repo sync -j 10
# mv /root/optee_repo_qemu_v8/optee_os /root/optee_repo_qemu_v8/optee_os_old
# sudo ln -s /root/optee_os /root/optee_repo_qemu_v8/optee_os

[Second terminal]
$ docker ps
... [copy container ID and use it instead of fff29bf11c00 below] ...
$ docker exec -it fff29bf11c00 bash
# /root/optee_repo_qemu_v8/soc_term/soc_term 54321
listening on port 54320

[Third terminal]
$ docker exec -it fff29bf11c00 bash
# /root/optee_repo_qemu_v8/soc_term/soc_term 54320
listening on port 54321

[First terminal]
# cd /root/optee_repo_qemu_v8/build
# make optee-os-clean arm-tf-clean
# make -j10 run LAUNCH_TERMINAL=false
(qemu) c

[Second terminal]
Welcome to Buildroot, type root or test to login
buildroot login: test
$ xtest
Run test suite with level=0

TEE test application started over default TEE instance
...
```

Izikd:
So actually I was referring to `make check` rather than `make run`, the `check` runs OK in detached mode.

Fixed the issues and executed successfully:
```
make check -j8 CFG_TEE_CORE_LOG_LEVEL=4

make arm-tf-clean
make check -j8 XEN_BOOT=y CFG_TEE_CORE_LOG_LEVEL=4
make check -j8 XEN_BOOT=y CFG_TEE_CORE_LOG_LEVEL=4 CFG_LPAE_ADDR_SPACE_BITS=44
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -817,19 +882,70 @@ void core_mmu_get_user_va_range(vaddr_t *base, size_t *size)
...
-		*base = (vaddr_t)user_va_idx << L1_XLAT_ADDRESS_SHIFT;
+		*base = (vaddr_t)user_va_idx << BASE_XLAT_ADDRESS_SHIFT;
 	if (size)
 		*size = BIT64(L1_XLAT_ADDRESS_SHIFT);
 }
 
+static uint64_t *core_mmu_get_user_mapping_entry(struct mmu_partition *prtn,
+						 unsigned int base_idx)
+{
+	uint64_t *tbl;
```
jenswi-linaro, line 902: Please initialize all local variables with a value.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -817,19 +882,70 @@ void core_mmu_get_user_va_range(vaddr_t *base, size_t *size)
...
+	assert(user_va_idx != -1);
+
+	tbl = prtn->base_tables[base_idx][get_core_pos()];
+	n = user_va_idx;
+
+	/*
+	 * If base level is 0, then user mapping entry is entry 0 in the level 1
+	 * table pointed by user_va_idx.
+	 */
+#if (CORE_MMU_BASE_TABLE_LEVEL == 0)
```
jenswi-linaro, line 919: We try to avoid `#if` and `#ifdef` inside the functions. There's a bit of the mix in the source, but this is the direction where working towards.
A plain `if (CORE_MMU_BASE_TABLE_LEVEL == 0)` should work here.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -817,19 +882,70 @@ void core_mmu_get_user_va_range(vaddr_t *base, size_t *size)
...
+	assert((tbl[n] & DESC_ENTRY_TYPE_MASK) == TABLE_DESC);
+
+	/* Copy bits 47:12 from tbl[n] to ntbl */
+	ntbl = tbl[n] & GENMASK_64(47, 12);
+
+#ifdef CFG_VIRTUALIZATION
```
jenswi-linaro, line 925: `if (IS_ENABLED(CFG_VIRTUALIZATION))` should work here.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -817,19 +882,70 @@ void core_mmu_get_user_va_range(vaddr_t *base, size_t *size)
...
+#endif
+
+	return &tbl[n];
+}
+
 bool core_mmu_user_mapping_is_active(void)
 {
 	bool ret;
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
+	uint64_t *entry;
```
jenswi-linaro, line 946: Please initialize all local variables with a value. Any value is OK as long as it isn't garbage from the stack.
This applies for the other function below that you've modified in this patch.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -484,6 +502,50 @@ void core_init_mmu_prtn(struct mmu_partition *prtn, struct tee_mmap_region *mm)
...
+ *
+ * If base table level is 1 - user_va_idx is already the index,
+ *                            so nothing to do.
+ * If base table level is 0 - user_va_idx is an level 0 index,
+ *                            so we need to create a level 1 table and TA entry
+ *                            will be the first entry there.
+ */
+static void core_init_mmu_prtn_ta(struct mmu_partition *prtn __maybe_unused)
+{
+#if (CORE_MMU_BASE_TABLE_LEVEL == 0)
```
jenswi-linaro, line 520: We normally try to avoid `#if` and `#ifdef` inside function, but this is OK since it's the entire body of the function.

jenswi-linaro (commented):
Comments for the commit "core: lpae: support user mapping when base level is 0"

jenswi-linaro:
The commit "core: lpae: support level 0 as base level" should probably be the last one in this PR.

Izikd:
Fixed the comments and and will make `core: lpae: support level 0 as base level` the last one on final squash and push force.

jenswi-linaro:
Looks good to me, please squash and apply:

jforissier:
There is a problem (data abort) with 32-bit TAs:
```
$ make check -j10 CFG_LPAE_ADDR_SPACE_BITS=44 COMPILE_S_USER=32 DUMP_LOGS_ON_ERROR=y
```

jforissier:
Also I noticed that execution is much, much slower than usual when pager is enabled and this feature is used, i.e.:
`CFG_WITH_PAGER=y` vs. `CFG_WITH_PAGER=y CFG_LPAE_ADDR_SPACE_BITS=44`. For example, `time xtest 4002` reports 6 seconds with pager only and 38 seconds with pager and 44 bit address space. Is this expected?

jenswi-linaro:
> Also I noticed that execution is much, much slower than usual when pager is enabled and this feature is used, i.e.:
> `CFG_WITH_PAGER=y` vs. `CFG_WITH_PAGER=y CFG_LPAE_ADDR_SPACE_BITS=44`. For example, `time xtest 4002` reports 6 seconds with pager only and 38 seconds with pager and 44 bit address space. Is this expected?

That's unexpected. I wonder if it's the same on real hardware.

Izikd:
> > Also I noticed that execution is much, much slower than usual when pager is enabled and this feature is used, i.e.:
> > `CFG_WITH_PAGER=y` vs. `CFG_WITH_PAGER=y CFG_LPAE_ADDR_SPACE_BITS=44`. For example, `time xtest 4002` reports 6 seconds with pager only and 38 seconds with pager and 44 bit address space. Is this expected?
> 
> That's unexpected. I wonder if it's the same on real hardware.

Maybe it's related to the new `core_mmu_get_user_mapping_entry()` that's being used during user mapping functions (e.g. `core_mmu_user_mapping_is_active()`, `core_mmu_get_user_map()`, `core_mmu_set_user_map()`).
With base level 0 it does that extra crunching that involves `phys_to_virt()`.

jforissier:
> > Also I noticed that execution is much, much slower than usual when pager is enabled and this feature is used, i.e.:
> > `CFG_WITH_PAGER=y` vs. `CFG_WITH_PAGER=y CFG_LPAE_ADDR_SPACE_BITS=44`. For example, `time xtest 4002` reports 6 seconds with pager only and 38 seconds with pager and 44 bit address space. Is this expected?
> 
> That's unexpected. I wonder if it's the same on real hardware.

It is much less dramatic, but still. On HiKey960: 0.53 vs. 1.04  seconds for test 4002.

jenswi-linaro:
It seems there's room for some optimizations, but I don't suppose this is a show stopper.
If you need that much va space you probably don't need to pager anyway.

Izikd:
> There is a problem (data abort) with 32-bit TAs:
> 
> ```
> $ make check -j10 CFG_LPAE_ADDR_SPACE_BITS=44 COMPILE_S_USER=32 DUMP_LOGS_ON_ERROR=y
> ```

After trying to debug and fix it, I'm not sure I deeply understand how AArch64 EL1 works with AArch32 EL0.
AArch64 EL1 should setup `TTBR0_EL1, TCR_EL1, etc.`, which have different meanings from AArch32 `TTBR0, TTBCR, etc.`.
Are we in a AArch64 MMU regime dictated by the AArch64 EL1 even if we're executing AArch32 EL0?

One obvious thing I had to fix is allowing TAs VA base be only in the range of [1GB, 4GB), as 32-bit TA can access VA only at the range [0GB, 4GB).
So it made me progress but now I'm stuck (not crash, but CPU 1 is reported to be stall by Linux kernel) in this test:
```
* regression_1013 Test concurrency with concurrent TA
o regression_1013.1 Using small concurrency TA
o regression_1013.1.1 Busy loop repeat 10000
```
I think this is the first test that uses CPU>0.

jenswi-linaro:
> > There is a problem (data abort) with 32-bit TAs:
> > ```
> > $ make check -j10 CFG_LPAE_ADDR_SPACE_BITS=44 COMPILE_S_USER=32 DUMP_LOGS_ON_ERROR=y
> > ```
> 
> After trying to debug and fix it, I'm not sure I deeply understand how AArch64 EL1 works with AArch32 EL0.
> AArch64 EL1 should setup `TTBR0_EL1, TCR_EL1, etc.`, which have different meanings from AArch32 `TTBR0, TTBCR, etc.`.
> Are we in a AArch64 MMU regime dictated by the AArch64 EL1 even if we're executing AArch32 EL0?

Yes, that's the way it has been until now at least. I'm not sure what effect the larger VA space has on AArch32 at EL0.

> 
> One obvious thing I had to fix is allowing TAs VA base be only in the range of [1GB, 4GB), as 32-bit TA can access VA only at the range [0GB, 4GB).

Makes sense.

> So it made me progress but now I'm stuck (not crash, but CPU 1 is reported to be stall by Linux kernel) in this test:
> 
> ```
> * regression_1013 Test concurrency with concurrent TA
> o regression_1013.1 Using small concurrency TA
> o regression_1013.1.1 Busy loop repeat 10000
> ```
> 
> I think this is the first test that uses CPU>0.

Yes. The stall is due to some CPU executing with non-secure interrupts masked for too long. Does the case complete?

Izikd:
> > So it made me progress but now I'm stuck (not crash, but CPU 1 is reported to be stall by Linux kernel) in this test:
> > ```
> > * regression_1013 Test concurrency with concurrent TA
> > o regression_1013.1 Using small concurrency TA
> > o regression_1013.1.1 Busy loop repeat 10000
> > ```   
> >   
> > I think this is the first test that uses CPU>0.
> 
> Yes. The stall is due to some CPU executing with non-secure interrupts masked for too long. Does the case complete?

No, exits with a timeout.
```
Running: time xtest ...
#################################################!!! Timeout
TIMEOUT - test case too long or hung? (last test started: regression_1013.1.1)
```

On serial1 side:
```
F/TC:? 1 plat_prng_add_jitter_entropy:72 0x1D
D/TC:? 1 tee_ta_close_session:512 csess 0xa8b1750e780 id 2
D/TC:? 1 tee_ta_close_session:531 Destroy session
D/TC:? 1 destroy_context:308 Destroy TA ctx (0xa8b1750d870)
D/TC:? 1 tee_ta_close_session:512 csess 0xa8b1750c5e0 id 3
D/TC:? 1 tee_ta_close_session:531 Destroy session
D/TC:? 1 destroy_context:308 Destroy TA ctx (0xa8b1750c580)
```

Izikd:
So I wasn't able to find the issue with `COMPILE_S_USER=32` (besides the one with 4GB limit).
If you have any more hints it will be helpful.

I've uploaded `core: lpae: search TAs VA base in the regions [1GB, 4GB)` which I think has a standalone importance, so this is why it's a separate commit that might be before any level 0 change.
The fixup for user mapping includes also caching user L1 entries instead of crunching the pointer each time (i.e. no more repetitive page table traversing by SW).
Please let me know if it helps performance.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -252,6 +251,7 @@ struct mmu_partition {
 	xlat_tbl_t *l2_ta_tables;
 	unsigned int xlat_tables_used;
 	unsigned int asid;
+	uint64_t *user_l1_entries[NUM_BASE_TABLES][CFG_TEE_CORE_NB_CORE];
```
jenswi-linaro, line 254: Would it be possible to use indexes with the type `uint8_t` instead of pointers in order to save some memory?

Izikd: Saving memory was my concern for not using such a solution in the first place :)
Do you mean indexes inside `xlat_tbl_t *xlat_tables;` array? Sounds good.
Can you confirm that it was indeed the root cause for the performance degradation? For some reason adding `time` to `xtest` doesn't show the time statistics for me.

jforissier:
> The fixup for user mapping includes also caching user L1 entries instead of crunching the pointer each time (i.e. no more repetitive page table traversing by SW).
> Please let me know if it helps performance.

I have tried commit edf4bd9 and unfortunately it crashes on boot:
```
$ make -j10 CFG_LPAE_ADDR_SPACE_BITS=44 CFG_TEE_CORE_LOG_LEVEL=2 CFG_WITH_PAGER=y run
...
I/TC: Secondary CPU 1 switching to normal world boot
E/TC:1 0
E/TC:1 0 Core data-abort at address 0xe119008 (translation fault)
E/TC:1 0  esr 0x96000007  ttbr0 0x0e12c100   ttbr1 0x00000000   cidr 0x0
E/TC:1 0  cpu #1          cpsr 0x400001c4
E/TC:1 0  x0  000000000e119008 x1  00000c1c9cff3070
E/TC:1 0  x2  0000000000000700 x3  0000008000000100
E/TC:1 0  x4  00000c1c9d01f13c x5  0000000000000000
E/TC:1 0  x6  0000000000000013 x7  00000c1c9cff5920
E/TC:1 0  x8  0000000000000070 x9  00000c1c9d013a00
E/TC:1 0  x10 0000000000000000 x11 00000c1c9cff5920
E/TC:1 0  x12 0000000000000000 x13 0000000000000000
E/TC:1 0  x14 0000000000000000 x15 0000000000000000
E/TC:1 0  x16 00000c1c9d03f138 x17 0000000000000000
E/TC:1 0  x18 0000000000000000 x19 0000000000000004
E/TC:1 0  x20 00000c1c9cff6078 x21 00000c1c9d01f110
E/TC:1 0  x22 00000c1c9d01f090 x23 00000c1c9d077f58
E/TC:1 0  x24 00000000ffffffff x25 00000c1c9d01f100
E/TC:1 0  x26 0000000000000000 x27 00000000ffff0008
E/TC:1 0  x28 00000c1c9cff6f58 x29 00000c1c9d077d60
E/TC:1 0  x30 00000c1c9cfeabd0 elr 00000c1c9cfeabe0
E/TC:1 0  sp_el0 00000c1c9d077d60
E/TC:1 0 TEE load address @ 0xc1c9cfe2000
E/TC:1 0 Call stack:
E/TC:1 0  0x00000c1c9cfeabe0 core_mmu_user_mapping_is_active at optee_os/core/arch/arm/mm/core_mmu_lpae.c:1009
E/TC:1 0  0x00000c1c9d0467f4 is_user_mode_ctx at optee_os/core/include/kernel/user_mode_ctx.h:19
E/TC:1 0  0x00000c1c9d049140 pseudo_ta_enter_open_session at optee_os/core/kernel/pseudo_ta.c:149
E/TC:1 0  0x00000c1c9d046490 tee_ta_open_session at optee_os/core/kernel/tee_ta_manager.c:743
E/TC:1 0  0x00000c1c9d04d124 entry_open_session at optee_os/core/tee/entry_std.c:372
E/TC:1 0  0x00000c1c9d03c148 std_smc_entry at optee_os/core/arch/arm/kernel/thread_optee_smc.c:194
E/TC:1 0 Panic 'unhandled pageable abort' at core/arch/arm/kernel/abort.c:553 <abort_handler>
E/TC:1 0 TEE load address @ 0xc1c9cfe2000
E/TC:1 0 Call stack:
E/TC:1 0  0x00000c1c9cfe80a0 print_kernel_stack at optee_os/core/arch/arm/kernel/unwind_arm64.c:80
E/TC:1 0  0x00000c1c9cfec44c __do_panic at optee_os/core/kernel/panic.c:24
E/TC:1 0  0x00000c1c9cfe7b0c get_fault_type at optee_os/core/arch/arm/kernel/abort.c:476
E/TC:1 0  0x00000c1c9cfe5a74 el1_sync_abort at optee_os/core/arch/arm/kernel/thread_a64.S:747
```

Same crash with pager disabled :-/

core/arch/arm/mm/core_mmu_lpae.c
jenswi-linaro: > Saving memory was my concern for not using such a solution in the first place :)

Aha

> Do you mean indexes inside `xlat_tbl_t *xlat_tables;` array? Sounds good.

Yes :-)

> Can you confirm that it was indeed the root cause for the performance degradation? For some reason adding `time` to `xtest` doesn't show the time statistics for me.

Sorry, I haven't tested this yet.

Izikd:
@jforissier  Very strange that I didn't see it before. Now I see it on my side too.
The plot thickens... it used to allocate TAs VA on base_table which are already separated per-core, while all other TEE tables are initialized once and copied to other cores.
Now we have a situation where we need to place TAs page inside a level 1 table which isn't a base table anymore, so it implies that the same level 1 table, which holds also TEE mappings, can't be just serve all cores as each core needs a different TA table pointer... :(
What are your suggestions?
Should this be executed for each core separately and not copied? It will be a shame to hold multiple tables for the parts that can be copied (e.g. TEE level 2/3 tables).
```
	for (n = 0; !core_mmap_is_end_of_table(mm + n); n++) {
		debug_print(" %010" PRIxVA " %010" PRIxPA " %10zx %x",
			    mm[n].va, mm[n].pa, mm[n].size, mm[n].attr);

		if (!IS_PAGE_ALIGNED(mm[n].pa) || !IS_PAGE_ALIGNED(mm[n].size))
			panic("unaligned region");
	}

	/* Clear table before use */
	memset(prtn->base_tables, 0, sizeof(base_xlation_table));

	for (n = 0; !core_mmap_is_end_of_table(mm + n); n++)
		if (!core_mmu_is_dynamic_vaspace(mm + n))
			core_mmu_map_region(prtn, mm + n);

	/*
	 * Primary mapping table is ready at index `get_core_pos()`
	 * whose value may not be ZERO. Take this index as copy source.
	 */
	for (n = 0; n < CFG_TEE_CORE_NB_CORE; n++) {
		if (n == get_core_pos())
			continue;

		memcpy(prtn->base_tables[0][n],
		       prtn->base_tables[0][get_core_pos()],
		       XLAT_ENTRY_SIZE * NUM_BASE_LEVEL_ENTRIES);
	}
```

jforissier:
@Izikd I am not too comfortable with the MMU code unfortunately, so I can only help with testing various configurations and mostly cosmetic code reviews :-/

Izikd:
* Changed `user_l1_entries` to be `uint8_t` indexes, and ifdefed this for base level 0 only, because if base level is 1 then the entries are from the base table and not from the xlat tables (hence indexes can't work).
* Added table copy function. All tables are copied from the initiator core to other cores as before, but at need (e.g. user table level 1, but base table is level 0) an entry can be applied this copy function that creates a copy of the original table.
* I think 32-bit TAs are working now, but maybe I'm wrong as I'm not fully sure on which config changes `make clean`/`make arm-tf-clean` is needed...

jforissier:
@Izikd thanks for the update! I'm going to test it.

>  I'm not fully sure on which config changes `make clean`/`make arm-tf-clean` is needed...

In theory, config changes do not require cleaning (whether you change `CFG_*` or `COMPILE_{S,NS}_{USER,KERNEL} it should be OK). Only `XEN_BOOT=y/n` introduced recently is thought to cause problems and requires `make arm-tf-clean`.
That being said when in doubt I often do `rm -rf ../out-br/build/optee* ../optee_os/out` to clean the OP-TEE components only. It does not take too long to rebuild.

jforissier:
@Izikd I have run quite a few tests with combinations of 32/44 bit space, pager/no pager,  arm32/arm64 TAs and this looks solid. Well done! ;) So this is:

`Tested-by: Jerome Forissier <jerome@forissier.org> (vexpress-qemu_armv8a)`

Shouldn't commit "core: lpae: support level 0 as base level" come last in the series? I think that Tested-by would apply ideally to that commit.

Izikd:
@jforissier Great news!
Indeed reordered the commits and squashed the fixup commits.

Izikd:
> Also I noticed that execution is much, much slower than usual when pager is enabled and this feature is used, i.e.:
> `CFG_WITH_PAGER=y` vs. `CFG_WITH_PAGER=y CFG_LPAE_ADDR_SPACE_BITS=44`. For example, `time xtest 4002` reports 6 seconds with pager only and 38 seconds with pager and 44 bit address space. Is this expected?

@jforissier Forgot to ask about it. Did latest patches resolve this?

jforissier:
> > Also I noticed that execution is much, much slower than usual when pager is enabled and this feature is used, i.e.:
> > `CFG_WITH_PAGER=y` vs. `CFG_WITH_PAGER=y CFG_LPAE_ADDR_SPACE_BITS=44`. For example, `time xtest 4002` reports 6 seconds with pager only and 38 seconds with pager and 44 bit address space. Is this expected?
> 
> @jforissier Forgot to ask about it. Did latest patches resolve this?

No, still the same more or less. But execution time largely depends on the amount pf paging required, if the pager pool size is increased (by decreasing `CFG_CORE_HEAP_SIZE`) the performance hit is smaller. So perhaps part of the reason the code is slower is because it is a bit bigger? As @jenswi-linaro mentioned, `CFG_WITH_PAGER=y` with `CFG_LPAE_ADDR_SPACE_BITS>=40` is probably not a very reasonable use case.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -484,6 +484,45 @@ void core_init_mmu_prtn(struct mmu_partition *prtn, struct tee_mmap_region *mm)
 	}
 }
 
+/*
+ * In order to support 32-bit TAs we will have to find
+ * a TAs VA base in the region (0, 4GB).
```
etienne-lms, line 526: ` * a user VA base in the region [0, 4GB].`

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -484,6 +484,45 @@ void core_init_mmu_prtn(struct mmu_partition *prtn, struct tee_mmap_region *mm)
...
+ * Due to OP-TEE design limitation, TAs page table should be an entry
+ * inside a level 1 page table.
+ *
+ * Available options are only these:
+ * - base level 0 entry 0 - [0GB, 512GB)
```
etienne-lms, line 531: replace closing parenthesis `)` with closing square brace`]`.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -507,13 +546,7 @@ void core_init_mmu(struct tee_mmap_region *mm)
...
 	}
 
-	for (n = 1; n < NUM_BASE_LEVEL_ENTRIES; n++) {
-		if (!default_partition.base_tables[0][0][n]) {
-			user_va_idx = n;
-			break;
-		}
-	}
-	assert(user_va_idx != -1);
+	core_init_mmu_prtn_ta_va_find(&default_partition);
```
etienne-lms, line 706: could shorten function label to `set_user_va_idx()`.

etienne-lms (commented):
Comments for commit "core: lpae: search TAs VA base in the regions [1GB, 4GB)".

By the way, is closing `)` means an excluded area upper address?
I would tend to use either `[1GB, 4GB]` or, if explicit exclusion is preferred, `[1GB 4GB[`.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -769,6 +763,33 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
...
+
+	if ((entry & DESC_ENTRY_TYPE_MASK) != TABLE_DESC ||
+	    level >= XLAT_TABLE_LEVEL_MAX)
+		return NULL;
+
+	/* Get original table */
+	pa = entry & OUTPUT_ADDRESS_MASK;
+
+	if (!IS_ENABLED(CFG_VIRTUALIZATION) ||
+	    prtn == &default_partition)
```
etienne-lms, line 959: fits in a single line:
```c
	if (!IS_ENABLED(CFG_VIRTUALIZATION) || prtn == &default_partition)
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -769,6 +763,33 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
...
+					     uint64_t entry)
+{
+	paddr_t pa = 0;
+	void *va = NULL;
+
+	if ((entry & DESC_ENTRY_TYPE_MASK) != TABLE_DESC ||
+	    level >= XLAT_TABLE_LEVEL_MAX)
+		return NULL;
+
+	/* Get original table */
```
etienne-lms, line 955: could remove this inline comment.

etienne-lms (commented):
For commit "core: lpae: add internal core_mmu_xlat_table_entry_pa2va()":
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with minor comments.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -763,6 +766,25 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
 	return ret;
 }
 
+static uint64_t *core_mmu_xlat_table_alloc(struct mmu_partition *prtn)
+{
+	uint64_t *new_table;
+
+	if (prtn->xlat_tables_used >= MAX_XLAT_TABLES) {
+		DMSG("out of xlat tables to allocate! total is %u",
```
etienne-lms, line 926: No need for bang `!` character.
Suggestion: use `EMSG()` and with a short message like `EMSG("%u xlat tables exhausted", MAX_XLAT_TABLES);`

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -841,6 +859,56 @@ bool core_mmu_entry_to_finer_grained(struct core_mmu_table_info *tbl_info,
 	return true;
 }
 
+/*
+ * For a table entry that points to a table - allocate and copy to
+ * a new pointed table. Does this only for the requested entry,
```
etienne-lms, line 1016: s/Does/Do/ or maybe `This is done for ...`.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -841,6 +859,56 @@ bool core_mmu_entry_to_finer_grained(struct core_mmu_table_info *tbl_info,
...
+	uint64_t *orig_table = NULL;
+	uint64_t *new_table = NULL;
+	uint64_t *entry = NULL;
+	struct mmu_partition *prtn = NULL;
+
+#ifdef CFG_VIRTUALIZATION
+	prtn = tbl_info->prtn;
+#else
+	prtn = &default_partition;
+#endif
```
etienne-lms, line 1033: prefer `if (IS_ENABLED(CFG_VIRTUALIZATION))`

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -841,6 +859,56 @@ bool core_mmu_entry_to_finer_grained(struct core_mmu_table_info *tbl_info,
...
+	/* Nothing to do for non-table entries */
+	if ((*entry & DESC_ENTRY_TYPE_MASK) != TABLE_DESC ||
+	    tbl_info->level >= XLAT_TABLE_LEVEL_MAX)
+		return true;
+
+	new_table = core_mmu_xlat_table_alloc(prtn);
+	if (!new_table)
+		return false;
+
+	/* Get original table */
```
etienne-lms, line 1050: Are this inline comment and to next one (line 903) really useful?

etienne-lms (commented):
Comments for commit "core: lpae: add internal core_mmu_entry_copy()".

I would rephrase a bit the commit log:
```patch
-core_mmu_entry_copy() to take a table entry that point to other
+core_mmu_entry_copy() takes a table entry that point to other
 table, allocate a new table, copy the content of the original table,
 and eventually make the higher level table point to the new table.
-This function is useful few some table is copied from core to core.
+This function is useful to copy mapping tables from core to core.
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -530,6 +588,94 @@ static void core_init_mmu_prtn_ta_va_find(struct mmu_partition *prtn)
...
+		 * Now base level table should be ready with a table descriptor
+		 */
+		assert((tbl[0] & DESC_ENTRY_TYPE_MASK) == TABLE_DESC);
+
+		tbl = core_mmu_xlat_table_entry_pa2va(prtn, 0, tbl[0]);
+
+		assert(tbl);
+	}
+
+	idx = ((uintptr_t)&tbl[user_va_idx] -
```
etienne-lms, line 656: (nitpicking) would prefer:
```c
 	idx = ((uintptr_t)&tbl[user_va_idx] - (uintptr_t)prtn->xlat_tables) /
	      sizeof(xlat_tbl_t);
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -961,10 +1133,13 @@ bool core_mmu_user_mapping_is_active(void)
 void core_mmu_get_user_map(struct core_mmu_user_map *map)
 {
 	struct mmu_partition *prtn = get_prtn();
+	uint64_t *entry = NULL;
 
 	assert(user_va_idx != -1);
```
etienne-lms, line 1142: Can remove this test since needed only from `core_mmu_get_user_mapping_entry()` and already asserted there.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -945,13 +1093,37 @@ void core_mmu_get_user_va_range(vaddr_t *base, size_t *size)
...
+}
+
 bool core_mmu_user_mapping_is_active(void)
 {
-	bool ret;
+	bool ret = false;
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
+	uint64_t *entry = NULL;
 
 	assert(user_va_idx != -1);
```
etienne-lms, line 1126: Can remove this test since needed only from `core_mmu_get_user_mapping_entry()` and already asserted there.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -975,9 +1150,11 @@ void core_mmu_get_user_map(struct core_mmu_user_map *map)
 
 void core_mmu_set_user_map(struct core_mmu_user_map *map)
 {
-	uint64_t ttbr;
+	uint64_t ttbr = 0;
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
 	struct mmu_partition *prtn = get_prtn();
+	uint64_t *entries[NUM_BASE_TABLES] = {};
```
etienne-lms, line 1160: s/`{}`/`{ }`/   (with a space char in between)

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -987,23 +1164,22 @@ void core_mmu_set_user_map(struct core_mmu_user_map *map)
...
 		ttbr |= ((uint64_t)map->asid << TTBR_ASID_SHIFT);
 		write_ttbr0_64bit(ttbr);
 		isb();
 	} else {
-		prtn->base_tables[0][get_core_pos()][user_va_idx] = 0;
-#ifdef CFG_CORE_UNMAP_CORE_AT_EL0
-		prtn->base_tables[1][get_core_pos()][user_va_idx] = 0;
-#endif
+		for (i = 0; i < NUM_BASE_TABLES; i++)
+			*entries[i] = 0;
```
etienne-lms, line 1185: `INVALID_DESC` isntead of `0`?

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -1061,33 +1240,34 @@ void core_mmu_get_user_map(struct core_mmu_user_map *map)
 
 void core_mmu_set_user_map(struct core_mmu_user_map *map)
 {
-	uint64_t ttbr;
+	uint64_t ttbr = 0;
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
 	struct mmu_partition *prtn = get_prtn();
+	uint64_t *entries[NUM_BASE_TABLES] = {};
```
etienne-lms, line 1250: s/{}/{ }/

etienne-lms (commented):
Comments for commit "core: lpae: support user mapping when base level is 0".

core/arch/arm/mm/core_mmu_lpae.c
Izikd: Unfortunately we can't do that, as existence of `prtn` inside `typeof(tbl_info)` is ifdefed already.

core/arch/arm/mm/core_mmu_lpae.c
Izikd: `)` means excluding in this case.

core/arch/arm/mm/core_mmu_lpae.c
Izikd: `)` means excluding in this case. Changed to `[1GB, 4GB)`

core/arch/arm/mm/core_mmu_lpae.c
Izikd: Fixed, but it will likely to fail checkpatch...
```
$ checkpatch.pl -g HEAD~
ERROR: space prohibited after that '-' (ctx:WxW)
#229: FILE: core/arch/arm/mm/core_mmu_lpae.c:649:
+       idx = ((uintptr_t)&tbl[user_va_idx] - (uintptr_t)prtn->xlat_tables) /
                                            ^

total: 1 errors, 0 warnings, 0 checks, 396 lines checked
```

Izikd:
@etienne-lms thanks for the deep review!

I think only comment left open is:
> By the way, is closing ) means an excluded area upper address?
> I would tend to use either [1GB, 4GB] or, if explicit exclusion is preferred, [1GB 4GB[.
So yes, `)`/`(` means excluded in these cases.
I don't recall seeing `[1GB 4GB[` as a syntax for exclusion, can you point me to some reference?

jforissier:
> I don't recall seeing `[1GB 4GB[` as a syntax for exclusion, can you point me to some reference?

https://en.wikipedia.org/wiki/ISO_31-11#Sets -- both `[` and `)` are acceptable as per this standard. Actually for me it is the opposite situation compared to you: I have never used `)` for exclusion, includling when I was studying math. I suppose it depends which country you are from. The problem with `(` and `)` is that `(a , b)` is ambiguous: open interval or ordered pair?
Anyways I think we can easily accept both notations here.

Izikd:
@jforissier TIL! Thanks for the details!
Updated the commits with the suggested syntax.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -248,6 +248,9 @@ static void *core_mmu_xlat_table_entry_pa2va(struct mmu_partition *prtn,
 					     unsigned int level,
 					     uint64_t entry);
 
+static bool core_mmu_entry_copy(struct core_mmu_table_info *tbl_info,
```
jenswi-linaro, line 280: We try to organize static functions in an order so we can avoid forward declarations.

jenswi-linaro (commented):
One comment the commit "core: lpae: add internal core_mmu_entry_copy()".

jenswi-linaro:
For the commit "core: lpae: search TAs VA base in the regions [1GB, 4GB[", please apply:

jenswi-linaro:
For "core: lpae: add internal core_mmu_entry_copy()" please apply:

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -737,7 +741,8 @@ bool core_mmu_entry_to_finer_grained(struct core_mmu_table_info *tbl_info,
 #endif
 	assert(prtn);
 
-	if (tbl_info->level >= 3 || idx > tbl_info->num_entries)
+	if (tbl_info->level >= XLAT_TABLE_LEVEL_MAX ||
+	    idx > tbl_info->num_entries)
```
etienne-lms, line 1023: by the way, i wonder if this is bugged. Shouldn't it  be `idx >= tbl_info->num_entries)`?

etienne-lms (commented):
Comments for commit "core: lpae: add internal core_mmu_entry_copy()" and 
commit "core: lpae: replace "3" with XLAT_TABLE_LEVEL_MAX", despite my R-b tag (which still stands, a specific commit should fix it IMO).

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "core: lpae: support user mapping when base level is 0" with minor comment addressed or not.

`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "core: lpae: search TAs VA base in the regions [1GB, 4GB[".

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -615,6 +672,93 @@ static void set_user_va_idx(struct mmu_partition *prtn)
 	assert(user_va_idx != -1);
 }
 
+/*
+ * TAs page table is an entry inside level 1 page table.
```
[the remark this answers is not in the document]
Izikd: better, indeed. done.

core/arch/arm/mm/core_mmu_lpae.c
Izikd: agree. maybe this bug is the origin for a similar bug introduced into the _copy() function above :)

Izikd:
@etienne-lms, fixed and added a new commit:
```
core: lpae: fix 'idx' boundary check in core_mmu_entry_to_finer_grained()
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -474,6 +492,55 @@ static void *core_mmu_xlat_table_entry_pa2va(struct mmu_partition *prtn,
...
+	    tbl_info->level >= XLAT_TABLE_LEVEL_MAX)
+		return true;
+
+	new_table = core_mmu_xlat_table_alloc(prtn);
+	if (!new_table)
+		return false;
+
+	orig_table = core_mmu_xlat_table_entry_pa2va(prtn,
+						     tbl_info->level,
+						     *entry);
```
etienne-lms, line 563: minor: prefer indent as:
```c
	orig_table = core_mmu_xlat_table_entry_pa2va(prtn, tbl_info->level,
						     *entry);
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -615,6 +672,93 @@ static void set_user_va_idx(struct mmu_partition *prtn)
...
+
+		/*
+		 * If this isn't the core that created the initial tables
+		 * mappings, then the level 1 table must be copied,
+		 * as it will hold pointer to the user mapping table
+		 * that changes per core.
+		 */
+		if (core != get_core_pos()) {
+			ret = core_mmu_entry_copy(&tbl_info, 0);
+			assert(ret);
```
etienne-lms, line 724: If xlat tables pool is exhausted, `core_mmu_entry_copy()` returns `false`. We can `assert()` here, must use an non debug test: `if (!core_mmu_entry_copy(&tbl_info, 0)) { panic(); }`

Note I think it would be better if `core_mmu_entry_copy()` could return a `TEE_Result` value instead of a boolean but that's not a strong opinion (as i.e. `core_mmu_entry_to_finer_grained()` already returns a boolean).

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -615,6 +672,93 @@ static void set_user_va_idx(struct mmu_partition *prtn)
...
+		}
+
+		ret = core_mmu_entry_to_finer_grained(&tbl_info, 0, true);
+		assert(ret);
```
etienne-lms, line 728: Same comment: can't `assert()`, must always check the result.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -615,6 +672,93 @@ static void set_user_va_idx(struct mmu_partition *prtn)
...
+
+		/*
+		 * Now base level table should be ready with a table descriptor
+		 */
+		assert((tbl[0] & DESC_ENTRY_TYPE_MASK) == TABLE_DESC);
+
+		tbl = core_mmu_xlat_table_entry_pa2va(prtn, 0, tbl[0]);
+
```
etienne-lms, line 736: minor: would prefer without the above empty line.

etienne-lms (commented):
For commit "core: lpae: add internal core_mmu_entry_copy()":
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment addressed or not.

For commit "core: lpae: fix 'idx' boundary check in core_mmu_entry_to_finer_grained()":

For commit "core: lpae: add internal core_mmu_entry_copy()":
The commit introduces a static function that is not called. Build may fail for this commit. Either merge it into the next commit (that uses the function) or add an `__unsued` attribute to the function in the commit and remove it in the next commit that makes this function being called.

Comment for commit "core: lpae: support user mapping when base level is 0".
Please remove my R-b tag for that commit until my comment is addressed (rejected or fixed)

Maybe remove Jerome's `Tested-by` tag from last commit, unless @jforissier tested the latest series since it was updated.

jforissier:
> For commit "core: lpae: add internal core_mmu_entry_copy()":
> The commit introduces a static function that is not called. Build may fail for this commit. Either merge it into the next commit (that uses the function) 

+1 

> Maybe remove Jerome's `Tested-by` tag from last commit, unless @jforissier tested the latest series since it was updated.

Yes, I will test again as soon as the PR passes the CI build phase.

jenswi-linaro:
For the commit "core: lpae: fix 'idx' boundary check in core_mmu_entry_to_finer_grain… " please apply:

core/arch/arm/mm/core_mmu_lpae.c
Izikd: replaced assert to panic. didn't change the return value.

Izikd:
I think I addressed all comments, hope didn't miss anything :)
Notice that checkpatch will fail due to a previous request to write this line in this format:
(IMHO, it's some checkpatch false positive)
```
$ checkpatch.pl -g HEAD~
ERROR: space prohibited after that '-' (ctx:WxW)
#229: FILE: core/arch/arm/mm/core_mmu_lpae.c:733:
+       idx = ((uintptr_t)&tbl[user_va_idx] - (uintptr_t)prtn->xlat_tables) /
                                            ^

total: 1 errors, 0 warnings, 0 checks, 396 lines checked
```

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -726,7 +730,6 @@ static void core_init_mmu_prtn_ta_core(struct mmu_partition *prtn
 		assert((tbl[0] & DESC_ENTRY_TYPE_MASK) == TABLE_DESC);
 
 		tbl = core_mmu_xlat_table_entry_pa2va(prtn, 0, tbl[0]);
-
```
etienne-lms: (nit) this change belongs to commit "core: lpae: support user mapping when base level is 0".

Izikd:
@etienne-lms oops, fixed.

etienne-lms:
Looks all good to me. 
I agree the checkpatch error is a false positive.
Can be merged IMO even if few review tags could be added.

@Izikd, you can add my R-b tag to commit "core: lpae: support user mapping when base level is 0".

@jenswi-linaro, i guess your R-b tag applies to commit "core: lpae: add internal core_mmu_xlat_table_entry_pa2va()"

In case it helps, `Tested-by: Etienne Carriere <etienne.carriere@linaro.org> (stm32mp1, qemuv8)`.

jenswi-linaro:
> @jenswi-linaro, i guess your R-b tag applies to commit "core: lpae: add internal core_mmu_xlat_table_entry_pa2va()"

Sure.

Izikd:
@etienne-lms @jenswi-linaro, I've applied the tags.

etienne-lms (approved):
Approving changes despite CI checkpatch failure.

jforissier:
Please add: `Tested-by: Jerome Forissier <jerome@forissier.org> (vexpress-qemu_armv8a)` to the last commit, thanks!

Izikd:
@jforissier @etienne-lms , rebased and applied your T-b tags to last commit.

jforissier:
@Izikd thanks for your work addressing all the comments and issues, I am merging this now.

PR #4766, opened 2021-07-28
arm64: format call stack addresses on 8 characters by default

jforissier:
print_stack_arm64() currently shows the full 64 bit addresses in
hexadecimal with leading zeros (0x0000000000000000). This is a bit hard
to read and is not necessary since virtual addresses are typically 32
or 36 bits (defined by CFG_LPAE_ADDR_SPACE_BITS), and in any case
nowhere near 64 bits.
Therefore, use a 32-bit format by default (0x00000000) and expand the
width as necessary. The new format is already used in ldelf to print
the region addresses so this changes brings consistency.


jenswi-linaro:

PR #4769, opened 2021-07-28
Update MBed TLS to 2.27.0

jforissier:
Updating the MBed TLS library to the latest stable release from upstream (2.27.0). The procedure described at https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#import-branches was followed. See also the description of the first commit. A number of commits from branch `import/mbedtls-2.22.0` were left aside when they were either already present upstream, or related to code outside of the updated directory (`lib/libmbedtls/mbedtls`) such as for example `lib/libmbedtls/core` or OP-TEE makefiles.


jforissier:
Sorry wrong target branch :-/

PR #4772, opened 2021-07-29
ta: pkcs11: Add operation state in session

ruchi393:
Add more operation states to take care of scenarios like failure
of an incremental (update) operation if a one-shot/final operation
has been started.

Signed-off-by: Ruchika Gupta <ruchika.gupta@linaro.org>

### DIFF
--- a/ta/pkcs11/src/pkcs11_attributes.c
+++ b/ta/pkcs11/src/pkcs11_attributes.c
@@ -87,8 +87,9 @@ check_mechanism_against_processing(struct pkcs11_session *session,
 		    !session->processing->relogged)
 			return PKCS11_CKR_USER_NOT_LOGGED_IN;
 
-		if (session->processing->updated) {
-			EMSG("Cannot perform one-shot on updated processing");
+		if (session->processing->step == PKCS11_FUNC_STEP_UPDATE ||
+		    session->processing->step == PKCS11_FUNC_STEP_FINAL) {
+			EMSG("Cannot perform one-shot on active processing");
 			return PKCS11_CKR_OPERATION_ACTIVE;
 		}
 
@@ -99,6 +100,12 @@ check_mechanism_against_processing(struct pkcs11_session *session,
 		if (session->processing->always_authen &&
 		    !session->processing->relogged)
 			return PKCS11_CKR_USER_NOT_LOGGED_IN;
+
+		if (session->processing->step == PKCS11_FUNC_STEP_ONESHOT ||
+		    session->processing->step == PKCS11_FUNC_STEP_FINAL) {
+			EMSG("Cannot perform update on finalized processing");
+			return PKCS11_CKR_OPERATION_ACTIVE;
+		}
 
 		allowed = !mechanism_is_one_shot_only(mechanism_type);
 		break;
@@ -118,6 +125,10 @@ check_mechanism_against_processing(struct pkcs11_session *session,
 		    !session->processing->relogged)
 			return PKCS11_CKR_USER_NOT_LOGGED_IN;
 
+		if (session->processing->step == PKCS11_FUNC_STEP_ONESHOT) {
+			EMSG("Cannot perform final on oneshot processing");
+			return PKCS11_CKR_OPERATION_ACTIVE;
+		}
 		return PKCS11_CKR_OK;
 
 	default:
--- a/ta/pkcs11/src/pkcs11_token.h
+++ b/ta/pkcs11/src/pkcs11_token.h
@@ -135,7 +135,7 @@ enum pkcs11_proc_state {
  * @mecha_type - mechanism type of the active processing
  * @always_authen - true if user need to login before each use
  * @relogged - true once client logged since last operation update
- * @updated - true once an active operation is updated
+ * @op_step - last active operation step - update, final or one-shot
  * @tee_op_handle - handle on active crypto operation or TEE_HANDLE_NULL
  * @tee_hash_algo - hash algorithm identifier.
  * @tee_hash_op_handle - handle on active hashing crypto operation or
@@ -145,9 +145,9 @@ enum pkcs11_proc_state {
 struct active_processing {
 	enum pkcs11_proc_state state;
 	uint32_t mecha_type;
+	enum processing_step step;
 	bool always_authen;
 	bool relogged;
-	bool updated;
 	TEE_OperationHandle tee_op_handle;
 	uint32_t tee_hash_algo;
 	TEE_OperationHandle tee_hash_op_handle;
--- a/ta/pkcs11/src/processing.c
+++ b/ta/pkcs11/src/processing.c
@@ -759,11 +759,18 @@ enum pkcs11_rc entry_processing_step(struct pkcs11_client *client,
 
 	if (rc == PKCS11_CKR_OK && (step == PKCS11_FUNC_STEP_UPDATE ||
 				    step == PKCS11_FUNC_STEP_UPDATE_KEY)) {
-		session->processing->updated = true;
+		session->processing->step = PKCS11_FUNC_STEP_UPDATE;
 		DMSG("PKCS11 session%"PRIu32": processing %s %s",
 		     session->handle, id2str_proc(mecha_type),
 		     id2str_function(function));
 	}
+
+	if (rc == PKCS11_CKR_BUFFER_TOO_SMALL &&
+	    step == PKCS11_FUNC_STEP_ONESHOT)
+		session->processing->step = PKCS11_FUNC_STEP_ONESHOT;
+
+	if (rc == PKCS11_CKR_BUFFER_TOO_SMALL && step == PKCS11_FUNC_STEP_FINAL)
+		session->processing->step = PKCS11_FUNC_STEP_FINAL;
 
 out:
 	switch (step) {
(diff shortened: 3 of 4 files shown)


ruchi393:
I couldn't find this explicitly stated in PKCS#11 spec but when running pkcs11-test, I came across few failures which are dependent on this state maintenance. Hence raising the PR as RFC to start a discussion if we need should cater to such errors.

List of pkcs11-tests which fail without this patch are:
[  FAILED  ] Ciphers/SecretKeyTest.EncryptModePolicing2/0, where GetParam() = "AES-ECB"
[  FAILED  ] Ciphers/SecretKeyTest.EncryptModePolicing2/1, where GetParam() = "AES-CBC"
[  FAILED  ] Digests/DigestTest.DigestFinalIntersperse/0, where GetParam() = "MD5"
[  FAILED  ] Digests/DigestTest.DigestFinalIntersperse/1, where GetParam() = "SHA-1"
[  FAILED  ] Digests/DigestTest.DigestFinalIntersperse/2, where GetParam() = "SHA-256"
[  FAILED  ] Digests/DigestTest.DigestFinalIntersperse/3, where GetParam() = "SHA-384"
[  FAILED  ] Digests/DigestTest.DigestFinalIntersperse/4, where GetParam() = "SHA-512"

vesajaaskelainen:
I was looking at what PKCS#11 3.0 brings up and there is something that might be a good to be aware -- cancelling of operations / session:
https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976627

vesajaaskelainen:
Tried to go thru the both 2.40 and 3.0 but did not spot anything.

One option I suppose is to ask from OASIS?
https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=pkcs11

Another option is to survey different implementations how they behave. Have you checked others?

I suppose clarification this in specification would be best so communication with OASIS probably would work best.

ruchi393:
> Tried to go thru the both 2.40 and 3.0 but did not spot anything.
> 
> One option I suppose is to ask from OASIS?
> https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=pkcs11
> 
> Another option is to survey different implementations how they behave. Have you checked others?
I did check SoftHSM. It doesn't seem to implement the additional states. 

> 
> I suppose clarification this in specification would be best so communication with OASIS probably would work best.
Ok

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -118,6 +126,11 @@ check_mechanism_against_processing(struct pkcs11_session *session,
 		    !session->processing->relogged)
 			return PKCS11_CKR_USER_NOT_LOGGED_IN;
 
+		if (session->processing->op_state == PKCS11_OP_FINAL_ONESHOT) {
+			EMSG("Cannot perform update on finalized on oneshot \
```
etienne-lms, line 130: `Cannot perform final on oneshot processing`

ta/pkcs11/src/pkcs11_token.h
```diff
@@ -128,14 +128,21 @@ enum pkcs11_proc_state {
 	PKCS11_SESSION_BUSY,
 };
 
+enum pkcs11_proc_op {
+	PKCS11_OP_INIT = 0,
+	PKCS11_OP_UPDATE,
+	PKCS11_OP_FINAL,
+	PKCS11_OP_FINAL_ONESHOT,
```
etienne-lms, line 135: maybe rename `PKCS11_OP_FINAL_ONESHOT` to `PKCS11_OP_ONESHOT`.

ta/pkcs11/src/pkcs11_token.h
```diff
@@ -128,14 +128,21 @@ enum pkcs11_proc_state {
...
+
 /*
  * Context of the active processing in the session
  *
  * @state - ongoing active processing function or ready state
  * @mecha_type - mechanism type of the active processing
  * @always_authen - true if user need to login before each use
  * @relogged - true once client logged since last operation update
- * @updated - true once an active operation is updated
+ * @op_state - last active operation state - update, final or one-shot
```
etienne-lms, line 145: I would replace `state` with `step` to prevent confusion (`PKCS11_FUNC_STEP_*`).
Operation state would rather refers the the state of the cyrpto context as per pkcs11 `C_{Get|Set}OperationState()`. What do you think?

etienne-lms (commented):
This P-R looks legitimate to me. I agree the issue must be addressed. There are mechanisms where implicitly a call to Final step to query output buffer size cannot be followed by a call to an Update step. I agree this to be addressed in a generic way as this P-R does.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -87,8 +87,10 @@ check_mechanism_against_processing(struct pkcs11_session *session,
 		    !session->processing->relogged)
 			return PKCS11_CKR_USER_NOT_LOGGED_IN;
 
-		if (session->processing->updated) {
-			EMSG("Cannot perform one-shot on updated processing");
+		if (session->processing->op_state == PKCS11_OP_UPDATE ||
+		    session->processing->op_state == PKCS11_OP_FINAL) {
+			EMSG("Cannot perform one-shot on updated or finalized \
```
etienne-lms, line 92: prefer shortened trace, i.e. `Cannot perform one-shot on active processing`.

etienne-lms:
> Out of interest what is the mecha where the problem would be?

The mechnism was aes-gsm decryption. Actaully AE stuff, where when we query the output size, we must complete the AE processing so that output size is known. Since we terminated the AE processing, it is hard to fill-in more input data afterward.

ta/pkcs11/src/pkcs11_token.h
ruchi393: Good Idea. I will do that.

ta/pkcs11/src/pkcs11_attributes.c
ruchi393: Ok

ta/pkcs11/src/pkcs11_token.h
ruchi393: I think what you say makes sense. I will change this to step.

ruchi393: Thinking more on it , we don't require the new enum pkcs11_proc_op at all. We can just save the last active operation and check on that. I will make the change and submit the PR fo review.

ta/pkcs11/src/pkcs11_token.h
```diff
@@ -147,7 +147,7 @@ struct active_processing {
 	uint32_t mecha_type;
 	bool always_authen;
 	bool relogged;
-	bool updated;
+	enum processing_step op_step;
```
etienne-lms, line 150: this field should be explicitly initialized to `PKCS11_FUNC_STEP_INIT` from `set_processing_state()` where `struct active_processing` is allocated. It happens that the struct is zero-filled and that `PKCS11_FUNC_STEP_INIT` equals 0, but I would prefer an explicit init:
```patch
 	proc = TEE_Malloc(sizeof(*proc), TEE_MALLOC_FILL_ZERO);
 	if (!proc)
 		return PKCS11_CKR_DEVICE_MEMORY;
 
 	/* Boolean are default to false and pointers to NULL */
 	proc->state = state;
+	proc->step = PKCS11_FUNC_STEP_INIT;
 	proc->tee_op_handle = TEE_HANDLE_NULL;
```

Also, I would maybe rename the field to `proc_step` or simply `step` and move it right below field `state` at line 146.

ta/pkcs11/src/processing.c
```diff
@@ -759,11 +759,16 @@ enum pkcs11_rc entry_processing_step(struct pkcs11_client *client,
 
 	if (rc == PKCS11_CKR_OK && (step == PKCS11_FUNC_STEP_UPDATE ||
 				    step == PKCS11_FUNC_STEP_UPDATE_KEY)) {
-		session->processing->updated = true;
+		session->processing->op_step = PKCS11_FUNC_STEP_UPDATE;
 		DMSG("PKCS11 session%"PRIu32": processing %s %s",
 		     session->handle, id2str_proc(mecha_type),
 		     id2str_function(function));
 	}
+	if (rc == PKCS11_CKR_BUFFER_TOO_SMALL && step == PKCS11_FUNC_STEP_ONESHOT)
```
etienne-lms, line 767: I suggest an empty line above since there are empty line in between and below the added code.

etienne-lms (commented):
> Thinking more on it , we don't require the new enum pkcs11_proc_op at all. We can just save the last active operation and check on that. I will make the change and submit the PR fo review.

I agree with you. The implementation looks better.

ta/pkcs11/src/pkcs11_token.h
ruchi393: Done, changed to step and initialised the variable

ta/pkcs11/src/processing.c
ruchi393: Ok

ruchi393:
All comments addressed

etienne-lms:

ruchi393:
Thanks @etienne-lms . Tags applied

PR #4780, opened 2021-08-04
plat-totalcompute: Introduce TC1 platform

uarif1:
The different platform parameters, such as GIC, UART, etc
are similar between TC0 and TC1. tc0_spmc_pm.c is also renamed to
tc_spmc_pm.c as its common to both platforms.

Signed-off-by: Usama Arif <usama.arif@arm.com>
Change-Id: I8263a37847266da0d0c0f1b112271ba701522131


core/arch/arm/plat-totalcompute/conf.mk
```diff
@@ -1,6 +1,4 @@
-PLATFORM_FLAVOR ?= tc0
-
-ifeq ($(PLATFORM_FLAVOR),tc0)
+ifneq ($(filter ${PLATFORM_FLAVOR}, tc0 tc1),)
```
jforissier, line 1: Minor: other platforms have the arguments swapped, i.e., the empty one first. I suggest doing the same for consistency. Also, better avoid spaces where not strictly necessary (`, tc0`). So:
```make
ifneq (,$(filter ${PLATFORM_FLAVOR},tc0 tc1)
```

core/arch/arm/plat-totalcompute/conf.mk
```diff
@@ -26,7 +24,7 @@ platform-cflags-debug-info = -gdwarf-2
 platform-aflags-debug-info = -gdwarf-2
 endif
 
-ifeq ($(PLATFORM_FLAVOR),tc0)
+ifneq ($(filter ${PLATFORM_FLAVOR}, tc0 tc1),)
```
jforissier, line 27: Same here

jforissier (commented):
Hi @uarif1,

Thanks for the patch. Two very minor comments below, then:

core/arch/arm/plat-totalcompute/conf.mk
uarif1: Thanks for the review! Updated in latest patch.

jforissier: You're welcome. Please don't forget to add my review tag next to you S-o-b.

uarif1: ah yes, sorry about that. Added in latest commit message.

PR #4785, opened 2021-08-06
core: svc: fix ta_time_prot_lvl always 100 problem

lubing521:
ta_time_prot_lvl value must be defined according
to the platform implementation,it may be 100 or 1000
that According to GP standards.

Signed-off-by: Lu Bing <lubing@eswin.com>
Tested-by: Lu Bing <lubing@eswin.com>

### DIFF
--- a/core/tee/tee_svc.c
+++ b/core/tee/tee_svc.c
@@ -70,7 +70,11 @@ static const char descr[] = TO_STR(CFG_TEE_IMPL_DESCR);
  * The real-time clock MUST be out of reach of software attacks
  * from the REE.
  */
+#ifdef CFG_SECURE_TIME_SOURCE_REE
 static const uint32_t ta_time_prot_lvl = 100;
+#elif CFG_SECURE_TIME_SOURCE_CNTPCT
+static const uint32_t ta_time_prot_lvl = 1000;
+#endif
 
 /* Elliptic Curve Cryptographic support */
 #ifdef CFG_CRYPTO_ECC


core/tee/tee_svc.c
```diff
@@ -70,7 +70,11 @@ static const char descr[] = TO_STR(CFG_TEE_IMPL_DESCR);
  * The real-time clock MUST be out of reach of software attacks
  * from the REE.
  */
+#ifdef CFG_SECURE_TIME_SOURCE_REE
 static const uint32_t ta_time_prot_lvl = 100;
+#elif CFG_SECURE_TIME_SOURCE_CNTPCT
```
etienne-lms, line 75: `CFG_SECURE_TIME_SOURCE_CNTPCT=y` is not sufficient here as the time reference is lost across reboots.

jforissier, line 75: No. This condition is not enough to fulfill the GP requirement, just read the comment above. CNTPCT (when available) is the "system clock" as in `TEE_GetSystemTime()`. Here it is the TA persistent real-time clock we re talking about (`gpd.tee.TAPersistentTime.protectionLevel`, not `gpd.tee.systemTime.protectionLevel`).
To support a level of 1000 here, we need both a real-time clock controlled by the TEE not the REE (which is not available in upstream OP-TEE), as well as TEE trusted storage (not clear what requirement this puts on storage; maybe the storage protection level can be considered by the TA if need be).

lubing521: thanks,Currently OpTEE-OS is not implemented to meet the GP 1000 level,so let it alway 100 level?

lubing521: thanks,Currently OpTEE-OS is not implemented to meet the GP 1000 level,so let it alway 100 level?

jforissier: Yes. If on your side you have platform-specific code for a real-time clock controlled by the TEE, then you can probably set it to 1000. But it cannot be 1000 with upstream code only.

lubing521: thanks,so this patch not suitable under conditon CFG_SECURE_TIME_SOURCE_CNTPCT .

lubing521:
this patch not suitable under conditon CFG_SECURE_TIME_SOURCE_CNTPCT .

PR #4792, opened 2021-08-10
FF-A add memory sharing for SPs (improved)

jellesels-arm:
FFA_MEM_SHARE currently only supports sharing memory from the Normal World to the OP-TEE endpoint. This PR add support to share a memory region from the Normal World with SPs or from a SP to another SP.

FFA_MEM_SHARE: is used to share a memory region with 1 or mote endpoints
FFA_MEM_RETRIEVE_REQ: is used by a SP to map a memory share in its memory map and to retrieve the VA address .
FFA_MEM_RELINQUISH: is used by a SP to relinquish access to a memory share, the memory region is unmapped form the SP memory map.
FFA_MEM_RECLAIM: is used by  the original owner of the memory region to reclaim the memory region. Afterwards none of the receiving SPs can retrieve the memory share anymore.

The following features are not supported and will be added later:
-Memory region attributes usage are not supported. Currently all shares have the cache as TEE_MATTR_CACHE_CACHED
-The memory management interfaces only support the use of the rxtx buffers when being called from a SP. The address parameter is not supported.

Manifest file that can be used to tests:
repo init -u https://review.trustedfirmware.org/OP-TEE/manifest -m psa-sp-fvp.xml -b refs/changes/99/10999/1
Changes made in the manifest:
Build: Change the build to be able to automatically build and add SPs to the image
optee_os: This PR
Trusted-service: Add the trusted services repo, this is where the test SPs live.
optee_test: The host component of the SPMC test
linux_poc: Proof of concept user space driver that is used to send FF-A messages from user space

We added a new bunch of regression tests in the optee_test repo under regression_9. This will contain all FF-A specific tests. The current PR passes all the xtest 90 tests. 
The SPs, which are used to execute the corresponding tests can be found in the trusted_service repo under deployments/spm_test/opteesp.

We are planning to upstream the tests them self, but are currently using a proof of concept Linux driver which is needed to send FF-A commands from userspace. The tests them self need to be cleaned but can be used to demonstrated the PR.

The following patches are added but not part of the PR:
[4732](https://github.com/OP-TEE/optee_os/pull/4732) is only added to be able to build the SPs but is not a dependency of this patch.
[4789](https://github.com/OP-TEE/optee_os/pull/4789) is needed to query the SP ids.


jenswi-linaro:
It seem that git.gitlab.arm.com is inaccessible.

jellesels-arm:
> It seem that git.gitlab.arm.com is inaccessible.

It seems that they changed the permission on gitlab. We are figuring out a solution for this

jellesels-arm:
> These TA_VASPACE -> TS_VASPACE changes are not related to memory sharing, please add them in a separate commit. Might as well do it in a separate pull request too to quickly get it out of the way.

Done [4801](https://github.com/OP-TEE/optee_os/pull/4801). I will rebase on top of master when the PR is merged.

core/include/mm/mobj_ffa.h
```diff
@@ -0,0 +1,75 @@
...
+#include <mm/fobj.h>
+#include <mm/mobj.h>
+#include <string_ext.h>
+#include <sys/queue.h>
+#include <tee_api_types.h>
+#include <types_ext.h>
+
+#include <optee_msg.h>
+
+SLIST_HEAD(sp_shared_mem_head_t, sp_mem_access_descr);
```
jenswi-linaro, line 23: This isn't needed without SPs enabled. We should have the mindset that if SPs aren't enabled it shouldn't have an impact either.

core/include/mm/mobj_ffa.h
```diff
@@ -0,0 +1,75 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
jenswi-linaro, line 1: Please create this file in a separate commit so it's easier to see what's added specifically for SPs.

core/arch/arm/mm/mobj_ffa.c
```diff
@@ -452,7 +499,7 @@ static size_t ffa_get_phys_offs(struct mobj *mobj,
 	return to_mobj_ffa(mobj)->page_offset;
 }
 
-static void *ffa_get_va(struct mobj *mobj, size_t offset)
+void *mobj_ffa_get_va(struct mobj *mobj, size_t offset)
```
jenswi-linaro, line 502: Why is this change needed?

core/arch/arm/mm/mobj_ffa.c
```diff
@@ -413,8 +460,8 @@ struct mobj *mobj_ffa_get_by_cookie(uint64_t cookie,
 	return &mf->mobj;
 }
 
-static TEE_Result ffa_get_pa(struct mobj *mobj, size_t offset,
-			     size_t granule, paddr_t *pa)
+TEE_Result ffa_get_pa(struct mobj *mobj, size_t offset,
```
jenswi-linaro, line 463: Why is this change needed, can't `mobj_get_pa()` be called instead?

jenswi-linaro (commented):
Some early comments on the commit "core: FF-A: ADD FFA_MEM_SHARE for SPs".

jellesels-arm:
> > It seem that git.gitlab.arm.com is inaccessible.
> 
> It seems that they changed the permission on gitlab. We are figuring out a solution for this

It should be again accessible.

PR #4793, opened 2021-08-11
core: kernel: interrupt: add type and prio for interrupts 

clementleger:
When describing a device in the device tree, it is sometimes necessary
to parse the interrupts properties and propagates them until adding
the interrupt. For instance some interrupt-cells allows to describe
priority and type of interrupt:

  interrupts = <67 IRQ_TYPE_LEVEL_HIGH 2>;

With existing support, only the interrupt number is returned by
`dt_get_irq()`. This patch adds type and prio parameter which are passed to
`dt_get_irq_type_prio()` and `itr_add_type_prio()`. This allows interrupt
drivers to fill this from devicetree in `dt_get_irq()` but also use these
information in the `add()` callback. Additionally, it allows to specify these
flags manually when not using devicetree.

These parameters can then be used by the interrupt controller driver to
setup the irq line correctly.

### DIFF
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -70,7 +70,8 @@
 #define GICC_IAR_CPU_ID_MASK	0x7
 #define GICC_IAR_CPU_ID_SHIFT	10
 
-static void gic_op_add(struct itr_chip *chip, size_t it, uint32_t flags);
+static void gic_op_add(struct itr_chip *chip, size_t it, uint32_t type,
+		       uint32_t prio);
 static void gic_op_enable(struct itr_chip *chip, size_t it);
 static void gic_op_disable(struct itr_chip *chip, size_t it);
 static void gic_op_raise_pi(struct itr_chip *chip, size_t it);
@@ -208,9 +209,16 @@ void gic_init(struct gic_data *gd, vaddr_t gicc_base __maybe_unused,
 #endif
 }
 
-static int gic_dt_get_irq(const uint32_t *properties, int count)
+static int gic_dt_get_irq(const uint32_t *properties, int count, uint32_t *type,
+			  uint32_t *prio)
 {
 	int it_num = DT_INFO_INVALID_INTERRUPT;
+
+	if (type)
+		*type = IRQ_TYPE_NONE;
+
+	if (prio)
+		*prio = 0;
 
 	if (!properties || count < 2)
 		return DT_INFO_INVALID_INTERRUPT;
@@ -428,7 +436,8 @@ void gic_it_handle(struct gic_data *gd)
 }
 
 static void gic_op_add(struct itr_chip *chip, size_t it,
-		       uint32_t flags __unused)
+		       uint32_t type __unused,
+		       uint32_t prio __unused)
 {
 	struct gic_data *gd = container_of(chip, struct gic_data, chip);
 
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -208,11 +208,11 @@ void gic_init(struct gic_data *gd, vaddr_t gicc_base __maybe_unused,
 #endif
 }
 
-static int gic_dt_get_irq(const uint32_t *properties, int len)
+static int gic_dt_get_irq(const uint32_t *properties, int count)
 {
 	int it_num = DT_INFO_INVALID_INTERRUPT;
 
-	if (!properties || len < 2)
+	if (!properties || count < 2)
 		return DT_INFO_INVALID_INTERRUPT;
 
 	it_num = fdt32_to_cpu(properties[1]);
--- a/core/include/kernel/dt.h
+++ b/core/include/kernel/dt.h
@@ -7,6 +7,7 @@
 #define KERNEL_DT_H
 
 #include <compiler.h>
+#include <kernel/interrupt.h>
 #include <kernel/panic.h>
 #include <stdint.h>
 #include <types_ext.h>
@@ -33,6 +34,10 @@
  * @reset: Device reset identifier (positive value) or DT_INFO_INVALID_CLOCK
  * @interrupt: Device interrupt identifier (positive value) or
  * DT_INFO_INVALID_INTERRUPT
+ * @type: IRQ_TYPE_* value parsed from interrupts properties or IRQ_TYPE_NONE if
+ * not present
+ * @prio: interrupt priority parsed from interrupts properties or 0 if not
+ * present
  */
 struct dt_node_info {
 	unsigned int status;
@@ -40,6 +45,8 @@ struct dt_node_info {
 	int clock;
 	int reset;
 	int interrupt;
+	uint32_t type;
+	uint32_t prio;
 };
 
 #if defined(CFG_DT)
--- a/core/include/kernel/interrupt.h
+++ b/core/include/kernel/interrupt.h
@@ -14,7 +14,7 @@
 
 struct itr_chip {
 	const struct itr_ops *ops;
-	int (*dt_get_irq)(const uint32_t *properties, int len);
+	int (*dt_get_irq)(const uint32_t *properties, int count);
 };
 
 struct itr_ops {
--- a/core/include/kernel/interrupt.h
+++ b/core/include/kernel/interrupt.h
@@ -5,6 +5,7 @@
 #ifndef __KERNEL_INTERRUPT_H
 #define __KERNEL_INTERRUPT_H
 
+#include <dt-bindings/interrupt-controller/irq.h>
 #include <types_ext.h>
 #include <sys/queue.h>
 #include <util.h>
@@ -14,11 +15,22 @@
 
 struct itr_chip {
 	const struct itr_ops *ops;
-	int (*dt_get_irq)(const uint32_t *properties, int count);
+	/*
+	 * dt_get_irq - parse a device tree interrupt property
+	 *
+	 * @properties raw interrupt property from device tree
+	 * @count number of elements in @properties
+	 * @type If not NULL, output interrupt type (IRQ_TYPE_* defines)
+	 * or IRQ_TYPE_NONE if unknown
+	 * @prio If not NULL, output interrupt priority value or 0 if unknown
+	 */
+	int (*dt_get_irq)(const uint32_t *properties, int count, uint32_t *type,
+			  uint32_t *prio);
 };
 
 struct itr_ops {
-	void (*add)(struct itr_chip *chip, size_t it, uint32_t flags);
+	void (*add)(struct itr_chip *chip, size_t it, uint32_t type,
+		    uint32_t prio);
 	void (*enable)(struct itr_chip *chip, size_t it);
 	void (*disable)(struct itr_chip *chip, size_t it);
 	void (*raise_pi)(struct itr_chip *chip, size_t it);
@@ -50,21 +62,37 @@ void itr_handle(size_t it);
 
 #ifdef CFG_DT
 /*
- * Get the DT interrupt property at @node. In the DT an interrupt
+ * Get the DT interrupt property at @node. In the DT an interrupt property can
+ * specify additional information which can be retrieved with @type and @prio.
  *
  * @fdt reference to the Device Tree
- * @node is the node offset to read
+ * @node is the node offset to read the interrupt property from
+ * @type interrupt type (IRQ_TYPE_* defines) if specified by interrupt property
+ * or IRQ_TYPE_NONE if not. Can be NULL if not needed
+ * @prio interrupt priority if specified by interrupt property or 0 if not. Can
+ * be NULL if not needed
  *
  * Returns the interrupt number if value >= 0
  * otherwise DT_INFO_INVALID_INTERRUPT
  */
-int dt_get_irq(const void *fdt, int node);
+int dt_get_irq_type_prio(const void *fdt, int node, uint32_t *type,
+			 uint32_t *prio);
+
+/*
+ * Get the DT interrupt property at @node
+ */
+static inline int dt_get_irq(const void *fdt, int node)
+{
+	return dt_get_irq_type_prio(fdt, node, NULL, NULL);
+}
 #endif
 
-struct itr_handler *itr_alloc_add(size_t it, itr_handler_t handler,
-				  uint32_t flags, void *data);
+struct itr_handler *itr_alloc_add_type_prio(size_t it, itr_handler_t handler,
+					    uint32_t flags, void *data,
+					    uint32_t type, uint32_t prio);
 void itr_free(struct itr_handler *hdl);
-void itr_add(struct itr_handler *handler);
+void itr_add_type_prio(struct itr_handler *handler, uint32_t type,
+		       uint32_t prio);
 void itr_enable(size_t it);
 void itr_disable(size_t it);
 /* raise the Peripheral Interrupt corresponding to the interrupt ID */
@@ -86,5 +114,18 @@ void itr_set_affinity(size_t it, uint8_t cpu_mask);
  * expects to receive secure interrupts should override this function.
  */
 void itr_core_handler(void);
+
+static inline void itr_add(struct itr_handler *handler)
+{
+	itr_add_type_prio(handler, IRQ_TYPE_NONE, 0);
+}
+
+static inline struct itr_handler *itr_alloc_add(size_t it,
+						itr_handler_t handler,
+						uint32_t flags, void *data)
+{
+	return itr_alloc_add_type_prio(it, handler, flags, data, IRQ_TYPE_NONE,
+				       0);
+}
 
 #endif /*__KERNEL_INTERRUPT_H*/
--- a/core/kernel/dt.c
+++ b/core/kernel/dt.c
@@ -282,7 +282,8 @@ void _fdt_fill_device_info(const void *fdt, struct dt_node_info *info, int offs)
 		dinfo.reset = (int)fdt32_to_cpu(*cuint);
 	}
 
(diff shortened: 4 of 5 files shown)


core/kernel/interrupt.c
```diff
@@ -76,6 +77,10 @@ struct itr_handler *itr_alloc_add(size_t it, itr_handler_t handler,
 		hdl->handler = handler;
 		hdl->flags = flags;
 		hdl->data = data;
+		if (spec)
+			memcpy(&hdl->spec, spec, sizeof(hdl->spec));
+		else
+			memset(&hdl->spec, 0, sizeof(hdl->spec));
```
etienne-lms, line 83: `memset()` not needed if allocated from `calloc()` above.

core/kernel/dt.c
```diff
@@ -265,6 +265,7 @@ void _fdt_fill_device_info(void *fdt, struct dt_node_info *info, int offs)
 		.clock = DT_INFO_INVALID_CLOCK,
 		.reset = DT_INFO_INVALID_RESET,
 		.interrupt = DT_INFO_INVALID_INTERRUPT,
+		.spec = {{0}, 0},
```
etienne-lms, line 268: not needed if all zeroes.

core/include/kernel/interrupt.h
```diff
@@ -12,13 +12,26 @@
...
 #define ITRF_SHARED			BIT(1)
 
+#define IRQ_MAX_SPEC		4
+
+/* struct itr_spec - Interrupt specifier
+ * @irq_spec: Specifier for interrupt
+ * @num_spec: number of specifiers in the irq_spec member
+ */
+struct itr_spec {
+	uint32_t irq_spec[IRQ_MAX_SPEC];
```
etienne-lms, line 22: This makes `struct itr_handler` to grow even for platforms when useless (not that there are many that care). Can the struct be allocated to the expected size when needed?

I don't really like such opaque an blob list. Things in _interrupt.h_ will be generic. It looks like this is expected to store a raw 32bit data series from the DT and pass to back-end driver. If data are driver specific then the driver should allocate a private data area in the interrupt instance and store there any driver specific information.

clementleger: Agreed, the idea was to be able to pass interrupt description (often described in the devicetree) to the driver. 
Currently, dt_get_irq returns only an int property, and even if not used, how do you pass additional information to the interrupt driver at itr_add call ? flags members seems a bit limiting.

clementleger: To be more clear, I need to pass priority and interrupt type (LEVEL, EDGE, HIGH, LOW) to the interrupt controller. This information is described in the device tree but might be provided by non dt code. These data are usually described using the interrupt cells (which are uint32_t).

One possibility is probably to drop this from struct itr_spec and pass it to itr_add and others. This way it won't take any space at all (except on the stack when declaring it). But it will be decoupled from the itr_handler which seems a bit odd.

etienne-lms: > But it will be decoupled from the itr_handler which seems a bit odd.

If `itr_add()` blindly passes it as an arg to the controller function handler (`struct itr_ops::add`), it looks fine to me. These are controller specific config data, so fine if opaque to `struct itr_handler`, It could even replace the `flags` argument in `itr_add()` that serves an equivalent purpose and is currently unused by the sole interrupt controller supported in optee.

clementleger: For dt_fill_device_info, the controller will add to allocate data because this part is generic and call dt_get_irq.
This means that dt_get_irq  will return an allocated struct by the controller. This is probably adding unnecessary allocations, and IMHO, the interrupt specifier that are used in the dt are simple (up to 4 uint32_t) and seems to fulfill their purpose to describe interrupts for almost all interrupt controllers.

etienne-lms: > This is probably adding unnecessary allocations 

These data are used only during `itr_add()` for configuration, right? I think they could be de-referenced/released once `itr_add()` completes.

> IMHO, the interrupt specifier that are used in the dt are simple (up to 4 uint32_t) and seems to fulfill their purpose to describe interrupts for almost all interrupt controllers.

The interrupt controller driver could read itself into the DT to get/process expected DT property cells. Maybe the DT parsing should not be done in generic `interrupt.c`.  Maybe over killing :).
Note the optee implements a single interrupt controller. If we need to support more, possibly cascade controllers, _interrupt.c|.h_ current API will not match.

As for `dt_fill_device_info()`, it is a convenient helper function and could remain simple IMO and always call `dt_get_irq()` with a NULL reference for extra interrupt configs.

I think we could remove field `struct itr_handler::flags` and replace existing argument `flags` of `itr_add()` with this `struct itr_dt_spec` reference, which could possibly be NULL.

clementleger: > > This is probably adding unnecessary allocations
> 
> These data are used only during `itr_add()` for configuration, right? I think they could be de-referenced/released once `itr_add()` completes.

> 
> > IMHO, the interrupt specifier that are used in the dt are simple (up to 4 uint32_t) and seems to fulfill their purpose to describe interrupts for almost all interrupt controllers.
> 
> The interrupt controller driver could read itself into the DT to get/process expected DT property cells. Maybe the DT parsing should not be done in generic `interrupt.c`. Maybe over killing :).

As I said, this might be done without devicetree, it is still necessary to passe information to the interrupt controller.

> Note the optee implements a single interrupt controller. If we need to support more, possibly cascade controllers, _interrupt.c|.h_ current API will not match.

The controller I've added does not requires cascaded interrupts but only needs more attributes for irq configuration so the current API is ok.

> 
> As for `dt_fill_device_info()`, it is a convenient helper function and could remain simple IMO and always call `dt_get_irq()` with a NULL reference for extra interrupt configs.

But then, device tree interrupts property won't be parsed and won't be retrieval by the drivers. This seems a bit deceiving.

> 
> I think we could remove field `struct itr_handler::flags` and replace existing argument `flags` of `itr_add()` with this `struct itr_dt_spec` reference, which could possibly be NULL.

I guess you talk about `itr_ops::add` since itr_add only takes a handler as parameter.

etienne-lms: > > As for `dt_fill_device_info(),` it is a convenient helper function and could remain simple IMO and always call `dt_get_irq()` with a NULL reference for extra interrupt configs.
> 
> But then, device tree interrupts property won't be parsed and won't be retrieval by the drivers. This seems a bit deceiving.

Sorry, i meant "(...) remain simple IMO and **could** always call `dt_get_irq()` with a NULL (...)".
I have no objection for `dt_fill_device_info()` to povide more help.


> > I think we could remove field `struct itr_handler::flags` and replace existing argument `flags` of `itr_add()` with this `struct itr_dt_spec` reference, which could possibly be NULL.
> 
> I guess you talk about `itr_ops::add` since itr_add only takes a handler as parameter.

Right. Add this `_spec` reference as argument to `itr_add()` and replace `itr_ops::add()` argument `flags` with the `_spec` reference.

clementleger: Ok, actually after looknig a bit more carefully at it, `flags` is used for more generic attributes and is used by the generic interrupt code:
```
			assert((hdl->flags & ITRF_SHARED) &&
			       (h->flags & ITRF_SHARED));
```
I'm not sure it would be a good idea to remove it. The struct `itr_spec` is for hardware interrupt line description and is meant for the interrupt controller driver.

etienne-lms: Me to blame, you're right. `flags` is internal. It already provides the edge/level config but is generic in `itr_` framework.

jenswi-linaro:
There's an overlap between `struct itr_spec` and `ITRF_TRIGGER_LEVEL` flag in `struct itr_handler::flags`. Given that we currently don't do anything with the `ITRF_TRIGGER_LEVEL` flag we could remove that and let that part be handled only via `struct itr_spec`.

core/include/kernel/interrupt.h
clementleger: Ok, so I will pass NULL when calling `dt_get_irq` in `dt_fill_device_info`. If needed, it could be added later.
I will also remove the flag argument for `.add` callback since flags should remain generic for the itr framework and keep the itr_spec for this callback which will contain the expected interrupt description (same as in the device tree) for the interrupt controller driver .

clementleger:
> There's an overlap between `struct itr_spec` and `ITRF_TRIGGER_LEVEL` flag in `struct itr_handler::flags`. Given that we currently don't do anything with the `ITRF_TRIGGER_LEVEL` flag we could remove that and let that part be handled only via `struct itr_spec`.

Indeed, this could be used to pass this flag. Maybe this should use the define that are used in the device tree, IRQ_TYPE_LEVEL_HIGH and other for instance instead of ITRF_TRIGGER_LEVEL, are you ok with that ?
BTW,  ITRF_TRIGGER_LEVEL seems unused in the GIC driver.
Globally, are you ok with the addition of this `struct itr_spec` or do you think it can be limitating to other usecases ?

jenswi-linaro:
> > There's an overlap between `struct itr_spec` and `ITRF_TRIGGER_LEVEL` flag in `struct itr_handler::flags`. Given that we currently don't do anything with the `ITRF_TRIGGER_LEVEL` flag we could remove that and let that part be handled only via `struct itr_spec`.
> 
> Indeed, this could be used to pass this flag. Maybe this should use the define that are used in the device tree, IRQ_TYPE_LEVEL_HIGH and other for instance instead of ITRF_TRIGGER_LEVEL, are you ok with that ?
> BTW, ITRF_TRIGGER_LEVEL seems unused in the GIC driver.
> Globally, are you ok with the addition of this `struct itr_spec` or do you think it can be limitating to other usecases ?
> 
> EDIT: However, I'm not sure if passing the `uint32_t properties` directly to itr_add would be better.

Perhaps we're trying to be too generic or flexible. Would it work redefining `struct itr_handler` as:
``` C
struct itr_handler {
        uint32_t it;
        uint8_t flags;
        uint8_t type; /* IRQ_TYPE_* from device-tree */
        itr_handler_t handler; 
        void *data;
        SLIST_ENTRY(itr_handler) link;
};
```
`itr_add()` and `itr_alloc_add()` would be changed as needed to take a `type` parameter.

clementleger:
In my case I need both the type and the interrupt priority so I would also need a priority argument but this is purely related to my controller so it seems strange to have that as an argument. I guess keeping it at least "a bit" more generic is needed. BTW, I don't think this is needed in the itr_handler struct (I removed it from my original PR) since this is potentially only once use at configure time (at least in my case).

jenswi-linaro:
Being able to set the priority of an interrupt seems reasonable.
We could add a prio argument too to `itr_add()`. In case we don't want to force everyone to pass a dummy prio we could have a `itr_add_prio()` instead where `itr_add()` would just be a wrapper function supplying a dummy or disabled value instead of a real `prio` argument. A bit like with `vm_map()` and `vm_map_pad()`.

clementleger:
> Being able to set the priority of an interrupt seems reasonable.
> We could add a prio argument too to `itr_add()`. In case we don't want to force everyone to pass a dummy prio we could have a `itr_add_prio()` instead where `itr_add()` would just be a wrapper function supplying a dummy or disabled value instead of a real `prio` argument. A bit like with `vm_map()` and `vm_map_pad()`.

Ok, sound good to me, please note that I will have to add these arguments to `dt_get_irq` also.

core/include/kernel/dt.h
```diff
@@ -33,13 +34,18 @@
  * @reset: Device reset identifier (positive value) or DT_INFO_INVALID_CLOCK
  * @interrupt: Device interrupt identifier (positive value) or
  * DT_INFO_INVALID_INTERRUPT
+ * @irq_type: IRQ_TYPE_* value parsed from interrupts properties or
```
etienne-lms, line 37: prefer `type` over `irq_type`. `irq` can be confusing on Arm (optee native (secure) interrupts can be IRQs or FIQs. See foreign/native interrutp).

core/include/kernel/dt.h
```diff
@@ -33,13 +34,18 @@
...
+ * IRQ_TYPE_NONE if not present
+ * @irq_prio: irq priority parsed from interrupts properties or 0 if not present
```
etienne-lms, line 39: ditto IMO: s/irq_prio/prio/g

core/include/kernel/interrupt.h
```diff
@@ -14,11 +15,23 @@
...
-	int (*dt_get_irq)(const uint32_t *properties, int len);
+	/*
+	 * dt_get_irq - parse a device tree interrupt property
+	 *
+	 * @properties raw interrupt property from device tree
+	 * @len number of properties
+	 * @type irq type (IRQ_TYPE_* defines) if specified by interrupt
+	 * property or IRQ_TYPE_NONE if not. Pointer is never NULL.
+	 * @prio irq priority if specified by interrupt property. Pointer is
+	 * never NULL.
```
etienne-lms, line 26:
```
	 * @prio Output interrupt priority or 0.
```
I think the callback should handle NULL `type` and `prio` pointers.
(` * @prio Output interrupt priority value (0 if unknown) if not NULL.`)

clementleger: Ok, I wanted to do this in the upper layer to avoid having each implementation to do it but if you are ok with that, the code that uses intermediate local variable can even be removed. I will modify that

clementleger: Ok I removed this and allows passing NULL pointers.

core/include/kernel/dt.h
clementleger: Done

core/include/kernel/dt.h
clementleger: Ok, renamed it and changed irq -> interrupt to be more generic.

core/kernel/interrupt.c
```diff
@@ -42,7 +44,17 @@ int dt_get_irq(const void *fdt, int node)
 	if (!prop)
 		return it_num;
 
-	return itr_chip->dt_get_irq(prop, len);
+	if (!type)
+		type = &ltype;
+	else
+		*type = IRQ_TYPE_NONE;
```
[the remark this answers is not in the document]
clementleger: Since I added the possibility of passing NULL pointers in dt_get_irq callback, I removed the intermediate variables and kept only the initialization part (Or do you want it to be removed also ?).

core/include/kernel/interrupt.h
```diff
@@ -87,4 +113,17 @@ void itr_set_affinity(size_t it, uint8_t cpu_mask);
  */
 void itr_core_handler(void);
 
+static inline void itr_add(struct itr_handler *handler, uint32_t type)
```
jenswi-linaro, line 116: We could pass the default type `IRQ_TYPE_NONE` also with this function. This would avoid many changes in this patch.
Opinions?

core/kernel/dt.c
clementleger: Removed

core/include/kernel/interrupt.h
etienne-lms: I agree.
Ditto for `itr_alloc_add()`: could pass `IRQ_TYPE_NONE`.  One needing to allocate/add an interrupt with more parameters should call `itr_alloc_add_prio()` straight.

By the way, maybe rename `itr_add_prio()` to `itr_add_type_prio()`
and `itr_alloc_add_prio()` to `itr_alloc_add_type_prio()`.

core/kernel/interrupt.c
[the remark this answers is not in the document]
etienne-lms: No need to set a default value here IMO.

core/include/kernel/interrupt.h
clementleger: Ok, I'll do that. so in the end `itr_add` stay the same and `itr_add_type_prio` uses `type` and `prio` parameters.

core/drivers/gic.c
```diff
@@ -208,7 +209,8 @@ void gic_init(struct gic_data *gd, vaddr_t gicc_base __maybe_unused,
 #endif
 }
 
-static int gic_dt_get_irq(const uint32_t *properties, int len)
+static int gic_dt_get_irq(const uint32_t *properties, int len,
+			  uint32_t *type __unused, uint32_t *prio __unused)
```
jenswi-linaro, line 213: Would it make sense to set `type` and `prio` to `IRQ_TYPE_NONE` and `0` respectively?
The point would be to make sure that we don't use garbage. It may help simplifying other functions too if they can depend on these always being set on a successful return.

core/kernel/interrupt.c
```diff
@@ -42,7 +43,13 @@ int dt_get_irq(const void *fdt, int node)
 	if (!prop)
 		return it_num;
 
-	return itr_chip->dt_get_irq(prop, len);
+	if (type)
```
jenswi-linaro, line 46: Wouldn't need this if we could depend on `itr_chip->dt_get_irq()` setting `type` and `prio`.

core/drivers/gic.c
clementleger: Yes, that was the point of the init in the generic dt_get_irq_type_prio.

core/kernel/interrupt.c
clementleger: Moved to gic.

jenswi-linaro:

clementleger:
New version:
- Add missing parenthesis in commit message for functions

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with minor comments addressed or not.

PR #4801, opened 2021-08-18
core: rename TA_VASPACE to TS_VASPACE

jellesels-arm:
The TA_VASPACE memory will be used by both TAs and SPs.
Rename it to TS_VASPACE so it is clearer that it can be used by both.
Split off from [#4792](https://github.com/OP-TEE/optee_os/pull/4792)


jenswi-linaro:

etienne-lms:

jforissier:
Merged.

PR #3514, opened 2020-01-07
RFC:Do not merge: Runtime Secure Keys Framework Proposal for RSA

sahilnxp:
Commit for Runtime Secure Key framework proposal without adding
new object types.
For more information about the discussion about this please
check https://github.com/OP-TEE/optee_os/issues/3287
rsa.c and bignum.c files are placed in core/lib/ just for now
to give an ides.
These will be moved to more accurate places when RSA support
will be added as a part of CAAM support work already in progress.

Signed-off-by: Sahil Malhotra <sahil.malhotra@nxp.com>
https://github.com/OP-TEE/optee_os/issues/3287

PR #4807, opened 2021-08-20
[RFC #2] Runtime Secure Keys Framework Proposal 

clementfaure:
Hello,

Here is a RFC pull-request for the integration of the runtime secure key within OPTEE as well as its interface for the Crypto HW driver. This work is a follow-up of the RFC made by Sahil and Ruchika a year ago for the RSA runtime secure key proposal. The runtime secure key concept has been extended to all cryptographic algorithms. https://github.com/OP-TEE/optee_os/pull/3514

The notion of encrypted/protected key is called here "rsec" for Runtime Secure key (after Sahil and Ruchika in the first PR). Nothing is definitive, all naming can be changed later.
Please note that the code attached does not feature all changes required for the actual implementation yet.

# Definition of a Runtime Secure Key (or RSEC)  

A Runtime Secure Key is a key protected and handled by the Crypto driver. A RSEC key would have the following properties:

-   The RSEC key buffer never holds the plain text key.
-   The key could only be wrapped in one way : plain text → RSEC. The opposite operation cannot be implemented.
-   Only the crypto driver would be able to use the RSEC key for crypto operations.
-   The RSEC key buffer content is the same whatever the object type (persistent or transient).

All operations related to rsec key are handled by the Crypto driver. By delegating the rsec key handling to the Crypto driver, we ensure that the key is never exposed unprotected in memory. It's up to the Crypto driver to implement the key protection. The crypto driver must implement:

-   Key allocation
-   Key generation and encapsulation
-   Key usage for cryptographic operations

To implement the RSEC key concept, we plan on:

-   Using attributes and object of the TEE Crypto API to do the distinction between RSEC and non-RSEC keys.
-   Introducing a common key object for all algorithms.
-   Adding the RSEC key concept to the Crypto HW API interface so any Crypto Driver could implement its own runtime secure key.

We aim at making the RSEC key implementation as generic as possible so a wide variety of Crypto driver could implement the RSEC key. The RSEC key can be an encrypted key stored in memory but it could also be a key stored in an external secure memory. Because the RSEC key data buffer is only handled by the crypto driver, it leaves the crypto driver a lot of flexibility on how the RSEC key is implemented.

# The implementation

## TEE Crypto API RSEC implementation

If we want to leave the choice to use or not an RSEC key for crypto operations, we need a way to designate the use of a RSEC key in the TA through the TEE Crypto API. Appending the TEE Crypto API for a new encrypted key concept would take a lot of time, the best would be to count on the Crypto API flexibility and "Object Usage".

What we suggest is the add a new **TEE_USAGE_PLAIN_KEY** to designate a RSEC or non-RSEC key. We would use the **TEE_RestrictObjectUsage1()** function to enable this new key usage. This idea was suggested in the previous RFC. In the previous RFC, using the already existing attribute **TEE_USAGE_EXTRACTABLE** as been suggested but it might be confusing to designate a RSEC key.

> The TEE_USAGE* would be called set by the `TEE_RestrictObjectUsage1()` function in 3 different scenarios:
> 
> 1.  When the object is Uninitialized [Transient Uninitialized Object]
> 2.  When object is Initialized and Not-persistent [Transient Initialized Object]
> 3.  When object is Initialized and Persistent. [Persistent Object]
> 
> The case 1 would generate a black key or wrap a plain key into a black key.
> 
> The cases 2 and 3 would encapsulate a plain key into an encrypted key. Since it is possible to call TEE_RestrictObjectUsage1() after the object initialization, the function TEE_RestrictObjectUsage1() will have to call the crypto driver to encapsulate the key. The crypto driver might need to reallocate the key accordingly.
> 
> (Credit @Ruchika)

Since **TEE_RestrictObjectUsage1()** defines the RSEC key and can be invoked at different stages of the object lifecycle, the **TEE_RestrictObjectUsage1()** will need to do a bit more than setting object flags. This function needs to do the following operations:

-   **If TEE_RestrictObjectUsage1(~TEE_USAGE_PLAIN_KEY ) and NOT** **TEE_HANDLE_FLAG_INITIALIZED then ..**
    -   **TEE_GenerateKey()**: call the appropriate Crypto driver to generate the key.
-   **If TEE_RestrictObjectUsage1(~TEE_USAGE_PLAIN_KEY) AND** **TEE_HANDLE_FLAG_INITIALIZED**
    -   **TEE_RestrictObjectUsage1():** call Crypto driver to wrap the key.

Adding a new TEE_USAGE will have an impact on already existing TA using **TEE_RestrictObjectUsage1().** If a TA restricts an object usage without including the newly added **TEE_USAGE_PLAIN_KEY** in the bitfield, then the object will be set as RSEC and that might be not wanted. Same thing would happen if we reuse **TEE_USAGE_EXTRACTABLE.** A TA using this attribute with **TEE_RestrictObjectUsage1****()** might ends up with a RSEC. Using a **TEE_USAGE*** to define a RSEC key or not, will have an impact on retro-compatibility.

Another straight forward solution would be to have a compilation flag for the Crypto Driver like **CFG_CRYPTO_DRV_[CIPHER, MAC, RSA ...] _RSEC** that enables the use by default of RSEC keys. Keys are generated as RSEC and wrapped as RSEC when populated. This way, there is no need to implement the RSEC key use in the TEE Crypto API.

## Introduce a new key object

Introducing a new key object in the code seems to be mandatory for a proper implementation of the rsec key. This object is an extension of obj_secret. Until now, key were handled differently depending of the algorithm. For the acipher, we have the bignum format and for the cipher, we have the combo buffer + length to handle the key. A good idea would be to unify the key format and use the following for all crypto keys:

``` C
struct tee_cryp_obj_key {
        /* Size of the security key */
	size_t size;
        /* Space allocated to store the key in memory */
	size_t alloc_size;
        /* Key buffer */
	uint8_t *data;
        /* A key attribute to know if the key is rsec or not (and potentially more info) */
        bool/uint32_t status;

        /* Key operation vectors */
	struct key_ops *ops;
};
```

-   **size_t size** : this variable holds the key security size. 128, 192, 256 ... for the Cipher and 1024, 2048 or 192, 224, 521 for the Asymmetric crypto. The size would be in bits.
-   **size_t alloc_size**: this variable holds the memory space allocated by the key buffer. Some crypto driver might need more space than the key security size to store metadata, checksum, nonce ... The Black key encapsulation in the CAAM adds at least to 12 more bytes to the encrypted key (6 bytes for the nonce and 6 bytes for the ICV). Because of the ICV, only a valid black key can be loaded. The size would be in bytes.
-   **uint8_t \*data**: data buffer holds the encrypted key and/or metadata.
-   **bool status**: flag that says if the key should be treated as rsec or plain text. We could also use a 32 bits value if we want to store more key attributes which is not the case currently.
-   **struct key_ops \*ops**: key operation vectors implemented by the crypto driver and initialized at key allocation/re-allocation. See bellow.

This key object implementation would discard the bignum format for the OPTEE core and asymmetric keys. For RSA keypair, we would have the following for example:

``` C
struct rsa_keypair {
	struct tee_cryp_obj_key *e;	/* Public exponent */
	struct tee_cryp_obj_key *d;	/* Private exponent */
	struct tee_cryp_obj_key *n;	/* Modulus */

	/* Optional CRT parameters (all NULL if unused) */
	struct tee_cryp_obj_key *p;	    /* N = pq */
	struct tee_cryp_obj_key *q;
	struct tee_cryp_obj_key *qp;	/* 1/q mod p */
	struct tee_cryp_obj_key *dp;	/* d mod (p-1) */
	struct tee_cryp_obj_key *dq;	/* d mod (q-1) */
}
```

The key operation vectors are initialized by the crypto driver upon key allocation. These functions are needed since some special operations must be done by the crypto driver to wrap the key.
``` C
struct key_ops {
	TEE_Result (*wrap)(struct tee_cryp_obj_key *key);
}
```

Currently, there is only one key operation :

-   **wrap() :** the "wrap" operation is the key encapsulation from plain text to RSEC.

## Add the RSEC key support in the Crypto HW API

**Add new cipher operation vectors**

Currently, the cipher and acipher don't have the same operation vectors. Basically, the cipher lacks key allocation and the key generation the acipher currently has. Extending the Crypto HW API to handle this part would allow the Crypto driver to handle the protected key allocation, generation and usage. This way the key generation is extended to the HW Crypto API and not restricted to a software RNG.

Here is what the cipher interface would look like:
``` C
struct drvcrypt_cipher {
	/* Allocate context */
	TEE_Result (*alloc_ctx)(void **ctx, uint32_t algo);
	/* Free context */
	void (*free_ctx)(void *ctx);
	/* Initialize the cipher operation */
	TEE_Result (*init)(struct drvcrypt_cipher_init *dinit);
	/* Update the cipher operation */
	TEE_Result (*update)(struct drvcrypt_cipher_update *dupdate);
	/* Finalize the cipher operation */
	void (*final)(void *ctx);
	/* Copy cipher context */
	void (*copy_state)(void *dst_ctx, void *src_ctx);

    + /* Generate key */
    + TEE_Result (*generate_key)(struct key_object{});
    + /* Allocate key */
    + TEE_Result (*alloc_key)(struct key_object{});
};
```
The same thing must be done for the MAC also ...

# CAAM features

## Black key and Black blob definitions

CAAM's black key mechanism is intended for protection of user keys against bus snooping and memory dumping while the keys are being written to or read from memory external to the SoC. The black key mechanism automatically encapsulates and decapsulates cryptographic keys on-the-fly in an encrypted data structure called a black key. Before a value is copied from a key register to memory, CAAM automatically encrypts the key as a black key (encrypted key) using as the encryption key the current value in the JDKEKR or TDKEKR, modified via the appropriate TZ/SDID value. Thus, each security domain (and TrustZone SecureWorld) has its own private black keys, which cannot be decrypted by the user of a different security domain identifier. When CAAM is instructed to use a black key as an encryption key, CAAM automatically decrypts the black key and places it directly into a key register before using the decrypted value in the user-specified cryptographic operation. Please note that JDKEKR and TDKEKR values are randomly re-generated at boot. The life expectancy of a black key is a power-cycle (reset or suspend/resume). A black key generated during a power cycle n cannot be decrypted and used starting the power cycle n+1. To make a black key power-cycle proof, a black key must be stored as a Blob (see bellow).

CAAM's built-in blob protocol provides a method for protecting user-defined data across system power cycles. The data to be protected is encrypted so that it can be safely placed into non-volatile storage before the chip is powered down. Each time that the blob protocol is used to protect data, a different randomly generated key is used to encrypt the data. This random key is itself encrypted using a key encryption key and the resulting encrypted key is then stored along with the encrypted data. The key-encryption key is derived from the chip's master secret key so the key-encryption key can be recreated when the chip powers up again. The combination of encrypted key and encrypted data is called a blob.

## Technical CAAM RSEC key implementation

On NXP platforms, the RSEC key will be stored as a black blob and not as a black key. The reason is that JDKEK and TDKEK and KEK nonce are regenerated upon every suspend/resume and system reset. Because of this, a black key will be unusable after a simple suspend/resume. Storing the key as blob instead of a black might have an impact on CAAM performance but it is negligible.

# Credits
@lenormandfranck @sahilnxp @cneveux

### DIFF
--- a/core/crypto/crypto.c
+++ b/core/crypto/crypto.c
@@ -198,6 +198,58 @@ void crypto_cipher_final(void *ctx)
 	cipher_ops(ctx)->final(ctx);
 }
 
+TEE_Result crypto_cipher_alloc_key(struct tee_cryp_obj_key *key,
+				   uint32_t key_type)
+{
+	TEE_Result res = TEE_SUCCESS;
+
+	/*
+	 * The Crypto driver might need more space than the security size of
+	 * the key. That's why we need to add the allocation() interface with
+	 * the crypto driver.
+	 */
+	res = drvcrypt_cipher_alloc_key(key, key_type);
+
+	if (res == TEE_ERROR_NOT_IMPLEMENTED) {
+		/*
+		 * The CIPHER driver does not have to implement this
+		 * alloc_key() operation.
+		 * If it is not implemented, we fallback to a simple sw
+		 * allocation where the size allocated for the key is equal
+		 * to the security size of the key.
+		 *
+		 * return calloc(key->size) ...
+		 */
+	}
+}
+
+TEE_Result crypto_cipher_gen_key(struct tee_cryp_obj_key *key)
+{
+	TEE_Result res = TEE_SUCCESS;
+
+	/*
+	 * Let the Crypto driver generate the key.
+	 * The key can be:
+	 * - plain text : in this case, we can call the crypto driver RNG or
+	 *                even return not implemented to use the regular
+	 *                SW implementation crypto_rng_read() below.
+	 * - encrypted : in this case, the crypto driver will generate the
+	 *               encrypted key accordingly.
+	 */
+	res = drvcrypt_cipher_gen_key(key);
+
+	if (res == TEE_ERROR_NOT_IMPLEMENTED) {
+		/*
+		 * The CIPHER driver does not have to implement this
+		 * gen_key() operation.
+		 * If it is not implemented, we fallback to a simple SW
+		 * generation like it was done for TEE_TYPE_GENERIC_SECRET.
+		 *
+		 * crypto_rng_read(key->data, key->size);
+		 */
+	}
+}
+
 TEE_Result crypto_cipher_get_block_size(uint32_t algo, size_t *size)
 {
 	uint32_t class = TEE_ALG_GET_CLASS(algo);
@@ -219,6 +271,47 @@ TEE_Result crypto_cipher_get_block_size(uint32_t algo, size_t *size)
 		return TEE_SUCCESS;
 	default:
 		return TEE_ERROR_NOT_SUPPORTED;
+	}
+}
+
+TEE_Result crypto_mac_alloc_key(struct tee_cryp_obj_key *key, uint32_t key_type)
+{
+	TEE_Result res = TEE_SUCCESS;
+
+	/*
+	 * The Crypto driver might need more space than the security size of
+	 * the key. That's why we need to add the allocation() interface with
+	 * the crypto driver.
+	 */
+	res = drvcrypt_mac_alloc_key(key, key_type);
+
+	if (res == TEE_ERROR_NOT_IMPLEMENTED) {
+		/*
+		 * The MAC driver does not have to implement this
+		 * alloc_key() operation.
+		 * If it is not implemented, we fallback to a simple sw
+		 * allocation.
+		 *
+		 * return calloc(key->size) ...
+		 */
+	}
+}
+
+TEE_Result crypto_mac_gen_key(struct tee_cryp_obj_key *key)
+{
+	TEE_Result res = TEE_SUCCESS;
+
+	res = drvcrypt_mac_gen_key(key);
+
+	if (res == TEE_ERROR_NOT_IMPLEMENTED) {
+		/*
+		 * The MAC driver does not have to implement this
+		 * gen_key() operation.
+		 * If it is not implemented, we fallback to a simple sw
+		 * generation like it was done for TEE_TYPE_GENERIC_SECRET.
+		 *
+		 * crypto_rng_read(key->data, key->size);
+		 */
 	}
 }
 
--- a/core/drivers/crypto/crypto_api/cipher/cipher.c
+++ b/core/drivers/crypto/crypto_api/cipher/cipher.c
@@ -169,6 +169,45 @@ static const struct crypto_cipher_ops cipher_ops = {
 	.copy_state = cipher_copy_state,
 };
 
+TEE_Result drvcrypt_cipher_alloc_key(struct tee_cryp_obj_key *key,
+				     uint32_t key_type)
+{
+	TEE_Result ret = TEE_ERROR_NOT_IMPLEMENTED;
+	struct crypto_cipher cipher = { };
+
+	if (!key) {
+		CRYPTO_TRACE("Parameters error (key @%p)", key);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	cipher.op = drvcrypt_get_ops(CRYPTO_CIPHER);
+	if (cipher.op && cipher.op->alloc_key)
+		ret = cipher.op->alloc_key(key, key_type);
+
+	CRYPTO_TRACE("Cipher alloc ret 0x%" PRIX32, ret);
+
+	return ret;
+}
+
+TEE_Result drvcrypt_cipher_gen_key(struct tee_cryp_obj_key *key)
+{
+	TEE_Result ret = TEE_ERROR_NOT_IMPLEMENTED;
+	struct crypto_cipher cipher = { };
+
+	if (!key) {
+		CRYPTO_TRACE("Parameters error (key @%p)", key);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	cipher.op = drvcrypt_get_ops(CRYPTO_CIPHER);
+	if (cipher.op && cipher.op->gen_key)
+		ret = cipher.op->gen_key(key);
+
+	CRYPTO_TRACE("Cipher alloc ret 0x%" PRIX32, ret);
+
+	return ret;
+}
+
 TEE_Result drvcrypt_cipher_alloc_ctx(struct crypto_cipher_ctx **ctx,
 				     uint32_t algo)
 {
--- a/core/drivers/crypto/crypto_api/include/drvcrypt_cipher.h
+++ b/core/drivers/crypto/crypto_api/include/drvcrypt_cipher.h
@@ -45,6 +45,11 @@ struct drvcrypt_cipher_update {
  * Crypto library cipher driver operations
  */
 struct drvcrypt_cipher {
+	/* Allocate the key */
+	TEE_Result (*alloc_key)(struct tee_cryp_obj_secret *key,
+				uint32_t key_type);
+	/* Generate the key */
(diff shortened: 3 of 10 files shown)


jenswi-linaro:
Thanks for picking up this. Over all I agree with this approach.

We need sort out how RSEC keys are enabled for a TA. With CAAM it seems that there is no limit on the number of black keys and that the overhead in terms of memory and time is nothing to worry too much about. So to keep things simple we could perhaps take the compilation flag approach mentioned above.

This feature is likely to take a while to get through with lots of comments. So we should try to divide the work into multiple PRs (perhaps one PR for each commit in this PR, starting with the first commit). We could keep this PR for to get the big picture and for more generic comments and keep the more detailed comments in the smaller PRs. Github reviews tend to run much smoother on smaller and short lived PRs.

jforissier:
I agree.

> So to keep things simple we could perhaps take the compilation flag approach mentioned above.

Yes, for simplicity I would recommend not adding flags to generic layers (I mean flags like `CFG_CRYPTO_DRV_[CIPHER, MAC, RSA ...] _RSEC`) and leave it to the actual crypto driver such as CAAM or whatever to use RSEC or not. Perhaps when the hardware supports it you don't even want to disable it.

clementfaure:
Hello,

Thanks for your feedback!

@jenswi-linaro 
Ok, we will go with the compilation flag to enable by default the use of RSEC keys in the crypto driver. The use of RSEC keys will be completely transparent for the TA. If we go this way, we can discard the commit 558f1733c7113c0b55c73460f8ad3ca825d46176 since the TEE Crypto API is no longer needed. We can also discard the `bool/uint32_t status;` in `struct tee_cryp_obj_key {}` since the key handling by the crypto driver will be based on the compilation flag.

At first, we wanted to have something integrated to the TEE Crypto API but as we mentioned, we are not satisfied with what we have come up to. We will go for the compilation flag for now ; nothing prevent us from adding the TEE Crypto API integration later if we have a better idea. Feel free to share if you have other suggestions :)

Now you have acknowledge the overall idea of the RFC, we will start the implementation ... For the upstreaming phase, I think we will start with 4bcd7bf393c3272b7ee2b8c30da34974d708ce69. This task will be pretty big and we will split it into multiple PRs to make the review easier as you suggest.

Yes, we will keep this RFC for generic comments and discussions.

@jforissier 
You're right, this flag should belong to the CAAM driver and not the Crypto HW API. We will to do that instead.

clementfaure:
Hello,
I'm re-opening the RFC to get some input about some concerns we have and our suggestion to get rid of the bignum format from the core.
In the GP TEE Internal Core API, the `Table 6-15: Object or Operation Attributes` specifies that all asymmetric key components must be stored as a `bignum` format.
However, our implementation for RSEC would have these components stored as `binary` instead (please refer to `Table 6-16: Attribute Format Definitions`).
Is such a change in the key format a problem for OPTEE implementation and its GP compliancy?

jenswi-linaro:
I guess the TAs will still expect to see attributes in bignum format, perhaps we can convert while copying out to user space.

qc-azarrabi:
I would like to reopen the discussion around this RFC, since the underlying problem still seems relevant.

Before proposing a specific next step, it would be very helpful to understand whether there is any updated maintainer view on this topic, whether the original authors still have plans around it, or whether there is any preferred direction for resuming the discussion.

If this topic is still of interest upstream, I would appreciate guidance on whether maintainers would prefer:
1. Continuing with the broader RSEC framework direction discussed here, or
2. Restarting from a smaller scope first.

jenswikl:
Unless NXP has any ambition here, I'd suggest restarting in a smaller scope.

PR #4811, opened 2021-08-24
ta: pkcs11: fix error code in asymmetric signature update sequence

etienne-lms:
Correct return code in asymmetric update sequence when digest of
the input data is updated on a multi-stage operation. Prior this
change, the implementation returned CKR_GENERAL_ERROR instead of
CKR_OK because the expected success return value was loaded for
that stage.

Fixes: fb279d8b608e ("ta: pkcs11: Add support for elliptic curve signing & verification")

### DIFF
--- a/ta/pkcs11/src/processing_asymm.c
+++ b/ta/pkcs11/src/processing_asymm.c
@@ -526,6 +526,7 @@ enum pkcs11_rc step_asymm_operation(struct pkcs11_session *session,
 
 			TEE_DigestUpdate(proc->tee_hash_op_handle, in_buf,
 					 in_size);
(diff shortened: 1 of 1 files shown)


ruchi393:
`Reviewed-by:Ruchika Gupta <ruchika.gupta@linaro.org>`

etienne-lms:
tag applied.

jforissier:
The QEMUv8 CI failure is unrelated (fix in https://github.com/OP-TEE/manifest/pull/198) but there is a checkpatch warning.

etienne-lms:
Thanks to notice, I'll fix.

PR #4812, opened 2021-08-24
Pkcs11 cfg

etienne-lms:
* mk/config.mk: describe CFG_PKCS11_TA_TOKEN_COUNT
Add a default value and a description for PKCS11 TA config switch CFG_PKCS11_TA_TOKEN_COUNT.

* ta: pkcs11: configurable heap size
Add a configuration switch for the PKCS11 TA heap size defaulting to 16kB as legacy.


ruchi393:
`Reviewed-by: Ruchika Gupta <ruchika.gupta@lianro.org>`

mk/config.mk
```diff
@@ -609,12 +609,18 @@ CFG_CRYPTOLIB_DIR ?= core/lib/libtomcrypt
...
+# Defines the number of PKCS11 token implemented by the PKCS11 TA
+CFG_PKCS11_TA_TOKEN_COUNT ?= 3
+
 # Enable PKCS#11 TA's TEE Identity based authentication support
 CFG_PKCS11_TA_AUTH_TEE_IDENTITY ?= y
 
 # Enable PKCS#11 TA's C_DigestKey support
 CFG_PKCS11_TA_ALLOW_DIGEST_KEY ?= y
 
+# Enable PKCS#11 TA's C_DigestKey support
```
jforissier, line 621: This comment is weird. So if `CFG_PKCS11_TA_HEAP_SIZE` is not defined or has a different value, C_DigestKey is not enabled?

ta/pkcs11/src/user_ta_header_defines.h
```diff
@@ -15,7 +15,12 @@
 					 TA_FLAG_INSTANCE_KEEP_ALIVE)
 
 #define TA_STACK_SIZE			(4 * 1024)
+
+#ifdef CFG_PKCS11_TA_HEAP_SIZE
```
jforissier, line 19: Why the `#ifdef` since `CFG_PKCS11_TA_HEAP_SIZE` has a default value in `mk/config.mk`? (I am suspecting a possible issue if the PKCS#11 TA had to be built outside the optee_os environment since `CFG_PKCS11_TA_HEAP_SIZE` is not exported; do we need something like what is done for `CFG_TEE_TA_LOG_LEVEL` [here](https://github.com/OP-TEE/optee_os/blob/3.14.0/ta/ta.mk#L41)?)

jforissier (commented):
For commit "mk/config.mk: describe CFG_PKCS11_TA_TOKEN_COUNT":

For the other one, comments below.

mk/config.mk
etienne-lms: oh sorry, looks I didn't double check this. obviously not the expected description.

ta/pkcs11/src/user_ta_header_defines.h
etienne-lms: I would prefer not to add TA specific switch in devkit generic _ta.mk_.
I think it is easier to treat `CFG_PKCS11_TA_HEAP_SIZE` (a value) with `#ifdef` pragma to handle case where it is not defined and don't add burden in mandatory configs, as other `CFG_PKCS11_TA_xxx` switch that don't really mandate a default value (since not defined means disabled) despite we set one in _mk/config.mk_.

mk/config.mk
etienne-lms: note we already do the same with `CFG_PKCS11_TA_TOKEN_COUNT` in [_ta/pkcs11/src/pkcs11_token.c_](https://github.com/OP-TEE/optee_os/blob/3.14.0/ta/pkcs11/src/pkcs11_token.c#L28-L32)

ta/pkcs11/src/user_ta_header_defines.h
etienne-lms: We already do the same with `CFG_PKCS11_TA_TOKEN_COUNT` in [_ta/pkcs11/src/pkcs11_token.c_](https://github.com/OP-TEE/optee_os/blob/3.14.0/ta/pkcs11/src/pkcs11_token.c#L28-L32).

etienne-lms:
updated

ta/pkcs11/src/user_ta_header_defines.h
jforissier: Right. How about moving the `CFG_PKCS11_TA*` definitions to `ta/pkcs11/sub.mk`? Then they would always get a default value, and the `#ifdef` could be removed. In addition it makes more sense to define them in the TA folder since they are used only by the TA.

etienne-lms: I through we didn't want to set `CFG_xxx` values from _sub.mk_ files.

jforissier: It is allowed, although rarely done I agree, because we prefer to use them only to list source files and sub-directories. But that's because there is often a separate `.mk` file where more complex things are done (such as `core.mk` etc.). From a `make` point of view it doesn't make a difference. `sub.mk` is even listed as a valid place to define `CFG_` flags in `mk/config.mk`, see [here](https://github.com/OP-TEE/optee_os/blob/3.14.0/mk/config.mk#L14).

etienne-lms: ok, great, thanks.
I'll update this P-R to move all `CFG_PKCS11_TA_*` switches into _ta/pkcs11/sub.mk_.

etienne-lms:
(force-)updated. review tags removed.

ta/pkcs11/src/pkcs11_token.c
```diff
@@ -24,12 +24,8 @@
 #include "serializer.h"
 #include "token_capabilities.h"
 
-/* Provide 3 slots/tokens, ID is token index */
-#ifndef CFG_PKCS11_TA_TOKEN_COUNT
-#define TOKEN_COUNT		3
-#else
+/* Number of tokens implemented by the TA. Token ID is the token index */
 #define TOKEN_COUNT		CFG_PKCS11_TA_TOKEN_COUNT
```
jforissier, line 28: You could as well use `CFG_PKCS11_TA_TOKEN_COUNT` in the code below, but I'm fine with keeping this define if you prefer (it is indeed shorter / easier to read).

jforissier (approved):

ta/pkcs11/src/pkcs11_token.c
etienne-lms: agree. The macro is used at several places in this file. Easier to read with `TOKEN_COUNT`.

PR #4818, opened 2021-08-27
core: arm64: remove duplicate definition of THREAD_CTX_SIZE

jforissier:
Following commit 1b302ac09816 ("core: enable FF-A with SPM Core at
S-EL1"), the THREAD_CTX_SIZE macro is now defined twice on arm64
(in $(out-dir)/core/include/generated/asm-defines.h).

Kill the definition in the #ifdef ARM64 block and keep the common
one.


jenswi-linaro:

PR #4825, opened 2021-08-31
core: improve version string shown at boot time when not using Git

jforissier:
When OP-TEE is built outside of a Git repository, the implementation
version is shown as "Unknown" at boot time. For example:

 I/TC: OP-TEE version: Unknown (gcc version 10.2.1 ...

Improve this a bit by appending the major/minor revision of OP-TEE,
which is hardcoded in mk/config.mk and thus always available.

The above example becomes:

 I/TC: OP-TEE version: Unknown_3.14 (gcc version 10.2.1 ...


jenswi-linaro (commented):

PR #4829, opened 2021-09-01
build: ldelf and TAs can rely on CFLAGS32/CFLAGS64

etienne-lms:
Defines arch-bits-$(sm) for ldelf and intree TAs sub components
so that they can build using CFLAGS32 (or CFLAGS64) directives
possibly passed by the build environment.

Defines arch-bits-ta_arm32 (resp. 64) in TA devkit to leverage
CFLAGS32 (reps. CFLAGS64) directive passed by the build process. This
change is needed for external package willing to pass specific
directive to TA build sequence as toolchain's sysroot path.

### DIFF
--- a/ta/mk/build-user-ta.mk
+++ b/ta/mk/build-user-ta.mk
@@ -1,4 +1,6 @@
 include mk/cleanvars.mk
+
+# Set $(sm) as the name of the in tree TA being built, for instance "avb" or "pkcs11"
 sm := $(lastword $(subst /, ,$(dir $(ta-mk-file))))
 sm-$(sm) := y
 
@@ -6,6 +8,9 @@ sm-$(sm) := y
 ta-target := $(strip $(if $(CFG_USER_TA_TARGET_$(sm)), \
 		$(filter $(CFG_USER_TA_TARGET_$(sm)), $(ta-targets)), \
 		$(default-user-ta-target)))
+
+arch-bits-ta_arm32 := 32
+arch-bits-ta_arm64 := 64
 
 ta-dev-kit-dir$(sm) := $(out-dir)/export-$(ta-target)
 link-out-dir$(sm) := $(out-dir)/$(patsubst %/,%, $(dir $(ta-mk-file)))
--- a/ta/mk/ta_dev_kit.mk
+++ b/ta/mk/ta_dev_kit.mk
@@ -26,6 +26,8 @@ libname := $(LIBNAME)
 shlibname := $(SHLIBNAME)
 shlibuuid := $(SHLIBUUID)
 
+arch-bits-ta_arm32 := 32
+arch-bits-ta_arm64 := 64
 
 ifneq ($V,1)
 q := @
--- a/ldelf/ldelf.mk
+++ b/ldelf/ldelf.mk
@@ -17,6 +17,7 @@ endif
 ifeq ($(CFG_ARM32_core),y)
 CFG_ARM32_$(sm) := y
 endif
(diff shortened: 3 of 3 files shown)


ta/mk/build-user-ta.mk
```diff
@@ -2,6 +2,13 @@ include mk/cleanvars.mk
 sm := $(lastword $(subst /, ,$(dir $(ta-mk-file))))
 sm-$(sm) := y
 
+ifeq ($(ta-target),ta_arm32)
```
jenswi-linaro, line 5: Will this work? `ta-target` is reassigned below.

etienne-lms: Indeed this works. Maybe not for the good reasons :)

jenswi-linaro: How about replacing with with
```
ifeq ($(sm),ta_arm32)
arch-bits-ta_arm32 := 32
endif
ifeq ($(sm),ta_arm64)
arch-bits-ta_arm64 := 64
endif
```

below the assignment to be more clear and do it the same way as in `ta/mk/ta_dev_kit.mk`

ta/mk/build-user-ta.mk
```diff
@@ -2,18 +2,18 @@ include mk/cleanvars.mk
 sm := $(lastword $(subst /, ,$(dir $(ta-mk-file))))
 sm-$(sm) := y
 
+# Select TA target (aka TA dev kit, when delivered)
+ta-target := $(strip $(if $(CFG_USER_TA_TARGET_$(sm)), \
+		$(filter $(CFG_USER_TA_TARGET_$(sm)), $(ta-targets)), \
+		$(default-user-ta-target)))
+
 ifeq ($(ta-target),ta_arm32)
 arch-bits-$(sm) := 32
```
jenswi-linaro, line 11: You're checking `$(ta-target)`, but then you're using `$(sm)`. I can't make sense of that.

etienne-lms: `sm`, from _build-user-ta.mk_ is the in-tree TA name (avb, pkcs11, ...).
As build process uses `CFLAGS$(arch-bits-$(sm))` (to refer to `CFLAGS32`/`CFLAGS64`), we must set `arch-bits-$(sm)` for each TA name.

External TAs use _ta_dev_kit.mk_, where `sm` is `ta_arm32` or `ta_arm64`, so a different path.

jenswi-linaro: Aha, thanks for the clarification. It might be worth adding a comment like:
_# $(sm) here is the name of the binary being built, for instance "avb" or "pkcs11"_

jenswi-linaro:

ta/mk/ta_dev_kit.mk
```diff
@@ -26,6 +26,12 @@ libname := $(LIBNAME)
 shlibname := $(SHLIBNAME)
 shlibuuid := $(SHLIBUUID)
 
+ifeq ($(sm),ta_arm32)
+arch-bits-ta_arm32 := 32
+endif
+ifeq ($(sm),ta_arm64)
+arch-bits-ta_arm64 := 64
+endif
```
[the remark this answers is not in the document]
etienne-lms: updated, in both _ta_dev_kit.mk_ and _build-user-ta.mk_.

jforissier (approved):
Looks good!

etienne-lms:
Squashed, commit log updated, review tags applied.

PR #4832, opened 2021-09-03
mk: fix test on gprof and shared libs configuration for TAs

etienne-lms:
Change test for when both CFG_TA_GPROF_SUPPORT and CFG_ULIBS_SHARED
are enabled to be more strict on switches expected value.

Reported-by: Clément Léger <clement.leger@bootlin.com>


jforissier (approved):
Perhaps give/explain an example for the problematic case in the commit description? Anyways:

etienne-lms:
updated commit log and applied tag.
@jforissier, is the description ok?

PR #4834, opened 2021-09-04
libutils: Handle zero-sized buffer allocations

Sadiq1811:
The GlobalPlatform TEE internal api specification mentions the following
about zero sized buffer allocations:
"The value returned is undefined but guaranteed to be different from NULL.
The Trusted Application SHALL NOT access the returned pointer.
The Trusted Application SHOULD panic if the memory pointed to by such a pointer is accessed for
either read or write"

But, we would never observe a TA panic because a zero size is internally
translated to 1 and finally to (2 * sizeof(long)) in the bget() function.

This patch handles this aspect so to naturally pass size=0 too across
the different layers.

Signed-off-by: Sadiq Hussain <sadiq.muchumarri@intel.com>

### DIFF
--- a/lib/libutee/tee_api.c
+++ b/lib/libutee/tee_api.c
@@ -12,6 +12,12 @@
 #include <utee_syscalls.h>
 #include "tee_api_private.h"
 
+/*
+ * return a known non-NULL invalid pointer when the
+ * requested size is zero
+ */
+#define TEE_NULL_SIZED_VA	((void *)1)
+
 static const void *tee_api_instance_data;
 
 /* System API - Internal Client API */
@@ -392,6 +398,9 @@ void TEE_GetREETime(TEE_Time *time)
 
 void *TEE_Malloc(uint32_t len, uint32_t hint)
 {
+	if (!len)
+		return TEE_NULL_SIZED_VA;
+
 	if (hint == TEE_MALLOC_FILL_ZERO)
 		return calloc(1, len);
 	else if (hint == TEE_USER_MEM_HINT_NO_FILL_ZERO)
@@ -404,12 +413,21 @@ void *TEE_Malloc(uint32_t len, uint32_t hint)
 
 void *TEE_Realloc(void *buffer, uint32_t newSize)
 {
+	if (!newSize) {
+		TEE_Free(buffer);
+		return TEE_NULL_SIZED_VA;
+	}
+
+	if (buffer == TEE_NULL_SIZED_VA)
+		return calloc(1, newSize);
+
 	return realloc(buffer, newSize);
 }
 
 void TEE_Free(void *buffer)
 {
-	free(buffer);
+	if (buffer != TEE_NULL_SIZED_VA)
+		free(buffer);
 }
 
 /* Cache maintenance support (TA requires the CACHE_MAINTENANCE property) */


etienne-lms (commented):
Are the bget changes bulletproof when playing with 0-byte allocation and releases? What pointer value do you expect from the allocation?

This change affects OP-TEE core and TAs whereas the spec you mention targets only the TA execution environment. Could this be handled in libutee?
Maybe patching `TEE_Malloc()` and friends to return a known invalid non NULL pointer (I.e. `(void *)1`) when requested size is 0. TA would segfault on access and `TEE_CheckMemoryAccessRights()` would properly return the address can't be accessed.

If patch bget, please are tests for 0-byte allocation in `self_test_malloc()`.

jenswi-linaro:
> This change affects OP-TEE core and TAs whereas the spec you mention targets only the TA execution environment. Could this be handled in libutee?
> Maybe patching `TEE_Malloc()` and friends to return a known invalid non NULL pointer (I.e. `(void *)1`) when requested size is 0. TA would segfault on access and `TEE_CheckMemoryAccessRights()` would properly return the address can't be accessed.

That was my thought too. Perhaps with some hints in the lower bits too for `TEE_Realloc()`.

Sadiq1811:
Thanks @etienne-lms and @jenswi-linaro for your feedback.
New changes have been pushed.

jenswi-linaro:
I don't think this will handle example very well:
``` C
p = TEE_Malloc(0, TEE_MALLOC_FILL_ZERO);
p = TEE_Realloc(p, 22);
```

lib/libutee/tee_api.c
```diff
@@ -392,6 +392,12 @@ void TEE_GetREETime(TEE_Time *time)
 
 void *TEE_Malloc(uint32_t len, uint32_t hint)
 {
+	/*
+	 * return a known non-NULL invalid pointer when the
+	 * requested size is zero
+	 */
+	if (len == 0)
+		return (void *)1;
```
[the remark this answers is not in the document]
jenswi-linaro: That macro should probably take the hints bits as argument too.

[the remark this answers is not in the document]
Sadiq1811: So, currently only TEE_MALLOC_FILL_ZERO hint is supported by OPTEE right?
There is no implementation yet for TEE_MALLOC_NO_FILL and TEE_MALLOC_NO_SHARE?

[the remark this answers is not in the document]
jenswi-linaro: Assume that we'll support `TEE_MALLOC_NO_FILL` and `TEE_MALLOC_NO_SHARE` in a not too distant future, that is, do make it harder than necessary to support those later on. Note that we have one extension, `TEE_USER_MEM_HINT_NO_FILL_ZERO`, which behaves like `TEE_MALLOC_NO_FILL`.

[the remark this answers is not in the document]
Sadiq1811: OK
Can we use a static function instead of a multi-line macro?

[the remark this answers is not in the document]
jenswi-linaro: Sure

lib/libutee/tee_api.c
```diff
@@ -404,12 +420,16 @@ void *TEE_Malloc(uint32_t len, uint32_t hint)
 
 void *TEE_Realloc(void *buffer, uint32_t newSize)
 {
-	return realloc(buffer, newSize);
```
jenswi-linaro, line 407: I was expected something like this:
``` C
if (is_zero_sized(buffer)) 
        return TEE_Malloc(newSize, get_zero_sized_hints(buffer));

return realloc(buffer, newSize);

Comment (jforissier) [reply lib/libutee/tee_api.c:407]: I don't want to derail the review, but... how does this handle the hint?
```
return realloc(buffer, newSize);
```

Comment (jenswi-linaro) [reply lib/libutee/tee_api.c:407]: It currently doesn't. I thought we were storing the hint somewhere, but as it turns out `realloc()` is always clearing the remaining part of a buffer.

With that in mind, perhaps I'm asking for a bit much.

Comment (etienne-lms) [reply lib/libutee/tee_api.c:407]: The hint is the address itself. Something like:

```c
void *TEE_Realloc(void *buffer, uint32_t newSize)
{
	if (!newSize) {
		free(buffer);
		return (void *)1;
	}

	if (buffer == (void *)1)
		return malloc(newSize);

	return realloc(buffer, newSize);
}
```

jenswi-linaro: Hah, it seems we all suggest half broken solutions. :-)
If the original zero sized buffer was obtained with `TEE_Malloc(0, TEE_MALLOC_FILL_ZERO)`, then the caller would be right to expect a zeroed buffer. So `calloc()` would be a better option than malloc() here.

etienne-lms: you're right:
```patch
+ #define NULLSIZE_VA ((void *)1)
+
void *TEE_Malloc(uint32_t len, uint32_t hint)
{
+	if (!len)
+		return NULLSIZE_VA;
+
	if (hint == TEE_MALLOC_FILL_ZERO)
		return calloc(1, len);
	else if (hint == TEE_USER_MEM_HINT_NO_FILL_ZERO)
		return malloc(len);
	EMSG("Invalid hint %#" PRIx32, hint);
	return NULL;
}

void *TEE_Realloc(void *buffer, uint32_t newSize)
{
+	if (!newSize) {
+		free(buffer);
+		return NULLSIZE_VA;
+	}
+
+	if (buffer == NULLSIZE_VA)
+		return calloc(1, newSize);
+
	return realloc(buffer, newSize);
}

void TEE_Free(void *buffer)
{
-	free(buffer);
+	if (buffer != NULLSIZE_VA)
+		free(buffer);
}

Comment (jenswi-linaro) [reply lib/libutee/tee_api.c:407]: Replace the call to `free()` with `TEE_Free()` in `TEE_Realloc()` and we should be done.

Comment (etienne-lms) [reply lib/libutee/tee_api.c:407]: right :)
@Sadiq1811, can you update your P-R with the outcome of the discussion?

Comment (Sadiq1811) [reply lib/libutee/tee_api.c:407]: done

File: lib/libutee/tee_api.c [hunk 3]
```diff
@@ -14,6 +14,11 @@
 
 static const void *tee_api_instance_data;
 
+static void *tee_null_sized_va(void)
+{
+	return (void *)1;
+}
+
```
jforissier, line 21: Why not simply `#define NULL_SIZED_VA (void *)1`?

Sadiq1811: tried to keep it modular for readability purpose

etienne-lms:
In commit log:
- missing `<` and `>` around mail address: `Signed-off-by: Real Name <some@address>`.
- prefer `API` over `api`.

Sadiq1811:
> In commit log:
> 
> * missing `<` and `>` around mail address: `Signed-off-by: Real Name <some@address>`.
> * prefer `API` over `api`.

Done

lib/libutee/tee_api.c
jforissier: I find the function slightly *less* readable if anything...

Sadiq1811: No problem - shall change it

etienne-lms: fine

lib/libutee/tee_api.c
```diff
@@ -12,6 +12,8 @@
 #include <utee_syscalls.h>
 #include "tee_api_private.h"
 
+#define TEE_NULL_SIZED_VA ((void *)1)
```
etienne-lms, line 15: minor: prefer a tab to better separates macro name from its value

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with minor comment addressed or not.

lib/libutee/tee_api.c
```diff
@@ -392,6 +394,13 @@ void TEE_GetREETime(TEE_Time *time)
 
 void *TEE_Malloc(uint32_t len, uint32_t hint)
 {
+	/*
+	 * return a known non-NULL invalid pointer when the
+	 * requested size is zero
+	 */
```
etienne-lms, line 400: also move this description to `TEE_NULL_SIZED_VA` definition.

Sadiq1811: done and tag added too

jforissier (commented):

Sadiq1811:
Tags have been applied to the commit message

jenswi-linaro:

Sadiq1811:
Anything else to be done from my side?
The build seems to have got auto cancelled due to time-out.

jenswi-linaro:
> Anything else to be done from my side?

No, hopefully not.

> The build seems to have got auto cancelled due to time-out.

It looks like it might be temporary problem, I've restarted the jobs.

Sadiq1811:
> > The build seems to have got auto cancelled due to time-out.
> 
> It looks like it might be temporary problem, I've restarted the jobs.

Thank you

jforissier:
@Sadiq1811 thanks for the patch

PR #4839, opened 2021-09-06
plat-stm32mp1: Don't call get_embedded_dt() without CFG_EMBED_DT

mrnuke:
Several pieces of stm32mp1 code call get_embedded_dt(), then use the
resulting pointer without checks, or initiate a panic if it is NULL.
Thus hitting this code results in a non-working binary. For example:

    "PLATFORM=stm32mp1 CFG_DT=y"

The get_embedded_dt() uses were #ifdef'd out based on CFG_DT. However,
as shown, this is problematic, as the calls assumed a valid fdt must
be returned. A non-NULL fdt can be guaranteed with CFG_EMBED_DT, so
use this as the basis for the #ifdefs.

Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>

### DIFF
--- a/core/arch/arm/plat-stm32mp1/main.c
+++ b/core/arch/arm/plat-stm32mp1/main.c
@@ -104,7 +104,7 @@ void console_init(void)
 	IMSG("Early console on UART#%u", CFG_STM32_EARLY_CONSOLE_UART);
 }
 
-#ifdef CFG_DT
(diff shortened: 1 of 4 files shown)


etienne-lms (commented):

jforissier:
@mrnuke can you please add @etienne-lms's R-b tag? Thanks.

PR #4844, opened 2021-09-07
Virt updates

jenswi-linaro:
A few random patches accumulated while working on the proof of concept FF-A mediator for Xen, https://linaro.atlassian.net/browse/STR-22 .


lorc:
Regarding "core: virt: forbid creating guest_id 0":

There are not only Type 1 hypervisors like Xen, but also Type 2 hypervisors like KVM. I believe, KVM-based hypervisor might want to create partition for itself.

core/arch/arm/kernel/virtualization.c
```diff
@@ -267,6 +267,8 @@ uint32_t virt_guest_created(uint16_t guest_id)
 
 	/* Initialize threads */
 	thread_init_threads();
+	/* Do the preinitcalls */
+	call_preinitcalls();
```
etienne-lms, line 276: It is strange that `preinitcalls()` are specifically handle in virtualization.  It would make more sense 
to add an explciit `virt-init` initcall level, specific to virtualization stuff. But I see the dynshm is registered as `preinit`  level aside CFG_VIRTUALIZATION consideration :(

core/arch/arm/kernel/virtualization.c
```diff
@@ -169,23 +169,23 @@ void virt_init_memory(struct tee_mmap_region *memory_map)
 }
 
 
-static int configure_guest_prtn_mem(struct guest_partition *prtn)
+static TEE_Result configure_guest_prtn_mem(struct guest_partition *prtn)
 {
-	int ret;
+	TEE_Result res;
```
etienne-lms, line 174: = TEE_xxx;  note the function either succeeds or returns OUT_OF_MEMORY.

core/include/initcall.h
```diff
@@ -30,12 +30,20 @@ struct initcall {
...
+			SCATTERED_ARRAY_BEGIN(preinitcall, struct initcall)
+#define preinitcall_end SCATTERED_ARRAY_END(preinitcall, struct initcall)
+
 #define initcall_begin	SCATTERED_ARRAY_BEGIN(initcall, struct initcall)
 #define initcall_end	SCATTERED_ARRAY_END(initcall, struct initcall)
 
 #define finalcall_begin	SCATTERED_ARRAY_BEGIN(finalcall, struct initcall)
 #define finalcall_end	SCATTERED_ARRAY_END(finalcall, struct initcall)
 
+#define preinit_early(fn)		__define_initcall(preinit, 1, fn)
```
lorc, line 43: I think you need to document somewhere that this calls are meant to be used by TEE part, not by Nexus

core/arch/arm/kernel/virtualization.c
```diff
@@ -322,14 +322,8 @@ uint32_t virt_guest_destroyed(uint16_t guest_id)
...
 {
-	struct guest_partition *prtn;
-	uint32_t exceptions;
-
-	prtn = get_current_prtn();
-
-	/* This can be true only if we return from IRQ RPC */
-	if (prtn && prtn->id == guest_id)
-		return true;
+	struct guest_partition *prtn = get_current_prtn();
```
lorc, line 328: Just to clarify. Are you sure that foreign IRQ handler calls `virt_unset_guest()`?

I can't find this call and when I wrote this part of the code, this wasn't the case.

core/arch/arm/kernel/virtualization.c
lorc: Well, initially there was somewhat different approach: we had Nexus (kernel) services, which should be initialized during the boot. Those services are shared by all the OP-TEE code.

And we had "TEE" services which were initialized for each VM instance.

jenswi-linaro:
> Regarding "core: virt: forbid creating guest_id 0":
> 
> There are not only Type 1 hypervisors like Xen, but also Type 2 hypervisors like KVM. I believe, KVM-based hypervisor might want to create partition for itself.

Yes, that would make sense. But until we have a working implementation for that let's be explicit.

core/arch/arm/kernel/virtualization.c
jenswi-linaro: Yes, that's done in `thread_state_suspend()`. Without this a thread could never migrate to another core.

lorc: Ah, yes, now I can see it. Thanks.

core/arch/arm/kernel/virtualization.c
jenswi-linaro: With the proof of concept implementation mentioned in https://linaro.atlassian.net/browse/STR-22 we need to be able allocate memory from the partition heap rather early (calling `mobj_ffa_sel1_spmc_new()` from `add_mem_share()`). At the preinit levels we're doing as little as possible to allow the guest to minimize time spent on guest initialization.

The idea was to keep the order of the preinit/init calls the same regardless or virtualization. The only difference is that preinit calls are called much earlier compared to the init calls in case of virtualization. 

Here's a few line from the secure log while booting.
This is printed when the guest is initialized:
```
D/TC:0   call_preinitcalls:21 level 1 virt_init_heap()
I/TC: Added guest 1
```

This is printed when the guest driver starts to probe.
```
D/TC:0   tee_entry_exchange_capabilities:102 Dynamic shared memory is enabled
D/TC:0 0 call_initcalls:40 level 1 register_time_source()
D/TC:0 0 call_initcalls:40 level 1 teecore_init_pub_ram()
D/TC:0 0 call_initcalls:40 level 3 mobj_mapped_shm_init()
D/TC:0 0 mobj_mapped_shm_init:439 Shared memory address range: fc00000, 11c00000
D/TC:0 0 call_initcalls:40 level 3 check_ta_store()
D/TC:0 0 check_ta_store:408 TA store: "Secure Storage TA"
D/TC:0 0 check_ta_store:408 TA store: "REE"
D/TC:0 0 call_initcalls:40 level 3 verify_pseudo_tas_conformance()
D/TC:0 0 call_initcalls:40 level 3 tee_cryp_init()
D/TC:0 0 call_initcalls:40 level 4 tee_fs_init_key_manager()
D/TC:0 0 call_initcalls:40 level 6 mobj_init()
D/TC:0 0 call_initcalls:40 level 6 default_mobj_init()
```

core/include/initcall.h
jenswi-linaro: Good point, I'll add something.

core/arch/arm/kernel/virtualization.c
jenswi-linaro: I'll fix.

jenswi-linaro:
Update

core/include/initcall.h
```diff
@@ -30,12 +30,67 @@ struct initcall {
...
+ *
+ *  +-------------------------------+-----------------------------------+
+ *  | Without virtualization        | With virtualization               |
+ *  +-------------------------------+-----------------------------------+
+ *  | At the end of boot_init_primary_late() just before the print:     |
+ *  | "Primary CPU switching to normal world boot"                      |
+ *  +-------------------------------+-----------------------------------+
+ *  | 1. call_preinitcalls()        | In the nexus                      |
+ *  | 2. call_initcalls()           +-----------------------------------+
+ *  | 3. call_finalcalls()          | 1. call_finalcalls()              |
```
etienne-lms, line 71: `bootfinal` callbacks are called before `*_init` and `preinit` levels. It sounds strange to me

Looking at the registered `boot_final` callbacks, i think they are misused.
`
core/arch/arm/kernel/secure_partition.c:boot_final(sp_init_all);
core/arch/arm/kernel/boot.c:boot_final(release_external_dt);
core/drivers/crypto/se050/glue/i2c.c:boot_final(load_trampoline);
`
The `secure_partition.c` callbacks could be `early_init` level, there are no dependencies.
The `se050/i2c.c` callback should be called only once a VM has been created, no? (`driver_init_late` level)
As for `release_external_dt()`, maybe it should be simply explicitly called from `boot_init_primary_late()` instead of `call_finalcalls()`.

core/include/initcall.h
```diff
@@ -30,12 +30,67 @@ struct initcall {
...
+ *  +-------------------------------+-----------------------------------+
+ *  | "Primary CPU switching to normal world boot" is printed           |
+ *  +-------------------------------+-----------------------------------+
+ *                                  | A guest is created and            |
+ *                                  | virt_guest_created() is called.   |
+ *                                  | After the partition has been      |
```
etienne-lms, line 77: I think you forgot to remove this line and the next one.

core/kernel/initcall.c
```diff
@@ -8,6 +8,25 @@
...
 #include <kernel/linker.h>
 
+/*
+ * Note: this function is weak just to make it possible to exclude it from
+ * the unpaged area.
+ */
+void __weak call_preinitcalls(void)
+{
+	const struct initcall *call = NULL;
+	TEE_Result ret = TEE_SUCCESS;
```
etienne-lms, line 18: s/ret/res/g

core/kernel/initcall.c
```diff
@@ -8,6 +8,25 @@
...
+
+	for (call = preinitcall_begin; call < preinitcall_end; call++) {
+		DMSG("level %d %s()", call->level, call->func_name);
+		ret = call->func();
+		if (ret != TEE_SUCCESS) {
+			EMSG("Preinitcall __text_start + 0x%08" PRIxVA
+			     " failed", (vaddr_t)call - VCORE_START_VA);
```
etienne-lms, line 25: nitpick:
```c
			EMSG("Preinitcall __text_start + %#08"PRIxVA" failed",
			     (vaddr_t)call - VCORE_START_VA);

**Comment by jenswi-linaro** [comment]
I'm taking the comments on `core/include/initcall.h` out of context here since I suppose this can take a wihle to sort out.

> `bootfinal` callbacks are called before `*_init` and `preinit` levels. It sounds strange to me

If we disregard the preinit levels this is an accurate picture of the master branch status.

> 
> Looking at the registered `boot_final` callbacks, i think they are misused:
> 
> ```
> core/arch/arm/kernel/secure_partition.c:boot_final(sp_init_all);
> core/arch/arm/kernel/boot.c:boot_final(release_external_dt);
> core/drivers/crypto/se050/glue/i2c.c:boot_final(load_trampoline);
> ```
> 
> The `secure_partition.c` callbacks could be `early_init` level, there are no dependencies.

This depends on  what will happen with SPs and `CFG_VIRTUALIZATION=y`. I don't think anyone have given much thought on that. I suspect in case it will ever be enabled with virtualization it will either be managed from the nexus or a special partition.

> The `se050/i2c.c` callback should be called only once a VM has been created, no? (`driver_init_late` level)

`boot_final()` was added to handle the I2C case, but I'm afraid that the virtualization case was overlooked. What will happen with I2C with virtualization? My guess is that one of the partitions will claim ownership. The same partition that matches the guest in normal world which takes ownership of the physical I2C controller. So for now I'd say that `load_trampoline` isn't compatible with virtualization.

> As for `release_external_dt()`, maybe it should be simply explicitly called from `boot_init_primary_late()` instead of `call_finalcalls()`.

`boot_final()` was agreed to be the last thing done before exiting to normal world for the first time, this a very good match for what's needed for `release_external_dt` regardless of virtualization.

**Comment by lorc** [comment]
> `boot_final()` was added to handle the I2C case, but I'm afraid that the virtualization case was overlooked. 

Well, problem is that virtualization is not compatible with access to HW devices. So, no I2C, no dedicated crypto accelerators, no RPMB (unless each guest can provide own PRMB partition), and so on.

This is standing issue for which I don't see easy solution, as access to HW requires locking and it is hard to implement locking between VMs in secure and reliable way.

File: core/include/kernel/virtualization.h [hunk 9]
```diff
@@ -42,10 +40,8 @@ uint32_t virt_guest_destroyed(uint16_t guest_id);
  * This function switches memory partitions, so TEE part of
  * OP-TEE will see memory associated with current guest.
  * It should be called on entry to OP-TEE
- *
- * Return: True if VM partition was found and set
  */
-bool virt_set_guest(uint16_t guest_id);
+TEE_Result virt_set_guest(uint16_t guest_id);
```
jforissier, line 45: Since all the functions modified here are only ever called in an `if (...)` statement, the return status could as well be `bool` especially for `virt_guest_created()` and `virt_guest_destroyed()` which sound like they return a boolean already, and making these two return a `TEE_Result` is counter-intuitive IMO: consider `if (virt_guest_created(id)) return <error>;`) above.

jforissier:
For commit "core: virtualization.h: add dummy static inline functions" and "core: replace ifdef CFG_VIRTUALIZATION":

core/include/kernel/virtualization.h
jenswi-linaro: For now it is, but in https://linaro.atlassian.net/browse/STR-22 https://github.com/jenswi-linaro/optee_os.git/tree/xen_ffa we check a bit more in the FF-A case.
STR-22 is just a proof of concept, but I thought that some patches where generic enough to pick up at this stage.

core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -32,7 +32,7 @@ void thread_handle_fast_smc(struct thread_smc_args *args)
 	thread_check_canaries();
 
 #ifdef CFG_VIRTUALIZATION
-	if (!virt_set_guest(args->a7)) {
+	if (virt_set_guest(guest_id)) {
```
lorc: Looks like some rebase artifact or copy-paste mistake. There is no `guest_id` variable introduced in this commit

core/arch/arm/kernel/virtualization.c
```diff
@@ -317,7 +320,7 @@ uint32_t virt_guest_destroyed(uint16_t guest_id)
 	return OPTEE_SMC_RETURN_OK;
```
etienne-lms, line 320: s/OPTEE_SMC_RETURN_OK/TEE_SUCCESS/

core/include/kernel/virtualization.h
```diff
@@ -4,9 +4,11 @@
 #ifndef KERNEL_VIRTUALIZATION_H
 #define KERNEL_VIRTUALIZATION_H
 
+#include <mm/core_mmu.h>
 #include <stdbool.h>
 #include <stdint.h>
-#include <mm/core_mmu.h>
+#include <tee_api_types.h>
+
```
etienne-lms, line 11: remove the empty line

etienne-lms (commented):
Comments for commit "core: virt: use TEE_Result return type for virt_*() functions".
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` once addressed.

Commit "core: virtualization.h: add dummy static inline functions":
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`.

Commit "core: virt: check pa at end of check_pa_matches_va()":

core/arch/arm/kernel/virtualization.c
jenswi-linaro: Thanks, I'll fix.

PR #4850, opened 2021-09-09
tee: entry_fast: support shm address exceed 4G with 32-bits kernel

PeikanTsai:
When using 32-bits ree kernel, shm base address may be truncated if it exceed 4G because the ree gpreg can't handle it.
Return the PFN instead of full physical address to fix it, then we can use the same fastcall to get shm config.
Also, it won't affect the old tee driver in linux kernel because arg1 is always 0.

### DIFF
--- a/core/arch/arm/tee/entry_fast.c
+++ b/core/arch/arm/tee/entry_fast.c
@@ -16,11 +16,26 @@
 #ifdef CFG_CORE_RESERVED_SHM
 static void tee_entry_get_shm_config(struct thread_smc_args *args)
 {
+	unsigned long page_shift, paddr, settings;
+
+	/*
+	 * If the page_shift is set from ree kernel, return the pfn and
+	 * set the corresponding bit in a3 to let the ree kernel know it.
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
It is a clever way of extending the current fastcall, but I'm afraid it might be a bit too clever.
I'd prefer a new fast call since that's less complicated.

PeikanTsai:
Agree that it's kind of tricky.
But I'd like to make the original tee driver compatible with new optee or the old optee with new tee driver.

4 GP registers are not enough for the original information plus the high 32 bits of paddr.
So does "a new fast call" means REE need 2 fast call to get the full shm config?

jenswi-linaro:
This is only an example:
```
a0  size of SHM in number of 4k pages, `0` if not supported
a1  upper 32 bits of a 64-bit physical address
a2  lower 32 bits of a 64-bit physical address
a3  cache settings
```
Or you could use the trick you used before, the point is to avoid complicating already present functions as you need to be very careful to maintain backwards compatibility. Also keep in mind that there may exist other drivers or such being involved in OP-TEE communication.

PR #4854, opened 2021-09-10
core: don't register heap pools when too small

etienne-lms:
Don't add heap1 pool when smaller than 64 bytes when CFG_WITH_PAGER=y.
Bget memory allocation stack fails when pools are too small. This change
does not correct the issue but works around it as losing 64 byte of heap
here is affordable even when OP-TEE pager is enabled.

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -416,8 +416,9 @@ static void init_runtime(unsigned long pageable_part)
 
 	init_asan();
 
-	malloc_add_pool(__heap1_start, __heap1_end - __heap1_start);
+	/* Add heap2 first as heap1 may be too small as initial bget pool */
 	malloc_add_pool(__heap2_start, __heap2_end - __heap2_start);
+	malloc_add_pool(__heap1_start, __heap1_end - __heap1_start);
 
 	/*
 	 * This needs to be initialized early to support address lookup
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -765,15 +765,19 @@ static void gen_malloc_add_pool(struct malloc_ctx *ctx, void *buf, size_t len)
 	uint32_t exceptions;
 	uintptr_t start = (uintptr_t)buf;
 	uintptr_t end = start + len;
-	const size_t min_len = ((sizeof(struct malloc_pool) + (SizeQuant - 1)) &
-					(~(SizeQuant - 1))) +
-				sizeof(struct bhead) * 2;
+	const size_t min_len = sizeof(struct bhead) + sizeof(struct bfhead);
 
 	start = ROUNDUP(start, SizeQuant);
 	end = ROUNDDOWN(end, SizeQuant);
 
 	if (start > end || (end - start) < min_len) {
 		DMSG("Skipping too small pool");
+		return;
+	}
+
+	/* First pool requires a bigger size */
+	if (!ctx->pool_len && (end - start) < MALLOC_INITIAL_POOL_MIN_SIZE) {
+		DMSG("Skipping too small initial pool");
 		return;
 	}
 
--- a/lib/libutils/isoc/include/malloc.h
+++ b/lib/libutils/isoc/include/malloc.h
@@ -8,6 +8,12 @@
 #include <stddef.h>
 #include <types_ext.h>
 
+/*
+ * Due to bget implementation, the first memory pool registered shall have
+ * a min size. Choose 1kB which is reasonable.
+ */
+#define MALLOC_INITIAL_POOL_MIN_SIZE	1024
+
 void free(void *ptr);
 
 #ifdef ENABLE_MDBG
--- a/ta/arch/arm/user_ta_header.c
+++ b/ta/arch/arm/user_ta_header.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2014, STMicroelectronics International N.V.
  */
 #include <compiler.h>
+#include <malloc.h>
 #include <tee_ta_api.h>
 #include <tee_internal_api_extensions.h>
 #include <trace.h>
@@ -90,6 +91,10 @@ const struct ta_head ta_head __section(".ta_head") = {
 };
 
 /* Keeping the heap in bss */
+#if TA_DATA_SIZE < MALLOC_INITIAL_POOL_MIN_SIZE
+#error TA_DATA_SIZE too small
+#endif
+
 uint8_t ta_heap[TA_DATA_SIZE];
 const size_t ta_heap_size = sizeof(ta_heap);
 


jenswi-linaro:
This is checked in `gen_malloc_add_pool()` too, but it results in a smaller number. Could there be an error in the calculation or some missed factor? I'd rather fix the problem in `gen_malloc_add_pool()` than trying to hide it with a magic number.

etienne-lms:
ok, i'll investigate deeper. I faced a issue with a 32 byte pool size.

jenswi-linaro:
Thanks

etienne-lms:
I've update the P-R with a fixup commit and a commit tagged [RFC] for which comment are more than welcome.


Few words about commit "[RFC] libutils: bget_malloc: fix test in pool min size".
It is still unclear to me why min size computed from [`gen_malloc_add_pool()`](https://github.com/OP-TEE/optee_os/blob/3.14.0/lib/libutils/isoc/bget_malloc.c#L768-L770) is expected as:
```c
   ROUNDUP(sizeof(struct malloc_pool), SizeQuant) +
   sizeof(struct bhead) * 2
```
`ROUNDUP(...)` relates to the first size that will be allocated in the pool. Added size relates to the min size bget needs in a pool to store allocation metadata. It is this size that is unclear to me.

From [bpool()](https://github.com/OP-TEE/optee_os/blob/3.14.0/lib/libutils/isoc/bget.c#L1097) implementation, it look like the min size of the pool must also leave room for a `struct bhead` instance and a `struct bfhead` instance. According to me the min size should be:
```c
   ROUNDUP(sizeof(struct malloc_pool), SizeQuant) +
   sizeof(struct bhead) + sizeof(struct bfhead)
```
Experimentation showed that it fails. This min size seems the right value but the first pool registered needs to provide few more bytes to store another `struct bhead` instance, if I understood correctly. A value that works on 32bit and 64bit seems:
```c
   ROUNDUP(sizeof(struct malloc_pool), SizeQuant) +
   sizeof(struct bhead) * 2 + sizeof(struct bfhead)
```
But i am not that confident in this value.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -765,9 +765,9 @@ static void gen_malloc_add_pool(struct malloc_ctx *ctx, void *buf, size_t len)
 	uint32_t exceptions;
 	uintptr_t start = (uintptr_t)buf;
 	uintptr_t end = start + len;
-	const size_t min_len = ((sizeof(struct malloc_pool) + (SizeQuant - 1)) &
-					(~(SizeQuant - 1))) +
-				sizeof(struct bhead) * 2;
+	const size_t min_len = ROUNDUP(sizeof(struct malloc_pool), SizeQuant) +
+			       sizeof(struct bhead) + sizeof(struct bfhead);
```
jenswi-linaro, line 769: This correction makes sense to me too. The old calculation doesn't seem right.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -765,9 +765,9 @@ static void gen_malloc_add_pool(struct malloc_ctx *ctx, void *buf, size_t len)
...
+	const size_t init_min_len = min_len + sizeof(struct bfhead);
```
jenswi-linaro, line 770: This one I can't understand. Are you compiling with `ENABLE_MDBG`, that is, `CFG_TEE_CORE_MALLOC_DEBUG=y` when trying to find the limits?

etienne-lms: No, I did not enable `ENABLE_MDBG`/`CFG_TEE_CORE_MALLOC_DEBUG`.
I tested calling `gen_malloc_add_pool()` for few small pools. The proposed `min_len` implementation worked for all pools added but the first one that needs this extra size, hence this `init_min_len`.

Note that i'm not that confident in the resulting bget_malloc implementation. Not sure it's bulletproof. In OP-TEE scope, I don't think we expect to add many small pools so I would propose to fix only `min_len` implementation and remove the added `min_len_init` part here.

jenswi-linaro: I agree, it's a bit hard to calculate the exactly how much memory the call to `realloc_unlocked()` will need.
How about saying that we'll need at least a 4k page in case it's the first call to `gen_malloc_add_pool()` for this context?

etienne-lms: 4KB is maybe overkilling but I agree an empiric value is better than nothing.

jenswi-linaro: Yes, it's for sure too much, but that's the point also. The thing is that the minimum value is quite hard to calculate since you'd need to take what `buf_get_pos()` does into account when predicting the call to `realloc_unlocked()`. For practical purposes it's enough to pick a number that is large enough to pass without any doubt. I suppose we could use 256 bytes instead if you think that's more reasonable.

etienne-lms: 256 byte or 4kB are both fine to me. Okay for 4kB. If a smaller value is needed at some point, we can adjust.
I'll update the P-R with your proposal.

etienne-lms: Thinking about that, that means TA heap size MUST be higher or equal to 4kB. It is set from `TA_DATA_SIZE` of each TA **user_ta_header_defines.h** header file. We should maybe add a guard like the below in **ta/arch/arm/user_ta_header.c**:
```patch
 /* Keeping the heap in bss */
+#if TA_DATA_SIZE < 4096
+#error TA_DATA_SIZE must be at least 4kByte
+#endif
 uint8_t ta_heap[TA_DATA_SIZE];
 const size_t ta_heap_size = sizeof(ta_heap);
```

jenswi-linaro: Makes sense, but it should be with a define from `<malloc.h>` or such.

etienne-lms: `optee_test/ta/large/include/user_ta_header_defines.h:#define TA_DATA_SIZE	(2 * 1024)` :face_with_head_bandage: 
Fixing OP-TEE test or require 1kB as min heap size?

jenswi-linaro: Let's go for 1kB then.

etienne-lms: updated

core/arch/arm/kernel/boot.c
```diff
@@ -416,8 +416,9 @@ static void init_runtime(unsigned long pageable_part)
 
 	init_asan();
 
-	malloc_add_pool(__heap1_start, __heap1_end - __heap1_start);
+
```
jenswi-linaro, line 419: A short comment on this order would be nice.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -765,9 +765,8 @@ static void gen_malloc_add_pool(struct malloc_ctx *ctx, void *buf, size_t len)
 	uint32_t exceptions;
 	uintptr_t start = (uintptr_t)buf;
 	uintptr_t end = start + len;
-	const size_t min_len = ((sizeof(struct malloc_pool) + (SizeQuant - 1)) &
-					(~(SizeQuant - 1))) +
-				sizeof(struct bhead) * 2;
+	const size_t min_len = ROUNDUP(sizeof(struct malloc_pool), SizeQuant) +
```
jenswi-linaro, line 768: I guess this part isn't needed any longer, that is, we don't care about `sizeof(struct malloc_pool)` if I get it right.

jenswi-linaro (commented):
Please squash the commit.

core/arch/arm/kernel/boot.c
etienne-lms: ok

lib/libutils/isoc/bget_malloc.c
etienne-lms: IIUC a pool must at least fit the size of `struct bhead` and `struct bfhead`. 
Keep at least `min_len = sizeof(struct bhead) + sizeof(struct bfhead) + SizeQuant`?

jenswi-linaro: Isn't `min_len = sizeof(struct bhead) + sizeof(struct bfhead)` enough?

jenswi-linaro: Yes, that should be enough.
One `struct bhead` is always used as a sentinel at the end of this buffer.
A free block is represented with a `struct bfhead`, while a used block only uses a `struct bhead`. So there should be room for at least `struct qlinks` or `2 * sizeof(void *)`.
That's ridiculously small, but if we reject that where should we draw the limit instead?

etienne-lms: Ok, makes sense to me. Thanks a lot for the help.

etienne-lms:
squashed

jenswi-linaro (commented):

You may want to rephrase this part in the second commit message a bit:
*The rational is that the initial pool min size is not straight forward to compute
seen bget implementation and we can expect at least 1kB are provided
which is far enough and yet reasonable.*

etienne-lms:
Do you mean removing the ROUNDUP reference (i'll do) or also rephrasing the sentence you mentioned?
It looks clear to me but it is not specifically intended to me ;)

jenswi-linaro:
After reading the commit message a few more times I can make sense of it, but perhaps it will be easier to read as (well, at least in my mind ;-) ):
```
Requires at least 1 kB for the initial malloc memory pool. The rational is
that the initial pool min size is not straightforward to compute as it
depends on the internals of the BGET implementation. However, with a requirement
of at least 1 kB in initial memory pool we'll have good margin while still
being reasonable.

Adds a build-time test in user_ta_header.c to ensure TA heap size do
match this constraint.
```

etienne-lms:
commit log updated and review tag applied

PR #4857, opened 2021-09-14
core: arm: fix test on CFG_CC_OPTIMIZE_FOR_SIZE value

etienne-lms:
Makefile should test disabled CFG_CC_OPTIMIZE_FOR_SIZE against not
being 'y' instead of being 'n'.

### DIFF
--- a/core/arch/arm/arm.mk
+++ b/core/arch/arm/arm.mk
@@ -154,10 +154,6 @@ ifeq ($(DEBUG),1)
 $(call force,CFG_CC_OPT_LEVEL,0)
 $(call force,CFG_DEBUG_INFO,y)
 endif
-ifeq ($(CFG_CC_OPTIMIZE_FOR_SIZE),n)
-# For backwards compatibility
-$(call force,CFG_CC_OPT_LEVEL,0)
-endif
 
 # Optimize for size by default, usually gives good performance too
 CFG_CC_OPT_LEVEL ?= s
--- a/core/arch/arm/arm.mk
+++ b/core/arch/arm/arm.mk
@@ -149,17 +149,8 @@ arm64-platform-cflags-no-hard-float ?= -mgeneral-regs-only
 arm64-platform-cflags-hard-float ?=
 arm64-platform-cflags-generic := -mstrict-align $(call cc-option,-mno-outline-atomics,)
 
-ifeq ($(DEBUG),1)
-# For backwards compatibility
-$(call force,CFG_CC_OPT_LEVEL,0)
-$(call force,CFG_DEBUG_INFO,y)
-endif
-
-# Optimize for size by default, usually gives good performance too
-CFG_CC_OPT_LEVEL ?= s
 platform-cflags-optimization ?= -O$(CFG_CC_OPT_LEVEL)
 
-CFG_DEBUG_INFO ?= y
 ifeq ($(CFG_DEBUG_INFO),y)
 platform-cflags-debug-info ?= -g3
 platform-aflags-debug-info ?= -g
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -40,6 +40,18 @@ PYTHON3 ?= python3
 
 # Define DEBUG=1 to compile without optimization (forces -O0)
 # DEBUG=1
+ifeq ($(DEBUG),1)
+# For backwards compatibility
+$(call force,CFG_CC_OPT_LEVEL,0)
+$(call force,CFG_DEBUG_INFO,y)
+endif
+
+# CFG_CC_OPT_LEVEL sets compiler optimization level passed with -O directive.
+# Optimize for size by default, usually gives good performance too.
+CFG_CC_OPT_LEVEL ?= s
+
+# Enabling CFG_DEBUG_INFO makes debug information embedded in core.
+CFG_DEBUG_INFO ?= y
 
 # If y, enable debug features of the TEE core (assertions and lock checks
 # are enabled, panic and assert messages are more verbose, data and prefetch


jforissier (approved):

jenswi-linaro:

etienne-lms:
CI failure is... surprising. Well, CI is right is that indeed, with this change `make CFG_WITH_USER_TA=n` fails.
The strange thing is why does it fail?

The issue is not related to this P-R.
Before this change `make CFG_WITH_USER_TA=n CFG_CC_OPT_LEVEL=0` already failed :confused:.
Building optee_os alone:
```
$ make CFG_WITH_USER_TA=n CFG_CC_OPT_LEVEL=0
(...)
arm-none-linux-gnueabihf-ld.bfd: out/arm-plat-vexpress/core/pta/system.o: in function `system_dlopen':
core/pta/system.c:227: undefined reference to `ldelf_dlopen'
arm-none-linux-gnueabihf-ld.bfd: out/arm-plat-vexpress/core/pta/system.o: in function `system_dlsym':
core/pta/system.c:260: undefined reference to `ldelf_dlsym'
core/arch/arm/kernel/link.mk:44: recipe for target 'out/arm-plat-vexpress/core/all_objs.o' failed 
```

Indeed with `CFG_SYSTEM_PTA=y` (default value) and `CFG_WITH_USER_TA=n`, _core/pta/system.c_ is built/linked but not _core/arch/arm/kernel/ldelf_loader.c_. Link stage should fail since the former relies on `ldelf_dlopen()` and `ldelf_dlsym()`, implemented in _ldelf_loader.c_. However build does not fail when building with `-Os`, and fails when building with `-O0`. Why?

The issue rises with this P-R because `CFG_CC_OPTIMIZE_FOR_SIZE` has no default value, hence default (qemu/armv7) build sets `CFG_CC_OPT_LEVEL=s` from [_arm.mk_'s instructions](https://github.com/OP-TEE/optee_os/blob/3.14.0/core/arch/arm/arm.mk#L157-L164) whereas with this P-R, now we default to `CFG_CC_OPT_LEVEL=0` and the strange (IMHO) linkage behavior described above.

I see 3 outcomes here:
* There is an issue when building `CFG_WITH_USER_TA=n` and `CFG_SYSTEM_PTA=y` (that is hidden by the `-O` compiler directive).
This can be fixed by forcing `CFG_SYSTEM_PTA=n` when `CFG_WITH_USER_TA=n`.
* This P-R changes the default build config for all platforms: `CFG_CC_OPT_LEVEL` would defaults to `0` instead of `s`. Is it expected? I don't think so. That was not my intention.
I propose to add default `CFG_CC_OPT_LEVEL ?= s` in _mk/config.mk_.
(Could set a default value for `CFG_CC_OPTIMIZE_FOR_SIZE` but this switch seems a legacy almost deprecated config switch: maybe remove it?)
* I don't understand the linker behavior. The 2 changes proposed right above will hide the issue but something is weird here and should be clarified.

jforissier:
> CI failure is... surprising. Well, CI is right is that indeed, with this change `make CFG_WITH_USER_TA=n` fails.
> The strange thing is why does it fail?
> 
> The issue is not related to this P-R.
> Before this change `make CFG_WITH_USER_TA=n CFG_CC_OPT_LEVEL=0` already failed confused.
> Building optee_os alone:
> 
> ```
> $ make CFG_WITH_USER_TA=n CFG_CC_OPT_LEVEL=0
> (...)
> arm-none-linux-gnueabihf-ld.bfd: out/arm-plat-vexpress/core/pta/system.o: in function `system_dlopen':
> core/pta/system.c:227: undefined reference to `ldelf_dlopen'
> arm-none-linux-gnueabihf-ld.bfd: out/arm-plat-vexpress/core/pta/system.o: in function `system_dlsym':
> core/pta/system.c:260: undefined reference to `ldelf_dlsym'
> core/arch/arm/kernel/link.mk:44: recipe for target 'out/arm-plat-vexpress/core/all_objs.o' failed 
> ```
> 
> Indeed with `CFG_SYSTEM_PTA=y` (default value) and `CFG_WITH_USER_TA=n`, _core/pta/system.c_ is built/linked but not _core/arch/arm/kernel/ldelf_loader.c_. 

Clearly, `system_dlopen()` and `system_dlsym()` should return `TEE_ERROR_NOT_SUPPORTED` when `CFG_WITH_USER_TA != y` (but perhaps a more drastic change is required, see below).

> Link stage should fail since the former relies on `ldelf_dlopen()` and `ldelf_dlsym()`, implemented in _ldelf_loader.c_. However build does not fail when building with `-Os`, and fails when building with `-O0`. Why?

Because when `CFG_WITH_USER_TA != y` then `uctx` is always NULL here:
```C
        struct user_mode_ctx *uctx = to_user_mode_ctx(s->ctx);
```
With `-Os` the compiler eliminates the dead code. With -O0` it doesn't. You can check the disassembly of the `invoke_command()` function with `arm-linux-gnueabihf-objdump -d out/arm-plat-vexpress/core/pta/system.o` in both cases.

`to_user_mode_ctx()` seems wrong here, shouldn't it be possible to able to call the system PTA from a PTA for instance? Or are we considering that kernel mode services already have direct access to core functions so they should use them instead?

> 
> The issue rises with this P-R because `CFG_CC_OPTIMIZE_FOR_SIZE` has no default value, hence default (qemu/armv7) build sets `CFG_CC_OPT_LEVEL=s` from [_arm.mk_'s instructions](https://github.com/OP-TEE/optee_os/blob/3.14.0/core/arch/arm/arm.mk#L157-L164) whereas with this P-R, now we default to `CFG_CC_OPT_LEVEL=0` and the strange (IMHO) linkage behavior described above.
> 
> I see 3 outcomes here:
> 
>     * There is an issue when building `CFG_WITH_USER_TA=n` and `CFG_SYSTEM_PTA=y` (that is hidden by the `-O` compiler directive).
>       This can be fixed by forcing `CFG_SYSTEM_PTA=n` when `CFG_WITH_USER_TA=n`.

I am fine with that solution if we consider the system PTA cannot be called from another PTA. But if possible I would prefer allowing that. Or is it an unreasonable use case?

> 
>     * This P-R changes the default build config for all platforms: `CFG_CC_OPT_LEVEL` would defaults to `0` instead of `s`. Is it expected? I don't think so. That was not my intention.
>       I propose to add default `CFG_CC_OPT_LEVEL ?= s` in _mk/config.mk_.

Sounds good.

>       (Could set a default value for `CFG_CC_OPTIMIZE_FOR_SIZE` but this switch seems a legacy almost deprecated config switch: maybe remove it?)

Yes this one is legacy, I am not opposed to removing it.

>     * I don't understand the linker behavior. The 2 changes proposed right above will hide the issue but something is weird here and should be clarified.

I hope my comment above clarifies it ;)

jenswi-linaro:
> `to_user_mode_ctx()` seems wrong here, shouldn't it be possible to able to call the system PTA from a PTA for instance? Or are we considering that kernel mode services already have direct access to core functions so they should use them instead?

Right, there's no need to be able to call PTAs from within OP-TEE core.

etienne-lms:
> Right, there's no need to be able to call PTAs from within OP-TEE core.

I agree.

> > I don't understand the linker behavior. The 2 changes proposed right above will hide the issue but something is weird here and should be clarified.
> 
> I hope my comment above clarifies it ;)

It does. Thanks for the investigation and sharing the details.

etienne-lms:
> This can be fixed by forcing CFG_SYSTEM_PTA=n when CFG_WITH_USER_TA=n.

See https://github.com/OP-TEE/optee_os/pull/4860.

jforissier:
> Remove deprecated CFG_CC_OPTIMIZE_FOR_SIZE 

That should be one commit...

> and move CFG_DEBUG_INFO and CFG_CC_OPT_LEVEL default value and description to in/config.mk.

...and that should be another one. Typo: `mk/config.mk`.

Then:

etienne-lms:
updated.
i removed the review tags.

jforissier (approved):

etienne-lms:
R-b tag applied.

jforissier:
Ping @jenswi-linaro since you initially reviewed and this changed totally ;)

jenswi-linaro:

etienne-lms:
R-b tag applied

etienne-lms:
rebased (to restart CI)

PR #4871, opened 2021-09-20
Add soc_is_imx8m*() functions and digprog driver rework

clementfaure:
Hello,

This PR adds the soc_is_imx8m*() function for mscale platforms. These will be used in future pull-requests.
It also introduces code rework of the digprog register reading for having a common implementation across all imx boards.

Thanks,
Clement


core/arch/arm/plat-imx/conf.mk
```diff
@@ -150,24 +150,28 @@ $(call force,CFG_NXP_CAAM,n)
 include core/arch/arm/cpu/cortex-a7.mk
 else ifneq (,$(filter $(PLATFORM_FLAVOR),$(mx8mq-flavorlist)))
 $(call force,CFG_MX8MQ,y)
+$(call force,CFG_MX8M,y)
```
jforissier, line 153: Since this variable is derived from other configuration values (in this case, `PLATFORM`/`PLATFORM_FLAVOR`), it should preferably be prefixed with an underscore: `_CFG_MX8M`. In this case, no need for `$(call force,...)` and `_CFG_MX8M := y` should be used instead (since internal variables are not supposed to be set from outside the makefiles). Of course other variables such as `CFG_MX8MQ`, `CFG_MX8MM` etc. would need the same thing. It was a (small) mistake to introduce them without the underscore.

Would you agree to rename all the `CFG_MX*` to `_CFG_MX*` before introducing the new one as `_CFG_MX8M`? As I said, not functionally required, just preferred.

clementfaure: That's a lot of files to modify ..  To be honest, I wasn't aware of this coding rule. Is it a Linux Kernel rule?

If you don't mind I would rather align ```CFG_MX8M``` on what has already been done with ```CFG_MX6``` and ```CFG_MX7```. However, I will take notes and make sure that future internal variables start with an underscore as you mentionned.

I like having ```$(call force,...)``` for this kind of variable because it makes the variable immutable and prevents any mistake. I hope that makes sense.

clementfaure:
Force-push:
* Fix compilation

core/arch/arm/plat-imx/conf.mk
jforissier: > That's a lot of files to modify .. To be honest, I wasn't aware of this coding rule. Is it a Linux Kernel rule?

No, it is completely OP-TEE specific, we do have a few _CFG already mostly for crypto. Unfortunately this is only documented in my head :wink: so no surprise the "rule" is not always followed... The main reason why I introduced this underscore at the time is to make them stand out, meaning they are not supposed to be set by the user. Another small advantage is to reduce the number of variables in `$(out-dir)/conf.mk`. Take a look again at the internal variables (`git grep _CFG_` and in particular `_CFG_CORE_LTC*`) and imagine them without the leading underscore, I think that would be confusing.

> I like having $(call force,...) for this kind of variable because it makes the variable immutable and prevents any mistake.

Me too, but it should be safe to assume any _CFG is reserved and should not be set from outside (command line or environment variable). In this case `:=` is the same.

If this is a bit controversial we can ignore this debate for now.

clementfaure: > The main reason why I introduced this underscore at the time is to make them stand out, meaning they are not supposed to be set by the user. Another small advantage is to reduce the number of variables in $(out-dir)/conf.mk. Take a look again at the internal variables (git grep _CFG_ and in particular _CFG_CORE_LTC*) and imagine them without the leading underscore, I think that would be confusing.

I see what you mean indeed ; that would have been confusing.

> Me too, but it should be safe to assume any _CFG is reserved and should not be set from outside (command line or environment variable). In this case := is the same.

I find ```$(call force,...)``` a bit more bullet proof IMO since there is nothing to assume from the user/developer. But I totally get your point on having designating internal values with having it started with an underscore. This is a common practice in many places.

I keep `CFG_MX8M` as it is now for consistency and we will try to comply with this rule for future PRs.

jforissier:

clementfaure:
Force-push
 * Apply tags (Thanks @jforissier )
 * Add appropriate "Fixes: ..." to commit message `core: imx: rework digprog driver `

PR #4872, opened 2021-09-20
core: dt: fix missing const attribute on fdt reference

etienne-lms:
The standard FDT reference in libfdt and friends is a const void *. Fix
few function prototypes that miss the const attribute.


jforissier (approved):

PR #4873, opened 2021-09-20
zynqmp: add minimal firmware driver

igoropaniuk:


core/arch/arm/plat-zynqmp/conf.mk
```diff
@@ -27,3 +27,7 @@ CFG_SHMEM_SIZE   ?= 0x10000000
 
 CFG_WITH_STATS ?= y
 CFG_CRYPTO_WITH_CE ?= y
+
+ifeq ($(CFG_ARM64_core),y)
+CFG_ZYNQMP_MMIO ?= y
+endif
```
jforissier, line 33: `CFG_ZYNQMP_MMIO ?= $(CFG_ARM64_core)`

core/drivers/zynqmp_mmio.c
```diff
@@ -0,0 +1,62 @@
...
+#include <tee_api_types.h>
+
+#define ZYNQMP_MMIO_READ	0xC2000014
+#define ZYNQMP_MMIO_WRITE	0xC2000013
+
+static uint32_t zynqmp_sip_call(uint32_t pm_api_id, uint32_t arg0,
+				uint32_t arg1, uint32_t arg2, uint32_t arg3,
+				uint32_t *payload)
+{
+	struct smccc_res res;
```
jforissier, line 18: `= { };`

core/drivers/zynqmp_mmio.c
```diff
@@ -0,0 +1,62 @@
...
+		*payload = res.a0 >> 32;
+
+	return res.a0;
+}
+
+TEE_Result zynqmp_mmio_write(const uint32_t address,
+			     const uint32_t mask,
+			     const uint32_t value)
+{
+	uint32_t ret;
```
jforissier, line 35: `= 0;`

core/drivers/zynqmp_mmio.c
```diff
@@ -0,0 +1,62 @@
...
+		return TEE_ERROR_GENERIC;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result zynqmp_mmio_read(const uint32_t address,
+			    uint32_t *value)
+{
+	uint32_t ret;
+	uint32_t payload;
```
jforissier, line 50: `= 0;`

jforissier (commented):
For commit "zynqmp: add simple mmio driver":
In the commit subject: s/mmio/MMIO/. In the description: s/mmio/MMIO/, s/zynqmp/ZynqMP/ (`zynqmp:` lowercase is OK as a platform prefix however).

More comments below, then:

igoropaniuk:
@ricardosalveti @ldts could you please also take a look
Thanks

jforissier:
```
 ZynqMP: add minimal firmware driver

Introduce minimal firmware driver with mmio_write/read and
efuse_access routines for zynqmp platform.
```

You changed the wrong `zynqmp`, introduced function names with missing parentheses and also it is best not to truncate function names for git grepability, I mean: `mmio_read()/mmio_write()` not `mmio_read()/write()`.
Please use:

```
zynqmp: add minimal firmware driver

Introduce minimal firmware driver with mmio_write()/mmio_read()
and efuse_access() routines for ZynqMP platform.
```

igoropaniuk:
@jforissier fixed, thanks Jerome

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+TEE_Result zynqmp_mmio_read(const uint32_t address,
+			    uint32_t *value);
+
+/*
+ * Stores all required details to
+ * read/write efuse memory.
+ * @src:	address of the buffer to store the data to be write/read
+ * @size:	number of words to be read/write
+ * @offset:	offset to be read/write
```
jforissier, line 36: `number of words to read or write`
What kind of words? 32-bit words?

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,57 @@
...
+ * Return a TEE_Result compliant status
+ */
+TEE_Result zynqmp_mmio_read(const uint32_t address,
+			    uint32_t *value);
+
+/*
+ * Stores all required details to
+ * read/write efuse memory.
+ * @src:	address of the buffer to store the data to be write/read
+ * @size:	number of words to be read/write
```
jforissier, line 35: `to be read or written`

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,57 @@
...
+ * @value: read value to store
+ * Return a TEE_Result compliant status
+ */
+TEE_Result zynqmp_mmio_read(const uint32_t address,
+			    uint32_t *value);
+
+/*
+ * Stores all required details to
+ * read/write efuse memory.
+ * @src:	address of the buffer to store the data to be write/read
```
jforissier, line 34: `to be read or written`

jforissier (commented):
More comments actually...

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/*
+ * Provide access to efuse memory
+ * @address: Address of the efuse params structure
+ * @value: returned output value
+ * Return a TEE_Result compliant status
+ */
+TEE_Result zynqmp_efuse_access(const uint64_t address,
+			       uint32_t *value);
```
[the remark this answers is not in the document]
jforissier: Oh OK. Then "access" is fine. Yes `struct xilinx_efuse *` is much better. I suggest defining macros for `flag` (`ZYNQMP_EFUSE_READ`/`ZYNQMP_EFUSE_WRITE` for instance), and for `pufuserfuse` too. `src` is misleading when reading from eFuse isn't it? (it is an output buffer in this case, right?), `buf` would be preferable. The value to be put in `size` and `offset` isn't clear to me (in which units are `size` and `offset`?).

core/arch/arm/kernel/asm-defines.c
```diff
@@ -127,4 +128,10 @@ DEFINES
 	       offsetof(struct boot_embdata, reloc_offset));
 	DEFINE(BOOT_EMBDATA_RELOC_LEN,
 	       offsetof(struct boot_embdata, reloc_len));
+#ifdef ARM64
```
etienne-lms, line 131: needs a specific section: inline comment above (`/* struct boot_embdata */`) is not accurate.

Can remove the line break, each fit in a single line.

core/arch/arm/kernel/sub.mk
```diff
@@ -22,6 +22,8 @@ srcs-$(CFG_PL310) += tee_l2cc_mutex.c
 srcs-$(CFG_ARM32_core) += thread_a32.S
 srcs-$(CFG_ARM64_core) += thread_a64.S
 srcs-y += thread.c
+
```
etienne-lms, line 25: can remove this empty line

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,83 @@
...
+				uint32_t *payload)
+{
+	struct smccc_res res = { };
+
+	smccc_smc(pm_api_id,
+		  ((uint64_t)arg1 << 32) | arg0,
+		  ((uint64_t)arg3 << 32) | arg2,
+		  0, 0, 0, 0, 0, &res);
+
+	if (payload)
```
etienne-lms, line 26: `if (payload && !res.a0)` ?

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,83 @@
...
+{
+	uint32_t ret = 0;
+	uint32_t payload = 0;
+
+	if (!value)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	ret = zynqmp_sip_call(ZYNQMP_MMIO_READ, address,
+			      0, 0, 0, &payload);
+	*value = payload;
```
etienne-lms, line 58: save output `payload` only upon a successful call.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,83 @@
...
+	uint32_t payload = 0;
+
+	if (!value)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	ret = zynqmp_sip_call(ZYNQMP_EFUSE_ACCESS,
+			      (uint32_t)(address >> 32),
+			      (uint32_t)(address),
+			      0, 0, &payload);
+	*value = payload;
```
etienne-lms, line 78: ditto

ldts:
I havent been able to read the eFUSEs with this code though: the call succeeds but it returns zeros.

I am looking at the usage in the Xilinx kernel tree:
https://github.com/Xilinx/linux-xlnx/blob/master/drivers/nvmem/zynqmp_nvmem.c

do we need to allocate some kind of special memory to be shared with the trusted firmware?

ldts:
maybe we need to recompile the pmufw with the ENABLE_EFUSE_ACCESS macro.
https://xilinx-wiki.atlassian.net/wiki/spaces/A/pages/18841682/Solution+ZynqMP+SoC+revision+read+mechanism

will check this out.

ldts:
@jforissier @jenswi-linaro  I did confirm that the issue we have is related to the memory being shared with the PMUFW.

On the read eFUSE call, the TEE needs to share two memory buffers with the PMUFW via TF-A ( a structure containing the command and a pointer to a buffer to be filled or read by the PMUFW)
 
So
1) TEE will pass a structure with the request to the TF-A
2) TF-A will tunnel the structure verbatim to the PMUFW

The structure  defines the eFUSE request -read/write, offset in the efuse map, size to read etc- and contains a pointer to the buffer to be written to as described in the PR

struct xilinx_efuse {
	uint64_t src; <--------pointer to hold the read or write data
	uint32_t size;
	uint32_t offset; <-- the efuse to act upon
	uint32_t flag;
	uint32_t pufuserfuse;
}

Is there a way to allocate from OP-TEE so the memory is visible to the PMUFW? 
This is early after boot so neither Linux nor U-Booot are executing (at that point SPL is just loading the TEE).

ldts:
@igoropaniuk  so two things:
1) all buffers must be cache aligned
2) caches need to be flushed/invalidated before handling the output

ldts:
I think the second commit adds no value to zynqmp and we should drop it (mmio reads are not necesary since optee can just  io_read/io_write to all registers)
then we should have a zynqmp_read_efuse function that handles the TEE_CACHEFLUSH operations and just calls __sip_call(..). reading the efuse ATM is only needed to generate the HUK but if it is required at a different level - ie to secure the board writing the PPK and RSA_EN efuses- then we can have those operations in a PTA IMO

ldts:
```
struct xilinx_efuse {
		uint64_t src;
		uint32_t size;
		uint32_t offset;
		uint32_t flag;
		uint32_t pufuserfuse;
	} *efuse = mem_alloc(sizeof(struct xilinx_efuse),
			      MEM_TYPE_ALIGN | MEM_TYPE_ZEROED);
	uint8_t *buffer = mem_alloc(12, MEM_TYPE_ALIGN | MEM_TYPE_ZEROED);

	efuse->src = (uint64_t) virt_to_phys(buffer);
	efuse->offset = 0x0c;
	efuse->size = 3;

	cache_operation(TEE_CACHEFLUSH, (void *) efuse->src, 12);
	cache_operation(TEE_CACHEFLUSH, (void *) efuse, sizeof(*efuse));
	efuse_read(virt_to_phys(efuse), &ret);
        DHEXDUMP(buffer, 12);
```

igoropaniuk:
I've done a complete overhaul of PR after discussion with Jorge and his help with debugging cache alignment issues:
1. Dropped `mmio_read()`/`mmio_write()` functions, as we don't plan to use them
2. Added allocation of mem aligned buffers inside `efuse_op()` implementation.
3. Added separate `efuse_read()` and `efuse_write()` functions with simplified formal params instead of explicitly filling xilinx_efuse structure by a callee.
4. Addressed existing comments

@jforissier @etienne-lms @ldts @ricardosalveti

ldts:
hi @igoropaniuk  efuse_write needs to have the same endianess than the CSU to avoid bricking the board when programming the PPK. please let's make sure that this requirement is met before merging.

igoropaniuk:
I've synced today with @ldts regarding endianess when using `zynqmp_efuse_write()`, and think there is misunderstanding here.

Taking into account that we provide buffer with whatever data it contains and size of the buffer, IMHO it's responsibility of caller of this function to provide input in the correct endianess.

Lets take for instance how it's done in Xilinx FSBL https://github.com/Xilinx/embeddedsw/blob/master/lib/sw_apps/zynqmp_fsbl/src/xfsbl_authentication.c#L437, yes, we For PPK we're expecting 32 bit unsigned integers, so we read every value moving by offset 4 and then for each 4 bytes use Xil_In32() for retrieving the value in correct endianess


Why not to do that the same way here then? We have a generic function for reading a buffer with whatever values (and we should not care if it's a sequence of u8 values or it's 32bit integers in little endian byte order) from efuse and then read/convert value into the proper format (endianess) in that caller function?

igoropaniuk:
I've synced today with @ldts regarding endianess when using `zynqmp_efuse_write()`, and I think there is misunderstanding here.

Taking into account that we provide buffer with whatever data it contains and size of the buffer as params, IMHO it's responsibility of a caller function to provide input in the correct endianess.

Lets take for instance how it's done in Xilinx FSBL https://github.com/Xilinx/embeddedsw/blob/master/lib/sw_apps/zynqmp_fsbl/src/xfsbl_authentication.c#L437. Yes, we For PPK we're expecting 32 bit unsigned integers, so we read every value moving by offset 4 and then for each 4 bytes use Xil_In32() for retrieving the value in correct endianess, but all that is done in function for specifically reading PPK hash.

Why not to do that the same way here then? We have a generic function for retrievin buffer with whatever values (and we should not care if it's a sequence of u8 values or it's 32bit integers in little endian byte order) from efuse and then read/convert value into the proper format (endianess) in that caller function?
For example:
```
void read_ppk(u32 *ppk_hash)
{
       u8 buf[48]; /* 4 * 12 */
       s32 num;
       u32 *data_read = ppk_hash;

       zynqmp_efuse_read (buf, 48, PPK_OFFSET, puf);
       for (num = 0; num < 12; num++) {
             /* 
              *or in case we have swapped byte order , then we should do 
              * (or use whatever similar function/macro optee provides) 
              *  *data_read = __bswap_32(buf[num * 4]) 
              */   
             *data_read = buf[num * 4]; 
           
             data_read++;            
       }
}
```

@ldts 
If that doesn't work for you, I'd suggest to drop that function for fusing efuses for now, and introduce all needed calls for PPK retrieval in a separate PR.

ldts:
um no, I am not sure where the misunderstanding lies: the interface we are providing for reading and writing efuses has to obey some byte order requirement (imposed by the xilskey library executing in the PMUFW)  or we risk bricking the board when the ROM fails to validate the signatures.

It is not clear to me what those requirements are at the moment, hence why I would not merge this PR: ie, what byte order does the OP-TEE read/write eFUSE interface accept? 

A clear use case is writing the PPK eFuses which will be used by the CSU to validate the bootloader signature: does your interface accept BE or LE? what should we use? and on read, does it return BE or LE?

All I am saying is that this need to be clarified and explicitly defined or the interface is incomplete.

igoropaniuk:
@ldts ok, I've dropped `zynqmp_efuse_write()`

lets address that in a separate PR, as all details about writing PPK efuses was out of initial scope of this PR.

igoropaniuk:
@ldts @jforissier @etienne-lms could you please take another look and leave you comments/suggestions if you have any?
As I've done a complete overhaul of the PR, some of existing comments are obsolete/not relevant anymore, however I've tried to address almost everything that was mentioned

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+		tee_ret = TEE_ERROR_GENERIC;
+	}
+
+	if (op == EFUSE_READ)
+		memcpy(buf, buf_aligned, sz);
+
+	free(buf_aligned);
+
+alloc_error:
+	free(efuse);
```
jforissier, line 135: Swap `free(buf_aligned);` and `free(efuse);`

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+
+	return ptr;
+}
+
+static TEE_Result efuse_op(enum efuse_op op, uint8_t *buf, size_t sz,
+			   uint32_t efuse_offset, uint32_t puf)
+{
+	paddr_t addr = 0;
+	void *buf_aligned = NULL;
+	uint32_t ret = 0;
```
jforissier, line 86: `res`

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+	TEE_Result tee_ret = TEE_SUCCESS;
```
jforissier, line 87: `ret`

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+
+	free(buf_aligned);
+
+alloc_error:
+	free(efuse);
+
+	return tee_ret;
+}
+
+TEE_Result zynqmp_efuse_read(uint8_t *buf, size_t sz, uint32_t efuse_offset,
```
jforissier, line 140: `size_t efuse_offset`?

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,26 @@
...
+/*
+ * Read efuse memory
+ * @buf: Buffer, where efuse date will be stored
+ * @sz: buffer size in bytes
+ * @offset: Offset of efuse register
+ * @puf: Is EFUSE PUF: ZYNQMP_NONPUF_EFUSE/ZYNQMP_PUF_EFUSE
+ * Return a TEE_Result compliant status
+ */
+TEE_Result zynqmp_efuse_read(uint8_t *buf, size_t sz, uint32_t efuse_offset,
+			     uint32_t puf);
```
jforissier, line 25: `bool puf_efuse`?

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+	efuse->src = (uint64_t)virt_to_phys(buf_aligned);
+	efuse->size = sz / sizeof(uint32_t);
+	efuse->offset = efuse_offset;
+	efuse->flag = op;
+	efuse->pufuserfuse = puf;
+
+	cache_operation(TEE_CACHEFLUSH, (void *)buf_aligned, sz);
+	cache_operation(TEE_CACHEFLUSH, (void *)efuse, sizeof(*efuse));
+
+	addr = virt_to_phys((void *)efuse);
```
jforissier, line 114: I think the casts to `void *` are not needed

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+	if (ret) {
+		if (ret == EFUSE_NOT_ENABLED)
+			EMSG("Efuse access is not enabled");
+		else
+			EMSG("Error in efuse access %x", ret);
+
+		tee_ret = TEE_ERROR_GENERIC;
+	}
+
+	if (op == EFUSE_READ)
```
jforissier, line 129: No need to copy if `ret` is nonzero, right?

core/arch/arm/kernel/smccc_call_a64.S
```diff
@@ -0,0 +1,25 @@
...
+ *		  unsigned long a6, unsigned long a7, struct smccc_res *res)
+ * a0~a7: input arguments
+ * res: output arguments from register 0-3
+ */
+FUNC smccc_smc , :
+	smc	#0
+	ldr	x4, [sp]
+	/* Put results to res struct */
+	stp	x0, x1, [x4, #SMCCC_RES_X0_OFFS]
+	stp	x2, x3, [x4, #SMCCC_RES_X2_OFFS]
```
jforissier, line 23: Can't `sp` be used directly instead of `x4`?

ldts:
> @ldts ok, I've dropped `zynqmp_efuse_write()`
> 
> lets address that in a separate PR, as all details about writing PPK efuses was out of initial scope of this PR.

but this is not about PPK efuses: this is  about making sure the zynqmp_efuse_write is defined properly and for that, we need to let the user know the format to use in the buffer passed in the interface (will be the same for read...)

igoropaniuk:
> > @ldts ok, I've dropped zynqmp_efuse_write()
> > 
> > lets address that in a separate PR, as all details about writing PPK efuses was out of initial scope of this PR.
> 
> but this is not about PPK efuses: this is about making sure the zynqmp_efuse_write is defined properly and for that, we need to let the user know the format to use in the buffer passed in the interface (will be the same for read...)

sure, I do agree that all endianess-specific handling should be done, but not in `zynqmp_efuse_read()`/`zynqmp_efuse_write()` rather in more high-level functions (that will put all values in proper byte order and then call `zynqmp_efuse_write()`, or vice-versa, will read buffer with data by calling `zynqmp_efuse_read()` and then extract the values from that buffer in proper byte order.)
 
Let's just take as an example https://github.com/Xilinx/linux-xlnx/blob/master/drivers/nvmem/zynqmp_nvmem.c, I don't see there any handling either.

ldts:
@igoropaniuk seems we are talking past each other. let me add it here for clarity (notice the explanation in the buffer field)
```
/*
 * Read efuse memory
 * @buf: Buffer, where efuse data will be stored. The data will be presented in BE/LE? format.
 * @sz: buffer size in bytes
 * @offset: Offset of efuse register
 * @puf: Is EFUSE PUF: ZYNQMP_NONPUF_EFUSE/ZYNQMP_PUF_EFUSE
 * Return a TEE_Result compliant status
 */
TEE_Result zynqmp_efuse_read(uint8_t *buf, size_t sz, uint32_t efuse_offset,
			     uint32_t puf);
```

the same should be done for the write function.
IMO that is all that is missing for users to know how to call the function.

core/drivers/zynqmp_pm.c
igoropaniuk: there is one ret on line above

core/drivers/zynqmp_pm.c
igoropaniuk: the problem here is that PMU processor (which is 32bit) is expecting this value to be 32bit  (this is why it's defined strictly as uint32_t in `xilinx_efuse` structure, whose values are read by PMU FW)

core/drivers/zynqmp_pm.c
igoropaniuk: ah, sorry, missed the comment above :)

core/arch/arm/kernel/smccc_call_a64.S
igoropaniuk: you mean in both lines?
well, not really, as `sp` value will change after pushing `x0`,`x1` values to stack on line 22.

if you mean on line 23 only, IMO I see no value for that

igoropaniuk: I'd prefer to do that the same way as it's done in Linux kernel source code
https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/smccc-call.S#L41

IMO Explicitly specifying the offsets (`SMCCC_RES_X0_OFFS` etc) makes that piece of code more readable.

igoropaniuk:
@ldts comment is addressed

@jforissier addressed all comments except two where I left replies

core/drivers/zynqmp_pm.c
jforissier: Fine then, thanks.

core/arch/arm/kernel/smccc_call_a64.S
jforissier: Actually we have a macro for that:
```
#include <arm64_macros.S>

store_xregs sp, SMCCC_RES_X0_OFFS, 0, 3
```
Then you can drop the definition of `SMCCC_RES_X2_OFFS`. BTW I suggest also dropping the `_OFFS` suffix, it is more or less implicit (check the names of other defines in `out/arm/core/include/generated/asm-defines.h`).

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,39 @@
...
+
+#define ZYNQMP_DNA_EFUSE_OFFSET		0xC
+
+/*
+ * Read efuse memory
+ * @buf: buffer, where efuse date will be stored. The data is returned
+ *       in LE byte order
+ * @sz: buffer size in bytes
+ * @offset: offset of efuse register
+ * @puf: is EFUSE PUF: ZYNQMP_NONPUF_EFUSE/ZYNQMP_PUF_EFUSE
```
jforissier, line 22: `s/: ZYNQMP_NONPUF_EFUSE/ZYNQMP_PUF_EFUSE//`

core/include/drivers/zynqmp_pm.h
```diff
@@ -0,0 +1,39 @@
...
+TEE_Result zynqmp_efuse_read(uint8_t *buf, size_t sz, uint32_t efuse_offset,
+			     bool puf);
+
+/*
+ * Write efuse memory
+ * @buf: buffer, contains data that will be written to efuse. The data
+ *       should be provided in LE byte order
+ * @sz: buffer size in bytes
+ * @offset: offset of efuse register
+ * @puf: is EFUSE PUF: ZYNQMP_NONPUF_EFUSE/ZYNQMP_PUF_EFUSE
```
jforissier, line 34: `s/: ZYNQMP_NONPUF_EFUSE/ZYNQMP_PUF_EFUSE//`

core/arch/arm/kernel/smccc_call_a64.S
jforissier: Actually we have a macro for that:
```
#include <arm64_macros.S>

store_xregs sp, SMCCC_RES_X0_OFFS, 0, 3
```
Then you can drop the definition of `SMCCC_RES_X2_OFFS`. BTW I suggest also dropping the `_OFFS` suffix, it is more or less implicit (check the names of other defines in `out/arm/core/include/generated/asm-defines.h`).

jforissier (approved):
With the comments below:

ldts:
@igoropaniuk have you tested the write function?

core/arch/arm/kernel/smccc_call_a64.S
igoropaniuk: Interesting, but that doesn't work

```
diff --git a/core/arch/arm/kernel/asm-defines.c b/core/arch/arm/kernel/asm-defines.c
index 31c38244..b83c665b 100644
--- a/core/arch/arm/kernel/asm-defines.c
+++ b/core/arch/arm/kernel/asm-defines.c
@@ -125,7 +125,6 @@ DEFINES
               offsetof(struct boot_embdata, reloc_len));
 #ifdef ARM64
        /* struct smccc_res */
-       DEFINE(SMCCC_RES_X0_OFFS, offsetof(struct smccc_res, a0));
-       DEFINE(SMCCC_RES_X2_OFFS, offsetof(struct smccc_res, a2));
+       DEFINE(SMCCC_RES_X0, offsetof(struct smccc_res, a0));
 #endif
 }
diff --git a/core/arch/arm/kernel/smccc_call_a64.S b/core/arch/arm/kernel/smccc_call_a64.S
index 3b6604f9..351ec1c4 100644
--- a/core/arch/arm/kernel/smccc_call_a64.S
+++ b/core/arch/arm/kernel/smccc_call_a64.S
@@ -4,6 +4,7 @@
  */
 
 #include <arm.h>
+#include <arm64_macros.S>
 #include <asm.S>
 #include <generated/asm-defines.h>
 
@@ -18,7 +19,6 @@ FUNC smccc_smc , :
        smc     #0
        ldr     x4, [sp]
        /* Put results to res struct */
-       stp     x0, x1, [x4, #SMCCC_RES_X0_OFFS]
-       stp     x2, x3, [x4, #SMCCC_RES_X2_OFFS]
+       store_xregs sp, SMCCC_RES_X0, 0, 3
        ret
 END_FUNC smccc_smc

```

Maybe that's because of `sp` aligment restrictions?
> This mechanism is simple in principle: if the alignment restrictions of sp are inconvenient, just use another register as your stack pointer. General-purpose registers have no special alignment restrictions.

https://community.arm.com/developer/ip-products/processors/b/processors-ip-blog/posts/using-the-stack-in-aarch64-implementing-push-and-pop

igoropaniuk:
@ldts for obvious reasons I only checked that tf-a PM handler is called and as params are provided in correct order (as I don't want to spoil any fuses on my board)

core/arch/arm/kernel/smccc_call_a64.S
jforissier: What doesn't work? Compilation or runtime? I get this disassembly:
```
0000000000000000 <smccc_smc>:
smccc_smc():
/home/jerome/work/optee_repo_qemu_v8/optee_os/core/arch/arm/kernel/smccc_call_a64.S:19
   0:   d4000003        smc     #0x0
/home/jerome/work/optee_repo_qemu_v8/optee_os/core/arch/arm/kernel/smccc_call_a64.S:21
   4:   a90007e0        stp     x0, x1, [sp]
   8:   a9010fe2        stp     x2, x3, [sp, #16]
/home/jerome/work/optee_repo_qemu_v8/optee_os/core/arch/arm/kernel/smccc_call_a64.S:22
   c:   d65f03c0        ret
```
Looks pretty good, doesn't it?

igoropaniuk: @jforissier looks like this is the problem, as this works:
```
ldr     x4, [sp]                                                        
store_xregs x4, SMCCC_RES_X0, 0, 3  
```

jforissier: Oh, did I miss an indirection? ;-) My assembly skills are not as good as I'd like... LGTM

igoropaniuk: runtime,  it hangs when trying to store value using directly the stack pointer as a base address

igoropaniuk:
Pushed updated version with @jforissier tags applied

jforissier:
> @igoropaniuk have you tested the write function?
> @ldts for obvious reasons I only checked that tf-a PM handler is called and as params are provided in correct order (as I don't want to spoil any fuses on my board)

If there is any doubt that the write function might be buggy it can be left aside for the moment, I see no issue with that.

ldts:
so it is not the memory shared with TF-A or recompiling the PMUFW. The issue seems to be that the buffer shared with the PMUFW is not v

> > @igoropaniuk have you tested the write function?
> > @ldts for obvious reasons I only checked that tf-a PM handler is called and as params are provided in correct order (as I don't want to spoil any fuses on my board)
> 
> If there is any doubt that the write function might be buggy it can be left aside for the moment, I see no issue with that

well no, there is nothing obvious in that reasoning @igoropaniuk; the platform provides user fuses that can be used to validate those write operations. Really the whole reason for this PR was to enable my set of PRs so I expected working/tested interfaces so I dont risk bricking my board. But I can take over once everything is merged, not a problem on my end -as a matter of fact, the sooner it is merged probably the better

igoropaniuk:
>>> @igoropaniuk have you tested the write function?

>> @ldts for obvious reasons I only checked that tf-a PM handler is called and as params are provided in correct order (as I don't want to spoil any fuses on my board)

> If there is any doubt that the write function might be buggy it can be left aside for the moment, I see no issue with that.

Removed it.

ldts:
@jforissier I dont think the write function is buggy but I have had tonnes of issues with this platform over the past 5 months - a long list hence why IMO testing makes sense (ie, remember the read_efuse function missing the cache operations...and the call was landing on TF-A)

jforissier:
Fine with me, let's give @etienne-lms an opportunity to comment further perhaps. And @ldts you may give R-b/A-b tags if you want.

ldts:
should we use the IS_ALIGNED() macro to decide whether to allocate our own buffers or not? 

`#define IS_ALIGNED(x, a) (((x) & ((typeof(x))(a) - 1)) == 0)`

ldts:
interesting, memalign doesnt work on my 3.10 backport...it returns unaligned addresses. I'd be curious to see what you get.

igoropaniuk:
@ldts nice catch, I also confirm that in some cases it doesn't return a buffer with the correct alignment (first buffer is aligned, 
subsequent is not)

```
E/TC:0 0 alloc_aligned_zeroed:76 ### ZYNQMP cacheline size = 0x40, address allocated 0x7e07cf80
E/TC:0 0 alloc_aligned_zeroed:76 ### ZYNQMP cacheline size = 0x40, address allocated 0x7e07cf10
```

jforissier:
I could not reproduce the `memalign()` issue on QEMUv8 with the following test (patch applied to current master branch).
```
$ make -j10 CFG_ENABLE_EMBEDDED_TESTS=y run
...
$ xtest 1001
```
```diff
diff --git a/core/pta/tests/misc.c b/core/pta/tests/misc.c
index cd9e5ee7..aaaffcc6 100644
--- a/core/pta/tests/misc.c
+++ b/core/pta/tests/misc.c
@@ -17,7 +17,7 @@
  * #define LOG     DMSG_RAW
  * #define LOG(...)
  */
-#define LOG(...)
+#define LOG DMSG_RAW
 
 static int self_test_add_overflow(void)
 {
@@ -311,6 +311,8 @@ static int self_test_malloc(void)
 	int *p3 = NULL, *p4 = NULL;
 	bool r;
 	int ret = 0;
+	int i = 0;
+	void *p[10] = { };
 
 	LOG("malloc tests:");
 	LOG("  p1=%p  p2=%p  p3=%p  p4=%p",
@@ -414,6 +416,22 @@ static int self_test_malloc(void)
 	p3 = NULL;
 	p4 = NULL;
 
+
+	for (i = 0; i < 10; i++) {
+		LOG(" - p[%d] = memalign(64, 24)", i);
+		p[i] = memalign(64, 24);
+		LOG("   p[%d]=%p", i, p[i]);
+		r = (p[i] && !((vaddr_t)p[i] % 64));
+		if (!r)
+			ret = -1;
+	}
+	for (i = 0; i < 10; i++) {
+		LOG(" - free(p[%d])", i);
+		free(p[i]);
+	}
+	LOG("  => test %s", r ? "ok" : "FAILED");
+	LOG("");
+
 	/* test free(NULL) */
 	LOG("- free NULL");
 	free(NULL);
```

ldts:
thanks @jforissier; it seems we have then a backport issue on 3.10 and that the alignment was not required by the PMUFW to access the fuses on read (we only had to manage the cache). can we then remove the memalign calls from the PR please?

vesajaaskelainen:
What kind of access protections are in places for PMUFW's access? Eg. there is Linux driver in Xilinx tree that would allow writing and reading some efuses when enabled -- and that is probably not the effect you are after.

ldts:
> What kind of access protections are in places for PMUFW's access? Eg. there is Linux driver in Xilinx tree that would allow writing and reading some efuses when enabled -- and that is probably not the effect you are after.

from SPL at EL3 we can access the eFUSES directly; once SPL loads the PMUFW and boots U-boot proper and the rest of the binaries, accessing PMUFW is done via TF-A (the linux driver you are commenting on also does just that). Inspecting the xilskey library (and with some minimal efuse testing) it seems to always be the A57 the one writing into the TF-A buffers - no DMA involved...so I doubt memalign is required (merely the flushing on the cache upon reads)

vesajaaskelainen:
> > What kind of access protections are in places for PMUFW's access? Eg. there is Linux driver in Xilinx tree that would allow writing and reading some efuses when enabled -- and that is probably not the effect you are after.
> 
> from SPL at EL3 we can access the eFUSES directly; once SPL loads the PMUFW and boots U-boot proper and the rest of the binaries, accessing PMUFW is done via TF-A (the linux driver you are commenting on also does just that). Inspecting the xilskey library (and with some minimal efuse testing) it seems to always be the A57 the one writing into the TF-A buffers - no DMA involved...so I doubt memalign is required (merely the flushing on the cache upon reads)

Small correction to above -- ZynqMP has A53 and Versal has A72.

Xilinx kernel seems to do aligned memory allocation. Haven't studied the PMUFW that deep what kind of Microblaze accesses are being done to that memory. It could easily need to be 32bit aligned at least.

OK. So we need to have TZ to access the PMUFW interface. Good.

Now next question is it OK to allow also Linux to access this too at the same time? Because I believe now both can access this. 

I suppose there can be multiple users for TF-A monitor from multiple cores? If I recall correctly OP-TEE had some guarantees for TA's for access singularity but does TF-A has the same thing? Eg. do we now possibly get race on the access?

Upstream kernel only allows access to read SOC identifier:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/nvmem/zynqmp_nvmem.c?h=v5.15-rc3#n19

And for the reference here is the Xilinx kernel version:
https://github.com/Xilinx/linux-xlnx/blob/xlnx_rebase_v5.10/drivers/nvmem/zynqmp_nvmem.c#L52

Then the next question is what kind of access control is needed for eFuse access in different applications.

There are user efuses that depending on your selected security model could have multiple uses.

PUF has own area or it can be user area if you cannot use PUF due to firmware encryption needs.

So those could be freely used in particular product -- but I suppose there needs to be some access gates that if someone decides to use some of free efuse for secretive stuff then that cannot be open for user space.

PMUFW eFuse default protection is there for a reason ;)

Eg. in our use case as we want to encrypt in example bitstream with eFuse AES key so we cannot use PUF for HUK.

That being said I am happy to collaborate on getting solution that works for both PUF users and for eFuse AES key users.

jenswi-linaro (commented):
Comment for the commit "core: arm64: smc call wrapper for requesting SM services"
We already have a `thread_smccc()` function which does more or less this.
Can that be used instead?

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+{
+	struct smccc_res res = { };
+
+	smccc_smc(pm_api_id,
+		  ((uint64_t)arg1 << 32) | arg0,
+		  ((uint64_t)arg3 << 32) | arg2,
+		  0, 0, 0, 0, 0, &res);
+
+	if (payload && !res.a0)
+		*payload = res.a0 >> 32;
```
jenswi-linaro, line 54: No need to shift, `res.a0` is always `0` if we reach here.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+static void *alloc_aligned_zeroed(size_t sz)
+{
+	size_t alloc_size = 0;
+	size_t cacheline_size = 0;
+	void *ptr = NULL;
+
+	alloc_size = sz;
+	cacheline_size = dcache_get_line_size();
+
+	if (ADD_OVERFLOW(alloc_size, ROUNDUP(alloc_size, cacheline_size),
```
jenswi-linaro, line 68: This doesn't look right, you should probably use `ROUNDUP_OVERFLOW()` instead.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+
+	efuse = alloc_aligned_zeroed(sizeof(struct xilinx_efuse));
+	if (!efuse) {
+		ret = TEE_ERROR_OUT_OF_MEMORY;
+		goto alloc_error;
+	}
+
+	memcpy(buf_aligned, buf, sz);
+
+	efuse->src = (uint64_t)virt_to_phys(buf_aligned);
```
jenswi-linaro, line 105: Is this cast needed?

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+	efuse->size = sz / sizeof(uint32_t);
+	efuse->offset = efuse_offset;
+	efuse->flag = op;
+	efuse->pufuserfuse = puf;
+
+	cache_operation(TEE_CACHEFLUSH, (void *)buf_aligned, sz);
```
jenswi-linaro, line 111: It seems a bit strange that this isn't handled by the firmware at EL3. If this SMC interface already is written in stone I guess that's no way around it, but otherwise I'd recommend fixing it.

By the way, it should be enough to clean the cache entries here.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+	if (res) {
+		if (res == EFUSE_NOT_ENABLED)
+			EMSG("Efuse access is not enabled");
+		else
+			EMSG("Error in efuse access %x", res);
+
+		ret = TEE_ERROR_GENERIC;
+	} else {
+		if (op == EFUSE_READ)
+			memcpy(buf, buf_aligned, sz);
```
jenswi-linaro, line 129: I believe we need to invalidate the cache entries first in case there has been some speculation.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	free(efuse);
+
+alloc_error:
```
jenswi-linaro, line 134: Since we pass this on success too. How about giving it a more neutral name, `out:` for instance?

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+#include <kernel/cache_helpers.h>
+#include <mm/core_memprot.h>
+#include <smccc.h>
+#include <string.h>
+#include <tee/cache.h>
+#include <tee_api_types.h>
+#include <types_ext.h>
+#include <utee_defines.h>
+
+#define EFUSE_ACCESS_SMC	0xC2000035
```
jenswi-linaro, line 17: Do we have any documentation on this interface? A link to some reference manual would be nice.

jenswi-linaro (commented):
Comments for the commit "zynqmp: add minimal firmware driver"

igoropaniuk:
@jenswi-linaro 
> Comment for the commit "core: arm64: smc call wrapper for requesting SM services"
> We already have a thread_smccc() function which does more or less this.
> Can that be used instead?

Fixed, I was initially doing implementation on 3.10 (where there wasn't such function), then up-ported to the latest OP-TEE to upstream that

core/drivers/zynqmp_pm.c
igoropaniuk: added

core/drivers/zynqmp_pm.c
igoropaniuk: removed

core/drivers/zynqmp_pm.c
igoropaniuk: removed that function all all, as it we don't need aligned buffers. Check @ldts comments above

core/drivers/zynqmp_pm.c
igoropaniuk: that wasn't correct at all, I've checked how it's handled in the linux kernel and corrected
res.a0 contains both return code and payload

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,144 @@
...
+};
+
+static uint32_t zynqmp_sip_call(uint32_t pm_api_id, uint32_t arg0,
+				uint32_t arg1, uint32_t arg2, uint32_t arg3,
+				uint32_t *payload)
+{
+	struct smccc_res res = { };
+
+	smccc_smc(pm_api_id,
+		  ((uint64_t)arg1 << 32) | arg0,
```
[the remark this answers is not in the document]
igoropaniuk: fixed

core/drivers/zynqmp_pm.c
igoropaniuk: > By the way, it should be enough to clean the cache entries here.
Right, fixed, thanks!

igoropaniuk:
@jforissier @ldts @jenswi-linaro all comments are addressed, thanks!

igoropaniuk:
Not related fail `OP-TEE.optee_os (Run regression tests (xtest) in QEMUv8) ` test job:
```
make[1]: Leaving directory '/root/optee_repo_qemu_v8/buildroot'
make[1]: Entering directory '/root/optee_repo_qemu_v8/out-br'
/usr/bin/make -j1 O=/root/optee_repo_qemu_v8/out-br HOSTCC="/usr/bin/gcc" HOSTCXX="/usr/bin/g++" syncconfig
make[3]: warning: -j1 forced in submake: resetting jobserver mode.
  GEN     /root/optee_repo_qemu_v8/out-br/Makefile

Your PATH contains spaces, TABs, and/or newline (\n) characters.
This doesn't work. Fix you PATH.
make[2]: *** [support/dependencies/dependencies.mk:27: dependencies] Error 1
make[1]: *** [Makefile:23: _all] Error 2
make: *** [common.mk:323: buildroot] Error 2

```

jforissier:
@igoropaniuk yes sorry about that see https://github.com/OP-TEE/build/pull/525#issuecomment-932345121

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,115 @@
...
+	if (res) {
+		if (res == EFUSE_NOT_ENABLED)
+			EMSG("Efuse access is not enabled");
+		else
+			EMSG("Error in efuse access %x", res);
+
+		ret = TEE_ERROR_GENERIC;
+	}
+
+	cache_operation(TEE_CACHEINVALIDATE, (void *)buf, sz);
```
jenswi-linaro, line 104: `buf` must be cacheline aligned and it's size a multiple of the cacheline size for this to work or you may invalidate unrelated data.
The cast isn't needed.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,115 @@
...
+	efuse->flag = op;
+	efuse->pufuserfuse = puf;
+
+	cache_operation(TEE_CACHECLEAN, (void *)buf, sz);
+	cache_operation(TEE_CACHECLEAN, (void *)efuse, sizeof(*efuse));
+
+	addr = virt_to_phys(efuse);
+
+	res = zynqmp_sip_call(EFUSE_ACCESS_SMC,
+			      (uint32_t)(addr >> 32),
```
jenswi-linaro, line 92: Is the cast needed?

core/drivers/zynqmp_pm.c
ldts: what is the problem with flushing more data from the cache than necessary? is it really worth asking for aligned memory to avoid flushing a few extra bytes?

core/drivers/zynqmp_pm.c
igoropaniuk: nope, will drop it

core/drivers/zynqmp_pm.c
jenswi-linaro: We are invalidating here. Flush (clean and invalidate) can't be used here due to cache speculation.

igoropaniuk: this is done after PMU writes data from caches, flushing can potentially overwrite this data in RAM written by PMU.
IMO having temporary buffers, like it was done from the very beginning would make our life much easier

igoropaniuk: @jenswi-linaro @ldts 

To address this comment I see two options:
1) bring back that temporary buffers, that are allocated within the function with memalign and size a multiple of the cacheline size
2) Add a check is buf is aligned here for the buffer provided as param as function (so imposes a the caller of that function to allocate it correctly), but that will require additional param to pass to the function (besides buffer size we should also provides amount of bytes we want to read from eFUSE):

```
       cacheline_size = dcache_get_line_size();
       if (!IS_ALIGNED((uintptr_t)buf,cacheline_size) ||
           (sz % cacheline_size)) {
               EMSG("Buffer should be cache aligned");
       }

```

igoropaniuk: @jenswi-linaro could you please provide your input about the preferred way to address this?  thanks

jenswi-linaro: Option 1 sounds good to me.

igoropaniuk: Fixed

igoropaniuk:
@jenswi-linaro @jforissier @ldts all comments are addressed

ldts:
@jenswi-linaro  interesting...just curious, why going through a memcopy and two separate allocations when it is enough to ask the user for a single aligned allocation? btw I'll remember the answer if we ever discuss performance issues :)
an alternative would be to return the buffer to the client and let the client do the free...
in my view, even better would be to impose on the client the alignment requirement though...

jenswi-linaro:
> @jenswi-linaro interesting...just curious, why going through a memcopy and two separate allocations when it is enough to ask the user for a single aligned allocation? btw I'll remember the answer if we ever discuss performance issues :) an alternative would be to return the buffer to the client and let the client do the free... in my view, even better would be to impose on the client the alignment requirement though...

I don't know how this function is going to be used so if you have a better solution go ahead with that. My primary focus here has been to avoid tricky memory corruption.

ldts:
> > @jenswi-linaro interesting...just curious, why going through a memcopy and two separate allocations when it is enough to ask the user for a single aligned allocation? btw I'll remember the answer if we ever discuss performance issues :) an alternative would be to return the buffer to the client and let the client do the free... in my view, even better would be to impose on the client the alignment requirement though...
> 
> I don't know how this function is going to be used so if you have a better solution go ahead with that. My primary focus here has been to avoid tricky memory corruption.

right so I think that either the efuse_read function does the allocation and returns the pointer to the allocated buffer back to the caller or the function just expects the buffer to be allocated with the right allocation requirements and simply validates the alignment and its length..

I think the later is better is a better practice; but I'll vote in favour of not having to memcpy between buffers unless it adds a lot of clarity to the code or is really necessary.

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,154 @@
...
+	return args.a0;
+}
+
+static void *alloc_aligned_zeroed(size_t *sz)
+{
+	size_t alloc_size = 0;
+	size_t cacheline_size = 0;
+	void *ptr = NULL;
+
+	if (!sz)
```
etienne-lms, line 73: `if (!sz)` or `if (!sz || !*sz)` ?

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,154 @@
...
+	cache_operation(TEE_CACHECLEAN, efuse, sizeof(*efuse));
+
+	addr = virt_to_phys(efuse);
+
+	res = zynqmp_sip_call(EFUSE_ACCESS_SMC, addr >> 32, addr, 0, 0, NULL);
+	if (res) {
+		if (res == EFUSE_NOT_ENABLED)
+			EMSG("Efuse access is not enabled");
+		else
+			EMSG("Error in efuse access %x", res);
```
etienne-lms, line 132: `EMSG("Error in efuse access %#"PRIx32, res);

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,154 @@
...
+#define EFUSE_ACCESS_SMC	0xC2000035
+#define EFUSE_NOT_ENABLED	29
+
+enum efuse_op {EFUSE_READ = 0, EFUSE_WRITE = 1};
+
+/*
+ * Stores all required details to
+ * read/write efuse memory.
+ * @src:        paddr_t of the buffer to store the data to be write/read
+ *              buffer should be cache aligned
```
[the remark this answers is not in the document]
igoropaniuk: in current implementation it must not, as we use temporary buffers (however there is discussion above)

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,154 @@
...
+#include <kernel/cache_helpers.h>
+#include <kernel/thread.h>
+#include <mm/core_memprot.h>
+#include <string.h>
+#include <tee/cache.h>
+#include <tee_api_types.h>
+#include <types_ext.h>
+#include <utee_defines.h>
+
+#define IS_ALIGNED(x, a) (((x) & ((typeof(x))(a) - 1)) == 0)
```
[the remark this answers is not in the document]
ldts: @etienne-lms isnt that name somewhat unusual? can we rename it to IS_ALIGNED?

core/drivers/zynqmp_pm.c
igoropaniuk: only: `if (!sz)`, in case `*sz == 0`, payload then will be filled with zeros in the next line, so no problem here

core/drivers/zynqmp_pm.c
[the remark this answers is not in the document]
igoropaniuk: not needed anymore at all, dropped it

igoropaniuk:
@etienne-lms thanks, all comments addressed

core/drivers/zynqmp_pm.c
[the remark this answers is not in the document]
ldts: actually the macros are different....so I wouldnt change it

ldts:
@igoropaniuk well I dont agree with having to memcopy and do two allocations to use this function but if the maintainers are ok with it then so be it

core/drivers/zynqmp_pm.c
```diff
@@ -0,0 +1,151 @@
...
+
+	if (!buf)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	buf_aligned_sz = sz;
+	buf_aligned = alloc_aligned_zeroed(&buf_aligned_sz);
+	if (!buf_aligned)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	efuse = malloc(sizeof(*efuse));
```
etienne-lms, line 107: I think you could remove this allocation and define `efuse` in the stack.
As the the `buf_aligned`, i guess you need dynamic allocation. Maybe preallocate a page during initialization, it should be enough for all PM messages and would ensure the data are isolated.

ldts: I would not recommend using a page: reads are usually a few words, no more. I would recommend allocating outside of the function.

ldts: the largest set of buffers that we will eventually have to write (maybe read) will be the  PPK ...that buffer will need to hold 12x32 bit words. that is the worst case scenario.

igoropaniuk: @etienne-lms added allocation of efuse on stack

igoropaniuk:
@jforissier @ldts @jenswi-linaro @etienne-lms any additional comments/objections/suggestions?

jenswi-linaro:
Looks good to me:
However, @ldts had concerns about efficiency. I leave that part to him since he knows better how this is going to be used.

ldts:
I would remove the aligned allocator and just validate the buffer alignment and the size; then we save on an unnecessary memcpy and the additional allocation. 
I just fail to see to reason to have those, that is all.

The typical use case will be 
```
uint32_t dna[3]__align(cache_line) = { 0 };

ret = zynqmp_read_efuse(dna, sizeof(dna), ZYNQMP_DNA_EFUSE_OFFSET, false);
if (ret)
    return ret;
```

PR #4875, opened 2021-09-22
Minor fixes i.MX platforms

clementfaure:
Hello,

Here are small fixes for i.MX platforms.

Thanks,
Clement


jforissier (commented):
- For commit " core: imx: remove useless header include":
- All other commits:

Question: is mx6sll OK regarding the errata? If I understand your comment correctly I think it is (but it mentions SL not SLL specifically). But if it is not the case then two commits should preferably be swapped.

clementfaure:
Hi @jforissier,
Yes, the mx6sl and the mx6sll are not concerned by the errata.
Only platforms imx6d/6q/6dp/6qp are affected by the PL310 issue because these platforms might feature faulty PL310 controllers.

Do you want me to add the link to the errata in the commit message ? https://www.nxp.com/docs/en/errata/IMX6DQCE.pdf

jforissier:
@clementfaure
>  Yes, the mx6sl and the mx6sll are not concerned by the errata.

Cool, then this series is fine in that order.

> Do you want me to add the link to the errata in the commit message ? https://www.nxp.com/docs/en/errata/IMX6DQCE.pdf

Yep good idea, the more information the better (well... up to a certain point!)

clementfaure:
Force-push:
* Link to the PL310 errata added
* Tags applied
* Rebase

PR #4878, opened 2021-09-23
ta: mk: introduce CFLAGS32 and CFLAGS64

clementfaure:
Hello,

With this commit recently merged : https://github.com/OP-TEE/optee_os/pull/4829
It is now required to export `CFLAGS64` and `CFLAGS32` in Yocto optee-test recipe to compile TAs.
This is due to the introduction of : 
https://github.com/OP-TEE/optee_os/blob/3469baa6e7a7b4920596e54ec4c7624314e32042/ta/mk/ta_dev_kit.mk#L29-L30
that make the export `CFLAGS64` and `CFLAGS32` required. Exporting `CFLAGS` only in the recipe won't work.
With this patch, if `CFLAGS` is defined, set `CFLAGS64` and `CFLAGS32` to `$(CFLAGS)`

For reference, the NXP recipes for OPTEE : https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-security/optee-imx?h=hardknott-5.10.52-2.1.0
Patch similar to a change introduced in b09cddcab1eee090886ca94ab9a4958ec6e174e0

Thanks!
Clement

### DIFF
--- a/ta/mk/ta_dev_kit.mk
+++ b/ta/mk/ta_dev_kit.mk
@@ -29,6 +29,12 @@ shlibuuid := $(SHLIBUUID)
 arch-bits-ta_arm32 := 32
 arch-bits-ta_arm64 := 64
 
+# For convenience
+ifdef CFLAGS
+CFLAGS32 ?= $(CFLAGS)
+CFLAGS64 ?= $(CFLAGS)
+endif
+
 ifneq ($V,1)
 q := @
 cmd-echo := true


jforissier (approved):

There is a typo in the commit description: s/CFGLAGS/CFLAGS/. May I suggest a more comprehensive description? The text below simply reflects the things I checked as I reviewed this.

```
ta_dev_kit.mk: set CFLAGS32 and CFLAGS64 to $(CFLAGS) by default

Prior to commit 9fdfcf617e3 ("build: ldelf and TAs can rely on
CFLAGS32/CFLAGS64") was $(arch-bits-$(sm)) not set for TAs (note: $(sm)
is the submodule name, i.e., ta_arm32 or ta_arm64 for TAs). As a result,
$$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) expanded to the value of CFLAGS
in mk/compile.mk.
With the introduction of the above commit, $(arch-bits-$(sm)) is always
set to 32 or 64 depending on the TA architecture, so mk/compile.mk now
evaluates $(CFLAGS32) or $(CFLAGS64). In order to keep the previous
behavior while still allowing specific overrides for 32 or 64-bit, set
CFLAGS32 and CFLAGS64 to $(CFLAGS) by default. This is consistent with
what was done for other submodules in commit b09cddcab1ee ("Introduce
CFLAGS32 and CFLAGS64").
```

clementfaure:
Hi @jforissier,
Yes, I admit that my commit message is too short. I will gladly take yours.
I will just fix the commit ID in your suggestion.
Thanks for your review!

jforissier:
> I will just fix the commit ID in your suggestion.

Oh yes sorry about that, should have been `commit 19fdfcf617e3 ("build: ldelf and TAs can rely on CFLAGS32/CFLAGS64")`.

etienne-lms:
thanks Clément for the fix.

PR #4881, opened 2021-09-24
drivers: imx: add OCOTP driver

clementfaure:
Hello,

Here is a small driver for the On-Chip OTP controller for all i.MX platforms (except 8qm and 8qxp).
This driver only supports the read of fuses through shadow registers for now.
It also implements the `tee_otp_get_die_id()` function.

Thanks!
Clement


core/arch/arm/plat-imx/conf.mk
```diff
@@ -437,6 +439,8 @@ CFG_NSEC_DDR_0_SIZE ?= ($(CFG_DDR_SIZE) - 0x02000000)
 CFG_CRYPTO_SIZE_OPTIMIZATION ?= n
 CFG_MMAP_REGIONS ?= 24
 
+CFG_IMX_OCOTP ?= y
```
jforissier, line 442: How about this:
```
ifeq ($(CFG_NXP_SE05X),y)
$(call force,CFG_IMX_OCOTP,n)
endif
CFG_IMX_OCOTP ?= y
```

.azure-pipelines.yml
```diff
@@ -135,7 +135,7 @@ jobs:
           _make PLATFORM=imx-mx6ulevk CFG_NXP_CAAM=y CFG_CRYPTO_DRIVER=y
           _make PLATFORM=imx-mx6ul9x9evk
           _make PLATFORM=imx-mx6ullevk CFG_WITH_SOFTWARE_PRNG=n CFG_IMX_RNGB=y
-          if [ -d $HOME/se050 ]; then _make PLATFORM=imx-mx6ullevk CFG_NXP_SE05X=y CFG_IMX_I2C=y CFG_STACK_{THREAD,TMP}_EXTRA=8192 CFG_CRYPTO_DRV_{CIPHER,ACIPHER}=y CFG_WITH_SOFTWARE_PRNG=n CFG_NXP_SE05X_{RNG,RSA,ECC,CTR}_DRV=y CFG_NXP_SE05X_PLUG_AND_TRUST_LIB=$HOME/se050/buildarm/libse050.a CFG_NXP_SE05X_PLUG_AND_TRUST=$HOME/se050; fi
+          if [ -d $HOME/se050 ]; then _make PLATFORM=imx-mx6ullevk CFG_NXP_SE05X=y CFG_IMX_I2C=y CFG_IMX_OCOTP=n CFG_STACK_{THREAD,TMP}_EXTRA=8192 CFG_CRYPTO_DRV_{CIPHER,ACIPHER}=y CFG_WITH_SOFTWARE_PRNG=n CFG_NXP_SE05X_{RNG,RSA,ECC,CTR}_DRV=y CFG_NXP_SE05X_PLUG_AND_TRUST_LIB=$HOME/se050/buildarm/libse050.a CFG_NXP_SE05X_PLUG_AND_TRUST=$HOME/se050; fi
```
jforissier, line 138: I think it would be preferable to avoid this, please see below (`conf.mk`).

jforissier (commented):
One thing below, otherwise LGTM.

core/arch/arm/plat-imx/conf.mk
clementfaure: That's what I wanted to do at first but in the end, the SE050 driver and the OCOTP driver can perfectly co-exist like so
`CFG_NXP_SE05X=y CFG_NXP_SE05X_HUK_DRV=n` with `CFG_IMX_OCOTP ?= y` enabled by default.
This is a valid i.MX configuration.

We could also have something like that in `conf.mk` instead:
```
# SE05X and OCOTP both implement tee_otp_get_die_id() 
CFG_IMX_OCOTP ?= y
ifeq ($(CFG_IMX_OCOTP),y)
$(call force,CFG_NXP_SE05X_HUK_DRV,n)
endif
```
but the code conditioned by `CFG_NXP_SE05X_HUK_DRV` would be never compiled by the CI.
That's why I prefer having it disabling the OCOTP in `.azure-pipelines.yml`.

Let me know what you think.

jforissier: I see. IMO the correct way to deal with this is to have something in the makefiles because (1) if we need to change the command line it means we have somehow broken something for users of that particular configuration and (2) it is better to protect from invalid combinations and error out during config check rather than later at link time.

What we want I believe is enable OCOTP by default unless SE05X HUK is enabled in which case it must be disabled, that is:
```
# SE05X HUK and OCOTP both implement tee_otp_get_die_id() 
ifeq ($(CFG_NXP_SE05X_HUK_DRV),y)
$(call force,CFG_IMX_OCOTP,n)
endif
CFG_IMX_OCOTP ?= y
```
...the problem is, `CFG_NXP_SE05X_HUK_DRV ?= y` is set in `core/drivers/crypto/se050/crypto.mk`, after the platform's `conf.mk` is included. That's annoying :disappointed: 

The only option I see is to put the above code at the end of `core/drivers/crypto/se050/crypto.mk`. Should work fine, although setting `CFG_IMX_OCOTP` in that file looks a bit out of place... But I'm afraid we can't do much better with our simplistic config system (if this thing happens a lot we could create a second platform config file included after the crypto stuff).

clementfaure: I agree with you. Setting `CFG_IMX_OCOTP` in `core/drivers/crypto/se050/crypto.mk` really looks out of place ...
Could we have instead
`CFG_IMX_OCOTP ?= y` in `plat-imx/conf.mk`
and
```
ifeq ($(CFG_NXP_SE05X_HUK_DRV),y)
CFG_IMX_OCOTP = n
endif
```
in `se050/crypto.mk` ?

jforissier: > `CFG_IMX_OCOTP = n`

This syntax wouldn't override `make CFG_NXP_SE05X_HUK_DRV=y CFG_IMX_OCOTP=y` and would not detect the conflict either (command line asking for `y`, Makefile silently setting `n`). `override CFG_IMX_OCOTP := n` would override, but not report the conflict from the command line (or even from an environment variable: `CFG_NXP_SE05X_HUK_DRV=y CFG_IMX_OCOTP=y make`).
We could write a more complex expression and test if  `$(origin CFG_IMX_OCOTP)` is `file` to know if we (probably) have the default definition of 'y` as opposed to a user-defined `y`, but that would be quite ugly TBH and not 100% correct either (since `file` origin could be from $(CFG_OPTEE_CONFIG) too...).

At this point I think the "least bad" option is `core/drivers/crypto/se050/crypto.mk`.

clementfaure: Ok I see. I will do as you suggest then.

jforissier: Sounds overly restrictive, but acceptable yes.

clementfaure: Thank you @jforissier.
Tag applied

core/drivers/imx_ocotp.c
```diff
@@ -0,0 +1,286 @@
...
+	unsigned char nb_words;
+	TEE_Result (*get_die_id)(uint64_t *ret_uid);
+};
+
+static vaddr_t g_base_addr;
+static struct mutex fuse_read = MUTEX_INITIALIZER;
+static const struct ocotp_instance *g_ocotp;
+
+static TEE_Result ocotp_get_die_id_mx(uint64_t *ret_uid);
+static TEE_Result ocotp_get_die_id_mx7ulp(uint64_t *ret_uid);
```
jforissier, line 30: Missed that... normally we try to avoid forward declarations, could be done by moving the struct definitions a bit later in the file. No big deal though.

clementfaure: I can move it, no problem

jforissier: OK

PR #4886, opened 2021-09-27
Fix FID for FFA_SECONDARY_EP_REGISTER_64

jforissier:
From: Sudeep Holla <sudeep.holla@arm.com>

Commit ef30482b4ad6 ("plat-vexpress: FF-A: update secondary core init")
updated secondary core init for the vexpress platform. In the process,
it used FFA_SECONDARY_EP_REGISTER with FID 0xC4000084 based on the ALP0
ABI of the spec and the TF-A upstream implementation at the time with a
note that the function ID needs to updated to the one finalied in the
spec and the TF-A implementation.

Assuming it is all finalised now, let us update the same with correct/
updated FID 0xC4000087.

Cc: Jens Wiklander <jens.wiklander@linaro.org>
Cc: Olivier Deprez <olivier.deprez@arm.com>
Cc: Balint Dobszay <balint.dobszay@arm.com>
Signed-off-by: Sudeep Holla <sudeep.holla@arm.com>

### DIFF
--- a/core/arch/arm/include/ffa.h
+++ b/core/arch/arm/include/ffa.h
@@ -69,7 +69,7 @@
 #define FFA_MEM_RECLAIM			U(0x84000077)
 #define FFA_MEM_FRAG_RX			U(0x8400007A)
 #define FFA_MEM_FRAG_TX			U(0x8400007B)
-#define FFA_SECONDARY_EP_REGISTER_64	U(0xC4000084)
+#define FFA_SECONDARY_EP_REGISTER_64	U(0xC4000087)
 
 /* Special value for traffic targeted to the Hypervisor or SPM */
 #define FFA_TARGET_INFO_MBZ		U(0x0)


jforissier:
The CI failure does not seem to be related to this PR but to the recent [rebasing of the Linux kernel onto v5.14](https://lists.trustedfirmware.org/pipermail/op-tee/2021-September/000848.html). `make XEN_BOOT=y check` is broken, the last line of the kernel boot is:
```
[xxx] optee: probing for conduit method.
```
I am investigating.

balint-dobszay-arm:
TF-A v2.5 still uses the old FID. Should we wait with this until TF-A v2.6 is released and https://github.com/OP-TEE/manifest/blob/master/fvp.xml can be updated?

jenswi-linaro:
> TF-A v2.5 still uses the old FID. Should we wait with this until TF-A v2.6 is released and https://github.com/OP-TEE/manifest/blob/master/fvp.xml can be updated?

Makes sense. Thanks for looking into this.

jforissier:
I am reponing this since I suppose we're still waiting on TF-A v2.6.

jforissier:
TFA v2.6 is now available (https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/tag/?h=v2.6).

jenswi-linaro:

jforissier:
> TFA v2.6 is now available (https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/tag/?h=v2.6).

PR to update manifest is https://github.com/OP-TEE/manifest/pull/208.

PR #4890, opened 2021-09-29
plat-stm32mp1: prepare for clock framework

etienne-lms:
A series of changes around resources to prepare support for generic clock API.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/drivers/stm32mp1_clk.c
+++ b/core/arch/arm/plat-stm32mp1/drivers/stm32mp1_clk.c
@@ -1306,20 +1306,28 @@ static TEE_Result stm32mp1_clk_early_init(void)
 	unsigned int i = 0;
 	int len = 0;
 	int ignored = 0;
+	bool rcc_secure = true;
 
-	fdt = get_embedded_dt();
-	node = fdt_node_offset_by_compatible(fdt, -1, DT_RCC_CLK_COMPAT);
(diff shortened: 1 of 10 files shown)


core/arch/arm/plat-stm32mp1/drivers/stm32mp1_clk.c
```diff
@@ -1306,20 +1306,28 @@ static TEE_Result stm32mp1_clk_early_init(void)
...
-	} else {
+	assert(_fdt_reg_base_address(fdt, node) == RCC_BASE);
+	assert(_fdt_get_status(fdt, node) != DT_STATUS_DISABLED);
+
+	if (!rcc_secure) {
 		if (io_read32(stm32_rcc_base() + RCC_TZCR) & RCC_TZCR_TZEN)
 			panic("Refuse to release RCC[TZEN]");
 
-		IMSG("RCC is non-secure");
+		disable_rcc_tzen();
```
jforissier: At this point in the patch series, this function doesn't exist.

jforissier (commented):
One issue below, otherwise LGTM:

core/arch/arm/plat-stm32mp1/drivers/stm32mp1_clk.c
etienne-lms: Oh! i must have missed something! My apologies.

etienne-lms:
I've fixed the faulty commit. Each build/boot ok now. Thanks for the catch.

Updated and A-b tag applied.

PR #4891, opened 2021-09-29
Keep the context of the last session

i-nevo:
Keep the context of the last session instead of deleting it when the
TA finishes executing an invoked commend, so if this TA is invoked
again, there's no need to retrieve the context.

Change-Id: Id2fa242e502a82c4e1b0dd6d9607dcad0a4c7a40


jenswi-linaro:
This is a bit crude. I'd rather try to preserve the involved translation tables. A bit like how translation tables referring to paged pages are treated. See `core/arch/arm/include/mm/pgt_cache.h` and `core/arch/arm/mm/pgt_cache.c`.

PR #4892, opened 2021-09-29
poplar: disable CFG_SECURE_DATA_PATH by default

igoropaniuk:
Since linaro-swg/linux.git branch optee [1] was rebased onto kernel
v5.12, Secure Data Path is broken in xtest [2] because the client side
is based on the ION allocator, which was removed from the kernel.

Therefore, disable SDP support by default.

Link: [1] https://github.com/linaro-swg/linux/tree/optee-v5.12-20210628
Link: [2] https://github.com/OP-TEE/optee_test/blob/3.13.0/host/xtest/regression_1000.c#L1220-L1263

Signed-off-by: Igor Opaniuk <igor.opaniuk@foundries.io>

### DIFF
--- a/core/arch/arm/plat-poplar/conf.mk
+++ b/core/arch/arm/plat-poplar/conf.mk
@@ -25,7 +25,6 @@ ifeq ($(CFG_PL061),y)
 core-platform-cppflags		+= -DPLAT_PL061_MAX_GPIOS=104
 endif
 
-CFG_SECURE_DATA_PATH ?= y
 CFG_TEE_SDP_MEM_BASE ?= 0x02800000
 CFG_TEE_SDP_MEM_SIZE ?= 0x00400000
 


core/arch/arm/plat-poplar/conf.mk
```diff
@@ -25,7 +25,7 @@ ifeq ($(CFG_PL061),y)
 core-platform-cppflags		+= -DPLAT_PL061_MAX_GPIOS=104
 endif
 
-CFG_SECURE_DATA_PATH ?= y
+CFG_SECURE_DATA_PATH ?= n
```
jforissier, line 28: Please just remove the line. We already have `CFG_SECURE_DATA_PATH ?= n` in `mk/config.mk`.

jforissier (commented):
With my comment addressed:

core/arch/arm/plat-poplar/conf.mk
igoropaniuk: done, pushed new version with tag applied. thanks

PR #4894, opened 2021-10-01
plat-vexpress: enable CFG_ENABLE_EMBEDDED_TESTS by default

jforissier:
The vexpress platform family is mainly used for development and testing
so it makes sense to enable internal tests by default. What this does
currently is xtest 1001 runs core internal tests and xtest 1006 runs
TA bget tests.

As a result, remove redundant CFG_ENABLE_EMBEDDED_TESTS=y in the Azure
CI build script and add one configuration with tests disabled.


etienne-lms:

jenswi-linaro:

jforissier:
@jbech-linaro it seems IBART is stuck, doesn't go past "01-HIKEY6220 - PRE TEST".

jbech-linaro:
> @jbech-linaro it seems IBART is stuck, doesn't go past "01-HIKEY6220 - PRE TEST".

I've restarted the IBART service and cancelled/restarted most of the pending jobs.

jbech-linaro:
> > @jbech-linaro it seems IBART is stuck, doesn't go past "01-HIKEY6220 - PRE TEST".
> 
> I've restarted the IBART service and cancelled/restarted most of the pending jobs (some Unicode-exception in the byte stream).

Didn't help, something triggers the same error message 😞 
```
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]: Exception in thread Thread-250:
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]: Traceback (most recent call last):
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:   File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:     self.run()
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:   File "/home/optee/devel/ibart/worker.py", line 301, in run
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:     current_status = status.d[self.start_job()]
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:   File "/home/optee/devel/ibart/worker.py", line 258, in start_job
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:     if not do_pexpect(child, c, e, cr, to):
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:   File "/home/optee/devel/ibart/worker.py", line 93, in do_pexpect
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]:     r = child.expect(e, timeout)
...
Oct 04 08:54:11 jbech-E6430s gunicorn[32117]: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xaa in position 44: invalid start byte
```

Right now I have no clue why this happens and I don't know how to reproduce it. I guess I have to take IBART down once again to see if I can figure out what is going on.

jforissier:
> > > @jbech-linaro it seems IBART is stuck, doesn't go past "01-HIKEY6220 - PRE TEST".
> > 
> > 
> > I've restarted the IBART service and cancelled/restarted most of the pending jobs (some Unicode-exception in the byte stream).
> 
> Didn't help, something triggers the same error message :disappointed:

...

> ```
> Oct 04 08:54:11 jbech-E6430s gunicorn[32117]: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xaa in position 44: invalid start byte
> ```
Perhaps the Buildroot upgrade has changed the console output somehow (coloring maybe?). I am about to do some release testing on HiKey so I will try to figure out where this `0xaa` byte comes from. Is there a way you could get more context, I mean log the whole string that is being parsed and causes an error?

jbech-linaro:
> Is there a way you could get more context, I mean log the whole string that is being parsed and causes an error?

That's the tricky thing here. To be able to figure that out I have to be able to reproduce it at least ones in a local build. I'm not sure if it's buildroot or not. I've had unicode exceptions in other scripts, in those cases it's been "non-standard" characters (like Swedish å, ä, ö and similar characters found in other languages). Once the cause is found, the fix is usually easy.

jforissier:
> > Is there a way you could get more context, I mean log the whole string that is being parsed and causes an error?
> 
> That's the tricky thing here. To be able to figure that out I have to be able to reproduce it at least ones in a local build. I'm not sure if it's buildroot or not. I've had unicode exceptions in other scripts, in those cases it's been "non-standard" characters (like Swedish å, ä, ö and similar characters found in other languages). Once the cause is found, the fix is usually easy.

I did not find any 0xaa sequence in the boot / xtest log on my side. How about working with a byte stream instead of UTF-8? That should work I think since the strings we expect are only from the ASCII subset? (or does the YAML code allow UTF-8?)
https://github.com/jbech-linaro/ibart/blob/master/worker.py#L134-L135

jbech-linaro:
> I did not find any 0xaa sequence in the boot / xtest log on my side. How about working with a byte stream instead of UTF-8?

I think we have to use unicode, since we have no clue really what is coming from the tools that we are using, just watch the amount of information in an IBART log and you'll see the vast amount of text from all tools. IIRC, even the Hikey flash uses Chinese characters

> That should work I think since the strings we expect are only from the ASCII subset? (or does the YAML code allow UTF-8?)

That's true, but if we'd set it to ASCII, then the log itself could look weird in the places where it uses non ASCII characters.

> https://github.com/jbech-linaro/ibart/blob/master/worker.py#L134-L135

I remember that this particular part was rather messy. Lots of issues with the sourcing the config files and setting the PS1.

Maybe via the webhooks, we can see which job it was that caused the issue and then by re-running that, it might be possible to pinpoint the issue. Unfortunately I cannot look into it right now, so I the meantime I've stopped all IBART jobs :/

jforissier:
> > I did not find any 0xaa sequence in the boot / xtest log on my side. How about working with a byte stream instead of UTF-8?
> 
> I think we have to use unicode, since we have no clue really what is coming from the tools that we are using, just watch the amount of information in an IBART log and you'll see the vast amount of text from all tools. IIRC, even the Hikey flash uses Chinese characters

On the contrary, if the issue is that not all output is valid UTF-8 then it is a good reason to specify the format as raw binary.
Another option: just add a try/except and ignore any parsing exception...

> Unfortunately I cannot look into it right now, so I the meantime I've stopped all IBART jobs :/

That's OK, we still have the Azure CI, that covers quite a lot already -- build, QEMUv8, Xen although not the GP suite.

PR #4896, opened 2021-10-01
core: new config switch CFG_PREALLOC_RPC_CACHE

etienne-lms:
Since recent changes in Linux kernel to add a shutdown callback to optee driver, I saw an issue with OP-TEE pager.
Disabling SHM cache is done from a fastcall SMC context that mandates unpaged resources. The issue is that moving SHM release resource to unpaged sections cost 6kB (without debug, on armv7-A/stm32mp1).

This change proposes to introduce CFG_PREALLOC_RPC_CACHE to allow disabling RPC shm preallocation. Enabling the switch (default) ensures required resources are resisdent to pager. Disabling the switch prevents these extra 6kB space in unpaged sections.

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -232,8 +232,8 @@ uint32_t __weak __thread_std_smc_entry(uint32_t a0, uint32_t a1, uint32_t a2,
 
 bool thread_disable_prealloc_rpc_cache(uint64_t *cookie)
 {
-	bool rv;
-	size_t n;
+	bool rv = false;
+	size_t n = 0;
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_FOREIGN_INTR);
 
 	thread_lock_global();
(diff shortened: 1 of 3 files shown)


jforissier:
Let's see if I understand this correctly ;-) Could we have two commits, the first one to fix the pager issue, and the second one to introduce `CFG_PREALLOC_RPC_CACHE`?

etienne-lms:
ok

core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -246,13 +246,16 @@ bool thread_disable_prealloc_rpc_cache(uint64_t *cookie)
...
 	rv = true;
-	for (n = 0; n < CFG_NUM_THREADS; n++) {
-		if (threads[n].rpc_arg) {
-			*cookie = mobj_get_cookie(threads[n].rpc_mobj);
-			mobj_put(threads[n].rpc_mobj);
-			threads[n].rpc_arg = NULL;
-			threads[n].rpc_mobj = NULL;
-			goto out;
+
+	if (IS_ENABLED(CFG_PREALLOC_RPC_CACHE)) {
```
jenswi-linaro, line 250: In case this is true we can return true directly at the beginning of the function without any other checks or locks.

core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -280,7 +283,8 @@ bool thread_enable_prealloc_rpc_cache(void)
 	}
 
 	rv = true;
-	thread_prealloc_rpc_cache = true;
+	if (IS_ENABLED(CFG_PREALLOC_RPC_CACHE))
```
jenswi-linaro, line 286: In case this is true we can return true directly at the beginning of the function without any other checks or locks.

etienne-lms: you're rigth.

core/arch/arm/kernel/thread_optee_smc.c
jenswi-linaro: Good point, I agree.

etienne-lms:
updated

jenswi-linaro:

etienne-lms:
fixup commit squashed & review tags applied

PR #4901, opened 2021-10-05
core: add call_preinitcalls()

jenswi-linaro:
Adds call_preinitcalls() for really early initcalls. This function is
supposed to be called before call_initcalls() is called. With
virtualization enabled it is called in a blocking context when the
OP-TEE partition is created.

This PR breaks out the initcall stuff from https://github.com/OP-TEE/optee_os/pull/4844

I hope to continue the discussion here. I believe that the hooks enabled with with are needed, but the names may have become a bit messy with the extensions added over the time. Perhaps we can find some better name.

### DIFF
--- a/core/include/initcall.h
+++ b/core/include/initcall.h
@@ -30,11 +30,66 @@ struct initcall {
 		{ .func = (fn), }
 #endif
 
+#define preinitcall_begin \
+			SCATTERED_ARRAY_BEGIN(preinitcall, struct initcall)
+#define preinitcall_end SCATTERED_ARRAY_END(preinitcall, struct initcall)
+
 #define initcall_begin	SCATTERED_ARRAY_BEGIN(initcall, struct initcall)
 #define initcall_end	SCATTERED_ARRAY_END(initcall, struct initcall)
 
 #define finalcall_begin	SCATTERED_ARRAY_BEGIN(finalcall, struct initcall)
 #define finalcall_end	SCATTERED_ARRAY_END(finalcall, struct initcall)
+
+/*
+ * The preinit_*(), *_init() and boot_final() macros are used to register
+ * callback functions to be called at different stages during
+ * initialization.
+ *
+ * Functions registered with preinit_*() are always called before functions
+ * registered with *_init().
+ *
+ * Functions registered with boot_final() are called before exiting to
+ * normal world the first time.
+ *
+ * Without virtualization this happens in the order of the defines below.
+ *
+ * However, with virtualization things are a bit different. boot_final()
+ * functions are called first before exiting to normal world the first
+ * time. Functions registered with boot_final() can only operate on the
+ * nexus. preinit_*() functions are called early before the first yielding
+ * call into the partition, in the newly created partition. *_init()
+ * functions are called at the first yielding call.
+ *
+ *  +-------------------------------+-----------------------------------+
+ *  | Without virtualization        | With virtualization               |
+ *  +-------------------------------+-----------------------------------+
+ *  | At the end of boot_init_primary_late() just before the print:     |
+ *  | "Primary CPU switching to normal world boot"                      |
+ *  +-------------------------------+-----------------------------------+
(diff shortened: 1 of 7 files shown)


jenswi-linaro:
Here's some parts from the previous discussion:

> @jenswi-linaro : I'm taking the comments on `core/include/initcall.h` out of context here since I suppose this can take a wihle to sort out.
> 
> > @etienne-lms : `bootfinal` callbacks are called before `*_init` and `preinit` levels. It sounds strange to me
> 
> If we disregard the preinit levels this is an accurate picture of the master branch status.
> 
> > @etienne-lms : Looking at the registered `boot_final` callbacks, i think they are misused:
> > ```
> > core/arch/arm/kernel/secure_partition.c:boot_final(sp_init_all);
> > core/arch/arm/kernel/boot.c:boot_final(release_external_dt);
> > core/drivers/crypto/se050/glue/i2c.c:boot_final(load_trampoline);
> > ```
> > The `secure_partition.c` callbacks could be `early_init` level, there are no dependencies.
> 
> This depends on what will happen with SPs and `CFG_VIRTUALIZATION=y`. I don't think anyone have given much thought on that. I suspect in case it will ever be enabled with virtualization it will either be managed from the nexus or a special partition.
> 
> >  @etienne-lms : The `se050/i2c.c` callback should be called only once a VM has been created, no? (`driver_init_late` level)
> 
> `boot_final()` was added to handle the I2C case, but I'm afraid that the virtualization case was overlooked. What will happen with I2C with virtualization? My guess is that one of the partitions will claim ownership. The same partition that matches the guest in normal world which takes ownership of the physical I2C controller. So for now I'd say that `load_trampoline` isn't compatible with virtualization.
> 
> > @etienne-lms : As for `release_external_dt()`, maybe it should be simply explicitly called from `boot_init_primary_late()` instead of `call_finalcalls()`.
> 
> `boot_final()` was agreed to be the last thing done before exiting to normal world for the first time, this a very good match for what's needed for `release_external_dt` regardless of virtualization.

core/include/initcall.h
```diff
@@ -30,12 +30,67 @@ struct initcall {
...
+ *  | 1. call_preinitcalls()        | In the nexus                      |
+ *  | 2. call_initcalls()           +-----------------------------------+
+ *  | 3. call_finalcalls()          | 1. call_finalcalls()              |
+ *  +-------------------------------+-----------------------------------+
+ *  | "Primary CPU switching to normal world boot" is printed           |
+ *  +-------------------------------+-----------------------------------+
+ *                                  | A guest is created and            |
+ *                                  | virt_guest_created() is called.   |
+ *                                  | After the partition has been      |
+ *                                  | created and activiated.           |
```
jforissier, line 78: activated

core/include/initcall.h
```diff
@@ -30,12 +30,67 @@ struct initcall {
...
+ *                                  +-----------------------------------+
+ *                                  | 2. call_preinitcalls()            |
+ *                                  +-----------------------------------+
+ *                                  | When the partition is recieving   |
```
jforissier, line 82: receiving

jforissier:
> I like the initcall way of registering functions to be called at a certain place, it's decentralized and has a minimal overhead.

Agreed.

> The biggest problem may be to give the different places good names to avoid confusion.

Yes, not an easy task especially since good names cannot just reflect the order in which the functions are called, due to the differences between virt and non-virt. The big comment block in this PR is very helpful in that respect.

> Should we try to sort out this now or wait until all the other pieces needed by https://linaro.atlassian.net/browse/STR-23 are in place?

IMO any suggestions are welcome at any time, we can try to think about better names now, but the current names are not really a problem and even less so if your description is merged. Also, the names are purely internal so no compatibility issue etc.

With the typos fixed this LGTM:

Note, I prefer to merge things when they are actually used, but I also understand it is desirable to split big PRs like 4844. Could we at least have some of the refactoring that uses the preinit calls?

jenswi-linaro:
Update

PR #4931, opened 2021-10-20
FF-A misc

jenswi-linaro:


jforissier:
Could you please wrap the long line in the second commit description and force push? That should quiet the checkpatch warning and also restart the QEMUv8 CI which should now be OK since I merged the xtest fix.

PR #4935, opened 2021-10-21
se050: plug and trust update 0.0.4

ldts:
https://github.com/foundriesio/plug-and-trust/releases/tag/v0.0.4

the apdu pta was tested with this uboot version on imx8mm
https://github.com/ldts/u-boot/commits/se050-apdu-ta

### DIFF
--- a/core/drivers/crypto/se050/adaptors/utils/utils.c
+++ b/core/drivers/crypto/se050/adaptors/utils/utils.c
@@ -18,12 +18,15 @@
 
 static void delete_transient_objects(void)
 {
-	Se05xSession_t *ctx = NULL;
+	SE05x_AttestationType_t att = kSE05x_AttestationType_None;
+	SE05x_SecureObjectType_t type = kSE05x_SecObjTyp_HMAC_KEY;
 	uint8_t more = kSE05x_MoreIndicator_NA;
+	smStatus_t status  = SM_NOT_OK;
+	Se05xSession_t *ctx = NULL;
 	uint8_t *list = NULL;
 	size_t len = 1024;
-	smStatus_t status  = SM_NOT_OK;
 	uint16_t offset = 0;
+	uint8_t mode = 0;
 	uint32_t id = 0;
 	size_t i = 0;
 
@@ -51,7 +54,11 @@ static void delete_transient_objects(void)
 			if (id >= OID_MAX || id == 0)
 				continue;
 
-			if (id & BIT(0))
+			status = Se05x_API_ReadType(ctx, id, &type, &mode, att);
+			if (status != SM_OK)
+				continue;
+
+			if (mode == kKeyObject_Mode_Transient)
 				Se05x_API_DeleteSecureObject(ctx, id);
 		}
 	} while (more == kSE05x_MoreIndicator_MORE);
@@ -59,7 +66,7 @@ static void delete_transient_objects(void)
 	free(list);
 }
 
-static sss_status_t generate_oid(sss_key_object_mode_t mode, uint32_t *val)
+static sss_status_t generate_oid(uint32_t *val)
 {
 	uint32_t oid = OID_MIN;
 	uint32_t random = 0;
@@ -73,11 +80,6 @@ static sss_status_t generate_oid(sss_key_object_mode_t mode, uint32_t *val)
 		if (oid > OID_MAX)
 			continue;
 
-		if (mode == kKeyObject_Mode_Transient)
-			oid |= BIT(0);
-		else
-			oid &= ~BIT(0);
-
 		if (!se050_key_exists(oid, &se050_session->s_ctx)) {
 			*val = oid;
 			return kStatus_SSS_Success;
@@ -87,7 +89,7 @@ static sss_status_t generate_oid(sss_key_object_mode_t mode, uint32_t *val)
 	return kStatus_SSS_Fail;
 }
 
-sss_status_t se050_get_oid(sss_key_object_mode_t mode, uint32_t *val)
+sss_status_t se050_get_oid(uint32_t *val)
 {
 	sss_status_t status = kStatus_SSS_Success;
 	uint16_t mem_t = 0;
@@ -103,7 +105,7 @@ sss_status_t se050_get_oid(sss_key_object_mode_t mode, uint32_t *val)
 	if (mem_t < TRANSIENT_MEMORY_THRESHOLD)
 		delete_transient_objects();
 
-	return generate_oid(mode, val);
+	return generate_oid(val);
 }
 
 static uint32_t se050_key(uint64_t key)
--- a/core/drivers/crypto/se050/core/scp03.c
+++ b/core/drivers/crypto/se050/core/scp03.c
@@ -4,11 +4,11 @@
  * Author: Jorge Ramirez <jorge@foundries.io>
  */
 
-#include <crypto/crypto.h>
+#include <crypto/crypto_se.h>
 #include <config.h>
 #include <se050.h>
 
-TEE_Result crypto_enable_scp03(unsigned int rotate_keys)
+TEE_Result crypto_se_enable_scp03(bool rotate_keys)
 {
 	sss_status_t status = kStatus_SSS_Success;
 
--- a/core/drivers/crypto/se050/core/sub.mk
+++ b/core/drivers/crypto/se050/core/sub.mk
@@ -11,3 +11,4 @@ srcs-$(CFG_NXP_SE05X_DIEID_DRV) += die_id.c
 srcs-$(CFG_NXP_SE05X_RNG_DRV) += rng.c
 srcs-$(CFG_NXP_SE05X_CIPHER_DRV) += cipher.c
 srcs-$(CFG_NXP_SE05X_SCP03_DRV) += scp03.c
+srcs-$(CFG_NXP_SE05X_APDU_DRV) += apdu.c
--- a/core/drivers/crypto/se050/core/sub.mk
+++ b/core/drivers/crypto/se050/core/sub.mk
@@ -3,7 +3,6 @@ include core/drivers/crypto/se050/cflags.mk
 incdirs-y += ../adaptors/include
 incdirs-y += include
 
-srcs-y += scp03.c
 srcs-y += storage.c
 srcs-$(CFG_NXP_SE05X_RSA_DRV) += rsa.c
 srcs-$(CFG_NXP_SE05X_ECC_DRV) += ecc.c
@@ -11,3 +10,4 @@ srcs-$(CFG_NXP_SE05X_CTR_DRV) += ctr.c
 srcs-$(CFG_NXP_SE05X_DIEID_DRV) += die_id.c
 srcs-$(CFG_NXP_SE05X_RNG_DRV) += rng.c
 srcs-$(CFG_NXP_SE05X_CIPHER_DRV) += cipher.c
+srcs-$(CFG_NXP_SE05X_SCP03_DRV) += scp03.c
--- a/core/drivers/crypto/se050/crypto.mk
+++ b/core/drivers/crypto/se050/crypto.mk
@@ -38,6 +38,12 @@ ifeq ($(CFG_NXP_SE05X_SCP03_DRV),y)
 $(call force,CFG_SCP03_PTA,y,Mandated by CFG_NXP_SE05X_SCP03)
 endif
 
+# Allow a secure client to send APDU raw frames
+CFG_NXP_SE05X_APDU_DRV ?= y
+ifeq ($(CFG_NXP_SE05X_APDU_DRV),y)
+$(call force,CFG_APDU_PTA,y,Mandated by CFG_NXP_SE05X_APDU)
+endif
+
 # Random Number Generator
 CFG_NXP_SE05X_RNG_DRV ?= y
 ifeq ($(CFG_NXP_SE05X_RNG_DRV),y)
--- a/core/drivers/crypto/se050/crypto.mk
+++ b/core/drivers/crypto/se050/crypto.mk
@@ -32,6 +32,12 @@ endif
 # SE05X Die Identifier
 CFG_NXP_SE05X_DIEID_DRV ?= y
 
+# Allow a secure client to enable the SCP03 session
+CFG_NXP_SE05X_SCP03_DRV ?= y
+ifeq ($(CFG_NXP_SE05X_SCP03_DRV),y)
+$(call force,CFG_SCP03_PTA,y,Mandated by CFG_NXP_SE05X_SCP03)
+endif
+
 # Random Number Generator
 CFG_NXP_SE05X_RNG_DRV ?= y
 ifeq ($(CFG_NXP_SE05X_RNG_DRV),y)
--- a/core/include/crypto/crypto.h
+++ b/core/include/crypto/crypto.h
@@ -81,10 +81,6 @@ void crypto_authenc_final(void *ctx);
 void crypto_authenc_free_ctx(void *ctx);
 void crypto_authenc_copy_state(void *dst_ctx, void *src_ctx);
 
-#if defined(CFG_NXP_SE05X)
-TEE_Result crypto_enable_scp03(unsigned int rotate_keys);
-#endif
-
 /* Informs crypto that the data in the buffer will be removed from storage */
 void crypto_storage_obj_del(uint8_t *data, size_t len);
 
--- a/core/include/crypto/crypto_se.h
+++ b/core/include/crypto/crypto_se.h
@@ -1,0 +1,27 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) Foundries Ltd. 2021 - All Rights Reserved
+ * Author: Jorge Ramirez <jorge@foundries.io>
+ */
+ /*
+  * This is the Cryptographic Secure Element API, part of the Cryptographic
+  * Provider API.
+  *
+  * These requests shall be handled in the secure element normally placed on
+  * a serial communication bus (SPI, I2C).
+  */
+#ifndef __CRYPTO_SE_H
+#define __CRYPTO_SE_H
+
+#include <tee_api_types.h>
+
+/*
+ * Enable Secure Channel Protocol 03 to communicate with the Secure Element.
+ *
+ * Since SCP03 uses symmetric encryption, this interface also allows the user to
+ * attempt the rotation the keys stored in the Secure Element.
+ *
+ * https://globalplatform.org/wp-content/uploads/2014/07/GPC_2.3_D_SCP03_v1.1.2_PublicRelease.pdf
+ */
+TEE_Result crypto_se_enable_scp03(bool rotate_keys);
+#endif
--- a/core/include/crypto/crypto_se.h
+++ b/core/include/crypto/crypto_se.h
@@ -16,6 +16,28 @@
 #include <tee_api_types.h>
 
 /*
+ * Type identifier for the APDU message as described by Smart Card Standard
+ * ISO7816-4 about ADPU message bodies decoding convention:
+ *
+ * https://cardwerk.com/smart-card-standard-iso7816-4-section-5-basic-organizations/#chap5_3_2
+ */
+enum crypto_apdu_type {
+	CRYPTO_APDU_CASE_NO_HINT,
+	CRYPTO_APDU_CASE_1,
+	CRYPTO_APDU_CASE_2,
+	CRYPTO_APDU_CASE_2E,
+	CRYPTO_APDU_CASE_3,
+	CRYPTO_APDU_CASE_3E,
+	CRYPTO_APDU_CASE_4,
+	CRYPTO_APDU_CASE_4E,
+};
+
+TEE_Result crypto_se_do_apdu(enum crypto_apdu_type type,
+			     uint8_t *header, size_t hdr_len,
+			     uint8_t *src_data, size_t src_len,
+			     uint8_t *dst_data, size_t *dst_len);
+
+/*
  * Enable Secure Channel Protocol 03 to communicate with the Secure Element.
  *
  * Since SCP03 uses symmetric encryption, this interface also allows the user to
--- a/core/pta/apdu.c
+++ b/core/pta/apdu.c
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) Foundries Ltd. 2021 - All Rights Reserved
+ * Author: Jorge Ramirez <jorge@foundries.io>
+ */
+#include <crypto/crypto_se.h>
+#include <kernel/pseudo_ta.h>
+#include <pta_apdu.h>
+
+#define PTA_NAME "pta.apdu"
+
+static TEE_Result get_apdu_type(uint32_t val, enum crypto_apdu_type *type)
+{
+	switch (val) {
+	case PTA_APDU_TXRX_CASE_NO_HINT:
+		*type = CRYPTO_APDU_CASE_NO_HINT;
+		break;
+	case PTA_APDU_TXRX_CASE_1:
+		*type = CRYPTO_APDU_CASE_1;
+		break;
+	case PTA_APDU_TXRX_CASE_2:
+		*type = CRYPTO_APDU_CASE_2;
+		break;
+	case PTA_APDU_TXRX_CASE_2E:
+		*type = CRYPTO_APDU_CASE_2E;
+		break;
+	case PTA_APDU_TXRX_CASE_3:
+		*type = CRYPTO_APDU_CASE_3;
+		break;
+	case PTA_APDU_TXRX_CASE_3E:
+		*type = CRYPTO_APDU_CASE_3E;
+		break;
+	case PTA_APDU_TXRX_CASE_4:
+		*type = CRYPTO_APDU_CASE_4;
+		break;
+	case PTA_APDU_TXRX_CASE_4E:
+		*type = CRYPTO_APDU_CASE_4E;
+		break;
+	default:
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+	return TEE_SUCCESS;
+}
+
+static TEE_Result invoke_command(void *session_context __unused,
+				 uint32_t command_id, uint32_t pt,
+				 TEE_Param params[TEE_NUM_PARAMS])
+{
+	const uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_MEMREF_INPUT,
+						TEE_PARAM_TYPE_MEMREF_INPUT,
+						TEE_PARAM_TYPE_MEMREF_OUTPUT);
+	enum crypto_apdu_type type = CRYPTO_APDU_CASE_NO_HINT;
+	TEE_Result ret = TEE_ERROR_NOT_IMPLEMENTED;
+	size_t len = 0;
+
+	FMSG("command entry point for pseudo-TA \"%s\"", PTA_NAME);
+
+	if (pt != exp_pt)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	switch (command_id) {
+	case PTA_CMD_TXRX_APDU_RAW_FRAME:
+		ret = get_apdu_type(params[0].value.a, &type);
+		if (ret)
+			return ret;
+
+		len = params[3].memref.size;
+		ret = crypto_se_do_apdu(type,
+					params[1].memref.buffer,
+					params[1].memref.size,
+					params[2].memref.buffer,
+					params[2].memref.size,
+					params[3].memref.buffer,
+					&len);
+		if (!ret)
+			params[3].memref.size = len;
+		break;
+	default:
+		return TEE_ERROR_NOT_IMPLEMENTED;
+	}
+
+	return ret;
+}
+
+pseudo_ta_register(.uuid = PTA_APDU_UUID, .name = PTA_NAME,
+		   .flags = PTA_DEFAULT_FLAGS,
+		   .invoke_command_entry_point = invoke_command);
--- a/core/pta/scp03.c
+++ b/core/pta/scp03.c
@@ -4,29 +4,33 @@
  * Author: Jorge Ramirez <jorge@foundries.io>
  */
 
-#include <crypto/crypto.h>
+#include <crypto/crypto_se.h>
 #include <kernel/pseudo_ta.h>
-#include <scp03_pta.h>
+#include <pta_scp03.h>
 
-#define PTA_NAME "scp03.pta"
+#define PTA_NAME "pta.scp03"
 
-static TEE_Result invoke_command(void *pSessionContext __unused,
-				 uint32_t nCommandID, uint32_t pt,
+static TEE_Result invoke_command(void *session_context __unused,
+				 uint32_t command_id, uint32_t pt,
 				 TEE_Param params[TEE_NUM_PARAMS])
 {
 	const uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
 						TEE_PARAM_TYPE_NONE,
 						TEE_PARAM_TYPE_NONE,
 						TEE_PARAM_TYPE_NONE);
+	bool rotate_keys = false;
 
 	FMSG("command entry point for pseudo-TA \"%s\"", PTA_NAME);
 
 	if (pt != exp_pt)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	switch (nCommandID) {
+	switch (command_id) {
 	case PTA_CMD_ENABLE_SCP03:
-		return crypto_enable_scp03(params[0].value.a);
+		if (params[0].value.a == PTA_SCP03_SESSION_ROTATE_KEYS)
+			rotate_keys = true;
+
+		return crypto_se_enable_scp03(rotate_keys);
 	default:
 		break;
 	}
--- a/lib/libutee/include/pta_apdu.h
+++ b/lib/libutee/include/pta_apdu.h
@@ -1,0 +1,47 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (C) 2021, Foundries Limited
+ * Author: Jorge Ramirez <jorge@foundries.io>
+ */
+
+#ifndef __PTA_APDU_H
+#define __PTA_APDU_H
+
+#define PTA_APDU_UUID { 0x3f3eb880, 0x3639, 0x11ec, \
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+
+/*
+ * ADPU based communication with the Secure Element
+ *
+ * [in]  value[0].a           Use APDU TXRX hints: PTA_APDU_TXRX_CASE_*
+ * [in]  memref[1].buffer     APDU header.
+ * [in]  memref[1].size       APDU header length.
+ * [in]  memref[2].buffer     request (APDU raw frame).
+ * [in]  memref[2].size       request length.
+ * [out] memref[3].buffer     response (APDU raw frame).
+ * [out] memref[3].size       response length.
+ *
+ * Result:
+ * TEE_SUCCESS - Invoke command success
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_NOT_IMPLEMENTED - Invoke command not implemented
+ * TEE_ERROR_GENERIC - Invoke command failure
+ */
+#define PTA_CMD_TXRX_APDU_RAW_FRAME		0
+
+/*
+ * Type identifier for the APDU message as described by Smart Card Standard
+ * ISO7816-4 about ADPU message bodies decoding convention:
+ *
+ * https://cardwerk.com/smart-card-standard-iso7816-4-section-5-basic-organizations/#chap5_3_2
+ */
+#define PTA_APDU_TXRX_CASE_NO_HINT	0
+#define PTA_APDU_TXRX_CASE_1		1
+#define PTA_APDU_TXRX_CASE_2		2
+#define PTA_APDU_TXRX_CASE_2E		3
+#define PTA_APDU_TXRX_CASE_3		4
+#define PTA_APDU_TXRX_CASE_3E		5
+#define PTA_APDU_TXRX_CASE_4		6
+#define PTA_APDU_TXRX_CASE_4E		7
+
+#endif /* __PTA_APDU_H */
--- a/lib/libutee/include/pta_scp03.h
+++ b/lib/libutee/include/pta_scp03.h
@@ -3,8 +3,8 @@
  * Copyright (C) 2020, Foundries Limited
  */
 
-#ifndef __SCP03_PTA_H
-#define __SCP03_PTA_H
+#ifndef __PTA_SCP03_H
+#define __PTA_SCP03_H
 
 #define PTA_SCP03_UUID { 0xbe0e5821, 0xe718, 0x4f77, \
 			{ 0xab, 0x3e, 0x8e, 0x6c, 0x73, 0xa9, 0xc7, 0x35 } }
@@ -12,7 +12,7 @@
 /*
  * Enable SCP03 support on the SE
  *
- * [in]     value[0].a    extra option (0 do not rotate keys, 1 rotate keys)
+ * [in]     value[0].a    Use session keys PTA_SCP03_SESSION_*
  *
  * Result:
  * TEE_SUCCESS - Invoke command success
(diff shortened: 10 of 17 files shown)


lib/libutee/include/apdu_pta.h
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright (C) 2021, Foundries Limited
+ */
+
+#ifndef __APDU_PTA_H
+#define __APDU_PTA_H
+
+#define PTA_APDU_UUID { 0x3f3eb880, 0x3639, 0x11ec, \
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+/*
+ * ADPU based communication with the SE05X secure element
```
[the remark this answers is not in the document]
ldts: I am writing this code as part of a more extensive integration with a service daemon. header, and buffers are generic but the request type might be only for the SE050 (although other implementations might want to use these params as well so the TXRX functions can be simplified)

This is how the types get handled on the Plug-and-Trust stack:
https://github.com/foundriesio/plug-and-trust/commit/5caf02c122f5fbb77aa41592d4b9ca0c6ef2c188#diff-4271878f3b672bd62c729c47fceec849f31ee69cc403df460d71bccd36e6ce7fR10

core/include/crypto/crypto.h
```diff
@@ -83,6 +83,22 @@ void crypto_authenc_copy_state(void *dst_ctx, void *src_ctx);
...
+
+enum crypto_apdu_type {
+	CRYPTO_APDU_NONE,
+	CRYPTO_APDU_TYPE_0,
+	CRYPTO_APDU_TYPE_1,
+	CRYPTO_APDU_TYPE_2,
+	CRYPTO_APDU_TYPE_3,
+	CRYPTO_APDU_TYPE_4,
+	CRYPTO_APDU_TYPE_5,
+	CRYPTO_APDU_TYPE_6,
```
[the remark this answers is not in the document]
ldts: they are just a commodity flag to simplify the actual functions that will handle the transfers...using them, since I am integrating with a generic NXP service, is simpler and will require less maintenance than modifying the service.

ldts:
@etienne-lms I will share tomorrow a libptteec (simple optee-client lib) to allow integration with the user space agent.

lib/libutee/include/apdu_pta.h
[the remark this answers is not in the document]
etienne-lms: My point here is to question whether this PTA exposes generic commands (in which case "APDU PTA" sounds ok), or if it the PTA API is dedicated to SE050 APDU support. Maybe the PTA expects to be generic but it must expose dedicated commands for SE050 support since some extra config info (as `SE050_APDU_TYPE_*` IDs are).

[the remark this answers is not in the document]
ldts: There is nothing specific to the SE50 in this software API. Communication via APDU frames consist on sending and receiving buffers. And I added an extra param to help with the handling of those buffers in whichever way the driver considers necessary.

But if you prefer I can rename it all and make it SE050 specific - just to save time (upstreaming is expensive as everyone knows). And then we can come up with a different API if ever any other secure element gets merged in op-tee....shall I do that?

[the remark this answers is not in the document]
etienne-lms: AFAICT generic APDU does not consider types info carried by `param[0].value.a` and this makes the API not generic. Could there be a way to have a generic PTA for APDU? Is there a need?
All in one, i think its not really clean to have an APDU PTA which interface only applies to SE05x devices.

This comment applies also to https://github.com/OP-TEE/optee_client/pull/290. Is libapduteec designed only for OP-TEE relying on a SE05x device?

I apology, I don't know anything about APDUs. Maybe someone can give some feedback on that concern.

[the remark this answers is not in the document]
ldts: But I disagree with that. The API is generic: ie it provides an interface to send and receive a buffer plus an enum to characterize - if necessary - the processing of those buffers.  
The SE050 processing benefits from having that enum but others might not (or might too). I guess I dont see why that makes it non-generic and not the opossite.
could you clarify? please

WRT to OP-TEE/optee_client#290 is just the same thing. Just as this PR is not SE05x specific.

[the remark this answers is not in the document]
etienne-lms: Maybe I'm the only one wondering about if this API allows to interface other devices. Should they use TYPE_NONE if not relevant? Client of the libapduteec is expected to know when it interfaces a SE05x and  set a non-TYPE_NONE?

If I'm the only one wondering about that, I guess I misunderstand the context. If so, fine with the interface but remove SE05x reference from the PTA API header file. Or change command name to highlight it's an APDU typed command that complies with SE05x (maybe other devices?) ADPU types definition.

core/include/crypto/crypto.h
```diff
@@ -83,6 +83,22 @@ void crypto_authenc_copy_state(void *dst_ctx, void *src_ctx);
 
 #if defined(CFG_NXP_SE05X)
```
jenswi-linaro, line 84: Would it make sense to move this entire block to a separate .h file, `<crypto/nxp_se05x.h>`? It seems that normal crypto users will not use these function so this only adds noise to ignore. A comment or two wouldn't hurt either.

ldts: @jenswi-linaro you mean temporarily until support for another APDU controlled secure element is added to OP-TEE?

ldts: both of these interfaces are generic but the se05x is the only user currently

[SCP03](:https://www.cardlogix.com/glossary/scp03-secure-channel-protocol-3/#:~:text=SCP03%20is%20a%20Asymmetric%20Secure,resistant%20to%20overhearing%20and%20tampering.&text=SCP03%20uses%20the%20generic%20composition,CPA%20and%20INT%2DCTXT%20secure.)

[APDU](https://en.wikipedia.org/wiki/ISO/IEC_7816#7816-8:_Commands_and_mechanisms_for_security_operations)

jenswi-linaro: `crypto/apdu_se.h` is fine if you prefer that.

ldts: ok

ldts: I changed it for crypto/secure_element.h..hope that is ok

core/include/crypto/crypto.h
```diff
@@ -22,6 +22,7 @@
 #ifndef __CRYPTO_CRYPTO_H
 #define __CRYPTO_CRYPTO_H
 
+#include <crypto/secure_element.h>
```
jenswi-linaro, line 25: I don't think this is needed here. Nothing in this file is supposed to depend on anything from `crypto/secure_element.h`.

ldts: right, I didnt add it for dependency but for consistency as I think it should be part of the crypto api. 
For example, retrieving the rng from a secure element would be just sending an APDU frame request using this interface...should the user include the superset crypto.h instead of the secure_element.h?

ldts: jfyi crypto_do_apdu (in the case of the se05x secure element) will - when SCP03 has been enabled- encrypt the raw APDU frame using the SPC03 keys using OP-TEE tomcrypt and send the frame to the linux kenel using the i2c trampoline

ldts: if SCP03 was not enabled, it will just use the trampoline to send the request and receive the response.

jenswi-linaro: To me it's as much part of the crypto api as `crypto/crypto_accel.h` is (which isn't included here).
There's a value in not exposing more than needed. I know this is a tradeoff, in this case I believe it makes sense to draw the line here.

ldts: right, and that makes little sense to me as well although is kind of different as the prefix does some of the changes ...

but anyway, ok, so to follow that same pattern I will have to rename the file to 
- crypto_secure_element.h

and the interfaces:
- crypto_secure_element_enable_scp03 
- crypto_secure_element_do_apdu

ok?

jenswi-linaro: Well, if you insist. I would have kept the function names as they are, I just wanted to get it out of the way of the "common case".

jenswi-linaro:

etienne-lms:
Apologies again but I still have concerns with parameter type in the APDU PTA API.
Maybe the API description could specify that type should be `APDU_TYPE_NONE` is the argument is not applicable. Is the argument mandatory when using an SE05x remove device?

By the way, please squash fixup commits "core: pta: scp03, beautify" and "core: crypto: Secure Element cryptographic interface" into commit "core: pta: APDU pseudo trusted application".

ldts:
> Apologies again but I still have concerns with parameter type in the APDU PTA API. Maybe the API description could specify that type should be `APDU_TYPE_NONE` is the argument is not applicable. Is the argument mandatory when using an SE05x remove device?
> 

@etienne-lms I understand your point. let me dig a bit more on this - I am porting the driver and maybe I can find some reference to some document. Maybe you are right and this is not platform specific - as it seems by inspecting the code- but part of some standard. Is a bit confusing because I cant find any references..anyway, let me come back to you.

ldts:
@etienne-lms 
https://cardwerk.com/smart-card-standard-iso7816-4-section-5-basic-organizations/#chap5_1_4_2

I think this is it. Check 5.3.1 Command APDU and 5.3.2

Which are the 4 types defined (plus the other types with extended); let me rework the interface.

etienne-lms (commented):
> https://cardwerk.com/smart-card-standard-iso7816-4-section-5-basic-organizations/#chap5_1_4_2
> 
> I think this is it. Check 5.3.1 Command APDU and 5.3.2
> 
> Which are the 4 types defined (plus the other types with extended); let me rework the interface.

Indeed now I understand what this type argument refers to. However, as such argument is not standardized at all, I think it would be better remove type argument from the Library and PTA API It would make the library and PTA far more portable.
I think  `crypto_secure_element_do_apdu()` from _.../se050/core/apdu.c_ could implement the logic that parses command data and set that type value SE05x seems to expect. 
This remains my opinion, for what it stands.

ldts:
> > https://cardwerk.com/smart-card-standard-iso7816-4-section-5-basic-organizations/#chap5_1_4_2
> > I think this is it. Check 5.3.1 Command APDU and 5.3.2
> > Which are the 4 types defined (plus the other types with extended); let me rework the interface.
> 
> Indeed now I understand what this type argument refers to. However, as such argument is not standardized at all, I think it would be better remove type argument from the Library and PTA API It would make the library and PTA far more portable. I think `crypto_secure_element_do_apdu()` from _.../se050/core/apdu.c_ could implement the logic that parses command data and set that type value SE05x seems to expect. This remains my opinion, for what it stands.

Right, and the SE050 library does extract the case from the APDU frame when the INVALID type is passed. However it also uses the explicit case requests on others hence why I'd rather keep it as is, it avoids parsing too frequently.

lib/libutee/include/apdu_pta.h
```diff
@@ -0,0 +1,30 @@
...
+#define __APDU_PTA_H
+
+#define PTA_APDU_UUID { 0x3f3eb880, 0x3639, 0x11ec, \
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+/*
+ * ADPU based communication with the Secure Element
+ *
+ * [in]  value[0].a           APDU request type.
+ * [in]  memref[1].buffer     APDU header.
+ * [in]  memref[1].size       APDU header length.
```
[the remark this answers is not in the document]
ldts: yes I could do it on the TA but I opted for doing that length check on the se050 processing call in the external Plug-and-Trust library - trying to be as generic as possible to other users who might want to use the header differently - maybe for workarounds.

ldts:

hi @jenswi-linaro I didnt add your ack yet since some other minor changes followed from @etienne-lms review. hope that is ok

lib/libutee/include/apdu_pta.h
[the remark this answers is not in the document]
etienne-lms: This function and this library is expected to follow APDU standard unless I missed something.
That said, passing a 32bit as a 4 byte buffer is fine. But it shall be 4 bytes, no more, no less.

core/include/crypto/crypto_secure_element.h
```diff
@@ -0,0 +1,44 @@
...
+  *
+  * These requests shall be handled in the secure element normally placed on
+  * a serial communication bus (SPI, I2C).
+  */
+#ifndef __CRYPTO_SECURE_ELEMENT_H
+#define __CRYPTO_SECURE_ELEMENT_H
+
+#include <tee_api_types.h>
+
+#if defined(CFG_NXP_SE05X)
```
etienne-lms, line 19: the functions are not SE05x specific. remove this directive.

core/include/crypto/crypto_secure_element.h
```diff
@@ -0,0 +1,44 @@
...
+TEE_Result crypto_secure_element_enable_scp03(unsigned int rotate_keys);
```
etienne-lms, line 20: This API function deserves an inline description comment

Prefer bool for rotate keys if it's a boolean information.

core/include/crypto/crypto_secure_element.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+enum crypto_apdu_type {
+	CRYPTO_APDU_CASE_1,
+	CRYPTO_APDU_CASE_2,
+	CRYPTO_APDU_CASE_2E,
+	CRYPTO_APDU_CASE_3,
+	CRYPTO_APDU_CASE_3E,
+	CRYPTO_APDU_CASE_4,
+	CRYPTO_APDU_CASE_4E,
+	CRYPTO_APDU_CASE_NO_HINT = 0xFF,
```
etienne-lms, line 36: no need for a specific value. I would put it in fisrt place in the enum.

core/pta/apdu.c
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (C) Foundries Ltd. 2021 - All Rights Reserved
+ * Author: Jorge Ramirez <jorge@foundries.io>
+ */
+
+#include <crypto/crypto_secure_element.h>
+#include <kernel/pseudo_ta.h>
+#include <apdu_pta.h>
```
etienne-lms, line 9: sort #includes please

core/pta/apdu.c
```diff
@@ -0,0 +1,51 @@
...
+
+#define PTA_NAME "apdu.pta"
+
+static TEE_Result invoke_command(void *pSessionContext __unused,
```
etienne-lms, line 13: snake_case (s/pSessionContext/session_context/)

core/pta/scp03.c
```diff
@@ -4,14 +4,14 @@
...
  */
 
-#include <crypto/crypto.h>
+#include <crypto/crypto_secure_element.h>
 #include <kernel/pseudo_ta.h>
 #include <scp03_pta.h>
 
 #define PTA_NAME "scp03.pta"
 
 static TEE_Result invoke_command(void *pSessionContext __unused,
```
etienne-lms, line 13: by the way, s/pSessionContext/pSessionContext/

lib/libutee/include/apdu_pta.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef __APDU_PTA_H
+#define __APDU_PTA_H
+
+#define PTA_APDU_UUID { 0x3f3eb880, 0x3639, 0x11ec, \
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+/*
+ * ADPU based communication with the Secure Element
+ *
+ * [in]  value[0].a           APDU request type.
```
etienne-lms, line 14: You mean explicit values for the IDs carries in this type parameter.

lib/libutee/include/apdu_pta.h
```diff
@@ -0,0 +1,30 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (C) 2021, Foundries Limited
+ */
+
+#ifndef __APDU_PTA_H
+#define __APDU_PTA_H
```
etienne-lms, line 7: again: almost all PTAs in optee_os use _pta__\<_some_naming_>.h_ as header file name (but scp03 PTA (cb1b1ecd336abb5c3cca92eb702f2a794b023e40) and rng PTA (b2ac1d0fa34ca28382fcedb30d938ccd4e8a5316). Please, help for the consistency in optee_os, and use standard _pta_apdu.h_.

@jforissier, @jenswi-linaro @ruchi393, once for all for my information: do you think it's worth having some consistency in optee_os PTAs header files naming/layout/API description format? There are the reference documentation for PTA embedded in OP-TEE core. If you don't think so, fine, I won't ask the question again.

core/include/crypto/crypto_secure_element.h
etienne-lms: and move the prototype below `enum crypto_apdu_type` definition

lib/libutee/include/apdu_pta.h
ldts: I agree. When I submitted those drivers I was asked to have the name inverted so I remember changing them. Now you are asking for the opposite (with an emphasis on 'again'). So yeah, while I personally don't care too much, having some consistency would be great. what is going to be then?

lib/libutee/include/apdu_pta.h
ldts: yes, once we agree on the libseteec names, I'll just import them here

core/pta/apdu.c
ldts: um, sorry

core/include/crypto/crypto_secure_element.h
ldts: sure, sounds good

core/include/crypto/crypto_secure_element.h
ldts: yes

core/include/crypto/crypto_secure_element.h
ldts: yeah, well is the only supported SE...but ok, do you want me to continue adding functionality and squashing? the list is growing...

lib/libutee/include/apdu_pta.h
ldts: I'll push again as it is. I can do another commit to rename these three header files - I probably should but please confirm.

etienne-lms (commented):
For commit "crypto: se050: Foundries Plug-and-Trust release 0.0.4"

For commit "crypto: se050: rename huk driver to die_id"
with a minor comment on commit message. I think it is preferable the commit log message (after header file) also describes what is done, even if repeating or rephrasing the commit log header line. Add a sentence to state the source file is renamed.

ldts:
do you want the header file renaming squashed or as a separate commit?

ldts:
> do you want the header file renaming squashed or as a separate commit?

I'll squash for consistency with a previous ask to squash  the scp03 changes on this PR

ldts:
@etienne-lms done.

core/include/crypto/crypto_secure_element.h
```diff
@@ -0,0 +1,49 @@
...
+	CRYPTO_APDU_CASE_1,
+	CRYPTO_APDU_CASE_2,
+	CRYPTO_APDU_CASE_2E,
+	CRYPTO_APDU_CASE_3,
+	CRYPTO_APDU_CASE_3E,
+	CRYPTO_APDU_CASE_4,
+	CRYPTO_APDU_CASE_4E,
+};
+
+TEE_Result crypto_secure_element_do_apdu(enum crypto_apdu_type type,
```
etienne-lms, line 35: add an inline description for the API function. header = 4 byte APDU header, src = tx message, dst = rx reponse message,

core/pta/apdu.c
```diff
@@ -0,0 +1,50 @@
...
+	default:
+		return TEE_ERROR_NOT_IMPLEMENTED;
+	}
+
+	return ret;
+}
+
+pseudo_ta_register(.uuid = PTA_APDU_UUID, .name = PTA_NAME,
+		   .flags = PTA_DEFAULT_FLAGS,
+		   .invoke_command_entry_point = invoke_command);
```
etienne-lms, line 50: APDU PTA can be freely called from REE clients without distinctions, or from TA clients? (no need for a open_session entry pointer?)

core/pta/apdu.c
```diff
@@ -0,0 +1,50 @@
...
+
+	FMSG("command entry point for pseudo-TA \"%s\"", PTA_NAME);
+
+	if (pt != exp_pt)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	switch (command_id) {
+	case PTA_CMD_TXRX_APDU_RAW_FRAME:
+		len = params[3].memref.size;
+		ret = crypto_secure_element_do_apdu(params[0].value.a,
```
etienne-lms, line 31: This reuses straight values of `PTA_APDU_TXRX_CASE_*` as values of `CRYPTO_APDU_CASE_*`. Consistency should be always asserted at build time.

The point is on one hand PTA API must define explicit values (0, 1, ...) for these IDs, on the other hand 
`enum crypto_apdu_type` from _crypto_secure_element.h_ cannt rely on the PTA IDs.
So either a conversion, or value compliance + assertion at build time.

lib/libutee/include/pta_scp03.h
```diff
@@ -22,4 +22,10 @@
  */
 #define PTA_CMD_ENABLE_SCP03		0
 
-#endif /* __SCP03_PTA_H */
+/* Enable the session using the current keys in the Secure Element */
+#define PTA_SCP03_SESSION_CURRENT_KEYS	0
```
etienne-lms, line 26: `PTA_SCP03_SESSION_CURRENT_KEY`  at singular, no?

lib/libutee/include/pta_scp03.h
```diff
@@ -22,4 +22,10 @@
...
+
+/* Enable the session after replacing the current keys in the Secure Element */
+#define PTA_SCP03_SESSION_ROTATE_KEYS	1
```
etienne-lms, line 29: There is only 1 key that is rotated, no?

etienne-lms (commented):
I think it would be better commit "core: pta: APDU pseudo trusted application" is split in 2. First introduce _crypto_secure_element.h_ and `crypto_secure_element_do_apdu()` then introduce APDU PTA.

core/pta/apdu.c
ldts: not that I can think of...any TA client to REE client can send APDUs to the Secure Element at any time

etienne-lms: ...and request or not key rotation, right?

core/pta/apdu.c
ldts: seems an overkill at this level of abstraction but ok..

core/pta/apdu.c
ldts: yes. In fact the the nxp agent linking to the libseteec might request rotation...and for sure the scp03 TA in uboot will need to be able to do it too.
see 
https://github.com/u-boot/u-boot/blob/master/cmd/scp03.c

ldts:
....

fatal: unable to access 'https://gerrit.googlesource.com/git-repo/': Failed to connect to gerrit.googlesource.com port 443: Connection refused
##[error]Bash exited with code '1'.
Finishing: CmdLine

ldts:
> I think it would be better commit "core: pta: APDU pseudo trusted application" is split in 2. First introduce _crypto_secure_element.h_ and `crypto_secure_element_do_apdu()` then introduce APDU PTA.

ah, no cmon, you asked me to squash those....

core/include/crypto/crypto_secure_element.h
ldts: dont we have to be coherent with the rest of the documents...have you checked crypto.h or crypto_accel.h? I think we are going to far here....

etienne-lms:
> ah, no cmon, you asked me to squash those....#4935 (comment)
>
> > By the way, please squash fixup commits "core: pta: scp03, beautify" and "core: crypto: Secure Element cryptographic interface" into commit "core: pta: APDU pseudo trusted application".

Keep as is. my bad, i'm wrong, thank to remind. I leave it to other to see whether a new API funct in crypto layer and an new PTA don't deserve each a specific commit in optee_os git revision history. My apologies if i made a bad comment at the time or now. It's quite recent that the split between se050, scp03 and apdu is (a bit) clear to me.

ldts:
> > ah, no cmon, you asked me to squash those....#4935 (comment)
> > > By the way, please squash fixup commits "core: pta: scp03, beautify" and "core: crypto: Secure Element cryptographic interface" into commit "core: pta: APDU pseudo trusted application".
> 
> Keep as is. my bad, i'm wrong, thank to remind. I leave it to other to see whether a new API funct in crypto layer and an new PTA don't deserve each a specific commit in optee_os git revision history. My apologies if i made a bad comment at the time or now. It's quite recent that the split between se050, scp03 and apdu is (a bit) clear to me.

um, no I think you are right...I shouldnt have agreed to squash in the first place. will do

core/drivers/crypto/se050/core/sub.mk
```diff
@@ -7,7 +7,7 @@ srcs-y += storage.c
 srcs-$(CFG_NXP_SE05X_RSA_DRV) += rsa.c
 srcs-$(CFG_NXP_SE05X_ECC_DRV) += ecc.c
 srcs-$(CFG_NXP_SE05X_CTR_DRV) += ctr.c
-srcs-$(CFG_NXP_SE05X_HUK_DRV) += huk.c
+srcs-$(CFG_NXP_SE05X_DIEID_DRV) += die_id.c
```
jforissier, line 10: 1. Minor: `DIE_ID` would be a bit easier to read. Your choice.
2. Should a few lines be changed in `die_id.c`? (`se050_huk`, etc.)

.azure-pipelines.yml
```diff
@@ -135,7 +135,7 @@ jobs:
           _make PLATFORM=imx-mx6ulevk CFG_NXP_CAAM=y CFG_CRYPTO_DRIVER=y
           _make PLATFORM=imx-mx6ul9x9evk
           _make PLATFORM=imx-mx6ullevk CFG_WITH_SOFTWARE_PRNG=n CFG_IMX_RNGB=y
-          if [ -d $HOME/se050 ]; then _make PLATFORM=imx-mx6ullevk CFG_NXP_SE05X=y CFG_IMX_I2C=y CFG_STACK_{THREAD,TMP}_EXTRA=8192 CFG_CRYPTO_DRV_{CIPHER,ACIPHER}=y CFG_WITH_SOFTWARE_PRNG=n CFG_NXP_SE05X_{RNG,RSA,ECC,CTR}_DRV=y CFG_NXP_SE05X_PLUG_AND_TRUST_LIB=$HOME/se050/buildarm/libse050.a CFG_NXP_SE05X_PLUG_AND_TRUST=$HOME/se050; fi
+          if [ -d $HOME/se050 ]; then _make PLATFORM=imx-mx6ullevk CFG_NXP_SE05X=y CFG_IMX_I2C=y CFG_STACK_{THREAD,TMP}_EXTRA=8192 CFG_CRYPTO_DRV_{CIPHER,ACIPHER}=y CFG_WITH_SOFTWARE_PRNG=n CFG_NXP_SE05X_{DIEID,RNG,RSA,ECC,CTR}_DRV=y CFG_NXP_SE05X_PLUG_AND_TRUST_LIB=$HOME/se050/buildarm/libse050.a CFG_NXP_SE05X_PLUG_AND_TRUST=$HOME/se050; fi
```
jforissier, line 138: I'd rather remove the useless flags in a separate commit first (all the ones that are `== y` by default).

.azure-pipelines.yml
```diff
@@ -161,7 +161,7 @@ jobs:
           _make PLATFORM=imx-mx7ulpevk
           _make PLATFORM=imx-mx8mmevk
           _make PLATFORM=imx-mx8mmevk CFG_NXP_CAAM=y CFG_CRYPTO_DRIVER=y
-          if [ -d $HOME/se050 ]; then _make PLATFORM=imx-mx8mmevk CFG_NXP_CAAM=y CFG_NXP_CAAM_RNG_DRV=y CFG_NXP_SE05X=y CFG_IMX_I2C=y CFG_STACK_{THREAD,TMP}_EXTRA=8192 CFG_CRYPTO_DRV_{CIPHER,ACIPHER}=y CFG_NXP_SE05X_RNG_DRV=n CFG_WITH_SOFTWARE_PRNG=n CFG_NXP_SE05X_{RSA,ECC,CTR}_DRV=y CFG_NXP_SE05X_PLUG_AND_TRUST_LIB=$HOME/se050/build/libse050.a CFG_NXP_SE05X_PLUG_AND_TRUST=$HOME/se050 ; fi
+          if [ -d $HOME/se050 ]; then _make PLATFORM=imx-mx8mmevk CFG_NXP_CAAM=y CFG_NXP_CAAM_RNG_DRV=y CFG_NXP_SE05X=y CFG_IMX_I2C=y CFG_STACK_{THREAD,TMP}_EXTRA=8192 CFG_CRYPTO_DRV_{CIPHER,ACIPHER}=y CFG_NXP_SE05X_RNG_DRV=n CFG_WITH_SOFTWARE_PRNG=n CFG_NXP_SE05X_{DIEID,RSA,ECC,CTR}_DRV=y CFG_NXP_SE05X_PLUG_AND_TRUST_LIB=$HOME/se050/build/libse050.a CFG_NXP_SE05X_PLUG_AND_TRUST=$HOME/se050 ; fi
```
jforissier, line 164: Same here

jforissier (commented):
A few small things, in any case:

.azure-pipelines.yml
ldts: ok. will do a separate commit for that.

ldts: I was thinking about this @jforissier the _issue_ here is that the se05x driver can coexist with other crypto drivers (ie, CAAM) and so having the visual information this way could be helpful - at least it is for me....ie, we can have the HUK from CAAM and RNG and not from the SE05x...is it a real problem to keep the pipeline as is? I understand that is not required but I think it helps.

jforissier: @ldts yes, it's OK.

core/drivers/crypto/se050/core/scp03.c
```diff
@@ -4,11 +4,11 @@
  * Author: Jorge Ramirez <jorge@foundries.io>
  */
 
-#include <crypto/crypto.h>
+#include <crypto/crypto_secure_element.h>
```
etienne-lms, line 7: For commit "core: crypto: Secure Element cryptographic interface": 
sorry for this late comment. what about _crypto_se.h_ for short? API function would shorten to `crypto_se_enable_scp03()`.

lib/libutee/include/pta_scp03.h
etienne-lms: ok

lib/libutee/include/pta_scp03.h
etienne-lms: ok

lib/libutee/include/pta_apdu.h
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright (C) 2021, Foundries Limited
+ * Author: Jorge Ramirez <jorge@foundries.io>
+ */
+
+#ifndef __PTA_APDU_H
+#define __PTA_APDU_H
+
+#define PTA_APDU_UUID { 0x3f3eb880, 0x3639, 0x11ec, \
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+/*
```
etienne-lms, line 12: For commit "core: pta: APDU pseudo trusted application": add an empty line above

etienne-lms (commented):
For commit "crypto: se050: APDU driver": 

Minor comments on few commits.

core/drivers/crypto/se050/core/scp03.c
ldts: it is just that I didnt want to use SE in case it generates confusion with Security Extensions...but I can change it if you think it will be ok

ldts:
hi @etienne-lms I'll wait for a few days for all comments at this point so we can wrap it up. Please let me know when you are done with the review so I push one last time.

core/drivers/crypto/se050/core/scp03.c
jforissier: We have used `_se` for Secure Element in the past, see commit e147a447.

ldts: ok

ldts:
I am not sure anymore what patches have been acked and which haven't since the sequence of changes has been long. Please Would it be possible @etienne-lms/ @jforissier  to confirm them?

core/include/crypto/crypto_se.h
```diff
@@ -15,6 +15,28 @@
...
+	CRYPTO_APDU_CASE_1,
+	CRYPTO_APDU_CASE_2,
+	CRYPTO_APDU_CASE_2E,
+	CRYPTO_APDU_CASE_3,
+	CRYPTO_APDU_CASE_3E,
+	CRYPTO_APDU_CASE_4,
+	CRYPTO_APDU_CASE_4E,
+};
+
+TEE_Result crypto_se_do_apdu(enum crypto_apdu_type type,
```
jenswi-linaro, line 35: This function declared here, but only defined in the next commit. I believe this could lead to a link error while bisecting.

jenswi-linaro (commented):
A commit for the commit "core: pta: APDU pseudo trusted application"

core/include/crypto/crypto_se.h
ldts: um, well ok, will push the se050 driver first...but this should not be enabled without a driver anyway

ldts: what I mean is this: https://github.com/OP-TEE/optee_os/pull/4935/commits/72df9ed590e76c589ba72325be87319596c54fa9 ...now (after I swapped the commits) this commit will be enabling a config that does nothing

ldts: so not sure what is worse TBH...probably what I just pushed is worse but let me know what you prefer

jenswi-linaro: The new order is fine, but I think that the changes in this file belongs in the commit "crypto: se050: APDU driver".

ldts: you are asking me to remove crypto_se.h from this commit?

jenswi-linaro: No, I'm expecting that you provide a declaration for `crypto_se_do_apdu()` when you introduce the global definition `crypto_se_do_apdu()` in the same commit.

ldts: That is totally unnecessary. I'll revert it to the way it was: building the PTA only makes sense when the definition is provided

jenswi-linaro: Noted

ldts: or if you let me know how to to generate the bisect error I'll fix of course! I just dont want to add unnecessary code...OP-TEE is complex as it is....

ldts:
so are we ready to merge? it has been a week with no comments other than todays clarification.

jforissier (commented):
@ldts I have merged the two commits with `Acked-by` tags. The other commits are not reviewed yet.

core/drivers/crypto/se050/crypto.mk
```diff
@@ -38,6 +38,12 @@ ifeq ($(CFG_NXP_SE05X_SCP03_DRV),y)
 $(call force,CFG_SCP03_PTA,y,Mandated by CFG_NXP_SE05X_SCP03)
 endif
 
+# Allow a secure client to send APDU raw frames
+CFG_NXP_SE05X_APDU_DRV ?= y
+ifeq ($(CFG_NXP_SE05X_APDU_DRV),y)
+$(call force,CFG_APDU_PTA,y,Mandated by CFG_NXP_SE05X_APDU)
```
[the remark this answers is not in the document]
ldts: I dont think that it is a better choice; I mean when we add support for another secure element, the commits will not be consistent. 

in one case,  this will have been added in "core: pta: APDU pseudo trusted application"
CFG_APDU_PTA ?= $(CFG_NXP_SE05X_APDU_DRV)
$(eval $(call cfg-depends-all,CFG_APDU_PTA,CFG_NXP_SE05X_APDU_DRV))

and in another case (hypotetical Xilinx device) this would hav e been added in "crypto: xxxx: APDU driver"

CFG_APDU_PTA ?= $(CFG_NXP_XILINX_XXX_APDU_DRV)
$(eval $(call cfg-depends-all,CFG_APDU_PTA,CFG_XILINX_XXX_APDU_DRV))

I think it is better as is. But I can change it if you want

ldts:
But the pta should never be enabled explicitly - and build should fail - unless there is a driver that can handle it. (which is why I defined the makefiles in the way I did, so there is no confusion).

But ok, instead of the two commits, I can do three

1. se050 apdu driver
2. generic apdu pta
3. se050 driver: enable pta

should I do that instead?

core/drivers/crypto/se050/crypto.mk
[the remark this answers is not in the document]
jforissier: In the hypothetical case of another driver for another APDU device, wouldn't we end up with `$(eval $(call cfg-depends-one,CFG_APDU_PTA,CFG_APDU_DRV1 CFG_APDU_DRV2))`?
Anyways it doesn't matter much for now.

jforissier:
> But the pta should never be enabled explicitly - and build should fail - unless there is a driver that can handle it. 

It was the intent of my proposal above https://github.com/OP-TEE/optee_os/pull/4935#discussion_r754901631 but maybe I'm missing something.

> (which is why I defined the makefiles in the way I did, so there is no confusion).
> 
> But ok, instead of the two commits, I can do three
> 
>     1. se050 apdu driver
> 
>     2. generic apdu pta
> 
>     3. se050 driver: enable pta
> 
> 
> should I do that instead?

Let's keep the code that you prefer. Please rebase and add:

to the commits: "core: pta: APDU pseudo trusted application" and "crypto: se050: APDU driver".

core/pta/apdu.c
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
etienne-lms, line 1: would prefer BSD-2-Clause is possible. Not a strong opinion.

etienne-lms (commented):
* "core: crypto: Secure Element cryptographic interface"
* "core: pta: APDU pseudo trusted application"
Also for that commit, fix commit log wrongly state:
"_Allow trusted applications to send APDU frames to a secure element_".
* "crypto: se050: APDU driver"

etienne-lms:
About SE API, there is the TEE SE API that was implemented long ago, removed from e147a447ccc06a6eca7caeddae22dec06c168d19. It implements the API exposed the TAs.
It included _tee/se/iso7816.h_, _tee/se/apdu.h_, a tee_se_manager with reads and channels.
This work is maybe better suited as generic definitions of SE API materials.

ldts:
> 

um, the SE050 could have used that - instead of having the Plug-and-Trust in a separate repository as we do now (https://github.com/foundriesio/plug-and-trust). 

Rewritting the se050 work to those interfaces now is not a small task (> ~6mm if I had to plan for it). I just didnt know it existed when I did the work so I approached the secure element more like a crypto device (like CAAM) which is why the SE stuff is mostly in the separate repository above.

I suppose we could bring the API back and progressively move towards it....but is a long term investment. what do you think?

etienne-lms (changes requested):
Please update commit message for commit "core: pta: APDU pseudo trusted application":
```patch
-Allow trusted applications to send APDU frames to a secure
element.
+Allow trusted applications and REE clients to send APDU frames to a secure
element.

**Comment by ldts** [comment]
@etienne-lms I think this also belongs in the crypto_se.h.

https://github.com/foundriesio/optee_os/blob/3.10%2Bfio/core/tee/tee_svc_storage.c#L530 

Basically when pkcs#11 stores a handle to a key (ie the key is in a secure element) instead of a key, the only way I could come up with a way to delete this key is by searching for it in the storage. 

Implementation detail aside, I think this case is generic to other SEs - so perhaps the interface belongs also the crypto_se.h ...what do you think?

File: core/drivers/crypto/se050/adaptors/utils/utils.c [hunk 27]
```diff
@@ -51,15 +54,20 @@ static void delete_transient_objects(void)
 			if (id >= OID_MAX || id == 0)
 				continue;
 
-			if (id & BIT(0))
+			status = Se05x_API_ReadType(ctx, id, &type, &mode,
+						    attestation);
```
jforissier, line 58: Doesn't this fit on one line?

jforissier (commented):
For commit "crypto: drivers: se050: object identifier persistence property":

One nit.

core/drivers/crypto/se050/adaptors/utils/utils.c
ldts: ok.

PR #4936, opened 2021-10-22
drivers: crypto: ecc/rsa/dsa: input parameter validation

ldts:
drivers: crypto: ecc/rsa/dsa: input parameter validation
To comply with the PKCS#11 convention for functions returning output
in a variable-length buffer, prefer to check the required size of the
output buffer before the existence of the output buffer itself.

This will save callers from having to allocate a buffer that might not
be used.


cneveux:
It's good for me.
`Acked-by: Cedric Neveux <cedric.neveux@nxp.com>`

PR #4938, opened 2021-10-25
[RFC] Proof of concept run-time integrity check

jenswi-linaro:
This is a proof of concept implementation for https://linaro.atlassian.net/browse/TS-138 (see a comment in this ticked for a kernel patch for some testing) and https://linaro.atlassian.net/browse/TS-120.

Here's patch for the Linux kernel OP-TEE driver to exercise this code a bit.
[0001-optee-silly-run-time-integrity-test-at-probe.patch.gz](https://github.com/OP-TEE/optee_os/files/7409690/0001-optee-silly-run-time-integrity-test-at-probe.patch.gz)

I'm creating this PR to discuss this and to see if it might be useful for normal world.

### DIFF
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -2495,6 +2495,8 @@ void *core_mmu_map_rti_check(paddr_t pa, size_t len, size_t *mapped_len)
 	uint32_t attr = 0;
 	size_t idx = 0;
 
+	DMSG("%#"PRIxPA":%#zx", pa, len);
+
 	/* Refuse to map memory which isn't know to be non-secure */
 	if (!core_pbuf_is(CORE_MEM_NON_SEC, pa, len))
 		return NULL;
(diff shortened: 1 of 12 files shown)


core/arch/arm/mm/core_mmu.c
```diff
@@ -2474,3 +2481,98 @@ void core_mmu_init_ta_ram(void)
...
+	if (attr & (TEE_MATTR_PR | TEE_MATTR_TABLE)) {
+		core_mmu_set_entry(&tbl_info, idx, 0, 0);
+		tlbi_mva_range(vabase, CORE_MMU_PGDIR_SIZE, old_granule);
+	}
+	if (!len)
+		return NULL;
+	/* We cannot map anything smaller than a page */
+	if ((pa | len) & SMALL_PAGE_MASK)
+		return NULL;
+
```
apalos, line 2533: Those two could be moved at the start

jenswi-linaro: I'd like a `NULL` return to mean that no memory is mapped any longer. That way there's no need for extra calls to make sure nothing is mapped.

PR #4942, opened 2021-10-25
[RFC] RAM optimized clk framework : reduce struct clk size : orphan clocks

etienne-lms:
This RFC P-R proposes changes in the clock framework to reduce the memory footprint of generic clock driver.
I propose an alternate implementation: see RFC P-R https://github.com/OP-TEE/optee_os/pull/4943.

This change proposes to define 2 clock alternate structures: full-fledged clocks (`struct clk_elt`) and orphan clocks (`struct clk`, no parent info). Clock drivers register either of these with properly bound operators depending on clock type.

This RFC start with 2 preparatory changes (same as in RFC P-R https://github.com/OP-TEE/optee_os/pull/4943) followed by orphan clock implementation over the 2 commits:
* "drivers: clk: prepare for orphan clocks"
* "drivers: clk: introduce orphan clocks"

As illustration, the 3 last commits port fixed clocks driver and stm32mp1 clock driver to orphan clocks:
* "drivers: clk: move fixed clock to orphan clocks"
* "[PR#4940] core: dt: add platform data per compatible identifier"
* "plat-stm32mp1: move clock driver to clock framework"

The goal for stm32mp1 platform is to move to the generic clock API, yet considering it runs with `CFG_PAGER=y` and current `struct clk` consumes a bit to much memory.

Comments are welcome.

### DIFF
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -32,10 +32,7 @@ struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 	struct clk_elt *clk_elt = NULL;
 	size_t parent = 0;
 
-	if (ops->id != CLK_OPS_ELT) {
-		EMSG("Panic on unexpected ops ID %u", ops->id);
-		panic();
-	}
+	assert(ops && ops->id == CLK_OPS_ELT);
 
 	clk_elt = calloc(1, sizeof(*clk_elt) +
 			 parent_count * sizeof(*parent_clks));
@@ -57,10 +54,7 @@ struct clk *clk_alloc_orphans(const struct clk_ops *ops, size_t count)
 	struct clk *clk = NULL;
 	size_t n = 0;
 
-	if (ops->id != CLK_OPS_ORPHAN) {
-		EMSG("Panic on unexpected ops ID %u", ops->id);
-		panic();
-	}
+	assert(ops && ops->id == CLK_OPS_ORPHAN);
 
 	clk = calloc(count, sizeof(*clk));
 	if (!clk)
@@ -195,9 +189,6 @@ static void clk_disable_no_lock(struct clk *clk)
 	if (clk->ops->disable)
 		clk->ops->disable(clk);
 
-	if (is_clk_orphan(clk))
-		return;
-
 	parent = clk_get_parent(clk);
 	if (parent)
 		clk_disable_no_lock(parent);
@@ -211,13 +202,11 @@ static TEE_Result clk_enable_no_lock(struct clk *clk)
 	if (refcount_inc(&clk->enabled_count))
 		return TEE_SUCCESS;
 
-	if (is_clk_elt(clk)) {
-		parent = clk_get_parent(clk);
-		if (parent) {
-			res = clk_enable_no_lock(parent);
-			if (res)
-				return res;
-		}
+	parent = clk_get_parent(clk);
+	if (parent) {
+		res = clk_enable_no_lock(parent);
+		if (res)
+			return res;
 	}
 
 	if (clk->ops->enable) {
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -101,9 +101,15 @@ TEE_Result clk_register(struct clk *clk)
 	clk_init_parent(clk);
 	clk_compute_rate_no_lock(clk);
 
-	DMSG("Registered clock %s, freq %lu", clk->name, clk_get_rate(clk));
+	DMSG("Registered clock %s, freq %lu", clk_get_name(clk),
+	     clk_get_rate(clk));
 
 	return TEE_SUCCESS;
+}
+
+const char *clk_get_name(struct clk *clk)
+{
+	return clk->name;
 }
 
 static bool clk_is_enabled_no_lock(struct clk *clk)
@@ -240,7 +246,9 @@ static TEE_Result clk_get_parent_idx(struct clk *clk, struct clk *parent,
 			return TEE_SUCCESS;
 		}
 	}
-	EMSG("Clock %s is not a parent of clock %s", parent->name, clk->name);
+
+	EMSG("Clock %s is not a parent of clock %s", clk_get_name(parent),
+	     clk_get_name(clk);
 
 	return TEE_ERROR_BAD_PARAMETERS;
 }
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -17,22 +17,28 @@ static unsigned int clk_lock = SPINLOCK_UNLOCK;
 struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		      struct clk **parent_clks, size_t parent_count)
 {
-	struct clk *clk = NULL;
+	struct clk_elt *clk_elt = NULL;
 	size_t parent = 0;
 
-	clk = calloc(1, sizeof(*clk) + parent_count * sizeof(clk));
-	if (!clk)
+	if (ops->id != CLK_OPS_ELT) {
+		EMSG("Panic on unexpected ops ID %u", ops->id);
+		panic();
+	}
+
+	clk_elt = calloc(1, sizeof(*clk_elt) +
+			 parent_count * sizeof(*parent_clks));
+	if (!clk_elt)
 		return NULL;
 
-	clk->num_parents = parent_count;
+	clk_elt->num_parents = parent_count;
 	for (parent = 0; parent < parent_count; parent++)
-		clk->parents[parent] = parent_clks[parent];
+		clk_elt->parents[parent] = parent_clks[parent];
 
-	clk->name = name;
-	clk->ops = ops;
-	refcount_set(&clk->enabled_count, 0);
+	clk_elt->name = name;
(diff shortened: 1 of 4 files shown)


core/drivers/clk/clk.c
```diff
@@ -119,17 +138,21 @@ TEE_Result clk_register(struct clk *clk)
...
 }
 
 const char *clk_get_name(struct clk *clk)
 {
-	if (clk_is_clk_elt(clk))
+	if (clk_is_clk_elt(clk) && clk_to_clk_elt(clk)->name)
 		return clk_to_clk_elt(clk)->name;
 
+	if (clk->ops->get_name)
+		return clk->ops->get_name(clk);
```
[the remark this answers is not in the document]
etienne-lms: not addressed yet.

etienne-lms:
Added 2 fixup commits to clarify how to retrieve a clock name.

core/drivers/clk/clk.c
```diff
@@ -14,23 +14,61 @@
 /* Global clock tree lock */
 static unsigned int clk_lock = SPINLOCK_UNLOCK;
 
+void clk_init_instance(struct clk *clk, const struct clk_ops *ops)
+{
+	assert(clk && is_clk_elt(clk) || is_clk_orphan(clk) && ops);
```
ldebieve, line 19: assert is incorrect as you may intend to access a clk->ops that is not yet initialized in is_clk_elt() and is_clk_orphan.

core/drivers/clk/clk.c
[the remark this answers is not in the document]
etienne-lms: addressed in https://github.com/OP-TEE/optee_os/pull/4942/commits/3e89ec12b097109b8a6c207963a66dcd16ea5fc8

core/drivers/clk/clk.c
etienne-lms: indeed something's wrong here. thanks.

etienne-lms: addressed in fixup commit.

core/drivers/clk/clk.c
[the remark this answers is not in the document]
jenswi-linaro: I'm sorry, this isn't what I was looking for. This function still contains special treatment for struct clk_elt. I'd rather that struct clk_elt always provided a `get_name()`. If that function just returns `clk_to_clk_elt(clk)->name`, that's fine the point is that we shouldn't care here.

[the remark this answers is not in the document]
etienne-lms: Would this [fixup commit](https://github.com/OP-TEE/optee_os/pull/4942/commits/532e0867f3fd2095032588b6410d965ccf2c22dd) do the job?
The _ugly_ part, IMO, is that it mandates a specific `get_name` operator for `clk_elt` instances,
instead of _ugly_ routing `clk_get_name()` according to the clock type.

core/drivers/clk/clk.c
```diff
@@ -169,17 +169,19 @@ TEE_Result clk_register(struct clk *clk)
...
-	/* clk_elt instance may have been registered with a name */
-	if (is_clk_elt(clk))
-		return clk_to_clk_elt(clk)->name;
-
 	return NULL;
 }
 
+const char *clk_elt_name(struct clk *clk)
+{
+	struct clk_elt *clk_elt = clk_to_clk_elt(clk);
```
jenswi-linaro, line 180: Or even
``` C
return clk_to_clk_elt(clk)->name;
```

core/drivers/clk/clk.c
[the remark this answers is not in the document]
jenswi-linaro: It's an improvement. I have a couple of comments though. :-)

core/drivers/clk/clk.c
```diff
@@ -37,8 +37,8 @@ struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		panic();
 	}
 
-	if (name && ops->get_name) {
-		EMSG("Operator get_name and name are exclusive");
+	if (ops->get_name != clk_elt_name) {
```
jenswi-linaro, line 40: Why do we care?

etienne-lms: I though about consistency in argument  `name` of this function and what `clk_get_name()` will return. However, I agree it's not a real issue, I can remove this test.

core/drivers/clk/clk.c
etienne-lms: true :)

etienne-lms:
2 last comments addressed.
@jenswi-linaro, I suggest I squash the fixup commits and we go into a new review round. Agree?

jenswi-linaro:
Sounds good, please squash the fixups.

etienne-lms:
Squashed back to 5 commits.

core/drivers/clk/clk.c
```diff
@@ -235,6 +235,14 @@ struct clk *clk_get_parent(struct clk *clk)
 	return clk->parent;
 }
 
+size_t clk_get_num_parents(struct clk *clk)
+{
+	if (clk_is_clk_elt(clk))
```
jenswi-linaro: struct clk_elt and friends are not available yet, they are introduced in the next commit.

jenswi-linaro (commented):
Comment for "drivers: clk: move clk_get_num_parents() implement in C source"

core/drivers/clk/clk.c
```diff
@@ -17,22 +17,28 @@ static unsigned int clk_lock = SPINLOCK_UNLOCK;
 struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		      struct clk **parent_clks, size_t parent_count)
 {
-	struct clk *clk = NULL;
+	struct clk_elt *clk_elt = NULL;
 	size_t parent = 0;
 
-	clk = calloc(1, sizeof(*clk) + parent_count * sizeof(clk));
-	if (!clk)
+	if (ops->id != CLK_OPS_ELT) {
```
jenswi-linaro, line 35: I suppose an assert to help while debugging could be justified, but this seems a bit much to me.

jenswi-linaro (commented):
Comment for "drivers: clk: prepare for orphan clocks"

core/drivers/clk/clk.c
etienne-lms: the function registers a clk_elt instance. If not of the right type, clk_alloc() should fail. Since it provides or errror code, returning NULL would tends to report heap exhaustion. I'll switch to an assertion.

core/drivers/clk/clk.c
etienne-lms: I've miss something when squashing. I'll fix that at next squash party if you're ok.

jenswi-linaro (commented):
Comments for "drivers: clk: introduce orphan clocks"

core/drivers/clk/clk.c
```diff
@@ -190,16 +223,23 @@ void clk_disable(struct clk *clk)
 
 unsigned long clk_get_rate(struct clk *clk)
 {
-	return clk->rate;
+	if (is_clk_elt(clk))
```
jenswi-linaro, line 261: Another comment for "drivers: clk: prepare for orphan clocks"
I'd rather use the `get_rate()` callback for this.

core/drivers/clk/clk.c
jenswi-linaro: Fine. :-)

core/drivers/clk/clk.c
etienne-lms: true. make sense. thanks.

etienne-lms: soory. i"m wrong. `get_rate` is an internal call asking driver to return the effective clock based on platform clock state and a provided (possible) parent rate. we would need 2 operators, it would make how clk_elt are defined more complex: would mandate platform clock does register the generic handler, as done for `get_name`.

jenswi-linaro:
I think we should strive to make the wrapper functions around the `ops` callbacks to be generic in the sense that they don't care what type of `struct clk` it's working on. This would make some of the wrapper functions just a return statement, those should preferably be `static inline` instead.

We use this approach at other places and it seems it could work here too.
If you agree please squash in the needed fixups in the commits and we can restart the review.

core/drivers/clk/clk.c
jenswi-linaro: The benefit would be a smaller code footprint for the orphan clocks.

etienne-lms:
`static inline` function can't be referenced by `ops`, do they?

jenswi-linaro:
They could actually, but it would be a bit unusual. But I was not referring to those function, it's `clk_get_name()` and friends, all those that does not much more than call the callback function.

etienne-lms:
Oh, yes, true. Those `clk_xxx()` API functions themselves would be inlined.
I'm getting the picture. It brings a sligtly different patch series. I'll create another P-R rather than fixup commits in this series.
Working at it, the big ugly "prepare orphan" will split in incremental changes, (if i do it right).

etienne-lms:
I've created https://github.com/OP-TEE/optee_os/pull/4978 with a new patch series for orphan clocks, renamed raw clocks.

PR #4943, opened 2021-10-25
[RFC] RAM optimized clk framework : weak functions

etienne-lms:
This RFC P-R proposes changes in the clock framework to reduce the memory footprint of generic clock driver.
I propose an alternate implementation: see RFC P-R https://github.com/OP-TEE/optee_os/pull/4942.

This idea of this P-R is to use clk driver API functions prototypes but not the implementation which is fully provided by a unique platform specific clock driver.

This RFC start with 2 preparatory changes, followed by a commit to introduce `_CFG_DRIVERS_CLK_WEAK`.
The 2 last commit is a port on stm32mp1 clock driver, using #4940. The goal for this platform is to move to the generic clock API, yet considering it runs with `CFG_PAGER=y`. 

Comments are welcome.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/drivers/stm32mp1_clk.c
+++ b/core/arch/arm/plat-stm32mp1/drivers/stm32mp1_clk.c
@@ -4,6 +4,9 @@
  */
 
 #include <assert.h>
+#include <config.h>
+#include <drivers/clk.h>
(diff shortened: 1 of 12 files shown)


etienne-lms:
I'm closing the issue. On can follow up from https://github.com/OP-TEE/optee_os/pull/4942.

PR #4949, opened 2021-10-27
Proposal for a driver probing sequence

etienne-lms:
**(edited atfer #4965 merge)**

This RFC proposes a unified driver probing sequence for compatible drivers registered as dt_driver. The main goal of this unified driver probing implementation is to resolve device driver dependencies without requiring specific static initcall levels.

This changes introduce implementation in _core/kernel/dt_drivers.c_, executed at `driver_init` initcall level. The function parses DT nodes and probes registered compatible drivers. Driver probe function can register as a device driver provider. Driver probe function can query device(s) to the driver provider which may in turn request caller probe to be deferred when the target device is not yet registered in the driver provider database.

**Preparatory changes:**

~~This scheme reuses and adapts the implementation from clk framework (device driver probing, registration and device queries) to propose a generic device driver model.~~
~~Commit _"core: factorize dt_driver probing helper function"_ moves some functions from _clk_dt.c_ to _dt_driver.c_.~~
_**(e)** addressed by #4965_

In this P-R (at least its initial state), ~~`TEE_ERROR_BUSY`~~ **New `TEE_ERROR_DEFER_DRIVER_INIT`** is used as return code from a driver probe function when the driver is waiting for device to be probed/initialized.
See commit _"core: dt_driver: get return code when querying a device"_.

**Driver probing loop**

The unified driver probing sequence is implemented in commit _"core: dt_driver probe sequence"_.
The implementation uses link lists to track DT nodes for which we want to probe a compatible driver, the devices that have been successfully initialized and initialized/registered device provider drivers.
The probe loop starts by scanning each subnode of the FDT root node. Compatible drivers, when probed, may add new nodes to the to-be-probed list. As far as nodes are added and drivers are probed, driver dependencies should be resolved.
The loop states that some dependencies can't be resolved if all remaining to-be-probed devices request probe deferral while no new node is added in the to-be-probed list.

Commit _"core: dt_driver: add simple bus driver"_ implement a simple bus driver. Simple bus is related to nodes for which each subnode is to be added to the to-be-probed list, typically used in `soc` node description (i.e [fsl-lx2160a](https://github.com/OP-TEE/optee_os/blob/3.15.0/core/arch/arm/dts/fsl-lx2160a.dtsi#L464-L465) and [stm32mp15](https://github.com/OP-TEE/optee_os/blob/3.15.0/core/arch/arm/dts/stm32mp151.dtsi#L110-L111))

**Add-on: post-init resource release**

3 last commits proposes an initcall specific level (`release_init_resource`) for releasing all resources needed only during core initialization. The driver probing model implementation uses it to release of memory allocated for resolving device dependencies. As an example, se050 driver is updated to leverage this new initcall level, when core initialization is completed.

### DIFF
--- a/core/drivers/clk/clk_dt.c
+++ b/core/drivers/clk/clk_dt.c
@@ -14,31 +14,34 @@
 #include <stddef.h>
 
 struct clk *clk_dt_get_by_name(const void *fdt, int nodeoffset,
-			       const char *name)
+			       const char *name, TEE_Result *res)
 {
 	int clk_id = 0;
 
 	clk_id = fdt_stringlist_search(fdt, nodeoffset, "clock-names", name);
-	if (clk_id < 0)
+	if (clk_id < 0) {
+		*res = TEE_ERROR_GENERIC;
 		return NULL;
+	}
 
-	return clk_dt_get_by_idx(fdt, nodeoffset, clk_id);
+	return clk_dt_get_by_idx(fdt, nodeoffset, clk_id, res);
 }
 
 static struct clk *clk_dt_get_by_idx_prop(const char *prop_name,
 					  const void *fdt, int nodeoffset,
-					  unsigned int clk_idx)
+					  unsigned int clk_idx, TEE_Result *res)
 {
 	void *device = dt_driver_device_from_node_idx_prop(prop_name, fdt,
-							   nodeoffset, clk_idx);
+							   nodeoffset, clk_idx,
+							   res);
 
 	return (struct clk *)device;
 }
 
 struct clk *clk_dt_get_by_idx(const void *fdt, int nodeoffset,
-			      unsigned int clk_idx)
+			      unsigned int clk_idx, TEE_Result *res)
 {
-	return clk_dt_get_by_idx_prop("clocks", fdt, nodeoffset, clk_idx);
+	return clk_dt_get_by_idx_prop("clocks", fdt, nodeoffset, clk_idx, res);
 }
 
 /* Recursively called from parse_clock_property() */
@@ -134,6 +137,7 @@ static void parse_assigned_clock(const void *fdt, int nodeoffset)
 	unsigned long rate = 0;
 	struct clk *parent = NULL;
 	const uint32_t *rate_prop = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
 
 	rate_prop = fdt_getprop(fdt, nodeoffset, "assigned-clock-rates",
 				&rate_len);
@@ -141,12 +145,12 @@ static void parse_assigned_clock(const void *fdt, int nodeoffset)
 
 	while (1) {
 		clk = clk_dt_get_by_idx_prop("assigned-clocks", fdt, nodeoffset,
-					     clock_idx);
+					     clock_idx, &res);
 		if (!clk)
 			return;
 
 		parent = clk_dt_get_by_idx_prop("assigned-clock-parents", fdt,
-						nodeoffset, clock_idx);
+						nodeoffset, clock_idx, &res);
 		if (parent) {
 			if (clk_set_parent(clk, parent)) {
 				EMSG("Could not set clk %s parent to clock %s",
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -146,7 +146,7 @@ static TEE_Result dt_driver_release_provider(void)
 	return TEE_SUCCESS;
 }
 
-driver_init_late(dt_driver_release_provider);
+release_init_resource(dt_driver_release_provider);
 
 /*
  * Helper functions for dt_drivers querying driver provider information
@@ -651,6 +651,30 @@ static TEE_Result probe_dt_drivers(void)
 }
 
 driver_init(probe_dt_drivers);
+
+static TEE_Result release_probe_lists(void)
+{
+	struct dt_driver_probe *elt = NULL;
+	struct dt_driver_probe *next = NULL;
+	const void * __maybe_unused fdt = NULL;
+
+	if (!IS_ENABLED(CFG_EMBED_DTB))
+		return TEE_SUCCESS;
+
+	fdt = get_embedded_dt();
+
+	assert(fdt && TAILQ_EMPTY(&dt_driver_probe_list));
+
+	TAILQ_FOREACH_SAFE(elt, &dt_driver_ready_list, link, next) {
+		DMSG("element: %s on node %s", elt->dt_drv->name,
+		     fdt_get_name(fdt, elt->nodeoffset, NULL));
+		free(elt);
+	}
+
+	return TEE_SUCCESS;
+}
+
+release_init_resource(release_probe_lists);
 
 /*
  * Simple bus support: handy to parse subnodes
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -151,7 +151,8 @@ struct dt_driver_provider *dt_driver_get_provider_by_phandle(uint32_t phandle)
 }
 
 static void *device_from_provider_prop(struct dt_driver_provider *prv,
-				       const uint32_t *prop)
+					  const uint32_t *prop,
+					  TEE_Result *res)
 {
 	struct dt_driver_phandle_args *pargs = NULL;
 	unsigned int n = 0;
@@ -159,14 +160,16 @@ static void *device_from_provider_prop(struct dt_driver_provider *prv,
 
 	pargs = calloc(1, prv->provider_cells * sizeof(uint32_t *) +
 		       sizeof(*pargs));
-	if (!pargs)
+	if (!pargs) {
+		*res = TEE_ERROR_OUT_OF_MEMORY;
 		return NULL;
+	}
 
 	pargs->args_count = prv->provider_cells;
 	for (n = 0; n < prv->provider_cells; n++)
 		pargs->args[n] = fdt32_to_cpu(prop[n + 1]);
 
-	device = prv->get_of_device(pargs, prv->priv_data);
+	device = prv->get_of_device(pargs, prv->priv_data, res);
 
 	free(pargs);
 
@@ -175,7 +178,8 @@ static void *device_from_provider_prop(struct dt_driver_provider *prv,
 
 void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 					  const void *fdt, int nodeoffset,
-					  unsigned int prop_idx)
+					  unsigned int prop_idx,
+					  TEE_Result *res)
 {
 	int len = 0;
 	int idx = 0;
@@ -186,16 +190,20 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 	struct dt_driver_provider *prv = NULL;
 
 	prop = fdt_getprop(fdt, nodeoffset, prop_name, &len);
-	if (!prop)
+	if (!prop) {
+		*res = TEE_ERROR_GENERIC;
 		return NULL;
+	}
 
 	while (idx < len) {
 		idx32 = idx / sizeof(uint32_t);
 		phandle = fdt32_to_cpu(prop[idx32]);
 
 		prv = dt_driver_get_provider_by_phandle(phandle);
-		if (!prv)
+		if (!prv) {
+			*res = TEE_ERROR_GENERIC;
 			return NULL;
+		}
 
 		prv_cells = dt_driver_provider_cells(prv);
 		if (prop_idx) {
@@ -204,9 +212,10 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 			continue;
 		}
 
-		return device_from_provider_prop(prv, prop + idx32);
+		return device_from_provider_prop(prv, prop + idx32, res);
 	}
 
+	*res = TEE_ERROR_GENERIC;
 	return NULL;
 }
 
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -2,15 +2,40 @@
 /*
  * Copyright (c) 2021, Linaro Limited
  * Copyright (c) 2021, Bootlin
+ * Copyright (c) 2021, Linaro Limited
+ * Copyright (c) 2021, STMicroelectronics
  */
 
+#include <assert.h>
+#include <config.h>
 #include <initcall.h>
+#include <kernel/boot.h>
 #include <kernel/dt.h>
 #include <kernel/dt_driver.h>
 #include <libfdt.h>
 #include <malloc.h>
 #include <sys/queue.h>
+#include <tee_api_defines_extensions.h>
 #include <tee_api_types.h>
+
+/*
+ * struct dt_driver_probe - Node instance in secure FDT to probe a driver for
+ *
+ * @link: List hook
+ * @nodeoffset: Node offset of device referenced in the FDT
+ * @type: One of DT_DRIVER_* or DT_DRIVER_NOTYPE.
+ * @deferrals: Driver probe deferrals count
+ * @dt_drv: Matching driver to probe if found or NULL
+ * @dm: Matching reference if applicable or NULL
+ */
+struct dt_driver_probe {
+	int nodeoffset;
+	enum dt_driver_type type;
+	unsigned int deferrals;
+	const struct dt_driver *dt_drv;
+	const struct dt_device_match *dm;
+	TAILQ_ENTRY(dt_driver_probe) link;
+};
 
 /*
  * struct dt_driver_provider - DT related info on probed device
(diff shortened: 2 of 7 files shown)


etienne-lms:
Rebased series and introduce a vendor specific `TEE_Result` code for reporting driver probing deferral requests:
`TEE_RESULT_DEPENDENT_DRIVER_NOT_INITIALIZED`, instead of using `TEE_ERROR_BUSY`.
(maybe shorter name `TEE_RESULT_DEFER_DRIVER_INIT` would be preferred?).

Still todo: embed a dummy DTB in qemu platforms to test the driver probing sequence.

We discussed how the order in which drivers are initialized (successfully probed) could be used to definesuspend/resume sequence order. This topic is addressed by P-R https://github.com/OP-TEE/optee_os/pull/4960.

jenswi-linaro:
> Rebased series and introduce a vendor specific `TEE_Result` code for reporting driver probing deferral requests: `TEE_RESULT_DEPENDENT_DRIVER_NOT_INITIALIZED`, instead of using `TEE_ERROR_BUSY`. (maybe shorter name `TEE_RESULT_DEFER_DRIVER_INIT` would be preferred?).

Yes, a slightly short name would be preferred. Perhaps `TEE_ERROR_DEFER_DRIVER_INIT`?

jforissier:
> Yes, a slightly short name would be preferred. Perhaps `TEE_ERROR_DEFER_DRIVER_INIT`?

+1 or even `TEE_ERROR_DEFER` (do we need to specify what needs to be deferred?)

jenswi-linaro:
> > Yes, a slightly short name would be preferred. Perhaps `TEE_ERROR_DEFER_DRIVER_INIT`?
> 
> +1 or even `TEE_ERROR_DEFER` (do we need to specify what needs to be deferred?)

Sure, the reason I suggested `TEE_ERROR_DEFER_DRIVER_INIT` was to have it stand out a bit. But `TEE_ERROR_DEFER` is OK too.

etienne-lms:
> Sure, the reason I suggested TEE_ERROR_DEFER_DRIVER_INIT was to have it stand out a bit. But TEE_ERROR_DEFER is OK too.

We specifically defer driver initialization hence so I prefer an explicit definition `TEE_ERROR_DEFER_DRIVER_INIT`.

I've use `TEE_RESULT_`  as prefix (not `TEE_ERROR_`) because it is actually not an error, and the Id is chosen in `non-error information` Id range specified by GTD TEE Int. Core API (v1.1).
`TEE_RESULT_` ?  `TEE_ERROR_`? (in favor for the former).

jforissier:
> because it is actually not an error

When the probe function returns this code, it has failed to accomplish the requested operation. That's what I would call an error :wink:

etienne-lms:
> > because it is actually not an error
>
> When the probe function returns this code, it has failed to accomplish the requested operation. That's what I would call an error :wink:

+1  :)

etienne-lms:
Please have i look at https://github.com/OP-TEE/optee_os/pull/4965
I've created this #4965 to ease review of the 2 first commits of this RFC series:
* "core: factorize dt_driver probing helper function"   (split inot small commits to ease review)
* "core: dt_driver: move probe callback to dt_driver"

etienne-lms:
Rebased over merged #4965 and using `TEE_ERROR_DEFER_DRIVER_INIT` extended `TEE_Result` code.

lib/libutee/include/tee_api_defines_extensions.h
```diff
@@ -1,12 +1,22 @@
...
 
+/*
+ * API extended result codes as per TEE_Result IDs defined in GPD TEE
+ * Internal Core API specification v1.1:
+ *
+ * 0x70000000 - 0x7FFFFFFF: Reserved for implementation-specific return
+ *			    code providing non-error information
+ * 0x80000000 - 0x8FFFFFFF: Reserved for implementation-specific errors
+ */
+#define TEE_ERROR_DEFER_DRIVER_INIT	0x70000000
```
jenswi-linaro, line 18: It no big deal, but in the commit message you describe this as an indication of a failure, but you have selected a value in the non-error range.

core/drivers/clk/clk_dt.c
```diff
@@ -134,20 +137,21 @@ static void parse_assigned_clock(const void *fdt, int nodeoffset)
...
 	rate_prop = fdt_getprop(fdt, nodeoffset, "assigned-clock-rates",
 				&rate_len);
 	rate_len /= sizeof(uint32_t);
 
 	while (1) {
 		clk = clk_dt_get_by_idx_prop("assigned-clocks", fdt, nodeoffset,
-					     clock_idx);
-		if (!clk)
+					     clock_idx, &res);
+		if (res || !clk)
```
jenswi-linaro, line 149: Checking `clk` should be enough. We should probably not even look at `res` unless `clk` is `NULL`.

core/drivers/clk/clk_dt.c
```diff
@@ -134,20 +137,21 @@ static void parse_assigned_clock(const void *fdt, int nodeoffset)
...
 			return;
 
 		parent = clk_dt_get_by_idx_prop("assigned-clock-parents", fdt,
-						nodeoffset, clock_idx);
-		if (parent) {
+						nodeoffset, clock_idx, &res);
+		if (!res && parent) {
```
jenswi-linaro, line 154: Same here, checking `parent` should be enough.

jenswi-linaro (commented):
Comments for "core: dt_driver: get return code when querying a device"

core/drivers/clk/clk_dt.c
etienne-lms: right

core/drivers/clk/clk_dt.c
etienne-lms: ok

lib/libutee/include/tee_api_defines_extensions.h
etienne-lms: Right. From previous discussion, we said the code reported a error case, so I'll change to `0x80000000`.

core/kernel/dt_driver.c
```diff
@@ -266,3 +296,387 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+			     fdt_get_name(fdt, subnode, NULL), res);
+			panic();
+		}
+
+		/* Rescursively parse the FDT, skipping disabled nodes */
+		if (IS_ENABLED(CFG_DRIVERS_DT_RECURSIVE_PROBE)) {
+			if (_fdt_get_status(fdt, subnode) == DT_STATUS_DISABLED)
+				continue;
+
+			parse_node(fdt, subnode);
```
jenswi-linaro, line 662: How deep will this go, worst case?

jenswi-linaro (commented):
Comments for the commit "core: dt_driver probe sequence"

lib/libutee/include/tee_api_defines_extensions.h
```diff
@@ -14,8 +14,11 @@
  * 0x70000000 - 0x7FFFFFFF: Reserved for implementation-specific return
  *			    code providing non-error information
  * 0x80000000 - 0x8FFFFFFF: Reserved for implementation-specific errors
+ *
+ * TEE_ERROR_DEFER_DRIVER_INIT - Device driver failed to initialize because
+ *	it depends on a device not yet initialized.
```
jenswi-linaro, line 19: What's "it" supposed to align with? It seems we have two choices, either "TEE_ERROR_DEFER_DRIVER_INIT" or "Device" at the row above.

jenswi-linaro (commented):
For the commit "libutee: define error core TEE_ERROR_DEFER_DRIVER_INIT"
with the indentation issue sorted:

jenswi-linaro (commented):
For the commit "core: dt_driver: get return code when querying a device" with fixup:

jenswi-linaro:
For commits "core: add initcall level 7 to release initialization resources" and "core: release driver probing resource after initialization"

jenswi-linaro (commented):
For commit "drivers: se050: use release_init_resource to enable i2c trampoline"

This is messy.

You say:
Release_init_resource is a better initcall level since it is called once other initialization are completed as per this initcall level name.

I say:
boot_final is better because I2C driver is after this point forbidden from tinkering directly with the hardware since normal world now owns the device.

I think we can agree that this will regardless not work with virtualization.

core/kernel/dt_driver.c
```diff
@@ -266,3 +296,387 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+ * probing drivers for device node. Non matching elements are removed during
+ * that DT nodes parsing loop.
+ *
+ * @added_count: increments when a new element is added to the list
+ */
+struct probe_list_info {
+	unsigned int added_count;
+};
+
+static struct probe_list_info probe_list_info;
```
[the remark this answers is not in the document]
etienne-lms: I get your point. Here, the count is not used to really count items in probe_list. It is used to detect if nodes to be probed were added in probe_list while some driver(s) defer their probing.
I could make this move explicit, replacing this count with a boolean, reset from `process_probe_list()` (before running a new round of probing) and set from `add_node_to_probe()` (when a node is added to probe_list):

```c
static bool node_added;

static TEE_Result add_node_to_probe(...)
{
	(...)
	// upon successful add of the node ref in probe_list:
	node_added = true;
	return TEE_SUCCESS;
}

static TEE_Result process_probe_list(const void *fdt)
{
	while (1 /* continue probing drivers for nodes in probe_list */) {
		/* Reset flag indicating new nodes where added in probe_list */
		node_added = false;

		/* possibly probe drivers */
		(...)

		/* If all remaining probes are deferred and no new driver to probe, we're stuck  */
		if (all_deferred && !node_added)
			goto err_missing_deps;
	}

}

File: core/kernel/dt_driver.c [hunk 7]
```diff
@@ -266,3 +296,387 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+
+static TAILQ_HEAD(dt_driver_probe_head, dt_driver_probe) dt_driver_probe_list =
+	TAILQ_HEAD_INITIALIZER(dt_driver_probe_list);
+
+static TAILQ_HEAD(, dt_driver_probe) dt_driver_ready_list =
+	TAILQ_HEAD_INITIALIZER(dt_driver_ready_list);
+
+static void add_to_ready_list(struct dt_driver_probe *elt)
```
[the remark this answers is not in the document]
etienne-lms: not really, I can remove that.

core/kernel/dt_driver.c
```diff
@@ -266,3 +296,387 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+
+	TAILQ_FOREACH(elt, &dt_driver_probe_list, link)
+		DMSG("- Driver %s probes on node %s",
+		     elt->dt_drv->name,
+		     fdt_get_name(fdt, elt->nodeoffset, NULL));
+
+	DMSG("Probe list end");
+}
+
+static bool is_already_in_probe_list(struct dt_driver_probe *candidate)
```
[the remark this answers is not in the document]
etienne-lms: +1

core/kernel/dt_driver.c
```diff
@@ -266,3 +296,387 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+}
+
+static TEE_Result process_probe_list(const void *fdt)
+{
+	struct dt_driver_probe *elt = NULL;
+	struct dt_driver_probe *prev = NULL;
+	unsigned int __maybe_unused loop_count = 0;
+	unsigned int __maybe_unused deferral_loop_count = 0;
+
+	while (1) {
```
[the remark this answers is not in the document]
etienne-lms: ;)

core/kernel/dt_driver.c
etienne-lms: as deep as the DT depth. 
You fear stack exhaustion?

lib/libutee/include/tee_api_defines_extensions.h
etienne-lms: Ok i'll rephrase more explicitly:
```
 * TEE_ERROR_DEFER_DRIVER_INIT - Device driver failed to initialize because
 *	the driver depends on a device not yet initialized.

**Comment by etienne-lms** [comment]
> For commit "drivers: se050: use release_init_resource to enable i2c trampoline"

My idea was that optee code releases resources once inits are done and I saw se050 I2C local driver as a resources needed by core only while it initializes. Once done, SE050 should be routed to the I2C RPC command, hence when init resources are released.
My reasoning is questionable :), i'll leave it to @ldts to consider if this really applies to this driver.

Regarding the impact on virtualization, is REE expected to get hand on the I2C before a VM is instantiated?
If yes, `boot_final` is the right place to re-route SE050 I2C. If not, `boot_final` is not the good place. Has SE050 support been tested with `CFG_VIRTUALIZATION=y`? IIUC se050 is initialized from [driver_init() stage](https://github.com/OP-TEE/optee_os/blob/3.15.0/core/drivers/crypto/se050/session.c#L59-L74), hence after `boot_final` when `CFG_VIRTUALIZATION=y`. I don't think the I2C REE RPC command is ready by then.

File: core/kernel/dt_driver.c (hunk 6 above)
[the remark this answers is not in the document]
Comment (jenswi-linaro) [reply core/kernel/dt_driver.c:324]: Yes, that's an improvement.

File: core/kernel/dt_driver.c (hunk 4 above)
Comment (jenswi-linaro) [reply core/kernel/dt_driver.c:662]: Not really, but it would be nice to have some guarantee against it.

File: core/kernel/dt_driver.c [hunk 10]
```diff
@@ -266,3 +296,387 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+				break;
+			default:
+				assert(0);
+			}
+		}
+
+		if (one_deferred)
+			deferral_loop_count++;
+
+		if (all_deferred && added_count == probe_list_info.added_count)
```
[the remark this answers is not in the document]
etienne-lms: addressed by https://github.com/OP-TEE/optee_os/pull/4949#discussion_r749242055?

core/kernel/dt_driver.c
etienne-lms: The DT we're parsing is expected secure. It is authenticated and taken care of with the same care as core execution env itself. Verifying DT consistency is a debug  concern, IMO. `assert()` and like would be enough?

> Not really, (...)

Do you have something else in mind?

jenswi-linaro: I guess the motivation above is good enough. Please add a comment like that.

etienne-lms:
updated

core/kernel/dt_driver.c
```diff
@@ -257,3 +408,293 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+		all_deferred = true;
+		added_node = false;
+
+		TAILQ_FOREACH_REVERSE_SAFE(elt, &dt_driver_probe_list,
+					   dt_driver_probe_head, link, prev) {
+			TAILQ_REMOVE(&dt_driver_probe_list, elt, link);
+
+			switch (probe_driver_node(fdt, elt)) {
+			case TEE_SUCCESS:
+				all_deferred = false;
```
jenswi-linaro, line 448: This negative logic is somewhat hard to follow. How about setting `one_probed_ok = true` instead?

core/kernel/dt_driver.c
```diff
@@ -257,3 +408,293 @@ TEE_Result dt_driver_probe_device_by_node(const void *fdt, int nodeoffset,
...
+	     loop_count, deferral_loop_count);
+
+	TAILQ_FOREACH(elt, &dt_driver_probe_list, link)
+		EMSG("- %s on node %s", elt->dt_drv->name,
+		     fdt_get_name(fdt, elt->nodeoffset, NULL));
+
+	panic();
+}
+
+static bool candidate_is_list_elt(struct dt_driver_probe *candidate,
```
jenswi-linaro, line 474: I can't make sense of this function name. How about `driver_probe_compare()` instead?

lib/libutee/include/tee_api_defines_extensions.h
```diff
@@ -1,12 +1,25 @@
...
+/*
+ * API extended result codes as per TEE_Result IDs defined in GPD TEE
+ * Internal Core API specification v1.1:
+ *
+ * 0x70000000 - 0x7FFFFFFF: Reserved for implementation-specific return
+ *			    code providing non-error information
+ * 0x80000000 - 0x8FFFFFFF: Reserved for implementation-specific errors
+ *
+ * TEE_ERROR_DEFER_DRIVER_INIT - Device driver failed to initialize because
+ *	the driver depends on a device not yet initialized.
```
jenswi-linaro, line 19: nit: this tab indentation is annoying, just a single space after the `*` should be better.

jenswi-linaro (commented):
Some comments.
How about dropping the commit "drivers: se050: use release_init_resource to enable i2c trampoline"? The commit doesn't fix any problem or bring any other significant improvement.

etienne-lms:
Ok, thanks. I'll apply the review tags you posted.

etienne-lms:
Fixup commits squashed. Review tags applied to relevant commits.

jenswi-linaro:
For the commits:
- core: dt_driver probe sequence
- core: dt_driver: add simple bus driver

Please apply:

etienne-lms:
Removed "[RFC] " from P-R header line.
Added a fixup commit to not panic inits when CFG_EMBED_DTB is disabled.

core/kernel/dt_driver.c
```diff
@@ -631,18 +631,19 @@ static void parse_node(const void *fdt, int node)
...
 }
 
-#ifdef CFG_EMBED_DTB
 /*
  * Parse FDT for nodes and save in probe list the node for which a dt_driver
  * matches node's compatible property.
  */
 static TEE_Result probe_dt_drivers(void)
 {
 	const void *fdt = get_embedded_dt();
```
jenswi-linaro, line 640: I'd prefer if this function was called after the if below.

core/kernel/dt_driver.c
```diff
@@ -655,6 +656,9 @@ static TEE_Result release_probe_lists(void)
 	struct dt_driver_probe *next = NULL;
 	const void * __maybe_unused fdt = get_embedded_dt();
```
jenswi-linaro, line 657: I'd prefer if this function was called after the if below.

etienne-lms:
updated

jenswi-linaro:
Fixup of the fixup looks good. :-)

etienne-lms:
@jenswi-linaro, are you ok I squash and apply your A-b tag on commit "core: dt_driver probe sequence"?

jenswi-linaro:
> @jenswi-linaro, are you ok I squash and apply your A-b tag on commit "core: dt_driver probe sequence"?

Yes, please go ahead.

etienne-lms:
Fixup commits squashed and review tags applied.

etienne-lms:
many thanks for the reviews

PR #4952, opened 2021-10-28
core: pta: socket: enable TA to query recv out buffer

jenswi-linaro:
Propagate out size for socket recv event when it's larger than the
supplied in size. Also enable passing a NULL buffer while querying the
size of the buffer.

### DIFF
--- a/core/tee/socket.c
+++ b/core/tee/socket.c
@@ -122,9 +122,9 @@ static TEE_Result socket_send(uint32_t instance_id, uint32_t param_types,
 static TEE_Result socket_recv(uint32_t instance_id, uint32_t param_types,
 			      TEE_Param params[TEE_NUM_PARAMS])
 {
-	struct mobj *mobj;
-	TEE_Result res;
-	void *va;
+	struct mobj *mobj = NULL;
+	TEE_Result res = TEE_SUCCESS;
+	void *va = NULL;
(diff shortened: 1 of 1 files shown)


core/tee/socket.c
```diff
@@ -136,11 +136,13 @@ static TEE_Result socket_recv(uint32_t instance_id, uint32_t param_types,
 		return TEE_ERROR_BAD_PARAMETERS;
 	}
 
-	va = thread_rpc_shm_cache_alloc(THREAD_SHM_CACHE_USER_SOCKET,
-					THREAD_SHM_TYPE_APPLICATION,
-					params[1].memref.size, &mobj);
-	if (!va)
-		return TEE_ERROR_OUT_OF_MEMORY;
+	if (params[1].memref.size) {
```
jforissier, line 139: Shouldn't `READ_ONCE()` be used systematically when accessing a memref?

jenswi-linaro: There's no need for that here, `params` is not in shared memory.

jforissier: OK I totally missed that. Fine then.

jforissier:

jenswi-linaro:
Tag applied

etienne-lms:

jenswi-linaro:
Tag applied

PR #4953, opened 2021-10-28
[RFC] core: pta: add remote attestation PTA

jforissier:
Add a PTA to perform remote attestation of user space TAs. Enabled with
CFG_ATTESTATION_PTA=y.

This feature allows a client application to request a measurement of a
TA running in secure user space before invoking a command. The
measurement is an SHA-256 hash of the virtual memory pages that are
specific to the TA and are non writable -- basically, code and read-only
data of the main application, as well as any shared library in use when
the measurement is made.

This is work in progress. The current code returns a simple hash; adding
a digital signature is TBD.

### DIFF
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -1,0 +1,226 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2021, Huawei Technologies Co., Ltd
+ */
+
+#include <crypto/crypto.h>
+#include <kernel/mutex.h>
+#include <kernel/pseudo_ta.h>
+#include <kernel/tee_ta_manager.h>
+#include <kernel/ts_manager.h>
+#include <kernel/user_ta.h>
+#include <mm/file.h>
+#include <mm/mobj.h>
+#include <pta_attestation.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/queue.h>
+#include <tee/entry_std.h>
+#include <tee/uuid.h>
+#include <user_ta_header.h>
+#include <utee_defines.h>
+
+#define PTA_NAME "attestation.pta"
+
+/*
+ * Is region valid for hashing?
+ * Exclude writable regions as well as those that are not specific to the TA
+ * (ldelf, kernel or temporary mappings).
+ */
+static bool is_region_valid(struct vm_region *r)
+{
+	uint32_t skip_flags = VM_FLAG_EPHEMERAL | VM_FLAG_PERMANENT |
+			      VM_FLAG_LDELF;
+
+	return !(r->flags & skip_flags || r->attr & TEE_MATTR_UW);
+}
+
+/*
+ * With this comparison function, we're hashing the smaller regions first.
+ * Regions of equal size are ordered based on their content (memcmp()).
+ * Identical regions can be in any order since they will yield the same hash
+ * anyways.
+ */
+static int cmp_regions(const void *a, const void *b)
+{
+	const struct vm_region *r1 = *(const struct vm_region **)a;
+	const struct vm_region *r2 = *(const struct vm_region **)b;
+
+	if (r1->size < r2->size)
+		return -1;
+
+	if (r1->size > r2->size)
+		return 1;
+
+	return memcmp((void *)r1->va, (void *)r2->va, r1->size);
+}
+
+static TEE_Result hash_update_mode_tags(void *ctx, struct vm_region *r)
+{
+	struct fobj *fobj = r->mobj->ops->get_fobj(r->mobj);
+	struct file *f = to_file_may_fail(r->mobj);
+	uint64_t offs_and_size[2] = { };
+	TEE_Result res = TEE_SUCCESS;
+	unsigned int poffs = 0;
+	uint8_t *tag = NULL;
+
+	/* Regions we're interested in should all have a struct file */
+	assert(f);
+
+	tag = file_get_tag(f);
+	assert(tag);
+
+	res = crypto_hash_update(ctx, tag, FILE_TAG_SIZE);
+	if (!res)
+		return res;
+
+	file_lock(f);
+	res = file_find_page_offset(f, fobj, &poffs);
+	file_unlock(f);
+	if (!res)
+		return res;
+
+	offs_and_size[0] = poffs * SMALL_PAGE_SIZE;
+	offs_and_size[1] = r->mobj->size;
+
+	return crypto_hash_update(ctx, (uint8_t *)offs_and_size,
+				  sizeof(offs_and_size));
+}
+
+static TEE_Result hash_update(void *ctx, struct vm_region *r, uint32_t mode)
+{
+	switch (mode) {
+	case PTA_ATTESTATION_HASH_MODE_FULL:
+		return crypto_hash_update(ctx, (uint8_t *)r->va, r->size);
+	case PTA_ATTESTATION_HASH_MODE_TAGS:
+		return hash_update_mode_tags(ctx, r);
+	default:
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+}
+
+static TEE_Result hash_regions(struct vm_info *vm_info, uint8_t *hash,
+			       uint32_t mode)
+{
+	TEE_Result res = TEE_SUCCESS;
+	struct vm_region *r = NULL;
+	struct vm_region **regions = NULL;
+	size_t nregions = 0;
+	void *ctx = NULL;
+	size_t i = 0;
+
+	res = crypto_hash_alloc_ctx(&ctx, TEE_ALG_SHA256);
+	if (res)
+		return res;
+
+	res = crypto_hash_init(ctx);
+	if (res)
+		goto out;
+
+	/*
+	 * Make an array of region pointers so we can use qsort() to order it.
+	 */
+
+	TAILQ_FOREACH(r, &vm_info->regions, link)
+		if (is_region_valid(r))
+			nregions++;
+
+	regions = malloc(nregions * sizeof(*regions));
+	if (!regions) {
+		res = TEE_ERROR_OUT_OF_MEMORY;
+		goto out;
+	}
+
+	TAILQ_FOREACH(r, &vm_info->regions, link)
+		if (is_region_valid(r))
+			regions[i++] = r;
+
+	/*
+	 * Sort regions so that they are in a consistent order even when TA ASLR
+	 * is enabled.
+	 */
+	qsort(regions, nregions, sizeof(*regions), cmp_regions);
+
+	/* Hash regions in order */
+	for (i = 0; i < nregions; i++) {
+		r = regions[i];
+		DMSG("va %p size %zu", (void *)r->va, r->size);
+		res = hash_update(ctx, r, mode);
+		if (res)
+			goto out;
+	}
+
+	res = crypto_hash_final(ctx, hash, TEE_SHA256_HASH_SIZE);
+out:
+	free(regions);
+	crypto_hash_free_ctx(ctx);
+	return res;
+}
+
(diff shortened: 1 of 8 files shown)


core/pta/attestation.c
```diff
@@ -0,0 +1,182 @@
...
+			      VM_FLAG_LDELF;
+
+	return !(r->flags & skip_flags || r->attr & TEE_MATTR_UW);
+}
+
+/*
+ * With this comparison function, we're hashing the smaller regions first.
+ * Regions of equal size are ordered based on their content (memcmp()).
+ * Identical regions can be in any order since they will yield the same hash
+ * anyways.
```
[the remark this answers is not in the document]
jforissier: @jenswi-linaro 

> This approach solves the problem we had before in a previous attempt at remote attestation. What if we instead did this with the "tags" instead? We wouldn't need to hash the entire TA and we'd get the initial read/write data too.

I looked and it's not clear to me how I can find the tags from the session :-/ you were referring to the struct file::tag, right?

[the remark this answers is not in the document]
jenswi-linaro: It's not trivial and you'll probably need to add a function to get a `struct mobj_with_fobj *` from a `struct mobj *` to find what `mobj_with_fobj_alloc()` stored. I'm not sure if it will end there though.

core/pta/attestation.c
```diff
@@ -0,0 +1,182 @@
...
+				     TEE_PARAM_TYPE_NONE))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (uuid_sz != sizeof(TEE_UUID))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!hash || hash_sz != TEE_SHA256_HASH_SIZE)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	nsec_sessions_list_head(&open_sessions);
```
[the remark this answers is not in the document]
etienne-lms: It looks like this command is reserved to REE kernel clients.
Session ID is a private information usually not shared with non-secure user land client. Do I get it wrong? 

I wonder also if session IDs related to TA clients should be visible/measurable by REE.

core/pta/attestation.c
```diff
@@ -0,0 +1,176 @@
...
+	ts_sess = ts_pop_current_session();
+	assert(ts_sess == &s->ts_sess);
+
+	tee_ta_put_session(s);
+	return ret;
+}
+
+static TEE_Result invoke_command(void *pSessionContext __unused,
+				 uint32_t nCommandID, uint32_t nParamTypes,
+				 TEE_Param pParams[TEE_NUM_PARAMS])
```
etienne-lms, line 162:
```patch
-				 uint32_t nCommandID, uint32_t nParamTypes,
+				 uint32_t command_id, uint32_t param_types,
-				 TEE_Param pParams[TEE_NUM_PARAMS])
```

core/pta/attestation.c
```diff
@@ -0,0 +1,176 @@
...
+
+static TEE_Result hash_ta(uint32_t types, TEE_Param params[TEE_NUM_PARAMS])
+{
+	struct tee_ta_session_head *open_sessions = NULL;
+	struct tee_ta_session *s = NULL;
+	struct ts_session *ts_sess = NULL;
+	uint32_t id = params[0].value.a;
+	uint8_t *hash = params[1].memref.buffer;
+	size_t hash_sz = params[1].memref.size;
+	TEE_Result ret = TEE_SUCCESS;
```
etienne-lms, line 121: nitpicking: prefer `res` over `ret` for consistency

core/pta/attestation.c
```diff
@@ -0,0 +1,176 @@
...
+		return TEE_ERROR_ACCESS_DENIED;
+	}
+
+	nsec_sessions_list_head(&open_sessions);
+	s = tee_ta_get_session(id, true, open_sessions);
+	if (!s)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	if (!is_user_ta_ctx(s->ts_sess.ctx))
+		return TEE_ERROR_ITEM_NOT_FOUND;
```
etienne-lms, line 146: `	tee_ta_put_session(s);`

core/pta/attestation.c
```diff
@@ -0,0 +1,176 @@
...
+		return hash_ta(nParamTypes, pParams);
+	default:
+		break;
+	}
+
+	return TEE_ERROR_NOT_IMPLEMENTED;
+}
+
+pseudo_ta_register(.uuid = PTA_ATTESTATION_UUID, .name = PTA_NAME,
+		   .flags = PTA_DEFAULT_FLAGS,
```
etienne-lms, line 175: Likely some feature that deserves enumeration.

etienne-lms (commented):
@jforissier you mentioned shared libs constraints. The PTA may need to return not-supported for TAs loaded with shared libs support. PTA should be disabled when `CFG_VIRTUALIZATION=y`.

Regarding the scope of the P-R, I think the TA measurements should be handled from the user_ta.c when TAs are loaded rather than from the PTA implementation itself.

jforissier:
> @jforissier you mentioned shared libs constraints. The PTA may need to return not-supported for TAs loaded with shared libs support. 

Why? The hash reflects the state of all the RO sections of the TA, shared library or not. Therefore I think such a hash can be relied upon as a measurement of the whole execution environment (ignoring the writable data).

> PTA should be disabled when `CFG_VIRTUALIZATION=y`.

Why?

> Regarding the scope of the P-R, I think the TA measurements should be handled from the user_ta.c when TAs are loaded rather than from the PTA implementation itself.

Why? Doing so would cause issues with `dlopen()`.

core/pta/attestation.c
jforissier: In my mind the caller is typically a user space CA, not the REE kernel so that might not be very useful.

core/pta/attestation.c
jforissier: ok

core/pta/attestation.c
jforissier: Right

core/pta/attestation.c
jforissier: OK

core/pta/attestation.c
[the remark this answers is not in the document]
jforissier: > It looks like this command is reserved to REE kernel clients.

No, it is meant for a CA wanting to authenticate the TA it is "talking" to.

> Session ID is a private information usually not shared with non-secure user land client. Do I get it wrong?

The `session_id` member of `TEEC_Session` is definitely public in `tee_client_api.h`. The exact content of `TEEC_Session` is implementation-defined as per the GP spec, so a CA using this feature would definitely not be portable to other TEEs, but we don't care much I suppose.

> I wonder also if session IDs related to TA clients should be visible/measurable by REE.

What do you mean? Measure the CA?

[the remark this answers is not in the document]
etienne-lms: A CA should not be allowed to measure other TA sessions than the one it's using. Here, a CA can invoke this PTA providing any session identifier value.

etienne-lms:
> > PTA should be disabled when CFG_VIRTUALIZATION=y.
> 
> Why?

A CA in a VM should not be able to measure (could say probe) sessions belonging to clients from other VMs.

core/pta/attestation.c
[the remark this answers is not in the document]
jforissier: > A CA should not be allowed to measure other TA sessions than the one it's using. Here, a CA can invoke this PTA providing any session identifier value.

I agree. I need to add something to prevent that.

jforissier:
> > > PTA should be disabled when CFG_VIRTUALIZATION=y.
> > 
> > 
> > Why?
> 
> A CA in a VM should not be able to measure (could say probe) sessions belonging to clients from other VMs.

OK, hopefully that is addressed by the check that makes sure the caller is the owner of the session that is being measured (still TBD).

etienne-lms:
I don't think OP-TEE core can verify that the client of PTA is the client of the session to the measured TA.
You would rather need a service from OP-TEE Linux driver that invokes this PTA, using the session ID owned and passed by userland CA.

core/pta/attestation.c
```diff
@@ -53,7 +55,28 @@ static int cmp_regions(const void *a, const void *b)
...
+	switch (mode) {
+	case PTA_ATTESTATION_HASH_MODE_FULL:
+		return crypto_hash_update(ctx, (uint8_t *)r->va, r->size);
+	case PTA_ATTESTATION_HASH_MODE_TAGS:
+		f = to_file_may_fail(r->mobj);
+		/* Regions we're interested in should all have a struct file */
+		assert(f);
+		tag = file_get_tag(f);
+		assert(tag);
+		return crypto_hash_update(ctx, tag, FILE_TAG_SIZE);
```
jenswi-linaro, line 72: This region probably only maps a part of the file, so we should try to dig out the offset into the file used and the size of this region and hash those values too.

jenswi-linaro:
> I don't think OP-TEE core can verify that the client of PTA is the client of the session to the measured TA. You would rather need a service from OP-TEE Linux driver that invokes this PTA, using the session ID owned and passed by userland CA.

Agree. Another option could be to let the TA to be attested call this PTA. A problem here is how to make sure that the TA cannot tamper with the attestation hash. Perhaps because it's MACed with a secret key which was supplied by the client to the PTA in advance.

core/pta/attestation.c
jforissier: Updated, not quite sure of my code though (locking?)

core/pta/attestation.c
```diff
@@ -55,21 +55,43 @@ static int cmp_regions(const void *a, const void *b)
...
+	uint64_t offs_and_size[2] = { 0, TO_LE64(r->mobj->size) };
+	struct fobj *fobj = r->mobj->ops->get_fobj(r->mobj);
+	struct file *f = to_file_may_fail(r->mobj);
+	TEE_Result res = TEE_SUCCESS;
+	unsigned int poffs = 0;
 	uint8_t *tag = NULL;
 
+	/* Regions we're interested in should all have a struct file */
+	assert(f);
+	file_lock(f);
```
jenswi-linaro, line 71: This can probably be moved down to only cover the call to `file_find_page_offset()`.

core/pta/attestation.c
```diff
@@ -55,21 +55,43 @@ static int cmp_regions(const void *a, const void *b)
...
+	res = crypto_hash_update(ctx, tag, FILE_TAG_SIZE);
+	if (!res) {
+		file_unlock(f);
+		return res;
+	}
+	res = file_find_page_offset(f, fobj, &poffs);
+	file_unlock(f);
+	if (!res)
+		return res;
+	offs_and_size[0] = TO_LE64(poffs * SMALL_PAGE_SIZE);
```
jenswi-linaro, line 83: This looks quite ambitious. Preparing for a big endian OP-TEE with little endian TAs?

jforissier: Yeah I thought about a big endian TEE core indeed... Probably overkill, it probably doesn't even matter if the hash changes in this case... However I'll keep the `uint64_t` conversions so that a 32-bit TA can have the same hash whatever the size of `size_t` in the TEE (Aarch32 vs. Aarch64 core).

core/pta/attestation.c
jforissier: OK

core/pta/attestation.c
jenswi-linaro: Makes sense.

jforissier:
Thanks for the comments so far. I have squashed the review commits to clean things a bit, and I will address the main remaining issue in a new RFC PR (not allowing any CA to measure any TA). I like Jens' [suggestion](https://github.com/OP-TEE/optee_os/pull/4953#issuecomment-960495569) and will explore this way.

PR #4954, opened 2021-10-29
Provide plat_scmi_clock_* using clock framework.

clementleger:
Integrating the clock framework with SCMI allows to avoid boilerplate
code to do so in platform specific files. This series adds a generic
layer that uses the generic clock framework to access and expose clocks.
A new CFG_SCMI_MSG_CLOCK_GENERIC configuration option is added to enable
this generic clock support.
Clock matching is done with device tree by using the following description:
    
```
        scmi-clock-ch0 {
            compatible = "scmi-clock-consumer";
            scmi-channel-id = <0>;
    
            scmi-clock-ch0@0 {
                reg = <0>;      // the scmi clock ID used to expose the clock
                clocks = <&some_clock>;   // reference to a platform clock
            }
            ...
        }
```

### DIFF
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -296,3 +296,12 @@ out:
 
 	return res;
 }
+
+TEE_Result clk_get_rates_array(struct clk *clk, size_t start_index,
+			       unsigned long *rates, size_t *nb_elts)
+{
+	if (!clk->ops->get_rates_array)
+		return TEE_ERROR_NOT_SUPPORTED;
+
+	return clk->ops->get_rates_array(clk, start_index, rates, nb_elts);
+}
--- a/core/include/drivers/clk.h
+++ b/core/include/drivers/clk.h
@@ -48,6 +48,7 @@ struct clk {
  * @get_parent: Get the current parent index of the clock
  * @set_rate: Set the clock rate
  * @get_rate: Get the clock rate
+ * @get_rates_array: Get the supported clock rates as array
  */
 struct clk_ops {
 	TEE_Result (*enable)(struct clk *clk);
@@ -58,6 +59,8 @@ struct clk_ops {
 			       unsigned long parent_rate);
 	unsigned long (*get_rate)(struct clk *clk,
 				  unsigned long parent_rate);
+	TEE_Result (*get_rates_array)(struct clk *clk, size_t start_index,
+				      unsigned long *rates, size_t *nb_elts);
 };
 
 /**
@@ -167,5 +170,18 @@ struct clk *clk_get_parent_by_index(struct clk *clk, size_t pidx);
  * Return a TEE_Result compliant value
  */
 TEE_Result clk_set_parent(struct clk *clk, struct clk *parent);
+
+/**
+ * clk_get_rates_array - Get supported rates as an array
+ *
+ * @clk: Clock for which the rates are requested
+ * @start_index: start index of requested rates
+ * @rates: Array of rates allocated by caller or NULL to query count of rates
+ * @nb_elts: Max number of elements that the array can hold as input. Contains
+ * the number of elements that was added in the array as output.
+ * Returns a TEE_Result compliant value
+ */
+TEE_Result clk_get_rates_array(struct clk *clk, size_t start_index,
+			       unsigned long *rates, size_t *nb_elts);
 
 #endif /* __DRIVERS_CLK_H */
--- a/core/drivers/scmi-msg/clock_generic.c
+++ b/core/drivers/scmi-msg/clock_generic.c
@@ -1,0 +1,279 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2021, Microchip
+ */
+
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <drivers/scmi.h>
+#include <drivers/scmi-msg.h>
+#include <kernel/boot.h>
+#include <libfdt.h>
+#include <string.h>
+#include <sys/queue.h>
+
+struct scmi_clk {
+	struct clk *clk;
+	unsigned int channel_id;
+	unsigned int scmi_id;
+	bool enabled;
+	SLIST_ENTRY(scmi_clk) link;
+};
+
+static SLIST_HEAD(, scmi_clk) scmi_clk_list =
+					SLIST_HEAD_INITIALIZER(scmi_clk_list);
+
+#define SCMI_MAX_CLK_NAME_LEN	16
+
+size_t plat_scmi_clock_count(unsigned int channel_id)
+{
+	unsigned int count = 0;
+	unsigned int max_id = 0;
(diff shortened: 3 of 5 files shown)


core/drivers/clk/clk.c
```diff
@@ -296,3 +296,38 @@ TEE_Result clk_set_parent(struct clk *clk, struct clk *parent)
...
+}
+
+TEE_Result clk_get_rates_array(struct clk *clk, size_t start_index,
+			       unsigned long *rates, size_t *nb_elts)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	/* Simply return the clock rate */
+	if (!clk->ops->get_rates_array)
+		return clk_dummy_get_rates(clk, start_index, rates, nb_elts);
```
etienne-lms, line 326: If driver does not provide a get_rate_array handler, i think this function should return an error. If not, this function may report different rate if clock rate changes, which is not the desired behavior of this function.

core/include/drivers/clk.h
```diff
@@ -168,4 +171,16 @@ struct clk *clk_get_parent_by_index(struct clk *clk, size_t pidx);
  */
 TEE_Result clk_set_parent(struct clk *clk, struct clk *parent);
 
+/**
+ * clk_get_rates_array - Get supported rates as an array
+ *
+ * @clk: Clock for which the rates are requested
+ * @start_index: start index of requested rates
+ * @rates: Array of rates allocated by caller or NULL to query count of rates
+ * @nb_elts: Max number of elements that the array can hold
```
etienne-lms, line 180: this descries the input value.
output value is the actual return rate array size.

etienne-lms (commented):
comment for commit "drivers: clk: add support for clk_get_rates_array"

core/drivers/clk/clk.c
```diff
@@ -296,3 +296,38 @@ TEE_Result clk_set_parent(struct clk *clk, struct clk *parent)
...
+
+	cpu_spin_lock(&clk_lock);
```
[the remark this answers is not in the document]
clementleger: Indeed, this won't protect it against rate changes. I'm not sure what can be done here. We can't "lock" globally the clock as the client might not request all the rate if he wants.

core/drivers/clk/clk.c
clementleger: Acked. But even if moved to SCMI clock handling, I'm not sure we can easily play well with rate changes.

etienne-lms: If a clock driver does not register a `get_rates_array` operator, SCMI server will not expose possible rates and the SCMI client/agent won't request rate changes. Therefore, supporting rate changes mandates clock driver to describe all possible rates for that clock, whatever its parent is. Note that if switching to a new rate requires the clock to change its parent, it will be transparent from the SCMI client/agent view.

All in one, I think supporting exposure of the clock through the SCMI clock protocol should mandate the driver to register a `get_rates_array` operator.

core/drivers/clk/clk.c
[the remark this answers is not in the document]
etienne-lms: I think the spinlock is useless and `clk_get_rates_array()` should call straight clock's `get_rates_array` handler, see comment below.

core/drivers/clk/clk.c
clementleger: Ok, I was inspired by what was done for the stm32mp1 but I guess this was kind of a quirk. I will do as you told.
ie, get_rates_array *should* return the list of rates supported whatever the parent.

core/include/drivers/clk.h
```diff
@@ -168,4 +171,17 @@ struct clk *clk_get_parent_by_index(struct clk *clk, size_t pidx);
  */
 TEE_Result clk_set_parent(struct clk *clk, struct clk *parent);
 
+/**
+ * clk_get_rates_array - Get supported rates as an array
+ *
+ * @clk: Clock for which the rates are requested
+ * @start_index: start index of requested rates
+ * @rates: Array of rates allocated by caller or NULL to query count of rates
+ * @nb_elts: Max number of elements that the array can hold as input. Contains
```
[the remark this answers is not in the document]
etienne-lms: @clementleger, if you prefer your proposal, fine with me and i'll updatethe 2 buggy descriptions in 
scmi-msg.h [clock](https://github.com/OP-TEE/optee_os/blob/3.15.0/core/include/drivers/scmi-msg.h#L183) and [voltage](https://github.com/OP-TEE/optee_os/blob/master/core/include/drivers/scmi-msg.h#L303-L304) :
```patch
- * @rates: If NULL, function returns, else output rates array
+ * @rates: Array of rates allocated by caller or NULL to query count of rates
- * @nb_elts: Array size of @levels.
+ * @nb_elts: Max number of elements that the array can hold as input. Contains
+ * the number of elements that was added in the array as output.
```

etienne-lms:
@jforissier, could you tag this as _enhancement_ or _wip_? This is a start point of discussions about SCMI service description in OP-TEE using DT.

jforissier:
@etienne-lms done

clementleger:
@etienne-lms: What can we do for this SCMI clock support to move forward ? I have the feeling the SCMI server bindings won't be defined so soon, leading to this never being integrated. Or do you have some news from the SCMI meetings about these bindings ?

clementleger:
@etienne-lms ping ?

etienne-lms:
Hi @clementleger,
Late answer and no new feedback on this :(
We (people working on SCMI development) have not reached to any consensus on how to describe the SCMI server side using DT but we strongly believe the description should start from the server basics and one that basis extend to the exposes services, that is having first a node defining agents/channels/static shared  memories/etc... while other DT nodes for service (as the exposed clocks) would refer to the channel defined from service main node. We did not progress at all since focused on other matters.
If you have ideas, please feel free update this P-R. I think starting with OP-TEE specific SCMI DT binding could help progressing on that topic but we'd need to make sure the bindings are strong enough to not be deprecated to early. FYI, next pubic meeting on SCMI development is planned May 5th, 11am CEST.

clementleger:
Hi @etienne-lms, I will rework this pull request to include the server description as proposed by Vincent Guittot. However, since there (seems to have) a lot of momentum to agree on the SCMI binding definition, I think it would be good to go toward some op-tee specific bindings. Reworking the complete bindings to support a complete device-tree description of the SCMI support seems a bit too much for what I need to do. What I'll try to do is at least to have a global description starting from the top SCMI node such as this:

```
scmi0: scmi0 {
	compatible = "optee,scmi-platform";
	#address-cells = <1>
	#size-cells = <0>

	scmi_clk: protocol@14 {
                compatible = "optee,scmi-clocks"
		reg = <0x14>;
		#clock-cells = <1>;

		scmi-clock-ch0@0 {
			reg = <0>;      // the scmi clock ID used to expose the clock
			clocks = <&some_clock>;   // reference to a platform clock
		};
		scmi-clock-ch0@1 {
			reg = <1>; 
			clocks = <&some_other_clock>;
		};
	};
};
```

PR #4962, opened 2021-11-03
dt_node_info: add the reg size to manage core virtual address map

ldebieve:
A series of changes that use the reg_size from the device tree to manage the call to phys_to_virt which now adds the len management.
Initial patch is to define the size as size_t to be compliant with the different functions using length as size_t and not ssize_t.
Update the driver using the _fdt_fill_device_info to retrieve the reg size and use it to get virtual address.

### DIFF
--- a/core/drivers/stm32_i2c.c
+++ b/core/drivers/stm32_i2c.c
@@ -290,7 +290,7 @@ struct i2c_request {
 
 static vaddr_t get_base(struct i2c_handle_s *hi2c)
 {
-	return io_pa_or_va_secure(&hi2c->base, I2C_SIZE);
+	return io_pa_or_va_secure(&hi2c->base, hi2c->reg_size);
 }
 
 static void notif_i2c_timeout(struct i2c_handle_s *hi2c)
@@ -692,11 +692,14 @@ int stm32_i2c_get_setup_from_fdt(void *fdt, int node,
 	memset(init, 0, sizeof(*init));
 
 	_fdt_fill_device_info(fdt, &info, node);
+	assert(info.reg != DT_INFO_INVALID_REG &&
+	       info.reg_size != DT_INFO_INVALID_REG_SIZE &&
+	       info.clock != DT_INFO_INVALID_CLOCK);
+
 	init->dt_status = info.status;
 	init->pbase = info.reg;
+	init->reg_size = info.reg_size;
 	init->clock = info.clock;
-	assert(info.reg != DT_INFO_INVALID_REG &&
-	       info.clock != DT_INFO_INVALID_CLOCK);
 
 	cuint = fdt_getprop(fdt, node, "i2c-scl-rising-time-ns", NULL);
 	if (cuint)
@@ -754,6 +757,7 @@ int stm32_i2c_init(struct i2c_handle_s *hi2c,
 
 	hi2c->dt_status = init_data->dt_status;
 	hi2c->base.pa = init_data->pbase;
+	hi2c->reg_size = init_data->reg_size;
 	hi2c->clock = init_data->clock;
 
 	rc = i2c_setup_timing(hi2c, init_data, &timing);
--- a/core/include/drivers/stm32_i2c.h
+++ b/core/include/drivers/stm32_i2c.h
@@ -31,6 +31,7 @@
  *
  * @dt_status: non-secure/secure status read from DT
  * @pbase: I2C interface base address
+ * @reg_size: I2C interface register map size
  * @clock: I2C bus/interface clock
  * @addr_mode_10b_not_7b: True if 10bit addressing mode, otherwise 7bit mode
  * @own_address1: 7-bit or 10-bit first device own address.
@@ -48,6 +49,7 @@
 struct stm32_i2c_init_s {
 	unsigned int dt_status;
 	paddr_t pbase;
+	size_t reg_size;
 	unsigned int clock;
 	bool addr_mode_10b_not_7b;
 	uint32_t own_address1;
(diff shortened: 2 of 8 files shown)


core/include/kernel/dt.h
```diff
@@ -199,10 +199,10 @@ static inline paddr_t _fdt_reg_base_address(const void *fdt __unused,
 	return (paddr_t)-1;
 }
 
-static inline ssize_t _fdt_reg_size(const void *fdt __unused,
-				    int offs __unused)
+static inline size_t _fdt_reg_size(const void *fdt __unused,
+				   int offs __unused)
 {
-	return -1;
+	return (size_t)-1;
```
etienne-lms, line 207: prefer `return DT_INFO_INVALID_REG_SIZE;`

etienne-lms (commented):
In commit logs, append parentheses pairs to function labels so that is easier to read-out they are functions: , `...the io_pa_or_va() function.`, `the phys_to_virt() function`.
Also s/mmu/MMU/.

Aside these comments, content looks good to me.

core/include/kernel/dt.h
ldebieve: Idea here is to be compliant with the previous function that return (paddr_t)-1 too instead of DT_INFO_INVALID_REG

jenswi-linaro:
Why is "core: kernel: use size_t instead of ssize_t for _fdt_reg_size" needed. I'd like to know which problem is solved or avoided.

ldebieve:
@jenswi-linaro This patch is a change to be coherent regarding the functions called later with the result of the _fdt_reg_size.
The functions are core_mmu_add_mapping(), io_pa_or_va(), phys_to_virt() which require size_t len rather than a ssize_t. Except the invalid size than may be return here, there is no need to use a signed type for this size.

jenswi-linaro:
> @jenswi-linaro This patch is a change to be coherent regarding the functions called later with the result of the _fdt_reg_size. The functions are core_mmu_add_mapping(), io_pa_or_va(), phys_to_virt() which require size_t len rather than a ssize_t. Except the invalid size than may be return here, there is no need to use a signed type for this size.

This sounds a bit fluffy. What do you mean by coherent? Is there a problem with the range of a ssize_t compared to a size_t in this context?

ldebieve:
In fact, there is a range issue as the size back from the device tree is defined as a u32 or u64. You cannot limit the range to ssize_t here whereas the DT doesn't limit to signed u32 ot u64. This is probably the best explanation.

jenswi-linaro:
> In fact, there is a range issue as the size back from the device tree is defined as a u32 or u64. You cannot limit the range to ssize_t here whereas the DT doesn't limit to signed u32 ot u64. This is probably the best explanation.

Fair enough, please update the commit message with something along that.

core/include/kernel/dt.h
etienne-lms: sorry, my comment does not apply to your P-R.
You should  have another patch to update all these casted `-1` to their related `DT_INFO_INVALID_*`, IMO.

ldebieve:
Updated commit comments based on review.

jenswi-linaro:
For the commits "core: kernel: use size_t instead of ssize_t for _fdt_reg_size()" and "core: kernel: add reg_size in the dt_node_info structure":

core/include/drivers/stm32_i2c.h
```diff
@@ -48,6 +49,7 @@
 struct stm32_i2c_init_s {
 	unsigned int dt_status;
 	paddr_t pbase;
+	ssize_t reg_size;
```
etienne-lms, line 52: s/ssize_t/size_t/

core/include/drivers/stm32_i2c.h
```diff
@@ -112,6 +115,7 @@ struct i2c_cfg {
  */
 struct i2c_handle_s {
 	struct io_pa_va base;
+	ssize_t reg_size;
```
etienne-lms, line 118: s/ssize_t/size_t/

etienne-lms (commented):
Comments for commit "drivers: stm32_i2c: use reg_size to map virtual addresses".

core/include/drivers/stm32_i2c.h
ldebieve: Correct, I'll fix it

core/include/drivers/stm32_i2c.h
ldebieve: ditto

ldebieve:
@etienne-lms Fix pushed

core/drivers/stm32_i2c.c
```diff
@@ -692,11 +692,14 @@ int stm32_i2c_get_setup_from_fdt(void *fdt, int node,
 	memset(init, 0, sizeof(*init));
 
 	_fdt_fill_device_info(fdt, &info, node);
+	assert(info.reg != DT_INFO_INVALID_REG &&
+	       info.reg_size != DT_INFO_INVALID_REG_SIZE &&
+	       info.clock != DT_INFO_INVALID_CLOCK);
```
jforissier, line 697: Shouldn't `panic()` be used rather than `assert()`? (the latter is a no-op when `-DNDEBUG`).

ldebieve: As previously done, we do not expect something wrong from the DT as it comes from the DT SoC level originally, that's why we use for most of our driver this check under assert only in debug mode.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for the whole series.

core/drivers/stm32_i2c.c
jforissier: OK

ldebieve:
Squashed fixup commit.
Applied review tag.

PR #4965, opened 2021-11-05
Driver probing sequence - step1 - factorize clk_dt code

etienne-lms:
This P-R implements the first commits from RFC P-R #4949 that moves code from **core/drivers/clk/clk_dt.c** to a newly created source file **core/kernel/dt_driver.c** to use the sequences and resources for other drivers than clocks.

PR #4969, opened 2021-11-08
drivers: clk: skip clock with no matching compatible

etienne-lms:
Clock DT node without a matching compatible in the registered DT drivers
should be ignored (return TEE_ERROR_ITEM_NOT_FOUND) instead of
returning a TEE_ERROR_GENERIC error code.

Fixes: dbe94a85c1a1 ("drivers: clk: add devicetree support")

Note this change conflicts with P-R https://github.com/OP-TEE/optee_os/pull/4965 that moves the related functino to dt_driver.c and doing so also addresses this change [(see proposed `dt_driver_probe_device_by_node()`)](https://github.com/OP-TEE/optee_os/pull/4965/commits/6244eab4145df6dfa4bd95a3ebc6425a4eec1f17#diff-5fca14f73705cc642bdbbd044cd99b88cfca252717f0ca813b7d6664eaba0bcaR218-R232). I think it's worth considering this change first, I'll update P-R https://github.com/OP-TEE/optee_os/pull/4965 accordingly.

@clementleger, do you agree?

### DIFF
--- a/core/drivers/clk/clk_dt.c
+++ b/core/drivers/clk/clk_dt.c
@@ -257,7 +257,7 @@ static TEE_Result clk_dt_node_clock_probe_driver(const void *fdt, int node)
 
 	count = fdt_stringlist_count(fdt, node, "compatible");
 	if (count < 0)
-		return TEE_ERROR_GENERIC;
+		return TEE_ERROR_ITEM_NOT_FOUND;
 
 	for (idx = 0; idx < count; idx++) {
 		compat = fdt_stringlist_get(fdt, node, "compatible", idx, &len);
@@ -278,7 +278,7 @@ static TEE_Result clk_dt_node_clock_probe_driver(const void *fdt, int node)
 		}
 	}
 
-	return TEE_ERROR_GENERIC;
+	return TEE_ERROR_ITEM_NOT_FOUND;
 }
 
 static TEE_Result clk_probe_clock_provider_node(const void *fdt, int node)


clementleger:
Completely ok with that, this is a bug.

`Acked-by: Clément Léger <clement.leger@bootlin.com>`

etienne-lms:
tag applied

etienne-lms:
rebased to fix CI failure + s/é/e/g in @clementleger review tag to fix checkpatch error report.

jforissier:
>  s/é/e/g in @clementleger review tag to fix checkpatch error report.

I would call it a checkpatch bug, Git commit messages are typically encoded in UTF-8 by default. Perhaps we don't use the proper checkpatch settings?

etienne-lms:
Thanks for acking this change. Since it's already part of https://github.com/OP-TEE/optee_os/pull/4965 series that has been reviewed, i'll discard this P-R (unless one prefers the change proposed in current P-R is inserted in https://github.com/OP-TEE/optee_os/pull/4965 series).

> I would call it a checkpatch bug, Git commit messages are typically encoded in UTF-8 by default. Perhaps we don't use the proper checkpatch settings?

I think we have the right checkpatch settings. Checkpatch complains on `Clément` but not on `Léger`, neither on other names as @vesajaaskelainen last name (`Jääskeläinen`). Checkpatch is not clement (if you forgive me this pun :)

etienne-lms:
Superseded by commit https://github.com/OP-TEE/optee_os/commit/ef20efc4448ce8a83ed420834b47b07e379545bb from #4965 that changed the error codes as this P-R proposed.

PR #4972, opened 2021-11-09
[RFC 2] core: pta: add remote attestation PTA

jforissier:
Add a PTA to perform remote attestation of user space TAs. Enabled with
CFG_ATTESTATION_PTA=y.

This feature allows a user space TA to request a measurement of itself
from the OP-TEE core. Typical use case: a client application invokes
its companion TA to obtain secure services and it needs some guarantees
that the code running in the TEE is indeed the expected one and has not
been tampered with or corrupted. The CA generates a random nonce and
sends it to the TA using a custom command. The TA opens a session to
the attestation PTA and passes the nonce. The PTA measures the caller,
thus generating a hash, which is then signed together with the nonce.
At this point the CA (or any other entity involved in the call flow,
such as a remote server for instance) can verify the signature. If it
is valid and the nonce is the expected one, then the TA hash can be
trusted and compared against a known good value.

TODO: implement the nonce and signature.

### DIFF
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -1,0 +1,212 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2021, Huawei Technologies Co., Ltd
+ */
+
+#include <crypto/crypto.h>
+#include <kernel/pseudo_ta.h>
+#include <kernel/tee_ta_manager.h>
+#include <kernel/ts_manager.h>
+#include <kernel/user_mode_ctx.h>
+#include <kernel/user_ta.h>
+#include <mm/file.h>
+#include <mm/mobj.h>
+#include <pta_attestation.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/queue.h>
+#include <tee/entry_std.h>
+#include <tee/uuid.h>
+#include <utee_defines.h>
+
+#define PTA_NAME "attestation.pta"
+
+/*
+ * Is region valid for hashing?
+ * Exclude writable regions as well as those that are not specific to the TA
+ * (ldelf, kernel or temporary mappings).
+ */
+static bool is_region_valid(struct vm_region *r)
+{
+	uint32_t skip_flags = VM_FLAG_EPHEMERAL | VM_FLAG_PERMANENT |
+			      VM_FLAG_LDELF;
+
+	return !(r->flags & skip_flags || r->attr & TEE_MATTR_UW);
+}
+
+/*
+ * With this comparison function, we're hashing the smaller regions first.
+ * Regions of equal size are ordered based on their content (memcmp()).
+ * Identical regions can be in any order since they will yield the same hash
+ * anyways.
+ */
+static int cmp_regions(const void *a, const void *b)
+{
+	const struct vm_region *r1 = *(const struct vm_region **)a;
+	const struct vm_region *r2 = *(const struct vm_region **)b;
+
+	if (r1->size < r2->size)
+		return -1;
+
+	if (r1->size > r2->size)
+		return 1;
+
+	return memcmp((void *)r1->va, (void *)r2->va, r1->size);
+}
+
+static TEE_Result hash_update_method_tags(void *ctx, struct vm_region *r)
+{
+	struct fobj *fobj = r->mobj->ops->get_fobj(r->mobj);
+	struct file *f = to_file_may_fail(r->mobj);
+	uint64_t offs_and_size[2] = { };
+	TEE_Result res = TEE_SUCCESS;
+	unsigned int poffs = 0;
+	uint8_t *tag = NULL;
+
+	/* Regions we're interested in should all have a struct file */
+	assert(f);
+
+	tag = file_get_tag(f);
+	assert(tag);
+
+	res = crypto_hash_update(ctx, tag, FILE_TAG_SIZE);
+	if (!res)
+		return res;
+
+	file_lock(f);
+	res = file_find_page_offset(f, fobj, &poffs);
+	file_unlock(f);
+	if (!res)
+		return res;
+
+	offs_and_size[0] = poffs * SMALL_PAGE_SIZE;
+	offs_and_size[1] = r->mobj->size;
+
+	return crypto_hash_update(ctx, (uint8_t *)offs_and_size,
+				  sizeof(offs_and_size));
+}
+
+static TEE_Result hash_update(void *ctx, struct vm_region *r, uint32_t method)
+{
+	switch (method) {
+	case PTA_ATTESTATION_HASH_METHOD_FULL:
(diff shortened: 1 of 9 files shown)


jforissier:
Followup to https://github.com/OP-TEE/optee_os/pull/4953. A slightly different approach to address the issue CAs being able to measure any running TA in my initial proposal. Here the PTA can only be called by a TA, not a CA, and the PTA measures its caller.

core/pta/attestation.c
```diff
@@ -0,0 +1,212 @@
...
+
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+				     TEE_PARAM_TYPE_MEMREF_OUTPUT,
+				     TEE_PARAM_TYPE_NONE, TEE_PARAM_TYPE_NONE))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!hash || hash_sz != TEE_SHA256_HASH_SIZE)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	s = ts_pop_current_session();
```
jenswi-linaro, line 176: This shouldn't be needed.

jforissier: Cargo-culted from https://github.com/OP-TEE/optee_os/blob/3.15.0/core/pta/system.c#L226-L228 :wink: 
Just to clarify the reason why it is needed there and not here: we're not switching to user mode here, right?

Will fix.

jenswi-linaro: When you activate a session you also activates the mappings of that context. In the case of PTAs they always borrow the mappings from the calling session so the PTA can access memory in the calling user TA.

etienne-lms:
This resembles to what fTPM does with measurements. Maybe the feature you target would nicely fit into a fTPM measurement of loaded TAs. The point is attesting TA through OP-TEE core and Linux OS assumes OP-TEE core and Linux are also attested, which is what fTPM can do, with a little help of OP-TEE's boot stage.
That said, your proposal looks fine and seems by far easier to setup than a TPM ecosystem, removing attestation of these execution contexts from the picture.

A question on your setup. Attestation key should be generate from OP-TEE (this PTA?). REE Client would get the attestation public key from the PTA?

core/pta/attestation.c
jforissier: Got it, thanks.

jforissier:
> A question on your setup. Attestation key should be generate from OP-TEE (this PTA?). REE Client would get the attestation public key from the PTA?

Yes, the key should belong to the TEE core. The public key cannot be returned by a call from the CA to the TEE because at this point we have no way of signing it to prevent tampering (chicken and egg problem). For simplicity I propose adding a key pair file specifically for this purpose, say `ATTESTATION_PTA_KEY ?= keys/attestation_pta.pem`. Similar to the TA signing key (`TA_SIGN_KEY ?= keys/default_ta.pem`) that would be a default key that can easily be overridden at build time.

What do you think?

etienne-lms:
I think it would be easier/safer that priv/pub key are generated at runtime by OP-TEE and CA gets the public part from OP-TEE. Your TA attestation already relies on OP-TEE core so a PTA service providing the pub key should be trustful from CA scope.

jforissier:
Yes the TEE core is trusted, but not the REE, so there could be a MITM attack between the CA and the TEE. Even worse if signature verification is done remotely, for instance a server wants to authenticate the TA before sending data.

etienne-lms:
ok.
I the end, this attestation could trap compromised TA on non-compromised OP-TEE. In such case, means the TA authen key was compromised. Would a second key pair be safer?

jforissier:
@etienne-lms yes it is additional security, note that a TA could loads and verifies fine (authentication key not compromised), but then an attacker could possibly alter the code at runtime (due to a vulnerability in the TEE core since code is supposed to be mapped RO or via DMA or whatever trick on memory...). In this case the runtime attestation would detect the issue.

etienne-lms:
A though: if TA participates in requesting the attestation, could it craft the pristine loaded images? ensuring back to corrupted image once PTA as generated the attestation stamp. It could be better REE kernel requests PTA attestation straight to OP-TEE core without involving the target TA, at any time.

jforissier:
Does it really make a difference? If a vulnerability allows the TA code/RO data to be altered after the attestation call, it could happen as well if a different mechanism is used to trigger the measurement. For example on entry of the next "invoke command" after the REE kernel has requested the measurement.

The problem I see with doing the measurement directly upon request from the REE is that we need to restrict which session is allowed to be measured, as you mentioned previously (in my initial proposal any CA could request the measuring of any existing session based on the numerical session ID, and that did not sound right). That means the TEE driver and/or the client library need to be modified, which I'd prefer to avoid if possible. This usage pattern [suggested](https://github.com/OP-TEE/optee_os/pull/4953#issuecomment-960495569) by Jens, having the TA invoke the PTA to measure itself, sounded like a good idea but perhaps it doesn't address some particular use case.

A different approach, and different use case too, would be to simply measure the TA binary file for a given UUID. The TEE core would do as if it was about to load the TA but instead it would just hash it (using the TA store API that should be pretty easy). With this, a remote entity could learn the state of the TAs and libraries installed on a device at any time, even if the TA is not running. In fact it may be a better solution for my initial requirement... I think I will open a 3rd RFC PR ;-) and I do have some signing code to share, too.

b49020:
Hi @jforissier,

Remote attestation of user TAs and corresponding user libraries sounds like an interesting use-case. Can we have the user or remote entity view of TEE measurements similar to IMA runtime measurements (sysfs entry: /sys/kernel/security/ima/ascii_runtime_measurements)? Maybe in a TEE specific sysfs entry: /sys/kernel/security/tee/ascii_runtime_measurements? I think this maybe a simple extension to the Linux TEE driver to request for measurements from PTA whenever the corresponding sysfs entry is read.

-Sumit

b49020:
Also, we should be able to restrict its access to Linux kernel only via TEE_IOCTL_LOGIN_REE_KERNEL method.

jforissier:
@b49020 the sysfs interface is a nice idea. Do you think the `ascii_runtime_measurements` file should list all the REE FS TAs? They would need to be enumerated by the REE, i.e. the driver or perhaps `tee-supplicant` because the TEE has no way to do so. That would exclude the early TAs (not an issue I guess since linked in the TEE image), and the secure storage TAs too.
Another approach would be to let the user space client form the UUID string and give it as part of the sysfs path. That is, request measurement of one TA binary only.

I plan to push a RFC v3 soon, we can further discuss there.

b49020:
> Do you think the ascii_runtime_measurements file should list all the REE FS TAs?

I would suggest that we only do runtime measurements of TAs that are already loaded into secure RAM as that's something we want to attest. Those could be early TAs or REE FS TAs.

PR #4973, opened 2021-11-09
Assorted fixes

parport0:
We have found several issues ranging from a function having no default return value to a double-free.

These issues were found with SVACE static analysis tool.

### DIFF
--- a/core/pta/tests/fs_htree.c
+++ b/core/pta/tests/fs_htree.c
@@ -32,8 +32,8 @@ static TEE_Result test_get_offs_size(enum tee_fs_htree_type type, size_t idx,
 {
 	const size_t node_size = sizeof(struct tee_fs_htree_node_image);
 	const size_t block_nodes = TEST_BLOCK_SIZE / (node_size * 2);
-	size_t pbn;
-	size_t bidx;
+	size_t pbn = 0;
+	size_t bidx = 0;
 
 	COMPILE_TIME_ASSERT(TEST_BLOCK_SIZE >
 			    sizeof(struct tee_fs_htree_node_image) * 2);
@@ -93,10 +93,10 @@ static TEE_Result test_read_init(void *aux, struct tee_fs_rpc_operation *op,
 				 enum tee_fs_htree_type type, size_t idx,
 				 uint8_t vers, void **data)
 {
-	TEE_Result res;
+	TEE_Result res = TEE_SUCCESS;
 	struct test_aux *a = aux;
-	size_t offs;
-	size_t sz;
+	size_t offs = 0;
+	size_t sz = 0;
 
 	res = test_get_offs_size(type, idx, vers, &offs, &sz);
 	if (res == TEE_SUCCESS) {
@@ -186,8 +186,8 @@ static uint32_t val_from_bn_n_salt(size_t bn, size_t n, uint8_t salt)
 
 static TEE_Result write_block(struct tee_fs_htree **ht, size_t bn, uint8_t salt)
 {
-	uint32_t b[TEST_BLOCK_SIZE / sizeof(uint32_t)];
-	size_t n;
+	uint32_t b[TEST_BLOCK_SIZE / sizeof(uint32_t)] = { 0 };
+	size_t n = 0;
 
 	for (n = 0; n < ARRAY_SIZE(b); n++)
 		b[n] = val_from_bn_n_salt(bn, n, salt);
@@ -197,9 +197,9 @@ static TEE_Result write_block(struct tee_fs_htree **ht, size_t bn, uint8_t salt)
 
 static TEE_Result read_block(struct tee_fs_htree **ht, size_t bn, uint8_t salt)
 {
-	TEE_Result res;
-	uint32_t b[TEST_BLOCK_SIZE / sizeof(uint32_t)];
-	size_t n;
+	TEE_Result res = TEE_SUCCESS;
+	uint32_t b[TEST_BLOCK_SIZE / sizeof(uint32_t)] = { 0 };
+	size_t n = 0;
 
 	res = tee_fs_htree_read_block(ht, bn, b);
 	if (res != TEE_SUCCESS)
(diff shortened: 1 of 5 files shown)


core/pta/tests/fs_htree.c
```diff
@@ -433,7 +433,7 @@ static struct test_aux *aux_alloc(size_t num_blocks)
 static TEE_Result test_write_read(size_t num_blocks)
 {
 	struct test_aux *aux = aux_alloc(num_blocks);
-	TEE_Result res;
+	TEE_Result res = TEE_SUCCESS;
```
etienne-lms, line 436: Note that currently this function is called with a non-zero argument, so we should not expect issue with current implementation. Nevertheless we should always have an init value for such local variables.

By the way, could you also add an init value to the 3 variables right below? Actually, the whole source file should be updated.

ta/pkcs11/src/processing_asymm.c
```diff
@@ -380,6 +380,7 @@ enum pkcs11_rc init_asymm_operation(struct pkcs11_session *session,
 {
 	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
 
+	assert(obj);
```
etienne-lms, line 383: Calling with `obj` being `NULL` is a programming error and is not expected. In such case, core would already catch the issue with a segfault. I don't think the assertion adds much value, but maybe to satisfy sanity source code tools. Otherwise, should all these input argument pointers be tested against `NULL` value?

ta/pkcs11/src/processing_symm.c
```diff
@@ -1019,8 +1019,10 @@ enum pkcs11_rc derive_key_by_symm_enc(struct pkcs11_session *session,
 		res = TEE_CipherDoFinal(proc->tee_op_handle, in_buf, in_size,
 					*out_buf, out_size);
 		rc = tee2pkcs_error(res);
-		if (rc)
+		if (rc) {
 			TEE_Free(*out_buf);
+			*out_buf = NULL;
```
etienne-lms, line 1024: `*out_buf` output value is not expected to be freed by caller since the function failed.
But I see that implementation of caller does free the buffer! Nice catch.

A better fix, IMO, would be to not load a reference in `*out_buf` unless the function succeeds.
Yet the fix proposed here does the job.

etienne-lms:
For commit "ta: pkcs11: fix a memory leak":
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`  with below addition in commit log:
```
Fixes: fa1ac7676f39 ("ta: pkcs11: Don't load objects that don't match the search during find")
```

etienne-lms (commented):
comments for the other commits

core/tee/tadb.c
```diff
@@ -172,13 +172,13 @@ static TEE_Result set_file(struct tee_tadb_dir *db, int idx)
 
 static void clear_file(struct tee_tadb_dir *db, int idx)
 {
-	if (idx < db->nbits)
+	if (db->files && idx < db->nbits)
```
jenswi-linaro, line 175: If `db->nbits` is larger than `0` then will `db->files` always be non-NULL, see `maybe_grow_files()`, so this test is meaningless. We can add a comment since it may not be obvious. Same for `test_file()` below.

jenswi-linaro (commented):
Comments for "core: fix a potential null pointer dereference"

core/pta/tests/fs_htree.c
etienne-lms: > By the way, could you also add an init value to the 3 variables right below? Actually, the whole source file should be updated.

I meant such change (updating the whole source file for local vars initialization) may not in the scope of your P-R. Tell us if so, we will create another P-R to address that.

parport0: > By the way, could you also add an init value to the 3 variables right below?

I don't think they are as critical as `res`, since `n`, `m`, `o` are just loop counters. But ok.

> updating the whole source file for local vars initialization

I'll add a commit to address that.

ta/pkcs11/src/processing_asymm.c
parport0: On closer inspection I see now that the logic in `entry_processing_init()` probably keeps `obj` not being NULL in sync with `processing_is_tee_digest()`. The switch between `init_symm_operation`/`init_asymm_operation`/`init_digest_operation` in `entry_processing_init()` was my main concern. Hence I can drop this commit.

ta/pkcs11/src/processing_symm.c
parport0: I agree that relying on `TEE_Free(NULL)` beind valid is not the prettiest solution. I'll change the fix to be more invasive, as you've suggested.

core/tee/tadb.c
parport0: I'll swap that for a comment. Thank you, these dependencies are not always obvious. Maybe some general documentation for `struct tee_tadb_dir` would be even better.

parport0:
Addressed the comments.

etienne-lms (commented):
* "pta: tests: add a just-in-case default return value to test_write_read()"
* "ta: pkcs11: fix a memory leak"
Please add `Fixes: ` tag as per https://github.com/OP-TEE/optee_os/pull/4973#issuecomment-964376960.
* "ta: pkcs11: fix a double-free"
assuming fixup patch "[Review] ta: pkcs11: fix a double-free differently" is squashed into
and you add the following tag to the commit log:
`Fixes: 8c499324e457 ("ta: pkcs11: Add function to set key data")`
* "pta: tests: fs_htree: make sure all local variables are initialized"
* ".gitignore: add tags and TAGS"

jenswi-linaro:
Please squash and fixup the commits.

parport0:
Squashed.

jenswi-linaro:
For the commits:
- core: assure the reader that clear_file(), test_file() are safe
- pta: tests: fs_htree: make sure all local variables are initialized
- pta: tests: add a just-in-case default return value to test_write_read()
- .gitignore: add tags and TAGS

Please apply:

parport0:
Done.

PR #4974, opened 2021-11-10
drivers: imx: enable CFG_TZC380 for imx6ull

andrejro:
Enable the region autoconfiguration for imx6ull by default.


jforissier:
LGTM, but could someone familiar with imx6ull confirm?

Emantor:
Yes, imx6ull uses a TZC-380 as do all i.MX processors I know of. Only special cases could be the "ulp" variants.

core/arch/arm/plat-imx/conf.mk
```diff
@@ -88,6 +88,7 @@ else ifneq (,$(filter $(PLATFORM_FLAVOR),$(mx6ull-flavorlist)))
 $(call force,CFG_MX6,y)
 $(call force,CFG_MX6ULL,y)
 $(call force,CFG_TEE_CORE_NB_CORE,1)
+$(call force,CFG_TZC380,y)
```
etienne-lms, line 91: Should be `CFG_TZC380 ?= y` as per commit log description `... by default.`.

etienne-lms (changes requested):

core/arch/arm/plat-imx/conf.mk
andrejro: The `call force` was used to align the `imx6ull` with the rest of `imx6` family.
Giving the security role TZC380 plays in the context of TEE it is preferable to force it to be enabled.

jforissier:
> Giving the security role TZC380 plays in the context of TEE it is preferable to force it to be enabled.

That's fine, in this case please mention "Always enable the region autoconfiguration for imx6ull." in the commit description. With that:

andrejro:
@jforissier done. Thanks!

PR #4978, opened 2021-11-15
RAM optimized clk framework, take 2: raw clocks

etienne-lms:
This P-R supersedes https://github.com/OP-TEE/optee_os/pull/4942 with orphan clocks renamed raw clocks.

### DIFF
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -197,7 +197,7 @@ void clk_disable(struct clk *clk)
 	cpu_spin_unlock_xrestore(&clk_lock, exceptions);
 }
 
-unsigned long clk_get_rate(struct clk *clk)
+unsigned long clk_elt_rate(struct clk *clk)
 {
 	return clk->rate;
 }
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -238,6 +238,11 @@ struct clk *clk_get_parent(struct clk *clk)
 	return clk->parent;
 }
 
+size_t clk_get_num_parents(struct clk *clk)
+{
+	return clk->num_parents;
+}
+
 static TEE_Result clk_get_parent_idx(struct clk *clk, struct clk *parent,
 				     size_t *pidx)
 {
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2021, Bootlin
  */
 
+#include <assert.h>
 #include <drivers/clk.h>
 #include <kernel/boot.h>
 #include <kernel/panic.h>
@@ -13,6 +14,17 @@
 
 /* Global clock tree lock */
 static unsigned int clk_lock = SPINLOCK_UNLOCK;
+
+void clk_init_instance(struct clk *clk, const struct clk_ops *ops)
+{
+	assert(clk && ops &&
+	       (ops->id == CLK_OPS_STANDARD ||
+		(ops->id == CLK_OPS_LIGHTWEIGHT && !ops->set_parent &&
+		 !ops->get_parent)));
+
+	clk->ops = ops;
+	refcount_set(&clk->enabled_count, 0);
+}
 
 struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		      struct clk **parent_clks, size_t parent_count)
@@ -34,8 +46,7 @@ struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		clk_std->parents[parent] = parent_clks[parent];
 
 	clk_std->name = name;
-	clk_std->clk.ops = ops;
-	refcount_set(&clk_std->clk.enabled_count, 0);
+	clk_init_instance(&clk_std->clk, ops);
 
 	return &clk_std->clk;
 }
@@ -44,6 +55,28 @@ void clk_std_free(struct clk *clk)
 {
 	if (clk)
 		free(clk_to_clk_std(clk));
+}
+
+struct clk *clk_lw_alloc(const struct clk_ops *ops, size_t count)
+{
+	struct clk *clk = NULL;
+	size_t n = 0;
+
+	assert(ops->id == CLK_OPS_LIGHTWEIGHT);
+
+	clk = calloc(count, sizeof(*clk));
+	if (!clk)
+		return NULL;
+
+	for (n = 0; n < count; n++)
+		clk_init_instance(clk + n, ops);
+
+	return clk;
+}
+
+void clk_lw_free(struct clk *clk)
+{
+	free(clk);
 }
 
 static bool __maybe_unused clk_check(struct clk *clk)
@@ -59,11 +92,21 @@ static bool __maybe_unused clk_check(struct clk *clk)
 
 		if (clk_std->num_parents > 1 && !clk->ops->get_parent)
 			return false;
-	} else {
-		return false;
+
+		return true;
 	}
 
-	return true;
+	if (is_clk_lw(clk)) {
+		if (clk->ops->set_parent || clk->ops->get_parent) {
+			DMSG("Unpexpected parent clock ops on clock %s",
+			     clk_get_name(clk));
+			return false;
+		}
+
+		return true;
+	}
+
+	return false;
 }
 
 static void cache_rate_no_lock(struct clk *clk)
@@ -126,7 +169,8 @@ TEE_Result clk_register(struct clk *clk)
 	clk_init_parent(clk);
 	cache_rate_no_lock(clk);
 
-	DMSG("Registered clock %s, freq %lu", clk_get_name(clk),
+	DMSG("Registered %sclock %s, freq %lu",
+	     is_clk_lw(clk) ? "lightweight " : "", clk_get_name(clk),
 	     clk_get_rate(clk));
 
 	return TEE_SUCCESS;
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -17,31 +17,33 @@ static unsigned int clk_lock = SPINLOCK_UNLOCK;
 struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		      struct clk **parent_clks, size_t parent_count)
 {
-	struct clk *clk = NULL;
+	struct clk_std *clk_std = NULL;
 	size_t parent = 0;
 
+	assert(ops->id == CLK_OPS_STANDARD);
 	if (name && !ops->get_name)
 		DMSG("Clock name unregistered for %s", name);
 
-	clk = calloc(1, sizeof(*clk) + parent_count * sizeof(clk));
-	if (!clk)
+	clk_std = calloc(1, sizeof(*clk_std) +
+			 parent_count * sizeof(*parent_clks));
+	if (!clk_std)
 		return NULL;
 
-	clk->num_parents = parent_count;
+	clk_std->num_parents = parent_count;
 	for (parent = 0; parent < parent_count; parent++)
-		clk->parents[parent] = parent_clks[parent];
+		clk_std->parents[parent] = parent_clks[parent];
 
-	clk->name = name;
-	clk->ops = ops;
-	refcount_set(&clk->enabled_count, 0);
+	clk_std->clk.name = name;
+	clk_std->clk.ops = ops;
+	refcount_set(&clk_std->clk.enabled_count, 0);
 
-	return clk;
+	return &clk_std->clk;
 }
 
-void clk_elt_free(struct clk *clk)
+void clk_std_free(struct clk *clk)
 {
 	if (clk)
-		free(clk_to_clk_elt(clk));
+		free(clk_to_clk_std(clk));
 }
 
 static bool __maybe_unused clk_check(struct clk *clk)
@@ -49,11 +51,17 @@ static bool __maybe_unused clk_check(struct clk *clk)
 	if (!clk->ops)
 		return false;
 
-	if (clk->ops->set_parent && !clk->ops->get_parent)
-		return false;
+	if (is_clk_std(clk)) {
+		struct clk_std *clk_std = clk_to_clk_std(clk);
 
-	if (clk->num_parents > 1 && !clk->ops->get_parent)
+		if (clk->ops->set_parent && !clk->ops->get_parent)
+			return false;
+
+		if (clk_std->num_parents > 1 && !clk->ops->get_parent)
+			return false;
+	} else {
 		return false;
+	}
 
 	return true;
 }
@@ -61,9 +69,14 @@ static bool __maybe_unused clk_check(struct clk *clk)
 static void cache_rate_no_lock(struct clk *clk)
 {
 	unsigned long parent_rate = 0;
+	struct clk_std *clk_std = NULL;
 
-	if (clk->parent)
-		parent_rate = clk_get_rate(clk->parent);
+	if (!is_clk_std(clk))
+		return;
+
+	clk_std = clk_to_clk_std(clk);
+	if (clk_std->parent)
+		parent_rate = clk_get_rate(clk_std->parent);
 
 	if (clk->ops->compute_rate)
 		clk->rate = clk->ops->compute_rate(clk, parent_rate);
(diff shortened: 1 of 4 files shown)


core/drivers/clk/clk.c
```diff
@@ -107,6 +110,11 @@ TEE_Result clk_register(struct clk *clk)
 	return TEE_SUCCESS;
 }
 
+const char *_clk_elt_name(struct clk *clk)
```
jenswi-linaro, line 192: I don't see the point with the leading `_`.

jenswi-linaro (commented):
Comment for the commit "drivers: clk: introduce get_name operator"

jenswi-linaro (commented):
Comment for "drivers: clk: introduce clk_elt"

jenswi-linaro (commented):
For commit "drivers: clk: introduce raw clocks"
> Raw clocks are lightweight clock...

How about calling them lightweight clocks instead? It could be shortened as `lw` or `lwt` where needed, not more cryptic than `elt`.

etienne-lms:
> > Raw clocks are lightweight clock...
> 
> How about calling them lightweight clocks instead? It could be shortened as lw or lwt where needed, not more cryptic than elt.

`CLK_OPS_LIGHTWEIGHT` / `is_clk_lw()` / `clk_lw_alloc()`.
Fine with me.

core/drivers/clk/clk.c
etienne-lms: ok

core/drivers/clk/clk.c
```diff
@@ -17,25 +17,27 @@ static unsigned int clk_lock = SPINLOCK_UNLOCK;
...
-	refcount_set(&clk->enabled_count, 0);
+	clk_elt->clk.name = name;
+	clk_elt->clk.ops = ops;
+	refcount_set(&clk_elt->clk.enabled_count, 0);
 
-	return clk;
+	return &clk_elt->clk;
 }
 
 void clk_free(struct clk *clk)
```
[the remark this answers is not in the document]
etienne-lms: What do you mean? Memory allocated from `clk_alloc()` are released by `clk_free()` without changes.
This comment is still valid with the raw/lw clock introduced later (allocated from `clk_raw/lw_alloc()`).
Did I miss something?.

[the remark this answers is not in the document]
jenswi-linaro: The `clk` is allocated in `clk_alloc()` as a `struct clk_elt`, but here you free it as a `struct clk` without first converting it to it's container struct. This happens to work here because `clk` is the first element in `struct clk_elt`, but that seems like an unexpected dependency.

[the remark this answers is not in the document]
etienne-lms: Actually, I expected the dependency and though it would be ok to use such shortcut.
I understand your concern. I'll update. Thanks for catching that.

etienne-lms:
updated

etienne-lms:
Issue reported by checkpatch CI test:
```
1bbed88b [review] drivers: clk: introduce clock elements
WARNING: line length of 81 exceeds 80 columns
#60: FILE: core/include/drivers/clk.h:22:
+ * CLK_OPS_ELT identities a full fledged clock. The struct clk * reference can be
```
I'll address that when squashing the fixup commits.

etienne-lms:
@clementleger, could you have a look at this series?

core/drivers/clk/clk.c
```diff
@@ -70,7 +70,10 @@ struct clk *clk_lw_alloc(const struct clk_ops *ops, size_t count)
...
 {
-	free(clk);
+	if (clk && is_clk_elt(clk))
```
jenswi-linaro, line 73: During this review I've seen a reluctance to add operators. Instead has the first approach often been like here, that is, test with `is_clk_elt()` and do special treatment. With operators you're able to take a more object oriented approach where the object (struct clk) provides operators for the needed special treatment. Is this reluctance because it's unexpected that we'll extend with other types of `struct clk_xxx` or because a hybrid approach seems easier right now?
 
Wouldn't it be easier with an operator for this?
``` C
if (clk)
    clk->ops->free(clk);
```

If you'd rather save an operator please add a comment somewhere so users of this framework understands how to use it.

etienne-lms: makes sense.
i face similar concern when considering other kind of drivers.
i'll add a `release` operator (`free` to too malloc() scope to me here).

core/drivers/clk/clk.c
```diff
@@ -14,82 +15,154 @@
...
 static void clk_init_parent(struct clk *clk)
 {
+	struct clk_elt *clk_elt = clk_to_clk_elt(clk);
 	size_t pidx = 0;
 
-	switch (clk->num_parents) {
+	if (!is_clk_elt(clk))
+		return;
+
+	clk_elt = clk_to_clk_elt(clk);
```
clementleger, line 157: `clk_elt` is already assigned ligne 148.
I guess the first cast is done a bit too early before checking the clock is indeed a clk_elt. You should remove the first one.

core/drivers/clk/clk.c
etienne-lms: Well, actually `free` is fine, API function is already `clk_free()`...

core/drivers/clk/clk.c
```diff
@@ -226,21 +322,37 @@ TEE_Result clk_set_rate(struct clk *clk, unsigned long rate)
...
+
+	return 0;
 }
 
 static TEE_Result clk_get_parent_idx(struct clk *clk, struct clk *parent,
 				     size_t *pidx)
 {
 	size_t i = 0;
 
+	if (!is_clk_elt(clk))
```
clementleger, line 347: IMHO, this check is redundant since `clk_get_num_parents()` would return 0 anyway and thus `TEE_ERROR_BAD_PARAMETERS` would be returned at line 357.
I know it would display the error message below but anyway, calling this with a lw clock would be an error.

core/drivers/clk/clk.c
etienne-lms: see fixup commit https://github.com/OP-TEE/optee_os/pull/4978/commits/3ce0d85e5733243ad8ede5a7bfa84e3d50417789.

core/drivers/clk/clk.c
```diff
@@ -285,18 +398,23 @@ TEE_Result clk_set_parent(struct clk *clk, struct clk *parent)
...
 	uint32_t exceptions = 0;
 	TEE_Result res = TEE_ERROR_GENERIC;
+	struct clk_elt *clk_elt = NULL;
 
-	if (clk_get_parent_idx(clk, parent, &pidx) || !clk->ops->set_parent)
+	if (!clk->ops->set_parent)
 		return TEE_ERROR_BAD_PARAMETERS;
 
+	if (clk_get_parent_idx(clk, parent, &pidx))
+		return TEE_ERROR_BAD_PARAMETERS;
```
clementleger, line 410: Not sure why you splat the `if` here, it still works with with lw_clk since clk_get_parent_idx would return `TEE_ERROR_BAD_PARAMETERS`. I don't think it's needed, it only save a call which is done in some error case.

core/include/drivers/clk.h
```diff
@@ -6,60 +6,116 @@
...
+#include <util.h>
 
 /* Flags for clock */
 #define CLK_SET_RATE_GATE	BIT(0) /* must be gated across rate change */
 #define CLK_SET_PARENT_GATE	BIT(1) /* must be gated across re-parent */
 
+/*
+ * Type of clock instantiated:
+ *
+ * CLK_OPS_ELT identities a full fledged clock. The struct clk * reference can be
```
clementleger, line 22: I'm really not sure "element" is the good word here, but I can't come up with something more meaningful...
child ? hierarchical ? tree_element ?

core/include/drivers/clk.h
```diff
@@ -6,60 +6,116 @@
...
+ * cast to struct clk_elt * to access clock element data.
+ *
+ * CLK_OPS_LIGHTWEIGHT identifies a simple clock, no parents, flags and
+ * features handle by strcut clk_elt instances to optimize struct clk
+ * memory footprint.
+ */
+enum clk_ops_id {
+	CLK_OPS_INVALID = 0,
+	CLK_OPS_ELT,
+	CLK_OPS_LIGHTWEIGHT,
```
clementleger, line 32: element is shortened in `ELT` but lightweight is kept entirely. Either use ELT/LW or ELEMENT/LIGHTWEIGHT

core/include/drivers/clk.h
```diff
@@ -68,22 +124,46 @@ struct clk_ops {
...
 struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		      struct clk **parent_clks, size_t parent_count);
 
+/**
+ * clk_lw_alloc - Allocate and initialize an array of lightweight clocks
+ *
+ * @ops: Clock operations
+ * @count: Number of clocks (> 0)
+ *
+ * Returns base address of a array of struct clk instances properly initialized
```
clementleger, line 158: *an array

core/include/drivers/clk.h
```diff
@@ -68,22 +124,46 @@ struct clk_ops {
...
+ * or NULL if allocation failed.
+ */
+struct clk *clk_lw_alloc(const struct clk_ops *ops, size_t count);
+
+/**
+ * clk_init_instance - Initialize a clock instance
+ *
+ * @clk: Reference to clock instance to initialize
+ * @ops: Clock operations pointner for the clock
```
clementleger, line 167: *pointer

core/include/drivers/clk.h
```diff
@@ -6,60 +6,116 @@
...
 };
 
+/* Generic helper clock operators */
+const char *clk_elt_name(struct clk *clk);
+unsigned long clk_elt_rate(struct clk *clk);
+
+/*
+ * Helper to identify clock operator type
+ */
+static inline bool is_clk_elt(struct clk *clk)
```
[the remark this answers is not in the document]
jenswi-linaro: We have already discussed this and for this kind of function we have the convention that `is_` comes first. Unless you see some other value in reordering the words.

clementleger:
> @clementleger, could you have a look at this series ?

Here are some comments. Do you have numbers regarding the additional text size vs reduced data size (out of curiosity)  ?

core/include/drivers/clk.h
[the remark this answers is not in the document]
clementleger: I just think it's more coherent to keep `clk_` in front of clk related function. These are the two only functions that denotes.
In this case why not renaming `clk_is_enabled_no_lock()` to `is_clk_enabled_no_lock()` ?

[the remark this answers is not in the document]
etienne-lms: `is_clk_elt()` is rather a clock framework internal function. It's ok it doesn't start with `clk_`.
`clk_is_enabled_no_lock()` is a _clk.c_ local function, we could even label it `is_enabled_no_lock()`.

core/drivers/clk/clk.c
etienne-lms: agree. get_parent will succeed.

core/include/drivers/clk.h
etienne-lms: `CLK_OPS_GENERIC` / `is_clk_gen()` ?
`CLK_OPS_STANDARD` / `is_clk_std()` ?
`CLK_OPS_FULLFEATURED` / `is_clk_ff()` ?

core/include/drivers/clk.h
etienne-lms: Ok. Better change to explicit ELEMENT (or GENERIC or ...) and LIGHTWEIGHT for `CLK_OPS_*`

core/drivers/clk/clk.c
etienne-lms: yes, could assert here.

core/drivers/clk/clk.c
etienne-lms: thanks, i should have removed the one at line 148.

core/include/drivers/clk.h
clementleger: `CLK_OPS_FULLFEATURED` /` is_clk_ff()` looks good to me ! Did not wanted it at first because it's too long but `ff`/`lw` is readable IMHO.

etienne-lms:
> Here are some comments. Do you have numbers regarding the additional text size vs reduced data size (out of curiosity) ?

On stm32mp1, 54 to 72 clocks, moving strcut clk from 32 bytes to 8 bytes saves 1.3 to 1.7kB of its resisdent SYSRAM (256kB). Saving a kByte on this platform is worth the gain. This allows allows new thing to be introduced in generic clock elements without impacting the lightweight clocks memory footprint.
I did not yet compare the unpaged sections sizes for the embedded images. I'll check how much code grows.

etienne-lms:
As for replacement of `CLK_OPS_ELT` as reference to standard full-fledged generic clock elements, where does your favor go:
* Vote :heart: for `CLK_OPS_GENERIC` / `is_clk_generic()`
* Vote :eyes: for `CLK_OPS_STANDARD` / `is_clk_std()`
* Vote :rocket: for `CLK_OPS_FULLFLEDGED` / `is_clk_ff()`
* Vote :+1: for `CLK_OPS_ELEMENT` / `is_clk_elt()`
* Vote :-1: if you don't have time to spend on this survey

core/include/drivers/clk.h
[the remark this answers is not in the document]
jenswi-linaro: > I just think it's more coherent to keep `clk_` in front of clk related function.

That's a matter of opinion.

> These are the two only functions that denotes. In this case why not renaming `clk_is_enabled_no_lock()` to `is_clk_enabled_no_lock()` ?

Those are different functions and not relevant for this discussion.

`is_clk_elt()` is a function to tell if the supplied `struct clk` is really a `struct clk_elt`. Notice the relation with `to_clk_elt()`. You'll find this patterns at other places too, `is_sp_ctx()`, `is_stmm_ctx()`, `is_pseudo_ta_ctx()`, `is_ta_ctx()`, `is_user_mode_ctx()`, `is_user_ta_ctx()`.

etienne-lms:
@ruchi393, @jforissier, do you want to participate to https://github.com/OP-TEE/optee_os/pull/4978#issuecomment-971467219?

jforissier:
> @ruchi393, @jforissier, do you want to participate to [#4978 (comment)](https://github.com/OP-TEE/optee_os/pull/4978#issuecomment-971467219)?

Not sure I can give a useful opinion as I don't know much about clocks and their usage... After giving a quick look at https://github.com/OP-TEE/optee_os/pull/4978/commits/bc371acc9480c508ffc9c1909277363fbc9b86c7, I *think* I would have used `struct clk_common` and `struct clk` for `struct clk` and `struct clk_elt`, respectively. And therefore,  `CLK_OPS_INVALID`/`CLK_OPS` and `is_clk()`.

etienne-lms:
> ... I would have used `struct clk_common` and `struct clk`
>  for `struct clk` and `struct clk_elt`, respectively.

Clock consumer driver would manipulate `struct clk_common` instances, for either `clk_elt` or lightweight clocks?
`TEE Result clk_enable(struct clk_common *clk);`
vs
`TEE Result clk_enable(struct clk *clk);`

It is easier consumer manipulate `struct clk` references.

etienne-lms:
No real consensus. I 'm fine with `CLK_OPS_STANDARD` / `is_clk_std()` as voted by @jenswi-linaro.
@clementleger, fine with you?
`CLK_OPS_STANDARD` / `is_clk_std()` / `clk_alloc()`
`CLK_OPS_LIGHTWEIGHT` / `is_clk_lw()` / `clk_lw_alloc()`

clementleger:
Agreed with @jforissier : `std` is fine !

jforissier:
@etienne-lms true. The important thing is for you and other clock reviewers/contributors to be comfortable with whatever names are chosen.

etienne-lms:
@clementleger, @jenswi-linaro : do I squash the existing fixup commits and apply the renaming straight into a new patch series?
Or do you prefer fixup commit on existing series for that renaming?

clementleger:
Maybe it makes more sense to be fixed before pushing it ?

jenswi-linaro:
I agree with @clementleger

etienne-lms:
appended a fixup commit for the renaming

jenswi-linaro:
> appended a fixup commit for the renaming

Whops, I thought we meant the opposite. Looks good, and no harm done I hope. :-)

etienne-lms:
I wasn't sure what you both expected :)

etienne-lms:
I can post the squashed series, it's ready...

jenswi-linaro:
OK from me.

etienne-lms:
Squashed fixup commit into a new 15 incremental patches series.

core/drivers/clk/clk.c
```diff
@@ -38,9 +38,10 @@ struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 	return clk;
 }
 
-void clk_free(struct clk *clk)
+void clk_elt_free(struct clk *clk)
 {
-	free(clk);
+	if (clk)
+		free(clk_to_clk_elt(clk));
```
etienne-lms: To fix. `clk_to_clk_elt()` not yet defined at this stage of the patch series.
I'll fix:
```c
void clk_elt_free(struct clk *clk)
{
	free(clk);
}

etienne-lms: Actually this is a bad idea.
Clock are allocated by grapes with `clk_alloc()` and cannot be freed individually.
I'll revert this patch and update `clk_alloc()` / `clk_free()` inline description accordingly:
```patch
- * clk_free - Free a clock structure
+ * clk_free - Free clock array reference returned by clk_alloc()
  * 
- * @clk: Clock to be freed or NULL
+ * @clk: A reference returned by clk_alloc() or NULL

etienne-lms:
Hi. No feedback on this. Well, after a second look at looking forward in some changes needed in the clock framework, this series just puts a mess on generic clock handling with clk to clk_elt. I'll close this and will hopefully propose another way to save precious bytes for memory constrained platforms.

jenswi-linaro:
> Hi. No feedback on this. Well, after a second look at looking forward in some changes needed in the clock framework, this series just puts a mess on generic clock handling with clk to clk_elt. I'll close this and will hopefully propose another way to save precious bytes for memory constrained platforms.

To be honest I'm quite overwhelmed by all these clock and dt driver changes back and forth all over the place. Perhaps it's possible to focus on one thing at a time.

etienne-lms:
I understand all these changes are a bit noisy. A issue with the dt driver is that it's useful once there are providers and consumers, hence changes to come (an i2c bus, a reset control, etc...). Let's see these lightweight clocks later.

Issue #4646, opened 2021-05-27
Overflow in core_mmu_init_ta_ram


ldebieve:
Hi All,

I'm facing an issue using my ST board 157C-EV1. This board (ARMv7 based) is using 1GB DDR size.
I'm using OP-TEE in DDR (no pager) and putting the core at the DDR end:
`CFG_TZDRAM_START = End of DDR - TZDRAM_SIZE`
Which means the TA_RAM will be from 0xFF****** to 0xFFFFFFFF

With this mapping, entering into the `core_mmu_init_ta_ram` it result on an issue:
	`pe = virt_to_phys((void *)(e - 1)) + 1;`

Here, the **pe return 0** because paddr_t  is defined as _unsigned long_.
pe is later used by `tee_mm_init` which is waiting for hi and low memory address.

I've for now workaround is using the `CFG_CORE_LARGE_PHYS_ADDR` but assuming that it increase the size of my final binary.

Should that be reworked to use a size rather than a end address?

BR

jenswi-linaro:
Good to hear that `CFG_CORE_LARGE_PHYS_ADDR` helped. Some increase is expected with this option, but I believe it shouldn't be too bad. So if you wish to run without `CFG_CORE_LARGE_PHYS_ADDR` might be more a matter of principle.

Anyway, I think we can agree that it is at least a bit annoying with this limitation. But there's a risk this is only one of several places that needs to be fixed. If you'd like to try fixing this I can help reviewing the patches.

ldebieve:
Hi @jenswi-linaro,

OK, Will do, just checking that enabling `CFG_CORE_LARGE_PHYS_ADDR` also shows that a build issue occurred with `LOG_LEVEL=4` and `PAGE=y`.
```
core/arch/arm/mm/tee_pager.c:1379:10: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
 1379 |          (void *)(ai->va & ~SMALL_PAGE_MASK));
```

Working on that an pushing a patch for review.

Thanks for feedback.
Lionel

PR #4979, opened 2021-11-15
core: mmu: fix overflow with high address in tee_mm_pool_t

ldebieve:
In case of TA_RAM defined at the end of address range,
the high address will be defined outside the vaddr_t limits
which ends in a 0 address usage.
The size must be used rather than the high address that will
avoid this overflow issue. Update the corresponding files due
to API modification.

Signed-off-by: Lionel Debieve <lionel.debieve@foss.st.com>

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -324,15 +324,15 @@ static void print_pager_pool_size(void)
 static void init_vcore(tee_mm_pool_t *mm_vcore)
 {
 	const vaddr_t begin = VCORE_START_VA;
-	vaddr_t end = begin + TEE_RAM_VA_SIZE;
+	size_t size = TEE_RAM_VA_SIZE;
 
 #ifdef CFG_CORE_SANITIZE_KADDRESS
 	/* Carve out asan memory, flat maped after core memory */
-	if (end > ASAN_SHADOW_PA)
-		end = ASAN_MAP_PA;
+	if (begin + size > ASAN_SHADOW_PA)
+		size = ASAN_MAP_PA - begin;
 #endif
 
-	if (!tee_mm_init(mm_vcore, begin, end, SMALL_PAGE_SHIFT,
+	if (!tee_mm_init(mm_vcore, begin, size, SMALL_PAGE_SHIFT,
 			 TEE_MM_POOL_NO_FLAGS))
 		panic("tee_mm_vcore init failed");
 }
@@ -497,7 +497,9 @@ static void init_runtime(unsigned long pageable_part)
 	 * TZSRAM.
 	 */
 	mm = tee_mm_alloc2(&tee_mm_vcore,
-		(vaddr_t)tee_mm_vcore.hi - TZSRAM_SIZE, TZSRAM_SIZE);
+			   (vaddr_t)tee_mm_vcore.lo +
+			   tee_mm_vcore.size - TZSRAM_SIZE,
+			   TZSRAM_SIZE);
 	assert(mm);
 	tee_pager_set_alias_area(mm);
 
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -2463,7 +2463,7 @@ void core_mmu_init_ta_ram(void)
 	vaddr_t s = 0;
 	vaddr_t e = 0;
 	paddr_t ps = 0;
-	paddr_t pe = 0;
+	size_t size = 0;
 
 	/*
 	 * Get virtual addr/size of RAM where TA are loaded/executedNSec
@@ -2475,14 +2475,14 @@ void core_mmu_init_ta_ram(void)
 		core_mmu_get_mem_by_type(MEM_AREA_TA_RAM, &s, &e);
 
 	ps = virt_to_phys((void *)s);
-	pe = virt_to_phys((void *)(e - 1)) + 1;
+	size = e - s;
 
 	if (!ps || (ps & CORE_MMU_USER_CODE_MASK) ||
-	    !pe || (pe & CORE_MMU_USER_CODE_MASK))
+	    !size || (size & CORE_MMU_USER_CODE_MASK))
 		panic("invalid TA RAM");
 
 	/* extra check: we could rely on core_mmu_get_mem_by_type() */
-	if (!tee_pbuf_is_sec(ps, pe - ps))
+	if (!tee_pbuf_is_sec(ps, size))
 		panic("TA RAM is not secure");
 
 	if (!tee_mm_is_empty(&tee_mm_sec_ddr))
@@ -2490,6 +2490,6 @@ void core_mmu_init_ta_ram(void)
 
 	/* remove previous config and init TA ddr memory pool */
 	tee_mm_final(&tee_mm_sec_ddr);
-	tee_mm_init(&tee_mm_sec_ddr, ps, pe, CORE_MMU_USER_CODE_SHIFT,
+	tee_mm_init(&tee_mm_sec_ddr, ps, size, CORE_MMU_USER_CODE_SHIFT,
 		    TEE_MM_POOL_NO_FLAGS);
 }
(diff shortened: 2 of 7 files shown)


jforissier:
> the high address will be defined outside the vaddr_t limits
> [...]
> The size must be used rather than the high address 

That sounds like a work around rather than a real fix... isn't `vaddr_t` supposed to be able to reach any memory address in the TEE/TA range? The contrary would seem a bit dangerous to me. Here I suppose the memory is touched in user mode only so with a different mapping and therefore different addresses?

core/arch/arm/mm/tee_mm.c
```diff
@@ -35,19 +35,19 @@ static void pfree(tee_mm_pool_t *pool, void *ptr)
...
-bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, paddr_t hi, uint8_t shift,
+bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, size_t size, uint8_t shift,
 		 uint32_t flags)
 {
 	if (pool == NULL)
 		return false;
 
 	lo = ROUNDUP(lo, 1 << shift);
-	hi = ROUNDDOWN(hi, 1 << shift);
+	size = ROUNDDOWN(size, 1 << shift);
```
jenswi-linaro, line 45: We must decrease `size` with the amount of bytes `lo` was rounded up with first.

ldebieve:
@jforissier OK, maybe my description is not really clear. 
This PR is linked to the old bug identify here: https://github.com/OP-TEE/optee_os/issues/4646.
With this mapping, entering into the core_mmu_init_ta_ram it result on an issue:
pe = virt_to_phys((void *)(e - 1)) + 1;

Here, the pe return 0 because paddr_t is defined as unsigned long.
pe is later used by tee_mm_init which is waiting for hi and low memory address.

Maybe a rephrase of the commit message is required.

jforissier:
Ah yes, if it is `paddr_t` that overflows then it should be clarified in the message, please.
But then, my question still stands: is it reasonable/sane that not all the TEE/TA memory can be addressed by `paddr_t`? Maybe the answer is "yes", I'm genuinely asking as a non-expert of MMU stuff ;-)

jenswi-linaro:
The problem is that we're trying to address one byte past the end of memory. So we can still address all valid memory, but not always 1 byte beyond.

jforissier:
@jenswi-linaro got it, thanks

core/include/mm/tee_mm.h
```diff
@@ -27,7 +27,7 @@ typedef struct _tee_mm_entry_t tee_mm_entry_t;
 struct _tee_mm_pool_t {
 	tee_mm_entry_t *entry;
 	paddr_t lo;		/* low boundary of the pool */
-	paddr_t hi;		/* high boundary of the pool */
+	size_t	size;		/* pool size */
```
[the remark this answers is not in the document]
ldebieve: Right, will do

core/arch/arm/mm/tee_mm.c
ldebieve: ok, I'll update

core/arch/arm/mm/tee_mm.c
```diff
@@ -362,7 +364,7 @@ tee_mm_entry_t *tee_mm_find(const tee_mm_pool_t *pool, paddr_t addr)
 	uint16_t offset = (addr - pool->lo) >> pool->shift;
 	uint32_t exceptions;
 
-	if (addr > pool->hi || addr < pool->lo)
+	if (addr > pool->lo + (pool->size - 1) || addr < pool->lo)
```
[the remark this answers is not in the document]
ldebieve: agree

core/include/mm/tee_mm.h
```diff
@@ -70,7 +70,7 @@ static inline bool tee_mm_validate(const tee_mm_pool_t *pool, paddr_t addr)
 uintptr_t tee_mm_get_smem(const tee_mm_entry_t *mm);
 
 /* Init managed memory area */
-bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, paddr_t hi, uint8_t shift,
+bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, size_t size, uint8_t shift,
```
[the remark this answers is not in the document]
ldebieve: consistent with the struct update, okay

ldebieve:
@jforissier Sure, I'll update the commit message regarding the `paddr_t `overflow.

core/arch/arm/kernel/boot.c
```diff
@@ -497,7 +497,9 @@ static void init_runtime(unsigned long pageable_part)
 	 * TZSRAM.
 	 */
 	mm = tee_mm_alloc2(&tee_mm_vcore,
-		(vaddr_t)tee_mm_vcore.hi - TZSRAM_SIZE, TZSRAM_SIZE);
+			   (vaddr_t)tee_mm_vcore.lo +
+			   (tee_mm_vcore.size - TZSRAM_SIZE),
```
etienne-lms, line 501: can remove parentheses

core/arch/arm/mm/core_mmu.c
```diff
@@ -2475,21 +2475,21 @@ void core_mmu_init_ta_ram(void)
 		core_mmu_get_mem_by_type(MEM_AREA_TA_RAM, &s, &e);
 
 	ps = virt_to_phys((void *)s);
-	pe = virt_to_phys((void *)(e - 1)) + 1;
+	size = (e - s);
```
etienne-lms, line 2478: parentheses

core/arch/arm/mm/mobj_dyn_shm.c
```diff
@@ -436,13 +436,16 @@ static TEE_Result mobj_mapped_shm_init(void)
...
 	core_mmu_get_mem_by_type(MEM_AREA_SHM_VASPACE, &pool_start, &pool_end);
 	if (!pool_start || !pool_end)
 		panic("Can't find region for shmem pool");
 
-	if (!tee_mm_init(&tee_mm_shm, pool_start, pool_end, SMALL_PAGE_SHIFT,
-		    TEE_MM_POOL_NO_FLAGS))
+	size = pool_end - pool_start;
+
+	if (!tee_mm_init(&tee_mm_shm, pool_start, size, SMALL_PAGE_SHIFT,
+			 TEE_MM_POOL_NO_FLAGS))
```
etienne-lms, line 448: suggestion:
```c
	if (!tee_mm_init(&tee_mm_shm, pool_start, pool_end - pool_start,
			 SMALL_PAGE_SHIFT, TEE_MM_POOL_NO_FLAGS))

File: core/arch/arm/mm/tee_mm.c [hunk 8]
```diff
@@ -329,7 +330,8 @@ size_t tee_mm_get_bytes(const tee_mm_entry_t *mm)
 
 bool tee_mm_addr_is_within_range(tee_mm_pool_t *pool, paddr_t addr)
 {
-	return (pool && ((addr >= pool->lo) && (addr <= pool->hi)));
+	return (pool && ((addr >= pool->lo) &&
+			 (addr <= pool->lo + (pool->size - 1))));
```
etienne-lms, line 334:
```c
	return pool && addr >= pool->lo && addr <= (pool->lo + pool->size - 1));

etienne-lms (commented):
in commit log:
```patch
-The size must be used rather than the high address that will
+The size must be used rather than the high address to
 avoid this overflow issue. Update the corresponding files due

ldebieve:
Fixing issue list in comments
Wait for the squash to update the commit message, replacing vaddr_t by paddr_t (comment from Jerome) and comment from Etienne too.

jenswi-linaro:

This is the start of getting rid of the pattern with an (exclusive) end address pointing one byte past the valid address range. I'd welcome patches addressing the other places. Some of the cases will probably be quite complicated so one PR at a time should be a good approach.

etienne-lms (commented):

ldebieve:
@jenswi-linaro Agree, I'll try to have further look at such exclusive end address usage and propose some fixes later.

jforissier:
@ldebieve cool, please finalize and I will merge.

ldebieve:
Rebase on last master.
Squash and fix commit message.

PR #4984, opened 2021-11-17
core: drivers: caam: disable jobring in dts only in case of external DTB

sahilnxp:
On LX2160 board, Embedded DTB is enabled.
While booting with CAAM enabled, a crash comes in OP-TEE because it
tries to disable the Job Ring in Embedded DTB, which is read only.
So disable Job ring only when using External DTB.

### DIFF
--- a/core/drivers/crypto/caam/hal/common/hal_cfg.c
+++ b/core/drivers/crypto/caam/hal/common/hal_cfg.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright 2017-2019 NXP
+ * Copyright 2017-2019, 2021 NXP
  *
  * Brief   CAAM Configuration.
  */
@@ -9,6 +9,7 @@
 #include <caam_hal_ctrl.h>
 #include <caam_hal_jr.h>
 #include <caam_jr.h>
+#include <config.h>
 #include <kernel/boot.h>
 #include <mm/core_memprot.h>
 #include <registers/jr_regs.h>
@@ -52,12 +53,13 @@ enum caam_status caam_hal_cfg_get_conf(struct caam_jrcfg *jrcfg)
 		jrcfg->offset = (CFG_JR_INDEX + 1) * JRX_BLOCK_SIZE;
 		jrcfg->it_num = CFG_JR_INT;
 
-#ifdef CFG_NXP_CAAM_RUNTIME_JR
-		if (fdt) {
-			/* Ensure Secure Job Ring is secure only into DTB */
-			caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
+		if (IS_ENABLED(CFG_NXP_CAAM_RUNTIME_JR) &&
+		    !is_embedded_dt(fdt)) {
+			if (fdt) {
+				/* Ensure Secure Job Ring is secure in DTB */
+				caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
+			}
 		}
-#endif
 	}
 
 	jrcfg->nb_jobs = NB_JOBS_QUEUE;
--- a/core/drivers/crypto/caam/hal/common/hal_cfg_dt.c
+++ b/core/drivers/crypto/caam/hal/common/hal_cfg_dt.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright 2017-2019 NXP
+ * Copyright 2017-2019, 2021 NXP
  *
  * Brief   CAAM Configuration.
  */
@@ -8,6 +8,7 @@
 #include <caam_hal_cfg.h>
 #include <caam_hal_jr.h>
 #include <caam_jr.h>
+#include <kernel/boot.h>
 #include <kernel/dt.h>
 #include <kernel/interrupt.h>
 #include <libfdt.h>
@@ -98,10 +99,12 @@ void caam_hal_cfg_get_jobring_dt(void *fdt, struct caam_jrcfg *jrcfg)
 
 	jr_offset = find_jr_offset(fdt, DT_STATUS_OK_SEC, &node);
 	if (jr_offset) {
-		/* Disable JR for Normal World */
-		if (dt_enable_secure_status(fdt, node)) {
-			EMSG("Not able to disable JR DTB entry");
-			return;
+		if (!is_embedded_dt(fdt)) {
+			/* Disable JR for Normal World */
+			if (dt_enable_secure_status(fdt, node)) {
+				EMSG("Not able to disable JR DTB entry");
+				return;
+			}
 		}
 
 		/* Get the job ring interrupt */
--- a/core/arch/arm/include/kernel/boot.h
+++ b/core/arch/arm/include/kernel/boot.h
@@ -89,4 +89,10 @@ unsigned long get_aslr_seed(void *fdt);
 
 void ffa_secondary_cpu_ep_register(vaddr_t secondary_ep);
 
+/* Returns true if passed DTB is same as Embedded DTB, otherwise false */
+static inline bool is_embedded_dt(void *fdt)
+{
+	return fdt && fdt == get_embedded_dt();
+}
+
 #endif /* __KERNEL_BOOT_H */


jenswi-linaro:
Could this problem be avoided by other means? For instance by supplying a working embedded DTB or to check some bit to see that the ring is read-only?

etienne-lms:
The issue is that the embedded DTB is, by implementation, read-only, hence one cannot modify its content `dt_enable_secure_status()` attempts to.
If the patched code is intended to work only with the external (non-secure) DTB, maybe the `caam_hal_cfg_get_conf()` should be change to use `get_external_dt()` instead of `get_dt()`.

sahilnxp:
> Could this problem be avoided by other means? For instance by supplying a working embedded DTB or to check some bit to see that the ring is read-only?

@jenswi-linaro  Currently we have just copied the DTS which is being used for Linux also, If we change here, we need to maintain multiple copies of same DTS. Regarding checking some bit to see if one ring is available or not - I don't think there is any feature like this. @clementfaure Do you know anything like that ?



> The issue is that the embedded DTB is, by implementation, read-only, hence one cannot modify its content `dt_enable_secure_status()` attempts to. If the patched code is intended to work only with the external (non-secure) DTB, maybe the `caam_hal_cfg_get_conf()` should be change to use `get_external_dt()` instead of `get_dt()`.

@etienne-lms No actually code is not intended to work only with external DTB, For LX2 it is reading the addresses of peripherals from Embedded DTB.

clementfaure:
> > Could this problem be avoided by other means? For instance by supplying a working embedded DTB or to check some bit to see that the ring is read-only?
> 
> @jenswi-linaro Currently we have just copied the DTS which is being used for Linux also, If we change here, we need to maintain multiple copies of same DTS. Regarding checking some bit to see if one ring is available or not - I don't think there is any feature like this. @clementfaure Do you know anything like that ?

The CAAM Kernel driver probes job rings depending on their status in the device tree. To choose whether a job ring must be probed or not, we must go through the device tree.

core/drivers/crypto/caam/hal/common/hal_cfg_dt.c
```diff
@@ -98,10 +99,12 @@ void caam_hal_cfg_get_jobring_dt(void *fdt, struct caam_jrcfg *jrcfg)
 
 	jr_offset = find_jr_offset(fdt, DT_STATUS_OK_SEC, &node);
 	if (jr_offset) {
-		/* Disable JR for Normal World */
-		if (dt_enable_secure_status(fdt, node)) {
-			EMSG("Not able to disable JR DTB entry");
-			return;
+		if (!IS_ENABLED(CFG_EMBED_DTB)) {
```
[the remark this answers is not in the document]
etienne-lms: ` dt_is_rw()` would be a nice helper however here the goal is the change the content of the DTB later passed to normal world, so a helper function like `dt_is_for_ree()`  seems more appropriate.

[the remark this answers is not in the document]
jenswi-linaro: Sure, or maybe even `dt_is_shared_with_nwd()`.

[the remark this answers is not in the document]
etienne-lms: Shared DT frightens me a bit. A shared DT should be placed in a safe location for secure world to get safe information from, and be placed in a non-secure memory for normal world to access it. This may requires platform specific means (e.g. using a "secure read/write" and "non-secure read-only" memory) or a dedicated SW sequence (e.g. using secure memory and moving the DTB to non-secure memory once OP-TEE inits are completed). We don't support these things in current OP-TEE implementation. Ideas and proposals are welcome.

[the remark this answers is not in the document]
jenswi-linaro: Yes, it seems more safe to work on a temporary secure copy of the DT which could be copied out again when we're done with it. I don't suppose that's in scope of this PR though.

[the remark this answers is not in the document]
sahilnxp: What about `dt_is_embedded()` function as name for function ?

```
bool dt_is_embedded()
{
       return (IS_ENABLED(CFG_EMBED_DTB));
}
```

> Yes, it seems more safe to work on a temporary secure copy of the DT which could be copied out again when we're done with it. I don't suppose that's in scope of this PR though.

Yes, scope of this PR is just to not call `dt_enable_secure_status(`) in case of Embedded DTB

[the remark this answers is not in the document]
etienne-lms: > Yes, it seems more safe to work on a temporary secure copy of the DT which could be copied out again when we're done with it.

Secure world cannot rely on a DTB loaded in a memory reacheable from the outer world (from OP-TEE view). For example, if the DTB in loaded in clear in DDR by some early boot stage, REE cannot access it since not booted yet but the DDR could be physically accessed and tampered before OP-TEE boots. In such case its too late for OP-TEE to copy the DT into some secure memory.

> I don't suppose that's in scope of this PR though.

Agree.

> ```
> bool dt_is_embedded()
> {
>        return (IS_ENABLED(CFG_EMBED_DTB));
> }
> ```

```c
bool is_embedded_dt(void *fdt)
{
	return fdt && fdt == get_embedded_dt();
}
```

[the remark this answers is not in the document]
sahilnxp: Tried testing with new suggestion on LX2 but facing a crash in board bootup.
Raised an issue for that: https://github.com/OP-TEE/optee_os/issues/5017 
Please have a look.

jenswi-linaro:
Looks good, please introduce `is_embedded_dt()` in a separate commit.

sahilnxp:
> Looks good, please introduce `is_embedded_dt()` in a separate commit.

Its done.

jenswi-linaro:

sahilnxp:

Tags updated.

core/arch/arm/kernel/boot.c
```diff
@@ -578,6 +578,11 @@ void *get_dt(void)
 	return fdt;
 }
 
+bool is_embedded_dt(void *fdt)
+{
+	return (fdt && fdt == get_embedded_dt());
```
etienne-lms, line 583: Can remove parentheses.
I would rather define this function as an static inline function in _boot.h_.

core/drivers/crypto/caam/hal/common/hal_cfg.c
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright 2017-2019 NXP
+ * Copyright 2017-2019, 2021 NXP
```
etienne-lms, line 3: s/2017-2019, 2021/2017-2021/

core/drivers/crypto/caam/hal/common/hal_cfg.c
```diff
@@ -52,12 +53,13 @@ enum caam_status caam_hal_cfg_get_conf(struct caam_jrcfg *jrcfg)
...
 
-#ifdef CFG_NXP_CAAM_RUNTIME_JR
-		if (fdt) {
+		if (IS_ENABLED(CFG_NXP_CAAM_RUNTIME_JR) &&
+		    !is_embedded_dt(fdt)) {
+			if (fdt) {
 			/* Ensure Secure Job Ring is secure only into DTB */
-			caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
+				caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
+			}
```
etienne-lms, line 61: Fix indentation of the inline comment. Suggestion:
```c
		if (IS_ENABLED(CFG_NXP_CAAM_RUNTIME_JR) &&
		    fdt && !is_embedded_dt(fdt)) {
			/* Ensure Secure Job Ring is secure only into DTB */
			caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
		}

etienne-lms (commented):
For commit "core: drivers: caam: disable jobring in dts only in case of external DTB"

Rephrase commit header line: s/dts/DT/:
"core: drivers: caam: disable jobring in DT only in case of external DTB"

core/drivers/crypto/caam/hal/common/hal_cfg.c
sahilnxp: Here in NXP we follows this approach, but if OP-TEE follows this approach you mentioned, I will change that.

etienne-lms: As you wish.

sahilnxp: Kept it NXP style.

core/drivers/crypto/caam/hal/common/hal_cfg.c
sahilnxp: Done

core/arch/arm/kernel/boot.c
sahilnxp: Done.

etienne-lms (commented):

sahilnxp:

Tag updated.

PR #4987, opened 2021-11-18
Introduce boot warning print for vanilla developer builds

jbech-linaro:
Recent security issues reported to the OP-TEE project covering iMX-devices has mostly been about current CSU configuration in the OP-TEE project. This purpose of this PR is to reduce the likelihood of configuring plat-imx incorrectly. It's simply a debug print. It might seem overkill to have something like this, but I think we've seen enough reports now that we can justify adding this.

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1253,6 +1253,10 @@ void __weak boot_init_primary_late(unsigned long fdt)
 	configure_console_from_dt();
 
 	IMSG("OP-TEE version: %s", core_v_str);
+	if (IS_ENABLED(CFG_WARN_INSECURE)) {
+		IMSG("WARNING: This OP-TEE configuration might be insecure!");
+		IMSG("WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html");
+	}
 	IMSG("Primary CPU initializing");
 #ifdef CFG_CORE_ASLR
 	DMSG("Executing at offset %#lx with virtual load address %#"PRIxVA,
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -723,5 +723,15 @@ CFG_DRIVERS_CLK ?= n
 CFG_DRIVERS_CLK_DT ?= $(call cfg-all-enabled,CFG_DRIVERS_CLK CFG_DT)
 CFG_DRIVERS_CLK_FIXED ?= $(CFG_DRIVERS_CLK_DT)
 
+# The purpose of this flag is to show a print when booting up the device that
+# indicates whether the board runs a standard developer configuration or not.
+# A developer configuration doesn't necessarily has to be secure. The intention
+# is that the one making products based on OP-TEE should override this flag in
+# plat-xxx/conf.mk for the platform they're basing their products on after
+# they've finalized implementing stubbed functionality (see OP-TEE
+# documentation/Porting guidelines) as well as vendor specific security
+# configuration.
+CFG_WARN_INSECURE ?= y
+
 $(eval $(call cfg-depends-all,CFG_DRIVERS_CLK_DT,CFG_DRIVERS_CLK CFG_DT))
 $(eval $(call cfg-depends-all,CFG_DRIVERS_CLK_FIXED,CFG_DRIVERS_CLK_DT))


core/arch/arm/plat-imx/drivers/imx_csu.c
```diff
@@ -103,6 +103,9 @@ static TEE_Result csu_init(void)
 	const struct csu_config *csu_config = NULL;
 	const struct csu_setting *csu_setting = NULL;
 
+	if (IS_ENABLED(CFG_IMX_INSECURE_CSU_SETTING))
+		IMSG("Device running with an insecure setting, consult the security reference manual from NXP");
```
jbech-linaro, line 107: I believe we allow long debug print lines since a while back?

core/arch/arm/plat-imx/conf.mk
```diff
@@ -460,5 +460,30 @@ CFG_IMX_CAAM ?= y
...
+#     under a NDA. I.e., partners will need to sign a NDA with NXP to get hold
+#     of the security reference manual. Because of this the OP-TEE project
+#     cannot and are not allowed to share the security reference manual.
+#  2. Once the CSU configuration has been completed and verified that it is
+#     working, this flag should be flipped to "n".
+#
+# For more information about some of the implications when not setting this
+# correctly can be found at:
+#   - https://github.com/OP-TEE/optee_os/security/advisories/GHSA-6q85-3ph3-rm47
+#   - https://github.com/OP-TEE/optee_os/security/advisories/GHSA-4pqr-q8rf-8464
```
jbech-linaro, line 485: This URL is currently only working for OP-TEE admins. But once we've moved it to "public" it will work for anyone.

core/arch/arm/plat-imx/drivers/imx_csu.c
jforissier, line 107: Yes I think there was a consensus. That being said, you could as well print two lines:
```suggestion
	 if (IS_ENABLED(CFG_IMX_INSECURE_CSU_SETTING)) {
		IMSG("Device running with an insecure setting!");
		IMSG("Consult the security reference manual from NXP");
	}
```

Would "*possibly* running" be more accurate?

core/arch/arm/plat-imx/conf.mk
```diff
@@ -460,5 +460,30 @@ CFG_IMX_CAAM ?= y
...
+# configuration at all. I.e., this flag doesn't affect any CSU settings at all.
+# It only enables or disables an informative print serving as a reminder to the
+# device manufacturer that they need to make changes to the CSU.
+#
+# The device manufacturer should:
+#  1. Consult the NXP security reference manual and configure the CSU
+#     accordingly. NXP shares the security reference manual with their partners
+#     under a NDA. I.e., partners will need to sign a NDA with NXP to get hold
+#     of the security reference manual. Because of this the OP-TEE project
+#     cannot and are not allowed to share the security reference manual.
```
etienne-lms, line 478: s/are/is/

jforissier (approved):
Like you (probably) I'd like we could do better, but I have nothing to suggest. This all looks reasonable and acceptable to me, so:

Pushing a bit further... shouldn't a similar message be printed for *all* devices? As you said, issues were reported related to the IMX CSU, but different platforms have equivalent mechanisms. Or you want to be able to point the downstream user to specific things?

core/arch/arm/plat-imx/drivers/imx_csu.c
etienne-lms: I don't really see how this helps. The embedded core image may contain mis-configured  security elements with or w/o the switch enabled.
From a platform user perspective, I think I would expect platform refuses to boot or provide some services when `CFG_INSECURE_SETTINGS=n` and something is badly configured.

core/arch/arm/plat-imx/conf.mk
```diff
@@ -460,5 +460,30 @@ CFG_IMX_CAAM ?= y
 endif
 endif
 
+# This is only here to force an informative print showing that the device runs
```
clementfaure, line 463: I would wait for NXP to publish the disclaimer before adding this chunk of text. Some information added in this file will also be specified in the disclaimer NXP is currently working on.

I would also personally prefer to have explanatory text in the documentation rather than in the source code.
Let me know if waiting for the disclaimer before merging is okay for you.

core/arch/arm/plat-imx/drivers/imx_csu.c
clementfaure: May I suggest something like:
```IMSG("Device running with an open security settings, please refer to OP-TEE i.MX platform documentation");``` ?
OPTEE i.MX documentation being the readthedocs page where NXP will add the disclaimer.

**Comment by jenswi-linaro** [comment]
> Pushing a bit further... shouldn't a similar message be printed for all devices?

Agree, that would include to use `keys/default_ta.pem` for instance.

**Review by clementfaure — COMMENTED** [review]
Hi @jbech-linaro, thanks for your help.
Please see my comments

File: core/arch/arm/plat-imx/conf.mk [hunk 5]
```diff
@@ -460,5 +460,30 @@ CFG_IMX_CAAM ?= y
...
+#     of the security reference manual. Because of this the OP-TEE project
+#     cannot and are not allowed to share the security reference manual.
+#  2. Once the CSU configuration has been completed and verified that it is
+#     working, this flag should be flipped to "n".
+#
+# For more information about some of the implications when not setting this
+# correctly can be found at:
+#   - https://github.com/OP-TEE/optee_os/security/advisories/GHSA-6q85-3ph3-rm47
+#   - https://github.com/OP-TEE/optee_os/security/advisories/GHSA-4pqr-q8rf-8464
+CFG_IMX_INSECURE_CSU_SETTING ?= y
```
Emantor, line 486: Please disable this for i.MX6UL, upstream OP-TEE configures the CSU correctly for this variant.

jbech-linaro:
@jenswi-linaro @jforissier @etienne-lms @clementfaure @Emantor , there was a lot of feedback in this PR. Let me try to summarize here:

1. Consensus seems to be that it's better to have a general warning enabled for our default configurations instead of having specific warnings spread out across the code base / platforms.
2. optee_docs is a better place to keep the information and we should mention that instead of vendor specific documentation.
3. A flag only enabling/disabling a print gives a false impression that a device is running in secure/insecure mode.

**General warning**
- Where to place the print? `__tee_entry_std()`? That would spam a lot. `call_initcalls()` might be better?
- How should we deal with devices that are behaving good (see @Emantor's [comment](https://github.com/OP-TEE/optee_os/pull/4987#discussion_r752368738) above). I'm inclined to say that we always should have it enabled and leave it up to the end user to change the flag when building in their environment.
- What should the print say? Probably easiest to just give a URL to the documentation. Something like
 -- `"OP-TEE running in insecure mode, see https://..."` or
 -- `"OP-TEE running in unconfigured mode, see https://..."`
- Are we saying that we should close this PR and create a new one only meant for a global warning?
- A global flag would be perfect, my only concern with a global flag is that it'll be too generic.

**optee_docs and not the code**
- I've been suggesting both, https://github.com/OP-TEE/optee_docs/pull/147 was a starting point for the docs, but just as an extra step I created this PR with the print. My intention was that if someone see that print, they'll look at the flag that enabled it and there we also have information about it.

**Print gives false impression**
- Agree, hence why I added a long comment about it around the flag itself.
- Initially I was thinking about checking against current CSU settings, but doing so is not trivial.
- A global warning would definitely suffer from this as well.

etienne-lms:
Thanks @jbech-linaro for the wrap-up :+1:
Better a trace than nothing can still be a argument (despite what i've said). Having an `IMSG()` warning trace that is enabled by default looks legitimate to me. 
E.g. `CFG_UNSAFE_WARNING_BANNER ?= y`
 `IMSG("This is a development OP-TEE: may expose sensitive data")`
Disabling the config explicitly only disables a banner.

jforissier:
@jbech-linaro here is my proposal.
- Introduce `CFG_WARN_INSECURE ?= y` in `mk/config.mk`, which would print:
```
WARNING: This OP-TEE configuration might be insecure!
WARNING: Please check https://...
```
- The message would conveniently be printed in `core/arch/arm/kernel/boot.c` immediately after the line:
```
OP-TEE version: xxx
```
- Platforms which are known to be secure by default would simply set `CFG_WARN_INSECURE ?= n` in their `plat-xxx/conf.mk` to silence the warning.

clementfaure:
HI @jbech-linaro ,
Thanks for the recap,
I would personally go with this proposal https://github.com/OP-TEE/optee_os/pull/4987#issuecomment-973902230
It's a generic warning and flexible enough for the maintainers to define if their platforms/platform flavors should be considered as secure or insecure.
A warning message + a link to the documentation is okay to me.

jbech-linaro:
> @jbech-linaro here is my proposal.

Looking good Jerome. I'll rework and reuse this PR. Stay tuned.

core/arch/arm/plat-imx/conf.mk
abarisani: This is not correct, the i.MX6UL configuration is also incorrect per upcoming advisory.

Emantor: > This is not correct, the i.MX6UL configuration is also incorrect per upcoming advisory.

Please either post the info directly or point me to the advisory as soon as it becomes available.

jbech-linaro:
@OP-TEE/maintainers , I've pushed another patch hopefully capturing the last set of comments. Please have a look.

b49020:
Looks reasonable to me:

Emantor:
`Acked-by: Rouven Czerwinski <r.czerwinski@pengutronix.de>`

core/arch/arm/plat-imx/conf.mk
abarisani: https://github.com/f-secure-foundry/advisories/blob/master/Security_Advisory-Ref_FSC-HWSEC-VR2021-0002-OP-TEE_TrustZone_bypass_at_wakeup.txt

ruchi393:
`Reviewed-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

clementfaure:
```Acked-by: Clement Faure <clement.faure@nxp.com>```

jbech-linaro:
Thanks, I've added the R-B/A-B tags found so far, so the PR should be ready to merge. But of others would like to have the tag as well, please let me know and I'll add them as well.

jforissier:

jbech-linaro:
> Acked-by: Jerome Forissier <jerome@forissier.org>

Added your tag as well, thanks!

jenswi-linaro:

jbech-linaro:
> Acked-by: Jens Wiklander <jens.wiklander@linaro.org>

Tag added!

jforissier:
@jbech-linaro thanks!

PR #4988, opened 2021-11-19
[FIX] core: dt_driver: pass driver type to clk_get_provider_by_*()

etienne-lms:
Adds driver type argument in clk_get_provider_by_*() functions to find
the target provider since a DT node may register providers of different
types. Updates dt_driver_device_from_node_idx_prop() accordingly.

Fixes: f498c4042931 ("core: dt_driver: factorize clk_get_provider_by_*()")

### DIFF
--- a/core/drivers/clk/clk_dt.c
+++ b/core/drivers/clk/clk_dt.c
@@ -33,6 +33,7 @@ static struct clk *clk_dt_get_by_idx_prop(const char *prop_name,
 {
 	void *device = dt_driver_device_from_node_idx_prop(prop_name, fdt,
 							   nodeoffset, clk_idx,
+							   DT_DRIVER_CLK,
 							   res);
 
 	return (struct clk *)device;
@@ -100,7 +101,7 @@ static TEE_Result clk_probe_clock_provider_node(const void *fdt, int node)
 		return TEE_ERROR_ITEM_NOT_FOUND;
 
 	/* Check if node has already been probed */
-	if (dt_driver_get_provider_by_node(node))
+	if (dt_driver_get_provider_by_node(node, DT_DRIVER_CLK))
 		return TEE_SUCCESS;
 
 	/* Check if the node has a clock property first to probe parent */
--- a/core/include/kernel/dt_driver.h
+++ b/core/include/kernel/dt_driver.h
@@ -68,6 +68,7 @@ TEE_Result dt_driver_register_provider(const void *fdt, int nodeoffset,
  * @fdt: FDT base address
  * @nodeoffset: node offset in the FDT
  * @prop_idx: index of the phandle data in the property
+ * @type: Driver type
  * @res: Output result code of the operation: TEE_ERROR_BUSY is target
  *	device is not yet initialized, otherwise any other compliant code.
  *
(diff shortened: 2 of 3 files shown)


etienne-lms:
I updated the P-R, removing the embedded test part that will come later.
Remains a single commit that propose a generic API for support reset controller in OP-TEE.

jforissier:
> Adds driver type argument in clk_get_provider_by_*() functions to find

"Adds driver type argument to dt_driver_get_provider_by_node()
and dt_driver_get_provider_by_phandle() ..."

(citing function names in full when reasonable helps with git log --grep=xyz)

> the target provider since a DT node may register providers of different
> types. 

Can you illustrate with a DT fragment? Not necessarily in the commit description but here in the PR so that I understand the fix better.

etienne-lms:
Below are 2 examples of DT nodes relating to devices that are both a clock provider and a reset controller provider. Extracted from Linux kernel v5.15 DTS files.

[`ao_ctrl@f7800000` from arch/arm64/boot/dts/hisilicon/hi6220.dtsi](https://github.com/torvalds/linux/blob/8bb7eca972ad531c9b149c0a51ab43a417385813/arch/arm64/boot/dts/hisilicon/hi6220.dtsi#L259-L264):
```dts
		ao_ctrl: ao_ctrl@f7800000 {
			compatible = "hisilicon,hi6220-aoctrl", "syscon";
			reg = <0x0 0xf7800000 0x0 0x2000>;
			#clock-cells = <1>;
			#reset-cells = <1>;
		};
```
[`rcc@50000000` from arch/arm64/boot/dts/hisilicon/hi6220.dtsi](https://github.com/torvalds/linux/blob/8bb7eca972ad531c9b149c0a51ab43a417385813/arch/arm/boot/dts/stm32mp151.dtsi#L1119-L1124)
```dts
		rcc: rcc@50000000 {
			compatible = "st,stm32mp1-rcc", "syscon";
			reg = <0x50000000 0x1000>;
			#clock-cells = <1>;
			#reset-cells = <1>;
		};
```

If such nodes were to probe clock and reset controller drivers in OP-TEE core, each of these drivers would register the DT node in the driver provider list: an entry for the clock provider driver and an entry for the reset controller provider driver.

When a clock consumer driver requests its clocks, we ends up calling `dt_driver_get_provider_by_phandle()` to find the driver in the driver provider list. Since both clock driver and reset controller driver relates to the same node (hence same node phandle), the function needs to use the phandle/node reference and to differentiate clock provider and reset provider drivers entries in the provider list.

Below is an example of a clock and reset device node consuming both a clock and a reset controller on stm32mp151.dtsi:
```dts
		hash1: hash@54002000 {
			compatible = "st,stm32f756-hash";
			reg = <0x54002000 0x400>;
			interrupts = <GIC_SPI 80 IRQ_TYPE_LEVEL_HIGH>;
	---------->	clocks = <&rcc HASH1>;
	---------->	resets = <&rcc HASH1_R>;
			dmas = <&mdma1 31 0x2 0x1000A02 0x0 0x0>;
			dma-names = "in";
			dma-maxburst = <2>;
			status = "disabled";
		};
```

jforissier:
@etienne-lms got it, thanks for the detailed explanation.

etienne-lms:
Thanks for the feedback. I've rephrase the commit log and apply review tag.
If commit log is not clean, please tell me. I'll remove the review tag and propose another phrasing.

jforissier:
> Thanks for the feedback. I've rephrase the commit log and apply review tag. If commit log is not clean, please tell me. I'll remove the review tag and propose another phrasing.

That's OK, the issue was more with me not knowing enough about this kind of DT nodes. If you want you may append a sentence to the first paragraph: "For example, a DT node may describe a device that acts both as a as a clock provider and a reset controller, for which two drivers are needed." or something like that...

etienne-lms:
Thanks. I've added the suggested description in the commit log.

jforissier:
Merged manually with a spelling mistake fixed in the commit description ("as a as a", my fault). Thanks!

etienne-lms:
Thanks @jforissier.

PR #4989, opened 2021-11-21
Fix SSK generation for CFG_CORE_HUK_SUBKEY_COMPAT

rpiasetskyi:
Legacy code has used the result of tee_otp_get_die_id function for SSK
generation. get_dummy_die_id function always returns default value and
this doesn't work for devices that initialized RPMB partition with old
OP-TEE version and have SoC specific tee_otp_get_die_id implementation.

Signed-off-by: Ruslan Piasetskyi <ruslan.piasetskyi@gmail.com>

### DIFF
--- a/core/kernel/huk_subkey.c
+++ b/core/kernel/huk_subkey.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2019, Linaro Limited
  */
 
+#include <config.h>
 #include <crypto/crypto.h>
 #include <kernel/huk_subkey.h>
 #include <kernel/tee_common_otp.h>
@@ -15,17 +16,18 @@ static TEE_Result mac_usage(void *ctx, uint32_t usage)
 }
 
 #ifdef CFG_CORE_HUK_SUBKEY_COMPAT
-/*
- * This gives the result of the default tee_otp_get_die_id()
- * implementation.
- */
-static void get_dummy_die_id(uint8_t *buffer, size_t len)
+static TEE_Result get_otp_die_id(uint8_t *buffer, size_t len)
 {
 	static const char pattern[4] = { 'B', 'E', 'E', 'F' };
 	size_t i;
 
+	if (IS_ENABLED(CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID))
+		return tee_otp_get_die_id(buffer, len);
+
 	for (i = 0; i < len; i++)
 		buffer[i] = pattern[i % 4];
+
+	return TEE_SUCCESS;
 }
 
 /*
@@ -42,7 +44,9 @@ static TEE_Result huk_compat(void *ctx, enum huk_subkey_usage usage)
 	case HUK_SUBKEY_RPMB:
 		return TEE_SUCCESS;
 	case HUK_SUBKEY_SSK:
-		get_dummy_die_id(chip_id, sizeof(chip_id));
+		res = get_otp_die_id(chip_id, sizeof(chip_id));
+		if (res)
+			return res;
 		res = crypto_mac_update(ctx, chip_id, sizeof(chip_id));
 		if (res)
 			return res;
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -648,6 +648,10 @@ endif
 # Enables backwards compatible derivation of RPMB and SSK keys
 CFG_CORE_HUK_SUBKEY_COMPAT ?= y
 
+# Use SoC specific tee_otp_get_die_id() implementation for SSK key generation.
+# This option depends on CFG_CORE_HUK_SUBKEY_COMPAT=y.
+CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID ?= n
+
 # Compress and encode conf.mk into the TEE core, and show the encoded string on
 # boot (with severity TRACE_INFO).
 CFG_SHOW_CONF_ON_BOOT ?= n


jenswi-linaro:
It looks like this will break with implementations depending on the current behaviour of the weak tee_otp_get_die_id().

rpiasetskyi:
`fda78375e37ff7341cd205bdc26c9ed9b991b927` changes default behavior for `tee_otp_get_die_id` function and breaks SSK generation for the devices that did not implement SoC specific `tee_otp_get_die_id`.

`df91a52223295c917ef5a8eb6bfaf0ad0544b6b8` changes SSK generation and breaks it for devices that does implement SoC specific `tee_otp_get_die_id`.

`178887362ae97acfd0f30653f15ac7e3f18b1e36` fixes SSK generation for devices without SoC specific `tee_otp_get_die_id`.

But there are still devices that use "old" protocol and implement SoC specific `tee_otp_get_die_id`. And the big problem is that these devices have already initialized RPMB partitions with some secrets inside. 

@jenswi-linaro, would it be a better solution to change `get_dummy_die_id` function to `get_compat_die_id` weak function? Then devices with SoC specific `tee_otp_get_die_id` can implement `get_compat_die_id` and old protocol will work.

jenswi-linaro:
> would it be a better solution to change `get_dummy_die_id` function to `get_compat_die_id` weak function? Then devices with SoC specific `tee_otp_get_die_id` can implement `get_compat_die_id` and old protocol will work.

It sounds like that should work, please name the function `tee_otp_get_compat_die_id()` with a short comment on when it's supposed to be used.

rpiasetskyi:
After some thinking, I've decided to introduce the new config instead of using a weak function to avoid additional unnecessary platform-specific code: in all cases, it will be just `return tee_otp_get_die_id(buffer, len);`

mk/config.mk
```diff
@@ -648,6 +648,9 @@ endif
 # Enables backwards compatible derivation of RPMB and SSK keys
 CFG_CORE_HUK_SUBKEY_COMPAT ?= y
 
+# Use SoC specific tee_otp_get_die_id implementation for SSK key generation
+#CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID ?= y
```
jenswi-linaro, line 652: Please add a comment that this depends on `CFG_CORE_HUK_SUBKEY_COMPAT` to be enabled like you describe it in the commit message.
Leave this as
`CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID ?= n`
without a `#` in front.

core/kernel/huk_subkey.c
```diff
@@ -15,18 +15,31 @@ static TEE_Result mac_usage(void *ctx, uint32_t usage)
...
+#else
 /*
  * This gives the result of the default tee_otp_get_die_id()
  * implementation.
  */
-static void get_dummy_die_id(uint8_t *buffer, size_t len)
+static TEE_Result get_otp_die_id(uint8_t *buffer, size_t len)
 {
-	static const char pattern[4] = { 'B', 'E', 'E', 'F' };
+	static const char pattern[] = { 'B', 'E', 'E', 'F' };
```
jenswi-linaro, line 34: This change is not relevant to this patch.

core/kernel/huk_subkey.c
```diff
@@ -15,18 +15,31 @@ static TEE_Result mac_usage(void *ctx, uint32_t usage)
...
 	size_t i;
 
 	for (i = 0; i < len; i++)
-		buffer[i] = pattern[i % 4];
+		buffer[i] = pattern[i % sizeof(pattern)];
```
jenswi-linaro, line 38: This change is not relevant to this patch.

core/kernel/huk_subkey.c
```diff
@@ -15,18 +15,31 @@ static TEE_Result mac_usage(void *ctx, uint32_t usage)
 }
 
 #ifdef CFG_CORE_HUK_SUBKEY_COMPAT
+#ifdef CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID
```
jenswi-linaro, line 18: Please avoid `ifdef` where possible, it is preferred to use the `IS_ENABLED()` macro, like:
``` C
static TEE_Result get_otp_die_id(uint8_t *buffer, size_t len)
{
	if (IS_ENABLED(CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID)) {
		return tee_otp_get_die_id(buffer, len);
	} else {
		static const char pattern...
		return TEE_SUCCESS;
	}
}

Comment (rpiasetskyi) [reply core/kernel/huk_subkey.c]: Done.

File: core/kernel/huk_subkey.c [hunk 5]
```diff
@@ -15,17 +16,19 @@ static TEE_Result mac_usage(void *ctx, uint32_t usage)
...
- * implementation.
- */
-static void get_dummy_die_id(uint8_t *buffer, size_t len)
+static TEE_Result get_otp_die_id(uint8_t *buffer, size_t len)
 {
-	static const char pattern[4] = { 'B', 'E', 'E', 'F' };
-	size_t i;
+	if (IS_ENABLED(CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID)) {
+		return tee_otp_get_die_id(buffer, len);
+	} else {
```
jenswi-linaro, line 23: It seems that checkpatch doesn't like this else since we're returning above, my mistake. I guess we'll need to do it like:
``` C
        static const char pattern...

        if (IS_ENABLED(CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID))
                return tee_otp_get_die_id(buffer, len);

        for... 
        return TEE_SUCCESS;
```

rpiasetskyi: Sorry, I forgot to run checkpatch before sending the change. My bad :(

jenswi-linaro: No worries, that's why we have it in CI. :-)

mk/config.mk
```diff
@@ -648,6 +648,10 @@ endif
 # Enables backwards compatible derivation of RPMB and SSK keys
 CFG_CORE_HUK_SUBKEY_COMPAT ?= y
 
+# Use SoC specific tee_otp_get_die_id implementation for SSK key generation.
```
jforissier, line 651: - `tee_otp_get_die_id implementation()`
- You mention only the SSK but the above comment mentions both RPMB key and SSK.
- Without going through the history of the changes in the code, it is hard to understand when one would want to enable this config. How about explaining it in more details in the commit description at least? Similar to what you did in https://github.com/OP-TEE/optee_os/pull/4989#issuecomment-975439147.

rpiasetskyi: * Done
* RPMB and SSK have different generation schemes.
  Old scheme:
  `SSK = HMAC_SHA256(HUK, Chip_ID || "ONLY_FOR_tee_fs_ssk")`
  `RPMB_KEY = HMAC_SHA256(HUK, eMMC_CID)`
  
  New scheme:
  `SSK = HMAC_SHA256(HUK, HUK_SUBKEY_SSK)`
  `RPMB_KEY = HMAC_SHA256(HUK, HUK_SUBKEY_RPMB || eMMC_CID)`
  
  Where:
  `HUK` - hardware unique key
  `Chip_ID` - `get_otp_die_id`
  `HMAC_SHA256(key, value)`
  
  `Chip_ID` was only used in the old implementation of the SSK generation scheme.
* Done.

P.S. Could you please explain what was the reason for changing the generation scheme? Because I do not see a big difference but I might have missed something. Thank you in advance.

etienne-lms (changes requested):

core/kernel/huk_subkey.c
```diff
@@ -15,17 +16,18 @@ static TEE_Result mac_usage(void *ctx, uint32_t usage)
...
-/*
- * This gives the result of the default tee_otp_get_die_id()
- * implementation.
- */
-static void get_dummy_die_id(uint8_t *buffer, size_t len)
+static TEE_Result get_otp_die_id(uint8_t *buffer, size_t len)
 {
+	if (IS_ENABLED(CFG_CORE_HUK_SUBKEY_COMPAT_USE_OTP_DIE_ID))
+		return tee_otp_get_die_id(buffer, len);
+
```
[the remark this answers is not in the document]
rpiasetskyi: Done.

mk/config.mk
jforissier: Thanks for claryfing.

> P.S. Could you please explain what was the reason for changing the generation scheme? Because I do not see a big difference but I might have missed something. Thank you in advance.

I believe the main reason is to avoid code that would manipulate the HUK in an unsafe manner and could potentially leak information about it. By providing the `huk_subkey_derive()` interface, this risk is minimized. See the description of commits fa0525fa and df91a522. @jenswi-linaro

jforissier:

jenswi-linaro:

jforissier:
@rpiasetskyi thanks!

mk/config.mk
jenswi-linaro: That and also to have a structured way of generating a suitable hardware unique key for each situation to avoid key reuse and such.

PR #4991, opened 2021-11-22
mediatek: define memory range and set SHMEM base address

massonju:
This Pull Request include the following changes for Mediatek platforms:
- registers the non-secure memory to support dynamic shm registering
- set SHMEM base address after TZDRAM


core/arch/arm/plat-mediatek/conf.mk
```diff
@@ -14,6 +14,9 @@ else
 $(call force,CFG_ARM32_core,y)
 endif
 
+# default DRAM size 1 GiB
+CFG_DRAM_SIZE ?= 0x40000000
+
```
jforissier, line 19: Add `CFG_DRAM_BASE ?= 0x40000000`

core/arch/arm/plat-mediatek/main.c
```diff
@@ -17,6 +17,8 @@ register_phys_mem_pgdir(MEM_AREA_IO_NSEC,
 
 static struct serial8250_uart_data console_data;
 
+register_ddr(DRAM_BASE, DRAM_SIZE);
```
jforissier, line 20: `register_ddr(CFG_DRAM_BASE, CFG_DRAM_SIZE);`

core/arch/arm/plat-mediatek/platform_config.h
```diff
@@ -17,6 +17,9 @@
 #endif
 #endif /*ARM64*/
 
+#define DRAM_BASE		0x40000000
+#define DRAM_SIZE		CFG_DRAM_SIZE
+
```
jforissier, line 22: Not needed

jforissier (commented):
Hi @massonju,

For commit "plat-mediatek: define memory range":
- Please see comments below
- (Nit) s/ram/RAM/ in the commit description
- Does the RAM address and size default setting apply to all the platform flavors?

jforissier (commented):
For commit "plat-mediatek: set SHMEM base address after TZDRAM":

massonju:
Hi @jforissier,

Thanks for the review.

> Does the RAM address and size default setting apply to all the platform flavors?

For now the default RAM base address has always been `0x40000000`.
But I can't tell you if it will change in the future...

The default RAM size can be different.
For example this mediatek board has 2 GiB:
https://ologicinc.com/portfolio/mediateki500/

At [baylibre](https://baylibre.com) for this kind of case we set CFG_DRAM_SIZE to the corresponding value at compilation.

jforissier:
@massonju OK that sounds fine. Please apply:

...to "plat-mediatek: define memory range". Thanks!

massonju:
@jforissier thanks

PR #4992, opened 2021-11-23
[RFC v3] Remote attestation PTA

jforissier:
Remote attestation Pseudo-TA, take 3. Compared to [v2](https://github.com/OP-TEE/optee_os/pull/4972), the use case is a bit different and more straightforward:
- The attestation request comes directly from the REE client application (like in v1), not from a TA
- The TA to be measured is identified by its UUID, it doesn't have to be a running TA with an open session
- The TA hash returned by the PTA is the same as the one that is stored in the TA header, and I provide a Python script to extract it so that a client application knows what to expect for a given `.ta` file
- Signing code has been added

Comments welcome as always.


jenswi-linaro (commented):
This is simpler, but puts the burden on the caller to keep track of which dynamic libs, if any, the TA is using. Also there is no check for a running TAs, only of the available TAs. The caller has to trust that TAs aren't upgraded before they are loaded.

jforissier:
> This is simpler, but puts the burden on the caller to keep track of which dynamic libs, if any, the TA is using. Also there is no check for a running TAs, only of the available TAs.

True, I believe the use case would be more of a system configuration validation, that is you expect some version of some packages to be present (TA1 version X and TA2 version Y for example), this translates into a list of known hashes for multiple binaries.

> The caller has to trust that TAs aren't upgraded before they are loaded.

Yes, attestation then invocation is subject to TOCTOU. Definitely not as tight as the previous models (in RFC v1 and v2) which were attesting an open session, but might still be useful. TBH I am not experienced with real world use cases for remote attestation and I do not have more precise requirements for now ;-)

jenswi-linaro:
> TBH I am not experienced with real world use cases for remote attestation and I do not have more precise requirements for now ;-)

Same here. :-)

core/pta/attestation/attestation.c
```diff
@@ -0,0 +1,179 @@
...
+#include <utee_defines.h>
+
+#define PTA_NAME "attestation.pta"
+
+/* Signing key is defined in signing_key.c */
+extern uint32_t pta_attestation_key_exponent;
+extern uint8_t pta_attestation_key_modulus[];
+extern size_t pta_attestation_key_modulus_size;
+extern uint8_t pta_attestation_key_private_exponent[];
+extern size_t pta_attestation_key_private_exponent_size;
```
etienne-lms, line 24: it it not recommended to store a private key in optee core. It mandates the optee image itslef cannot be read in clear from a boot media accessible from REE which cannot be guaranteed by OP-TEE core.

It would be better that the private key gets provisioned in core secure storage.

core/pta/attestation/attestation.c
```diff
@@ -0,0 +1,179 @@
...
+	 */
+
+	res = hash_binary(uuid, out);
+	if (res)
+		return res;
+	res = digest_nonce_and_hash(digest, nonce, nonce_sz, out);
+	if (res)
+		return res;
+	return sign_digest(out + TEE_SHA256_HASH_SIZE,
+			   out_sz - TEE_SHA256_HASH_SIZE, digest);
```
etienne-lms, line 173: set `params[2].memref.size = min_out_sz;`  before returning

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,46 @@
...
+ *                           should be at least 32 + modulus size bytes. For
+ *                           example, for a 3072 bit key (384 bytes) the minimum
+ *                           buffer size is 416 bytes.
+ *
+ * Return codes:
+ * TEE_SUCCESS
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_SHORT_BUFFER - Output buffer size less than required
+ */
+#define PTA_ATTESTATION_HASH_TA		0x0
```
etienne-lms, line 44: Would be nice the API passes a Hash ID as argument.

core/pta/attestation/attestation.c
jforissier: Right

lib/libutee/include/pta_attestation.h
jforissier: Do you mean the hash algorithm?

etienne-lms: Yes, a hash algo identifier. A new dedicated command for a new hash algo would do the job, but maybe the TA API can already provide the flexibility.
I guess a command requesting a signed hash should also defines the signing scheme used, or allow a param to set the desired signing scheme.

jforissier: Totally agree but given that the initial requirements are already quite imprecise I thought I would hard-code some "reasonable" algorithms first. A more flexible command can easily be provided later.

b49020:
> > The caller has to trust that TAs aren't upgraded before they are loaded.
> 
> Yes, attestation then invocation is subject to TOCTOU. Definitely not as tight as the previous models (in RFC v1 and v2) which were attesting an open session, but might still be useful. TBH I am not experienced with real world use cases for remote attestation and I do not have more precise requirements for now ;-)

TBH, I would prefer the previous models (as in RFC v1 and v2) as I am not sure if there is any value-add to attest TAs that aren't loaded since we already have TA signature check and version rollback prevention during load time.

AFAIK remote attestation is useful for remote entity to verify that the platform is running the same code (TA in this case) as intended, so a runtime measurement would be useful IMO.

core/pta/attestation/attestation.c
b49020: Yeah for private key to be bundled in optee core, it is required to employ firmware encryption techniques like the one provided by TF-A [1].

So how about OP-TEE to rather generate this attestation key if its not already present in core secure storage during device provisioning?

[1] https://trustedfirmware-a.readthedocs.io/en/latest/design/trusted-board-boot.html#authenticated-encryption-framework

jforissier: @etienne-lms @b49020 agreed. I will update the PTA to generate the key on first call and save it to secure storage. I will also add a command to retrieve the public key.

PR #4994, opened 2021-11-23
core: dt_driver: allow driver probe/init failure without panicking

etienne-lms:
Following recent discussions, a driver initialization failure (aside from probe deferral error case) should not panic core.
These patches changes the common driver probing sequence to not panic in such case.


etienne-lms:
ping

jforissier:

etienne-lms:
tag applied

PR #4995, opened 2021-11-23
core: dt_driver: fix missing probe deferral report

etienne-lms:
During discussions around P-R https://github.com/OP-TEE/optee_os/pull/4965 I missed an instruction in the implementation: where dt_driver provider reports a probe deferral error case: that is when expected provider is not yet registered.
This is fixed by commit "core: dt_driver: return TEE_ERROR_DEFER_DRIVER_INIT if no provider".

**(edited) the driver probing test part has been removed from this P-R**
~~This P-R also proposes some test support for driver probing, based on a dedicated embedded DTB. It adds a command to the invocation test PTA, to retrieve the status of the DT drivers probing tests.~~

~~Because the test currently relies only on clock resources, a commit adds config switch `CFG_DRIVERS_CLK_EARLY_PROBE` to have clock drivers probed with other registered compatible drivers so we can test deferring a clock consumer due to its clock device dependency.~~

### DIFF
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -257,7 +257,8 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 
 		prv = dt_driver_get_provider_by_phandle(phandle);
 		if (!prv) {
-			*res = TEE_ERROR_GENERIC;
+			/* No provider registered yet */
+			*res = TEE_ERROR_DEFER_DRIVER_INIT;
 			return NULL;
 		}
 


etienne-lms:
Updated P-R to remove all the embedded test support parts (that has dependencies on other P-Rs).
Embedded DT driver probing test will be proposed later through another P-R.

This P-R now covers only the fixup commit to defer caller initialization when expected driver provider is not yet registered.

jenswi-linaro:

etienne-lms:
tag applied

jforissier:
@etienne-lms 
> Changes dt_driver_device_from_node_idx_prop() to have clk_dt_get_by_*()

Sorry I have not looked closely, is `clck_` relevant here?

etienne-lms:
Thanks to spot that. The change is expected to cover all deferrable drivers. I'll update the commit log.

etienne-lms:
commit log updated

PR #4996, opened 2021-11-23
core: reset controller driver interface

etienne-lms:
This P-R proposes a reset controller driver interface and means to register in the DT driver provider framework ~~and some embedded tests~~:
* commit "_drivers: add reset controller framework_"
* ~~commit "_core: dt_driver: test reset-controller probing_"~~

~~This P-R reuses other changes alrerady under in others P-R:~~
~~* commit "_core: dt_driver: pass driver type to clk_get_provider_by_*()_" (from [P-R 4988](https://github.com/OP-TEE/optee_os/pull/4988)) is needed for provider to possibly differentiate providers relate to a same device node in the embedded DT.~~
~~* commits "core: dt_driver: return TEE_ERROR_DEFER_DRIVER_INIT if no provider",
"drivers: clk: add CFG_DRIVERS_CLK_EARLY_PROBE" and "core: dt_driver: drivers to test probe deferral"
(from [P-R 4995](https://github.com/OP-TEE/optee_os/pull/4995)) are foundation of DT driver probing tests, needed to add test on reset controller drivers probing.~~

### DIFF
--- a/core/drivers/rstctrl/rstctrl.c
+++ b/core/drivers/rstctrl/rstctrl.c
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2021, Linaro Limited
+ */
+
+#include <assert.h>
+#include <drivers/rstctrl.h>
+#include <io.h>
+#include <kernel/dt.h>
+#include <kernel/dt_driver.h>
+#include <kernel/spinlock.h>
+#include <libfdt.h>
+#include <stdint.h>
+
+/* Global reset controller access lock */
+
+TEE_Result rstctrl_get_exclusive(struct rstctrl *rstctrl)
+{
+	uint32_t exceptions = 0;
+	TEE_Result res = TEE_ERROR_ACCESS_CONFLICT;
+	static unsigned int rstctrl_lock = SPINLOCK_UNLOCK;
+
+	exceptions = cpu_spin_lock_xsave(&rstctrl_lock);
+
+	if (!rstctrl->exclusive) {
+		rstctrl->exclusive = true;
+		res = TEE_SUCCESS;
+	}
+
+	cpu_spin_unlock_xrestore(&rstctrl_lock, exceptions);
+
+	return res;
+}
+
+void rstctrl_put_exclusive(struct rstctrl *rstctrl)
+{
+	assert(rstctrl->exclusive);
+
+	WRITE_ONCE(rstctrl->exclusive, false);
+}
+
+TEE_Result rstctrl_dt_get_by_name(const void *fdt, int nodeoffset,
+				  const char *name, struct rstctrl **rstctrl)
+{
+	int index = 0;
+
+	index = fdt_stringlist_search(fdt, nodeoffset, "reset-names", name);
+	if (index < 0)
+		return TEE_ERROR_GENERIC;
+
+	return rstctrl_dt_get_by_index(fdt, nodeoffset, index, rstctrl);
+}
--- a/core/drivers/sub.mk
+++ b/core/drivers/sub.mk
@@ -50,5 +50,6 @@ srcs-$(CFG_ZYNQMP_HUK) += zynqmp_huk.c
 subdirs-y += crypto
 subdirs-$(CFG_BNXT_FW) += bnxt
 subdirs-$(CFG_DRIVERS_CLK) += clk
+subdirs-$(CFG_DRIVERS_RSTCTRL) += rstctrl
 subdirs-$(CFG_SCMI_MSG_DRIVERS) += scmi-msg
 subdirs-y += imx
--- a/core/include/drivers/rstctrl.h
+++ b/core/include/drivers/rstctrl.h
@@ -1,0 +1,234 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2021, Linaro Limited
+ */
+
+#ifndef __DRIVERS_RSTCTRL_H
+#define __DRIVERS_RSTCTRL_H
+
+#include <kernel/dt.h>
+#include <kernel/dt_driver.h>
+#include <stdint.h>
+#include <tee_api_types.h>
+
+struct rstctrl;
+
+struct rstctrl_ops {
+	/*
+	 * Operators on reset control(s) exposed by a reset controller
+	 *
+	 * @assert_level: Assert reset level on control with a timeout hint
+	 * @deassert_level: Deassert reset level on control with a timeout hint
+	 * @get_name: Get a string name for the controller, or NULL is none
+	 *
+	 * Operator functions @assert_level and @deassert_level use arguments:
+	 * @rstctrl: Reset controller
+	 * @id: Identifier for the reset level control in the reset controller
+	 * @to_ms: Timeout in microseconds or RSTCTRL_NO_TIMEOUT, may be ignored
+	 * by reset controller.
+	 * Return a TEE_Result compliant code.
+	 */
+	TEE_Result (*assert_level)(struct rstctrl *rstctrl, unsigned int to_us);
+	TEE_Result (*deassert_level)(struct rstctrl *rstctrl,
+				     unsigned int to_us);
+	const char *(*get_name)(struct rstctrl *rstctrl);
+};
+
+/*
+ * struct rstctrl - Instance of a control exposed by a reset controller
+ * @ops: Operators of the reset controller
+ * @exclusive: Set when a consumer has exclusive control on the reset level
+ */
+struct rstctrl {
+	const struct rstctrl_ops *ops;
+	bool exclusive;
+};
+
+/**
+ * RSTCTRL_DECLARE - Declare a reset controller driver with a single
+ * device tree compatible string.
+ *
+ * @__name: Reset controller driver name
+ * @__compat: Compatible string
+ * @__probe: Reset controller probe function
+ */
+#define RSTCTRL_DT_DECLARE(__name, __compat, __probe) \
+	static const struct dt_device_match __name ## _match_table[] = { \
+		{ .compatible = __compat }, \
+		{ } \
+	}; \
+	DEFINE_DT_DRIVER(__name ## _dt_driver) = { \
+		.name = # __name, \
+		.type = DT_DRIVER_RSTCTRL, \
+		.match_table = __name ## _match_table, \
+		.probe = __probe, \
+	}
+
+/*
+ * Platform driver may ignore the timeout hint according to their
+ * capabilities. RSTCTRL_NO_TIMEOUT specifies no timeout hint.
+ */
+#define RSTCTRL_NO_TIMEOUT	0
+
+/*
+ * rstctrl_assert_to - Assert reset control possibly with timeout
+ * rstctrl_assert - Assert reset control
+ * rstctrl_deassert_to - Deassert reset control possibly with timeout
+ * rstctrl_deassert - Deassert reset control
+ *
+ * @rstctrl: Reset controller
+ * @to_us: Timeout in microseconds
+ * Return a TEE_Result compliant code
+ */
+static inline TEE_Result rstctrl_assert_to(struct rstctrl *rstctrl,
+					   unsigned int to_us)
+{
+	return rstctrl->ops->assert_level(rstctrl, to_us);
+}
+
+static inline TEE_Result rstctrl_assert(struct rstctrl *rstctrl)
+{
+	return rstctrl_assert_to(rstctrl, RSTCTRL_NO_TIMEOUT);
+}
+
+static inline TEE_Result rstctrl_deassert_to(struct rstctrl *rstctrl,
+					     unsigned int to_us)
+{
+	return rstctrl->ops->deassert_level(rstctrl, to_us);
+}
+
+static inline TEE_Result rstctrl_deassert(struct rstctrl *rstctrl)
+{
+	return rstctrl_deassert_to(rstctrl, RSTCTRL_NO_TIMEOUT);
+}
+
+/*
+ * rstctrl_name - Get a name for the reset level control or NULL
+ *
+ * @rstctrl: Reset controller
+ * Return a pointer to controller name or NULL
+ */
+static inline const char *rstctrl_name(struct rstctrl *rstctrl)
+{
+	if (rstctrl->ops->get_name)
+		return rstctrl->ops->get_name(rstctrl);
+
+	return NULL;
+}
+
+/**
+ * rstctrl_dt_get_exclusive - Get exclusive access to reset controller
+ *
+ * @rstctrl: Reset controller
+ * Return a TEE_Result compliant value
+ */
+TEE_Result rstctrl_get_exclusive(struct rstctrl *rstctrl);
+
+/**
+ * rstctrl_put_exclusive - Release exclusive access to target
+ *
+ * @rstctrl: Reset controller
+ */
+void rstctrl_put_exclusive(struct rstctrl *rstctrl);
+
+/**
+ * rstctrl_ops_is_valid - Check reset controller ops is valid
+ *
+ * @ops: Reference to reset controller operator instance
+ */
+static inline bool rstctrl_ops_is_valid(const struct rstctrl_ops *ops)
+{
+	return ops && ops->assert_level && ops->deassert_level;
+}
+
+#ifdef CFG_DT
+/**
+ * rstctrl_dt_get_by_index - Get a reset controller at a specific index in
+ * 'resets' property
+ *
+ * @fdt: Device tree to work on
+ * @nodeoffset: Node offset of the subnode containing a 'resets' property
+ * @index: Reset controller index in 'resets' property
+ * @rstctrl: Output reset controller reference upon success
+ *
+ * Return TEE_SUCCESS in case of success
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if reset controller is not initialized
+ * Return a TEE_Result compliant code in case of error
+ */
+static inline TEE_Result rstctrl_dt_get_by_index(const void *fdt,
+						 int nodeoffset,
+						 unsigned int index,
+						 struct rstctrl **rstctrl)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	*rstctrl = dt_driver_device_from_node_idx_prop("resets", fdt,
+						       nodeoffset, index,
+						       DT_DRIVER_RSTCTRL, &res);
+	return res;
+}
+#else
+static inline TEE_Result rstctrl_dt_get_by_index(const void *fdt __unused,
+						 int nodeoffset __unused,
+						 unsigned int index __unused,
+						 struct rstctrl **rstctrl)
+{
+	*rstctrl = NULL;
+	return TEE_ERROR_NOT_SUPPORTED;
+}
+#endif /*CFG_DT*/
+
+/**
+ * rstctrl_dt_get_by_name - Get a reset controller matching a name in the
+ * 'reset-names' property
+ *
+ * @fdt: Device tree to work on
+ * @nodeoffset: Node offset of the subnode containing a 'resets' property
+ * @name: Reset controller name to get
+ * @rstctrl: Output reset controller reference upon success
+ *
+ * Return TEE_SUCCESS in case of success
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if reset controller is not initialized
+ * Return a TEE_Result compliant code in case of error
+ */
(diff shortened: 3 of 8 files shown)


etienne-lms:
I updated the P-R, removing the embedded test part that will come later.
Remains a single commit that propose a generic API for supporting reset controller in OP-TEE.

etienne-lms:
Rebased on master.

Any comment on that. Please tell me is this seems ok as a first step for reset controller support, or if it rather looks like a bad idea or API design.

Checkpatch `CHECK` report is not blocking IMO:
```
CHECK: Macro argument reuse '_ops' - possible side-effects?
#202: FILE: core/include/drivers/rstctrl.h:72:
+#define RSTCTRL_OPS_IS_VALID(_ops) \
+	((_ops)->assert_level && (_ops)->deassert_level)
```

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,217 @@
...
+		{ } \
+	}; \
+	const struct dt_driver __name ## _dt_driver __dt_driver = { \
+		.name = # __name, \
+		.type = DT_DRIVER_RSTCTRL, \
+		.match_table = __name ## _match_table, \
+		.probe = __probe, \
+	}
+
+#define RSTCTRL_OPS_IS_VALID(_ops) \
```
jenswi-linaro, line 72: Any reason why this is a macro instead of an inline function? The latter should take care of the checkpatch issue.

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,217 @@
...
+
+/*
+ * struct rstctrl - Instance of a control exposed by a reset controller
+ * @ops: Operators of the reset controller
+ * @priv: Reset control private data (unsigned value or memory pointer)
+ * @exclusive: Set when a consumer has exclusive control on the reset level
+ */
+struct rstctrl {
+	const struct rstctrl_ops *ops;
+	uintptr_t priv;
```
jenswi-linaro, line 48: Do we need this field? I guess it wouldn't if we intended to include this struct in a larger device specific struct.

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,217 @@
...
+	bool exclusive;
```
jenswi-linaro, line 49: How will this field be used? Is it needed for all instances?

core/include/drivers/rstctrl.h
etienne-lms: Indeed, could be removed.

core/include/drivers/rstctrl.h
etienne-lms: In case a reset controller is shared by several devices, one may request exclusive access to the controller when using the related resource and release exclusive access once resource is no more used. Likely needed only for specific shared reset controllers. I've added here to have a generic structure.

core/include/drivers/rstctrl.h
etienne-lms: Right, I'll move to an inline function. (I should have done it seen checkpatch report)

core/include/drivers/rstctrl.h
etienne-lms: > How will this field be used? 

Sorry, I didn't answer that question. It is used through `rstctrl_{get|put}_exclusive()`.

core/drivers/rstctrl/rstctrl.c
```diff
@@ -0,0 +1,63 @@
...
+	cpu_spin_unlock_xrestore(&rstctrl_lock, exceptions);
+
+	return res;
+}
+
+void rstctrl_put_exclusive(struct rstctrl *rstctrl)
+{
+	assert(rstctrl->exclusive);
+
+	rstctrl->exclusive = false;
```
jenswi-linaro, line 38: I would prefer if this was assigned using a `WRITE_ONCE()` macro. We don't have that macro yet, but it should be easy to implement using `__compiler_atomic_store()`

core/drivers/rstctrl/rstctrl.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+struct rstctrl *rstctrl_dt_get_by_idx(const void *fdt, int nodeoffset,
+				      unsigned int idx, TEE_Result *res)
+{
+	void *device = dt_driver_device_from_node_idx_prop("resets", fdt,
+							   nodeoffset, idx,
+							   res);
+
+	return (struct rstctrl *)device;
```
jenswi-linaro, line 48: Cast is not needed.

core/include/drivers/rstctrl.h
jenswi-linaro: Thanks

etienne-lms:
comments addressed

jenswi-linaro:
For "io.h: add WRITE_ONCE macro" please apply:

For the rest:

etienne-lms:
Many thanks for the review and feedback.
Review tags applied.

etienne-lms:
I've added a fixup commit.

jenswi-linaro:
The fixup looks good:

etienne-lms:
thanks @jenswi-linaro, i'll squash it.

jforissier:
The state of this PR looks good enough to be merged but since it is a RFC perhaps more feedback is needed. @clementleger since you submitted #5028, what do you think? Does this look useful to you? Thanks.

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,228 @@
...
+	return ops && ops->assert_level && ops->deassert_level;
+}
+
+/**
+ * rstctrl_dt_get_by_idx - Get a reset controller at a specific index in
+ * 'resets' property
+ *
+ * @fdt: Device tree to work on
+ * @nodeoffset: Node offset of the subnode containing a 'resets' property
+ * @index: Reset controller index in 'resets' property
```
clementleger, line 153: should be `idx`

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,228 @@
...
+static inline TEE_Result rstctrl_deassert(struct rstctrl *rstctrl)
+{
+	return rstctrl_deassert_to(rstctrl, RSTCTRL_NO_TIMEOUT);
+}
+
+/*
+ * rstctrl_name - Get a name for the reset level control or NULL
+ *
+ * @rstctrl: Reset controller
+ * Return a pointer to  anem or NULL
```
clementleger, line 112: typo: `anem`

core/drivers/sub.mk
```diff
@@ -47,5 +47,6 @@ srcs-$(CFG_ZYNQMP_HUK) += zynqmp_huk.c
 subdirs-y += crypto
 subdirs-$(CFG_BNXT_FW) += bnxt
 subdirs-$(CFG_DRIVERS_CLK) += clk
+subdirs-$(CFG_DRIVERS_RSTCTRL) += rstctrl
```
clementleger, line 50: Shouldn't `CFG_DRIVERS_RSTCTRL` be documented in `config.mk` ?

clementleger (commented):
This support seems a good starting point but the lack of multiple reset line support probably lacks.

etienne-lms:
> Driver I submitted is a global reset for the SoC. This RFC targets reset controller which can reset individual device AFAIU.

I fully agree.

> EDIT: I would say that it could be used to reset the chip but then, maybe some additional property should be added (...)
> Or maybe something on top of that could make use of it to reset the chip and this generic code could be used by PSCI for reset for instance.

I would not go that way. Reset controllers are designed to be use by consumers: drivers (using a dt "resets = " property) over a reset controller provider driver.  PSCI does not get system or CPU reset controllers from DT, not for the CPU of the executing REE/TEE.

> This support seems a good starting point but lacks support of multiple reset line per controller.

Either consumer driver store both reset controller instance reference (pointer) and a controller ID.
=> `rstctrl_[de]assert()` gets an extra argument.
Or the reset controller is 1 instance per reset line: this proposal.
=> current way, diver use private means to bind the struct rstctrl * reference to a target line.
Alternate to current state: restore `priv` field in `struct rstctrl` to hold the reset line ref in the reset controller driver.

core/include/drivers/rstctrl.h
etienne-lms: do we restore this generic private ref for the driver?

In current series state, it is up to the platform reset controller driver to bind the struct address to its local reset line reference. 

Example: planned for [_stm32_reset.c_ (ephemeral link)](https://github.com/etienne-lms/optee_os/commit/2e12ef74f202562574fff9dc42009d3145b73d67?diff=unified#diff-1ceec50dd256ff8eba2db4dd1c4f52dd1ba4214b40b3e5af0ea89966c4748d1bR22)
```c
/* Exposed rstctrl instance */
struct stm32_rstline {
	unsigned int id;
	struct rstctrl rstctrl;
	SLIST_ENTRY(stm32_rstline) link;
};

static struct stm32_rstline *to_rstline(struct rstctrl *rstctrl)
{
	assert(rstctrl);

	return container_of(rstctrl, struct stm32_rstline, rstctrl);
}
```

Adding `priv` in `struct rstctrl` increases the struct size, may be be easier for platforms willing to save a reference (local pointer or numeric identifer.
Note the numeric identifer must be set by the provider driver: it may be the DT binding Id for 1 cell `#reset-cells` providers, it may need a conversion when DT binding rely or more IDs (`#reset-cells = 2;`).

core/drivers/sub.mk
etienne-lms: yes it should. i'll add. thanks.

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,228 @@
...
+	/*
+	 * Operators on reset control(s) exposed by a reset controller
+	 *
+	 * assert_level: Assert reset level on control with a timeout hint
+	 * deassert_level: Deassert reset level on control with a timeout hint
+	 * get_name: Get a string name for the controller, or NULL is none
+	 *
+	 * @rstctrl: Reset control reference
+	 * @id: Identifier for the reset level control in the reset controller
+	 * @to_ms: Timeout in microsencond or RSTCTRL_NO_TIMEOUT, may be ignored
```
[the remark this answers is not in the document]
etienne-lms: > id and to_ms are not parameters

Right, I described them here since common arguments to the operators.
You would prefer another way to describe them?

> BTW, I do not see where you can specify the reset line to use. 

Consumer calls `rstctrl_dt_get_by_index()` (or `rstctrl_dt_get_by_name()`, as clocks) and gets a reset line, I.e. a `struct rstctrl` pointer.
It then calls `rstctrl_assert()` to request request.
It may (often) need to call `rstctrl_deassert()` when the line shall be manually released.

Should rstctrl_dt_get_by_index() return a rstctrl reference + a line ID reference, to be used in rstctrl_*() API functions?

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,228 @@
...
+ * @exclusive: Set when a consumer has exclusive control on the reset level
+ */
+struct rstctrl {
+	const struct rstctrl_ops *ops;
+	bool exclusive;
+};
+
+/**
+ * RSTCTRL_DECLARE - Declare a reset controller driver with a single
+ * DTT compatible string.
```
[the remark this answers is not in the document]
etienne-lms: will fix. thanks.

core/include/drivers/rstctrl.h
etienne-lms: thanks

core/include/drivers/rstctrl.h
etienne-lms: right, thanks.

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @ops: Reference to reset controller operator instance
+ */
+static inline bool rstctrl_ops_is_valid(const struct rstctrl_ops *ops)
+{
+	return ops && ops->assert_level && ops->deassert_level;
+}
+
+/**
+ * rstctrl_dt_get_by_idx - Get a reset controller at a specific index in
```
etienne-lms, line 148: I want to rename this to: `dt_get_rstctrl_by_index()` with get `TEE_result` return value:
```c
TEE_Result dt_get_rstctrl_by_index(const void *fdt, int nodeoffset,
				   unsigned int index,
				   struct rstctrl **rstctrl);
```
It will look better in caller's instruction block:
```c
	struct rstctrl *rstctrl_ref = NULL;

	res = dt_get_rstctrl_by_index(fdt, node, index, &rstctrl_ref);
	if (res)
		return res;
	res = rstctrl_get_exclisive(rstctrl_ref);
	if (res)
		return res;
	(...)
```

I'll propose the same for clk: 
```c
TEE_Result dt_get_clk_by_index(const void *fdt, int nodeoffset,
			       unsigned int index, struct clk **clk);
TEE_Result dt_get_clk_by_name(const void *fdt, int nodeoffset,
			      chanr *name, struct clk **clk);
```

etienne-lms (commented):
Many thanks for the feedback and review.

core/include/drivers/rstctrl.h
clementleger: Looks good to me for the clock functions.

etienne-lms:
@clementleger, regarding `priv` field in the reset controller structure (`struct rstctrl`), would you you prefer the field inside the structure or let platform driver include `struct rstcltr` inside its own instance structure (and use `container_of()` like means, as in current P-R state)?

clementleger:
Maybe letting the platform driver decides where to put it will allow more flexibility ? `container_of()` will allow all of them to get it easily and it will also allow all drivers to put this in the memory they want.

etienne-lms:
Do I understand the current proposal is fine, without a `priv` field in `struct rstctrl`?
```c
struct rstctrl {
	const struct rstctrl_ops *ops;
	bool exclusive;
};
```

clementleger:
> Do I understand the current proposal is fine, without a `priv` field in `struct rstctrl`?
> 
> ```c
> struct rstctrl {
> 	const struct rstctrl_ops *ops;
> 	bool exclusive;
> };
> ```

Yes !

etienne-lms:
ok, thanks to confirm. I'll update this P-R and create one for the clock APIs.

etienne-lms:
Comments addressed.

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,227 @@
...
+static inline const char *rstctrl_name(struct rstctrl *rstctrl)
+{
+	if (rstctrl->ops->get_name)
+		return rstctrl->ops->get_name(rstctrl);
+
+	return NULL;
+}
+
+/**
+ * rstctrl_dt_get_exclusive - Get exclusive control or target
```
clementleger, line 124: typo: `of` instead of `or`

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,227 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2021, Linaro Limited
+ *
+ * Self contained reset controller framework in rstctrl.h
```
clementleger, line 5: Is it still self contained in the `.h` since there is a `rstctrl.c` file ?

mk/config.mk
```diff
@@ -727,6 +727,13 @@ CFG_DRIVERS_CLK ?= n
 CFG_DRIVERS_CLK_DT ?= $(call cfg-all-enabled,CFG_DRIVERS_CLK CFG_DT)
 CFG_DRIVERS_CLK_FIXED ?= $(CFG_DRIVERS_CLK_DT)
 
+$(eval $(call cfg-depends-all,CFG_DRIVERS_CLK_DT,CFG_DRIVERS_CLK CFG_DT))
+$(eval $(call cfg-depends-all,CFG_DRIVERS_CLK_FIXED,CFG_DRIVERS_CLK_DT))
+
+# When enabled, CFG_DRIVERS_RSTCTRL embeds a reset controller framework in
+# OP-TEE core to provide reset controls on subsystems of the devices.
+CFG_DRIVERS_RSTCTRL ?= n
```
clementleger, line 735: Don't you have a dependency on `CFG_DT` ?

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,227 @@
...
+	if (rstctrl->ops->get_name)
+		return rstctrl->ops->get_name(rstctrl);
+
+	return NULL;
+}
+
+/**
+ * rstctrl_dt_get_exclusive - Get exclusive control or target
+ *
+ * @rstctrl: Reference to reset level control
```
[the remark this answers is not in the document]
etienne-lms: ok

core/include/drivers/rstctrl.h
etienne-lms: yes

core/include/drivers/rstctrl.h
```diff
@@ -0,0 +1,227 @@
...
+ * Return a TEE_Result compliant value
+ */
+TEE_Result rstctrl_get_exclusive(struct rstctrl *rstctrl);
+
+/**
+ * rstctrl_put_exclusive - Release exclusive access to target
+ *
+ * @rstctrl: Reference to reset level control
```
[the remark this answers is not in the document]
etienne-lms: ok

mk/config.mk
etienne-lms: There should not be but for the rstctrl_dt_get_by_`*()` API functions.
I'll check that it build ok with and without `CFG_DT`.

etienne-lms:
Comments addressed.
I've rebased the series with its fixup commits to latest master to check build when `CFG_DT` is disabled.
I've added some more fixup commits to address latest review comments.

jforissier:
For commit "mk/config.mk: fix layout":
s/a assigned/are assigned/, then

etienne-lms:
Updated the series with reviewed commits + non-reviewed changes (API function change) as a fixup commit.

etienne-lms:
CI seems to fails on a [python errors](https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=2068&view=logs&j=705748f3-7146-5e86-79af-1c0266d20a8c&t=d3458a7d-1a81-5639-8e63-8927d807e53a&l=24447) in the build log?
```
ImportError: cannot import name 'SystemDependency' from 'mesonbuild.dependencies.base' (snip)
```

jforissier:
> CI seems to fails on a [python errors](https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=2068&view=logs&j=705748f3-7146-5e86-79af-1c0266d20a8c&t=d3458a7d-1a81-5639-8e63-8927d807e53a&l=24447) in the build log?
> 
> ```
> ImportError: cannot import name 'SystemDependency' from 'mesonbuild.dependencies.base' (snip)
> ```

Yes, I think it is just a matter of rebuilding the Docker image (Buildroot changed and this required cleaning or rebuilding from scratch). I'm on it.

etienne-lms:
thanks @jforissier.

etienne-lms:
@jenswi-linaro, sorry to bother. You've acked the main commit but from discussions, I changed an API function prototype and a fixup commit remains un-reviewed.
Could you ack the fixup if looking good to you, so that i could get the whole merged?

core/include/drivers/rstctrl.h
```diff
@@ -141,30 +141,42 @@ static inline bool rstctrl_ops_is_valid(const struct rstctrl_ops *ops)
...
- * index in which case @res provides the error code.
+ * Return TEE_SUCCESS in case of success
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if reset controller is not initialized
+ * Return a TEE_Result compliant code in case of error
  */
-static inline struct rstctrl *rstctrl_dt_get_by_idx(const void *fdt,
-						    int nodeoffset,
-						    unsigned int idx,
-						    TEE_Result *res)
+static inline TEE_Result rstctrl_dt_get_by_index(const void *fdt,
```
jenswi-linaro, line 158: Mindor: How about keeping this name as `rstctrl_dt_get_by_idx()` instead? Same for the idx parameter.

jenswi-linaro (commented):
Only a minor comment on the fixup. With that fixed or not:

core/include/drivers/rstctrl.h
etienne-lms: It is rather for consistency with the change made in clk_dt_get_by*() that I change here `idx` to plain `index`.

etienne-lms:
Thanks for the review.
Tags applied.

jforissier:
@clementleger any more feedback?

clementleger:
Nope, that is ok for me.

etienne-lms:
I need to rebase since introduction of macro `DEFINE_DT_DRIVER()`

PR #4999, opened 2021-11-24
[RFC] Use -Wdeclaration-after-statement in CI

jenswi-linaro:

### DIFF
--- a/core/tee/socket.c
+++ b/core/tee/socket.c
@@ -19,9 +19,10 @@ static uint32_t get_instance_id(struct ts_session *sess)
 static TEE_Result socket_open(uint32_t instance_id, uint32_t param_types,
 			      TEE_Param params[TEE_NUM_PARAMS])
 {
-	struct mobj *mobj;
-	TEE_Result res;
-	void *va;
+	struct thread_param tpm[4] = { };
+	struct mobj *mobj = NULL;
+	TEE_Result res = TEE_SUCCESS;
+	void *va = NULL;
 	uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
 					  TEE_PARAM_TYPE_MEMREF_INPUT,
 					  TEE_PARAM_TYPE_VALUE_INPUT,
@@ -41,16 +42,13 @@ static TEE_Result socket_open(uint32_t instance_id, uint32_t param_types,
 
 	memcpy(va, params[1].memref.buffer, params[1].memref.size);
 
-	struct thread_param tpm[4] = {
-		[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_SOCKET_OPEN,
-					 instance_id, 0),
-		[1] = THREAD_PARAM_VALUE(IN,
-				params[0].value.b, /* server port number */
-				params[2].value.a, /* protocol */
-				params[0].value.a  /* ip version */),
-		[2] = THREAD_PARAM_MEMREF(IN, mobj, 0, params[1].memref.size),
-		[3] = THREAD_PARAM_VALUE(OUT, 0, 0, 0),
-	};
+	tpm[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_SOCKET_OPEN, instance_id, 0);
+	tpm[1] = THREAD_PARAM_VALUE(IN,
+				    params[0].value.b, /* server port number */
+				    params[2].value.a, /* protocol */
+				    params[0].value.a  /* ip version */);
+	tpm[2] = THREAD_PARAM_MEMREF(IN, mobj, 0, params[1].memref.size);
+	tpm[3] = THREAD_PARAM_VALUE(OUT, 0, 0, 0);
 
 	res = thread_rpc_cmd(OPTEE_RPC_CMD_SOCKET, 4, tpm);
 	if (res == TEE_SUCCESS)
@@ -62,6 +60,7 @@ static TEE_Result socket_open(uint32_t instance_id, uint32_t param_types,
 static TEE_Result socket_close(uint32_t instance_id, uint32_t param_types,
 			       TEE_Param params[TEE_NUM_PARAMS])
 {
+	struct thread_param tpm = { };
 	uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
 					  TEE_PARAM_TYPE_NONE,
 					  TEE_PARAM_TYPE_NONE,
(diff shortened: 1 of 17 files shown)


core/core.mk
```diff
@@ -60,6 +60,10 @@ ifneq ($(CFG_TEE_CORE_DEBUG),y)
 cppflags$(sm)  += -DNDEBUG
 endif
 
+ifeq ($(CFG_WARN_DECL_AFTER_STATEMENT),y)
+cflags$(sm)	+= $(call cc-option,-Wdeclaration-after-statement)
+endif
```
jforissier, line 65: Since you have fixed all occurrences in the prior patches, perhaps this could be done always (no need for a CFG_)?

jenswi-linaro: I was thinking the same at first, but there's the problem with out of tree code. The optee_os/core part may not be so bad but this would have the potential to affect TAs built with TA dev-kit. So just as with `CFG_WERROR` I thought would be safer to let this be disabled by default.

jforissier: This sets the flags for the core only, so there should be no problem for external TAs. And IMO in-tree TAs should have the flag too:
```diff
diff --git a/ta/mk/build-user-ta.mk b/ta/mk/build-user-ta.mk
index 8138d474..7a8c1c1c 100644
--- a/ta/mk/build-user-ta.mk
+++ b/ta/mk/build-user-ta.mk
@@ -30,6 +30,7 @@ include mk/$(COMPILER_$(sm)).mk
 
 cppflags$(sm)  := $(cppflags$(ta-target)) -I$(ta-dev-kit-dir$(sm))/include
 cflags$(sm)    := $(cflags$(ta-target))
+cflags$(sm)    += $(call cc-option,-Wdeclaration-after-statement)
 aflags$(sm)    := $(aflags$(ta-target))
 
 ifeq ($(CFG_ULIBS_SHARED),y)
```

Another option would be to add it to `mk/compile.mk` only so that any module has it, but guarded with the `CFG_` so that it can be disabled when TA dev kit is used externally:
```diff
diff --git a/mk/compile.mk b/mk/compile.mk
index 9803f7e2..1e60cfe6 100644
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -36,6 +36,9 @@ comp-cflags-warns-high = \
        -Wshadow -Wstrict-prototypes -Wswitch-default \
        -Wwrite-strings \
        -Wno-missing-field-initializers -Wno-format-zero-length
+ifeq ($(CFG_WARN_DECL_AFTER_STATEMENT),y)
+comp-cflags-warn-high += $(call cc-option,-Wdeclaration-after-statement)
+endif
 comp-cflags-warns-medium = \
        -Wredundant-decls
 comp-cflags-warns-low = \
```

jenswi-linaro: I like the second option better, but we could enable `CFG_WARN_DECL_AFTER_STATEMENT` by default instead. I'll give a try, stay tuned... :-)

core/tee/tadb.c
```diff
@@ -709,10 +708,8 @@ static TEE_Result ta_load(struct tee_tadb_ta_read *ta)
 	ta->ta_buf = mobj_get_va(ta->ta_mobj, 0);
 	assert(ta->ta_buf);
 
-	struct thread_param params[] = {
-		[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_FS_READ, ta->fd, 0),
-		[1] = THREAD_PARAM_MEMREF(OUT, ta->ta_mobj, 0, sz),
-	};
+	params[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_FS_READ, ta->fd, 0),
+	params[1] = THREAD_PARAM_MEMREF(OUT, ta->ta_mobj, 0, sz),
```
etienne-lms, line 712: typo: s/,/;/ here and above

core/tee/socket.c
```diff
@@ -105,13 +104,11 @@ static TEE_Result socket_send(uint32_t instance_id, uint32_t param_types,
...
-					 params[0].value.a /* handle */),
-		[1] = THREAD_PARAM_MEMREF(IN, mobj, 0, params[1].memref.size),
-		[2] = THREAD_PARAM_VALUE(INOUT, params[0].value.b, /* timeout */
-					 0, 0),
-	};
+	tpm[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_SOCKET_SEND, instance_id,
+				    params[0].value.a /* handle */);
+	tpm[1] = THREAD_PARAM_MEMREF(IN, mobj, 0, params[1].memref.size),
+	tpm[2] = THREAD_PARAM_VALUE(INOUT, params[0].value.b, /* timeout */
+				     0, 0),
```
etienne-lms, line 111: typo: s/,/;/ here and above

core/tee/socket.c
```diff
@@ -144,13 +142,10 @@ static TEE_Result socket_recv(uint32_t instance_id, uint32_t param_types,
...
-		[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_SOCKET_RECV, instance_id,
-					 params[0].value.a /* handle */),
-		[1] = THREAD_PARAM_MEMREF(OUT, mobj, 0, params[1].memref.size),
-		[2] = THREAD_PARAM_VALUE(IN, params[0].value.b /* timeout */,
-					 0, 0),
-	};
+	tpm[0] = THREAD_PARAM_VALUE(IN, OPTEE_RPC_SOCKET_RECV, instance_id,
+				    params[0].value.a /* handle */),
+	tpm[1] = THREAD_PARAM_MEMREF(OUT, mobj, 0, params[1].memref.size),
+	tpm[2] = THREAD_PARAM_VALUE(IN, params[0].value.b /* timeout */, 0, 0),
```
etienne-lms, line 148: ditto

jenswi-linaro:
Update with `CFG_WARN_DECL_AFTER_STATEMENT` enabled by default, both for core and the in-tree TAs. It should not be enabled by default for out of tree TAs, those in xtest for instance.

etienne-lms (commented):
**(edited: updated with explicit reviewed commit list)**
* libmbedtls: disable -Wdeclaration-after-statement 
* libtomcrypt: disable -Wdeclaration-after-statement 
* libutee: Fix warning in bigint_binary_mod() 
* libutee: Fix warning in TEE_BigIntInvMod() 
* core: fix warning in tee_time_get_ree_time() 
* core: Fix warning in mobj_reg_shm_inc_map() 
* core: Fix warning in find_chosen_node() 
* core: Fix warning in __wq_rpc() 
* core: Fix warning in register_shm() 
* core: Fix warnings in core/tee/tee_svc_cryp.c 
* core: Fix warnings in core/tee/tee_fs_rpc.c 
* core: Fix warning in gprof_send_rpc()
* core: Fix warnings in core/kernel/lockdep.c

Comments for commits:
* "core: Fix warnings in core/tee/tadb.c" and "core: Fix warnings in core/tee/socket.c"

jforissier:
Looks good! For commit "Add config option CFG_WARN_DECL_AFTER_STATEMENT":

jenswi-linaro:
Comments addressed and tags applied. Note that a few commit still needs to be reviewed.

etienne-lms:
* "Add config option CFG_WARN_DECL_AFTER_STATEMENT"
* "avb: Fix warnings in ta/avb/entry.c"
* "core: Fix warnings in core/tee/socket.c"
* "core: Fix warnings in core/tee/tadb.c"

jenswi-linaro:
Update with the last tags. All commits should have been reviewed now.

PR #5001, opened 2021-11-25
plat-stm32mp1: BSEC drivers fixes and updates

etienne-lms:

### DIFF
--- a/core/drivers/stm32_bsec.c
+++ b/core/drivers/stm32_bsec.c
@@ -672,4 +672,4 @@ static TEE_Result initialize_bsec(void)
 	return TEE_SUCCESS;
 }
 
-driver_init(initialize_bsec);
+early_init(initialize_bsec);
--- a/core/drivers/stm32_bsec.c
+++ b/core/drivers/stm32_bsec.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-3-Clause
 /*
- * Copyright (c) 2017-2020, STMicroelectronics
+ * Copyright (c) 2017-2021, STMicroelectronics
  */
 
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


core/include/drivers/stm32_bsec.h
```diff
@@ -10,6 +10,9 @@
 #include <stdint.h>
 #include <tee_api.h>
 
+#define BSEC_BITS_PER_WORD		(8U * sizeof(uint32_t))
+#define BSEC_BYTES_PER_WORD		(sizeof(uint32_t))
```
jforissier, line 14: This one is not used

jforissier (commented):
- For commit "drivers: stm32_bsec: define BSEC word size in header file":
You do not say why you move the definition to the header file. Is this patch really useful? Which C file other than `stm32_bsec.h` (Edit: I mean stm32_bsec.**c**) would need this definition?
- For other commits:

etienne-lms (commented):
> You do not say why you move the definition to the header file. Is this patch really useful? Which C file other than stm32_bsec.h would need this definition?

Well.. my apologies, dinging a bit, it's useless. I'll discard the change. Thanks for pointing that :), Further change are expected to provide more flexibility in OTPs identification. But actaully this change in even useless in that scope.
Further change will ease definition of OTPs using a nvmem layout, binding nvmem areas (OTP ranges) to identifiers (string names). This

etienne-lms:
Removed commit "drivers: stm32_bsec: define BSEC word size in header file" from the series.
Applied tag to commits.

etienne-lms:
Right, i'll fix the commit message.
Thanks

etienne-lms:
Commit log fixed.
Tags applied.

PR #5002, opened 2021-11-25
dt drive rprobing: probe before/after crypto initialization

etienne-lms:
This series changes the driver probing sequence to probe drivers before and after core crypto and rng services are initialized.

Driver registering the the drvcrypt framework must be registered before crypto API is used. Drivers depending on the crypto service should be deferred to after crypto init.

### DIFF
--- a/core/include/kernel/dt_driver.h
+++ b/core/include/kernel/dt_driver.h
@@ -82,6 +82,21 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 					  TEE_Result *res);
 
 /*
+ * dt_driver_get_crypto() - Request crypto support for driver initialization
+ *
+ * Return TEE_SUCCESS if cryptography services are initialized, otherwise return
+ * TEE_ERROR_DEFER_DRIVER_INIT.
+ */
+TEE_Result dt_driver_get_crypto(void);
+
+#ifdef CFG_DT
+/* Inform DT driver probe sequence that core crypto support is initialized */
+void dt_driver_crypt_init_complete(void);
+#else
+static inline void dt_driver_crypt_init_complete(void) {}
+#endif
+
+/*
  * Return driver provider reference from its node offset value in the FDT
  */
 struct dt_driver_provider *
--- a/core/tee/tee_cryp_utl.c
+++ b/core/tee/tee_cryp_utl.c
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2014, Linaro Limited
+ * Copyright (c) 2014-2021, Linaro Limited
  * Copyright (c) 2021, SumUp Services GmbH
  */
 
 #include <crypto/crypto.h>
 #include <initcall.h>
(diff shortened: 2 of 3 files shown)


toromanoSTM (commented):
minor typo.

etienne-lms:
Fixed inline description typos.

etienne-lms:
Squashed fixup commits and rebased to run CI with regression_2004.4.2 failure fixed (https://github.com/OP-TEE/optee_test/issues/569).

etienne-lms:
rebased (resolved merge conflict)

core/tee/tee_cryp_utl.c
```diff
@@ -16,6 +16,13 @@
 #include <trace.h>
 #include <utee_defines.h>
 
+static bool tee_cryp_inited;
```
jenswi-linaro, line 19: This is something internal to the DT driver probing. Could it be kept there instead?
`probe_dt_drivers_early()` and `probe_dt_drivers()` already depends on `tee_cryp_init()` being called somewhere in between so it could just as well be assumed that it has been called when `probe_dt_drivers()` is called.

etienne-lms: Would be ok but I feared _tee_cryp_util.c_ is modified, for example changing `tee_cryp_init()` from `service_init` to another initcall level. Do you think it's worth to assume it won't change?

jenswi-linaro: OK, so this is for debug purposes.
How about calling a `dt_driver_crypt_init_complete()` at the end of `tee_cryp_init()`? In case the DT driver framework isn't used it can do nothing else take some action in case asserts are enabled or something.

etienne-lms: +1, thanks for the suggestion.

core/include/kernel/dt_driver.h
```diff
@@ -12,7 +12,9 @@
 #include <kernel/dt.h>
 #include <stdint.h>
 #include <sys/queue.h>
+#include <tee_api_defines_extensions.h>
 #include <tee_api_types.h>
+#include <tee/tee_cryp_utl.h>
```
jenswi-linaro, line 17: This might not be needed any longer.

jenswi-linaro (commented):

core/include/kernel/dt_driver.h
etienne-lms: done

etienne-lms:
Fixup commit squashed + rebased (deps on 670fa4aa34cbd8b66424ba6d121f26a45a0b7a76).

etienne-lms:
I didn't apply the review tag as I've modified the commit logs.

jenswi-linaro:

etienne-lms:
Tag applied. @toromanoSTM, are you ok with the changes?

toromanoSTM:
Ok with the changes.
`Reviewed-by: Nicolas Toromanoff <nicolas.toromanoff@foss.st.com>`

etienne-lms:
tags applied. thanks.

PR #5006, opened 2021-11-26
plat-sam: add sama5d2 TRNG driver

clementleger:
Add TRNG driver for sama5d2.

### DIFF
--- a/core/arch/arm/plat-sam/conf.mk
+++ b/core/arch/arm/plat-sam/conf.mk
@@ -42,3 +42,10 @@ CFG_TEE_RAM_VA_SIZE ?= 0x100000
 # Device tree related configuration
 CFG_DT_ADDR ?= 0x21500000
 CFG_GENERATE_DTB_OVERLAY ?= y
+
+CFG_WITH_SOFTWARE_PRNG ?= n
+CFG_ATMEL_TRNG ?= y
+ifeq ($(CFG_ATMEL_TRNG),y)
+CFG_HWRNG_PTA ?= y
+$(call force,CFG_HWRNG_QUALITY,1024)
+endif
--- a/core/drivers/atmel_trng.c
+++ b/core/drivers/atmel_trng.c
@@ -1,0 +1,156 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2021 Microchip
+ */
+
+#include <assert.h>
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <io.h>
+#include <kernel/boot.h>
+#include <kernel/spinlock.h>
+#include <libfdt.h>
+#include <matrix.h>
+#include <platform_config.h>
+#include <rng_support.h>
+#include <string.h>
+#include <tee/tee_cryp_utl.h>
+
+/* Registers */
+#define TRNG_CTRL		0x0
+#define TRNG_CTRL_WAKEY_OFFSET	8
+#define TRNG_CTRL_WAKEY_VALUE	0x524E47
+
+#define TRNG_IER		0x10
+#define TRNG_ISR		0x1C
+#define TRNG_ODATA		0x50
+
+static unsigned int trng_lock = SPINLOCK_UNLOCK;
+static unsigned int trng_random_val_lock = SPINLOCK_UNLOCK;
+static vaddr_t trng_base;
+static uint8_t random_byte_pos;
+static union {
+	uint32_t val;
+	uint8_t byte[sizeof(uint32_t)];
+} random_data;
+
+static uint32_t atmel_trng_read32(void)
+{
+	uint32_t exceptions = 0;
+	uint32_t value = 0;
+
+	exceptions = cpu_spin_lock_xsave(&trng_lock);
+
+	while (!io_read32(trng_base + TRNG_ISR))
+		;
+
+	value = io_read32(trng_base + TRNG_ODATA);
+
+	cpu_spin_unlock_xrestore(&trng_lock, exceptions);
+
+	return value;
+}
+
+TEE_Result crypto_rng_read(void *buf, size_t len)
+{
+	uint8_t *rngbuf = buf;
+	uint32_t val = 0;
+	size_t len_to_copy = 0;
+
+	assert(buf);
+	assert(trng_base);
+
+	while (len) {
+		val = atmel_trng_read32();
+		len_to_copy = MIN(len, sizeof(uint32_t));
+		memcpy(rngbuf, &val, len_to_copy);
+		rngbuf += len_to_copy;
+		len -= len_to_copy;
+	}
+
+	return TEE_SUCCESS;
+}
+
+uint8_t hw_get_random_byte(void)
+{
+	uint32_t exceptions = 0;
+	uint8_t data = 0;
+
+	assert(trng_base);
+
+	exceptions = cpu_spin_lock_xsave(&trng_random_val_lock);
+
+	/*
+	 * The TRNG generates a whole 32 bits word every 84 cycles. To avoid
+	 * discarding 3 bytes at each request, request 4 bytes of random data
+	 * and return only 1 at each request until there is no more bytes in the
+	 * random_data "cache".
+	 */
+	if (!random_byte_pos)
+		random_data.val = atmel_trng_read32();
+
+	data = random_data.byte[random_byte_pos++];
+	if (random_byte_pos == sizeof(uint32_t))
+		random_byte_pos = 0;
+
+	cpu_spin_unlock_xrestore(&trng_random_val_lock, exceptions);
+
+	return data;
+}
+
+/* This is a true RNG, no need for seeding */
+void plat_rng_init(void)
+{
+	assert(trng_base);
+}
+
+static void atmel_trng_reset(void)
+{
+	uint32_t ctrl_val = TRNG_CTRL_WAKEY_VALUE << TRNG_CTRL_WAKEY_OFFSET;
+
+	/* Disable TRNG */
+	io_setbits32(trng_base + TRNG_CTRL, ctrl_val);
+	/* Enable interrupt */
+	io_setbits32(trng_base + TRNG_IER, 1);
+	/* Enable TRNG */
+	io_setbits32(trng_base + TRNG_CTRL, ctrl_val | 1);
+}
+
+static TEE_Result trng_node_probe(const void *fdt, int node,
+				  const void *compat_data __unused)
+{
+	int status = _fdt_get_status(fdt, node);
+	size_t size = 0;
+	struct clk *clk = NULL;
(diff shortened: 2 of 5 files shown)


jenswi-linaro (commented):
There's a spell error in the "plat-sam: include sama5d2.h file necessary for some values" commit message, "probnlems".

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,155 @@
...
+	uint32_t val = 0;
+	uint8_t len_to_copy = 0;
+
+	assert(buf);
+
+	assert(trng_base);
+
+	while (len) {
+		val = atmel_trng_read32();
+		len_to_copy = len > TRNG_BYTE_READ ? TRNG_BYTE_READ : len;
```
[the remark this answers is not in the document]
clementleger: Agreed, this does not add any useful information.

clementleger:
Update version:
- Used `while()` instead of `do { } while()` for ISR
- Used MIN
- Replace TRNG_BYTES_READ with `sizeof(uint32_t)`.
- Fix typo in commit

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,150 @@
...
+{
+	uint8_t *rngbuf = buf;
+	uint32_t val = 0;
+	uint8_t len_to_copy = 0;
+
+	assert(buf);
+	assert(trng_base);
+
+	while (len) {
+		val = atmel_trng_read32();
```
jenswi-linaro, line 49: Do we need to protect this function from concurrent calls?

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,150 @@
...
+	 * and return only 1 at each request until there is no more bytes in the
+	 * trng_valid_bytes "cache".
+	 */
+	if (trng_valid_bytes == 0) {
+		trng_last_val = atmel_trng_read32();
+		trng_valid_bytes = sizeof(uint32_t);
+	}
+
+	data = trng_last_val & 0xFF;
+	trng_last_val >>= 8;
```
jenswi-linaro, line 77: It looks like this global variable needs to be protected from concurrent access.

core/drivers/atmel_trng.c
clementleger: ISR is a cleared on read register so even if two thread enters this function, only one of them will read 0 and then read ODATA. The second one will read ISR after that and will also get a data from odata. ISR will only be set after reading ODATA. So there is need to protect this function.

core/drivers/atmel_trng.c
clementleger: Most probably yes, `trng_last_val` & `trng_valid_bytes` needs to be protected.

clementleger:
New version:
 - After looking at Hikey driver, switched to union for 32 bits reading
 - Switched to @etienne-lms driver probing mecanism

core/drivers/atmel_trng.c
jenswi-linaro: I guess you meant to say that there isn't a need to protect this function.
What if the CPU breaking out of the loop is interrupted before it's able to read ODATA and then resumes at the same time as the other CPU is breaking out of the loop?
It seems more safe to rely on synchronization mechanisms designed for the purpose.

clementleger: Sorry, I meant that there were no need to protect this indeed.

> What if the CPU breaking out of the loop is interrupted before it's able to read ODATA and then resumes at the same time as the other CPU is breaking out of the loop?

If this happens,  the other CPU will not drop out of the loop since ISR won't be set until first CPU reads ODATA. As I said, ISR is set only **after** reading ODATA and ISR is cleared on read.

I can add a spinlock if needed but according to the hardware specification, there is normally no need for it.

jenswi-linaro: Is the hardware specification publicly available?

clementleger: Yes: https://ww1.microchip.com/downloads/en/DeviceDoc/DS60001476B.pdf
Section 63.

clementleger: Hum actually, after reading it again, you might be right. It seems like the ISR will be trigged again even if not reading ODATA.

jenswi-linaro: > 

Thanks

clementleger:
New version:
- Add locking around TRNG reading after @jenswi-linaro comment

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,155 @@
...
+static uint8_t random_byte_pos;
+static union {
+	uint32_t val;
+	uint8_t byte[sizeof(uint32_t)];
+} random_data;
+
+static uint32_t atmel_trng_read32(void)
+{
+	uint32_t exceptions = 0;
+	uint32_t value;
```
jenswi-linaro, line 40: Please initialize value.

jenswi-linaro (commented):
With the comment fixed:

core/drivers/atmel_trng.c
clementleger: Fixed.

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,155 @@
...
+
+static uint32_t atmel_trng_read32(void)
+{
+	uint32_t exceptions = 0;
+	uint32_t value = 0;
+
+	exceptions = cpu_spin_lock_xsave(&trng_lock);
+
+	while (!io_read32(trng_base + TRNG_ISR))
+		;
```
etienne-lms, line 45: timeout?  (hmm, IIRC you said the platform as no available timer support for optee. If so, discard my comment)

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,155 @@
...
+		random_byte_pos = 0;
+
+	cpu_spin_unlock_xrestore(&trng_random_val_lock, exceptions);
+
+	return data;
+}
+
+/* This is a true RNG, no need for seeding */
+void plat_rng_init(void)
+{
```
etienne-lms, line 103: maybe `assert(trng_base);` or like as a debug helper to check the driver has already been initialized.

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,155 @@
...
+	clk_enable(clk);
+
+	atmel_trng_reset();
+
+	return TEE_SUCCESS;
+}
+
+static const struct dt_device_match atmel_trng_match_table[] = {
+	{ .compatible = "atmel,at91sam9g45-trng" },
+	{ 0 }
```
etienne-lms, line 147: remove `0`

core/arch/arm/plat-sam/conf.mk
```diff
@@ -42,3 +42,8 @@ CFG_TEE_RAM_VA_SIZE ?= 0x100000
 # Device tree related configuration
 CFG_DT_ADDR ?= 0x21500000
 CFG_GENERATE_DTB_OVERLAY ?= y
+
+CFG_WITH_SOFTWARE_PRNG ?= n
+CFG_ATMEL_TRNG ?= y
+CFG_HWRNG_PTA ?= y
+CFG_HWRNG_QUALITY ?= 1024
```
etienne-lms, line 49: `?=` ?
Atmel TRNG quality is not something configurable. I would rather use:
```
CFG_WITH_SOFTWARE_PRNG ?= n
CFG_ATMEL_TRNG ?= y
ifeq ($(CFG_ATMEL_TRNG),y)
CFG_HWRNG_PTA ?= y
$(call force,CFG_HWRNG_QUALITY,1024)
endif

File: core/drivers/atmel_trng.c (hunk 5 above)
Comment (clementleger) [reply core/drivers/atmel_trng.c:45]: Indeed, the platform does not have the ARM generic timer extension.

File: core/drivers/atmel_trng.c (hunk 6 above)
Comment (clementleger) [reply core/drivers/atmel_trng.c:103]: `plat_rng_init()` is called in `tee_cryp_init()` which is a service initcall and thus called before drivers initcall. So adding an assert here will always assert.

File: core/drivers/atmel_trng.c (hunk 7 above)
Comment (clementleger) [reply core/drivers/atmel_trng.c:147]: Ok

File: core/arch/arm/plat-sam/conf.mk (hunk 8 above)
Comment (clementleger) [reply core/arch/arm/plat-sam/conf.mk:49]: Indeed, looks better to force it.

File: core/drivers/atmel_trng.c [hunk 9]
```diff
@@ -0,0 +1,156 @@
...
+	cpu_spin_unlock_xrestore(&trng_lock, exceptions);
+
+	return value;
+}
+
+TEE_Result crypto_rng_read(void *buf, size_t len)
+{
+	uint8_t *rngbuf = buf;
+	uint32_t val = 0;
+	uint8_t len_to_copy = 0;
```
etienne-lms, line 58: (late minor comment) no gain in specifically using `uint8_t` here, `size_t` would do the job.

core/drivers/atmel_trng.c
etienne-lms: Ok. Note you could consider removing `assert(trng_base);` from `crypto_rng_read()`. However, I think it can be handy to have such check in case you change the driver init sequence, e.g. using dt_driver probing loop.

That said, my comment related to debug support. Feel free to keep as-is.

etienne-lms (commented):
Minor comment in commit log for:
* commit "plat-sam: include sama5d2.h file necessary for some values": in commit log:
s/platform_config/platform_config.h/
* commit "plat-sam: enable atmel trng driver"
s/atmel/Atmel/
Could rephrase to "Default embed Atmel TRNG in sama5d2 platform."

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with source file minor comments addressed or not.

clementleger:
New version:
 - Switch `uint8_t` to `size_t`
 - Reword commits

core/drivers/atmel_trng.c
```diff
@@ -0,0 +1,156 @@
...
+	struct clk *clk = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (status != DT_STATUS_OK_SEC)
+		return TEE_ERROR_GENERIC;
+
+	matrix_configure_periph_secure(AT91C_ID_TRNG);
+
+	clk = clk_dt_get_by_idx(fdt, node, 0, &res);
+	if (res)
```
etienne-lms, line 133: Prefer check `clk` rather than `res` since you expect a non-NULL `clk`.

etienne-lms: Looking back at `clk_dt_get_by_idx()`  inline description, `res` being `TEE_SUCCESS` is indeed a success case and you can expect the return reference is valid and non-NULL. So fine as is. Can discard my comment.

clementleger:
Any other comments ?

etienne-lms (approved):

PR #5007, opened 2021-11-29
build: use CFLAGS32/CFLAGS64 when building in-tree TAs

jforissier:
Commit 19fdfcf617e3 ("build: ldelf and TAs can rely on CFLAGS32/CFLAGS64")
only partially implements what is mentioned in the commit description.
The ldelf part is OK, but in-tree TAs still don't use CFLAGS32 or
CFLAGS64. The reason is that the submodule name $(sm) is not ta_arm32 or
ta_arm64 like for TAs build with the "dev kit". Instead, $(sm) is the
name of the directory (such as "avb" for core/ta/avb, "pkcs11" for
core/ta/pkcs11, etc.). Therefore, it is not arch-bits-ta_arm32 or
arch-bits-ta_arm64 that needs to be set but arch-bits-$(sm).

Fixes: 19fdfcf617e3 ("build: ldelf and TAs can rely on CFLAGS32/CFLAGS64")

### DIFF
--- a/ta/mk/build-user-ta.mk
+++ b/ta/mk/build-user-ta.mk
@@ -9,8 +9,12 @@ ta-target := $(strip $(if $(CFG_USER_TA_TARGET_$(sm)), \
 		$(filter $(CFG_USER_TA_TARGET_$(sm)), $(ta-targets)), \
 		$(default-user-ta-target)))
 
-arch-bits-ta_arm32 := 32
-arch-bits-ta_arm64 := 64
+ifeq ($(ta-target),ta_arm32)
+arch-bits-$(sm) := 32
+endif
+ifeq ($(ta-target),ta_arm64)
+arch-bits-$(sm) := 64
+endif
 
 ta-dev-kit-dir$(sm) := $(out-dir)/export-$(ta-target)
 link-out-dir$(sm) := $(out-dir)/$(patsubst %/,%, $(dir $(ta-mk-file)))


jforissier:
@rfs613 would you mind checking if this fixes the issue you reported? If it does, feel free to give a `Tested-by:` and also a `Reported-by:` if you want (no obligation).

jforissier:
I am restarting IBART which reported an error that doesn't seem related to this change (random timeout?)
```
  1375:  o regression_2004.4.2 UDP Socket recv
  1376:  regression_2000.c:697: socket_recv(session, sh, buf, &blen, 0xFFFFFFFF, &ret_orig) has an unexpected value: 0xf1007003 = TEE_ISOCKET_ERROR_TIMEOUT, expected 0x0 = TEEC_SUCCESS
  1377:  regression_2000.c:779: test_2004_send_recv(c, &server_iostate, &local_iostate, &session, &sh, 64, 32) has an unexpected value: 0x0 = false, expected 0x1 = true
  1378:    regression_2004.4.2 FAILED
```

rfs613:
> @rfs613 would you mind checking if this fixes the issue you reported? If it does, feel free to give a `Tested-by:` and also a `Reported-by:` if you want (no obligation).

Indeed that does the trick, thanks!

    Tested-by: Ralph Siemsen <ralph.siemsen@linaro.org>

jenswi-linaro:

PR #5011, opened 2021-11-30
plat-stm32mp1: fixes for CRYP1 support

etienne-lms:
Don't enable CRYP1 support in OP-TEE core from ST boards supporting stm32mp1 platform to comply with the upstream boards DTS files since v5.11 in Linux kernel that enables CRYP1 in non-secure world.

Adds registering of CRYP1 device as a secure peripheral when it is registered as a crypto driver.

### DIFF
--- a/core/arch/arm/dts/stm32mp157c-dk2.dts
+++ b/core/arch/arm/dts/stm32mp157c-dk2.dts
@@ -26,10 +26,6 @@
 	};
 };
 
-&cryp1 {
-	status = "okay";
-};
-
 &dsi {
 	#address-cells = <1>;
 	#size-cells = <0>;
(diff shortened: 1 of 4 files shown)


etienne-lms:
@toromanoSTM, could you have a look at these changes? There are needed to comply with upstream Linux kernel. Assigning CRYP1 to secure world would need to update Linux in-tree DTS for these platforms which is not desired in the upstream Linux kernel source tree.

etienne-lms:
IBART failure relates to https://github.com/OP-TEE/optee_test/issues/569.

toromanoSTM:
Seems nice for me

For the serie.
`Reviewed-by: Nicolas Toromanoff <nicolas.toromanoff@foss.st.com>`

etienne-lms:
R-b tag applied.

PR #5013, opened 2021-11-30
[RFC] core: Use the embedded DT with dynamic shared memory

linnj:
Dynamic shared memory only uses the external DT at this time.
With this change, the embedded DT or the external DT can be
used for non-secure memory.

Signed-off-by: John Linn <linnj@xilinx.com>

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1133,7 +1133,7 @@ static void discover_nsec_memory(void)
 	const struct core_mmu_phys_mem *mem_begin = NULL;
 	const struct core_mmu_phys_mem *mem_end = NULL;
 	size_t nelems;
-	void *fdt = get_external_dt();
+	void *fdt = get_dt();
 
 	if (fdt) {
 		mem = get_nsec_memory(fdt, &nelems);


core/arch/arm/kernel/boot.c
```diff
@@ -1133,7 +1133,7 @@ static void discover_nsec_memory(void)
 	const struct core_mmu_phys_mem *mem_begin = NULL;
 	const struct core_mmu_phys_mem *mem_end = NULL;
 	size_t nelems;
-	void *fdt = get_external_dt();
+	void *fdt = get_dt();
```
etienne-lms, line 1136: This change would break platforms that embed a DTB in core but that expects REE memory areas are described from REE DTB (currently the external DT), not OP-TEE DTB.
I would have rather see a scheme like getting the secure DTB, find memory nodes and if none found fallback to external DTB to look for memory nodes.

etienne-lms (commented):
Thanks @linnj for the proposal.
DT nodes `memory@...` are expected to define system memory. It should be clarified in the scope of OP-TEE core if such nodes relate to overall system memory (REE+TEE+other) or to OP-TEE core memory only. Should one use a reserved-memory node in the DT to carve out some non-OP-TEE memory areas? Should the secure DTB distinguish secure memory areas from non-secure memory areas (REE dynamic shm ranges)?

core/arch/arm/kernel/boot.c
linnj: Thanks @etienne-lms, I'm still trying to get all the DT stuff into my head clearly as the docs are a big vague or I'm overlooking the details.  From the code I think we are saying that non-secure memory, right now, must always come from the external device tree and there is no option for it to come from the embedded DT and maybe that makes sense.  This implies that you would typically use both CFG_EMBED_DT and an external DT (CFG_DT_ADDR) at the same time, true?

linnj: I also see some drivers pulling driver information from only the embedded DT and failing if it's not there.

etienne-lms: > From the code I think we are saying that non-secure memory, right now, must always come from the external device tree and there is no option for it to come from the embedded DT and maybe that makes sense.

Sorry I was not clear on that. IMO location of the non-secure memory normal world could use as TEE/REE shared memory could be retrieved from either the embedded/secure DT or the external/non-secure DT.
I think the implementation should allow OP-TEE to find this information from the secure DT and if not found then fallback to fetch it from the non-secure DT.
My question is how should the secure DT distinguish whether memory nodes define secure memory address ranges or overall system memory address ranges (whatever it is secure or not).

> This implies that you would typically use both CFG_EMBED_DT and an external DT (CFG_DT_ADDR) at the same time, true?

Yes, platforms could rely on both the embedded/secure DT and an external/non-secure DT. Secure config information are read from the secure DT. Non-secure information (mainly debug stuf) could be read from the non-secure DT.

This is my view, feel free to argue, there are rooms for changes I think.

etienne-lms:
@linnj, some thoughts.

Current state: Platform accesses ext/non-secure DT to get dynshm areas
* Core gets overall main RAM (let's say DDR), carves out secure areas (tee, ta, sdp, rsvshm) resulting in a dynshm area list.
dynshm is phys memory address ranges that are legitimate TEE shm candidate pages. (phys mem addr as per core view).
* Core also adds optee firmware node and optee reserved memroy nodes in the non-sec DTB.
* That is core don't care about any reserved-memory nodes existing in the non-sec DTB.
* That is embedded DTB does not partiticipate to dynshm list

A proposal

**-** Platform embeds a (secure) DTB:
* Prior reading non-sec DTB, core can parse secure DTB to get memory nodes and the reserved-memory node.
* If secure DTB defines memory nodes : use them as generic memory, cadidate to dynshm.
* If secure DTB defines reserved-memory subnodes; shall be carved out from system memory areas to defines the dynshm areas list.

**-** Platform also access a non-secure DTB (read/changed before non-secure boots).
* Core possibly already has a list a dynshm areas. It is reliable. No other areas can be added, but some areas can shorten/changed.
* Core reads non-sec DTB and refines dynshm based on found memory nodes (intersect of the areas).
* Core reads non-sec DTB and refines dynshm based on found reserved-memory subnodes (carved out).
* Core updates non-secure DTB to add reserved a node for all areas not covered by dynshm list, based on system memory defined by  memory nodes in the the non-secure DTB.

After the 2 DTB passes, core gets the final dynshm lst allowing non-secure to safely register shm buffers.

@jforissier , @jenswi-linaro , @ruchi393 , makes sense? Feedback is welcome.

PR #5014, opened 2021-12-01
libtomcrypt/sha256_accel: fix stringop-overflow error

martinetd:
Attempting to build optee-os with gcc11 fails with the following error
```
In function ‘sha256_compress_nblocks’,
    inlined from ‘sha256_compress’ at core/lib/libtomcrypt/sha256_accel.c:81:11,
    inlined from ‘sha256_done’ at core/lib/libtomcrypt/sha256_accel.c:158:5:
core/lib/libtomcrypt/sha256_accel.c:75:5: error: ‘crypto_accel_sha256_compress’ accessing 32 bytes in a region of size 20 [-Werror=stringop-overflow=]
   75 |     crypto_accel_sha256_compress(state, buf, blocks);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
core/lib/libtomcrypt/sha256_accel.c: In function ‘sha256_done’:
core/lib/libtomcrypt/sha256_accel.c:75:5: note: referencing argument 1 of type ‘uint32_t *’ {aka ‘unsigned int *’}
In file included from core/lib/libtomcrypt/sha256_accel.c:41:
core/include/crypto/crypto_accel.h:45:6: note: in a call to function ‘crypto_accel_sha256_compress’
   45 | void crypto_accel_sha256_compress(uint32_t state[8], const void *src,
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Indeed, the 'state' argument here is taken from sha1.state which is a
uint32_t state[5], so 20 bytes long instead of the uint32_t state[8]
crypto_accel_sha256_compress expects.

OTOH we're in a sha256 function, and sha256.state conveniently is of the
correct size, so use sha256.state as appropriate instead.

Note that hash_state is a union and sha{1,256}.state are at the same
offset, so this is actually a no-op change.

Signed-off-by: Dominique Martinet <dominique.martinet@atmark-techno.com>

### DIFF
--- a/core/lib/libtomcrypt/sha256_accel.c
+++ b/core/lib/libtomcrypt/sha256_accel.c
@@ -68,9 +68,9 @@ int sha256_ce_transform(ulong32 *state, const unsigned char *buf, int blocks);
 static int sha256_compress_nblocks(hash_state *md, const unsigned char *buf,
 				   int blocks)
 {
-   void *state = md->sha1.state;
+   void *state = md->sha256.state;
 
-   COMPILE_TIME_ASSERT(sizeof(md->sha1.state[0]) == sizeof(uint32_t));
+   COMPILE_TIME_ASSERT(sizeof(md->sha256.state[0]) == sizeof(uint32_t));
 
     crypto_accel_sha256_compress(state, buf, blocks);
     return CRYPT_OK;


jenswi-linaro:

jforissier:
Thanks for the patch. Let's ignore the checkpatch error, it looks like it is upset by the format of the error messages emitted by GCC 11.

martinetd:
Thanks for the feedback!
Yes it looks like the gcc warning lines are too long. I'd rather keep these as is so I fully agree with ignoring checkpatch :)

This is my first contribution here so I wasn't too sure, but looking at older PRs it looks like I'm supposed to amend the commit message with acks received here?
I've just done so and force pushed.

PR #5015, opened 2021-12-01
crypto: drivers: se050: object identifier range [FIX]

ldts:
With the introduction of the se050 APDU driver, external clients can
create persistent objects on the secure element non-volatile memory.

The unique identifiers for these objects do not necessarily need to
fall within the range defined for objects created using the
cyptographic operation interfaces (keypair_gen).

This commit fixes the use case where a key stored in the SE05x device
(for example via a cloud service communicating to the optee-client's
libseteec) is imported into the secure world (ie the pkcs#11 database) and then used for
authentication (ie, EC sign)

### DIFF
--- a/core/drivers/crypto/se050/adaptors/utils/utils.c
+++ b/core/drivers/crypto/se050/adaptors/utils/utils.c
@@ -115,7 +115,8 @@ static uint32_t se050_key(uint64_t key)
 	if (!IS_WATERMARKED(key))
 		return 0;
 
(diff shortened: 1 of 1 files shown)


ldts:
hi @jforissier, just a quick reminder in case this went unnoticed. it is safe to merge it.

jforissier:
OK,

ldts:
thanks. just updated with the tag

PR #5025, opened 2021-12-03
Remote attestation PTA

jforissier:
Differences with [v4](https://github.com/OP-TEE/optee_os/pull/5010):
- The hash of TA memory pages from [v2](https://github.com/OP-TEE/optee_os/pull/4972) is back, in addition to obtaining the digest from the TA secure header (struct shdr::hash) is still possible. This way more use cases can be supported.
- Attestation of TEE core memory is added, but see the limitation with ASLR in the commit description (might or might not be an big issue depending on the use case).

~~Feel free to formally review the commits that are not tagged "[RFC]", although they are needed only for the attestation PTA, they also make sense on their own (except perhaps the first one "scripts: add dump_ta_header.py" and last one "kernel/linker.h: export __text_end").~~

Update: "RFC v5" tag removed.


core/tee/tee_fs_rpc.c
```diff
@@ -23,6 +24,27 @@ struct tee_fs_dir {
 	struct tee_fs_dirent d;
 };
 
+/* "/dirf.db" or "/<file number>" */
+static TEE_Result
+tee_svc_storage_create_filename_dfh(void *buf, size_t blen,
```
jenswi-linaro, line 29: Would it make sense to rename this function to `create_filename()` also?

core/tee/tee_svc_storage.c
```diff
@@ -104,29 +103,6 @@ TEE_Result tee_svc_storage_create_filename(void *buf, size_t blen,
...
-	if (pos >= blen)
-		return TEE_ERROR_SHORT_BUFFER;
-
-	l = blen - pos;
-	return tee_fs_dirfile_fileh_to_fname(dfh, file + pos, &l);
-}
-#endif
-
 /* "/TA_uuid" */
 TEE_Result tee_svc_storage_create_dirname(void *buf, size_t blen,
```
jenswi-linaro, line 107: How about moving this function and `tee_svc_storage_create_filename()` to `tee_rpmb_fs.c` following the same logic?

jenswi-linaro (commented):
Comments for "core: move tee_svc_storage_create_filename_dfh() to core/tee/tee_fs_rpc.c"

core/tee/sub.mk
```diff
@@ -24,25 +24,29 @@ srcs-$(CFG_CRYPTO_CONCAT_KDF) += tee_cryp_concat_kdf.c
...
 srcs-y += tee_time_generic.c
 srcs-$(CFG_SECSTOR_TA) += tadb.c
 srcs-$(CFG_GP_SOCKETS) += socket.c
 srcs-y += tee_ta_enc_manager.c
-
 endif #CFG_WITH_USER_TA,y
 
+ifeq ($(_CFG_WITH_SECURE_STORAGE),y)
+srcs-y += tee_fs_key_manager.c
+srcs-$(CFG_RPMB_FS) += tee_rpmb_fs.c
```
jenswi-linaro, line 40: Given that `_CFG_WITH_SECURE_STORAGE` will be `y` if `CFG_RPMB_FS` is `y` it seems unnecessary to have this inside the if.

jenswi-linaro (commented):
Comment for "core: move FS files under _CFG_WITH_SECURE_STORAGE"

core/tee/tee_fs_rpc.c
jforissier: Absolutely.

core/tee/tee_svc_storage.c
jforissier: Yes, will do.

core/tee/sub.mk
jforissier: Correct, I will simplify.

jforissier:
Comments addressed.

jenswi-linaro:
For the following commits with their fixups please apply:

jforissier:
Fixup commits squashed and review tag applied. I have changed the subject of "core: move tee_svc_storage_create_filename() ..." and improved the description too, it is now "core: move functions from tee_svc_storage.c to tee_rpmb_fs.c".

jenswi-linaro:
I think it would make sense to merge these reviewed commits separately. I should also make it easier to focus on what this PR is about.

jforissier:
> I think it would make sense to merge these reviewed commits separately. I should also make it easier to focus on what this PR is about.

Sure. I have merged them just now and rebased the branch. Thanks for reviewing.

core/arch/arm/kernel/kern.ld.S
```diff
@@ -93,6 +93,9 @@ SECTIONS
 	.text : {
 		KEEP(*(.text._start))
 		__identity_map_init_start = .;
+		__text_data_start = .;
+		*(.text.data)
```
jenswi-linaro, line 97: If there's ever a function, static or global, named data it will go into this too.
How about using `.identify_map.data` instead? That way we'd have full control of what's going in here.

jforissier: Fair enough, and it also makes it obvious that the data need to be reachable using the identity mapping which is good.
Can I still keep `__text_data_start`/`__text_data_end`? To the "outside world" (tools etc.) it's essentially data embedded in `.text`.

jenswi-linaro: Sure

jenswi-linaro:
Please squash the fixups.

jforissier:
> Please squash the fixups.

Will do. Some more work is needed to deal with the ASLR case, because there is still relocatable data in both the `.text` and `.rodata` sections. I'll refrain from pushing to this branch until I have something that works and is in a good shape to be reviewed. Thanks for the comments so far.

jforissier:
Update. The attestation commit is mostly unchanged but I have added a number of commits to suppress text relocations and a few more to avoid storing relocatable data into `.rodata`. @jenswi-linaro despite your advice to use the stack, I couldn't figure out how to deal with the transition from reduced to full mapping though :-/ please see commit "FIXME: core: arm: disable CFG_UNMAP_CORE_AT_EL0 by default". I wouldn't mind some more help on that one...

jenswi-linaro:
Perhaps something like (untested) this would work. This stuff is a bit hairy so let me know if you'd rather I did this.

``` diff
diff --git a/core/arch/arm/include/kernel/thread.h b/core/arch/arm/include/kernel/thread.h
index d2b0d006c8de..ce8069fbec25 100644
--- a/core/arch/arm/include/kernel/thread.h
+++ b/core/arch/arm/include/kernel/thread.h
@@ -43,6 +43,7 @@ struct thread_core_local {
 	uint64_t x[4];
 #endif
 	vaddr_t tmp_stack_va_end;
+	long kcode_offset;
 	short int curr_thread;
 	uint32_t flags;
 	vaddr_t abt_stack_va_end;
diff --git a/core/arch/arm/kernel/asm-defines.c b/core/arch/arm/kernel/asm-defines.c
index 76c87de881a0..c1436baa937c 100644
--- a/core/arch/arm/kernel/asm-defines.c
+++ b/core/arch/arm/kernel/asm-defines.c
@@ -94,6 +94,8 @@ DEFINES
 	/* struct thread_core_local */
 	DEFINE(THREAD_CORE_LOCAL_X0, offsetof(struct thread_core_local, x[0]));
 	DEFINE(THREAD_CORE_LOCAL_X2, offsetof(struct thread_core_local, x[2]));
+	DEFINE(THREAD_CORE_LOCAL_KCODE_OFFSET,
+	       offsetof(struct thread_core_local, kcode_offset));
 #endif /*ARM64*/
 
 	/* struct thread_ctx */
diff --git a/core/arch/arm/kernel/thread.c b/core/arch/arm/kernel/thread.c
index 05dd8697e856..e89d85092c48 100644
--- a/core/arch/arm/kernel/thread.c
+++ b/core/arch/arm/kernel/thread.c
@@ -1045,6 +1045,15 @@ static void init_thread_stacks(void)
 }
 #endif /*CFG_WITH_PAGER*/
 
+static void set_core_local_kcode_offset(struct thread_core_local *cls,
+					long offset)
+{
+	size_t n = 0;
+
+	for (n = 0; n < CFG_TEE_CORE_NB_CORE; n++)
+		cls[n].kcode_offset = offset;
+}
+
 static void init_user_kcode(void)
 {
 #ifdef CFG_CORE_UNMAP_CORE_AT_EL0
@@ -1058,7 +1067,11 @@ static void init_user_kcode(void)
 	core_mmu_get_user_va_range(&v, NULL);
 	thread_user_kcode_offset = thread_user_kcode_va - v;
 
+	set_core_local_kcode_offset(thread_core_local,
+				    thread_user_kcode_offset);
 #if defined(CFG_CORE_WORKAROUND_SPECTRE_BP_SEC) && defined(ARM64)
+	set_core_local_kcode_offset((void *)thread_user_kdata_page,
+				    thread_user_kcode_offset);
 	/*
 	 * When transitioning to EL0 subtract SP with this much to point to
 	 * this special kdata page instead. SP is restored by add this much
diff --git a/core/arch/arm/kernel/thread_a64.S b/core/arch/arm/kernel/thread_a64.S
index 5d70d9ed24fe..9d13f81ae327 100644
--- a/core/arch/arm/kernel/thread_a64.S
+++ b/core/arch/arm/kernel/thread_a64.S
@@ -188,7 +188,13 @@ END_FUNC thread_unwind_user_mode
 		isb
 
 		/* Jump into the full mapping and continue execution */
+#if 0
 		ldr	x0, =1f
+#else
+		adr	x0, 1f
+		ldr	x1, [sp, #THREAD_CORE_LOCAL_KCODE_OFFSET]
+		add	x0, x0, x1
+#endif
 		br	x0
 	1:
 BTI(		bti	j)
@@ -484,7 +490,11 @@ eret_to_el0:
 #endif
 
 	/* Jump into the reduced mapping and continue execution */
+#if 0
 	ldr	x1, =1f
+#else
+	adr	x1, 1f
+#endif
 	sub	x1, x1, x0
 	br	x1
 1:
```

jforissier:
@jenswi-linaro awesome, your suggestion works fine and eliminates 6 relocations in `thread_a64.S`. There is one left however :-/
In `icache_inv_user_range()`, there is a `ldr x0, =1f`. I suspect the same trick can be applied, but how to find the location if the offset? Not based on `sp` at this point, right?

jenswi-linaro:
This (untested) code might work:
``` diff
diff --git a/core/arch/arm/kernel/thread_a64.S b/core/arch/arm/kernel/thread_a64.S
index 5d70d9ed24fe..1356fd53304f 100644
--- a/core/arch/arm/kernel/thread_a64.S
+++ b/core/arch/arm/kernel/thread_a64.S
@@ -535,16 +535,16 @@ icache_inv_user_range:
 
 #ifdef CFG_CORE_UNMAP_CORE_AT_EL0
 	/* Point to the vector into the reduced mapping */
-	adr_l	x2, thread_user_kcode_offset
-	ldr	x2, [x2]
+	adr_l	x7, thread_user_kcode_offset
+	ldr	x7, [x7]	/* this register must be preserved */
 	mrs	x4, vbar_el1	/* this register must be preserved */
-	sub	x3, x4, x2
+	sub	x3, x4, x7
 	msr	vbar_el1, x3
 	isb
 
 	/* Jump into the reduced mapping and continue execution */
-	ldr	x3, =1f
-	sub	x3, x3, x2
+	adr	x3, 1f
+	sub	x3, x3, x7
 	br	x3
 1:
 BTI(	bti	j)
@@ -588,7 +588,8 @@ BTI(	bti	j)
 	isb
 
 	/* Jump into the full mapping and continue execution */
-	ldr	x0, =1f
+	adr	x0, 1f
+	add	x0, x0, x7
 	br	x0
 1:
 BTI(	bti	j)
```

jforissier:
> This (untested) code might work:

Guess what? It does! :+1: I will introduce theses changes at the right place in the series. Thanks again!

core/arch/arm/kernel/thread_a64.S
```diff
@@ -535,16 +535,16 @@ icache_inv_user_range:
...
 	mrs	x4, vbar_el1	/* this register must be preserved */
-	sub	x3, x4, x2
+	sub	x3, x4, x7
 	msr	vbar_el1, x3
 	isb
 
 	/* Jump into the reduced mapping and continue execution */
-	ldr	x3, =1f
-	sub	x3, x3, x2
+	adr_l	x3, 1f
```
jenswi-linaro, line 548: A plain `adr` should be enough, same below. This works because it's not very far away.

core/arch/arm/plat-rcar/romapi_call.S
```diff
@@ -31,9 +32,11 @@ FUNC __plat_romapi_wrapper , : , .identity_map
 	/*
 	 * We are running at identity location, so we can't use bl there,
 	 * because assembler will generate relative address to virt_to_phys(),
-	 * which is not identity mapped. So we need to use absolute address.
+	 * which is not identity mapped.
 	 */
-	ldr	x9, =virt_to_phys
+	adr     x9, virt_to_phys
```
jenswi-linaro, line 37: This should probably be an `adr_l` instead since we don't know how far away it might be, same below.

jenswi-linaro (commented):
Comment for "plat-rcar: suppress text relocations in romapi_call.S"

core/arch/arm/kernel/kern.ld.S
```diff
@@ -234,6 +232,7 @@ SECTIONS
 		__data_start = .;
 		*(.data .data.* .gnu.linkonce.d.*)
 		. = ALIGN(8);
+		KEEP(*(SORT(.scattered_array*)));
```
jenswi-linaro, line 230: This will prevent the scattered arrays from being paged any longer. Another drawback is that they will become less protected at runtime since they are now mapped read-write.

jenswi-linaro (commented):
Comments for "core: kern.ld.S: move .scattered_array* into .data"

core/arch/arm/kernel/kern.ld.S
```diff
@@ -233,6 +228,12 @@ SECTIONS
 		*(.data .data.* .gnu.linkonce.d.*)
 		. = ALIGN(8);
 		KEEP(*(SORT(.scattered_array*)));
+#ifdef CFG_DT
+		. = ALIGN(8);
```
jenswi-linaro, line 232: Same problem here as with scattered arrays. While looking at this I wonder why the dtdrv stuff isn't implemented using scattered arrays.

jenswi-linaro (commented):
Comment for "core: kern.ld.S: rename .rodata.dtdrv to .data.dtdrv, move to .data"

jenswi-linaro:
With this we actually removes some protection by making data that where supposed to remain constant after relocation read-write mapped instead. Perhaps we instead should collect all this data in a special section that can be mapped read-only and still be paged.

core/arch/arm/plat-rcar/romapi_call.S
jforissier: Yes, will fix.

core/arch/arm/kernel/thread_a64.S
jforissier: Correct, fixed.

jforissier:
> With this we actually removes some protection by making data that where supposed to remain constant after relocation read-write mapped instead. Perhaps we instead should collect all this data in a special section that can be mapped read-only and still be paged.

That's a valid concern indeed. I will think about it.

jforissier:
Rebased onto master which gets rid of the dtdrv stuff. New commits added to address comments are 8e403217 "[Review] core: kern.ld.S: move .scattered_array* into .data" and later.

core/arch/arm/kernel/kern.ld.S
```diff
@@ -134,11 +134,21 @@ SECTIONS
...
+#ifndef CFG_CORE_ASLR
+		. = ALIGN(8);
+		KEEP(*(SORT(.scattered_array*)));
+#endif
 #endif
 		. = ALIGN(8);
 		__rodata_end = .;
 	}
 
+#if !defined(CFG_WITH_PAGER) && defined(CFG_CORE_ASLR)
```
jenswi-linaro, line 146: Why can't this always be used even when ASLR isn't enabled?

jforissier: It can, but doesn't have to. I thought having a section called `.data.rel.ro` could be slightly misleading when ASLR is disabled (and therefore the executable is not relocatable).
I can simplify if you prefer.

jenswi-linaro: Makes sense, let's keep it as it is now.

b49020:
@jforissier Thanks again for clarifying queries at OP-TEE LOC meeting. I think from overall design aspect, this feature looks in better shape now. So we should be good to remove RFC tag at this point. If anyone else has different views, please feel free to comment.

Also, can you squash all the review and update commits? As I would like to spend some time to review and test actual code afresh.

BTW, if anybody missed OP-TEE LOC meeting discussion regarding this feature, [here](https://docs.google.com/presentation/d/10vrFfxCt7Th8OL05jq-_2-uNVaJXQaH3omP97AsCVy4/edit#slide=id.p) is the reference link to the slides.

jforissier:
@b49020 yes I will squash the updates and rebase onto master. I noticed I forgot to hash `.text_init` and `.rodata_init` (present when pager is enabled), I will fix that too.

b49020:
While looking at this patch-set, following commit caught mine attention.

`Commit: "core: do not tag ops structures with __rodata_unpaged"`

It looks like after this commit with `CFG_CORE_ASLR=y`, we are essentially allowing constant operations structs writable. IOW, no MMU RO protection which was there earlier with `__rodata_unpaged`. This will make it easier for any vulnerability found in TEE to modify function pointers and redirect code flow. Although having measurement for these ops structs included would be the best solution but this is where ASLR makes it difficult (two security features limiting each other's capability rather than attacker's :)).

However, we shouldn't remove MMU RO protection for these ops struct and atleast can have a special section for them to be excluded from measurement. Any better ideas are much appreciated.

b49020 (commented):
This review doesn't include `core/pta/attestation.c` as that's something I will look into next.

jforissier:
> While looking at this patch-set, following commit caught mine attention.
> 
> `Commit: "core: do not tag ops structures with __rodata_unpaged"`
> 
> It looks like after this commit with `CFG_CORE_ASLR=y`, we are essentially allowing constant operations structs writable. IOW, no MMU RO protection which was there earlier with `__rodata_unpaged`. This will make it easier for any vulnerability found in TEE to modify function pointers and redirect code flow. Although having measurement for these ops structs included would be the best solution but this is where ASLR makes it difficult (two security features limiting each other's capability rather than attacker's :)).
> 
> However, we shouldn't remove MMU RO protection for these ops struct and atleast can have a special section for them to be excluded from measurement. Any better ideas are much appreciated.

You're right. In fact these structures are quite similar to the scattered arrays, so we can probably store them in `.data.rel.ro*` when ASLR is enabled. I will propose something.

core/arch/arm/kernel/kern.ld.S
```diff
@@ -323,13 +331,18 @@ SECTIONS
...
-		__rodata_init_end = .;
+#endif
 	}
+#ifdef CFG_CORE_ASLR
+	.data.rel.ro_init : ALIGN(8) {
+		KEEP(*(SORT(.scattered_array*)));
+	}
+#endif
+	. = ALIGN(8);
+	__rodata_init_end = .;
```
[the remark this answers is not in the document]
jforissier: Good catch ;-) This is actually fixed by later commit "core: pager: export __{text,rodata}_{init,pageable}_{start,end}". At this point it doesn't matter since the attestation PTA doesn't exist yet.
Still I suppose it would make more sense to introduce `__ro_and_relro_data_init_end` at the same time as the new relro section. I will fix that.

mk/config.mk
```diff
@@ -574,6 +574,17 @@ $(eval $(call cfg-depends-all,CFG_SYSTEM_PTA,CFG_WITH_USER_TA))
 # world OS.
 CFG_DEVICE_ENUM_PTA ?= y
 
+# The attestation pseudo TA provides an interface to request a measurement of
+# a given TA binary or shared library given its UUID.
+CFG_ATTESTATION_PTA ?= $(_CFG_WITH_SECURE_STORAGE)
```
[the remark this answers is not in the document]
jforissier: Maybe. I would like it to be tested by CI, but I can `export CFG_ATTESTATION_PTA=y` in `.azure-pipelines.yml` instead.

jforissier:
@b49020 I have fixed your comments in place rather than add fixup commits because it would have been hard to understand. Also rebased on master.
Changes since last review:
- Introduce __ro_and_relro_data_init_end in "core: kern.ld.S: move .scattered_array* into .data.rel.ro" (note: subject previously said '.data', it was a mistake)
- Discard "core: do not tag ops structures with __rodata_unpaged"
- Instead, add "core: compiler.h: introduce __relrodata_unpaged(x)" and "core: tag ops structures with __relrodata_unpaged". Question: is it OK to still use __rodata_dummy in `link_dummies_paged.c`?
- Fix a small issue in print_tee_hash.py: do not fail when a section is not found, it does happen when pager is disabled(commit: "scripts: add print_tee_hash.py").

Tested with and without pager, with and without ASLR.

b49020:
@jforissier Thanks for the update.

For commits:
```
core: suppress text relocation in thread.c
core: arm64: suppress text relocations in thread_optee_smc_a64.S
core: arm64: suppress some text relocations in thread_a64.S
core: arm64: suppress text relocations in icache_inv_user_range()
core: arm64: suppress text relocations caused by restore_mapping macro
plat-rcar: suppress text relocations in romapi_call.S
kernel/linker.h: export __text_end
core: add delimited area in .text to store data
core: arm64: move boot_mmu_config and cached_mem_end to .identity_map.data
core: kern.ld.S: move .scattered_array* into .data.rel.ro
core: compiler.h: introduce __relrodata_unpaged(x)
core: tag ops structures with __relrodata_unpaged
core: arm64: aslr: pass '-z text' linker option
core: pager: export __{text,rodata}_{init,pageable}_{start,end}
```
feel free to add:

> Question: is it OK to still use __rodata_dummy in link_dummies_paged.c?

I don't see a point to have declarations separately in this file. So shouldn't those be dropped entirely?

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+#ifndef __PTA_ATTESTATION_H
+#define __PTA_ATTESTATION_H
+
+#define PTA_ATTESTATION_UUID { 0x39800861, 0x182a, 0x4720, \
+		{ 0x9b, 0x67, 0x2b, 0xcd, 0x62, 0x2b, 0xc0, 0xb5 } }
+
+/*
+ * Get the RSA public key use for signing the values returned by other commands.
```
b49020, line 17: s/signing/verification/

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,99 @@
...
+ * [out]    memref[0]        Public key exponent in big endian order
+ * [out]    memref[1]        Modulus in big endian order
+ *
+ * Return codes:
+ * TEE_SUCCESS
+ * TEE_ERROR_GENERIC - Internal error
+ * TEE_ERROR_SHORT_BUFFER - One or both buffers are too small, required size
+ *                          is provided in memref[i].size
+ */
+#define PTA_ATTESTATION_GET_PUBKEY 0x0
```
b49020, line 28: Although currently we are using RSA as the signing scheme but it should be possible in future to extend this pTA to use ECDSA scheme instead in a backwards compatible way. So I would suggest you to add an output value parameter to say that is an RSA public key.

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,99 @@
...
+ *                           bytes. For example, for a 32-byte SHA256 digest and
+ *                           2048 bit key (256 bytes) the minimum buffer size is
+ *                           288 bytes.
+ *
+ * Return codes:
+ * TEE_SUCCESS
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_SHORT_BUFFER - Output buffer size less than required
+ */
+#define PTA_ATTESTATION_GET_TA_SHDR_DIGEST 0x1
```
b49020, line 61: Here as well, we should have an output value parameter to say that its an RSA signature.

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,99 @@
...
+ *                           signature. See PTA_ATTESTATION_GET_TA_HDR_DIGEST
+ *                           for a description of the signature.
+ *
+ * Return codes:
+ * TEE_SUCCESS
+ * TEE_ERROR_ACCESS_DENIED - Caller is not a user space TA
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_SHORT_BUFFER - Output buffer size less than required
+ */
+#define PTA_ATTESTATION_HASH_TA_MEMORY 0x2
```
b49020, line 79: Here as well, we should have an output value parameter to say that its an RSA signature.

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,99 @@
...
+ *                           signature. See PTA_ATTESTATION_GET_TA_HDR_DIGEST
+ *                           for a description of the signature.
+ *
+ * Return codes:
+ * TEE_SUCCESS
+ * TEE_ERROR_ACCESS_DENIED - Caller is not a user space TA
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_SHORT_BUFFER - Output buffer size less than required
+ */
+#define PTA_ATTESTATION_HASH_TEE_MEMORY 0x3
```
b49020, line 97: Here as well, we should have an output value parameter to say that its an RSA signature.

mk/config.mk
```diff
@@ -574,6 +574,17 @@ $(eval $(call cfg-depends-all,CFG_SYSTEM_PTA,CFG_WITH_USER_TA))
 # world OS.
 CFG_DEVICE_ENUM_PTA ?= y
 
+# The attestation pseudo TA provides an interface to request a measurement of
+# a given TA binary or shared library given its UUID.
+CFG_ATTESTATION_PTA ?= n
+$(eval $(call cfg-depends-all,CFG_ATTESTATION_PTA,_CFG_WITH_SECURE_STORAGE))
+
+# RSA key size (in bits) for the attestation PTA. Must be at least 2048 which
+# is the size of the data to be signed (an SHA-256 hash).
```
b49020, line 583: > Must be at least 2048 which is the size of the data to be signed (an SHA-256 hash).

Isn't size of SHA-256 hash 256 bits? Although I agree that choosing 2048 as RSA key size is better default.

lib/libutee/include/pta_attestation.h
```diff
@@ -0,0 +1,99 @@
...
+ * read-only data).
+ *
+ * [in]     memref[0]        Nonce
+ * [out]    memref[1]        SHA256 hash of the TEE memory followed by a
+ *                           signature. See PTA_ATTESTATION_GET_TA_HDR_DIGEST
+ *                           for a description of the signature.
+ *
+ * Return codes:
+ * TEE_SUCCESS
+ * TEE_ERROR_ACCESS_DENIED - Caller is not a user space TA
```
b49020, line 93: Is this error code relevant here?

b49020 (commented):
Comments for pTA interface: `core: pta: add remote attestation PTA`

mk/config.mk
```diff
@@ -574,6 +574,17 @@ $(eval $(call cfg-depends-all,CFG_SYSTEM_PTA,CFG_WITH_USER_TA))
 # world OS.
 CFG_DEVICE_ENUM_PTA ?= y
 
+# The attestation pseudo TA provides an interface to request a measurement of
+# a given TA binary or shared library given its UUID.
```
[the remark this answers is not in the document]
jforissier: Will fix

lib/libutee/include/pta_attestation.h
jforissier: Right

lib/libutee/include/pta_attestation.h
jforissier: I agree that other encryption schemes may be useful in the future, but I'd rather avoid introducing too much complexity now. I think what you are asking can easily be done with minimal code changes by adding a new command.

mk/config.mk
jforissier: Actually what I wrote is complete BS :open_mouth: :smile: I will update the text and explain the key size constraint more precisely.

lib/libutee/include/pta_attestation.h
jforissier: Good catch, I'll remove.

jforissier:
Update

lib/libutee/include/pta_attestation.h
b49020: Yeah adding a new info command would be good alternative. I think it would be worth to add at this point. As it would avoid to hard code signature type and algorithm used in both pseudo TA and remote client when that info can be fetched at runtime.

jforissier: How about `PTA_ATTESTATION_GET_PUBKEY` would return `TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA256` as `[out] value[2].a`?

b49020 (commented):
pTA implementation comments.

lib/libutee/include/pta_attestation.h
b49020: That's fine with me as well.

core/pta/attestation.c
```diff
@@ -0,0 +1,761 @@
...
+
+static struct rsa_keypair *key;
+
+static const uint8_t key_file_name[] = "key";
+
+static TEE_Result allocate_key(void)
+{
+	assert(!key);
+
+	key = calloc(sizeof(*key), 1);
```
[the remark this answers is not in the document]
jforissier: You're right, I will fix all occurrences.

core/pta/attestation.c
```diff
@@ -0,0 +1,761 @@
...
+	return res;
+}
+
+static TEE_Result load_key(void)
+{
+	/*
+	 * Serialized key pair is 3 bignums (e, p and n) plus their sizes
+	 * encoded as uint16_t. e is 65537 so it needs only 3 bytes.
+	 */
+	size_t size = 3 + 2 * MAX_KEY_SIZE / 8 + 3 * sizeof(uint16_t);
```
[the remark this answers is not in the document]
jforissier: - I used `MAX_KEY_SIZE` rather than `CFG_ATTESTATION_PTA_KEY_SIZE` because I thought maybe we should be able to read a larger key that might have been generated and saved by a prior version of the TEE.
- I will move the formula to a macro (or even better a static function)
- I will also allocate the buffer within `init_key()` as suggested.

core/pta/attestation.c
```diff
@@ -0,0 +1,761 @@
...
+
+{
+	const struct tee_file_operations *fops = NULL;
+	struct tee_file_handle *fh = NULL;
+	TEE_Result res = TEE_SUCCESS;
+	struct tee_pobj *po = NULL;
+
+	fops = tee_svc_storage_file_ops(storage_id);
+	if (!fops)
+		return TEE_ERROR_ITEM_NOT_FOUND;
```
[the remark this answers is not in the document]
jforissier: Should not make a difference in practice, but "not implemented" does sound better.

core/pta/attestation.c
```diff
@@ -0,0 +1,761 @@
...
+	void *ctx = NULL;
+
+	res = crypto_hash_alloc_ctx(&ctx, TEE_ALG_SHA256);
+	if (res)
+		return res;
+
+	res = crypto_hash_init(ctx);
+	if (res)
+		goto out;
+	if (nonce) {
```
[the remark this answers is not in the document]
jforissier: No, but it would make the documentation more complex :wink "SIG(SHA256(Nonce | digest)) except when no Nonce is provided in which case it is SIG(digest)". Perhaps I should simply reject `nonce == NULL`? It doesn't seem safe to run the algorithm without a nonce.

core/pta/attestation.c
```diff
@@ -0,0 +1,761 @@
...
+	if (res)
+		return res;
+
+	min_out_sz = TEE_SHA256_HASH_SIZE + crypto_bignum_num_bytes(key->n);
+	params[1].memref.size = min_out_sz;
+	if (out_sz < min_out_sz)
+		return TEE_ERROR_SHORT_BUFFER;
+	out_sz = min_out_sz;
+
+	/* Check that we're called from a user TA */
```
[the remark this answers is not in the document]
jforissier: Would be better yes.

core/pta/attestation.c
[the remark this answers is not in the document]
b49020: Makes sense to me.

core/pta/attestation.c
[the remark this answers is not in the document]
b49020: I agree, we should reject to perform a signature without nonce.

b49020:
With comments (https://github.com/OP-TEE/optee_os/pull/5025#discussion_r797596335, https://github.com/OP-TEE/optee_os/pull/5025#discussion_r798518264) addressed, feel free to add:

for commits:

```
scripts: add print_tee_hash.py
core: pta: add remote attestation PTA
```

jforissier:
@b49020 thanks for the review!

b49020:
@jforissier you are welcome but still I don't this https://github.com/OP-TEE/optee_os/pull/5025#discussion_r797596335 addressed.

jforissier:
@b49020 sorry I forgot to add the change :-/ now done. I have also once again revised the explanation for the minimal key size, because I had referred to the wrong section in RFC8017...

jforissier:
Rebased to resolve conflicts.

jforissier:
If there is no more comment I will merge this. @jenswi-linaro?

core/arch/arm/kernel/entry_a64.S
```diff
@@ -29,8 +30,7 @@
 		adr_l	x1, stack_tmp_stride
 		ldr	w1, [x1]
 		mul	x1, x0, x1
-		adr_l	x0, stack_tmp_export
-		ldr	x0, [x0]
+		bl	thread_get_stack_tmp_export
```
jenswi-linaro, line 33: SP is unassigned at this point. How can you know that `thread_get_stack_tmp_export()` doesn't use the stack?

jforissier: Good point. It happens to be the case (I checked the disassembly), but there is obviously no guarantee so I need to find another way. I think I'll rewrite the function in assembly unless you have a better idea.

jenswi-linaro: In that case we could just add the needed instructions here.

jforissier:
Update:
- Commit "core: suppress text relocations in thread.c" reverted and replaced with "core: suppress text relocation on stack_tmp_export"
- Rebased onto master

core/arch/arm/kernel/entry_a32.S
```diff
@@ -823,9 +829,9 @@ LOCAL_FUNC enable_mmu , : , .identity_map
 	bx	lr
 END_FUNC enable_mmu
 
-LOCAL_DATA stack_tmp_export_rel , :
-	.word	stack_tmp_export - stack_tmp_export_rel
-END_DATA stack_tmp_export_rel
+LOCAL_DATA stack_tmp_rel , :
+	.word	stack_tmp - stack_tmp_rel
```
jenswi-linaro, line 833: I'm not sure but it seems that making this `stack_tmp - stack_tmp_rel - STACK_TMP_GUARD` should allow us to skip the entire `#if (STACK_TMP_GUARD != 0)` above.

jforissier: It does work. Updated.

jenswi-linaro: Great, please do the same in the _a64.S.

jforissier: Done.

jenswi-linaro:

jforissier:
@b49020 @jenswi-linaro thanks for your time, I am merging this now.

PR #5032, opened 2021-12-07
core: arm: lpae: fix build with large number of CPU cores

jforissier:
A compile time assertion is triggered by the following command:

 $ make -j10 -s PLATFORM=vexpress-qemu_armv8a CFG_TEE_CORE_NB_CORE=128 \
                CFG_LPAE_ADDR_SPACE_BITS=40
 In file included from core/arch/arm/mm/core_mmu_lpae.c:61:
 core/arch/arm/mm/core_mmu_lpae.c: In function ‘core_init_mmu_prtn_ta_core’:
 lib/libutils/isoc/include/assert.h:30:24: error: duplicate case value
    30 |   switch (0) { case 0: case ((x) ? 1: 0): default : break; } \
       |                        ^~~~
 core/arch/arm/mm/core_mmu_lpae.c:697:2: note: in expansion of macro ‘COMPILE_TIME_ASSERT’
   697 |  COMPILE_TIME_ASSERT(MAX_XLAT_TABLES <
       |  ^~~~~~~~~~~~~~~~~~~
 lib/libutils/isoc/include/assert.h:30:16: note: previously used here
    30 |   switch (0) { case 0: case ((x) ? 1: 0): default : break; } \
       |                ^~~~
 core/arch/arm/mm/core_mmu_lpae.c:697:2: note: in expansion of macro ‘COMPILE_TIME_ASSERT’
   697 |  COMPILE_TIME_ASSERT(MAX_XLAT_TABLES <
       |  ^~~~~~~~~~~~~~~~~~~
 core/arch/arm/mm/core_mmu_lpae.c:708:8: warning: unused variable ‘ret’ [-Wunused-variable]
   708 |   bool ret = false;
       |        ^~~

The type used for struct mmu_partition::user_l1_table_idx, currently
uint8_t, is not wide enough. Fix the issue by using uint8_t or uint16_t
based on the value of MAX_XLAT_TABLES.

### DIFF
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -216,6 +216,16 @@
 				 XLAT_TABLE_USER_EXTRA)
 #endif /*!MAX_XLAT_TABLES*/
 
+#if (CORE_MMU_BASE_TABLE_LEVEL == 0)
+#if (MAX_XLAT_TABLES <= UINT8_MAX)
+typedef uint8_t l1_idx_t;
+#elif (MAX_XLAT_TABLES <= UINT16_MAX)
+typedef uint16_t l1_idx_t;
+#else
+#error MAX_XLAT_TABLES is suspiciously large, please check
+#endif
+#endif
+
 typedef uint64_t base_xlat_tbls_t[CFG_TEE_CORE_NB_CORE][NUM_BASE_LEVEL_ENTRIES];
 typedef uint64_t xlat_tbl_t[XLAT_TABLE_ENTRIES];
 
@@ -257,7 +267,7 @@ struct mmu_partition {
 	 * Indexes of the L1 table from 'xlat_tables'
 	 * that points to the user mappings.
 	 */
-	uint8_t user_l1_table_idx[NUM_BASE_TABLES][CFG_TEE_CORE_NB_CORE];
+	l1_idx_t user_l1_table_idx[NUM_BASE_TABLES][CFG_TEE_CORE_NB_CORE];
 #endif
 };
 


core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -216,6 +216,16 @@
 				 XLAT_TABLE_USER_EXTRA)
 #endif /*!MAX_XLAT_TABLES*/
 
+#if (CORE_MMU_BASE_TABLE_LEVEL == 0)
+#if (MAX_XLAT_TABLES < (1 << 8))
+#define l1_idx_type uint8_t
```
jenswi-linaro, line 221: How about using
``` C
typedef uint8_t l1_idx_t;
```
instead? Using `#define` instead of typedef looks like a funny way of avoiding a checkpatch warning.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -216,6 +216,16 @@
 				 XLAT_TABLE_USER_EXTRA)
 #endif /*!MAX_XLAT_TABLES*/
 
+#if (CORE_MMU_BASE_TABLE_LEVEL == 0)
+#if (MAX_XLAT_TABLES < (1 << 8))
```
jenswi-linaro, line 220: `BIT(8)`?

core/arch/arm/mm/core_mmu_lpae.c
Izikd: agree. If the typedef check is irrelevant for this project, then it can be disabled in `.checkpatch.conf` in the root of the project.

jforissier: @jenswi-linaro agreed.
@Izikd we could disable the check easily, yes, but I think I'd rather keep it. Most of the time a new typedef is not what we want and this case is an exception.

core/arch/arm/mm/core_mmu_lpae.c
jforissier: Yup

jforissier:
Update

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -216,6 +216,16 @@
 				 XLAT_TABLE_USER_EXTRA)
 #endif /*!MAX_XLAT_TABLES*/
 
+#if (CORE_MMU_BASE_TABLE_LEVEL == 0)
+#if (MAX_XLAT_TABLES <= UINT8_MAX)
+typedef uint8_t l1_idx_type;
```
jenswi-linaro, line 221: I'd prefer `l1_idx_t`.

jforissier: Sure.

jenswi-linaro (commented):

PR #5036, opened 2021-12-08
core: pm: fix incorrect flag check in pm_callback

ldebieve:
Fix test error check that always return true with current
the condition. The check must be done to identify if the
SUSPENDED flag as been set.

Signed-off-by: Lionel Debieve <lionel.debieve@foss.st.com>

### DIFF
--- a/core/kernel/pm.c
+++ b/core/kernel/pm.c
@@ -65,7 +65,7 @@ static TEE_Result do_pm_callback(enum pm_op op, uint32_t pm_hint,
 	TEE_Result res = TEE_ERROR_GENERIC;
 	bool suspending = op == PM_OP_SUSPEND;
 
-	if (suspending == (bool)(hdl->flags | PM_FLAG_SUSPENDED))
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):
Nice catch! Thanks for the fix.
Could you add a `Fixes:` tag in the commit log?

etienne-lms:
also fix commit log:
> Fix test error check that always return**s** true with current
> **~~the~~** condition. The check must be done to identify if the
> SUSPENDED flag **h**as been set.

`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with commit log fixed.

ldebieve:
@etienne-lms Fixing commit message + Fixes, thanks

jenswi-linaro:

jforissier:
Nit: no need for a blank line between Fixes: and the other tags.

ldebieve:
Remove blank line + Tags applied.

PR #5043, opened 2021-12-09
drivers: sam: add shutdown controller driver

clementleger:
This PR adds driver for atmel SAM shutdown controller.

### DIFF
--- a/core/drivers/atmel_shdwc.c
+++ b/core/drivers/atmel_shdwc.c
@@ -1,0 +1,182 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2015 Atmel Corporation,
+ *                    Nicolas Ferre <nicolas.ferre@atmel.com>
+ * Copyright (c) 2021, Microchip
+ */
+
+#include <drivers/atmel_shdwc.h>
+#include <drivers/sam/at91_ddr.h>
+#include <io.h>
+#include <kernel/dt.h>
+#include <kernel/thread.h>
+#include <libfdt.h>
+#include <stdbool.h>
+#include <tee_api_defines.h>
+#include <tee_api_types.h>
+#include <trace.h>
+#include <types_ext.h>
+#include <util.h>
+
+#include "at91_clk.h"
+
+#define SHDW_WK_PIN(reg, cfg)	((reg) & \
+					AT91_SHDW_WKUPIS((cfg)->wkup_pin_input))
+#define SHDW_RTCWK(reg, cfg)	(((reg) >> ((cfg)->sr_rtcwk_shift)) & 0x1)
+#define SHDW_RTTWK(reg, cfg)	(((reg) >> ((cfg)->sr_rttwk_shift)) & 0x1)
+#define SHDW_RTCWKEN(cfg)	BIT((cfg)->mr_rtcwk_shift)
+#define SHDW_RTTWKEN(cfg)	BIT((cfg)->mr_rttwk_shift)
+
+#define SLOW_CLK_FREQ		32768ULL
+#define DBC_PERIOD_US(x)	DIV_ROUND_UP((1000000ULL * (x)), SLOW_CLK_FREQ)
+
+static vaddr_t shdwc_base;
+static vaddr_t mpddrc_base;
+
+bool atmel_shdwc_available(void)
+{
+	return shdwc_base != 0;
+}
+
+void __noreturn atmel_shdwc_shutdown(void)
+{
+	vaddr_t pmc_base = at91_pmc_get_base();
+
+	/*
+	 * Mask exception before entering assembly which does not expect to be
+	 * interrupted.
+	 */
+	thread_mask_exceptions(THREAD_EXCP_ALL);
+
+	__atmel_shdwc_shutdown(mpddrc_base, shdwc_base, pmc_base);
+
+	/* We are going to shutdown the CPU so we will never hit this loop */
+	while (true)
+		;
+}
+
+static const unsigned long long sdwc_dbc_period[] = {
+	0, 3, 32, 512, 4096, 32768,
+};
+
+static uint32_t at91_shdwc_debouncer_value(uint32_t in_period_us)
+{
+	int i = 0;
+	int max_idx = ARRAY_SIZE(sdwc_dbc_period) - 1;
+	uint64_t period_us = 0;
+	uint64_t max_period_us = DBC_PERIOD_US(sdwc_dbc_period[max_idx]);
+
+	if (in_period_us > max_period_us) {
+		DMSG("debouncer period %"PRIu32" too big, using %"PRIu64" us",
+		     in_period_us, max_period_us);
+		return max_idx;
+	}
+
+	for (i = max_idx - 1; i > 0; i--) {
+		period_us = DBC_PERIOD_US(sdwc_dbc_period[i]);
+		if (in_period_us > period_us)
+			break;
+	}
+
+	return i + 1;
+}
+
+static uint32_t at91_shdwc_get_wakeup_input(const void *fdt, int np)
+{
+	const uint32_t *prop = NULL;
+	uint32_t wk_input_mask = 0;
+	uint32_t wuir = 0;
+	uint32_t wk_input = 0;
+	int child = 0;
+	int len = 0;
+
+	fdt_for_each_subnode(child, fdt, np) {
+		prop = fdt_getprop(fdt, child, "reg", &len);
+		if (!prop || len != sizeof(uint32_t)) {
+			DMSG("reg property is missing for node %s",
+			     fdt_get_name(fdt, child, NULL));
+			continue;
+		}
+		wk_input = fdt32_to_cpu(*prop);
+		wk_input_mask = BIT32(wk_input);
+		if (!(wk_input_mask & AT91_SHDW_WKUPEN_MASK)) {
+			DMSG("wake-up input %"PRId32" out of bounds ignore",
+			     wk_input);
+			continue;
+		}
+		wuir |= wk_input_mask;
+
+		if (fdt_getprop(fdt, child, "atmel,wakeup-active-high", NULL))
+			wuir |= AT91_SHDW_WKUPT(wk_input);
+	}
+
+	return wuir;
+}
+
+static void at91_shdwc_dt_configure(const void *fdt, int np)
+{
+	const uint32_t *prop = NULL;
+	uint32_t mode = 0;
+	uint32_t tmp = 0;
+	uint32_t input = 0;
+	int len = 0;
+
+	prop = fdt_getprop(fdt, np, "debounce-delay-us", &len);
+	if (prop && len == sizeof(uint32_t)) {
+		tmp = fdt32_to_cpu(*prop);
+		mode |= AT91_SHDW_WKUPDBC(at91_shdwc_debouncer_value(tmp));
+	}
+
+	if (fdt_getprop(fdt, np, "atmel,wakeup-rtc-timer", &len))
+		mode |= AT91_SHDW_RTCWKEN;
+
+	io_write32(shdwc_base + AT91_SHDW_MR, mode);
+
+	input = at91_shdwc_get_wakeup_input(fdt, np);
+	io_write32(shdwc_base + AT91_SHDW_WUIR, input);
+}
+
+static TEE_Result atmel_shdwc_probe(const void *fdt, int node,
+				    const void *compat_data __unused)
+{
+	int ddr_node = 0;
+	size_t size = 0;
+	uint32_t ddr = AT91_DDRSDRC_MD_LPDDR2;
+
+	/*
+	 * Assembly code relies on the fact that there is only one CPU to avoid
+	 * any other one to invalidate TLB/I-Cache.
+	 */
+	COMPILE_TIME_ASSERT(CFG_TEE_CORE_NB_CORE == 1);
+
+	if (dt_map_dev(fdt, node, &shdwc_base, &size) < 0)
+		return TEE_ERROR_GENERIC;
+
+	ddr_node = fdt_node_offset_by_compatible(fdt, -1,
+						 "atmel,sama5d3-ddramc");
+	if (ddr_node < 0)
+		return TEE_ERROR_GENERIC;
+
+	if (dt_map_dev(fdt, ddr_node, &mpddrc_base, &size) < 0)
+		return TEE_ERROR_GENERIC;
+
+	ddr = io_read32(mpddrc_base + AT91_DDRSDRC_MDR) & AT91_DDRSDRC_MD;
+	if (ddr != AT91_DDRSDRC_MD_LPDDR2 && ddr != AT91_DDRSDRC_MD_LPDDR3)
+		mpddrc_base = 0;
+
+	at91_shdwc_dt_configure(fdt, node);
+
+	return TEE_SUCCESS;
+}
+
+static const struct dt_device_match atmel_shdwc_match_table[] = {
+	{ .compatible = "atmel,sama5d2-shdwc" },
+	{ }
+};
+
+const struct dt_driver atmel_shdwc_dt_driver __dt_driver = {
+	.name = "atmel_shdwc",
+	.type = DT_DRIVER_NOTYPE,
+	.match_table = atmel_shdwc_match_table,
+	.probe = atmel_shdwc_probe,
+};
--- a/core/drivers/atmel_shdwc_a32.S
+++ b/core/drivers/atmel_shdwc_a32.S
@@ -1,0 +1,79 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2015 Atmel Corporation,
+ *                    Nicolas Ferre <nicolas.ferre@atmel.com>
+ * Copyright (c) 2021, Microchip
+ */
+
+
+#include <arm.h>
+#include <arm32_macros.S>
+#include <asm.S>
+#include <drivers/atmel_shdwc.h>
+#include <drivers/sam/at91_ddr.h>
+
+#include "at91_pmc.h"
+
+/*
+ * Code size of shutdown assembly must fit in a single Cortex-A5 cache
+ * line which is 8 words long (32 bytes) since SDRAM might be disabled and thus
+ * not accessible to fetch code or data from it.
+ */
+.macro check_fit_in_cacheline since
+	.if (. - \since) > 32
+		.error "Shutdown assembly code exceeds cache line size"
+	.endif
+.endm
+
+/**
+ * Shutdown the CPU
+ *
+ * This function is in assembly to be sure the code fits in a single cache line.
+ * We are going to power down the SDRAM and thus we can't fetch code from it
+ * once powered down.
+ *
+ * r0 = mpddrc_base
+ * r1 = shdwc_base
+ * r2 = pmc_base
+ */
+FUNC __atmel_shdwc_shutdown , :
+
+	mov_imm r3, AT91_DDRSDRC_LPDDR2_PWOFF
+	mov_imm r4, (AT91_SHDW_KEY | AT91_SHDW_SHDW)
+
+	/*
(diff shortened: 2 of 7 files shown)


core/drivers/atmel_shdwc.c
```diff
@@ -0,0 +1,194 @@
...
+bool atmel_shdwc_available(void)
+{
+	return shdwc_base != 0;
+}
+
+void __noreturn atmel_shdwc_shutdown(void)
+{
+	vaddr_t pmc_base = at91_pmc_get_base();
+
+	asm volatile (
```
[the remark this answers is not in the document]
clementleger: Ok, best solution for this is to move it in assembly file. The output must be controlled to avoid being larger than a cache line since we power down the SDRAM.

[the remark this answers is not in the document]
jforissier: Sounds good. Would you mind adding a comment in the assembly file to explain why it has to be written in assembly?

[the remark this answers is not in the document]
clementleger: Sure !

clementleger:
New version:
- Move inline assembly to assembly file .S

core/include/drivers/sam/at91_ddr.h
```diff
@@ -0,0 +1,187 @@
...
+ *	Nicolas Ferre <nicolas.ferre@atmel.com>
+ */
+
+#ifndef AT91_DDRSDR_H
+#define AT91_DDRSDR_H
+
+/* Mode Register */
+#define AT91_DDRSDRC_MR				0x00
+/* Command Mode */
+#define	AT91_DDRSDRC_MODE			(0x7 << 0)
```
jenswi-linaro, line 15: Please use the `SHIFT_U32()` macro instead of `<<` in this file.

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,59 @@
...
+ * r2 = pmc_base
+ */
+FUNC atmel_shdwc_shutdown_asm , :
+
+	ldr r3, =AT91_DDRSDRC_LPDDR2_PWOFF
+	ldr r4, =(AT91_SHDW_KEY | AT91_SHDW_SHDW)
+
+/* Align to cache line */
+.balign 32
+	/* Ensure AT91_SHDW_CR is in the TLB by reading it */
```
jenswi-linaro, line 34: For how long do you expect it will stay here?
Any chance that it can get evicted earlier than expected?

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,59 @@
...
+ * We are going to power down the SDRAM and thus we can't fetch code from it
+ * once powered down.
+ *
+ * r0 = mpddrc_base
+ * r1 = shdwc_base
+ * r2 = pmc_base
+ */
+FUNC atmel_shdwc_shutdown_asm , :
+
+	ldr r3, =AT91_DDRSDRC_LPDDR2_PWOFF
```
jenswi-linaro, line 29: Please use the `mov_imm` assembly macro instead.

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,59 @@
...
+	ldr r4, =(AT91_SHDW_KEY | AT91_SHDW_SHDW)
+
+/* Align to cache line */
+.balign 32
+	/* Ensure AT91_SHDW_CR is in the TLB by reading it */
+	ldr	r6, [r1, #AT91_SHDW_CR]
+
+	/* Power down SDRAM0 */
```
jenswi-linaro, line 37: Well, maybe power down...

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,59 @@
...
+#include <asm.S>
+#include <at91_pmc.h>
+#include <drivers/atmel_shdwc.h>
+#include <drivers/sam/at91_ddr.h>
+
+/**
+ * Shutdown the CPU
+ *
+ * This function is in assembly to be sure the code fits in a single cache line.
+ * We are going to power down the SDRAM and thus we can't fetch code from it
```
jenswi-linaro, line 20: So you're depending on the cache line staying in the icache? Any guarantees for that?

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,59 @@
...
+ * once powered down.
+ *
+ * r0 = mpddrc_base
+ * r1 = shdwc_base
+ * r2 = pmc_base
+ */
+FUNC atmel_shdwc_shutdown_asm , :
```
jenswi-linaro, line 27: I'd prefer `__atmel_shdwc_shutdown` or `atmel_shdwc_shutdown_helper`.

core/drivers/atmel_shdwc_a32.S
clementleger: If no other access are made after that, the entry should still be in the TLB after that since the micro TLB is 10 entries wide.
So at least, PMC, MPDDRC, SHDWC and current code translation will fit in it.

To be sure of that, I will also disable exception before entering this piece of code so nothing will happen in this path. I also noticed that with this code, the PMC is not accessed before powering down the SDRAM but this does not affect the sama5d2_xplained because we don't have LPDDR on it.

core/drivers/atmel_shdwc_a32.S
clementleger: This was the intended purpose of this block of code taken from Linux sama5d2 shutdown controller. But after looking at it carefully, it probably does not work as-is.
The cache line on Cortex A5 is 8 words long (32 bytes) The actual code is 40 bytes long. So there is no guarantee that the code will fit in a single cache line. The code was originally working but ended up being modified non working IMHO.
I would say that if the code is aligned on 32 bytes and is smaller than 8 bytes, there is no reason for it to be evicted (unless there are some exceptions happening which as I say will be disabled) . I could also add a compile time check to be sure the code is indeed less than 32 bytes.

In order to make this code working, I will modified it to align on a cache line just before powering the SDRAM (`str	r3, [r0, #AT91_DDRSDRC_LPR]`), this result in the end of the code being 8 bytes long (minus the infinite loop). But either the SDRAM is not powered down and the infinite loop will be taken, or the SDRAM is powered down and this line will probably never be loaded from SDRAM and the processor will stall.

core/drivers/atmel_shdwc_a32.S
clementleger: Ok.

core/drivers/atmel_shdwc_a32.S
clementleger: `/* Power down SDRAM0 if mpddrc_base is set */`

core/drivers/atmel_shdwc_a32.S
jenswi-linaro: What if there's an icache invalidate from another CPU as part of some non-secure attack? Can that cause the cache line to be removed?
Perhaps that's not a security issue, but only a DOS.

core/drivers/atmel_shdwc_a32.S
jenswi-linaro: What if another CPU invalidates the TLB?
I guess this might become more of a DOS than a security issue.

core/drivers/atmel_shdwc_a32.S
jenswi-linaro: OK, thanks.

jenswi-linaro:

clementleger:
New version:
- Added Acked-by
- Add missing stdint include in header

core/drivers/atmel_shdwc.c
```diff
@@ -0,0 +1,183 @@
...
+				    const void *compat_data __unused)
+{
+	int ddr_node = 0;
+	size_t size = 0;
+	uint32_t ddr = AT91_DDRSDRC_MD_LPDDR2;
+
+	/* Code size of assembly code for shutdown must fit in a single
+	 * Cortex-A5 cache line which is 45 words long.
+	 */
+	assert(__atmel_shdwc_shutdown_cache_sz <= 32);
```
jforissier, line 146: Why not use a compile-time assertion in assembly like https://github.com/OP-TEE/optee_os/blob/3.15.0/core/arch/arm/kernel/entry_a64.S#L468-L478?

jforissier (commented):
- For commit "drivers: sam: add ddr.h file with DDR register definition":
s/ddr.h/at91_ddr.h
- See one more comment below

With that addressed:

core/drivers/atmel_shdwc.c
clementleger: Because I did not know how to do it :) I will use that.

clementleger:
@jforissier : Not sure if your Acked-by is meant for "drivers: atmel-shdwc: add atmel shdwc driver" or only for "drivers: sam: add ddr.h file with DDR register definition" since your comment targets the first one.

jforissier:
@clementleger my Ack is for all commits up to the point the Ack is given, it is what should be understood in general when no specific commit is mentioned. So, same for Jen's Ack, you can apply it to all commits.

clementleger:
Ok thanks

jforissier:
Would you mind replacing ddr.h with at91_ddr.h in the commit subject: "drivers: sam: add ddr.h file with DDR register definition"?

clementleger:
> Would you mind replacing ddr.h with at91_ddr.h in the commit subject: "drivers: sam: add ddr.h file with DDR register definition"?

Done !

clementleger:
New version:
- Rename ddr.h -> at91_ddr.h
- Rename first commit to use `drivers: clk: sam:` instead of `plat-sam:`

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,78 @@
...
+	tst	r6, #AT91_PMC_MCKRDY
+	beq	2b
+
+	/* Shutdown CPU */
+	str	r4, [r1, #AT91_SHDW_CR]
+
+	check_fit_in_cacheline __atmel_shdwc_shutdown_sdram_disabled
+
+	/* We should never reach this since we shut down the CPU */
+	b	.
```
etienne-lms, line 77: With speculation, could CPU speculatively execution this instruction while SDRAM is off? (in which case maybe prefer `b .` before `check_fit_...`.

core/include/drivers/atmel_shdwc.h
```diff
@@ -0,0 +1,65 @@
...
+#include <compiler.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+/* Shut Down Control Register */
+#define AT91_SHDW_CR		0x00
+/* Shut Down command */
+#define AT91_SHDW_SHDW		BIT(0)
+/* KEY Password */
+#define AT91_SHDW_KEY		(0xa5UL << 24)
```
etienne-lms, line 20: `SHIFT_U32()`? (from previous equiv comments)

core/include/drivers/atmel_shdwc.h
```diff
@@ -0,0 +1,65 @@
...
+#define AT91_SHDW_WKUPDBC(x)	(((x) << AT91_SHDW_WKUPDBC_SHIFT) \
+						& AT91_SHDW_WKUPDBC_MASK)
+#define AT91_SHDW_RTCWKEN	BIT32(17)
+
+/* Shut Down Status Register */
+#define AT91_SHDW_SR		0x08
+#define AT91_SHDW_WKUPIS_SHIFT	16
+#define AT91_SHDW_WKUPIS_MASK	GENMASK_32(31, 16)
+#define AT91_SHDW_WKUPIS(x)	((1 << (x)) << AT91_SHDW_WKUPIS_SHIFT \
+						& AT91_SHDW_WKUPIS_MASK)
```
etienne-lms, line 35: `define AT91_SHDW_WKUPIS(x)	BIT32((x) + AT91_SHDW_WKUPIS_SHIFT)`?

core/include/drivers/atmel_shdwc.h
```diff
@@ -0,0 +1,65 @@
...
+#define AT91_SHDW_SHDW		BIT(0)
+/* KEY Password */
+#define AT91_SHDW_KEY		(0xa5UL << 24)
+
+/* Shut Down Mode Register */
+#define AT91_SHDW_MR		0x04
+#define AT91_SHDW_WKUPDBC_SHIFT	24
+#define AT91_SHDW_WKUPDBC_MASK	GENMASK_32(26, 24)
+#define AT91_SHDW_WKUPDBC(x)	(((x) << AT91_SHDW_WKUPDBC_SHIFT) \
+						& AT91_SHDW_WKUPDBC_MASK)
```
etienne-lms, line 27:
```patch
-#define AT91_SHDW_WKUPDBC(x)	(((x) << AT91_SHDW_WKUPDBC_SHIFT) \
-						& AT91_SHDW_WKUPDBC_MASK)
+#define AT91_SHDW_WKUPDBC(x)	(((x) << AT91_SHDW_WKUPDBC_SHIFT) & \
+				 AT91_SHDW_WKUPDBC_MASK)
```

Other occurrences below.  (maybe used `BIT32()` in thoses)

etienne-lms (commented):
LGTM but few questions (and minor comments to address or not).

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	ldr	r6, [r1, #AT91_SHDW_CR]
+	ldr	r6, [r2, #AT91_PMC_MCKR]
+
+	/* Power down SDRAM0 if mpddrc_base is set */
+	tst	r0, #0
+	beq	1f
+
+/* Align to cache line to ensure the rest of code fits in a single line */
+.balign 32
```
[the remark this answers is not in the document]
clementleger: Yes, nops are generated. The GAS manual state that the void is filled either with 0s or on some systems with nops. On ARM32, nops are generated.

https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_7.html#SEC74

core/drivers/atmel_shdwc_a32.S
clementleger: Cortex A5 does not have speculative execution nor out-of-order execution capabilities.
https://en.wikipedia.org/wiki/Comparison_of_CPU_microarchitectures

Moreover, I don't really think this kind of thing can happen. The store would at least be issued by the processor before entering the loop.

core/drivers/atmel_shdwc.c
```diff
@@ -0,0 +1,179 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2015 Atmel Corporation,
+ *                    Nicolas Ferre <nicolas.ferre@atmel.com>
+ * Copyright (c) 2021, Microchip
+ */
+
+#include <at91_clk.h>
```
etienne-lms, line 8: sorry, i missed that.
should be `#include "at91_clk.h"`, below the diamond `#include <>`'s.

core/drivers/atmel_shdwc_a32.S
```diff
@@ -0,0 +1,78 @@
...
+ * Copyright (c) 2015 Atmel Corporation,
+ *                    Nicolas Ferre <nicolas.ferre@atmel.com>
+ * Copyright (c) 2021, Microchip
+ */
+
+
+#include <arm.h>
+#include <arm32_macros.S>
+#include <asm.S>
+#include <at91_pmc.h>
```
etienne-lms, line 12: ditto: `#include "at91_pmc.h"` below line 14.

core/include/drivers/atmel_shdwc.h
```diff
@@ -0,0 +1,65 @@
...
+#define AT91_SHDW_SHDW		BIT(0)
+/* KEY Password */
+#define AT91_SHDW_KEY		SHIFT_U32(0xa5UL, 24)
+
+/* Shut Down Mode Register */
+#define AT91_SHDW_MR		0x04
+#define AT91_SHDW_WKUPDBC_SHIFT	24
+#define AT91_SHDW_WKUPDBC_MASK	GENMASK_32(26, 24)
+#define AT91_SHDW_WKUPDBC(x)	(SHIFT_U32((x), AT91_SHDW_WKUPDBC_SHIFT) \
+				 & AT91_SHDW_WKUPDBC_MASK)
```
etienne-lms, line 27:
```patch
-#define AT91_SHDW_WKUPDBC(x)	(SHIFT_U32((x), AT91_SHDW_WKUPDBC_SHIFT) \
-				 & AT91_SHDW_WKUPDBC_MASK)
+#define AT91_SHDW_WKUPDBC(x)	(SHIFT_U32((x), AT91_SHDW_WKUPDBC_SHIFT) & \
+				 AT91_SHDW_WKUPDBC_MASK)

File: core/include/drivers/atmel_shdwc.h [hunk 17]
```diff
@@ -0,0 +1,65 @@
...
+#define AT91_SHDW_RTCWKEN	BIT32(17)
+
+/* Shut Down Status Register */
+#define AT91_SHDW_SR		0x08
+#define AT91_SHDW_WKUPIS_SHIFT	16
+#define AT91_SHDW_WKUPIS_MASK	GENMASK_32(31, 16)
+#define AT91_SHDW_WKUPIS(x)	(BIT32((x) + AT91_SHDW_WKUPIS_SHIFT) \
+				 & AT91_SHDW_WKUPIS_MASK)
```
etienne-lms, line 35:
```patch
-#define AT91_SHDW_WKUPIS(x)	(BIT32((x) + AT91_SHDW_WKUPIS_SHIFT) \
-				 & AT91_SHDW_WKUPIS_MASK)
+#define AT91_SHDW_WKUPIS(x)	(BIT32((x) + AT91_SHDW_WKUPIS_SHIFT) & \
+				 AT91_SHDW_WKUPIS_MASK)
```
since the mask value, could use:
`#define AT91_SHDW_WKUPIS(x)	BIT32((x) + AT91_SHDW_WKUPIS_SHIFT)`

core/drivers/atmel_shdwc_a32.S
etienne-lms: ok, thanks.

core/drivers/atmel_shdwc_a32.S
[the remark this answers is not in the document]
etienne-lms: ok. thanks for the link.

core/include/drivers/atmel_shdwc.h
clementleger: Ditto for `AT91_SHDW_WKUPT`

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with the 2 last comments addressed.

core/drivers/atmel_shdwc.c
```diff
@@ -0,0 +1,181 @@
...
+		prop = fdt_getprop(fdt, child, "reg", &len);
+		if (!prop || len != sizeof(uint32_t)) {
+			DMSG("reg property is missing for node %s",
+			     fdt_get_name(fdt, child, NULL));
+			continue;
+		}
+		wk_input = fdt32_to_cpu(*prop);
+		wk_input_mask = BIT32(wk_input);
+		if (!(wk_input_mask & AT91_SHDW_WKUPEN_MASK)) {
+			DMSG("wake-up input %d out of bounds ignore", wk_input);
```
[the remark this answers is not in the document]
clementleger: Fixed other DMSG line 70 too.

clementleger:
New version:
 - Fixed after last comment
 - Has a 80 column error happens on a DMSG string too long, does it need to be splitted in some say ?

jforissier:
> Has a 80 column error happens on a DMSG string too long, does it need to be splitted in some way ?

It can be ignored, but how about replacing "reduced to" by "using"? Should fit then.

clementleger:
New version:
-  No warning remaining

jforissier:
IBART failed due to `https://git.savannah.gnu.org/git/grub.git` not reachable (not the first time IIRC...). Let's ignore this hopefully temporary failure. Thanks @clementleger.

PR #5044, opened 2021-12-09
driver: atmel_trng: remove wrong assertion

clementleger:
Remove mistakenly added assert which will always trigger in debug
mode.


jforissier (approved):

PR #4985, opened 2021-11-18
plat-stm32mp1: port platform clock driver to clk framework

etienne-lms:
This P-R implements changes in stm32mp1 clock driver to register clocks to the standard clock framework.
This platform will move to lightweight clocks (https://github.com/OP-TEE/optee_os/pull/4978) if/once their are integrated, so as to lower pressure on unpaged resisdent secure memory.

PR #5047, opened 2021-12-11
Helper macro DEFINE_RODATA_UNPAGED()

etienne-lms:
Follow-up of comment thread starting here https://github.com/OP-TEE/optee_os/pull/4985#issuecomment-985582516 with @jforissier, here is a proposal for a helper macro `DEFINE_RODATA_UNPAGED()` as a complement to existing `__rodata_unpage()` and the need to define the variable twice.

Note checkpatch is not happy with the macro definition. It looks good to me however and i would treat reported error as a false positive. That said, suggestions are welcome if one sees a way to fix the issues.

```c
Checking commit(s):
ea0a4f66c core: DEFINE_RODATA_UNPAGED() helper macro
WARNING: Possible unwrapped commit description (prefer a maximum 75 chars per line)
#13: 
ERROR: Macros with multiple statements should be enclosed in a do - while loop

ERROR: Macros with multiple statements should be enclosed in a do - while loop
#53: FILE: core/include/keep.h:55:
+#define DEFINE_RODATA_UNPAGED(_type, _label) \
+	const _type _label __rodata_dummy; \
+	const _type _label __weak __rodata_unpaged(#_label)

CHECK: Macro argument reuse '_type' - possible side-effects?
#53: FILE: core/include/keep.h:55:
+#define DEFINE_RODATA_UNPAGED(_type, _label) \
+	const _type _label __rodata_dummy; \
+	const _type _label __weak __rodata_unpaged(#_label)

CHECK: Macro argument reuse '_label' - possible side-effects?
#53: FILE: core/include/keep.h:55:
+#define DEFINE_RODATA_UNPAGED(_type, _label) \
+	const _type _label __rodata_dummy; \
+	const _type _label __weak __rodata_unpaged(#_label)

total: 1 errors, 1 warnings, 2 checks, 19 lines checked
```

### DIFF
--- a/core/include/keep.h
+++ b/core/include/keep.h
@@ -33,6 +33,11 @@
 		__section("__keep_meta_vars_pager") = (unsigned long)&(sym)
 
 #define __DECLARE_KEEP_PAGER1(sym, file_id) __DECLARE_KEEP_PAGER2(sym, file_id)
+
+/*
+ * DECLARE_KEEP_PAGER() - Resource and its dependencies are linked in
+ * an unpaged section
+ */
 #define DECLARE_KEEP_PAGER(sym) __DECLARE_KEEP_PAGER1(sym, __FILE_ID__)
 
 #define __DECLARE_KEEP_INIT2(sym, file_id) \
@@ -41,6 +46,11 @@
 		__section("__keep_meta_vars_init") = (unsigned long)&(sym)
 
 #define __DECLARE_KEEP_INIT1(sym, file_id) __DECLARE_KEEP_INIT2(sym, file_id)
+
+/*
+ * DECLARE_KEEP_INIT() - Resource and its dependencies are linked in
+ * an init (a.k.a pageable pre-mapped) section
+ */
 #define DECLARE_KEEP_INIT(sym) __DECLARE_KEEP_INIT1(sym, __FILE_ID__)
 
 /*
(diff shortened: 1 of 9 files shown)


core/include/keep.h
```diff
@@ -43,6 +43,19 @@
...
+ *
+ * Define a const variable linked into an unpaged read-only section without
+ * propagating its unpaged constrain to the references the variable may refer
+ * to. This requires defining it twice with specific attributes for pager
+ * linker management. The variable is global (not static) since using __weak
+ * attribute for the dual definition.
+ */
+#define DEFINE_RODATA_UNPAGED(_type, _label) \
+	const _type _label __rodata_dummy; \
+	const _type _label __weak __rodata_unpaged(#_label)
```
jforissier, line 67: You are defining the same variable twice with different sections attributes, in the same compilation unit. Does this even compile? In the current code we have separate C files (`link_dummies_paged.c` and `link_dummies_init.c`).

etienne-lms: that's right. didn't check... this is stupidly a to simple macro. doesn't work.
I close the P-R.

etienne-lms:
pointless, closing.

core/include/keep.h
jforissier: No worries, thanks for trying to simplify things :)

PR #5048, opened 2021-12-11
crypto: drivers: se050 updates

ldts:
ECC sign fix:
----------------

The crypto API validates the size of the buffer that will hold the
resulting signature. This means that the SE05X driver can not use the
variable length buffer mechanism to request extra bytes to handle the
DER format.

To address this situation, this patch allocates a temporary buffer to
get the signature from the Plug-and-Trust subsystem; then, upon doing
the DER to binary conversion, copies the resulting data to the output
buffer.

OEFID runtime detection:
----------------------------------

The CFG_CORE_SE05X_OEFID definition is not required as the SE05X OEFID
can be read during early init - before the SCP03 session has been
established.
    
However we can continue to define it optionally: when that occurs, the OP-TEE driver 
will only initialize when such OEFID is  available.

FIX rsa encrypt/decrypt
------------------------------
 Fix input/output buffers (they were swapped).
 Fix algorithm selection for RSAES

Test:
  
```
  echo "data to sign (max 100 bytes)" > data

  openssl rsautl -encrypt -inkey rsa-pubkey.pub \
                  -in data -pubin -out data.crypt

  pkcs11-tool --module /usr/lib/libckteec.so.0.1 \
          --pin 87654321 --decrypt --id 01 \
          --token-label fio --mechanism RSA-PKCS \
	  --input-file data.crypt > data.decrypted

 diff data data.decrypted
```

### DIFF
--- a/core/drivers/crypto/se050/adaptors/utils/info.c
+++ b/core/drivers/crypto/se050/adaptors/utils/info.c
@@ -11,17 +11,16 @@
 #include <se05x_tlv.h>
 #include <smCom.h>
 #include <string.h>
+#include <util.h>
 
 /* Force the output until the P&T stack fixes its verbosity */
 #define LOG_MAU8_I(msg, buf, len) nLog_au8("Info", 0xff, msg, buf, len)
 #define LOG_I(format, ...) nLog("Info", 0xff, format, ##__VA_ARGS__)
-
 #define LOG_E(format, ...) nLog("Info", NX_LEVEL_ERROR, format, ##__VA_ARGS__)
-
 #define LOG_MAU8_E(msg, buf, len) \
 	nLog_au8("Info", NX_LEVEL_ERROR, msg, buf, len)
 
-static sss_status_t jcop4_get_id(void *ctx)
+static sss_status_t jcop4_get_id(void *ctx, bool display)
 {
 	char jcop_platform_id[17] = { 0 };
 	smStatus_t ret = SM_OK;
@@ -76,6 +75,10 @@ static sss_status_t jcop4_get_id(void *ctx)
 		return kStatus_SSS_Fail;
 	}
 
+	memcpy(se050_ctx.se_info.oefid, &rsp.vConfiguration_ID[2], 2);
+	if (!display)
+		return kStatus_SSS_Success;
+
 	LOG_I("SE050 JCOP4 Information:");
 	LOG_I("%s = 0x%02X", "Tag value - proprietary data 0xFE",
 	      rsp.vTag_value_proprietary_data);
@@ -92,7 +95,6 @@ static sss_status_t jcop4_get_id(void *ctx)
 	      rsp.vLength_configuration_ID);
 	LOG_MAU8_I("Configuration ID",
 		   rsp.vConfiguration_ID, sizeof(rsp.vConfiguration_ID));
-
 	LOG_MAU8_I("OEF ID", &rsp.vConfiguration_ID[2], 2);
 	LOG_I("%s = 0x%02X", "Tag patch ID (Must be 0x02)", rsp.vTag_patch_ID);
 	LOG_I("%s = 0x%02X", "Length patch ID 0x08", rsp.vLength_patch_ID);
@@ -149,7 +151,7 @@ static void show_config(uint16_t cfg)
 	}
 }
 
-static sss_status_t applet_get_id(sss_se05x_session_t *session)
+static sss_status_t applet_get_id(sss_se05x_session_t *session, bool display)
 {
 	SE05x_Result_t result = kSE05x_Result_NA;
 	smStatus_t ret = SM_OK;
@@ -170,8 +172,6 @@ static sss_status_t applet_get_id(sss_se05x_session_t *session)
 	if (ret != SM_OK)
 		return kStatus_SSS_Fail;
 
-	LOG_MAU8_I("Applet ID", uid, uidLen);
-
 	/*
 	 * VersionInfo is a 7 - byte value consisting of:
 	 * - 1 - byte Major applet version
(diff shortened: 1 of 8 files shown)


jforissier (commented):

ldts:
@jforissier updated with tags.

BTW do you think we could replace the current way we build the se050 driver -  ie linking the plug-and-trust. library [https://github.com/foundriesio/plug-and-trust] - with a git submodule instead so the SE05x driver is built in one pass together with the rest of OP-TEE? 

I think it will make it easier for everyone. shall I take a stab at it?

jforissier:
We don't have submodules currently so we should carefully consider what it means to add one. Ideally it shouldn't change anything except for the users of the submodule.
What do others think?

jenswi-linaro:
> We don't have submodules currently so we should carefully consider what it means to add one. Ideally it shouldn't change anything except for the users of the submodule. What do others think?

I'm not so keen on git submodules in the optee_os git. They tend to be a bit messy in my opinion. I don't think we need a git submodule to be able to reach code outside of the optee_os git.

ldts:
> > We don't have submodules currently so we should carefully consider what it means to add one. Ideally it shouldn't change anything except for the users of the submodule. What do others think?
> 
> I'm not so keen on git submodules in the optee_os git. They tend to be a bit messy in my opinion. I don't think we need a git submodule to be able to reach code outside of the optee_os git.

so @ricardosalveti was suggesting a CFG_SE05X_PLUG_AND_TRUST_SRC ....the OP-TEE user wanting to enable the SE050 would have to pull the tree/files somewhere before building OP-TEE;  would that be an acceptable solution? I'd really like to replace the current implementation which requires that we build the library pointing to the OP-TEE tree and then build OP-TEE linking against the library files we just built...is really a pain and somewhat fragile because all the macros.

core/drivers/crypto/se050/adaptors/utils/info.c
```diff
@@ -199,10 +204,37 @@ static sss_status_t applet_get_id(sss_se05x_session_t *session)
...
+			return ret;
+		}
+
+#ifdef CFG_CORE_SE05X_OEFID
+		/* validate the requested OEFID against the runtime detected */
+		oefid = SHIFT_U32(se050_ctx.se_info.oefid[0], 8) |
+			SHIFT_U32(se050_ctx.se_info.oefid[1], 0);
+
+		if (oefid != CFG_CORE_SE05X_OEFID) {
+			EMSG("OEFID configuration error, 0x%x != 0x%x",
```
etienne-lms, line 231: `EMSG("OEFID configuration error, 0x%x != 0x%"PRIx32,`

core/drivers/crypto/se050/session.c
```diff
@@ -39,10 +39,16 @@ TEE_Result se050_core_early_init(struct se050_scp_key *keys)
...
-	se050_display_board_info(se050_session);
-	/* the session must be closed after accessing board information */
+	sss_status_t status = kStatus_SSS_Success;
+
+	status = se050_get_se_info(se050_session,
+				   IS_ENABLED(CFG_CORE_SE05X_DISPLAY_INFO));
+	if (status != kStatus_SSS_Success)
+		return TEE_ERROR_GENERIC;
+
+	/* the session must be closed after accessing the board information */
```
etienne-lms, line 51: if so, should this close the session before returning an error at line 49?

etienne-lms (commented):
Commit "crypto: drivers: se050: ecc sign" 

Comment for "crypto: drivers: se050: OEFID runtime detection".

jenswi-linaro:
> > > We don't have submodules currently so we should carefully consider what it means to add one. Ideally it shouldn't change anything except for the users of the submodule. What do others think?
> > 
> > 
> > I'm not so keen on git submodules in the optee_os git. They tend to be a bit messy in my opinion. I don't think we need a git submodule to be able to reach code outside of the optee_os git.
> 
> so @ricardosalveti was suggesting a CFG_SE05X_PLUG_AND_TRUST_SRC ....the OP-TEE user wanting to enable the SE050 would have to pull the tree/files somewhere before building OP-TEE; would that be an acceptable solution? I'd really like to replace the current implementation which requires that we build the library pointing to the OP-TEE tree and then build OP-TEE linking against the library files we just built...is really a pain and somewhat fragile because all the macros.

Sounds good to me.

core/drivers/crypto/se050/adaptors/utils/info.c
ldts: ok

core/drivers/crypto/se050/session.c
ldts: ok. althought it doesnt really matter at this point. the device cant be recovered....will have to reboot.

ldts:
> > > > We don't have submodules currently so we should carefully consider what it means to add one. Ideally it shouldn't change anything except for the users of the submodule. What do others think?
> > > 
> > > 
> > > I'm not so keen on git submodules in the optee_os git. They tend to be a bit messy in my opinion. I don't think we need a git submodule to be able to reach code outside of the optee_os git.
> > 
> > 
> > so @ricardosalveti was suggesting a CFG_SE05X_PLUG_AND_TRUST_SRC ....the OP-TEE user wanting to enable the SE050 would have to pull the tree/files somewhere before building OP-TEE; would that be an acceptable solution? I'd really like to replace the current implementation which requires that we build the library pointing to the OP-TEE tree and then build OP-TEE linking against the library files we just built...is really a pain and somewhat fragile because all the macros.
> 
> Sounds good to me.

ok

ldts:
@etienne-lms @jforissier anything else pending on these? there is an important fix for rsa enc/dec - it was broken before.

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -463,13 +463,21 @@ static TEE_Result do_encrypt(struct drvcrypt_rsa_ed *rsa_data)
 {
 	switch (rsa_data->rsa_id) {
 	case DRVCRYPT_RSA_PKCS_V1_5:
+		return encrypt_es(TEE_ALG_RSAES_PKCS1_V1_5,
```
jforissier, line 466: Why is `rsa_data->hash_algo` not `TEE_ALG_RSAES_PKCS1_V1_5` in this case? It looks weird that you need to throw in a constant here while the OAEP case is dynamic.

jforissier (commented):
- For commit "crypto: drivers: se050: OEFID runtime detection":
- One question on "crypto: drivers: se050: fix rsa encrypt/decrypt".

core/drivers/crypto/se050/core/rsa.c
ldts: yes. I do agree, that looked weird to me too...I think the hash_algo field in the interface should have been just named algo and the algorithm always set.... but that is the way the API was written so I kind of assumed it was intentional.

maybe @clementfaure can comment?

ldts: @jforissier so to be clear, the patch under review just obeys the API. 

if there is a problem, it is with the Crypto API itself so its maintainers should comment.

jforissier: Ok then for "crypto: drivers: se050: fix rsa encrypt/decrypt":

ldts:
updated tags

ldts:
@jforissier anything else?

jforissier:
@ldts no, I am merging this now.

PR #5050, opened 2021-12-14
plat-sam: add psci support for reset and shutdown

clementleger:
This PR adds support for reset and shutdown via PSCI for plat-sam. It depends on reset controller and shutdown controller support which is currently under review.

### DIFF
--- a/core/arch/arm/plat-sam/pm/psci.c
+++ b/core/arch/arm/plat-sam/pm/psci.c
@@ -3,16 +3,38 @@
  * Copyright (c) 2021, Microchip
  */
 
+#include <drivers/atmel_rstc.h>
+#include <kernel/panic.h>
 #include <sm/psci.h>
 #include <sm/std_smc.h>
 #include <stdint.h>
+#include <trace.h>
+
+void __noreturn psci_system_reset(void)
+{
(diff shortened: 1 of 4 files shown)


core/arch/arm/plat-sam/pm/psci.c
```diff
@@ -3,16 +3,39 @@
  * Copyright (c) 2021, Microchip
  */
 
+#include <drivers/atmel_rstc.h>
```
etienne-lms, line 6: In commit "plat-sam: enable PSCI reset support": `console.h` should be before `drivers/xxx.h`.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment on commit "plat-sam: enable PSCI reset support" addressed.

core/arch/arm/plat-sam/pm/psci.c
clementleger: Even better, removed it since not needed anymore.

jenswi-linaro:

clementleger:
New version:
- Added Acked-by Jens

PR #5052, opened 2021-12-14
core: dt_driver: fix result argument description get_of_device_func

etienne-lms:
Fixes inline description of type get_of_device_func that falsely mentions
TEE_ERROR_BUSY instead of TEE_ERROR_DEFER_DRIVER_INIT when expected
resource requests deferral of the driver probing.

Fixes: d8b14b46af9d ("core: dt_driver: get return code when querying a device")


jforissier (commented):

etienne-lms:
tag applied.

PR #5056, opened 2021-12-14
lib: libutils: ext: mempool.c: remove unused thread.h

maroueneboubakri:
This commit removes `thread.h` from `lib/libutils/ext/mempool.c` since it is
not used, but also to eliminate the dependency between generic libraries
code and architecture-specific core functions.


jenswi-linaro:

PR #5057, opened 2021-12-14
drivers: clk: change API functions clk_dt_get_by_*()

etienne-lms:
* drivers: clk: change `clk_dt_get_by_*()` prototype
Changes `clk_dt_get_by_idx()` and `clk_dt_get_by_name()` to return a the `TEE_Result` code and use an output argument to pass back clock reference rather than the opposite. This change makes `clk_dt_get_by_*()` function more consistent with the other OP-TEE core API functions.
Also renames`clk_dt_get_by_idx()` to `clk_dt_get_by_index()`.
Updates _sama5d2_clk.c_ and _atmel_trng.c_ accordingly.

* drivers: clk: use `while (true)`
Changes _clk_dt.c_ to use `while (true)` instead of `while (1)` for consistency in optee_os implementation.

### DIFF
--- a/core/drivers/clk/clk_dt.c
+++ b/core/drivers/clk/clk_dt.c
@@ -13,36 +13,36 @@
 #include <libfdt.h>
 #include <stddef.h>
 
-struct clk *clk_dt_get_by_name(const void *fdt, int nodeoffset,
-			       const char *name, TEE_Result *res)
+TEE_Result clk_dt_get_by_name(const void *fdt, int nodeoffset,
+			      const char *name, struct clk **clk)
 {
 	int clk_id = 0;
 
 	clk_id = fdt_stringlist_search(fdt, nodeoffset, "clock-names", name);
 	if (clk_id < 0) {
-		*res = TEE_ERROR_GENERIC;
-		return NULL;
+		*clk = NULL;
+		return TEE_ERROR_GENERIC;
 	}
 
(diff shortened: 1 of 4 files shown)


core/drivers/clk/clk_dt.c
```diff
@@ -13,36 +13,36 @@
...
+					 struct clk **clk)
 {
-	void *device = dt_driver_device_from_node_idx_prop(prop_name, fdt,
-							   nodeoffset, clk_idx,
-							   DT_DRIVER_CLK,
-							   res);
+	TEE_Result res = TEE_ERROR_GENERIC;
 
-	return (struct clk *)device;
+	*clk = dt_driver_device_from_node_idx_prop(prop_name, fdt, nodeoffset,
```
jenswi-linaro, line 36: I suppose the same logic could be applied to `dt_driver_device_from_node_idx_prop()` then, but I guess we can save that for another PR.

jenswi-linaro (commented):

core/drivers/clk/clk_dt.c
etienne-lms: Yes, `dt_driver_device_from_node_idx_prop()`, could be changed accordingly. Returning `void *` is only needed at `xxx_dt_get()` callback level. I'll address in a next commit.

etienne-lms:
tag applied

PR #5059, opened 2021-12-15
ta: pkcs11: Clean up temporary_object_list in err condition

cmcgcc:
When handle_get() failed in create_object(), ssession obj is not removed from temporary_object_list.
Only token obj is removed inside cleanup_persistent_object().
Add LIST_REMOVE for session obj.

Signed-off-by: Mengchi Cheng <mengcc@amazon.com>

### DIFF
--- a/ta/pkcs11/src/object.c
+++ b/ta/pkcs11/src/object.c
@@ -85,6 +85,8 @@ static void cleanup_volatile_obj_ref(struct pkcs11_object *obj)
 	if (!obj)
 		return;
 
+	LIST_REMOVE(obj, link);
+
 	if (obj->key_handle != TEE_HANDLE_NULL)
 		TEE_FreeTransientObject(obj->key_handle);
 
@@ -119,8 +121,6 @@ void cleanup_persistent_object(struct pkcs11_object *obj,
 	obj->attribs_hdl = TEE_HANDLE_NULL;
 	destroy_object_uuid(token, obj);
 
-	LIST_REMOVE(obj, link);
-
 	cleanup_volatile_obj_ref(obj);
 }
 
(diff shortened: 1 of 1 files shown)


ta/pkcs11/src/object.c
```diff
@@ -291,8 +291,10 @@ enum pkcs11_rc create_object(void *sess, struct obj_attrs *head,
 	handle_put(get_object_handle_db(session), obj_handle);
 	if (get_bool(head, PKCS11_CKA_TOKEN))
 		cleanup_persistent_object(obj, session->token);
-	else
+	else {
+		LIST_REMOVE(obj, link);
```
etienne-lms, line 295: Thanks for the finding. It does the job but I think we could end up calling `LIST_REMOVE(obj, link)` only from `cleanup_volatile_obj_ref()`. It would replace this added call and those at lines 122 and 143. What do you think?

cmcgcc: cleanup_volatile_obj_ref() is also called inside cleanup_persistent_object().  LIST_REMOVE() will be called multiple times for token obj especially in the destroy_object(). I do not think it is a good behavior although it does not create error for now. Meanwhile, obj is explicitly inserted to list inside create_object(), it is much clearer to read if it is also removed inside the same function.

cmcgcc: Yeah, it is a good idea. Let me update the patch.

cmcgcc:
I have updated patch. Is there any more comments?

etienne-lms:
You need to fix with the commit message as per checkpatch. 76 char/line max.
I would rephrase a bit the message, my suggestion:
```
ta: pkcs11: Clean up temporary_object_list on object creation failure

Calls LIST_REMOVE() only from cleanup_volatile_obj_ref() this is always
called to finalize object release. Allocated objects are always inserted
into a list after being created and therefore need to be removed for its
owner list before the memory is released.

This changes fixes an issue when handle_get() failed in create_object()
and does not remove the reference from from temporary_object_list.
```

jforissier:
@cmcgcc would you please add Etienne's Reviewed-by? Then I can merge this PR. Thanks!

cmcgcc:
> @cmcgcc would you please add Etienne's Reviewed-by? Then I can merge this PR. Thanks!

Done

PR #5060, opened 2021-12-15
core: stmm: Increase the shared number of pages

apalos:
Currently we only allow single page sharing for the StanAloneMM non-secure
world buffer.  There are cases on EFI variables though which this isn't
enough.  For example an EFI signature list (.esl) containing more than
two keys would fail since the payload is larger than a single page.  So
let's bump the number to something more reasonable.

Note that we could add this on a config option with an upper boundary, but
four pages should be enough for embedded use cases

Signed-off-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>

### DIFF
--- a/core/arch/arm/kernel/stmm_sp.c
+++ b/core/arch/arm/kernel/stmm_sp.c
@@ -69,8 +69,8 @@ static const uint16_t ffa_storage_id = 4U;
 
 static const unsigned int stmm_stack_size = 4 * SMALL_PAGE_SIZE;
 static const unsigned int stmm_heap_size = 398 * SMALL_PAGE_SIZE;
-static const unsigned int stmm_sec_buf_size = SMALL_PAGE_SIZE;
-static const unsigned int stmm_ns_comm_buf_size = SMALL_PAGE_SIZE;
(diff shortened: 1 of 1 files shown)


grandpaul:
``
Tested-by: Ying-Chun Liu (PaulLiu) <paul.liu@linaro.org>
``

etienne-lms (commented):
I would remove the suggestion from the commit log.

I agree `CFG_STMM_xxx` config switches would be nice for the STMM static stack, heap and the 2 shm's. The values may depend on changes in StMM services or use.

apalos:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` I would remove the suggestion from the commit log.
> 
> I agree `CFG_STMM_xxx` config switches would be nice for the STMM static stack, heap and the 2 shm's. The values may depend on changes in StMM services or use.

Thanks Etienne!

Yea we could switch to those if we eventually need them.  However since this functionality depends on u-boot/op-tee and EDK2 I'd rather hardcode it something reasonable for now and make everyone lives easier.

jenswi-linaro:

Issue #5022, opened 2021-12-03
NXP: i.MX and LS Platforms: Stability issue with PKCS11 TA


sahilnxp:
Some errors and stability issues affect the PKCS11 TA which happens after several xtest runs. 
This issue can be seen on multiple LS and i.MX board with following commits:
optee_os: https://github.com/OP-TEE/optee_os/commit/d803b8857e50cc2a4d7fa86dcbcfe8c983cdbc0c
optee_client: https://github.com/OP-TEE/optee_client/commit/df537dd7b9147bd5aaba620f76cbf9dd7a07f0a3
optee_test: https://github.com/OP-TEE/optee_test/commit/5b319f83a8ffec1f7efed992fe37f29559c6544f

This issue can be reproduced without CAAM driver.
Please do the following to reproduce the issue:

`i=0; while xtest -l 15 pkcs11; do ((i++)) && echo "$i"; done`

ISSUE #1: The PKCS11 TA runs out of memory when generating the RSA keys.
The default heap size of the TA seems not enough to do several test runs.
Increasing the TA heap size seems to solve the issue:

```
ta/pkcs11/sub.mk

-CFG_PKCS11_TA_HEAP_SIZE ?= (16 * 1024) 
+CFG_PKCS11_TA_HEAP_SIZE ?= (64 * 1024)
```
ISSUE #2: TA panicked with code 0xffff0006
The following TA crash would sometime appear after few xtest runs:

```
o pkcs11_1021.18 RSA-3072: Sign & verify tests - oneshot - CKM_RSA_PKCS
E/TC:? 0
E/TC:? 0 TA panicked with code 0xffff0006
E/LD:  Status of TA fd02c9da-306c-48c7-a49c-bbd827ae86ee
E/LD:   arch: aarch64
E/LD:  region  0: va 0x40004000 pa 0xfe200000 size 0x002000 flags rw-s (ldelf)
E/LD:  region  1: va 0x40006000 pa 0xfe202000 size 0x008000 flags r-xs (ldelf)
E/LD:  region  2: va 0x4000e000 pa 0xfe20a000 size 0x001000 flags rw-s (ldelf)
E/LD:  region  3: va 0x4000f000 pa 0xfe20b000 size 0x004000 flags rw-s (ldelf)
E/LD:  region  4: va 0x40013000 pa 0xfe20f000 size 0x001000 flags r--s
[...]

PR #5065, opened 2021-12-17
core: libtomcrypt: increase MPI_MEMPOOL_SIZE to 46kB

sahilnxp:
This value is increased because xtest pkcs11_1019 when run
in loop, leads to extensive use of memory pool which
sometimes leads to memory allocation failure.

Problem is the way mempool_alloc() is implemented in combination
with how it's used from ltc_ecc_projective_add_point().

mempool_alloc() has a stack like allocation scheme. When freeing
the top element that memory is returned to the pool, but if memory
further down is free it's not returned until all elements above has
been freed. If two or more elements gets allocated and freed in a
cycle they can continue to use more and more memory with nothing
returned.

Fixes: https://github.com/OP-TEE/optee_os/issues/5022

### DIFF
--- a/core/lib/libtomcrypt/mpi_desc.c
+++ b/core/lib/libtomcrypt/mpi_desc.c
@@ -19,7 +19,7 @@
 #endif
 
 /* Size needed for xtest to pass reliably on both ARM32 and ARM64 */
-#define MPI_MEMPOOL_SIZE	(42 * 1024)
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

sahilnxp:

Tag updated

etienne-lms (commented):

sahilnxp:

Tag updated

PR #5066, opened 2021-12-20
mempool based on bget

jenswi-linaro:
This relates to the issue https://github.com/OP-TEE/optee_os/issues/5022 and the recently merged PR https://github.com/OP-TEE/optee_os/pull/5065

Reply to https://github.com/OP-TEE/optee_os/issues/5022#issuecomment-996885144
> > Interesting, I wonder how much it can increase. Perhaps it's time to start looking for a alternative implementation to the mempool. I have an idea where bget could actually be reused with this.
> 
> Yes, we should look for some other mempool implementation as it hogs memory which is already free in some cases. It would be great if we can improve this implementation with bget.

This PR attempts to do this. Note that I haven't tried to tune `MPI_MEMPOOL_SIZE`.

The purpose of mempool is to be able to guarantee that temporary memory allocations will succeed during asymmetric cipher operations. As long as we return the memory allocated with mempool_alloc() we can use it in other functions too. There's a couple of examples of that in the code already. So the rule is: if you allocate memory with mempool_alloc() you should free it again before returning.

### DIFF
--- a/lib/libutils/ext/mempool.c
+++ b/lib/libutils/ext/mempool.c
@@ -54,10 +54,10 @@
 
 struct mempool {
 	size_t size;  /* size of the memory pool, in bytes */
-	ssize_t last_offset;   /* offset to the last one */
 	vaddr_t data;
+	struct malloc_ctx *mctx;
 #ifdef CFG_MEMPOOL_REPORT_LAST_OFFSET
-	ssize_t max_last_offset;
+	size_t max_allocated;
 #endif
 #if defined(__KERNEL__)
 	void (*release_mem)(void *ptr, size_t size);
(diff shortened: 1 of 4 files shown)


lib/libutils/ext/mempool.c
```diff
@@ -117,47 +139,25 @@ mempool_alloc_pool(void *data, size_t size,
...
-	pool->last_offset = offset;
+	p = raw_malloc(0, 0, size, pool->mctx);
+	if (p) {
 #ifdef CFG_MEMPOOL_REPORT_LAST_OFFSET
-	if (pool->last_offset > pool->max_last_offset) {
-		pool->max_last_offset = pool->last_offset;
-		DMSG("Max memory usage increased to %zu",
-		     (size_t)pool->max_last_offset);
-	}
+		struct malloc_stats stats = { };
```
sahilnxp, line 149: This code block should also be in CFG_WITH_STATS along with CFG_MEMPOOL_REPORT_LAST_OFFSET

lib/libutils/isoc/bget_malloc.c
```diff
@@ -890,6 +890,29 @@ static bool gen_malloc_buffer_overlaps_heap(struct malloc_ctx *ctx,
...
+	memset(ctx, 0, sizeof(*ctx));
+	ctx->poolset.freelist.ql.flink = &ctx->poolset.freelist;
+	ctx->poolset.freelist.ql.blink = &ctx->poolset.freelist;
+}
+
+void raw_malloc_add_pool(struct malloc_ctx *ctx, void *buf, size_t len)
+{
+	gen_malloc_add_pool(ctx, buf, len);
+}
+
```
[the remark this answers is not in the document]
jenswi-linaro: I'll fix.

lib/libutils/ext/mempool.c
jenswi-linaro: I'll enable CFG_WITH_STATS from `mk/config.mk` when `CFG_MEMPOOL_REPORT_LAST_OFFSET=y`.

jenswi-linaro:
It seems I've mixed up which change should go into which commit. I'll fix that and force push. This needs to be rebased so I'll do that too.

jenswi-linaro:
Updated

sahilnxp:
`Acked-by: Sahil Malhotra <sahil.malhotra@nxp.com>`

jenswi-linaro:
Tags applied. Rebased on master to resolve a conflict.

PR #5067, opened 2021-12-20
core: use adr_l for global r/w variable threads

jenswi-linaro:
Replace an adr instruction with adr_l in thread_foreign_intr_exit to
make sure that the r/w global variable threads is accessible even if the
optee binary is very large.


jforissier (approved):

jenswi-linaro:
Tag applied

PR #5069, opened 2021-12-20
core: arch: kernel: move spmc functions from thread.h to thread_spmc.h

maroueneboubakri:
It is more relevant to declare `thread_spmc_populate_mobj_from_rx()` and
`thread_spmc_relinquish()` in `thread_spmc.h` instead of `thread.h`
Source file `mobj_ffa.c` makes use of these two functions, hence include
`kernel/secure_partition.h` header.

### DIFF
--- a/core/arch/arm/mm/mobj_ffa.c
+++ b/core/arch/arm/mm/mobj_ffa.c
@@ -9,6 +9,7 @@
 #include <keep.h>
 #include <kernel/refcount.h>
 #include <kernel/spinlock.h>
+#include <kernel/thread_spmc.h>
 #include <mm/mobj.h>
 #include <sys/queue.h>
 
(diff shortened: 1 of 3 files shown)


core/arch/arm/mm/mobj_ffa.c
```diff
@@ -9,6 +9,7 @@
 #include <keep.h>
 #include <kernel/refcount.h>
 #include <kernel/spinlock.h>
+#include <kernel/secure_partition.h>
```
jenswi-linaro, line 12: Shouldn't this be `thread_spmc.h`?

jenswi-linaro:
Please update the commit message with `kernel/thread_spmc.h` instead. With that fixed:

PR #5071, opened 2021-12-21
GitHub actions: General updates

jbech-linaro:
- Update 'stale' from v3.0.7 to v4.0.1.
- Give 'write' permissions to actions (pull requests and issues).
- Add and set (to true)
    remove-issue-stale-when-updated
    remove-pr-stale-when-updated
  which will automatically remove the 'Stale' label when someone adds a
  new comment to a ticket marked as stale.

Signed-off-by: Joakim Bech <joakim.bech@linaro.org>

### DIFF
--- a/.github/workflows/stales.yml
+++ b/.github/workflows/stales.yml
@@ -3,11 +3,15 @@ on:
   schedule:
   - cron: "15 00 * * *"
 
+permissions:
+  issues: write
+  pull-requests: write
+
 jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/stale@v3.0.7
+    - uses: actions/stale@v4.0.1
 
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
@@ -20,3 +24,5 @@ jobs:
         days-before-stale: 30
         days-before-close: 5
         remove-stale-when-updated: true
+        remove-issue-stale-when-updated: true
+        remove-pr-stale-when-updated: true


.github/workflows/stales.yml
```diff
@@ -20,3 +24,5 @@ jobs:
         days-before-stale: 30
         days-before-close: 5
         remove-stale-when-updated: true
```
vchong, line 26: Based on https://github.com/actions/stale, `remove-stale-when-updated` should cover both `remove-issue-stale-when-updated` and `remove-pr-stale-when-updated`?

jbech-linaro: I was thinking the same at first (since it mentions override), but I believe that the two ones added here also triggers on comments, when the former only triggers if someone push a new commit to the PR. If I recall, we've always have to manually remove the stale label when someone has added a comment to the tickets. Based on my comment here and if you re-read again, do you interpret it the same way as I do?

jbech-linaro: On the other hand, it might simply be that this hasn't worked because we've had older versions of the stale plugin and not having the "permissions" enabled. So, maybe these lines are superflous in the end.

vchong: > Based on my comment here and if you re-read again, do you interpret it the same way as I do?

Yes, I think you're right. `remove-stale-when-updated` only mentioned 'updates', not 'updates/comments'. Sorry for the noise.

jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

PR #5073, opened 2021-12-21
libutils: provide __sprintf_chk implementation

b49020:
While building optee_test CXX test-cases natively on aarch64, OP-TEE
build relies on toolchain provided by buildroot. The buildroot toolchain
is built with flag: -fstack-protector-strong which requires
__sprintf_chk symbol provided by standard glibc. For OP-TEE we use a
customized libc which leads to below error:

CC out/init.o
CC out/os_test.o
CC out/ta_entry.o
CXX out/cxx_tests.o
CC out/user_ta_header.o
CPP out/ta.lds
LD out/5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.elf
/home/sumit/optee_br/build/../toolchains/aarch64/bin/aarch64-linux-ld.bfd: /home/sumit/optee_br/toolchains/aarch64/bin/../lib/gcc/aarch64-buildroot-linux-gnu/10.3.0/../../../../aarch64-buildroot-linux-gnu/lib/../lib64/libstdc++.a(cp-demangle.o): in function d_append_num': cp-demangle.c:(.text+0x830): undefined reference to __sprintf_chk'

Fix this issue by providing __sprintf_chk implementation.

Signed-off-by: Sumit Garg <sumit.garg@linaro.org>

### DIFF
--- a/lib/libutils/isoc/sprintf.c
+++ b/lib/libutils/isoc/sprintf.c
@@ -3,8 +3,10 @@
  * Copyright (c) 2020, Huawei Technologies Co., Ltd
  */
 
+#include <compiler.h>
 #include <printk.h>
 #include <stdio.h>
+#include <stdlib.h>
 
 int sprintf(char *str, const char *fmt, ...)
 {
@@ -14,6 +16,25 @@ int sprintf(char *str, const char *fmt, ...)
 	va_start(ap, fmt);
 	retval = __vsprintf(str, fmt, ap);
 	va_end(ap);
+
+	return retval;
+}
+
+int __sprintf_chk(char *str, int flag __unused, size_t slen,
+		  const char *fmt, ...)
+{
+	int retval;
+	va_list ap;
+
+	if (slen == 0)
+		abort();
+
+	va_start(ap, fmt);
+	retval = __vsnprintf(str, slen, fmt, ap, false);
+	va_end(ap);
+
+	if (retval > 0 && (size_t)retval >= slen)
+		abort();
 
 	return retval;
 }
--- a/lib/libutils/isoc/include/stdio.h
+++ b/lib/libutils/isoc/include/stdio.h
@@ -20,6 +20,8 @@ int snprintf(char *str, size_t size, const char *fmt, ...)
                     __attribute__ ((__format__ (__printf__, 3, 4)));
 int vsnprintf (char *str, size_t size, const char *fmt, va_list ap)
                     __attribute__ ((__format__ (__printf__, 3, 0)));
+int __sprintf_chk(char *str, int flag, size_t slen, const char *fmt, ...)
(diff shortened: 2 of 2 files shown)


lib/libutils/isoc/sprintf.c
```diff
@@ -17,3 +18,16 @@ int sprintf(char *str, const char *fmt, ...)
...
 }
+
+int __sprintf_chk(char *str, int flag __unused, size_t slen __unused,
+		  const char *fmt, ...)
+{
+	int retval;
+	va_list ap;
+
+	va_start(ap, fmt);
+	retval = __vsprintf(str, fmt, ap);
```
jenswi-linaro, line 29: Surely `vsnprintf()` is a better candidate to call here. Then we'd be able to check overflow in `str` which i guess is the most common error.

b49020: Makes sense, I will switch to `vsnprintf()`.

lib/libutils/isoc/sprintf.c
```diff
@@ -17,3 +18,16 @@ int sprintf(char *str, const char *fmt, ...)
...
 }
+
+int __sprintf_chk(char *str, int flag __unused, size_t slen,
+		  const char *fmt, ...)
+{
+	int retval;
+	va_list ap;
+
+	va_start(ap, fmt);
+	retval = __vsnprintf(str, slen, fmt, ap, false);
```
jenswi-linaro, line 29: http://refspecs.linux-foundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/libc---sprintf-chk-1.html

mentions a few checks that should be performed. Checking for slen == 0 or that retval > = slen shouldn't be too hard.

b49020: > Checking for slen == 0 or that retval > = slen shouldn't be too hard.

Although I did see a check for `slen == 0` in glibc implementation [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=debug/sprintf_chk.c;h=c4444f7ff8dc64a1d95f2270b281ceadbc36fa65;hb=HEAD) but I couldn't find check for `retval > = slen`. Should we just return `-1` for these condition checks?

jenswi-linaro: The spec seems to suggest that we should panic instead.

jforissier (commented):
I agree with Jens' comments about adding more checks if we can, but on the principle:

lib/libutils/isoc/sprintf.c
b49020: Added corresponding checks.

b49020:
Tag applied with Jens' comments addressed.

jenswi-linaro:

b49020:
Tag applied.

Issue #4231, opened 2020-11-27
[DOC] Porting guide


maroueneboubakri:
Hi,

Is there a document which helps porting OP-TEE to another architecture  (not ARM) ? 
Thank you :) 

Regards
Maro

jenswi-linaro:
I'm afraid not. Which architecture do you have in mind?

maroueneboubakri:
@jenswi-linaro  Thank you for your quick reply, the arch is RISC-V.

jenswi-linaro:
Do you intend to run it on a dedicated CPU or does RISC-V have some suitable special privilege level?

maroueneboubakri:
We are experimenting both use-cases.  RISC-V has 4 privilege mode. 
To enable TEE, the hardware should has at least a user mode and privileged mode. 
The plan is to run seL4 on the machine mode (to run as a Secure Monitor) and in top of that we could have a Linux and/or OP-TEE running in the supervisor mode. We could have a physical memory speration between Linux and OP-TEE. So, running OP-TEE like on a dedicated CPU is fine for us as a starting point. If there is any  useful documentation that would be highly appreciated.

jenswi-linaro:
The seL4 setup would be similar to the EL3 setup used in Armv8, at least from operating system point of view.
The documentation we have is available at https://optee.readthedocs.io/en/latest/ . Perhaps you've already seen that.
Since we only have OP-TEE on Arm platforms today what goes into `/core` versus `/core/arch/arm` is not very clear. I suspect that quite a bit could be moved from `/core/arch/arm` to `/core`, but in some cases we'd need to copy only portions of files.

By the way, do you think it would be possible to run it all on QEMU, the full package with seL4, OP-TEE and Linux?

jforissier:
> Since we only have OP-TEE on Arm platforms today

Someone mentioned this to me a while ago, but it doesn't seem to be maintained recently: https://01.org/op-tee-ia, https://github.com/op-tee-ia.

maroueneboubakri:
> The seL4 setup would be similar to the EL3 setup used in Armv8, at least from operating system point of view.
> The documentation we have is available at https://optee.readthedocs.io/en/latest/ . Perhaps you've already seen that.
> Since we only have OP-TEE on Arm platforms today what goes into `/core` versus `/core/arch/arm` is not very clear. I suspect that quite a bit could be moved from `/core/arch/arm` to `/core`, but in some cases we'd need to copy only portions of files.
[...]

PR #5080, opened 2021-12-27
optee_os risc-v port

maroueneboubakri:
This is a first PR towards introducing support for RISC-V. This PR builds a dummy tee core file which does nothing. 
Requirements:
- Toolchain: gcc-riscv64-linux-gnu
- Simulator:  [Spike](https://github.com/riscv-software-src/riscv-isa-sim)

To build:
`make ARCH=riscv PLATFORM=spike CROSS_COMPILE64=riscv64-linux-gnu- ta-targets=ta_rv64
`

To run (simulating 2 processors):
`spike -m0x40008000:0x1000,0x80000000:0x40000000 -p2 out/riscv-plat-spike/core/tee.elf
`

Where 0x80000000 is the DRAM base and 0x40008000 is the HTIF base.

Output:
```
I/TC: OP-TEE version: 3.15.0-274-g99ca1618 (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #2 Wed Jan  5 23:11:14 UTC 2022 riscv
I/TC: Primary CPU switching to REE boot
I/TC: Secondary CPU 1 initializing
I/TC: Secondary CPU 1 switching to REE boot
```

Still not sure about the copyright in the header as some files are partially or totally recopied form the other sources.

There is some checkpatch failures which are a bit complicated to handle as well.


core/arch/riscv/plat-spike/platform_config.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#ifndef PLATFORM_CONFIG_H
+#define PLATFORM_CONFIG_H
+
+#define TEE_RAM_START		CFG_TZDRAM_START
```
jenswi-linaro, line 6: TZ in `CFG_TZDRAM_START` means trustzone. I'd prefer a more accurate name.

jenswi-linaro (commented):
Comment for "core: arch: riscv: plat-spike: add platform configuration header file"

core/arch/riscv/kernel/kern.ld.S
```diff
@@ -0,0 +1,485 @@
...
+			 * symbol that may be overridden by platform
+			 * specific code.
+			 */
+		  .text.get_core_pos_mpidr)
+		__identity_map_init_end = .;
+		KEEP(*(.text.init .text.plat_cpu_reset_early \
+		       .text.reset .text.reset_primary .text.unhandled_cpu \
+		       .text.__assert_flat_mapped_range))
+
+#ifdef CFG_WITH_PAGER
```
jenswi-linaro, line 104: Do you see that paging will ever be used for risc-v?

core/arch/riscv/kernel/kern.ld.S
```diff
@@ -0,0 +1,485 @@
...
+	. = ALIGN(SMALL_PAGE_SIZE);
+
+	__flatmap_nex_rw_size = . - __flatmap_nex_rw_start;
+	__flatmap_nex_rw_end = .;
+#endif
+
+	__flatmap_unpg_rw_start = .;
+
+  /*--------------------------------------------------------------------*/
+  /* HTIF, isolated onto separate page                                  */
```
jenswi-linaro, line 209: What is HTIF?

jenswi-linaro (commented):
Comments for "core: arch: riscv: kernel: setup linker for the risc-v core module"

core/arch/riscv/plat-spike/platform_config.h
maroueneboubakri: I would use Trusted Domain, does `CFG_TDDRAM_START` work ?

core/arch/riscv/kernel/kern.ld.S
maroueneboubakri: Spike doesn't yet model a UART but relies on RISC-V Host-Target Interface (HTIF) to perform all I/O. It is a protocol allowing the target to access host to perform console, storage etc. It requires special ELF symbols. This works also for QEMU which will be added later.

core/arch/riscv/kernel/kern.ld.S
maroueneboubakri: Yes, the plan is to be able to run in embedded profile (MMU-less) and application profile.
Also RISC-V provides virtual addressing mechanism with and without MMU support.

jenswi-linaro:
For commits:
- Makefile: allow building for a new target architecture
- riscv: create makefiles and directories tree for riscv
- core: arch: riscv: riscv.mk: setup compiler for the risc-v core module

please apply:

I'd rather replace the `core: arch: riscv` prefix with `core: riscv`, it should be pretty clear what it's about anyway. These prefixes are not a strict rule, they are rather intended as a quick way of understanding roughly which part of the code is touched.

Generally we try to avoid using the word kernel when we mean for instance the tee binary or privileged mode. This is because it's otherwise easily mistaken for the Linux kernel.

I'm not so sure about the "core: core.mk: optionally build core source files" commit. It seems we're delaying the problem, but not breaking it into smaller problems. I'd prefer stubbing as much as possible instead.

ldts:
hi @maroueneboubakri 
just followed the instructions on your PR and I am not sure the console is being initialized

```
jramirez@trex plat-spike (dev-riscv *)$ git diff
diff --git a/core/arch/riscv/plat-spike/main.c b/core/arch/riscv/plat-spike/main.c
index 433c1d7a..ed5b40b7 100644
--- a/core/arch/riscv/plat-spike/main.c
+++ b/core/arch/riscv/plat-spike/main.c
@@ -7,6 +7,9 @@ static struct htif_console_data console_data __nex_bss;
 
 void console_init(void)
 {
+ while(1)
+   ;
+     
        htif_console_init(&console_data);
-   register_serial_console(&console_data.chip);
+ register_serial_console(&console_data.chip);  
 }
```

then build and executed

```
jramirez@trex optee$ ./build.riscv 
  CHK     out/riscv-plat-spike/conf.mk
  UPD     out/riscv-plat-spike/conf.mk
  CHK     out/riscv-plat-spike/include/generated/conf.h
  CC      out/riscv-plat-spike/core/arch/riscv/plat-spike/main.o
core/arch/riscv/plat-spike/main.c: In function ‘console_init’:
core/arch/riscv/plat-spike/main.c:8:6: warning: function might be candidate for attribute ‘noreturn’ [-Wsuggest-attribute=noreturn]
    8 | void console_init(void)
      |      ^~~~~~~~~~~~
  LD      out/riscv-plat-spike/core/all_objs.o
  GEN     out/riscv-plat-spike/core/unpaged_entries.txt
  LD      out/riscv-plat-spike/core/unpaged.o
  GEN     out/riscv-plat-spike/core/text_unpaged.ld.S
  GEN     out/riscv-plat-spike/core/rodata_unpaged.ld.S
  GEN     out/riscv-plat-spike/core/init_entries.txt
  UPD     out/riscv-plat-spike/core/.buildcount
  GEN     out/riscv-plat-spike/core/version.o
  LD      out/riscv-plat-spike/core/init.o
  GEN     out/riscv-plat-spike/core/text_init.ld.S
  GEN     out/riscv-plat-spike/core/rodata_init.ld.S
  CPP     out/riscv-plat-spike/core/kern.ld
  LD      out/riscv-plat-spike/core/tee.elf
  OBJDUMP out/riscv-plat-spike/core/tee.dmp
  GEN     out/riscv-plat-spike/core/tee.symb_sizes
...
core   0: 0x0000000000001000 (0x00000297) auipc   t0, 0x0
core   0: 0x0000000000001004 (0x02028593) addi    a1, t0, 32
core   0: 0x0000000000001008 (0xf1402573) csrr    a0, mhartid
core   0: 0x000000000000100c (0x0182b283) ld      t0, 24(t0)
core   0: 0x0000000000001010 (0x00028067) jr      t0
core   0: 0x00000000bdb00000 (0x0000006f) j       pc + 0x0
```


The output I see is just the same that I get when I dont apply my hack to check the console: ie, either way I never see the message in your comment

_`warning: tohost and fromhost symbols not in ELF; can't communicate with target`_

am I missing something?

maroueneboubakri:
Hi @ldts,

The core being built at this level is a dummy core which does nothing as you can see from the spike trace.
Regarding HITF symbols, I moved the HTIF section to a sperate linker script file inside the plat-spike 

`#include "../kernel/kern.ld.S"

SECTIONS
{
        .htif : {
    PROVIDE( __htif_base = . );
                        KEEP(*(.htif))
        }
        . = ALIGN(0x1000);
}
`
In top of that thread exceptions masking should be implemented (even dummy) since these are used by the spinlock. 

I can push a new commit which does initialize the console and enable logging. But I think this way the PR is will stall. The purpose of this PR is to introduce a new architecture, setup the CI for building.

ldts:
> I can push a new commit which does initialize the console and enable logging. But I think this way the PR is will stall. The purpose of this PR is to introduce a new architecture, setup the CI for building.

ah I see. IMO some "riscv hello world " console output is always nice but sure, if it is going to get in the way and require some temporary work around to be removed later I agree it makes little sense....too much already

maroueneboubakri:
Sure, I add more commits to enable console.

core/arch/riscv/include/kernel/tlb_helpers.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: copyright

core/arch/riscv/kernel/spinlock.c
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
etienne-lms, line 1: copyright

core/arch/riscv/plat-spike/drivers/htif.c
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
etienne-lms, line 1: copyright

core/arch/riscv/plat-spike/drivers/htif.c
```diff
@@ -0,0 +1,62 @@
...
+
+#include <stdint.h>
+#include <compiler.h>
```
etienne-lms, line 4: sort in alphabet order

core/arch/riscv/plat-spike/drivers/htif.c
```diff
@@ -0,0 +1,62 @@
...
+
+static void __maybe_unused tohost_cmd(uintptr_t dev,
+				      uintptr_t cmd, uintptr_t data)
+{
+	while (tohost)
+		asm volatile ("" :: "r" (fromhost));
+
+	*((volatile uint64_t *)&tohost) = (((uint64_t)(dev) << 56) |
+	((uint64_t)(cmd) << 48) |
+	(uint64_t)(data));
```
etienne-lms, line 32:
```c
	*((volatile uint64_t *)&tohost) = ((uint64_t)(dev) << 56) |
					  ((uint64_t)(cmd) << 48) |
					  (uint64_t)(data);

File: core/arch/riscv/plat-spike/drivers/htif.c [hunk 9]
```diff
@@ -0,0 +1,62 @@
...
+#endif
+}
+
+static void htif_console_flush(struct serial_chip *chip __unused)
+{
+}
+
+static const struct serial_ops htif_console_ops = {
+	.flush = htif_console_flush,
+	.getchar = 0,
```
etienne-lms, line 53: can remove this line

core/arch/riscv/plat-spike/drivers/htif.c
```diff
@@ -0,0 +1,62 @@
...
+	.putc = htif_console_putc,
+};
+
```
etienne-lms, line 56: remove this empty line

core/arch/riscv/plat-spike/drivers/htif.h
```diff
@@ -0,0 +1,21 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: copyright

core/arch/riscv/plat-spike/drivers/htif.h
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef __HTIF_H__
+#define __HTIF_H__
+#include <types_ext.h>
+#include <drivers/serial.h>
```
etienne-lms, line 6: sort in alphabetical order

core/arch/riscv/plat-spike/drivers/htif.h
```diff
@@ -0,0 +1,21 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#ifndef __HTIF_H__
```
etienne-lms, line 3: I recommend `__DRIVERS_HTIF_H__`

core/arch/riscv/plat-spike/kern.ld.S
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: (BSD-2-Clause) */
```
etienne-lms, line 1: copyright

core/arch/riscv/plat-spike/main.c
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
etienne-lms, line 1: copyright

core/arch/riscv/include/kernel/secure_partition.h
```diff
@@ -0,0 +1,34 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: copyright

core/arch/riscv/include/kernel/stmm_sp.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: copyright

core/arch/riscv/include/riscv.h
```diff
@@ -0,0 +1,24 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: copyright

core/arch/riscv/include/kernel/thread.h
```diff
@@ -86,19 +82,35 @@ struct thread_param {
...
+ * thread_*_exceptions() functions below.
+ * These definitions are for targets running U/M modes. Consider external
+ * interrupts as foreign interrupts for now.
  */
-uint32_t thread_rpc_cmd(uint32_t cmd, size_t num_params,
-			struct thread_param *params);
+
+#define THREAD_EXCP_FOREIGN_INTR	(MIP_UEIP | MIP_MEIP)
+#define THREAD_EXCP_NATIVE_INTR		(MIP_USIP | MIP_MSIP | MIP_UTIP  \
+					| MIP_MTIP)
```
etienne-lms, line 94: optee_os source files rather places operator before the line break:
```c
#define THREAD_EXCP_NATIVE_INTR		(MIP_USIP | MIP_MSIP | MIP_UTIP | \
					 MIP_MTIP)

File: core/arch/riscv/kernel/thread.c [hunk 20]
```diff
@@ -0,0 +1,89 @@
...
+	((vaddr_t)&(stack)[n] + STACK_CANARY_SIZE / 2)
+#define GET_STACK_TOP_SOFT(stack, n) \
+	(GET_STACK_TOP_HARD(stack, n) + STACK_CHECK_EXTRA)
+#define GET_STACK_BOTTOM(stack, n) ((vaddr_t)&(stack)[n] + \
+	 sizeof(stack[n]) - STACK_CANARY_SIZE / 2)
+
+const void *stack_tmp_export __section(".identity_map.stack_tmp_export") =
+	(void *)(GET_STACK_BOTTOM(stack_tmp, 0) - STACK_TMP_OFFS);
+const uint32_t stack_tmp_stride __section(".identity_map.stack_tmp_stride")
+	= sizeof(stack_tmp[0]);
```
etienne-lms:
```c
const uint32_t stack_tmp_stride __section(".identity_map.stack_tmp_stride") =
	sizeof(stack_tmp[0]);

**Review by etienne-lms — COMMENTED** [review]
* those already acked by @jenswi-linaro
"lib: libutils: isoc: remove unused thread.h from bget_malloc.c",
"core: kernel: move spin_lock_debug.c to core/kernel",
"core: kernel: move spinlock.h to core/include/kernel",
"core: kernel: include: remove multiple blank lines in spinlock.h"
* "lib: libutils: compiler.h: set trailing comment according to target"
* "core: kernel: move abort.h to core/include/kernel"
* "core: mm: move pgt_cache.h to core/include/mm"
* "core: mm: move tee_pager.h to core/include/mm"
* "core: sm: move optee_smc.h to core/include/sm":
(I'm not sure OPTEE SMC IDs can be arch generic, but why not)
* "core: mm: vm.c: remove arm.h and include config.h"
* "core: kernel: move boot.h to core/include/kernel"
* "core: mm: move core_mmu.h to core/include/mm"
(i'm not sure everything is really arch generic, but ok)
*"core: kernel: move thread_defs.h to core/include/kernel"
* "core: riscv: generic_ram_layout.h: ram layout configuration directives"
* "core: mm: move tee_mm.c to core/mm"
* "core: kernel: tee_ta_manager.c: remove arm.h from includes"
* 

* "core: riscv: define risc-v instruction set architecture in encoding.h"

LGTM for the following commits but missing copyright notices in source/header files:
* "

File: core/arch/riscv/plat-spike/conf.mk [hunk 21]
```diff
@@ -0,0 +1,76 @@
+CFG_RV64_core=y
```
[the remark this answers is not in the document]
maroueneboubakri: Yes at some point this will no longer be needed.

jforissier, line 1: `=` is never a valid way of setting a `CFG_` variable. Your two options are `?=` and `$(call force,...)` as explained in https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#configuration-variables and https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#the-force-macro.

core/arch/riscv/plat-spike/platform_config.h
jenswi-linaro: Sounds good.

core/arch/riscv/kernel/kern.ld.S
jenswi-linaro: `CFG_WITH_PAGER` is not needed to support MMU, see also https://optee.readthedocs.io/en/latest/architecture/core.html#pager

Do you really plan to implement this for risc-v?

core/arch/riscv/kernel/kern.ld.S
jenswi-linaro: Who uses these special ELF symbols?

jenswi-linaro:
I'm not so keen on "core: sm: move optee_smc.h to core/include/sm". I'd prefer if riscv defined an independent ABI for communication with the "normal world". That way it's also more clear what's really supported. Note that even on arm `optee_smc.h` is unused in case we're using FF-A (`optee_ffa.h`) instead.

core/arch/riscv/include/encoding.h
```diff
@@ -0,0 +1,4830 @@
+/* SPDX-License-Identifier: BSD-4-Clause-UC */
```
jenswi-linaro: Does this file have an upstream? Is it copied from somewhere?

How about calling this file `riscv.h` instead? It seems to serve more or less the same purpose as the `arm.h` file.

core/arch/riscv/include/encoding.h
```diff
@@ -0,0 +1,4830 @@
...
+#define HSTATUS_HU          0x00000200
+#define HSTATUS_SPVP        0x00000100
+#define HSTATUS_SPV         0x00000080
+#define HSTATUS_GVA         0x00000040
+#define HSTATUS_VSBE        0x00000020
+
+#define USTATUS_UIE         0x00000001
+#define USTATUS_UPIE        0x00000010
+
+#define DCSR_XDEBUGVER      (3U << 30)
```
jenswi-linaro: `SHIFT_U32()` is preferred. Same for all the other places below.

jenswi-linaro (commented):
Comments for the commit "core: riscv: define risc-v instruction set architecture in encoding.h"

maroueneboubakri:
> I'm not so keen on "core: sm: move optee_smc.h to core/include/sm". I'd prefer if riscv defined an independent ABI for communication with the "normal world". That way it's also more clear what's really supported. Note that even on arm `optee_smc.h` is unused in case we're using FF-A (`optee_ffa.h`) instead.

However, many sources outside arm uses it such as core/kernel/notif.c, core/mm/vm.c, core/mm/mobj.c etc. The plan was to stay as close as possible to what it is used today in term of communication with REE (Linux) as stated here [issue#comment ](https://github.com/OP-TEE/optee_os/issues/4231#issuecomment-735649743)

maroueneboubakri:
The whole `encoding.h` is automatically generated by [https://github.com/riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes). We added slight changes to satisfy `checkpatch` and add the license information. This is subject to update each time a new extension or updated in the ISA specifications occurs.

core/arch/riscv/kernel/spinlock.c
```diff
@@ -0,0 +1,29 @@
...
+void __cpu_spin_lock(unsigned int *lock)
+{
+	while (__cpu_spin_trylock(lock))
+		;
+}
+
+void __cpu_spin_unlock(unsigned int *lock)
+{
+	asm volatile ("fence rw,w" ::: "memory");
+	asm volatile ("amoswap.w x0, x0, %0" : "=A" (*lock) :: "memory");
```
jenswi-linaro, line 17: We try to avoid inline assembly instead we code the entire function in an assembly file or use some helper wrapper functions.

In this case I suppose it would be better to just implement it in an assembly file. Same for `__cpu_spin_trylock()` below.

jenswi-linaro (commented):
Comment for the commit "core: riscv: spinlock.c: implement spin-locking primitives"

core/arch/riscv/plat-spike/drivers/htif.c
```diff
@@ -0,0 +1,63 @@
...
+{
+	cpu_spin_lock(&htif_global_lock);
+}
+
+void htif_unlock_global(void)
+{
+	cpu_spin_unlock(&htif_global_lock);
+}
+
+static void __maybe_unused tohost_cmd(uintptr_t dev,
```
jenswi-linaro, line 27: Why are you using the `uintptr_t` type for all the arguments? None of the arguments carry any address.

core/arch/riscv/plat-spike/drivers/htif.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+static void htif_console_putc(struct serial_chip *chip __unused,
+			      int ch __maybe_unused)
+{
+#if __riscv_xlen == 32
+#warning HTIF is not supported on RV32
+#else
+	htif_lock_global();
+	tohost_cmd(1, 1, ch);
```
jenswi-linaro, line 45: Please use some defines instead of these magic numbers.

jenswi-linaro (commented):
Comments for the commit "core: riscv: plat-spike: console driver based on host-target interface"

jenswi-linaro:
> > I'm not so keen on "core: sm: move optee_smc.h to core/include/sm". I'd prefer if riscv defined an independent ABI for communication with the "normal world". That way it's also more clear what's really supported. Note that even on arm `optee_smc.h` is unused in case we're using FF-A (`optee_ffa.h`) instead.
> 
> However, many sources outside arm uses it such as core/kernel/notif.c, core/mm/vm.c, core/mm/mobj.c etc.

Actually, it's only `core/tee/entry_std.c` also, nothing more. We could try to resolve those dependencies.

> The plan was to stay as close as possible to what it is used today in term of communication with REE (Linux) as stated here [issue#comment ](https://github.com/OP-TEE/optee_os/issues/4231#issuecomment-735649743)

Since then we've gained the FF-A ABI so adding yet another ABI shouldn't be too hard. The kernel driver has also been restructured in a way that it should be quite easy to add one more ABI.

jenswi-linaro:
> The whole `encoding.h` is automatically generated by https://github.com/riscv/riscv-opcodes. We added slight changes to satisfy `checkpatch` and add the license information. This is subject to update each time a new extension or updated in the ISA specifications occurs.

OK, I guess I was confused by the "/* Automatically generated by parse_opcodes.  */" in the middle of the file. Please add something in the commit message or close to the top of the file how it's supposed to be updated. Also it seems the the SPDX-License-Identifier isn't accurate.

core/arch/riscv/kernel/entry.S
```diff
@@ -3,9 +3,165 @@
...
+.equ SEM_CPU_READY, 1
+#endif
+
+	/*
+	 * Setup sp to point to the top of the tmp stack for the current CPU:
+	 * sp is assigned stack_tmp_export + cpu_id * stack_tmp_stride
+	 */
+.macro set_sp
+	/* Unsupported CPU, park it before it breaks something */
+	li   t1, CFG_TEE_CORE_NB_CORE
```
jenswi-linaro: Please separate the instruction and its arguments with a single tab. Please use this style in all assembly files.

jenswi-linaro (commented):
Comment for the commit "core: riscv: entry.S: platform initialisation"

core/arch/riscv/kernel/thread.c
```diff
@@ -103,9 +105,54 @@ short int thread_get_id(void)
...
+struct thread_core_local *__nostackcheck thread_get_core_local(void)
+{
+	/*
+	 * Rely on thread pointer register instead of hart id which is available
+	 * only in machine mode. This allows retrieving the core local whether
+	 * running core in M-Mode or S-Mode.
+	 */
+	struct thread_core_local *l;
+
+	asm volatile("mv %0, tp" : "=&r"(l));
```
jenswi-linaro, line 251: We try to avoid using inline assembly in normal functions. Please wrap this in a static inline function you can put in <riscv.h>. A good name could be `read_tp()` if we chose to follow the same convention as in `<arm.h>` and friends.

jenswi-linaro (commented):
Comment for the commit "core: riscv: thread core local structure and functions"

jenswi-linaro:
For how long do you plan to add new commits to this PR?

maroueneboubakri:
> For how long do you plan to add new commits to this PR?

Until it gets merged :) The first plan was to stop until we have a dummy core, then had been asked to have console output. Would be okay to handle all comment above than merge ?

jenswi-linaro:
>>> For how long do you plan to add new commits to this PR?

>> @jenswi-linaro : Until it gets merged :) 

Haha! Was starting to suspect that. :-)

>> The first plan was to stop until we have a dummy core, then had been asked to have console output. Would be okay to handle all comment above then merge ?

Sure, let's focus on getting what we have now ready to be merged.

jenswi-linaro (commented):
Comment on the commit "core: riscv: thread context and stack management"

core/arch/riscv/kernel/thread.c
```diff
@@ -523,3 +524,231 @@ bool thread_init_stack(uint32_t thread_id, vaddr_t sp)
...
+{
+	panic("unhandled exception");
+}
+
+static void dump_trap_frame(struct thread_trap_frame *frame)
+{
+	IMSG(
+	" x0  0x%016" PRIxPTR " ra  0x%016" PRIxPTR
+	" sp  0x%016" PRIxPTR " gp  0x%016" PRIxPTR
+	"\n",
```
jenswi-linaro, line 602: No need for an explicit newline.

core/arch/riscv/kernel/thread.c
```diff
@@ -523,3 +524,231 @@ bool thread_init_stack(uint32_t thread_id, vaddr_t sp)
...
+
+static void thread_unhandled_trap(struct thread_trap_frame *frame __unused)
+{
+	panic("unhandled exception");
+}
+
+static void dump_trap_frame(struct thread_trap_frame *frame)
+{
+	IMSG(
+	" x0  0x%016" PRIxPTR " ra  0x%016" PRIxPTR
```
jenswi-linaro, line 600: Please use `lx` instead of `PRIxPTR` since all the registers are defined as `unsigned long`.

jenswi-linaro (commented):
Comments for the commit "core: riscv: thread: trap handling and thread initialization"

jenswi-linaro:
I never understood if you see that paging will ever be used on risc-v.
I'm not sure that it would make sense, but I'm open to be convinced otherwise.  :-)

maroueneboubakri:
> I never understood if you see that paging will ever be used on risc-v. I'm not sure that it would make sense, but I'm open to be convinced otherwise. :-)

Since it is straightforward to me, what would be the requirements to support paging in OP-TEE, could you kindly provide a checklist and we can discuss them.

jenswi-linaro:
For paging to make sense you need to have some special memory that you really what to use instead of the infinite amount of DDR otherwise available. If you use the pager with the special memory you still need to use DDR as backing store. And not least of all you must not forget all the hard work that will be needed to get this to work and to keep it working.
Perhaps we can start with why paging on risc-v should be supported.

maroueneboubakri:
> For paging to make sense you need to have some special memory that you really what to use instead of the infinite amount of DDR otherwise available. If you use the pager with the special memory you still need to use DDR as backing store. And not least of all you must not forget all the hard work that will be needed to get this to work and to keep it working. Perhaps we can start with why paging on risc-v should be supported.

If we look at Spike which is the main platform to be used for development (Qemu also has spike machine) there is the possibility to simulate many memories. You can have an additional memory mapped at an address of your choice which could simulate the SRAM in addition to the DDR memory mapped at 0x80000000. Moreover, It supports MMU. I believe that's enough to test paging on OP-TEE. But the goal at this level is to sort-out the common codes in thread.c and core_mmu.c without being intrusive in removing special parts related to pager and vfp. That's why I removed some commits, I want instead of recopying the existing codes from arm to riscv, put the common code outside the arch folder. But this needs more commits and a motivation/reason.

To be honest, at this level I'm not sure if it worth working more on this, the steps how to proceed are not clear to me. I mean to have a port we need many commits, in the other hand many commits can't be reviewed. One commit with big change is also intrusive and can't be reviewed. We have the first step where we can build a core which outputs to the console but it will stall. Without these commits can't proceed further with a new PR. 

Maybe OP-TEE could be the defacto TEE stack for ARM but not sure if there is an intention to have it also for RISC-V as in the other hand Trusty seems to run already on RISC-V since LK already supports RISC-V (not sure if fully) but yes.

I suggest to open a new PR to sort out common codes for thread and mm before going to RISC-V, but want to have this PR merged before to make sure that there is an interest to have it.

jenswi-linaro:
I was more looking for a yes or no answer, or to be honest only no. :-)

If you say yes to CFG_WITH_PAGER=y on Risc-V you need a stronger argument than that it's possible to emulate some special memory. For instance, you have a platform in mind which must use paging or the TEE will never fit into that limited memory. If it's not clear to you what `CFG_WITH_PAGER=y`does, then you don't want it.

The reason I'm insisting on this is that you have a few Risc-V specific files with `#ifdef CFG_WITH_PAGER` and that's dead code which will remain so for a foreseeable future.

maroueneboubakri:
> 

Yes maybe that's easier to suppress support for pager.

jenswi-linaro:
> Yes maybe that's easier to suppress support for pager.

Agreed. Please all the `#ifdef CFG_WITH_PAGER` in the Risc-V specific files in the PR.

core/arch/riscv/riscv.mk
```diff
@@ -0,0 +1,177 @@
...
+rv64-platform-cppflags += -mcmodel=medany -march=rv64imafd -mabi=lp64d
+rv64-platform-cppflags += -Wno-missing-include-dirs
+
+rv64-platform-cppflags += -DRV64=1 -D__LP64__=1
+rv32-platform-cppflags += -DRV32=1 -D__ILP32__=1
+
+platform-cflags-generic ?= -ffunction-sections -fdata-sections -pipe
+platform-aflags-generic ?= -pipe
+
+rv64-platform-cflags-generic := -mstrict-align $(call cc-option,-mno-outline-atomics,)
```
jforissier, line 50: `-mno-outline-atomics` makes no sense on RISC-V (it is related to the ARMv8.1-A instruction set).

jforissier:
- For commits:
"lib: libutils: isoc: remove unused thread.h from bget_malloc.c"
"core: kernel: move spin_lock_debug.c to core/kernel"
"core: kernel: move spinlock.h to core/include/kernel"

- For commits:
"core: kernel: include: remove multiple blank lines in spinlock.h"
"Makefile: allow building for a new target architecture"

- Comments for commit "riscv: create makefiles and directories tree for riscv"
s/for riscv/for RISC-V/ in the commit subject
  Why introduce empty files at this stage? Why not add them with the proper content
as needed? This commit doesn't allow a clean build:
  ```bash
  $ make -j10 ARCH=riscv
  ta/mk/build-user-ta.mk:65: ta/arch/riscv/link.mk: No such file or directory
  make: *** No rule to make target 'ta/arch/riscv/link.mk'.  Stop.
  ```

- For commit "core: riscv: plat-spike: default configuration for spike platform"
s/for spike/for Spike/ and same in the description. Spike should always be written
with an uppercase S except in the subject prefix which are always lowercase
(plat-spike:, spike:).
  With that fixed:

jforissier (commented):
For commit "core: riscv: riscv.mk: setup compiler for the risc-v core module"

s/the risc-v/the RISC-V/ in the subject, plus a couple of comments below, then:

core/arch/riscv/kernel/entry.S
```diff
@@ -0,0 +1,11 @@
...
+/*
+ * Copyright 2022 NXP
+ */
+
+#include <platform_config.h>
+#include <asm.S>
+
+FUNC _start , :
+	j	.
+END_FUNC _start
```
jforissier, line 124: I don't think it is a good idea to introduce code that is completely changed in the same series. This whole commit should be omitted.

jforissier (commented):
Reviewing commit "core: riscv: kernel: build a dummy tee core".

core/arch/riscv/include/kernel/thread.h
```diff
@@ -0,0 +1,180 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
jforissier, line 1: This files duplicates quite a few things from `core/arch/arm/include/kernel/thread.h`. Isn't there a way to use a common file for non arch-specific things?

jforissier (commented):
Reviewing "core: riscv: include: add thread.h header file".

jforissier:
For commits:
  "core: kernel: move abort.h to core/include/kernel"
  "core: mm: move pgt_cache.h to core/include/mm"
  "core: mm: move tee_pager.h to core/include/mm"
  "core: mm: vm.c: remove arm.h and include config.h"

Acked-by: Jerome Forissier <jerome@forissier.org>

I propose you move these commits as well as the first five in the series to a new PR, where they can be merged immediately.
```
lib: libutils: isoc: remove unused thread.h from bget_malloc.c
core: kernel: move spin_lock_debug.c to core/kernel
core: kernel: move spinlock.h to core/include/kernel
core: kernel: include: remove multiple blank lines in spinlock.h
Makefile: allow building for a new target architecture
```

maroueneboubakri:
> Reviewing "core: riscv: include: add thread.h header file".

There is a lot of code form `thread.h` and `thread.c` to be duplicated. The same applies for `core_mmu.h` and `core_mmu.c`. There is 2 reasons why makes making common code move to a sperate files a bit complicated, **and needs to be done by you team to make sure we don't break things.**
 
1) Many objects are not handled as opaque. For example, `static_memory_map` is used by functions which are generic but also with functions which are architecture-specific. Functions `core_mmu_add|remove_mapping()` access `static_memory_map` and these should be implemented according to the architecture. The way how to access xlat table is different, therefore, `core_mmu_find_table()` and `core_mmu_get|set_entry()` can't be used/implemented for RISC-V and is much easier to re-implement  `core_mmu_add|remove_mapping()`.

2) Flags such as `CFG_WITH_PAGER `flag which will make significant change, and woul be the main cause of code duplciation, for me (unless I misunderstand it), this is a feature and does not depend on the architecture, let's take `thread.c` and remove any code which depends on `CFG_WITH_PAGER `, the result will be a duplicate copy of `thread.c`  which will be hard to maintain the feature.  [This is needs to be decided ](https://github.com/OP-TEE/optee_os/pull/5080#issuecomment-1021991864).

Also, calls to `thread_lazy_(restore|save)_ns_vfp()`  are also added to many functions which can be re-used. I'm not sure if these can be replaced with generic names instead.

Regards
Maro

jenswi-linaro:
Here's PR aiming to take care of much of the threads.c problem https://github.com/OP-TEE/optee_os/pull/5157

jenswi-linaro:
Here's another PR aiming at taking care of the core_mmu.c problem https://github.com/OP-TEE/optee_os/pull/5164

fanghuaqi:
Hi @maroueneboubakri , I tried to build your dev-riscv branch, it seems that it is not working, is there any steps to reproduce the result you present in this [video](https://www.youtube.com/watch?v=uR1YwN47yY8), I would like to take a try with it. Thanks in advance.

maroueneboubakri:
> Hi @maroueneboubakri , I tried to build your dev-riscv branch, it seems that it is not working, is there any steps to reproduce the result you present in this [video](https://www.youtube.com/watch?v=uR1YwN47yY8), I would like to take a try with it. Thanks in advance.

@fanghuaqi the work is ongoing to upstream the port, yet there is no public reference for the full port to get and try it.

fanghuaqi:
> > Hi @maroueneboubakri , I tried to build your dev-riscv branch, it seems that it is not working, is there any steps to reproduce the result you present in this [video](https://www.youtube.com/watch?v=uR1YwN47yY8), I would like to take a try with it. Thanks in advance.
> 
> @fanghuaqi the work is ongoing to upstream the port, yet there is no public reference for the full port to get and try it.

Ok, thank you. If there is any workable commit or guide, could you kindly mention me in this thread.

core/arch/riscv/kernel/entry.S
```diff
@@ -176,7 +170,7 @@ END_FUNC reset_secondary
 DECLARE_KEEP_PAGER reset_secondary
 
 LOCAL_FUNC unhandled_cpu , :
-	j	cpu_idle
+	wfi
```
jforissier, line 173: This patch ("core: riscv: implement cpu_idle() function ") is weird:
- `cpu_idle()` already exists (in assembly), the patch it essentially moving it to C (which is OK). The subject needs to be modified accordingly.
- The description mentions "a panicked CPU", but this is `unhandled_cpu()` => "unhandled CPU" would sounds better (what is a panicked CPU at this point?)
- You're replacing `j cpu_idle` with `wfi` which is OK but should be a separate patch, and then explain why `cpu_idle()` is not as good (I suspect the reason is because `mb()` is useless)

core/arch/riscv/kernel/idle.c
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2022 NXP
+ */
+
+#include <riscv.h>
+#include <kernel/panic.h>
```
jforissier, line 7: `<kernel/panic.h>` not needed?

jforissier (commented):
Some comments on "core: riscv: implement cpu_idle() function".

core/arch/riscv/include/kernel/thread.h
```diff
@@ -0,0 +1,418 @@
...
+/*
+ * Copyright (c) 2014, STMicroelectronics International N.V.
+ * Copyright (c) 2016-2017, Linaro Limited
+ * Copyright (c) 2020-2021, Arm Limited
+ * Copyright 2022 NXP
+ */
+
+#ifndef KERNEL_THREAD_H
```
jforissier, line 9: If I understand the arch model we're putting in place, I think this file should be called `core/arch/riscv/include/kernel/thread_arch.h` and it should include `<kernel/thread.h>`. Then you can get rid of many definitions below.

jforissier (commented):
Comment on "core: riscv: thread: thread management routines for RISC-V".

jenswi-linaro:
I'd rather not have "core: sm: move optee_smc.h to core/include/sm". You can build the interface on top of optee_msg.h instead, that's what we're doing for FF-A in optee_ffa.h.

maroueneboubakri:
> 

That's to satisfy:
```
core/kernel/notif.c:#include <sm/optee_smc.h>
core/mm/mobj.c:#include <sm/optee_smc.h>
core/mm/vm.c:#include <sm/optee_smc.h>
core/tee/entry_std.c:#include <sm/optee_smc.h>
```
Should these be includes be replaced with optee_msg.h ?

jenswi-linaro:
Regarding optee_smc.h usage outside core/arch/arm, https://github.com/OP-TEE/optee_os/pull/5172

maroueneboubakri:
> Regarding optee_smc.h usage outside core/arch/arm, #5172

Commit "core: sm: move optee_smc.h to core/include/sm" dropped.

maroueneboubakri:
@jenswi-linaro the latest 2 commits ,"ldelf: add support for RISC-V" and "core: riscv: add support building user TAs for RISC-V" demonstrate a need to refactor `ldelf`, `libutee` and `ta` such that code can be re-used by another architecture implementation.  Code includes, `ldelf_loader.c`, `utee_syscalls_asm.S`, `user_ta_entry.c`, `tcb.c` and `gprof`.
These commit allows building TAs for the current PR but the goal is experimental as described above, rather than providing functional code.

jenswi-linaro:
I think it's time to start merging risc-v commits. To make it easier to progress I propose that you select a handful of commits from this PR and create a new PR with those. Once that's merged, repeat.

Stuff that we merge doesn't have to be immediately testable from risc-v point of, but they must not break anything that works. Once we have something that compiles we can look at CI builds for risc-v.

jforissier:
@jenswi-linaro +1

> Stuff that we merge doesn't have to be immediately testable from risc-v point of, but they must not break anything that works.

Yes, and @maroueneboubakri please insert usage instructions where it makes sense in the commit descriptions, so that what can be done is documented for everyone else to try. For example in the first commit that allows to compile something, explain how to build (`make ARCH=riscv PLATFORM=spike` I guess?), then when something can be run explain how to run the spike tool, etc.

> Once we have something that compiles we can look at CI builds for risc-v.

I can take care of that, or at least help if it involves preinstalling a cross compiler in the CI Docker image for example.

maroueneboubakri:
Work is still ongoing

jenswi-linaro:
How about start merging reviewed commits?

jforissier:
@jenswi-linaro yes, I can merge up to (and including) "core: riscv: add stmm and sp prototypes". I will add my R-b to "core: include: io.h: define io_read64() and io_write64() helpers" which has no review tag but looks fine to me.

jenswi-linaro:
For commit "core: include: io.h: define io_read64() and io_write64() helpers" please apply:

I was actually hoping that @maroueneboubakri would do as you suggested above:
> Yes, and @maroueneboubakri please insert usage instructions where it makes sense in the commit descriptions, so that what can be done is documented for everyone else to try. For example in the first commit that allows to compile something, explain how to build (make ARCH=riscv PLATFORM=spike I guess?), then when something can be run explain how to run the spike tool, etc.

jforissier:
> I was actually hoping that @maroueneboubakri would do as you suggested above:
> 
> > Yes, and @maroueneboubakri please insert usage instructions where it makes sense in the commit descriptions, so that what can be done is documented for everyone else to try. For example in the first commit that allows to compile something, explain how to build (make ARCH=riscv PLATFORM=spike I guess?), then when something can be run explain how to run the spike tool, etc.

True, I had forgotten about that. I tried building and noticed that `CROSS_COMPILE{32,64}` don't have the proper default values (they are `arm-...` and `aarch64-...`), they should be set based on `$(ARCH)`.

core/arch/riscv/kernel/spinlock.S
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
jforissier, line 1: Please use C style comments `/* SPDX-License-Identifier: BSD-2-Clause */` in assembly files.

lubing521:
> This is a first PR towards introducing support for RISC-V. This PR builds a dummy tee core file which does nothing. Requirements:
> 
> * Toolchain: gcc-riscv64-linux-gnu
> * Simulator:  [Spike](https://github.com/riscv-software-src/riscv-isa-sim)
> 
> To build: `make ARCH=riscv PLATFORM=spike CROSS_COMPILE64=riscv64-linux-gnu- ta-targets=ta_rv64 `
> 
> To run (simulating 2 processors): `spike -m0x40008000:0x1000,0x80000000:0x40000000 -p2 out/riscv-plat-spike/core/tee.elf `
> 
> Where 0x80000000 is the DRAM base and 0x40008000 is the HTIF base.
> 
> Output:
> 
> ```
> I/TC: OP-TEE version: 3.15.0-274-g99ca1618 (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #2 Wed Jan  5 23:11:14 UTC 2022 riscv
> I/TC: WARNING: This OP-TEE configuration might be insecure!
> I/TC: WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html
> I/TC: Primary CPU initializing
> I/TC: Primary CPU switching to REE boot
> I/TC: Secondary CPU 1 initializing
> I/TC: Secondary CPU 1 switching to REE boot
> ```
> 
> Still not sure about the copyright in the header as some files are partially or totally recopied form the other sources.
> 
> There is some checkpatch failures which are a bit complicated to handle as well.
> 
> Signed-off-by: Marouene Boubakri [marouene.boubakri@nxp.com](mailto:marouene.boubakri@nxp.com)

Hi maroueneboubakri，
    thanks your patch，it is grate job！ there may be a bug on memory manager。I  clone your project from
http://github.com/maroueneboubakri/optee_os and checkout on branch dev-riscv，The core built ok，but not
work well。my log as below：
···
lubing@lubing-VirtualBox:~/work/riscv/optee_os$ spike -d  -m0x40008000:0x1000,0x80000000:0x50000000 -p2 out/riscv-plat-spike/core/tee.elf
: rs
D/TC:?   add_phys_mem:554 HTIF_BASE type IO_NSEC 0x40008000 size 0x00200000
D/TC:?   add_phys_mem:554 TEE_SHMEM_START type NSEC_SHM 0x7fe00000 size 0x00200000
D/TC:?   add_phys_mem:554 TA_RAM_START type TA_RAM 0xbdf00000 size 0x00b00000
D/TC:?   add_phys_mem:554 VCORE_UNPG_RW_PA type TEE_RAM_RW 0xbdb16000 size 0x003ea000
D/TC:?   add_phys_mem:554 VCORE_UNPG_RX_PA type TEE_RAM_RX 0xbdb00000 size 0x00016000
D/TC:?   add_va_space:594 type RES_VASPACE size 0x00a00000
D/TC:?   add_va_space:594 type SHM_VASPACE size 0x02000000
D/TC:?   dump_mmap_table:707 type NSEC_SHM     va 0xb9a00000..0xb9bfffff pa 0x7fe00000..0x7fffffff size 0x00200000 (pgdir)
D/TC:?   dump_mmap_table:707 type SHM_VASPACE  va 0xb9c00000..0xbbbfffff pa 0x00000000..0x01ffffff size 0x02000000 (pgdir)
D/TC:?   dump_mmap_table:707 type RES_VASPACE  va 0xbbe00000..0xbc7fffff pa 0x00000000..0x009fffff size 0x00a00000 (pgdir)
D/TC:?   dump_mmap_table:707 type TA_RAM       va 0xbcb00000..0xbd5fffff pa 0xbdf00000..0xbe9fffff size 0x00b00000 (smallpg)
D/TC:?   dump_mmap_table:707 type IO_NSEC      va 0xbd800000..0xbd9fffff pa 0x40008000..0x40207fff size 0x00200000 (smallpg)
D/TC:?   dump_mmap_table:707 type TEE_RAM_RX   va 0xbdb00000..0xbdb15fff pa 0xbdb00000..0xbdb15fff size 0x00016000 (smallpg)
D/TC:?   dump_mmap_table:707 type TEE_RAM_RW   va 0xbdb16000..0xbdefffff pa 0xbdb16000..0xbdefffff size 0x003ea000 (smallpg)
D/TC:?   core_mmu_find_table:326 PTE invalid 0x0 idx 0
E/TC:0   Panic 'can't find table for mapping' at ?:0 
···
I found that pte  is zero when core_mmu_find_table function run。 
···
diff --git a/core/arch/riscv/mm/core_mmu_arch.c b/core/arch/riscv/mm/core_mmu_arch.c
index 6deae22c..a4f20d41 100644
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -322,6 +322,10 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
                        break;
                idx = core_mmu_pgt_idx(va, level);
                pte = &(pgt->entries[idx]);
+               if(!pte) {
+                       DMSG("PTE invalid %p idx %ld",pte,idx);
+                       return false;
+               }
                if(core_mmu_entry_is_branch(pte))
                {
                        pgt = (pgt_t*)(core_mmu_pte_ppn(pte) << RISCV_PGSHIFT);

···

jforissier:
@maroueneboubakri could you please move the approved commits (those with R-b/A-b) to a new PR so that we can make some progress on this?

maroueneboubakri:
work in progress.

jenswi-linaro:
> @maroueneboubakri could you please move the approved commits (those with R-b/A-b) to a new PR so that we can make some progress on this?

What are we waiting for with these commits? It's much easier to work together if we merge what we have agreed on.

jforissier:
> > @maroueneboubakri could you please move the approved commits (those with R-b/A-b) to a new PR so that we can make some progress on this?
> 
> What are we waiting for with these commits? It's much easier to work together if we merge what we have agreed on.

I have gone through the list of commits and cherry-picked the ones that have been reviewed with no pending comments. They are merged to master now. @maroueneboubakri please rebase further work onto master. Thanks!

sichunqin:
Is there any plan to make the risc-v porting completed and available to the public?

PR #5081, opened 2021-12-30
qemu: Secure bootlog capture and display support for QEMUv7

DevendraDevadiga:
1MB of secure memory will be used to store the secure log and can be fetch and display from TA.
In Linux "dmesg" is the command to get the bootlog. Similarly "tmesg" - TEE Message command is provided.

In QEMU secure memory will be increased to another 1MB as below (hw/arm/virt.c):
[VIRT_SECURE_MEM] =         { 0x0e000000, 0x01100000 },
 
The trusted application for "tmesg" is provided here:
https://github.com/DevendraDevadiga/optee-secure-bootlog.git

User need to enable below macro in QEMU conf.mk file:
CFG_BOOT_LOG_PTA ?= y

Signed-off-by: Devendra Devadiga <devendradevadiga01@gmail.com>


jenswi-linaro:
I'm not so sure about this. To start with you're depending on a change in the emulator.
Then I don't see how keeping secure world debug logs in secure memory is so important that we should sacrifice so much memory. What's wrong with logging and be done with it?

DevendraDevadiga:
Okay. I thought if debug console is not always ON and some TA is in running state and later user/developer wants to observe the log, this will be helpfull.

Yes I tested by doing changes in emulator. I thought to submit patch for that also if these changes are okay.

Mean time I thought to reduce 1MB  in CFG_TZDRAM_SIZE also for test as below:

CFG_TZDRAM_START ?= 0x0e100000
-CFG_TZDRAM_SIZE  ?= 0x00f00000
Change size as below:
+ CFG_TZDRAM_SIZE  ?= 0xE00000

And then use that 1MB for secure log.
CFG_TEE_BOOT_LOG_START ?= (CFG_TZDRAM_START + CFG_TZDRAM_SIZE)
CFG_TEE_BOOT_LOG_SIZE ?= 0x100000U

Anyway thanks for your suggestion.
If it is not useful I will close the pull request.

Thank you
Devendra

PR #5089, opened 2022-01-06
crypto: drivers: se050: ecc shared secret

ldts:
Allow clients to inject their own keypairs to derive the secret - the
previous implementation only allowed for secure element NVM based
keypairs to be used.

By default, the secure element does not store all the possible EC
curves in its internal memory; however attempting to inject a keypair
when the curve is not in the secure element would cause the injection
to fail.

This commit addresses that situation by generating those curves in the
SE whenever they are not available.

Tested with:
  TEE_ALG_ECDH_P192, 
  TEE_ALG_ECDH_P224, 
  TEE_ALG_ECDH_P256
  TEE_ALG_ECDH_P384 
   TEE_ALG_ECDH_P521 

**fixes xtest 4009**

### DIFF
--- a/core/drivers/crypto/se050/core/ecc.c
+++ b/core/drivers/crypto/se050/core/ecc.c
@@ -212,6 +212,14 @@ static TEE_Result se050_inject_public_key(sss_se05x_object_t *k_object,
 	if (st != kStatus_SSS_Success)
 		return TEE_ERROR_BAD_PARAMETERS;
 
+	/*
+	 * This function might return an error if the curve already
+	 * exists in the secure element. An actual error creating the
+	 * curve will be caught when attempting to set the key.
+	 */
+	sss_se05x_key_store_create_curve(&se050_session->s_ctx,
+					 curve_tee2se050(key->curve));
+
 	st = se050_get_oid(&oid);
 	if (st != kStatus_SSS_Success)
 		return TEE_ERROR_GENERIC;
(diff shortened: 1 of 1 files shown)


core/drivers/crypto/se050/core/ecc.c
```diff
@@ -212,6 +212,9 @@ static TEE_Result se050_inject_public_key(sss_se05x_object_t *k_object,
 	if (st != kStatus_SSS_Success)
 		return TEE_ERROR_BAD_PARAMETERS;
 
+	(void)sss_se05x_key_store_create_curve(&se050_session->s_ctx,
```
jenswi-linaro, line 215: What happens if this function fails?

ldts: if the curve doesnt exist in the SE, the next function will fail to inject the key.
if the curve already exist in the SE and this function for whatever reason fails, it all will work as it should.

Apparently some versions of the API used to return an error if the curve was already in the SE (which would be a pass for what we need here) hence why I coded it this way. Not a big deal I dont think.

jenswi-linaro: Makes sense. Please drop the `(void)` and add a comment along what you just said.

ldts: @jenswi-linaro pushed

jenswi-linaro (commented):

ldts:
@jenswi-linaro added tag

PR #5090, opened 2022-01-07
[RFC] Add initial TPM2 driver

vchong:
@apalos FYI


core/include/tpm2.h
```diff
@@ -0,0 +1,138 @@
...
+
+#include <stdint.h>
+#include <types_ext.h>
+#include <util.h>
+
+#define TPM2_REG_SIZE 0x5000
+
+#define TPM2_HDR_LEN 10
+
+#define TPM2_ACCESS(l) SHIFT_U32(l, 12)
```
jenswi-linaro, line 18: Please add a tab between `TPM2_ACCESS(l)` and `SHIFT_U32(l, 12)`, same below.

core/include/tpm2.h
```diff
@@ -0,0 +1,138 @@
...
+
+enum tpm2_int_flags {
+	TPM2_INT_DATA_AVAIL_INT = BIT(0),
+	TPM2_INT_STS_VALID_INT = BIT(1),
+	TPM2_INT_LOCALITY_CHANGE_INT = BIT(2),
+	TPM2_INT_CMD_READY_INT = BIT(7),
+	TPM2_GLOBAL_INT_ENABLE = BIT(31),
+};
+
+enum tpm2_result {
```
jenswi-linaro, line 33: Does these values follow some standard? If so please add a reference.

core/arch/arm/plat-vexpress/tpm2_platform.h
```diff
@@ -0,0 +1,16 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Linaro Limited
+ *
```
jenswi-linaro, line 4: This line isn't needed, please remove it here and in the other files in the patch.

core/include/tpm2.h
```diff
@@ -0,0 +1,138 @@
...
+};
+
+enum {
+	TPM2_CMD_COUNT_OFFSET = BIT(1),
+	TPM2_CMD_ORDINAL_OFFSET = SHIFT_U32(3, 1),
+	TPM2_MAX_BUF_SIZE = 1260,
+};
+
+struct tpm2_chip {
+	struct tpm2_ops *ops;
```
jenswi-linaro, line 106: const struct.

jenswi-linaro (commented):
Comments for the commit "tpm2: add header files"

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+	return buf[0] << 24 | buf[1] << 16 | buf[2] << 8 | buf[3];
+}
+
+enum tpm2_result tpm2_init(struct tpm2_chip *chip)
+{
+	enum tpm2_result ret = TPM2_OK;
+	struct tpm2_ops *ops = chip->ops;
+	uint32_t flags = 0;
+
+	if (!tpm2_check_ops(ops)) {
```
jenswi-linaro, line 173: I guess `assert(tpm2_check_ops(ops));` should be good enough here.

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+		t_cnt += TPM2_TIMEOUT_MS;
+	} while (t_cnt < chip->timeout_a);
+
+	EMSG("Timeout t_cnt = %" PRIu32 " ms", t_cnt);
+	return TPM2_ERR_TIMEOUT;
+}
+
+uint32_t tpm2_convert2be(uint8_t *buf)
+{
+	return buf[0] << 24 | buf[1] << 16 | buf[2] << 8 | buf[3];
```
jenswi-linaro, line 164: Please use the `SHIFT_U32()` macro instead.

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+
+	chip->ops->rx8(chip, TPM2_RID(chip->locality), 1, &chip->rid);
+	chip->ops->rx32(chip, TPM2_DID_VID(chip->locality), &chip->vend_dev);
+
+	return tpm2_free_locality(chip);
+}
+
+enum tpm2_result tpm2_end(struct tpm2_chip *chip)
+{
+	tpm2_get_ready(chip);
```
jenswi-linaro, line 205: What happens if these two functions fails?

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+	if (!(status & TPM2_STS_COMMAND_READY)) {
+		ret = tpm2_get_ready(chip);
+		if (ret) {
+			EMSG("Previous cmd cancel failed");
+			goto free_locality;
+		}
+		ret = tpm2_wait_for_status(chip, TPM2_STS_COMMAND_READY,
+					   chip->timeout_b, &status);
+		if (ret) {
+			EMSG("Module not ready\n");
```
jenswi-linaro, line 268: `\n` at the end isn't needed.

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+	size += tmp_size;
+	if (ret || size < expected) {
+		EMSG("Unable to rx remaining data");
+		size = TPM2_ERR_IO;
+		goto out;
+	}
+
+out:
+	tpm2_get_ready(chip);
+	/* obtained from tpm2_tx() */
```
jenswi-linaro, line 386: ??

jenswi-linaro (commented):
Comments for the commit "tpm2: add core driver functionalities"

core/drivers/tpm2_mmio.c
```diff
@@ -0,0 +1,91 @@
...
+
+	return TPM2_OK;
+}
+
+static enum tpm2_result tpm2_mmio_rx8(struct tpm2_chip *chip, uint32_t adr,
+				      uint16_t len, uint8_t *buf)
+{
+	vaddr_t base = tpm2_chip_to_base(chip);
+
+	while (len--)
```
jenswi-linaro, line 44: I'd prefer something less error prone, like:
``` C
for (n = 0; n < len; n++)
         buf[n] = io_read8(base + adr);
```

By the way, you're doing byte reads from the same device mapped address multiple times. What happens if the producer can't keep up? Are the reads just stalling until there's something to read or does something else happens?

core/drivers/tpm2_mmio.c
```diff
@@ -0,0 +1,91 @@
...
+		io_write8(base + adr, *buf++);
+
+	return TPM2_OK;
+}
+
+static struct tpm2_ops tpm2_mmio_ops = {
```
jenswi-linaro, line 61: static const struct

jenswi-linaro (commented):
Comments for the commit "tpm2: add MMIO platform driver"

core/drivers/sub.mk
```diff
@@ -5,6 +5,8 @@ srcs-$(CFG_TZC380) += tzc380.c
 srcs-$(CFG_GIC) += gic.c
 srcs-$(CFG_PL061) += pl061_gpio.c
 srcs-$(CFG_PL022) += pl022_spi.c
+srcs-$(CFG_TPM2) += tpm2.c
```
jenswi-linaro, line 8: This should go into the commit "tpm2: add core driver functionalities"

core/drivers/sub.mk
```diff
@@ -5,6 +5,8 @@ srcs-$(CFG_TZC380) += tzc380.c
...
+srcs-$(CFG_TPM2) += tpm2_mmio.c
```
jenswi-linaro, line 9: This should go into the commit "tpm2: add MMIO platform driver"

jenswi-linaro (commented):
Comments for the commit "tpm2: enable TPM2 driver build"

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+	struct tpm2_ops *ops = chip->ops;
+	uint32_t flags = 0;
+
+	if (!tpm2_check_ops(ops)) {
+		EMSG("No rx tx functions defined");
+		return TPM2_ERR_INIT;
+	}
+
+	/*
+	 * chip->timeout_a start out as 0 in tpm2_get_locality()
```
etienne-lms, line 179: is it intentional to bypass timeout detection?

core/include/tpm2.h
```diff
@@ -0,0 +1,138 @@
...
+	int32_t is_open;
```
etienne-lms, line 107: bool?

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+	uint32_t burstcnt = 0;
+	uint32_t sent = 0;
+	uint32_t wr_size = 0;
+	uint8_t data = TPM2_STS_GO;
+	uint8_t status = 0;
+
+	if (!chip)
+		return TPM2_ERR_GENERIC;
+
+	/* free in tpm2_rx() */
```
etienne-lms, line 250: comment?

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+
+	return true;
+}
+
+static bool tpm2_check_locality(struct tpm2_chip *chip, int loc)
+{
+	struct tpm2_ops *ops = chip->ops;
+	uint8_t locality = 0;
+
+	ops->rx8(chip, TPM2_ACCESS(loc), 1, &locality);
```
etienne-lms, line 24: It would be nice to have helper functions like so this call implement `tpm2_rx8(chip, args....);`

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+	if (status & TPM2_STS_DATA_EXPECT) {
+		ret = TPM2_ERR_IO;
+		goto free_locality;
+	}
+
+	ret = ops->tx8(chip, TPM2_STS(chip->locality), 1, &data);
+	if (ret)
+		goto free_locality;
+
+	return sent;
```
etienne-lms, line 313: `return TPM2_OK`

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+
+enum tpm2_result tpm2_rx(struct tpm2_chip *chip, uint8_t *buf, uint32_t len)
+{
+	enum tpm2_result ret = TPM2_OK;
+	uint32_t expected = 0;
+	uint32_t size = 0;
+	uint32_t tmp_size = 0;
+
+	if (len < TPM2_HDR_LEN)
+		return TPM2_ERR_ARG_LIST_TOO_LONG;
```
etienne-lms, line 359: `TPM2_ERR_INVALID_ARG`?

core/drivers/tpm2.c
```diff
@@ -0,0 +1,391 @@
...
+		size = TPM2_ERR_IO;
+		goto out;
+	}
+
+out:
+	tpm2_get_ready(chip);
+	/* obtained from tpm2_tx() */
+	tpm2_free_locality(chip);
+
+	return size;
```
etienne-lms, line 389: should return `ret` which shall be loaded with the expected return value.

etienne-lms (commented):
for commit "tpm2: add core driver functionalities"

core/drivers/tpm2_mmio.c
```diff
@@ -0,0 +1,91 @@
...
+	.rx32 = tpm2_mmio_rx32,
+	.tx32 = tpm2_mmio_tx32,
+	.rx8 = tpm2_mmio_rx8,
+	.tx8 = tpm2_mmio_tx8,
+};
+DECLARE_KEEP_PAGER(tpm2_mmio_ops);
+
+enum tpm2_result tpm2_mmio_init(struct tpm2_mmio_data *md, paddr_t pbase)
+{
+	md->base.pa = pbase;
```
etienne-lms, line 71: i would start with `assert(!md->base.pa);` as init guard.

core/drivers/tpm2_mmio.c
```diff
@@ -0,0 +1,91 @@
...
+	md->chip.ops = &tpm2_mmio_ops;
+
+	md->base.va = io_pa_or_va(&md->base, TPM2_REG_SIZE);
+	DMSG("TPM2 MMIO pbase: 0x%" PRIxVA, md->base.pa);
+	DMSG("TPM2 MMIO vbase: 0x%" PRIxVA, md->base.va);
+
+	if (tpm2_init(&md->chip)) {
+		EMSG("Init failed");
+		return TPM2_ERR_INIT;
```
etienne-lms, line 80: should propagate the error code from `tpm2_init()`. Ditto for `tpm2_open()` below.

core/drivers/tpm2_mmio.c
```diff
@@ -0,0 +1,91 @@
...
+	.tx8 = tpm2_mmio_tx8,
+};
+DECLARE_KEEP_PAGER(tpm2_mmio_ops);
+
+enum tpm2_result tpm2_mmio_init(struct tpm2_mmio_data *md, paddr_t pbase)
+{
+	md->base.pa = pbase;
+	md->chip.ops = &tpm2_mmio_ops;
+
+	md->base.va = io_pa_or_va(&md->base, TPM2_REG_SIZE);
```
etienne-lms, line 74: I would rather enforce this maps a secure memory.
```
	void *va = NULL;

	md->base.pa = pbase;
	va = io_pa_or_va_secure(...);
	assert(va);
```

core/drivers/sub.mk
etienne-lms: deserves a specific `CFG_TPM2_MMIO`?

core/arch/arm/plat-vexpress/main.c
```diff
@@ -24,12 +24,18 @@
...
+#if defined(PLATFORM_FLAVOR_qemu_armv8a) && defined(CFG_TPM2)
+#include <tpm2_platform.h>
+#endif
 #include <trace.h>
 
 static struct gic_data gic_data __nex_bss;
 static struct pl011_data console_data __nex_bss;
 
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, PL011_REG_SIZE);
+#if defined(PLATFORM_FLAVOR_qemu_armv8a) && defined(CFG_TPM2)
```
etienne-lms, line 36: I suggest `#if defined(TPM2_MMIO)`

core/include/tpm2.h
apalos: Yea they follow https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClientTPMInterfaceSpecification_TIS__1-3_27_03212013.pdf and specifically "5.9.1 Interrupt Enable"

vchong:
Most comments addressed. Some have to be done during rebase. @jenswi-linaro @etienne-lms Would you prefer to review just the fixups or a new PR altogether, i.e. v2?

ldts:
hi @vchong I believe the standard use case will require TPM access from u-boot and linux (ie, I dont think op-tee will initiate any TPM requests during early boot?). 

Being that so, why not use the TPM mmio/spi/i2c drivers from those subsystems instead of having our own...just like was done on the se050 ([rpc_io_i2c.c](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/rpc_io_i2c.c))
Then all of a sudden we get support for all currently available devices. what do you think?

vchong:
> hi @vchong I believe the standard use case will require TPM access from u-boot and linux (ie, I dont think op-tee will initiate any TPM requests during early boot?).
> 
> Being that so, why not use the TPM mmio/spi/i2c drivers from those subsystems instead of the TEE having its own mmio driver; I was thinking just like was done on the se050 ([rpc_io_i2c.c](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/rpc_io_i2c.c)): then we get support for all currently available devices. what do you think?

Hi @ldts I think there's a requirement/need to do this for reading and updating the PCR as part of the boot process. @ruchi393 can probably provide more details?

ruchi393:
> > hi @vchong I believe the standard use case will require TPM access from u-boot and linux (ie, I dont think op-tee will initiate any TPM requests during early boot?).
> > Being that so, why not use the TPM mmio/spi/i2c drivers from those subsystems instead of the TEE having its own mmio driver; I was thinking just like was done on the se050 ([rpc_io_i2c.c](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/rpc_io_i2c.c)): then we get support for all currently available devices. what do you think?
> 
> Hi @ldts I think there's a requirement/need to do this for reading and updating the PCR as part of the boot process. @ruchi393 can probably provide more details?

Hi @ldts we are looking at a use-case where where eventlog is handed over to op-tee and it extends the PCR's in TPM. At this time u-boot is not up as yet.

ldts:
> > > hi @vchong I believe the standard use case will require TPM access from u-boot and linux (ie, I dont think op-tee will initiate any TPM requests during early boot?).
> > > Being that so, why not use the TPM mmio/spi/i2c drivers from those subsystems instead of the TEE having its own mmio driver; I was thinking just like was done on the se050 ([rpc_io_i2c.c](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/rpc_io_i2c.c)): then we get support for all currently available devices. what do you think?
> > 
> > 
> > Hi @ldts I think there's a requirement/need to do this for reading and updating the PCR as part of the boot process. @ruchi393 can probably provide more details?
> 
> Hi @ldts we are looking at a use-case where where eventlog is handed over to op-tee and it extends the PCR's in TPM. At this time u-boot is not up as yet.

ah ok, interesting thanks!
 the use case I am a bit more more familiar with is when a ROM verified SPL boots a signed OP-TEE and signed u-boot proper and then u-boot proper **_scripts_** the access to the TPM which could be routed to OP-TEE which would then could just use the uboot drivers.

PR #5093, opened 2022-01-10
RFC: Extending scripts/sign_encrypt.py to verify a signed TA

hoiho-amzn:
Looking for feedbacks on whether to extend `sign_encrypt.py` for
the verification functionality or create a new script.

### DIFF
--- a/scripts/sign_encrypt.py
+++ b/scripts/sign_encrypt.py
@@ -14,6 +14,11 @@ algo = {'TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA256': 0x70414930,
 enc_key_type = {'SHDR_ENC_KEY_DEV_SPECIFIC': 0x0,
                 'SHDR_ENC_KEY_CLASS_WIDE': 0x1}
 
+SHDR_BOOTSTRAP_TA = 1
+SHDR_ENCRYPTED_TA = 2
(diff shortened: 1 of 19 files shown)


scripts/sign_encrypt.py
```diff
@@ -14,6 +14,12 @@
 enc_key_type = {'SHDR_ENC_KEY_DEV_SPECIFIC': 0x0,
                 'SHDR_ENC_KEY_CLASS_WIDE': 0x1}
 
+SHDR_MAGIC = 0x4f545348
+SHDR_SIZE = 20
+
+SHDR_BOOTSTRAP_TA = 1
+SHDR_ENCRYPTED_TA = 2
```
jforissier, line 21: Nitpicking... this should be a separate commit (with the changes to lines 201-205)

jforissier (commented):
> Looking for feedbacks on whether to extend `sign_encrypt.py` for the verification functionality or create a new script.

No strong opinion, but as far as I'm concerned it is OK to extend the script.

scripts/sign_encrypt.py
hoiho-amzn: these enhancements will be on PR https://github.com/OP-TEE/optee_os/pull/5098

PR #5096, opened 2022-01-10
crypto: drivers: se050: rsa: various fixes

ldts:
crypto: drivers: se050: rsa: add RSA_NOPAD enc/dec support
---
Commit 8563cdc ("drivers: crypto: se050: limitations to RSA
crypto") removed RSA_NOPAD support based on the Plug And Trust MW
documentation, Release v02,14,00 (Apr 03, 2020).

That documentation was incorrect as RSA_NOPAD is indeed supported by
the secure element as described in the SE050 APDU specification [1],
section 4.3.14, table 32.

This commit restores the functionality and fixes previous bugs.

Validated on xtest 4006 and 4011.

[1] https://www.nxp.com/docs/en/application-note/AN12413.pdf

crypto: drivers: se050: rsa: fix OAEP and revert regression
--
Revert a regression introduced in the encrypt operation when swapping
buffers (fixes part of 'commit e1c70d7 ("crypto: drivers: se050:
fix rsa encrypt/decrypt")'

Fix misuse of the hash_algo field during OAEP encrypt/decrypt.

All tests passing
  * xtest -t regression 4006

crypto: drivers: se050: rsa: decrypt_es, validate the output buffer
--
The size of the decrypted output is not known until decryption has
happened.

Use an intermediate buffer large enough to guarantee that the
decrypted message will fit.

This allows the driver to validate the size of the output buffer
passed in the interface.

pkcs11 test passing now
 * xtest pkcs11_1023.

 crypto: drivers: se050: rsa: sign_ssa error handling
--
SE NVM keys shall only be deleted using either the pkcs#11 interface
(if the key was created by pkcs#11) or the free_keypair crypto API
interface and never as a result of some error handling operation.
    
Notice that calling free_keypair will invalidate any copy made of that
keypair since the keypair for a SE only holds a handle to the key
stored in the SE NVM.

### DIFF
--- a/core/drivers/crypto/se050/core/rsa.c
+++ b/core/drivers/crypto/se050/core/rsa.c
@@ -193,6 +193,8 @@ static TEE_Result decrypt_es(uint32_t algo, struct rsa_keypair *key,
 	sss_se05x_asymmetric_t ctx = { };
 	sss_se05x_object_t kobject = { };
 	TEE_Result res = TEE_SUCCESS;
+	uint8_t *buf = NULL;
+	size_t buf_len = src_len;
 
 	res = se050_inject_keypair(&kobject, key);
 	if (res)
@@ -207,14 +209,33 @@ static TEE_Result decrypt_es(uint32_t algo, struct rsa_keypair *key,
 		return TEE_ERROR_BAD_PARAMETERS;
 	}
 
-	st = sss_se05x_asymmetric_decrypt(&ctx, src, src_len, dst, dst_len);
-	if (st != kStatus_SSS_Success)
-		res = TEE_ERROR_BAD_PARAMETERS;
+	/* we don't know the size of the decrypted data, just the upper limit */
+	buf = mempool_calloc(mempool_default, 1, buf_len);
+	if (!buf) {
+		res = TEE_ERROR_OUT_OF_MEMORY;
+		goto out;
+	}
 
+	st = sss_se05x_asymmetric_decrypt(&ctx, src, src_len, buf,  &buf_len);
+	if (st != kStatus_SSS_Success) {
+		res = TEE_ERROR_BAD_PARAMETERS;
+		goto out;
+	}
+
+	if (buf_len > *dst_len) {
+		*dst_len = buf_len;
+		res = TEE_ERROR_SHORT_BUFFER;
+		goto out;
+	}
+
+	*dst_len = buf_len;
+	memcpy(dst, buf, buf_len);
+out:
 	if (!se050_rsa_keypair_from_nvm(key))
 		sss_se05x_key_store_erase_key(se050_kstore, &kobject);
 
 	sss_se05x_asymmetric_context_free(&ctx);
+	mempool_free(mempool_default, buf);
 
 	return res;
 }
--- a/core/drivers/crypto/se050/core/rsa.c
+++ b/core/drivers/crypto/se050/core/rsa.c
@@ -9,6 +9,7 @@
 #include <drvcrypt_acipher.h>
 #include <drvcrypt_math.h>
 #include <initcall.h>
+#include <mempool.h>
 #include <se050.h>
 #include <string.h>
 #include <tee/cache.h>
@@ -97,6 +98,11 @@ static TEE_Result se050_inject_public_key(sss_se05x_object_t *k_object,
 	if (st != kStatus_SSS_Success)
 		return TEE_ERROR_GENERIC;
 
+	/*
+	 * Keys 2048 and above MUST to be placed on persistent storage even
+	 * though the keys will be deleted after the operation. This is a
+	 * memory restriction in the secure element.
+	 */
 	st = sss_se05x_key_object_allocate_handle(k_object, oid,
 						  kSSS_KeyPart_Public,
 						  kSSS_CipherType_RSA, 0,
@@ -143,10 +149,11 @@ static TEE_Result se050_inject_keypair(sss_se05x_object_t *k_object,
 	if (st != kStatus_SSS_Success)
 		return TEE_ERROR_GENERIC;
 
+	/* Keys 2048 and above need to be placed on persistent storage */
 	st = sss_se05x_key_object_allocate_handle(k_object, oid,
 						  kSSS_KeyPart_Pair,
 						  kSSS_CipherType_RSA, 0,
-						  kKeyObject_Mode_Transient);
+						  kKeyObject_Mode_Persistent);
 	if (st != kStatus_SSS_Success)
 		return TEE_ERROR_BAD_PARAMETERS;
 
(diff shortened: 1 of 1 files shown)


core/drivers/crypto/se050/core/rsa.c
```diff
@@ -15,6 +15,12 @@
 #include <tee/tee_cryp_utl.h>
 #include <tee_api_defines_extensions.h>
 
+static inline uint32_t oaep_alg_from_hash_alg(uint32_t hash_alg)
+{
+	return (TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA1 & ~GENMASK_32(21, 20)) |
```
jenswi-linaro, line 20: I expect that it's bits [23:20] which are to replaced by bits [3:0] from `hash_alg`.
Please update the code accordingly and add a comment on what's going on.

jenswi-linaro (commented):
> All tests (level 0) except the unsupported (and unsafe) RSA_NOPAD passing
>
> xtest -t regression 4006

What does that mean? All tests are not passing? Any new regressions? What's the plan for eventual failures?

core/drivers/crypto/se050/core/rsa.c
etienne-lms: An explicit switch/case for the supported hash algo would maybe be easier to maintain.

ldts:
> What does that mean? All tests are not passing? Any new regressions? What's the plan for eventual failures?


right I did want to have this conversation - should xtest **not** yield the suite when the crypto operation is not supported?

RSA_NOPAD is currently not supported on the SE05X driver - the driver returns TEE_ERROR_NOT_IMPLEMENTED to the crypto API. 

However xtest 4006 will abort on that failure which is not what we need. So I am kind of forced to do this in order to successfully execute 4006 (ie: disable RSA_NOPAD encrypt/decrypt requests)

```
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -2781,6 +2781,7 @@ struct xtest_ac_case {
 
 static const struct xtest_ac_case xtest_ac_cases[] = {
        /* RSA test without crt parameters */
+#ifndef CFG_CRYPTO_SE05X
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_ENCRYPT,
                          ac_rsassa_vect1, NULL_ARRAY, WITHOUT_SALT),
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_DECRYPT,
@@ -2797,6 +2798,7 @@ static const struct xtest_ac_case xtest_ac_cases[] = {
                          ac_rsassa_vect19, NULL_ARRAY, WITHOUT_SALT),
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_DECRYPT,
                          ac_rsassa_vect19, NULL_ARRAY, WITHOUT_SALT),
+#endif
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSASSA_PKCS1_V1_5_SHA1, TEE_MODE_SIGN,
                          ac_rsassa_vect3, NULL_ARRAY, WITHOUT_SALT),
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSASSA_PKCS1_V1_5_SHA1, TEE_MODE_VERIFY,
@@ -2916,6 +2918,7 @@ static const struct xtest_ac_case xtest_ac_cases[] = {
                          ac_rsaes_oaep_vect10, NULL_ARRAY, WITHOUT_SALT),
 
        /* RSA test with crt parameters */
+#ifndef CFG_CRYPTO_SE05X
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_ENCRYPT,
                          ac_rsassa_vect1, ARRAY, WITHOUT_SALT),
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_DECRYPT,
@@ -2924,6 +2927,7 @@ static const struct xtest_ac_case xtest_ac_cases[] = {
                          ac_rsassa_vect2, ARRAY, WITHOUT_SALT),
        XTEST_AC_RSA_CASE(1, TEE_ALG_RSA_NOPAD, TEE_MODE_DECRYPT,
                          ac_rsassa_vect2, ARRAY, WITHOUT_SALT),
+#endif
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSASSA_PKCS1_V1_5_SHA1, TEE_MODE_SIGN,
                          ac_rsassa_vect3, ARRAY, WITHOUT_SALT),
        XTEST_AC_RSA_CASE(0, TEE_ALG_RSASSA_PKCS1_V1_5_SHA1, TEE_MODE_VERIFY,


```

doing the above, 4006 works as expected

core/drivers/crypto/se050/core/rsa.c
ldts: > An explicit switch/case for the supported hash algo would maybe be easier to maintain.

I dont see why....I'd rather handle the algorithm rather than the hash algorithm. Seems cleaner to me.

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -15,6 +15,12 @@
 #include <tee/tee_cryp_utl.h>
 #include <tee_api_defines_extensions.h>
 
+static inline uint32_t oaep_alg_from_hash_alg(uint32_t hash_alg)
```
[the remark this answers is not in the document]
ldts: um. ok...maybe I'll just write a macro for this.

jenswi-linaro:
> right I did want to have this conversation - should xtest not yield the suite when the crypto operation is not supported?

Updating `TEE_IsAlgorithmSupported()` and the check around:
https://github.com/OP-TEE/optee_test/blob/01322455ed79f3c9a862a222df194d962a2093f3/host/xtest/regression_4000.c#L3638-L3644
should do it.

ldts:
> > right I did want to have this conversation - should xtest not yield the suite when the crypto operation is not supported?
> 
> Updating `TEE_IsAlgorithmSupported()` and the check around: https://github.com/OP-TEE/optee_test/blob/01322455ed79f3c9a862a222df194d962a2093f3/host/xtest/regression_4000.c#L3638-L3644 should do it.

Um, this CONFIG filtering is not ideal - btw it seems it wont work for RSA_NOPAD as it is not controlled by config (so it cant be disabled unless we disable it for everyone)

It would be great the check could be done at runtime for each _request_ :not the checking for the algorithm being available but the actual request with the parameters being supported or not.
The SE050 supports RSA keygen but not for all key sizes...ie: 256, 384, 640 are not supported so those tests would  also fail.

Runtime validation is not a simple xtest modification so perhaps it can be added to some planning tool/JIRA?

ldts:
BTW incidentally I am going to add RSA_NOPAD support back again to the SE05x driver: it  is showing as supported on the list of APDU requests even though it doesnt show in the plug-and-trust stack documentation.

ldts:
hi @jenswi-linaro @etienne-lms 
I restored the support for RSA_NOPAD encryption and fixed the comments regarding the inline function

Validated using
* xtest 4006 
* xtest 4011

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -15,6 +15,10 @@
 #include <tee/tee_cryp_utl.h>
 #include <tee_api_defines_extensions.h>
 
+#define TEE_OAEP_ALG(__hash_alg) \
```
jenswi-linaro, line 18: What this macro does with a few words.

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -15,6 +15,10 @@
...
+	((TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA1 & ~GENMASK_32(23, 20)) | \
+	SHIFT_U32(__hash_alg, 20))
```
jenswi-linaro, line 20: Since we're only interested in the lowest four bits we should use a mask for that.

core/drivers/crypto/se050/core/rsa.c
ldts: from an OAEP hash algorithm, generates the OAEP algorithm 
(just undo what the crypto API did)

core/drivers/crypto/se050/core/rsa.c
ldts: um ok, but it is unnecessary code since the hash_algo must have zeros in those values anyway.

core/drivers/crypto/se050/core/rsa.c
jenswi-linaro: I was more looking for a comment like:
/* Makes a TEE_ALG_RSAES_PKCS1_OAEP_MGF1_* from input hash function */

core/drivers/crypto/se050/core/rsa.c
jenswi-linaro: Yeah, just as unnecessary as GENMASK_32 above where it's only a single bit that you need to clear.

ldts: nah, that is silly: the one you want added does nothing while the one that I added is required (even is clearing a couple of bits that will be zeros anyway).
but anyway, I am going to remove the macro altogether since only SHA1 seems to be required according to latest spec. - though that might change in the future. 

https://www.nxp.com/docs/en/application-note/AN12413.pdf (page 40)

ldts:
@jenswi-linaro updated removing macro

ldts:
would be great to have this merged before the next release

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -247,6 +252,129 @@ static TEE_Result encrypt_es(uint32_t algo, struct rsa_public_key *key,
...
+	st = sss_se05x_asymmetric_context_init(&ctx, se050_session, &kobject,
+					       kAlgorithm_SSS_RSASSA_NO_PADDING,
+					       kMode_SSS_Decrypt);
+	if (st != kStatus_SSS_Success) {
+		sss_se05x_key_store_erase_key(se050_kstore, &kobject);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	blen = CFG_CORE_BIGNUM_MAX_BITS / 8;
+	buf = calloc(1, blen);
```
jenswi-linaro, line 301: Please use `mempool_calloc(mempool_default, 1, blen)` instead as this would pretty much guarantee to succeed and a calling TA doesn't have to panic.

Same for the other short-lived allocations that are freed before the calling function returns.

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -247,6 +274,129 @@ static TEE_Result encrypt_es(uint32_t algo, struct rsa_public_key *key,
...
+		offset++;
+
+	if (*dst_len < blen - offset) {
+		*dst_len = blen - offset;
+		res = TEE_ERROR_SHORT_BUFFER;
+		goto out;
+	}
+
+	*dst_len = blen - offset;
+	memcpy(dst, (char *)buf + offset, *dst_len);
```
jenswi-linaro, line 329: Cast is not needed, buf is already of a suitable type.

etienne-lms (commented):
Comment for  commit "crypto: drivers: se050: rsa: add RSA_NOPAD enc/dec support".

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commits
"crypto: drivers: se050: rsa: fix OAEP and revert regression"
"crypto: drivers: se050: rsa: decrypt_es, validate the output buffer"

core/drivers/crypto/se050/core/rsa.c
```diff
@@ -247,6 +252,129 @@ static TEE_Result encrypt_es(uint32_t algo, struct rsa_public_key *key,
...
+out:
+	free(buf);
+	if (!se050_rsa_keypair_from_nvm(key))
```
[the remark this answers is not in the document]
ldts: they are different use cases. let me explain

If line 272 fails to initialize the context - with any type of key - I am assuming there is a problem with the key (either it is invalid or it is corrupted). Hence why I think it is acceptable to delete it on all cases.

in line 311 we certainly do not want to delete a working key that resides in the SE NVM as the owner will very likely request it again. Keys that are not in NVM can be deleted at this point as from the POV of the SE they no longer matter.

So the code is correct.

jenswi-linaro:

core/drivers/crypto/se050/core/rsa.c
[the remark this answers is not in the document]
etienne-lms: Maybe add an inline comment could clarify that. The implemented sequence is not obvious (to me :).
Note test on `!se050_rsa_keypair_from_nvm(key)` is not done at line 374 (in `encrypt_nopad()`).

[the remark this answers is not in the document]
ldts: ok let me clarify.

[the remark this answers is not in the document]
ldts: will do the same to the sign operation. then it is consistent all over.

jforissier:
@jenswi-linaro @etienne-lms could you please review/ack the last commit "crypto: drivers: se050: rsa: sign_ssa error handling"? then this could be merged for 3.16.0-rc1.

jenswi-linaro:
Do we care about a fixes tag on "crypto: drivers: se050: rsa: sign_ssa error handling"?
Regardless:

jforissier:
> Do we care about a fixes tag on "crypto: drivers: se050: rsa: sign_ssa error handling"?

Certainly wouldn't hurt ;) If anything, Fixes: is informational, and in case the original commit is part of a release then it can also help with back-porting.

ldts:
ok

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commits
"crypto: drivers: se050: rsa: add RSA_NOPAD enc/dec support"
"crypto: drivers: se050: rsa: sign_ssa error handling" (nitpick: remove the empty line between `Fixes:` and `S-o-b` tags in commit log)

jforissier:
> Acked-by: Etienne Carriere <etienne.carriere@linaro.org>

Done and pushed to master.

PR #5098, opened 2022-01-10
scripts/sign_encrypt.py: readability improvements

hoiho-amzn:
A few constant values would have been better if replaced with constant
definitions

Signed-off-by: Donald Chan <hoiho@amazon.com>


jforissier (approved):
Thanks for the patch!

PR #5102, opened 2022-01-11
FF-A device memory

jellesels-arm:
Add support for FF-A SPs to map device memory. 
For SPs to be able to map memory as a device memory region the following changes have to be made:

-Extend `mmu_lpae` to be able to map the different device types.
-Extend sp_mem to allow to map device memory
-Extend the FF-A sharing messages to support memory to be mapped as a device region.

### DIFF
--- a/core/arch/arm/include/ffa.h
+++ b/core/arch/arm/include/ffa.h
@@ -78,6 +78,12 @@
 #define FFA_NORMAL_MEM_REG_ATTR		U(0x2f)
 
 /* Device memory attributes */
+#define FFA_DEV_MEM_REG_ATTR_MASK	U(0x1c)
+#define FFA_DEV_MEM_REG_ATTR		BIT(4)
+
+/* FF-A Device memory attributes */
+#define FFA_DEV_MEM_SHIFT		U(2)
+#define FFA_DEV_MEM_MASK		U(0x3)
 #define FFA_DEV_MEM_nGnRnE		U(0x0)
 #define FFA_DEV_MEM_nGnRE		U(0x1)
 
--- a/core/arch/arm/include/ffa.h
+++ b/core/arch/arm/include/ffa.h
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
  * Copyright (c) 2020, Linaro Limited
- * Copyright (c) 2018-2021, Arm Limited. All rights reserved.
+ * Copyright (c) 2018-2022, Arm Limited. All rights reserved.
  */
 
 #ifndef __FFA_H
@@ -76,6 +76,10 @@
 
 /* Memory attributes: Normal memory, Write-Back cacheable, Inner shareable */
 #define FFA_NORMAL_MEM_REG_ATTR		U(0x2f)
+
+/* Device memory attributes */
+#define FFA_DEV_MEM_nGnRnE		U(0x0)
+#define FFA_DEV_MEM_nGnRE		U(0x1)
 
 /* Memory access permissions: Read-write */
 #define FFA_MEM_ACC_RW			BIT(1)
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2020-2021, Arm Limited.
+ * Copyright (c) 2020-2022, Arm Limited.
  */
 #include <bench.h>
 #include <crypto/crypto.h>
@@ -240,6 +240,8 @@ TEE_Result sp_map_shared(struct sp_session *s,
 	struct sp_ctx *ctx = NULL;
 	uint32_t perm = TEE_MATTR_UR;
 	struct sp_mem_map_region *reg = NULL;
+	uint32_t cache_type = 0;
+	bool is_device = smem->mem_reg_attr & FFA_DEV_MEM_REG_ATTR;
 
 	ctx = to_sp_ctx(s->ts_sess.ctx);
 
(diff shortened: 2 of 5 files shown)


core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -121,10 +121,22 @@
...
+#define ATTR_DEVICE_IO_INDEX		0x2
+#define ATTR_DEVICE_nGnRnE_INDEX	0x3
+#define ATTR_DEVICE_nGRE_INDEX		0x4
+#define ATTR_DEVICE_GRE_INDEX		0x5
 #define ATTR_INDEX_MASK			0x7
 
+#define ATTR_DEVICE_MASK		(0xf)
 #define ATTR_DEVICE			(0x4)
+
+#define ATTR_DEVICE_nGnRnE		(0x0)
```
jenswi-linaro, line 135: Do we need all these types? So far we've been able to keep the number of different types down by only adding what's really needed.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -121,10 +121,22 @@
 #define LOWER_ATTRS(x)			(((x) & 0xfff) << LOWER_ATTRS_SHIFT)
 
 #define ATTR_DEVICE_INDEX		0x0
+#define ATTR_DEVICE_nGnRE_INDEX		0x0
 #define ATTR_IWBWA_OWBWA_NTR_INDEX	0x1
+#define ATTR_DEVICE_IO_INDEX		0x2
```
jenswi-linaro, line 126: What's this?

jellesels-arm: Not needed

core/arch/arm/mm/core_mmu_lpae.c
jellesels-arm: The spec allows normally all of them. But I am checking internally if we need to support all of them.

jenswi-linaro: Thanks.

jellesels-arm: I checked and we can give an error back for the none supported once. However the spec states:

jenswi-linaro: OK, no problem.

jellesels-arm: I update the patch and only allow nGnRE and nGnRnE

core/mm/vm.c
```diff
@@ -266,7 +266,7 @@ TEE_Result vm_map_pad(struct user_mode_ctx *uctx, vaddr_t *va, size_t len,
 	struct vm_region *reg = NULL;
 	uint32_t attr = 0;
 
-	if (prot & ~TEE_MATTR_PROT_MASK)
+	if (prot & ~TEE_MATTR_PROT_MASK && prot & ~TEE_MATTR_DEVICE_PROT_MASK)
```
jenswi-linaro, line 269: This is not the type of check I expect that this function should perform.

core/include/mm/tee_mmu_types.h
```diff
@@ -29,6 +30,12 @@
 
 #define TEE_MATTR_GLOBAL		BIT(10)
 #define TEE_MATTR_SECURE		BIT(11)
+#define TEE_MATTR_DEVICE		BIT(13)
```
jenswi-linaro, line 36: Wouldn't it make more sense to group this device stuff as different kinds of cache settings?
If you don't want to call it `TEE_MATTR_CACHE*` we could rename it to `TEE_MATTR_MEM_TYPE*`

core/mm/vm.c
jellesels-arm: Do you want me to just remove:
`if (prot & ~TEE_MATTR_PROT_MASK)`
Or do you have anything else specifically in mind?

jenswi-linaro: Please remove `&& prot & ~TEE_MATTR_DEVICE_PROT_MASK`.

jellesels-arm: The reason I add the check for the `TEE_MATTR_DEVICE_PROT_MASK` is that for device regions we have some extra flags that are not inside the `TEE_MATTR_PROT_MASK`. I thought having two different flags on for testing the device regions and one for testing the normal region would be the best approach.  Since the normal regions  don't have the TEE_MATTR_DEVICE_nGnRE and TEE_MATTR_DEVICE_nGnRnE flags.
Do you want me to update the `TEE_MATTR_PROT_MASK` flag instead to also cover the device regions?

jenswi-linaro: The flags you're mentioning aren't protection or permission flags. I'd expect device mapping attributes to  come from the `mobj` describing the memory to map. Check how `attr` is constructed below.

jellesels-arm: Ack

core/include/mm/tee_mmu_types.h
jellesels-arm: ACK

jenswi-linaro (commented):
Comments for the commit "core: Add device memory mapping to mmu_lpae"

Until now we've had only two memory types, cached and non-cached. Since
we're about to change that we should do it in a way to cause as little
confusion as possible.

Please create a new PR with this commit only to give this change a bit more
attention and also to allow us to get this through without more delay than 
necessary.

jellesels-arm:
> Comments for the commit "core: Add device memory mapping to mmu_lpae"
> 
> Until now we've had only two memory types, cached and non-cached. Since we're about to change that we should do it in a way to cause as little confusion as possible.
> 
> Please create a new PR with this commit only to give this change a bit more attention and also to allow us to get this through without more delay than necessary.

I created the new PR: https://github.com/OP-TEE/optee_os/pull/5193

jellesels-arm:
> > Comments for the commit "core: Add device memory mapping to mmu_lpae"
> > Until now we've had only two memory types, cached and non-cached. Since we're about to change that we should do it in a way to cause as little confusion as possible.
> > Please create a new PR with this commit only to give this change a bit more attention and also to allow us to get this through without more delay than necessary.
> 
> I created the new PR: #5193
PR is rebased as #5193 was merged.

jenswi-linaro:
If normal world still has this memory mapped we may run into incoherent memory between the two worlds. How is this going to be managed?

jellesels-arm:
> If normal world still has this memory mapped we may run into incoherent memory between the two worlds. How is this going to be managed?

I would leave it up to the Normal World to make sure that this doesn't happen. If this is not sufficient we could always disable device sharing from the Normal World. This PR would still be needed as it lays the foundation for [5215]( https://github.com/OP-TEE/optee_os/pull/5215). Here we use this code to map device memory regions from the manifest file. 

@jenswi-linaro: let me know what you think.

jenswi-linaro:
I don't understand why you need to device map normal memory which also is shared with normal world.

jellesels-arm:
Closing this PR as it make more sense to combine it inside [5215](https://github.com/OP-TEE/optee_os/pull/5215).

PR #5105, opened 2022-01-11
Fix race in mobj_inc_map

jenswi-linaro:

### DIFF
--- a/core/arch/arm/mm/mobj_dyn_shm.c
+++ b/core/arch/arm/mm/mobj_dyn_shm.c
@@ -114,6 +114,7 @@ static void *mobj_reg_shm_get_va(struct mobj *mobj, size_t offst)
 
 static void reg_shm_unmap_helper(struct mobj_reg_shm *r)
 {
+	assert(r->mm);
 	assert(r->mm->pool->shift == SMALL_PAGE_SHIFT);
 	core_mmu_unmap_pages(tee_mm_get_smem(r->mm), r->mm->size);
 	tee_mm_free(r->mm);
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
These problems are usually not visible with the current setup, but I'm working on a kernel patch where these problems suddenly appears. So it would be quite nice to have these in this release to be proactive.

core/arch/arm/mm/mobj_dyn_shm.c
```diff
@@ -183,28 +184,41 @@ static TEE_Result mobj_reg_shm_inc_map(struct mobj *mobj)
 	uint32_t exceptions = 0;
 	size_t sz = 0;
 
-	if (refcount_inc(&r->mapcount))
-		return TEE_SUCCESS;
+	while (true) {
+		if (refcount_inc(&r->mapcount))
+			return TEE_SUCCESS;
```
jforissier, line 189: "When val is 0, refcount_inc() does not change the value and returns false."
So when `r->mapcount` is initially 0, how will any thread proceed?

jenswi-linaro: If val is 0 then we take the spinlock below and check that it's still 0 now that we have a lock. Then we map the mobj and initialize the mapcount to 1.

In case the val wasn't 0 any longer when we have the lock we instead increase the mapcount and return.

jforissier: Oh I'm sorry I totally missed the opening brace... :facepalm: thanks for clarifying

jenswi-linaro: No worries

jforissier (approved):

jenswi-linaro:
Tag applied.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for both commits (with comment addressed).

jenswi-linaro:
Addressed comment, applied tags.

PR #5109, opened 2022-01-13
core: the FF-A ABI is now a stable ABI

jenswi-linaro:
The OP-TEE FF-A driver in the Linux kernel has been merged, so the
changes in the ABI towards the Linux kernel from now on have to be
backwards compatible.


jforissier (commented):

jenswi-linaro:
Tag applied.

PR #5117, opened 2022-01-17
Compilation error/warning for wdog imx driver

clementfaure:


core/drivers/imx_wdog.c
```diff
@@ -154,18 +154,16 @@ static TEE_Result imx_wdog_base(vaddr_t *wdog_vbase)
 	*wdog_vbase = (vaddr_t)phys_to_virt(WDOG_BASE, MEM_AREA_IO_SEC, 1);
 #if defined(CFG_IMX_WDOG_EXT_RESET)
 	ext_reset_output = true;
+#else
+	ext_reset_output = false;
```
jforissier, line 158: Global variables are always initialized to `false` by default, so this is not desirable. I'd much prefer you add `__maybe_unused` to the definition of `ext_reset_output`.

clementfaure:
Should I wrap the compilation error message in the commit message ?

jforissier:
> Should I wrap the compilation error message in the commit message ?

I prefer not. Checkpatch has a special rule to match compiler messages, but this syntax (` | <some text>`) is not handled yet it seems...

jforissier (commented):

Two minor comments:
- A better prefix may be `core: drivers: imx_wdog:` for both commits
- "compilation warning" not "warning compilation" in the second commit subject

clementfaure:
@jforissier Tags applied
- I've a slight preference for `drivers: imx_wdog` rather than `core: drivers: imx_wdog`. I hope it's fine for you.
- Fixed typo

PR #5123, opened 2022-01-18
core: imx: add support for i.MX8DXL

clementfaure:
Hello!
A patch to add the i.MX 8DXL platform support.
Please note that the CAAM is currently disabled and will be enabled later.

Thanks,
Clement


jforissier (approved):

PR #5124, opened 2022-01-18
scripts/sign_encrypt.py: Support signing TAs using AWS KMS

hoiho-amzn:
This adds support for signing trusted applications (TAs) using
a customer owned AWS KMS asymmetric key.

When the option to --key points to a valid Amazon Resource Name (ARN),
the signing operation will be delegated to AWS KMS. IAM credentials are
provided via environment variables.

Requires boto3 to work correctly.

Signed-off-by: Donald Chan <hoiho@amazon.com>

### DIFF
--- a/scripts/sign_encrypt.py
+++ b/scripts/sign_encrypt.py
@@ -98,7 +98,9 @@ def get_args(logger):
     parser.add_argument('--uuid', required=True,
                         type=uuid_parse, help='String UUID of the TA')
     parser.add_argument('--key', required=True,
-                        help='Name of signing key file (PEM format)')
+                        help='Name of signing key file (PEM format) or an ' +
+                             'Amazon Resource Name (arn:) of an AWS KMS ' +
+                             'asymmetric key')
     parser.add_argument('--enc-key', required=False,
                         help='Encryption key string')
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


scripts/sign_encrypt.py
```diff
@@ -182,15 +182,22 @@ def main():
 
     args = get_args(logger)
 
-    with open(args.key, 'rb') as f:
-        data = f.read()
+    if args.key.startswith('arn'):
```
jforissier, line 185: Could you make this `arn:` instead? Rationale: the colon makes it more unlikely to be a PEM file.

jforissier (commented):
Hi @hoiho-amzn,

Thanks for the patch. Could you please update the help string for `--key`? Something like: "Name of signing key file (PEM format) or Amazon Resource Name (arn:) of an AWS KMS asymmetric key.". With that and the comment below addressed:

scripts/sign_encrypt.py
hoiho-amzn: agreed

hoiho-amzn:

Tried my best to put the help string in the most appropriate locations....but the entire help section might need some revamping eventually - for example, the `digest` command shouldn't be requiring the `--key` argument since it just outputs a hash digest.

jforissier:
> All comments have been addressed.
> 
> Tried my best to put the help string in the most appropriate locations....

Honestly the best place would be https://github.com/OP-TEE/optee_os/pull/5124/commits/dfd9eb285ac52c691bf37f9a1209390c5c2460b1#diff-d1c7598c2461aa2f6bf8d62e319fbfc02e576955e7bb1ed8e82394c728ba4eaaR108-R109 and do not change the big block of help text.

> but the entire help section might need some revamping eventually - for example, the `digest` command shouldn't be requiring the `--key` argument since it just outputs a hash digest.

I agree. One way might be to use sub-parsers (`ArgumentParser.add_subparsers()`), I think it could help structure things better.

hoiho-amzn:
OK - this help string change looks the cleanest to me.

PR #5125, opened 2022-01-19
drivers: crypto: se050: build plug-and-trust in tree

ldts:
Building the Plug-and-Trust library required building OP-TEE first in
order to get some architecture specific definitions. This makes the
integration with Yocto metas unnecessarily complex.

The following commit simplifies the build sequence: the user would
clone the Plug-and-Trust tree [1] into the external (ext) folder and
build OP-TEE as usual.

[1] https://github.com/foundriesio/plug-and-trust.git

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -72,7 +72,7 @@ jobs:
           function upload_cache() { if [ ! -e .uploaded ]; then echo Uploading cache && tar c -C $HOME .ccache | gzip -1 | ssh $SCP_OPT optee_os_ci@cache.forissier.org "cat >ccache-$PROJ.tar.gz" && touch .uploaded || echo Nevermind; fi; }
           function check_upload_cache() { NOW=$(date +%s); if [ $(expr $NOW - $START) -gt 3000 ]; then upload_cache; fi; }
           function _make() { make -j$(getconf _NPROCESSORS_ONLN) -s O=out $* && check_upload_cache; }
-          function download_plug_and_trust() { mkdir -p $HOME/se050 && curl -L https://github.com/foundriesio/plug-and-trust/releases/download/v0.0.4/se050-0.0.4.tar.bz2 | tar jxf - --strip-components=1 -C $HOME/se050 || (rm -rf $HOME/se050; echo Nervermind); }
+          function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.1.0 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
 
           download_cache
           ccache -s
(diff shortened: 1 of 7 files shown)


ldts:
@jforissier not sure if you think it will be possible but this would be really nice to have on 3.16. It does simplify yocto builds by a lot.

core/drivers/crypto/se050/ext/sub.mk
```diff
@@ -0,0 +1,42 @@
+include core/drivers/crypto/se050/cflags.mk
```
jforissier, line 1: Shouldn't this whole file be part of the upstream library code? (if a file is renamed for example, this would need updating).

jforissier (commented):
> @jforissier not sure if you think it will be possible but this would be really nice to have on 3.16. It does simplify yocto builds by a lot.

I have one comment below, but I don't mind taking this even if it's not addressed right now.
@jenswi-linaro what do you think?

jenswi-linaro:
> > @jforissier not sure if you think it will be possible but this would be really nice to have on 3.16. It does simplify yocto builds by a lot.
> 
> I have one comment below, but I don't mind taking this even if it's not addressed right now. @jenswi-linaro what do you think?

Sure. I don't understand why the plug and trust library needs to be unpacked inside the optee_os git though. That makes cleaning with `git clean -fdx` a bit more inconvenient to use.

core/drivers/crypto/se050/ext/sub.mk
ldts: yes, will fix that.

ldts:
> > > @jforissier not sure if you think it will be possible but this would be really nice to have on 3.16. It does simplify yocto builds by a lot.
> > 
> > 
> > I have one comment below, but I don't mind taking this even if it's not addressed right now. @jenswi-linaro what do you think?
> 
> Sure. I don't understand why the plug and trust library needs to be unpacked inside the optee_os git though. That makes cleaning with `git clean -fdx` a bit more inconvenient to use.

yes...ok, let me address that as well.

ldts:
> > > > @jforissier not sure if you think it will be possible but this would be really nice to have on 3.16. It does simplify yocto builds by a lot.
> > > 
> > > 
> > > I have one comment below, but I don't mind taking this even if it's not addressed right now. @jenswi-linaro what do you think?
> > 
> > 
> > Sure. I don't understand why the plug and trust library needs to be unpacked inside the optee_os git though. That makes cleaning with `git clean -fdx` a bit more inconvenient to use.
> 
> yes...ok, let me address that as well.

is there a way to pass an absolute path to subdirs?

jenswi-linaro:
> is there a way to pass an absolute path to subdirs? I suppose I could just go down the current directory but is not straight forward and seems hacky

Yes, I hope that works at least.

ldts:
> > is there a way to pass an absolute path to subdirs? I suppose I could just go down the current directory but is not straight forward and seems hacky
> 
> Yes, I hope that works at least.

um...no, I think this will need some .mk change but I havent figured out what yet.

jenswi-linaro:
Looks good to me. I was first considering checking if the patch started with a `/`, but it's actually quite nice to be explicit about whether path is supposed to be in the tree or outside the tree.

jforissier:
@ldts I agree with @jenswi-linaro, your proposal looks good :+1:

jforissier (approved):
LGTM

ldts:
@jforissier tags updated

PR #5126, opened 2022-01-19
core: Fix flags and tag check_retrieve_request()

jellesels-arm:
Inside check_retrieve_request() the flags and tag where swapped.
Store them in the correct variable.

Signed-off-by: Jelle Sels <jelle.sels@arm.com>

### DIFF
--- a/core/arch/arm/kernel/spmc_sp_handler.c
+++ b/core/arch/arm/kernel/spmc_sp_handler.c
@@ -377,8 +377,8 @@ static TEE_Result check_retrieve_request(struct sp_mem_receiver *receiver,
 	struct ffa_mem_access *retr_access = NULL;
 	uint8_t share_perm = receiver->perm.perm;
 	uint32_t retr_perm = 0;
-	uint32_t retr_flags = READ_ONCE(retr_dsc->tag);
-	uint64_t retr_tag = READ_ONCE(retr_dsc->flags);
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
It looks like we've merged untested code.
Please add a fixes tag for the original commit.
With that:

jellesels-arm:
> It looks like we've merged untested code. Please add a fixes tag for the original commit. With that: `Acked-by: Jens Wiklander <jens.wiklander@linaro.org>`
Added the fixes tag.
I will have a look why this was not catches by the tests, cause I ran the my tests. I assume that in my original tests the value of the flags and tag was the same.

jenswi-linaro:
Please make the tag
`Fixes: de66193d9849 ("core: FF-A: ADD FFA_MEM_RETRIEVE for SPs")`

jellesels-arm:
> Please make the tag `Fixes: de66193d9849 ("core: FF-A: ADD FFA_MEM_RETRIEVE for SPs")`

Done

jforissier:
@jellesels-arm small nits for future commits: no need for a blank line between `Fixes:` and other tags (`Signed-off-by:`, etc.) and also it would be good to identify the affected "subsystem" more precisely in the subject line, for example by adding `ff-a:` or `spmc:` after `core:`.

PR #5130, opened 2022-01-20
ta: pkcs11: support ECDH derive

ldts:
Add support for ECDH derive operation.

This commit integrates previous work done by Etienne Carriere and Vesa
Jääskeläinen.

Tested with pkcs11_tool -m ECDH1-DERIVE

ref https://github.com/OP-TEE/optee_client/pull/298


ta/pkcs11/include/pkcs11_ta.h
```diff
@@ -1248,6 +1248,9 @@ enum pkcs11_mechanism_id {
 	PKCS11_CKM_ECDSA_SHA224			= 0x01043,
 	PKCS11_CKM_ECDSA_SHA256			= 0x01044,
 	PKCS11_CKM_ECDSA_SHA384			= 0x01045,
+	PKCS11_CKM_ECDH1_DERIVE = 0x1050,
```
etienne-lms, line 1251: move below line 1254 and indent `= ...`.

ta/pkcs11/src/processing.c
```diff
@@ -981,13 +981,27 @@ enum pkcs11_rc entry_processing_key(struct pkcs11_client *client,
...
+			goto out;
+
+		rc = init_asymm_operation(session, function, proc_params,
+					  parent);
+		if (rc)
+			goto out;
+
+		rc = do_asymm_derivation(session, proc_params, &head);
+		if (rc)
+			goto out;
```
etienne-lms, line 995: nitpickiong but I would prefer a standard way.
```
	} else if (processing_is_tee_asymm()) {
		assert(function == PKCS11_FUNCTION_DERIVE);

		rc = init_asymm_operation(session, function, proc_params,
					  parent);
		if (rc)
			goto out;

		rc = do_asymm_derivation(session, proc_params, &head);
		if (rc)
			goto out;

		goto done;
	} else {
		rc = PKCS11_CKR_MECHANISM_INVALID;
		goto out;
	}

File: ta/pkcs11/src/processing_ec.c [hunk 3]
```diff
@@ -646,3 +646,95 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
 		return 0;
 	}
 }
+
+enum pkcs11_rc pkcs2tee_param_ecdh(struct pkcs11_attribute_head *proc_params,
+				   void **pub_data, size_t *pub_size)
+{
+	struct serialargs args;
```
etienne-lms, line 653: prefer `= { }` here at remove `TEE_MemFill()` at line 657.

ta/pkcs11/src/processing_ec.c
```diff
@@ -646,3 +646,95 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+	}
+
+	*pub_size -= sizeof(uint8_t);
+
+	if (*pub_size >= 0x80) {
+		EMSG("DER long definitive form not yet supported");
+		return PKCS11_CKR_ARGUMENTS_BAD;
+	}
+
+	return serialargs_get_ptr(&args, pub_data, *pub_size);
```
etienne-lms, line 697: perfer be strict and end the function with:
```c
	if (serialargs_remaining_bytes(&args))
		return PKCS11_CKR_ARGUMENTS_BAD;
```

Ditto for function `pkcs2tee_algo_ecdh()` below.

etienne-lms (commented):
some minor comments

ta/pkcs11/src/processing_ec.c
```diff
@@ -646,3 +646,103 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+	struct serialargs args = { };
+	uint32_t kdf = 0;
+
+	serialargs_init(&args, proc_params->data, proc_params->size);
+
+	rc = serialargs_get(&args, &kdf, sizeof(uint32_t));
+	if (rc)
+		return rc;
+
+	/* Remaining arguments are extracted by pkcs2tee_ecdh_param_pub() */
```
etienne-lms, line 719: s/pkcs2tee_ecdh_param_pub/pkcs2tee_param_ecdh/

ta/pkcs11/src/processing_ec.c
```diff
@@ -646,3 +646,103 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+		break;
+	case 521:
+		*tee_id = TEE_ALG_ECDH_P521;
+		break;
+	default:
+		TEE_Panic(0);
+		break;
+	}
+
+	if (serialargs_remaining_bytes(&args))
```
etienne-lms, line 744: Sorry, I confused you in https://github.com/OP-TEE/optee_os/pull/5130#discussion_r789744788. Should not check `serialargs_remaining_bytes()` here as we're expecting more arguments (as stated at line 719).

ta/pkcs11/src/processing_asymm.c
```diff
@@ -366,6 +383,8 @@ init_tee_operation(struct pkcs11_session *session,
 	case PKCS11_CKM_RSA_PKCS_OAEP:
 		rc = pkcs2tee_proc_params_rsa_oaep(proc, proc_params);
 		break;
+	case PKCS11_CKM_ECDH1_DERIVE:
+	case PKCS11_CKM_ECDH1_COFACTOR_DERIVE:
```
etienne-lms, line 387: no need to explicit these I think.

ta/pkcs11/src/processing_asymm.c
```diff
@@ -125,6 +129,9 @@ pkcs2tee_algorithm(uint32_t *tee_id, uint32_t *tee_hash_id,
 	case PKCS11_CKM_ECDSA_SHA512:
 		rc = pkcs2tee_algo_ecdsa(tee_id, proc_params, obj);
 		break;
+	case PKCS11_CKM_ECDH1_DERIVE:
```
etienne-lms, line 132: Missing  `case PKCS11_CKM_ECDH1_COFACTOR_DERIVE:` here to support the mechanism.
Note GP TEE Internal API only consider the cofactor deriveration scheme, so both `PKCS11_CKM_ECDH1_DERIVE` and PKCS11_CKM_ECDH1_COFACTOR_DERIVE` end in a cofactor derivation scheme (IIUC).

ta/pkcs11/src/processing_ec.c
ldts: um, is then something incorrect lurking? because I retested with these changed and it was all good.

ldts: > um, is then something incorrect lurking? because I retested with these changes and it was all good - ie, the key derivation suceeded.

hold on, let me make sure I updated the correct TA. I think I didnt

ldts: right, so yes, this breaks so it makes sense. ok, fixing and reposting

ldts:
retested

ta/pkcs11/src/processing_ec.c
```diff
@@ -646,3 +646,100 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+	uint32_t kdf = 0;
+
+	serialargs_init(&args, proc_params->data, proc_params->size);
+
+	rc = serialargs_get(&args, &kdf, sizeof(uint32_t));
+	if (rc)
+		return rc;
+
+	/* Remaining arguments are extracted by pkcs2tee_param_ecdh */
+	if (kdf != PKCS11_CKD_NULL)
```
etienne-lms, line 720: A last comment: maybe  worth to add a debug trace stating this implementation only supports CKD_NULL derivation, known the PKCS#11 specification also defines CKD_SHA1_KDF scheme.
E.g.:
```c
	if (kdf != PKCS11_CKD_NULL) {
		DMSG("Only support CKD_NULL key derivation for ECDH");
		return PKCS11_CKR_MECHANISM_PARAM_INVALID;
	}

**Review by etienne-lms — COMMENTED** [review]
with a last suggestion (see review  comment) and some rewording of the commit log. I would remove the sentence:
> This commit integrates previous work done by Etienne Carriere and Vesa Jääskeläinen.

I don't think you really need to mention us. If you prefer to do so, a `Cc:` tag is fine to me (unless @vesajaaskelainen has some preferences).
Maybe worth also mentioning that CKD_SHA1_KDF is intentionally not supported since X9.63 deprecation.

File: ta/pkcs11/src/processing_ec.c (hunk 9 above)
Comment (ldts) [reply ta/pkcs11/src/processing_ec.c:720]: Sure I'll add that comment. The thing is that I didn't have to do much on this commit - it only took me a few hours to re-assemble the pieces you and Vesa did and then just validated them. 
I'll wait for @vesajaaskelainen  comment on the tag (I dont think we use Co-developed-by in OP-TEE? should I do that instead?)

Comment (jforissier) [reply ta/pkcs11/src/processing_ec.c:720]: > (I dont think we use Co-developed-by in OP-TEE? should I do that instead?)

`Co-developed-by:` has been used occasionally, it's a good way to acknowledge other people's work in my opinion.
```
$ git log --oneline --grep=Co-developed-by | wc -l
35
```

Comment (ldts) [reply ta/pkcs11/src/processing_ec.c:720]: ah yeah, ok, let me do that instead then.

**Comment by ldts** [comment]
> Maybe worth also mentioning that CKD_SHA1_KDF is intentionally not supported since X9.63 deprecation.

I am sorry @etienne-lms  I cant find where this has been announced. So are all the KFDs been deprecated?  ie

```
CKD_SHA1_KDF
CKD_SHA224_KDF
CKD_SHA256_KDF
CKD_SHA384_KDF
CKD_SHA512_KDF
```

**Comment by etienne-lms** [comment]
The pkcs11 TA implementation is based on pkcs11 v2.40-e01 [1] that defines IDs for CKD_SHAx_KDF but from all of them only mentions AINSI X9.63 as reference which (IIUC) only refers to SHA1. It's not clear how other SHA's should be used.
Maybe referring SHA1_KDF deprecation is not really  usefull and commit could only mention that no CKD_SHAx_KDF schemes are implemented.

[1] http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/pkcs11-curr-v2.40.html

File: ta/pkcs11/src/processing_ec.c (hunk 9 above)
Comment (etienne-lms) [reply ta/pkcs11/src/processing_ec.c:720]: As for me, I let you choose between either
```
Co-developed-by: Etienne Carriere <etienne.carriere@linaro.org>
```
or
```
```
at your will. But note that both are somewhat exclusive.

**Comment by vesajaaskelainen** [comment]
I'll try to check this out -- but did we have tests for this? At least they were not linked in.

**Comment by ldts** [comment]
> I'll try to check this out -- but did we have tests for this? At least they were not linked in.

I only validated with pkcs11_tool - didnt write a test for it.

File: ta/pkcs11/src/processing_ec.c (hunk 9 above)
Comment (ldts) [reply ta/pkcs11/src/processing_ec.c]: ok, so I'll add you as reviewer and @vesajaaskelainen as Co-developed-by (unless he prefers the review tag as well).

File: ta/pkcs11/src/processing_asymm.c (hunk 8 above)
Comment (vesajaaskelainen) [reply ta/pkcs11/src/processing_asymm.c:132]: I suppose TEE internal API method is defined with:
```
TEE_ALG_ECDH_DERIVE_SHARED_SECRET - NIST SP800-56A, Cofactor Static Unified Model
```

When I have a look at said specification there is following defined:
- 5.7.1.1 Finite Field Cryptography Diffie-Hellman (FFC DH) Primitive
- 5.7.1.2 Elliptic Curve Cryptography Cofactor Diffie-Hellman (ECC CDH) 
Primitive
- 5.7.2.1 Finite Field Cryptography MQV (FFC MQV) Primitive
- 5.7.2.2 ECC MQV Associate Value Function
- 5.7.2.3 Elliptic Curve Cryptography MQV (ECC MQV) Primitive

Then we have in PKCS#11:

- 2.3.9 Elliptic curve Diffie-Hellman key derivation (CKM_ECDH1_DERIVE)
- 2.3.10 Elliptic curve Diffie-Hellman with cofactor key derivation (CKM_ECDH1_COFACTOR_DERIVE)
- 2.3.11 Elliptic curve Menezes-Qu-Vanstone key derivation (CKM_ECMQV_DERIVE)

These all refer to ANSI X9.63.

I am not in anyway a math expert but just looking at those documents I would say that:
- TEE_ALG_ECDH_DERIVE_SHARED_SECRET would match to NIST SP800-56A/5.7.1.2 
- and that would match to CKM_ECDH1_COFACTOR_DERIVE.

Also looking at different variants described in NIST SP800-56A I would say that 5.7.1.1 and 5.7.1.2 are different methods.

That being said I believe CKM_ECDH1_DERIVE would match to NIST SP800-56A/5.7.1.1.

Now this code in here assumes that both CKM_ECDH1_DERIVE and CKM_ECDH1_COFACTOR_DERIVE are calculated the same.

Now I may be missing something but to me that is wrong.

Comment (vesajaaskelainen) [reply ta/pkcs11/src/processing_asymm.c:132]: If I understood correctly the cofactor comes from domain parameters -- eg. from curve:
```
$ openssl ecparam -noout -text -param_enc explicit -name P-192
...
Cofactor:  1 (0x1)
...
$ openssl ecparam -noout -text -param_enc explicit -name P-224
...
Cofactor:  1 (0x1)
...
$ openssl ecparam -noout -text -param_enc explicit -name P-256
...
Cofactor:  1 (0x1)
...
$ openssl ecparam -noout -text -param_enc explicit -name P-384
...
Cofactor:  1 (0x1)
...
$ openssl ecparam -noout -text -param_enc explicit -name P-521
...
Cofactor:  1 (0x1)
```

So as we currently support only those curves then the `CKM_ECDH1_DERIVE` and `CKM_ECDH1_COFACTOR_DERIVE` would give same result. But when we would support some other curves then this would become a bigger issue?

**Comment by vesajaaskelainen** [comment]
> > I'll try to check this out -- but did we have tests for this? At least they were not linked in.
> 
> I only validated with pkcs11_tool - didnt write a test for it.
> 
> ```
> pkcs11-tool --module /usr/lib/libckteec.so.0.1 --keypairgen --key-type EC:prime256v1 --id 01 --label tls --token-label fio --pin 87654321
> 
> pkcs11-tool --module /usr/lib/libckteec.so.0.1 --keypairgen --key-type EC:prime256v1 --id 02 --label tls2 --token-label fio --pin 87654321
> 
> pkcs11-tool --module /usr/lib/libckteec.so.0.1 --read-object --type pubkey --id 02 --token-label fio --pin 87654321 -o /tmp/pub.der
> 
> pkcs11-tool --module /usr/lib/libckteec.so.0.1 --derive -m ECDH1-DERIVE --id 01 --label tls --token-label fio --pin 87654321 --input-file /tmp/pub.der --output-file /tmp/bytes
> ```

Now with this we do get result -- but is it correct?

I would say that cross verification with other tooling is a must.

As there are different methods for calculation I would say this requires test cases in test suite otherwise it is quite easy to get broken and it also verifies that actual method is correct. Recommend to use external crypto library (like openssl) to assist in that.

Also we should have test cases to test out handling of different forms of public key input.

**Comment by vesajaaskelainen** [comment]
> The pkcs11 TA implementation is based on pkcs11 v2.40-e01 [1] that defines IDs for CKD_SHAx_KDF but from all of them only mentions AINSI X9.63 as reference which (IIUC) only refers to SHA1. It's not clear how other SHA's should be used. Maybe referring SHA1_KDF deprecation is not really usefull and commit could only mention that no CKD_SHAx_KDF schemes are implemented.
> 
> [1] http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/pkcs11-curr-v2.40.html

Openssl wiki says:
> Note that shared secrets derived in this way may not be evenly distributed within the key space. For this reason a shared secret is typically passed through some further function such as a message digest such as SHA2 (possibly combining the shared secret with other well defined data first). Using a shared secret directly as an encryption key could lead to biases in your encryption implementation which in turn is likely to lead to security weaknesses. Refer to section 2.1.2 of RFC 2631 for an example of this in practice.

https://wiki.openssl.org/index.php/EVP_Key_Agreement

https://datatracker.ietf.org/doc/html/rfc2631#section-2.1.2

Seems reasonable and seems that `kdf` and `shared data` should be used in practice.

**Comment by vesajaaskelainen** [comment]
Oh... please rebase the PR on the master. Cherry picked the commit to try it out.

File: ta/pkcs11/src/processing_asymm.c (hunk 8 above)
Comment (vesajaaskelainen) [reply ta/pkcs11/src/processing_asymm.c:132]: OP-TEE's both libtomcrypt and mbedtls variants seems to be implementing `CKM_ECDH1_DERIVE` without cofactor and not the one with cofactor as one would believe based on Global Platform spec.

File: ta/pkcs11/src/processing_ec.c [hunk 10]
```diff
@@ -646,3 +646,102 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+	*pub_size = temp;
+
+	rc = serialargs_get(&args, &temp, sizeof(uint8_t));
+	if (rc)
+		return rc;
+
+	if (temp != 0x02 && temp != 0x03 && temp != 0x04)
+		return PKCS11_CKR_ARGUMENTS_BAD;
+
+	if (temp != 04) {
```
vesajaaskelainen, line 683: something fishy in here. `temp` is the same as what was above `if` and here we have octal 04 which is same as hex 0x04 above.

Please see `tee2pkcs_ec_attributes()` for an idea how DER encoded form has compression specified.

Also:
> A token MUST be able to accept this value encoded as a raw octet string (as per section A.5.2 of [ANSI X9.62]).  A token MAY, in addition, support accepting this value as a DER-encoded ECPoint (as per section E.6 of [ANSI X9.62]) i.e. the same as a CKA_EC_POINT encoding.  The calling application is responsible for converting the offered public key to the compressed or uncompressed forms of these encodings if the token does not support the offered form.

ta/pkcs11/src/processing_ec.c
```diff
@@ -646,3 +646,102 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+	if (rc || !temp)
+		return rc;
+
+	*pub_size = temp;
+
+	rc = serialargs_get(&args, &temp, sizeof(uint8_t));
+	if (rc)
+		return rc;
+
+	if (temp != 0x02 && temp != 0x03 && temp != 0x04)
```
vesajaaskelainen, line 680: I suppose this should either be raw data or octet string (0x04).

Not integer (0x02) or bit string (0x03).

I suppose doing this properly would require some heuristics as raw vs. der is not given as argument which would have been much nicer approach.

ldts:
> > > I'll try to check this out -- but did we have tests for this? At least they were not linked in.
> > 
> > 
> > I only validated with pkcs11_tool - didnt write a test for it.
> > ```
> > pkcs11-tool --module /usr/lib/libckteec.so.0.1 --keypairgen --key-type EC:prime256v1 --id 01 --label tls --token-label fio --pin 87654321
> > 
> > pkcs11-tool --module /usr/lib/libckteec.so.0.1 --keypairgen --key-type EC:prime256v1 --id 02 --label tls2 --token-label fio --pin 87654321
> > 
> > pkcs11-tool --module /usr/lib/libckteec.so.0.1 --read-object --type pubkey --id 02 --token-label fio --pin 87654321 -o /tmp/pub.der
> > 
> > pkcs11-tool --module /usr/lib/libckteec.so.0.1 --derive -m ECDH1-DERIVE --id 01 --label tls --token-label fio --pin 87654321 --input-file /tmp/pub.der --output-file /tmp/bytes
> > ```
> 
> Now with this we do get result -- but is it correct?

I did use the snippet above on several curves (prime256v1, secp224r1, prime192v1..) using the SE051 and the shared secret for alice and bob was always matching - the code also generated the expected number of bits.  IMO it is just unlikely that the result is incorrect

> 
> I would say that cross verification with other tooling is a must.

ah ok.

> 
> As there are different methods for calculation I would say this requires test cases in test suite otherwise it is quite easy to get broken and it also verifies that actual method is correct. Recommend to use external crypto library (like openssl) to assist in that.
> 
> Also we should have test cases to test out handling of different forms of public key input.

ok

ta/pkcs11/src/processing_ec.c
ldts: right, the check should have been temp!= 0x04

ta/pkcs11/src/processing_ec.c
```diff
@@ -646,3 +646,102 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
...
+		return rc;
+
+	/* Public data size and content */
+	rc = serialargs_get(&args, &temp, sizeof(uint32_t));
+	if (rc || !temp)
+		return rc;
+
+	*pub_size = temp;
+
+	rc = serialargs_get(&args, &temp, sizeof(uint8_t));
```
jenswi-linaro, line 676: `temp` is a `uint32_t`, you should probably use some temporary `uint8_t` variable instead.

ldts:
@etienne-lms did you do any tests I could reuse back when you did your implementation?

ta/pkcs11/src/processing_ec.c
ldts: sorry I dont quite understand the recommendation. you are asking the checks for 0x02 and 0x03 to be removed?

ldts:
@vesajaaskelainen I have removed the codeveloped tags I added for you to avoid confusion during the review. will add them back once I get your final ack.

etienne-lms:
> @etienne-lms did you do any tests I could reuse back when you did your implementation?

At the time, i was running a test based on vectors and sequence used in existing `xtest_tee_test_4009()`.
It injects a EC keypair, derives a generic secret and checks the secret value.
The test is still available in my branch [sks-to-pkcs](https://github.com/etienne-lms/optee_test/tree/sks-to-pkcs11). It was named [`regression_4218`](https://github.com/etienne-lms/optee_test/blob/sks-to-pkcs11/host/xtest/regression_4200.c#L2507-L2508), implemented by function `run_xtest_tee_test_4218()` (after few redirections), see [regression_4000.c here](https://github.com/etienne-lms/optee_test/blob/d7deefe021610b4113eaf8ec50d48058fbcae2d7/host/xtest/regression_4000.c#L6890-L6900).

ldts:
> > @etienne-lms did you do any tests I could reuse back when you did your implementation?
> 
> At the time, i was running a test based on vectors and sequence used in existing `xtest_tee_test_4009()`. It injects a EC keypair, derives a generic secret and checks the secret value. The test is still available in my branch [sks-to-pkcs](https://github.com/etienne-lms/optee_test/tree/sks-to-pkcs11). It was named [`regression_4218`](https://github.com/etienne-lms/optee_test/blob/sks-to-pkcs11/host/xtest/regression_4200.c#L2507-L2508), implemented by function `run_xtest_tee_test_4218()` (after few redirections), see [regression_4000.c here](https://github.com/etienne-lms/optee_test/blob/d7deefe021610b4113eaf8ec50d48058fbcae2d7/host/xtest/regression_4000.c#L6890-L6900).

thanks a lot @etienne-lms . I'll try spinning something out of this so this PR can  be merged.

ldts:
@etienne-lms @vesajaaskelainen finally I got around to doing this. will send it later today for review.

ldts:
https://github.com/OP-TEE/optee_test/pull/587

ldts:
@jforissier @etienne-lms @vesajaaskelainen what is pending at this point?

ta/pkcs11/src/processing_asymm.c
jforissier: OP-TEE currently implements only the NIST curves (`TEE_ECC_CURVE_NIST_P{192,224,256,384,521}`) with cofactor 1. SM2 is also supported but only for its specific key exchange protocol (SM2 KEP) and signature algorithm (SM2 DSA).

You can find ECDH tests in `xtest 4009`: see [here](https://github.com/OP-TEE/optee_test/blob/3.16.0/host/xtest/regression_4000.c#L4793) and [here](https://github.com/OP-TEE/optee_test/blob/3.16.0/host/xtest/regression_4000_data.h#L6481).

ldts:
should I close this PR? hard to know where we stand when nobody is following up

ldts:
@ricardosalveti maybe we should continue carrying this patch in our repos?

vesajaaskelainen:
As far I can say problem here is as discussed above is: `CKM_ECDH1_DERIVE` vs `CKM_ECDH1_COFACTOR_DERIVE`.

I don't think we can say that we support `CKM_ECDH1_COFACTOR_DERIVE` as OP-TEE is not doing cofactor calculation in its implementation.

ldts:
@vesajaaskelainen  @etienne-lms OP-TEE does not support curves with a cofactor different than 1 AFICS. Am I wrong?

```
/* List of Supported ECC Curves */
#define TEE_CRYPTO_ELEMENT_NONE             0x00000000
#define TEE_ECC_CURVE_NIST_P192             0x00000001
#define TEE_ECC_CURVE_NIST_P224             0x00000002
#define TEE_ECC_CURVE_NIST_P256             0x00000003
#define TEE_ECC_CURVE_NIST_P384             0x00000004
#define TEE_ECC_CURVE_NIST_P521             0x00000005
#define TEE_ECC_CURVE_SM2   
```

So should the issue you are raising really be preventing this PR to be merged? It seems more like an extension to me.

ldts:
Although even if it OP-TEE supported curves with cofactors 4 or 8, wouldn't it be better to have something supporting 99% of the use cases than none?

vesajaaskelainen:
Did a quick survey for some crypto libraries and how do they handle ECDH.

| library | ecdh | ecdh w/cofactor | notes |
|-------|----|----|---|
| openssl  | yes | yes | flag used to select whether cofactor is taken in use |
| wolfcrypt | yes | yes | flag used to select whether cofactor is take in use |
| bouncy castle | yes | yes | different class |
| mbedtls | yes | no | cofactor not really used for much |
| libtomcrypt | yes | no | cofactor not really used for much |
| crypto++ | yes | no | cofactor not really used for much |

OP-TEE has `libtomcrypt`, `mbedtls` **and** other crypto providers which I cannot say how they function.

Some comments on topics comment also that it is not well documented in standards what is being used.

If I understood from security point of view the cofactor should be preferred to match curve properties -- but we do not have that support in our crypto libraries.

Based on current implementation in OP-TEE -- I would keep `CKM_ECDH1_DERIVE` and drop `CKM_ECDH1_COFACTOR_DERIVE` as we do not do proper math for cofactor and it will become a hidden problem when someone adds support for new curves specified in new GP spec. Alternative for this would be that we have filtered list of curves that support "transparently" both each derive method so that new curves would need to be added to the lists then with that we do not get surprises (eg. the NIST curves have cofactor 1 so both can be transparently supported).

Lingering problem here is what GP spec expects for key derivation. When Edwards curves are added they should calculate ECDH with the cofactor. @jforissier could you get clarification from GP on how to handle the ECDH with/without cofactor (and what it currently means) as I suppose it would mean extending the GP spec (and may also affect OP-TEE's internal cryptoapi).

jforissier:
@vesajaaskelainen GP supports Edwards curve ED25519 since version 1.2, but that support is optional and as you said OP-TEE doesn't implement it. 1.3 adds ED448. In any case, in the latest GP spec the word "cofactor" doesn't occur. So my understanding is that the cofactor is a fixed value for a given curve and therefore doesn't need to appear as a parameter anywhere at the GP API level.

etienne-lms:
Thanks @vesajaaskelainen and @ldts for digging into this issue.

> OP-TEE has libtomcrypt, mbedtls and other crypto providers which I cannot say how they function.

As I understand it, I agree with @vesajaaskelainen. We should remove support for `CKM_ECDH1_COFACTOR_DERIVE` from this P-R since not supported by the reference crypto libs embedded in OP-TEE.

> So my understanding is that the cofactor is a fixed value for a given curve and therefore doesn't need to appear as a parameter anywhere at the GP API level.

For what I can say, the PKCS#11 specifically neither defines APIs to pass a co-factor parameter, it is a defacto agreeed value per curve definition.

ldts:
thanks @vesajaaskelainen . Right, so  updated the patch removing that operation - just so we can continue with the discussion. Incidentally the SE05x supports some of those curves (ie ED25519) not supported by OP-TEE....so I do have a way of testing in the future.

ldts:
@jforissier @vesajaaskelainen @etienne-lms so what is next?

vesajaaskelainen:
> > So my understanding is that the cofactor is a fixed value for a given curve and therefore doesn't need to appear as a parameter anywhere at the GP API level.
> 
> For what I can say, the PKCS#11 specifically neither defines APIs to pass a co-factor parameter, it is a defacto agreeed value per curve definition.

In order to be able to select whether to use cofactor or not I suppose there would need to be similar feature like openssl/wolfcrypt are doing with configurable mode flag. But that kinda looks like it can either be:
- Official change in GP spec
- Local extension

I would prefer GP spec change but someone part for the GP process would need to work on that.

In order to progress on that you probably would need good real life examples where such feature would be needed. There was some indication that some PIV cards are using cofactor mode (and only that):

ref: https://github.com/OpenSC/OpenSC/issues/1609#issuecomment-493557343
> NIST 800-73-4 says in part 2:
> "A.5.2.1 ECDH with the PIV KMK
> The PIV Card supports ECDH key agreement by performing the elliptic curve cryptography cofactor Diffie-Hellman (ECC CDH) > primitive [SP800-56A, Section 5.7.1.2] using its ECC KMK private key and an ECC public key that is provided as input to the
> GENERAL AUTHENTICATE command. All other procedures required to complete key agreement are performed by the
> cardholder’s client application and its associated cryptographic module."

Then the next step would be to figure out what to do with math libraries in OP-TEE... mbedtls seem to be dropping out legacy crypto support code in new major release that could hit us.

Libtcomcrypt is also missing the feature so I suppose one would need to contribute it in there before we get it in OP-TEE.

Then we have a question about other crypto backends that might have an issue with it (or they might actually be doing always cofactor part).

vesajaaskelainen:
> thanks @vesajaaskelainen . Right, so updated the patch removing that operation - just so we can continue with the discussion. Incidentally the SE05x supports some of those curves (ie ED25519) not supported by OP-TEE....so I do have a way of testing in the future.

Thanks.

About Edwards curves. I suppose we need to first have that implemented in software crypto libs and exposed thru GP API then cross testing with external crypto provides would be good. GP API major upgrade would be a bit bigger job but some specific cryptos could perhaps be introduced earlier.

But lets leave that out of this PR.

vesajaaskelainen:
> @jforissier @vesajaaskelainen @etienne-lms so what is next?

Checked out the comments above and I suppose only hindering issue was the public key encoding. Just wondering if that is best left out of this PR and go with the current feature set. Public key encoding would need other refactoring and same code should be used in more than one place.

> ok, so I'll add you as reviewer and @vesajaaskelainen as Co-developed-by (unless he prefers the review tag as well).

You can have me as Co-developed-by as proposed.

And if @etienne-lms is happy still with the reviewed-by then I suppose we are go.

etienne-lms:
Hi, sorry not contributing much here. My R-b still applies.
Thanks for looking into the cofactor handling and i'm fine with the proposed P-R, excluding cofactor.
There is just the [test](https://github.com/OP-TEE/optee_test/pull/587) part I'm not found of. Wouldn't it better better to test upon some reference vectors, as those used for regression tests (`derive_key_ecdh_t`)?

ldts:
@etienne-lms RE tests I still dont understand why...testing against a reference vector offers no better guarantee that the implementation is right ....seems like a aesthetic choice to me. where as the test I propose its provides a better functional reference (ie, is a real-world use case). but in any case, your call of course

ldts:
just added tags

PR #5131, opened 2022-01-20
Add support for i.MX8ULP

clementfaure:


core/drivers/crypto/caam/hal/imx_8ulp/hal_jr.c
```diff
@@ -0,0 +1,89 @@
...
+	HAL_TRACE("JR%" PRIu32 "DID_MS value 0x%" PRIx32, jr_idx, val);
+
+	/* Prepare the Job Ring MS/LS registers */
+	if (owner & JROWNER_SECURE) {
+		/* Configuration only locked for the Secure JR */
+		cfg_ms = JRxDID_MS_PRIM_DID(owner & ~JROWNER_SECURE);
+		cfg_ms |= JRxDID_MS_PRIM_TZ | JRxDID_MS_TZ_OWN;
+		cfg_ms |= JRxDID_MS_AMTD;
+		cfg_ms |= JRxDID_MS_PRIM_ICID(owner & ~JROWNER_SECURE);
+#ifdef CFG_NXP_CAAM_RUNTIME_JR
```
jforissier, line 43:
```C
#include <config.h>
...
if IS_ENABLED(xx) {
        ...
}
```
...is always preferred when possible. In this case, I think it is.

jforissier (commented):
One minor comment, other than that looks good!

jforissier:
@clementfaure Jens is about to create the release tag for 3.16.0, so if you'd like this to make it into the release, please apply my Ack and I will merge immediately. Thanks!

clementfaure:
Tags applied and comment addressed! Hoping it is not too late ...

PR #5132, opened 2022-01-20
core: use a paged scattered array for DT drivers list

etienne-lms:
Fixes macro DEFINE_DT_DRIVER() to use SCATTERED_ARRAY_DEFINE_PG_ITEM()
that defines a paged list whereas SCATTERED_ARRAY_DEFINE_ITEM()
assumes array must be linked in the rodata_init section. Indeed
DT driver probing is done after pager is initialized.

Fixes: 61bdedea9452 ("core: define DT drivers using scattered arrays")

### DIFF
--- a/core/include/kernel/dt.h
+++ b/core/include/kernel/dt.h
@@ -103,7 +103,7 @@ struct dt_driver {
 };
 
 #define DEFINE_DT_DRIVER(name) \
-		SCATTERED_ARRAY_DEFINE_ITEM(dt_drivers, struct dt_driver)
+		SCATTERED_ARRAY_DEFINE_PG_ITEM(dt_drivers, struct dt_driver)
 
 /*
  * Find a driver that is suitable for the given DT node, that is, with


etienne-lms:
I agree with this change, `scatter_array_dt_drivers_...` is still in the `rodata_init` section but not what is references (and all their related dependencies) will go in pageable sections which is is main issue this P-R wants to address.
I think it is not a major issue that the scatter array itself is in the init section. Such array should be quite small (no thousands of device drivers registered).

jenswi-linaro:

jforissier:
@etienne-lms OK, I believe I have a bad understanding of the pager "init" sections :-/ Can you please clarify 3 things:
- Does the data in `.rodata_init` remain accessible after pager is initialized, and during the whole life of the TEE?
- If so, is `.rodata_init` pageable or unpaged?
- Since the array items (i.e., the structures passed to `SCATTERED_ARRAY_DEFINE_ITEM()` or `SCATTERED_ARRAY_DEFINE_PG_ITEM()`)  are still in `.rodata_init` even after this change, I think this sentence in the description is misleading: "SCATTERED_ARRAY_DEFINE_ITEM() assumes array must be linked in the rodata_init section", wouldn't it be better to say "SCATTERED_ARRAY_DEFINE_ITEM() assumes array **and its dependencies** must be linked in the rodata_init section"?

etienne-lms:
Hi @jforissier,
`.rodata_init` and `.text_init`  sections are fully pageable. The sole difference is that such page are pre-loaded at boot-time.
They remain loaded until the pager is initialized running (and decide to unload such a page).
The init sections is designed for resources that don't need to be resident in secure memory at runtime need need to be resident during early inits of the core, before mmu and pager are initialized.

>  (...) I think this sentence in the description is misleading: (...)

Ok, i'll fix the commit log accordingly.

jforissier:

etienne-lms:
Commit log updated.

jforissier:
@etienne-lms thanks for the clarification. Now I get it.

etienne-lms:
tags applied

PR #5134, opened 2022-01-21
core: crypto: DSA key size is given by TEE_ATTR_DSA_PRIME

jforissier:
The size of a DSA key is the size of the prime number (p) and not the
size of the subprime (q). Therefore the size indicator flag in the
definition of DSA cryptographic object attributes should be associated
with TEE_ATTR_DSA_PRIME, not with TEE_ATTR_DSA_SUBPRIME.


jenswi-linaro:

etienne-lms (commented):

PR #5136, opened 2022-01-24
crypto: drivers: se050: remove implicit dependency

ldts:
The SE05X device is platform independent and therefore does not need
the iMX I2C driver but the actual driver for the particular platform
is connected into.

Implementing these changes required a fix in the Plug-and-Trust tree
(the addition of a missing dependency), therefore we will also bump
the Plug-and-Trust version used in the Azure pipeline.


jforissier (commented):

ldts:
@jforissier all good?

jforissier:
@ldts yes, thanks.

PR #5140, opened 2022-01-25
libutee: Update handle-state after TEE_AllocateOperation()

Sadiq1811:
The TEE internal core API specification mentions the
following in the description of TEE_GetOperationInfo():
"TEE_HANDLE_FLAG_INITIALIZED: Set for multi-stage
operations and for Digest operations"

TEE_AllocateOperation() is the first API which needs
to be called before calling other crypto APIs. Hence,
the handle-state should be set to
TEE_HANDLE_FLAG_INITIALIZED atleast (excluding
asymmetric operations) after invoking
TEE_AllocateOperation().

Signed-off-by: Sadiq Hussain <sadiq.muchumarri@intel.com>

### DIFF
--- a/lib/libutee/tee_api_operations.c
+++ b/lib/libutee/tee_api_operations.c
@@ -44,8 +44,49 @@ TEE_Result TEE_AllocateOperation(TEE_OperationHandle *operation,
 	if (!operation)
 		TEE_Panic(0);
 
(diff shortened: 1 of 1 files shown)


Sadiq1811:
Changes are not needed.
Hence, closing this PR.

PR #4277, opened 2020-12-16
core: reserve devices for exclusive OP-TEE usage

ldts:
Use the device tree to reserve devices away from the normal world. The
devices should be **comma separated strings** identifying the dt nodes to
reserve.

For example (a single driver in imx6ull would be a single string):
 CFG_CORE_RESERVE_DRIVERS="\"/soc/aips-bus@2200000/crypto@2280000\""

This initial implementation has strtok in boot.c for simplicity (clearly it doesn't belong there). 

If there is interest in having this feature merged in the project, please could you suggest a place where to locate the strtok function and I'll repost?

PR #5143, opened 2022-01-26
Add JR device tree overlay support for imx8

clementfaure:
Hello!

This patch set allows the CAAM driver to set some JR status to "disabled" through the device tree overlay.

We've put the `dt_overlay_disable_node()` in `kernel/boot.c` thinking that would be leveraged by other drivers. Let me know if that works for you.

Thanks,
Clement


core/drivers/crypto/caam/hal/common/hal_cfg.c
```diff
@@ -54,11 +54,10 @@ enum caam_status caam_hal_cfg_get_conf(struct caam_jrcfg *jrcfg)
...
 		if (IS_ENABLED(CFG_NXP_CAAM_RUNTIME_JR) &&
-		    !is_embedded_dt(fdt)) {
-			if (fdt) {
-				/* Ensure Secure Job Ring is secure in DTB */
-				caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
-			}
+		    IS_ENABLED(CFG_CAAM_JR_DISABLE_NODE) &&
+		    fdt && !is_embedded_dt(fdt)) {
+			/* Ensure Secure Job Ring is secure in DTB */
+			caam_hal_cfg_disable_jobring_dt(fdt, jrcfg);
```
jforissier, line 60: Unrelated change, should be a separate commit

core/arch/arm/kernel/boot.c
```diff
@@ -671,6 +671,42 @@ static int init_dt_overlay(struct dt_descriptor *dt, int __maybe_unused dt_size)
...
+
+	offs = fdt_add_subnode(dt->blob, offs, frag);
+	if (offs < 0)
+		return offs;
+
+	dt->frag_id += 1;
+
+	ret = fdt_setprop_string(dt->blob, offs, "target-path", target);
+	if (ret < 0)
+		return -1;
```
jforissier, line 698: `return res;`?

jforissier (commented):
Hi @clementfaure, a few minor things below. Note, in commit "drivers: caam: disable job ring through fdt overlay": I think 'via' would be more correct than 'through' (and typo: 'thought'). And maybe also s/fdt/DT/.

core/arch/arm/kernel/boot.c
clementfaure: Ok

clementfaure:
Since you gave me your `Acked-by` I squashed my changes

ldts:
@Emantor some months ago I tried to upstream a feature that allowed the platform to explicitly reserve drivers for exclusive TEE access.https://github.com/OP-TEE/optee_os/pull/4277 
This was rejected because the DT paths werent considered stable back then.

Looking at this PR it seems this has now changed. Please could you comment as to why the paths werent stable then and why they are now?

jforissier:
> however this was rejected upstream because the DT paths werent considered stable back then.

Good point, but is there a better option? Commit "drivers: caam: disable job ring via DT overlay" mentions that "On imx8m platforms, OP-TEE has no direct access to the Linux device tree", so I suppose the DT node path cannot be created dynamically?

clementfaure:
For imx8 platforms (SPL->ATF->OPTEE->Uboot), OPTEE has no access to the Linux device tree as it is not loaded in memory yet. There is no way for OPTEE to parse the Linux device and look for compatible strings and eventually modify the node.
I believe the only way to modify subnodes from OTPEE via the device tree overlay is to work with hardcoded paths.

ldts:
but isnt that just one of the possible boot scenarios? and in that case, why can't the FIT image handled by SPL contain the Linux device tree and this be passed to OP-TEE?

ldts:
if it helps I might have somewhere the changes required to pass a string of data from SPL to  TF-A and from there to OP-TEE. Had to do those for my initial Xilinx DNA investigation. if it helps.

clementfaure:
I agree with you that device trees are a bit tricky to deal with and require some maintenance to ensure a semblance of consistency between all the boot components.
However, working with device tree overlays seems to be reasonable to me. The overlay could be in the FIT image or somewhere else, from an OPTEE point of view, it does not really matter as the overlay address is passed by ATF to OPTEE.

clementfaure:
I don't know what to say. Looks like OPTEE support device tree overlays and to my knowledge, hardcoded paths are the only way to build overlays. That's not ideal, I know.
However, I would certainly prefer working with compatible strings whenever the Linux device tree is available to OPTEE.

ldts:
I think @Emantor had a proposal but I never saw the details, something along the lines of 

`requiring an include of the __symbols__ node for DT overlays and then using the phandles.`

clementfaure:
@Emantor @ldts Do you have something to share with us so we can go forward with this pull-request ?

ldts:
Sorry no, I haven't been working on this any further @clementfaure 

From the discussion back then when I implemented #4277 @jenswi-linaro did agree with @Emantor 's proposal (instead of hardcoding DT paths) so I'd like to get his input on this now.

jenswi-linaro:
> From the discussion back then when I implemented #4277 @jenswi-linaro did agree with @Emantor 's proposal (instead of hardcoding DT paths) so I'd like to get his input on this now.

I gave a thumbs up, indicating that I thought it an important point to consider.
I currently don't have any input for this.

ldts:
> > From the discussion back then when I implemented #4277 @jenswi-linaro did agree with @Emantor 's proposal (instead of hardcoding DT paths) so I'd like to get his input on this now.
> 
> I gave a thumbs up, indicating that I thought it an important point to consider. I currently don't have any input for this.

if this is not an issue for anyone then I think we are all good then - we opted for carrying over patches similar to this one since we worked under the understanding that OP-TEE's direction was different. Maybe that is no longer the case.

ldts:
@jforissier shall I resend https://github.com/OP-TEE/optee_os/pull/4277 appending the DT __symbols__ ?
it should also cover the use case this PR is addressing.

ldts:
Ok I'll have a go, In the case that what I am submitting is not merged,  we can always get the symbols stuff as a separate PR and merge it with this one...so I think we should keep this open.

ldts:
I have several patches on the fly, nothing complex, just normal stuff
1. SPL - pass DTB to TF-A and build dtb with __symbols__
2. TF-A: I dont believe it requires RESET_TO_BL31 (I think it is a bug which is what is causing all the pain here), pass DTB to OP-TEE.
3. handle a DTB on params and generate an overlay on a CFG address.
4. work on the overaly [next]

will post something next week.

ldts:
https://github.com/OP-TEE/optee_os/pull/5143

ldts:
About hard-coding paths, u-boot is a _minefield_... I wonder if this is correct?
https://github.com/u-boot/u-boot/blob/master/arch/arm/mach-imx/imx8m/soc.c#L830

```
int disable_vpu_nodes(void *blob)
{
	const char *nodes_path_8mq[] = {
		"/vpu@38300000",
		"/soc@0/vpu@38300000"
	};

	const char *nodes_path_8mm[] = {
		"/vpu_g1@38300000",
		"/vpu_g2@38310000",
		"/vpu_h1@38320000"
	};

	const char *nodes_path_8mp[] = {
		"/vpu_g1@38300000",
		"/vpu_g2@38310000",
		"/vpu_vc8000e@38320000"
	};

	if (is_imx8mq())
		return disable_fdt_nodes(blob, nodes_path_8mq, ARRAY_SIZE(nodes_path_8mq));
	else if (is_imx8mm())
		return disable_fdt_nodes(blob, nodes_path_8mm, ARRAY_SIZE(nodes_path_8mm));
	else if (is_imx8mp())
		return disable_fdt_nodes(blob, nodes_path_8mp, ARRAY_SIZE(nodes_path_8mp));
	else
		return -EPERM;

}

```
given that in my DTB
```
jramirez@trex repo,u-boot.build$ cat u-boot.dts | grep vpu
                vpumix-pd {
                        domain-name = "vpumix";
                vpug1-pd {
                        domain-name = "vpu_g1";
                vpug2-pd {
                        domain-name = "vpu_g2";
                vpuh1-pd {
                        domain-name = "vpu_h1";
                vpumix_pd = "/power-domains/vpumix-pd";
                vpu_g1_pd = "/power-domains/vpug1-pd";
                vpu_g2_pd = "/power-domains/vpug2-pd";
                vpu_h1_pd = "/power-domains/vpuh1-pd";
```

PR #3721, opened 2020-04-02
[Don't merge] imx: Enable prediction resistance for CAAM RNG

Emantor:
The [upstream kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/crypto/caam?id=358ba762d9f1d4ba99ab31ef12bc28014b22f4c9) merged a patch series to enable prediction resistance for all RNG state handles. Enable prediction resistance for the OP-TEE CAAM RNG handling as well.

PR #5145, opened 2022-01-26
drivers: caam: instantiate RNG state handle with prediction resistance

clementfaure:
Instantiate RNG state handles with Prediction Resistance (PR) support.
This way SW further downstream (e.g. Rich OS, boot loader etc.) is able
to use the "PR" bit in RNG generation descriptors (forcing TRNG
re-seeding before PRNG / DRBG outputs random data).

Note: current patch does not deal with RNG state handles that have
already been initialized, but without PR support (this could happen if
U-boot would run before OP-TEE etc.). In this case, RNG state handle
would have to be deinstantiated first, and then reinstantiated with
PR support.

Signed-off-by: Franck LENORMAND <franck.lenormand@nxp.com>
Signed-off-by: Horia Geantă <horia.geanta@nxp.com>

### DIFF
--- a/core/drivers/crypto/caam/include/caam_desc_defines.h
+++ b/core/drivers/crypto/caam/include/caam_desc_defines.h
@@ -412,6 +412,9 @@
 /* State Handle */
 #define ALGO_RNG_SH(sh)		SHIFT_U32((sh) & 0x3, 4)
 
+/* Prediction Resistance */
+#define ALGO_RNG_PR BIT32(1)
+
 /* State */
 #define AS_RNG_GENERATE		0x0
 #define AS_RNG_INSTANTIATE	0x1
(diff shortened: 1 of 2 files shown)


Emantor:
Doesn't the rng gen data also need a modification as done in #3721?

clementfaure:
Hi @Emantor,
This patch only enables the prediction resistance support for the initialized RNG state handles.
You can still generate new data without prediction resistance with that "prediction resistance enabled" RNG state handle

jforissier:
TBH, I don't know much about how this stuff works but the description makes sense to me and coding style is OK :wink: so:

clementfaure:
Thank you @jforissier! Rebase and tag applied.

Issue #5144, opened 2022-01-26
Xilinx Zynq MPSoC does not boot anymore with new DDR configurations


vesajaaskelainen:
Using latest versions Xilinx ZCU102 devkit (nor our own board) does not boot.

During the boot gives following error on trace:
```
E/TC:0 0 std_smc_entry:183 Bad arg address 0x803989000
```

Seems that this commit causes the issue -- if we revert this then zcu102 once again boots:
[a7474d1](https://github.com/OP-TEE/optee_os/commit/a7474d1d0b780fc4f8a93883ae3f21320b155998) - plat: zynqmp: register ddr for dyn shm support

Xilinx Zynq MPSoC does have two DDR bases:
- First 2 GiB of DRAM (DDR Low): 0x0000_0000 .. 0x8000_0000
- And then rest of memory (DDR High):  0x0008_0000_0000 .. 0x000F_FFFF_FFFF

Real memory size is either detected from DDR DIMM or hard coded in device tree. There could also be aliasing so perhaps the right course of action would be to get this detail from device tree assuming previous stages has detected the memory size or has fixed size.

ricardosalveti:
How much memory do you have in your devkit? The board I test with has 2GB, which is enough for the hardcoded values, and enough for dyn shm to work (optee needs to register the entire memory region otherwise with dyn shm the kernel might use a region that is unknown to optee and cause the error you're getting).

vesajaaskelainen:
On this ZCU102 where we did testing it had 4 GiB of memory.

ricardosalveti:
OP-TEE can also identify based on what gets provided by the device tree, can you check if the one given to optee also contain the 4GBs described there?

vesajaaskelainen:
Our current bootflow is:
FSBL->ATF->OP-TEE->u-boot...

And currently we do not have device tree in Xilinx image blob containing ATF. We will see what effect adding device tree to the Xilinx image blob has.

ricardosalveti:
Without dtb I think it worked before for you because optee was probably disabling dyn_shm. If you build with CFG_CORE_DYN_SHM=n you should get it to boot with latest rev.

To fix this properly (and enable dyn_shm) you can either provide a dtb or extend plat-zynqmp/main.c to register DRAM1_BASE (via another CFG), as done with other targets, e.g.:

```
core/arch/arm/plat-stm/main.c:register_ddr(DRAM0_BASE, DRAM0_SIZE);
[...]

PR #5146, opened 2022-01-26
Xilinx Zynq MPSoC fix memory allocation

vesajaaskelainen:
Xilinx Zynq MPSoC features two base addresses for DDR memories:
- First 2 GiB of DRAM (DDR Low): 0x0000_0000 .. 0x8000_0000
- And then rest of memory (DDR High): 0x0008_0000_0000 .. 0x000F_FFFF_FFFF

This change configures DDR base addresses depending on the size of DDR memory configured during build time and also helps on taking in use device tree to capture memory size from there.

Fixes physical address space bits to 40 bits allowing all peripherals to be used in OP-TEE.

This change also fixes DDR memory size for Xilinx ZCU102 board to 4 GB.

Tested that ZCU102 boots with and without device tree being used.

Also tested with custom board with larger DDR memory.

Fixes: https://github.com/OP-TEE/optee_os/issues/5144

### DIFF
--- a/core/arch/arm/plat-zynqmp/conf.mk
+++ b/core/arch/arm/plat-zynqmp/conf.mk
@@ -23,8 +23,20 @@ else
 $(call force,CFG_ARM32_core,y)
 endif
 
-# Configure DDR size either by using device tree or override it with:
+ifneq (,$(filter $(PLATFORM_FLAVOR),zcu102 zc1751_dc1 zc1751_dc2))
+# ZCU102 features 4 GiB of DDR
+ifeq ($(CFG_ARM64_core),y)
+CFG_DDR_SIZE ?= 0x100000000
+else
+# On 32 bit build limit to 2 GiB of RAM
 CFG_DDR_SIZE ?= 0x80000000
+endif
+endif
+
+ifneq (,$(filter $(PLATFORM_FLAVOR),ultra96))
+# Ultra96 features 2 GiB of DDR
+CFG_DDR_SIZE ?= 0x80000000
+endif
 
 # By default use DT address as specified by Xilinx
 CFG_DT_ADDR ?= 0x100000
--- a/core/arch/arm/plat-zynqmp/conf.mk
+++ b/core/arch/arm/plat-zynqmp/conf.mk
@@ -16,6 +16,9 @@ $(call force,CFG_CORE_ASLR,n)
 
 ifeq ($(CFG_ARM64_core),y)
 $(call force,CFG_WITH_LPAE,y)
+
+# ZynqMP supports up to 40 bits of physical addresses
+CFG_CORE_ARM64_PA_BITS ?= 40
 else
 $(call force,CFG_ARM32_core,y)
 endif
--- a/core/arch/arm/plat-zynqmp/conf.mk
+++ b/core/arch/arm/plat-zynqmp/conf.mk
@@ -23,6 +23,9 @@ else
 $(call force,CFG_ARM32_core,y)
 endif
 
+# By default use DT address as specified by Xilinx
+CFG_DT_ADDR ?= 0x100000
+
 CFG_TZDRAM_START ?= 0x60000000
 CFG_TZDRAM_SIZE  ?= 0x10000000
 CFG_SHMEM_START  ?= 0x70000000
--- a/core/arch/arm/plat-zynqmp/conf.mk
+++ b/core/arch/arm/plat-zynqmp/conf.mk
@@ -23,6 +23,9 @@ else
 $(call force,CFG_ARM32_core,y)
 endif
 
+# Configure DDR size either by using device tree or override it with:
+CFG_DDR_SIZE ?= 0x80000000
+
 # By default use DT address as specified by Xilinx
 CFG_DT_ADDR ?= 0x100000
 
--- a/core/arch/arm/plat-zynqmp/main.c
+++ b/core/arch/arm/plat-zynqmp/main.c
@@ -65,8 +65,16 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC,
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CSU_BASE, CSU_SIZE);
 #endif
 
-#ifdef DRAM0_BASE
-register_ddr(DRAM0_BASE, DRAM0_SIZE);
+#if CFG_DDR_SIZE > 0x80000000
+
+#ifdef CFG_ARM32_core
+#error DDR size over 2 GiB is not supported in 32 bit ARM mode
+#endif
+
+register_ddr(DRAM0_BASE, 0x80000000);
+register_ddr(DRAM1_BASE, CFG_DDR_SIZE - 0x80000000);
+#else
+register_ddr(DRAM0_BASE, CFG_DDR_SIZE);
 #endif
 
 void main_init_gic(void)
--- a/core/arch/arm/plat-zynqmp/platform_config.h
+++ b/core/arch/arm/plat-zynqmp/platform_config.h
@@ -39,6 +39,14 @@
 #error "Pager not supported for zynqmp"
 #endif
 
+/* DDR Low area base */
+#define DRAM0_BASE		0
+
+#ifdef ARM64
+/* DDR High area base is only available when compiling for 64 bits */
+#define DRAM1_BASE		0x800000000
+#endif
+
 #if defined(PLATFORM_FLAVOR_zc1751_dc1) || \
 	defined(PLATFORM_FLAVOR_zc1751_dc2) || \
 	defined(PLATFORM_FLAVOR_zcu102)
@@ -55,9 +63,6 @@
 #define CONSOLE_UART_BASE	UART0_BASE
 #define IT_CONSOLE_UART		IT_UART0
 #define CONSOLE_UART_CLK_IN_HZ	UART0_CLK_IN_HZ
-
-#define DRAM0_BASE		0
-#define DRAM0_SIZE		0x80000000
 
 #define GICD_OFFSET		0
 #define GICC_OFFSET		0x20000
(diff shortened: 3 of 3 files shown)


core/arch/arm/plat-zynqmp/platform_config.h
```diff
@@ -56,8 +59,9 @@
 #define IT_CONSOLE_UART		IT_UART0
 #define CONSOLE_UART_CLK_IN_HZ	UART0_CLK_IN_HZ
 
-#define DRAM0_BASE		0
-#define DRAM0_SIZE		0x80000000
+#if !defined(CFG_DDR_SIZE)
+#define CFG_DDR_SIZE		0x80000000
```
jenswi-linaro: How about assigning the default value in `core/arch/arm/plat-zynqmp/conf.mk` instead?

core/arch/arm/plat-zynqmp/conf.mk
```diff
@@ -23,6 +23,9 @@ endif
 # ZynqMP supports up to 40 bits of physical addresses
 CFG_CORE_ARM64_PA_BITS ?= 40
 
+# Configure DDR size either by using device tree or override it with:
+# CFG_DDR_SIZE ?= 0x80000000
```
jenswi-linaro, line 27: A new `CFG_` should be documented somewhere, for instance `mk/config.mk`
`CFG_DDR_SIZE` sounds a bit ambiguous. Is it the total amount of physical DDR? Including eventual secure and possibly other reserved portions? Or is it only the non-secure parts usable by normal world?

core/arch/arm/plat-zynqmp/platform_config.h
vesajaaskelainen: I believe like in imx it is done like that. I am OK for that.

core/arch/arm/plat-zynqmp/conf.mk
vesajaaskelainen: CFG_DDR_SIZE is used in:
- plat-imx/conf.mk:CFG_DDR_SIZE
- plat-stm/conf.mk:CFG_DDR_SIZE

And in those define it as being bytes of available DDR memory. So I have continued the treat.

Other similar things that I spotted:
- plat-hikey/conf.mk:CFG_DRAM_SIZE_GB
- plat-poplar/conf.mk:CFG_DRAM_SIZE_GB
- plat-ls/conf.mk:CFG_DRAM0_SIZE
- plat-uniphier/conf.mk:CFG_DRAM0_SIZE
- plat-mediatek/conf.mk:CFG_DRAM_SIZE
- plat-stm32mp1/conf.mk:CFG_DRAM_SIZE

Probably would be a good idea to commonize these a bit and document them.

I selected the DDR variant as that's how it is called in manuals too.

In file `./core/arch/arm/include/mm/generic_ram_layout.h` there are defines that are used to configure secure RAM locations and sizes so I would re-use those for defining other features.

As there is a need to configure DDR dimensions for OP-TEE then I suppose device tree and if that is not available then configurations like DDR memory size should be used.

core/arch/arm/plat-zynqmp/platform_config.h
vesajaaskelainen: Added commit for that. Please check it out.

core/arch/arm/plat-zynqmp/conf.mk
vesajaaskelainen: If you have direction you want to go I can add necessary notes to `mk/config.mk`. Thou probably not a good idea to set default value for it.

etienne-lms: i think your current implementation (using `CFG_DDR_SIZE` straight from your _plat-xxx/conf.mk_) is fine.
What's done with these `CFG_DRAM_SIZE`'-like config switches is really platform specific. Unless there is generic code using a unified `CFG_DDR_{BASE|SIZE}` or like, i don't think there is value factorizing the platforms configs (IMHO).

core/arch/arm/plat-zynqmp/platform_config.h
```diff
@@ -39,6 +39,9 @@
 #error "Pager not supported for zynqmp"
 #endif
 
+#define DRAM0_BASE		0
+#define DRAM1_BASE		0x800000000
```
jenswi-linaro, line 43: Is this valid if `CFG_DDR_SIZE` is <= 0x80000000?

core/arch/arm/plat-zynqmp/conf.mk
jenswi-linaro: Fair enough.

core/arch/arm/plat-zynqmp/platform_config.h
vesajaaskelainen: DRAM1_BASE memory map should not be configured when DRAM size is not over 2GiB.

DRAM0_BASE address gives view for first 2 GiB and then rest is in DRAM1_BASE. 

Note that there are quite many 0 digits in DRAM1_BASE ;).

vesajaaskelainen:
I'll fix next the compiler warning.

core/arch/arm/plat-zynqmp/platform_config.h
jenswi-linaro: OK, please drop the DRAM1_BASE when it's not valid.

jforissier:
> Is there something hidden behind the builds as I am unable to reproduce the compiler issue on my machine?

Not really... except perhaps `CFG_WERROR=y` which is burried in the `_make()` function definition in `.azure-pipelines.yml` :)

```bash
$ git describe
3.15.0-318-g51d27b1f
$ make -s PLATFORM=zynqmp-zcu102 CFG_WERROR=y
In file included from core/include/mm/core_memprot.h:9,
                 from core/include/drivers/serial.h:11,
                 from core/include/drivers/cdns_uart.h:33,
                 from core/arch/arm/plat-zynqmp/main.c:35:
core/arch/arm/include/mm/core_mmu.h:225:61: error: unsigned conversion from ‘long long int’ to ‘long unsigned int’ changes value from ‘34359738368’ to ‘0’ [-Werror=overflow]
  225 |                 { .name = (_name), .type = (_type), .addr = (_addr), \
      |                                                             ^
core/arch/arm/include/mm/core_mmu.h:283:17: note: in expansion of macro ‘__register_memory’
  283 |                 __register_memory(#addr, MEM_AREA_DDR_OVERALL, (addr), \
      |                 ^~~~~~~~~~~~~~~~~
core/arch/arm/plat-zynqmp/main.c:70:1: note: in expansion of macro ‘register_ddr’
   70 | register_ddr(DRAM1_BASE, UINT64_C(CFG_DDR_SIZE) - UINT64_C(0x80000000));
      | ^~~~~~~~~~~~
cc1: all warnings being treated as errors
make: *** [mk/compile.mk:162: out/arm-plat-zynqmp/core/arch/arm/plat-zynqmp/main.o] Error 1
```

> Is that failing thing 32 bit build?

Yes!

```bash
$ grep CFG_ARM out/arm-plat-zynqmp/conf.mk
CFG_ARM32_core=y
CFG_ARM32_ldelf=y
CFG_ARM32_ta_arm32=y
```

core/arch/arm/plat-zynqmp/platform_config.h
vesajaaskelainen: While resolving 32 vs 64 bit compiling issue decided it is a good idea to hide 64 bit addresses when compiling for 32 bits so added guard for that.

etienne-lms (commented):

ricardosalveti:
`Acked-by: Ricardo Salveti <ricardo@foundries.io>`

core/arch/arm/plat-zynqmp/main.c
```diff
@@ -65,8 +65,11 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC,
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CSU_BASE, CSU_SIZE);
 #endif
 
-#ifdef DRAM0_BASE
-register_ddr(DRAM0_BASE, DRAM0_SIZE);
+#if CFG_DDR_SIZE > 0x80000000
+register_ddr(DRAM0_BASE, 0x80000000);
+register_ddr(DRAM1_BASE, CFG_DDR_SIZE - 0x80000000);
```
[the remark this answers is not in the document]
jenswi-linaro: The very least you should have a comment explaining this, but I'd prefer a:
``` C
#ifdef CFG_ARM32_core
#error DRAM1 inaccessible for ARM32
#else
...
#endif
By the way, how come dram1 isn't defined for arm32, some restriction in the platform?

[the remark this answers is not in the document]
Comment (vesajaaskelainen) [reply core/arch/arm/plat-zynqmp/main.c:70]: My thinking here is:
```
0x8 00000000

Where address space is over 32 bits.

And this value does not fit into uintptr_t.
```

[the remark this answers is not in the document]
jenswi-linaro: Agree, but that's what we have `paddr_t` for. However, that's not fully implemented in `register_ddr()` so I understand that it's confusing. How about just leaving a comment that it's a known problem/restriction on arm32? After the release we can see if there's a way of updating `register_ddr()` to handle this.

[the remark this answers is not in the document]
vesajaaskelainen: I can add the proposed error too. No problem on that.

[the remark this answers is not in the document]
vesajaaskelainen: Added #ifdef version with #error.

vesajaaskelainen:

Thanks!

vesajaaskelainen:
> `Acked-by: Ricardo Salveti <ricardo@foundries.io>`

Thanks!

jenswi-linaro:

vesajaaskelainen:

Thanks!

vesajaaskelainen:
Applied all tags.

PR #5147, opened 2022-01-26
core: panic at unexpected smc return

jenswi-linaro:
Certain smc's are not expected to return. Prior to this patch in order
to guard against unexpected return a "b ." instruction was added after
each such smc to at least capture the cpu. With the introduction of FF-A
TF-A may in case there's a mismatch between OP-TEE and TF-A
configuration return some error code when an unrecognized smc is
encountered. The result is typically that the boot hangs after the print:
I/TC: Primary CPU switching to normal world boot

To help diagnosing such errors a call to panic is added after
each smc which isn't expected to return. The result becomes instead:
I/TC: Primary CPU switching to normal world boot
E/TC:0   Panic at core/arch/arm/kernel/boot.c:122 <__panic_at_smc_return>
E/TC:0   TEE load address @ 0xe100000
E/TC:0   Call stack:
E/TC:0    0x0e10d23c
E/TC:0    0x0e124848
E/TC:0    0x0e10be60

### DIFF
--- a/core/arch/arm/include/arm32_macros.S
+++ b/core/arch/arm/include/arm32_macros.S
@@ -17,3 +17,10 @@
 		.endif
 	.endm
 
+	.macro panic_at_smc_return
+#if defined(CFG_TEE_CORE_DEBUG)
+		bl	__panic_at_smc_return
+#else
+		b	.
+#endif
+	.endm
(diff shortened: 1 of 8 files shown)


etienne-lms (commented):

jenswi-linaro:
Tag applied. As far as I'm concerned this can wait until after the release to be merged.

Issue #5142, opened 2022-01-25
pkcs11 aes_cbc padding questions


cmcgcc:
We are trying to use PKCS11_CKM_AES_CBC_PAD for encryption. But it failed without padding by ourselves.
I did not see any padding implementation in the TA source code.
Did I miss anything? Or is it still under development?

And is aes gcm also in the road map?

Thanks

etienne-lms:
The implementation is bugged. From 3.16.0-rc1:
https://github.com/OP-TEE/optee_os/blob/a5ea52c21e0de82830dbf3beaa95895b2697deb5/ta/pkcs11/src/processing_symm.c#L70

GPD TEE API spec does not define AES CBC pad operation. It seems the TA should be changed to state CKM_AES_CBC_PAD is not supported. Do you want to propose a fix?
@jforissier, could you tag this issue as a **bug**?

As for AES GCM, there is no one working on that up to my knowledge. There are implementation studies (e.g. [here](https://github.com/etienne-lms/optee_os/blob/sks-to-pkcs11/ta/pkcs11/src/processing_aes.c)) but nothing yet was submitted to the upstream. Contributions are welcome.

jforissier:
> @jforissier, could you tag this issue as a **bug**?

Done.

etienne-lms:
@cmcgcc, while looking at this, i made this [change](https://github.com/OP-TEE/optee_os/pull/5154). It is fine to you?

cmcgcc:
@etienne-lms @jforissier  I made a draft pull request for encryption padding. Let me know your thoughts. Thanks.
https://github.com/OP-TEE/optee_os/pull/5239

PR #5154, opened 2022-01-31
ta: pkcs11: PKCS11_CKM_AES_CBC_PAD is not supported

etienne-lms:
Remove PKCS11_CKM_AES_CBC_PAD from the list of the supported mechanism
as it is not implemented by the TA.

Fixes: https://github.com/OP-TEE/optee_os/issues/5142


cmcgcc:
Would it be better just add some comments to mark it unfinished? The patch will eventually be reverted.
In terms of the cr itself, these two places also need removal.
https://github.com/OP-TEE/optee_os/blob/e42199d0ac15b260f35fb1a3d076fd5313b292ea/ta/pkcs11/src/token_capabilities.c#L73
https://github.com/OP-TEE/optee_os/blob/e42199d0ac15b260f35fb1a3d076fd5313b292ea/ta/pkcs11/src/token_capabilities.c#L374

etienne-lms:
Thanks for the feedback.

`MECHANISM(PKCS11_CKM_AES_CBC_PAD, ...)` defines the mechanism in array `pkcs11_modes[]` ([token_capabilities.c](https://github.com/OP-TEE/optee_os/blob/e42199d0ac15b260f35fb1a3d076fd5313b292ea/ta/pkcs11/src/token_capabilities.c#L69)) that lists the PKCS#11 mechanism properties as per the specification, for reference. So we should keep AES_CBC_PAD reference here.

My change removes the AES_CBC_PAD entry from `token_mechanism[]` (`TA_MECHANISM(...)`) which lists the mechanisms/operations supported by the TA, hence to be removed.

I agree  `case PKCS11_CKM_AES_CBC_PAD:` could be removed from `pkcs11_mechanism_supported_key_sizes()`. I'll update this P-R.

vesajaaskelainen:
Hmm :grin: https://github.com/vesajaaskelainen/optee_os/commit/b06bf1901b54140aad869dcd2995e7caa4339b07

Wondering should we polish that instead?

cmcgcc:
We are implementing one internally for encryption. We can probably share it if no one is working on it.

etienne-lms:
@vesajaaskelainen and @cmcgcc: are you ok we merge the current P-R as is, to reflect current implementation?

cmcgcc:
We have a proposal for encryption padding. It is under internal reviewing process for oss publishing. Would you like to wait a day or two?

vesajaaskelainen:
> We have a proposal for encryption padding. It is under internal reviewing process for oss publishing. Would you like to wait a day or two?

I don't think we are that hurry. Waiting to see updated PR with the support.

etienne-lms:
No real hurry. Let's at least target an update before next 3.17.0 tag ([planned mid April](https://optee.readthedocs.io/en/latest/general/releases.html?highlight=3.17.0)).

etienne-lms:
@vesajaaskelainen, @cmcgcc, do you ack this change for 3.17.0 tag?

vesajaaskelainen:
Sure.

`Acked-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

etienne-lms:
tag applied

PR #5155, opened 2022-01-31
watchdog: add glue and arm-smc-wdt smc handler

clementleger:
This commits adds a thin glue layer for watchdog access and adds a SMC handler which handles the `arm-smc-wdt` compatible Linux driver [1].

[1] https://lwn.net/Articles/819480/

### DIFF
--- a/core/arch/arm/kernel/delay.c
+++ b/core/arch/arm/kernel/delay.c
@@ -27,8 +27,11 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <kernel/boot.h>
 #include <kernel/delay.h>
+#include <kernel/misc.h>
 
+#ifdef CFG_CORE_HAS_GENERIC_TIMER
 void udelay(uint32_t us)
 {
 	uint64_t target = timeout_init_us(us);
@@ -36,6 +39,22 @@ void udelay(uint32_t us)
 	while (!timeout_elapsed(target))
 		;
 }
+#else
+
+void udelay(uint32_t us)
+{
+	uint64_t cycles = 0;
+	uint32_t cycles_to_wait = 0;
+
+	cycles = (uint64_t)us * ((uint64_t)plat_get_freq() / 1000000ULL);
+
+	while (cycles) {
+		cycles_to_wait = MIN(cycles, UINT32_MAX);
+		wait_cycles(cycles_to_wait);
+		cycles -= cycles_to_wait;
+	}
+}
+#endif
 
 void mdelay(uint32_t ms)
 {
--- a/core/arch/arm/plat-sam/freq.c
+++ b/core/arch/arm/plat-sam/freq.c
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2022 Microchip
+ */
+
+#include <assert.h>
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <kernel/boot.h>
+#include <kernel/panic.h>
+#include <libfdt.h>
+
+static unsigned long freq;
+
+static TEE_Result get_freq_from_dt(void)
+{
+	int node;
+	struct clk *clk;
+	const void *fdt = get_embedded_dt();
+
+	if (!fdt)
+		panic();
+
+	node = fdt_node_offset_by_compatible(fdt, -1, "arm,cortex-a5");
+	if (!node)
+		panic();
+
+	if (clk_dt_get_by_name(fdt, node, "cpu", &clk))
+		panic();
+
+	freq = clk_get_rate(clk);
+
+	return TEE_SUCCESS;
+}
+early_init_late(get_freq_from_dt);
+
+unsigned long plat_get_freq(void)
+{
+	assert(freq);
+
+	return freq;
+}
--- a/core/drivers/atmel_wdt.c
+++ b/core/drivers/atmel_wdt.c
@@ -1,0 +1,281 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2022 Microchip
+ */
+
+#include <assert.h>
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <drivers/wdt.h>
+#include <io.h>
+#include <kernel/delay.h>
+#include <kernel/dt.h>
+#include <kernel/pm.h>
+#include <matrix.h>
+#include <sama5d2.h>
+#include <tee_api_types.h>
+
+#define WDT_CR			0x0
+#define WDT_CR_KEY		SHIFT_U32(0xA5, 24)
+#define WDT_CR_WDRSTT		BIT(0)
+
+#define WDT_MR			0x4
+#define WDT_MR_WDV		GENMASK_32(11, 0)
+#define WDT_MR_WDV_SET(val)	((val) & WDT_MR_WDV)
+#define WDT_MR_WDFIEN		BIT(12)
+#define WDT_MR_WDRSTEN		BIT(13)
+#define WDT_MR_WDDIS		BIT(15)
+#define WDT_MR_WDD_SHIFT	16
+#define WDT_MR_WDD_MASK		GENMASK_32(11, 0)
+#define WDT_MR_WDD		SHIFT_U32(WDT_MR_WDD_MASK, WDT_MR_WDD_SHIFT)
+#define WDT_MR_WDD_SET(val) \
+			SHIFT_U32(((val) & WDT_MR_WDD_MASK), WDT_MR_WDD_SHIFT)
+#define WDT_MR_WDDBGHLT		BIT(28)
+#define WDT_MR_WDIDLEHLT	BIT(29)
+
+#define WDT_SR			0x8
+#define WDT_SR_DUNF		BIT(0)
+#define WDT_SR_DERR		BIT(1)
+
+/*
+ * The watchdog is clocked by a 32768Hz clock/128 and the counter is on
+ * 12 bits.
+ */
+#define SLOW_CLOCK_FREQ		(32768)
+#define WDT_CLOCK_FREQ		(SLOW_CLOCK_FREQ / 128)
+#define WDT_MIN_TIMEOUT		1
+#define WDT_MAX_TIMEOUT		(BIT(12) / WDT_CLOCK_FREQ)
+
+#define WDT_DEFAULT_TIMEOUT	WDT_MAX_TIMEOUT
+
+/*
+ * We must wait at least 3 clocks period before accessing registers MR and CR.
+ * Ensure that we see at least 4 edges
+ */
+#define WDT_REG_ACCESS_UDELAY	(1000000ULL / SLOW_CLOCK_FREQ * 4)
+
+#define SEC_TO_WDT(sec)		(((sec) * WDT_CLOCK_FREQ) - 1)
+
+#define WDT_ENABLED(mr)		(!((mr) & WDT_MR_WDDIS))
+
+struct atmel_wdt {
+	struct wdt_chip chip;
+	vaddr_t base;
+	unsigned long rate;
+	uint32_t mr;
+	bool enabled;
+};
+
+static void atmel_wdt_write_sleep(struct atmel_wdt *wdt, uint32_t reg,
+				  uint32_t val)
+{
+	udelay(WDT_REG_ACCESS_UDELAY);
+
+	io_write32(wdt->base + reg, val);
+}
+
+static TEE_Result atmel_wdt_settimeout(struct wdt_chip *chip,
+				       unsigned long timeout)
+{
+	struct atmel_wdt *wdt = container_of(chip, struct atmel_wdt, chip);
+
+	wdt->mr &= ~WDT_MR_WDV;
+	wdt->mr |= WDT_MR_WDV_SET(SEC_TO_WDT(timeout));
+
+	/* WDV and WDD can only be updated when the watchdog is running */
+	if (WDT_ENABLED(wdt->mr))
+		atmel_wdt_write_sleep(wdt, WDT_MR, wdt->mr);
+
+	return TEE_SUCCESS;
+}
+
+static void atmel_wdt_ping(struct wdt_chip *chip)
+{
+	struct atmel_wdt *wdt = container_of(chip, struct atmel_wdt, chip);
+
+	atmel_wdt_write_sleep(wdt, WDT_CR, WDT_CR_KEY | WDT_CR_WDRSTT);
+}
+
+static void atmel_wdt_start(struct atmel_wdt *wdt)
+{
+	wdt->mr &= ~WDT_MR_WDDIS;
+	atmel_wdt_write_sleep(wdt, WDT_MR, wdt->mr);
+}
+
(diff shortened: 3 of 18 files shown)


core/drivers/wdt/watchdog.c
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2022 Microchip
+ */
+
+#include <drivers/wdt.h>
+
+struct wdt_chip *wdt_chip;
+
+TEE_Result watchdog_register(struct wdt_chip *chip)
```
jenswi-linaro, line 10: No one's calling this function.

core/drivers/wdt/watchdog_sm.c
```diff
@@ -0,0 +1,68 @@
...
+	SMCWD_SET_TIMEOUT	= 1,
+	SMCWD_ENABLE		= 2,
+	SMCWD_PET		= 3,
+	SMCWD_GET_TIMELEFT	= 4,
+};
+
+static unsigned long wdt_min_timeout;
+static unsigned long wdt_max_timeout;
+
+enum sm_handler_ret wdt_sm_handler(struct thread_smc_args *args)
```
jenswi-linaro, line 21: No one's calling this function.

core/drivers/wdt/watchdog_sm.c
```diff
@@ -0,0 +1,68 @@
...
+{
+	if (OPTEE_SMC_FUNC_NUM(args->a0) !=
```
jenswi-linaro, line 23: The entire id should match, no need to mask out anything.

core/drivers/wdt/watchdog_sm.c
jenswi-linaro, line 21: I'd like to see a definition of the ABI provided by this function. From what I've been able to find so it seems that this is specific for mt8173.

mk/config.mk
```diff
@@ -789,3 +789,11 @@ $(eval $(call cfg-enable-all-depends,CFG_MEMPOOL_REPORT_LAST_OFFSET, \
 CFG_TA_PAUTH ?= n
 
 $(eval $(call cfg-depends-all,CFG_TA_PAUTH,CFG_ARM64_core))
+
+# Enable support for generic watchdog registration
+# This watchdog will then be usable by non-secure world through SMC calls.
+CFG_WDT ?= n
+CFG_WDT_SM_HANDLER ?= n
+CFG_WDT_SM_HANDLER_ID ?= 0x82003D06
```
jenswi-linaro, line 797: The id seems to be specific to mt8173. It's likely going to be platform specific.

jenswi-linaro (commented):
This code isn't even built in this PR.

clementleger:
> This code isn't even built in this PR.

We already did so for the clock framework so I thought it was ok to do it like this. I will add the watchdog driver that uses it and the call `to wdt_sm_handler()` then.

jenswi-linaro:
> I will add the watchdog driver that uses it and the call to wdt_sm_handler() then.

Thanks.

core/drivers/wdt/watchdog_sm.c
clementleger: I will add a commit adding the call to `wdt_sm_handler()`

core/drivers/wdt/watchdog_sm.c
clementleger: Ack

core/drivers/wdt/watchdog_sm.c
clementleger: Actually, the definition does not seems to be specified by ARM. But the driver that was submitted states that is a generic SMC watchdog driver [1]. If not, were should I place this file ?

[1] https://lore.kernel.org/linux-arm-kernel/20200220155159.GB29658@roeck-us.net/T/

mk/config.mk
jenswi-linaro: It is quite ugly to be honest. For mt8173 it's the next free SIP function id. Who knows what's suitable for another platform?

core/drivers/wdt/watchdog_sm.c
jenswi-linaro: I assume that this might be reused for other platforms. Let's keep it here, but we should be clear on what we're defining.

mk/config.mk
clementleger: Agreed, I have switched the config option to the following which forces the user to define it:

```
ifeq ($(CFG_WDT_SM_HANDLER_ID),)
$(error CFG_WDT_SM_HANDLER_ID must be defined when enabling CFG_WDT_SM_HANDLER)
endif
```

clementleger:
New version:
- Added call to `wdt_sm_handler()` in core/sm.c
- Added atmel watchdog driver
- Added udelay support without generic timers
- Enabled watchdog on sama5d2

clementleger:
New version:
- Fixed merge conflict

clementleger:
New version: 
- Fix build when `CFG_WDT_SM_HANDLER_ID` is not defined

core/drivers/wdt/watchdog_sm.c
```diff
@@ -0,0 +1,67 @@
...
+	SMCWD_PET		= 3,
+	SMCWD_GET_TIMELEFT	= 4,
+};
+
+static unsigned long wdt_min_timeout;
+static unsigned long wdt_max_timeout;
+
+enum sm_handler_ret wdt_sm_handler(struct thread_smc_args *args)
+{
+	if (args->a0 != CFG_WDT_SM_HANDLER_ID)
```
jenswi-linaro, line 23: To better optimize the code in `sm_from_nsec()` we could do this test in an inline function which would then be executed as part of `sm_from_nsec()` avoiding the jump to this function only to return most often. Do you get the idea?

clementleger: Yes totally. This would avoid a function call each time we do the test, acked.

clementleger:
New version:
- Apply Jens recommendation to do the SMC ID check in an inlined function

core/arch/arm/kernel/delay.c
```diff
@@ -27,15 +27,30 @@
...
 		;
 }
+#else
+
+void udelay(uint32_t us)
+{
+	uint64_t cycles = 0;
+
+	cycles = (uint64_t)us * ((uint64_t)plat_get_freq() / 1000000ULL);
+	assert(cycles < (unsigned long)-1);
```
jenswi-linaro, line 49: What is the purpose of this assert? When do we care if there's an overflow or not?

core/arch/arm/plat-sam/freq.c
```diff
@@ -0,0 +1,39 @@
...
+
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <kernel/boot.h>
+#include <kernel/panic.h>
+#include <libfdt.h>
+
+static unsigned long freq;
+
+static unsigned long get_freq_from_dt(void)
```
jenswi-linaro, line 14: I'd prefer doing this once at boot so we can forget about the embedded DT once boot is done.

core/drivers/atmel_wdt.c
```diff
@@ -0,0 +1,222 @@
...
+	struct wdt_chip chip;
+	vaddr_t base;
+	unsigned long rate;
+	uint32_t mr;
+};
+
+static void atmel_wdt_write_sleep(struct atmel_wdt *wdt, uint32_t reg,
+				  uint32_t val)
+{
+	udelay(WDT_REG_ACCESS_UDELAY);
```
jenswi-linaro, line 70: You should perhaps call `wait_cycles()` instead with the number of cycles you'd like to wait.

core/drivers/wdt/watchdog_sm.c
```diff
@@ -0,0 +1,64 @@
...
+	SMCWD_PET		= 3,
+	SMCWD_GET_TIMELEFT	= 4,
+};
+
+static unsigned long wdt_min_timeout;
+static unsigned long wdt_max_timeout;
+
+enum sm_handler_ret __wdt_sm_handler(struct thread_smc_args *args)
+{
+	switch (args->a1) {
```
jenswi-linaro, line 23: What happens if this function is called concurrently on two different CPUs?

core/arch/arm/plat-sam/freq.c
clementleger: Since I'm using the clock framework, it is not ready at boot time but just after the clock framework is probed. Doing so allows to defer the clock retrieval.

core/arch/arm/kernel/delay.c
clementleger: Since the `udelay()` is implemented in assembly, it uses 32bits registers and I did not wanted to handle 32/64bits in assembly code, that's why I added this assert here.

core/drivers/atmel_wdt.c
clementleger: Hum, wouldn't this be breaking the fact that the driver can work on multiple platforms ? That why I added the udelay layer that uses the platform frequency.

core/drivers/wdt/watchdog_sm.c
clementleger: Main concern about that is `wdt_min_timeout` and `wdt_max_timeout` values that are potentially accessed concurrently even if the same values will be returned. I'm not sure where the concurrency should be handled. From what I already saw in OP-TEE, it seems you prefer to handle this in drivers but I'm not entirely sure.

core/include/drivers/wdt.h
```diff
@@ -27,10 +29,76 @@ struct wdt_chip {
...
 	void (*stop)(struct wdt_chip *chip);
 	void (*ping)(struct wdt_chip *chip);
 	TEE_Result (*set_timeout)(struct wdt_chip *chip, unsigned long timeout);
 };
 
+#ifdef CFG_WDT
+extern struct wdt_chip *wdt_chip;
+
+/* Register a watchdog as the system watchdog */
+TEE_Result watchdog_register(struct wdt_chip *chip);
```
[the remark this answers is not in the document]
clementleger: I will actually squash the two first commits il does not makes a lot of sens to register a driver but not being able to really use it with a proper API.

core/drivers/atmel_wdt.c
jenswi-linaro: Sorry, my mistake.

core/arch/arm/kernel/delay.c
jenswi-linaro: Using an assert means that you don't care about this in production builds, only during development.
Perhaps it would be better to just set a max value instead in case of overflow.

core/arch/arm/plat-sam/freq.c
jenswi-linaro: The point is that I'd like to avoid depending on an embedded DT once boot is done. Let's say that the embedded DT is replaced by a DT which is only available until we do the first exit to normal world.

core/arch/arm/kernel/delay.c
clementleger: Yes, sorry, indeed, using a maximum value would actually be better.

core/arch/arm/plat-sam/freq.c
clementleger: Hum ok, I understand. So indeed, I will try to add an early initicall to retrieve the frequency from the DT with the clock framework once and thus setup the plat_freq variable at init only.

core/arch/arm/kernel/delay.c
clementleger: Actually, what can be done is simply to loop on the `wait_cycles` function by calling it with `MIN(cycles, UINT32T_MAX)` and this will handle the 64 bits values.

jenswi-linaro: Does it make sense to spin for such a long time? If we're close to UINT32_MAX clearly something is wrong.

clementleger: Actually, the `uint32_t` is meant to hold the number of cycles which might be way greater than the number of usec.

clementleger:
New version:
- Allow to handle large cycles count
- Read freq at boot time and not when calling `plat_get_freq()`
- Squashed missing suspend support for sama5d2 watchdog in watchdog driver

core/arch/arm/kernel/delay.c
jenswi-linaro: OK, got it. Looping as you suggest makes sense.

core/drivers/wdt/watchdog_sm.c
jenswi-linaro: `wdt_min_timeout` and `wdt_max_timeout` must be protected here since they are read in this function. Other aspects should probably be handled inside the driver in case that driver is accessed somewhere else too.

clementleger:
New version:
- Add locking for min/max timeout variables

jenswi-linaro:

clementleger:
BTW, do you think the `atmel_wdt` driver should be moved in drivers/wdt ? There are already some watchdog drivers in drivers/ but maybe it's time to start organizing watchdog drivers in drivers/wdt/

clementleger:
New version:
- Added Jens Acked-by
- Fixed conflicts

clementleger:
Finally just replaced `assert(wdt_chip)` in watchdog interface with `if (!wdt_chip) return;`. Without that, if compiled without debug, then the normal world could potentially crash OP-TEE by calling `wdt_start()` or others when no watchdog device is present. This does not seems a good things to allow.

core/include/drivers/wdt.h
```diff
@@ -27,10 +31,99 @@ struct wdt_chip {
...
+
+static inline
+TEE_Result watchdog_init(unsigned long *min_timeout, unsigned long *max_timeout)
+{
+	if (!wdt_chip)
+		return TEE_ERROR_NOT_SUPPORTED;
+
+	if (wdt_chip->ops->init)
```
jenswi-linaro, line 54: How about combining the check above with this to become
``` C
if (wdt_chip && wdt_chip->ops->init)
```
It saves a few lines looks better in my opinion. Same in for instance `watchdog_stop()` below.

core/include/drivers/wdt.h
```diff
@@ -27,10 +31,99 @@ struct wdt_chip {
...
+
+	return TEE_ERROR_NOT_SUPPORTED;
+}
+
+static inline void watchdog_start(void)
+{
+	if (!wdt_chip)
+		return;
+
+	wdt_chip->ops->start(wdt_chip);
```
jenswi-linaro, line 65: How about
``` C
if (wdt_chip)
         wdt_chip->ops->start(wdt_chip);
```
instead? Same in `watchdog_ping()` and `watchdog_settimeout()` below.

core/include/drivers/wdt.h
clementleger: Acked, actually lack a `init` function should not be fatal and we should simply return `TEE_SUCCESS`

core/include/drivers/wdt.h
clementleger: Agreed

clementleger:
New version:
- Simplify ops check for readability

jenswi-linaro:
Looks good.

PR #5156, opened 2022-01-31
drivers: add atmel_saic driver

clementleger:
Add a driver for sama5d2 interrupt handler and initialize it in sama5d2 platform.

### DIFF
--- a/core/arch/arm/plat-sam/main.c
+++ b/core/arch/arm/plat-sam/main.c
@@ -28,9 +28,11 @@
 
 #include <arm32.h>
 #include <console.h>
+#include <drivers/atmel_saic.h>
 #include <drivers/atmel_uart.h>
 #include <io.h>
 #include <kernel/boot.h>
+#include <kernel/interrupt.h>
 #include <kernel/misc.h>
 #include <kernel/panic.h>
 #include <kernel/tz_ssvce_def.h>
@@ -350,4 +352,15 @@ static int matrix_init(void)
 void plat_primary_init_early(void)
 {
 	matrix_init();
+}
+
+void itr_core_handler(void)
+{
+	atmel_saic_it_handle();
+}
+
+void main_init_gic(void)
+{
+	if (atmel_saic_setup())
+		panic("Failed to init interrupts\n");
 }
--- a/core/drivers/atmel_saic.c
+++ b/core/drivers/atmel_saic.c
@@ -1,0 +1,357 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2021, Microchip
+ */
+
+#include <assert.h>
+#include <drivers/atmel_saic.h>
+#include <dt-bindings/interrupt-controller/irq.h>
+#include <io.h>
+#include <kernel/boot.h>
+#include <kernel/dt.h>
+#include <kernel/interrupt.h>
+#include <kernel/pm.h>
+#include <libfdt.h>
+#include <sam_sfr.h>
+#include <tee_api_types.h>
+#include <trace.h>
+
+#define AT91_AIC_MAX_PRIO	8
+
+#define SAMA5D2_AIC_MAX_IRQS	77
+
+#define SAMA5D2_AIC_MAX_IRQS32	((SAMA5D2_AIC_MAX_IRQS + 31) / 32)
+
+struct saic_data {
+	struct itr_chip chip;
+	vaddr_t base;
+	size_t nr_irqs;
+	uint32_t external[SAMA5D2_AIC_MAX_IRQS32];
+};
+
+static struct saic_data saic = {0};
+
+static void saic_register_pm(void);
+
+static void saic_write_reg(uint32_t reg, uint32_t val)
+{
+	io_write32(saic.base + reg, val);
+}
+
+static uint32_t saic_read_reg(uint32_t reg)
+{
+	return io_read32(saic.base + reg);
+}
+
+void atmel_saic_it_handle(void)
+{
+	uint32_t irqnr = saic_read_reg(AT91_AIC_IVR);
+
+	itr_handle(irqnr);
+	saic_write_reg(AT91_AIC_EOICR, 0);
+}
+
+static void saic_select_it(size_t it)
+{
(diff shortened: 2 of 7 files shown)


core/drivers/atmel_saic.c
```diff
@@ -0,0 +1,360 @@
...
+	int it_grp = it / 32;
+	int it_off = it % 32;
+
+	if (it >= saic.nr_irqs)
+		panic();
+
+	return saic.external[it_grp] & BIT32(it_off);
+}
+
+static uint32_t saic_get_src_type(int dt_level, int it, uint32_t *src_type)
```
jenswi-linaro, line 78: It looks like you should probably use `int` instead as return type.

core/drivers/atmel_saic.c
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	return 0;
+}
+
+static void saic_add(struct itr_chip *chip __unused, size_t it,
+		     uint32_t type, uint32_t prio)
+{
+	uint32_t src_type = AT91_AIC_SMR_HIGH_LEVEL;
+	uint32_t res = TEE_ERROR_GENERIC;
```
jenswi-linaro, line 110: This has nothing to do with the TEE_ERRROR_* codes.

core/drivers/atmel_saic.c
```diff
@@ -0,0 +1,360 @@
...
+	io_write32(sfr_base + AT91_SFR_AICREDIR, aicredir_val);
+}
+
+static void saic_init_external(const void *fdt, int node)
+{
+	int i = 0;
+	int len = 0;
+	int it_grp = 0;
+	int it_off = 0;
+	size_t it = 0xFFFFFFFF;
```
jenswi-linaro, line 223: This looks like a magic number.

core/drivers/atmel_saic.c
```diff
@@ -0,0 +1,360 @@
...
+
+	external = fdt_getprop(fdt, node, "atmel,external-irqs", &len);
+	if (!external)
+		return;
+
+	len /= sizeof(uint32_t);
+	for (i = 0; i < len; i++) {
+		it = fdt32_to_cpu(external[i]);
+
+		DMSG("IRQ %d is external", it);
```
jenswi-linaro, line 234: %zu

core/drivers/atmel_saic.c
clementleger: Agreed, to be more coherent with other value, I will switch change this function to return  a real `TEE_Result` and check that in other callers

core/drivers/atmel_saic.c
clementleger: Acked and there is no need for such value.

clementleger:
New version:
- Fixed return type and check for `saic_get_src_type()`
- Remove useless `0xFFFFFFFF` init value
- Simplify check for `saic_init()` in `main_init_gic()` (remove TEE_Result variable)

core/arch/arm/plat-sam/main.c
```diff
@@ -351,3 +353,14 @@ void plat_primary_init_early(void)
 {
 	matrix_init();
 }
+
+void itr_core_handler(void)
+{
+	atmel_saic_it_handle();
+}
+
+void main_init_gic(void)
```
jenswi-linaro, line 362: Hmm, now it's clear that we could have picked a better name for this function. :-)
We should change it in a later PR.

core/include/drivers/atmel_saic.h
```diff
@@ -0,0 +1,47 @@
...
+#include <tee_api_types.h>
+
+#define AT91_AIC_WPKEY	0x414943
+
+#define AT91_AIC_SSR		0x0
+#define	AT91_AIC_SSR_ITSEL_MASK	GENMASK_32(6, 0)
+
+#define AT91_AIC_SMR		0x4
+#define	AT91_AIC_SMR_SRC_SHIFT	5
+#define	AT91_AIC_SMR_LEVEL	(0 << AT91_AIC_SMR_SRC_SHIFT)
```
jenswi-linaro, line 18: It is generally more safe to use the macro `SHIFT_U32()` for this

core/include/drivers/atmel_saic.h
```diff
@@ -0,0 +1,47 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2021, Microchip
+ */
+
+#ifndef __DRIVERS_ATMEL_SAIC_H
+#define __DRIVERS_ATMEL_SAIC_H
+
+#include <tee_api_types.h>
+
```
jenswi-linaro, line 10: Please include <util.h> too.

core/arch/arm/plat-sam/main.c
clementleger: Yep :)

core/include/drivers/atmel_saic.h
clementleger: Ack

core/include/drivers/atmel_saic.h
clementleger: Ok

core/drivers/atmel_saic.c
```diff
@@ -0,0 +1,356 @@
...
+}
+
+static void saic_select_it(size_t it)
+{
+	assert(!(it & ~AT91_AIC_SSR_ITSEL_MASK));
+
+	saic_write_reg(AT91_AIC_SSR, it);
+}
+
+static void saic_configure_it(size_t it, int src_type, int priority)
```
jenswi-linaro, line 61: I would have expected `uint32_t` instead of `int` here.
Any particular reason why `int` is used here?

clementleger: No, you are right, using `uint32_t` would be better.

clementleger: I wil lalso change the `it` parameter of `saic_get_src_type()` to use a `size_t`

jenswi-linaro:

clementleger:
New version:
- Added @jenswi-linaro `Acked-by`

PR #5159, opened 2022-02-02
cfg: Add restrictions with CFG_TA_PAUTH

ruchi393:
Linux boot fails with XEN on CPU's with FEAT_PAUTH. So,
CFG_TA_PAUTH currently has not been tested and is incompatible
with CFG_VIRTUALIZATION.

It is also incompatible with debug options like FTRACE and GPROF.

If these incompatible options are selected together by a platform,
compilation will stop and user will be warned with an error
message.

Signed-off-by: Ruchika Gupta <ruchika.gupta@linaro.org>


jenswi-linaro:

jforissier:

ruchi393:
PR updated with tags. Thanks for the review.

PR #5164, opened 2022-02-08
Split core mmu

jenswi-linaro:
This PR is mostly moves some code around. All behaviour is supposed to remain unchanged.
Easy to fix checkpatch complaints has also been addressed.


core/arch/arm/include/barrier.h
```diff
@@ -0,0 +1,16 @@
...
+
+#ifndef __BARRIER_H
+#define __BARRIER_H
+
+#include <compiler.h>
+#include <arm.h>
+
+#define barrier_mem()		dsb()
+#define barrier_write()		dsb_ishst()
+#define barrier_read()		dsb_ishld()
```
jforissier, line 14: This commit raised a lot of interrogations and I have spent some time reading the Arm ARM as well as the Linux kernel sources... here are my questions:

1. Wouldn't it be easier to use the same names as the Linux kernel? See [`mb(), rmb(), wmb()`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/memory-barriers.txt?h=v5.17-rc3#n1816).
2. Is "inner shareable" the correct domain here? I would rather expect "full system" shareability for such generic barriers, that is `dsb_st()` and `dsb_ld()`. Related to this: before this series, `core_mmu_map_pages()` uses `dsb_ishst()` which you replace with `barrier_write()`. Is "inner shareable" correct in this context? Shouldn't it be at least "outer shareable" so that cores from a different cluster can see the update?
3. Wouldn't a `dmb` be good enough, rather than a `dsb`?
4. Finally, I notice in the Linux kernel doc that the memory barrier macros imply a compiler barrier, and their implementations indeed have a `"memory"` constraint (see [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/include/asm/barrier.h?h=v5.17-rc3#n22) for arm64 for example). Shouldn't we update our definitions of `dmb()` and `dsb()` accordingly?

jenswi-linaro: 1. Yes, I considered that. The problem is that they are not exactly the same thing, more on that below.
4. Good point, I'll fix.

How about we drop these barriers and provide a special `core_mmu_table_write_barrier()` in `core_mmu_private.h` only to be used for this special purpose?

We wait with generic barriers outside architecture specific code until we have a better use case. Or the solution is perhaps always to provide special purpose barriers? We'll see.

For most cases we have Arm recommendations on which barrier to use, to try to cover this with generic barriers as a glue layer might be too much for us to handle.

jforissier: I agree, let's do the minimum we need at this point, i.e., `core_mmu_table_write_barrier()` sounds good to me.

jenswi-linaro:
Dropped "core: add barrier.h", replaced by the new "core: add core_mmu_table_write_barrier()".
Added "core: arm{32,64}.h: add "memory" constraint".

core/arch/arm/include/mm/core_mmu_arch.h
```diff
@@ -7,8 +7,10 @@
 #define __CORE_MMU_ARCH_H
 
 #ifndef __ASSEMBLER__
+#include <arm.h>
 #include <assert.h>
 #include <compiler.h>
+#include <config.h>
```
jforissier, line 13: Nitpick: this include belongs in commit "core: split core/arch/arm/mm/core_mmu.c" rather than "core: add core_mmu_table_write_barrier()".

jforissier (commented):
"core: arm{32,64}.h: add "memory" constraint"
"core: split core/arch/arm/include/mm/core_mmu.h"
"core: split core_mmu_private.h"
"core: introduce TRUSTED_{S,D}RAM_*"
"core: add core_mmu_table_write_barrier()"

"core: split core/arch/arm/mm/core_mmu.c"

core/arch/arm/include/mm/core_mmu_arch.h
jenswi-linaro: I'll fix.

jenswi-linaro:
Comment addressed, tags applied.

core/mm/core_mmu.c
```diff
@@ -0,0 +1,2298 @@
...
+			map->region_size = SMALL_PAGE_SIZE;
+		else
+			panic("Impossible memory alignment");
+
+		if (map_is_tee_ram(map))
+			map->region_size = SMALL_PAGE_SIZE;
+	}
+}
+
+static unsigned int get_va_width(void)
```
maroueneboubakri, line 888: Maybe move this to arch instead ? not sure about `IS_ENABLED(ARM64)`  since we get `error: ‘CFG_LPAE_ADDR_SPACE_BITS’ undeclared`

jenswi-linaro: OK

jenswi-linaro:
Update

jforissier:
Fine with me.

jenswi-linaro:
Squashed in the update

jenswi-linaro:
Tag applied

maroueneboubakri:
@jenswi-linaro, sorry for coming back to this later, I have 2 points to raise

1) I believe that `core_mmu_va2idx` and `core_mmu_idx2va` in `core_mmu.h` should move to `core_mmu_arch.h` on RISCV-V the index is computed over bits per page level.

2) There is another concern which I have raised before, which is about iterating through page table levels, in RISC-V for example, it is always a count down, meaning that for sv39 format for example we have 3 levels we start iterating from level 2 to 0. In https://github.com/OP-TEE/optee_os/blob/master/core/mm/core_mmu.c#L1567 we start form the `CORE_MMU_BASE_TABLE_LEVEL` and we increment. That could work if we map everything at 4K pages which level 0. In case we want to map some pages at level 1 = 2M pages or level 2 = 1G pages that would not work, since we break when the page level shift is equal to small page which is the case for level 0. Do you think that we can introduce some sort of constant to decide whether to increment or decrement levels ? `level+= CORE_MMU_TABLE_LEVEL_STEP` ?
CORE_MMU_TABLE_LEVEL_STEP will be 1 for ARM and -1 for RISC-V.

@jforissier 

Maro

jenswi-linaro:
> > > 3\. There is another concern which I have raised before, which is about iterating through page table levels, in RISC-V for example, it is always a count down, meaning that for sv39 format for example we have 3 levels we start iterating from level 2 to 0. In https://github.com/OP-TEE/optee_os/blob/master/core/mm/core_mmu.c#L1567 we start form the `CORE_MMU_BASE_TABLE_LEVEL` and we increment. That could work if we map everything at 4K pages which level 0. In case we want to map some pages at level 1 = 2M pages or level 2 = 1G pages that would not work, since we break when the page level shift is equal to small page which is the case for level 0. Do you think that we can introduce some sort of constant to decide whether to increment or decrement levels ? `level+= CORE_MMU_TABLE_LEVEL_STEP` ?
> > >     CORE_MMU_TABLE_LEVEL_STEP will be 1 for ARM and -1 for RISC-V.
> > 
> > 
> > I see two approaches to this translation table level problem.
> > 
> > 1. We let value used for level be the same as used in the cpu.
> > 2. We make a cpu architecture indepedendant definition of what a translation table level is and then leave it for the architecture specific code to interpret that number when needed.
> > 
> > I'm inclined to start with 1 leave it to whoever has had enough first to fix it. If you'd like to go ahead with `CORE_MMU_TABLE_LEVEL_STEP` I will not object.
> 
> Option 2 sounds interesting.

Yes, but it might be a bit painful in core/arch/arm, we have a few different configurations to consider there.

jenswi-linaro:
> > > 1. I believe that `core_mmu_va2idx` and `core_mmu_idx2va` in `core_mmu.h` should move to `core_mmu_arch.h` on RISCV-V the index is computed over bits per page level.
> > 
> > 
> > I don't understand why `(va - tbl_info->va_base) >> tbl_info->shift` doesn't work. Is the number of entries in the table not a multiple of 2 or something? How should these functions be implemented for risc-v?
> 
> This how the `core_mmu_va2idx` function should be implemented in RISC-V:
> 
> ```
> static inline unsigned core_mmu_va2idx(struct core_mmu_table_info *tbl_info,
> 			vaddr_t va)
> {
> 	return (va  >> tbl_info->shift)  & ((1 << PG_LEVEL_BITS) - 1);
> }
> ```
> 
> Where `PG_LEVEL_BITS` is 9, meaning that number of entries in a page table is 512. The virtual address of the page table is not required at all. A page table may points to next level page table.
> 

It sounds like you've misunderstood what `core_mmu_va2idx()` does. The `va_base` is not the virtual address of the translation table, it's the virtual address of the start of the vm space which is supposed to be spanned by this table. Risc-v uses a radix tree just as Arm does. The layout may be different, but the principle is the same.

jenswi-linaro:
> We still need that implementation to compute the index of an address at a given page table level. We don't care about the page table itself.

For now I assume, on the longer term I think it would be an advantage to try to use translation tables in the same way as we do in OP-TEE today. When skimming through the description of the translation tables I see nothing that suggests that they need to be treated differently compared to the LPAE translation tables used on Arm. As far as I can tell they are from a higher level more or less the same. Please let me know if I've misunderstood it.

> Having `CORE_MMU_TABLE_LEVEL_STEP` or a helper function to iterate through page table levels in reverse order is enough for now.

OK

Issue #5165, opened 2022-02-09
pkcs11 missing ec_point in sign init when using an imported ec keypair


StevenCai-Gallagher:
Hi Everyone,

I am trying to import an OpenSSL generated EC keypair into the trust store, and then use the keypair to do sign/verify. 

After creating a public key object and a private key object, I get the "Missing EC_POINT attribute" error message when I am doing the c_signinit. It seems TEE_PopulateTransientObject requires public x/y values, so it will try to load the ec_point attribute from the private key object in load_tee_ec_key_attrs() when it tries to do signature initialization. But the ec_piont attribute can not be put from the template into the attributes, which attributes were used to create the private key object. It will be filtered, even though I add an ec_point attribute in the template for private key object creating. 

I wonder if I misunderstand something? If not, I don't know if we can add pkcs11_cka_ec_point in ec_private_key_opt_or_null[].

Thank you.

vesajaaskelainen:
If one just looks at the code one could believe that following is incorrect:
https://github.com/OP-TEE/optee_os/blob/master/ta/pkcs11/src/processing_ec.c#L373-L381

Eg. it should not load public key values from private key as per PKCS#11 specs.

But this seems to be a bit more complex issue:
https://github.com/OP-TEE/optee_os/blob/master/ta/pkcs11/src/processing_ec.c#L533-L540

Gives indication that OP-TEE's internal functions require us to provide EC point.

Now we should not import them as it could be wrong -- but I suppose we should calculate EC points X & Y when we import the private key -- but that gets hairy as we have utilized TEE Internal Core API for that and now that is in mix of things.

We have the code of course for that:
https://github.com/OP-TEE/optee_os/blob/master/core/lib/libtomcrypt/src/pk/ecc/ecc_make_key.c#L64-L67

But it is a bit far away... and crosses crypto api layer.

We do have `lib/libmbedtls` that we could perhaps link the TA with. We could perhaps utilize DER parser from there too.

Now need to ponder this a bit -- @etienne-lms / @ruchi393 any opinions how should we handle this?

vesajaaskelainen:
I was thinking about this and is it overkill to get external math lib involved? It would feel better to use external math lib as then it would be maintained.

One option I was thinking that perhaps we could add in `processing_ec.c`'s curve list the other curve parameters so we could calculate with GP's TEE Arithmetic API math functions the public key out of provided private key and store that also in object.

[...]

PR #5166, opened 2022-02-10
pkcs11: keep EC_POINT in attributes for private object creation

StevenCai-Gallagher:
EC_POINT attribute should be stored in both private and public object.
Remove the empty EC_POINT attribute when TA try to generate EC keys.

Signed-off-by: Steven Cai <steven.cai@gallagher.com>

When using imported EC keypair to sign a digest. It will be failed in c_signinit, which
unable to find the EC_POINT attribute in the private key object. The root cause of that
 is EC_POINT doesn't be stored in the private key object when it's created. This PR
 tries to add the EC_POINT attribute from input templates if it's available. And remove 
the empty EC_POINT attribute when it tries to generate an EC key.

issue: #5165

### DIFF
--- a/ta/pkcs11/src/pkcs11_attributes.c
+++ b/ta/pkcs11/src/pkcs11_attributes.c
@@ -467,6 +467,7 @@ static const uint32_t ec_private_key_mandated[] = {
 
 static const uint32_t ec_private_key_opt_or_null[] = {
 	PKCS11_CKA_EC_PARAMS,
+	PKCS11_CKA_EC_POINT,
 	PKCS11_CKA_VALUE,
 };
 
--- a/ta/pkcs11/src/processing_ec.c
+++ b/ta/pkcs11/src/processing_ec.c
@@ -568,6 +568,7 @@ enum pkcs11_rc generate_ec_keys(struct pkcs11_attribute_head *proc_params,
 		return PKCS11_CKR_TEMPLATE_INCONSISTENT;
 
 	if (remove_empty_attribute(pub_head, PKCS11_CKA_EC_POINT) ||
+	    remove_empty_attribute(priv_head, PKCS11_CKA_EC_POINT) ||
 	    remove_empty_attribute(priv_head, PKCS11_CKA_VALUE) ||
 	    remove_empty_attribute(priv_head, PKCS11_CKA_EC_PARAMS)) {
 		EMSG("Unexpected attribute(s) found");


vesajaaskelainen:
As far I can tell EC_POINT is not supposed to be stored in private key.

From PKCS#2.40:
[2.3.5 Elliptic curve key pair generation](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850451)

> The mechanism contributes the CKA_CLASS, CKA_KEY_TYPE, and CKA_EC_POINT attributes to the new public key and the CKA_CLASS, CKA_KEY_TYPE, CKA_EC_PARAMS or CKA_ECDSA_PARAMS and CKA_VALUE attributes to the new private key.

Also object specs do not specify it being so:
- [2.3.3 ECDSA public key objects](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850449)
- [2.3.4 Elliptic curve private key objects](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850450)

ldts:
um, does this request comply with the PKCS#11 standard? 

I am asking because other useful changes along the lines of this one - i.e: importing pre-provisioned keys from secure elements into the pkcs11 secure storage - have been proposed and rejected for not complying.

vesajaaskelainen:
> um, does this request comply with the PKCS#11 standard?
> 
> I am asking because other useful changes along the lines of this one - i.e: importing pre-provisioned keys from secure elements into the pkcs11 secure storage - have been proposed and rejected for not complying.

Problem with that is that you leave door open to operate with the _same_ key from other environments bypassing PKCS#11 TA's security policies who can access/operate with the keys.

In here the one imports EC Public & Private Key from clear text forms. In that operation one transfers the ownership to the PKCS#11 TA and then the key material is owned. If you happen to have copy of such material elsewhere that is not a problem for PKCS#11 TA as such.

cdo82:
For the moment the PKCS11 specification didn't allow EC_POINT for EC Private key.
But The behavior of PKCS11 TA need to be coherent.
When PKCS11 TA generate an EC keypair, the EC_POINT is set for  public AND private object.
https://github.com/OP-TEE/optee_os/blob/83a3d56a2c82888c86dd03bf6851cde815936020/ta/pkcs11/src/processing_ec.c#L551C1-L556C1
https://github.com/OP-TEE/optee_os/blob/83a3d56a2c82888c86dd03bf6851cde815936020/ta/pkcs11/src/processing_ec.c#L376C1-L403C1

If a people want import EC Private keys (for exemple from another TA to PKCS11 TA using PKCS11 API), this key will be not usable due to PKCS11 TA implementation.

vesajaaskelainen:
> For the moment the PKCS11 specification didn't allow EC_POINT for EC Private key. But The behavior of PKCS11 TA need to be coherent. When PKCS11 TA generate an EC keypair, the EC_POINT is set for public AND private object. https://github.com/OP-TEE/optee_os/blob/83a3d56a2c82888c86dd03bf6851cde815936020/ta/pkcs11/src/processing_ec.c#L551C1-L556C1 https://github.com/OP-TEE/optee_os/blob/83a3d56a2c82888c86dd03bf6851cde815936020/ta/pkcs11/src/processing_ec.c#L376C1-L403C1
> 
> If a people want import EC Private keys (for exemple from another TA to PKCS11 TA using PKCS11 API), this key will be not usable due to PKCS11 TA implementation.

Yes and this is a problem that needs to be fixed too.

The problem here is that TEE Internal API requires one to have both public and private properties when doing operations.

Kinda the right thing would be to create hidden property for public key part when private key is imported or when it is generated -- but we have now fleet of devices with the current behavior so right thing probably would be to make it generate/import the public key properties to private key object but with new hidden attribute. And then first look for that and if not found then look for "older" way.

When one generates new key pair:
- private key object:
   - PKCS11_CKA_EC_POINT_HIDDEN: public key
   - PKCS11_CKA_VALUE: private key
- public key object:
  - PKCS11_CKA_EC_POINT: public key

When one imports private key:

- private key object:
   - PKCS11_CKA_EC_POINT_HIDDEN: <calculate from private key>
   - PKCS11_CKA_VALUE: private key

Now when key is opened then:
- private key object:
  - read PKCS11_CKA_EC_POINT_HIDDEN
    - backup read: PKCS11_CKA_EC_POINT
  - read PKCS11_CKA_VALUE

vesajaaskelainen:
I have above mentioned changes in development that should replace this PR.

vesajaaskelainen:
PR https://github.com/OP-TEE/optee_os/pull/6204 fixes the problem. Please check it out.

PR #5168, opened 2022-02-11
plat: rcar: enable GIC support and add initial port for Renesas RCar Gen4 

lorc:
There are two patches, first one adds GIC support for Gen3, second one adds support for Gen4.

### DIFF
--- a/core/arch/arm/plat-rcar/main.c
+++ b/core/arch/arm/plat-rcar/main.c
@@ -37,15 +37,18 @@
 #include <rng_support.h>
 
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, SCIF_REG_SIZE);
-register_phys_mem_pgdir(MEM_AREA_IO_SEC, PRR_BASE, SMALL_PAGE_SIZE);
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
-register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICC_BASE, GIC_DIST_REG_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICC_BASE, GIC_CPU_REG_SIZE);
+#ifdef PRR_BASE
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, PRR_BASE, SMALL_PAGE_SIZE);
+#endif
 
 /* Legacy platforms */
 #if defined(PLATFORM_FLAVOR_salvator_h3) || \
 	defined(PLATFORM_FLAVOR_salvator_h3_4x2g) || \
 	defined(PLATFORM_FLAVOR_salvator_m3) || \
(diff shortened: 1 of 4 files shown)


core/arch/arm/plat-rcar/main.c
```diff
@@ -76,3 +82,25 @@ unsigned long plat_get_aslr_seed(void)
...
+
+	if (!gicc_base || !gicd_base)
+		panic();
+
+	/* Initialize GIC */
+	gic_init(&gic_data, gicc_base, gicd_base);
+	itr_init(&gic_data.chip);
+}
+
+#if CFG_TEE_CORE_NB_CORE > 1
```
jenswi-linaro, line 123: I don't suppose this is needed. The amount of memory saved must be minimal. Finding out a way of excluding `reset_secondary()` should enable saving a few more bytes.

jenswi-linaro: No worries. When it's needed, it's needed.

core/arch/arm/plat-rcar/main.c
```diff
@@ -82,9 +89,23 @@ unsigned long plat_get_aslr_seed(void)
...
+				    GIC_DIST_REG_SIZE);
+
+	if (!gicd_base)
+		panic();
+
+	/* Initialize GIC */
+	gic_init(&gic_data, 0, gicd_base);
+	itr_init(&gic_data.chip);
+#else
 	vaddr_t gicc_base;
```
jenswi-linaro, line 109: This can be simplified to have less duplicated code with something like:
``` C
vaddr_t gicc_base = 0;
vaddr_t gicd_base = 0;

gicd_base = ...;
if (!gicd_base)
          panic();
if (!IS_ENABLED(CFG_ARM_GICV3)) {
          gicc_base = ...;
          if (!gicc_base)
                    panic();
}
...

Comment (lorc) [reply core/arch/arm/plat-rcar/main.c]: Thanks for the suggestion. Reworked and moved all the GIC-related changes to the first patch.

Looks like all boards implement GIC initialization in the same way. Maybe it is worth to provide default initialization function? I can prepare patch.

Comment (jenswi-linaro) [reply core/arch/arm/plat-rcar/main.c]: Sure, let's see what you have in mind.

File: core/arch/arm/plat-rcar/main.c [hunk 3]
```diff
@@ -76,3 +83,32 @@ unsigned long plat_get_aslr_seed(void)
...
 }
+
+void main_init_gic(void)
+{
+	vaddr_t gicc_base = 0;
+	vaddr_t gicd_base = 0;
+
+	gicd_base = core_mmu_get_va(GICD_BASE, MEM_AREA_IO_SEC,
+				    GIC_DIST_REG_SIZE);
+
```
jenswi-linaro, line 102: This empty line isn't needed, same below the next call to `core_mmu_get_va()`.

lorc:
Rebased onto new master, with reworked GIC initialization

jenswi-linaro:

lorc:
Thank you. Added the tag.

PR #5172, opened 2022-02-14
Optee_smc_h

jenswi-linaro:
Don't include optee_smc.h outside of core/arch/arm.

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -9,6 +9,7 @@
 #include <io.h>
 #include <kernel/misc.h>
 #include <kernel/msg_param.h>
+#include <kernel/notif.h>
 #include <kernel/thread.h>
 #include <kernel/thread_private.h>
 #include <kernel/virtualization.h>
@@ -25,6 +26,9 @@
 
 static bool thread_prealloc_rpc_cache;
 static unsigned int thread_rpc_pnum;
+
+static_assert(NOTIF_VALUE_DO_BOTTOM_HALF ==
+	      OPTEE_SMC_ASYNC_NOTIF_VALUE_DO_BOTTOM_HALF);
 
 void thread_handle_fast_smc(struct thread_smc_args *args)
 {
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -190,7 +190,10 @@ static uint32_t std_smc_entry(uint32_t a0, uint32_t a1, uint32_t a2,
 
 	arg = mobj_get_va(mobj, 0, OPTEE_MSG_GET_ARG_SIZE(num_params));
 	assert(arg && mobj_is_nonsec(mobj));
-	rv = tee_entry_std(arg, num_params);
+	if (tee_entry_std(arg, num_params))
+		rv = OPTEE_SMC_RETURN_EBADCMD;
+	else
+		rv = OPTEE_SMC_RETURN_OK;
 	mobj_put(mobj);
 
 	return rv;
--- a/lib/libutils/isoc/include/assert.h
+++ b/lib/libutils/isoc/include/assert.h
@@ -25,9 +25,30 @@ void _assert_log(const char *expr, const char *file, const int line,
 	} while (0)
 #endif
 
+/* This macro is deprecated, please use static_assert instead */
 #define COMPILE_TIME_ASSERT(x) \
 	do { \
 		switch (0) { case 0: case ((x) ? 1: 0): default : break; } \
 	} while (0)
 
+#endif
+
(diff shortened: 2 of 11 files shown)


maroueneboubakri:
`Acked-by: Marouene Boubakri <marouene.boubakri@nxp.com>`

lib/libutils/isoc/include/assert.h
```diff
@@ -31,3 +31,5 @@ void _assert_log(const char *expr, const char *file, const int line,
 	} while (0)
 
 #endif
+
+#define static_assert _Static_assert
```
jforissier, line 35: Should `COMPILE_TIME_ASSERT` be deprecated? I think it should. `static_assert` is better because it can be used in some places where `COMPILE_TIME_ASSERT` cannot, and it is standard. One drawback is the C11 version requires a string argument, but the C2x version doesn't. GCC doesn't seem to care but Clang in `-std=gnu99` mode emits a warning:

```
$ clang -std=gnu99 -o test test.c
test.c:7:20: warning: '_Static_assert' with no message is a C2x extension [-Wc2x-extensions]
static_assert(0 < 1);
                   ^
                   , ""
1 warning generated.

```

So, how about adding this:
```diff
diff --git a/mk/compile.mk b/mk/compile.mk
index b5ab62b0..51ba074c 100644
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -35,7 +35,8 @@ comp-cflags-warns-high = \
        -Wmissing-prototypes -Wnested-externs -Wpointer-arith \
        -Wshadow -Wstrict-prototypes -Wswitch-default \
        -Wwrite-strings \
-       -Wno-missing-field-initializers -Wno-format-zero-length
+       -Wno-missing-field-initializers -Wno-format-zero-length \
+       -Wno-c2x-extensions \
 ifeq ($(CFG_WARN_DECL_AFTER_STATEMENT),y)
 comp-cflags-warns-high += $(call cc-option,-Wdeclaration-after-statement)
 endif
```

...as well as a comment near `COMPILE_TIME_ASSERT` to mention that `static_assert` should be used instead?

core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -26,6 +27,10 @@
 static bool thread_prealloc_rpc_cache;
 static unsigned int thread_rpc_pnum;
 
+static_assert(NOTIF_VALUE_DO_BOTTOM_HALF ==
+	      OPTEE_SMC_ASYNC_NOTIF_VALUE_DO_BOTTOM_HALF,
+	      "NOTIF_VALUE_DO_BOTTOM_HALF mismatch");
```
jforissier, line 32: The message could be omitted if you agree with my previous proposal.

jforissier (commented):
"core: vm.c: don't include sm/optee_smc.h"
"core: remove references to OPTEE_SMC_SHM_CACHED"
  
 "core: change return type for tee_entry_std() TEE_Result"

More comments below.

lib/libutils/isoc/include/assert.h
jenswi-linaro: Hmm, it seems that GCC 8.2 acts funny when omitting the message to `static_assert`.
```
core/arch/arm/kernel/thread_optee_smc.c:31:50: error: expected ‘,’ before ‘)’ token
        OPTEE_SMC_ASYNC_NOTIF_VALUE_DO_BOTTOM_HALF);
                                                  ^
cc1: error: unrecognized command line option ‘-Wno-c2x-extensions’ [-Werror]
```

I suppose it can be worked around with some macro trickery, but is it worth it? I'll try to see how hard it is.

jforissier: Oh, perhaps we need `$(call cc-option,-Wno-c2x-extensions)` instead of plain `-Wno-c2x-extensions`.

jenswi-linaro:
Update for single argument static_assert().

lib/libutils/isoc/include/assert.h
jenswi-linaro: > Oh, perhaps we need `$(call cc-option,-Wno-c2x-extensions)` instead of plain `-Wno-c2x-extensions`.

I'm not sure that helps, the message seems to be shown only when the static_assert is fails.

jenswi-linaro:
Should we go with this workaround for older versions of GCC?

jforissier:
> Should we go with this workaround for older versions of GCC?

Yes, I like it. The pre-processor trick is either quite nice or somewhat ugly depending on how you look at it :wink: but in any case it is very localized (so it will not hurt maintainability) and it is convenient. So...

For "assert.h: add static_assert" and "core: move NOTIF_VALUE_DO_BOTTOM_HALF assert to thread_optee_smc.c":

jenswi-linaro:
Tags applied.

PR #5175, opened 2022-02-15
drivers: crypto: add parameter checks for RSA signature

clementfaure:
Add size check in the crypto driver for RSA sign and verify functions.
For both functions, the encoded message length has some size
constraints [1].

[1]: Public-Key Cryptography Standards (PKCS) #1: RSA Cryptography
https://datatracker.ietf.org/doc/html/rfc3447#section-9.1.1

Fixes: f5a70e3ef ("drivers: crypto: generic resources for crypto device driver - RSA")
Signed-off-by: Cedric Neveux <cedric.neveux@nxp.com>


etienne-lms:
Looks good to me.
CI failure seems unrelated to this change.

clementfaure:
Thanks @etienne-lms !

PR #5177, opened 2022-02-15
[Hikey] Define SDP in embedded dtb

omasse-linaro:
Allow definition of the Secure Data Path memory region in an embedded dtb
It remove SDP memory intersection checking as embedded DTB is not
available during init of tee core memory mapping

Comply with reserved memory bindings
Linux documentation file:
Documentation/devicetree/bindings/reserved-memory/reserved-memory.yaml

Signed-off-by: Olivier Masse <olivier.masse@nxp.com>

### DIFF
--- a/core/arch/arm/dts/hikey_sdp.dts
+++ b/core/arch/arm/dts/hikey_sdp.dts
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2021, NXP. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/dts-v1/;
+
+/ {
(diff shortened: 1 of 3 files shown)


jenswi-linaro:
I tested this on hikey and OP-TEE seems to file the entry in the DTB:
```
I/TC: No non-secure external DT
D/TC:0 0 discover_nsec_memory:1204 Warning register_dynamic_shm() is deprecated, please use register_ddr() instead
I/TC: Embedded DTB found
D/TC:0 0 configure_sdp_dtb:413 region base 0x3e800000 4194304
D/TC:0 0 carve_out_phys_mem:342 No need to carve out 0x3e800000 size 0x400000
```

PR #5178, opened 2022-02-15
plat: arm: refactor GIC initialization

lorc:
All platforms (except STM32MP1) follow the same pattern during GIC
initialization: get virtual addresses for distributor (and optionally,
for CPU interface), check that they are not NULL, call either
`gic_init()` or `gic_init_base_addr()`.

We can move most of this logic into `gic_init_base_addr()` while platform
code will supply only base physical addresses for distributor and CPU
interface. This will simplify and align platform code.

ST32MP1 had more complex logic, as it used `io_pa_or_va_secure()` to get
MMIO range addresses. However, as `main_init_gic()` called
`assert(cpu_mmu_enabled())`, there is no sense in using
`io_pa_or_va_secure()`, because we already ensured that VA will be
always used. Thus `assert()` call was moved to `gic_init_base_addr()`, and
STM32MP1 were aligned with other platforms.

Signed-off-by: Volodymyr Babchuk <volodymyr_babchuk@epam.com>

---

This PR is the result of discussion in https://github.com/OP-TEE/optee_os/pull/5168

### DIFF
--- a/core/arch/arm/plat-aspeed/platform_ast2600.c
+++ b/core/arch/arm/plat-aspeed/platform_ast2600.c
@@ -61,17 +61,7 @@ static struct gic_data gic_data;
 
 void main_init_gic(void)
 {
-	vaddr_t gicc_base = 0;
-	vaddr_t gicd_base = 0;
-
-	gicc_base = core_mmu_get_va(GIC_BASE + GICC_OFFSET,
-				    MEM_AREA_IO_SEC, SMALL_PAGE_SIZE);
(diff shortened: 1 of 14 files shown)


jenswi-linaro (commented):

I would be nice with acks from a few other platform maintainers.

lorc:
@etienne-lms, may I ask you to take a look at this PR?

STM was affected most of all platforms.

etienne-lms:
Hi @lorc. LGTM. Thanks for the stm32mp1 update.

lorc:
Thank you for review.

I have added the tags.

Issue #4687, opened 2021-06-16
i.MX 8M: Allocate more memory for TA


JamesMenetrey:
Dear OP-TEE developers,

I know this is a recurring theme, but I would need to allocate big chunks of memory for my TAs for my i.MX 8M evaluation board.

Based on the default configuration of OP-TEE, I am able to allocate at most a buffer with a size of 7MB. Changing the definition of `PGT_CACHE_SIZE` enabled me to allocate approximately 27MB, which seems to be capped to that value.

After reading various issues on GitHub and some documentations, I understood that changing the size allocated to the TZDRMAM (namely [CFG_TZDRAM_SIZE](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-imx/conf.mk#L423) in the case of the i.MX board) would be sufficient, but modifying the value from 0x01e00000 to 0x02500000 throws an error at boot time.

Here is the full log I have when I boot with the change:

```
U-Boot SPL 2020.10-rc1 (Jun 16 2021 - 18:00:08 +0000)
PMIC:  PFUZE100 ID=0x10
Normal Boot
Trying to boot from MMC2
D/TC:0   plat_get_aslr_seed:109 Warning: no ASLR seed
D/TC:0   add_phys_mem:564 TEE_SHMEM_START type NSEC_SHM 0x00500000 size 0x00200000
D/TC:0   add_phys_mem:564 TA_RAM_START type TA_RAM 0xfe200000 size 0x02300000
D/TC:0   add_phys_mem:564 VCORE_UNPG_RW_PA type TEE_RAM_RW 0xfe062000 size 0x0019e000
D/TC:0   add_phys_mem:564 VCORE_UNPG_RX_PA type TEE_RAM_RX 0xfe000000 size 0x00062000
D/TC:0   add_phys_mem:564 ROUNDDOWN(0x38800000, CORE_MMU_PGDIR_SIZE) type IO_SEC 0x38800000 size 0x00200000
D/TC:0   add_phys_mem:564 ROUNDDOWN(0x30360000, CORE_MMU_PGDIR_SIZE) type IO_SEC 0x30200000 size 0x00400000
D/TC:0   add_phys_mem:564 ROUNDDOWN((0x30860000), CORE_MMU_PGDIR_SIZE) type IO_NSEC 0x30800000 size 0x00400000
D/TC:0   add_va_space:604 type RES_VASPACE size 0x00a00000
[...]

PR #5179, opened 2022-02-16
RTC: add basic RTC support in OP-TEE

clementleger:
This PR adds support for the following:
- RTC API
- sama5d2 RTC driver
- RTC PTA

On sama5d2, the RTC is contained in a SYS peripheral group which contains the WDT, the RSTC, the PIT and the RTC. This group can only be set secure as a whole and if doing so, the RTC won't be accessible anymore by the non-secure world. This PR adds support to expose the RTC through a PTA which will allow to communicate with a non-secure world driver (see [1]).

[1] (https://github.com/clementleger/linux/commit/78c118b0b3b9701ab5edb2b6d1964e700d3ee8ba)


lib/libutils/ext/include/util.h
```diff
@@ -86,6 +86,15 @@
 	(__extension__ ({ __typeof__(x) _x = (x); \
 	  __typeof__(y) _y = (y); \
 	  (_x + (_y / 2)) / _y; }))
+
+/*
+ * Divide a positive dividend with a positive or negative divisor and round
+ * to the nearest value.
+ */
+#define DIV_ROUND_NEAREST(x, y) \
```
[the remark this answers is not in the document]
clementleger: Ok.

clementleger:
New version:
- Rebased on master
- Remove DIV_ROUND_NEAREST commit
- Use UDIV_ROUND_NEAREST function instead in RTC driver

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_OUTPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	time = params[0].memref.buffer;
+	if (!time || params[0].memref.size < sizeof(*time))
```
jenswi-linaro, line 31: Do we need to be flexible with the size?

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	time = params[0].memref.buffer;
```
jenswi-linaro, line 52: Size check missing.

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+{
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_OUTPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	return rtc_get_offset((long *)&params[0].value.a);
```
jenswi-linaro: Doesn't work on 64bit platforms.

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_OUTPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	info = params[0].memref.buffer;
+	if (!info || params[0].memref.size < sizeof(*info))
```
jenswi-linaro, line 113: Do we need to be flexible with the size?

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	time = params[0].memref.buffer;
+	if (!time)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	rtc_set_time(time);
```
jenswi-linaro, line 56: What happens if the info in `time` doesn't make sense?

core/include/drivers/rtc.h
```diff
@@ -0,0 +1,136 @@
...
+#define DRIVERS_RTC_H
+
+#include <tee_api_types.h>
+
+#define RTC_CORRECTION_FEATURE	BIT(0)
+
+#define TIME_INIT(year, month, day, hour, min, sec) \
+					{sec, min, hour, day, month, year, 0}
+
+struct optee_rtc_time {
```
jenswi-linaro, line 16: This constitutes the ABI provided by the PTA in a later patch. How can we know that this struct is enough? When comparing with `struct rtc_time` in Linux there's a few more fields.

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2018, Linaro Limited
+ * Copyright (c) 2021, EPAM Systems. All rights reserved.
+ *
+ * Based on plat-synquacer/rtc_pta.c
```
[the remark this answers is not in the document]
clementleger: Yes

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+static TEE_Result rtc_pta_get_time(uint32_t types,
+				   TEE_Param params[TEE_NUM_PARAMS])
+{
+	struct optee_rtc_time *time = NULL;
+
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_OUTPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		DMSG("bad parameters types: 0x%" PRIx32, types);
```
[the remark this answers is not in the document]
clementleger: I don't think so, I just started from the hwrng.c and these were present but they can be left out for sure.

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
...
+
+	return rtc_get_info(&info->features, &info->range_min,
+			    &info->range_max);
+}
+
+static TEE_Result invoke_command(void *session __unused,
+				 uint32_t cmd, uint32_t ptypes,
+				 TEE_Param params[TEE_NUM_PARAMS])
+{
+	FMSG(PTA_NAME" command %#"PRIx32" ptypes %#"PRIx32, cmd, ptypes);
```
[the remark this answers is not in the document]
clementleger: Nope, I will remove all the print

core/pta/rtc.c
```diff
@@ -0,0 +1,141 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2018, Linaro Limited
+ * Copyright (c) 2021, EPAM Systems. All rights reserved.
```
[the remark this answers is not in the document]
clementleger: Yes, I'll add the Microchip one.

core/pta/rtc.c
clementleger: Actually no, I think this could be fixed for the moment and be a "==" check

core/pta/rtc.c
clementleger: Hum yes, I thought about this. How can I convey a 64 bit value here ? Use a memref like what I did for the struct ?
I saw that the tee_params in Linux are 64bits each but then in OP-TEE, they are 32 bits. Did I missed something ?
I'm also not sure it makes sense to have an offset larger that 32 bits.

core/pta/rtc.c
clementleger: Again, I think no, we can check for the size to be strictly equal.

core/pta/rtc.c
clementleger: It actually depends on the driver. The AT91 RTC will actually return an error while programming it. Maybe I could add a sanity check to at least avoid total non sense but it will probably be a bit difficult (leap year, february duration and so on)

core/include/drivers/rtc.h
clementleger: Yes, in the linux `rtc_time` struct, there is `yday` which is the year day and `is_dst` for daylight saving time. The `yday` member can be computed based on the other informations and `is_dst` is actually not used. I asked the maintainer about these fields and he says they can be left out.

core/pta/rtc.c
clementleger: Check added

clementleger:
New version:
- Use SHIFT_U32 were applicable in atmel_rtc driver
- Removed all debug print in rtc_pta.c
- Use MIN() in atmel_rtc driver
- Update copyrights

core/include/drivers/rtc.h
clementleger: Ok thanks @jenswi-linaro !

core/pta/rtc.c
```diff
@@ -0,0 +1,144 @@
...
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result rtc_pta_set_time(uint32_t types,
+				   TEE_Param params[TEE_NUM_PARAMS])
+{
+	struct optee_rtc_time *time = NULL;
+
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INPUT,
```
jenswi-linaro, line 44: We are transmitting 7 32-bit fields for struct optee_rtc_time. We can pass those fields as value parameters instead since we have 8 such fields available. With this the kernel RTC driver wouldn't need to use any shared memory for communication and here we wouldn't have to worry about the passed parameters being changed while used.

lib/libutee/include/rtc_pta_client.h
```diff
@@ -0,0 +1,94 @@
...
+ * Result:
+ * TEE_SUCCESS - Invoke command success
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ */
+#define PTA_CMD_RTC_GET_OFFSET		0x3
+
+/*
+ * PTA_CMD_RTC_SET_OFFSET - Set RTC offset
+ *
+ * param[0] (in value) - value.a: RTC offset to be set
```
jenswi-linaro, line 83: Can the offset be negative?

clementleger: Yes the offset can be negative.

core/pta/rtc.c
```diff
@@ -0,0 +1,144 @@
...
+				     TEE_Param params[TEE_NUM_PARAMS])
+{
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	rtc_set_offset(params[0].value.a);
```
jenswi-linaro, line 70: Since an offset can be negative I suppose that even though the value is a `uint32_t` it should be casted into a `int32_t` in order to be properly upgraded to a `long` on 64bit platforms.

core/pta/rtc.c
jenswi-linaro: Hmm, that's unfortunate. Then I guess we need some struct definition, but I'd rather have a separate one so internals of OP-TEE doesn't become part of a TA ABI.

clementleger: Agreed, I did the same on Linux side to avoid reusing the `struct rtc_time`. I can probably move this definition in `rtc_pta_client.h` and add another one in rtc.h. They will be identical though.

etienne-lms (commented):
Some comments. The API looks good to me, only some descriptions are missing.
As for file _rtc_pta_client.h_. most PTA header file (where their API/ABI is defined) are named _pta_foo.h_. Prefer renaming it to _pta_rtc.h_?

core/include/drivers/rtc.h
```diff
@@ -0,0 +1,136 @@
...
+/*
+ * Copyright 2022 Microchip.
+ */
+
+#ifndef DRIVERS_RTC_H
+#define DRIVERS_RTC_H
+
+#include <tee_api_types.h>
+
+#define RTC_CORRECTION_FEATURE	BIT(0)
```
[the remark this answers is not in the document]
clementleger: Sure

core/include/drivers/rtc.h
```diff
@@ -0,0 +1,136 @@
...
+
+#define TIME_INIT(year, month, day, hour, min, sec) \
+					{sec, min, hour, day, month, year, 0}
```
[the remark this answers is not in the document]
clementleger: Ok

core/pta/rtc.c
```diff
@@ -0,0 +1,182 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2018, Linaro Limited
+ * Copyright (c) 2021, EPAM Systems. All rights reserved.
+ * Copyright (c) 2022, Microchip
+ *
+ * Based on pta/hwrng.c
```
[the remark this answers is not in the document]
clementleger: Ack

core/pta/rtc.c
```diff
@@ -0,0 +1,182 @@
...
+	if (!info || params[0].memref.size != sizeof(*info))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	res = rtc_get_info(&features, &range_min, &range_max);
+	if (res)
+		return res;
+
+	info->version = PTA_RTC_INFO_VERSION;
+
+	if (features & RTC_CORRECTION_FEATURE)
```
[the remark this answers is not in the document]
clementleger: Actually, If if the core reported that RTC_CORRECTION_FEATURE is supported, then `set_offset/get_offset` are implemented. And even if not implemented, an error will be returned. However, I think i will modify the way features are set to allow the RTC API to determine that by itself.

lib/libutee/include/rtc_pta_client.h
```diff
@@ -0,0 +1,106 @@
...
+	uint32_t tm_sec;
+	uint32_t tm_min;
+	uint32_t tm_hour;
+	uint32_t tm_mday;
+	uint32_t tm_mon;
+	uint32_t tm_year;
+	uint32_t tm_wday;
+} __packed;
+
+struct pta_rtc_info {
```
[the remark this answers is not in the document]
clementleger: Thanks for the description !

core/drivers/atmel_rtc.c
```diff
@@ -0,0 +1,309 @@
...
+	atmel_rtc_write(RTC_CR, cr | RTC_CR_UPDCAL | RTC_CR_UPDTIM);
+
+	do {
+		sr = atmel_rtc_read(RTC_SR);
+	} while (!(sr & RTC_SR_ACKUPD));
+
+	atmel_rtc_write(RTC_SCCR, RTC_SCCR_ACKCLR);
+
+	atmel_rtc_write(RTC_TIMR,
+			RTC_SET_VAL(TIME_SEC, bcd_encode(tm->tm_sec))
```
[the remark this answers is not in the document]
clementleger: Acked

core/drivers/atmel_rtc.c
```diff
@@ -0,0 +1,309 @@
...
+}
+
+static TEE_Result atmel_rtc_get_offset(struct rtc *rtc __unused, long *offset)
+{
+	uint32_t mr = atmel_rtc_read(RTC_MR);
+	long val = RTC_VAL(MR_CORR, mr);
+
+	if (!val) {
+		*offset = 0;
+		return 0;
```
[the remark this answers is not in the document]
clementleger: Done

core/pta/rtc.c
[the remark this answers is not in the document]
clementleger: Fixed this by computing features in `rtc_get_info()`

lib/libutee/include/rtc_pta_client.h
```diff
@@ -0,0 +1,106 @@
...
+#define __RTC_PTA_CLIENT_H
+
+#include <tee_api_types.h>
+
+#define PTA_RTC_UUID { 0xf389f8c8, 0x845f, 0x496c, \
+		{ 0x8b, 0xbe, 0xd6, 0x4b, 0xd2, 0x4c, 0x92, 0xfd } }
+
+#define PTA_RTC_INFO_VERSION		0x1
+
+#define PTA_RTC_CORRECTION_FEATURE	BIT(0)
```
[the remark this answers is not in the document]
clementleger: Done

lib/libutee/include/rtc_pta_client.h
```diff
@@ -0,0 +1,106 @@
...
+ * param[3] unused
+ *
+ * Result:
+ * TEE_SUCCESS - Invoke command success
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect struct size for RTC time
+ */
+#define PTA_CMD_RTC_GET_TIME		0x1
+
+/*
+ * PTA_CMD_RTC_GET_TIME - Get time from RTC
```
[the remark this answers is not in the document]
clementleger: Fixed

lib/libutee/include/rtc_pta_client.h
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * param[0] (in memref) - RTC buffer memory reference containing a struct
+ *			  optee_rtc_time
```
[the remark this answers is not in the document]
clementleger: Fixed renaming

clementleger:
> Some comments. The API looks good to me, only some descriptions are missing. As for file _rtc_pta_client.h_. most PTA header file (where their API/ABI is defined) are named _pta_foo.h_. Prefer renaming it to _pta_rtc.h_?

I named it after the rng pta one which is name `rng_pta_client.h`. I can rename it but it won't be coherent with what already exists.

clementleger:
New version:
- Fixed Etienne comments
- Rework RTC feature to be set based on set/get_offset presence
- Add a check for rtc ops to be set in `rtc_register()`
- Rename `rtc_pta_client.h` to `pta_rtc.h`

Loveself7070:
تم الإرسال من هاتف HONOR الخاص بي-------- الرسالة الأصلية --------من:‏ Clément Léger ***@***.***>التاريخ:‏ الثلاثاء، ٢٢ فبراير، ٢٠٢٢ ٤:٥٥ مإلى:‏ OP-TEE/optee_os ***@***.***>نسخة إلى:‏ Subscribed ***@***.***>الموضوع:‏ Re: [OP-TEE/optee_os] RTC: add basic RTC support in OP-TEE (PR #5179)
New version:
Fixed Etienne commentsRework RTC feature to be set based on set/get_offset presenceAdd a check for rtc ops to be set in rtc_register()Rename rtc_pta_client.h to pta_rtc.h

clementleger:
New version:
- Removed return value from `rtc_register()` and add assert for checking `ops`. Using a returned value does not really makes sense. The fact that a RTC driver provides the correct operations should be done during development and thus, there is no need to check that at runtime.

clementleger:
Also, renamed `CFG_RTC` to `CFG_DRIVERS_RTC` to match what was done for other frameworks. (Unfortunately, I missed that for the watchdog)

etienne-lms (commented):
Hi @clementleger, I though I posted some comments but they were still drafts. Here there are. Only minor comments.

core/drivers/rtc/rtc.c
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright 2022 Microchip
+ */
+
+#include <drivers/rtc.h>
+#include <tee_api_types.h>
+
+struct rtc *rtc_device;
+
+TEE_Result rtc_register(struct rtc *rtc)
+{
```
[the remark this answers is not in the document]
clementleger: Yes

lib/libutee/include/pta_rtc.h
```diff
@@ -0,0 +1,117 @@
...
+#define PTA_RTC_UUID { 0xf389f8c8, 0x845f, 0x496c, \
+		{ 0x8b, 0xbe, 0xd6, 0x4b, 0xd2, 0x4c, 0x92, 0xfd } }
+
+#define PTA_RTC_INFO_VERSION		0x1
+
+/*
+ * RTC provides set/get offset and thus command PTA_CMD_RTC_GET_OFFSET and
+ * PTA_CMD_RTC_SET_OFFSET might be called
+ */
+#define PTA_RTC_CORRECTION_FEATURE	BIT(0)
```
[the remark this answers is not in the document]
clementleger: Completely ok with that, I missed it.

lib/libutee/include/pta_rtc.h
```diff
@@ -0,0 +1,117 @@
...
+struct pta_rtc_info {
+	uint64_t version;
+	uint64_t features;
+	struct pta_rtc_time range_min;
+	struct pta_rtc_time range_max;
+} __packed;
+
+/*
+ * PTA_CMD_RTC_GET_INFO - Get RTC information
+ *
```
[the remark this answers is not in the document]
clementleger: Acked, I shouldn't have started from this rng pta I guess.

core/pta/rtc.c
```diff
@@ -0,0 +1,180 @@
...
+
+	return res;
+}
+
+static TEE_Result rtc_pta_get_info(uint32_t types,
+				   TEE_Param params[TEE_NUM_PARAMS])
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct pta_rtc_info *info = NULL;
+	struct optee_rtc_time range_min = {};
```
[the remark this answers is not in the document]
clementleger: Done

core/pta/rtc.c
```diff
@@ -0,0 +1,180 @@
...
+	struct optee_rtc_time range_min = { };
+	struct optee_rtc_time range_max = { };
+	uint64_t features = 0;
+
+	if (types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_OUTPUT,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
```
etienne-lms, line 134: braces

core/pta/rtc.c
```diff
@@ -0,0 +1,180 @@
...
+
+	info = params[0].memref.buffer;
+	if (!info || params[0].memref.size != sizeof(*info))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	res = rtc_get_info(&features, &range_min, &range_max);
+	if (res)
+		return res;
+
+	info->version = PTA_RTC_INFO_VERSION;
```
etienne-lms, line 144: can you add `memset(info, 0, sizeof(*info));` before filling with data?

lib/libutee/include/pta_rtc.h
```diff
@@ -0,0 +1,103 @@
...
+
+struct pta_rtc_time {
+	uint32_t tm_sec;
+	uint32_t tm_min;
+	uint32_t tm_hour;
+	uint32_t tm_mday;
+	uint32_t tm_mon;
+	uint32_t tm_year;
+	uint32_t tm_wday;
+} __packed;
```
etienne-lms, line 29: is `__packed` needed? I remember @jenswi-linaro worrying about such useless attribute.

etienne-lms (commented):
The ability the set date and offset is opened to any client application invoking the PTA. I wonder if should there be some identification passed to restrict such accesses upon 'root' or like login identification?

lib/libutee/include/pta_rtc.h
clementleger: In theory no since both OP-TEE and the OS runs on the same platform. I added them because I was asked to on some other PR which add the same kind of communication pattern.

core/pta/rtc.c
clementleger: Yes

lib/libutee/include/pta_rtc.h
clementleger: Will wait for @jenswi-linaro answer before removing them.

core/pta/rtc.c
```diff
@@ -0,0 +1,180 @@
...
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE,
+				     TEE_PARAM_TYPE_NONE)) {
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	res = rtc_get_offset(&offset);
+	if (res)
+		return res;
+
```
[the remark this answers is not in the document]
clementleger: Yes, it could be better to handle that.

etienne-lms:
On Linux side, we can see 2 ways to set the RTC data. Either using a kernel driver plugged to /dev/rtc* standard tools or an application that would  straight invoke the PTA.
If you plan to use the first case, the kernel driver can open the session with [`TEE_IOCTL_LOGIN_REE_KERNEL`](https://elixir.bootlin.com/linux/v5.15/source/include/uapi/linux/tee.h#L187) ID while the PTA would restrict set time/offset commands to the sessions opened with `TEE_LOGIN_REE_KERNEL`. The kernel uses [`CAP_SYS_TIME`](https://elixir.bootlin.com/linux/v5.15/source/drivers/rtc/dev.c#L223) to identify applications allowed to alter time reference.
In the second case, I think the PTA needs some `CFG_RTC_PTA_LOGIN_USER_ID` or `CFG_RTC_PTA_LOGIN_GROUP_ID` or like configuration means.

clementleger:
> On Linux side, we can see 2 ways to set the RTC data. Either using a kernel driver plugged to /dev/rtc* standard tools or an application that would straight invoke the PTA. If you plan to use the first case, the kernel driver can open the session with [`TEE_IOCTL_LOGIN_REE_KERNEL`](https://elixir.bootlin.com/linux/v5.15/source/include/uapi/linux/tee.h#L187) ID while the PTA would restrict set time/offset commands to the sessions opened with `TEE_LOGIN_REE_KERNEL`. The kernel uses [`CAP_SYS_TIME`](https://elixir.bootlin.com/linux/v5.15/source/drivers/rtc/dev.c#L223) to identify applications allowed to alter time reference.

Ok great, thanks for the explanations ! I already have a kernel driver for that RTC curently using `TEE_IOCTL_LOGIN_PUBLIC` so I will consider to open the session using `TEE_IOCTL_LOGIN_REE_KERNEL`.

clementleger:
New version:
- Fix trailing newline
- Add open_session callback to allow only kernel clients.

core/include/drivers/rtc.h
```diff
@@ -0,0 +1,141 @@
...
+#ifndef DRIVERS_RTC_H
+#define DRIVERS_RTC_H
+
+#include <tee_api_types.h>
+#include <util.h>
+
+/* The RTC allows to set/get offset for correction */
+#define RTC_CORRECTION_FEATURE	BIT(0)
+
+#define TIME_INIT(year, month, day, hour, min, sec) \
```
jenswi-linaro, line 15: This macro seems a bit unnecessary. If you don't like the order in the struct you can change it there instead.

core/arch/arm/plat-sam/conf.mk
```diff
@@ -80,3 +80,7 @@ ifeq ($(CFG_WDT_SM_HANDLER),y)
 CFG_WDT_SM_HANDLER_ID := 0x2000500
 endif
```
Loveself7070, line 81: #4687 CFG_WDT_SM_HANDLER_ID : = 0x2000500

core/include/drivers/rtc.h
clementleger: Yes, you are right, it was used previously and was meant to generate a timestamp rather than the struct. Will remove.

clementleger:
New version:
- Remove `TIME_INIT()` macro and reorder optee_rtc_time struct members to be more logical

jenswi-linaro:

clementleger:
New version:
- Added Acked-by Jens

core/drivers/atmel_rtc.c
```diff
@@ -0,0 +1,309 @@
...
+	.get_time = atmel_rtc_get_time,
+	.set_time = atmel_rtc_set_time,
+	.get_offset = atmel_rtc_get_offset,
+	.set_offset = atmel_rtc_set_offset,
+};
+
+static struct rtc atmel_rtc = {
+	.ops = &atmel_rtc_ops,
+	.range_min = {1900, 1, 1, 0, 0, 0, 0},
+	.range_max = {2099, 12, 31, 23, 59, 59, 0},
```
etienne-lms, line 269: minor: add space char in braces:
```c
	.range_min = { 1900, 1, 1, 0, 0, 0, 0 },
	.range_max = { 2099, 12, 31, 23, 59, 59, 0 },

etienne-lms (commented):
"drivers: rtc: add RTC API"
"pta: add PTA for RTC"

"driver: atmel_rtc: add driver for atmel RTC" (with minor comment)
"dts: sama5d2: set RTC as secure"
"plat-sam: enable RTC support"

clementleger:
New version:
- Added Acked/Reviewed-By Etienne
- Add spaces before/after braces

jforissier:
@clementleger thanks!

PR #5180, opened 2022-02-16
plat-imx: Add SA settings for i.MX7DS

marckleinebudde:
The Secure Access register configures the access mode for
non-TrustZone aware DMA masters. To ensure that no DMA master can read
the secure memory for OP-TEE, we set access for all masters except the
ARM CP15 register to non-secure only and lock the settings afterwards.

Signed-off-by: Marc Kleine-Budde <mkl@pengutronix.de>


jforissier:
CC @OP-TEE/plat-imx

clementfaure:
Hello,
I would like to remind that OP-TEE support for i.MX platforms does not aim at being secure out of the box. The already existing configuration of the SoC security modules should not mislead the user into thinking that i.MX default security configuration is fully secure in adequation with his requirements.
You can find more information here : https://optee.readthedocs.io/en/latest/architecture/platforms/nxp.html
With that being said,

PR #5183, opened 2022-02-18
drivers: crypto: se050: panic on initialization error

ldts:
Failure to initialize the SE05x device is a critical operation as it will
effectively disable ciphers configured at build time.

This also matches the behaviour implemented by the other crypto drivers.


jforissier (approved):
Makes sense.

ldts:
@jforissier thanks. tags updated

PR #5187, opened 2022-02-22
dirvers: stm32_gpio: Fix register access before enabling clocks

loicbauer-socomec:
The valid_gpio_config function did access the GPIO register
before the clock was enabled, which did lead to the assert always
failing when using stm32_gpio_set_output_level.

Signed-off-by: Loïc Bauer <loic.bauer@socomec.com>

### DIFF
--- a/core/drivers/stm32_gpio.c
+++ b/core/drivers/stm32_gpio.c
@@ -394,9 +394,9 @@ int stm32_gpio_get_input_level(unsigned int bank, unsigned int pin)
 	struct clk *clk = stm32_get_gpio_bank_clk(bank);
 	int rc = 0;
 
-	assert(valid_gpio_config(bank, pin, true));
-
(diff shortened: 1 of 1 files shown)


jforissier:
Hi @loicbauer-socomec, thanks for the patch. Two minor things about the commit description:
- We prefer to use the present tense to describe the situation before the patch is applied
- Add parentheses after function names

So I recommend: "The valid_gpio_config() function accesses the GPIO register before the clock is enabled, which leads to the assert always failing when using stm32_gpio_set_output_level()."

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with the changes suggested by @jforissier in the commit log.

loicbauer-socomec:
Thank you, I updated the commit message.

PR #5203, opened 2022-03-04
Extending watchdog timeout

etienne-lms:
This series proposes to allow OP-TEE to offer a new watchdog service to non-secure world: extension of the watchdog programmable timeout. 

The need comes from some REE distributions that may take a long time between bootloader sequence completion and operating system installation. During that time, few mintues, operating system may not be able to refresh the watchdog. To overcome that, i watchdog service proposes to use a timer interrupt to refresh watchdog during the desired period.

This series starts with implementation of generic timer interrupt support for 32bit Arm architecture.
Follows changes in watchdog framework to implement the watchdog timeout extension.
Last commit proposes a PTA service for watchdog resources that mimics the services recently merged through a SMC interface.

### DIFF
--- a/core/arch/arm/kernel/sub.mk
+++ b/core/arch/arm/kernel/sub.mk
@@ -7,8 +7,10 @@ srcs-y += idle.c
 
 srcs-$(CFG_SECURE_TIME_SOURCE_CNTPCT) += tee_time_arm_cntpct.c
 srcs-$(CFG_SECURE_TIME_SOURCE_REE) += tee_time_ree.c
+ifeq ($(CFG_CORE_HAS_GENERIC_TIMER),y)
+srcs-$(CFG_ARM32_core) += timer_a32.c
+endif
 srcs-$(CFG_ARM64_core) += timer_a64.c
-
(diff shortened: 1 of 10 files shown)


jforissier:
Could this mechanism be made totally transparent to the caller? I mean:
- When the feature is enabled, make `watchdog_init()` return `LONG_MAX` in `*max_value`,
- Whenever `watchdog_settimeout()` is called with a value larger than the chip's max timeout, compute the number of internal refreshes needed (`n` times the chip's max timeout plus a remainder) and use that when `watchdog_start()` is called.

etienne-lms:
> Could this mechanism be made totally transparent to the caller? I mean:

Could do yes but it's not really handy when using Arm generic timer that hit only the core that armed the event. Using this timer makes some constraints when timeout can be extended. I initially made the feature to be available from boot time, while boot core remains executing. `watchdog_extend_timeout_timer_is_available()` queries when conditions are met. It's not really clean I admit.
An evolution would use a platform secure timer resource to tick the event, in a generic way. But there's no secure timer driver or interface defined.

PR #5206, opened 2022-03-06
ldelf: allow ELFs without DT_HASH to be loaded

parport0:
It is possible for an ELF to be compiled with --hash-style=gnu;
some distros even have that as their default. This produces an ELF
with no .hash section, but with a .gnu.hash section in its stead.

As a first step towards detaching table-traversal code from
the rest of the symbol resolution code, do not fail in the case
of a missing DT_HASH, but use the old bruteforce lookup code
when elf->hashtab is NULL.

Signed-off-by: Elvira Khabirova <e.khabirova@omp.ru>

### DIFF
--- a/ldelf/ta_elf.c
+++ b/ldelf/ta_elf.c
@@ -283,9 +283,11 @@ static void save_hashtab(struct ta_elf *elf)
 						  phdr[n].p_memsz);
 	}
 
-	check_hashtab(elf, elf->hashtab, 0, 0);
-	hashtab = elf->hashtab;
-	check_hashtab(elf, elf->hashtab, hashtab[0], hashtab[1]);
+	if (elf->hashtab) {
+		check_hashtab(elf, elf->hashtab, 0, 0);
+		hashtab = elf->hashtab;
+		check_hashtab(elf, elf->hashtab, hashtab[0], hashtab[1]);
+	}
 }
 
 static void save_soname_from_segment(struct ta_elf *elf, unsigned int type,
--- a/ldelf/ta_elf.c
+++ b/ldelf/ta_elf.c
@@ -218,6 +218,9 @@ static void save_hashtab_from_segment(struct ta_elf *elf, unsigned int type,
 		if (tag == DT_HASH) {
 			elf->hashtab = (void *)(val + elf->load_addr);
 			break;
+		} else if (tag == DT_GNU_HASH) {
+			elf->gnu_hashtab = (void *)(val + elf->load_addr);
+			break;
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


jenswi-linaro:
I'm not sure that we need this flexibility. Surely it must be possible to convince the toolchain to add a DT_HASH with some flag?

parport0:
`--hash-style=gnu` is used in some places for a reason; GNU-style hash sections have better performance than old SYSV-style ones (see https://www.akkadia.org/drepper/dsohowto.pdf). It is expected that, as time goes on, it becomes the new default.

jforissier:
IMO it would be preferable to add support for `DT_GNU_HASH`, would you want to give it a try? That being said, I am not totally opposed to this patch if it helps deal with problematic situations, but at a minimum ldelf should log a warning when defaulting to the inefficient symbol lookup.

jenswi-linaro:
With the commit "ldelf: add DT_GNU_HASH support" added, do we really need the previous commit "ldelf: allow ELFs without DT_HASH to be loaded"?

parport0:
Yeah, if the consensus here is to remove the fallback code, then I'll do it. I based the second commit on it as it performs some of the necessary separations.

The issue right now are the failing tests at IBART. I haven't managed to reproduce that yet. I'm going to need some time before I can get back to this.

jforissier (commented):
Hi @parport0, thanks for the update.

Like @jenswi-linaro said, since you have implemented DT_GNU_HASH, I don't think it makes much sense to keep the first commit.

I have found an issue on QEMUv8 with the `os_test` TA (for example `xtest 1022`), which links against a C++ code. With `--hash-style=gnu`, ldelf reports: "symbol __pthread_key_create not found". That symbol should not be an issue; it is weak undefined and should therefore be ignored. You can reproduce the issue with:
```diff

diff --git a/ta/arch/arm/link.mk b/ta/arch/arm/link.mk
index ec5c8db6..a0236de3 100644
--- a/ta/arch/arm/link.mk
+++ b/ta/arch/arm/link.mk
@@ -36,6 +36,7 @@ link-ldflags += $(call ld-option,-z force-bti) --fatal-warnings
 endif
 link-ldflags += --as-needed # Do not add dependency on unused shlib
 link-ldflags += $(link-ldflags$(sm))
+link-ldflags += --hash-style=gnu

 $(link-out-dir$(sm))/dyn_list:
        @$(cmd-echo-silent) '  GEN     $@'
```
...and `make run` as usual.

Some more comments below.

jforissier:
Hi @parport0,

I am following-up on this PR which I'd like to get merged eventually. If you can't spend more time on this it's OK, just let me know and I will carry on with a new PR. Thanks!

> I have found an issue on QEMUv8 with the `os_test` TA (for example `xtest 1022`), which links against a C++ code. With `--hash-style=gnu`, ldelf reports: "symbol __pthread_key_create not found".

That problem is addressed by https://github.com/OP-TEE/optee_os/pull/5373.

Another small fix is needed in this PR so that DT_GNU_HASH is used instead of DT_HASH when both are present and DT_HASH happens to be enumerated first:
```diff
diff --git a/ldelf/ta_elf.c b/ldelf/ta_elf.c
index f1ae475e..e7b4a2df 100644
--- a/ldelf/ta_elf.c
+++ b/ldelf/ta_elf.c
@@ -217,10 +217,8 @@ static void save_hashtab_from_segment(struct ta_elf *elf, unsigned int type,
                read_dyn(elf, addr, n, &tag, &val);
                if (tag == DT_HASH) {
                        elf->hashtab = (void *)(val + elf->load_addr);
-                       break;
                } else if (tag == DT_GNU_HASH) {
                        elf->gnu_hashtab = (void *)(val + elf->load_addr);
-                       break;
                }
        }
 }
```

PR #5207, opened 2022-03-08
dt driver: test for driver probing sequence

etienne-lms:
Implement _dt_driver_test.c_ with emulated resources (starting with clk's and rstctrl's) and a consumer device that gets those resources (possibly deferring its probe). A test DTSI defines the topology. After initcall completion, collect the status and let core self test (`xtest -t regression 1001`) report any failure (with a companion error traces).

Former change introduces `CFG_DRIVERS_CLK_EARLY_PROBE=y|n` to ease probe deferral (clocks not always early probe).
Note: can be useful when a clock depends on a pinctrl, probed after clock, but no occurrence in existing platforms.

Later change adds a missing description for CFG_DRIVERS_DT_RECURSIVE_PROBE.

The series default enables the test for qemu's platforms. The don't embed any DTB so use a template DTS file.

### DIFF
--- a/core/kernel/dt_driver_test.c
+++ b/core/kernel/dt_driver_test.c
@@ -1,0 +1,525 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2022, Linaro Limited
+ *
+ * Tests introduce dummy test drivers and assiciated devices defined in
+ * dt_driver_test.dtsi file with device resource dependencies.
+ */
+
+#include <assert.h>
+#include <config.h>
+#include <crypto/crypto.h>
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <drivers/rstctrl.h>
+#include <initcall.h>
+#include <kernel/dt.h>
+#include <kernel/dt_driver.h>
+#include <libfdt.h>
+#include <malloc.h>
+#include <sys/queue.h>
+#include <tee_api_defines_extensions.h>
+#include <tee_api_types.h>
+
+#define DT_TEST_MSG(...)	FMSG("(dt-driver-test) " __VA_ARGS__)
+
+/* Test state IDs */
+enum dt_test_sid { DEFAULT = 0, IN_PROGRESS, SUCCESS, FAILED };
+
+/*
+ * DT tests state to be reported from PTA_INVOKE_TESTS_CMD_DT_TEST_STATUS
+ * possibly printed to console. A test can be skipped (DEFAULT) or be
+ * successful (SUCCESS) orthewise it has failed (IN_PROGRESS, FAILED).
+ */
+struct dt_test_state {
+	enum dt_test_sid probe_deferral;
+	enum dt_test_sid probe_clocks;
+	enum dt_test_sid probe_resets;
+	enum dt_test_sid crypto_dependencies;
+};
+
+/*
+ * References allocated from heap to be free once test completed
+ * dt_test_alloc(), dt_test_free(), dt_test_free_all()
+ */
+struct dt_test_free_ref {
+	void *p;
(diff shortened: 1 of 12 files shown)


core/kernel/dt_driver_test.c
```diff
@@ -0,0 +1,529 @@
...
+	{ }
+};
+
+DEFINE_DT_DRIVER(dt_test_consumer_driver) = {
+	.name = "dt-test-crypt-consumer",
+	.match_table = dt_test_crypt_consumer_match_table,
+	.probe = dt_test_crypt_consumer_probe,
+};
+
+#ifdef CFG_DRIVERS_CLK
```
jenswi-linaro, line 326: Why is this ifdef needed?

core/kernel/dt_driver_test.c
```diff
@@ -0,0 +1,529 @@
...
+err:
+	dt_test_free(clk);
+	return res;
+}
+
+CLK_DT_DECLARE(dt_test_clock_provider, "linaro,dt-test-provider",
+	       dt_test_clock_provider_probe);
+#endif /* CFG_DRIVERS_CLK */
+
+#ifdef CFG_DRIVERS_RSTCTRL
```
jenswi-linaro, line 411: Why is this ifdef needed?

core/pta/tests/misc.c
```diff
@@ -548,14 +550,26 @@ static int self_test_nex_malloc(void)
...
+
 /* exported entry points for some basic test */
 TEE_Result core_self_tests(uint32_t nParamTypes __unused,
 		TEE_Param pParams[TEE_NUM_PARAMS] __unused)
 {
 	if (self_test_mul_signed_overflow() || self_test_add_overflow() ||
 	    self_test_sub_overflow() || self_test_mul_unsigned_overflow() ||
 	    self_test_division() || self_test_malloc() ||
-	    self_test_nex_malloc()) {
+	    self_test_nex_malloc() || self_test_dt_driver()) {
```
jenswi-linaro, line 572: I'd rather have this in a separate test case.
I suppose we can keep it in 1001, but via another subcase.

etienne-lms: ok

etienne-lms: Is https://github.com/OP-TEE/optee_test/pull/591 what you had in mind?

core/kernel/dt_driver_test.c
jenswi-linaro: Some flexibility is good. What happens without this ifdef, does something break?

core/pta/tests/misc.c
jenswi-linaro: Yes, thanks.

core/kernel/dt_driver_test.c
jenswi-linaro: OK, let's keep it then.

core/kernel/dt_driver_test.c
```diff
@@ -0,0 +1,529 @@
...
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct clk *clk = NULL;
+
+	DT_TEST_MSG("Register clocks");
+
+	clk = dt_test_alloc(DT_TEST_CLK_COUNT * sizeof(*clk));
+	if (!clk)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
```
jenswi-linaro, line 381:
``` C
for (n = 0; n < DT_TEST_CLK_COUNT; n++) {
...
```
should avoid these hard coded indexes below.

etienne-lms: ok

jenswi-linaro (commented):

etienne-lms:
Squashed fixup commits and  applied review tag.
Rebased (for CI).

Issue #5201, opened 2022-03-04
Possiable bug in core_mmu_add_mapping()?


tiger20081015:
Hi, we met an issue related to core_mmu_add_mapping with CFG_WITH_LPAE=y in armv7 platform. The details are:

1. We place optee start address to 0xe000_0000 by setting CFG_TZDRAM_START ?= 0xe0000000 and CFG_TZDRAM_SIZE  ?= 0x02000000.  We also set CFG_RESERVED_VASPACE_SIZE  to 1G for later mapping. 
With these settings, dump_mmap_table shows below mapping table:

_E/TC:0 0 dump_mmap_table:753 type TA_RAM       va 0x3c400000..0x3dffffff pa 0xe0400000..0xe1ffffff size 0x01c00000 (pgdir)
E/TC:0 0 dump_mmap_table:753 type IO_SEC       va 0x3e000000..0x3e0fffff pa 0x00200000..0x002fffff size 0x00100000 (smallpg)
E/TC:0 0 dump_mmap_table:753 type TEE_RAM_RX   va 0xe0000000..0xe00b1fff pa 0xe0000000..0xe00b1fff size 0x000b2000 (smallpg)
E/TC:0 0 dump_mmap_table:753 type TEE_RAM_RW   va 0xe00b2000..0xe03fffff pa 0xe00b2000..0xe03fffff size 0x0034e000 (smallpg)
E/TC:0 0 dump_mmap_table:753 type SHM_VASPACE  va 0xe0400000..0xe23fffff pa 0x00000000..0x01ffffff size 0x02000000 (pgdir)
E/TC:0 0 dump_mmap_table:753 type IO_SEC       va 0xe2400000..0xec3fffff pa 0x01000000..0x0affffff size 0x0a000000 (pgdir)
E/TC:0 0 dump_mmap_table:753 type IO_SEC       va 0xec400000..0xfc3fffff pa 0x70000000..0x7fffffff size 0x10000000 (pgdir)
**E/TC:0 0 dump_mmap_table:753 type RES_VASPACE  va 0xfc400000..0x3c3fffff pa 0x00000000..0x3fffffff size 0x40000000 (pgdir)**_

Please note the RES_VASPACE  overwraps 4G with "va 0xfc400000..0x3c3fffff" . 

2. Then our drivers call core_mmu_add_mapping with 8M length multi times.  But we met error only after 2~3 times. Debug log shows it is caused by below code in core_mmu_add_mapping:
```
[...]

PR #5214, opened 2022-03-10
core: mm: fix core virtual address range constraint in lpae

etienne-lms:
Removes constraint in LPAE implementation that expects core virtual
address range to fit a single block of the MMU base table used to
map the area. The implementation allows virtual addresses to spread
over several blocks of the base table. Therefor the limitation can be
removed.

As a result, selecting whether core virtual address range spreads
above or below the chosen base address depends on its location
against the virtual range mid address.

Fixes: https://github.com/OP-TEE/optee_os/issues/5201

### DIFF
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -813,13 +813,16 @@ void core_init_mmu(struct tee_mmap_region *mm)
 	assert(max_va < BIT64(CFG_LPAE_ADDR_SPACE_BITS));
 }
 
-bool core_mmu_place_tee_ram_at_top(paddr_t paddr)
+#ifdef CFG_WITH_PAGER
+/* Prefer to consume only 1 base xlat table for the whole mapping */
+bool core_mmu_prefer_tee_ram_at_top(paddr_t paddr)
 {
 	size_t base_level_size = BASE_XLAT_BLOCK_SIZE;
 	paddr_t base_level_mask = base_level_size - 1;
 
 	return (paddr & base_level_mask) > (base_level_size / 2);
 }
+#endif
 
 #ifdef ARM32
 void core_init_mmu_regs(struct core_mmu_config *cfg)
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -886,31 +886,27 @@ static void assign_mem_granularity(struct tee_mmap_region *memory_map)
 	}
 }
 
-static bool assign_mem_va(vaddr_t tee_ram_va,
-			  struct tee_mmap_region *memory_map)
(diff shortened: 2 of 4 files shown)


core/mm/core_mmu.c
```diff
@@ -886,6 +886,11 @@ static void assign_mem_granularity(struct tee_mmap_region *memory_map)
 	}
 }
 
+static bool place_tee_ram_at_top(paddr_t paddr)
+{
+	return paddr > core_mmu_get_va_width() / 2;
```
jenswi-linaro, line 891: BIT64(core_mmu_get_va_width()) ?

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -813,14 +813,6 @@ void core_init_mmu(struct tee_mmap_region *mm)
 	assert(max_va < BIT64(CFG_LPAE_ADDR_SPACE_BITS));
 }
 
-bool core_mmu_place_tee_ram_at_top(paddr_t paddr)
-{
-	size_t base_level_size = BASE_XLAT_BLOCK_SIZE;
-	paddr_t base_level_mask = base_level_size - 1;
-
-	return (paddr & base_level_mask) > (base_level_size / 2);
```
jenswi-linaro, line 821: With this we were trying to avoid crossing boundaries of the top level tables, typically 1G on Armv7.
Crossing such a boundary costs 4k since we need to use one more translation table.

Perhaps this small memory saving isn't relevant on LPAE enabled platforms.

etienne-lms:
review comments addressed

core/mm/core_mmu.c
```diff
@@ -1003,18 +1024,29 @@ static bool assign_mem_va(vaddr_t tee_ram_va,
...
 
 	return true;
+
+maybe_retry:
+	if (IS_ENABLED(CFG_WITH_PAGER) && !retry &&
+	    tee_ram_at_top != place_tee_ram_at_top(tee_ram_va)) {
+		/* Retry to setup mapping consumming an extra xlat table */
+		tee_ram_at_top = place_tee_ram_at_top(tee_ram_va);
+		retry = true;
+		goto define_vmap;
```
jenswi-linaro, line 1046: This is not how loops are implemented. Please find another way.

etienne-lms: i agree it is not clean. i'll update...

etienne-lms: I've split `assign_mem_va()` in 2 functions to ease integration.

jenswi-linaro (commented):

etienne-lms:
squashed fixup commits and applied review tag.
I rebased for CI dependency purpose.

Note I rephrased the commit message since original was not accurate.

PR #5215, opened 2022-03-11
FF-A map manifest region

jellesels-arm:
Map the device and memory regions from the SP manifest file into the SP memory space.
Iterate over all device and manifest regions in the manifest file. Map the regions inside the SP memory space.
With this patch the Trusted Service crypto can use the hardware TRNG.

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -134,15 +134,20 @@ bool sp_has_exclusive_access(struct sp_mem_map_region *mem,
 	return !sp_mem_is_shared(mem);
 }
 
+/*
+ * sp_init_info allocates and maps the sp_ffa_init_info for the SP. It will copy
+ * the fdt into the allocates page(s) and return a pointer to the new location
+ * of the fdt. This pointer can be used to update data inside the fdt.
+ */
 static TEE_Result sp_init_info(struct sp_ctx *ctx, struct thread_smc_args *args,
-			       const void * const fdt, vaddr_t *va,
-			       size_t *num_pgs)
+			       const void * const input_fdt, vaddr_t *va,
+			       size_t *num_pgs, void **fdt_copy)
 {
 	struct sp_ffa_init_info *info = NULL;
 	int nvp_count = 1;
 	size_t nvp_size = sizeof(struct sp_name_value_pair) * nvp_count;
 	size_t info_size = sizeof(*info) + nvp_size;
-	size_t fdt_size = fdt_totalsize(fdt);
+	size_t fdt_size = fdt_totalsize(input_fdt);
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t perm = TEE_MATTR_URW | TEE_MATTR_PRW;
 	struct fobj *fo = NULL;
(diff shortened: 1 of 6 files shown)


core/arch/arm/kernel/secure_partition.c
```diff
@@ -299,7 +301,28 @@ TEE_Result sp_map_shared(struct sp_session *s,
...
+
 	SLIST_FOREACH(reg, &smem->regions, link) {
+		/* Check device region attributes */
+		if (is_device) {
+			uint32_t mcache_type = 0;
+
+			if (mobj_get_cattr(reg->mobj, &mcache_type))
+				return FFA_INVALID_PARAMETERS;
+
+			if (cache_type != mcache_type) {
```
jenswi-linaro, line 377: Couldn't this check have been performed earlier when the smem was created? It seems that the smem is in an inconsistent state that never should have been allowed if this test fails.

jenswi-linaro (commented):
Comment for the commit "core: Allow SPs to map share device memory"

core/arch/arm/mm/sp_mem.c
```diff
@@ -126,12 +125,25 @@ void  sp_mem_set_cattr(struct mobj *mobj, uint32_t cache_type)
...
+
+	m->is_secure = is_secure;
+}
+
 static bool mobj_sp_matches(struct mobj *mobj __maybe_unused,
 			    enum buf_is_attr attr)
 {
+	struct mobj_sp *m = to_mobj_sp(mobj);
+
 	assert(mobj->ops == &mobj_sp_ops);
```
jenswi-linaro, line 158: This assert can be removed, it's done inside `to_mobj_sp()`.

core/arch/arm/mm/sp_mem.c
```diff
@@ -82,9 +84,6 @@ int sp_mem_add_pages(struct mobj *mobj, unsigned int *idx,
 	if (ADD_OVERFLOW(*idx, num_pages, &n) || n > tot_page_count)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	if (!core_pbuf_is(CORE_MEM_NON_SEC, pa, num_pages * SMALL_PAGE_SIZE))
```
jenswi-linaro, line 82: We will need some way of checking that the security state of the `mobj_sp` matches the one of the physical pages represented.

jenswi-linaro (commented):
Comments for "core: sp_mem: Allow secure memory mapping"

core/arch/arm/mm/sp_mem.c
jellesels-arm: Ack

core/arch/arm/mm/sp_mem.c
jellesels-arm: Ack

core/arch/arm/kernel/secure_partition.c
jellesels-arm: This check is not needed like you are saying, smem is the one which is setting the cache_type before.

jforissier:
Superseded by https://github.com/OP-TEE/optee_os/pull/5223.

Issue #5198, opened 2022-03-03
ftrace does not work because mobj_get_cattr function fails


zhanlej:
I found that the performance is very poor when using the RSA interface of OPTEE CORE, so I want to locate where the performance bottleneck is. I tried to open the ftrace function of OPTEE by referring to the [ftrace documentation](https://optee.readthedocs.io/en/latest/debug/ftrace.html), but I encountered some problems, which are described as follows:

I am using a chip from TI, and TI provides ti-optee-os modified based on optee_os. Although there are some differences between the two code bases, the core logic of OPTEE should not be modified, because the bl32 compiled by me using github's optee_os can also run normally.

Hardware version: TI TDA4
OPTEE OS version: [ti-optee-os tag ti2020.00](https://git.ti.com/cgit/optee/ti-optee-os/tag/?h=ti2020.00) (similar to version 3.8.0)
OPTEE CLIENT version: [ti-optee-client tag ti2020.00](https://git.ti.com/cgit/optee/ti-optee-client/tag/?h=ti2020.00)
OPTEE OS compilation parameters:

```
O=out/arm 
CFG_ARM64_core=y 
CROSS_COMPILE="/usr/bin/ccache /opt/gw-gcc/gcc/bin/aarch64-none-linux-gnu-" 
CROSS_COMPILE_core="/usr/bin/ccache /opt/gw-gcc/gcc/bin/aarch64-none-linux-gnu-" 
CROSS_COMPILE_ta_arm64="/usr/bin/ccache /opt/gw-gcc/gcc/bin/aarch64-none-linux-gnu-" 
CROSS_COMPILE_ta_arm32="/usr/bin/ccache /home/zlj/WorkSpace/icode/baidu/adu/tee_security/build/../toolchains/aarch32/bin/arm-linux-gnueabihf-" 
CFG_TEE_CORE_LOG_LEVEL=3 
CFG_TEE_TA_LOG_LEVEL=1 
DEBUG=1 
ta-targets=ta_arm64 
CFG_TEE_BENCHMARK=n 
CFG_TEE_TA_MALLOC_DEBUG=y 
CFG_FTRACE_SUPPORT=y 
CFG_ULIBS_MCOUNT=y 
CFG_TA_MCOUNT=y 
CFG_FTRACE_BUF_SIZE=100000  
PLATFORM=k3-j721e 
CFG_ARM64_core=y 
```

[...]

PR #5220, opened 2022-03-15
core: mm: fix mobj_shm_ops support .get_cattr.

zhanlej:
ftrace use static shared memory returns an object of type mobj_shm_ops.
But the get_cattr function is not implemented in mobj_shm_ops.This will
cause ftrace to not work properly.

Signed-off-by: Lejia Zhang <zhanlej@gmail.com>
Suggested-by: Sumit Garg <sumit.garg@linaro.org>

See #5198 for detail.

### DIFF
--- a/core/mm/mobj.c
+++ b/core/mm/mobj.c
@@ -396,6 +396,17 @@ static bool mobj_shm_matches(struct mobj *mobj __unused, enum buf_is_attr attr)
 	return attr == CORE_MEM_NSEC_SHM || attr == CORE_MEM_NON_SEC;
 }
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

zhanlej:
> 
> Please fix the checkpatch error.

Tag applied. And the checkpatch error fixed.

jforissier:
Thanks @zhanlej. I have modified the commit subject slightly, adding parentheses after `.get_cattr` to denote a function (it is the convention we use), and I have merged the patch.

PR #5224, opened 2022-03-21
TZASC minor fixes for the drivers and i.MX support

clementfaure:
Hello,

In this PR, I add the TZASC support for imx8m platforms that have some specific behavior regarding the regions base address.
I also do some minor fixes in the TZASC driver and imx TZASC driver.

Thanks!


core/arch/arm/plat-imx/drivers/tzc380.c
```diff
@@ -21,8 +22,30 @@
  */
 #ifndef TZASC2_BASE
 #define TZASC2_BASE			0
+#else
+register_phys_mem(MEM_AREA_IO_SEC, TZASC2_BASE, TZASC_SIZE);
 #endif
 
+register_phys_mem(MEM_AREA_IO_SEC, TZASC_BASE, TZASC_SIZE);
```
jforissier, line 29: How could this driver work without this? It doesn't seem platform-specific code was registering the memory either?
Should this be a separate commit?

jforissier (commented):
Hi @clementfaure, I have one question below. Plus two minor things:
- In the description for "drivers: imx: tzc380: add support for 8mscale platforms": "have an offset equals to" -> "have an offset relative to" perhaps?
- To avoid the checkpatch warnings, use Link: tags such as:
  ```
   Blah blah [1], [2] ...

   Link: [1] https://...
   Link: [2] https://...
   Signed-off-by: ...
   ```

core/arch/arm/plat-imx/drivers/tzc380.c
clementfaure: The TZASC registers are registered by https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-imx/main.c#L64
`AIPSx_BASE` and `AIPSx_SIZE` cover multiple peripheral/IP registers. That's how the TZASC registers were registered.

clementfaure:
Force-push: reformat commit message with `Link: [x] https://...`

core/arch/arm/plat-imx/drivers/tzc380.c
jforissier: Got it. So wouldn't it make sense to have one commit just for these two `register_phys_mem()`?

Anyways, for all commits:

clementfaure: Ok, no problem!

clementfaure:
Force-push:

- Rebase
- I remove some support for i.MX platforms, as it might change soon. I will submit the changes later in another PR

PR #5225, opened 2022-03-21
Add support for SNVS on imx6/7/8m platforms and enable OTPMK selection

clementfaure:
Hello,

Add the SNVS support for all imx6, imx7 and imx8m platforms.
To achieve that I choose to delete `imx_rpmb.c` driver that was part of the SNVS driver in the end. I've merged it with `imx_snvs.c`. I believe it makes more sense this way.

I've also added the selection of OTP key as master key for the CAAM when the device is closed.

Regards,
Clement


ldts (changes requested):
why does this need the CAAM crypto driver? when the SE05x is enabled, we do enable CAAM (ie CFG_NXP_CAAM=y and CFG_NXP_CAAM_RNG_DRV=y) but not the crytpo driver.

jforissier:
About commit "drivers: imx_snvs: fix SNVS security configuration values": it seems conditional compilation could be avoided and `if (IS_ENABLED(CFG_...))` used instead?

jforissier:
Commit "core: drivers: merge i.MX SNVS driver files" leaves the file `core/include/drivers/imx_snvs.h` essentially empty, this doesn't make much sense does it?

jforissier:
For commits:
- core: imx6: add SNVS_SIZE value
- core: imx7: add SNVS_SIZE value
- core: imx8m: add SNVS_SIZE value
- drivers: imx_snvs: fix SNVS_SSM_MODE_SECURE value
- core: imx: force CFG_IMX_UART to y for all imx6/7 and imx8m
- plat: imx: enable SNVS driver by default

Note, sometimes it would be nice to better explain "why" some changes are done, for example why force CFG_IMX_UART? Is it invalid to have it disabled?

clementfaure:
> About commit "drivers: imx_snvs: fix SNVS security configuration values": it seems conditional compilation could be avoided and `if (IS_ENABLED(CFG_...))` used instead?

Here if find `#ifdef ... #endif` cleaner but I can use `IS_ENABLED` if required

ldts:
> @ldts The SNVS master key is used by the CAAM to derive its blob key encryption key (BKEK). This key is then used by the CAAM to encapsulation and decapsulate blobs. Therefore, setting the OTP as master key in the SNVS without using the CAAM for crypto is pointless.

um, isn't it required for the HUK generation then?

jforissier:
> > Commit "core: drivers: merge i.MX SNVS driver files" leaves the file `core/include/drivers/imx_snvs.h` essentially empty, this doesn't make much sense does it?
> 
> Well ... it's true that I leave `imx_snvs.h` empty in this commit but I populate this file in a following commit. What I wanted to do here, was to keep the file with it previous copyright (that is not NXP). I'm not sure, if I'm allowed to delete a file with a copyright that is not mine and then re-create the file without it ...

Good question. I am tempted to say we should just delete the file, but let's preserve the copyrights as much as possible instead.

jforissier:
> > About commit "drivers: imx_snvs: fix SNVS security configuration values": it seems conditional compilation could be avoided and `if (IS_ENABLED(CFG_...))` used instead?
> 
> Here if find `#ifdef ... #endif` cleaner but I can use `IS_ENABLED` if required

OK with me.

core/drivers/imx_snvs.c
```diff
@@ -40,7 +56,85 @@ enum snvs_security_cfg {
...
+
+	if ((hp_lr & SNVS_HPLR_MKS_SL) || (lp_lr & SNVS_LPLR_MKS_HL))
+		return true;
+
+	return false;
+}
+
+/*
+ * Set the Master key to use OTPMK and lock it. Return true if the operation is
+ * successful. Return false if the master mey is already locked.
```
ldts, line 102: typo s/mey/key

core/drivers/imx_snvs.c
```diff
@@ -40,7 +56,85 @@ enum snvs_security_cfg {
...
+ */
+static bool is_otpmk_valid(void)
+{
+	uint32_t hp_status = 0;
+	vaddr_t base = core_mmu_get_va(SNVS_BASE, MEM_AREA_IO_SEC, SNVS_SIZE);
+
+	hp_status = io_read32(base + SNVS_HPSR);
+
+	if (hp_status & (SNVS_HPSR_OTPMK_ZERO | SNVS_HPSR_OTPMK_SYND)) {
+		EMSG("OTPK Key is not valid");
```
ldts, line 130: why do we need to return an error?

core/drivers/imx_snvs.c
```diff
@@ -61,6 +155,12 @@ static enum snvs_security_cfg snvs_get_security_cfg(void)
 	return SNVS_SECURITY_CFG_FAB;
 }
 
+static bool snvs_is_device_closed(void)
```
ldts, line 190: can we use this function on plat_rpmb_key_is_ready replacing the "security == SNVS_SECURITY_CFG_CLOSED"at the end of the function?

core/drivers/imx_snvs.c
```diff
@@ -96,3 +196,27 @@ bool plat_rpmb_key_is_ready(void)
...
+	/*
+	 * If the board is closed, setting the OTP as the master key should
+	 * not fail.
+	 */
+	if (!status & snvs_is_device_closed()) {
+		EMSG("Device is closed");
+		EMSG("Master Key cannot be set to OTPMK");
+		return TEE_ERROR_BAD_STATE;
+	}
+
```
ldts, line 252: is this function correct? 
When the OTPMK is not valid and the device is not closed, we are returning TEE_SUCCESS.

clementfaure:
> > @ldts The SNVS master key is used by the CAAM to derive its blob key encryption key (BKEK). This key is then used by the CAAM to encapsulation and decapsulate blobs. Therefore, setting the OTP as master key in the SNVS without using the CAAM for crypto is pointless.
> 
> um, isn't it required for the HUK generation then?

Yes it is used to generate the HUK in this function `caam_blob_mkvb_init()` which is called from `crypto_driver_init()`

ldts:
> > > @ldts The SNVS master key is used by the CAAM to derive its blob key encryption key (BKEK). This key is then used by the CAAM to encapsulation and decapsulate blobs. Therefore, setting the OTP as master key in the SNVS without using the CAAM for crypto is pointless.
> > 
> > 
> > um, isn't it required for the HUK generation then?
> 
> Yes it is used to generate the HUK in this function `caam_blob_mkvb_init()` which is called from `crypto_driver_init()`

So setting the OTP as the master key in the SNVS without using the CAAM for crypto is not totally pointless then :)

Anyway, it doesnt matter. It cant remain as it is, it will not impact the SE05x use case: the CAAM init function does get executed before the SE05x then takes over as crypto so it is all good in the end (I had forgotten about it). thanks for the info!

clementfaure:
> Anyway, it doesnt matter. It cant remain as it is, it will not impact the SE05x use case: the CAAM init function does get executed before the SE05x then takes over as crypto so it is all good in the end (I had forgotten about it). thanks for the info!

Yes, you're right. I misunderstood what you meant. The CAAM is initialized as usual by the crypto engine could be something else at runtime.

core/drivers/imx_snvs.c
clementfaure: Ok

core/drivers/imx_snvs.c
clementfaure: Because the OTP key fused is considered invalid by the SNVS and therefore, it cannot be used as a master key.

core/drivers/imx_snvs.c
clementfaure: Yes, we could. I will fix that

core/drivers/imx_snvs.c
clementfaure: Yes, this is correct.
When the device is not closed, the OTP key could be not fused yet (so all zeroes ...) triggering a key violation. So, in the open state, we simply ignore the error.
Plus, in the open state, the CAAM does not use the master key provided by the SNVS by a dummy test key instead.

core/drivers/imx_snvs.c
ldts: right but why do we need to write an error message to the console? the function is not failing, is just doing what is supposed to do isnt it? so the caller of the function could be expecting an invalid response and we would be writing an error out to the console...seems weird

clementfaure: Oh, you mean printing an error message? Well, I don't have a strong preference for this. It's true that's not really an error. I can remove it and downgrade the trace level to debug mode.

jforissier:
> > Note, sometimes it would be nice to better explain "why" some changes are done, for example why force CFG_IMX_UART? Is it invalid to have it disabled?
> 
> Having it disabled triggers a compilation error ... I will add it to the commit description

That link error could as well be fixed like so:
```diff
diff --git a/core/include/drivers/imx_uart.h b/core/include/drivers/imx_uart.h
index b9df517e..6fe294f9 100644
--- a/core/include/drivers/imx_uart.h
+++ b/core/include/drivers/imx_uart.h
@@ -29,6 +29,7 @@
 #ifndef IMX_UART_H
 #define IMX_UART_H

+#include <compiler.h>
 #include <types_ext.h>
 #include <drivers/serial.h>

@@ -37,6 +38,12 @@ struct imx_uart_data {
        struct serial_chip chip;
 };

+#ifdef CFG_IMX_UART
 void imx_uart_init(struct imx_uart_data *pd, paddr_t base);
+#else
+static inline void imx_uart_init(struct imx_uart_data *pd __unused,
+                                paddr_t base __unused)
+{}
+#endif

 #endif /* IMX_UART_H */
```
...but some changes in `core/arch/arm/plat-imx/conf.mk` are desirable anyways for consistency. I'm OK with your current approach of forcing it everywhere if you prefer.

clementfaure:
> ...but some changes in `core/arch/arm/plat-imx/conf.mk` are desirable anyways for consistency. I'm OK with your current approach of forcing it everywhere if you prefer.

Yes, you're right. Actually, I might include this fix later also but I rather do it in another PR.

core/drivers/imx_snvs.c
ldts: um but then, the function will return success but the master OTPMK will not have been set; the semantics will be violated making the code difficult to read IMO.

can we not return something else in that case and handle it with a comment?

clementfaure: Would something like this more readable?
```suggestion

TEE_Result imx_snvs_set_master_otpmk(void)
{
	bool status = false;

	/* The OTPMK can only be set on closed devices */
	if (!snvs_is_device_closed())
		return TEE_ERROR_BAD_STATE;

	if (is_otpmk_valid())
		return TEE_ERROR_BAD_STATE;

	if (is_otpmk_selected() && is_mks_locked())
		return TEE_SUCCESS;

	if (set_mks_otpmk())
		return TEE_SUCCESS;
	else
		return TEE_ERROR_BAD_STATE;
}
```

ldts: yes. something like that.

I personally would do something like this - and remove the static set_mks_otpmk() but what you did is the sort of thing I had in mind.

```
TEE_Result imx_snvs_set_master_otpmk(void){

	vaddr_t base = core_mmu_get_va(SNVS_BASE, MEM_AREA_IO_SEC, SNVS_SIZE);

	if (!snvs_is_device_closed())
		return TEE_ERROR_BAD_STATE;

	if (is_otpmk_valid())
		return TEE_ERROR_BAD_STATE;

	if (is_mks_locked()) {
		if (is_otpmk_selected())
			return TEE_SUCCESS;
		return TEE_ERROR_BAD_STATE;
	}

	/* set the key */
	io_setbits32(base + SNVS_HPCOMR, SNVS_HPCOMR_MKS_EN);
	io_clrbits32(base + SNVS_LPMKCR, SNVS_LPMKCR_MKCR_MKS_SEL);
	io_clrbits32(base + SNVS_HPLR, SNVS_HPLR_MKS_SL);
	io_setbits32(base + SNVS_LPLR, SNVS_LPLR_MKS_HL);

	return TEE_SUCCES;
}

```

clementfaure: I'm okay with your implementation. I will just keep the register access for the key select and lock behind `set_mks_otpmk()` function.

clementfaure:
Push
* Address @ldts comments

core/drivers/imx_snvs.c
```diff
@@ -2,37 +2,191 @@
...
+	if ((hp_lr & SNVS_HPLR_MKS_SL) || (lp_lr & SNVS_LPLR_MKS_HL))
+		return true;
+
+	return false;
+}
+
+/*
+ * Set the Master key to use OTPMK and lock it. Return true if the operation is
+ * successful. Return false if the master key is already locked.
```
Emantor, line 102: This does not return anything, function is `static void`, was the intention to check something here or was this changed during refactoring?

core/drivers/imx_snvs.c
```diff
@@ -2,37 +2,191 @@
...
+#endif
 
-	return SNVS_SECURITY_CFG_FAB;
+static bool snvs_is_device_closed(void)
+{
+	return (snvs_get_security_cfg() == SNVS_SECURITY_CFG_CLOSED);
 }
 
-enum snvs_ssm_mode snvs_get_ssm_mode(void)
+#ifdef CFG_RPMB_FS
```
Emantor, line 188: Shouldn't this compile without the `#ifdef`?

Emantor (changes requested):

core/drivers/imx_snvs.c
clementfaure: It was changed during the review. Thanks for noticing it

core/drivers/imx_snvs.c
clementfaure: `snvs_get_ssm_mode()` is only used in `plat_rpmb_key_is_ready()` and the latter is implemented only if `CFG_RPMB_FS=y`

clementfaure:
@Emantor @ldts 
I've squashed fixup commits for an eventual second review. Don't hesitate to add your `Acked-by`

clementfaure:
Force-push: fix the compilation warning

clementfaure:
Since no recent comments have been submitted, could we move forward with this PR?

jforissier:
For "core: drivers: merge i.MX SNVS driver files":

For "drivers: imx_snvs: use snvs_is_device_closed() for RPMB key status":

@Emantor @ldts would you mind taking a quick look and acknowledging especially the commits without a review tag yet?

clementfaure:
Hello @jforissier 
These following commits are still not tagged:
* `drivers: caam: set OTP as master key`
* `drivers: imx_snvs: fix SNVS security configuration values`
* `drivers: imx_snvs: add master key selection`

What's your suggestion on how to move forward with this PR? 

Force-push: tags applied and rebase.

core/drivers/crypto/caam/caam_ctrl.c
```diff
@@ -29,6 +30,14 @@ static TEE_Result crypto_driver_init(void)
 	/* Enable the CAAM Clock */
 	caam_hal_clk_enable(true);
 
+	/* Set OTP as master key if the platform is closed */
+	if (snvs_is_device_closed()) {
+		retresult = imx_snvs_set_master_otpmk();
+		if (retresult != TEE_SUCCESS &&
```
jforissier, line 36: `if (retresult && retresult != TEE_ERROR_NOT_IMPLEMENTED)`

core/drivers/imx_snvs.c
```diff
@@ -2,37 +2,188 @@
...
+	vaddr_t base = core_mmu_get_va(SNVS_BASE, MEM_AREA_IO_SEC, SNVS_SIZE);
+
+	hp_lr = io_read32(base + SNVS_HPLR);
+	lp_lr = io_read32(base + SNVS_LPLR);
+
+	if ((hp_lr & SNVS_HPLR_MKS_SL) || (lp_lr & SNVS_LPLR_MKS_HL))
+		return true;
+
+	return false;
+}
```
jforissier, line 98:
```C
static bool is_mks_locked(void)
{
	vaddr_t base = core_mmu_get_va(SNVS_BASE, MEM_AREA_IO_SEC, SNVS_SIZE);

	return io_read32(base + SNVS_HPLR) & SNVS_HPLR_MKS_SL ||
	       io_read32(base + SNVS_LPLR) & SNVS_LPLR_MKS_HL;
}                                                                         
```

core/drivers/imx_snvs.c
```diff
@@ -2,37 +2,188 @@
...
+	uint32_t hp_status = 0;
+	vaddr_t base = core_mmu_get_va(SNVS_BASE, MEM_AREA_IO_SEC, SNVS_SIZE);
+
+	hp_status = io_read32(base + SNVS_HPSR);
+
+	if (hp_status & (SNVS_HPSR_OTPMK_ZERO | SNVS_HPSR_OTPMK_SYND))
+		return false;
+
+	return true;
+}
```
jforissier, line 125: Same here, can be made more compact

jforissier (commented):
> Hello @jforissier 
> These following commits are still not tagged:
> * `drivers: caam: set OTP as master key`
> * `drivers: imx_snvs: fix SNVS security configuration values`
> * `drivers: imx_snvs: add master key selection`
>
> What's your suggestion on how to move forward with this PR? 

I'm fine with this, minor suggestions below. In any case you can apply:

...to the 3 commits.

core/drivers/crypto/caam/caam_ctrl.c
clementfaure: Ok

core/drivers/imx_snvs.c
clementfaure: I'm OK with that

core/drivers/imx_snvs.c
clementfaure: Ok, I will do.

jforissier:
Fine with me, you may squash the fix-ups.

ldts:
> For "core: drivers: merge i.MX SNVS driver files": `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`
> 
> For "drivers: imx_snvs: use snvs_is_device_closed() for RPMB key status": `Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`
> 
> @Emantor @ldts would you mind taking a quick look and acknowledging especially the commits without a review tag yet?

sorry never saw this notification.

clementfaure:
Hello @ldts,

The CAAM derives the HUK from the SNVS master key.
When closing the board, the SNVS goes into secure/trusted state, and it selects the OTP key as the master key, by default. In this pull-request, we ensure that OTP key selection is enforced and locked. It does not change the key selection.

If your closed devices have the OTP key selected as the master key (default behavior), then the generated HUK should remain the same. Please note that I have not tested this case my self.

clementfaure:
@ldts I will also gladly take your review tags for the commits you've checked

clementfaure:
@jforissier I guess we won't have any additional tags for this PR. Could we proceed to the merge?

jforissier:
@clementfaure yes. I have restarted the failed CI job (the error was unrelated to this PR and now fixed). When complete I will merge.

PR #5235, opened 2022-03-25
Spectre bhb workaround

jenswi-linaro:


jenswi-linaro:
I've tested this on QEMU emulating a Cortex-A72. These patches would normally use the Spectre-V2 workaround for this CPU so I had to hack `get_excp_vect()` to call `select_vector_wa_spectre_bhb(8)` instead of `select_vector_wa_spectre_v2()` to test the new exception vector and such.

I'm using the approach recommended at https://developer.arm.com/support/arm-security-updates/speculative-processor-vulnerability/spectre-bhb

jenswi-linaro:
Addressed comments and applied tags.

PR #5236, opened 2022-03-27
overlay: use the __symbols__ node in DTB to help overlay generation

ldts:
Summary
-------------
The following patchset reads the `__symbols__` node in a DTB passed in bootargs to help generating the kernel DTB overlay and avoid having to hardcode node paths (node's path-name are not stable and may change thus requiring firmware updates to handle those changes)

**Tested in imx8m - preparing changes for upstreaming to the imx-atf project**

iMX-ATF changes required to support this feature on iMX8m
--------------------------------------------------------------------------------
Trusted Firmware for iMX: https://source.codeaurora.org/external/imx/imx-atf

The imx8mm boot sequence is as follows:
imx8mm

The current imx8m  TF-A 1) does not expect any params from SPL 2) therefore [hardcodes](https://source.codeaurora.org/external/imx/imx-atf/commit/?h=imx_5.4.3_2.0.0&id=f1a195b5cce64365a7227557a9009a4f545aa02d) the addresses where the FDT overlay will be located for OP-TEE (BL32) and U-Boot (BL33) 3) and clears the FDT overlay memory region

This could be modified so that SPL handles the DTB it reads from a FIT to TF-A:

```
diff --git a/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c b/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c
index 049192880..66af6de5f 100644
--- a/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c
+++ b/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c
@@ -176,7 +176,10 @@ void bl31_early_platform_setup2(u_register_t arg0, u_register_t arg1,
        /* Make sure memory is clean */
        mmio_write_32(BL32_FDT_OVERLAY_ADDR, 0);
        bl33_image_ep_info.args.arg3 = BL32_FDT_OVERLAY_ADDR;
-   bl32_image_ep_info.args.arg3 = BL32_FDT_OVERLAY_ADDR;
+ /* bl32 could be receiving a DTB on input while also asked to generate
+  * an overlay
+  */
+ bl32_image_ep_info.args.arg3 = arg1 > 0 ? arg1 : BL32_FDT_OVERLAY_ADDR;
 #endif
 #endif

```

Then to actually receive the arguments TF-A could do:

1) if the HW/Bootrom allows for it, do the reset actions in SPL and then

```
diff --git a/plat/imx/imx8m/imx8mm/platform.mk b/plat/imx/imx8m/imx8mm/platform.mk
index 600a3c449..0161ff8e2 100644
--- a/plat/imx/imx8m/imx8mm/platform.mk
+++ b/plat/imx/imx8m/imx8mm/platform.mk
@@ -50,7 +50,7 @@ BL31_SOURCES          +=      plat/imx/common/imx8_helpers.S                  \
                                ${IMX_GIC_SOURCES}
 
 USE_COHERENT_MEM       :=      1
-RESET_TO_BL31              :=      1
+RESET_TO_BL31                :=      0
 A53_DISABLE_NON_TEMPORAL_HINT := 0
 
 ERRATA_A53_835769      :=      1

```

2) Alternatively (**preferred**) we could not clear the arguments passed by SPL while keeping RESET_TO_BL31. For instance just define a new macro for imx8m (KEEP_FIRST_BOOTLOADER_ARGUMENTS) and continue to enter via reset.

```
diff --git a/bl31/aarch64/bl31_entrypoint.S b/bl31/aarch64/bl31_entrypoint.S
index ed058648f..f62b60520 100644
--- a/bl31/aarch64/bl31_entrypoint.S
+++ b/bl31/aarch64/bl31_entrypoint.S
@@ -72,11 +72,14 @@ func bl31_entrypoint
         * there's no argument to relay from a previous bootloader. Zero the
         * arguments passed to the platform layer to reflect that.
         * ---------------------------------------------------------------------
-    */
+ */
+#ifndef KEEP_FIRST_BOOTLOADER_ARGUMENTS
        mov     x20, 0
        mov     x21, 0
        mov     x22, 0
        mov     x23, 0
+#endif
+

```

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -115,6 +115,15 @@ __weak unsigned long plat_get_aslr_seed(void)
 	return 0;
 }
 
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak const char * const *main_get_optee_reserved_drivers(size_t *num)
+{
+	if (num)
+		*num = 0;
+
+	return NULL;
+}
+
 /*
  * This function is called as a guard after each smc call which is not
  * supposed to return.
@@ -1133,6 +1142,138 @@ static int mark_tzdram_as_reserved(struct dt_descriptor *dt)
 				   CFG_TZDRAM_SIZE);
 }
 
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+static int add_descriptor_node(struct dt_descriptor *dt, const char *driver)
+{
+	unsigned int i = 0;
+	unsigned int j = 0;
+	char root_name[256] = { };
+	char buffer[512] = { };
+	char *item[128] = { };
+	char *root = NULL;
+	char *p = buffer;
+	int offs = 0;
+
+	offs = fdt_path_offset(dt->blob, driver);
+	if (offs >= 0)
+		return offs;
+
+	strncpy(buffer, driver, sizeof(buffer));
+
+	/* strok modifies the buffer */
+	root = strtok_r(p, "/", &p);
+	if (!root)
+		return -1;
+
+	for (i = 0; i < ARRAY_SIZE(item); i++) {
+		item[i] = strtok_r(p, "/", &p);
+		if (!item[i])
+			break;
+	}
+
+	if (i >= ARRAY_SIZE(item))
+		return -1;
+
+	/* add the root node name */
+	snprintf(root_name, sizeof(root_name), "/%s", root);
+	offs = fdt_path_offset(dt->blob, root_name);
+	if (offs < 0) {
+		offs = add_dt_path_subnode(dt, "/", root);
+		if (offs < 0)
+			return -1;
+	}
+
+	for (j = 0; j < i; j++) {
+		offs = fdt_add_subnode(dt->blob, offs, item[j]);
+		if (offs < 0)
+			return -1;
+	}
+
+	return offs;
+}
+#endif
+
+/*
+ * Guarantee that these drivers will not be used by either EL1 or EL0: set the
+ * device tree node status in the FDT and overlay to disabled and the
+ * secure-status to okay.
+ * Return values:
+ *    0: num_drivers have been reserved (in DTB and optionally in overlay)
+ *   -1: no __symbols__ in DTB, no HEAP left, error creating overlay nodes,
+ *       driver node not found in DTB (but present as a __symbol__),
+ */
+static int reserve_platform_drivers_dt_nodes(struct dt_descriptor *dt)
+{
+	static const char **drivers;
+	static size_t num_drivers;
+	const char * const *drvs = { NULL };
+	const char *path = NULL;
+	unsigned int i = 0;
+	int offs = 0;
+	int len = 0;
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	static bool init;
+
+	if (dt->is_overlay || init) {
+		/*
+		 * OP-TEE handles first a DTB populated with a __symbols__ node
+		 * passed in parameters. On the next (and last!) call to
+		 * this method, uses those symbols to generate the FDT overlay.
+		 */
+		goto handle;
+	}
+#endif
+	drvs = main_get_optee_reserved_drivers(&num_drivers);
+	if (!drvs || !num_drivers)
+		return 0;
+
+	offs = fdt_path_offset(dt->blob, "/__symbols__");
+	if (offs < 0)
+		return -1;
+
+	drivers = malloc(num_drivers * sizeof(drivers[0]));
+	if (!drivers)
+		return -1;
+
+	for (i = 0; i < num_drivers; i++) {
+		path = fdt_getprop_namelen(dt->blob, offs,
+					   drvs[i], strlen(drvs[i]), &len);
+		if (!path)
+			return -1;
+
+		drivers[i] = malloc(len);
+		if (!drivers[i])
+			return -1;
+
+		memcpy((void *)drivers[i], (void *)path, len);
+	}
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	init = true;
+handle:
+#endif
+	for (i = 0; i < num_drivers; i++) {
+		if (!drivers[i])
+			break;
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+		if (dt->is_overlay)
+			offs = add_descriptor_node(dt, drivers[i]);
+		else
+#endif
+			offs = fdt_path_offset(dt->blob, drivers[i]);
+
+		if (offs < 0)
+			return -1;
+
+		if (fdt_setprop_string(dt->blob, offs, "status", "disabled"))
+			return -1;
+
+		if (fdt_setprop_string(dt->blob, offs, "secure-status", "okay"))
+			return -1;
+	}
+
+	return 0;
+}
+
 static void update_external_dt(void)
 {
 	struct dt_descriptor *dt = &external_dt;
@@ -1145,6 +1286,9 @@ static void update_external_dt(void)
 
 	if (config_psci(dt))
 		panic("Failed to config PSCI");
+
+	if (reserve_platform_drivers_dt_nodes(dt))
+		panic("Failed to reserve drivers for OP-TEE exclusive access");
 
 #ifdef CFG_CORE_RESERVED_SHM
 	if (mark_static_shm_as_reserved(dt))
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -78,6 +78,9 @@ struct dt_descriptor {
 	void *blob;
 #ifdef _CFG_USE_DTB_OVERLAY
 	int frag_id;
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	bool is_overlay;
+#endif
 #endif
 };
 
@@ -652,6 +655,10 @@ static int add_dt_overlay_fragment(struct dt_descriptor *dt, int ioffs)
 	int offs;
 	int ret;
 
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	if (!dt->is_overlay)
+		return ioffs;
+#endif
 	snprintf(frag, sizeof(frag), "fragment@%d", dt->frag_id);
 	offs = fdt_add_subnode(dt->blob, ioffs, frag);
 	if (offs < 0)
@@ -670,6 +677,10 @@ static int init_dt_overlay(struct dt_descriptor *dt, int __maybe_unused dt_size)
 {
 	int fragment;
 
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	if (!dt->is_overlay)
+		return 0;
+#endif
 	if (IS_ENABLED(CFG_EXTERNAL_DTB_OVERLAY)) {
 		if (!fdt_check_header(dt->blob)) {
 			fdt_for_each_subnode(fragment, dt->blob, 0)
@@ -1287,6 +1298,9 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
  */
 void __weak boot_init_primary_late(unsigned long fdt)
 {
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	external_dt.is_overlay = false;
+#endif
 	init_external_dt(fdt);
 	tpm_map_log_area(get_external_dt());
 	discover_nsec_memory();
@@ -1310,8 +1324,16 @@ void __weak boot_init_primary_late(unsigned long fdt)
 		IMSG("Initializing virtualization support");
 		core_mmu_init_virtualization();
 	} else {
+		/* use runtime init calls to update DTB */
 		init_tee_runtime();
 	}
+
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
+	release_external_dt();
+	external_dt.is_overlay = true;
+	init_external_dt(CFG_NSEC_DTB_OVERLAY_ADDR);
+	update_external_dt();
+#endif
 	call_finalcalls();
 	IMSG("Primary CPU switching to normal world boot");
 }
--- a/lib/libutils/isoc/include/string.h
+++ b/lib/libutils/isoc/include/string.h
@@ -30,6 +30,7 @@ char *strstr(const char *big, const char *little);
 char *strcpy(char *dest, const char *src);
 char *strncpy(char *dest, const char *src, size_t n);
 char *strrchr(const char *s, int i);
+char *strtok_r(char *str, const char *delim, char **saveptr);
 
 void *memchr(const void *buf, int c, size_t length);
 
--- a/lib/libutils/isoc/newlib/strtok_r.c
+++ b/lib/libutils/isoc/newlib/strtok_r.c
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 1988 Regents of the University of California.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the University nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE REGENTS AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE REGENTS OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include <string.h>
+
+static char *
+__strtok_r (register char *s,
+	register const char *delim,
+	char **lasts,
+	int skip_leading_delim)
+{
+	register char *spanp;
+	register int c, sc;
+	char *tok;
+
+
+	if (s == NULL && (s = *lasts) == NULL)
+		return (NULL);
+
+	/*
+	 * Skip (span) leading delimiters (s += strspn(s, delim), sort of).
+	 */
+cont:
+	c = *s++;
+	for (spanp = (char *)delim; (sc = *spanp++) != 0;) {
+		if (c == sc) {
+			if (skip_leading_delim) {
+				goto cont;
+			}
+			else {
+				*lasts = s;
+				s[-1] = 0;
+				return (s - 1);
+			}
+		}
+	}
+
+	if (c == 0) {		/* no non-delimiter characters */
+		*lasts = NULL;
+		return (NULL);
+	}
+	tok = s - 1;
+
+	/*
+	 * Scan token (scan for delimiters: s += strcspn(s, delim), sort of).
+	 * Note that delim must have one NUL; we stop if we see that, too.
+	 */
+	for (;;) {
+		c = *s++;
+		spanp = (char *)delim;
+		do {
+			if ((sc = *spanp++) == c) {
+				if (c == 0)
+					s = NULL;
+				else
+					s[-1] = 0;
+				*lasts = s;
+				return (tok);
+			}
+		} while (sc != 0);
+	}
+	/* NOTREACHED */
+}
+
+char *
+strtok_r (register char *__restrict s,
+	register const char *__restrict delim,
+	char **__restrict lasts)
+{
+	return __strtok_r (s, delim, lasts, 1);
+}
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -439,22 +439,52 @@ CFG_DT ?= n
 CFG_DTB_MAX_SIZE ?= 0x10000
 
 # Device Tree Overlay support.
-# CFG_EXTERNAL_DTB_OVERLAY allows to append a DTB overlay into an existing
-# external DTB. The overlay is created when no valid DTB overlay is found.
-# CFG_GENERATE_DTB_OVERLAY allows to create a DTB overlay at external
-# DTB location.
-# External DTB location (physical address) is provided either by boot
-# argument arg2 or from CFG_DT_ADDR if defined.
+#
+# Use Case 1: CFG_GENERATE_DTB_OVERLAY, creates an overlay.
+#
+# Use Case 2: CFG_EXTERNAL_DTB_OVERLAY, applies an overlay into an existing DTB
+#             overlay. If no valid overlay is found, create it.
+#
+#             The  DTB location (physical address) is provided either by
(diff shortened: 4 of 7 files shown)


ldts:
@jforissier checkpatch seems to have a problem with _CFG_USE_DTB_OVERLAY (I didnt add this option btw)

jforissier:
> @jforissier checkpatch seems to have a problem with _CFG_USE_DTB_OVERLAY (I didnt add this option btw)

Yes, checkpatch assumes there is only one possible prefix for configuration variables (the default being `CONFIG_` as in the Linux kernel, we [set it](https://github.com/OP-TEE/optee_os/blob/3.16.0/.checkpatch.conf#L13) to `CFG_` for OP-TEE). Let's ignore the warning.

lib/libutils/isoc/include/string.h
```diff
@@ -30,6 +30,7 @@ char *strstr(const char *big, const char *little);
 char *strcpy(char *dest, const char *src);
 char *strncpy(char *dest, const char *src, size_t n);
 char *strrchr(const char *s, int i);
+char* strtok(char *str, const char *delims);
```
jforissier, line 33: I'd rather have `strtok_r()` instead, so it can be easily used in thread context.
Style: `char *strtok(...);`.

lib/libutils/isoc/strtok.c
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2022 Foundries Ltd <jorge@foundries.io>
+ */
+#include <stdlib.h>
+#include <string.h>
+
+char* strtok(char *str, const char *delims)
```
jenswi-linaro, line 8: Why not copy this function (or `strtok_r()` as suggested by @jforissier) from newlib instead?

lib/libutils/isoc/newlib/strtok_r.c
```diff
@@ -0,0 +1,101 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#undef __STRICT_ANSI__
+#include "_ansi.h"
+#include <string.h>
+
+static char *
+_DEFUN (__strtok_r, (s, delim, lasts, skip_leading_delim),
```
jforissier, line 35: The latest upstream version doesn't have this ugly `_DEFUN` anymore, I'd rather we use that instead (although we do have some older code too). https://github.com/bminor/newlib/blob/master/newlib/libc/string/strtok_r.c

ldts: sure, let me pull and update.

jforissier:
"libutils: Import strtok_r from newlib" is 

(if you don't mind and while you're at it, add parentheses after the function name in the subject and commit description: `strtok_r()` not `strtok_r`)

core/arch/arm/kernel/boot.c
```diff
@@ -652,6 +655,10 @@ static int add_dt_overlay_fragment(struct dt_descriptor *dt, int ioffs)
 	int offs;
 	int ret;
 
+#ifdef CFG_OVERLAY_ADDR
+	if (!dt->is_overlay)
+		return ioffs;
```
etienne-lms, line 669: should return an `FDT_ERR_xxx` value, or -1 as done at line 671.
(note -1 is not really nice: it relates to `FDT_ERR_NOTFOUND`. At least it makes the implementation consistent.

core/arch/arm/kernel/boot.c
```diff
@@ -1287,6 +1301,11 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
  */
 void __weak boot_init_primary_late(unsigned long fdt)
 {
+#ifdef CFG_OVERLAY_ADDR
+	struct dt_descriptor *dt = &external_dt;
+
+	dt->is_overlay = 0;
```
etienne-lms, line 1454: `external_dt.is_overlay = true;`

mk/config.mk
```diff
@@ -439,22 +439,52 @@ CFG_DT ?= n
...
+# Use Case 2: CFG_EXTERNAL_DTB_OVERLAY, applies an overlay into an existing DTB
+#             overlay. If no valid overlay is found, create it.
+#
+#             The  DTB location (physical address) is provided either by
+#             1. boot argument arg2.
+#             2. CFG_DT_ADDR.
+#
+# Use Case 3: CFG_OVERLAY_ADDR, creates an overlay at the physical address
+#             provided with this config AND updates the DTB.
+#             i,e: SPL passes the U-boot DTB address in params; and the kernel
```
etienne-lms, line 454: prefer `CFG_OVERLAY_ADDR` is renamed to something like `CFG_NSEC_DTB_OVERLAY_ADDR`, that is with `_DTB_OVERLAY_ADDR` rather than only `_OVERLAY_ADDR`.

mk/config.mk
```diff
@@ -439,22 +439,52 @@ CFG_DT ?= n
...
+#             CFG_DT_ADDR
+#
+#
 # A subsequent boot stage can then merge the generated overlay DTB into a main
 # DTB using the standard fdt_overlay_apply() method.
+
 CFG_EXTERNAL_DTB_OVERLAY ?= n
 CFG_GENERATE_DTB_OVERLAY ?= n
 
+ifeq ($(CFG_OVERLAY_ADDR),)
```
etienne-lms, line 468: `ifneq ($(CFG_OVERLAY_ADDR),y)`

core/arch/arm/kernel/boot.c
```diff
@@ -1310,8 +1329,16 @@ void __weak boot_init_primary_late(unsigned long fdt)
...
 	} else {
+		/* use runtime init calls to update DTB */
 		init_tee_runtime();
 	}
+
+#ifdef CFG_OVERLAY_ADDR
+	release_external_dt();
+	dt->is_overlay = 1;
+	init_external_dt(CFG_OVERLAY_ADDR);
+	update_external_dt();
```
etienne-lms, line 1487: complete the sequence with `release_external_dt()`.

mk/config.mk
```diff
@@ -439,22 +439,52 @@ CFG_DT ?= n
...
+#             The  DTB location (physical address) is provided either by
+#             1. boot argument arg2.
+#             2. CFG_DT_ADDR.
+#
+# Use Case 3: CFG_OVERLAY_ADDR, creates an overlay at the physical address
+#             provided with this config AND updates the DTB.
+#             i,e: SPL passes the U-boot DTB address in params; and the kernel
+#                  overlay address is received in the config.
+#
+#             Condition: do NOT configure
```
etienne-lms, line 457: maybe more explicit as: `Exclusive with CFG_EXTERNAL_DTB_OVERLAY, CFG_GENERATE_DTB_OVERLAY and CFG_DT_ADDR`.

mk/config.mk
```diff
@@ -439,22 +439,52 @@ CFG_DT ?= n
...
+
+ifneq ($(strip $(CFG_OVERLAY_ADDR)),)
+ifeq ($(CFG_EXTERNAL_DTB_OVERLAY),y)
+$(error Cannot implement OVERLAY_ADDR and EXTERNAL_DTB_OVERLAY)
+endif
+ifeq ($(CFG_GENERATE_DTB_OVERLAY),y)
+$(error Cannot implement OVERLAY_ADDR and GENERATE_DTB_OVERLAY)
+endif
+ifneq ($(strip $(CFG_DT_ADDR)),)
+$(error Cannot implement OVERLAY_ADDR and CFG_DT_ADDR)
```
etienne-lms, line 484: I think here `CFG_OVERLAY_ADDR` is still compatible with `CFG_DT_ADDR`. The later can point to the DTB loaded by earlier boot stage (as SPL) while `CFG_OVERLAY_ADDR` points to the DTB overlay binary generated by OP-TEE. Am I right?

core/arch/arm/kernel/boot.c
```diff
@@ -997,13 +997,17 @@ static int get_nsec_memory_helper(void *fdt, struct core_mmu_phys_mem *mem)
...
-	addr_size = fdt_address_cells(fdt, 0);
-	if (addr_size < 0)
-		return 0;
-
-	len_size = fdt_size_cells(fdt, 0);
-	if (len_size < 0)
-		return 0;
+	if (IS_ENABLED(_CFG_USE_DTB_OVERLAY)) {
+		len_size = sizeof(paddr_t) / sizeof(uint32_t);
+		addr_size = sizeof(paddr_t) / sizeof(uint32_t);
```
etienne-lms, line 1011: When `_CFG_USE_DTB_OVERLAY` is enabled and the target memory already stores a DTB, OP-TEE is expected to update this DTB. In that case, size/address cell counts should still be read from the DTB image.

etienne-lms (commented):
Comments for commits "core: overlay: process DTB and generate DTB overlay" 
and "core: calculate size/address cells with overlay".

"libutils: Import strtok_r() from newlib"

core/arch/arm/kernel/boot.c
ldts: this is taken care off by the boot final stuff - I think it is good as is.,

mk/config.mk
ldts: um not at the moment (see entry_a64.S) copying the value to x20... I could make it compatible with a little bit more of work but I'd rather have that as a separate enhancement.

clementfaure:
Hi @ldts,
Could you explain this change in ATF?
```
-RESET_TO_BL31              :=      1
+RESET_TO_BL31                :=      0
```

ldts:
> Hi @ldts, Could you explain this change in ATF?
> 
> ```
> -RESET_TO_BL31              :=      1
> +RESET_TO_BL31                :=      0
> ```

Sorry yes, this deserves an explanation.

I haven't posted the change yet for review since I am looking a bit more into the full set of changes needed in SPL; having said that the imx8m TF-A changes discussion should be independent from this PR IMO: I just used imx8m since is the board I had at hand. 

on imx8m SPL boots TF-A so if the bootrom starts SPL at the right exception level I see no reason why SPL cant take the necessary steps to warm boot TF-A.

At the moment, the change that gets me to a working Linux shell (ie all cores initialized) just requires the following instead of disabling RESET_TO_BL31 (which will require a bit more investigation)...I dont know for sure but starting with RESET_TO_BL31 might have been a shortcut to avoid SPL changes (I might be wrong but this has been done in the past)

```
diff --git a/bl31/aarch64/bl31_entrypoint.S b/bl31/aarch64/bl31_entrypoint.S
index ed058648f..f62b60520 100644
--- a/bl31/aarch64/bl31_entrypoint.S
+++ b/bl31/aarch64/bl31_entrypoint.S
@@ -72,11 +72,14 @@ func bl31_entrypoint
         * there's no argument to relay from a previous bootloader. Zero the
         * arguments passed to the platform layer to reflect that.
         * ---------------------------------------------------------------------
-    */
+ */
+#ifndef KEEP_FIRST_BOOTLOADER_ARGUMENTS
        mov     x20, 0
        mov     x21, 0
        mov     x22, 0
        mov     x23, 0
+#endif
+

```

core/arch/arm/kernel/boot.c
```diff
@@ -1153,6 +1293,9 @@ static void update_external_dt(void)
 	if (config_psci(dt))
 		panic("Failed to config PSCI");
 
+	if (reserve_platform_drivers_dt_nodes(dt))
```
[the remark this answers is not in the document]
ldts: Right that is one way of doing it. I think a better way would be have macros in some i`mxX-reserve.h` files.

We could also add runtime support without having to change much (add a line to boot_init_primary_late)..we would just need to make sure the reserve_drivers array is fully populated after the driver has initialized.I can add support for that if you see value?

core/arch/arm/plat-imx/main.c
```diff
@@ -115,6 +115,21 @@ void console_init(void)
 #endif
 }
 
+static const char *const reserved[] = {
+#if defined(CFG_NXP_CAAM)
+#if defined(CFG_MX8MM) || defined(CFG_MX8MQ) || defined(CFG_MX8MP)
```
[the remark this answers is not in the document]
ldts: I think they could be extended to all iMX8m products..

core/arch/arm/kernel/boot.c
etienne-lms: That's right.

etienne-lms (commented):
Comments for commit "arm: boot: reserve devices for exclusive access".

Note: init functions playing with DT  in _boot.c_ are really growing. I think they would deserve their own source file. What do you think? Could be addressed aside or after this P-R.

core/arch/arm/kernel/boot.c
```diff
@@ -1140,6 +1149,137 @@ static int mark_tzdram_as_reserved(struct dt_descriptor *dt)
...
+		if (dt->is_overlay)
+			offs = add_descriptor_node(dt, drivers[i]);
+		else
+#endif
+			offs = fdt_path_offset(dt->blob, drivers[i]);
+
+		if (offs < 0)
+			return -1;
+
+		if (!offs)
```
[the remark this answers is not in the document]
ldts: you are right. this is a bug. if it was the overlay it would have been ok (it meant it had already been reserved) But not for a DTB. will remove this check. thanks!

core/arch/arm/kernel/boot.c
```diff
@@ -1140,6 +1149,137 @@ static int mark_tzdram_as_reserved(struct dt_descriptor *dt)
...
+		/* OP-TEE handles first a DTB populated with a __symbols__ node
+		 * passed in parameters. Then, using those symbols, it generates
+		 * the overlay.
+		 */
+		goto handle;
+	}
+
+	offs = fdt_path_offset(dt->blob, "/__symbols__");
+	if (offs < 0)
+		goto handle;
```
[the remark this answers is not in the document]
ldts: 100% right. will fix.

core/arch/arm/kernel/boot.c
```diff
@@ -1140,6 +1149,137 @@ static int mark_tzdram_as_reserved(struct dt_descriptor *dt)
...
+
+	return offs;
+}
+#endif
+
+/* Guarantee that these drivers will not be used by either EL1 or EL0: set the
+ * device tree node status in the FDT and overlay to disabled and the
+ * secure-status to okay.
+ */
+static int reserve_platform_drivers_dt_nodes(struct dt_descriptor *dt)
```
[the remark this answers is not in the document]
ldts: I did follow the pattern used on similar functions invoked from the same context (ie `update_external_dt()`). I think we should keep it as int.

mk/config.mk
ldts: but CFG_OVERLAY_ADDR is an actual physical address

core/arch/arm/kernel/boot.c
```diff
@@ -1140,6 +1149,137 @@ static int mark_tzdram_as_reserved(struct dt_descriptor *dt)
...
+		goto handle;
+	}
+
+	offs = fdt_path_offset(dt->blob, "/__symbols__");
+	if (offs < 0)
+		goto handle;
+
+	drvs = main_get_optee_reserved_drivers(&num_drivers);
+	if (!drvs || !num_drivers)
+		return 0;
```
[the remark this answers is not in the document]
ldts: thanks, this would have introduced a nasty global regression.

core/arch/arm/kernel/boot.c
ldts: @etienne-lms um no I dont think so. If this is not the overlay when CFG_OVERLAY_ADDR is enabled, this could behave as if it was a DTB (ie, as if _CFG_USE_DTB_OVERLAY is not defined); and that code return offs.

core/arch/arm/kernel/boot.c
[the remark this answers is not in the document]
etienne-lms: Ok, but I think this is an inheritance from the first function that returned _libdft.h_ negative codes on error. Ok, let's consider moving to `TEE_Result` later, for all applicable functions.

ldts:
@etienne-lms thanks a lot for all your comments. I did push a first update but will continue with your review (you had comments that I didnt look into yet)

ldts:
even though all TF-A imx8m support is in Codeaurora I am proposing this change in trusted-firmware-a:
https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/14596

ldts:
@clementfaure @etienne-lms any further comments?

ldts:
@jforissier I realize everyone is busy with the release but just checking that this PR is just following its course and not forgotten? no rush from my end of course

jforissier:
@ldts I doubt the review can be completed before the freeze for 3.17.0-rc1 (Friday), so this will probably have to go in later.

ldts:
jfyi seems the TF-A PR is moving ahead.

ldts:
@jforissier so what should we do with this PR?

jforissier:
For commit "ci: azure: specify use case for CFG_OVERLAY_ADDR":
In the commit subject, s/CFG_OVERLAY_ADDR/CFG_NSEC_DTB_OVERLAY_ADDR/ then:

As for the other unreviewed commits, I'd like opinions form @etienne-lms and @clementfaure who commented already.

clementfaure:
Hello guys,
I've asked some folks in the BSP team to check what the impact of the ATF changes would be on imx8 bootflow (NXP internal tree). Depending on the result, we will use this feature or not.
@ldts If you want to go forward with this PR, you can remove `plat-imx: caam: iMX8m reserve job ring` as I cannot approve it for now.

core/arch/arm/plat-imx/main.c
```diff
@@ -115,6 +115,21 @@ void console_init(void)
...
+	"sec_jr0",
```
clementfaure, line 121: The job ring to disable is not always `jr0`. The job ring to disable is `CFG_JR_INDEX` defined in `crypto_conf.mk`

ldts:
@clementfaure ok thanks. I have removed the support for iMX from this PR. 
Since this is moving ahead https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/14596, adding support for the imx tf-a to pass the dtb with the symbols to op-tee should be trivial (as explained in this PR)...but no rush, I understand it might require some planning if agreed.

@jforissier I did add your Acked-by to the azure commit.

ldts:
change has been merged

ldts:
what is happening with this? when can a decision be made?
the alternative is to hardcode paths

clementfaure:
Hi @ldts,
I don't have any feedback from BSP maintainers about the option you introduced in your TF-A pull-request.
Just one question : could the two methods to modify the DTB overlay coexist in optee-os ?

ldts:
> Hi @ldts, I don't have any feedback from BSP maintainers about the option you introduced in your TF-A pull-request. Just one question : could the two methods to modify the DTB overlay coexist in optee-os ?

Right, the TF-A PR was merged some weeks ago so the functionality is upstream. But sure, both options could coexist: ie hardcoding paths or using the DTB. Just that hardcoding paths has a higher potential of failing (ie, an update could lead to a security issue if a path changes and OP-TEE cant reserve the device)

ldts:
@jenswi-linaro @jbech-linaro would appreciate an answer on this.

@ricardosalveti we might have to carry this patch off-tree to keep our system secure. maybe we should tag it.

jenswi-linaro:
@ldts, answer on what?

ldts:
> @ldts, answer on what?

TL;DR 
1) Should OP-TEE hardcode paths to dt nodes or 2) should OP-TEE use __symbols"
if 1) I'll drop this PR
if 2) when can this PR be merged - what is missing.

ldts:
> From what I've understood `2` is the preferred option. Any downside to that?

Platforms wanting to use 2 obviously must be able to pass the DTB to OP-TEE (ie, that requires a TF-A change on top of the one I already upstreamed for imx8. It is nothing major but someone needs to do it). If platforms cant pass a DTB then they wont be able to reserve - maybe for those platforms hardcoding the path could be an alternative (but this could be abused)

clementfaure:
> Platforms wanting to use 2 obviously must be able to pass the DTB to OP-TEE (ie, that requires a TF-A change on top of the one I already upstreamed for imx8.

When you say that, you mean that you tested your TF-A patch on imx8m, right? You didn't literally add the support to imx8m in upstream, right?

> It is nothing major but someone in NXP needs to do it.

It's not guaranteed that NXP will add the support in the TF-A. I guess the BSP team will review this option during the next rebase.

Modifying the overlay with __symbols is a great thing to have, for sure. However, my main concern is that the feature requires to have the right platform support in TF-A ... If we enforce the use of `__symbols` to edit the DTB overlay, in the end, `CFG_EXTERNAL_DTB_OVERLAY=y` would require `RESET_TO_BL31_WITH_PARAMS=0` in the TF-A ?

ldts:
> > Platforms wanting to use 2 obviously must be able to pass the DTB to OP-TEE (ie, that requires a TF-A change on top of the one I already upstreamed for imx8.
> 
> When you say that, you mean that you tested your TF-A patch on imx8m, right? You didn't literally add the support to imx8m in upstream, right?

I added generic support upstream to enable TF-A to pass the parameters to OP-TEE (applicable to the imx8 case)
https://github.com/ARM-software/arm-trusted-firmware/commit/25844ff728e4a0e5430ba2032457aba7b780a701

The TF-A changes required could be as simple as skipping the current hardcoding of the addresses when there is something being sent from TF-A (see below) I would have been open to help proposing alternatives at the time but I cant really spend more time on this (2 months is enough...as far as I am concerned I dont really care what they do now, no longer my problem, they should have engaged back then). 

```
diff --git a/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c b/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c
index 049192880..66af6de5f 100644
--- a/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c
+++ b/plat/imx/imx8m/imx8mm/imx8mm_bl31_setup.c
@@ -176,7 +176,10 @@ void bl31_early_platform_setup2(u_register_t arg0, u_register_t arg1,
        /* Make sure memory is clean */
        mmio_write_32(BL32_FDT_OVERLAY_ADDR, 0);
        bl33_image_ep_info.args.arg3 = BL32_FDT_OVERLAY_ADDR;
-   bl32_image_ep_info.args.arg3 = BL32_FDT_OVERLAY_ADDR;
+ /* bl32 could be receiving a DTB on input while also asked to generate
+  * an overlay
+  */
+ bl32_image_ep_info.args.arg3 = arg1 > 0 ? arg1 : BL32_FDT_OVERLAY_ADDR;
 #endif
 #endif

```
which is what I used to test on imx8mm.

> 
> > It is nothing major but someone in NXP needs to do it.
> 
> It's not guaranteed that NXP will add the support in the TF-A. I guess the BSP team will review this option during the next rebase.
> 

sure. the BSP team might wish to continue hardcoding stuff. IMO is not very flexible and in this case it kind of limits what OP-TEE has to offer but that is the platform's choice.

> Modifying the overlay with __symbols is a great thing to have, for sure. However, my main concern is that the feature requires to have the right platform support in TF-A ... If we enforce the use of `__symbols` to edit the DTB overlay, in the end, `CFG_EXTERNAL_DTB_OVERLAY=y` would require `RESET_TO_BL31_WITH_PARAMS=0` in the TF-A ?

Kind of depend on each platform. But for iMX yes, it needs that so the param registers are not cleared when RESET_TO_BL31 is enabled.

jenswi-linaro:
It seems that we need both options. If I understand this correctly, you're implementing option 2 here? If we take this then we can save option number 1 to whoever needs it enough to implement and upstream it.

ldts:
> It seems that we need both options. If I understand this correctly, you're implementing option 2 here? If we take this then we can save option number 1 to whoever needs it enough to implement and upstream it.

right.

core/arch/arm/kernel/boot.c
```diff
@@ -115,6 +115,15 @@ __weak unsigned long plat_get_aslr_seed(void)
 	return 0;
 }
 
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak const char * const *main_get_optee_reserved_drivers(size_t *num)
```
jenswi-linaro, line 119: I'd prefer `plat_get_reserved_drivers()`.

core/arch/arm/kernel/boot.c
```diff
@@ -1133,6 +1142,138 @@ static int mark_tzdram_as_reserved(struct dt_descriptor *dt)
...
+static int reserve_platform_drivers_dt_nodes(struct dt_descriptor *dt)
+{
+	static const char **drivers;
+	static size_t num_drivers;
+	const char * const *drvs = { NULL };
+	const char *path = NULL;
+	unsigned int i = 0;
+	int offs = 0;
+	int len = 0;
+#ifdef CFG_NSEC_DTB_OVERLAY_ADDR
```
jenswi-linaro, line 1214: Please use `if (IS_ENALBED(CFG_NSEC_DTB_OVERLAY_ADDR)` instead of ifdefs where possible.

jenswi-linaro (commented):
Comments for the commit "arm: boot: reserve devices for exclusive access"

ldts:
@jenswi-linaro before investing more time on this, is this going to be merged or are users going to end up hardcoding paths? I'd rather not waste the reviewers time (or my time) otherwise.

jforissier:
> I believe that at least @jforissier and I agree that this is useful.

Yes. And I have merged the `strtok_r()` commit.

PR #5237, opened 2022-03-28
Forwarding TPM event log to FF-A SPs

imre-kis-arm:
**The first three commits of this PR are being reviewed in a separate PR: #5223** 

Forward TPM event log to FF-A SPs, so the initial attestation service of trusted-services can access it.
The PR consists of two parts:

- Enable OP-TEE to read the external device tree from secure memory. If OP-TEE acts like a SEL-1 SPMC Trusted Firmware-A passes a device tree in the secure memory (TOS FW config) and it forwards the details of TPM event log buffer in this device tree. Compared to the "traditional" OP-TEE boot (non-FF-A) the standard "tpm_event_log_addr" property is being used instead of "tpm_event_log_sm_addr".
- Map the TPM event log into the FF-A SP if its manifest has an "arm,tpm_event_log" compatible node indicating its need for the event log.

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -12,6 +12,7 @@
 #include <kernel/spmc_sp_handler.h>
 #include <kernel/thread_private.h>
 #include <kernel/thread_spmc.h>
+#include <kernel/tpm.h>
 #include <kernel/ts_store.h>
 #include <ldelf.h>
 #include <libfdt.h>
@@ -525,6 +526,79 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
 	return TEE_SUCCESS;
 }
 
+static TEE_Result handle_tpm_event_log(struct sp_ctx *ctx, void *fdt)
+{
+	uint32_t perm = TEE_MATTR_URW | TEE_MATTR_PRW;
+	uint32_t dummy_size __maybe_unused = 0;
+	TEE_Result res = TEE_SUCCESS;
+	size_t page_count = 0;
+	struct fobj *f = NULL;
+	struct mobj *m = NULL;
+	vaddr_t log_addr = 0;
+	size_t log_size = 0;
+	int node = 0;
+
+	node = fdt_node_offset_by_compatible(fdt, 0, "arm,tpm_event_log");
+	if (node < 0)
+		return TEE_SUCCESS;
+
+	/* Checking the existence and size of the event log properties */
+	if (sp_dt_get_u64(fdt, node, "tpm_event_log_addr", &log_addr)) {
+		EMSG("tpm_event_log_addr not found or has invalid size");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	if (sp_dt_get_u32(fdt, node, "tpm_event_log_size", &dummy_size)) {
+		EMSG("tpm_event_log_size not found or has invalid size");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	/* Validating event log */
+	res = tpm_get_event_log_size(&log_size);
+	if (res)
+		return res;
+
+	if (!log_size) {
+		EMSG("Empty TPM event log was provided");
+		return TEE_ERROR_ITEM_NOT_FOUND;
+	}
+
+	/* Allocating memory area for the event log to share with the SP */
+	page_count = ROUNDUP_DIV(log_size, SMALL_PAGE_SIZE);
+
+	f = fobj_sec_mem_alloc(page_count);
+	m = mobj_with_fobj_alloc(f, NULL);
+	fobj_put(f);
+	if (!m)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	res = vm_map(&ctx->uctx, &log_addr, log_size, perm, 0, m, 0);
+	mobj_put(m);
+	if (res)
+		return res;
+
+	/* Copy event log */
+	res = tpm_get_event_log((void *)log_addr, &log_size);
+	if (res)
+		goto err_unmap;
+
+	/* Setting event log details in the manifest */
+	res = fdt_setprop_u64(fdt, node, "tpm_event_log_addr", log_addr);
+	if (res)
+		goto err_unmap;
+
+	res = fdt_setprop_u32(fdt, node, "tpm_event_log_size", log_size);
+	if (res)
+		goto err_unmap;
+
+	return TEE_SUCCESS;
+
+err_unmap:
+	vm_unmap(&ctx->uctx, log_addr, log_size);
+
+	return res;
+}
+
 static TEE_Result handle_fdt(const void * const fdt, const TEE_UUID *uuid)
 {
 	int len = 0;
@@ -593,6 +667,12 @@ static TEE_Result sp_init_uuid(const TEE_UUID *uuid, const void * const fdt)
 	res = handle_fdt_dev_regions(ctx, fdt_copy);
 	if (res)
 		goto out;
+
+	if (IS_ENABLED(CFG_CORE_TPM_EVENT_LOG)) {
+		res = handle_tpm_event_log(ctx, fdt_copy);
+		if (res)
+			goto out;
+	}
 
 	ts_pop_current_session();
 
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -2,6 +2,7 @@
 /*
  * Copyright (c) 2016, 2022 Linaro Limited
  * Copyright (c) 2014, STMicroelectronics International N.V.
+ * Copyright (c) 2022, Arm Limited and Contributors. All rights reserved.
  */
 
 #include <assert.h>
@@ -642,6 +643,15 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_NSEC_SHM:
 		return attr | TEE_MATTR_PRW | cached;
 	case MEM_AREA_EXT_DT:
+		/*
+		 * If CFG_MAP_EXT_DT_SECURE is enabled map the external device
+		 * tree as secure non-cached memory, otherwise, fall back to
+		 * non-secure mapping.
+		 */
+		if (IS_ENABLED(CFG_MAP_EXT_DT_SECURE))
+			return attr | TEE_MATTR_SECURE | TEE_MATTR_PRW |
+			       noncache;
+		fallthrough;
 	case MEM_AREA_IO_NSEC:
 		return attr | TEE_MATTR_PRW | noncache;
 	case MEM_AREA_IO_SEC:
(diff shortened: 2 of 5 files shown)


core/mm/core_mmu.c
```diff
@@ -642,6 +643,10 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_NSEC_SHM:
 		return attr | TEE_MATTR_PRW | cached;
 	case MEM_AREA_EXT_DT:
+	/* If CFG_DT_SECURE is disabled, fall back to non-secure mapping. */
+#ifdef CFG_DT_SECURE
```
jenswi-linaro, line 647: Give how this is used I'd prefer a bit more elaborate name, perhaps `CFG_MAP_EXT_DT_SECURE`.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,92 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
...
+		goto err;
+
+	if (!event_log_size) {
+		EMSG("Empty TPM event log was provided");
+		res = TEE_ERROR_ITEM_NOT_FOUND;
+		goto err;
+	}
+
+	/* Allocating memory area for the event log to share with the SP */
+	page_count = ROUNDUP(event_log_size, SMALL_PAGE_SIZE) / SMALL_PAGE_SIZE;
```
jenswi-linaro, line 574: Please use `ROUNDUP_DIV()` instead.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,92 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
...
+	struct mobj *m = NULL;
+	struct sp_ctx *ctx = NULL;
+	uint32_t perm = TEE_MATTR_URW | TEE_MATTR_PRW;
+
+	node = fdt_node_offset_by_compatible(fdt, 0, "arm,tpm_event_log");
+	if (node < 0)
+		return TEE_SUCCESS;
+
+	/* Checking the existence and size of the event log properties */
+	if (!fdt_getprop(fdt, node, "tpm_event_log_addr", &prop_len) ||
```
jenswi-linaro, line 548: Please use the new helper function `sp_dt_get_u64()`

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,92 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
...
+	    prop_len != sizeof(uint64_t)) {
+		EMSG("tpm_event_log_addr not found or has invalid size");
+		res = TEE_ERROR_BAD_FORMAT;
+		goto err;
+	}
+
+	if (!fdt_getprop(fdt, node, "tpm_event_log_size", &prop_len) ||
```
jenswi-linaro, line 555: Please use the new helper function `sp_dt_get_u32()`.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,92 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
...
+	    prop_len != sizeof(uint32_t)) {
+		EMSG("tpm_event_log_size not found or has invalid size");
+		res = TEE_ERROR_BAD_FORMAT;
+		goto err;
```
jenswi-linaro, line 559: Nothing special is happening at `err:` so you could just return the error directly instead. This comment applies to all the `goro err;` below too.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,92 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
 	return TEE_SUCCESS;
 }
 
+#ifdef CFG_CORE_TPM_EVENT_LOG
+static TEE_Result handle_tpm_event_log(void *fdt, struct sp_session *s)
+{
+	int node = 0;
+	TEE_Result res = TEE_SUCCESS;
+	int prop_len = 0;
+	vaddr_t event_log_addr = 0;
```
jenswi-linaro, line 535: `log_addr` and `log_size` would be easier to use and clear enough.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,92 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
 	return TEE_SUCCESS;
 }
 
+#ifdef CFG_CORE_TPM_EVENT_LOG
+static TEE_Result handle_tpm_event_log(void *fdt, struct sp_session *s)
+{
+	int node = 0;
```
jenswi-linaro, line 532: An upside down christmas tree isn't a requirement, but in cases like this it tends to be a bit easier on the eyes.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -594,6 +681,12 @@ static TEE_Result sp_init_uuid(const TEE_UUID *uuid, const void * const fdt)
 	if (res)
 		goto out;
 
+#ifdef CFG_CORE_TPM_EVENT_LOG
+	res = handle_tpm_event_log(fdt_copy, sess);
```
balint-dobszay-arm, line 685: nit: instead of `sess` you could pass `ctx` like in `handle_fdt_dev_regions()`

core/mm/core_mmu.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/mm/core_mmu.c
```diff
@@ -642,6 +643,13 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_NSEC_SHM:
 		return attr | TEE_MATTR_PRW | cached;
 	case MEM_AREA_EXT_DT:
+	/*
+	 * If CFG_MAP_EXT_DT_SECURE is enabled map the external device tree as
+	 * secure non-cached memory, otherwise, fall back to non-secure mapping.
+	 */
+#ifdef CFG_MAP_EXT_DT_SECURE
```
jenswi-linaro, line 650: Please use `if (IS_ENABLED(CFG_MAP_EXT_DT_SECURE))` instead.

core/mm/core_mmu.c
```diff
@@ -642,6 +643,13 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_NSEC_SHM:
 		return attr | TEE_MATTR_PRW | cached;
 	case MEM_AREA_EXT_DT:
+	/*
```
jenswi-linaro, line 646: This should probably be indented one more step.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -594,6 +670,12 @@ static TEE_Result sp_init_uuid(const TEE_UUID *uuid, const void * const fdt)
 	if (res)
 		goto out;
 
+#ifdef CFG_CORE_TPM_EVENT_LOG
```
jenswi-linaro, line 673: Please use `if (IS_ENABLED(CFG_CORE_TPM_EVENT_LOG))` instead

core/mm/core_mmu.c
imre-kis-arm: Done, also added fallthrough; to suppress compiler warning.

core/mm/core_mmu.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done, added empty handle_tpm_event_log implementation for CFG_CORE_TPM_EVENT_LOG=n.

imre-kis-arm:
I've rebased to patch on master as #5223 got merged.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,87 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
 	return TEE_SUCCESS;
 }
 
+#ifdef CFG_CORE_TPM_EVENT_LOG
```
jenswi-linaro, line 529: I looks like this ifdef could be removed and leave it to the compile to optimize it away. That we we still have this function compile tested even without CFG_CORE_TPM_EVENT_LOG.

imre-kis-arm: Sounds reasonable, done.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -525,6 +526,79 @@ static TEE_Result handle_fdt_dev_regions(struct sp_ctx *ctx, void *fdt)
 	return TEE_SUCCESS;
 }
 
+static TEE_Result handle_tpm_event_log(struct sp_ctx *ctx, void *fdt)
+{
+	uint32_t perm = TEE_MATTR_URW | TEE_MATTR_PRW;
+	TEE_Result res = TEE_SUCCESS;
+	uint32_t dummy_size = 0;
```
etienne-lms, line 533: __maybe_unused

etienne-lms (commented):
for commit "core: Enable mapping DT from secure memory".

for commit "core: FF-A: Map TPM event log for FF-A SPs" with minor comment.

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

jenswi-linaro:

imre-kis-arm:
The only change I did was adding the Reviewed-by tag to the commit message and now the test times out. Do you have ideas why this would happen?
I'm also trying to run the same tests locally but it will take some time.

jforissier:
@imre-kis-arm all looks good now unless I missed something, can I merge this?

imre-kis-arm:
Yes, you can merge it, I think all comments have been addressed. Thank you for the quick review of this PR!

PR #5239, opened 2022-03-29
ta: pkcs11: Add padding for aes_cbc encryption

cmcgcc:
Implemented pkcs#7 padding at the end of aes_cbc encryption.

Tested-by: Mengchi Cheng <mengcc@amazon.com>

### DIFF
--- a/ta/pkcs11/src/processing_symm.c
+++ b/ta/pkcs11/src/processing_symm.c
@@ -25,6 +25,11 @@ struct input_data_ref {
 	void *data;
 };
 
+struct aec_cbc_padding {
+	uint32_t last_block_size;
+	uint8_t pad_block[TEE_AES_BLOCK_SIZE];
+};
+
 bool processing_is_tee_symm(enum pkcs11_mechanism_id proc_id)
 {
 	switch (proc_id) {
@@ -533,6 +538,7 @@ init_tee_operation(struct pkcs11_session *session,
 {
 	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
 	uint32_t *pkcs11_data = NULL;
+	struct aec_cbc_padding *pad_block = NULL;
 
 	switch (proc_params->id) {
 	case PKCS11_CKM_AES_CMAC:
@@ -584,7 +590,6 @@ init_tee_operation(struct pkcs11_session *session,
 		rc = PKCS11_CKR_OK;
 		break;
 	case PKCS11_CKM_AES_CBC:
-	case PKCS11_CKM_AES_CBC_PAD:
 	case PKCS11_CKM_AES_CTS:
 		if (proc_params->size != 16)
 			return PKCS11_CKR_MECHANISM_PARAM_INVALID;
@@ -592,6 +597,21 @@ init_tee_operation(struct pkcs11_session *session,
 		TEE_CipherInit(session->processing->tee_op_handle,
 			       proc_params->data, 16);
 		rc = PKCS11_CKR_OK;
+		break;
+	case PKCS11_CKM_AES_CBC_PAD:
+		if (proc_params->size != 16)
+			return PKCS11_CKR_MECHANISM_PARAM_INVALID;
+
+		pad_block = TEE_Malloc(sizeof(struct aec_cbc_padding),
+				       TEE_MALLOC_FILL_ZERO);
+		if (!pad_block)
+			return PKCS11_CKR_DEVICE_MEMORY;
+
+		session->processing->extra_ctx = pad_block;
+		TEE_CipherInit(session->processing->tee_op_handle,
+			       proc_params->data, 16);
+		rc = PKCS11_CKR_OK;
+
 		break;
 	case PKCS11_CKM_AES_CTR:
 		rc = tee_init_ctr_operation(session->processing,
@@ -730,6 +750,9 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
 	uint32_t hmac_len = 0;
 	uint8_t computed_mac[TEE_MAX_HASH_SIZE] = { 0 };
 	uint32_t computed_mac_size = TEE_MAX_HASH_SIZE;
+	struct aec_cbc_padding *last_block = proc->extra_ctx;
+	int out_size_temp = 0;
+	size_t in_size_padded = 0;
 
 	if (TEE_PARAM_TYPE_GET(ptypes, 1) == TEE_PARAM_TYPE_MEMREF_INPUT) {
 		in_buf = params[1].memref.buffer;
@@ -808,7 +831,6 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
 
 	case PKCS11_CKM_AES_ECB:
 	case PKCS11_CKM_AES_CBC:
-	case PKCS11_CKM_AES_CBC_PAD:
 	case PKCS11_CKM_AES_CTS:
 	case PKCS11_CKM_AES_CTR:
 		if (step == PKCS11_FUNC_STEP_FINAL ||
@@ -826,6 +848,45 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
 			res = TEE_CipherUpdate(proc->tee_op_handle,
 					       in_buf, in_size,
 						out_buf, &out_size);
+			output_data = true;
+			rc = tee2pkcs_error(res);
+			break;
+		default:
+			TEE_Panic(function);
+			break;
+		}
+		break;
+
+	case PKCS11_CKM_AES_CBC_PAD:
+		if (step == PKCS11_FUNC_STEP_FINAL ||
+		    step == PKCS11_FUNC_STEP_ONESHOT)
+			break;
+
+		if (!in_buf) {
+			EMSG("No input data");
+			return PKCS11_CKR_ARGUMENTS_BAD;
+		}
+
+		switch (function) {
+		case PKCS11_FUNCTION_ENCRYPT:
+			assert(proc->extra_ctx);
+
+			res = TEE_CipherUpdate(proc->tee_op_handle,
+					       in_buf, in_size,
+					       out_buf, &out_size);
+
+			output_data = true;
+			rc = tee2pkcs_error(res);
+
+			/* Update text size still in the OS buffer. */
+			if (rc == PKCS11_CKR_OK)
+				last_block->last_block_size = (in_size + last_block->last_block_size) & 0xF;
+
+			break;
+		case PKCS11_FUNCTION_DECRYPT:
+			res = TEE_CipherUpdate(proc->tee_op_handle,
+					       in_buf, in_size,
+					       out_buf, &out_size);
 			output_data = true;
 			rc = tee2pkcs_error(res);
 			break;
@@ -947,7 +1008,6 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
 
 	case PKCS11_CKM_AES_ECB:
 	case PKCS11_CKM_AES_CBC:
-	case PKCS11_CKM_AES_CBC_PAD:
 	case PKCS11_CKM_AES_CTS:
 	case PKCS11_CKM_AES_CTR:
 		if (step == PKCS11_FUNC_STEP_ONESHOT && !in_buf) {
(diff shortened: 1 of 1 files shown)


cmcgcc:
Hello,
I made this as a draft because it is only encryption part.
I have not got time to finish the decryption part. I would probably realize it in a similar way.
So I think it might not be bad to collect feedback first.

The test is conducted by running aws example code with various plaintext length and chunk size for C_EncryptUpdate.
https://github.com/aws-samples/aws-cloudhsm-pkcs11-examples/blob/master/src/encrypt/aes_cbc.c

ta/pkcs11/src/pkcs11_token.h
```diff
@@ -141,6 +141,7 @@ enum pkcs11_proc_state {
  * @tee_hash_op_handle - handle on active hashing crypto operation or
  * TEE_HANDLE_NULL
  * @extra_ctx - context for the active processing
+ * @last_block_size - size of the last block for padding purpose.
```
etienne-lms, line 144: This field should be allocated and referenced using `extra_ctx` field.
A block could also be allocated, to store padded bytes, so that it.s allocated from init and won't fail at final stage due to memory exhaustion.

ta/pkcs11/src/processing_symm.c
```diff
@@ -835,6 +837,43 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+
+			output_data = true;
+			rc = tee2pkcs_error(res);
+
+			/* Update text size still in the os buffer. */
+			if (rc == PKCS11_CKR_OK)
+				proc->last_block_size = (in_size + proc->last_block_size) & 0xF;
+
+			break;
+		case PKCS11_FUNCTION_DECRYPT:
```
etienne-lms, line 864: Prefer not introduce this if AES_PAD decryption is not supported

ta/pkcs11/src/processing_symm.c
```diff
@@ -969,6 +1007,51 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+	case PKCS11_CKM_AES_CBC_PAD:
+		if (step == PKCS11_FUNC_STEP_ONESHOT && !in_buf) {
+			EMSG("No input data");
+			return PKCS11_CKR_ARGUMENTS_BAD;
+		}
+
+		switch (function) {
+		case PKCS11_FUNCTION_ENCRYPT:;
+			in_size_padded = ((in_size + TEE_AES_BLOCK_SIZE) & ~0xF)
+                                              - (proc->last_block_size & 0xF);
```
etienne-lms, line 1019: Minus sign `-` at end of previous line.

etienne-lms (commented):
Hi @cmcgcc, thanks for the proposal. 
Since there's no decryption yet, i'm not sure this change will be merged for 3.17.0 tag.
If so, are you ok we merge https://github.com/OP-TEE/optee_os/pull/5154 for the tag and add AES_PAD afterward?

jforissier:
> Since there's no decryption yet, i'm not sure this change will be merged for 3.17.0 tag.

I plan to create the 3.17.0-rc1 tag in a few hours and it does not make much sense to merge encryption without decryption indeed.

ta/pkcs11/src/processing_symm.c
cmcgcc: > Hi @cmcgcc, thanks for the proposal. Since there's no decryption yet, i'm not sure this change will be merged for 3.17.0 tag. If so, are you ok we merge #5154 for the tag and add AES_PAD afterward?

Sorry for the late reply. I made some change to address the comments. I am working on the decryption part and will patch later.

ta/pkcs11/src/processing_symm.c
```diff
@@ -969,6 +1030,68 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+	case PKCS11_CKM_AES_CBC_PAD:
+		if (step == PKCS11_FUNC_STEP_ONESHOT && !in_buf) {
+			EMSG("No input data");
+			return PKCS11_CKR_ARGUMENTS_BAD;
+		}
+
+		switch (function) {
+		case PKCS11_FUNCTION_ENCRYPT:
+			if (!proc->extra_ctx)
+				return PKCS11_CKR_ARGUMENTS_BAD;
```
etienne-lms, line 1042: Rather `PKCS11_CKR_GENERAL_ERROR` (or even `TA_Panic()`) , This is not expect.

ta/pkcs11/src/processing_symm.c
```diff
@@ -969,6 +1030,68 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+					output_data = true;
+					break;
+				}
+				out_size_temp = out_size;
+				res = TEE_CipherUpdate(proc->tee_op_handle,
+						       in_buf, in_size,
+						       out_buf, &out_size);
+
+				output_data = true;
+				rc = tee2pkcs_error(res);
```
etienne-lms, line 1057: prefer:
```c
				rc = tee2pkcs_error(res);
				if (rc)
 					break;

				/* Update text size still in the os buffer. */
				last_block->last_block_size = in_size & 0xF;
				out_buf += out_size;
				out_size = out_size_temp - out_size;
				out_size_temp -= out_size;
			}

File: ta/pkcs11/src/processing_symm.c [hunk 6]
```diff
@@ -969,6 +1030,68 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+					last_block->last_block_size = in_size & 0xF;
+					out_buf += out_size;
+					out_size = out_size_temp - out_size;
+					out_size_temp -= out_size;
+				}
+				else
+					break;
+			}
+
+			in_size_padded = TEE_AES_BLOCK_SIZE - (last_block->last_block_size & 0xF);
```
etienne-lms, line 1070: can remove `& 0xF`, this value is always <= 15.

ta/pkcs11/src/processing_symm.c
```diff
@@ -533,6 +538,7 @@ init_tee_operation(struct pkcs11_session *session,
 {
 	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
 	uint32_t *pkcs11_data = NULL;
+	struct aec_cbc_padding *pad_block = NULL;
```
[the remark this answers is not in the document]
cmcgcc: void *extra_ctx = NULL; ? Could you elaborate a bit?

ta/pkcs11/src/processing_symm.c
```diff
@@ -969,6 +1030,68 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+			return PKCS11_CKR_ARGUMENTS_BAD;
+		}
+
+		switch (function) {
+		case PKCS11_FUNCTION_ENCRYPT:
+			if (!proc->extra_ctx)
+				return PKCS11_CKR_ARGUMENTS_BAD;
+
+			if (step == PKCS11_FUNC_STEP_ONESHOT) {
+				if (out_size < ((in_size & ~0xF) + TEE_AES_BLOCK_SIZE)){
```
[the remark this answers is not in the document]
cmcgcc: The out size should be 32. The extra bock saves 0x10 padding.

ta/pkcs11/src/processing_symm.c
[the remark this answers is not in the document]
etienne-lms: well... weird idea of mine. This is fine. Sorry for the noise.

ta/pkcs11/src/processing_symm.c
etienne-lms: hi @cmcgcc. thanks for the update. you can close/hide these comment thread now if you want.

ta/pkcs11/src/processing_symm.c
```diff
@@ -835,6 +857,46 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+
+		if (!in_buf) {
+			EMSG("No input data");
+			return PKCS11_CKR_ARGUMENTS_BAD;
+		}
+
+		switch (function) {
+		case PKCS11_FUNCTION_ENCRYPT:
+			if (!proc->extra_ctx)
+				return PKCS11_CKR_ARGUMENTS_BAD;
```
etienne-lms, line 873: `assert(proc->extra_ctx);`

ta/pkcs11/src/processing_symm.c
etienne-lms: hmm... `assert(proc->extra_ctx);` is far enough

ta/pkcs11/src/processing_symm.c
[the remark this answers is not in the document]
etienne-lms: right, sorry.
(please fix missing space before `{`, it will hurt checkpatch)

ta/pkcs11/src/processing_symm.c
```diff
@@ -584,14 +590,28 @@ init_tee_operation(struct pkcs11_session *session,
...
 		TEE_CipherInit(session->processing->tee_op_handle,
 			       proc_params->data, 16);
 		rc = PKCS11_CKR_OK;
+		break;
+	case PKCS11_CKM_AES_CBC_PAD:
+		if (proc_params->size != 16)
+			return PKCS11_CKR_MECHANISM_PARAM_INVALID;
+
+		pad_block = TEE_Malloc(sizeof(struct aec_cbc_padding),
+							 TEE_MALLOC_FILL_ZERO);
```
etienne-lms, line 606: indentation:
```c
		pad_block = TEE_Malloc(sizeof(struct aec_cbc_padding),
					TEE_MALLOC_FILL_ZERO);

File: ta/pkcs11/src/processing_symm.c [hunk 11]
```diff
@@ -835,6 +857,45 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+			assert(proc->extra_ctx);
+
+			res = TEE_CipherUpdate(proc->tee_op_handle,
+					       in_buf, in_size,
+					       out_buf, &out_size);
+
+			output_data = true;
+			rc = tee2pkcs_error(res);
+
+			/* Update text size still in the os buffer. */
```
etienne-lms, line 881: c/os/OS/
ditto at line 1059.

etienne-lms (commented):
encryption part looks good to me.
remaining minor comments.

ta/pkcs11/src/processing_symm.c
```diff
@@ -969,6 +1029,65 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+			EMSG("No input data");
+			return PKCS11_CKR_ARGUMENTS_BAD;
+		}
+
+		switch (function) {
+		case PKCS11_FUNCTION_ENCRYPT:
+			assert(proc->extra_ctx);
+
+			if (step == PKCS11_FUNC_STEP_ONESHOT) {
+				if (out_size < ((in_size & ~0xF) + TEE_AES_BLOCK_SIZE)) {
```
etienne-lms, line 1043: we must conform to 80char/line. Maybe introduce a local variable:
```c
				pad_sz = (in_size & ~0xF) + TEE_AES_BLOCK_SIZE;
```

ta/pkcs11/src/processing_symm.c
```diff
@@ -835,6 +857,45 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+			res = TEE_CipherUpdate(proc->tee_op_handle,
+					       in_buf, in_size,
+					       out_buf, &out_size);
+
+			output_data = true;
+			rc = tee2pkcs_error(res);
+
+			/* Update text size still in the OS buffer. */
+			if (rc == PKCS11_CKR_OK)
+				last_block->last_block_size = (in_size + last_block->last_block_size) & 0xF;
```
etienne-lms, line 883: Break line to fit in 80 char/line.
It does not produce a nice layout but we must conform to optee_os coding style.
Maybe 
```c
				last_block->last_block_size =
					(in_size +
					 last_block->last_block_size) & 0xF;
```

ta/pkcs11/src/processing_symm.c
```diff
@@ -969,6 +1029,65 @@ enum pkcs11_rc step_symm_operation(struct pkcs11_session *session,
...
+			out_size += out_size_temp;
+
+			output_data = true;
+			rc = tee2pkcs_error(res);
+
+			break;
+		case PKCS11_FUNCTION_DECRYPT:
+			res = TEE_CipherDoFinal(proc->tee_op_handle,
+						in_buf, in_size,
+						out_buf, &out_size);
```
etienne-lms, line 1082: My question may be stupid: is this implementation expected to successfully decrypt an AES CBC padded cipher text? I guess some post-processing is needed, maybe more.
If this piece of code is not expected to be reviewed, could you remove implementation of case`PKCS11_FUNCTION_DECRYPT` until you expect we review it.

etienne-lms (changes requested):

ta/pkcs11/src/processing_symm.c
cmcgcc: Sorry for the delay. I am having the decrypt under internal review right now. Is there a way to reopen this pull request?
The problem for decrypt I could not solve is without decryption it can only return the out_size including padding. Is it acceptable?

etienne-lms: For sure, decryption should return the right clear text size, excluding padded bytes artifact.
The issue I foresee is that caller is required to provide an output buffer that is large enough to hold those padded bytes, even if effective output buffer size will exclude these extra bytes. The weird thing is that a caller knowing the expected end size (without padding) and providing a buffer that excludes them will get a TEE_ERROR_SHORT_BUFFER he/she would not expect. This could be an compatibility issue with existing software packages that don't handle this.

A way out could be PKCS11 TA to always decrypt last block (at each step/update) into a private secure buffer and copy to effective output buffer when expected. That would add a bit of logic in the implementation but would behave better with any client implementation.

Another way out could be to manage that from cryptoki client library that would decrypt AES_CBC_PAD in temporary buffer, enlarged by 1 AES block.

cmcgcc: A temp buffer at client side might be the way to go. If my memory is not wrong, AES_CBC decryption normally needs cypher text of previous block. Only decrypt the last cypher block would not return correct plain text.

cmcgcc: Did some experiments. I do not think C_DecryptFinal can be used to remove padding. It does not take encrypt bytes. Meanwhile the encrypt text length of C_DecryptUpdate has to be multiple of 16, which means the decrypt buff length should be same as encrypt text length or else CKR_BUFFER_TOO_SMALL will return. So the decrypt buffer length of C_DecryptFinal can only be 0.

PR #5241, opened 2022-03-29
arm: aspeed: fix kernel panic while getting gicd & gicc va

Neal-liu:
608016966 commit unify GIC initialization flow into common
gic code and get GIC distributor/CPU interface virtual addresses
with 64KB granularity.

However, Aspeed SoC hardware design only used 4KB granularity for
each of them. Revise register GICD/GICC physical memory size to meet
gic init requirement. (from 4KB to 64KB)
This commit would result in memory map overlaps warning.

D/TC:0   add_phys_mem:554 GIC_BASE + GICC_OFFSET type IO_SEC 0x40462000 size 0x00010000
D/TC:0   add_phys_mem:554 GIC_BASE + GICD_OFFSET type IO_SEC 0x40461000 size 0x00010000
D/TC:0   add_phys_mem:568 Physical mem map overlaps 0x40461000


jenswi-linaro:
We could perhaps define `GIC_DIST_REG_SIZE` and `GIC_CPU_REG_SIZE` as `SMALL_PAGE_SIZE` if `CFG_ARM_GICV3=n`. However, that would leave `GICC_DIR` unmapped. I'm not sure that's a problem though. What does others think? @etienne-lms @jforissier ?

jforissier:
> '608016966' commit unify GIC initialization flow into common

@Neal-liu the proper way to cite a previous commit is:
```
Commit 60801696667d ("plat: arm: refactor GIC initialization") unifies
GIC initialization flow...
```
I have the following alias in my `~/.gitconfig`, quite convenient IMO:
```
[alias]
	cite = show --abbrev=12 --format='commit %h (\"%s\")' -s
```

Neal-liu:
@jforissier Thanks for sharing. This is really helpful. I'll revise it for the upcoming patchset.
@jenswi-linaro Is it better to define it as config and let platform owners to define its own setting?
For example, somewhere common mk:
```
GIC_DIST_REG_SIZE ?= 0x10000
GIC_CPU_REG_SIZE ?= 0x10000
```

core/arch/arm/plat-aspeed/conf.mk
```
GIC_CPU_REG_SIZE = 0x1000
GIC_CPU_REG_SIZE = 0x1000
```

jenswi-linaro:
We would like to configure as much as possible by default. Since OP-TEE isn't using `GICC_DIR` it should be safe to define small values in case of GICv2 which I assume we're using with the Aspeed SoC.

etienne-lms:
@Neal-liu, FYI, "Physical mem map overlaps" trace message is not a warning trace. It is only debug information, harmless.

It is true GICD and GICC IOMEM sizes used to map these registers should match the size values used in `gic_init_base_addr()` to retrieve their base VA  (at least be higher or equal). If some platforms have specific IOMEM sizes, either passed these to `gic_init_base_addr()` or use a CFG_GIC_REG_SIZE config switch.

Neal-liu:
Currently, I'll follow the common code to register GICD and GICC IOMEM sizes as 64KB and leave the debug message alone since it's harmless. Once `gic_init_base_addr()` or somewhere can customize the size, I'll have corresponding modification.
If this is workable for you, I'll update this PR for further review.

jenswi-linaro:
Sounds good. Please apply:

Neal-liu:
> Sounds good. Please apply:

Tag applied.
Thanks for the help on reviewing the patch and suggestions.

PR #5243, opened 2022-03-30
Fix device-tree assigned-clocks properties parsing

clementleger:
This small series fixes parsing problems when using `assigned-clocks` properties:
- Wrong check for assigned-clock-rates property
- Handling of phandle == 0

### DIFF
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -266,6 +266,13 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 	while (idx < len) {
 		idx32 = idx / sizeof(uint32_t);
 		phandle = fdt32_to_cpu(prop[idx32]);
+		if (!phandle) {
+			if (!prop_idx)
+				break;
+			idx += sizeof(phandle);
+			prop_idx--;
+			continue;
+		}
 
 		prv = dt_driver_get_provider_by_phandle(phandle, type);
 		if (!prv) {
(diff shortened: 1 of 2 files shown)


core/kernel/dt_driver.c
```diff
@@ -266,6 +266,11 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 	while (idx < len) {
 		idx32 = idx / sizeof(uint32_t);
 		phandle = fdt32_to_cpu(prop[idx32]);
+		if (phandle == 0) {
```
jenswi-linaro, line 269: While this style isn't rejected, please don't mix it with the other style in the same function.
So make this `if (!phandle)...`

etienne-lms (commented):

core/kernel/dt_driver.c
clementleger: Indeed, I'll fixed that.

clementleger:
New version:
- Use `(!phandle)`.
- Add early return, it worked previously due to some wrong check... (prop_idx being an unsigned, - 1, is not a value we want to handle...)

@etienne-lms I did not added your `Reviewed-by` on the first commit since I modified it.

jenswi-linaro:

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for the 2 commits.

PR #5245, opened 2022-04-01
plat-stm32mp1: expose SCMI voltage domains for some PMIC regulators

etienne-lms:
Defines the DT binding ID and SCMI resources for SCMI voltage domain exposed by platform stm32mp1 when secure hardeing of the resources are enabled.


core/include/drivers/stpmic1_regulator.h
```diff
@@ -0,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2021, STMicroelectronics
+ */
+
+#ifndef STPMIC1_REGULATOR_H
```
jenswi-linaro, line 6: __DRIVERS_STPMIC1_REGULATOR_H?

core/include/drivers/stpmic1_regulator.h
```diff
@@ -0,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2021, STMicroelectronics
```
jenswi-linaro, line 3: 2022 also?

core/include/drivers/stpmic1_regulator.h
etienne-lms: right.

jenswi-linaro:

PR #5248, opened 2022-04-04
TPM2 fix

ruchi393:

PR #5249, opened 2022-04-04
Eventlog Parsing and PCR Extension in TPM

ruchi393:
The patch set  adds patches to parse the eventlog passed to OP-TEE and extend the PCR's of a TCG TPM provider is registered.
The framework also provides generic framework for devices other than TPM to register as TCG PCR providers.
The eventlog parsing code and TCG things which are originally authored by me and @apalos submitted to u-boot code are re-used over here.

The code has been tested on QEMU with sw-tpm as backend. I plan to send further patches in
- TF-A to add support to create eventlog framework in QEMU [ I have local patches for that right now]
- to optee build repo to help user replicate the demo


ruchi393:
The initial 2 patches in the PR are part of another PR - https://github.com/OP-TEE/optee_os/pull/5248, so please ignore them during review.

ruchi393:
@ldts , I have added a generic PCR provider for eventlog. Please have a look and let me know if it works.

@apalos , can you also help review.

ldts:
> @ldts , I have added a generic PCR provider for eventlog. Please have a look and let me know if it works.

thanks @ruchi393. I'll check but it seems good (need to switch my board)

core/drivers/tpm2/tpm2_chip.c
```diff
@@ -28,13 +31,186 @@ enum tpm2_result tpm2_chip_recv(uint8_t *buf, uint32_t *len,
...
+
+/*
+ * Get the following information from TPM and store it in chip's capability
+ * structure.
+ * Number of banks available in TPM
+ * Number of active banks and the corresponding digest algo
+ */
+static enum tpm2_result tpm2_get_bank_info(struct tpm2_chip *chip)
+{
+	uint32_t i, j = 0;
```
etienne-lms, line 92: 1 definition per line

core/drivers/tpm2/tpm2_chip.c
```diff
@@ -57,5 +233,24 @@ enum tpm2_result tpm2_chip_register(struct tpm2_chip *chip)
 	if (ret)
 		EMSG("TPM2 Self Test Failed");
 
+	ret = tpm2_populate_capability(chip);
+	if (!ret)
+		tpm2_dump_capability(chip);
+
+	/* Register TPM2 as tcg provider */
```
etienne-lms, line 240: s/tcg/TCG/ ?

core/drivers/tpm2/tpm2_cmd.c
```diff
@@ -147,3 +147,284 @@ enum tpm2_result tpm2_selftest(uint8_t full)
...
+enum tpm2_result tpm2_get_capability(uint32_t capability, uint32_t property,
+				     uint32_t prop_cnt, void *prop,
+				     uint32_t *prop_len)
+{
+	uint8_t buf[32] = { };
+	uint32_t buf_len = sizeof(buf);
+	uint8_t *resp_buf = NULL;
+	uint32_t resp_len = 256;
+	uint32_t prop_offset = 0;
+	enum tpm2_result rc = TPM2_OK;
```
etienne-lms, line 160: s/rc/ret/g for consistency?

core/drivers/tpm2/tpm2_cmd.c
```diff
@@ -147,3 +147,284 @@ enum tpm2_result tpm2_selftest(uint8_t full)
...
+	tpm2_cmd_add_u32(buf, buf_len, prop_cnt);
+
+	rc = tpm2_transmit(buf, tpm2_cmd_len((struct tpm2_cmd *)buf), resp_buf,
+			   &resp_len);
+	if (rc)
+		goto out;
+
+	resp_len = tpm2_cmd_len((struct tpm2_cmd *)resp_buf);
+
+	DHEXDUMP(resp_buf, tpm2_cmd_len((struct tpm2_cmd *)resp_buf));
```
etienne-lms, line 182: Remove?

core/include/tpm2.h
```diff
@@ -29,6 +36,125 @@
...
+/* The following set of ABI constants were chosen by the TSS Working Group.
+ * They represent reasonable, future-proof values.
+ */
+#define TPM2_NUM_PCR_BANKS	16
+#define TPM2_MAX_PCRS		32
+#define TPM2_PCR_SELECT_MAX	((TPM2_MAX_PCRS + 7) / 8)
+
+/* Table 78 - Definition of TPMU_HA Union */
+union tpmu_ha {
+	uint8_t sha[TPM2_SHA_DIGEST_SIZE];
```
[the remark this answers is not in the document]
ruchi393: Not directly, but this is needed to determine the max size of digestlist which is allocated in tpml_digest_values hen parsing the eventlog.

ruchi393:
PR updated with review comments addressed.

ruchi393:
@etienne-lms , @jenswi-linaro , @jforissier , @apalos , can you help review the PR

core/include/tpm2.h
```diff
@@ -29,6 +36,125 @@
...
+/* Table 101 TPMS_TAGGED_PROPERTY */
+struct tpms_tagged_property {
+	uint32_t property;
+	uint32_t value;
+} __packed;
+
+/* Table 113 TPML_TAGGED_TPM_PROPERTY */
+struct tpml_tagged_tpm_property {
+	uint32_t count;
+	struct tpms_tagged_property tpm_property[];
```
jenswi-linaro, line 132: I can't make sense of this. What would `p->tpm_property[1]` mean? `struct tpms_tagged_property` has also a flexible array memory at the end.

core/include/tpm2.h
```diff
@@ -29,6 +36,125 @@
...
+struct tpms_pcr_selection  {
+	uint16_t hash;
+	uint8_t size_of_select;
+	uint8_t pcr_select[];	/* Max index TPM2_PCR_SELECT_MAX */
+} __packed;
+
+/* Table 111 - Definition of TPML_PCR_SELECTION Structure */
+struct tpml_pcr_selection {
+	uint32_t count;
+	struct tpms_pcr_selection pcr_selections[]; /* Max TPM2_NUM_PCR_BANKS */
```
jenswi-linaro, line 120: I can't make sense of this. What would `p->pcr_selections[1]` mean? `struct tpms_pcr_selection` also has a flexible array memory at the end.

core/include/kernel/tcg.h
```diff
@@ -0,0 +1,163 @@
...
+ * @event_size: Size of event
+ * @event:      Event data
+ */
+struct tcg_pcr_event {
+	uint32_t pcr_index;
+	uint32_t event_type;
+	uint8_t digest[TPM2_SHA1_DIGEST_SIZE];
+	uint32_t event_size;
+	uint8_t event[];
+} __packed;
```
jenswi-linaro, line 32: The `__packed` attributes make access to the struct inefficient since it also means that the compiler should generate code for unaligned access. In some cases it may be a reasonable compromise,  but is it really needed here?

core/kernel/tcg.c
```diff
@@ -0,0 +1,393 @@
...
+	struct tpml_digest_values *digest_list = NULL;
+
+	if (!pcr_provider) {
+		EMSG("No provider available for PCR's");
+		return TEE_ERROR_GENERIC;
+	}
+
+	buffer = malloc(TPM2_EVENT_LOG_SIZE);
+	if (!buffer) {
+		EMSG("Error allocating mem\n");
```
jenswi-linaro, line 305: `\n` not needed.

core/kernel/tcg.c
```diff
@@ -0,0 +1,393 @@
...
+ * Return:		status code
+ * @pos			Offset in the eventlog where the specID event ends
+ * @digest_list:	list of digests in the event
+ */
+static TEE_Result tcg_parse_specid_event(void *buffer, uint32_t log_size,
+					 uint32_t *pos,
+					 struct tpml_digest_values *digest_list)
+{
+	struct tcg_efi_spec_id_event *spec_event = NULL;
+	struct tcg_pcr_event *event_header = (struct tcg_pcr_event *)buffer;
```
jenswi-linaro, line 119: cast is not needed.

core/include/tpm2.h
ruchi393: Good catch. I am currently hard coding count as 1 in the commands, so didn't face the issue. I will change elements in both the structures from variable length array to fixed size arrays.

core/include/kernel/tcg.h
ruchi393: Since digest arrays size is fixed to 20 bytes, not having this __packed won;t have any impact. I will remove it

core/kernel/tcg.c
ruchi393: Sorry, I didn't notice it. I will remove it from here as well as other occurrence in the file for \n

ruchi393:
PR updated with comments addressed

core/kernel/tcg.c
```diff
@@ -0,0 +1,393 @@
...
+ * @buffer:		Pointer to the start of the eventlog
+ * @log_size:		Size of the eventlog
+ * @pos:		[in] Offset of specID event in the eventlog buffer
+ *			[out] Return offset of the next event in the buffer
+ *			after the specID
+ * @digest_list:	list of digests in the event
+ *
+ * Return:		status code
+ * @pos			Offset in the eventlog where the specID event ends
+ * @digest_list:	list of digests in the event
```
etienne-lms, line 112: remove the 2 lines above

core/drivers/tpm2/tpm2_chip.c
```diff
@@ -28,13 +31,185 @@ enum tpm2_result tpm2_chip_recv(uint8_t *buf, uint32_t *len,
...
+		return true;
+
+	return false;
+}
+
+/* Get value of property (TPM_PT) from capability TPM_CAP_TPM_PROPERTIES */
+static enum tpm2_result tpm2_get_tpm_property(uint32_t property, uint32_t *val)
+{
+	uint8_t *prop = NULL;
+	uint32_t num_prop = 1;
```
etienne-lms, line 61: is this needed?

etienne-lms (commented):
minor comments only....

core/drivers/tpm2/tpm2_chip.c
```diff
@@ -57,5 +232,24 @@ enum tpm2_result tpm2_chip_register(struct tpm2_chip *chip)
 	if (ret)
 		EMSG("TPM2 Self Test Failed");
 
+	ret = tpm2_populate_capability(chip);
```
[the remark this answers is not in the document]
ruchi393: Yes, we continue even if self tests fail

core/kernel/tcg.c
```diff
@@ -0,0 +1,393 @@
...
+ */
+static TEE_Result tcg_parse_specid_event(void *buffer, uint32_t log_size,
+					 uint32_t *pos,
+					 struct tpml_digest_values *digest_list)
+{
+	struct tcg_efi_spec_id_event *spec_event = NULL;
+	struct tcg_pcr_event *event_header = buffer;
+	uint32_t active = 0, supported = 0, alg_count = 0, pcr_count = 0;
```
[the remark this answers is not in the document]
ruchi393: I am ok wither ways...splitting it to multiple defines

core/drivers/tpm2/tpm2_chip.c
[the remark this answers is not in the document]
etienne-lms: For clarify and ease maintenance, if think it should be obvious that failing test does not abort the sequence. Either an inline comment, or using a debug trace level or explicitly not scratching `ret`. Or all of these :)
```
	/* Self test result is informative */
	if (tpm2_selftest(full))
		DMSG("TPM2 Self Test Failed");
```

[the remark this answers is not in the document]
ruchi393: Thanks @etienne-lms . Your suggestions LGTM. I have made the change

ruchi393:
Thanks for the review @etienne-lms . PR updated with all comments addressed.

core/drivers/tpm2/tpm2_chip.c
```diff
@@ -28,13 +31,180 @@ enum tpm2_result tpm2_chip_recv(uint8_t *buf, uint32_t *len,
...
+		return TPM2_ERR_GENERIC;
+
+	caps->pcr_select_min = pcr_select_min;
+
+	return TPM2_OK;
+}
+
+static void tpm2_dump_capability(struct tpm2_chip *chip)
+{
+	struct tpm2_caps *caps = &chip->capability;
```
jenswi-linaro, line 192: __maybe_unused

core/kernel/tcg.c
```diff
@@ -0,0 +1,395 @@
...
+	size_t pos = 0;
+
+	event_size = tcg_event_final_size(digest_list);
+	if (*offset >= log_size || *offset + event_size > log_size) {
+		EMSG("Event exceeds log size");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	event = (struct tcg_pcr_event2 *)((uintptr_t)buffer + *offset);
+	*pcr = get_unaligned_le32(&event->pcr_index);
```
jenswi-linaro, line 242: `struct tcg_pcr_event2` has the `__packed` attribute so you can access `event->pcr_index` directly.

By the way, if `struct tcg_pcr_event2` wouldn't had had the `__packed` attribute then would this `event` pointer be undefined.

core/kernel/tcg.c
```diff
@@ -0,0 +1,395 @@
...
+
+	/* get the count */
+	count = get_unaligned_le32(&event->digests.count);
+	if (count != digest_list->count)
+		return TEE_ERROR_BAD_FORMAT;
+
+	pos = offsetof(struct tcg_pcr_event2, digests);
+	pos += offsetof(struct tpml_digest_values, digests);
+
+	for (i = 0; i < digest_list->count; i++) {
```
jenswi-linaro, line 252: What if `digest_list->count` isn't `TPM2_NUM_PCR_BANKS`, are we still accessing `struct tcg_pcr_event2` is it's defined in C or do we deviate?

core/include/kernel/tcg.h
```diff
@@ -0,0 +1,164 @@
...
+/*
+ * Crypto Agile Log Entry Format
+ *
+ * @pcr_index:	PCRIndex event extended to
+ * @event_type:	Type of event
+ * @digests:	List of digestsextended to PCR index
+ * @event_size: Size of the event data
+ * @event:	Event data
+ */
+struct tcg_pcr_event2 {
```
jenswi-linaro, line 43: What does the `2` in the name signify?

core/include/kernel/tcg.h
```diff
@@ -0,0 +1,164 @@
...
+struct tcg_efi_spec_id_event {
+	uint8_t signature[16];
+	uint32_t platform_class;
+	uint8_t spec_version_minor;
+	uint8_t spec_version_major;
+	uint8_t spec_errata;
+	uint8_t uintn_size;
+	uint32_t number_of_algorithms;
+	struct tcg_efi_spec_id_event_algorithm_size digest_sizes[];
+} __packed;
```
jenswi-linaro, line 94: Why is this packed?

jenswi-linaro (commented):
I find the usage of the `__packed` attribute confusing and also suspect that it may not be needed at all the places. It's not clear to me why it's used at the various places.

core/arch/arm/kernel/boot.c
```diff
@@ -19,6 +19,7 @@
 #include <kernel/misc.h>
 #include <kernel/panic.h>
 #include <kernel/tee_misc.h>
+#include <kernel/tcg.h>
```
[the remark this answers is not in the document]
ruchi393: Not now. I will remove it.

core/kernel/tcg.c
ruchi393: Offset in the buffer at which tcg_pcr_event2 is located is not necessarily aligned to a word boundary.

For clarity, I would like to keep the get_unaligned_le32 for the pcr_index

core/kernel/tcg.c
ruchi393: digest_lists count gets populated when parsing spec ID event,  Check that the count isn't TPM2_NUM_PCR_BANKS should happen there. I have added the check. Thanks for pointing this out.

There will be deviation from struct tcg_pcr_event2 , as actual size of digest (of type tcg_pcr_event2) would depend on the value of count which comes from spec ID event. That is the reason elements event_size and event are not accessed directly.

core/include/kernel/tcg.h
ruchi393: The name comes from the specification. Spec mentions SHA1 Event Log Entry Format as tcg_pcr_event. This was for devices with only single hash bank. This is still used to specify the header in event log.
So event log has
- event log header (of type tcg_pcr_event)
- spec ID event [ type tcg_efi_spec_id_event]
- multiple events of type tcg_pcr_event2 [ count of digest and hash algorithms in this needs to match what is parsed from spec ID event above)
-
For devices which support multiple hash banks, event log format entry was changed and was called tcg_pcr_event2.

core/include/kernel/tcg.h
ruchi393: Frankly, @apalos used it as __packed in u-boot, so I reused it as it is over here, :-). TPM structures are quiet unaligned. Eventlog header has a variable array at it's end and spec ID event is placed after that in event log. So it's starting boundary may be unaligned.

ruchi393:
PR updated with comments addressed. I will take care of the 2 checkpatch reported issues when squashing the patches.

ruchi393:
Commits rebased to master and squashed with all comments addressed

core/drivers/tpm2/tpm2_cmd.c
```diff
@@ -135,15 +135,291 @@ enum tpm2_result tpm2_selftest(uint8_t full)
...
+	 * capability (uin32_t)
+	 * capability data [ This is the property data to be returned ]
+	 */
+	prop_offset = sizeof(struct tpm2_cmd_hdr) + sizeof(uint8_t) +
+		      sizeof(uint32_t);
+
+	if (*prop_len >= resp_len - prop_offset) {
+		memcpy(prop, &resp_buf[prop_offset], resp_len - prop_offset);
+	} else {
+		EMSG("Response Buffer size for property small");
```
jenswi-linaro, line 195: too small?

core/drivers/tpm2/tpm2_cmd.c
```diff
@@ -135,15 +135,291 @@ enum tpm2_result tpm2_selftest(uint8_t full)
...
+	 * corresponding to a specific PCR. Octet index is found by dividing the
+	 * PCR number by 8.
+	 */
+	pcr_select_idx = pcr_idx >> 3;
+
+	/*
+	 * pcr_select_size indicates the number of octets in pcr_select. It's
+	 * minimum value is available in caps structure.
+	 */
+	if (pcr_select_idx > caps.pcr_select_min)
```
jenswi-linaro, line 248: Isn't this `pcr_select_size = MAX(pcr_select_idx, caps.pcr_select_min)`?

core/kernel/tcg.c
```diff
@@ -0,0 +1,399 @@
...
+	uint16_t hash_alg = 0;
+	uint32_t active = 0;
+	uint32_t alg_count = 0;
+	uint32_t i = 0;
+	uint32_t pcr_count = 0;
+	uint32_t spec_active = 0;
+	uint32_t supported = 0;
+	size_t spec_event_size = 0;
+
+	if (*pos >= log_size || (*pos + sizeof(*spec_event)) > log_size)
```
jenswi-linaro, line 129: Wouldn't this be the same as `if ((*pos + sizeof(*spec_event)) > log_size)`?

core/kernel/tcg.c
```diff
@@ -0,0 +1,399 @@
...
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	if (spec_event->spec_version_minor !=
+			TCG_EFI_SPEC_ID_EVENT_SPEC_VERSION_MINOR_TPM2 ||
+	    spec_event->spec_version_major !=
+			TCG_EFI_SPEC_ID_EVENT_SPEC_VERSION_MAJOR_TPM2)
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (spec_event->number_of_algorithms < 1) {
```
jenswi-linaro, line 147: Looks like a complicated way of testing for zero.

core/kernel/tcg.c
jenswi-linaro: So we have a struct definition that can't be used? That deserves at least a comment with a warning. We should also consider if there's a better way of documenting layout of binary data.

core/kernel/tcg.c
ruchi393: Yes..Will change

core/kernel/tcg.c
ruchi393: Changed it

jenswi-linaro:

ruchi393:
Thanks @jenswi-linaro . Patches squashed and A-by tag applied

jforissier:
@etienne-lms I think all your comments have been addressed, would you like to give tags before I merge this?

etienne-lms:
Yes, thanks.
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for the series.

jforissier:
Tags applied and merged, thanks.

PR #5253, opened 2022-04-06
Add support for Versal ACAP

ldts:
Evaluation board [link](https://www.xilinx.com/products/boards-and-kits/vck190.html)
Register reference [link](https://www.xilinx.com/htmldocs/registers/am012/am012-versal-register-reference.html#mod___apu_gic_redist_sgispi.html)

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -274,6 +274,11 @@ R:	Ricardo Salveti <ricardo@foundries.io> [@ricardosalveti]
 S:	Maintained
 F:	core/arch/arm/plat-zynqmp/
 
+Xilinx Versal ACAP
+R:	Jorge Ramirez-Ortiz <jorge@foundries.io> [@ldts]
+S:	Maintained
+F:	core/arch/arm/plat-versal/
+
 Virtualization support
 R:	Volodymyr Babchuk <vlad.babchuk@gmail.com>
 S:	Maintained


core/arch/arm/plat-versal/conf.mk
```diff
@@ -0,0 +1,31 @@
...
+$(call force,CFG_PL011,y)
+$(call force,CFG_GIC,y)
+$(call force,CFG_ARM_GICV3,y)
+$(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
+$(call force,CFG_WITH_ARM_TRUSTED_FW,y)
+
+# Disable core ASLR for two reasons:
+# 1. There is no source for ALSR seed, as ATF does not provide a
+#    DTB to OP-TEE. Hardware RNG is also not currently supported.
+# 2. OP-TEE does not boot with enabled CFG_CORE_ASLR.
```
jforissier, line 15: 1 is fair enough but 2 is slightly worrying, any idea what may be going wrong?

core/arch/arm/plat-versal/conf.mk
```diff
@@ -0,0 +1,31 @@
...
+$(call force,CFG_ARM32_core,y)
+endif
+
+CFG_TZDRAM_START ?= 0x60000000
+CFG_TZDRAM_SIZE  ?= 0x10000000
+CFG_SHMEM_START  ?= 0x70000000
+CFG_SHMEM_SIZE   ?= 0x10000000
+
+CFG_WITH_STATS ?= y
+CFG_CRYPTO_WITH_CE ?= n
```
jforissier, line 31: Interesting. If CE is supported at all, why not enable it by default?

core/arch/arm/plat-versal/platform_config.h
```diff
@@ -0,0 +1,69 @@
...
+#define PLATFORM_CONFIG_H
+
+#include <mm/generic_ram_layout.h>
+
+/* Make stacks aligned to data cache line length */
+#define STACK_ALIGNMENT		64
+
+#ifdef CFG_WITH_PAGER
+#error "Pager not supported for Versal"
+#endif
```
jforissier, line 16: If "not supported" is only due to not having any boot loader config for that purpose, I'd rather avoid that error. It could be useful to accept `CFG_WITH_PAGER=y` if only for build testing.

.azure-pipelines.yml
```diff
@@ -221,6 +221,7 @@ jobs:
           _make PLATFORM=hisilicon-hi3519av100_demo
           _make PLATFORM=amlogic
           _make PLATFORM=rzn1
+          _make PLATFORM=versal CFG_ARM64_core=y CFG_CORE_DYN_SHM=y
```
jforissier, line 224: Please make sure `CFG_ARM64_core=y CFG_CORE_DYN_SHM=y` are set by default and remove them from here.

jforissier (commented):
Hi @ldts,

Comments below. For commit "MAINTAINERS: maintain the Xilinx Versal ACAP platform":

core/arch/arm/plat-versal/conf.mk
ldts: right, so the tool that generates the boot.bin (Xilinx's bootgen) cant find the entry point from the ELF file when this option is enabled. I could provide it at build time to TF-A  and that would get it running....I'll remove the second comment and expand.

jforissier: Alright no need to make it work (although it would be nice of course!), but yes please explain the limitation.

core/arch/arm/plat-versal/conf.mk
ldts: my bad, just reused it from zynqmp. will remove as it doesnt seem to be supported.

jforissier: Then `$(call force,CFG_CRYPTO_WITH_CE,n)` would be appropriate. But that seems weird, 64-bit core without crypto extensions?

ldts: my mistake (again!) - I did enable support and assumed that the crash I was seeing was due to that and didnt think twice.
CEs are not imposed by the architecture so even though optional I agree it would be weird not to have them.

Just checked the TRM and it is indeed supported. So yes, will force it 

Versal ACAP
Technical Reference Manual
AM011 (v1.3) October 27, 2021

```
Application Processing Unit
The APU MPCore processor is integrated into the FPD subsystem. The implementation and
functionality of the Cortex®-A72 processor is detailed in the Application Processing Unit chapter
with additional information in the Arm documents.
The main features of the APU:
• Dual Arm Cortex-A72 cores
• VFPv4 floating point, NEON, Crypto extension
• 48 KB instruction, 32 KB data caches
• GIC-500 interrupt controller
• 1 MB L2 Cache Coherent Interconnect (CCI)
```

jforissier: `CFG_CRYPTO_WITH_CE ?= y` would be slightly more correct (allowing the developer to disable it at build time), but it doesn't really matter.

ldts: ok

core/arch/arm/plat-versal/platform_config.h
ldts: ok will do that but currently it just wont boot (I havent looked into it TBH)

core/arch/arm/plat-versal/platform_config.h
```diff
@@ -0,0 +1,65 @@
...
+#define UART_BAUDRATE		115200
+#endif
+#ifndef CONSOLE_BAUDRATE
+#define CONSOLE_BAUDRATE	UART_BAUDRATE
+#endif
+
+/* For virtual platforms where there isn't a clock */
+#ifndef CONSOLE_UART_CLK_IN_HZ
+#define CONSOLE_UART_CLK_IN_HZ	1
+#endif
```
jforissier, line 63: Looks copied from `core/arch/arm/plat-zynqmp/platform_config.h`. Can it happen? I don't think so.

core/arch/arm/plat-versal/main.c
```diff
@@ -0,0 +1,61 @@
...
+#include <drivers/pl011.h>
+
+#include <arm.h>
+#include <console.h>
+#include <kernel/boot.h>
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
+#include <kernel/tee_time.h>
+#include <mm/core_memprot.h>
+#include <trace.h>
```
jforissier, line 21: Minor: prefer all includes in one block and in alphabetical order

jforissier (commented):
Two more small things, then for the first two commits please apply:

core/arch/arm/plat-versal/main.c
ldts: ah sure, sorry I wasnt sure if this rule had been changed recently. I remember seeing some crypto commit (crypto/aspeed) with all the includes unsorted. will fix.

core/arch/arm/plat-versal/platform_config.h
ldts: yeah, I dont think so either. will remove

ldts: just wondering if it will break qemu? I am developing on the target so not looking on virtual platforms yet.
https://support.xilinx.com/s/article/1166606?language=en_US

PR #5255, opened 2022-04-07
core: fix use after free in tee_ta_open_session()

jenswi-linaro:
Fixes a use after free where the session pointer 's' was used after
tee_ta_close_session() while recovering from an error.

Fixes: 82061b8d7b34 ("core: store TA params in session struct")


jforissier (approved):
Nice catch!

jenswi-linaro:
Yeah, I trapped it while experimenting with the Memory Tagging Extension.

Tag applied.

PR #5259, opened 2022-04-08
mk: update CROSS_COMPILE and CROSS_COMPILE64 defaults

bryanbrattlof:
After ARM's 9.2-2019.12 release of their GNU toolchain in 2019, ARM has
consistently named their toolchains aarch64-none-linux-gnu and
arm-none-linux-gnueabihf including the latest 11.2-2022.02 released
this February. Currently optee still uses the old naming for the
default CROSS_COMPILE and CROSS_COMPILE64 flags, requiring many who use
the newest toolchain versions to override these flags.

Update the default CROSS_COMPILE and CROSS_COMPILE64 flags to reduce
the number of compiler variables needed to compile optee.

Signed-off-by: Bryan Brattlof <bb@ti.com>

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -17,9 +17,9 @@
 # (CFG_* variables only).
 
 # Cross-compiler prefix and suffix
-CROSS_COMPILE ?= arm-linux-gnueabihf-
+CROSS_COMPILE ?= arm-none-linux-gnueabihf-
 CROSS_COMPILE32 ?= $(CROSS_COMPILE)
-CROSS_COMPILE64 ?= aarch64-linux-gnu-
+CROSS_COMPILE64 ?= aarch64-none-linux-gnu-
 COMPILER ?= gcc
 
 # For convenience


jforissier:
HI @bryanbrattlof,

I don't know what to think about this change. Although I can see that this might help a few developers, I have the following arguments against:

- IMO that Arm decided to rename the compiler commands was a pretty bad idea because they basically tried to fix something that was not broken (unless I missed something -- what is the rationale for the name change?)
- I do not think most Linux distributions have followed suit. For example on my Ubuntu 21.10:
  ```
  jerome@jerome-XPS-13-9310:~$ /usr/bin/aarch64-linux-gnu-gcc --version | head -1
  aarch64-linux-gnu-gcc (Ubuntu 11.2.0-5ubuntu1) 11.2.0
  jerome@jerome-XPS-13-9310:~$ dpkg -S /usr/bin/aarch64-linux-gnu-gcc
  gcc-aarch64-linux-gnu: /usr/bin/aarch64-linux-gnu-gcc
  ```
  Why should we favor the toolchains compiled by Arm over the ones included in the distros?
- There is also at least another place that more or less assumes the "old" names, it is `scripts/symbolize.py`. More generally, external projects are also more likely to have the old name as the default than the new one.
- We have put in place a [trivial step in `toolchains.mk`](https://github.com/OP-TEE/build/blob/3.16.0/toolchain.mk#L22) to restore compatibility with the established names when using the toolchains built by Arm and I think it's a much less painful way of dealing with the issue.
- There is no guarantee that Arm won't change their mind again (either to revert to the old naming convention, or to choose yet another one!).

That being said, if you or anyone has a different opinion please speak up and we'll pick the option that makes life easier to the larger number of people.

bryanbrattlof:
Interesting... I may have borked my system at some point in time and started using the `aarch64-none-linux-gnu` binaries and just assumed the world had accepted the name change. Looking around, it seems most of the popular distros (the 3 that came to my mind first) still use the `aarch64-linux-gnu` naming.[0][1][2]

So it seems like this can be dropped until a consensus is reached or ARM gives up and changes the names back. :) Either way, I agree, this would be a pointless changes until then.

Thanks for such a quick review and reply @jforissier !

[0] https://archlinux.org/packages/community/x86_64/aarch64-linux-gnu-gcc/
[1] https://packages.ubuntu.com/jammy/gcc-aarch64-linux-gnu
[2] https://packages.fedoraproject.org/pkgs/cross-binutils/binutils-aarch64-linux-gnu/

PR #5261, opened 2022-04-11
scripts/gen_tee_bin.py: inlude .ctors in included sections

jenswi-linaro:
Includes the section .ctors in sections included in the pager binary.
This fixes a problem where the constructor function pointers are all
zeroes. Constructors are normally not used by OP-TEE, but with
CFG_CORE_SANITIZE_KADDRESS=y it's used to initialize access to global
variables.


jforissier (commented):

jforissier:
Unrelated to this change -- I tried to reproduce the CI error locally (regression_2002.1), but haven't succeeded. It happens occasionally in CI so there must be something wrong...

jenswi-linaro:
> Unrelated to this change -- I tried to reproduce the CI error locally (regression_2002.1), but haven't succeeded. It happens occasionally in CI so there must be something wrong...

Yes, I was a bit surprised. It seems like something we need to sort out.

jforissier:
@jenswi-linaro although this is a fix, it concerns only a debug configuration (`CFG_CORE_SANITIZE_KADDRESS=y`) so I suggest we should take it after the release. What do you think?

PR #5262, opened 2022-04-12
drivers: caam: fix function definition when CFG_CAAM_NO_ITR=y

clementfaure:
There is a bug in the CAAM JR interruption enablement logic. When
CFG_CAAM_NO_ITR=y, the JR interruptions are used and when
CFG_CAAM_NO_ITR=n, the JR interruptions are not used.

Even with this wrong logic, the CAAM is still able to enqueue jobs.
When no JR interruptions are received, the CAAM will manually dequeue
jobs from the jobring by checking the number of jobs done in the output
ring slots full register.

CAAM JR interruptions are not mandatory for the CAAM to work properly
but it makes the dequeuing faster than polling the output ring slot full
register.

Fixes: 3f45afc31 ("drivers: caam: disable the use of interrupts for some platforms")

### DIFF
--- a/core/arch/arm/plat-imx/crypto_conf.mk
+++ b/core/arch/arm/plat-imx/crypto_conf.mk
@@ -51,13 +51,16 @@ $(call force,CFG_CAAM_SIZE_ALIGN,1)
 $(call force,CFG_JR_BLOCK_SIZE,0x1000)
 $(call force,CFG_JR_INDEX,2)
 $(call force,CFG_JR_INT,114)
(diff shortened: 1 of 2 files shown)


jforissier:
> Yeah, common pitfall ... I took your suggestion instead

Indeed, I like this way better. One thing though, changing the name now might break external build scripts... you know better than me if it is likely to be a problem or not, so from my point of view:

clementfaure:
It will be fine!
Thanks for the quick review @jforissier

PR #5266, opened 2022-04-13
Versal ACAP TRNG driver

ldts:
The Versal™ device contains a true random number generator (TRNG). 

The TRNG enables applications to be compliant with AIS-20/31 and NIST-800-90A/B/C standards. 

For additional details, see the Versal ACAP Security Manual (UG1508). This manual requires an active NDA to download from the [Design Security Lounge](https://www.xilinx.com/member/design_security.html).

To support these standards, the TRNG operates in three modes as shown in the following figure.

download

Features include:

Generates cryptographically secure random numbers
Generates data blocks with a 32-bit wide interface
Provides security strength of 256 bits
Ring oscillator and PLL random sources

### DIFF
--- a/core/drivers/versal_trng.c
+++ b/core/drivers/versal_trng.c
@@ -1,0 +1,1158 @@
+// SPDX-License-Identifier: MIT
+/*
+ * Copyright (C) 2022 Xilinx, Inc.  All rights reserved.
+ * Copyright (C) 2022 Foundries Ltd.
+ *
+ * Driver port from Xilinx's FSBL implementation, jorge@foundries.io
+ *
+ * The Xilinx True Random Number Generator(TRNG) module in Versal - PMC TRNG
+ * consists of an entropy source, a deterministic random bit generator (DRBG)
+ * and health test logic, which tests the randomness of the generated data.
+ * The entropy source for the unit is an array of Ring Oscillators.
+ *
+ * The Versal PMC TRNG is envisaged to operate in three basic modes: DRNG, PTRNG
+ * and HRNG . Each of these can be operated with or without Derivative Function
+ * (DF), resulting in a total of 6 different modes of operation.
+ *
+ * NIST SP-800-90A practically requires the true random generators based on
+ * CTR_DRBG to include a derivation function (DF). This is expected to be
+ * implemented inside the Silicon (TRNG IP). However, the version of the IP used
+ * in Versal PMC does not have this capability. Hence, a software
+ * implementation of the DF is done in this driver.
+ *
+ * DRNG mode: Deterministic Random Number Generator mode.
+ *            In this mode, the DRBG portion of the TRNG is used. User provides
+ *            the (external) seed.
+ * PTRNG mode: Physical True Random Number Generator mode (aka Entropy mode).
+ *            In this mode digitized Entropy source is output as random number.
+ * HRNG mode: Hybrid Random Number Generator mode.
+ *            This is combination of above two modes in which the Entropy source
+ *            is used to provide the seed, which is fed to the DRBG, which in
+ *            turn generates the random number.
+ *
+ * DRNG mode with DF: It may not be common usecase to use the DF with DRNG as
+ * the general expectation would be that the seed would have sufficient entropy.
+ * However, the below guideline from section 10.2.1 of NIST SP-800-90A implies
+ * that need for DF for DRNG mode too: "..the DRBG mechanism is specified to
+ * allow an implementation tradeoff with respect to the use of this derivation
+ * function. The use of the derivation function is optional if either an
+ * approved RBG or an entropy source provides full entropy output when entropy
+ * input is requested by the DRBG mechanism. Otherwise, the derivation function
+ * shall be used". Sufficient large entropy data from user is fed to DF to
+ * generate the seed which will be loaded into the external seed registers.
+ * From here, it is similar to regular DRNG mode.
+ *
+ * PTRNG mode with DF: This mode is similar to PTRNG mode, however, the entropy
+ * data from the core output registers are accumulated and fed to the DF
+ * (instead of directly consuming it). The output of the DF would be final
+ * random data. In this mode, the output of DF is not seed but the random data.
+ *
+ * HRNG mode with DF: This mode is the combination of the above two modes.
+ * The entropy data is fed to the DF to produce seed. This seed is loaded to the
+ * external seed registers which provide seed to the DRBG.
+ */
+#include <arm.h>
+#include <crypto/crypto.h>
+#include <initcall.h>
+#include <io.h>
+#include <kernel/delay.h>
+#include <kernel/panic.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <rng_support.h>
+#include <stdlib.h>
+#include <string.h>
+#include <tee/tee_cryp_utl.h>
+#include <trace.h>
+
+#define TRNG_STATUS			0x04
+#define TRNG_STATUS_QCNT_SHIFT		9
+#define TRNG_STATUS_QCNT_MASK		(BIT(9) | BIT(10) | BIT(11))
+#define TRNG_STATUS_CERTF_MASK		BIT(3)
+#define TRNG_STATUS_DTF_MASK		BIT(1)
+#define TRNG_STATUS_DONE_MASK		BIT(0)
+#define TRNG_CTRL			0x08
+#define TRNG_CTRL_EUMODE_MASK		BIT(8)
+#define TRNG_CTRL_PRNGMODE_MASK		BIT(7)
+#define TRNG_CTRL_PRNGSTART_MASK	BIT(5)
+#define TRNG_CTRL_PRNGXS_MASK		BIT(3)
+#define TRNG_CTRL_TRSSEN_MASK		BIT(2)
+#define TRNG_CTRL_PRNGSRST_MASK		BIT(0)
+#define TRNG_EXT_SEED_0			0x40
+/* Below registers are not directly referenced in driver but are accessed
+ * with offset from TRNG_EXT_SEED_0
+ * Register: TRNG_EXT_SEED_1		0x00000044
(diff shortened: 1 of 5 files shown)


ldts:
I don't really have a solution for the checks - not warnings, not errors - raised by checkpatch. perhaps we would just discard them?

jforissier:
> I don't really have a solution for the checks - not warnings, not errors - raised by checkpatch. perhaps we would just discard them?

We are not too strict on checkpatch but here I think there are easy fixes for most (all?) of the issues. For instance using a statement expression with local variables `({ ... })`.

ldts:
> > I don't really have a solution for the checks - not warnings, not errors - raised by checkpatch. perhaps we would just discard them?
> 
> We are not too strict on checkpatch but here I think there are easy fixes for most (all?) of the issues. ~For instance using a statement expression with local variables `({ ... })`~. Edit: some macros definitely nee to use their args multiple times so indeed we cannot clear all warnings. Please just do your best ;)

ok I'll have a look see if there is something that sticks out

ldts:
in the end I just got rid of the macros....better to have checkpatch clean as a reference so the file doesn't degrade

ldts:
@jforissier just a small reminder

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+{
+	while (max_blk > 0) {
+		xorb(iv, in);
+		encrypt(iv, iv);
+		in += BLK_SIZE;
+		max_blk -= 1;
+	}
+}
+
+static void setup_key(const unsigned char *k, int klen)
```
jforissier, line 316: Make `klen` and `i` a `size_t` then the cast in `memcpy()` in not needed.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+{
+	unsigned char rcon = 1;
+	int i, sch_size = 240;
```
jforissier, line 319: `size_t` and one declaration per line

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+#define TRNG_BYTES_PER_REG	4
+#define TRNG_MAX_QCNT		4
+#define TRNG_RESEED_TIMEOUT	15000
+#define TRNG_GENERATE_TIMEOUT	8000
+#define TRNG_MIN_DFLENMULT	2
+#define TRNG_MAX_DFLENMULT	9
+#define PRNGMODE_RESEED		0
+#define PRNGMODE_GEN		TRNG_CTRL_PRNGMODE_MASK
+#define RESET_DELAY		10
+#define TRNG_SWAP_ENDIAN	TEE_U32_BSWAP
```
jforissier, line 131: Strictly speaking, shouldn't this be `TEE_U32_TO_BIG_ENDIAN`? (which you should probably use directly, no need for `TRNG_SWAP_ENDIAN`)

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+		trng->dfin.ivc[0] = TRNG_SWAP_ENDIAN(index / BLK_SIZE);
+		checksum((unsigned char *)&trng->dfin,
+			 trng->dfout + index, dfin_len / BLK_SIZE);
+	}
+
+	/* DF algorithm - step 2 */
+	setup_key(trng->dfout, DF_KEY_LEN);
+	for (index = 0; index < TRNG_SEED_LEN; index += BLK_SIZE) {
+		inp_blk = index == 0 ? &dfout[TRNG_SEC_STRENGTH_LEN] :
+				       &dfout[index - BLK_SIZE];
```
jforissier, line 459: Same here, `if / else`

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+static TEE_Result trng_wait_for_event(vaddr_t addr, uint32_t off,
+				      uint32_t mask, uint32_t event,
+				      uint32_t time_out)
+{
+	uint64_t tref = timeout_init_us(time_out);
+
+	do {
+		if (timeout_elapsed(tref))
+			panic();
+
```
jforissier, line 513: Remove empty line

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+	/* Loop as many times based on len requested. In each burst 128 bits
+	 * are generated, which is reflected in QCNT value of 4 by hardware.
+	 */
+	for (bcnt = 0; bcnt < bursts; bcnt++) {
+		if (trng_wait_for_event(trng->cfg.addr,
+					TRNG_STATUS, TRNG_STATUS_QCNT_MASK,
+					TRNG_MAX_QCNT << TRNG_STATUS_QCNT_SHIFT,
+					TRNG_GENERATE_TIMEOUT)) {
+			EMSG("Time out");
```
jforissier, line 582: The `EMSG()` output doesn't show file/line number by default so a bit of context would be nice. How about "versal_trng: Time out"? and same in other places. You may define a macro to add the prefix.

jforissier (commented):
Hello @ldts,

A few comments below. Thanks.

ldts:
thanks @jforissier I'll update the PR over the weekend,

@jforissier One thing, I need a processor to DMA to OP-TEE memory (this is before Linux or U-boot are running). how would I allocate this memory? I would have thought the DMA could access the OP-TEE stack data but is not what I see (this is on Versal, I have a few more PR on the way)

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+	uint32_t buf[RAND_BUF_LEN];   /* buffer of random bits      */
+	size_t len;
+	struct trng_dfin dfin;
+	uint8_t dfout[TRNG_SEED_LEN]; /* output of the DF operation */
+};
+
+/* Derivative function variables */
+static unsigned char SBx1[256];
+static unsigned char SBx2[256];
+static unsigned char SBx3[256];
```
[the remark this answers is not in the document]
ldts: yes in this case I thought it made more sense to keep the original variables...I'd prefer to keep it this way if possible.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+		max_blk -= 1;
+	}
+}
+
+static void setup_key(const unsigned char *k, int klen)
+{
+	unsigned char rcon = 1;
+	int i, sch_size = 240;
+
+	rounds = 14;
```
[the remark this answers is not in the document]
ldts: I suppose it must be ( the algorithm is pretty much verbatim from the original FSBL code). But I see no reason why we could not just initialize it to 14 although I am not sure it matters much?
yes, will use MAX_ROUNDS

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+
+static TEE_Result trng_wait_for_event(vaddr_t addr, uint32_t off,
+				      uint32_t mask, uint32_t event,
+				      uint32_t time_out)
+{
+	uint64_t tref = timeout_init_us(time_out);
+
+	do {
+		if (timeout_elapsed(tref))
+			panic();
```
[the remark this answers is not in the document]
ldts: great. thanks!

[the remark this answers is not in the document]
ldts: will also add a comment for that

core/drivers/versal_trng.c
ldts: Um is this new? AFAIR it always showed the file and line

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+			EMSG("Time out");
+			return TEE_ERROR_GENERIC;
+		}
+
+		/* DTF flag set during generate indicates catastrophic
+		 * condition, which needs to be checked for every time unless we
+		 * are in PTRNG mode
+		 */
+		if (trng->usr_cfg.mode == TRNG_PTRNG)
+			goto skip_dtf;
```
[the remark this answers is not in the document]
ldts: Can I use the 'maintainability card' on this?...I think reading the code is easier with this helper.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+{
+	trng_clrset32(trng->cfg.addr, TRNG_CTRL,
+		      TRNG_CTRL_PRNGSRST_MASK, TRNG_CTRL_PRNGSRST_MASK);
+	trng_write32(trng->cfg.addr, TRNG_RESET, TRNG_RESET_VAL_MASK);
+	udelay(RESET_DELAY);
+}
+
+static TEE_Result trng_check_seed(uint8_t *entropy, uint32_t len)
+{
+	uint32_t *p = (void *)entropy;
```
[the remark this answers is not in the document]
ldts: not sure about this. why does it matter if the memory can be referenced to?

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1142 @@
...
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result trng_collect_random(struct versal_trng *trng, uint8_t *dst,
+				      size_t len)
+{
+	const size_t bursts = len / TRNG_BURST_SIZE;
+	const size_t words = TRNG_BURST_SIZE_BITS / TRNG_REG_SIZE;
+	uint32_t *p = (void *)dst;
```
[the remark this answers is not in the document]
ldts: same as above.

core/drivers/versal_trng.c
jforissier: About five years :wink: https://github.com/OP-TEE/optee_os/blob/3.17.0/mk/config.mk#L99-L104
I introduced this mainly for "INFO" messages, and thought "ERROR" level output should probably not include source-code level information either because the end-user should be able to understand the error without checking the code. That being said, we're loosing the function name which often gives some context. Maybe we need a macro to show the driver name like in the Linux kernel?

PR #5267, opened 2022-04-13
drivers: caam: add JR interruption only if CFG_CAAM_ITR=y

clementfaure:
Adding the JR interruption in the OPTEE CAAM driver, even if not used in
OPTEE, prevents the Linux CAAM driver from using the JR interruption on
platforms sharing the same line of interruption for all job rings.

To dequeue job from the job ring, the Linux CAAM driver would pull the
number of jobs done from the output ring slot full register.

The fix is to add the JR interruption only if CFG_CAAM_ITR=y. This
allows the Linux CAAM driver from dequeuing job faster than polling from
a register.


jforissier:
s/interruption/interrupt/ I think

Would you like this merged for the release?

clementfaure:
Comment addressed!
Yes, that would be great to have it merged for the release

PR #5274, opened 2022-04-19
Enable mbedtls pk parse and write

LuminousXLB:
This enables PK parse and write functions in TEE Kernel, when ASN1 parse and write are enabled.

Signed-off-by: Shen Jiamin <shen_jiamin@comp.nus.edu.sg>


jenswi-linaro:
Those functions are not supposed to be called directly within OP-TEE, "everything" should go via the <crypto/crypto.h> API. The reason for this is that the crypto library can be replaced by another crypto library.

PR #5279, opened 2022-04-20
Multiple small fixes for issues in the CAAM driver and Core

clementfaure:


jenswi-linaro:
Regarding "drivers: caam: avoid arithmetic operation for pointer assignment", either style is OK. Which is easier to read depends. Anyway, for all commits please apply:

clementfaure:
Hello @jenswi-linaro 
Thanks for the review! Tag applied + rebase

PR #5285, opened 2022-04-26
arm: vfp_a32: check the existence of FPU NEON by CPACR

Neal-liu:
Almost all Cortex-A processors come with a Floating-Point Unit (FPU)
and most also have a NEON unit. However, use of registers D16-D31 of
the VFP register file depends on the processor's FPU implementation.

Arm coprocessor access control register (CPACR.D32DIS) indicates if
FPU and Advanced SIMD is implemented. Check this bit to see if the
processor is supported or not.

Signed-off-by: Neal Liu <neal_liu@aspeedtech.com>

-----------------
Fix the issue of undefined instruction abort.
```
E/LD:  Call stack:
E/LD:   0x40112710
  regression_1010.4 OK
o regression_1010.5 Invalid memory access 5
E/TC:0 0
E/TC:0 0 Core undef-abort at address 0x88002fd4
E/TC:0 0  fsr 0x00000000  ttbr0 0x8806786a  ttbr1 0x8806006a  cidr 0x2
E/TC:0 0  cpu #0          cpsr 0x400001df
E/TC:0 0  r0 0x880416c8      r4 0x00000000    r8 0x001510c0   r12 0xfffd4e40
E/TC:0 0  r1 0x00000001      r5 0x88041640    r9 0x00112730    sp 0x880684a8
E/TC:0 0  r2 0x00000000      r6 0x00000000   r10 0x0000000a    lr 0x88002fb7
E/TC:0 0  r3 0x00000000      r7 0x00112740   r11 0x00235708    pc 0x88002fd4
E/TC:0 0 TEE load address @ 0x88000000
E/TC:0 0 Call stack:
E/TC:0 0  0x88002fd4
E/TC:0 0  0x88002fb7
E/TC:0 0 Panic '[abort] abort in abort handler (trap CPU)' at core/arch/arm/kernel/abort.c:468 <get_fault_type>
```

### DIFF
--- a/core/arch/arm/kernel/vfp_a32.S
+++ b/core/arch/arm/kernel/vfp_a32.S
@@ -3,6 +3,8 @@
  * Copyright (c) 2015-2016, Linaro Limited
  */
 
+#include <arm32.h>
+#include <arm32_macros.S>
 #include <asm.S>
 
 	.fpu	neon
@@ -10,6 +12,9 @@
 /* void vfp_save_extension_regs(uint64_t regs[VFP_NUM_REGS]); */
 FUNC vfp_save_extension_regs , :
 	vstm	r0!, {d0-d15}
+	read_cpacr r1
+	tst	r1, #CPACR_D32DIS
+	bxne	lr
 	vstm	r0, {d16-d31}
 	bx	lr
 END_FUNC vfp_save_extension_regs
@@ -17,6 +22,9 @@ END_FUNC vfp_save_extension_regs
 /* void vfp_restore_extension_regs(uint64_t regs[VFP_NUM_REGS]); */
 FUNC vfp_restore_extension_regs , :
 	vldm	r0!, {d0-d15}
+	read_cpacr r1
+	tst	r1, #CPACR_D32DIS
+	bxne	lr
 	vldm	r0, {d16-d31}
 	bx	lr
 END_FUNC vfp_restore_extension_regs


core/arch/arm/kernel/vfp_a32.S
```diff
@@ -4,20 +4,30 @@
...
+#include <arm32.h>
+#include <generated/arm32_sysreg.S>
 
 	.fpu	neon
 
 /* void vfp_save_extension_regs(uint64_t regs[VFP_NUM_REGS]); */
 FUNC vfp_save_extension_regs , :
-	vstm	r0!, {d0-d15}
-	vstm	r0, {d16-d31}
+	mov	r4, r0
```
jenswi-linaro, line 14: The calling convention requires `r4` to be preserved.
Why not keep `r0` as indexing register and use `r1` when reading CPACR?
Same comment applies below for `vfp_restore_extension_regs()`.

Neal-liu: According to ARM calling convention, `r0-r3` are used for function arguments/return values. `r4-r11` are used for local variables.
That is why I choose `r4` instead of `r1` to prevent function parameter extension.

jenswi-linaro: Not completely correct. Some quotes from section 5.1.1 Core registers below that table:

*The first four registers r0-r3 (a1-a4) are used to pass argument values into a subroutine and to return a result value from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).*

*A subroutine must preserve the contents of the registers r4-r8, r10, r11 and SP (and r9 in PCS variants that designate r9 as v6).*

So you can do whatever you like with r0-r3, an eventual return value is ignored for this function as it's supposed to return `void`.

Neal-liu: Got your point. I'll revise it. Thanks for your suggestion.

core/arch/arm/kernel/vfp_a32.S
```diff
@@ -4,20 +4,30 @@
...
+#include <arm32.h>
+#include <generated/arm32_sysreg.S>
 
 	.fpu	neon
 
 /* void vfp_save_extension_regs(uint64_t regs[VFP_NUM_REGS]); */
 FUNC vfp_save_extension_regs , :
-	vstm	r0!, {d0-d15}
-	vstm	r0, {d16-d31}
+	mov	r1, r0
```
jenswi-linaro, line 14: We could save one instruction by using `r0` directly.
`read_cpacr` can operate with `r1` instead.

jenswi-linaro: `read_cpacr()` is just an assembly macro, the `reg` argument is the chosen destination register.
```
        .macro read_cpacr reg
        mrc p15, 0, \reg, c1, c0, 2
        .endm
```

Neal-liu: I check the wrong file (arm32_sysreg.h) which is inline assembly.
```
/* Coprocessor Access Control Register */
static inline __noprof uint32_t read_cpacr(void)
{
        uint32_t v;

        asm volatile ("mrc p15, 0, %0, c1, c0, 2" : "=r"  (v));

        return v;
}
```

I'll simplify the patch as you suggested. Thanks !

jenswi-linaro:

core/arch/arm/kernel/vfp_a32.S
```diff
@@ -4,19 +4,27 @@
  */
 
 #include <asm.S>
+#include <arm32.h>
+#include <generated/arm32_sysreg.S>
```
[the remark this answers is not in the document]
Neal-liu: okay, I'll update it as you suggested.

Neal-liu:

tag applied, thanks for your review.

etienne-lms (commented):

Neal-liu:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`

tag applied, thanks for your review.

PR #5297, opened 2022-04-29
core:crypto: Add X25519 support

sohhass:
Hey folks!

This PR adds X25519 support for OP-TEE as defined in GlobalPlatform TEE Internal Core API v1.2 using libtomcrypt as the backend crypto implementation.

Since the Curve25519 implementation does not support bignums, there is also a new key attribute type (octet reference) as defined in GlobalPlatform TEE Internal Core API v1.2 Table 6-15.

### DIFF
--- a/core/crypto.mk
+++ b/core/crypto.mk
@@ -47,6 +47,10 @@ CFG_CRYPTO_ECC ?= y
 CFG_CRYPTO_SM2_PKE ?= y
 CFG_CRYPTO_SM2_DSA ?= y
 CFG_CRYPTO_SM2_KEP ?= y
+# X25519 is only supported by libtomcrypt
+ifeq ($(CFG_CRYPTOLIB_NAME),tomcrypt)
+CFG_CRYPTO_X25519 ?= y
+endif
 
 # Authenticated encryption
 CFG_CRYPTO_CCM ?= y
@@ -163,6 +167,7 @@ core-ltc-vars += SIZE_OPTIMIZATION
 core-ltc-vars += SM2_PKE
 core-ltc-vars += SM2_DSA
 core-ltc-vars += SM2_KEP
+core-ltc-vars += X25519
 # Assigned selected CFG_CRYPTO_xxx as _CFG_CORE_LTC_xxx
 $(foreach v, $(core-ltc-vars), $(eval _CFG_CORE_LTC_$(v) := $(CFG_CRYPTO_$(v))))
 _CFG_CORE_LTC_MPI := $(CFG_CORE_MBEDTLS_MPI)
@@ -186,6 +191,7 @@ _CFG_CORE_LTC_SHA512_DESC := $(CFG_CRYPTO_DSA)
 _CFG_CORE_LTC_XTS := $(CFG_CRYPTO_XTS)
 _CFG_CORE_LTC_CCM := $(CFG_CRYPTO_CCM)
 _CFG_CORE_LTC_AES_DESC := $(call cfg-one-enabled, CFG_CRYPTO_XTS CFG_CRYPTO_CCM)
+$(call force,CFG_CRYPTO_X25519,n,not supported by mbedtls)
 endif
 
 ###############################################################
@@ -226,6 +232,7 @@ _CFG_CORE_LTC_HASH := $(call ltc-one-enabled, MD5 SHA1 SHA224 SHA256 SHA384 \
 _CFG_CORE_LTC_MAC := $(call ltc-one-enabled, HMAC CMAC CBC_MAC)
 _CFG_CORE_LTC_CBC := $(call ltc-one-enabled, CBC CBC_MAC)
 _CFG_CORE_LTC_ASN1 := $(call ltc-one-enabled, RSA DSA ECC)
+_CFG_CORE_LTC_EC25519 := $(call ltc-one-enabled, X25519)
 
 ###############################################################
 # Platform independent crypto-driver configuration
--- a/core/lib/libtomcrypt/src/pk/x25519/x25519_make_key.c
+++ b/core/lib/libtomcrypt/src/pk/x25519/x25519_make_key.c
@@ -27,7 +27,6 @@ int x25519_make_key(prng_state *prng, int wprng, curve25519_key *key)
 {
    int err;
 
-   LTC_ARGCHK(prng != NULL);
    LTC_ARGCHK(key  != NULL);
 
    if ((err = prng_is_valid(wprng)) != CRYPT_OK) {
--- a/core/lib/libtomcrypt/src/sub.mk
+++ b/core/lib/libtomcrypt/src/sub.mk
@@ -6,3 +6,4 @@ subdirs-$(_CFG_CORE_LTC_ACIPHER) += math
 subdirs-y += misc
 subdirs-y += modes
 subdirs-$(_CFG_CORE_LTC_ACIPHER) += pk
+subdirs-$(_CFG_CORE_LTC_EC25519) += pk
--- a/core/lib/libtomcrypt/sub.mk
+++ b/core/lib/libtomcrypt/sub.mk
@@ -101,6 +101,8 @@ ifneq (,$(filter y,$(_CFG_CORE_LTC_SM2_DSA) $(_CFG_CORE_LTC_SM2_PKE)))
    cppflags-lib-y += -DLTC_ECC_SM2
 endif
 
+cppflags-lib-$(_CFG_CORE_LTC_X25519) += -DLTC_CURVE25519
+
 cppflags-lib-y += -DLTC_NO_PKCS
 
 cppflags-lib-y += -DLTC_DER
@@ -133,7 +135,7 @@ endif
 srcs-$(_CFG_CORE_LTC_SM2_DSA) += sm2-dsa.c
 srcs-$(_CFG_CORE_LTC_SM2_PKE) += sm2-pke.c
 srcs-$(_CFG_CORE_LTC_SM2_KEP) += sm2-kep.c
-
+srcs-$(_CFG_CORE_LTC_X25519) += x25519.c
 ifeq ($(_CFG_CORE_LTC_ACIPHER),y)
 srcs-y += mpi_desc.c
 endif
--- a/core/lib/libtomcrypt/x25519.c
+++ b/core/lib/libtomcrypt/x25519.c
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2022, Technology Innovation Institute (TII)
+ */
+
+#include <crypto/crypto.h>
+#include <stdlib.h>
+#include <string.h>
+#include <tee_api_types.h>
+#include <trace.h>
+#include <utee_defines.h>
+
+#include "acipher_helpers.h"
+
+/* X25519 key is an octet string of 32 bytes */
+#define X25519_KEY_SIZE_BYTES UL(32)
+
+TEE_Result crypto_acipher_alloc_x25519_keypair(struct x25519_keypair *key,
+					       size_t key_size)
+{
+	size_t key_size_bytes = key_size / 8;
+
+	if (!key)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	memset(key, 0, sizeof(*key));
+
+	if (key_size_bytes != X25519_KEY_SIZE_BYTES)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	key->priv = calloc(1, key_size_bytes);
+	key->pub = calloc(1, key_size_bytes);
+
+	if (!key->priv || !key->pub) {
+		free(key->priv);
+		free(key->pub);
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result crypto_acipher_gen_x25519_key(struct x25519_keypair *key,
+					 size_t key_size)
+{
+	curve25519_key ltc_tmp_key = { };
+	size_t key_size_bytes = key_size / 8;
+
+	if (key_size_bytes != X25519_KEY_SIZE_BYTES)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (x25519_make_key(NULL, find_prng("prng_crypto"), &ltc_tmp_key) !=
+	    CRYPT_OK)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (key_size_bytes < sizeof(ltc_tmp_key.pub) ||
+	    key_size_bytes < sizeof(ltc_tmp_key.priv))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	memcpy(key->pub, ltc_tmp_key.pub, sizeof(ltc_tmp_key.pub));
+	memcpy(key->priv, ltc_tmp_key.priv, sizeof(ltc_tmp_key.priv));
+	memzero_explicit(&ltc_tmp_key, sizeof(ltc_tmp_key));
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result crypto_acipher_x25519_shared_secret(struct x25519_keypair
+					       *private_key,
+					       void *public_key,
+					       void *secret,
+					       unsigned long *secret_len)
+{
+	curve25519_key ltc_private_key = {
+		.type = PK_PRIVATE,
+		.algo = PKA_X25519,
+	};
+	curve25519_key ltc_public_key = {
+		.type = PK_PUBLIC,
+		.algo = PKA_X25519,
+	};
+
+	if (!private_key || !public_key || !secret || !secret_len)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	static_assert(sizeof(ltc_public_key.pub) == X25519_KEY_SIZE_BYTES &&
+		      sizeof(ltc_public_key.priv) == X25519_KEY_SIZE_BYTES);
+
+	memcpy(ltc_public_key.pub, public_key, X25519_KEY_SIZE_BYTES);
+	memcpy(ltc_private_key.priv, private_key->priv, X25519_KEY_SIZE_BYTES);
+
+	if (x25519_shared_secret(&ltc_private_key, &ltc_public_key,
+				 secret, secret_len) != CRYPT_OK)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	/* Clear private key from the stack */
+	memzero_explicit(&ltc_private_key, sizeof(ltc_private_key));
+
+	/*
+	 * RFC 7748, sec 6.1, check for all zero shared secret output to reject
+	 * input points of low order.
+	 */
+	if (*secret_len != X25519_KEY_SIZE_BYTES ||
+	    !consttime_memcmp(secret, ltc_private_key.pub,
+			       X25519_KEY_SIZE_BYTES))
+		return TEE_ERROR_SECURITY;
+
+	return TEE_SUCCESS;
+}
--- a/core/tee/tee_svc_cryp.c
+++ b/core/tee/tee_svc_cryp.c
@@ -2,6 +2,7 @@
 /*
  * Copyright (c) 2014, STMicroelectronics International N.V.
  * Copyright (c) 2020, 2022 Linaro Limited
+ * Copyright (c) 2022, Technology Innovation Institute (TII)
  */
 
 #include <assert.h>
@@ -79,6 +80,11 @@ struct tee_cryp_obj_secret {
 #define ATTR_OPS_INDEX_BIGNUM     1
     /* Convert to/from value attribute depending on direction */
 #define ATTR_OPS_INDEX_VALUE      2
+    /* Convert to/from curve25519 attribute depending on direction */
+#define ATTR_OPS_INDEX_25519      3
+
+    /* Curve25519 key bytes size is always 32 bytes*/
+#define KEY_SIZE_BYTES_25519 UL(32)
 
 struct tee_cryp_obj_type_attrs {
 	uint32_t attr_id;
@@ -423,6 +429,23 @@ static const struct tee_cryp_obj_type_attrs tee_cryp_obj_sm2_keypair_attrs[] = {
 	},
 };
 
+static
+const struct tee_cryp_obj_type_attrs tee_cryp_obj_x25519_keypair_attrs[] = {
+	{
+	.attr_id = TEE_ATTR_X25519_PRIVATE_VALUE,
+	.flags = TEE_TYPE_ATTR_REQUIRED,
+	.ops_index = ATTR_OPS_INDEX_25519,
+	RAW_DATA(struct x25519_keypair, priv)
+	},
+
+	{
+	.attr_id = TEE_ATTR_X25519_PUBLIC_VALUE,
+	.flags = TEE_TYPE_ATTR_REQUIRED,
+	.ops_index = ATTR_OPS_INDEX_25519,
+	RAW_DATA(struct x25519_keypair, pub)
+	},
+};
+
 struct tee_cryp_obj_type_props {
 	TEE_ObjectType obj_type;
 	uint16_t min_size;	/* may not be smaller than this */
@@ -551,6 +574,10 @@ static const struct tee_cryp_obj_type_props tee_cryp_obj_props[] = {
 	PROP(TEE_TYPE_SM2_KEP_KEYPAIR, 1, 256, 256,
 	     sizeof(struct ecc_keypair),
 	     tee_cryp_obj_sm2_keypair_attrs),
+
+	PROP(TEE_TYPE_X25519_KEYPAIR, 1, 256, 256,
+	     sizeof(struct x25519_keypair),
+	     tee_cryp_obj_x25519_keypair_attrs),
 };
 
 struct attr_ops {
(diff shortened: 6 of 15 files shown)


core/tee/tee_svc_cryp.c
```diff
@@ -78,6 +79,11 @@ struct tee_cryp_obj_secret {
 #define ATTR_OPS_INDEX_BIGNUM     1
     /* Convert to/from value attribute depending on direction */
 #define ATTR_OPS_INDEX_VALUE      2
+    /* Convert to/from curve25519 attribute depending on direction */
+#define ATTR_OPS_INDEX_25519      3
```
jforissier, line 83: Why is this needed? All other algorithms use bignum and value.

jforissier:
For commit "libutee: add X25519 support": with one minor comment:

core/tee/tee_svc_cryp.c
sohhass: > Why is this needed? All other algorithms use bignum and value.

The X(Ed)25519 curve operations arguments are  32-byte octet strings as defined in
[RFC 7748, pg 7](https://datatracker.ietf.org/doc/html/rfc7748#page-7). Also from TEE Internal Core API Specification v1.2, Table Table 6-15, `TEE_ATTR_X25519_*` is a binary octet string (while other asym crypto attrs are bignum). The underlying implementation of libtomcrypt for Curve25519 operations does not use bignums (AFAIK most of the library implementations follow this convention). The X25519 byte level encoding is `little-endian` according to the `RFC-7748`, and if I am not mistaken for bignums it is big-endian. If we go the bignum route, we should then add conversions from bignum to appropriate binary format before calling the `X25519` ops.

jforissier: Makes sense, thanks for the details. Let's keep this new define.

core/tee/tee_svc_cryp.c
```diff
@@ -871,6 +898,121 @@ static void op_attr_value_clear(void *attr)
...
+	if (s < key_size || !buffer)
+		return TEE_ERROR_SHORT_BUFFER;
+
+	return copy_to_user(buffer, *key, key_size);
+}
+
+static TEE_Result op_attr_25519_to_binary(void *attr, void *data,
+					  size_t data_len, size_t *offs)
+{
+	TEE_Result res;
```
etienne-lms, line 942:
```c
	size_t next_offs = 0;

File: core/tee/tee_svc_cryp.c [hunk 3]
```diff
@@ -871,6 +898,121 @@ static void op_attr_value_clear(void *attr)
...
+	(*offs) = next_offs;
+
+	return TEE_SUCCESS;
+}
+
+static bool op_attr_25519_from_binary(void *attr, const void *data,
+				      size_t data_len, size_t *offs)
+{
+	uint8_t **key = attr;
+	uint32_t s;
```
etienne-lms, line 965: = 0;

core/tee/tee_svc_cryp.c
```diff
@@ -871,6 +898,121 @@ static void op_attr_value_clear(void *attr)
...
+}
+
+static void op_attr_25519_clear(void *attr)
+{
+	uint8_t **key = attr;
+	size_t _size = KEY_SIZE_BYTES_25519;
+
+	assert(*key);
+
+	memzero_explicit(*key, _size);
```
etienne-lms, line 1013: `memzero_explicit(*key, KEY_SIZE_BYTES_25519);` is fine (IMHO)

etienne-lms, line 1013: `memzero_explicit(*key, KEY_SIZE_BYTES_25519);` is fine (IMHO)

core/tee/tee_svc_cryp.c
```diff
@@ -3226,6 +3420,41 @@ TEE_Result syscall_cryp_derive_key(unsigned long state,
 		crypto_acipher_free_ecc_public_key(&peer_key);
 		crypto_acipher_free_ecc_public_key(&peer_eph_key);
 	}
+#endif
+#if defined(CFG_CRYPTO_X25519)
+	else if (cs->algo == TEE_ALG_X25519) {
+		uint8_t *x25519_pub_key;
+		uint8_t *pt_secret;
+		unsigned long pt_secret_len;
```
etienne-lms, line 3428:
```c
		uint8_t *x25519_pub_key = NULL;
		uint8_t *pt_secret = NULL;
		unsigned long pt_secret_len = 0;

Comment (etienne-lms) [inline core/tee/tee_svc_cryp.c:3428]:
```c
		uint8_t *x25519_pub_key = NULL;
		uint8_t *pt_secret = NULL;
		unsigned long pt_secret_len = 0;

File: core/lib/libtomcrypt/src/sub.mk [hunk 6]
```diff
@@ -6,3 +6,7 @@ subdirs-$(_CFG_CORE_LTC_ACIPHER) += math
 subdirs-y += misc
 subdirs-y += modes
 subdirs-$(_CFG_CORE_LTC_ACIPHER) += pk
+
+ifeq ($(_CFG_CORE_LTC_ACIPHER),n)
```
etienne-lms, line 10: prefer: `ifneq ($(_CFG_CORE_LTC_ACIPHER),y)`

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+
+	memcpy(ltc_public_key.pub, public_key,
+		    X25519_KEY_SIZE_BYTES);
+
+	memcpy(ltc_private_key.priv, private_key->priv,
+		    X25519_KEY_SIZE_BYTES);
+
+	if (x25519_shared_secret(&ltc_private_key, &ltc_public_key,
+		    secret, secret_len) != CRYPT_OK)
+				return TEE_ERROR_BAD_PARAMETERS;
```
etienne-lms, line 104:
```c
	if (x25519_shared_secret(&ltc_private_key, &ltc_public_key,
				 secret, secret_len) != CRYPT_OK)

File: core/lib/libtomcrypt/x25519.c [hunk 8]
```diff
@@ -0,0 +1,109 @@
...
+	if (sizeof(ltc_public_key.pub) < X25519_KEY_SIZE_BYTES
+		|| sizeof(ltc_private_key.priv) < X25519_KEY_SIZE_BYTES)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+
+	memcpy(ltc_public_key.pub, public_key,
+		    X25519_KEY_SIZE_BYTES);
+
+	memcpy(ltc_private_key.priv, private_key->priv,
+		    X25519_KEY_SIZE_BYTES);
```
etienne-lms, line 100: each of the 2 fits in a single line
```c
	memcpy(ltc_public_key.pub, public_key, X25519_KEY_SIZE_BYTES);
	memcpy(ltc_private_key.priv, private_key->priv, X25519_KEY_SIZE_BYTES);

File: core/lib/libtomcrypt/x25519.c [hunk 9]
```diff
@@ -0,0 +1,109 @@
...
+
+	if (x25519_shared_secret(&ltc_private_key, &ltc_public_key,
+		    secret, secret_len) != CRYPT_OK)
+				return TEE_ERROR_BAD_PARAMETERS;
+
+	memset(&ltc_private_key, 0, sizeof(ltc_private_key));
+	memset(&ltc_public_key, 0, sizeof(ltc_public_key));
+	return TEE_SUCCESS;
```
etienne-lms, line 108: add an empty line above `return` instruction.

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+	};
+	curve25519_key ltc_public_key = {
+		.type = PK_PUBLIC,
+		.algo = PKA_X25519
+	};
+
+	if (private_key == NULL
+		|| public_key == NULL
+		|| secret == NULL
+		|| secret_len == NULL)
```
etienne-lms, line 87:
```c
	if (!private_key || !public_key || !secret || !secret_len)

	COMPILE_TIME_ASSERT(sizeof(ltc_public_key.pub) ==
			    X25519_KEY_SIZE_BYTES &&
			    sizeof(ltc_public_key.priv) ==
			    X25519_KEY_SIZE_BYTES);
```

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+TEE_Result crypto_acipher_x25519_shared_secret(struct x25519_keypair
+					       *private_key,
+					       void *public_key,
+					       void *secret,
+					       unsigned long *secret_len)
+{
+
+	curve25519_key ltc_private_key = {
+		.type = PK_PRIVATE,
+		.algo = PKA_X25519
```
etienne-lms, line 77: prefer with a terminal coma for each line (ditto line 81)

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+	return TEE_SUCCESS;
+}
+
+TEE_Result crypto_acipher_x25519_shared_secret(struct x25519_keypair
+					       *private_key,
+					       void *public_key,
+					       void *secret,
+					       unsigned long *secret_len)
+{
+
```
etienne-lms, line 74: remove empty line

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+	memset(&ltc_tmp_key, 0, sizeof(ltc_tmp_key));
+
+	if (x25519_make_key(&dummy, find_prng("prng_crypto"),
+	    &ltc_tmp_key)!= CRYPT_OK)
+			return TEE_ERROR_BAD_PARAMETERS;
+
+	if (key_size_bytes < sizeof(ltc_tmp_key.pub)
+		|| key_size_bytes < sizeof(ltc_tmp_key.priv))
+		return TEE_ERROR_BAD_PARAMETERS;
```
etienne-lms, line 59: if (key_size_bytes < sizeof(ltc_tmp_key.pub) ||
	    key_size_bytes < sizeof(ltc_tmp_key.priv))

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+			return TEE_ERROR_OUT_OF_MEMORY;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result crypto_acipher_gen_x25519_key(struct x25519_keypair *key,
+					 size_t key_size)
+{
+	curve25519_key ltc_tmp_key = { };
+	size_t key_size_bytes = (size_t)(key_size / 8);
```
etienne-lms, line 45: remove cast

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+	prng_state dummy;
```
etienne-lms, line 46: `= { };`

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+
+	if (key_size_bytes != X25519_KEY_SIZE_BYTES)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	memset(&ltc_tmp_key, 0, sizeof(ltc_tmp_key));
+
+	if (x25519_make_key(&dummy, find_prng("prng_crypto"),
+	    &ltc_tmp_key)!= CRYPT_OK)
```
etienne-lms, line 54:
```c
	if (x25519_make_key(&dummy, find_prng("prng_crypto"), &ltc_tmp_key) !=
	    CRYPT_OK)

File: core/lib/libtomcrypt/x25519.c [hunk 17]
```diff
@@ -0,0 +1,109 @@
...
+			return TEE_ERROR_BAD_PARAMETERS;
+
+	if (key_size_bytes < sizeof(ltc_tmp_key.pub)
+		|| key_size_bytes < sizeof(ltc_tmp_key.priv))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	memcpy(key->pub, ltc_tmp_key.pub, sizeof(ltc_tmp_key.pub));
+	memcpy(key->priv, ltc_tmp_key.priv, sizeof(ltc_tmp_key.priv));
+	memset(&ltc_tmp_key, 0, sizeof(ltc_tmp_key));
```
etienne-lms, line 63: candidate for `memzero_explicit()`?

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+/* X25519 key is an octet string of 32 bytes */
+#define X25519_KEY_SIZE_BYTES UL(32)
+
+TEE_Result crypto_acipher_alloc_x25519_keypair(struct x25519_keypair *key,
+					       size_t key_size)
+{
+
+	size_t key_size_bytes = (size_t)(key_size / 8);
+
+	if (key == NULL)
```
etienne-lms, line 24: prefer `if (!key)` for consistency in optee_os

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,109 @@
...
+
+	memset(key, 0, sizeof(*key));
+
+	if (key_size_bytes != X25519_KEY_SIZE_BYTES)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	key->priv = calloc(1, key_size_bytes);
+	key->pub = calloc(1, key_size_bytes);
+
+	if (key->priv == NULL || key->pub == NULL)
```
etienne-lms, line 35:
```c
	if (!key->priv || !key->pub) {
		free(key->priv);
		free(key->pub);
		return TEE_ERROR_OUT_OF_MEMORY;
	}

File: core/lib/libtomcrypt/x25519.c [hunk 20]
```diff
@@ -0,0 +1,109 @@
...
+#include "acipher_helpers.h"
+
+/* X25519 key is an octet string of 32 bytes */
+#define X25519_KEY_SIZE_BYTES UL(32)
+
+TEE_Result crypto_acipher_alloc_x25519_keypair(struct x25519_keypair *key,
+					       size_t key_size)
+{
+
+	size_t key_size_bytes = (size_t)(key_size / 8);
```
etienne-lms, line 22: cast not needed

core/tee/tee_svc_cryp.c
sohhass: `op_attr_25519_to_binary` should we initialize all `*_to_binary` (bignum, value) functions? was here for just consistency

core/tee/tee_svc_cryp.c
sohhass: `op_attr_25519_from_binary`:  ditto as `op_attr_25519_to_binary` comment

core/lib/libtomcrypt/x25519.c
sohhass: good catch, will fix!

core/tee/tee_svc_cryp.c
```diff
@@ -871,6 +898,121 @@ static void op_attr_value_clear(void *attr)
...
+	uint64_t key_size = (uint64_t)KEY_SIZE_BYTES_25519;
+
+	res = copy_from_user(&s, size, sizeof(s));
+	if (res != TEE_SUCCESS)
+		return res;
+
+	res = copy_to_user(size, &key_size, sizeof(key_size));
+	if (res != TEE_SUCCESS)
+		return res;
+	if (!key_size)
```
etienne-lms, line 930: is it possible that `key_size == 0` here? Did you mean `if (!s)`?

core/tee/tee_svc_cryp.c
etienne-lms: We should fix other `*_to_binary()` functions, but that may be the subject for another P-R. As here you introduce new code, prefer have it following optee_os coding style from the start.

core/lib/libtomcrypt/x25519.c
sohhass: @etienne-lms Thanks for the review! One comment on this one, the struct `prng_state` here is passed since the libtomcrypt X25519 implementation expects a valid object due to this check `LTC_ARGCHK(prng != NULL)` (https://github.com/OP-TEE/optee_os/blob/master/core/lib/libtomcrypt/src/pk/x25519/x25519_make_key.c#L30), in reality it is never used as OP-TEE has its own `prng_crypto_desc` ("prng_crypto") and all the `prng`functions declare `prng_state` as `__unused`. 

Also I am not sure if it is safe to initialize `prng_state` this way since the initialization needs to properly handle the mutex. 

Making this change produces this error during key generation:
```
D/LD:  ldelf:168 ELF (cb3e5ba0-adf1-11e0-998b-0002a5d5c51b) at 0x40014000
F/TC:? 0 trace_syscall:151 syscall #8 (syscall_check_access_rights)
F/TC:? 0 trace_syscall:151 syscall #27 (syscall_cryp_obj_alloc)
F/TC:? 0 trace_syscall:151 syscall #54 (syscall_obj_generate_key)
E/TC:1   Dead canary at end of 'stack_abt[3]' (0xe1a01fc)
E/TC:1   Panic at core/kernel/thread.c:124 <thread_check_canaries>
E/TC:1   TEE load address @ 0xe100000
E/TC:1   Call stack:
E/TC:1    0x0e108900
E/TC:1    0x0e113ec8
E/TC:1    0x0e116e58
E/TC:1    0x0e106a5c
```

One option is to remove this check `LTC_ARGCHK(prng != NULL)` from libtomcrypt implementation, but I avoid making any changes to the library.
Also note, this check is not present in other asymmetric crypto key generation functions, so this case is only valid for `X25519`.

core/tee/tee_svc_cryp.c
sohhass: thanks for pointing out , this check `if (!key_size)` is not needed, will remove!

core/lib/libtomcrypt/x25519.c
jenswi-linaro: I believe removing the `LTC_ARGCHK(prng != NULL)` check is the best option. I don't know how large `prng_state` is, but I'm afraid it can become quite large in certain configurations.

sohhass: Thanks for the comment, `LTC_ARGCHK(prng != NULL)` removed

sohhass:
@etienne-lms thanks for the your review, all your comments are addressed in the latest changes and rebased on top of #5301

jforissier:
> rebased on top of #5301

We still need to figure how to sort these conflicting IDs, 5301 is not the right way it seems.

sohhass:
> > rebased on top of #5301
> 
> We still need to figure out how to address the issue of conflicting IDs, 5301 is not the right way it seems.

Currently OP-TEE does not defineAPI Version headers that would allow compile time selection of specific version. This is defined in TEE Internal Core API v1.2 section 3.1.1

lib/libutee/include/tee_api_defines.h
```diff
@@ -289,7 +294,7 @@
 #define TEE_ECC_CURVE_NIST_P384             0x00000004
 #define TEE_ECC_CURVE_NIST_P521             0x00000005
 #define TEE_ECC_CURVE_SM2                   0x00000300
-
+#define TEE_ECC_CURVE_25519                 0x00000300
```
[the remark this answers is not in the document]
jforissier: Move up one line to keep numerical order

core/crypto.mk
```diff
@@ -226,6 +228,7 @@ _CFG_CORE_LTC_HASH := $(call ltc-one-enabled, MD5 SHA1 SHA224 SHA256 SHA384 \
 _CFG_CORE_LTC_MAC := $(call ltc-one-enabled, HMAC CMAC CBC_MAC)
 _CFG_CORE_LTC_CBC := $(call ltc-one-enabled, CBC CBC_MAC)
 _CFG_CORE_LTC_ASN1 := $(call ltc-one-enabled, RSA DSA ECC)
+_CFG_CORE_LTC_EC25519 := $(call ltc-one-enabled, X25519 ED25519)
```
jforissier, line 231: Do we really want `ED25519` here? It doesn't seem that `CFG_CRYPTO_ED25519` (and therefore `_CFG_CORE_LTC_ED25519`) is supported in this PR?

core/lib/libtomcrypt/acipher_helpers.h
```diff
@@ -35,7 +35,7 @@ static inline TEE_Result convert_ltc_verify_status(int ltc_res, int ltc_stat)
 	}
 }
 
-#ifdef CFG_CRYPTOLIB_NAME_tomcrypt
+#ifdef _CFG_CORE_LTC_ECC
```
jforissier, line 38: This is not strictly related to 25519 support, please make a separate patch for this fix.

core/lib/libtomcrypt/src/sub.mk
```diff
@@ -6,3 +6,7 @@ subdirs-$(_CFG_CORE_LTC_ACIPHER) += math
 subdirs-y += misc
 subdirs-y += modes
 subdirs-$(_CFG_CORE_LTC_ACIPHER) += pk
+
+ifneq ($(_CFG_CORE_LTC_ACIPHER),y)
+subdirs-$(_CFG_CORE_LTC_EC25519) += pk
+endif
```
jforissier, line 12: One single statement like this would be better IMO:
```make
ifneq (,$(filter y,$(_CFG_CORE_LTC_ACIPHER) $(_CFG_CORE_LTC_EC25519)))
subdirs-y += pk
endif
```

Slightly off-topic but perhaps we should consider allowing multiple occurrences of the same sub-directory? I mean the following change:
```diff
diff --git a/mk/subdir.mk b/mk/subdir.mk
index ffa61d4e..d280c309 100644
--- a/mk/subdir.mk
+++ b/mk/subdir.mk
@@ -21,6 +21,8 @@ srcs :=
 gen-srcs :=
 asm-defines-files :=
 
+uniq = $(if $1,$(firstword $1) $(call uniq,$(filter-out $(firstword $1),$1)))
+
 define process-subdir-srcs-y
 ifeq ($$(sub-dir),.)
 srcs                           += $1
@@ -174,8 +176,8 @@ this-out-dir :=
 asm-defines-y :=
 
 # Process subdirectories in current directory
-$$(foreach sd, $$(sub-subdirs), $$(eval $$(call process-subdir,$$(sd))))
+$$(foreach sd, $$(call uniq,$$(sub-subdirs)), $$(eval $$(call process-subdir,$$(sd))))
 endef #process-subdir
 
 # Top subdirectories
-$(foreach sd, $(subdirs), $(eval $(call process-subdir,$(sd))))
+$(foreach sd, $(call uniq,$(subdirs)), $(eval $(call process-subdir,$(sd))))
``` 
...would allow:
```make
subdirs-$(_CFG_CORE_LTC_ACIPHER) += pk
subdirs-$(_CFG_CORE_LTC_EC25519) += pk
```
Opinions welcome (I can create a separate PR)

core/lib/libtomcrypt/sub.mk
```diff
@@ -100,6 +100,9 @@ endif
 ifneq (,$(filter y,$(_CFG_CORE_LTC_SM2_DSA) $(_CFG_CORE_LTC_SM2_PKE)))
    cppflags-lib-y += -DLTC_ECC_SM2
 endif
+ifeq ($(_CFG_CORE_LTC_X25519),y)
+   cppflags-lib-y += -DLTC_CURVE25519
+endif
```
jforissier, line 105: Conditional not needed.
```
cppflags-lib-$(_CFG_CORE_LTC_X25519) += -DLTC_CURVE25519
```

core/crypto.mk
sohhass: not really, this was just a placeholder; `ED25519` will be a different PR, will remove from here

core/lib/libtomcrypt/src/sub.mk
jenswi-linaro: I like the idea @jforissier.

core/crypto.mk
```diff
@@ -47,6 +47,7 @@ CFG_CRYPTO_ECC ?= y
 CFG_CRYPTO_SM2_PKE ?= y
 CFG_CRYPTO_SM2_DSA ?= y
 CFG_CRYPTO_SM2_KEP ?= y
+CFG_CRYPTO_X25519 ?= y
```
jforissier, line 50: Add `ifeq ($(CFG_CRYPTOLIB_NAME),tomcrypt)` / `endif` and also `$(call force,CFG_CRYPTO_X25519,n,not supported by mbedtls)` in `core/core.mk` (same as for `CFG_CRYPTO_RSASSA_NA1` which is also supported by LTC only).

jforissier (commented):
Some more comments

sohhass:
> Some more comments

thanks, addressed!

core/lib/libtomcrypt/x25519.c
```diff
@@ -0,0 +1,106 @@
...
+			    X25519_KEY_SIZE_BYTES);
+
+	memcpy(ltc_public_key.pub, public_key, X25519_KEY_SIZE_BYTES);
+	memcpy(ltc_private_key.priv, private_key->priv, X25519_KEY_SIZE_BYTES);
+
+	if (x25519_shared_secret(&ltc_private_key, &ltc_public_key,
+				 secret, secret_len) != CRYPT_OK)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	memzero_explicit(&ltc_private_key, sizeof(ltc_private_key));
```
jforissier, line 99: I still don't understand this, `crypto_acipher_dh_shared_secret()` doesn't clear the private key for instance.

core/tee/tee_svc_cryp.c
```diff
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
  * Copyright (c) 2014, STMicroelectronics International N.V.
- * Copyright (c) 2020, 2022 Linaro Limited
+ * Copyright (c) 2020, Linaro Limited
```
etienne-lms, line 4: leave the Linaro copyright unchanged.

core/tee/tee_svc_cryp.c
```diff
@@ -872,6 +899,116 @@ static void op_attr_value_clear(void *attr)
...
+		return res;
+
+	if (ADD_OVERFLOW(*offs, key_size, &next_offs))
+		return TEE_ERROR_OVERFLOW;
+
+	if (data && next_offs <= data_len)
+		memcpy((uint8_t *)data + *offs, *key, key_size);
+	*offs = next_offs;
+
+	return res;
```
etienne-lms, line 957: return  TEE_SUCCESS;

sohhass:
> The overall looks good to me. Some comments.

Thanks! comments addressed

core/lib/libtomcrypt/x25519.c
jforissier: @sohhass can you please explain why this is needed?

jforissier: @sohhass can you please explain why this is needed?

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with some minor comments.

core/tee/tee_svc_cryp.c
```diff
@@ -872,6 +899,112 @@ static void op_attr_value_clear(void *attr)
...
+
+	assert(*key);
+
+	memzero_explicit(*key, KEY_SIZE_BYTES_25519);
+}
+
+static void op_attr_25519_free(void *attr)
+{
+	op_attr_25519_clear(attr);
+	free(*(uint8_t**)attr);
```
etienne-lms, line 1005: checkpatch complains for a missing space char: `free(*(uint8_t **)attr);`

For consistency, maybe:
```c
	uint8_t **key = attr;

	op_attr_25519_clear(attr);
	free(*key);
```

core/tee/tee_svc_cryp.c
```diff
@@ -1916,6 +2066,35 @@ static TEE_Result tee_svc_obj_generate_key_ecc(
...
+static TEE_Result
+tee_svc_obj_generate_key_x25519(struct tee_obj *o,
+				const struct tee_cryp_obj_type_props
+							*type_props,
+				uint32_t key_size,
+				const TEE_Attribute *params,
+				uint32_t param_count)
+{
+	TEE_Result res;
+	struct x25519_keypair *tee_x25519_key;
```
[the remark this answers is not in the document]
etienne-lms: to fix

etienne-lms (changes requested):
There is still a coding style issue (reported by checkpatch)  and a missing local variable initialization. Once addressed, i'm fine with this P-R.

There is another issue CI complains on (when building optee_rust_examples_ext) but it is not related to this P-R (see https://github.com/OP-TEE/optee_os/issues/5361).

core/lib/libtomcrypt/src/pk/x25519/x25519_make_key.c
```diff
@@ -27,7 +27,6 @@ int x25519_make_key(prng_state *prng, int wprng, curve25519_key *key)
 {
    int err;
 
-   LTC_ARGCHK(prng != NULL);
```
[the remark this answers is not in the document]
jforissier: @etienne-lms previous discussion: https://github.com/OP-TEE/optee_os/pull/5297#discussion_r864622745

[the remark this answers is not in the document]
jforissier: > @etienne-lms previous discussion: [#5297 (comment)](https://github.com/OP-TEE/optee_os/pull/5297#discussion_r864622745)

@sohhass I believe the reason why the check is removed needs to be better documented. Please create a separate commit. It will also minimize the risk of this line being inadvertently dropped if a conflict occurs when rebasing to a more recent version of LibTomCrypt.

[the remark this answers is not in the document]
etienne-lms: Right, sorry, i didn't look enough in the P-R history.

etienne-lms (commented):

jforissier:
@sohhass I have merged this PR manually because I wanted to re-order the commits in a sequence that makes more sense in my opinion: the commit which allows to enable `CFG_CRYPTO_X25519=y` needs to come last so that everything is ready at that point. Thanks again for your contribution!

sohhass:
> Thanks again for your contribution!

my pleasure!

PR #5299, opened 2022-05-01
Add AM64x and AM62x device support

glneo:
Nothing extra is needed yet from the OP-TEE side to run on these devices (we have been using the `am65x` flavor for these SoCs so far), but we are starting to add drivers that will have different configurations on these two devices. Add named support here.


jforissier:
LGTM but the subject of the commit is slightly misleading, like if it was adding code to support the platforms. Could you split in two commits: "ci: azure: build for k3-am62x and k3-am64x" and "MAINTAINERS: add TI AM62x and AM64x"? With that:

PR #5300, opened 2022-05-01
Unified hardware random number get bytes

glneo:
When in the process of adding a new RNG driver, I noticed a couple oddities we have in the driver-side HW RNG API.

The normal way to implement a HW RNG driver is to override `hw_get_random_byte()`. The issue here is that all the supported RNG hardware outputs more than one byte at a time, leading to the need for buffering of bytes to be read out by this interface in the drivers. Reading out one byte at a time has a bit of a performance overhead, so instead several drivers also override `crypto_rng_read()` (which can read out multiple bytes in one call). Since all users of HW RNG read out in multiple bytes in a loop anyway, there is no need for this single byte API.

Switch all users of RNG over to only `crypto_rng_read()`. Have all drivers implement only `hw_get_random_bytes()`. When HW RNG is enabled `crypto_rng_read()` simply calls `hw_get_random_bytes()`.

DRA7x RNG for instance, the performance was improved while simplifying the driver logic in the process.

The process of converting over most drivers was straightforward but some sanity checks from owners would be nice.

Performance and correctness can be checked by enabling the HW RNG PTA on your plat:

```
CFG_HWRNG_QUALITY ?= 1024
CFG_HWRNG_PTA ?= y
```

In Linux check that `/dev/hwrng` is using the optee_rng driver:

`# cat /sys/class/misc/hw_random/rng_current`

Then run:

`# rngtest -c 100 < /dev/hwrng`

### DIFF
--- a/core/arch/arm/plat-rcar/hw_rng.c
+++ b/core/arch/arm/plat-rcar/hw_rng.c
@@ -22,27 +22,34 @@ static unsigned int spin_lock __nex_data = SPINLOCK_UNLOCK;
  */
 static uint8_t rng_cache[PLAT_RND_VECTOR_SZ] __nex_bss
 					__aligned(RCAR_CACHE_LINE_SZ);
-static uint8_t rng_cache_pos __nex_data = PLAT_RND_VECTOR_SZ;
+static uint8_t rng_cache_pos __nex_data;
 
-uint8_t hw_get_random_byte(void)
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
 {
-	uint32_t exceptions = cpu_spin_lock_xsave(&spin_lock);
+	uint32_t exceptions;
+	uint8_t *buffer = buf;
+	size_t buffer_pos = 0;
 	uint8_t ret_val = 0;
 
-	assert(rng_cache_pos <= PLAT_RND_VECTOR_SZ);
+	assert(rng_cache_pos < PLAT_RND_VECTOR_SZ);
 
-	if (rng_cache_pos == PLAT_RND_VECTOR_SZ) {
-		uint32_t ret = plat_rom_getrndvector(rng_cache, scratch_buf,
-						     sizeof(scratch_buf));
+	while (buffer_pos < len) {
+		exceptions = cpu_spin_lock_xsave(&spin_lock);
+		/* Refill our FIFO */
+		if (rng_cache_pos == 0) {
+			uint32_t ret = plat_rom_getrndvector(rng_cache,
+							scratch_buf,
+							sizeof(scratch_buf));
+			if (ret != 0)
+				panic("ROM_GetRndVector() returned error!");
+		}
 
-		if (ret != 0)
-			panic("ROM_GetRndVector() returned error!");
+		buffer[buffer_pos++] = rng_cache[rng_cache_pos++];
+		if (rng_cache_pos == PLAT_RND_VECTOR_SZ)
+			rng_cache_pos = 0;
 
-		rng_cache_pos = 0;
+		cpu_spin_unlock_xrestore(&spin_lock, exceptions);
 	}
-
-	ret_val = rng_cache[rng_cache_pos++];
-	cpu_spin_unlock_xrestore(&spin_lock, exceptions);
 
 	return ret_val;
 }
--- a/core/arch/arm/plat-rcar/main.c
+++ b/core/arch/arm/plat-rcar/main.c
@@ -27,6 +27,7 @@
  */
 
 #include <console.h>
+#include <crypto/crypto.h>
 #include <kernel/boot.h>
 #include <kernel/panic.h>
 #include <mm/core_memprot.h>
@@ -34,7 +35,6 @@
 #include <stdint.h>
 #include <drivers/scif.h>
 #include <drivers/gic.h>
-#include <rng_support.h>
 
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, SCIF_REG_SIZE);
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
@@ -73,14 +73,16 @@ void console_init(void)
 }
 
 #ifdef CFG_RCAR_ROMAPI
+/* Should only seed from a hardware random number generator */
+static_assert(!IS_ENABLED(CFG_WITH_SOFTWARE_PRNG));
+
 unsigned long plat_get_aslr_seed(void)
 {
 	unsigned long seed = 0;
-	size_t i;
 
-	/* On RCAR we can call hw_get_random_byte() on early boot stages */
-	for (i = 0; i < sizeof(seed); i++)
-		seed = (seed << 8) | hw_get_random_byte();
+	/* On RCAR we can get hw random bytes on early boot stages */
+	if (crypto_rng_read(&seed, sizeof(seed)))
+		panic();
 
 	return seed;
 }
--- a/core/arch/arm/plat-stm/rng_support.c
+++ b/core/arch/arm/plat-stm/rng_support.c
@@ -46,7 +46,7 @@ static inline int hwrng_waithost_fifo_full(void)
 	return 0;
 }
 
-uint8_t hw_get_random_byte(void)
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
 {
 	/*
 	 * Only the HW RNG IP is used to generate the value through the
@@ -85,41 +85,47 @@ uint8_t hw_get_random_byte(void)
 #define _LOCAL_FIFO_SIZE 8     /* min 2, 4, 6, max 8 */
 
 	static uint8_t lfifo[_LOCAL_FIFO_SIZE];     /* local fifo */
-	static int pos = -1;
+	static int pos;
 
 	static int nbcall;  /* debug purpose - 0 is the initial value*/
 
 	volatile uint32_t tmpval[_LOCAL_FIFO_SIZE/2];
-	uint8_t value;
 	int i;
+
+	uint8_t *buffer = buf;
+	size_t buffer_pos = 0;
 
 	nbcall++;
 
-	/* Retrieve data from local fifo */
-	if (pos >= 0) {
-		pos++;
-		value = lfifo[pos];
-		if (pos == (_LOCAL_FIFO_SIZE - 1))
-			pos = -1;
-		return value;
+	while (buffer_pos < len) {
+		/* Refill our FIFO */
+		if (pos == 0) {
+			if (hwrng_waithost_fifo_full())
+				return TEE_ERROR_GENERIC;
+
+			/*
+			 * Read the FIFO according to the number of
+			 * expected elements
+			 */
+			for (i = 0; i < _LOCAL_FIFO_SIZE / 2; i++)
+				tmpval[i] = io_read32(rng_base() +
+						      RNG_VAL_OFFSET) & 0xFFFF;
+
+			/* Update the local SW fifo for next request */
+			pos = 0;
+			for (i = 0; i < _LOCAL_FIFO_SIZE / 2; i++) {
+				lfifo[pos] = tmpval[i] & 0xFF;
+				pos++;
+				lfifo[pos] = (tmpval[i] >> 8) & 0xFF;
+				pos++;
+			}
+			pos = 0;
+		}
+
+		buffer[buffer_pos++] = lfifo[pos++];
+		if (pos == _LOCAL_FIFO_SIZE)
+			pos = 0;
 	}
 
-	if (hwrng_waithost_fifo_full())
-		return 0;
-
-	/* Read the FIFO according the number of expected element */
-	for (i = 0; i < _LOCAL_FIFO_SIZE / 2; i++)
-		tmpval[i] = io_read32(rng_base() + RNG_VAL_OFFSET) & 0xFFFF;
-
-	/* Update the local SW fifo for next request */
-	pos = 0;
-	for (i = 0; i < _LOCAL_FIFO_SIZE / 2; i++) {
-		lfifo[pos] = tmpval[i] & 0xFF;
-		pos++;
-		lfifo[pos] = (tmpval[i] >> 8) & 0xFF;
-		pos++;
-	};
-
-	pos = 0;
-	return lfifo[pos];
+	return TEE_SUCCESS;
 }
--- a/core/crypto/rng_hw.c
+++ b/core/crypto/rng_hw.c
@@ -8,29 +8,6 @@
 #include <tee/tee_cryp_utl.h>
 #include <types_ext.h>
 
-/*
- * This is only here to keep the compiler happy while we convert over
- * platforms. Either hw_get_random_bytes() is overridden or this
- * function is, in either case this is never called.
- */
-uint8_t __weak hw_get_random_byte(void)
-{
-	panic();
-	return 4; // chosen by fair dice roll.
-		  // guaranteed to be random.
-}
-
-TEE_Result __weak hw_get_random_bytes(void *buf, size_t blen)
-{
-	uint8_t *b = buf;
-	size_t n = 0;
-
-	for (n = 0; n < blen; n++)
-		b[n] = hw_get_random_byte();
-
-	return TEE_SUCCESS;
-}
-
 /* This is a HW RNG, no need for seeding */
 TEE_Result __weak crypto_rng_init(const void *data __unused,
 				  size_t dlen __unused)
--- a/core/crypto/rng_hw.c
+++ b/core/crypto/rng_hw.c
@@ -7,12 +7,25 @@
 #include <tee/tee_cryp_utl.h>
 #include <types_ext.h>
 
+TEE_Result __weak hw_get_random_bytes(void *buf, size_t blen)
+{
+	uint8_t *b = buf;
+	size_t n = 0;
+
+	for (n = 0; n < blen; n++)
+		b[n] = hw_get_random_byte();
+
+	return TEE_SUCCESS;
+}
+
+/* This is a HW RNG, no need for seeding */
 TEE_Result __weak crypto_rng_init(const void *data __unused,
 				  size_t dlen __unused)
 {
 	return TEE_SUCCESS;
 }
 
+/* This is a HW RNG, no need to add entropy */
 void __weak crypto_rng_add_event(enum crypto_rng_src sid __unused,
 				 unsigned int *pnum __unused,
 				 const void *data __unused,
@@ -22,15 +35,8 @@ void __weak crypto_rng_add_event(enum crypto_rng_src sid __unused,
 
 TEE_Result __weak crypto_rng_read(void *buf, size_t blen)
 {
-	uint8_t *b = buf;
-	size_t n;
-
-	if (!b)
+	if (!buf)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	for (n = 0; n < blen; n++)
-		b[n] = hw_get_random_byte();
-
-	return TEE_SUCCESS;
+	return hw_get_random_bytes(buf, blen);
 }
-
--- a/core/crypto/rng_hw.c
+++ b/core/crypto/rng_hw.c
@@ -3,9 +3,22 @@
 
 #include <compiler.h>
 #include <crypto/crypto.h>
+#include <kernel/panic.h>
 #include <rng_support.h>
 #include <tee/tee_cryp_utl.h>
 #include <types_ext.h>
+
+/*
+ * This is only here to keep the compiler happy while we convert over
+ * platforms. Either hw_get_random_bytes() is overridden or this
+ * function is, in either case this is never called.
+ */
+uint8_t __weak hw_get_random_byte(void)
+{
+	panic();
+	return 4; // chosen by fair dice roll.
+		  // guaranteed to be random.
+}
 
 TEE_Result __weak hw_get_random_bytes(void *buf, size_t blen)
 {
--- a/core/crypto/rng_hw.c
+++ b/core/crypto/rng_hw.c
@@ -9,21 +9,20 @@
 #include <types_ext.h>
 
 /* This is a HW RNG, no need for seeding */
-TEE_Result __weak crypto_rng_init(const void *data __unused,
-				  size_t dlen __unused)
+TEE_Result crypto_rng_init(const void *data __unused, size_t dlen __unused)
 {
 	return TEE_SUCCESS;
 }
 
 /* This is a HW RNG, no need to add entropy */
-void __weak crypto_rng_add_event(enum crypto_rng_src sid __unused,
-				 unsigned int *pnum __unused,
-				 const void *data __unused,
-				 size_t dlen __unused)
+void crypto_rng_add_event(enum crypto_rng_src sid __unused,
+			  unsigned int *pnum __unused,
+			  const void *data __unused,
+			  size_t dlen __unused)
 {
 }
 
-TEE_Result __weak crypto_rng_read(void *buf, size_t blen)
+TEE_Result crypto_rng_read(void *buf, size_t blen)
 {
 	if (!buf)
 		return TEE_ERROR_BAD_PARAMETERS;
--- a/core/drivers/atmel_trng.c
+++ b/core/drivers/atmel_trng.c
@@ -26,13 +26,7 @@
 #define TRNG_ODATA		0x50
 
 static unsigned int trng_lock = SPINLOCK_UNLOCK;
-static unsigned int trng_random_val_lock = SPINLOCK_UNLOCK;
 static vaddr_t trng_base;
-static uint8_t random_byte_pos;
-static union {
-	uint32_t val;
-	uint8_t byte[sizeof(uint32_t)];
-} random_data;
 
 static uint32_t atmel_trng_read32(void)
 {
@@ -51,13 +45,12 @@ static uint32_t atmel_trng_read32(void)
 	return value;
 }
 
-TEE_Result crypto_rng_read(void *buf, size_t len)
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
 {
 	uint8_t *rngbuf = buf;
 	uint32_t val = 0;
 	size_t len_to_copy = 0;
 
-	assert(buf);
 	assert(trng_base);
 
 	while (len) {
@@ -69,33 +62,6 @@ TEE_Result crypto_rng_read(void *buf, size_t len)
 	}
 
 	return TEE_SUCCESS;
-}
-
-uint8_t hw_get_random_byte(void)
-{
-	uint32_t exceptions = 0;
-	uint8_t data = 0;
-
-	assert(trng_base);
-
-	exceptions = cpu_spin_lock_xsave(&trng_random_val_lock);
-
-	/*
-	 * The TRNG generates a whole 32 bits word every 84 cycles. To avoid
-	 * discarding 3 bytes at each request, request 4 bytes of random data
-	 * and return only 1 at each request until there is no more bytes in the
-	 * random_data "cache".
-	 */
-	if (!random_byte_pos)
-		random_data.val = atmel_trng_read32();
-
-	data = random_data.byte[random_byte_pos++];
-	if (random_byte_pos == sizeof(uint32_t))
-		random_byte_pos = 0;
-
-	cpu_spin_unlock_xrestore(&trng_random_val_lock, exceptions);
-
-	return data;
 }
 
 /* This is a true RNG, no need for seeding */
--- a/core/drivers/crypto/caam/caam_rng.c
+++ b/core/drivers/crypto/caam/caam_rng.c
@@ -556,22 +556,12 @@ enum caam_status caam_rng_init(vaddr_t ctrl_addr)
 }
 
 #ifdef CFG_NXP_CAAM_RNG_DRV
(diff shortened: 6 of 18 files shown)


core/crypto/rng_hw.c
```diff
@@ -7,12 +7,25 @@
 #include <tee/tee_cryp_utl.h>
 #include <types_ext.h>
 
+TEE_Result __weak hw_get_random_bytes(void *buf, size_t blen)
```
etienne-lms: I think it is not nice to have 2 functions name that identical
`hw_get_random_byte()`
`hw_get_random_bytes()`

core/crypto/rng_hw.c
```diff
@@ -7,12 +7,25 @@
...
+{
+	uint8_t *b = buf;
+	size_t n;
```
etienne-lms: = 0;

etienne-lms (commented):
@Hi glneo,

I would like to port SMCCC TRNG interface into OP-TEE. It is a standard 'fastcall' SMC to retrieve random byte. In that scope, i need to introduce a new API for HW RNG drivers:
`TEE_Result hw_get_available_entropy(void *buf, size_t *blen);`

I wonder it could also replace the new API you propose `hw_get_random_bytes()`. A generic implementation of `hw_get_random_byte()` and `crypto_rng_read()` would loop on this function to get expected random bytes.
What do you think?

core/crypto/rng_hw.c
glneo: Which is why I removed the singular version by the end of the series, so there is only one of these functions.

core/crypto/rng_hw.c
glneo: ACK

glneo:
How is this `hw_get_available_entropy()` supposed to behave, I'm guessing return as many bytes as it has available up to some set limit? We might want to have drivers implement another function that reports how many bytes are usually (or maximally) returned per call, so we can size the buffer `hw_get_available_entropy()` appropriately.

Either way, this series would still be the first step, getting all the drivers over to a single API. Then we can add a generic `hw_get_random_bytes()` that does as you suggest. And move drivers over one by one, dropping their own `hw_get_random_bytes()` and implementing `hw_max_entropy_size()`/`hw_get_available_entropy()` which would get used by the generic `hw_get_random_bytes()`.

core/arch/arm/plat-rcar/hw_rng.c
```diff
@@ -22,26 +22,32 @@ static unsigned int spin_lock __nex_data = SPINLOCK_UNLOCK;
...
-		uint32_t ret = plat_rom_getrndvector(rng_cache, scratch_buf,
-						     sizeof(scratch_buf));
-
-		if (ret != 0)
-			panic("ROM_GetRndVector() returned error!");
-
-		rng_cache_pos = 0;
+	assert(rng_cache_pos < PLAT_RND_VECTOR_SZ);
+
+	while (buffer_pos < len) {
```
etienne-lms, line 36: is it ok to hold a spinlock  during this loop?

Maybe use
```patch
-uint8_t hw_get_random_byte(void)
+static uint8_t __hw_get_random_byte(void)
 (...)
+{
+   while (len--)
+       *((uint8_t *)buf)++ = __hw_get_random_byte();
+}

File: core/crypto/rng_hw.c (hunk 1 above)
Comment (etienne-lms) [reply core/crypto/rng_hw.c]: ok

File: core/pta/hwrng.c [hunk 4]
```diff
@@ -8,6 +8,7 @@
  */
 
 #include <kernel/pseudo_ta.h>
+#include <crypto/crypto.h>
```
etienne-lms, line 11: move above previous line

core/drivers/atmel_trng.c
```diff
@@ -51,13 +45,12 @@ static uint32_t atmel_trng_read32(void)
...
 }
 
-TEE_Result crypto_rng_read(void *buf, size_t len)
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
 {
 	uint8_t *rngbuf = buf;
 	uint32_t val = 0;
 	size_t len_to_copy = 0;
 
-	assert(buf);
```
etienne-lms, line 60: why removing this assertion?

core/drivers/imx_rngb.c
```diff
@@ -175,8 +175,6 @@ TEE_Result crypto_rng_read(void *buf, size_t len)
 	if (!rngb.ready)
 		return TEE_ERROR_BAD_STATE;
 
-	assert(buf);
```
etienne-lms, line 178: could keep it

core/drivers/dra7_rng.c
```diff
@@ -102,26 +102,30 @@ static void dra7_rng_read64(uint32_t *low_word, uint32_t *high_word)
...
+	static size_t fifo_pos;
+	uint8_t *buffer = (uint8_t *)buf;
+	size_t buffer_pos = 0;
 
 	uint32_t exceptions = cpu_spin_lock_xsave(&rng_lock);
 
-	if (!pos)
-		dra7_rng_read64(&random.val[0], &random.val[1]);
-	ret = random.byte[pos];
+	while (buffer_pos < len) {
```
etienne-lms, line 117: This brings the loop inside the spinlocked block. Make sure it is OK.

etienne-lms (commented):
Ok, I agree. `hw_get_available_entropy()` could be complementary to `hw_get_random_bytes()`.

"core: Add hw_get_random_bytes()" (with comments addressed)
"core: pta: Use common crypto_rng_read() over hw_get_random_byte()" (with comments addressed)
"plar: rcar: Use common crypto_rng_read() over hw_get_random_byte()"
"core: rng_hw: Add dummy hw_get_random_byte()"
"drivers: caam: rng: Switch to hw_get_random_bytes()"
"drivers: se050: rng: Switch to hw_get_random_bytes()"
"drivers: dra7_rng: Use cpu_spin_lock_xsave helper function"

"drivers: dra7_rng: Only fetch RNG base VA once"
"drivers: dra7_rng: Factor out dra7_rng_read64()"


I have a review comment that pops of some of the last commits of this series that update platform drivers: a concern when looping with a spinlock help. For these implementation, I think you should use a less intrusive patch like:
```patch
-uint8_t hw_get_random_byte(void)
+static uint8_t __hw_get_random_byte(void)
 
 (...)
 
+{
+	while (len--)
+		*((uint8_t *)buf)++ = __hw_get_random_byte();
+
+	return TEE_SUCCESS;
+}

File: core/drivers/atmel_trng.c (hunk 5 above)
Comment (glneo) [reply core/drivers/atmel_trng.c:60]: We already check the buffer before calling `hw_get_random_bytes()`, it is guaranteed to be valid by the caller at this point.

File: core/arch/arm/plat-rcar/hw_rng.c (hunk 3 above)
Comment (glneo) [reply core/arch/arm/plat-rcar/hw_rng.c:36]: Not sure why it wouldn't be okay, unless you are suggesting we use something lighter like a regular `mutex_lock`? We need to serialize inside these calls anyway since they update a shared buffer/index.

**Comment by glneo** [comment]
I've added my attempt at this `hw_get_available_entropy()` function and switched the DRA7xx driver over to it.

Drivers can now either implement `hw_get_random_bytes()` xor `hw_get_available_entropy()` and both functions will work with the weak default for the other. After we switch over all drivers over we can remove the weak `hw_get_random_bytes()` and move all its code into 'crypto_rng_read()'.

File: core/arch/arm/plat-rcar/hw_rng.c (hunk 3 above)
Comment (etienne-lms) [reply core/arch/arm/plat-rcar/hw_rng.c:36]: For these drivers, the existing spinlock already protects IOMEM accesses. I don't think all those platform are okay to hold a spinlock and more specifically mask all interrupts by the time a caller gets all its expected random bytes.
All in one, I think the spinlock_xsave/xrestore are ok but if locked/unlocked inside the loop, for each chunk of bytes read.

File: core/crypto/rng_hw.c [hunk 8]
```diff
@@ -3,19 +3,89 @@
...
+static struct mutex rng_fifo_mutex = MUTEX_INITIALIZER;
+
+TEE_Result __weak hw_get_max_available_entropy(size_t *blen)
+{
+	*blen = sizeof(size_t);
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result __weak hw_get_available_entropy(void *buf)
```
etienne-lms, line 25: I think this function should also output the number of bytes returned.
This idea of this function of not to allow driver to issue bigger burst of random byte, but rather let the function output the entropy that is available in the HW, and let caller call again to get the remaining.

So it is a bit strange to propose an implementation of `hw_get_available_entropy()` on top of `hw_get_hw_bytes()`. It should rather be the opposite.

core/crypto/rng_hw.c
```diff
@@ -3,19 +3,89 @@
...
+	 * We do not need to seed our HW RNG, but we do need to allocate
+	 * a buffer for storing extra entropy, set that up here.
+	 */
+	ret = hw_get_max_available_entropy(&rng_fifo_size);
+	if (ret != TEE_SUCCESS) {
+		rng_fifo_size = 0;
+		return ret;
+	}
+
+	rng_fifo = malloc(rng_fifo_size);
```
etienne-lms, line 81: I don't think this is needed. No need to cache the available entropy.

glneo: We need to store the extra entropy somewhere. That is the core of the performance boost. The HW produces a different amount per cycle than the SW consumes. Without this we have to throw away the extra entropy every time. All the drivers already do this, all we are doing here is centralizing it so the logic is not repeated in each driver.

core/crypto/rng_hw.c
glneo: All the hardware I looked at produces entropy at a fixed amount of bytes per iteration. And it is either available or not. So it's easier to have `hw_get_available_entropy()` return a fixed size every time, if it is not available then return nothing. We can query for that size using `hw_get_max_available_entropy()`.

I have made an implementation in this patch that does perform `hw_get_random_bytes()` on top of `hw_get_available_entropy()`. So both ways work and are interoperable.

This `__weak` version `hw_get_available_entropy()` is here only temporary until it is supported natively by all the drivers, then we remove this temporary gasket version. This allows us in the mean time have a common non-blocking API that can implement both SMCCC TRNG and `crypto_rng_read()`.

core/crypto/rng_hw.c
```diff
@@ -3,19 +3,89 @@
...
+					break;
+				else if (ret != TEE_ERROR_BUSY)
+					goto out;
+			}
+		}
+
+		buffer[buffer_pos++] = rng_fifo[rng_fifo_pos++];
+		if (rng_fifo_pos == rng_fifo_size)
+			rng_fifo_pos = 0;
+	}
```
etienne-lms, line 59: The simplest implementation of `hw_get_random_bytes()` over `hw_get_available_entropy()` is rather;
```c
TEE_Result __weak hw_get_random_bytes(void *buf, size_t blen)
{
	uint8_t *ptr = buf;
	size_t rem = blen;
	size_t s = 0;

	while (rem) {
		s = rem;
		res = hw_get_available_entropy(ptr, &s);
		if (res)
			return res;

		ptr += s;
		rem -= s;
	}

	return TEE_SUCCESS;
}
```

And let `hw_get_available_entropy()` use an internal spinlock to read already comulated data.

core/crypto/rng_hw.c
etienne-lms: Some drivers have bigger FIFO and we should not wait the HW FIFO is full to provide random byte.

etienne-lms (commented):
> All the hardware I looked at produces entropy at a fixed amount of bytes per iteration. And it is either available or not. So it's easier to have hw_get_available_entropy() return a fixed size every time, if it is not available then return nothing.

Looking at supported HW, all produce 32bit random burst but some have a deeper FIFO (16bytes for imx_rng.c and stm32_rng.c, up 255 IIUC for bcm_hwrng.c). My intention with `hw_get_available_entropy()` was rather to collect random bytes that the HW has ready, not specifically waiting for its FIFO to be full before providing samples.

core/crypto/rng_hw.c
glneo: With your suggested `hw_get_available_entropy()` we do not wait.

But this FIFO here is used by the function `crypto_rng_read()`/`hw_get_random_bytes()` waiting until the requested amount of random bytes is ready is part of the API.

core/crypto/rng_hw.c
glneo: If we call this version of `hw_get_random_bytes()` asking for 2 bytes for instance, then `hw_get_available_entropy()` would have to either throw out the extra entropy that wasn't requested (HW always reads in bursts of at least 32bits) or buffer the extra for the next call. Former is slower and wastes entropy, the latter all we have done with this "simple" implementation is push the FIFO complexity down into each and every driver instead.

core/pta/hwrng.c
```diff
@@ -31,8 +31,7 @@ static TEE_Result rng_get_entropy(uint32_t types,
 	if (!e)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	for (i = 0; i < params[0].memref.size; i++)
-		e[i] = hw_get_random_byte();
+	crypto_rng_read(e, params[0].memref.size);
```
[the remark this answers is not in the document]
glneo: Was thinking about this, I wasn't sure if we actually wanted to force only HW RNG with this PTA. For instance some plats (stm32mp1) seed the PRNG with a HW RNG, or add entropy from secure time sources, in that case the randomness produced by the PRNG might be "better" than otherwise. And we might want to feed that down into the REE. But maybe not..

I'll add this check here and in the plat-rcar case below.

core/arch/arm/plat-rcar/main.c
```diff
@@ -76,11 +76,9 @@ void console_init(void)
 unsigned long plat_get_aslr_seed(void)
 {
 	unsigned long seed = 0;
-	size_t i;
 
-	/* On RCAR we can call hw_get_random_byte() on early boot stages */
-	for (i = 0; i < sizeof(seed); i++)
-		seed = (seed << 8) | hw_get_random_byte();
+	/* On RCAR we can get hw random bytes on early boot stages */
+	crypto_rng_read(&seed, sizeof(seed));
```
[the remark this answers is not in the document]
glneo: ACK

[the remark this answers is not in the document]
glneo: Interestingly adding this check I found RCAR was not disabling the PRNG, will fix that for them (new second patch in this series).

core/crypto/rng_hw.c
```diff
@@ -3,10 +3,23 @@
...
 
+/*
+ * This is only here to keep the compiler happy while we convert over
+ * platforms. Either hw_get_random_bytes() is overridden or this
+ * function is, in either case this is never called.
+ */
+uint8_t __weak hw_get_random_byte(void)
+{
+	panic();
+	return 4; // chosen by fair dice roll.
```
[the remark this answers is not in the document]
glneo: This function is a temporary to preserve bisectability, it is removed at in the end of the series.

And how often does one get to make life imitate art :smile: https://xkcd.com/221/

core/drivers/hi16xx_rng.c
```diff
@@ -62,15 +62,16 @@ static TEE_Result hi16xx_rng_init(void)
...
 {
 	static vaddr_t r;
 	static int pos;
 	static union {
 		uint32_t val;
 		uint8_t byte[4];
 	} random;
-	uint8_t ret;
+	size_t buffer_pos = 0;
+	uint8_t *buffer = (uint8_t *)buf;
```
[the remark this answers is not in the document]
glneo: ACK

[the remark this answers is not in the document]
glneo: Fixed this in all the patches.

core/crypto/rng_hw.c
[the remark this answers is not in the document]
jenswi-linaro: Haha, let's keep it then. :-)

core/crypto/rng_hw.c
etienne-lms: I think we can have a generic cache for HW accumulated entropy we don't want to trash.

A RNG driver could expose 2 functions:
- `hw_get_entropy_burst_size()` to get no the full HW RNG FIFO size but rather the size of a random data burst, at HW level.
- `hw_get_available_entropy()` to get one or more burst of entropy from the HW and caching them.

A generic RNG API would be
- An generic internal help function `cache_available_entropy()` to get HW ready entropy and accumulated it in the internal cache.
- `crypto_rng_read_available(void *buf, size_t* sz)` would consume only from the cached random bytes.
It would return an error only if the HW reports it. Would use spinlock protection to access cache buffer.
- `crypto_rng_read()` would loop on calling `crypto_rng_read_available()` until the target buffe ris filled or an erro was returned. No need for spinlock or mutex, since this gets handled from `crypto_rng_read_available()`.

I think the legacy function hw_get_random_byte() could be fully removed. Caller should only go through `crypto_rng_read()`.

glneo: Sounds like we are getting closer to an alignment here. Couple minor points.

I'm not sure I'm seeing the need for `hw_get_available_entropy()` to return the amount read, it should always be the full burst size if available or TEE_BUSY if not. So `crypto_rng_read()` can loop on that, and SMCCC TRNG can return instantly as needed. Both cache the extra entropy in a common location, but maybe that is what you are suggesting with `cache_available_entropy()`. (I'm ignoring the bcm_hwrng.c driver here since it doesn't use the normal RNG framework for some reason and so needs fixed separately).

Arguably the SMCCC TRNG API is backwards, it should return the amount available (up to a set amount), not request an amount and fail when not instantly available. The issue being for platforms that generate only up to 32bits without blocking, will always fail if we request 64bits (which the Linux kernel does) as we cannot get that much without blocking for more. So even with partial caching, SMCCC will still need to be called twice every time before it returns a success the second time. Not much we can do about that now, just an observation.

> I think the legacy function hw_get_random_byte() should be fully removed. Callers should only go through crypto_rng_read().

Agree, that is what this series already does. If we are aligned on this, then this series is still the base work that needs to be done either way to get these two tasks done. Then the non-blocking API can be added in the next series on top of this (this one is already longer than like so I don't want to combine them if I can help it).

jforissier:
For commits:
"drivers: hi16xx_rng: Switch to hw_get_random_bytes()"
"core: rng_hw: Remove hw_get_random_byte()"
"core: rng_hw: Remove __weak attribute from HW RNG functions"

jenswi-linaro:
For the commits:
core: Add hw_get_random_bytes()
core: pta: Use common crypto_rng_read() over hw_get_random_byte()
core: rng_hw: Add dummy hw_get_random_byte()
core: rng_hw: Remove hw_get_random_byte()
core: rng_hw: Remove __weak attribute from HW RNG functions
please apply:

For the rest, **excluding**:
plat-stm: Switch to hw_get_random_bytes()
core: Add hw_get_max_available_entropy() and hw_get_available_entropy()
drivers: dra7_rng: Switch to hw_get_available_entropy()
please apply:

I'd rather complete this PR with
"core: rng_hw: Remove __weak attribute from HW RNG functions"
as the last commit. We can take the remaining commits in a new PR.

core/crypto/rng_hw.c
etienne-lms: > Sounds like we are getting closer to an alignment here. Couple minor points.

Yes. Thanks for the discussion. Rather go in the same direction.

> Arguably the SMCCC TRNG API is backwards, it should return the amount available (up to a set amount), not request an amount and fail when not instantly available.

Sorry, I was not clear. Getting available entropy can return 0 bytes without failing.
This is why it provides a buffer of a given size (passed as `len` input argument value) and on return knows how many bytes where found (passed as `len` output argument value).

core/pta/hwrng.c
```diff
@@ -31,8 +36,7 @@ static TEE_Result rng_get_entropy(uint32_t types,
...
 
 	return TEE_SUCCESS;
```
etienne-lms, line 41: `return crypto_rng_read(e, params[0].memref.size);`

etienne-lms (commented):
for commit "core: pta: Use common crypto_rng_read() over hw_get_random_byte()", my R-b tag still applies once comment addressed.

core/arch/arm/plat-rcar/main.c
```diff
@@ -73,14 +73,15 @@ void console_init(void)
...
```
etienne-lms, line 84:
```
	if (crypto_rng_read(&seed, sizeof(seed))
		panic();

**Review by etienne-lms — COMMENTED** [review]
for commit "plar: rcar: Use common crypto_rng_read() over hw_get_random_byte()".
My R-b tag still applies once comment addressed.

File: core/drivers/crypto/caam/caam_rng.c [hunk 17]
```diff
@@ -556,24 +556,11 @@ enum caam_status caam_rng_init(vaddr_t ctrl_addr)
 }
 
 #ifdef CFG_NXP_CAAM_RNG_DRV
-TEE_Result crypto_rng_read(void *buf, size_t blen)
+TEE_Result hw_get_random_bytes(void *buf, size_t blen)
 {
```
etienne-lms, line 560: Preserve `if (!buf) return TEE_ERROR_BAD_PARAMETERS;` (IMHO)

etienne-lms (commented):
comment for commit "drivers: caam: rng: Switch to hw_get_random_bytes()".

core/arch/arm/plat-stm/rng_support.c
```diff
@@ -46,7 +46,7 @@ static inline int hwrng_waithost_fifo_full(void)
 	return 0;
 }
 
-uint8_t hw_get_random_byte(void)
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
```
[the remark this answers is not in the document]
etienne-lms: `plat-stm` is still used. It's still maintained with basic OP-TEE features.

core/crypto/rng_hw.c
```diff
@@ -9,21 +9,21 @@
 #include <types_ext.h>
 
 /* This is a HW RNG, no need for seeding */
-TEE_Result __weak crypto_rng_init(const void *data __unused,
-				  size_t dlen __unused)
+TEE_Result crypto_rng_init(const void *data __unused,
```
etienne-lms, line 66: fits in a single line

core/crypto/rng_hw.c
```diff
@@ -3,19 +3,89 @@
...
+	*blen = sizeof(size_t);
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result __weak hw_get_available_entropy(void *buf)
+{
+	TEE_Result ret;
+
+	ret = hw_get_random_bytes(buf, sizeof(size_t));
```
etienne-lms, line 29: this call may block. Getting available entropy is not expected to block. I think the sole weak implementation of `hw_get_available_entropy()` is to return `TEE_ERROR_NOT_SUPPORTED`.

If driver provide an implementation for `hw_get_available_entropy()`, then fine. If not, there ain't any.

We can still implement a weak generic `hw_get_random_bytes()` that calls `hw_get_available_entropy()`. No need for mutex in that function.

Here the question is does platform driver `hw_get_available_entropy()` caches generated random bytes or is it done from a generic function, in _rng_hw.c_.

core/arch/arm/plat-stm/rng_support.c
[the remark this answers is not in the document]
jenswi-linaro: OK, thanks for the update.

core/arch/arm/plat-stm/rng_support.c
```diff
@@ -85,41 +85,48 @@ uint8_t hw_get_random_byte(void)
...
-	for (i = 0; i < _LOCAL_FIFO_SIZE / 2; i++) {
-		lfifo[pos] = tmpval[i] & 0xFF;
-		pos++;
-		lfifo[pos] = (tmpval[i] >> 8) & 0xFF;
-		pos++;
-	};
-
-	pos = 0;
-	return lfifo[pos];
+	return value;
```
jenswi-linaro, line 131: This is never initialized and also of wrong type.

core/crypto/rng_hw.c
```diff
@@ -3,19 +3,89 @@
...
+	if (ret != TEE_SUCCESS)
+		return ret;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result __weak hw_get_random_bytes(void *buf, size_t blen)
+{
+	uint8_t *buffer = (uint8_t *)buf;
```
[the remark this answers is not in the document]
glneo: ACK

core/drivers/crypto/caam/caam_rng.c
glneo: ACK

core/drivers/crypto/se050/core/rng.c
```diff
@@ -28,20 +28,7 @@ void plat_rng_init(void)
 {
 }
 
-TEE_Result crypto_rng_read(void *buf, size_t blen)
+TEE_Result hw_get_random_bytes(void *buf, size_t blen)
 {
-	if (!buf)
```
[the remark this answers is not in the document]
glneo: ACK

core/crypto/rng_hw.c
glneo: ACK

core/arch/arm/plat-stm/rng_support.c
glneo: Ah, yeah that was meant to just be `return TEE_SUCCESS;`, fixed.

core/crypto/rng_hw.c
glneo: Sounds fair, lets take Jens' suggestion and end this PR before we add this `hw_get_available_entropy()` and take the last two patches of this series into their own. We can continue there.

etienne-lms (commented):
Comment on spin locking for commits:
"drivers: dra7_rng: Switch to hw_get_random_bytes()"
"drivers: hi16xx_rng: Switch to hw_get_random_bytes()"
"plat-rcar: Switch to hw_get_random_bytes()"

Comment on return value for commit:
"plat-stm: Switch to hw_get_random_bytes()"

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit:
"plat-rcar: Software PRNG should be disabled when using hardware generator"

"drivers: atmel_trng: Switch to hw_get_random_bytes()"
"drivers: imx_rngb: Switch to hw_get_random_bytes()"
"core: rng_hw: Remove hw_get_random_byte()"
"core: rng_hw: Remove __weak attribute from HW RNG functions"

jenswi-linaro:
Has this PR stalled?

core/drivers/hi16xx_rng.c
```diff
@@ -79,17 +80,19 @@ uint8_t hw_get_random_byte(void)
 		r = (vaddr_t)phys_to_virt(RNG_BASE, MEM_AREA_IO_SEC, 1) +
 			RNG_NUM;
 
-	if (!pos)
-		random.val = io_read32(r);
+	while (buffer_pos < len) {
```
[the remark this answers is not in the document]
glneo: Okay, moved.

core/arch/arm/plat-stm/rng_support.c
```diff
@@ -85,41 +85,47 @@ uint8_t hw_get_random_byte(void)
...
+			if (hwrng_waithost_fifo_full())
+				return 0;
+
+			/*
+			 * Read the FIFO according to the number of
+			 * expected elements
+			 */
+			for (i = 0; i < _LOCAL_FIFO_SIZE / 2; i++)
+				tmpval[i] = io_read32(rng_base() +
+						RNG_VAL_OFFSET) & 0xFFFF;
```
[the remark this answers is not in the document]
glneo: Fixed.

glneo:
Some new RNG drivers got added and only just now found some free time to convert them.

etienne-lms (commented):
LGTM

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit
"plat-rcar: Software PRNG should be disabled when using hardware generator"

"plat-rcar: Software PRNG should be disabled when using hardware generator"
"drivers: smccc_trng: Switch to hw_get_random_bytes()"
"drivers: stm32_rng: Switch to hw_get_random_bytes()"
"drivers: dra7_rng: Switch to hw_get_random_bytes()"
"drivers: hi16xx_rng: Switch to hw_get_random_bytes()"
"drivers: xiphera_trng: Switch to hw_get_random_bytes()"
"plat-k3: Factor out sa2ul_rng_read128()"
"plat-k3: Switch to hw_get_random_bytes()"
"plat-stm: Switch to hw_get_random_bytes()"

core/arch/arm/plat-rcar/main.c
```diff
@@ -73,14 +73,16 @@ void console_init(void)
...
+
 unsigned long plat_get_aslr_seed(void)
 {
 	unsigned long seed = 0;
-	size_t i;
 
-	/* On RCAR we can call hw_get_random_byte() on early boot stages */
-	for (i = 0; i < sizeof(seed); i++)
-		seed = (seed << 8) | hw_get_random_byte();
+	/* On RCAR we can get hw random bytes on early boot stages */
```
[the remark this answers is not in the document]
glneo: This is not a function name anymore, I'm actually just saying literally that we can "get hw random bytes" here :)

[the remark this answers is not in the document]
etienne-lms: right! i read that too quickly. My apologies for the noise.

PR #5301, opened 2022-05-02
libutee: fix value of TEE_ECC_CURVE_SM2

jforissier:
The GlobalPlatform TEE Interbal Core API specification v1.3 has the
following text:

 In TEE Internal Core API v1.2 and v1.2.1, TEE_ECC_CURVE_25519 and
 TEE_ECC_CURVE_SM2 were incorrectly assigned the same identifier.

Indeed, both were 0x00000300. In v1.3, TEE_ECC_CURVE_SM2 is now
0x00000400. Update the code accordingly.

This is an API and ABI change, but note that this value is used only
in TEE_IsAlgorithmSupported(). Therefore, only TAs that dynamically
check for algorithm support at runtime may be impacted.

### DIFF
--- a/lib/libutee/include/tee_api_defines.h
+++ b/lib/libutee/include/tee_api_defines.h
@@ -288,7 +288,7 @@
 #define TEE_ECC_CURVE_NIST_P256             0x00000003
 #define TEE_ECC_CURVE_NIST_P384             0x00000004
 #define TEE_ECC_CURVE_NIST_P521             0x00000005
-#define TEE_ECC_CURVE_SM2                   0x00000300
+#define TEE_ECC_CURVE_SM2                   0x00000400
 
 
 /* Panicked Functions Identification */


etienne-lms (commented):

jenswi-linaro:
> This is an API and ABI change, but note that this value is used only
> in TEE_IsAlgorithmSupported(). Therefore, only TAs that dynamically
> check for algorithm support at runtime may be impacted.

I don't see how this limits the impacted TAs much. How about the OP-TEE Core usage? That must surely affect the ABI.

jforissier:
> I don't see how this limits the impacted TAs much. How about the OP-TEE Core usage? That must surely affect the ABI.

The value is used internally for instance when the TA generates a key pair, the TA doesn't use `TEE_ECC_CURVE_25519` in the call, it gives `TEE_TYPE_SM2_PKE_KEYPAIR` etc. not sure however what happens when a key is saved then reloaded...

Should we simply swap the values instead? I.e., keep `0x00000300` for `TEE_ECC_CURVE_SM2` and use `0x00000400` for `TEE_ECC_CURVE_25519`. That would be the opposite that GP defines, but since we implemented SM2 prior to X25519 I think it would make more sense.

etienne-lms:
> That would be the opposite that GP defines, 

That would be an issue I think.

> but since we implemented SM2 prior to X25519 I think it would make more sense.

I don't agree. I should stick to the spec, or fully move to vendor defined IDs for these 2 algo, if we fear confusion.
Using TEE_ECC_CURVE_SM2 operations on 25519 objects and vice-versa will report errors. Can't we rely on this?

jenswi-linaro:
GP defines a way for compatible updates of the API. I don't recall exactly how that works but a client can define which version of the API it expects. We should probably start to use that scheme.

For these IDs I agree with @etienne-lms, we can't use anything than what's in the spec. Worst case we break the ABI and increase major version. If we take that path then we should make an inventory of what else can be cleaned out.

sohhass:
I noticed the issue while working on #5297. Currently OP-TEE does not define`API Version` headers that would allow compile time selection of specific version. This is defined in TEE Internal Core API v1.2 section 3.1.1.

jenswi-linaro:
> The value is used internally for instance when the TA generates a key pair, the TA doesn't use `TEE_ECC_CURVE_25519` in the call, it gives `TEE_TYPE_SM2_PKE_KEYPAIR` etc. not sure however what happens when a key is saved then reloaded...

I suppose you meant `TEE_ECC_CURVE_SM2`. After a closer look at this it seems that this might not affect the ABI after all. To be honest I confused it with an object type or attribute. Perhaps a simple test is enough to see that an older TA works with a newer OP-TEE OS.

jforissier:
> > The value is used internally for instance when the TA generates a key pair, the TA doesn't use `TEE_ECC_CURVE_25519` in the call, it gives `TEE_TYPE_SM2_PKE_KEYPAIR` etc. not sure however what happens when a key is saved then reloaded...
> 
> I suppose you meant `TEE_ECC_CURVE_SM2`.

Yes, sorry.

> After a closer look at this it seems that this might not affect the ABI after all. To be honest I confused it with an object type or attribute. Perhaps a simple test is enough to see that an older TA works with a newer OP-TEE OS.

Right. I have tested that, i.e.: build everything for QEMUv8 and run xtest (`make run`); modify optee_os as per this PR and rebuild only the firmware including OP-TEE (`make arm-tf`); run the previously built xtest with its unmodified TAs (`make run-only`). All tests still pass.

jenswi-linaro:

jforissier:
Thanks, I have added a few words about the non-regression test to the commit description.

PR #5302, opened 2022-05-02
core: sp_mem: fix get_cattr() callback name

balint-dobszay-arm:
Commit 8afe7a7c5220 ("core: rename mobj_get_cattr() to mobj_get_mem_type()")
renames the get_cattr() callback in struct mobj_ops(). However, sp_mem
wasn't updated as part of this change, so currently it doesn't compile.
Fix this and get aligned with the new naming.

### DIFF
--- a/core/arch/arm/mm/sp_mem.c
+++ b/core/arch/arm/mm/sp_mem.c
@@ -44,8 +44,7 @@ static size_t mobj_sp_size(size_t num_pages)
 	return s;
 }
 
-struct mobj *sp_mem_new_mobj(uint64_t pages, uint32_t cache_type,
-			     bool is_secure)
+struct mobj *sp_mem_new_mobj(uint64_t pages, uint32_t mem_type, bool is_secure)
 {
 	struct mobj_sp *m = NULL;
 	size_t s = 0;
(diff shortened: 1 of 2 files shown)


balint-dobszay-arm:
Not sure how to make checkpatch happy in this case. If I shorten the referred commit's hash, I get an error that 12+ chars are needed to describe the hash. Now I get a warning of maximum 75 chars per line. But breaking the quoted commit msg into two lines doesn't feel right.

jforissier:
> 



> But breaking the quoted commit msg into two lines doesn't feel right.

...yet, it is the appropriate solution I think :wink:

jforissier (commented):

Would it be appropriate to add a build config to CI?

balint-dobszay-arm:
> > But breaking the quoted commit msg into two lines doesn't feel right.
> 
> ...yet, it is the appropriate solution I think 😉

Okay, fixed 🙂

balint-dobszay-arm:
> 
> Would it be appropriate to add a build config to CI?

Thanks, I added the tag. Yes, AFAICS there is a build config for `CFG_CORE_SEL1_SPMC=y`, but not for `CFG_SECURE_PARTITION=y` too. I'll get back to this.

PR #5307, opened 2022-05-03
plat-ti: Enable hardware RNG PTA

glneo:
When the hardware RNG is available we should also enable
the HWRNG PTA so we can use it from the REE.


jenswi-linaro:

PR #5308, opened 2022-05-03
plat-ti: Add SM handler for setting AMBA IF in the WUGEN

glneo:
This mirrors what is available as a ROM API. We implement this here as
OP-TEE's monitor replaces the ROM's monitor functions and we would like
to keep feature parity.


jenswi-linaro:

PR #5309, opened 2022-05-04
libutils: fix MTE error in realloc()

jenswi-linaro:
With memory tagging realloc() should only untag the old buffer if the
allocation of a new buffer is successful.

The implementation of realloc() prior to this patch is based on bgetr().
This is not so convenient with MTE enabled since the old buffer must be
untagged before calling bgetr(). If realloc() fails the memory tags must
be restored, but this was not done.

Restoring the tags is made somewhat more complicated by the fact that
the size reported by bget_buf_size() may actually be larger than the
size of the tagged buffer. Memalign() is sometimes forced to allocate
buffers larger than requested in order to honor the requested alignment.

So instead of using bgetr(), realloc() is implemented using bget(),
memcpy(), memset() and brel() as needed. This is the same thing that
bgetr() did, except that the old buffer is not untagged until a new
buffer has been successfully allocated.

Fixes: 08a5c4f9ae42 ("libutils: add MTE support in malloc() and friends")

### DIFF
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -499,7 +499,24 @@ void *raw_realloc(void *ptr, size_t hdr_size, size_t ftr_size,
 	if (!s)
 		s++;
 
-	p = bgetr(maybe_untag_buf(ptr), 0, 0, s, &ctx->poolset);
+	p = bget(0, 0, s, &ctx->poolset);
+
+	if (p && ptr) {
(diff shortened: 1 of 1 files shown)


jforissier (approved):
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (vexpress-qemu_armv8a gp_20056)`

jenswi-linaro:
Tags applied.

PR #5310, opened 2022-05-05
symbolize.py: translate TA panic codes

jforissier:
When a TA panics, the TEE core shows the panic code as follows:

 E/TC:? 0 TA panicked with code 0xffff000f

The value is whatever was passed to TEE_Panic() by the TA or the library
that caused the panic; typically, the TEE Intenal Core API functions
implemented in libutee. It quite often happens that this value is a
TEE_Result code. Therefore it is convenient to translate the hexadecimal
value to a symbolic name (TEE_ERROR_*). With this commit, if a match is
found, the name is shown in parentheses like so:

 E/TC:? 0 TA panicked with code 0xffff000f (TEE_ERROR_SECURITY)

If the code doesn't correspond to a known TEE_Result value however
nothing is appended to the output.

Suggested-by: Peter Griffin <peter.griffin@linaro.org>

### DIFF
--- a/scripts/symbolize.py
+++ b/scripts/symbolize.py
@@ -21,6 +21,7 @@ TEE_LOAD_ADDR_RE = re.compile(r'TEE load address @ (?P<load_addr>0x[0-9a-f]+)')
 STACK_ADDR_RE = re.compile(
     r'[UEIDFM]/(TC|LD):(\?*|[0-9]*) [0-9]* +(?P<addr>0x[0-9a-f]+)')
 ABORT_ADDR_RE = re.compile(r'-abort at address (?P<addr>0x[0-9a-f]+)')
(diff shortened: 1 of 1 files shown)


jforissier:
CC @petegriffin

petegriffin:
IMO that is a very helpful additional feature to the script :) As I often find myself having to grep what the error code is not being a every day optee developer.

Acked-by: Peter Griffin <peter.griffin@linaro.org>

jenswi-linaro:
Nice. We could perhaps decode ESR too? Example https://esr.arm64.dev/?#0x623a0400

jenswi-linaro:

jforissier:
> Nice. We could perhaps decode ESR too? Example https://esr.arm64.dev/?#0x623a0400

Good idea, ESR is quite painful to decode manually. I'm adding that to my to-do list :)

PR #5313, opened 2022-05-06
core: decrease register size for gicv2

jenswi-linaro:
The mapped size for GIC distributor and cpu registers is currently
defined to the size used for GICv3. GICv2 doesn't need such large sizes,
in fact some platforms has the distributor and cpu registers next to
each other in the physical memory map. This causes an overlap that can
be confusing. Fix this by selecting a smaller size when a GICv2 is used
instead.

It should be noted GICC_DIR is at offset 0x1000 in the cpu interface so
this register will not be accessible, but this should not be a problem
since OP-TEE doesn't use that register.

### DIFF
--- a/core/include/drivers/gic.h
+++ b/core/include/drivers/gic.h
@@ -9,8 +9,13 @@
 #include <types_ext.h>
 #include <kernel/interrupt.h>
 
+#if defined(CFG_ARM_GICV3)
 #define GIC_DIST_REG_SIZE	0x10000
 #define GIC_CPU_REG_SIZE	0x10000
+#else
+#define GIC_DIST_REG_SIZE	0x1000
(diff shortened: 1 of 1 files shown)


vishnu-banavath (approved):

vishnu-banavath:
@jenswi-linaro  - is there any chance this PR can be rebased to latest master so that I can rebase my changes on top of this commit?

jenswi-linaro:
Rebased. @vishnu-banavath, it seems you have reviewed this. Would you mind giving a tag "Reviewed-by:" or "Acked-by:"?

vishnu-banavath (commented):
Reviewed-by: Vishnu Banavath <vishnu.banavath@arm.com>

jenswi-linaro:
Tag applied, thanks.

PR #5317, opened 2022-05-06
plat-corstone1000: add corstone1000 platform

vishnu-banavath:
These changes are to add corstone1000 platform to optee
core.
arch/arm/plat-vexpress is taken as a reference to make
these changes.

Signed-off-by: Vishnu Banavath <vishnu.banavath@arm.com>

### DIFF
--- a/core/arch/arm/plat-corstone1000/conf.mk
+++ b/core/arch/arm/plat-corstone1000/conf.mk
@@ -1,0 +1,37 @@
+PLATFORM_FLAVOR ?= mps3
+
+$(call force,CFG_HWSUPP_MEM_PERM_WXN,y)
+$(call force,CFG_HWSUPP_MEM_PERM_PXN,y)
+$(call force,CFG_ENABLE_SCTLR_RR,n)
+$(call force,CFG_ENABLE_SCTLR_Z,n)
+
+arm64-platform-cpuarch := cortex-a35
+arm64-platform-cflags += -mcpu=$(arm64-platform-cpuarch)
+arm64-platform-aflags += -mcpu=$(arm64-platform-cpuarch)
+platform-flavor-armv8 := 1
+
+$(call force,CFG_GIC,y)
+$(call force,CFG_PL011,y)
+$(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
+
+$(call force,CFG_ARM64_core,y)
+
+CFG_WITH_STATS ?= y
+
+CFG_WITH_ARM_TRUSTED_FW ?= y
+CFG_WITH_LPAE ?=y
+
+CFG_TEE_CORE_NB_CORE = 1
+CFG_TZDRAM_START ?= 0x02002000
+CFG_TZDRAM_SIZE  ?= 0x000FE000
+CFG_TEE_RAM_VA_SIZE ?= 0x00AF000
+CFG_SHMEM_START  ?= 0x86000000
+CFG_SHMEM_SIZE   ?= 0x00200000
+
+CFG_DDR_SIZE ?= 0x80000000
+CFG_DT_ADDR ?= 0x82100000
+CFG_DTB_MAX_SIZE ?= 0x100000
+
+$(call force,CFG_PSCI_ARM64,y)
+$(call force,CFG_DT,y)
+$(call force,CFG_EXTERNAL_DTB_OVERLAY,y)
--- a/core/arch/arm/plat-corstone1000/main.c
+++ b/core/arch/arm/plat-corstone1000/main.c
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2022, Arm Limited
+ */
+
+#include <console.h>
+#include <drivers/gic.h>
+#include <drivers/pl011.h>
+#include <initcall.h>
+#include <keep.h>
+#include <kernel/boot.h>
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
+#include <kernel/panic.h>
+#include <kernel/tee_time.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <stdint.h>
+#include <trace.h>
+
+static struct gic_data gic_data __nex_bss;
+static struct pl011_data console_data __nex_bss;
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, PL011_REG_SIZE);
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICC_BASE, GIC_DIST_REG_SIZE);
+
+void main_init_gic(void)
+{
+	gic_init_base_addr(&gic_data, GIC_BASE + GICC_OFFSET,
+			   GIC_BASE + GICD_OFFSET);
+
+	itr_init(&gic_data.chip);
+}
+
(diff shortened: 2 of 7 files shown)


core/arch/arm/plat-corstone1000/platform_config.h
```diff
@@ -0,0 +1,46 @@
...
+
+#define CONSOLE_UART_BASE	UART1_BASE
+
+#define DRAM0_BASE		0x80000000
+#define DRAM0_SIZE		0x7f000000
+
+#define GICD_OFFSET		0x10000
+#define GICC_OFFSET		0x2f000
+
+#ifdef GIC_BASE
```
jenswi-linaro, line 28: This ifdef can be removed. Same below with UART_BAUDRATE and CONSOLE_BAUDRATE

core/arch/arm/plat-corstone1000/platform_config.h
```diff
@@ -0,0 +1,46 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Arm Limited
+ */
+
+#ifndef PLATFORM_CONFIG_H
+#define PLATFORM_CONFIG_H
+
+#include <mm/generic_ram_layout.h>
```
jenswi-linaro, line 9: It looks like these two included aren't needed.

core/arch/arm/plat-corstone1000/main.c
```diff
@@ -0,0 +1,77 @@
...
+
+static struct gic_data gic_data __nex_bss;
+static struct pl011_data console_data __nex_bss;
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, PL011_REG_SIZE);
+#ifdef DRAM0_BASE
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+#endif
+
+#ifdef GIC_BASE
```
jenswi-linaro, line 34: Can drop this ifdef, GIC_BASE is always defined. Same above with DRAM0_BASE.

core/arch/arm/plat-corstone1000/main.c
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2022, Arm Limited
+ */
+
+#include <arm64.h>
```
jenswi-linaro, line 6: Please remove the includes below which are not needed.

core/arch/arm/plat-corstone1000/main.c
```diff
@@ -0,0 +1,77 @@
...
+void console_init(void)
+{
+	pl011_init(&console_data, CONSOLE_UART_BASE, CONSOLE_UART_CLK_IN_HZ,
+		   CONSOLE_BAUDRATE);
+	register_serial_console(&console_data.chip);
+}
+
+void ffa_secondary_cpu_boot_req(vaddr_t secondary_ep, uint64_t cookie)
+{
+	DMSG("This is single core platform\n");
```
jenswi-linaro, line 75: The `\n` is not needed, the trace macros adds that if it's not there.

jenswi-linaro (commented):
Would you like to add an entry in `/MAINTAINERS`?

jenswi-linaro:
Please fix the checkpatch complaints.

core/arch/arm/plat-corstone1000/platform_config.h
```diff
@@ -0,0 +1,37 @@
...
+
+#define GICD_BASE		(GIC_BASE + GICD_OFFSET)
+#define GICC_BASE		(GIC_BASE + GICC_OFFSET)
+
+#define UART_BAUDRATE		115200
+#define CONSOLE_BAUDRATE	UART_BAUDRATE
+
+#define SYS_COUNTER_FREQ_IN_TICKS	UL(50000000) /* 32MHz */
+
+#define CONSOLE_UART_CLK_IN_HZ	UL(50000000) /* 32MHz*/
```
jforissier, line 35: 50,000,000 Hz isn't 32 MHz

core/arch/arm/plat-corstone1000/conf.mk
```diff
@@ -0,0 +1,37 @@
...
+
+CFG_WITH_ARM_TRUSTED_FW ?= y
+CFG_WITH_LPAE ?=y
+
+CFG_TEE_CORE_NB_CORE = 1
+CFG_TZDRAM_START ?= 0x02002000
+CFG_TZDRAM_SIZE  ?= 0x000FE000
+CFG_TEE_RAM_VA_SIZE ?= 0x00AF000
+CFG_SHMEM_START  ?= 0x86000000
+CFG_SHMEM_SIZE   ?= 0x00200000
```
jforissier, line 29: Open question: does it really make sense for newer platforms to carve out some memory for OP-TEE SHM? Now we have `CFG_CORE_DYN_SHM=y` by default.

jenswi-linaro: I suppose there could be some special case where you really want it, but I believe that's quite uncommon.

core/arch/arm/plat-corstone1000/main.c
```diff
@@ -0,0 +1,68 @@
...
+static struct gic_data gic_data __nex_bss;
+static struct pl011_data console_data __nex_bss;
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, PL011_REG_SIZE);
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICC_BASE, GIC_DIST_REG_SIZE);
+
+void main_init_gic(void)
```
jenswi-linaro, line 37: Please compare with with https://github.com/OP-TEE/optee_os/blob/2e1b85fe61f3a2ff52bb8df91c47732e5108e8df/core/arch/arm/plat-vexpress/main.c#L56-L66

It seems that https://github.com/OP-TEE/optee_os/pull/5313 is relevant here too.

jenswi-linaro:

etienne-lms (commented):
for commit "plat-corstone1000: add corstone1000 platform"

core/arch/arm/plat-corstone1000/conf.mk
```diff
@@ -0,0 +1,37 @@
...
+$(call force,CFG_GIC,y)
+$(call force,CFG_PL011,y)
+$(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
+
+$(call force,CFG_ARM64_core,y)
+
+CFG_WITH_STATS ?= y
+
+CFG_WITH_ARM_TRUSTED_FW ?= y
+CFG_WITH_LPAE ?=y
```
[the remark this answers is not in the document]
jforissier: @etienne-lms this is a common pattern, I think it should be in `core/arch/arm/arm.mk` instead. I will propose a patch.

jenswi-linaro:
Please rebase this PR to catch the fixes we've mentioned here.

core/arch/arm/plat-corstone1000/platform_config.h
jforissier: Yes, @etienne-lms may be able to explain better but I think including `<mm/generic_ram_layout.h>` allows platforms to make sure that `TEE_RAM_VA_START` and `TEE_RAM_VA_SIZE` (and possibly others) are defined based on CFG values or default values. But since some platforms (used to?) hard-code values in their `platform_config.h`, the callers always include `<platform.h>` not `<mm/generic_ram_layout.h>`. There is one exception which does not look correct though: `core/arch/arm/plat-imx/drivers/tzc380.c` (but it probably doesn't matter because all i.MX flavors use the generic mechanism).

PR #5323, opened 2022-05-09
core: rpmb: __weak access implementation selectable at build time

ldts:
Avoid having to patch the source code to prevent accessing RPMB by
mistake.

### DIFF
--- a/core/tee/tee_rpmb_fs.c
+++ b/core/tee/tee_rpmb_fs.c
@@ -3081,5 +3081,5 @@ TEE_Result tee_rpmb_fs_raw_open(const char *fname, bool create,
 
 bool __weak plat_rpmb_key_is_ready(void)
 {
-	return true;
+	return false;
 }


core/tee/tee_rpmb_fs.c
```diff
@@ -3081,5 +3081,8 @@ TEE_Result tee_rpmb_fs_raw_open(const char *fname, bool create,
 
 bool __weak plat_rpmb_key_is_ready(void)
 {
-	return true;
+	if (IS_ENABLED(CFG_RPMP_IS_READY_WEAK))
```
jenswi-linaro, line 3084: Wasn't the point with the `__weak` implementation that it should be easily overridden?

ldts: yes, well kind of:  we should not have to modify the source code to override the platform's weak behavior: some platforms might want to return false to prevent the RPMB persistent keys to be written (or even just return false for some builds for the same platform).

ldts: it is often the case that the RPMB keys depend on the SoC security state so having a more restricted behavior for the weak implementation seems sensible to avoid being locked out.

jenswi-linaro: It seems that the behaviour you're looking for is actually platform specific. The point with this `__weak` function is that the platform should override it if it needs anything but the default behaviour.
I suggest that you provide this function in the platform you're referring to.

Another option to achieve the result you're looking for is to compile with `CFG_RPMB_FS=n`

ldts: This is default is basically saying yes, we allow you to access RPMB and we hope you used the keys that you wanted to use because otherwise we are just breaking your hardware; instead, the default should be no, we do not give you access to RPMB unless you are sure that you have the right keys and for that, please do use a platform specific implementation.

jforissier: Don't we have `CFG_RPMB_WRITE_KEY ?= n` by default already?

ldts: yes and that is a good default: so why is the final `plat_rpmb_key_is_ready` just  opposite? that is what I want to know...because it is burning a fuse and it is clearly inconsitent.

I see no contradiction on a user wanting to exercise the RPMB subsystem fully and then gatekeep at platform level.
That config allows the RPMB subsystem to just attempt burning the key; the platform - usually depending on its security state - then might chose to allow or deny it.

jenswi-linaro: > If you can provide a solid argument for the current default that is all I need. I do not think you can.

It seems that you want RPMB when enabled to be by default unusable. My use case is that I'd like RPMB when enabled to be usable.

Do note that the dangerous flag here is `CFG_RPMB_WRITE_KEY`. It would in my opinion be much more safe to not program the RPMB key automatically (for multiple reasons), instead it should only be done if explicitly requested. The details of this scheme remains to be defined.

ldts: I can see how this can be convenient:  platform agnostic access want the RPMB enabled flag (which is disabled by default) to bypass the platform (hence why it is enabled by default). 

why not then just articulate it more explicitly, something like this?

```
bool __weak plat_rpmb_key_is_ready(void)
{
-	return true;
+	if (IS_ENABLED(CFG_RPMB_WRITE_KEY))
+                     return true;
```

For me the platform - just as the subsystem- should also disable access by default as enabling it might require certain knowledge (could be security states, fuses, whatever). Fusing is always a risky operation.

IMO we could make OP-TEE more coherent when writing fuses at OS but also at platform level: it is weird that at subsystem level we use one default and at platform level just the opposite when they both control access to the same resource. Just messy and prone to errors....even thought it might be convenient not having to consider the platform sometimes.

jenswi-linaro: `plat_rpmb_key_is_ready()` is called only if `CFG_RPMB_WRITE_KEY` is `y`.

ldts: of course. but it doesnt mean it needs to return true. it could return false.

it returns true because it is convenient for developers that do not worry about the platform....hence the patch above: it explicitly maps the behavior that is being enforced (ie, ignore the platform, we just enable it because the RPMB subsystem is the gatekeeper)

I would still merge my PR or simply remove the __weak function.

ldts: of course, that is the way the code is working today (with a level of indirection for the platform to have its saying which makes sense). The only thing we disagree is with the default which is clearly wrong when you care about the platform.

ldts: that default should be false.

ldts: let me try from another angle: what does `plat_rpmb_key_is_ready` even mean? because of course RPMB is ready - why wouldnt it?....this is not about RPMB being ready, this is about protecting the user from the risk of losing the RPMB access some time in the future.

usually, _platform ready_ bases its decision in knowing how the RPMB keys are generated and knowing that there are some deterministic actions - ie closing the board- that will affect the key generation.

When the weak implementation defaults `plat_rpmb_key_is_ready` is because there is a guarantee that access will not be lost no matter what you do: ie, that the keys will always be accessible.

That should not be the default, that is a lot to ask to the average user: and yes CFG_RPMB_WRITE_KEY is completely unrelated to plat_rpmb_key_is_ready and yes they both need to be set separately.

Emantor: The original intention of `plat_rpmb_key_is_ready` was to add a platform switch to prevent users from flashing their boards in the wrong state. The only user was the i.MX platform, where the HUK key is insecure as long as the platform is not in Trusted Boot mode.

This isn't necessarily the case on other SoCs, since its very SoC specific. Therefore I chose to implement the `__weak` default with a true return value, since this doesn't change anything on existing other platforms.

Whats your usecase for setting the new switch `CFG_RPMB_IS_READY_WEAK` to `n`? Because at the moment it defaults to the previous default of true in the default configuration in the current iteration of the patch.

ldts: so what is the outcome? revert to default? use config?

Emantor: I vote for changing the default. My only worry is that there is somebody out there using the current default, however care has to be taken anyway when switching to newer versions of OP-TEE. It should also be documented in the release notes.

jenswi-linaro: I don't see how this helps at all. For the key to be written you currently need to change two flags `CFG_RPMB_FS` and `CFG_RPMB_WRITE_KEY` from their default `n` value to `y`. How will it help to have a third flag which you need to do the same thing with? The problem IMO is that the key is written automatically regardless of how many flags you add to tell that you really really mean business this time.

Emantor: I think my comment is not quite clear. I am for removing the CFG and instead switching the weak implementation from returning true to returning false.

ldts: RPMB has a functional layer and a platform layer. The platform layer policy tries to guarantee that the access keys wont be lost when securing the platform. The fact that the user sets CFG_RPMB_WRITE_KEY is orthogonal to the platform decision. And by default, the platform should apply the more restrictive policy when dealing with persistent operations.
I wouldn't recommend taking shortcuts and mixing up responsibility layers that is all...just complicates maintenance and adds issues.

jenswi-linaro: What is this platform layer, `tee_otp_get_hw_unique_key()`?

ldts: `plat_rpmb_key_is_ready` is the code that provides the guarantees to the user that she will not lose access to RPMB once the board has been secured (or whatever condition the platform wants to protect against).
Since the RPMB key depends on the HUK and the HUK depends on the security state on iMX CAAM supported boards and Zynqmp platform, usually the condition is the board being closed.
so really, that is why it gates the access to RPMB at the very last step..

jenswi-linaro: `plat_rpmb_key_is_ready()` is changed to always return `false`, then must all platforms that wishes to use RPMB override that function. If so then we could just as well remove it and require the platform to provide it if `CFG_RPMB_WRITE_KEY=y`. That will for sure trigger a lot of questions.

Overriding `plat_rpmb_key_is_ready()` is only useful if there's some hardware state that can be checked to tell if it seems that the generated key will be the final key.

Having another config variable to control the default value of the `__weak` `plat_rpmb_key_is_ready()` would mean essentially adding a third flag to tell that you really really mean business this time. That doesn't make sense, you're just moving the problem.

I understand that it's quite annoying to burn the wrong key into RPMB and worst case make the board unusable. I'm still of the opinion that the problem is that `CFG_RPMB_WRITE_KEY=y` triggers automatic burning of that key. From security point of view `CFG_RPMB_WRITE_KEY` isn't optimal either, you're forced to flash the board with an OP-TEE binary that must never leak on a production device. It can be hard to guarantee that stuff you erase from flash can't be found again.

jenswi-linaro: It looks like the documentation should be updated.
https://optee.readthedocs.io/en/latest/architecture/secure_storage.html#important-caveats
and
https://optee.readthedocs.io/en/latest/building/efi_vars/stmm.html?highlight=CFG_RPMB_WRITE_KEY#op-tee-build-instructions
could be more clear.

jenswi-linaro: Thanks

Emantor: @ldts please do, I am at a conference ATM.

ldts: @jenswi-linaro https://github.com/OP-TEE/optee_docs/pull/157 not sure if this is enough or if you wanted a different angle to the issue we have been discussing

ldts: > @ldts please do, I am at a conference ATM.

ah nice.

core/tee/tee_rpmb_fs.c
```diff
@@ -3081,5 +3081,5 @@ TEE_Result tee_rpmb_fs_raw_open(const char *fname, bool create,
 
 bool __weak plat_rpmb_key_is_ready(void)
 {
-	return true;
+	return false;
```
jenswi-linaro, line 3084: If `plat_rpmb_key_is_ready()` is changed to always return `false`, then must all platforms that wishes to use RPMB override that function. If so then we could just as well remove it and require the platform to provide it if `CFG_RPMB_WRITE_KEY=y`. That will for sure trigger a lot of questions.

Overriding `plat_rpmb_key_is_ready()` is in my opinion only useful if there's some hardware state that can be checked to tell if it seems that the generated key will be the final key.

ldts: Yes if the default is set to false, it will impact current users. But for some it might be for the best, protecting their RPMB hardware. For others is an inconvenience with no monetary cost...which could be worked around with an additional configuration.

When you work enabling hardware on different platforms, the current situation is nasty. Securing a board is not a trivial activity and this just adds pain to injury.

jenswi-linaro: It's a good point, but if we're to impact all users of RPMB we should try to make the best of it.
Is this the best thing we can do?
Do you have any comments to https://github.com/OP-TEE/optee_os/issues/5338?

ldts:
I am closing this PR.
Worth mentioning that this is impacting users that need RPMB. And not in a positive way.
IMO this should have been merged while other implementations take place. Bad decision.

jforissier:
@ldts changes get merged when a consensus is reached. Here I don't think it is the case. So until other discussions reach a conclusion (#5338), I consider the status quo is preferable.

PR #5326, opened 2022-05-11
drivers: caam: fix cache operation on SGT table

clementfaure:
The cache operation of the SGT table in caam_sgt_derive() was wrong and
it did not take into account the CAAM "burst" defined by the value
CFG_CAAM_SGT_ALIGN.
The cache operation of the SGT table in caam_sgt_cache_op() is done
correctly however.

This patch adds caam_sgt_entries_cache_op() to do this operation and
avoid implementation errors.

Signed-off-by: Franck LENORMAND <franck.lenormand@nxp.com>

### DIFF
--- a/core/drivers/crypto/caam/utils/utils_sgt.c
+++ b/core/drivers/crypto/caam/utils/utils_sgt.c
@@ -15,6 +15,18 @@
 #include <tee/cache.h>
 #include <util.h>
 
+/*
+ * Perform cache management clean operation on the SGT table entry
+ *
+ * @sgtbuf     SGT table to manage
+ */
+static void caam_sgt_entries_cache_clean(const struct caamsgtbuf *sgtbuf)
+{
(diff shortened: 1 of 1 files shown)


core/drivers/crypto/caam/utils/utils_sgt.c
```diff
@@ -15,16 +15,26 @@
 #include <tee/cache.h>
 #include <util.h>
 
+/*
+ * Perform cache management clean operation on the SGT table entry
+ *
+ * @sgtbuf     SGT table to manage
+ */
+static void caam_sgt_entries_cache_op(const struct caamsgtbuf *sgtbuf)
```
jforissier, line 23: `clean` instead of `op`?

jforissier (commented):
One suggestion below.

core/drivers/crypto/caam/utils/utils_sgt.c
clementfaure: Ok

PR #5327, opened 2022-05-11
core: imx: link: generate entry_point_address.txt for uTee image

clementfaure:
For some platforms like mx7ulpevk, the `UL` attribute for CFG_DRAM_BASE
is necessary to avoid the following error:

$ PLATFORM=imx-mx7ulpevk make uTee
core/mm/mobj.c: In function ‘mobj_init’:
./out/arm-plat-imx/include/generated/conf.h:222:64: warning: integer overflow in expression of type ‘int’ results in ‘-1644167168’ [-Woverflow]
  222 | #define CFG_TZDRAM_START (0x60000000 - 0x01e00000 - 0x00200000 + 0x40000000)
      |                                                                ^

However, this `UL` attribute prevents the shell from properly computing
the CFG_TZDRAM_START value :

$ PLATFORM=imx-mx7ulpevk make uTee
bash: (UL(0x60000000) - 0x01e00000 - 0x00200000 + 0x40000000): missing `)' (error token is "(0x60000000) - 0x01e00000 - 0x00200000 + 0x40000000)")

To address both issues, the OP-TEE OS start address is read from tee.elf
and stored in entry_point_address.txt file.

Fixes: 0f8347dcafe ("core: imx: generate uImage for imx6 and imx7 platforms")

### DIFF
--- a/core/arch/arm/plat-imx/link.mk
+++ b/core/arch/arm/plat-imx/link.mk
@@ -7,5 +7,5 @@ uTee: $(link-out-dir)/uTee
 cleanfiles += $(link-out-dir)/uTee
 $(link-out-dir)/uTee: $(link-out-dir)/tee-raw.bin
 	@$(cmd-echo-silent) '  MKIMAGE $@'
-	$(q)ADDR=`printf 0x%x $$(($(CFG_TZDRAM_START)))`; \
+	$(q)ADDR=`printf 0x%x $$(($(subst UL,,$(CFG_TZDRAM_START))))`; \
 		mkimage -A arm -O linux -C none -a $$ADDR -e $$ADDR -d $< $@


core/arch/arm/plat-imx/link.mk
```diff
@@ -2,10 +2,15 @@ include core/arch/arm/kernel/link.mk
...
+	$(q)$(READELFcore) -h $(link-out-dir)/tee.elf | awk '/Entry point address/ {print $$4}' > $@
+
 .PHONY: uTee
 uTee: $(link-out-dir)/uTee
 cleanfiles += $(link-out-dir)/uTee
-$(link-out-dir)/uTee: $(link-out-dir)/tee-raw.bin
+$(link-out-dir)/uTee: $(link-out-dir)/tee-raw.bin $(link-out-dir)/entry_point_address.txt
 	@$(cmd-echo-silent) '  MKIMAGE $@'
-	$(q)ADDR=`printf 0x%x $$(($(CFG_TZDRAM_START)))`; \
+	$(q)@ADDR=`cat $(link-out-dir)/entry_point_address.txt`; \
```
jforissier, line 15: `$(subst UL,,$(CFG_TZDRAM_START))` would work too...

clementfaure: That would work indeed. I thought about doing something similar but since there are other similar attributes like `U`, `ULL`, `L` ... We would also have to consider the other cases ... no?

jforissier: Why would a physical address ever need to be encoded as an unsigned long long? IMO it should always be a unsigned long, and often an unsigned (int) is enough and we drop the UL. But that's it, no?

clementfaure: Knowing that `CFG_TZDRAM_START` could be the result of multiple additions, I thought that considering all these attributes was also a good idea.
But yes, we should only get `UL`.

jforissier (approved):

clementfaure:
Tag applied! Thanks!

PR #5328, opened 2022-05-11
K3 SA2UL true random-number generator driver

glneo:
This series adds a couple commands to the TI-SCI driver for device control and firewall setup. We use these to enable and secure the SA2UL device available on all K3 platforms. One of the parts of the SA2UL being a TRNG which we add support for here. Support for the other SA2UL HW crypto engines will be added in subsequent series.

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul.c
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Texas Instruments K3 SA2UL Driver
+ *
+ * Copyright (C) 2022 Texas Instruments Incorporated - https://www.ti.com/
+ *	Andrew Davis <afd@ti.com>
+ */
+
+#include <drivers/ti_sci.h>
+#include <initcall.h>
+#include <io.h>
+#include <keep.h>
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
+#include <kernel/spinlock.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <rng_support.h>
+
+#include "sa2ul.h"
+
+#define	SA2UL_ES                0x0008
+#define SA2UL_ES_TRNG           BIT(3)
+#define	SA2UL_EEC               0x1000
+#define SA2UL_EEC_TRNG          BIT(3)
+
+#define FW_ENABLE_REGION        0x0a
+#define FW_BIG_ARM_PRIVID       0x01
+#define FW_WILDCARD_PRIVID      0xc3
+#define FW_SECURE_ONLY          GENMASK_32(8, 0)
+#define FW_NON_SECURE           GENMASK_32(16, 0)
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, SA2UL_BASE, SA2UL_REG_SIZE);
+
+static TEE_Result sa2ul_init(void)
+{
+	vaddr_t sa2ul = (vaddr_t)phys_to_virt(SA2UL_BASE, MEM_AREA_IO_SEC,
+					      RNG_REG_SIZE);
+	uint16_t fwl_id = SA2UL_TI_SCI_FW_ID;
+	uint16_t sa2ul_region = SA2UL_TI_SCI_FW_RGN_ID;
+	uint16_t rng_region = RNG_TI_SCI_FW_RGN_ID;
+	uint8_t owner_index = OPTEE_HOST_ID;
+	uint8_t owner_privid = 0;
+	uint16_t owner_permission_bits = 0;
+	uint32_t control = 0;
+	uint32_t permissions[FWL_MAX_PRIVID_SLOTS] = { };
+	uint64_t start_address = 0;
+	uint64_t end_address = 0;
+	uint32_t val = 0;
+	TEE_Result result = TEE_SUCCESS;
+	int ret = 0;
+
+	/* Power on the SA2UL device */
+	ret = ti_sci_device_get(SA2UL_TI_SCI_DEV_ID);
+	if (ret) {
+		EMSG("Failed to get SA2UL device");
+		return TEE_ERROR_GENERIC;
+	}
+
+	IMSG("Activated SA2UL device");
+
+	/* Try to claim the SA2UL firewall for ourselves */
+	ret = ti_sci_change_fwl_owner(fwl_id, sa2ul_region, owner_index,
+				      &owner_privid, &owner_permission_bits);
+	if (ret) {
+		/*
+		 * This is not fatal, it just means we are on an HS device
+		 * where the DMSC already owns the SA2UL. On GP we need
+		 * to do additional setup for access permissions below.
+		 */
+		DMSG("Could not change SA2UL firewall owner");
+	} else {
+		DMSG("Fixing SA2UL firewall owner for GP device");
+
+		/* Get current SA2UL firewall configuration */
+		ret = ti_sci_get_fwl_region(fwl_id, sa2ul_region, 1,
+					    &control, permissions,
+					    &start_address, &end_address);
+		if (ret) {
+			EMSG("Could not get firewall region information");
+			return TEE_ERROR_GENERIC;
+		}
+
+		/* Modify SA2UL firewall to allow all others access*/
+		control = FW_ENABLE_REGION;
+		permissions[0] = (FW_WILDCARD_PRIVID << 16) | FW_NON_SECURE;
+		ret = ti_sci_set_fwl_region(fwl_id, sa2ul_region, 1,
+					    control, permissions,
+					    start_address, end_address);
+		if (ret) {
+			EMSG("Could not set firewall region information");
+			return TEE_ERROR_GENERIC;
+		}
+	}
+
+	/* Claim the TRNG firewall for ourselves */
+	ret = ti_sci_change_fwl_owner(fwl_id, rng_region, owner_index,
+				      &owner_privid, &owner_permission_bits);
+	if (ret) {
+		EMSG("Could not change TRNG firewall owner");
+		return TEE_ERROR_GENERIC;
+	}
+
+	/* Get current TRNG firewall configuration */
+	ret = ti_sci_get_fwl_region(fwl_id, rng_region, 1,
+				    &control, permissions,
+				    &start_address, &end_address);
+	if (ret) {
+		EMSG("Could not get firewall region information");
+		return TEE_ERROR_GENERIC;
+	}
+
+	/* Modify TRNG firewall to block all others access */
+	control = FW_ENABLE_REGION;
+	permissions[0] = (FW_BIG_ARM_PRIVID << 16) | FW_SECURE_ONLY;
+	ret = ti_sci_set_fwl_region(fwl_id, rng_region, 1,
+				    control, permissions,
+				    start_address, end_address);
+	if (ret) {
+		EMSG("Could not set firewall region information");
+		return TEE_ERROR_GENERIC;
+	}
+
+	IMSG("Enabled firewalls for SA2UL TRNG device");
+
+	/* Enable RNG engine in SA2UL if not already enabled */
+	val = io_read32(sa2ul + SA2UL_ES);
+	if (!(val & SA2UL_ES_TRNG)) {
+		IMSG("Enabling SA2UL TRNG engine");
+		io_setbits32(sa2ul + SA2UL_EEC, SA2UL_EEC_TRNG);
+	}
+
+	/* Initialize the RNG Module */
+	result = sa2ul_rng_init();
+	if (result != TEE_SUCCESS)
+		return result;
+
+	IMSG("SA2UL Drivers initialized");
+
+	return TEE_SUCCESS;
+}
+driver_init(sa2ul_init);
--- a/core/arch/arm/plat-k3/drivers/sa2ul.h
+++ b/core/arch/arm/plat-k3/drivers/sa2ul.h
@@ -1,0 +1,16 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Texas Instruments K3 SA2UL Driver
+ *
+ * Copyright (C) 2022 Texas Instruments Incorporated - https://www.ti.com/
+ *	Andrew Davis <afd@ti.com>
+ */
+
+#ifndef __DRIVERS_SA2UL_H
+#define __DRIVERS_SA2UL_H
+
+#include <tee_api_types.h>
+
+TEE_Result sa2ul_rng_init(void);
+
+#endif /* __DRIVERS_SA2UL_H */
--- a/core/arch/arm/plat-k3/drivers/sa2ul_rng.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul_rng.c
@@ -1,0 +1,142 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Texas Instruments K3 SA2UL RNG Driver
+ *
+ * Copyright (C) 2022 Texas Instruments Incorporated - https://www.ti.com/
+ *	Andrew Davis <afd@ti.com>
+ */
+
+#include <initcall.h>
+#include <io.h>
+#include <keep.h>
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
+#include <kernel/spinlock.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <rng_support.h>
+
+#include "sa2ul.h"
+
+#define	RNG_OUTPUT_0            0x00
+#define	RNG_OUTPUT_1            0x04
+#define	RNG_OUTPUT_2            0x08
+#define	RNG_OUTPUT_3            0x0C
+#define	RNG_STATUS              0x10
+#define RNG_READY               BIT(0)
+#define SHUTDOWN_OFLO           BIT(1)
+#define	RNG_INTACK              0x10
+#define	RNG_CONTROL             0x14
+#define ENABLE_TRNG             BIT(10)
+#define	RNG_CONFIG              0x18
+#define	RNG_ALARMCNT            0x1C
+#define	RNG_FROENABLE           0x20
+#define	RNG_FRODETUNE           0x24
+#define	RNG_ALARMMASK           0x28
+#define	RNG_ALARMSTOP           0x2C
+#define	RNG_OPTIONS             0x78
+#define	RNG_EIP_REV             0x7C
+
+#define RNG_CONTROL_STARTUP_CYCLES_SHIFT        16
+#define RNG_CONTROL_STARTUP_CYCLES_MASK         GENMASK_32(31, 16)
+
+#define RNG_CONFIG_MAX_REFIL_CYCLES_SHIFT       16
+#define RNG_CONFIG_MAX_REFIL_CYCLES_MASK        GENMASK_32(31, 16)
+#define RNG_CONFIG_MIN_REFIL_CYCLES_SHIFT       0
+#define RNG_CONFIG_MIN_REFIL_CYCLES_MASK        GENMASK_32(7, 0)
+
+#define RNG_ALARMCNT_ALARM_TH_SHIFT             0
+#define RNG_ALARMCNT_ALARM_TH_MASK              GENMASK_32(7, 0)
+#define RNG_ALARMCNT_SHUTDOWN_TH_SHIFT          16
+#define RNG_ALARMCNT_SHUTDOWN_TH_MASK           GENMASK_32(20, 16)
+
+#define RNG_CONTROL_STARTUP_CYCLES              0xff
+#define RNG_CONFIG_MIN_REFIL_CYCLES             0x5
+#define RNG_CONFIG_MAX_REFIL_CYCLES             0x22
+#define RNG_ALARM_THRESHOLD                     0xff
+#define RNG_SHUTDOWN_THRESHOLD                  0x4
+
+#define RNG_FRO_MASK    GENMASK_32(23, 0)
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, RNG_BASE, RNG_REG_SIZE);
+
+static unsigned int rng_lock = SPINLOCK_UNLOCK;
+static vaddr_t rng;
+
+uint8_t hw_get_random_byte(void)
+{
+	static int pos;
+	static union {
+		uint32_t val[4];
+		uint8_t byte[16];
+	} random;
+	uint32_t exceptions = 0;
+	uint8_t ret = 0;
+
+	assert(rng);
+
+	exceptions = cpu_spin_lock_xsave(&rng_lock);
+
+	if (!pos) {
+		/* Is the result ready (available)? */
+		while (!(io_read32(rng + RNG_STATUS) & RNG_READY)) {
+			/* Is the shutdown threshold reached? */
+			if (io_read32(rng + RNG_STATUS) & SHUTDOWN_OFLO) {
+				uint32_t alarm = io_read32(rng + RNG_ALARMSTOP);
+				uint32_t tune = io_read32(rng + RNG_FRODETUNE);
+
+				/* Clear the alarm events */
+				io_write32(rng + RNG_ALARMMASK, 0x0);
+				io_write32(rng + RNG_ALARMSTOP, 0x0);
+				/* De-tune offending FROs */
+				io_write32(rng + RNG_FRODETUNE, tune ^ alarm);
+				/* Re-enable the shut down FROs */
+				io_write32(rng + RNG_FROENABLE, RNG_FRO_MASK);
+				/* Clear the shutdown overflow event */
+				io_write32(rng + RNG_INTACK, SHUTDOWN_OFLO);
+
+				DMSG("Fixed FRO shutdown");
+			}
+		}
+		/* Read random value */
+		random.val[0] = io_read32(rng + RNG_OUTPUT_0);
+		random.val[1] = io_read32(rng + RNG_OUTPUT_1);
+		random.val[2] = io_read32(rng + RNG_OUTPUT_2);
+		random.val[3] = io_read32(rng + RNG_OUTPUT_3);
+		/* Acknowledge read complete */
+		io_write32(rng + RNG_INTACK, RNG_READY);
+	}
+
+	ret = random.byte[pos];
+
+	pos = (pos + 1) % 16;
+
+	cpu_spin_unlock_xrestore(&rng_lock, exceptions);
+
+	return ret;
+}
+
+TEE_Result sa2ul_rng_init(void)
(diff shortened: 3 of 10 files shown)


core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -179,6 +179,63 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info)
...
+	return 0;
+}
+
+/**
+ * ti_sci_device_get() - Get a device managed by TISCI
+ *
+ * @id:		Device Identifier
+ *
+ * Return: 0 if all goes well, else appropriate error message
+ */
```
jforissier, line 221: It is something I had not noticed in previous TI driver... you should avoid duplicating the function documentation. The doc block should preferably be in the header file (except for static functions of course).

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -179,6 +179,63 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info)
...
+ * @state:	State to move the device to
+ *
+ * Return: 0 if all goes well, else appropriate error message
+ */
+static int ti_sci_device_set_state(uint32_t id, uint32_t flags, uint8_t state)
+{
+	struct ti_sci_msg_req_set_device_state req = { };
+	struct ti_sci_msg_resp_set_device_state resp = { };
+	struct ti_sci_xfer xfer = { };
+	int ret;
```
jforissier, line 196: `= 0;`

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -179,6 +179,63 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info)
...
+		return ret;
+
+	req.id = id;
+	req.state = state;
+
+	ret = ti_sci_do_xfer(&xfer);
+	if (ret)
+		return ret;
+
+	return 0;
```
jforissier, line 212:
```suggestion
	return ti_sci_do_xfer(&xfer);
```

jforissier (commented):
For commit "plat-k3: drivers: ti-sci: Add support for setting device state", with the comments below addressed:

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -236,6 +236,162 @@ int ti_sci_device_put(uint32_t id)
...
+ *                      firewall, this field indicates the channel in question
+ *                      (index starting from 0)
+ * @n_permission_regs:  Number of permission registers to set
+ * @control:            Contents of the firewall CONTROL register to set
+ * @permissions:        Contents of the firewall PERMISSION register to set
+ * @start_address:      Contents of the firewall START_ADDRESS register to set
+ * @end_address:        Contents of the firewall END_ADDRESS register to set
+ *
+ * Return: 0 if all went well, else returns appropriate error value.
+ */
```
jforissier, line 257: Remove this doc block

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -236,6 +236,162 @@ int ti_sci_device_put(uint32_t id)
...
+	req.permissions[1] = permissions[1];
+	req.permissions[2] = permissions[2];
+	req.start_address = start_address;
+	req.end_address = end_address;
+
+	ret = ti_sci_do_xfer(&xfer);
+	if (ret)
+		return ret;
+
+	return 0;
```
jforissier, line 289:
```suggestion
	return ti_sci_do_xfer(&xfer);
```

jforissier (commented):
For commit "plat-k3: drivers: ti-sci: Add support for setting firewall state", with the comments below and the checkpatch warnings addressed:

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -0,0 +1,143 @@
...
+#include <platform_config.h>
+#include <rng_support.h>
+
+#include "drivers/ti_sci.h"
+#include "sa2ul.h"
+
+#define	SA2UL_ES                0x0008
+#  define SA2UL_ES_TRNG         BIT(3)
+#define	SA2UL_EEC               0x1000
+#  define SA2UL_EEC_TRNG        BIT(3)
```
jforissier, line 26: No spaces please (`#define`)

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -0,0 +1,143 @@
...
+	uint16_t rng_region = RNG_TI_SCI_FW_RGN_ID;
+	uint8_t owner_index = OPTEE_HOST_ID;
+	uint8_t owner_privid;
+	uint16_t owner_permission_bits;
+	uint32_t control;
+	uint32_t permissions[FWL_MAX_PRIVID_SLOTS];
+	uint64_t start_address;
+	uint64_t end_address;
+	uint32_t val;
+	int ret;
```
jforissier, line 51: `= 0;` / `= { };`

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -0,0 +1,148 @@
...
+#define	RNG_OUTPUT_0            0x00
+#define	RNG_OUTPUT_1            0x04
+#define	RNG_OUTPUT_2            0x08
+#define	RNG_OUTPUT_3            0x0C
+#define	RNG_STATUS              0x10
+#  define RNG_READY             BIT(0)
+#  define SHUTDOWN_OFLO         BIT(1)
+#define	RNG_INTACK              0x10
+#define	RNG_CONTROL             0x14
+#  define ENABLE_TRNG           BIT(10)
```
jforissier, line 31: `#define`

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -0,0 +1,148 @@
...
+static vaddr_t rng;
+
+uint8_t hw_get_random_byte(void)
+{
+	static int pos;
+	static union {
+		uint32_t val[4];
+		uint8_t byte[16];
+	} random;
+	uint8_t ret;
```
jforissier, line 74: Initializations missing

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -0,0 +1,148 @@
...
+	pos = (pos + 1) % 16;
+
+	cpu_spin_unlock_xrestore(&rng_lock, exceptions);
+
+	return ret;
+}
+
+TEE_Result sa2ul_rng_init(void)
+{
+	uint32_t val;
```
jforissier, line 119: `= 0;` here rather than below

jforissier (commented):
For commit "plat-k3: drivers: Add SA2UL RNG driver", with the comments below and the checkpatch warnings addressed:

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: ACK

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: Many of these functions have to read in returned data after the `ti_sci_do_xfer`. Keeping the format this way makes all the functions consistent, even the ones that don't read out anything. It also matches the same in TF-A. I'd like to keep the two drivers in sync as close as I can, so fixes in one can be matched and go into the other as easily as possible.

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: ACK

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: Same comment as above.

core/arch/arm/plat-k3/drivers/ti_sci.c
jforissier: The functions consistency argument is a bit weak IMO, but the consistency with TF-A is reasonable for easier maintenance. You can leave the code as is.

etienne-lms (commented):
minor comments only.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -179,6 +179,40 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info)
...
+{
+	struct ti_sci_msg_req_set_device_state req = { };
+	struct ti_sci_msg_resp_set_device_state resp = { };
+	struct ti_sci_xfer xfer = { };
+	int ret = 0;
+
+	ret = ti_sci_setup_xfer(TI_SCI_MSG_SET_DEVICE_STATE, flags,
+				&req, sizeof(req),
+				&resp, sizeof(resp),
+				&xfer);
```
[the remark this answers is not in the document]
glneo: Would mess with TF-A code sync.

core/arch/arm/plat-k3/drivers/ti_sci.h
```diff
@@ -22,6 +22,24 @@
  */
 int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info);
 
+/**
+ * Device control operations
+ *
+ * - ti_sci_device_get - command to request for device managed by TISCI
```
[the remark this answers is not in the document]
glneo: These are the non-exclusive access versions of these functions. We cant get the SA2UL exclusively on HS (shared with DMSC) so I didn't add the exclusive versions just yet.

Will cleanup the comment a bit anyway.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -213,6 +213,105 @@ int ti_sci_device_put(uint32_t id)
...
+	return 0;
+}
+
+int ti_sci_change_fwl_owner(uint16_t fwl_id, uint16_t region,
+			    uint8_t owner_index, uint8_t *owner_privid,
+			    uint16_t *owner_permission_bits)
+{
+	struct ti_sci_msg_req_fwl_change_owner_info req;
+	struct ti_sci_msg_resp_fwl_change_owner_info resp;
+	struct ti_sci_xfer xfer;
```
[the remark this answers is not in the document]
glneo: ACK

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -213,6 +213,105 @@ int ti_sci_device_put(uint32_t id)
...
+	req.owner_index = owner_index;
+
+	ret = ti_sci_do_xfer(&xfer);
+	if (ret)
+		return ret;
+
+	*owner_privid = resp.owner_privid;
+	*owner_permission_bits = resp.owner_permission_bits;
+
+	return ret;
```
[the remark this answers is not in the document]
glneo: ACK

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -0,0 +1,142 @@
...
+	io_write32(rng + RNG_CONFIG, val);
+
+	/* Configure the desired FROs */
+	io_write32(rng + RNG_FRODETUNE, 0x0);
+
+	/* Enable all FROs */
+	io_write32(rng + RNG_FROENABLE, 0xffffff);
+
+	val = ENABLE_TRNG;
+	io_write32(rng + RNG_CONTROL, val);
```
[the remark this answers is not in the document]
glneo: Sure

core/arch/arm/plat-k3/drivers/sa2ul.h
```diff
@@ -0,0 +1,18 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Texas Instruments K3 SA2UL Driver
+ *
+ * Copyright (C) 2022 Texas Instruments Incorporated - https://www.ti.com/
+ *	Andrew Davis <afd@ti.com>
+ */
+
+#ifndef SA2UL_H
+#define SA2UL_H
```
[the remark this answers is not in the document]
glneo: But we are not in drivers/..

Will fix anyway

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -213,6 +213,105 @@ int ti_sci_device_put(uint32_t id)
...
+				&resp, sizeof(resp),
+				&xfer);
+	if (ret)
+		return ret;
+
+	req.fwl_id = fwl_id;
+	req.region = region;
+	req.n_permission_regs = n_permission_regs;
+	req.control = control;
+	req.permissions[0] = permissions[0];
```
jenswi-linaro, line 237: Wouldn't it make sense to use `memcmp(req.permissions, permissions, sizeof(req.permissions))` instead?

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -213,6 +213,105 @@ int ti_sci_device_put(uint32_t id)
...
+	ret = ti_sci_do_xfer(&xfer);
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
+int ti_sci_get_fwl_region(uint16_t fwl_id, uint16_t region,
+			  uint32_t n_permission_regs, uint32_t *control,
+			  uint32_t (*permissions)[FWL_MAX_PRIVID_SLOTS],
```
jenswi-linaro, line 252: I'm not sure what this means, but it looks like `uint32_t permissions[FWL_MAX_PRIVID_SLOTS]` should be enough and avoid the funny looking assignments below.

jenswi-linaro: This is the first time I've seen an array passed in such a way (as a symbolic gesture) so I very much doubt that this is something people in general would expect for an output array.

A simple array would be preferable. Input only arrays are normally `const` while output arrays are without that attribute.

core/arch/arm/plat-k3/drivers/ti_sci.c
jenswi-linaro: Yeah, it's not a big deal. My main concern is that `FWL_MAX_PRIVID_SLOTS` is used above, but here we rely on the knowledge that `FWL_MAX_PRIVID_SLOTS` always will be `3`.

If wrote this code I would have left a `static_assert(FWL_MAX_PRIVID_SLOTS == 3)` around here. Do as you like.

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: It makes it easier for the user of the function too, they just pass all the output variables prefixed with `&`, even the array `permissions`:

```
ti_sci_get_fwl_region(fwl_id, sa2ul_region, 1,
					    &control, &permissions,
					    &start_address, &end_address);
```

Otherwise it would be the only output that doesn't have `&`.

I'm not inflexible here, if we want const and pass only the array itself then I can switch.

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: `FWL_MAX_PRIVID_SLOTS` will always be 3, I just make it a define to avoid using 3 as a magic number where I can, I'll add that `static_assert` for sanity sake.

glneo: Hmm, looking at it, since `n_permission_regs` defines the amount of these slots we set, I can just loop on that. Plus a check for `n_permission_regs <= FWL_MAX_PRIVID_SLOTS` and I think we solve both issues.

core/arch/arm/plat-k3/drivers/ti_sci.c
jenswi-linaro: Please use const instead.

jenswi-linaro:

core/arch/arm/plat-k3/drivers/ti_sci.h
```diff
@@ -40,6 +40,75 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info);
 int ti_sci_device_get(uint32_t id);
 int ti_sci_device_put(uint32_t id);
 
+/**
+ * ti_sci_set_fwl_region() - Request for configuring a firewall region
+ *
+ * @fwl_id:             Firewall ID in question. fwl_id is defined in the TRM.
+ * @region:             Region or channel number to set config info This field
```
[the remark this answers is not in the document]
glneo: Fixed.

core/arch/arm/plat-k3/drivers/ti_sci.h
```diff
@@ -40,6 +40,75 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info);
...
+ *                      is unused in case of a simple firewall and must be
+ *                      initialized to zero. In case of a region based
+ *                      firewall, this field indicates the region in question.
+ *                      (index starting from 0) In case of a channel based
```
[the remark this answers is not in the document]
glneo: Fixed.

core/arch/arm/plat-k3/drivers/ti_sci.h
```diff
@@ -40,6 +40,75 @@ int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info);
...
+ *                      firewall, this field indicates the channel in question
+ *                      (index starting from 0)
```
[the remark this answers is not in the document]
glneo: Fixed.

core/arch/arm/plat-k3/drivers/ti_sci_protocol.h
```diff
@@ -139,6 +142,126 @@ struct ti_sci_msg_resp_set_device_state {
...
+#define FWL_MAX_PRIVID_SLOTS 3U
+
+/**
+ * struct ti_sci_msg_req_fwl_set_firewall_region - Set region firewall permissions
+ * @hdr:		Generic Header
+ * @fwl_id:		Firewall ID
+ * @region:		Region or channel number to set config info.
+ *			This field is unused in case of a simple firewall and
+ *			must be initialized to zero.  In case of a region based
+ *			firewall, this field indicates the region. (index
```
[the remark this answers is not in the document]
glneo: Fixed.

core/arch/arm/plat-k3/drivers/ti_sci_protocol.h
```diff
@@ -139,6 +142,126 @@ struct ti_sci_msg_resp_set_device_state {
...
+ *			starting from 0) In case of a channel based firewall,
+ *			this field indicates the channel (index starting
+ *			from 0)
```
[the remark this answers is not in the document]
glneo: Fixed.

core/arch/arm/plat-k3/main.c
```diff
@@ -19,10 +19,8 @@
 #include <stdint.h>
 #include <string_ext.h>
 
-#if defined(PLATFORM_FLAVOR_am65x) || defined(PLATFORM_FLAVOR_j721e)
 #include "drivers/sec_proxy.h"
 #include "drivers/ti_sci.h"
```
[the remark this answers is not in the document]
glneo: Moved.

core/arch/arm/plat-k3/main.c
```diff
@@ -61,14 +59,10 @@ void console_init(void)
...
-#if defined(PLATFORM_FLAVOR_am65x) || defined(PLATFORM_FLAVOR_j721e)
-TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+static TEE_Result init_ti_sci(void)
 {
-	uint8_t dkek[SA2UL_DKEK_KEY_LEN] = { 0 };
 	int ret = 0;
 
-	assert(SA2UL_DKEK_KEY_LEN >= HW_UNIQUE_KEY_LENGTH);
-
 	ret = k3_sec_proxy_init();
```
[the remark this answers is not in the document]
glneo: Will switch to TEE_Result as ret type. Some later patch can convert `ti_sci_init()` over, that issue wasn't introduced in this series (neither was this issue, but while I'm here I'll fix this one up)

core/arch/arm/plat-k3/main.c
```diff
@@ -77,14 +71,27 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
+	uint8_t dkek[SA2UL_DKEK_KEY_LEN] = { };
+	int ret = 0;
+
+	assert(SA2UL_DKEK_KEY_LEN >= HW_UNIQUE_KEY_LENGTH);
+
 	ret = ti_sci_get_dkek(0, "OP-TEE", "DKEK", dkek);
-	if (ret)
+	if (ret) {
+		EMSG("Could not get HUK");
 		return ret;
```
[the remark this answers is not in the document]
glneo: `TEE_ERROR_SECURITY` seems to be more commonly returned when HUK fails, will return that.

core/arch/arm/plat-k3/main.c
```diff
@@ -77,14 +71,27 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 	if (ret)
```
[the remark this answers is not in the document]
glneo: Fixed.

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -0,0 +1,142 @@
...
+#include <keep.h>
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
+#include <kernel/spinlock.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <rng_support.h>
+
+#include "drivers/ti_sci.h"
```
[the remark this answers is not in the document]
glneo: Moved.

core/arch/arm/plat-k3/drivers/sa2ul.h
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Copyright (C) 2022 Texas Instruments Incorporated - https://www.ti.com/
+ *	Andrew Davis <afd@ti.com>
+ */
+
+#ifndef __DRIVERS_SA2UL_H
+#define __DRIVERS_SA2UL_H
+
+#include <stddef.h>
+#include <stdint.h>
```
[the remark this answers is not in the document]
glneo: Removed.

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -0,0 +1,142 @@
...
+	uint64_t start_address = 0;
+	uint64_t end_address = 0;
+	uint32_t val = 0;
+	int ret = 0;
+
+	/* Power on the SA2UL device */
+	ret = ti_sci_device_get(SA2UL_TI_SCI_DEV_ID);
+	if (ret) {
+		EMSG("Failed to get SA2UL device");
+		return TEE_ERROR_GENERIC;
```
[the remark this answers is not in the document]
etienne-lms: ok

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -0,0 +1,142 @@
...
+				io_write32(rng + RNG_ALARMMASK, 0x0);
+				io_write32(rng + RNG_ALARMSTOP, 0x0);
+				/* De-tune offending FROs */
+				io_write32(rng + RNG_FRODETUNE, tune ^ alarm);
+				/* Re-enable the shut down FROs */
+				io_write32(rng + RNG_FROENABLE, RNG_FRO_MASK);
+				/* Clear the shutdown overflow event */
+				io_write32(rng + RNG_INTACK, SHUTDOWN_OFLO);
+
+				DMSG("Fixed FRO shutdown\n");
```
etienne-lms, line 99: remove trailing `\n`

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -0,0 +1,142 @@
...
+		random.val[3] = io_read32(rng + RNG_OUTPUT_3);
+		/* Acknowledge read complete */
+		io_write32(rng + RNG_INTACK, RNG_READY);
+	}
+
+	ret = random.byte[pos];
+
+	pos = (pos + 1) % 16;
+
+	cpu_spin_unlock_xrestore(&rng_lock, exceptions);
```
etienne-lms, line 115: this masks interrupt during the loop processing. Make sure it is ok and that interrupts will not be masked for too long.

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -0,0 +1,142 @@
...
+	val = io_read32(sa2ul + SA2UL_ES);
+	if (!(val & SA2UL_ES_TRNG)) {
+		IMSG("Enabling SA2UL TRNG engine");
+		io_setbits32(sa2ul + SA2UL_EEC, SA2UL_EEC_TRNG);
+	}
+
+	/* Initialize the RNG Module */
+	ret = sa2ul_rng_init();
+	if (ret != TEE_SUCCESS)
+		return ret;
```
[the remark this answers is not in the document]
glneo: 'sa2ul_rng_init()' already returns a `TEE_Result`, to avoid the conversion to int, I'll add a `TEE_Result` var and use that here.

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
glneo: That is okay for now, we don't wait for results that long in practice. Plus this gets fixed in a better way after #5300, which I can start back work on after this series goes in.

glneo:
Thanks, ACKs added.

jforissier:
Added Jens' Ack and merged, thanks!

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -0,0 +1,143 @@
...
+	/* Try to claim the SA2UL firewall for ourselves */
+	ret = ti_sci_change_fwl_owner(fwl_id, sa2ul_region, owner_index,
+				      &owner_privid, &owner_permission_bits);
+	if (ret) {
+		/*
+		 * This is not fatal, it just means we are on an HS device
+		 * where the DMSC already owns the SA2UL. On GP we need
+		 * to do additional setup for access permissions below.
+		 */
+		DMSG("Could not change SA2UL firewall owner");
```
BaochengSu, line 72: On HS device, there will be some error printed from inside `ti_sci_change_fwl_owner`:

```
I/TC: Activated SA2UL device
E/TC:0 0 ti_sci_get_response:120 Message not acknowledged
E/TC:0 0 ti_sci_do_xfer:149 Failed to get response (-65536)
I/TC: Enabled firewalls for SA2UL TRNG device
```

This is really annoying. At the first glance, people would wonder if there is something wrong with their device.

Is it possible to check if the current device is not HS, then call this function?

PR #5329, opened 2022-05-12
[Hikey] Define SDP in embedded dtb

omasse-linaro:
Allow definition of the Secure Data Path memory region in an embedded dtb
It remove SDP memory intersection checking as embedded DTB is not
available during init of tee core memory mapping

Comply with reserved memory bindings
Linux documentation file:
Documentation/devicetree/bindings/reserved-memory/reserved-memory.yaml

Signed-off-by: Olivier Masse <olivier.masse@nxp.com>


omasse-linaro:
Tested on Hikey960.

core/mm/core_mmu.c
```diff
@@ -451,28 +504,47 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
 #ifdef CFG_SECURE_DATA_PATH
 static bool pbuf_is_sdp_mem(paddr_t pbuf, size_t len)
 {
+	if (sec_sdp.size != 0)
```
jenswi-linaro, line 507: if (sec_sdp.size)

core/mm/core_mmu.c
```diff
@@ -451,28 +504,47 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
+	const struct core_mmu_phys_mem *mem = NULL;
+	struct mobj **mobj_base = NULL;
+	struct mobj **mobj = NULL;
+	paddr_t addr = sec_sdp.paddr;
+	paddr_size_t size = sec_sdp.size;
+	int cnt = 1;
+	int i = 0;
 
 	/* SDP mobjs table must end with a NULL entry */
 	mobj_base = calloc(cnt + 1, sizeof(struct mobj *));
```
jenswi-linaro, line 527: This should be done after `cnt` has been determined.

The logic below isn't very easy to follow. Perhaps it would be better to not mix the two cases in the loop.

omasse-linaro: ok, maybe let's split the logic like that
`struct mobj **core_sdp_mem_create_mobjs(void)
{
	struct mobj **mobj_base = NULL;
	struct mobj **mobj = NULL;
	int i = 0;

	/* SDP mobjs table must end with a NULL entry */
	mobj_base = calloc(cnt + 1, sizeof(struct mobj *));
	if (!mobj_base)
		panic("Out of memory");

	mobj = mobj_base;

	if (!sec_sdp.size) {
		int cnt = phys_sdp_mem_end - phys_sdp_mem_begin;
		const struct core_mmu_phys_mem *mem = phys_sdp_mem_begin;

		for (i = 0; i < cnt; i++) {
			mobj[i] = mobj_phys_alloc(mem[i].addr, mem[i].size,
						  TEE_MATTR_MEM_TYPE_CACHED,
						  CORE_MEM_SDP_MEM);
			if (!mobj[i])
				panic("can't create SDP physical memory object");
		}
	} else {
		mobj[0] = mobj_phys_alloc(sec_sdp.paddr, sec_sdp.size,
					  TEE_MATTR_MEM_TYPE_CACHED,
					  CORE_MEM_SDP_MEM);
		if (!mobj[0])
			panic("can't create SDP physical memory object");
	}

	return mobj_base;
}
`

jenswi-linaro: Yes, something like that.

core/mm/core_mmu.c
omasse-linaro: done

core/mm/core_mmu.c
omasse-linaro: done

core/mm/core_mmu.c
```diff
@@ -451,30 +504,52 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
+
 struct mobj **core_sdp_mem_create_mobjs(void)
 {
-	const struct core_mmu_phys_mem *mem;
-	struct mobj **mobj_base;
-	struct mobj **mobj;
-	int cnt = phys_sdp_mem_end - phys_sdp_mem_begin;
+	struct mobj **mobj_base = NULL;
+	struct mobj **mobj = NULL;
+	int cnt = (sec_sdp.size ? 1 : (phys_sdp_mem_end - phys_sdp_mem_begin));
```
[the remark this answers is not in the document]
omasse-linaro: done

core/arch/arm/dts/hikey_sdp.dts
```diff
@@ -0,0 +1,22 @@
...
+/ {
+	#address-cells = <1>;
+	#size-cells = <0>;
+
+	reserved-memory {
+		#address-cells = <1>;
+		#size-cells = <1>;
+
+		sdp@0x3e800000 {
+			compatible = "linaro,secure-heap";
```
jenswi-linaro, line 17: Where is this "compatible" documented?

omasse-linaro: vendor prefix was taken from bindings documentation
https://www.kernel.org/doc/Documentation/devicetree/bindings/vendor-prefixes.yaml

secure-heap property here is added as a new feature.

jenswi-linaro: OK I understand that to mean that anything "linaro," can be documented elsewhere.
I guess it should still be documented somewhere. Since the kernel is going to use this it seems natural to document it in the kernel tree to. Or am I missing something?

omasse-linaro: probably something like that in Documentation/devicetree/bindings/reserved-memory/linaro,secure-heap.yaml
# SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
%YAML 1.2
---
$id: http://devicetree.org/schemas/reserved-memory/linaro,secure-heap.yaml#
$schema: http://devicetree.org/meta-schemas/core.yaml#

title: Linaro Secure DMABUF Heap

maintainers:
  - Olivier Masse <olivier.masse@nxp.com>

description: On Linaro OPTEE, the Secure Data Path feature
  need a reserved memory.

allOf:
  - $ref: "reserved-memory.yaml"

properties:
  compatible:
    const: linaro,secure-heap

  reg:
    description: region of memory reserved for OPTEE SDP feature

unevaluatedProperties: false

required:
  - compatible
  - reg

omasse-linaro: done here
https://github.com/linaro-swg/linux/pull/101/commits/0c7ef9e676c79ca3f6b565944bb0d39f23750f6b

jenswi-linaro: Thanks

jenswi-linaro:
Please split this into two commits. One which is generic (with a "core: " prefix) and one which is hikey specific with a "plat-hikey: " prefix.

core/mm/core_mmu.c
```diff
@@ -73,6 +75,8 @@ static struct memaccess_area nsec_shared[] __nex_data = {
 };
 
 #if defined(CFG_SECURE_DATA_PATH)
+static const char *tz_sdp_match = "linaro,secure-heap";
+static struct memaccess_area sec_sdp;
 #ifdef CFG_TEE_SDP_MEM_BASE
```
jforissier, line 80: Could `CFG_TEE_SDP_MEM_BASE` and `CFG_TEE_SDP_MEM_SIZE` be removed from the code and the config files since the new method is to use the embedded DT? I see no benefit in allowing both methods; the upgrade seems easy enough.

jforissier (commented):
One comment regarding "core: Define SDP in embedded dtb" (should be DTB not dtb by the way), then for both commits:

core/mm/core_mmu.c
omasse-linaro: ok, do we still need to support TEE_SDP_TEST_MEM_BASE ?

core/mm/core_mmu.c
```diff
@@ -255,6 +259,48 @@ static struct tee_mmap_region *find_map_by_pa(unsigned long pa)
 	return NULL;
 }
 
+#if defined(CFG_SECURE_DATA_PATH)
```
etienne-lms, line 262: maybe remove if `#ifdef` and return `false` when not enabled at line 270?

core/mm/core_mmu.c
```diff
@@ -255,6 +259,48 @@ static struct tee_mmap_region *find_map_by_pa(unsigned long pa)
...
+	if (!IS_ENABLED(CFG_EMBED_DTB))
+		return false;
+
+	fdt = get_embedded_dt();
+	if (!fdt)
+		panic("No DTB found");
+
+	node = fdt_node_offset_by_compatible(fdt, node, tz_sdp_match);
+	if (node < 0) {
+		EMSG("Cannot find %s node in the device tree",
```
etienne-lms, line 279: This is not an error message. Use `DMSG()`.

BTW, fits in a single line.

core/mm/core_mmu.c
etienne-lms: I agree with you but some platforms to use `CFG_TEE_SDP_MEM_BASE/_SIZE` (hikey :), marvel, poplar, vexpress). We should get their feedback first.

core/mm/core_mmu.c
```diff
@@ -373,8 +419,15 @@ void core_mmu_set_discovered_nsec_ddr(struct core_mmu_phys_mem *start,
...
-		carve_out_phys_mem(&m, &num_elems, pmem->addr, pmem->size);
+	if (dtb_get_sdp_region()) {
+		DMSG("region base %p %zu", (void *)sec_sdp.paddr, sec_sdp.size);
+
+		carve_out_phys_mem(&m, &num_elems, sec_sdp.paddr, sec_sdp.size);
+	} else {
+		for (pmem = phys_sdp_mem_begin; pmem < phys_sdp_mem_end; pmem++)
+			carve_out_phys_mem(&m, &num_elems, pmem->addr,
+					   pmem->size);
+	}
```
etienne-lms, line 430: Should we report an error (trace or even panic) when a platform defines SDP memory areas both in DT and in its build config switches? I don't think one should take precedence, it would rather look like a platform misconfiguration that should be reported in some way.

omasse-linaro: in fact, the current implementation is looking for a SDP DTB definition in any, by default. 
But it will fallback to the build config switches otherwise, which allow usage of TEE_SDP_TEST_MEM_BASE/TEE_SDP_TEST_MEM_SIZE.

core/mm/core_mmu.c
jforissier: @omasse-linaro the test pool could still be created if SDP is enabled but not defined in the embedded DT I suppose.
Anyway let's leave that for later.

jforissier (commented):
More comments on "plat-hikey: Add embedded DTB to define SDP" which I reviewed a bit too quickly ;)

core/mm/core_mmu.c
etienne-lms: Maybe the implementation should support both: SDP mem pools defined from embedded DT and those defined with `register_sdp_mem()`.

omasse-linaro: yes it's actually the current behavior.
for now the SDP mem pools defined from embedded DT is used if it exist, otherwise the register_sdp_mem() from build config switches is taken.
do you mean that we could define multiple SDP area from DT and/or from build config ?

etienne-lms: Optee_os current implementation allows several occurrences of `register_sdp_mem()`.
The DT binding you propose also allows several heap pools to be defined.
I guess optee_os could take all of them into account to forge a list of SDP areas.

omasse-linaro: I would see one use case to such a list of SDP areas, it could allow SDP memory separation for several Trusted Application.
However have you another use case for supporting multiple SDP areas ?

etienne-lms: Flexibility.
If a platform defines SDP pools from both embedded DTB and from built-in platform implementation, there no reason to exclude some.

omasse-linaro: done here 804bd35faadac5521310fa07e32d693e2c59814c

core/mm/core_mmu.c
```diff
@@ -255,6 +259,47 @@ static struct tee_mmap_region *find_map_by_pa(unsigned long pa)
...
+	size_t tmp_size = 0;
+
+	if (!IS_ENABLED(CFG_EMBED_DTB))
+		return false;
+
+	fdt = get_embedded_dt();
+	if (!fdt)
+		panic("No DTB found");
+
+	node = fdt_node_offset_by_compatible(fdt, node, tz_sdp_match);
```
etienne-lms, line 277: nitpicking: prefer use an explicit `-1` 2nd argument rather than `node` inited to -1.

Or even better, use a loop to support several compatible nodes if found, and create an array of `struct core_mmu_phys_mem`. But maybe in a later change if you want this initial support to be merged before coming 3.18.0 tag.

core/mm/core_mmu.c
```diff
@@ -451,30 +502,57 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
-					CORE_MEM_SDP_MEM);
-		if (!*mobj)
-			panic("can't create SDP physical memory object");
-	}
+	mobj = mobj_base;
+
+	for (mem = phys_sdp_mem_begin;
+	     mem < phys_sdp_mem_end; mem++, mobj++)
+		*mobj = core_sdp_mem_alloc_mobj(mem->addr,
+						mem->size);
```
etienne-lms, line 552: restore previous instructions that checked the allocation succeeds.

core/mm/core_mmu.c
```diff
@@ -255,6 +259,47 @@ static struct tee_mmap_region *find_map_by_pa(unsigned long pa)
...
+
+	fdt = get_embedded_dt();
+	if (!fdt)
+		panic("No DTB found");
+
+	node = fdt_node_offset_by_compatible(fdt, -1, tz_sdp_match);
+	if (node < 0) {
+		DMSG("Cannot find %s node in the device tree", tz_sdp_match);
+		return false;
+	}
```
etienne-lms, line 281: Could you just add a test like the following:
```c
	while (node >= 0) {
		node = fdt_node_offset_by_compatible(fdt, node, tz_sdp_match);
		if (node >= 0)
			DMSG("Ingore SDP pool node %s, supports only 1 node",
	}
```

core/mm/core_mmu.c
```diff
@@ -451,30 +501,57 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
 static bool pbuf_is_sdp_mem(paddr_t pbuf, size_t len)
 {
-	return pbuf_is_special_mem(pbuf, len, phys_sdp_mem_begin,
-				   phys_sdp_mem_end);
+	bool is_sdp_mem = false;
+
+	if (sec_sdp.size)
+		is_sdp_mem = core_is_buffer_inside(pbuf, len,
+						   sec_sdp.paddr,
+						   sec_sdp.size);
```
etienne-lms, line 509:
```c
		is_sdp_mem = core_is_buffer_inside(pbuf, len, sec_sdp.paddr,
						    sec_sdp.size);

File: core/mm/core_mmu.c [hunk 13]
```diff
@@ -373,6 +418,11 @@ void core_mmu_set_discovered_nsec_ddr(struct core_mmu_phys_mem *start,
 	 */
 
 #ifdef CFG_SECURE_DATA_PATH
+	if (dtb_get_sdp_region()) {
+		DMSG("region base %p %zu", (void *)sec_sdp.paddr, sec_sdp.size);
```
etienne-lms, line 422: prefer without a cast
```c
		DMSG("region base %#"PRIxPA" %zu", sec_sdp.paddr, sec_sdp.size);
```
I don't think it's the best place to print existence of the area.
It would be better either where found, from `dtb_get_sdp_region()` or no specific `DMSG()`.

core/mm/core_mmu.c
```diff
@@ -255,6 +259,47 @@ static struct tee_mmap_region *find_map_by_pa(unsigned long pa)
 	return NULL;
 }
 
+#if defined(CFG_SECURE_DATA_PATH)
+static bool dtb_get_sdp_region(void)
+{
+	void *fdt = NULL;
+	int node;
```
etienne-lms, line 266: `= 0;`  (must be initialized as per optee_os coding style)

core/mm/core_mmu.c
etienne-lms: ok as-is

core/mm/core_mmu.c
```diff
@@ -451,30 +501,57 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
+
+	return is_sdp_mem;
+}
+
+static struct mobj *core_sdp_mem_alloc_mobj(paddr_t pa, size_t size)
+{
+	struct mobj *mobj = NULL;
+
+	mobj = mobj_phys_alloc(pa, size,
+			       TEE_MATTR_MEM_TYPE_CACHED,
```
etienne-lms, line 523: prefer spread over less lines if possible (not a strong opinion, but more consistent with most optee_os implementation).
```c
	mobj = mobj_phys_alloc(pa, size, TEE_MATTR_MEM_TYPE_CACHED,
				CORE_MEM_SDP_MEM);
```
or 
```c
static struct mobj *core_sdp_mem_alloc_mobj(paddr_t pa, size_t size)
{
	struct mobj *mobj = mobj_phys_alloc(pa, size, TEE_MATTR_MEM_TYPE_CACHED,
					     CORE_MEM_SDP_MEM);
					    
	if (!mobj)
		panic("can't create SDP physical memory object");

	return mobj;			       
}
```

core/mm/core_mmu.c
```diff
@@ -451,30 +501,57 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
+			       CORE_MEM_SDP_MEM);
+	if (!mobj)
+		panic("can't create SDP physical memory object");
+
+	return mobj;
 }
 
 struct mobj **core_sdp_mem_create_mobjs(void)
 {
 	const struct core_mmu_phys_mem *mem;
```
etienne-lms, line 533: by the way, could you please initialize `const struct core_mmu_phys_mem *mem = NULL;`

core/mm/core_mmu.c
etienne-lms: Ok.
By the way, the loop header and body fits in single lines:
```c
	for (mem = phys_sdp_mem_begin; mem < phys_sdp_mem_end; mem++, mobj++)
		*mobj = core_sdp_mem_alloc_mobj(mem->addr, mem->size);
```

etienne-lms (commented):
nitpicking comments again (:confused:)

core/mm/core_mmu.c
omasse-linaro: ok let's remove this log.

core/mm/core_mmu.c
etienne-lms: my fault: this implementation scratches `node` and line 289 will fail. Need another temp node reference:
```c
	int n = 0;
	(...)

	n = node;
	while ((n = fdt_node_offset_by_compatible(fdt, n, tz_sdp_match) >= 0)
		DMSG("Ignoring SDP pool node %s", fdt_get_name(fdt, n, NULL))

File: core/mm/core_mmu.c [hunk 17]
```diff
@@ -451,30 +505,53 @@ static bool pbuf_is_nsec_ddr(paddr_t pbuf __unused, size_t len __unused)
...
+		is_sdp_mem = pbuf_is_special_mem(pbuf, len, phys_sdp_mem_begin,
+						 phys_sdp_mem_end);
+
+	return is_sdp_mem;
+}
+
+static struct mobj *core_sdp_mem_alloc_mobj(paddr_t pa, size_t size)
+{
+	struct mobj *mobj = mobj_phys_alloc(pa, size, TEE_MATTR_MEM_TYPE_CACHED,
+			       CORE_MEM_SDP_MEM);
```
etienne-lms, line 524: indentation:
 ```patch
 	struct mobj *mobj = mobj_phys_alloc(pa, size, TEE_MATTR_MEM_TYPE_CACHED,
-			       CORE_MEM_SDP_MEM);
+					     CORE_MEM_SDP_MEM);

core/mm/core_mmu.c
etienne-lms: ok

etienne-lms (changes requested):

core/mm/core_mmu.c
omasse-linaro: indeed, I did not notice it either

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "core: Define SDP in embedded DTB" with a last comment as a trace mentions `node` (implicitly node name?) but prints a compatible property string ID.

For that commit: I have another comment (for later). Compatible node are expected from `reserved-memory` node. I think the line 128 replace `node = fdt_node_offset_by_compatible(fdt, -1, tz_sdp_match)` with a call to a local helper function that finds `reserved-memory` nodes and parses sub-nodes calling `fdt_node_check_compatible()`. Such function could also check `status` property in case the node is disabled (despite the fact Linux kernel/u-boot don't test that).
The SDP active node uniqueness test should better also move to a local helper function possibly asserted (preventing unnecessary scans on release builds).

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "plat-hikey: Add embedded DTB to define SDP".

jforissier:
> TODO: Add a local helper function that finds reserved-memory nodes and

Please remove this comment. With that, for both commits:

etienne-lms (approved):

jforissier:
@omasse-linaro thanks!

PR #5334, opened 2022-05-13
scmi-msg: fixes in scmi base protocol

etienne-lms:
SCMI fixes from a issue reported in the LKML.


jforissier:
I suggest adding a link to the mentioned LKML discussion to the first commit, I think it should be:
```
Link: https://lore.kernel.org/lkml/1698297.NAKyZzlH2u@archbook
```

etienne-lms:
ok, i'll do.

etienne-lms:
tag applied

PR #5336, opened 2022-05-13
build: introduce CFG_OPTEE_REVISION_EXTRA

jforissier:
Adds CFG_OPTEE_REVISION_EXTRA (default: empty) which can be used to
append a custom string to the revision string shown in the boot banner.
A typical use case is build environments such as Yocto/OpenEmbedded
which check out a particular version of the optee_os repository and may
add patches on top. In this case the revision string is something like:

  3.17.0-dev (gcc version ...

which doesn't give any information on what modifications are added and
therefore makes it difficult to know for sure if a deployed binary is
indeed the expected one (even more so when the build date is fixed via
SOURCE_DATE_EPOCH for reproducible builds).
CFG_OPTEE_REVISION_EXTRA allows to append a specific build identifier.
For example:

  $ make CFG_OPTEE_REVISION_EXTRA=-mybuild_1234

would give:

  3.17.0-dev-mybuild_1234


etienne-lms (commented):

PR #5348, opened 2022-05-23
crypto_api: acipher: rsa: provide a software implementation

ldts:
Provide a software implementation via LTC to unsupported operations.

This functionality is required for platforms that only offer hardware support for a minimum set of RSA crypto operations.

### DIFF
--- a/core/drivers/crypto/crypto_api/acipher/rsa.c
+++ b/core/drivers/crypto/crypto_api/acipher/rsa.c
@@ -6,6 +6,7 @@
  */
 #include <drvcrypt.h>
 #include <crypto/crypto.h>
+#include <crypto/crypto_impl.h>
 #include <tee_api_defines_extensions.h>
 #include <tee/tee_cryp_utl.h>
 #include <utee_defines.h>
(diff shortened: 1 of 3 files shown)


ldts:
@jforissier normally I would have use the linker to wrap the LTC function names when building libtomcrypt but I dont know how to do it with these makefiles hence the SW(__x) macro....

jenswi-linaro:
I'm sorry, this is just a too ugly brute force approach. It will likely confuse cscope too. Could it be done with some clever attribute?

ldts:
I fully agree on he level of ugliness (was really testing the waters with the minimum set of changes I needed for the prototyping i am doing for Versal). 
I think we could have an interface and a separate CFG_ that enables it (ie a global crypto_acipher_rsa_sw_ops ? )

jforissier:
Not sure what it would look like with a `crypto_acipher_rsa_sw_ops`, feel free to try and propose. But it seems we could simply rename the software implementations (in LTC and MBedTLS) as `sw_crypto_*()` and when no hardware driver is configured simply define wrapper functions `crypto_rsa_*() { return sw_crypto_rsa_*(); }`, they could be `static inline` even to avoid an extra function call.

PR #5350, opened 2022-05-24
Add support to use external device tree and then map specifically device as secure

vesajaaskelainen:
We are using system level device tree where we use Xilinx Vivado tooling style device tree as a base (cleaned up version of that) .
This allows flexibility for our FPGA developers to change base addresses of devices more freely and then device trees just match that.

In order to load our Xiphera TRNG devices and our own drivers in OP-TEE this PR enabled support to load also devices from external device tree.

Other feature that this PR adds is possibility to map TrustZone aware IP core we have developed as secure in OP-TEE and then Linux will use it in non-secure. With standard device mapping logic device would be mapped as non-secure and OP-TEE would not allow the access.

### DIFF
--- a/core/kernel/dt.c
+++ b/core/kernel/dt.c
@@ -125,6 +125,39 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
 	return 0;
 }
 
+int dt_map_secure_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
+{
+	paddr_t pbase;
+	vaddr_t vbase;
+	size_t sz;
+	int st;
+
+	assert(cpu_mmu_enabled());
+
+	st = _fdt_get_status(fdt, offs);
+	if (st == DT_STATUS_DISABLED)
+		return -1;
+
+	pbase = _fdt_reg_base_address(fdt, offs);
+	if (pbase == DT_INFO_INVALID_REG)
+		return -1;
+	sz = _fdt_reg_size(fdt, offs);
+	if (sz == DT_INFO_INVALID_REG_SIZE)
+		return -1;
+
+	/* Check if we have a mapping, create one if needed */
+	vbase = (vaddr_t)core_mmu_add_mapping(MEM_AREA_IO_SEC, pbase, sz);
+	if (!vbase) {
+		EMSG("Failed to map %zu bytes at PA 0x%"PRIxPA,
+		     (size_t)sz, pbase);
+		return -1;
+	}
+
+	*base = vbase;
+	*size = sz;
+	return 0;
+}
+
 /* Read a physical address (n=1 or 2 cells) */
 static paddr_t _fdt_read_paddr(const uint32_t *cell, int n)
 {
--- a/core/include/kernel/dt.h
+++ b/core/include/kernel/dt.h
@@ -131,6 +131,20 @@ const struct dt_driver *dt_find_compatible_driver(const void *fdt, int offs);
 int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size);
 
 /*
+ * Map a device into secure memory and return the base VA and the mapping size.
+ * If the mapping already exists, the function simply returns the @base and
+ * @size information.
+ *
+ * @offs is the offset of the node that describes the device in @fdt.
+ * @base receives the base virtual address corresponding to the base physical
+ * address of the "reg" property
+ * @size receives the size of the mapping
+ *
+ * Returns 0 on success or -1 in case of error.
+ */
+int dt_map_secure_dev(const void *fdt, int offs, vaddr_t *base, size_t *size);
+
+/*
  * Check whether the node at @offs contains the property with propname or not.
  *
  * @offs is the offset of the node that describes the device in @fdt.
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -671,11 +671,9 @@ static TEE_Result probe_dt_drivers_early(void)
 	TEE_Result res = TEE_ERROR_GENERIC;
 	const void *fdt = NULL;
 
-	if (!IS_ENABLED(CFG_EMBED_DTB))
+	fdt = get_dt();
+	if (!fdt)
 		return TEE_SUCCESS;
-
-	fdt = get_embedded_dt();
-	assert(fdt);
 
 	parse_node(fdt, fdt_path_offset(fdt, "/"));
 
@@ -694,11 +692,9 @@ static TEE_Result probe_dt_drivers(void)
 	TEE_Result res = TEE_ERROR_GENERIC;
 	const void *fdt = NULL;
 
-	if (!IS_ENABLED(CFG_EMBED_DTB))
+	fdt = get_dt();
+	if (!fdt)
 		return TEE_SUCCESS;
-
-	fdt = get_embedded_dt();
-	assert(fdt);
 
 	res = process_probe_list(fdt);
 	if (res || !TAILQ_EMPTY(&dt_driver_failed_list)) {
@@ -720,12 +716,11 @@ static TEE_Result release_probe_lists(void)
 	struct dt_driver_probe *next = NULL;
 	struct dt_driver_provider *prov = NULL;
 	struct dt_driver_provider *next_prov = NULL;
-	const void * __maybe_unused fdt = NULL;
+	const void *fdt = NULL;
 
-	if (!IS_ENABLED(CFG_EMBED_DTB))
+	fdt = get_dt();
+	if (!fdt)
 		return TEE_SUCCESS;
-
-	fdt = get_embedded_dt();
 
 	assert(fdt && TAILQ_EMPTY(&dt_driver_probe_list));
 


core/kernel/dt.c
```diff
@@ -125,6 +125,39 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
 	return 0;
 }
 
+int dt_map_secure_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
+{
+	paddr_t pbase;
+	vaddr_t vbase;
+	size_t sz;
+	int st;
```
etienne-lms, line 133: `= 0;`

core/kernel/dt.c
```diff
@@ -125,6 +125,39 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
...
+	if (!vbase) {
+		EMSG("Failed to map %zu bytes at PA 0x%"PRIxPA,
+		     (size_t)sz, pbase);
+		return -1;
+	}
+
+	*base = vbase;
+	*size = sz;
+	return 0;
+}
```
etienne-lms, line 159: could be factorized with `dt_map_dev()`?

etienne-lms (commented):
I have a generic comment on this change. When external DTB is located in non-secure memory, it could be modified (e.g. from a physical external accesses) before OP-TEE initializes. In this case, is it safe to configure OP-TEE secure drivers using an unsafe DTB content?

jenswi-linaro:
> I have a generic comment on this change. When external DTB is located in non-secure memory, it could be modified (e.g. from a physical external accesses) before OP-TEE initializes. In this case, is it safe to configure OP-TEE secure drivers using an unsafe DTB content?

I share your concern. Even if security could be guaranteed on one system it could very well fail on another. I would seem more safe to keep the DTB in secure memory until secure world is done with it.

vesajaaskelainen:
> > I have a generic comment on this change. When external DTB is located in non-secure memory, it could be modified (e.g. from a physical external accesses) before OP-TEE initializes. In this case, is it safe to configure OP-TEE secure drivers using an unsafe DTB content?
> 
> I share your concern. Even if security could be guaranteed on one system it could very well fail on another. I would seem more safe to keep the DTB in secure memory until secure world is done with it.

In ARMv8 with ATF we have the boot flow which is also secure component in the flow.

With Xilinx ZynqMP SoC boot ROM verifies the FSBL/U-Boot SPL and then you either use Xilinx 'bootgen' based signing or FIT signing to load further components (like ATF, OP-TEE, device tree, u-boot).

One must trust all components before OP-TEE is loaded as they can adjust the system (secure) state (like enable debug interfaces or so).

With FPGA based devices DTB is more or less managed with the FPGA design tools like (Xilinx Vivado). They often provide tooling for device tree generation. As FPGA designs are more or less always proprietary this is not suitable to be shared in upstream project. In our case for DTB logistics it is though that it was clearest to minimize amount of device trees and use common device tree for U-Boot and OP-TEE as OP-TEE will then inject its own details there which is needed in u-boot.

If you feel that it is better to not to enable this kind of behavior by default then one option could be is to have new function for getting device tree (get_device_driver_dt?) and then new CFG option for allowing external device tree to be used for device driver loading.

Just a side note that similar external device tree support change is then needed in SCMI support with Xilinx Zynq-7000 improvements. With SCMI there are lots of other nodes that need to be loaded from common device tree as they are managed pretty directly from Xilinx Vivado/FPGA design.

core/kernel/dt.c
vesajaaskelainen: Problem with `dt_map_dev()` is that if device tree has both non-secure status and secure status enabled then device gets mapped as non-secure. And if it is mapped as non-secure then HW access for FPGA peripheral happens as non-secure thus OP-TEE cannot access thus TrustZone aware registers.

If we invent new settings into device tree then one could extend `dt_map_dev()`.

core/kernel/dt.c
vesajaaskelainen: Should we go to modify the `dt_map_dev()` too as this was more or less copy from there?

jenswi-linaro:
> > > I have a generic comment on this change. When external DTB is located in non-secure memory, it could be modified (e.g. from a physical external accesses) before OP-TEE initializes. In this case, is it safe to configure OP-TEE secure drivers using an unsafe DTB content?
> > 
> > 
> > I share your concern. Even if security could be guaranteed on one system it could very well fail on another. I would seem more safe to keep the DTB in secure memory until secure world is done with it.
> 
> In ARMv8 with ATF we have the boot flow which is also secure component in the flow.
> 
> With Xilinx ZynqMP SoC boot ROM verifies the FSBL/U-Boot SPL and then you either use Xilinx 'bootgen' based signing or FIT signing to load further components (like ATF, OP-TEE, device tree, u-boot).
> 
> One must trust all components before OP-TEE is loaded as they can adjust the system (secure) state (like enable debug interfaces or so).

I'm maybe a bit out of date here, but I thought that debugging could only be configured in one direction, from more open to less open. So if we want non-secure debug it must never be disabled.

I agree that the must trust the secure components, but for the non-secure components loaded before OP-TEE which can affect the security state I guess we should hope that they have done the right thing. But trusting them may be a bit much since they may have been tampered with using a debugger or something. If OP-TEE has nothing to do with those component we can turn a blind eye to that since it's not our job to secure all platforms in the world.

A reason I'd rather have the DTB in secure memory while OP-TEE is using it is to keep the interface to outside world a bit easier. The loader can very well copy the DTB into secure memory just before OP-TEE is about to use it, that's outside of OP-TEE's control. But if we use a non-secure DTB which is clearly non-secure as that's how we mapped it then there's a high risk that this pattern will be copied to other platforms.

> 
> With FPGA based devices DTB is more or less managed with the FPGA design tools like (Xilinx Vivado). They often provide tooling for device tree generation. As FPGA designs are more or less always proprietary this is not suitable to be shared in upstream project. In our case for DTB logistics it is though that it was clearest to minimize amount of device trees and use common device tree for U-Boot and OP-TEE as OP-TEE will then inject its own details there which is needed in u-boot.

I don't see a problem with copying the DTB into non-secure memory for non-secure consumption once OP-TEE is done with it.

> 
> If you feel that it is better to not to enable this kind of behavior by default then one option could be is to have new function for getting device tree (get_device_driver_dt?) and then new CFG option for allowing external device tree to be used for device driver loading.

That would add even more complexity to the jungle of DT CFG options we already have.

> 
> Just a side note that similar external device tree support change is then needed in SCMI support with Xilinx Zynq-7000 improvements. With SCMI there are lots of other nodes that need to be loaded from common device tree as they are managed pretty directly from Xilinx Vivado/FPGA design.

OK

vesajaaskelainen:
> > > > I have a generic comment on this change. When external DTB is located in non-secure memory, it could be modified (e.g. from a physical external accesses) before OP-TEE initializes. In this case, is it safe to configure OP-TEE secure drivers using an unsafe DTB content?
> > > 
> > > 
> > > I share your concern. Even if security could be guaranteed on one system it could very well fail on another. I would seem more safe to keep the DTB in secure memory until secure world is done with it.
> > 
> > 
> > In ARMv8 with ATF we have the boot flow which is also secure component in the flow.
> > With Xilinx ZynqMP SoC boot ROM verifies the FSBL/U-Boot SPL and then you either use Xilinx 'bootgen' based signing or FIT signing to load further components (like ATF, OP-TEE, device tree, u-boot).
> > One must trust all components before OP-TEE is loaded as they can adjust the system (secure) state (like enable debug interfaces or so).
> 
> I'm maybe a bit out of date here, but I thought that debugging could only be configured in one direction, from more open to less open. So if we want non-secure debug it must never be disabled.

There has been several CPU's that I have worked that has supported "soft disable" for JTAG. When you enable secure boot with eFuses debug peripherals are kept in reset/not enabled.

This can be handy especially when you are early in the enabling security features for your device.

How CPUs seem to work today is that after POR reset JTAG functionality is kept in reset/disabled until boot ROM (or similar) goes and enables those -- and in secure boot this never happens. Some of them have additional lock bits that boot ROM sets so you cannot go to play around with the mechanisms later on.

Most of today's CPU's (all?) also have kill switch eFuse for JTAG functionality so you may want to consider that for release quality devices.

> I agree that the must trust the secure components, but for the non-secure components loaded before OP-TEE which can affect the security state I guess we should hope that they have done the right thing. But trusting them may be a bit much since they may have been tampered with using a debugger or something. If OP-TEE has nothing to do with those component we can turn a blind eye to that since it's not our job to secure all platforms in the world.
> 
> A reason I'd rather have the DTB in secure memory while OP-TEE is using it is to keep the interface to outside world a bit easier. The loader can very well copy the DTB into secure memory just before OP-TEE is about to use it, that's outside of OP-TEE's control. But if we use a non-secure DTB which is clearly non-secure as that's how we mapped it then there's a high risk that this pattern will be copied to other platforms.

In this particular case DTB is also modified by OP-TEE which is then used by u-boot. This actual device tree also came from FPGA design tools -- so one could consider that as non-trusted -- but then again you just have to trust FPGA developer (and Xilinx) doing the right thing as that has the ultimate control on how the device behaves.

Xilinx has even defined this magic address (https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-zynqmp/conf.mk#L40) where device tree is supposed to be loaded.

I suppose alternative for this would be then to build new OP-TEE only device tree and then inject it during the build time into OP-TEE OS and then use that for OP-TEE's device drivers. This is a bit problematic from logistics point of view but doable.

Then later on OP-TEE would access the device tree located in Xilinx reserved memory address and inject OP-TEE's "handoff" information for the next stages?

> > With FPGA based devices DTB is more or less managed with the FPGA design tools like (Xilinx Vivado). They often provide tooling for device tree generation. As FPGA designs are more or less always proprietary this is not suitable to be shared in upstream project. In our case for DTB logistics it is though that it was clearest to minimize amount of device trees and use common device tree for U-Boot and OP-TEE as OP-TEE will then inject its own details there which is needed in u-boot.
> 
> I don't see a problem with copying the DTB into non-secure memory for non-secure consumption once OP-TEE is done with it.

Catch there is that device tree is pretty complex -- also meaning that it is large. 

So I suppose there would then need to be following set of device trees:

- U-Boot DTB (what ever u-boot needs) -- loaded at 0x100000 by FSBL/SPL
- OP-TEE DTB (very small set of what is needed) -- included within OP-TEE OS binary
- Kernel DTB -- loaded from FIT image and modified by u-boot

U-boot plays around with `fdtgrep` and such tools to filter out device tree's for SPL. perhaps that is something to try out -- but I suppose that would be out side of OP-TEE OS build?

What has been nice with SPL is that we have had a code that checks HWID resistors/eFuse what board revision it is and then selects device tree based on that one. Now I suppose we are missing such thing in OP-TEE? Don't probably need that for Zynq MPSoC as only changing part would be FPGA bitstream and that is kinda implicit within the image (as is psu init code generated by Xilinx tools).

> > If you feel that it is better to not to enable this kind of behavior by default then one option could be is to have new function for getting device tree (get_device_driver_dt?) and then new CFG option for allowing external device tree to be used for device driver loading.
> 
> That would add even more complexity to the jungle of DT CFG options we already have.

OK.

jenswi-linaro:
> > A reason I'd rather have the DTB in secure memory while OP-TEE is using it is to keep the interface to outside world a bit easier. The loader can very well copy the DTB into secure memory just before OP-TEE is about to use it, that's outside of OP-TEE's control. But if we use a non-secure DTB which is clearly non-secure as that's how we mapped it then there's a high risk that this pattern will be copied to other platforms.
> 
> In this particular case DTB is also modified by OP-TEE which is then used by u-boot. This actual device tree also came from FPGA design tools -- so one could consider that as non-trusted -- but then again you just have to trust FPGA developer (and Xilinx) doing the right thing as that has the ultimate control on how the device behaves.
> 
> Xilinx has even defined this magic address (https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-zynqmp/conf.mk#L40) where device tree is supposed to be loaded.
> 
> I suppose alternative for this would be then to build new OP-TEE only device tree and then inject it during the build time into OP-TEE OS and then use that for OP-TEE's device drivers. This is a bit problematic from logistics point of view but doable.
> 
> Then later on OP-TEE would access the device tree located in Xilinx reserved memory address and inject OP-TEE's "handoff" information for the next stages?

This is more or less what we're doing today on some platforms if I recall correctly.

We'd like to be able to replace this embedded DTB with one supplied in secure memory instead, to among things make logistics a bit easier.

Is there a problem with letting the OP-TEE loader copy the DTB into secure memory and then once OP-TEE returns after reading and possibly updating the DTB it's copied into non-secure memory for later boot stages?

It seems that there are platforms capable of delivering a DTB in non-secure memory while still remaining as secure as if the DTB would have resided in secure memory. But I doubt that we within OP-TEE can find a generic solution along this that is secure on all platforms. That's why I'd like to keep it simple from OP-TEE point of view, the fewer different ways of doing something we have the easier it will be to get it right even if some platform could have used a more direct approach.

It could be and has been debated if OP-TEE should use the same DTB as the normal world and especially the kernel is using. We should try to use an approach in OP-TEE that works in both cases without special tweaks within OP-TEE.

vesajaaskelainen:
> > > A reason I'd rather have the DTB in secure memory while OP-TEE is using it is to keep the interface to outside world a bit easier. The loader can very well copy the DTB into secure memory just before OP-TEE is about to use it, that's outside of OP-TEE's control. But if we use a non-secure DTB which is clearly non-secure as that's how we mapped it then there's a high risk that this pattern will be copied to other platforms.
> > 
> > 
> > In this particular case DTB is also modified by OP-TEE which is then used by u-boot. This actual device tree also came from FPGA design tools -- so one could consider that as non-trusted -- but then again you just have to trust FPGA developer (and Xilinx) doing the right thing as that has the ultimate control on how the device behaves.
> > Xilinx has even defined this magic address (https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-zynqmp/conf.mk#L40) where device tree is supposed to be loaded.
> > I suppose alternative for this would be then to build new OP-TEE only device tree and then inject it during the build time into OP-TEE OS and then use that for OP-TEE's device drivers. This is a bit problematic from logistics point of view but doable.
> > Then later on OP-TEE would access the device tree located in Xilinx reserved memory address and inject OP-TEE's "handoff" information for the next stages?
> 
> This is more or less what we're doing today on some platforms if I recall correctly.

OK

> We'd like to be able to replace this embedded DTB with one supplied in secure memory instead, to among things make logistics a bit easier.
> 
> Is there a problem with letting the OP-TEE loader copy the DTB into secure memory and then once OP-TEE returns after reading and possibly updating the DTB it's copied into non-secure memory for later boot stages?

I do not see a problem of copying DTB into OP-TEE's secure memory as long as memory allocator can be used at that stage. For device driver initialization only concern is the size of the data to copy -- but in theory that is in the DTB header (`totalsize`) and then we can have some sanity value to check against (eg. `CFG_DTB_MAX_SIZE`).

Then we can just have original reference for non-secure DTB where OP-TEE will eventually inject handoff information.

I can try this if it sounds as a good path?

This model has added value in example if there would be DDR DIMM that's size is auto detected from its SPD EEPROM by FSBL/SPL and then updated into device tree during the boot.

> It seems that there are platforms capable of delivering a DTB in non-secure memory while still remaining as secure as if the DTB would have resided in secure memory. But I doubt that we within OP-TEE can find a generic solution along this that is secure on all platforms. That's why I'd like to keep it simple from OP-TEE point of view, the fewer different ways of doing something we have the easier it will be to get it right even if some platform could have used a more direct approach.

Concept of secure and non-secure memory is a bit moot during the boot process as previous processes are executing in secure state. ATF or OP-TEE is often the ones that first setup the security firewalls and loader is executed before those.

If you would have large amount of OCM memory then that would be best but can be challenging to fit the ATF, OP-TEE's pager and additional data in structured way (fixed addresses) in there for boot to succeed. So at some point of time one needs to start to resort in DDR. In here OP-TEE pager + "swapped out" pages encrypted in DDR could be a solution (unless there is HW accelerated solution for that) -- but that is available **after** OP-TEE OS kernel has started.

In theory DDR encryption could be bootstrapped earlier but that would need to be coordinated from start of FSBL/SPL being executed within OCM. Loading on top of FSBL/SPL while it is executing is a bit troublesome too. Bootstrapping with other AUX core could be an option where available.

> It could be and has been debated if OP-TEE should use the same DTB as the normal world and especially the kernel is using. We should try to use an approach in OP-TEE that works in both cases without special tweaks within OP-TEE.

OK

vesajaaskelainen:
keep-alive

vesajaaskelainen:
Getting a headache on trying to understand discussion ;) ... so I draw a picture to make it easier to discuss. And also tried to summarize the information so far.

To make it more concrete I have drawn USB boot case in here.

```mermaid
sequenceDiagram
    participant PC
    participant ROM as CSU-ROM
    participant FSBL
    participant ATF
    participant DTB
    participant OPTEE as OP-TEE
    participant UBoot as U-Boot
    participant OCM
    participant PL as FPGA (PL)
    participant DDR

    Note right of PC: Target is configured for USB boot
    Note right of PC: User powers the target
    activate PMU
    PMU->>PMU: Execute from PMU-ROM
    activate ROM
    ROM->>ROM: Start USB-DFU
    PC->>ROM: Sending Xilinx Image with FSBL
    ROM-->>OCM: Store image
    ROM->>ROM: Verify image
    ROM->>ROM: Decrypt image
    ROM->>PC: Done
    ROM-->>OCM: Execute code from OCM (with Cortex-A53 #35;0)
    OCM-->>FSBL: Start from FSBL's entry point
    deactivate ROM
    activate FSBL
    Note right of FSBL: FSBL is considered secure
    FSBL-->>DDR: Configure DDR controller
    FSBL->>PMU: Send PMUCFG with some HW "isolation" firewalls
    FSBL->>FSBL: Start USB-DFU
    PC->>FSBL: Sending Xilinx Image with PMU,Bitstram,ATF,DTB,OP-TEE,U-Boot
    FSBL-->>DDR: Store image
    FSBL->>FSBL: Verify image
    Note right of FSBL: Decrypt image on-the-go
    FSBL->>PMU: Program PMUFW image and start
    PMU->>PMU: Start PMUFW
    Note right of PMU: PMU+PMUFW must be considered secure due to its access level
    FSBL->>+PMU: Program bitstream image and activate
    PMU-->>PL: Program bitstream image and activate
    activate PL
    PL-->>PL: Activate
    Note right of PL: FPGA can be both secure and in-secure. Design & HW firewalls dictate in here.
    PMU->>-FSBL: Done
    FSBL-->>DDR: Store ATF image (0x1000..0x7ffff)
    FSBL-->>DDR: Store OP-TEE image (0x60000000)
    FSBL-->>DDR: Store DTB image (0x100000..0x2fffff)
    FSBL-->>DDR: Store U-Boot image (0x08000000)
    FSBL->>PC: Done
    FSBL-->>DDR: Execute code from DDR (ATF)
    DDR-->>ATF: Start from ATF's entry point
    deactivate FSBL
    activate ATF
    Note right of ATF: ATF is considered secure
    ATF-->>DDR: Fetch DTB (0x100000)
    ATF->>ATF: Configure system
    Note right of ATF: Now some more firewalls are set up
    ATF->>ATF: Configure secure monitor
    ATF->>ATF: Inject entries into device tree
    ATF-->>DDR: Apply DTB modifications (0x100000)
    ATF->>OPTEE: Start from OP-TEE's entry point
    activate OPTEE
    Note right of OPTEE: OP-TEE is considered secure
    OPTEE-->>DDR: Fetch DTB (0x100000)
    OPTEE->>OPTEE: Load drivers based on DTB
    OPTEE->>OPTEE: Inject entries into device tree
    OPTEE-->>DDR: Apply DTB modifications (0x100000)
    OPTEE->>ATF: Return
    ATF-->>UBoot: Start from U-Boot's entry point
    activate UBoot
    Note right of UBoot: This is first non-secure code being executed
    UBoot-->>DDR: Fetch DTB (0x100000)
    UBoot->>UBoot: Load drivers based on DTB
    UBoot->>UBoot: Start USB-Fastboot (or so)
    Note right of UBoot: And receive FIT image with kernel, ramdisk, kernel dtb
    deactivate PL
    deactivate PMU
    deactivate UBoot
    deactivate OPTEE
    deactivate ATF
```

> Special case for FPGA based devices is that device configuration can be changed with FPGA design tools. Eg. change HW isolation configuration or add new IP cores in custom addresses.

> Note to self: study more how those DDR memory + HW firewalls are configured.

My understanding is that the problem here is what we consider DDR memory for DTB to be? Is it safe for OP-TEE to use it?

In ATF we have (plat/xilinx/zynqmp/bl31_zynqmp_setup.c :: bl31_plat_arch_setup()):
```c
        const mmap_region_t bl_regions[] = {
#if (BL31_LIMIT < PLAT_DDR_LOWMEM_MAX)
                MAP_REGION_FLAT(XILINX_OF_BOARD_DTB_ADDR, XILINX_OF_BOARD_DTB_MAX_SIZE,
                        MT_MEMORY | MT_RW | MT_NS),
#endif
                MAP_REGION_FLAT(BL31_BASE, BL31_END - BL31_BASE,
                        MT_MEMORY | MT_RW | MT_SECURE),
                MAP_REGION_FLAT(BL_CODE_BASE, BL_CODE_END - BL_CODE_BASE,
                                MT_CODE | MT_SECURE),
                MAP_REGION_FLAT(BL_RO_DATA_BASE, BL_RO_DATA_END - BL_RO_DATA_BASE,
                                MT_RO_DATA | MT_SECURE),
                MAP_REGION_FLAT(BL_COHERENT_RAM_BASE,
                                BL_COHERENT_RAM_END - BL_COHERENT_RAM_BASE,
                                MT_DEVICE | MT_RW | MT_SECURE),
                {0}
        };
```

Almost like there is something missing... BL32 setup?

Which makes the DTB memory as non-secure which it should be so that `U-Boot` can then later access it.

Now basic problem here is that U-Boot must be aware of all previous DTB modifications done by `ATF` and by `OP-TEE` so that it can later on modify `kernel's DTB`.

Now if the memory is not shared between parties `ATF` - `OP-TEE` - `U-Boot` it makes life a bit harder.

Now from injection point of view as long as OP-TEE does not otherwise utilize it for its operations I suppose this is not a problem?

If OP-TEE would use it to load drivers for itself it becomes as a problem?

Now as more or less all the code is executed on this platform from DDR only protection we have is CPU's internal protections. Eg. if someone goes to DDR bus itself and modifies the contents then we cannot do much to protect against that (expect on HW design to make it harder to access the DDR bus).

OCM size here is 256 KiB which sounds fine but we have a bit chicken egg problem in here. With OP-TEE enabled ATF it consumes 100 KiB. which would otherwise fit in OCM memory but it is at that point of time consumed by FSBL image which is being executed from there (it is a bit fat).

There seem to have been some idea for ATF to reside in OCM if there is no secure payloads present (would take 88 KiB) but we are 12 KiB over that. And that would also require quite fit FSBL binary.

If we would go to execute code from DDR either by FSBL trampoline then we could load ATF directly into OCM and be happy about that part but then we have used time in DDR memory which can be considered dangerous.

Then the next question here would be where OP-TEE's OS would be stored. Now it is also in DDR and there is hefty amount of memory reserved for that.

In theory in this platform there is `Cortex-R5F` where `CSU` can also transition execution into during the boot flow. But OCM is still the place where the binary is loaded. There is TCM memory (64, 128 or in 256 KiB configurations) for R5F's which could be used as a trampoline to make OCM free when loading more code when being executed from TCM. But if your product has other plans for R5F's then this can become a bit problematic.

---

I'll try to summarize our options in here:

a) Have embedded DTB in OP-TEE

Embedded DTB would be used to configure OP-TEE's drivers while external DTB would be used to inject OP-TEE's part back into it.

U-Boot would have fresh information in DTB.

Properties: All of these would be stored in DDR memory.

Pros:
- Not much need to be changed in OP-TEE -- existing functionality

Cons:
- Part of the devices are within FPGA that we want to load (like TRNG and our own IP) and configuration can change for that (eg. base addresses).
- There is extra device tree / product to maintain manually.

b) Utilize DTB from shared location

FSBL would load DTB from Xilinx Image container into specified address (0x100000).

ATF would then go and inject its own information into that.

OP-TEE would then use it also for its own drivers and also inject own information into that.

U-Boot would have fresh information in DTB.

Properties: All of these would be stored in DDR memory.

Pros:
- Only one or two DTBs to be maintained (system DTB and kernel DTB). Don't know if it is feasible to have them as same.
- Quite small code change
- Seems to be way Xilinx has designed it to be

Cons:
- OP-TEE would trust information in DTB for driver loading
  - Do we need to copy DTB in this case to OP-TEE memory side so that if there are references they are for secure side?
  - Seem to work without copy today

c) Make OP-TEE copy DTB into non-secure location

Transition DTB to be loaded in memory location that is considered secured.
- Change ATF so that this new secure memory location is considered secure (and keep old as non-secure?)
- Change memory address in ATF for DTB
- Change OP-TEE to load drivers from that location (similar to this PR)

Add new operation in OP-TEE to copy DTB into non-secure location (old 0x100000?).
- This is needed so that we get ATF's injected information carried over
- Set this non-secure location as external DTB so that OP-TEE would inject details into it

U-Boot would look from old DTB location and continue the boot flow.

Properties: All of these would be stored in DDR memory.

Pros:
- Any?

Cons:
- More changes in OP-TEE to also perform DTB copying
- Still need changes like in this PR for driver loading
- Changes in ATF that would need to be conditional whether OP-TEE is booted or not so that ATF injects its information in proper locations

d) Your option in here :)

At least in this platform it seems that we must trust DDR contents. There just ain't enough internal OCM memory for consumption even thou there is "plenty of it".

---

To me at least the change proposed in this PR seems to be good compromise for platform where there isn't enough OCM to keep OP-TEE (and friends) in OCM.

core/kernel/dt.c
vesajaaskelainen: Looking at the file more. It is full of missing variable initializations. Thou there does not seem to be problematic cases.

Should I make fix it all commit before adding a new one? Or shall we continue with existing style within the file?

vesajaaskelainen:
@jforissier re-open the PR to continue the discussion?

jforissier:
@vesajaaskelainen for whatever reason I can't re-open it...

vesajaaskelainen:
> @vesajaaskelainen for whatever reason I can't re-open it...

OK. Splitted the PR to two new PRs.

PR #5352, opened 2022-05-24
Fixes for the SNVS driver

clementfaure:
Hello,

Here are two fixes for silly mistakes in the SNVS driver.
These two problems have been found after testing on a 8mscale closed platform with OTP selected as a master key.


clementfaure:
Thanks for the quick review @jforissier 
Tags applied

PR #5355, opened 2022-05-26
SP: Add dump state

jellesels-arm:
Add dump state for SPs. This will make it possible for the symbolize
script to print SP call stack on a panic,

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -807,6 +807,20 @@ static bool sp_handle_svc(struct thread_svc_regs *regs)
 	return false;
 }
 
+static void sp_dump_state(struct ts_ctx *ctx)
+{
+	struct sp_ctx *utc = to_sp_ctx(ctx);
+
+	if (utc->uctx.dump_entry_func) {
+		TEE_Result res = ldelf_dump_state(&utc->uctx);
+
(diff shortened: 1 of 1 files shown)


core/arch/arm/kernel/secure_partition.c
```diff
@@ -807,13 +807,36 @@ static bool sp_handle_svc(struct thread_svc_regs *regs)
...
+{
+	struct sp_ctx *utc = to_sp_ctx(ctx);
+
+	if (utc->uctx.dump_entry_func) {
+		TEE_Result res = ldelf_dump_state(&utc->uctx);
+
+		if (!res || res == TEE_ERROR_TARGET_DEAD)
+			return;
+		/*
+		 * Fall back to dump_state_no_ldelf_dbg() if
```
jforissier, line 820: ...except `dump_state_no_ldelf_dbg()` doesn't exist here.

jforissier (commented):
One comment, otherwise:

jenswi-linaro:
Please add a `core:` prefix to to the subject line in the commit message. With that and Jerome's comment addressed please apply:

core/arch/arm/kernel/secure_partition.c
jellesels-arm: Done

jforissier:
Merged.

PR #5359, opened 2022-05-31
FF-A and virtualization

jenswi-linaro:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,6 +130,7 @@ jobs:
           _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SEL1_SPMC=y
           _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SEL2_SPMC=y CFG_CORE_PHYS_RELOCATABLE=y CFG_TZDRAM_START=0x0d304000 CFG_TZDRAM_SIZE=0x00cfc000
           _make PLATFORM=vexpress-qemu_armv8a CFG_{ATTESTATION,DEVICE_ENUM,RTC,SCMI,SECSTOR_TA_MGT}_PTA=y CFG_WITH_STATS=y CFG_TA_STATS=y
+          _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SEL1_SPMC=y CFG_NS_VIRTUALIZATION=y
           dd if=/dev/urandom of=BL32_AP_MM.fd bs=2621440 count=1 && _make PLATFORM=vexpress-qemu_armv8a CFG_STMM_PATH=BL32_AP_MM.fd CFG_RPMB_FS=y CFG_CORE_HEAP_SIZE=524288 CFG_TEE_RAM_VA_SIZE=0x00400000
           if [ -d $HOME/scp-firmware ]; then _make PLATFORM=vexpress-qemu_armv8a CFG_SCMI_SCPFW=y CFG_SCP_FIRMWARE=$HOME/scp-firmware; fi
           _make PLATFORM=stm-b2260
--- a/core/arch/arm/include/ffa.h
+++ b/core/arch/arm/include/ffa.h
@@ -169,9 +169,17 @@
 #define FFA_PART_PROP_NOTIF_DESTROYED	BIT(7)
 #define FFA_PART_PROP_AARCH64_STATE	BIT(8)
 
+#define FFA_MEMORY_HANDLE_HYPERVISOR_BIT	BIT64(63)
+#define FFA_MEMORY_HANDLE_SECURE_BIT		BIT64(45)
+#define FFA_MEMORY_HANDLE_NON_SECURE_BIT	BIT64(44)
+/*
+ * Codes the OP-TEE partition/guest ID into a cookie in order to know which
+ * partition to activate when reclaiming the shared memory. This field is 0
+ * unless CFG_NS_VIRTUALIZATION is enabled.
+ */
+#define FFA_MEMORY_HANDLE_PRTN_SHIFT		16
+#define FFA_MEMORY_HANDLE_PRTN_MASK		GENMASK_32(16, 0)
 
-#define FFA_MEMORY_HANDLE_SECURE_BIT	BIT64(45)
-#define FFA_MEMORY_HANDLE_NONE_SECURE_BIT	BIT64(44)
 
 #define FFA_BOOT_INFO_NAME_LEN		U(16)
 
--- a/core/arch/arm/include/ffa.h
+++ b/core/arch/arm/include/ffa.h
@@ -28,12 +28,15 @@
 #define FFA_VERSION_MAJOR		U(1)
 #define FFA_VERSION_MAJOR_SHIFT		U(16)
 #define FFA_VERSION_MAJOR_MASK		U(0x7FFF)
-#define FFA_VERSION_MINOR		U(0)
+#define FFA_VERSION_MINOR		U(1)
 #define FFA_VERSION_MINOR_SHIFT		U(0)
 #define FFA_VERSION_MINOR_MASK		U(0xFFFF)
 #define MAKE_FFA_VERSION(major, minor)	\
 	((((major) & FFA_VERSION_MAJOR_MASK) << FFA_VERSION_MAJOR_SHIFT) | \
 	 ((minor) & FFA_VERSION_MINOR_MASK))
+
+#define FFA_VERSION_1_0			MAKE_FFA_VERSION(1, 0)
+#define FFA_VERSION_1_1			MAKE_FFA_VERSION(1, 1)
 
 /* Function IDs */
 #define FFA_ERROR			U(0x84000060)
@@ -48,9 +51,11 @@
 #define FFA_RXTX_UNMAP			U(0x84000067)
 #define FFA_PARTITION_INFO_GET		U(0x84000068)
 #define FFA_ID_GET			U(0x84000069)
+#define FFA_SPM_ID_GET			U(0x84000085)
 #define FFA_MSG_WAIT			U(0x8400006B)
 #define FFA_MSG_YIELD			U(0x8400006C)
 #define FFA_MSG_RUN			U(0x8400006D)
+#define FFA_MSG_SEND2			U(0x84000086)
 #define FFA_MSG_SEND			U(0x8400006E)
 #define FFA_MSG_SEND_DIRECT_REQ_32	U(0x8400006F)
 #define FFA_MSG_SEND_DIRECT_REQ_64	U(0xC400006F)
@@ -70,14 +75,42 @@
 #define FFA_MEM_RECLAIM			U(0x84000077)
 #define FFA_MEM_FRAG_RX			U(0x8400007A)
 #define FFA_MEM_FRAG_TX			U(0x8400007B)
+#define FFA_NOTIFICATION_BITMAP_CREATE	U(0x8400007D)
+#define FFA_NOTIFICATION_BITMAP_DESTROY	U(0x8400007E)
+#define FFA_NOTIFICATION_BIND		U(0x8400007F)
+#define FFA_NOTIFICATION_UNBIND		U(0x84000080)
+#define FFA_NOTIFICATION_SET		U(0x84000081)
+#define FFA_NOTIFICATION_GET		U(0x84000082)
+#define FFA_NOTIFICATION_INFO_GET_32	U(0x84000083)
+#define FFA_NOTIFICATION_INFO_GET_64	U(0xC4000083)
 #define FFA_SECONDARY_EP_REGISTER_64	U(0xC4000087)
 #define FFA_MEM_PERM_GET_32		U(0x84000088)
 #define FFA_MEM_PERM_GET_64		U(0xC4000088)
 #define FFA_MEM_PERM_SET_32		U(0x84000089)
 #define FFA_MEM_PERM_SET_64		U(0xC4000089)
 
+#define FFA_FEATURE_NOTIF_PEND_INTR	U(0x1)
+#define FFA_FEATURE_SCHEDULE_RECV_INTR	U(0x2)
+#define FFA_FEATURE_MANAGED_EXIT_INTR	U(0x3)
+
 /* Special value for traffic targeted to the Hypervisor or SPM */
 #define FFA_TARGET_INFO_MBZ		U(0x0)
+
+#define FFA_MSG_FLAG_FRAMEWORK		BIT(31)
+#define FFA_MSG_TYPE_MASK		GENMASK_32(7, 0)
+#define FFA_MSG_PSCI			U(0x0)
+#define FFA_MSG_SEND_VM_CREATED		U(0x4)
+#define FFA_MSG_RESP_VM_CREATED		U(0x5)
+#define FFA_MSG_SEND_VM_DESTROYED	U(0x6)
+#define FFA_MSG_RESP_VM_DESTROYED	U(0x7)
+#define FFA_MSG_VERSION_REQ		U(0x8)
+#define FFA_MSG_VERSION_RESP		U(0x9)
+
+/*
+ * Flag used as parameter to FFA_PARTITION_INFO_GET to return partition
+ * count only.
+ */
+#define FFA_PARTITION_INFO_GET_COUNT_FLAG	BIT(0)
 
 /* Memory attributes: Normal memory, Write-Back cacheable, Inner shareable */
 #define FFA_NORMAL_MEM_REG_ATTR		U(0x2f)
@@ -120,12 +153,22 @@
  * Flags used for the FFA_PARTITION_INFO_GET return message:
  * BIT(0): Supports receipt of direct requests
  * BIT(1): Can send direct requests
- * BIT(2): Cannot send and receive indirect messages
- * BIT(3): Does not support receipt of notifications
+ * BIT(2): Can send and receive indirect messages
+ * BIT(3): Supports receipt of notifications
  * BIT(4-5): Partition ID is a PE endpoint ID
  */
-#define FFA_PARTITION_DIRECT_REQ_RECV_SUPPORT BIT(0)
-#define FFA_PARTITION_DIRECT_REQ_SEND_SUPPORT BIT(1)
+#define FFA_PART_PROP_DIRECT_REQ_RECV	BIT(0)
+#define FFA_PART_PROP_DIRECT_REQ_SEND	BIT(1)
+#define FFA_PART_PROP_INDIRECT_MSGS	BIT(2)
+#define FFA_PART_PROP_RECV_NOTIF	BIT(3)
+#define FFA_PART_PROP_IS_PE_ID		SHIFT_U32(0, 4)
+#define FFA_PART_PROP_IS_SEPID_INDEP	SHIFT_U32(1, 4)
+#define FFA_PART_PROP_IS_SEPID_DEP	SHIFT_U32(2, 4)
+#define FFA_PART_PROP_IS_AUX_ID		SHIFT_U32(3, 4)
+#define FFA_PART_PROP_NOTIF_CREATED	BIT(6)
+#define FFA_PART_PROP_NOTIF_DESTROYED	BIT(7)
+#define FFA_PART_PROP_AARCH64_STATE	BIT(8)
+
 
 #define FFA_MEMORY_HANDLE_SECURE_BIT	BIT64(45)
 #define FFA_MEMORY_HANDLE_NONE_SECURE_BIT	BIT64(44)
@@ -183,7 +226,7 @@ struct ffa_mem_access {
 };
 
 /* Lend, donate or share memory transaction descriptor */
-struct ffa_mem_transaction {
+struct ffa_mem_transaction_1_0 {
 	uint16_t sender_id;
 	uint8_t mem_reg_attr;
 	uint8_t reserved0;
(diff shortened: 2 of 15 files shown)


jenswi-linaro:
Can be tested as described in https://github.com/OP-TEE/build/pull/570

core/arch/arm/kernel/entry_a64.S
```diff
@@ -128,18 +128,17 @@ FUNC _start , :
 #if defined(CFG_CORE_SEL1_SPMC)
 	/*
 	 * With OP-TEE as SPMC at S-EL1 the SPMD (SPD_spmd) in TF-A passes
-	 * the DTB in x0, pagaeble part in x1 and the rest of the registers
-	 * are unused
+	 * the SPMC manifest in x0, the pagaeble part in x1, the system DTB
+	 * in x2 and the rest of the registers are unused.
 	 */
 	mov	x19, x1		/* Save pagable part */
-	mov	x20, x0		/* Save DT address */
```
balint-dobszay-arm, line 135: Are we dropping the value passed in `x0`, or is it saved somewhere else?

jenswi-linaro: We are dropping it since we currently don't look at the SPMC manifest.
If or when we need it we should save it.

balint-dobszay-arm: Looking more into this, I think there might be some breakage here.

- The SPMD by default was always passing the SPMC manifest (`TOS_FW_CONFIG`) in `x0`, at least on FVP. There might be some qemu specific code overriding this (I'm not sure?) but AFAICS the code for this is not platform specific [1] [2].

- OP-TEE saves this to `x20` at this current line, and later (mis)interprets this as the system dtb (or `HW_CONFIG` in TF-A's terminology) [3].

- On FVP platform the SPMD passes the Event Log buffer address in the SPMC manifest (`TOS_FW_CONFIG`) [4].

- OP-TEE is looking for the Event Log address in `HW_CONFIG`, but it's actually passed in the SPMC manifest, but OP-TEE misinterprets the SPMC manifest as the `HW_CONFIG`. So luckily we end up at the same place, the event log is correctly found, and our use case of getting the event log (on FVP, with OP-TEE as S-EL1 SPMC) is working correctly [5].

This is my understanding currently, does this make any sense?
By dropping `x0` we break this use case, so either the SPMD should pass the Event Log in `HW_CONFIG` or OP-TEE should look for the Event Log in the SPMC manifest, not sure which one is better.

[1]: https://github.com/ARM-software/arm-trusted-firmware/blob/master/common/desc_image_load.c#L249
[2]: https://trustedfirmware-a.readthedocs.io/en/latest/components/secure-partition-manager.html#spmc-boot
[3]: https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/boot.c#L1300
[4]: https://github.com/ARM-software/arm-trusted-firmware/blob/master/plat/arm/board/fvp/fvp_bl2_measured_boot.c#L212
[5]: https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/secure_partition.c#L581

jenswi-linaro: Thanks for looking into this. I'd like to define a common register usage for OP-TEE with SPMC at S-EL1. This thing with getting it right by mistake should also be resolved. We could if it helps break out all commits needed for FF-A 1.1 and take those in a separate PR.

https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/15339 should be updated to follow what we agree on. Thanks for the link to `populate_next_bl_params_config()`, I should use that one on QEMU too.

balint-dobszay-arm: > I'd like to define a common register usage for OP-TEE with SPMC at S-EL1.

If OP-TEE is the S-EL1 SPMC, I think the current solution on FVP (`x0`: SPMC manifest, `x1`: HW_CONFIG) sounds reasonable. Is it feasible for QEMU to align with this? I'm not familiar with that platform.

> This thing with getting it right by mistake should also be resolved.

I plan to open a PR soon (next 1-2 weeks) which should address this too.

> We could if it helps break out all commits needed for FF-A 1.1 and take those in a separate PR.

Sounds good to me, thanks.

jforissier:
- "core: add virt_get_current_guest_id()"
-  "core: add memory type MEM_AREA_NEX_NSEC_SHM"
"core: register spmc_init() with boot_final()"
- "ci: azure: build CFG_CORE_SEL1_SPMC=y CFG_VIRTUALIZATION=y"
When changed to GitHub Actions and `s/azure: //`:

jforissier:
@jenswi-linaro I am re-opening this since I suppose it got stale mainly because higher priority stuff came in.

jenswi-linaro:
Thanks.

jforissier:
- For commits:
  "core: add virt_get_current_guest_id()"
  "core: add memory type MEM_AREA_NEX_NSEC_SHM":

- For "core: register spmc_init() with boot_final()" :

jenswi-linaro:
Rebased and tags applied.

jenswi-linaro:
Rebased and added "core: arm64: use adr_l for __nex_bss_start and __nex_bss_end"

.github/workflows/ci.yml
```diff
@@ -127,6 +127,7 @@ jobs:
           _make PLATFORM=vexpress-qemu_armv8a CFG_NS_VIRTUALIZATION=y
           _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_PREALLOC_EL0_TBLS=y
           _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SEL1_SPMC=y
+          _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SEL1_SPMC=y CFG_VIRTUALIZATION=y
```
jforissier, line 130: This is now called `CFG_NS_VIRTUALIZATION`.

jenswi-linaro: Thanks, I'll fix.

jenswi-linaro:
Tag applied and comment addressed.

jforissier:
For "ci: azure: build CFG_CORE_SEL1_SPMC=y CFG_NS_VIRTUALIZATION=y":

jenswi-linaro:
Tag applied

jenswi-linaro:
The corresponding Xen patches are still under review, but I don't expect ABI changes for anything but eventual bugs in the ABI itself.

jenswi-linaro:
Please help reviewing this pull request, especially
- "core: spmc: improve checks in handle_mem_share()"
- "core: spmc: support FF-A 1.1 "
- "core: spmc: support virtualization with SPMC at S-EL1 "

is in need of attention. It's starting to be inconvenient to have this unmerged and it also puts us at a disadvantage with getting the FF-A patches in Xen merged.

balint-dobszay-arm:
> Please help reviewing this pull request, especially
> 
> * "core: spmc: improve checks in handle_mem_share()"
> * "core: spmc: support FF-A 1.1 "
> * "core: spmc: support virtualization with SPMC at S-EL1 "
> 
> are in need of attention. It's starting to be inconvenient to have this unmerged and it also puts us at a disadvantage with getting the FF-A patches in Xen merged.

I'll review these patches in the next 1-2 weeks. @imre-kis-arm might also chime in.

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -322,36 +359,52 @@ static bool is_my_uuid(uint32_t w0, uint32_t w1, uint32_t w2, uint32_t w3)
...
+	struct ffa_partition_info_x *fpi = NULL;
+
+	if (!count_only) {
+		fpi = rxtx->tx;
+		/* Add OP-TEE SP */
+		spmc_fill_partition_entry(rxtx->ffa_vers, fpi, my_endpoint_id,
+					  CFG_TEE_CORE_NB_CORE, my_part_props,
+					  my_uuid_words);
+		rxtx->tx_is_mine = false;
+		fpi++;
```
balint-dobszay-arm, line 450: This doesn't take into account that there might be `FFA_UUID_SIZE` number of extra bytes at the end of the `struct ffa_partition_info_x`. I think the increment size should depend on the FF-A version, like in `sp_partition_info_get()`. Probably it'd be easier to move the increment size calculation to a common place, e.g. `spmc_fill_partition_entry()` since there we already have different scenarios based on the FF-A version?

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -139,12 +173,15 @@ static void handle_features(struct thread_smc_args *args)
...
 #endif
+	case FFA_FEATURES:
 	case FFA_MEM_FRAG_TX:
 	case FFA_MEM_RECLAIM:
 	case FFA_MSG_SEND_DIRECT_REQ_32:
 	case FFA_INTERRUPT:
 	case FFA_PARTITION_INFO_GET:
+	case FFA_RXTX_UNMAP:
 	case FFA_RX_RELEASE:
+	case FFA_FEATURE_MANAGED_EXIT_INTR:
```
balint-dobszay-arm, line 194: I'm not sure if this is true for `CFG_CORE_SEL1_SPMC` case?

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -1528,6 +1694,8 @@ static TEE_Result spmc_init(void)
 	my_endpoint_id = spmc_get_id();
 	DMSG("My endpoint ID %#x", my_endpoint_id);
 
+	nw_rxtx.ffa_vers = INITIAL_NW_FFA_VERSION;
```
balint-dobszay-arm, line 1814: The `nw_rxtx` here is actually the RXTX buffer between OP-TEE S-EL1 SP and the S-EL2 SPMC, it's not related to the NWd right?

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -1528,6 +1694,8 @@ static TEE_Result spmc_init(void)
 	my_endpoint_id = spmc_get_id();
```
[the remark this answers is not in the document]
jenswi-linaro: Good point, I'll fix.

core/arch/arm/include/ffa.h
```diff
@@ -48,6 +51,7 @@
 #define FFA_RXTX_UNMAP			U(0x84000067)
 #define FFA_PARTITION_INFO_GET		U(0x84000068)
 #define FFA_ID_GET			U(0x84000069)
+#define FFA_SPM_ID_GET			U(0x84000085)
```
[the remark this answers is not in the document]
jenswi-linaro: I'll fix it.

core/arch/arm/include/ffa.h
```diff
@@ -48,6 +51,7 @@
...
 #define FFA_MSG_WAIT			U(0x8400006B)
 #define FFA_MSG_YIELD			U(0x8400006C)
 #define FFA_MSG_RUN			U(0x8400006D)
```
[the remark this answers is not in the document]
jenswi-linaro: OK

core/arch/arm/include/ffa.h
```diff
@@ -70,15 +74,43 @@
...
+#define FFA_FEATURE_SCHEDULE_RECV_INTR	U(0x2)
+#define FFA_FEATURE_MANAGED_EXIT_INTR	U(0x3)
+
 /* Special value for traffic targeted to the Hypervisor or SPM */
 #define FFA_TARGET_INFO_MBZ		U(0x0)
 
+#define FFA_MSG_FLAG_FRAMEWORK		BIT(31)
+#define FFA_MSG_TYPE_MASK		GENMASK_32(7, 0)
+#define FFA_MSG_PSCI			U(0x0)
+#define FFA_MSG_SEND_VM_CREATED		U(0x4)
```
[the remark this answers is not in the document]
jenswi-linaro: The corresponding defines has been reviewed at https://lore.kernel.org/xen-devel/AS8PR08MB79916CDCB5DB7CA86E8077A592989@AS8PR08MB7991.eurprd08.prod.outlook.com/

It would be confusing to use different names in OP-TEE compared to Xen and I'm not so keen on making changes to patches which has already been reviewed by someone.

If you still want to rename them, would you mind proposing that in that mail thread on the xen list?

core/arch/arm/kernel/thread_spmc.c
jenswi-linaro: Good catch, I'll fix it.

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -961,7 +1134,7 @@ void thread_spmc_msg_recv(struct thread_smc_args *args)
 	switch (args->a0) {
 #if defined(CFG_CORE_SEL1_SPMC)
 	case FFA_VERSION:
```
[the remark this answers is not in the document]
jenswi-linaro: You're right. I'll add a separate commit to fix this.

core/arch/arm/include/ffa.h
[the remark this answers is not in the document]
jenswi-linaro: Thanks

core/arch/arm/kernel/thread_spmc.c
jenswi-linaro: Right, I'll drop the `NW_` part.

jenswi-linaro:
Update

jenswi-linaro:
Rebased and tag applied to "core: spmc: improve checks in handle_mem_share()".

jenswi-linaro:
CI seems happy again.

jforissier:
- For "ci: azure: build CFG_CORE_SEL2_SPMC=y": please drop "azure:" (not used anymore), then:
  Please also drop "azure:" from "ci: azure: build CFG_CORE_SEL1_SPMC=y CFG_NS_VIRTUALIZATION=y".
- For " core: ffa: rename spmc_get_id() to get_my_id()" and "core: ffa: rename nw_rxtx to my_rxtx":
- For "core: ffa: support FFA_SPM_ID_GET", "core: ffa: support FFA_VERSION from S-EL0 SPs only" and "core: ffa: initialize my_rxtx.size":

jenswi-linaro:
Tags applied. It looks like it's mostly the commit "core: spmc: support FF-A 1.1" and its "[review]" updates that may need more review. Please let me know if I should squash in the fixups.

balint-dobszay-arm (commented):
- For "core: ffa: rename spmc_get_id() to get_my_id()" and "core: ffa: support FFA_SPM_ID_GET" and "core: ffa: initialize my_rxtx.size":
- For "core: ffa: support FFA_VERSION from S-EL0 SPs only" after addressing my comment:
- For "core: ffa: rename nw_rxtx to my_rxtx" I think there is an error in the commit msg: `s/SPMC at S-EL1/SPMC at S-EL2/`. With this fixed:
- For "core: spmc: support FF-A 1.1" and its fixups:
Please squash the fixups.

balint-dobszay-arm:
I'll finish reviewing "core: spmc: support virtualization with SPMC at S-EL1" this week, AFAICS that's the only commit left that has no review so far.

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -1073,6 +1073,14 @@ static void handle_mem_perm_set(struct thread_smc_args *args,
 		      FFA_PARAM_MBZ, FFA_PARAM_MBZ);
 }
 
+static void spmc_handle_version(struct thread_smc_args *args,
+				struct ffa_rxtx *rxtx)
+{
+	spmc_set_args(args, spmc_exchange_version(args->a0, rxtx),
```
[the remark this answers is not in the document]
jenswi-linaro: Thanks, I'll fix it.

jenswi-linaro:
Squashed and tags applied

jenswi-linaro:
Rebased on master to resolve the conflict.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -132,20 +125,19 @@ TEE_Result sp_partition_info_get(uint32_t ffa_vers, size_t buf_size,
...
 			continue;
-		if (!count_only && count < count_limit) {
-			spmc_fill_partition_entry(ffa_vers, fpi,
-						  s->endpoint_id, 1,
-						  part_props, NULL);
-			fpi = (void *)((vaddr_t)fpi + fpi_size);
+		if (!count_only && !res) {
+			uint32_t uuid_words[4] = { 0 };
+
+			tee_uuid_to_octets((uint8_t *)uuid_words, ffa_uuid);
```
balint-dobszay-arm, line 131: `ffa_uuid` is NULL when we're querying all SPs, I think the argument should be `&s->ffa_uuid` instead.

Also it looks like these changes got mixed up during the squashing, shouldn't these be part of the previous commit instead?

jenswi-linaro: I'm sorry, I don't know what happened. I'll sort it out.

jenswi-linaro: > `ffa_uuid` is NULL when we're querying all SPs, I think the argument should be `&s->ffa_uuid` instead.

You're right, I'll fix it.

jenswi-linaro:
Update

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -1044,6 +1134,14 @@ static void handle_mem_frag_tx(struct thread_smc_args *args,
 	uint32_t ret_w2 = 0;
 	uint32_t ret_w3 = 0;
 	uint32_t ret_fid = 0;
+	int rc = 0;
+
+	if (IS_ENABLED(CFG_NS_VIRTUALIZATION)) {
+		uint16_t guest_id = endpoint_id >> 16;
+
+		if (!guest_id || virt_set_guest(guest_id))
+			goto out_set_rc;
```
balint-dobszay-arm, line 1147: At this point rc is still 0, i.e. we will return from the function with FFA_SUCCESS. Is this intentional?

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -956,9 +1041,15 @@ static int handle_mem_share_rxbuf(size_t blen, size_t flen,
 		goto out;
 	}
 
+	if (IS_ENABLED(CFG_NS_VIRTUALIZATION) &&
+	    virt_set_guest(mem_trans.sender_id))
```
balint-dobszay-arm, line 1049: Shouldn't we add this in `handle_mem_share_tmem()` too? Before calling `add_mem_share()` from that function, just like here.

core/arch/arm/kernel/secure_partition.c
balint-dobszay-arm: Thanks, please squash the fixup.

balint-dobszay-arm:
In the current state of the PR I get an error when running `xtest -t ffa_spmc`:

```
* ffa_spmc_1001 Test FF-A communication
o ffa_spmc_1001.1 SP1 comms check
ffa_spmc_1000.c:186: Expression "args.args[0] == SPMC_TEST_OK" (255 == 170) is false
  ffa_spmc_1001.1 FAILED
  ffa_spmc_1001 FAILED
```

```
I/TC: E/SPM-TEST1: test_ffa_partition_info_get:199 ffa_rx_release(): unknown error -6
```

The rest of the tests are okay. I'll investigate a bit more to see what causes the issue, it's not present on the master branch so the regression is introduced somewhere in this PR.

core/arch/arm/kernel/thread_spmc.c
jenswi-linaro: Good catch, I'll update.

core/arch/arm/kernel/thread_spmc.c
jenswi-linaro: You're right, I'll fix it.

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -431,12 +488,14 @@ void spmc_handle_partition_info_get(struct thread_smc_args *args,
 	}
 
 	ret_fid = FFA_SUCCESS_32;
-	rxtx->tx_is_mine = false;
+	if (!count_only)
+		rxtx->tx_is_mine = false;
```
balint-dobszay-arm, line 547: I think this should be moved after the `out` label. From the `if (is_nil_uuid...` case we always jump to `out` which means the updating of `tx_is_mine` is skipped. I'm not sure why this didn't cause any errors before, but it looks like this is the reason for SPMC test failure I was getting.

jenswi-linaro: Makes sense, thanks for looking into this.

jenswi-linaro:
Update

balint-dobszay-arm:
- For "core: spmc: support FF-A 1.1" including fixups:

- For "core: spmc: support virtualization with SPMC at S-EL1" including fixups:

jenswi-linaro:
Squashed, tags applied, and rebased.

jenswi-linaro:
Does anyone know what the `pkcs11_1007` error is about?

jenswi-linaro:
I ran the QEMU (v7) test locally and xtest passes, not sure what's wrong with IBART.

jforissier:
> I ran the QEMU (v7) test locally and xtest passes, not sure what's wrong with IBART.

Looks like an out-of-memory condition. I could not reproduce it locally either. I have restarted IBART, we'll see.

jenswi-linaro:
Rebased on master.

jenswi-linaro:
IBART is happy again. :-)
This should be ready to be merged now.

PR #5363, opened 2022-06-03
core: pta: attestation: fix buffer size for generated key

clementfaure:
There is assertion error in the attestation PTA when it comes to
deserialize the key. Here are the logs:

* regression_1037 Remote attestation
o regression_1037.1 Get public key
E/TC:? 0 assertion '!buf_sz' failed at core/pta/attestation.c:199 <deserialize_key>
E/TC:3 0 Panic at core/kernel/assert.c:28 <_assert_break>
E/TC:3 0 TEE load address @ 0xbe000000
E/TC:3 0 Call stack:
E/TC:3 0  0xbe0091b4
E/TC:3 0  0xbe024b5c
E/TC:3 0  0xbe02292c
E/TC:3 0  0xbe02fde4
E/TC:3 0  0xbe0300c4
E/TC:3 0  0xbe029a3c
E/TC:3 0  0xbe025e70
E/TC:3 0  0xbe0336e0
E/TC:3 0  0xbe007070
E/TC:3 0  0xbe0071ec

To reproduce the issue, you need a persistent storage and follow these
steps:
 $ xtest 1037
 *reboot the platform*
 $ xtest 1037

When allocating the key buffer, the maximum buffer size is allocated
(1033 bytes) whatever the size specified by CFG_ATTESTATION_PTA_KEY_SIZE.

With default attestation key size (CFG_ATTESTATION_PTA_KEY_SIZE) of
3072 bits, only 777 bytes is needed to store the key but the allocated
buffer size is still 1033 bytes.

When the key has already been generated, the key stored is deserialized
and the deserialize_key() function expects the size of the previously
generated key to be equal to the maximum buffer size.

This issue can be fixed by allocating only the required buffer size of
the key size specified by CFG_ATTESTATION_PTA_KEY_SIZE.

Fixes: 7e05ec25b ("core: pta: add remote attestation PTA")

### DIFF
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -180,23 +180,26 @@ static TEE_Result deserialize_key(uint8_t *buf, size_t buf_sz)
 
 	res = allocate_key();
 	if (res)
-		goto out;
+		return res;
 
 	sz = deserialize_bignum(p, buf_sz, key->e);
 	if (!sz)
-		goto out;
+		goto err;
 	p += sz;
 	buf_sz -= sz;
 	sz = deserialize_bignum(p, buf_sz, key->d);
 	if (!sz)
-		goto out;
+		goto err;
 	p += sz;
 	buf_sz -= sz;
 	sz = deserialize_bignum(p, buf_sz, key->n);
 	if (!sz)
-		goto out;
-out:
-	return res;
+		goto err;
+
+	return TEE_SUCCESS;
+err:
+	free_key();
+	return TEE_ERROR_GENERIC;
 }
 
 static TEE_Result sec_storage_obj_read(TEE_UUID *uuid, uint32_t storage_id,
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -195,8 +195,6 @@ static TEE_Result deserialize_key(uint8_t *buf, size_t buf_sz)
 	sz = deserialize_bignum(p, buf_sz, key->n);
 	if (!sz)
 		goto out;
-	buf_sz -= sz;
-	assert(!buf_sz);
 out:
 	return res;
 }


core/pta/attestation.c
```diff
@@ -330,7 +330,8 @@ static TEE_Result init_key(void)
 		 * e is 65537 so its bignum size is 3 bytes. d and n can be up
 		 * to MAX_KEY_SIZE bits.
 		 */
-		size = bufsize(3, MAX_KEY_SIZE / 8, MAX_KEY_SIZE / 8);
+		size = bufsize(3, CFG_ATTESTATION_PTA_KEY_SIZE / 8,
+			       CFG_ATTESTATION_PTA_KEY_SIZE / 8);
```
jforissier, line 334: Wouldn't it be better to change the assertion instead? I mean `CFG_ATTESTATION_PTA_KEY_SIZE` should be understood as the size of the key in case it needs to be generated only, otherwise whatever has been saved previously has the priority. That would be more friendly to existing devices in case the TEE is upgraded and `CFG_ATTESTATION_PTA_KEY_SIZE` has been changed in the mean time.

clementfaure: You're right, I've changed the condition to ensure retro-compatibility.
Are you okay with always allocating the maximum size for the key?

core/pta/attestation.c
```diff
@@ -186,17 +187,17 @@ static TEE_Result deserialize_key(uint8_t *buf, size_t buf_sz)
 	if (!sz)
 		goto out;
 	p += sz;
-	buf_sz -= sz;
+	sz_done += sz;
 	sz = deserialize_bignum(p, buf_sz, key->d);
```
jforissier, line 191: If `buf_sz` is not decremented, then the size check in `deserialize_bignum()` becomes useless, doesn't it?

core/pta/attestation.c
jforissier: > Are you okay with always allocating the maximum size for the key?

Seems acceptable, yes.

core/pta/attestation.c
clementfaure: Yeah, I missed this check in `deserialize_bignum()`. I guess I can just remove the `assert()` in this `deserialize_key()`

core/pta/attestation.c
```diff
@@ -195,8 +195,6 @@ static TEE_Result deserialize_key(uint8_t *buf, size_t buf_sz)
 	sz = deserialize_bignum(p, buf_sz, key->n);
```
clementfaure, line 195: Btw @jforissier, why are we not returning a `TEE_ERROR` code when `deserialize_bignum()` returns `0` ?

jforissier (commented):
WIth the below thing addressed:

Thanks!

etienne-lms (commented):
It looks like `deserialize_key()` also lacks freeing `key` on failure: `free_key()` is called only when `generate_key()` fails. It is intended?

If it is, this P-R could slightly simplify the code replacing `goto err;` instructions with `return TEE_ERROR_GENERIC;` straight (but the 1st occurrence that should `return res;`).

clementfaure:
@etienne-lms I've addressed your comment. We should free the allocated key IMO. Let's see what @jforissier says about this

jforissier:
> @etienne-lms I've addressed your comment. We should free the allocated key IMO. Let's see what @jforissier says about this

I agree.

clementfaure:
Tags applied

etienne-lms (commented):

clementfaure:
@etienne-lms Tag applied

PR #5369, opened 2022-06-09
core: armv7: fix ASLR error

jenswi-linaro:
With commit 528dabb28254 ("core: suppress text relocation on
stack_tmp_export") the stack pointer is calculated using a relative
address instead of based on an absolute address which is relocated with
ASLR enabled.

Prior to this on Armv7 we compensate for a relocation update for
stack_tmp_export_rel in reset_secondary() just after the stack pointer
was initialized. So now when the relocation update of stack_tmp_export_rel
is gone remove the compensating code too.

Fixes: 528dabb28254 ("core: suppress text relocation on stack_tmp_export")

### DIFF
--- a/core/arch/arm/kernel/entry_a32.S
+++ b/core/arch/arm/kernel/entry_a32.S
@@ -878,17 +878,6 @@ UNWIND(	.cantunwind)
 	wait_primary
 
 	set_sp
-#ifdef CFG_CORE_ASLR
-	/*
-	 * stack_tmp_export which is used as base when initializing sp has
-	 * been relocated to the new offset. Since MMU isn't enabled on
-	 * this CPU yet we need to restore the corresponding physical
-	 * address.
-	 */
-	adr	r0, boot_mmu_config
-	ldr	r0, [r0, #CORE_MMU_CONFIG_LOAD_OFFSET]
-	sub	sp, sp, r0
-#endif
 
 #if defined (CFG_BOOT_SECONDARY_REQUEST)
 	/* if L1 is not invalidated before, do it here */


jforissier (approved):

I'm thinking perhaps we should run the CI QEMUv8 check job in 32-bit mode too? It would cover part of the ARMv7 code Although it wouldn't have caught this particular bug -- the code was outside the `CFG_WITH_ARM_TRUSTED_FW` case.

jenswi-linaro:

Tag applied

> I'm thinking perhaps we should run the CI QEMUv8 check job in 32-bit mode too? It would cover part of the ARMv7 code Although it wouldn't have caught this particular bug -- the code was outside the `CFG_WITH_ARM_TRUSTED_FW` case.

Wouldn't it make more sense to run a ARMv7 job instead then?

jforissier:
> > I'm thinking perhaps we should run the CI QEMUv8 check job in 32-bit mode too? It would cover part of the ARMv7 code Although it wouldn't have caught this particular bug -- the code was outside the `CFG_WITH_ARM_TRUSTED_FW` case.
> 
> Wouldn't it make more sense to run a ARMv7 job instead then?

You mean https://github.com/OP-TEE/manifest/blob/master/default.xml instead of https://github.com/OP-TEE/manifest/blob/master/qemu_v8.xml? Probably. It could be a separate job (good because it could run in parallel; less good because we would need a separate Docker image, at least if we adopt the same model as for v8 in order to improve build speed).

jenswi-linaro:
> You mean https://github.com/OP-TEE/manifest/blob/master/default.xml instead of https://github.com/OP-TEE/manifest/blob/master/qemu_v8.xml?

Yes

> Probably. It could be a separate job (good because it could run in parallel; less good because we would need a separate Docker image, at least if we adopt the same model as for v8 in order to improve build speed).

You're the expert here. :-)

PR #5373, opened 2022-06-10
ldelf: improve handling of weak undefined symbols

jforissier:
When processing a dynamic relocation for a weak undefined symbol, it is
not an error if the symbol is not found (such a symbol should resolve
to zero). The current code doesn't make a special case of this; it sets
the value to zero *after* the symbol lookup. That's OK when the binary
uses the legacy dynamic symbol table (DT_HASH format), because undefined
symbols are present in it. However, things will break when we switch to
the newer DT_GNU_HASH format which doesn't include them.

Update the code to prepare the DT_GNU_HASH support.

### DIFF
--- a/ldelf/ta_elf_rel.c
+++ b/ldelf/ta_elf_rel.c
@@ -186,7 +186,8 @@ success:
 
 static void e32_get_sym_name(const Elf32_Sym *sym_tab, size_t num_syms,
 			     const char *str_tab, size_t str_tab_size,
-			     Elf32_Rel *rel, const char **name)
+			     Elf32_Rel *rel, const char **name,
+			     bool *weak_undef)
 {
 	size_t sym_idx = 0;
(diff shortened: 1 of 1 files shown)


ldelf/ta_elf_rel.c
```diff
@@ -414,6 +433,14 @@ static void e64_get_sym_name(const Elf64_Sym *sym_tab, size_t num_syms,
 	if (name_idx >= str_tab_size)
 		err(TEE_ERROR_BAD_FORMAT, "Name index out of range");
 	*name = str_tab + name_idx;
+
+	if (!weak_undef)
```
jenswi-linaro, line 437: This is only non-NULL. Wouldn't it make more sense to always assume it's non-NULL?

jenswi-linaro:

PR #5377, opened 2022-06-13
core: sp: fix SP manifest UUID endianness

balint-dobszay-arm:
The Trusted Services project writes the SP UUIDs into the SP manifest in
big-endian format. However, tee_uuid_from_octets() expects little-endian
input. Fix this mismatch.


balint-dobszay-arm:
After more discussions it seems that in order to get better aligned with other consumers of the SP manifest (TF-A, Hafnium, etc.), it's better to change the endianness at the origin instead of converting here. Closing this PR, sorry for the noise.

PR #5379, opened 2022-06-15
Fix SA2UL background firewall size

glneo:


core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -83,11 +84,11 @@ static TEE_Result sa2ul_init(void)
 		}
 
 		/* Modify SA2UL firewall to allow all others access*/
-		control = FW_ENABLE_REGION;
+		control = FW_BACKGROUND_REGION | FW_ENABLE_REGION;
 		permissions[0] = (FW_WILDCARD_PRIVID << 16) | FW_NON_SECURE;
 		ret = ti_sci_set_fwl_region(fwl_id, sa2ul_region, 1,
 					    control, permissions,
-					    start_address, end_address);
+					    0x0, GENMASK_64(31, 0));
```
etienne-lms, line 91: `0, UINT32_MAX);`

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment addressed.

core/arch/arm/plat-k3/drivers/sa2ul.c
glneo: Okay. My original thought was it might need expanded to 44 bits for some platforms. If I ever do need to change it I can go back to GENMASK then.

PR #5380, opened 2022-06-16
crypto-api: ecc keypairgen for NIST P521

ldts:
Provide the valid number of bits for the curve to the underlying interface.
This allows the call to be routed to software implementations.

### DIFF
--- a/core/drivers/crypto/crypto_api/acipher/ecc.c
+++ b/core/drivers/crypto/crypto_api/acipher/ecc.c
@@ -39,6 +39,28 @@ static size_t get_ecc_key_size_bytes(uint32_t curve)
 }
 
 /*
+ * Returns the key size in bits for the given ECC curve
+ *
+ * @curve   ECC Curve ID
+ */
+static size_t get_ecc_key_size_bits(uint32_t curve)
+{
+	switch (curve) {
+	case TEE_ECC_CURVE_NIST_P192:
+	case TEE_ECC_CURVE_NIST_P224:
+	case TEE_ECC_CURVE_NIST_P256:
+	case TEE_ECC_CURVE_NIST_P384:
+		return 8 * get_ecc_key_size_bytes(curve);
+
+	case TEE_ECC_CURVE_NIST_P521:
+		return 521;
+
+	default:
+		return 0;
+	}
+}
+
+/*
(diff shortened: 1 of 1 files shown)


core/drivers/crypto/crypto_api/acipher/ecc.c
```diff
@@ -107,6 +107,8 @@ static TEE_Result ecc_generate_keypair(struct ecc_keypair *key,
 	}
 
 	key_size_bits = get_ecc_key_size_bytes(key->curve) * 8;
```
jenswi-linaro, line 109: Wouldn't it be more convenient with a `get_ecc_key_size_bits()` which does the right thing?

core/drivers/crypto/crypto_api/acipher/ecc.c
```diff
@@ -38,6 +38,28 @@ static size_t get_ecc_key_size_bytes(uint32_t curve)
...
+{
+	switch (curve) {
+	case TEE_ECC_CURVE_NIST_P192:
+	case TEE_ECC_CURVE_NIST_P224:
+	case TEE_ECC_CURVE_NIST_P256:
+	case TEE_ECC_CURVE_NIST_P384:
+		return 8 * get_ecc_key_size_bytes(curve);
+
+	case TEE_ECC_CURVE_NIST_P521:
+		return 521;
```
etienne-lms, line 56: For consistency of sizes defined in `get_ecc_key_size_bits()` and `get_ecc_key_size_bytes()`, maybe define sizes only in `get_ecc_key_size_bits()` and have `get_ecc_key_size_bytes()` do the conversion?

ldts: yes I know but I didnt like the way it was done before (and which I have replicated in other places)...IMO this is much more clear and easier to read as it immediately flags a curve that needs to be handled with care - some HW ciphers might require padding or tricks.
unless it is a strong request  from the file maintainer I'd rather keep it this way.

jenswi-linaro: I agree it certainly flags that something funny is going on, but I'd prefer if that was explained in a good comment since it's not obvious what it is from this.

ldts:
will continue in https://github.com/OP-TEE/optee_os/pull/5426

PR #5381, opened 2022-06-16
drivers: scmi-msg: Fix parameter type

ludvigparsson:
In order to make clock.c and voltage_domain.c compile on
64 bit architecture, we cannot use unsigned int in the
function prototype and size_t in the function definition.

Signed-off-by: Ludvig Pärsson <ludvig.parsson@axis.com>


etienne-lms (commented):
Thanks @ludvigpa for the fix.

ludvigparsson:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` Thanks @ludvigpa for the fix.

Added tag

PR #5383, opened 2022-06-16
ldelf: add DT_GNU_HASH support

jforissier:
It is possible for an ELF to be compiled with --hash-style=gnu;
some distros even have that as their default. This produces an ELF
with no .hash section, but with a .gnu.hash section in its stead.

GNU-style hash sections have better performance than old SYSV-style ones.
It can be expected that, as time goes on, it becomes the new default.

Use the DT_GNU_HASH table for symbol lookup when present, then fall
back onto DT_HASH.

Co-developed-by: Elvira Khabirova <e.khabirova@omp.ru>
Signed-off-by: Elvira Khabirova <e.khabirova@omp.ru>

### DIFF
--- a/ldelf/ta_elf.c
+++ b/ldelf/ta_elf.c
@@ -218,6 +218,9 @@ static void save_hashtab_from_segment(struct ta_elf *elf, unsigned int type,
 		if (tag == DT_HASH) {
 			elf->hashtab = (void *)(val + elf->load_addr);
 			break;
+		} else if (tag == DT_GNU_HASH) {
+			elf->gnu_hashtab = (void *)(val + elf->load_addr);
+			break;
 		}
 	}
 }
@@ -262,6 +265,55 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
 	check_range(elf, "DT_HASH", ptr, sz);
 }
 
+static void check_gnu_hashtab(struct ta_elf *elf, void *ptr)
+{
+	struct gnu_hashtab *h = ptr;
+	uint32_t *bucket = NULL;
+	size_t last_sym = 0;
+	size_t num_words = 4; /* nbuckets, symoffset, bloom_size, bloom_shift */
+	size_t bloom_words = 0;
+	size_t sz = 0;
+	size_t i = 0;
+
+	if (!IS_ALIGNED_WITH_TYPE(ptr, uint32_t))
+		err(TEE_ERROR_BAD_FORMAT, "Bad alignment of DT_GNU_HASH %p",
+		    ptr);
+
+	check_range(elf, "DT_GNU_HASH", h, sizeof(*h));
+
+	if (elf->is_32bit)
+		bloom_words = h->bloom_size;
+	else
+		bloom_words = h->bloom_size * 2;
+
+	if (ADD_OVERFLOW(num_words, h->nbuckets, &num_words) ||
+	    ADD_OVERFLOW(num_words, bloom_words, &num_words) ||
+	    MUL_OVERFLOW(num_words, sizeof(uint32_t), &sz))
+		err(TEE_ERROR_BAD_FORMAT, "DT_GNU_HASH overflow");
+
+	/*
+	 * There is no total number of exported symbols in DT_GNU_HASH,
+	 * so we have to find the highest bucket value manually.
+	 * See http://deroko.phearless.org/dt_gnu_hash.txt.
+	 */
+	bucket = (uint32_t *)h + 4 + bloom_words;
+	check_range(elf, "DT_GNU_HASH", bucket, h->nbuckets * sizeof(uint32_t));
+	last_sym = 0;
+	for (i = 0; i < h->nbuckets; i++)
+		if (bucket[i] > last_sym)
+			last_sym = bucket[i];
+
+	if (MUL_OVERFLOW(last_sym, sizeof(uint32_t), &last_sym) ||
+	    ADD_OVERFLOW(num_words, last_sym, &num_words))
+		err(TEE_ERROR_BAD_FORMAT, "DT_GNU_HASH overflow");
+
+	/* Shouldn't be less than the first dynsym offset */
+	if (last_sym < h->symoffset)
+		err(TEE_ERROR_BAD_FORMAT, "Bad DT_GNU_HASH");
+
+	check_range(elf, "DT_GNU_HASH", ptr, num_words * sizeof(uint32_t));
+}
+
 static void save_hashtab(struct ta_elf *elf)
 {
 	uint32_t *hashtab = NULL;
@@ -283,9 +335,13 @@ static void save_hashtab(struct ta_elf *elf)
 						  phdr[n].p_memsz);
 	}
 
-	check_hashtab(elf, elf->hashtab, 0, 0);
-	hashtab = elf->hashtab;
-	check_hashtab(elf, elf->hashtab, hashtab[0], hashtab[1]);
+	if (elf->hashtab) {
+		check_hashtab(elf, elf->hashtab, 0, 0);
+		hashtab = elf->hashtab;
+		check_hashtab(elf, elf->hashtab, hashtab[0], hashtab[1]);
+	}
+	if (elf->gnu_hashtab)
+		check_gnu_hashtab(elf, elf->gnu_hashtab);
 }
 
 static void save_soname_from_segment(struct ta_elf *elf, unsigned int type,
--- a/ldelf/ta_elf_rel.c
+++ b/ldelf/ta_elf_rel.c
@@ -32,10 +32,21 @@ static uint32_t elf_hash(const char *name)
 	return h;
 }
 
-static bool __resolve_sym(struct ta_elf *elf, unsigned int st_bind,
-			  unsigned int st_type, size_t st_shndx,
-			  size_t st_name, size_t st_value, const char *name,
-			  vaddr_t *val, bool weak_ok)
+static uint32_t gnu_hash(const char *name)
+{
+	const unsigned char *p = (const unsigned char *)name;
+	uint32_t h = 5381;
+
+	while (*p)
+		h = (h << 5) + h + *p++;
+
+	return h;
+}
+
+static bool sym_compare(struct ta_elf *elf, unsigned int st_bind,
+			unsigned int st_type, size_t st_shndx,
+			size_t st_name, size_t st_value, const char *name,
+			vaddr_t *val, bool weak_ok)
 {
 	bool bind_ok = false;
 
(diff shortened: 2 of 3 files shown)


jforissier:
Replaces https://github.com/OP-TEE/optee_os/pull/5206.

ldelf/ta_elf.c
```diff
@@ -262,6 +265,62 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
...
+	 */
+	size_t num_words = 4;
+	size_t sz = 0;
+	size_t i = 0;
+
+	if (!IS_ALIGNED_WITH_TYPE(ptr, uint32_t))
+		err(TEE_ERROR_BAD_FORMAT, "Bad alignment of DT_GNU_HASH %p",
+		    ptr);
+
+	check_range(elf, "DT_GNU_HASH", ptr, sz);
```
etienne-lms, line 285: replace `sz` with `num_words * sizeof(uint32_t)`
(`3 * sizeof(uint32_t)` seems enough seen next instruction, but `num_words` (4) seems more consistent)

ldelf/ta_elf.c
```diff
@@ -262,6 +265,62 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
...
+		err(TEE_ERROR_BAD_FORMAT, "DT_GNU_HASH overflow");
+
+	/*
+	 * There is no total number of exported symbols in DT_GNU_HASH,
+	 * so we have to find the highest bucket value manually.
+	 * See http://deroko.phearless.org/dt_gnu_hash.txt.
+	 */
+	bucket = hashtab + 4 + hashtab[2] * (elf->is_32bit ? 1 : 2);
+	last_sym = 0;
+	for (i = 0; i < hashtab[0]; i++)
```
etienne-lms, line 309: `check_range(elf, "DT_GNU_HASH", bucket, hashtab[0] * sizeof(uint32_t));` before this loop.

ldelf/ta_elf.c
```diff
@@ -283,9 +342,13 @@ static void save_hashtab(struct ta_elf *elf)
...
-	check_hashtab(elf, elf->hashtab, 0, 0);
-	hashtab = elf->hashtab;
-	check_hashtab(elf, elf->hashtab, hashtab[0], hashtab[1]);
+	if (elf->hashtab) {
+		check_hashtab(elf, elf->hashtab, 0, 0);
+		hashtab = elf->hashtab;
+		check_hashtab(elf, elf->hashtab, hashtab[0], hashtab[1]);
+	}
+	if (elf->gnu_hashtab)
+		check_gnu_hashtab(elf, elf->gnu_hashtab);
```
etienne-lms, line 351: It looks like we expect at least one of `elf->hashtab` or `elf->gnu_hashtab` is used. Should this be checked?

ldelf/ta_elf_rel.c
```diff
@@ -78,63 +89,122 @@ static bool __resolve_sym(struct ta_elf *elf, unsigned int st_bind,
...
-				     bool weak_ok)
+static bool check_found_sym(struct ta_elf *elf, const char *name,
+			    vaddr_t *val, bool weak_ok,
+			    size_t n, bool is_32)
 {
+	Elf32_Sym *sym32 = NULL;
+	Elf64_Sym *sym64 = NULL;
+
+	if (n >= elf->num_dynsyms)
+		err(TEE_ERROR_BAD_FORMAT, "Index out of range");
```
etienne-lms, line 100: prefer with an empty line below

ldelf/ta_elf_rel.c
```diff
@@ -78,63 +89,122 @@ static bool __resolve_sym(struct ta_elf *elf, unsigned int st_bind,
 	return true;
 }
 
-static TEE_Result resolve_sym_helper(uint32_t hash, const char *name,
-				     vaddr_t *val, struct ta_elf *elf,
-				     bool weak_ok)
+static bool check_found_sym(struct ta_elf *elf, const char *name,
+			    vaddr_t *val, bool weak_ok,
+			    size_t n, bool is_32)
```
etienne-lms, line 94: not really needed to pass `is_32` since this is `elf->is_32bit`

ldelf/ta_elf_rel.c
```diff
@@ -78,63 +89,122 @@ static bool __resolve_sym(struct ta_elf *elf, unsigned int st_bind,
...
+		uint32_t symskip = hashtab[1];
+		uint32_t bloom_size = hashtab[2];
+		uint32_t *bucket = NULL;
+		uint32_t *chain = NULL;
+		uint32_t hashval = 0;
+
+		hash = gnu_hash(name);
+
+		if (elf->is_32bit) {
+			uint32_t *bloom = (void *)&hashtab[4];
```
etienne-lms, line 154: Cast not needed, here and at line 161. 
Do you prefer to keep them for consistency with the 64bit case instruction sequence?

ldelf/ta_elf.c
```diff
@@ -262,6 +265,62 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
...
+
+	if (MUL_OVERFLOW(last_sym, sizeof(uint32_t), &last_sym) ||
+	    ADD_OVERFLOW(num_words, last_sym, &num_words))
+		err(TEE_ERROR_BAD_FORMAT, "DT_GNU_HASH overflow");
+
+	/* Shouldn't be less than the first dynsym offset */
+	if (last_sym < hashtab[1])
+		err(TEE_ERROR_BAD_FORMAT, "Bad DT_GNU_HASH");
+
+	check_range(elf, "DT_GNU_HASH", ptr, sz);
```
etienne-lms, line 321: s/sz/num_words * sizeof(uint32_t)/

etienne-lms (commented):
in commit log: s/in its stead/instead/

ldelf/ta_elf.c
jforissier: Right, I'll use `num_words`.

ldelf/ta_elf.c
jforissier: OK

ldelf/ta_elf_rel.c
jforissier: No, useless is useless ;) I'll fix.

ldelf/ta_elf.c
jforissier: Yes, will fix

ldelf/ta_elf.c
```diff
@@ -262,6 +265,63 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
...
+	     ADD_OVERFLOW(num_words, hashtab[2], &num_words)) ||
+	    MUL_OVERFLOW(num_words, sizeof(uint32_t), &sz))
+		err(TEE_ERROR_BAD_FORMAT, "DT_GNU_HASH overflow");
+
+	/*
+	 * There is no total number of exported symbols in DT_GNU_HASH,
+	 * so we have to find the highest bucket value manually.
+	 * See http://deroko.phearless.org/dt_gnu_hash.txt.
+	 */
+	bucket = hashtab + 4 + hashtab[2] * (elf->is_32bit ? 1 : 2);
```
jenswi-linaro, line 307: I'd prefer helper variables instead of using `hashtab[0..3]` directly.

jforissier: Even better (I think): I have introduced `struct gnu_hashtab`. Please check again.

ldelf/ta_elf.c
```diff
@@ -282,7 +282,7 @@ static void check_gnu_hashtab(struct ta_elf *elf, void *ptr)
 		err(TEE_ERROR_BAD_FORMAT, "Bad alignment of DT_GNU_HASH %p",
 		    ptr);
 
-	check_range(elf, "DT_GNU_HASH", ptr, sz);
+	check_range(elf, "DT_GNU_HASH", ptr, num_words);
```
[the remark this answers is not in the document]
jforissier: Fixed.

ldelf/ta_elf_rel.c
```diff
@@ -78,63 +89,120 @@ static bool __resolve_sym(struct ta_elf *elf, unsigned int st_bind,
...
+	uint32_t n = 0;
+	uint32_t hash = 0;
+
+	if (elf->gnu_hashtab) {
+		/*
+		 * Using uint32_t here for convenience because both Elf64_Word
+		 * and Elf32_Word are 32-bit types
+		 */
+		uint32_t *hashtab = elf->gnu_hashtab;
+		uint32_t nbuckets = hashtab[0];
```
jenswi-linaro, line 144: Use `struct gnu_hashtab`?

jenswi-linaro:

parport0:
@jforissier , thank you for picking up my changes. I'm glad there was a way to not let it go to waste.

ldelf/ta_elf.c
```diff
@@ -262,6 +265,55 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
...
+	 * See http://deroko.phearless.org/dt_gnu_hash.txt.
+	 */
+	bucket = (uint32_t *)h + 4 + bloom_words;
+	check_range(elf, "DT_GNU_HASH", bucket, h->nbuckets * sizeof(uint32_t));
+	last_sym = 0;
+	for (i = 0; i < h->nbuckets; i++)
+		if (bucket[i] > last_sym)
+			last_sym = bucket[i];
+
+	if (MUL_OVERFLOW(last_sym, sizeof(uint32_t), &last_sym) ||
```
etienne-lms, line 306: Is this multiplication needed? IIUC `last_sym` is a word count hence should be added to `num_words`.

ldelf/ta_elf.c
```diff
@@ -262,6 +265,55 @@ static void check_hashtab(struct ta_elf *elf, void *ptr, size_t num_buckets,
...
+	    ADD_OVERFLOW(num_words, last_sym, &num_words))
```
etienne-lms, line 307: According to referred http://deroko.phearless.org/dt_gnu_hash.txt, `last_sym` at this stage gives the (relative) index of the last chain start, but does give last chain end. I seems to me `check_range()` below should consider the last chain size (until bit0 of a chain word is set).

ldelf/ta_elf.c
jforissier: Sounds right, will fix.

ldelf/ta_elf.c
jforissier: The problem is we don't know the end of the chain at this point so we can't iterate (calling `check_range()` to make sure data can be accessed **after** accessing them would not make sense of course). I am not sure how useful these range checks really are. We could either ignore this part of the check, or (better) obtain the size of the GNU hash section by other means and check against it during symbol resolution. In fact I think I can simplify the code in `save_hashtab()` by getting hash table info from the section headers (which include start and size info) rather than from the dynamic section (which contains only the start).

etienne-lms:
@jforissier, can close this P-R.

PR #5385, opened 2022-06-17
[RFC] embed SCMI service implementation from SCP-firmware

etienne-lms:
This P-R proposes to build an SCMI server stack using [SCP-firmware](https://github.com/ARM-Software/SCP-firmware.git) reference implementation. Since this component is not ready yet to be embedded in OP-TEE core, the P-R synchronizes with a fork of the SCP-firmware Git repository until the mainline reference implementation merges required changes.

SCP-firmware is the reference open source implementation of SCMI services, maintained by Arm. It targets full implementation of the SCMI specification. This software component was initially designed for Cortex-M targets and is currently evolving to be embedded in more execution environment, including OP-TEE core, Linux based virtual machine and more. In OP-TEE scope, this implementation is an alternate to the SCMI service implementation based on scmi-msg drivers that only offers basic SCMI services.

The P-R shows integration of this implementation in platform stm32mp1.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "core/scmi-server/SCP-firmware"]
 	path = core/scmi-server/SCP-firmware
-	url = https://github.com/ARM-Software/SCP-firmware.git
+	url = https://github.com/etienne-lms/SCP-firmware.git
+	branch = optee-stm32mp1
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "core/scmi-server/SCP-firmware"]
+	path = core/scmi-server/SCP-firmware
+	url = https://github.com/ARM-Software/SCP-firmware.git
--- a/core/arch/arm/include/scmi/scmi_server.h
+++ b/core/arch/arm/include/scmi/scmi_server.h
@@ -1,0 +1,91 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2019-2022, Linaro Limited
+ */
+#ifndef SCMI_SERVER_H
+#define SCMI_SERVER_H
+
+#include <tee_api_types.h>
+#include <types_ext.h>
+
+#ifdef CFG_SCMI_SERVER
+/*
+ * Request processing of an incoming event in the SCMI server for a target
+ * MHU/SMT mailbox.
+ * The function is exported by scmi-server.
+ *
+ * @channel_id: SCMI channel handler
+ */
+TEE_Result scmi_server_smt_process_thread(unsigned int channel_id);
+
+/*
+ * Request processing of an incoming event in the SCMI server for a target
+ * MHU/MSG mailbox.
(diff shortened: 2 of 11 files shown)


etienne-lms:
Update with SCP-firmware implementing generic modules for OP-TEE clock and reset controller resources. This update is also based on added clk API function `clk_get_rates_array()` proposed in P-R https://github.com/OP-TEE/optee_os/pull/4954. This update does not yet include description of exposed SCMI clocks from OP-TEE secure DT, as proposed by https://github.com/OP-TEE/optee_os/pull/4954.

jenswi-linaro:
Regarding "core: scmi_server: host SCP-firmware source tree for SCMI services". I'm not sure git submodules are the solution here, we've said no to that a few times already.

jforissier:
> Regarding "core: scmi_server: host SCP-firmware source tree for SCMI services". I'm not sure git submodules are the solution here, we've said no to that a few times already.

IIRC the main reason was that submodules do not allow to carry local modifications easily (a Git repo has to exist somewhere). It was problematic for both LibTomCrypt and MBed TLS, for which we do have small patches on top of upstream. Would things be a bit different here? Such as, we would not expect any local changes in the SCP-firmware?

Personally I have nothing against submodules (note that in our build environments, some projects do have submodules: QEMU and EDK2 for instance). We just need to make sure we use them wisely I suppose :)

jenswi-linaro:
I'm not so keen on git submodules in optee_os.git.
Can't we use the same method as with `CFG_NXP_SE05X_PLUG_AND_TRUST`?

jenswi-linaro:
> > Can't we use the same method as with `CFG_NXP_SE05X_PLUG_AND_TRUST`?
> 
> If we do so, we'll need a specific build env in SCP-firmware for optee, as plug-and-trust does: https://github.com/foundriesio/plug-and-trust/blob/optee_lib/sub.mk. Such a change, specifically for OP-TEE would be quite invasive in SCP-firmware filetree and build process.

I meant using a `CFG_SCMI_SERVER_SCP_PATH` replacing `$(libdir)/$(scmi-server-scp-path)` in `core/scmi-server/sub.mk` and then point `CFG_SCMI_SERVER_SCP_PATH` somewhere outside optee_os.git.

> 
> For info, SCP-firmware currently expect to be built from either a entry GNU Makefile file or an entry CMakeList.txt and the GNU Makefile build env will soon be deprecated in favor of a CMake only build env. When building for OP-TEE, I would expect OP-TEE build directives are passed to SCP-firmware build process, yet relying on SCP-firmware native build sequence.

Ouch, that doesn't sound very nice. If this is intended for more than just one odd platform it should be worth considering an OP-TEE contained build including having the source in optee_os.git. For the normal reasons:
1. Being able to tell what you have built
2. Have something reproducible
3. Have an environment you understand and which works in a consistent way, not two, tree or four all with different quirks.

If we can do this for LTC and mbedTLS I'm sure we can do it here too, if there's a will.

I understand why the Linux kernel has its requirement to be self contained.

PR #5387, opened 2022-06-21
core: dt_driver: rework dt_driver_register_provider()

GseoC:
This change makes dt_driver_register_provider() function return a TEE_SUCCESS instead of an error. Not getting a phandle should result in the provider not being registered with no error returned.


etienne-lms (commented):

GseoC:
Tag applied

PR #5388, opened 2022-06-21
plat-stm32mp1: use stm32mp_dt_bindings.h helper

GseoC:
This P-R implements the use of the stm32mp_dt_bindings.h helper which was upstreamed but not used.

This header groups all the device tree bindings necessary to compile a particular platform. This way, drivers do not include some headers under compilation switches: they remain in this helper.


etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for the 3 commits.

For commit "core: drivers: fix stm32mp_dt_bindings.h":
Could add a `Fixes:` tag in commit message:
```raw
Fixes: 19a4632e0f17 ("dt-bindings: stm32: add stm32mp13 clock and reset bindings")

PR #5389, opened 2022-06-21
plat-stm32mp1: stm32mp15/13 platforms conf update

GseoC:
This P-R brings various configuration updates for the STM32MP1 platforms. They aim to optimize each of them.

RAM configuration is aligned with TF-A. We accept to "loose" the shared-memory size on STM32MP15 for now.


GseoC:
> ```
> 
> plat-stm32mp1: conf: default disable ASLR
> 
> Default disable CFG_CORE_ASLR on stm32mp1. The platform memory
> firewall does not allow secure world to access external DTB in
> non-secure memory when MMU is OFF, which is what the software attempts
> to do when CFG_CORE_ASLR=y.
> 
> Signed-off-by: Gatien Chevallier <gatien.chevallier@foss.st.com>
> ```
> 
> Jens' comment regarding having the DTB always in secure memory is relevant: [#5350 (comment)](https://github.com/OP-TEE/optee_os/pull/5350#issuecomment-1136958516)
> 
> It is unfortunate that ASLR should be disabled only for this reason...

Hello Jérôme,

We are currently investigating on this point and looking for a way to have our DTB in secure memory. For now, it would imply too much change but we will come to that soon.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -174,13 +174,13 @@ ifeq ($(CFG_WITH_PAGER),y)
...
 CFG_UNWIND ?= n
 CFG_LOCKDEP ?= n
-CFG_CORE_ASLR ?= n
 CFG_TA_BGET_TEST ?= n
 # Default disable early TA compression to support a smaller HEAP size
 CFG_EARLY_TA_COMPRESS ?= n
 CFG_CORE_HEAP_SIZE ?= 49152
 endif
 
+CFG_CORE_ASLR ?= n
```
etienne-lms, line 183: nitpicking: for consistency of _conf.mk_ layout, I would prefer this line is moved next to generic configs, around lines 46/50:
```patch
 CFG_WITH_LPAE ?= y
 CFG_MMAP_REGIONS ?= 23
+CFG_CORE_ASLR ?= n
 
 ifeq ($(CFG_EMBED_DTB_SOURCE_FILE),)
 # Some drivers mandate DT support
```

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for the series with comment preferably  addressed.

GseoC:
Tag applied

PR #5390, opened 2022-06-22
core: handle large holes in S-EL0 map 

jenswi-linaro:


jenswi-linaro:
I wouldn't mind getting this included in the release, but then we should be able to merge it sooner rather than later.

jenswi-linaro:
Tag applied

PR #5393, opened 2022-06-23
plat-stm32mp1: rework the use of shared resources

GseoC:
This P-R introduce the use of the shared IO driver which use is to isolate the use of shared resources that aren't relevant for STM32MP13 platform.

Now, the use of such resources is under the CFG_STM32MP1_SHARED_RESOURCES compilation flag, which is disabled by
default but enabled for STM32MP15.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -69,10 +69,12 @@ endif
 ifeq ($(CFG_STM32MP13),y)
 $(call force,CFG_STM32MP15,n)
 $(call force,CFG_STM32MP_CLK_CORE,y)
+$(call force,CFG_STM32MP1_SHARED_RESOURCES,n)
 $(call force,CFG_STM32MP13_CLK,y)
 $(call force,CFG_STM32MP15_CLK,n)
 CFG_STM32MP_OPP_COUNT ?= 2
 else
+$(call force,CFG_STM32MP1_SHARED_RESOURCES,y)
 $(call force,CFG_STM32MP15,y)
 $(call force,CFG_STM32MP15_CLK,y)
 endif
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -41,6 +41,7 @@ $(call force,CFG_PSCI_ARM32,y)
 $(call force,CFG_SECONDARY_INIT_CNTFRQ,y)
 $(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
 $(call force,CFG_SM_PLATFORM_HANDLER,y)
+$(call force,CFG_STM32_SHARED_IO,y)
 $(call force,CFG_WITH_SOFTWARE_PRNG,y)
 
 CFG_TEE_CORE_NB_CORE ?= 2
--- a/core/arch/arm/plat-stm32mp1/stm32_util.h
+++ b/core/arch/arm/plat-stm32mp1/stm32_util.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
- * Copyright (c) 2018-2019, STMicroelectronics
+ * Copyright (c) 2018-2022, STMicroelectronics
  */
 
 #ifndef __STM32_UTIL_H__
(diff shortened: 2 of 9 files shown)


core/arch/arm/plat-stm32mp1/stm32_util.h
```diff
@@ -265,4 +281,59 @@ static inline void stm32mp_register_clock_parents_secure(unsigned long clock_id
 }
 #endif
```
etienne-lms, line 282: can simplify these:
```patch
 #ifdef CFG_STM32MP1_SHARED_RESOURCES
 /* Register resource @id as a secure peripheral */
 void stm32mp_register_secure_periph(enum stm32mp_shres id);
 
 (...)
 
 /* Return true if and only if GPIO bank @bank is registered as non-secure */
 bool stm32mp_gpio_bank_is_non_secure(unsigned int bank);
 
-#if defined(CFG_STM32MP15)
 /* Register parent clocks of @clock (ID used in clock DT bindings) as secure */
 void stm32mp_register_clock_parents_secure(unsigned long clock_id);
-#else
-static inline void stm32mp_register_clock_parents_secure(unsigned long clock_id
-							 __unused)
-{
-}
-#endif
 #else /* CFG_STM32MP1_SHARED_RESOURCES */
-
 static inline void stm32mp_register_secure_periph(enum stm32mp_shres id
 						  __unused)
 {
 }
 
 (...)
 
 static inline bool stm32mp_gpio_bank_is_non_secure(unsigned int bank __unused)
 {
 	return false;
 }
 
-void stm32mp_register_clock_parents_secure(unsigned long clock_id);
+static inline
+void stm32mp_register_clock_parents_secure(unsigned long clock_id __unused)
+{
+}
 #endif /* CFG_STM32MP1_SHARED_RESOURCES */
-
 #endif /*__STM32_UTIL_H__*/

File: core/arch/arm/plat-stm32mp1/stm32_util.h [hunk 2]
```diff
@@ -265,4 +281,59 @@ static inline void stm32mp_register_clock_parents_secure(unsigned long clock_id
...
+{
+	return false;
+}
+
+static inline bool stm32mp_gpio_bank_is_non_secure(unsigned int bank __unused)
+{
+	return false;
+}
+
+void stm32mp_register_clock_parents_secure(unsigned long clock_id);
```
etienne-lms, line 335: see above comment

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -127,6 +128,9 @@ CFG_WDT ?= $(CFG_STM32_IWDG)
 # Platform specific configuration
 CFG_STM32MP_PANIC_ON_TZC_PERM_VIOLATION ?= y
 
+# Default disable shared resources
+CFG_STM32MP1_SHARED_RESOURCES ?= n
```
etienne-lms, line 132: as CFG_STM32MP13=y does not expect to enable `CFG_STM32MP1_SHARED_RESOURCES`, I would suggest to force it disabled in bertween lines 69 to 75:
```patch
 ifeq ($(CFG_STM32MP13),y)
 $(call force,CFG_STM32MP15,n)
 $(call force,CFG_STM32MP_CLK_CORE,y)
+$(call force,CFG_STM32MP1_SHARED_RESOURCES,n)
 $(call force,CFG_STM32MP13_CLK,y)
 $(call force,CFG_STM32MP15_CLK,n)
 CFG_STM32MP_OPP_COUNT ?= 2
 else
```

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit
"drivers: stm32_shared_io: introduce shared io driver."
(but that you can remove the period `.` from the commit log header line)

Comments for commit "plat-stm32mp1: introduce CFG_STM32MP1_SHARED_RESOURCES"

core/arch/arm/plat-stm32mp1/conf.mk
GseoC: > as CFG_STM32MP13=y does not expect to enable `CFG_STM32MP1_SHARED_RESOURCES`, I would suggest to force it disabled in bertween lines 69 to 75:
> 
> ```diff
>  ifeq ($(CFG_STM32MP13),y)
>  $(call force,CFG_STM32MP15,n)
>  $(call force,CFG_STM32MP_CLK_CORE,y)
> +$(call force,CFG_STM32MP1_SHARED_RESOURCES,n)
>  $(call force,CFG_STM32MP13_CLK,y)
>  $(call force,CFG_STM32MP15_CLK,n)
>  CFG_STM32MP_OPP_COUNT ?= 2
>  else
> ```


It is disabled by default but I guess we can force it to prevent further issues

etienne-lms: remove these lines since config switch is forced above.

etienne-lms (changes requested):

GseoC:
Tag updated

PR #5396, opened 2022-06-24
dts: sama5d2: Set TCB1 as secure

clementleger:
While checking for OP-TEE 3.18, found that the device-tree modifications
to enable the TCB were missing. Add missing status-okay line to enable
TCB1 for OP-TEE usage.


jforissier (commented):
If you say so :wink:

PR #5397, opened 2022-06-24
core: make sure build date is always in English

jforissier:
Setting LANG=C before invoking the date command doesn't always result in
the "C" (English) locale being selected. The correct way is to set
LC_ALL. As explained in the locale(7) man page:

 If the second argument to setlocale(3) is an empty string, "", for the
 default locale, it is determined using the following steps:

 1. If there is a non-null environment variable LC_ALL, the value of
    LC_ALL is used.

 2. If an environment variable with the same name as one of the
    categories above exists and is non-null, its value is used for that
    category.

 3. If there is a non-null environment variable LANG, the value of LANG
    is used.

Fixes: 3e2b963515c1 ("core: use C locale when generating the build date")

### DIFF
--- a/core/arch/arm/kernel/link.mk
+++ b/core/arch/arm/kernel/link.mk
@@ -148,7 +148,7 @@ version-o-cflags = $(filter-out -g3,$(core-platform-cflags) \
 ifneq ($(SOURCE_DATE_EPOCH),)
 date-opts = -d @$(SOURCE_DATE_EPOCH)
 endif
-DATE_STR = `LANG=C date -u $(date-opts)`
+DATE_STR = `LC_ALL=C date -u $(date-opts)`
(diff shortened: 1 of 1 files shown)


igoropaniuk:
```
I/TC: OP-TEE version: 3.17.0-150-g37010ab7-dev (gcc version 10.2.1 20201103 (GNU Toolchain for the A-profile Architecture 10.2-2020.11 (arm-10.16))) #3 Sat Jun 25 11:14:01 UTC 2022 arm
```

```
Reviewed-by: Igor Opaniuk <igor.opaniuk@foundries.io>
Tested-by: Igor Opaniuk <igor.opaniuk@foundries.io>
```

jenswi-linaro:

PR #5398, opened 2022-06-25
imx_ocotp: correct ocotp fuse address computation

Tim-Anderson:
Not all the imx platforms have the same bank spacing. In particular
the imx6 and imx7ulp both have bank sizes that are twice as large
as the imx8m.

This will set the correct computation.

Signed-off-by: Tim Anderson <tim.anderson@foundries.io>

### DIFF
--- a/core/drivers/imx_ocotp.c
+++ b/core/drivers/imx_ocotp.c
@@ -14,7 +14,12 @@
 #define OCOTP_CTRL		    0x0
 #define OCOTP_CTRL_ERROR	    BIT32(9)
 #define OCOTP_CTRL_BUSY		    BIT32(8)
+
+#if defined(CFG_MX6) || defined(CFG_MX7ULP)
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

clementfaure:
You're right, the bank size differs depending on the platform.
However, you also need to change the bank and word numbers in `ocotp_get_die_id_mx7ulp()`. This function was written with a bank size of `0x40`. It should not be a problem for `ocotp_get_die_id_mx()`

jforissier:
I suppose one commit is not correct without the other? If so the second one should be titled "[Review] imx_ocotp: correct ocotp fuse address computation" to indicate it is a fixup commit pushed for review, to be later squashed into the first one once reviewers have approved.

clementfaure:
`Reviewed-by: Clement Faure <clement.faure@nxp.com>`

etienne-lms (commented):
my A-b still applies

jforissier:
@Tim-Anderson thanks.

PR #5404, opened 2022-06-28
core:tee: fix TA crash if RPMB key is not provisioned

wangjudyw:
If optee-os is set with CFG_RPMB_WRITE_KEY=n, returning
TEE_ERROR_BAD_STATE will result in panic in optee-os and TA
will crash when trying to access RPMB. Instead returning
TEE_ERROR_STORAGE_NOT_AVAILABLE to signal TA RPMB not ready for use,
and allow TA having chance to perform error handling.

Signed-off-by: Judy Wang <wangjudy@microsoft.com>


jforissier (approved):
> will result in panic in optee-os and TA will crash when trying to access RPMB

This is ambiguous, there is no panic in the OP-TEE core, please rephrase as "will result in a TA panic when trying to access RPMB". Also please use "core: " rather than "core:tee: " as a subject prefix. With that LGTM:

jforissier:
Yes but still "panic in optee-os" sounds like the TEE core (kernel) would panic, which is a [different thing](https://github.com/OP-TEE/optee_os/blob/3.17.0/core/include/kernel/panic.h#L34). The panic we are talking about happens in user space (in the TA, in the libutee library as you mentioned) and in this case we always talk about "TA panic".

wangjudyw:
log is modified accordingly.

jforissier:
@wangjudyw excellent, thank you.

jenswi-linaro:

etienne-lms (commented):

wangjudyw:
updated log to add "Reviewed-by"

PR #5406, opened 2022-06-28
Rename HWRNG PTA header file

etienne-lms:
Renames rng_pta_client.h to pta_rng.h for consistency with other PTA header filess exported in OP-TEE SDK.


lib/libutee/include/rng_pta_client.h
```diff
@@ -1,45 +1,14 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
- * Copyright (C) 2018, Linaro Limited
+ * Copyright (C) 2018-2022, Linaro Limited
+ *
+ * This header file will sooner or later be deprecated, unless endlessly
+ * maintained for backward compatibility.
```
jforissier, line 6: With this PR, it is deprecated already, and may be *removed* sooner or later. So I suggest:
"This header file is deprecated and kept for backward compatibility."

core/arch/arm/plat-synquacer/main.c
```diff
@@ -16,7 +16,7 @@
 #include <kernel/timer.h>
 #include <mm/core_memprot.h>
 #include <platform_config.h>
-#include <rng_pta.h>
+#include <synq_rng_pta.h>
```
jforissier, line 19: I'd rather use double quotes to make it clear it's a local file (and keep the current name). If the name is change then `synquacer_rng_pta.h` seems preferable to `synq_rng_pta.h`.

jforissier (commented):
- For commit: "pta: rng: upgrade pta header file to optee standard"
I prefer commit subject to be as precise as possible regarding the "what" the commit does. How about: "pta: rng: rename rng_pta_client.h to pta_rng.h", then mention in the commit description that it is the naming convention used by all other PTAs (the "why"). With that and the minor comment below addressed:
- For the other commit, see comment.

core/arch/arm/plat-synquacer/main.c
etienne-lms: double :+1:

etienne-lms:
Updated commit messages. @jforissier, tell me if they look good to you.

etienne-lms:
I've fixed the header file inline comment (forgot to in previous update):
```patch
- * This header file will sooner or later be deprecated, unless endlessly
- * maintained for backward compatibility.
+ * This header file is deprecated and kept for backward compatibility.
```
Tag applied.

etienne-lms:
Updated 1st commit to fix against checkpatch report. Review taq remains.
```patch
- * [inout]   memref[0] - Entropy buffer memory reference
+ * [in/out]   memref[0] - Entropy buffer memory reference
```

PR #5409, opened 2022-06-28
Preallocate translation tables for S-EL0 contexts

jenswi-linaro:

### DIFF
--- a/core/arch/arm/mm/pgt_cache.c
+++ b/core/arch/arm/mm/pgt_cache.c
@@ -166,57 +166,53 @@ static bool pgt_entry_matches(struct pgt *p, vaddr_t begin, vaddr_t last)
 				     last - begin);
 }
 
-void pgt_flush_ctx_range(struct pgt_cache *pgt_cache,
-			 struct ts_ctx *ctx __unused,
-			 vaddr_t begin, vaddr_t last)
+void pgt_flush_range(struct user_mode_ctx *uctx, vaddr_t begin, vaddr_t last)
 {
-	if (pgt_cache) {
-		struct pgt *p;
-		struct pgt *next_p;
+	struct pgt_cache *pgt_cache = &uctx->pgt_cache;
+	struct pgt *next_p = NULL;
+	struct pgt *p = NULL;
 
-		/*
-		 * Do the special case where the first element in the list is
-		 * removed first.
-		 */
+	/*
+	 * Do the special case where the first element in the list is
+	 * removed first.
+	 */
+	p = SLIST_FIRST(pgt_cache);
+	while (pgt_entry_matches(p, begin, last)) {
+		SLIST_REMOVE_HEAD(pgt_cache, link);
+		free_pgt(p);
 		p = SLIST_FIRST(pgt_cache);
-		while (pgt_entry_matches(p, begin, last)) {
-			SLIST_REMOVE_HEAD(pgt_cache, link);
-			free_pgt(p);
-			p = SLIST_FIRST(pgt_cache);
+	}
+
+	/*
+	 * p either points to the first element in the list or it's NULL,
+	 * if NULL the list is empty and we're done.
+	 */
+	if (!p)
+		return;
+
+	/*
+	 * Do the common case where the next element in the list is
+	 * removed.
+	 */
+	while (true) {
+		next_p = SLIST_NEXT(p, link);
+		if (!next_p)
+			break;
+		if (pgt_entry_matches(next_p, begin, last)) {
+			SLIST_REMOVE_AFTER(p, link);
+			free_pgt(next_p);
+			continue;
 		}
 
-		/*
-		 * p either points to the first element in the list or it's
-		 * NULL, if NULL the list is empty and we're done.
-		 */
-		if (!p)
-			return;
-
-		/*
-		 * Do the common case where the next element in the list is
-		 * removed.
-		 */
-		while (true) {
-			next_p = SLIST_NEXT(p, link);
-			if (!next_p)
-				break;
-			if (pgt_entry_matches(next_p, begin, last)) {
-				SLIST_REMOVE_AFTER(p, link);
-				free_pgt(next_p);
-				continue;
-			}
-
-			p = SLIST_NEXT(p, link);
-		}
+		p = SLIST_NEXT(p, link);
 	}
 }
 
-void pgt_flush_ctx(struct ts_ctx *ctx)
+void pgt_flush(struct user_mode_ctx *uctx)
 {
-	struct pgt_cache *pgt_cache = NULL;
+	struct pgt_cache *pgt_cache = &uctx->pgt_cache;
 	struct pgt *p = NULL;
 
-	pgt_cache = &to_user_mode_ctx(ctx)->pgt_cache;
 	while (true) {
 		p = SLIST_FIRST(pgt_cache);
 		if (!p)
@@ -226,32 +222,29 @@ void pgt_flush_ctx(struct ts_ctx *ctx)
 	}
 }
 
-void pgt_clear_ctx_range(struct pgt_cache *pgt_cache,
-			 struct ts_ctx *ctx __unused,
-			 vaddr_t begin, vaddr_t end)
+void pgt_clear_range(struct user_mode_ctx *uctx, vaddr_t begin, vaddr_t end)
 {
-	if (pgt_cache) {
-		struct pgt *p = NULL;
-	#ifdef CFG_WITH_LPAE
-		uint64_t *tbl = NULL;
-	#else
-		uint32_t *tbl = NULL;
-	#endif
-		unsigned int idx = 0;
-		unsigned int n = 0;
+	struct pgt_cache *pgt_cache = &uctx->pgt_cache;
+	struct pgt *p = NULL;
+#ifdef CFG_WITH_LPAE
+	uint64_t *tbl = NULL;
+#else
+	uint32_t *tbl = NULL;
+#endif
+	unsigned int idx = 0;
+	unsigned int n = 0;
 
-		SLIST_FOREACH(p, pgt_cache, link) {
-			vaddr_t b = MAX(p->vabase, begin);
-			vaddr_t e = MIN(p->vabase + CORE_MMU_PGDIR_SIZE, end);
+	SLIST_FOREACH(p, pgt_cache, link) {
+		vaddr_t b = MAX(p->vabase, begin);
+		vaddr_t e = MIN(p->vabase + CORE_MMU_PGDIR_SIZE, end);
 
-			if (b >= e)
-				continue;
+		if (b >= e)
+			continue;
 
-			tbl = p->tbl;
-			idx = (b - p->vabase) / SMALL_PAGE_SIZE;
-			n = (e - b) / SMALL_PAGE_SIZE;
-			memset(tbl + idx, 0, n * sizeof(*tbl));
-		}
+		tbl = p->tbl;
+		idx = (b - p->vabase) / SMALL_PAGE_SIZE;
+		n = (e - b) / SMALL_PAGE_SIZE;
+		memset(tbl + idx, 0, n * sizeof(*tbl));
 	}
 }
 
@@ -275,8 +268,10 @@ static struct pgt *prune_before_va(struct pgt_cache *pgt_cache, struct pgt *p,
 	return p;
 }
 
-bool pgt_check_avail(struct pgt_cache *pgt_cache, struct vm_info *vm_info)
+bool pgt_check_avail(struct user_mode_ctx *uctx)
 {
+	struct pgt_cache *pgt_cache = &uctx->pgt_cache;
+	struct vm_info *vm_info = &uctx->vm_info;
 	struct pgt *p = SLIST_FIRST(pgt_cache);
 	struct vm_region *r = NULL;
 	struct pgt *pp = NULL;
@@ -285,10 +280,10 @@ bool pgt_check_avail(struct pgt_cache *pgt_cache, struct vm_info *vm_info)
 
 	/*
 	 * Prune unused tables. This is normally not needed since
-	 * pgt_flush_ctx_range() does this too, but in the error path of
-	 * for instance vm_remap() such calls may not be done. So for
-	 * increased robustness remove all unused translation tables before
-	 * we may allocate new ones.
+	 * pgt_flush_range() does this too, but in the error path of for
+	 * instance vm_remap() such calls may not be done. So for increased
+	 * robustness remove all unused translation tables before we may
+	 * allocate new ones.
 	 */
 	TAILQ_FOREACH(r, &vm_info->regions, link) {
 		for (va = ROUNDDOWN(r->va, CORE_MMU_PGDIR_SIZE);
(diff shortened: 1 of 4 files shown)


etienne-lms:
@jenswi-linaro, do you think these should hit 3.18.0?

jenswi-linaro:
> @jenswi-linaro, do you think these should hit 3.18.0?

No, this can wait. I just wanted to publish before my laptop self destructs or something. :-)

core/include/mm/pgt_cache.h
```diff
@@ -66,28 +67,23 @@ bool pgt_check_avail(struct pgt_cache *pgt_cache, struct vm_info *vm_info);
...
  * pgt_put_all() - informs the translation table manager that these tables
  *		   will not be needed for while
  * @pgt_cache:	list of translation table to put to rest
  */
 #if defined(CFG_CORE_PREALLOC_EL0_TBLS)
-static inline void pgt_put_all(struct pgt_cache *pgt_cache __unused) { }
+static inline void pgt_put_all(struct user_mode_ctx *uctx __unused) { }
 #else
-void pgt_put_all(struct pgt_cache *pgt_cache);
+void pgt_put_all(struct user_mode_ctx *uctx);
```
jforissier, line 82: [commit "core: simplify pgt interface provided by pgt_cache.h"]

The comment needs updating to reflect the new signature.

core/include/mm/pgt_cache.h
```diff
@@ -66,28 +67,23 @@ bool pgt_check_avail(struct pgt_cache *pgt_cache, struct vm_info *vm_info);
...
 #if defined(CFG_CORE_PREALLOC_EL0_TBLS)
-static inline void pgt_get_all(struct pgt_cache *pgt_cache __unused,
-			       struct ts_ctx *owning_ctx __unused,
-			       struct vm_info *vm_info __unused) { }
+static inline void pgt_get_all(struct user_mode_ctx *uctx __unused) { }
 #else
-void pgt_get_all(struct pgt_cache *pgt_cache, struct ts_ctx *owning_ctx,
-		 struct vm_info *vm_info);
+void pgt_get_all(struct user_mode_ctx *uctx);
 #endif
```
jforissier, line 73: [commit "core: simplify pgt interface provided by pgt_cache.h"]

The comment needs updating to reflect the new signature.

jforissier (commented):
"core: remove save_ctx parameter from pgt_free()"
"core: initialize struct user_mode_ctx with vm_info_init()"
"core: call tee_pager_rem_um_regions() from vm_info_final()"
"core: tee_pager.h: provide stubbed tee_pager_pgt_save_and_release_entries()"

"core: move pgt_cache to struct user_mode_ctx"
One question below, then:

"core: call pgt_flush_ctx() from vm_info_final()"
"core: pgt: use pgt_cache_list without pager too"

"core: pgt: rename to pgt_put_all() and pgt_get_all()"
See comments below.

"core: pgt: support preallocated translation tables for S-EL0"
What are the benefits of CFG_CORE_PREALLOC_EL0_TBLS=y? Is it only good for SPs? I tried it with xtest, expecting perhaps some speed improvements, but nothing obvious or perhaps even a slight decrease in performance...
`time xtest` on QEMUv8: with CFG_CORE_PREALLOC_EL0_TBLS=n 2m29.32s / 2m28.31s ; with CFG_CORE_PREALLOC_EL0_TBLS=y 2m34.78s / 2m37.90s

"ci: azure: add build for CFG_CORE_PREALLOC_EL0_TBLS=y"
Once moved to .github/worflows/ci.yml:

"core: simplify pgt interface provided by pgt_cache.h"
One minor comment, then:

core/mm/vm.c
```diff
@@ -1258,20 +1255,24 @@ TEE_Result vm_check_access_rights(const struct user_mode_ctx *uctx,
...
+
+	if (is_user_mode_ctx(tsd->ctx)) {
+		/*
+		 * We're coming from a user mode context so we must make
+		 * the pgts available for reuse.
+		 */
+		uctx = to_user_mode_ctx(tsd->ctx);
+		pgt_free(&uctx->pgt_cache);
+	}
+	tsd->ctx = NULL;
```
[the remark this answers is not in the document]
jenswi-linaro: You're right, I just added it as a safeguard during development. I'll remove it.

core/include/mm/pgt_cache.h
```diff
@@ -53,9 +53,23 @@ SLIST_HEAD(pgt_cache, pgt);
...
+ *
+ * Guaranteed to succeed, but may need to sleep for a while to get all the
+ * needed translation tables.
+ */
+void pgt_get_all(struct pgt_cache *pgt_cache, struct ts_ctx *owning_ctx,
+		 struct vm_info *vm_info);
+/*
+ * pgt_put_all() - informs the translation table manager that these tables
+ *		   will not be needed for while
+ * @pgt_cache:	list of translation table to put to rest
```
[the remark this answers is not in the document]
jenswi-linaro: put to rest, as in inactivate. I'll rephrase it to make it more clear.

core/include/mm/pgt_cache.h
```diff
@@ -53,9 +53,23 @@ SLIST_HEAD(pgt_cache, pgt);
 
 bool pgt_check_avail(struct vm_info *vm_info);
 
-void pgt_alloc(struct pgt_cache *pgt_cache, struct ts_ctx *owning_ctx,
-	       struct vm_info *vm_info);
-void pgt_free(struct pgt_cache *pgt_cache);
```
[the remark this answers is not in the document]
jenswi-linaro: No, those functions are doing what their names suggest.

jenswi-linaro:
Comments addressed, tags applied, rebased on master.

jenswi-linaro:
Note that a few of the commits are missing tags.

jforissier:
I've run a few tests on a RockPi4B board with `CFG_CORE_PREALLOC_EL0_TBLS=n` and `CFG_CORE_PREALLOC_EL0_TBLS=y`. Unfortunately no noticeable speed improvement :cry: 

| Command | n | y |
|-|-|-|
| `time { for i in $(seq 1 100); do taskset -c 5 xtest regression_1011 >/dev/null; done }`  | 0m28.906s, 0m28.393s, 0m27.944s | 0m27.426s, 0m27.801s, 0m28.031s |
| `time { for i in $(seq 1 100); do taskset -c 5 xtest regression_1022 >/dev/null; done }` | 0m15.689s, 0m14.954s, 0m16.568s | 0m15.919s, 0m15.759s, 0m16.052s |
| `time { for i in $(seq 1 1000); do taskset -c 5 xtest pkcs11_1002 >/dev/null; done }` | 0m26.703s, 0m26.317s, 0m26.736s | 0m26.211s, 0m26.273s, 0m26.046s |
| `time taskset -c 5 xtest` | 2m41.418s, 2m43.966s, 2m46.632s | 2m43.293s, 2m45.438s, 2m42.727s |

So except for a few cleanup commits I'm not sure it makes sense to merge this (diffstat +682 -161 so that's a significant amount of code) unless of course we have a proper use case. Thoughts?

jenswi-linaro:
The primary objective with this is not speed improvement but rather guarantees that we can map the S-EL0 context once we've been able to initialize it. This might not be very important for TAs, but is essential for SPs executed at S-EL0 as they can't wait in normal world for translation tables to become available. Systems that are not memory-constrained will also scale much better since we're not limited by `PGT_CACHE_SIZE` any longer.

etienne-lms (commented):
core: remove save_ctx parameter from pgt_free()

core/arch/arm/mm/pgt_cache.c
```diff
@@ -34,12 +37,312 @@
...
+		}
+	}
+prune_done:
+
+	p = SLIST_FIRST(pgt_cache);
+	pp = NULL;
+	TAILQ_FOREACH(r, &vm_info->regions, link) {
+		for (va = ROUNDDOWN(r->va, CORE_MMU_PGDIR_SIZE);
+		     va < r->va + r->size; va += CORE_MMU_PGDIR_SIZE) {
+			if (p  && p->vabase < va) {
```
etienne-lms, line 314: typo: 2 space chars

core/arch/arm/mm/pgt_cache.c
```diff
@@ -204,30 +217,38 @@ static struct pgt *pop_from_cache_list(vaddr_t vabase, void *ctx)
...
+	return 0;
+#endif
+}
+
 static struct pgt *pop_least_used_from_cache_list(void)
 {
 	struct pgt *pgt;
 	struct pgt *p_prev = NULL;
 	size_t least_used;
+	size_t next_used;
```
etienne-lms, line 532: `= 0;` maybe a append fixup commit to the series to fix the 2 missing init values above?

core/arch/arm/mm/pgt_cache.c
```diff
@@ -244,15 +265,7 @@ static void pgt_free_unlocked(struct pgt_cache *pgt_cache)
...
-		if (p->num_used_entries) {
-			push_to_cache_list(p);
-		} else {
-			tee_pager_pgt_save_and_release_entries(p);
-			p->ctx = NULL;
-			p->vabase = 0;
-
-			push_to_free_list(p);
-		}
+		push_to_cache_list(p);
```
etienne-lms, line 566: here pager (with 32bit mmu) seems to treat unused pgts differently eventually releasing rw paged memory. Can the situation happen here (`p->num_used_entries==0`)?

core/mm/vm.c
```diff
@@ -684,16 +761,23 @@ TEE_Result vm_set_prot(struct user_mode_ctx *uctx, vaddr_t va, size_t len,
...
-		vm_set_ctx(uctx->ts_ctx);
+		if (!mobj_is_paged(r->mobj)) {
+			need_sync = true;
+			set_um_region(uctx, r);
+			/*
+			 * Normally when set_um_region() is called we
+			 * change from no mapping to some mapping, but in
+			 * this case we change the permissions on an
+			 * already present mapping so some TLB invalidation
+			 * is needed.  We also depend on the dsb()
```
etienne-lms, line 772: 2 space chars

core/include/mm/pgt_cache.h
```diff
@@ -53,41 +53,36 @@ struct pgt {
...
 
-bool pgt_check_avail(struct pgt_cache *pgt_cache, struct vm_info *vm_info);
+bool pgt_check_avail(struct user_mode_ctx *uctx);
 
 /*
  * pgt_get_all() - makes all needed translation tables available
- * @pgt_cache:	list of translation tables for the owning context
- * @owning_ctx:	the context to own the tables
+ * @uctx:	the context to own the tables
  * @vm_info:	VM map for the context
```
etienne-lms, line 63: commit "core: simplify pgt interface provided by pgt_cache.h":
`vm_info` was also removed from function's params

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for the whole series but commit 'core: pgt: use pgt_cache_list without pager too' where i have a question.

Minor comments here and there.

core/include/mm/pgt_cache.h
```diff
@@ -53,9 +53,23 @@ SLIST_HEAD(pgt_cache, pgt);
...
+ * pgt_get_all() - makes all needed translation tables available
+ * @pgt_cache:	list of translation tables for the owning context
+ * @owning_ctx:	the context to own the tables
+ * @vm_info:	VM map for the context
+ *
+ * Guaranteed to succeed, but may need to sleep for a while to get all the
+ * needed translation tables.
+ */
+void pgt_get_all(struct pgt_cache *pgt_cache, struct ts_ctx *owning_ctx,
+		 struct vm_info *vm_info);
```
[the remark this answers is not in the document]
jenswi-linaro: OK, I'll fix up the commit in place.

core/arch/arm/mm/pgt_cache.c
```diff
@@ -34,12 +37,312 @@
...
+SLIST_HEAD(pgt_parent_list, pgt_parent);
+static struct pgt_parent_list parent_list = SLIST_HEAD_INITIALIZER(parent_list);
+static unsigned int parent_spinlock = SPINLOCK_UNLOCK;
+
+static void free_pgt(struct pgt *pgt)
+{
+	struct pgt_parent *parent = NULL;
+	uint32_t exceptions = 0;
+
+	exceptions = cpu_spin_lock_xsave(&parent_spinlock);
```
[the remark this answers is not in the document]
jenswi-linaro: OK, I'll fix up the commit in place.

core/arch/arm/mm/pgt_cache.c
jenswi-linaro: Before this, it only made sense to cache pgt's where the pager had pages assigned. Now we're going to use this as a way of avoiding to reinitialized the translation tables where possible.

jenswi-linaro:

core/arch/arm/mm/pgt_cache.c
etienne-lms: I agree when pager is off, caching pgt's allows to possibly reuse them, but when pager is enable, we don't have much secure RAM and releasing secure memory can be quite useful for smooth optee execution. Pgt's are pager locked rw pages. With this change they remain paged until the TA context is released. This could prevent paged op-tee from executing several TAs  or make pager very inefficient because of limited number of available physical pages in the pager pool.

jenswi-linaro: They can still be re-used, it's just that they need to be picked up from `pgt_cache_list` instead of `pgt_free_list`. `pop_from_some_list()` handles this.

etienne-lms: The previous implementation released the pgt's rw locked pages once TA invocation completes. With the new implementation (this P-R), the pgt's rw locked pages are still in place, in pgt cache list, until the TA context dies, hence lowering the number of available physical pages for the service next invocation. It is a matter of balance I think on, when pager is enable, whether we prefer to speed up time to setup back TA mapping, or spend more time but relax the pager pool pressure.

I've tested this P-R against `xtest regression_1005` that attempts the created many sessions on a multi instance TA. I've increased the number of opened sessions from 3 to 10. I've compared the behavior with current change and with the hack below to restore release of pager locked pgt's from `pgt_free_unlocked()`:
```patch
 	while (!SLIST_EMPTY(pgt_cache)) {
 		struct pgt *p = SLIST_FIRST(pgt_cache);
 
 		SLIST_REMOVE_HEAD(pgt_cache, link);
 
+#ifdef CFG_PAGED_USER_TA
+		if (p->num_used_entries) {
+			/* Release paged locked PGT's now from pager */
+			tee_pager_pgt_save_and_release_entries(p);
+			p->ctx = NULL;
+			p->vabase = 0;
+
+			push_to_free_list(p);
+			continue;
+		}
+#endif
 		push_to_cache_list(p);
 	}
```
On this very test and on my paged platform, releasing pgt's on TA context destruction (the P-R as-is) rather than on TA invocation completion (P-R + the above patch) makes this test about 25% slower. Tracking the number of unlocked pages in page pool, I see its a it about 20% higher with the above patch applied (because pgts are released).

I am thinking also of keep alive TAs that, IIUC, will have their cache pgt's assigned forever, preventing pager from using the related rw lock pages.

Maybe I missing something but high pressure on pager pool can make OP-TEE very very slow to operate.

jenswi-linaro: Thanks, I see the problem now.

jenswi-linaro: In your diff above you're freeing pages more aggressively (all physical pages referred by this table too) than before this PR.

I've added an update restoring the old behavior so it's slightly different from what you suggested above.

jenswi-linaro:
I'll rebase as soon as we have agreed on the updates to "core: pgt: use pgt_cache_list without pager too"

etienne-lms:
Thanks @jenswi-linaro for addressing my concern. Your fixup looks good to me (but that is fixes commit "core: pgt: use pgt_cache_list without pager too").
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for that commit with its fix.

jenswi-linaro:
Squashed, rebased and tag applied.

PR #5410, opened 2022-06-29
plat-stm32mp1: restore reserved shared memory

etienne-lms:
Restores OP-TEE reserved shared memory (the last MBytes of DRAM located
on top of the secure DDR) in STM32MP15 default configuration. This
default configuration is needed to support latest mainline TF-A in conjunction
with mainline Linux kernel.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -31,7 +31,6 @@ endif
 
 include core/arch/arm/cpu/cortex-a7.mk
 
-$(call force,CFG_CORE_RESERVED_SHM,n)
 $(call force,CFG_BOOT_SECONDARY_REQUEST,y)
 $(call force,CFG_DRIVERS_CLK,y)
 $(call force,CFG_DRIVERS_CLK_FIXED,n)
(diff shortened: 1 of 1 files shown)


GseoC (approved):
Acked-by: Gatien Chevallier <gatien.chevallier@foss.st.com>

jenswi-linaro:
I assume that the kernel is told to stay out of this memory in the DTB. Does this mean that with this you somehow patch the DTB passed to the kernel, but still initializes with dynamic shared memory in the driver or the kernel would use this memory as shared memory?

etienne-lms:
Hi @jenswi-linaro. Platform stm32mp1 uses the generic scheme where OP-TEE adds optee reserved-memory nodes if optee node not found in the DT.

In the OP-TEE distrib (mainfest/build), the linux fork includes a patch in Linux DTS to define OP-TEE nodes.
On the other side, mainline Linux source tree does not include these information by legacy, so we need optee to adds thoses nodes in the DTB when optee is booted (for consistency between tzc400 config and DTB reserved-memory description).

Plan is to have tf-a/linux/u-boot mainline DTSs to not use rsv-shm and define the optee reserved-memory node for TZDRAM secure range. This should come with kernel v5.19. Until we have release tags for TF-A/U-Boot/Linux that allows optee_os/stm32mp1 default config to disable rsv-shm, let's have it enable by default.

I would like this to hit v3.18.0.

etienne-lms:
@GseoC tag applied.

jenswi-linaro:
Thanks for the explanation.

PR #5415, opened 2022-06-30
SPMC: Enable handling 64-bit direct messages

jellesels-arm:
Enable using the 64-bit version of the FF-A direct request and response
calls.


jenswi-linaro:

etienne-lms (commented):
In commit message header line: prefer use prefix `core: ffa: ...` or `core: spmc: ...` rather than `SPMC: ...` for consistency with other commit messages related to FFA and SPMC in revision history.

jforissier:
@jellesels-arm please, no blank line between tags (R-b:, A-b:, S-o-b:). There should be one blank line after the description, then all the tags. Thanks.

jellesels-arm:
> @jellesels-arm please, no blank line between tags (R-b:, A-b:, S-o-b:). There should be one blank line after the description, then all the tags. Thanks.

Done

Issue #5417, opened 2022-06-30
ASLR panic on boot with pager


jforissier:
I got the following crash on boot on QEMUv8 with `CFG_WITH_PAGER=y`:
```
E/TC:0 0 Panic 'can't find mmu tables' at core/arch/arm/mm/tee_pager.c:549 <tee_pager_early_init>
E/TC:0 0 TEE load address @ 0x50b9000
E/TC:0 0 Call stack:
E/TC:0 0  0x050bf144
E/TC:0 0  0x050c2ff4
E/TC:0 0  0x050f0d78
E/TC:0 0  0x050f0284
```
The problem is not easily reproducible because it seems related to ASLR and more specifically the random seed value. From the load address above I figured out the seed value and indeed the following patch reproduces the issue:
```diff
diff --git a/core/mm/core_mmu.c b/core/mm/core_mmu.c
index 2bd4d38c..896ca4be 100644
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -1207,6 +1207,8 @@ static unsigned long init_mem_map(struct tee_mmap_region *memory_map,
        unsigned long offs = 0;
        size_t last = 0;

+       seed = 0xf6fb9000UL;
+
        last = collect_mem_ranges(memory_map, num_elems);
        assign_mem_granularity(memory_map);

```

jenswi-linaro:
I'll take a look at this.

PR #5420, opened 2022-06-30
Aslr fix

jenswi-linaro:
Fixes https://github.com/OP-TEE/optee_os/issues/5417


jforissier (approved):
That was fast! :wink:

`Acked-by: Jerome Forissier <jerome.forissier@linaro.org>` (for both commits)
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (vexpress-qemu_armv8a)` (for the last one)

Thanks.

jenswi-linaro:
Yeah, it helps quite a lot when it's easy so easy to reproduce.
Tags applied.

core/mm/core_mmu.c
```diff
@@ -2060,11 +2060,16 @@ void *core_mmu_add_mapping(enum teecore_memtypes type, paddr_t addr, size_t len)
 }
 
 #ifdef CFG_WITH_PAGER
-static vaddr_t get_linear_map_end(void)
+static paddr_t get_linear_map_end_va(void)
```
etienne-lms, line 2063: s/paddr_t/vaddr_t/

core/mm/core_mmu.c
```diff
@@ -2060,11 +2060,16 @@ void *core_mmu_add_mapping(enum teecore_memtypes type, paddr_t addr, size_t len)
...
 {
 	/* this is synced with the generic linker file kern.ld.S */
 	return (vaddr_t)__heap2_end;
 }
+
+static paddr_t get_linear_map_end_pa(void)
+{
+	return get_linear_map_end_va() - VCORE_START_VA + TEE_LOAD_ADDR;
```
etienne-lms, line 2071: cast?
`return  (paddr_t)(get_linear_map_end_va() - VCORE_START_VA + TEE_LOAD_ADDR);`

etienne-lms (changes requested):
typo in function return type.

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` once addressed.

core/mm/core_mmu.c
jenswi-linaro: I'll fix.

core/mm/core_mmu.c
jenswi-linaro: I'm sorry, I can't see the point with that.

jenswi-linaro:
tag applied.

core/mm/core_mmu.c
etienne-lms: do we need here to explicitly cast from `vaddr_t` to `paddr_t`?

jenswi-linaro: I'm not sure what would be gained by that. The type is already converted automatically.

etienne-lms: ok, fine.

PR #5427, opened 2022-07-07
core: plat-ls: set CFG_NUM_THREADS=2 in case of LS1012ARDB

sahilnxp:
xtest regression_1009.3 was failing on LS1012ARDB, so setting
CFG_NUM_THREADS=2 in case of ls1012ardb.

### DIFF
--- a/core/arch/arm/plat-ls/conf.mk
+++ b/core/arch/arm/plat-ls/conf.mk
@@ -33,6 +33,7 @@ include core/arch/arm/cpu/cortex-armv8-0.mk
 $(call force,CFG_TEE_CORE_NB_CORE,1)
 $(call force,CFG_DRAM0_SIZE,0x40000000)
 $(call force,CFG_CORE_CLUSTER_SHIFT,2)
+CFG_NUM_THREADS ?= 2
 CFG_SHMEM_SIZE ?= 0x00200000
 endif
 


core/arch/arm/plat-ls/conf.mk
```diff
@@ -33,6 +33,7 @@ include core/arch/arm/cpu/cortex-armv8-0.mk
 $(call force,CFG_TEE_CORE_NB_CORE,1)
 $(call force,CFG_DRAM0_SIZE,0x40000000)
 $(call force,CFG_CORE_CLUSTER_SHIFT,2)
+$(call force,CFG_NUM_THREADS,2)
```
jforissier, line 36: What is the default value without this change? I guess 1 but I'm not sure, especially since the below line: `CFG_NUM_THREADS ?= CFG_TEE_CORE_NB_CORE` is wrong: missing `$()` -- to be fixed in a separate patch?

I believe regression_1009 expects `CFG_NUM_THREADS >= 2` indeed, but *forcing* 2 might be a bit too strong. How about writing `CFG_NUM_THREADS ?= 2` instead? Then if one wants to build with 3 or 4 or whatever for special use cases, it is easier.

sahilnxp:
> I guess you'll face the same problem with the other flavors with `CFG_TEE_CORE_NB_CORE=1`.

Yes, that is only ls1012afrwy and we are not supporting OP-TEE on ls1012afrwy board.
I will be raising the pull request for removal of its support.

core/arch/arm/plat-ls/conf.mk
sahilnxp: Yes, It is set as 1 by default without this change.

_CFG_NUM_THREADS ?= CFG_TEE_CORE_NB_CORE is wrong: missing $() -- to be fixed in a separate patch?_
Yeah, I didn't noticed that, will fix in a separate patch.


_I believe regression_1009 expects CFG_NUM_THREADS >= 2 indeed, but forcing 2 might be a bit too strong. How about writing CFG_NUM_THREADS ?= 2 instead? Then if one wants to build with 3 or 4 or whatever for special use cases, it is easier._
Yes, your point is valid, but what if person gave CFG_NUM_THREADS as 1, this test will again fail.
That's why forced it to 2, but I think we are in a way by setting it as CFG_NUM_THREADS ?= 2 indicating that we need at least 2 threads.
Will do the change.

jforissier:


Commit descriptions are preferably written in the present tense to describe the situation before the commit, and the imperative mood to describe the change. May I suggest:

```diff
-  core: plat-ls: set CFG_NUM_THREADS=2 in case of LS1012ARDB
- 
- xtest regression_1009.3 was failing on LS1012ARDB, so setting
- CFG_NUM_THREADS =? 2 in case of ls1012ardb.
- Need atleast 2 threads for proper functioning of ls1012ardb.
+ core: plat-ls: set CFG_NUM_THREADS ?= 2 for LS1012ARDB
+
+ xtest regression_1009.3 fails on LS1012ARDB because the test needs
+ at least two threads but the default configuration for the platform
+ enables only one. Set CFG_NUM_THREADS =? 2 to fix the issue.
```

jenswi-linaro:

sahilnxp:
Thanks @jenswi-linaro, tag updated.

jforissier:
Merged manually with typo fixed (`=?`). Thanks.

sahilnxp:
> Merged manually with typo fixed (`=?`). Thanks.

Thanks @jforissier

PR #5431, opened 2022-07-11
core: arm: kernel: fix rpc shm free when prealloc is disable

etienne-lms:
This change fixes an issue found in tag 3.17.0.

> Fixes commit [1] that changed implementation of the standard invocation
> SMC command to introduce invocation with RPC shared memory reference.
> A wrong logic was implemented to free RPC buffer on standard invocation
> completion. This change fixes that by freeing the shared memory when
> pre-allocation is disable by config switch or runtime service, not when
> it is enable.
>
>  Fixes: feb290a51087 ("core: add OPTEE_SMC_CALL_WITH_RPC_ARG")

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -133,14 +133,12 @@ static uint32_t get_msg_arg(struct mobj *mobj, size_t offset,
 	return OPTEE_SMC_RETURN_OK;
 }
 
-static void maybe_clear_prealloc_rpc_cache(struct thread_ctx *thr)
+static void clear_prealloc_rpc_cache(struct thread_ctx *thr)
 {
-	if (IS_ENABLED(CFG_PREALLOC_RPC_CACHE) && thread_prealloc_rpc_cache) {
-		thread_rpc_free_arg(mobj_get_cookie(thr->rpc_mobj));
-		mobj_put(thr->rpc_mobj);
-		thr->rpc_arg = NULL;
-		thr->rpc_mobj = NULL;
-	}
+	thread_rpc_free_arg(mobj_get_cookie(thr->rpc_mobj));
+	mobj_put(thr->rpc_mobj);
+	thr->rpc_arg = NULL;
+	thr->rpc_mobj = NULL;
 }
 
 static uint32_t call_entry_std(struct optee_msg_arg *arg, size_t num_params,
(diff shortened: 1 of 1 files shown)


etienne-lms:
Thanks @jenswi-linaro. Indeed I miss the passed rpc ref case.

etienne-lms:
(fixed missing S-o-b tag)

core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -157,9 +157,7 @@ static uint32_t call_entry_std(struct optee_msg_arg *arg, size_t num_params,
 		 * struct. But if it is we must use the supplied struct and
 		 * at the same time make sure to not break anything.
 		 */
-		if (IS_ENABLED(CFG_PREALLOC_RPC_CACHE) |
-		    thread_prealloc_rpc_cache)
-			clear_prealloc_rpc_cache(thr);
+		clear_prealloc_rpc_cache(thr);
```
jenswi-linaro, line 160: I'm not so keen on doing this unconditionally. This if-case is our new optimized common path and that shouldn't do more than what's needed.

core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -172,8 +170,7 @@ static uint32_t call_entry_std(struct optee_msg_arg *arg, size_t num_params,
 	if (rpc_arg)
 		thr->rpc_arg = NULL;
 
-	if (rv == OPTEE_SMC_RETURN_OK &&
-	    !(IS_ENABLED(CFG_PREALLOC_RPC_CACHE) && thread_prealloc_rpc_cache))
+	if (rv == OPTEE_SMC_RETURN_OK && !thread_prealloc_rpc_cache)
```
jenswi-linaro, line 173: With this change it looks like it might be harder for the compiler to determine that `thread_prealloc_rpc_cache` is unused with `CFG_PREALLOC_RPC_CACHE=n`.
If you think the if-statement needs to be simplified feel free to add a
``` C
bool prealloc_rpc_cache_is_eanbled()

**Review by jenswi-linaro — COMMENTED** [review]
Comments for the "simplification".

File: core/arch/arm/kernel/thread_optee_smc.c [hunk 3]
```diff
@@ -159,7 +157,9 @@ static uint32_t call_entry_std(struct optee_msg_arg *arg, size_t num_params,
 		 * struct. But if it is we must use the supplied struct and
 		 * at the same time make sure to not break anything.
 		 */
-		maybe_clear_prealloc_rpc_cache(thr);
+		if (IS_ENABLED(CFG_PREALLOC_RPC_CACHE) |
```
jenswi-linaro: `s/|/||/`

etienne-lms:
fixup commits squashed and tag applied.

PR #5439, opened 2022-07-14
libutee: Add more TEE_MALLOC hints

gagachang:
This patch adds the definitions and related code with the hint
values: TEE_MALLOC_NO_SHARE and TEE_MALLOC_NO_FILL. They were
added from TEE Internal Core API v1.2.

Also, TEE_MALLOC_NO_FILL SHALL be used in conjunction with
TEE_MALLOC_NO_SHARE. If user uses it in wrong way, a TEE panic is
triggered.

The original TEE_USER_MEM_HINT_NO_FILL_ZERO is preserved for
backward compatibility.

Signed-off-by: Che-Chia Chang <alvinga@andestech.com>

### DIFF
--- a/lib/libutee/include/tee_api_defines.h
+++ b/lib/libutee/include/tee_api_defines.h
@@ -83,6 +83,8 @@
 
 /* Memory Management Constant */
 #define TEE_MALLOC_FILL_ZERO               0x00000000
+#define TEE_MALLOC_NO_FILL                 0x00000001
+#define TEE_MALLOC_NO_SHARE                0x00000010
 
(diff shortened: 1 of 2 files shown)


jforissier:
Maybe add to the commit description that `TEE_MALLOC_NO_SHARE` does nothing special?
Also in the subject `s/TEE_MALLOC/TEE_Malloc()/`.

gagachang:
Yes you guys are right.
Since I did not implement the features about taging on allocated memory, I close this PR.
Thanks for hint.

PR #5444, opened 2022-07-18
libutils: util.h: add get_field_u{32,64}() and set_field_u{32,64}()

maroueneboubakri:
This commit defines `get_field_u{32,64}()` and `set_field_u{32,64}()` macros for respectively getting and setting bit fields.

### DIFF
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -172,6 +172,27 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg0 = val >> 32;
 	*reg1 = val;
 }
+
+/* Get and set bit fields  */
+static inline uint32_t get_field_u32(uint32_t reg, uint32_t mask)
+{
+	return (reg & mask) / (mask & ~(mask - 1));
+}
+
+static inline uint32_t set_field_u32(uint32_t reg, uint32_t mask, uint32_t val)
+{
+	return (reg & ~mask) | (val * (mask & ~(mask - 1)));
+}
+
+static inline uint64_t get_field_u64(uint64_t reg, uint64_t mask)
+{
+	return (reg & mask) / (mask & ~(mask - 1));
+}
+
+static inline uint64_t set_field_u64(uint64_t reg, uint64_t mask, uint64_t val)
+{
+	return (reg & ~mask) | (val * (mask & ~(mask - 1)));
+}
 #endif
 
 #endif /*UTIL_H*/


etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with indentation fixed.

lib/libutils/ext/include/util.h
```diff
@@ -140,6 +140,21 @@
 #define GENMASK_64(h, l) \
 	(((~UINT64_C(0)) << (l)) & (~UINT64_C(0) >> (64 - 1 - (h))))
 
+/* Get and set bit fields  */
+#define GET_FIELD_U32(reg, mask) \
+	(((reg) & UINT32_C(mask)) / ((mask) & ~((mask) << 1)))
```
jenswi-linaro, line 145: Requiring `mask` to be a constant instead may make this macro somewhat less useful.
The other uses of `mask` should also be made into an unsigned 32-bit integer. If we don't care about assembly it's probably easier to rewrite it as (with the bonus that `mask` is only evaluated once):
``` C
#define GET_FIELD_U32(reg, mask) \
        (__extension__({ \
                uint32_t _mask  = (mask); \
                \
        ￼       ((reg) & _mask) / (_mask & ~SHIFT_U32(_mask, 1); \
        }))

File: lib/libutils/ext/include/util.h [hunk 2]
```diff
@@ -140,6 +140,41 @@
...
+
+#define SET_FIELD_U64(reg, mask, val) \
+        (__extension__({ \
+                uint64_t _mask  = (mask); \
+				uint64_t _val  = (val); \
+				\
+               ((reg & ~_mask) |\
+	           ((_val * (_mask & ~SHIFT_U64(_mask, 1))) & _mask)); \
+	    }))
+
```
jforissier, line 177: You're mixing tabs and spaces :-/

In fact this looks unnecessarily complex... how about a static inline function instead? (and same for the other ones of course).

```C
static inline uint64_t SET_FIELD_U64(uint64_t reg, uint64_t mask, uint64_t val)          
{                                                                               
        return (reg & ~mask) | ((val * (mask & ~SHIFT_U64(mask, 1))) & mask);   
}
```

maroueneboubakri: Comment addressed.

jenswi-linaro:

jforissier (commented):

maroueneboubakri:

jforissier:
Thanks @maroueneboubakri, I'd like confirmation from @etienne-lms that he's still OK with applying his R-b since the code was changed substantially.

etienne-lms (approved):
Hi, LGTM, my R-b applies.

PR #5451, opened 2022-07-25
ci: migrate from Azure pipelines to GitHub actions

jforissier:
The Azure CI recently stopped working with frequent "no space left on
device" errors when fetching the Docker image used by the QEMUv8_check
and QEMUv8_build_Rust jobs [1]. Take this opportunity to migrate to
GitHub Actions which has similar or better capabilities (better
integration with GitHub UI, no need for an Azure account or project,
built-in caching capability for build artifacts, more concurrent jobs
and longer timeouts [2]).

No functional change except for cosmetic things (job names) and the
handling of the build cache in the multi-platform build job. Previously,
files could be uploaded to the cache server before the end of the
script. As a result, if a build would time out it would usually be
enough to just restart it, then the new run would obtain files from the
previous run and be fast enough to complete. With the GitHub cache
action used here, this doesn't happen. Data are only uploaded to the
cache when the build completes with no error. It means some manual work
may be needed if the maximum build time is reached: push a truncated
build job to a temporary branch, wait for it to complete, then restart
the timed out job. However, the full build currently takes about 1 hour
with an empty cache and the maximum build time is 6 hours so it should
not be a problem anytime soon.

Link: [1] https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=3395&view=logs&j=4b042fc3-edf6-5596-582e-7ecd0cce7842&t=9368a7c9-f53c-4bb5-bbbf-b04d402537af&l=53
Link: [2] https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,14 +1,24 @@
+name: CI
+on: [push, pull_request]
 jobs:
-  - job: code_style
-    displayName: 'Check code style'
-    pool:
-      vmImage: ubuntu-18.04
-    container:
-      image: jforissier/optee_os_ci
+  code_style:
+    name: Code style
+    runs-on: ubuntu-latest
+    container: jforissier/optee_os_ci
     steps:
-      - script: |
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 50 # >= the highest number of commits expected in a single PR
+      - name: Update Git config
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

jforissier:
Thanks @jenswi-linaro. I have added caching to the QEMUv8_check job too, saves a few minutes (not much, most is execution time).

PR #5456, opened 2022-07-27
Rework i.MX MU driver for imx8ulp support

clementfaure:
Hello,

The imx8ulp features an MU to communicate with the external security controller.
This MU is slightly different than the imx8qm/qxp MU.
Rework the MU driver and add a HAL for all supported platforms.

I will upstream the security enclave driver for imx8ulp in another PR to keep this one small.

Thanks!


jenswi-linaro:
You already have a `core/drivers/imx` directory, please use that instead.
You're adding yet another global include directory, I assume that it's only for internal shared .h files but I would have preferred if we wouldn't need a new directory each time this is needed.
Having a file to include as `#include <local.h>` doesn't seem right. How about naming it imx_mu_private.h or something?
I'm not so keen on calling a driver HAL, all drivers are hardware abstraction layers. Please give it a proper name instead.

clementfaure:
Ok, I will reorganize things differently.
Regarding the HAL, I mean, a driver could have several level of abstraction (IP, platform, ..). We did something similar with the CAAM. I will come up with another naming instead of `hal` if that's your preference.

clementfaure:
If you don't mind, I will keep `_hal_` keyword for platform specific functions like you did for the CAAM.

core/drivers/imx/mu/imx_mu.c
```diff
@@ -0,0 +1,150 @@
...
+	unsigned int nb_channel = 0;
+
+	assert(base && msg);
+
+	if (msg->header.size > IMX_MU_MSG_SIZE) {
+		EMSG("msg->size is > than IMX_MU_MSG_SIZE");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	res = imx_mu_hal_send(base, 0, *(uint32_t *)msg);
```
etienne-lms, line 100: This cast looks a bit strange. One need to look into `struct imx_mu_msg` to understand the cast.
I would suggest to use an local helper function like `uint32_t get_32bit_header(struct imx_mu_msg *msg)` to do the trick. Or maybe define `struct imx_mu_msg_header` as an union of a `uint32_t` cell and a 4 `uint8_t` cells struct.

core/drivers/imx/mu/imx_mu_8q.c
```diff
@@ -33,15 +38,27 @@ static TEE_Result mu_wait_for(vaddr_t addr, uint32_t mask)
...
-					   MU_CR_F_MASK);
+	return MU_MAX_CHANNEL;
 }
 
-TEE_Result mu_send_msg(vaddr_t base, unsigned int index, uint32_t msg)
+void imx_mu_hal_init(vaddr_t base)
 {
+	io_clrbits32(base + MU_ACR_OFFSET,
+		     MU_CR_GIE_MASK | MU_CR_RIE_MASK | MU_CR_TIE_MASK |
+			     MU_CR_GIR_MASK | MU_CR_F_MASK);
```
etienne-lms, line 55: indentation:
```c
	io_clrbits32(base + MU_ACR_OFFSET,
		     MU_CR_GIE_MASK | MU_CR_RIE_MASK | MU_CR_TIE_MASK |
		     MU_CR_GIR_MASK | MU_CR_F_MASK);
```

core/drivers/imx/mu/imx_mu_8ulp.c
```diff
@@ -0,0 +1,77 @@
...
+#define MU_MAX_RX_CHANNEL 4
+#define MU_MAX_TX_CHANNEL 8
+
+static TEE_Result mu_wait_for(vaddr_t addr, uint32_t mask)
+{
+	uint64_t timeout = timeout_init_us(1000);
+
+	while (!(io_read32(addr) & mask)) {
+		if (timeout_elapsed(timeout))
+			return TEE_ERROR_BUSY;
```
etienne-lms, line 30: with a 1ms timeout, if this is executed in a thread context, I strongly recommend to prevent from spurious timeout reports due to thread scheduling delays in the system:
```c
	while (!(io_read32(addr) & mask))
		if (timeout_elapsed(timeout))
			break;

	if (io_read32(addr) & mask)
		return TEE_SUCCESS;
	else
		return TEE_ERROR_BUSY;
```

core/drivers/imx/mu/imx_mu_8ulp.c
```diff
@@ -0,0 +1,77 @@
...
+#include "imx_mu_common.h"
+
+#define MU_TCR		  0x120
+#define MU_TSR		  0x124
+#define MU_RCR		  0x128
+#define MU_RSR		  0x12C
+#define MU_TRn(n)	  (0x200 + 0x4 * (n))
+#define MU_RRn(n)	  (0x280 + 0x4 * (n))
+#define MU_TSR_TEn(n)	  SHIFT_U32(0x1, (n))
+#define MU_RSR_RFn(n)	  SHIFT_U32(0x1, (n))
```
etienne-lms, line 20: `BIT(n)`?

By the way, really prefer full uppercase names for macros. Could these be:
```c
#define MU_TSR_TE(n)	  BIT32(n)
#define MU_RSR_RF(n)	  BIT32(n)
```
?

core/arch/arm/plat-imx/conf.mk
```diff
@@ -214,6 +214,8 @@ CFG_DRAM_BASE ?= 0x80000000
 CFG_TEE_CORE_NB_CORE ?= 2
 $(call force,CFG_NXP_SNVS,n)
 $(call force,CFG_IMX_OCOTP,n)
+$(call force,CFG_SC_IPC_BASE,SC_IPC0_BASE)
```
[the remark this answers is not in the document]
clementfaure: I will remove it

core/drivers/imx/mu/imx_mu.c
clementfaure: I agree, this is a bit strange.
We could also do a `memcpy(&word, &msg->header, sizeof(uint32_t));`

core/drivers/imx/mu/imx_mu_8q.c
clementfaure: Ok

core/drivers/imx/mu/imx_mu_8ulp.c
clementfaure: Ok

core/drivers/imx/mu/imx_mu_8ulp.c
clementfaure: That makes sense

jenswi-linaro:
> If you don't mind, I will keep `_hal_` keyword for platform specific functions like we did for the CAAM.

I think that was a mistake as "hal" says absolutely nothing about the function. I'd prefer to replace "hal" with "plat" if there's not a better alternative.

jenswi-linaro:
The change looks good to me.

clementfaure:
@jenswi-linaro Great! Let me know if you have more comments

jenswi-linaro:
Please squash the fixup and I'll go through it again.

clementfaure:
Force-push : Rebase and conflict resolution

core/drivers/imx/mu/imx_mu.c
```diff
@@ -0,0 +1,152 @@
...
+#include <kernel/delay.h>
+#include <kernel/spinlock.h>
+#include <string.h>
+#include <trace.h>
+
+#include "imx_mu_platform.h"
+
+#define RX_TIMEOUT (100 * 1000)
+
+static unsigned int s_mu_spinlock = SPINLOCK_UNLOCK;
```
jenswi-linaro, line 16: What's the story with the `s_` prefix? If it's to some variant of coding attributes into the name of the variable please drop it.

jenswi-linaro (commented):

core/drivers/imx/mu/imx_mu.c
clementfaure: Ok, I will remove it

clementfaure:
Could this PR be merged?

clementfaure:
Thanks!

PR #5459, opened 2022-07-29
plat-stm32mp1: add STM32MP157C based DHCOM SoM on PDK2 baseboard

jneuhauser:
The dts(i) files are imported from Linux 5.19-rc6.

Changes made to the imported dts(i) files:
- Drop GPLv2 licensed resources using comments or their explicit values
- Drop cryp1 okay status as on other boards
- Drop unsupported special rcc clocks definition using comments
- Enable rcc as on other boards
- Set board_id as non secure otp
- Allow iwdg2 for usage in non-secure world

Signed-off-by: Johann Neuhauser <jneuhauser@dh-electronics.com>

----------------------------------------

Ping @etienne-lms as stm32mp1 platform maintainer.

To support the special rcc clock definition, we need support that a node can be its own clock provider. So we comment this for now and I'll open a issue to discuss how this should be implemented/supported.

### DIFF
--- a/core/arch/arm/dts/stm32mp157c-dhcom-pdk2.dts
+++ b/core/arch/arm/dts/stm32mp157c-dhcom-pdk2.dts
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: GPL-2.0+ OR BSD-3-Clause
+/*
+ * Copyright (C) 2019-2020 Marek Vasut <marex@denx.de>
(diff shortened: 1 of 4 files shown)


jneuhauser:
Close this pull request in favor of https://github.com/OP-TEE/optee_os/pull/5479 with an additional board as second commit.

PR #5460, opened 2022-07-29
drivers: xiphera_trng: Allow interrupts while waiting for random

vesajaaskelainen:
If for some reason getting new random values take a bit longer toggle
interrupt masks on/off while waiting.

This allows pending interrupts to be served faster in REE side as getting
new random might not be that important. At the same time it gives more
time for random number generation to complete and not just performing
spinning and waiting.

It was originally recommended by TRNG vendor not to cache previous partial
results in memory. TRNG itself is rather fast so there should
be always fresh bytes available for consumption. Thus to simplify the code
remove the FIFO construct.

### DIFF
--- a/core/drivers/xiphera_trng.c
+++ b/core/drivers/xiphera_trng.c
@@ -37,30 +37,27 @@ static unsigned int trng_lock = SPINLOCK_UNLOCK;
 
 static vaddr_t xiphera_trng_base;
 
+static bool xiphera_trng_random_available(void)
+{
+	uint32_t status = 0;
+
+	status = io_read32(xiphera_trng_base + STATUS_REG);
+
+	return status == TRNG_NEW_RAND_AVAILABLE;
+}
+
 static uint32_t xiphera_trng_read32(void)
 {
 	uint32_t value = 0;
-	uint32_t exceptions = 0;
-	uint32_t status = 0;
 
-	while (true) {
-		/* Wait until we have value available */
-		status = io_read32(xiphera_trng_base + STATUS_REG);
-		if (status != TRNG_NEW_RAND_AVAILABLE)
-			continue;
+	value = io_read32(xiphera_trng_base + RAND_REG);
 
-		value = io_read32(xiphera_trng_base + RAND_REG);
-
-		/*
-		 * Ack that RNG value has been consumed and trigger new one to
-		 * be generated
-		 */
-		io_write32(xiphera_trng_base + CONTROL_REG, HOST_TO_TRNG_READ);
-		io_write32(xiphera_trng_base + CONTROL_REG,
-			   HOST_TO_TRNG_ENABLE);
-
-		break;
-	}
+	/*
+	 * Ack that RNG value has been consumed and trigger new one to be
+	 * generated
+	 */
+	io_write32(xiphera_trng_base + CONTROL_REG, HOST_TO_TRNG_READ);
+	io_write32(xiphera_trng_base + CONTROL_REG, HOST_TO_TRNG_ENABLE);
 
 	return value;
 }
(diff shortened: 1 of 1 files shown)


vesajaaskelainen:
@etienne-lms were you thinking something like this?

vesajaaskelainen:
Note: this also fixes compiler warning about unused variable `exceptions' which was introduced in random number generation re-factoring.

@jforissier Wondering should these kinda of optional drivers be included in some configuration in CI so that they get checked out?

jforissier:
> @jforissier Wondering should these kinda of optional drivers be included in some configuration in CI so that they get checked out?

Yes, that would be a good thing. Luckily with the migration from Azure to GitHub Actions, the build step has sped up a bit (from ~35 to ~20 minutes), so we have some leeway here.

vesajaaskelainen:
> > @jforissier Wondering should these kinda of optional drivers be included in some configuration in CI so that they get checked out?
> 
> Yes, that would be a good thing. Luckily with the migration from Azure to GitHub Actions, the build step has sped up a bit (from ~35 to ~20 minutes), so we have some leeway here.

Created PR about that:
https://github.com/OP-TEE/optee_os/pull/5462

core/drivers/xiphera_trng.c
```diff
@@ -37,30 +37,31 @@ static unsigned int trng_lock = SPINLOCK_UNLOCK;
...
-			   HOST_TO_TRNG_ENABLE);
+static uint32_t xiphera_trng_read32(void)
+{
+	uint32_t value = 0;
 
-		break;
-	}
+	/* Sanity check that we actually have valid RNG available */
+	if (!xiphera_trng_random_available())
+		panic("RNG access error");
```
jforissier, line 55: IIUC, it is possible for `xiphera_trng_random_available()` to return false more or less at anytime (given the read loop that restores interrupts below). So how is it guaranteed that this particular call cannot fail?

vesajaaskelainen: Looked at the VHDL for CPU interface part (other part is protected). There is a state machine in here. I don't think this can never fail. Will take it out so there is no sanity check for it.

vesajaaskelainen: Removed and updated the PR.

core/drivers/xiphera_trng.c
```diff
@@ -85,6 +82,12 @@ TEE_Result hw_get_random_bytes(void *buf, size_t len)
 	while (buffer_pos < len) {
 		uint32_t exceptions = cpu_spin_lock_xsave(&trng_lock);
 
+		/* Wait until we have value available */
+		if (!xiphera_trng_random_available()) {
+			cpu_spin_unlock_xrestore(&trng_lock, exceptions);
```
jenswi-linaro, line 87: I'd prefer matching each lock with a single unlock when possible to make it easier to read the code.

core/drivers/xiphera_trng.c
```diff
@@ -85,6 +82,12 @@ TEE_Result hw_get_random_bytes(void *buf, size_t len)
 	while (buffer_pos < len) {
 		uint32_t exceptions = cpu_spin_lock_xsave(&trng_lock);
 
+		/* Wait until we have value available */
+		if (!xiphera_trng_random_available()) {
```
jenswi-linaro, line 86: This check is only needed if `xiphera_trng_read32()` is going to be called below.

vesajaaskelainen: Right. Let's see what can be done for that.

core/drivers/xiphera_trng.c
vesajaaskelainen: Ok. I'll see how that could look.

vesajaaskelainen: This is now changed. Actually dropped the whole FIFO construct at the same time as it simplified it quite a bit with this construct.

Originally we were advised by TRNG vendor not to cache previous results if everything is not consumed. So this returns the original approach.

core/drivers/xiphera_trng.c
vesajaaskelainen: Addressed.

vesajaaskelainen:
Rebased and updated the PR.

glneo:
> TRNG itself is rather fast so there should be always fresh bytes available for consumption.

So which is it? Will we need to re-enable interrupts while while waiting if we don't plan on needing to wait?

My only concern with this patch is that it undoes the result caching, which seem unnecessary. But also it makes it harder to move to the newer driver style presented here: #5492 . After that goes in, then we get the benefit of this PR for free. All drivers behave asynchronously and do not poll for entropy, preventing long interrupt locking.

vesajaaskelainen:
Hmm... seems that my comment has been lost :|

Here is the some public details of the TRNG core and its performace:
https://xiphera.com/products/pdf/XIP8001B_PB.pdf

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with a minor comment.

core/drivers/xiphera_trng.c
```diff
@@ -37,30 +37,27 @@ static unsigned int trng_lock = SPINLOCK_UNLOCK;
 
 static vaddr_t xiphera_trng_base;
 
-static uint32_t xiphera_trng_read32(void)
+static bool xiphera_trng_random_available(void)
 {
-	uint32_t value = 0;
-	uint32_t exceptions = 0;
-	uint32_t status = 0;
+	uint32_t status;
```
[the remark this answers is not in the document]
vesajaaskelainen: OK

vesajaaskelainen:
Sorry for a bit of delay on activity.

Rebased, resolved review comments, ran xtest, applied tags.

Thanks for the review!

PR #5472, opened 2022-08-03
core: arm: add MPIDR affinity shift and mask for 32-bit

vishnu-banavath:
This change is to add MPIDR affinity shift and mask for
32-bit

Signed-off-by: Vishnu Banavath <vishnu.banavath@arm.com>


jforissier:
> This change is to add MPIDR affinity shift and mask for 32-bit

This is confusing. I'd rather describe the change as: "Add MPIDR shift and mask for MPIDR_EL1 Aff3", because this applies only to the Aarch64 state and NOT to Aarch32 in which Aff3 doesn't exist. Should the macros be named `MPIDR_EL1_*` rather than `MPIDR_*` to reflect this? Maybe not. Should the definitions be guarded by `#ifdef ARM64 .. #endif`? Probably.

jenswi-linaro:
> Should the definitions be guarded by `#ifdef ARM64 .. #endif`? Probably.

I agree, but that's also enough.

core/arch/arm/plat-n1sdp/n1sdp_core_pos.S
```diff
@@ -0,0 +1,32 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Arm Limited
+ */
+
+#include <asm.S>
+#include <arm.h>
+#include "platform_config.h"
+
+FUNC get_core_pos_mpidr , :
```
jenswi-linaro, line 10: Would it be possible to update the weak `get_core_pos_mpidr()` instead?

jenswi-linaro (commented):
Comments for the commit "plat-n1sdp: add N1SDP platform support"
Please add Neoverse somewhere in the commit message. A link to some description of this platform would be welcome too.

PR #5474, opened 2022-08-05
core: link: add --no-warn-rwx-segments

jforissier:
binutils ld.bfd generates one RWX LOAD segment by merging several sections
with mixed R/W/X attributes (.text, .rodata, .data). After version 2.38 it
also warns by default when that happens [1], which breaks the build due to
--fatal-warnings. The RWX segment is not a problem for the TEE core, since
that information is not used to set memory permissions. Therefore, silence
the warning.

Link: [1] https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=ba951afb99912da01a6e8434126b8fac7aa75107

### DIFF
--- a/core/arch/arm/kernel/link.mk
+++ b/core/arch/arm/kernel/link.mk
@@ -31,6 +31,7 @@ link-ldflags += -T $(link-script-pp) -Map=$(link-out-dir)/tee.map
 link-ldflags += --sort-section=alignment
 link-ldflags += --fatal-warnings
 link-ldflags += --gc-sections
+link-ldflags += $(call ld-option,--no-warn-rwx-segments)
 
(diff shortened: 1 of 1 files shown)


jforissier:
@martinetd feel free to give your `Reported-by:` or any other tag/comment.

martinetd:
Thanks! I agree this makes sense given the behaviour isn't new, the version of optee-os I was looking at didn't have ld-options but with that I have no remorse to use this.
Tested-by or other tags don't really make sense given how simple it is, so let's go with reported-by only -- thanks for the suggestion :)

`Reported-by: Dominique Martinet <dominique.martinet@atmark-techno.com>`

jenswi-linaro:

PR #5478, opened 2022-08-08
core: spmc: add SPMC manifest handling

balint-dobszay-arm:
If the SPMD is enabled in TF-A, the register usage at BL31 -> BL32
handover is the following [1]:
  - x0: TOS_FW_CONFIG DT (SPMC manifest)
  - x1: HW_CONFIG DT

This commit modifies the entry function to get aligned with this
convention if OP-TEE is the S-EL1 SPMC. It also adds functionality to
map the SPMC manifest and retrieve it later.

Link: [1] https://trustedfirmware-a.readthedocs.io/en/latest/components/secure-partition-manager.html#spmc-boot
Signed-off-by: Balint Dobszay <balint.dobszay@arm.com>

### DIFF
--- a/core/arch/arm/arm.mk
+++ b/core/arch/arm/arm.mk
@@ -98,6 +98,7 @@ ifeq ($(CFG_CORE_SEL1_SPMC),y)
 $(call force,CFG_CORE_FFA,y)
 $(call force,CFG_CORE_SEL2_SPMC,n)
 $(call force,CFG_CORE_EL3_SPMC,n)
+$(call force,CFG_DT,y)
 endif
 # SPMC configuration "S-EL2 SPMC" where SPM Core is implemented at S-EL2,
 # that is, the hypervisor sandboxing OP-TEE
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
  * Copyright (c) 2015-2022, Linaro Limited
+ * Copyright (c) 2022, Arm Limited
  */
 
 #include <arm.h>
@@ -83,6 +84,9 @@ struct dt_descriptor {
 };
 
 static struct dt_descriptor external_dt __nex_bss;
+#ifdef CFG_CORE_SEL1_SPMC
+static struct dt_descriptor spmc_manifest_dt __nex_bss;
+#endif
 #endif
 
 #ifdef CFG_SECONDARY_INIT_CNTFRQ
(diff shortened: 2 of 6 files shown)


balint-dobszay-arm:
@jenswi-linaro @jforissier, related to https://github.com/OP-TEE/optee_os/pull/5359.

This PR currently marked as draft because there is an issue: TF-A v2.6 passes the `HW_CONFIG` in NS memory (at least on FVP, I'm not sure on qemu?). The `get_aslr_seed()` function will try to access this before the MMU is enabled, which won't work of course.

My temporary workaround is `CFG_CORE_ASLR=n`. However, in TF-A v2.7 this is changed, the `HW_CONFIG` is passed in S memory so there is no issue anymore. Should I just update the FVP platform to TF-A v2.7? Or is it preferred to wait and update all platforms at once?

jforissier:
> This PR currently marked as draft because there is an issue: TF-A v2.6 passes the `HW_CONFIG` in NS memory (at least on FVP, I'm not sure on qemu?). The `get_aslr_seed()` function will try to access this before the MMU is enabled, which won't work of course.

Please make it clear in the subject that this introduces DT support too: "core: spmc: add SPMC manifest **and DT** handling". Please also add your comment about the TF-A v2.6 vs. v2.7 situation to the commit description. It can be quite helpful for troubleshooting I think ;)

> My temporary workaround is `CFG_CORE_ASLR=n`. However, in TF-A v2.7 this is changed, the `HW_CONFIG` is passed in S memory so there is no issue anymore. Should I just update the FVP platform to TF-A v2.7? Or is it preferred to wait and update all platforms at once?

It is acceptable to update platforms at different times.

balint-dobszay-arm:
> > This PR currently marked as draft because there is an issue: TF-A v2.6 passes the `HW_CONFIG` in NS memory (at least on FVP, I'm not sure on qemu?). The `get_aslr_seed()` function will try to access this before the MMU is enabled, which won't work of course.
> 
> Please make it clear in the subject that this introduces DT support too: "core: spmc: add SPMC manifest **and DT** handling". Please also add your comment about the TF-A v2.6 vs. v2.7 situation to the commit description. It can be quite helpful for troubleshooting I think ;)
> 

On second thought, if OP-TEE is S-EL1 SPMC AFAICS we're not relying on the `HW_CONFIG` DT at all. For sure it has been working so far, because currently `x1` is misinterpreted as the pageable part's address, so nothing from that DT could be used. I think I'll just do a modification to drop the address passed in `x1`.

> > My temporary workaround is `CFG_CORE_ASLR=n`. However, in TF-A v2.7 this is changed, the `HW_CONFIG` is passed in S memory so there is no issue anymore. Should I just update the FVP platform to TF-A v2.7? Or is it preferred to wait and update all platforms at once?
> 
> It is acceptable to update platforms at different times.

Okay, thanks, I'll sync with my team if we want to move FVP to v2.7 now or later.

jenswi-linaro:
How boot information is passed to OP-TEE can be put into two categories:
1. Small backward compatible extensions of the upstream scheme
2. A new take on the problem, but tries to be backward compatible where it matters

1 is what we've done so far and it has so far worked rather well, but it's also getting quite complicated if we consider all the different DT options we have.

2 is a new standardized method based on for instance Transfer Lists from the Handoff specification or Boot information protocol from the FF-A spec. This should make it easier to add custom or platform-specific boot parameters without breaking compatibility. It is important that we chose **one** method and then stick to that or we'll end up with ever-increasing baggage. I've made one prototype based on the Handoff specification described in https://linaro.atlassian.net/browse/TS-243, the intention is not to merge this just to show what it takes with a unified approach.

What you're proposing here falls into category 1, but is not backward compatible. I can't see any reason why we shouldn't maintain backward compatibility at this stage. You're also making it impossible to use paging with SPMC at S-EL1. I'd rather just pass the SPMC manifest in some unused register. That way we wouldn't even need any ifdefs in the assembly code, just preserve and pass on yet another register.

balint-dobszay-arm:
It makes sense, thanks for the detailed reply. I think this should be discussed further including the TF-A (SPMD) team, I'll organise a meeting when I'm back from holiday.

balint-dobszay-arm:
> It makes sense, thanks for the detailed reply. I think this should be discussed further including the TF-A (SPMD) team, I'll organise a meeting when I'm back from holiday.

As discussed in the meeting, the FF-A v1.1 boot protocol should be used to have a generic solution for this topic. Details and scheduling of the implementation itself is TBD.

PR #5484, opened 2022-08-12
core: mbedtls: Use a DRBG to speed up RSA key generation

larperaxis:
We are doing performance tuning on the PKSC11 TA and the RSA key generation speed was one area that needed improvements. The main bottleneck is the source of random bits that in any real system will be a slow hardware TRNG.
Let's do like most crypto libraries do and put a DRBG between the key generator and the TRNG. Our work focuses only on the performance with mbedtls as the core crypto library since it had the overall best performance for asymmetric crypto.

For mbedtls there was a choice between CTR and HMAC DRBG. I have no preference on one or the other, the choice of CTR was simply because its header file already was included in the rsa.c file. Perhaps a hint that someone had intended to do this work earlier.

Performance diff on RSA 2048 key generation for our system with Cortex A53 @ 1.2 GHz and 1e5 bits/s TRNG:
Before patch (14 samples)
mean time 14.02 sec stddev. 7.91

With patch (14 samples):
mean time 1.67 sec stddev. 1.24

### DIFF
--- a/lib/libmbedtls/core/rsa.c
+++ b/lib/libmbedtls/core/rsa.c
@@ -203,8 +203,13 @@ TEE_Result crypto_acipher_gen_rsa_key(struct rsa_keypair *key, size_t key_size)
 {
 	TEE_Result res = TEE_SUCCESS;
 	mbedtls_rsa_context rsa;
+	mbedtls_ctr_drbg_context rngctx;
 	int lmd_res = 0;
 	uint32_t e = 0;
+
+	mbedtls_ctr_drbg_init(&rngctx);
+	if (mbedtls_ctr_drbg_seed(&rngctx, mbd_rand, NULL, NULL, 0))
+		return TEE_ERROR_BAD_STATE;
 
 	memset(&rsa, 0, sizeof(rsa));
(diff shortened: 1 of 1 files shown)


larperaxis:
The IBART job failed on a missing u-boot git tag. Can the maintainers check that and retrigger the workflow ?

jbech-linaro:
> The IBART job failed on a missing u-boot git tag. Can the maintainers check that

I will have a look, IBART leverage the manifests as any other builds and U-Boot is often rather static. But I will have a check.


> and retrigger the workflow ?

Yes I can do that, as an FYI if you have a Google account, even you yourself can re-trigger the job (just press login [here](https://optee.mooo.com:5000) and then restart your build).

jbech-linaro:
Didn't find anything obvious, it worked fine on my own end, but I had an issue with U-boot as well when updating the reference tree that we store locally. So I've wiped U-boot from there, re-synced it (was OK) and now I've restarted your build, let's see if that fixes the issue. If not, I'll have another go.

jbech-linaro:

jenswi-linaro:

PR #5485, opened 2022-08-13
drivers: se050: optional I2C access via trampoline

ldts:
Platforms with secure I2C buses (i.e: STM32MP1) or those with only
a secure element on the bus might prefer not to delegate the I2C
traffic to the REE.


jforissier (approved):

PR #5489, opened 2022-08-18
Enable CAAM driver on mx7ulp

clementfaure:


jforissier (commented):

But I suppose "core: imx: enable the CAAM driver on mx7ulpevk" should come after the other commit, no?

clementfaure:
@jforissier You're right! I reordered the commits and applied your tag. Thanks!

PR #5490, opened 2022-08-18
clang.mk: fix build on AArch64 host

jforissier:
When building on an AArch64 host and using OP-TEE's build.git [1], the
AArch64 cross compiler prefix is "aarch64-linux-" instead of the usual
"aarch64-linux-gnu-". This happens due to [2]. Clang still expects
--target=aarch64-linux-gnu so for convenience map the prefix accordingly
in mk/clang.mk.

Link: [1] https://github.com/OP-TEE/build.git
Link: [2] https://github.com/OP-TEE/build/blob/3.18.0/toolchain.mk#L97


jenswi-linaro:

PR #5495, opened 2022-08-19
Miscellaneous K3 improvements and fixups

glneo:
Some small cleanups that didn't each warrant their own PRs, should be all relatively straightforward.


core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -115,6 +115,8 @@ static TEE_Result sa2ul_init(void)
 	/* Modify TRNG firewall to block all others access */
 	control = FW_ENABLE_REGION;
 	permissions[0] = (FW_BIG_ARM_PRIVID << 16) | FW_SECURE_ONLY;
+	start_address = RNG_BASE;
+	end_address = RNG_BASE + RNG_REG_SIZE - 1;
```
jforissier, line 119: TBH I don't like this kind of defensive programming too much. It makes the code more difficult to understand. Since `ti_sci_get_fwl_region()` is supposed to always provide the requested address unless it returns an error, I suggest either doing nothing here or adding an `assert()`.

jforissier (commented):
Hi @glneo,

One comment on "plat-k3: drivers: Set SA2UL firewall region addresses", see below. The other commits are fine and are:

Thanks!

core/arch/arm/plat-k3/drivers/sa2ul.c
glneo: Yes `ti_sci_get_fwl_region()` is going to return an address or an error, but the address might not cover the peripheral we want. In AM64x/AM62x the firewall regions are much more flexible and a given firewall region can be used to cover multiple different IPs, the addresses are not fixed in place.

In AM64x for example, Firewall 35 region 3 is expected to be already setup to cover the RNG, as that is how ROM/SYSFW sets it up, but nothing prevents it from being changed before OP-TEE starts (for instance the R5 core in AM64x starts up and performs tasks before the A53 core that runs OP-TEE is brought up). In this case being defensive and forcing the region address back to cover the RNG is a core part of securing our RNG for OP-TEE's exclusive use.

jforissier: Ah, that's interesting. I wouldn't mind having at least a part of this explanation in the commit description... In any case:

glneo: Okay, updated the commit description to be more clear.

PR #5496, opened 2022-08-23
core: arm: use CFG_STACK_THREAD_EXTRA also with CFG_WITH_PAGER

ldts:
The Plug-and-Trust[1] companion project to the se050 crypto driver
has high stack requirements.
Due to this the driver is configured with CFG_STACK_THREAD_EXTRA.

This configuration however is not enforced by OP-TEE when the pager
is running - as it happens on the STM32MP1 platform; this causes the
driver to hit core data-aborts during some cryptographic operations
due to the Plug-and-Trust overflowing the stack.

Fix this by applying CFG_STACK_THREAD_EXTRA to the STACK_THREAD_SIZE
also when the pager is enabled.

[1] https://github.com/NXP/plug-and-trust

### DIFF
--- a/core/arch/arm/include/kernel/thread_private_arch.h
+++ b/core/arch/arm/include/kernel/thread_private_arch.h
@@ -25,7 +25,12 @@
 #else
 #define STACK_TMP_SIZE		(2048 + STACK_TMP_OFFS)
 #endif
+
+#ifdef CFG_WITH_PAGER
+#define STACK_THREAD_SIZE	(8192 + CFG_STACK_THREAD_EXTRA)
+#else
 #define STACK_THREAD_SIZE	8192
+#endif
 
 #if defined(CFG_CORE_SANITIZE_KADDRESS) || defined(__clang__) || \
 	!defined(CFG_CRYPTO_WITH_CE)
@@ -42,7 +47,12 @@
 #else
 #define STACK_TMP_SIZE		(2048 + STACK_TMP_OFFS)
 #endif
+
+#ifdef CFG_WITH_PAGER
+#define STACK_THREAD_SIZE	(8192 + CFG_STACK_THREAD_EXTRA)
+#else
 #define STACK_THREAD_SIZE	8192
+#endif
 
 #if TRACE_LEVEL > 0
 #define STACK_ABT_SIZE		3072


ldts:
Tested on STM32MP157 using OP-TEE 3.18.

Before the change, the RSA path generates core data-aborts due to stack overflows.

core/arch/arm/include/kernel/thread_private_arch.h
```diff
@@ -25,7 +25,12 @@
 #else
 #define STACK_TMP_SIZE		(2048 + STACK_TMP_OFFS)
 #endif
+
+#ifdef CFG_WITH_PAGER
+#define STACK_THREAD_SIZE	(8192 + CFG_STACK_THREAD_EXTRA)
```
jforissier, line 30: I think `CFG_STACK_THREAD_EXTRA` should be included in `STACK_THREAD_SIZE` regardless of pager configuration. In fact the current code looks buggy, see `print_stack_limits()` / `thread_stack_start()` / `thread_stack_size()` / `get_stack_limits()`. They all ignore `CFG_STACK_THREAD_EXTRA`.

ldts: I agree (hence why I didnt just add it to init_thread_stacks as it is done on the non pager case)

So based on this, 

```
#ifndef CFG_WITH_PAGER
DECLARE_STACK(stack_thread, CFG_NUM_THREADS,
	      STACK_THREAD_SIZE + CFG_STACK_THREAD_EXTRA, static);
#endif

```

I could have done

```
#ifdef CFG_WITH_PAGER
static void init_thread_stacks(void)
{
	size_t n = 0;

	/*
	 * Allocate virtual memory for thread stacks.
	 */
	for (n = 0; n < CFG_NUM_THREADS; n++) {
		tee_mm_entry_t *mm = NULL;
		vaddr_t sp = 0;
		size_t num_pages = 0;
		struct fobj *fobj = NULL;

		/* Find vmem for thread stack and its protection gap */
		mm = tee_mm_alloc(&tee_mm_vcore,
				  SMALL_PAGE_SIZE + STACK_THREAD_SIZE + CFG_STACK_THREAD_EXTRA );
		assert(mm);
```

but It seemed wrong.

But this code area being so sensitive I was hoping someone with more experience to comment ; to me adding it to STACK_THREAD_SIZE is bullet-proof and easier to test and manage.

should we do that then?

jforissier: Let me try something, I'll share with you so that you can test on STM32MP1 with pager.

jforissier: https://github.com/OP-TEE/optee_os/pull/5500

ldts: will test this in an hour or two but the patch is pretty much what I had in  mind. nice!

ldts: @jforissier works for me with the pager on STM32MP1. 
I'll close this PR then

PR #5505, opened 2022-08-29
gcc12 fixes

quaresmajose:

### DIFF
--- a/core/arch/arm/plat-stm32mp1/main.c
+++ b/core/arch/arm/plat-stm32mp1/main.c
@@ -139,9 +139,9 @@ static TEE_Result init_console_from_dt(void)
 	/* Replace early console with the new one */
 	console_flush();
 	console_data = *pd;
-	free(pd);
 	register_serial_console(&console_data.chip);
 	IMSG("DTB enables console (%ssecure)", pd->secure ? "" : "non-");
+	free(pd);
 
 	return TEE_SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


jforissier (approved):
Hi @quaresmajose, thanks for the patches.

quaresmajose:
> Hi @quaresmajose, thanks for the patches.
> 

Thanks for the review, added the `Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`

clementfaure:
For `eb12a9916c1812fb79a2bbb4d6a6fe460ee4580b`

quaresmajose:
The `CI / Code style (pull_request) ` job is falling because the line is bigger than 75 chars but this line is the compiler stdout and IMO is better to have it as is.
```
dd2c90e6 drivers: imx: dcp: fix compilation address error
WARNING: Possible unwrapped commit description (prefer a maximum 75 chars per line)
#15: 
| core/drivers/imx/dcp/dcp_huk.c:71:23: error: the comparison will always evaluate as 'true' for the address of 'data' will never be NULL [-Werror=address]
```

jforissier:
> The `CI / Code style (pull_request) ` job is falling because the line is bigger than 75 chars but this line is the compiler stdout and IMO is better to have it as is.

I agree. It seems checkpatch is confused by the new error reporting style of GCC (with a `|` character at the start of the line). Let's just ignore that for the moment. Longer term it would be good to update the regexp in checkpatch.

PR #5426, opened 2022-07-07
Versal ACAP

ldts:
This patch-set extends the current support for the [Versal ACAP platform.](https://www.xilinx.com/products/silicon-devices/acap/versal.html) with a TRNG, access to eFuses and PUF and certain hardware supported cryptographic operations. It also provides a configurable HUK to help generating a secure key for RPMB accesses.

Crypto-operations not supported by the hardware are routed to their software implementations in libtomcrypt.

To understand how this port works, the PLM firmware services used/mentioned by these OP-TEE drivers can be obtained in this repository:
https://github.com/Xilinx/embeddedsw/tree/master/lib/sw_services 

The versal_mbox.c driver requests are handled by the server side on those services:
i.e. in the case of eFuse (NVM) =>  https://github.com/Xilinx/embeddedsw/tree/master/lib/sw_services/xilnvm/src/server

Some of the functionality -i.e. PUF access or NVM access - is restricted by the PLM itself at build time; even inside those services the available calls might be controlled by another level of PLM build time configs ( i.e. see comments on versal_nvm.c regarding  `efuse_write_puf_as_user_fuse`). 

Introduction to the OP-TEE port:
https://github.com/OP-TEE/optee_docs/pull/163

Tested on VCK-190 https://www.xilinx.com/products/boards-and-kits/vck190.html with BSP xilinx-vck190-2022.1 
xtests: https://github.com/OP-TEE/optee_test/pull/606 
with the above change:
```
29219 subtests of which 0 failed
120 test cases of which 0 failed
0 test cases were skipped
TEE test application done!

```

Some unit tests used during development can be found in this commit:
https://github.com/ldts/optee_os/commit/15d27821d6e561697cbc0603c90d93a2f7a59e11
This was worth mentioning since xtests will not validate the SHA3-384 driver.

```
I/TC: Versal: Test NVM                                                                                                                                                                                                                                                                                                                                                                                                                                                                
I/TC: ---- EFUSE wr puf userf:                           [DISABLED]                                                                                                                                                                                                                                                                                                                                                                                                                   
I/TC: ---- EFUSE wr aes bbram:                           [OK]                                                                                                                                                                                                                                                                                                                                                                                                                         
I/TC: ---- EFUSE wr aes usr1:                            [DISABLED]                                                                                                                                                                                                                                                                                                                                                                                                                   
I/TC: ---- EFUSE wr usr:                                 [DISABLED]                                                                                                                                                                                                                                                                                                                                                                                                                   
[...]

PR #5507, opened 2022-08-29
Versal - platform updates

ldts:

### DIFF
--- a/core/arch/arm/plat-versal/link.mk
+++ b/core/arch/arm/plat-versal/link.mk
@@ -1,0 +1,3 @@
+include core/arch/arm/kernel/link.mk
+
+all: $(link-out-dir)/tee-raw.bin
--- a/core/arch/arm/plat-versal/conf.mk
+++ b/core/arch/arm/plat-versal/conf.mk
@@ -2,6 +2,8 @@ PLATFORM_FLAVOR ?= generic
 
 include core/arch/arm/cpu/cortex-armv8-0.mk
 
+CFG_MMAP_REGIONS ?= 24
+
 $(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
 $(call force,CFG_WITH_ARM_TRUSTED_FW,y)
 $(call force,CFG_TEE_CORE_NB_CORE,2)


core/arch/arm/plat-versal/link.mk
```diff
@@ -0,0 +1,11 @@
+include core/arch/arm/kernel/link.mk
+
+all: $(link-out-dir)/tee-raw.bin
```
jforissier, line 3: Shouldn't that be `all: $(link-out-dir)/uTee`? In other words, why should `tee-raw.bin` be generated by default but not `uTee`?

jforissier (commented):
> `Increase the number of regions`

Why, or perhaps more precisely, why by default?

core/arch/arm/plat-versal/link.mk
ldts: sorry yes. we dont really need uTee

I think I can just 

`all: $(link-out-dir)/tee-raw.bin  `

jforissier: OK, so with just the above two lines (include + all:) this commit is:

ldts:
> > `Increase the number of regions`
> 
> Why, or perhaps more precisely, why by default?

sorry didnt notice your question: yes so just as done on other platforms - probably more than what is actually needed but I had to increase it while developing some of the drivers (IIRC when I added support for the GPIO driver)...so picked 24 as this value seemed the norm and allowed my drivers to be mapped.

jforissier:
> > > `Increase the number of regions`
> > 
> > 
> > Why, or perhaps more precisely, why by default?
> 
> sorry didnt notice your question: yes so just as done on other platforms - probably more than what is actually needed but I had to increase it while developing some of the drivers (IIRC when I added support for the GPIO driver)...so picked 24 as this value seemed the norm and allowed my drivers to be mapped.

Alright, makes sense of course. Perhaps you could add a few words in the commit description, for example mentioning that the new value is a reasonable number to use when enabling a few more drivers than are currently available upstream?

For that commit:

ldts:
For reference, there is now an easy way to build a release besides Foundries.io v88 https://foundries.io/products/releases/88/
https://github.com/OP-TEE/build/pull/590

ldts:
@jforissier updated with tags. Added additional information on both commits.

jforissier:
> @jforissier updated with tags. Added additional information on both commits.

You forgot the Ack on the last one :wink:

ldts:
argh sorry. doing it now

PR #5508, opened 2022-08-29
drivers: versal: true random number generator

ldts:
Configure the TRNG driver to operate in Hybrid mode with derivative
function.

This driver was ported from its original FSBL implementation (1)

(1) https://github.com/Xilinx/embeddedsw

### DIFF
--- a/core/drivers/versal_trng.c
+++ b/core/drivers/versal_trng.c
@@ -1,0 +1,1149 @@
+// SPDX-License-Identifier: MIT
+/*
+ * Copyright (C) 2022 Xilinx, Inc.  All rights reserved.
+ * Copyright (C) 2022 Foundries Ltd.
+ *
+ * Driver port from Xilinx's FSBL implementation, jorge@foundries.io
+ *
+ * The Xilinx True Random Number Generator(TRNG) module in Versal - PMC TRNG
+ * consists of an entropy source, a deterministic random bit generator (DRBG)
+ * and health test logic, which tests the randomness of the generated data.
+ * The entropy source for the unit is an array of Ring Oscillators.
+ *
+ * The Versal PMC TRNG is envisaged to operate in three basic modes: DRNG, PTRNG
+ * and HRNG . Each of these can be operated with or without Derivative Function
+ * (DF), resulting in a total of 6 different modes of operation.
+ *
+ * NIST SP-800-90A practically requires the true random generators based on
+ * CTR_DRBG to include a derivation function (DF). This is expected to be
+ * implemented inside the Silicon (TRNG IP). However, the version of the IP used
+ * in Versal PMC does not have this capability. Hence, a software
+ * implementation of the DF is done in this driver.
+ *
+ * DRNG mode: Deterministic Random Number Generator mode.
+ *            In this mode, the DRBG portion of the TRNG is used. User provides
+ *            the (external) seed.
+ * PTRNG mode: Physical True Random Number Generator mode (aka Entropy mode).
+ *            In this mode digitized Entropy source is output as random number.
+ * HRNG mode: Hybrid Random Number Generator mode.
+ *            This is combination of above two modes in which the Entropy source
+ *            is used to provide the seed, which is fed to the DRBG, which in
+ *            turn generates the random number.
+ *
+ * DRNG mode with DF: It may not be common usecase to use the DF with DRNG as
+ * the general expectation would be that the seed would have sufficient entropy.
+ * However, the below guideline from section 10.2.1 of NIST SP-800-90A implies
+ * that need for DF for DRNG mode too: "..the DRBG mechanism is specified to
+ * allow an implementation tradeoff with respect to the use of this derivation
+ * function. The use of the derivation function is optional if either an
+ * approved RBG or an entropy source provides full entropy output when entropy
+ * input is requested by the DRBG mechanism. Otherwise, the derivation function
+ * shall be used". Sufficient large entropy data from user is fed to DF to
+ * generate the seed which will be loaded into the external seed registers.
+ * From here, it is similar to regular DRNG mode.
+ *
+ * PTRNG mode with DF: This mode is similar to PTRNG mode, however, the entropy
+ * data from the core output registers are accumulated and fed to the DF
+ * (instead of directly consuming it). The output of the DF would be final
+ * random data. In this mode, the output of DF is not seed but the random data.
+ *
+ * HRNG mode with DF: This mode is the combination of the above two modes.
+ * The entropy data is fed to the DF to produce seed. This seed is loaded to the
+ * external seed registers which provide seed to the DRBG.
+ */
+#include <arm.h>
+#include <crypto/crypto.h>
+#include <initcall.h>
+#include <io.h>
+#include <kernel/delay.h>
+#include <kernel/panic.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <rng_support.h>
+#include <stdlib.h>
+#include <string.h>
+#include <tee/tee_cryp_utl.h>
+#include <trace.h>
+
+#define TRNG_BASE            0xF1230000
+#define TRNG_SIZE            0x10000
+
+#define TRNG_STATUS			0x04
+#define TRNG_STATUS_QCNT_SHIFT		9
+#define TRNG_STATUS_QCNT_MASK		(BIT(9) | BIT(10) | BIT(11))
+#define TRNG_STATUS_CERTF_MASK		BIT(3)
+#define TRNG_STATUS_DTF_MASK		BIT(1)
+#define TRNG_STATUS_DONE_MASK		BIT(0)
+#define TRNG_CTRL			0x08
+#define TRNG_CTRL_EUMODE_MASK		BIT(8)
+#define TRNG_CTRL_PRNGMODE_MASK		BIT(7)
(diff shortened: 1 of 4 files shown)


etienne-lms (commented):
some comments. LGTM otherwise.

ldts:
@etienne-lms thanks a lot for your comments!. will update the PR in the morning (sorry for the late responses but you saw all the other stuff pending at the same time)

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1151 @@
...
+ * approved RBG or an entropy source provides full entropy output when entropy
+ * input is requested by the DRBG mechanism. Otherwise, the derivation function
+ * shall be used". Sufficient large entropy data from user is fed to DF to
+ * generate the seed which will be loaded into the external seed registers.
+ * From here, it is similar to regular DRNG mode.
+ *
+ * PTRNG mode with DF: This mode is similar to PTRNG mode, however, the entropy
+ * data from the core output registers are accumulated and fed to the DF
+ * (instead of directly consuming it). The output of the DF would be final
+ * random data. In this mode, the output of DF is not seed but the random data.
```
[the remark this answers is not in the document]
ldts: I think it is right: usually the DF is used to just generate a seed. The comment is saying that when used in this mode, the DF output is not a seed but the actual random data.

ldts:
@etienne-lms updated and rebased

ldts:
@etienne-lms maybe it can be merged?

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1151 @@
...
+	return TEE_SUCCESS;
+}
+
+void plat_rng_init(void)
+{
+}
+
+static TEE_Result trng_hrng_mode_init(void)
+{
+	uint8_t pers_str[TRNG_PERS_STR_LEN] = {
```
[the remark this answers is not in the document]
jforissier: @ldts this comment wasn't addressed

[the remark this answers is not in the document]
ldts: @jforissier thanks, didnt notice your comment earlier.... just pushed it.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1147 @@
...
+#define TRNG_STATUS_DONE_MASK		BIT(0)
+#define TRNG_CTRL			0x08
+#define TRNG_CTRL_EUMODE_MASK		BIT(8)
+#define TRNG_CTRL_PRNGMODE_MASK		BIT(7)
+#define TRNG_CTRL_PRNGSTART_MASK	BIT(5)
+#define TRNG_CTRL_PRNGXS_MASK		BIT(3)
+#define TRNG_CTRL_TRSSEN_MASK		BIT(2)
+#define TRNG_CTRL_PRNGSRST_MASK		BIT(0)
+#define TRNG_EXT_SEED_0			0x40
+/* Below registers are not directly referenced in driver but are accessed
```
etienne-lms, line 85:
```
/*
 * ...
 *
```
Ditto at line 100.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1147 @@
...
+		xfer_len = DF_PAD_DATA_LEN + TRNG_PERS_STR_LEN;
+		src = (uintptr_t)trng->dfin.pstr;
+		offset = MAX_PRE_DF_LEN - trng->len;
+	}
+
+	/* Move back into the dfin structure */
+	dst = src - offset;
+	reminder = (uintptr_t)&trng->dfin + sizeof(trng->dfin) - offset;
+	if (offset) {
+		memcpy((void *)dst, (void *)src, xfer_len);
```
etienne-lms, line 456: `memove()` seems a better option.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1147 @@
...
+
+	if (trng_reseed(&versal_trng, NULL, usr_cfg.dfmul)) {
+		EMSG("Reseed Failed");
+		goto error;
+	}
+
+	return TEE_SUCCESS;
+error:
+	panic();
+	return TEE_ERROR_GENERIC;
```
etienne-lms, line 1144: Remove the return instruction.

Maybe replace `goto error` with `panic` straight? Whatever, do as you prefer.

etienne-lms (commented):
Comment on commit message: fix link using a `Link:` tag.
```
This driver was ported from its original FSBL implementation [1].

[1] https://github.com/Xilinx/embeddedsw
```

etienne-lms:
Hi @ldts, sorry i was a bit busy and left P-R reviews for a while and only report now these 2 coding style issues (inline comments) and must importantly a `memcpy()` that should be replaced with a `memmove()`.

core/drivers/versal_trng.c
ldts: um, why, how do you know src is not being used somewhere else. I dont feel comfortable with this change. why do you feel it is required?

etienne-lms: src and dst are in the same address range. As stated in the inline comment, this move data inside a single buffer, hence this should use `memmove()` IIUC.

etienne-lms: yes, i was thinking of overlaps. From memcpy() man page:
> DESCRIPTION
>      The  memcpy()  function  copies n bytes from memory area src to memory area dest.
>      **The memory areas must not overlap.  Use memmove(3) if the memory areas do overlap.**

Maybe `Xil_SMemCpy()` in the referred code ensures destination address is lower than source when buffer do overlap. Using `memcpy()` you do not have this ensurance.

ldts: um, I think you are right.
maybe we should just do likewise?
```
int Xil_SMemCpy(void *Dest, const u32 DestSize,
	const void *Src, const u32 SrcSize, const u32 CopyLen)
{
	int Status = XST_FAILURE;
	const u8 *Src8 = (const u8 *) Src;
	const u8 *Dst8 = (u8 *) Dest;
	void * volatile DestTemp = Dest;
	const void * volatile SrcTemp = Src;

	if ((Dest == NULL) || (Src == NULL)) {
		Status =  XST_INVALID_PARAM;
	}
	else if ((CopyLen == 0U) || (DestSize < CopyLen) || (SrcSize < CopyLen)) {
		Status =  XST_INVALID_PARAM;
	}
	/* Return error for overlap string */
	else if ((Src8 < Dst8) && (&Src8[CopyLen - 1U] >= Dst8)) {
		Status =  XST_INVALID_PARAM;
	}
	else if ((Dst8 < Src8) && (&Dst8[CopyLen - 1U] >= Src8)) {
		Status =  XST_INVALID_PARAM;
	}
	else {
		(void)memcpy(DestTemp, SrcTemp, CopyLen);
		Status = XST_SUCCESS;
	}

	return Status;
}
```

ldts: I think when I rewrote this code early this year I already made all the proper corrections thought so we are probably are going in circles...but it is a good thing to review.

ldts: just pushed a safeguard (in any case the algorithm expected the remainder to be set to 0 so truncating seems to be fine.

jforissier:
For "MAINTAINERS: add Versal ACAP TRNG maintainer":

ldts:
> Hi @ldts, sorry i was a bit busy and left P-R reviews for a while and only report now these 2 coding style issues (inline comments) and must importantly a `memcpy()` that should be replaced with a `memmove()`.

hi thanks! @etienne-lms I followed up where you made the original comments. safeguarding against a potential overlap seems a better choice as we need to remain in the structure.

ldts:
done then?

ldts:
and rebased

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1150 @@
...
+	trng->status = TRNG_ERROR;
+	return TEE_ERROR_GENERIC;
+}
+
+static struct versal_trng versal_trng = {
+	.cfg.base = TRNG_BASE,
+	.cfg.len = TRNG_SIZE,
+};
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
```
zjuzzy, line 1070: Does it need a header file to declare this interface? so that others can use this interface.

ldts: no it is not required. the interface is already in core/include/rng_support.h (it is not driver specific)

ldts:
so all good? @jforissier ?

jforissier:
@ldts LGTM you may add:

But I'd like @etienne-lms's ack or review tag since he did most of the review.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with proposal on remaining comment applied.

core/drivers/versal_trng.c
```diff
@@ -0,0 +1,1150 @@
...
+		offset = MAX_PRE_DF_LEN - trng->len;
+	}
+
+	/* Move back into the dfin structure */
+	dst = src - offset;
+	reminder = (uintptr_t)&trng->dfin + sizeof(trng->dfin) - offset;
+	if (offset) {
+		/* safeguard memcpy operation */
+		if (xfer_len > offset)
+			xfer_len = offset;
```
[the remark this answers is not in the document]
ldts: No that is ok, I'll panic and configure the default multiplier as per the reference code. thanks!

ldts:
@jforissier @etienne-lms thanks for the review. updated!

PR #5513, opened 2022-08-30
Remove support for LS1021A-QDS and LS1021A-TWR platform

sahilnxp:
LS1021A-TWR and LS1021A-QDS platforms doesn't support OP-TEE anymore, so removing their support.


jforissier:
> LS1021A-TWR and LS1021A-QDS platforms doesn't support OP-TEE anymore

Do you mean the current OP-TEE code doesn't work on these boards anymore? Or NXP doesn't want to maintain OP-TEE on them? I'd like to better understand the reason for removing the  code and also the consequences on current (downstream) users.

Apart from that, technically this looks fine so:

sahilnxp:
Yes, you are right, current OP-TEE code doesn't work on these boards anymore that's why removing their support.

Tags applied.

PR #5514, opened 2022-08-30
Imx8mn i2c driver

RobertKrikke:


ldts:
@RobertKrikke looks good but could you repost using a single commit please? with that you can add my 
`Acked-by: Jorge Ramirez-Ortiz <jorge@foundries.io>`

ldts:
@jforissier could we pick this up please?

PR #5516, opened 2022-09-01
RSA route unsupported crypto API operations to their software implementations if needed

ldts:


ldts:
@cneveux @clementfaure any comments?

nathan-menhorn:
@cneveux @clementfaure any updates on this pull request?

core/lib/libtomcrypt/rsa.c
```diff
@@ -84,6 +85,10 @@ static TEE_Result tee_algo_to_ltc_hashindex(uint32_t algo, int *ltc_hashindex)
 
 TEE_Result crypto_acipher_alloc_rsa_keypair(struct rsa_keypair *s,
 					    size_t key_size_bits __unused)
+__weak_alias("sw_crypto_acipher_alloc_rsa_keypair");
```
jforissier, line 88: Indent with 1 tab?

lib/libutils/ext/include/compiler.h
```diff
@@ -20,6 +20,7 @@
 #define __packed	__attribute__((packed))
 #endif
 #define __weak		__attribute__((weak))
+#define __weak_alias(x) __attribute__((weak, alias(x)))
```
jforissier, line 23: Thinking about it, I'd prefer `#define __alias(x) __attribute__((alias(x)))`. Reason being, I find the resulting macro ambiguous. Consider `void f(void) __weak_alias("g");`. What is weak? `f()` or `g()`? With `void f(void) __weak __alias("g");` it is clear that we are adding an attribute (the name of the alias) to a weak function.

jforissier (commented):
I'm not sure I understand the purpose of "crypto: se05x: provide RSA software crypto operations":

> The SE05X driver however explicitly disables [the fallback software implementations].

Where does it disable them? What happens without this patch?

core/lib/libtomcrypt/rsa.c
ldts: > I'm not sure I understand the purpose of "crypto: se05x: provide RSA software crypto operations":
> 
> > The SE05X driver however explicitly disables [the fallback software implementations].
> 
> Where does it disable them? What happens without this patch?

without this patch, SE05x non-supported RSA operations (certain key sizes) will fail.

this was one of the reasons for needing this 
https://github.com/OP-TEE/optee_test/commit/b657ea448eb96c936d62f5fc8ea97b747170034d 

once/if this is merged we could revert it

jforissier: I still don't get it. You're providing SW fallbacks that return not implemented and that would make things work? The crypto framework is supposed to detect not implemented from the driver then call the SW fallback in this case, no?

ldts: um no, sorry I misspoke (I went into automatic mode without thinking on this use case which is the opposite of everything else in the series...)

Right .so the SE05x driver explicitly [disables](https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/crypto.mk#L70) RSA software operations in its config. 
So the fallback aliases are not available at build time hence why providing them this way is required.

ldts: Ok but notice that the question you are raising is not about any new code but the _current_ behavior of the driver instead, (which I am trying to no alter in this commit)
You are asking why did we need to disable LTC RSA in order to enable the SE050 RSA driver.

I dont remember the specifics but being the SE05x driver platform independent there was a build conflict with the RSA software implementations (they share the function names). Since fallback didn't exist at the time I opted for just disabling the code that would not be required.

Coming back to this series - lets not forget this is geared to enable Versal like systems, where HW crypto support might just be limited to a few sizes and cypers - I didnt want to add even more functionality to the series: ~7000 LOC to review was already quite enough; as a matter of fact it has been over two months in since I sent the PR and I still haven't been able to merge more than a few of the simpler commits (this is not criticism btw, it is just stating what needs to be ... reviews must be as thorough as possible to maintain the quality of the software, there are no shortcuts to be taken).

So I don't think that right now is the time to change the way the SE050 driver works: so fallback to RSA software is not enabled/supported by this commit on the SE05x crypto driver (_yet_).

if you think it has to be done now, I can spend a few days working on it. But IMO I see no real need having so many other commits pending for review in my work trees (besides the one currently posted)

But that this - SE05x RSA fallback- will need to be supported at some point I am sure of it.

ldts: I understand what you are saying but there is nothing inconsistent on this:
The series enables the _possibility_ of falling back to software operations: the series does **not** enable the actual fallback on any platform, not even on the SE05x (that should be up to each platform maintainer I'd imagine)

And to enable the possibility of software fallbacks on SE05x enabled platforms  we must provide stub calls for the undefined due to `_CFG_CORE_LTC_RSA = n `.

I can revisit and dig a bit more why disabling  _CFG_CORE_LTC_RSA was required (probably related to the way the crypto driver is built leaving no other choice when we wanted to use the CAAM at the same time). It was complex.

I feel that we are going to on a tangent and that this investigation that you are requesting could have been done at a later time since all changes are contained within the crypto driver itself. And as I said, we will have to do it when we integrate SE05x versions with reduced RSA support.

But ok. Will look into this over the coming weeks. I suppose Versal up-streaming can continue to wait a bit more, there doesn't seem to be a sense of urgency anywhere.

ldts:
once I look into the SE050 support for this feature, what about the actual PR?

I'd rather not continue working on it if it is going to be rejected anyway.

Need input..it has been over 2 months since the first post as part of the larger Versal PR, more than 5 weeks since the second one and not much other than @jenswi-linaro rejecting it initially; I understand everyone is busy but if this is a no-go then I'd rather stop at this point.

 Just so we are on the same page, I'll have to also press the brakes on the Versal crypto drivers since they rely on being able to fallback to software operations.

anyway, any input will be appreciated.

ldts:
hey @jenswi-linaro at least that is some common ground. that is good, right?

Also maybe you - or maybe some maintainers- would like to propose something rather than not having an implementation that we all agreed is needed. Often it is a good thing to leave personal aesthetics aside - we are not talking about issues in this PR, at least I havent heard of any coherently articulated - and bring in functionality.

Would be nice to have a good story for ECC/RSA instead of what we have today (different APIs and some glitches WRT of how keys are removed and such).

what do you think? could you propose something? I dont mind doing some validation for you if you need me to.

jenswi-linaro:
@ldts, how about this https://github.com/OP-TEE/optee_os/pull/5578 ?

ldts:
> @ldts, how about this #5578 ?

thanks a lot @jenswi-linaro . Let me have a look and do some validation on Versal

ldts:
since the crypto api wide discussions for global fallback requires more people to participate on the discussion - and test- I am proposing this first step.

once merged - and once Versal AES support is also merged since it depends on the ipi - I can post the RSA driver.

core/lib/libtomcrypt/rsa.c
ldts: to complete this thread: contrary to what I said earlier, the RSA SE05x driver will never be able to fallback to software operations that require private keys since the ARM cores can not access theme in the secure element.

it might do in the future on public key operations (ie to accelerate starting TAs (rsa verify) but not at this time.

ldts:
@jenswi-linaro just to be clear, with this new patch-set, every individual RSA crypto driver can decide whether to delegate the processing of the request to the software versions of the algorithms.

I think it is rather simple and can cause no regressions so I am assuming it should be quick to merge.

core/drivers/crypto/crypto_api/include/drvcrypt_acipher.h
```diff
@@ -61,6 +61,7 @@ struct drvcrypt_rsa_ssa {
  */
 struct drvcrypt_rsa_ed {
 	enum drvcrypt_rsa_id rsa_id; /* RSA Algorithm Id */
+	uint32_t algo;
```
jenswi-linaro, line 64: Please add a comment on which algorithm this is.

jenswi-linaro (commented):
Comments for "crypto-api: rsa: pass algorithm to implementation"
please add a "core:" prefix

With the comments fixed:

jenswi-linaro:
For "core: compiler.h: introduce __alias(x)"
please use a "libutils:" prefix instead of "core:", with that fixed:

For "crypto_impl: rsa: support the crypto driver"
please replace the "crypto_impl" prefix with "core:", with that fixed:

For
- "core: ltc: rsa: support the crypto driver"
- "lib: mbedtls: rsa: support the crypto driver"
Please apply:

ldts:
tags applied. thanks

clementfaure:
For "core: crypto-api: rsa: pass algorithm to implementation"

ldts:
thanks @clementfaure

PR #5518, opened 2022-09-02
plat: rcar: fix core pos calculation for H3 boards

lorc:
Due to mistake, cluster position wasn't shifted left if chip is not M3W. This
led to erroneous core ID calculation on chips that are not M3W. Actually, this
affected only H3, as only this chip has two clusters.

Fix this by always shifting x1 (cluster ID) to the left, before doing one
additional shift for non-M3W chips.

Fixes: 572afdce53ea ("plat: rcar: Derive core map from PRR")

Reported-by: Oleksandr Grytsov <oleksandr_grytsov@epam.com>
Tested-by: Volodymyr Babchuk <volodymyr_babchuk@epam.com> (R-Car M3)
Tested-by: Oleksandr Grytsov <oleksandr_grytsov@epam.com> (R-Car H3)
Signed-off-by: Volodymyr Babchuk <volodymyr_babchuk@epam.com>

### DIFF
--- a/core/arch/arm/plat-rcar/core_pos_a64.S
+++ b/core/arch/arm/plat-rcar/core_pos_a64.S
@@ -64,9 +64,10 @@ FUNC get_core_pos_mpidr , :
 1:	mov	w2, #PRR_PRODUCT_M3W
 	and	w3, w3, #PRR_PRODUCT_MASK
 	cmp	w2, w3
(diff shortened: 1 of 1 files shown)


jforissier (commented):

Please wrap the long line in the commit description so we get a clean checkpatch report.

lorc:
Thanks. Added tag and fixed commit message formatting.

Issue #5254, opened 2022-04-07
Weird secure monitor issue with mixed fast smc and std smc calls


vesajaaskelainen:
I have been working on refurbishing Xilinx Zynq-7000 SoC (Cortex-A9 x 2) support in OP-TEE to make it possible to boot u-boot and Linux kernel. OP-TEE OS is from tag 3.16.0 + changes.

With the work I have started using SCMI support that @etienne-lms has been introducing in different components.

I have implemented clock driver for zynq7k and several fast smc calls for quick path to query certain hardware details and reworked zynq7k support altogether.

System works so that U-Boot's SPL loads OP-TEE and u-boot proper and then SPL boots into OP-TEE and then when it is done then starts executing u-boot proper and then u-boot's TEE queries with fast smc calls features supported and then switches to SCMI support to fetch information about what clocks are there with std smc calls and to SCMI server and then u-boot gets relocated and then it goes to ask again what features are supported but that fails with:

```
E/TC:0   assertion 'thread_get_exceptions() == THREAD_EXCP_ALL' failed at core/arch/arm/kernel/thread_optee_smc.c:47 <thread_handle_fast_smc>
E/TC:0   Panic at core/kernel/assert.c:28 <_assert_break>
```

I have been looking with the Trace32 what could be the issue and currently it looks like that effect starts from `sm_from_nsec()`:

https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/sm/sm.c#L93

and then later in:

https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/sm/sm_a32.S#L196

actually causes FIQ bit to be unmasked.

and then actual operation is performed but thread exception check happens and causes panic.

Seems that contents of `struct sm_ctx *ctx` is wrong.

I am a bit puzzled what would be the right course of action as there seems to be quite a bit in play in here.

While stepping in secure monitor I can see that in some cases the FIQ bit is not set and my understanding is that if there would be IRQ handler for secure peripheral then FIQ would be used to signal it.

Any ideas what to try or look?

I have added some debug prints to figure out what kind of SMC calls we have.
```
U-Boot SPL 2022.01 (Feb 14 2022 - 16:26:27 +0000)
Trying to boot from RAM
## Checking hash(es) for config conf ... OK
## Checking hash(es) for Image optee ... sha256+ OK
## Checking hash(es) for Image uboot ... sha256+ OK
## Checking hash(es) for Image fdt ... sha256+ OK
[...]

PR #5520, opened 2022-09-04
ARMv7 SCTRL.NMFI workaround support

vesajaaskelainen:
If the ARMv7 Cortex-A core is configured with Non-maskable FIQ (NMFI) support there are side effects that FIQ can only be masked during exception entry and once unmasked by software it cannot anymore be masked.

Side effects of this is that on systems with Non-maskable FIQ (NMFI) support enabled OP-TEE's atomic contexts are trying to mask all interrupts including FIQ but FIQ cannot be masked thus leaving it non-masked. This state is then carried back into secure monitor's sm_ctx structure thus causing next call to SMC entry having incorrect FIQ masking configured.

This issue can represent itself as:
```
E/TC:0   assertion 'thread_get_exceptions() == THREAD_EXCP_ALL' failed at core/arch/arm/kernel/thread_optee_smc.c:50 <thread_handle_fast_smc>
```

Having FIQ non-masked during atomic contexts can cause random side effects.

In order to safeguard from the issue FIQ needs to be un-used in system design. When workaround is enabled and if FIQ would be enabled OP-TEE will cause panic to detect the issue.

As it can be unclear which systems do have the NMFI enabled there is verification code in OP-TEE's startup when debug levels are enabled to check for the matter.

Xilinx Zynq-7000's ARM Cortex-A9 is configured with NMFI support so enable the workaround for it.

ARMv8 architecture is not affected as the Non-maskable FIQ support is not available in there.

Fixes: https://github.com/OP-TEE/optee_os/issues/5254


vesajaaskelainen:
TODO: Verify that FIQ actually causes the panic.

jenswi-linaro:
From the ARM ARMv7 regarding non-maskable FIQs:
_Some ARMv7 implementations can be configured so that the CPSR.F bit cannot be set to 1 by an MSR or CPS instruction._

What if we set the F-bit by other means, that is as we do in an exception return using the `movs` instruction? That should make FIQ a bit more useful on those platforms.

vesajaaskelainen:
> From the ARM ARMv7 regarding non-maskable FIQs: _Some ARMv7 implementations can be configured so that the CPSR.F bit cannot be set to 1 by an MSR or CPS instruction._
> 
> What if we set the F-bit by other means, that is as we do in an exception return using the `movs` instruction? That should make FIQ a bit more useful on those platforms.

This is partly what is done in here. We prepare so that just before handler entry we set the F flag. (ctx->sec.mon_spsr).

In theory one could perhaps do some exception tiggering operation that would then do some similar magic.

jenswi-linaro:
If I have understood it correctly we must avoid using the `cps` instruction to set the F bit, `movs` is another way of doing that. I suppose it partially defeats the purpose with NMFI though.

jforissier:
Is there any way the problematic situation could be detected at runtime without need for any new `CFG_`?

vesajaaskelainen:
> Is there any way the problematic situation could be detected at runtime without need for any new `CFG_`?

You can `read_sctlr()` and check for the bit. But that adds overhead for all invocations for all platforms so I am wondering if it is a good idea or not.

core/arch/arm/sm/sm.c
```diff
@@ -93,6 +93,10 @@ uint32_t sm_from_nsec(struct sm_ctx *ctx)
 	sm_restore_unbanked_regs(&ctx->sec.ub_regs);
 
 	memcpy(&ctx->sec.r0, args, sizeof(*args));
+#ifdef CFG_CORE_WORKAROUND_ARM_NMFI
+	/* Make sure FIQ is masked when jumping to SMC entry. */
+	ctx->sec.mon_spsr |= CPSR_F;
```
jforissier, line 98: Couldn't it be done once per CPU in `init_sec_mon()` where we currently have `sec_ctx->mon_spsr = CPSR_MODE_SVC | CPSR_I;`?

vesajaaskelainen: Unfortunately no. In first times we enter in here (and without the feature) the FIQ is masked.

But due to FIQ unmasking and then trying to mask it again (which fails) then it stores back the MON SPSR without the FIQ mask.

jenswi-linaro:
When revisiting this PR with fresh eyes I think it makes a lot of sense. I'm going to give a few comments but overall it's OK.

core/arch/arm/kernel/abort.c
```diff
@@ -568,3 +568,15 @@ void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs)
 		break;
 	}
 }
+
+#ifdef CFG_CORE_WORKAROUND_ARM_NMFI
+void __noreturn itr_core_handler(void)
+{
+	/*
+	 * Note: removed __weak so that when workaround is enabled and if there
```
jenswi-linaro, line 576: Note: overrides the default implementation of this function so that ...

core/arch/arm/kernel/abort.c
```diff
@@ -568,3 +568,15 @@ void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs)
 		break;
 	}
 }
+
+#ifdef CFG_CORE_WORKAROUND_ARM_NMFI
```
jenswi-linaro, line 572: This code should probably be put in `core/arch/arm/kernel/thread.c` instead.

core/arch/arm/kernel/boot.c
```diff
@@ -1292,6 +1292,17 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
...
 }
 
+#if defined(ARM32) && (CFG_TEE_CORE_LOG_LEVEL >= TRACE_DEBUG)
+static bool cpu_nmfi_enabled(void)
+{
+	uint32_t sctlr = 0;
+
+	sctlr = read_sctlr();
+
+	return !!(sctlr & SCTLR_NMFI);
```
jenswi-linaro, line 1302: `return sctlr & SCTLR_NMFI;` should be enough, the C standard already delivers what you try to do with `!!`.

core/arch/arm/kernel/boot.c
```diff
@@ -1292,6 +1292,17 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
 	init_sec_mon(nsec_entry);
 }
 
+#if defined(ARM32) && (CFG_TEE_CORE_LOG_LEVEL >= TRACE_DEBUG)
+static bool cpu_nmfi_enabled(void)
+{
+	uint32_t sctlr = 0;
```
jenswi-linaro, line 1298:
``` C
if (IS_ENALBED(ARM32))
    return read_sctlr() & SCTLR_NMFI;
return false;
```

core/arch/arm/kernel/boot.c
```diff
@@ -1318,6 +1329,19 @@ void __weak boot_init_primary_late(unsigned long fdt)
 		DMSG("Memory tagging %s",
 		     memtag_is_enabled() ?  "enabled" : "disabled");
 
+	/* Check if platform needs NMFI workaround */
+#if defined(ARM32) && (CFG_TEE_CORE_LOG_LEVEL >= TRACE_DEBUG)
+	if (cpu_nmfi_enabled())	{
+#ifndef CFG_CORE_WORKAROUND_ARM_NMFI
```
jenswi-linaro, line 1335: `if (!IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))`

core/arch/arm/kernel/boot.c
```diff
@@ -1318,6 +1329,19 @@ void __weak boot_init_primary_late(unsigned long fdt)
...
+		IMSG("WARNING: This ARM core has NMFI enabled, please apply workaround!");
+#endif
+	} else {
+#ifdef CFG_CORE_WORKAROUND_ARM_NMFI
```
jenswi-linaro, line 1339: `if (IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))`

core/arch/arm/sm/sm.c
```diff
@@ -93,6 +93,10 @@ uint32_t sm_from_nsec(struct sm_ctx *ctx)
 	sm_restore_unbanked_regs(&ctx->sec.ub_regs);
 
 	memcpy(&ctx->sec.r0, args, sizeof(*args));
+#ifdef CFG_CORE_WORKAROUND_ARM_NMFI
```
jenswi-linaro, line 96: Please use `if (IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))` instead.

vesajaaskelainen: OK

core/arch/arm/kernel/boot.c
vesajaaskelainen: OK

core/arch/arm/kernel/boot.c
vesajaaskelainen: OK

core/arch/arm/kernel/boot.c
vesajaaskelainen: OK

core/arch/arm/kernel/boot.c
vesajaaskelainen: OK

core/arch/arm/kernel/abort.c
vesajaaskelainen: Ok. Lets see.

core/arch/arm/kernel/abort.c
vesajaaskelainen: OK

core/arch/arm/kernel/boot.c
vesajaaskelainen: I backtrack a bit in here as `SCTLR_NMFI` is not defined for `ARM64`.

vesajaaskelainen:
Testing with Xilinx Zynq-7000 without debugs enabled. 

Note: In top of these changes I have my other Xilinx Zynq-7000 changes that are not yet in PRs.

System booted and then ran `xtest`:
```
29910 subtests of which 0 failed
123 test cases of which 0 failed
0 test cases were skipped
TEE test application done!
```

Enabling debugs with `CFG_TEE_CORE_LOG_LEVEL=4`.

Verified that there were no warnigns:
```
I/TC: OP-TEE version: 3.19.0-64-gfe0d15fd05-dev (gcc version 9.3.0 (GCC)) #1 Fri Oct 21 09:28:08 UTC 2022 arm
```

Simulating that `cpu_nmfi_enabled()` returns `false` (with `CFG_CORE_WORKAROUND_ARM_NMFI`).

```
I/TC: OP-TEE version: 3.19.0-65-gb95dddb802-dev (gcc version 9.3.0 (GCC)) #1 Fri Oct 21 09:43:35 UTC 2022 arm
D/TC:0 0 boot_init_primary_late:1346 Executing at offset 0 with virtual load address 0x3c000000
I/TC: WARNING: This ARM core does not have NMFI enabled, no need for workaround.
```

And device boots to Linux. Ran `xtest` and it succeeded (as workaround is still enabled) after a while as logs slow it down quite a bit.

Then returning `cpu_nmfi_enabled()` back to normal (returns true on this SoC) with `CFG_CORE_WORKAROUND_ARM_NMFI` disabled. Eg. simulating a SoC where the developer is not yet aware that NMFI workaround needs to be present.

Then we get the expected warning:
```
I/TC: OP-TEE version: 3.19.0-65-gf83afec52f-dev (gcc version 9.3.0 (GCC)) #1 Fri Oct 21 09:54:52 UTC 2022 arm
D/TC:0 0 boot_init_primary_late:1346 Executing at offset 0 with virtual load address 0x3c000000
I/TC: WARNING: This ARM core has NMFI enabled, please apply workaround!
```

And on this SoC we also get the panic during the boot:
```
E/TC:0   assertion 'thread_get_exceptions() == THREAD_EXCP_ALL' failed at core/arch/arm/kernel/thread_optee_smc.c:50 <thread_handle_fast_smc>
E/TC:0   Panic at core/kernel/assert.c:28 <_assert_break>
E/TC:0   TEE load address @ 0x3c000000
E/TC:0   Call stack:
E/TC:0    0x3c004465
E/TC:0    0x3c00cba1
E/TC:0    0x3c00b133
E/TC:0    0x3c002a91
E/TC:0    0x3c0004b8
```

All seems behave as expected on this SoC.

core/arch/arm/kernel/boot.c
```diff
@@ -1338,6 +1350,17 @@ void __weak boot_init_primary_late(unsigned long fdt)
...
 		     memtag_is_enabled() ?  "enabled" : "disabled");
 
+	/* Check if platform needs NMFI workaround */
+#if CFG_TEE_CORE_LOG_LEVEL >= TRACE_DEBUG
+	if (cpu_nmfi_enabled())	{
+		if (!IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))
+			IMSG("WARNING: This ARM core has NMFI enabled, please apply workaround!");
+	} else {
+		if (IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))
+			IMSG("WARNING: This ARM core does not have NMFI enabled, no need for workaround.");
```
jforissier, line 1360: Nit: preferred style is to omit full stop (`.`) when there is single sentence in log messages. Same in comments: `/* One sentence, no full stop */` but `/* Two sentences. Full stops needed. */`.

jforissier (approved):
One minor comment, other than that:

core/arch/arm/kernel/boot.c
vesajaaskelainen: OK. I am not native English speaker -- there are two parts in here -- but I assume this is what you are looking for:
```
IMSG("WARNING: This ARM core does not have NMFI enabled, no need for workaround");
```
eg. dot removed from the end. 

Will update as recommended.

vesajaaskelainen:
Updated the PR to address review comments. Applied tags.

Rebased on top of master.

Thanks for the review!

vesajaaskelainen:
@jforissier I suppose IBART failure is not related to this PR ?

jforissier:
> @jforissier I suppose IBART failure is not related to this PR ?

Probably not, I've seen similar failures occasionally where IBART doesn't go past the clone phase. @jbech-linaro any idea?

core/arch/arm/kernel/boot.c
```diff
@@ -1338,6 +1350,17 @@ void __weak boot_init_primary_late(unsigned long fdt)
 		DMSG("Memory tagging %s",
 		     memtag_is_enabled() ?  "enabled" : "disabled");
 
+	/* Check if platform needs NMFI workaround */
+#if CFG_TEE_CORE_LOG_LEVEL >= TRACE_DEBUG
```
jenswi-linaro, line 1354: Is this `#if` needed?

vesajaaskelainen: My idea was that it is a waste in a way and in normal images there is no such need. Whoever is porting OP-TEE most likely runs with traces enabled and thus would get the notes.

In a way we could even have separate porting check code that tries to probe some things that needs configuration -- if that is doable / viable.

jenswi-linaro: You're concerned that the call to `read_sctlr()` in `cpu_nmfi_enabled()` might not be optimized away by the kernel even if the `IMSG()` doesn't do anything?

etienne-lms: > My idea was that it is a waste in a way and in normal images there is no such need. Whoever is porting OP-TEE most likely runs with traces enabled and thus would get the notes.

I think `IMSG()` already provides a accurate trace level for these messages. I don't neither think the `#if/#endif` in `TRACE_LEVEL`.

As for the optimization view, as this is optee initialization, i guess it's ok reading few more CPU registers in this sequence.

etienne-lms (commented):
"arm32: correct bit define for SCTLR.NMFI setting"
"arm.mk: Add CFG_CORE_WORKAROUND_ARM_NMFI for NMFI problem"
"arm32: kernel: Add common itr_core_handler() for NMFI workaround" (in commit message: s/in system level/at system level/)

`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit:
plat-zynq7k: Configure platform needing NMFI workaround

core/arch/arm/kernel/boot.c
```diff
@@ -1357,7 +1357,7 @@ void __weak boot_init_primary_late(unsigned long fdt)
 			IMSG("WARNING: This ARM core has NMFI enabled, please apply workaround!");
 	} else {
 		if (IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))
-			IMSG("WARNING: This ARM core does not have NMFI enabled, no need for workaround.");
+			IMSG("WARNING: This ARM core does not have NMFI enabled, no need for workaround");
```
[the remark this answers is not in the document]
vesajaaskelainen: mybad... sorry for that.

core/arch/arm/sm/sm.c
```diff
@@ -93,6 +93,11 @@ uint32_t sm_from_nsec(struct sm_ctx *ctx)
 	sm_restore_unbanked_regs(&ctx->sec.ub_regs);
 
 	memcpy(&ctx->sec.r0, args, sizeof(*args));
+
+	if (IS_ENABLED(CFG_CORE_WORKAROUND_ARM_NMFI))
+		/* Make sure FIQ is masked when jumping to SMC entry. */
+		ctx->sec.mon_spsr |= CPSR_F;
```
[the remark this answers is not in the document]
vesajaaskelainen: OK. Will go for upper one.

core/arch/arm/kernel/boot.c
```diff
@@ -1312,6 +1312,18 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
 	init_sec_mon(nsec_entry);
 }
 
+#if CFG_TEE_CORE_LOG_LEVEL >= TRACE_DEBUG
+static bool cpu_nmfi_enabled(void)
```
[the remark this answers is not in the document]
vesajaaskelainen: When removing #if check then it always needs to be there.

core/arch/arm/kernel/boot.c
vesajaaskelainen: OK. Removing those #if's.

jenswi-linaro:

vesajaaskelainen:

Thanks for the review! Applied tags.

Updated the commit `arm32: kernel: Add check whether core needs NMFI workaround` to have matching description after the changes.

PR #5522, opened 2022-09-05
ffa: Export ffa.h to be used by host

jellesels-arm:
The ffa.h file is needed by the OP-TEE test suite. Export it so it can
be used by it. The corresponding test PR can be found [here](https://github.com/OP-TEE/optee_test/pull/614) 

Signed-off-by: Jelle Sels <jelle.sels@arm.com>

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -909,3 +909,6 @@ CFG_DRIVERS_TPM2_MMIO ?= n
 ifeq ($(CFG_CORE_TPM_EVENT_LOG),y)
 CFG_CORE_TCG_PROVIDER ?= $(CFG_DRIVERS_TPM2)
 endif
+
+# Enable the FF-A SPMC tests in xtests
+CFG_SPMC_TESTS ?= n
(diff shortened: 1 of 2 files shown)


core/arch/arm/include/ffa.h
```diff
@@ -9,7 +9,9 @@
 
 /* This is based on the FF-A 1.0 EAC specification */
 
+#if defined(ARM64) || defined(ARM32)
```
jforissier, line 12: Unrelated change?

jellesels-arm: No, it wasn't. However I changed it to be more clear.
For the spmc tests in xtest we need the ffa.h file, the ffa.h file needs the smccc.h. So I changed it to copy both files when the SPMC_TEST are selected.

jforissier (commented):

jenswi-linaro (commented):

etienne-lms (commented):
Mention `CFG_SPMC_TESTS` in _mk/config.mk_?

jellesels-arm:
> Mention `CFG_SPMC_TESTS` in _mk/config.mk_?

Ack

jellesels-arm:
Any more comments on this? I would we good if we could merge the SPMC testing before the 3.19 release.

PR #5524, opened 2022-09-07
plat-stm32mp1: cleaning ST boards DTS files

etienne-lms:


core/arch/arm/dts/stm32mp157c-ev1.dts
```diff
@@ -250,7 +244,7 @@
 	pinctrl-names = "default", "sleep";
 	pinctrl-0 = <&m_can1_pins_a>;
 	pinctrl-1 = <&m_can1_sleep_pins_a>;
-	status = "okay";
+	s	status = "disabled";atus = "okay";
```
jneuhauser, line 247: Seems broken

etienne-lms: indeed! thanks.

etienne-lms:
I think OP-TEE DTS files will have to diverge at some point from the Linux kernel reference DTS files because OP-TEE does not use the same devices as Linux. I preferred here to disable unused nodes rather than removing then to ease re-sync for Linux DTS's.

PR #5525, opened 2022-09-07
plat-stm32mp1: reset system using rstctrl device

etienne-lms:
Implements PSCI system reset support in STM32MP13 by using rstctrl API to reset the system rather than a SoC specific resource in plat-stm32mp1.


jenswi-linaro (commented):

etienne-lms:
Tag applied
Thanks

PR #5527, opened 2022-09-08
core: boot: do not allocate memory region if optee node already exists

GseoC:
This P-R introduces the removal of the addition of OP-TEE node in U-Boot device-tree if it already exists. It is needed after:

http://patchwork.ozlabs.org/project/uboot/patch/20220907134201.1.I8c8568a1dada00ca09ce1f3d5abaf38e6fdf66d5@changeid/

If the OP-TEE firmware node is already present in the device tree, there is no need to duplicate the node. Until now, there
was no issue with that because the duplicated node (addition of "optee_core" node by mark_tzdram_as_reserved())  covered the same memory region. With the attachment change, OP-TEE base address is now 0xdd000000, which is considered as an overlap because the other node describes a base address of 0xde000000.

OP-TEE is our default security solution and its node should be present in U-Boot/Kernel device-tree.

This changes skips any device-tree updates such as PSCI, resmem, etc... Can we consider that if the OP-TEE node is present,
then the device-tree is fine as it is? What should be the scope?


jenswi-linaro:
A problem with all these "generic" tweaks on how we deal with DTB in various forms is that it has become quite complex. It makes sense, in a way, what you propose. However, you're also proposing an ABI change for all platforms that rely on an updated DTB. We must double-check that all the other platforms are fine with this. Listing the platforms here would be a good start.

etienne-lms (commented):
> This changes skips any device-tree updates such as PSCI, resmem, etc... Can we consider that if the OP-TEE node is present,
then the device-tree is fine as it is? What should be the scope?

I don't think this scheme will match all platforms. I would rather have a simple logic to test existing reserved-memory as proposed by @jforissier, or introduce a config switch allowing an OP-TEE platform not bother with the DTB passed as boot argument (and which 32bit monitor passes to BL33).

GseoC:
Thank you all for you answers.

There is no real reason for us to update the non-secure device-tree in our implementation. I think that in an ecosystem where OP-TEE is embedded by default, having OP-TEE nodes in the device tree should be mandatory and that what is in the device tree should be considered as correct information. All PSCI/OP-TEE/Reserved memory nodes are already in U-Boot/kernel device-tree. Therefore, OP-TEE should only pass the NS device-tree as a boot argument to U-Boot. What do you think of a CFG_EXTERNAL_DT switch around external device-tree functions? I can submit an implementation with such a switch.

I just wonder what would happen if there's a mismatch between OP-TEE conf and NS world on OP-TEE region range. But maybe that is another debate and maybe we should consider that U-Boot/Kernel OP-TEE nodes and CFG_TZDRAM_START/SIZE are always aligned (Else, it is the check that @jforissier was talking about, but that would mean reading the external device tree).

core/arch/arm/kernel/boot.c
```diff
@@ -1296,13 +1308,19 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
...
+void __weak boot_init_primary_late(unsigned long __maybe_unused fdt)
 {
+#ifdef CFG_EXTERNAL_DT
 	init_external_dt(fdt);
 	tpm_map_log_area(get_external_dt());
 	discover_nsec_memory();
 	update_external_dt();
 	configure_console_from_dt();
+#else
+	tpm_map_log_area(get_embedded_dt());
```
etienne-lms, line 1320: This is not applicable. TPM log buffer info cannot be found from the embedded DT.
Note `tpm_map_log_area(NULL);` will not harm the system but only emit info trace messages.

core/arch/arm/kernel/boot.c
```diff
@@ -1296,13 +1308,19 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
  * Note: this function is weak just to make it possible to exclude it from
  * the unpaged area.
  */
-void __weak boot_init_primary_late(unsigned long fdt)
+void __weak boot_init_primary_late(unsigned long __maybe_unused fdt)
 {
+#ifdef CFG_EXTERNAL_DT
```
etienne-lms, line 1313: I think we should remove this condition and change `init_external_dt()`/`update_external_dt()` to do nothing when `!IS_ENABLED(CFG_EXTERNAL_DT)`.

core/arch/arm/kernel/boot.c
```diff
@@ -82,7 +82,9 @@ struct dt_descriptor {
 #endif
 };
 
+#ifdef CFG_EXTERNAL_DT
 static struct dt_descriptor external_dt __nex_bss;
```
etienne-lms, line 86: I would suggest to remove this condition and add `__maybe_unused` attribute:
`static struct dt_descriptor external_dt __nex_bss __maybe_unused;`

core/arch/arm/kernel/boot.c
```diff
@@ -625,6 +627,7 @@ void *get_embedded_dt(void)
 #endif /*CFG_EMBED_DTB*/
 
 #if defined(CFG_DT)
+#ifdef CFG_EXTERNAL_DT
```
etienne-lms, line 630:
```c
void *get_external_dt(void)
{
	if (!IS_ENABLED(CFG_EXTERNAL_DT))
		retun NULL;

	(...)
```
and same for `update_external_dt()` and `release_external_dt()`.

core/arch/arm/kernel/boot.c
```diff
@@ -1184,7 +1192,11 @@ static void discover_nsec_memory(void)
 	const struct core_mmu_phys_mem *mem_begin = NULL;
 	const struct core_mmu_phys_mem *mem_end = NULL;
 	size_t nelems;
+#ifdef CFG_EXTERNAL_DT
 	void *fdt = get_external_dt();
+#else /* CFG_EXTERNAL_DT */
+	void *fdt = get_embedded_dt();
```
etienne-lms, line 1198: this should be in a dedicated commit: get NSEC memory from embedded DT when CFG_EXTERNAL_DT is disabled.

core/arch/arm/plat-stm32mp1/main.c
```diff
@@ -121,7 +121,7 @@ static TEE_Result init_console_from_dt(void)
 	fdt = get_embedded_dt();
 	res = get_console_node_from_dt(fdt, &node, NULL, NULL);
 	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
-		fdt = get_external_dt();
+		fdt = get_dt();
```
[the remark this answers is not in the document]
GseoC: Oops

etienne-lms (commented):
Looks ok to me. @GseoC, could you address latest comments and squash straight the fixup commits?

GseoC:
Updated with Etienne's comments

GseoC:
Fixed a "Signed-off-by"

core/arch/arm/kernel/boot.c
```diff
@@ -82,7 +82,7 @@ struct dt_descriptor {
 #endif
 };
 
-static struct dt_descriptor external_dt __nex_bss;
+static struct dt_descriptor external_dt __nex_bss __maybe_unused;
```
etienne-lms, line 85: `__maybe_unused` not needed since access to the resource is conditioned with `if (...IS_ENABLED(...)...)` instead of `#if`/`#endif`.

core/arch/arm/kernel/boot.c
```diff
@@ -1196,7 +1196,11 @@ static void discover_nsec_memory(void)
 	const struct core_mmu_phys_mem *mem_begin = NULL;
 	const struct core_mmu_phys_mem *mem_end = NULL;
 	size_t nelems;
+#ifdef CFG_EXTERNAL_DT
 	void *fdt = get_external_dt();
+#else /* CFG_EXTERNAL_DT */
+	void *fdt = get_embedded_dt();
+#endif /* CFG_EXTERNAL_DT */
```
etienne-lms, line 1203: prefer:
```patch
 	void *fdt = get_external_dt();
 
+	if (!fdt)
+		fdt = get_embedded_dt();
 
 	if (fdt) {
```

I have some trouble with this sequence: here we try external and fallback to embedded. That is not very consistent with [`configure_console_from_dt()`](https://elixir.bootlin.com/op-tee/3.18.0/source/core/kernel/console.c#L129) and other places calling [`get_dt()`](https://elixir.bootlin.com/op-tee/3.18.0/source/core/arch/arm/kernel/boot.c#L592) that does the opposite.
As we have quite a legacy here on getting DRAM info from external DT, I would not replace line 1200 with a call to `get_dt()`. Since platform can use `register_ddr()` to have these information embedded, I think this change is not mandated by introduction of the new config switch `CFG_EXTERNAL_DT` and should be further discussed.

etienne-lms (commented):
For commit "core: boot: do not force implement the external device-tree ABI":
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment addressed.
Maybe be more explicitly in commit message aoubt CFG_EXTERNAL_DT defaulting to CFG_DT value:
```patch
-This ABI stays implemented by default when a device tree is present
-for legacy purpose.
+CFG_EXTERNAL_DT default value is defined by CFG_DT for backward compatibility
+of OP-TEE default configuration.
```

For commit "plat-stm32mp1: conf: default disable CFG_EXTERNAL_DT": 
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`. Could you rephrase the commit message? E.g.:
```patch
-Default configuration for stm32mp1 platforms shouldn't use
-the external device tree as it is never accessed.
+Default configuration for stm32mp1 does not access external DTB.
```

core/arch/arm/kernel/boot.c
GseoC: Ok, dropping this patch

core/arch/arm/kernel/boot.c
GseoC: Indeed

GseoC:
Updated with @etienne-lms 's comments

core/arch/arm/kernel/boot.c
GseoC: IMO, it is not an issue if it falls back every time when CFG_EXTERNAL_DT=n since every platform have these configs embedded.

Maybe it should be explicitly said to be careful to double-check the embedded values as there will be no further fallback

etienne-lms (commented):
Could rephrase the commit message, the issue addressed by this change is rather 32bit arch related.
```
Do not implement external device tree ABI if CFG_EXTERNAL_DT=n. Some
ecosystem implementations do not require OP-TEE to modify or use this
external device tree. This change is useful on 32bit systems where 
OP-TEE only needs to pass BL33 DTB base address provided by earlier
boot stage as TF-A BL2.

CFG_EXTERNAL_DT default value is defined by CFG_DT for backward
compatibility of OP-TEE default configuration.
```

GseoC:
Rephrased

etienne-lms (commented):
Looks good the me. Thanks.
@jenswi-linaro, @jforissier and others, what do you think about this new switch?

mk/config.mk
```diff
@@ -441,6 +441,7 @@ ifeq ($(CFG_EMBED_DTB),y)
 $(call force,CFG_DT,y)
 endif
 CFG_EMBED_DTB ?= n
+CFG_EXTERNAL_DT ?= $(CFG_DT)
```
jenswi-linaro, line 444: This line must be below the `CFT_DT ?=` line.

GseoC:
Updated with @jenswi-linaro 's comment

mk/config.mk
```diff
@@ -442,6 +442,7 @@ $(call force,CFG_DT,y)
 endif
 CFG_EMBED_DTB ?= n
 CFG_DT ?= n
+CFG_EXTERNAL_DT ?= $(CFG_DT)
```
jenswi-linaro, line 445: Please add a description of this in the comment above.

GseoC: Added a description in the comment

jenswi-linaro:

GseoC:
Thank you for the review.

Updated with Jens' tag.

jforissier:
I think it makes sense to take this in 3.19.0. I will merge the PR if there is no objection.

PR #5530, opened 2022-09-09
core: rpc: log error when allocation fails in non-secure world

jforissier:
The error messages printed when the non-secure world fails to allocate shared memory may be misleading misleading:

 E/LD:  init_elf:453 sys_open_ta_bin(cb3e5ba0-adf1-11e0-998b-0002a5d5c51b)
 E/TC:? 0 ldelf_init_with_ldelf:131 ldelf failed with res: 0xffff000c

This looks like an out-of-memory condition (0xffff000c = TEE_ERROR_OUT_OF_MEMORY) in the TEE core ("E/TC"), in other words insufficient core heap space. Add the following message to help pinpoint the issue:

 [Edit]~~E/TC:? 0 get_rpc_alloc_res:645 RPC allocation failed in non-secure world: ret=0xffff000c ret_origin=0x2~~
  E/TC:? 0 get_rpc_alloc_res:645 RPC allocation failed. Non-secure world result: ret=0xffff000c ret_origin=0x2

This situation can be reproduced by killing tee-supplicant and invoking a TA. When the CA is interrupted (Ctrl-C), the above log is printed.

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -616,18 +616,18 @@ static struct mobj *get_rpc_alloc_res(struct optee_msg_arg *arg,
 	paddr_t p = 0;
 
 	if (arg->ret || arg->num_params != 1)
-		return NULL;
+		goto err;
 
 	if (arg->params[0].attr != OPTEE_MSG_ATTR_TYPE_TMEM_OUTPUT  &&
 	    arg->params[0].attr != (OPTEE_MSG_ATTR_TYPE_TMEM_OUTPUT |
 				    OPTEE_MSG_ATTR_NONCONTIG))
-		return NULL;
+		goto err;
 
 	p = arg->params[0].u.tmem.buf_ptr;
 	sz = READ_ONCE(arg->params[0].u.tmem.size);
 	cookie = arg->params[0].u.tmem.shm_ref;
 	if (sz < size)
-		return NULL;
+		goto err;
 
 	if (arg->params[0].attr == OPTEE_MSG_ATTR_TYPE_TMEM_OUTPUT)
 		mobj = rpc_shm_mobj_alloc(p, sz, cookie);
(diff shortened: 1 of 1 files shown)


core/arch/arm/kernel/thread_optee_smc.c
```diff
@@ -636,11 +636,14 @@ static struct mobj *get_rpc_alloc_res(struct optee_msg_arg *arg,
...
-		return NULL;
+		goto out;
 	}
 
 	assert(mobj_is_nonsec(mobj));
-
+out:
+	if (!mobj)
+		EMSG("RPC allocation failed in non-secure world: ret=%#"PRIx32
+		     " ret_origin=%#"PRIx32, arg->ret, arg->ret_origin);
```
etienne-lms, line 646: The log message seems ok when `ret` reports an error but when `ret` is 0, the failure comes from another origin, either issue with the RPC ABI or out-of-mem in TEE core. We may not be able to distinguish the 2. Maybe rephrase:
```patch
-		EMSG("RPC allocation failed in non-secure world: ret=%#"PRIx32
+		EMSG("RPC allocation failed. Non-secure world result: ret=%#"PRIx32

Comment (jforissier) [reply core/arch/arm/kernel/thread_optee_smc.c]: Agreed.

**Review by etienne-lms — COMMENTED** [review]

File: core/arch/arm/kernel/thread_optee_smc.c [hunk 2]
```diff
@@ -636,11 +636,14 @@ static struct mobj *get_rpc_alloc_res(struct optee_msg_arg *arg,
 
 	if (!mobj) {
 		thread_rpc_free(bt, cookie, mobj);
-		return NULL;
+		goto out;
 	}
 
 	assert(mobj_is_nonsec(mobj));
-
+out:
```
jenswi-linaro, line 643: Since we only do `goto out` on error and there's not any common cleanup anyway, let's ahve an `err:` label instead which doesn't need to test if `mobj` is NULL or not since it always will be so.

jforissier: OK, updated

etienne-lms (commented):
My R-b still applies.

PR #5538, opened 2022-09-19
core: add missing pauth key saving in foreign interrupt handler

jenswi-linaro:
When a foreign interrupt (non-secure) is trapped in OP-TEE the state of the current thread is saved similarly to when an RPC is performed.

With pointer authentication enabled two more registers which are part of the current context, APIAKEYHI-EL1 and APIAKEYLO-EL1, are added. These registers contains a key needed for pointer authentication. This key is unique per context so it must always be saved and restored when switching context.

Prior to this patch the step where this key is saved in the foreign interrupt handler was missing, so fix this by adding the missing step.

Fixes: 2b06f9dede33 ("Add basic pointer authentication support for TA's")

### DIFF
--- a/core/arch/arm/kernel/thread_a64.S
+++ b/core/arch/arm/kernel/thread_a64.S
@@ -1071,6 +1071,12 @@ END_FUNC el0_sync_abort
 	/* Save original x0..x3 */
 	store_xregs x0, THREAD_CTX_REGS_X0, 10, 13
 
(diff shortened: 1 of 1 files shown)


jforissier (commented):

etienne-lms (commented):

jenswi-linaro:
Tags applied.

PR #5541, opened 2022-09-20
core: storage: improve SE05x deletion of NVM objects

ldts:
Keys created on the Secure Element NVM via the PKCS#11 TA are removed by scanning the data buffer holding the reference to the key during the release of the object.

The storage allocated to hold those keys (ECC/RSA) is always below the page size length which seems like a reasonable figure to use for future extensions.

This commit avoids scanning objects larger than that length.

Incidentally also fixes regression 6018: this test releases an object of size 0xA0000 which can't be scanned due to this part of the code hitting an OOM.

### DIFF
--- a/core/drivers/crypto/se050/core/storage.c
+++ b/core/drivers/crypto/se050/core/storage.c
@@ -5,22 +5,49 @@
  */
 
 #include <crypto/crypto.h>
+#include <mm/core_mmu.h>
 #include <se050.h>
 #include <se050_utils.h>
 #include <string.h>
+#include <tee/tee_fs.h>
+#include <tee/tee_obj.h>
+#include <tee/tee_pobj.h>
 
-void crypto_storage_obj_del(uint8_t *data, size_t len)
+TEE_Result crypto_storage_obj_del(struct tee_obj *o)
 {
 	sss_status_t status = kStatus_SSS_Success;
 	uint32_t val = SE050_KEY_WATERMARK;
+	TEE_Result ret = TEE_ERROR_GENERIC;
 	sss_se05x_object_t k_object = { };
+	uint8_t *data = NULL;
+	uint8_t *p = NULL;
 	bool found = false;
-	uint8_t *p = data;
+	size_t len = 0;
 
-	if (!p)
-		return;
+	if (!o)
+		return TEE_ERROR_BAD_PARAMETERS;
 
-	while (len > sizeof(uint64_t) && !found) {
+	len = o->info.dataSize;
+
+	/* Supported keys (ECC/RSA) require less than 4KB of storage */
+	if (len > SMALL_PAGE_SIZE || len <= sizeof(uint64_t))
+		return TEE_SUCCESS;
+
+	data = calloc(1, len);
+	if (!data)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	/* Read the object into memory */
+	ret = o->pobj->fops->read(o->fh, o->info.dataPosition, data, &len);
+	if (ret) {
+		EMSG("se05x: can not read the object prior removal");
+		free(data);
+		goto out;
+	}
+
+	/* Scan the object for the watermark */
+	p = data;
+	while (len >= sizeof(uint32_t) && !found) {
 		if (memcmp(p, &val, sizeof(val)) != 0) {
 			p++;
 			len--;
@@ -29,22 +56,53 @@ void crypto_storage_obj_del(uint8_t *data, size_t len)
 		found = true;
 	}
 
-	if (!found)
-		return;
+	if (!found) {
+		free(data);
+		return TEE_SUCCESS;
+	}
 
+	/* Retrieve the object identifier */
 	p = p - 4;
 	memcpy((void *)&val, p, sizeof(val));
+	free(data);
 
 	if (val < OID_MIN || val > OID_MAX)
-		return;
+		return TEE_SUCCESS;
 
 	status = sss_se05x_key_object_init(&k_object, se050_kstore);
-	if (status != kStatus_SSS_Success)
-		return;
+	if (status != kStatus_SSS_Success) {
+		ret = TEE_ERROR_BAD_STATE;
+		goto out;
+	}
 
 	status = sss_se05x_key_object_get_handle(&k_object, val);
-	if (status != kStatus_SSS_Success)
-		return;
+	if (status != kStatus_SSS_Success) {
+		EMSG("se05x: can not communicate with the secure element");
+		ret = TEE_ERROR_BAD_STATE;
+		goto out;
+	}
 
-	sss_se05x_key_store_erase_key(se050_kstore, &k_object);
+	status = sss_se05x_key_store_erase_key(se050_kstore, &k_object);
+	if (status != kStatus_SSS_Success) {
+		EMSG("se05x: can not communicate with the secure element");
+		ret = TEE_ERROR_BAD_STATE;
+		goto out;
+	}
+
+out:
+	/*
+	 * Users can delete the SE05X NVM objects during boot using a built
+	 * time configuration flag (CFG_CORE_SE05X_INIT_NVM).
+	 *
+	 * This could cause the deletion of the secure storage objects holding
+	 * references to those IDs via crypto_storage_obj_del() to fail, leaving
+	 * broken links in the file system.
+	 *
+	 * Therefore we only permit this call to block the deletion upon an
+	 * additional specific config.
+	 */
+	if (ret && IS_ENABLED(CFG_CORE_SE05X_BLOCK_OBJ_DEL_ON_ERROR))
+		return ret;
+
+	return TEE_SUCCESS;
 }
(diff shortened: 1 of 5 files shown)


core/tee/tee_svc_storage.c
```diff
@@ -444,7 +445,8 @@ TEE_Result syscall_storage_obj_del(unsigned long obj)
 	if (o->pobj == NULL || o->pobj->obj_id == NULL)
 		return TEE_ERROR_BAD_STATE;
 
-	if (IS_ENABLED(CFG_NXP_SE05X)) {
+	/* Scan buffers that might contain a Secure Element key identifier */
+	if (IS_ENABLED(CFG_NXP_SE05X) && o->info.dataSize < SMALL_PAGE_SIZE) {
 		len = o->info.dataSize;
 		data = calloc(1, len);
```
jenswi-linaro, line 451: You should probably use `mempool_calloc()` and `mempool_free()` with `mempool_default` if can may need to alloc 4k. `mempool_calloc()` is more or less guaranteed to succeed on such small allocation.

core/tee/tee_svc_storage.c
```diff
@@ -444,7 +445,8 @@ TEE_Result syscall_storage_obj_del(unsigned long obj)
 	if (o->pobj == NULL || o->pobj->obj_id == NULL)
 		return TEE_ERROR_BAD_STATE;
 
-	if (IS_ENABLED(CFG_NXP_SE05X)) {
+	/* Scan buffers that might contain a Secure Element key identifier */
+	if (IS_ENABLED(CFG_NXP_SE05X) && o->info.dataSize < SMALL_PAGE_SIZE) {
```
jenswi-linaro, line 449: I'd rather have a special define for this, with a check where these objects are written also to not exceed this.

etienne-lms: This is really SE05X specific.
A comment a bit out of scope of this P-R: as `crypto_storage_obj_del()` is implemented only for SE05X, could this function be renamed to a se050 API function and do what se050 driver expects, leaving this generic `syscall_storage_obj_del()` function with:
```c
	if (IS_ENABLED(CFG_NXP_SE05X))
		se050_maybe_delete_keyid(o);
```
`se050_maybe_delete_keyid()` in _core/drivers/crypto/se050/core/storage.c_ would do whatever SE050 needs to, including what `crypto_storage_obj_del()` does.

ldts: I think `crypto_storage_obj_del` should be left generic (it ha a weak implementation).

core/tee/tee_svc_storage.c
ldts: you are right. doing it next

core/tee/tee_svc_storage.c
```diff
@@ -444,18 +442,8 @@ TEE_Result syscall_storage_obj_del(unsigned long obj)
...
-			return TEE_ERROR_OUT_OF_MEMORY;
-
-		res = o->pobj->fops->read(o->fh, o->info.dataPosition,
-					  data, &len);
-		if (res == TEE_SUCCESS)
-			crypto_storage_obj_del(data, len);
-		free(data);
-	}
+	/* The cryptographic layer might need to do house-keeping */
+	crypto_storage_obj_del(o);
```
jenswi-linaro, line 446: I prefer a configuration where the overhead can be reduced to zero when not needed. With this weak function always being called, we're still adding a few useless instructions for every platform. If this call was guarded with some IS_ENABLED() test, like before, it could be optimized away completely if not needed.

jenswi-linaro:

core/drivers/crypto/se050/core/storage.c
```diff
@@ -5,21 +5,40 @@
...
+void crypto_storage_obj_del(struct tee_obj *o)
 {
 	sss_status_t status = kStatus_SSS_Success;
 	uint32_t val = SE050_KEY_WATERMARK;
 	sss_se05x_object_t k_object = { };
+	uint8_t *data = NULL;
 	bool found = false;
-	uint8_t *p = data;
+	size_t len = 0;
+	uint8_t *p;
```
etienne-lms, line 25: = NULL

core/drivers/crypto/se050/core/storage.c
```diff
@@ -5,21 +5,40 @@
...
 
-	if (!p)
+	if (!o)
 		return;
 
+	len = o->info.dataSize;
+	data = calloc(1, len);
```
etienne-lms: Such failure will silebntly not free data. This might be an issue!
Should return OUT_OF_MEMORY.

core/drivers/crypto/se050/core/storage.c
```diff
@@ -5,21 +5,40 @@
...
+	if (!data)
+		return;
+
+	/* Read the object into memory */
+	if (o->pobj->fops->read(o->fh, o->info.dataPosition, data, &len)) {
```
etienne-lms, line 41: This is not expected. Should at least warn. Panicking is overkilling?
Otherwise, return with a failure status.

etienne-lms (commented):
I think both commits should be squashed to not introduce a commit that may fail on out-of-memory.

ldts:
@etienne-lms I thought necessary to add an additional CFG_ to handle the error. I did my best to explain it in crypto.mk. Please let me know if it is not clear.

ldts:
> I think both commits should be squashed to not introduce a commit that may fail on out-of-memory.

um, ok. I thought it made sense to have them separate but since the SE05x is the only user I guess it wont be an issue.

core/tee/tee_svc_storage.c
ldts: Actually this is a terrible idea leading to a deadlock on 6016. will leave it the way it is. 
This mempool thread dependent allocator is pure evil IMO

jenswi-linaro: Memory leakage will lead to deadlock, fast, but other than that I don't see how this could cause a deadlock.

etienne-lms (commented):
Looks ok to me. `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with very minor remaining comments addressed.

core/drivers/crypto/se050/core/storage.c
```diff
@@ -29,22 +57,52 @@ void crypto_storage_obj_del(uint8_t *data, size_t len)
...
+
+	status = sss_se05x_key_store_erase_key(se050_kstore, &k_object);
+	if (status != kStatus_SSS_Success) {
+		EMSG("se05x: can not communicate with the secure element");
+		ret = TEE_ERROR_BAD_STATE;
+		goto out;
+	}
+
+out:
+	/* Users can delete the SE05X NVM objects during boot using a built
```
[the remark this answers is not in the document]
ldts: sorry I always make this mistake,

ldts:
@etienne-lms @jforissier tags added

ldts:
rebased on master

core/tee/tee_svc_storage.c
ldts: it is easy to reproduce - just run 6016 and it hits in a second or two

```
E/TC:? 3 ree_fs_read:368
E/TC:? 3 ree_fs_read_primitive:310
E/TC:? 3 get_pool:93            -- WAIT LOCK POOL
E/TC:? 3 get_pool:95            -- POOL LOCKED
E/TC:? 3 put_pool:114           -- POOL FINAL UNLOCK
E/TC:? 3 ree_fs_read_primitive:357
E/TC:? 3 ree_fs_read:370
E/TC:? 1 ree_fs_read:368
E/TC:? 1 ree_fs_read_primitive:310
E/TC:? 1 get_pool:93            -- WAIT LOCK POOL
E/TC:? 1 get_pool:95            -- POOL LOCKED
E/TC:? 1 put_pool:114           -- POOL FINAL UNLOCK
E/TC:? 1 ree_fs_read_primitive:357
E/TC:? 1 ree_fs_read:370
E/TC:? 2 get_pool:93            -- WAIT LOCK POOL
E/TC:? 2 get_pool:95            -- POOL LOCKED
E/TC:? 1 get_pool:93            -- WAIT LOCK POOL
E/TC:? 2 put_pool:114           -- POOL FINAL UNLOCK
E/TC:? 1 get_pool:95            -- POOL LOCKED
E/TC:? 2 read_dent:102 !!
E/TC:? 2 ree_fs_read_primitive:310
E/TC:? 2 get_pool:93            -- WAIT LOCK POOL

/* at this point the system locks up  */

```

ldts: I can instrument it further if you think it is worth addressing the issue

jenswi-linaro: Please do, or if you can find a way of reproducing it on QEMU and I'll investigate.

I wonder what thread 1 is doing, waiting for some other mutex?

ldts: the issue is with getting the mempool before taking the ree_fs_lock...is expected to be the other way around

jenswi-linaro: Thanks. The problem is with the locking order, sometimes `ree_fs_mutex` is locked first and sometimes `mempool_default->mu`.

It seems there's no easy way of using the mempool API at this stage.

etienne-lms (commented):

jenswi-linaro:

ldts:
thanks. tags updated.

ldts:
@jforissier rrdy for merge

jforissier:
I'm taking this in 3.19.0 since it fixes a issue with xtest regression_6018.

PR #5543, opened 2022-09-22
drivers: versal: PM service

ldts:
Calls the TF-A exported SiP services.

reference https://github.com/OP-TEE/optee_os/pull/5426

### DIFF
--- a/core/drivers/versal_pm.c
+++ b/core/drivers/versal_pm.c
@@ -1,0 +1,215 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2022 Foundries.io Ltd
+ * Jorge Ramirez-Ortiz <jorge@foundries.io>
+ */
+
+#include <arm.h>
+#include <drivers/versal_mbox.h>
+#include <drivers/versal_pm.h>
+#include <initcall.h>
+#include <kernel/cache_helpers.h>
+#include <kernel/delay.h>
+#include <kernel/panic.h>
+#include <kernel/thread.h>
+#include <mm/core_memprot.h>
+#include <string.h>
+#include <tee/cache.h>
+#include <tee_api_types.h>
+#include <utee_defines.h>
+
+/* VERSAL_SIP_UID: 2ab9e4ec-93b9-11e7-a019dfe0dbad0ae0 */
+#define VERSAL_SIP_UID_0 U(0xece4b92a)
+#define VERSAL_SIP_UID_1 U(0xe711b993)
+#define VERSAL_SIP_UID_2 U(0xe0df19a0)
+#define VERSAL_SIP_UID_3 U(0xe00aaddb)
+#define VERSAL_SIP_MAJOR  0
+#define VERSAL_SIP_MINOR  1
+
+#define VERSAL_SIP_SVC_VERSION		0x8200ff03
+#define VERSAL_SIP_SVC_UID		0x8200ff01
+#define VERSAL_SIP_SVC			0xc2000000
+
+#define PAYLOAD_ARG_CNT		8
+
+/* MBOX IPI */
+#define PM_MODULE_SHIFT		8
+#define PM_MODULE		2
+#define PM_API_ID(x)		((PM_MODULE << PM_MODULE_SHIFT) | (x))
+#define VERSAL_PM_MAJOR		0
+#define VERSAL_PM_MINOR		1
+
+/* PM API ids */
+#define PM_GET_API_VERSION		1
+#define PM_GET_DEVICE_STATUS		3
+#define PM_GET_OP_CHARACTERISTIC	4
+#define PM_REGISTER_NOTIFIER		5
+#define PM_REQ_SUSPEND			6
+#define PM_SELF_SUSPEND			7
+#define PM_FORCE_POWERDOWN		8
+#define PM_ABORT_SUSPEND		9
+#define PM_REQ_WAKEUP			10
+#define PM_SET_WAKEUP_SOURCE		11
+#define PM_SYSTEM_SHUTDOWN		12
+#define PM_REQUEST_DEVICE		13
+#define PM_RELEASE_DEVICE		14
+#define PM_SET_REQUIREMENT		15
+#define PM_SET_MAX_LATENCY		16
+#define PM_RESET_ASSERT			17
+#define PM_RESET_GET_STATUS		18
+#define PM_INIT_FINALIZE		21
+#define PM_GET_CHIPID			24
+#define	PM_PINCTRL_REQUEST		28
+#define	PM_PINCTRL_RELEASE		29
+#define	PM_PINCTRL_GET_FUNCTION		30
+#define	PM_PINCTRL_SET_FUNCTION		31
+#define	PM_PINCTRL_CONFIG_PARAM_GET	32
+#define	PM_PINCTRL_CONFIG_PARAM_SET	33
+#define PM_IOCTL			34
+#define PM_QUERY_DATA			35
+#define PM_CLOCK_ENABLE			36
+#define PM_CLOCK_DISABLE		37
+#define PM_CLOCK_GETSTATE		38
+#define PM_CLOCK_SETDIVIDER		39
+#define PM_CLOCK_GETDIVIDER		40
+#define PM_CLOCK_SETRATE		41
+#define PM_CLOCK_GETRATE		42
+#define PM_CLOCK_SETPARENT		43
+#define PM_CLOCK_GETPARENT		44
+#define PM_PLL_SET_PARAMETER		48
+#define PM_PLL_GET_PARAMETER		49
+#define PM_PLL_SET_MODE			50
+#define PM_PLL_GET_MODE			51
+#define PM_FEATURE_CHECK		63
+
+/* Loader API id */
+#define PM_LOAD_PDI			0x701
+
+/* PDI sources */
+#define PDI_SRC_JTAG		0x0
+#define PDI_SRC_QSPI24		0x1
+#define PDI_SRC_QSPI32		0x2
+#define PDI_SRC_SD0		0x3
+#define PDI_SRC_EMMC0		0x4
+#define PDI_SRC_SD1		0x5
+#define PDI_SRC_EMMC1		0x6
+#define PDI_SRC_USB		0x7
+#define PDI_SRC_OSPI		0x8
+#define PDI_SRC_SBI		0x9
+#define PDI_SRC_SMAP		0xA
+#define PDI_SRC_PCIE		0xB
+#define PDI_SRC_SD1_LS		0xE
+#define PDI_SRC_DDR		0xF
+
+struct versal_sip_payload {
+	uint32_t data[PAYLOAD_ARG_CNT];
+};
+
+static uint32_t versal_sip_call(uint32_t smc_fid, uint32_t arg0, uint32_t arg1,
+				uint32_t arg2, uint32_t arg3,
+				struct versal_sip_payload *payload)
+{
+	struct thread_smc_args args = {
+		.a0 = smc_fid,
+		.a1 = reg_pair_to_64(arg1, arg0),
+		.a2 = reg_pair_to_64(arg3, arg2),
+	};
+
+	thread_smccc(&args);
+
+	if (payload) {
+		reg_pair_from_64(args.a0, &payload->data[1], &payload->data[0]);
+		reg_pair_from_64(args.a1, &payload->data[3], &payload->data[2]);
+		reg_pair_from_64(args.a2, &payload->data[5], &payload->data[4]);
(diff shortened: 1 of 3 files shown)


core/drivers/versal_pm.c
```diff
@@ -0,0 +1,147 @@
...
+
+	return args.a0;
+}
+
+TEE_Result versal_write_fpga(paddr_t pa)
+{
+	struct ipi_cmd cmd = { };
+
+	if (IS_ENABLED(CFG_VERSAL_FPGA_PM_DRIVER))
+		return versal_sip_call(PM_LOAD_PDI, PDI_SRC_DDR, pa, pa >> 32,
```
jenswi-linaro, line 116: After a bit of searching I found this comment in TF-A where the `PM_LOAD_PDI` call is handled:
```
/**
 * eemi_for_compatibility() - EEMI calls handler for deprecated calls
 *
 * @return - If EEMI API found then, uintptr_t type address, else 0
 *
 * Some EEMI API's use case needs to be changed in Linux driver, so they
 * can take advantage of common EEMI handler in TF-A. As of now the old
 * implementation of these APIs are required to maintain backward compatibility
 * until their use case in linux driver changes.
 */
```
It seems you're calling a deprecated function. Any chance of using a non-deprecated function instead?

By the way, it would be nice with a reference to the ABI and which class of platform can expect this ABI to be available.

core/drivers/versal_pm.c
```diff
@@ -0,0 +1,147 @@
...
+TEE_Result versal_soc_version(uint8_t *version)
+{
+	struct versal_sip_payload payload = { };
+	const uint32_t version_shift = 12;
+	uint32_t res = 0;
+
+	if (!version)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	res = versal_sip_call(PM_GET_CHIPID, 0, 0, 0, 0, &payload);
```
jenswi-linaro, line 138: Is there a way to check that we're using the expected ABI?

core/drivers/versal_pm.c
ldts: right. I could just remove it - just thought it might be useful as a backup. 

The preferred call is the one right under it (direct to the PLM instead of going to the TF-A)

core/drivers/versal_pm.c
ldts: Not that I am aware of - at the time of writing, a few months ago - and that is going to be a problem (ie, newer PLM firmware will fix the DMA issue with unaligned accesses to the AES-GCM engine....but at the time the code was written there was no way to check for a version).

this is certainly something that I will need to track and enfoce

core/drivers/versal_pm.c
jenswi-linaro: Backup sounds almost like it's not going to be used or tested regularly.
If it works for you without this perhaps it's better to remove it.

core/drivers/versal_pm.c
ldts: ahh...sorry of course, you meant for the TF-A (I was focused on the PLM which is what is really important TBH, the SIP is only used for the version)

sure yeah, I can query that but I think it is a bit of an overkill in this instance. What we really need is that same call for the PLM

jenswi-linaro: The SIP commands are chip specific so if you're unlucky they might do damage if used on the wrong hardware.

ldts: @nathan-menhorn does AMD Xilinx release the PLM and the TF-A firmwares together or does it do it independently?

I am wondering if we could use TF-A **VERSAL_SIP_SVC_UID** (0x8200ff01) and **VERSAL_SIP_SVC_VERSION**(0x8200ff03) not only to verify the ABI with the secure monitor but also with the PLM and so guarantee that the mailbox driver remains functional.

Perhaps we should do that while the PLM doesnt provide an interface that mbox can use to validate the ABI (I suppose there will be one soon). what do you think?

nathan-menhorn: Hi @ldts, it is safe to make an assumption of use that if you are using any piece of software from one version it is required to use all other pieces of software from the same version. I don't think *VERSAL_SIP_SVC_VERSION* would be the best macro to use as it appears this hasn't been updated for quite some time.

ldts: yes I agree with you but I dont think there is a better choice.

I am not too concerned about about the TF-A but for mailbox it would be great to have a version service to identify regressions due to ABI changes... Do you think it could come on the next PLM release?

ldts: I think there is one IPI that I can use to get to the version (from LIBPM_MODULE_ID).

ok will check both for Mailbox (PLM) and TF-A.

ldts:
```

/****************************************************************************/
/**
 * @brief  This function is used to request the version number of the API
 * running on the platform management controller.
 *
 * @param  Version Returns the API 32-bit version number.
 *
 * @return XST_SUCCESS if successful else XST_FAILURE or an error code
 * or a reason code
 *
 *
 *
 ****************************************************************************/
XStatus XPm_GetApiVersion(u32 *Version)
{
	XStatus Status = (s32)XST_FAILURE;
	u32 Payload[PAYLOAD_ARG_CNT];

	PACK_PAYLOAD0(Payload, PM_GET_API_VERSION);

	/* Send request to the target module */
	Status = XPm_IpiSend(PrimaryProc, Payload);
	if (XST_SUCCESS != Status) {
		goto done;
	}

	/* Return result from IPI return buffer */
	Status = Xpm_IpiReadBuff32(PrimaryProc, Version, NULL, NULL);

done:
	return Status;
}
```

ldts:
@jenswi-linaro I updated with a check to the version.
will do the same for the mbox driver....and for zynqmp

core/drivers/versal_pm.c
```diff
@@ -0,0 +1,192 @@
...
+
+static TEE_Result versal_check_pm_abi(void)
+{
+	struct versal_sip_payload p = { };
+	struct ipi_cmd cmd = { };
+	struct ipi_cmd rsp = { };
+
+	/* Check with TF-A */
+	if (versal_sip_call(VERSAL_SIP_SVC_VERSION, 0, 0, 0, 0, &p))
+		panic();
```
jenswi-linaro, line 174: Is panic necessary, can't we just complain a bit and avoid using this?
By the way, I guess we should check `VERSAL_SIP_SVC_UID` to see that it really is a Versal SIP we're talking to and not some other SIP.

ldts: Yes for our PM usage panic is not necessary - maybe for some of the other services, will just report the error.

Sure I can use the UID. BTW should the checks against the values be with CFG_* instead of built-in macros? Just to avoid having to modify code for updates

jenswi-linaro: No, it's enough to check that the major and minor versions are larger or equal to the expected value. This is explained in the SMC Calling Convention specification.

I don't expect the SIP UID to change.

core/drivers/versal_pm.c
```diff
@@ -0,0 +1,239 @@
...
+static TEE_Result check_sip_uuid(struct versal_sip_payload *p)
+{
+	uint32_t time_low = TEE_U32_BSWAP(p->data[0]);
+	uint16_t time_mid = TEE_U32_BSWAP(p->data[2]) >> 16;
+	uint16_t time_hi  = TEE_U32_BSWAP(p->data[2]);
+	uint8_t uuid[16] = { };
+	size_t i = 0;
+	size_t j = 0;
+
+	for (i = 0; i < sizeof(time_low) && j < sizeof(uuid); i++, j++)
```
jenswi-linaro, line 172: You could simplify this by just checking that the four 32-bit words hold the expected values.
See for example optee_msg_api_uid_is_optee_api() at:

https://github.com/torvalds/linux/blob/2bca25eaeba6190efbfcb38ed169bd7ee43b5aaf/drivers/tee/optee/smc_abi.c#L1140-L1150

You should probably use `tee_uuid_from_octets()` instead if you want to convert this into a TEE_UUID.

core/drivers/versal_pm.c
ldts: ok. I followed what was done in `optee_msg_api_revision_is_compatible` that you pointed to below (so a match on expected major and minor >= expected minor)

core/include/drivers/versal_pm.h
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (C) 2022 Foundries.io Ltd
+ */
+
+#ifndef __DRIVERS_VERSAL_PM_H__
+#define __DRIVERS_VERSAL_PM_H__
+
+#include <platform_config.h>
+#include <tee_api_types.h>
+#include <types_ext.h>
+#include <util.h>
```
etienne-lms, line 12: noitpicking: _platform_config.h_ and _util.h_ are not needed.

core/drivers/versal_pm.c
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	cmd.data[0] = PM_API_ID(PM_GET_API_VERSION);
+	if (versal_mbox_notify(&cmd, &rsp, NULL))
+		return TEE_ERROR_GENERIC;
+
+	major = rsp.data[1] & 0xFFFF;
+	minor = rsp.data[1] >> 16;
+	if (major != VERSAL_PM_MAJOR || minor < VERSAL_PM_MINOR) {
+		EMSG("Invalid PM version: Major %d, Minor %d", major, minor);
```
etienne-lms, line 213: s/%d/%"PRId32"/
or more simply define `major` and `minor` with type `unsigned int`.

core/drivers/versal_pm.c
nathan-menhorn: > 

Hi @ldts, the next release is scheduled to be released in a few weeks - Xilinx has  Spring/Fall release schedule -  and we  won't be able to add this in right away.

ldts:
@etienne-lms updated @jenswi-linaro

jenswi-linaro:
There's still https://github.com/OP-TEE/optee_os/pull/5543/files#r978362366

core/drivers/versal_pm.c
```diff
@@ -0,0 +1,147 @@
...
+{
+	struct ipi_cmd cmd = { };
+
+	if (IS_ENABLED(CFG_VERSAL_FPGA_PM_DRIVER))
+		return versal_sip_call(PM_LOAD_PDI, PDI_SRC_DDR, pa, pa >> 32,
+				       0, NULL);
+
+	cmd.data[0] = PM_LOAD_PDI;
+	cmd.data[1] = PDI_SRC_DDR;
+	cmd.data[2] = pa >> 32;
```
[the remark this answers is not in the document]
ldts: argh sorry. yes thanks for that....I'll use this syntax on all the other files.

core/drivers/versal_pm.c
```diff
@@ -0,0 +1,220 @@
...
+struct versal_sip_payload {
+	uint32_t data[PAYLOAD_ARG_CNT];
+};
+
+static uint32_t versal_sip_call(uint32_t pm_api_id, uint32_t arg0,
+				uint32_t arg1, uint32_t arg2, uint32_t arg3,
+				struct versal_sip_payload *payload)
+{
+	struct thread_smc_args args = {
+		.a0 = pm_api_id > PM_LOAD_PDI ?
```
jenswi-linaro, line 113: I didn't notice this before. Is this needed?

ldts: I am following the separation present in TF-A:
https://review.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/7737fdf0edd4ede45b23a4aef0ae7b2f78a37ad7/plat/xilinx/versal/sip_svc_setup.c#56

jenswi-linaro: Possibly, but here you're mixing some API ID with proper SMC FID, we don't have that mix TF-A.
It would be cleaner to only supply SMC FIDs here, the caller will know if it first needs to make the API ID into an SMC ID or not.

ldts: I dont understand how you would want to reach VERSAL_SIP_SVC_UID or VERSAL_SIP_SVC_VERSION and then what clarity would that set of changes bring to the caller?

jenswi-linaro: Instead of calling as `versal_sip_call(PM_GET_CHIPID, ...)` do `versal_sip_call(VERSAL_SIP_SVC | PM_GET_CHIPID, ...)`. That way we don't need to use heuristics to determine if `VERSAL_SIP_SVC` needs to be or:ed in or not.

ldts: yes you are right

jenswi-linaro:

ldts:
@jforissier this one is ready

etienne-lms (commented):

ldts:
now with tags

PR #5544, opened 2022-09-22
drivers: versal: non volatile memory (eFuse and BBRAM)

ldts:
Provide an interface to access the xilnvm service executing in the PLM firmware running on the Microblaze processor.

reference https://github.com/OP-TEE/optee_os/pull/5426 

### DIFF
--- a/core/drivers/versal_nvm.c
+++ b/core/drivers/versal_nvm.c
@@ -1,0 +1,1005 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2022 Foundries.io Ltd
+ * Jorge Ramirez-Ortiz <jorge@foundries.io>
+ */
+
+#include <arm.h>
+#include <drivers/versal_nvm.h>
+#include <drivers/versal_mbox.h>
+#include <initcall.h>
+#include <kernel/panic.h>
+#include <kernel/tee_misc.h>
+#include <mm/core_memprot.h>
+#include <string.h>
+#include <tee/cache.h>
+
+#include "drivers/versal_nvm.h"
+
+#define NVM_WORD_LEN 4
+
+/* Protocol API with the remote processor */
+#define NVM_MODULE_SHIFT		8
+#define NVM_MODULE			11
+#define NVM_API_ID(_id) ((NVM_MODULE << NVM_MODULE_SHIFT) | (_id))
+
+#define __aligned_efuse			__aligned(CACHELINE_LEN)
+
+/* Internal */
+struct versal_efuse_puf_fuse_addr {
+	uint64_t data_addr;
+	uint32_t start_row;
+	uint32_t num_rows;
+	uint8_t env_monitor_dis;
+	uint8_t prgm_puf_fuse;
+	uint8_t pad[46];
+};
+
+/*
+ * Max size of the buffer needed for the remote processor to DMA efuse _data_
+ * to/from
+ */
+#define EFUSE_MAX_LEN (EFUSE_MAX_USER_FUSES * sizeof(uint32_t))
+
+enum versal_nvm_api_id {
+	API_FEATURES				= 0,
+	BBRAM_WRITE_AES_KEY			= 1,
+	BBRAM_ZEROIZE				= 2,
+	BBRAM_WRITE_USER_DATA			= 3,
+	BBRAM_READ_USER_DATA			= 4,
+	BBRAM_LOCK_WRITE_USER_DATA		= 5,
+	EFUSE_WRITE				= 6,
+	EFUSE_WRITE_PUF				= 7,
+	EFUSE_PUF_USER_FUSE_WRITE		= 8,
+	EFUSE_READ_IV				= 9,
+	EFUSE_READ_REVOCATION_ID		= 10,
+	EFUSE_READ_OFFCHIP_REVOCATION_ID	= 11,
+	EFUSE_READ_USER_FUSES			= 12,
+	EFUSE_READ_MISC_CTRL			= 13,
+	EFUSE_READ_SEC_CTRL			= 14,
+	EFUSE_READ_SEC_MISC1			= 15,
+	EFUSE_READ_BOOT_ENV_CTRL		= 16,
+	EFUSE_READ_PUF_SEC_CTRL			= 17,
+	EFUSE_READ_PPK_HASH			= 18,
+	EFUSE_READ_DEC_EFUSE_ONLY		= 19,
+	EFUSE_READ_DNA				= 20,
+	EFUSE_READ_PUF_USER_FUSES		= 21,
+	EFUSE_READ_PUF				= 22,
+	EFUSE_INVALID				= 23,
+};
+
+/* uint64_t are memory addresses */
+struct versal_efuse_data {
+	uint64_t env_mon_dis_flag;
+	uint64_t aes_key_addr;
+	uint64_t ppk_hash_addr;
+	uint64_t dec_only_addr;
+	uint64_t sec_ctrl_addr;
+	uint64_t misc_ctrl_addr;
+	uint64_t revoke_id_addr;
+	uint64_t iv_addr;
+	uint64_t user_fuse_addr;
+	uint64_t glitch_cfg_addr;
+	uint64_t boot_env_ctrl_addr;
+	uint64_t misc1_ctrl_addr;
+	uint64_t offchip_id_addr;
+	uint8_t pad[24];
+};
+
+/* Helper read and write requests (not part of the protocol) */
+struct versal_nvm_buf {
+	size_t len;
+	void *buf;
+};
+
+struct versal_nvm_read_req {
+	enum versal_nvm_api_id efuse_id;
+	enum versal_nvm_revocation_id revocation_id;
+	enum versal_nvm_offchip_id offchip_id;
+	enum versal_nvm_ppk_type ppk_type;
+	enum versal_nvm_iv_type iv_type;
+	struct versal_nvm_buf ibuf[VERSAL_MAX_IPI_BUF];
+};
+
+struct versal_bbram_data {
+	size_t aes_key_len;
+	uint32_t user_data;
+};
+
+struct versal_nvm_write_req {
+	struct versal_efuse_data data;
+	struct versal_bbram_data bbram;
+	struct versal_nvm_buf ibuf[VERSAL_MAX_IPI_BUF];
+	enum versal_nvm_api_id efuse_id;
+};
+
+static TEE_Result
+prepare_cmd(struct versal_ipi_cmd *cmd, enum versal_nvm_api_id efuse,
+	    struct versal_nvm_buf *ibufs, uint32_t *arg)
+{
+	uint32_t a = 0;
+	uint32_t b = 0;
+	size_t i = 0;
+
+	cmd->data[i++] = NVM_API_ID(efuse);
+	if (arg)
+		cmd->data[i++] = *arg;
+
+	if (!ibufs[0].buf)
+		return TEE_SUCCESS;
+
+	reg_pair_from_64(virt_to_phys(ibufs[0].buf), &b, &a);
+
+	cmd->data[i++] = a;
+	cmd->data[i++] = b;
+
+	for (i = 0; i < VERSAL_MAX_IPI_BUF; i++) {
+		cmd->ibuf[i].mem.alloc_len = ibufs[i].len;
+		cmd->ibuf[i].mem.buf = ibufs[i].buf;
+	}
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result efuse_req(enum versal_nvm_api_id efuse,
+			    struct versal_nvm_buf *ibufs, uint32_t *arg)
+{
+	struct versal_ipi_cmd cmd = { };
+	TEE_Result ret = TEE_SUCCESS;
+
+	ret = prepare_cmd(&cmd, efuse, ibufs, arg);
+	if (ret)
+		return ret;
+
+	ret = versal_mbox_notify(&cmd, NULL, NULL);
+	if (ret)
+		EMSG("Mailbox error");
+
+	return ret;
+}
+
+static TEE_Result versal_alloc_read_buffer(struct versal_nvm_read_req *req)
+{
+	assert(req);
+	req->ibuf[0].len = 1024;
+	req->ibuf[0].buf = alloc_cache_aligned(req->ibuf[0].len);
+	if (!req->ibuf[0].buf)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	return TEE_SUCCESS;
+}
+
+static void versal_free_read_buffer(struct versal_nvm_read_req *req)
+{
+	assert(req);
+	free(req->ibuf[0].buf);
+}
+
+static void *versal_get_read_buffer(struct versal_nvm_read_req *req)
+{
+	assert(req);
+	return req->ibuf[0].buf;
+}
+
+static TEE_Result versal_nvm_read(struct versal_nvm_read_req *req)
+{
+	uint32_t *arg = NULL;
+	uint32_t val = 0;
+
+	if (!req)
+		return TEE_ERROR_GENERIC;
+
+	switch (req->efuse_id) {
+	case EFUSE_READ_DNA:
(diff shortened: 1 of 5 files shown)


core/include/drivers/versal_nvm.h
```diff
@@ -0,0 +1,290 @@
...
+	struct versal_bbram_write_ops const * const write;
+	struct versal_bbram_read_ops const * const read;
+};
+
+struct versal_efuse_ops {
+	struct versal_efuse_write_ops const * const write;
+	struct versal_efuse_read_ops const * const read;
+};
+
+struct versal_efuse_ops const *versal_get_efuse_ops(void);
```
jenswi-linaro, line 287: How come you have chosen a function pointer based interface? Do you expect other implementations of this interface?

By the way, given that here's zero documentation of the different functions perhaps you could give a reference to the used interface at least?

ldts: > How come you have chosen a function pointer based interface? Do you expect other implementations of this interface?

No not really. Was just a matter of readability,  encapsulating the interface in a way that is easy to access and update if needed. Is this an issue

ldts: > By the way, given that here's zero documentation of the different functions perhaps you could give a reference to the used interface at least?

The information contained in the file is of a similar nature of what you will find on the PLM firmware - I simply removed the redundancy

Take this as an example:
```

/*****************************************************************************/
/**
 * brief	This function sends IPI request to read MiscCtrlBits
 * 		requested by the user
 *
 * param	InstancePtr Pointer to the client instance
 * param	MiscCtrlBits	Address of the output buffer to store the
 * 				MiscCtrlBits eFuses data
 *
 * return	- XST_SUCCESS - If the read is successful
 * 		- XST_FAILURE - If there is a failure
 *
 ******************************************************************************/
int XNvm_EfuseReadMiscCtrlBits(XNvm_ClientInstance *InstancePtr, const u64 MiscCtrlBits)
```

Do you want me to populate OP-TEE with this sort of comments? (not trying to be ironic here...I honestly do not know what you need)

jenswi-linaro: > Do you want me to populate OP-TEE with this sort of comments?

No, that doesn't help much. But given that this is an ABI it should be nice to have some kind of documentation to fall back to. Surely there's something you can link to.

jenswi-linaro: > > How come you have chosen a function pointer based interface? Do you expect other implementations of this interface?
> 
> No not really. Was just a matter of readability, encapsulating the interface in a way that is easy to access and update if needed. Is this an issue? I can send the HUK driver which uses the interface if it helps with the usability case.

I'm not so keen on function pointer based interfaces when they aren't needed. One of the drawbacks is that linker optimization like garbage collection is prevented with this. Another is that it always raises the question of where are the other implementations, so it's a bit confusing.
We used to have various crypto ops structs but I replaced those in:
commit 3bbd3ce9058c ("core: remove struct crypto_ops")
commit 291e5450a1e9 ("Replace struct acipher_ops with function interface")
commit 33790cc18342 ("Replace struct bignum_ops with function interface")
commit 8875ce46f413 ("Replace struct authenc_ops with function interface")
commit e9eaba5c0a3d ("Replace struct mac_ops with function interface")
commit 82d91db11f01 ("Replace struct cipher_ops with function interface")
commit b887bd8f271b ("Replace struct hash_ops with function interface")

ldts: oh man, I wish I had known about this restriction. I guess you are asking me to get rid of the pointer based interface then?

> > Do you want me to populate OP-TEE with this sort of comments?
> 
> No, that doesn't help much. But given that this is an ABI it should be nice to have some kind of documentation to fall back to. Surely there's something you can link to.


@nathan-menhorn is there any publicly available document that describes each one of the eFuses?

ldts: > > > How come you have chosen a function pointer based interface? Do you expect other implementations of this interface?
> > 
> > 
> > No not really. Was just a matter of readability, encapsulating the interface in a way that is easy to access and update if needed. Is this an issue? I can send the HUK driver which uses the interface if it helps with the usability case.
> 
> I'm not so keen on function pointer based interfaces when they aren't needed. One of the drawbacks is that linker optimization like garbage collection is prevented with this. Another is that it always raises the question of where are the other implementations, so it's a bit confusing. We used to have various crypto ops structs but I replaced those in: commit [3bbd3ce](https://github.com/OP-TEE/optee_os/commit/3bbd3ce9058cb0f6c4e7d4725823a3e70029d049) ("core: remove struct crypto_ops") commit [291e545](https://github.com/OP-TEE/optee_os/commit/291e5450a1e9cf4e495d292f40a8ab9b901d9c78) ("Replace struct acipher_ops with function interface") commit [33790cc](https://github.com/OP-TEE/optee_os/commit/33790cc18342231b13f7e6791196bd7c1f41e33f) ("Replace struct bignum_ops with function interface") commit [8875ce4](https://github.com/OP-TEE/optee_os/commit/8875ce46f413b5f5b4e6d73707672d8a3f51f92c) ("Replace struct authenc_ops with function interface") commit [e9eaba5](https://github.com/OP-TEE/optee_os/commit/e9eaba5c0a3d27075b1e995294fd42de65c0b912) ("Replace struct mac_ops with function interface") commit [82d91db](https://github.com/OP-TEE/optee_os/commit/82d91db11f0188df939ce52f6106a39f026de74c) ("Replace struct cipher_ops with function interface") commit [b887bd8](https://github.com/OP-TEE/optee_os/commit/b887bd8f271b056210e964fe7e86dbafe7f6d752) ("Replace struct hash_ops with function interface")

I hear you. I am just the opposite...having an interface to retrieve when needed instead of a direct function call makes the software partitioning cleaner to my eyes. 

Having said I started doing the code with the usual funcion calls but the names of the functions where too long so I switched to that pointer interface...ie: `versal_efuse_read_offchip_revoke_id` or `versal_efuse_write_puf_as_user_fuse`. I think the code is going to be quite ugly hence why I need to make sure this is what you are asking me to do. if you could confirm I'll just rename it all.

ldts: um I think linker garbage collection can not be used as a silver bullet rule to every interface; in fact I don't think it is that relevant on Versal due to this architecture not being so memory constrained as other platforms (i.e: stm32mp1, imx6ull..). 
Is there any other measurable drawback apart from this one? 

I would like to align with the overall design that you have in mind for OP-TEE as much as possible but at the same time this driver is not part of the OP-TEE generic core. And as I mentioned - and as driver maintainer - in this instance I might value more the clarity in the interface than the few hundreds of bytes that we might save using direct calls.

I suppose if I don't get review comments - and if you dont want to proceed with the review - I will just have to rename the functions and move on. Still have a lot of code pending for Versal in my trees (HUK, ECC, RSA, SHA3...)

ldts:
@etienne-lms the reference is contained in the original series  https://github.com/OP-TEE/optee_os/pull/5426 
In particular https://github.com/OP-TEE/optee_os/pull/5426/commits/2bc7d7135f7f1af0b2dd618879907568b88756c8, which I haven't posted separately because it depends on other things and I didnt want to overcomplicate the review again.

The issue with this interface is that there are hundreds of fields only documented under NDA and so there is little value in attempting to describe the interface - it would be worthless info.

For instance we  wont be able to go into detail with this definition

```
sruct versal_efuse_sec_ctrl_bits {
	uint8_t aes_dis;
	uint8_t jtag_err_out_dis;
	uint8_t jtag_dis;
	uint8_t ppk0_wr_lk;
	uint8_t ppk1_wr_lk;
	uint8_t ppk2_wr_lk;
	uint8_t aes_crc_lk;
	uint8_t aes_wr_lk;
	uint8_t user_key0_crc_lk;
	uint8_t user_key0_wr_lk;
	uint8_t user_key1_crc_lk;
	uint8_t user_key1_wr_lk;
	uint8_t sec_dbg_dis;
	uint8_t sec_lock_dbg_dis;
	uint8_t boot_env_wr_lk;
	uint8_t reg_init_dis;
	uint8_t pad[48];
};

```

So what will be the point of documenting the write operation of these bits? what will I document other than `this function below writes versal_efuse_sec_ctrl_bits`?

```
struct versal_efuse_write_ops {
[...]
TEE_Result (*sec_ctrl)(struct versal_efuse_sec_ctrl_bits *p)`
[...]
}
```

I suppose I can add a comment saying details can be obtained via NDA with AMD/Xilinx?

But I see no value in incorporating the comments as described here for instance:
https://github.com/Xilinx/embeddedsw/blob/master/lib/sw_services/xilnvm/src/client/xnvm_efuseclient.c

Yes each versal platform will only build the versal-nvm driver but I dont know if all the eFuses will be available on each platform variant. Having a pointer based interface allows for finer tuning and easier extensibility. But I cant think ahead how/if this will happen.
Pretty much the same than when a **struct keys** is defined with just one key inside just because in the future it might hold more keys: that struct incorporates a nicer view of the software through the abstraction but is functionally unnecessary.

ldts:
@jenswi-linaro @etienne-lms I think my disagreement with your POVs is irrelevant at this point.

what do you need me to do? just remove pointers and write those long function names? anything else at this point?

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+	size_t i = 0;
+
+	cmd->data[i++] = NVM_API_ID(efuse);
+	if (arg)
+		cmd->data[i++] = *arg;
+
+	if (!ibufs[0].buf)
+		return TEE_SUCCESS;
+
+	cmd->data[i++] = virt_to_phys(ibufs[0].buf);
```
jenswi-linaro, line 117: Use `reg_pair_from_64()` instead.

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+
+	if (first + num > EFUSE_MAX_USER_FUSES || len < num * sizeof(uint32_t))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	req.ibuf[0].buf = &cfg;
+	req.ibuf[0].len = sizeof(cfg);
+	req.ibuf[1].buf = lbuf;
+	req.ibuf[1].len = sizeof(lbuf);
+
+	cfg.addr = (paddr_t)virt_to_phys((void *)cfg.addr);
```
jenswi-linaro, line 257: Cast to `paddr_t` isn't needed. `virt_to_phys()` already returns a `paddr_t`, please fix the other places in this file too.

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+
+	memcpy(buf, lbuf, num * sizeof(uint32_t));
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result versal_efuse_read_iv(uint32_t *buf, size_t len,
+				enum versal_nvm_iv_type type)
+{
+	uint8_t lbuf[1024] __aligned_efuse = { 0 };
```
jenswi-linaro, line 270: That's quite a bit to store on the stack. Can it be done differently with a smaller buffer or by having the caller supply a well-aligned buffer? Another option is to allocate using the mempool API. The mempool API doesn't have a `mempool_memalign()`, but that shouldn't be hard to add.

There are quite a few other functions also doing large stack allocations. I'm not sure how big this problem is. If these are only called during boot initialization I guess we're safe, but if it can happen later during syscall processing then it's more concerning.

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+		.data.env_mon_dis_flag = 1,
+		.efuse_id = EFUSE_WRITE,
+	};
+	uint32_t row = 0;
+	uint32_t bit = 0;
+
+	row = id >> (NVM_WORD_LEN + 1);
+	bit = id & (NVM_WORD_LEN - 1);
+
+	cfg.revoke_id[row] = 1 << bit;
```
jenswi-linaro, line 582: Use the `BIT()` macro instead.

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+
+	return TEE_SUCCESS;
+}
+
+/*
+ *  This functionality requires building the PLM with XNVM_ACCESS_PUF_USER_DATA
+ *  Calls will fail otherwise.
+ *  When available, efuse_read_puf becomes unavailable.
+ */
+TEE_Result versal_efuse_read_puf_as_user_fuse(struct versal_efuse_puf_user_fuse
```
jenswi-linaro, line 776: Please break the line as:
``` C
TEE_Result 
versal_efuse_read_puf_as_user_fuse(struct versal_efuse_puf_user_fuse *p)
```

Same for all the other functions where you're breaking the line between the type and parameter name.

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+	return TEE_SUCCESS;
+}
+
+TEE_Result versal_efuse_read_offchip_revoke_id(uint32_t *buf, size_t len,
+					       enum versal_nvm_offchip_id id)
+{
+	uint8_t lbuf[1024]__aligned_efuse = { 0 };
+	struct versal_nvm_read_req req = {
+		.efuse_id = EFUSE_READ_OFFCHIP_REVOCATION_ID,
+		.offchip_id = id, };
```
etienne-lms, line 695: mising new line before `}`

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+/*
+ *  This functionality requires building the PLM with XNVM_ACCESS_PUF_USER_DATA
+ *  Calls will fail otherwise.
+ *  When available, efuse_read_puf becomes unavailable.
+ */
+TEE_Result versal_efuse_read_puf_as_user_fuse(struct versal_efuse_puf_user_fuse
+					      *p)
+{
+	uint32_t fuses[PUF_EFUSES_WORDS]__aligned_efuse = { 0 };
+	struct versal_efuse_puf_fuse_addr {
```
etienne-lms, line 780: this is use at 2 places. Better define at source file top?

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,944 @@
...
+}
+
+TEE_Result versal_efuse_read_user_data(uint32_t *buf, size_t len,
+				       uint32_t first, size_t num)
+{
+	uint8_t lbuf[1024] __aligned_efuse = { 0 };
+	struct versal_efuse_user_data cfg __aligned_efuse = {
+		.addr = (uintptr_t)lbuf,
+		.start = first,
+		.num = num, /* fuses needs to be at least 40 bytes */
```
etienne-lms, line 243: return with an error if less than 40 bytes?

core/drivers/versal_nvm.c
etienne-lms: I have the same comment on stack size. Stacks are few kByte by default. 1kB is quite a lot.

core/drivers/versal_nvm.c
ldts: I'll just remove the comment. the firmware handle function might change and this requirement is not in the API...was just sort of soemething I noticed

ldts:
updated

ldts:
rebased

core/drivers/versal_nvm.c
```diff
@@ -0,0 +1,1001 @@
...
+	size_t i = 0;
+
+	cmd->data[i++] = NVM_API_ID(efuse);
+	if (arg)
+		cmd->data[i++] = *arg;
+
+	if (!ibufs[0].buf)
+		return TEE_SUCCESS;
+
+	reg_pair_from_64(virt_to_phys(ibufs[0].buf, &b, &a);
```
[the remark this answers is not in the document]
ldts: sorry yes, last minute changes done after the code was tested. fixing that and adding it to CI.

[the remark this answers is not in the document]
ldts: worst than that, pushed the wrong branch

jenswi-linaro:

jforissier:
Ping @etienne-lms in case you would like to Ack. Thanks.

etienne-lms (commented):

ldts:
sorry didnt notice. updating now

ldts:
@jforissier updated

jforissier:
@ldts can you please resolve the conflict? Then I will merge this. Thanks!

PR #5548, opened 2022-09-23
core: add pointer authentication support

jenswi-linaro:
Previously pointer authentication was only supported for TAs. With this patch add a configuration option CFG_CORE_PAUTH to enable support for core. Each privileged thread has its own APIA key. There are also a separate APIA key for each physical core used when handling an abort or when using the tmp stack.

### DIFF
--- a/core/arch/arm/include/kernel/thread_arch.h
+++ b/core/arch/arm/include/kernel/thread_arch.h
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
  * Copyright (c) 2014, STMicroelectronics International N.V.
- * Copyright (c) 2016-2017, Linaro Limited
+ * Copyright (c) 2016-2022, Linaro Limited
  * Copyright (c) 2020-2021, Arm Limited
  */
 
@@ -29,6 +29,16 @@
 
 struct mobj;
 
+/*
+ * Storage of keys used for pointer authentication. FEAT_PAuth supports a
+ * number of keys of which only the APIA key is currently used, depending on
+ * configuration.
+ */
+struct thread_pauth_keys {
+	uint64_t apia_hi;
+	uint64_t apia_lo;
+};
+
 struct thread_core_local {
 #ifdef ARM32
 	uint32_t r[2];
@@ -36,6 +46,9 @@ struct thread_core_local {
 #endif
 #ifdef ARM64
 	uint64_t x[4];
+#endif
+#ifdef CFG_CORE_PAUTH
+	struct thread_pauth_keys keys;
 #endif
 	vaddr_t tmp_stack_va_end;
 	long kcode_offset;
@@ -71,11 +84,6 @@ struct thread_user_vfp_state {
 	struct vfp_state vfp;
 	bool lazy_saved;
 	bool saved;
-};
-
-struct thread_pauth_keys {
-	uint64_t hi;
-	uint64_t lo;
 };
 
 #ifdef ARM32
@@ -161,6 +169,10 @@ struct thread_abort_regs {
 	uint64_t elr;
 	uint64_t spsr;
 	uint64_t sp_el0;
+#if defined(CFG_TA_PAUTH) || defined(CFG_CORE_PAUTH)
+	uint64_t apiakey_hi;
+	uint64_t apiakey_lo;
+#endif
 };
 #endif /*ARM64*/
 
@@ -216,7 +228,7 @@ struct thread_svc_regs {
 	uint64_t x28;
 	uint64_t x29;
 #endif
-#ifdef CFG_TA_PAUTH
+#if defined(CFG_TA_PAUTH) || defined(CFG_CORE_PAUTH)
 	uint64_t apiakey_hi;
 	uint64_t apiakey_lo;
 #endif
@@ -256,7 +268,7 @@ struct thread_ctx_regs {
 	uint64_t cpsr;
 	uint64_t x[31];
 	uint64_t tpidr_el0;
-#ifdef CFG_TA_PAUTH
+#if defined(CFG_TA_PAUTH) || defined(CFG_CORE_PAUTH)
 	uint64_t apiakey_hi;
 	uint64_t apiakey_lo;
 #endif
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1252,6 +1252,14 @@ void init_tee_runtime(void)
 	if (!IS_ENABLED(CFG_VIRTUALIZATION))
 		call_preinitcalls();
 	call_initcalls();
+
+	/*
+	 * These two functions uses crypto_rng_read() to initialize the
+	 * pauth keys. Once call_initcalls() returns we're guaranteed that
+	 * crypto_rng_read() is ready to be used.
+	 */
+	thread_init_core_local_pauth_keys();
+	thread_init_thread_pauth_keys();
 }
 
 static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -858,15 +858,27 @@ CFG_CORE_ASYNC_NOTIF ?= n
 $(eval $(call cfg-enable-all-depends,CFG_MEMPOOL_REPORT_LAST_OFFSET, \
 	 CFG_WITH_STATS))
 
-# Pointer Authentication (part of ARMv8.3 Extensions) provides
-# instructions for signing and authenticating pointers against secret keys.
-# These can be used to mitigate ROP (Return oriented programming) attacks.
-# This option enables these instructions for TA's at EL0. When this option is
-# enabled , TEE core will initialize secret keys per TA.
-CFG_TA_PAUTH ?= n
+# Pointer Authentication (part of ARMv8.3 Extensions) provides instructions
+# for signing and authenticating pointers against secret keys. These can
+# be used to mitigate ROP (Return oriented programming) attacks. This is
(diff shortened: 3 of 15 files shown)


core/arch/arm/include/kernel/thread_arch.h
```diff
@@ -29,13 +29,21 @@
 
 struct mobj;
 
+struct thread_pauth_keys {
```
ldts, line 32: why is it plural? this is just a key

core/arch/arm/kernel/boot.c
```diff
@@ -1252,6 +1252,10 @@ void init_tee_runtime(void)
 	if (!IS_ENABLED(CFG_VIRTUALIZATION))
 		call_preinitcalls();
 	call_initcalls();
+
+	/* crypto_rng_read() should be ready to be used by now */
```
ldts, line 1256: not sure this comment adds a lot of value.

core/arch/arm/kernel/boot.c
```diff
@@ -1252,6 +1252,10 @@ void init_tee_runtime(void)
...
+	thread_init_core_local_pauth_keys();
```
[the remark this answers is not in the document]
etienne-lms: the functions are stubbed when config switches are disabled.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`. LGTM AFAICT.

ldts:
> `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`. LGTM AFAICT.

um, why? the structure name is clearly wrong.

core/arch/arm/kernel/boot.c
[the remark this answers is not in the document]
ldts: sure but just paraphrassing @jenswi-linaro review to one of my PR
"I prefer a configuration where the overhead can be reduced to zero when not needed. With this weak function always being called, we're still adding a few useless instructions for every platform. If this call was guarded with some IS_ENABLED() test, like before, it could be optimized away completely if not needed."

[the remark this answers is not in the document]
ldts: more over since the PR is disabling the feature by default..
https://github.com/OP-TEE/optee_os/pull/5548/commits/1fd4e7de4c3858a41ee3afbcc2da04f68be39a36#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54aR871

core/arch/arm/include/kernel/thread_arch.h
jenswi-linaro: I'm just moving the definition so I can use it a bit earlier.
I suppose we could rename it if you think it's important, it should probably go into a separate patch though.

core/arch/arm/kernel/boot.c
jenswi-linaro: It answers the question of why this function is called so late, one would normally expect this to be called earlier.

core/arch/arm/kernel/boot.c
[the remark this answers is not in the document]
jenswi-linaro: > sure but just paraphrassing @jenswi-linaro review to one of my PR "I prefer a configuration where the overhead can be reduced to zero when not needed. With this weak function always being called, we're still adding a few useless instructions for every platform. If this call was guarded with some IS_ENABLED() test, like before, it could be optimized away completely if not needed."
> 
> why should this be any different?

It's different because it's done differently. I'm stubbing using a static inline function, that has zero overhead with optimizations enabled.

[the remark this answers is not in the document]
jenswi-linaro: > should it be surrounded with an `if (IS_ENABLED(CFG_CORE_PAUTH)))` to make it easier to read? I realize it is functionally equivalent...

That can be discussed, I chose to just call the function in this case. I thought it would be pretty clear what's going on in this case.

core/arch/arm/include/kernel/thread_arch.h
jenswi-linaro: On the other hand, this is supposed to abstract the PAUTH key handling and PAUTH has multiple keys.

ldts: I think naming single key in its plural form might be confusing when reading the code; and when it comes to nitpicks I'd say it is an important one easy to address. For me, personally, functionality comes first but in my experience with OP-TEE these details have always mattered hence my point. 
I suppose it makes sense renaming it in a different PR too.

core/arch/arm/kernel/boot.c
ldts: Ah now it makes sense. Can we mention this explicitly perhaps? I don't think it is obvious

jenswi-linaro:
Adding a fix for the native interrupt handler.

core/arch/arm/include/kernel/thread_arch.h
jenswi-linaro: So you mean depending on how many keys we abstract away we need to change the name of the abstraction?

I can add a comment explaining what's the deal with the pauth keys, I hope that should be more helpful.

jenswi-linaro:
Update

core/arch/arm/include/kernel/thread_arch.h
ldts: Sure the multiplicity in this case should be give by the name of the variable not the structure definition.
multiple keys would be  `struct thread_pauth_key *keys` or a pointer to the single key would be `struct thread_pauth_key *key`

Otherwise why not have `struct thread_vector_tables,` `struct thread_user_vfp_states`  and so forth.

ldts: In fact
```
	regs->apiakey_hi = keys->hi;
	regs->apiakey_lo = keys->lo;
```

regs surely points to a structure containing a set of registers, where as keys points to a structure containing a single key

so in fact

```
struct user_mode_ctx {
	struct vm_info vm_info;
	struct vm_paged_region_head *regions;
#if defined(CFG_WITH_VFP)
	struct thread_user_vfp_state vfp;
#endif
#if defined(CFG_TA_PAUTH)
	struct thread_pauth_keys keys;
#endif
	struct ts_ctx *ts_ctx;
	uaddr_t entry_func;
	uaddr_t dump_entry_func;
#ifdef CFG_FTRACE_SUPPORT
	uaddr_t ftrace_entry_func;
#endif
	uaddr_t dl_entry_func;
	uaddr_t ldelf_stack_ptr;
	bool is_32bit;
	bool is_initializing;
	vaddr_t stack_ptr;
};
```

should also have ` struct thread_pauth_key key;` instead

ldts: Unless you expect to add more keys to the structure sometime in the future of course
```
struct thread_pauth_keys {
	uint64_t akey_hi;
	uint64_t akey_lo;
	uint64_t bkey_hi;
	uint64_t bkey_lo;
       [....]
};
```

ldts: quickly coming back to this just to wrap it up from my end, I think you should at least change to a way that is more specific, maybe also add a comment


```
/* OP-TEE only uses PAUTH_AKEY */
struct thread_pauth_keys {
	uint64_t akey_hi;
	uint64_t akey_lo;
};
```

jenswi-linaro: Fair enough, I'll update.

jenswi-linaro:
Update

mk/config.mk
```diff
@@ -858,15 +858,25 @@ CFG_CORE_ASYNC_NOTIF ?= n
...
 
-# Pointer Authentication (part of ARMv8.3 Extensions) provides
-# instructions for signing and authenticating pointers against secret keys.
-# These can be used to mitigate ROP (Return oriented programming) attacks.
-# This option enables these instructions for TA's at EL0. When this option is
-# enabled , TEE core will initialize secret keys per TA.
-CFG_TA_PAUTH ?= n
+# Pointer Authentication (part of ARMv8.3 Extensions) provides instructions
+# for signing and authenticating pointers against secret keys. These can
+# be used to mitigate ROP (Return oriented programming) attacks.
```
jforissier, line 863: Practically speaking, and typically, this means having `paciasp`/`autiasp` at the beginning and end of functions, inserted by the compiler, in order to sign then verify ELR, doesn't it?
Assuming this is all there is to it at the moment, it would be nice to write it explicitly here I think.

jforissier (approved):
Nice complement to `CFG_TA_PAUTH` :) One suggestion below.
Also, what is the reason for choosing one key per thread and CPU core, and one key per TA? My understanding is the more keys the better and it was just convenient to do it like so, correct?

jenswi-linaro:
> Also, what is the reason for choosing one key per thread and CPU core, and one key per TA? My understanding is the more keys the better and it was just convenient to do it like so, correct?

One key per TA was from before and I assume the reason was isolation to avoid leakage of shared keys.
For the Core part I'm practically using one key for each stack also for isolation and avoiding leakage of keys. We should ideally regenerate the keys from time to time since we are encrypting with the same key and IV (stack pointer) more than once, but that can be saved for future improvements.

mk/config.mk
jenswi-linaro: OK, I'll add something.

jenswi-linaro:
Squashed and tags applied. @ldts, do you want to give a tag?

ldts:
> Squashed and tags applied. @ldts, do you want to give a tag?

Sure, thanks

`Reviewed-by: Jorge Ramirez-Ortiz <jorge@foundries.io>
`

jenswi-linaro:
Tag applied

Issue #5448, opened 2022-07-22
STM32MP1 Hardware Unique Key


ldts:
Has anyone given any thought to generating a HUK for the STM32MP1 platform? Maybe @etienne-lms ?

Intuitively perhaps using the [Secure Secret Provisioning process ](https://www.st.com/resource/en/application_note/dm00713831-.pdf) to write a 256 bit secret key to a pre-defined OTP, deny it access from NSEC and then using it to encrypt the SoC unique identifier should be enough. 

Just as it is done on the Xilinx/AMD platforms.

Any other thoughts for a HUK driver? Shall I propose something along those lines?

etienne-lms:
Hi @ldts,
Indeed using BSEC OTP fuses to store secret HUK is the recommended way. The reference you provide is the good one for burning the target fuses. Yet there is no predefined default location for the OTP words that shall store the HUK, for stm32mp1 platforms. Defining fuses would be nice but not all people really want to burn fuses. There are free locations in the secure part of the BSEC OTP fuse memory.

That said, I've been asked at few times some example code. I've prepared this [patch](https://github.com/etienne-lms/optee_os/commit/cfec862b017d0a97e184f32abfdd75c3d0000eef) (non-permanent link) to define 4 OTPs words for that purpose, expecting `CFG_xxx` switches to define the location, yet allowing to use an all 0's test key. IHIH. Maybe it would be wroth merging this code in mainline.

By the way, OP-TEE OS currently expects a 128bit HUK. The referred patch expects 4 32bit OTPs words to store HUK.

ldts:
thanks @etienne-lms; I was thinking we could perhaps use the SoC unique identifier - 96 bits - to make it unique and then just have the OEMs  write an AES encryption key on the OTP eFuse to generate the HUK - rather than having the actual HUK in OTP? So maybe extending your patch we could retrieve a 256 encryption key?

then maybe the OP-TEE core could have a generic driver that generates the HUK as
1) gets some public uniqueness for the SoC
2) encrypts it with a secret key (from either eFuse or already enabled in the crypto engine)

what do you think?

etienne-lms:
Do you propose an alternate implementation of `huk_subkey_derive()` to do HUK derivation with different inputs (a HUK read from fuses that is not 128bit large, a platform specific unique ID, ...)? 
[...]

PR #5555, opened 2022-09-27
plat-stm32mp15: generate HUK using an AES-GCM key from BSEC OTP

ldts:
The algorithm used simplifies the device provisioning phase because
it does not require a unique per device secret to be fused: just a key
common to all devices.

The algorithm uses a 128 bit symmetric key stored as four 32 bit words
read from OTP fuses.
The HUK is calculated by AES-GCM encrypting the device UID (96 bits).
Since the UID is persistent - and so should be the key - the NONCE can
be reused and hold any value.
The OTP values must be secrets but don't need to be unique per-device.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -261,6 +261,9 @@ CFG_WITH_NSEC_UARTS ?= y
 # UART instance used for early console (0 disables early console)
 CFG_STM32_EARLY_CONSOLE_UART ?= 4
 
+# Disable the HUK by default as it requires a product specific configuration
+CFG_STM32MP15_HUK ?= n
+
 # Sanity on choice config switches
 ifeq ($(call cfg-all-enabled,CFG_STM32MP15 CFG_STM32MP13),y)
 $(error CFG_STM32MP13_CLK and CFG_STM32MP15_CLK are exclusive)
--- a/core/drivers/stm32mp15_huk.c
+++ b/core/drivers/stm32mp15_huk.c
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2022, Linaro Limited
+ * Copyright (c) 2022, Foundries.io Limited
+ */
+
+#include <assert.h>
+#include <config.h>
+#include <crypto/crypto.h>
+#include <drivers/stm32_bsec.h>
+#include <kernel/tee_common_otp.h>
+#include <mempool.h>
+#include <platform_config.h>
+#include <stm32_util.h>
+#include <string.h>
+#include <string_ext.h>
+
+static bool stm32mp15_huk_init;
+
+static TEE_Result stm32mp15_read_uid(uint32_t *uid)
+{
+	uint32_t *q = uid;
+
+	/*
+	 * Shadow memory for UID words might not be locked: to guarante that
+	 * the final values are read we must lock them.
+	 */
+	if (stm32_bsec_set_sw_lock(UID0_OTP) ||
+	    stm32_bsec_shadow_read_otp(q++, UID0_OTP))
+		return TEE_ERROR_GENERIC;
+
+	if (stm32_bsec_set_sw_lock(UID1_OTP) ||
+	    stm32_bsec_shadow_read_otp(q++, UID1_OTP))
+		return TEE_ERROR_GENERIC;
+
+	if (stm32_bsec_set_sw_lock(UID2_OTP) ||
+	    stm32_bsec_shadow_read_otp(q++, UID2_OTP))
+		return TEE_ERROR_GENERIC;
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result stm32mp15_read_otp(uint32_t otp, uint32_t *key, bool *locked)
+{
+	bool tmp = true;
+
+	if (!stm32mp_is_closed_device()) {
+		/*
+		 * When the device is not closed, the shadow memory for these
+		 * words might not be locked: check and report them
+		 */
+		if (stm32_bsec_read_permanent_lock(otp, &tmp))
+			return TEE_ERROR_GENERIC;
+
+		if (tmp && stm32_bsec_read_sw_lock(otp, &tmp))
+			return TEE_ERROR_GENERIC;
+	}
+
+	if (stm32_bsec_shadow_read_otp(key, otp))
+		return TEE_ERROR_GENERIC;
+
+	*locked = *locked && tmp;
+
+	return TEE_SUCCESS;
+}
+
+/*
+ *  AES-GCM: nonce must be unique per message and key.
+ *
+ *  This function always uses the same key - once its locked - with the same
+ *  unique message hence the nonce can be any constant.
+ */
+static TEE_Result aes_gcm_encrypt_uid(uint8_t *key, size_t key_len,
+				      uint8_t *out, size_t *out_len)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	const uint8_t nonce[12] = { 0x55 };
+	uint32_t uid[4] = { 0 };
+	uint8_t tag[16] = { 0 };
+	size_t nonce_len = sizeof(nonce);
+	size_t tag_len = sizeof(tag);
+	size_t uid_len = sizeof(uid);
+	void *ctx = NULL;
+
+	ret = stm32mp15_read_uid(uid);
+	if (ret)
+		goto out;
+
+	ret = crypto_authenc_alloc_ctx(&ctx, TEE_ALG_AES_GCM);
+	if (ret)
+		goto out;
+
+	ret = crypto_authenc_init(ctx, TEE_MODE_ENCRYPT, key, key_len, nonce,
+				  nonce_len, TEE_AES_BLOCK_SIZE, 0, uid_len);
+	if (ret)
+		goto out_free_ctx;
+
+	ret = crypto_authenc_enc_final(ctx, (uint8_t *)uid, sizeof(uid),
+				       out, out_len, tag, &tag_len);
+	if (ret)
+		goto out_free_ctx;
+
+	crypto_authenc_final(ctx);
+out_free_ctx:
+	crypto_authenc_free_ctx(ctx);
+out:
+	if (ret)
+		memzero_explicit(out, *out_len);
+
+	return ret;
+}
+
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	uint32_t otp_key[4] = { 0 };
+	size_t len = sizeof(otp_key);
+	TEE_Result ret = TEE_SUCCESS;
+	uint32_t *key = otp_key;
+	bool lock = true;
+
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_0 < STM32MP1_OTP_MAX_ID);
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_1 < STM32MP1_OTP_MAX_ID);
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_2 < STM32MP1_OTP_MAX_ID);
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_3 < STM32MP1_OTP_MAX_ID);
+
+	ret = stm32mp15_read_otp(CFG_STM32MP15_HUK_BSEC_KEY_0, key++, &lock);
+	if (ret)
+		goto out;
+
+	ret = stm32mp15_read_otp(CFG_STM32MP15_HUK_BSEC_KEY_1, key++, &lock);
+	if (ret)
+		goto out;
+
+	ret = stm32mp15_read_otp(CFG_STM32MP15_HUK_BSEC_KEY_2, key++, &lock);
+	if (ret)
+		goto out;
+
+	ret = stm32mp15_read_otp(CFG_STM32MP15_HUK_BSEC_KEY_3, key++, &lock);
+	if (ret)
+		goto out;
+
+	ret = aes_gcm_encrypt_uid((uint8_t *)otp_key, len, hwkey->data, &len);
+
+	if (len != HW_UNIQUE_KEY_LENGTH)
+		ret = TEE_ERROR_GENERIC;
+out:
+	memzero_explicit(otp_key, HW_UNIQUE_KEY_LENGTH);
+
+	if (!ret && !stm32mp15_huk_init) {
+		stm32mp15_huk_init = true;
+		IMSG("HUK %slocked", lock ? "" : "un");
+		DHEXDUMP(hwkey->data, HW_UNIQUE_KEY_LENGTH);
+	}
+
+	return ret;
+}
+
--- a/core/drivers/sub.mk
+++ b/core/drivers/sub.mk
@@ -60,6 +60,7 @@ srcs-$(CFG_ARM_SMCCC_TRNG) += smccc_trng.c
 srcs-$(CFG_VERSAL_GPIO) += versal_gpio.c
 srcs-$(CFG_VERSAL_MBOX) += versal_mbox.c
 srcs-$(CFG_VERSAL_PM) += versal_pm.c
+srcs-$(CFG_STM32MP15_HUK) += stm32mp15_huk.c
 
 subdirs-y += crypto
 subdirs-$(CFG_BNXT_FW) += bnxt


ldts:
@etienne-lms this is what I meant in https://github.com/OP-TEE/optee_os/issues/5448

ldts:
tested on stm32mp157

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,100 @@
...
+				       tag, &tag_len);
+	if (res)
+		return res;
+
+	crypto_authenc_final(ctx);
+	crypto_authenc_free_ctx(ctx);
+
+	huk.ready = true;
+
+	if (!stm32mp_is_closed_device())
```
etienne-lms, line 91: To ensure key don't change, we should test the permanent-lock of each read OTP:
```patch
+	huk_fuses_locked = false;
 	for (n = 0; n < sizeof(key) / sizeof(uint32_t); n++) {
+		boot lock = false;
+
+		res = stm32_bsec_read_permanent_lock(bsec_base + n, &lock);
+		if (res)
+			return res;
+		if (!lock)
+			huk_fuses_locked = false;	
+
 		res = stm32_bsec_shadow_register(bsec_base + n);
 		if (res)
 			return res;
```

The test (at line 91) should also depend on CFG_STM32MP1_HUK_BSEC_SHADOW_TESTKEY.

etienne-lms (commented):
Hi Jorge, Thanks for the proposal. Please find some comments.
We may add an alternate way in the future: defining the fuse location with the DT. Would that sound good to you?

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		res = stm32_bsec_read_otp(&word, bsec_base + n);
+		if (res)
+			return res;
+
+		memcpy(key + n * sizeof(uint32_t), &word, sizeof(uint32_t));
+	}
+
+	for (n = 0; n < (UID2_OTP - UID0_OTP + 1); n++) {
```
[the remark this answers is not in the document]
ldts: I just shrinked the word nonce which could be any value and doesn't need to be secret. Maybe I should leave it as nonce?

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,100 @@
...
+#include <string.h>
+
+#if !defined(CFG_STM32MP1_HUK_BSEC_BASE)
+#error "CFG_STM32MP1_HUK_BSEC_BASE undefined"
+#endif
+
+static struct {
+	uint8_t key[HW_UNIQUE_KEY_LENGTH];
+	bool ready;
+} huk;
```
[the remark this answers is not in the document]
ldts: um all other platforms tend to do this. I always worked under the impression that the memory would be secured. Is this not the case?

ldts:
@etienne-lms I made some adjustments - let me know what you think (mainly about zeroizing the HUK).
With the original code, if the OTP were locked, writing a zero to them might sometime silently fail and we end up reading back whatever value was locked. So I opted for forcing the zeroes in a simpler way

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,124 @@
...
+			return TEE_ERROR_GENERIC;
+		}
+	} else {
+		memcpy(hwkey->data, otp_data, HW_UNIQUE_KEY_LENGTH);
+	}
+
+	if (!initialized) {
+		initialized = true;
+		IMSG("HUK [%s %s]", type, state);
+		DHEXDUMP(hwkey->data, HW_UNIQUE_KEY_LENGTH);
```
etienne-lms, line 120: Prefer without this DHEXDUMP() trace.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,121 @@
...
+	for (n = 0; n < (UID2_OTP - UID0_OTP + 1); n++) {
+		res = stm32_bsec_read_otp(&word, UID0_OTP + n);
+		if (res)
+			return res;
+
+		memcpy(uid + n * sizeof(uint32_t), &word, sizeof(uint32_t));
+	}
+
+	/* HUK is the AES-GCM encrypted UID using a 128 bit key from OTP */
+	res = crypto_authenc_alloc_ctx(&ctx, TEE_ALG_AES_CCM);
```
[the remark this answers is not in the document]
ldts: typo! GCM

core/arch/arm/plat-stm32mp1/plat_huk.c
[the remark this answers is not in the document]
ldts: Right, AMD/Xilinx proposed this scheme to protect public uniqueness (ie SoC ID) via a secret key- it is what is being used in Zynqmp and in Versal ACAP so I thikn it makes sense to use it globally.

It makes provisioning after manufacturing easier since the secret key can be shared across all devices.

if we used SHA and the SoC Id became known then we would have an issue (I thikn this can be read from U-boot?) ; but then again derivation is disabled by default so for the simple scheme the user can read the uniqueness from those OTPs instead

core/arch/arm/plat-stm32mp1/plat_huk.c
ldts: but I think this is useful for development( ie if enabling RPMB might be a good idea to remember the HUK)

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -254,6 +254,20 @@ CFG_WITH_NSEC_UARTS ?= y
...
+CFG_STM32MP1_HUK_BSEC ?= n
+ifeq ($(CFG_STM32MP1_HUK_BSEC),y)
+# Zeroize the HUK: if enabled, the HUK will be set to Zeroes.
+CFG_STM32MP1_HUK_ZEROIZED ?= n
+# Otherwise: use four words starting at CFG_STM32MP1_HUK_BSEC_BASE as follows
+#
+# if DERIVED: the AES-GCM encryption key is retrieved from OTP
+#             and used to encrypt the UID OTP to generate the HUK.
+# else: the HUK is read from OTP (default case).
+CFG_STM32MP1_HUK_BSEC_DERIVED ?= n
```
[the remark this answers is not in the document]
ldts: what about just GCM? it signifies that we are using some private key to encrypt possibly public unique information

core/arch/arm/plat-stm32mp1/plat_huk.c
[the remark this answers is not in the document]
ldts: sure, I suppose reading shadow memory and encrypting 16 bytes doesn't take long. and since the huk is only used to derive other keys performance shouldn't be that critical anyway (getting the HUK cant not be in any realtime path)

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+
+		memcpy(otp_data + n * sizeof(uint32_t), &word,
+		       sizeof(uint32_t));
+	}
+
+	if (IS_ENABLED(CFG_STM32MP1_HUK_BSEC_DERIVE_GCM)) {
+		/* The secret is not per-device unique: derive using the UID */
+		type = "Derived";
+		res = aes_gcm_encrypt_uid(otp_data, HW_UNIQUE_KEY_LENGTH,
+					  hwkey->data,  &len);
```
etienne-lms, line 120: remove 1 of the 2 space chars before `&len`

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+			return res;
+		}
+
+		res = stm32_bsec_read_otp(&word, bsec_base + n);
+		if (res) {
+			EMSG("huk: read OTP secret");
+			return res;
+		}
+
+		res = stm32_bsec_write_otp(tmp, bsec_base + n);
```
etienne-lms, line 106: Why this process of loading (read shadow mem then reload shadow mem and write shadow into shadow mem?

Could call `stm32_bsec_shadow_otp()` once at at initialization (to load shadow mem with fuse data) and use `stm32_bsec_read_otp()` for each call to read the shadowed data.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+		if (res) {
+			EMSG("huk: shadow the secret");
+			return res;
+		}
+
+		memcpy(otp_data + n * sizeof(uint32_t), &word,
+		       sizeof(uint32_t));
```
etienne-lms, line 113: nitpicking: fits in a single line

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+		/* The secret is unique: use it as the HUK */
+		memcpy(hwkey->data, otp_data, HW_UNIQUE_KEY_LENGTH);
+	}
+
+	if (!initialized) {
+		initialized = true;
+		IMSG("HUK [%s %s]", type, state);
+		DHEXDUMP(hwkey->data, HW_UNIQUE_KEY_LENGTH);
+	}
+
```
etienne-lms, line 135: It would be nice to have `memzero_explicit(otp_data, sizeof(otp_data))` + `memzero_explicit(&tmp, sizeof(tmp))` on exit (including some error paths)?

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+			return res;
+		}
+
+		memcpy(otp_data + n * sizeof(uint32_t), &word,
+		       sizeof(uint32_t));
+	}
+
+	if (IS_ENABLED(CFG_STM32MP1_HUK_BSEC_DERIVE_GCM)) {
+		/* The secret is not per-device unique: derive using the UID */
+		type = "Derived";
```
etienne-lms, line 118: "Derived-GCM"? to reflect `CFG_STM32MP1_HUK_BSEC_DERIVE_GCM`

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+		0xe0, 0x42, 0x6c, 0x0f, 0xa1, 0x33,
+	};
+	uint8_t tag[16] = { 0 };
+	TEE_Result res = TEE_ERROR_GENERIC;
+	size_t tag_len = sizeof(tag);
+	size_t uid_len = sizeof(uid);
+	uint32_t word = 0;
+	void *ctx = NULL;
+	size_t n = 0;
+
```
etienne-lms, line 34: for consistency:
`static_assert(HW_UNIQUE_KEY_LENGTH >= (UID2_OTP - UID0_OTP + 1) * sizeof(uint32_t));`

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+	for (n = 0; n < (UID2_OTP - UID0_OTP + 1); n++) {
+		res = stm32_bsec_read_otp(&word, UID0_OTP + n);
```
etienne-lms, line 36: I missed that before. First call should load fuses into shadow-mem before it's read (e.g. `stm32_bsec_shadow_otp()`).
Or could use `stm32_bsec_shadow_read_otp()` each time UID

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -254,6 +254,15 @@ CFG_WITH_NSEC_UARTS ?= y
 # UART instance used for early console (0 disables early console)
 CFG_STM32_EARLY_CONSOLE_UART ?= 4
 
+# Disable the HUK by default as it requires CFG_STM32MP1_HUK_BSEC_BASE
+# which is product specific
+CFG_STM32MP1_HUK_BSEC ?= n
+ifeq ($(CFG_STM32MP1_HUK_BSEC),y)
+# An AES-GCM encryption key is retrieved from OTP and used to encrypt the UID
+# OTP in order to generate the HUK. Otherwise the HUK is read from OTP (default)
+CFG_STM32MP1_HUK_BSEC_DERIVE_GCM ?= n
```
etienne-lms, line 263: Nitpicking again: i would rather name it `CFG_STM32MP1_HUK_BSEC_DERIVE_DEVUID` as the purpose is more explicit, that is deriving HUK using device UID.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+
+	crypto_authenc_final(ctx);
+	crypto_authenc_free_ctx(ctx);
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	uint32_t bsec_base = CFG_STM32MP1_HUK_BSEC_BASE;
```
etienne-lms, line 67: I wonder whether the configuration should also define the number of fuses used to store the key.
I could then ensure it complies with OP-TEE expectation:
`static_assert(CFG_STM32MP1_HUK_BSEC_COUNT * sizeof(uint32_t) == HW_UNIQUE_KEY_LENGTH)`

core/arch/arm/plat-stm32mp1/plat_huk.c
etienne-lms: Not exactly. https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-stm32mp1/nsec-service/bsec_svc.c#L41-L50 attempts the read the fuse, bypassing shadow mem.

Mainline OP-TEE/stm32mp1 has no specific strategy on using the BSEC locks. OP-TEE components only read fuse values, that is load fuse in shadow mem and read them from that shadow memory. The values you can read from fuses depend on the state of the several BSEC locks. If an early boot stage or an earlier power cycle sequence burned some BSEC locks, OP-TEE will end in using the values found in shadow mem that may differ from fuses effective values. There are 4 locks around OTPs (with few specific constraints). One is persistent across power cylces and forbids programming of a fuse unit (permanent_lock). 3 locks are sticky (unlocked at system reset): one to forbid raw programming of shadow memory (sw_lock), one to forbid reloading of fuse in shadow mem (sr_lock) and last one to forbid programming of fuse unit (sp_lock, it's a sticky variant of permanent lock).

So if you want to ensure using fuses definitive values:
- check permanent_lock is set for each unit read.
- check sw_lock is set for each unit read.
- load fuse units in shardow mem and read from there.
- lock sw_lock if you want to forbid further modification on HUK value
- if all permanent_lock's and sw_lock's were set and we've locked sw_lock's
  then we're using "Locked" keys, otherwise, we're using "Devel" keys.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -254,6 +254,15 @@ CFG_WITH_NSEC_UARTS ?= y
 # UART instance used for early console (0 disables early console)
 CFG_STM32_EARLY_CONSOLE_UART ?= 4
 
+# Disable the HUK by default as it requires CFG_STM32MP1_HUK_BSEC_BASE
+# which is product specific
+CFG_STM32MP1_HUK_BSEC ?= n
```
ldebieve, line 259: I would prefer to call the switch with CFG_STM32MP15_*** as, STM32MP13 as a real HUK that will be managed in another way.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+	/* HUK is the AES-GCM encrypted UID using a secret key from OTP */
+	res = crypto_authenc_alloc_ctx(&ctx, TEE_ALG_AES_GCM);
+	if (res)
+		return res;
+
+	res = crypto_authenc_init(ctx, TEE_MODE_ENCRYPT, key, key_len,
+				  nonce, sizeof(nonce),
+				  TEE_AES_BLOCK_SIZE, 0, uid_len);
+	if (res)
+		return res;
```
ldebieve, line 52: I would expect in that error case that the context must be cleaned before returning error.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,138 @@
...
+
+	res = crypto_authenc_enc_final(ctx, uid, sizeof(uid), out, out_len,
+				       tag, &tag_len);
+	if (res)
+		return res;
```
ldebieve, line 57: ditto

core/arch/arm/plat-stm32mp1/sub.mk
```diff
@@ -3,6 +3,7 @@ global-incdirs-y += .
 srcs-y += main.c
 srcs-y += reset.S
 srcs-$(CFG_SCMI_MSG_DRIVERS) += scmi_server.c
+srcs-$(CFG_STM32MP1_HUK_BSEC) += plat_huk.c
```
ldebieve, line 6: As already comment, the file can be better called stm32mp15_huk.c as it will be MP15 specific.

core/arch/arm/plat-stm32mp1/plat_huk.c
ldts: um no the semicolon will land at 81

core/arch/arm/plat-stm32mp1/conf.mk
ldts: sure I can do that. CFG_STM32MP1_HUK_BSEC_DERIVE_DEV_UID

core/arch/arm/plat-stm32mp1/plat_huk.c
ldts: yeah, seems safer doing that.

core/arch/arm/plat-stm32mp1/sub.mk
ldts: Perhaps we could keep the same file name and resolve the implementation internally?

ldebieve: we already design a solution that will not use at all this way of working, upstream will come soon for MP13.

ldts: you will still need to provide `tee_otp_get_hw_unique_key`

ldebieve: On stm32mp13, the HUK is not accessible by the CPU. We will proposed on that a weak on the `huk_subkey_derive` function rather. In that case the `tee_otp_get_hw_unique_key` implementation will not be used at all.

ldts: Seems weird to me to break that interface...so what will that huk interface provide fomr mp13? because it has to do something....
in any case, I see no reason for not merging what I am proposing... you can propose your change later on and do the necessary modifications....I believe this is the way the open source projects usually works.

ldebieve: You're right, the future implementation is out of scope of the discussion. My point is that the file named plat_huk.c means usable for all stm32mp1 which is the purpose of my comment because this is not usable on MP13 where the huk is not coming directly from OTP. So, I'd prefer, for more clarity to keep that under specific stm32mp**15** implementation.

ldts: right but regardless - no matter what smt32mp13 does - OP-TEE must have a` tee_otp_get_hw_unique_key` function which is all that this file provides for stm32mp1

I think it makes sense this way. Then for mp13 we can just add comments - if that is all it is needed - on the file perhaps describing its implementation.

Otherwise this plat-huk.c will have to go into core/drivers/stmp32mp15-huk.c and users will have to dig for the huk information for mp13...it makes it all more difficult to follow IMO.

etienne-lms: As Lionel mentioned in https://github.com/OP-TEE/optee_os/pull/5555#discussion_r987711168, mp13 will not implement `tee_otp_get_hw_unique_key()` but rather `huk_subkey_derive()` (which needs to be overridable) subce the hw will do the huk derivation, OP-TEE won't manipulate the huk itself. Hence another implementation.

I would suggest renaming s/CFG_STM32MP1_HUK/CFG_STM32MP15_HUK/.

I think our dear maintainers prefer to have drivers in _core/drivers/_ rather than _core/.../plat-xxxx/_.  In that scope _core/drivers/stm32mp15_huk.c_ could be a good candidate.

ldts: Sorry but no, that is not true: mp13 - as every OP-TEE platform - has to have an implementation of tee_otp_get_hw_unique_key(): it can provide it explicitly or fallback to the weak implementation. But there has to be one implementation.

This PR is just explicit about it. It doesn't preclude changes to huk_subkey_derive()....so I dont see what is wrong with it.

etienne-lms: I don't agree. For a platform not accessing a huk but only able to perfom derivation, there should be not need (no point) to implement  `tee_otp_get_hw_unique_key()`. In the generic code, `tee_otp_get_hw_unique_key()` is only called from _huk_subkey.c_ in the purpose of generating the derived keys. This where function `huk_subkey_derive()` will need to be overrideable by a platform specific implementation.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,182 @@
...
+#include <drivers/stm32_bsec.h>
+#include <kernel/tee_common_otp.h>
+#include <platform_config.h>
+#include <string.h>
+#include <string_ext.h>
+
+#if !defined(CFG_STM32MP15)
+#define CFG_STM32MP15_HUK_BSEC_BASE (STM32MP1_OTP_MAX_ID + 1)
+#define CFG_STM32MP15_HUK_BSEC_COUNT (0)
+#endif
```
etienne-lms, line 19: remove this part. This code will be only embedded when `CFG_STM32MP15=y`.

core/arch/arm/plat-stm32mp1/plat_huk.c
```diff
@@ -0,0 +1,182 @@
...
+	memzero_explicit(otp_data, sizeof(otp_data));
+	if (res) {
+		EMSG("HUK unavailable");
+		memzero_explicit(hwkey->data, HW_UNIQUE_KEY_LENGTH);
+	}
+
+	return res;
+}
+
+static TEE_Result stm32mp13_hw_unique_key(struct tee_hw_unique_key *hwkey)
```
etienne-lms, line 160: You can remove this function.

core/arch/arm/plat-stm32mp1/plat_huk.c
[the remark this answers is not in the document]
etienne-lms: Ok. Having this derivation specifically under `CFG_STM32MP15_HUK_BSEC_DERIVE_DEV_UID` looks right to me.

core/arch/arm/plat-stm32mp1/sub.mk
jforissier: Let's not anticipate too much here. I tend to agree with @etienne-lms that the proper interface between the generic code and the platform-specific code shall become `huk_subkey_derive()` rather than `tee_otp_get_hw_unique_key()`, but until then let's keep things as simple and clear as possible. Referring to Lionel's comment:

>  As already comment, the file can be better called stm32mp15_huk.c as it will be MP15 specific.

...and Etienne's one:

> I would suggest renaming s/CFG_STM32MP1_HUK/CFG_STM32MP15_HUK/.

...both seem totally reasonable to me. This code is intended for use by MP15 so call it MP15, period. Whether some MP15 code will fall under MP13 or MP1 we can decide later.

ldts: ok lets do that and keep it simple: let me send my PR in the simplest possible terms and if someone needs additional functionality it can happen later.

ldts:
Minimal version updated.

core/drivers/stm32mp15_huk.c
```diff
@@ -0,0 +1,152 @@
...
+				       out, out_len, tag, &tag_len);
+	if (ret)
+		goto auth_err;
+
+	crypto_authenc_final(ctx);
+auth_err:
+	crypto_authenc_free_ctx(ctx);
+err:
+	memzero_explicit(uid, uid_len);
+	memzero_explicit(tag, tag_len);
```
etienne-lms, line 94: `uid` is not sensitive ,no need for memzeor_explicit.
I don't think `tag` is neither.

core/drivers/stm32mp15_huk.c
```diff
@@ -0,0 +1,152 @@
...
+	uint32_t *key = NULL;
+	uint8_t *p = NULL;
+	bool lock = false;
+
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_0 < STM32MP1_OTP_MAX_ID);
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_1 < STM32MP1_OTP_MAX_ID);
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_2 < STM32MP1_OTP_MAX_ID);
+	static_assert(CFG_STM32MP15_HUK_BSEC_KEY_3 < STM32MP1_OTP_MAX_ID);
+
+	key = mempool_calloc(mempool_default, 1, len);
```
etienne-lms, line 114: prefer `malloc()` for 16 bytes. Even a 4 32bit array is fine allocated in the stack is ok.

etienne-lms (commented):
Thanks @ldts for the update.

Could you bring back `CFG_STM32MP15_HUK_DERIVE_DEVUID` to condition the derivation?

You prefer 4 `CFG_STM32MP15_HUK_BSEC_KEY_x` (0/1/2/3) config switches rather than 2 `CFG_STM32MP15_HUK_BSEC_BASE/_COUNT`? Why not, I have not strong opinion.

ldts:
> Thanks @ldts for the update.
> 
> Could you bring back `CFG_STM32MP15_HUK_DERIVE_DEVUID` to condition the derivation?
> 

Right, well it is no longer needed. this commit is just about a simple HUK that does not required unique per-device data on the fuses.
We can add more flexibility on another commit- just want to keep this simple

> You prefer 4 `CFG_STM32MP15_HUK_BSEC_KEY_x` (0/1/2/3) config switches rather than 2 `CFG_STM32MP15_HUK_BSEC_BASE/_COUNT`? Why not, I have not strong opinion.

right. as you can see the BSEC_ now mentions a KEY and ...I think it is easier to read.

etienne-lms:
> > Could you bring back CFG_STM32MP15_HUK_DERIVE_DEVUID to condition the derivation?
> 
> Right, well it is no longer needed. this commit is just about a simple HUK that does not required unique per-device data on the fuses.
> We can add more flexibility on another commit- just want to keep this simple

What should be platform default config, HUK from raw OTPs or HUK from OTPs with a devID derivation?
I think it is the former, hence a config switch that is default disabled as in your previous post.

Maybe you prefer that I create a P-R on top of your change to introduce:
CFG_STM32MP15_HUK_BSEC_DERIVE_DEVUID=y|n
CFG_STM32MP15_HUK_BSEC_RAW=y|n
and mandate one and only one one is enabled when CFG_STM32MP15_HUK=y.


> > You prefer 4 CFG_STM32MP15_HUK_BSEC_KEY_x (0/1/2/3) config switches rather than 2 
> > CFG_STM32MP15_HUK_BSEC_BASE/_COUNT? Why not, I have not strong opinion.
> 
> right. as you can see the BSEC_ now mentions a KEY and ...I think it is easier to read.

`_KEY_0/1/2/3` are BSEC word indices, not key IDs.
And that make 4 configs to set instead of 1 (assuming the `_COUNT` one is kept to its default value).

ldts:
In my view the less restrictive option to the user should be the default: and that is the one that does not need any fuses with per-device unique secrets, just one set of common values across the fleet of devices. That makes manufacturing/provisioning easier.

Yes please, I think both should deserve their own commits with their specific information...so they can be reviewed later on: adding too many options in a single commit makes documenting and explaining too expensive as I have been experiencing first hand. And it doesn't need to be like this.

In this commit, those configs represent words (32 bits) of a 128 bit key: so really key_word0, key_word1, key_word2 and key_word3

core/drivers/stm32mp15_huk.c
etienne-lms: From the SoC spec, UID fuses are in the "non-secure" fuses area, meaning non-secure world can read it any time.
For the tag, I don't think it can evade information from derived key.

As for the zynqmp drier, sorry i did not catch this issue. 
Not a big deal here, but let's have code that do what's needded and don't do what is not.

core/drivers/stm32mp15_huk.c
ldts: but isnt mempool supposed to be used as a stack replacement? short lived memory? at least that is what the commit adverstises...


```
+ * Allocation of temporary memory buffers which are used in a stack like
+ * fashion....


```

ldts: https://github.com/OP-TEE/optee_os/blob/master/lib/libutils/ext/mempool.c#L20

jenswi-linaro: No, it's supposed to be used when you need larger chunks. Only one thread at a time can use the mempool, that's why it's not likely to fail.

core/drivers/stm32mp15_huk.c
```diff
@@ -0,0 +1,150 @@
...
+#include <string.h>
+#include <string_ext.h>
+
+static bool stm32mp15_huk_init;
+
+static TEE_Result stm32mp15_read_uid(uint32_t *uid)
+{
+	uint32_t *q = uid;
+
+	if (stm32_bsec_shadow_read_otp(q++, UID0_OTP))
```
etienne-lms, line 23: For a real locked state result in `tee_otp_get_hw_unique_key()`, you would need here to call `stm32mp15_read_otp()` for each `OTP_UIDx`.
That's not a mandatory fix. Locked/unlocked state only triggers a console trace.

ldts: um, also on the UIDs? Is there a risk to this values changing between reads?

core/drivers/stm32mp15_huk.c
ldts: ah I can see the pool's mutex...I get it now

core/drivers/stm32mp15_huk.c
etienne-lms: Reading again the spec, there are some useful details.

IIUC the UID words (words 13-15: ID0, ID1, ID2) are always permantent locked but the shadow memory for UID words can be sticky locked of not. To ensure you use fuses definitive values, optee should test sw_lock is set, or more simply lock sw_lock for thoses words.

Regdarding where HUK will be placed (so-called upper OTPs), when the device is in "secure-closed" state, the shadow memory for these words is fully locked (no reload/reprogam) so no need to shadow and to test locks. When not "secure-closed" state, the several word locks may be released. In this case, using definitive fuses requires the sequence you implmented (shadow value + test locks).

ldts: thanks a lot for checking these details @etienne-lms . I have incorporated your additions.

ldts:
@jenswi-linaro @etienne-lms what is missing on this PR? anything else for me to add?

ldts:
anything?

core/drivers/stm32mp15_huk.c
```diff
@@ -0,0 +1,157 @@
...
+
+	ret = stm32mp15_read_otp(CFG_STM32MP15_HUK_BSEC_KEY_2, key++, &lock);
+	if (ret)
+		goto out;
+
+	ret = stm32mp15_read_otp(CFG_STM32MP15_HUK_BSEC_KEY_3, key++, &lock);
+	if (ret)
+		goto out;
+
+	ret = aes_gcm_encrypt_uid((uint8_t *)otp_key, len, hwkey->data, &len);
```
etienne-lms, line 142: LGTM.
As per discussion, I will create a P-R to condition the uid encryption upon `CFG_STM32MP15_HUK_BSEC_DERIVE_UID?=n` so that when enabled, the default HUK used is the one programmed in the fuses. @ldts, is that Ok with you?

core/drivers/stm32mp15_huk.c
```diff
@@ -0,0 +1,157 @@
...
+
+	ret = crypto_authenc_enc_final(ctx, (uint8_t *)uid, sizeof(uid),
+				       out, out_len, tag, &tag_len);
+	if (ret)
+		goto auth_err;
+
+	crypto_authenc_final(ctx);
+auth_err:
+	crypto_authenc_free_ctx(ctx);
+err:
```
etienne-lms, line 106: sorry I missed that. this is not a error only path so relabel to `out_free_ctx` annd `out`.

etienne-lms (changes requested):

core/drivers/stm32mp15_huk.c
ldts: yep that seems like a good plan. Do you wish to maintain this driver?

core/drivers/stm32mp15_huk.c
ldts: sorry I'll never get used to this...ok will change,.

ldts:
updated

core/drivers/stm32mp15_huk.c
ldts: @etienne-lms ^^

etienne-lms: Ok for me.

core/drivers/stm32mp15_huk.c
```diff
@@ -0,0 +1,157 @@
...
+
+	ret = crypto_authenc_enc_final(ctx, (uint8_t *)uid, sizeof(uid),
+				       out, out_len, tag, &tag_len);
+	if (ret)
+		goto out_free_ctx;
+
+	crypto_authenc_final(ctx);
+out_free_ctx:
+	crypto_authenc_free_ctx(ctx);
+out:
```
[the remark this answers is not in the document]
ldts: I dont see any problems with it - also -_Wall_ wont say a thing either.

core/drivers/stm32mp15_huk.c
ldts: ok so when you send your PR maybe you can add yourself to the maintainer file for this one? thanks for this too.

etienne-lms (commented):

PR #5557, opened 2022-09-27
plat-stm32mp1: add support for i2c5 bus

igoropaniuk:
This allows stm32_i2c driver to properly initialize and use i2c5 bus on stm32mp15 SoC.

Signed-off-by: Michael Scott <mike@foundries.io>
Signed-off-by: Igor Opaniuk <igor.opaniuk@foundries.io>


etienne-lms (commented):
On stm32mp15, I2C5 SoC subsystem cannot be assigned to secure world. Hence the I2C5 clock and reset contollers from the SoC RCC subsystem can be modified by non-secure world at will. This is why optee/stm32mp1 only defines I2C4/I2C6 clocks/reset and not the ones for I2C1/2/3/5.

igoropaniuk:
@etienne-lms the reason we need that is we use NXP SE050 on STM32MP1-disco, connected via Arduino connector (which exposes I2C5 bus).

Initialial initialization of SE050 chip is done by OP-TEE, then further interaction is delegated to REE. Related commits done by @ldts :
https://github.com/OP-TEE/optee_os/commit/613c63090b00ac971fb02ac74f7f23e6c751972c
https://github.com/linaro-swg/linux/commit/c05210ab975771e161427eb47696b869d820bdaf

Considering information from above
> I2C5 SoC subsystem cannot be assigned to secure world.

is not a problem.

> Hence the I2C5 clock and reset contollers from the SoC RCC subsystem can be modified by non-secure world at will.

just like this one. This setup obviously implies the possibility of denial of service attacks, however not disclosure of data transfering between OP-TEE OS core and SE050 chip.

core/drivers/clk/clk-stm32mp15.c
```diff
@@ -353,6 +354,8 @@ static const struct stm32mp1_clk_gate stm32mp1_clk_gate[] = {
 	_CLK_FIXED(SEC, RCC_DDRITFCR, 9, DDRPHYCAPB, _PCLK4),
 	_CLK_FIXED(SEC, RCC_DDRITFCR, 10, DDRPHYCAPBLP, _PCLK4),
 
+	_CLK_SC2_SELEC(SEC, RCC_MP_APB1ENSETR, I2C5EN, I2C5_K, _I2C35_SEL),
```
etienne-lms, line 357: I would suggest you add I2C5 clock under a config switch `#ifdef CFG_WITH_NSEC_I2CS`, with s/`SEC`/``N_S`/ as 1st argument.

core/arch/arm/plat-stm32mp1/shared_resources.c
```diff
@@ -596,6 +610,8 @@ static void set_etzpc_secure_configuration(void)
 			    shres2decprot_attr(STM32MP1_SHRES_SPI6));
 	config_lock_decprot(STM32MP1_ETZPC_I2C4_ID,
 			    shres2decprot_attr(STM32MP1_SHRES_I2C4));
+	config_lock_decprot(STM32MP1_ETZPC_I2C5_ID,
+			    shres2decprot_attr(STM32MP1_SHRES_I2C5));
```
etienne-lms, line 614: I2C5 subsystem will no be affected by this change. I think `STM32MP1_SHRES_I2C5` should not be added as the underlying HW will always be accessible from the non secure world.

Removing definition and support of `STM32MP1_SHRES_I2C5`, I think the change you need in shared_resrouces.c is:
```patch
--- a/core/arch/arm/plat-stm32mp1/shared_resources.c
+++ b/core/arch/arm/plat-stm32mp1/shared_resources.c
@@ -319,6 +319,10 @@ static void register_periph_iomem(vaddr_t base, enum shres_state state)
        case UART7_BASE:
        case UART8_BASE:
        /* Fall through */
+#endif
+#ifdef CFG_WITH_NSEC_I2CS
+       case I2C5_BASE:
+       /* Fall through */
 #endif
        case IWDG2_BASE:
                /* Allow drivers to register some non-secure resources */

**Review by etienne-lms — COMMENTED** [review]
Hi @igoropaniuk, sorry for the late feedback.
I think you should not define this extra shared-resrouce and define straight I2C5 clock as a non-secure clock. I suggest a new config switch CFG_WITH_NSEC_I2CS to enable use of I2C non secure interfaces, like existing `CFG_WITH_NSEC_GPIOS`/`CFG_WITH_NSEC_UARTS` for gpios and uarts.

File: core/drivers/clk/clk-stm32mp15.c [hunk 3]
```diff
@@ -467,6 +476,7 @@ static const uint8_t rtc_parents[] = {
 static const struct stm32mp1_clk_sel stm32mp1_clk_sel[_PARENT_SEL_NB] = {
 	/* Secure aware clocks */
 	_CLK_PARENT(_STGEN_SEL, RCC_STGENCKSELR, 0, 0x3, stgen_parents),
+	_CLK_PARENT(_I2C35_SEL, RCC_I2C35CKSELR, 0, 0x7, i2c35_parents),
```
etienne-lms, line 479: this one should fall in the `/* Always non-secure clocks ... */` list right below.
Would you mind moving it and condition with `#ifdef CFG_WITH_NSEC_I2CS`?

core/drivers/clk/clk-stm32mp15.c
```diff
@@ -421,6 +426,10 @@ static const uint8_t stgen_parents[] = {
 	_HSI_KER, _HSE_KER
 };
 
+static const uint8_t i2c35_parents[] = {
+	_PCLK1, _PLL4_R, _HSI_KER, _CSI_KER
+};
```
etienne-lms, line 431:
```
#ifdef CFG_WITH_NSEC_I2CS
#endif
```

etienne-lms (commented):
LGTM with the 2 remaining comments addressed.

etienne-lms:
> Let me know if it makes sense to address it.

I think `/* FALLTHROUGH */` or `/* fallthrough */` is preferred. First for consistency in optee_os source tree as these are used,  and second for consistency also: our inherited checkpatch rule set this constraint, fine, it serves our source tree consistency.

jforissier:
@etienne-lms no, we do have a `fallthrough` "keyword" (a macro actually), it should be used in new code since Linux/checkpatch prefers it.

igoropaniuk:
@jforissier @etienne-lms addressed `fallthrough` warning

etienne-lms:

igoropaniuk:
@etienne-lms R-b tag applies, thanks!

igoropaniuk:
@jforissier @etienne-lms there were other instances of Fallthrough comments in `core/arch/arm/plat-stm32mp1/shared_resources.`, so replaced them in a separate commit.

etienne-lms (commented):
Thanks @igoropaniuk for tha additional  fix.
for commit "plat-stm32mp1: fallthrough macro instead of comment".

igoropaniuk:
@etienne-lms tag applied, thanks

PR #5561, opened 2022-09-29
ci: QEMUv8_check: also run tests with Cryptographic Extension enabled

jforissier:
Adds a line to build and run regression tests with CFG_CRYPTO_WITH_CE=y. Many Armv8 CPUs support this extension so it is very commonly used and deserves to be tested here.

While we're at it, set CFG_TEE_CORE_LOG_LEVEL=0 at the beginning to avoid duplication.


jenswi-linaro:

PR #5565, opened 2022-10-03
Adding support for configuring ZCU102 with 4 or 16GB RAM

thehoustonian:
The Xilinx ZCU102 can be configured with 16GB of RAM, even though the default is 4GB. This pull request adds the ability to configure the ZCU102 with 16GB of RAM via the `CFG_DRAM_SIZE_GB` variable. I've tried testing it locally, and it seems to work for me.

### DIFF
--- a/core/arch/arm/plat-zynqmp/conf.mk
+++ b/core/arch/arm/plat-zynqmp/conf.mk
@@ -27,9 +27,13 @@ CFG_UART_BASE ?= UART0_BASE
 CFG_UART_IT ?= IT_UART0
 CFG_UART_CLK_HZ ?= UART0_CLK_IN_HZ
 
-# ZCU102 features 4 GiB of DDR
+# ZCU102 can be configured with 4GB or 16GB RAM modules.
 ifeq ($(CFG_ARM64_core),y)
+ifeq ($(CFG_DRAM_SIZE_GB),16)
+CFG_DDR_SIZE ?= 0x400000000
+else
 CFG_DDR_SIZE ?= 0x100000000
+endif
 else
 # On 32 bit build limit to 2 GiB of RAM
 CFG_DDR_SIZE ?= 0x80000000


jforissier:
CC @ricardosalveti 

@thehoustonian thanks for the patch, we will need a Signed-off-by: (see https://optee.readthedocs.io/en/latest/general/contribute.html) as well as some text in the commit description, for example "The Xilinx ZCU102 can be configured with 16GB of RAM, even though the default is 4GB. Introduce CFG_DRAM_SIZE_GB=16 to support 16GB." or something like that.

thehoustonian:
@jforissier No problem, I updated the commit message. Let me know if there's anything else I need to do. Thanks!

ricardosalveti:
This is the first time I noticed such config, as it seems some other targets are just using CFG_DDR_SIZE directly, which could also be used here since it got a weak assignment.

@jforissier I'm ok with the change, but not sure what is the preferred CFG variable to use here from the project perspective.

jforissier:
@ricardosalveti thanks for your feedback, TBH I would also prefer not to introduce a new CFG, even though `CFG_DDR_SIZE=0x400000000` is a bit hard to parse for humans -- if that's an issue we could `#define SZ_1G 0x40000000` in some header file so that `CFG_DDR_SIZE="(16*SZ_1G)"` would be an acceptable syntax.
Another option is to define a platform flavor for the 16G version, but is it worth it?
@thehoustonian what do you think?

thehoustonian:
@jforissier @ricardosalveti I don't think it's worth creating a new platform flavor, since other RAM sizes may be legitimate (the RAM module is removable, and I'm not sure Xilinx necessarily limits boards to just 4 or 16GB sticks). 

I chose `CFG_DRAM_SIZE_GB` based on what I saw [here](https://github.com/OP-TEE/optee_os/blob/203ee23d005b2cec2f21b5de334c5a246be32599/core/arch/arm/plat-hikey/platform_config.h#L114), but it seems like the `CFG_DDR_SIZE` may be more common across the other platform configs. 

I think it makes sense to align with what's more common. If `CFG_DDR_SIZE` is preferred, then maybe this patch is not needed and I can manually set that when building?

jforissier:
> I think it makes sense to align with what's more common. If `CFG_DDR_SIZE` is preferred, then maybe this patch is not needed and I can manually set that when building?

Absolutely. All you need to do is append `CFG_DDR_SIZE=0x400000000` to your `make` command, or you can also set it as an environment variable.

thehoustonian:
@jforissier Alright, that sounds like a good plan. Thanks for the feedback! I'm going to close this pull request then, since it is not needed.

PR #5568, opened 2022-10-05
plat-stm32mp1: conf: fix tzdram default size when w/o rsv-shm

etienne-lms:
Fix the default TZDRAM size that is 32MByte when CFG_CORE_RESERVED_SHM
is disable, not 30MByte.


jforissier (approved):

etienne-lms:
tag applied.
thanks.

Issue #5338, opened 2022-05-16
Programming RPMB key, or how to avoid CFG_RPMB_WRITE_KEY=y


jenswi-linaro:
Hi all!

Problems with programing the RPMB key has been discussed recently in 
https://github.com/OP-TEE/optee_os/pull/5323 and https://github.com/OP-TEE/optee_docs/pull/157.

The config variable `CFG_RPMB_WRITE_KEY` must be set to `y` for OP-TEE to do this. But there are a few problems with this. From security point of view it may lead to a class break if ever such a signed OP-TEE binary is leaked. Then there's the accidents with writing unusable keys which in worst case may brick the device. Vendors will do well to never use `CFG_RPMB_WRITE_KEY=y`, but since the option is there it's a risk that it's just too convenient to use.

The best would be if this wasn't done by OP-TEE at all. If secure fuses etc used to derive the RPMB key are known the RPMB key can be calculated offline, possibly with help from a script for the last step. Typically the steps done by `huk_subkey_derive()` after the HUK has been retrieved with `tee_otp_get_hw_unique_key()`.

With the key calculated independent of OP-TEE we could let U-Boot and/or some user space application in Linux program the key during production. As a safety step to avoid accidents we can add a PTA function to verify that the calculated RPMB key matches OP-TEE's view. Providing such a function should not be a security problem since an eventual attacker just as well could do a brute force on the RPMB itself.

From testing point of view in CI we could add the capability to xtest to write the RPMB key if needed on QEMU. For real hardware I assume that the key will be programmed before it's added to a CI loop.

Thoughts, comments?

Thanks,
Jens

jforissier:
> If secure fuses etc used to derive the RPMB key are known the RPMB key can be calculated offline

This may not always be true. If the device uses a derivation mechanism where the root key cannot leave the chip, then the key can only be generated by software running on the device (and probably only in secure world?). In this case some special firmware would be needed to program the RPMB key no matter what, be it a special build of OP-TEE or something else.

jenswi-linaro:
OK. I had the impression that even if the root key doesn't leave the chip it would still be known because it was known how the chip was fused during production. But if some firmware randomizes and program this early in the life cycle of the platform then it will truly unknown to the outside.

[...]

PR #5570, opened 2022-10-05
core: huk definition

ldts:
The RPMB key derivation algorithm consumes the HUK and therefore is impacted by changes in its value.

The platform specific function plat_rpmb_key_is_ready() should be allowed to query the HUK in order to allow/deny burning the RPMB key.

### DIFF
--- a/core/include/kernel/tee_common_otp.h
+++ b/core/include/kernel/tee_common_otp.h
@@ -13,6 +13,8 @@
 
 struct tee_hw_unique_key {
 	uint8_t data[HW_UNIQUE_KEY_LENGTH];
+	/* The hardware unique key is now inmmutable */
+	bool locked;
 };
 
 TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey);
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -18,6 +18,7 @@
 #include <kernel/linker.h>
 #include <kernel/misc.h>
 #include <kernel/panic.h>
+#include <kernel/tee_common_otp.h>
 #include <kernel/tee_misc.h>
 #include <kernel/thread.h>
 #include <kernel/tpm.h>
@@ -1298,6 +1299,8 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
  */
 void __weak boot_init_primary_late(unsigned long fdt)
 {
+	struct tee_hw_unique_key huk = { };
+
 	init_external_dt(fdt);
 	tpm_map_log_area(get_external_dt());
 	discover_nsec_memory();
@@ -1305,10 +1308,6 @@ void __weak boot_init_primary_late(unsigned long fdt)
 	configure_console_from_dt();
 
 	IMSG("OP-TEE version: %s", core_v_str);
-	if (IS_ENABLED(CFG_WARN_INSECURE)) {
-		IMSG("WARNING: This OP-TEE configuration might be insecure!");
-		IMSG("WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html");
-	}
 	IMSG("Primary CPU initializing");
 #ifdef CFG_CORE_ASLR
 	DMSG("Executing at offset %#lx with virtual load address %#"PRIxVA,
@@ -1328,6 +1327,15 @@ void __weak boot_init_primary_late(unsigned long fdt)
 	}
 	call_finalcalls();
 	IMSG("Primary CPU switching to normal world boot");
+
+	if (IS_ENABLED(CFG_WARN_INSECURE)) {
+		/* Check drivers for unsafe configurations */
+		if (tee_otp_get_hw_unique_key(&huk))
+			huk.locked = false;
+
+		IMSG("WARNING: This OP-TEE configuration %s insecure!", !huk.locked ? "IS" : "might be");
+		IMSG("WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html");
+	}
 }
 
 static void init_secondary_helper(unsigned long nsec_entry)
--- a/core/arch/arm/kernel/otp_stubs.c
+++ b/core/arch/arm/kernel/otp_stubs.c
@@ -19,6 +19,7 @@
 
 __weak TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 {
+	huk.locked = false;
 	memset(&hwkey->data[0], 0, sizeof(hwkey->data));
 	return TEE_SUCCESS;
 }


ldts:
if this change is accepted, I will update the zynqmp and the stm32mp15_huk PR to use this value.

core/include/kernel/tee_common_otp.h
```diff
@@ -13,6 +13,8 @@
 
 struct tee_hw_unique_key {
 	uint8_t data[HW_UNIQUE_KEY_LENGTH];
+	/* The hardware unique key is now inmmutable */
+	bool locked;
```
jenswi-linaro, line 17: How does this relate to `plat_rpmb_key_is_ready()`?

jenswi-linaro:
I would rather delegate the decision of when to write the RPMB key to the normal world like in https://github.com/OP-TEE/optee_os/issues/5338

core/include/kernel/tee_common_otp.h
ldts: `plat_rpmb_key_is_ready` often depends on the platform being considered closed/secured  - ie, the bootrom booting signed firmware. However it should also depend on the HUK key being immutable (so the RPMB derived key doesnt change on a subsequent boot causing the RPMB not to be accesible). 

So IMO having some information with respect to the RPMB key derivation mutability is required. I propose to do it here since the only reason for variability in the RPMB key is the HUK  changing,  but really what plat_rpmb_key_is_ready needs to know is if the  key that has been generated can change; then it can decide whether to allow the key to be written or not.

etienne-lms: I agree with Jens. It's up to the platform to refuse or not to provide a HUK. There is no unique definition of what is a locked/closed device, possibly several levels of device closure.

ldts:
> I would rather delegate the decision of when to write the RPMB key to the normal world like in #5338

If OP-TEE calculates the key, OP-TEE will still need to let the normal world know if the key is to be trusted - which is the problem we are discussing.

Or if OP-TEE is not going to provide a key, then someone else will need to figure out how to derive (or where to store!) the RPMB key. 

The good thing about OP-TEE deriving it is that the key doesn't need to be stored.

 The problem of enforcing a secure HUK across all architectures and products is harder than the problem of letting the platform know if the HUK - and consequently the RPMB key- is a secret....I find this  a more liberal/flexible approach to a problem with to many variants that is hard to attack head on.

Products and platforms can then make their own decisions.

jenswi-linaro:
> > I would rather delegate the decision of when to write the RPMB key to the normal world like in #5338
> 
> If OP-TEE calculates the key, OP-TEE will still need to let the normal world know if the key is to be trusted - which is the problem we are discussing.

Thanks for the clarification.

> 
> Or if OP-TEE is not going to provide a key, then someone else will need to figure out how to derive (or where to store!) the RPMB key.
> 
> The good thing about OP-TEE deriving it is that the key doesn't need to be stored.

Agreed

> 
> The problem of enforcing a secure HUK across all architectures and products is harder than the problem of letting the platform know if the HUK - and consequently the RPMB key- is a secret....I find this a more liberal/flexible approach to a problem with to many variants that is hard to attack head on.
> 
> Products and platforms can then make their own decisions.

I don't see the point of the notion of an untrusted HUK. Either you have it and it can be used or you don't have it. If it's not trusted then you don't have it.

We should try to avoid being flexible as that means more ifs and buts.

ldts:
anyway, many ways of doing this but the key requirement is that HUK users need to know if the HUK is to be trusted.

So what do we do? Shall I just close this PR? do we feel it deserves a larger discussion? should we merge this proposal?

jenswi-linaro:
We should not merge this. As for the discussion, it seems we don't understand each other.

ldts:
1) Your position is that the HUK must be trusted and it should not be used without that guarantee (agree 100%)
2) Your position (I think) is also that it is acceptable to merge HUK drivers that do not deliver that guarantee. 
if 2) is true, I do not understand that: maybe you could build on it?
if 2) is not true then I need to fix https://github.com/OP-TEE/optee_os/pull/5555  
so need your input.

jenswi-linaro:
No, I disagree on 1 and 2.

ldts:
This is kind of surprising since you wrote: "I don't see the point of the notion of an untrusted HUK. Either you have it and it can be used or you don't have it. If it's not trusted then you don't have it."

what did you men then? your position is totally cryptic to me.

ldts:
This discussion is important IMO. please could we resume with it.

To provide with some background -aside of RPMB- take for example the [SCP03](https://globalplatform.org/wp-content/uploads/2014/07/GPC_2.3_D_SCP03_v1.1.2_PublicRelease.pdf) support we have in the SE05x crypto driver controlling a whole family of HSMs

SCP03 defines a set of symmetric keys that must be rotated once the platform is deemed secured. The rotation scheme that we (you and I during the review process) agreed was, rather than storing these keys in memory, we would derive their values from a secure key.  In OP-TEE this is the HUK.
As a consequence, if the SCP03 keys are rotated and the HUK then changes, the secure element will no longer be reachable (and it can not be reset, so it is lost forever...all keys and certificates and blobs no longer accessible).

So the SE05x key rotation scheme - just as the RPMB key generation scheme - would benefit from having an indication if the HUK is stable before going ahead.

Incidentally, the current __weak definition of HUK is silly for most purposes as it doesnt generate a per-device unique key- something that causes issues when dealing with fleets of devices. So perhaps that ___weak_ should use just a CID if available (so at least it is unique...at the moment is not unique nor secure so it would be a step forward) _and_ OP-TEE requires HUK drivers to generate immutable secret value and raises a security warning when that is not the case.

what do you think?

jenswi-linaro:
Why not just let the platform return an error if it detects that the HUK can't be used yet?

On QEMU and Hikey it's from a testing point of good enough to use zeroes.

ldts:
Right and that would be an option if we had a __weak implementation that provides a unique id.

What NXP does in their hardware is that while the platform is not booting secured firmware (the SoC hasnt been fused/closed), the hardware provides a real unique id. But then, once the board is secured (fused) it generates a different one which is final and therefore secured.

This is the ideal situation we have been trying to mimic on the other HUK drivers.. And the reason why `plat_rpmb_is_ready` usually checks for the closed state on those platforms. But it really is all about the HUK.

jenswi-linaro:
> Right and that would be an option if we had a __weak implementation that provides a unique id.

I suppose we could remove the `__weak` implementation and make it truly platform specific.

A HUK is supposed to be unique and secure. I don't see a point in using a non-secure HUK as NXP does, but perhaps I'm missing a use case.

For testing purposes, you can of course use any non-changing HUK, but OP-TEE still treats it as if it's unique and secure.

ldts:
I think the root of the problem is that a HUK is needed during development that is truly unique - not just a string of zeroes common to all OP-TEE devices.

if the __weak implementation returned a unique ID when executing on real hardware we could then enforce/recommend that any platform specific HUK driver must be unique and immutable and whenever possible a secret (ie, only readable when the BOOTROM is booting a signed firmware)

having a generic HUK returning a unique id when executing on various hardware should be doable (ie just use the CID value use for booting). 

Then it would be clear at build time if the HUK derived keys are secured.

jenswi-linaro:
> I think the root of the problem is that a HUK is needed during development that is truly unique - not just a string of zeroes common to all OP-TEE devices.

Why can't zeroes be used?

> 
> if the __weak implementation returned a unique ID when executing on real hardware we could then enforce/recommend that any platform specific HUK driver must be unique and immutable and whenever possible a secret (ie, only readable when the BOOTROM is booting a signed firmware)

The `__weak` implementation is generic, I don't see how it can be unique. To me, it seems complicated to have a HUK that changes depending on the security state of the platform. NXP has chosen that approach, but as I said they may have a use case I'm not seeing. What's your use case with a changing HUK?

> 
> having a generic HUK returning a unique id when executing on various hardware should be doable (ie just use the CID value use for booting).

What CID? The one from the eMMC? How can that be secure or even non-changing?

It seems I'm missing a key point in your use case.

PR #5573, opened 2022-10-06
core: crypto: remove TEE_ATTR_ECC_CURVE as an attribute of TEE_TYPE_ED25519_KEYPAIR

jforissier:
The ECC curve is not an attribute of an Ed25519 key pair, let alone a mandatory one for key generation. It was mistakenly added by commit 03e07432b68f ("ta: pkcs11: Add Ed25519 support"), thus breaking xtest regression_4007_ed25519 (subcase .1 Generate Ed25519 key). Remove that attribute from the key type definition.

Fixes: 03e07432b68f ("ta: pkcs11: Add Ed25519 support")


jenswi-linaro:

PR #5577, opened 2022-10-09
core: remove invalid check

ldts:
Current session can not return NULL as panic() aborts the execution in that case.

### DIFF
--- a/core/kernel/ts_manager.c
+++ b/core/kernel/ts_manager.c
@@ -57,11 +57,7 @@ struct ts_session *ts_pop_current_session(void)
 
 struct ts_session *ts_get_calling_session(void)
 {
-	struct ts_session *s = ts_get_current_session();
-
-	if (s)
-		s = TAILQ_NEXT(s, link_tsd);
-	return s;
+	return TAILQ_NEXT(ts_get_current_session(), link_tsd);
 }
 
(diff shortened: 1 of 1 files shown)


ldts:
any issues with this PR?

jforissier:
The subject and description are unclear, so that I had to go and check the code to understand. How about:
```
 core: ts_get_calling_session(): remove useless check

ts_get_current_session() can not return NULL, therefore remove the NULL
check in ts_get_calling_session().
```

ldts:
@jforissier thanks. I know is a bit of a silly PR but I think in the end it helps bringing up coherence to the whole pull-review process (meaning that merged code should not have the sort of issues - even if they are non-functional -  that are later on discussed during pull requests - ie, poorly chosen variable names, unneeded checks...trivial stuff like this...). TIA!

jforissier:
@ldts true, every bit matters, and that's why I prefer using exact function names rather than higher level english text in commit descriptions. For example "ts_get_current_session() can not return NULL" vs. "Current session cannot return NULL", or citing the name of the modified function (`ts_get_calling_session()`) in the text so that `git grep` can easily spot the change. No big deal though.

ldts:
@jforissier tag applied

PR #5578, opened 2022-10-10
[RFC] Sw RSA ops fallback

jenswi-linaro:
The commits:
- core: compiler.h: introduce __weak_alias(x) 
- crypto_impl: rsa: support the crypto driver 
- core: ltc: rsa: support the crypto driver
are copied from https://github.com/OP-TEE/optee_os/pull/5516, please keep the review of those to https://github.com/OP-TEE/optee_os/pull/5516

The two commits:
- core: crypto: add drvcrypt_get_rsa_ops() 
- core: crypto: add drvcrypt_rsa_sw_ops 
explores a way of using a software fallback transparent to the crypto_acipher_*rsa_*() functions in `core/drivers/crypto/crypto_api/acipher/rsa.c` and `core/drivers/crypto/crypto_api/acipher/rsassa.c `. The new function drvcrypt_get_rsa_ops() takes a key_size parameter to help selecting the right RSA ops struct.

We could if desired extend this further and store the RSA ops pointer in struct rsa_keypair and struct rsa_public_key to avoid a few calls to drvcrypt_get_rsa_ops().

### DIFF
--- a/core/drivers/crypto/caam/acipher/caam_rsa.c
+++ b/core/drivers/crypto/caam/acipher/caam_rsa.c
@@ -1543,6 +1543,7 @@ static TEE_Result do_decrypt(struct drvcrypt_rsa_ed *rsa_data)
 /*
  * Registration of the RSA Driver
  */
+static const struct drvcrypt_rsa *driver_rsa_ptr;
 static const struct drvcrypt_rsa driver_rsa = {
 	.alloc_keypair = do_allocate_keypair,
 	.alloc_publickey = do_allocate_publickey,
@@ -1555,6 +1556,11 @@ static const struct drvcrypt_rsa driver_rsa = {
 	.optional.ssa_verify = NULL,
 };
 
+const struct drvcrypt_rsa *drvcrypt_get_rsa_ops(size_t key_size_bytes __unused)
+{
+	return driver_rsa_ptr;
+}
+
 enum caam_status caam_rsa_init(struct caam_jrcfg *caam_jrcfg)
 {
 	enum caam_status retstatus = CAAM_FAILURE;
@@ -1564,8 +1570,8 @@ enum caam_status caam_rsa_init(struct caam_jrcfg *caam_jrcfg)
 		caam_era = caam_hal_ctrl_era(jr_base);
 		RSA_TRACE("CAAM Era %d", caam_era);
 
-		if (!drvcrypt_register_rsa(&driver_rsa))
-			retstatus = CAAM_NO_ERROR;
+		driver_rsa_ptr = &driver_rsa;
+		retstatus = CAAM_NO_ERROR;
 	}
 
 	return retstatus;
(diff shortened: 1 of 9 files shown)


ldts:
Hardware might also only support some algorithms and will need to call back the relevant software operations - ie see the snippet below for the Versal case where returning TEE_ERROR_NOT_IMPLEMENTED invoked their software alternatives from the driver API framework

in this case we'll need to add more parameters to the get_rsa_ops (for instance the rsa_id) to cover that scenario. Would you be open to that?


in do_encrypt
```
	if (rsa_data->key.n_size == 128)
		return TEE_ERROR_NOT_IMPLEMENTED;

	switch (rsa_data->rsa_id) {
	case DRVCRYPT_RSASSA_PSS:
	case DRVCRYPT_RSA_OAEP:
	case DRVCRYPT_RSA_NOPAD:
		return TEE_ERROR_NOT_IMPLEMENTED;
	case DRVCRYPT_RSASSA_PKCS_V1_5:
		/* IP too supports SHA-3-384 - but requires TEE Core API 1.3 */
		if (rsa_data->hash_algo == TEE_ALG_SHA1)
			return TEE_ERROR_NOT_IMPLEMENTED;

		break;
	default:
		panic();
	}

```

ldts:
I think _`drvcrypt_get_rsa_ops`_ should also request the operation ie (encrypt/decrypt/ssa_sign/ssa_verify) to make sure the parsing is proper.

jenswi-linaro:
Hmm, that's a problem because we don't know that we're going to use that algorithm in for instance crypto_acipher_alloc_rsa_keypair(). So we could potentially end up with mixed ops pointers, that is, having bignums allocated for se050, but then used by the software fallback.

jenswi-linaro:
Another option is to skip the approach with drvcrypt_get_rsa_ops() and instead let for instance `do_encrypt()` call the software function directly.
This means that se050 would have to use the bignum format of the software fallback, but I guess that's the case already and not a problem.

ldts:
@jenswi-linaro yes I think that option would be better and will make it all clear: do_encrypt() calling the final function with all the parameters required.
then each driver can decide what to provide and implement any locking if needed (we could have concurrent encrypt operations...once on hardware, others on software at driver level. Might be a good thing performance wise.

jenswi-linaro:
Good, let's close this PR then. It looks like we agree on the way forward here. Will you make the updates to do_encrypt() as needed?

ldts:
@jenswi-linaro dont we need some crypto_api owners to comment. before investing the time?

jenswi-linaro:
It might be easier to explain the plan with an easy example. But sure, if we can get some early feedback that's good.

jforissier:
> That is a very important point which I kind of forgot and explains some discussions (@jforissier, ie why LTC RSA can not be enabled with the SE05x driver); so the SE05x RSA/ECC driver can never fallback to software RSA/ECC operations since the keys generated by the driver are not real keys usable by software

Totally makes sense, thanks for the reminder.

PR #5582, opened 2022-10-13
plat-stm: rng: initialize stack variables.

ldts:
Stack variables should be initialized.

### DIFF
--- a/core/arch/arm/plat-stm/rng_support.c
+++ b/core/arch/arm/plat-stm/rng_support.c
@@ -34,7 +34,7 @@ static vaddr_t rng_base(void)
 
 static inline int hwrng_waithost_fifo_full(void)
 {
-	uint32_t status;
+	uint32_t status = 0;
 
 	do {
 		status = io_read32(rng_base() + RNG_STATUS_OFFSET);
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-stm/rng_support.c
```diff
@@ -89,8 +89,8 @@ TEE_Result hw_get_random_bytes(void *buf, size_t len)
 
 	static int nbcall;  /* debug purpose - 0 is the initial value*/
 
-	volatile uint32_t tmpval[_LOCAL_FIFO_SIZE/2];
-	int i;
+	volatile uint32_t tmpval[_LOCAL_FIFO_SIZE / 2] = { 0 };
```
jforissier, line 92: That should be `= { };`, and the checkpatch warning makes sense: why is `volatile` needed here? What kind of unwanted optimization is it supposed to avoid? (this is more a question for @etienne-lms, although this is 8 year old code :wink:)

etienne-lms: I think it's not needed at all :)

jforissier:
With `volatile` removed, preferably in a separate commit since it is a totally different purpose:

PR #5586, opened 2022-10-14
plat-stm32mp1: conf: default enable CFG_EXTERNAL_DT for mp15

etienne-lms:
Changes stm32mp1 MP15 variant default configuration for CFG_EXTERNAL_DT that is now default enabled. This is needed as mainline U-Boot and Linux may not yet define the necessary optee nodes in their DT. Therefore prefer external DT be accessed by default and let external OP-TEE configuration disable the switch if desired.

This change does not modify MP13 variant default configuration where CFG_EXTERNAL_DT is default disabled.

Acked-by: Gatien Chevallier <gatien.chevallier@foss.st.com>


jforissier:
Is this strictly identical to #5585?

etienne-lms:
yes

etienne-lms:
...but an empty line remove in the commit message.

PR #5593, opened 2022-10-17
core: libtomcrypt: fix ed25519ctx signature

etienne-lms:
Fixes signature size value not properly set from caller argument in crypto_acipher_ed25519_sign() and crypto_acipher_ed25519ctx_sign().

Prior this patch could execution fail or not fail, depending on content previously loaded in siglen stack memory cell.

### DIFF
--- a/core/lib/libtomcrypt/ed25519.c
+++ b/core/lib/libtomcrypt/ed25519.c
@@ -63,14 +63,16 @@ TEE_Result crypto_acipher_ed25519_sign(struct ed25519_keypair *key,
 				       uint8_t *sig, size_t *sig_len)
 {
 	int err;
-	unsigned long siglen;
+	unsigned long siglen = 0;
 	curve25519_key private_key = {
 		.type = PK_PRIVATE,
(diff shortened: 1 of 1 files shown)


jforissier (commented):
Oops. Can you add a Fixes: tag with the buggy commit?

etienne-lms:
ok, done. tag applied.

jenswi-linaro:

etienne-lms:
Tag added. I also added a sentence in the commit log:

> Prior this patch could execution fail or not fail **and possibly overflow
> caller passed buffer**, depending on content previously loaded in siglen
> stack memory cell.

jforissier:
> core: libtomcrypt: fix ed25519ctx signature

That is not only Ed25519ctx,  but also plain Ed25519 and Ed25519ph, right? (BTW `crypto_acipher_ed25519ctx_sign()` would be better named `crypto_acipher_ed25519_ph_ctx_sign()` since it handles both, but that's another story). I suggest "core: libtomcrypt: fix Ed25519 signature".

PR #5597, opened 2022-10-20
plat-k3: am62x: add SA2UL and TRNG support

ti-kamlesh:
Add SA2UL and TRNG support for TI SoC AM62X through OP-TEE.

Signed-off-by: Kamlesh Gurudasani <kamlesh@ti.com>

### DIFF
--- a/core/arch/arm/plat-k3/platform_config.h
+++ b/core/arch/arm/plat-k3/platform_config.h
@@ -63,26 +63,38 @@
 #define SA2UL_BASE		0x04e00000
 #define SA2UL_TI_SCI_DEV_ID	136
 #define SA2UL_TI_SCI_FW_ID	2112
+#define SA2UL_TI_SCI_FW_RGN_ID	0
 #elif defined(PLATFORM_FLAVOR_j721e)
 #define SA2UL_BASE		0x40900000
 #define SA2UL_TI_SCI_DEV_ID	265
 #define SA2UL_TI_SCI_FW_ID	1196
+#define SA2UL_TI_SCI_FW_RGN_ID	0
 #elif defined(PLATFORM_FLAVOR_j784s4)
 #define SA2UL_BASE		0x40900000
 #define SA2UL_TI_SCI_DEV_ID	-1
 #define SA2UL_TI_SCI_FW_ID	1196
+#define SA2UL_TI_SCI_FW_RGN_ID	0
 #elif defined(PLATFORM_FLAVOR_am64x)
 #define SA2UL_BASE		0x40900000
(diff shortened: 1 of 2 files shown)


jforissier:
The subject says this adds AM62x support but it is not true since commit bf1abbb3cd01 has already introduced AM62x in CI. So I would prefer: `plat-k3: am62x: add SA2UL and TRNG support`. Or would AM62x build but not run properly with the current config? In that case this change would be a fix (`plat-k3: am62x: fix SA2UL and TRNG support`)?
It may not matter much but with the current description it is impossible to know what is wrong and what is added or fixed.

Other than that:

core/arch/arm/plat-k3/platform_config.h
```diff
@@ -63,26 +63,38 @@
...
 #elif defined(PLATFORM_FLAVOR_j784s4)
 #define SA2UL_BASE		0x40900000
 #define SA2UL_TI_SCI_DEV_ID	-1
 #define SA2UL_TI_SCI_FW_ID	1196
+#define SA2UL_TI_SCI_FW_RGN_ID	0
 #elif defined(PLATFORM_FLAVOR_am64x)
 #define SA2UL_BASE		0x40900000
 #define SA2UL_TI_SCI_DEV_ID	133
 #define SA2UL_TI_SCI_FW_ID	35
+#define SA2UL_TI_SCI_FW_RGN_ID	0
```
jforissier, line 81: Sorry to comment on a commit that I have already acked... but those `#define SA2UL_TI_SCI_FW_RGN_ID 0` affect all non-am62x flavors. Are they legitimate here? Shouldn't they be part of a different commit? (what does this fix?)

ti-kamlesh: Not an issue @jforissier, anything to improve the quality. Thanks for the comment.

#define SA2UL_TI_SCI_FW_RGN_ID 0 had same value for all flavors earlier
am62x has different value SA2UL_TI_SCI_FW_RGN_ID.

There is no functional change for non am62x flavors, just added separate entry for SA2UL_TI_SCI_FW_RGN_ID.
though there is some duplication since most devices are the same, but it looks cleaner and more consistent.

Had discussion with Andrew Davis internally on this, he suggested the same. 

Please let me know your thoughts.

jforissier: Oh, I'm sorry, I missed the fact that you removed the common line:
```diff
- #define SA2UL_TI_SCI_FW_RGN_ID	0
```

So indeed all is fine. I will merge this. Thanks!

PR #5603, opened 2022-10-23
core: dt: Make it possible to alter device mapping

vesajaaskelainen:
In case where IP core device is TrustZone aware and is used by both REE
and TEE dt_map_dev() would normally cause non-secure mapping for the
device.

When selected registers in IP core are only accessible by TrustZone device
needs to be mapped with MEM_AREA_IO_SEC to cause actual AXI memory access
be made with AWPROT[1] and ARPROT[1] bits configured properly.

This adds new argument for dt_map_dev() to enable forcing mapping to be
secure or non-secure.

### DIFF
--- a/core/include/kernel/dt.h
+++ b/core/include/kernel/dt.h
@@ -70,6 +70,17 @@ enum dt_driver_type {
 };
 
 /*
+ * DT_MAP_AUTO: Uses status properties from device tree to determine mapping.
+ * DT_MAP_SECURE: Force mapping for device to be secure.
+ * DT_MAP_NON_SECURE: Force mapping for device to be non-secure.
+ */
+enum dt_map_dev_directive {
+	DT_MAP_AUTO,
+	DT_MAP_SECURE,
+	DT_MAP_NON_SECURE
+};
+
+/*
  * dt_driver_probe_func - Callback probe function for a driver.
  *
  * @fdt: FDT base address
@@ -125,10 +136,12 @@ const struct dt_driver *dt_find_compatible_driver(const void *fdt, int offs);
  * @base receives the base virtual address corresponding to the base physical
  * address of the "reg" property
  * @size receives the size of the mapping
+ * @mapping what kind of mapping is done for memory.
  *
  * Returns 0 on success or -1 in case of error.
  */
-int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size);
+int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size,
+	       enum dt_map_dev_directive mapping);
 
 /*
  * Check whether the node at @offs contains the property with propname or not.
@@ -234,7 +247,8 @@ static inline const struct dt_driver *dt_find_compatible_driver(
 }
 
 static inline int dt_map_dev(const void *fdt __unused, int offs __unused,
-			     vaddr_t *vbase __unused, size_t *size __unused)
+			     vaddr_t *vbase __unused, size_t *size __unused,
+			     enum dt_map_dev_directive mapping __unused)
 {
 	return -1;
 }
--- a/core/kernel/dt.c
+++ b/core/kernel/dt.c
@@ -86,7 +86,8 @@ int dt_enable_secure_status(void *fdt, int node)
 	return 0;
 }
 
-int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
+int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size,
+	       enum dt_map_dev_directive mapping)
 {
 	enum teecore_memtypes mtype;
 	paddr_t pbase;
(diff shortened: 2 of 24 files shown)


vesajaaskelainen:
This is split from PR https://github.com/OP-TEE/optee_os/pull/5350 as we could not re-open that.

Also decided to split the commits to own PRs.

core/kernel/dt.c
```diff
@@ -125,6 +125,39 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
 	return 0;
 }
 
+int dt_map_secure_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
```
jenswi-linaro, line 128: This duplicates most of the code in `dt_map_dev()`. Would it be possible to refactor the two functions to use a common helper function or something?

core/kernel/dt.c
```diff
@@ -125,6 +125,39 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size)
...
+	st = _fdt_get_status(fdt, offs);
+	if (st == DT_STATUS_DISABLED)
+		return -1;
+
+	pbase = _fdt_reg_base_address(fdt, offs);
+	if (pbase == DT_INFO_INVALID_REG)
+		return -1;
+	sz = _fdt_reg_size(fdt, offs);
+	if (sz == DT_INFO_INVALID_REG_SIZE)
+		return -1;
```
zjuzzy, line 146: The err_code(-1) returned by different exception branches. Maybe inconvenient for debug.

etienne-lms: Right. This is legacy and could move to `TEE_Result` return value.

etienne-lms (commented):
`dt_map_dev()` maps with ns=0 when `secure-status = "okay"` & `status = "disabled"`. @vesajaaskelainen, you don't want to rely on this scheme? You need an API function that enforces the secure state?

vesajaaskelainen:
> `dt_map_dev()` maps with ns=0 when `secure-status = "okay"` & `status = "disabled"`. @vesajaaskelainen, you don't want to rely on this scheme? You need an API function that enforces the secure state?

The problem here is that the real device tree would be:
```
status = "enabled"
secure-status = "enabled"
```

Eg. part of the registers are accessible for Linux kernel drivers and full set is available for OP-TEE driver.

If you do this the MMU mapping will be non-secure resulting non-secure access in FPGA side and we cannot access the TZ protected registers.

vesajaaskelainen:
Alternative for this function would be I suppose:
- invent new device tree keyword
- re-work logic work when both statuses are enabled so that actual mapping is secure. But don't know what that would break.

etienne-lms (commented):
> invent new device tree keyword

I don't think it is the preferred way...

> re-work logic when both statuses are enabled so that actual mapping is secure. But don't know what that would break.

That would be sowhat more consistent IMO. There are impacted drivers, we should get an Acked-by from them.

core/kernel/dt.c
jenswi-linaro: Sounds good to me.

vesajaaskelainen: I can try the proposed method.

core/include/kernel/dt.h
```diff
@@ -69,6 +69,18 @@ enum dt_driver_type {
...
 };
 
+/*
+ * DT_MAP_FROM_NODE_STATUS: Uses status properties from device tree to
+ * determine mapping.
+ * DT_MAP_SECURE: Force mapping for device to be secure.
+ * DT_MAP_NON_SECURE: Force mapping for device to be non-secure.
+ */
+enum dt_map_dev_directive {
+	DT_MAP_FROM_NODE_STATUS,
```
vesajaaskelainen, line 79: Was thinking should this be `DT_MAP_AUTO` or so.

vesajaaskelainen:
Now I have updated the PR based on comments.

Please check it out if it matches your ideas.

core/include/kernel/dt.h
jenswi-linaro: Yes, that's a good name.

jenswi-linaro:
Please change the prefix in the commit subject from kernel to core.

vesajaaskelainen:
> Please change the prefix in the commit subject from kernel to core.

OK

core/include/kernel/dt.h
vesajaaskelainen: I wait a bit if there are other comments and then go with this (needs a change in quite a many places :sweat:)

jforissier: FWIW... `DT_MAP_AUTO` sounds good to me too

vesajaaskelainen:
Updated the PR based on comments.

jenswi-linaro:

vesajaaskelainen:

Thanks for the review! Applied tag.

etienne-lms (commented):

vesajaaskelainen:
About failed checks. These passed nicely on my github fork's actions.

So some transient failure somewhere else?

jforissier:
> About failed checks. These passed nicely on my github fork's actions.
> 
> So some transient failure somewhere else?

Possible. Or perhaps there is a race condition somewhere -- `2002.1` is a multi-threaded TCP socket test. Unfortunately CI doesn't produce detailed logs and even with those it might be difficult to understand what happened. I have restarted the job.

PR #5606, opened 2022-10-24
A set of commits to start compiling for RISC-V

maroueneboubakri:
This PR adds commits populating required header files with minimalist definitions and some C files with stub functions to start compiling for RISC-V target. 
To build:
```
make ARCH=riscv PLATFORM=spike CROSS_COMPILE64=riscv64-linux-gnu-
```

This results a minimum OP-TEE OS core able to run on Spike and it prints to console.
To run:
```
$ spike -m0x40008000:0x1000,0x80000000:0x40000000 -p2 out/riscv-plat-spike/core/tee.elf
I/TC: OP-TEE version: 3.19.0-32-gcfc61d91 (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #1 Wed Nov  2 11:47:31 UTC 2022 riscv
I/TC: Primary CPU switching to REE boot
I/TC: Secondary CPU 1 initializing
I/TC: Secondary CPU 1 switching to REE boot
```


zjuzzy (approved):

core/arch/riscv/include/kernel/thread_arch.h
```diff
@@ -0,0 +1,134 @@
...
+
+/*
+ * These flags should vary according to the privilege mode selected
+ * to run OP-TEE core on (M/HS/S). For now default to S-Mode.
+ */
+#define THREAD_EXCP_FOREIGN_INTR	MIP_SEIP
+#define THREAD_EXCP_NATIVE_INTR		MIP_SSIP
+#define THREAD_EXCP_ALL			(THREAD_EXCP_FOREIGN_INTR	\
+					| THREAD_EXCP_NATIVE_INTR	\
+					| MIP_STIP)
```
etienne-lms, line 113: prefer:
```c
define THREAD_EXCP_ALL			(THREAD_EXCP_FOREIGN_INTR |	\
					 THREAD_EXCP_NATIVE_INTR |	\
					 MIP_STIP)

File: core/arch/riscv/include/mm/core_mmu_arch.h [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+
+static inline void core_mmu_table_write_barrier(void)
+{
+	/* Invoke memory barrier */
+	mb();
+}
+
+TEE_Result cache_op_inner(enum cache_op op, void *va, size_t len);
+
+static inline bool core_mmu_check_max_pa(paddr_t pa __maybe_unused)
```
etienne-lms, line 120: remove `__maybe_unused`

core/arch/riscv/include/mm/core_mmu_arch.h
```diff
@@ -0,0 +1,131 @@
...
+	return pa <= (BIT64(RISCV_MMU_PA_WIDTH) - 1);
+}
+
+static inline unsigned int core_mmu_get_va_width(void)
+{
+	return RISCV_MMU_VA_WIDTH;
+}
+#endif /*__ASSEMBLER__*/
+
+#endif /* CORE_MMU_H */
```
etienne-lms, line 131: `/*__CORE_MMU_ARCH_H*/` (btw, ditto in arm header file)

etienne-lms (commented):
With comments adressed, `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`.

jforissier:
> This results linking errors related to unimplemented symbols.

That's not great.

>  riscv: link.mk: linking parameters for OP-TEE core
> This recopies the make file from ARM architecture.

That's not great either.

We should try to avoid code duplication as much as possible. For example `get_core_pos()` could be moved to common code, leaving only `__get_core_pos()` arch- or plat-specific.

jenswi-linaro:
"riscv: link.mk: linking parameters for OP-TEE core" is much more complicated than needed since Risc-V doesn't use paging and doesn't need to be able to partition the binary into unpaged, init, and paged parts.

jenswi-linaro:
"riscv: link_dummy.ld: initial OP-TEE dummy core linking script" shouldn't be needed at all since Risc-V doesn't use paging.

maroueneboubakri:
> > This results linking errors related to unimplemented symbols.
> 
> That's not great.

@jforissier to avoid implementing everything in one PR with bunch of commits to review.
Should make linker ignore that for the moment to help building a minimalist core to run ?

> > riscv: link.mk: linking parameters for OP-TEE core
> > This recopies the make file from ARM architecture.
> 
> That's not great either.
> 
> We should try to avoid code duplication as much as possible. For example `get_core_pos()` could be moved to common code, leaving only `__get_core_pos()` arch- or plat-specific.

Agree

maroueneboubakri:
> "riscv: link.mk: linking parameters for OP-TEE core" is much more complicated than needed since Risc-V doesn't use paging and doesn't need to be able to partition the binary into unpaged, init, and paged parts.

@jenswi-linaro is it better to minimalise link.mk ? at some point will have a lot copied from the original link.mk, maybe move link.mk and make it generic ?

jforissier:
> > > This results linking errors related to unimplemented symbols.
> > 
> > 
> > That's not great.
> 
> @jforissier to avoid implementing everything in one PR with bunch of commits to review. Should make linker ignore that for the moment to help building a minimalist core to run ?

Up to a reasonable point I'd rather have a few more commits right away and no link errors. If that's too much code then it's OK to have stubs. A linker flag to ignore unresolved symbols is the least preferred way because it may hide truly unexpected errors (such as issues with some CFG_ combinations etc.).

jenswi-linaro:
> > "riscv: link.mk: linking parameters for OP-TEE core" is much more complicated than needed since Risc-V doesn't use paging and doesn't need to be able to partition the binary into unpaged, init, and paged parts.
> 
> @jenswi-linaro is it better to minimalise link.mk ? at some point will have a lot copied from the original link.mk, maybe move link.mk and make it generic ?

Yes, start with a minimal copy. Once we know more we can try to extract the generic parts from the arm link.mk.

maroueneboubakri:
@jenswi-linaro could consider the request mentioned here [PR#5164](https://github.com/OP-TEE/optee_os/pull/5164) to let arch implement its own `core_mmu_va2idx()`. This is yet mandatory.

jenswi-linaro:
@maroueneboubakri, I fail to see how the generic implementations of `core_mmu_va2idx()` and `core_mmu_idx2va()` don't work or are unsuitable for risc-v. These functions are designed to be usable with any page size (of power of 2) or level of table. It could be worth checking how the LPAE tables used in arch/arm compares with what arch/risc-v will use. From the little I've seen in risc-v specification I don't expect it to be too different.

maroueneboubakri:
> @maroueneboubakri, I fail to see how the generic implementations of `core_mmu_va2idx()` and `core_mmu_idx2va()` don't work or are unsuitable for risc-v. These functions are designed to be usable with any page size (of power of 2) or level of table. It could be worth checking how the LPAE tables used in arch/arm compares with what arch/risc-v will use. From the little I've seen in risc-v specification I don't expect it to be too different.

@jenswi-linaro In RISCV-V we can have several MMU modes, Sv32, Sv39, Sv48, Sv57 and probably Sv64. Ech mode has different number of possible page levels, however the number of bits representing the page index in a level is always 9. Therefore each level has a table of 512 entries. In all, the page size is always multiple of 2 not an issue.

Given the the MMU mode is Sv39, the virtual address is represented as following:

```
38    30 29   21 20   12 11         0
+-------+-------+-------+-----------+
| VPN[2]|VPN[1] |VPN[0] |Page offset|
+-------+-------+-------+-----------+
   9       9       9         12
```
Given a virtual address, To get the index in first level (level 0).  Is (va >> 12) & 0x1ff.  This is the difference.
0x1ff is the 9-bits mask 0b1_1111_1111 (9 bits).

We still can keep the function as it is and mask the the result of `core_mmu_va2idx()` when needed but tracking this seems complicated.

jenswi-linaro:
> We still can keep the function as it is and mask the the result of `core_mmu_va2idx()` when needed but tracking this seems complicated.

What's more complicated is to invent a new scheme to do the same thing. There's no need to mask the result of `core_mmu_va2idx()`, if there is then something is wrong.

jenswi-linaro:
Please let us know when you're done pushing commits so we can start reviewing.

maroueneboubakri:
> Please let us know when you're done pushing commits so we can start reviewing.

@jforissier @etienne-lms  @jenswi-linaro  Kindly review:

```
riscv: include: kernel: add tee_l2cc_mutex.h
riscv: core: define CFG_MAX_CACHE_LINE_SHIFT in riscv.mk
riscv: include: add cache_helpers.h
riscv: include: add thread_arch.h
riscv: include: add thread_private_arch.h
riscv: include: add core_mmu_arch.h
riscv: kernel: boot.c: implement generic boot functions
riscv: kernel: main.c: implement tee_otp_get_hw_unique_key()
riscv: kernel: idle.c: implement cpu_idle()
riscv: kernel: tee_time.c: implement tee_time_get_sys_time()
riscv: kernel: trace_ext.c: implement tracer functions 
riscv: kernel: asm-defines.c: defines for a minimalist core 
riscv: kernel: mentry.S: boot optee_os in machine mode
riscv: plat-spike: conf.mk: set CFG_TEE_RAM_VA_SIZE to 4Mb
riscv: mm: tlb_helpers_rv.S: translation look-aside buffer invalidate 
riscv: plat-spike: conf.mk: set CFG_TEE_CORE_LOG_LEVEL to default
```

core/arch/riscv/include/kernel/thread_arch.h
```diff
@@ -0,0 +1,142 @@
...
+#include <compiler.h>
+#include <riscv.h>
+#include <types_ext.h>
+#endif
+
+#ifndef __ASSEMBLER__
+
+#define THREAD_CORE_LOCAL_ALIGNED __aligned(2 * RISCV_XLEN_BYTES)
+
+struct thread_pauth_keys {
```
jenswi-linaro, line 19: Is this used? I see that you add it in `struct thread_core_local` below, but is it used from there?

core/arch/riscv/include/kernel/thread_arch.h
```diff
@@ -0,0 +1,142 @@
...
+	uint64_t apia_hi;
+	uint64_t apia_lo;
+};
+
+struct thread_core_local {
+	unsigned long x[4];
```
jenswi-linaro, line 25: Is this used?

jenswi-linaro (commented):
Comments for "riscv: include: add thread_arch.h"

core/arch/riscv/include/kernel/thread_private_arch.h
```diff
@@ -0,0 +1,71 @@
...
+#define STACK_THREAD_SIZE	(8192 + CFG_STACK_THREAD_EXTRA)
+
+#if TRACE_LEVEL > 0
+#define STACK_ABT_SIZE		3072
+#else
+#define STACK_ABT_SIZE		1024
+#endif
+
+#ifdef CFG_CORE_DEBUG_CHECK_STACKS
+#define STACK_CHECK_EXTRA	1536
```
jenswi-linaro, line 30: It seems that one needs to read the Arm code in order to understand the purpose of this. If you don't use this keep it as `0` for simplicity.

jenswi-linaro (commented):
Comment for "riscv: include: add thread_private_arch.h"

core/arch/riscv/include/mm/core_mmu_arch.h
```diff
@@ -0,0 +1,132 @@
...
+#define RISCV_MMU_PA_WIDTH	U(32)
+#define RISCV_MMU_VA_WIDTH	U(32)
+#else
+#error unknown or unsupported mmu mode
+#endif
+
+#define RISCV_PTES_PER_PT (U(1) << RISCV_PGLEVEL_BITS)
+#define RISCV_PGLEVELS ((RISCV_MMU_VA_WIDTH - RISCV_PGSHIFT) \
+			/ RISCV_PGLEVEL_BITS)
+#define RISCV_MMU_VPN_MASK	((U(1) << RISCV_PGLEVEL_BITS) - 1)
```
jenswi-linaro, line 69: Please use the `BIT()` macro, or even better `GENMASK_32()`.

core/arch/riscv/include/mm/core_mmu_arch.h
```diff
@@ -0,0 +1,132 @@
...
+/*
+ * Level 0: shift = 12 4K pages
+ * Level 1: shift = 21 2M pages
+ * Level 2: shift = 30 1G pages
+ */
+
+#define CORE_MMU_USER_CODE_SHIFT	SMALL_PAGE_SHIFT
+#define CORE_MMU_USER_PARAM_SHIFT	SMALL_PAGE_SHIFT
+
+#define CORE_MMU_PGDIR_LEVEL	U(0)
```
jenswi-linaro, line 82: Strange indentation for the values here.

core/arch/riscv/include/mm/core_mmu_arch.h
```diff
@@ -0,0 +1,132 @@
...
+#define RISCV_SATP_ASID_SIZE	U(9)
+#define RISCV_SATP_ASID_MASK	0x01FF
+#define RISCV_MMU_PA_WIDTH	U(32)
+#define RISCV_MMU_VA_WIDTH	U(32)
+#else
+#error unknown or unsupported mmu mode
+#endif
+
+#define RISCV_PTES_PER_PT (U(1) << RISCV_PGLEVEL_BITS)
+#define RISCV_PGLEVELS ((RISCV_MMU_VA_WIDTH - RISCV_PGSHIFT) \
```
jenswi-linaro, line 67: I'd rather have this as:
``` C
#define RISCV_PGLEVELS          ((RISCV_MMU_VA_WIDTH - RISCV_PGSHIFT) / \
                                 RISCV_PGLEVEL_BITS)
```
Note that the operator is at the end of a line instead of the beginning.

core/arch/riscv/include/mm/core_mmu_arch.h
```diff
@@ -0,0 +1,132 @@
...
+
+#define CORE_MMU_USER_CODE_SHIFT	SMALL_PAGE_SHIFT
+#define CORE_MMU_USER_PARAM_SHIFT	SMALL_PAGE_SHIFT
+
+#define CORE_MMU_PGDIR_LEVEL	U(0)
+#define CORE_MMU_PGDIR_SHIFT U(12)
+
+#define CORE_MMU_BASE_TABLE_LEVEL	0
+#define CORE_MMU_BASE_TABLE_SHIFT	(RISCV_PGLEVEL_BITS * \
+		CORE_MMU_BASE_TABLE_LEVEL + RISCV_PGSHIFT)
```
jenswi-linaro, line 87: This could be indented better.

core/arch/riscv/include/mm/core_mmu_arch.h
```diff
@@ -0,0 +1,132 @@
...
+	DCACHE_AREA_INVALIDATE,
+	ICACHE_INVALIDATE,
+	ICACHE_AREA_INVALIDATE,
+	DCACHE_CLEAN_INV,
+	DCACHE_AREA_CLEAN_INV,
+};
+
+static inline void core_mmu_table_write_barrier(void)
+{
+	/* Invoke memory barrier */
```
jenswi-linaro, line 115: This comment doesn't help much, but I'm missing the comment explaining the purpose of this function. In `arch/arm` it's a very specific instruction so it might be worth explaining what's expected of the barrier in case someone decides to try to optimize this.

jenswi-linaro (commented):
Comments for "riscv: include: add core_mmu_arch.h"

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,130 @@
...
+
+#define PADDR_INVALID               ULONG_MAX
+
+uint32_t sem_cpu_sync[CFG_TEE_CORE_NB_CORE];
+
+static void init_runtime(unsigned long pageable_part __unused)
+{
+}
+
+void init_sec_mon(unsigned long nsec_entry __maybe_unused)
```
jenswi-linaro, line 32: Do you need this?

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,130 @@
...
+	 * Mask asynchronous exceptions before switch to the thread vector
+	 * as the thread handler requires those to be masked while
+	 * executing with the temporary stack. The thread subsystem also
+	 * asserts that the foreign interrupts are blocked when using most of
+	 * its functions.
+	 */
+	thread_set_exceptions(THREAD_EXCP_ALL);
+
+	/*
+	 * Pager: init_runtime() calls thread_kernel_enable_vfp() so we must
```
jenswi-linaro, line 50: Do you expect that arch/riscv ever will need to implement a pager?
I find it highly unlikely, better to keep things simple and assume that it will never happen.

jenswi-linaro (commented):
For commit "riscv: kernel: boot.c: implement generic boot functions":
You mention a security monitor in the commit message. Will we ever have such a thing? If so, what will it do? Anything at all like the Armv7 monitor?

core/arch/riscv/kernel/trace_ext.c
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
jenswi-linaro, line 1: This file looks identical to the arch/arm version of the same. We should move `core/arch/arm/kernel/trace_ext.c` to `core/kernel/trace_ext.c` instead.

jenswi-linaro (commented):
Comment for "riscv: kernel: trace_ext.c: implement tracer functions"

core/arch/riscv/kernel/asm-defines.c
```diff
@@ -0,0 +1,47 @@
...
+	       offsetof(struct thread_core_local, curr_thread));
+	DEFINE(THREAD_CORE_LOCAL_FLAGS,
+	       offsetof(struct thread_core_local, flags));
+	DEFINE(THREAD_CORE_LOCAL_ABT_STACK_VA_END,
+	       offsetof(struct thread_core_local, abt_stack_va_end));
+
+	DEFINE(STACK_TMP_GUARD, STACK_CANARY_SIZE / 2 + STACK_TMP_OFFS);
+
+	/* struct thread_ctx_regs */
+	DEFINE(RISCV_CTX_REG_STATUS, offsetof(struct thread_ctx_regs, status));
```
jenswi-linaro, line 29: All the `RISCV_CTX_REG_*` seem to be unused.

jenswi-linaro (commented):
Comments for "riscv: kernel: asm-defines.c: defines for a minimalist core"

core/arch/riscv/kernel/mentry.S
```diff
@@ -0,0 +1,204 @@
...
+	lla	t0, __bss_start
+	lla	t1, __bss_end
+	beq	t0, t1, 1f
+0:
+	STR	zero, (t0)
+	add	t0, t0, RISCV_XLEN_BYTES
+	bne	t0, t1, 0b
+1:
+
+	csrw satp, 0
```
jenswi-linaro, line 145: Please add indent the argument for the instruction using a tab character. Same below with the `jal` instruction.

core/arch/riscv/kernel/mentry.S
```diff
@@ -0,0 +1,204 @@
...
+	set_sp
+	set_tp
+	cpu_is_ready
+
+	jal boot_init_secondary
+	j	.
+END_FUNC reset_secondary
+
+LOCAL_FUNC unhandled_cpu , :
+	wfi
```
jenswi-linaro, line 180: No need for a `j unhandled_cpu` after this instruction?

core/arch/riscv/kernel/mentry.S
```diff
@@ -0,0 +1,204 @@
...
+	bge	t0, t1, unhandled_cpu
+	la	t2, stack_tmp
+	lw	t1, stack_tmp_stride
+	/*
+	 * t0 is core pos + 1
+	 * t1 is value of stack_tmp_stride
+	 * t2 is value of stack_tmp
+	 */
+	mv	sp, t2
+	/* Don't use mul instruction because not all harts support it. */
```
jenswi-linaro, line 37: Could the multiplication be implemented using a macro instead, at least if there's a chance we'll need multiplications in assembly at other places too?

jenswi-linaro (commented):
Comments for "riscv: kernel: mentry.S: boot optee_os in machine mode"

jenswi-linaro (commented):
One more comment for "riscv: kernel: mentry.S: boot optee_os in machine mode":
In the description, you mention machine code, but this is assembly code.

Edit: nevermind, my bad, you said machine mode.

jenswi-linaro:
For commit "riscv: include: kernel: add tee_l2cc_mutex.h"
Commit description isn't correct, it's core/mm/core_mmu.c and core/mm/vm.c
which needs this header file.
With that fixed:
Acked-by: Jens Wiklander <jens.wiklander@linaro.org>

For the commits:
- "riscv: core: define CFG_MAX_CACHE_LINE_SHIFT in riscv.mk"
- "riscv: include: add cache_helpers.h"
- "riscv: kernel: main.c: implement tee_otp_get_hw_unique_key()"
- "riscv: kernel: idle.c: implement cpu_idle()"
- "riscv: kernel: tee_time.c: implement tee_time_get_sys_time()"
- "riscv: plat-spike: conf.mk: set CFG_TEE_RAM_VA_SIZE to 4Mb"
- "riscv: mm: tlb_helpers_rv.S: translation look-aside buffer invalidate "
- "riscv: plat-spike: conf.mk: set CFG_TEE_CORE_LOG_LEVEL to default"
Please apply:
Acked-by: Jens Wiklander <jens.wiklander@linaro.org>

core/arch/riscv/include/kernel/thread_arch.h
maroueneboubakri: `thread_resume() `is not part of this commitset yet. But referenced [here ](https://github.com/OP-TEE/optee_os/pull/5606/commits/73ca858cdfd469ed36deb81f771469b2a36f2515#diff-bcc8474c5ac865bd696bc2b4329c46f8616c5d60ad1a712f94def838e291f8afR135).

core/arch/riscv/include/mm/core_mmu_arch.h
maroueneboubakri: Comment just added to to satisfy checkpatch, since it is used by `core/mm/core_mmu.c `we implemented it using fence. Maybe this is not helpful as answer.

core/arch/riscv/kernel/boot.c
maroueneboubakri: Will decide it later, if it to use OpenSBI with security monitor extension or to implement security monitor in OP-TEE.

core/arch/riscv/kernel/mentry.S
maroueneboubakri: Which instruction ?

core/arch/riscv/include/kernel/thread_arch.h
jenswi-linaro: Will that ever be used? I find it somewhat unlikely that pointer authentication is implemented in the same way in Riscv as in Arm.

core/arch/riscv/kernel/mentry.S
jenswi-linaro: `wfi`

core/arch/riscv/include/kernel/thread_arch.h
maroueneboubakri: Better to remove them for now.

core/arch/riscv/kernel/boot.c
maroueneboubakri: I mean the current implementation we have uses it.  But not sure to keep it in the end, as there are many approches for security monitor. The one we implement now is based on OpenSBI extension, but those are not specifications yet.

jenswi-linaro:
Are these last two commits:
- riscv: core: riscv.mk: select privilege mode of OP-TEE core
- riscv: kernel: sbi.c: add RISC-V SBI interface 

necessary at this stage? I mean, things compile without them. I'd prefer to not grow the scope at this stage.

By the way, there are a few warnings:
```
core/arch/riscv/kernel/boot.c: In function ‘init_primary’:
core/arch/riscv/kernel/boot.c:34:40: warning: unused parameter ‘pageable_part’ [-Wunused-parameter]
   34 | static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
      |                          ~~~~~~~~~~~~~~^~~~~~~~~~~~~
core/arch/riscv/kernel/boot.c: At top level:
cc1: note: unrecognized command-line option ‘-Wno-c2x-extensions’ may have been intended to silence earlier diagnostics
core/arch/riscv/kernel/thread.c:313:33: warning: ‘struct thread_pauth_keys’ declared inside parameter list will not be visible outside of this definition or declaration
  313 |                          struct thread_pauth_keys *keys __unused)
      |                                 ^~~~~~~~~~~~~~~~~
core/arch/riscv/kernel/thread.c:388:13: warning: ‘thread_rpc_free’ defined but not used [-Wunused-function]
  388 | static void thread_rpc_free(unsigned int bt __unused, uint64_t cookie __unused,
      |             ^~~~~~~~~~~~~~~
core/arch/riscv/kernel/thread.c:182:22: warning: ‘ftrace_suspend’ defined but not used [-Wunused-function]
  182 | static void __noprof ftrace_suspend(void)
      |                      ^~~~~~~~~~~~~~
cc1: note: unrecognized command-line option ‘-Wno-c2x-extensions’ may have been intended to silence earlier diagnostics
core/arch/riscv/mm/core_mmu_arch.c: In function ‘core_mmu_entry_to_finer_grained’:
core/arch/riscv/mm/core_mmu_arch.c:317:66: warning: unused parameter ‘tbl_info’ [-Wunused-parameter]
  317 | l core_mmu_entry_to_finer_grained(struct core_mmu_table_info *tbl_info,
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~

core/arch/riscv/mm/core_mmu_arch.c:318:51: warning: unused parameter ‘idx’ [-Wunused-parameter]
  318 |                                      unsigned int idx, bool secure __maybe_unused)
      |                                      ~~~~~~~~~~~~~^~~
core/arch/riscv/mm/core_mmu_arch.c: At top level:
cc1: note: unrecognized command-line option ‘-Wno-c2x-extensions’ may have been intended to silence earlier diagnostics
```

The note about "-Wno-c2x-extensions" can be ignored, it goes away when there are no warnings in the code.

jenswi-linaro:
OK, please let us know when you've fixed everything for a final review of the remaining commits.

maroueneboubakri:
@jenswi-linaro kindly review the following commits:

```
core: kernel: move trace_ext.c to core/kernel 
riscv: kernel: spinlock.S: make __cpu_spin_trylock visible 
riscv: core: riscv.mk: select privilege mode of OP-TEE core 
riscv: kernel: sbi: add RISC-V SBI interface 
riscv: kernel: add console driver for S-Mode 
riscv: include: riscv.h: define generic CSR registers
```
 Those are the last commits in this PR, before working the others like:
```
riscv: kernel: implement thread management routines 
riscv: mm: core_mmu_arch.c: stub functions for memory management unit
```

Regards
Maro

core/arch/arm/kernel/sub.mk
```diff
@@ -37,7 +37,6 @@ ifeq ($(CFG_WITH_VFP),y)
 srcs-$(CFG_ARM32_core) += vfp_a32.S
```
jforissier, line 37: For commit "core: kernel: move trace_ext.c to core/kernel":

core/arch/riscv/kernel/spinlock.S
```diff
@@ -33,7 +33,7 @@ FUNC __cpu_spin_unlock , :
 END_FUNC __cpu_spin_unlock
 
 /* unsigned int __cpu_spin_trylock(unsigned int *lock) */
-LOCAL_FUNC __cpu_spin_trylock , :
+FUNC __cpu_spin_trylock , :
```
jforissier, line 36: For commit "riscv: kernel: spinlock.S: make __cpu_spin_trylock visible":

core/arch/riscv/riscv.mk
```diff
@@ -35,6 +35,18 @@ endif
 
 CFG_MAX_CACHE_LINE_SHIFT ?= 6
 
+ifeq ($(CFG_RISCV_S_MODE),y)
+$(info Info: Builing core for S-Mode)
+CFG_RISCV_M_MODE := n
```
jforissier, line 40: $(call force,CFG_RISCV_M_MODE,n) and same below for `CFG_RISCV_S_MODE`.

That's in commit "riscv: core: riscv.mk: select privilege mode of OP-TEE core". With that fixed, the commit is:

core/arch/riscv/kernel/sbi.c
```diff
@@ -0,0 +1,45 @@
...
+void sbi_console_putchar(int ch)
+{
+	sbi_ecall(SBI_EXT_0_1_CONSOLE_PUTCHAR, (unsigned long)ch);
+}
+
+int sbi_boot_hart(uint32_t hart_id, paddr_t start_addr, unsigned long arg)
+{
+	struct sbiret ret;
+
+	ret	= sbi_ecall(SBI_EXT_HSM, SBI_EXT_HSM_HART_START, hart_id, start_addr, arg);
```
jforissier, line 42: Use space not tab before `=`

Commit "riscv: kernel: sbi: add RISC-V SBI interface". With my comments fixed:

core/arch/riscv/kernel/console.c
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
jforissier, line 1: Commit "riscv: kernel: add console driver for S-Mode":

core/arch/riscv/include/riscv.h
```diff
@@ -25,6 +25,27 @@
...
 #endif
 
+#if defined(CFG_RISCV_M_MODE)
+#define CSR_MODE_OFFSET	PRV_M
+#define XRET			mret
+#elif defined(CFG_RISCV_S_MODE)
+#define CSR_MODE_OFFSET	PRV_S
+#define XRET			sret
+#else
+#error CFG_RISCV_M_MODE or CFG_RISCV_S_MODE should be defined
```
jforissier, line 35: Commit " riscv: include: riscv.h: define generic CSR registers"
You do not need the error message here, it is already checked in the `.mk` file.

jforissier (commented):
Hello @maroueneboubakri, a few more comments from me. Thanks.

core/arch/riscv/riscv.mk
maroueneboubakri: This will allow selecting the mode from command line ? `make ... CFG_RISCV_S_MODE=y`
The default is building for M-Mode, but Spike supports (M/U or M/S/U), or should modify the config in case we want to build for S-Mode ?

jforissier: > This will allow selecting the mode from command line ?

Yes. As I understand it, `CFG_RISCV_S_MODE` and `CFG_RISCV_M_MODE` are mutually exclusive. So if one is enabled (`=y`), then the other must be disabled (either undefined, empty or `=n`). By using `$(call force,...)` we make sure that such an inconsistent configuration is detected. For example `make CFG_RISCV_S_MODE=y CFG_RISCV_M_MODE=y` will report an error. With your proposed code (`:=`), it would report nothing and `CFG_RISCV_M_MODE` would still be `y` due to the higher priority of the command line over `:=`.

There is a very simple rule for `CFG_` in OP-TEE Makefiles: you set them either with `?=` or with `$(call force...)`. Any other way is a mistake. (That's only for `CFG_` ,other variables may be a bit different).

See: https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#configuration-variables (we should perhaps add the above rule more clearly there...)

maroueneboubakri: Yes understood, the issue comes here:

```
$ make ARCH=riscv PLATFORM=spike CROSS_COMPILE64=riscv64-linux-gnu-  CFG_RISCV_S_MODE=y
Info: Builing core for S-Mode
core/arch/riscv/riscv.mk:40: *** CFG_RISCV_M_MODE is set to 'y' (from file) but its value must be 'n'.  Stop.
```
In platform conf.mk:
```
CFG_RISCV_M_MODE ?= y
```
Command line supposed to have more priority than file ?

jforissier: Sorry it is indeed a bit more complicated because the two variables are related. Please try the following:
```
CFG_RISCV_M_MODE ?= y
ifeq ($(CFG_RISCV_M_MODE),y)
ifeq ($(CFG_RISCV_S_MODE),y)
$(error CFG_RISCV_M_MODE and CFG_RISCV_S_MODE cannot be both 'y')
else
$(call force,CFG_RISCV_S_MODE,n)
endif
endif
ifeq ($(CFG_RISCV_S_MODE),y)
$(call force,CFG_RISCV_M_MODE,n)
endif
ifneq (y,$(call cfg-one-enabled,CFG_RISCV_M_MODE M CFG_RISCV_S_MODE))
$(error Either CFG_RISCV_M_MODE or CFG_RISCV_S_MODE must be 'y')
endif
```

With this, the command to build M mode is `make` and for S mode `make CFG_RISCV_S_MODE=y CFG_RISCV_M_MODE=n`. And hopefully invalid combinations should be caught too.

jforissier:
For "core: include: misc.h: divide into misc.h and misc_arch.h":

core/arch/riscv/include/kernel/misc.h
```diff
@@ -0,0 +1,28 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
jforissier, line 1: This file should be removed from commit "riscv: include: add misc_arch.h". Once done that commit is:

jforissier (commented):
For commit "core: include: cache_helpers.h: allow reusing architecture-dependent code":

core/arch/riscv/plat-spike/conf.mk
```diff
@@ -73,3 +73,4 @@ CFG_TDDRAM_SIZE  ?= 0x00f00000
 CFG_SHMEM_START  ?= 0x7fe00000
 CFG_SHMEM_SIZE   ?= 0x00200000
 CFG_TEE_RAM_VA_SIZE ?= 0x00400000
+CFG_RISCV_M_MODE ?= y
```
jforissier, line 76: Something I missed in "riscv: core: riscv.mk: select privilege mode of OP-TEE core"

Can Spike work in S mode?
If it can, this line is not needed because it is the default set in `riscv.mk`
If it cannot then this line should be:
```
$(call force,CFG_RISCV_M_MODE,y)
```
(otherwise someone could do `make M=n S=y` and the checks in `riscv.mk` would not complain, not knowing that this config is not supported by the platform)

jforissier:
```
 riscv: plat-spike: conf.mk: set CFG_TEE_RAM_VA_SIZE to 4MB

Set CFG_TEE_RAM_VA_SIZE to 0x00400000 in core/arch/riscv/plat-spike/conf.mk

Acked-by: Jens Wiklander <jens.wiklander@linaro.org>
cked-by: Jerome Forissier <jerome.forissier@linaro.org>
```
^--- missing A

Please fix the checkpatch warnings too. Most of them are legitimate it seems.

core/arch/riscv/plat-spike/conf.mk
maroueneboubakri: Yes, Spike supports all privilege levels.

jenswi-linaro:
For commit "core: include: misc.h: divide into misc.h and misc_arch.h"
It seems that `core/arch/arm/include/kernel/misc.h` is still there,
I was expecting that it would be removed with this.

For commit "riscv: kernel: spinlock.S: make __cpu_spin_trylock visible"
It's cpu_spin_trylock() that needs __cpu_spin_trylock(). Please
use () after function names.

For commits:
- "core: include: cache_helpers.h: allow reusing architecture-dependent code"
- "core: kernel: move trace_ext.c to core/kernel"

Please apply:

For commits:
- "riscv: include: riscv.h: define generic CSR registers"
- "riscv: kernel: add console driver for S-Mode"
- "riscv: kernel: sbi: add RISC-V SBI interface"
- "riscv: core: riscv.mk: select privilege mode of OP-TEE core"

Please apply:

jforissier:
@maroueneboubakri please apply Jens' tags, then I will merge as many reviewed commits as possible. The remaining ones can be handled in a new PR.

maroueneboubakri:
@jforissier done.

maroueneboubakri:
Added slight changes to allow deciding to use SBI or not, same for console driver using legacy SBI extension.
This will allow using different uart driver instead of legacy SBI calls.

jforissier:
@maroueneboubakri I have picked all the commits having a review/ack tag and applied them to master. You should be able to cherry-pick the remaining ones with only trivial conflicts in the `sub.mk` files. Note, I have fixed the Author field in several commits to use the same name/email as in the `Reviewed-by:`. Thanks!

maroueneboubakri:
> @maroueneboubakri I have picked all the commits having a review/ack tag and applied them to master. You should be able to cherry-pick the remaining ones with only trivial conflicts in the `sub.mk` files. Note, I have fixed the Author field in several commits to use the same name/email as in the `Reviewed-by:`. Thanks!

Thanks @jforissier !

maroueneboubakri:
@jforissier @jenswi-linaro , Kindly review the following commits

```
riscv: plat-virt: add qemu generic virtual platform 
riscv: core: implement platform-level interrupt controller driver 
riscv: core: implement core local interruptor (clint) driver 
core: kernel: move tee_time.c and tee_time_ree.c to core/kernel 
riscv: kernel: add time source based on time registers 
```

Thanks
Maro

maroueneboubakri:
@jforissier @jenswi-linaro , Kindly review the following commits

```
riscv: plat-virt: add qemu generic virtual platform 
riscv: core: implement platform-level interrupt controller driver 
riscv: core: implement core local interruptor (clint) driver 
core: kernel: move tee_time.c and tee_time_ree.c to core/kernel 
riscv: kernel: add time source based on time registers 
```

Thanks
Maro

jforissier:
> Why are you not starting a new PR with new patches instead? It seems the
> way of working has become that when Jerome think you have too many reviewed
> patches he just merges manually them to get them out of the way. That's not
> how we work normally

Agreed, and I indeed asked for a new PR in https://github.com/OP-TEE/optee_os/pull/5606#issuecomment-1326152226.

jenswi-linaro:
> > You could perhaps skip the `riscv` part
> > of these paths, but I believe it might be better to have them.
> 
> Not a strong opinion but I would prefer the opposite (not have `riscv`). Linux for example doesn't have arch sub-directories under drivers and I believe the reason is some hardware may very well be used by different CPUs.

Fair enough, let's skip the `riscv` part then.

maroueneboubakri:
Reopened new PRs:

- [PR#5701](https://github.com/OP-TEE/optee_os/pull/5701)
- [PR#5714](https://github.com/OP-TEE/optee_os/pull/5714)
- [PR#5716](https://github.com/OP-TEE/optee_os/pull/5716)
- [PR#5718](https://github.com/OP-TEE/optee_os/pull/5718)

And close this.

PR #5607, opened 2022-10-25
ci: code_style: run pycodestyle even if checkpatch has failed

jforissier:
When checkpatch reports issues, it returns an error status which make the code_style job fail as expected. However, by default when a job step fails the subsequent ones are not executed. Therefore, pycodestyle is skipped which is bad because we sometimes want to ignore some checkpatch errors and we still want to detect Python issues.

This problem is fixed by adding a condition to the "Run pycodestyle" step.


jenswi-linaro:

PR #5621, opened 2022-11-02
core:Add support for Hisilicon D06 (PLATFORM=d06)

xiaoxuZeng:
D06 is a server-class development board equipped with a Hisilicon Phosphor processor.

Signed-off-by: Zengxx <zengxiaoxu@huawei.com>


jenswi-linaro:
I notice that we have a `plat-d02` which I assume relates to this. Would it make sense to group these two under a new `plat-???` and use `PLATFORM_FLAVOR=d06` instead?

jforissier:
> I notice that we have a `plat-d02` which I assume relates to this. Would it make sense to group these two under a new `plat-???` and use `PLATFORM_FLAVOR=d06` instead?

As far as I can tell, `d02` is pretty much obsolete now. I am listed as the maintainer, but I don't have access to that board anymore. 

D06 could go under `plat-hisilicon` (`PLATFORM=hisilicon-d06` sounds good) but in practice it doesn't make much sense since what is already there (`hi3519av100_demo`) is totally unrelated to D06. There is no code to share. In fact I think it using manufacturer names in `plat-*` is a bad idea, it is too broad.

So all in all I don't mind `PLATFORM=d06`. We could use `hisilicon_d06` if it sounds too generic.

jenswi-linaro:
OK, let's keep it as it is then.

core/arch/arm/plat-d06/core_pos_a64.S
```diff
@@ -0,0 +1,48 @@
...
+
+	mov     x5, x4
+#if (CFG_TEE_CORE_NB_CORE == 96)
+	lsl     x5, x5, 1
+	add     x5, x5, x4
+	lsl     x5, x5, 1
+#elif (CFG_TEE_CORE_NB_CORE == 128)
+	lsl     x5, x5, 3
+#else
+	COMPILE_TIME_ASSERT(0);
```
jforissier, line 39: `static_assert(0);` is preferred since 4516e38db4.

core/arch/arm/plat-d06/platform_config.h
```diff
@@ -0,0 +1,43 @@
...
+ *    TA RAM: 32 MiB                             |
+ *  0x2600_0000                                  | TZDRAM
+ *    TEE RAM: 32 MiB (TEE_RAM_VA_SIZE)		 |
+ *  0x2400_0000 [TZDRAM_BASE, TEE_LOAD_ADDR]     -
+ *    Shared memory: 64 MB                       | SHMEM
+ *  0x2000_0000                                  -
+ *    Linux/other                                | DRAM0
+ *  0x0000_0000 [DRAM0_BASE]                     -
+ *
+ */
```
jforissier, line 41: It looks this is copied from D02. The values don't match what is in `conf.mk`. I suggest removing the whole block altogether.

jforissier:
Also @zengxx-HUAWEI feel free to add yourself to `MAINTAINERS` if you wish.

zengxx-HUAWEI:
> Also @zengxx-HUAWEI feel free to add yourself to `MAINTAINERS` if you wish.

Thanks, I will submit an application to the maintainer list.

xiaoxuZeng:
> @zengxx-HUAWEI please see my comments below. Also please add a space in the commit subject:
> 
> ```
> core:Add support for Hisilicon D06 (PLATFORM=d06)
>      ^here
> ```
> 
> Then you can add: `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`
> 
> Thanks!

done, Thanks.

xiaoxuZeng:
> OK, let's keep it as it is then. `Acked-by: Jens Wiklander <jens.wiklander@linaro.org>`

done, Thanks.

MAINTAINERS
```diff
@@ -113,6 +113,11 @@ HiSilicon D02
 S:	Orphan
 F:	core/arch/arm/plat-d02/
 
+HiSilicon D06
+R:	zengxx <zengxiaoxu@huawei.com> [@xiaoxuZeng]
```
jforissier, line 117: We expect full names as "Firstname Lastname", that is: `Xiaoxu Zeng` rather than `zengxx` if you don't mind. Same in `Signed-off-by:` and git Author.

jforissier (commented):
One more comment then:

MAINTAINERS
xiaoxuZeng: Nice advice. thank you.

wgzh1987 (approved):

PR #5627, opened 2022-11-07
SCMI server from SCP-firmware

etienne-lms:
The P-R allows a platform to build an SCMI server stack to run in OP-TEE core. The SCMI server is built from SCP-firmware reference  implementation https://github.com/ARM-software/SCP-firmware.git which source are to be fetch externally to optee_os and path set from `CFG_SCP_FIRMWARE`. 

This P-R implements an SCMI for platforms fvp and stm32mp this is embedded when `CFG_SCMI_SERVER=y`.
P-Rs https://github.com/OP-TEE/manifest/pull/220 and https://github.com/OP-TEE/build/pull/600 allows one to fetch and build stm32mp1 with SCP-firmware SCMI server.

### DIFF
--- a/core/arch/arm/plat-vexpress/conf.mk
+++ b/core/arch/arm/plat-vexpress/conf.mk
@@ -69,6 +69,11 @@ CFG_SHMEM_SIZE   ?= 0x00200000
 # DRAM1 is defined above 4G
 $(call force,CFG_CORE_LARGE_PHYS_ADDR,y)
 $(call force,CFG_CORE_ARM64_PA_BITS,36)
+ifeq ($(CFG_SCMI_SERVER),y)
+$(call force,CFG_SCMI_SERVER_PRODUCT,optee-fvp)
+$(call force,CFG_SCMI_SERVER_SMT_HEADER,n)
+$(call force,CFG_SCMI_SERVER_MSG_HEADER,y)
+endif
 endif
 
 ifeq ($(PLATFORM_FLAVOR),juno)
@@ -125,6 +130,11 @@ CFG_SHMEM_SIZE  ?= 0x00200000
 CFG_TEE_SDP_MEM_SIZE ?= 0x00400000
 $(call force,CFG_DT,y)
 CFG_DTB_MAX_SIZE ?= 0x100000
+ifeq ($(CFG_SCMI_SERVER),y)
+$(call force,CFG_SCMI_SERVER_PRODUCT,optee-fvp)
+$(call force,CFG_SCMI_SERVER_SMT_HEADER,n)
+$(call force,CFG_SCMI_SERVER_MSG_HEADER,y)
+endif
 endif
 
 ifneq (,$(filter $(PLATFORM_FLAVOR),qemu_virt qemu_armv8a))
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -757,6 +757,13 @@ endif
 # configuration mandates target product identifier is configured with
 # CFG_SCMI_SERVER_PRODUCT and the SCP-firmware source tree path with
 # CFG_SCP_FIRMWARE.
+#
+# There are several protocols supported for exchanging SCMI message through
+# shared memory.
+# CFG_SCMI_SERVER_SMT_HEADER, when set, indicates to SCMI PTA that SCMI SMT
+# protocol based statically allocated IOMEM shared memory is supported.
+# CFG_SCMI_SERVER_MSG_HEADER, when set, indicates to SCMI PTA that SCMI MSG_SMT
+# protocol based generic dynamially allocated OP-TEE shared memory is supported.
 CFG_SCMI_SERVER ?= n
 
 ifeq ($(CFG_SCMI_SERVER),y)
@@ -767,6 +774,8 @@ endif
 ifeq (,$(wildcard $(CFG_SCP_FIRMWARE)/CMakeLists.txt))
 $(error CFG_SCMI_SERVER=y mandates CFG_SCP_FIRMWARE configuration)
 endif
+CFG_SCMI_SERVER_SMT_HEADER ?= y
+CFG_SCMI_SERVER_MSG_HEADER ?= y
 endif #CFG_SCMI_SERVER
 
 ifeq ($(CFG_SCMI_MSG_DRIVERS)-$(CFG_SCMI_SERVER),y-y)
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -752,6 +752,27 @@ $(eval $(call cfg-depends-all,CFG_SCMI_MSG_SMT_INTERRUPT_ENTRY,CFG_SCMI_MSG_SMT)
 $(eval $(call cfg-depends-one,CFG_SCMI_MSG_SMT_THREAD_ENTRY,CFG_SCMI_MSG_SMT CFG_SCMI_MSG_SHM_MSG))
 endif
 
+# CFG_SCMI_SERVER, when enabled, embeds the reference SCMI server implementation
+# from SCP-firmware package as an built-in SCMI stack in core. This
+# configuration mandates target product identifier is configured with
+# CFG_SCMI_SERVER_PRODUCT and the SCP-firmware source tree path with
+# CFG_SCP_FIRMWARE.
+CFG_SCMI_SERVER ?= n
+
+ifeq ($(CFG_SCMI_SERVER),y)
+$(call force,CFG_SCMI_PTA,y,Required by CFG_SCMI_SERVER)
+ifeq (,$(CFG_SCMI_SERVER_PRODUCT))
+$(error CFG_SCMI_SERVER=y mandates CFG_SCMI_SERVER_PRODUCT configuration)
+endif
+ifeq (,$(wildcard $(CFG_SCP_FIRMWARE)/CMakeLists.txt))
+$(error CFG_SCMI_SERVER=y mandates CFG_SCP_FIRMWARE configuration)
+endif
+endif #CFG_SCMI_SERVER
+
+ifeq ($(CFG_SCMI_MSG_DRIVERS)-$(CFG_SCMI_SERVER),y-y)
+$(error CFG_SCMI_MSG_DRIVERS=y and CFG_SCMI_SERVER=y are mutually exclusive)
+endif
+
 # Enable SCMI PTA interface for REE SCMI agents
 CFG_SCMI_PTA ?= n
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,9 +81,11 @@ jobs:
 
           function _make() { make -j$(nproc) -s O=out $*; }
           function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.4.0 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
+          function download_scp_firmware() { git clone --single-branch -b add-optee-arch https://github.com/etienne-lms/SCP-firmware.git $HOME/scp-firmware || echo Nervermind; }
 
           ccache -s -v
           download_plug_and_trust
+          download_scp_firmware
 
           _make
           _make COMPILER=clang
@@ -130,6 +132,7 @@ jobs:
           _make PLATFORM=stm32mp1
           _make PLATFORM=stm32mp1-135F_DK
           _make PLATFORM=stm32mp1-157C_DK2
+          if [ -d $HOME/scp-firmware ]; then _make PLATFORM=stm32mp1-157C_DK2 CFG_SCMI_SERVER=y CFG_SCP_FIRMWARE=$HOME/scp-firmware; fi
           _make PLATFORM=vexpress-fvp
           _make PLATFORM=vexpress-fvp CFG_ARM64_core=y
           _make PLATFORM=vexpress-fvp CFG_ARM64_core=y CFG_CORE_SEL1_SPMC=y CFG_SECURE_PARTITION=y
--- a/core/arch/arm/include/scmi/scmi_server.h
+++ b/core/arch/arm/include/scmi/scmi_server.h
@@ -1,0 +1,75 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2019-2022, Linaro Limited
+ */
+#ifndef SCMI_SERVER_H
+#define SCMI_SERVER_H
+
+#include <tee_api_types.h>
+#include <types_ext.h>
+
+#ifdef CFG_SCMI_SERVER
+/*
+ * Request processing of an incoming event in the SCMI server for a target
+ * MHU/SMT mailbox.
+ * The function is exported by scmi-server.
+ *
+ * @channel_id: SCMI channel handler
+ */
+TEE_Result scmi_server_smt_process_thread(unsigned int channel_id);
+
+/*
+ * Request processing of an incoming event in the SCMI server for a target
+ * MHU/MSG mailbox.
+ * The function is exported by scmi-server.
+ *
+ * @id: SCMI channel handler
+ * @in_buf: Input message MSG buffer
+ * @in_size: Input message MSG buffer size
+ * @out_buf: Output message MSG buffer
+ * @out_size: Reference to output message MSG buffer size
+ */
+TEE_Result scmi_server_msg_process_thread(unsigned int channel_id, void *in_buf,
+					  size_t in_size, void *out_buf,
+					  size_t *out_size);
+
+/*
+ * Get SCP-firmware channel device ID from the client channel ID.
+ *
+ * @channel_id: SCMI channel handler
+ * @handle: Output SCP-firmware device ID for the target SCMI mailbox
+ */
+TEE_Result scmi_server_get_channel(unsigned int channel_id, int *handle);
+
+/* Get number of channels supported by the SCMI platform/server */
+int scmi_server_get_channels_count(void);
+
+#else /* CFG_SCMI_SERVER */
+static inline
+TEE_Result scmi_server_smt_process_thread(unsigned int channel_id __unused)
+{
+	return TEE_ERROR_NOT_SUPPORTED;
+}
+
+static inline
+TEE_Result scmi_server_msg_process_thread(unsigned int channel_id __unused,
+					  void *in_buf __unused,
+					  size_t in_size __unused,
+					  void *out_buf __unused,
+					  size_t *out_size __unused)
+{
+	return TEE_ERROR_NOT_SUPPORTED;
+}
+
+static inline TEE_Result scmi_server_get_channel(unsigned int id __unused,
+						 int *handle __unused)
+{
+	return TEE_ERROR_NOT_SUPPORTED;
+}
+
+static inline int scmi_server_get_channels_count(void)
+{
+	return 0;
+}
+#endif /* CFG_SCMI_SERVER */
+#endif /* SCMI_SERVER_H */
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -190,6 +190,11 @@ CFG_WDT ?= $(CFG_STM32_IWDG)
 # Platform specific configuration
 CFG_STM32MP_PANIC_ON_TZC_PERM_VIOLATION ?= y
 
+# Default enable scmi-msg server if SCP-firmware SCMI server is disabled
+ifneq ($(CFG_SCMI_SERVER),y)
+CFG_SCMI_MSG_DRIVERS ?= y
+endif
+
 # SiP/OEM service for non-secure world
 CFG_STM32_BSEC_SIP ?= y
 CFG_STM32MP1_SCMI_SIP ?= n
@@ -202,10 +207,18 @@ endif
 # Default enable SCMI PTA support
 CFG_SCMI_PTA ?= y
 ifeq ($(CFG_SCMI_PTA),y)
+ifneq ($(CFG_SCMI_SERVER),y)
 $(call force,CFG_SCMI_MSG_DRIVERS,y,Mandated by CFG_SCMI_PTA)
 $(call force,CFG_SCMI_MSG_SMT_THREAD_ENTRY,y,Mandated by CFG_SCMI_PTA)
 CFG_SCMI_MSG_SHM_MSG ?= y
 CFG_SCMI_MSG_SMT ?= y
+endif # !CFG_SCMI_SERVER
+endif # CFG_SCMI_PTA
+
+CFG_SCMI_SERVER ?= n
+ifeq ($(CFG_SCMI_SERVER),y)
+$(call force,CFG_SCMI_SERVER_PRODUCT,optee-stm32mp1)
+cflags-plat-scmi-server-$(CFG_STPMIC1) += -DCFG_STPMIC1
 endif
 
 CFG_SCMI_MSG_DRIVERS ?= n
--- a/core/arch/arm/plat-stm32mp1/main.c
+++ b/core/arch/arm/plat-stm32mp1/main.c
@@ -14,6 +14,7 @@
 #include <drivers/stm32_uart.h>
 #include <drivers/stm32mp1_etzpc.h>
 #include <drivers/stm32mp_dt_bindings.h>
+#include <io.h>
 #include <kernel/boot.h>
 #include <kernel/dt.h>
 #include <kernel/interrupt.h>
@@ -24,6 +25,7 @@
 #include <platform_config.h>
 #include <sm/psci.h>
 #include <stm32_util.h>
+#include <string.h>
 #include <trace.h>
 
 register_phys_mem_pgdir(MEM_AREA_IO_NSEC, APB1_BASE, APB1_SIZE);
@@ -44,6 +46,11 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, APB6_BASE, APB6_SIZE);
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, AHB4_BASE, AHB4_SIZE);
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, AHB5_BASE, AHB5_SIZE);
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, GIC_BASE, GIC_SIZE);
+
+#ifdef CFG_STM32MP1_SCMI_SHM_BASE
+register_phys_mem(MEM_AREA_IO_NSEC, CFG_STM32MP1_SCMI_SHM_BASE,
+		  CFG_STM32MP1_SCMI_SHM_SIZE);
+#endif
 
 register_ddr(DDR_BASE, CFG_DRAM_SIZE);
 
@@ -307,6 +314,18 @@ static TEE_Result init_stm32mp1_drivers(void)
 
 	etzpc_configure_tzma(1, SYSRAM_SEC_SIZE >> SMALL_PAGE_SHIFT);
 	etzpc_lock_tzma(1);
+
+	if (SYSRAM_SIZE > SYSRAM_SEC_SIZE) {
+		size_t nsec_size = SYSRAM_SIZE - SYSRAM_SEC_SIZE;
+		paddr_t nsec_start = SYSRAM_BASE + SYSRAM_SEC_SIZE;
+		uint8_t *va = phys_to_virt(nsec_start, MEM_AREA_IO_NSEC,
+					   nsec_size);
+
+		IMSG("Non-secure SYSRAM [%p %p]", va, va + nsec_size - 1);
+
+		/* Clear content from the non-secure part */
+		memset(va, 0, nsec_size);
+	}
 
 	return TEE_SUCCESS;
 }
--- a/core/core.mk
+++ b/core/core.mk
@@ -146,6 +146,12 @@ libname = unw
 libdir = lib/libunw
 include mk/lib.mk
 
+ifeq ($(CFG_SCMI_SERVER),y)
+libname = scmi-server
+libdir = core/lib/scmi-server
+include mk/lib.mk
+endif
+
 #
 # Do main source
 #
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -301,3 +301,12 @@ out:
 
 	return res;
 }
+
+TEE_Result clk_get_rates_array(struct clk *clk, size_t start_index,
+			       unsigned long *rates, size_t *nb_elts)
+{
+	if (!clk->ops->get_rates_array)
+		return TEE_ERROR_NOT_SUPPORTED;
+
+	return clk->ops->get_rates_array(clk, start_index, rates, nb_elts);
+}
--- a/core/include/drivers/clk.h
+++ b/core/include/drivers/clk.h
@@ -48,6 +48,7 @@ struct clk {
  * @get_parent: Get the current parent index of the clock
  * @set_rate: Set the clock rate
  * @get_rate: Get the clock rate
+ * @get_rates_array: Get the supported clock rates as array
  */
 struct clk_ops {
 	TEE_Result (*enable)(struct clk *clk);
@@ -58,6 +59,8 @@ struct clk_ops {
 			       unsigned long parent_rate);
 	unsigned long (*get_rate)(struct clk *clk,
 				  unsigned long parent_rate);
+	TEE_Result (*get_rates_array)(struct clk *clk, size_t start_index,
+				      unsigned long *rates, size_t *nb_elts);
 };
 
 /**
@@ -177,5 +180,18 @@ struct clk *clk_get_parent_by_index(struct clk *clk, size_t pidx);
  * Return a TEE_Result compliant value
  */
 TEE_Result clk_set_parent(struct clk *clk, struct clk *parent);
+
+/**
+ * clk_get_rates_array - Get supported rates as an array
+ *
+ * @clk: Clock for which the rates are requested
+ * @start_index: start index of requested rates
+ * @rates: Array of rates allocated by caller or NULL to query count of rates
+ * @nb_elts: Max number of elements that the array can hold as input. Contains
+ * the number of elements that was added in the array as output.
+ * Returns a TEE_Result compliant value
+ */
+TEE_Result clk_get_rates_array(struct clk *clk, size_t start_index,
+			       unsigned long *rates, size_t *nb_elts);
 
 #endif /* __DRIVERS_CLK_H */
--- a/core/lib/scmi-server/include/optee_scmi.h
+++ b/core/lib/scmi-server/include/optee_scmi.h
@@ -1,0 +1,21 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2019-2022, Linaro Limited
+ */
+#ifndef OPTEE_SCMI_H
+#define OPTEE_SCMI_H
+
+#include <tee_api_types.h>
+#include <types_ext.h>
+
+/*
+ * Return virtual address mapped for target SMT IOMEM address range
+ *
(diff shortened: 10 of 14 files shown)


core/scmi-server/sub.mk
```diff
@@ -0,0 +1,45 @@
...
+scmi-server-flags-y += -DDISABLE_CPPCHECK=1
+scmi-server-flags-y += -DCFG_NUM_THREADS=$(CFG_NUM_THREADS)
+scmi-server-flags-y += -DSCP_OPTEE_DIR:PATH=$(CURDIR)
+
+scmi-server-flags-y += -DCFG_CROSS_COMPILE=$(lastword $(CROSS_COMPILE_core))
+
+CFLAGS_OPTEE = $(cflags$(sm))
+CFLAGS_OPTEE += $(subst -include $(conf-file),,$(cflags$(sm)))
+CFLAGS_OPTEE += $(cflags-plat-scmi-server-y)
+scmi-server-flags-y += -DCFG_CFLAGS_OPTEE="$(CFLAGS_OPTEE)"
```
jenswi-linaro, line 34: `$(cflags$(sm))` can change even after this `sub.mk` has been processed.
Have you verified that `$(CFLAGS_OPTEE)` is updated as needed if/when that happens?

mk/config.mk
```diff
@@ -742,6 +742,27 @@ $(eval $(call cfg-depends-all,CFG_SCMI_MSG_SMT_INTERRUPT_ENTRY,CFG_SCMI_MSG_SMT)
...
+ifeq (,$(CFG_SCMI_SERVER_PRODUCT))
+$(error CFG_SCMI_SERVER=y mandates CFG_SCMI_SERVER_PRODUCT configuration)
+endif
+ifeq (,$(wildcard $(CFG_SCP_FIRMWARE)/CMakeLists.txt))
+$(error CFG_SCMI_SERVER=y mandates CFG_SCP_FIRMWARE configuration)
+endif
+endif #CFG_SCMI_SERVER
+
+ifeq ($(CFG_SCMI_MSG_DRIVERS)-$(CFG_SCMI_SERVER),y-y)
+$(error CFG_SCMI_MSG_DRIVERS=y and CFG_SCMI_SERVER=y are exclusive)
```
jenswi-linaro, line 763: mutually exclusive

jenswi-linaro (commented):
Comments for "core: scmi_server: Build a SCMI server from SCP-firmware"
In the commit message:
`s/axternal/external/`

core/arch/arm/plat-vexpress/conf.mk
```diff
@@ -55,6 +55,11 @@ endif
 CFG_WITH_STATS ?= y
 CFG_ENABLE_EMBEDDED_TESTS ?= y
 
+ifeq ($(CFG_SCMI_SERVER),y)
+CFG_SCMI_SERVER_PRODUCT=fvp
```
jenswi-linaro, line 59: In the commit "plat-stm32mp1: support building with CFG_SCMI_SERVER=y" you're using force instead, any particular reason?

jenswi-linaro:
For commit "drivers: clk: add support for clk_get_rates_array", please remove the `Change-Id:` tag.

core/scmi-server/scmi_server.c
```diff
@@ -0,0 +1,86 @@
...
+
+TEE_Result scmi_server_get_channel(unsigned int channel_id, int *handle)
+{
+	int fwk_id = 0;
+
+	fwk_id = scmi_get_device(channel_id);
+	if (fwk_id < 0)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (handle)
```
xiaoxuZeng, line 41: In the case `handle == NULL` I think it would be better to return `TEE_ERROR_BAD_PARAMETERS`. :)
Maybe:
```c
if (handle == NULL)

*handle = fwk_id;
```

core/scmi-server/sub.mk
```diff
@@ -0,0 +1,45 @@
...
+
+define build-SCP-firmware-cmake
+$(scp-firmware-output): FORCE
```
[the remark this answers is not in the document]
etienne-lms: right. I think we can remove this.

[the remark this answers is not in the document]
etienne-lms: We need `FORCE`. We need to run CMake on SCP-firmware to know if something needs to be rebuilt, optee_ios as no visibility on whether its source file tree content was modified in some ways.

core/arch/arm/plat-vexpress/conf.mk
etienne-lms: no, but here I should have used `$(call force,...)` or `?=`. I'll fix, using `$(call force,...)`.

core/scmi-server/sub.mk
etienne-lms: Latest change do evaluate `$(cflagscore)` when target is executed, taking into account changes to its value after this _sub.mk_ file is parsed.

jenswi-linaro: Thanks for checking.

core/scmi-server/scmi_server.c
xiaoxuZeng: > Actaully there is a call to `scmi_server_get_channel()` from the SCMI PTA that only checks the channel ID is valid but does not retrieve the SCP-fmw related handle, [here in current patch](https://github.com/OP-TEE/optee_os/pull/5627/commits/8b96706fe7a5d7ed0033304fa3a5c5aec6997985#diff-b02cf9d13860a7b8eb818540e5776a98cba205eecc11519571d540cac8b5fa61R193-R196).

That‘s ok

xiaoxuZeng (approved):
lgtm

core/scmi-server/sub.mk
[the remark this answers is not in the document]
jenswi-linaro: Fair enough, I see that you at least got rid of the `rm -rf`

etienne-lms:
@xiaoxuZeng, can I add your Acked-by or Reviewed-by tag to the commits?
To do so I would need your tag ID [`Full Name <mail-address>`](https://optee.readthedocs.io/en/latest/general/contribute.html#developer-certificate-of-origin).

xiaoxuZeng:
> @xiaoxuZeng, can I add your Acked-by or Reviewed-by tag to the commits? To do so I would need your tag ID [`Full Name <mail-address>`](https://optee.readthedocs.io/en/latest/general/contribute.html#developer-certificate-of-origin).
Sure

Reviewed-by: Xiaoxu Zeng <zengxiaoxu@huawei.com>

thanks.

etienne-lms:
Tag appllied + fix include order:
```patch
--- a/core/scmi-server/scmi_server.c
+++ b/core/scmi-server/scmi_server.c
@@ -8,5 +8,5 @@
 #include <initcall.h>
 #include <kernel/panic.h>
 #include <mm/core_memprot.h>
+#include <optee_scmi.h>
 #include <scmi/scmi_server.h>
-#include <optee_scmi.h>

**Comment by etienne-lms** [comment]
rebased

**Comment by etienne-lms** [comment]
I've updated _ci.yml_ but it requires CMake. Can you add it in your optee_os_ci image?
CI messsage:
```
if [ -d $HOME/scp-firmware ]; then _make PLATFORM=stm32mp1-157C_DK2 CFG_SCMI_SERVER=y CFG_SCP_FIRMWARE=$HOME/scp-firmware; fi
bash: line 1: cmake: command not found
make: *** [core/scmi-server/sub.mk:28: out/core/scmi-server/build/product/optee-stm32mp1/fw/libscmi-fw-all.a] Error 127
Error: Process completed with exit code 2.

jforissier:
> I've updated _ci.yml_ but it requires CMake. Can you add it in your optee_os_ci image?

Done. For commit "ci: build stm32mp15 with scp-firmware scmi-server":

etienne-lms:
Rebased to trigger a new CI round.

etienne-lms:
CI build failed. Some missing glibc things:
```
fwk_id.c:(.text.__fwk_id_str+0x72): undefined reference to `__snprintf_chk'
/usr/bin/arm-linux-gnueabihf-ld.bfd: fwk_id.c:(.text.__fwk_id_str+0x12e): undefined reference to `__snprintf_chk'
fwk_io.c:(.text.fwk_io_vprintf+0x30): undefined reference to `__vsnprintf_chk'
```
Browsing a bit this could be related to use of musl in the doccker image.

jforissier:
> Browsing a bit this could be related to use of musl in the doccker image.

Does it build successfully for you? Can you give the full command line?
CI doesn't use the GCC compiler built by Arm (the one we get via `make toolchains` in build git), it might be the reason. That being said, which libc the compiler provides should not make a difference since OP-TEE has its own (libutils).

jforissier:
It looks like the C flags for the SCP firmware are missing `-nostdinc`. Therefore the compiler uses the headers for the default C library instead of the OP-TEE ones (from libutils). Refer to the objdump output of `fwk_log.c.obj`:
```
fwk_log_vsnprintf():
/tmp/SCP-firmware/framework/src/fwk_log.c:124
  46:   9701            str     r7, [sp, #4]
snprintf():
/usr/arm-linux-gnueabihf/include/bits/stdio2.h:71
  48:   f7ff fffe       bl      0 <__snprintf_chk>
                        48: R_ARM_THM_CALL      __snprintf_chk
  4c:   682a            ldr     r2, [r5, #0]
  4e:   4605            mov     r5, r0
```

The file `/usr/arm-linux-gnueabihf/include/bits/stdio2.h` should not be used. Instead ``snprintf()` should be found in `lib/libutils/isoc/include/stdio.h`. Since libutils provides no implementation for `__snprintf_chk()` there is a link error.

etienne-lms:
> It looks like the C flags for the SCP firmware are missing -nostdinc. (...)

Many thanks for the analysis and to the OP-TEE CI that issue was reported.

I've have changed _scmi-server/sub.mk_ to set core nostdinc and libutils include paths to CFLAGS directives passed through CMake.

Changing SCP-firmware to build with nostdinc, i faced 2 issues.
First _assert.h_ and _stdlib.h_ header files from SCP-fmw and optee_os are using the same `ASSERT_H` / `STDLIB_H` marco name for header guards. I've added an RFC commit to this P-R describing the issue and a change proposal in our header files.
Second is `aligned_alloc()` needed by SCP-firmware. As is it a standard function, the easiest way to address its declaration and implementation is to implement it in libutils. See related RFC commit.

I've updated _ci.yml_ in the series to point to my SCP-firmare branch to haved needed changes in that repo and check CI results. I'll sync with @vingu-linaro to update P-R https://github.com/ARM-software/SCP-firmware/pull/664 with related changes.

lib/libutils/isoc/include/stdlib.h
```diff
@@ -15,6 +15,8 @@
 #include <stddef.h>
 #include <malloc.h>
 
+/* aligend_alloc() is declared in included malloc.h */
```
jforissier, line 18: I would prefer to keep things as close to the standard as possible, that is: declare `aligned_alloc()` in `<stdlib.h>` not in `<malloc.h>` and use the proper C11 guard:
```C
#if __STDC_VERSION__ >= 201112L
void *aligned_alloc(size_t alignment, size_t size);
#endif
```

[Upstream newlib](https://github.com/bminor/newlib/blob/master/newlib/libc/include/stdlib.h#L331) uses `__ISO_C_VISIBLE` which seems to be the BSD libc way of doing things, I think it makes more sense to use `__STDC_VERSION__ defined by the compiler.

jforissier (changes requested):
Reviewing "[RFC] libutils: implement aligned_alloc()"

jforissier:
Commit "[RFC] libutils: prefix system header guard for assert.h and stdlib.h":

LGTM, but it means all our libc (libutils) header should probably do the same... we can leave it until another conflict appears though.

etienne-lms:
I can apply the change to all std header from libutils. I also think it would be more consistent.

jforissier:
> I can apply the change to all std header from libutils. I also think it would be more consistent.

Fine by me.

etienne-lms:
Is `#if __STDC_VERSION__ >= 201112L` really needed to protect `aligned_alloc()` declaration? We can implement to function whatever the toolchain used, no?
Moreover I saw libutils build for ldelf fails when we have these guards.

etienne-lms:
I have added fixup commits to move `aligned_alloc()` declaration to _stdlib.h_ (without guards)
and to add prefix `_` (single underscore char) to standard header files (not optee specific _trace.h_, _util.h_, etc...

jforissier:
> Is `#if __STDC_VERSION__ >= 201112L` really needed to protect `aligned_alloc()` declaration? We can implement to function whatever the toolchain used, no? Moreover I saw libutils build for ldelf fails when we have these guards.

It is a dangerous path to take IMO. What if some external code depends on this (which is a standard) to provide its own implementation? Then we would have a conflict.

etienne-lms:
We build with `-std=gnu99`, not c11. Does that mean we can't implement `aligned_alloc()` in libutils?

jforissier:
I suppose we can easily switch to `-std=c11` or `std=gnu11`.

etienne-lms:
I've sqaushed the fixup commits and rebased the series since `aligned_alloc()` is merged.
I've added 3 new fixup commits related to discussion with @vingu-linaro.

@xiaoxuZeng and @jforissier, can you check these 3 added changes. I'll apply your review tag upon your feedback.

core/scmi-server/sub.mk
```diff
@@ -14,6 +14,7 @@ incdirs_ext-y += $(scmi-server-scp-path)/arch/none/optee/include
 scp-firmware-output = $(scmi-server-out-path)/build/product/optee-$(scmi-server-product)/fw/libscmi-fw-all.a
 
 libdeps += $(scp-firmware-output)
+libdeps += out/arm/core-lib/libutils/libutils.a
```
jforissier, line 17: You don't want to hard-code the output directory here! It is the reason of the CI failure, which uses a custom output directory ([link](https://github.com/OP-TEE/optee_os/actions/runs/3542522124/jobs/5948110592#step:5:183)).

The fix should be simple enough though. I suggest to define `lib-libfile-$(libname)-$(sm)` similarly to the `-shlibfile-` version which already exists:
```diff
diff --git a/mk/lib.mk b/mk/lib.mk
index 30f2040a3..65b575ec5 100644
--- a/mk/lib.mk
+++ b/mk/lib.mk
@@ -27,6 +27,7 @@ endif
 include mk/compile.mk

 lib-libfile    := $(out-dir)/$(base-prefix)$(libdir)/lib$(libname).a
+lib-libfile-$(libname)-$(sm) := $(lib-libfile)
 ifeq ($(CFG_ULIBS_SHARED),y)
 lib-shlibfile  := $(out-dir)/$(base-prefix)$(libdir)/lib$(libname).so
 lib-shlibstrippedfile := $(out-dir)/$(base-prefix)$(libdir)/lib$(libname).stripped.so
```

Then you should be able to use: `libdeps += $(lib-libfile-libutils-core)` here.

jforissier (commented):
@etienne-lms one comment, once addressed LGTM.

core/scmi-server/sub.mk
etienne-lms: > EDIT: But why is this dependency even needed? libscmi-fw-all.a is an archive, it should not need to be linked against libutils.

Actually... you're right, we don't need it. Libutils is already in the system deps, as the scim-server lib we're building. I'll crosscheck with @vingu-linaro, maybe related to a specific build env.

vingu-linaro: It's not about linking with archive but the order of the lib when optee_os is linked into the final binary.
Without this dependency, libutils.a is before libscmi-fw-all.a some functions of libutils are removed because not used at this point. With this, we ensure that libscmi-fw-all.a is before and add a ref to libutils.a function

jenswi-linaro: > It's not about linking with archive but the order of the lib when optee_os is linked into the final binary. Without this dependency, libutils.a is before libscmi-fw-all.a some functions of libutils are removed because not used at this point. With this, we ensure that libscmi-fw-all.a is before and add a ref to libutils.a function

That is surprising, `libutils.a` is usually the last lib (expect `libgcc.a`) on the command line.

jforissier: > It's not about linking with archive but the order of the lib when optee_os is linked into the final binary.

Right, I was a bit confused by the name (`libdeps`), as if the library was needed to produce the SCP .a. It's not -- very well.

> That is surprising, `libutils.a` is usually the last lib (except `libgcc.a`) on the command line.

Correct, this is supposed to be achieved by:
```
mk/lib.mk:libdeps               := $(lib-libfile) $(libdeps)
```
...and the fact that `libutils` is listed first in `core.mk`:

```
# Do libraries
...
libname = utils
libdir = lib/libutils
include mk/lib.mk 
...
ifeq ($(CFG_SCMI_SERVER),y)
libname = scmi-server
libdir = core/scmi-server
include mk/lib.mk
endif
```

vingu-linaro: without this line i have : 
/usr/bin/ccache /home/vingu/Linaro/SCMI/optee-project-qemu-latest/build/../toolchains/aarch64/bin/aarch64-linux-gnu-ld.bfd -T core/arch/arm/kernel/link_dummy.ld --no-check-sections  out/arm/core/arch/arm/kernel/ldelf_loader.o out/arm/core/arch/arm/kernel/tee_time.o out/arm/core/arch/arm/kernel/rpc_io_i2c.o out/arm/core/arch/arm/kernel/otp_stubs.o out/arm/core/arch/arm/kernel/delay.o out/arm/core/arch/arm/kernel/idle.o out/arm/core/arch/arm/kernel/tee_time_arm_cntpct.o out/arm/core/arch/arm/kernel/timer_a64.o out/arm/core/arch/arm/kernel/spin_lock_a64.o out/arm/core/arch/arm/kernel/tlb_helpers_a64.o out/arm/core/arch/arm/kernel/cache_helpers_a64.o out/arm/core/arch/arm/kernel/thread_a64.o out/arm/core/arch/arm/kernel/thread.o out/arm/core/arch/arm/kernel/thread_optee_smc.o out/arm/core/arch/arm/kernel/thread_optee_smc_a64.o out/arm/core/arch/arm/kernel/abort.o out/arm/core/arch/arm/kernel/vfp.o out/arm/core/arch/arm/kernel/vfp_a64.o out/arm/core/arch/arm/kernel/trace_ext.o out/arm/core/arch/arm/kernel/misc_a64.o out/arm/core/arch/arm/kernel/boot.o out/arm/core/arch/arm/kernel/entry_a64.o out/arm/core/arch/arm/kernel/unwind_arm64.o out/arm/core/arch/arm/mm/core_mmu.o out/arm/core/arch/arm/mm/core_mmu_lpae.o out/arm/core/arch/arm/mm/pgt_cache.o out/arm/core/arch/arm/mm/mobj_dyn_shm.o out/arm/core/arch/arm/tee/arch_svc_a64.o out/arm/core/arch/arm/tee/arch_svc.o out/arm/core/arch/arm/tee/entry_fast.o out/arm/core/arch/arm/tee/cache.o out/arm/core/arch/arm/plat-vexpress/main.o out/arm/core/arch/arm/plat-vexpress/vendor_props.o out/arm/core/crypto/crypto.o out/arm/core/crypto/aes-gcm.o out/arm/core/crypto/aes-gcm-sw.o out/arm/core/crypto/aes-gcm-ghash-tbl.o out/arm/core/crypto/signed_hdr.o out/arm/core/crypto/rng_fortuna.o out/arm/core/crypto/cbc-mac.o out/arm/core/crypto/aes-cts.o out/arm/core/crypto/sm2-kdf.o out/arm/core/crypto/sm3.o out/arm/core/crypto/sm3-hash.o out/arm/core/crypto/sm3-hmac.o out/arm/core/crypto/sm4.o out/arm/core/crypto/sm4-ecb.o out/arm/core/crypto/sm4-cbc.o out/arm/core/crypto/sm4-ctr.o out/arm/core/drivers/pl011.o out/arm/core/drivers/gic.o out/arm/core/drivers/clk/clk.o out/arm/core/drivers/clk/clk_dt.o out/arm/core/drivers/clk/fixed_clk.o out/arm/core/drivers/rstctrl/rstctrl.o out/arm/core/kernel/spin_lock_debug.o out/arm/core/kernel/assert.o out/arm/core/kernel/console.o out/arm/core/kernel/dt.o out/arm/core/kernel/dt_driver.o out/arm/core/kernel/dt_driver_test.o out/arm/core/kernel/pm.o out/arm/core/kernel/handle.o out/arm/core/kernel/interrupt.o out/arm/core/kernel/ldelf_syscalls.o out/arm/core/kernel/msg_param.o out/arm/core/kernel/panic.o out/arm/core/kernel/refcount.o out/arm/core/kernel/tee_misc.o out/arm/core/kernel/tee_ta_manager.o out/arm/core/kernel/ts_manager.o out/arm/core/kernel/scattered_array.o out/arm/core/kernel/huk_subkey.o out/arm/core/kernel/user_mode_ctx.o out/arm/core/kernel/initcall.o out/arm/core/kernel/user_access.o out/arm/core/kernel/mutex.o out/arm/core/kernel/wait_queue.o out/arm/core/kernel/notif.o out/arm/core/kernel/thread.o out/arm/core/kernel/user_ta.o out/arm/core/kernel/ree_fs_ta.o out/arm/core/kernel/early_ta.o out/arm/core/kernel/secstor_ta.o out/arm/core/kernel/embedded_ts.o out/arm/core/kernel/pseudo_ta.o out/arm/core/mm/mobj.o out/arm/core/mm/fobj.o out/arm/core/mm/file.o out/arm/core/mm/vm.o out/arm/core/mm/core_mmu.o out/arm/core/mm/tee_mm.o out/arm/core/pta/device.o out/arm/core/pta/secstor_ta_mgmt.o out/arm/core/pta/stats.o out/arm/core/pta/system.o out/arm/core/pta/scmi.o out/arm/core/pta/tests/fs_htree.o out/arm/core/pta/tests/invoke.o out/arm/core/pta/tests/misc.o out/arm/core/pta/tests/mutex.o out/arm/core/pta/tests/aes_perf.o out/arm/core/tee/entry_std.o out/arm/core/tee/tee_cryp_utl.o out/arm/core/tee/tee_cryp_hkdf.o out/arm/core/tee/tee_cryp_concat_kdf.o out/arm/core/tee/tee_cryp_pbkdf2.o out/arm/core/tee/tee_obj.o out/arm/core/tee/tee_svc.o out/arm/core/tee/tee_svc_cryp.o out/arm/core/tee/tee_svc_storage.o out/arm/core/tee/tee_time_generic.o out/arm/core/tee/tadb.o out/arm/core/tee/socket.o out/arm/core/tee/tee_ta_enc_manager.o out/arm/core/tee/tee_fs_key_manager.o out/arm/core/tee/tee_ree_fs.o out/arm/core/tee/fs_dirfile.o out/arm/core/tee/fs_htree.o out/arm/core/tee/tee_fs_rpc.o out/arm/core/tee/tee_pobj.o out/arm/core/tee/uuid.o out/arm/core/tee/tee_supp_plugin_rpc.o out/arm/core/ta_pub_key.o out/arm/core/ldelf_hex.o out/arm/core/early_ta_f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c.o out/arm/core/arch/arm/dts/embedded_dtb_test.o   out/arm/core/scmi-server/libscmi-server.a out/arm/lib/libunw/libunw.a out/arm/core/lib/zlib/libzlib.a out/arm/core/lib/libfdt/libfdt.a out/arm/core/lib/libtomcrypt/libtomcrypt.a out/arm/core-lib/libmbedtls/libmbedtls.a out/arm/core-lib/libutils/libutils.a  out/arm/core/scmi-server/build/product/optee-fvp/fw/libscmi-fw-all.a /home/vingu/Linaro/SCMI/optee-project-qemu-latest/toolchains/aarch64/bin/../lib/gcc/aarch64-none-linux-gnu/10.2.1/libgcc.a -o out/arm/core/all_objs.o
/home/vingu/Linaro/SCMI/optee-project-qemu-latest/build/../toolchains/aarch64/bin/aarch64-linux-gnu-ld.bfd: /home/vingu/Linaro/SCMI/optee-project-qemu-latest/optee_os/out/arm/core/scmi-server/build/framework/libframework.a(fwk_string.c.obj): in function `fwk_str_strncpy':
/home/vingu/Linaro/SCMI/optee-project-qemu-latest/SCP-firmware/framework/src/fwk_string.c:52: undefined reference to `strncpy'
core/arch/arm/kernel/link.mk:68: recipe for target 'out/arm/core/all_objs.o' failed

With the line it becomes: 

/usr/bin/ccache /home/vingu/Linaro/SCMI/optee-project-qemu-latest/build/../toolchains/aarch64/bin/aarch64-linux-gnu-ld.bfd -T core/arch/arm/kernel/link_dummy.ld --no-check-sections  out/arm/core/arch/arm/kernel/ldelf_loader.o out/arm/core/arch/arm/kernel/tee_time.o out/arm/core/arch/arm/kernel/rpc_io_i2c.o out/arm/core/arch/arm/kernel/otp_stubs.o out/arm/core/arch/arm/kernel/delay.o out/arm/core/arch/arm/kernel/idle.o out/arm/core/arch/arm/kernel/tee_time_arm_cntpct.o out/arm/core/arch/arm/kernel/timer_a64.o out/arm/core/arch/arm/kernel/spin_lock_a64.o out/arm/core/arch/arm/kernel/tlb_helpers_a64.o out/arm/core/arch/arm/kernel/cache_helpers_a64.o out/arm/core/arch/arm/kernel/thread_a64.o out/arm/core/arch/arm/kernel/thread.o out/arm/core/arch/arm/kernel/thread_optee_smc.o out/arm/core/arch/arm/kernel/thread_optee_smc_a64.o out/arm/core/arch/arm/kernel/abort.o out/arm/core/arch/arm/kernel/vfp.o out/arm/core/arch/arm/kernel/vfp_a64.o out/arm/core/arch/arm/kernel/trace_ext.o out/arm/core/arch/arm/kernel/misc_a64.o out/arm/core/arch/arm/kernel/boot.o out/arm/core/arch/arm/kernel/entry_a64.o out/arm/core/arch/arm/kernel/unwind_arm64.o out/arm/core/arch/arm/mm/core_mmu.o out/arm/core/arch/arm/mm/core_mmu_lpae.o out/arm/core/arch/arm/mm/pgt_cache.o out/arm/core/arch/arm/mm/mobj_dyn_shm.o out/arm/core/arch/arm/tee/arch_svc_a64.o out/arm/core/arch/arm/tee/arch_svc.o out/arm/core/arch/arm/tee/entry_fast.o out/arm/core/arch/arm/tee/cache.o out/arm/core/arch/arm/plat-vexpress/main.o out/arm/core/arch/arm/plat-vexpress/vendor_props.o out/arm/core/crypto/crypto.o out/arm/core/crypto/aes-gcm.o out/arm/core/crypto/aes-gcm-sw.o out/arm/core/crypto/aes-gcm-ghash-tbl.o out/arm/core/crypto/signed_hdr.o out/arm/core/crypto/rng_fortuna.o out/arm/core/crypto/cbc-mac.o out/arm/core/crypto/aes-cts.o out/arm/core/crypto/sm2-kdf.o out/arm/core/crypto/sm3.o out/arm/core/crypto/sm3-hash.o out/arm/core/crypto/sm3-hmac.o out/arm/core/crypto/sm4.o out/arm/core/crypto/sm4-ecb.o out/arm/core/crypto/sm4-cbc.o out/arm/core/crypto/sm4-ctr.o out/arm/core/drivers/pl011.o out/arm/core/drivers/gic.o out/arm/core/drivers/clk/clk.o out/arm/core/drivers/clk/clk_dt.o out/arm/core/drivers/clk/fixed_clk.o out/arm/core/drivers/rstctrl/rstctrl.o out/arm/core/kernel/spin_lock_debug.o out/arm/core/kernel/assert.o out/arm/core/kernel/console.o out/arm/core/kernel/dt.o out/arm/core/kernel/dt_driver.o out/arm/core/kernel/dt_driver_test.o out/arm/core/kernel/pm.o out/arm/core/kernel/handle.o out/arm/core/kernel/interrupt.o out/arm/core/kernel/ldelf_syscalls.o out/arm/core/kernel/msg_param.o out/arm/core/kernel/panic.o out/arm/core/kernel/refcount.o out/arm/core/kernel/tee_misc.o out/arm/core/kernel/tee_ta_manager.o out/arm/core/kernel/ts_manager.o out/arm/core/kernel/scattered_array.o out/arm/core/kernel/huk_subkey.o out/arm/core/kernel/user_mode_ctx.o out/arm/core/kernel/initcall.o out/arm/core/kernel/user_access.o out/arm/core/kernel/mutex.o out/arm/core/kernel/wait_queue.o out/arm/core/kernel/notif.o out/arm/core/kernel/thread.o out/arm/core/kernel/user_ta.o out/arm/core/kernel/ree_fs_ta.o out/arm/core/kernel/early_ta.o out/arm/core/kernel/secstor_ta.o out/arm/core/kernel/embedded_ts.o out/arm/core/kernel/pseudo_ta.o out/arm/core/mm/mobj.o out/arm/core/mm/fobj.o out/arm/core/mm/file.o out/arm/core/mm/vm.o out/arm/core/mm/core_mmu.o out/arm/core/mm/tee_mm.o out/arm/core/pta/device.o out/arm/core/pta/secstor_ta_mgmt.o out/arm/core/pta/stats.o out/arm/core/pta/system.o out/arm/core/pta/scmi.o out/arm/core/pta/tests/fs_htree.o out/arm/core/pta/tests/invoke.o out/arm/core/pta/tests/misc.o out/arm/core/pta/tests/mutex.o out/arm/core/pta/tests/aes_perf.o out/arm/core/tee/entry_std.o out/arm/core/tee/tee_cryp_utl.o out/arm/core/tee/tee_cryp_hkdf.o out/arm/core/tee/tee_cryp_concat_kdf.o out/arm/core/tee/tee_cryp_pbkdf2.o out/arm/core/tee/tee_obj.o out/arm/core/tee/tee_svc.o out/arm/core/tee/tee_svc_cryp.o out/arm/core/tee/tee_svc_storage.o out/arm/core/tee/tee_time_generic.o out/arm/core/tee/tadb.o out/arm/core/tee/socket.o out/arm/core/tee/tee_ta_enc_manager.o out/arm/core/tee/tee_fs_key_manager.o out/arm/core/tee/tee_ree_fs.o out/arm/core/tee/fs_dirfile.o out/arm/core/tee/fs_htree.o out/arm/core/tee/tee_fs_rpc.o out/arm/core/tee/tee_pobj.o out/arm/core/tee/uuid.o out/arm/core/tee/tee_supp_plugin_rpc.o out/arm/core/ta_pub_key.o out/arm/core/ldelf_hex.o out/arm/core/early_ta_f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c.o out/arm/core/arch/arm/dts/embedded_dtb_test.o   out/arm/core/scmi-server/libscmi-server.a out/arm/lib/libunw/libunw.a out/arm/core/lib/zlib/libzlib.a out/arm/core/lib/libfdt/libfdt.a out/arm/core/lib/libtomcrypt/libtomcrypt.a out/arm/core-lib/libmbedtls/libmbedtls.a out/arm/core-lib/libutils/libutils.a  out/arm/core/scmi-server/build/product/optee-fvp/fw/libscmi-fw-all.a out/arm/core-lib/libutils/libutils.a /home/vingu/Linaro/SCMI/optee-project-qemu-latest/toolchains/aarch64/bin/../lib/gcc/aarch64-none-linux-gnu/10.2.1/libgcc.a -o out/arm/core/all_objs.o

The only difference is the position of libscmi-fw-all.a vs libutils.a

etienne-lms:
@vingu-linaro, I've added your change to add support for qemuv8.

core/scmi-server/sub.mk
etienne-lms: moving `scmi-server` before `libutils` in _core/core.mk_ would resolve the dependency. But maybe a bit fragile.

vingu-linaro: I have also considered this change but I agree that it's a bit fragile

jforissier: > moving `scmi-server` before `libutils` in _core/core.mk_ would resolve the dependency. But maybe a bit fragile.

The libs should be in reverse order of appearance in core.mk. `mk/lib.mk` should take care of that, perhaps there is a bug in it or in the way it is used.

The naming is a bit weird too, I see no reason not stick more closely with the lib model and have all this code under `core/lib/libscmi-fwk-all`. Any deviation can only be painful when looking for stuff etc.

jenswi-linaro: The problem is one line above. `mk/lib.mk` adds to `libdeps` in front and here you're appending.

etienne-lms: thanks @jenswi-linaro, indeed changing line 16 should fix your issue @vingu-linaro:
```patch
-libdeps += $(scp-firmware-output)
+libdeps := $(scp-firmware-output) $(libdeps)
```

I'll push a fixup commit.

etienne-lms:
I've pushed 2 more fixup commits: moving _scmi-server/_ to _core/lib/_ and fixing the issue discussed in https://github.com/OP-TEE/optee_os/pull/5627#discussion_r1031737099. 
@vingu-linaro: can you update https://github.com/ARM-software/SCP-firmware/pull/664 with this change
https://github.com/etienne-lms/SCP-firmware/commit/454eb02250a92f5cc5299d2b794e4f88368745f2?

CI should succeed to build.

jforissier:
> CI should succeed to build.

It does! :+1: 

Honestly this is somewhat abusing the `mk/lib.mk` mechanism, because the build rules are not supposed to be in the `sub.mk` file (i.e., `core/lib/scmi-server/sub.mk`). As a result `mk/lib.mk` does things that are not needed. A cleaner approach would be to create a dedicated file for external sources built using cmake, such as `mk/lib_ext_cmake.mk`, which could be fed a couple of parameters form the `sub.mk`. But I'm OK with this until we meet another such case.

jforissier (approved):

jenswi-linaro:
> A cleaner approach would be to create a dedicated file for external sources built using cmake, such as `mk/lib_ext_cmake.mk`, which could be fed a couple of parameters form the `sub.mk`. But I'm OK with this until we meet another such case.

I disagree, a clean approach would include the needed parts of SCP-firmware in the optee_os git and compile it as any other library. The way the se050 stuff is done is a compromise, but at least we have a well-integrated build system.

This interaction with cmake is terrible, it's bad enough to build with a sub-make, and the only way to make it worse would be to include autotools in the stew.

jforissier:
> > A cleaner approach would be to create a dedicated file for external sources built using cmake, such as `mk/lib_ext_cmake.mk`, which could be fed a couple of parameters form the `sub.mk`. But I'm OK with this until we meet another such case.
> 
> I disagree, a clean approach would include the needed parts of SCP-firmware in the optee_os git and compile it as any other library. The way the se050 stuff is done is a compromise, but at least we have a well-integrated build system.

Hmmm. Yeah. But it seems a bit much to ask the SCP Firmware project to adopt the OP-TEE build system, because OP-TEE is only one possible target, right?

> This interaction with cmake is terrible, it's bad enough to build with a sub-make, and the only way to make it worse would be to include autotools in the stew.

Hey, bitbake would be even worse! :wink:

jenswi-linaro:
> Hmmm. Yeah. But it seems a bit much to ask the SCP Firmware project to adopt the OP-TEE build system, because OP-TEE is only one possible target, right?

That hasn't stopped us with mbedtls, LTC, zlib, libfdt and se050.

etienne-lms:
Fixup commit squashed.
I applied @jforissier review tag.

@xiaoxuZeng, could you have a look at the 2 below commits?
_"core: scmi_server: Build a SCMI server from SCP-firmware"_
_"core: pta: scmi: support SCP-firmware SCMI resources"_
These commits have changed since your review tag was sent so I removed your tag from these 2 commits. Please, tell me if I can apply you review tag.

etienne-lms:
After dicsussions with maintainers, it's been decided to change how we build SCP-firmware, not using CMake native build of the package but rather using optee_os sub.mk based makefiles. I've created https://github.com/OP-TEE/optee_os/pull/5686 for that purpose and will close current P-R.

PR #5630, opened 2022-11-08
ldelf: supports riscv ldelf startup and riscv TAs loading

liushiwei007:
Added 64-bit riscv ldelf startup assembly and
parsing 64-bit elf files.

Tested-by: Liu Shiwei <liushiwei@eswincomputing.com>

### DIFF
--- a/ldelf/ldelf.ld.S
+++ b/ldelf/ldelf.ld.S
@@ -12,6 +12,9 @@ OUTPUT_ARCH(arm)
 OUTPUT_FORMAT("elf64-littleaarch64")
 OUTPUT_ARCH(aarch64)
 #endif
+#ifdef RV64
+OUTPUT_ARCH(riscv)
+#endif
 
 ENTRY(_ldelf_start)
 SECTIONS {
--- a/ldelf/start_rv64.S
+++ b/ldelf/start_rv64.S
@@ -1,0 +1,61 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright  2022  Beijing ESWIN Computing Technology Co., Ltd.
+ */
+
+#include <asm.S>
+#include <elf_common.h>
+
+/*
+ * _ldelf_start() - Entry of ldelf
+ *
+ * See include/ldelf.h for details on TEE Core interaction.
+ *
+ * void _ldelf_start(struct ldelf_arg *arg);
+ */
+FUNC _ldelf_start , :
+	/*
+	 * First ldelf needs to be relocated. The binary is compiled to
+	 * contain only a minimal number of R_RISCV_RELATIVE relocations in
+	 * read/write memory, leaving read-only and executable memory
+	 * untouched.
+	 */
+	lla	a1, reloc_end_rel
+	lw	a3, 0(a1)
+	lla	a1, reloc_begin_rel
+	lw	a2, 0(a1)
+	add	a2, a2, a1
+	add	a3, a3, a1
+	beq	a2, a3, 2f
+
+	lla	a1, _ldelf_start	/* Get the load offset */
+
+	/* Loop over the relocations (Elf64_Rela) and process all entries */
+1:
+	ld	t1, 0(a2)	/* t1 = r_offset */
+	ld	t2, 8(a2)	/* t2 = r_info */
+	ld	t3, 16(a2)	/* t3 = r_addend */
+	addi	a2, a2, 24
+	and	t2, t2, 0xff
+	addi	t4, zero, R_RISCV_RELATIVE
+	bne	t2, t4, 3f
+
+	/* Update the pointer at r_offset + load offset */
+	add	t1, a1, t1
+	ld	t4, 0(t1)
+	add	t4, t4, t3
+	add	t4, t4, a1
+	sd	t4, 0(t1)
+	ble	a2, a3, 1b
+2:
+	jal	ldelf
+	addi	a0, a0, 0
+	jal	_ldelf_return
+3:
+	addi	a0, a0, 0
+	jal	_ldelf_panic
+reloc_begin_rel:
+    .word __reloc_begin - reloc_begin_rel
+reloc_end_rel:
+    .word __reloc_end - reloc_end_rel
+END_FUNC _ldelf_start
--- a/ldelf/ta_elf.c
+++ b/ldelf/ta_elf.c
@@ -98,6 +98,7 @@ struct ta_elf *ta_elf_find_elf(const TEE_UUID *uuid)
 	return NULL;
 }
 
+#if defined(ARM32) || defined(ARM64)
 static TEE_Result e32_parse_ehdr(struct ta_elf *elf, Elf32_Ehdr *ehdr)
 {
 	if (ehdr->e_ident[EI_VERSION] != EV_CURRENT ||
@@ -156,6 +157,38 @@ static TEE_Result e64_parse_ehdr(struct ta_elf *elf __unused,
 	return TEE_ERROR_NOT_SUPPORTED;
 }
 #endif /*ARM64*/
+#endif /* ARM32 || ARM64 */
+
+#if defined(RV64)
+static TEE_Result e32_parse_ehdr(struct ta_elf *elf __unused,
+				 Elf32_Ehdr *ehdr __unused)
+{
+		return TEE_ERROR_BAD_FORMAT;
+}
+
+static TEE_Result e64_parse_ehdr(struct ta_elf *elf, Elf64_Ehdr *ehdr)
+{
+	if (ehdr->e_ident[EI_VERSION] != EV_CURRENT ||
+	    ehdr->e_ident[EI_CLASS] != ELFCLASS64 ||
+	    ehdr->e_ident[EI_DATA] != ELFDATA2LSB ||
+	    ehdr->e_ident[EI_OSABI] != ELFOSABI_NONE ||
+	    ehdr->e_type != ET_DYN || ehdr->e_machine != EM_RISCV ||
+	    ehdr->e_phentsize != sizeof(Elf64_Phdr) ||
+	    ehdr->e_shentsize != sizeof(Elf64_Shdr))
+		return TEE_ERROR_BAD_FORMAT;
+
+	elf->is_32bit = false;
+	elf->e_entry = ehdr->e_entry;
+	elf->e_phoff = ehdr->e_phoff;
+	elf->e_shoff = ehdr->e_shoff;
+	elf->e_phnum = ehdr->e_phnum;
+	elf->e_shnum = ehdr->e_shnum;
+	elf->e_phentsize = ehdr->e_phentsize;
+	elf->e_shentsize = ehdr->e_shentsize;
+
+	return TEE_SUCCESS;
+}
+#endif /* RV64 */
 
 static void check_phdr_in_range(struct ta_elf *elf, unsigned int type,
 				vaddr_t addr, size_t memsz)
--- a/ldelf/ta_elf_rel.c
+++ b/ldelf/ta_elf_rel.c
@@ -478,7 +478,7 @@ static void e32_relocate(struct ta_elf *elf, unsigned int rel_sidx)
 	}
 }
 
-#ifdef ARM64
+#if defined(ARM64) || defined(RV64)
 static void e64_get_sym_name(const Elf64_Sym *sym_tab, size_t num_syms,
 			     const char *str_tab, size_t str_tab_size,
 			     Elf64_Rela *rela, const char **name,
@@ -518,6 +518,7 @@ static void e64_process_dyn_rela(const Elf64_Sym *sym_tab, size_t num_syms,
 	*where = val;
 }
 
+#ifdef ARM64
 static void e64_process_tls_tprel_rela(const Elf64_Sym *sym_tab,
 				       size_t num_syms, const char *str_tab,
 				       size_t str_tab_size, Elf64_Rela *rela,
@@ -570,6 +571,7 @@ static void e64_process_tlsdesc_rela(const Elf64_Sym *sym_tab, size_t num_syms,
 	e64_process_tls_tprel_rela(sym_tab, num_syms, str_tab, str_tab_size,
 				   rela, where + 1, elf);
 }
+#endif /*ARM64*/
 
 static void e64_relocate(struct ta_elf *elf, unsigned int rel_sidx)
 {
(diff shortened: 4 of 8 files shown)


ldelf/start_rv64.S
```diff
@@ -0,0 +1,62 @@
...
+ * void start(struct ldelf_arg *arg);
+ */
+FUNC _ldelf_start , :
+	/*
+	 * First ldelf needs to be relocated. The binary is compiled to
+	 * contain only a minimal number of R_RISCV_RELATIVE relocations in
+	 * read/write memory, leaving read-only and executeble memory
+	 * untouched.
+	 */
+	lla a1, reloc_end_rel
```
jenswi-linaro, line 24: Please add a tab character between the instruction and its arguments. This comment applies to all the instructions in this file.

ldelf/ta_elf.c
```diff
@@ -149,7 +149,30 @@ static TEE_Result e64_parse_ehdr(struct ta_elf *elf, Elf64_Ehdr *ehdr)
...
-#else /*ARM64*/
+#elif defined RV64
+static TEE_Result e64_parse_ehdr(struct ta_elf *elf, Elf64_Ehdr *ehdr)
+{
+	if (ehdr->e_ident[EI_VERSION] != EV_CURRENT ||
+	    ehdr->e_ident[EI_CLASS] != ELFCLASS64 ||
+	    ehdr->e_ident[EI_DATA] != ELFDATA2LSB ||
+	    ehdr->e_ident[EI_OSABI] != ELFOSABI_NONE ||
+	    ehdr->e_type != ET_DYN || ehdr->e_machine != EM_RISCV ||
+	    ehdr->e_phentsize != sizeof(Elf64_Phdr) ||
```
jenswi-linaro, line 160: Don't care about `e_flags`?

ldelf/ta_elf.c
```diff
@@ -149,7 +149,30 @@ static TEE_Result e64_parse_ehdr(struct ta_elf *elf, Elf64_Ehdr *ehdr)
 
 	return TEE_SUCCESS;
 }
-#else /*ARM64*/
+#elif defined RV64
```
jenswi-linaro, line 152: `#elif defined(RV64)`

You should probably add some `#if defined(ARM32) || defined(ARM64)` above `e32_parse_ehdr()` and provide a Risc-V alternative for that too.

jforissier, line 152: Addparentheses: `defined(RV64)`

lib/libutils/ext/include/confine_array_index.h
```diff
@@ -142,4 +142,11 @@ static inline size_t confine_array_index(size_t index, size_t size) {
   return safe_index;
 }
 #endif
+#ifdef __riscv
+static inline size_t confine_array_index(size_t index, size_t size) {
+	(void)index;
+	(void)size;
+	return index;
+}
```
jforissier, line 150: Please use `__unused` instead of `(void)size;`, and note that `index` is used.
```C
static inline size_t confine_array_index(size_t index, size_t size __unused) {
	return index;
}
```

jforissier (commented):
Hi @liushiwei007, thanks for the patch!

Please use the spelling "RISC-V" in the commit description, and "ELF" not "elf". I suggest:
```
ldelf: support RISC-V

Add 64-bit RISC-V ldelf startup assembly and parsing 64-bit ELF files.
```

WIth this and the comments below addressed:

ldelf/ta_elf_rel.c
```diff
@@ -659,6 +661,14 @@ static void e64_relocate(struct ta_elf *elf, unsigned int rel_sidx)
 			 * here in case such entries exist.
 			 */
 			break;
+		case R_RISCV_RELATIVE:
```
jenswi-linaro, line 664: It's nice that so much of the 64-bit ELF code can be reused. However, we should not try to relocate a Risc-V ELF with AArch64 relocations types and vice versa. Please do something like:
``` C
#ifdef ARM64
    case R_AARCH64_NONE:
...
    case R_AARCH64_TLSDESC:
...
#endif
#ifdef RV64
    case R_RISCV_NONE:
...
    case R_RISCV_JUMP_SLOT:
...
#endif

File: lib/libutee/include/elf_common.h [hunk 6]
```diff
@@ -683,6 +684,13 @@ typedef struct {
 #define	R_386_TLS_TPOFF32	37	/* GOT entry of -ve static TLS offset */
 #define	R_386_IRELATIVE		42	/* PLT entry resolved indirectly at runtime */
 
+#define	R_RISCV_NONE		0	/* No relocation. */
+#define	R_RISCV_64			2	/* 64-bit relocation. */
```
jenswi-linaro, line 688: Strange indentation

lib/libutee/include/elf_common.h
```diff
@@ -683,6 +684,13 @@ typedef struct {
...
+#define	R_RISCV_RELATIVE	3	/* Adjust a link address (A) to its
+									load address (B + A). */
```
jenswi-linaro, line 690: Please align with the start of "Adjust" on the line above. Same comment two lines down.

lib/libutils/ext/include/confine_array_index.h
```diff
@@ -142,4 +142,11 @@ static inline size_t confine_array_index(size_t index, size_t size) {
   return safe_index;
 }
 #endif
+#ifdef __riscv
```
jenswi-linaro, line 145: Please add an empty line above. By the way, changes in this file should go into a separate commit.

lib/libutils/ext/include/confine_array_index.h
```diff
@@ -142,4 +142,11 @@ static inline size_t confine_array_index(size_t index, size_t size) {
...
+static inline size_t confine_array_index(size_t index, size_t size) {
+	(void)index;
+	(void)size;
```
jenswi-linaro, line 148: Risc-V is not immune to Spectre V1 type of attacks. Please try to provide a real implementation for this function.

lib/libutils/ext/include/confine_array_index.h
liushiwei007: Ok, I'll analyze the Spectre V1 and put together a separate commit

ldelf/ta_elf.c
liushiwei007: Hi @jenswi-linaro, RISC-V e_flags contains four parts RVC, Float ABI, RVE and TSO(please refer to [https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc](url) ), but none of these four parts are relevant to the current processing logic.

jenswi-linaro: OK

ldelf/ldelf.ld.S
```diff
@@ -12,6 +12,10 @@ OUTPUT_ARCH(arm)
 OUTPUT_FORMAT("elf64-littleaarch64")
 OUTPUT_ARCH(aarch64)
 #endif
+#ifdef RV64
+OUTPUT_ARCH(riscv)
+#endif
+
```
etienne-lms, line 18: remove empty line

ldelf/start_rv64.S
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * See include/ldelf.h for details on TEE Core interaction.
+ *
+ * void start(struct ldelf_arg *arg);
+ */
+FUNC _ldelf_start , :
+	/*
+	 * First ldelf needs to be relocated. The binary is compiled to
+	 * contain only a minimal number of R_RISCV_RELATIVE relocations in
+	 * read/write memory, leaving read-only and executeble memory
```
etienne-lms, line 21: s/executeble/executable/

ldelf/start_rv64.S
```diff
@@ -0,0 +1,62 @@
...
+	/* Update the pointer at r_offset + load offset */
+	add t1, a1, t1
+	ld t4, 0(t1)
+	add t4, t4, t3
+	add t4, t4, a1
+	sd t4, 0(t1)
+	ble a2, a3, 1b
+2:
+	jal ldelf
+	addi a0, a0,0
```
etienne-lms, line 53: space char after `,`. ditto below.

etienne-lms (commented):
few minor comments

jforissier:
@liushiwei007 please do not add `Acked-by:` or `Reviewed-by:` unless given explicitly in the comments. Mine is OK, but Jens and Etienne probably want to check that their comments have been addressed before their tags are added. Thanks!

jenswi-linaro:

jenswi-linaro:
I don't suppose this compiles at the moment since `confine_array_index()` is missing.
However, it has been tested with some out-of-tree patches and the patch itself looks nice.

jforissier:
> I don't suppose this compiles at the moment since `confine_array_index()` is missing. However, it has been tested with some out-of-tree patches and the patch itself looks nice.

It does, but I'd rather not merge any "advanced" stuff that is not at least compile-tested before basic platform support is in (#5606), and before RISC-V is added to CI (I can take care of that).

liushiwei007:
> > I don't suppose this compiles at the moment since `confine_array_index()` is missing. However, it has been tested with some out-of-tree patches and the patch itself looks nice.
> 
> It does, but I'd rather not merge any "advanced" stuff that is not at least compile-tested before basic platform support is in (#5606), and before RISC-V is added to CI (I can take care of that).

So I continue to submit my other patch and wait for (#5606) merge?

jforissier:
@liushiwei007 yes, feel free to submit other PRs so they can be reviewed sooner.

lib/libutils/ext/include/confine_array_index.h
```diff
@@ -142,4 +142,19 @@ static inline size_t confine_array_index(size_t index, size_t size) {
   return safe_index;
 }
 #endif
+
+#ifdef __riscv
+static inline size_t confine_array_index(size_t index, size_t size) {
+	size_t safe_index = 0;
+	__asm__(
+	"slt	t0, %2, %1\n"
```
jenswi-linaro, line 150: Is this implementation speculation safe? I would have expected a barrier or something. Would you mind adding a short comment to give a hint as to what's going on?

By the way, you should probably not use explicit registers like `t0`, instead, you should use some `%x` allocated by the compiler.

jforissier: This may be useful: https://arxiv.org/abs/2206.04507

lib/libutils/ext/include/confine_array_index.h
```diff
@@ -142,4 +142,24 @@ static inline size_t confine_array_index(size_t index, size_t size) {
...
 #endif
+
+#ifdef __riscv
+static inline size_t confine_array_index(size_t index, size_t size) {
+	size_t safe_index = 0;
+  // Use a Set if Less Than to enforce validation of |index|.
+  // slt rd, rs1, rs2, Compare the numbers in x[rs1] and x[rs2].
+  // If x[rs1] is smaller, write 1 to x[rd], otherwise write 0.
+  // Subsequent instructions depend on the results of the comparison
+  // and cannot be executed until the comparison is resolved.
```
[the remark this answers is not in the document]
liushiwei007: Not explicitly stated, It's just that the assembly code doesn't choose branching.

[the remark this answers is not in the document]
liushiwei007: I just found out that my message entered pengding，I'm sorry 。

ldelf/ta_elf_rel.c
```diff
@@ -478,7 +478,7 @@ static void e32_relocate(struct ta_elf *elf, unsigned int rel_sidx)
 	}
 }
 
-#ifdef ARM64
+#if defined(ARM64) || defined(RV64)
```
etienne-lms, line 481: can you update also line 726 and 732?   s/`ARM64`/`ARM64 || RV64`/

liushiwei007: ok

jenswi-linaro:
Please create a new PR for the riscv implementation of `confine_array_index()`.  That should make it easier to keep the focus on the topic in the PR.

lib/libutils/ext/include/confine_array_index.h
jenswi-linaro: This is a tricky subject, keep in mind that the `confine_array_index()` function is only supposed to be called with an index out of bound while speculating. So for a CPU that doesn't speculate the function will only see valid indexes.

liushiwei007: Yes, my logic is the same, return 0 for out-of-range indexes, otherwise return index itself, and the computation itself is not branching。

liushiwei007: Is this submission OK？

jenswi-linaro: I'm not sure this is a speculation safe implementation.

jforissier: > I'm not sure this is a speculation safe implementation.

How about we leave this function unimplemented until some real hardware is supported by OP-TEE? Then we'll have more information on whether or not the CPU is vulnerable and what mitigations are effective.

jenswi-linaro: It's needed by ldelf. I suppose we could have a naive C implementation as fallback until it's properly implemented on riscv.

jforissier: @jenswi-linaro yes, it is what I meant ("unimplemented" was not the proper word).

liushiwei007: @jenswi-linaro  hi，I don't quite understand “ a naive C implementation”。Does it mean to determine whether index is smaller than size use C language？

jenswi-linaro: Yes, to make things work, but without protection against side-channel attacks.

liushiwei007: Ok，I see。

jforissier (commented):
For commit "libutils: confine_array_index: add support for RISC-V", with my comments addressed:

Please move that commit before the other one, it makes more sense.

liushiwei007:
> For commit "libutils: confine_array_index: add support for RISC-V", with my comments addressed:
> Please move that commit before the other one, it makes more sense.
You mean to leave both the confine_array_index assembly implementation and the C implementation, and put the assembly implementation first?

liushiwei007:
> For commit "libutils: confine_array_index: add support for RISC-V", with my comments addressed:
> 
> 
> Please move that commit before the other one, it makes more sense.

You mean to leave both the confine_array_index assembly implementation and the C implementation, and put the assembly implementation first?

jforissier:
> You mean to leave both the confine_array_index assembly implementation and the C implementation, and put the assembly implementation first?

No, I mean in the Git history, "libutils: confine_array_index: add support for RISC-V" should come first, then "ldelf: support RISC-V" should come on top. Because the latter depends on the former.

jenswi-linaro:

jforissier:
@liushiwei007 thanks!

PR #5639, opened 2022-11-09
core: ree_fs: check ta size early

jenswi-linaro:
As soon as TA signature has been verified check that the total loaded size matches what is in the sign header. This prevents an eventual attacker from providing arbitrary values in the img_size field of the signed header.

### DIFF
--- a/core/kernel/ree_fs_ta.c
+++ b/core/kernel/ree_fs_ta.c
@@ -424,6 +424,14 @@ static TEE_Result ree_fs_ta_open(const TEE_UUID *uuid,
 			goto error_free_hash;
 		}
 
+		/*
+		 * This is checked further down too, but we must sanity
+		 * check shdr->img_size before it's used below.
+		 */
+		if (ta_size != sz + shdr->img_size) {
+			res = TEE_ERROR_SECURITY;
+			goto error_free_hash;
+		}
 
 		ehdr = malloc(ehdr_sz);
 		if (!ehdr) {
--- a/core/kernel/ree_fs_ta.c
+++ b/core/kernel/ree_fs_ta.c
@@ -439,7 +439,10 @@ static TEE_Result ree_fs_ta_open(const TEE_UUID *uuid,
 			goto error_free_hash;
 		}
 
-		memcpy(ehdr, ((uint8_t *)ta + offs), ehdr_sz);
+		*ehdr = img_ehdr;
+		memcpy((uint8_t *)ehdr + sizeof(img_ehdr),
+		       (uint8_t *)ta + offs + sizeof(img_ehdr),
+		       ehdr_sz - sizeof(img_ehdr));
 
 		res = crypto_hash_update(hash_ctx, (uint8_t *)ehdr, ehdr_sz);
 		if (res != TEE_SUCCESS)


core/kernel/ree_fs_ta.c
```diff
@@ -229,6 +229,31 @@ static TEE_Result rpc_load(const TEE_UUID *uuid, struct shdr **ta,
...
+		break;
+	case SHDR_ENCRYPTED_TA:
+		hdr_sz = sizeof(struct shdr_bootstrap_ta) +
+			 sizeof(struct shdr_encrypted_ta);
+		break;
+	default:
+		return TEE_ERROR_SECURITY;
+	}
+
+	if (ta_size != offs + hdr_sz + shdr->img_size)
```
etienne-lms, line 251: we should add the signed heade size also I think (from `SHDR_GET_SIZE(shdr)` or like).

Sometthing like:
```c
static TEE_Result check_ta_type_and_size(const struct shdr *shdr, size_t offs,
					 size_t ta_size, uint8_t *ta)
{
	struct shdr_encrypted_ta img_ehdr = { };
	size_t hdr_sz = SHDR_GET_SIZE(shdr);

	switch (shdr->img_type) {
	case SHDR_TA:
		break;
	case SHDR_BOOTSTRAP_TA:
		hdr_sz += sizeof(struct shdr_bootstrap_ta);
		break;
	case SHDR_ENCRYPTED_TA:
		hdr_sz += sizeof(struct shdr_bootstrap_ta);
		memcpy(&img_ehdr, ta + offs + hdr_sz, sizeof(img_ehdr));
		hdr_sz += SHDR_ENC_GET_SIZE(&img_ehdr);
		break;
	default:
		return TEE_ERROR_SECURITY;
	}

	if (ta_size != offs + hdr_sz + shdr->img_size)
		return TEE_ERROR_SECURITY;

	return TEE_SUCCESS;
}
```
Maybe some ADD_OVERFLOW are missing.

core/kernel/ree_fs_ta.c
```diff
@@ -229,6 +229,31 @@ static TEE_Result rpc_load(const TEE_UUID *uuid, struct shdr **ta,
...
+
+	switch (shdr->img_type) {
+	case SHDR_TA:
+		break;
+	case SHDR_BOOTSTRAP_TA:
+		hdr_sz = sizeof(struct shdr_bootstrap_ta);
+		break;
+	case SHDR_ENCRYPTED_TA:
+		hdr_sz = sizeof(struct shdr_bootstrap_ta) +
+			 sizeof(struct shdr_encrypted_ta);
```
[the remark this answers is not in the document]
jenswi-linaro: You're right.

core/kernel/ree_fs_ta.c
jenswi-linaro: The initial `SHDR_GET_SIZE(shdr)` isn't needed, that's already added to offs.

etienne-lms: It was added only for the subkey headers found in the binary, not for the TA image (bootstrap, encrypted or installed) part were looking into here.

jbech-linaro:
```
```

jenswi-linaro:
This became more complicated than anticipated. I'm going to start over.

jenswi-linaro:
Updated with a new approach. I found another small issue that I'm addressing in the last patch.

etienne-lms (commented):

jenswi-linaro:
Tag applied

PR #5640, opened 2022-11-09
core: shdr: check that hash algorithm is strong enough

jenswi-linaro:
Until now shdr_verify_signature() accepted any hash GP algorithm known to OP-TEE. A few of those (MD5 and SHA-1) are known to be weak. So add an extra check to only allow algorithms strong enough.

### DIFF
--- a/core/crypto/signed_hdr.c
+++ b/core/crypto/signed_hdr.c
@@ -48,12 +48,19 @@ struct shdr *shdr_alloc_and_copy(size_t offs, const void *img, size_t img_size)
 	return shdr;
 }
 
+static bool is_weak_hash_algo(uint32_t algo)
+{
+	return algo == TEE_ALG_MD5 || algo == TEE_ALG_SHA1 ||
+	       algo == TEE_ALG_MD5SHA1;
+}
+
 TEE_Result shdr_verify_signature(const struct shdr *shdr)
 {
-	struct rsa_public_key key;
-	TEE_Result res;
+	struct rsa_public_key key = { };
+	TEE_Result res = TEE_SUCCESS;
 	uint32_t e = TEE_U32_TO_BIG_ENDIAN(ta_pub_key_exponent);
-	size_t hash_size;
+	size_t hash_size = 0;
+	size_t hash_algo = 0;
 
 	if (shdr->magic != SHDR_MAGIC)
 		return TEE_ERROR_SECURITY;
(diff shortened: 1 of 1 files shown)


core/crypto/signed_hdr.c
```diff
@@ -48,21 +48,31 @@ struct shdr *shdr_alloc_and_copy(size_t offs, const void *img, size_t img_size)
 	return shdr;
 }
 
+static bool is_strong_hash_algo(uint32_t algo)
+{
+	return algo == TEE_ALG_SHA224 || algo == TEE_ALG_SHA256 ||
+	       algo == TEE_ALG_SHA384 || algo == TEE_ALG_SHA512;
+}
```
jforissier, line 55: How about excluding the known weak algorithms instead? There would be only two entries here, and the code would only need changing when an algorithm becomes weak, not when a new (strong) algorithm is added.

jforissier (commented):
One comment, then:

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with a minor coding style fix and with or without @jforissier comment addressed.

core/crypto/signed_hdr.c
jenswi-linaro: Assume that we check against weak algorithms instead. Should we check against MD5 even if it can occur?

It seems more clear to check which algorithms we're prepared to accept.

jforissier: > Assume that we check against weak algorithms instead. Should we check against MD5 even if it can occur?

Yes.

> It seems more clear to check which algorithms we're prepared to accept.

...which is "*all* algorithms except the weak ones". For example, GP 1.3.1 defines `TEE_ALG_SHA3_{224,256,384,512}` which would be perfectly acceptable if OP-TEE implemented them. So with the following:

```C
static bool is_weak_hash_algo(uint32_t algo)
{
	return algo == TEE_ALG_MD5 || algo == TEE_ALG_SHA1;
}
```
...we could add support for SHA3 without changing anything here.

jbech-linaro:
```
```

jenswi-linaro:
Tags applied, except @jforissier's

core/crypto/signed_hdr.c
jenswi-linaro: I'm sorry, I didn't see your reply until now.
I suppose it should be safe to assume that we're not going to add any new weak algorithms, only strong ones.

jenswi-linaro: We have a definition for `TEE_ALG_MD5SHA1` too. I'm not sure what to make of that though.

jforissier: > I suppose it should be safe to assume that we're not going to add any new weak algorithms, only strong ones.

Yes, and that is the reason for my proposal. One of us must be missing something obvious in this discussion :-/
Let me try to state it again: if we're not going to add any new weak algorithm (we agree on that), why impose to change this code every time a new strong algorithm is added (your proposal), when we could just do nothing (mine)?

jforissier: > We have a definition for `TEE_ALG_MD5SHA1` too. I'm not sure what to make of that though.

Ha! Indeed. Non-standard algo. It is MD5 and SHA-1 concatenated, so if both are considered weak I guess it should be considered weak too. The hashes are not even combined in any subtle way.

jenswi-linaro:
Updated according to @jforissier's comment and applied his tag.

etienne-lms (approved):

jenswi-linaro:
Wait, I just noticed that we missed one function.

jenswi-linaro:
With the last update, we're covering `shdr_verify_signature2()` too.

jforissier:
> With the last update, we're covering `shdr_verify_signature2()` too.

etienne-lms (commented):

jenswi-linaro:
Squashed and tags applied.

PR #5641, opened 2022-11-09
core: fix potential integer overflow in syscall_log()

jenswi-linaro:
Fixes a potential integer overflow in syscall_log(). Note that an eventual overflow would still be caught by copy_from_user(), but it's preferable to catch this earlier.

### DIFF
--- a/core/tee/tee_svc.c
+++ b/core/tee/tee_svc.c
@@ -29,25 +29,26 @@
 
 vaddr_t tee_svc_uref_base;
 
-void syscall_log(const void *buf __maybe_unused, size_t len __maybe_unused)
+void syscall_log(const void *buf, size_t len)
 {
-#ifdef CFG_TEE_CORE_TA_TRACE
-	char *kbuf;
+	if (IS_ENABLED(CFG_TEE_CORE_TA_TRACE)) {
+		char *kbuf = NULL;
+		size_t sz = 0;
 
-	if (len == 0)
-		return;
+		if (!len || ADD_OVERFLOW(len, 1, &sz))
+			return;
 
-	kbuf = malloc(len + 1);
-	if (kbuf == NULL)
-		return;
+		kbuf = malloc(sz);
+		if (!kbuf)
+			return;
 
-	if (copy_from_user(kbuf, buf, len) == TEE_SUCCESS) {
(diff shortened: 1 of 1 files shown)


core/tee/tee_svc.c
```diff
@@ -31,23 +31,24 @@ vaddr_t tee_svc_uref_base;
 
 void syscall_log(const void *buf __maybe_unused, size_t len __maybe_unused)
 {
-#ifdef CFG_TEE_CORE_TA_TRACE
-	char *kbuf;
+	if (IS_ENABLED(CFG_TEE_CORE_TA_TRACE)) {
```
etienne-lms, line 34: can remove the `__maybe_unused` attributes at line 32.

etienne-lms (commented):

core/tee/tee_svc.c
xiaoxuZeng: if CFG_TEE_CORE_TA_TRACE not define, len maybe unused.

jbech-linaro:
```
```

jenswi-linaro:
Tags applied

core/tee/tee_svc.c
jforissier: @xiaoxuZeng just to clarify, `__maybe_unused` is needed when there is conditional compilation (`#ifdef CFG_SOMETHING`) but not when `IS_ENABLED(CFG_SOMETHING)` is used, because the later if effectively an `if ()` construct so the compiler will see the variables used inside the block even if the condition is false (which is why we prefer using `IS_ENABLED()` whenever possible).

xiaoxuZeng: Got， Thanks very much!

xiaoxuZeng: > @xiaoxuZeng just to clarify, `__maybe_unused` is needed when there is conditional compilation (`#ifdef CFG_SOMETHING`) but not when `IS_ENABLED(CFG_SOMETHING)` is used, because the latter if effectively an `if ()` construct so the compiler will see the variables used inside the block even if the condition is false (which is why we prefer using `IS_ENABLED()` whenever possible).

Got， Thanks very much!

PR #5643, opened 2022-11-13
AMD/Xilinx Versal ACAP: SHA3-384

ldts:


nathan-menhorn:
Hi @jforissier any updates on this PR?

core/drivers/crypto/versal/hash.c
```diff
@@ -0,0 +1,246 @@
...
+struct update_request {
+	uint8_t *data;
+	size_t len;
+	STAILQ_ENTRY(update_request) link;
+};
+
+struct versal_hash_ctx {
+	struct crypto_hash_ctx hash_ctx;
+	enum versal_sha3_state state;
+	STAILQ_HEAD(update_request_list, update_request) req_list;
```
jenswi-linaro, line 37: The `STAILQ_HEAD()` macro expands to:
``` C
struct update_request_list { ... }
```
I would prefer to have this done at the top level instead of inside `struct versal_hash_ctx` in case you ever need to use `struct update_request_list` at another place.

core/drivers/crypto/versal/hash.c
```diff
@@ -0,0 +1,246 @@
...
+}
+
+static TEE_Result do_hash_update(struct crypto_hash_ctx *ctx,
+				 const uint8_t *data, size_t len)
+{
+	struct versal_hash_ctx *c = to_versal_ctx(ctx);
+	struct update_request *req = NULL;
+
+	/*
+	 * The engine does not have the concept of context, so all update
```
jenswi-linaro, line 68: It seems that this implementation should target a special use case and should normally be avoided in a generic case where we don't have control over how much data is to be hashed.

If this is accessible from user space it can't be too hard to cause a panic somewhere in this driver.

ldts: Right, I did my best to provide standard functionality with the limitations it has. WRT panic, I suppose as easy as making OP-TEE panic...not sure if that is a bad thing as long as it is a controlled panic? sorry not sure what do you mean, maybe I miss your point

jenswi-linaro: Given how limited this is perhaps it's time to take a step back and consider the use cases we'd like to cover with this driver. For general purpose usage, I guess the system will panic so often that it will be more or less impossible to work with.

ldts: Not sure about that, in fact I doubt the standard user who allocates a context, processes and leaves will hit any issue.

https://github.com/ldts/optee_os/blob/xlnx_v2022.1.versal.001/core/drivers/crypto/versal/hash_test.c this is what I did to test.

jenswi-linaro: The way to make it pass is to pass a tiny payload. Try to check that hash of a 2meg TA or something.

ldts: yes that wouldn't be good. That would have been a good test to have in xtest. But yes, running out of memory is a risk with this architecture.

Having said that how is that use case is  different from say creating keys using the the pkcs#11 TA?. It is equally easy to panic the TA by creating keys in a loop. Is this a bug then?

But I understand your point. Unfortunately I have no work-around that...so what do you propose? shall I just drop it?

jenswi-linaro: A TA that panics isn't good, but if we panic and hang the entire system that's not good on a different level.

> But I understand your point. Unfortunately I have no work-around for that...so what do you propose? shall I just drop it?

Perhaps that's best for now since it can't be used as the primary generic hash implementation.

ldts: ok.

SHA3-384 was being used by the Versal HUK driver (which I'll post after these) but I can do it a bit differently

jenswi-linaro: Makes sense

core/drivers/crypto/versal/hash.c
```diff
@@ -0,0 +1,246 @@
...
+	arg.ibuf[0].mem = p;
+	arg.data[0] = NEXT_PACKET | init_mask | len;
+	arg.dlen = 1;
+
+	if (versal_crypto_request(VERSAL_SHA3_UPDATE, &arg, &err))
+		ret = TEE_ERROR_GENERIC;
+	else
+		c->state = SHA3_RUN;
+
+	free(p.buf);
```
etienne-lms, line 112: `versal_mbox_free()` helper functions would really be nice. It helps checking balanced malloc/free calls.

core/drivers/crypto/versal/hash.c
```diff
@@ -0,0 +1,246 @@
...
+	struct update_request *src_req = NULL;
+	struct update_request *dst_req = NULL;
+
+	src_hctx = container_of(src_ctx, struct versal_hash_ctx, hash_ctx);
+	dst_hctx = container_of(dst_ctx, struct versal_hash_ctx, hash_ctx);
+
+	dst_hctx->hash_ctx = src_hctx->hash_ctx;
+	dst_hctx->state = src_hctx->state;
+
+	/* duplicate the queue requests */
```
etienne-lms, line 176: s/duplicate/Duplicate/

core/drivers/crypto/versal/hash.c
```diff
@@ -0,0 +1,246 @@
...
+	/* Requests were pushed but never finalized */
+	STAILQ_FOREACH_SAFE(req, &hctx->req_list, link, next) {
+		free(req->data);
+		free(req);
+	}
+	free(hctx);
+}
+
+static TEE_Result versal_hash_alloc(struct crypto_hash_ctx **ctx,
+				    uint32_t algo __unused)
```
etienne-lms, line 209: remove `_unused`

ldts:
@etienne-lms sorry I am going to drop this driver from crypto and add it as a TEE specific driver due to the memory issue that @jenswi-linaro raised. But will take your comments into account.

core/drivers/crypto/versal/hash.c
nathan-menhorn: > yes that wouldn't be good. That would have been a good test to have in xtest. But yes, running out of memory is a risk with this architecture.
> 
> Having said that, how is that use case different from say creating keys using the the pkcs#11 TA?. It is equally easy to panic the TA by creating keys in a loop. Is this a bug then?
> 
> But I understand your point. Unfortunately I have no work-around for that...so what do you propose? shall I just drop it?

Hi @ldts specifically where are you concerned about running out of memory and which memory in particular? There's quite a bit of memory to chose from in Versal.

ldts: hi @nathan-menhorn. right so since the hardware engine doesn't keep the context, the TEE needs to keep contexts requests and updates in its internal memory until the active context finalizes. 
Imagine a thread starts an operation, and then stops while other threads start queuing sha3-384 requests. We will end up starving the memory available in the TEE.

Having said that, maybe we could just associate a memory pool to the sha3 driver if you think this driver is important - AFAIR it is main use case is for the HUK and not much more so I was planing of moving it to core/drivers so it is just internal to the TEE.

@jenswi-linaro would you be ok if I allocated a memory pool for this driver?

ldts: the idea is that big payloads will not affect the rest of the system, just the operation

jenswi-linaro: > @jenswi-linaro would you be ok if I allocated a memory pool for this driver?

I don't see how you can make that work. Whatever amount of memory you sacrifice you will still be able to run out of it.

nathan-menhorn: > hi @nathan-menhorn. right so since the hardware engine doesn't keep the context, the TEE needs to keep contexts requests and updates in its internal memory until the active context finalizes. Imagine a thread starts an operation, and then stops while other threads start queuing sha3-384 requests. We will end up starving the memory available in the TEE.
> 
> Having said that, maybe we could just associate a memory pool to the sha3 driver if you think this driver is important - AFAIR it is main use case is for the HUK and not much more so I was planing of moving it to core/drivers so it is just internal to the TEE.
> 

Hi @ldts that makes sense. There's use cases outside of the HUK but moving it core/drivers makes sense. However, I also see you are working on a shared memory pool. I'll just keep an eye on this PR. Thanks!

ldts: @jenswi-linaro I added some info to the commit messagge  https://github.com/OP-TEE/optee_os/pull/5643/commits/46f13a439273beea37071098d5f663b069484b99

ldts: it seems acceptable to me.

jenswi-linaro: No, this can't be used as a generic hash to the GlobalPlatform TEE Internal Core API.
It seems that this hardware can only be used in a more limited use case.

ldts: yeah I now fully agree. I tried to implement support for multithread and is just a huge pain on top of the memory consumption...it just wont scale. ok will move to core/drivers as a companion driver

ldts:
@jenswi-linaro @etienne-lms updated for review

ldts:
@nathan-menhorn this is a much simpler version which will not be reachable from user-space via pkcs#11.

nathan-menhorn:
> @nathan-menhorn this is a much simpler version which will not be reachable from user-space via pkcs#11.

Thanks for the update @ldts. I'll browse this code either end of today or tomorrow. Have you tested this in your branch or via TA or some other means?

ldts:
> > @nathan-menhorn this is a much simpler version which will not be reachable from user-space via pkcs#11.
> 
> Thanks for the update @ldts. I'll browse this code either end of today or tomorrow. Have you tested this in your branch or via TA or some other means?

right, just tested internally with a reference sha3-384 KAT. It is now very simple really since we dont need to simulate contexts.

ldts:
so can we merge this one? there is not much to it

nathan-menhorn:
> 

Hi @jenswi-linaro and @etienne-lms I would agree with @ldts. I've reviewed the patch and it was also tested against some KATs.

jenswi-linaro:

jforissier:
@nathan-menhorn that's good. As the one who pushes the merge button PRs, I am essentially waiting for Acked-by / Reviewed-by / Tested-by tags, so feel free to give yours if you want.

ldts:
@jforissier tags added

jforissier:
Thanks, merged with s/sha3-384/SHA3-384/ in the second commit subject.

PR #5646, opened 2022-11-14
Fault mitigation

jenswi-linaro:
This is https://github.com/OP-TEE/optee_os/pull/5247 rebased and updated.

The use case is to make `buf_ta_open()` more resilient to fault injection attacks on the hardware, more specifically glitching attacks. The basic assumption is that glitch can cause one or more instructions in sequence to act as nops when executed. Anything can happen of course but I'm using this approximation to have something more concrete to work with.

The fault mitigations are supposed to add zero overhead unless enabled and if enabled have a  modest overhead.

In this PR I've made fault mitigations enabled by default with `CFG_CORE_FAULT_MITIGATION?=y` in `mk/config.mk`. Should it be disabled by default instead?

### DIFF
--- a/core/crypto/signed_hdr.c
+++ b/core/crypto/signed_hdr.c
@@ -4,6 +4,7 @@
  */
 
 #include <crypto/crypto.h>
+#include <fault_mitigation.h>
 #include <kernel/panic.h>
 #include <mempool.h>
 #include <signed_hdr.h>
@@ -59,46 +60,56 @@ TEE_Result shdr_verify_signature(const struct shdr *shdr)
 	struct rsa_public_key key = { };
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t e = TEE_U32_TO_BIG_ENDIAN(ta_pub_key_exponent);
+	struct ftmn ftmn = { };
+	unsigned int err_incr = 2;
 	size_t hash_size = 0;
 	size_t hash_algo = 0;
 
 	if (shdr->magic != SHDR_MAGIC)
-		return TEE_ERROR_SECURITY;
+		goto err;
 
 	if (TEE_ALG_GET_MAIN_ALG(shdr->algo) != TEE_MAIN_ALGO_RSA)
-		return TEE_ERROR_SECURITY;
+		goto err;
 
 	hash_algo = TEE_DIGEST_HASH_TO_ALGO(shdr->algo);
 	if (is_weak_hash_algo(hash_algo))
-		return TEE_ERROR_SECURITY;
+		goto err;
 
 	res = tee_alg_get_digest_size(hash_algo, &hash_size);
 	if (res)
-		return TEE_ERROR_SECURITY;
+		goto err;
 	if (hash_size != shdr->hash_size)
-		return TEE_ERROR_SECURITY;
+		goto err;
 
 	res = crypto_acipher_alloc_rsa_public_key(&key,
 						  ta_pub_key_modulus_size * 8);
 	if (res)
-		return TEE_ERROR_SECURITY;
+		goto err;
 
 	res = crypto_bignum_bin2bn((uint8_t *)&e, sizeof(e), key.e);
 	if (res)
-		goto out;
+		goto err;
 	res = crypto_bignum_bin2bn(ta_pub_key_modulus, ta_pub_key_modulus_size,
 				   key.n);
 	if (res)
-		goto out;
+		goto err;
 
-	res = crypto_acipher_rsassa_verify(shdr->algo, &key, shdr->hash_size,
-					   SHDR_GET_HASH(shdr), shdr->hash_size,
-					   SHDR_GET_SIG(shdr), shdr->sig_size);
+	FTMN_CALL_FUNC(res, &ftmn, FTMN_INCR0,
+		       crypto_acipher_rsassa_verify, shdr->algo, &key,
+		       shdr->hash_size, SHDR_GET_HASH(shdr), shdr->hash_size,
+		       SHDR_GET_SIG(shdr), shdr->sig_size);
+	if (!res) {
+		ftmn_checkpoint(&ftmn, FTMN_INCR0);
+		goto out;
+	}
+	err_incr = 1;
+err:
+	res = TEE_ERROR_SECURITY;
+	FTMN_SET_CHECK_RES_NOT_ZERO(&ftmn, err_incr * FTMN_INCR0, res);
 out:
+	FTMN_CALLEE_DONE_CHECK(&ftmn, FTMN_INCR0, FTMN_STEP_COUNT(2), res);
 	crypto_acipher_free_rsa_public_key(&key);
-	if (res)
-		return TEE_ERROR_SECURITY;
-	return TEE_SUCCESS;
+	return res;
 }
 
 static const struct shdr_subkey_attr *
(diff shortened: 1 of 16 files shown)


ldts:
@jenswi-linaro are there any performance tests you plan to run? I have several boards (xilinx, nxp, stm32...) that I could use if you need help. Not that performance is more critical than security in this case so may be irrelevant.

lib/libutils/ext/include/fault_mitigation.h
```diff
@@ -0,0 +1,687 @@
...
+ *	added to record anything but a zero result.
+ * 8. FAULT.NESTED.CHECK
+ *	The linked calls performed by for instance FTMN_CALL_FUNC() addresses
+ *	this by relying on the called function to update a state in
+ *	struct ftmn_func_arg which is checked when the function has returned.
+ * 11. FAULT.PENALTY
+ *	This is implicit since we're normally trying to protect things post
+ *	boot and booting takes quite some time.
+ *
+ * [1] https://www.riscure.com/uploads/2020/05/Riscure_Whitepaper_Fault_Mitigation_Patterns_final.pdf
```
ldts, line 51: @jenswi-linaro the link is not visible.

jforissier: There is a copy at https://web.archive.org/web/20220616035354/https://www.riscure.com/uploads/2020/05/Riscure_Whitepaper_Fault_Mitigation_Patterns_final.pdf perhaps we could use that link?

jenswi-linaro:
> @jenswi-linaro are there any performance tests you plan to run? I have several boards (xilinx, nxp, stm32...) that I could use if you need help. Not that performance is more critical than security in this case so may be irrelevant.

@ldts, thanks for the offer. I haven't planned any performance testing because I believe the added overhead will not be noticeable combined with signature verification. That said, if you notice anything odd please let me know.

lib/libutils/ext/include/fault_mitigation.h
jenswi-linaro: Thanks, I'll update.

jenswi-linaro:
Rebased to resolve a merge conflict.

lib/libutils/ext/include/fault_mitigation.h
```diff
@@ -0,0 +1,687 @@
...
+ *
+ * This function is called whenever the FTMN function detects an
+ * inconsistency.  An inconsistency is able to occur if the system is
+ * subject to an fault injection attack, in this case doing a panic() isn't
+ * an extreme measure.
+ */
+#ifdef __KERNEL__
+#define FTMN_PANIC()	panic();
+#else
+#define FTMN_PANIC()	TEE_Panic(0);
```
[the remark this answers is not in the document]
etienne-lms: `TEE_ERROR_SECURITY` or `TEE_ERROR_BAD_STATE` could be good candidates.
But i admit it will not be specifically obvious it specifically from the fault injection countermeasure.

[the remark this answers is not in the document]
jforissier: The only occasion when I have found specific values to be helpful in `TEE_Panic()`, is when running negative tests in CI, when the `xtest` output is visible but not the secure world log. Not very applicable to this case I suppose. But FWIW either ways are OK with me (0 or something else).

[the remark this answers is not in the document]
jenswi-linaro: I don't care which we take either. @etienne-lms which do you prefer?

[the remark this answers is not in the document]
etienne-lms: Not willing to track glitch attacks on CI runs? :)
Ok, `0` is fine.

jforissier (commented):
LGTM. I find the modified code a bit difficult to read unfortunately, but I suppose there is not much that can be done. The impact could have been much worse ;)

A couple of comments below, then for the whole:

lib/libutils/ext/include/fault_mitigation.h
```diff
@@ -0,0 +1,687 @@
...
+
+#ifndef __KERNEL__
+extern struct ftmn_func_arg *__ftmn_global_func_arg;
+#endif
+
+static inline struct ftmn_func_arg **__ftmn_get_tsd_func_arg_pp(void)
+{
+#if defined(CFG_CORE_FAULT_MITIGATION) && defined(__KERNEL__)
+	return &thread_get_tsd()->ftmn_arg;
+#elif defined(CFG_CORE_FAULT_MITIGATION)
```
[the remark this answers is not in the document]
jenswi-linaro: Good point.

lib/libutils/ext/include/fault_mitigation.h
```diff
@@ -0,0 +1,687 @@
...
+	FTMN_INCR2 = 7879,
+	FTMN_INCR3 = 7883,
+	FTMN_INCR4 = 7901,
+	FTMN_INCR5 = 7907,
+	FTMN_INCR_RESERVED = 7919,
+};
+
+typedef int (*ftmn_memcmp_t)(const void *p1, const void *p2, size_t nb);
+
+/* The default hash used when xoring the result in struct ftmn_check */
```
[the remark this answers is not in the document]
jenswi-linaro: No, just random.

jenswi-linaro:
Squashed and tags applied.

etienne-lms:
LGTM.

jenswi-linaro:
Tag applied.

core/crypto/signed_hdr.c
```diff
@@ -59,46 +60,56 @@ TEE_Result shdr_verify_signature(const struct shdr *shdr)
...
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t e = TEE_U32_TO_BIG_ENDIAN(ta_pub_key_exponent);
+	struct ftmn ftmn = { };
+	unsigned int err_incr = 2;
 	size_t hash_size = 0;
 	size_t hash_algo = 0;
 
 	if (shdr->magic != SHDR_MAGIC)
-		return TEE_ERROR_SECURITY;
+		goto err;
```
rejoiceliberty, line 69: I changed this to 'goto err' on OP-TEE 3.14 and ran 'xtest -t regression 1008' and encountered an 'unhandled pageable abort' error. After analysis, I think it is because crypto_acipher_alloc_rsa_public_key is not executed but crypto_acipher_free_rsa_public_key is executed. Not sure if the latest version is OK.

etienne-lms: Looking at the (latest) implementation, I think it currently works fine as `key` is 0 initialized and all public key free possible callback handles (ltc, mbedtls, versal, caam and se050 do behave nicely in such context. Yet, I may have missed something and all in one, it looks a bit fragile and I think you're right, `shdr_verify_signature()` exit error sequence should not always call `crypto_acipher_free_rsa_public_key()`.

rejoiceliberty: The error is really because the key was not initialized to 0 on the old version. After initialization, all xtests passed. Thanks!

etienne-lms: For info: looking into `struct rsa_keypair` management, few other places assume it is safe to call `crypto_acipher_free_rsa_public_key()` on a key ref that was zero initialized and not allocated or partially allocated, for example `sw_crypto_acipher_alloc_rsa_keypair()` in _lib/libtomcrypt/rsa.c_ and _lib/libmbedtls/core/rsa.c_. So contrary to above my comment, current latest implementation is fine.

PR #5650, opened 2022-11-15
core: ffa: Allow multiple SPs with same UUID

jellesels-arm:
The FF-A spec doesn't forbid multiple SPs to have the same UUID. This makes it possible to use the FF-A UUID as a identifier for the protocol on top of the FF-A layer.
To achieve this we have to make sure that the FFA_PARTITION_INFO_GET can return more then one endpoint id if we pass a UUID. We also need to change the naming convention of the SPs. This used to be {UUID}.stripped.elf. This has been changed to be
{SP_NAME}_{UUID}.stripped.elf. To make sure we don't use any special characters when creating the name of the SP array we take the hash of the name first.

Build change: https://github.com/OP-TEE/build/pull/603
Trusted Services changes:
https://review.trustedfirmware.org/c/TS/trusted-services/+/16587/
https://review.trustedfirmware.org/c/TS/trusted-services/+/16586

Testing can be done by using:
repo init -u https://github.com/jellesels-arm/manifest.git -m fvp-spmc-test.xml -b multiple_uuid && repo sync -j8 --no-clone-bundle

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -94,23 +94,6 @@ static void set_sp_ctx_ops(struct ts_ctx *ctx)
 	ctx->ops = &sp_ops;
 }
 
-TEE_Result sp_find_session_id(const TEE_UUID *uuid, uint32_t *session_id)
-{
-	struct sp_session *s = NULL;
-
-	TAILQ_FOREACH(s, &open_sp_sessions, link) {
-		if (!memcmp(&s->ts_sess.ctx->uuid, uuid, sizeof(*uuid))) {
-			if (s->state == sp_dead)
-				return TEE_ERROR_TARGET_DEAD;
-
-			*session_id  = s->endpoint_id;
-			return TEE_SUCCESS;
-		}
-	}
-
-	return TEE_ERROR_ITEM_NOT_FOUND;
-}
-
 struct sp_session *sp_get_session(uint32_t session_id)
 {
 	struct sp_session *s = NULL;
@@ -123,18 +106,23 @@ struct sp_session *sp_get_session(uint32_t session_id)
 	return NULL;
 }
 
-TEE_Result sp_partition_info_get_all(struct ffa_partition_info *fpi,
-				     size_t *elem_count)
+TEE_Result sp_partition_info_get(struct ffa_partition_info *fpi,
+				 const TEE_UUID *ffa_uuid, size_t *elem_count)
 {
 	size_t in_count = *elem_count;
 	struct sp_session *s = NULL;
 	size_t count = 0;
 
 	TAILQ_FOREACH(s, &open_sp_sessions, link) {
+		if (ffa_uuid &&
+		    memcmp(&s->ffa_uuid, ffa_uuid, sizeof(*ffa_uuid)))
+			continue;
+
 		if (s->state == sp_dead)
 			continue;
 		if (count < in_count) {
-			spmc_fill_partition_entry(fpi, s->endpoint_id, 1);
+			spmc_fill_partition_entry(fpi, s->endpoint_id,
+						  1);
 			fpi++;
 		}
 		count++;
@@ -184,7 +172,7 @@ static uint16_t new_session_id(struct sp_sessions_head *open_sessions)
 	return id;
 }
 
-static TEE_Result sp_create_ctx(const TEE_UUID *uuid, struct sp_session *s)
+static TEE_Result sp_create_ctx(const TEE_UUID *bin_uuid, struct sp_session *s)
 {
 	TEE_Result res = TEE_SUCCESS;
 	struct sp_ctx *spc = NULL;
@@ -196,7 +184,7 @@ static TEE_Result sp_create_ctx(const TEE_UUID *uuid, struct sp_session *s)
 
 	spc->open_session = s;
 	s->ts_sess.ctx = &spc->ts_ctx;
-	spc->ts_ctx.uuid = *uuid;
+	spc->ts_ctx.uuid = *bin_uuid;
 
 	res = vm_info_init(&spc->uctx, &spc->ts_ctx);
 	if (res)
@@ -212,7 +200,7 @@ err:
 }
 
 static TEE_Result sp_create_session(struct sp_sessions_head *open_sessions,
-				    const TEE_UUID *uuid,
+				    const TEE_UUID *bin_uuid,
 				    struct sp_session **sess)
 {
 	TEE_Result res = TEE_SUCCESS;
@@ -227,8 +215,8 @@ static TEE_Result sp_create_session(struct sp_sessions_head *open_sessions,
 		goto err;
 	}
 
-	DMSG("Loading Secure Partition %pUl", (void *)uuid);
-	res = sp_create_ctx(uuid, s);
+	DMSG("Loading Secure Partition %pUl", (void *)bin_uuid);
+	res = sp_create_ctx(bin_uuid, s);
 	if (res)
 		goto err;
 
@@ -319,16 +307,17 @@ TEE_Result sp_unmap_ffa_regions(struct sp_session *s, struct sp_mem *smem)
 
 static TEE_Result sp_open_session(struct sp_session **sess,
 				  struct sp_sessions_head *open_sessions,
-				  const TEE_UUID *uuid)
+				  const TEE_UUID *ffa_uuid,
+				  const TEE_UUID *bin_uuid)
 {
 	TEE_Result res = TEE_SUCCESS;
 	struct sp_session *s = NULL;
 	struct sp_ctx *ctx = NULL;
 
-	if (!find_secure_partition(uuid))
+	if (!find_secure_partition(bin_uuid))
 		return TEE_ERROR_ITEM_NOT_FOUND;
 
-	res = sp_create_session(open_sessions, uuid, &s);
+	res = sp_create_session(open_sessions, bin_uuid, &s);
 	if (res != TEE_SUCCESS) {
 		DMSG("sp_create_session failed %#"PRIx32, res);
 		return res;
@@ -355,6 +344,8 @@ static TEE_Result sp_open_session(struct sp_session **sess,
 	s->state = sp_idle;
 	s->caller_id = 0;
 	sp_init_set_registers(ctx);
+
+	memcpy(&s->ffa_uuid, ffa_uuid, sizeof(*ffa_uuid));
 	ts_pop_current_session();
 
 	return TEE_SUCCESS;
@@ -410,11 +401,10 @@ static TEE_Result sp_dt_get_uuid(const void *fdt, int node,
 	return TEE_SUCCESS;
 }
 
-static TEE_Result check_fdt(const void * const fdt, const TEE_UUID *uuid)
+static TEE_Result fdt_get_uuid(const void * const fdt, TEE_UUID *uuid)
 {
 	const struct fdt_property *description = NULL;
 	int description_name_len = 0;
-	TEE_UUID fdt_uuid = { };
 
 	if (fdt_node_check_compatible(fdt, 0, "arm,ffa-manifest-1.0")) {
 		EMSG("Failed loading SP, manifest not found");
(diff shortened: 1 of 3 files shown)


jforissier:
> The FF-A spec doesn't forbid multiple SPs to have the same UUID. 

Admittedly I don't know much about SPs... but doesn't that sounds like a recipe for future troubles? Which impact will it have on common code? (SP vs TA vs TS).

jellesels-arm:
> > The FF-A spec doesn't forbid multiple SPs to have the same UUID.
> 
> Admittedly I don't know much about SPs... but doesn't that sounds like a recipe for future troubles? Which impact will it have on common code? (SP vs TA vs TS).

It doesn't have any impact on the common code. TA and TS are not impacted. Internally we only use the SP UUID with the `FFA_PARTITION_INFO_GET`, which will return the endpoint id for the corresponding SP if an UUID is passed. Otherwise we only used endpoint id's to identify the SPs.

jellesels-arm:
@jbech-linaro I will respond here to your question to have the discussion about this topic at the same place. 
> Perhaps this has been discussed in the past, but why do we want to allow multiple SP's to have the same UUID in the first place? It's sounds wrong to me, but still I can see that this might make it easier from a (non-secure) user client point of view. I.e, a client sees "an interface" that belongs to a certain UUID, but under the hood on the secure side it has been split up into different SPs.

This is one of the reason we want to implement it. We want to use the UUID as a identified for the underlying protocol. This would make it easier for the integrator and normal world applications. They don't need to know which SPs are loaded in the system but only which protocol they understand. Depending on the UUID they will select the underlying protocol to communicate with the SP. 
The underlying protocol will then take care off  figuring out how to discover the SPs.  Trusted Services  for example are developed in such a way that multiple Trusted Services can run in one or in Multiple SPs. It has a discovery mechanism on top of FF-A which allows use to discover the Trusted Services services. The protocol itself then figures out if this service lives in the same or a different SP (as well as which SP).

jenswi-linaro:
I'm sorry, I don't get it. If two SPs have the same UUID, how will the client know which to communicate with? And also, what's the point with the other SP that the client chooses not to communicate with?

jellesels-arm:
> I'm sorry, I don't get it. If two SPs have the same UUID, how will the client know which to communicate with? And also, what's the point with the other SP that the client chooses not to communicate with?
The client only needs to know the endpoint id of a SP to be able to communicate with it. Endpoint ids of SP are mapped at boot time, they don't change. 
There are 2 ways the client can figure out  the endpoint of a SP:
-The endpoint id can be hard coded inside the manifest of the SP.
-Trusted Services have a protocol where they can ask each SPs what there services is. They will will call FFA_GET_PARTION_INFO with the protocol UUID. This will return all the endpoint ids of the SPs which understand the Trusted Services protocol. The protocol itself will call each SP to see what services it has. Depending on this response the client knows which service lives in which SP. 

The client can still communicate with both SPs. 
image

jenswi-linaro:
OK, so the UUID isn't used to identify some SPs any longer. Instead, it tells the protocol the SP understands and the client is supposed to ask all SPs that speak this protocol if it might have the service the client is looking for.  Even if this isn't specifically prohibited by the FF-A spec it still seems you're pushing the limit. Have you checked with @achingupta that this indeed is supposed to be permitted and not a loophole? You'll lose the ability to identify an SP without first exchanging messages with it.

If I understand this correctly you're trying to solve the problem of finding the endpoint ID of a particular service and using only a UUID isn't flexible enough. Another option that comes to mind is to have a special service lookup SP with a well-known UUID.

However, the changes in OP-TEE are not intrusive so my only concern is if this is in the spirit of the FF-A spec or not.

jellesels-arm:
> 

Yes, you are correct about this. We had some internal discussions about this with @achingupta and this fits into the FF-A spec.
This change also doesn't forbid people to use the UUID in the more traditional way where every UUID is matched to 1 SP.

achingupta:
@jenswi-linaro 

Just to add to @jellesels-arm's explanation...

Yes. The spec supports this to primarily cater for multiple EFI Standalone MM SPs. Each SP is a collection of EFI Standalone MM drivers (services) that can be accessed over the MM communication protocol. We expect the UUID to identify the MM communication protocol. Hence, EFI runtime can use this UUID to determine which SPs are StMM SPs. 

The next step is to determines which services are implemented in which SP. Admittedly, we need to define messages for this in the same vein as the TS services messages. But the first step was to be able to identify the SPs.

I hope this seems reasonable.

jenswi-linaro:
Thanks for the explanations. I had a chat with @jforissier today and we agreed that as long as this is fine from an FF-A spec point of view and doesn't limit or break anything else it's OK from an OP-TEE point of view too.

However, there are a few things to keep in mind. Until now we've assumed that a UUID is unique for a binary, at least in its domain (SP in this case). `scripts/symbolize.py` may need some patching to handle SPs with the same UUID. If we even enable dynamic linking of SPs we have to be careful that there are no clashes of UUID with libraries.

jellesels-arm:
> However, there are a few things to keep in mind. Until now we've assumed that a UUID is unique for a binary, at least in its domain (SP in this case). `scripts/symbolize.py` may need some patching to handle SPs with the same UUID. If we even enable dynamic linking of SPs we have to be careful that there are no clashes of UUID with libraries.

I will have a look at the `scripts/symbolize.py`  script.

jenswi-linaro:
I'm still a bit uneasy about having multiple SPs with the same UUID. With that, there's no good way of identifying the binary file. I suppose the problem could be worked around or avoided, but we would diverge even further from how TAs are treated and I'm not sure that's a good thing.

Would it make sense to give an SP an alias UUID. FFA_PARTITION_INFO_GET could present alias UUIDs in addition to the real UUIDs. I guess it could be configurable if the real UUID should remain hidden if there's an alias UUID.

jellesels-arm:
> I'm still a bit uneasy about having multiple SPs with the same UUID. With that, there's no good way of identifying the binary file. I suppose the problem could be worked around or avoided, but we would diverge even further from how TAs are treated and I'm not sure that's a good thing.
> 

We do encapsulate the SP manifest together with the SP in the binary. We could add some extra information inside the SP manifest file.
Or are you more specifically taking about the binary name? I get around this by adding the hash of the SP name to the file name.

> Would it make sense to give an SP an alias UUID. FFA_PARTITION_INFO_GET could present alias UUIDs in addition to the real UUIDs. I guess it could be configurable if the real UUID should remain hidden if there's an alias UUID.

This would require the FFA_PARTITION_INFO_GET message to be changed inside the FF-A spec. 
There was some talk about allowing a SP to have more then 1 UUID linked to it. I don't fully remember the details of this to be honest. If this is aloud we could have a convention where we use one UUID for the protocol and one for the SP itself. 

@achingupta any thoughts on this?

jellesels-arm:
There are some other examples where the FF-A spec allows multiple SPs to have the same UUID:

- Hafnium for example can run OP-TEE as multiple independent S-EL1 SP. This would have the same behavior where multiple SPs will be identified by the same UUID.

- This change is also needed if someone wants to run the same SP multiple times. TAs can be run multiple times as they are dynamically loaded. However this is not the case for SPs. They are loaded at boot and can't be loaded afterwards. To be able to load a SP multiple times we need to add the SP multiple times to the image.

jforissier:
> To be able to load a SP multiple times we need to add the SP multiple times to the image

Does this result in the image file having several physical copies of the SP binary?

jellesels-arm:
> > To be able to load a SP multiple times we need to add the SP multiple times to the image
> 
> Does this result in the image file having several physical copies of the SP binary?

At the moment it does. However I don't think the FF-A spec disallow only having one image for multiple SPs. We could add this in the future.

gyuri-szing:
To sum our view:
"The meaning of FF-A UUID."
The FF-A UUID identities SPs which are VM like compartments. An SP can run an entire trusted OS, and thus the UUID cannot be bound to the "functionality" or "features" as this may depend on the S-EL0 workloads the TOS is running. Even in case of a static configuration of the set of workloads, this could be unique for each deployment. Thus, the NWd clients would need to maintain a list of compatible UUIDs. This is error prone and does not scale (e.g. backwards compatibility issues).
Thus, the best is if the FF-A UUID identifies the "protocol", and "generic behavior" of the SP and further discovery is solved in a higher layer. In this case the UUID only tells how a client can talk to the SP. I think this is already true for S-EL1 SPs. E.g. when having multiple OP-TEE OS images running, each is having the same FF-A UUID.

I think the same principles need to apply to S-EL0 SPs, independent of the fact, these are running in the application layer. Otherwise, the clients may need to be aware of the SWd SW structure as the client might need to behave different depending on if a service is deployed to an S-EL0 or and S-EL1 SP. (StMM might be an example for that.)

Some S-EL0 and TS specific examples:
- In TS, the service to SP mapping can vary. E.g. the Corstone-1000 platform deploys all services to a single proxy SP. Service to SP mapping in S-EL0 SPs depends more on peripheral sharing problems, which may push towards a set-up where services are deployed to a single SP. This is due to the micro kernel architecture mandated by the nature of S-EL0 SPs
- The VM to SP mapping might not always be n:1. E.g. if I wish to separate VM assets stronger, then I might want to have 1:1 mapping, and that would mean running multiple instances of the same SP. In that case multiple SPs would have the same UUID.

"OP-TEE "machinery" challenges."
The other topic is OP-TEE "machinery", meaning the build system, tooling and perhaps how the UUID is used during boot.
There are challenges as the "machinery" is shared with TAs, but trying to keep SP design in par with TA design just because of that is IMHO not correct. If the best choice is to extend the machinery to support handling both TAs and SPs, or to it is to split is to be decided. My gut feeling says extending might be better.

jenswi-linaro:
I believe the solution is to use two UUIDs, one to identify the binary making up the SP and one to present with `FFA_PARTITION_INFO_GET`.

jellesels-arm:
> 

I changed the patch so we have 2 UUID. 

1. The FF-A uuid which is used in the manifest file and with FFA_PARTITION_INFO_GET.
2. Binary uuid. Is used to identify the SP inside OP-TEE and the build system
The two UUID's can be the same if there is only one SP using the FF-A uuid.

Manifest file was updated so it can be used to test everything.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -123,21 +106,25 @@ struct sp_session *sp_get_session(uint32_t session_id)
...
 	struct sp_session *s = NULL;
 	size_t count = 0;
 
 	TAILQ_FOREACH(s, &open_sp_sessions, link) {
-		if (s->state == sp_dead)
-			continue;
-		if (count < in_count) {
-			spmc_fill_partition_entry(fpi, s->endpoint_id, 1);
-			fpi++;
+		if (!ffa_uuid || !memcmp(&s->ffa_uuid, ffa_uuid,
```
jenswi-linaro, line 117: You could do:
``` C
if (ffa_uuid &&
    memcmp(&s->ffa_uuid, ffa_uuid, sizeof(*ffa_uuid))
         continue;
```
instead and leave the lines below unchanged, that should make the code a bit easier to read.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -426,16 +415,11 @@ static TEE_Result check_fdt(const void * const fdt, const TEE_UUID *uuid)
 	if (description)
 		DMSG("Loading SP: %s", description->data);
 
-	if (sp_dt_get_uuid(fdt, 0, "uuid", &fdt_uuid)) {
+	if (!uuid ||  sp_dt_get_uuid(fdt, 0, "uuid", uuid)) {
```
jenswi-linaro, line 418: Is it meaningful to supply `uuid` as NULL? I guess not since `sp_init_uuid()` always gives a valid pointer.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -410,11 +400,10 @@ static TEE_Result sp_dt_get_uuid(const void *fdt, int node,
 	return TEE_SUCCESS;
 }
 
-static TEE_Result check_fdt(const void * const fdt, const TEE_UUID *uuid)
+static TEE_Result check_fdt(const void * const fdt, TEE_UUID *uuid)
```
jenswi-linaro, line 403: This function does a bit more than "checking" the fdt now and should preferrably have a name to reflect that.

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -404,24 +404,26 @@ void spmc_handle_partition_info_get(struct thread_smc_args *args,
...
 
 		uuid_array[0] = args->a1;
 		uuid_array[1] = args->a2;
 		uuid_array[2] = args->a3;
 		uuid_array[3] = args->a4;
 		tee_uuid_from_octets(&uuid, (uint8_t *)uuid_array);
 
-		res = sp_find_session_id(&uuid, &endpoint_id);
+		res = sp_partition_info_get(fpi, &uuid, &count);
+		rc = count;
```
jenswi-linaro, line 421: It would make more sense to assign this after `res` has been checked.

jellesels-arm: Ack

core/arch/arm/kernel/secure_partition.c
jellesels-arm: Ack

core/arch/arm/kernel/secure_partition.c
jellesels-arm: Ack

core/arch/arm/kernel/secure_partition.c
jellesels-arm: Ack

PR #5653, opened 2022-11-16
make X25519 & Ed25519 available for mbedtls configurations

sa-kib:
Make LTC implementations of X25519 & Ed25519 algorithms available when crypto library is chosen to be mbedtls (which yet lacks support for these algorithms).

### DIFF
--- a/core/crypto.mk
+++ b/core/crypto.mk
@@ -47,11 +47,8 @@ CFG_CRYPTO_ECC ?= y
 CFG_CRYPTO_SM2_PKE ?= y
 CFG_CRYPTO_SM2_DSA ?= y
 CFG_CRYPTO_SM2_KEP ?= y
-# X25519 and Ed25519 are only supported by libtomcrypt
-ifeq ($(CFG_CRYPTOLIB_NAME),tomcrypt)
 CFG_CRYPTO_ED25519 ?= y
 CFG_CRYPTO_X25519 ?= y
-endif
 
 # Authenticated encryption
 CFG_CRYPTO_CCM ?= y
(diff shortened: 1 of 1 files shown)


jforissier:
> Make LTC implementations of X25519 & Ed25519 algorithms available when crypto library is chosen to be mbedtls (which yet lacks support for these algorithms).

That's fine (`Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`), but wouldn't it make more sense to have this commit **after** some other commit actually adding/enabling support in MBed TLS ?

sa-kib:
> > Make LTC implementations of X25519 & Ed25519 algorithms available when crypto library is chosen to be mbedtls (which yet lacks support for these algorithms).
> 
> That's fine (`Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`), but wouldn't it make more sense to have this commit **after** some other commit actually adding/enabling support in MBed TLS ?

@jforissier this PR is actually a follow up of this [discussion](https://github.com/OP-TEE/optee_os/pull/5486#pullrequestreview-1075526662). Probably the commit description is a bit misleading, so I changed it for more clarity.

jforissier:
@sa-kib thanks for the update but I still find the commit description unclear. I'd like to make sure I understand well: this patch allows LibTomCrypt to be used for X25519/Ed25519 when MBedTLS is selected, correct? In this case how about:
```
crypto: fallback to LibTomCrypt for X25519/Ed25519 when CFG_CRYPTOLIB_NAME=mbedtls

Mbed TLS currently does not support X25519/Ed25519. This commit allows the
use of LibTomCrypt for these algorithms when CFG_CRYPTO_ED25519=y and/or
CFG_CRYPTO_X25519=y, and CFG_CRYPTOLIB_NAME=mbedtls.
```

sa-kib:
> @sa-kib thanks for the update but I still find the commit description unclear. I'd like to make sure I understand well: this patch allows LibTomCrypt to be used for X25519/Ed25519 when MBedTLS is selected, correct? In this case how about:
> 
> ```
> crypto: fallback to LibTomCrypt for X25519/Ed25519 when CFG_CRYPTOLIB_NAME=mbedtls
> 
> Mbed TLS currently does not support X25519/Ed25519. This commit allows the
> use of LibTomCrypt for these algorithms when CFG_CRYPTO_ED25519=y and/or
> CFG_CRYPTO_X25519=y, and CFG_CRYPTOLIB_NAME=mbedtls.
> ```

I think your version describes what's going on very nice, I shall use it (only added word 'implementation' to clarify that only a part of LibTomCrypt gets enabled)

PR #5657, opened 2022-11-18
core: arm: stmm: use mempool to decompress stmm image

etienne-lms:
Changes StMM management to have zlib using default mempool to allocate buffers for StMM image decompression. This is useful as the process can require buffer of several kilobytes.

### DIFF
--- a/core/arch/arm/kernel/stmm_sp.c
+++ b/core/arch/arm/kernel/stmm_sp.c
@@ -11,6 +11,7 @@
 #include <kernel/stmm_sp.h>
 #include <kernel/thread_private.h>
 #include <kernel/user_mode_ctx.h>
+#include <mempool.h>
 #include <mm/fobj.h>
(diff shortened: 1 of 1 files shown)


etienne-lms:
tag applied
I have few other places, i wonder if it's worth to cover also, [temp link here](https://github.com/etienne-lms/optee_os/commit/4c29a4ae41d40174b19ffc99ba7a566f85ef1cc4).

jenswi-linaro:
> tag applied I have few other places, i wonder if it's worth to cover also, [temp link here](https://github.com/etienne-lms/optee_os/commit/4c29a4ae41d40174b19ffc99ba7a566f85ef1cc4).

It is tempting, but as @ldts found out mempool is tricky to use if you happen to hold another lock. If we chose to use mempool allocations in RPMB then we must make sure that we don't hold for instance `tee_ta_mutex`.

PR #5658, opened 2022-11-18
core: arm: mm: introduce CFG_MMU_XLAT_TABLES

etienne-lms:
Adds configuration switch CFG_MMU_XLAT_TABLES to set the number of translation tables statically allocated in OP-TEE core.

This switch is useful for example when one builds OP-TEE with large embedded (early) TAs which may require custom TEE RAM virtual address range size (with CFG_TEE_RAM_VA_SIZE) and a custom number of MMU translation tables.

Legacy MAX_XLAT_TABLES is still considered as some platform_config.h do define that macro, but CFG_MMU_XLAT_TABLES takes precedence when defined.


jenswi-linaro:
How about fixing the two platforms which use `MAX_XLAT_TABLES`? It should make everything much easier.

etienne-lms:
Ok, i'll push an update.

etienne-lms:
Ok, that would be better indeed.

PR #5659, opened 2022-11-18
[RFC] StMM: 32bit mode requires cpu usr_lr to be preserved

etienne-lms:
This P-R tries to fix an issue in StMM management in 32bit mode where CPU register USR_LR is not preserved in StMM execution context.

### DIFF
--- a/core/arch/arm/include/kernel/thread_private_arch.h
+++ b/core/arch/arm/include/kernel/thread_private_arch.h
@@ -181,6 +181,8 @@ void thread_set_fiq_sp(vaddr_t sp);
 
 /* Read usr_sp banked CPU register */
 uint32_t thread_get_usr_sp(void);
+uint32_t thread_get_usr_lr(void);
+void thread_set_usr_lr(uint32_t usr_lr);
 #endif /*ARM32*/
 
 void thread_alloc_and_run(uint32_t a0, uint32_t a1, uint32_t a2, uint32_t a3,
--- a/core/arch/arm/kernel/stmm_sp.c
+++ b/core/arch/arm/kernel/stmm_sp.c
@@ -111,7 +111,18 @@ static TEE_Result stmm_enter_user_mode(struct stmm_ctx *spc)
 	exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
 	cntkctl = read_cntkctl();
 	write_cntkctl(cntkctl | CNTKCTL_PL0PCTEN);
+
+#ifdef ARM32
+	/* Handle usr_lr in place of __thread_enter_user_mode() */
+	thread_set_usr_lr(spc->regs.usr_lr);
+#endif
+
 	__thread_enter_user_mode(&spc->regs, &panicked, &panic_code);
+
+#ifdef ARM32
+	spc->regs.usr_lr = thread_get_usr_lr();
+#endif
+
 	write_cntkctl(cntkctl);
 	thread_unmask_exceptions(exceptions);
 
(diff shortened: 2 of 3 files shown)


jenswi-linaro:
When is this needed? Is OP-TEE breaking some calling convention or is it StMM that expects a bit too much? Any chance of doing this internally in StMM instead?

etienne-lms:
When a client invokes an OP-TEE GPD TA, it is entered through specific `__utee_entry()`  entry point, hence not relying  on CPU user mode LR value.
On the other hand, StMM behaves differently as it has a specific entrypoint at init but afterward, when entrered back, core is expected to retrun to where StMM left at previous invocation. In 32bit, **urs_lr** value at previous invocation completion must be preserved for StMM next invocation.

Note that in the above, i don't talk about tee thread suspend/resume scheme that already fully saves/restores CPU registers for StMM, as expected.

jenswi-linaro:
So StMM expects `svc` to preserve `lr`. Any chance of wrapping the `svc`'s or that's not feasible?

etienne-lms:
Saving `usr_lr` to stmm context can move from  `stmm_enter_user_mode()` to `svc`'s handler we still need to set `usr_lr` in CPU once `stmm_enter_user_mode()` locked the thread, e.g.:
```patch
static TEE_Result stmm_enter_user_mode(struct stmm_ctx *spc)
{
	uint32_t exceptions = 0;
	uint32_t panic_code = 0;
	uint32_t panicked = 0;
	uint64_t cntkctl = 0;

	exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
	cntkctl = read_cntkctl();
	write_cntkctl(cntkctl | CNTKCTL_PL0PCTEN);

#ifdef ARM32
	/* Handle usr_lr in place of __thread_enter_user_mode() */
	thread_set_usr_lr(spc->regs.usr_lr);
#endif

	__thread_enter_user_mode(&spc->regs, &panicked, &panic_code);

	write_cntkctl(cntkctl);
	thread_unmask_exceptions(exceptions);
	(...)
```

jenswi-linaro:
> Saving `usr_lr` to stmm context can move from `stmm_enter_user_mode()` to `svc`'s handler we still need to set `usr_lr` in CPU once `stmm_enter_user_mode()` locked the thread, e.g.:

I'm sorry for being unclear, I mean in the StMM code. I assume that StMM already have OP-TEE adaptations so saving lr on the stack before an `svc` shouldn't be much different from calling another function.

etienne-lms:
StMM is based on edk2. Edk2 implements a generic arm `ArmCallSvc()` function from its _ArmSvcLib_ library.
Here is the 32bit [`ArmCallSvc()` implementation](https://github.com/tianocore/edk2/blob/edk2-stable202208/ArmPkg/Library/ArmSvcLib/Arm/ArmSvc.S#L13). 
StMM completes an invocation by calling `ArmCallSvc()` from from [`DelegatedEventLoop()`](https://github.com/tianocore/edk2/blob/edk2-stable202208/StandaloneMmPkg/Library/StandaloneMmCoreEntryPoint/Arm/StandaloneMmCoreEntryPoint.c#L122). I see no clean way to address that but modifying `ArmCallSvc()` but I doubt edk2 maintainers will accept such a modification because of an OP-TEE issue.

jenswi-linaro:
OK, thanks for the details.

etienne-lms:
Thanks for the review.
Fixup commit squashed and tag applied.

PR #5660, opened 2022-11-18
Add support for compiler stack protector

maximus64:
This change add support for CFG_CORE_STACK_PROTECTOR and CFG_TA_STACK_PROTECTOR. This flag enable the compiler stack overflow protection feature -fstack-protector-strong and also generate random stack canary value on kernel boot and TA entry.

Weak function plat_get_random_stack_canary() should be override by platform to provide random stack canary value for the core kernel.

Signed-off-by: Khoa Hoang <admin@khoahoang.com>

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -113,6 +113,31 @@ __weak unsigned long plat_get_aslr_seed(void)
 	return 0;
 }
 
+#if defined(_CFG_CORE_STACK_PROTECTOR)
+/* Generate random stack canary value on boot up */
+__weak uintptr_t plat_get_random_stack_canary(void)
+{
+	uintptr_t canary = 0xbaaaad00;
+	TEE_Result ret = TEE_ERROR_GENERIC;
+
+	/*
+	 * With virtualization the RNG is not initialized in Nexus core.
+	 * Need to override with platform specific implementation.
+	 */
+	if (IS_ENABLED(CFG_VIRTUALIZATION)) {
+		IMSG("WARNING: Using fixed value for stack canary");
+		return canary;
+	}
+
+	ret = crypto_rng_read(&canary, sizeof(canary));
+	if (ret != TEE_SUCCESS)
+		panic("Failed to generate random stack canary");
+
+	/* Leave null byte in canary to prevent string base exploit */
+	return canary & ~0xffUL;
+}
+#endif /*_CFG_CORE_STACK_PROTECTOR*/
+
 /*
  * This function is called as a guard after each smc call which is not
  * supposed to return.
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -350,6 +350,13 @@ clear_nex_bss:
 	mov	sp, x21
 #endif
 
+#ifdef _CFG_CORE_STACK_PROTECTOR
+	/* Update stack canary value */
+	bl	plat_get_random_stack_canary
+	adr_l	x5, __stack_chk_guard
+	str	x0, [x5]
+#endif
+
 	/*
 	 * In case we've touched memory that secondary CPUs will use before
 	 * they have turned on their D-cache, clean and invalidate the
--- a/core/include/kernel/boot.h
+++ b/core/include/kernel/boot.h
@@ -69,6 +69,9 @@ void plat_cpu_reset_early(void);
 void plat_primary_init_early(void);
 unsigned long plat_get_aslr_seed(void);
 unsigned long plat_get_freq(void);
+#if defined(_CFG_CORE_STACK_PROTECTOR)
+uintptr_t plat_get_random_stack_canary(void);
+#endif
 void arm_cl2_config(vaddr_t pl310);
 void arm_cl2_enable(vaddr_t pl310);
 
--- a/lib/libutils/isoc/stack_check.c
+++ b/lib/libutils/isoc/stack_check.c
@@ -3,13 +3,27 @@
  * Copyright (c) 2014, STMicroelectronics International N.V.
  */
 #include <compiler.h>
+#include <trace.h>
+
+#if defined(__KERNEL__)
+# include <kernel/panic.h>
+# define PANIC() panic()
+#elif defined(__LDELF__)
+# include <ldelf_syscalls.h>
+# define PANIC() _ldelf_panic(2)
+#else
+# include <utee_syscalls.h>
+# define PANIC() _utee_panic(TEE_ERROR_OVERFLOW)
+#endif
+
 void *__stack_chk_guard __nex_data = (void *)0x00000aff;
 
 void __attribute__((noreturn)) __stack_chk_fail(void);
 
 void __stack_chk_fail(void)
 {
+	EMSG_RAW("stack smashing detected");
 	while (1)
-		;
+		PANIC();
 }
 
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -313,6 +313,37 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
 # corruption vulnerabilities more difficult.
 CFG_CORE_ASLR ?= y
 
+# Stack Protection for TEE Core
+# This flag enables the compiler stack protection mechanisms -fstack-protector.
+# It will check the stack canary value before returning from a function to
+# prevent buffer overflow attacks. Stack protector canary logic will be added
+# for vulnerable functions that contain:
+# - A character array larger than 8 bytes.
+# - An 8-bit integer array larger than 8 bytes.
+# - A call to alloca() with either a variable size or a constant size bigger
+#   than 8 bytes.
+CFG_CORE_STACK_PROTECTOR ?= n
+# This enable stack protector flag -fstack-protector-strong. Stack protector
+# canary logic will be added for vulnerable functions that contain:
+# - An array of any size and type.
+# - A call to alloca().
+# - A local variable that has its address taken.
+CFG_CORE_STACK_PROTECTOR_STRONG ?= y
+# This enable stack protector flag -fstack-protector-all. Stack protector canary
+# logic will be added to all functions regardless of their vulnerability.
+CFG_CORE_STACK_PROTECTOR_ALL ?= n
+# Stack Protection for TA
+CFG_TA_STACK_PROTECTOR ?= n
+CFG_TA_STACK_PROTECTOR_STRONG ?= y
+CFG_TA_STACK_PROTECTOR_ALL ?= n
+
+_CFG_CORE_STACK_PROTECTOR := $(call cfg-one-enabled, CFG_CORE_STACK_PROTECTOR \
+						     CFG_CORE_STACK_PROTECTOR_STRONG \
+						     CFG_CORE_STACK_PROTECTOR_ALL)
+_CFG_TA_STACK_PROTECTOR := $(call cfg-one-enabled, CFG_TA_STACK_PROTECTOR \
+						   CFG_TA_STACK_PROTECTOR_STRONG \
+						   CFG_TA_STACK_PROTECTOR_ALL)
+
 # Load user TAs from the REE filesystem via tee-supplicant
 CFG_REE_FS_TA ?= y
 
--- a/ta/arch/arm/user_ta_header.c
+++ b/ta/arch/arm/user_ta_header.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2014, STMicroelectronics International N.V.
  */
 #include <compiler.h>
+#include <config.h>
 #include <malloc.h>
 #include <tee_ta_api.h>
 #include <tee_internal_api_extensions.h>
@@ -10,6 +11,8 @@
 #include <user_ta_header.h>
 #include <user_ta_header_defines.h>
 #include <utee_syscalls.h>
+
+extern void *__stack_chk_guard;
 
 int trace_level = TRACE_LEVEL;
 
@@ -46,7 +49,25 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
 					struct utee_params *up,
 					unsigned long cmd_id)
 {
-	TEE_Result res = __utee_entry(func, session_id, up, cmd_id);
+	static bool stack_canary_inited;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (IS_ENABLED(_CFG_TA_STACK_PROTECTOR) && !stack_canary_inited) {
+		uintptr_t canary = 0;
+
+		res = _utee_cryp_random_number_generate(&canary,
+							sizeof(canary));
+		if (res != TEE_SUCCESS)
+			_utee_return(res);
+
+		/* Leave null byte in canary to prevent string base exploit */
+		canary &= ~0xffUL;
+
+		__stack_chk_guard = (void *)canary;
+		stack_canary_inited = true;
+	}
+
+	res = __utee_entry(func, session_id, up, cmd_id);
 
 #if defined(CFG_FTRACE_SUPPORT)
 	/*
--- a/ta/ta.mk
+++ b/ta/ta.mk
@@ -12,6 +12,11 @@ include mk/$(COMPILER_$(sm)).mk
 # Config flags from mk/config.mk
 #
 
+ta-stackp-cflags-$(CFG_TA_STACK_PROTECTOR) := -fstack-protector
+ta-stackp-cflags-$(CFG_TA_STACK_PROTECTOR_STRONG) := -fstack-protector-strong
+ta-stackp-cflags-$(CFG_TA_STACK_PROTECTOR_ALL) := -fstack-protector-all
+$(sm)-platform-cflags += $(ta-stackp-cflags-y)
+
 ifeq ($(CFG_TA_MBEDTLS_SELF_TEST),y)
 $(sm)-platform-cppflags += -DMBEDTLS_SELF_TEST
 endif
--- a/core/arch/arm/kernel/entry_a32.S
+++ b/core/arch/arm/kernel/entry_a32.S
@@ -567,6 +567,13 @@ shadow_stack_access_ok:
 	mov	sp, r9
 #endif
 
+#ifdef _CFG_CORE_STACK_PROTECTOR
+	/* Update stack canary value */
+	bl	plat_get_random_stack_canary
+	ldr	r1, =__stack_chk_guard
+	str	r0, [r1]
+#endif
+
 	/*
 	 * In case we've touched memory that secondary CPUs will use before
 	 * they have turned on their D-cache, clean and invalidate the
--- a/core/core.mk
+++ b/core/core.mk
@@ -24,6 +24,12 @@ cppflags$(sm)	+= -include $(conf-file)
 cppflags$(sm)	+= -I$(out-dir)/core/include
 cppflags$(sm)	+= $(core-platform-cppflags)
 cflags$(sm)	+= $(core-platform-cflags)
+
+core-stackp-cflags-$(CFG_CORE_STACK_PROTECTOR) := -fstack-protector
+core-stackp-cflags-$(CFG_CORE_STACK_PROTECTOR_STRONG) := -fstack-protector-strong
+core-stackp-cflags-$(CFG_CORE_STACK_PROTECTOR_ALL) := -fstack-protector-all
+cflags$(sm)	+= $(core-stackp-cflags-y)
+
 ifeq ($(CFG_CORE_SANITIZE_UNDEFINED),y)
 cflags$(sm)	+= -fsanitize=undefined
 endif


core/arch/arm/kernel/entry_a64.S
```diff
@@ -265,6 +265,13 @@ clear_nex_bss:
 	/* Enable Console */
 	bl	console_init
 
+#ifdef CFG_CORE_STACK_PROTECTOR
```
jenswi-linaro, line 268: This should probably be moved just before we're cleaning the cache to let platforms use higher level functions if possible. Same for the 32-bit code.

core/arch/arm/kernel/boot.c
```diff
@@ -113,6 +113,14 @@ __weak unsigned long plat_get_aslr_seed(void)
 	return 0;
 }
 
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak unsigned long plat_get_random_stack_canary(void)
```
jenswi-linaro, line 117: I wonder if this function shouldn't be mandatory for platforms to implement if configured with `CFG_CORE_STACK_PROTECTOR`. Without a random value we could just as well use the default value in `lib/libutils/isoc/stack_check.c`.

ta/arch/arm/user_ta_header.c
```diff
@@ -46,7 +48,21 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
...
+	TEE_Result res;
+
+#if defined(CFG_TA_STACK_PROTECTOR)
+	uintptr_t canary;
+
+	res = _utee_cryp_random_number_generate(&canary, sizeof(canary));
+	if (res != TEE_SUCCESS)
+		goto out;
+
+	canary &= ~0xffUL;
```
jenswi-linaro, line 60: Why? This isn't obvious.

ta/arch/arm/user_ta_header.c
```diff
@@ -46,7 +48,21 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
 					struct utee_params *up,
 					unsigned long cmd_id)
 {
-	TEE_Result res = __utee_entry(func, session_id, up, cmd_id);
+	TEE_Result res;
+
+#if defined(CFG_TA_STACK_PROTECTOR)
+	uintptr_t canary;
+
+	res = _utee_cryp_random_number_generate(&canary, sizeof(canary));
```
jenswi-linaro, line 56: Doing this each time we enter a TA might be a bit expensive.

ta/arch/arm/user_ta_header.c
maximus64: This is to clearing a byte in the canary value to prevent string base overflow attack e.g scanf, strcpy. Because it doesn't let you put a null byte in middle of the string. In some rare case, it also prevent leaking the canary value when print non-terminated string from stack. This reduce the entropy but I think 24bits is still enough for 32bits platforms. I will put comment in the patch to make it clear.

core/arch/arm/kernel/entry_a64.S
maximus64: I'm a bit confuse here. Not sure how moving it before the cache clean (line 253) would allow platforms to use higher level functions. Maybe you mean move it after we enable the MMU and re-initialize the serial console at line 322?

jenswi-linaro: I was thinking of line 352.

ta/arch/arm/user_ta_header.c
maximus64: Good point. I added check to only set the stack canary once

core/arch/arm/kernel/entry_a64.S
maximus64: I see that make sense now. I have moved this in the updated patch.

core/arch/arm/kernel/boot.c
maximus64: Since we move the set random canary call to after boot_init_primary_late() we can use the crypto_rng_read() to generate stack canary value and doesn't need platform specific implementation. Fixed in updated patch

maximus64: This change currently broke the virtualization due to boot flow differences. `init_tee_runtime()` doesn't get call in `boot_init_primary_late()` when virtualization is enable.

ta/arch/arm/user_ta_header.c
jenswi-linaro: Makes sense, thanks for the explanation.

ta/arch/arm/user_ta_header.c
```diff
@@ -46,7 +48,28 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
 					struct utee_params *up,
 					unsigned long cmd_id)
 {
-	TEE_Result res = __utee_entry(func, session_id, up, cmd_id);
+	TEE_Result res;
+
+#if defined(CFG_TA_STACK_PROTECTOR)
```
jenswi-linaro, line 53: Please use `if (IS_ENABLED(CFG_TA_STACK_PROTECTOR)) {` instead.

ta/arch/arm/user_ta_header.c
```diff
@@ -46,7 +48,28 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
...
+
+		res = _utee_cryp_random_number_generate(&canary,
+							sizeof(canary));
+		if (res != TEE_SUCCESS)
+			_utee_return(res);
+
+		/* leave null byte in canary to prevent string base exploit */
+		canary &= ~0xffUL;
+
+		__stack_chk_guard = (void *)canary;
```
jenswi-linaro, line 67: If we set this here we must make sure that this function, `__ta_entry()`, doesn't use the guard or stuff will break. How is this done?

core/arch/arm/kernel/boot.c
```diff
@@ -1473,6 +1473,31 @@ struct ns_entry_context *boot_core_hpen(void)
 }
 #endif
 
+#if defined(CFG_CORE_STACK_PROTECTOR)
+/* generate random stack canary value on boot up */
+__weak unsigned long boot_get_random_stack_canary(void)
+{
+	uintptr_t canary;
```
jenswi-linaro, line 1480: Please initialize the local variables, it's required by our coding standard.

ta/arch/arm/user_ta_header.c
```diff
@@ -46,7 +48,28 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
 					struct utee_params *up,
 					unsigned long cmd_id)
 {
-	TEE_Result res = __utee_entry(func, session_id, up, cmd_id);
+	TEE_Result res;
```
jenswi-linaro, line 51: This should be initialized with a value.

ta/arch/arm/user_ta_header.c
maximus64: The `__ta_entry()` function doesn't return and also defined with `__noreturn` attribute so if the compiler insert stack canary check in function epilogue, that code won't be reachable. If someone add silly code that return from `__ta_entry()`, it would generate compiler warning. If they ignore the warning then the code would crash even w/o stack protector because they can't do that from the TA entry point.

jenswi-linaro: OK, thanks.

jenswi-linaro:

core/arch/arm/kernel/boot.c
```diff
@@ -1473,6 +1473,31 @@ struct ns_entry_context *boot_core_hpen(void)
...
 
+#if defined(CFG_CORE_STACK_PROTECTOR)
+/* generate random stack canary value on boot up */
+__weak unsigned long boot_get_random_stack_canary(void)
+{
+	uintptr_t canary = 0xbaaaad00;
+	TEE_Result ret = TEE_ERROR_GENERIC;
+
+	/*
+	 * With virtualization the RNG doesn't get initialize in Nexus core.
```
etienne-lms, line 1484: s/doesn't get initialize/is not initialized/

core/arch/arm/kernel/boot.c
```diff
@@ -1473,6 +1473,31 @@ struct ns_entry_context *boot_core_hpen(void)
 }
 #endif
 
+#if defined(CFG_CORE_STACK_PROTECTOR)
+/* generate random stack canary value on boot up */
```
etienne-lms, line 1477: s/generate/Generate/

core/arch/arm/kernel/boot.c
```diff
@@ -1473,6 +1473,31 @@ struct ns_entry_context *boot_core_hpen(void)
...
+	if (IS_ENABLED(CFG_VIRTUALIZATION)) {
+		IMSG("WARNING: Using fixed value for stack canary");
+		return canary;
+	}
+
+	ret = crypto_rng_read(&canary, sizeof(canary));
+	if (ret != TEE_SUCCESS)
+		panic("Failed to generate random stack canary");
+
+	/* leave null byte in canary to prevent string base exploit */
```
etienne-lms, line 1496: s/leave/Leave/

core/arch/arm/kernel/boot.c
```diff
@@ -1473,6 +1473,31 @@ struct ns_entry_context *boot_core_hpen(void)
 }
 #endif
 
+#if defined(CFG_CORE_STACK_PROTECTOR)
+/* generate random stack canary value on boot up */
+__weak unsigned long boot_get_random_stack_canary(void)
```
etienne-lms, line 1478: Change here to `uintptr_t` for con, consistency?

If platform can legitimatly override this function, it should better be prefixed `plat_` (e.g. `plat_get_random_stack_canary()`).

mk/config.mk
```diff
@@ -313,6 +313,15 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
 # corruption vulnerabilities more difficult.
 CFG_CORE_ASLR ?= y
 
+# Stack Protection for TEE Core
+#
+# This flag enable the compiler stack protection mechanisms
```
etienne-lms, line 318: s/enable/enables/

can remove the empty line above.

mk/config.mk
```diff
@@ -313,6 +313,15 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
...
+# -fstack-protector-strong. It will check the stack canary value before return
```
etienne-lms, line 319: s/return/returning/ ?

lib/libutils/isoc/stack_check.c
```diff
@@ -3,13 +3,27 @@
  * Copyright (c) 2014, STMicroelectronics International N.V.
  */
 #include <compiler.h>
+#include <trace.h>
+
+#if defined(__KERNEL__)
+# include <kernel/panic.h>
+# define PANIC() panic("stack smashing detected")
```
etienne-lms, line 10: There already is a `stack smashing detected` error trace in `__stack_chk_fail()`.

core/include/kernel/boot.h
```diff
@@ -43,6 +43,8 @@ struct boot_embdata {
 extern uint8_t embedded_secure_dtb[];
 extern const struct core_mmu_config boot_mmu_config;
 
+unsigned long boot_get_random_stack_canary(void);
```
etienne-lms, line 46: condition declaration to `CFG_CORE_STACK_PROTECTOR`

etienne-lms (commented):
Some minor comments.

jforissier:
- I think these features should be called `CFG_{CORE,TA}_STACK_PROTECTOR_STRONG` since there are several variants of the `-stack-protector*` compiler flag.

- How does this compare to `CFG_CORE_DEBUG_CHECK_STACKS=y`? Could you provide an example of what happens when a stack overflow occurs in the TEE core? For an example with `CFG_CORE_DEBUG_CHECK_STACKS=y` see https://github.com/OP-TEE/optee_os/commit/28d6e35a8dd5481da71270c28d8d321e9dbe009f. If this can provide a useful stack trace right at the point of the stack overflow, then I suppose we could get rid of `CFG_CORE_DEBUG_CHECK_STACKS` which is clearly more complicated that this PR.

maximus64:
> * I think these features should be called `CFG_{CORE,TA}_STACK_PROTECTOR_STRONG` since there are several variants of the `-stack-protector*` compiler flag.

I updated the patch to also include `CFG_CORE_STACK_PROTECTOR_STRONG` and `CFG_CORE_STACK_PROTECTOR_ALL` to choose between different stack protector modes. Same for TA `CFG_TA_STACK_PROTECTOR_STRONG` and `CFG_TA_STACK_PROTECTOR_ALL`
 
>     * How does this compare to `CFG_CORE_DEBUG_CHECK_STACKS=y`? Could you provide an example of what happens when a stack overflow occurs in the TEE core? For an example with `CFG_CORE_DEBUG_CHECK_STACKS=y` see [28d6e35](https://github.com/OP-TEE/optee_os/commit/28d6e35a8dd5481da71270c28d8d321e9dbe009f). If this can provide a useful stack trace right at the point of the stack overflow, then I suppose we could get rid of `CFG_CORE_DEBUG_CHECK_STACKS` which is clearly more complicated that this PR.

`CFG_CORE_DEBUG_CHECK_STACKS` is difference than the compiler stack protector. It check if the stack memory region for underflow and overflow. But the `CFG_TA_STACK_PROTECTOR*` check the canary value in stack frame before returning. It for prevent buffer overflow attack and doesn't check if the stack exceed the stack memory region

mk/config.mk
```diff
@@ -313,6 +313,35 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
...
+# logic will be add to all functions regardless of their vulnerability.
+CFG_CORE_STACK_PROTECTOR_ALL ?= n
+# Stack Protection for TA
+CFG_TA_STACK_PROTECTOR ?= y
+CFG_TA_STACK_PROTECTOR_STRONG ?= y
+CFG_TA_STACK_PROTECTOR_ALL ?= n
+$(eval $(call cfg-depends-all,CFG_CORE_STACK_PROTECTOR_STRONG,CFG_CORE_STACK_PROTECTOR))
+$(eval $(call cfg-depends-all,CFG_CORE_STACK_PROTECTOR_ALL,CFG_CORE_STACK_PROTECTOR))
+$(eval $(call cfg-depends-all,CFG_TA_STACK_PROTECTOR_STRONG,CFG_TA_STACK_PROTECTOR))
+$(eval $(call cfg-depends-all,CFG_TA_STACK_PROTECTOR_ALL,CFG_TA_STACK_PROTECTOR))
```
jforissier, line 342: This makes things slightly more complex to use than necessary and a bit confusing since `-fstack-protector-strong` implies `-fstack-protector` so I suggest the `CFG_` should work the same. I propose to introduce a helper `_CFG`:
```
_CFG_CORE_STACK_PROTECTOR := $(eval $(call cfg-one-enabled,CFG_CORE_STACK_PROTECTOR CFG_CORE_STACK_PROTECTOR_STRONG CFG_CORE_STACK_PROTECTOR_ALL))
```

Same for TA of course.

ta/ta.mk
```diff
@@ -29,6 +34,9 @@ ta-mk-file-export-vars-$(sm) += CFG_SECURE_DATA_PATH
 ta-mk-file-export-vars-$(sm) += CFG_TA_MBEDTLS_SELF_TEST
 ta-mk-file-export-vars-$(sm) += CFG_TA_MBEDTLS
 ta-mk-file-export-vars-$(sm) += CFG_TA_MBEDTLS_MPI
+ta-mk-file-export-vars-$(sm) += CFG_TA_STACK_PROTECTOR
+ta-mk-file-export-vars-$(sm) += CFG_TA_STACK_PROTECTOR_STRONG
+ta-mk-file-export-vars-$(sm) += CFG_TA_STACK_PROTECTOR_ALL
```
jforissier, line 39: Not sure these need to be exported. They change the TA compilation flags, so I think it should be left to the user to set or not some of these when building the TA (assuming objects with and without stack protector can be mixed?).

mk/config.mk
```diff
@@ -313,6 +313,35 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
...
+#   than 8 bytes.
+CFG_CORE_STACK_PROTECTOR ?= y
+# This enable stack protector flag -fstack-protector-strong. Stack protector
+# canary logic will be add for vulnerable functions that contain:
+# - An array of any size and type.
+# - A call to alloca().
+# - A local variable that has its address taken.
+CFG_CORE_STACK_PROTECTOR_STRONG ?= y
+# This enable stack protector flag -fstack-protector-all. Stack protector canary
+# logic will be add to all functions regardless of their vulnerability.
```
jforissier, line 333: added

mk/config.mk
```diff
@@ -313,6 +313,35 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
...
+# It will check the stack canary value before returning from a function to
+# prevent buffer overflow attacks. Stack protector canary logic will be add for
+# vulnerable functions that contain:
+# - A character array larger than 8 bytes.
+# - An 8-bit integer array larger than 8 bytes.
+# - A call to alloca() with either a variable size or a constant size bigger
+#   than 8 bytes.
+CFG_CORE_STACK_PROTECTOR ?= y
+# This enable stack protector flag -fstack-protector-strong. Stack protector
+# canary logic will be add for vulnerable functions that contain:
```
jforissier, line 327: added

mk/config.mk
```diff
@@ -313,6 +313,35 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
 # corruption vulnerabilities more difficult.
 CFG_CORE_ASLR ?= y
 
+# Stack Protection for TEE Core
+# This flag enables the compiler stack protection mechanisms -fstack-protector.
+# It will check the stack canary value before returning from a function to
+# prevent buffer overflow attacks. Stack protector canary logic will be add for
```
jforissier, line 319: added

jforissier (commented):
> I updated the patch to also include CFG_CORE_STACK_PROTECTOR_STRONG and CFG_CORE_STACK_PROTECTOR_ALL to choose between different stack protector modes. Same for TA CFG_TA_STACK_PROTECTOR_STRONG and CFG_TA_STACK_PROTECTOR_ALL

Thanks for doing this, but please note: when making non-trivial changes to a commit you should not keep the `Reviewed-by:` tags.

More comments below.

mk/config.mk
```diff
@@ -313,6 +313,35 @@ CFG_TA_ASLR_MAX_OFFSET_PAGES ?= 128
...
+# Stack Protection for TA
+CFG_TA_STACK_PROTECTOR ?= y
+CFG_TA_STACK_PROTECTOR_STRONG ?= y
+CFG_TA_STACK_PROTECTOR_ALL ?= n
+$(eval $(call cfg-depends-all,CFG_CORE_STACK_PROTECTOR_STRONG,CFG_CORE_STACK_PROTECTOR))
+$(eval $(call cfg-depends-all,CFG_CORE_STACK_PROTECTOR_ALL,CFG_CORE_STACK_PROTECTOR))
+$(eval $(call cfg-depends-all,CFG_TA_STACK_PROTECTOR_STRONG,CFG_TA_STACK_PROTECTOR))
+$(eval $(call cfg-depends-all,CFG_TA_STACK_PROTECTOR_ALL,CFG_TA_STACK_PROTECTOR))
+
+
```
etienne-lms, line 344: remove extra empty line

ta/arch/arm/user_ta_header.c
```diff
@@ -46,7 +49,25 @@ void __noreturn _C_FUNCTION(__ta_entry)(unsigned long func,
...
+
+	if (IS_ENABLED(CFG_TA_STACK_PROTECTOR) && !stack_canary_inited) {
+		uintptr_t canary = 0;
+
+		res = _utee_cryp_random_number_generate(&canary,
+							sizeof(canary));
+		if (res != TEE_SUCCESS)
+			_utee_return(res);
+
+		/* leave null byte in canary to prevent string base exploit */
```
etienne-lms, line 63: s/leave/Leave/

ta/ta.mk
etienne-lms: CFG_TA_STACK_PROTECTOR affects the library. It should be exported for TA to know wether or not the lib in the devkit embeds stack protection.

jforissier: Why would you need to know unless it has an impact on how you build your TA? This information is available in the generated OP-TEE config file anyways (`<TA dev kit>/host_include/conf.mk`)

maximus64: okay I removed the TA stack protector config export

maximus64:
Look like there some issue with the checkpatch CI. It wanted me to use `CFG__CFG_TA_STACK_PROTECTOR` instead of `_CFG_TA_STACK_PROTECTOR`. I tried and it doesn't work. Not sure how to make this warning go away.
```
 d12f2608 core: add support for compiler stack protector
WARNING: IS_ENABLED(_CFG_TA_STACK_PROTECTOR) is normally used as IS_ENABLED(CFG__CFG_TA_STACK_PROTECTOR)
#200: FILE: ta/arch/arm/user_ta_header.c:55:
+	if (IS_ENABLED(_CFG_TA_STACK_PROTECTOR) && !stack_canary_inited) {
```

etienne-lms (commented):
In commit message: i think you can remove "core:" prefix since the change proposes both core and TA stack protection.

jforissier:
Ping @jenswi-linaro since you gave a R-b previously.

jenswi-linaro:

PR #5668, opened 2022-11-23
core: ffa: remove pager annotations

jenswi-linaro:
Configuration with pager and FF-A is currently not supported. Supporting this would require extensions to the FF-A specification to be able to load OP-TEE with paging enabled. So far we don't have any platforms with FF-A which are memory constrained enough that paging can be motivated. If this would change we'll have a good use case to test with when adding pager support for FF-A.

Currently we have a few pager annotations (DECLARE_KEEP_PAGER() and __*_unpaged) which are effectively unused. So save us from adding yet more unused annotations by removing the few we have in the FF-A specific code.

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -60,7 +60,7 @@ struct sp_pkg_header {
 
 struct fip_sp_head fip_sp_list = STAILQ_HEAD_INITIALIZER(fip_sp_list);
 
-const struct ts_ops sp_ops;
+static const struct ts_ops sp_ops;
 
 /* List that holds all of the loaded SP's */
 static struct sp_sessions_head open_sp_sessions =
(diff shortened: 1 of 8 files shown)


core/arch/arm/mm/sp_mem.c
```diff
@@ -174,7 +174,7 @@ static void inactivate(struct mobj *mobj)
 	cpu_spin_unlock_xrestore(&sp_mem_lock, exceptions);
 }
 
-const struct mobj_ops mobj_sp_ops __weak __relrodata_unpaged("mobj_sp_ops") = {
+const struct mobj_ops mobj_sp_ops = {
```
etienne-lms, line 177: can be `static` now the weak constraint has been removed (there is a hint to update at line 20 :)

`mobj_ffa_ops` and `sp_ops` should be `static` in their respecitive source file.

jenswi-linaro:
Update

jforissier (commented):

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1089,11 +1089,7 @@ static void sp_dump_state(struct ts_ctx *ctx)
 	user_mode_ctx_print_mappings(&utc->uctx);
 }
 
-/*
- * Note: this variable is weak just to ease breaking its dependency chain
- * when added to the unpaged area.
- */
-const struct ts_ops sp_ops __weak __relrodata_unpaged("sp_ops") = {
+const struct ts_ops sp_ops = {
```
etienne-lms, line 1092: `sp_ops` should also be `static`.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with remaining comment addressed.

core/arch/arm/kernel/secure_partition.c
jenswi-linaro: Thanks.

jenswi-linaro:
Comment addressed, tags applied.

PR #5669, opened 2022-11-23
libutils: implement aligned_alloc() and bump to C11 (gnu11)

etienne-lms:

### DIFF
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -739,6 +739,17 @@ void *mdbg_memalign(const char *fname, int lineno, size_t alignment,
 	return gen_mdbg_memalign(&malloc_ctx, fname, lineno, alignment, size);
 }
 
+#if __STDC_VERSION__ >= 201112L
+void *mdbg_aligned_alloc(const char *fname, int lineno, size_t alignment,
+			 size_t size)
+{
+	if (size % alignment)
+		return NULL;
+
+	return gen_mdbg_memalign(&malloc_ctx, fname, lineno, alignment, size);
+}
+#endif /* __STDC_VERSION__ */
+
 void mdbg_check(int bufdump)
 {
 	gen_mdbg_check(&malloc_ctx, bufdump);
@@ -823,6 +834,16 @@ void *memalign(size_t alignment, size_t size)
 	malloc_unlock(&malloc_ctx, exceptions);
 	return p;
 }
+
+#if __STDC_VERSION__ >= 201112L
+void *aligned_alloc(size_t alignment, size_t size)
+{
+	if (size % alignment)
+		return NULL;
+
+	return memalign(alignment, size);
+}
+#endif /* __STDC_VERSION__ */
 
 static void *get_payload_start_size(void *ptr, size_t *size)
 {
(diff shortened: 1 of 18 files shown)


lib/libutils/isoc/bget_malloc.c
```diff
@@ -824,6 +833,14 @@ void *memalign(size_t alignment, size_t size)
 	return p;
 }
 
+#if __STDC_VERSION__ >= 201112L
+void *aligned_alloc(size_t alignment, size_t size)
+{
+	assert(!(size % alignment));
```
jforissier, line 839: No, in this case NULL should be returned (and `errno` set to `EINVAL` if we had an `errno` variable, but we don't :wink:)

jforissier (commented):
Reviewing commit "libutils: implement aligned_alloc()":
s/enable/enabled/ (twice) in the commit description + one comment below. With that:

jforissier:
For commit "libutils: prefix system header guard with an underscore":

jforissier:
For commit "mk: compile with gnu11 instead of gnu99 standard":

Prefer "mk: compile with -std=gnu11 instead of -std=gnu99". Perhaps you could say it is supported since GCC 4.7 and link to https://gcc.gnu.org/wiki/C11Status?

lib/libutils/isoc/bget_malloc.c
etienne-lms: ok. done (but for `errno`). Thanks.

etienne-lms:

lib/libutils/isoc/include/stdlib.h
```diff
@@ -29,4 +29,14 @@ int rand(void);
 unsigned long _strtoul (const char *nptr, char **endptr, int base);
 unsigned long strtoul (const char *s, char **ptr, int base);
 
-#endif /* STDLIB_H */
+#if __STDC_VERSION__ >= 201112L
+void *aligned_alloc(size_t alignment, size_t size);
```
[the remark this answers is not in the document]
jforissier: I looked at the man pages, everything is supposed to be made available via `<stdlib.h>` except `memalign()` which requires `<malloc.h>` (and AFAICT should not be pulled by `<stdlib.h>` alone). Our `<stdlib.h>` includes `<malloc.h>`, so it doesn't strictly comply regarding `memalign()`. glibc and newlib declare `calloc()`, `malloc()` and `realloc()` in  `<stdlib.h>` AND `<malloc.h>` (yes the prototypes are present twice). `aligned_alloc()` however is only in `<stdlib.h>`. So if we want `aligned_alloc()` NOT to be available ONLY when `<stdlib.h>`, it belongs to `<stdlib.h>`. 

Now I realize this explanation is quite convoluted and perhaps we should not care too much about this level of detail. Availability (or rather non-availability) of function should be based on `-sdt=xxx` and macros rather than which header exactly is included I suppose. So I won't object if you prefer to group `aligned_alloc()` with the others in `malloc.h`.

etienne-lms:
updated

jenswi-linaro:

etienne-lms:
Fiuxp commit squashed and tag applied.
@jforissier, I assumed your review tags sill apply.

jforissier:
> Fiuxp commit squashed and tag applied. @jforissier, I assumed your review tags sill apply.

Yes, thanks

PR #5671, opened 2022-11-23
optee-os: The sm4-xts algorithm is supported in the GP process.

XIEPINGAN:
In this patch, The sm4-xts algorithm is supported in the GP process.

Signed-off-by: Pingan Xie <xiepingan3@huawei.com>
Reviewed-by: Xiaoxu Zeng <zengxiaoxu@huawei.com>

### DIFF
--- a/core/crypto/sm4.c
+++ b/core/crypto/sm4.c
@@ -268,3 +268,84 @@ void sm4_crypt_ctr(struct sm4_context *ctx, size_t length, uint8_t ctr[16],
 		length -= 16;
 	}
 }
+
+static void xts_multi(unsigned char *in, unsigned char *out)
+{
+	uint8_t tt = 0;
+	uint8_t t = 0;
+	int i = 0;
+
+	for (i = 0; i < 16; i++) {
+		tt = in[i] >> 7;
+		out[i] = ((in[i] << 1) | t) & 0xFF;
+		t = tt;
+	}
+
+	out[0] ^= (0x87 & (0 - tt));
+}
+
+static void xor_128(const uint8_t a[16], const uint8_t b[16], uint8_t c[16])
+{
+	int i = 0;
+
+	for (i = 0; i < 16; i++)
+		c[i] = a[i] ^ b[i];
+}
+
+void sm4_crypt_xts(struct sm4_context *ctx, struct sm4_context *ctx_ek,
+		   struct sm4_context *ctx_dk, size_t len, uint8_t *iv,
+		   const uint8_t *input, uint8_t *output)
+{
+	uint8_t tweak[16] = { };
+	uint8_t tweak1[16] = { };
+	uint8_t ct[16] = { };
+	size_t i = 0;
+
+	assert(len >= 16);
+
+	sm4_one_round(ctx_ek->sk, iv, tweak);
+
+	if (ctx->mode == SM4_DECRYPT && (len % 16))
+		len -= 16;
+
+	while (len >= 16) {
+		xor_128(input, tweak, ct);
+		sm4_one_round(ctx->sk, ct, ct);
+		xor_128(ct, tweak, output);
+
+		xts_multi(tweak, tweak);
+		len -= 16;
+		if (len == 0) {
+			sm4_one_round(ctx_dk->sk, tweak, iv);
+			return;
+		}
+		input += 16;
+		output += 16;
+	}
+
+	if (ctx->mode == SM4_ENCRYPT) {
+		memcpy(ct, output - 16, 16);
+		for (i = 0; i < len; i++) {
+			output[i] = ct[i];
+			ct[i] = input[i];
+		}
+
+		xor_128(ct, tweak, ct);
+		sm4_one_round(ctx->sk, ct, ct);
+		xor_128(ct, tweak, ct);
+		memcpy(output - 16, ct, 16);
+	} else {
+		xts_multi(tweak, tweak1);
+		xor_128(input, tweak1, ct);
+		sm4_one_round(ctx->sk, ct, ct);
+		xor_128(ct, tweak1, ct);
+
+		for (i = 0; i < len; ++i) {
+			output[16 + i] = ct[i];
+			ct[i] = input[16 + i];
+		}
+		xor_128(ct, tweak, ct);
+		sm4_one_round(ctx->sk, ct, ct);
+		xor_128(ct, tweak, output);
+	}
+}
--- a/core/tee/tee_svc_cryp.c
+++ b/core/tee/tee_svc_cryp.c
@@ -2613,8 +2613,10 @@ TEE_Result syscall_cryp_state_alloc(unsigned long algo, unsigned long mode,
 
 	switch (TEE_ALG_GET_CLASS(algo)) {
 	case TEE_OPERATION_CIPHER:
-		if ((algo == TEE_ALG_AES_XTS && (key1 == 0 || key2 == 0)) ||
-		    (algo != TEE_ALG_AES_XTS && (key1 == 0 || key2 != 0))) {
+		if ((TEE_ALG_GET_CHAIN_MODE(algo) == TEE_CHAIN_MODE_XTS &&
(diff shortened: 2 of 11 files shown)


core/tee/tee_svc_cryp.c
```diff
@@ -2613,8 +2613,10 @@ TEE_Result syscall_cryp_state_alloc(unsigned long algo, unsigned long mode,
...
 	case TEE_OPERATION_CIPHER:
-		if ((algo == TEE_ALG_AES_XTS && (key1 == 0 || key2 == 0)) ||
-		    (algo != TEE_ALG_AES_XTS && (key1 == 0 || key2 != 0))) {
+		if (((algo == TEE_ALG_AES_XTS || algo == TEE_ALG_SM4_XTS) &&
+		    (key1 == 0 || key2 == 0)) ||
+		    ((algo != TEE_ALG_AES_XTS && algo != TEE_ALG_SM4_XTS) &&
+		    (key1 == 0 || key2 != 0))) {
 			res = TEE_ERROR_BAD_PARAMETERS;
 		} else {
 			res = crypto_cipher_alloc_ctx(&cs->ctx, algo);
```
jenswi-linaro, line 2622: It seems that `crypto_cipher_alloc_ctx()` doesn't know `TEE_ALG_SM4_XTS`.

lib/libutee/include/tee_api_defines.h
```diff
@@ -137,6 +137,7 @@
 #define TEE_ALG_DES3_CBC_MAC_PKCS5              0x30000513
 #define TEE_ALG_SM4_ECB_NOPAD                   0x10000014
 #define TEE_ALG_SM4_CBC_NOPAD                   0x10000114
+#define TEE_ALG_SM4_XTS                         0x10000414
```
jenswi-linaro, line 140: From which specification?

XIEPINGAN: Thanks for your reply. The 0x10000414 is assembled according to the assembly rules. Bits[31:28] are TEE_OPERATION_CIPHER, Bits[11:8] are TEE_CHAIN_MODE_XTS, and Bits[7:0] are TEE_MAIN_ALGO_SM4.

jenswi-linaro: Yes, that's the logical choice. But it would also be nice to know if this is explicitly defined too since this way of constructing an ID was removed in version 1.2 of the spec.

I can't find this in any public spec, but it would be good to know if you know that it's part of a draft or otherwise planned.

XIEPINGAN: Yes. The spec doesn't have this algorithm, but we really need to use it. SM4-XTS is widely used and our customers have the same request. If I want to add this algorithm, how do I define it?

jenswi-linaro: We should be able to find a way to accommodate this. @jforissier, what's your take on this?

jforissier: `TEE_ALG_SM4_XTS` should be added to `tee_api_defines_extensions.h` instead of `tee_api_defines.h`, like was done for  [`TEE_ALG_DES3_CMAC`](https://github.com/OP-TEE/optee_os/commit/eee637e7c) for example. And the value should be in the implementation-specific range: `0xF0000414` would be fine I think.

core/tee/tee_svc_cryp.c
XIEPINGAN: The software computing support for sm4-xts has been added to the second commit.

jenswi-linaro: Good, please change the order of the two commits.

core/crypto/sm4.c
```diff
@@ -268,3 +268,82 @@ void sm4_crypt_ctr(struct sm4_context *ctx, size_t length, uint8_t ctr[16],
...
+		c[i] = a[i] ^ b[i];
+}
+
+void sm4_crypt_xts(struct sm4_context *ctx, struct sm4_context *ctx_ek,
+		   struct sm4_context *ctx_dk, size_t len, uint8_t *iv,
+		   const uint8_t *input, uint8_t *output)
+{
+	uint8_t tweak[16], tweak1[16];
+	uint8_t ct[16];
+	size_t i;
```
jforissier, line 300:
```C
	uint8_t tweak[16] = { };
	uint8_t tweak1[16] = { };
	uint8_t ct[16] = { };
	size_t i = 0;
```

core/tee/tee_svc_cryp.c
```diff
@@ -2613,8 +2613,10 @@ TEE_Result syscall_cryp_state_alloc(unsigned long algo, unsigned long mode,
 
 	switch (TEE_ALG_GET_CLASS(algo)) {
 	case TEE_OPERATION_CIPHER:
-		if ((algo == TEE_ALG_AES_XTS && (key1 == 0 || key2 == 0)) ||
-		    (algo != TEE_ALG_AES_XTS && (key1 == 0 || key2 != 0))) {
+		if (((algo == TEE_ALG_AES_XTS || algo == TEE_ALG_SM4_XTS) &&
+		    (key1 == 0 || key2 == 0)) ||
+		    ((algo != TEE_ALG_AES_XTS && algo != TEE_ALG_SM4_XTS) &&
+		    (key1 == 0 || key2 != 0))) {
```
jforissier, line 2619: Please use `TEE_ALG_GET_CHAIN_MODE()`:
```C
		if ((TEE_ALG_GET_CHAIN_MODE(algo) == TEE_CHAIN_MODE_XTS && 
			(key1 == 0 || key2 == 0)) ||
		    (TEE_ALG_GET_CHAIN_MODE(algo) != TEE_CHAIN_MODE_XTS &&
		    	(key1 == 0 || key2 != 0))) {
```

jforissier (commented):
- For commit "tee: The sm4-xts algorithm is supported in the GP process."
Please change subject to "core, libutee: introduce TEE_ALG_SM4_XTS"
- For commit "crypto: Add software computing support for sm4-xts."
Please change subject to "core: crypto: implement SM4 XTS"

More comments below.

jforissier:
@XIEPINGAN do you have test vectors for SM4 XTS? It would be nice to add a test case to `xtest 4003`: https://github.com/OP-TEE/optee_test/blob/3.19.0/host/xtest/regression_4000.c#L2123

core/tee/tee_svc_cryp.c
XIEPINGAN: Is it possible not to change the order of commit, because the first commit is to introduce sm4-xts in the gp process, and the second commit is to implement sm4-xts software computing.

jenswi-linaro: Each commit in the git history should build, so you must change the order. You can add the TEE_ALG_SM4_XTS define first in a separate commit.

XIEPINGAN: Can I merge these two commit?

jenswi-linaro: I would prefer to take them as separate commits.

XIEPINGAN:
The sm4-xts test case has been added locally. After this submission is incorporated, I will submit it to xtest 4003. @jforissier

core/tee/tee_svc_cryp.c
jforissier: @jenswi-linaro commit "introduce TEE_ALG_SM4_XTS" comes first so both commits build fine. Can you clarify what you expect? Would you like the first commit to introduce the define and touch libutee only, then the second one to add the implementation in core?

jenswi-linaro: My bad, the order is good.

jenswi-linaro:

jforissier:
@XIEPINGAN please add the missing space to fix the checkpatch message. With that:

jforissier:
@XIEPINGAN correct, the CI errors are unrelated and the issue has been fixed now. I am merging this PR, thank you for your contribution.

PR #5672, opened 2022-11-23
core: add overflow checks in crypto_aes_ccm_init()

jenswi-linaro:
aad_len and payload_len are of the type size_t which has a greater range than int which is used for the corresponding arguments when passed to ccm_init(). So to guard against wrapped or truncated values check that the variables can be in a int first before calling ccm_init().

Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>

### DIFF
--- a/core/lib/libtomcrypt/ccm.c
+++ b/core/lib/libtomcrypt/ccm.c
@@ -85,6 +85,10 @@ static TEE_Result crypto_aes_ccm_init(struct crypto_authenc_ctx *aectx,
 	if (nonce_len > TEE_CCM_NONCE_MAX_LENGTH)
 		return TEE_ERROR_BAD_PARAMETERS;
 
+	/* Check that payload_len and aad_len will fit into ints */
+	if (payload_len > INT_MAX || aad_len > INT_MAX)
+		return TEE_ERROR_BAD_PARAMETERS;
+
 	/* check the tag len */
 	if ((tag_len < 4) || (tag_len > TEE_CCM_TAG_MAX_LENGTH) ||
 	    (tag_len % 2 != 0))


core/lib/libtomcrypt/ccm.c
```diff
@@ -85,6 +85,10 @@ static TEE_Result crypto_aes_ccm_init(struct crypto_authenc_ctx *aectx,
 	if (nonce_len > TEE_CCM_NONCE_MAX_LENGTH)
 		return TEE_ERROR_BAD_PARAMETERS;
 
+	/* Check that payload_len and aad_len will fit into ints */
```
etienne-lms, line 88: I seems these sizes are handle as `unsigned long` in this source file, as other possibly long sizes. Only `ccm_init()` uses `int` types in the function ABI. Maybe we should better change the function prototype.

jenswi-linaro: The problem is with `ccm_init()` which takes these two as ints.

etienne-lms: sorry, I meant `long` (`signed long`).
Should maybe be discussed with ltc folks since it changes an api in the lib.
In the meantime, `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`.
Compiler does not complain on 32bit that the below test is always false?

jenswi-linaro: `INT_MAX < `SIZE_MAX`

jenswi-linaro:
Tag applied

core/lib/libtomcrypt/ccm.c
etienne-lms: right! (as usual :)

PR #5678, opened 2022-11-29
Gp update

jenswi-linaro:
This is the first step in updating the TEE Internal Core API. 
More will follow, but this should be nice to get out of the way.


lib/libutee/tee_api_objects.c
```diff
@@ -37,34 +37,52 @@ void __utee_from_attr(struct utee_attribute *ua, const TEE_Attribute *attrs,
...
 		objectInfo->keySize = 0;
 		objectInfo->maxKeySize = 0;
 		objectInfo->objectUsage = 0;
 		objectInfo->dataSize = 0;
 		objectInfo->dataPosition = 0;
 		objectInfo->handleFlags = 0;
+	} else {
+		objectInfo->objectType = info.obj_type;
+		objectInfo->keySize = info.obj_size;
+		objectInfo->maxKeySize = info.max_obj_size;
```
jforissier, line 58: How about first switching to the new names (`objectSize`/`maxObjectSize`) everywhere prior to this commit?

jforissier (commented):
Reviewing commit "Use struct utee_object_info in ta syscall abi:

- s/ta/TA/ in the subject
- One comment below

With that:

jforissier:
For commit "Move tee_api.h and tee_ta_api.h into tee_internal_api.h":

lib/libutee/tee_api_objects.c
jenswi-linaro: I'd rather keep them as is here. These functions will need to be duplicated with a 1.1 version before we're at version 1.3.1. So it's easier to keep track of what's v1.1 if we wait with updating the new names until have copied out a version 1.1 function.

I hope it's not a big deal in the PR and it will be pretty clear what I'm meaning soon in the next PR.

jforissier: Got it, makes sense.

jenswi-linaro:
Tags applied, note that "core: update to new field names in TEE_ObjectInfo" doesn't have any A-B or R-B.

jforissier:
> Tags applied, note that "core: update to new field names in TEE_ObjectInfo" doesn't have any A-B or R-B.

It looks OK except that you don't say why you don't update everywhere:

>  core: update to new field names in TEE_ObjectInfo
> 
> With GP TEE Internal Core API version 1.1.1 the two fields keySize and
> maxKeySize was renamed to objectSize and maxObjectSize respectively.
> Update the rest of the code to reflect that change as a preparation for

Here. It is not all the rest.

> future updates.

jenswi-linaro:
> > Tags applied, note that "core: update to new field names in TEE_ObjectInfo" doesn't have any A-B or R-B.
> 
> It looks OK except that you don't say why you don't update everywhere:
> 
> > core: update to new field names in TEE_ObjectInfo
> > With GP TEE Internal Core API version 1.1.1 the two fields keySize and
> > maxKeySize was renamed to objectSize and maxObjectSize respectively.
> > Update the rest of the code to reflect that change as a preparation for
> 
> Here. It is not all the rest.
> 
> > future updates.

Well, it was in the context of `/core` given the `core:` prefix. Would it be more clear to say "the rest of the core code"?

jforissier:
> Well, it was in the context of /core given the core: prefix. Would it be more clear to say "the rest of the core code"?

Yes I think so. In any case that commit is:

Thanks!

jenswi-linaro:
Thanks, tag applied.

Note there's a bugfix added to "Use struct utee_object_info in TA syscall abi":
``` diff
--- a/core/tee/tee_svc_cryp.c
+++ b/core/tee/tee_svc_cryp.c
@@ -1127,7 +1127,7 @@ TEE_Result syscall_cryp_obj_get_info(unsigned long obj,
 	o_info.data_size = o->info.dataSize;
 	o_info.data_pos = o->info.dataPosition;
 	o_info.handle_flags = o->info.handleFlags;
-	res = copy_to_user_private(info, &o_info, sizeof(o->info));
+	res = copy_to_user_private(info, &o_info, sizeof(o_info));
 
 exit:
 	return res;
```

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for the series.

jenswi-linaro:
Tag applied.

PR #5684, opened 2022-12-01
 AMD/Xilinx Versal ACAP: Physical Unclonable Function

ldts:
The Versal® device contains a physically unclonable function (PUF). The PUF creates a signature (or fingerprint) of each device that is unique to that device. The KEK can only be directly fed into the AES-GCM engine and cannot be read out of the device.

The KEK is 256 bits in length with 256 bits of entropy and is used to encrypt the users red key allowing its storage in black (encrypted) form. The black key can be stored in either eFUSEs, BBRAM, or external storage.

Enhanced from the previous generation, the Versal device PUF also outputs a user accessible unique ID that is cryptographically isolated from the PUF KEK despite using the same entropy source. While unique to each device, it is not considered a “secret” and does not have the same access protections as the KEK.

https://docs.xilinx.com/r/en-US/am011-versal-acap-trm/Physically-Unclonable-Function?tocId=Xhon0EQi82Y2A1DuolwnfQ

### DIFF
--- a/core/drivers/versal_puf.c
+++ b/core/drivers/versal_puf.c
@@ -1,0 +1,277 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2022 Foundries.io Ltd
+ * Jorge Ramirez-Ortiz <jorge@foundries.io>
+ */
+
+#include <arm.h>
+#include <confine_array_index.h>
+#include <drivers/versal_mbox.h>
+#include <drivers/versal_nvm.h>
+#include <drivers/versal_puf.h>
+#include <initcall.h>
+#include <mm/core_memprot.h>
+#include <string.h>
+#include <tee/cache.h>
+
+/* Protocol API with the remote processor */
+#define VERSAL_PUF_MODULE_SHIFT		8
+#define VERSAL_PUF_MODULE		12
+#define PUF_API_ID(_id) ((VERSAL_PUF_MODULE << VERSAL_PUF_MODULE_SHIFT) | (_id))
+
+enum versal_puf_error {
+	/* Registration */
+	ERROR_INVALID_PARAM = 0x02,
+	ERROR_INVALID_SYNDROME_MODE = 0x03,
+	ERROR_SYNDROME_WORD_WAIT_TIMEOUT = 0x04,
+	ERROR_PUF_DONE_WAIT_TIMEOUT = 0x07,
+	ERROR_REGISTRATION_INVALID = 0x08,
+	SHUTTER_GVF_MISMATCH = 0x09,
+	ERROR_SYN_DATA_ERROR = 0x0A,
+	IRO_FREQ_WRITE_MISMATCH = 0x0B,
+	/* Regeneration */
+	ERROR_CHASH_NOT_PROGRAMMED = 0x10,
+	ERROR_PUF_STATUS_DONE_TIMEOUT = 0x11,
+	ERROR_INVALID_REGENERATION_TYPE = 0x12,
+	ERROR_INVALID_PUF_OPERATION = 0x13,
+	ERROR_REGENERATION_INVALID = 0x14,
+	ERROR_REGEN_PUF_HD_INVALID = 0x15,
+	ERROR_INVALID_READ_HD_INPUT = 0x16,
+	ERROR_PUF_DONE_KEY_NT_RDY = 0x17,
+	ERROR_PUF_DONE_ID_NT_RDY = 0x18,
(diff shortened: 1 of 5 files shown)


core/include/drivers/versal_puf.h
```diff
@@ -0,0 +1,78 @@
...
+enum versal_puf_api {
+	VERSAL_PUF_API_FEATURES = 0U,
+	VERSAL_PUF_REGISTER,
+	VERSAL_PUF_REGENERATE,
+	VERSAL_PUF_CLEAR_ID,
+};
+
+#define __aligned_puf	__aligned(CACHELINE_LEN)
+
+TEE_Result versal_puf_regenerate(struct versal_puf_data *buf,
```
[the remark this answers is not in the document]
ldts: @jenswi-linaro just added some additional comments. The reason for this PUF is essentially the same than what we did for Zynqmp (the HUK generation)

core/drivers/versal_puf.c
```diff
@@ -0,0 +1,270 @@
...
+		VERSAL_PUF_ERROR(ERROR_REGEN_PUF_HD_INVALID),
+		VERSAL_PUF_ERROR(ERROR_INVALID_READ_HD_INPUT),
+		VERSAL_PUF_ERROR(ERROR_PUF_DONE_KEY_NT_RDY),
+		VERSAL_PUF_ERROR(ERROR_PUF_DONE_ID_NT_RDY),
+		VERSAL_PUF_ERROR(ERROR_PUF_ID_ZERO_TIMEOUT),
+	};
+
+	if (err <= ERROR_PUF_ID_ZERO_TIMEOUT && err >= ERROR_INVALID_PARAM) {
+		if (elist[err - ERROR_INVALID_PARAM].name)
+			return elist[err - ERROR_INVALID_PARAM].name;
```
jenswi-linaro, line 76: After seeing this pattern a few times in your code I now realize that this can serve as a spectre gadget. Perhaps not very likely to be exploitable, but still.

What does others think? Do we care here?

ldts: hey, sure, I care :) maybe you can share a bit more of your thinking?

jenswi-linaro: If the CPU speculates past the checks at line 74 and 75 it may cause the caller to (while still speculating) dereference `name` via an out-of-bounds index and populate the cache based on that address.

For other examples search for where `confine_array_index()` is called.

ldts: ah ok. I wasnt familiar with this. should I then confine the index?

so the attacker, once it knows how to get that cache information, would have to add a jump address on the error function stack and cause the PUF to fail? um, is this easy to do?...seems someone able to do it (populate a function stack with pointers) could anyways attack any other part of OP-TEE, no?

For consistency I'd vote in favor of confining the index everywhere, but please let me know if you are ok with that.

jenswi-linaro: > so the attacker, once it knows how to get that cache information, would have to add a jump address on the error function stack and cause the PUF to fail? um, is this easy to do?...seems someone able to do it (populate a function stack with pointers) could anyways attack any other part of OP-TEE, no?

If an attacker can do that then we're in even deeper trouble. So to justify the usage of `confine_array_index()` it must be something that happens before OP-TEE is compromised. The index is received from outside of OP-TEE so I guess that OP-TEE doesn't have to be compromised in this case.

ldts: so what is next? having some guidelines in the docs would be really useful IMO.

jenswi-linaro: Please use `confine_array_index()` to cap the index and we should be good here.

ldts: ok.

ldts:
updated and rebased

ldts:
@jenswi-linaro can I add your review tag? once this is merged I can post the HUK driver and I think we are pretty much done with the Versal ACAP support for the time being.

nathan-menhorn:
Hi @jenswi-linaro any updates on this PR?

nathan-menhorn:
Hi @ldts any updates on this PR?

nathan-menhorn:
> > Hi @ldts any updates on this PR?
> 
> No not from from my side (I have nothing more to add to it ). With the vacation period approaching many will be OoO already or trying to wrap up their on PRs so we just have to wait really.

Thanks @ldts.

jenswi-linaro (commented):

ldts:
@jenswi-linaro thanks, tag added

ldts:
@jforissier I think this is ready.

ldts:
fixed the conflict due to a previous merge.

PR #5685, opened 2022-12-02
ta: pkcs11: fix debug message

sa-kib:
Minor mistake of printing key handle values after reset.

Signed-off-by: Sergiy Kibrik <Sergiy_Kibrik@epam.com>


etienne-lms (commented):

PR #5686, opened 2022-12-02
SCMI server from SCP-firmware (take 2)

etienne-lms:
This P-R superseeds https://github.com/OP-TEE/optee_os/pull/5627.

The main change is the build process. This P-R builds SCP-firmware using optee_os native build scripts instead of using SCP-firmware native CMake build process.

I have also updated CI script (_ci.yml_) to get SCP-firmware master branch tip for CI tests and testing build of vexpress-fvp and vexpress-qemuv8a with SCP-firmware SCMI stack.

This P-R is tested with latest master of SCP-firmware that integrates https://github.com/ARM-software/SCP-firmware/pull/664.

I've remove the review tags from commit already reviewed in #5627 since content sowhat changed, but for 1st commit "drivers: clk: add support for clk_get_rates_array" that has not changed.

I expected 2 checkpatch warnings due to use of `IS_ENABLED()` with `_CFG_xxx` macros:
```
WARNING: IS_ENABLED(_CFG_SCMI_PTA_SMT_HEADER) is normally used as IS_ENABLED(CFG__CFG_SCMI_PTA_SMT_HEADER)
#30: FILE: core/pta/scmi.c:18:
+	if (IS_ENABLED(_CFG_SCMI_PTA_SMT_HEADER))

WARNING: IS_ENABLED(_CFG_SCMI_PTA_MSG_HEADER) is normally used as IS_ENABLED(CFG__CFG_SCMI_PTA_MSG_HEADER)
#33: FILE: core/pta/scmi.c:21:
+	if (IS_ENABLED(_CFG_SCMI_PTA_MSG_HEADER))
```


core/lib/scmi-server/sub.mk
```diff
@@ -0,0 +1,174 @@
...
+		-Wno-declaration-after-statement
+
+# The below directives will be removed once SCP-firmware pull requests
+# 728, 729 and 730 are merged.
+cflags-lib-y += -Wno-undef \
+		-Wno-maybe-uninitialized \
+		-Wno-unused-but-set-variable \
+		-Wno-suggest-attribute=format \
+
+cflags-lib-$(_CFG_SCPFW_NOTIFICATION) += -DBUILD_HAS_NOTIFICATION \
```
jenswi-linaro, line 62: This should go into `cppflags-lib-...`, same below.

core/lib/scmi-server/conf-optee-stm32mp1.mk
```diff
@@ -0,0 +1,18 @@
+_CFG_SCPFW_MOD_RESET_DOMAIN := y
```
jenswi-linaro, line 1: Why do you use a `_CFG_` prefix instead of `CFG_`?

core/lib/scmi-server/include/fwk_assert.h
```diff
@@ -0,0 +1,39 @@
...
+/*
+ * Copyright (c) 2022, Linaro Limited
+ *
+ */
+
+#ifndef SCMI_SERVER_FWK_ASSERT_H
+#define SCMI_SERVER_FWK_ASSERT_H
+
+#include <assert.h>
+#include_next <fwk_assert.h>
```
jenswi-linaro, line 11: Is this a temporary workaround?

core/lib/scmi-server/sub.mk
```diff
@@ -0,0 +1,174 @@
...
+# be explicitly added in srcs-y, as for generic module scmi below.
+#
+# scpfw-embed-mod takes 4 arguments:
+# $1 module name, lowercase
+# $2 module directory name
+# $3 module parent directory relative path in scpfw tree
+# $4 module name, uppercase, relates to _CFG_SCPFW_MOD_$4
+define scpfw-embed-mod
+ifneq (,$$(wildcard $(scpfw-path)/$3/$2/include/*))
+incdirs_ext-y += $(scpfw-path)/$3/$2/include
```
jenswi-linaro, line 114: This could lead to some very long command lines. This is more of an observation, I'm not sure it's a problem.

mk/config.mk
```diff
@@ -750,10 +750,41 @@ CFG_SCMI_MSG_VOLTAGE_DOMAIN ?= n
...
+# from SCP-firmware package as an built-in SCMI stack in core. This
+# configuration mandates target product identifier is configured with
+# CFG_SCMI_SCPFW_PRODUCT and the SCP-firmware source tree path with
+# CFG_SCP_FIRMWARE.
+CFG_SCMI_SCPFW ?= n
+
+ifeq ($(CFG_SCMI_SCPFW),y)
+$(call force,CFG_SCMI_PTA,y,Required by CFG_SCMI_SCPFW)
+ifeq (,$(CFG_SCMI_SCPFW_PRODUCT))
+$(error CFG_SCMI_SCPFW=y mandates CFG_SCMI_SCPFW_PRODUCT configuration)
```
jenswi-linaro, line 771: I prefer to use the word "requires" for this.

core/lib/scmi-server/conf-optee-stm32mp1.mk
etienne-lms: I initially did not want to add a lot of `CFG_` if not necessary. These switches are set from the platform and I don't expect a build env to override the platform selected config. I though `_CFG` would apply.

core/lib/scmi-server/include/fwk_assert.h
etienne-lms: I don't think so. I think it will stay.
I can try to move this code in SCP-firmware arch/none/opee/include to get it maintained (and possibly removed one day) from there. Makes sense.

core/lib/scmi-server/sub.mk
etienne-lms: ok, thanks.

mk/config.mk
etienne-lms: ok

core/lib/scmi-server/conf-optee-stm32mp1.mk
jenswi-linaro: In my view we use a `_CFG` prefix for internal helper variables that are derived by some logic from other flags, that is, they are normally not set directly in a configuration. I'd say when in doubt, use an ordinary `CFG` prefix instead.

core/lib/scmi-server/include/fwk_assert.h
jenswi-linaro: OK, thanks. The optee part of the patch seems to suggest that it has something to do with OP-TEE adaptations.

etienne-lms: I've not removed this header file yet. It's planned but I prefer to be sure SCP-fmw is OK to host that wrapper header before doing so.

jenswi-linaro: OK

core/lib/scmi-server/sub.mk
```diff
@@ -0,0 +1,176 @@
...
+		      -DDISABLE_CPPCHECK=1 \
+		      -DCFG_NUM_THREADS=$(CFG_NUM_THREADS) \
+		      -DSCP_OPTEE_DIR:PATH=$(CURDIR) \
+		      -DCFG_CROSS_COMPILE=$(lastword $(CROSS_COMPILE_core))
+
+gensrcs-y += fwk_module_list
+force-gensrc-fwk_module_list := y
+produce-fwk_module_list = build/framework/src/fwk_module_list.c
+depends-fwk_module_list = FORCE
+recipe-fwk_module_list = cmake -S $(scpfw-path) -B $(scpfw-out-path)/build $(scpfw-cmake-flags-y)
```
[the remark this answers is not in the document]
etienne-lms: Using cmake here makes it easier to describe the scp-fmw modules embedded with the framework and generate the related source/header files pair. This part is described per 'product' (a-k-a platform) in SCP-firmware using CMake description file. Note we could possibly sete the dependency on:
```
depends-fwk_module_list = $(scpfw-path)/product/$(scpfw-product)/fw/Firmware.cmake
```

To get rid of the cmake processing, we need to replicate it in _scmi-server/sub.mk_ and a python script. Doable but I think it complicates the integration, but still doable.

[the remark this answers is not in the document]
jenswi-linaro: Let's try this first and see if it causes any trouble.

jenswi-linaro (commented):

etienne-lms:
I'll squash the fixup commits and rebase to go through a CI run.

etienne-lms:
Updated, review tag applied.
I still expect a CI/Code style failure related to checkpatch warnings report:
```
WARNING: IS_ENABLED(_CFG_SCMI_PTA_SMT_HEADER) is normally used as IS_ENABLED(CFG__CFG_SCMI_PTA_SMT_HEADER)
#30: FILE: core/pta/scmi.c:18:
+	if (IS_ENABLED(_CFG_SCMI_PTA_SMT_HEADER))

WARNING: IS_ENABLED(_CFG_SCMI_PTA_MSG_HEADER) is normally used as IS_ENABLED(CFG__CFG_SCMI_PTA_MSG_HEADER)
#33: FILE: core/pta/scmi.c:21:
+	if (IS_ENABLED(_CFG_SCMI_PTA_MSG_HEADER))

**Comment by jforissier** [comment]
>  I still expect a CI/Code style failure related to checkpatch warnings report:

@etienne-lms we could `#define IS_ENABLED2(x) IS_ENABLED(x)` to silence that pesky checkpatch warning, because I can't think of a fix that would be acceptable upstream (Linux).

**Comment by etienne-lms** [comment]
I have pushed a proposal for `IS_ENABLED()`. See the 2 commits "xxx: IS_ENABLED2() ..." in the new series, inserted before the patch that now uses this macro.
I've appended 2 fixup commits.
I can't get rid of Cmake traces 
```
-- Configuring done
-- Generating done
-- Build files have been written to: ...
```

File: lib/libutils/ext/include/config.h [hunk 7]
```diff
@@ -29,6 +29,11 @@
  */
 #define IS_ENABLED(config_macro) Z_IS_ENABLED1(config_macro)
 
+/**
+ * @brief As IS_ENABLED() but to be used with _CFG_* internal switches
```
jforissier, line 33: add "to avoid triggering false positives with checkpatch" maybe?

jforissier (commented):
For "libutils: IS_ENABLED2() for _CFG_* switches":

jforissier:
> I have pushed a proposal for `IS_ENABLED()`. See the 2 commits "xxx: IS_ENABLED2() ..." in the new series, inserted before the patch that now uses this macro. I've appended 2 fixup commits. I can't get rid of Cmake traces
> 
> ```
> -- Configuring done
> -- Generating done
> -- Build files have been written to: ...
> ```

Maybe like so: (untested)

```make
ifeq ($(cmd-echo-silent),true)
cmake-redirect = >/dev/null
endif
recipe-fwk_module_list = cmake -S $(scpfw-path) -B $(scpfw-out-path)/build \
                         $(scpfw-cmake-flags-y) --log-level=WARNING $(cmake-redirect)
```

In theory it should insert the redirection only when `make -s` (silent) is used.

etienne-lms:
Slient Works fine. Thanks for the tip.

jforissier:
Two more things below, then you can apply:
...to all commits.

etienne-lms:
Commit "libutils: IS_ENABLED2() for CFG* switches": updated inline description with "...to avoid triggering false positives with checkpatch".
@jforissier, I applied your review tag. @jenswi-linaro, yours is still applied

Don't merge yet, I've rebased and squashed the fixup commits but appended a new one.

Added fixup commit is to prevent the CI from failing when https://github.com/ARM-software/SCP-firmware/pull/733 is merged as there'll be 2 wrapper header file present to define `fwk_expect()`. With this fixup commit, SCP-fimware one will take precendence when present and build will be fine. Once that scpfw P-R is merged, I'll remove the header file from optee_os source tree.

etienne-lms:
I update this P-R (and rebased). The series is unmodified but the appended fixup last commit.

https://github.com/ARM-software/SCP-firmware/pull/733 has been rejected therefore we must host _fwk_assert.h_ wrapper file in _core/lib/scmi-server/include/_ until we find a better solution so I removed the fixup commit mentionned in comment https://github.com/OP-TEE/optee_os/pull/5686#issuecomment-1347224111.

I've added a fixup commit for "core: lib: scmi-server: Build a SCMI server from SCP-firmware" since a macro was missing for SCMI base protocol to be embedded in SCP-firmware. Please can one check this fixup. All commits alread have review tag. If the fixup is ok and CI is fine, i'll squash and request this P-R to be merged.

jenswi-linaro:
Why do we need a `fwk_assert.h` at all, doesn't the SCP version work?

etienne-lms:
> Why do we need a fwk_assert.h at all, doesn't the SCP version work?
>
 
SCP fmw implementation is based on toolchain _assert.h_ implementation. AFAICT GCC, LLVM, Arm compiler are based on something like the below:
```c
#define assert(cond)    ((cond) ? (void)0 : __assert_fail(#cond, __FILE__, __LINE__))
```
Such makes `assert()` to possibly return a value despite the function manpage that says "_**RETURN VALUE** No value is returned._". SCP-fmw maintainers prefer to preserve this legacy implementation. Using this possible return value of `assert()` allows SCP-fmw framework to implement a helper function named `fwk_expect()`:
```c
/*!
 * \def fwk_expect
 *
 * \brief Expect the success of a condition in an expression.
 *
 * \details In debug builds, the macro will evaluate the condition and trigger
 *      an assertion if it fails. In release builds, or if running tests, the
 *      macro will evaluate the condition and discard its result.
 *
 *      #### Example
 *
 *      \code{.c}
 *      if (!fwk_expect(rand() != 42))
 *          return FWK_E_STATE;
 *
 *      return FWK_E_SUCCESS;
 *      \endcode
 *
 *      In this example, the code will assert in a debug build if `rand` is
 *      equal to 42. In a release build the branch will be taken.
 *
 * \param condition Condition to test.
 *
 * \return The value of `condition`.
 */

#if defined(NDEBUG) || defined(BUILD_TESTS)
#    define fwk_expect(condition) (bool)__builtin_expect((condition), 1)
#else
#    define fwk_expect(condition) (bool)(fwk_check(condition), 1)
#endif
```
Changing OP-TEE OS `assert()` definition the same way would resolve the issue.

etienne-lms:
I've added 2 fixup commits for issues reported by  @vingu-linaro.

etienne-lms:
Ok. Please have a look at https://github.com/OP-TEE/optee_os/pull/5750.

etienne-lms:
I'll rebase this P-R once #5750 is merged.
@jforissier, @jenswi-linaro, do you ack the 5 appended fixup commits?

jforissier:
> I'll rebase this P-R once #5750 is merged. @jforissier, @jenswi-linaro, do you ack the 5 appended fixup commits?

Fine with me, aside 5750 the whole series is:

core/lib/scmi-server/sub.mk
```diff
@@ -70,11 +70,14 @@ cflags-lib-y += -Wno-undef \
 		-Wno-unused-but-set-variable \
 		-Wno-suggest-attribute=format
 
+# Notification implementation has strict aliasing issues
+cppflags-lib-$(CFG_SCPFW_NOTIFICATION) += -Wno-strict-aliasing
```
jenswi-linaro, line 74: This should be cflags-lib...

jenswi-linaro (commented):
With the comment addressed:

core/lib/scmi-server/sub.mk
etienne-lms: Right, OK, thanks.

etienne-lms:
Thanks for the reviews.
Removed commit that added _fwk_assert.h_.
Fixed _core/lib/scmi-server/sub.mk_ :  s/cppflags/cflags/.
Rebased, tag applied.

PR #5689, opened 2022-12-06
Update reported version to GP Core Internal API v1.3.1

jenswi-linaro:
Updates the reported version to 1.3.1. Three new defines:
- TEE_CORE_API_REQUIRED_MAJOR_VERSION
- TEE_CORE_API_REQUIRED_MINOR_VERSION
- TEE_CORE_API_REQUIRED_MAINTENANCE_VERSION are added by the standard as a way for the TA to specify required version of the API. OP-TEE only supports downgrading to version 1.1.

A simplified OP-TEE specific method is also provided: Adds the configuration option CFG_TA_OPTEE_CORE_API_COMPAT_1_1 which enables TEE Internal Core API v1.1 compatibility for in-tree TAs.

The TA dev kit is also updated to recognize
CFG_TA_OPTEE_CORE_API_COMPAT_1_1 and set define
__OPTEE_CORE_API_COMPAT_1_1 to 1 if set.

These new defines does not do anything yet, but in following commits functions and types will be updated gradually until all functions and types changed in version 1.3.1 compared to the ones in v1.1 have been updated.

### DIFF
--- a/lib/libutee/include/tee_api_defines.h
+++ b/lib/libutee/include/tee_api_defines.h
@@ -1,14 +1,96 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
  * Copyright (c) 2014, STMicroelectronics International N.V.
+ * Copyright (c) 2022, Linaro Limited
  */
 
-/* Based on GP TEE Internal Core API Specification Version 1.1 */
+/* Based on GP TEE Internal Core API Specification Version 1.3.1 */
 
 #ifndef TEE_API_DEFINES_H
 #define TEE_API_DEFINES_H
 
-#define TEE_INT_CORE_API_SPEC_VERSION     0x0000000A
+#define TEE_CORE_API_MAJOR_VERSION		1U
+#define TEE_CORE_API_MINOR_VERSION		3U
+#define TEE_CORE_API_MAINTENANCE_VERSION	1U
+#define TEE_CORE_API_VERSION \
+			((TEE_CORE_API_MAJOR_VERSION << 24) | \
+			 (TEE_CORE_API_MINOR_VERSION << 16) | \
+			 (TEE_CORE_API_MAINTENANCE_VERSION << 8))
+#define TEE_CORE_API_1_3_1
+
+/*
+ * Below follows the GP defined way of letting a TA define that it wants an
+ * API compatible with version 1.1 or the latest. An alternative approach
+ * is to set __OPTEE_CORE_API_COMPAT_1_1, but that's an OP-TEE extension.
+ *
+ * The GP specs (>= 1.2) requires that only APIs defined in the indicated
+ * version SHALL be made available when using this mechanism. However, that
+ * is far beyond what ordinary standards requires as they permit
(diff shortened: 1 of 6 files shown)


jenswi-linaro:
This should fix a recent regression due to https://github.com/OP-TEE/optee_test/pull/629.
Note that the commit "Update reported version to GP Core Internal API v1.3.1" also is part of https://github.com/OP-TEE/optee_os/pull/5688.

lib/libutee/include/tee_api_defines.h
```diff
@@ -1,14 +1,96 @@
...
 
-#define TEE_INT_CORE_API_SPEC_VERSION     0x0000000A
+#define TEE_CORE_API_MAJOR_VERSION		1U
+#define TEE_CORE_API_MINOR_VERSION		3U
+#define TEE_CORE_API_MAINTENANCE_VERSION	1U
+#define TEE_CORE_API_VERSION \
+			((TEE_CORE_API_MAJOR_VERSION << 24) | \
+			 (TEE_CORE_API_MINOR_VERSION << 16) | \
+			 (TEE_CORE_API_MAINTENANCE_VERSION << 8))
+#define TEE_CORE_API_1_3_1
```
jforissier, line 19: Is this useful?

lib/libutee/include/tee_api_defines.h
```diff
@@ -1,14 +1,96 @@
...
+#if defined(TEE_CORE_API_REQUIRED_MAINTENANCE_VERSION) && \
+	TEE_CORE_API_REQUIRED_MAINTENANCE_VERSION != 0
+#error "Required maintenance version not supported"
+#endif
+#endif
+#endif
+
+/*
+ * For backwards compatibility with v1.1 as provided by OP-TEE back
+ * then, define __OPTEE_CORE_API_COMPAT_1_1 to 1.
```
jforissier, line 61: Should we say "up to 3.19.0" rather than "back then"?

lib/libutee/include/tee_api_defines.h
```diff
@@ -1,14 +1,96 @@
...
+ *
+ * Some versions of the GP specs has introduced backwards incompatible
```
jforissier, line 63: s/has/have/ (plural)

jforissier (commented):
A few comments below, in any case:

Thanks for extracting this commit from the GP 1.3.1 series :+1:

lib/libutee/include/tee_api_defines.h
jenswi-linaro: It's required by the specification.

lib/libutee/include/tee_api_defines.h
jenswi-linaro: Makes sense, I'll fix.

jenswi-linaro:
Comments address, squashed and tag applied.

jforissier:
@etienne-lms please check if this is good with you. I'd like this merged ASAP to "un-break" the CI. Thanks!

jenswi-linaro:
Squashed the update.

PR #5693, opened 2022-12-08
libutee: support RISC-V syscall

liushiwei007:
Add 64-bit RISC-V TA system call function.
Modify the utee_syscalls_asm.S file to be generic.

Signed-off-by: liushiwei <liushiwei@eswincomputing.com>
Reviewed-by: chenchaokai <chenchaokai@eswincomputing.com>


lib/libutee/arch/riscv/utee_syscalls_rv64.S
```diff
@@ -0,0 +1,32 @@
...
+
+        .section .text
+
+        .macro UTEE_SYSCALL name, scn, num_args
+	FUNC \name , :
+
+	.if \num_args > TEE_SVC_MAX_ARGS || \num_args > 8
+	.error "Too many arguments for syscall"
+	.endif
+	addi	sp, sp, -16
```
jenswi-linaro, line 17: Why are you doing anything with the stack pointer here?

liushiwei007: When there are more parameters, use the stack to pass two parameters。

jenswi-linaro: There will never be more than eight, I believe it should be possible to copy the pattern from `utee_syscalls_a64.S`.

liushiwei007: Following our original logic, RISC-V uses the a0-a7 register to pass parameters, 8 parameters, plus the system call number and number of parameters, at most 10 parameters need to be passed, so the stack is used to pass 2 parameters.  As you suggested, the number of arguments is not passed, and the system call number is passed as t0 so that stack memory is not used

jenswi-linaro (commented):

PR #5697, opened 2022-12-08
core: crypto: Fix tee_obj_attr_copy_from for X25519

sohhass:
Hey folks!

This PR fixes a missing check for X25519 in `tee_obj_attr_copy_from`.

We should make sure in `tee_obj_attr_copy_from` if the source object is of type `TEE_TYPE_X25519_KEYPAIR` the destination object should be of type `TEE_TYPE_X25519_PUBLIC_KEY` i.e. the function should extract public key attributes if presented with a key pair object.

The `TEE_CopyObjectAttributes1` API as specified in the `TEE Internal Core API Specification v1.2` should check if the source object is of type *_KEYPAIR then the destination is a subset of type *_PUBLIC_KEY when populating the destination attributes. This condition is listed in `Table 5-11` of the specifications.


jenswi-linaro:

Please add a `Fixes: 90040fa4c81c ("core: crypto: add X25519 support")` in the commit message.

etienne-lms (commented):
I suggest you replace `'tee_obj_attr_copy_from'` with `tee_obj_attr_copy_from()` in commit message.

PR #5702, opened 2022-12-14
plat-totalcompute: change entrypoint of secure partition

davidson-k:
The default image offset is changed from 0x1000 to 0x4000 to accomodate the boot protocol information.

Signed-off-by: Davidson K <davidson.kumaresan@arm.com>


jenswi-linaro (commented):

PR #5703, opened 2022-12-14
[plat-sam] sama5d2 improvements

clementleger:
This series embeds multiple improvements for the sama5d2 support. It mainly target at improving secure status for multiple devices as well as enabling reboot on watchdog reset.

### DIFF
--- a/core/drivers/atmel_rstc.c
+++ b/core/drivers/atmel_rstc.c
@@ -6,6 +6,8 @@
 #include <drivers/atmel_rstc.h>
 #include <io.h>
 #include <kernel/dt.h>
+#include <matrix.h>
+#include <sama5d2.h>
 #include <tee_api_defines.h>
 #include <tee_api_types.h>
 #include <types_ext.h>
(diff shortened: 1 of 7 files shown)


clementleger:
Update: Fix unwrapped commit

etienne-lms (commented):

core/drivers/atmel_rstc.c
```diff
@@ -43,6 +45,11 @@ static TEE_Result atmel_rstc_probe(const void *fdt, int node,
 {
 	size_t size = 0;
 
+	if (_fdt_get_status(fdt, node) != DT_STATUS_OK_SEC)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	matrix_configure_periph_secure(AT91C_ID_SYS);
```
ldts, line 51: is it not an issue if it fails? shouldn't we at least pop an error on the console?

clementleger: @ldts: If it fails, then the `atmel_rstc_available()` will return false. This is used by the code that can invoke reset support. So it is actually "supported", the same can happen if the rstc node is lacking.

clementleger:
New version:
- Added  Acked-by: Etienne Carriere <etienne.carriere@linaro.org>

PR #5704, opened 2022-12-14
[RFC] User libname.map as version script if available

jenswi-linaro:
If a shared library is linked and a version script is available with the name `$(libname).map`, pass it to the linker as: `--version-script=$(libname).map`

This enables symbol versioning of the shared library.


etienne-lms:
reopen

jforissier:
A slightly modified/fixed/improved version of this is commit "lib: link shared user space libraries with version script if it exists" in draft PR#5833. Therefore I am closing this one.

PR #5707, opened 2022-12-15
core: crypto: fix TEE_ATTR_EDDSA_PREHASH interpretation

jenswi-linaro:
Commit 0aaad418ac8b ("core: crypto: add Ed25519 support") introduced support for the ED25519 algorithm. This included parsing a TEE_ATTR_EDDSA_PREHASH parameter that unfortunately was not fully compliant with the standard.  So fix this with a more strict interpretation of TEE_ATTR_EDDSA_PREHASH as described in the specification.

Fixes: 0aaad418ac8b ("core: crypto: add Ed25519 support")

@varder FYI

### DIFF
--- a/core/tee/tee_svc_cryp.c
+++ b/core/tee/tee_svc_cryp.c
@@ -2192,7 +2192,14 @@ tee_svc_obj_ed25519_parse_params(const TEE_Attribute *params, size_t num_params,
 	for (n = 0; n < num_params; n++) {
 		switch (params[n].attributeID) {
 		case TEE_ATTR_EDDSA_PREHASH:
-			*ph_flag = true;
+			if (params[n].content.value.b)
+				return TEE_ERROR_BAD_PARAMETERS;
+			if (!params[n].content.value.a)
+				*ph_flag = false;
+			else if (params[n].content.value.a == 1)
+				*ph_flag = true;
+			else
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

jenswi-linaro:
Tag applied.

jenswi-linaro:
@xiaoxuZeng in OP-TEE we work with tags as a way of tracking how it has been reviewed.
After a deep review "Reviewed-by: Name <mailaddress>"
After a not so deep review "Acked-by: Name <mailaddress>"

Feel free to add a tag in a comment if you agree with this patch. You can see how Etienne supplied his tag above.

ldts:
isnt it easier to read in this form?

```
	if (params[n].content.value.a > 1 ) 

	if (params[n].content.value.a)
		*ph_flag = true;
	else
		*ph_flag = false; 
```

jenswi-linaro:
You could write it as
``` C
	if (params[n].content.value.a > 1 || params[n].content.value.b) 

	*ph_flag = params[n].content.value.a;
```
too. I still prefer my original proposal, but if there's a strong push for an alternative I'll update it.

varder:
Reviewed-by: Valerii Chubar <valerii_chubar@epam.com>

jenswi-linaro:
Tag applied.

PR #5708, opened 2022-12-15
drivers: stm32_bsec: miscellaneous fixes on BSEC driver

GseoC:
This P-R implements various fixes and coding style updates for the BSEC driver.


core/include/drivers/stm32_bsec.h
```diff
@@ -10,6 +10,9 @@
 #include <stdint.h>
 #include <tee_api.h>
 
+#define BSEC_BITS_PER_WORD		(8U * sizeof(uint32_t))
+#define BSEC_BYTES_PER_WORD		(sizeof(uint32_t))
```
etienne-lms, line 24: can remove outer parentheses

core/drivers/stm32_bsec.c
```diff
@@ -430,14 +423,13 @@ TEE_Result stm32_bsec_permanent_lock_otp(uint32_t otp_id)
 TEE_Result stm32_bsec_write_debug_conf(uint32_t value)
 {
 	TEE_Result result = TEE_ERROR_GENERIC;
-	uint32_t masked_val = value & BSEC_DEN_ALL_MSK;
```
etienne-lms, line 423: Previous code prevented caller from writting into reserved bits (that should remain to 0 as per ref manual). I think we should rpeserved the mask or at least assert that `value` sets only valid bits (`assert(!(value & ~BSEC_DEN_ALL_MSK));`).

etienne-lms (commented):
* "drivers: stm32_bsec: rework and move BITS_PER_WORD"
 (with a comment)
* "drivers: stm32_bsec: use U() for unsigned constants"
* "drivers: stm32_bsec: correct address in lock support for any upper_base value"
(with a comment: header line too large + an emprty line to remove in commit message)
* "drivers: stm32_bsec: use _fdt_reg* API"

core/include/drivers/stm32_bsec.h
GseoC: Ok :)

core/drivers/stm32_bsec.c
GseoC: Actually, the BSEC_DEN_ALL_MSK isn't correct atm as it covers platform-dependant reserved bits that shouldn't be set (Bit 8 and 0 for STM32MP13x and Bit 0 for STM32MP15x).
A way of doing it would be to create a specific mask for STM32MP13x and a different one for 15x platforms. But then, i

The check is fully removed here to simplify code and transfer the responsibility to check what has been written in this register to the caller of the function. Now, it should be managed by caller with a sequence read/modify/write to avoid update of reserved bits

etienne-lms: Ok. Fine it is caller responsibility.
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "drivers: stm32_bsec: move debug defines to include".

GseoC:
Pushed with comment addressed and tags applied

GseoC:
Corrected checkpatch error

jforissier:
Thanks. Ignoring the CI error, it is fixed in master.

PR #5709, opened 2022-12-15
FIX build failure: versal_pm

ldts:

### DIFF
--- a/core/arch/arm/plat-versal/conf.mk
+++ b/core/arch/arm/plat-versal/conf.mk
@@ -64,3 +64,6 @@ CFG_VERSAL_CRYPTO_DRIVER ?= y
 
 # SHA3-384 crypto engine
 CFG_VERSAL_SHA3_384 ?= y
+
+# PM driver
+CFG_VERSAL_PM ?= y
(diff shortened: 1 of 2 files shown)


ldts:
hi @jforissier could we merge this fix before EOY please?

jforissier:
For commit "drivers: versal_pm: fix compilation issue"

> The structure used "struct ipi_cmd" was renamed to
> "struct versal_ipi_cmd" causing a build failure.

When was it renamed? I can see that `struct versal_ipi_cmd` was introduced in commit 4b46e0e81 but I find no trace of a `struct ipi_cmd` being defined anywhere. That said and with a proper description:

I think we need more build configurations in `.github/workflows/ci.yml` to avoid these kind of errors.

jforissier:
For commit "versal: enable the PM driver."

No trailing `.` please.

ldts:
@jforissier yes I forgot to enable in ci hence why it went unnoticed. Now that it is enabled by default this should not happen again.

The initial ipi driver defined `struct ipi_cmd` which after review became `struct versal_ipi_cmd`; when we reviewed the PM driver I might have had some of this old code in the driver

jforissier:
> @jforissier yes I forgot to enable in ci hence why it went unnoticed. Now that it is enabled by default this should not happen again.

Yeah sorry I found out later when reading the second patch ;)
 
> The initial ipi driver defined `struct ipi_cmd` which after review became `struct versal_ipi_cmd`; when we reviewed the PM driver I might have had some of this old code in the driver

OK, got it, it would be less confusing to just say "versal_pm.c uses a non existent struct name: ipi_cmd, should be versal_ipi_cmd" or something like that.

ldts:
@jforissier I think this is done.

jforissier:
@ldts indeed, thanks.

PR #5713, opened 2022-12-16
Add BSEC otp provisionning

GseoC:
Add a way for non-secure world to provision One Time Programmed resources using "st,non-secure-otp" device tree property.

### DIFF
--- a/core/drivers/stm32_bsec.c
+++ b/core/drivers/stm32_bsec.c
@@ -713,8 +713,40 @@ static void bsec_dt_otp_nsec_access(void *fdt, int bsec_node)
 			}
 		}
 
-		if (!fdt_getprop(fdt, bsec_subnode, "st,non-secure-otp", NULL))
+		/* Handle different kinds of non-secure accesses */
+		if (fdt_getprop(fdt, bsec_subnode,
+				"st,non-secure-otp-provisioning", NULL)) {
+			bool locked = false;
+			bool locked_2 = false;
+
+			/* Check if write of OTP is locked */
+			if (stm32_bsec_read_permanent_lock(otp_id, &locked))
+				panic("Cannot read permanent lock");
+
+			/*
+			 * Check if fuses of the subnode
+			 * have the same lock status
+			 */
+			for (i = 1; i < (reg_size / sizeof(uint32_t)); i++) {
+				if (stm32_bsec_read_permanent_lock(otp_id + i,
+								   &locked_2))
+					panic("Cannot read permanent lock");
+
+				if (locked != locked_2) {
+					EMSG("Inconsistent status OTP ID %u",
+					     otp_id + i);
+					locked = true;
+				}
+			}
+
+			if (locked) {
+				DMSG("BSEC: OTP locked");
+				continue;
+			}
+		} else if (!fdt_getprop(fdt, bsec_subnode, "st,non-secure-otp",
+					NULL)) {
 			continue;
+		}
 
 		if ((reg_offset % sizeof(uint32_t)) ||
 		    (reg_size % sizeof(uint32_t)))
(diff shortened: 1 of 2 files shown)


core/drivers/stm32_bsec.c
```diff
@@ -616,8 +616,41 @@ static void bsec_dt_otp_nsec_access(void *fdt, int bsec_node)
 			}
 		}
 
-		if (!fdt_getprop(fdt, bsec_subnode, "st,non-secure-otp", NULL))
+		/* Handle different kinds of non-secure accesses */
+		if (fdt_getprop(fdt, bsec_subnode,
+				"st,non-secure-otp-provisioning", NULL)) {
+			bool locked = false;
+			bool lcked_next = false;
```
etienne-lms, line 720: s/lcked_next/locked_next/g

core/drivers/stm32_bsec.c
```diff
@@ -713,8 +713,41 @@ static void bsec_dt_otp_nsec_access(void *fdt, int bsec_node)
...
+
+			/* Check if write of OTP is locked */
+			if (stm32_bsec_read_permanent_lock(otp_id, &locked))
+				panic("BSEC: Couldn't read permanent lock at init");
```
etienne-lms, line 724: maybe use the same rtace message as below: `Cannot read permanent lock`

core/drivers/stm32_bsec.c
```diff
@@ -713,8 +713,41 @@ static void bsec_dt_otp_nsec_access(void *fdt, int bsec_node)
...
+			 * Check if fuses of the subnode
+			 * have the same lock status
+			 */
+			for (i = 1; i < (reg_size / sizeof(uint32_t)); i++) {
+				if (stm32_bsec_read_permanent_lock(otp_id + i,
+								   &lcked_next))
+					panic("Cannot read permanent lock");
+
+				if (locked != lcked_next) {
+					EMSG("Inconsistent status OTP id %u",
```
etienne-lms, line 736: s/id/ID/

core/drivers/stm32_bsec.c
```diff
@@ -713,8 +713,41 @@ static void bsec_dt_otp_nsec_access(void *fdt, int bsec_node)
...
+					     otp_id + i);
+					locked = true;
+					continue;
```
etienne-lms, line 739: `continue` has no effect here even if the implementation looks correct to me. I suggest to remove it.

core/drivers/stm32_bsec.c
GseoC: Change to locked_2. I did this because of character limitation

core/drivers/stm32_bsec.c
GseoC: Sure

core/drivers/stm32_bsec.c
GseoC: Ack

core/drivers/stm32_bsec.c
GseoC: Yes indeed...

GseoC:
Hello @jforissier @etienne-lms , is there a reason QEMU check failed? I'm not finding an explicit reason in the log

jforissier:
> Hello @jforissier @etienne-lms , is there a reason QEMU check failed? I'm not finding an explicit reason in the log

Nothing obvious indeed... the fact that we don't have the console output doesn't help. I'm restarting the job and also running it locally to see if I can reproduce the issue. In any case there is next to zero chance that the failure could be caused by this PR I think.

etienne-lms (commented):

GseoC:
Force-pushed with tags applied

PR #5716, opened 2022-12-19
riscv: kernel: implement time and delay routines for RISC-V

maroueneboubakri:
Adds a core local interruptor (CLINT) driver and implements TEE time source based on RISC-V's time registers. Finally it implements architecture-specific functions for delay.


jforissier (commented):
Reviewing commit "riscv: core: implement core local interruptor (clint) driver":
With my comments addressed/anwered:

core/arch/riscv/include/riscv.h
```diff
@@ -292,6 +292,29 @@ static inline __noprof void uret(void)
 	asm volatile("uret");
 }
 
+static inline __noprof uint64_t read_time(void)
+{
+	uint64_t time;
+	uint32_t hi __maybe_unused, lo __maybe_unused;
```
jforissier, line 298: Coding style: one variable per line and initialize all variables (= 0).

core/arch/riscv/include/riscv.h
```diff
@@ -292,6 +292,29 @@ static inline __noprof void uret(void)
...
+#ifdef CFG_RISCV_M_MODE
+	time = clint_get_mtime();
+#else /*CFG_RISCV_S_MODE*/
+#ifdef RV32
+	do {
+		hi = read_csr(timeh);
+		lo = read_csr(time);
+	} while (hi != read_csr(timeh));
+
+	time =  SHIFT_U32((uint64_t)hi, 32) | lo;
```
jforissier, line 309: `SHIFT_U64(hi, 32) | lo`

core/arch/riscv/include/riscv.h
```diff
@@ -292,6 +292,29 @@ static inline __noprof void uret(void)
...
+#else /*RV64*/
+	time = read_csr(time);
```
jforissier, line 311: How about `time = rdtime();` instead? (the two different `time` are  bit confusing)

jforissier (commented):
Comments for "riscv: include: riscv.h: implement read_time() function".

core/arch/riscv/include/riscv.h
maroueneboubakri: Correct.

jforissier (commented):
For "core: kernel: move tee_time.c and tee_time_ree.c to core/kernel":

I would personally drop the `kernel:` prefix => "core: move tee_time.c and tee_time_ree.c to core/kernel", anyways:

core/arch/riscv/kernel/tee_time_rdtime.c
```diff
@@ -0,0 +1,31 @@
...
+#include <kernel/tee_time.h>
+#include <kernel/time_source.h>
+#include <riscv.h>
+#include <utee_defines.h>
+
+static TEE_Result riscv_get_sys_time(TEE_Time *time)
+{
+	uint64_t _time;
+
+	_time = read_time();
```
jforissier, line 16: `uint64_t tm =  read_time();` because we want all variables initialized (coding rule), and the underscore is not really justified here (not like reserved names in user libraries).

core/arch/riscv/plat-spike/platform_config.h
```diff
@@ -14,4 +14,10 @@
 #define HTIF_BASE	0x40008000
 #endif
 
+#ifdef CFG_RISCV_MTIME_RATE
+#define RISCV_MTIME_RATE CFG_RISCV_MTIME_RATE
+#else
+#define RISCV_MTIME_RATE 1000000
+#endif
+
```
jforissier, line 22: Drop this. `CFG_RISCV_MTIME_RATE` is always defined.

core/arch/riscv/kernel/tee_time_rdtime.c
```diff
@@ -0,0 +1,31 @@
...
+
+	time->seconds = _time / RISCV_MTIME_RATE;
+	time->millis = (_time % RISCV_MTIME_RATE) /
+		       (RISCV_MTIME_RATE  / TEE_TIME_MILLIS_BASE);
```
jforissier, line 20: Use `CFG_RISCV_MTIME_RATE` directly, and also a helper variable would help:
```C
	uint64_t rate = CFG_RISCV_MTIME_RATE;

	time->seconds = tm / rate;
	time->millis = (tm % rate) / (rate / TEE_TIME_MILLIS_BASE);
```

jforissier (commented):
Reviewing "riscv: kernel: add time source based on time registers"

jforissier (commented):
For " core: kernel: delay: sort-out architecture-independant code from arch dir":

core/arch/riscv/include/kernel/delay_arch.h
```diff
@@ -0,0 +1,24 @@
...
+#define __KERNEL_DELAY_ARCH_H
+
+#include <platform_config.h>
+#include <riscv.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+static inline uint64_t timeout_init_us(uint32_t us)
+{
+	return read_time() + (((uint64_t)us * RISCV_MTIME_RATE) / 1000000ULL);
```
jforissier, line 16: So given my previous comment this shoud use `CFG_RISCV_MTIME_RATE`, but what if `CFG_RISCV_M_MODE` is not set?

jforissier (commented):
Comment for "riscv: kernel: include: add delay_arch.h".

core/arch/riscv/include/riscv.h
maroueneboubakri: `read_time()` moved to a separate new header file `time.h`

core/arch/riscv/include/kernel/delay_arch.h
maroueneboubakri: Didn't get it.

jforissier: Sorry...
1. `RISCV_MTIME_RATE` does not exist (assuming my [comment](https://github.com/OP-TEE/optee_os/pull/5716#discussion_r1061332653) is addressed) so that should be `CFG_RISCV_MTIME_RATE`.
2. But then I am wondering, is `CFG_RISCV_MTIME_RATE` always defined and meaningful? I am wondering this for two reasons: (a) it is currently only in `plat-spike`, what about other platforms? and (b) there is MTIME in the name which implies M mode doesn't it? so what about platforms configured for a different mode?

maroueneboubakri: Maybe the name `CFG_RISCV_MTIME_RATE ` is confusing due to the M, but that variable holds the clock frequency of `mtime` which is different across platforms and should be always defined.  The `mtime` register depends on a global real time clock. In S-Mode we read it through CLINT, in M-Mode we directly read the CSR, We can find better name otherwise.

jforissier: Ah, that makes sense. Excuse my poor knowledge of the RISC-V architecture :-/ I suppose the name is OK then.

maroueneboubakri: All comments addressed, any tag to apply to the remaining commits ?

jforissier: You should also drop the extra parentheses `+ ( ... );`. Then please apply:

to "riscv: kernel: include: add delay_arch.h"

maroueneboubakri:

jenswi-linaro:
For patches touching stuff contained under `/core` please use a `core: `  prefix.
It helps to get a quick view of the scope. The `kernel: ` part that you add is usually less useful, but it doesn't harm either. A `riscv: ` prefix only gives less information as it could be something riscv specific for TAs.

maroueneboubakri:
> For patches touching stuff contained under `/core` please use a `core: ` prefix. It helps to get a quick view of the scope. The `kernel: ` part that you add is usually less useful, but it doesn't harm either. A `riscv: ` prefix only gives less information as it could be something riscv specific for TAs.

Done.

jforissier:
"core: riscv: provide time.h":

jforissier:
" core: riscv: add time source based on time registers":

maroueneboubakri:
@jforissier done.

jenswi-linaro:

maroueneboubakri:

Applied.

PR #5721, opened 2022-12-20
riscv: kernel: implement boot flow for RISC-V

maroueneboubakri:
Implement boot flow in boot.c and entry.S

**entry.S** provides core's single entry point for RV32/RV64 in S/M Modes. For now it performs: booting primary and secondary harts. Setting stack pointer, thread pointer (to `thread_core_local`), supervisor address translation and protection register, clearing BSS...etc and calls to appropriate functions to initialize the MMU and continue to boot flow from boot.c.

**boot.c** provides an implementation of `init_tee_runtime()`, `plat_primary_init_early()`, `boot_init_primary_early()`, `boot_init_primary_late()`, `boot_init_secondary()` and helper functions.  For now `init_sec_mon()` is kept to be replaced later by a routine to initialize SBI implementation (to for example, probe for available SBI extensions).

### DIFF
--- a/core/arch/riscv/kernel/boot.c
+++ b/core/arch/riscv/kernel/boot.c
@@ -1,0 +1,151 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2022-2023 NXP
+ */
+
+#include <assert.h>
+#include <compiler.h>
+#include <config.h>
+#include <console.h>
+#include <keep.h>
+#include <kernel/boot.h>
+#include <kernel/linker.h>
+#include <kernel/misc.h>
+#include <kernel/panic.h>
+#include <kernel/thread.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <mm/tee_mm.h>
+#include <mm/tee_pager.h>
+#include <platform_config.h>
+#include <riscv.h>
+#include <sbi.h>
+#include <stdio.h>
+#include <trace.h>
+#include <util.h>
+
+#define PADDR_INVALID               ULONG_MAX
+
+paddr_t start_addr;
+unsigned long boot_args[4];
+
+uint32_t sem_cpu_sync[CFG_TEE_CORE_NB_CORE];
+
+void init_sec_mon(unsigned long nsec_entry __maybe_unused)
+{
+	assert(nsec_entry == PADDR_INVALID);
+	/* Do nothing as we don't have a secure monitor */
+}
+
+#ifdef CFG_RISCV_S_MODE
+static void start_secondary_cores(void)
+{
+	size_t i = 0;
+	size_t pos = get_core_pos();
+
+	for (i = 0; i < CFG_TEE_CORE_NB_CORE; i++)
+		if (i != pos && IS_ENABLED(CFG_RISCV_SBI) &&
+		    sbi_boot_hart(i, start_addr, i))
+			EMSG("Error starting secondary hart %zu", i);
+}
+#endif
+
+static void init_runtime(void)
+{
+	malloc_add_pool(__heap1_start, __heap1_end - __heap1_start);
+
+	IMSG_RAW("\n");
+}
+
+void init_tee_runtime(void)
+{
+	core_mmu_init_ta_ram();
+	call_preinitcalls();
+	call_initcalls();
+}
+
+static void init_primary(unsigned long nsec_entry)
+{
+	/*
+	 * Mask asynchronous exceptions before switch to the thread vector
+	 * as the thread handler requires those to be masked while
+	 * executing with the temporary stack. The thread subsystem also
+	 * asserts that the foreign interrupts are blocked when using most of
+	 * its functions.
+	 */
+	thread_set_exceptions(THREAD_EXCP_ALL);
+
+	init_runtime();
+	thread_init_boot_thread();
+	thread_init_primary();
+	thread_init_per_cpu();
+	init_sec_mon(nsec_entry);
+}
+
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak void plat_primary_init_early(void)
+{
+}
+
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak void main_init_plic(void)
+{
+}
+
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak void main_secondary_init_plic(void)
+{
+}
+
+void boot_init_primary_early(unsigned long pageable_part __unused,
+			     unsigned long nsec_entry __unused)
+{
+	unsigned long e = PADDR_INVALID;
+
+	init_primary(e);
+}
+
+void boot_init_primary_late(unsigned long fdt __unused)
+{
+	IMSG("OP-TEE version: %s", core_v_str);
+	if (IS_ENABLED(CFG_WARN_INSECURE)) {
+		IMSG("WARNING: This OP-TEE configuration might be insecure!");
+		IMSG("WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html");
+	}
+	IMSG("Primary CPU initializing");
+	main_init_plic();
+	init_tee_runtime();
+	call_finalcalls();
+	IMSG("Primary CPU initialized");
+
+#ifdef CFG_RISCV_S_MODE
+	start_secondary_cores();
+#endif
+}
+
+static void init_secondary_helper(unsigned long nsec_entry)
+{
+	size_t pos = get_core_pos();
+
+	IMSG("Secondary CPU %zu initializing", pos);
+
+	/*
+	 * Mask asynchronous exceptions before switch to the thread vector
+	 * as the thread handler requires those to be masked while
+	 * executing with the temporary stack. The thread subsystem also
+	 * asserts that the foreign interrupts are blocked when using most of
+	 * its functions.
+	 */
+	thread_set_exceptions(THREAD_EXCP_ALL);
+
+	thread_init_per_cpu();
+	init_sec_mon(nsec_entry);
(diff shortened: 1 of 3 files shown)


core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak void main_secondary_init_plic(void)
+{
+}
+
+/*
+ * Note: this function is weak just to make it possible to exclude it from
+ * the unpaged area so that it lies in the init area.
+ */
+void __weak boot_init_primary_early(unsigned long pageable_part __unused,
```
jenswi-linaro, line 114: This function don't have to be `__weak` since riscv doesn't support paging. Same below for `boot_init_primary_late()`

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+#ifdef CFG_RISCV_S_MODE
+static void start_secondary_cores(void)
+{
+	int ret;
+	uint32_t i;
+	size_t pos = get_core_pos();
+
+	for (i = 0; i < CFG_TEE_CORE_NB_CORE; i++) {
+		if (i != pos) {
+#ifdef CFG_RISCV_SBI
```
jenswi-linaro, line 50: Please us `if (IS_ENABLED(CFG_RISCV_SBI))` instead.

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+			ret = sbi_boot_hart(i, _start_addr, (unsigned long)i);
```
jenswi-linaro, line 51: The case is not needed

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+#include <sbi.h>
+#include <stdio.h>
+#include <trace.h>
+#include <util.h>
+
+#include <platform_config.h>
+
+#define PADDR_INVALID               ULONG_MAX
+
+unsigned long _start_addr;
```
jenswi-linaro, line 30: Why do you have a `_` prefix? Same below with `_boot_args`.

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void init_tee_runtime(void)
+{
+	core_mmu_init_ta_ram();
+	call_preinitcalls();
+	call_initcalls();
+}
+
+static void init_primary(unsigned long pageable_part __unused,
```
jenswi-linaro, line 76: I don't suppose `pageable_part` will ever be needed so you could drop it.

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+#include <platform_config.h>
+
+#define PADDR_INVALID               ULONG_MAX
+
+unsigned long _start_addr;
+unsigned long _boot_args[4];
+
+uint32_t sem_cpu_sync[CFG_TEE_CORE_NB_CORE];
+
+void init_sec_mon(unsigned long nsec_entry __maybe_unused)
```
jenswi-linaro, line 35: Will `nsec_entry` ever be needed here?

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+{
+	assert(nsec_entry == PADDR_INVALID);
+	/* Do nothing as we don't have a secure monitor */
+}
+
+#ifdef CFG_RISCV_S_MODE
+static void start_secondary_cores(void)
+{
+	int ret;
+	uint32_t i;
```
jenswi-linaro, line 45: Why not `size_t` as for `pos` below?

core/arch/riscv/kernel/boot.c
```diff
@@ -0,0 +1,169 @@
...
+	size_t pos = get_core_pos();
+
+	for (i = 0; i < CFG_TEE_CORE_NB_CORE; i++) {
+		if (i != pos) {
+#ifdef CFG_RISCV_SBI
+			ret = sbi_boot_hart(i, _start_addr, (unsigned long)i);
+#else
+			ret = -1;
+#endif /*CFG_RISCV_SBI*/
+			if (ret != 0)
```
jenswi-linaro, line 55: What's the point with this when `sbi_boot_hart()` isn't called?

core/arch/riscv/kernel/boot.c
maroueneboubakri: Is `__weak ` related to paging only ? my understanding exporting this function as `__weak ` such cab be overridden, by platform/main.c for example.

core/arch/riscv/kernel/boot.c
maroueneboubakri: We will not able to start the secondary hart when we are in S-Mode.

core/arch/riscv/kernel/boot.c
maroueneboubakri: Depends on how REE and TEE will communicate (though SBI like ARM's SMC or through messaging like RPMSG).
Let's keep it for now.

core/arch/riscv/kernel/boot.c
maroueneboubakri: Removed.

core/arch/riscv/kernel/boot.c
jenswi-linaro: Sure, but in this case it's pretty clear from the comment what was the intent with `__weak` for this function.

maroueneboubakri: Comment removed.

jenswi-linaro: > Comment removed.

That wasn't my point. You seem to try to keep a bit of the structure from arch/arm booting. In arch/arm platforms are not supposed to override `boot_init_primary_early()`. So why do you insist that it must be possible in arch/riscv?

maroueneboubakri: Ok removed.

core/arch/riscv/kernel/entry.S
```diff
@@ -0,0 +1,233 @@
...
+
+LOCAL_FUNC reset_primary , : , .identity_map
+UNWIND(	.cantunwind)
+
+	bootargs_entry
+
+	/*
+	 * Zero bss
+	 */
+1:
```
jenswi-linaro, line 134: Is this used?

core/arch/riscv/kernel/entry.S
```diff
@@ -0,0 +1,233 @@
...
+	csrw	CSR_SATP, zero
+	set_sp
+	set_tp
+
+	jal	thread_init_thread_core_local
+
+	jal	plat_primary_init_early
+	jal	console_init
+
+	/* Get number of ASID bits in HW */
```
jenswi-linaro, line 159: Couldn't this be done in C instead, at some point in `core_init_mmu_map()` before MMU is enabled?

core/arch/riscv/kernel/entry.S
```diff
@@ -0,0 +1,233 @@
...
+	la	tp, thread_core_local
+	mult	a0, a1, t0
+	add	tp, tp, t0
+	csrr	a0, CSR_XSCRATCH
+	sw	a0, THREAD_CORE_LOCAL_HART_ID(tp)
+.endm
+
+.macro set_satp
+	la	a1, boot_mmu_config
+	LDR	a0, CORE_MMU_CONFIG_SATP(a1)
```
jenswi-linaro, line 66: What does this mean? Is LDR a pseudo instruction that is expanded by the assembler? Where is `CORE_MMU_CONFIG_SATP` defined?

maroueneboubakri: Defined [here ](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/include/riscv.h#L20-L26) to load a word or doubleword depending on the target.
`CORE_MMU_CONFIG_SATP`  is defined in asm-defines, I'll add the commit to [PR#5701](https://github.com/OP-TEE/optee_os/pull/5701/).

jenswi-linaro:

maroueneboubakri:
Tag applied.

core/arch/riscv/kernel/entry.S
```diff
@@ -0,0 +1,221 @@
...
+
+	bootargs_entry
+
+	/*
+	 * Zero bss
+	 */
+1:
+	lla	t0, __bss_start
+	lla	t1, __bss_end
+	beq	t0, t1, 1f
```
etienne-lms, line 145: shouldn't this be `1b` not `1f`?

core/arch/riscv/kernel/entry.S
```diff
@@ -0,0 +1,221 @@
...
+	set_satp
+	cpu_is_ready
+
+	jal	boot_init_secondary
+	j	.
+END_FUNC reset_secondary
+
+LOCAL_FUNC unhandled_cpu , :
+	wfi
+	j unhandled_cpu
```
etienne-lms, line 197: prefer with a tab:
```
	j	unhandled_cpu

maroueneboubakri:
> 

Applied.

core/arch/riscv/kernel/entry.S
etienne-lms: is `1:` missing at line 150?

maroueneboubakri: Yes, added and removed the top `1:` label.

etienne-lms: ok, right. latest post is fine.

etienne-lms (approved):

PR #5730, opened 2022-12-27
core: mbedtls: fix use of SHA-256 crypto accelerated routines

jenswi-linaro:
The wrong name of a configuration was used to test in mbedtls if the accelerated SHA-256 routines should be used. Fix this by using the correct name CFG_CORE_CRYPTO_SHA256_ACCEL instead.

Fixes: 2fc5dc95a949 ("core: mbedtls: use SHA-256 crypto accelerated routines")


jforissier (approved):

jenswi-linaro:
Tag applied

PR #5735, opened 2023-01-02
plat-k3: Increase size of high DDR memory region

manorit2001:
With the addition of j784s4 in K3 devices, DRAM size is increased to 32GB.

Update the size of higher memory addresses to handle this.

### DIFF
--- a/core/arch/arm/plat-k3/platform_config.h
+++ b/core/arch/arm/plat-k3/platform_config.h
@@ -19,7 +19,7 @@
 #define DRAM0_SIZE      0x80000000
 
 #define DRAM1_BASE      0x880000000
-#define DRAM1_SIZE      0x80000000
+#define DRAM1_SIZE      0x780000000
 
 #define SCU_BASE        0x01800000
 #if defined(PLATFORM_FLAVOR_j721e) || defined(PLATFORM_FLAVOR_j784s4)


core/arch/arm/plat-k3/platform_config.h
```diff
@@ -19,7 +19,7 @@
 #define DRAM0_SIZE      0x80000000
 
 #define DRAM1_BASE      0x880000000
-#define DRAM1_SIZE      0x80000000
```
jforissier, line 22: Isn't there any drawback to supplying a larger `DRAM1_SIZE` than actually exists on some platforms?

manorit2001: I had confirmed with @glneo on this and he mentioned that it shouldn't create any issues for other platforms IIRC

jforissier: Thanks, then FWIW:

jenswi-linaro: No issues apart from allowing normal world to trick secure world into mapping non-existing physical memory.

jforissier: > No issues apart from allowing normal world to trick secure world into mapping non-existing physical memory.

That would be a DoS, not more severe as many other things that a compromised NW could do I suppose.

jenswi-linaro: > > No issues apart from allowing normal world to trick secure world into mapping non-existing physical memory.
> 
> That would be a DoS, not more severe as many other things that a compromised NW could do I suppose.

This is a new policy for how to handle non-secure memory.  We can keep this as a platform decision for now and later make it global once we're certain it's only a matter of DoS.

etienne-lms: Why not:
```c
#if defined(PLATFORM_FLAVOR_j784s4)
#define DRAM1_SIZE      0x780000000
#else
#define DRAM1_SIZE      0x80000000
#endif

manorit2001: > Why not:
> 
> ```c
> #if defined(PLATFORM_FLAVOR_j784s4)
> #define DRAM1_SIZE      0x780000000
> #else
> #define DRAM1_SIZE      0x80000000
> #endif
> ```

We could do that just didn't chose to as didn't see any critical security implications of increasing the DRAM size for the other platforms too, if it is a blocker then would re-roll the patch with the above change, let me know if this needs to be done

manorit2001: > > Why not:
> > ```c
> > #if defined(PLATFORM_FLAVOR_j784s4)
> > #define DRAM1_SIZE      0x780000000
> > #else
> > #define DRAM1_SIZE      0x80000000
> > #endif
> > ```
> 
> We could do that just didn't chose to as didn't see any critical security implications of increasing the DRAM size for the other platforms too, if it is a blocker then would re-roll the patch with the above change, let me know if this needs to be done

Hi @etienne-lms , is this required?

etienne-lms: In my opinion, we should not create a execution path in OP-TEE that allows normal world to freeze a bus or panic the system on demand. Maybe a boot loader can provide the information to the platform with a DT. Maybe a config switch.

etienne-lms: The easiest to get this change merged the one in https://github.com/OP-TEE/optee_os/pull/5735#discussion_r1063273293 unless @glneo acks this change, IMHO.

etienne-lms: I missed https://github.com/OP-TEE/optee_os/pull/5735#discussion_r1059861680. This change is fine IM(Strong)O.
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` if it helps.

glneo: The size of DDR is not related to the SoC, but the board the SoC lives on, it just so happens the J784S4-EVM was the first of our EVMs to ship with this much DDR. All K3 SoCs support this much(and more) DDR and should support this larger size (at least until we can set this size dynamically).

`Acked-by: Andrew Davis <afd@ti.com>`

jforissier:
Ack tags applied and merged for 3.20.0. Thanks!

PR #5736, opened 2023-01-03
drivers: stm32_bsec: use DT NVMEM layout API

GseoC:
Use of NVMEM layout in BSEC  driver instead of hardcoded OTP numbers.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -43,6 +43,7 @@ $(error Invalid platform flavor $(PLATFORM_FLAVOR))
 endif
 CFG_EMBED_DTB_SOURCE_FILE ?= $(flavor_dts_file-$(PLATFORM_FLAVOR))
 endif
+CFG_EMBED_DTB_SOURCE_FILE ?= stm32mp157c-dk2.dts
 
 ifneq ($(filter $(CFG_EMBED_DTB_SOURCE_FILE),$(flavorlist-no_cryp)),)
 $(call force,CFG_STM32_CRYP,n)
@@ -75,6 +76,7 @@ endif
 include core/arch/arm/cpu/cortex-a7.mk
 
 $(call force,CFG_DRIVERS_CLK,y)
+$(call force,CFG_DRIVERS_CLK_DT,y)
 $(call force,CFG_GIC,y)
 $(call force,CFG_INIT_CNTVOFF,y)
 $(call force,CFG_PSCI_ARM32,y)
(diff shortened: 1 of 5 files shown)


etienne-lms (changes requested):
This change makes stm32mp1 strictly depending on an embedded DTB. I would initally argue that it should be fixed but actually I think it's a valid constraint for that platform, to require an embedded DTB. In order to address that, we need a preliminatry fix that handles this requirement (few line in platform _conf.mk_).

GseoC:
Updated with removal of support for STM32MP1x platforms lacking an embedded device tree.

Applied @etienne-lms 's suggestion.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -42,6 +42,9 @@ ifeq ($(flavor_dts_file-$(PLATFORM_FLAVOR)),)
 $(error Invalid platform flavor $(PLATFORM_FLAVOR))
 endif
 CFG_EMBED_DTB_SOURCE_FILE ?= $(flavor_dts_file-$(PLATFORM_FLAVOR))
+else
+PLATFORM_FLAVOR ?= 157C_DK2
+CFG_EMBED_DTB_SOURCE_FILE ?= $(flavor_dts_file-$(PLATFORM_FLAVOR))
```
etienne-lms, line 47: Simpler:
```patch
+PLATFORM_FLAVOR ?= 157C_DK2
-ifneq ($(PLATFORM_FLAVOR),)
 ifeq ($(flavor_dts_file-$(PLATFORM_FLAVOR)),)
 $(error Invalid platform flavor $(PLATFORM_FLAVOR))
 endif
 CFG_EMBED_DTB_SOURCE_FILE ?= $(flavor_dts_file-$(PLATFORM_FLAVOR))
-endif
```

That said, build directive could config for, for example, a MP15-EV1 board wihtout setting `PLATFORM_FLAVOR`, for example `make CFG_EMBED_DTB_SOURCE_FILE=stm32mp157c-ev1.dts`
or for an MP13 variant: `make CFG_EMBED_DTB_SOURCE_FILE=stm32mp135f-dk.dts CFG_STM32MP13=y`

In those case, having `PLATFORM_FLAVOR=157C_DK2` would have not effect but would be confusing (it would appears in the generate config files, next to non 157c-dk2 configs). Therefore, I would prefer we set a default embedded DTB if none is provided, rather than forcing a `PLATFORM_CONFIG` value:
```patch
 ifneq ($(PLATFORM_FLAVOR),)
 ifeq ($(flavor_dts_file-$(PLATFORM_FLAVOR)),)
 $(error Invalid platform flavor $(PLATFORM_FLAVOR))
 endif
 CFG_EMBED_DTB_SOURCE_FILE ?= $(flavor_dts_file-$(PLATFORM_FLAVOR))
 endif
+CFG_EMBED_DTB_SOURCE_FILE ?= stm32mp157c-dk2.dts
```

What do you think?

etienne-lms (commented):
"drivers: stm32mp15_huk: fix use of stm32mp_is_closed_device()"
"ci: remove duplicate build for STM32MP157C-DK2"
and "drivers: stm32_bsec: use DT NVMEM layout API"

core/arch/arm/plat-stm32mp1/conf.mk
GseoC: I'm ok with your suggestion.

Only thing is we can have a flavor different from the embedded device tree. But that was already the case so... :)

I will rebusmit next week

GseoC:
Updated with tags applied.

Please note that:

- drivers: stm32mp15_huk: fix use of stm32mp_is_closed_device() was part of another P-R that has been merged. 
- I rephrased commit message for commit ci: remove duplicate build for STM32MP157C-DK2 to match the updates for commit plat-stm32mp1: conf: mandate the use of device tree on STM32MP1x platforms

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`  for commit
"plat-stm32mp1: conf: mandate the use of device tree on STM32MP1x platforms".

Could you simplifiy the commit message? I would shorten it to:

```
plat-stm32mp1: conf: mandate an embedded DTB

STM32MP1x platforms now mandate an embedded device tree using
CFG_EMBED_DTB_SOURCE_FILE. This decision simplifies platform
configuration and complies with existing flavors that all define an
embedded DT. This change makes stm32mp157c-dk2.dts the default
embedded DTB when none is set.

S-o-b: ...
```

PR #5737, opened 2023-01-03
plat-stm32mp1: enable debug feature on non secure-closed chip

GseoC:
This P-R enables full debug features if the board is not in secure-closed mode and CFG_TEE_CORE_DEBUG is enabled.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -310,6 +310,9 @@ $(error CFG_STM32MP15_HUK_BSEC_KEY and CFG_STM32MP15_HUK_BSEC_DERIVE_UID are exc
 endif
 endif # CFG_STM32MP15_HUK
 
+CFG_TEE_CORE_DEBUG ?= y
+CFG_STM32_DEBUG_ACCESS ?= $(CFG_TEE_CORE_DEBUG)
+
 # Sanity on choice config switches
 ifeq ($(call cfg-all-enabled,CFG_STM32MP15 CFG_STM32MP13),y)
 $(error CFG_STM32MP13_CLK and CFG_STM32MP15_CLK are exclusive)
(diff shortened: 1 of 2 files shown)


core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -310,6 +310,10 @@ $(error CFG_STM32MP15_HUK_BSEC_KEY and CFG_STM32MP15_HUK_BSEC_DERIVE_UID are exc
 endif
 endif # CFG_STM32MP15_HUK
 
+ifeq ($(CFG_TEE_CORE_DEBUG),y)
+CFG_STM32_DEBUG_ACCESS ?= y
+endif
```
etienne-lms, line 315: `CFG_TEE_CORE_DEBUG` may not be set at this point and is default enabled in _mk/config.mk`. So this change may not enable CFG_STM32_DEBUG_ACCESS while CFG_TEE_CORE_DEBUG is.

Suggestion:
```patch
-ifeq ($(CFG_TEE_CORE_DEBUG),y)
-CFG_STM32_DEBUG_ACCESS ?= y
-endif
+CFG_TEE_CORE_DEBUG ?= y
+CFG_STM32_DEBUG_ACCESS ?= $(CFG_TEE_CORE_DEBUG)
```

GseoC:
```
include $(platform-dir)/conf.mk
include mk/config.mk
```

Thought the other way around... ok with your suggestion

GseoC:
Force-pushed with suggestion applied, thanks

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`. 
Could you mention the new config switch in the commit message?

GseoC:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`. Could you mention the new config switch in the commit message?

Sure, updated with tag and commit message rephrased.

etienne-lms (changes requested):
Could you remove _"which is default set to 'y' for this platform."_ from the commit message? It is not true, MP15 variant default disables debug for RAM size considerations.

GseoC:
> Could you remove _"which is default set to 'y' for this platform."_ from the commit message? It is not true, MP15 variant default disables debug for RAM size considerations.

Done :)

etienne-lms (approved):

GseoC:
Everything up to date and approved

PR #5743, opened 2023-01-05
plat-stm32mp1: default restore BSEC SMC SiP on STM32MP15

etienne-lms:


GseoC (commented):
In commit message:
s/that uses/that use
s/The service will be default disables/The service will be default disabled

GseoC:
Indeed, it has been deprecated too fast :)

With minor comments addressed:

etienne-lms:
Thanks.
Commit message rephrased and review tag applied.

PR #5750, opened 2023-01-09
libutils: change assert() to conform with stdlib implementation

etienne-lms:
Changes assert() definition to return a (dummy) value when expression to true. This change allows to integrate external libraries which assume assert() conforms to such implementation, as found in GCC or LLVM toolchains.

### DIFF
--- a/lib/libutils/isoc/include/assert.h
+++ b/lib/libutils/isoc/include/assert.h
@@ -12,17 +12,19 @@ void __noreturn _assert_break(void);
 void _assert_log(const char *expr, const char *file, const int line,
 			const char *func);
 
-/* assert() specs: generates a log but does not panic if NDEBUG is defined */
+static inline void __noreturn _assert_trap(const char *expr_str,
+					   const char *file, const int line,
+					   const char *func)
+{
+	_assert_log(expr_str, file, line, func);
+	_assert_break();
+}
+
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

lib/libutils/isoc/include/assert.h
```diff
@@ -12,17 +12,20 @@ void __noreturn _assert_break(void);
...
+					   const char *func)
+{
+	_assert_log(expr_str, file, line, func);
+	_assert_break();
+}
+
 /* assert() specs: generates a log but does not panic if NDEBUG is defined */
 #ifdef NDEBUG
-#define assert(expr)	do { } while (0)
+#define assert(expr)	((void)0)
```
jforissier, line 25: The comment is wrong, no log is ever generated when `NDEBUG` is defined. Please just remove the comment.

jforissier (commented):
On minor comment, otherwise:

lib/libutils/isoc/include/assert.h
etienne-lms: Ok, agree.

etienne-lms:
Comment addressed and tags applied.

jforissier:

jforissier (approved):

PR #5757, opened 2023-01-11
libutee: _utee_panic: add register info

liushiwei007:
Added the ability to save register information
on the stack in function _utee_panic.

Signed-off-by: liushiwei <liushiwei@eswincomputing.com>

### DIFF
--- a/lib/libutee/arch/riscv/utee_syscalls_rv.S
+++ b/lib/libutee/arch/riscv/utee_syscalls_rv.S
@@ -5,6 +5,8 @@
 
 #include <tee_syscall_numbers.h>
 #include <asm.S>
+#include <riscv.h>
+#include <riscv_macros.S>
 
         .section .text
 
@@ -22,6 +24,10 @@
 	.endm
 
 	FUNC _utee_panic, :
+	addi	sp, sp , -(30*RISCV_XLEN_BYTES)
+	STR	ra, 0(sp)
+	store_xregs sp, RISCV_XLEN_BYTES, 3, 31
+	mv a1, sp
 	j	__utee_panic
 	END_FUNC _utee_panic
 
--- a/lib/libutils/ext/arch/riscv/sub.mk
+++ b/lib/libutils/ext/arch/riscv/sub.mk
@@ -1,0 +1,1 @@
+global-incdirs-y += include


liushiwei007:
@maroueneboubakri  Please see if this meets your requirements for panic function.

lib/libutee/arch/riscv/sub.mk
```diff
@@ -1,3 +1,4 @@
 cppflags-y += -I$(sub-dir)/../..
+global-incdirs-y += ../../../../core/arch/riscv/include
```
jenswi-linaro, line 2: We don't reach into other parts of the source tree like this. If there's something in `core/arch/riscv/include` that's needed in libutils then it should probably be kept in a libutils include dir instead.

liushiwei007: > We don't reach into other parts of the source tree like this. If there's something in `core/arch/riscv/include` that's needed in libutils then it should probably be kept in a libutils include dir instead.

I refer to the macro store_xregs, It is kept in the core/arch/riscv/include/riscv_macros. S, compile ldelf can access to, but when compile user mode library can't access.  Can I store a copy of riscv_macros. S in the library header file?

jenswi-linaro: No copies, please. If riscv.h and riscv_macros.S are needed outside /core then I believe `lib/libutils/ext/arch/riscv/include` should be a good place to keep them.

liushiwei007: I don't see this directory yet, do you mean to create this directory, change some compilation options, and make it accessible to outside and core?

jenswi-linaro: Yes, please.

liushiwei007: Ok, I try it.

liushiwei007:
@maroueneboubakri @jenswi-linaro 
riscv_macros.S and riscv.h are updated in #5714 ，and I also cited the changes in this update，Whether wait  #5714 merged, then move the two files to the lib/libutils/ext/arch/riscv/include.

maroueneboubakri:
> @maroueneboubakri @jenswi-linaro riscv_macros.S and riscv.h are updated in #5714 ，and I also cited the changes in this update，Whether wait #5714 merged, then move the two files to the lib/libutils/ext/arch/riscv/include.

Not sure if you read my comments above but this is not needed at all.

liushiwei007:
What is not needed?  this PR  or this move of the header file?

> > @maroueneboubakri @jenswi-linaro riscv_macros.S and riscv.h are updated in #5714 ，and I also cited the changes in this update，Whether wait #5714 merged, then move the two files to the lib/libutils/ext/arch/riscv/include.
> 
> Not sure if you read my comments above but this is not needed at all.

What is not needed,  this PR  or this move of the header file?

maroueneboubakri:
> What is not needed? this PR or this move of the header file?
> 
> > > @maroueneboubakri @jenswi-linaro riscv_macros.S and riscv.h are updated in #5714 ，and I also cited the changes in this update，Whether wait #5714 merged, then move the two files to the lib/libutils/ext/arch/riscv/include.
> > 
> > 
> > Not sure if you read my comments above but this is not needed at all.
> 
> What is not needed, this PR or this move of the header file?

There is no need to move these header since you don't need to store the panic context because it is already saved.

Issue #5759, opened 2023-01-11
Core panic with MEMTAG=y


jforissier:
On [QEMUv8](https://optee.readthedocs.io/en/latest/building/devices/qemu.html#qemu-v8) with `make MEMTAG=y run`, OP-TEE panics on boot:
```
E/TC:0 0 Core data-abort at address 0x6000000456c4ea0 (tag check fault)
E/TC:0 0  esr 0x96000051  ttbr0 0x0e1a9000   ttbr1 0x00000000   cidr 0x0
E/TC:0 0  cpu #0          cpsr 0x800003c4
E/TC:0 0  x0  06000000456c4ea0 x1  0000000000000001
E/TC:0 0  x2  0000000000000000 x3  0000000000000010
E/TC:0 0  x4  09000000456c4e40 x5  0000000000000010
E/TC:0 0  x6  ffffffffffffffff x7  09000000456c4e60
E/TC:0 0  x8  0000000000000020 x9  00000000456b4e90
E/TC:0 0  x10 000000004568d118 x11 0000000000000258
E/TC:0 0  x12 00000000456d906c x13 0000000000000258
E/TC:0 0  x14 00000000456d9164 x15 000000004568d118
E/TC:0 0  x16 0000000045684bd4 x17 0000000000000000
E/TC:0 0  x18 00000000456d9134 x19 0000000000000258
E/TC:0 0  x20 000000004568d118 x21 0000000000000000
E/TC:0 0  x22 0000000000000000 x23 000000004568d3d0
E/TC:0 0  x24 0000000000000000 x25 00000000456a5000
E/TC:0 0  x26 00000000456a5588 x27 00000000456a5608
E/TC:0 0  x28 00000000456a39c0 x29 00000000456d90a0
E/TC:0 0  x30 000000004563ea0c elr 000000004563ea24
E/TC:0 0  sp_el0 00000000456d90a0
E/TC:0 0 TEE load address @ 0x4562b000
E/TC:0 0 Call stack:
E/TC:0 0  0x4563ea24 add_node_to_probe at optee_os/core/kernel/dt_driver.c:559
E/TC:0 0  0x4563eb7c parse_node at optee_os/core/kernel/dt_driver.c:650
E/TC:0 0  0x45642398 call_initcalls at optee_os/core/kernel/initcall.c:41
E/TC:0 0  0x45633798 boot_init_primary_late at optee_os/core/arch/arm/kernel/boot.c:1395
[...]

PR #5760, opened 2023-01-12
libutils: malloc: fix MDBG combined with MEMTAG

jenswi-linaro:
Fixes problems with MDBG when combined with memory tagging. MDBG uses an extra header which wasn't taken into account when tagging the allocated memory. This leads to tag check fault. So fix it by adding the size of the extra header when tagging the allocated memory.

Fixes: 08a5c4f9ae42 ("libutils: add MTE support in malloc() and friends")

### DIFF
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -209,7 +209,7 @@ static __maybe_unused bufsize bget_buf_size(void *buf)
 	return osize;
 }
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Fixes https://github.com/OP-TEE/optee_os/issues/5759

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with minor comment addressed.

jenswi-linaro:

PR #5761, opened 2023-01-12
Interrupt notification over async notif

etienne-lms:
This P-R proposes 2 changes related to async notification.

The first one is the ability to use a per-cpu interrupt (PPI) as notification vehicle in normal world instead of an SPI. The related change in Linux kernel is available in the LKML:
https://lore.kernel.org/lkml/20230112145424.3791276-2-etienne.carriere@linaro.org/

The second one is an extension of async notif to allow OP-TEE to generate an interrupt event for a Linux kenrel interrupt consumer. Async notif was initially designed to schedule or wake an OP-TEE thread. This new "interrupt notification" feature is designed for when the consumer expects a interrupt event. The related Linux kernel change is available in the LKML:
DT bindings: https://lore.kernel.org/lkml/20230112145424.3791276-3-etienne.carriere@linaro.org/
optee driver: https://lore.kernel.org/lkml/20230112145424.3791276-4-etienne.carriere@linaro.org/


core/kernel/notif.c
```diff
@@ -35,6 +51,8 @@ TEE_Result notif_alloc_async_value(uint32_t *val)
 
 	if (!alloc_values_inited) {
 		bit_set(notif_alloc_values, NOTIF_VALUE_DO_BOTTOM_HALF);
+		if (IS_ENABLED(CFG_CORE_IT_NOTIF))
+			bit_set(notif_alloc_values, NOTIF_VALUE_DO_IT);
```
jenswi-linaro, line 55: This should probably be done unconditionally to make sure it's not found by `bit_ffc()` below.

core/kernel/notif.c
```diff
@@ -106,6 +124,73 @@ void notif_send_async(uint32_t value)
...
+	*it_valid = (bit >= 0);
+	if (!*it_valid) {
+		*it_pending = false;
+		goto out;
+	}
+
+	res = bit;
+	bit_clear(notif_it_pending, res);
+	bit_ffs(notif_it_pending, (int)NOTIF_IT_VALUE_MAX + 1, &bit);
+	*it_pending = true;
```
jenswi-linaro, line 147: `s/true/(bit >= 0)/`

etienne-lms:
Updated above description with LKML links.

core/arch/arm/tee/entry_fast.c
```diff
@@ -110,6 +117,14 @@ static void tee_entry_exchange_capabilities(struct thread_smc_args *args)
 
 	args->a1 |= OPTEE_SMC_SEC_CAP_RPC_ARG;
 	args->a3 = THREAD_RPC_MAX_NUM_PARAMS;
+
+#ifdef CFG_CORE_IT_NOTIF
```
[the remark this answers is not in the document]
etienne-lms: `CFG_CORE_IT_NOTIF_MAX` maybe undefined when `CFG_CORE_IT_NOTIF` is disabled.
Unless you prefer we force `CFG_CORE_IT_NOTIF_MAX` to a dummy value when so?

core/arch/arm/tee/entry_fast.c
```diff
@@ -285,6 +323,20 @@ void __tee_entry_fast(struct thread_smc_args *args)
 			args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
 		break;
 
+	case OPTEE_SMC_GET_IT_NOTIF_VALUE:
+		if (IS_ENABLED(CFG_CORE_ASYNC_NOTIF))
+			get_it_value(args);
```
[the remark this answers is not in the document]
etienne-lms: ok

core/kernel/notif.c
etienne-lms: Legacy implementation allows async notif value 1 be used for standard (see linux kernel impl.).

core/kernel/notif.c
etienne-lms: At this stage, we know `bit >= 0` (lines 138..141).

core/arch/arm/tee/entry_fast.c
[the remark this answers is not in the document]
jenswi-linaro: OK, let's keep it as it is then.

core/kernel/notif.c
jenswi-linaro: I'm sorry, I don't follow. Can you give a pointer?

core/kernel/notif.c
jenswi-linaro: No, you have just cleared that bit. Now you're checking if there's any other bit to instruct the caller to call again to get the next bit.

core/kernel/notif.c
etienne-lms: Latest OP-TEE and Linux kernel consider notif value 1 as a generic purpose aysnc notif. So we cannot change Linux to unconditionally consider this value as reserved for "it notif". Therefore, on Linux side, this value is reserved for "it notif" only upon IT_NOTIF capability being enabled. In related Linux kernel patch (https://lore.kernel.org/lkml/20230112145424.3791276-4-etienne.carriere@linaro.org/):
```c
+		else if (optee->smc.sec_caps & OPTEE_SMC_SEC_CAP_IT_NOTIF &&
+			 value == OPTEE_SMC_ASYNC_NOTIF_VALUE_DO_IT)
+			handle_optee_it(optee);
```

I agree we could have the OP-TEE side to always reserve this value 1 for IT_NOTIF, but we cannot do that on Linux side. So for consistency, I bound it to the capabiliy on both side. Do you think it adds confusions?

core/kernel/notif.c
etienne-lms: you're right! indeed. i'll fix, thanks.

core/kernel/notif.c
jenswi-linaro: Thanks for the explanation. Yes, I think it's slightly confusing on the OP-TEE side, with a value that is only sometimes reserved. On the Linux side, we have no choice, but on the OP-TEE side, I believe it would be clearer to have it permanently reserved.

etienne-lms:
Created https://github.com/OP-TEE/optee_os/pull/5772 for the PPI change. I'll remove that commit from this series and will focus on `CFG_CORE_IT_NOTIF` in this P-R.

etienne-lms:
I re-implemented the feature (PPI part removed) rebased on master (where PPI patch was merged) and split into several commits. The related Linux kernel changes to support interrupt notification must be updated. I'll send a PATCH v2 post and update this comment with a LKML ref (**TODO**).

core/kernel/notif.c
etienne-lms: I forgot to fix this. I'll add a fixup commit.
My apologies.

core/kernel/notif.c
etienne-lms: With latest proposal, there is no more reserved async value related to interrupt notification.

This is at the cost of non-secure world will always invoke (smc fastcall) optee to query pending IRQ after what pending async value (thread related) will will queried.

etienne-lms:
Digging deeper, here are improvements to do on the API. I want to minimize the number of world transition during the non-secure interrupt execution. I'll update. I also should come with a test setup we can exercice (e.g. CI with Qemu) implying a test pathc in our OP-TEE Linux fork.

etienne-lms:
Closing. Superseded by https://github.com/OP-TEE/optee_os/pull/5793.

PR #5762, opened 2023-01-12
Log capabilities on boot and check shared memory support

jforissier:
It took me a bit more time than it should have to troubleshoot an error on RockPi4 with `CFG_CORE_RESERVED_SHM=n`. The Linux kernel reports:
```
optee: probe of firmware:optee failed with error -22
```
(22 is `EINVAL`)

What happens with the RockPi4 build is, it doesn't register any memory range with `register_ddr()` and no DT is passed to OP-TEE by the bootloader either. Therefore, dynamic shm is disabled. So if reserved shm is disabled at compile time, there is no shared memory communication possible.

The two patches in this PR:
- Upgrade the existing `DMSG()` in `tee_entry_exchange_capabilities()` to `IMSG()`,
- Add similar `IMSG()` for virtualization and reserved shm,
- Panic the core if no shm is enabled.

### DIFF
--- a/core/arch/arm/tee/entry_fast.c
+++ b/core/arch/arm/tee/entry_fast.c
@@ -64,6 +64,7 @@ static void tee_entry_fastcall_l2cc_mutex(struct thread_smc_args *args)
 
 static void tee_entry_exchange_capabilities(struct thread_smc_args *args)
 {
+	bool res_shm_en = IS_ENABLED(CFG_CORE_RESERVED_SHM);
 	bool dyn_shm_en __maybe_unused = false;
 
 	/*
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
For "core: log capabilities with INFO level":

I'm not so sure about the panic part. It seems the main purpose is to make it easier to spot a log entry. I can imagine use cases where it doesn't matter if we're able to communicate with for instance a boot loader.

jforissier:
> I'm not so sure about the panic part. It seems the main purpose is to make it easier to spot a log entry.

Indeed.

> I can imagine use cases where it doesn't matter if we're able to communicate with for instance a boot loader.

I thought about special cases too, but wasn't sure if anything useful could be done without shared memory. Are you thinking about a boot loader that would only pass parameters by value (in and out)? And then no use of the TEE by the REE? It would be a pretty contrived scenario I believe. That said I'm OK with dropping the second patch to be safe.

jenswi-linaro:
> > I'm not so sure about the panic part. It seems the main purpose is to make it easier to spot a log entry.
> 
> Indeed.
> 
> > I can imagine use cases where it doesn't matter if we're able to communicate with for instance a boot loader.
> 
> I thought about special cases too, but wasn't sure if anything useful could be done without shared memory. Are you thinking about a boot loader that would only pass parameters by value (in and out)? And then no use of the TEE by the REE? It would be a pretty contrived scenario I believe. That said I'm OK with dropping the second patch to be safe.

Please do. It should IMO be a normal world decision to stop the boot on anything but critical errors in secure world.

jforissier:
Panic patch dropped.

core/arch/arm/tee/entry_fast.c
```diff
@@ -64,6 +64,7 @@ static void tee_entry_fastcall_l2cc_mutex(struct thread_smc_args *args)
 
 static void tee_entry_exchange_capabilities(struct thread_smc_args *args)
 {
+	bool res_shm_en = IS_ENABLED(CFG_CORE_RESERVED_SHM);
```
etienne-lms, line 67: could be `const`

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` (with minor comment addressed or not)

core/arch/arm/tee/entry_fast.c
jforissier: True but it would not make any practical difference (it's on the stack not global; I would definitely add `const` for a global variable).

PR #5771, opened 2023-01-17
crypto: se050: update policies

ldts:
Certain devices (depending on the applet) will use the common policies to for the object's read/write permissions.

Make sure both are supported.


jenswi-linaro (commented):

ldts:
updated tag and fixed wording error on the commit. 

this can be merged now - it will make sure devices with newer applets are supported (otherwise access to their keys will be forbidden by the policy)

PR #5772, opened 2023-01-17
core: notif: allow GIC_PPI usage for async notif

etienne-lms:
Allows to use GIC_PPI interrupts for asynchronous notification.

This change replace macro COMPILE_TIME_ASSERT() with static_assert() as the former is deprecated.

### DIFF
--- a/core/kernel/notif.c
+++ b/core/kernel/notif.c
@@ -94,7 +94,7 @@ void notif_send_async(uint32_t value)
 {
 	uint32_t old_itr_status = 0;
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

etienne-lms:
Tag applied. Thanks.
It would be nice that this change is part of 3.20.0. It would ease reviews and discussions on Linux optee driver changes for GIC PPI support has we can point to an OP-TEE tag that embeds the feature. Note this change makes no functional changes.

etienne-lms:
Thanks @jenswi-linaro & @jforissier.

PR #5774, opened 2023-01-18
plat-sam: retrieve matrix id from device-tree

clementleger:
When device have multiple instances, it is needed to identify which one we are going to
use in the driver. For that purpose, add `matrix_dt_get_id()`  function which matches the device-tree
reg property with the correct matrix id.


clementleger:
This is not a Linux property since Linux has no use of this (no matrix driver). Moreover, adding this property would need to modify all the devices bindings. This is the only solution I found to make something flexible rather than comparing physical addresses to determine the matrix id.

jforissier:
> This is not a Linux property since Linux has no use of this (no matrix driver). 

OK, but even though the property doesn't apply to the kernel I still believe it should be proposed upstream because Linux is the reference repository for DT bindings. We can't expect people working on a different secure OS and using DT to come here and find this property, all the more since it is not formally documented but just used by some code. And I'd rather avoid having OP-TEE specific bindings.
Perhaps people who no more about the DT process in Linux can comment.

> Moreover, adding this property would need to modify all the devices bindings.

Could it be made optional perhaps?

>  This is the only solution I found to make something flexible rather than comparing physical addresses to determine the matrix id.

I'm fine with having the info in DT, it clearly is hardware description.

clementleger:
> > This is not a Linux property since Linux has no use of this (no matrix driver).
> 
> OK, but even though the property doesn't apply to the kernel I still believe it should be proposed upstream because Linux is the reference repository for DT bindings. We can't expect people working on a different secure OS and using DT to come here and find this property, all the more since it is not formally documented but just used by some code. And I'd rather avoid having OP-TEE specific bindings. Perhaps people who no more about the DT process in Linux can comment.

Agreed.

> 
> > Moreover, adding this property would need to modify all the devices bindings.
> 
> Could it be made optional perhaps?

What I meant is that it's not really a per-device thing and when thinking more about it, with have another option which might be to create a new DT node for the matrix device and associate the device in that node because these ids are actually not tied to the devices but rather to the matrix controller itself. This would allow to keep the device bindings as is. We could have subnodes defining the matrix ids for instance:

```
matrix {
   compatible = "microchip,matrix";
   tcb0 {
      id = <&tcb0 35>
   },
   tcb1 {
      id = <&tcb1 36>
   },
  ...
}
```

clementleger:
New version:
- Discard "atmel,matrix-id" from device-tree and use reg content to match the node with the id

clementleger:
Ping ?

jforissier:

PR #5775, opened 2023-01-18
plat-sam: use device tree "assigned-clock-rates" property rather than hardcoding rate

clementleger:


core/include/dt-bindings/clock/at91.h
```diff
@@ -27,6 +27,7 @@
 #define PMC_PLLBCK		8
 #define PMC_AUDIOPLLCK		9
 #define PMC_MCK_PRES		10
+#define PMC_CORE_CLK_COUNT	11
```
etienne-lms: What about SAMA7G5 variant with specific belo clocks ?

etienne-lms (commented):
Commit "drivers: clk: sam: use define for count of main clocks",
in commit message: s/Use define/Adds a macro/.

With this fix, `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for the series.

I've made some other comments but you can discard them.

core/include/dt-bindings/clock/at91.h
clementleger: You are right, this should actually be named `PMC_SAMA5D2_CORE_CLK_COUNT`since this is specific to sama5d2 platform.

clementleger:
New version:
- Rename `PMC_CORE_CLK_COUNT` into `PMC_SAMA5D2_CORE_CLK_COUNT`
- Add Etienne `Acked-by`

PR #5782, opened 2023-01-23
core: crypto: add support MD5 hashes in RSA sign/verify

SOuajih:
Introduce support of using MD5 hashes in RSA sign/verify operations, which is required by AOSP Keymaster.

This patch has been inspired by the following commit: 199d0b7310d1705661a106358f1f0b46e4c5c587

Signed-off-by: Julien Masson <jmasson@baylibre.com>
Signed-off-by: Safae Ouajih <souajih@baylibre.com>


lib/libutee/include/tee_api_defines.h
```diff
@@ -232,12 +232,14 @@
...
 #define TEE_ALG_RSASSA_PKCS1_V1_5_SHA512        0x70006830
 #define TEE_ALG_RSASSA_PKCS1_V1_5_MD5SHA1       0x7000F830
+#define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_MD5       0x70111930
 #define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA1      0x70212930
 #define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA224    0x70313930
 #define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA256    0x70414930
 #define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA384    0x70515930
 #define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA512    0x70616930
 #define TEE_ALG_RSAES_PKCS1_V1_5                0x60000130
+#define TEE_ALG_RSAES_PKCS1_OAEP_MGF1_MD5       0x60110230
```
jforissier, line 242: Both defines should go in `tee_api_defines_extensions.h` because they are non-standard (not in the GlobalPlatform spec).

jforissier (commented):
Hello @SOuajih, thanks for the patch.

GlobalPlatform has deprecated the use of MD5 [1], so I am wondering if adding this kind of support here is a good thing to do. Although I do understand there may be valid use cases (for compatibility with older products).

One more comment below.

[1] "Cryptographic Algorithm Recommendations – Public Release v2.0" ([link](https://globalplatform.org/specs-library/globalplatform-technology-cryptographic-algorithm-recommendations/))

lib/libutee/include/tee_api_defines.h
SOuajih: Applied, Thank you for the review.

lib/libutee/include/tee_api_defines_extensions.h
```diff
@@ -7,6 +7,10 @@
 #ifndef TEE_API_DEFINES_EXTENSIONS_H
 #define TEE_API_DEFINES_EXTENSIONS_H
 
+/* Algorithm Identifiers */
```
jforissier, line 10: Please replace this comment with something more meaningful such as `/* RSA signatures with MD5 hash */`

jforissier (commented):
See one comment in the code. About the commit description:

> ```
> Introduce support of using MD5 hashes in RSA sign/verify operations,
> which is required by AOSP Keymaster.
>
> This patch has been inspired by the following commit:
> 
> 199d0b7310d1705661a106358f1f0b46e4c5c587
> ```

The proper format is:

```
by commit 199d0b7310d1 ("core: crypto: add support MD5 hashes in RSA
sign/verify").
```

Also the link to the AOSP test would be nice to have. Context always helps.

With that:

lib/libutee/include/tee_api_defines_extensions.h
```diff
@@ -7,6 +7,10 @@
 #ifndef TEE_API_DEFINES_EXTENSIONS_H
 #define TEE_API_DEFINES_EXTENSIONS_H
 
+/* RSA signatures with MD5 hash */
+#define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_MD5       0x70111930
+#define TEE_ALG_RSAES_PKCS1_OAEP_MGF1_MD5       0x60110230
```
etienne-lms, line 12: This commit is about RSA sign/verify but `TEE_ALG_RSAES_PKCS1_OAEP_MGF1_MD5` is about RSa ciphering. I think related added lines should be removed from this change.

etienne-lms (commented):
For commit message, prefer use a `Link:` tag and not wrap the referred commit header line:
```
This patch has been inspired by commit [1].

Link: [1] https://github.com/OP-TEE/optee_os/commit/199d0b7310d1705661a106358f1f0b46e4c5c587 ("core: crypto: add support MD5 hashes in RSA
sign/verify").

File: lib/libutee/include/tee_api_defines_extensions.h (hunk 3 above)
Comment (jforissier) [reply lib/libutee/include/tee_api_defines_extensions.h:12]: I agree, either the commit could be split in two, or the commit description needs to be fixed to mention both signing and cipher (it's a small change so it is acceptable to have both in one commit IMO).

Comment (SOuajih) [reply lib/libutee/include/tee_api_defines_extensions.h:12]: Applied: 
- Mention both Sign and Cipher in commit description
- Add a link to the AOSP vts test
- Use "Link:" tag to refer for the commit in the description
@jforissier Please let me know if the Acked-by is still ok to be added

Thank you

Comment (jforissier) [reply lib/libutee/include/tee_api_defines_extensions.h:12]: Absolutely. You don't need the blank line between the `Link:` and `Signed-off-by:` lines though.

Comment (SOuajih) [reply lib/libutee/include/tee_api_defines_extensions.h:12]: Applied, Thank you

**Review by etienne-lms — COMMENTED** [review]
Please unwrap `Link: [2]` (and remove terminal dot) in commit message:
```patch
-Link: [2] 199d0b7310d1 ("core: crypto: add support MD5 hashes in RSA
-sign/verify").
+Link: [2] 199d0b7310d1 ("core: crypto: add support MD5 hashes in RSA sign/verify")

File: lib/libutee/include/tee_api_defines_extensions.h [hunk 4]
```diff
@@ -7,6 +7,10 @@
 #ifndef TEE_API_DEFINES_EXTENSIONS_H
 #define TEE_API_DEFINES_EXTENSIONS_H
 
+/* RSA signatures with MD5 hash */
+#define TEE_ALG_RSASSA_PKCS1_PSS_MGF1_MD5       0x70111930
```
jenswi-linaro, line 11: Implementation-defined algorithm identifiers should be in the range 0xF0000000 – 0xF0FFFFFF.

SOuajih: I will make an update as soon as possible
Thank you

SOuajih: Hello @jenswi-linaro . I am having some issues with this range of algorithm identifiers. Should I adapt all necessary optee-os code and add very specific algorithm code to make it work ? One small example of this is :  reworking all syscall_cryp_state_alloc() since TEE_ALG_GET_CLASS(algo) will be returning 0xF for that range and that is not defined to be an operation (I could define it but it will be a definition for all the 0xF0000000 – 0xF0FFFFFF identifier range ). Could you please tell what is the right way to do this implementation ?

jforissier: > One small example of this is : reworking all syscall_cryp_state_alloc() since TEE_ALG_GET_CLASS(algo) will be returning 0xF for that range and that is not defined to be an operation (I could define it but it will be a definition for all the 0xF0000000 – 0xF0FFFFFF identifier range )

`TEE_ALG_GET_CLASS()` is implemented as a static inline function: `__tee_alg_get_class()`. You can easily return whatever value is needed, and only for that particular algorithm identifier. Same for `TEE_ALG_GET_MAIN_ALG()` and possibly other things if needed.

See https://github.com/OP-TEE/optee_os/blob/3.20.0/lib/libutee/include/utee_defines.h#L57-L71.

SOuajih: Thank you, it is clear to me now. I will make an update

lib/libutee/include/tee_api_defines_extensions.h
```diff
@@ -7,6 +7,10 @@
 #ifndef TEE_API_DEFINES_EXTENSIONS_H
 #define TEE_API_DEFINES_EXTENSIONS_H
 
+/* RSA signatures with MD5 hash */
```
etienne-lms, line 10: I thin this should explain why these magic numbers:
`/* Value prefixed with vendor ID bit31 with by TEE bitfields IDs */`

SOuajih: Applied, Thank you

jenswi-linaro:

PR #5784, opened 2023-01-23
ta: avb: fix object leakage at lock state write

ikhorn:
Should close object created for reading, so better use seek and then close the object.

Fixes: b29b41950 ("ta: add AVB TA")
Signed-off-by: Ivan Khoronzhuk <ivan.khoronzhuk@globallogic.com>


jenswi-linaro (commented):

PR #5785, opened 2023-01-23
mm: core_memprot: align convenience macroses usage

ikhorn:
As the macroses exist, why not to use them.

Signed-off-by: Ivan Khoronzhuk <ivan.khoronzhuk@globallogic.com>

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -189,7 +189,7 @@ static uint32_t std_entry_with_parg(paddr_t parg, bool with_rpc_arg)
 	uint32_t rv = 0;
 
 	/* Check if this region is in static shared space */
-	if (core_pbuf_is(CORE_MEM_NSEC_SHM, parg, sz)) {
+	if (tee_pbuf_is_nsec_shm(parg, sz)) {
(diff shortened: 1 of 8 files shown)


etienne-lms (commented):
Should rephrase the commit log to state this change adds helper functions, extending existing tee_pbuf_is_*() and tee_vbuf_is_*().

Note also `tee_pbuf_is_nsec_shm()` can be misleading: it targets buffers in the static reserved shm area (`CFG_CORE_RESERVED_SHM=y`), not any shared memory buffer (e.g. dyn shm).

jenswi-linaro:
You're introducing most of the macros you claim already existed. I'm not sure this is that useful, I'd rather address the bigger picture than add another set of macros.

ikhorn:
> It not just adds macros, it tries to decrease misleading by using same style verification macros. The misleading between static reserved shm and dynamic shared memory added historically before and should be got used to already I guess, but deferentially not by this change.

ikhorn:
> I'm not sure this is that useful, I'd rather address the bigger picture than add another set of macros.
Absolutely agree. But then better drop the existent macros and use same style everywhere.

jenswi-linaro:
Sure, go ahead and remove the `tee_?buf_*()` macros. I've always found the cast in them troublesome.

PR #5787, opened 2023-01-23
Asyn notif: support PPI: fix DT node

etienne-lms:
P-R https://github.com/OP-TEE/optee_os/pull/5772 allows OP-TEE to use a PPI instead of an SPI for async notification of normal world, but I forgot to update `add_optee_dt_node()` that can populate the DT optee node with interrupt config. This P-R addresses this issue.


core/arch/arm/kernel/boot.c
```diff
@@ -798,22 +799,43 @@ static int add_optee_dt_node(struct dt_descriptor *dt)
...
-		 * first cell.
+		 * first cell. A PPI type is indicated with value 1.
 		 *
 		 * The interrupt number goes in the second cell where
-		 * SPIs ranges from 0 to 987.
+		 * SPIs ranges from 0 to 987 and PPI ranges from 0 to 15.
 		 *
 		 * Flags are passed in the third cell where a 1 means edge
-		 * triggered.
+		 * triggered for SPI types and (8|mask<<8) means a lowlevel
```
jenswi-linaro, line 808: s/lowlevel/level/
If you want to go into more detail please do so, but then we should do the same for the edge triggered.

core/arch/arm/kernel/boot.c
```diff
@@ -798,22 +799,43 @@ static int add_optee_dt_node(struct dt_descriptor *dt)
...
+			val[1] = CFG_CORE_ASYNC_NOTIF_GIC_INTID - GIC_SPI_BASE;
+			/* Type level */
+			val[2] = IRQ_TYPE_EDGE_RISING;
+		} else {
+			/* GIC_PPI as per DT bindings */
+			val[0] = GIC_PPI;
+			/* PPI interrupt number as per DT bindings */
+			val[1] = CFG_CORE_ASYNC_NOTIF_GIC_INTID - GIC_PPI_BASE;
+			/* Type/levem */
+			val[2] = IRQ_TYPE_LEVEL_LOW |
```
jenswi-linaro, line 832: Does this have to be level triggered? Edge triggered interrupts are easier to handle since they are fire and forget.
If you're keeping it as level triggered please add a comment explaining why the PPIs must be treated different from SPIs.

core/arch/arm/kernel/boot.c
```diff
@@ -798,22 +799,43 @@ static int add_optee_dt_node(struct dt_descriptor *dt)
...
+			/* SPI interrupt number as per DT bindings */
+			val[1] = CFG_CORE_ASYNC_NOTIF_GIC_INTID - GIC_SPI_BASE;
+			/* Type level */
+			val[2] = IRQ_TYPE_EDGE_RISING;
+		} else {
+			/* GIC_PPI as per DT bindings */
+			val[0] = GIC_PPI;
+			/* PPI interrupt number as per DT bindings */
+			val[1] = CFG_CORE_ASYNC_NOTIF_GIC_INTID - GIC_PPI_BASE;
+			/* Type/levem */
```
jenswi-linaro, line 831: levem?

core/arch/arm/kernel/boot.c
```diff
@@ -798,22 +799,43 @@ static int add_optee_dt_node(struct dt_descriptor *dt)
...
+
+		/* PPI are visible only in current CPU cluster */
+		static_assert(!CFG_CORE_ASYNC_NOTIF_GIC_INTID ||
+			      CFG_CORE_ASYNC_NOTIF_GIC_INTID >= GIC_SPI_BASE ||
+			      (CFG_CORE_ASYNC_NOTIF_GIC_INTID >= GIC_PPI_BASE &&
+			       CFG_TEE_CORE_NB_CORE <= 4));
+
+		if (CFG_CORE_ASYNC_NOTIF_GIC_INTID >= GIC_SPI_BASE) {
+			/* GIC_SPI as per DT bindings */
+			val[0] = GIC_SPI;
```
jenswi-linaro, line 821: It would be more clear with separate helper variables for this, for instance, `itr_type`, `itr_id`, and `itr_trigger`.

core/arch/arm/kernel/boot.c
etienne-lms: ok, i'll more less verbose. The info are in the DT bindings.

core/arch/arm/kernel/boot.c
etienne-lms: ok

core/arch/arm/kernel/boot.c
etienne-lms: sorry, i'll fix.

core/arch/arm/kernel/boot.c
etienne-lms: I'm a bit unshure here. I'll check.

jenswi-linaro (commented):

core/arch/arm/kernel/boot.c
```diff
@@ -798,22 +799,43 @@ static int add_optee_dt_node(struct dt_descriptor *dt)
...
-					      GIC_SPI_BASE),
-			TEE_U32_TO_BIG_ENDIAN(irq_type_edge),
-		};
+		uint32_t val[3] = { };
+
+		/* PPI are visible only in current CPU cluster */
+		static_assert(!CFG_CORE_ASYNC_NOTIF_GIC_INTID ||
+			      CFG_CORE_ASYNC_NOTIF_GIC_INTID >= GIC_SPI_BASE ||
+			      (CFG_CORE_ASYNC_NOTIF_GIC_INTID >= GIC_PPI_BASE &&
+			       CFG_TEE_CORE_NB_CORE <= 4));
```
jforissier, line 819: Does this mean PPIs cannot be used for async notif on CPUs with more than 4 cores?

etienne-lms: I saw cortex-A can connect to up to 8 cores on a single GIC for PPI routing, but I guess it depends on the silicon implementation. So here, i think `CFG_TEE_CORE_NB_CORE <= 4` could be removed. It is a platform choise to decide whether an SPI or PPI can be used for optee async notifs.

etienne-lms:
I increased `CFG_TEE_CORE_NB_CORE` max value from 4 to 8 as PPI cpu masking supports up to 8 cores. @jenswi-linaro, does your A-b tag still apply?

etienne-lms:
Appended another fixup commit to address CI checkpatch warnings.

jenswi-linaro:
> I increased `CFG_TEE_CORE_NB_CORE` max value from 4 to 8 as PPI cpu masking supports up to 8 cores. @jenswi-linaro, does your A-b tag still apply?

Yes

etienne-lms:
Thanks.
Fixup commits squashed and tag applied.

PR #5788, opened 2023-01-23
crypto: drivers: se050: fix generation of oid values

ldts:
Converting the OID watermarked value to a bignum removes the first byte if this is different than zero.

The failing case observed the value 0x57.72.15.66.1a.f2.9d.00 being retrieved as 0x57.72.15.66.1a.f2.9d after having been transformed into a bignum and back to its original binary value.

This will cause cryptographic operations to fail as the secured keys and objects become not addressable.

### DIFF
--- a/core/drivers/crypto/se050/adaptors/utils/utils.c
+++ b/core/drivers/crypto/se050/adaptors/utils/utils.c
@@ -77,7 +77,12 @@ static sss_status_t generate_oid(uint32_t *val)
 			return kStatus_SSS_Fail;
 
 		oid = OID_MIN + (random & OID_MAX);
-		if (oid > OID_MAX)
+
+		/*
+		 * The less significant byte different than zero prevents bignum
+		 * conversions from discarding it
+		 */
+		if (oid > OID_MAX || !(oid & 0xFF))
 			continue;
 
 		if (!se050_key_exists(oid, &se050_session->s_ctx)) {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
           export CFG_WERROR=y
 
           function _make() { make -j$(nproc) -s O=out $*; }
-          function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.4.0 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
+          function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.4.1 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
           function download_scp_firmware() { git clone --single-branch https://github.com/ARM-software/SCP-firmware.git $HOME/scp-firmware || echo Nervermind; }
 
           ccache -s -v


core/drivers/crypto/se050/adaptors/utils/utils.c
```diff
@@ -77,7 +77,7 @@ static sss_status_t generate_oid(uint32_t *val)
 			return kStatus_SSS_Fail;
 
 		oid = OID_MIN + (random & OID_MAX);
-		if (oid > OID_MAX)
+		if (oid > OID_MAX || !(oid & 0xFF))
```
jforissier, line 80: This would deserve a comment I think. Why is 0 a the less significant byte causes an issue? Why can't this particular case be fixed somewhere else? (or why is is better to do it like this anyways?)

jforissier (commented):
You're giving a 64-bit value as an example but `oid` is an `uint32_t`, I don't get it.

ldts:
thanks @jforissier 

When the operation is to be performed with a key stored in the secure element, the key is accessed via a **watermarked handle stored in the key-pair**: the handle value is [**WATERMARK:HANDLE**] where each part is 4 bytes.
https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/adaptors/utils/utils.c#L140.

Before operating with the keypair, the driver checks if the private part of the key is indeed a handle and not an actual key
https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/core/ecc.c#L328

In practice, a handle such as` 0x1a.f2.9d.00 `(4 bytes) becomes WATERMARK `0x57.72.15.66.1a.f2.9d.00 `(8 bytes).

This value then then gets converted to a bignum and passed as a private key in the key-pair.

When the ECC/RSA operation is requested, the driver checks whether the key is a handle; however on the conversion back to binary the bignum turns into `0x57.72.15.66.1a.f2.9d` (7 bytes) therefore losing the first byte.

The check at https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/adaptors/utils/utils.c#L147 will fail and so will the operation.

I think I mentioned this  in another thread but perhaps we could consider having a _handle_ field in the keypair definition; and then of course revisit the GP Secure Element API in OP-TEE that was removed 4 years ago when we didn't have any users.

core/drivers/crypto/se050/adaptors/utils/utils.c
ldts: I only did it here because it is contained to the driver's use case - I havent seen complains/issues with the way bignum operate.

jforissier:
> thanks @jforissier
> 
> When the operation is to be performed with a key stored in the secure element, the key is accessed via a **watermarked handle stored in the key-pair**: the handle value is [**WATERMARK:HANDLE**] where each part is 4 bytes. https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/adaptors/utils/utils.c#L140.

Ah, that makes sense, thank you.

> The check at https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/adaptors/utils/utils.c#L147 will fail and so will the operation.

Isn't this the problem? You know the bignum should to be 64-bit. If it's not, the only reason is leading zeroes, so when converting from bignum back to 64-bit binary, you could just shift the pointer to properly pad with zeroes?

ldts:
> > thanks @jforissier
> > When the operation is to be performed with a key stored in the secure element, the key is accessed via a **watermarked handle stored in the key-pair**: the handle value is [**WATERMARK:HANDLE**] where each part is 4 bytes. https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/adaptors/utils/utils.c#L140.
> 
> Ah, that makes sense, thank you.
> 
> > The check at https://github.com/OP-TEE/optee_os/blob/master/core/drivers/crypto/se050/adaptors/utils/utils.c#L147 will fail and so will the operation.
> 
> Isn't this the problem? You know the bignum should to be 64-bit. If it's not, the only reason is leading zeroes, so when converting from bignum back to 64-bit binary, you could just shift the pointer to properly pad with zeroes?

no not always - the bignum could have an actual proper key and not a 8 byte watermarked handle...so I cant assume the size difference is a side effect...it is a bit of a pain having to do this watermark so I hope I can remove it soon.

jforissier (commented):

PR #5789, opened 2023-01-24
Provision TEE threads for system invocations

etienne-lms:
Proposal to add an SMC function ID to OP-TEE entry for system operations invocation using provisioned thread contexts.
Related to ~~https://github.com/linaro-swg/linux/pull/108~~ Linux [[PATCH 1/2]](https://lore.kernel.org/lkml/20230130094157.1082712-1-etienne.carriere@linaro.org/) and [[PATCH 2/2]](https://lore.kernel.org/lkml/20230130094157.1082712-2-etienne.carriere@linaro.org/).

(edited: updated Linux counterpart  refs)


core/include/kernel/thread.h
```diff
@@ -19,6 +19,7 @@
 #define THREAD_FLAGS_COPY_ARGS_ON_RETURN	BIT(0)
 #define THREAD_FLAGS_FOREIGN_INTR_ENABLE	BIT(1)
 #define THREAD_FLAGS_EXIT_ON_FOREIGN_INTR	BIT(2)
+#define THREAD_FLAGS_SYSTEM_SERVICE		BIT(3)
```
jenswi-linaro, line 22: What's the point of this?

mk/config.mk
```diff
@@ -112,6 +112,9 @@ CFG_WITH_SOFTWARE_PRNG ?= y
 # Number of threads
 CFG_NUM_THREADS ?= 2
 
+# Number of threads among CFG_NUM_THREADS provisioned for system invocation
```
jenswi-linaro, line 115: Please add something like. ", for instance, SMCI server."

core/arch/arm/kernel/thread.c
```diff
@@ -215,35 +216,77 @@ static void init_regs(struct thread_ctx *thread, uint32_t a0, uint32_t a1,
...
 
 	assert(l->curr_thread == THREAD_ID_INVALID);
 
 	thread_lock_global();
 
-	for (n = 0; n < CFG_NUM_THREADS; n++) {
-		if (threads[n].state == THREAD_STATE_FREE) {
-			threads[n].state = THREAD_STATE_ACTIVE;
-			found_thread = true;
+	for (n = 0; n < CFG_NUM_THREADS; n++)
```
jenswi-linaro, line 255: This seems quite complicated and scans the threads more than once. How about something like (which also takes care of the layer violation with regards to OPTEE_SMC_CALL_SYSTEM_WITH_REGD_ARG):
``` diff
--- a/core/arch/arm/kernel/thread.c
+++ b/core/arch/arm/kernel/thread.c
@@ -215,65 +215,77 @@ static void init_regs(struct thread_ctx *thread, uint32_t a0, uint32_t a1,
 }
 #endif /*ARM64*/
 
-static void __thread_alloc_and_run(uint32_t a0, uint32_t a1, uint32_t a2,
-				   uint32_t a3, uint32_t a4, uint32_t a5,
-				   uint32_t a6, uint32_t a7,
+static int find_free_thread(size_t start_idx, size_t count)
+{
+	size_t n = 0;
+
+	for (n = start_idx; n < start_idx + count; n++)
+		if (threads[n].state == THREAD_STATE_FREE) {
+			return n;
+
+	return -1;
+}
+
+static void __thread_alloc_and_run(bool sys_thread, uint32_t a0, uint32_t a1,
+				   uint32_t a2, uint32_t a3, uint32_t a4,
+				   uint32_t a5, uint32_t a6, uint32_t a7,
 				   void *pc)
 {
 	struct thread_core_local *l = thread_get_core_local();
-	bool found_thread = false;
-	size_t n = 0;
+	int i = -1;
 
 	assert(l->curr_thread == THREAD_ID_INVALID);
 
 	thread_lock_global();
 
-	for (n = 0; n < CFG_NUM_THREADS; n++) {
-		if (threads[n].state == THREAD_STATE_FREE) {
-			threads[n].state = THREAD_STATE_ACTIVE;
-			found_thread = true;
-			break;
-		}
-	}
+	if (sys_thread)
+		i = find_free_thread(CFG_NUM_THREADS - CFG_NUM_SYSTEM_THREADS,
+				     CFG_NUM_SYSTEM_THREADS);
+	if (i == -1)
+		i = find_free_thread(0, CFG_NUM_THREADS -
+					CFG_NUM_SYSTEM_THREADS);
+
+	if (i >= 0)
+		threads[i].state = THREAD_STATE_ACTIVE;
 
 	thread_unlock_global();
 
-	if (!found_thread)
+	if (i == -1)
 		return;
 
-	l->curr_thread = n;
+	l->curr_thread = i;
 
-	threads[n].flags = 0;
-	init_regs(threads + n, a0, a1, a2, a3, a4, a5, a6, a7, pc);
+	threads[i].flags = 0;
+	init_regs(threads + i, a0, a1, a2, a3, a4, a5, a6, a7, pc);
 #ifdef CFG_CORE_PAUTH
 	/*
 	 * Copy the APIA key into the registers to be restored with
 	 * thread_resume().
 	 */
-	threads[n].regs.apiakey_hi = threads[n].keys.apia_hi;
-	threads[n].regs.apiakey_lo = threads[n].keys.apia_lo;
+	threads[i].regs.apiakey_hi = threads[i].keys.apia_hi;
+	threads[i].regs.apiakey_lo = threads[i].keys.apia_lo;
 #endif
 
 	thread_lazy_save_ns_vfp();
 
 	l->flags &= ~THREAD_CLF_TMP;
-	thread_resume(&threads[n].regs);
+	thread_resume(&threads[i].regs);
 	/*NOTREACHED*/
 	panic();
 }
 
-void thread_alloc_and_run(uint32_t a0, uint32_t a1, uint32_t a2, uint32_t a3,
-			  uint32_t a4, uint32_t a5)
+void thread_alloc_and_run(bool sys_thread, uint32_t a0, uint32_t a1,
+			  uint32_t a2, uint32_t a3, uint32_t a4, uint32_t a5)
 {
-	__thread_alloc_and_run(a0, a1, a2, a3, a4, a5, 0, 0,
+	__thread_alloc_and_run(sys_thread, a0, a1, a2, a3, a4, a5, 0, 0,
 			       thread_std_smc_entry);
 }
 
 #ifdef CFG_SECURE_PARTITION
 void thread_sp_alloc_and_run(struct thread_smc_args *args __maybe_unused)
 {
-	__thread_alloc_and_run(args->a0, args->a1, args->a2, args->a3, args->a4,
-			       args->a5, args->a6, args->a7,
+	__thread_alloc_and_run(false, args->a0, args->a1, args->a2, args->a3,
+			       args->a4, args->a5, args->a6, args->a7,
 			       spmc_sp_thread_entry);
 }
 #endif
diff --git a/core/arch/arm/kernel/thread_optee_smc.c b/core/arch/arm/kernel/thread_optee_smc.c
index e4b803e0169e..1f3c8c3113ab 100644
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -70,7 +70,9 @@ uint32_t thread_handle_std_smc(uint32_t a0, uint32_t a1, uint32_t a2,
 		thread_resume_from_rpc(a3, a1, a2, a4, a5);
 		rv = OPTEE_SMC_RETURN_ERESUME;
 	} else {
-		thread_alloc_and_run(a0, a1, a2, a3, 0, 0);
+		bool sys_thread = (a0 == OPTEE_SMC_CALL_SYSTEM_WITH_REGD_ARG);
+
+		thread_alloc_and_run(sys_thread, a0, a1, a2, a3, 0, 0);
 		rv = OPTEE_SMC_RETURN_ETHREAD_LIMIT;
 	}
 
diff --git a/core/arch/arm/kernel/thread_spmc.c b/core/arch/arm/kernel/thread_spmc.c
index 3b4ac0b4e35c..41f197ebde6c 100644
--- a/core/arch/arm/kernel/thread_spmc.c
+++ b/core/arch/arm/kernel/thread_spmc.c
@@ -451,8 +451,8 @@ static void handle_yielding_call(struct thread_smc_args *args)
 				       0);
 		res = TEE_ERROR_BAD_PARAMETERS;
 	} else {
-		thread_alloc_and_run(args->a1, args->a3, args->a4, args->a5,
-				     args->a6, args->a7);
+		thread_alloc_and_run(false, args->a1, args->a3, args->a4,
+				     args->a5, args->a6, args->a7);
 		res = TEE_ERROR_BUSY;
 	}
 	spmc_set_args(args, FFA_MSG_SEND_DIRECT_RESP_32,

File: core/arch/arm/include/sm/optee_smc.h [hunk 4]
```diff
@@ -317,6 +322,9 @@
 /* Secure world supports pre-allocating RPC arg struct */
 #define OPTEE_SMC_SEC_CAP_RPC_ARG		BIT(6)
 
+/* Secure world provisions thread for system service invocation */
+#define OPTEE_SMC_SEC_CAP_SYSTEM_THREAD		BIT(8)
```
jenswi-linaro, line 326: What happened with `BIT(7)` and why is this separated from the other "SEC_CAPs"?

core/include/kernel/thread.h
etienne-lms: This flags allows to track which active system threads are consumed, in function `enough_free_system_thread()`.

core/arch/arm/kernel/thread.c
etienne-lms: Ok, simpler. I'll use your sequence.

mk/config.mk
etienne-lms: ok

core/arch/arm/include/sm/optee_smc.h
etienne-lms: indeed. sorry, a leftover  from https://github.com/OP-TEE/optee_os/pull/5793 that adds another funcID.

etienne-lms:
Comment addressed (in the right way, i hope).
I've added a tee_caps.c driver i tried to make generic. Tell me if it makes sense.

mk/config.mk
etienne-lms: I missed this comment. I've added a fixup commit.

core/include/tee/tee_caps.h
```diff
@@ -0,0 +1,43 @@
...
+#include <stdbool.h>
+
+/*
+ * OP-TEE capabilities negocatied with normal world (REE)
+ * These IDs are bit positions in a bitstring object.
+ *
+ * OPTEE_CAP_SYSTEM_THREAD
+ *	When enabled, normal world and TEE agreed generic yield invocations
+ *	consume the standard thread pool and system invocations consume the
+ *	system threads.
```
etienne-lms, line 18: s/thread pool/thread contexts pool/
s/system threads/system thread contexts pool/

jenswi-linaro:
>I've added a tee_caps.c driver i tried to make generic. Tell me if it makes sense.

I don't think it's needed. In principle, all the capabilities we announce to the normal world are new features that it doesn't matter if the normal world uses or not. I mean, if normal world doesn't use one particular feature there's nothing OP-TEE can do about that. We should try to keep things as simple as possible.

core/arch/arm/kernel/thread.c
```diff
@@ -215,65 +217,83 @@ static void init_regs(struct thread_ctx *thread, uint32_t a0, uint32_t a1,
...
 
 	thread_lock_global();
 
-	for (n = 0; n < CFG_NUM_THREADS; n++) {
-		if (threads[n].state == THREAD_STATE_FREE) {
-			threads[n].state = THREAD_STATE_ACTIVE;
-			found_thread = true;
-			break;
-		}
+	if (CFG_NUM_SYSTEM_THREADS &&
```
jenswi-linaro, line 243: In my patch, I wrote this part in 6 lines and I believe they are both easier to read and for the compiler to optimize.

etienne-lms:
>> I've added a tee_caps.c driver i tried to make generic. Tell me if it makes sense.
>
> I don't think it's needed. In principle, all the capabilities we announce to the normal world are new features that it doesn't matter if the normal world uses or not. I mean, if normal world doesn't use one particular feature there's nothing OP-TEE can do about that. We should try to keep things as simple as possible.

I agree with making things simple.... hmmm, okay. I'll drop that. That means updating device images with this feature needs linux kernel image to be updated first then TEE image to be updated unless what the provisioned TEE system threads will not be used by the old Linux which may affect the platform.

etienne-lms:
Seen that the series makes changes that are reverted afterward, are you okay i squash fixup commits for the next proposal?

core/arch/arm/kernel/thread.c
etienne-lms: ok

jenswi-linaro:
> Seen that the series makes changes that are reverted afterward, are you okay i squash fixup commits for the next proposal?

Please do.

etienne-lms:
Udpdated.

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +572,9 @@
 /* See OPTEE_SMC_CALL_WITH_REGD_ARG above */
 #define OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG	U(19)
 
+/* See OPTEE_SMC_CALL_SYSTEM_WITH_REGD_ARG above */
+#define OPTEE_SMC_FUNCID_CALL_SYSTEM_WITH_REGD_ARG	U(22)
```
etienne-lms, line 576: s/22/20/

jenswi-linaro:
With OPTEE_SMC_FUNCID_CALL_SYSTEM_WITH_REGD_ARG updated:

Please wait with merging this until we know how the counterpart in the Linux kernel driver is received.

jenswi-linaro:
Feel free to squash and apply the tag.

etienne-lms:
Tags applied (@jenswi-linaro, i changed the plat-stm32mp1 R-b into an A-b)

etienne-lms:
Updated to address comments at upstream.
Linux driver patches v2 can be [found here](https://lore.kernel.org/lkml/20230214152047.1143106-1-etienne.carriere@linaro.org/).

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -203,6 +209,8 @@
 	OPTEE_SMC_STD_CALL_VAL(OPTEE_SMC_FUNCID_CALL_WITH_RPC_ARG)
 #define OPTEE_SMC_CALL_WITH_REGD_ARG \
 	OPTEE_SMC_STD_CALL_VAL(OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG)
+#define OPTEE_SMC_CALL_SYSTEM_WITH_REGD_ARG \
```
jenswi-linaro, line 212: Sumit's suggestion with a bit to indicate if it's a system request made a lot of sense. That way we don't add arbitrary limitations.

jenswi-linaro (commented):
I'm also responding in the kernel thread for the kernel side of things.

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +572,55 @@
...
+ * Possible return values:
+ * OPTEE_SMC_RETURN_UNKNOWN_FUNCTION	Trusted OS does not recognize this
+ *                                      function.
+ * OPTEE_SMC_RETURN_OK			Call successfully completed.
+ * OPTEE_SMC_RETURN_ETHREAD_LIMIT	Number of Trusted OS threads exceeded
+ *                                      for the request.
+ */
+#define OPTEE_SMC_FUNCID_CALL_RESERVE_SYS_THREAD	U(21)
+#define OPTEE_SMC_CALL_RESERVE_SYS_THREAD \
+	OPTEE_SMC_STD_CALL_VAL(OPTEE_SMC_FUNCID_CALL_RESERVE_SYS_THREAD)
```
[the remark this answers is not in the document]
etienne-lms: right, I messed this.

core/arch/arm/tee/entry_fast.c
```diff
@@ -217,6 +221,32 @@ static void get_async_notif_value(struct thread_smc_args *args)
...
+static void request_system_thread_context(struct thread_smc_args *args)
+{
+	if (IS_ENABLED(CFG_RESERVED_SYSTEM_THREAD)) {
+		if (reserve_sys_thread())
+			args->a0 = OPTEE_SMC_RETURN_ETHREAD_LIMIT;
+		else
+			args->a0 = OPTEE_SMC_RETURN_OK;
+
+	} else {
+		args->a0 = OPTEE_SMC_RETURN_OK;
```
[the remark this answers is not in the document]
etienne-lms: right!

core/arch/arm/kernel/thread.c
```diff
@@ -35,6 +35,22 @@
 #include <trace.h>
 #include <util.h>
 
+/*
```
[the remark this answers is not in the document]
etienne-lms: With the new scheme, we expect REE to call optee to reserve system threads. This could be done there are already TEE threads assigned, so we can have the thread context pool split in 2 parts (upper indices for system invocation, lower part for the rest). We must count how many threads are free hence to logic proposed here.

core/arch/arm/include/sm/optee_smc.h
etienne-lms: I don't see where to store that bit. ABI for `OPTEE_SMC_CALL_WITH*_ARG` leave no room for an extra bit, we need new funcIDs, reversing an ABI register to carry that information (e.g. bit0 of input `a4` argument).

jenswi-linaro: Sumit suggested BIT(15) in the SMC ID if I remember correctly.

core/arch/arm/kernel/thread.c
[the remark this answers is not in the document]
jenswi-linaro: I don't see why you need to count. If threads are already in use it's likely they are staying off the higher ranges. Are you making all this trouble to guard against the pathological case where all threads have been used and the last of these happens to be one that runs forever?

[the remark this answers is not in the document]
etienne-lms: With the implementation proposed here, there is no specific pre-provisioning of thread contextes on OP-TEE side. It's the REE that requests some system contexts reservation. With this scheme, when REE requests reservation of a system thread context, OP-TEE has to deal with already allocated threads. A pros is that running OP-TEE with a kernel not supporting TEE system context, will be able to use all available TEE thread context.
I'll try to simplify but I fear cases where OP-TEE accept to reserve a system context but fails to let REE use it because, as you said, the upper thread(s) are already in use.

core/arch/arm/include/sm/optee_smc.h
etienne-lms: I didn't address this comment yet.

jenswi-linaro:
At some point, you mentioned that you only wanted to reserve this thread if it was going to be used by normal world. And the reason for that is that you can't afford to waste an unused thread. With paging enabled an unused thread uses ~700 bytes if I remember correctly. Of these ~700 bytes are ~500 bytes VFP state which in theory could be saved on the stack instead. So we're basically doing this to save 200-700 bytes of nonpaged memory?

etienne-lms:
A thread costs 912 bytes of bss (ctx's and pgt entries) and 4kB of nozi (mmu tbl) on 32b.
On 64b, it costs 768 bytes of bss and 80B of nozi (1 mmu ul1 table).
Numbers from current post 3.20.0 revision.

etienne-lms:
sorry, I messed up retrieving the numbers... I'll update.

etienne-lms:
Cost of a thread (`CFG_NUM_THREADS`) in static mapping. Numbers from current post 3.20.0 revision.
64b.LPAE: 1872 bytes of bss + 4kbyte of nozi
32b/LPAE: 912 bytes of bss + 4kbyte of nozi
32b/32b-mmu: 768 bytes of bss + 128 bytes of nozi
bss stores ctx's and pgt entries.  nozi store mmu tables.

jenswi-linaro:
Thanks for the detailed figures. Are you using 32b/32b-mmu in your use case?

etienne-lms:
No, the platform is using LPAE. But could be an argument for using 32b-mmu on that platform.

core/arch/arm/include/sm/optee_smc.h
b49020: How about we start calling **top 4 bits** for funcIDs as flags in OP-TEE SMC ABI? Since they are already zeros and would give us enough flexibility to add features like system threads etc. Otherwise the function IDs list will keep on increasing given all the combinatorial involved and hard to keep track off. Also, still there will be 12 bits for real funcIDs which allows that list to go upto value: **8191**.

jenswi-linaro: I'd like to have both, combinations explicitly defined so:
``` C
#define OPTEE_SMC_CALL_SYSTEM_THREAD_FLAG BIT(15)
#define OPTEE_SMC_CALL_SYSTEM_WITH_ARG \
        OPTEE_SMC_STD_CALL_VAL(OPTEE_SMC_FUNCID_CALL_WITH_ARG | \
                               OPTEE_SMC_CALL_SYSTEM_THREAD_FLAG)
#define OPTEE_SMC_CALL_SYSTEM_WITH_RPC_ARG \
        OPTEE_SMC_STD_CALL_VAL(OPTEE_SMC_FUNCID_CALL_WITH_RPC_ARG | \
                               OPTEE_SMC_CALL_SYSTEM_THREAD_FLAG)
#define OPTEE_SMC_CALL_SYSTEM_WITH_REGD_ARG \
        OPTEE_SMC_STD_CALL_VAL(OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG | \
                               OPTEE_SMC_CALL_SYSTEM_THREAD_FLAG)
```

The `OPTEE_SMC_CALL_SYSTEM_THREAD_FLAG` doesn't make sense for all SMCs so we should still define all SMCs, but with the bit defined there's still some structure in how an SMC is constructed so masking, etc can still be done where that's convenient.

b49020: Okay that's fine with me.

etienne-lms:
From recent discussions, Linux should have all info needed to monitor TEE thread entry and provision system entry contexts for dedicated services. I've created https://github.com/linaro-swg/linux/pull/109 on Linux kernel where no change is needed in OP-TEE OS.

etienne-lms:
Closing this P-R in favor to a REE only system thread management.

PR #5793, opened 2023-01-25
(take 2) NS Interrupt notification over async notif

etienne-lms:
Supersedes https://github.com/OP-TEE/optee_os/pull/5761

Adds OP-TEE interrupt event notification to normal world with SMC transport, using existing async notif support.
Adds 2 SMC fastcall funcIDs to retrieve interrupt events and to mask an interrupt.
Adds an Itr Notif PTA to expose interrupt notif config means: set active state, set wakeup state.
Adds Itr Notif tests in Invoke test PTA: triggers and test some interrupt notifications.
Related regression tests in [PR-640](https://github.com/OP-TEE/optee_test/pull/640), tested on qemu_virt and stm32mp15.

Related to a change series in Linux kernel posted to LKML:
[PATCH v2 1/3: dt bindings](https://lore.kernel.org/lkml/20230124105643.1737250-1-etienne.carriere@linaro.org/), [PATCH v2 2/3: optee irqchip](https://lore.kernel.org/lkml/20230124105643.1737250-2-etienne.carriere@linaro.org/) and [PATCH v2 3/3: state/wakeup services](https://lore.kernel.org/lkml/20230124105643.1737250-3-etienne.carriere@linaro.org/).


lib/libutils/ext/include/bitstring.h
```diff
@@ -144,4 +144,30 @@ typedef	unsigned char bitstr_t;
 	*(value) = _value; \
 } while (0)
 
+#define	bit_ffs_from(name, nbits, startbit, value) do { \
```
jenswi-linaro, line 147: It looks like `bit_ffs()` could then be defined as:
``` C
 #define bit_ffs(name, nbits, value) bit_ffs_from((name), (nbits), 0, (value))
```

I think it would be useful to get rid of some of the duplicated code.

lib/libutils/ext/include/bitstring.h
```diff
@@ -144,4 +144,30 @@ typedef	unsigned char bitstr_t;
...
+	register bitstr_t *_name = (name); \
```
jenswi-linaro, line 148: Please drop the `register` keyword, it doesn't add any value.

lib/libutils/ext/include/bitstring.h
```diff
@@ -144,4 +144,30 @@ typedef	unsigned char bitstr_t;
...
+	register int _byte, _nbits = (nbits); \
+	register int _stopbyte = _bit_byte(_nbits - 1), _value = -1; \
+	register int _startbyte = _bit_byte(startbit); \
+	register int _test_bit_mask = (uint8_t)(0xff << ((startbit) % 8)); \
```
jenswi-linaro, line 152: With `uint8_t` you add a dependency on `<stdint.h>`, I would probably have used `bitstr_t`  here instead to avoid that, especially since you already have `bitstr_t _lb` below.

jenswi-linaro (commented):
Comments for "libutils: add bit_ffs_from()"

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+		bit_ffs_from(notif_it_pending, max_bit, bit + 1, &bit);
+	} while (bit >= 0 && bit_test(notif_it_masked, bit));
+
+	*it_valid = (bit >= 0);
+	if (!*it_valid) {
+		*it_pending = false;
+		goto out;
+	}
+
+	/* Clear pending interrupt from the record */
```
jenswi-linaro, line 180: This comment should be after the next line.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+	*it_pending = (bit >= 0);
+
+out:
+	cpu_spin_unlock_xrestore(&it_lock, old_itr_status);
+
+	return res;
+}
+
+/* This function can execute in a fastcall interrupt masked context */
+static void set_it_mask(unsigned int id, bool do_mask)
```
jenswi-linaro, line 199: We have "id", "it value", and "it number" for the same thing.  I'd prefer if we could stick to one name, I think "it number" is the clearest.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+		bit_clear(notif_it_masked, id);
+
+		if (bit_test(notif_it_pending, id))
+			itr_raise_pi(CFG_CORE_ASYNC_NOTIF_GIC_INTID);
+	}
+
+	cpu_spin_unlock_xrestore(&it_lock, old_itr_status);
+}
+
+/* Speculation safe search of notif_it instance from a ID value */
```
jenswi-linaro, line 222: It seems this comment is an answer to a question during an earlier review.
Why shouldn't this function (or any other for that matter) be speculation safe?

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+ * Notify an interrupt event to normal world
+ *
+ * @notif_it Reference to registered notifier
+ */
+void notif_it_raise_event(struct notif_it *notif_it)
+{
+	uint32_t old_itr_status = 0;
+
+	if (find_notif_it(notif_it->id) != notif_it) {
+		DMSG("Invalid notif handle");
```
jenswi-linaro, line 336: Can this happen? Perhaps an assert would be better here.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+ * @id The interrupt number/ID
+ * @ops Operations for that interrupt notifier
+ * @notif_it Preallocated instance address or output reference
+ *   Input: if not NULL and pointing to a non NULL cell, this is
+ *   the preallocated struct notif_it to be used.
+ *   Input: if not NULL and pointing to NULL, struct notif_it
+ *   instance must be allocated.
+ *   Output: if not NULL, points to registered struct notif_it
+ *
+ * Return a TEE_Result compliant code
```
jenswi-linaro, line 362: This line doesn't add any value.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+ */
+TEE_Result notif_it_register(unsigned int id, const struct notif_it_ops *ops,
```
jenswi-linaro, line 364: How about letting the caller populate `struct notif_it` instead like it's done with `notif_register_driver()`?
It would simplify things a bit and this is likely a static configuration anyway.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+
+	assert(!bit_test(notif_it_pending, id));
+	old_itr_status = cpu_spin_lock_xsave(&it_lock);
+	bit_clear(notif_it_pending, id);
+	cpu_spin_unlock_xrestore(&it_lock, old_itr_status);
+
+	if (notif_it_ref && *notif_it_ref) {
+		notif = *notif_it_ref;
+		memset(notif, 0, sizeof(*notif));
+		if (!is_unpaged(notif)) {
```
jenswi-linaro, line 388: Assert would make more sense.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+			DMSG("notif instance must be unpaged");
+			return TEE_ERROR_GENERIC;
+		}
+	} else {
+		notif = calloc(1, sizeof(*notif));
```
jenswi-linaro, line 393: I don't think we need two ways of registering an interrupt notifier.

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+ * Return a TEE_Result compliant code
+ */
+TEE_Result notif_it_unregister(struct notif_it *notif)
+{
+	uint32_t old_itr_status = 0;
+
+	if (!notif)
+		return TEE_SUCCESS;
+
+	if (find_notif_it(notif->id) != notif)
```
jenswi-linaro, line 430: Assert?

core/kernel/notif.c
```diff
@@ -127,6 +411,14 @@ void notif_register_driver(struct notif_driver *ndrv)
 	SLIST_INSERT_HEAD(&notif_driver_head, ndrv, link);
 
 	cpu_spin_unlock_xrestore(&notif_lock, old_itr_status);
+
+#if defined(CFG_CORE_IT_NOTIF)
+	old_itr_status = cpu_spin_lock_xsave(&it_lock);
+
+	bit_nset(notif_it_masked, 0, (int)NOTIF_IT_VALUE_MAX);
```
jenswi-linaro, line 471: Why is this done each time we register a new driver to be notified of an event?

mk/config.mk
```diff
@@ -935,6 +935,19 @@ endif
...
+# for sending asynchronous notifications to normal world of dedicated
+# interrupt events to be handle in an interrupt context in normal world.
+# CFG_CORE_IT_NOTIF depends on CFG_CORE_ASYNC_NOTIF.
+# When enabled, OP-TEE can notify interrupt events identified by an
+# interrupt number from 0 to CFG_CORE_IT_NOTIF_MAX.
+CFG_CORE_IT_NOTIF ?= n
+
+$(eval $(call cfg-depends-all,CFG_CORE_IT_NOTIF,CFG_CORE_ASYNC_NOTIF))
+ifeq ($(CFG_CORE_IT_NOTIF),y)
+CFG_CORE_IT_NOTIF_MAX ?= 31
```
jenswi-linaro, line 953: Would it make sense to enable `CFG_CORE_IT_NOTIF` only if a value is provided for `CFG_CORE_IT_NOTIF_MAX`?

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+	notif_it_hdl[notif_it->id] = notif_it;
+}
+
+/*
+ * Mask/unmask a interrupt notification
+ *
+ * @id Interrupt identifier
+ * @do_mask True to mask the event, false to unmask the event
+ *
+ * This function is called from an fastcall context.
```
jenswi-linaro, line 252: s/an/a/

jenswi-linaro (commented):
Comments for "core: notif: interrupt notification"

jenswi-linaro (commented):
For commit "core: notif: fix input comment typo"

core/arch/arm/tee/entry_fast.c
```diff
@@ -217,6 +239,69 @@ static void get_async_notif_value(struct thread_smc_args *args)
...
 
+static void get_pending_notif(struct thread_smc_args *args)
+{
+	bool it_valid = false;
+	bool it_pending = false;
+	int it[5] = { -1, -1, -1, -1, -1 };
+	ssize_t cnt = 0;
+
+	do {
+		uint32_t it_num  = notif_it_get_value(&it_valid, &it_pending);
```
jenswi-linaro, line 249: I'd rather have one single function to dig out all the needed information, something like:
``` C
/*
 * Report pending interrupts and asynchronous notification values.
 * @do_bottom_half: set to true if NOTIF_VALUE_DO_BOTTOM_HALF is pending, and
 *                  clears it as delivered.
 * @async_value:    set to true if any other asynchronous notification value
 *                  is pending.
 * @it_nums:        array of size *@it_count to report pending interrupts
 * @it_count:       at most @it_count interrupts are reported in @it_nums,
 *                  @it_count is updated with the number of reported interrupts.
 */
void notif_get_pending(bool *do_bottom_half, bool *async_value,
                       uint16_t *it_nums, size_t *it_count);
```

jenswi-linaro (commented):
Comments for "core: arm: smc: entry service for interrupt notification"

Why must normal world use a capability to tell OP-TEE that it wants to use interrupt notifications? Isn't it enough that normal world tries to enable or configure it to tell that?

core/pta/it_notif.c
```diff
@@ -0,0 +1,98 @@
...
+static TEE_Result activate_it(uint32_t param_types,
+			      TEE_Param params[TEE_NUM_PARAMS])
+{
+	if (param_types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+					   TEE_PARAM_TYPE_NONE,
+					   TEE_PARAM_TYPE_NONE,
+					   TEE_PARAM_TYPE_NONE))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	return notif_it_set_state(params[0].value.a, true /* enable */);
```
jenswi-linaro, line 22: Since we have a single `notif_it_set_state()` why not supply `params[0].value.b` as the second parameter? It should reduce the interface a bit.

lib/libutee/include/pta_it_notif.h
```diff
@@ -0,0 +1,40 @@
...
+#define __PTA_IT_NOTIF_H
+
+#define PTA_IT_NOTIF_UUID { \
+		0x4461e5c7, 0xb523, 0x4b73, \
+		{ 0xac, 0xed, 0x75, 0xad, 0x2b, 0x9b, 0x59, 0xa1 } \
+	}
+/*
+ * PTA_IT_NOTIF_ACTIVATE_RESOURCE - Enable a interrupt notification
+ *
+ * [in]  params[0].value.a     Interrupt ID
```
jenswi-linaro, line 15: Please use a single term for the interrupt and also describe what it represents and how values are determined.

mk/config.mk
```diff
@@ -941,7 +941,10 @@ CFG_CORE_ASYNC_NOTIF ?= n
 # CFG_CORE_IT_NOTIF depends on CFG_CORE_ASYNC_NOTIF.
 # When enabled, OP-TEE can notify interrupt events identified by an
 # interrupt number from 0 to CFG_CORE_IT_NOTIF_MAX.
+# CFG_IT_NOTIF_PTA embeds interrupt notification configuration services
```
jenswi-linaro, line 944: s/embeds/enables/

jenswi-linaro (commented):
Comments for "pta: it_notif: non-services for interrupt notification"

core/kernel/notif.c
```diff
@@ -263,6 +263,25 @@ void notif_it_set_mask(unsigned int id, bool do_mask)
...
+ * Return whether interrupt notification is masked or unmasked
+ *
+ * @id Interrupt identifier
+ *
+ * Return true or false upon interrupt notif mask value
+ */
+bool notif_it_is_masked(unsigned int id)
+{
+	assert(find_notif_it(id));
+	return bit_test(notif_it_masked, id) != 0;
```
jenswi-linaro, line 275: We should test that `id` is smaller than `NOTIF_IT_VALUE_MAX` too, same below.

core/kernel/notif.c
```diff
@@ -263,6 +263,25 @@ void notif_it_set_mask(unsigned int id, bool do_mask)
 		set_it_mask(id, do_mask);
 }
 
+/*
```
jenswi-linaro, line 265: We're normally documenting the interface in the .h files.

jenswi-linaro (commented):
Comments for "core: notif: add status helper functions for embedded tests"

core/pta/tests/misc.c
```diff
@@ -21,6 +25,226 @@
  */
 #define LOG(...)
 
+#ifdef CFG_IT_NOTIF_TEST
+static struct notif_it test_it_notif[TEST_IT_NOTIF_CNT];
+/* Helper to release registered resource in case of error */
+static bool test_it_notif_registered[TEST_IT_NOTIF_CNT];
+
+static unsigned int idx2id(unsigned int test_idx)
```
jenswi-linaro, line 33: So now we have interrupt indexes too? This is starting to get really confusing.

jenswi-linaro (commented):
Comment for "core: pta: test: invoke pta command to test interrupt notif"

core/include/kernel/notif.h
```diff
@@ -108,11 +108,16 @@ struct notif_driver {
  * @ops Notifier callback operations or NULL if none
  * @flags: NOTIF_IT_FLAG_* properties
  * @priv Driver private data of NULL
+ * @set_mask Copy of set_mask entry for pager optims
```
jenswi-linaro, line 111: This is a big ugly in my opinion. Better to deal with the problem in place, or accept the penalty.

jenswi-linaro (commented):
Comment for "core: notif: add set_mask() fastcall callback handler"

jenswi-linaro:
It's unfortunate that we started to abbreviate interrupt as "IT", here it's clearly too short without a proper context. At other places, we abbreviate it as "ITR" or "INTR". The latter is the clearest but "ITR" isn't too bad either.

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +565,94 @@
...
+ * in normal world.
+ *
+ * Call requests usage:
+ * a0	SMC Function ID, OPTEE_SMC_CONFIGURE_NOTIF_IT
+ * a1	Interrupt identifier value
+ * a2	Bit[0]: 1 to configure interrupt state with a3[bit 0], else 0
+ *	Bit[1]: 1 to configure interrupt mask  with a3[bit 1], else 0
+ *	Bit[2]: 1 to set if wakeup from low power state, upon a3[bit 2], else 0
+ *      Bit[31:3] Reserved for future use, MBZ
+ * a3	Bit[0]: 1 to enable the interrupt, 0 disabled. Applies if a2[bit 0]=1
```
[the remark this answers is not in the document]
etienne-lms: The API described here and implementation don't match.

The initial goal of this funcID is to mask/unmask an ITR notif. Then I though the funcID could be later extended to act on more controls, hence a bit to say is we mant to mask/unmask (a2[bit0]) and a bit to state the desired mask level (a3[bit0]).

Do you prefer 1 funcID for 1 purpose and I rename `OPTEE_SMC_CONFIG_NOTIF_ITR` into `OPTEE_SMC_SET_ITR_MASK` and use it only for mask/unmask requests?

core/arch/arm/tee/entry_fast.c
```diff
@@ -66,6 +67,18 @@ static void tee_entry_exchange_capabilities(struct thread_smc_args *args)
 {
 	bool res_shm_en = IS_ENABLED(CFG_CORE_RESERVED_SHM);
 	bool dyn_shm_en __maybe_unused = false;
+	unsigned int notif_it_max_lines = 0;
+	uint32_t nsec_caps_mask = OPTEE_SMC_NSEC_CAP_UNIPROCESSOR |
+				  OPTEE_SMC_SEC_CAP_IT_NOTIF;
```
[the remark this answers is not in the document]
etienne-lms: Then the cap falg should be renamed `OPTEE_SMC_CAP_ITR_NOTIF`.
To enable the feature, non-secure should set the cap and secure world should also set the cap.

core/arch/arm/tee/entry_fast.c
```diff
@@ -116,6 +129,15 @@ static void tee_entry_exchange_capabilities(struct thread_smc_args *args)
 
 	args->a1 |= OPTEE_SMC_SEC_CAP_RPC_ARG;
 	args->a3 = THREAD_RPC_MAX_NUM_PARAMS;
+
+	if (IS_ENABLED(CFG_CORE_IT_NOTIF) &&
+	    nsec_caps & OPTEE_SMC_SEC_CAP_IT_NOTIF) {
```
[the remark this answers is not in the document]
etienne-lms: I think we should. It's better for OS upgrades that OP-TEE knows what NS is capable of. At least, OP-TEE can know non-secure won't be able to use ITR notifs.

core/arch/arm/tee/entry_fast.c
etienne-lms: ok, i'll do that.

core/arch/arm/tee/entry_fast.c
```diff
@@ -217,6 +239,69 @@ static void get_async_notif_value(struct thread_smc_args *args)
...
+
+	args->a1 = cnt;
+	if (it_pending)
+		args->a1 |= OPTEE_SMC_NOTIF_IT_PENDING;
+	if (notif_do_bottom_half_retrieved())
+		args->a1 |= OPTEE_SMC_NOTIF_DO_BOTTOM_HALF;
+	if (notif_async_value_is_pending())
+		args->a1 |= OPTEE_SMC_NOTIF_VALUE_PENDING;
+
+	if (--cnt >= 0)
```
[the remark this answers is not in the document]
etienne-lms: ok

core/include/kernel/notif.h
etienne-lms: Ok, i'll accept the penalty to make thing simpler.
If one needs an optim, let's deal with it by then.

core/kernel/notif.c
etienne-lms: fair

core/kernel/notif.c
```diff
@@ -106,6 +123,273 @@ void notif_send_async(uint32_t value)
...
+	*it_valid = (bit >= 0);
+	if (!*it_valid) {
+		*it_pending = false;
+		goto out;
+	}
+
+	/* Clear pending interrupt from the record */
+	res = bit;
+	bit_clear(notif_it_pending, res);
+	bit_ffs(notif_it_pending, max_bit, &bit);
```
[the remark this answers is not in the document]
etienne-lms: nothing. i'll remove. sorry.

core/kernel/notif.c
etienne-lms: ok `itr_num` here and there?

core/kernel/notif.c
etienne-lms: The comment should maybe be at line 227, before the for loop.

Antoher implementation could be:
```c
	if (itr_num > NOTIF_IT_VALUE_MAX)
		return NULL;
	itr_num = confine_array_index(itr_num, NOTIF_IT_VALUE_MAX + 1);

	return notif_it_hdl + itr_num;
```

core/kernel/notif.c
etienne-lms: ok

core/kernel/notif.c
etienne-lms: this is asserted at line above.

core/kernel/notif.c
etienne-lms: I agree

core/kernel/notif.c
etienne-lms: ok

core/kernel/notif.c
etienne-lms: Ok

core/kernel/notif.c
etienne-lms: :)

core/kernel/notif.c
etienne-lms: +1

core/kernel/notif.c
etienne-lms: This is the main async notif registering function (`notif_register_driver()`), not the itr notif registering function (`notif_it_register()`). The later masks its own itr with `set_it_mask(id, 1);`.

core/pta/it_notif.c
etienne-lms: Ok. And same for wakeup state commands.

core/pta/tests/misc.c
etienne-lms: sorry... I'll try to simplify that.

lib/libutee/include/pta_it_notif.h
etienne-lms: OK

core/arch/arm/tee/entry_fast.c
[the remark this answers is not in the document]
jenswi-linaro: Why? What's the difference of normal world supporting interrupt notifications and not configuring/enable any and normal world not supporting interrupt notifications? What would be different from OP-TEE point of view?

core/kernel/notif.c
jenswi-linaro: So this interface will only be called with an already controlled "id"?
My point is that assert catches things that "can't happen", but if they do happen it's due to a flaw inside OP-TEE only and not due to external input. If external input can cause this, then it must always be checked and we either panic() or return an error.

jenswi-linaro:
Please squash the updates.

etienne-lms (commented):
I'll squash the fixup commits.

etienne-lms:
Squashed fixup commits (+ few added fixes) and rebased on master.

etienne-lms:
Appended 2 fixup commits for issues reported by CI builds.

core/pta/itr_notif.c
```diff
@@ -0,0 +1,70 @@
...
+	case PTA_ITR_NOTIF_ACTIVATION:
+		return activate_itr_notifier(param_types, params);
+	case PTA_ITR_NOTIF_WAKEUP_SOURCE:
+		return activate_wakeup_source(param_types, params);
+	default:
+		return TEE_ERROR_NOT_IMPLEMENTED;
+	}
+}
+
+pseudo_ta_register(.uuid = PTA_ITR_NOTIF_UUID, .name = "itr_notif.pta",
```
jenswi-linaro, line 67: I'm having second thoughts about this PTA. Why is it implemented as a PTA instead of using fast-calls?

etienne-lms: The interrupt controller may be distant, e.g. on a bus, so power up/down and set_wakeup for the interrupt must be done from a tee thread.

jenswi-linaro: This is something that is tied closely to `OPTEE_SMC_SEC_CAP_ITR_NOTIF`, right? I mean if we have `OPTEE_SMC_SEC_CAP_ITR_NOTIF` then we can assume that this service is enabled too?

If so then I believe we could add an `OPTEE_MSG_CMD_SET_ITR_NOTIF` command to replace this PTA. The advantage is that we don't need to keep track of another session and it should also simplify the kernel driver a bit.

etienne-lms: ok, you mean an yielded SMC funcID for itr notif operations? I guess a single SMC should be enough for the 2 supported operations.

jenswi-linaro: Yes

etienne-lms:
keep alive.
the linux part is still in progress...

etienne-lms:
thanks

etienne-lms:
Rebased on master (fixup commits not squashed).
Linux kernel counterpart patches v3 are available on the lkml: see https://lore.kernel.org/lkml/20230315113201.1343781/.

core/pta/itr_notif.c
jenswi-linaro: Are you going to do something about replacing the PTA with an SMC?

etienne-lms:
I've replaced the ITR_NOTIF PTA with SMC functions for interrupt notif activation/wakeup configuration. This change reveverts a commit from the series. I think I should squash the fixup commit...
Linux counterpart patches is [v4 in LKML](https://lore.kernel.org/lkml/20230317171003.974801-2-etienne.carriere@linaro.org/).

jenswi-linaro:
Please squash all the updates into their respective commits.

core/kernel/notif.c
```diff
@@ -307,7 +307,7 @@ TEE_Result notif_itr_register(struct notif_itr *notif)
 {
 	unsigned int itr_num = 0;
 	uint32_t exceptions = 0;
-	const struct notif_itr_ops *ops = NULL;
+	const __maybe_unused struct notif_itr_ops *ops = NULL;
```
jenswi-linaro, line 310: I believe this should go into an earlier commit.

core/pta/tests/misc.c
```diff
@@ -21,6 +25,338 @@
...
+	mutex_lock(&itr_notif_test_lock);
+
+	res = register_test1_itr_notif();
+	if (res)
+		goto out;
+
+	IMSG("Itr-notif test1: check all test interrupt notifs are masked");
+	for (n = 0; n < TEST_ITR_NOTIF_CNT; n++) {
+		if (!notif_itr_is_masked(test_itr_notif[n].itr_num)) {
+			DMSG("ITR notifier %u is not default masked", n);
```
jenswi-linaro, line 133: %zu

core/arch/arm/plat-vexpress/conf.mk
```diff
@@ -120,6 +120,8 @@ $(call force,CFG_DT,y)
 CFG_DTB_MAX_SIZE ?= 0x100000
 CFG_CORE_ASYNC_NOTIF ?= y
 CFG_CORE_ASYNC_NOTIF_GIC_INTID ?= 219
+CFG_CORE_ITR_NOTIF ?= y
```
jenswi-linaro, line 123: Better to test this on QEMU v8.

core/kernel/notif.c
```diff
@@ -106,6 +126,217 @@ void notif_send_async(uint32_t value)
...
+	uint32_t exceptions = 0;
+	bool do_bh = false;
+	size_t cnt = 0;
+	int bit = 0;
+
+	exceptions = cpu_spin_lock_xsave(&notif_itr_lock);
+
+	/* Retrieve at most *itr_num's pending event */
+	for (cnt = 0; cnt < *itr_count; cnt++) {
+		bit = get_pending_itr_num();
```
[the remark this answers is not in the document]
etienne-lms: I'll consider that. I foresee some corner cases, I wonder if it's worth it.

core/kernel/notif.c
```diff
@@ -106,6 +126,217 @@ void notif_send_async(uint32_t value)
...
+		itr_nums[cnt] = bit;
+	}
+
+	/* Report if there are other pending interrupts */
+	if (cnt == *itr_count && get_pending_itr_num() >= 0)
+		cnt++;
+
+	/* Retrieve bottom half event if pending */
+	if (bit_test(notif_values, NOTIF_VALUE_DO_BOTTOM_HALF)) {
+		bit_clear(notif_values, bit);
```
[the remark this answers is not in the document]
etienne-lms: indeed, thanks.

core/kernel/notif.c
```diff
@@ -106,6 +126,217 @@ void notif_send_async(uint32_t value)
...
+	cpu_spin_unlock_xrestore(&notif_itr_lock, exceptions);
+}
+
+static struct notif_itr *find_notif_itr(unsigned int itr_num)
+{
+	assert(itr_num <= NOTIF_ITR_VALUE_MAX);
+	return notif_itr_hdl[itr_num];
+}
+
+static struct notif_itr *find_ree_notif_itr(unsigned int itr_num)
```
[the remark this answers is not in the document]
etienne-lms: yes. thanks.

core/kernel/notif.c
```diff
@@ -241,6 +241,22 @@ void notif_itr_set_mask(unsigned int itr_num, bool do_mask)
 	}
 }
 
+bool notif_itr_is_masked(unsigned int itr_num)
+{
+	if (!find_notif_itr(itr_num))
+		panic("Unsafe itr number value");
```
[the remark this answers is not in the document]
etienne-lms: rephrased to `Invalid interrupt number`

core/kernel/notif.c
etienne-lms: right, i'll need to fix that when squashing fixup commits.

core/arch/arm/plat-vexpress/conf.mk
etienne-lms: ok, i'll make this be tested on both qemu_virt and qemu_armv8a.

etienne-lms (commented):
Thanks for the review comments.

etienne-lms:
keep alive

etienne-lms:
Squashed fixup commits
+ added/removed  few `__maybe_unused` attributes in commits
+ split commit on plat-vexpress between qemu_virt and qemuv8a flavors

Note this is still related to a change in Linux kernel (see https://github.com/linaro-swg/linux/pull/112)  and https://github.com/OP-TEE/optee_test/pull/663)https://github.com/OP-TEE/optee_test/pull/663.

jenswi-linaro:
For "libutils: add bit_ffs_from()" please apply:

How about submitting "libutils: add bit_ffs_from()" and "core: notif: fix input comment typo" in a separate PR to get them out of the way?

etienne-lms:
Ok, I've created https://github.com/OP-TEE/optee_os/pull/6009.

core/include/kernel/notif.h
```diff
@@ -89,6 +102,30 @@ struct notif_driver {
...
+	unsigned int itr_num;
+	const struct notif_itr_ops *ops;
+};
+
+/*
+ * struct notif_itr_ops - Interrupt notifier operations
+ *
+ * @set_mask Fasctcall callback for (un)masking the event or NULL if not used
+ * @set_state Callback for enabling/disabling the event or NULL if not used
+ * @set_wakeup Callback for enabling/disabling standby wakeup source or NULL
```
jenswi-linaro, line 121: Do you have an example of what this callback should do and when it's called?

etienne-lms: It is aimed to be called by non-secure world from `enable_irq_wake()` and `disable_irq_wake()` in Linux kernel (see https://elixir.bootlin.com/linux/v6.3/source/include/linux/interrupt.h#L482) or like. Such a call is relayed to OP-TEE core through [optee_itr_notif_set_wake() handler](https://github.com/linaro-swg/linux/pull/112/commits/9a724e5484e5f5440547a46afca119e433969f14#diff-e496739c0d15953f39d2accb0bf7aa442921cdb6b76b91847e4d850d0c27d371R1016).

core/include/kernel/notif.h
```diff
@@ -89,6 +102,30 @@ struct notif_driver {
...
+struct notif_itr {
+	unsigned int itr_num;
+	const struct notif_itr_ops *ops;
+};
+
+/*
+ * struct notif_itr_ops - Interrupt notifier operations
+ *
+ * @set_mask Fasctcall callback for (un)masking the event or NULL if not used
+ * @set_state Callback for enabling/disabling the event or NULL if not used
```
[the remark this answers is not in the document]
etienne-lms: This is intended to enable or disable the interrupt when the Linux driver consuming the interrupt need to.

core/include/kernel/notif.h
```diff
@@ -152,6 +189,118 @@ static inline uint32_t notif_get_value(bool *value_valid, bool *value_pending)
...
+ */
+void notif_itr_raise_event(struct notif_itr *notif);
+
+/*
+ * Mask/unmask an interrupt notification
+ *
+ * @itr_num Interrupt identifier provided by normal world
+ * @do_mask True to mask the event, false to unmask the event
+ *
+ * This function is called from a fastcall context
```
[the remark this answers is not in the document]
etienne-lms: Only `notif_itr_set_mask()` and `notif_get_pending()` (below) are called from a fastcall SMC  function ID execution context. Other functions are call from thread contexts.

core/include/kernel/notif.h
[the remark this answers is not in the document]
jenswi-linaro: Yes, so from an OP-TEE point of view, what is the difference? What will be done differently inside OP-TEE?

core/include/kernel/notif.h
etienne-lms: `struct notif_itr_ops` is intended to 'interrupt notification', a service exposed to non-secure world. 
For wake up source handled inside OP-TEE, OP-TEE drivers should use OP-TEE internal means (for example we could add a `.set_wakeup` handler field in `struct itr_chip`, once https://github.com/OP-TEE/optee_os/pull/5954 is reviewed), not this ops.

core/include/kernel/notif.h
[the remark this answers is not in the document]
etienne-lms: `struct notif_itr_ops` is intended to 'interrupt notification', a service exposed to non-secure world.

For wake up source handled inside OP-TEE, OP-TEE drivers should use OP-TEE internal means that are `interrupt_mask()`, `interrupt_unmask()`, `interrupt_enable()` and `interrupt_disable()` (API functions based on https://github.com/OP-TEE/optee_os/pull/5954 is reviewed). We could also add a `.set_wakeup` handler field in `struct itr_chip` for wakeup source support, if needed.

etienne-lms:
I've created https://github.com/OP-TEE/optee_docs/pull/198 to describe the feature.

etienne-lms:
keep alive

PR #5794, opened 2023-01-25
Introduce binary format SP support

imre-kis-arm:
This pull request introduces support for an SPMC agnostic SP format. In order to make it easier to implement support for this format, a simple flat binary format was chosen so it won't require an ELF loader. The SPMC only has to copy the SP binary into an RWX memory area then it can jump to its entry point. The access rights of different sections of the SP can be set by either via the load relative memory region definitions of the SP manifest or by invoking the `FFA_MEM_PERM_SET` FF-A interface from the SP during the boot phase. Setting up the stack, clearing the .bss section, handling relocations, etc. are all handled by the SP startup code.
The ELF format SPs remain supported by OP-TEE. The SP loader can distinguish between the two formats by the `legacy-elf-format` field of the SP manifest.
trusted-services already provides SP deployments in this new format (config: default-sp).
This PR requires the patches of https://github.com/OP-TEE/build/pull/613 in the build repository.
For more detailed description see the [PR in optee_docs](https://github.com/OP-TEE/optee_docs/pull/181).

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -46,6 +46,8 @@
 					 SP_MANIFEST_ATTR_WRITE | \
 					 SP_MANIFEST_ATTR_EXEC)
 
+#define SP_MANIFEST_FLAG_NOBITS	BIT(0)
+
 #define SP_PKG_HEADER_MAGIC (0x474b5053)
 #define SP_PKG_HEADER_VERSION_V1 (0x1)
 #define SP_PKG_HEADER_VERSION_V2 (0x2)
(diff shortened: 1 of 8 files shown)


imre-kis-arm:
The checkpatch warning is caused by a longer error message. Unless it's not absolutely necessary I wouldn't split it into multiple lines, as it prevents finding the line of the error via grep-ing the error message.

jenswi-linaro:
Do you have any more documentation on the "load-address-relative-offset" property? I'm interested in the rationale behind this.

imre-kis-arm:
[Arm Firmware Framework for Arm A-profile v1.1](https://developer.arm.com/documentation/den0077/latest) adds "Load address relative offset" field as an alternative to the Base address field in the SP manifest's memory regions (Table 5.2). This way the SPMC can configure the memory regions (segments) of position independent SPs in cases where the SPMC chooses the load address.
For example this is manifest of the Protected Storage SP of trusted-services. The region list is generated based on the segments of the original ELF file.

```
/dts-v1/;

/ {
	compatible = "arm,ffa-manifest-1.0";
	ffa-version = <0x10000>;
	uuid = <0x1f81b75 0x6847de3d 0x100f14a5 0x9017edae>;
	description = "PS";
	execution-ctx-count = <0x01>;
	exception-level = <0x01>;
	execution-state = <0x00>;
	xlat-granule = <0x00>;
	messaging-method = <0x03>;

	memory-regions {
		compatible = "arm,ffa-manifest-memory-regions";

		region_0 {
			load-address-relative-offset = <0x00 0x00>;
			pages-count = <0x11>;
			attributes = <0x05>;
			load-flags = <0x00>;
		};

		region_11000 {
			load-address-relative-offset = <0x00 0x11000>;
			pages-count = <0x01>;
			attributes = <0x03>;
			load-flags = <0x00>;
		};

		region_12000 {
			load-address-relative-offset = <0x00 0x12000>;
			pages-count = <0x20>;
			attributes = <0x03>;
			load-flags = <0x01>;
		};
	};
};
```

Normally the segment details would be packed into the ELF file, but as now we use a flat binary, this metadata is now represented as load address relative memory regions in the SP manifest.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -422,6 +379,196 @@ static TEE_Result sp_dt_get_uuid(const void *fdt, int node,
...
+	struct mobj *mobj = NULL;
+	uaddr_t base_addr = 0;
+	uint32_t vm_flags = 0;
+	unsigned int idx = 0;
+	vaddr_t va = 0;
+
+	if (!s || !uctx)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	DMSG("Loading generic format SP %pUl", &uctx->ts_ctx->uuid);
```
jenswi-linaro, line 430: s/generic/raw binary/

core/arch/arm/kernel/secure_partition.c
```diff
@@ -422,6 +379,196 @@ static TEE_Result sp_dt_get_uuid(const void *fdt, int node,
 	return TEE_SUCCESS;
 }
 
+static TEE_Result sp_is_elf_format(const void *fdt, int sp_node,
+				   bool *is_elf_format)
+{
+	TEE_Result res = TEE_SUCCESS;
+	uint32_t elf_format = 0;
+
+	res = sp_dt_get_u32(fdt, sp_node, "legacy-elf-format", &elf_format);
```
jenswi-linaro, line 388: Why not just "elf-format"?

core/arch/arm/kernel/secure_partition.c
```diff
@@ -422,6 +379,196 @@ static TEE_Result sp_dt_get_uuid(const void *fdt, int node,
...
+	ts_push_current_session(&s->ts_sess);
+
+	res = sp_is_elf_format(fdt, 0, &is_elf_format);
+	if (res == TEE_SUCCESS) {
+		if (is_elf_format) {
+			/* Load the SP using ldelf. */
+			ldelf_load_ldelf(&ctx->uctx);
+			res = ldelf_init_with_ldelf(&s->ts_sess, &ctx->uctx);
+		} else {
+			/* Generic binary format SP */
```
jenswi-linaro, line 550: s/Generic/Raw/

jenswi-linaro (commented):
As few comments, with those addressed please apply:

The reason we switched from a raw binary format for TAs was that we had to
reinvent the wheel again and again so at some point it was easier to just
switch to ELF and get almost everything for free. Here you seem confident
that you can match everything you'd ever need from ELF with a raw custom
binary format. I don't think you should call this a generic binary format,
it's not so generic compared to ELF.

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done, also changed the commit message to follow the same naming convention.

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: The main reason was to emphasize that the ELF format is not the preferable one anymore. I you would prefer "elf-format" I can change it.

jenswi-linaro: I would prefer "elf-format" as that's more neutral. But that's just my opinion if it's important to emphasize that this is less desired even in the name of the variable keep it as it is.

imre-kis-arm: Changed to "elf-format".

imre-kis-arm:
> Famous last words. :-)

I know, I know... :)

jforissier:
@imre-kis-arm please rebase onto master and get rid of the `Change-Id:` tag, thanks.

imre-kis-arm:
> @imre-kis-arm please rebase onto master and get rid of the `Change-Id:` tag, thanks.

Done and fixed.

Please note that after rebasing onto master I was only able to test it on an FVP platform if I temporarily reverted f1f431c7a92671b4fa397976d381cc5ad8adacc4. This issue will be addressed in a separate PR.

PR #5795, opened 2023-01-25
dts: stm32: clean device tree for STM32MP13

GseoC:
Unused peripherals shouldn't be supported at board level for STM32MP13.

### DIFF
--- a/core/arch/arm/dts/stm32mp13-pinctrl.dtsi
+++ b/core/arch/arm/dts/stm32mp13-pinctrl.dtsi
@@ -43,17 +43,4 @@
 			bias-pull-up;
 		};
 	};
-
-	uart8_pins_a: uart8-0 {
-		pins1 {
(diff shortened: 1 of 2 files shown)


core/arch/arm/dts/stm32mp13-pinctrl.dtsi
```diff
@@ -44,16 +44,4 @@
 		};
 	};
```
etienne-lms, line 46: remove also the above empty line

etienne-lms (commented):

core/arch/arm/dts/stm32mp13-pinctrl.dtsi
GseoC: Yep, missed that

GseoC:
Updated and tag applied

etienne-lms (approved):

PR #5799, opened 2023-01-26
versal: enable the crypto driver

ldts:
The crypto driver API provides an extra indirection level to enable different ciphers.

Since Versal ACAP supports acipher and authenc, enable them.

### DIFF
--- a/core/drivers/crypto/versal/crypto.mk
+++ b/core/drivers/crypto/versal/crypto.mk
@@ -1,0 +1,11 @@
+ifeq ($(CFG_VERSAL_CRYPTO_DRIVER),y)
+# Enable the crypto driver
+$(call force,CFG_CRYPTO_DRIVER,y)
+
+CFG_CRYPTO_DRIVER_DEBUG ?= 0
+$(call force,CFG_CRYPTO_DRV_ACIPHER,y)
+$(call force,CFG_CRYPTO_DRV_ECC,y)
+$(call force,CFG_CRYPTO_DRV_RSA,y)
+$(call force,CFG_CRYPTO_DRV_AUTHENC,y)
+
+endif
--- a/core/drivers/crypto/versal/ecc.c
+++ b/core/drivers/crypto/versal/ecc.c
@@ -99,13 +99,18 @@ static TEE_Result ecc_get_key_size(uint32_t curve, size_t *bytes, size_t *bits)
 	return TEE_SUCCESS;
 }
 
+static void memcpy_swp(uint8_t *to, const uint8_t *from, size_t len)
+{
+	size_t i = 0;
+
+	for (i = 0; i < len; i++)
+		to[i] = from[len - 1 - i];
+}
+
 static void crypto_bignum_bn2bin_eswap(uint32_t curve,
 				       struct bignum *from, uint8_t *to)
 {
 	uint8_t pad[66] = { 0 };
-	uint8_t tmp = 0;
-	size_t i = 0;
-	size_t j = 0;
 	size_t len = crypto_bignum_num_bytes(from);
 	size_t bytes = 0;
 	size_t bits = 0;
@@ -114,52 +119,28 @@ static void crypto_bignum_bn2bin_eswap(uint32_t curve,
 		panic();
 
 	crypto_bignum_bn2bin(from, pad + bytes - len);
-	for (i = 0, j = bytes - 1; i < j; i++, j--) {
-		tmp = pad[i];
-		pad[i] = pad[j];
-		pad[j] = tmp;
-	}
-	memcpy(to, pad, bytes);
+	memcpy_swp(to, pad, bytes);
 }
 
 static TEE_Result ecc_prepare_msg(uint32_t algo, const uint8_t *msg,
 				  size_t msg_len, struct versal_mbox_mem *p)
 {
-	uint8_t buf[TEE_SHA512_HASH_SIZE + 2] = { 0 };
-	TEE_Result ret = TEE_SUCCESS;
+	uint8_t swp[TEE_SHA512_HASH_SIZE + 2] = { 0 };
 	size_t len = 0;
 
-	switch (algo) {
-	case TEE_ALG_ECDSA_P384:
+	if (msg_len > TEE_SHA512_HASH_SIZE + 2)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (algo == TEE_ALG_ECDSA_P384)
 		len = TEE_SHA384_HASH_SIZE;
-
-		if (msg_len == len)
-			return versal_mbox_alloc(len, msg, p);
-
-		ret = tee_hash_createdigest(TEE_ALG_SHA384, msg, msg_len,
-					    buf, sizeof(buf));
-		if (ret)
-			return ret;
-		break;
-	case TEE_ALG_ECDSA_P521:
+	else if (algo == TEE_ALG_ECDSA_P521)
 		len = TEE_SHA512_HASH_SIZE + 2;
+	else
+		return TEE_ERROR_NOT_SUPPORTED;
 
-		if (msg_len == len)
-			return versal_mbox_alloc(len, msg, p);
-
-		ret = tee_hash_createdigest(TEE_ALG_SHA512, msg, msg_len, buf,
-					    sizeof(buf));
-		if (ret)
-			return ret;
-		break;
-	default:
-		return TEE_ERROR_GENERIC;
-	}
-
-	/* provide the hash */
-	versal_mbox_alloc(len, buf, p);
-
-	return TEE_SUCCESS;
+	/* Swap the hash/message and pad if necessary */
+	memcpy_swp(swp, msg, msg_len);
+	return versal_mbox_alloc(len, swp, p);
 }
 
 static TEE_Result verify(uint32_t algo, struct ecc_public_key *key,
@@ -176,6 +157,9 @@ static TEE_Result verify(uint32_t algo, struct ecc_public_key *key,
 	uint32_t err = 0;
 	size_t bytes = 0;
 	size_t bits = 0;
+
+	if (sig_len % 2)
+		return TEE_ERROR_SIGNATURE_INVALID;
 
 	ret = ecc_get_key_size(key->curve, &bytes, &bits);
 	if (ret != TEE_SUCCESS) {
@@ -206,8 +190,11 @@ static TEE_Result verify(uint32_t algo, struct ecc_public_key *key,
 	}
 	memset(&arg, 0, sizeof(arg));
 
-	/* Verify the message with the validated key */
-	versal_mbox_alloc(sig_len, sig, &s);
+	versal_mbox_alloc(sig_len, NULL, &s);
+	/* Swap the {R,S} components */
+	memcpy_swp(s.buf, sig, sig_len / 2);
+	memcpy_swp((uint8_t *)s.buf + sig_len / 2, sig + sig_len / 2,
+		   sig_len / 2);
 	versal_mbox_alloc(sizeof(*cmd), NULL, &cmd_buf);
 
 	cmd = cmd_buf.buf;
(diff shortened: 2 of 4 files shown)


ldts:
@jenswi-linaro the first thing it happens when enabling the crypto driver is this.

* regression_1004 Test User Crypt TA                                                                                                                                                                                                                                                                                                                                                                                                                                                  
E/TC:? 0 assertion 'f->arg.hash == (f->my_hash ^ f->called_hash)' failed at lib/libutils/ext/include/fault_mitigation.h:298 <__ftmn_copy_linked_call_res>                                                                                                                                                                                                                                                                                                                             
E/TC:1 0 Panic at core/kernel/assert.c:28 <_assert_break>                                                                                                                                                                                                                                                                                                                                                                                                                             
E/TC:1 0 TEE load address @ 0x60000000                                                                                                                                                                                                                                                                                                                                                                                                                                                
E/TC:1 0 Call stack:                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
E/TC:1 0  0x600086d4                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
E/TC:1 0  0x6001b1d8                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
E/TC:1 0  0x60019f88                  


Just `CFG_FAULT_MITIGATION=n `seems to be needed to pass these tests.

core/drivers/crypto/versal/crypto.mk
```diff
@@ -0,0 +1,10 @@
+ifeq ($(CFG_VERSAL_CRYPTO_DRIVER),y)
+# Enable the crypto driver
+$(call force,CFG_CRYPTO_DRIVER,y)
+
+CFG_CRYPTO_DRIVER_DEBUG ?= 0
+$(call force,CFG_CRYPTO_DRV_ACIPHER,y)
+$(call force,CFG_CRYPTO_DRV_ECC,y)
+$(call force,CFG_CRYPTO_DRV_RSA,y)
+$(call force,CFG_CRYPTO_DRV_AUTHENC,y)
```
jforissier, line 9: Why `force` and not `?=`

jforissier (commented):
One comment on "versal: enable the crypto driver", otherwise for both commits:

core/drivers/crypto/versal/crypto.mk
ldts: So the way I thought about this is that when Versal sets CFG_VERSAL_CRYPTO_DRIVER=y, the three drivers are mandatory: ie versal ecc, versal rsa and versal auth-enc. 

Those however can only be enabled if the crypto API is also configured to allow calling them (hence the forcing)

etienne-lms (commented):

I suggest you swap the 2 commits order to provision enough heap before your platform enable these crypto functions.

etienne-lms:
It will not be easy for the fault migitation sequence to verify the execution path if it doesn't know which exec path is to be used. Yet, I think this is out of scope of this P-R.

ldts:
Validating with openssl I noticed a problem with the signatures which wasn't caught with xtests alone. I added the fix to this patchset.

core/drivers/crypto/versal/ecc.c
```diff
@@ -99,13 +100,30 @@ static TEE_Result ecc_get_key_size(uint32_t curve, size_t *bytes, size_t *bits)
...
+static void eswap(size_t len, const uint8_t *from, uint8_t *to)
 {
-	uint8_t pad[66] = { 0 };
+	uint8_t *pad = calloc(1, len);
 	uint8_t tmp = 0;
 	size_t i = 0;
 	size_t j = 0;
+
+	if (!pad)
+		panic();
```
etienne-lms, line 111: It seems overkilling to panic here. Should better return a error code.

That said, you don't need a allocation here, simply loop on `to[i++] = from[j--]` or like.

core/drivers/crypto/versal/ecc.c
```diff
@@ -207,7 +229,15 @@ static TEE_Result verify(uint32_t algo, struct ecc_public_key *key,
 	memset(&arg, 0, sizeof(arg));
 
 	/* Verify the message with the validated key */
-	versal_mbox_alloc(sig_len, sig, &s);
+	swp = calloc(1, sig_len);
+	if (!swp) {
+		ret = TEE_ERROR_OUT_OF_MEMORY;
+		goto out;
+	}
+	eswap(sig_len / 2, sig, swp);
```
etienne-lms, line 237: Check at function entry that `sig_len` is even?

core/drivers/crypto/versal/crypto.mk
```diff
@@ -0,0 +1,14 @@
...
+
+# Disable Fault Mitigation: currently affected by
+# the need to fallback to software operations .
+CFG_FAULT_MITIGATION = n
```
etienne-lms, line 13: `CFG_FAULT_MITIGATION ?= n`
or 
`$(call force,CFG_FAULT_MITIGATION,n)`

etienne-lms (commented):
For commit "crypto: versal: ecc sign/verify fix",  header line could be more explicit, e.g.:
`crypto: versal: hash and signature byte swap in ecc sign/verify`.

 For that commit, the issue was nout found using `xtest`, even with `xtest -l 15` (which tests more vectors than default `xtest -l 0`)?

core/drivers/crypto/versal/crypto.mk
ldts: I had tried both options:

1) the first one will not work as crypto.mk runs last and someone else enabled CFG_FAULT_MITIGATION (so it wont force it)

2) returned the following error probably for the same reason
 core/drivers/crypto/versal/crypto.mk:13: *** CFG_FAULT_MITIGATION is set to 'y' (from file) but its value must be 'n'.  Stop.

the simpler choice I see, since this make runs last, is to do as I did.

core/drivers/crypto/versal/ecc.c
ldts: ECDSA signatures are 2 times the key length but I'll check anyway,

core/drivers/crypto/versal/crypto.mk
jforissier: Set `CFG_FAULT_MITIGATION ?= n` in `core/arch/arm/plat-versal/conf.mk`.

core/drivers/crypto/versal/ecc.c
ldts: yeah I can do that, just need to remove the case where I used the the same buffer in from/to fields. let me do that.

core/drivers/crypto/versal/crypto.mk
ldts: right...but since the issue is related to crypto I thought it would make sense to set it there.

why is doing that a problem? what is wrong with setting it to n in crypto.mk?

ldts:
btw while working on this I also validated RSA with openssl and that is good - so just ECC was sign/verify was impacted by the need to swap the input.

core/drivers/crypto/versal/crypto.mk
ldts: btw I dont mind doing what you propose @jforissier, just would like to understand the reason behind it.

jforissier: Sure, see: https://github.com/OP-TEE/optee_docs/pull/182/files

ldts:
@jforissier @etienne-lms please let me know if I can add your tags so the crypto driver can be enabled by default

core/drivers/crypto/versal/crypto.mk
ldts: ahh, yes, passing CFG_ through the environment or build line..ok thanks!

jforissier (commented):
For "crypto: versal: ecc sign/verify fix":
Two comments below.

jforissier (commented):
For "versal: increase CFG_CORE_HEAP_SIZE":

jforissier (commented):
For "versal: enable the crypto driver":

core/drivers/crypto/versal/ecc.c
```diff
@@ -114,40 +119,42 @@ static void crypto_bignum_bn2bin_eswap(uint32_t curve,
...
 					    buf, sizeof(buf));
 		if (ret)
 			return ret;
 		break;
 	case TEE_ALG_ECDSA_P521:
 		len = TEE_SHA512_HASH_SIZE + 2;
 
-		if (msg_len == len)
-			return versal_mbox_alloc(len, msg, p);
+		if (msg_len == len || msg_len == TEE_SHA512_HASH_SIZE) {
```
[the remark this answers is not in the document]
ldts: the IP needs the length of the hash to be 66 bytes but the hash size is 64 bytes. 
For for 64 bytes the driver pads with zeros.

Accepting 66 is to cover the case where the crypto API might be padding the hash to this length (maybe in the future?). But the driver assumes that the msg is an actual hash.

 Any other values are not being accepted as hashes but interpreted as a byte array.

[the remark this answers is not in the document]
jforissier: If by "crypto API" you mean the one that is internal to OP-TEE then I suggest making things as simple as possible and decide that `msg_len` is just... the message length and handle the +2 only locally (and with a comment).

Besides... what if we want to sign a message of size 64 (or 66) bytes?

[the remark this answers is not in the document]
ldts: yes. ok let me simplify just making sure xtests and local tests pass ignore anything else.

ldts:
the last update continues to pass xtest -l4 -t 4006 and openssl validation

jforissier:
So the `tee_hash_createdigest()` calls that you removed were effectively not used? That said this LGTM.

For commit "crypto: versal: ecc sign/verify fix":

jforissier:
Please fix the checkpatch issues too.

ldts:
> So the `tee_hash_createdigest()` calls that you removed were effectively not used? That said this LGTM.

yes, it was not being used. some sort of fallback

> 
> For commit "crypto: versal: ecc sign/verify fix":
> 

ldts:
@jforissier thanks.

Just added one last fix - to RSA: sign/verify is clearly supported but encrypt/decrypt is not - so added a patch to disable it.

jforissier:
> Just added one last fix - to RSA: sign/verify is clearly supported but encrypt/decrypt is not - so added a patch to disable it.

You mean "crypto: versal: rsa: only support sign/verify operations"? How does it disable anything?

jforissier:
You left a checkpatch warning. I don't see why we should make an exception here. `keypair` is a term that we agreed to accept in the source code, but it doesn't make it more correct in English so chekpatch does have a point in claiming you should write `printf "OP-TEE: create key pair"`.

ldts:
> > Just added one last fix - to RSA: sign/verify is clearly supported but encrypt/decrypt is not - so added a patch to disable it.
> 
> You mean "crypto: versal: rsa: only support sign/verify operations"? How does it disable anything?

sign/verify is done via returning TEE_ERROR_NOT_IMPLEMENTED from do_ssa_verify/do_ssa_sign: when that happens, 
the internal crypto API does the pkcs padding (in the driver's behalf, which is quite nice) and then calls the driver again via encrypt/decrypt.

this patch causes encrypt/decrypt to be routed to software operations except in the case where the algorithm is DRVCRYPT_RSASSA_PKCS_V1_5 (the only one supported by the PLM firmware).

core/drivers/crypto/versal/crypto.mk
etienne-lms: In this _crypto.mk_ file you could still check the value:
```makefile
$(call force,CFG_FAULT_MITIGATION,n,Required for Versal support)
```
 but as @jforissier said, platform config shall enforce the expected value knowing they enable the hw device.

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for the series

ldts:
thanks @etienne-lms, your tags applied

PR #5810, opened 2023-02-03
ci: qemuv8: enable PKCS#11

jforissier:
Enable the PKCS#11 tests in the QEMUv8 jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,6 +265,8 @@ jobs:
           export LC_ALL=C
           export BR2_CCACHE_DIR=/github/home/.cache/ccache
           export CFG_TEE_CORE_LOG_LEVEL=0
+          export CFG_PKCS11_TA=y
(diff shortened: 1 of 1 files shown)


jforissier:
On second thought, perhaps we could enable `CFG_PKCS11_TA` by default in `mk/config.mk`, or in `core/arm/plat-vexpress` if that's too much. But then we would also need to build the in-tree TAs for both `ta_arm32` and `ta_arm64`, otherwise there could be a mismatch with the test environment. For example `qemu_v8.mk` sets `COMPILE_S_USER := 64` ([link](https://github.com/OP-TEE/build/blob/3.20.0/qemu_v8.mk#L8)), thus expecting 64-bit TAs, but the default target for in-tree TAs is `ta_arm32`.

jforissier:
Closing this in favor of: https://github.com/OP-TEE/optee_os/pull/5812

PR #5812, opened 2023-02-06
plat-vexpress: enable PKCS#11 TA by default

jforissier:
Configure `plat-vexpress` to build the PKCS#11 TA by default and select the default TA target of `vexpress-qemu_armv8a` to be `ta_arm64`.


jenswi-linaro:

PR #5815, opened 2023-02-07
libutee: add TEE_MAIN_ALGO_SHAKE values

clementfaure:
The CAAM driver relies on TEE_ALG_GET_MAIN_ALG() macro to retrieve the main algorithm ID from the TEE_ALG_* value.

With the addition of TEE_ALG_SHAKE128 and TEE_ALG_SHAKE256, TEE_ALG_GET_MAIN_ALG() would return 0x01 (TEE_MAIN_ALGO_MD5) and 0x02 (TEE_MAIN_ALGO_SHA1). These returned values are wrong.

Add TEE_MAIN_ALGO_SHAKE128 and TEE_MAIN_ALGO_SHAKE256 values for respectively TEE_ALG_SHAKE128 and TEE_ALG_SHAKE256.

@jforissier I'm not sure about the `TEE_MAIN_ALGO_SHAKE` values tho ...

### DIFF
--- a/lib/libutee/include/utee_defines.h
+++ b/lib/libutee/include/utee_defines.h
@@ -44,6 +44,8 @@
 #define TEE_MAIN_ALGO_CONCAT_KDF 0xC1 /* OP-TEE extension */
 #define TEE_MAIN_ALGO_PBKDF2     0xC2 /* OP-TEE extension */
 #define TEE_MAIN_ALGO_X25519     0x44 /* Not in v1.2 spec */
+#define TEE_MAIN_ALGO_SHAKE128   0xC3 /* OP-TEE extension */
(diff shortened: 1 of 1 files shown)


jforissier:
> @jforissier I'm not sure about the `TEE_MAIN_ALGO_SHAKE` values tho ...

Should be OK. The concept of "main algo" doesn't exist anymore since the GP v1.2 spec so we can use whichever values work for our use case...

clementfaure:
Got it! Thanks!

jenswi-linaro:
> Should be OK. The concept of "main algo" doesn't exist anymore since the GP v1.2 spec so we can use whichever values work for our use case...

Since we're still using the concept perhaps we should add a comment describing how we're using it in OP-TEE and which values to use etc. Not in scope of this PR.

Anyway:

PR #5817, opened 2023-02-09
core: tpm: fix syntax in trace message

etienne-lms:
Fixes build warning (trace message below) when CFG_CORE_TPM_EVENT_LOG=y.


jenswi-linaro:

etienne-lms:
tag applied. thanks.

PR #5822, opened 2023-02-10
Add TZASC support for imx8mscale

clementfaure:


core/arch/arm/plat-imx/drivers/tzc380.c
```diff
@@ -25,7 +26,23 @@
 register_phys_mem(MEM_AREA_IO_SEC, TZASC2_BASE, TZASC_SIZE);
 #endif
 
-register_phys_mem(MEM_AREA_IO_SEC, TZASC_BASE, TZASC_SIZE);
```
jforissier, line 28: I'm confused. Doesn't this temporarily break things until the next commit? Why does this have to be removed and added back later?

jforissier (commented):
- For the first two commits, one question below 
- For commit: "drivers: imx: tzc380: dump TZASC state after lockdown"
What it's really doing is "drivers: imx: tzc380: **do not** dump TZASC state **before** lockdown". With that:
- For commit "drivers: imx: tzc380: re-configure TZ380 upon PM resume":

jforissier:

jforissier:
Please rebase onto master while you're at it. It should fix the IBART error.

jforissier (approved):

PR #5829, opened 2023-02-16
core: ffa: add TOS_FW_CONFIG handling

balint-dobszay-arm:
At boot TF-A passes two DT addresses (HW_CONFIG and TOS_FW_CONFIG), but currently only the HW_CONFIG address is saved, the other one is dropped. This commit adds functionality to save the TOS_FW_CONFIG too, so we can retrieve it later. This is necessary for the CFG_CORE_SEL1_SPMC use case, because the SPMC manifest is passed in this DT.

### DIFF
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
  * Copyright (c) 2015-2022, Linaro Limited
- * Copyright (c) 2021, Arm Limited
+ * Copyright (c) 2021-2023, Arm Limited
  */
 
 #include <platform_config.h>
@@ -320,7 +320,11 @@ clear_nex_bss:
 	bl	core_mmu_set_default_prtn_tbl
 #endif
 
+#ifdef CFG_CORE_SEL1_SPMC
+	mov	x0, xzr		/* pager not used */
+#else
 	mov	x0, x19		/* pagable part address */
+#endif
 	mov	x1, #-1
 	bl	boot_init_primary_early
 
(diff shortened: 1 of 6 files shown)


core/arch/arm/kernel/entry_a64.S
```diff
@@ -338,6 +338,7 @@ clear_nex_bss:
 	str	wzr, [x22, #THREAD_CORE_LOCAL_FLAGS]
 #endif
 	mov	x0, x20		/* DT address */
+	mov	x1, x19		/* TOS_FW_CONFIG DT address */
```
jenswi-linaro, line 341: If not `CFG_CORE_SEL1_SPMC` then we should supply zeroes in `x1` and vice versa in `x0` above where `boot_init_primary_early()` is called with pageable part address.
`boot_init_primary_late()` is called form `entry_a32.S` too, there we must make sure to supply zeroes in `r1`.

core/arch/arm/kernel/entry_a64.S
```diff
@@ -338,6 +338,7 @@ clear_nex_bss:
 	str	wzr, [x22, #THREAD_CORE_LOCAL_FLAGS]
 #endif
 	mov	x0, x20		/* DT address */
```
jenswi-linaro, line 340: `/* DT address also known as HW_CONFIG */`

core/arch/arm/kernel/entry_a64.S
balint-dobszay-arm: Fixed

core/arch/arm/kernel/entry_a64.S
balint-dobszay-arm: Fixed

jenswi-linaro (commented):

PR #5830, opened 2023-02-16
core: arm: link.mk: produce tee-raw.bin by default

jenswi-linaro:
Adds tee-raw.bin as a dependency of all. This produces a tee-raw.bin for all platforms when building. tee-raw.bin is more useful than for instance tee-pager_v2.bin which often is used when a raw binary is needed. Platforms with a link.mk only to produce tee-raw.bin have their link.mk removed since the generic version suffices now.


clementfaure:

jforissier (commented):

apalos:
Acked-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>

jenswi-linaro:
Tags applied

PR #5833, opened 2023-02-20
[RFC] Symbol versioning in libutee.so, the easy part

jforissier:
This RFC/Draft pull request applies versions `GP_1.1` and `GP_1.3.1` to GP API symbols when `libutee` is built as a shared object (dynamic library). It cannot work as is, because the dynamic loader (ldelf) lacks support for versioned symbols, see commit description for "[WIP] libutee: add symbol versioning for shared library". Such support has to be introduced first.

If anyone feels like working on the dynamic loader part, please feel free to do it and re-use these commits.

More details in https://docs.google.com/document/d/1qvN4WIIcmk3MGWdpdB3-hpg3oOErI7tgYbfGx9VR6FA.


mk/lib.mk
```diff
@@ -62,13 +63,17 @@ ifeq ($(CFG_ULIBS_SHARED),y)
 ifeq ($(sm)-$(CFG_TA_BTI),ta_arm64-y)
 lib-ldflags$(lib-shlibfile) += $$(call ld-option,-z force-bti) --fatal-warnings
 endif
-$(lib-shlibfile): $(objs) $(lib-needed-so-files)
+ifneq (,$(wildcard $(lib-verscript)))
```
jenswi-linaro, line 66: I'm not so sure about the `wildcard`, that could hide an invalid value in `lib-verscript`.

jenswi-linaro:
"build: mk/lib.mk: use full path to shared library in flags variable" would make sense upstream by itself.

jforissier:
> "build: mk/lib.mk: use full path to shared library in flags variable" would make sense upstream by itself.

Indeed. https://github.com/OP-TEE/optee_os/pull/5834.

PR #5834, opened 2023-02-20
build: mk/lib.mk: use full path to shared library in flags variable

jforissier:
To add link flags for a shared library, a makefile variable is used that is called lib-ldflags$(libuuid). That's incorrect because the UUID is not enough to uniquely identify a shared library in the build. For example when both 32-bit and 64-bit user space is generated there are two versions of the shared library with the same UUID. It is not a problem at the moment because lib-ldflags$(libuuid) is used only for one target: ta_arm64, but fix this anyways so that the variable may be used for more complex cases.

### DIFF
--- a/mk/lib.mk
+++ b/mk/lib.mk
@@ -60,14 +60,14 @@ $(lib-libfile): $(objs)
 endif
 ifeq ($(CFG_ULIBS_SHARED),y)
 ifeq ($(sm)-$(CFG_TA_BTI),ta_arm64-y)
-lib-ldflags$(libuuid) += $$(call ld-option,-z force-bti) --fatal-warnings
+lib-ldflags$(lib-shlibfile) += $$(call ld-option,-z force-bti) --fatal-warnings
 endif
 $(lib-shlibfile): $(objs) $(lib-needed-so-files)
 	@$(cmd-echo-silent) '  LD      $$@'
 	@mkdir -p $$(dir $$@)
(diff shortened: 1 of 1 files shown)


mk/lib.mk
```diff
@@ -60,14 +60,14 @@ $(lib-libfile): $(objs)
 endif
 ifeq ($(CFG_ULIBS_SHARED),y)
 ifeq ($(sm)-$(CFG_TA_BTI),ta_arm64-y)
-lib-ldflags$(libuuid) += $$(call ld-option,-z force-bti) --fatal-warnings
+lib-ldflags$(lib-shlibfile) += $$(call ld-option,-z force-bti) --fatal-warnings
```
jenswi-linaro, line 63: ` lib-shlibfile` is empty when `CFG_ULIBS_SHARED` isn't `y`

jenswi-linaro (commented):
Can drop the `build: ` prefix from the commit.

mk/lib.mk
jenswi-linaro: Hmm, this comment would be more relevant below at line 70.

jforissier: Line 70 is stilll in the `ifeq ($(CFG_ULIBS_SHARED),y) .. endif` block

jenswi-linaro: My bad, you're right.

jenswi-linaro:

Issue #5841, opened 2023-02-22
CAAM crypo driver doesn't return TEE_ERROR_SHORT_BUFFER for RSA PKCS#1 v1.5 decryption


anta5010:
This issue is a follow-up to https://github.com/OP-TEE/optee_os/issues/4971

  In https://github.com/OP-TEE/optee_os/commit/8bddeb8511ba0d00a18bb603bfd816825adea3f1 Crypto API was updated to check buffer size for encryption operation where Crypto API can make a decision based on key length. But, for decryption operation Crypto API left it for drivers to decide. 

  The caam driver checks buffer size for RSA OAEP decryption and returns TEE_ERROR_SHORT_BUFFER if the buffer is not large enough:
https://github.com/OP-TEE/optee_os/blob/809fa817ae6331d98b55f7afaa3c20f8407822e4/core/drivers/crypto/caam/acipher/caam_rsa.c#L922

 For RSA PKCS#1 v1.5 decryption it always returns TEE_SUCCESS regardless of buffer size
https://github.com/OP-TEE/optee_os/blob/809fa817ae6331d98b55f7afaa3c20f8407822e4/core/drivers/crypto/caam/acipher/caam_rsa.c#L1459

PR #5842, opened 2023-02-22
core: drivers: crypto: caam: Check PKCS_V1_5 decryption buffer size

anta5010:
Check if original buffer is large enough for a result of RSA PKCS_V1_5 decryption operation.
With this change PKCS11 variable length buffers are supported for all RSA operations:
- Crypto API checks it for PKCS_V1_5 and OAEP encryptions.
- OAEP decryption already supports it.

This fixes: https://github.com/OP-TEE/optee_os/issues/5841

### DIFF
--- a/core/drivers/crypto/caam/acipher/caam_rsa.c
+++ b/core/drivers/crypto/caam/acipher/caam_rsa.c
@@ -1456,6 +1456,14 @@ static TEE_Result do_caam_decrypt(struct drvcrypt_rsa_ed *rsa_data,
 			cache_operation(TEE_CACHEINVALIDATE, size_msg.data,
 					size_msg.length);
 
+			/* Check if the original buffer size is large enough */
+			if (msg.orig.length < caam_read_val32(size_msg.data)) {
+				rsa_data->message.length =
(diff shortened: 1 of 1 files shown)


anta5010:
@clementfaure @etienne-lms could you have a look.
The "code style" checks fails on commit description, but this commit will be squashed after reviews anyway.

Testing I've done:

Parsec service uses PKCS11 variable buffer length functionality for
- Encryption: https://github.com/parallaxsecond/rust-cryptoki/blob/617efa846d67755d48708eb74e3d8149e184e83e/cryptoki/src/session/encryption.rs#L32
- Decryption: https://github.com/parallaxsecond/rust-cryptoki/blob/617efa846d67755d48708eb74e3d8149e184e83e/cryptoki/src/session/decryption.rs#L32

For Prasec PKCS11 provider backend I'm using PKCS11 TA with CAAM drivers.
Without the fix RSA PKCS#1 v1.5 encryption works (because Crypto API checks for buffer length) and decryption fails. With this fix decryption works as well.

clementfaure:
For such small patch, you can squash your fixup.

PR #5852, opened 2023-02-28
drivers: pinctrl: add device-tree pinctrl support 

clementleger:
Add pinctrl support code for device-tree parsing and to allow drivers to apply pin muxing configuration based on the `pinctrl-<x>`and `pinctrl-names`. Also add a PIO driver for the sama5d2 pin muxing controller that provides a way to apply these configurations. This support will be used for the upcoming mcp16502 I2C connected PMIC support.

### DIFF
--- a/core/drivers/pinctrl/atmel_pio.c
+++ b/core/drivers/pinctrl/atmel_pio.c
@@ -1,0 +1,240 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright 2019 Microchip.
+ */
+#include <assert.h>
+#include <drivers/clk.h>
+#include <drivers/clk_dt.h>
+#include <drivers/pinctrl.h>
+#include <initcall.h>
+#include <io.h>
+#include <kernel/dt.h>
+#include <libfdt.h>
+#include <malloc.h>
+#include <matrix.h>
+#include <platform_config.h>
+#include <trace.h>
+#include <util.h>
+
+#define PIO_GROUP_COUNT		4
+#define PIO_GROUP_OFFSET	0x40
+#define PIO_REG(reg, group)	((reg) + ((group) * PIO_GROUP_OFFSET))
+/* Mask register */
+#define PIO_MSKR(group)		PIO_REG(0x0, (group))
+/* Configuration register */
+#define PIO_CFGR(group)		PIO_REG(0x4, (group))
+#define PIO_CFGR_FUNC		GENMASK(2, 0)
+#define PIO_CFGR_PUEN		BIT(9)
+#define PIO_CFGR_PDEN		BIT(10)
+
+/* Non-Secure configuration register */
+#define PIO_SIONR(group)	PIO_REG(0x30, (group))
+/* Secure configuration register */
+#define PIO_SIOSR(group)	PIO_REG(0x34, (group))
+
+#define DT_GET_PIN_NO(val)	((val) & 0xFF)
+#define DT_GET_FUNC(val)	(((val) >> 16) & 0xF)
+
+struct atmel_pio {
+	vaddr_t base;
+};
+
+struct atmel_pio_pin_conf {
+	uint32_t pin_mask;
+	uint32_t pin_cfg;
+	uint8_t pio_group;
+	struct atmel_pio *pio;
+};
+
+static void pio_write(struct atmel_pio *pio, unsigned int offset, uint32_t val)
+{
+	io_write32(pio->base + offset, val);
+}
+
+static TEE_Result pio_conf_apply(struct pinconf *conf)
+{
+	struct atmel_pio_pin_conf *pio_conf = conf->priv;
+	struct atmel_pio *pio = pio_conf->pio;
+
+	DMSG("Apply cfg %#" PRIx32 " on group %" PRIu8 ", pins %#" PRIx32,
+	     pio_conf->pin_cfg, pio_conf->pio_group, pio_conf->pin_mask);
+
+	pio_write(pio, PIO_SIOSR(pio_conf->pio_group), pio_conf->pin_mask);
+	pio_write(pio, PIO_MSKR(pio_conf->pio_group), pio_conf->pin_mask);
+	pio_write(pio, PIO_CFGR(pio_conf->pio_group), pio_conf->pin_cfg);
+
+	return TEE_SUCCESS;
+}
+
+static void pio_conf_free(struct pinconf *conf)
+{
+	free(conf);
+}
+
+static const struct pinctrl_ops pio_pinctrl_ops = {
+	.conf_apply = pio_conf_apply,
+	.conf_free = pio_conf_free,
+};
+
+static struct pinconf *pio_pinctrl_dt_get(struct dt_driver_phandle_args *a,
+					  void *data, TEE_Result *res)
+{
+	int i = 0;
+	int func = 0;
+	int group = 0;
+	int pin_no = 0;
+	uint32_t cfg = 0;
+	int prop_count = 0;
+	int pio_group = -1;
+	uint32_t pinmux = 0;
+	uint32_t pin_mask = 0;
+	bitstr_t *cfg_modes = NULL;
+	const uint32_t *prop = NULL;
+	struct pinconf *pinconf = NULL;
+	struct atmel_pio *atmel_pio = data;
+	struct atmel_pio_pin_conf *pio_conf = NULL;
+
+	prop = fdt_getprop(a->fdt, a->phandle_node, "pinmux", &prop_count);
+	if (!prop) {
+		*res = TEE_ERROR_ITEM_NOT_FOUND;
+		return NULL;
+	}
+
+	prop_count /= sizeof(uint32_t);
+	for (i = 0; i < prop_count; i++) {
+		pinmux = fdt32_to_cpu(prop[i]);
+
+		pin_no = DT_GET_PIN_NO(pinmux);
+		func = DT_GET_FUNC(pinmux);
+
+		group = pin_no / 32;
+		if (pio_group == -1) {
+			pio_group = group;
+		} else {
+			if (group != pio_group) {
+				EMSG("Unexpected group %d vs %d", group,
+				     pio_group);
+				*res = TEE_ERROR_GENERIC;
+				return NULL;
+			}
+		}
+
+		pin_mask |= BIT(pin_no % 32);
+	}
+
+	cfg = func;
+
+	*res = pinctrl_parse_dt_pin_modes(a->fdt, a->phandle_node, &cfg_modes);
+	if (*res)
+		return NULL;
+
+	for (i = 0; i < PINCTRL_DT_PROP_MAX; i++) {
+		if (!bit_test(cfg_modes, i))
+			continue;
+
+		switch (i) {
+		case PINCTRL_DT_PROP_BIAS_PULL_UP:
+			cfg |= PIO_CFGR_PUEN;
+			cfg &= ~PIO_CFGR_PDEN;
+			break;
+		case PINCTRL_DT_PROP_BIAS_PULL_DOWN:
+			cfg |= PIO_CFGR_PDEN;
+			cfg &= ~PIO_CFGR_PUEN;
+			break;
+		case PINCTRL_DT_PROP_BIAS_DISABLE:
+			break;
+		default:
+			EMSG("Unhandled config %u", i);
+			break;
+		}
+	}
+
+	free(cfg_modes);
+
+	pinconf = calloc(1, sizeof(*pinconf) + sizeof(*pio_conf));
+	if (!pinconf) {
+		*res = TEE_ERROR_OUT_OF_MEMORY;
+		return NULL;
+	}
+
+	pio_conf = (struct atmel_pio_pin_conf *)(pinconf + 1);
+
+	pio_conf->pin_mask = pin_mask;
+	pio_conf->pin_cfg = cfg;
+	pio_conf->pio = atmel_pio;
+	pio_conf->pio_group = pio_group;
+	pinconf->priv = pio_conf;
+	pinconf->ops = &pio_pinctrl_ops;
+
+	*res = TEE_SUCCESS;
+
+	return pinconf;
+}
+
+static void pio_init_hw(struct atmel_pio *pio)
+{
+	int i = 0;
+
+	/* Set all IOs as non-secure */
+	for (i = 0; i < PIO_GROUP_COUNT; i++)
+		pio_write(pio, PIO_SIONR(PIO_GROUP_COUNT), GENMASK_32(31, 0));
+}
+
+static TEE_Result pio_node_probe(const void *fdt, int node,
+				 const void *compat_data __unused)
+{
+	size_t size = 0;
+	struct clk *clk = NULL;
+	struct atmel_pio *pio = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (fdt_get_status(fdt, node) != DT_STATUS_OK_SEC)
+		return TEE_ERROR_BAD_STATE;
+
+	pio = calloc(1, sizeof(*pio));
+	if (!pio)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	res = clk_dt_get_by_index(fdt, node, 0, &clk);
+	if (res)
+		goto free_pio;
+
+	if (dt_map_dev(fdt, node, &pio->base, &size, DT_MAP_AUTO) < 0)
+		goto free_pio;
+
+	res = clk_enable(clk);
+	if (res)
+		goto free_pio;
+
+	matrix_configure_periph_secure(AT91C_ID_PIOA);
+	matrix_configure_periph_secure(AT91C_ID_PIOB);
+	matrix_configure_periph_secure(AT91C_ID_PIOC);
+	matrix_configure_periph_secure(AT91C_ID_PIOD);
+
+	pio_init_hw(pio);
+
+	res = pinctrl_register_provider(fdt, node, pio_pinctrl_dt_get, pio);
+	if (res)
+		goto disable_clock;
+
+	return TEE_SUCCESS;
+
+disable_clock:
+	clk_disable(clk);
+free_pio:
+	free(pio);
+
+	return res;
+}
+
+static const struct dt_device_match atmel_pio_match_table[] = {
+	{ .compatible = "atmel,sama5d2-pinctrl" },
+	{ }
+};
+
+DEFINE_DT_DRIVER(atmel_pio_dt_driver) = {
+	.name = "atmel_pio",
+	.type = DT_DRIVER_PINCTRL,
+	.match_table = atmel_pio_match_table,
+	.probe = pio_node_probe,
+};
--- a/core/drivers/pinctrl/pinctrl.c
+++ b/core/drivers/pinctrl/pinctrl.c
@@ -1,0 +1,132 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2023, Microchip
+ */
+
+#include <assert.h>
+#include <drivers/pinctrl.h>
+#include <libfdt.h>
+#include <stdio.h>
+#include <tee_api_defines.h>
+#include <tee_api_defines_extensions.h>
+#include <tee_api_types.h>
+#include <util.h>
+
+static const char * const pin_modes[PINCTRL_DT_PROP_MAX] = {
+	[PINCTRL_DT_PROP_BIAS_DISABLE] = "bias-disable",
+	[PINCTRL_DT_PROP_BIAS_PULL_UP] = "bias-pull-up",
+	[PINCTRL_DT_PROP_BIAS_PULL_DOWN] = "bias-pull-down",
+};
+
+TEE_Result pinctrl_parse_dt_pin_modes(const void *fdt, int node,
+				      bitstr_t **modes)
+{
+	unsigned int i = 0;
+	bitstr_t *modes_ptr = NULL;
+
+	modes_ptr = bit_alloc(PINCTRL_DT_PROP_MAX);
+	if (!modes_ptr)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	for (i = 0; i < ARRAY_SIZE(pin_modes); i++)
+		if (fdt_getprop(fdt, node, pin_modes[i], NULL))
+			bit_set(modes_ptr, i);
+
+	*modes = modes_ptr;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pinctrl_apply_state(struct pinctrl_state *state)
+{
+	unsigned int i = 0;
+	struct pinconf *conf = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	for (i = 0; i < state->conf_count; i++) {
+		conf = state->confs[i];
+
+		res = conf->ops->conf_apply(conf);
+		if (res) {
+			EMSG("Failed to apply pin conf");
+			return res;
+		}
+	}
+
+	return TEE_SUCCESS;
+}
+
+void pinctrl_free_state(struct pinctrl_state *state)
+{
+	unsigned int i = 0;
+
+	for (i = 0; i < state->conf_count; i++)
+		state->confs[i]->ops->conf_free(state->confs[i]);
+
+	free(state);
+}
+
+TEE_Result pinctrl_get_state_by_idx(const void *fdt, int nodeoffset,
+				    unsigned int pinctrl_index,
+				    struct pinctrl_state **state_ret)
+{
+	int bw = 0;
+	unsigned int conf_id = 0;
+	const uint32_t *prop = NULL;
+	unsigned int conf_count = 0;
+	/* Enough char to hold "pinctrl-<max_int>" */
+	char prop_name[8 + 20 + 1] = { };
+	struct pinctrl_state *state = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	bw = snprintf(prop_name, sizeof(prop_name), "pinctrl-%d",
+		      pinctrl_index);
+	if (bw >= (int)sizeof(prop_name))
+		return TEE_ERROR_OVERFLOW;
+
+	prop = fdt_getprop(fdt, nodeoffset, prop_name, (int *)&conf_count);
+	if (!prop)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	conf_count /= sizeof(uint32_t);
+	state = calloc(1, sizeof(struct pinctrl_state) +
+			  conf_count * sizeof(struct pinconf *));
+	if (!state)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	state->conf_count = conf_count;
+	for (conf_id = 0; conf_id < conf_count; conf_id++) {
+		state->confs[conf_id] =
+			dt_driver_device_from_node_idx_prop(prop_name, fdt,
+							    nodeoffset, conf_id,
+							    DT_DRIVER_PINCTRL,
+							    &res);
+		if (res) {
+			free(state);
+			return res;
+		}
+	}
+
+	*state_ret = state;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pinctrl_get_state_by_name(const void *fdt, int nodeoffset,
+				     const char *name,
+				     struct pinctrl_state **state)
+{
+	int pinctrl_index = 0;
+
+	if (!name)
+		name = "default";
+
+	pinctrl_index = fdt_stringlist_search(fdt, nodeoffset, "pinctrl-names",
+					      name);
+	if (pinctrl_index < 0) {
+		*state = NULL;
+		return TEE_ERROR_GENERIC;
+	}
+
+	return pinctrl_get_state_by_idx(fdt, nodeoffset, pinctrl_index, state);
+}
--- a/core/include/drivers/pinctrl.h
+++ b/core/include/drivers/pinctrl.h
@@ -1,0 +1,176 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022-2023, Microchip
+ */
+
+#ifndef __DRIVERS_PINCTRL_H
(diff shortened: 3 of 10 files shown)


clementleger:
One checkpatch error is remaining but I did not found an easy way to fix it. Splitting it accross lines yields another checkpatch error.

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,94 @@
...
+ */
+struct pinctrl {
+	int node;
+	TEE_Result (*apply_state)(struct pinctrl *pinctrl, const void *fdt,
+				  int node);
+	SLIST_ENTRY(pinctrl) link;
+};
+
+#ifdef CFG_DRIVERS_PINCTRL
+
```
etienne-lms, line 33: can remove this empty line

etienne-lms (commented):
Thanks @clementleger for this pinctrl framework.
I have many comments. Maybe some could be addressed later. Feel free to argue.

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @fdt: Device-tree which contains the pinctrl configuration node to parse
+ * @node: Pinctrl configuration node to parse
+ * @cfg: Returned configuration
+ * @count: Count of elements in the @cfg array.
+ *
+ * Return a TEE_Result compliant value
+ */
+TEE_Result pinctrl_parse_dt_pin_mode_config(const void *fdt, int node,
+					    uint32_t **cfg,
```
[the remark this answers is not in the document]
etienne-lms: My comment is rather stupid as this function explicitly outputs the config mode value.  Discard it.

That said, `cfg` argument should rather be of type `enum pin_config_mode **`.
Or maybe the function could be more generic `pinctrl_parse_dt_info()` and use a `struct pinctrl_info` initially holding a unique field `enum pin_mode_config mode`.
What do you think?

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,94 @@
...
+					    unsigned int *count);
+#else
```
GseoC, line 69: You can recall here it's

#else /* CFG_DRIVERS_PINCTRL */

It helps reading :)

ditto for #endif

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * pinctrl_apply_state - Apply a pinctrl state
+ *
+ * @fdt: Device-tree containing the pinctrl node
+ * @node: Node of the device for which to apply the pinctrl state
+ * @name: Name of the pinctrl state to apply or NULL for the "default" one
+ *
+ * Return a TEE_Result compliant value
+ */
+TEE_Result pinctrl_apply_state(const void *fdt, int node, const char *name);
```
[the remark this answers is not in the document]
GseoC: +1

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,94 @@
...
+#include <tee_api_types.h>
+
+enum pin_config_mode {
+	PIN_MODE_BIAS_DISABLE,
+	PIN_MODE_BIAS_PULL_UP,
+	PIN_MODE_BIAS_PULL_DOWN,
+};
+
+/*
+ * struct pinctrl - Describe a pinctrl controller
```
GseoC, line 19: "Describe a pin controller" or "Describe a pinctrl" is enough

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,191 @@
...
+			break;
+		default:
+			EMSG("Unhandled mode %"PRIu32"\n", configs[i]);
+			break;
+		}
+	}
+
+	free(configs);
+
+	DMSG("Applying configuration on group %d, pins %x, cfg %x\n", pio_group,
```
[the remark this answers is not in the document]
clementleger: Will use `DMSG("Apply cfg %x on group %d, pins %x", cfg, pio_group, pin_mask);`

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,191 @@
...
+			cfg &= (~PIO_CFGR_PDEN);
+			break;
+		case PIN_MODE_BIAS_PULL_DOWN:
+			cfg |= PIO_CFGR_PDEN;
+			cfg &= (~PIO_CFGR_PUEN);
+			break;
+		case PIN_MODE_BIAS_DISABLE:
+			break;
+		default:
+			EMSG("Unhandled mode %"PRIu32"\n", configs[i]);
```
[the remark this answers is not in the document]
clementleger: Will display "config"

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,191 @@
...
+		pio_write(pio, PIO_SIONR(PIO_GROUP_COUNT), GENMASK_32(31, 0));
+}
+
+static TEE_Result pio_node_probe(const void *fdt, int node,
+				 const void *compat_data __unused)
+{
+	size_t size = 0;
+	struct clk *clk = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct atmel_pio *pio = NULL;
```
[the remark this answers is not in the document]
clementleger: Acked

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,133 @@
...
+		if (pinctrl->node == nodeoffset)
+			return pinctrl;
+	}
+
+	return NULL;
+}
+
+static struct pinctrl *pinctrl_find_parent(const void *fdt, int node)
+{
+	int parent_node = -1;
```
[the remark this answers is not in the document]
clementleger: Since we want a sane init value and that for a node value, -1 is expected as an error.

core/include/drivers/pinctrl.h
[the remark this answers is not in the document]
clementleger: The return value is a mask of multiple `enum pin_config_mode`, that's why I used a `uint32_t`. But passing `struct pinconf_info`  could be done (pinctrl refers to the pin controller, a pinconf however could refer to a pinconf)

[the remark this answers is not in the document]
clementleger: yes, maybe passing a `struct pinconf_info` might be done (pinctrl refers to the pin controller itself, pinconf to the pin configuration)

core/include/drivers/pinctrl.h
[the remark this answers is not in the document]
clementleger: Regarding the dynamic applying of pinctrl, This is something rarely used, even in Linux driver. Do you have such use case ?

About the `DT_DRIVER_PINCTRL` and clk like probing/callbacks, unless the other type of controller that are using `*-cells` properties, pinctrl do not use such notation but rather a phandle to a subnode of the parent pinctrl node. 
Moreover, rather than the `get_of_device` callback is not applicable since  the pinctrl apply function needs a node to be applied and not some `struct dt_driver_phandle_args` that contaisn a few cells value.  So, this method of looking up is not easily applicable to pinctrl nodes unless modifying your existing DT code to handle that exact specific case and passing a node  in the phandle args rather than cells value. Not sure if this is better though.

Regarding` TEE_Result pinctrl_dt_get(fdt, node, struct pinctrl **pinctrl);`, what is this function supposed to do ?  A node might have multiple pinctrl subnode references so I'm not sure what you expect here and in this PR `struct pinctrl` refers to a pin controller. So maybe you meant something else.

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2022, Microchip
```
etienne-lms, line 3: 2022-2023

core/include/drivers/pinctrl.h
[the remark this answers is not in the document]
etienne-lms: > Regarding the dynamic applying of pinctrl, This is something rarely used, even in Linux driver. Do you have such use case ?

I am thinking of suspend time here specifc config can be applied by drivers.

> About the DT_DRIVER_PINCTRL and clk like probing/callbacks ...
> ... So, this method of looking up is not easily applicable to pinctrl

Ok, understood, the device lists of dt_driver will not help much.

> Regarding TEE_Result pinctrl_dt_get(fdt, node, struct pinctrl **pinctrl);, what is this function supposed to do ? ...

If I understood correctly your explainations, I would expect it gets the consumer node as input arg and outputs a handle for a pinmiux config list (e.g. `pinconf_list *`) which can be use at run or PM time to load alternate pinmux configs, as argument to `pinconf_load_by_index()`.

core/include/drivers/pinctrl.h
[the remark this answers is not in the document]
etienne-lms: +1

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,191 @@
...
+	uint32_t pin_mask = 0;
+	uint32_t *configs = NULL;
+	unsigned int prop_count = 0;
+	const uint32_t *prop = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct atmel_pio *pio = container_of(pinctrl, struct atmel_pio,
+					     pinctrl);
+	unsigned int configs_count = 0;
+
+	prop = fdt_getprop(fdt, node, "pinmux", (int *)&prop_count);
```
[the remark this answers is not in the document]
etienne-lms: ok

core/include/drivers/pinctrl.h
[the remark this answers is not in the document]
clementleger: > > Regarding the dynamic applying of pinctrl, This is something rarely used, even in Linux driver. Do you have such use case ?
> 
> I am thinking of suspend time here specifc config can be applied by drivers.
> 
> > About the DT_DRIVER_PINCTRL and clk like probing/callbacks ...
> > ... So, this method of looking up is not easily applicable to pinctrl
> 
> Ok, understood, the device lists of dt_driver will not help much.
Actually, after thinking a bit about it, we might potentially modify your DT parsing code with the following modification:
- Return 0 as cells count for DT_DRIVER_PINCTRL rather than parsing a cells property
- Add a "int node_phandle" in the `struct dt_driver_phandle_args` which will contains the phandle of the controller/node in case of pinctrl (ie, the phandle specified before the cells values when refering to them)
- In dt device lookup, when resolving the references, in case we handle a DT_DRIVER_PINCTRL, get the parent instead of looking for the phandle to resolve the provider

If you are ok with that, I think that with these modifications, we can get rid of the custom code I have and be integrated with your DT driver framework. There will be some special handling but that seems rather limited to me. I'll try that and resubmit it. 

> > Regarding TEE_Result pinctrl_dt_get(fdt, node, struct pinctrl **pinctrl);, what is this function supposed to do ? ...
> 
> If I understood correctly your explainations, I would expect it gets the consumer node as input arg and outputs a handle for a pinmiux config list (e.g. `pinconf_list *`) which can be use at run or PM time to load alternate pinmux configs, as argument to `pinconf_load_by_index()`.

Ok, get it, I was confuse by the use of the `struct pinctrl` I think this can also be done at the cost of increase use of memory due to some intermediate structure. I think this is also cleaner and will thus match a bit more the use of other frameworks.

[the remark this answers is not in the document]
etienne-lms: I think we're aligned. I see we need some clear vocabulary, i am also confusing pinctrl provider refs, pinctrl config refs and pinctrl consumer config-list refs.

As for the DT bnidings that allow to refer to a pinmux config by_name() or by_index(), we could keep it simple and only support by_index() at runtime, with a helper name2index() converter helper function to not malloc/strdup names  at init.

clementleger:
New version:
- Totally rework pinctrl to use DT infrastructure.
- Split the way pinctrl configurations are get/applied.
- Add a new pinctrl_get_by_idx/name function
- Add a new pinctrl_apply_state function

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,148 @@
...
+	cfg_modes_ptr->modes_count = prop_count;
+
+	prop_count = 0;
+	for (i = 0; i < ARRAY_SIZE(pin_modes); i++) {
+		if (!fdt_getprop(fdt, node, pin_modes[i], NULL))
+			continue;
+
+		cfg_modes_ptr->modes[prop_count] = i;
+		prop_count++;
+	}
```
etienne-lms, line 52: I don't understand why you are storing several mode.
Each pin in a pinctrl has 1 of the 3 bias related cfg, right?

core/kernel/dt_driver.c
```diff
@@ -286,6 +289,19 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
 			return NULL;
 		}
 
+		/* Pinctrl states refers to a subnode of the controller node */
```
etienne-lms, line 292: Getting `phandle_node` is usefull only for pinctrl (up to now). `fdt_node_offset_by_phandle()` as well as `fdt_parent_offset()` reparses FDT, better avoid if not usefull.

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,148 @@
...
+#include <drivers/pinctrl.h>
+#include <libfdt.h>
+#include <stdio.h>
+#include <sys/queue.h>
+#include <tee_api_defines.h>
+#include <tee_api_defines_extensions.h>
+#include <tee_api_types.h>
+#include <util.h>
+
+SLIST_HEAD(pinctrl_list, pinctrl) pinctrl_list;
```
etienne-lms, line 16: seems unused

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,230 @@
...
+				     pio_group);
+				*res = TEE_ERROR_GENERIC;
+				return NULL;
+			}
+		}
+
+		pin_mask |= BIT(pin_no % 32);
+	}
+
+	cfg = func;
```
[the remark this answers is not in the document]
clementleger: Yes, for the sama5d2 pinctrl driver, this is expected to be true.

core/drivers/pinctrl/pinctrl.c
clementleger: Indeed.

core/drivers/pinctrl/pinctrl.c
clementleger: These property are actually just a few amongst the supported ones by the different bindings. So for the moment, there is only bias supported since the atmel driver only handles that, but if you look at the bindings (https://elixir.bootlin.com/linux/latest/source/Documentation/devicetree/bindings/pinctrl/pincfg-node.yaml), there are much more that can potentially be supported later.

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	*cfg_modes = cfg_modes_ptr;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pinctrl_apply_state(struct pinctrl_state *state)
+{
+	unsigned int conf_id = 0;
```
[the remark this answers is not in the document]
clementleger: Acked.

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,148 @@
...
+	unsigned int i = 0;
+	unsigned int prop_count = 0;
+	struct pin_modes *cfg_modes_ptr;
+
+	for (i = 0; i < ARRAY_SIZE(pin_modes); i++) {
+		if (!fdt_getprop(fdt, node, pin_modes[i], NULL))
+			continue;
+
+		prop_count++;
+	}
```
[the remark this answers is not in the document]
clementleger: Acked

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,148 @@
...
+	conf_count /= sizeof(uint32_t);
+	state = calloc(1, sizeof(struct pinctrl_state) +
+		       conf_count * sizeof(struct pinconf *));
+	if (!state)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	state->conf_count = conf_count;
+	for (conf_id = 0; conf_id < conf_count; conf_id++) {
+		state->confs[conf_id] =
+			dt_driver_device_from_node_idx_prop(prop_name, fdt,
```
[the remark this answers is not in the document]
clementleger: Can't get everything right all the time ;)

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,230 @@
...
+	clk_enable(clk);
+
+	matrix_configure_periph_secure(AT91C_ID_PIOA);
+	matrix_configure_periph_secure(AT91C_ID_PIOB);
+	matrix_configure_periph_secure(AT91C_ID_PIOC);
+	matrix_configure_periph_secure(AT91C_ID_PIOD);
+
+	pio_init_hw(pio);
+
+	return pinctrl_register_provider(fdt, node, pio_pinctrl_dt_get, pio);
```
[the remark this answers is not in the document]
clementleger: I guess we would need so yes. would need dt_unmap and so on ...

core/kernel/dt_driver.c
clementleger: Ok, will move this in the `if (type == DT_DRIVER_PINCTRL)`

core/drivers/pinctrl/pinctrl.c
etienne-lms: ok, right.
suggestion: bitstring.h could help to ease passing boolean properties. Here with:
```c
	for (i = 0; i < PINCTRL_DT_PROPS_MAX; i++)
		if (fdt_getprop(fdt, node, pinctrl_dt_prop[i], NULL))
			bit_set(i, PINCTRL_DT_PROPS_MAX);
```
with
```
enum pinctrl_dt_prop {
	PINCTRL_DT_PROP_BIAS_DISABLE,
	PINCTRL_DT_PROP_BIAS_PULL_UP,
	PINCTRL_DT_PROP_BIAS_PULL_DOWN,
	PINCTRL_DT_PROPS_MAX
};

static const char * const pinctrl_dt_prop[] = {
	[PINCTRL_DT_PROP_BIAS_DISABLE] = "bias-disable",
	...
```

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,148 @@
...
+
+	snprintf(prop_name, sizeof(prop_name), "pinctrl-%d", pinctrl_id);
+
+	prop = fdt_getprop(fdt, nodeoffset, prop_name, (int *)&conf_count);
+	if (!prop)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	conf_count /= sizeof(uint32_t);
+	state = calloc(1, sizeof(struct pinctrl_state) +
+		       conf_count * sizeof(struct pinconf *));
```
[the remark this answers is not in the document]
etienne-lms: see line 32

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,126 @@
...
+#include <tee_api_types.h>
+
+enum pin_config_mode {
+	PIN_MODE_BIAS_DISABLE,
+	PIN_MODE_BIAS_PULL_UP,
+	PIN_MODE_BIAS_PULL_DOWN,
+};
+
+struct pinconf {
+	const struct pinctrl_ops *ops;
```
etienne-lms, line 20: instead of passing `ops` in each pinconf, it could be made common in `struct pinctrl_state`

core/kernel/dt_driver.c
```diff
@@ -277,6 +283,25 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
...
+		if (type == DT_DRIVER_PINCTRL) {
+			phandle_node = fdt_node_offset_by_phandle(fdt, phandle);
+			if (phandle_node < 0) {
+				*res = TEE_ERROR_GENERIC;
+				return NULL;
+			}
+
+			nodeoffset = fdt_parent_offset(fdt, phandle_node);
+			if (nodeoffset < 0)
+				return NULL;
```
etienne-lms, line 296: error case

clementleger: Acked

core/drivers/pinctrl/pinctrl.c
clementleger: Ok acked, i'll switch to a bitmap instead of multiple `uint32_t`

core/include/drivers/pinctrl.h
etienne-lms: ok

clementleger:
New version:
- Used `bitstr_t` instead of uint32_t array for pin modes.
- Added `dt_unmap_dev()` to unmap a previously mapped device
- Add error path handling in atmel_pio driver

core/kernel/dt_driver.c
```diff
@@ -280,6 +283,27 @@ void *dt_driver_device_from_node_idx_prop(const char *prop_name,
...
+			}
+
+			nodeoffset = fdt_parent_offset(fdt, phandle_node);
+			if (nodeoffset < 0) {
+				*res = TEE_ERROR_GENERIC;
+				return NULL;
+			}
+
+			phandle = fdt_get_phandle(fdt, nodeoffset);
+			if (!phandle) {
```
etienne-lms, line 301: `if (phandle <= 0)`

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,126 @@
...
+	unsigned int i = 0;
+
+	for (i = 0; i < state->conf_count; i++)
+		state->confs[i]->ops->conf_free(state->confs[i]);
+
+	free(state);
+}
+
+TEE_Result pinctrl_get_state_by_idx(const void *fdt, int nodeoffset,
+				    unsigned int pinctrl_id,
```
etienne-lms, line 70: s/pinctrl_id/pinctrl_index/

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,126 @@
...
+
+	state->conf_count = conf_count;
+	for (conf_id = 0; conf_id < conf_count; conf_id++) {
+		state->confs[conf_id] =
+			dt_driver_device_from_node_idx_prop(prop_name, fdt,
+							    nodeoffset, conf_id,
+							    DT_DRIVER_PINCTRL,
+							    &res);
+		if (res)
+			return res;
```
etienne-lms, line 101: since `dt_driver_device_from_node_idx_prop()` can return `TEE_ERROR_DEFER_DRIVER_INIT`, this error case should free `state`.

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,126 @@
...
+	*state_ret = state;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pinctrl_get_state_by_name(const void *fdt, int nodeoffset,
+				     const char *name,
+				     struct pinctrl_state **state)
+{
+	int pinctrl_id = 0;
```
etienne-lms, line 113: s/pinctrl_id/pinctrl_index/

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,153 @@
...
+#include <bitstring.h>
+#include <kernel/dt_driver.h>
+#include <sys/queue.h>
+#include <tee_api_types.h>
+
+enum pinctrl_dt_prop {
+	PINCTRL_DT_PROP_BIAS_DISABLE,
+	PINCTRL_DT_PROP_BIAS_PULL_UP,
+	PINCTRL_DT_PROP_BIAS_PULL_DOWN,
+	PINCTRL_DT_PROP_MAX,
```
etienne-lms, line 18: can remove the above coma (_MAX is the last enumerated value)

core/kernel/dt.c
```diff
@@ -139,6 +142,24 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size,
...
+	size_t sz = 0;
+	int st = DT_STATUS_DISABLED;
+
+	st = _fdt_get_status(fdt, offs);
+	if (st == DT_STATUS_DISABLED)
+		return TEE_ERROR_BAD_STATE;
+
+	sz = _fdt_reg_size(fdt, offs);
+	if (sz == DT_INFO_INVALID_REG_SIZE)
+		return -1;
```
etienne-lms, line 157: `return TEE_ERROR_GENERIC;`

core/kernel/dt.c
```diff
@@ -139,6 +142,24 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size,
...
+
+	return core_mmu_remove_mapping(dt_to_type(st, mapping), (void *)base,
+				       sz);
```
etienne-lms, line 160: i don't think this function is always reliable: `core_mmu_add_mapping()` does not count consumers of a mapped area and reuses already mapped VA ranges when possible. `core_mmu_remove_mapping()` may therefore unmap a VA range used by another component.

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,243 @@
...
+{
+	io_write32(pio->base + offset, val);
+}
+
+static TEE_Result pio_conf_apply(struct pinconf *conf)
+{
+	struct atmel_pio_pin_conf *pio_conf = conf->priv;
+	struct atmel_pio *pio = pio_conf->pio;
+
+	DMSG("Apply cfg %x on group %d, pins %x\n", pio_conf->pin_cfg,
```
etienne-lms, line 59: `DMSG("Apply cfg %#"PRIx32" on group %"PRIu8, pins %#"PRIx32, ...`

core/drivers/pinctrl/atmel_pio.c
```diff
@@ -0,0 +1,243 @@
...
+	uint32_t pin_mask = 0;
+	bitstr_t *cfg_modes = NULL;
+	unsigned int prop_count = 0;
+	const uint32_t *prop = NULL;
+	struct pinconf *pinconf = NULL;
+	struct atmel_pio *atmel_pio = data;
+	struct atmel_pio_pin_conf *pio_conf = NULL;
+
+	prop = fdt_getprop(a->fdt, a->phandle_node, "pinmux",
+			   (int *)&prop_count);
```
etienne-lms, line 98: prefer remove this cast and use `int` for both `prop_count` and `i`.

core/drivers/pinctrl/pinctrl.c
clementleger: Oups indeed

core/kernel/dt.c
clementleger: Hum ok, so what should I do for that series ? Remove this and assume even if the driver fails we keep the memory mapping ?

clementleger:
Ping ? That would be better to be able to merge this series first before I2C since the former relies on commits that are provided by this one.

core/include/drivers/pinctrl.h
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * pinctrl_apply_state - apply a pinctrl state
+ *
+ * @state: State to be applied
+ * Return a TEE_Result compliant value
+ */
+TEE_Result pinctrl_apply_state(struct pinctrl_state *state);
+
+TEE_Result pinctrl_parse_dt_pin_modes(const void *fdt, int node,
```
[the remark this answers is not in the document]
clementleger: Acked, I'll replace "node" with "nodeoffset"

etienne-lms (commented):
A late comment for commit "drivers: pinctrl: add pinctrl support".

core/drivers/pinctrl/pinctrl.c
```diff
@@ -0,0 +1,128 @@
...
+
+	free(state);
+}
+
+TEE_Result pinctrl_get_state_by_idx(const void *fdt, int nodeoffset,
+				    unsigned int pinctrl_index,
+				    struct pinctrl_state **state_ret)
+{
+	/* Enough char to hold "pinctrl-<max_int>" */
+	char prop_name[19] = { };
```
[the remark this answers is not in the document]
clementleger: I'll actually do both, checking the return value of `snprintf()` is better in any case.

etienne-lms (approved):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "drivers: pinctrl: add pinctrl support".

etienne-lms:
@jforissier, could you trigger CI test for this P-R? I think it is clean.

etienne-lms:
Checkpatch will complain on commit "drivers: pinctrl: add pinctrl support". Please address the issue:
```
bb4f6fde3 drivers: pinctrl: add pinctrl support
WARNING: line length of 81 exceeds 80 columns
#324: FILE: core/include/drivers/pinctrl.h:131:
+				     pinctrl_dt_get_func get_dt_pinctrl __unused,
```

Hint: you can run checkpatch from your git tree:
```shell
$ export CHECKPATCH=/patch/to/linux/source/tree/scripts/checkpatch.pl
$ ./scripts/checkpatch.sh bootlin-sam/sama5d2_pinctrl~4..bootlin-sam/sama5d2_pinctrl
Checking commit(s):
65f5b1724 (bootlin-sam/sama5d2_pinctrl) plat-sam: enable pinctrl and atmel_pio driver for wlsom1 board
9a7a3e261 dts: at91: set PIO controller as secure
339880b0d drivers: pinctrl: atmel_pio: add pio controller for sama5d2
bb4f6fde3 drivers: pinctrl: add pinctrl support
WARNING: line length of 81 exceeds 80 columns
#324: FILE: core/include/drivers/pinctrl.h:131:
+				     pinctrl_dt_get_func get_dt_pinctrl __unused,

total: 0 errors, 1 warnings, 0 checks, 327 lines checked
```
See [OP-TEE checkpatch documentation](https://optee.readthedocs.io/en/latest/general/coding_standards.html#running-checkpatch).

tprrt:
> ```shell
> core/include/drivers/pinctrl.h
> ```

I renamed `get_dt_pinctrl` to `get_pinctrl`, to fix this checkpatch issue.

PR #5853, opened 2023-02-28
[RFC] drivers: i2c: add I2C controller and devices support

clementleger:
In order to support the i2c mcp16502 PMIC that is available on the wlsom1 platform, add I2C support to sama5d2. This PMIC is defined in the device-tree and uses gpio and pinctrl capabilities. This series relies on the pinctrl series that is in another PR.


clementleger:
Actually needs some rework.

PR #5859, opened 2023-03-03
 plat-totalcompute: Remap console logs 

annsai01:


jforissier:
Please force-push the change as one commit based on latest master, thanks.

PR #5865, opened 2023-03-08
core: crypto: change supported HMAC key size ranges

SOuajih:
Change supported HMAC key size range from 64 to 1024, required to successfully pass AOSP Keymaster VTS tests.

From TEE Internal Core API specificaion:
"Table 5-9, support for other sizes or algorithms is implementation-defined."

Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>
Signed-off-by: Safae Ouajih <souajih@baylibre.com>
(cherry picked from commit ab43191a4d4df481ff42df1e9bba72b7d76cb220)


jbech-linaro:
Even though this solves a problem, it'll violate the GP spec, that says for example:
```
TEE_TYPE_HMAC_SHA1 Between 80 and 512 bits, multiple of 8 bits
...
```
See "TEE_AllocateTransientObject Object Types and Key Sizes" table in the spec.

jforissier:
Perhaps the ranges could be made configurable via `CFG_*` so that an AOSP-VTS-compliant OP-TEE can be built easily (with the proper note that doing so would violate some GP requirements)?

SOuajih:
Thank you for the remarks, I will make an update using a new  `CFG_*` config

mk/config.mk
```diff
@@ -1030,3 +1030,7 @@ CFG_FAULT_MITIGATION ?= y
 # compatibility with version v1.1 as the value of this variable is not
 # preserved in the TA dev-kit.
 CFG_TA_OPTEE_CORE_API_COMPAT_1_1 ?= n
+
+# Change supported HMAC key size range, from 64 to 1024.
+# Note that this violates GP requirements of HMAC size range.
```
jforissier, line 1035: add "but is needed to pass AOSP Keymaster VTS tests" maybe? (and specify version, give link or any useful context, because it will help when times come to perhaps simplify things...)

jforissier (commented):
The line "(cherry picked from commit ...)" is not useful because it does not give any more information than is already in the commit (the referenced commit wasn't part of any official release I assume). So you should remove it.

With that and my comment below addressed:

mk/config.mk
SOuajih: Done , Thank you

jenswi-linaro (commented):

PR #5866, opened 2023-03-08
Stop using deprecated constants TEE_ALG_ECDSA_P* and TEE_ALG_ECDH_P*

jforissier:
Updates related to the deprecation of ECDSA and ECDH algorithm identifiers. This is a draft because of the last commit ("ta: pkcs11: do not use deprecated TEE_ALG_ECDSA_P* and TEE_ALG_ECDH_P*") which is incomplete. Help from PKCS#11 experts appreciated ;)

### DIFF
--- a/ta/pkcs11/src/processing_ec.c
+++ b/ta/pkcs11/src/processing_ec.c
@@ -425,11 +425,10 @@ enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
 	switch (proc_params->id) {
 	case PKCS11_CKM_ECDSA:
 		/*
-		 * FIXME
 		 * ECDSA without hashing. The TEE algo to use is unknown at this
-		 * point.
+		 * point. Return the one with the largest digest size.
 		 */
-		return PKCS11_CKR_GENERAL_ERROR;
+		return TEE_ALG_ECDSA_SHA512;
 	case PKCS11_CKM_ECDSA_SHA1:
 		*tee_id = TEE_ALG_ECDSA_SHA1;
 		break;
--- a/ta/pkcs11/src/processing_ec.c
+++ b/ta/pkcs11/src/processing_ec.c
@@ -420,43 +420,33 @@ enum pkcs11_rc load_tee_ec_key_attrs(TEE_Attribute **tee_attrs,
 
 enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
 				   struct pkcs11_attribute_head *proc_params,
-				   struct pkcs11_object *obj)
+				   struct pkcs11_object *obj __unused)
 {
 	switch (proc_params->id) {
 	case PKCS11_CKM_ECDSA:
+		/*
+		 * FIXME
+		 * ECDSA without hashing. The TEE algo to use is unknown at this
+		 * point.
+		 */
+		return PKCS11_CKR_GENERAL_ERROR;
 	case PKCS11_CKM_ECDSA_SHA1:
+		*tee_id = TEE_ALG_ECDSA_SHA1;
+		break;
 	case PKCS11_CKM_ECDSA_SHA224:
+		*tee_id = TEE_ALG_ECDSA_SHA224;
+		break;
 	case PKCS11_CKM_ECDSA_SHA256:
+		*tee_id = TEE_ALG_ECDSA_SHA256;
+		break;
 	case PKCS11_CKM_ECDSA_SHA384:
+		*tee_id = TEE_ALG_ECDSA_SHA384;
+		break;
 	case PKCS11_CKM_ECDSA_SHA512:
+		*tee_id = TEE_ALG_ECDSA_SHA512;
 		break;
 	default:
 		return PKCS11_CKR_GENERAL_ERROR;
-	}
-
-	/*
-	 * TODO: Fixing this in a way to support also other EC curves would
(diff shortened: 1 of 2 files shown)


core/drivers/crypto/versal/ecc.c
```diff
@@ -131,9 +131,9 @@ static TEE_Result ecc_prepare_msg(uint32_t algo, const uint8_t *msg,
 	if (msg_len > TEE_SHA512_HASH_SIZE + 2)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	if (algo == TEE_ALG_ECDSA_P384)
+	if (algo == TEE_ALG_ECDSA_SHA384 || algo == __OPTEE_ALG_ECDSA_P384)
```
jenswi-linaro, line 134: the `|| algo == __OPTEE_ALG_ECDSA_P384` part shouldn't be needed due to the call to `translate_compat_algo()` in `syscall_cryp_state_alloc()`, same below.

jenswi-linaro:
For the commits:
- core: mbedtls: ecc_get_keysize(): do not check algorithm against curve
- libutee: add TEE_ALG_ECDSA_SHA* to TEE_ALG_GET_DIGEST_SIZE()
please apply:

core/drivers/crypto/versal/ecc.c
jforissier: OK, I will remove it. I suppose the same applies equally to "drivers: crypto: se050: do not use deprecated algorithm macros".

lib/libutee/include/utee_defines.h
```diff
@@ -207,32 +207,37 @@ static inline size_t __tee_alg_get_digest_size(uint32_t algo)
 	case TEE_ALG_HMAC_SHA1:
 	case TEE_ALG_DSA_SHA1:
 	case TEE_ALG_ECDSA_SHA1:
+	case __OPTEE_ALG_ECDSA_P192:
```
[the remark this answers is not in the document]
jforissier: I introduced a call to `TEE_ALG_GET_DIGEST_SIZE()` in "[WIP] ta: pkcs11: do not use deprecated TEE_ALG_ECDSA_P* and TEE_ALG_ECDH_P*" but indeed the PKCS#11 TA does not use the compat values so it is useless. I will drop commit "libutee: add deprecated algorithms __OPTEE_ALG_ECDSA_P* to TEE_ALG_GET_DIGEST_SIZE()".

ta/pkcs11/src/processing_ec.c
```diff
@@ -420,45 +420,35 @@ enum pkcs11_rc load_tee_ec_key_attrs(TEE_Attribute **tee_attrs,
...
-				   struct pkcs11_object *obj)
+				   struct pkcs11_object *obj __unused)
 {
 	switch (proc_params->id) {
 	case PKCS11_CKM_ECDSA:
+		/*
+		 * FIXME
+		 * ECDSA without hashing. The TEE algo to use is unknown at this
+		 * point.
+		 */
```
etienne-lms, line 431: Here you could use the key size to define `TEE_ALG_ECDSA_SHAx` for the largest hash supported by the key (as done for `PKCS11_CKM_ECDSA` before this patch. Selected TEE algo will be used only to set/check hash/signature size (done in [`step_asym_operation()`](https://elixir.bootlin.com/op-tee/3.20.0/source/ta/pkcs11/src/processing_asymm.c#L679)). The ECDSA part, based on hash/sign. size, is done later in that function ([signing](https://elixir.bootlin.com/op-tee/3.20.0/source/ta/pkcs11/src/processing_asymm.c#L777) and [verifying](https://elixir.bootlin.com/op-tee/3.20.0/source/ta/pkcs11/src/processing_asymm.c#L787)).

etienne-lms: Actually, we could set `TEE_ALG_ECDSA_SHA512` for `PKCS11_CKM_ECDSA`. It should do the job.

`xtest pkcs11_1019` currently fails with this P-R because of below switch cases. `pkcs11_1019` uses a 384bit ECC key for several ECDSA flavors but our OP-TEE implementation mandates `ECDSA_SHAx` flavor exactly matches the ECC key size. For example, `TEE_ALG_ECDSA_SHA1` mandates a 192 bit key:
https://github.com/OP-TEE/optee_os/blob/fe2fd3ff46c0bada3dc9b2ae8c1025bfd8f57f3f/lib/libutee/tee_api_operations.c#L72-L77
Maybe the code could be modified:
```patch
-		if (maxKeySize != 192)
+ 		if (maxKeySize < 192)
```

etienne-lms: hmmm... I think i got confused in ECDSA_SHA1/P192 IDs and meanings.
I think the right way is:
```c
 case TEE_ALG_ECDSA_SHA1: 
 	if (maxKeySize < 192) 
 	break; 

 case __OPTEE_ALG_ECDSA_P192: 
 case __OPTEE_ALG_ECDH_P192: 
 	if (maxKeySize != 192) 
 	break;

jforissier: The only constraints I see in the spec on key size for `TEE_ALG_ECDSA_SHA*` is 160 <= size <= 521 (bits) so that's what I implement in the latest update (as well as your suggestion for the PKCS#11 TA). Let's see if it passes CI.

PR #5876, opened 2023-03-10
SHA-3 acceleration 

jenswi-linaro:


jenswi-linaro:
There are a few checkpatch warnings in `sha3_armv8a_ce_a64.S` for long lines, but I'm not sure if it's worth it to fix those.

jenswi-linaro:
> > There are a few checkpatch warnings in `sha3_armv8a_ce_a64.S` for long lines, but I'm not sure if it's worth it to fix those.
> 
> Isn't it just enough to split the lines with a line continuation character? (`\`) That would not look bad I believe.

Yeah, that was easier than expected. I thought that the `.irp` line would cause trouble.

jforissier (commented):

jenswi-linaro:
Tag applied

jenswi-linaro:
Good to merge?

jforissier:
> Good to merge?

Yes!

PR #5878, opened 2023-03-10
core: fix out-of-bounds access of dump_ctx

weizhaojiang:
Problem: in the case of no UTA running, the buffer of dump_ctx will be allocated with 0 size and passed to init_dump_ctx(). That causes buffer overrunning.

Solution: Check buffer size before allocate the buffer. If it's 0, return TEE_ERROR_ITEM_NOT_FOUND.

### DIFF
--- a/core/kernel/tee_ta_manager.c
+++ b/core/kernel/tee_ta_manager.c
@@ -982,7 +982,10 @@ TEE_Result tee_ta_instance_stats(void *buf, uint32_t *buf_size)
 			ta_count++;
 
 	sz = sizeof(struct tee_ta_dump_stats) * ta_count;
-	if (!buf || *buf_size < sz) {
+	if (!sz) {
+		/* sz = 0 means there is no UTA, return no item found. */
+		res = TEE_ERROR_ITEM_NOT_FOUND;
+	} else if (!buf || *buf_size < sz) {
 		/*
 		 * buf is null or pass size less than actual size
 		 * means caller try to query the buffer size.


core/kernel/tee_ta_manager.c
```diff
@@ -982,7 +982,14 @@ TEE_Result tee_ta_instance_stats(void *buf, uint32_t *buf_size)
 			ta_count++;
 
 	sz = sizeof(struct tee_ta_dump_stats) * ta_count;
-	if (!buf || *buf_size < sz) {
+	if (sz == 0) {
+		/*
+		 * sz = 0 means there is no UTA, set buf_size
+		 * as 0 and return no item found.
+		 */
+		*buf_size = 0;
```
jforissier, line 990: Updating `buf_size` is not necessary I believe. It is supposed to be used only when return value is `TEE_SUCCESS` or `TEE_ERROR_SHORT_BUFFER`.

jforissier (commented):
Good catch! One comment below, then:

core/kernel/tee_ta_manager.c
weizhaojiang: done. thanks for the reviewing.

core/kernel/tee_ta_manager.c
```diff
@@ -982,7 +982,10 @@ TEE_Result tee_ta_instance_stats(void *buf, uint32_t *buf_size)
 			ta_count++;
 
 	sz = sizeof(struct tee_ta_dump_stats) * ta_count;
-	if (!buf || *buf_size < sz) {
+	if (sz == 0) {
```
jenswi-linaro, line 985: `if (!sz)` is the preferred style.

jenswi-linaro (commented):
With the comment addressed, please apply:

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with @jenswi-linaro comment addressed, for consistency.

core/kernel/tee_ta_manager.c
weizhaojiang: done, thanks for reviewing.

PR #5881, opened 2023-03-13
plat-stm32mp1: pager resident size optims

etienne-lms:
Few changes it optimize a bit resident memory size when using pager.

### DIFF
--- a/core/arch/arm/mm/core_mmu_v7.c
+++ b/core/arch/arm/mm/core_mmu_v7.c
@@ -794,7 +794,6 @@ void core_init_mmu_regs(struct core_mmu_config *cfg)
 	 */
 	cfg->ttbcr = TTBCR_N_VALUE;
 }
-DECLARE_KEEP_PAGER(core_init_mmu_regs);
 
 enum core_mmu_fault core_mmu_get_fault_type(uint32_t fsr)
 {
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -119,10 +119,18 @@ CFG_WITH_PAGER ?= y
 CFG_WITH_SOFTWARE_PRNG ?= y
 endif # CFG_STM32MP15
 
+ifeq ($(CFG_WITH_PAGER),y)
+CFG_WITH_LPAE ?= n
+endif
 CFG_WITH_LPAE ?= y
 CFG_MMAP_REGIONS ?= 23
 CFG_DTB_MAX_SIZE ?= (256 * 1024)
 CFG_CORE_ASLR ?= n
+
+ifneq ($(CFG_WITH_LPAE),y)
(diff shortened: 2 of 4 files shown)


GseoC:
For ST-specific commits:

core/arch/arm/mm/tee_pager.c
```diff
@@ -865,7 +865,6 @@ static void split_region(struct vm_paged_region *reg,
 
 	pager_unlock(exceptions);
 }
-DECLARE_KEEP_PAGER(split_region);
```
jenswi-linaro, line 868: `DECLARE_KEEP_PAGER()` is used here to make sure that we don't get pagefault for this code while holing the pager lock, same below.

core/arch/arm/mm/core_mmu_v7.c
```diff
@@ -794,7 +794,7 @@ void core_init_mmu_regs(struct core_mmu_config *cfg)
 	 */
 	cfg->ttbcr = TTBCR_N_VALUE;
 }
-DECLARE_KEEP_PAGER(core_init_mmu_regs);
+DECLARE_KEEP_INIT(core_init_mmu_regs);
```
jenswi-linaro, line 797: I wonder if we could remove `DECLARE_KEEP_INIT()`, compare with the other implementations of this function.

core/arch/arm/sm/sm_a32.S
```diff
@@ -406,7 +406,7 @@ FUNC sm_init , :
 
 	bx	lr
 END_FUNC sm_init
-DECLARE_KEEP_PAGER sm_init
+DECLARE_KEEP_INIT sm_init
```
jenswi-linaro, line 409: We can probably remove `DECLARE_KEEP_INIT()` too. It will however still have to remain in the unpaged area since secondary CPUs may initialize after the primary CPU has started paging.

core/arch/arm/mm/tee_pager.c
etienne-lms: Ok, i'll have a second look. Thanks.

core/arch/arm/sm/sm_a32.S
etienne-lms: right!

core/arch/arm/mm/core_mmu_v7.c
etienne-lms: You're right. `core_init_mmu_map()` makes it an init section already. I'll update.

etienne-lms:
Squashed and apply @GseoC tag on stm32mp1 commits.
Updated core patch to only remove keep pager directive on `core_init_mmu_regs()`.

jenswi-linaro:
For the commit "core: remove keep pager directive on core_init_mmu_regs()" please apply:

etienne-lms:
tag applied

PR #5882, opened 2023-03-13
libutee: increase MPI_MEMPOOL_SIZE to 14Kb

Narflex:
With the Widevine v17/v18 OPKs using their new Provisioning 4.0 technique, this pool size needs to be increased to 14Kb from 12Kb.

The sequence that is being executed that requires this is as follows:
1. TEE_GenerateKey(key_handle, 2048, NULL, 0)
2. TEE_GetObjectBufferAttribute(key, TEE_ATTR_RSA_MODULUS, modulus_data, &modulus_len), same for TEE_ATTR_RSA_PUBLIC_EXPONENT and TEE_ATTR_RSA_PRIVATE_EXPONENT.
3. mbedtls_rsa_complete() on a pk object created from the extracted modulus, public exp and private exp.

### DIFF
--- a/lib/libutee/tee_api_arith_mpi.c
+++ b/lib/libutee/tee_api_arith_mpi.c
@@ -13,7 +13,7 @@
 #include <utee_syscalls.h>
 #include <util.h>
 
-#define MPI_MEMPOOL_SIZE	(12 * 1024)
+#define MPI_MEMPOOL_SIZE	(14 * 1024)
 
 static void __noreturn api_panic(const char *func, int line, const char *msg)
 {


jforissier (approved):

jenswi-linaro:

etienne-lms:

Narflex:
Not entirely sure, but based on the documentation...this may be approved at this point. Is that correct? (so then I'd just need to add all the Acked-by lines to the commit message and re-push it and then it can get merged)

jenswi-linaro:
> Not entirely sure, but based on the documentation...this may be approved at this point. Is that correct? (so then I'd just need to add all the Acked-by lines to the commit message and re-push it and then it can get merged)

Correct.

Narflex:
OK, should be ready for merge now, thanks!

PR #5886, opened 2023-03-15
core: crypto: fix pkcs_1_v1_5_decode when empty message

SOuajih:
In case of EME-PKCS1-v1_5 decoding, the encoded message format is as follow : EM = 0x00 || 0x02 || PS || 0x00 || M. When using an empty message, the 0x00 octet that separates the padding string and message is located at the end. Thus, update the condition to pass the check in case of empty message.

This fixes the following AOSP cts test:
Module: CtsKeystoreTestCases
Test: testEmptyPlaintextEncryptsAndDecrypts
Link: https://android.googlesource.com/platform/cts/+/refs/tags/android-cts-12.0_r6/tests/tests/keystore/src/android/keystore/cts/CipherTest.java

### DIFF
--- a/core/lib/libtomcrypt/src/pk/pkcs1/pkcs_1_v1_5_decode.c
+++ b/core/lib/libtomcrypt/src/pk/pkcs1/pkcs_1_v1_5_decode.c
@@ -58,7 +58,7 @@ int pkcs_1_v1_5_decode(const unsigned char *msg,
     }
     ps_len = i++ - 2;
 
-    if (i >= modulus_len) {
+    if (i > modulus_len) {
       /* There was no octet with hexadecimal value 0x00 to separate ps from m.
        */
       result = CRYPT_INVALID_PACKET;


jforissier:
Hello @SOuajih,

Thanks for the patch. Two things:
- This is a change to the libtomcrypt library itself, so the commit subject would be better as: "libtomcrypt: fix pkcs_1_v1_5_decode() when empty message" ("libtomcrypt:" rather than "core: crypto:"; also note the parentheses after the function name).
- Have you considered submitting this change upstream (https://github.com/libtom/libtomcrypt)?

SOuajih:
> Hello @SOuajih,
> 
> Thanks for the patch. Two things:
> 
>     * This is a change to the libtomcrypt library itself, so the commit subject would be better as: "libtomcrypt: fix pkcs_1_v1_5_decode() when empty message" ("libtomcrypt:" rather than "core: crypto:"; also note the parentheses after the function name).
> 

Hello, 
Thank you for the review, 
Done

>     * Have you considered submitting this change upstream (https://github.com/libtom/libtomcrypt)?

Thank you, I was not aware of this, do you think this should be submitted upstream to libtomcrypt before ?

jforissier:
> Thank you, I was not aware of this, do you think this should be submitted upstream to libtomcrypt before ?

Yes it would be good, for the additional review that libtomcrypt experts can provide, and also to prevent our local copy from diverging too much. Thanks!

jforissier:
This commit was accepted upstream (https://github.com/libtom/libtomcrypt/commit/caf35002883365e2bd61e9ab6ccc78ad584048a4) so I have merged it. Thanks!

PR #636, opened 2016-02-02
Separate pager from abort handling

jenswi-linaro:
Refactoring abort handling to make it easier to integrate the coming hard-float support which needs special treatment in aborts.

The commits are intended to be reviewed one by one.

PR #5898, opened 2023-03-20
core: ffa: Allow multiple SPs with same UUID

jellesels-arm:
The FF-A spec doesn't forbid multiple SPs to have the same UUID. This makes it possible to use the FF-A UUID as a identifier for the protocol on top of the FF-A layer.
To achieve this we have to make sure that the FFA_PARTITION_INFO_GET can return more then one endpoint id if we pass a UUID. To make sure that there is no collision between the SP binaries names, we distinguish between the FF-A UUID and the SP UUID. The SP UUID is used to identify the SP itself. While the FF-A UUID is used as part of the FF-A protocol.

Build changes can be found [#636](https://github.com/OP-TEE/build/pull/636)
This is a follow-up on [#5650](https://github.com/OP-TEE/build/pull/603)

### DIFF
--- a/core/arch/arm/include/kernel/secure_partition.h
+++ b/core/arch/arm/include/kernel/secure_partition.h
@@ -43,6 +43,7 @@ struct sp_session {
 	unsigned int spinlock;
 	const void *fdt;
 	bool is_initialized;
+	TEE_UUID ffa_uuid;
 	TAILQ_ENTRY(sp_session) link;
 };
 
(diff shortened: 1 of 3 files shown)


jforissier (commented):

jenswi-linaro:
In the commit message you say:
_The FF-A spec doesn't forbid_
That sounds a bit negative, how about:
_The FF-A spec allows_
instead?

Anyway, please apply:

PR #5900, opened 2023-03-20
core: spmc: fix FIP SP loading

balint-dobszay-arm:
The memory management in process_sp_pkg() function contains errors. It tries to add new mappings for the SP packages that reside in the TA_RAM PA range, but this range is already mapped so this is unnecessary and wrong. Fix the code by simply using phys_to_virt() instead.

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -1590,6 +1590,12 @@ static TEE_Result sp_init_all(void)
 		}
 	}
 
+	/*
+	 * At this point all FIP SPs are loaded by ldelf or by the raw binary SP
+	 * loader, so the original images (loaded by BL2) are not needed anymore
+	 */
+	fip_sp_deinit_all();
+
 	/* Continue the initialization and run the SP */
 	TAILQ_FOREACH(s, &open_sp_sessions, link) {
 		res = sp_first_run(s);
@@ -1600,12 +1606,6 @@ static TEE_Result sp_init_all(void)
 				panic();
 		}
 	}
-
-	/*
-	 * At this point all FIP SPs are loaded by ldelf or by the raw binary SP
-	 * loader, so the original images (loaded by BL2) are not needed anymore
-	 */
-	fip_sp_deinit_all();
 
 	return TEE_SUCCESS;
 }


core/arch/arm/kernel/secure_partition.c
```diff
@@ -1645,11 +1612,6 @@ static TEE_Result sp_init_all(void)
 				panic();
 		}
 	}
-	/*
-	 * At this point all FIP SPs are loaded by ldelf so the original images
-	 * (loaded by BL2 earlier) can be unmapped
-	 */
-	fip_sp_unmap_all();
```
jforissier, line 1652: IIUC moving this call up (before the the `sp_first_run()` calls) is unrelated to this fix, in which case it would be better to create a separate commit.

jforissier (commented):
Hi @balint-dobszay-arm,

One small comment, in any case LGTM:

core/arch/arm/kernel/secure_partition.c
balint-dobszay-arm: Okay, I'll move this to a separate commit.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1418,111 +1418,76 @@ static const struct ts_ops sp_ops = {
...
-	tee_mm_entry_t *mm = NULL;
 	struct fip_sp *sp = NULL;
 	uint64_t sp_fdt_end = 0;
 	size_t sp_pkg_size = 0;
 	vaddr_t sp_pkg_va = 0;
-	size_t num_pages = 0;
 
-	/* Map only the first page of the SP package to parse the header */
+	/* Check the first page which contains the SP package header */
 	if (!tee_pbuf_is_sec(sp_pkg_pa, SMALL_PAGE_SIZE))
```
jenswi-linaro, line 1429: This isn't needed, `phys_to_virt()` below will fail if the PA isn't in the expected range.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1418,111 +1418,76 @@ static const struct ts_ops sp_ops = {
...
-		goto err_unmap;
+		return TEE_ERROR_BAD_FORMAT;
 	}
 
-	core_mmu_unmap_pages(sp_pkg_va, 1);
-	tee_mm_free(mm);
-
-	/* Map the whole package */
+	/* Check the whole SP package now that the size is known */
 	if (!tee_pbuf_is_sec(sp_pkg_pa, sp_pkg_size))
```
jenswi-linaro, line 1464: This is also taken care of implicitly by `phys_to_virt()` below.

jenswi-linaro (commented):
With the comments addressed:

core/arch/arm/kernel/secure_partition.c
balint-dobszay-arm: Thanks, fixed.

core/arch/arm/kernel/secure_partition.c
balint-dobszay-arm: Fixed

balint-dobszay-arm:
Thanks for the review, all comments addressed and tags applied.

PR #5904, opened 2023-03-22
core: dt: add kernel DT API functions and cleaning

GseoC:
Adds _fdt_read_uint32_index() device-tree API to easily ready at an index into a multi-value property.

This completes _fdt_read_uint32_array() and _fdt_read_uint32()

Disclaimer:
New API is not used ATM but could be by anyone interested. The last past is simply cleaning the header file

Edit:
Also adds _fdt_get_reg_props_by_index() and _fdt_get_reg_props_by_name()

Removes _fdt_check_node() ghost prototype

Edit2:

Replaced all  `_fdt_`  prefixes by  fdt_

### DIFF
--- a/core/drivers/clk/clk-stm32mp13.c
+++ b/core/drivers/clk/clk-stm32mp13.c
@@ -1439,7 +1439,7 @@ static int clk_stm32_parse_oscillator_fdt(const void *fdt, int node,
 			return ret;
 
 		if (strncmp(cchar, name, (size_t)ret) ||
-		    _fdt_get_status(fdt, subnode) == DT_STATUS_DISABLED)
+		    fdt_get_status(fdt, subnode) == DT_STATUS_DISABLED)
 			continue;
 
 		cuint = fdt_getprop(fdt, subnode, "clock-frequency", &ret);
@@ -1457,8 +1457,8 @@ static int clk_stm32_parse_oscillator_fdt(const void *fdt, int node,
 		if (fdt_getprop(fdt, subnode, "st,css", NULL))
 			osci->css = true;
 
-		osci->drive = _fdt_read_uint32_default(fdt, subnode, "st,drive",
-						       LSEDRV_MEDIUM_HIGH);
+		osci->drive = fdt_read_uint32_default(fdt, subnode, "st,drive",
+						      LSEDRV_MEDIUM_HIGH);
 
 		return 0;
 	}
@@ -1498,13 +1498,13 @@ static int clk_stm32_load_vco_config_fdt(const void *fdt, int subnode,
 {
 	int ret = 0;
 
-	ret = _fdt_read_uint32_array(fdt, subnode, "divmn", vco->div_mn,
-				     PLL_DIV_MN_NB);
+	ret = fdt_read_uint32_array(fdt, subnode, "divmn", vco->div_mn,
+				    PLL_DIV_MN_NB);
 	if (ret != 0)
 		return ret;
 
-	ret = _fdt_read_uint32_array(fdt, subnode, "csg", vco->csg,
-				     PLL_CSG_NB);
+	ret = fdt_read_uint32_array(fdt, subnode, "csg", vco->csg,
+				    PLL_CSG_NB);
 
 	vco->csg_enabled = (ret == 0);
 
@@ -1517,9 +1517,9 @@ static int clk_stm32_load_vco_config_fdt(const void *fdt, int subnode,
 	vco->status = RCC_PLLNCR_DIVPEN | RCC_PLLNCR_DIVQEN |
 		      RCC_PLLNCR_DIVREN | RCC_PLLNCR_PLLON;
 
-	vco->frac = _fdt_read_uint32_default(fdt, subnode, "frac", 0);
+	vco->frac = fdt_read_uint32_default(fdt, subnode, "frac", 0);
 
-	vco->src = _fdt_read_uint32_default(fdt, subnode, "src", UINT32_MAX);
+	vco->src = fdt_read_uint32_default(fdt, subnode, "src", UINT32_MAX);
 
 	return 0;
 }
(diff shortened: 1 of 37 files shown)


jenswi-linaro:
Why do we use these "_fdt_" prefixes?

jforissier:
> Why do we use these `_fdt_` prefixes?

IIRC I'm the one who introduced that naming convention. `fdt_` is the prefix used by libfdt. I used `_fdt_` for simple helper function that do very similar things than the `fdt_` ones, just simple extensions, to mean "that should really belong in libfdt". Then I used `dt_` for more complex, OP-TEE specific manipulations.

But TBH this doesn't make much sense and I'll be happy with a unified convention, i.e., `dt_` everywhere. What do you think?

jenswi-linaro:
Thanks for the background. Yes, I would also prefer `dt_` everywhere.
If there's an intent to extend libfdt please upstream those functions with proper function names.

GseoC:
> Thanks for the background. Yes, I would also prefer `dt_` everywhere. If there's an intent to extend libfdt please upstream those functions with proper function names.

Hello,
I'll come back with dt_ prefix then :)

etienne-lms:
Note the error code returned by `fdt_xxx()` and `_fdt_xxx()` fucntion are libfdt error number, which is quite convient.
There are `dt_xxx()` function that return `TEE_Result` codes. I fear confusions in these return codes.

jforissier:
@etienne-lms good point...

jforissier:
> How about just dropping the `_` prefix then? There's something special about functions having a `_` prefix suggesting that it only should be used if you know what you're doing or at least with a warning flag.

Works for me.

GseoC:
Forced-pushed with Edit 2 modifications (See P-R description)

core/include/kernel/dt.h
```diff
@@ -214,6 +214,14 @@ void fdt_fill_device_info(const void *fdt, struct dt_node_info *info,
 int fdt_read_uint32_array(const void *fdt, int node, const char *prop_name,
 			  uint32_t *array, size_t count);
 
+/*
+ * Read one cell from a given multi-value property of the given node.
+ * Returns 0 on success, or a negative
```
etienne-lms, line 219: unwrap line

core/include/kernel/dt.h
```diff
@@ -303,6 +311,14 @@ static inline uint32_t fdt_read_uint32_default(const void *fdt __unused,
 	return dflt_value;
 }
 
+static inline int fdt_read_uint32_index(const void *fdt __unused,
+					int node __unused,
+					const char *prop_name __unused,
+					int index __unused,
+					uint32_t *value __unused) {
```
etienne-lms, line 326: new line char before brace `{`

core/kernel/dt.c
```diff
@@ -316,6 +316,24 @@ int fdt_read_uint32_array(const void *fdt, int node, const char *prop_name,
...
 
+int fdt_read_uint32_index(const void *fdt, int node, const char *prop_name,
+			  int index, uint32_t *value)
+{
+	const fdt32_t *cuint = NULL;
+	int len = 0;
+
+	cuint = fdt_getprop(fdt, node, prop_name, &len);
+	if (!cuint)
+		return -FDT_ERR_NOTFOUND;
```
etienne-lms, line 334: `return len;`

core/kernel/dt.c
```diff
@@ -356,3 +356,61 @@ uint32_t fdt_read_uint32_default(const void *fdt, int node,
...
 	return ret;
 }
+
+int fdt_get_reg_props_by_index(const void *fdt, int node, int index,
+			       paddr_t *base, size_t *size)
+{
+	const fdt32_t *prop = NULL;
+	int parent = 0;
+	int len = 0;
+	int ac = 0;
```
etienne-lms, line 366: s/ac/address_cells/g
s/sc/size_cells/g
(all lines fit in 80char :)

core/drivers/clk/clk-stm32mp13.c
```diff
@@ -1517,17 +1517,17 @@ static int clk_stm32_load_vco_config_fdt(const void *fdt, int subnode,
...
+	vco->src = fdt_read_uint32_default(fdt, subnode, "src", UINT32_MAX);
 
 	return 0;
 }
 
 static int clk_stm32_load_output_config_fdt(const void *fdt, int subnode,
 					    struct stm32_pll_output *output)
 {
-	return _fdt_read_uint32_array(fdt, subnode, "st,pll_div_pqr",
+	return fdt_read_uint32_array(fdt, subnode, "st,pll_div_pqr",
```
etienne-lms, line 1530: indentation below.

etienne-lms (commented):
Commit "core: dt: add fdt_read_size() API" does not export `fdt_read_size()`. It rather makes a local helper fucntion. Could you update the  commit message?

- "core: replace _fdt_ prefix with fdt_ for device tree API" with comment addressed.
- "core: dt: remove fdt_check_node()".

core/drivers/clk/clk-stm32mp13.c
GseoC: Ok, I still apply your tag on the patch and make the fixup

core/kernel/dt.c
GseoC: I added another patch to fix it in fdt_read_uint32_array() as well

GseoC:
Force-pushed with fixups made and comments applied. I made an extra patch to fix return value in fdt_read_uint32_array(). Tags applied

etienne-lms:
"core: dt: add kernel DT API fdt_read_uint32_index()",
"core: dt: add fdt_read_size() helper",
"core: dt: add kernel DT API to retrieve regs by name",
and "core: dt: fix return value in fdt_read_uint32_array()".

GseoC:
Fixups squashed and tags applied. Thank you for the review

PR #5911, opened 2023-03-24
se050: ecc: SE050-F shared secret

ldts:
The SE050-F does not support shared secret generation. 
This commit allows the operation to fallback to its software implementation.

Fixes: regression 4009
Note: sw fallback will only execute if the private key is _not _ in the secure element (obviously)

### DIFF
--- a/core/drivers/crypto/se050/core/ecc.c
+++ b/core/drivers/crypto/se050/core/ecc.c
@@ -36,9 +36,6 @@ static bool oefid_key_supported(size_t bits)
 
 static bool oefid_algo_supported(uint32_t algo)
 {
-	if (!algo)
-		return true;
-
 	switch (se050_get_oefid()) {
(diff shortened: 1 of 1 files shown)


jforissier (commented):

ldts:
thanks @jforissier 
I noticed that there were changes in this file (sorry for not having reviewed, been a bit busy over the past few weeks). Now back to normal.

jforissier:
@ldts if this fixes a regression introduced by a recent change it would good to mention the commit in the description (with a `Fixes:` tag).

ldts:
@jforissier ping

jforissier:
> @jforissier ping

Merged manually with two minor fixes to the commit description: I removed the blank line between the tags and changed `xtest regression 4009` to `xtest regression 4009` since it is the actual command that need to be run. 

Thanks!

PR #5921, opened 2023-03-28
ci: compile-test as many PTAs as possible on QEMU/QEMUv8

jforissier:
Enable as many PTAs as possible in the QEMU/QEMUv8 CI builds in order to catch compile issues. Some PTAs are not applicable to QEMU though.


jforissier:
Rebased onto master, CI should pass now.

jenswi-linaro:

PR #5922, opened 2023-03-28
core: imx: remove duplicate driver_init() call

clementfaure:
Remove duplicated call to driver_init().
The previous driver_init() call would only initialize the driver. The new driver_init() call initializes the driver and its power management callback.

Fixes: 97eb916803fe ("drivers: imx: tzc380: re-configure TZ380 upon PM resume")


jenswi-linaro (commented):

PR #5925, opened 2023-03-29
ta: collated ta compilation

liushiwei007:
Sort out the common code compiled by ta,
abstract RISCV and ARM compile the common part of ta.

Reviewed-by: chenchaokai <chenchaokai@eswincomputing.com>


ta/mk/build-user-ta.mk
```diff
@@ -66,7 +66,7 @@ additional-compile-deps := $(ta_dev_kit-files-include)
 include mk/compile.mk
 # Install TA libraries before in-tree TAs can be linked
 additional-link-deps := $(ta_dev_kit-files-lib)
-include  ta/arch/$(ARCH)/link.mk
+include  ta/arch/link.mk
```
jforissier, line 69: This does not make much sense does it? If it's under `arch` then it applies to some specific architecture.
Same for `link_shlib.mk` and `user_ta_header.c`. I suggest moving them higher up into `ta/`.

liushiwei007: > This does not make much sense does it? If it's under `arch` then it applies to some specific architecture. Same for `link_shlib.mk` and `user_ta_header.c`. I suggest moving them higher up into `ta/`.

done

jforissier (approved):

Please replace `ta` with `TA` in the subject and description except for the `ta:` subsystem prefix which is fine in lowercase

jforissier:
And perhaps `build: ta: add RISC-V linker script` would be more descriptive?

Issue #5905, opened 2023-03-22
OP-TEE core exception with TA memstats


weizhaojiang:
General guidance when creating issues:
I enable `CFG_TA_STATS ?= y` to have ta memory dump. It can work in most cases expect below one:
Shell1 execute:
console# xtest 6016
Shell2 execute:
console# xtest --stats --ta

optee core will data abort with below log:
[   35.607698]<6> (6)[2254:xtest]E/TC:6 2 Core data-abort at address 0xffffffffffffff80 (translation fault)

The issue can be reproduced in latest version of master.

I debugged on it and found happens when core try to switch the user ta to its user context. However, the usr_stack from context is 0 – this is a fatal error. After that it does a  ROUNDUP() and get the fake stack pointer – 0xffffffffffffff80(64bits). Data abort happens when core access the user ta's stack.

Any suggestion to figure out why the stack pointer will be changed to 0 in this case?

weizhaojiang:
With below patch, the crash will be gone. But still like to figure out why it gets 0 stack_pointer.

`@@ -151,11 +151,30 @@ static TEE_Result user_ta_enter(struct ts_session *session,
                        goto out;
        }

+       /* Make room for usr_params at top of stack */
+       usr_stack = utc->uctx.stack_ptr;
+       DMSG("!!utc->uctx.stack_ptr = 0x%x", utc->uctx.stack_ptr);
+       if (usr_stack == 0) {
+               res = TEE_ERROR_CORRUPT_OBJECT;
+               DMSG("!!!!!!session = 0x%x", session);
+               DMSG("!!!!!!utc = 0x%x", utc);
+               DMSG("!!!!!!usr_stack == 0");
+               DMSG("!!!!!!utc->uctx.is_initializing = %d", utc->uctx.is_initializing);
+               DMSG("!!!!!!utc->uctx.is_32bit = %d", utc->uctx.is_32bit);
+               DMSG("!!!!!!utc->uctx.ts_ctx = 0x%x", utc->uctx.ts_ctx);
+               DMSG("!!!!!!utc->uctx.dl_entry_func = 0x%x", utc->uctx.dl_entry_func);
+               DMSG("!!!!!!utc->uctx.load_addr = 0x%x", utc->uctx.load_addr);
+               struct tee_ta_ctx *ta_ctx = &utc->ta_ctx;
+               DMSG("!!!!!!ta_ctx->panicked = %d", ta_ctx->panicked);
[...]

PR #5926, opened 2023-03-29
core: check TA initialization completion before dump it

weizhaojiang:
Problem: In some concurrent cases, ta dump will try to dump a TA which has not completed TA initialization and the TA stack pointer isn't set. That causes a data abort when accessing its stack.

Solution: Check the user TA initialization is completed or not. If it is still being initialized, return TEE_ERROR_BAD_STATE. https://github.com/OP-TEE/optee_os/issues/5905

### DIFF
--- a/core/kernel/tee_ta_manager.c
+++ b/core/kernel/tee_ta_manager.c
@@ -854,6 +854,10 @@ static TEE_Result dump_ta_memstats(struct tee_ta_session *s,
 	if (!ts_ctx)
 		return TEE_ERROR_ITEM_NOT_FOUND;
 
+	if (is_user_ta_ctx(ts_ctx) &&
+	    to_user_ta_ctx(ts_ctx)->uctx.is_initializing)
+		return TEE_ERROR_BAD_STATE;
+
 	ctx = ts_to_ta_ctx(ts_ctx);
 
 	if (ctx->panicked)


jforissier (commented):
Hi @weizhaojiang,

This LGTM. Please change the commit subject to "core: dump_ta_memstats(): check TA initialization completion before accessing it" and s/ta dump/TA dump/ in the description.

With that:

core/kernel/tee_ta_manager.c
```diff
@@ -854,6 +854,13 @@ static TEE_Result dump_ta_memstats(struct tee_ta_session *s,
 	if (!ts_ctx)
 		return TEE_ERROR_ITEM_NOT_FOUND;
 
+	/*
+	 * if the user TA is still being initialized,
+	 * return BAD_STATE to avoid accessing it.
+	 */
+	if (to_user_ta_ctx(ts_ctx)->uctx.is_initializing)
```
jenswi-linaro, line 861: You need to check that `ts_ctx` really is a user_ta_ctx with `is_user_ta_ctx()` first.

core/kernel/tee_ta_manager.c
```diff
@@ -854,6 +854,13 @@ static TEE_Result dump_ta_memstats(struct tee_ta_session *s,
 	if (!ts_ctx)
 		return TEE_ERROR_ITEM_NOT_FOUND;
 
+	/*
+	 * if the user TA is still being initialized,
```
etienne-lms, line 858: (nit) s/if/If/
maybe even remove the inline comment: the implementation is explicit.

weizhaojiang: thanks for suggestion. done

jenswi-linaro:

PR #5931, opened 2023-03-31
CAAM compilation fix

clementfaure:
Hello,

Here is a quick rework for the CAAM `crypto.mk` that was IMO cumbersome to read.
@ldts Let me know that configuration is still compatible with SE050 configuration.
With that, I did some minor fixes on CAAM compilation flags.

Thanks!


jenswi-linaro:

clementfaure:
Hi @jenswi-linaro ,
I removed the commit for the `crypto.mk` re-work. There are some CAAM configuration that are not compatible, especially when CAAM capabilities are not all enabled. I need to give more thought.

Tag applied for the other commits, since there is no change.

PR #5933, opened 2023-04-01
core: mm: Fix idx truncation bug

LeiSen62:
When the idx of TA memory mapping is assigned, it is saved in the l1_idx_t, which is the uint8_t or uint16_t type. But when it is parsed, it is save in the only uint8_t that will lead to error. To solve this problem, the idx should be saved in the uint16_t type when parsing the idx.

Signed-off-by: LeiSen <leisen1@huawei.com>

### DIFF
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -1138,7 +1138,7 @@ static uint64_t *core_mmu_get_user_mapping_entry(struct mmu_partition *prtn,
 						 unsigned int base_idx)
 {
 #if (CORE_MMU_BASE_TABLE_LEVEL == 0)
-	uint8_t idx = 0;
+	l1_idx_t idx = 0;
 	uint64_t *tbl = NULL;
 #endif
 


core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -1138,7 +1138,7 @@ static uint64_t *core_mmu_get_user_mapping_entry(struct mmu_partition *prtn,
 						 unsigned int base_idx)
 {
 #if (CORE_MMU_BASE_TABLE_LEVEL == 0)
-	uint8_t idx = 0;
+	uint16_t idx = 0;
```
jforissier, line 1141: How about using `l1_idx_t` instead?

LeiSen62: That`s a good idea

jforissier:
With `l1_idx_t` instead of `uint16_t`:

core/arch/arm/mm/core_mmu_lpae.c
LeiSen62: That`s a good idea.

jenswi-linaro:

jforissier:
Tags applied and merged. Thanks.

PR #5666, opened 2022-11-23
crypto_api: add ecc encrypt and decrypt to crypto

yuzexiyzx:
1. add ecc encrypt to ecc_public_key_ops and add ecc decrypt to ecc_keypair_ops.
2. define struct drvcrypt_encrypt_data

Signed-off-by: yuzexi <yuzexi@hisilicon.com>
Reviewed-by: Xiaoxu Zeng <zengxiaoxu@huawei.com>

PR #5937, opened 2023-04-03
CAAM SM2 encrypt/decrypt fallback

clementfaure:
Hello,

With the recent SM2 encryption and decryption operations in the crypto driver API: https://github.com/OP-TEE/optee_os/pull/5666, the SM2 crypto tests in xtest (`regression_4006.43`) fail when the CAAM is enabled.
The CAAM does not implement (`.encrypt` and `.decrypt` operations).

The ECC software fallback mechanism could be used as suggested in this PR, but it requires some modification in the crypto driver API to make it work correctly.
Even if I do understand that this SW mechanism is needed in some cases/specific configuration, I'm not entirely convinced by how it works. The fallback implementation in the CAAM might also be invalidated as the black key implementation will be submitted soon.
So ... I'm also tempted to just set `CFG_CRYPTO_SM2_PKE` to `n` if the CAAM driver is enabled.

Thoughts?


jenswi-linaro:
> I'm also tempted to just set `CFG_CRYPTO_SM2_PKE` to `n` if the CAAM driver is enabled.

It makes sense to give it some time to work out a good solution.

clementfaure:
> > I'm also tempted to just set `CFG_CRYPTO_SM2_PKE` to `n` if the CAAM driver is enabled.
> 
> It makes sense to give it some time to work out a good solution.

Let's do that then

PR #5940, opened 2023-04-05
drivers: caam: disable CFG_CRYPTO_SM2_* when ECC CAAM driver is enabled

clementfaure:
Follow-up : https://github.com/OP-TEE/optee_os/pull/5937

### DIFF
--- a/core/crypto.mk
+++ b/core/crypto.mk
@@ -6,6 +6,13 @@ CFG_CRYPTO_SIZE_OPTIMIZATION ?= y
 
 ifeq (y,$(CFG_CRYPTO))
 
+###############################################################
+# Platform crypto-driver configuration. It has a higher priority over the
+# generic crypto configuration below.
+###############################################################
+CRYPTO_MAKEFILES := $(sort $(wildcard core/drivers/crypto/*/crypto.mk))
+include $(CRYPTO_MAKEFILES)
+
 # Ciphers
 CFG_CRYPTO_AES ?= y
 CFG_CRYPTO_DES ?= y
@@ -283,12 +290,6 @@ _CFG_CORE_LTC_MAC := $(call ltc-one-enabled, HMAC CMAC CBC_MAC)
 _CFG_CORE_LTC_CBC := $(call ltc-one-enabled, CBC CBC_MAC)
 _CFG_CORE_LTC_ASN1 := $(call ltc-one-enabled, RSA DSA ECC)
 _CFG_CORE_LTC_EC25519 := $(call ltc-one-enabled, ED25519 X25519)
-
-###############################################################
-# Platform independent crypto-driver configuration
-###############################################################
-CRYPTO_MAKEFILES := $(sort $(wildcard core/drivers/crypto/*/crypto.mk))
-include $(CRYPTO_MAKEFILES)
 
 # Enable TEE_ALG_RSASSA_PKCS1_V1_5 algorithm for signing with PKCS#1 v1.5 EMSA
 # without ASN.1 around the hash.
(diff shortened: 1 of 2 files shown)


core/drivers/crypto/caam/crypto.mk
```diff
@@ -175,5 +175,11 @@ $(call force, CFG_NXP_CAAM_ACIPHER_DRV, $(call cryphw-one-enabled, RSA ECC DH DS
 $(call force, CFG_CRYPTO_DRV_MAC, $(call cryphw-one-enabled, HMAC CMAC))
 CFG_CRYPTO_DRV_ACIPHER ?= $(CFG_NXP_CAAM_ACIPHER_DRV)
 
+# Disable SM2 as it is not supported by the CAAM driver
+ifeq ($(CFG_NXP_CAAM_ECC_DRV),y)
+CFG_CRYPTO_SM2_PKE=n
+CFG_CRYPTO_SM2_DSA=n
```
jforissier, line 181: `$(call force,CFG_xx,n)`

https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#there-are-only-two-ways

jenswi-linaro:
The default value of `CFG_CRYPTO_SM2_PKE` is already assigned in `core/crypto.mk`. It looks like we have a problem with the order values are assigned.

clementfaure:
I didn't anticipate that ...
Would it make sense to move
```
CRYPTO_MAKEFILES := $(sort $(wildcard core/drivers/crypto/*/crypto.mk))
include $(CRYPTO_MAKEFILES)
```
to the top of `mk/crypto.mk` ? To give the HW crypto driver the priority over the global crypto configuration?

jforissier:
@clementfaure I think it would make sense indeed. Platform-specific settings should always have higher priority.

clementfaure:
Force-push
* Rebase
* Add [d2f3139](https://github.com/OP-TEE/optee_os/pull/5940/commits/d2f31393d23090486dd94c7095a0ef83d4cbfd97)
* Add `CFG_CRYPTO_SM2_KEP=n`

jforissier:
Perhaps the second commit could mention it is a temporary fix until proper software fallback is implemented. Anyways for both commits:

core/crypto.mk
```diff
@@ -6,6 +6,13 @@ CFG_CRYPTO_SIZE_OPTIMIZATION ?= y
 
 ifeq (y,$(CFG_CRYPTO))
 
+###############################################################
+# Platform independent crypto-driver configuration
```
jenswi-linaro, line 10: In the commit message, you say that the platform crypto configuration should have higher priority than the generic stuff, yet we talk about platform-independent stuff here. I can't make sense of it.

clementfaure: Would that make more sense?
```
###############################################################
# Platform crypto-driver configuration. It has a higher priority over the
# generic crypto configuration below.
###############################################################
```

jenswi-linaro: Yes, thanks.

jenswi-linaro:
With my comment addressed:

jforissier:
This being a fix, I am tempted to take it for 3.21.0, however I am a bit concerned that the first commit ("core: crypto: give the platform configuration a higher priority") might break some platforms/configurations (although it is definitely the right thing to do).

@clementfaure can this wait until after the release? I suppose you can set the proper `CFG_` values on the command line anyways?

clementfaure:
From an i.MX maintainer point of view, I would like to see this PR merged, of course. Especially since we will rebase our internal tree on top of this 3.21 release.
But I do get your point : that's an important change, especially for a RC1.
If the release is merged without the fix, we should at least mention that i.MX platforms compiled with the ECC CAAM driver should be compiled with `CFG_CRYPTO_SM2_{PKE,KEP,DSA}=n`

jforissier:
> If the release is merged without the fix, we should at least mention that i.MX platforms compiled with the ECC CAAM driver should be compiled with `CFG_CRYPTO_SM2_{PKE,KEP,DSA}=n`

Yes I'd rather do that. I will take this immediately after the release.

PR #5945, opened 2023-04-06
TZASC: Allow 32KB region size

philip-ro:
According to the ARM TZC-380 Technical Reference Manual, 32KB is the minimum region size [1]. But before this patch, tzc_auto_configure() allowed only 64KB as minimum.

(pow > 15) implies the following:
	region_size = (1ULL << pow) = (2^pow) > 32KB

After this patch, (pow >= 15) gives us region_size >= 32KB.

[1] TZC-380 TRM, Table 3.16 (https://developer.arm.com/documentation/ddi0431/c/programmers-model/register-descriptions/region-attributes--n--register?lang=en)

Tested on i.MX6UL
Signed-off-by: Philip Oberfichtner <pro@denx.de>

### DIFF
--- a/core/drivers/tzc380.c
+++ b/core/drivers/tzc380.c
@@ -242,7 +242,7 @@ int tzc_auto_configure(vaddr_t addr, vaddr_t size, uint32_t attr,
 	 */
 	pow = tzc.addr_width;
 
-	while (lsize != 0 && pow > 15) {
+	while (lsize != 0 && pow >= 15) {
 		region_size = 1ULL << pow;
 
 		/* Case region fits alignment and covers requested area */


jenswi-linaro:
@MrVan , do you have any comments?

clementfaure:
Hi @philip-ro,
Your commit matches what's in the RM. It's ok for me.

philip-ro:
Hi @clementfaure,

What I tested is to protect a 16KB region. This won't work before this patch with `pow = 16`, because of these lines:

https://github.com/OP-TEE/optee_os/blob/32b94ed4b63fdac094b1a6b7719f7cb8979695f6/core/drivers/tzc380.c#L263-L265

With `sub_region_size = 8KB` and `lsize = 16KB`.

After this patch the TZASC uses a 32KB region with multiple subregions to successfully cover the 16 KB.

jenswi-linaro:

jforissier:
Please use a `Link: [1] <url>` tag to fix the checkpatch error.

PR #5947, opened 2023-04-06
core: define a OP-TEE SMC ID for watchdog fast call

ldebieve:
Add a dedicated OP-TEE ID for a SMC fast call entry to manage the ARM watchdog service.

### DIFF
--- a/core/arch/arm/include/sm/optee_smc.h
+++ b/core/arch/arm/include/sm/optee_smc.h
@@ -597,11 +597,15 @@
  *
  * Return codes:
  * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_DISABLED - The watchdog is not enabled
  *
  * Command SMCWD_GET_TIMELEFT : Get time left
  * [in]		a1	Set to 4
+ * [out]	a1	The timeout value in seconds before watchdog expires
  *
  * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_DISABLED - The watchdog is not enabled
  * PSCI_RET_NOT_SUPPORTED - Function not supported
  *
  * Other commands
--- a/core/arch/arm/include/sm/optee_smc.h
+++ b/core/arch/arm/include/sm/optee_smc.h
@@ -560,36 +560,59 @@
 #define OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG	U(19)
 
 /*
- * Access to the secure watchdog.
+ * Secure watchdog management service
  *
  * Call requests usage:
- * a0	SMC Function ID, OPTEE_SMC_WATCHDOG
- * a1	Watchdog command
- *	0 : Init watchdog
- *	1 : Set timeout
- *	2 : Enable watchdog
- *	3 : Ping the watchdog for refresh
- *	4 : Get time left
- * a2	if a1 == 1, the timeout value to set
- *	if a1 == 2, 0 to stop the watchdog, 1 to enable it.
- *	preserved otherwise
- * a3-6	Not used
- * a7	Hypervisor Client ID register
+ * [in]		a0	SMC Function ID, OPTEE_SMC_WATCHDOG
+ * [out]	a0	PSCI error code return
  *
- * Normal return register usage:
- * a0 PSCI_RET_SUCCESS
- * a1	if a1 == 0, the minimal timeout value supported by watchdog
- *	if a1 == 4, the remaining time before watchdog expires
- *	preserved otherwise.
- * a2	if a1 == 0, the maximum timeout value supported by watchdog
- *	preserved otherwise.
- * a3-7	Preserved
+ * Watchdog commands
+ * Command SMCWD_INIT : Watchdog initialization
+ * [in]		a1	Set to 0
+ * [out]	a1	The minimal timeout value in seconds supported
+ *		a2	The maximal timeout value in seconds supported
  *
- * Error return:
- * a0	PSCI_RET_NOT_SUPPORTED		Function not supported
- *	PSCI_RET_INVALID_PARAMETERS	Invalid arguments
- *	PSCI_RET_INTERNAL_FAILURE	Device error
- * a1-7 Preserved
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_INTERNAL_FAILURE - Initialization failure
+ *
+ * Command SMCWD_SET_TIMEOUT : Watchdog set timeout
+ * [in]		a1	Set to 1
+ *		a2	The timeout value in seconds to set
+ *
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_INVALID_PARAMETERS - Incorrect input param
+ *
+ * Command SMCWD_ENABLE : Watchdog enable
+ * [in]		a1	Set to 2
+ *		a2	Set to 0 to stop the watchdog, 1 to enable it
+ *
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_INVALID_PARAMETERS - Incorrect input param
+ *
+ * Command SMCWD_PET : Ping the watchdog for refresh
+ * [in]		a1	Set to 3
+ *
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ *
+ * Command SMCWD_GET_TIMELEFT : Get time left
+ * [in]		a1	Set to 4
+ *
+ * Return codes:
+ * PSCI_RET_NOT_SUPPORTED - Function not supported
+ *
+ * Other commands
+ * [in]		a1	Other values
+ *
+ * Return codes:
+ * PSCI_RET_NOT_SUPPORTED - Function not supported
+ *
+ * a3-6		Not used
+ * a7		Hypervisor Client ID register
+ *
  *
  * By SMCCC convention:
  * PSCI_RET_SUCCESS, OPTEE_SMC_RETURN_OK and ARM_SMCCC_RET_SUCCESS
--- a/core/arch/arm/include/sm/optee_smc.h
+++ b/core/arch/arm/include/sm/optee_smc.h
@@ -563,18 +563,21 @@
  * Access to the secure watchdog.
  *
  * Call requests usage:
- * a0	SMC Function ID, OPTEE_SMC_FUNCID_WATCHDOG
+ * a0	SMC Function ID, OPTEE_SMC_WATCHDOG
  * a1	Watchdog command
  *	0 : Init watchdog
  *	1 : Set timeout
  *	2 : Enable watchdog
  *	3 : Ping the watchdog for refresh
  *	4 : Get time left
- * a2-6	Not use
+ * a2	if a1 == 1, the timeout value to set
+ *	if a1 == 2, 0 to stop the watchdog, 1 to enable it.
+ *	preserved otherwise
+ * a3-6	Not used
  * a7	Hypervisor Client ID register
  *
  * Normal return register usage:
- * a0	OPTEE_SMC_RETURN_OK
+ * a0 PSCI_RET_SUCCESS
  * a1	if a1 == 0, the minimal timeout value supported by watchdog
  *	if a1 == 4, the remaining time before watchdog expires
  *	preserved otherwise.
@@ -582,12 +585,21 @@
  *	preserved otherwise.
  * a3-7	Preserved
  *
- * Not supported return register usage:
- * a0	OPTEE_SMC_RETURN_UNKNOWN_FUNCTION
- * a1-7	Preserved
+ * Error return:
+ * a0	PSCI_RET_NOT_SUPPORTED		Function not supported
+ *	PSCI_RET_INVALID_PARAMETERS	Invalid arguments
+ *	PSCI_RET_INTERNAL_FAILURE	Device error
+ * a1-7 Preserved
+ *
+ * By SMCCC convention:
+ * PSCI_RET_SUCCESS, OPTEE_SMC_RETURN_OK and ARM_SMCCC_RET_SUCCESS
+ * are equal.
+ * PSCI_RET_NOT_SUPPORTED, OPTEE_SMC_RETURN_UNKNOWN_FUNCTION and
+ * ARM_SMCCC_RET_NOT_SUPPORTED are equal.
+ *
  */
 #define OPTEE_SMC_FUNCID_WATCHDOG		U(20)
-#define OPTEE_SMC_FUNCID_WATCHDOG_MASK \
+#define OPTEE_SMC_WATCHDOG \
 	OPTEE_SMC_FAST_CALL_VAL(OPTEE_SMC_FUNCID_WATCHDOG)
 
 /*
--- a/core/arch/arm/include/sm/optee_smc.h
+++ b/core/arch/arm/include/sm/optee_smc.h
@@ -560,6 +560,37 @@
 #define OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG	U(19)
 
 /*
+ * Access to the secure watchdog.
+ *
+ * Call requests usage:
+ * a0	SMC Function ID, OPTEE_SMC_FUNCID_WATCHDOG
+ * a1	Watchdog command
+ *	0 : Init watchdog
+ *	1 : Set timeout
+ *	2 : Enable watchdog
+ *	3 : Ping the watchdog for refresh
+ *	4 : Get time left
+ * a2-6	Not use
+ * a7	Hypervisor Client ID register
+ *
+ * Normal return register usage:
+ * a0	OPTEE_SMC_RETURN_OK
+ * a1	if a1 == 0, the minimal timeout value supported by watchdog
+ *	if a1 == 4, the remaining time before watchdog expires
+ *	preserved otherwise.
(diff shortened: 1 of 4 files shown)


core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,32 @@
...
+ *
+ * Call requests usage:
+ * a0	SMC Function ID, OPTEE_SMC_FUNCID_WATCHDOG
+ * a1	Watchdog command
+ *	0 : Init watchdog
+ *	1 : Set timeout
+ *	2 : Enable watchdog
+ *	3 : Ping the watchdog for refresh
+ *	4 : Get time left
+ * a2-6	Not used
```
jforissier, line 573: How is the timeout passed (command 1) and how is the time left returned (command 4)?

ldebieve: Sorry, missed some part, I'll update it quickly

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,37 @@
...
+ *
+ * Call requests usage:
+ * a0	SMC Function ID, OPTEE_SMC_FUNCID_WATCHDOG
+ * a1	Watchdog command
+ *	0 : Init watchdog
+ *	1 : Set timeout
+ *	2 : Enable watchdog
+ *	3 : Ping the watchdog for refresh
+ *	4 : Get time left
+ * a2-6	Not use
```
jforissier, line 573: Wrong. a2 is used to set the timeout when a0 == 1 and to enable/disable the watchdog when a0 == 2.
s/use/used/

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,37 @@
...
+ * a7	Hypervisor Client ID register
+ *
+ * Normal return register usage:
+ * a0	OPTEE_SMC_RETURN_OK
```
jforissier, line 577: a0 may also be `PSCI_RET_INTERNAL_FAILURE`, `PSCI_RET_INVALID_PARAMETER` or `PSCI_RET_NOT_SUPPORTED`.

etienne-lms: Should this description states that RET_OK and UNKNOWN_FONCTION IDs are SMCCC standard values that make this OPTEE_SMC_RETURN/PSCI_RET mix to be consistent?

```
 * Normal return register usage:
 * a0	PSCI_RET_SUCCESS
 * a1-7	Preserved
 *
 * Error return:
 * a0	PSCI_RET_NOT_SUPPORTED      Function not supported
 * 	PSCI_RET_INVALID_PARAMETERS Invalid arguments
 * 	PSCI_RET_INTERNAL_FAILURE   Device error
 * a1-7	Preserved
 *
 * By SMCCC convention:
 * PSCI_RET_SUCCESS, OPTEE_SMC_RETURN_OK and ARM_SMCCC_RET_SUCCESS
 * are equal.
 * PSCI_RET_NOT_SUPPORTED, OPTEE_SMC_RETURN_UNKNOWN_FUNCTION and  
 * ARM_SMCCC_RET_NOT_SUPPORTED are equal.
 */

**Review by etienne-lms — COMMENTED** [review]
In commit message header line: "core: arm: ..."

_mk/config.mk_ requires a value for  `CFG_WDT_SM_HANDLER_ID`. I think we need an extra switch like `CFG_WDT_SM_HANDLER_OPTEE_ID`.
```patch
ifeq (y-,$(CFG_WDT_SM_HANDLER)-$(CFG_WDT_SM_HANDLER_ID))
$(error CFG_WDT_SM_HANDLER_ID must be defined when enabling CFG_WDT_SM_HANDLER)
endif
```

core/arch/arm/include/sm/optee_smc.h
ldebieve: Agree, I'll add it.

core/arch/arm/include/sm/optee_smc.h
ldebieve: Looks good to me to mix both PSCI and OPTEE return args., let's follow that.

ldebieve:
> In commit message header line: "core: arm: ..."
> 
> _mk/config.mk_ requires a value for `CFG_WDT_SM_HANDLER_ID`. I think we need an extra switch like `CFG_WDT_SM_HANDLER_OPTEE_ID`.
> 
> ```diff
> ifeq (y-,$(CFG_WDT_SM_HANDLER)-$(CFG_WDT_SM_HANDLER_ID))
> $(error CFG_WDT_SM_HANDLER_ID must be defined when enabling CFG_WDT_SM_HANDLER)
> endif
> ```

I was expecting, when ok with this proposal, to keep the default OPTEE_SMC_FUNCID_WATCHDOG but keep the CFG_WDT_SM_HANDLER_ID available for override.

CFG_WDT_SM_HANDLER could be defined without HANDLER_ID to use the OP-TEE generic one and keep the override on:
<pre>
#ifdef CFG_WDT_SM_HANDLER_ID
enum sm_handler_ret wdt_sm_handler(struct thread_smc_args *args)
{
	if (args->a0 != CFG_WDT_SM_HANDLER_ID)
		return SM_HANDLER_PENDING_SMC;

	return __wdt_sm_handler(args);
}
#endif
</pre>

Do you want the associated code to the review?

etienne-lms:
> Do you want the associated code to the review?

Yes please. Having the full picture would be better.

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,49 @@
 /* See OPTEE_SMC_CALL_WITH_REGD_ARG above */
 #define OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG	U(19)
 
+/*
+ * Access to the secure watchdog.
```
etienne-lms, line 563: Suggestion:
` * Secure watchdog management service`

etienne-lms (commented):
some comments on ABI description.

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,49 @@
...
+ * a2	if a1 == 1, the timeout value to set
+ *	if a1 == 2, 0 to stop the watchdog, 1 to enable it.
+ *	preserved otherwise
+ * a3-6	Not used
+ * a7	Hypervisor Client ID register
+ *
+ * Normal return register usage:
+ * a0 PSCI_RET_SUCCESS
+ * a1	if a1 == 0, the minimal timeout value supported by watchdog
+ *	if a1 == 4, the remaining time before watchdog expires
```
etienne-lms, line 582: "... the remaining time **in seconds** before ..."

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,76 @@
...
+ * PSCI_RET_NOT_SUPPORTED - Function not supported
+ *
+ * Other commands
+ * [in]		a1	Other values
+ *
+ * Return codes:
+ * PSCI_RET_NOT_SUPPORTED - Function not supported
+ *
+ * a3-6		Not used
+ * a7		Hypervisor Client ID register
```
etienne-lms, line 618: The API is not that easy to read.
I would propose the below layout (without tabs in inline comments):
```
/*
 * Secure watchdog management service
 *
 * Call requests usage (for all watchdog command):
 * a0      [in]         SMC Function ID, OPTEE_SMC_WATCHDOG
 *         [out]        PSCI error code return
 * a1      [in]         Watchdog command (one of SMCWD_*)
 *         [out]        Depends on watchdog command (input a1)
 * a2      [in/out]     Depends on watchdog command (input a1)
 * a3-6    [in/out]     Not used
 * a7      [in/out]     Hypervisor Client ID register
 *
 * For watchdog command SMCWD_INIT: Watchdog initialization
 * [in]         a0      SMC Function ID (OPTEE_SMC_WATCHDOG)
 * [in]         a1      0 (command SMCWD_INIT)
 * [out]        a0      PSCI_RET_SUCCESS - Command success
 *                      PSCI_RET_INTERNAL_FAILURE - Initialization failure
 *                      PSCI_RET_NOT_SUPPORTED - Function ID (a0) not supported
 * [out]        a1      The minimal timeout supported value in seconds
 *                      Applicable only upon a0 is PSCI_RET_SUCCESS
 * [out]        a2      The maximal timeout supported value in seconds
 *                      Applicable only upon a0 is PSCI_RET_SUCCESS
 *
 * For watchdog command SMCWD_SET_TIMEOUT: Watchdog set timeout
 * [in]         a0      SMC Function ID (OPTEE_SMC_WATCHDOG)
 * [in]         a1      1 (command SMCWD_SET_TIMEOUT)
 * [in]         a2      The timeout value in seconds to set
 * [out]        a0      PSCI_RET_SUCCESS - Command success
 *                      PSCI_RET_INVALID_PARAMETERS - Incorrect input param
 *                      PSCI_RET_NOT_SUPPORTED - Function ID (a0) not supported
 *
 * For watchdog command SMCWD_ENABLE: Watchdog enable
 * [in]         a0      SMC Function ID (OPTEE_SMC_WATCHDOG)
 * [in]         a1      2 (command SMCWD_ENABLE)
 * [in]         a2      Set to 0 to stop the watchdog, 1 to enable it
 * [out]        a0      PSCI_RET_SUCCESS - Command success
 *                      PSCI_RET_INVALID_PARAMETERS - Incorrect input param
 *                      PSCI_RET_NOT_SUPPORTED - Function ID (a0) not supported
 *
 * For watchdog command SMCWD_PET: Ping the watchdog for refresh
 * [in]         a0      SMC Function ID (OPTEE_SMC_WATCHDOG)
 * [in]         a1      3 (command SMCWD_PET)
 * [out]        a0      PSCI_RET_SUCCESS - Command success
 *                      PSCI_RET_DISABLED - The watchdog is not enabled
 *                      PSCI_RET_NOT_SUPPORTED - Function ID (a0) not supported
 *
 * For watchdog command SMCWD_GET_TIMELEFT: Get time left
 * [in]         a0      SMC Function ID (OPTEE_SMC_WATCHDOG)
 * [in]         a1      4 (command SMCWD_GET_TIMELEFT)
 * [out]        a0      PSCI_RET_SUCCESS - Command success
 *                      PSCI_RET_DISABLED - The watchdog is not enabled
 *                      PSCI_RET_NOT_SUPPORTED - Function ID (a0) not supported
 * [out]        a1      Guaranteed time in seconds before watchdog expires
 *                      Applicable only upon a0 is PSCI_RET_SUCCESS
 *
 * For any other command value
 * [in]         a0      SMC Function ID (OPTEE_SMC_WATCHDOG)
 * [in]         a1      > 4
 * [out]        a0      PSCI_RET_INVALID_PARAMETERS - Invalid watchdog command
 *                      PSCI_RET_NOT_SUPPORTED - Function ID (a0) not supported
 *
 * By SMCCC convention:
 * PSCI_RET_SUCCESS, OPTEE_SMC_RETURN_OK and ARM_SMCCC_RET_SUCCESS
 * are equal.
 * PSCI_RET_NOT_SUPPORTED, OPTEE_SMC_RETURN_UNKNOWN_FUNCTION and
 * ARM_SMCCC_RET_NOT_SUPPORTED are equal.
 */
#define OPTEE_SMC_FUNCID_WATCHDOG		U(20)
#define OPTEE_SMC_WATCHDOG \
	OPTEE_SMC_FAST_CALL_VAL(OPTEE_SMC_FUNCID_WATCHDOG)

#define SMCWD_INIT		0
#define SMCWD_SET_TIMEOUT	1
#define SMCWD_ENABLE		2
#define SMCWD_PET		3
#define SMCWD_GET_TIMELEFT	4

Comment (jenswi-linaro) [reply core/arch/arm/include/sm/optee_smc.h:618]: How about keeping to the established style in this file? I believe it's clear enough.

File: core/arch/arm/include/sm/optee_smc.h [hunk 7]
```diff
@@ -559,6 +559,76 @@
...
+ * [out]	a0	PSCI error code return
+ *
+ * Watchdog commands
+ * Command SMCWD_INIT : Watchdog initialization
+ * [in]		a1	Set to 0
+ * [out]	a1	The minimal timeout value in seconds supported
+ *		a2	The maximal timeout value in seconds supported
+ *
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
```
jenswi-linaro, line 576: PSCI?  The SMCs here use OPTEE_SMC_RETURN_*

core/arch/arm/include/sm/optee_smc.h
ldebieve: @jenswi-linaro It was the initial proposal, after discussion with @etienne-lms I've tried to use a similar description based on PTA description that seems to be useful for such kind of "API" description. 
I'm okay to come back to previous style used in this file if it's clear enough.

core/arch/arm/include/sm/optee_smc.h
ldebieve: Keeping PSCI return code to be compatible with the existing service. Linux driver (and U-boot too) are using PSCI code return.

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,76 @@
...
+ *
+ * Call requests usage:
+ * [in]		a0	SMC Function ID, OPTEE_SMC_WATCHDOG
+ * [out]	a0	PSCI error code return
+ *
+ * Watchdog commands
+ * Command SMCWD_INIT : Watchdog initialization
+ * [in]		a1	Set to 0
+ * [out]	a1	The minimal timeout value in seconds supported
+ *		a2	The maximal timeout value in seconds supported
```
GseoC, line 573: minor: s/maximal/maximum

core/arch/arm/include/sm/optee_smc.h
```diff
@@ -559,6 +559,76 @@
 /* See OPTEE_SMC_CALL_WITH_REGD_ARG above */
 #define OPTEE_SMC_FUNCID_CALL_WITH_REGD_ARG	U(19)
 
+/*
+ * Secure watchdog management service
+ *
+ * Call requests usage:
+ * [in]		a0	SMC Function ID, OPTEE_SMC_WATCHDOG
+ * [out]	a0	PSCI error code return
```
GseoC, line 567: with [out], I think "return" is unnecessary

GseoC (commented):
For commit "driver: wdt: override the WDT_HANDLER_ID if defined", I would change:
s/override/overriden

core/arch/arm/include/sm/optee_smc.h
jenswi-linaro: That's a mess, to be honest. I don't see what PSCI has to do with the OPTEE_SMC range of SMCs, PSCI uses a different range of SMCs. When looking at the implementation I see that you still use `OPTEE_SMC_RETURN_UNKNOWN_FUNCTION`.

core/arch/arm/include/sm/optee_smc.h
jenswi-linaro: I'm sorry to change the direction. This looks small enough to make part of the normal OPTEE_SMC ABI, but then it should also follow the style of the ABI.
By the way, do you have examples of how this is used from the Linux kernel driver and U-Boot?

core/arch/arm/include/sm/optee_smc.h
etienne-lms: That is true but this is the ABI described in the [Linux kernel driver](https://elixir.bootlin.com/linux/v6.3/source/drivers/watchdog/arm_smc_wdt.c#L45) (and the [U-Boot one](https://elixir.bootlin.com/u-boot/v2023.07-rc1/source/drivers/watchdog/arm_smc_wdt.c#L45) by mimethism). Should these be fixed?

jenswi-linaro: That's relevant information. This doesn't have anything to do with the OPTEE_SMC interface at all, in fact, the ABI is already defined. In the kernel, it's accessed as a SiP service unless the SMC ID is overridden in device tree. I believe we should be better off with implementing it in the SiP service range here too, defined in a separate .h file and so on.

etienne-lms: As we have all means in OP-TEE to provide such service in a native way. I think it's a shame we can't define a generic OP-TEE service for watchdog management. The main objective of this P-R is to have an standard SMC func ID for this instead of requiring each platform to redefine its own func ID for the very exact same purpose.
This P-R also clarifies the service ABI that is not  explicit when one reads OP-TEE OS and Linux kernel implementations.

jenswi-linaro: There is already one defacto standard SMC id `0x82003D06` used if `arm,smc-id` isn't found in device tree. If the SiP SMC range isn't good enough and you must use the Trusted OS SMC range instead please use a sub-range that doesn't overlap with the range used by the OPTEE_SMC ABI range. I don't see any benefit in mixing the ABI with the OPTEE_SMC ABI.

OPTEE_SMC ABI uses decimal `50` as the sub-range for requests from normal world and decimal `62` for some signaling to OPTEED in TF-A.
Google is using decimal `63` for the `NSSMC_OPTEED_CALL_LOAD_IMAGE` so perhaps it's best to stay out of that sub-range entirely too.
Maybe decimal `60` is a suitable sub-range.

etienne-lms:
keep alive

PR #5953, opened 2023-04-11
[RFC]  Physically relocatable OP-TEE binary

jenswi-linaro:
This depends on https://github.com/OP-TEE/optee_os/pull/5952, please ignore any commit before "core: add core_mmu_tee_load_pa address".

This is an alternative to https://github.com/OP-TEE/optee_os/pull/5948.


core/arch/arm/kernel/boot.c
```diff
@@ -1621,3 +1627,89 @@ unsigned long __weak get_aslr_seed(void *fdt __unused)
...
+	if (hdr->version != FFA_BOOT_INFO_VERSION) {
+		EMSG("Bad boot info version %#"PRIx32, hdr->version);
+		panic();
+	}
+	if (hdr->desc_count != 1) {
+		EMSG("Bad boot info descriptor count %#"PRIx32,
+		     hdr->desc_count);
+		panic();
+	}
+	desc = (void *)((vaddr_t)hdr + hdr->desc_offset);
```
apalos, line 1653: depending on the desc_offset desc might end up being unaligned, do we care here?

core/mm/core_mmu.c
```diff
@@ -1346,6 +1432,10 @@ void __weak core_init_mmu_map(unsigned long seed, struct core_mmu_config *cfg)
 	struct tee_mmap_region *tmp_mmap = get_tmp_mmap();
 	unsigned long offs = 0;
 
+	if (IS_ENABLED(CFG_CORE_RELOCATABLE) &&
+	    (core_mmu_tee_load_pa & SMALL_PAGE_MASK))
```
apalos, line 1436: Do we have to panic here, or can we just load op-tee on the rounded up page aligned address

core/arch/arm/kernel/boot.c
jenswi-linaro: I don't think we care or we should check the alignment of `hdr` also. We could perhaps add a few asserts to help debugging, but I'm not sure it's worth it.

core/mm/core_mmu.c
jenswi-linaro: Do you mean to memmove the entire OP-TEE binary? If so then we should do that before we call relocate() for the first time in `core/arch/arm/kernel/entry_a64.S`. In fact, we should do it before moving `struct boot_embdata`. Do you think it's worth the trouble?

jenswi-linaro:
Rebased on master now that #5952 has been merged.

jenswi-linaro:
Force-push to fix one checkpatch issue.

etienne-lms (commented):
Very few minor comments. LGTM FWIW.
I've tested these changes on stm32mp1 w/ pager (i.e. with `CFG_CORE_RELOCATABLE` disabled), no regressions, but I guess you didn't expect any :)

Commit "core: add core_mmu_tee_load_pa address": 
in commit message: core: s/core_mmu_tee_base_pa/core_mmu_tee_load_pa/

core/arch/arm/kernel/boot.c
```diff
@@ -1625,3 +1627,89 @@ unsigned long __weak get_aslr_seed(void *fdt __unused)
...
+		panic();
+	}
+	ret = dt_getprop_as_number(fdt, 0, "load-address", &num);
+	if (ret < 0) {
+		EMSG("Can't read \"load-address\" from FF-A manifest at %p: error %d",
+		     fdt, ret);
+		panic();
+	}
+	*base = num;
+	ret = dt_getprop_as_number(fdt, 0, "mem-size", &num);
```
shpark, line 1696: Where can I find specification for the FF-A manifest format that include `mem-size` subnode? On https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html, there seems to be no `mem-size`.

shpark:
Just a clarification question. It seems like when `CFG_CORE_RELOCATABLE` is set, the whole trusted DRAM region (i.e., `{ .paddr = TRUSTED_DRAM_BASE, .size = TRUSTED_DRAM_SIZE`) is to be registered as TA RAM. Can this region potentially overlap with the TEE RX/RW regions? If so, how's the overlapped regions between TEE and TA regions are being removed from the TA memory pool (`tee_mm_sec_ddr`)?

core/mm/core_mmu.c
```diff
@@ -110,6 +120,53 @@ void core_mmu_get_secure_memory(paddr_t *base, paddr_size_t *size)
...
+	if (ARRAY_SIZE(secure_only) == 1) {
+		vaddr_t load_offs = 0;
+
+		assert(core_mmu_tee_load_pa >= secure_only[0].paddr);
+		load_offs = core_mmu_tee_load_pa - secure_only[0].paddr;
+
+		assert(secure_only[0].size >
+		       load_offs + TEE_RAM_VA_SIZE + TEE_SDP_TEST_MEM_SIZE);
+		b = secure_only[0].paddr + load_offs + TEE_RAM_VA_SIZE;
+		s = secure_only[0].size - load_offs - TEE_RAM_VA_SIZE -
```
shpark, line 156: It seems like the lower part of the secure memory (below `core_mmu_tee_load_pa`) is not being taken account into the TA RAM. Am I right? If so, do you plan to submit patches to also take account the lower memory?

jenswi-linaro: That part is currently expected to be quite small, 16kB or so in a Hafnium setup. I haven't made up my mind about what to do with that yet.

jforissier:
Sorry for closing, pushed the wrong button...
A couple of checkpatch errors can be fixed, let's ignore the line too long and of course the "ba" false positive.

jenswi-linaro:
Updated with checkpatch fixes.

jenswi-linaro:
Should I squash in the fixups?

core/arch/arm/include/ffa.h
```diff
@@ -187,5 +209,27 @@ struct ffa_mem_relinquish {
...
+
+/* FF-A v1.1 boot information header */
+struct ffa_boot_info_header {
+	uint32_t signature;
+	uint32_t version;
+	uint32_t blob_size;
+	uint32_t desc_size;
+	uint32_t desc_count;
+	uint32_t desc_offset;
+	uint64_t resserved;
```
balint-dobszay-arm, line 231: s/resserved/reserved/

jenswi-linaro:
Updated

jenswi-linaro:
Let me know if I should squash the fixups.

core/include/mm/core_mmu.h
```diff
@@ -661,6 +661,8 @@ static inline bool core_mmu_check_end_pa(paddr_t pa, size_t len)
  */
 void core_mmu_get_secure_memory(paddr_t *base, paddr_size_t *size);
 
+void core_mmu_get_ta_range(paddr_t *base, size_t *size);
```
etienne-lms, line 677:
```c
/*
 * core_mmu_get_ta_range() - get physical memory range reserved for TAs
 * @base: [out] range base address ref or NULL
 * @size: [out] range size ref or NULL
 */

File: core/include/kernel/dt.h [hunk 7]
```diff
@@ -96,6 +96,9 @@ enum dt_map_dev_directive {
 typedef TEE_Result (*dt_driver_probe_func)(const void *fdt, int nodeoffset,
 					   const void *compat_data);
 
+int dt_getprop_as_number(const void *fdt, int nodeoffset, const char *name,
```
etienne-lms, line 99:
```c
/*
 * dt_getprop_as_number() - get a DT property a unsigned number
 * @fdt: DT base address
 * @nodeoffset: node offset
 * @name: property string name
 * @num: output number read
 * Return 0 on success and a negative FDT error value on error
 * 
 * The size of the property determines if it is read as an unsigned 32-bit
 * or 64-bit integer.
 */

File: core/arch/arm/kernel/boot.c [hunk 8]
```diff
@@ -1625,3 +1627,91 @@ unsigned long __weak get_aslr_seed(void *fdt __unused)
...
+	else if (name_fmt == FFA_BOOT_INFO_FLAG_NAME_FORMAT_UUID)
+		DMSG("Boot info descriptor UUID %pUl", (void *)desc->name);
+	else
+		DMSG("Boot info descriptor: unknown name format %"PRIu8,
+		     name_fmt);
+
+	content_fmt = (desc->flags & FFA_BOOT_INFO_FLAG_CONTENT_FORMAT_MASK) >>
+		      FFA_BOOT_INFO_FLAG_CONTENT_FORMAT_SHIFT;
+	if (content_fmt != FFA_BOOT_INFO_FLAG_CONTENT_FORMAT_ADDR) {
+		EMSG("Bad boot info content format %"PRIu8", expected 0 (address)",
```
etienne-lms, line 1666: Suggestion for consistency:
```
		EMSG("Bad boot info content format %"PRIu8", expected %u (address)",
		     content_fmt, FFA_BOOT_INFO_FLAG_CONTENT_FORMAT_ADDR);

File: core/arch/arm/kernel/entry_a64.S [hunk 9]
```diff
@@ -275,7 +289,7 @@ clear_nex_bss:
 #endif
 
 #ifdef CFG_CORE_ASLR
-	mov	x0, x20
+	mov	x0, x20		/* dt address */
```
etienne-lms, line 310: s/dt/DT/

etienne-lms (commented):
"core: add core_mmu_tee_load_pa address"
"core: add core_mmu_get_ta_range()" (with an inline description comment suggestion)
"core: core_mmu.c: use secure_only[] where possible"
"core: virt: use core_mmu_get_ta_range()"
"core: arm: generic_ram_layout.h: remove TA_RAM_*"
"core: core_mmu.c: only try to add pager vaspace when enabled"
"core: add dt_getprop_as_number()" (with an inline description comment suggestion)
"core: add core_mmu_set_secure_memory()"


"core: ffa: add boot info structs and defines"
"core: ffa: parse boot info" (with typos in message (see below) + 2 minor comments)
"core: support physically relocatable OP-TEE binary"
"ci: add job with SPMC at S-EL2 and relocatable"

Typos in commit message for "core: ffa: parse boot info":
s/as an boot arguments/as a boot argument/
s/These covers/These cover/
s/core and TA memory/core and TA memories/

etienne-lms:
> Skipped the `s/core and TA memory/core and TA memories/` part since I don't think plural isn't needed for that.

I believe you're right :)

jforissier:
@apalos @shpark @balint-dobszay-arm @LeiSen62 any comment on this? Looks ready for merging.

balint-dobszay-arm:
> any comment on this? Looks ready for merging.

No more comments from my side.

core/arch/arm/arm.mk
```diff
@@ -122,6 +123,15 @@ $(call force,CFG_CORE_SEL2_SPMC,n)
 $(call force,CFG_CORE_SEL1_SPMC,n)
 endif
 
+ifeq ($(CFG_CORE_RELOCATABLE)-$(CFG_WITH_PAGER),y-y)
+$(error CFG_CORE_RELOCATABLE and CFG_WITH_PAGER are not compatible)
+endif
+ifeq ($(CFG_CORE_RELOCATABLE),y)
+ifneq ($(CFG_CORE_SEL2_SPMC),y)
+$(error CFG_CORE_RELOCATABLE depends on CFG_CORE_SEL2_SPMC)
```
LeiSen62, line 131: Whether this feature can be decoupled from CFG_CORE_SEL2_SPMC? If CFG_CORE_SEL2_SPMC is not supported, the OP-TEE also needs to be loaded to different addresses.

shpark: +1 I am also curious if this patch can be generalized to be used for physical KASLR, but that might be out of scope of this PR?

jenswi-linaro: OP-TEE needs a way to find out which memory it can use when we don't hardcode the range of available memory any longer. For `CFG_CORE_SEL2_SPMC` we have the SP manifest via boot info. Other configurations are more implementation specific.

To enable other configurations, we need something corresponding to `get_sec_mem_from_manifest()`. But that's not in the scope of this PR.

core/arch/arm/arm.mk
```diff
@@ -113,6 +113,7 @@ $(call force,CFG_CORE_FFA,y)
 $(call force,CFG_CORE_SEL1_SPMC,n)
 $(call force,CFG_CORE_EL3_SPMC,n)
 CFG_CORE_HAFNIUM_INTC ?= y
+CFG_CORE_RELOCATABLE ?= y
```
jforissier, line 116: I think this flag deserves some description here, or maybe a different name, because with ASLR the core is already relocatable in the VA space, but that's a different relocation we're dealing with here. How about `CFG_CORE_PHYS_RELOCATABLE`?

jenswi-linaro:
Renamed `CFG_CORE_RELOCATABLE` to `CFG_CORE_PHYS_RELOCATABLE` and added a short description. Updated the affected commits with the new name.

jforissier:
```
 core: support physically relocatable OP-TEE binary

With CFG_CORE_RELOCATABLE=y enable support in OP-TEE to relocate itself
```
s/CFG_CORE_RELOCATABLE/CFG_CORE_PHYS_RELOCATABLE/

jenswi-linaro:
Updated commit message in "core: support physically relocatable OP-TEE binary"

core/arch/arm/arm.mk
LeiSen62: I think you can obtain the base address and size in other ways, and there is no need to bind it to CFG_CORE_SEL2_SPMC, which reduces the universality of CFG_CORE_RELOCATABLE.

LeiSen62: In this feature, we pay more attention to the loading base address. The loading base address can be obtained through "adrp x3, __text_start" and then saved to the global variable.

jenswi-linaro: Once this is merged, feel free to propose a PR where the `CFG_CORE_SEL2_SPMC` dependency is removed. I believe that broadening the scope of this PR now that it's ready to be merged would delay this feature further without much gain.

LeiSen62: Will you have a new PR to decouple CFG_CORE_RELOCATABLE from CFG_CORE_SEL2_SPMC in the future? If not, will the community accept such PRs?

jforissier: I agree with @jenswi-linaro. @LeiSen62 do you think you could propose something after this is merged?

Nitpicking... Related to that, maybe the commit description in "core: support physically relocatable OP-TEE binary" could hint at what would be required to lift the limitation. Something like:
"This feature is currently only supported with CFG_CORE_SEL2_SPMC=y since the TEE core has to know the range of available memory. With SPMC at EL2 this is accomplished via get_sec_mem_from_manifest()."

core/arch/arm/arm.mk
```diff
@@ -113,6 +113,9 @@ $(call force,CFG_CORE_FFA,y)
 $(call force,CFG_CORE_SEL1_SPMC,n)
 $(call force,CFG_CORE_EL3_SPMC,n)
 CFG_CORE_HAFNIUM_INTC ?= y
+# Enable support in OP-TEE to relocate itself to allow it to run from
+# physical address that differs from the link address
```
etienne-lms, line 117: nitpicking: `... from **a** physical address that differs ...`

core/arch/arm/arm.mk
etienne-lms: Maybe move description in _mk/config.mk_?
```make
# Enable support in OP-TEE to relocate itself to allow it to run from a
# physical address that differs from the link address. This feature is
# currently supported only when CFG_CORE_SEL2_SPMC is enabled.
CFG_CORE_RELOCATABLE ?= n
$(eval $(call cfg-depends-one,CFG_CORE_RELOCATABLE,CFG_CORE_SEL2_SPMC))
```

etienne-lms (commented):
Suggestion for commit message of patch "ci: add job with SPMC at S-EL2 and relocatable":
replace `Adds a job with` with `Build with`. The change does not add a job in the CI.

core/arch/arm/arm.mk
jenswi-linaro: I'm not sure that's the best place. With `CFG_CORE_SEL2_SPMC` we should use `y` as the default value, else `n`. This is likely to change as more and more configurations can support `CFG_CORE_RELOCATABLE`. The number of configuration options enabled with `CFG_CORE_SEL2_SPMC=y` is also expected to grow and it's not nice to have that in too many different places.

jenswi-linaro:
> Suggestion for commit message of patch "ci: add job with SPMC at S-EL2 and relocatable": replace `Adds a job with` with `Build with`. The change does not add a job in the CI.

Addressed

core/arch/arm/arm.mk
etienne-lms: I think it is legitimate for _arm.mk_ to default enable the config and for _mk/config.mk_ to default disable that config.
The only point is that there should be only one place where the config switch description can be found.

That said, as currnetly only SPM'ed OP-TEE can support this feature, I'm fine having the config and inline description as currently proposed by this P-R.

jenswi-linaro: OK, thanks.

jenswi-linaro:
This should be ready to merge now.

core/arch/arm/arm.mk
jenswi-linaro: @LeiSen62, this is merged now feel free to propose the improvement you had in mind.

PR #5954, opened 2023-04-12
interrupt chip controller drivers

etienne-lms:
This P-R proposes a interrupt (controller) chip framework and moves GIC driver to that framework.

These change allows to integrate interrupt chip that multiplex interrupts on an interrupt of a parent interrupt controller (as the GIC or another interrupt chip). Interrupt consumer drivers can use generic DT bindings (properties "interrupts" and "interrupts-extended") to get their interrupt resources.

These changes preserve exiting `itr_xxx()` API functions and friends related to the unique root CPU interrupt controller (likely GIC based on Arm arch). They add `interrupt_xxx()` functions where interrupt controller is an `struct itr_chip`, not specifically the GIC.

Comments are welcome. Later addition should bring some dt_driver tests means for this interrupt chip framework.


jenswi-linaro:
This is interesting. It would help with a picture and some description in the documentation to better understand how fits together.

I don't think we should need two sets of functions, with a good abstraction layer we should be able to have only one set of functions.

etienne-lms:
> It would help with a picture and some description in the documentation

Ok

> I don't think we should need two sets of functions, with a good abstraction layer we should be able to have only one set of functions.

This series preserves legacy API functions `itr_xxx()` for backward compat of the prototypes... but maybe that's useless. 
I could change the functions to assume that when argument `struct itr_chip *` is NULL, it relates to the root CPU interrupt controller, unless one thinks it is an ugly shortcut.

jenswi-linaro:
I think we should make the API what we'd like it to be, that way it's easier to understand and use. If we need to change all the users in the tree, so be it.

etienne-lms:
Rebased on matser tip.

I've created https://github.com/OP-TEE/optee_docs/pull/192 to address https://github.com/OP-TEE/optee_os/pull/5954#issuecomment-1512810626 about documentation.

etienne-lms:
This series is not ready to use. I did not address deprecation of legacy function `itr_xxx()` and there interrupt configuration means must be refined. Please consider this is still work in progress.
I plan to squash all fixup commits in my next update.

etienne-lms:
I have updated the series, squashing the fixup commits, changing all `itr_core_` label prefixes to interrupt_main_` label prefixes. The 2 last commits remove old `itr_xxx()` functions, leaving only the new `interrupt_xxx()` API functions that can be used with main interrupt controller (as the GIC) or a secondary interrupt controller.

Note the serie introduces dt_driver support for description of interrupt controller chains from a DT blob.

The series is still long, at least it gives the big picture. If we agree on moving forward, i'll extract the first commits from this series and create dedicated P-Rs to ease step by step review.

core/include/kernel/interrupt.h
```diff
@@ -329,4 +344,113 @@ TEE_Result interrupt_alloc_add_main_handler(size_t it_num,
  * This function may panic on non-NULL invalid @hdl reference.
  */
 void interrupt_remove_free_handler(struct itr_handler *hdl);
+
+/*
+ * dt_get_itr_func - Typedef of function to get an interrupt line in DT node
+ *
+ * @args	Pointer to devicetree description of the clock to parse
```
etienne-lms, line 345: `Phandler arguments reference`

core/include/kernel/interrupt.h
```diff
@@ -329,4 +344,113 @@ TEE_Result interrupt_alloc_add_main_handler(size_t it_num,
...
+ * calls interrupt_enable().
+ * Upon success, struct itr_line must point to memory allocated with malloc()
+ * or like as it is freed prior returning to caller function.
+ */
+typedef struct itr_line *(*dt_get_itr_func)(struct dt_driver_phandle_args *args,
+					    void *data, TEE_Result *res);
+
+#ifdef CFG_DT
+/**
+ * dt_register_itr_provider - Register an interrupt provider
```
etienne-lms, line 368: s/dt_register_itr_provider/dt_register_interrupt_provider/

etienne-lms:
May I rebase on mater tip and squash the fixup commits?

Once done, as the series is quite long (19 patches), maybe I should create smaller P-Rs to ease review.

jenswi-linaro:
> May I rebase on mater tip and squash the fixup commits?

Yes, please.

etienne-lms:
Fixup commit squashed and series rebased on master branch tip.

etienne-lms:
But I see I missed a lot of comments I posted.
I'll fix that and force push again...

etienne-lms:
Series is now back ready for review.

core/include/kernel/interrupt.h
```diff
@@ -65,6 +65,9 @@ void interrupt_main_init(struct itr_chip *data);
 
 void itr_handle(size_t it);
 
+/* Retrieve main interrupt controller reference */
+struct itr_chip *interrupt_main_get(void);
```
jenswi-linaro, line 123: How about `interrupt_get_main_chip()` instead?

core/include/kernel/interrupt.h
```diff
@@ -28,6 +34,15 @@ struct itr_chip {
 			  uint32_t *prio);
 };
 
+/*
+ * struct itr_ops - Interrupt controller operations
+ * @add		Register and configure a interrupt line
```
jenswi-linaro, line 47: The word "line" used here and in the following lines is redundant, please remove it.
s/a interrupt/an interrupt/

core/include/kernel/interrupt.h
```diff
@@ -6,21 +6,28 @@
...
 #define ITRF_SHARED		BIT(1)
 
+struct itr_handler;
+
 /*
  * struct itr_chip - Interrupt controller
  *
  * @ops Operation callback functions
+ * @name Controller name, for debug purpose
+ * @handlers Registered hadnlers list head
```
jenswi-linaro, line 25: s/hadnlers/handlers/

core/include/kernel/interrupt.h
```diff
@@ -164,4 +181,117 @@ static inline struct itr_handler *itr_alloc_add(size_t it,
...
+
+/*
+ * interrupt_mask() - Mask a interrupt
+ * @chip	Interrupt controller
+ * @itr_num	Interrupt number
+ * This function may be called in interrupt context
+ */
+static inline void interrupt_mask(struct itr_chip *chip, size_t itr_num)
+{
+	if (chip->ops->mask)
```
jenswi-linaro, line 215: Why not make this function pointer mandatory? Same for the other ops below.

core/include/kernel/interrupt.h
```diff
@@ -164,4 +181,117 @@ static inline struct itr_handler *itr_alloc_add(size_t it,
...
+static inline void interrupt_disable(struct itr_chip *chip, size_t itr_num)
+{
+	if (chip->ops->disable)
+		chip->ops->disable(chip, itr_num);
+	else
+		interrupt_mask(chip, itr_num);
+}
+
+/*
+ * interrupt_configure() - Configure an interrupt line in interrupt controller
```
jenswi-linaro, line 258: The word "line" is redundant.

core/kernel/interrupt.c
```diff
@@ -143,3 +143,125 @@ void __weak __noreturn interrupt_main_handler(void)
...
+		if (h->it == itr_num) {
+			if (h->handler(h) == ITRR_HANDLED)
+				was_handled = true;
+			else if (!(h->flags & ITRF_SHARED))
+				break;
+		}
+	}
+
+	if (!was_handled) {
+		EMSG("mask %s:%zu", chip->name, itr_num);
```
jenswi-linaro, line 146: Disabling unhandled...

core/include/kernel/interrupt.h
```diff
@@ -66,6 +67,20 @@ struct itr_ops {
...
+/*
+ * struct itr_line - Interrupt line description
+ * @chip	Interrupt controller
+ * @itr_num	Interrupt number
+ *
+ * This struct is used for binding interrupt device data between
+ * drivers when using DT_DRIVERS means. See dt_get_itr_func type
+ * definition.
+ */
+struct itr_line {
```
jenswi-linaro, line 79: How about "struct itr_desc" instead?

core/include/kernel/interrupt.h
```diff
@@ -265,6 +280,23 @@ TEE_Result interrupt_configure(struct itr_chip *chip, size_t itr_num,
...
 TEE_Result interrupt_add_handler(struct itr_handler *hdl);
 
+/*
+ * interrupt_add_main_handler() - Register a main controller interrupt handler
+ * @hdl		Interrupt handler to register
+ *
+ * @hdl.chip is expected either already set to main interrupt controller
+ * or NULL in which case it is set to interrupt_main_get().
+ */
+static inline TEE_Result interrupt_add_main_handler(struct itr_handler *hdl)
```
jenswi-linaro, line 284: I'm not sure we need `interrupt_add_main_handler()` and `interrupt_alloc_add_main_handler()`. It's not too much for the caller to call `interrupt_main_get()` first.

core/include/kernel/interrupt.h
```diff
@@ -287,11 +319,137 @@ TEE_Result interrupt_alloc_add_handler(struct itr_chip *chip, size_t it_num,
...
+ * code. The interrupt matches the reference found with properties of the
+ * pointer FDT node.
+ *
+ * Upon success, the interrupt is configured and consumer can add a handler
+ * function to the interrupt. Yet, the interrupt is not enabled until consumer
+ * calls interrupt_enable().
+ * Upon success, struct itr_line must point to memory allocated with malloc()
+ * or like as it is freed prior returning to caller function.
+ */
+typedef struct itr_line *(*dt_get_itr_func)(struct dt_pargs *args, void *data,
```
jenswi-linaro, line 362: I know we have a few cases where we return the error code like this, but they are not very nice. I would prefer if we returned the TEE_Result and instead had an `**out_itr` or such.

core/include/kernel/interrupt.h
```diff
@@ -287,11 +319,137 @@ TEE_Result interrupt_alloc_add_handler(struct itr_chip *chip, size_t it_num,
...
+ * Return TEE_SUCCESS in case of success
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if interrupt controller is not yet inited
+ * Return TEE_ERROR_ITEM_NOT_FOUND if the DT does not reference target interrupt
+ * Return any other TEE_Result compliant code in case of error
+ */
+TEE_Result dt_get_interrupt_by_name(const void *fdt, int node, const char *name,
+				    struct itr_chip **chip, size_t *itr_num);
+#else
+static inline
+TEE_Result dt_register_interrupt_provider(const void *dt __unused,
```
jenswi-linaro, line 414: Both `clang-format` and I prefer breaking the line as:
``` C
static inline TEE_Result
dt_register_interrupt_provider(...
```
It's also as far as I know the way it's normally done in the kernel when needed.

core/include/kernel/interrupt.h
```diff
@@ -287,11 +319,137 @@ TEE_Result interrupt_alloc_add_handler(struct itr_chip *chip, size_t it_num,
...
+						  int node __unused,
+						  const char *name __unused,
+						  struct itr_chip **ch __unused,
+						  size_t *itr_num __unused)
+{
+	return TEE_ERROR_NOT_IMPLEMENTED;
+}
+#endif /*CFG_DT*/
+
+/* Helper function for when interrupt index is 0 */
```
jenswi-linaro, line 442: What's an interrupt index? Is it something DT specific?

jenswi-linaro:
For "core: interrupt: rename internal itr_chip to itr_main_chip" please apply:

For "core: dt_driver: define interrupt controller drivers identifier" please apply:

Comment on commit message for "drivers: move to interrupt_call_handlers()":
Replace "Deprecates itr_handle()" with "Removes itr_handle()"

For the entire PR:
s/interrupt line/interrupt/g where you can't find a better word than "line".

core/include/kernel/interrupt.h
etienne-lms: ok

core/include/kernel/interrupt.h
etienne-lms: fair

core/include/kernel/interrupt.h
etienne-lms: right, let's make it mandatory

core/include/kernel/interrupt.h
etienne-lms: ok.

core/include/kernel/interrupt.h
etienne-lms: This prototype is related to how type `get_of_device_func` was introduced with _dt_driver.h_, without a `TEE_Result` status:
https://github.com/OP-TEE/optee_os/blob/8c0c44c9d2f768a95beacd4605776cfe29c9bff6/core/include/kernel/dt_driver.h#L35-L36
which was introduced later as output argument:
https://github.com/OP-TEE/optee_os/blob/d8b14b46af9d2c7cee7f875e389e1a659bd123da/core/include/kernel/dt_driver.h#L41-L42

I admit there is not real need to use a `void *` return value and a `TEE_Result *` output value instead of the opposite. I can create a P-R to switch these back if you think it would be more consistent regarding the overall core implementation.

core/include/kernel/interrupt.h
etienne-lms: ok, i'll update.

core/include/kernel/interrupt.h
etienne-lms: Interrupt index relates to the index of the interrupt reference by DT binding property `interrupts` that can list several interrupts a driver can consume , see [DT binding docs](https://elixir.bootlin.com/linux/latest/source/Documentation/devicetree/bindings/interrupt-controller/interrupts.txt#L17). It is not related to interrupt number identifier value used by an interrupt controllers.

core/kernel/interrupt.c
etienne-lms: right!

core/include/kernel/interrupt.h
jenswi-linaro: Yes, please.

core/arch/arm/plat-vexpress/main.c
```diff
@@ -156,7 +156,8 @@ static TEE_Result init_console_itr(void)
 {
 	TEE_Result res = TEE_ERROR_GENERIC;
 
-	res = interrupt_add_main_handler(&console_itr);
+	console_itr.chip = interrupt_get_main_chip();
+	res = interrupt_add_handler(&console_itr);
```
etienne-lms, line 160: Her, i think having a `interrupt_add_main_handler()` wold be nice, to not have this 2 steps: definition/init of `struct itr_handler` outside the function but needed to loading `.chip` field from the function right before calling `interrupt_add_handler()`.

Would you be ok we define an `interrupt_add_main_chip_handler() helper function?

core/include/kernel/interrupt.h
jenswi-linaro: Thanks for the explanation. Please add something short about what index means.

core/arch/arm/plat-vexpress/main.c
jenswi-linaro: I see your point, but I would prefer something more generic. How about
``` C
TEE_Result interrupt_add_handler_with_chip(struct itr_chip *chip,
                                           struct itr_handler *h);
```
Then it could be called as:
``` C
res = interrupt_add_handler_with_chip(interrupt_get_main_chip(),
                                      &console_itr);
```
My point is that the pattern will not differ much if another interrupt controller is used. You just need to supply a different chip pointer.

etienne-lms: ok, i'll post a fixup commit

core/include/kernel/interrupt.h
etienne-lms: Ok, i'll create a P-R for that.

core/include/kernel/interrupt.h
```diff
@@ -133,4 +199,224 @@ static inline struct itr_handler *itr_alloc_add(size_t it,
...
+ * @chip	Interrupt controller
+ * @itr_num	Interrupt number
+ *
+ * This function is called from an interrupt context by a primary interrupt
+ * handler. This function calls the handlers registered for that interrupt
+ */
+void interrupt_call_handlers(struct itr_chip *chip, size_t itr_num);
+
+/*
+ * interrupt_mask() - Mask a interrupt
```
etienne-lms, line 217: missed that: s/a interrupt/an interrupt/
ditto for the 3 below occurrences.

jenswi-linaro:
Feel free to squash in the updates.

etienne-lms:
Sorry I missed your last comment.
Ok, i'll squash commits.

etienne-lms:
I added late fixup commit to address https://github.com/OP-TEE/optee_os/pull/5954#discussion_r1211841887 that I missed when fixing & squashing.

core/include/kernel/interrupt.h
```diff
@@ -61,26 +72,41 @@ enum itr_return {
...
 };
 
+#define ITR_HANDLER(_chip, _itr_num, _flags, _fn, _priv) \
+	((struct itr_handler){ \
+		.chip = (_chip), .it = (_itr_num), .flags = (_flags), \
+		.handler = (_fn), .data = (_priv), \
+	})
+
+/*
+ * Return true only is interrupt chip provides required handlers
```
borneoa, line 117: s/is/if/

core/include/kernel/interrupt.h
```diff
@@ -164,4 +190,129 @@ static inline struct itr_handler *itr_alloc_add(size_t it,
...
+static inline void interrupt_enable(struct itr_chip *chip, size_t itr_num)
+{
+	if (chip->ops->enable)
+		chip->ops->enable(chip, itr_num);
+	else
+		interrupt_unmask(chip, itr_num);
+}
+
+/*
+ * interrupt_disable() - Enable an interrupt
```
borneoa, line 252: s/Enable/Disable/

core/include/kernel/interrupt.h
```diff
@@ -315,4 +330,120 @@ TEE_Result interrupt_alloc_add_handler(struct itr_chip *chip, size_t it_num,
...
+/**
+ * dt_get_interrupt_by_index() - Get an interrupt from DT by interrupt index
+ *
+ * Interrupt index (@index) refers to the index of the target interrupt to be
+ * retrieved as DT binding property "interrupts" may define several
+ * interrupts.
+ *
+ * @fdt		Device tree to work on
+ * @node	Node offset of the subnode containing interrupt(s) references
+ * @index	Index in "interrupts" or "exteneded-interrupts" property list
```
borneoa, line 374: s/exteneded/extended/

core/include/kernel/interrupt.h
```diff
@@ -315,4 +330,120 @@ TEE_Result interrupt_alloc_add_handler(struct itr_chip *chip, size_t it_num,
...
+ * code. The interrupt matches the reference found with properties of the
+ * pointer FDT node.
+ *
+ * Upon success, the interrupt is configured and consumer can add a handler
+ * function to the interrupt. Yet, the interrupt is not enabled until consumer
+ * calls interrupt_enable().
+ * Upon success, struct itr_desc must point to memory allocated with malloc()
+ * or like as it is freed prior returning to caller function.
+ */
+typedef struct itr_desc *(*dt_get_itr_func)(struct dt_pargs *args, void *data,
```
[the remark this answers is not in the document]
borneoa: No, I'm wrong! This typedef is required in the #else case too!

core/kernel/interrupt.c
```diff
@@ -265,3 +265,123 @@ void interrupt_remove_free_handler(struct itr_handler *hdl)
...
+	/* Get interrupt chip/number from phandle and "interrupts" property */
+	*desc = dt_driver_device_from_node_idx_prop_phandle("interrupts", fdt,
+							    node, index,
+							    DT_DRIVER_INTERRUPT,
+							    phandle, &res);
+	return res;
+}
+
+/*
+ * Provide an itr_desc based on "interrupts-extended " property bindings.
```
borneoa, line 308: Please remove the space from last char of the string

core/include/kernel/interrupt.h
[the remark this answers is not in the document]
borneoa: No, I'm wrong! This typedef is required in the #else case too!

core/kernel/interrupt.c
```diff
@@ -265,3 +265,123 @@ void interrupt_remove_free_handler(struct itr_handler *hdl)
...
+ * May return TEE_ERROR_DEFER_DRIVER_INIT if parent controller is found
+ * by not yet initiallized.
```
borneoa, line 310: s/by/but/ ?
s/initiallized/initialized/

core/include/kernel/interrupt.h
```diff
@@ -61,26 +72,41 @@ enum itr_return {
...
+		.handler = (_fn), .data = (_priv), \
+	})
+
+/*
+ * Return true only is interrupt chip provides required handlers
+ * @chip: Interrupt controller reference
+ */
+static inline bool itr_chip_is_valid(struct itr_chip *chip)
+{
+	return chip && chip->ops && chip->ops->mask && chip->ops->unmask;
```
etienne-lms: `.add` is also required. I'll add it.

core/kernel/interrupt.c
```diff
@@ -23,13 +23,15 @@ static struct itr_chip *itr_main_chip __nex_bss;
...
 {
+	assert(itr_chip_is_valid(chip));
+
 	itr_main_chip = chip;
 }
 
 struct itr_chip *interrupt_get_main_chip(void)
 {
-	return itr_main_chip;
 	SLIST_INIT(&itr_main_chip->handlers);
```
etienne-lms, line 33: this is not a correction, it's rather another bug!
I'll fix that.

etienne-lms:
I added a commit to implement mask/unmask operation handlers in GIC itr_chip since these are now required.

tprrt:
> Fixup commits squashed. Review tag applied on the 2 related commits and fixed commit message for "drivers: move to interrupt_call_handlers()" as per [#5954 (comment)](https://github.com/OP-TEE/optee_os/pull/5954#issuecomment-1570403576).
> 
> @borneoa, you told me you were interested in that framework. Could you have a look at these patches? @tprrt, could you test this series on your boards? This series impacts Atmel main interrupt controller integration.

Hello,

Yes, I can test this serie on my boards, at the beginning of next week.

Kind regards,
Thomas

core/include/kernel/interrupt.h
etienne-lms: @jenswi-linaro: Should we really mandate mask/unmask operators? GIC, Atmel SAIC and Hafnium HFIC don't provide specific interrupt masking means so I need to use their relate interrupt disable/enable function for these operators. Does it make sence for these main interrupt controllers?

jenswi-linaro: I guess my confusion regarding the mask/unmask operators stems from this, I don't see what it's supposed to do.

core/include/kernel/interrupt.h
```diff
@@ -58,11 +82,15 @@ struct itr_handler {
...
- * Initialise core interrupt controller driver
- * @data Core controller main data reference to register
+ * Initialise main interrupt controller driver
+ * @data Main controller main data reference to register
  */
 void interrupt_main_init(struct itr_chip *data);
 
+/*
+ * Call handlers registered for that interrupt in core interrupt controller
+ * @it Interrupt line number
```
jenswi-linaro: Interrupt number.

jenswi-linaro (commented):
With the comment for "core: interrupt: add inline descriptions" addressed please apply:

core/kernel/interrupt.c
```diff
@@ -143,3 +143,125 @@ void __weak __noreturn interrupt_main_handler(void)
...
+	struct itr_handler *h = NULL;
+
+	assert(hdl && hdl->chip->ops);
+
+	SLIST_FOREACH(h, &hdl->chip->handlers, link)
+		if (h->it == hdl->it &&
+		    (!(hdl->flags & ITRF_SHARED) || !(h->flags & ITRF_SHARED)))
+			return TEE_ERROR_GENERIC;
+
+	interrupt_configure(hdl->chip, hdl->it, IRQ_TYPE_NONE, 0);
```
jenswi-linaro, line 183: What type `IRQ_TYPE_NONE` and prio `0` isn't good enough, how is it configured for the interrupt?

core/kernel/interrupt.c
```diff
@@ -143,3 +143,125 @@ void __weak __noreturn interrupt_main_handler(void)
...
+TEE_Result interrupt_add_handler(struct itr_handler *hdl)
+{
+	struct itr_handler *h = NULL;
+
+	assert(hdl && hdl->chip->ops);
+
+	SLIST_FOREACH(h, &hdl->chip->handlers, link)
+		if (h->it == hdl->it &&
+		    (!(hdl->flags & ITRF_SHARED) || !(h->flags & ITRF_SHARED)))
+			return TEE_ERROR_GENERIC;
```
jenswi-linaro, line 181: How is the caller expected to recover from this?

jenswi-linaro (commented):
Comments for "core: interrupt: interrupt chip framework".

The paragraph
"This change implements new interrupt_*() function in source file
 irq_chip.c but I think this could straight be located in interrupt.c."
could be removed.

Regarding:
"   This change does not modify existing interrupt API function that
    allow a driver to get an interrupt from the root interrupt controller.
    That said, legacy API function (itr_enable(), itr_add(), ...) could be
    renamed to emphasize they relate to root controller (itr_main)."
Wouldn't it make more sense to remove the legacy API at the end of this PR?

jenswi-linaro (commented):
For commit "core: interrupt: core controller uses irq_chip list head" and its fixups, please apply:

core/include/kernel/interrupt.h
etienne-lms: Some interrupt controllers are able to mask/unmask interrupt, so it's fair to provide a handler for that.

For consistency, I think we should mandate `mask`/`unmask` operators and change `interrupt_call_handlers()` to mask unhandled interrupt, not disable them, as `.disable` operator has only mask.unmask are expected to be called from uninterruptible contexts.

Regarding GIC, SAIC and other main controller driver, they can't mask individual interrupts, and no one ask them to do so, so let use their disable/enable function handlers as mask/unmask handlers, as best effort. It think it is fair this way.

core/include/kernel/interrupt.h
etienne-lms: addressed

etienne-lms:
in reply to https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462263885:
> Regarding:
> " This change does not modify existing interrupt API function that
> allow a driver to get an interrupt from the root interrupt controller.
> That said, legacy API function (itr_enable(), itr_add(), ...) could be
> renamed to emphasize they relate to root controller (itr_main)."
> Wouldn't it make more sense to remove the legacy API at the end of this PR?

I agree. There are few more  patches need for that. I'll appedn them to the series.

core/kernel/interrupt.c
etienne-lms: Faced that with Atmel SAIC :)
For chip that require specific type/prio i'll add a helper function in a appended fixup commit.

core/kernel/interrupt.c
etienne-lms: I think it is a programming error to mix shared and non-shared interrupt handlers on a same interrupt. At least caller will know something is bad. Existing API function `itr_add_type_prio()` did such verification and panicked in debug mode. I think it is better to simply report the error and let caller panic at its will.

etienne-lms:
I've squashed the fixup commits and applied review tag on 3 commits (https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462257113, https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462321608  and https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462325071). The previously reviewed series can be found [here: branch itr-chip-prev](https://github.com/etienne-lms/optee_os/commits/itr-chip-prev).

For commit: "core: interrupt: interrupt chip framework", I updated the commit message according to https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462263885 and https://github.com/OP-TEE/optee_os/pull/5954#issuecomment-1577183123) and I addressed review comments with an new appended 1st fixup commit:
"[review] core: interrupt: interrupt chip framework"

Fixed Atmel driver issues (https://github.com/OP-TEE/optee_os/pull/5954#issuecomment-1576826879) with 2 fixup commits:
"[review] drivers: atmel_piobu: upgrade to new interrupt framework"
"[review] drivers: atmel_wdt: upgrade to new interrupt framework"

I appended 3 new commits to remove the old interrupt API functions `itr_xxx()`:
"[review] plat-stm32mp1: upgrade to new interrupt framework"
"core: notif: upgrade to new interrupt framework"
"core: interrupt: remove old API functions"

Last, I appendeded another fixup commit "[review] core: interrupt: interrupt chip framework" to mask unhandled interrupt instead of disabling them. This is needed as mask operation handler is expected to be atomic while disable operation handler does not.

@tprrt: cuold you test this updated series? Atmel SAIC controller shoulld be provided the expect configuration data now.

etienne-lms:
Should I create smaller P-Rs to get the series reviewed (~22 commits)

core/kernel/interrupt.c
jenswi-linaro: Fair enough, please add a debug print.

core/include/kernel/interrupt.h
jenswi-linaro: > Some interrupt controllers are able to mask/unmask interrupt, so it's fair to provide a handler for that.

Out of curiosity, could you name one?

core/kernel/dt_driver.c
```diff
@@ -292,6 +292,38 @@ void *dt_driver_device_from_parent(const void *fdt, int nodeoffset,
...
+		return NULL;
+	}
+
+	prv = dt_driver_get_provider_by_phandle(phandle, type);
+	if (!prv) {
+		*res = TEE_ERROR_DEFER_DRIVER_INIT;
+		return NULL;
+	}
+
+	prop_index *= dt_driver_provider_cells(prv);
```
jenswi-linaro, line 321: We should check that this is in the range of the `prop` array.

jenswi-linaro (commented):
Comment for "core: dt_driver: add helper for old fashion interrupt bindings". With that addressed
please apply:

jenswi-linaro (commented):
Comment for "core: dt_driver: add helper for old fashion interrupt bindings". With that addressed
please apply:

jenswi-linaro:
For commit "core: interrupt: interrupt_get_main_chip() returns main controller":

jenswi-linaro:
> Should I create smaller P-Rs to help this series (~22 commits) review?

Sure, feel free to chip off a few patches. :-)

tprrt:
> I've squashed the fixup commits and applied review tag on 3 commits ([#5954 (review)](https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462257113), [#5954 (review)](https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462321608) and [#5954 (review)](https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462325071)). The previously reviewed series can be found [here: branch itr-chip-prev](https://github.com/etienne-lms/optee_os/commits/itr-chip-prev).
> 
> For commit: "core: interrupt: interrupt chip framework", I updated the commit message according to [#5954 (review)](https://github.com/OP-TEE/optee_os/pull/5954#pullrequestreview-1462263885) and [#5954 (comment)](https://github.com/OP-TEE/optee_os/pull/5954#issuecomment-1577183123)) and I addressed review comments with an new appended 1st fixup commit: "[review] core: interrupt: interrupt chip framework"
> 
> Fixed Atmel driver issues ([#5954 (comment)](https://github.com/OP-TEE/optee_os/pull/5954#issuecomment-1576826879)) with 2 fixup commits: "[review] drivers: atmel_piobu: upgrade to new interrupt framework" "[review] drivers: atmel_wdt: upgrade to new interrupt framework"
> 
> I appended 3 new commits to remove the old interrupt API functions `itr_xxx()`: "[review] plat-stm32mp1: upgrade to new interrupt framework" "core: notif: upgrade to new interrupt framework" "core: interrupt: remove old API functions"
> 
> Last, I appendeded another fixup commit "[review] core: interrupt: interrupt chip framework" to mask unhandled interrupt instead of disabling them. This is needed as mask operation handler is expected to be atomic while disable operation handler does not.
> 
> @tprrt: cuold you test this updated series? Atmel SAIC controller shoulld be provided the expect configuration data now.

Hello @etienne-lms,
It's good now, no more issue occurs on sama5.
Kind regards,
Thomas

core/kernel/dt_driver.c
etienne-lms: okay, i'll add a fixup commit.

etienne-lms:
thanks @jenswi-linaro for the review.
thanks @tprrt for the test.
I've added few more fixup commits. I'll create P-Rs for the 1st patches of this series.

core/kernel/dt_driver.c
```diff
@@ -319,6 +324,10 @@ void *dt_driver_device_from_node_idx_prop_phandle(const char *prop_name,
 	}
 
 	prop_index *= dt_driver_provider_cells(prv);
+	if (prop_index * sizeof(*prop) >= (size_t)len) {
+		*res = TEE_ERROR_GENERIC;
```
etienne-lms, line 328: should be `TEE_ERROR_ITEM_NOT_FOUND`

etienne-lms:
I have created https://github.com/OP-TEE/optee_os/pull/6087 and https://github.com/OP-TEE/optee_os/pull/6088 for the 5 first commits of the current series.

core/kernel/dt_driver.c
etienne-lms: done

core/kernel/interrupt.c
etienne-lms: Addressed through helper function `interrupt_add_configure_handler()`.

core/include/kernel/interrupt.h
etienne-lms: Actually, I must admit I did not find any. I was confused by documentation that sometimes uses mask and enable terminology for the same configuration means. Do you think we can remove mask/unmask and keep only the enable/disable operators? Until someone request we splits these couples.

jenswi-linaro: Thanks for your effort. Yes, I think we can remove mask/unmask for now.

borneoa: @jenswi-linaro @etienne-lms
I have a real use case that requires mask/unmask to be split from enable/disable.

An I2C based consumer sends itr's through GPIO to the itr hierarchy made by:
- edge triggered EXTI itr controller, that in turn sends its itr to
- level triggered ARM GIC.

To ack the itr in the consumer through I2C I have to rely on a bottom half using async notif.

I cannot "disable" the itr in the EXTI otherwise (due to edge base trigger) I will loose a new incoming itr
between the consumer's ack and the EXTI "enable".
So I have to rely on the "disable" of GIC, while keeping EXTI enabled.

The top/bottom half sequence is:
- top half:
  .1) "mask" itr, that means "ack the edge" on EXTI and "disable" the itr on GIC
  .2) notify bottom half
- bottom half:
  .3) ack the consumer's itr through I2C
  .4) "unmask" itr, that "does nothing" on EXTI and "enables" GIC

A new edge itr occurring between 3) and 4) is recorded by EXTI but propagated by GIC only after 4).

Level triggered itr controllers don't need to differentiate enable/disable from mask/unmask
because they will not loose an incoming level itr; if the level is still there, it will trigger the itr
once controller re-enables.
Edge triggered itr controllers need mask to block the propagation of new itr and ack internally the old
itr, while still sensing for new edges. Unmask will eventually propagate the newly recorded itr.

In my hierarchy above, EXTI does not have a mask feature, but relies on GIC for this.

GIC should register mask/unmask that corresponds to current gic_op_enable/gic_op_disable, as
already done in patch "driver: gic: implement mask/unmask handler".
EXTI will register enable/disable that operates as real enable/disable of EXTI and also
register an empty "unmask" and a "mask" that ack the itr.

etienne-lms: Thank @borneoa for the details.
So I think it is better to keep the 4 mask/unmaks/enable/disable handlers and relates API function `interrupt_xxx()`. An interrupt controller that has registered to handler to the GIC can offer real interrupt mask/unmask/enable/disable services to its own consumers.

I have created P-R https://github.com/OP-TEE/optee_os/pull/6096 to  specifically review these API functions.

etienne-lms:
I created 2 P-Rs:
https://github.com/OP-TEE/optee_os/pull/6096: 5 commits for the new interrupt API functions `interrupt_xxx()`
https://github.com/OP-TEE/optee_os/pull/6097: 2 commits for the dt_driver registering/retrieving

Will remain the updates in few drivers and platform code to bump to the new API and a last cmmit to remove the old interrupt API functions `itr_xxx()`.

core/include/kernel/interrupt.h
etienne-lms: addressed by https://github.com/OP-TEE/optee_os/pull/6098.

etienne-lms:
As #6096 has been merged, I created https://github.com/OP-TEE/optee_os/pull/6113 to finalize the new interrupt API.

The part related to use of DT means to register interrupt driver providers and for consumer to get their interrupt is still pending: once https://github.com/OP-TEE/optee_os/pull/6097 is merged, I will create a P-R for a last commit to make GIC driver to register as a `DT_DRIVER_INTERRUPT`.

Once, done, all change proposed in this P-R will have been addressed.

The next step regarding interrupt management wil likely be support for bottom half interrupt handlers, see https://github.com/OP-TEE/optee_docs/pull/198.

core/kernel/interrupt.c
```diff
@@ -62,79 +90,223 @@ void itr_handle(size_t it)
...
+}
+
+TEE_Result interrupt_add_configure_handler(struct itr_handler *hdl,
+					   uint32_t type, uint32_t prio)
+{
+	struct itr_handler *h = NULL;
+
+	assert(hdl && hdl->chip->ops);
+
+	SLIST_FOREACH(h, &hdl->chip->handlers, link) {
```
ldebieve, line 113: Should it be a SAFE loop has the remove will update the list?

etienne-lms: This part was merged with https://github.com/OP-TEE/optee_os/pull/6096.
That said, I agree, theree should be some protection here in case interrupt handlers are added/removed at runtime.

PR #5955, opened 2023-04-12
plat-zynqmp: fixes interrupt controller

### LINKED ISSUE #5949: Erroneous ZynqMP Interrupt IDs
Have there been any previous attempts at using interrupt within OP-TEE OS on the ZynqMP platform? It appears that there is [some initialization](https://github.com/OP-TEE/optee_os/blob/32b94ed4b63fdac094b1a6b7719f7cb8979695f6/core/arch/arm/plat-zynqmp/main.c#L83) that takes place, but there don't seem to be any devices that currently enable any interrupts. It doesn't even look like the GIC is being added to the kernel's interupt handling mechanisms, [as done for other platforms](https://github.com/OP-TEE/optee_os/blob/32b94ed4b63fdac094b1a6b7719f7cb8979695f6/core/arch/arm/plat-zynq7k/main.c#L148).

I'm attempting to enable interrupts for GEM3, and have done so following a similar method to what is done for UART interrupts on other platforms. However, after enabling interrupts within the GEM itself, it seems like interrupts are coming in on ID 0 (SGI1), rather than the expected ID 95 (GEM3). This causes additional problems since the software generated interrupt belongs to group 1, and the kernel interrupt handling has a number of assertions that interrupts belong to group 0. It seems like what is happening is:
- Interrupt gets registered with kernel during boot
- GEM gets enabled, and interrupts start being generated on ID 95
- Kernel misinterprets interrupt as being ID 0
- SGI1 has no handler registered, so kernel attempts to disable the interrupt
- Disable function asserts that the interrupt being disabled is a group 0 interrupt. Since SGI1 is group 1, the assert fails and causes a panic

Notably, the panic doesn't actually appear in the console unless stepping through the code. Possibly because the interrupt is being constantly re-handled, since no actual handler gets called, and the disable call also fails. If you let the system continue running freely, execution will eventually end up out in undefined memory.

Using a debugger, it appears that the proper interrupt is being asserted in the pending interrupt registers. Are there any thoughts on why OP-TEE might be trying to handle the wrong interrupt?

zach-clark:
Updates GICC_OFFSET to account for the already-offset GIC_BASE. Additionally initializes the interrupt controller with a pointer to the interrupt chip.

Fixes #5949

### DIFF
--- a/core/arch/arm/plat-zynqmp/main.c
+++ b/core/arch/arm/plat-zynqmp/main.c
@@ -82,6 +82,7 @@ void main_init_gic(void)
 	/* On ARMv8, GIC configuration is initialized in ARM-TF */
 	gic_init_base_addr(&gic_data, GIC_BASE + GICC_OFFSET,
 			   GIC_BASE + GICD_OFFSET);
+	itr_init(&gic_data.chip);
 }
 
 void itr_core_handler(void)
--- a/core/arch/arm/plat-zynqmp/platform_config.h
+++ b/core/arch/arm/plat-zynqmp/platform_config.h
@@ -70,7 +70,7 @@
 #define UART1_CLK_IN_HZ		100000000
 
 #define GICD_OFFSET		0
-#define GICC_OFFSET		0x20000
+#define GICC_OFFSET		0x10000
 
 #elif defined(PLATFORM_FLAVOR_ultra96)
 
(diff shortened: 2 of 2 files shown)


ricardosalveti:
`Reviewed-by: Ricardo Salveti <ricardo@foundries.io>`

PR #5957, opened 2023-04-13
core：add  a new pbkdf2 implementation

yuzexiyzx:
Add a new cryptographically secure pbkdf2 function known as pbkdf2_f
. The implementation is based on hardware, if support, which is full
process uninstallation with mac simulations.


yuzexiyzx:
I can add a derive key driver after pull request "drivers: implement SEC driver" merged.  Before that，I wonder if my change is ok without driver part.

core/crypto/pbkdf2_hw.c
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/* Copyright (c) 2022, HiSilicon Limited. */
+
+#include <crypto/crypto.h>
+#include <crypto/pbkdf2_support.h>
+
+/* This is a hardware pbkdf2 function, without mac simulation */
+TEE_Result __weak hw_pbkdf2(uint32_t hash_id __unused,
```
jforissier, line 8: I don't think this weak function is needed, because when `CFG_CRYPTO_HW_PBKDF2=y` there will be a platform-specific implementation, and when `CFG_CRYPTO_HW_PBKDF2 != y` it is not used.

jforissier (commented):
Comments for commit "core: add new pbkdf2 implementation".

>  core: add new pbkdf2 implementation

PBKDF2 (upper case)
In fact this commit does not add the implementation, just the config flag and the function prototype. Therefore it would be better described as "core: add hook for accelerated PBKDF2". And then it would be nice to have an actual implementation for a Hisilicon platform.

> Add a new cryptographically secure pbkdf2 function known as pbkdf2_f.

Any link to some specification? Is this function just one implementation by Hisilicon of the PBKDF2 spec, or is it different?

> The implementation is based on hardware, if support, 

OK

> which is full
> process uninstallation with mac simulations.

I don't understand this part.

jforissier:
Commit "lib: add pbkdf2 derive type define" looks OK (with s/pbkdf2/PBKDF2/) but it does not make much sense until it is actually used by another commit.

jenswi-linaro:
@yuzexiyzx, do you plan to resume this PR?

yuzexiyzx:
> @yuzexiyzx, do you plan to resume this PR?
Maybe later.I think it would be better to resume this PR after pull request "drivers: implement SEC driver" merged. Then, I will Add a derive key driver in the HW crypto API.

jenswi-linaro:
OK, I leave up to you to keep this PR open then.

PR #5960, opened 2023-04-13
core: arm64: virt: fix adr link error

jenswi-linaro:
Replace the two remaining "adr" instructions with "adr_l" to avoid link error with virtualization and debugging enabled:
out/arm/core/arch/arm/kernel/entry_a64.o: in function `clear_bss': core/arch/arm/kernel/entry_a64.S:237:(.text._start+0x8c): relocation truncated to fit: R_AARCH64_ADR_PREL_LO21 against symbol `__nex_bss_start' defined in .bss.mempool_default section in out/arm/core/all_objs.o core/arch/arm/kernel/entry_a64.S:238:(.text._start+0x90): relocation truncated to fit: R_AARCH64_ADR_PREL_LO21 against symbol `__nex_bss_end' defined in .bss.mempool_default section in out/arm/core/all_objs.o

### DIFF
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -234,8 +234,8 @@ clear_bss:
 	 * Clear .nex_bss, this code obviously depends on the linker keeping
 	 * start/end of .bss at least 8 byte aligned.
 	 */
-	adr	x0, __nex_bss_start
-	adr	x1, __nex_bss_end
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
This patch is the same fix as in "core: arm64: use adr_l for __nex_bss_start and __nex_bss_end" in https://github.com/OP-TEE/optee_os/pull/5359. Closing this as a duplicate.

Issue #5946, opened 2023-04-06
Please help understand security implication of TEE_AEDecryptFinal()  defined by the GP spec.


zlhk100:
Hi team,

When I try to figure the reason of TA panic,  it brought me to[ this line](https://github.com/OP-TEE/optee_os/blob/master/core/tee/tee_svc_cryp.c#L4093):
inside TEE_AEDecryptFinal().    Somehow,  due to tag parameter was passed in from REE side share memory same as cipher-text source and nonce buffer/parameters.   Then memory access sanity-check marks CCM TAG buffer as ACCESS_DENIED.

After checking GP internal API Spec(Thanks @jforissier for sharing and points),  it does define this API's behaviour as implemented.  

GP Spec seems viewing TAG different from cipher-text src.     Essentially meaning that:
from API user's perspective:  
     1.  you can directly pass in any memory type in for cipher-text src.
      2. For tag field,  you need to do a memcpy() to TA owned private memory before passing into this syscall.

However from **security coding perspective**,  how to make sense out of it?     

Would be interesting to know it's motivation since CCM/GCM like cipher means to mitigate chosen cipher-text attack.     I'm aware of TOCTOU attack but seems not relevant here,  please correct me if i'm wrong?

Maybe some side channel attacks existing if tag field is under bad actor's control?

Look forward to your feedbacks!

jenswi-linaro:
Comparing with `TEE_MACCompareFinal()` it is curious that `mac` has the `[inbuf]` attribute instead.

jforissier:
This is something that should be clarified with GlobalPlatform I believe. @etienne-lms I think you have access to the private GP mailing list for the TEE group, any chance you could ask?

jforissier:
> Actually, I saw this is already planned to be fixed next Core API, so I think we can fix the implementation

Thanks for checking :+1: I agree. @zlhk100 would you like to create a pull request?

zlhk100:
thanks @jforissier @etienne-lms @jenswi-linaro for checking and commenting.  Sure,  I will create a PR on this fix as commented.

zlhk100:
https://github.com/OP-TEE/optee_os/pull/5968.  is created and under reviewing.

PR #5968, opened 2023-04-19
core: crypto: change TEE_AEDecryptFinal() tag param's attribute

zlhk100:
Refer to https://github.com/OP-TEE/optee_os/issues/5946. Due to tag parameter was passed in from REE side share memory same as cipher-text source and nonce buffer/parameters. Then memory access sanity-check marks CCM TAG buffer as ACCESS_DENIED, which triggers user TA panic.

Change tag parameter's attribute from [in] to [inbuf], aligned with next GP TEE Internal core API specification Update.


zlhk100:
@jforissier @etienne-lms please help reviewing.   thanks, Lei

jforissier:
Hello @zlhk100, thanks for the patch. LGTM.

jenswi-linaro (commented):

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with comments on commit message:
Use `Fixes:` tag or `Link:` tag rather than plain text `Refer to ...`.
There is no official "next GP TEE Internal core API specification" so (IMO) you should rephrase that sentence, see suggestion below:
```patch
 core: crypto: change TEE_AEDecryptFinal() tag param's attribute
 
-Refer to https://github.com/OP-TEE/optee_os/issues/5946.
 Due to tag ...

-Change tag parameter's attribute from [in] to [inbuf], aligned
-with next GP TEE Internal core API specification Update.
+Change tag parameter's attribute from [in] to [inbuf]. This fix is
+expected to be addressed in next GP TEE Internal Core API specification.
 
+Fixes: https://github.com/OP-TEE/optee_os/issues/5946
 Signed-off-by: ...
```

etienne-lms:
My mistake: replace `Fixes:` tag with a `Link:` tag, it should fix the issue reported by CI code style test.
You should also remove the empty line between that tag and your `Signed-off-by:` tag in commit message. All commit message tags should be contiguous (IIUC).

PR #5977, opened 2023-04-24
CRC32 HW support discovery

balint-dobszay-arm:
Add support for discovering if the CRC32 instruction family is implemented by the PE.
Convey this information to Secure Partitions through the SP manifest DT.

### DIFF
--- a/core/arch/arm/include/arm64.h
+++ b/core/arch/arm/include/arm64.h
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
  * Copyright (c) 2015, Linaro Limited
+ * Copyright (c) 2023, Arm Limited
  */
 #ifndef ARM64_H
 #define ARM64_H
@@ -237,6 +238,11 @@
 #define FEAT_MTE2_IMPLEMENTED		U(0x2)
 #define FEAT_MTE3_IMPLEMENTED		U(0x3)
 
+#define ID_AA64ISAR0_EL1_CRC32_MASK	UL(0xf)
+#define ID_AA64ISAR0_EL1_CRC32_SHIFT	U(16)
+#define FEAT_CRC32_NOT_IMPLEMENTED	U(0x0)
+#define FEAT_CRC32_IMPLEMENTED		U(0x1)
+
 #define ID_AA64ISAR1_GPI_SHIFT		U(28)
 #define ID_AA64ISAR1_GPI_MASK		U(0xf)
 #define ID_AA64ISAR1_GPI_NI		U(0x0)
@@ -427,6 +433,7 @@ DEFINE_U64_REG_READ_FUNC(par_el1)
 DEFINE_U64_REG_WRITE_FUNC(mair_el1)
 
 DEFINE_U64_REG_READ_FUNC(id_aa64pfr1_el1)
+DEFINE_U64_REG_READ_FUNC(id_aa64isar0_el1)
 DEFINE_U64_REG_READ_FUNC(id_aa64isar1_el1)
 DEFINE_REG_READ_FUNC_(apiakeylo, uint64_t, S3_0_c2_c1_0)
 DEFINE_REG_READ_FUNC_(apiakeyhi, uint64_t, S3_0_c2_c1_1)
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -1181,6 +1181,36 @@ err_unmap:
 	return res;
 }
 
+/*
+ * Note: this function is called only on the primary CPU. It assumes that the
+ * features present on the primary CPU are available on all of the secondary
+ * CPUs as well.
+ */
+static TEE_Result handle_hw_features(void *fdt)
+{
+	uint32_t val __maybe_unused = 0;
+	TEE_Result res = TEE_SUCCESS;
+	int node = 0;
+
+	/*
+	 * HW feature descriptions are optional in the SP manifest, it's not an
+	 * error if we don't find any.
+	 */
+	node = fdt_node_offset_by_compatible(fdt, 0, "arm,hw-features");
+	if (node < 0)
+		return TEE_SUCCESS;
+
+	/* Modify the crc32 property only if it's already present */
+	if (!sp_dt_get_u32(fdt, node, "crc32", &val)) {
+		res = fdt_setprop_u32(fdt, node, "crc32",
+				      feat_crc32_implemented());
+		if (res)
+			return res;
+	}
+
+	return TEE_SUCCESS;
+}
+
 static TEE_Result sp_init_uuid(const TEE_UUID *bin_uuid, const void * const fdt)
 {
 	TEE_Result res = TEE_SUCCESS;
@@ -1245,6 +1275,10 @@ static TEE_Result sp_first_run(struct sp_session *sess)
 		if (res)
 			goto out;
 	}
+
+	res = handle_hw_features(fdt_copy);
+	if (res)
+		goto out;
 
 	ts_pop_current_session();
 
--- a/core/arch/arm/include/arm.h
+++ b/core/arch/arm/include/arm.h
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
  * Copyright (c) 2015, Linaro Limited
- * Copyright (c) 2019, Arm Limited. All rights reserved.
+ * Copyright (c) 2019-2023, Arm Limited. All rights reserved.
  */
 #ifndef ARM_H
 #define ARM_H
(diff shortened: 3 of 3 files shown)


jforissier (approved):
- For "core: arm64: add CRC32 HW support discovery":
- For "core: spmc: handle HW feature info in SP manifest":

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1181,6 +1181,31 @@ static TEE_Result handle_tpm_event_log(struct sp_ctx *ctx, void *fdt)
...
+	 * error if we don't find any.
+	 */
+	node = fdt_node_offset_by_compatible(fdt, 0, "arm,hw-features");
+	if (node < 0)
+		return TEE_SUCCESS;
+
+	/* Modify the crc32 property only if it's already present */
+	if (!sp_dt_get_u32(fdt, node, "crc32", &val)) {
+		res = fdt_setprop_u32(fdt, node, "crc32",
+				      feat_crc32_implemented());
```
jforissier, line 1201: Here you're assuming that all PEs either support or do not support CRC32, aren't you? A mixed configuration may lead to troubles otherwise. Should detection happen on boot for each core and be consolidated here instead?

balint-dobszay-arm: Yes, that's correct. I understand your concern and I was thinking about the same. The problem is that the SPs get initialized before the secondary cores are switched on, so the detection on the secondary cores would be too late already. It's an unfortunate situation, I'm not sure what would be the right solution.

jforissier: Why not initialize the SPs after all the cores are online?

balint-dobszay-arm: The SPs must be available already when control is first handed over to the NWd, since e.g. the NWd bootloader might rely on services provided by SPs. AFAIU the secondary cores are only initialized later, during Linux boot.

jforissier: > The SPs must be available already when control is first handed over to the NWd, since e.g. the NWd bootloader might rely on services provided by SPs.

OK.

> AFAIU the secondary cores are only initialized later, during Linux boot.

Correct. Then perhaps the HW capabilities in the SP could be set once for the boot CPU, then updated when all the secondary cores are up and running? The validity of this depends on how the "crc32" node is specified, and on the actual use cases of course.

balint-dobszay-arm: Normally the SP processes its config once when it boots up and it's not possible to change it later. I have to think about this, if we can come up with a mechanism to workaround this.

balint-dobszay-arm: After a bit of investigation, it looks like it should rarely or never happen in a real world system, that some cores have CRC32 and some don't. Still, we cannot rule it out completely. My proposed fix:
- when the primary core boots check if CRC is available and store the result somewhere,
- the SPs are booted with the info that's available on the primary core,
- when the secondary cores come up, check if CRC availability matches with the primary core,
- if there is a mismatch, panic.

This is rather a workaround than a solution, but at least it prevents the SPs from executing an illegal instruction in the situation when the boot core has CRC but later the SP is scheduled on a core which doesn't. All of the above could be guarded with `CFG_SECURE_PARTITION` so it doesn't affect other use cases. What do you think?

jenswi-linaro: Makes sense to me. In fact, I believe it should be a quite common approach to expect all CPUs we're executing on to be compatible with each other.

jenswi-linaro:

core/arch/arm/kernel/secure_partition.c
jforissier: @jenswi-linaro hopefully, yes. I agree with the proposed approach.

balint-dobszay-arm: Thanks for the feedback, I implemented the workaround described above. TBH the code I added looks a bit out of place, I'm wondering if it'd be better to move it to `entry_a64.S`?

core/arch/arm/kernel/boot.c
```diff
@@ -1423,6 +1423,29 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
...
+	}
+
+	/* If available on the boot core, make sure other cores match */
+	if (available) {
+		if (!feat_crc32_implemented()) {
+			EMSG("CRC32 not available on core %ld", get_core_pos());
+			panic();
+		}
+	}
+}
```
jforissier, line 1446:
```C
#else
static void check_crc32_feature(void)
{}
```
and remove the ifdefs in `boot_init_primary_late()` and `init_secondary_helper()`

balint-dobszay-arm: Done

core/arch/arm/kernel/boot.c
```diff
@@ -1423,6 +1423,33 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
 	init_sec_mon(nsec_entry);
 }
 
+#ifdef CFG_SECURE_PARTITION
+static void check_crc32_feature(void)
+{
+	static bool checked;
```
balint-dobszay-arm, line 1429: I removed the initialisation for these variables to get rid of the checkpatch warning

jforissier:

core/arch/arm/kernel/boot.c
```diff
@@ -1423,6 +1423,33 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
...
+	/* Check once (i.e. on the boot core) and store the result */
+	if (!checked) {
+		available = feat_crc32_implemented();
+		checked = true;
+		return;
+	}
+
+	/* If available on the boot core, make sure other cores match */
+	if (available) {
+		if (!feat_crc32_implemented()) {
```
jenswi-linaro, line 1441: Do we really support systems where the different CPUs we execute code on are incompatible with each other?

jenswi-linaro: If so, how are the incompatibilities bounded?

balint-dobszay-arm: I'm not sure, it is possible in theory but I don't know if it exists in real HW. As discussed in the thread above, Jerome had concerns about it, so I added this check to catch the error if we ever run on such asymmetrical system. Should I  keep it or remove it? I'm okay with both approaches, it depends on what you and @jforissier prefer.

core/arch/arm/kernel/boot.c
```diff
@@ -1423,6 +1423,33 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
...
+
+	/* Check once (i.e. on the boot core) and store the result */
+	if (!checked) {
+		available = feat_crc32_implemented();
+		checked = true;
+		return;
+	}
+
+	/* If available on the boot core, make sure other cores match */
+	if (available) {
```
etienne-lms, line 1440: `if (available != feat_crc32_implemented())`

(assuming Jens below comment is addressed in some way)

core/arch/arm/kernel/boot.c
```diff
@@ -1482,6 +1509,7 @@ void __weak boot_init_primary_late(unsigned long fdt,
 	} else {
 		init_tee_runtime();
 	}
+	check_crc32_feature();
```
etienne-lms, line 1512: I think `check_crc32_feature()` should rather be called from a `boot_final()` registered handler.

core/arch/arm/kernel/boot.c
etienne-lms: caller could pass a argument to state whether it is the primary core that calls it (from `boot_init_primary_late()` or like) or it is a secondary core (`from init_secondary_helper()`).

core/arch/arm/kernel/boot.c
jenswi-linaro: Let's wait with these kinds of checks until we know they are needed if you don't mind @jforissier.

jforissier: OK but I would at least like a comment in `handle_hw_features()` to state that it is called from the primary CPU and it assumes that if the feature is present then all secondary CPUs have the feature enabled too.

balint-dobszay-arm: Okay, I'll update it.

core/arch/arm/kernel/boot.c
balint-dobszay-arm: As discussed above, I removed this commit from the PR.

core/arch/arm/kernel/boot.c
balint-dobszay-arm: As discussed above, I removed this commit from the PR.

core/arch/arm/include/arm64.h
```diff
@@ -456,4 +463,3 @@ DEFINE_REG_WRITE_FUNC_(icc_igrpen1, uint32_t, S3_0_C12_C12_7)
 #endif /*__ASSEMBLER__*/
 
 #endif /*ARM64_H*/
-
```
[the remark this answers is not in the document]
balint-dobszay-arm: Fixed

core/arch/arm/kernel/boot.c
balint-dobszay-arm: Done

balint-dobszay-arm:
Addressed all comments, rebased on master and applied tags.

Issue #5924, opened 2023-03-28
Virtualization and core_mmu_add_mapping


patrickjauernig:
Hi,

I am using a setup with Xen + OP-TEE with CFG_NS_VIRTUALIZATION. While Xen is initializing (before the first VM booted), I want to call OP-TEE (core) and use cryptographic functions. However, they do not have a "context" at this time (and hence, fail), so I am creating a VM (corresponding to the hypervisor) to create a context (shortened):

```
thread_get_core_local()->curr_thread = 0;
virt_guest_created(id);
virt_set_guest(id);
```

This works, but then `core_mmu_add_mapping` and `core_mmu_remove_mapping` (with the `MEM_AREA_NSEC_SHM` type) do not work correctly anymore when trying to (un)map Xen memory (I think it is mainly due to the absence of MEM_AREA_RES_VASPACE). 

Is there anything to create the context for the cryptographic functions without creating a VM, or to fix the memory mappings? Sorry, I could not find any information on this topic.

jenswi-linaro:
It sounds like you're on the right track. Why isn't `MEM_AREA_RES_VASPACE` available?

patrickjauernig:
It was there before I created the VM but is not there anymore after the "hypervisor VM" was created and set as guest. I guess it is because (based on the current implementation) VM contexts are not expected to need it? Based on what I saw, virtualization is heavily focusing on TA execution.

As a side observation, in the "hypervisor VM", `core_mmu_add_mapping` does not work the first time I invoke it (taking [this branch](https://github.com/OP-TEE/optee_os/blob/db8ca286ea44786a19e344ca02b47cc8be04924e/core/mm/core_mmu.c#L2050)), i.e., `find_map_by_va` returns NULL . When invoking it a second time, it takes the [other branch](https://github.com/OP-TEE/optee_os/blob/db8ca286ea44786a19e344ca02b47cc8be04924e/core/mm/core_mmu.c#L2058), and I can get the map containing the virtual address. 

Then, when removing the mapping later, [it errors out because it cannot find MEM_AREA_RES_VASPACE](https://github.com/OP-TEE/optee_os/blob/db8ca286ea44786a19e344ca02b47cc8be04924e/core/mm/core_mmu.c#L1939).

jenswi-linaro:
We've done just enough to have the current use cases working with virtualization. By the way, what you call "hypervisor VM" we call partition.

[...]

PR #5978, opened 2023-04-24
core: thread: Add support for canary value randomization

mingxien:
Currently hardcoded magic number is used as thread stack canary, an attacker with full control over the overflow can embed the hardcoded canary value on the right location to bypass the overflow detection.

To add extra layer of security, redefine the canary value as variable, such that the canary can be initialized during runtime.

Weak function plat_set_thread_canary() can be overridden by platform to initialize the thread canary with random numbers from TRNG hardware.

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -120,12 +120,14 @@ __weak unsigned long plat_get_aslr_seed(void)
 	return 0;
 }
 
-#if defined(_CFG_CORE_STACK_PROTECTOR)
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
 /* Generate random stack canary value on boot up */
-__weak uintptr_t plat_get_random_stack_canary(void)
+__weak void plat_get_random_stack_canaries(void *buf, size_t ncan, size_t size)
 {
-	uintptr_t canary = 0xbaaaad00;
 	TEE_Result ret = TEE_ERROR_GENERIC;
+	size_t i = 0;
+
+	assert(buf && ncan && size);
 
 	/*
 	 * With virtualization the RNG is not initialized in Nexus core.
@@ -133,17 +135,20 @@ __weak uintptr_t plat_get_random_stack_canary(void)
 	 */
 	if (IS_ENABLED(CFG_NS_VIRTUALIZATION)) {
 		IMSG("WARNING: Using fixed value for stack canary");
-		return canary;
+		memset(buf, 0xab, ncan * size);
+		goto out;
 	}
 
-	ret = crypto_rng_read(&canary, sizeof(canary));
+	ret = crypto_rng_read(buf, ncan * size);
 	if (ret != TEE_SUCCESS)
 		panic("Failed to generate random stack canary");
 
+out:
 	/* Leave null byte in canary to prevent string base exploit */
-	return canary & ~0xffUL;
+	for (i = 0; i < ncan; i++)
+		*((uint8_t *)buf + size * i) = 0;
 }
-#endif /*_CFG_CORE_STACK_PROTECTOR*/
+#endif /* _CFG_CORE_STACK_PROTECTOR || CFG_WITH_STACK_CANARIES */
 
 /*
  * This function is called as a guard after each smc call which is not
@@ -1385,6 +1390,16 @@ void init_tee_runtime(void)
 	 */
 	thread_init_core_local_pauth_keys();
 	thread_init_thread_pauth_keys();
+
+	/*
+	 * Reinitialize canaries around the stacks with crypto_rng_read().
+	 *
+	 * TODO: Updating canaries when CFG_NS_VIRTUALIZATION is enabled will
+	 * require synchronization between thread_check_canaries() and
+	 * thread_update_canaries().
+	 */
+	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION))
+		thread_update_canaries();
 }
 
 static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
--- a/core/arch/arm/kernel/entry_a32.S
+++ b/core/arch/arm/kernel/entry_a32.S
@@ -571,9 +571,15 @@ shadow_stack_access_ok:
 
 #ifdef _CFG_CORE_STACK_PROTECTOR
 	/* Update stack canary value */
-	bl	plat_get_random_stack_canary
+	sub	sp, sp, #0x8
+	mov	r0, sp
+	mov	r1, #1
+	mov	r2, #0x4
+	bl	plat_get_random_stack_canaries
+	ldr	r0, [sp]
 	ldr	r1, =__stack_chk_guard
 	str	r0, [r1]
+	add	sp, sp, #0x8
 #endif
 
 	/*
--- a/core/include/kernel/boot.h
+++ b/core/include/kernel/boot.h
@@ -70,8 +70,16 @@ void plat_cpu_reset_early(void);
 void plat_primary_init_early(void);
 unsigned long plat_get_aslr_seed(void);
 unsigned long plat_get_freq(void);
-#if defined(_CFG_CORE_STACK_PROTECTOR)
-uintptr_t plat_get_random_stack_canary(void);
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
+/*
+ * plat_get_random_stack_canaries() - Get random values for stack canaries.
+ * @buf:	Pointer to the buffer where to store canaries
+ * @ncan:	The number of canaries to generate.
+ * @size:	The size (in bytes) of each canary.
+ *
+ * This function has a __weak default implementation.
+ */
+void plat_get_random_stack_canaries(void *buf, size_t ncan, size_t size);
 #endif
 void arm_cl2_config(vaddr_t pl310);
 void arm_cl2_enable(vaddr_t pl310);
--- a/core/kernel/thread.c
+++ b/core/kernel/thread.c
@@ -8,6 +8,7 @@
 #include <config.h>
 #include <crypto/crypto.h>
 #include <kernel/asan.h>
+#include <kernel/boot.h>
 #include <kernel/lockdep.h>
 #include <kernel/misc.h>
 #include <kernel/panic.h>
@@ -31,8 +32,8 @@ struct thread_core_local thread_core_local[CFG_TEE_CORE_NB_CORE] __nex_bss;
  */
 
 #ifdef CFG_WITH_STACK_CANARIES
-#define START_CANARY_VALUE	0xdededede
-#define END_CANARY_VALUE	0xabababab
+static uint32_t start_canary_value = 0xdedede00;
+static uint32_t end_canary_value = 0xababab00;
 #define GET_START_CANARY(name, stack_num) name[stack_num][0]
 #define GET_END_CANARY(name, stack_num) \
 	name[stack_num][sizeof(name[stack_num]) / sizeof(uint32_t) - 1]
@@ -81,8 +82,8 @@ void thread_init_canaries(void)
 		uint32_t *start_canary = &GET_START_CANARY(name, n);	\
 		uint32_t *end_canary = &GET_END_CANARY(name, n);	\
 									\
-		*start_canary = START_CANARY_VALUE;			\
-		*end_canary = END_CANARY_VALUE;				\
+		*start_canary = start_canary_value;			\
+		*end_canary = end_canary_value;				\
 	}
 
 	INIT_CANARY(stack_tmp);
@@ -92,6 +93,27 @@ void thread_init_canaries(void)
 #endif
 #endif/*CFG_WITH_STACK_CANARIES*/
 }
+
+#if defined(CFG_WITH_STACK_CANARIES)
+void thread_update_canaries(void)
+{
+	uint32_t canary[2] = { };
+	uint32_t exceptions = 0;
+
+	plat_get_random_stack_canaries(canary, ARRAY_SIZE(canary),
+				       sizeof(canary[0]));
+
+	exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
+
+	thread_check_canaries();
+
+	start_canary_value = canary[0];
+	end_canary_value = canary[1];
+	thread_init_canaries();
+
+	thread_unmask_exceptions(exceptions);
+}
+#endif
 
 #define CANARY_DIED(stack, loc, n, addr) \
 	do { \
@@ -108,28 +130,28 @@ void thread_check_canaries(void)
 
 	for (n = 0; n < ARRAY_SIZE(stack_tmp); n++) {
 		canary = &GET_START_CANARY(stack_tmp, n);
-		if (*canary != START_CANARY_VALUE)
+		if (*canary != start_canary_value)
 			CANARY_DIED(stack_tmp, start, n, canary);
 		canary = &GET_END_CANARY(stack_tmp, n);
-		if (*canary != END_CANARY_VALUE)
+		if (*canary != end_canary_value)
 			CANARY_DIED(stack_tmp, end, n, canary);
 	}
 
 	for (n = 0; n < ARRAY_SIZE(stack_abt); n++) {
 		canary = &GET_START_CANARY(stack_abt, n);
-		if (*canary != START_CANARY_VALUE)
+		if (*canary != start_canary_value)
 			CANARY_DIED(stack_abt, start, n, canary);
 		canary = &GET_END_CANARY(stack_abt, n);
-		if (*canary != END_CANARY_VALUE)
+		if (*canary != end_canary_value)
 			CANARY_DIED(stack_abt, end, n, canary);
 	}
 #if !defined(CFG_WITH_PAGER) && !defined(CFG_NS_VIRTUALIZATION)
 	for (n = 0; n < ARRAY_SIZE(stack_thread); n++) {
 		canary = &GET_START_CANARY(stack_thread, n);
-		if (*canary != START_CANARY_VALUE)
+		if (*canary != start_canary_value)
 			CANARY_DIED(stack_thread, start, n, canary);
 		canary = &GET_END_CANARY(stack_thread, n);
-		if (*canary != END_CANARY_VALUE)
+		if (*canary != end_canary_value)
 			CANARY_DIED(stack_thread, end, n, canary);
 	}
 #endif
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -396,9 +396,15 @@ clear_nex_bss:
 
 #ifdef _CFG_CORE_STACK_PROTECTOR
 	/* Update stack canary value */
-	bl	plat_get_random_stack_canary
+	sub	sp, sp, #0x10
+	mov	x0, sp
+	mov	x1, #1
+	mov	x2, #0x8
+	bl	plat_get_random_stack_canaries
+	ldr	x0, [sp]
 	adr_l	x5, __stack_chk_guard
 	str	x0, [x5]
+	add	sp, sp, #0x10
 #endif
 
 	/*
--- a/core/include/kernel/thread.h
+++ b/core/include/kernel/thread.h
@@ -52,6 +52,12 @@ void thread_init_canaries(void);
 void thread_init_primary(void);
 void thread_init_per_cpu(void);
 
+#if defined(CFG_WITH_STACK_CANARIES)
+void thread_update_canaries(void);
+#else
+static inline void thread_update_canaries(void) { }
+#endif
+
 struct thread_core_local *thread_get_core_local(void);
 
 /*


jenswi-linaro:
Good idea, how about instead setting new values to these after the PRNG has been initialized? That way we can avoid yet another platform-specific function needing to be implemented to keep up with security.

mingxien:
The commit has been addressed.

There are three modifications has been done in the latest commit.
1. Rename plat_'s'et_thread_canary to plat_'g'et_thread_canary.
2. Implement random canary get mechanism in plat_get_thread_canary. (reference plat_get_random_stack_canary)
3. Change the timing that thread_init_canaries is called.

jenswi-linaro:
I would really like to avoid yet another platform-specific function needing to be implemented to keep up with security.

mingxien:
> I would really like to avoid yet another platform-specific function needing to be implemented to keep up with security.
Does this mean random canary should be implemented within the thread_init_canaries instead of plat_get_thread_canary?
Otherwise, within commit 3e9f25a, the plat_set_thread_canary in core/kernel/thread.c provides basic
implementation for random canary by using crypto_rng_read if platform-specific function isn't implemented.
This could keep up with securtiy even specific platform doesn't implement plat_set_thread_canary.

jenswi-linaro:
My point is that if a platform-specific hook can be avoided because we can implement it generically instead, then that's a much preferred solution since it doesn't add to the ever-growing list of platform-specific overrides.

mingxien:
[d839074](https://github.com/OP-TEE/optee_os/commit/d8390740dd25658bfaec70827c262a32f1235733) removed the platform-specific function and relocated thread_init_canaries into init_tee_runtime().

The reason for moving thread_init_canaries to after call_initcalls is to ensure that crypto_rng_read has been initialized and is available for use.

core/kernel/thread.c
```diff
@@ -76,13 +76,22 @@ void thread_init_canaries(void)
 {
 #ifdef CFG_WITH_STACK_CANARIES
 	size_t n;
+	uint32_t canary[2] = {0xdedede00, 0xababab00};
```
etienne-lms, line 79: nitpicking on coding style: add space char after/before starting/closing braces:
`uint32_t canary[2] = { 0xdedede00, 0xababab00 };`

etienne-lms (commented):
@mingxien, could you squash all commits into a signle one?

With `CFG_NS_VIRTUALIZATION=y`, crypto layer initialization appens after `boot_init_primary_late()` completes. Could the canaries be initialized with a static value from `thread_init_primary()` as before and updated to dynamically drawn values from a `boot_final` initcall handler when `CFG_NS_VIRTUALIZATION=y`?

This change will also prevent to enable native interrupt during OP-TEE inits. We could address that once a platform requests native interrupt support during drivers initialization but I think it would be nice that canaries management are ready for this.

mingxien:
Squashed commits in [6094638](https://github.com/OP-TEE/optee_os/commit/609463860d0bfe0e5cdbb7b2795e12a6d952307e).

As far as I know, crypto initialization occurs in `tee_cryp_init` and registers `service_init`. With `CFG_NS_VIRTUALIZATION=y`, `service_init` is executed for the first standard call and occurs after `boot_final` as we can see in [initcall.h](https://github.com/OP-TEE/optee_os/blob/master/core/include/initcall.h#L64). Therefore, initializing canaries during `boot_final` might not obtain adequate random numbers.

Please correct me if I am wrong or provide more explanation.

jenswi-linaro:
I'm sorry, it seems my suggestion wasn't the best. The original proposal with a platform-specific function to provide random for the canaries seems less complicated.

The problem with an ever-growing list of functions for platforms to override can be limited somewhat by later overriding it in the configured TRNG driver if available early enough, for instance, `core/drivers/smccc_trng.c`. But that can be saved for another PR.

So if you or someone else don't mind, please change back to a platform-specific function to provide random, perhaps `void plat_early_get_random_stack_canaries(unsigned long *canary1, unsigned long *canary2)`.

mingxien:
Let's combine all our thinking and concerns together. There are three directions we'd like to head for.

1. Keep canaries initialization with a static value from `thread_init_primary` as before to prevent a platform from requesting native interrupt support during drivers initialization and add new api, perhaps `thread_update_canaries`, with interrupt mask/unmask to update canaries with random numbers from `crypto_rng_read` at the `init_tee_runtime`. The reason why we need to add `thread_update_canaries` at the [init_tee_runtime](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/boot.c#L1379-L1383) is most of implementations of TRNG initialization are initialized from an `call_initcalls` according to `optee_os/core/drivers`.
2. Add `plat_early_get_random_stack_canaries` with `crypto_rng_read` in the `thread_init_canaries`. Any platform with unavailabled `crypto_rng_read` at early boot stage must override `plat_early_get_random_stack_canaries` by platfrom-specific implementation.
3. Combine above two together. Initialize canaries from `thread_init_primary` as before with available `plat_early_get_random_stack_canaries`, otherwise static value. Then also update random canaries if canaries are set as a static value earlier.

The solution is an open mind for discussion.

jenswi-linaro:
Thanks for your patience. May I add a fourth option?
4. Piggy-back on _CFG_CORE_STACK_PROTECTOR in `entry_a{32,64}.S` and update the stack canaries with the retrieved random values using the new function `thread_update_canaries()`. `thread_update_canaries()` can temporarily mask interrupts while updating the canaries. As in option 1. above, we can start with static canary values. `plat_get_random_stack_canary()` would need to be updated to return three words or random instead of just one.

mingxien:
Extract the point of view from all. We have [995e5ef](https://github.com/OP-TEE/optee_os/commit/995e5ef90fe6161a99e5bc5af5bdf688a8bbb40e) contains three key changes.
1. `plat_get_random_stack_canary` is renamed and refactored to support to get multiple canaries. In the function, heap was applied to get random value with `crypto_rng_read` to prevent stack protector from checking stack guard failed.
2. Add new function `thread_update_canaries` with temporarily mask interrupts and insert it in the `init_tee_runtime`.
3. Change variable type of thread stack canaries to support 64/32-bits platforms.

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,41 @@ __weak unsigned long plat_get_aslr_seed(void)
...
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
 /* Generate random stack canary value on boot up */
-__weak uintptr_t plat_get_random_stack_canary(void)
+__weak void __nostackcheck plat_get_random_stack_canaries(void *buf, size_t blen)
 {
-	uintptr_t canary = 0xbaaaad00;
+	assert(buf && blen >= sizeof(uintptr_t) && !(blen % sizeof(uintptr_t)));
+
 	TEE_Result ret = TEE_ERROR_GENERIC;
+	void *canaries_buf = malloc(blen);
```
jenswi-linaro, line 128: A temporary buffer shouldn't be needed, better to update the assembly caller to supply a buffer on the stack.

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,41 @@ __weak unsigned long plat_get_aslr_seed(void)
...
+
+	if (!canaries_buf)
+		panic("Failed to malloc canaries tmp buffer");
+
+	memset(canaries_buf, 0xab, blen);
```
jenswi-linaro, line 133: Why?

core/arch/arm/kernel/boot.c
```diff
@@ -1383,6 +1394,9 @@ void init_tee_runtime(void)
 	 */
 	thread_init_core_local_pauth_keys();
 	thread_init_thread_pauth_keys();
+
+	/* Initialize canaries around the stacks with crypto_rng_read() */
```
jenswi-linaro, line 1398: Reinitialize?

core/kernel/thread.c
```diff
@@ -31,17 +32,17 @@ struct thread_core_local thread_core_local[CFG_TEE_CORE_NB_CORE] __nex_bss;
...
+static uintptr_t end_canary_value = 0xababab00;
 #define GET_START_CANARY(name, stack_num) name[stack_num][0]
 #define GET_END_CANARY(name, stack_num) \
-	name[stack_num][sizeof(name[stack_num]) / sizeof(uint32_t) - 1]
+	name[stack_num][sizeof(name[stack_num]) / sizeof(uintptr_t) - 1]
 #endif
 
 #define DECLARE_STACK(name, num_stacks, stack_size, linkage) \
-linkage uint32_t name[num_stacks] \
+linkage uintptr_t name[num_stacks] \
```
jenswi-linaro, line 43: We've used `uint32_t` on 64-bit platforms before this. If you want to change this, please do that in a separate PR to keep the focus on what's essential.

core/kernel/thread.c
```diff
@@ -61,7 +62,7 @@ DECLARE_STACK(stack_thread, CFG_NUM_THREADS, STACK_THREAD_SIZE, static);
 #define GET_STACK_BOTTOM(stack, n) ((vaddr_t)&(stack)[n] + sizeof(stack[n]) - \
 				    STACK_CANARY_SIZE / 2)
 
-const uint32_t stack_tmp_stride __section(".identity_map.stack_tmp_stride") =
+const uintptr_t stack_tmp_stride __section(".identity_map.stack_tmp_stride") =
```
jenswi-linaro, line 65: No need to change this, in fact, the assembly code depends on a `uint32_t`.

core/arch/arm/kernel/boot.c
mingxien: The platform with `CFG_NS_VIRTUALIZATION` need to fill fixed value to initialize cancaries' buffer. I'll move this inside `IS_ENABLED(CFG_VIRTUALIZATION)` case.

core/kernel/thread.c
mingxien: Got it. I'll revert these changes

mingxien:
Replied the comments.
- Revert 64-bits canaries change in [f1a77c4](https://github.com/OP-TEE/optee_os/pull/5978/commits/f1a77c4216772c33df6c53bdd6df712570d36a71).
- Supply canaries' buffer with stack and fix others reference from comments in [a2e3fbe](https://github.com/OP-TEE/optee_os/pull/5978/commits/a2e3fbe8479d3811490ef7abbfce91d43bc42300).

jenswi-linaro:
Are you still working on the review comments?

jenswi-linaro:
It looks a bit funny, but I'll give it another round.

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,34 @@ __weak unsigned long plat_get_aslr_seed(void)
 	return 0;
 }
 
-#if defined(_CFG_CORE_STACK_PROTECTOR)
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
 /* Generate random stack canary value on boot up */
-__weak uintptr_t plat_get_random_stack_canary(void)
+__weak void plat_get_random_stack_canaries(void *buf, size_t blen, size_t clen)
```
jenswi-linaro, line 123: The relationship between blen and clen isn't fully clear. How about using the `qsort()`/`bsearch()` approach instead with `size_t nmemb, size_t size` instead? No big deal though, but the parameters must be fully described.

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,34 @@ __weak unsigned long plat_get_aslr_seed(void)
...
+		memset(buf, 0xab, blen);
+		goto exit;
 	}
 
-	ret = crypto_rng_read(&canary, sizeof(canary));
+	ret = crypto_rng_read(buf, blen);
 	if (ret != TEE_SUCCESS)
 		panic("Failed to generate random stack canary");
 
+exit:
```
jenswi-linaro, line 143: In OP-TEE we would usually call this label `out:`

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,34 @@ __weak unsigned long plat_get_aslr_seed(void)
...
 	/* Leave null byte in canary to prevent string base exploit */
-	return canary & ~0xffUL;
+	for (size_t i = 0; i < blen / clen; i++)
```
jenswi-linaro, line 145: Please declare `i` at the start of a block. Variables should not be declared mixed with code.

core/arch/arm/kernel/entry_a32.S
```diff
@@ -571,9 +571,17 @@ shadow_stack_access_ok:
 
 #ifdef _CFG_CORE_STACK_PROTECTOR
 	/* Update stack canary value */
-	bl	plat_get_random_stack_canary
+	mov	r7, sp
+	sub	sp, sp, #0x8
+	bic	sp, sp, #7	/* make sure it's a multiple of 8 */
```
jenswi-linaro, line 576: This shouldn't be needed, the SP must always be a multiple of 8 when calling functions.

core/arch/arm/kernel/entry_a32.S
```diff
@@ -571,9 +571,17 @@ shadow_stack_access_ok:
 
 #ifdef _CFG_CORE_STACK_PROTECTOR
 	/* Update stack canary value */
-	bl	plat_get_random_stack_canary
+	mov	r7, sp
```
jenswi-linaro, line 574: This isn't needed, you know how much to add to the SP after the function call.

core/include/kernel/boot.h
```diff
@@ -69,8 +69,8 @@ void plat_cpu_reset_early(void);
 void plat_primary_init_early(void);
 unsigned long plat_get_aslr_seed(void);
 unsigned long plat_get_freq(void);
-#if defined(_CFG_CORE_STACK_PROTECTOR)
-uintptr_t plat_get_random_stack_canary(void);
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
+void plat_get_random_stack_canaries(void *buf, size_t blen, size_t clen);
```
etienne-lms, line 73: Please add an inline description, stating what `blen` and `clen` relate to. It should tell that the function is expected to panic upon error.

If this function is expected to always provide 2 32bit values, prefer:
`void plat_get_random_stack_canaries(uint32_t *start_canary, unint32_t *end_canary);`

core/kernel/thread.c
```diff
@@ -6,6 +6,7 @@
  */
 
 #include <config.h>
+#include <kernel/boot.h>
```
etienne-lms, line 9: move 2 lines below

core/include/kernel/boot.h
mingxien: OK, I'll add an inline description for the function.
About the definition of `plat_get_random_stack_canaries`, it's defined to obtain canaries of variable number and individual size in byte. We hope this provides convenience for further expansion of other canary functions with different requirements.
Currently, thread canaries require two 32-bit canaries, while compiler canaries require a single 32- or 64-bit canary.

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,38 @@ __weak unsigned long plat_get_aslr_seed(void)
...
+__weak void plat_get_random_stack_canaries(void *buf, size_t ncan, size_t size)
 {
-	uintptr_t canary = 0xbaaaad00;
 	TEE_Result ret = TEE_ERROR_GENERIC;
+	size_t i = 0;
+	size_t blen = 0;
+
+	assert(buf && ncan && size);
+
+	MUL_OVERFLOW(ncan, size, &blen);
```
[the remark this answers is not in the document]
mingxien: `MUL_OVERFLOW` is indeed unnecessary, and we removed it in [e7d0184](https://github.com/OP-TEE/optee_os/pull/5978/commits/e7d018429a4554c9c78bf66a8c6a795373ea6edd).

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,36 @@ __weak unsigned long plat_get_aslr_seed(void)
...
-#if defined(_CFG_CORE_STACK_PROTECTOR)
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
 /* Generate random stack canary value on boot up */
-__weak uintptr_t plat_get_random_stack_canary(void)
+__weak void plat_get_random_stack_canaries(void *buf, size_t ncan, size_t size)
 {
-	uintptr_t canary = 0xbaaaad00;
 	TEE_Result ret = TEE_ERROR_GENERIC;
+	size_t i = 0;
+	size_t blen = 0;
```
jenswi-linaro, line 127: I guess `blen` is unused now.

core/include/kernel/boot.h
```diff
@@ -69,8 +69,17 @@ void plat_cpu_reset_early(void);
 void plat_primary_init_early(void);
 unsigned long plat_get_aslr_seed(void);
 unsigned long plat_get_freq(void);
-#if defined(_CFG_CORE_STACK_PROTECTOR)
-uintptr_t plat_get_random_stack_canary(void);
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
+/*
+ * plat_get_random_stack_canaries() - get a specified number (ncan) of stack
+ * canaries, each canary has size in bytes.
+ * @buf:	Pointer to the buffer which stored canaries
```
etienne-lms, line 76: s/stored/stores/
maybe: "Pointer to the buffer where to store canaries"

mingxien:
The comments above have been addressed.
Remove unused variables and polish descriptions in commit [4edd063](https://github.com/OP-TEE/optee_os/pull/5978/commits/4edd0630388ae4291a63ada2825af90111333554).

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,35 @@ __weak unsigned long plat_get_aslr_seed(void)
...
-	ret = crypto_rng_read(&canary, sizeof(canary));
+	ret = crypto_rng_read(buf, ncan * size);
 	if (ret != TEE_SUCCESS)
 		panic("Failed to generate random stack canary");
 
+out:
 	/* Leave null byte in canary to prevent string base exploit */
-	return canary & ~0xffUL;
+	for (i = 0; i < ncan; i++)
+		*((uint8_t *)buf + size * i) &= ~0xffUL;
```
etienne-lms, line 147: Can replace `UL` with `U`, but even prefer `~U(0xff)` altough I think `~0xff` is enough.

core/include/kernel/boot.h
```diff
@@ -69,8 +69,17 @@ void plat_cpu_reset_early(void);
 void plat_primary_init_early(void);
 unsigned long plat_get_aslr_seed(void);
 unsigned long plat_get_freq(void);
-#if defined(_CFG_CORE_STACK_PROTECTOR)
-uintptr_t plat_get_random_stack_canary(void);
+#if defined(_CFG_CORE_STACK_PROTECTOR) || defined(CFG_WITH_STACK_CANARIES)
+/*
+ * plat_get_random_stack_canaries() - get a specified number (ncan) of stack
+ * canaries, each canary has size in bytes.
```
etienne-lms, line 75: `ncan` and `size` are already described below. I think you can simplify with:
` * plat_get_random_stack_canaries() - Get random values for stack canaries`

core/kernel/thread.c
```diff
@@ -93,6 +94,25 @@ void thread_init_canaries(void)
...
+#if defined(CFG_WITH_STACK_CANARIES)
+void thread_update_canaries(void)
+{
+	uint32_t canary[2] = { };
+	uint32_t exceptions = 0;
+
+	plat_get_random_stack_canaries(canary, ARRAY_SIZE(canary),
+				       sizeof(canary[0]));
+
+	exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
```
etienne-lms, line 106: Should canaries be checed before there are updated?

mingxien: Are you suggesting that we should check the canaries are assigned random values correctly by plat_get_random_stack_canaries, rather than the initial value of '0'?

etienne-lms: No, I mean verifiying that canaries used (load in stacks) before this function execute, are still alive before we scratch them with new canary values.

core/arch/arm/kernel/boot.c
mingxien: I attempted to use `~0xff`, but the compiler generated a warning about overflow.

>   CC      out/arm-plat-mediatek/core/arch/arm/kernel/boot.o
core/arch/arm/kernel/boot.c: In function ‘plat_get_random_stack_canaries’:
core/arch/arm/kernel/boot.c:147:35: warning: overflow in implicit constant conversion [-Woverflow]
   *((uint8_t *)buf + size * i) &= ~0xff;
                                   ^

mingxien:
Comments are addressed in [601a8f3](https://github.com/OP-TEE/optee_os/pull/5978/commits/601a8f3449f67b0e64fa2248ea95c56d10a3cf44).
- Add `thread_check_canaries`
- Replace `UL` with `U`
- Simplify descriptions

jenswi-linaro:
Looks good to me, please squash all commits into one single commit for a final review.

etienne-lms (commented):
fine with me too

mingxien:
Squashed all commits and acked-by tags collected. Thanks!

core/arch/arm/kernel/boot.c
```diff
@@ -118,30 +118,35 @@ __weak unsigned long plat_get_aslr_seed(void)
...
-	ret = crypto_rng_read(&canary, sizeof(canary));
+	ret = crypto_rng_read(buf, ncan * size);
 	if (ret != TEE_SUCCESS)
 		panic("Failed to generate random stack canary");
 
+out:
 	/* Leave null byte in canary to prevent string base exploit */
-	return canary & ~0xffUL;
+	for (i = 0; i < ncan; i++)
+		*((uint8_t *)buf + size * i) &= ~U(0xff);
```
jenswi-linaro, line 147: Since you're addressing a byte that should be zeroed, how about just assigning `0`?

mingxien: That's much simpler. Would it be better to use `0x00` to make it more readable?

etienne-lms: 0 is 0, prefer `0`.

mingxien:
> Since you're addressing a byte that should be zeroed, how about just assigning `0`?

I squashed this change in [e438920](https://github.com/OP-TEE/optee_os/pull/5978/commits/e43892035af352bd18b58852abf36d4fdc00a861).
Thanks!

etienne-lms:
@mingxien, could you rebase your patch on top of master branch tip? It should fix IBART CI test.

jenswi-linaro:
That merge commit is not the same as rebasing on top of the master branch. To rebase on the master branch you typically do:
``` bash
git fetch github
git rebase github/master
```
and you should be done

jenswi-linaro:

mingxien:

Add in the commit message.
Thanks.

etienne-lms:
thanks

jforissier:
```
I/TC: WARNING: Using fixed value for stack canary
D/TC:3 0 core_mmu_xlat_table_alloc:526 xlat tables used 9 / 11
E/TC:3   Dead canary at start of 'stack_tmp[0]' (0xcc8816c0)
E/TC:3   Panic at core/kernel/thread.c:134 <thread_check_canaries>
E/TC:3   TEE load address @ 0xcc7fc000
E/TC:3   Call stack:
E/TC:3    0xcc804658 print_kernel_stack at optee_os/core/arch/arm/kernel/unwind_arm64.c:89
E/TC:3    0xcc812af0 __do_panic at optee_os/core/kernel/panic.c:24
E/TC:3    0xcc815f58 thread_check_canaries at optee_os/core/kernel/thread.c:137
E/TC:3    0xcc8022e4 thread_handle_std_smc at optee_os/core/arch/arm/kernel/thread_optee_smc.c:56
E/TC:3    0xcc7fd798 vector_std_smc_entry at optee_os/core/arch/arm/kernel/thread_optee_smc_a64.S:40
```

:skull: :bird:

mingxien:
> ```
> I/TC: WARNING: Using fixed value for stack canary
> D/TC:3 0 core_mmu_xlat_table_alloc:526 xlat tables used 9 / 11
> E/TC:3   Dead canary at start of 'stack_tmp[0]' (0xcc8816c0)
> E/TC:3   Panic at core/kernel/thread.c:134 <thread_check_canaries>
> E/TC:3   TEE load address @ 0xcc7fc000
> E/TC:3   Call stack:
> E/TC:3    0xcc804658 print_kernel_stack at optee_os/core/arch/arm/kernel/unwind_arm64.c:89
> E/TC:3    0xcc812af0 __do_panic at optee_os/core/kernel/panic.c:24
> E/TC:3    0xcc815f58 thread_check_canaries at optee_os/core/kernel/thread.c:137
> E/TC:3    0xcc8022e4 thread_handle_std_smc at optee_os/core/arch/arm/kernel/thread_optee_smc.c:56
> E/TC:3    0xcc7fd798 vector_std_smc_entry at optee_os/core/arch/arm/kernel/thread_optee_smc_a64.S:40
> ```
> 
> 💀 🐦

Got it! We'll take a look at why canaries are dead when `CFG_NS_VIRTUALIZATION=y`.
Thanks!

HsuChihYu:
> ```
> I/TC: WARNING: Using fixed value for stack canary
> D/TC:3 0 core_mmu_xlat_table_alloc:526 xlat tables used 9 / 11
> E/TC:3   Dead canary at start of 'stack_tmp[0]' (0xcc8816c0)
> E/TC:3   Panic at core/kernel/thread.c:134 <thread_check_canaries>
> E/TC:3   TEE load address @ 0xcc7fc000
> E/TC:3   Call stack:
> E/TC:3    0xcc804658 print_kernel_stack at optee_os/core/arch/arm/kernel/unwind_arm64.c:89
> E/TC:3    0xcc812af0 __do_panic at optee_os/core/kernel/panic.c:24
> E/TC:3    0xcc815f58 thread_check_canaries at optee_os/core/kernel/thread.c:137
> E/TC:3    0xcc8022e4 thread_handle_std_smc at optee_os/core/arch/arm/kernel/thread_optee_smc.c:56
> E/TC:3    0xcc7fd798 vector_std_smc_entry at optee_os/core/arch/arm/kernel/thread_optee_smc_a64.S:40
> ```
> 
> 💀 🐦

We've looked into the issue and found the problem:

1. In current implementation, `thread_update_canaries` is called when partition is receiving the first yielding call and virt_on_stdcall() is called. As a result, `thread_update_canaries` is actually updating `start_canary_value` and `end_canary_value` inside the `.data` section of this one particular partition only.
2. When OP-TEE receive next STD SMC, `thread_check_canaries` is called before entering thread context, and it's checking the canary value on tmp stack against the `start_canary_value` and `end_canary_value` inside the `.data` section of default partition.

*************
To solve the problem, I think we can move `start_canary_value` and `end_canary_value` into ``.nex_data`` section by adding `__nex_data` tag such that these value are [shared between partitions](https://github.com/OP-TEE/optee_os/issues/5924#issuecomment-1491351013).

We've verified this patch on the QEMUv8, Xen environment and boot up ok with xtest pass.

We will update the P-R for review today. Thanks.

mingxien:
The `__nex_data` tag was added, and commit was squashed.
Thanks!

core/arch/arm/kernel/boot.c
```diff
@@ -1385,6 +1390,9 @@ void init_tee_runtime(void)
 	 */
 	thread_init_core_local_pauth_keys();
 	thread_init_thread_pauth_keys();
+
+	/* Reinitialize canaries around the stacks with crypto_rng_read() */
+	thread_update_canaries();
```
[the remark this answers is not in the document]
mingxien: Yes, calling `thread_update_canaries` once with `CFG_NS_VIRTUALIZATION=y` should be sufficient. We'll add an initialized flag for checking purposes. This change will be submitted at a later time.
Thanks.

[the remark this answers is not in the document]
jenswi-linaro: Better disable this call for `CFG_NS_VIRTUALIZATION=y` since it can break things.

core/kernel/thread.c
```diff
@@ -98,20 +99,29 @@ void thread_init_canaries(void)
...
 {
 	uint32_t canary[2] = { };
-	uint32_t exceptions = 0;
+	static bool inited __nex_data = false;
+	uint32_t old_itr_status = 0;
+
+	old_itr_status = cpu_spin_lock_xsave(&thread_canary_lock);
+
+	if (inited) {
+		IMSG("The thread canaries have been updated");
```
etienne-lms, line 108: is the message needed?

core/kernel/thread.c
```diff
@@ -98,20 +99,29 @@ void thread_init_canaries(void)
 void thread_update_canaries(void)
 {
 	uint32_t canary[2] = { };
-	uint32_t exceptions = 0;
+	static bool inited __nex_data = false;
```
etienne-lms, line 102: suggestion: s/inited/updated/g

mingxien:
Because the update of `start_canary_value`/`end_canary_value` and the canary value on stack is not atomic, and `thread_check_canaries` is performed at the very beginning of SMC entry point, it's possible that core 0 is in the middle of updating canaries and at the same time other core is doing `thread_check_canaries`, which would fail when `CFG_NS_VIRTUALIZATION=y`.
One possible solution is that we modify `thread_update_canaries` and `thread_check_canaries` such that only the stack of the running core is updated/checked when `CFG_NS_VIRTUALIZATION=y`.
But until we fully figure out how to safely update the canaries, it make sense to only update canaries when `CFG_NS_VIRTUALIZATION=n` and left `CFG_NS_VIRTUALIZATION` support in another P-R.

I rebase the change wchich is same as [ee775f6](https://github.com/OP-TEE/optee_os/commit/ee775f64c953e2ad47349116f41348570a84003d), and disable the function `thread_update_canaries` for `CFG_NS_VIRTUALIZATION=y`.

jenswi-linaro:
Looks good, please squash the commits.

mingxien:
The commits have been squashed. Thanks!

core/arch/arm/kernel/boot.c
```diff
@@ -1385,6 +1390,10 @@ void init_tee_runtime(void)
 	 */
 	thread_init_core_local_pauth_keys();
 	thread_init_thread_pauth_keys();
+
+	/* Reinitialize canaries around the stacks with crypto_rng_read() */
```
jforissier, line 1394: It would be nice to say why the canaries are not updated when virtualization is enabled.

mingxien:
Add comments to descript why the canaries are not updated when `CFG_NS_VIRTUALIZATION=y`.
Thanks.

core/arch/arm/kernel/boot.c
```diff
@@ -1385,6 +1390,16 @@ void init_tee_runtime(void)
 	 */
 	thread_init_core_local_pauth_keys();
 	thread_init_thread_pauth_keys();
+
+	/*
+	 * Reinitialize canaries around the stacks with crypto_rng_read().
+	 * For virtualization, we do not update canaries because otherwise we
+	 * will need a lock to avoid racing between thread_check_canaries
+	 * operate on nexus and thread_update_canaries operate on partition,
```
jforissier, line 1398: thread_update_canaries()

core/arch/arm/kernel/boot.c
```diff
@@ -1385,6 +1390,16 @@ void init_tee_runtime(void)
...
+	 * which will induce huge performance penalty.
```
jforissier, line 1399: Really? It is hard to believe. My understanding was that is was more of a "TODO" thing, left for later...

HsuChihYu: > 

Yes, we believe there must be ways to get it done, we just don't have a clever solution yet and rather to save it for another P-R.
We will rephrase the comment, such that it won't discourage other contributor from adding support of update canaries when `CFG_NS_VIRTUALIZATION` is enabled in the future.

jforissier: Very good, thanks

mingxien:
We rephrased the comment for `thread_update_canaries`.

mingxien:
@jenswi-linaro @etienne-lms, it seems like both of you are ok on this P-R, may I add your reviewed by tag in the commit message again?

jenswi-linaro:

mingxien:
Tags applied, thanks.

jforissier:
@etienne-lms has acknowledged ([link](https://github.com/OP-TEE/optee_os/pull/5978#issuecomment-1592933217)) and nothing significant has changed since, so you may add: `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`. Thanks.

mingxien:
Tags updated, thanks.

etienne-lms (approved):

PR #5980, opened 2023-04-25
Cbmem console

Narflex:
This adds a console driver to OP-TEE for logging to the coreboot mem console. It does this by passing a device tree from TF-A into OP-TEE that contains the firmware/coreboot node as describe in the Linux device tree specification. OP-TEE then parses the table to extract the address of the coreboot console.

A second commit adds support for logging to multiple serial consoles, which is useful when developing with the coreboot mem console driver enabled.

### DIFF
--- a/core/drivers/cbmem_console.c
+++ b/core/drivers/cbmem_console.c
@@ -1,0 +1,167 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2023, Linaro Limited
+ */
+
+#include <compiler.h>
+#include <console.h>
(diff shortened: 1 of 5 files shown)


Narflex:
Whoops...forgot to recompile after processing checkpatch corrections (and this doesn't compile now). Will submit a fixup commit shortly.

Narflex:
The force push was to fix the commit messages.

Narflex:
I'm abandoning this one since it appears now that the second commit breaks w/ ASLR enabled. I'll redo this so it only has the first commit for cbmem console.

PR #5989, opened 2023-05-02
Re-work CAAM Crypto configuration file + move CFG_WITH_SOFTWARE_PRNG definition

clementfaure:
Hello,

This PR combines the re-work of the CAAM crypto configuration file for i.MX/LS and the move of CFG_WITH_SOFTWARE_PRNG definition from mk/config.mk to mk/crypto.mk.
The latter allows the HW Crypto driver to overide the default CFG_WITH_SOFTWARE_PRNG value. Let me know what you think about this change.

Thanks!

### DIFF
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -517,6 +517,5 @@ ifneq (,$(filter y, $(CFG_MX6) $(CFG_MX7) $(CFG_MX7ULP)))
 CFG_IMX_CAAM ?= y
 endif
 
-CFG_WITH_SOFTWARE_PRNG ?= y
 endif
 
--- a/core/arch/arm/plat-ls/conf.mk
+++ b/core/arch/arm/plat-ls/conf.mk
@@ -124,7 +124,3 @@ CFG_CRYPTO_SIZE_OPTIMIZATION ?= n
 # NXP CAAM support is not enabled by default and can be enabled
 # on the command line
 CFG_NXP_CAAM ?= n
-
-ifneq ($(CFG_NXP_CAAM),y)
-$(call force,CFG_WITH_SOFTWARE_PRNG,y)
-endif
--- a/core/crypto.mk
+++ b/core/crypto.mk
@@ -73,6 +73,12 @@ CFG_CRYPTO_AES_GCM_FROM_CRYPTOLIB ?= n
 
 endif
 
+# PRNG configuration
+# If CFG_WITH_SOFTWARE_PRNG is enabled, crypto provider provided
+# software PRNG implementation is used.
+# Otherwise, you need to implement hw_get_random_bytes() for your platform
+CFG_WITH_SOFTWARE_PRNG ?= y
+
 ifeq ($(CFG_WITH_PAGER),y)
 ifneq ($(CFG_CRYPTO_SHA256),y)
 $(warning Warning: Enabling CFG_CRYPTO_SHA256 [required by CFG_WITH_PAGER])
--- a/core/drivers/crypto/caam/crypto.mk
+++ b/core/drivers/crypto/caam/crypto.mk
@@ -171,9 +171,8 @@ endif
 
 endif # !CFG_LS
 
-$(call force, CFG_NXP_CAAM_ACIPHER_DRV, $(call cryphw-one-enabled, RSA ECC DH DSA))
+$(call force, CFG_CRYPTO_DRV_ACIPHER , $(call cryphw-one-enabled, RSA ECC DH DSA))
 $(call force, CFG_CRYPTO_DRV_MAC, $(call cryphw-one-enabled, HMAC CMAC))
-CFG_CRYPTO_DRV_ACIPHER ?= $(CFG_NXP_CAAM_ACIPHER_DRV)
 
 # Disable SM2 as it is not supported by the CAAM driver
 ifeq ($(CFG_NXP_CAAM_ECC_DRV),y)
--- a/core/drivers/crypto/caam/crypto.mk
+++ b/core/drivers/crypto/caam/crypto.mk
@@ -1,5 +1,4 @@
 ifeq ($(CFG_NXP_CAAM),y)
-#
 # CAAM Debug: define 3x32 bits value (same bit used to debug a module)
 # CFG_DBG_CAAM_TRACE  Module print trace
 # CFG_DBG_CAAM_DESC   Module descriptor dump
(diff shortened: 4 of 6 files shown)


jenswi-linaro:
Please update the commit message in
"core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms"
to reflect the actual change.

Other than that:

etienne-lms (commented):
Commit "drivers: caam: remove CFG_NXP_CAAM_ACIPHER compilation flag":
Maybe state in commit message that `CFG_CRYPTO_DRV_ACIPHER` is used instead as it already does the job.


Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition"
I don't think this change is needed. _km/config.mk_ is already included last among config makefiles.

Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms":
If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?

clementfaure:
> 



> Commit "drivers: caam: remove CFG_NXP_CAAM_ACIPHER compilation flag": Maybe state in commit message that `CFG_CRYPTO_DRV_ACIPHER` is used instead as it already does the job.

Ok, I will do.
> 
> Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition" I don't think this change is needed. _km/config.mk_ is already included last among config makefiles.

That's needed in this case. The `$(call force, CFG_WITH_SOFTWARE_PRNG,n)` must come before the `CFG_WITH_SOFTWARE_PRNG ?= y`. I've added the compilation error in the commit message.
> 
> Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?

I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.

clementfaure:
> 



> Commit "drivers: caam: remove CFG_NXP_CAAM_ACIPHER compilation flag": Maybe state in commit message that `CFG_CRYPTO_DRV_ACIPHER` is used instead as it already does the job.

Ok, I will do.
> 
> Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition" I don't think this change is needed. _km/config.mk_ is already included last among config makefiles.

That's needed in this case. The `$(call force, CFG_WITH_SOFTWARE_PRNG,n)` must come before the `CFG_WITH_SOFTWARE_PRNG ?= y`. I've added the compilation error in the commit message.
> 
> Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?

I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.

clementfaure:
> 



> Commit "drivers: caam: remove CFG_NXP_CAAM_ACIPHER compilation flag": Maybe state in commit message that `CFG_CRYPTO_DRV_ACIPHER` is used instead as it already does the job.

Ok, I will do.
> 
> Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition" I don't think this change is needed. _km/config.mk_ is already included last among config makefiles.

That's needed in this case. The `$(call force, CFG_WITH_SOFTWARE_PRNG,n)` must come before the `CFG_WITH_SOFTWARE_PRNG ?= y`. I've added the compilation error in the commit message.
> 
> Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?

I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.

clementfaure:
> 



> Commit "drivers: caam: remove CFG_NXP_CAAM_ACIPHER compilation flag": Maybe state in commit message that `CFG_CRYPTO_DRV_ACIPHER` is used instead as it already does the job.

Ok, I will do.
> 
> Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition" I don't think this change is needed. _km/config.mk_ is already included last among config makefiles.

That's needed in this case. The `$(call force, CFG_WITH_SOFTWARE_PRNG,n)` must come before the `CFG_WITH_SOFTWARE_PRNG ?= y`. I've added the compilation error in the commit message.
> 
> Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?

I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.

clementfaure:
> 



> Commit "drivers: caam: remove CFG_NXP_CAAM_ACIPHER compilation flag": Maybe state in commit message that `CFG_CRYPTO_DRV_ACIPHER` is used instead as it already does the job.

Ok, I will do.
> 
> Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition" I don't think this change is needed. _km/config.mk_ is already included last among config makefiles.

That's needed in this case. The `$(call force, CFG_WITH_SOFTWARE_PRNG,n)` must come before the `CFG_WITH_SOFTWARE_PRNG ?= y`. I've added the compilation error in the commit message.
> 
> Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?

I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.

clementfaure:
Tags applied + rebase

etienne-lms:
>> Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?
>>
>I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.

When `CFG_NXP_CAAM=y` you must force `CFG_WITH_SOFTWARE_PRNG=y`, not rely on the default value. If one build this flavor with `CFG_WITH_SOFTWARE_PRNG=n` (e.g. `make PLATFORM=<your-platform-flavor> CFG_WITH_SOFTWARE_PRNG=n`), you won't catch the issue from the config perspective, unless you do as I suggested to.


>> Commit "core: move CFG_WITH_SOFTWARE_PRNG default definition" I don't think this change is needed. km/config.mk is already included last among config makefiles.
>>
> That's needed in this case. The `$(call force, CFG_WITH_SOFTWARE_PRNG,n)` must come before the `CFG_WITH_SOFTWARE_PRNG ?= y`. I've added the compilation error in the commit message.

My mistake. _mk/config.mk_ is currently included after inclusion of _core/crypto.mk_.
This is were our Makefile base management of config switches dependencies comes to its limitation. We really need to move of a real efficient configuration framework.

clementfaure:
> > > Commit "core: ls: remove CFG_WITH_SOFTWARE_PRNG default definition for LS platforms": If `CFG_NXP_CAAM=y` requires `CFG_WITH_SOFTWARE_PRNG=y`, prefer keep it explicitly, no?
> > 
> > 
> > I mean, CFG_WITH_SOFTWARE_PRNG has a default value. I don't really see the point of explicitly re-setting the flag to its default value.
> 
> When `CFG_NXP_CAAM=y` **it looks like** you must force `CFG_WITH_SOFTWARE_PRNG=y`, **you can**not rely on the default value. If one build**s** this flavor with `CFG_WITH_SOFTWARE_PRNG=n`, you won't catch the issue from the config perspective, unless you do as I suggested to: E.g. `make PLATFORM=<your-platform-flavor> CFG_WITH_SOFTWARE_PRNG=n`.
> 

Actually, when  `CFG_NXP_CAAM=y`, `CFG_WITH_SOFTWARE_PRNG` must be disabled

etienne-lms:
Ok, i messed up with the switches value but I think the problem remains. That said, if you think it is ok to build some ls platforms with `CFG_NXP_CAAM=n` and `CFG_WITH_SOFTWARE_PRNG=n` without getting into unexpected problems, fine with me.

jforissier:
@etienne-lms we are all aware of the limitations of the current config system but switching to a different one such as Kconfig would require a lot of work I think, hence why it has never been done, considering the current system is "good enough"...

jenswi-linaro:
> @etienne-lms we are all aware of the limitations of the current config system but switching to a different one such as Kconfig would require a lot of work I think, hence why it has never been done, considering the current system is "good enough"...

I'm not sure it's so hard to do the switch, as long as it's accepted that we may need to configure it slightly differently before building. However, Kconfig isn't a silver bullet, so some problems may remain or others surface.

PR #2702, opened 2018-12-13
core: configure for a single user mode architecture

etienne-lms:
This change introduces `CFG_USER_TA_TARGET` to allow a configuration to select a unique user TA execution architecture even if the platform supports several (Aarch32/Aarch64).

`CFG_USER_TA_TARGET` shall be set to either `ta_arm32` or `ta_arm64`.

This feature can be used in build environments that do not support several cross compilation toolchains when building a single target. Buildroot (at least up to 2018.11) is such of an environment. The environment will likely use an Aarch64 toolchain for the TEE/REE which toolchain may not be able to compile Aarch32 targets and build to fail with a error message like:
```shell
    CC      out/ta_arm32-lib/libmbedtls/mbedtls/library/aes.o
  aarch64-linux-gnu-gcc: error: unrecognized command line option ‘-mthumb’
```

This change adds a bit of documentation on this configuration, in file **mk/config.mk** and adds few info about the related `CFG_USER_TA_TARGET_<ta-name>`.

PR #5990, opened 2023-05-02
CAAM Black key support for asymmetric crypto

clementfaure:
# CAAM Black/Blob support in OPTEE

## CAAM Black key overview
The CAAM provides a key encapsulation mechanism called Black key. This is a protection of user keys against bus snooping attacks when the keys are being written or read from external memory (outside the SoC). The black key mechanism encapsulates/decapsulates keys on the fly in an encrypted data structure called a black key. Before copying the plain text key to memory, the CAAM automatically encrypts the key in the black key format using a key encryption key. This key encryption key is called JDKEK - Job ring descriptor key encryption key.

## CAAM Black key types
There are two different black key formats
 - AES-ECB : The key is encrypted using AES-ECB. The resulting black key will a multiple of 16 bytes. => AES-ECB(plain text key + zero padding)
 - AES-CCM : The key is encrypted using AES-CCM (slower than ECB). This format also includes a MAC to ensure integrity and authenticity. This black key is at least 12 bytes longer than the plain text key. Key modulo(8 bytes) + 6 bytes nonce + 6 bytes ICV => AES-CCM(plain text key + zero padding + nonce) + ICV

Please note that in the CAAM context, a plain text key is called a Red key.
The CAAM driver requires the user to specify the given key type in the job descriptor. The CAAM cannot tell if the key given is Red or Black-ECB/CCM.
Once encapsulated, a black key cannot be decapsulated to memory. The CAAM will prevent any Black key to plain text key in memory operation. A black key can only be decapsulated to CAAM key registers.

## JDKEK
Each job ring has its own key encryption key to encapsulate keys into black key format. This key is randomly re-generated upon each reset. The black key encapsulation is modified via the appropriate Trustzone status and SDID value. With this, each domain has its own black keys and cannot use black keys generated by other domains. When the CAAM is instructed to use a black key, the CAAM automatically decrypts the incoming black key in its own key register to do the crypto operations.
JDKEK has a limited lifetime, and it is re-generated (randomly) upon each power cycle (reset, suspend/resume). A black key generated in a power cycle will be unusable after the next power cycle.
Please note that trusted descriptor key encryption keys - TDKEK can also be used to encapsulate black keys, but this is not supported in this PR.

## Black blobs
To retain encrypted keys across power cycles, the CAAM is able to encrypt keys using a secure non-volatile key encryption key. This non-volatile KEK is device specific and cannot be extracted from the device. The non-volatile KEK is derived from the device master key. There are several methods to define the master key.
Black blobs provide confidentiality and integrity protection, and keys can be stored in a long-term storage. To make a black key power cycle proof, the CAAM generates a random blob key to encrypt the black key. This random blob key is then encrypted with the non-volatile Blob KEK.

A black key can be encapsulated into a black blob, and a black blob can be decapsulated to a black key (in memory) or plain text key (CAAM key registers).
```

                                                                BKEK
                                                                  │
                                                                  ▼
                                                            ┌────────────┐
                                                            │ AES-ECB    │
                     JDKEK or               RNG ───────────►│ Encryption │
                       TDKEK                 │              └─────┬──────┘
                         │                   │ Blob key           │
                         │                   │                    ▼
                         │                   │              ┌────────────┐
                         ▼                   ▼              │  Key blob  │
                  ┌─────────────┐      ┌────────────┐       ├────────────┤
┌───────────┐     │ AES-ECB/CCM │      │ AES-CCM    │       │            │
│ Black key ├────►│  Decryption ├─────►│ Encryption ├──────►│ Data blob  │
└───────────┘     └─────────────┘      └──────────┬─┘       │            │
                                                  │         │            │
                                                  │         ├────────────┤
                                                  └────────►│   MAC      │
                                                            └────────────┘
```

# CAAM Black key/blob in OPTEE

## Pull-request overview
This PR aims at adding the support of Black key/blob to CAAM asymmetric cryptographic operations. The key generation and the cryptographic operations being linked by the same API vectors, the support for asymmetric crypto does not require modification in the OPTEE core.
This PR also provides to the CAAM driver the ability to handle plain text keys and black keys. Indeed, the CAAM driver is able to use a previously generated black key or a plain text key given by the user as an input.
Please note that, in the case of asymmetric keys, only the private part of the key pair is encapsulated as a black key/blob.

## Black key supported formats
This PR adds the support for AES-ECB and AES-CCM black keys. Only one key generation format is supported at the same time. It must be chosen at compilation. It defines the default format for CAAM key generation.
There are 3 formats available:
 - Plain text
 - AES-ECB
 - AES-CCM
Even if the CAAM supports only one key generation format by default, the CAAM is still able to use any key format (plain text, AES-ECB/CCM ...) for its crypto operations.

I would advise sticking to AES-CCM. It provides better security, adds authenticity and performance regarding the black key encapsulation/decapsulation are similar to AES-ECB.

Because a cryptographic key pair is a TEE object and can be saved in persistent memory at any time, private key pairs will always be stored as a black blob (which is power cycle proof).
```
┌─────────────────────────┐
│                         │
│   CAAM Key generation   ├──────────────►   Key format defined by `caam_key_default_key_gen_type()`
│                         │
└─────────────────────────┘

┌─────────────────────────┐
│                         │
│  CAAM Crypto operations │◄──────────────   Plain text key, black key ECB, CCM, ...
│                         │
└─────────────────────────┘
```

## How to handle both Black keys and plain text keys?
Along the key data, a header with some metadata must be added to tell the CAAM how to handle the incoming key.
Metadata:
 - Magic number (32 bits): a magic number the CAAM driver can detect on the first 32 bits of the key buffer is needed. If a magic number is detected, the key is a black key generated by the CAAM. If not, this is a plain text key.
 - Key type (32 bits): specify the CAAM black key type. This info is needed by the CAAM for the black key decapsulation.
 - Key security size (32 bits): this is the security key size of the black key. This is different from the actual buffer key size. This info is needed by the CAAM for the black key decapsulation.

When generating a key pair, the CAAM driver encapsulates the generated private key in the following format `struct caam_key_serialized {}` :

```
Black blob key generated by the CAAM:
+-----------------------------------------------------+
| MAGIC number - 0xCAAFBFFB (32 bits)                 |
+-----------------------------------------------------+
| Key type (if magic number is present)               |
| -  CAAM_KEY_PLAIN_TEXT                              |
| -  CAAM_KEY_BLACK_ECB                               |
| -  CAAM_KEY_BLACK_CCM                               |
+-----------------------------------------------------+
| Plain text key size (if magic number is present)    |
+-----------------------------------------------------+
| Black blob                                          |
+-----------------------------------------------------+
```

```
Regular key generated by the CAAM:
+-----------------------------------------------------+
| Plain text data                                     |
+-----------------------------------------------------+
```

## Black blobs as a bignum in OPTEE core
Because a black key or blob is always bigger than the original plain text key size, the maximum bignum size must be increased. This is especially true for RSA-4096 key pair. `CFG_CORE_BIGNUM_MAX_BITS` value must be increased accordingly.

The private part of the key pair can be any size : the OPTEE core does not rely on the buffer size of the private key to determine the key pair size. The CAAM black key size overhead is not a problem for asymmetric crypto.

## Supported algos?
This PR adds the black key/blob support for the following algorithms:
 - DH
 - DSA
 - RSA
 - ECC
The CAAM directly outputs the private part of the key pair encapsulated in a black key format. The plain text private key is never exposed in the memory.
Note for RSA: currently, the RSA key pair generation CAAM function outputs the private part of the key pair in plain text format to memory. The plain text private key is manually encapsulated as a black key right after key pair generation. This will be solved in the future with the support of CAAM internal secure memory.
For the other algorithms, the key generation operations directly output the private key part in a black format.

## What about the support for symmetric algos?
The black key support for symmetric algos is another story and will most likely require changes in OPTEE Core. There are multiples obstacles:

 1. **The key buffer size is not equal to the key security size.**
 In the case of an encapsulated symmetric key, the actual key buffer size does not reflect the actual security key size. For the CAAM, a 256 bits key requires at least a 640 bits buffer to encapsulate the key as a black blob.
 An HW crypto driver could also refer to key IDs if it has its own keystore/internal memory (like the SE050).

 One way to work around this core/GP limitation would be to have a crypto HW driver callback that returns the actual security key size of the given encapsulated key. This way, the OPTEE core would ask the HW crypto driver for its key size checks ... If such callback is not implemented by the HW driver, the core would rely on the buffer size to determine the security key size.

 2.**The black key generation cannot be done through TEE Crypto API**
 There is no clean way to specify the crypto HW driver the type of key to generate. Going through `TEE_GenerateKey()` would require to add new `TEE_TYPE_...` like `TEE_TYPE_AES_BLACK_KEY` or something ... which is not very GP compliant.
 The only way to generate black keys would be a PTA.

Feel free to ask questions.

Regards,
Clement

### DIFF
--- a/core/drivers/crypto/caam/acipher/caam_dh.c
+++ b/core/drivers/crypto/caam/acipher/caam_dh.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright 2018-2021 NXP
+ * Copyright 2018-2021, 2023 NXP
  *
  * Implementation of DH
  */
@@ -8,6 +8,7 @@
 #include <caam_common.h>
 #include <caam_hal_ctrl.h>
 #include <caam_jr.h>
+#include <caam_key.h>
 #include <caam_utils_mem.h>
 #include <caam_utils_status.h>
 #include <drvcrypt.h>
@@ -30,7 +31,7 @@
 struct caam_dh_keypair {
 	struct caambuf g; /* Generator */
 	struct caambuf p; /* Prime Number Modulus */
-	struct caambuf x; /* Private key */
+	struct caamkey x; /* Private key */
 	struct caambuf y; /* Public key */
 };
 
@@ -43,7 +44,7 @@ static void do_keypair_free(struct caam_dh_keypair *key)
 {
 	caam_free_buf(&key->g);
 	caam_free_buf(&key->p);
-	caam_free_buf(&key->x);
+	caam_key_free(&key->x);
 	caam_free_buf(&key->y);
 }
 
@@ -114,12 +115,11 @@ static enum caam_status do_keypriv_conv(struct caam_dh_keypair *outkey,
 	cache_operation(TEE_CACHECLEAN, outkey->p.data, outkey->p.length);
 
 	/* Private Key X */
-	retstatus = caam_calloc_buf(&outkey->x, key_size);
+	retstatus = caam_key_deserialize_from_bn(inkey->x, &outkey->x, 0);
 	if (retstatus != CAAM_NO_ERROR)
 		return retstatus;
 
-	crypto_bignum_bn2bin(inkey->x, outkey->x.data);
-	cache_operation(TEE_CACHECLEAN, outkey->x.data, outkey->x.length);
+	caam_key_cache_op(TEE_CACHECLEAN, &outkey->x);
 
 	return CAAM_NO_ERROR;
 }
@@ -176,7 +176,7 @@ static TEE_Result do_allocate_keypair(struct dh_keypair *key, size_t size_bits)
 		goto err;
 
 	/* Allocate Private key X */
-	key->x = crypto_bignum_allocate(size_bits);
+	key->x = crypto_bignum_allocate(CFG_CORE_BIGNUM_MAX_BITS);
 	if (!key->x)
 		goto err;
 
@@ -226,6 +226,7 @@ static TEE_Result do_gen_keypair(struct dh_keypair *key,
 	uint32_t *desc = NULL;
 	uint32_t desclen = 0;
 	int counter = 0;
+	enum caam_key_type key_type = caam_key_default_key_gen_type();
 
 	l_bytes = crypto_bignum_num_bytes(key->p);
 	if (!l_bytes)
@@ -256,13 +257,17 @@ static TEE_Result do_gen_keypair(struct dh_keypair *key,
 	}
 
 	/* Allocate Private Key to be generated */
-	retstatus = caam_calloc_align_buf(&caam_dh_key.x, n_bytes);
+	caam_dh_key.x.key_type = key_type;
+	caam_dh_key.x.sec_size = n_bytes;
+	caam_dh_key.x.is_blob = false;
+
+	retstatus = caam_key_alloc(&caam_dh_key.x);
 	if (retstatus != CAAM_NO_ERROR) {
 		ret = caam_status_to_tee_result(retstatus);
 		goto out;
 	}
-	cache_operation(TEE_CACHEFLUSH, caam_dh_key.x.data,
-			caam_dh_key.x.length);
+
+	caam_key_cache_op(TEE_CACHEFLUSH, &caam_dh_key.x);
 
 	/* Allocate Public Key to be generated */
 	retstatus = caam_calloc_align_buf(&caam_dh_key.y, l_bytes);
@@ -301,9 +306,23 @@ static TEE_Result do_gen_keypair(struct dh_keypair *key,
 	caam_desc_add_ptr(desc, caam_dh_key.p.paddr);
 	caam_desc_add_ptr(desc, dh_r.paddr);
 	caam_desc_add_ptr(desc, caam_dh_key.g.paddr);
-	caam_desc_add_ptr(desc, caam_dh_key.x.paddr);
+	caam_desc_add_ptr(desc, caam_dh_key.x.buf.paddr);
 	caam_desc_add_ptr(desc, caam_dh_key.y.paddr);
-	caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL));
+
+	switch (key_type) {
+	case CAAM_KEY_PLAIN_TEXT:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, NONE));
+		break;
+	case CAAM_KEY_BLACK_ECB:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, ECB));
+		break;
+	case CAAM_KEY_BLACK_CCM:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, CCM));
+		break;
+	default:
+		ret = TEE_ERROR_GENERIC;
+		goto out;
+	}
 
 	desclen = caam_desc_get_len(desc);
 	caam_desc_update_hdr(desc, DESC_HEADER_IDX(desclen, desclen - 1));
@@ -321,15 +340,12 @@ static TEE_Result do_gen_keypair(struct dh_keypair *key,
 		retstatus = caam_jr_enqueue(&jobctx, NULL);
 
 		if (retstatus == CAAM_NO_ERROR) {
-			cache_operation(TEE_CACHEINVALIDATE, caam_dh_key.x.data,
-					caam_dh_key.x.length);
+			caam_key_cache_op(TEE_CACHEINVALIDATE, &caam_dh_key.x);
 			cache_operation(TEE_CACHEINVALIDATE, caam_dh_key.y.data,
 					caam_dh_key.y.length);
 
 			/* Copy Private and Public keypair */
-			ret = crypto_bignum_bin2bn(caam_dh_key.x.data,
-						   caam_dh_key.x.length,
-						   key->x);
+			ret = caam_key_serialize_to_bn(key->x, &caam_dh_key.x);
 			if (ret != TEE_SUCCESS)
 				goto out;
 
@@ -432,7 +448,7 @@ static TEE_Result do_shared_secret(struct drvcrypt_secret_data *sdata)
 	caam_desc_add_word(desc, DESC_HEADER(0));
 	caam_desc_add_word(desc, pdb_sgt_flags |
 				 PDB_DL_KEY_L_SIZE(caam_dh_key.y.length) |
-				 PDB_DL_KEY_N_SIZE(caam_dh_key.x.length));
+				 PDB_DL_KEY_N_SIZE(caam_dh_key.x.sec_size));
 	/* Prime */
 	caam_desc_add_ptr(desc, caam_dh_key.p.paddr);
 	/* Modulus - Not used */
@@ -440,11 +456,25 @@ static TEE_Result do_shared_secret(struct drvcrypt_secret_data *sdata)
 	/* Public key */
 	caam_desc_add_ptr(desc, caam_dh_key.y.paddr);
 	/* Private key */
-	caam_desc_add_ptr(desc, caam_dh_key.x.paddr);
+	caam_desc_add_ptr(desc, caam_dh_key.x.buf.paddr);
 	/* Output secret */
 	caam_desc_add_ptr(desc, secret.sgtbuf.paddr);
 
-	caam_desc_add_word(desc, SHARED_SECRET(DL));
+	switch (caam_dh_key.x.key_type) {
+	case CAAM_KEY_PLAIN_TEXT:
+		caam_desc_add_word(desc, SHARED_SECRET(DL, NONE));
+		break;
+	case CAAM_KEY_BLACK_ECB:
+		caam_desc_add_word(desc, SHARED_SECRET(DL, ECB));
+		break;
+	case CAAM_KEY_BLACK_CCM:
+		caam_desc_add_word(desc, SHARED_SECRET(DL, CCM));
+		break;
+	default:
+		ret = TEE_ERROR_GENERIC;
+		goto out;
+	}
+
 	desclen = caam_desc_get_len(desc);
 	caam_desc_update_hdr(desc, DESC_HEADER_IDX(desclen, desclen - 1));
 
--- a/core/drivers/crypto/caam/acipher/caam_dsa.c
+++ b/core/drivers/crypto/caam/acipher/caam_dsa.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright 2019-2021 NXP
+ * Copyright 2019-2021, 2023 NXP
  *
  * Implementation of DSA functions
  */
@@ -8,6 +8,7 @@
 #include <caam_common.h>
 #include <caam_hal_ctrl.h>
 #include <caam_jr.h>
+#include <caam_key.h>
 #include <caam_utils_mem.h>
 #include <caam_utils_status.h>
 #include <drvcrypt.h>
@@ -17,6 +18,7 @@
 #include <string.h>
 
 #include "local.h"
+#include "utee_types.h"
 
 #ifdef CFG_CAAM_64BIT
 #define MAX_DESC_KEY_GEN 14
@@ -38,7 +40,7 @@ struct caam_dsa_keypair {
 	struct caambuf g; /* Generator */
 	struct caambuf p; /* Prime Number (L bits) */
 	struct caambuf q; /* Subprime Number (N bits) */
-	struct caambuf x; /* Private key */
+	struct caamkey x; /* Private key */
 	struct caambuf y; /* Public key */
 };
 
@@ -52,7 +54,7 @@ static void do_keypair_free(struct caam_dsa_keypair *key)
 	caam_free_buf(&key->g);
 	caam_free_buf(&key->p);
 	caam_free_buf(&key->q);
-	caam_free_buf(&key->x);
+	caam_key_free(&key->x);
 	caam_free_buf(&key->y);
 }
 
@@ -184,19 +186,14 @@ static enum caam_status do_keypriv_conv(struct caam_dsa_keypair *outkey,
 	field_size = crypto_bignum_num_bytes(inkey->q);
 	crypto_bignum_bn2bin(inkey->q, outkey->q.data + n_bytes - field_size);
 
-	/* Private key is only scalar x of n bytes */
-	retstatus = caam_calloc_buf(&outkey->x, n_bytes);
+	retstatus = caam_key_deserialize_from_bn(inkey->x, &outkey->x, n_bytes);
 	if (retstatus != CAAM_NO_ERROR)
 		return retstatus;
-
-	/* Get the number of bytes of x to pad with 0's */
-	field_size = crypto_bignum_num_bytes(inkey->x);
-	crypto_bignum_bn2bin(inkey->x, outkey->x.data + n_bytes - field_size);
 
 	cache_operation(TEE_CACHECLEAN, outkey->g.data, outkey->g.length);
 	cache_operation(TEE_CACHECLEAN, outkey->p.data, outkey->p.length);
 	cache_operation(TEE_CACHECLEAN, outkey->q.data, outkey->q.length);
-	cache_operation(TEE_CACHECLEAN, outkey->x.data, outkey->x.length);
+	caam_key_cache_op(TEE_CACHECLEAN, &outkey->x);
 
 	return CAAM_NO_ERROR;
 }
@@ -295,7 +292,7 @@ static TEE_Result do_allocate_keypair(struct dsa_keypair *key, size_t l_bits,
 		goto err;
 
 	/* Allocate Private key X */
-	key->x = crypto_bignum_allocate(n_bits);
+	key->x = crypto_bignum_allocate(CFG_CORE_BIGNUM_MAX_BITS);
 	if (!key->x)
 		goto err;
 
@@ -385,6 +382,7 @@ static TEE_Result do_gen_keypair(struct dsa_keypair *key, size_t l_bits,
 	uint32_t desclen = 0;
 	size_t l_bytes = l_bits / 8;
 	size_t n_bytes = n_bits / 8;
+	enum caam_key_type key_type = caam_key_default_key_gen_type();
 
 	DSA_TRACE("Generate Key - Private (%zu bits) and Public (%zu bits)",
 		  n_bits, l_bits);
@@ -397,13 +395,17 @@ static TEE_Result do_gen_keypair(struct dsa_keypair *key, size_t l_bits,
 	}
 
 	/* Allocate Private Key to be generated */
-	retstatus = caam_calloc_align_buf(&caam_dsa_key.x, n_bytes);
+	caam_dsa_key.x.key_type = caam_key_default_key_gen_type();
+	caam_dsa_key.x.sec_size = n_bytes;
+	caam_dsa_key.x.is_blob = false;
+
+	retstatus = caam_key_alloc(&caam_dsa_key.x);
 	if (retstatus != CAAM_NO_ERROR) {
 		ret = caam_status_to_tee_result(retstatus);
 		goto out;
 	}
-	cache_operation(TEE_CACHEFLUSH, caam_dsa_key.x.data,
-			caam_dsa_key.x.length);
+
+	caam_key_cache_op(TEE_CACHEFLUSH, &caam_dsa_key.x);
 
 	/* Allocate Public Key to be generated */
 	retstatus = caam_calloc_align_buf(&caam_dsa_key.y, l_bytes);
@@ -430,9 +432,23 @@ static TEE_Result do_gen_keypair(struct dsa_keypair *key, size_t l_bits,
 	caam_desc_add_ptr(desc, caam_dsa_key.p.paddr);
 	caam_desc_add_ptr(desc, caam_dsa_key.q.paddr);
 	caam_desc_add_ptr(desc, caam_dsa_key.g.paddr);
-	caam_desc_add_ptr(desc, caam_dsa_key.x.paddr);
+	caam_desc_add_ptr(desc, caam_dsa_key.x.buf.paddr);
 	caam_desc_add_ptr(desc, caam_dsa_key.y.paddr);
-	caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL));
+
+	switch (key_type) {
+	case CAAM_KEY_PLAIN_TEXT:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, NONE));
+		break;
+	case CAAM_KEY_BLACK_ECB:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, ECB));
+		break;
+	case CAAM_KEY_BLACK_CCM:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, CCM));
+		break;
+	default:
+		ret = TEE_ERROR_GENERIC;
+		goto out;
+	}
 
 	desclen = caam_desc_get_len(desc);
 	caam_desc_update_hdr(desc, DESC_HEADER_IDX(desclen, desclen - 1));
@@ -443,14 +459,13 @@ static TEE_Result do_gen_keypair(struct dsa_keypair *key, size_t l_bits,
 	retstatus = caam_jr_enqueue(&jobctx, NULL);
 
 	if (retstatus == CAAM_NO_ERROR) {
-		cache_operation(TEE_CACHEINVALIDATE, caam_dsa_key.x.data,
-				caam_dsa_key.x.length);
+		caam_key_cache_op(TEE_CACHEINVALIDATE, &caam_dsa_key.x);
+
 		cache_operation(TEE_CACHEINVALIDATE, caam_dsa_key.y.data,
 				caam_dsa_key.y.length);
 
 		/* Copy Private and Public keypair */
-		ret = crypto_bignum_bin2bn(caam_dsa_key.x.data,
-					   caam_dsa_key.x.length, key->x);
+		ret = caam_key_serialize_to_bn(key->x, &caam_dsa_key.x);
 		if (ret != TEE_SUCCESS)
 			goto out;
 
@@ -459,7 +474,8 @@ static TEE_Result do_gen_keypair(struct dsa_keypair *key, size_t l_bits,
 		if (ret != TEE_SUCCESS)
 			goto out;
 
-		DSA_DUMPBUF("X", caam_dsa_key.x.data, caam_dsa_key.x.length);
+		DSA_DUMPBUF("X", caam_dsa_key.x.buf.data,
+			    caam_dsa_key.x.buf.length);
 		DSA_DUMPBUF("Y", caam_dsa_key.y.data, caam_dsa_key.y.length);
 	} else {
 		DSA_TRACE("CAAM Status 0x%08" PRIx32, jobctx.status);
@@ -569,7 +585,7 @@ static TEE_Result do_sign(struct drvcrypt_sign_data *sdata, size_t l_bytes,
 	/* Generator */
 	caam_desc_add_ptr(desc, dsakey.g.paddr);
 	/* Secret key */
-	caam_desc_add_ptr(desc, dsakey.x.paddr);
+	caam_desc_add_ptr(desc, dsakey.x.buf.paddr);
 	/* Input message */
 	caam_desc_add_ptr(desc, msg.sgtbuf.paddr);
 	/* Signature 1st part */
@@ -579,7 +595,20 @@ static TEE_Result do_sign(struct drvcrypt_sign_data *sdata, size_t l_bytes,
 	/* Message length */
 	caam_desc_add_word(desc, sdata->message.length);
 
-	caam_desc_add_word(desc, DSA_SIGN(DL, HASHED));
+	switch (dsakey.x.key_type) {
+	case CAAM_KEY_PLAIN_TEXT:
+		caam_desc_add_word(desc, DSA_SIGN(DL, HASHED, NONE));
+		break;
+	case CAAM_KEY_BLACK_ECB:
+		caam_desc_add_word(desc, DSA_SIGN(DL, HASHED, ECB));
+		break;
+	case CAAM_KEY_BLACK_CCM:
+		caam_desc_add_word(desc, DSA_SIGN(DL, HASHED, CCM));
+		break;
+	default:
+		ret = TEE_ERROR_GENERIC;
+		goto out;
+	}
 
 	desclen = caam_desc_get_len(desc);
 	caam_desc_update_hdr(desc, DESC_HEADER_IDX(desclen, desclen - 1));
--- a/core/drivers/crypto/caam/acipher/caam_ecc.c
+++ b/core/drivers/crypto/caam/acipher/caam_ecc.c
@@ -8,13 +8,17 @@
 #include <caam_common.h>
 #include <caam_hal_ctrl.h>
 #include <caam_jr.h>
+#include <caam_key.h>
+#include <caam_trace.h>
 #include <caam_utils_mem.h>
 #include <caam_utils_status.h>
 #include <drvcrypt.h>
 #include <drvcrypt_acipher.h>
 #include <mm/core_memprot.h>
+#include <stdint.h>
 #include <string.h>
 #include <tee/cache.h>
+#include <utee_types.h>
 
 #ifdef CFG_CAAM_64BIT
 #define MAX_DESC_KEY_GEN 8
@@ -35,7 +39,7 @@
  */
 struct caam_ecc_keypair {
 	struct caambuf xy;
-	struct caambuf d;
+	struct caamkey d;
 };
 
 /*
@@ -46,7 +50,7 @@ struct caam_ecc_keypair {
 static void do_keypair_free(struct caam_ecc_keypair *key)
 {
 	caam_free_buf(&key->xy);
-	caam_free_buf(&key->d);
+	caam_key_free(&key->d);
 }
 
 /*
@@ -96,20 +100,18 @@ static enum caam_status do_keypair_conv(struct caam_ecc_keypair *outkey,
 					size_t size_sec)
 {
 	enum caam_status retstatus = CAAM_OUT_MEMORY;
-	size_t d_size = 0;
 
 	ECC_TRACE("ECC Convert Keypair size %zu bytes", size_sec);
 
 	/* Private key is only scalar d of sec_size bytes */
-	retstatus = caam_calloc_buf(&outkey->d, size_sec);
-	if (retstatus != CAAM_NO_ERROR)
+	retstatus = caam_key_deserialize_from_bn(inkey->d, &outkey->d,
+						 size_sec);
+	if (retstatus)
 		return retstatus;
 
-	/* Get the number of bytes of d to pad with 0's */
-	d_size = crypto_bignum_num_bytes(inkey->d);
-	crypto_bignum_bn2bin(inkey->d, outkey->d.data + size_sec - d_size);
+	caam_key_cache_op(TEE_CACHEFLUSH, &outkey->d);
 
-	cache_operation(TEE_CACHECLEAN, outkey->d.data, outkey->d.length);
+	ECC_DUMPBUF("Outkey", outkey->d.buf.data, outkey->d.buf.length);
 
 	return CAAM_NO_ERROR;
 }
@@ -161,7 +163,7 @@ static TEE_Result do_allocate_keypair(struct ecc_keypair *key,
 	memset(key, 0, sizeof(*key));
 
 	/* Allocate Secure Scalar */
-	key->d = crypto_bignum_allocate(size_bits);
+	key->d = crypto_bignum_allocate(CFG_CORE_BIGNUM_MAX_BITS);
 	if (!key->d)
 		goto err;
 
@@ -252,11 +254,12 @@ static TEE_Result do_gen_keypair(struct ecc_keypair *key, size_t key_size)
 	TEE_Result ret = TEE_ERROR_GENERIC;
 	enum caam_status retstatus = CAAM_FAILURE;
 	enum caam_ecc_curve curve = CAAM_ECC_UNKNOWN;
-	struct caambuf d = { };
+	struct caamkey d = { };
 	struct caambuf xy = { };
 	struct caam_jobctx jobctx = { };
 	uint32_t *desc = NULL;
 	uint32_t desclen = 0;
+	enum caam_key_type key_type = caam_key_default_key_gen_type();
 
 	ECC_TRACE("Generate Keypair of %zu bits", key_size);
 
@@ -276,28 +279,47 @@ static TEE_Result do_gen_keypair(struct ecc_keypair *key, size_t key_size)
 	}
 
 	/*
-	 * Allocate secure and public keys in one buffer
+	 * Allocate secure and public keys in two buffers
 	 * Secure key size = key_size align in bytes
 	 * Public key size = (key_size * 2) align in bytes
 	 */
-	retstatus = caam_alloc_align_buf(&d, (key_size / 8) * 3);
+	d.key_type = key_type;
+	d.sec_size = ROUNDUP_DIV(key_size, 8);
+	d.is_blob = false;
+
+	retstatus = caam_key_alloc(&d);
 	if (retstatus != CAAM_NO_ERROR) {
 		ret = caam_status_to_tee_result(retstatus);
 		goto out;
 	}
 
-	/* Build the xy buffer to simplify the code */
-	xy.data = d.data + key_size / 8;
-	xy.length = 2 * (key_size / 8);
-	xy.paddr = d.paddr + key_size / 8;
+	retstatus = caam_alloc_align_buf(&xy, (key_size / 8) * 2);
+	if (retstatus != CAAM_NO_ERROR) {
+		ret = caam_status_to_tee_result(retstatus);
+		goto out;
+	}
 
 	/* Build the descriptor using Predifined ECC curve */
 	caam_desc_init(desc);
 	caam_desc_add_word(desc, DESC_HEADER(0));
 	caam_desc_add_word(desc, PDB_PKGEN_PD1 | PDB_ECC_ECDSEL(curve));
-	caam_desc_add_ptr(desc, d.paddr);
+	caam_desc_add_ptr(desc, d.buf.paddr);
 	caam_desc_add_ptr(desc, xy.paddr);
-	caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC));
+
+	switch (key_type) {
+	case CAAM_KEY_PLAIN_TEXT:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC, NONE));
+		break;
+	case CAAM_KEY_BLACK_ECB:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC, ECB));
+		break;
+	case CAAM_KEY_BLACK_CCM:
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC, CCM));
+		break;
+	default:
+		ret = TEE_ERROR_GENERIC;
+		goto out;
+	}
 
 	desclen = caam_desc_get_len(desc);
 	caam_desc_update_hdr(desc, DESC_HEADER_IDX(desclen, desclen - 1));
@@ -305,16 +327,21 @@ static TEE_Result do_gen_keypair(struct ecc_keypair *key, size_t key_size)
 	ECC_DUMPDESC(desc);
 
 	jobctx.desc = desc;
-	cache_operation(TEE_CACHEFLUSH, d.data, d.length);
+	caam_key_cache_op(TEE_CACHEFLUSH, &d);
+	cache_operation(TEE_CACHEFLUSH, xy.data, xy.length);
+
 	retstatus = caam_jr_enqueue(&jobctx, NULL);
 
 	if (retstatus == CAAM_NO_ERROR) {
-		cache_operation(TEE_CACHEINVALIDATE, d.data, d.length);
+		caam_key_cache_op(TEE_CACHEINVALIDATE, &d);
+		cache_operation(TEE_CACHEINVALIDATE, xy.data, xy.length);
 
 		/* Copy all keypair parameters */
-		ret = crypto_bignum_bin2bn(d.data, key_size / 8, key->d);
-		if (ret != TEE_SUCCESS)
+		retstatus = caam_key_serialize_to_bn(key->d, &d);
+		if (retstatus) {
+			ret = caam_status_to_tee_result(retstatus);
 			goto out;
+		}
 
 		ret = crypto_bignum_bin2bn(xy.data, xy.length / 2, key->x);
 		if (ret != TEE_SUCCESS)
@@ -325,7 +352,7 @@ static TEE_Result do_gen_keypair(struct ecc_keypair *key, size_t key_size)
 		if (ret != TEE_SUCCESS)
 			goto out;
 
-		ECC_DUMPBUF("D", d.data, key_size / 8);
+		ECC_DUMPBUF("D", d.buf.data, key_size / 8);
 		ECC_DUMPBUF("X", xy.data, xy.length / 2);
 		ECC_DUMPBUF("Y", xy.data + xy.length / 2, xy.length / 2);
 	} else {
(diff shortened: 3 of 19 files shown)


jenswi-linaro:
What are black keys?

sahilnxp:
> What are black keys?

@jenswi-linaro That is CAAM terminology for Runtime Secure Keys.

jenswi-linaro:
Yes, but how are they used? I need to understand what I'm reviewing. I'd rather not read a spec of 1000 pages first.

sahilnxp:
Sorry for that, It is in context of `[RFC] Runtime Secure Keys Framework Proposal` https://github.com/OP-TEE/optee_os/pull/4807

clementfaure:
Hi @jenswi-linaro , I will provide a description of the CAAM black key mechanism shortly

clementfaure:
Hello all, please have a look at PR description for more info about the black implementation.

jenswi-linaro (commented):
Comments for "drivers: caam: add CAAM key driver"

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+ * @size	Input buffer size
+ */
+static uint32_t get_magic_number(const uint8_t *data, size_t size)
+{
+	struct caam_key_serialized key = {};
+
+	assert(data && size);
+	assert(size > sizeof(key.header));
+
+	memcpy(&key, data, sizeof(key.header));
```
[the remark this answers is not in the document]
clementfaure: This corresponds to an older design, I will fix it indeed.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+ * @size	Input buffer size
+ */
+static size_t get_key_sec_size(const uint8_t *data, size_t size)
+{
+	struct caam_key_serialized key = {};
+
+	assert(data && size);
+	assert(size > sizeof(key.header));
+
+	if (get_magic_number(data, size) != MAGIC_NUMBER)
```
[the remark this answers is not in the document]
clementfaure: A plain text key, in this design, doesn't feature any header. It's the magic number that defines if a key is plain or encrypted.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+	case CAAM_KEY_PLAIN_TEXT:
+		/*
+		 * If the key is plain text, the allocation size is equal to the
+		 * key size and no blob operation on this key is possible.
+		 */
+		return key->sec_size;
+	case CAAM_KEY_BLACK_ECB:
+		/* ECB-black key must be a multiple of 16 bytes */
+		return ROUNDUP(key->sec_size, 16);
+		break;
```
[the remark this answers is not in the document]
clementfaure: Ok

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+		 */
+		return ROUNDUP(key->sec_size, 8) + BLACK_KEY_NONCE_SIZE +
+			     BLACK_KEY_ICV_SIZE;
+		break;
+	default:
+		return 0;
+	}
+}
+
+void caam_key_to_str(const char *trace , const struct caamkey *key)
```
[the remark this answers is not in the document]
clementfaure: Ok

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+{
+	if (!key || !trace)
+		return;
+
+	KEY_TRACE("%s key_type:%s key_size:%zu is_blob:%s addr:%p",
+		  caam_key_type_to_str[key->key_type], trace, key->sec_size,
```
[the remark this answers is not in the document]
clementfaure: Ok, I will do it

clementfaure:
I will wait for additional feedback/review before addressing comments. Same thing for checkpatch issues. By the way, I'm not sure about this check:
```
CHECK: Macro argument reuse 'alg' - possible side-effects?
#2702: FILE: core/drivers/crypto/caam/include/caam_desc_helper.h:515:
+#define SHARED_SECRET(type, alg)                                               \
 	(CMD_OP_TYPE | OP_TYPE(UNI) | PROTID(SHARED_SECRET) |                  \
+	 PROT_PK_TYPE(type) | PROT_PRI_ENC(alg) | PROT_PRI_EXT(alg))
```
Is it really a problem?

jforissier:
> I will wait for additional feedback/review before addressing comments. Same thing for checkpatch issues. By the way, I'm not sure about this check:
> 
> ```
> CHECK: Macro argument reuse 'alg' - possible side-effects?
> #2702: FILE: core/drivers/crypto/caam/include/caam_desc_helper.h:515:
> +#define SHARED_SECRET(type, alg)                                               \
>  	(CMD_OP_TYPE | OP_TYPE(UNI) | PROTID(SHARED_SECRET) |                  \
> +	 PROT_PK_TYPE(type) | PROT_PRI_ENC(alg) | PROT_PRI_EXT(alg))
> ```
> 
> Is it really a problem?

In this case I think you could easily use a temporary and the `({ .. })` syntax like here:

https://github.com/OP-TEE/optee_os/blob/23419643ecf2dcce31ccc2ee03dfa74096a45de5/core/include/signed_hdr.h#L183-L184

clementfaure:
> > I will wait for additional feedback/review before addressing comments. Same thing for checkpatch issues. By the way, I'm not sure about this check:
> > ```
> > CHECK: Macro argument reuse 'alg' - possible side-effects?
> > #2702: FILE: core/drivers/crypto/caam/include/caam_desc_helper.h:515:
> > +#define SHARED_SECRET(type, alg)                                               \
> >  	(CMD_OP_TYPE | OP_TYPE(UNI) | PROTID(SHARED_SECRET) |                  \
> > +	 PROT_PK_TYPE(type) | PROT_PRI_ENC(alg) | PROT_PRI_EXT(alg))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is it really a problem?
> 
> In this case I think you could easily use a temporary and the `({ .. })` syntax like here:
> 
> https://github.com/OP-TEE/optee_os/blob/23419643ecf2dcce31ccc2ee03dfa74096a45de5/core/include/signed_hdr.h#L183-L184

Unfortunately, it doesn't work
```

In file included from core/drivers/crypto/caam/acipher/../include/caam_common.h:11,
                 from core/drivers/crypto/caam/acipher/../include/caam_acipher.h:10,
                 from core/drivers/crypto/caam/acipher/caam_rsa.c:8:
core/drivers/crypto/caam/acipher/caam_rsa.c: In function ‘do_gen_keypair’:
core/drivers/crypto/caam/acipher/caam_rsa.c:668:69: error: ‘NONE’ undeclared (first use in this function)
  668 |                         caam_desc_add_word(desc, RSA_FINAL_KEY(ALL, NONE));
      |                                                                     ^~~~
core/drivers/crypto/caam/acipher/../include/caam_desc_helper.h:488:19: note: in definition of macro ‘RSA_FINAL_KEY’
  488 |         ({ typeof(alg) _alg = (alg);                                           \
      |                   ^~~
```
`alg` is not a variable but a macro extension `#define PROT_PRI_ENC(alg)	SHIFT_U32(PROT_PRI_ENC_##alg, 2)`

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
clementfaure: After adding `data_to_serialized_key()`, I must keep it that way

clementfaure:
Force-push:
* Address some comments. Some comments are still open to discussion.
* Rebase

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+ * }
+ */
+struct caam_key_serialized {
+	struct {
+		uint32_t magic_number; /* Magic number */
+		uint32_t key_type;     /* Black key type */
+		uint32_t sec_size;     /* The original plain text key size */
+	} header;
+	uint8_t key[];
+} __packed;
```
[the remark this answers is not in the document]
jenswi-linaro: The proper way to do that is to make a struct layout that is free from padding by nature and then add a `static_assert()` to make sure that there really isn't any unexpected padding. Dropping the inner struct should be sufficient.

That leads to more efficient machine code, better warnings, and also saves some time for maintainers since there's less chance that others copy the slightly flawed `__packed` pattern. :-)

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,721 @@
...
+static struct caam_key_serialized *data_to_serialized_key(const uint8_t *data,
+							  size_t size)
+{
+	struct caam_key_serialized *key = NULL;
+
+	assert(data && size);
+	assert(size > sizeof(key->header));
+	assert(IS_ALIGNED_WITH_TYPE(data, struct caam_key_serialized));
+
+	key = (struct caam_key_serialized *)data;
```
jenswi-linaro, line 101: `return (struct caam_key_serialized *)data;` ?

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,721 @@
...
+	*size = key->buf.length;
+
+	/*
+	 * For black keys, the serialized key includes the header and must be
+	 * in a blob format
+	 */
+	if (key->key_type != CAAM_KEY_PLAIN_TEXT) {
+		size_t alloc = 0;
+		struct caam_key_serialized key_serialized = { };
+
```
jenswi-linaro, line 535: Empty line isn't needed.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,739 @@
...
+ * @size	Input buffer size
+ */
+static enum caam_key_type get_key_type(const uint8_t *data, size_t size)
+{
+	struct caam_key_serialized key = {};
+
+	assert(data && size);
+	assert(size > sizeof(key.header));
+
+	if (get_magic_number(data, size) != MAGIC_NUMBER)
```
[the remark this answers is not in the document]
jenswi-linaro: Who generates the key? If it's only done by the CAAM then we could check for that and regenerate in the unlikely event. If it can be supplied externally then we have to assume that it might be crafted.

jenswi-linaro:
> `alg` is not a variable but a macro extension `#define PROT_PRI_ENC(alg) SHIFT_U32(PROT_PRI_ENC_##alg, 2)`

The drawback with this approach is that cscope and grep become a bit less efficient. How about spelling out the constants instead?

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
clementfaure: If I drop `__packed` I will get a cast warning
```

core/drivers/crypto/caam/caam_key.c: In function ‘data_to_serialized_key’:
core/drivers/crypto/caam/caam_key.c:99:15: warning: cast increases required alignment of target type [-Wcast-align]
   99 |         key = (struct caam_key_serialized *)data;
      |               ^
core/drivers/crypto/caam/caam_key.c: At top level:
cc1: note: unrecognized command-line option ‘-Wno-c2x-extensions’ may have been intended to silence earlier diagnostics
```
A static assert can make sure it won't happen, but I will still get the warning, right?

clementfaure:
> PROT_PRI_ENC

I agree 

> > `alg` is not a variable but a macro extension `#define PROT_PRI_ENC(alg) SHIFT_U32(PROT_PRI_ENC_##alg, 2)`
> 
> The drawback with this approach is that cscope and grep become a bit less efficient. How about spelling out the constants instead?

The CAAM already has an extensive use of macro extension when it comes to build job descriptor, so I didn't think it was a problem. What do you suggest? I would like to avoid adding `if else` conditions ..

core/drivers/crypto/caam/caam_key.c
clementfaure: Ok

core/drivers/crypto/caam/caam_key.c
clementfaure: ok

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
jenswi-linaro: If you allow externally supplied plain text private keys then you'd need a magic larger than a plain text key to be able to block forged keys.

As I understand it, it's not about probability like hash collisions, it seems trivial to forge a plain text private key with the magic number as a prefix.

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
jenswi-linaro: You can cast it to `void *` instead since you have checked that the address is well aligned.

jenswi-linaro:
> > > `alg` is not a variable but a macro extension `#define PROT_PRI_ENC(alg) SHIFT_U32(PROT_PRI_ENC_##alg, 2)`
> > 
> > 
> > The drawback with this approach is that cscope and grep become a bit less efficient. How about spelling out the constants instead?
> 
> The CAAM already has an extensive use of macro extension when it comes to build job descriptor, so I didn't think it was a problem. What do you suggest? I would like to avoid adding `if else` conditions ..

In this case, it's clearly a false positive so I suppose we could ignore it for now. Perhaps someone can find a way later to avoid such false positives.

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
clementfaure: > If you allow externally supplied plain text private keys then you'd need a magic larger than a plain text key to be able to block forged keys.

Yes, I want to maintain support for external plain text keys
> 
> As I understand it, it's not about probability like hash collisions, it seems trivial to forge a plain text private key with the magic number as a prefix.

Forging and trying to make a plain text key look like a black key, won't do much. If the key is forged to look like an ECB Black key, the singing operation will output a corrupted result. If the key is forged to look like a CCM Black key, the import key will fail (because the MAC verification) and no signing operation will be done.

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
clementfaure: I would personally go with either `__packed` or a `memcpy()`. I find it tricky to make a `void *` cast dependent of `assert()` in another line of code somewhere else.
I guess if we go that way, it must be properly documented in the code.

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
jenswi-linaro: OK, so it shouldn't be a security problem then.

core/drivers/crypto/caam/caam_key.c
jenswi-linaro: OK what?

clementfaure:
Up

clementfaure:
Could you please remove the stall label? Thanks

clementfaure:
@embetrix Enable the CAAM driver and all private keys generated by the CAAM will be encrypted.

clementfaure:
@embetrix Have you tested on your side? Any feedback?

clementfaure:
@jforissier @jenswi-linaro I know that's a big chunk of code to review, but if you could spend some time on it, that would be great. Feel free to ask questions if this PR remains unclear to you. Thanks a lot.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,721 @@
...
+#include <stdint.h>
+#include <string.h>
+#include <tee/cache.h>
+#include <tee_api_defines.h>
+#include <trace.h>
+#include <utee_types.h>
+
+/*
+ * CAAM Key magic number.
+ * This 32 bits value added to the first 32 bits tells the CAAM driver if
```
jforissier, line 23: "Added" sounds a bit strange here, how about "When the first 32 bits of a key buffer are equal to this value, the buffer is a serialized CAAM key structure"?

More fundamentally, what happens if a "plain" key happens to have MAGIC_NUMBER in its first 32-bits?

core/drivers/crypto/caam/crypto.mk
```diff
@@ -132,6 +132,12 @@ CFG_NXP_CAAM_RSA_KEY_FORMAT ?= 3
 # Disable device tree status of the secure job ring
 CFG_CAAM_JR_DISABLE_NODE ?= y
 
+# Define the default CAAM private key encryption generation and the bignum
+# maximum size needed.
+# CAAM_KEY_PLAIN_TEXT    -> 4096 bits
+# CAAM_KEY_BLACK_ECB|CCM -> 4156 bits
+CFG_CORE_BIGNUM_MAX_BITS = 4156
```
jforissier, line 139: `?=`

Are 4096 and 4156 the only valid values? If so it may be good to check it with `ifneq($(CFG_CORE_BIGNUM_MAX_BITS),...) ... $(error ...) ... endif`

core/drivers/crypto/caam/include/caam_desc_helper.h
```diff
@@ -387,6 +387,13 @@ static inline void dump_desc(uint32_t *desc)
 	(CMD_KEY_TYPE | CMD_CLASS(CLASS_2) | KEY_DEST(MDHA_SPLIT) |            \
 	 KEY_LENGTH(len))
 
+/*
+ * Load a class cla key of length len to register dst.
+ */
+#define LD_KEY(cla, dst, len)                                                  \
+	(CMD_KEY_TYPE | CMD_CLASS(cla) | KEY_DEST(dst) |                       \
+	 KEY_LENGTH(len))
```
jforissier, line 395: Fits on the previous line I think

core/drivers/crypto/caam/include/caam_key.h
```diff
@@ -0,0 +1,155 @@
...
+	enum caam_key_type key_type; /* CAAM Key type */
+	size_t sec_size; /* Security key size */
+	bool is_blob; /* Shows if the key is in blob format */
+};
+
+/*
+ * Returns the default key type for CAAM key generation.
+ * The CAAM can only generate one key type.
+ */
+inline enum caam_key_type caam_key_default_key_gen_type(void)
```
jforissier, line 36: `static inline`

core/drivers/crypto/caam/include/caam_trace.h
```diff
@@ -279,6 +280,29 @@
...
 
+/* KEY */
+#if CAAM_DBG_TRACE(KEY)
+#define KEY_TRACE DRV_TRACE
+#if CAAM_DBG_DESC(KEY)
+#define KEY_DUMPDESC(desc)                                                     \
+	do {                                                                   \
+		KEY_TRACE("KEY Descriptor");                                   \
+		DRV_DUMPDESC(desc);                                            \
+	} while (0)
```
jforissier, line 291: Preferably move the line continuation symbols `\` to the left (align on the next tab stop after the "KEY Descriptor" line)

jforissier (commented):
A few comments on "drivers: caam: add CAAM key driver"

jforissier:
For commits:
* drivers: caam: add class field to FIFO_ST macro
* drivers: caam: add missing header
* drivers: caam: fix DSA_DUMPDESC macro

core/drivers/crypto/caam/acipher/caam_rsa.c
```diff
@@ -539,18 +546,32 @@ static TEE_Result do_gen_keypair(struct rsa_keypair *key, size_t key_size)
...
-	retstatus = caam_calloc_align_buf(&genkey.p, key_size / 8);
+	size_d_gen_val_ptr = caam_calloc_align(sizeof(uint32_t));
+	if (!size_d_gen_val_ptr) {
+		ret = TEE_ERROR_OUT_OF_MEMORY;
+		goto exit_gen_keypair;
+	}
+
+	/* First allocate primes p and q */
+	genkey.p.key_type = CAAM_KEY_PLAIN_TEXT;
+	genkey.p.sec_size = (key_size / 8) / 2;
```
jforissier, line 557: It might be good to have a macro for that, so the `/ 2` could be documented.

jforissier (commented):
Commenting on "drivers: caam: add CAAM key support for RSA"

core/drivers/crypto/caam/caam_key.c
clementfaure: I will rephrase it. Regarding your question, let me know if this previous discussion answers your question : https://github.com/OP-TEE/optee_os/pull/5990#discussion_r1189978494

core/drivers/crypto/caam/crypto.mk
clementfaure: 4096 and 4156 are the minimum values required for 4096 RSA keypairs and black keypairs. You could still be able to 
decrease these values if you don't want to generate 4096 keypairs.

Here I check if `CFG_CORE_BIGNUM_MAX_BITS` can store a 4096 bits RSA black keypair
https://github.com/OP-TEE/optee_os/pull/5990/files#diff-2af92fa52b0087bf848ecd8b9b9de69c7330aa1b63cb31c3388387e1dcdb5bd7R723-R727

clementfaure: 4096 and 4156 are the minumum value s

Here I check if `CFG_CORE_BIGNUM_MAX_BITS` can store a 4096 bits RSA black keypair
https://github.com/OP-TEE/optee_os/pull/5990/files#diff-2af92fa52b0087bf848ecd8b9b9de69c7330aa1b63cb31c3388387e1dcdb5bd7R723-R727

clementfaure: I have to set it like this `=` instead of `?=` otherwise it won't override the default value.

core/drivers/crypto/caam/include/caam_key.h
clementfaure: Yup

core/drivers/crypto/caam/include/caam_trace.h
clementfaure: I wanted to align with the style in the file but okay, I will fix

core/drivers/crypto/caam/acipher/caam_prime_rsa.c
```diff
@@ -373,6 +373,9 @@ static void do_desc_setup(uint32_t *desc, struct prime_data_rsa *data,
 			  const struct caambuf *small_prime,
 			  const paddr_t desc_prime)
 {
+	size_t key_size_bits = data->key_size / 2;
```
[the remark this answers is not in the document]
clementfaure: That's my mistake. It does not make sense to divide it by 2 here. I will divide it by 2 here instead: 
`caam_desc_add_word(desc, 5 * data->key_size/ 2);`

core/drivers/crypto/caam/acipher/caam_rsa.c
clementfaure: p, q, dp, dq, qp are half the size of the keypair size (defined by n, d). I will define a macro if you want.

core/drivers/crypto/caam/acipher/caam_prime_rsa.c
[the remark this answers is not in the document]
clementfaure: It does not make sense to divide the key size here indeed. I will divide it in the function input argument.

core/drivers/crypto/caam/crypto.mk
jforissier: But that's not allowed (see [here](https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#there-are-only-two-ways) and a in the paragraphs above for details). It's either `?=` or `$(call force,CFG_CORE_BIGNUM_MAX_BITS,4156)`. In this case, if a default value is set somewhere else, you should try to set it earlier so that it becomes the actual default.

clementfaure:
Force-push: rebase and tags applied

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
jenswi-linaro: Feel free to add a comment.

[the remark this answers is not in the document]
clementfaure: On?

core/drivers/crypto/caam/caam_key.c
jforissier: > I will rephrase it. Regarding your question, let me know if this previous discussion answers your question : [#5990 (comment)](https://github.com/OP-TEE/optee_os/pull/5990#discussion_r1189978494)

It does, thanks ;) would you mind adding a comment to explain that collisions are not expected to be a problem, both for plain key generation, and regarding potential attacks (you comment [here](https://github.com/OP-TEE/optee_os/pull/5990#discussion_r1219291868) is especially useful IMO).

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
jenswi-linaro: > I guess if we go that way, it must be properly documented in the code.

clementfaure:
Hi, can we re-open the PR?

jforissier:
> Hi, can we re-open the PR?

Sure.

core/drivers/crypto/caam/crypto.mk
clementfaure: Fix suggestion here : ba8ef3776

core/drivers/crypto/caam/caam_key.c
clementfaure: Done

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
clementfaure: https://github.com/OP-TEE/optee_os/pull/5990/files#diff-2af92fa52b0087bf848ecd8b9b9de69c7330aa1b63cb31c3388387e1dcdb5bd7R109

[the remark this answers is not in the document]
jenswi-linaro: I'm sorry, I don't understand, what are you proposing?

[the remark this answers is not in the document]
clementfaure: Oops sorry, wrong link. I mean, I went with the `(void *)` cast and add comments : https://github.com/OP-TEE/optee_os/blob/21b78605d29a8a1d62f4cc483607ae429ccf7e94/core/drivers/crypto/caam/caam_key.c#L109

clementfaure:
Any more comments on this pull-request?

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,746 @@
...
+			.key_type = key->key_type,
+			.sec_size = key->sec_size,
+			.is_blob = true,
+		};
+
+		alloc = caam_key_get_alloc_size(&tmp);
+		if (!alloc)
+			return CAAM_FAILURE;
+
+		*size = alloc + sizeof(key_serialized);
```
jenswi-linaro, line 570: `sizeof(struct caam_key_serialized)`, and remove the otherwise unused `key_serialized`.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,746 @@
...
+
+	caam_desc_init(desc);
+	caam_desc_add_word(desc, DESC_HEADER(0));
+	caam_desc_add_word(desc, LD_KEY(CLASS_1, PKHA_E, key->sec_size));
+	caam_desc_add_ptr(desc, input_buf.paddr);
+
+	if (key_type == CAAM_KEY_BLACK_ECB)
+		caam_desc_add_word(desc, FIFO_ST(CLASS_NO, PKHA_E_AES_ECB_JKEK,
+						 key->sec_size));
+	else if (key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 692: Add nothing if `key_type` isn't `CAAM_KEY_BLACK_CCM` either?

jenswi-linaro (commented):
Comments for "drivers: caam: add CAAM key driver"

There are a few checkpatch complaints that aren't false positives.

core/drivers/crypto/caam/acipher/caam_rsa.c
```diff
@@ -618,29 +649,39 @@ static TEE_Result do_gen_keypair(struct rsa_keypair *key, size_t key_size)
...
-
+		caam_desc_add_ptr(desc, genkey.dp.buf.paddr);
+		caam_desc_add_ptr(desc, genkey.dq.buf.paddr);
+		caam_desc_add_ptr(desc, genkey.qp.buf.paddr);
+
+		if (key_type == CAAM_KEY_PLAIN_TEXT)
+			caam_desc_add_word(desc, RSA_FINAL_KEY(ALL, NONE));
+		else if (key_type == CAAM_KEY_BLACK_ECB)
+			caam_desc_add_word(desc, RSA_FINAL_KEY(ALL, ECB));
+		else if (key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 672: Add no word, if `key_type` isn't `CAAM_KEY_BLACK_CCM` either?

Same below.

core/drivers/crypto/caam/acipher/caam_rsa.c
```diff
@@ -105,17 +112,13 @@ static void do_keypair_free(struct caam_rsa_keypair *key)
...
-		caam_free_buf(&key->p);
-	}
-
-	if (key->format > RSA_PRIVATE_KEY_FORMAT_2 && key->dp.data) {
-		key->dp.length += key->dq.length + key->qp.length;
-		caam_free_buf(&key->dp);
-	}
+	caam_key_free(&key->d);
+	caam_key_free(&key->p);
+	caam_key_free(&key->q);
```
jenswi-linaro, line 118: Did we leak memory here before?

core/drivers/crypto/caam/acipher/caam_rsa.c
```diff
@@ -703,19 +757,19 @@ static TEE_Result do_oaep_decoding(struct drvcrypt_rsa_ed *rsa_data)
...
 	TEE_Result ret = TEE_ERROR_GENERIC;
 	enum caam_status retstatus = CAAM_FAILURE;
-	struct caambuf DB = { };
-	struct caambuf lHash = { };
-	struct caambuf seed = { };
-	struct caambuf dbMask = { };
-	struct caambuf maskedDB = { };
-	struct caambuf maskedSeed = { };
-	struct caambuf EM = { };
+	struct caambuf DB = {};
```
jenswi-linaro, line 760: Why are you changing this?

core/drivers/crypto/caam/acipher/caam_rsa.c
```diff
@@ -1413,7 +1512,24 @@ static TEE_Result do_caam_decrypt(struct drvcrypt_rsa_ed *rsa_data,
...
+		goto exit_decrypt;
+	}
+
+	/* Set private key encryption type */
+	if (g_key_type == CAAM_KEY_BLACK_ECB)
+		operation |= PROT_RSA_KEY_ENC(ECB);
+	else if (g_key_type == CAAM_KEY_BLACK_CCM)
+		operation |= PROT_RSA_KEY_ENC(CCM);
+	else if (g_key_type == CAAM_KEY_PLAIN_TEXT)
+		operation |= PROT_RSA_KEY_ENC(NONE);
```
jenswi-linaro, line 1530: Do nothing if `g_key_type` isn't `CAAM_KEY_PLAIN_TEXT` either?

jenswi-linaro (commented):
Comments for "drivers: caam: add CAAM key support for RSA"

core/drivers/crypto/caam/acipher/caam_dsa.c
```diff
@@ -430,9 +435,15 @@ static TEE_Result do_gen_keypair(struct dsa_keypair *key, size_t l_bits,
...
-	caam_desc_add_ptr(desc, caam_dsa_key.x.paddr);
+	caam_desc_add_ptr(desc, caam_dsa_key.x.buf.paddr);
 	caam_desc_add_ptr(desc, caam_dsa_key.y.paddr);
-	caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL));
+
+	if (key_type == CAAM_KEY_PLAIN_TEXT)
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, NONE));
+	else if (key_type == CAAM_KEY_BLACK_ECB)
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, ECB));
+	else if (key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 445: What if `key_type` isn't `CAAM_KEY_BLACK_CCM` either?

jenswi-linaro (commented):
Comment for "drivers: caam: add CAAM key support for DSA"

core/drivers/crypto/caam/acipher/caam_ecc.c
```diff
@@ -258,45 +263,61 @@ static TEE_Result do_gen_keypair(struct ecc_keypair *key, size_t key_size)
...
-	caam_desc_add_ptr(desc, d.paddr);
+	caam_desc_add_ptr(desc, d.buf.paddr);
 	caam_desc_add_ptr(desc, xy.paddr);
-	caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC));
+
+	if (key_type == CAAM_KEY_PLAIN_TEXT)
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC, NONE));
+	else if (key_type == CAAM_KEY_BLACK_ECB)
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(ECC, ECB));
+	else if (key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 297: What if `key_type` isn't `CAAM_KEY_BLACK_CCM` either?

core/drivers/crypto/caam/acipher/caam_ecc.c
```diff
@@ -660,11 +694,17 @@ static TEE_Result do_shared_secret(struct drvcrypt_secret_data *sdata)
...
+	caam_desc_add_ptr(desc, ecckey.d.buf.paddr);
 	/* Output secret */
 	caam_desc_add_ptr(desc, secret.sgtbuf.paddr);
 
-	caam_desc_add_word(desc, SHARED_SECRET(ECC));
+	if (ecckey.d.key_type == CAAM_KEY_PLAIN_TEXT)
+		caam_desc_add_word(desc, SHARED_SECRET(ECC, NONE));
+	else if (ecckey.d.key_type == CAAM_KEY_BLACK_ECB)
+		caam_desc_add_word(desc, SHARED_SECRET(ECC, ECB));
+	else if (ecckey.d.key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 705: What if `ecckey.d.key_type` isn't `CAAM_KEY_BLACK_CCM` either.

jenswi-linaro (commented):
Comments for "drivers: caam: add CAAM key support for ECC"

core/drivers/crypto/caam/acipher/caam_dh.c
```diff
@@ -226,6 +226,14 @@ static TEE_Result do_gen_keypair(struct dh_keypair *key,
 	uint32_t *desc = NULL;
 	uint32_t desclen = 0;
 	int counter = 0;
+	/*
+	 * Set DH CAAM key type to plain text as regression_4007_dh does not
+	 * support black keys.
+	 */
+	enum caam_key_type key_type = CAAM_KEY_PLAIN_TEXT;
+
+	if (key_type == CAAM_KEY_MAX_VALUE)
```
jenswi-linaro, line 235: Why?

core/drivers/crypto/caam/acipher/caam_dh.c
```diff
@@ -301,9 +313,15 @@ static TEE_Result do_gen_keypair(struct dh_keypair *key,
...
-	caam_desc_add_ptr(desc, caam_dh_key.x.paddr);
+	caam_desc_add_ptr(desc, caam_dh_key.x.buf.paddr);
 	caam_desc_add_ptr(desc, caam_dh_key.y.paddr);
-	caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL));
+
+	if (key_type == CAAM_KEY_PLAIN_TEXT)
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, NONE));
+	else if (key_type == CAAM_KEY_BLACK_ECB)
+		caam_desc_add_word(desc, PK_KEYPAIR_GEN(DL, ECB));
+	else if (key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 323: What if `key_type` isn't `CAAM_KEY_BLACK_CCM` either?

core/drivers/crypto/caam/acipher/caam_dh.c
```diff
@@ -432,19 +447,25 @@ static TEE_Result do_shared_secret(struct drvcrypt_secret_data *sdata)
...
+	caam_desc_add_ptr(desc, caam_dh_key.x.buf.paddr);
 	/* Output secret */
 	caam_desc_add_ptr(desc, secret.sgtbuf.paddr);
 
-	caam_desc_add_word(desc, SHARED_SECRET(DL));
+	if (caam_dh_key.x.key_type == CAAM_KEY_PLAIN_TEXT)
+		caam_desc_add_word(desc, SHARED_SECRET(DL, NONE));
+	else if (caam_dh_key.x.key_type == CAAM_KEY_BLACK_ECB)
+		caam_desc_add_word(desc, SHARED_SECRET(DL, ECB));
+	else if (caam_dh_key.x.key_type == CAAM_KEY_BLACK_CCM)
```
jenswi-linaro, line 466: What if `caam_dh_key.x` isn't `CAAM_KEY_BLACK_CCM` either?

jenswi-linaro (commented):
Comments for "drivers: caam: add CAAM key support for DH"

jenswi-linaro (commented):
For commit "core: move CFG_CORE_BIGNUM_MAX_BITS default definition"

This commit should probably be first in the pull request.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,746 @@
...
+
+	caam_free_buf(&key->buf);
+}
+
+void caam_key_cache_op(enum utee_cache_operation op, const struct caamkey *key)
+{
+	if (!key)
+		return;
+
+	if (key->buf.nocache == 0)
```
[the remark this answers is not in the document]
clementfaure: Ok

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,746 @@
...
+	assert(data && size && key);
+
+	KEY_TRACE("Deserialization binary buffer");
+	KEY_DUMPBUF("Deserialize key buffer input", data, size);
+
+	/*
+	 * If a security key size is given, use it. Otherwise, rely on
+	 * the buffer size.
+	 */
+	if (sec_size == 0)
```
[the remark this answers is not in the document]
clementfaure: This is needed because some keys will have leading zeros we must preserve. There is a comment few lines below. Let me know if you want me to move it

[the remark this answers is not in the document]
clementfaure: I'm gonna check it

core/drivers/crypto/caam/caam_key.c
clementfaure: Indeed

core/drivers/crypto/caam/caam_key.c
clementfaure: If the key isn't black, no encapsulation is needed. There is an `assert()` at the beginning of the `caam_key_black_encapsulation()` function.

jenswi-linaro: Please skip the "if (key_type == CAAM_KEY_BLACK_CCM)" part and only use an `else`.

If the answer is the same for all the other similar comments I've made, do the same there.

core/drivers/crypto/caam/acipher/caam_rsa.c
clementfaure: `key_type` have three valid values. This `if .. else if` covers all valid values.

core/drivers/crypto/caam/acipher/caam_rsa.c
clementfaure: I make sure `g_key_type` has a valid value above.

core/drivers/crypto/caam/acipher/caam_dsa.c
clementfaure: Same, I check if `key_type` has a valid value.

core/drivers/crypto/caam/include/caam_desc_helper.h
```diff
@@ -234,8 +234,9 @@ static inline void dump_desc(uint32_t *desc)
 /*
  * FIFO Store from register src of length len
  */
-#define FIFO_ST(src, len)                                                      \
-	(CMD_FIFO_STORE_TYPE | FIFO_STORE_OUTPUT(src) | FIFO_STORE_LENGTH(len))
+#define FIFO_ST(cla, src, len)                                                 \
+	(CMD_FIFO_STORE_TYPE | CMD_CLASS(cla) | FIFO_STORE_OUTPUT(src) |       \
+	FIFO_STORE_LENGTH(len))
```
etienne-lms, line 239: mind indentation (for consistency with other macro definitions)

etienne-lms (commented):
"drivers: caam: add class field to FIFO_ST macro" (with comment addressed),
"drivers: caam: add missing header" (add `Fixes: 38923d487567 ("drivers: caam: implement CAAM DMA Object")` in commit message),
"drivers: caam: fix DSA_DUMPDESC macro" (add `Fixes: 4b383f736e9e ("drivers: caam: implement NXP CAAM Driver - DSA"` is commit message)

"core: move CFG_CORE_BIGNUM_MAX_BITS default definition".

core/drivers/crypto/caam/acipher/caam_dh.c
clementfaure: Ok so, currently xtest `xtest_tee_test_4007_dh()` does not support CAAM black keys. During DH keypair generation, xtest checks if the returned buffer size is equal to the DH key size. A CAAM black key being bigger than the key itself, one of the check fails.
For xtest compatibility, DH generation is forced to plain text.
Let me know what you think

core/drivers/crypto/caam/include/caam_key.h
```diff
@@ -0,0 +1,155 @@
...
+enum caam_status caam_key_operation_blob(const struct caamkey *in_key,
+					 struct caamkey *out_key);
+
+/*
+ * Deserialize CAAM key structure from binary buffer
+ *
+ * @data	Buffer input
+ * @size	Buffer input size
+ * @key		CAAM key structure to populate
+ * @size_sec	Security key size to deserialize, optional. If not needed,
```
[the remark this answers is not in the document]
clementfaure: Fixed

core/drivers/crypto/caam/include/caam_desc_helper.h
clementfaure: Fixed

core/drivers/crypto/caam/caam_key.c
clementfaure: Ok for this one.

jenswi-linaro: When you're testing for as many different values as in the last case I'd prefer a switch-case construct, with a default for error handling.

At this location, I'd prefer just
``` C
if (key_type == CAAM_KEY_BLACK_ECB)
         ...
else
         ...
```
Please try to avoid testing for the same thing twice in the same function since it then in the second test "can't happen" if it was an error condition.

clementfaure: Ok, I will check other places where `key_type` is tested

core/drivers/crypto/caam/acipher/caam_rsa.c
clementfaure: Sorry, yes, I will remove
```
else if (g_key_type == CAAM_KEY_PLAIN_TEXT)
		operation |= PROT_RSA_KEY_ENC(NONE);
```

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
clementfaure: We cannot rely on the bignum size as it could be smaller than the security key size. In this case, we need to specify the security key to properly do the bignum to bin translation and eventually pad the result. It was already done before :
https://github.com/OP-TEE/optee_os/blob/8e9d8acca4f8141b72d2d32846317954124c17ec/core/drivers/crypto/caam/acipher/caam_ecc.c#L108-L110

clementfaure:
Force-push: rebase
Push: fixup commits:
- @jenswi-linaro I tried to follow your recommandations. I went with `switch` instead of `if else` and remove duplicated testing.
- @jforissier I still get checkpatch warnings about macros `Macro argument reuse 'alg' - possible side-effects?` Should I go with GNU extensions?

clementfaure:
Force-push : checkpatch warnings fixed and macros reworked

clementfaure:
Any comments?

core/drivers/crypto/caam/caam_key.c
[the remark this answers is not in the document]
jenswi-linaro: Please leave an explanation like this in the comment above for future reference.

jenswi-linaro (commented):
One comment for now. Please squash in the fixups, it's too long ago for me to remember what I looked at so I'll need to go through all the (unreviewed) patches again anyway.

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,753 @@
...
+		/*
+		 * For a blob operation, the input key size is the original key
+		 * size of the black key.
+		 * The output key size is the final blob size.
+		 */
+		input_buffer_size = in_key->sec_size;
+		output_buffer_size = out_key->buf.length;
+	} else {
+		/*
+		 * For an un-blob operation, the input key size is the original
```
jenswi-linaro, line 354: s/un-blob/non-blob/

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,753 @@
...
+	blob.sec_size = key->sec_size;
+	blob.key_type = key->key_type;
+
+	/*
+	 * Check if the destination is big enough for the black blob buffer and
+	 * header.
+	 */
+	status = caam_key_serialized_size(&blob, &serialized_size);
+	if (status)
+		goto out;
```
jenswi-linaro, line 515: Nothing is allocated yet, why not just return the error?

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,753 @@
...
+
+	if (size < serialized_size) {
+		KEY_TRACE("Destination buffer is too short %zu < %zu", size,
+			  serialized_size);
+		status = CAAM_OUT_MEMORY;
+		goto out;
```
jenswi-linaro, line 521: Nothing is allocated yet, why not just return the error?

core/drivers/crypto/caam/caam_key.c
```diff
@@ -0,0 +1,753 @@
...
+	}
+
+	/* Blob the given key */
+	status = caam_key_operation_blob(key, &blob);
+	if (status) {
+		KEY_TRACE("Blob operation fail");
+		goto out;
```
jenswi-linaro, line 528: Is freeing needed? Else just return the error.

core/drivers/crypto/caam/acipher/caam_dh.c
jenswi-linaro: It seems that the test is broken. Is it possible to fix the test? It would be a shame to enforce limits here due to broken tests.

jenswi-linaro (commented):
With the comments addressed, for the commits:
- "drivers: caam: add CAAM key driver"
- "drivers: caam: add CAAM key support for RSA"
- "drivers: caam: add CAAM key support for DSA"
- "drivers: caam: add CAAM key support for ECC"
- "drivers: caam: add CAAM key support for DH"
please apply:

core/drivers/crypto/caam/caam_key.c
clementfaure: Ok

core/drivers/crypto/caam/caam_key.c
clementfaure: Let's return the error indeed

core/drivers/crypto/caam/caam_key.c
clementfaure: Let's return the error indeed

core/drivers/crypto/caam/caam_key.c
clementfaure: same

clementfaure:
Force-push: rebase + fix compilation warning.

jforissier:
@clementfaure thank you, this is ready for merging now!

PR #5991, opened 2023-05-03
crypto: se050: compare symmetric keys in constant time

ldts:
Symmetric keys should be compared in constant time to protect against side channel attacks.


jforissier (approved):

ldts:
@jforissier i think it is ok to merge

PR #5994, opened 2023-05-03
Removing TPM driver from OP-TEE core

etienne-lms:
Using a TPM driver from OP-TEE prevent non-secure world to leverage TPM software stack and features.
For OP-TEE to use TPM for secure services, OP-TEE core will rather rely on a non-secure remote service on the same principles as the RPMB based secure storage. This change does not address this later case, it only removes code currently not used in OP-TEE core.

Comments are welcome of course if one needs such TPM driver from OP-TEE core and proposes some service integration.


jenswi-linaro (commented):

etienne-lms:
tag applied

jforissier:
Is it not a problem to remove things before the alternative is implemented?

jenswi-linaro:
This is as far as I know not used or complete.

jenswi-linaro:
This is as far as I know not used or complete.

jforissier:
> This is as far as I know not used or complete.

OK

PR #6002, opened 2023-05-05
Non-secure interrupt handling through FF-A

imre-kis-arm:
Add FFA_INTERRUPT and FFA_RUN support for signaling non-secure interrupts and for resuming to the secure world. If a secure partition is preempted by a non-secure interrupt OP-TEE saves the SP's state and sends an FFA_INTERRUPT to the normal world. After handling the interrupt the normal world should send an FFA_RUN to OP-TEE so it can continue running the SP.
If OP-TEE is the active FF-A endpoint (i.e. it is running TAs) the non-secure interrupts are signaled by the existing OPTEE_FFA_YIELDING_CALL_RETURN_INTERRUPT message instead of FFA_INTERRUPT.
The way of handling NS interrupts (signaled or queued) when a given SP is running is determined by the ns-interrupts-action SP manifest property. If an SP is called by another SP but which has stricter NS interrupt handling mode than the callee, the callee inherits the mode of the caller.


core/arch/arm/include/kernel/secure_partition.h
```diff
@@ -44,6 +44,8 @@ struct sp_session {
 	const void *fdt;
 	bool is_initialized;
 	TEE_UUID ffa_uuid;
+	uint32_t ns_interrupts_action;
```
jenswi-linaro, line 47: The names of these fields are too long.

imre-kis-arm: Ok, I'm changing them to something shorter.

core/arch/arm/kernel/thread_spmc_a64.S
```diff
@@ -171,13 +185,29 @@ END_FUNC thread_rpc
  * The current thread as indicated by @thread_index has just been
  * suspended.  The job here is just to inform normal world the thread id to
  * resume when returning.
+ * If the active FF-A endpoint is OP-TEE (or a TA) then an this function send an
+ * OPTEE_FFA_YIELDING_CALL_RETURN_INTERRUPT message to the normal world via the
```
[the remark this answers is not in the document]
imre-kis-arm: Is that a use case that someone calls OP-TEE or its TAs directly via the FF-A driver? If yes, the issue is already present without this patch, because OP-TEE currently sends `OPTEE_FFA_YIELDING_CALL_RETURN_INTERRUPT` in an `FFA_MSG_SEND_DIRECT_RESP` in case of a non-secure interrupt.
Until now the SPs were not preemptible by non-secure interrupts. This PR adds this and if an SP was running when the non-secure interrupt happened it sends `FFA_INTERRUPT`.
We still need to keep the original behavior for all other cases, because otherwise the OP-TEE driver won't be able to wake up thread waiters here: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/tee/optee/ffa_abi.c?h=v6.1.27#n593
Here I've relied on the active FF-A endpoint ID stored in rpc_target_info to distinguish between the two cases.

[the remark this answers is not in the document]
jenswi-linaro: > Here I've relied on the active FF-A endpoint ID stored in thread->tsd.rpc_target_info to distinguish between the two cases.

I'm not so keen on that since there is actually an overlap in case non-secure interrupts are unmasked at S-EL1. As long as OP-TEE core doesn't unmask non-secure interrupts while in S-EL1 while handling S-EL0 SPs it should still work. However, it's a bit fragile.

A new `THREAD_FLAGS_FFA_ONLY` flag to indicate that the normal OP-TEE MSG ABI isn't in use is more explicit. This can be used to stop any RPC while handling S-EL0 SPs.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1351,8 +1353,23 @@ static TEE_Result sp_enter_invoke_cmd(struct ts_session *s,
 	sp_regs->cpsr = read_daif() & (SPSR_64_DAIF_MASK << SPSR_64_DAIF_SHIFT);
 
 	exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
+
+	/*
+	 * Store endpoint ID and thread ID in rpc_target_info. This will be used
+	 * as w1 in FFA_INTERRUPT in case of a NWd interrupt.
```
etienne-lms, line 1428: s/NWd inteerupt/foreign interrupt/  (unless NWd is FF-A defined acronym)

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1351,8 +1353,23 @@ static TEE_Result sp_enter_invoke_cmd(struct ts_session *s,
...
+	 */
+	rpc_target_info = thread_get_tsd()->rpc_target_info;
+	thread_id = thread_get_id();
+	assert((thread_id & ~0xffff) == 0);
```
etienne-lms, line 1432: `assert(thread_id <= UINT16_MAX)` would be better (IMH-nitpciking-O)

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1351,8 +1353,23 @@ static TEE_Result sp_enter_invoke_cmd(struct ts_session *s,
...
+	thread_get_tsd()->rpc_target_info = (sp_s->endpoint_id << 16) |
+					    (thread_id & 0xffff);
```
etienne-lms, line 1434: masking `thread_id` is needed?

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1351,8 +1353,23 @@ static TEE_Result sp_enter_invoke_cmd(struct ts_session *s,
...
+
 	__thread_enter_user_mode(sp_regs, &panicked, &panic_code);
+
 	sp_regs->cpsr = cpsr;
+	/* Restore rpc_target_info */
```
etienne-lms, line 1438: empty line above

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -366,6 +366,32 @@ int spmc_sp_add_share(struct ffa_rxtx *rxtx,
 	return res;
 }
 
+void spmc_sp_set_to_preempted(struct ts_session *ts_sess)
+{
+	if (ts_sess && is_sp_ctx(ts_sess->ctx)) {
+		struct sp_session *sp_sess = to_sp_session(ts_sess);
+
+		assert(sp_sess->state == sp_busy);
+
```
etienne-lms, line 375: could remove that empty line

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -45,7 +45,7 @@ struct mem_frag_state {
 #endif
 
 /* Initialized in spmc_init() below */
-static uint16_t my_endpoint_id;
+uint16_t my_endpoint_id;
```
etienne-lms, line 48: declaration missing in some header file

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -438,6 +438,32 @@ void spmc_handle_partition_info_get(struct thread_smc_args *args,
 		      FFA_PARAM_MBZ, FFA_PARAM_MBZ);
 	cpu_spin_unlock(&rxtx->spinlock);
 }
+
+static void spmc_handle_run(struct thread_smc_args *args)
+{
+	uint16_t endpoint = (args->a1 >> 16) & 0xffff;
+	uint16_t thread_id = (args->a1 & 0xffff);
```
etienne-lms, line 445: can remove parentheses.

I think a `SP_THREAD_ID_MASK` macro would be nice as used at several places.

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -438,6 +438,32 @@ void spmc_handle_partition_info_get(struct thread_smc_args *args,
...
+	uint32_t rc = 0;
```
etienne-lms, line 446: `FFA_OK`?

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

core/arch/arm/kernel/spmc_sp_handler.c
imre-kis-arm: Done

core/arch/arm/kernel/thread_spmc.c
imre-kis-arm: Done

core/arch/arm/kernel/thread_spmc.c
imre-kis-arm: The [discussion above](https://github.com/OP-TEE/optee_os/pull/6002#discussion_r1188694570) can affect this part, we might not need to use this variable outside of the file. I would wait with this until we figure it out.

core/arch/arm/kernel/thread_spmc.c
imre-kis-arm: Actually, this part is defined in the FF-A spec. I've added macros for packing and unpacking these fields.

core/arch/arm/kernel/thread_spmc_a64.S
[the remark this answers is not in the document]
jenswi-linaro: It should be added to `threads[n].state` when called from `thread_sp_alloc_and_run()`. I guess that means adding a `flags` parameter to `__thread_alloc_and_run()`.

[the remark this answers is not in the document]
imre-kis-arm: Ok, thank you, now I get it.
Done

core/arch/arm/kernel/thread_spmc_a64.S
```diff
@@ -171,13 +185,32 @@ END_FUNC thread_rpc
...
+	/* load threads[w0].flags into w2 */
+	ldr	w2, [x1, #THREAD_CTX_FLAGS]
+	/* load threads[w0].tsd.rpc_target_info into w1 */
+	ldr	w1, [x1, #THREAD_CTX_TSD_RPC_TARGET_INFO]
+	and     w2, w2, #THREAD_FLAGS_FFA_ONLY
+	cmp	x2, 0
+	bne	thread_ffa_interrupt
+#else
+	/* load threads[w0].tsd.rpc_target_info into w1 */
 	ldr	w1, [x1, #THREAD_CTX_TSD_RPC_TARGET_INFO]
```
jenswi-linaro, line 212: I'd rather not do this in an `#else`, please refactor to be able to initialize `w1` before the `#if` above.

core/arch/arm/kernel/thread_spmc.c
imre-kis-arm: I reverted this line.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1211,6 +1215,36 @@ static TEE_Result handle_hw_features(void *fdt)
...
+	case SP_MANIFEST_NS_INT_SIGNALED:
+		/* OK */
+		break;
+
+	case SP_MANIFEST_NS_INT_MANAGED_EXIT:
+		EMSG("Managed exit is not implemented");
+		return TEE_ERROR_NOT_IMPLEMENTED;
+
+	default:
+		EMSG("Invalid ns-interrupts-action value: %d", s->ns_int_mode);
```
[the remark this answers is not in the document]
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1329,17 +1367,44 @@ TEE_Result sp_enter(struct thread_smc_args *args, struct sp_session *sp)
...
+		struct sp_session *caller_sp = to_sp_session(caller);
+
+		s->ns_int_mode_inherited = MIN(caller_sp->ns_int_mode_inherited,
+					       s->ns_int_mode);
+	} else {
+		s->ns_int_mode_inherited = s->ns_int_mode;
+	}
+
+	if (s->ns_int_mode_inherited == SP_MANIFEST_NS_INT_QUEUED)
+		*cpsr |= (THREAD_EXCP_FOREIGN_INTR << ARM32_CPSR_F_SHIFT);
```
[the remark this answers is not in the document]
imre-kis-arm: Done

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1329,17 +1367,44 @@ TEE_Result sp_enter(struct thread_smc_args *args, struct sp_session *sp)
...
 
+/*
+ * According to FF-A v1.1 section 8.3.1.4 if a caller requires less permissive
+ * active on NS interrupt than the callee, the callee must inherit the caller's
+ * configuration.
+ * Each SP's own NS action setting is stored in ns_int_mode. The effective
+ * action will be MIN([self action], [caller's action]) which is stored in the
+ * ns_int_mode_inherited field.
+ */
+static void sp_cpsr_configure_foreing_interrupts(struct sp_session *s,
```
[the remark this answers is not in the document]
imre-kis-arm: Done

core/arch/arm/kernel/thread_spmc_a64.S
```diff
@@ -171,13 +185,32 @@ END_FUNC thread_rpc
...
 	mov	x1, #THREAD_CTX_SIZE
 	adr_l	x2, threads
 	madd	x1, x1, x0, x2
+#if CFG_SECURE_PARTITION
+	/* load threads[w0].flags into w2 */
+	ldr	w2, [x1, #THREAD_CTX_FLAGS]
+	/* load threads[w0].tsd.rpc_target_info into w1 */
+	ldr	w1, [x1, #THREAD_CTX_TSD_RPC_TARGET_INFO]
+	and     w2, w2, #THREAD_FLAGS_FFA_ONLY
+	cmp	x2, 0
```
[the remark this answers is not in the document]
imre-kis-arm: Done

core/arch/arm/kernel/thread_spmc_a64.S
imre-kis-arm: I've refactored this part.

core/arch/arm/kernel/spmc_sp_handler.c
```diff
@@ -1132,6 +1157,12 @@ void spmc_sp_msg_handler(struct thread_smc_args *args,
 			handle_mem_perm_set(args, caller_sp);
 			sp_enter(args, caller_sp);
 			break;
+		case 0xdeadbeef:
```
jenswi-linaro, line 1160: What's this?

jenswi-linaro:
Please drop the `Change-Id` tags, with that:

PR #6006, opened 2023-05-10
core: riscv: Fix macro for inline assembly

gagachang:
The "CSR_XXX" definitions are not only used for pure assembly code, but also used for inline assembly in C code. But using SHIFT_U64() in inline assembly will lead to compilation error, since this macro casts value to (uint64_t) first.

For example, "read_csr(CSR_XSTATUS)" will be expanded by pre-processor as "csrr a5,(((uint64_t)(1)<<(8))|0x000)", which is not acceptable for assembly language.

Fix this by using UINT64_C macro instead, since it suffixes the value with ULL instead of C type.

### DIFF
--- a/core/arch/riscv/include/encoding.h
+++ b/core/arch/riscv/include/encoding.h
@@ -316,37 +316,6 @@
 #define RISCV_PGSHIFT 12
 #define RISCV_PGSIZE (1 << RISCV_PGSHIFT)
 
-#ifndef __ASSEMBLER__
-
-#ifdef __GNUC__
-
-#define read_csr(reg) ({ unsigned long __tmp; \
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
If you look in `lib/libutils/ext/include/util.h` you see that we have two implementations of the macro `SHIFT_U64()`, one for C and one for assembly. I guess you could save some time by making sure that the correct version is selected for riscv too.

gagachang:
Dear @jenswi-linaro 

Update. Now those macros regard given `csr` as input operand of inline assembly, instead of using direct stringizing.

jforissier:
Clever fix :)

gagachang:
@jenswi-linaro @jforissier 
Thank you! Tags applied.

PR #6009, opened 2023-05-10
bit_ffs_from() + fix typo in notif inline comment

etienne-lms:
Patches extracted from https://github.com/OP-TEE/optee_os/pull/5793 and all ready reviewed (R-b tag applied).


jenswi-linaro (approved):

jforissier:
QEMU with Xen has timed out, I think I have seen this happen occasionally, we might have a bug hiding somewhere... restarting the job for now.

Issue #6005, opened 2023-05-09
PKCS#11 destroy object can cause access to freed pointer in multisession access


vting-seq:
If there are two PKCS#11 sessions running, a token object being destroyed in one of those sessions can result in the access of that freed pointer in the other session causing a data-abort.

The token object database is global w.r.t all sessions (see init_persistent_db).  In object.c : entry_find_objects_init, the global token database is scanned for matching objects.  If one is found (and a handle to it is not found in the session), a handle is created for it, and the raw pointer is added to the handle list for that session.

The problem is that if the _other_ session destroys that token object, that pointer (and handle) should no longer be valid.  If the session is in the middle of using that handle - pkcs11_handle2object will still return that destroyed pointer - leading to a data-abort.

I can sort of fix this by iterating through **all** the sessions and doing a handle_put for the object (i.e. removing the handle --> pointer reference) in destroy_object.  The other session using the handle will then return an error (invalid handle) the next time it is used.  However I am not sure if it is really safe, as it might still be possible if one session has already converted the handle to the pointer, during which the object is destroyed.

To see this behaviour, in one session:

```
C_FindObjectsInit(session_0,template,1);
C_FindObjects(session_0,&object,1,&objcount);

------(second session)----------------------->  C_DestroyObject(session_1,object)

C_GetAttributeValue(session_0,object,objtemplate,objtemplatecount);  <---- will cause data-abort
```

Note:  object must be a persistent token object.

etienne-lms:
You are right. There is an issue here.
I've create a test case to trigger the issue (https://github.com/OP-TEE/optee_test/pull/665) and a change in optee_os to addressed it (https://github.com/OP-TEE/optee_os/pull/6018).

PR #6018, opened 2023-05-14
ta: pkcs11: invalidate handles on destroyed token object

etienne-lms:

### DIFF
--- a/ta/pkcs11/src/handle.c
+++ b/ta/pkcs11/src/handle.c
@@ -67,11 +67,16 @@ uint32_t handle_get(struct handle_db *db, void *ptr)
 	return n;
 }
 
+static bool handle_is_valid(struct handle_db *db, uint32_t handle)
+{
+	return db && handle && handle < db->max_ptrs;
+}
+
 void *handle_put(struct handle_db *db, uint32_t handle)
 {
 	void *p = NULL;
 
-	if (!db || !handle || handle >= db->max_ptrs)
+	if (!handle_is_valid(db, handle))
 		return NULL;
 
 	p = db->ptrs[handle];
@@ -81,7 +86,7 @@ void *handle_put(struct handle_db *db, uint32_t handle)
 
 void *handle_lookup(struct handle_db *db, uint32_t handle)
 {
-	if (!db || !handle || handle >= db->max_ptrs)
+	if (!handle_is_valid(db, handle))
 		return NULL;
 
 	return db->ptrs[handle];
--- a/ta/pkcs11/src/handle.c
+++ b/ta/pkcs11/src/handle.c
@@ -17,6 +17,9 @@
  */
 #define HANDLE_DB_INITIAL_MAX_PTRS	4
 
+/* Specific pointer ~0 denotes a still allocated but invalid handle */
+#define INVALID_HANDLE_PTR	((void *)~0)
+
 void handle_db_init(struct handle_db *db)
 {
 	TEE_MemFill(db, 0, sizeof(*db));
@@ -37,7 +40,7 @@ uint32_t handle_get(struct handle_db *db, void *ptr)
 	void *p = NULL;
 	uint32_t new_max_ptrs = 0;
 
-	if (!db || !ptr)
+	if (!db || !ptr || ptr == INVALID_HANDLE_PTR)
 		return 0;
 
 	/* Try to find an empty location (index 0 is reserved as invalid) */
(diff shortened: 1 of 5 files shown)


etienne-lms:
Dear pkcs11 TA contributors (@vesajaaskelainen, @ldts, @varder and few others), may I ask some review comments on this change?

ta/pkcs11/src/handle.c
```diff
@@ -67,11 +67,16 @@ uint32_t handle_get(struct handle_db *db, void *ptr)
 	return n;
 }
 
+static bool db_and_handle_are_valid(struct handle_db *db, uint32_t handle)
```
vesajaaskelainen, line 73: Even thou this is a static I would prefer to keep `handle_` prefix for method to make debugging life nicer.

Suggesting:
```c
static bool handle_is_valid(struct handle_db *db, uint32_t handle)
```

ta/pkcs11/src/handle.h
```diff
@@ -47,4 +47,6 @@ void *handle_lookup(struct handle_db *db, uint32_t handle);
 /* Return the handle associated to a pointer if found, else return 0 */
 uint32_t handle_lookup_handle(struct handle_db *db, void *ptr);
 
+/* Invalidate the reference referred by an allocated handle */
+void handle_invalidate(struct handle_db *db, uint32_t handle);
 #endif /*PKCS11_TA_HANDLE_H*/
```
vesajaaskelainen, line 52: add empty like before `#endif.`..

ta/pkcs11/src/handle.c
```diff
@@ -86,17 +89,28 @@ void *handle_put(struct handle_db *db, uint32_t handle)
...
 	return db->ptrs[handle];
 }
 
+void handle_invalidate(struct handle_db *db, uint32_t handle)
+{
+	if (db_and_handle_are_valid(db, handle)) {
+		if (!db->ptrs[handle])
+			TEE_Panic(TEE_ERROR_GENERIC);
+
+		db->ptrs[handle] = INVALID_HANDLE_PTR;
```
vesajaaskelainen, line 105: If I read correctly -- this will effectively mark handle ID "consumed" forever?

Eg. we have `(uint32_t)~0 - 2` handles and if they have been invalidated then game is over?

vesajaaskelainen:
> Dear pkcs11 TA contributors (@vesajaaskelainen, @ldts, @varder and few others), may I ask some review comments on this change?

Sure... I will think again this with a bit fresher brain.

varder:
Reviewed-by: Valerii Chubar <valerii_chubar@epam.com>

varder:
Reviewed-by: Valerii Chubar <valerii_chubar@epam.com>

varder:
'Reviewed-by: Valerii Chubar <valerii_chubar@epam.com>'

etienne-lms:
Thanks @varder for the feedback.

ta/pkcs11/src/handle.c
vesajaaskelainen: What I see happening in our devices is that the connection could be active for months -- meaning the connection to PKCS#11 TA is open for months :)

BUT I do not expect to see too much activity. There could be new keys pairs created in limited fashion, some certs added/remove. So in real life I suppose if some one would really try to exhaust the handles then the TA's HEAP size would be the real limit. That being said HEAP size limit comes much much faster.

vesajaaskelainen:
@etienne-lms can you squash the review commits to correct commits.

Looks OK so far.

ta/pkcs11/src/handle.c
vesajaaskelainen: Actually we could compress the list... eg. ID just going forward and then use memory for those objects that are actually available. But again -- perhaps monitor the situation first.

etienne-lms: The objects are destroyed so there is no problem with memory used for objects. It is only the handles that remain, and the space allocated for the client handle database which could grow and exhaust the TA heap. Franlky i see no way out but a redesign of the session object handles database.

That said, the issue reported reported by @vting-seq (https://github.com/OP-TEE/optee_os/issues/6005) is real and I think this P-R resolves it. So I would greatly prefer we merge this P-R and take time to address the heap exhaustion  you foresee. Or maybe I did not understand the actuallity of this heap issue.

etienne-lms:
@vesajaaskelainen, may I apply you Acked-by tag on this series?

etienne-lms:
fixup commmits squashed

vesajaaskelainen:
For commits:
- https://github.com/OP-TEE/optee_os/pull/6018/commits/234389ca402397ddaaddb7946a10153a442e615c - ta: pkcs11: factorize handle reference sanity test
- https://github.com/OP-TEE/optee_os/pull/6018/commits/1fec008ad5a78a41d469859f936e7103fbfb0533 - ta: pkcs11: invalidating handle in a database
- https://github.com/OP-TEE/optee_os/pull/6018/commits/c37fd2415d35d3bc2b9b89c35334fecbc66bc76d - ta: pkcs11: invalidate handle of destroyed token objects

`Reviewed-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

etienne-lms:
review tags applied. thanks.

PR #6024, opened 2023-05-15
Add missing conditional compilation for RISC-V

gagachang:
RV64 also uses kern_sp. The elf.h is also used by RV32 and RV64.


gagachang:
Tags applied!

PR #6025, opened 2023-05-15
plat-rzn1: increase DDR size to 1GB

rfs613:
There are now some RZ/N1 devices with 1GB rather than 256MB. The
first-stage bootloader does not support passing a DT to OP-TEE, so
static values are set at compile time. Increase the DDR size so as to
avoid OP-TEE calls failing with TEEC_ERROR_OUT_OF_MEMORY.

Signed-off-by: Ralph Siemsen <ralph.siemsen@linaro.org>

### DIFF
--- a/core/arch/arm/plat-rzn1/platform_config.h
+++ b/core/arch/arm/plat-rzn1/platform_config.h
@@ -11,7 +11,7 @@
 
 /* DRAM */
 #define DRAM_BASE			0x80000000
-#define DRAM_SIZE			0x10000000
+#define DRAM_SIZE			0x40000000 /* 1GB, and support 256MB */
 
 /* GIC */
 #define GIC_BASE			0x44100000


core/arch/arm/plat-rzn1/platform_config.h
```diff
@@ -11,7 +11,7 @@
 
 /* DRAM */
 #define DRAM_BASE			0x80000000
-#define DRAM_SIZE			0x10000000
+#define DRAM_SIZE			0x40000000
```
etienne-lms, line 14: Maybe on a platform flavor basis?
Up to @b49020 recommendation.

b49020: I would have better liked the DT based auto discovery of RAM approach, but since the proprietary boot-loader on RZN1 doesn't have the DT support yet and we have to live with it for the time being.

Also, having a different platform flavour just on the basis of different RAM size doesn't seems worth as compared to building platform specific binaries when a common tee.bin can work on both platforms.

AFAIK, the non-secure RAM is registered in OP-TEE just to service dynamic SHM requests from Linux. So just having a large enough memory space registered in OP-TEE such that it can handle differing RAM size requirements should be sufficient.

b49020: @rfs613 Its worth to add a code comment for `DRAM_SIZE` to clarify that it supports both `256MB` and `1GB` RAM sizes.

rfs613: Done

b49020:

rfs613:
I fixed the checkpatch warning.

jforissier:
Merged with Sumit's review tag added. Thanks!

PR #6026, opened 2023-05-16
core: riscv: kernel: Add missing initialization for core local stacks

gagachang:
The thread core local stacks should be initialized when the primary core performs system initialization.

Fixes: ca82589 ("core: split core/arch/arm/kernel/thread.c")

### DIFF
--- a/core/arch/riscv/kernel/boot.c
+++ b/core/arch/riscv/kernel/boot.c
@@ -66,6 +66,8 @@ void init_tee_runtime(void)
 
 static void init_primary(unsigned long nsec_entry)
 {
+	thread_init_core_local_stacks();
+
 	/*
 	 * Mask asynchronous exceptions before switch to the thread vector
 	 * as the thread handler requires those to be masked while


jforissier (commented):

core/arch/riscv/kernel/boot.c
```diff
@@ -66,6 +66,7 @@ void init_tee_runtime(void)
 
 static void init_primary(unsigned long nsec_entry)
 {
+	thread_init_core_local_stacks();
```
etienne-lms, line 69: nitpicking: an empty line below would be great.

etienne-lms (commented):
minor comment (`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment addressed or not)

core/arch/riscv/kernel/boot.c
gagachang: done

gagachang:
Tags applied!

PR #6027, opened 2023-05-16
[RFC PATCH] introduction of a remoteproc framework to load signed coprocessor firmware images

arnopo:
The aim of this PR is to propose the introcuction  of a remoteproc framework in OP-TEE to add secure services to:
- authenticate and load a signed firmware,
- manage the coprocessor live cycle,
- Isolate/protect the associated memories and resources during the live cycle.

This PR is quite huge, trying to give a whole picture of the implementation.  So it can be spitted on demand to
ease the review.

The implementation is made of:

1) A signature script (sign_rproc_fw.py) with an associated default key
  (default_rproc.pem)
  - The script allows to add several ELF firmware images  in one signed binary.
  - As the MCUBOOT firmware it add TLV (type, length, value) properties in a
    trailer blob, to provide information needed to authenticate and load the
    firmware.

2) A generic Trusted Application
  - it is an early TA, that  can be used by U-BOOT or Linux built-in driver.
  - only ELF format is supported yet.

3) A platform Pseudo TA
  - provide basic services platform dependent such as the signature verification
    the load and verification of a section.

4) a remoteproc driver
  - this platform driver is in charge of the remoteproc resources based on device tree,
  - In a second step it should also be able to manage the access right to the memories
    to ensure its isolations.

Code associated to the Linux Kernel is proposed as RFC for upstream:

Related pull request:

- https://github.com/OP-TEE/optee_os/pull/6048
- https://github.com/OP-TEE/optee_os/pull/6128
- https://github.com/OP-TEE/optee_os/pull/6300
- https://github.com/OP-TEE/optee_os/pull/6470


keys/default_rproc.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
jenswi-linaro, line 1: Why do we need a key different from the `default_ta.pem`?

arnopo: > Why do we need a key different from the `default_ta.pem`?

Mainly to compartmentalize the keys used for the remoteproc firmware image and the Trusted Application.  For example, the remoteproc firmware key could be shared with a third party responsible for providing the coprocessor firmware.

jenswi-linaro: In that case, have you considered https://optee.readthedocs.io/en/latest/architecture/subkeys.html?

etienne-lms (commented):
few minor comments

keys/default_rproc.pem
jenswi-linaro: > Thanks for the link. No i did not consider that. If I well understood your proposal is that the default_rproc key has to be considered as a subkey of the remoteproc TA , right?

No, the idea with subkeys is that you supply the chain of signed keys in the header. That way you can delegate signing of certain binaries without sharing the root key.

> So if that simplify the picture no issue to use the default_ta.pem'. Should I rename is to default_key.pem` in such case?

Sure, it might break signing of TAs in xtest, but that can be fixed also then.

arnopo:
CI Code style fail analysis:

  - `Macro argument reuse 'x' - possible side-effects?`
    check done, no side effect on macro

  - ` CHECK: Alignment should match open parenthesis`
i will add a fixup! to fix alignements

keys/default_rproc.pem
arnopo: Related fixup added on top :
https://github.com/OP-TEE/optee_os/pull/6027/commits/fe6990cade8508af04721205e372755067d59ce9
https://github.com/OP-TEE/optee_os/pull/6027/commits/961a570631fe81ab54e45e9b1365551947062b08

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -137,6 +139,18 @@ CFG_MMAP_REGIONS ?= 23
 CFG_DTB_MAX_SIZE ?= (256 * 1024)
 CFG_CORE_ASLR ?= n
 
+CFG_RPROC_PTA ?= n
+ifeq ($(CFG_RPROC_PTA),y)
+# Remoteproc early TA for coprocessor firmware management in boot stages
+CFG_IN_TREE_EARLY_TAS += remoteproc/80a4c275-0a47-4905-8285-1486a9771a08
+
+# Disable early TA compression to limit HEAP size
+CFG_EARLY_TA_COMPRESS ?= n
```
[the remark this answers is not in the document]
arnopo: done in https://github.com/OP-TEE/optee_os/pull/6027/commits/91d6a61f9950e6c71d374dbbce7343f8e25cdbbc

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,391 @@
...
+#include <kernel/dt_driver.h>
+#include <kernel/tee_misc.h>
+#include <libfdt.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+
+#define TIMEOUT_US_1MS	U(1000)
+
+/**
+ * struct stm32_rproc_mem describes memory regions used by the remote processor.
```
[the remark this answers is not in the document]
arnopo: done in https://github.com/OP-TEE/optee_os/pull/6027/commits/84cd779e63dd55985581f2f23defb2335f6c3720

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -137,6 +139,18 @@ CFG_MMAP_REGIONS ?= 23
 CFG_DTB_MAX_SIZE ?= (256 * 1024)
 CFG_CORE_ASLR ?= n
 
+CFG_RPROC_PTA ?= n
+ifeq ($(CFG_RPROC_PTA),y)
+# Remoteproc early TA for coprocessor firmware management in boot stages
+CFG_IN_TREE_EARLY_TAS += remoteproc/80a4c275-0a47-4905-8285-1486a9771a08
+
```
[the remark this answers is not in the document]
arnopo: done in https://github.com/OP-TEE/optee_os/pull/6027/commits/91d6a61f9950e6c71d374dbbce7343f8e25cdbbc

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,350 @@
...
+	memcpy(dst, src, size);
+
+	/* Verify that loaded segment is valid */
+	res = hash_sha256_check(hash, dst, size);
+	if (res) {
+		memset(dst, 0, size);
+		goto out;
+	}
+
+	res = TEE_SUCCESS;
```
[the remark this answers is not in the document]
arnopo: https://github.com/OP-TEE/optee_os/pull/6027/commits/ab2eb653742669d07ee3bfd813f84c846efbcb56

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,350 @@
...
+	case PTA_RPROC_FIRMWARE_START:
+		return rproc_pta_start(param_types, params);
+	case PTA_RPROC_FIRMWARE_STOP:
+		return rproc_pta_stop(param_types, params);
+	case PTA_RPROC_FIRMWARE_DA_TO_PA:
+		return rproc_pta_da_to_pa(param_types, params);
+	case PTA_RPROC_VERIFY_DIGEST:
+		return rproc_pta_verify_digest(param_types, params);
+	default:
+		break;
```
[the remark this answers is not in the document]
arnopo: done in https://github.com/OP-TEE/optee_os/pull/6027/commits/ab2eb653742669d07ee3bfd813f84c846efbcb56

ta/remoteproc/elf_parser.c
```diff
@@ -0,0 +1,188 @@
...
+#include <string.h>
+#include <trace.h>
+#include <types_ext.h>
+#include <util.h>
+
+static bool va_in_fwm_image_range(void *va, uint8_t *fw, size_t fw_size)
+{
+	uint8_t *vaddr = va;
+
+	assert(fw + fw_size >= fw);
```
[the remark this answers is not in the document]
arnopo: done in https://github.com/OP-TEE/optee_os/pull/6027/commits/7879cf77b77eac51f05e946bd387122ef6a6c4ed

keys/default_rproc.pem
arnopo: > CI Code style fail analysis:
> 
> * `Macro argument reuse 'x' - possible side-effects?`
>   check done, no side effect on macro
> * ` CHECK: Alignment should match open parenthesis`
>   i will add a fixup! to fix alignements

Alignement done in https://github.com/OP-TEE/optee_os/pull/6027/commits/84cd779e63dd55985581f2f23defb2335f6c3720

ta/remoteproc/elf_parser.c
```diff
@@ -0,0 +1,190 @@
...
+
+	if (!load_seg || fw + fw_size <= fw)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!IS_ALIGNED_WITH_TYPE(phdr, uint32_t) ||
+	    !va_in_fwm_image_range(phdr, fw, fw_size))
+		return TEE_ERROR_BAD_FORMAT;
+
+	for (i = 0; i < ehdr->e_phnum; i++, phdr++) {
+		uint32_t dst = phdr->p_paddr;
```
etienne-lms, line 84: `paddr_t` as physical address may not be 32bit.
Here and at several other place.

ta/remoteproc/elf_parser.c
```diff
@@ -0,0 +1,190 @@
...
+		uint8_t *src = NULL;
+
+		if (!va_in_fwm_image_range((void *)((vaddr_t)(phdr + 1) - 1),
+					   fw, fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		if (phdr->p_type != PT_LOAD)
+			continue;
```
etienne-lms, line 92: maybe move this below line 98, to catch first corrupted data.
Or move straight before line 87.

ta/remoteproc/elf_parser.c
```diff
@@ -0,0 +1,190 @@
...
+	unsigned int i = 0;
+	char *name_table = NULL;
+	struct resource_table *table = NULL;
+	Elf32_Ehdr *ehdr = (Elf32_Ehdr *)(void *)fw;
+	uint8_t *elf_data = fw;
+
+	if (fw + fw_size <= fw)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	shdr = (void *)(fw + ehdr->e_shoff);
```
etienne-lms, line 124: check if `fw + ehdr->e_shoff > fw`

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+					     paddr_t *rsc_pa,
+					     size_t *rsc_size)
+{
+	uint32_t da = 0;
+	TEE_Result res = TEE_ERROR_GENERIC;
+	Elf32_Word size = 0;
+
+	res = e32_parser_find_rsc_table(fw_img, fw_img_sz, &da, &size);
+	if (res)
+		goto err;
```
etienne-lms, line 706: if no ressource table (`TEE_ERROR_NO_DATA`), is that an error?

ta/remoteproc/include/elf_parser.h
```diff
@@ -0,0 +1,60 @@
...
+
+TEE_Result e32_parse_ehdr(uint8_t *fw, size_t size);
+TEE_Result e32_parser_load_elf_image(uint8_t *fw,  size_t fw_size,
+				     TEE_Result (*load_seg)(uint8_t *src,
+							    uint32_t size,
+							    uint32_t da,
+							    uint32_t mem_size,
+							    void *priv),
+				     void *priv_data);
+TEE_Result e32_parser_find_rsc_table(uint8_t *fw, size_t fw_size,
```
etienne-lms, line 56: add an empty line bewteen each declaration.
an inline description comment is also welcome.

ta/remoteproc/include/user_ta_header_defines.h
```diff
@@ -0,0 +1,29 @@
...
+
+#define TA_FLAGS			(TA_FLAG_DEVICE_ENUM | \
+					 TA_FLAG_SINGLE_INSTANCE | \
+					 TA_FLAG_INSTANCE_KEEP_ALIVE)
+
+/* Provisioned stack size */
+#define TA_STACK_SIZE			(4 * 1024)
+
+/* Provisioned heap size for TEE_Malloc() and friends */
+#define TA_DATA_SIZE			(4 * 1024)
```
etienne-lms, line 21: 4KByte is not much, is it enough for the TA?

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+#define RPROC_TLV_IMGTYPE	U(0x04)
+#define RPROC_TLV_IMGSIZE	U(0x05)
+#define RPROC_TLV_HASHTABLE	U(0x10)
+#define RPROC_TLV_PKEYINFO	U(0x11)
+
+#define RPROC_TLV_SIGNTYPE_LGTH U(1)
+
+#define LE16_TO_CPU(x) (((x)[1] << 8) | (x)[0])
+#define LE32_TO_CPU(x) (((x)[3] << 24) | ((x)[2] << 16) | \
+			((x)[1] << 8)  | (x)[0])
```
etienne-lms, line 112: could use `TEE_U32_BSWAP()` and `TEE_U16_BSWAP()` from _utee_defines.h_

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+/*
+ * struct remoteproc_tlv - Type-Length-Value structure
+ * @type: type of data
+ * @length: size of the data.
+ * @value: pointer to the data.
+ */
+struct remoteproc_tlv {
+	uint16_t type;
+	uint16_t length;
+	uint8_t value[0];
```
etienne-lms, line 126: s/[0]/[]/ for consistency in optee_os implementation.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+};
+
+/*
+ * struct remoteproc_segment - program header with hash structure
+ * @phdr: program header
+ * @hash: hash associated to the program segment.
+ */
+struct remoteproc_segment {
+	Elf32_Phdr phdr;
+	unsigned char hash[TEE_SHA256_HASH_SIZE];
```
etienne-lms, line 146: s/char/uint8_t/
ditto at all other place where hash is declare as a char array.

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,179 @@
...
+ * struct rproc_pta_key_info - public key information
+ * @algo:	Algorithm, defined by public key algorithms TEE_ALG_*
+ *		from TEE Internal API specification
+ * @info_size:	Byte size of the @info
+ * @info:	Append key information data
+ */
+struct rproc_pta_key_info {
+	uint32_t algo;
+	uint32_t info_size;
+	char info[];
```
etienne-lms, line 51: s/char/uint8_t/

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,179 @@
...
+{
+	size_t s = 0;
+
+	if (!keyinf || ADD_OVERFLOW(sizeof(*keyinf), keyinf->info_size, &s))
+		return 0;
+
+	return s;
+}
+
+#define RPROC_PTA_GET_KEYINFO_SIZE(x)	rproc_pta_get_keyinfo_size((x))
```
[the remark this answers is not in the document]
arnopo: I am not comfortable with this inline service. I don't see any inlin function in PTA API.
- either should it be a service,
- or should I call rproc_pta_get_keyinfo_size in the remoteproc TA,
- or should I duplicate the inline function in remoteproc_core .c and remoteproc_pta.c?

ta/remoteproc/elf_parser.c
```diff
@@ -0,0 +1,190 @@
...
+#include <string.h>
+#include <trace.h>
+#include <types_ext.h>
+#include <util.h>
+
+static bool va_in_fwm_image_range(void *va, uint8_t *fw, size_t fw_size)
+{
+	uint8_t *vaddr = va;
+
+	return vaddr >= fw && vaddr < fw + fw_size;
```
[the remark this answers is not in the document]
arnopo: This test should be  not done in this function. It tests that the fw area is  correct.
I have already put test in functions which call this one.
Please have look in https://github.com/OP-TEE/optee_os/commit/7879cf77b77eac51f05e946bd387122ef6a6c4ed to check if update answers to your point

ta/remoteproc/elf_parser.c
arnopo: `dst` is a  remote 32-bit remote processor address, that correspond to addresses stored in the ELF32 format.
`paddr_t` is the CPU address so can be either 32 bits or 64 bits.

In my opinion,  the 32 bit ELF parser we should not use 'paddr_t'...

ta/remoteproc/include/user_ta_header_defines.h
arnopo: I have never seen any overflow. 
on the other side  it depends on the firmware/ platform  (for instance `TEE_MALLOC `to store the  header and the trailer of the image).
It might make more sense to define a config (as done for `CFG_PKCS11_TA_HEAP_SIZE`), no ?

ta/remoteproc/remoteproc_core.c
arnopo: If I am not mistaken, `TEE_U32_BSWAP`  allows to swap between little and big endian

Here we want to convert from Little endian to CPU endian( so LE or BE).
So using TEE_U32_BSWAP should not work.

To answer to such requirement we should need equivalent of the Linux helper:
https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/byteorder

ta/remoteproc/elf_parser.c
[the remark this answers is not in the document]
etienne-lms: ok, right. It's done from all other functions that call this helper function.
Discard my comment.

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,390 @@
...
+
+	res = rstctrl_assert_to(rproc->hold_boot, TIMEOUT_US_1MS);
+	if (res)
+		return res;
+
+	res = rstctrl_assert_to(rproc->mcu_rst, TIMEOUT_US_1MS);
+	if (res)
+		return res;
+
+	return TEE_SUCCESS;
```
etienne-lms, line 115: minor suggestion, to shorten the function:
```c
	res = rstctrl_assert_to(rproc->hold_boot, TIMEOUT_US_1MS);
	if (!res)
		res = rstctrl_assert_to(rproc->mcu_rst, TIMEOUT_US_1MS);

	return res;
```

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,390 @@
...
+	 */
+	res = rstctrl_deassert_to(rproc->hold_boot, TIMEOUT_US_1MS);
+	if (res)
+		return res;
+
+	res = rstctrl_assert_to(rproc->hold_boot, TIMEOUT_US_1MS);
+	if (res)
+		return res;
+
+	return TEE_SUCCESS;
```
etienne-lms, line 97: minor suggestion, to shorten the function:
```c
	res = rstctrl_deassert_to(rproc->hold_boot, TIMEOUT_US_1MS);
	if (!res)
		res = rstctrl_assert_to(rproc->hold_boot, TIMEOUT_US_1MS);

	return res;

File: ta/remoteproc/remoteproc_core.c (hunk 11 above)
Comment (etienne-lms) [reply ta/remoteproc/remoteproc_core.c:706]: Ok, makes sense. In that case you should update the trace message accordingly:
`IMSG("Ignore empty resource table section");` in _ta/remoteproc/elf_parser.c_ line 145.

File: core/drivers/remoteproc/stm32_remoteproc.c [hunk 22]
```diff
@@ -0,0 +1,390 @@
...
+		return TEE_ERROR_CORRUPT_OBJECT;
+
+	mems = rproc->regions;
+
+	for (i = 0; i < rproc->n_regions; i++) {
+		if (!core_is_buffer_inside(pa, size, mems[i].addr,
+					   mems[i].size))
+			continue;
+
+		*va = core_mmu_add_mapping(MEM_AREA_RAM_NSEC, pa, size);
```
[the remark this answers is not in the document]
etienne-lms: Ok. Maybe a trace message to mention that:
`IMSG("Warning: experimental feature, coprocess memory is not protected");

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,390 @@
...
+
+	for (i = 0; i < nranges; i++) {
+		uint32_t da = fdt32_to_cpu(list[i++]);
+		uint32_t pa = fdt32_to_cpu(list[i++]);
+		uint32_t size = fdt32_to_cpu(list[i]);
+
+		if (core_is_buffer_inside(region->addr, region->size,
+					  pa, size)) {
+			region->da = da + (region->addr - pa);
+			DMSG("device address found : %#x", region->da);
```
etienne-lms, line 255: `%#"PRIxPA, region->da)`

lib/libutee/include/remoteproc_pta.h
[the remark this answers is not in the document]
etienne-lms: inline functions  are fine.  They are better then macros: compiler can better check arguments types, return value etc...

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,179 @@
...
+ * @info:	Append key information data
+ */
+struct rproc_pta_key_info {
+	uint32_t algo;
+	uint32_t info_size;
+	char info[];
+};
+
+static inline size_t
+	rproc_pta_get_keyinfo_size(struct rproc_pta_key_info *keyinf)
```
etienne-lms, line 55: remove the heading tab. My suggestion: shorten the function label:
```c
static inline size_t rproc_pta_keyinfo_size(struct rproc_pta_key_info *keyinf)
```

ta/remoteproc/elf_parser.c
etienne-lms: You're right. Elf32 is... 32bit.

ta/remoteproc/include/user_ta_header_defines.h
etienne-lms: Yes, you can add such a config switch.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+ *                        +-------------+
+ *                        +-------------+
+ *                        |     TLV     |  TLV structures
+ *                        |   payload   |
+ *                        +-------------+
+ */
+
+/* Firmware state */
+enum remoteproc_state {
+	REMOTEPROC_OFF,
```
etienne-lms, line 73: Prefer not having 2 enum values with the same name (`REMOTEPROC_OFF` from _core/pta/stm32mp/remoteproc_pta.c_) but that not a strong opinion as each is in the scope of its respective source file.

ta/remoteproc/remoteproc_core.c
etienne-lms: `TEE_U32_BSWAP()` and `TEE_U16_BSWAP()` do the very exact same thing as done here by `LE32_TO_CPU()` and `LE16_TO_CPU()`.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+	if (!params[1].memref.buffer || !params[1].memref.size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	DMSG("Got base addr: %p size %x", params[1].memref.buffer,
+	     params[1].memref.size);
+
+	res = remoteproc_verify_firmware(ctx, params[1].memref.buffer,
+					 params[1].memref.size);
+	if (res) {
+		EMSG("Can't Authenticate the firmware, res = %#"PRIx32, res);
```
etienne-lms, line 1017: should print the firmware ID.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+TEE_Result TA_CreateEntryPoint(void)
+{
+	return TEE_SUCCESS;
+}
+
+void TA_DestroyEntryPoint(void)
+{
+}
+
+ /*
```
etienne-lms, line 1154: remove the extra space char starting this line as well as those starting lines 1155 to 1159.

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,179 @@
...
+
+/* Platform predefined TLV ID */
+/* boot address of secure fw */
+#define PTA_RPROC_TLV_SBOOTADDR            U(0x21)
+/* boot address of non-secure fw */
+#define PTA_RPROC_TLV_NSBOOTADDR           U(0x22)
+
+/* Platform predefined TLV LENGTH (byte) */
+#define PTA_RPROC_TLV_SBOOTADDR_LGTH       U(4)
+#define PTA_RPROC_TLV_NSBOOTADDR_LGTH      U(4)
```
etienne-lms, line 39: prefer tabs over space chars in the 4 macro definitions above

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+	res = remoteproc_load_elf(ctx);
+	if (res)
+		goto out;
+
+	ctx->state = REMOTEPROC_LOADED;
+
+out:
+	/* Clear reference to firmware image from shared memory */
+	ctx->fw_img = NULL;
+	ctx->fw_img_sz =  0;
```
etienne-lms, line 1030: Replace the 2 space chars with a single one.

There are several other occurrences of 2 space chars that could be removed in this P-R:
ta/remoteproc/remoteproc_core.c lines 112, 324, 424, 562, 576, 583, 743, 976, 1030 and 1031, 
lib/libutee/include/remoteproc_pta.h lines 93 and 171
ta/remoteproc/include/elf_parser.h lines 39 to 42, line 49
core/include/drivers/stm32_remoteproc.h lines 13 and 52

ta/remoteproc/remoteproc_core.c
arnopo: I did not use the same name as the state enum could be different in the TA and the PTA, means we could have specific states in PTA depending on the platform implementation.

ta/remoteproc/remoteproc_core.c
arnopo: What about `IMSG("Warning: No resource table found");` Print an error here seems too strong

core/drivers/remoteproc/stm32_remoteproc.c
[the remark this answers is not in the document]
arnopo: Done

core/include/drivers/stm32_remoteproc.h
```diff
@@ -0,0 +1,63 @@
...
+void *stm32_rproc_get(uint32_t fw_id);
+
+/*
+ * stm32_rproc_da_to_pa() - Convert the coprocessor device address to a loCPU
+ *                        physical address.
+ * @firmware_id unique identifier of the firmware
+ * @da		device memory address from the remote processor space
+ *		perspective.
+ * @size	size of the memory
+ * @pa		Output CPU physical address associated to the @da.
```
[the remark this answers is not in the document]
arnopo: Done

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	/* Copy the segment to the remote processor memory */
+	memcpy(dst, src, size);
+
+	/* Verify that loaded segment is valid */
+	res = hash_sha256_check(hash, dst, size);
+	if (res) {
+		memset(dst, 0, size);
+		goto out;
```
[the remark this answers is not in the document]
arnopo: Done

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,346 @@
...
+	const uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_VALUE_INPUT);
+	TEE_Result res = TEE_ERROR_GENERIC;
+	paddr_t pa = 0;
+	void *dst = NULL;
+	paddr_t da = params[1].value.a;
+	size_t size = params[2].value.a;
+	char value = (char)params[3].value.a;
```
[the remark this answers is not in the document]
arnopo: Done

lib/libutee/include/remoteproc_pta.h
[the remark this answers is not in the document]
arnopo: Done

ta/remoteproc/include/elf_parser.h
arnopo: Done

ta/remoteproc/include/user_ta_header_defines.h
arnopo: Done

ta/remoteproc/remoteproc_core.c
arnopo: Done

lib/libutee/include/remoteproc_pta.h
arnopo: Done

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+	TEE_DigestUpdate(op, hdr, sizeof(*hdr));
+	res = TEE_DigestDoFinal(op, ctx->tlr, U64_ALIGN_SZ(hdr->tlv_len) +
+				U64_ALIGN_SZ(hdr->plat_tlv_len),
+				hash, &hash_len);
+
+	if (res != TEE_SUCCESS)
+		goto out;
+
+	/*
+	 * TODO:
```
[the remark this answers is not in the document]
arnopo: Done

core/drivers/remoteproc/stm32_remoteproc.c
arnopo: Done

core/drivers/remoteproc/stm32_remoteproc.c
arnopo: Done

lib/libutee/include/remoteproc_pta.h
arnopo: Done

ta/remoteproc/remoteproc_core.c
arnopo: Done

ta/remoteproc/remoteproc_core.c
arnopo: Done

lib/libutee/include/remoteproc_pta.h
arnopo: Done

keys/default_rproc.pem
arnopo: > > Thanks for the link. No i did not consider that. If I well understood your proposal is that the default_rproc key has to be considered as a subkey of the remoteproc TA , right?
> 
> No, the idea with subkeys is that you supply the chain of signed keys in the header. That way you can delegate signing of certain binaries without sharing the root key.
> 
> > So if that simplify the picture no issue to use the default_ta.pem'. Should I rename is to default_key.pem` in such case?
> 
> Sure, it might break signing of TAs in xtest, but that can be fixed also then.

I created a PR to rename the key file name: https://github.com/OP-TEE/optee_os/pull/6048

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+		if (res)
+			goto out;
+
+		/* Take opportunity to get the resource table address */
+		res = remoteproc_parse_rsc_table(ctx, ctx->fw_img + offset,
+						 img_size, &rsc_pa, &rsc_size);
+		if (res != TEE_SUCCESS && res != TEE_ERROR_NO_DATA)
+			goto out;
+
+		if (res != TEE_ERROR_NO_DATA) {
```
etienne-lms, line 953: prefer `if (res == TEE_SUCCESS)`

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+	}
+
+	/*
+	 * If res == TEE_ERROR_NO_DATA then no resource table found.
+	 * Force to TEE_SUCCESS as the resource table is optional.
+	 */
+	res = TEE_SUCCESS;
+
+out:
+	/* TODO: clean-up the memories in case of fail */
```
etienne-lms, line 978: You can clear the memory content. Use `memzero_explicit()` if necessary.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+
+	if (ctx->state == REMOTEPROC_OFF)
+		return TEE_ERROR_BAD_STATE;
+
+	reg_pair_from_64((uint64_t)ctx->rsc_pa,
+			 &params[1].value.b, &params[1].value.a);
+	reg_pair_from_64((uint64_t)ctx->rsc_size,
+			 &params[2].value.b, &params[2].value.a);
+
+	return TEE_SUCCESS;
```
etienne-lms, line 1142: I suggest to return `TEE_ERROR_NO_DATA` in case `rsc_size` is zero.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+	res = remoteproc_get_tlv(ctx->tlr, RPROC_TLV_NUM_IMG,
+				 &tlv, &length);
+	if (res)
+		goto out;
+	if (length != sizeof(uint8_t)) {
+		res = TEE_ERROR_BAD_FORMAT;
+		goto out;
+	}
+
+	num_img = *tlv;
```
etienne-lms, line 925: check if `num_img` is 0

arnopo: Done

ta/remoteproc/remoteproc_core.c
arnopo: Unfortunately  the Linux kernel does not have errors definition, so it not possible to treat the specific use case where no resource table is define with that error. 
Using TEE_ERROR_NO_DATA  will also be in contradiction with https://github.com/OP-TEE/optee_os/pull/6027#discussion_r1206679595

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,176 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2023, STMicroelectronics - All Rights Reserved
+ */
+
+#ifndef __REMOTEPROC_PTA_H
+#define __REMOTEPROC_PTA_H
+
+#include <util.h>
```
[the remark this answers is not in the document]
arnopo: Done

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,343 @@
...
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (rproc_ta_state != REMOTEPROC_ON)
+		return TEE_ERROR_BAD_STATE;
+
+	res = stm32_rproc_stop(params[0].value.a);
+	if (res)
+		return res;
+
+	/* TODO clean the memory and change protection right*/
```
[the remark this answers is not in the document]
arnopo: yes i will add a memory clean

ta/remoteproc/remoteproc_core.c
arnopo: No  possible here we don't know which memory should be cleaned. the memory should be cleaned by the platform. storing a list of memory segment filled seems not appropriate. 
The question is does we need to clean the memory , I changed the TODO by a comment with a question.
if yes either we do nothing here or we call a new PTA service to clean-up the memory

ta/remoteproc/remoteproc_core.c
arnopo: done

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+		if (peh->phdr.p_paddr != da)
+			continue;
+
+		/*
+		 * Segment is read from a non secure memory. Crosscheck it using
+		 * the hash table to verify that the segment has not been
+		 * corrupted.
+		 */
+		if (peh->phdr.p_type != PT_LOAD)
+			return TEE_ERROR_CORRUPT_OBJECT;
```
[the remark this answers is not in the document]
arnopo: done

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1212 @@
...
+		}
+	}
+
+	*rsc_size = size;
+
+	return TEE_SUCCESS;
+
+err:
+	*rsc_size = 0;
+	*rsc_pa = 0;
```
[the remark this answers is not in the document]
arnopo: done

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,176 @@
...
+ * struct rproc_pta_key_info - public key information
+ * @algo:	Algorithm, defined by public key algorithms TEE_ALG_*
+ *		from TEE Internal API specification
+ * @info_size:	Byte size of the @info
+ * @info:	Append key information data
+ */
+struct rproc_pta_key_info {
+	uint32_t algo;
+	uint32_t info_size;
+	uint8_t  info[];
```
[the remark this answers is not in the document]
arnopo: done

core/pta/stm32mp/sub.mk
```diff
@@ -1 +1,10 @@
 srcs-$(CFG_STM32_BSEC_PTA) += bsec_pta.c
+srcs-$(CFG_RPROC_PTA) += remoteproc_pta.c
+
+ifeq ($(CFG_RPROC_PTA),y)
+gensrcs-y += rproc_pub_key
+produce-rproc_pub_key = rproc_pub_key.c
+depends-rproc_pub_key = $(RPROC_SIGN_KEY) scripts/pem_to_pub_c.py
+recipe-rproc_pub_key = $(PYTHON3) scripts/pem_to_pub_c.py --prefix rproc_pub_key \
+        --key $(RPROC_SIGN_KEY) --out $(sub-dir-out)/rproc_pub_key.c
+endif
```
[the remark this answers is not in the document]
arnopo: Done

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,343 @@
...
+	struct ts_session *s = ts_get_calling_session();
+	const uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE);
+
+	if (pt != exp_pt)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	/* TODO: check that we're called by the remoteproc TA (check UUID) */
```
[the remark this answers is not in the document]
arnopo: Done

arnopo:
@jforissier, @jenswi-linaro, @etienne-lms 

I still have to address @etienne-lms  comment https://github.com/OP-TEE/optee_os/pull/6027#discussion_r1206664166 
but I propose to also 

- [ ] Squash and rebase this pull request to integrate https://github.com/OP-TEE/optee_os/pull/6048

- [ ] create in parallel a PR to satart the review  of the signature script  (https://github.com/OP-TEE/optee_os/pull/6027/commits/16a4633e6d134f1cc906e1f413b2bdf682fe08f7)

Are you ok with this approach?

etienne-lms:
I'm ok you squash the fixup commits in this series.
I'm also in favor of you creating a dedicated P-R for the coproc firmware signature script. It will allow to focus on own it's expected to be signed which will drive how it's expected to be verified.

arnopo:
I have rebased the PR on master(https://github.com/OP-TEE/optee_os/commit/9a54d4848dae866aa3f1e79fd4203dfda3c30cd9)

Check patch still fails due to  `CHECK: Macro argument reuse 'x' - possible side-effects?` reports, that are (in my opinion) a false positive

jforissier:
> Check patch still fails due to `CHECK: Macro argument reuse 'x' - possible side-effects?` reports, that are (in my opinion) a false positive

Indeed we have accepted those kind of warnings in the past, but I do think they are not that hard to fix in this case. For example:
```C
#define LE16_TO_CPU(x) ({ uint8_t *_x = x; ((_x)[1] << 8) | (_x)[0]; })
```
etc.

arnopo:
> > Check patch still fails due to `CHECK: Macro argument reuse 'x' - possible side-effects?` reports, that are (in my opinion) a false positive
> 
> Indeed we have accepted those kind of warnings in the past, but I do think they are not that hard to fix in this case. For example:
> 
> ```c
> #define LE16_TO_CPU(x) ({ uint8_t *_x = x; ((_x)[1] << 8) | (_x)[0]; })
> ```
> 
> etc.

Thank for the advice§ I tried different approach to fix that, without success.  I test your proposal

ta/remoteproc/include/elf_parser.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * This structure shall be consistent with the Linux kernel structure
+ * definition from include/linux/remoteproc.h.
+ */
+struct resource_table {
+	uint32_t ver;
+	uint32_t num;
+	uint32_t reserved[2];
+	uint32_t offset[];
+} __packed;
```
jenswi-linaro, line 33: Why `__packed`?

jenswi-linaro:
How come the TLVs are placed after the images? You needlessly get two "headers" to keep track of instead of just one at the beginning. Then there's also the alignment issue, do you pad the images to ensure the TLVs are at a well-aligned offset?

ta/remoteproc/include/elf_parser.h
arnopo: This table is defined in a shared memory and used by OPTEE but also the remote processor firmware. 
To avoid alignement issue we pack it. 
Same is done in linux:
https://elixir.bootlin.com/linux/latest/source/include/linux/remoteproc.h#L73

arnopo:
> How come the TLVs are placed after the images? You needlessly get two "headers" to keep track of instead of just one at the beginning. Then there's also the alignment issue, do you pad the images to ensure the TLVs are at a well-aligned offset?

FYI I created a PR dedicated to the signature script I propose to answer to this in the dedicated PR #6128

ta/remoteproc/include/elf_parser.h
jenswi-linaro: Fair enough, I'll skip my usual rant against `__packed` this time. But I'd appreciate a short comment motivating the `__packed` usage to avoid people copying this without understanding why. For instance that it's required by the ELF binary format, interoperability, or something along that.

arnopo: I've added a comment. I hope this is enough to clarify usage.

ta/remoteproc/include/elf_parser.h
```diff
@@ -24,6 +24,12 @@
  *
  * This structure shall be consistent with the Linux kernel structure
  * definition from include/linux/remoteproc.h.
+ *
+ * The resource_table structure is stored in a shared memory, accessed by the
```
jenswi-linaro, line 28: If this is the only reason for the packed attribute then you could just add a `static_assert(sizeof(struct resource_table) == 4 * sizeof(uint32_t))` instead. As a nice bonus, you'd allow the compile to generate efficient code.

However, I suspect that in the kernel the struct may end up at an unaligned address or offset and that's the main reason for the `__packed` attribute. If this can't happen in OP-TEE then there's no need for a `__packed` attribute that only causes confusion and inefficient code. If it can happen, just mention how and I'm sure the `__packed` attribute will make sense to a reader.

Code is much easier to understand when you don't add things for good measure or because it was added somewhere else for reasons unknown.

arnopo: I reanalyze the use cases. The Linux Kernel  parses the structure to get resources declared just after the `resource_table` structure in memory. It gets the resource addresses thanks to the `offset` field. The size of the `offset` field depends on the number of resources declared in `num` field.  In this case the __packed makes sense to avoid memory corruption, as the memory region is structured in several structures not 64-bit aligned.

The OP-TEE doesn't parse the resources. It just gets the start address of the structure and performs some basic checks on the `resource_table` structure. The size is retrieved from the `.resource_table` elf section.

So just removing the `__packed ` attribute seems OK. Adding the test with  `static_assert` seems useless and not accurate  due to the variability of offset field size.
I will test it and reworks the fixup

Thanks for your advise!

arnopo: I reworked the last fixup! ( force update) to remove the "__packed" and add a related comment

jenswi-linaro: OK, thanks.

arnopo:
> > How come the TLVs are placed after the images? You needlessly get two "headers" to keep track of instead of just one at the beginning. Then there's also the alignment issue, do you pad the images to ensure the TLVs are at a well-aligned offset?
> 
> FYI I created a PR dedicated to the signature script I propose to answer to this in the dedicated PR #6128

Answer post here :https://github.com/OP-TEE/optee_os/pull/6128#issuecomment-1602872882

arnopo:
@jenswi-linaro, @jforissier
"Could you please reopen the PR? It is still valid, even though I have sent some other PRs to split the upstream of this work. It contains the remaining code that I expect to upstream.

arnopo:
I re-based the Work on the master branch that integrates https://github.com/OP-TEE/optee_os/pull/6128
A new PR  https://github.com/OP-TEE/optee_os/pull/6300 has been created for the remoteproc TA .

arnopo:
Remove stale label, this PR is a used to show the full picture of the remoteproc loader implementation.

arnopo:
I close the RFC as the feature is now merged in following patches

https://github.com/OP-TEE/optee_os/pull/6048
https://github.com/OP-TEE/optee_os/pull/6128
https://github.com/OP-TEE/optee_os/pull/6300
https://github.com/OP-TEE/optee_os/pull/6470

@etienne-lms  @jenswi-linaro @jforissier: Thank you for your time and your reviews

PR #6030, opened 2023-05-17
core: interrupt: factorization & renaming

etienne-lms:
This change prepares changes to move to the interrupt management API function proposed in https://github.com/OP-TEE/optee_docs/pull/192 and to the generic interrupt chip framework proposed by https://github.com/OP-TEE/optee_os/pull/5954.

Commit "core: define main interrupt controller data from its driver" factorizes CPU **~~core~~ main** interrupt controller initialization and **~~root~~ main native interrupt** handling.

Commit "drivers: gic: factorize call to gic_init() or gic_init_base_addr()" makes `gic_init()` a unique entry point to initialize GIC driver.

Commit "core: interrupt: rename itr_core_handler()" renames `itr_core_handler()` to **~~`interrupt_core_handler()`~~ `interrupt_main_handler()`**.

Commit "core: interrupt: rename itr_init()" renames `itr_init()` to **~~`interrupt_core_init()`~~ `interrupt_main_init()`**.

**(edited) updated from review comments**

### DIFF
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -497,8 +497,8 @@ static void __maybe_unused gic_native_itr_handler(void)
 }
 
 #ifndef CFG_CORE_WORKAROUND_ARM_NMFI
-/* Override itr_core_handler() with core interrupt controller implementation */
-void itr_core_handler(void)
+/* Override interrupt_main_handler() with driver implementation */
+void interrupt_main_handler(void)
 {
 	gic_native_itr_handler();
 }
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -7,6 +7,7 @@
 #include <arm.h>
 #include <assert.h>
 #include <config.h>
+#include <compiler.h>
 #include <drivers/gic.h>
 #include <keep.h>
 #include <kernel/dt.h>
@@ -72,6 +73,15 @@
 #define GICC_IAR_CPU_ID_MASK	0x7
 #define GICC_IAR_CPU_ID_SHIFT	10
 
+struct gic_data {
+	vaddr_t gicc_base;
+	vaddr_t gicd_base;
+	size_t max_it;
+	struct itr_chip chip;
+};
+
+static struct gic_data gic_data __nex_bss;
+
 static void gic_op_add(struct itr_chip *chip, size_t it, uint32_t type,
 		       uint32_t prio);
 static void gic_op_enable(struct itr_chip *chip, size_t it);
@@ -135,8 +145,10 @@ out:
 	return ret;
 }
 
-void gic_cpu_init(struct gic_data *gd)
+void gic_cpu_init(void)
 {
+	struct gic_data *gd = &gic_data;
+
 #if defined(CFG_ARM_GICV3)
 	assert(gd->gicd_base);
 #else
@@ -166,11 +178,12 @@ void gic_cpu_init(struct gic_data *gd)
 #endif
 }
 
-void gic_init(struct gic_data *gd, paddr_t gicc_base_pa, paddr_t gicd_base_pa)
+void gic_init(paddr_t gicc_base_pa, paddr_t gicd_base_pa)
 {
+	struct gic_data *gd = &gic_data;
 	size_t n;
 
-	gic_init_base_addr(gd, gicc_base_pa, gicd_base_pa);
+	gic_init_base_addr(gicc_base_pa, gicd_base_pa);
 
 	for (n = 0; n <= gd->max_it / NUM_INTS_PER_REG; n++) {
 		/* Disable interrupts */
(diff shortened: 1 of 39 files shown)


core/drivers/gic.c
```diff
@@ -454,12 +493,22 @@ void gic_it_handle(struct gic_data *gd)
 	gic_write_eoir(gd, iar);
 }
 
+#ifndef CFG_CORE_WORKAROUND_ARM_NMFI
+/* Override itr_core_handler() with core interrupt controller implementation */
+void itr_core_handler(void)
+{
+	gic_native_itr_handler();
```
jenswi-linaro, line 503: Why not open code `gic_native_itr_handler()` here?

jenswi-linaro (commented):
Comment for "core: define main interrupt controller data from its driver"

jenswi-linaro (commented):
Comment for "core: define main interrupt controller data from its driver"

core/include/kernel/interrupt.h
```diff
@@ -57,7 +57,12 @@ struct itr_handler {
 	SLIST_ENTRY(itr_handler) link;
 };
 
-void itr_init(struct itr_chip *data);
+/*
+ * Initialise core interrupt controller driver
```
jenswi-linaro, line 61: I think `s/core/main/` would be more clear in case there is more than one interrupt controller.

jenswi-linaro (commented):
Comment for "core: define main interrupt controller data from its driver"

core/drivers/gic.c
etienne-lms: what do you mean by "open"?

jenswi-linaro: Copy the content of `gic_native_itr_handler()`.

etienne-lms: Because of `#ifndef CFG_CORE_WORKAROUND_ARM_NMFI`, I had to tag `gic_read_iar()` and `gic_write_eoir()` as `__maybe_unused`. I prefer the way implemented here where only `gic_native_itr_handler()` is tagged `__maybe_unused`. This way, GIC drvier handler function has a nice `gic_` prefixed label.
That said, I have no strong opinon here.

core/include/kernel/interrupt.h
etienne-lms: Agree. Are you ok I address your comment when squashing the fixup commits, instead of appending a messy fixup commit that affect all those platform files?

Note: I plan to update OP-TEE docs with this new labels.

etienne-lms: Sorry, I meesed-up. The many changed platform fiel are not impatced. I'll post a fixup commit.

core/drivers/gic.c
jenswi-linaro: OK, let's keep it as it is then.

core/drivers/gic.c
```diff
@@ -296,6 +315,8 @@ static void gic_it_set_cpu_mask(struct gic_data *gd, size_t it,
 	vaddr_t itargetsr = gd->gicd_base +
 			    GICD_ITARGETSR(it / NUM_TARGETS_PER_REG);
 
+	assert(gd == &gic_data);
```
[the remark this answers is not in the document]
jenswi-linaro: There's no need to assert in these functions which are called from the `gic_op_*()` functions.
I don't mind asserting that the passed `chip` in the `gic_op_*()` functions is supposed to match what we have in `gic_data`, it makes it clear what's expected.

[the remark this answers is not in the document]
etienne-lms: ok, so I keep as is.

jforissier:
> @jforissier, do these changes make sense to you? I'll wait for some status in this P-R before rebasing #5954.

They do seem to simplify things quite a bit and the diffstat is nice (+193 -339), so I'd say: yes, I'm fine with that.

jenswi-linaro:

etienne-lms:
fixup commit squashed and review tag applied.

jforissier:
It looks like RockPi4 was broken by this PR. When testing 4.7.0-rc1, I get a panic "Secure interrupt handler not defined". Anyone has an idea what's missing? Thanks.

etienne-lms:
You get this panic trace message when no `interrupt_main_handler()` defines the CPU main interrupt controller.
IIUC, RockPi4 is `PLATFORM=rockchip-rk3399` and embeds GIC driver **core/drivers/gic.c** without enabling `CFG_CORE_WORKAROUND_ARM_NMFI=y` hence `interrupt_main_handler()` should come from **gic.c** driver.

jforissier:
> You get this panic trace message when no `interrupt_main_handler()` defines the CPU main interrupt controller. IIUC, RockPi4 is `PLATFORM=rockchip-rk3399` and embeds GIC driver **core/drivers/gic.c** without enabling `CFG_CORE_WORKAROUND_ARM_NMFI=y` hence `interrupt_main_handler()` should come from **gic.c** driver.

`gic.c` is not compiled because `CFG_GIC` is not enabled due to https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-rockchip/conf.mk#L4 which was introduced by commit 447c5f6bc49f. If I add this file to the build anyways, `xtest` hangs in `regression_1003  Core internal read/write mutex`.

etienne-lms:
Right! indeed. I was fooled by https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-rockchip/conf.mk#L33 enforcing `CFG_ARM_GICV3=y`.
`interrupt_main_handler()` is called when the platform receives a secure (native) interrupt. What does the platform expect to do if there is no interrupt driver?

PR #6031, opened 2023-05-17
build: add -Werror to cc-option check

jforissier:
Clang emits a warning, not an error, and returns a success status when passed a -Wxxx option that it does not recognize. This makes the cc-option macro behave incorrectly with Clang when testing such options.

 $ aarch64-linux-gnu-gcc -Wdoes-not-exist -c -x c /dev/null
 aarch64-linux-gnu-gcc: error: unrecognized command-line option ‘-Wdoes-not-exist’
 $ echo $?
 1

 $ clang --target=aarch64-linux-gnu -Wdoes-not-exist -c -x c /dev/null
 warning: unknown warning option '-Wdoes-not-exist'; did you mean '-Wmicrosoft-exists'? [-Wunknown-warning-option]
 1 warning generated.
 $ echo $?
 0

Fix that by adding -Werror to the test command.


jenswi-linaro (commented):

PR #6035, opened 2023-05-18
core: riscv: Add more ISA into -march compiler option

gagachang:
Add c, zicsr and zifencei ISA extension strings into -march compiler option.


gagachang:
> It would be nice to say why this is needed in the commit description.

Hi @jforissier, I have a better solution for this.
I will create another PR, thanks!

PR #6036, opened 2023-05-19
RISC-V: Add Makefile variables for platform ISA, ABI, and code model

gagachang:
In RISC-V, each platform may have different supported ISA extensions, ABI, and code model.

In the first commit, we define the default variables of ISA extensions, ABI, and code model in RISC-V core Makefile `riscv.mk`.
In the second commit, we show that how the virt platform overrides the ISA in its `conf.mk`.

### DIFF
--- a/core/arch/riscv/riscv.mk
+++ b/core/arch/riscv/riscv.mk
@@ -71,9 +71,19 @@ core-platform-cppflags	+= -I$(arch-dir)/include
 core-platform-subdirs += \
 	$(addprefix $(arch-dir)/, kernel mm tee) $(platform-dir)
 
-# more convenient to move it to platform instead
-rv64-platform-cppflags += -mcmodel=medany -march=rv64imafd -mabi=lp64d
+# Default values for "-mcmodel", "-march", and "-abi" compiler flags.
+# Platform-specific overrides are in core/arch/riscv/plat-*/conf.mk.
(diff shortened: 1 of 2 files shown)


core/arch/riscv/riscv.mk
```diff
@@ -71,9 +71,20 @@ core-platform-cppflags	+= -I$(arch-dir)/include
...
+rv64-platform-abi ?= lp64d
+rv32-platform-isa ?= rv32imafd
+rv32-platform-abi ?= ilp32d
+
+rv64-platform-cppflags += -mcmodel=$(riscv-platform-mcmodel)
+rv64-platform-cppflags += -march=$(rv64-platform-isa) -mabi=$(rv64-platform-abi)
 rv64-platform-cppflags += -Wno-missing-include-dirs
+rv32-platform-cppflags += -mcmodel=$(riscv-platform-mcmodel)
+rv32-platform-cppflags += -march=$(rv32-platform-isa) -mabi=$(rv32-platform-abi)
+rv32-platform-cppflags += -Wno-missing-include-dirs
```
jenswi-linaro, line 87: Why is this needed?

gagachang: @jenswi-linaro Resolved in third commit.

jenswi-linaro:

gagachang:
Tag applied!

PR #6039, opened 2023-05-22
core: moves dt_driver_test from core/kernel/ to core/pta/test/

etienne-lms:


jforissier:
For commit "core: dt_driver_test: fix test failure status"

> Before this change, failing clock/reset/gpios tests printed their status

No need to use the past tense, it makes descriptions more complex. "Failing x/y/z tests print their status as..."

> as "not passed" because related status dt_test_state.probe_xxx remained

"not passed" should probably be "not run"? Should be modified in the source code in a separate patch (of course the commit description does not need to be changed since it is how the text was printed).

> to DEFAULT value and was never set to FAILED. Fix this and move each test
> to a specific local function to ease error cases management using 'goto'
> instructions.

With that:

jforissier:
Commit "core: dt_driver_test: move source file to core/pta/test/" is:

etienne-lms:
Forced update:
* added a commit to address s/"not passed"/"not run"/  review comment;
* rephrased commit message of commit "core: dt_driver_test: fix test failure status";
* applied @jforissier review tags to the 2 latter patches.

jforissier:
"core: dt_driver_test: clarify "not run" status message":

etienne-lms:
review tags applied

PR #6048, opened 2023-05-25
keys: rename default key

arnopo:
The default key can have some other uses than the TA signature. Rename the file to have a generic name.
Instead of specifying the use in the name, specifies the crypto algorithm.

This pull requested is related to [the remoteproc framework RFC](https://github.com/OP-TEE/optee_os/pull/6027) 
and especially discussion https://github.com/OP-TEE/optee_os/pull/6027#discussion_r1195089865

I identified impacts in 2 other repositories ( but probaly some other in CI):

- [optee_test](https://github.com/OP-TEE/optee_test.git)
  - [subkey_notes.txt](https://github.com/OP-TEE/optee_test/blob/master/ta/subkey_notes.txt)

- [optee_docs](https://github.com/OP-TEE/optee_docs)
  - [porting_guidelines.rst](https://github.com/OP-TEE/optee_docs/blob/master/architecture/porting_guidelines.rst) 
  - [trusted_applications.rst](https://github.com/OP-TEE/optee_docs/blob/master/building/trusted_applications.rst)
 
 I plan to create associated PRs after this one will be approved (if approved).


jenswi-linaro:
What do others think, is it worth the trouble to rename `keys/default_ta.pem`?
Do we need the name of the algorithm key name, wouldn't  `keys/default.pem` be enough?

jforissier:
> What do others think, is it worth the trouble to rename `keys/default_ta.pem`?

Maybe we could have a symlink to avoid breaking external projects.

> Do we need the name of the algorithm key name, wouldn't `keys/default.pem` be enough?

+1 for `default.pem`

jenswi-linaro:
> Maybe we could have a symlink to avoid breaking external projects.

Good idea.

etienne-lms:
I think it's worth renaming the file.
Ok for `default.pem` and a symlink for backward compat with external build environments.

arnopo:
@jenswi-linaro , @jforissier 

Please find update based on symbolic link solution.
Unfortunately it is not quite straightforward using current makefiles. So perhaps this will also fails the CI...
 
The issue seems coming from   [ta/ta.mk#]141(https://github.com/OP-TEE/optee_os/blob/master/ta/ta.mk#L141).
"cp -P" make a copy in the build directory  preserving the symlink.

Another issue found in ta/ta.mk  [ta/ta.mk#]186(https://github.com/OP-TEE/optee_os/blob/master/ta/ta.mk#L186).
Seems to me that we should have this fix in any case to  avoid to have the key file name  hard-coded. I can extract it in a separate PR on request.

jforissier:
> "cp -P" make a copy in the build directory preserving the symlink.

What's the problem? Then copy both the symlink and the new file, it should work.

```
ta-keys = keys/default_ta.pem keys/default.pem
```

arnopo:
> > "cp -P" make a copy in the build directory preserving the symlink.
> 
> What's the problem? Then copy both the symlink and the new file, it should work.
> 
> ```
> ta-keys = keys/default_ta.pem keys/default.pem
> ```

Yes that works if ta-keys is hardcoded in ta.mk

Before doing that, could  you please confirm that you are not interested in having
`ta-keys = $(TA_SIGN_KEY)` as I proposed in https://github.com/OP-TEE/optee_os/pull/6048/commits/65dabf2dd356e75d4798a5df336b25b2c1980402 ?

jforissier:
> > > "cp -P" make a copy in the build directory preserving the symlink.
> > 
> > 
> > What's the problem? Then copy both the symlink and the new file, it should work.
> > ```
> > ta-keys = keys/default_ta.pem keys/default.pem
> > ```
> 
> Yes that works if ta-keys is hardcoded in ta.mk
> 
> Before doing that, could you please confirm that you are not interested in having `ta-keys = $(TA_SIGN_KEY)` as I proposed in [65dabf2](https://github.com/OP-TEE/optee_os/commit/65dabf2dd356e75d4798a5df336b25b2c1980402) ?

Taking `TA_SIGN_KEY` into account is fine, but it is a different matter. How about:

```make
ta-keys := $(TA_SIGN_KEY)
ifeq ($(TA_SIGN_KEY),keys/default.pem)
# Copy symlink too for backwards compatibility
ta-keys += keys/default_ta.pem
endif
```

jenswi-linaro:
Please squash and fix up commits as needed.

arnopo:
> Please squash and fix up commits as needed.

Done

jenswi-linaro:

jforissier:

Preferably remove the upercase K in "Keys: rename the default key to default.pem" -- all subsystem prefixes are lowercase.

arnopo:
Reviewed-by added, and commit uppercase issue fixed. I also fixed another typo in same commit message ( s/become/becomes/)

Thank!

ta/ta.mk
```diff
@@ -184,6 +184,14 @@ $(foreach f, $(ta-srcfiles), \
 
 # Copy keys
 ta-keys := $(TA_SIGN_KEY)
+#
+# default_ta.pem is a symlink to default.pem, for backwards compatibility.
+# If default_ta.pem is used, copy both files.
+#
+ifeq ($(TA_SIGN_KEY),keys/default_ta.pem)
+	ta-keys += keys/default.pem
```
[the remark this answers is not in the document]
arnopo: bad reflex, fixed!

ta/ta.mk
```diff
@@ -184,6 +184,14 @@ $(foreach f, $(ta-srcfiles), \
 
 # Copy keys
 ta-keys := $(TA_SIGN_KEY)
+#
+# default_ta.pem is a symlink to default.pem, for backwards compatibility.
+# If default_ta.pem is used, copy both files.
+#
```
etienne-lms, line 190: prefer with empty comment lines removed, useless IMO.

arnopo: Done

etienne-lms (approved):

arnopo:
> `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`

Added in the commit message
Thanks!

PR #6055, opened 2023-05-26
drivers: stm32_rng: fix case when RNG is not ready

etienne-lms:
Checks RNG data ready status bit before each read of a 32bit sample from the RNG FIFO. Indeed the data ready status bit tells that the RNG FIFO contains random bytes by burst of 32bit word, not by burst of 4 32bit words.


jforissier (approved):

jenswi-linaro (commented):

etienne-lms:
Review tags applied.
Thanks

PR #6056, opened 2023-05-26
gpio framework: add a put operator

etienne-lms:
Adds a `.put` operation handler in `struct gpio_ops` for a GPIO consumer driver to be able to release the GPIO resources it has allocated when getting its GPIO using a DT drvier GPIO provider (`gpio_dt_alloc_pin()`). This is needed for example, when such a consumer driver must defer its probe sequence hence release resources as it will be probed again later and will request again its GPIO resources.

### DIFF
--- a/core/include/drivers/gpio.h
+++ b/core/include/drivers/gpio.h
@@ -55,16 +55,22 @@ struct gpio_chip {
 };
 
 struct gpio_ops {
+	/* Get GPIO direction current configuration */
 	enum gpio_dir (*get_direction)(struct gpio_chip *chip,
 				       unsigned int gpio_pin);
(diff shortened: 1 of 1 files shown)


etienne-lms:
@tprrt, could you have a look at this P-R?

tprrt:
Hello @etienne-lms,

It looks good to me.

Kind regards,
Thomas

etienne-lms:
@tprrt, do I understand I can apply you `Acked-by` review tag on both commits?

tprrt:
> @tprrt, do I understand I can apply you `Acked-by` review tag on both commits?

Yes, you can.

etienne-lms:
A-b review tag applied.

PR #6062, opened 2023-05-29
arm64: memtag: strip tag from crash dumps

jforissier:
The MTE tag is not really useful when displaying a crash dump. In fact it makes it more complicated for tools like script/symbolize.py to retrieve the actual (untagged) virtual address, especially as we now support non-Arm architectures and therefore blindly stripping the MTE bits is not possible.

This commit strips the tag in call stacks displayed by print_stack_arm64(). It also removes it from the virtual address shown on abort (__print_abort_info()) since symbolize.py does try to resolve the address as symbol + offset ; but in this case the tagged address is printed as well, because it can be helpful to diagnose tag check faults.

### DIFF
--- a/core/arch/arm/kernel/abort.c
+++ b/core/arch/arm/kernel/abort.c
@@ -11,6 +11,7 @@
 #include <kernel/tee_ta_manager.h>
 #include <kernel/thread_private.h>
 #include <kernel/user_mode_ctx.h>
+#include <memtag.h>
 #include <mm/core_mmu.h>
 #include <mm/mobj.h>
 #include <mm/tee_pager.h>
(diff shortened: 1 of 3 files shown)


core/arch/arm/kernel/abort.c
```diff
@@ -11,6 +11,8 @@
 #include <kernel/tee_ta_manager.h>
 #include <kernel/thread_private.h>
 #include <kernel/user_mode_ctx.h>
+#include <memtag.h>
+#include <memtag.h>
```
jenswi-linaro, line 15: Duplicated

jenswi-linaro (commented):

core/arch/arm/kernel/abort.c
jforissier: Fixed, thanks

etienne-lms (approved):

PR #6063, opened 2023-05-30
plat-stm32mp1: add missing inclusion of stm32_gpio.h

etienne-lms:
Adds inclusion of stm32_gpio.h from main.c that is needed when CFG_STM32MP13 is enabled.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/main.c
+++ b/core/arch/arm/plat-stm32mp1/main.c
@@ -9,6 +9,7 @@
 #include <console.h>
 #include <drivers/gic.h>
 #include <drivers/stm32_etzpc.h>
+#include <drivers/stm32_gpio.h>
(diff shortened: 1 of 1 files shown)


jforissier (approved):

Update CI?

etienne-lms:
CI already builds with with `PLATFORM=stm32mp1-135F_DK` which sets [`CFG_STM32MP13=y`](https://github.com/OP-TEE/optee_os/blob/3.21.0/core/arch/arm/plat-stm32mp1/conf.mk#L67-L69). I found this issue when porting _stm32_gpio.c_ to pinctrl framework where I removed inclusion of _stm32_gpio.h_ from _stm32_uart.h_ which is included by _main.c_ hence masking the issue.

etienne-lms:
review tag applied

PR #6067, opened 2023-05-31
stm32_gpio: bump to GPIO framework

etienne-lms:

### DIFF
--- a/core/drivers/stm32_gpio.c
+++ b/core/drivers/stm32_gpio.c
@@ -99,7 +99,9 @@ static enum gpio_level stm32_gpio_get_level(struct gpio_chip *chip,
 	unsigned int mode = 0;
 
 	assert(gpio_pin < bank->ngpios);
-	clk_enable(bank->clock);
+
+	if (clk_enable(bank->clock))
+		panic();
 
 	mode = (io_read32(bank->base + GPIO_MODER_OFFSET) >> (gpio_pin << 1)) &
 	       GPIO_MODE_MASK;
@@ -131,7 +133,9 @@ static void stm32_gpio_set_level(struct gpio_chip *chip, unsigned int gpio_pin,
 	struct stm32_gpio_bank *bank = gpio_chip_to_bank(chip);
 
 	assert(gpio_pin < bank->ngpios);
-	clk_enable(bank->clock);
+
+	if (clk_enable(bank->clock))
+		panic();
 
 	assert(((io_read32(bank->base + GPIO_MODER_OFFSET) >>
 		 (gpio_pin << 1)) & GPIO_MODE_MASK) == GPIO_MODE_OUTPUT);
@@ -151,7 +155,9 @@ static enum gpio_dir stm32_gpio_get_direction(struct gpio_chip *chip,
 	uint32_t mode = 0;
 
 	assert(gpio_pin < bank->ngpios);
-	clk_enable(bank->clock);
+
+	if (clk_enable(bank->clock))
+		panic();
 
 	mode = (io_read32(bank->base + GPIO_MODER_OFFSET) >> (gpio_pin << 1)) &
 	       GPIO_MODE_MASK;
@@ -183,7 +189,8 @@ static void stm32_gpio_set_direction(struct gpio_chip *chip,
 	else
 		mode = GPIO_MODE_OUTPUT;
 
-	clk_enable(bank->clock);
+	if (clk_enable(bank->clock))
+		panic();
 	exceptions = cpu_spin_lock_xsave(&gpio_lock);
 	io_clrsetbits32(bank->base + GPIO_MODER_OFFSET,
 			SHIFT_U32(GPIO_MODE_MASK, gpio_pin << 1),
--- a/core/drivers/stm32_gpio.c
+++ b/core/drivers/stm32_gpio.c
@@ -9,6 +9,7 @@
 #include <assert.h>
 #include <drivers/clk.h>
 #include <drivers/clk_dt.h>
+#include <drivers/gpio.h>
 #include <drivers/stm32_gpio.h>
 #include <io.h>
 #include <kernel/dt.h>
@@ -56,6 +57,7 @@
  *
  * @base: base address of the GPIO controller registers.
  * @clock: clock identifier.
+ * @gpio_chip: GPIO chip reference for that GPIO bank
  * @ngpios: number of GPIOs.
  * @bank_id: Id of the bank.
  * @lock: lock protecting the GPIO bank access.
@@ -66,6 +68,7 @@
 struct stm32_gpio_bank {
 	vaddr_t base;
 	struct clk *clock;
+	struct gpio_chip gpio_chip;
 	unsigned int ngpios;
 	unsigned int bank_id;
 	unsigned int lock;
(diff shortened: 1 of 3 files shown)


etienne-lms:
@ldebieve, @GseoC: could you have a look at these changes?

core/drivers/stm32_gpio.c
```diff
@@ -77,6 +80,132 @@ static unsigned int gpio_lock;
...
+static enum gpio_level stm32_gpio_get_level(struct gpio_chip *chip,
+					    unsigned int gpio_pin)
+{
+	struct stm32_gpio_bank *bank = gpio_chip_to_bank(chip);
+	enum gpio_level level = GPIO_LEVEL_HIGH;
+	unsigned int reg_offset = 0;
+	unsigned int mode = 0;
+
+	assert(gpio_pin < bank->ngpios);
+	clk_enable(bank->clock);
```
GseoC, line 100: Same question here on return code

core/drivers/stm32_gpio.c
```diff
@@ -77,6 +80,132 @@ static unsigned int gpio_lock;
...
+
+static enum gpio_level stm32_gpio_get_level(struct gpio_chip *chip,
+					    unsigned int gpio_pin)
+{
+	struct stm32_gpio_bank *bank = gpio_chip_to_bank(chip);
+	enum gpio_level level = GPIO_LEVEL_HIGH;
+	unsigned int reg_offset = 0;
+	unsigned int mode = 0;
+
+	assert(gpio_pin < bank->ngpios);
```
GseoC, line 99: gpio_pin is an argument that is provided by a framework user, therefore is not static like device tree configuration. assert will disappear when debug isn't enabled. Maybe a if/panic is better here.

(several occurences)

core/drivers/stm32_gpio.c
```diff
@@ -77,6 +80,132 @@ static unsigned int gpio_lock;
...
+	       GPIO_MODE_MASK;
+
+	switch (mode) {
+	case GPIO_MODE_INPUT:
+		reg_offset = GPIO_IDR_OFFSET;
+		break;
+	case GPIO_MODE_OUTPUT:
+		reg_offset = GPIO_ODR_OFFSET;
+		break;
+	default:
```
GseoC, line 112: Do we want to panic if we ask a level for these modes?

10: Alternate function mode
11: Analog mode

core/drivers/stm32_gpio.c
```diff
@@ -77,6 +80,132 @@ static unsigned int gpio_lock;
...
+}
+
+static void stm32_gpio_set_level(struct gpio_chip *chip, unsigned int gpio_pin,
+				 enum gpio_level level)
+{
+	struct stm32_gpio_bank *bank = gpio_chip_to_bank(chip);
+
+	assert(gpio_pin < bank->ngpios);
+	clk_enable(bank->clock);
+
```
GseoC, line 133: Is it necessary to check if the GPIO is in output mode?

core/drivers/stm32_gpio.c
```diff
@@ -77,6 +80,132 @@ static unsigned int gpio_lock;
...
+static enum gpio_dir stm32_gpio_get_direction(struct gpio_chip *chip,
+					      unsigned int gpio_pin)
+{
+	struct stm32_gpio_bank *bank = gpio_chip_to_bank(chip);
+	uint32_t mode = 0;
+
+	assert(gpio_pin < bank->ngpios);
+	clk_enable(bank->clock);
+
+	mode = (io_read32(bank->base + GPIO_MODER_OFFSET) >> (gpio_pin << 1)) &
```
GseoC, line 151: SHIFT_U32

core/drivers/stm32_gpio.c
```diff
@@ -77,6 +80,132 @@ static unsigned int gpio_lock;
...
+	clk_enable(bank->clock);
+	exceptions = cpu_spin_lock_xsave(&gpio_lock);
+	io_clrsetbits32(bank->base + GPIO_MODER_OFFSET,
+			SHIFT_U32(GPIO_MODE_MASK, gpio_pin << 1),
+			SHIFT_U32(mode, gpio_pin << 1));
+	cpu_spin_unlock_xrestore(&gpio_lock, exceptions);
+	clk_disable(bank->clock);
+}
+
+static bool is_stm32_gpio_chip(struct gpio_chip *chip);
```
GseoC, line 190: Prototype at the beginning of file?

core/drivers/stm32_gpio.c
```diff
@@ -328,6 +457,65 @@ static int get_pinctrl_from_fdt(void *fdt, int node,
...
+	io_clrsetbits32(bank->base + GPIO_PUPDR_OFFSET,
+			SHIFT_U32(GPIO_PUPD_PULL_MASK, shift_2b),
+			SHIFT_U32(pupd, shift_2b));
+
+	cpu_spin_unlock_xrestore(&gpio_lock, exceptions);
+
+	gpio->chip = &bank->gpio_chip;
+
+	*res = TEE_SUCCESS;
+	return gpio;
```
GseoC, line 519: Miss empty line before return

core/drivers/stm32_gpio.c
etienne-lms: which framework do you foresee on manipulating GPIOs?
A consumer driver should not get/set a GPIO level if it did not get the GPIO first. So I think the assertion above is even overkilling.

core/drivers/stm32_gpio.c
etienne-lms: yes, because there is a conflict between `struct gpio_ops stm32_gpio_ops` and `is_stm32_gpio_chip()` definition, both depend on the other.

core/drivers/stm32_gpio.c
etienne-lms: `gpio_pin` is an unsigned int, not a u32.

core/drivers/stm32_gpio.c
etienne-lms: make sense.

core/drivers/stm32_gpio.c
etienne-lms: Same as above: a consumer should not get/set a GPIO level if it did not get the GPIO first.

core/drivers/stm32_gpio.c
etienne-lms: I'll make the function panicking, and consider `TEE_Result` return code for GPIO operation handlers in another P-R.

core/drivers/stm32_gpio.c
etienne-lms: ok, i'll add one.

etienne-lms:
comments addressed

core/drivers/stm32_gpio.c
GseoC: Maybe when piloting a GPIO regulator?

Whatever the response, I think your statement : _A consumer driver should not get/set a GPIO level if it did not get the GPIO first. So I think the assertion above is even overkilling._ is valid. It can even be removed

GseoC:
I'm ok with the changes: `Acked-by: Gatien Chevallier <gatien.chevallier@foss.st.com>`

PR #6069, opened 2023-05-31
FF-A v1.1 boot protocol for SPs

balint-dobszay-arm:
Implement the FF-A v1.1 boot protocol for passing boot info to Secure Partitions.

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -1289,6 +1289,7 @@ static TEE_Result sp_first_run(struct sp_session *sess)
 
 	ctx = to_sp_ctx(sess->ts_sess.ctx);
 	ts_push_current_session(&sess->ts_sess);
+	sess->is_initialized = false;
 
 	/*
 	 * Load relative memory regions must be handled before doing any other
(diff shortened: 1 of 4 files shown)


balint-dobszay-arm:
> Please add a "Fixes:" tag to "core: sp: fix unmap in sp_first_run"

Done

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1322,11 +1408,17 @@ static TEE_Result sp_first_run(struct sp_session *sess)
...
+				       &boot_info_size);
+	if (res)
+		goto out;
+
 	ts_pop_current_session();
 
 	sess->is_initialized = false;
 	if (sp_enter(&args, sess)) {
-		vm_unmap(&ctx->uctx, va, sz);
+		vm_unmap(&ctx->uctx, boot_info_va, boot_info_size);
```
jenswi-linaro, line 1420: When calling these functions, the current session should be `&sess->ts_sess`.
Perhaps `sp_enter()` could be refactored so we have a special version of the function to call above where the current session is already set. This refactoring should perhaps be done before this patch since it's a bug fix.

balint-dobszay-arm: I see your point, however I'm not sure what you mean by the refactoring. Removing the `ts_pop_current_session()` call above, and moving it to after these calls would also solve the issue, wouldn't it?

balint-dobszay-arm: Okay I realised we have `sp_enter()`->`sp_enter_invoke_cmd()`->`ts_push_current_session()`, so we do have to keep the `ts_pop_current_session()` before, my mistake.

How about just surrounding these two `vm_unmap()` calls with `ts_push_current_session()` and `ts_pop_current_session()`? Similarly to how it's done at the end of the function.

balint-dobszay-arm: I think we should call `spmc_sp_msg_handler()` without the session activated? So the `ts_push_current_session()` right after `sp_enter()` seems wrong to me. I pushed a patch that's similar to your proposal, let me know what you think. It also addresses the FF-A error code vs `TEE_Result` issue.

jenswi-linaro:

balint-dobszay-arm:
Updated, rebased on master. The PR now includes another bugfix (`core: sp: fix session handling in sp_first_run`) as suggested by Jens in the thread above.

jenswi-linaro:

balint-dobszay-arm:
Thanks, tag applied.

PR #6074, opened 2023-06-01
core: io: add access with memory barrier functions

patrickdelaunay:
Add IO functions with memory barrier to be used by driver when order of accesses need to be enforced (see readl() & writel() / readl_relaxed() & writel_relaxed() functions in Linux file io.h).

Change-Id: I25c8713ba66fd3d5b1d15750b9752c08254856ed

### DIFF
--- a/core/include/io.h
+++ b/core/include/io.h
@@ -20,6 +20,25 @@
 #define READ_ONCE(p)		__compiler_atomic_load(&(p))
 #define WRITE_ONCE(p, v)	__compiler_atomic_store(&(p), (v))
 
+/*
+ *  IO port access primitives
+ *  -------------------------
+ *
+ * Arm A-profile cores don't have special IO access instructions.
+ * OP-TEE maps IO memory as device, shareable and bufferable but not cacheable.
+ *
+ * io_readX(), io_writeX() io_clr/setbitsX() functions do not guarantee
+ * that the IO request reached the device in the code order.
+ *
+ * The barrier version of each function - io_readX_barrier(), io_writeX_barier()
+ * and io_clr/setbitsX_barrier() - guarantee that IO access are
+ * observable for system (outer shareable):
+ * - Make sure the write above is visible
+ * - Make sure the read is performed
+ */
+#define __rmb()		dsb()
+#define __wmb()		dsb()
+
 static inline void io_write8(vaddr_t addr, uint8_t val)
 {
 	*(volatile uint8_t *)addr = val;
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):
LGTM. 
Please remove the Change-Id tag **~~adn~~ and the** reference to Linux io.h function in commit message and you can add my R-b tag.

**(edited)**

patrickdelaunay:
Ok, done

core/include/io.h
```diff
@@ -20,6 +20,25 @@
...
+ * io_readX(), io_writeX() io_clr/setbitsX() functions do not guarantee
+ * that the IO request reached the device in the code order.
+ *
+ * The barrier version of each function - io_readX_barrier(), io_writeX_barier()
+ * and io_clr/setbitsX_barrier() - guarantee that IO access are
+ * observable for system (outer shareable):
+ * - Make sure the write above is visible
+ * - Make sure the read is performed
+ */
+#define __rmb()		dsb()
```
jenswi-linaro, line 39: `core/include/io.h` is an architecture common file so we should at the very least surround this with ifdefs. Then we'd also have to do the same with an `#include <arm.h>` above so perhaps it's better to put this in `core/arch/arm/include/io_arch.h` or such.

I'd prefer a more descriptive name for these barriers, such as `io_before_write_barrier()`, `io_after_write_barrier()`, and so on.

Shouldn't this be a `dsb_oshld` and the write barrier `dsb_oshst`?

core/include/io.h
```diff
@@ -40,6 +59,11 @@ static inline void io_write64(vaddr_t addr, uint64_t val)
 	*(volatile uint64_t *)addr = val;
 }
 
+#define io_write8_barrier(v, c)		({ __wmb(); io_write8(v, c); })
```
jenswi-linaro, line 62: I guess it might be a matter of taste, but in these situations, I'd prefer using static inline functions instead since they are specific with regard to types of arguments and return value. Same for the other function like macros below.

core/include/io.h
```diff
@@ -75,6 +105,10 @@ static inline void io_mask32(vaddr_t addr, uint32_t val, uint32_t mask)
 	io_write32(addr, (io_read32(addr) & ~mask) | (val & mask));
 }
 
+#define io_mask8_barrier(v, c, m)	({ __wmb(); io_mask8(v, c, m); })
```
jenswi-linaro, line 108: `io_mask8()` does both read and write, is it enough with a write barrier like this?

etienne-lms:
keep alive

PR #6080, opened 2023-06-05
mk: Support user static lib for aosp build

taddk:
For AOSP environment to build user static libraries, there are two issues we had met,

1. the static library is default writing to the current source folder which is not allowed by AOSP build environment.
2. the static library is not necessarily installed to vendor lib folder like a TA or a shared library.

Propose to use `$(out-link-dir$(sm))` (passed by `$O`) to place the generated user static library, and change LOCAL_MODULE_CLASS to STATIC_LIBRARIES for it in AOSP build to avoid unnecessarily installation.

### DIFF
--- a/mk/aosp_optee.mk
+++ b/mk/aosp_optee.mk
@@ -106,7 +106,11 @@ include $(CLEAR_VARS)
 LOCAL_MODULE := $(local_module)
 LOCAL_PREBUILT_MODULE_FILE := $(OPTEE_TA_OUT_DIR)/$(LOCAL_MODULE)
 LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/lib/optee_armtz
+ifneq ($(filter %.ta, $(local_module)),)
 LOCAL_MODULE_CLASS := EXECUTABLES
+else
+LOCAL_MODULE_CLASS := STATIC_LIBRARIES
+endif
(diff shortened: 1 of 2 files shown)


mk/aosp_optee.mk
```diff
@@ -106,7 +106,11 @@ include $(CLEAR_VARS)
 LOCAL_MODULE := $(local_module)
 LOCAL_PREBUILT_MODULE_FILE := $(OPTEE_TA_OUT_DIR)/$(LOCAL_MODULE)
 LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/lib/optee_armtz
+ifneq ($(filter %.ta, $(local_module)),)
```
jenswi-linaro, line 109: I'm trusting that you're doing the right thing here.
AOSP support is more or less unmaintained so if this works for you...

jenswi-linaro (commented):

mk/aosp_optee.mk
taddk: I hope so. Since TAs and shared libraries recognized by tee-supplicant are with file extension ".ta", this should generally work for most cases. We did test by building optee_test suite.

Or maybe we can add an optional variable, `local_module_class`, for the module to specify its class. That could make it more straightforward.

PR #6081, opened 2023-06-05
drivers: pinctrl: pinctrl name may not be found

etienne-lms:
Changes pinctrl_get_state_by_name() to return TEE_ERROR_ITEM_NOT_FOUND instead of TEE_ERROR_GENERIC when an pinctrl state index is not found in the DT.

Fixes: 9aec039ec0d7 ("drivers: pinctrl: add pinctrl support")


GseoC:

PR #6098, opened 2023-06-08
core: dt_driver: swap TEE_result and retrieved device reference

etienne-lms:
Changes dt_driver callback function to return a `TEE_Result` value and pass retrieved device reference by a output argument rather than the opposite.

This change updates _dt_driver.c_, _dt_driver.h_, _dt_driver_test.c_ and all drivers implementing related dt_driver callback functions.

As a consequence, this change removes all type definition related to device specific callback handler function types which are useless as all these now comply with type `dt_driver_probe_func` defined in _dt_driver.h_.

This P-R is related to discussion thread https://github.com/OP-TEE/optee_os/pull/5954#discussion_r1211837322.

### DIFF
--- a/core/drivers/pinctrl/pinctrl.c
+++ b/core/drivers/pinctrl/pinctrl.c
@@ -96,15 +96,18 @@ TEE_Result pinctrl_get_state_by_idx(const void *fdt, int nodeoffset,
 
 	state->conf_count = conf_count;
 	for (conf_id = 0; conf_id < conf_count; conf_id++) {
-		state->confs[conf_id] =
-			dt_driver_device_from_node_idx_prop(prop_name, fdt,
-							    nodeoffset, conf_id,
-							    DT_DRIVER_PINCTRL,
-							    &res);
+		void *pinconf = NULL;
+
+		res = dt_driver_device_from_node_idx_prop(prop_name, fdt,
+							  nodeoffset, conf_id,
+							  DT_DRIVER_PINCTRL,
+							  &pinconf);
 		if (res) {
 			free(state);
 			return res;
 		}
+
+		state->confs[conf_id] = pinconf;
 	}
 
 	*state_ret = state;
--- a/core/drivers/atmel_piobu.c
+++ b/core/drivers/atmel_piobu.c
@@ -204,24 +204,26 @@ static const struct gpio_ops atmel_piobu_ops = {
 	.set_interrupt = secumod_gpio_set_interrupt,
 };
 
-static struct gpio *secumod_dt_get(struct dt_pargs *pargs, void *data,
-				   TEE_Result *res)
+static TEE_Result secumod_dt_get(struct dt_pargs *pargs, void *data,
+				 struct gpio **out_gpio)
 {
+	TEE_Result res = TEE_ERROR_GENERIC;
 	struct gpio *gpio = NULL;
 	struct gpio_chip *chip = data;
 
-	gpio = gpio_dt_alloc_pin(pargs, res);
-	if (*res)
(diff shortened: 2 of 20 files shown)


etienne-lms:
Checkpatch warns on `nodeoffset` not being a identifiable name. IMO it's a false positive. This change series is not expected to rename those argument names.
```
WARNING: function definition argument 'int nodeoffset' should also have an identifier name
#944: FILE: core/include/kernel/dt_driver.h:172:
+TEE_Result dt_driver_device_from_parent(const void *fdt, int nodeoffset,
```

etienne-lms:
This P-R is related to discussion thread https://github.com/OP-TEE/optee_os/pull/5954#discussion_r1211837322.

etienne-lms:
Bug found in _dt_driver_test.c_ and fixed. CI make check should run ok now.

etienne-lms:
I pushed 2 fixup. A bug introduced in `fixed_clock_probe()` and a big fixup commit to restore per driver framework callback functions (clk, reset controller, gpio, pinctrl, i2c bus) to not have consumer callback using `void **` device reference types but rather driver specific type (`struct clk **`, `struct rstctrl **` etc...) as before this change series.
Sorry for the incoviences, i think i would be easier for the reviews if I could squash straight those fixup commits.

core/drivers/pinctrl/pinctrl.c
```diff
@@ -96,11 +96,12 @@ TEE_Result pinctrl_get_state_by_idx(const void *fdt, int nodeoffset,
...
-			dt_driver_device_from_node_idx_prop(prop_name, fdt,
-							    nodeoffset, conf_id,
-							    DT_DRIVER_PINCTRL,
-							    &res);
+		struct pinconf **pinconf = state->confs + conf_id;
+
+		res = dt_driver_device_from_node_idx_prop(prop_name, fdt,
+							  nodeoffset, conf_id,
+							  DT_DRIVER_PINCTRL,
+							  (void **)pinconf);
```
jenswi-linaro, line 104: This cast might be a problem, depending on the compiler. It may not be a problem in practice since all pointers use the same size, but a zealous compiler could complain about strict aliasing rules. I guess this is the reason why this somewhat backward way of returning the error value was chosen.

I'd prefer
``` C
                void *pinconf = NULL;

                res = dt_driver_device_from_node_idx_prop(prop_name, fdt,
                                                          nodeoffset, conf_id,
                                                          DT_DRIVER_PINCTRL,
                                                          &pinconf);
                 if (res) {
                        free(state);
                        return res;
                 }
                 state->confs[conf_id] = pinconf;
```
to avoid that ugly cast.

jenswi-linaro (commented):
What do others think? Is it worth it to swap the TEE_Result and device reference?

jforissier:
> What do others think? Is it worth it to swap the TEE_Result and device reference?

No strong opinion, but judging by the diffstat (+269 -334) it makes the code a bit more compact, and having a TEE_Result as a return value is indeed much more natural. So I'd say yes, it's worth it.

> I'd prefer
> ```C
>                void *pinconf = NULL;
>
>                res = dt_driver_device_from_node_idx_prop(prop_name, fdt,
>                                                          nodeoffset, conf_id,
>                                                          DT_DRIVER_PINCTRL,
>                                                          &pinconf);
> ```

+1

jenswi-linaro:
Thanks @jforissier.
@etienne-lms, please address my comment and squash the commits as needed. I'll review the stuff once you're done.

jenswi-linaro:

etienne-lms:
review tag applied

PR #6100, opened 2023-06-08
core: riscv: Fix misuse of cppflags

gagachang:
The -mxxx and -Wxxx are not preprocessor flags. Fix it by defining them as C flags.


gagachang:
Tag applied!

Issue #6099, opened 2023-06-08
[RFC] Rename "mva" to "va"


gagachang:
The terminology "mva" is specific for older ARM architecture which has FCSE extension [1].
To support multiple architecture, it would be good to rename "mva" to common terminology, such as "va".

Suggest renaming related functions in `core/mm/vm.c` and `core/include/mm/core_mmu.h`:
As is:
```
tlbi_mva_range()
tlbi_mva_range_asid()
```
To be:
```
tlbi_va_range()
tlbi_va_range_asid()
```

[1] https://developer.arm.com/documentation/ddi0406/c/Appendices/Fast-Context-Switch-Extension--FCSE-

etienne-lms:
I'm fine with such a change. As Arm refer the mva for these instruction, the Arm implementation could wrap `tlbi_va_range()` to a call to arch specific `tlbi_mva_range()` and preserve its existing tlb_helpers functions implementation that is consistent with the architectures reference manuals.

gagachang:
Dear @etienne-lms 

I have a local branch with commits which apply your suggestion. 
https://github.com/gagachang/optee_os/tree/rename-mva

Please take a look, thanks!

jenswi-linaro:
Please propose this change in a pull request. It's much easier to review it in that format.

gagachang:
Hi @jenswi-linaro 

Here is the proposed PR: https://github.com/OP-TEE/optee_os/pull/6101

gagachang:
Since https://github.com/OP-TEE/optee_os/pull/6101 has already been merged, I close this ticket.

PR #6101, opened 2023-06-09
[RFC] Rename "mva" to "va"

gagachang:
This PR simply applies the discussion in https://github.com/OP-TEE/optee_os/issues/6099

### DIFF
--- a/core/include/mm/core_mmu.h
+++ b/core/include/mm/core_mmu.h
@@ -567,18 +567,18 @@ struct tee_mmap_region *
 core_mmu_find_mapping_exclusive(enum teecore_memtypes type, size_t len);
 
 /*
- * tlbi_mva_range() - Invalidate TLB for virtual address range
+ * tlbi_va_range() - Invalidate TLB for virtual address range
  * @va:		start virtual address, must be a multiple of @granule
  * @len:	length in bytes of range, must be a multiple of @granule
  * @granule:	granularity of mapping, supported values are
  *		CORE_MMU_PGDIR_SIZE or SMALL_PAGE_SIZE. This value must
  *		match the actual mappings.
  */
-void tlbi_mva_range(vaddr_t va, size_t len, size_t granule);
+void tlbi_va_range(vaddr_t va, size_t len, size_t granule);
 
 /*
- * tlbi_mva_range_asid() - Invalidate TLB for virtual address range for
- *			   a specific ASID
+ * tlbi_va_range_asid() - Invalidate TLB for virtual address range for
+ *			  a specific ASID
  * @va:		start virtual address, must be a multiple of @granule
  * @len:	length in bytes of range, must be a multiple of @granule
  * @granule:	granularity of mapping, supported values are
@@ -586,7 +586,7 @@ void tlbi_mva_range(vaddr_t va, size_t len, size_t granule);
  *		match the actual mappings.
  * @asid:	Address space identifier
  */
-void tlbi_mva_range_asid(vaddr_t va, size_t len, size_t granule, uint32_t asid);
+void tlbi_va_range_asid(vaddr_t va, size_t len, size_t granule, uint32_t asid);
 
 /* Check cpu mmu enabled or not */
 bool cpu_mmu_enabled(void);
--- a/core/arch/arm/include/arm64.h
+++ b/core/arch/arm/include/arm64.h
@@ -224,7 +224,7 @@
 #define PAR_PA_SHIFT		U(12)
 #define PAR_PA_MASK		(BIT64(36) - 1)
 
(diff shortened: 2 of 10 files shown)


core/arch/arm/include/mm/core_mmu_arch.h
```diff
@@ -210,6 +210,11 @@ static inline unsigned int core_mmu_get_va_width(void)
 	}
 	return 32;
 }
+
+void tlbi_mva_range(vaddr_t va, size_t len, size_t granule);
```
etienne-lms, line 214: Prefer static inline functions:

```
static inline void tlbi_va_range_asid(vaddr_t va, size_t len, size_t granule, uint32_t asid)
{
	tlbi_mva_range_asid(va, len, granule, asid);
}

static inline void tlbi_va_range(vaddr_t va, size_t len, size_t granule)
{
	tlbi_mva_range(va, len, granule);
}
```

Since core_mmu.h does include _mm/core_mmu_arch.h_, everything will be in place for callers.

etienne-lms (commented):
please either squash to 2 commits  into a signle one or swap their order in the series, otherwise build fails on the former.

gagachang:
> CI test error:
> 
> ```
> core/include/mm/core_mmu.h:589:6: error: redundant redeclaration of 'tlbi_va_range_asid' [-Werror=redundant-decls]
>   589 | void tlbi_va_range_asid(vaddr_t va, size_t len, size_t granule, uint32_t asid);
>       |      ^~~~~~~~~~~~~~~~~~
> (...)
> core/arch/arm/include/mm/core_mmu_arch.h:223:20: note: previous definition of 'tlbi_va_range_asid' with type 'void(vaddr_t,  size_t,  size_t,  uint32_t)' {aka 'void(long unsigned int,  unsigned int,  unsigned int,  unsigned int)'}
>   223 | static inline void tlbi_va_range_asid(vaddr_t va, size_t len, size_t granule,
>       |                    ^~~~~~~~~~~~~~~~~~
> ```
> 
> Sorry my suggestion of using inline function was stupid 😕.

Inline should be better to reduce function call cost.
Maybe they should added into `core_mmu.h` ?

```
#if defined (ARM32) || defined(ARM64)
static inline void tlbi_va_range_asid(vaddr_t va, size_t len, size_t granule, uint32_t asid)
{
	tlbi_mva_range_asid(va, len, granule, asid);
}

static inline void tlbi_va_range(vaddr_t va, size_t len, size_t granule)
{
	tlbi_mva_range(va, len, granule);
}
#else ...

#endif
```

jforissier:
> [RFC] Rename "mva" to "va"

Why not just do exactly that? (in `core/arch/arm/mm/core_mmu.c` as well)

gagachang:
> > [RFC] Rename "mva" to "va"
> 
> Why not just do exactly that? (in `core/arch/arm/mm/core_mmu.c` as well)

Both are OK to me.
@etienne-lms how about this ?

etienne-lms (commented):
all good to me

core/include/mm/core_mmu.h
```diff
@@ -567,26 +567,41 @@ struct tee_mmap_region *
...
  * @len:	length in bytes of range, must be a multiple of @granule
  * @granule:	granularity of mapping, supported values are
  *		CORE_MMU_PGDIR_SIZE or SMALL_PAGE_SIZE. This value must
  *		match the actual mappings.
  */
-void tlbi_mva_range(vaddr_t va, size_t len, size_t granule);
+#if defined(ARM32) || defined(ARM64)
+static inline void tlbi_va_range(vaddr_t va, size_t len, size_t granule)
+{
+	tlbi_mva_range(va, len, granule);
```
jforissier, line 580: I meant: why not get rid of `tlbi_mva_range()`?

Even so, this PR leaves `_mva` in many places (including in `core/arch/riscv/mm/tlb_helpers_rv.S`):
```
$ git grep _mva | wc -l
38
```
If `va` is not a good term to begin with, even on modern Arm, I think that *all* occurrences should be replaced by `va`.

gagachang: OK, I can submit more commits to replace `mva` with `va`.

gagachang: Hi @jforissier 

I made a commit for this, please check it.
Note that some "mva" are still reserved for ARM32.

jenswi-linaro:
How about doing it in a single commit?

gagachang:
@jforissier @jenswi-linaro
Thank you!

Dear @etienne-lms 
I remove your tag because the commit is different with original one.
Please feel free to review it again.

etienne-lms (approved):

PR #6102, opened 2023-06-09
mk/aosp_optee.mk: fix build dependency for static libraries

taddk:
After skipping installation for static libraries by commit  667213e21e8b ("mk: Support user static lib for aosp build").  The dependency set by $(local_module_deps) needs different file paths for static libraries. Use aosp macro to build the dependency with the intermediate file instead of installed file.

The dependency change also affects user TAs and shared libraries. Since installed location could be changed for some customization requirement, I think using the intermediate-dir should make it easier if we would like to change the `$(LOCAL_MODULE_PATH)` for TAs.

### DIFF
--- a/mk/aosp_optee.mk
+++ b/mk/aosp_optee.mk
@@ -112,6 +112,7 @@ else
 LOCAL_MODULE_CLASS := STATIC_LIBRARIES
 endif
 LOCAL_MODULE_TAGS := optional
+LOCAL_REQUIRED_MODULES := $(local_module_deps)
 
 TA_TMP_DIR := $(subst /,_,$(LOCAL_PATH))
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Would it make sense to add a 
```
Fixes: d1b003febbcd ("mk: Support user static lib for aosp build")
```
tag to the commit message?

taddk:
Sounds a good idea. I will update commit message to add the tag and remove the wrong commit sha1 mentioned. Thanks for the suggestion.

taddk:
I found that the previous change doesn't install the shared libraries when `local_module_deps` is set. It is a problem due to the dependency is not set to the installed file in the previous change. So, the latest patch sets the module dependency with `LOCAL_REQUIRED_MODULES` provided by AOSP build-system is proposed.

taddk:
Thanks, updated the `Link:` tag in commit message.

jenswi-linaro:

Issue #5725, opened 2022-12-21
Mapping user VA space and understanding the use of CORE_MMU_PGDIR_LEVEL and CORE_MMU_PGDIR_SIZE


maroueneboubakri:
Hi @jenswi-linaro, @jforissier,

I was looking at `vm.c` and `pgt_cache.c` I'm a bit confused about the usage of `CORE_MMU_PGDIR_LEVEL` and `CORE_MMU_PGDIR_SIZE`.
My use case is simple, I want to decide at which level I want to map the user-space. I want to map it using 2M pages. I'm not sure if this could be done easily ? as there is strong bound between `CORE_MMU_PGDIR_LEVEL` ,`CORE_MMU_PGDIR_SIZE` and memory regions. 

These do not have sens in RISC-V as there is no PGDIRs. Not sure if this could replaced with more generic definitions likes mega pages, giga pages in addition to small pages ? When working with memory regions and page tables ? 
I noticed that despite changing value of CORE_MMU_PGDIR_LEVEL, setting VA base to 2M aligned address in `core_mmu_get_user_va_range()`:
`ldelf_loader.c` uses number of pages of `SMALL_PAGE_SIZE `?
`mobj_get_pa()` returns always 4K aligned PAs ? 
`select_va_in_range()` returns 4K aligned adresses or padded with 4K pages.

Short question is, what is the appropriate way to map user VA space using giga pages or mega pages. In ARM is it enough to replace `L1_XLAT_ADDRESS_SHIFT` with `L0_XLAT_ADDRESS_SHIFT ` for example ? 

Appreciate your help.
Maro

maroueneboubakri:
Also here, what `(pgdir)` means ?

```
D/TC:?   add_phys_mem:635 MU_BASE type IO_NSEC 0x88000000 size 0x00008000
D/TC:?   add_phys_mem:635 ROUNDDOWN(0x0c000000, CORE_MMU_PGDIR_SIZE) type IO_SEC 0x0c000000 size 0x00200000
D/TC:?   add_phys_mem:635 ROUNDDOWN(0x10000000, CORE_MMU_PGDIR_SIZE) type IO_NSEC 0x10000000 size 0x00200000
D/TC:?   add_phys_mem:635 TEE_SHMEM_START type NSEC_SHM 0x88f00000 size 0x00200000
D/TC:?   add_phys_mem:635 TA_RAM_START type TA_RAM 0x80500000 size 0x00b00000
[...]

PR #6104, opened 2023-06-09
riscv: mm: initial implementation of memory management routines for RISC-V

maroueneboubakri:
Happy to see, that there is an intention to make slight changes to the current OP-TEE OS's memory management layer, to make it easier for RISC-V MMU driver integration, I re-open this PR to check progressively what we can adapt to have this big chunk merged.

PR #6105, opened 2023-06-11
core: mm: Refine architecture specific translation level

gagachang:
This PR is initiated from the discussion in https://github.com/OP-TEE/optee_os/issues/5725

To decouple the architecture specific address translation rule from core mm.
The `next_level` filed is added in the `struct core_mmu_table_info` to represent the next finer grained level.

Some ARM is also modified based on this change.

### DIFF
--- a/core/arch/arm/include/mm/core_mmu_arch.h
+++ b/core/arch/arm/include/mm/core_mmu_arch.h
@@ -210,6 +210,17 @@ static inline unsigned int core_mmu_get_va_width(void)
 	}
 	return 32;
 }
+
+static inline bool core_mmu_level_in_range(unsigned int level)
+{
+#if CORE_MMU_BASE_TABLE_LEVEL == 0
+	return level <= CORE_MMU_PGDIR_LEVEL;
+#else
+	return level >= CORE_MMU_BASE_TABLE_LEVEL &&
+	       level <= CORE_MMU_PGDIR_LEVEL;
+#endif
+}
+
 #endif /*__ASSEMBLER__*/
 
 #endif /* CORE_MMU_H */
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -948,6 +948,7 @@ void core_mmu_set_info_table(struct core_mmu_table_info *tbl_info,
 		unsigned level, vaddr_t va_base, void *table)
 {
 	tbl_info->level = level;
+	tbl_info->next_level = level + 1;
 	tbl_info->table = table;
 	tbl_info->va_base = va_base;
 	tbl_info->shift = XLAT_ADDR_SHIFT(level);
@@ -1028,6 +1029,7 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
 			tbl_info->table = tbl;
 			tbl_info->va_base = va_base;
 			tbl_info->level = level;
+			tbl_info->next_level = level + 1;
 			tbl_info->shift = level_size_shift;
 			tbl_info->num_entries = num_entries;
 #ifdef CFG_NS_VIRTUALIZATION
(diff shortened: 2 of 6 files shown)


core/arch/arm/include/mm/core_mmu_arch.h
```diff
@@ -210,6 +210,12 @@ static inline unsigned int core_mmu_get_va_width(void)
 	}
 	return 32;
 }
+
+static inline bool core_mmu_table_level_not_exceed(unsigned int level)
```
etienne-lms, line 214: replicate this in _core/arch/riscv/include/mm/core_mmu_arch.h_

etienne-lms (commented):
for commit "core: mm: Reduce size of struct core_mmu_table_info"
and commit "core: mm: Introduce next_level field of struct core_mmu_table_info".

For commit "core: mm: Introduce core_mmu_table_level_not_exceed()":
In commit message: s/rule/rules/
The riscv part is missing, it will not build. the code LGTM.

core/arch/arm/include/mm/core_mmu_arch.h
gagachang: Hi @etienne-lms 

RISC-V uses decreasing style in address translation level, which is different with ARM.
And since new RISC-V mm implementation https://github.com/OP-TEE/optee_os/pull/6104 is not ready to be merged, I hope RISC-V's version can be implemented in another PR. Some table level definitions are changed in #6104.

etienne-lms: riscv builds _boot.c_ that now requires `core_mmu_table_level_not_exceed()`.

gagachang: I add a new commit for RISC-V's `core_mmu_table_level_not_exceed()`.
Please check it.

etienne-lms (commented):
Squash "[Review] Add RISC-V's core_mmu_table_level_not_exceed()"
into "core: mm: Introduce core_mmu_table_level_not_exceed()"
and you can add my R-b for that commit.

gagachang:
@etienne-lms 

Tag applied! Thanks for reviews.

etienne-lms (approved):

core/arch/arm/include/mm/core_mmu_arch.h
jenswi-linaro, line 214: How about making this:
``` C
static inline bool core_mmu_table_level_in_range(unsigned int level)
{
        return level >= CORE_MMU_BASE_TABLE_LEVEL &&
               level <= CORE_MMU_PGDIR_LEVEL;
}
```
With the advantage that we avoid confusion about which side we're checking in the interval.

gagachang: It makes sense.
But if `CORE_MMU_BASE_TABLE_LEVEL` equals 0, compiler will generate the warning:
```
warning: comparison of unsigned expression in ‘>= 0’ is always true [-Wtype-limits]
```
What about this version:
```
static inline bool core_mmu_table_level_in_range(unsigned int level)
{
#if CORE_MMU_BASE_TABLE_LEVEL == 0
	return level <= CORE_MMU_PGDIR_LEVEL;
#else
	return level >= CORE_MMU_BASE_TABLE_LEVEL &&
	       level <= CORE_MMU_PGDIR_LEVEL;
#endif
}
```

jenswi-linaro: OK.

gagachang: Resolved in new commit.

core/arch/arm/include/mm/core_mmu_arch.h
```diff
@@ -210,6 +210,17 @@ static inline unsigned int core_mmu_get_va_width(void)
 	}
 	return 32;
 }
+
+static inline bool core_mmu_table_level_in_range(unsigned int level)
```
jenswi-linaro, line 214: I'm sorry for picking on this again, but I'd prefer to use the name `core_mmu_level_in_range()`.

jenswi-linaro (commented):

core/arch/arm/include/mm/core_mmu_arch.h
gagachang: That's OK~
Changes and tags are included in new commits.

jenswi-linaro:
Looks good.

PR #6115, opened 2023-06-15
plat-stm32mp1: restore SYSRAM for SCMI message on STM32MP13

etienne-lms:
Restores use of SYSRAM last page for STM32MP13 for SCMI communication as U-Boot and Linux kernel device trees are not yet updated to use OP-TEE native shared memory instead.

Fixes: 89ba3422ee80 ("plat-stm32mp1: scmi_server: default use OP-TEE shared memory")


GseoC:

etienne-lms:
Tag applied.
Thanks.

etienne-lms:
thanks @jforissier.

PR #6116, opened 2023-06-15
Bounce buffers for syscall processing

jenswi-linaro:
This relates to https://github.com/OP-TEE/optee_os/pull/6061 and tries to solve the problem with unpredictable memory allocation for bounce buffers. There's some example usage in the commit "core: use user-access functions in ldelf interaction".

### DIFF
--- a/core/include/kernel/user_access.h
+++ b/core/include/kernel/user_access.h
@@ -7,6 +7,7 @@
 #define __KERNEL_USER_ACCESS_H
 
 #include <assert.h>
+#include <kernel/user_access_arch.h>
 #include <tee_api_types.h>
 #include <types_ext.h>
 
--- a/core/include/kernel/user_access.h
+++ b/core/include/kernel/user_access.h
@@ -95,6 +95,24 @@ TEE_Result bb_memdup_user(const void *src, size_t len, void **p);
  */
 TEE_Result bb_memdup_user_private(const void *src, size_t len, void **p);
 
+/*
+ * bb_strndup_user() - Duplicate a user-space string into a bounce buffer
+ * @src:    Pointer to the user string to be duplicated.
+ * @maxlen: Maximum length of the user string
+ * @dst:    Holds duplicated string on success, or unchanged on failure.
+ * @dstlen: Length of string, excluding the terminating zero, returned in
+ *          @dst.
+ *
+ * Note that the returned buffer is allocated by bb_alloc() and normally
+ * doesn't have to be freed. But if it is to be freed the supplied length
+ * to bb_free() should be dstlen + 1.
+ *
+ * Return TEE_SUCCESS on success.
+ * Return TEE_ERROR_OUT_OF_MEMORY or TEE_ERROR_ACCESS_DENIED on error.
+ */
+TEE_Result bb_strndup_user(const char *src, size_t maxlen, char **dst,
+			   size_t *dstlen);
+
 TEE_Result copy_kaddr_to_uref(uint32_t *uref, void *kaddr);
 
 uint32_t kaddr_to_uref(void *kaddr);
--- a/core/include/kernel/user_access.h
+++ b/core/include/kernel/user_access.h
@@ -31,6 +31,38 @@ static inline TEE_Result copy_from_user(void *kaddr __unused,
 
 #endif
 
+/*
+ * bb_alloc() - Allocate a bounce buffer
+ * @len:	Length of bounce buffer
+ *
+ * The bounce buffer is allocated from a per user TA context region reserved
+ * for bounce buffers. Buffers are allocated in a stack like fashion so
+ * only the last buffer can be free. Buffers generally don't have to be
+ * freed, all bounce buffer allocations are reset on each syscall entry.
+ *
+ * Return NULL on failure or a valid pointer on success.
+ */
+void *bb_alloc(size_t len);
+
+/*
+ * bb_free() - Free a bounce buffer
+ * @bb:		Buffer
+ * @len:	Length of buffer
+ *
+ * The bounce buffer is only freed if it is last on the stack of allocated
+ * bounce buffers. This function does normally not need to be called, see
+ * description of bb_alloc().
+ */
+void bb_free(void *bb, size_t len);
+
+/*
+ * bb_reset() - Reset bounce buffer allocation
+ *
+ * Resets the bounce buffer allocatation state, old pointers allocated
+ * with bb_alloc() should not be used any longer.
+ */
+void bb_reset(void);
+
 TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len);
 TEE_Result copy_to_user(void *uaddr, const void *kaddr, size_t len);
 
--- a/core/include/kernel/user_access.h
+++ b/core/include/kernel/user_access.h
@@ -66,6 +66,35 @@ void bb_reset(void);
 TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len);
 TEE_Result copy_to_user(void *uaddr, const void *kaddr, size_t len);
 
+TEE_Result clear_user(void *uaddr, size_t n);
+
+size_t strnlen_user(const void *s, size_t n);
+
+/*
+ * bb_memdup_user() - Duplicate a user-space buffer into a bounce buffer
+ * @src:    Pointer to the user buffer to be duplicated.
+ * @len:    Length of the user buffer to be duplicated.
+ * @p:      Holds duplicated bounce buffer on success, or unchanged on failure.
+ *          Note that the returned buffer is allocated by bb_alloc() and
+ *          normally doesn't have to be freed.
+ * Return TEE_SUCCESS on success.
+ * Return TEE_ERROR_OUT_OF_MEMORY or TEE_ERROR_ACCESS_DENIED on error.
+ */
+TEE_Result bb_memdup_user(const void *src, size_t len, void **p);
+
+/*
+ * bb_memdup_user_private() - Duplicate a private user-space buffer
+ * @src:    Pointer to the user buffer to be duplicated. The buffer should
+ *          be private to current TA (i.e., !TEE_MEMORY_ACCESS_ANY_OWNER).
+ * @len:    Length of the user buffer to be duplicated.
+ * @p:      Holds duplicated kernel buffer on success, or unchanged on failure.
+ *          Note that the returned buffer is allocated by bb_alloc() and
+ *          normally doesn't have to be freed.
+ * Return TEE_SUCCESS on success.
+ * Return TEE_ERROR_OUT_OF_MEMORY or TEE_ERROR_ACCESS_DENIED on error.
+ */
+TEE_Result bb_memdup_user_private(const void *src, size_t len, void **p);
+
 TEE_Result copy_kaddr_to_uref(uint32_t *uref, void *kaddr);
 
 uint32_t kaddr_to_uref(void *kaddr);
@@ -74,5 +103,26 @@ static inline void *uref_to_kaddr(uint32_t uref)
 {
 	return (void *)uref_to_vaddr(uref);
 }
+
+#define GET_USER_SCALAR(_x, _p) ({					\
+	TEE_Result __res = TEE_SUCCESS;					\
+	__typeof((_x)) __tmp = 0;					\
+									\
+	static_assert(sizeof(_x) == sizeof(*(_p)));			\
+									\
+	__res = copy_from_user(&__tmp, (const void *)_p, sizeof(_x));	\
+	(_x) = (__tmp);							\
+	__res;								\
+})
+
+#define PUT_USER_SCALAR(_x, _p) ({					\
+	TEE_Result __res = TEE_SUCCESS;					\
+	__typeof((_x)) __x = (_x);					\
+									\
+	static_assert(sizeof(_x) == sizeof(*(_p)));			\
+									\
+	__res = copy_to_user((void *)_p, &(__x), sizeof(*(_p)));	\
+	__res;								\
+})
 
 #endif /*__KERNEL_USER_ACCESS_H*/
--- a/core/include/kernel/user_mode_ctx_struct.h
+++ b/core/include/kernel/user_mode_ctx_struct.h
@@ -28,6 +28,9 @@
  * @is_32bit:		True if 32-bit TS, false if 64-bit TS
  * @is_initializing:	True if TS is not fully loaded
  * @stack_ptr:		Stack pointer
+ * @bbuf:		Bounce buffer for user buffers
+ * @bbuf_size:		Size of bounce buffer
+ * @bbuf_offs:		Offset to unused part of bounce buffer
  */
 struct user_mode_ctx {
 	struct vm_info vm_info;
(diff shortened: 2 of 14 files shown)


core/include/kernel/user_access.h
```diff
@@ -31,6 +31,38 @@ static inline TEE_Result copy_from_user(void *kaddr __unused,
 
 #endif
 
+/*
+ * bb_alloc() - Allocate a bounce buffer
+ * @len:	Length of bounce buffer
+ *
+ * The bounce buffer is allocated from a per user TA context region reserved
+ * for bounce buffers. Buffers are allocated and a stack like fashion so
```
etienne-lms, line 39: s/and/in/

core/include/kernel/user_access.h
```diff
@@ -31,6 +31,38 @@ static inline TEE_Result copy_from_user(void *kaddr __unused,
...
+ * Return NULL on failure or a valid pointer on success.
+ */
+void *bb_alloc(size_t len);
+
+/*
+ * bb_free() - Free a bounce buffer
+ * @bb:		Buffer
+ * @len:	Length of buffer
+ *
+ * The bounce buffer is only freed if it last on the stack of allocated
```
etienne-lms, line 52: s/if it last/if it is last/

core/include/kernel/user_mode_ctx_struct.h
```diff
@@ -51,6 +54,9 @@ struct user_mode_ctx {
 	bool is_32bit;
 	bool is_initializing;
 	vaddr_t stack_ptr;
+	uint8_t *bbuf;
+	size_t bbuf_size;
+	size_t bbuf_offs;
```
etienne-lms, line 59: maybe a struct for bounce buffer fields.

core/kernel/user_access.c
```diff
@@ -74,6 +94,41 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+		bb = uctx->bbuf + uctx->bbuf_offs;
+		uctx->bbuf_offs = ROUNDUP(offs, sizeof(long) * 2);
+	}
+	return bb;
+}
+
+void bb_free(void *bb, size_t len)
+{
+	struct user_mode_ctx *uctx = get_current_uctx();
+	size_t offs = ROUNDUP((vaddr_t)bb + len, sizeof(long) * 2);
```
etienne-lms, line 114: `off = ROUNDUP((vaddr_t)bb + len, sizeof(long) * 2) - uctx->bbuf;` ?

etienne-lms (commented):
Comments for commit "core: add bounce buffer to user mode context".

core/include/kernel/user_access.h
```diff
@@ -66,6 +66,35 @@ void bb_reset(void);
...
+
+/*
+ * bb_memdup_user() - Duplicate a user-space buffer into a bounce buffer
+ * @src:    Pointer to the user buffer to be duplicated.
+ * @len:    Length of the user buffer to be duplicated.
+ * @p:      Holds duplicated bounce buffer on success, or unchanged on failure.
+ *          Note that the returned buffer is allocated by bb_alloc() and
+ *          normally doesn't have to be freed.
+ * Return TEE_SUCCESS on success.
+ * Return TEE_ERROR_OUT_OF_MEMORY or TEE_ERROR_ACCESS_DENIEND on error.
```
etienne-lms, line 81: `TEE_ERROR_ACCESS_DENIED`
ditto at line 94.

etienne-lms (commented):
Comment for commit "core: add more user access functions".

core/include/kernel/user_access.h
```diff
@@ -95,6 +95,24 @@ TEE_Result bb_memdup_user(const void *src, size_t len, void **p);
...
+ * @dst:    Holds duplicated string on success, or unchanged on failure.
+ * @dstlen: Length of string, excluding the terminating zero, returned in
+ *          @dst.
+ *
+ * Note that the returned buffer is allocated by bb_alloc() and normally
+ * doesn't have to be freed. But if it is to be freed the supplied length
+ * to bb_free() should be dstlen + 1.
+ *
+ * Return TEE_SUCCESS on success.
+ * Return TEE_ERROR_OUT_OF_MEMORY or TEE_ERROR_ACCESS_DENIEND on error.
```
etienne-lms, line 111: `TEE_ERROR_ACCESS_DENIED`

core/include/kernel/user_access.h
```diff
@@ -83,7 +83,7 @@ size_t strnlen_user(const void *s, size_t n);
 TEE_Result bb_memdup_user(const void *src, size_t len, void **p);
 
 /*
- * bb_memdup_user_private() - Duplicate a private user-space buffer.
+ * bb_memdup_user_private() - Duplicate a private user-space buffer
```
etienne-lms, line 86: belongs to previous commit

etienne-lms (commented):
Comments for "core: add bb_strndup_user()".

core/include/kernel/user_mode_ctx_struct.h
jenswi-linaro: I'm not so keen on that, but I'll add it if you think it's important.

core/kernel/user_access.c
jenswi-linaro: You're right. I'll rewrite the function to make it easier to follow.

jenswi-linaro:
Update with force push.

core/pta/system.c
```diff
@@ -239,25 +243,28 @@ static TEE_Result system_dlsym(struct user_mode_ctx *uctx, uint32_t param_types,
...
 
-	sym = params[1].memref.buffer;
-	if (!sym)
-		return TEE_ERROR_BAD_PARAMETERS;
-	maxlen = params[1].memref.size;
+	res = copy_from_user(&uuid, params[0].memref.buffer, sizeof(uuid));
+	if (res)
+		return res;
+
+	res = bb_strndup_user(params[1].memref.buffer, params[1].memref.size,
```
etienne-lms, line 261:
```c
	if (!params[1].memref.buffer)

**Review by etienne-lms — COMMENTED** [review]
1 comment for commit "core: use user-access functions in ldelf interaction".

File: core/kernel/user_access.c [hunk 9]
```diff
@@ -74,6 +96,137 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+		if (prev_offs + ROUNDUP(len, BB_ALIGNMENT) == uctx->bbuf_offs)
+			uctx->bbuf_offs = prev_offs;
+	}
+}
+
+void bb_free(void *bb, size_t len)
+{
+	struct user_mode_ctx *uctx = get_current_uctx();
+
+	if (uctx)
```
etienne-lms, line 129: maybe allow `bb_free()` to do nothing when `bb == NULL`?

jenswi-linaro:
Update

etienne-lms (commented):
"arm64: Introduce permissive PAN implementation".

"core: add bounce buffer to user mode context" (late suggestion addressed or not),
"core: add more user access functions",
"core: add bb_strndup_user()" and 
"core: use user-access functions in ldelf interaction".

etienne-lms:
I agree checkpatch warning is a false positive. I could not dump CI checkpatch results so I ran checkpatch from my env.

core/kernel/user_access.c
jenswi-linaro: It will do nothing if `bb == NULL` since `bb` now is smaller than `uctx->bbuf`.

jenswi-linaro:
Tags applied.
@seonghp does this make sense to you too?

seonghp:
Overall, I like this approach to bump-allocate bounce buffers from the TA memory, instead of the core memory. These are few (minor) concerns here.

1. Would it be acceptable to allocate 4KiB buffer for each user context?
2. Even with this patch, ldelf system calls will not be covered by PAN patch as it may potentially allocate large buffer (e.g., we discussed 6KiB in #6061).
3. With this patch, will we still need to check size of user-provided buffers? 

Besides that, I think dynamically expanding `uctx->bbuf` would be a nice feature to have, but that should be the future work imho

core/kernel/user_access.c
```diff
@@ -74,6 +96,137 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+{
+	vaddr_t bbuf = (vaddr_t)uctx->bbuf;
+
+	if (bb > bbuf && IS_ALIGNED(bb, BB_ALIGNMENT)) {
+		size_t prev_offs = bb - bbuf;
+
+		if (prev_offs + ROUNDUP(len, BB_ALIGNMENT) == uctx->bbuf_offs)
+			uctx->bbuf_offs = prev_offs;
+	}
+}
```
seonghp, line 123: I am curious if this is guaranteed to work

jenswi-linaro: This should work for freeing each buffer is done in reverse order. However, the main target is dealing with attributes in `core/tee/tee_svc_cryp.c` so it's might only the last allocation that should be freeable.

jenswi-linaro:
> Overall, I like this approach to bump-allocate bounce buffers from the TA memory, instead of the core memory. These are few (minor) concerns here.
> 
> 1. Would it be acceptable to allocate 4KiB buffer for each user context?

Yes, this is allocated from TA memory pool so we normally have several megabytes of this memory.

> 2. Even with this patch, ldelf system calls will not be covered by PAN patch as it may potentially allocate large buffer (e.g., we discussed 6KiB in [[RFC] Introduce initial PAN support for OP-TEE OS #6061](https://github.com/OP-TEE/optee_os/pull/6061)).

I agree. I would prefer if we for now just disable PAN during that code path. I can provide another pull request to fix that.

> 3. With this patch, will we still need to check size of user-provided buffers?

No, with this the TA is only affecting itself if unexpected buffer sizes are supplied.

> 
> Besides that, I think dynamically expanding `uctx->bbuf` would be a nice feature to have, but that should be the future work imho

Agreed, might even be something to add as a TA property so a TA can be tuned for its use case.

seonghp:
Thank you for your comments. These all good look to me. I am now looking for potential information leakage via bounce buffer (it's used by the core but can be re-used by the TA iiuc). Let me give you update after looking at this.

Besides that, I think all my aforementioned concerns and questions are addressed.

jenswi-linaro:
When a TA is assigned new memory from the pool of physical TA memory it's always cleared first. The risk of information leakage should be minimal, limited to bugs and possibly side-channel attacks.

seonghp:
> When a TA is assigned new memory from the pool of physical TA memory it's always cleared first.

Thank you for the information! sgtm

jenswi-linaro:
We have a release due pretty soon so we should probably not merge this pull request now since it's not complete.

What's the best way forward?

@seonghp do you want to start a new pull request with these commits and your changes from https://github.com/OP-TEE/optee_os/pull/6061 reworked as needed on top?

seonghp:
Thank you for sharing info about the release schedule. I agree that we'll need more time and better not merge this too quickly.

Yes I can start a new PR built on top of this branch. Feel free to tell me if that's the way you would like 😄

jenswi-linaro:
Yes, please. Try split the "uaccess: correctly apply user-access functions" commit into smaller commits each taking care of more closely related changes as I did with the "core: use user-access functions in ldelf interaction". Perhaps three or four new commits depending on what you think makes sense. It's easier to review when a commit doesn't do many different things at the same time.

seonghp:
Okay, I will submit a new PR when ready.

core/kernel/user_access.c
```diff
@@ -74,6 +96,137 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+		uctx->bbuf_offs = ROUNDUP(offs, BB_ALIGNMENT);
+	}
+	return bb;
+}
+
+static void bb_free_helper(struct user_mode_ctx *uctx, vaddr_t bb, size_t len)
+{
+	vaddr_t bbuf = (vaddr_t)uctx->bbuf;
+
+	if (bb > bbuf && IS_ALIGNED(bb, BB_ALIGNMENT)) {
```
seonghp, line 117: The first bounce buffer will be allocated at `bbuf`. We are missing to free them as we check `bb > bbuf`, here.

core/kernel/user_access.c
```diff
@@ -74,6 +96,137 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+	uint32_t flags = TEE_MEMORY_ACCESS_READ | TEE_MEMORY_ACCESS_ANY_OWNER;
+	TEE_Result res = TEE_SUCCESS;
+	size_t l = 0;
+	char *d = NULL;
+
+	src = memtag_strip_tag_const(src);
+	res = check_access(flags, src, maxlen);
+	if (res)
+		return res;
+	l = strnlen(src, maxlen);
```
seonghp, line 218: how about using `strnlen_user()`? Otherwise, it would raise PAN error.

core/kernel/user_access.c
```diff
@@ -74,6 +96,137 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+	d = bb_alloc(l + 1);
+	if (!d)
+		return TEE_ERROR_OUT_OF_MEMORY;
+	memcpy(d, src, l);
```
seonghp, line 222: how about using `copy_from_user()`? Same reason as above.

core/kernel/user_access.c
```diff
@@ -74,6 +96,137 @@ TEE_Result copy_to_user_private(void *uaddr, const void *kaddr, size_t len)
...
+TEE_Result bb_strndup_user(const char *src, size_t maxlen, char **dst,
+			   size_t *dstlen)
+{
+	uint32_t flags = TEE_MEMORY_ACCESS_READ | TEE_MEMORY_ACCESS_ANY_OWNER;
+	TEE_Result res = TEE_SUCCESS;
+	size_t l = 0;
+	char *d = NULL;
+
+	src = memtag_strip_tag_const(src);
+	res = check_access(flags, src, maxlen);
```
seonghp, line 215: by using `strlen_user()` and `copy_from_user()`, this check can be elided.

core/kernel/user_access.c
jenswi-linaro: Good catch. We should use `>=` instead.

core/kernel/user_access.c
jenswi-linaro: We should try to manage to do this with a single call to `check_access()`.

core/kernel/user_access.c
jenswi-linaro: I'd expect that `bb_strndup_user()` should also be updated to use `enter_user_access()` and `exit_user_access()`.

core/kernel/user_access.c
seonghp: Oh I get your point. I will wrap `strnlen()` and `memcpy()` with `enter_user_access()` and `exit_user_access()` in the last commit of upcoming PR.

jenswi-linaro: Good, thanks!

core/kernel/user_access.c
seonghp: Are you going to fix this on `jenswi-linaro/bounce_buffer`, or shall I fix that in a new PR?

jenswi-linaro:
Replaced by https://github.com/OP-TEE/optee_os/pull/6135

PR #6119, opened 2023-06-15
ftrace: change implementation to use binary circular buffer

jforissier:
The current implementation of function tracing (CFG_FTRACE_SUPPORT) produces human-readable text into the output buffer that is passed to tee-supplicant and ultimately saved to the Linux filesystem. Two main issues with that:

1. The string formatting code is somewhat complex. It introduces significant overhead in the execution time of the instrumented functions.
2. The various policies about how to handle a buffer full condition (CFG_FTRACE_BUF_WHEN_FULL) are not very convenient. In particular, "shift" is typically the most desirable option because it always keeps the most recent entries, but it is very inefficient to the point of not being usable in practice.

This commit addresses the above concerns by making the ftrace buffer circular one, each entry being a 16-byte binary structure. The formatting code is offloaded to a new Python script: scripts/ftrace_format.py. The output is mostly unchanged except for an added field showing the current depth in the call stack and indentation being two spaces instead of just one.

Typical usage (captured on QEMUv8):
```
 build$ mkdir -p ../tmp
 build$ chmod a+w ../tmp
 build$ make CFG_FTRACE_SUPPORT=y CFG_FTRACE_BUF_SIZE=15000 \
             CFG_TA_MCOUNT=y CFG_ULIBS_MCOUNT=y CFG_SYSCALL_FTRACE=y \
             QEMU_VIRTFS_AUTOMOUNT=y run
 $ xtest regression_1004
 ...
 $ cp /tmp/ftrace-cb3e5ba0-adf1-11e0-998b-0002a5d5c51b.out /mnt/host/tmp
 build$ cd ..
 optee$ optee_os/scripts/ftrace_format.py \
           tmp/ftrace-cb3e5ba0-adf1-11e0-998b-0002a5d5c51b.out |
        optee_os/scripts/symbolize.py \
           -d optee_os/out/arm/core \
           -d out-br/build/optee_test_ext-1.0/ta/*/out | less
 TEE load address @ 0x5ab04000
 Function graph for TA: cb3e5ba0-adf1-11e0-998b-0002a5d5c51b @ 80085000
             |   1 |   __ta_entry() {
             |   2 |     __utee_entry() {
   43.840 us |   3 |       ta_header_get_session()
    7.216 us |   3 |       tahead_get_trace_level()
   14.480 us |   3 |       trace_set_level()
             |   3 |       malloc_add_pool() {
             |   4 |         raw_malloc_add_pool() {
   46.032 us |   5 |           bpool()
             |   5 |           raw_realloc() {
  166.256 us |   6 |             bget()
   23.056 us |   6 |             raw_malloc_return_hook()
  267.952 us |   5 |           }
  398.720 us |   4 |         }
  426.992 us |   3 |       }
             |   3 |       TEE_GetPropertyAsU32() {
   23.600 us |   4 |         is_propset_pseudo_handle()
             |   4 |         __utee_check_instring_annotation() {
   26.416 us |   5 |           strlen()
             |   5 |           check_access() {
             |   6 |             TEE_CheckMemoryAccessRights() {
             |   7 |               _utee_check_access_rights() {
             |   8 |                 syscall_check_access_rights() {
             |   9 |                   ts_get_current_session() {
    4.304 us |  10 |                     ts_get_current_session_may_fail()
   10.976 us |   9 |                   }
             |   9 |                   to_user_ta_ctx() {
    2.496 us |  10 |                     is_user_ta_ctx()
    8.096 us |   9 |                   }
             |   9 |                   vm_check_access_rights() {
             |  10 |                     vm_buf_is_inside_um_private() {
             |  11 |                       core_is_buffer_inside() {
 ...
```

### DIFF
--- a/ldelf/ftrace.c
+++ b/ldelf/ftrace.c
@@ -5,8 +5,10 @@
 
 #include <assert.h>
 #include <printk.h>
+#include <string.h>
 #include <sys/queue.h>
 #include <types_ext.h>
+#include <user_ta_header.h>
 #include <util.h>
 
 #include "ftrace.h"
@@ -25,6 +27,9 @@ bool ftrace_init(struct ftrace_buf **fbuf_ptr)
 	vaddr_t val = 0;
 	int count = 0;
 	size_t fbuf_size = 0;
+	size_t pad = 0;
+	char *p = NULL;
+	char magic[] = { 'F', 'T', 'R', 'A', 'C', 'E', 0x00, 0x01 };
 
 	res = ta_elf_resolve_sym("__ftrace_info", &val, NULL, NULL);
 	if (res)
@@ -45,17 +50,32 @@ bool ftrace_init(struct ftrace_buf **fbuf_ptr)
 
 	fbuf = (struct ftrace_buf *)(vaddr_t)finfo->buf_start.ptr64;
 	fbuf->head_off = sizeof(struct ftrace_buf);
-	count = snprintk((char *)fbuf + fbuf->head_off, MAX_HEADER_STRLEN,
+	p = (char *)fbuf + fbuf->head_off;
+	count = snprintk(p, MAX_HEADER_STRLEN,
 			 "Function graph for TA: %pUl @ %lx\n",
 			 (void *)&elf->uuid, elf->load_addr);
 	assert(count < MAX_HEADER_STRLEN);
+	p += count;
 
 	fbuf->ret_func_ptr = finfo->ret_ptr.ptr64;
 	fbuf->ret_idx = 0;
 	fbuf->lr_idx = 0;
 	fbuf->suspend_time = 0;
 	fbuf->buf_off = fbuf->head_off + count;
-	fbuf->curr_size = 0;
+	/* For proper alignment of uint64_t values in the ftrace buffer  */
+	pad = 8 - (vaddr_t)p % 8;
+	if (pad == 8)
+		pad = 0;
+	while (pad--) {
+		*p++ = 0;
+		fbuf->buf_off++;
+		count++;
+	}
+	/* Delimiter for easier decoding */
+	memcpy(p, magic, sizeof(magic));
+	fbuf->buf_off += sizeof(magic);
+	count += sizeof(magic);
+	fbuf->curr_idx = 0;
 	fbuf->max_size = fbuf_size - sizeof(struct ftrace_buf) - count;
 	fbuf->syscall_trace_enabled = false;
 	fbuf->syscall_trace_suspended = false;
@@ -70,11 +90,30 @@ void ftrace_copy_buf(void *pctx, void (*copy_func)(void *pctx, void *b,
 {
 	if (fbuf) {
 		struct ta_elf *elf = TAILQ_FIRST(&main_elf_queue);
-		size_t dump_size = fbuf->buf_off - fbuf->head_off +
-				   fbuf->curr_size;
+		char *hstart = (char *)fbuf + fbuf->head_off;
+		char *cstart = (char *)fbuf + fbuf->buf_off;
+		char *ccurr = cstart + fbuf->curr_idx * sizeof(uint64_t);
+		size_t csize = 0;
+		size_t dump_size = 0;
+		char *end = NULL;
 
 		assert(elf && elf->is_main);
-		copy_func(pctx, (char *)fbuf + fbuf->head_off, dump_size);
+
+		if (fbuf->overflow)
+			csize = fbuf->max_size;
+		else
+			csize = fbuf->curr_idx * sizeof(uint64_t);
+		dump_size = fbuf->buf_off - fbuf->head_off + csize;
+		end = hstart + dump_size;
+
+		/* Header */
+		copy_func(pctx, hstart, fbuf->buf_off - fbuf->head_off);
+		if (fbuf->overflow) {
+			/* From current index to end of circular buffer */
+			copy_func(pctx, ccurr, end - ccurr);
+		}
+		/* From start of circular buffer to current index */
+		copy_func(pctx, cstart, ccurr - cstart);
 	}
 }
 
--- a/lib/libutee/include/user_ta_header.h
+++ b/lib/libutee/include/user_ta_header.h
@@ -67,12 +67,13 @@ struct ftrace_buf {
 	uint32_t lr_idx;	/* lr index used for stack unwinding */
 	uint64_t begin_time[FTRACE_RETFUNC_DEPTH]; /* Timestamp */
 	uint64_t suspend_time;	/* Suspend timestamp */
-	uint32_t curr_size;	/* Size of ftrace buffer */
+	uint32_t curr_idx;	/* Current entry in the (circular) buffer */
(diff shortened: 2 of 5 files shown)


jenswi-linaro:
Nice!

b49020:
Thanks @jforissier for the rework. I did liked the reduction in execution time overhead via this approach. I will have a bit more deep dive here but from ftrace output format vise I have following comment:

> The output is mostly unchanged except for an added field showing the current depth in the call stack and indentation being two spaces instead of just one.

Currently we support upto maximum function call depth being 50. So if we have a long function depth then 2 spaces might lead to line overflow. Also, with current depth being displayed, I am not sure if 2 spaces does any value add here.

BTW, it would be better to have some framework to notify maintainers who maintains particular frameworks within OP-TEE OS. As in this case I only got to know about this change by chance.

lib/libutee/include/user_ta_header.h
```diff
@@ -68,13 +68,28 @@ struct ftrace_buf {
...
+ * Structure pushed to the ftrace circular buffer on entry and exit of any
+ * instrumented function.
+ * - When entering the function, @stack_depth is non zero and @val is the
+ *   function address
+ * - When returning, @stack_depth is zero and @val is the time spent in the
+ *   function and its callees, in nanoseconds
+ */
+struct __ftrace_elem {
+	uint8_t stack_depth;
+	uint8_t reserved[7];
```
b49020, line 89: This looks like wastage of space within ftrace buffer. Can't we encode stack depth in the top most nibble of function address?

lib/libutils/ext/ftrace/ftrace.c
```diff
@@ -62,112 +56,43 @@ static __noprof struct ftrace_buf *get_fbuf(void)
...
 
-	*str++ = '0';
-	*str++ = 'x';
+	elem = (struct __ftrace_elem *)((vaddr_t)fbuf + fbuf->buf_off) + idx;
+	elem->stack_depth = level;
+	elem->val = val;
 
-	for (i = 0; i < addr_size; i++)
-		*str++ = hex_str[(pc >> 4 * (addr_size - i - 1)) & 0xf];
+	if (fbuf->curr_size < idx * sizeof(*elem))
```
b49020, line 71: This check looks redundant. Shouldn't we unconditionally do the assignment below?

lib/libutils/ext/ftrace/ftrace.c
```diff
@@ -62,112 +56,43 @@ static __noprof struct ftrace_buf *get_fbuf(void)
...
+		fbuf->curr_size = idx * sizeof(*elem);
```
b49020, line 72: Shouldn't this assignment be after we have incremented `idx`? As otherwise we would miss the leaf node from trace.

lib/libutils/ext/ftrace/ftrace.c
```diff
@@ -62,112 +56,43 @@ static __noprof struct ftrace_buf *get_fbuf(void)
...
-	*str++ = '(';
-	*str++ = ')';
-	*str++ = ' ';
-	*str++ = '{';
-	*str++ = '\n';
-	*str = '\0';
+	idx++;
+	if ((idx + 1) * sizeof(*elem) > fbuf->max_size) {
+		idx = 0;
+		fbuf->curr_size = fbuf->max_size;
```
b49020, line 77: Thinking more about this, I think we should rather have a `overflow` flag instead of `curr_size` since we can very well use `idx` to retrieve the current size. Then we don't need the assignment above too.

ldelf/ftrace.c
```diff
@@ -72,9 +92,22 @@ void ftrace_copy_buf(void *pctx, void (*copy_func)(void *pctx, void *b,
...
+		char *cstart = (char *)fbuf + fbuf->buf_off;
+		char *ccurr = cstart +
+				fbuf->curr_idx * sizeof(struct __ftrace_elem);
 
 		assert(elf && elf->is_main);
-		copy_func(pctx, (char *)fbuf + fbuf->head_off, dump_size);
+
+		/* Header */
+		copy_func(pctx, hstart, fbuf->buf_off - fbuf->head_off);
+		if (fbuf->curr_size == fbuf->max_size) {
```
b49020, line 105: Here the check for overflow flag would be self describing as compared to current one.

b49020:
> I wonder if we could write some GitHub action to automatically add a comment and tag the maintainers.

That would be helpful as otherwise it would be difficult to keep an eye on every PR unless you are watching the whole Github project.

lib/libutee/include/user_ta_header.h
jforissier: I tried to avoid making assumptions on the values of virtual addresses but it is reasonable to assume the topmost byte is zero, especially since ftrace and PAuth are currently incompatible. Not too sure about what values ASLR may bring but it can be disabled if need be.
I will push an update.

lib/libutils/ext/ftrace/ftrace.c
jforissier: Agreed.

jforissier:
Update

b49020:
The changes look good to me now. Feel free to squash the review commits and add:

scripts/ftrace_format.py
```diff
@@ -8,9 +8,12 @@
 #
 #  <ASCII text> <zero or more nul bytes> FTRACE\x00\x01 <binary data>...
 #
-# <binary data> is an array of struct __ftrace_elem.
+# <binary data> is an array of 64-bit integers.
+# - When the topmost byte is 0, the entry indicates a function return and the
+# remainaing bytes are a duration in nanoseconds.
```
etienne-lms, line 13: s/remainaing/remaining/

ldelf/ftrace.c
```diff
@@ -70,11 +90,31 @@ void ftrace_copy_buf(void *pctx, void (*copy_func)(void *pctx, void *b,
 {
 	if (fbuf) {
 		struct ta_elf *elf = TAILQ_FIRST(&main_elf_queue);
-		size_t dump_size = fbuf->buf_off - fbuf->head_off +
-				   fbuf->curr_size;
+		char *hstart = (char *)fbuf + fbuf->head_off;
+		char *cstart = (char *)fbuf + fbuf->buf_off;
+		char *ccurr = cstart +
+				fbuf->curr_idx * sizeof(uint64_t);
```
etienne-lms, line 96: nitpicking: fit in a sinlge line

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with typo fixed.

jforissier:
@jenswi-linaro may I keep your ack?

jenswi-linaro:
> @jenswi-linaro may I keep your ack?

Sure

b49020:
Please update `ftrace` documentation [here](https://optee.readthedocs.io/en/latest/debug/ftrace.html) with the updated command line once you merge this PR.

jforissier:
> Please update `ftrace` documentation [here](https://optee.readthedocs.io/en/latest/debug/ftrace.html) with the updated command line once you merge this PR.

Will do, thanks form reminding me.

PR #6121, opened 2023-06-16
core: riscv: Complete virtual address related TLB operations

gagachang:
Add TLB invalidate function which is corresponding to virtual address and ASID.
Also the two wrappers are added for memory management implementation.

CC @maroueneboubakri for review.

### DIFF
--- a/core/arch/riscv/include/kernel/tlb_helpers.h
+++ b/core/arch/riscv/include/kernel/tlb_helpers.h
@@ -9,7 +9,9 @@
 #ifndef __ASSEMBLER__
 
 void tlbi_all(void);
+void tlbi_va_allasid(vaddr_t va);
 void tlbi_asid(unsigned long asid);
+void tlbi_va_asid(vaddr_t va, uint32_t asid);
 
 #endif /*!__ASSEMBLER__*/
 
--- a/core/arch/riscv/mm/tlb_helpers_rv.S
+++ b/core/arch/riscv/mm/tlb_helpers_rv.S
@@ -1,11 +1,10 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
+ * Copyright (c) 2023 Andes Technology Corporation
  * Copyright 2022 NXP
  */
 
 #include <asm.S>
-#include <riscv.h>
-#include <util.h>
 
 /* void tlbi_all(void); */
 FUNC tlbi_all , :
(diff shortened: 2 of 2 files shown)


core/arch/riscv/include/kernel/tlb_helpers.h
```diff
@@ -9,7 +9,19 @@
 #ifndef __ASSEMBLER__
 
 void tlbi_all(void);
+void tlbi_va_allasid(vaddr_t va);
 void tlbi_asid(unsigned long asid);
+void tlbi_va_asid(vaddr_t va, uint32_t asid);
+
+static inline void tlbi_va_allasid_nosync(vaddr_t va)
```
jenswi-linaro, line 16: What's the point of these _nosync functions? Do you plan to provide more primitive versions later that will be more efficient for bulk operations?

gagachang: @jenswi-linaro 
Hmm.. In fact, _nosync are meaningless for RISC-V.

Hi @maroueneboubakri 

Do we need to flush icache after flushing TLB by sfence.vma?
If not, I think we do not need _nosync related functions, what do you think?
In `core_mmu_arch.c`:

```C
void tlbi_va_range(vaddr_t va, size_t len, size_t granule)
{
	assert(granule == CORE_MMU_PGDIR_SIZE || granule == SMALL_PAGE_SIZE);
	assert(!(va & (granule - 1)) && !(len & (granule - 1)));

	mb();
	while (len) {
		tlbi_va_allasid(va);
		len -= granule;
		va += granule;
	}
	mb();
}

void tlbi_va_range_asid(vaddr_t va, size_t len, size_t granule, uint32_t asid)
{
	assert(granule == CORE_MMU_PGDIR_SIZE || granule == SMALL_PAGE_SIZE);
	assert(!(va & (granule - 1)) && !(len & (granule - 1)));

	mb();
	while (len) {
		tlbi_va_asid(va, asid);
		len -= granule;
		va += granule;
	}
	mb();
}
```

maroueneboubakri: @gagachang I agree, I remove `_nosync` then.

gagachang:
Hi @maroueneboubakri 

I remove declarations of `_nosync` functions.
Please take a look at the new commit, thanks!

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`, FWIW.
Commit message could state the commit also adds missing declaration of `tlbi_va_allasid()`.

jenswi-linaro:

gagachang:
@etienne-lms @jenswi-linaro 

Thank you. Tags applied

PR #6127, opened 2023-06-19
libutils: malloc: fix raw_malloc_buffer_overlaps_heap()

vsatoes:
When checking if there's an overlap between allocated buffer and heap, raw_malloc_buffer_overlaps_heap() considers two cases: when buffer comes before the pool and the opposite. On the first case, overlap will happen if the buffer end after the loop start. Since buf_end is computed as buf_start + len, it will point to the address of the first byte after the memory region allocated to the buffer.

Fix raw_malloc_buffer_overlaps_heap() by considering overlap only when buffer end is bigger than the pool start.

Fixes: 12d739bd5028 (libutils: use raw_malloc_*() as more primitive bget wrappers)
Signed-off-by: Vitor Sato Eschholz <vsatoes@baylibre.com>


jenswi-linaro:

etienne-lms (commented):

PR #6128, opened 2023-06-19
Introduce new python script to sign a remote processor firmware

arnopo:
This pull request is extract from  https://github.com/OP-TEE/optee_os/pull/6027

It introduces a python script to sign coprocessor firmwares.

- It accepts as input several firmware images ( for instance to support Cortex-M33 with a TF-M + a Zephyr), in ELF format.
- It adds a header an a trailer. The trailer contains properties based on a Type Lenght Data (TLV) format.
- It adds in the trailer a computed table that contains an HASS for each  segment to load
- It signs only the header and the trailer to optimize the authentication process and the images load. During the load step, the firmware images will be authenticated thanks to the HASS table generated (the hass table is signed as part of the trailer) .


scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,520 @@
...
+    tlv = TLV(TLV_INFO_MAGIC)
+
+    sign_type = ENUM_SIGNATURE_TYPE[args.key_type]
+    get_key = key_type.get(sign_type, lambda: "Invalid sign type")
+
+    key = get_key(args.keyf)
+
+    if not key.has_private():
+        logger.error('Provided key cannot be used for signing, ' +
+                     'please use offline-signing mode.')
```
etienne-lms, line 465: offline signing is not supported I think. Remove this last sentence.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,520 @@
...
+    if args.key_infof:
+        with open(args.key_infof, 'rb') as f:
+            key_info = f.read()
+        tlv.add('PKEYINFO', key_info)
+
+    # Compute the Trailer containing TLV (with 64 bit alignment)
+    trailer_buff = tlv.get()
+    s_header.tlv_length = len(trailer_buff)
+
+    # TODO: create a function for alignment
```
etienne-lms, line 532: can remove this inline comment.

scripts/sign_rproc_fw.py
```diff
@@ -381,6 +394,50 @@ def ecc_sig_size(key):
 }
 
 
+class custom_tlv(object):
+    '''
+        custom tlv provided as argument
```
etienne-lms, line 399: `Custom TLV`

scripts/sign_rproc_fw.py
```diff
@@ -381,6 +394,50 @@ def ecc_sig_size(key):
...
+        self.tlvs = {}
+        self.custom_tlvs = TLV(PLAT_TLV_INFO_MAGIC)
+        self.tlv_buff = bytearray()
+
+    def generate_tlv(self, cust_tlv):
+        # Get list of custom protected TLVs from the command-line
+        for tlv in cust_tlv:
+            if tlv[0].isalpha():
+                if tlv[0] in PLAT_TLV_PREDEF_VALUES.keys():
+                    tag = PLAT_TLV_PREDEF_VALUES[tlv[0]]
```
etienne-lms, line 412: rename `tag` into `type` or `type_id` for consistency.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -139,6 +139,10 @@ CFG_MMAP_REGIONS ?= 23
 CFG_DTB_MAX_SIZE ?= (256 * 1024)
 CFG_CORE_ASLR ?= n
 
+
+# Embed public part of this key in OP-TEE OS
+RPROC_SIGN_KEY ?= keys/default.pem
```
etienne-lms, line 144: I think `RPROC_SING_KEY` should be introduced later, with the patch introducing the RPROC PTA, where it is used, hence outside of this P-R.

etienne-lms (commented):
Comments on the 3 commits.

For commit "scripts: add remote processor firmware signature tools".
In commit message, the below sentence is missing something:
`This tool adds a binary header, a signature and a trailer`

For commit "ta: export remoteproc signature script and default key"
Remove Changed-Id tag from the commit message.

The python script looks ok, but I'm not a python expert, far from.

Using TLV structures for passing data to remoteproc TA and remote proc PTA looks fine to me.

arnopo:
Answer to the comment https://github.com/OP-TEE/optee_os/pull/6027#issuecomment-1602295508
> How come the TLVs are placed after the images? You needlessly get two "headers" to keep track of instead of just one at the beginning. Then there's also the alignment issue, do you pad the images to ensure the TLVs are at a well-aligned offset?

Trailer and TLV are inspired from the MCUBOOT signature. 
Do you propose to include TLV in the HEADER or to move the trailer before the image ( and rename it)? or to include it in the header?

One advantage of the trailer at the end is to detect image troncature early in the process, as authentication is made on this blob.
One drawback is that I have two copy in secure memory, one for the header, one for the trailer

But I can also update to have following structure:
- header
- trailer (renamed TLVS)
- signature
- images


Concerning alignement. Yes i padded  images here:
https://github.com/OP-TEE/optee_os/pull/6128/commits/becaeefa5505a236edf95ed3d750f5a0a9ab89c7#diff-9bdba1e9db0cb738cd8533a11965accfd5d9508d97437f6d9e1581dcef405c22R513
Each blob should be 64-bit aligned.

And each TLV is 64-bit aligned:
https://github.com/OP-TEE/optee_os/pull/6128/commits/becaeefa5505a236edf95ed3d750f5a0a9ab89c7#diff-9bdba1e9db0cb738cd8533a11965accfd5d9508d97437f6d9e1581dcef405c22R154

arnopo (commented):
@etienne-lms comments addressed + improvement of the inline help print

core/arch/arm/plat-stm32mp1/conf.mk
etienne-lms: I think copying the python script into the devkit makes sense in this series.

arnopo: done

arnopo:
> Answer to the comment [#6027 (comment)](https://github.com/OP-TEE/optee_os/pull/6027#issuecomment-1602295508)
> 
> > How come the TLVs are placed after the images? You needlessly get two "headers" to keep track of instead of just one at the beginning. Then there's also the alignment issue, do you pad the images to ensure the TLVs are at a well-aligned offset?
> 
> Trailer and TLV are inspired from the MCUBOOT signature. Do you propose to include TLV in the HEADER or to move the trailer before the image ( and rename it)? or to include it in the header?
> 
> One advantage of the trailer at the end is to detect image troncature early in the process, as authentication is made on this blob. One drawback is that I have two copy in secure memory, one for the header, one for the trailer
> 
> But I can also update to have following structure:
> 
> * header
> * trailer (renamed TLVS)
> * signature
> * images
> 
> Concerning alignement. Yes i padded images here: [becaeef#diff-9bdba1e9db0cb738cd8533a11965accfd5d9508d97437f6d9e1581dcef405c22R513](https://github.com/OP-TEE/optee_os/commit/becaeefa5505a236edf95ed3d750f5a0a9ab89c7#diff-9bdba1e9db0cb738cd8533a11965accfd5d9508d97437f6d9e1581dcef405c22R513) Each blob should be 64-bit aligned.
> 
> And each TLV is 64-bit aligned: [becaeef#diff-9bdba1e9db0cb738cd8533a11965accfd5d9508d97437f6d9e1581dcef405c22R154](https://github.com/OP-TEE/optee_os/commit/becaeefa5505a236edf95ed3d750f5a0a9ab89c7#diff-9bdba1e9db0cb738cd8533a11965accfd5d9508d97437f6d9e1581dcef405c22R154)

@jenswi-linaro :
Any conclusion on this topic? Is it ok to keep trailer at the end or should I move it between the header and the signature?

jenswi-linaro:
Unless there's a strong argument against it, I'd prefer having the TLVs before the images.

arnopo:
> Unless there's a strong argument against it, I'd prefer having the TLVs before the images.

Thanks for the prompt answer! No strong argument against, so I will update in this way.

arnopo:
> Unless there's a strong argument against it, I'd prefer having the TLVs before the images.

Done!

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,568 @@
...
+#                  \     +-------------+   (in bytes, 64-bit aligned)
+#                   \    +-------------+
+#                    \   | PLAT TLV sz |   Platform TLV size
+#                     \--+-------------+   (in bytes, 64-bit aligned)
+#                        +-------------+
+#                        |     TLV     |   Information used to authenticate the
+#                        |             |   firmware, stored in
+#                        +-------------+   Type-Length-Value format.
+#                        +-------------+
+#                        |   Platform  |   Specific platform information,
```
jenswi-linaro, line 65: Why do the Platform TLVs need to be separated from the other TLVs? By reserving a type namespace for each they could be kept apart in that way instead when needed.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,568 @@
...
+#                               ...
+#                        +-------------+
+#                        |   Firmware  |
+#                        |    image n  |
+#                        +-------------+
+#
+#  TLV and platform TLV chunk:
+#
+#                  ------+-------------+
+#                  |     |  TLV Magic  |
```
jenswi-linaro, line 85: I assume TLV is supposed to be short for Type, Length, and Value. Why not use those names for the fields too?

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,568 @@
...
+import os
+import sys
+import struct
+import logging
+import binascii
+
+#  Generated file structure:
+#
+#                    ----+-------------+
+#                   /    |    Magic    |
```
jenswi-linaro, line 43: Where a field has a fixed size, please write the size here for each field.

scripts/sign_rproc_fw.py
arnopo: Please could you clarify? What fields are you talking about? 
The TLV structures are stored in TLV payload, the TLV header is inspired from MCUBOOT

scripts/sign_rproc_fw.py
jenswi-linaro: As I see it you're already using two namespaces `PLAT_TLV_PREDEF_TYPES` uses a `0x0000002` prefix and `TLV_TYPES` anything below `0x00000020`. I suppose that could be improved by giving larger ranges, but the principle is there already. By looking at a type field you can already tell Platform TLVs apart from the others.

scripts/sign_rproc_fw.py
jenswi-linaro: You call this field `TLV Magic` instead of just `Type`, same below `TLV size` vs `Length` and so on.

scripts/sign_rproc_fw.py
arnopo: The `PLAT_TLV_PREDEF_TYPES `and `PLAT_TLV_TYPES` are 2 independent structure.

- The `PLAT_TLV_TYPES` is used by the `TLV` class
- The PLAT_TLV_PREDEF_VALUES is used by the `custom_tlv` class. It lists predefined TLV for platforms. The offset x 0x20 (range could be higner) is use to give a namespace starting at 0x0 for platform TLVs. User can add platform TLV in argument of the script, either using prredefined type or giving it own type ID defined in the platform PTA.

To be honest, I wonder if having predefined type for platform TLV in the TA makes sense...

scripts/sign_rproc_fw.py
arnopo: The `TLV magic ` is not the TLV, but part of the header of the chunk ( or blob). 
- the `TLV header` contains 2 fields:
  - `Magic`: a magic value  that correxpond to the ID of the chunk
  - `size` the size of the TLV payload it is 'n' * sizeof(TLV structure), n is the number of TLV structures added.
-  the ` TLV payload `contains the TLV structures. It can contain the 'n' TLV structures  
 
I  will add  more comments

arnopo:
@jenswi-linaro:
I added more information in the format description. Hope that help to understand

scripts/sign_rproc_fw.py
jenswi-linaro: I'd rather drop the "platform TLV" area and only have a single TLV area. To make it extendable you could reserve the 0x0000_FFxx range of types for extensions, while the 0x0000_00xx range is for official use and the rest are reserved for future use. Upstream OP-TEE should typically only use the official range of types as a way of keeping fragmentation in check.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,576 @@
...
+#                               ...
+#                        +-------------+
+#                        |   Firmware  |
+#                        |    image n  |
+#                        +-------------+
+#
+#  TLV and platform TLV chunk structure:
+#
+#                  ------+-------------+
+#                  |     |  TLV Magic  |  32-bit word, magic value 0x69070000
```
jenswi-linaro, line 85: I don't see why this is needed

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,576 @@
...
+#     +------------+     +-------------+  (TLV) or 0x69080000 (platform TLV)
+#     | TLV header |     +-------------+
+#     +------------+     |  TLV size   |  32-bit word, size of the TLV payload
```
jenswi-linaro, line 88: "TLV size" is already defined in the main header, no need to repeat it here. So the TLV header isn't needed.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,576 @@
...
+#                  |-----+-------------+
+#                  ------+-------------+
+#                  |     |    TLV 1    |  The payload contains a list of TLV.
+#     +------------+     |.............|  A TLV structure contains fields:
+#     | TLV payload|     |     ...     |    - Type (8-bit)
```
jenswi-linaro, line 93: Using 32-bit numbers here instead should make the value part 64-bit aligned always.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,576 @@
...
+#                   /    |    Magic    |   32-bit word, magic value equal to
+#                  /     +-------------+   0x3543A468
+#                 /      +-------------+
+#                /       |   version   |   32-bit word, version of the format
+#               /        +-------------+
+#              /         +-------------+
+# +-----------+          |  sign size  |   32-bit word, size of the signature
+# |   Header  |          +-------------+   (in bytes, 64-bit aligned)
+# +-----------+          +-------------+
+#              \         | images size |   32 bit word, size of the images to
```
jenswi-linaro, line 52: If this is 64-bit aligned, where is the padding?

arnopo (commented):
Thanks for the advice!
I posted few questions to be sure to go in the right direction.
FYI, I will be out of the office for coming weeks, so I will address your points when I come back.

scripts/sign_rproc_fw.py
jenswi-linaro: Got it, thanks.

scripts/sign_rproc_fw.py
jenswi-linaro: TLV type from 0x00020000 to 0xFFFF_FFFF should be reserved.

The TLV types sent to the PTA should probably also be official and standardized in some way.
Perhaps we need one more range for unofficial TLV types for downstream use, but this just a small detail that we can work out later if needed.

arnopo: I confirm that we will need flexibility to add unofficial TLV for downstream as we address Mass Market. We are already considering using this platform TLV to provide some boot configurations.
In short term mandatory platform TLVs are the boot address of the secure and non secure Firmware.

Standardize platform TLV would means list all of them, right? As they are platform dependent that would means having an exhaustive list of platform TLV taking account all platforms needs/constraints.  In this case I assume that  both the script and the remoteproc core TA should know this list.

An alternative would have only range for platform TLV and each platform define is own offical TLV in its PTA in upstream.
In downstream if extra plat TLV is added,  we can assume that the platform PTA will be impacted to add the new TLV definition in the list but also associated functions to treat it. 
The script today is oriented to this alternative using the " "--plat-tlv" parameter that add a platform TLV in the list without checking the match between the type and the associate data. the control will be done in

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,509 @@
...
+#                   -----+-------------+  (aligned on 64-bit), in bytes.
+#
+#                        +-------------+  Information used to authenticate the
+#                        |     TLV     |  images and boot the remote processor,
+#                        |             |  stored in Type-Length-Value format.
+#                        +-------------+  'Type' and 'Length' are 32-bit words.
+#
+#                        +-------------+
+#                        | Signature   |   Signature of the header, the TLV,
+#                        +-------------+   and the platform TLV.
```
jenswi-linaro, line 65: I guess we can skip "and the platform TLV" since that's now part of the TLV.

jenswi-linaro (commented):
I'm OK with the binary format now, thanks for your patience.
@etienne-lms would you like to continue reviewing the script?

scripts/sign_rproc_fw.py
arnopo: Right, I 'm waiting @etienne-lms feedback and I will fix this.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,520 @@
...
+        raise Exception(("Actual signature length is not equal to ",
+                         "the computed one: {} != {}".
+                         format(len(signature), sign_size)))
+
+    s_header.dump()
+
+    with open(args.outf, 'wb') as f:
+        f.write(header)
+        f.write(signature)
+        f.write(bytearray(sign_size - s_header.sign_length))
```
etienne-lms, line 497: Is it possible `sign_size != s_header.sign_length`? This line can be removed IMO.

scripts/sign_rproc_fw.py
etienne-lms: I agree with Jens, TLV covers cells for TA and PTA. Can remove "and the platform TLV" here.

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,509 @@
...
+#              /         +-------------+
+# +-----------+          +-------------+
+# |   Header  |          |  sign size  |  32-bit word, size of the signature
+# +-----------+          +-------------+  (aligned on 64-bit), in bytes.
+#              \         +-------------+
+#               \        | images size |  32-bit word, size of the images to
+#                \       +-------------+  load (aligned on 64-bit), in bytes.
+#                 \      +-------------+
+#                  \     |  TLV size   |  32-bit word, size of the generic TLV
+#                   -----+-------------+  (aligned on 64-bit), in bytes.
```
etienne-lms, line 56: nitpicking: since payloads order is header;TLVs;signature;fmw binaries, would it make sense to use that ordering in the header:  magic;version;tlv-size;sign-size;images-size?

arnopo (commented):
I take opportunity to squash the fixes

etienne-lms (commented):
For commits "scripts: add remote processor firmware signature tool".
3 typos in inline comments.
Please remove the `Change-Id:` tag from commit messages.
The structure and script look OK to me. 
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with my comments addressed.

"ta: export remoteproc signature script and default key".
Ditto for `Change-Id:` tag.
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

arnopo:
> For commits "scripts: add remote processor firmware signature tool". 3 typos in inline comments. Please remove the `Change-Id:` tag from commit messages. The structure and script look OK to me. `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with my comments addressed.
> 
> "ta: export remoteproc signature script and default key". Ditto for `Change-Id:` tag. `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

Done, thanks for the review!

scripts/sign_rproc_fw.py
```diff
@@ -0,0 +1,511 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Copyright (C) 2023, STMicroelectronics - All Rights Reserved
```
jenswi-linaro, line 4: Do you need the "All Rights Reserved" clause? It's rather pointless, but OK if required by your legal team.

jenswi-linaro (commented):
Please add
`Acked-by: Jens Wiklander <jens.wiklander@linaro.org>` for
 "scripts: add remote processor firmware signature tool"
and
`Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>` for
"ta: export remoteproc signature script and default key".

scripts/sign_rproc_fw.py
arnopo: I removed it

arnopo:
> Please add `Acked-by: Jens Wiklander <jens.wiklander@linaro.org>` for "scripts: add remote processor firmware signature tool" and `Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>` for "ta: export remoteproc signature script and default key".


Thanks for the review !
I added your `acked-by` and `reviewed-by`, take opportunity to rebase on master branch, and fix the log level from `debug` to `info` in the script

arnopo:
Thanks!
I'm preparing the next PR containing the remoteproc TA + the remoteproc PTA API. I will send it this week or at the beginning of next week.

PR #6133, opened 2023-06-23
core: mm: shm: fix assert when unmap object

omasse-linaro:
An assert could occur in reg_shm_unmap_helper when performing a simple stress test with two threads which invoke in a infinite loop a ping command to a TA with an input shared memory as parameter.

```
E/TC:1 1 assertion 'r->mm' failed at core/arch/arm/mm/mobj_dyn_shm.c:117 <reg_s>
E/TC:1 1 Panic at core/kernel/assert.c:28 <_assert_break>
E/TC:1 1 TEE load address @ 0xe1b32000

```
reg_shm_unmap_helper is called by mobj_reg_shm_dec_map when the mapcount reach 0 with an already unmapped object.

Fixed by taking the reg_shm_map_lock spin lock before incrementing or decrementing the map refcount.

### DIFF
--- a/core/arch/arm/mm/mobj_dyn_shm.c
+++ b/core/arch/arm/mm/mobj_dyn_shm.c
@@ -184,20 +184,10 @@ static TEE_Result mobj_reg_shm_inc_map(struct mobj *mobj)
 	uint32_t exceptions = 0;
 	size_t sz = 0;
 
-	while (true) {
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
I'd like to find out why `refcount_inc()` and `refcount_dec()` can't be used without a spinlock. The functions are pretty useless if they can't be used without locks.

PR #6136, opened 2023-06-26
plat-k3: Reduce logging verbosity during runtime

glneo:
The default UART can be disabled at runtime for power management. Any prints after boot from OP-TEE could cause a bus data abort if the UART has been disabled. After boot lower the log level to only print error level messages. This matches what is done in TF-A with the "runtime" console.


core/arch/arm/plat-k3/main.c
```diff
@@ -49,6 +49,15 @@ void console_init(void)
 	register_serial_console(&console_data.chip);
 }
 
+static TEE_Result console_runtime(void)
+{
+	/* Console is only used for errors after this point */
+	trace_set_level(TRACE_ERROR);
```
jforissier, line 55: Isn't this a bit inconvenient when doing development? Wouldn't a build-time conditional (`if IS_ENABLED(CFG_FOO)`) help?

glneo: Sure, I'll add a build flag to disable this feature when debugging.

jforissier:
> The default console can be disabled at runtime for power management. Any
> prints after boot from OP-TEE could cause a bus data abort if the UART has
> been disabled. After boot lower the log level to only print error level
> messages.

How can error level message not cause a data abort then?

core/arch/arm/plat-k3/conf.mk
```diff
@@ -1,6 +1,7 @@
 CFG_WITH_STATS ?= y
 CFG_CRYPTO_WITH_CE ?= y
 CFG_CONSOLE_UART ?= 0
+CFG_CONSOLE_RUNTIME ?= y
```
etienne-lms, line 4: This config switch name looks very generic. It should be documented in _mk/config.mk_ or be renamed to something like `CFG_K3_CONSOLE_RUNTIME`.

Maybe a generic switch would set a specific level post boot, like `CFG_CORE_RUNTIME_LOG_LEVEL`.

jforissier: > Maybe a generic switch would set a specific level post boot, like `CFG_CORE_RUNTIME_LOG_LEVEL`.

+1. Note that it would waste some memory space, for example if `CFG_TEE_CORE_LOG_LEVEL=3` and the post-boot log level is 2 then many debug messages (level 3) would be there but never used.

etienne-lms: True. 
Thinking about this, would be make sense to use for example the stats PTA or device PTA to modify the log level at runtime?

jforissier: I like the idea although it is probably a different use case than what this PR is supposed to address.

core/arch/arm/plat-k3/main.c
```diff
@@ -49,6 +49,17 @@ void console_init(void)
 	register_serial_console(&console_data.chip);
 }
 
+static TEE_Result console_runtime(void)
+{
+	/* Console is only used for errors after this point */
+	if (IS_ENABLED(CFG_CONSOLE_RUNTIME))
```
etienne-lms, line 55: if this is to be generic, this function could be implemented in _core/kernel/console.c_

Issue #6120, opened 2023-06-15
optee 3.21.0: core: mobj_reg_shm_dec_map


omasse-linaro:
Hi

I'm getting an assert in reg_shm_unmap_helper when performing a simple stress test with two threads which invoke in a infinite loop a ping command to a TA with an input shared memory as parameter.

E/TC:1 1 assertion 'r->mm' failed at core/arch/arm/mm/mobj_dyn_shm.c:117 <reg_s>
E/TC:1 1 Panic at core/kernel/assert.c:28 <_assert_break>
E/TC:1 1 TEE load address @ 0xe1b32000

it seems that reg_shm_unmap_helper is called by mobj_reg_shm_dec_map when the mapcount reach 0 with an already unmapped object.

looking at the refcount.h (core/include/kernel/refcount.h) it is specify that when refcount_dec(r) is true, the object o is uninitialize if refcount_val(r) is still 0 AND o is not NULL.

however mobj_reg_shm_dec_map in core/arch/arm/mm/mobj_dyn_shm.c do not check if r->mm is not NULL before calling reg_shm_unmap_helper.

doing so, my stress test run fine without assert.
could we have a race condition when two threads call to mobj_reg_shm_dec_map occur at the same time ?

after doing refcount_dec on mapcount, thread 1 take the lock before thread 2 and call reg_shm_unmap_helper and release the lock.
thread 2 then take the lock and call reg_shm_unmap_helper as well, which fail with assert(r->mm).

jenswi-linaro:
Each call to `mobj_dec_map()` must be matched by a previous call to `mobj_inc_map()`. When that rule is followed then the race you're describing can't occur.

I believe the real problem is that the `mobj_inc_map()` and `mobj_dec_map()` calls are unbalanced for some reason.

omasse-linaro:
indeed, I see that you have fix a race issue in mobj_reg_shm_inc_map, but why not just get the reg_shm_map_lock spin lock before incrementing or decrementing the map refcount ?

omasse-linaro:
Yes, I just reproduce the issue with qemu as well.
![Screenshot from 2023-06-21 11-27-52](https://github.com/OP-TEE/optee_os/assets/4939714/c1073263-4398-485f-9802-279e12c8fd7a)

omasse-linaro:
draft submitted here https://github.com/OP-TEE/optee_os/pull/6133

jenswi-linaro:
What's regression case 1040?

[...]

PR #6137, opened 2023-06-26
core: fix race in mobj_reg_shm_dec_map()

jenswi-linaro:
Fixes a race in mobj_reg_shm_dec_map() when r->mm is NULL. This is similar to the race fixed by commit 06ea466f9c19 ("core: fix race in mobj_reg_shm_inc_map()"), but with one more possibility.

The problem goes like:
A. Thread 1 calls mobj_reg_shm_dec_map() at the same time as thread 2
   calls mobj_reg_shm_inc_map().
B. Thread 1 decreases mapcount to zero and tries to take the spinlock,
   but thread 1 is suspended before it has acquired the spinlock.
C. Thread 2 sees that mapcount is zero and takes the spinlock and maps
   the memory.
D. Thread 2 calls mobj_reg_shm_dec_map(), mapcount reaches zero again
   and the shared memory is unmapped and r->mm is set to NULL.
E. Thread 1 is finally resumed and acquires the spinlock, mapcount is still
   zero but r->mm is also NULL.

To fix the problem at step E above check that r->mm is still non-NULL.

Fixes: 06ea466f9c19 ("core: fix race in mobj_reg_shm_inc_map()")

### DIFF
--- a/core/arch/arm/mm/mobj_dyn_shm.c
+++ b/core/arch/arm/mm/mobj_dyn_shm.c
@@ -200,7 +200,7 @@ static TEE_Result mobj_reg_shm_inc_map(struct mobj *mobj)
 	}
 
 	/*
-	 * If we have beated another thread calling mobj_reg_shm_dec_map()
+	 * If we have beaten another thread calling mobj_reg_shm_dec_map()
 	 * to get the lock we need only to reinitialize mapcount to 1.
 	 */
 	if (!r->mm) {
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Another fix for https://github.com/OP-TEE/optee_os/issues/6120

jenswi-linaro:
Updated the commit message for `ffa_dec_map()`.

omasse-linaro (approved):

jforissier (commented):

jenswi-linaro:
Tag applied.

omasse-linaro:
Acked-by: Olivier Masse <olivier.masse@nxp.com>

jenswi-linaro:
Tag applied

PR #6139, opened 2023-06-27
qemu_armv8a: fix build with CFG_USER_TA_TARGETS=ta_arm32

jforissier:
The proper way to build in-tree TAs in 64-bit mode by default is to set supported-ta-targets to "ta_arm64 ta_arm32". Indeed, the default target is always defined as the first entry in supported-ta-targets, as documented in mk/config.mk.

Fixes the following build error:

 $ make CFG_USER_TA_TARGETS=ta_arm32 PLATFORM=vexpress-qemu_armv8a
 bash: -W: invalid option
 ...

default-user-ta-target is not to be used by the platform configuration files. It is meant to be set by the main Makefile. For this reason, replace the conditional assignment (?=) with $(call force, ...) in order to catch inconsistencies in a more friendly way.

Fixes: 07031b23de23 ("qemu_armv8a: set default-user-ta-target ?= ta_arm64")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ include core/core.mk
 
 # Platform/arch config is supposed to assign the targets
 ta-targets ?= invalid
-default-user-ta-target ?= $(firstword $(ta-targets))
+$(call force,default-user-ta-target,$(firstword $(ta-targets)))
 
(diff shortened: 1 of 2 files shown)


etienne-lms (commented):

jenswi-linaro (commented):

PR #6147, opened 2023-06-30
RISC-V: Support device tree

gagachang:
This series add support for device tree.
The device tree may be provided by early boot stage (i.e., M-mode firmware such as openSBI), or embedded in OP-TEE core.
RISC-V boot stage will initialize and update the device tree if it is necessary.
For example, add reserved-memory node for secure memory.

### DIFF
--- a/core/arch/riscv/kernel/boot.c
+++ b/core/arch/riscv/kernel/boot.c
@@ -43,6 +43,43 @@ static struct dt_descriptor external_dt __nex_bss;
 #endif
 
 #if defined(CFG_DT)
+void *get_external_dt(void)
+{
+	if (!IS_ENABLED(CFG_EXTERNAL_DT))
+		return NULL;
+
+	assert(cpu_mmu_enabled());
+	return external_dt.blob;
+}
+
+static TEE_Result release_external_dt(void)
+{
+	int ret = 0;
+
+	if (!IS_ENABLED(CFG_EXTERNAL_DT))
+		return TEE_SUCCESS;
+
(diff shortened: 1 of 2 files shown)


jforissier:
Hi @gagachang,
That is a lot of code in common with Arm... How about moving as much as possible to `core/kernel/dt.c`?

gagachang:
> Hi @gagachang,
> 
> That is a lot of code in common with Arm... How about moving as much as possible to `core/kernel/dt.c`?

Sounds great!
It seems that only init and update should be architecture-specific.
Other functions could be common code.
I will prepare for that.

gagachang:
Hi @jforissier 

I found that `get_dt()`, `get_external_dt()`, ... etc functions are declared in `core/include/kernel/boot.h`.
And `core/kernel/dt.c` is more likely a location for common dt APIs.

How about creating a new file, i. e., `core/kernel/boot.c` for `get_dt()`, `get_external_dt()`, etc?

jforissier:
> Hi @jforissier
> 
> I found that `get_dt()`, `get_external_dt()`, ... etc functions are declared in `core/include/kernel/boot.h`. And `core/kernel/dt.c` is more likely a location for common dt APIs.
> 
> How about creating a new file, i. e., `core/kernel/boot.c` for `get_dt()`, `get_external_dt()`, etc?

Sounds good to me.

gagachang:
Hi @jforissier 

I create v2 PR https://github.com/OP-TEE/optee_os/pull/6150 for this.
Let's move to that PR!

PR #6149, opened 2023-06-30
mk/aosp_optee.mk: allow building ta with specified target architecture

taddk:
Introduce $(local_optee_ta_target) to specify the target architecture of the built TA in aosp build environment.
Since some ta might prefer better performance with ta_arm64, and some prefer smaller code size with ta_arm32.


mk/aosp_optee.mk
```diff
@@ -115,6 +115,10 @@ LOCAL_MODULE_TAGS := optional
 LOCAL_REQUIRED_MODULES := $(local_module_deps)
 
 TA_TMP_DIR := $(subst /,_,$(LOCAL_PATH))
+ifneq ($(filter-out $(OPTEE_TA_TARGETS),$(local_optee_ta_target)),)
+TA_DEV_KIT_DIR := $(TA_DEV_KIT_DIR:$(OPTEE_TA_TARGETS)=$(local_optee_ta_target))
```
jenswi-linaro, line 119: Why not assign the desired value the first time around line 58?
What is `$(TA_DEV_KIT_DIR:$(OPTEE_TA_TARGETS)=$(local_optee_ta_target))` supposed to do?

mk/aosp_optee.mk
```diff
@@ -115,6 +115,10 @@ LOCAL_MODULE_TAGS := optional
...
+TA_TMP_DIR := $(addsuffix _$(local_optee_ta_target), $(TA_TMP_DIR))
```
jenswi-linaro, line 120: Why is this needed?

mk/aosp_optee.mk
taddk: Indeed, it is possible to move this to line 58.
We do the substitution to replace the default configuration set by `$(OPTEE_TA_TARGETS)` to `$(local_optee_ta_target)` specified by module Android.mk. I think we can move this to line 58 like,
```
ifneq ($(local_optee_ta_target),)
TA_DEV_KIT_DIR := $(ABS_OPTEE_OS_OUT_DIR)/export-$(local_optee_ta_target)
else
TA_DEV_KIT_DIR := $(ABS_OPTEE_OS_OUT_DIR)/export-${OPTEE_TA_TARGETS}
endif
```

mk/aosp_optee.mk
taddk: For a common static library that could be linked by several TAs (in different architectures), we would put the intermediate object and library files in different architecture to different folder. Also, it helps the TA in specified architecture (different from the default one) to find the correct folder of the static library in the same architecture.

mk/aosp_optee.mk
jenswi-linaro: Sounds good

mk/aosp_optee.mk
jenswi-linaro: OK

taddk: Would that be acceptable if we also change the original path by adding the `"_ta_{target_arch}"` to the `$(TA_TMP_DIR)` no matter `$(local_optee_ta_target)` holds a value? like,

```
ifneq ($(local_optee_ta_target),)
TA_TMP_DIR := $(addsuffix _$(local_optee_ta_target), $(subst /,_,$(LOCAL_PATH)))
else
TA_TMP_DIR := $(addsuffix _$(OPTEE_TA_TARGETS), $(subst /,_,$(LOCAL_PATH)))
endif
```

jenswi-linaro: Sure. In fact that should allow using a helper variable instead of `local_optee_ta_target` and `OPTEE_TA_TARGETS`.

taddk: Good idea. Updated.

jenswi-linaro:

taddk:
Tag applied.

Sorry for late reply. if anything need to be improved or changed, just let me know.

PR #6155, opened 2023-07-05
change load address of OPTEE-OS from 0x36000000 to 0x02100000

rutigl:
Changes load address of OPTEE-OS from 0x36000000 to 0x02100000

Co-developed-by: Hila Miranda-Kuzi <hila.miranda.kuzi1@gmail.com>

### DIFF
--- a/core/include/drivers/pinctrl.h
+++ b/core/include/drivers/pinctrl.h
@@ -155,7 +155,6 @@ TEE_Result pinctrl_get_state_by_idx(const void *fdt __unused,
 static inline
 void pinctrl_free_state(struct pinctrl_state *state __unused)
 {
-	return TEE_ERROR_NOT_SUPPORTED;
 }
 
 static inline


jbech-linaro:

core/arch/arm/plat-nuvoton/conf.mk
```diff
@@ -21,14 +23,14 @@ CFG_NPCM_DEBUG ?= n
...
+CFG_TZDRAM_START ?= 0x02100000
+CFG_TZDRAM_SIZE  ?= 0x03f00000
 # SHM chosen arbitrary, in a way that it does not interfere
 # with initial location of linux kernel, dtb and initrd.
-CFG_SHMEM_START ?= 0x2FD00000
+CFG_SHMEM_START ?= 0x06000000
 CFG_SHMEM_SIZE  ?= 0x00200000
 # When Secure Data Path is enable, last MByte of TZDRAM is SDP test memory.
-CFG_TEE_SDP_MEM_SIZE ?= 0x00400000
+CFG_TEE_SDP_MEM_SIZE ?= 0x05F00000
```
etienne-lms, line 33: It is strange that `CFG_TEE_SDP_MEM_SIZE` is higher than  `CFG_TZDRAM_SIZE`. Please crosscheck this value.

etienne-lms (commented):
Change commit header line to somethoing like `plat-nuvoton: update configuration`.

core/arch/arm/plat-nuvoton/conf.mk
```diff
@@ -12,6 +12,8 @@ $(call force,CFG_GIC,y)
 $(call force,CFG_ARM_GICV2,y)
 $(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
 $(call force,CFG_16550_UART,y)
+$(call force,CFG_EXTERNAL_DT,n)
+$(call force,CFG_WARN_INSECURE,n)
```
jenswi-linaro, line 16: It would be nice to have this change in a separate commit with an explanation of why it's not needed.

jenswi-linaro:
Please add a `plat-nuvoton: ` prefix to the title of the patches.

How can BootBlock make the configuration secure if for instance `tee_otp_get_hw_unique_key()` only returns zeroes? The purpose of `CFG_WARN_INSECURE` is to warn until all such security problems have been addressed. If you have out-of-tree patches that replace `tee_otp_get_hw_unique_key()` among other things to make the configuration secure, then I believe setting `CFG_WARN_INSECURE=n` should be in that list too, and not upstream.

core/arch/arm/plat-nuvoton/conf.mk
rutigl: Agree, I confused between CFG_TEE_SDP_MEM_SIZE and CFG_TEE_SDP_MEM_BASE

rutigl:
> Please add a `plat-nuvoton: ` prefix to the title of the patches.
> 
> How can BootBlock make the configuration secure if for instance `tee_otp_get_hw_unique_key()` only returns zeroes? The purpose of `CFG_WARN_INSECURE` is to warn until all such security problems have been addressed. If you have out-of-tree patches that replace `tee_otp_get_hw_unique_key()` among other things to make the configuration secure, then I believe setting `CFG_WARN_INSECURE=n` should be in that list too, and not upstream.

BootBlock validates if every image (OPTEE, TF-A, U-Boot, Linux) is signed using ROM private/public key pair and crc is OK. This is a security check made in our platform.

jenswi-linaro:
> > Please add a `plat-nuvoton: ` prefix to the title of the patches.
> > How can BootBlock make the configuration secure if for instance `tee_otp_get_hw_unique_key()` only returns zeroes? The purpose of `CFG_WARN_INSECURE` is to warn until all such security problems have been addressed. If you have out-of-tree patches that replace `tee_otp_get_hw_unique_key()` among other things to make the configuration secure, then I believe setting `CFG_WARN_INSECURE=n` should be in that list too, and not upstream.
> 
> BootBlock validates if every image (OPTEE, TF-A, U-Boot, Linux) is signed using ROM private/public key pair and crc is OK. This is a security check made in our platform.

That's good, but there's still the problem with a HUK that isn't unique and known to the entire world. It's quite practical on a development device, but perhaps not so good on a production device.

jenswi-linaro:
The patch "plat-nuvoton: update configuration" does quite a bit more than it's supposed to.

jenswi-linaro:
The patch "plat-nuvoton: configuration change" also does a bit more than expected.

jenswi-linaro:
> I have implemented tee_otp_get_hw_unique_key() function and tested it, but if I remove $(call force,CFG_WARN_INSECURE,n) from conf.mk, I still receive Insecure warning for OPTEE, so I returned this setting.

By setting `CFG_WARN_INSECURE=n` in upstream without securing the platform in upstream you defeat the purpose of the commit 9e42008d9512 ("config: add flag CFG_WARN_INSECURE"). Do you really think that the upstream users of your platform benefit from not being warned that their software might be insecure? If they don't like the warning then I also assume that they don't like that their software isn't secure.

rutigl:
I have nothing against CFG_WARN_INSECURE flag. I am just asking what should I change in my code to not receive this warning

rutigl:
> The patch "plat-nuvoton: configuration change" also does a bit more than expected.
No problem, do you want me to make three commits: 
- 1 for change load address,
- 2 for remove DT warning
- 3 for implementing HUK reading?

jenswi-linaro:
Once no further security configuration is needed for the platform it would make sense to remove the `CFG_WARN_INSECURE` flag. However, as far as I know, we don't have any upstream platform with that flag disabled.

And now you closed this PR so I'm not sure what you're trying to upstream.

rutigl:
The same thing as before, just without CFG_WARN_INSECURE flag, and
rearranged with three separate commits.

2023-07-18 12:23 GMT+03:00, Jens Wiklander ***@***.***>:

PR #6161, opened 2023-07-09
Fix PAN in ldelf and user_ta

gagachang:
When `CFG_PAN` is enabled, the kernel can not directly touch the user memory.
These two commits fix the invalid accesses on user stack by applying user-access functions to duplicate the user data into the bounce buffers.

### DIFF
--- a/core/kernel/ldelf_loader.c
+++ b/core/kernel/ldelf_loader.c
@@ -120,6 +120,8 @@ TEE_Result ldelf_init_with_ldelf(struct ts_session *sess,
 	uint32_t panic_code = 0;
 	uint32_t panicked = 0;
 	uaddr_t usr_stack = 0;
+	struct ldelf_arg *arg_bbuf = NULL;
+	void *bbuf = NULL;
 
 	usr_stack = uctx->ldelf_stack_ptr;
 	usr_stack -= ROUNDUP(sizeof(*arg), STACK_ALIGNMENT);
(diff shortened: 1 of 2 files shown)


gagachang:
Hello @jenswi-linaro @seonghp 

I encountered some faults when I enabled `CFG_PAN`.
It seems that some user parameters, which are put on the user stack, are not correctly accessed by using user-access functions.
Including `struct ldelf_arg` and `struct utee_params`.

Please take a look at this PR to fix this issue.
Thanks!

seonghp:
Hello @gagachang, thank you for having interest in PAN support patch 😄. In both locations, the kernel is trying to access to the memory region that were previously accessible to user. At those points (after returning from the `thread_enter_user_mode()`), it was safe (at least no PAN exception was raised by QEMU) to access those addresses. However, I struggled to pinpoint the exact line in code where the user-access is revoked. Let me get back to you when I find it.

gagachang:
Hello @seonghp 

Thanks for quick reply!
I am not sure how ARM's PAN works. Does it raise exception when kernel accesses user's page?
If yes, PAN may work until user pages are unmapped, even if the thread is returning from thread_enter_user_mode().

seonghp:
Yes, when PAN is enabled and the kernel tries to access the user memory, it will result in a permission fault.

After some debugging, it seems like maybe I was trusting QEMU too much.

I will investigate this issue further, test on different platforms and see what happens (tbh it will take some time).

Thank you for catching this.

gagachang:
Hi @seonghp 
Thanks!
In fact, I tested CFG_PAN on QEMU RISC-V virtual platform and find these issues.
If the issues do not happen on your side, maybe they are specific on RISC-V architecture.
Just want to confirm it.

seonghp:
Thank you for testing and sharing results. Could you also share if these were the only issues that you have found when testing with QEMU RISC-V?

gagachang:
> Thank you for testing and sharing results. Could you also share if these were the only issues that you have found when testing with QEMU RISC-V?

Yes, without these patches, the page fault will be raised by QEMU RISC-V.
The issues are gone by adopting these patches.

seonghp:
I observed that the user memory regions were still mapped into the page table after returning from the `thread_enter_user_mode()`. Provided that, those accesses to the user memory regions (which are replaced with user-access functions in this PR) are supposed to raise exceptions, but they didn't on QEMU ARMv8 for currently unknown reason (probably some optimizations in QEMU skips PAN checks sometimes).

I have not tested this issue on other platforms yet. However IMO the changes suggested by this PR has be merged to ensure that the previous patch for restricting kernel's access to unprivileged memory, works on other platforms (e.g, QEMU RISC-V).

That being said, I also want to hear what the maintainers think, too.

core/kernel/ldelf_loader.c
```diff
@@ -159,27 +161,37 @@ TEE_Result ldelf_init_with_ldelf(struct ts_session *sess,
...
+	arg_bbuf = (struct ldelf_arg *)bbuf;
+
 	if (is_user_ta_ctx(uctx->ts_ctx)) {
 		/*
 		 * This is already checked by the elf loader, but since it runs
 		 * in user mode we're not trusting it entirely.
 		 */
-		if (arg->flags & ~TA_FLAGS_MASK)
+		if (arg_bbuf->flags & ~TA_FLAGS_MASK) {
+			bb_free(bbuf, sizeof(struct ldelf_arg));
```
jenswi-linaro, line 176: This call isn't needed since the `uctx` is going to be removed anyway.
The rules for the `bb_*()` functions are different from the normal `malloc()` / `free()` pattern since the bounce buffer state is reset for each syscall.

core/kernel/user_ta.c
```diff
@@ -97,27 +97,38 @@ static TEE_Result init_utee_param(struct utee_params *up,
...
+	TEE_Result res = TEE_SUCCESS;
+	size_t n = 0;
+	struct utee_params *up_bbuf = NULL;
+	void *bbuf = NULL;
+
+	res = bb_memdup_user(up, sizeof(struct utee_params), &bbuf);
+	if (res)
+		return;
+
+	up_bbuf = (struct utee_params *)bbuf;
```
jenswi-linaro, line 111: Cast isn't needed.

jenswi-linaro (commented):
Please use more compact titles, maybe:
- "core: fix ldelf invalid access to user stack with CFG_PAN=y"
- "core: fix update from user parameters with CFG_PAN=y"

It would be nice with "Fixes:" tags too,
Fixes: 4e154320e47c ("core: Apply finer-grained PAN")

jenswi-linaro:
> I observed that the user memory regions were still mapped into the page table after returning from the `thread_enter_user_mode()`. Provided that, those accesses to the user memory regions (which are replaced with user-access functions in this PR) are supposed to raise exceptions, but they didn't on QEMU ARMv8 for currently unknown reason (probably there might be some optimizations in QEMU to skip PAN checks).

I'm planning to play around a bit with PAN on QEMU v8. I'll take a look at why PAN checks might be skipped sometimes.

core/kernel/ldelf_loader.c
gagachang: Clear explanation, thanks!

gagachang:
Hi @jenswi-linaro 
Comments are addressed in latest commits. Please check it

jenswi-linaro:
Looks good.

PR #486, opened 2015-10-06
v1.0.0 Release Notes

pascal-brand-st-dev:
Signed-off-by: Pascal Brand <pascal.brand@st.com>

PR #418, opened 2015-07-28
arm64: SHA-1 using ARMv8-A cryptographic extensions

jforissier:
- LibTomCrypt: add a new macro, HASH_PROCESS_NBLOCKS, similar to
HASH_PROCESS but accepts a function that digests n blocks of data, not
just 1.
- Import sha1_ce_transform() from the Linux kernel (Linaro contribution)
which implements the main SHA-1 transform in assembler using the ARMv-8
cryptographic extensions.
- Acceleration is enabled by setting CFG_CRYPTO_SHA1_ARM64_CE=y (this
is the default when PLATFORM=hikey).

Performance was compared to the plain C version using sha-perf
(https://github.com/linaro-swg/sha-perf.git). Average hashing speed on
HiKey is (MiB/s):

```
   Size | Accelerated?
  (KiB) |    No    Yes
  ------+-------------
      1 |  14.2   22.9
      2 |  20.6   43.9
      4 |  26.6   81.2
      8 |  31.1  140.2
     16 |  34.0  220.7
     32 |  35.8  311.5
     64 |  36.7  393.2
    128 |  37.2  452.5
    256 |  37.4  489.1
    384 |  37.4  501.5
```

Signed-off-by: Jerome Forissier <jerome.forissier@linaro.org>

Issue #332, opened 2015-06-05
about optee-os task scheduling


louisjohn0416:
i found there is no task scheduling in opee-os, So we cannot run TUI on optee-os ?

jenswi-linaro:
There's on explicit task scheduling as we're scheduled by normal world. Whenever we receive an IRQ we exit to normal world to let it handle the IRQ through the normal world vector. This IRQ can be a timer interrupt or whatever. When the thread that tries to reenter secure world get CPU time again we'll enter secure world and resume the work there.

louisjohn0416:
Got it , thank you !

PR #6168, opened 2023-07-11
crypto : custom HUK derviation with SAES on STM32MP13

tbourgoi:
This pull request adds the driver SAES for the platform STM32MP13
SAES is used to derive a key for the HUK.
To do that,  we override the function huk_subkey_derive().

One false positive will be found by the CI, the macro CLRSETBITS changes one of its parameters

### DIFF
--- a/core/drivers/crypto/stm32/common.h
+++ b/core/drivers/crypto/stm32/common.h
@@ -8,7 +8,31 @@
 
 #include <tee_api_types.h>
 
+enum stm32_cipher_ip_id {
+	CRYP_IP,
+	SAES_IP,
+};
+
+/*
+ * Crypto algorithm common macro used in stm32_saes and stm32_cryp driver
+ */
+
+#define INT8_BIT			U(8)
+#define AES_BLOCK_SIZE_BIT		U(128)
+#define AES_BLOCK_SIZE			(AES_BLOCK_SIZE_BIT / INT8_BIT)
+#define AES_BLOCK_NB_U32		(AES_BLOCK_SIZE / sizeof(uint32_t))
+#define DES_BLOCK_SIZE_BIT		U(64)
+#define DES_BLOCK_SIZE			(DES_BLOCK_SIZE_BIT / INT8_BIT)
+#define DES_BLOCK_NB_U32		(DES_BLOCK_SIZE / sizeof(uint32_t))
+#define MAX_BLOCK_SIZE_BIT		AES_BLOCK_SIZE_BIT
+#define MAX_BLOCK_SIZE			AES_BLOCK_SIZE
+#define MAX_BLOCK_NB_U32		AES_BLOCK_NB_U32
+#define AES_KEYSIZE_128			U(16)
+#define AES_KEYSIZE_192			U(24)
+#define AES_KEYSIZE_256			U(32)
+#define AES_IVSIZE			U(16)
+
 TEE_Result stm32_register_authenc(void);
-TEE_Result stm32_register_cipher(void);
+TEE_Result stm32_register_cipher(enum stm32_cipher_ip_id);
 
 #endif /* __DRIVERS_CRYPTO_STM32_COMMON_H */
--- a/core/drivers/crypto/stm32/stm32_saes.c
+++ b/core/drivers/crypto/stm32/stm32_saes.c
@@ -11,11 +11,13 @@
 #include <kernel/boot.h>
 #include <kernel/delay.h>
 #include <kernel/dt.h>
+#include <kernel/huk_subkey.h>
 #include <kernel/mutex.h>
 #include <libfdt.h>
 #include <mm/core_memprot.h>
 #include <stdint.h>
 #include <stm32_util.h>
+#include <string_ext.h>
 #include <utee_defines.h>
 #include <util.h>
 
@@ -235,6 +237,22 @@ static void clear_computation_completed(uintptr_t base)
 	io_setbits32(base + _SAES_ICR, _SAES_I_CC);
 }
 
+static TEE_Result wait_key_valid(vaddr_t base)
+{
+	uint64_t timeout_ref = timeout_init_us(SAES_TIMEOUT_US);
+
+	while (!(io_read32(base + _SAES_SR) & _SAES_SR_KEYVALID))
+		if (timeout_elapsed(timeout_ref))
+			break;
+
+	if (!(io_read32(base + _SAES_SR) & _SAES_SR_KEYVALID)) {
+		DMSG("CCF timeout");
+		return TEE_ERROR_GENERIC;
+	}
+
+	return TEE_SUCCESS;
+}
+
 static TEE_Result saes_start(struct stm32_saes_context *ctx)
 {
 	uint64_t timeout_ref = 0;
@@ -325,6 +343,8 @@ static void saes_write_key(struct stm32_saes_context *ctx)
 
 static TEE_Result saes_prepare_key(struct stm32_saes_context *ctx)
 {
+	TEE_Result res = TEE_ERROR_GENERIC;
+
 	/* Disable the SAES peripheral */
 	io_clrbits32(ctx->base + _SAES_CR, _SAES_CR_EN);
 
@@ -336,14 +356,16 @@ static TEE_Result saes_prepare_key(struct stm32_saes_context *ctx)
 
 	saes_write_key(ctx);
 
+	res = wait_key_valid(ctx->base);
+	if (res)
+		return res;
+
 	/*
 	 * For ECB/CBC decryption, key preparation mode must be selected
 	 * to populate the key.
 	 */
 	if ((IS_CHAINING_MODE(ECB, ctx->cr) ||
 	     IS_CHAINING_MODE(CBC, ctx->cr)) && is_decrypt(ctx->cr)) {
-		TEE_Result res = TEE_SUCCESS;
-
 		/* Select Mode 2 */
 		io_clrsetbits32(ctx->base + _SAES_CR, _SAES_CR_MODE_MASK,
 				SHIFT_U32(_SAES_CR_MODE_KEYPREP,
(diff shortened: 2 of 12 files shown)


core/drivers/crypto/stm32/common.h
```diff
@@ -8,7 +8,42 @@
...
+
+/*
+ * Macro to manage bit manipulation when we work on a local variable
+ * before writing only once to the hardware register.
+ */
+#define CLRBITS(v, bits)		((v) &= ~(bits))
+#define SETBITS(v, bits)		((v) |= (bits))
+#define CLRSETBITS(v, mask, bits)	(*(v) = (*(v) & ~(mask)) | (bits))
+
+#define TOBE32(x)			TEE_U32_BSWAP(x)
```
jenswi-linaro, line 24: We already have `TEE_U32_TO_BIG_ENDIAN()` and `TEE_U32_FROM_BIG_ENDIAN()`. They are perhaps a bit long, but until we add new more convenient generic macros that's what we use.

core/drivers/crypto/stm32/common.h
```diff
@@ -8,7 +8,42 @@
...
+enum stm32_cipher_ip_id {
+	CRYP_IP,
+	SAES_IP,
+};
+
+/*
+ * Macro to manage bit manipulation when we work on a local variable
+ * before writing only once to the hardware register.
+ */
+#define CLRBITS(v, bits)		((v) &= ~(bits))
```
jenswi-linaro, line 20: This macro seems a bit dangerous. What if `v` is a 64-bit value and `bits` is a 32-bit value?
Generally, I think that the quality of helper macros like these would be improved by making them available via `libutils`. They would receive more reviews. Naturally, it's quite likely that they'd have to be changed a bit to make everyone happy too.

core/kernel/huk_subkey.c
```diff
@@ -105,3 +105,12 @@ TEE_Result huk_subkey_derive(enum huk_subkey_usage usage,
 	crypto_mac_free_ctx(ctx);
 	return res;
 }
+
+__weak TEE_Result huk_subkey_derive(enum huk_subkey_usage usage,
+				    const void *const_data,
+				    size_t const_data_len,
+				    uint8_t *subkey, size_t subkey_len)
```
jenswi-linaro, line 112: If we need to do this at all it should probably be `__weak __alias("__huk_subkey_derive")` instead like we do with for instance `crypto_acipher_rsaes_encrypt()`.

core/drivers/crypto/stm32/stm32_saes.c
```diff
@@ -1096,6 +1119,236 @@ TEE_Result stm32_saes_update(struct stm32_saes_context *ctx, bool last_block,
...
+out:
+	free(data);
+	if (res)
+		memzero_explicit(subkey, subkey_size);
+
+	return res;
+}
+
+/* Implement hardware HUK derivation using SAES resources */
+TEE_Result huk_subkey_derive(enum huk_subkey_usage usage,
```
jenswi-linaro, line 1287: I'd expect this to be compatible with the default implementation, but it looks like you get different results depending on implementation. These subkeys are in a way part of different ABIs provided by OP-TEE, if they change depending on the configuration you also break the ABIs.

tbourgoi: Hello,
I'm not sure to fully understand your comment, for you the HW HUK derivation should be the same algorithm and produce the same DHUK as the SW implementation in OPTEE ?

If yes, I don't think it's possible. The SAES does not support HMAC operation and the SW implementation uses HMAC.
Why is it so important that several configurations of OP-TEE derive the HUK the same way ? If a developer choose to change the derivation algorithm the secure storage and any other services must be reinitialised to use the new DHUKs produce with the new algorithm. Data encoded with the DHUK from the previous algorithm are not relevant anymore.

core/drivers/crypto/stm32/common.h
tbourgoi: Ok update in SAES and CRYP driver. I also added a commit to update authenc.c : https://github.com/OP-TEE/optee_os/pull/6168/commits/4c74bce77e568739678b6f64d0773c0aff400213

core/drivers/crypto/stm32/stm32_saes.c
jenswi-linaro: If you're not going to give the same result as the SW version, then you shouldn't fall back to it either or it may lead to a very bad user experience.

tbourgoi: Ok I see your point.
On STM32MP13 the HUK is only accessible by SAES and cannot be read by the SW, therefore the key derivation has to be done with the HW.
By default, we compile all the drivers compatible for the platform and the configuration in the DT probes or not the drivers. 
For SAES, compiling the driver leads to the redefinition of the HUK derivation algorithm. If the peripheral is disabled in the DT then when we will derive a key we will be using an unprobbed peripheral and OP-TEE will panic. Hence the default implementation if SAES is disabled. (without the SW fall back, we would need to know at compile time if SAES will be enabled or not which is a really bad user experience)

Maybe, I try to fix the issue the wrong way and a weak function is not the right solution. 
We could add an ops that by default is pointing to the SW derivation provided by OP-TEE and when we probe SAES we change the ops to point to the SAES HW implementation.
What do you think ? Do you have a other propositions ?

jenswi-linaro: Where do you keep the DTB, compiled into the OP-TEE binary? If so, you know at compile-time whether SAES will be enabled. As long as there's no impact on generic code I'm not too concerned about what you're doing with plat-stm*. But when you need generic changes based on somewhat dubious reasons there's a problem. Having `huk_subkey_derive()` weak is OK, but adding a generic wrapper function too means forcing everyone else to use an extra indirection for no good reason.

tbourgoi: For now yes the DTB is in the OP-TEE binary but in the long term we will externalize the DTB to avoid recompiling the core when we want to change the configuration. So Parsing the DTB during the compilation to see if SAES is enabled or not is a no go for us.
And even now, we don't want to change the compilation of OP-TEE depending of the device tree. The code compiled should work independently of the status of SAES.

Yes there is an extra indirection but GCC will most probably optimize this, so I don't think it has an impact on the performance. And in term of support I suppose it will be similar to  `crypto_acipher_rsaes_encrypt()`.

So there are 2 good reasons to had the indirection : 
- Avoid a custom command line if we disable the HW that implements the HUK derivation.
- Avoid a panic if the HUK derivation is implemented with a disabled HW in the DT.

core/kernel/huk_subkey.c
jenswi-linaro: Please make this:
``` diff
+TEE_Result huk_subkey_derive(enum huk_subkey_usage usage,
+                            const void *const_data, size_t const_data_len,
+                            uint8_t *subkey, size_t subkey_len)
+__weak __alias("__huk_subkey_derive");
```

tbourgoi: Done in https://github.com/OP-TEE/optee_os/pull/6168/commits/0dfcd0a834b0217f4643c7a297db38133c8ed35f

core/drivers/crypto/stm32/common.h
tbourgoi: The register of the CRYP and SAES are only 32 bits so the mismatch 64 bits/32 bits is not a huge concern right now for us but nevertheless it's a good point.
If this kind of helper could be integrated in `libutils`, I need some rework before proposing it.

tbourgoi: `libutils` already provide macros to manage bit field so I rework stm32_saes to use `set_field_u32` and remove the macro in common.h. done in [`68f29e7`](https://github.com/OP-TEE/optee_os/pull/6168/commits/68f29e7fa616a35d8ac77c41ee781bd8d808ce52)

jenswi-linaro:
Please squash in the fixup for "core: allow redefinition of function huk_subkey_derive()".

jenswi-linaro:
For "core: allow redefinition of function huk_subkey_derive()", please apply:

tbourgoi:
There is a warning with `DOUT`, checkpatch thinks it is misspelled with `DOUBT` : 

> WARNING: 'DOUT' may be misspelled - perhaps 'DOUBT'?
> #486: FILE: core/drivers/crypto/stm32/stm32_cryp.c:700:
> `+	 * TEE_U32_FROM_BIG_ENDIAN after reading from DOUT register.`

I think we can count that as a false positive.

the last CHECK : with macro `CLRSETBITS` should disappear with the squash of this fixup : https://github.com/OP-TEE/optee_os/pull/6168/commits/bd3dfc200fff820b5580073c68bf790a2a0d4663.
> 398087c6f crypto: stm32: SAES cipher support
> CHECK: Macro argument reuse 'v' - possible side-effects?
> #332: FILE: core/drivers/crypto/stm32/common.h:22:
> `+#define CLRSETBITS(v, mask, bits)	(*(v) = (*(v) & ~(mask)) | (bits))`

jenswi-linaro:
So far I'm the only one reviewing this PR. I think an STM32 maintainer should review the rest of the patches, in the meantime it's probably a good idea to squash in all the fixups to make it easier for the next reviewer.

core/drivers/crypto/stm32/stm32_saes.c
```diff
@@ -0,0 +1,1427 @@
...
+		write_aligned_block(ctx->base, ctx->extra);
+
+		res = wait_computation_completed(ctx->base);
+		if (res)
+			return res;
+
+		clear_computation_completed(ctx->base);
+
+		ctx->assoc_len += ctx->extra_size * INT8_BIT;
+		ctx->extra_size = 0U;
```
etienne-lms, line 507: s/0U/0/   or `U(0)`

core/drivers/crypto/stm32/stm32_saes.c
```diff
@@ -0,0 +1,1427 @@
...
+ * @param data_size: data size
+ *
+ * @retval 0 if OK.
+ */
+TEE_Result stm32_saes_update_assodata(struct stm32_saes_context *ctx,
+				      uint8_t *data, size_t data_size)
+{
+	TEE_Result res = TEE_SUCCESS;
+	unsigned int i = 0U;
+	uint32_t previous_phase = 0U;
```
etienne-lms, line 703: use 0 or U(0).

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with the few minor comments addressed. I'm fine you to squash the fixup commits.

core/drivers/crypto/stm32/stm32_saes.c
```diff
@@ -0,0 +1,1427 @@
...
+ * @param data_size: tag size
+ *
+ * @retval TEE_SUCCESS if OK.
+ */
+TEE_Result stm32_saes_final(struct stm32_saes_context *ctx, uint8_t *tag,
+			    size_t tag_size)
+{
+	TEE_Result res = TEE_SUCCESS;
+	uint32_t tag_u32[4] = { };
+	uint32_t previous_phase = 0U;
```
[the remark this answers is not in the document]
tbourgoi: done in https://github.com/OP-TEE/optee_os/pull/6168/commits/35271a821c51fd89a42ec95359370daf1233f7f4

tbourgoi:
Hello @jenswi-linaro 

Do you see any other update required to merge the series ?
In my opinion, the CI Code style return a false positive : 

> checkpatch 0cff53b29fe6b5412d8049fc05a5b8ec3201f3b8
>0cff53b29 crypto: stm32: use generic macro in authenc.c
> checkpatch 2d600ca13fcdd17e4b8ed905df9f7[35](https://github.com/OP-TEE/optee_os/actions/runs/5751761424/job/15591244711?pr=6168#step:5:36)539d9d2b3
>2d600ca13 plat-stm32mp1: conf: default enable SAES
> checkpatch 07b8ce2bc0575c2770a1efca9f69fc9e016651f0
>07b8ce2bc drivers: stm32_saes: redefine huk_subkey_derive()
> checkpatch 80834d028b71fc04ed4bed1d152b4a1ad88bd3f8
>80834d028 core: allow redefinition of function huk_subkey_derive()
> checkpatch 35271a821c51fd89a42ec95359[37](https://github.com/OP-TEE/optee_os/actions/runs/5751761424/job/15591244711?pr=6168#step:5:38)0daf1233f7f4
>35271a821 crypto: stm32: SAES cipher support
>WARNING: 'DOUT' may be misspelled - perhaps 'DOUBT'?
>#418: FILE: core/drivers/crypto/stm32/stm32_cryp.c:700:
>+	 * TEE_U32_FROM_BIG_ENDIAN after reading from DOUT register.
> 	                                              ^^^^
>
>total: 0 errors, 1 warnings, 0 checks, 1633 lines checked
> checkpatch c51cf3d199c09fd11492[38](https://github.com/OP-TEE/optee_os/actions/runs/5751761424/job/15591244711?pr=6168#step:5:39)6c8fcd2e843b01fefc
>c51cf3d19 plat-stm32mp1: allocate SAES to secure world
> checkdiff 9c4d09553ce9c[40](https://github.com/OP-TEE/optee_os/actions/runs/5751761424/job/15591244711?pr=6168#step:5:41)a14e6efab34562784b65a2a30 0cff53b29fe6b5[41](https://github.com/OP-TEE/optee_os/actions/runs/5751761424/job/15591244711?pr=6168#step:5:42)2d8049fc05a5b8ec3201f3b8
>WARNING: 'DOUT' may be misspelled - perhaps 'DOUBT'?
>#486: FILE: core/drivers/crypto/stm32/stm32_cryp.c:700:
>+	 * TEE_U32_FROM_BIG_ENDIAN after reading from DOUT register.
> 	                                              ^^^^
>
>total: 0 errors, 1 warnings, 0 checks, 1995 lines checked

Thomas

jenswi-linaro:
Looks like false positives to me. All commits are reviewed, I guess this should be OK to be merged.

jforissier:
Yes, all good, thanks!

PR #6169, opened 2023-07-11
Pan fix

jenswi-linaro:
Fixes the missing PSTATE.PAN bit when returning with `syscall_sys_return()`. Did crash as expected without PR https://github.com/OP-TEE/optee_os/pull/6161.


gagachang:
Dear @jenswi-linaro 

I have a question about `ldelf_syscall_*` functions.
Take `ldelf_syscall_open_bin()` for example, it seems that the `const TEE_UUID *uuid` is located within user memory.
If the kernel wants to access it, it needs user-access functions.
Am I correct?

I got the following stack trace when I tried `CFG_PAN=y`.
I seems that the exception is raised when kernel accesses `uuid`.
```
E/TC:0 0 Core data-abort at address 0x40013fa8 (translation fault)
E/TC:0 0 cpu	#0
E/TC:0 0 cause	000000000000000d epc	000000008e0639e2
E/TC:0 0 tval	0000000040013fa8 status	8000000200006100
E/TC:0 0 ra	000000008e064124 sp	0000000000000000
E/TC:0 0 gp	0000000000000000 tp	0000000000000aff
E/TC:0 0 t0	0000000000000001 t1	0000000000000000
E/TC:0 0 t2	0000000000000000 s0	000000008e0d78d0
E/TC:0 0 s1	000000008e0d77c0 a0	000000008e0d7960
E/TC:0 0 a1	0000000000000025 a2	0000000040013fa8
E/TC:0 0 a3	0000000000000050 a4	0000000040013fa8
E/TC:0 0 a5	0000000040013fa8 a5	0000000040013fa8
E/TC:0 0 a6	000000008e074060 a7	0000000000000000
E/TC:0 0 s2	000000008e061254 s3	000000008e0d7810
E/TC:0 0 s4	000000008e0d7810 s5	000000008e0d7830
E/TC:0 0 s6	000000008e099ac0 s7	000000008e0d7830
E/TC:0 0 s8	000000008e00bb5c s9	000000008e0b5220
E/TC:0 0 s10	000000008e0d79d8 s11	000000008e0d7a7f
E/TC:0 0 t3	0000000000000000 t4	0000000000000000
E/TC:0 0 t5	0000000000000000 t6	0000000000000000
E/TC:0 0 TEE load address @ 0x8e000000
E/TC:0 0 Call stack:
E/TC:0 0  0x8e0639e2 uuid2str at optee_os/lib/libutils/ext/snprintk.c:229
E/TC:0 0  0x8e064120 kprintf at optee_os/lib/libutils/ext/snprintk.c:427
E/TC:0 0  0x8e063958 __vsnprintf at optee_os/lib/libutils/ext/snprintk.c:143
E/TC:0 0  0x8e0638f6 vsnprintk at optee_os/lib/libutils/ext/snprintk.c:131
E/TC:0 0  0x8e064d7c trace_vprintf at optee_os/lib/libutils/ext/trace.c:163
E/TC:0 0  0x8e064b26 trace_printf at optee_os/lib/libutils/ext/trace.c:108
E/TC:0 0  0x8e0145ac ldelf_syscall_open_bin at optee_os/core/kernel/ldelf_syscalls.c:142
E/TC:0 0  0x8e00067a scall_do_call at optee_os/core/arch/riscv/kernel/arch_scall_rv.S:32
E/TC:0 0  0x8e001b82 thread_scall_handler at optee_os/core/arch/riscv/kernel/thread_arch.c:126
E/TC:0 0  0x8e001daa thread_user_ecall_handler at optee_os/core/arch/riscv/kernel/thread_arch.c:175
E/TC:0 0  0x8e0021fe thread_exception_handler at optee_os/core/arch/riscv/kernel/thread_arch.c:251
E/TC:0 0  0x8e0022da thread_trap_handler at optee_os/core/arch/riscv/kernel/thread_arch.c:297
E/TC:0 0  0x8e000358 trap_from_user at optee_os/core/arch/riscv/kernel/thread_rv.S:132
```

jenswi-linaro:
I assume you have removed the enter_user_access() and exit_user_access() from scall_handle_ldelf().
Yes, some fixing is needed for the ldelf syscalls too. I'm currently looking at that, I've also run into the uuid abort when testing some small changes.

gagachang:
> I assume you have removed the enter_user_access() and exit_user_access() from scall_handle_ldelf().

Thanks for hint.
After deeper debugging I found it is our trusted-firmware's issue, not OP-TEE.
Now the exception is gone.

seonghp:
Oh, IIUC so it was not fault of the QEMU (it's always easy to blame others), but the `SPSR_EL1.PAN` was not set properly before `eret` to EL1 after handling `return()` or `panic()` system calls, so the PAN exceptions were not raised. Thank you for catching this.

gagachang:
Do you guys encountered PAN exception on xtest 1033 ?

It seems that in the `pass value` test, TA provides user buffer to system PTA.
(https://github.com/OP-TEE/optee_test/blob/master/ta/supp_plugin/ta_entry.c#L60)
The system PTA will crash when it accesses that user buffer.

jenswi-linaro:
> Did you guys encounter PAN exception on xtest 1033 ?
> 
> It seems that in the `pass value` test, TA provides user buffer to system PTA. (https://github.com/OP-TEE/optee_test/blob/master/ta/supp_plugin/ta_entry.c#L60) The system PTA will crash when it accesses that user buffer.

Oh, another SPSR fabrication problem hiding that. I'll take care of that too in the PR.

gagachang:
> Oh, another SPSR fabrication problem hiding that. I'll take care of that too in the PR.

Thanks!

I just found that xtest 2001 has the same problem.
`socket_open()`, `socket_send()` and `socket_recv()` also access user buffers.

jenswi-linaro:
Yeah, I just noticed that too. I'm taking care of that. Then there's also xtest 6001, I'm looking at that at the moment.

jenswi-linaro:
Rewrote "core: arm64: preserve PSTATE.PAN when making SPSR" to address another SPSR fabrication issue. Now we have a common function to fabricate an SPSR based on PSTATE. Also added a few more patches fixing more PAN problems discovered with the SPSR fabrication problem out of the way.

This update fixes all issues with PAN I'm currently aware of. I'm planning to add more fine-grained PAN for ldelf syscalls too, but that will go into another pull request.

jenswi-linaro:
All checks has passed now.

jenswi-linaro:
Please help to review these PAN fixes.

seonghp:
Thank you for the work! I will take time to further review the PR.

At the first glance, regarding the commit "core: arm64: preserve PSTATE.PAN when making SPSR", IIUC the function `handle_user_mode_panic()`, also setup the SPSR, too, but the commit seems to not take into account this case.

jenswi-linaro:
> At the first glance, regarding the commit "core: arm64: preserve PSTATE.PAN when making SPSR", IIUC the function `handle_user_mode_panic()`, also setup the SPSR, too, but the commit seems to not take into account this case.

Thanks, I'll fix that.



> Regarding "core: add user buffer to tee_invoke_supp_plugin_rpc()", do we have any call-sites where `buf_core` is not `NULL`? Is `data_core` argument left for this added for future-proofing?

Yes, that's for future-proofing.

jenswi-linaro:
Update to take care of the `handle_user_mode_panic()` issue. Let me know if I should squash in the "review" commits.

core/arch/arm/kernel/abort.c
```diff
@@ -379,10 +379,6 @@ static void handle_user_mode_panic(struct abort_info *ai)
 	ai->regs->apiakey_hi = tc->keys.apia_hi;
 	ai->regs->apiakey_lo = tc->keys.apia_lo;
 #endif
-
-	daif = (ai->regs->spsr >> SPSR_32_AIF_SHIFT) & SPSR_32_AIF_MASK;
-	/* XXX what about DAIF_D? */
```
seonghp, line 384: One thing I am not sure about is that if it is okay to remove this line. It seems like `el0_sync_abort` does not touch the memory at the offset `THREAD_ABG_REG_SPSR`, but I am just wondering if I am missing something. Other than this, it looks good to me and I think it can be squashed into previous commit.

jenswi-linaro: `THREAD_ABT_REG_SPSR` is updated at:
https://github.com/OP-TEE/optee_os/blob/f7e4fc1e3c341532465a8e0c602e77d1cc7a3633/core/arch/arm/kernel/thread_a64.S#L1027-L1029

However, there is a problem. We might need to create the new SPSR based on the EL0 SPSR since for instance (some of?) the DAIF bits are masked when taking an exception.

seonghp: I am wondering if we need to keep the previous EL0 SPSR value here. If the `abort_handler()` hits the `handle_user_mode_panic()` then it will never return to the EL0 iiuc, and the previous `spsr` value has been already printed (in `get_fault_type()`).

jenswi-linaro: I changed back to the old way of setting the SPSR, except that I preserve the PAN bit in the same way as the DAIF bits.

jenswi-linaro:
Squashed in the updates and added another sentence to "core: arm64: preserve PSTATE.PAN when making SPSR"

gagachang:
Applied parts of this PR and tested on RISC-V architecture with `CFG_PAN=y`.
Now we can pass `xtest` without PAN exception.
Thanks @jenswi-linaro !
```
26795 subtests of which 0 failed
104 test cases of which 0 failed
0 test cases were skipped
```

jenswi-linaro:
I noticed that we don't have the PAN bit set while at S-EL0 so SPSR_EL1.PAN isn't set. So we need to copy the PSTATE.PAN bit instead.

seonghp:
I am wondering if this PR is on radar of other maintainers for review.

jforissier:
@seonghp it is on my to-do list and hopefully others can review too, but note this is summer time and some people are on vacation :wink:

seonghp:
Good news! Thank you for the confirmation 👍

core/arch/arm/kernel/thread.c
```diff
@@ -412,6 +412,23 @@ void thread_resume_from_rpc(uint32_t thread_id, uint32_t a0, uint32_t a1,
 }
 
 #ifdef ARM64
+static uint64_t spsr_from_pstate(void)
+{
+	uint64_t spsr = SPSR_64(SPSR_64_MODE_EL1, SPSR_64_MODE_SP_EL0, 0);
+
+	spsr = SPSR_64(SPSR_64_MODE_EL1, SPSR_64_MODE_SP_EL0, 0);
```
jforissier, line 419: Duplicated initialization

jenswi-linaro:
Comments addressed, tag applied.

gagachang:
> @gagachang @seonghp any comment?

LGTM, I tested this PR on RISC-V and worked well.

seonghp:
LGTM. Thanks for your work!

PR #6186, opened 2023-07-20
drivers: caam: fix size of cmac update data

clementfaure:
In case of an update operation, the total input data size processed must be a multiple of a block size. The total block size is equal to the input data size and the saved buffer size.

If the reallocation DMA buffer is less than the input data size, buffer length plus saved buffer size need to be adjusted to align on multiple of a block size.

Depending on the memory buffer input configuration, the function caam_dmaobj_sgtbuf_build() might modify the data size to be processed in the loop.

This case happens sometimes on i.MX platforms where the input buffer physical address in above 32 bits. This implies reporting the data size re-ajustment when data is saved in the context buffer.

### DIFF
--- a/core/drivers/crypto/caam/cipher/caam_cipher_mac.c
+++ b/core/drivers/crypto/caam/cipher/caam_cipher_mac.c
@@ -503,6 +503,15 @@ static TEE_Result do_update_cmac(struct drvcrypt_cipher_update *dupdate)
 			if (ret)
 				goto end_cmac;
 
+			/*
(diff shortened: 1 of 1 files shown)


jforissier (commented):
Preferably, s/cmac/CMAC/ in the subject.

clementfaure:
Hi @jforissier, is another ack needed for this patch? I would like cherry pick this patch in our tree .. Thanks!

jforissier:
Hello @clementfaure, not it's fine, I'm merging it right now. Thanks!

PR #6187, opened 2023-07-20
riscv: support page table cache

maroueneboubakri:
To support page table cache we re-use the original implementation. We move `pgt_cache.c` under `core/mm` and we set `CFG_WITH_LPAE ` if `CFG_CORE_LARGE_PHYS_ADDR` is set (which is the case for virt platform).
Flag names are kept as they are despite they are ARM related (EL0, LPAE).


gagachang:
Maybe some comments could be added  into `riscv.mk` to describe the reuse of `CFG_WITH_LPAE`

maroueneboubakri:
WIP

jenswi-linaro:
For "core: mm: move pgt_cache.c to core/mm":

jforissier:
@gagachang does that looks OK to you?

gagachang:
> @gagachang does that looks OK to you?

Yes, LGTM.

If any RISC-V guys' tag is needed, please apply

maroueneboubakri:
@jforissier tags applied.

jforissier:
> @jforissier tags applied.

I believe the tag is applicable to the 1st commit too.

PR #5718, opened 2022-12-19
riscv: driver: implement platform-level interrupt controller (PLIC) driver for RISC-V

maroueneboubakri:
An initial implementation of RISC-V PLIC driver conforming to the specification.
`CFG_RISCV_PLIC` flag allows building it or not for platforms with custom PLIC IP.

PR #6189, opened 2023-07-24
RISC-V: Refine PLIC related code

gagachang:
This PR renames some existed functions.
Also it refines PLIC hartid to PLIC context, since the PLIC specification uses context instead of HART.

Hello, @maroueneboubakri 
Please review this patch, thanks!


core/arch/riscv/kernel/boot.c
```diff
@@ -120,7 +120,7 @@ __weak void primary_init_intc(void)
 }
 
 /* May be overridden in plat-$(PLATFORM)/main.c */
-__weak void secondary_init_intc(void)
+__weak void main_secondary_init_intc(void)
```
jforissier, line 123: This name does not make much sense IMO, what is the main word supposed to represent? I'd rather do the opposite replacement in all Arm platforms.

gagachang: It is originated from traditional ARM code

I can see @maroueneboubakri  just rename `main_secondary_init_gic()` to `main_secondary_init_intc()` in commit https://github.com/OP-TEE/optee_os/pull/5718/commits/7cfd2b24191d9fc63993c571b6f21401656580fc

jforissier: That's commit https://github.com/OP-TEE/optee_os/commit/ef50391ee92b850d6a6f9280969ed4a9857480af in the master branch actually. Well I wasn't careful enough when reviewing, because I agreed with the description which states "Rename secondary_init_gic() to secondary_init_intc()", but it is obviously not what the code does.

I still believe `main_secondary_*()` is a bad name unless someone explains what is it supposed to mean.

gagachang: Find a commit 7 years ago https://github.com/OP-TEE/optee_os/commit/05efe1e1222d69cfbb68a1aeca9a9af2c310a7aa

gagachang: Resolved in https://github.com/OP-TEE/optee_os/pull/6207

core/drivers/plic.c
```diff
@@ -35,23 +35,38 @@
...
 
+/*
+ * We assume that each hart has M-mode and S-mode, so the contexts look like:
+ * PLIC context 0 is hart 0 M-mode
+ * PLIC context 1 is hart 0 S-mode
+ * PLIC context 2 is hart 1 M-mode
+ * PLIC context 3 is hart 1 S-mode
+ * ...
+ */
+static uint32_t plic_get_context(size_t hartid)
```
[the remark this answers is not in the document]
gagachang: Good idea. thanks!

maroueneboubakri:

gagachang:
Tags applied

PR #6195, opened 2023-07-26
pta: attestation: fixes

jforissier:
Several fixes for the attestation PTA. One commit excludes a region that should not be hashed, which was introduced by a commit related to PAN. The other ones fix invalid accesses to user space memory (TA memory pages and memref parameters) when `CFG_PAN=y`.

### DIFF
--- a/core/include/kernel/pseudo_ta.h
+++ b/core/include/kernel/pseudo_ta.h
@@ -7,7 +7,9 @@
 
 #include <assert.h>
 #include <compiler.h>
+#include <config.h>
 #include <kernel/tee_ta_manager.h>
+#include <kernel/user_ta.h>
 #include <scattered_array.h>
 #include <tee_api_types.h>
 #include <user_ta_header.h>
@@ -59,6 +61,39 @@ static inline struct pseudo_ta_ctx *to_pseudo_ta_ctx(struct ts_ctx *ctx)
 
 TEE_Result tee_ta_init_pseudo_ta_session(const TEE_UUID *uuid,
 			struct tee_ta_session *s);
+
+/*
+ * Helper functions for PTAs to support calls from a TA when CFG_PAN=y
+ */
+
+static inline bool is_caller_ta_with_pan(void)
+{
+	struct ts_session *s = NULL;
+
+	if (!IS_ENABLED(CFG_PAN))
+		return false;
+	s = ts_get_calling_session();
+	return s && is_user_ta_ctx(s->ctx);
+}
+
+/*
+ * If caller is a TA and PAN is enabled, allocate bounce buffers for each
+ * memref in @params and build @bparams, then make *@oparams point to @bparams.
+ * Otherwise just make *@oparams point to @params.
+ */
+TEE_Result to_bounce_params(uint32_t param_types,
+			    TEE_Param params[TEE_NUM_PARAMS],
+			    TEE_Param bparams[TEE_NUM_PARAMS],
+			    TEE_Param **oparams);
+
+/*
+ * If @eparams == @bparams, copy data from @bparams to @params. Otherwise, do
+ * nothing.
+ */
+TEE_Result from_bounce_params(uint32_t param_types,
+			      TEE_Param params[TEE_NUM_PARAMS],
+			      TEE_Param bparams[TEE_NUM_PARAMS],
+			      TEE_Param *eparams);
 
 #endif /* KERNEL_PSEUDO_TA_H */
 
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -483,10 +483,11 @@ static TEE_Result sign_buffer(uint8_t *buf, size_t buf_sz, uint8_t *nonce,
  */
 static bool is_region_valid(struct vm_region *r)
 {
-	uint32_t skip_flags = VM_FLAG_EPHEMERAL | VM_FLAG_PERMANENT |
-			      VM_FLAG_LDELF;
+	uint32_t dontwant = VM_FLAG_EPHEMERAL | VM_FLAG_PERMANENT |
+			    VM_FLAG_LDELF;
+	uint32_t want = VM_FLAG_READONLY;
 
-	return !(r->flags & skip_flags || r->attr & TEE_MATTR_UW);
+	return ((r->flags & want) == want && !(r->flags & dontwant));
 }
 
 /*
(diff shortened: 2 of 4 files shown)


jenswi-linaro:

jenswi-linaro:
> Adds missing enter_user_access()/exit_user_access() when the PTA
accesses the TA memory for hashing purposes, and use bounce buffers
for reading the nonce and writing the output from/to user space.

> FIXME: shouldn't the bounce buffers be involved in some common code
so that all PTAs can benefit? Whenever the caller of a PTA is a TA
(as opposed to a CA or another PTA) and PAN is enabled, then bounce
buffers must be used.

You just answered that question above. There are situations where bounce buffers aren't the solution to the problem so it can't be applied by default.

core/pta/attestation.c
```diff
@@ -620,6 +628,8 @@ static TEE_Result cmd_hash_ta_memory(uint32_t param_types,
 	struct user_mode_ctx *uctx = NULL;
 	TEE_Result res = TEE_SUCCESS;
 	struct ts_session *s = NULL;
+	uint8_t *nonce_bbuf = NULL;
```
jforissier, line 631: This is the part which I think should be in `core/kernel/pseudo_ta.c` rather than here, I suppose in `copy_in_param()` / `update_out_param()`? Indeed this PTA also has issues in other commands, for example if `cmd_get_pubkey()` is invoked by a TA instead of a CA, and `CFG_PAN=y`, it will crash. @jenswi-linaro what do you think?

jforissier:
> > Adds missing enter_user_access()/exit_user_access() when the PTA
> > accesses the TA memory for hashing purposes, and use bounce buffers
> > for reading the nonce and writing the output from/to user space.
> 
> > FIXME: shouldn't the bounce buffers be involved in some common code
> > so that all PTAs can benefit? Whenever the caller of a PTA is a TA
> > (as opposed to a CA or another PTA) and PAN is enabled, then bounce
> > buffers must be used.
> 
> You just answered that question above. There are situations where bounce buffers aren't the solution to the problem so it can't be applied by default.

Yet passing a memref to a PTA is a very common case, and by default PTAs can be called by CAs, PTAs and TAs, so I think there should be a way to automatically use a bounce buffer when the caller is a TA and PAN is enabled. Otherwise it will lead to code duplication and it will be error prone.

jforissier:
Another question: when a PAN fault occurs, could we avoid crashing the whole TEE and instead just panic the caller TA?

jenswi-linaro:
> Another question: when a PAN fault occurs, could we avoid crashing the whole TEE and instead just panic the caller TA?

What if there are locks that would lead to a deadlock a bit later? It might be more trouble than it's worth because each time a PAN fault occurs it is a bug that should be fixed. I'm not sure if it can be used as an attack vector, but we can be sure that it doesn't improve security.

jenswi-linaro:
> > > Adds missing enter_user_access()/exit_user_access() when the PTA
> > > accesses the TA memory for hashing purposes, and use bounce buffers
> > > for reading the nonce and writing the output from/to user space.
> > 
> > 
> > > FIXME: shouldn't the bounce buffers be involved in some common code
> > > so that all PTAs can benefit? Whenever the caller of a PTA is a TA
> > > (as opposed to a CA or another PTA) and PAN is enabled, then bounce
> > > buffers must be used.
> > 
> > 
> > You just answered that question above. There are situations where bounce buffers aren't the solution to the problem so it can't be applied by default.
> 
> Yet passing a memref to a PTA is a very common case, and by default PTAs can be called by CAs, PTAs and TAs, so I think there should be a way to automatically use a bounce buffer when the caller is a TA and PAN is enabled. Otherwise it will lead to code duplication and it will be error prone.

We could add a helper function to automatically deal with bounce buffers based on some parameters. The point is that it must be possible to exclude bounce buffering for some memrefs when calling certain PTA functions. Unless we want to change how `pseudo_ta_register()` is used this should probably be done in the PTA invoke function.

jforissier:
> > Another question: when a PAN fault occurs, could we avoid crashing the whole TEE and instead just panic the caller TA?
> 
> What if there are locks that would lead to a deadlock a bit later?

Difficult thing to deal with, indeed.

> It might be more trouble than it's worth because each time a PAN fault occurs it is a bug that should be fixed. 

Agreed.

> I'm not sure if it can be used as an attack vector, but we can be sure that it doesn't improve security.

OK. But as I said if there is no way to make the case 'TA with PAN calling a PTA and passing memrefs" work by default (even in a sub-optimal way), I'm afraid we will keep opening DOS attack opportunities in the TEE core, that's pretty much guaranteed :-/ User space should never be able to crash the kernel...

jforissier:
> We could add a helper function to automatically deal with bounce buffers based on some parameters. The point is that it must be possible to exclude bounce buffering for some memrefs when calling certain PTA functions. Unless we want to change how `pseudo_ta_register()` is used this should probably be done in the PTA invoke function.

OK let me try the helper function route. Then we'll see if more factoring is possible/desirable.

core/include/kernel/pseudo_ta.h
```diff
@@ -60,5 +62,44 @@ static inline struct pseudo_ta_ctx *to_pseudo_ta_ctx(struct ts_ctx *ctx)
...
+
+static inline bool is_caller_ta_with_pan(void)
+{
+	struct ts_session *s = NULL;
+	struct user_mode_ctx *uctx = NULL;
+
+	if (!IS_ENABLED(CFG_PAN))
+		return false;
+	s = ts_get_calling_session();
+	if (!s)
```
jenswi-linaro, line 77: The rest of this function can be replaced with `return s && is_user_ta_ctx(s->ctx)`;

jenswi-linaro (commented):
The extra `bparams` uses quite a bit of stack only to maybe hold four pointers.
While looking at this I got an idea of how we can move almost all of this into generic pre- and post-processing of a PTA invoke. I'll share something shortly so we can compare the approaches.

core/include/kernel/pseudo_ta.h
jforissier: Done

jenswi-linaro:
> The extra `bparams` uses quite a bit of stack only to maybe hold four pointers. While looking at this I got an idea of how we can move almost all of this into generic pre- and post-processing of a PTA invoke. I'll share something shortly so we can compare the approaches.

I ran into a few rabbit holes and it will take two weeks before I'm able to resume this. The approach here fixes the error. I believe we can improve PAN support further, but that's not in the scope of this PR.

Please apply:

PR #6211, opened 2023-08-01
core: small fixes

ghost:
### There are two points of correction:

1. Issue with DH (Diffie-Hellman):
    The DH secret, due to its mechanism, becomes a value less than the key length of the Private key. Consequently, when obtaining the maximum size of the secret from the current key length of the Public key, the secret length falls short.

1. Typo in the ECC Crypt API header file:
    According to the OP-TEE specification, the size passed as an argument to gen_keypair should be in terms of bit length. However, the variable name for the argument is bytes.

### DIFF
--- a/core/drivers/crypto/crypto_api/include/drvcrypt_acipher.h
+++ b/core/drivers/crypto/crypto_api/include/drvcrypt_acipher.h
@@ -157,7 +157,7 @@ struct drvcrypt_ecc {
 	/* Free ECC public key */
 	void (*free_publickey)(struct ecc_public_key *key);
 	/* Generates the ECC keypair */
-	TEE_Result (*gen_keypair)(struct ecc_keypair *key, size_t size_bytes);
+	TEE_Result (*gen_keypair)(struct ecc_keypair *key, size_t size_bits);
 	/* ECC Sign a message and returns the signature */
 	TEE_Result (*sign)(struct drvcrypt_sign_data *sdata);
 	/* ECC Verify a message's signature */


jenswi-linaro:
Since these two patches are fixes, please add "Fixes: " tags too, pointing out the commit which is fixed.

jenswi-linaro:
Please fix the checkpatch complaints.
If you're not sure how to proceed, please let us know and we'll help you.

TE-0000106830:
@jenswi-linaro 
Than you for your kind follow. I'm trying to contribute with Contribute Section in OP-TEE Docs but I may miss something to do because this is my first contribute in my life.

I fixed commit message first. And I'll fix `checkpatch complaints` .

jenswi-linaro:
Please fix up the commit messages along something like this:
```
drivers: crypto: fix to extract DH secret length from modulus

Prior to this patch, the DH secret, due to its mechanism, becomes a value
less than the key length of the Private key. Consequently, when obtaining
the maximum size of the secret from the current key length of the Public
key, the secret length falls short. So change this to extract DH secret
length from modulus instead of public key length.

Fixes: f6e2b9e2d1a2 ("drivers: crypto: implement crypto driver - DH")
Signed-off-by: Katsunori Kikuchi <Katsunori.Kikuchi@sony.com>
```

and this:
```
core: crypto_api: fixes typo "bytes" to "bits"

Fixes a typo for *gen_keypair() in struct drvcrypt_ecc where the last
parameter was "size_bytes" while the value represents bits, so fix this by
renaming it to "size_bits".

Fixes: d29cd2efcd46 ("core: driver: generic resources for crypto device driver - ECC")
Signed-off-by: Katsunori Kikuchi <Katsunori.Kikuchi@sony.com>
```

TE-0000106830:
@jenswi-linaro
Thank you for your support. I misunderstood the "fixes:" tag usage.
And can I use the commit message sample you gave? Because the sample seems perfect and I don't think that I can write commit messages better than the ones.

jenswi-linaro:
> @jenswi-linaro Thank you for your support. I misunderstood the "fixes:" tag usage.

No problem.

> And can I use the commit message sample you gave? Because the sample seems perfect and I don't think that I can write commit messages better than the ones.

Yes, feel free to use them as is.

TE-0000106830:
Thank you for your kind support.
I just pushed commits with your perfect sample commit messages!

TE-0000106830:
Thank you for your continuous support.
I jsut add it to the commits.

PR #6216, opened 2023-08-02
crypto: add a new pbkdf2 implementation

loubaihui:
We add a new pbkdf2 implementation.

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -1038,3 +1038,8 @@ CFG_TA_OPTEE_CORE_API_COMPAT_1_1 ?= n
 #              - PerInstance/AttestationTest#
 # Note that this violates GP requirements of HMAC size range.
 CFG_HMAC_64_1024_RANGE ?= n
+
+# Enable a hardware pbkdf2 function
+# By default use standard pbkdf2 implementation
+CFG_CRYPTO_HW_PBKDF2 ?= n
+$(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
--- a/core/tee/sub.mk
+++ b/core/tee/sub.mk
@@ -21,7 +21,9 @@ srcs-y += entry_std.c
 srcs-y += tee_cryp_utl.c
 srcs-$(CFG_CRYPTO_HKDF) += tee_cryp_hkdf.c
 srcs-$(CFG_CRYPTO_CONCAT_KDF) += tee_cryp_concat_kdf.c
+ifneq ($(CFG_CRYPTO_HW_PBKDF2),y)
 srcs-$(CFG_CRYPTO_PBKDF2) += tee_cryp_pbkdf2.c
+endif
 
 ifeq ($(CFG_WITH_USER_TA),y)
 srcs-y += tee_obj.c


jenswi-linaro:
How about just:
``` diff
--- a/core/tee/sub.mk
+++ b/core/tee/sub.mk
@@ -21,7 +21,9 @@ srcs-y += entry_std.c
 srcs-y += tee_cryp_utl.c
 srcs-$(CFG_CRYPTO_HKDF) += tee_cryp_hkdf.c
 srcs-$(CFG_CRYPTO_CONCAT_KDF) += tee_cryp_concat_kdf.c
+ifneq($(CFG_CRYPTO_HW_PBKDF2),y)
 srcs-$(CFG_CRYPTO_PBKDF2) += tee_cryp_pbkdf2.c
+endif
 
 ifeq ($(CFG_WITH_USER_TA),y)
 srcs-y += tee_obj.c
```
instead?

We should also document the `CFG_CRYPTO_HW_PBKDF2` flag, probably in `mk/config.mk`

loubaihui:
> How about just:
> 
> ```diff
> --- a/core/tee/sub.mk
> +++ b/core/tee/sub.mk
> @@ -21,7 +21,9 @@ srcs-y += entry_std.c
>  srcs-y += tee_cryp_utl.c
>  srcs-$(CFG_CRYPTO_HKDF) += tee_cryp_hkdf.c
>  srcs-$(CFG_CRYPTO_CONCAT_KDF) += tee_cryp_concat_kdf.c
> +ifneq($(CFG_CRYPTO_HW_PBKDF2),y)
>  srcs-$(CFG_CRYPTO_PBKDF2) += tee_cryp_pbkdf2.c
> +endif
>  
>  ifeq ($(CFG_WITH_USER_TA),y)
>  srcs-y += tee_obj.c
> ```
> 
> instead?
> 
> We should also document the `CFG_CRYPTO_HW_PBKDF2` flag, probably in `mk/config.mk`

Sounds great, according to your suggestions, we made the revision.

core/tee/tee_cryp_pbkdf2.c
```diff
@@ -112,3 +112,4 @@ TEE_Result tee_cryp_pbkdf2(uint32_t hash_id, const uint8_t *password,
 	crypto_mac_free_ctx(hmac_parms.ctx);
 	return res;
 }
+
```
jenswi-linaro, line 115: Unrelated change.

core/crypto/sub.mk
```diff
@@ -42,4 +42,5 @@ srcs-$(CFG_CRYPTO_ECB) += sm4-ecb.c
 srcs-$(CFG_CRYPTO_CBC) += sm4-cbc.c
 srcs-$(CFG_CRYPTO_CTR) += sm4-ctr.c
 srcs-$(CFG_CRYPTO_XTS) += sm4-xts.c
+srcs-$(CFG_CRYPTO_HW_PBKDF2) += pbkdf2_hw.c
```
jenswi-linaro, line 45: I don't think `pbkdf2_hw.c` or `pbkdf2_support.h` is needed at all. The hardware implementation can just implement the `tee_cryp_pbkdf2()`

loubaihui: Is that mean:
If the two files are not needed, the hardware implementation also put in `tee_cryp_pbkdf2.c` and use CFG_CRYPTO_HW_PBKDF2 to distinguish?

jenswi-linaro: No, a hardware non-stubbed implementation is provided in some other file.
It doesn't make sense to use a stubbed hardware implementation when a working software version is available.

loubaihui: The hardware implementation will be different from the current software, we will use different hash algorithm etc.
So, I think the  `pbkdf2_hw.c` and `pbkdf2_support.h` is necessary, then specific implementation of `hw_pbkdf2()` function will be added in `core/ drivers/crypto` @jenswi-linaro

jenswi-linaro: Do you mean that the same input will yield different results? I don't see how a stubbed implementation helps against that.

loubaihui: > Do you mean that the same input will yield different results? I don't see how a stubbed implementation helps against that.

I'm sorry for misunderstanding your previous comment. Now, I got it and the PR is ready for another round of review.

mk/config.mk
```diff
@@ -1038,3 +1038,7 @@ CFG_TA_OPTEE_CORE_API_COMPAT_1_1 ?= n
 #              - PerInstance/AttestationTest#
 # Note that this violates GP requirements of HMAC size range.
 CFG_HMAC_64_1024_RANGE ?= n
+
+#Enable a hardware pbkdf2 function
```
jenswi-linaro, line 1042: Please add a space after `#`, the same on the line below.

mk/config.mk
```diff
@@ -1038,3 +1038,7 @@ CFG_TA_OPTEE_CORE_API_COMPAT_1_1 ?= n
...
+#By default use standard pbkdf2 implementation
+CFG_CRYPTO_HW_PBKDF2 ?= n
```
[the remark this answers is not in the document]
jforissier: Yes, I agree

mk/config.mk
loubaihui: Modifications are made.

mk/config.mk
```diff
@@ -1038,3 +1038,7 @@ CFG_TA_OPTEE_CORE_API_COMPAT_1_1 ?= n
 #              - PerInstance/AttestationTest#
 # Note that this violates GP requirements of HMAC size range.
 CFG_HMAC_64_1024_RANGE ?= n
+
+# Enable a hardware pbkdf2 function
+CFG_CRYPTO_HW_PBKDF2 ?= y
+$(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
```
jenswi-linaro, line 1044: Please add a newline character at the end of the file to avoid this ugly warning from git.

mk/config.mk
```diff
@@ -1038,3 +1038,7 @@ CFG_TA_OPTEE_CORE_API_COMPAT_1_1 ?= n
 #              - PerInstance/AttestationTest#
 # Note that this violates GP requirements of HMAC size range.
 CFG_HMAC_64_1024_RANGE ?= n
+
+# Enable a hardware pbkdf2 function
+CFG_CRYPTO_HW_PBKDF2 ?= y
```
jenswi-linaro, line 1043: The default value should be `n`.

loubaihui: OK

jenswi-linaro (commented):

loubaihui:
@jenswi-linaro @jforissier

loubaihui:
> @loubaihui the code looks good to me so please add:
> 
> 
> However I would like the commit description changed because it does not add a new implementation, just a config flag to support a hardware implementation. How about:
> 
> ```
> crypto: introduce CFG_CRYPTO_HW_PBKDF2
> 
> Add a new configuration flag to support hardware implementation of
> PBKDF2.
> ```

Ok, thanks! @jforissier

jforissier:
The CI error with "QEMUv8, Xen" is unrelated to this PR, and probably caused by insufficient disk space. This will be addressed separately. Therefore I am merging this, thanks @loubaihui.

PR #6217, opened 2023-08-02
core: drivers: add HiSilicon trng gp implement.

loubaihui:
We add a new trng module which is match with  HiSilicon hardware.


core/arch/arm/plat-d06/conf.mk
```diff
@@ -17,6 +17,8 @@ $(call force,CFG_ARM_GICV3,y)
 $(call force,CFG_LPAE_ADDR_SPACE_BITS,48)
 $(call force,CFG_LPC_UART,y)
 
+CFG_HISILICON_PLAT_ACC ?= y
```
jenswi-linaro, line 20: What's the logic behind this name?

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+			return TEE_ERROR_BAD_STATE;
+		}
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	size_t buf_pos = 0, fifo_pos = 0;
```
jenswi-linaro, line 58: Only one variable per line please.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if (!buf || !len) {
+		EMSG("Invalid input parameter!\n");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	while (buf_pos < len) {
+		if (fifo_pos == 0) {
+			ret = trng_read(&fifo.val);
```
jenswi-linaro, line 75: I believe this call needs to be protected by a spinlock.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	size_t buf_pos = 0, fifo_pos = 0;
+	uint8_t *out = (uint8_t *)buf;
+	union trng_fifo fifo = { };
```
jenswi-linaro, line 60: Could this be static to save a partial read between calls? Compare with other implementations of this function.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	return ret;
+}
+
+static TEE_Result trng_init(void)
+{
+	struct hisi_trng *trng;
+	uint32_t val;
+
+	IMSG("TRNG driver init start, version %s!\n", TRNG_MODULE_VERSION);
+	trng = (struct hisi_trng *)malloc(sizeof(struct hisi_trng));
```
jenswi-linaro, line 93: Cast is not needed.

How about skipping the malloc() have `trng_dev` as a `struct hisi_trng` instead of a pointer?

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+#define HTRNG_DATA_ST_REG 0x1B4
+#define HTRNG_VERSION_REG 0x1B8
+#define HTRNG_VERSION_V2 GENMASK_32(31, 0)
+#define HTRNG_VERSION_V3 0x30
+#define HTRNG_DRBG_ALG_SM4 BIT(1)
+#define HTRNG_RANDATA_ERR_ST GENMASK_32(15, 1)
+#define HTRNG_BYTES 4
+
+struct hisi_trng {
+	uintptr_t base;
```
jenswi-linaro, line 23: This is where we use a `vaddr_t` instead.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+union trng_fifo {
+	uint32_t val;
+	uint8_t data_byte[HTRNG_BYTES];
+};
+
+static struct hisi_trng *trng_dev;
+
+static TEE_Result trng_read(uint32_t *val)
+{
+	uint32_t status;
```
jenswi-linaro, line 36: Please initialize this.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+		out[buf_pos++] = fifo.data_byte[fifo_pos++];
+		fifo_pos %= HTRNG_BYTES;
+	}
+
+	return ret;
+}
+
+static TEE_Result trng_init(void)
+{
+	struct hisi_trng *trng;
```
jenswi-linaro, line 89: These needs to be initialized.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	uint32_t val;
+
+	IMSG("TRNG driver init start, version %s!\n", TRNG_MODULE_VERSION);
+	trng = (struct hisi_trng *)malloc(sizeof(struct hisi_trng));
+	if (!trng) {
+		EMSG("Fail to malloc trng dev!\n");
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+
+	if (cpu_mmu_enabled()) {
```
jenswi-linaro, line 99: We should be able to depend on MMU to be enabled at this stage.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	size_t buf_pos = 0, fifo_pos = 0;
+	uint8_t *out = (uint8_t *)buf;
+	union trng_fifo fifo = { };
+	TEE_Result ret = 0;
+
+	if (!trng_dev) {
+		EMSG("No valid TRNG device!\n");
```
jenswi-linaro, line 64: The `\n` is not needed, it's added automatically by `EMSG()` and friends. Please remove the `\n` at the end of all logs in this file.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	} else
+		IMSG("TRNG hardware version V2!\n");
+
+	trng_dev = trng;
+	IMSG("TRNG driver init done!\n");
+
+	return TEE_SUCCESS;
+}
+
+early_init(trng_init);
```
jenswi-linaro, line 124: Please add a newline at the end of this line.

xiaoxuZeng:
> We add a new trng module which is match witch Huawei hardware.

"Huawei hardware" sounds a bit strange.

jenswi-linaro:
Please let us know (with an explicit comment or we might miss it) when the PR is ready for another round of review.

loubaihui:
> Please let us know (with an explicit comment or we might miss it) when the PR is ready for another round of review.

Actually, we haven't got ready for another new round of review yet, because we do not fully understand the following two reviews you have listed. Could you please provide more detailed explanations?
1、
image
2、
image

jenswi-linaro:
Please post the comments instead of a screenshot of pending comments.

jenswi-linaro:
Anyway, `trng_read()` first reads some register and then another depending on the output. What if two threads are calling this function in parallel, what you happen?

Regarding the call to `cpu_mmu_enabled()`, since `trng_init()` is called after the MMU has been enabled, we can assume that the MMU is enabled.

loubaihui:
> Please post the comments instead of a screenshot of pending comments.



> Anyway, `trng_read()` first reads some register and then another depending on the output. What if two threads are calling this function in parallel, what do you think will happen?
> 
> Regarding the call to `cpu_mmu_enabled()`, since `trng_init()` is called after the MMU has been enabled, we can assume that the MMU is enabled.

Can the `io_read32` interface guarantee an atomic read operation?

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+static TEE_Result trng_read(uint32_t *val)
+{
+	uint32_t status;
+
+	*val = io_read32(trng_dev->base + HTRNG_RANDATA_REG);
+	if (*val == 0) {
+		if (trng_dev->ver == HTRNG_VERSION_V3) {
+			status = io_read32(trng_dev->base + HTRNG_DATA_ST_REG);
+			if (status & HTRNG_RANDATA_ERR_ST) {
+				EMSG("TRNG hardware error 0x%x happened!\n",
```
etienne-lms, line 43: `EMSG("TRNG hardware error 0x%"PRIx32" happened", ...`.

You can remove both bang `!` and new-line \n` termination.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	size_t buf_pos = 0, fifo_pos = 0;
+	uint8_t *out = (uint8_t *)buf;
+	union trng_fifo fifo = { };
+	TEE_Result ret = 0;
```
etienne-lms, line 61: init to a `TEE_xxx` macro. either `TEE_SUCCESS` or `TEE_ERROR_GENERIC`.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+		}
+	} else
+		trng->base = HISI_TRNG_BASE;
+
+	trng->ver = io_read32(trng->base + HTRNG_VERSION_REG);
+	if (trng->ver == HTRNG_VERSION_V3) {
+		IMSG("TRNG hardware version V3!\n");
+		val = io_read32(trng->base + HTRNG_ALG_MODE_REG);
+		val |= HTRNG_DRBG_ALG_SM4;
+		io_write32(trng->base + HTRNG_ALG_MODE_REG, val);
```
etienne-lms, line 114: you can use `io_setbits32(trng->base + HTRNG_ALG_MODE_REG, HTRNG_DRBG_ALG_SM4)`

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	} else
+		IMSG("TRNG hardware version V2!\n");
+
+	trng_dev = trng;
+	IMSG("TRNG driver init done!\n");
```
etienne-lms, line 119: Prefer `DMSG()` for driver generic init info. `IMSG()` is intended to be short at boot time.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,124 @@
...
+	uint32_t val;
+	uint8_t data_byte[HTRNG_BYTES];
+};
+
+static struct hisi_trng *trng_dev;
+
+static TEE_Result trng_read(uint32_t *val)
+{
+	uint32_t status;
+
```
etienne-lms, line 37: The function entry should test `trng_dev` and exit with an error NULL.

etienne-lms (commented):
Some added comments.

> Can the io_read32() guarantee an atomic read operation

The issue is with the error recorver path that rely on the `io_read32()` output. It should be protected. 2 threads could call the function and conflict with HW expected behaviour.

loubaihui:
Many modifications are submitted, the PR is ready for another round of review. @jenswi-linaro 
Due to the PR [https://github.com/OP-TEE/optee_os/pull/6297](url) is merged , we use `IO_READ32_POLL_TIMEOUT` to read random number.

jenswi-linaro:
We don't do merge commits in PRs, please drop the merge commit and rebase instead.

loubaihui:
> We don't do merge commits in PRs, please drop the merge commit and rebase instead.

Ok

core/arch/arm/plat-d06/conf.mk
```diff
@@ -26,3 +26,5 @@ CFG_TZDRAM_START ?=    0x20C0000000
 CFG_TZDRAM_SIZE ?=     0x32000000
 CFG_SHMEM_START ?=     0x50000000
 CFG_SHMEM_SIZE ?=      0x04000000
+
+CFG_HISILICON_PLAT_ACC ?= y
```
jenswi-linaro, line 30: This is unused

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,84 @@
...
+	return ret;
+}
+
+static TEE_Result trng_init(void)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+
+	IMSG("TRNG driver init start.");
+	trng_dev = (struct hisi_trng *)malloc(sizeof(struct hisi_trng));
```
jenswi-linaro, line 70: No need to cast the result from `malloc()`. You need to check the result from the memory allocation though. You should perhaps use `calloc()` instead since you don't initialize `trng_dev->ver`, if you need dynamic memory allocation at all.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,84 @@
...
+
+	trng->base = (uintptr_t)phys_to_virt_io(HISI_TRNG_BASE, HISI_TRNG_SIZE);
+	if (!trng_dev->base) {
+		EMSG("Fail to get trng io_base.");
+		free(trng_dev);
```
jenswi-linaro, line 75: `trng_dev` should be set to NULL also.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,84 @@
...
+	uint32_t val;
+	uint8_t data_byte[HTRNG_BYTES];
+};
+
+static struct hisi_trng *trng_dev;
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	uint8_t *out = (uint8_t *)buf;
```
jenswi-linaro, line 32: Casting is not needed.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if (!buf || !len) {
+		EMSG("Invalid input parameter.");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	while (buf_pos < len) {
+		if (fifo_pos == 0) {
+			if (IO_READ32_POLL_TIMEOUT(trng_dev->base +
```
jenswi-linaro, line 49: Doesn't this access need to be serialized in case `hw_get_random_bytes()` called by multiple cores at the same time?

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,84 @@
...
+	uint8_t data_byte[HTRNG_BYTES];
+};
+
+static struct hisi_trng *trng_dev;
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	uint8_t *out = (uint8_t *)buf;
+	static union trng_fifo fifo = { };
```
jenswi-linaro, line 33: You use dynamic memory allocation to keep track of the base address which is more or less static, but static memory allocation to keep the dynamic part of the state.

I'd prefer a more direct approach with global:
``` C
static struct hisi_trng trng_dev; /* or perhaps just "static vaddr_t trng_base" */
static union trng_fifo trng_fifo;
static unsigned int trng_lock = SPINLOCK_UNLOCK
```

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,81 @@
...
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+
+	IMSG("TRNG driver init start.");
+	trng_dev = calloc(1, sizeof(struct hisi_trng));
+
+	trng->base = (uintptr_t)phys_to_virt_io(HISI_TRNG_BASE, HISI_TRNG_SIZE);
+	if (!trng_dev->base) {
+		EMSG("Fail to get trng io_base.");
+		trng_dev = NULL;
```
etienne-lms, line 71: clear `trng_dev` after it's freed

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,81 @@
...
+		return TEE_ERROR_NOT_SUPPORTED;
+	}
+
+	if (!buf || !len) {
+		EMSG("Invalid input parameter.");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	do {
+		ret = IO_READ32_POLL_TIMEOUT(trng_dev->base + HTRNG_RANDATA_REG
```
etienne-lms, line 40: `IO_READ32_POLL_TIMEOUT()` returns a boolean value, not a `TEE_Result` value.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,81 @@
...
+					    val, val, POLL_PERIOD, POLL_TIMEOUT)
+		if (ret) {
+			EMSG("Hardware busy.");
+			return TEE_ERROR_BUSY;
+		}
+		if (len - current_len >= HTRNG_BYTES) {
+			mencpy((uint8_t)buf + current_len, &val, HTRNG_BYTES);
```
etienne-lms, line 47: typo: s/mencpy/memcpy/

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,81 @@
...
+			current_len += HTRNG_BYTES;
+
+			if (current_len == len)
+				return TEE_SUCCESS;
+			continue;
+		}
+		memcpy((uint8_t)buf + current_len, &val, len - current_len);
+		current_len = len;
```
etienne-lms, line 55: a `break` instruction would be more explicit.
The whole seems easier to read as:
```c
	while (current_len < len) {
		ret = IO_READ32_POLL_TIMEOUT(trng_dev->base + HTRNG_RANDATA_REG
					    val, val, POLL_PERIOD, POLL_TIMEOUT)
		if (ret) {
			EMSG("Hardware busy.");
			return TEE_ERROR_BUSY;
		}

		size = MIN(HTRNG_BYTES, len - current_len);
		memcpy((uint8_t)buf + current_len, &val, size);
		current_len += size;
	}

	return TEE_SUCCESS;

**Review by etienne-lms — CHANGES_REQUESTED** [review]

**Review by etienne-lms — COMMENTED** [review]
You need some kind of mutex in `hw_get_random_bytes()` to protect against concurrent accesses to RNG, as pointed in https://github.com/OP-TEE/optee_os/pull/6217#discussion_r1282810444 and https://github.com/OP-TEE/optee_os/pull/6217#issuecomment-1702470222.

**Comment by loubaihui** [comment]
A spinklock is used in `trng_read`，and some fix is made based on your comments. 
Thank your for another round of review.

**Comment by loubaihui** [comment]
> A spinklock is used in `trng_read`，and some fix is made based on your comments. 
Thank your for another round of review. @jenswi-linaro @etienne-lms

File: core/drivers/hisi_trng.c [hunk 27]
```diff
@@ -0,0 +1,94 @@
...
+static TEE_Result trng_read(uint32_t *val)
+{
+	uint32_t exceptions = 0;
+
+	exceptions = cpu_spin_lock_xsave(&trng_lock);
+	if (IO_READ32_POLL_TIMEOUT(trng_dev->base + HTRNG_RANDATA_REG,
+				   val, val, POLL_PERIOD, POLL_TIMEOUT)) {
+		EMSG("Hardware busy.");
+		cpu_spin_unlock_xrestore(&trng_lock, exceptions);
+		return TEE_ERROR_BUSY;
```
jenswi-linaro, line 34: How about using a local variable `TEE_Result res = TEE_SUCCESS` and assigning it with `TEE_ERROR_BUSY` here? That way you could remove the `cpu_spin_unlock_xrestore()` at the line above.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,94 @@
...
+
+static TEE_Result trng_init(void)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+
+	IMSG("TRNG driver init start.");
+	trng_dev = calloc(1, sizeof(struct hisi_trng));
+	if (!trng_dev) {
+		EMSG("Fail to calloc trng device.");
+		return TEE_ERROR_STORAGE_NO_SPACE;
```
jenswi-linaro, line 78: TEE_ERROR_OUT_OF_MEMORY

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	trng->base = (uintptr_t)phys_to_virt_io(HISI_TRNG_BASE, HISI_TRNG_SIZE);
+	if (!trng_dev->base) {
+		EMSG("Fail to get trng io_base.");
+		free(trng_dev);
+		trng_dev = NULL;
+		return TEE_ERROR_BAD_FORMAT;
```
jenswi-linaro, line 86: Would TEE_ERROR_ACCESS_DENIED make sense?

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,94 @@
...
+	TEE_Result ret = TEE_ERROR_GENERIC;
+
+	IMSG("TRNG driver init start.");
+	trng_dev = calloc(1, sizeof(struct hisi_trng));
+	if (!trng_dev) {
+		EMSG("Fail to calloc trng device.");
+		return TEE_ERROR_STORAGE_NO_SPACE;
+	}
+
+	trng->base = (uintptr_t)phys_to_virt_io(HISI_TRNG_BASE, HISI_TRNG_SIZE);
```
jenswi-linaro, line 81: Please cast this to `vaddr_t`.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2021 HiSilicon Limited. */
+
+#include <initcall.h>
+#include <io.h>
+#include <malloc.h>
+#include <mm/core_memprot.h>
+#include <platform_config.h>
+#include <rng_support.h>
+#include <kernel/spinlock.h>
```
jenswi-linaro, line 10: Please keep these sorted, that is, move the line before the `malloc.h` line.

core/drivers/hisi_trng.c
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2021 HiSilicon Limited. */
```
jenswi-linaro, line 2: The year is 2023 now.

jenswi-linaro (commented):
With my last comments addressed, please apply:

loubaihui:
> With my last comments addressed, please apply

OK, thank you!

core/drivers/sub.mk
```diff
@@ -77,6 +77,7 @@ srcs-$(CFG_VERSAL_PUF) += versal_puf.c
 srcs-$(CFG_VERSAL_HUK) += versal_huk.c
 srcs-$(CFG_CBMEM_CONSOLE) += cbmem_console.c
 srcs-$(CFG_RISCV_PLIC) += plic.c
+srcs-$(CFG_HISILICON_PLAT_ACC) += hisi_trng.c
```
jforissier, line 80: This name (`CFG_HISILICON_PLAT_ACC`) is not specific enough. A few questions to help select a better name:
- We have `CFG_HI16xx_RNG` for `plat-d02` already, do you need a really new driver for `plat-d06`?
- If yes, could a similar naming scheme be used for both drivers? What are the official SoC names for D02 and D06? I believe D02 is HIP05 and D06 is HIP08. So how about `CFG_HIP05_RNG` and `CFG_HIP08_RNG`?

xiaoxuZeng:
> FWIW, `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` (I assume Jerome's comment [#6217 (comment)](https://github.com/OP-TEE/optee_os/pull/6217#discussion_r1371772177) will be fixed in some way). @xiaoxuZeng, could you ack the change, as d06 platform maintainer? (this change modifies the platform default configuration).

@etienne-lms 

I checked the changes, and it's okay.

Trng is a hardware module that generates random numbers in d06. The configuration of loubaihui is correct.
```
Acked-by: Xiaoxu Zeng <zengxiaoxu@huawei.com>
```

loubaihui:
@jforissier

jforissier:
Merged with minor edits to the commit description. Thanks!

loubaihui:
> Merged with minor edits to the commit description. Thanks!

Thank you!

PR #6219, opened 2023-08-03
Multiple Coverity fixes for CAAM Driver, Crypto API and Core

clementfaure:
Hello,
Our CI reported multiple Coverity issues in optee-os. The issues fixed in this pull-request are rated `High` and `Medium` impact. As a maintainer of the crypto driver API and the CAAM, I want to push 2878ec621 b3b05676a 1ba0f6dc6 d34d5c50b.

At the same time, I also fixed some issues in libutee, PTA attestation and the core. Feel free to comment regarding commits  d68119a17 dae33b69b a6c6ab0fe 726f19ea9 b98fa53d1

Thanks!

### DIFF
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -56,9 +56,11 @@ static TEE_Result generate_key(void)
 
 	res = allocate_key();
 	if (res)
-		return res;
+		goto err;
 
-	crypto_bignum_bin2bn((uint8_t *)&e, sizeof(e), key->e);
+	res = crypto_bignum_bin2bn((uint8_t *)&e, sizeof(e), key->e);
+	if (res)
+		goto err;
 
 	/*
 	 * For security reasons, the RSA modulus size has to be at least the
@@ -68,9 +70,12 @@ static TEE_Result generate_key(void)
 	COMPILE_TIME_ASSERT(CFG_ATTESTATION_PTA_KEY_SIZE >=
 			    TEE_SHA256_HASH_SIZE);
 	res = crypto_acipher_gen_rsa_key(key, CFG_ATTESTATION_PTA_KEY_SIZE);
-	if (res)
-		free_key();
+	if (!res)
+		goto out;
 
+err:
+	free_key();
+out:
 	return res;
 }
 
--- a/core/tee/entry_std.c
+++ b/core/tee/entry_std.c
@@ -351,13 +351,13 @@ static TEE_Result get_open_session_meta(size_t num_params,
 
 static void entry_open_session(struct optee_msg_arg *arg, uint32_t num_params)
 {
-	TEE_Result res;
+	TEE_Result res = TEE_ERROR_GENERIC;
 	TEE_ErrorOrigin err_orig = TEE_ORIGIN_TEE;
 	struct tee_ta_session *s = NULL;
-	TEE_Identity clnt_id;
-	TEE_UUID uuid;
+	TEE_Identity clnt_id = { };
+	TEE_UUID uuid = { };
 	struct tee_ta_param param = { };
-	size_t num_meta;
+	size_t num_meta = 0;
 	uint64_t saved_attr[TEE_NUM_PARAMS] = { 0 };
 
 	res = get_open_session_meta(num_params, arg->params, &num_meta, &uuid,
--- a/core/tee/entry_std.c
+++ b/core/tee/entry_std.c
@@ -356,7 +356,7 @@ static void entry_open_session(struct optee_msg_arg *arg, uint32_t num_params)
 	struct tee_ta_session *s = NULL;
 	TEE_Identity clnt_id;
 	TEE_UUID uuid;
-	struct tee_ta_param param;
+	struct tee_ta_param param = { };
 	size_t num_meta;
 	uint64_t saved_attr[TEE_NUM_PARAMS] = { 0 };
 
--- a/lib/libutee/tee_api.c
+++ b/lib/libutee/tee_api.c
@@ -754,12 +754,16 @@ static bool addr_is_in_no_share_heap(void *p)
 void *TEE_Realloc(void *buffer, size_t newSize)
 {
 	if (!newSize) {
-		TEE_Free(buffer);
+		void *ret = NULL;
 
 		if (addr_is_in_no_share_heap(buffer))
-			return TEE_NULL_SIZED_NO_SHARE_VA;
+			ret = TEE_NULL_SIZED_NO_SHARE_VA;
 		else
-			return TEE_NULL_SIZED_VA;
+			ret = TEE_NULL_SIZED_VA;
+
+		TEE_Free(buffer);
+
+		return ret;
 	}
 
 	if (buffer == TEE_NULL_SIZED_VA)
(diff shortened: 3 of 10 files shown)


jforissier (commented):
For "drivers: caam: free resource upon dmaobj initialization failure":

jforissier (commented):
For "drivers: caam: mp: fix memory on CAAM descriptor allocation failure":

jforissier (commented):
For "crypto: rsamgf: initialize allocated buffer":

jforissier (commented):
For "drivers: caam: remove dead code":

> Use array index syntax instead of pointer arithmetic for better
> readability.

This is a matter of preference, I believe @jenswi-linaro  prefers the latter :wink:

jforissier (commented):
For "lib: libutee: fix use after free":

lib/libutee/tee_api_property.c
```diff
@@ -201,7 +201,7 @@ TEE_Result TEE_GetPropertyAsString(TEE_PropSetHandle propsetOrEnumerator,
 {
 	TEE_Result res;
 	size_t l;
-	enum user_ta_prop_type type;
+	enum user_ta_prop_type type = 0;
```
jforissier, line 204: IIUC the important thing here is that the initial value be `!= USER_TA_PROP_TYPE_BINARY_BLOCK`. Should we introduce `USER_TA_PROP_TYPE_INVALID`? Or maybe just add a `static_assert(USER_TA_PROP_TYPE_BINARY_BLOCK != 0);`?

jforissier (commented):
Comments on "lib: libutee: initialize TA property type".

Please use prefix: "lib: libutee: TEE_GetPropertyAsString():"

jforissier (commented):
For "core: tee: initialize tee_ta_param object":

Please use prefix: "core: tee: entry_open_session():".

jforissier (commented):
For "core: tee: initialize dirfile_entry objects":

core/pta/attestation.c
```diff
@@ -55,9 +55,11 @@ static TEE_Result generate_key(void)
 
 	res = allocate_key();
 	if (res)
-		return res;
+		goto err;
```
jforissier, line 58: Not needed, if `allocate_key()` failed there should be nothing to free.

jforissier (commented):
For "core: pta: attestation: check return value of crypto_bignum_bin2bn()":

One comment below, with that:

core/pta/attestation.c
clementfaure: In `allocate_key()`, `crypto_acipher_alloc_rsa_keypair()` could fail but `key` still needs to be freed no?

jforissier: You're right. The `goto err;` is correct here.

lib/libutee/tee_api_property.c
clementfaure: I went for `USER_TA_PROP_TYPE_INVALID`

jforissier:
For "lib: libutee: initialize TA property type" + fixup:

clementfaure:
Force push: squash + rebase

lib/libutee/tee_api.c
```diff
@@ -754,12 +754,16 @@ static bool addr_is_in_no_share_heap(void *p)
 void *TEE_Realloc(void *buffer, size_t newSize)
 {
 	if (!newSize) {
-		TEE_Free(buffer);
+		void *ret = NULL;
 
 		if (addr_is_in_no_share_heap(buffer))
```
jenswi-linaro, line 759: For the record, this is a false positive. But this may look a bit odd so I believe we're better off checking the pointer first.

jenswi-linaro (commented):
For "lib: libutee: fix use after free":
You're fixing a false positive, I think that should be mentioned in the commit message.

lib/libutee/tee_api_property.c
```diff
@@ -201,7 +201,7 @@ TEE_Result TEE_GetPropertyAsString(TEE_PropSetHandle propsetOrEnumerator,
 {
 	TEE_Result res;
```
jenswi-linaro, line 202: While you're at it, please initialize all the locals here.

jenswi-linaro (commented):
For "lib: libutee: initialize TA property type"

I believe this is a false positive, but it's a good idea to plug it.
Especially since our coding guidelines require local variables to be initialized.

If you agree that it's a false positive please mention that in the commit message, or point out where the error is so I can double check.

jenswi-linaro (commented):
For "core: tee: initialize dirfile_entry objects"

All these looks like false positives, please update the commit message accordingly or point out where the error are.

lib/libutee/tee_api_property.c
clementfaure: Ok, I will do that in a separate commit

clementfaure: I don't think this is a false positive. We can reach the following if statement without `type` being initialized
https://github.com/OP-TEE/optee_os/blob/58d050167425a8513ceee9b8b21878ae0b13ee8f/lib/libutee/tee_api_property.c#L228

jenswi-linaro: How? I followed the different functions (and syscalls) and couldn't find the spot.

core/tee/entry_std.c
```diff
@@ -356,7 +356,7 @@ static void entry_open_session(struct optee_msg_arg *arg, uint32_t num_params)
 	struct tee_ta_session *s = NULL;
 	TEE_Identity clnt_id;
 	TEE_UUID uuid;
```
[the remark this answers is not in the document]
clementfaure: Ok, I will do in a separate commit.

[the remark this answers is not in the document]
clementfaure: According to Coverity, this is not a false positive. With `param` uninitialized, you reach
https://github.com/OP-TEE/optee_os/blob/ee8a977dddc04e28be96aed87fe23e0476588464/core/tee/entry_std.c#L388-L390
then read `param` with `mobj_put()`
https://github.com/OP-TEE/optee_os/blob/ee8a977dddc04e28be96aed87fe23e0476588464/core/tee/entry_std.c#L262-L270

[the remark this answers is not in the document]
jenswi-linaro: My bad, you're right.

lib/libutee/tee_api_property.c
clementfaure: In its analysis, Coverity does not follow the reference to `type` given as an argument to `propget_get_property()`. The analysis stays in the scope of `TEE_GetPropertyAsString()`

clementfaure:
> For "core: tee: initialize dirfile_entry objects"
> 
> All these looks like false positives, please update the commit message accordingly or point out where the error are.

Coverity does not follow variable references given as an argument to a function parameter. A function could potentially initialize a variable through its pointer, Coverity won't take that into account.

lib/libutee/tee_api_property.c
clementfaure: Ok, I will add it

clementfaure: Wait, it looks like `propget_get_property()` can return without initializing `type` no? It would not a false positive in that case.

jenswi-linaro: Yes, but it doesn't return TEE_ERROR_SHORT_BUFFER. However, it's a bit hard to tell with 100% certainty so I'm fine with this commit either way.

clementfaure:
Force-push : tag applied + re-word commits since uninitialized scalar value issues are not false positive

PR #6221, opened 2023-08-04
RISC-V: Refine TP restoring in trap handler

gagachang:
RISC-V kernel uses TP register to store thread_core_local structure. When the thread enters user mode, the value of TP is used by user mode. Therefore, when CPU enters trap handler, it needs to restore TP to get thread_core_local structure.

In previous implementation, the value of TP is saved under kernel SP before entering user mode, and the trap handler restores TP from that stack location. However, the value of TP has already been saved into the thread_user_mode_rec structure (https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/kernel/thread_rv.S#L202), which is also upon kernel SP, before entering user mode. So the value of TP can be restored just from thread_user_mode_rec, instead of saving into another location which is under the kernel SP.

### DIFF
--- a/core/arch/riscv/kernel/asm-defines.c
+++ b/core/arch/riscv/kernel/asm-defines.c
@@ -52,6 +52,8 @@ DEFINES
 	       offsetof(struct thread_user_mode_rec, ctx_regs_ptr));
 	DEFINE(THREAD_USER_MODE_REC_X1,
 	       offsetof(struct thread_user_mode_rec, x[0]));
+	DEFINE(THREAD_USER_MODE_REC_X4,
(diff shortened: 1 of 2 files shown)


gagachang:
Hi @maroueneboubakri 

This would improve the restoring of TP register in the trap handler.
Please check this PR, thanks!

maroueneboubakri:
> @maroueneboubakri Kindly request for review, thanks!

@gagachang kindly apply:

```
Tested-by: Marouene Boubakri <marouene.boubakri@nxp.com>
Reviewed-by: Marouene Boubakri <marouene.boubakri@nxp.com>
```

gagachang:
Tags applied

PR #6228, opened 2023-08-08
core: arm64: write_64bit_pair()

xiaoxuZeng:
Implement write_64bit_pair that write two 64 bits data together.

### DIFF
--- a/core/arch/arm/include/arm64.h
+++ b/core/arch/arm/include/arm64.h
@@ -355,6 +355,13 @@ static inline __noprof void tlbi_vale1is(uint64_t va)
 	asm volatile ("tlbi	vale1is, %0" : : "r" (va));
 }
 
+static inline void write_64bit_pair(uint64_t dst, uint64_t hi, uint64_t lo)
+{
+	/* 128bits should be written to hardware at one time */
+	asm volatile ("stp %1, %0, [%2]" : :
+		      "r" (hi), "r" (lo), "r" (dst) : "memory");
+}
+
 /*
  * Templates for register read/write functions based on mrs/msr
  */


core/include/io.h
```diff
@@ -40,6 +40,17 @@ static inline void io_write64(vaddr_t addr, uint64_t val)
 	*(volatile uint64_t *)addr = val;
 }
 
+static inline void write_64bit_pair(uintptr_t addr, uint64_t val_l,
```
jenswi-linaro, line 43: This function should go into `core/arch/arm/include/arm64.h`, with the `dmb oshst` instruction excluded.
Something along:
``` diff
--- a/core/arch/arm/include/arm64.h
+++ b/core/arch/arm/include/arm64.h
@@ -355,6 +355,12 @@ static inline __noprof void tlbi_vale1is(uint64_t va)
 	asm volatile ("tlbi	vale1is, %0" : : "r" (va));
 }
 
+static inline void write_64bit_pair(uint64_t dst, uint64_t hi, uint64_t lo)
+{
+	asm volatile ("stp	%0, %1, [%2]" : :
+		      "r" (hi), "r" (lo), "r" (dst) : "memory" );
+}
+
```
Note that the high argument should be before the low to harmonize with for instance `reg_pair_to_64()`.

core/arch/arm/include/arm64.h
```diff
@@ -355,6 +355,14 @@ static inline __noprof void tlbi_vale1is(uint64_t va)
 	asm volatile ("tlbi	vale1is, %0" : : "r" (va));
 }
 
+static inline void write_64bit_pair(uintptr_t dst, uint64_t hi,
```
jenswi-linaro, line 358: Please replace `uintptr_t` with `uint64_t`. Functions in this file are supposed to be as primitive as possible and the assembly instruction expects a 64-bit register.

core/include/io.h
xiaoxuZeng: Done, Thank you adivce.

core/arch/arm/include/arm64.h
xiaoxuZeng: Done

core/arch/arm/include/arm64.h
```diff
@@ -355,6 +355,14 @@ static inline __noprof void tlbi_vale1is(uint64_t va)
 	asm volatile ("tlbi	vale1is, %0" : : "r" (va));
 }
 
+static inline void write_64bit_pair(uint64_t dst, uint64_t hi,
+				    uint64_t lo)
```
etienne-lms, line 359: fits in 1 line

xiaoxuZeng: Done, Thank you adivce.

PR #6230, opened 2023-08-08
ta: pkcs11: Fix PKCS11 TA failures in ECDSA processing

brijeshamzn:
Fixes problem in ECDSA processing when ECDSA signing is chosen it need not depend on key size but a chosen algorithm need to be returned. Whereas during verification the key size need to be returned. Fixing these subtelities passes the PKCS11 test on platform using newer version of the global platform. The patch try to be backward compatible by conditionally compiling when the defines are expected for newer implementations.


ta/pkcs11/src/processing_asymm.c
```diff
@@ -88,13 +88,22 @@ pkcs2tee_algorithm(uint32_t *tee_id, uint32_t *tee_hash_id,
 		  TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA384, TEE_ALG_SHA384 },
 		{ PKCS11_CKM_SHA512_RSA_PKCS_PSS,
 		  TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA512, TEE_ALG_SHA512 },
+#if defined(TEE_ALG_ECDSA_SHA256) && (TEE_ALG_ECDSA_SHA256 == TEE_ALG_ECDSA_P256)
+ 		{ PKCS11_CKM_ECDSA, TEE_ALG_ECDSA_SHA256, TEE_ALG_SHA256},
```
etienne-lms, line 92: For `PKCS11_CKM_ECDSA`, the algorithm depends on key size and `.tee_hash_id` should be 0 so that `allocate_tee_operation()` does not compte the hash.

vesajaaskelainen:
What are you actually trying to improve/fix? Eg. what is the thing you want to do?

From my point of view the code works today? (I think we need to do changes here for other reasons but what is yours :))

I do not see it valuable for OP-TEE's PKCS#11 TA that we support different GP API versions.

jenswi-linaro:
> I do not see it valuable for OP-TEE's PKCS#11 TA that we support different GP API versions.

+1

brijeshamzn:
> What are you actually trying to improve/fix? Eg. what is the thing you want to do?
> 
> From my point of view the code works today? (I think we need to do changes here for other reasons but what is yours :))
> 
> I do not see it valuable for OP-TEE's PKCS#11 TA that we support different GP API versions.

brijeshamzn:
> > Hi,
> > I am seeing that PKCS11 TA would crash if newer version of GP API is used by a platform. So this code attempts to resolve that issue. It is funny that all algorithms themselves work fine only problem is in this part where parameters are vetted.
> 
> Repository optee_test implements ECDSA tests with the PKCS11 TA. Since release tag 3.21.0, PKCS11 TA is default build with GP TEE Internal Core API v1.3.1. No issues seen so far. Can you reproduce your issue with the existing tests?

I think you need to first admit that what should be a client responsibility and then TA's responsibilities. With GP API after 1.1.1 the algorithm type need to be coming from client for signing and then on verifying part the input size need to be dependent on key size. But in the TA those logics are seeming inverted probably fine for older API but not for newer API and any platform which implements the GP API properly which is in our case will fail after algorithm type is deduced on signing which can cause mismatch. I have tried to do right thing. However I am in favor of taking review comments which are reasonable and post a rework

etienne-lms:
Sorry if my comment seemed rude, it did not mean to be. I just try to understand the issue. PKCS11 TA should work, what ever GP TEE Internal Core API it is based on. If existing pkcs11 tests in optee_test do not properly cover ECDSA, we should fixed it, and fix the obsiouly fix TA if needed.

brijeshamzn:
> Sorry if my comment seemed rude, it did not mean to be. I just try to understand the issue. PKCS11 TA should work, what ever GP TEE Internal Core API it is based on. If existing pkcs11 tests in optee_test do not properly cover ECDSA, we should fixed it, and fix the obsiouly fix TA if needed.

The test client creates a 192 bit, 224, 256, 384 and 521 bits keys and for each key it does the ECDSA signing using 192, 224, 256, 384 and 512 SHA types which should be fine because ECDSA algorithm can use any key size with any SHA type. But once a particular key size is chosen it is clients responsibility to tell TA the type of SHA it is using and TA must serve with signing or verifying. But at our platforms TA kept failing and after debugging we found that TA even if the SHA type is told to it finds the SHA type based on keysize during signing and finds the input signature size based on SHA type, as a result 192 bit key will incorrectly work with only and only 192 SHA and so on but in practice it should not be the case. with this fix which on the signing part simply returns the SHA type being input and on verification part I try to use the key size as a guide to check the input size which must be 2* key size rounded off bytes. I am not sure why this issue is there in the TA and why it is not flagged as incorrect but it is wrong unfortunately.

brijeshamzn:
Should we proceed with this change ? Any other comments ?

etienne-lms:
My apologies @brijeshamzn. I think you're right but I have not spent enough time on this to make sure it's the right way to handle this GP TEE API change. What would help is an test example highlighting why current implementation is too restrictive regarding payload hash size VS signature key size.

At least, @jforissier, can you remove the Stale label from this P-R?

jforissier:
> At least, @jforissier, can you remove the Stale label from this P-R?

Sure, done!

vesajaaskelainen:
On point of OP-TEE's PKCS#11 TA I think right approach would be to upgrade OP-TEE's GP API to newer version and at that time adjust the PKCS#11 TA (and others where needed) instead of having those #ifdef's.

If there is a bug in current implementation then we should not require #ifdef's either.

I agree that some areas where this PR touches would need more work like better supporting different Edward curve variants vs. ECDSA curves. (Edwards curve support in PKCS#11 specification is also a bit broken).

Please note that OP-TEE and OP-TEE's PKCS#11 are bundled together in same release and used GP API version is coming from OP-TEE's GP API version.

brijeshamzn:
So What is the suggestion ? While I see the .h files they give an illusion that the support under OPTEE has advanced to the later version of GP API ? but in fact it seems that OPTEE still supports older GP API? Is there any broader plan to support newer GP API version ? 1.0 sounds too old and outdated ? what do you think ? Also what involves in advancing the GP API version ? How do you suggest to fix the problems you just described ?

jenswi-linaro:
OP-TEE provides v1.3.1 of GP Core Internal API and that's what the PKCS#11 TA is using also.

brijeshamzn:
@jenswi-linaro 
If that is true then concern raised by @vesajaaskelainen  is not valid. Instead I would argue that we need to merge the changes without ifdef's as the changes fix a genuine bug in the PKCS11 implementation at vetting its arguments. However it would also require change under the test case as the test case need to provide the input straight away and then PKCS11 should get rid of incorrect parameters check and while in doing so it must be understood that codes or platforms out in the world need to adjust accordingly. I am new to the party so I will leave upto you guys to decide and I am ready to do work whatever you guys decide and change this PR accordingly for your review.

jenswi-linaro:
Commit c7f1b4f7881f ("Upgrade in-tree TAs to use latest version of TEE Internal Core API") only did the bare minimum to make the in-tree TAs work with a 1.3.1 API.

etienne-lms:
I think the pkcs11 TA implementation should change regarding `PKCS11_CKM_ECDSA` so that it allows any wide enough key to sign the pre-hashed blob if the pre-hashed blob is not bigger than the key itselft. I think this is an expected behavior of how this mechanism can be used. 
Does that sound good to all of you? @brijeshamzn, would it fix your current issue?

brijeshamzn:
@etienne-lms  have you made the suggested patch? could you share the pull request or the commit with me ?

etienne-lms:
Is the issue related to this change already addressed? I think it is still to be addressed.
@brijeshamzn, should this change still be considered?

PR #6232, opened 2023-08-09
core: arm: plat-versal: Add maybe_unused attribute to constant strings

sriramsriram-msft:
If log level is set to print only EMSGs, constant strings can be unused. Add maybe_unused attribute to prevent compilation errors.


jforissier (commented):

sriramsriram-msft:
This PR can be merged.

PR #6233, opened 2023-08-10
drivers: clk: avoid the assert failure when there's "assigned-clocks"

TonyHan11:
Once "assigned-clocks" is parsed correctly variable "clk" will retain the non-NULL value and skip "return", when "res" is non-zero for new "clock_idx" assert(false) will happen.


core/drivers/clk/clk_dt.c
```diff
@@ -179,6 +179,7 @@ static void parse_assigned_clock(const void *fdt, int nodeoffset)
 		}
 
 		clock_idx++;
+		clk = NULL;
```
jenswi-linaro, line 182: This obviously fixes the issue, but I wonder if the root of the problem isn't further up around lines 158-162.
Perhaps something like:
``` C
if (res)
     return;
assert(clk);
```

TonyHan11: Updated, thank you for your comment.

core/drivers/clk/clk_dt.c
```diff
@@ -157,9 +157,9 @@ static void parse_assigned_clock(const void *fdt, int nodeoffset)
 	while (true) {
 		res = clk_dt_get_by_idx_prop("assigned-clocks", fdt, nodeoffset,
 					     clock_idx, &clk);
-		if (!clk)
+		if (res)
```
jenswi-linaro, line 160: With this, I believe that we can remove `__maybe_unused` from `res` above.

jenswi-linaro (commented):

etienne-lms (approved):

PR #6250, opened 2023-08-22
core: fix bug when packing RPMB read frame

hunter-Ding:
- when read RPMB, should contain block count in packet according to spec

### DIFF
--- a/core/tee/tee_rpmb_fs.c
+++ b/core/tee/tee_rpmb_fs.c
@@ -1242,6 +1242,7 @@ static TEE_Result tee_rpmb_read(uint16_t dev_id, uint32_t addr, uint8_t *data,
 	rawdata.msg_type = msg_type;
 	rawdata.nonce = nonce;
 	rawdata.blk_idx = &blk_idx;
+	rawdata.block_count = &blkcnt;
 	res = tee_rpmb_req_pack(req, &rawdata, 1, dev_id, NULL, NULL);
 	if (res != TEE_SUCCESS)
 		goto func_exit;


hunter-Ding:
> Hi @hunter-Ding, do you have a test case for this fix?

sorry for saying no.
I debug code about RPMB, and found the issue by dumping frames from OP-TEE.
image
"Block Count" is nessary according to the pic

jforissier:
OK. While this looks good I must say it is some time since I have not dived in the RPMB code and I fail to see why it works in the current state even for multiple block reads. That's a bit annoying. Test case: QEMUv8, `make CFG_REE_FS=n CFG_RPMB_FS=y CFG_RPMB_WRITE_KEY=y CFG_RPMB_TESTKEY=y run` and `xtest 6001`, you can see debug traces: "Read 8 blocks at index 2".

jforissier:
Regarding the checkpatch error: please fix the commit description
```
 core: fix bug when packing RPMB read frame
```
You need a blank line here
```
- when read RPMB, should contain block count
```
...and replace `- when` by `When`

```
  in packet according to spec

Signed-off-by: dingfm <fangming123123123@163.com>
```
Please use your real name. `dingfm` doesn't sound like one.

etienne-lms (commented):
~~`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`~~

**(edited) A-b tag removed as I think the proposed change is not needed.**

etienne-lms:
Looking at the Jedec spec, I seems to me the bock count field patched by the P-R is expected as 0.

In JEDEC Standard No. 84-A441, March 2010, section "7.6.16.4 Accesses to the Replay Protected Memory Block", in bullet sub-section "Authenticated Data Read" (page 78), it's written that a CMD25 initiates the read request (with request type 0x0004) and is followed by a CMD18 to fetch the effective data. 
The 1st table in that bullet sub-section shows CMD25 frame content. "Block Count" (2Bytes, position [5:4]) is explicitly set to "0x00". Note that for this message type, MAC is not computed.

If I understood it right, this P-R is not required.

I've done few experiments with the eMMC device I have. For that read request command, block count field seems completly ignored, any values (from 0 to 0xFFFFF) passes and RPMB block(s) are always succesfully read.

PR #6256, opened 2023-08-24
ci: qemuv8: add check Hafnium

jenswi-linaro:
Adds a check with Hafnium, that is, testing with Hafnium as SPMC at S-EL2 and OP-TEE as an SP at S-EL1. Building with SPMC_AT_EL=2.


jenswi-linaro:
And CI passed finally.

jenswi-linaro:
Tag applied

PR #6257, opened 2023-08-24
core: support larger values for CFG_TEE_CORE_NB_CORE

jenswi-linaro:
With larger values of CFG_TEE_CORE_NB_CORE (for example, 18 on the marvell-cnf10ka platform) CORE_MMU_BASE_TABLE_OFFSET becomes to large to be used as an immediate value in add and sub assembly instructions. This is handled by populating a spare register using the mov_imm macro and use the spare register instead. But the mov_imm macro can handle complex defines so CORE_MMU_BASE_TABLE_OFFSET must be evaluated in asm-defines.c first.

This should fix errors like:
core/arch/arm/kernel/thread_a64.S: Assembler messages: core/arch/arm/kernel/thread_a64.S:339: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:347: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:355: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:372: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:379: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:386: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:660: Error: immediate out of range core/arch/arm/kernel/thread_a64.S:732: Error: immediate out of range make: *** [mk/compile.mk:165: out/core/arch/arm/kernel/thread_a64.o] Error 1

### DIFF
--- a/core/arch/arm/include/arm64_macros.S
+++ b/core/arch/arm/include/arm64_macros.S
@@ -116,6 +116,24 @@
 		movk    \_reg, :abs_g0_nc:\_val
 	.endm
 
+	.macro add_imm _reg, _val
+	.if ((\_val) > 0xfff)
+		add	\_reg, \_reg, ((\_val) >> 12), LSL #12
+	.endif
+	.if (((\_val) & 0xfff) > 0)
+		add	\_reg, \_reg, ((\_val) & 0xfff)
+	.endif
+	.endm
+
+	.macro sub_imm _reg, _val
+	.if ((\_val) > 0xfff)
+		sub	\_reg, \_reg, ((\_val) >> 12), LSL #12
+	.endif
+	.if (((\_val) & 0xfff) > 0)
+		sub	\_reg, \_reg, ((\_val) & 0xfff)
+	.endif
+	.endm
+
 	/*
 	 * Load address of <sym> into <reg>, <sym> being in the range
 	 * +/- 4GB of the PC (note that 'adr reg, sym' is limited to +/- 1MB).
--- a/core/arch/arm/kernel/asm-defines.c
+++ b/core/arch/arm/kernel/asm-defines.c
@@ -7,6 +7,7 @@
 #include <kernel/boot.h>
 #include <kernel/thread.h>
 #include <kernel/thread_private.h>
+#include <mm/core_mmu_arch.h>
 #include <sm/pm.h>
 #include <sm/sm.h>
 #include <types_ext.h>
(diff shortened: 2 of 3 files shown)


jforissier:
> `use the spare register instead. But the mov_imm macro can handle complex`

s/can/can't/

With that:

jenswi-linaro:
Commit message update and tag applied.

jenswi-linaro:
I was a bit quick to think that the previous patch would work. I'm trying a new approach with add_imm and sub_imm instead to avoid using an extra register for sub or add.

core/arch/arm/kernel/asm-defines.c
```diff
@@ -157,4 +158,13 @@ DEFINES
 	       offsetof(struct boot_embdata, reloc_offset));
 	DEFINE(BOOT_EMBDATA_RELOC_LEN,
 	       offsetof(struct boot_embdata, reloc_len));
+
+#ifdef CORE_MMU_BASE_TABLE_OFFSET
+	/*
+	 * This define is too complex to be used as an argument for mov_imm
```
jforissier, line 164: add_imm/sub_imm

jforissier:
> I was a bit quick to think that the previous patch would work. I'm trying a new approach with add_imm and sub_imm instead to avoid using an extra register for sub or add.

Hmmm it was clobbering some registers that should have been preserved, right? It wasn't too obvious.

For "core: arm64: add add_imm and sub_imm assembly macros" please apply:

For "core: support larger values for CFG_TEE_CORE_NB_CORE" with my comment addressed:

jforissier (commented):
And here's the comment ;)

jenswi-linaro:
Update

gowthamithiagu (commented):
Thanks for taking up this one at the earliest. With this change, not seeing the immediate out of range error in the platforms I have tested.

gowthamithiagu:
With this,

Tested-by: Gowthami <gthiagarajan@marvell.com>

core/arch/arm/include/arm64_macros.S
```diff
@@ -116,6 +116,24 @@
 		movk    \_reg, :abs_g0_nc:\_val
 	.endm
 
+	.macro add_imm _reg, _val
+	.if ((\_val) > 0xfff)
+		add	\_reg, \_reg, ((\_val) >> 12), LSL #12
+	.endif
+	.if (((\_val) & 0xfff) > 0)
```
etienne-lms, line 123: Is `> 0` needed?

etienne-lms (approved):

core/arch/arm/include/arm64_macros.S
jenswi-linaro: We can potentially save one instruction.

jenswi-linaro:
Tags applied.

PR #6258, opened 2023-08-24
plat-k3: Default to 2 core per cluster only for AM65x

glneo:
All other SoCs have 4 cores per cluster, which is the default, or they only have one cluster in which case this value is unimportant.


jenswi-linaro:

jforissier:
@glneo would you mind applying Jens' Acked-by tag? Thanks!

PR #6263, opened 2023-08-28
core: assign memory tags to bounce buffers

jenswi-linaro:
Just as the heap uses memory tags (CFG_MEMTAG=y) use memory tags for the temporary bounce buffers. This should catch problems with out of bounds accesses, using already freed, reset bounce buffers.


jforissier (commented):

jenswi-linaro:
tag applied.

PR #6271, opened 2023-08-30
Optee watchdog

ldebieve:
Add a dedicated OP-TEE ID for a SMC fast call entry to manage the ARM watchdog service.

Recreating this pull request based on first review [https://github.com/OP-TEE/optee_os/pull/5947].
Too late for an update, comment has been taken and cleanup done to define a dedicated OP-TEE SMC ID.

### DIFF
--- a/core/arch/arm/include/sm/watchdog_smc.h
+++ b/core/arch/arm/include/sm/watchdog_smc.h
@@ -1,0 +1,81 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) STMicroelectronics 2023
+ */
+#ifndef SM_WATCHDOG_SMC_H
+#define SM_WATCHDOG_SMC_H
+
+/*
+ * This file describes the secure watchdog management service.
+ * It exposes the SMC supported commands.
+ * We're following ARM SMC Calling Convention as specified in
+ * https://developer.arm.com/documentation/den0028.
+ */
+
+/*
+ * Overall global call for watchdog interface
+ * Call requests usage:
+ * a0		[in]		SMC Function ID, CFG_WDT_SM_HANDLER_ID
+ *		[out]		PSCI error code return
+ * a1		[in]		Watchdog command (one of SMCWD_*)
+ *		[out]		Depends on watchdog command (input a1)
+ * a2		[in/out]	Depends on watchdog command (input a1)
+ * a3-6		[in/out]	Not used
+ * a7		[in/out]	Hypervisor Client ID register
+ */
+
+/* Watchdog supported commands */
+
+#define SMCWD_INIT		0
+#define SMCWD_SET_TIMEOUT	1
+#define SMCWD_ENABLE		2
+#define SMCWD_PET		3
+#define SMCWD_GET_TIMELEFT	4
+
+/*
+ * Command SMCWD_INIT : Watchdog initialization
+ * [in]		a1	Set to SMCWD_INIT
+ * [out]	a1	The minimal timeout value in seconds supported
+ *		a2	The maximum timeout value in seconds supported
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_INTERNAL_FAILURE - Initialization failure
+ *
+ * Command SMCWD_SET_TIMEOUT : Watchdog set timeout
+ * [in]		a1	Set to SMCWD_SET_TIMEOUT
+ *		a2	The timeout value in seconds to set
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_INVALID_PARAMETERS - Incorrect input param
+ *
+ * Command SMCWD_ENABLE : Watchdog enable
+ * [in]		a1	Set to SMCWD_ENABLE
+ *		a2	Set to 0 to stop the watchdog, 1 to enable it
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_INVALID_PARAMETERS - Incorrect input param
+ *
+ * Command SMCWD_PET : Ping the watchdog for refresh
+ * [in]		a1	Set to SMCWD_PET
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_DISABLED - The watchdog is not enabled
+ *
+ * Command SMCWD_GET_TIMELEFT : Get time left
+ * [in]		a1	Set to SMCWD_GET_TIMELEFT
+ * [out]	a1	The timeout value in seconds before watchdog expires
+ * Return codes:
+ * PSCI_RET_SUCCESS - Command success
+ * PSCI_RET_DISABLED - The watchdog is not enabled
+ * PSCI_RET_NOT_SUPPORTED - Function not supported
+ *
+ * Other commands
+ * [in]		a1	Other values
+ * Return codes:
+ * PSCI_RET_NOT_SUPPORTED - Function not supported
+ *
+ * a3-6		Not used
+ * a7		Hypervisor Client ID register
+ */
+
+#endif /* SM_WATCHDOG_SMC_H */
--- a/core/arch/arm/tee/entry_fast.c
+++ b/core/arch/arm/tee/entry_fast.c
@@ -5,6 +5,7 @@
  */
 
 #include <config.h>
+#include <drivers/wdt.h>
 #include <kernel/boot.h>
 #include <kernel/misc.h>
 #include <kernel/notif.h>
@@ -13,6 +14,7 @@
 #include <mm/core_mmu.h>
 #include <optee_msg.h>
 #include <sm/optee_smc.h>
+#include <sm/watchdog_smc.h>
 #include <tee/entry_fast.h>
 
 #ifdef CFG_CORE_RESERVED_SHM
@@ -217,6 +219,15 @@ static void get_async_notif_value(struct thread_smc_args *args)
 		args->a2 |= OPTEE_SMC_ASYNC_NOTIF_PENDING;
 }
 
+static void tee_entry_watchdog(struct thread_smc_args *args)
+{
+#if defined(CFG_WDT_SM_HANDLER)
+	__wdt_sm_handler(args);
+#else
+	args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
+#endif
+}
+
 /*
  * If tee_entry_fast() is overridden, it's still supposed to call this
  * function.
@@ -289,6 +300,11 @@ void __tee_entry_fast(struct thread_smc_args *args)
 			get_async_notif_value(args);
 		else
 			args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
+		break;
+
+	/* Watchdog entry if handler ID is defined in TOS range */
+	case CFG_WDT_SM_HANDLER_ID:
+		tee_entry_watchdog(args);
 		break;
 
 	default:
--- a/core/drivers/wdt/watchdog_sm.c
+++ b/core/drivers/wdt/watchdog_sm.c
@@ -7,14 +7,7 @@
 #include <kernel/spinlock.h>
 #include <sm/optee_smc.h>
 #include <sm/psci.h>
-
-enum smcwd_call {
-	SMCWD_INIT		= 0,
-	SMCWD_SET_TIMEOUT	= 1,
-	SMCWD_ENABLE		= 2,
-	SMCWD_PET		= 3,
-	SMCWD_GET_TIMELEFT	= 4,
-};
+#include <sm/watchdog_smc.h>
 
 static unsigned long wdt_min_timeout;
 static unsigned long wdt_max_timeout;
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -986,13 +986,15 @@ endif
 CFG_WDT ?= n
 
 # Enable watchdog SMC handling compatible with arm-smc-wdt Linux driver
-# When enabled, CFG_WDT_SM_HANDLER_ID must be defined with a SMC ID
 CFG_WDT_SM_HANDLER ?= n
 
 $(eval $(call cfg-enable-all-depends,CFG_WDT_SM_HANDLER,CFG_WDT))
-ifeq (y-,$(CFG_WDT_SM_HANDLER)-$(CFG_WDT_SM_HANDLER_ID))
-$(error CFG_WDT_SM_HANDLER_ID must be defined when enabling CFG_WDT_SM_HANDLER)
-endif
+
+# When CFG_WDT_SM_HANDLER=y, SMC function ID 0x82003D06 default implements
+# arm-smc-wdt service. Platform can also override this ID with a platform
+# specific SMC function ID to access arm-smc-wdt service thanks to
+# optional config switch CFG_WDT_SM_HANDLER_ID.
+CFG_WDT_SM_HANDLER_ID ?= 0x82003D06
 
 # Allow using the udelay/mdelay function for platforms without ARM generic timer
 # extension. When set to 'n', the plat_get_freq() function must be defined by


core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
...
+#ifndef OPTEE_EXTRA_SMC_H
+#define OPTEE_EXTRA_SMC_H
+
+#include <sm/optee_smc.h>
+
+/*
+ * This file describes the trusted OS additional part.
+ * It exposes the SMC function ID used and the associated functions.
+ * We're following ARM SMC Calling Convention as specified in
+ * http://infocenter.arm.com/help/topic/com.arm.doc.den0028a/index.html
```
etienne-lms, line 14: Updated URL: https://developer.arm.com/documentation/den0028

core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
...
+#define OPTEE_SMC_OWNER_EXTRA_CALL_VAL(func_num) \
+	OPTEE_SMC_CALL_VAL(OPTEE_SMC_32, OPTEE_SMC_FAST_CALL, \
+			   OPTEE_SMC_OWNER_TRUSTED_OS_EXTRA, (func_num))
+
+#define OPTEE_SMC_FUNCID_EXTRA_CALLS_COUNT	U(0xFF00)
+
+/*
+ * Secure watchdog management service
+ * SMC ID uses a Fast Call in the Trusted OS Calls ranges.
+ *
```
etienne-lms, line 29: can remove this extra empty line

core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
...
+ * This file describes the trusted OS additional part.
+ * It exposes the SMC function ID used and the associated functions.
+ * We're following ARM SMC Calling Convention as specified in
+ * http://infocenter.arm.com/help/topic/com.arm.doc.den0028a/index.html
+ *
+ */
+
+#define OPTEE_SMC_OWNER_TRUSTED_OS_EXTRA	U(60)
+
+#define OPTEE_SMC_OWNER_EXTRA_CALL_VAL(func_num) \
```
etienne-lms, line 20: this macro seems not used. Can remove.

core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) STMicroelectronics 2023 - All Rights Reserved
+ */
+#ifndef OPTEE_EXTRA_SMC_H
+#define OPTEE_EXTRA_SMC_H
```
etienne-lms, line 6: prefer `SM_EXTRA_SMC_H` for consistency with the header file patch/name.

core/drivers/wdt/watchdog_sm.c
```diff
@@ -6,16 +6,9 @@
 #include <drivers/wdt.h>
 #include <kernel/spinlock.h>
 #include <sm/optee_smc.h>
+#include <sm/extra_smc.h>
```
etienne-lms, line 9: swap the 2 above lines

core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
...
+/* Watchdog supported commands */
+
+#define SMCWD_INIT		0
+#define SMCWD_SET_TIMEOUT	1
+#define SMCWD_ENABLE		2
+#define SMCWD_PET		3
+#define SMCWD_GET_TIMELEFT	4
+
+/*
+ * Command SMCWD_ENABLE : Watchdog initialization
```
etienne-lms, line 58: s/SMCWD_ENABLE/SMCWD_INIT/

etienne-lms (commented):
Few minor comments, otherwise looks good to me.

mk/config.mk
```diff
@@ -984,13 +984,13 @@ endif
...
-# When enabled, CFG_WDT_SM_HANDLER_ID must be defined with a SMC ID
+# When CFG_WDT_SM_HANDLER=y, SMC function ID 0xb2000014 implements
+# arm-smc-wdt service. Platform can also define an additional platform
+# specific SMC function ID to access arm-smc-wdt service thanks to optional
+# config switch CFG_WDT_SM_HANDLER_ID.
 CFG_WDT_SM_HANDLER ?= n
 
 $(eval $(call cfg-enable-all-depends,CFG_WDT_SM_HANDLER,CFG_WDT))
-ifeq (y-,$(CFG_WDT_SM_HANDLER)-$(CFG_WDT_SM_HANDLER_ID))
-$(error CFG_WDT_SM_HANDLER_ID must be defined when enabling CFG_WDT_SM_HANDLER)
```
jenswi-linaro, line 992: I don't see why don't always require `CFG_WDT_SM_HANDLER_ID` to define the id.

etienne-lms: SMC funcID  defined with `CFG_WDT_SM_HANDLER_ID` is trapped by OP-TEE Monitor (_core/arch/arm/sm/sm.c_). The proposed change allows to route Watchdog SMC functions to OP-TEE even when using TF-A BL31 Monitor.

core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
...
+
+/*
+ * This file describes the trusted OS additional part.
+ * It exposes the SMC function ID used and the associated functions.
+ * We're following ARM SMC Calling Convention as specified in
+ * http://infocenter.arm.com/help/topic/com.arm.doc.den0028a/index.html
+ *
+ */
+
+#define OPTEE_SMC_OWNER_TRUSTED_OS_EXTRA	U(60)
```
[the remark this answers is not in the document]
etienne-lms: Using Trusted OS range makes TF-A BL31 to route the SMC to OP-TEE, whereas a SIP would not reach OP-TEE SPD of TF-A. This is useful when watchdog driver is owned by Trusted OS, not by BL31.

core/arch/arm/include/sm/extra_smc.h
```diff
@@ -0,0 +1,104 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) STMicroelectronics 2023 - All Rights Reserved
```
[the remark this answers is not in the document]
ldebieve: no, I'll remove it

core/arch/arm/include/sm/extra_smc.h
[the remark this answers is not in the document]
ldebieve: The idea is to keep opened this file for other TOS SMC services (if any), not limited to watchdog.
For the range, cf @etienne-lms reply.

mk/config.mk
ldebieve: @etienne-lms is too fast. Confirm the reply too.

jenswi-linaro: That doesn't answer the question. Why can't we always use `CFG_WDT_SM_HANDLER_ID` to define the watchdog SMC ID? We could define a sensible default value, like `0x82003D06`. If it's not good enough for some platform, just override it.

core/arch/arm/include/sm/extra_smc.h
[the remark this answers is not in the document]
jenswi-linaro: We have the de facto standard SMC id `0x82003D06` which works perfectly well on v7 systems where I suppose this will be used. If you need another value you can override it in the platform configuration using `CFG_WDT_SM_HANDLER_ID`.

> The idea is to keep opened this file for other TOS SMC services (if any), not limited to watchdog.

My hope is that we'll need none extra at all.

> Using Trusted OS range makes TF-A BL31 to route the SMC to OP-TEE, whereas a SIP would not reach OP-TEE SPD of TF-A. This is useful when watchdog driver is owned by Trusted OS, not by BL31.

Adding random SMC services to OP-TEE is not compatible with FF-A, so in the longer term, it will become a legacy thing. If you're aiming for v8 systems perhaps it's better to try to avoid the legacy track.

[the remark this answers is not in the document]
ldebieve: >We have the de facto standard SMC id 0x82003D06 which works perfectly well on v7 systems where I suppose this will be used. If you need another value you can override it in the platform configuration using CFG_WDT_SM_HANDLER_ID.

You're right, it became standard as put in the Linux driver but the initial target was TF-A SMC call, not TOS implementation. This is only limited to Linux side as OP-TEE doesn't define any HANDLER_ID by default for v7 implementation. Whatever the value, you must define it when WDT is enable.
The first idea was to use a standard ID for Trusted OS as the watchdog is managed in the OP-TEE side and put it as an TOS SMC SIP range.
Going in that way will allow a generic approach for platform using OP-TEE that use the same ID between 32 and 64 bit platform.

>>The idea is to keep opened this file for other TOS SMC services (if any), not limited to watchdog.

> My hope is that we'll need none extra at all.

I hope so..., bet on the future, no issue to rename and only keep it for WDT for now.

> Adding random SMC services to OP-TEE is not compatible with FF-A, so in the longer term, it will become a legacy thing. If you're aiming for v8 systems perhaps it's better to try to avoid the legacy track.

Not agree with you. Platform working on pre armv8.4 will not switch to FF-A (FF-A will not be legacy for such platforms). I'll expect a watchdog ID for FF-A later for those which use it but we need for 64bit platform a way to address the secure watchdog manage by the TOS that will be redirect by the TF-A (current implementation).

mk/config.mk
ldebieve: Defining this one will make it a standard that is not better that giving a OP-TEE SMC TOS ID instead. Only working again for armv7 not supported on armv8 64bit.

core/arch/arm/include/sm/extra_smc.h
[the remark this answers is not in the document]
jenswi-linaro: I don't share your view, but it's pointless to argue about that here. Please make this file exclusive for WDT and we should be done here.

mk/config.mk
jenswi-linaro: Use whatever default value you find useful, but it should be defined as the SMC ID used in the current configuration, and platforms are supposed to be able to override it. No need for a second way to define the SMC ID.

core/arch/arm/include/sm/watchdog_smc.h
```diff
@@ -0,0 +1,93 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) STMicroelectronics 2023
+ */
+#ifndef WATCHDOG_SMC_H
+#define WATCHDOG_SMC_H
```
etienne-lms, line 6: `SM_WATCHDOG_SMC_H`

core/arch/arm/include/sm/watchdog_smc.h
```diff
@@ -0,0 +1,93 @@
...
+ * It exposes the SMC function ID used and the associated functions.
+ * We're following ARM SMC Calling Convention as specified in
+ * https://developer.arm.com/documentation/den0028
+ */
+
+/*
+ * Secure watchdog management service
+ * SMC ID uses a Fast Call in the Trusted OS Calls ranges.
+ */
+#define OPTEE_SMC_OWNER_TRUSTED_OS_WATCHDOG	U(60)
```
etienne-lms, line 21: For consistency, I propose to explicit the function number with a macro:
```patch
 #define OPTEE_SMC_OWNER_TRUSTED_OS_WATCHDOG	U(60)
+#define OPTEE_SMC_FUNC_NUM_WATCHDOG		U(0)
 #define OPTEE_SMC_WATCHDOG \
 	OPTEE_SMC_CALL_VAL(OPTEE_SMC_32, OPTEE_SMC_FAST_CALL, \
 			   OPTEE_SMC_OWNER_TRUSTED_OS_WATCHDOG, \
-			   U(0))
+			   OPTEE_SMC_FUNC_NUM_WATCHDOG)

File: mk/config.mk [hunk 12]
```diff
@@ -984,13 +984,16 @@ endif
...
-endif
+
+# When CFG_WDT_SM_HANDLER=y and OP-TEE is used as monitor,
+# SMC function ID 0x82003D06 implements arm-smc-wdt service.
+# Platform can also define an additional platform specific SMC function ID
```
etienne-lms, line 993: Proposal for rephrasing:
\# When CFG_WDT_SM_HANDLER=y and OP-TEE is used as monitor **(i.e. OP-TEE 32bit Secure Monitor)**
\# SMC function ID 0x82003D06 **default** implements arm-smc-wdt service.
\# Platform can **~~also define an additional~~ override this ID with a** platform specific SMC function ID
\# to access arm-smc-wdt service thanks to optional config switch
\# CFG_WDT_SM_HANDLER_ID.

etienne-lms (commented):
Hi @ldebieve, 
To easy review, I suggest you squash the fixup commits.

core/arch/arm/include/sm/watchdog_smc.h
ldebieve: I've made same as optee_smc.h which use OPTEE_SMC_H.
Only SM and PM use this, I'm ok but difficult to get consistency with other file.
No issue to put SM whatever, not seeing an explicit rule to use the folder name (which is not used in the other include folders)

mk/config.mk
ldebieve: Agree

core/arch/arm/include/sm/watchdog_smc.h
etienne-lms: _optee_smc.h_ header file is there since a while, before OP-TEE tried to harmonize the header file guards.
I think _optee_smc.h_  could be updated accordingly. I guess some other existing legacy header files also.

mk/config.mk
ldebieve: @jenswi-linaro Any comment before I update and squash ?

jenswi-linaro: Please go ahead and squash.

core/arch/arm/tee/entry_fast.c
```diff
@@ -291,6 +302,10 @@ void __tee_entry_fast(struct thread_smc_args *args)
 			args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
 		break;
 
+	case OPTEE_SMC_WATCHDOG:
```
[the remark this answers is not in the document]
ldebieve: We will catch here the OPTEE predefined ID for 64bit mode, the CFG_WDT_SM_HANDLER_ID purpose is limited to 32bit mode (defined in sm/sm.c).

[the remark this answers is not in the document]
jenswi-linaro: So for armv7 you support configuring a custom SMC ID, but not for armv8. I would prefer if there's a single way of configuring the SMC ID, that is, that `CFG_WDT_SM_HANDLER_ID` always is used.

core/arch/arm/include/sm/watchdog_smc.h
```diff
@@ -0,0 +1,81 @@
...
+ * This file describes the secure watchdog management service.
+ * It exposes the SMC supported commands.
+ * We're following ARM SMC Calling Convention as specified in
+ * https://developer.arm.com/documentation/den0028.
+ */
+
+/*
+ * Overall global call for watchdog interface
+ * Call requests usage:
+ * a0		[in]		SMC Function ID, OPTEE_SMC_WATCHDOG
```
jenswi-linaro, line 18: CFG_WDT_SM_HANDLER_ID

jenswi-linaro (commented):

core/arch/arm/tee/entry_fast.c
```diff
@@ -217,6 +219,15 @@ static void get_async_notif_value(struct thread_smc_args *args)
 		args->a2 |= OPTEE_SMC_ASYNC_NOTIF_PENDING;
 }
 
+static void tee_entry_watchdog(struct thread_smc_args *args)
+{
+#if defined(CFG_WDT_SM_HANDLER)
+		__wdt_sm_handler(args);
```
etienne-lms, line 225: remove 1 tab

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with tab fixed

jenswi-linaro:

ldebieve:
Review tags applied
BR,

PR #6276, opened 2023-09-01
libutils: mempool: fix unbalanced put_pool()

jenswi-linaro:
Prior to this patch mempool_free() unconditionally called put_pool(), but if the "ptr" argument is NULL it means that there hasn't been a corresponding call to get_pool(). Fix this only calling put_pool() for non-NULL pointers.

Fixes: a51d45b52503 ("libutils: mempool based raw malloc functions")


jforissier (commented):

jenswi-linaro:
Wrapped commit message, tag applied

Issue #6226, opened 2023-08-07
Dead lock for multi-session


charles1129:
Dear Sirs,
> Recently we noticed a issue with our custom TA. If CA was killed using killall command and then restarted, sometimes it would stuck in uninterruptible sleep. It should not be a problem during normal work, but there might be a risk of CA being stuck on sleep after a kill by OOM killer or restarted for some other reason. 
It seems like there is some deadlock in TEE code causing this, so I conducted a stress test for hello world example.

### hello_world TA : 

- Add multi session flag in user_ta_header_defines.h:

```
#define TA_FLAGS (TA_FLAG_USER_MODE | TA_FLAG_EXEC_DDR | \
                 TA_FLAG_SINGLE_INSTANCE | TA_FLAG_MULTI_SESSION)
```

- Trigger panic in invoke:

```
    TEE_Result TA_InvokeCommandEntryPoint(void *sess_ctx __unused, uint32_t cmd_id,
                                      uint32_t param_types,
                                      TEE_Param params[4] __unused)
    {
        uint32_t exp_param_types = TEE_PARAM_TYPES(TEE_PARAM_TYPE_NONE,
                                                   TEE_PARAM_TYPE_NONE,
                                                   TEE_PARAM_TYPE_NONE,
                                                   TEE_PARAM_TYPE_NONE);

        DMSG("%s(): begin called with cmd_id:%#" PRIx32 "\n", __func__, cmd_id);

        TEE_Panic(-1);
        ...
    }
```

### hello_world CA:
Execute hello_world CA in two while loop

    Process 1 run CA in while loop;
    Process 2 run CA in while loop;

### Debug:
After debug, I find it's often blocked in core/kernel/tee_ta_manager.c:
```
    static void tee_ta_unlink_session(struct tee_ta_session *s,
                        struct tee_ta_session_head *open_sessions)
    {
        mutex_lock(&tee_ta_mutex);

        ...

        while (s->ref_count != 1)
                 condvar_wait(&s->refc_cv, &tee_ta_mutex); // blocked here 

        TAILQ_REMOVE(open_sessions, s, link);

        mutex_unlock(&tee_ta_mutex);
    }

```

Do you have any better suggestions?

Best Regards

jenswi-linaro:
It seems that the thread holding `tee_ta_mutex` either is killed by the OOM killer or perhaps is blocked by something else in the kernel driver.

charles1129:
hi @jenswi-linaro 

[...]

PR #6281, opened 2023-09-04
Fix ta panic race

jenswi-linaro:
This should fix the issue reported in https://github.com/OP-TEE/optee_os/issues/6226


jenswi-linaro:
> Having some test cases in `xtest` would be great, but probably not simple since we're talking race conditions here...

Sorry, forgot to make the test PR too. https://github.com/OP-TEE/optee_test/pull/693

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`  with minor comments.
Commit message for "core: add release_state to struct ts_ops" : s/release/released/
Commit message for "core: fix race in handling TA panic" message: s/where/were/

jenswi-linaro:

jforissier:
Added Tested-by: tag from https://github.com/OP-TEE/optee_os/issues/6226#issuecomment-1709417774 and merged manually.

PR #6286, opened 2023-09-06
core: dt_driver: fix error handling in probe_dt_drivers()

GseoC:
Failed probed sequence did not panic if drivers failed to probe.


etienne-lms (commented):
I agree with your change. I would expect core initialization to fail if an expected device driver failed to initialize.

GseoC:
Tag applied :)

PR #6289, opened 2023-09-08
Manifest dt

jenswi-linaro:

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -78,7 +78,7 @@ uint32_t sem_cpu_sync[CFG_TEE_CORE_NB_CORE];
 DECLARE_KEEP_PAGER(sem_cpu_sync);
 #endif
 
-#ifdef CFG_CORE_SEL1_SPMC
+#ifdef CFG_CORE_FFA
 static void *manifest_dt __nex_bss;
 #endif
 
@@ -967,7 +967,7 @@ static struct core_mmu_phys_mem *get_nsec_memory(void *fdt __unused,
 #endif /*CFG_CORE_DYN_SHM*/
 #endif /*!CFG_DT*/
 
-#if defined(CFG_CORE_SEL1_SPMC) || defined(CFG_CORE_SEL2_SPMC)
+#if defined(CFG_CORE_FFA)
 void *get_manifest_dt(void)
 {
 	return manifest_dt;
@@ -1021,7 +1021,7 @@ void *get_manifest_dt(void)
 static void init_manifest_dt(unsigned long pa __unused)
 {
 }
-#endif /*CFG_CORE_SEL1_SPMC && CFG_DT*/
+#endif /*CFG_CORE_FFA*/
 
 #ifdef CFG_CORE_DYN_SHM
 static void discover_nsec_memory(void)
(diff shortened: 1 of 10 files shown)


core/arch/arm/kernel/boot.c
```diff
@@ -1165,12 +1171,18 @@ static bool cpu_nmfi_enabled(void)
  * the unpaged area.
  */
 void __weak boot_init_primary_late(unsigned long fdt,
-				   unsigned long tos_fw_config)
+				   unsigned long manifest_fdt)
```
etienne-lms, line 1174: This local `manifest_fdt` and the global `manifest_dt` quite look alike, it a bit confusing.

etienne-lms (commented):
The second commit does many things. I felt a bit lost in the implication of requiring a manifest DTB in all the SPMC configurations. Is it always read-only, even when `CFG_CORE_SEL1_SPMC=y`?

core/mm/core_mmu.c
```diff
@@ -755,6 +756,8 @@ uint32_t core_mmu_type_to_attr(enum teecore_memtypes t)
 	case MEM_AREA_NSEC_SHM:
 	case MEM_AREA_NEX_NSEC_SHM:
 		return attr | TEE_MATTR_PRW | cached;
+	case MEM_AREA_MANIFEST_DT:
+		return attr | TEE_MATTR_SECURE | TEE_MATTR_PR;
```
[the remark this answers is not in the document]
jenswi-linaro: Something is missing, it should be `cached` since we have coherent memory when using this. I'll fix.

jenswi-linaro:
> The second commit does many things. I felt a bit lost in the implication of requiring a manifest DTB in all the SPMC configurations. Is it always read-only, even when `CFG_CORE_SEL1_SPMC=y`?

I'm sorry, I'll try to break it into smaller pieces. Yes, the manifest is always supposed to be read-only, it's not to be confused with the system DTB.

core/arch/arm/kernel/boot.c
jenswi-linaro: I'll think of something.

jenswi-linaro:
I've split "core: rename tos_fw_config_dt to manifest_dt" into multiple commits.
I've also added "libutils: add runtime_assert()" for better stubbing of `core_mmu_set_secure_memory()` depending on configuration.

LeiSen62:
Acked-by: Leisen <leisen1@huawei.com>

etienne-lms (commented):

jenswi-linaro:
Tags applied.

balint-dobszay-arm:
Hi Jens,
I get a crash when `CFG_CORE_SEL1_SPMC=y` and `CFG_CORE_TPM_EVENT_LOG=y`. After your changes the manifest DT is mapped as RO: https://github.com/OP-TEE/optee_os/pull/6289/commits/aca027763ffaf76c8a36ae787a085b402d212185#diff-e50fbd95497200fea2a8eb3e91c406c07bc915bd20bdc6d76b88bd7c06ee1a74R760

But the TPM event log handling tries to write to it here: https://github.com/OP-TEE/optee_os/blob/master/core/kernel/tpm.c#L50

Making the region RW and/or removing the mentioned `fdt_setprop()` solves the issue. When the DT is in S memory I don't really see a reason to explicitly zero the event log address, so probably removing the write is the better solution.
What do you think?

jenswi-linaro:
I agree, removing the writing makes sense. Do you mind making a PR to fix it or shall I do it?

balint-dobszay-arm:
I'll create a PR. It looks like this only affects the S-EL1 SPMC use case, since otherwise `tpm_map_log_area()` operates on the external DT which is RW.

jenswi-linaro:
Thank you

balint-dobszay-arm:
FYI https://github.com/OP-TEE/optee_os/pull/6310

Issue #6280, opened 2023-09-03
Secure storage object(with PKCS11) usage 


coach-bin:
When we used other secure OS(not OPTEE), a client application had APIs such as write_certificate(idx, data), read_certificate(idx, data). For example, 3 certificates were stored in external storage permanently, and they had own index: 0, 1, 2. so we accessed the certifcates via those index.

As we migrate to OPTEE, I'd like to see if I can use pkcs11 in a same way. (considering to use secure storage instead)

I think the scenario like below. 
1. In initial booting, CA creates 3 persistent objects with C_CreateObject(CKA_LABLE={0,1,2}, CKA_TOKEN=TRUE, CKA_MODIFIABLE=TRUE, CKA_DESTROYABLE=FALSE) : This is just to populate 3 objects, and it will write dummy data and will write real data later.
2. CA manages a map of object index and label. 
3. CA can access the object via index.
   Pseudo code:
   write_certificate(idx,data) {
      label = get_label(idx);
      template = {
         {CKA_LABEL, label, ...},
         {CKA_MODIFIABLE, FALSE, ...} // at this point, make it impossible to modify
      };
      C_CreateObject(template);
   }
   read_certificate(idx,data) {
      label = get_label(idx);
      template = {
         {CKA_LABEL, label, ...},
      };
       C_FindObjects(template);
   }
5. From 2nd booting, CA checks the label corresponding to index 0. If the object corresponding to the label exists, it doesn't create persistent objects any more. And it can writre/read the object via index

Does this make sense? Or is there any concern for this scenario? (Please consider this is first time to try to use pkcs11 :))

etienne-lms:
Hello @coach-bin, you description looks good to me. Using object label is a good way to identify object.

coach-bin:
@etienne-lms  Thanks for reviewing. I'll test this way

coach-bin:
@etienne-lms 
I'm checking the step 3 that I described above, but it's failing. Once I created an object with template below, if I tried to modify its attribute, C_SetAttributeValue() returns CKR_ATTRIBUTE_READ_ONLY, but I can't find any clew. Is there anything that I need to check? I treid CKO_DATA class.

	CK_ATTRIBUTE data_template[] = {
		{ CKA_LABEL, (CK_UTF8CHAR_PTR)label1, strlen(label1) },
		{ CKA_CLASS,	&(CK_OBJECT_CLASS){CKO_DATA},
[...]

PR #6291, opened 2023-09-12
ta: pkcs11: raw data object attributes can be modified

etienne-lms:
Fixes modification permissions for raw data object (PKCS11_CKO_DATA) attributes. These are modifiable, with exception of boolean attributes TOKEN, MODIFIABLE, DESTROYABLE and PRIVATE, that are already handled by attribute_is_modifiable() function.

Link: https://github.com/OP-TEE/optee_os/issues/6280

### DIFF
--- a/ta/pkcs11/src/pkcs11_attributes.c
+++ b/ta/pkcs11/src/pkcs11_attributes.c
@@ -2146,8 +2146,7 @@ static bool attribute_is_modifiable(struct pkcs11_session *session,
 			return true;
 		break;
 	case PKCS11_CKO_DATA:
-		/* None of the data object attributes are modifiable */
-		return false;
+		return true;
 	case PKCS11_CKO_CERTIFICATE:
 		return attr_is_modifiable_certificate(req_attr, session, obj);
 	default:


coach-bin:
Thanks for fixing. Asking just for curiosity, does this modification permission dependent on the HSM implementation? (which means it's not a PKCS11 spec)

vesajaaskelainen:
I have been wondering the wording in the specification what actually can be modified.

I am observing:
http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html
https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html

> 4.2 Common attributes
> (8) May be modified after object is created with a C_SetAttributeValue call, or in the process of copying object with a C_CopyObject call.  However, it is possible that a particular token may not permit modification of the attribute during the course of a C_CopyObject call.

Some entries for (8) are defined in:
> 4.7.2 Overview
> 4.8 Public key objects
> 4.9 Private key objects
> 4.10 Secret key objects

But not elsewhere?

CKA_VALUE is not among things that are modifyable? Only reference for CKA_VALUE to be modified is realtime clock adjustment and I don't think we should support that at all.

In:
> 4.4 Storage Objects

> Only the CKA_LABEL attribute can be modified after the object is created. (The CKA_TOKEN, CKA_PRIVATE, and CKA_MODIFIABLE attributes can be changed in the process of copying an object, however.)

But not specified as (8)

In:
> 4.5 Data objects

There is no any indication that those values can be modified.

coach-bin:
> In:
> 
> > 4.5 Data objects
> 
> There is no any indication that those values can be modified.
This is what I'm curious about : I'd like to believe that this depends on the HSM implementation, but I'm not sure, so I'm asking about it. 

I think if CKA_VALUE attribute is read-only, it should be specified as read only like 4.3.4.2 (Monotonic counter object), but for data object, there is no indication about it.

Also in a perspective of the purpose of using data object:
Data objects hold information defined by an application. Other than providing access to it, Cryptoki does not attach any special meaning to a data object. 
--> Data object usage is defined by an application, so I think the data object should be flexible. I think an application should have the flexibility to set CKA_MODIFIABLE to FALSE when it wants to create an object that cannot be modified after creation, and to TRUE otherwise. I'm a newbie on PKCS11, but I expected it to work that way as a user.


4.3.3.2
The CKA_VALUE attribute may be set using the C_SetAttributeValue function if permitted by the device.

4.3.4.2
CKA_VALUE1
1Read Only
The CKA_VALUE attribute may not be set by the client.

etienne-lms:
I also found the spec is not clear on whether `CKA_VALUE` for a "Data object" can be modified or not: it does not say it is neither it says it is not. From my reading, I understand that the attribute is modifiable as the spec does not say it is not. My statement would apply also to `CKA_APPLICATION` and `CKA_OBJECT_ID` for "Data objects".

That said, I have not strong opinion on whether OP-TEE pkcs11 TA should change it's current behavior or not on that topic. The current implementation does not break the spec.

coach-bin:
@etienne-lms Copying my comment from #6280 

I agree to describe this token restriction in the doc. And I have no strong opinion on whether we need to change current behavior. In my scenario it's good to change this and all I'm worried about is reverting this change again. As this is my first experience with pkcs11 and the spec is ambiguous, I'm cautious.

etienne-lms:
Ok, i'll close this P-R. Thanks for the feedback.

PR #6292, opened 2023-09-12
driver: gic: support sgi raise for gicv3

PeikanTsai:
Use icc_sgi1r and icc_asgi1r to send SGI for gicv3.


core/drivers/gic.c
```diff
@@ -389,22 +389,40 @@ static void gic_it_set_pending(struct gic_data *gd, size_t it)
...
+			SHIFT_U64(mask_aff1, 16) |
+			SHIFT_U64(mask_id, 24) |
+			SHIFT_U64(mask_aff2, 32));
+#else
+	uint32_t mask_group = group & 0x1;
 	uint32_t mask = (mask_id | SHIFT_U32(mask_group, 15) |
 		SHIFT_U32(mask_cpu, 16));
+#endif
 
 	assert(gd == &gic_data);
```
jenswi-linaro, line 412: This function is not easy to read with the last ifdefs. How about moving these asserts to the calling function and make this function:
``` C
{
#if defined(CFG_ARM_GICV3)
...
#else
...
#endif
}
```
with no code outside the ifdef?

core/drivers/gic.c
```diff
@@ -389,22 +389,40 @@ static void gic_it_set_pending(struct gic_data *gd, size_t it)
...
+			     uint8_t cpu_mask, uint8_t group)
 {
 	uint32_t mask_id = it & 0xf;
-	uint32_t mask_group = group & 0x1;
 	uint32_t mask_cpu = cpu_mask & 0xff;
+#if defined(CFG_ARM_GICV3)
+	/* Only support sending SGI to the cores in the same cluster now */
+	uint64_t mpidr = read_mpidr();
+	uint64_t mask_aff1 = (mpidr & MPIDR_AFF1_MASK) >> MPIDR_AFF1_SHIFT;
+	uint64_t mask_aff2 = (mpidr & MPIDR_AFF2_MASK) >> MPIDR_AFF2_SHIFT;
```
jenswi-linaro, line 401: Aff3 missing?

core/arch/arm/include/arm64.h
```diff
@@ -484,6 +484,8 @@ DEFINE_REG_WRITE_FUNC_(icc_eoir0, uint32_t, S3_0_c12_c8_1)
 DEFINE_REG_WRITE_FUNC_(icc_eoir1, uint32_t, S3_0_c12_c12_1)
 DEFINE_REG_WRITE_FUNC_(icc_igrpen0, uint32_t, S3_0_C12_C12_6)
 DEFINE_REG_WRITE_FUNC_(icc_igrpen1, uint32_t, S3_0_C12_C12_7)
+DEFINE_REG_WRITE_FUNC_(icc_sgi1r, uint64_t, S3_0_C12_C11_5)
```
jenswi-linaro, line 487: Please make this change in a separate commit.

PeikanTsai: Got it. I'll push a separate commit for it.

core/drivers/gic.c
PeikanTsai: Agree. I'll update the patch as you suggested.

core/drivers/gic.c
PeikanTsai: I'll add it in the updated patch also, thanks!

jenswi-linaro (commented):

PeikanTsai:
@jenswi-linaro  Could you please submit theses patches? Thanks!

jforissier:
@PeikanTsai merged, thanks.

PR #6293, opened 2023-09-13
crypto: stm32: fix SAES reset at probe time

etienne-lms:
Uses SAES internal reset sequence only when external reset controller is not available. This change fixes a non-systematic SAES error seen when SAES internal reset is triggered right after external reset sequence. Whereas a fixe could be to add a delay between external reset and internal reset sequences, this change simplifies the sequence as internal reset sequence is not needed when SAES instance is reset using its external reset controller.

Fixes: 4320f5cf30c5 ("crypto: stm32: SAES cipher support")

### DIFF
--- a/core/drivers/crypto/stm32/stm32_saes.c
+++ b/core/drivers/crypto/stm32/stm32_saes.c
@@ -1388,8 +1388,8 @@ static TEE_Result stm32_saes_probe(const void *fdt, int node,
 	if (clk_enable(saes_pdata.clk))
 		panic();
 
-	/* External reset of SAES */
 	if (saes_pdata.reset) {
+		/* External reset of SAES */
 		if (rstctrl_assert_to(saes_pdata.reset, TIMEOUT_US_1MS))
 			panic();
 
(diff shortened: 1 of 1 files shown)


core/drivers/crypto/stm32/stm32_saes.c
```diff
@@ -1397,13 +1396,12 @@ static TEE_Result stm32_saes_probe(const void *fdt, int node,
 
 		if (rstctrl_deassert_to(saes_pdata.reset, TIMEOUT_US_1MS))
 			panic();
+	} else {
+		io_setbits32(saes_pdata.base + _SAES_CR, _SAES_CR_IPRST);
+		udelay(SAES_RESET_DELAY);
+		io_clrbits32(saes_pdata.base + _SAES_CR, _SAES_CR_IPRST);
```
jforissier, line 1402: Minor: it would be nice to keep the comments:
```C
if () {
        /* External reset of SAES */
        ...
} else {
        /* Internal reset of SAES */
        ...
}
```

jforissier (approved):
One comment below, please also fix the checkpatch error.

core/drivers/crypto/stm32/stm32_saes.c
etienne-lms: right, thanks.

etienne-lms:
Comments addressed and commit message fixed.
Tag applied.

tbourgoi:
Ok for me, thanks for the fix.
`Acked-by: Thomas Bourgoin <thomas.bourgoin@foss.st.com>`

PR #6294, opened 2023-09-13
plat-k3: drivers: Open TRNG firewall for TIFS

ti-kamlesh:
On devices with PLATFORM=k3-am62x, there is only one SA2UL instance, which is being shared between TIFS and OP-TEE.

Blocking access to TRNG from all other entities other than OP-TEE is causing firewall exception when being accessed by TIFS.

Allow access to TIFS to use SA2UL TRNG along with OP-TEE.

While at it, correct values for FW_SECURE_ONLY and FW_NON_SECURE.

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul.c
@@ -28,6 +28,7 @@
 #define FW_ENABLE_REGION        0x0a
 #define FW_BACKGROUND_REGION    BIT(8)
 #define FW_BIG_ARM_PRIVID       0x01
+#define FW_TIFS_PRIVID          0xca
 #define FW_WILDCARD_PRIVID      0xc3
 #define FW_SECURE_ONLY          GENMASK_32(7, 0)
 #define FW_NON_SECURE           GENMASK_32(15, 0)
@@ -46,6 +47,7 @@ static TEE_Result sa2ul_init(void)
 	uint16_t owner_permission_bits = 0;
 	uint32_t control = 0;
 	uint32_t permissions[FWL_MAX_PRIVID_SLOTS] = { };
+	uint32_t num_perm = 0;
 	uint64_t start_address = 0;
 	uint64_t end_address = 0;
 	uint32_t val = 0;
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -28,9 +28,10 @@
 #define FW_ENABLE_REGION        0x0a
 #define FW_BACKGROUND_REGION    BIT(8)
 #define FW_BIG_ARM_PRIVID       0x01
+#define FW_TIFS_PRIVID          0xca
 #define FW_WILDCARD_PRIVID      0xc3
-#define FW_SECURE_ONLY          GENMASK_32(8, 0)
-#define FW_NON_SECURE           GENMASK_32(16, 0)
+#define FW_SECURE_ONLY          GENMASK_32(7, 0)
+#define FW_NON_SECURE           GENMASK_32(15, 0)
```
manorit2001, line 34: Hi Kamlesh, I see that although both the fixes are related but are still not apt to come in a same commit. Probably split this fix out into a separate commit.

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -116,10 +117,15 @@ static TEE_Result sa2ul_init(void)
 
 	/* Modify TRNG firewall to block all others access */
 	control = FW_ENABLE_REGION;
-	permissions[0] = (FW_BIG_ARM_PRIVID << 16) | FW_SECURE_ONLY;
 	start_address = RNG_BASE;
 	end_address = RNG_BASE + RNG_REG_SIZE - 1;
+	permissions[0] = (FW_BIG_ARM_PRIVID << 16) | FW_SECURE_ONLY;
+#if defined(PLATFORM_FLAVOR_am62x)
+	permissions[1] = (FW_TIFS_PRIVID << 16) | FW_NON_SECURE;
+	ret = ti_sci_set_fwl_region(fwl_id, rng_region, 2,
```
manorit2001, line 125: A stylistic comment but we should not split the function call like this when we can create a variable `num_permissions` and that might look cleaner.

manorit2001 (changes requested):

core/arch/arm/plat-k3/drivers/sa2ul.c
jenswi-linaro: +1

glneo:
> On devices with PLATFORM=k3-am62x, there is only one SA2UL instance, which is being shared between TIFS and OP-TEE.

This is true for several TI platforms. You should be specific here on why you want this only for AM62.

glneo:
> Hello, out of curiosity, how do you manage concurrent accesses to you TRNG entropy? Does your SA2UL module handles it so it can always give entropy to the requester?

No, it is still single user at a time. The only time TIFS uses the TRNG is during the low power mode exit sequence, and at that time OP-TEE is not active, the rest of the time it is exclusively used by OP-TEE. TIFS is our security controller, and already has control of all firewalls so it can be (has to be) trusted to do the right thing and not use its TRNG access to subvert OP-TEE's security.

manorit2001 (approved):
`Reviewed-by: Manorit Chawdhry <m-chawdhry@ti.com>`

manorit2001:
Hi @jforissier , would this change be in 4.0.0 tag?

jforissier:
> Hi @jforissier , would this change be in 4.0.0 tag?

Yes it will :)

PR #6297, opened 2023-09-15
core: io: IO_READ32_POLL_TIMEOUT()

xiaoxuZeng:
Implement Polling Read Register Interface inspired by linux kernel.

### DIFF
--- a/core/include/io.h
+++ b/core/include/io.h
@@ -273,4 +273,31 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
 	io_write8(addr, (io_read8(addr) & ~clear_mask) | set_mask);
 }
 
+/*
+ * Poll on a IO memory content or timeout
+ *
+ * @_addr is the address of the memory cell accessed
+ * @_val represents the val of the memory cell accessed
+ * @_cond represents the condition to get the correct value
+ * @_delay_us represents the read interval in mircorseconds
+ * @_timeout_us represents the timeout period in microseconds
+ *
+ * @return nonzero value means timeout, 0 means got right value
+ */
+#define IO_READ32_POLL_TIMEOUT(_addr, _val, _cond, _delay_us, _timeout_us) \
+	({ \
+		uint32_t __timeout = 0; \
+		uint32_t __delay = (_delay_us); \
+		\
+		while (__timeout < (_timeout_us)) { \
+			(_val) = io_read32(_addr); \
+			if (_cond) \
+				break; \
+			__timeout += (__delay); \
+			udelay(__delay); \
+		} \
+		(_val) = io_read32(_addr); \
+		!(_cond); \
+	})
+
 #endif /*IO_H*/


core/include/io.h
```diff
@@ -273,4 +273,22 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+	while (timeout < (timeout_us)) { \
+		(val) = io_read32(addr); \
+		if (cond) { \
+			flag = true; \
+			break; \
+		} \
+		timeout += (_delay_us); \
+		udelay(_delay_us); \
+	} \
+	(flag) ? HISI_QM_DRVCRYPT_NO_ERR : HISI_QM_DRVCRYPT_ETMOUT; \
```
jforissier, line 291: You cannot use platform specific status codes here. `true` and `false` may be better options.

core/include/io.h
```diff
@@ -273,4 +273,22 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
 	io_write8(addr, (io_read8(addr) & ~clear_mask) | set_mask);
 }
 
+#define readl_poll_timeout(addr, val, cond, delay_us, timeout_us) \
```
etienne-lms, line 276: An inline description comment is welcome.
Perfer use uppercase characters for macros (`READL_POLL_TIMEOUT()`)
As this relies on `io_read32()`, I think macro name `IO_READ32_POLL_TIMEOUT()` would be better.

core/include/io.h
```diff
@@ -273,4 +273,22 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+({ \
+	uint64_t timeout = 0; \
+	uint64_t _delay_us = delay_us; \
+	bool flag = false; \
```
etienne-lms, line 280: Variables `timeout` and `flag` may conflict with variables with same label already defined in caller function. Could you add an underscore prefix to address that, as you did for `_delay_us` (`_timeout`, `_flag`)?

core/include/io.h
```diff
@@ -273,4 +273,22 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+	while (timeout < (timeout_us)) { \
+		(val) = io_read32(addr); \
+		if (cond) { \
+			flag = true; \
+			break; \
+		} \
+		timeout += (_delay_us); \
+		udelay(_delay_us); \
+	} \
+	(flag) ? true : false; \
```
etienne-lms, line 291: flag is already a boolean:
```patch
		udelay(_delay_us); \
 	} \
-	(flag) ? true : false; \
+	flag; \
 })

File: core/include/io.h [hunk 5]
```diff
@@ -273,4 +273,33 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+	while (_timeout < (timeout_us)) { \
+		(val) = io_read32(addr); \
+		if (cond) { \
+			_flag = false; \
+			break; \
+		} \
+		_timeout += (_delay_us); \
+		udelay(_delay_us); \
+	} \
+	(_flag); \
```
etienne-lms, line 302: If the thread from where `IO_READ32_POLL_TIMEOUT()` is called is rescedule by non-secure world, we may timeout while the target IO memory cell is at the expected state. Could you add a final test on `cond` test once out of the loop? Something like:
```patch
 		_timeout += (_delay_us); \
 		udelay(_delay_us); \
 	} \
+	(val) = io_read32(addr); \
+	if (cond) { \
+		_flag = false; \
 	(_flag); \
```

Note: can remove the parentheses around `flags` in the last line.

core/include/io.h
```diff
@@ -273,4 +273,37 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+ * @delay_us represents the read interval in mircorseconds
+ * @_timeout_us represents the time-out period in microseconds
+ *
+ * @return true means time-out, false means got right value
+ */
+#define IO_READ32_POLL_TIMEOUT(addr, val, cond, delay_us, _timeout_us) \
+	({ \
+		uint64_t _timeout = 0; \
+		uint64_t _delay_us = delay_us; \
+		bool _flag = true; \
```
jenswi-linaro, line 291: I'd prefer `__ret` or `__rc`. We should probably prefix the locals with `__` instead.

core/include/io.h
```diff
@@ -273,4 +273,37 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+ * @cond represents the condition to get the correct value
+ * @delay_us represents the read interval in mircorseconds
+ * @_timeout_us represents the time-out period in microseconds
+ *
+ * @return true means time-out, false means got right value
+ */
+#define IO_READ32_POLL_TIMEOUT(addr, val, cond, delay_us, _timeout_us) \
+	({ \
+		uint64_t _timeout = 0; \
+		uint64_t _delay_us = delay_us; \
```
jenswi-linaro, line 290: Please use `(delay_us)`.

core/include/io.h
```diff
@@ -273,4 +273,37 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+				_flag = false; \
+				break; \
+			} \
+			_timeout += (_delay_us); \
+			udelay(_delay_us); \
+		} \
+		(val) = io_read32(addr); \
+		if (cond) { \
+			_flag = false; \
+		} \
```
etienne-lms, line 305: Could simply:
```c
#define IO_READ32_POLL_TIMEOUT(_addr, _val, _cond, _delay_us, _timeout_us) \
	({ \
		uint64_t __timeout = 0; \
		uint64_t __delay = (_delay_us); \
		\
		while (__timeout < (_timeout_us)) { \
			(_val) = io_read32(_addr); \
			if (_cond) \
				break; \
			__timeout += (__delay); \
			udelay(__delay); \
		} \
		(_val) = io_read32(_addr); \
		(_cond); \
	})

Comment (jenswi-linaro) [reply core/include/io.h:305]: I'm not sure we want to evaluate `_cond` more times than necessary.

Comment (etienne-lms) [reply core/include/io.h:305]: We need to evaluate `cond` even once timeout has elapsed: https://github.com/OP-TEE/optee_os/pull/6297#discussion_r1330511820.
Otherwise, it will be the caller responsibility to check again its condition in case the macro returns `true` and the caller executes in a theraded context with foreign interrupts unmasked.

Comment (jenswi-linaro) [reply core/include/io.h:305]: Got it, good point.

Comment (xiaoxuZeng) [reply core/include/io.h]: > Could **~simply~ simplify**:
> 
> ```c
> #define IO_READ32_POLL_TIMEOUT(_addr, _val, _cond, _delay_us, _timeout_us) \
> 	({ \
> 		uint64_t __timeout = 0; \
> 		uint64_t __delay = (_delay_us); \
> 		\
> 		while (__timeout < (_timeout_us)) { \
> 			(_val) = io_read32(_addr); \
> 			if (_cond) \
> 				break; \
> 			__timeout += (__delay); \
> 			udelay(__delay); \
> 		} \
> 		(_val) = io_read32(_addr); \
> 		(_cond); \
> 	})
> ```

It's awesome, far better than the my version!!!

File: core/include/io.h (hunk 3 above)
Comment (xiaoxuZeng) [reply core/include/io.h]:
```c
CHECK: Macro argument reuse '_addr' - possible side-effects?
```
checkpatch got such warning,`_val` and `_cond' have same warning.

Maybe i should change such argument  to addr_?

xiaoxuZeng:
I noticed that the `checkpatch` picked up some alerts. but I think that has little side-effects.

```c
CHECK: Macro argument reuse 'cond_' - possible side-effects?
#32: FILE: core/include/io.h:2[87](https://github.com/OP-TEE/optee_os/actions/runs/6269566318/job/17026189382?pr=6297#step:5:88):
+#define IO_READ32_POLL_TIMEOUT(addr_, val_, cond_, delay_us_, timeout_us_) \
+	({ \
+		uint64_t __timeout = 0; \
+		uint64_t __delay = (delay_us_); \
+		\
+		while (__timeout < (timeout_us_)) { \
+			(val_) = io_read32(addr_); \
+			if (cond_) \
+				break; \
+			__timeout += (__delay); \
+			udelay(__delay); \
+		} \
+		(val_) = io_read32(addr_); \
+		!(cond_); \
+	})
```

core/include/io.h
```diff
@@ -273,4 +273,31 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
 	io_write8(addr, (io_read8(addr) & ~clear_mask) | set_mask);
 }
 
+/*
+ * Poll on a IO memory content or timeout
+ *
+ * @_addr is the address of the memory cell accessed
+ * @_val represents the val  of the memory cell accessed
```
etienne-lms, line 280: replace the 2 spaces chars with a single one

core/include/io.h
```diff
@@ -273,4 +273,31 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+ * @_cond represents the condition to get the correct value
+ * @_delay_us represents the read interval in mircorseconds
+ * @_timeout_us represents the time-out period in microseconds
+ *
+ * @return nonzero value means time-out, 0 means got right value
+ */
+#define IO_READ32_POLL_TIMEOUT(_addr, _val, _cond, _delay_us, _timeout_us) \
```
etienne-lms, line 287: really need `_delay_us` argument?

core/include/io.h
```diff
@@ -273,4 +273,31 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
 }
 
+/*
+ * Poll on a IO memory content or timeout
+ *
+ * @_addr is the address of the memory cell accessed
+ * @_val represents the val  of the memory cell accessed
+ * @_cond represents the condition to get the correct value
+ * @_delay_us represents the read interval in mircorseconds
+ * @_timeout_us represents the time-out period in microseconds
```
etienne-lms, line 283: Prefer `timeout` instead of `time-out`. More consistent within optee_os source files.
Here and at line below.

core/include/io.h
```diff
@@ -273,4 +273,31 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
...
+ *
+ * @return nonzero value means time-out, 0 means got right value
+ */
+#define IO_READ32_POLL_TIMEOUT(_addr, _val, _cond, _delay_us, _timeout_us) \
+	({ \
+		uint64_t __timeout = 0; \
+		uint64_t __delay = (_delay_us); \
```
etienne-lms, line 290: Prefer `uint32_t`: `udelay()` argument is `uint32_t`.
`__timeout` also could be `uint32_t`.

etienne-lms (commented):
> CHECK: Macro argument reuse 'cond_' - possible side-effects?

We already faced such CHECK reports. I think it is ok and we can consider the report as a false positive.

Could you update commit message: s/readl_poll_timeout()/IO_READ32_POLL_TIMEOUT()/ ?

core/include/io.h
xiaoxuZeng: yeah，I just discussed it with workmates, we found different hardward' delay time is different, sec module would be 10 ms, but some hardware 5 ms is enought.

etienne-lms (commented):

jenswi-linaro:

etienne-lms (approved):
OK to merge?

jforissier:
> OK to merge?

Yes, thanks for reminding.

PR #6298, opened 2023-09-15
zynqmp: remove redundant platform config code

ierkiaga:
The hardware description is identical in all the platforms, there is no need for specific ultra96 code to define base addresses.

### DIFF
--- a/core/arch/arm/plat-zynqmp/platform_config.h
+++ b/core/arch/arm/plat-zynqmp/platform_config.h
@@ -57,22 +57,8 @@
 	defined(PLATFORM_FLAVOR_zc1751_dc2) || \
 	defined(PLATFORM_FLAVOR_zcu102) || \
 	defined(PLATFORM_FLAVOR_zcu104) || \
-	defined(PLATFORM_FLAVOR_zcu106)
-
-#define GIC_BASE		0xF9010000
-#define UART0_BASE		0xFF000000
-#define UART1_BASE		0xFF010000
-
-#define IT_UART0		53
-#define IT_UART1		54
-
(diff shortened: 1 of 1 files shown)


jbech-linaro:
`Acked-by: Joakim Bech <joakim.bech@linaro.org>`

vesajaaskelainen:
Actually Ultra96 has different setup for UARTs?

Did you try with Ultra96 too? Because it might be that Ultra96 config is just wrong.

ierkiaga:
> Actually Ultra96 has different setup for UARTs?
> 
Yes, but that's handled in the config.mk file where the CFG_UART_BASE is set to either UART0_BASE or UART1_BASE.

> Did you try with Ultra96 too? Because it might be that Ultra96 config is just wrong.
No but code inspection should be enough here. Both conditional compilation statements are identical so no need for duplicating code. As you can see in both cases is defining the base address/offset/interruptid of the same peripherals.

vesajaaskelainen:
I agree that it looks correct change. I recall we had some issues with Ultra96 bootup at some point of time but never finalized the debugging.

I suppose best action is to clean it up as proposed. And then fix Ultra96 support later if there is a need.

```
Acked-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>
```

ierkiaga:
Thanks @vesajaaskelainen,
I'm planning to do some updates to the build repository so I will test the Ultra96v1 board soon. I do recall some information about issues with v2 board but as I do not have v2 will not be able to test it.

jenswi-linaro:

PR #6300, opened 2023-09-18
Introduce remote proc trusted application

arnopo:
This series introduces the remoteproc Trusted Application, which enables the authentication of signed images through the scripts/sign_rproc_fw.py script. 

The remote proc TA depends on a remoteproc PTA that must implement platform-specific requirements. To achieve this, a remoteproc PTA API is also introduced.

A ELF32 parser is also introduced. This parser, inspired from the Linux remoteproc parser,  get loadable sections and the resource table specific section.

Related pull request:
- https://github.com/OP-TEE/optee_os/pull/6027

Code associated to the Linux Kernel is proposed as RFC for upstream:

### DIFF
--- a/ta/remoteproc/sub.mk
+++ b/ta/remoteproc/sub.mk
@@ -1,0 +1,3 @@
+global-incdirs-y += include
+global-incdirs-y += src
+subdirs-y += src
--- a/lib/libutee/include/utee_defines.h
+++ b/lib/libutee/include/utee_defines.h
@@ -302,12 +302,18 @@ static inline size_t __tee_alg_get_digest_size(uint32_t algo)
 #endif
 
 /* If we we're on a big endian platform we'll have to update these */
-#define TEE_U64_FROM_BIG_ENDIAN(x)  TEE_U64_BSWAP(x)
-#define TEE_U32_FROM_BIG_ENDIAN(x)  TEE_U32_BSWAP(x)
-#define TEE_U16_FROM_BIG_ENDIAN(x)  TEE_U16_BSWAP(x)
-#define TEE_U64_TO_BIG_ENDIAN(x)    TEE_U64_BSWAP(x)
-#define TEE_U32_TO_BIG_ENDIAN(x)    TEE_U32_BSWAP(x)
-#define TEE_U16_TO_BIG_ENDIAN(x)    TEE_U16_BSWAP(x)
+#define TEE_U64_FROM_LITTLE_ENDIAN(x)  ((uint64_t)(x))
+#define TEE_U32_FROM_LITTLE_ENDIAN(x)  ((uint32_t)(x))
+#define TEE_U16_FROM_LITTLE_ENDIAN(x)  ((uint16_t)(x))
+#define TEE_U64_TO_LITTLE_ENDIAN(x)    ((uint64_t)(x))
+#define TEE_U32_TO_LITTLE_ENDIAN(x)    ((uint32_t)(x))
+#define TEE_U16_TO_LITTLE_ENDIAN(x)    ((uint16_t)(x))
+#define TEE_U64_FROM_BIG_ENDIAN(x)     TEE_U64_BSWAP(x)
+#define TEE_U32_FROM_BIG_ENDIAN(x)     TEE_U32_BSWAP(x)
+#define TEE_U16_FROM_BIG_ENDIAN(x)     TEE_U16_BSWAP(x)
+#define TEE_U64_TO_BIG_ENDIAN(x)       TEE_U64_BSWAP(x)
+#define TEE_U32_TO_BIG_ENDIAN(x)       TEE_U32_BSWAP(x)
+#define TEE_U16_TO_BIG_ENDIAN(x)       TEE_U16_BSWAP(x)
 
 #define TEE_TIME_MILLIS_BASE    1000
 
--- a/ta/remoteproc/Makefile
+++ b/ta/remoteproc/Makefile
@@ -1,0 +1,18 @@
+# The UUID for the Trusted Application
+BINARY=80a4c275-0a47-4905-8285-1486a9771a08
+
+ifdef TA_CROSS_COMPILE
+CROSS_COMPILE ?= $(TA_CROSS_COMPILE)
+endif
+export CROSS_COMPILE
+
+CFG_TEE_TA_LOG_LEVEL ?= 2
+CPPFLAGS += -DCFG_TEE_TA_LOG_LEVEL=$(CFG_TEE_TA_LOG_LEVEL)
+
+-include $(TA_DEV_KIT_DIR)/mk/ta_dev_kit.mk
+
+ifeq ($(wildcard $(TA_DEV_KIT_DIR)/mk/ta_dev_kit.mk), )
+clean:
+	@echo 'Note: $$(TA_DEV_KIT_DIR)/mk/ta_dev_kit.mk not found, cannot clean TA'
+	@echo 'Note: TA_DEV_KIT_DIR=$(TA_DEV_KIT_DIR)'
+endif
--- a/ta/remoteproc/include/ta_remoteproc.h
+++ b/ta/remoteproc/include/ta_remoteproc.h
@@ -1,0 +1,64 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2023, STMicroelectronics
+ */
+
+#ifndef TA_REMOTEPROC_H
+#define TA_REMOTEPROC_H
+
+/*
+ * This UUID is generated with uuidgen
+ * the ITU-T UUID generator at http://www.itu.int/ITU-T/asn1/uuid.html
+ */
+#define TA_REMOTEPROC_UUID { 0x80a4c275, 0x0a47, 0x4905, \
+			     { 0x82, 0x85, 0x14, 0x86, 0xa9, 0x77, 0x1a, 0x08} }
+
+/* The command IDs implemented in this TA */
+
+/*
+ * Authentication of the firmware and load in the remote processor memory.
+ *
+ * [in]  params[0].value.a:	Unique 32bit remote processor identifier
+ * [in]  params[1].memref:	buffer containing the image of the firmware
+ */
+#define TA_RPROC_CMD_LOAD_FW		1
+
+/*
+ * Start the remote processor.
+ *
+ * [in]  params[0].value.a:	Unique 32bit remote processor identifier
+ */
+#define TA_RPROC_CMD_START_FW		2
+
+/*
+ * Stop the remote processor.
+ *
+ * [in]  params[0].value.a:	Unique 32bit remote processor identifier
+ */
+#define TA_RPROC_CMD_STOP_FW		3
+
+/*
+ * Return the physical address of the resource table, or 0 if not found
+ * No check is done to verify that the address returned is accessible by the
+ * non-secure world. If the resource table is loaded in a protected memory,
+ * then accesses from non-secure world will likely fail.
+ *
+ * [in]  params[0].value.a:	Unique 32bit remote processor identifier
+ * [out] params[1].value.a:	32bit LSB resource table memory address
+ * [out] params[1].value.b:	32bit MSB resource table memory address
+ * [out] params[2].value.a:	32bit LSB resource table memory size
+ * [out] params[2].value.b:	32bit MSB resource table memory size
+ */
+#define TA_RPROC_CMD_GET_RSC_TABLE	4
+
+/*
+ * Get remote processor firmware core dump. If found, return either
+ * TEE_SUCCESS on successful completion or TEE_ERROR_SHORT_BUFFER if output
+ * buffer is too short to store the core dump.
+ *
+ * [in]  params[0].value.a:	Unique 32bit remote processor identifier
+ * [out] params[1].memref:	Core dump, if found
+ */
+#define TA_RPROC_CMD_GET_COREDUMP	5
+
+#endif /*TA_REMOTEPROC_H*/
--- a/ta/remoteproc/src/elf_parser.c
+++ b/ta/remoteproc/src/elf_parser.c
@@ -1,0 +1,175 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2023, STMicroelectronics
+ */
+
+#include <assert.h>
+#include <elf_parser.h>
+#include <string.h>
+#include <trace.h>
+#include <types_ext.h>
+#include <util.h>
+
+static bool va_in_fwm_image_range(void *va, uint8_t *fw, size_t fw_size)
+{
+	uint8_t *vaddr = va;
+
+	return vaddr >= fw && vaddr < fw + fw_size;
+}
+
+TEE_Result e32_parse_ehdr(uint8_t *fw, size_t size)
+{
+	Elf32_Ehdr *ehdr = (Elf32_Ehdr *)(void *)fw;
+
+	if (!fw || !IS_ALIGNED_WITH_TYPE(fw, uint32_t)) {
+		EMSG("Invalid firmware address");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	if (size < sizeof(Elf32_Ehdr) ||
+	    size < (ehdr->e_shoff + sizeof(Elf32_Shdr)))
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (!IS_ELF(*ehdr) ||
+	    ehdr->e_ident[EI_VERSION] != EV_CURRENT ||
+	    ehdr->e_ident[EI_CLASS] != ELFCLASS32 ||
+	    ehdr->e_phentsize != sizeof(Elf32_Phdr) ||
+	    ehdr->e_shentsize != sizeof(Elf32_Shdr)) {
+		EMSG("Invalid header");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	if (ehdr->e_phnum == 0) {
+		EMSG("No loadable segment found");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result e32_parser_load_elf_image(uint8_t *fw, size_t fw_size,
+				     TEE_Result (*load_seg)(uint8_t *src,
+							    uint32_t size,
+							    uint32_t da,
+							    uint32_t mem_size,
+							    void *priv),
+				     void *priv_data)
+{
+	Elf32_Ehdr *ehdr = (Elf32_Ehdr *)(void *)fw;
+	Elf32_Phdr *phdr = (void *)((int8_t *)ehdr + ehdr->e_phoff);
+	TEE_Result res = TEE_SUCCESS;
+	unsigned int i = 0;
+
+	if (!load_seg || fw + fw_size <= fw)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!IS_ALIGNED_WITH_TYPE(phdr, uint32_t) ||
+	    !va_in_fwm_image_range(phdr, fw, fw_size))
+		return TEE_ERROR_BAD_FORMAT;
+
+	for (i = 0; i < ehdr->e_phnum; i++, phdr++) {
+		uint32_t dst = phdr->p_paddr;
+		uint8_t *src = NULL;
+
+		if (phdr->p_type != PT_LOAD)
+			continue;
+
+		if (!va_in_fwm_image_range((void *)((vaddr_t)(phdr + 1) - 1),
+					   fw, fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		src = (uint8_t *)fw + phdr->p_offset;
+
+		if (!va_in_fwm_image_range(src, fw, fw_size) ||
+		    !va_in_fwm_image_range(src + phdr->p_filesz, fw, fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		res = load_seg(src, phdr->p_filesz, dst, phdr->p_memsz,
+			       priv_data);
+		if (res)
+			return res;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result e32_parser_find_rsc_table(uint8_t *fw, size_t fw_size,
+				     Elf32_Addr *rsc_addr,
+				     Elf32_Word *rsc_size)
+{
+	Elf32_Shdr *shdr = NULL;
+	unsigned int i = 0;
+	char *name_table = NULL;
+	struct resource_table *table = NULL;
+	Elf32_Ehdr *ehdr = (Elf32_Ehdr *)(void *)fw;
+	uint8_t *elf_data = fw;
+
+	if (fw + fw_size <= fw || fw + ehdr->e_shoff < fw)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	shdr = (void *)(fw + ehdr->e_shoff);
+	if (!IS_ALIGNED_WITH_TYPE(shdr, uint32_t) ||
+	    !va_in_fwm_image_range(shdr, fw, fw_size))
+		return TEE_ERROR_BAD_FORMAT;
+
+	name_table = (char *)elf_data + shdr[ehdr->e_shstrndx].sh_offset;
+	if (!va_in_fwm_image_range(name_table, fw, fw_size))
+		return TEE_ERROR_BAD_FORMAT;
+
+	for (i = 0; i < ehdr->e_shnum; i++, shdr++) {
+		size_t size = shdr->sh_size;
+		size_t offset = shdr->sh_offset;
+		size_t s = 0;
+
+		if (!va_in_fwm_image_range(shdr, fw, fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		if (strcmp(name_table + shdr->sh_name, ".resource_table"))
+			continue;
+
+		if (!shdr->sh_size)
+			return TEE_ERROR_NO_DATA;
+
+		if (offset + size > fw_size || offset + size < size) {
+			EMSG("Resource table truncated");
+			return TEE_ERROR_BAD_FORMAT;
+		}
+
+		if (sizeof(struct resource_table) > size) {
+			EMSG("No header found in resource table");
+			return TEE_ERROR_BAD_FORMAT;
+		}
+
+		table = (struct resource_table *)(void *)(elf_data + offset);
+		if (!IS_ALIGNED_WITH_TYPE(table, uint32_t))
+			return TEE_ERROR_CORRUPT_OBJECT;
+
+		if (table->ver != 1) {
+			EMSG("Unsupported firmware version %"PRId32,
+			     table->ver);
+			return TEE_ERROR_BAD_FORMAT;
+		}
+
+		if (table->reserved[0] || table->reserved[1]) {
+			EMSG("Non zero reserved bytes");
+			return TEE_ERROR_BAD_FORMAT;
+		}
+
+		if (MUL_OVERFLOW(table->num, sizeof(*table->offset), &s) ||
+		    ADD_OVERFLOW(s, sizeof(struct resource_table), &s) ||
+		    s > size) {
+			EMSG("Resource table incomplete");
+			return TEE_ERROR_BAD_FORMAT;
+		}
+
+		DMSG("Resource table address %#"PRIx32", size %"PRIu32,
+		     shdr->sh_addr, shdr->sh_size);
+
+		*rsc_addr = shdr->sh_addr;
+		*rsc_size = shdr->sh_size;
+
+		return TEE_SUCCESS;
+	}
+
+	return TEE_ERROR_NO_DATA;
+}
--- a/ta/remoteproc/src/elf_parser.h
+++ b/ta/remoteproc/src/elf_parser.h
@@ -1,0 +1,91 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2023, STMicroelectronics
+ */
+
+#ifndef ELF_PARSER_H
+#define ELF_PARSER_H
+
+#include <elf32.h>
+#include <stdint.h>
+#include <tee_api_types.h>
+
+/**
+ * struct resource_table - firmware resource table header
+ * @ver: version number
+ * @num: number of resource entries
+ * @reserved: reserved (must be zero)
+ * @offset: array of offsets pointing at the various resource entries
+ *
+ * A resource table is essentially a list of system resources required
+ * by the remote processor. It may also include configuration entries.
+ * If needed, the remote processor firmware should contain this table
+ * as a dedicated ".resource_table" ELF section.
+ *
+ * This structure shall be consistent with the Linux kernel structure
+ * definition from include/linux/remoteproc.h.
+ *
+ * The resource_table structure does not need to be packed in OP-TEE. The ELF
+ * parser does not access the system resources structures that follow this
+ * structure in shared memory (no memory alignment constraint on structures).
+ */
+struct resource_table {
+	uint32_t ver;
+	uint32_t num;
+	uint32_t reserved[2];
+	uint32_t offset[];
+};
+
+struct fw_elf32 {
+	uintptr_t e_entry;
+	uintptr_t e_phoff;
+	uintptr_t e_shoff;
+	uint32_t e_phnum;
+	uint32_t e_shnum;
+	uint32_t e_phentsize;
+	uint32_t e_shentsize;
+
+	Elf32_Phdr *phdr;
+	Elf32_Shdr *shdr;
+};
+
+/*
+ * e32_parse_ehdr() - Check and parse the ELF header
+ *
+ * @fw:   Firmware ELF file image
+ * @size: Byte size of firmware ELF file image
+ * Return TEE_SUCCESS or appropriate error.
+ */
+TEE_Result e32_parse_ehdr(uint8_t *fw, size_t size);
+
+/*
+ * e32_parser_load_elf_image() - simple ELF loader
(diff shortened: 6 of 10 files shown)


ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1079 @@
...
+
+	for (i = 0; i < nb_entry; peh++, i++) {
+		if (peh->phdr.p_paddr != da)
+			continue;
+
+		/*
+		 * Segment is read from a non secure memory. Crosscheck it using
+		 * the hash table to verify that the segment has not been
+		 * corrupted.
+		 */
```
etienne-lms, line 672: I think this inline comment is misplaced or misphrased. It explains why we need a hash table, which rather belongs to `remoteproc_load_segment()` function, right below.
I think here the inline comment could state that segment info are crosschecked (load addres, load size, segment size, ...) against same info saved in hash table segment (read here from secure memory).

Suggestion (IIUC):
```
		/*
		 * Segment metadata were read from a non-secure memory.
                 * Validate them using hash table data stored in secure memory.
		 */

File: ta/remoteproc/remoteproc_core.c [hunk 2]
```diff
@@ -0,0 +1,1079 @@
...
+	unsigned int nb_entry = ctx->nb_segment;
+
+	peh = (void *)(ctx->hash_table);
+
+	for (i = 0; i < nb_entry; peh++, i++) {
+		if (peh->phdr.p_paddr != da)
+			continue;
+
+		/*
+		 * Segment is read from a non secure memory. Crosscheck it using
```
etienne-lms, line 669: s/non secure/non-secure/  (for consistency)

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1079 @@
...
+	if (res != TEE_SUCCESS)
+		goto out;
+
+	/*
+	 * This implementation could be enhanced by providing alternative to
+	 * verify the signature in the TA. This could be done for instance by
+	 * getting the key object from secure storage.
+	 */
+
+	/* By default ask the pta to verify the signature. */
```
etienne-lms, line 420: /pta/remoteproc PTA/

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1079 @@
...
+		 * corrupted.
+		 */
+		if (peh->phdr.p_type != PT_LOAD)
+			return TEE_ERROR_BAD_PARAMETERS;
+
+		if (peh->phdr.p_filesz != size || peh->phdr.p_memsz != mem_size)
+			return TEE_ERROR_BAD_PARAMETERS;
+
+		if (src < ctx->fw_img ||
+		    (src + peh->phdr.p_filesz) > (ctx->fw_img + ctx->fw_img_sz))
```
etienne-lms, line 680: should check `src + peh->phdr.p_filesz` agianst overflow?

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1079 @@
...
+	params[2].value.a = da;
+	params[3].memref.buffer = hash;
+	params[3].memref.size = TEE_SHA256_HASH_SIZE;
+
+	if (size) {
+		res = TEE_InvokeTACommand(pta_session, TEE_TIMEOUT_INFINITE,
+					  PTA_RPROC_LOAD_SEGMENT_SHA256,
+					  param_types, params, NULL);
+		if (res != TEE_SUCCESS) {
+			EMSG("Fails to load segment, res = 0x%x", res);
```
etienne-lms, line 728: `... res = %#"PRIx32, res)`
Ditto at line 747

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1079 @@
...
+			EMSG("Fails to clear segment, res = 0x%x", res);
+	}
+
+	return res;
+}
+
+static TEE_Result remoteproc_load_elf(struct remoteproc_context *ctx)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	uint8_t num_img = 0, i = 0;
```
etienne-lms, line 756:
```c
	uint8_t num_img = 0;
	uint8_t i = 0;
```
Note that `num_img` and `i` could be `unsigned int` or `size_t`.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1079 @@
...
+ *                               ...
+ *                        +-------------+
+ *                        |   Firmware  |
+ *                        |    image n  |
+ *                        +-------------+
+ */
+
+/* Firmware state */
+enum remoteproc_state {
+	REMOTEPROC_OFF,
```
etienne-lms, line 60: explicit `= 0` since it is the expected default state when context is allocated (and zero initiallized)?

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,166 @@
...
+ * Interface to the pseudo TA which provides platform implementation
+ * of the remote processor management
+ */
+
+#define PTA_RPROC_UUID { 0x54af4a68, 0x19be, 0x40d7, \
+		{ 0xbb, 0xe6, 0x89, 0x50, 0x35, 0x0a, 0x87, 0x44 } }
+
+/* Firmware format */
+#define PTA_RPROC_PROPRIETARY_FMT	BIT32(0)
+#define PTA_RPROC_ELF_FMT		BIT32(1)
```
etienne-lms, line 22: Maybe sing enumerated values above and below would better clarify which capability bits are expected from `PTA_RPROC_HW_CAPABILITIES` arguments `value[1].a` and `value[2].a`.

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,166 @@
...
+
+/* Firmware image protection */
+/* The platform supports copy of the input firmware image in secure memory */
+#define PTA_RPROC_FW_SECURE_COPY	BIT32(0)
+/* The platform supports load of segment with hash protection */
+#define PTA_RPROC_FW_WITH_HASH_TABLE	BIT32(1)
+/* The platform is able to change access to secure the firmware input image */
```
etienne-lms, line 29: a bit unclear.
`The platform is can restrict access remoteproc firmware memories`?

etienne-lms (commented):
"ta: add remote processor firmware ELF parser" and 
"ta: add the support of the start and stop command in remoteproc TA".

arnopo (commented):
Thanks @etienne-lms for the review

ta/remoteproc/remoteproc_core.c
```diff
@@ -540,6 +544,64 @@ static TEE_Result remoteproc_verify_firmware(struct remoteproc_context *ctx,
...
+	params[0].value.a = ctx->fw_id;
+	params[1].value.a = da;
+	params[2].value.a = size;
+
+	res = TEE_InvokeTACommand(pta_session, TEE_TIMEOUT_INFINITE,
+				  PTA_RPROC_FIRMWARE_DA_TO_PA,
+				  param_types, params, NULL);
+	if (res != TEE_SUCCESS) {
+		EMSG("Failed to translate device address %#"PRIx32, da);
+		return 0;
```
etienne-lms, line 572: maybe add an inline comment here: `/* By convention in core, physical address 0 is invalid */`.

lib/libutee/include/remoteproc_pta.h
```diff
@@ -0,0 +1,150 @@
...
+
+	return s;
+}
+
+/*
+ * Platform capabilities.
+ *
+ * Get Platform firmware loader service capabilities.
+ *
+ * [in]  params[0].value.a:	Unique 32bit firmware identifier
```
etienne-lms, line 54: Actually, I wonder if `Unique 32bit remote processor identifier` is better.
The TA/PTA don't really care which system is implemented in the binary, rather where it's expected to be loaded, no?

ta/remoteproc/remoteproc_core.c
etienne-lms: ok

etienne-lms (commented):
"ta: add the support of the resource table in remoteproc TA".

I've added few review comments for the 2 remaining commits.

ta/remoteproc/remoteproc_core.c
```diff
@@ -94,14 +94,23 @@ enum remoteproc_img_type {
...
-#define LE32_TO_CPU(x) ({ uint8_t *_x = x; ((_x)[3] << 24) | ((_x)[2] << 16) | \
-					   ((_x)[1] << 8) | (_x)[0]; })
-
-#define U64_ADD_TO_ALIGN(s) ({ size_t _s = s; (_s) + sizeof(uint64_t) - \
-					     ((_s) % sizeof(uint64_t)); })
-#define U64_ALIGN_SZ(x) ({ size_t _x = x; (_x) % (sizeof(uint64_t)) ? \
-					  U64_ADD_TO_ALIGN(_x) : (_x); })
+#define LE16_TO_CPU(x) ({		\
+		uint8_t *__x = x;	\
+		(__x[1] << 8) | __x[0]; \
```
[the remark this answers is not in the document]
jenswi-linaro: It's confusing with yet another way of doing this. How about exporting `core/include/io.h` to `libutils` and using `get_unaligned_le*()` instead?

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1102 @@
...
+	})
+#define LE32_TO_CPU(x) ({					         \
+		uint8_t *__x = x;					 \
+		(__x[3] << 24) | (__x[2] << 16) | (__x[1] << 8) | __x[0]; \
+	})
+
+#define U64_ADD_TO_ALIGN(s) ({					   \
```
jenswi-linaro, line 106: I'm not sure what these macros are supposed to do. How about adding a comment and implementing them using `ROUNDUP()` and friends instead?

ta/remoteproc/sub.mk
```diff
@@ -0,0 +1,6 @@
+# REMOTEPROC TA heap size can be customized if 4kB is not enough
+CFG_REMOTEPROC_TA_HEAP_SIZE ?= (4 * 1024)
```
jenswi-linaro, line 2: No, CFG variables should not change once we start to process the `sub.mk` files.

ta/remoteproc/remoteproc_core.c
arnopo: done

ta/remoteproc/sub.mk
arnopo: This is inspired from the PKCS11 TA : [`ta/pkcs11/sub.mk`](https://github.com/OP-TEE/optee_os/blob/master/ta/pkcs11/sub.mk#L7)

Please, could you provide some advice on how to accomplish this in a clean manner?

jenswi-linaro: Ouch, I didn't see that one before. I'd prefer to put this in `mk/config.mk`.

ta/remoteproc/remoteproc_core.c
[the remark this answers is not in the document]
jenswi-linaro: > A discussion was initiate with @etienne-lms on this macro. [#6027 (comment)](https://github.com/OP-TEE/optee_os/pull/6027#discussion_r1203612792) the point was that macro should support conversion from LE to LE or BE.

Except that it does a little bit more so it's doing the exact thing (minus eventual bugs) as get_unaligned_le16().

> 
> The question from Etienne was: @jforissier, @jenswi-linaro, may I ask your advice on this? Do we need such macro? Can we assume OP-TEE OS and TAs are always little endian?

No, we can't. However, we can depend on the TA devkit to be updated to cover this. So The macros TEE_U*_{FROM,TO}_BIG_ENDIAN() should be safe in that regard.

ta/remoteproc/sub.mk
jforissier: I also prefer not setting configuration variables in the `sub.mk` files and use them only to list the files to be compiled.

On the other hand it is convenient to keep the TA config under the TA source directory (i.e., `ta/remoteproc`). How about doing that in `ta/remoteproc/user_ta.mk`? It is included here: https://github.com/OP-TEE/optee_os/blob/3.22.0/ta/mk/build-user-ta.mk#L25. We could also introduce `ta/remoteproc/config.mk` and include it before `user_ta.mk`.

jenswi-linaro: It's still somewhat late in the middle of sub.mk loading for TAs. I don't suppose we need to make a big problem out of this. So if you prefer this fine with me for now.

ta/remoteproc/remoteproc_core.c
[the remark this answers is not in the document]
jenswi-linaro: Why do you need to make unaligned accesses?

Yes, the endian conversion functions will be updated if needed.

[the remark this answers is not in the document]
arnopo: > Why do you need to make unaligned accesses?

Not necessary, I just need to ensure a conversion from the LE to the CPU endian which can be LE or BE.

An alternative is to create a TEE_U*_FROM_LITTLE_ENDIAN in utee_define.h

```c
/* If we we're on a big endian platform we'll have to update these */
#define TEE_U64_FROM_BIG_ENDIAN(x)  TEE_U64_BSWAP(x)
#define TEE_U32_FROM_BIG_ENDIAN(x)  TEE_U32_BSWAP(x)
#define TEE_U16_FROM_BIG_ENDIAN(x)  TEE_U16_BSWAP(x)
#define TEE_U64_TO_BIG_ENDIAN(x)    TEE_U64_BSWAP(x)
#define TEE_U32_TO_BIG_ENDIAN(x)    TEE_U32_BSWAP(x)
#define TEE_U16_TO_BIG_ENDIAN(x)    TEE_U16_BSWAP(x)
+ #define TEE_U64_FROM_LITTLE_ENDIAN(x)  x
+ #define TEE_U32_FROM_LITTLE_ENDIAN(x)  x
+ #define TEE_U16_FROM_LITTLE_ENDIAN(x)  x
+ #define TEE_U64_TO_LITTLE_ENDIAN(x)  x
+ #define TEE_U32_TO_LITTLE_ENDIAN(x)  x
+ #define TEE_U16_TO_LITTLE_ENDIAN(x)  x
```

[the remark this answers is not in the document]
jenswi-linaro: Sounds good.

etienne-lms (commented):
some minor comments

ta/remoteproc/remoteproc_core.c
arnopo: done

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1093 @@
...
+	if (!ctx)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	if (ctx->state != REMOTEPROC_OFF)
+		return TEE_ERROR_BAD_STATE;
+
+	if (!params[1].memref.buffer || !params[1].memref.size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	DMSG("Got base addr: %p size %zx", params[1].memref.buffer,
```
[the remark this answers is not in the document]
arnopo: done

ta/remoteproc/remoteproc_core.c
```diff
@@ -239,8 +231,11 @@ static TEE_Result remoteproc_get_tlv(void *tlv_chunk, size_t tlv_size,
 
 	/* Parse the TLV area */
 	while (p_tlv < p_end_tlv) {
-		tlv_type = LE32_TO_CPU(p_tlv);
-		tlv_length = LE32_TO_CPU(&p_tlv[RPROC_TLV_LENGTH_OF]);
+		p_tlv_d = (uint32_t *)(void *)p_tlv;
```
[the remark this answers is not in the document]
etienne-lms: I meant here:
```c
	while (p_tlv < p_end_tlv) {
		uint32_t d32 = 0;

		memcpy(&d32, p_tlv, sizeof(d32));
		tlv_type = TEE_U32_FROM_LITTLE_ENDIAN(d32);

		memcpy(&d32, p_tlv + RPROC_TLV_LENGTH_OF, sizeof(d32));
		tlv_length = TEE_U32_FROM_LITTLE_ENDIAN(d32);

		(...)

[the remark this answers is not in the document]
Comment (arnopo) [reply ta/remoteproc/remoteproc_core.c:239]: yes much better than my implementation, thanks for the proposal.

File: ta/remoteproc/remoteproc_core.c [hunk 17]
```diff
@@ -557,7 +559,12 @@ static paddr_t remoteproc_da_to_pa(uint32_t da, size_t size,
 		return 0;
 	}
 
-	return (paddr_t)reg_pair_to_64(params[3].value.b, params[3].value.a);
+	pa = (paddr_t)reg_pair_to_64(params[3].value.b, params[3].value.a);
+
+	/* On success, assert that the pa address is not 0 */
+	assert(res != TEE_SUCCESS || pa != 0);
```
etienne-lms, line 565: You can remove the test on `res`. It's always `TEE_SUCCESS` when we're here.

ta/remoteproc/remoteproc_core.c
```diff
@@ -589,6 +582,7 @@ static TEE_Result get_tlv_images_size(struct remoteproc_context *ctx,
 {
 	TEE_Result res = TEE_ERROR_GENERIC;
 	uint8_t *tlv_value = NULL;
+	uint32_t *tlv_v = NULL;
```
etienne-lms, line 652:
```patch
-	uint32_t *tlv_v = NULL;
+	uint32_t tlv_v = 0;

File: ta/remoteproc/remoteproc_core.c [hunk 19]
```diff
@@ -0,0 +1,1096 @@
...
+	uint32_t version;
+	uint32_t tlv_len;
+	uint32_t sign_len;
+	uint32_t img_len;
+};
+
+#define FW_TLV_PTR(img, hdr)  ((img) + sizeof(*(hdr)))
+#define FW_SIGN_PTR(img, hdr) ({					 \
+		struct remoteproc_fw_hdr *__hdr = hdr;			 \
+		FW_TLV_PTR((img), __hdr) + ROUNDUP_64(__hdr->tlv_len); \
```
etienne-lms, line 141: sorry nitpicking. As per coding style, there needs an empty line above.
```c
#define FW_SIGN_PTR(img, hdr) ({ \
		struct remoteproc_fw_hdr *__hdr = hdr; \
		\
		FW_TLV_PTR((img), __hdr) + ROUNDUP_64(__hdr->tlv_len); \
	})
#define FW_IMG_PTR(img, hdr) ({ \
		struct remoteproc_fw_hdr *___hdr = hdr; \
		\
		FW_SIGN_PTR((img), (___hdr)) + ROUNDUP_64(___hdr->sign_len); \
	})
```

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1096 @@
...
+	})
+#define FW_IMG_PTR(img, hdr) ({						      \
+		struct remoteproc_fw_hdr *___hdr = hdr;			      \
+		FW_SIGN_PTR((img), (___hdr)) +	ROUNDUP_64(___hdr->sign_len); \
```
etienne-lms, line 145: typo: replace the tab with a space char.

ta/remoteproc/remoteproc_core.c
arnopo: done

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1098 @@
...
+}
+
+static TEE_Result
+remoteproc_save_fw_header_and_tlvs(struct remoteproc_context *ctx,
+				   void *fw_orig, uint32_t fw_orig_size)
+{
+	struct remoteproc_fw_hdr *hdr = fw_orig;
+	uint32_t length = 0;
+
+	if (ADD_OVERFLOW(sizeof(*hdr), ROUNDUP_64(hdr->tlv_len), &length) ||
```
etienne-lms, line 357: Thi should check `hdr` alignment `!IS_ALIGNED_WIT(fw_orig, struct remoteproc_fw_hdr))`.

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1098 @@
...
+	ctx->fw_img_sz = hdr->img_len;
+	ctx->fw_img = FW_IMG_PTR(fw_orig, hdr);
+
+	DMSG("Firmware images addr: %p size: %zu", ctx->fw_img,
+	     ctx->fw_img_sz);
+
+	return TEE_SUCCESS;
+
+free_sec_cpy:
+	TEE_Free(ctx->sec_cpy);
```
etienne-lms, line 530: I see that `ctx->sec_cpy` is used only in the scope of this function `remoteproc_verify_firmware()`.
I think the field could be removed from `struct remoteproc_context` and be referenced from `remoteproc_verify_firmware()` stack, passed as argument to `remoteproc_save_fw_header_and_tlvs()`, `remoteproc_verify_header()` and `remoteproc_verify_signature()`.

arnopo: The secure copy contains help hash table and TLV. I implemented in this way to be able to reuse it later out of this function.
As today I have no code that reuses it ( ST downstream) I can change this as you propose.
we can come back to a context pointer later if needed.

arnopo:
I rebased on master and squashed fixup before taking into account new comments from @etienne-lms

ta/remoteproc/remoteproc_core.c
```diff
@@ -135,13 +135,15 @@ struct remoteproc_fw_hdr {
 };
 
 #define FW_TLV_PTR(img, hdr)  ((img) + sizeof(*(hdr)))
-#define FW_SIGN_PTR(img, hdr) ({					 \
-		struct remoteproc_fw_hdr *__hdr = hdr;			 \
-		FW_TLV_PTR((img), __hdr) + ROUNDUP_64(__hdr->tlv_len); \
+#define FW_SIGN_PTR(img, hdr) ({					\
+		struct remoteproc_fw_hdr *__hdr = hdr;			\
```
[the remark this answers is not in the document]
arnopo: done

ta/remoteproc/remoteproc_core.c
arnopo: So here if I well understood, I keep the pointer but set it to NULL before free it.
I'm not sureto  understand your point concerning `assert(ctx->sec_cpy)` as function that use the pointer are called by `remoteproc_load_fw()`.
It seems  to me some over checking. But perhaps you prefer that add it anyway?

etienne-lms: At least, in your latest proposal, `ctx->sec_cpy` is cleared when the buffer is freed. So fine with me.

etienne-lms (commented):
Looks good to me.

`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits:
"pta: stm32mp: add new remoteproc pta API",
"libutee: add TEE_U*_FROM/TO_LITTLE_ENDIAN()" and
"ta: add the support of the start and stop command in remoteproc TA".

"ta: add remote processor firmware ELF parser",
"ta: add remote processor Trusted Application" and
"ta: add the support of the resource table in remoteproc TA".

arnopo:
> Looks good to me.
> 
> `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits: "pta: stm32mp: add new remoteproc pta API", "libutee: add TEE_U*_FROM/TO_LITTLE_ENDIAN()" and "ta: add the support of the start and stop command in remoteproc TA".
> 
> `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits "ta: add remote processor firmware ELF parser", "ta: add remote processor Trusted Application" and "ta: add the support of the resource table in remoteproc TA".

Done, thanks a lot for your reviews!

jforissier (commented):
Hi @arnopo,

Please see comments below. More comments...

- For "pta: stm32mp: add new remoteproc pta API", I suggest "libutee: add remoteproc PTA API"
Why did you use a stm32mp: prefix? Isn't this API generic? It looks like other platforms could re-use this API, only the implementation will be platform-specific.. If not, the file name should probably contain the platform (remoteproc_stm32mp_pta.h).
- For "libutee: add TEE_U*_FROM/TO_LITTLE_ENDIAN()"
- For the following commits I also suggest changing subjects:
  "ta: add remote processor firmware ELF parser" => "ta: remoteproc: add firmware ELF parser"
  "ta: add remote processor Trusted Application " => " ta: remoteproc: add remote processor Trusted Application"
  "ta: add the support of the start and stop command in remoteproc TA" => " ta: remoteproc: add support for start and stop commands"
  "ta: add the support of the resource table in remoteproc TA" => "ta: remoteproc: add support for resource table"
  Then: 

ta/remoteproc/elf_parser.c
```diff
@@ -0,0 +1,174 @@
...
+
+	return vaddr >= fw && vaddr < fw + fw_size;
+}
+
+TEE_Result e32_parse_ehdr(uint8_t *fw, size_t size)
+{
+	Elf32_Ehdr *ehdr = (Elf32_Ehdr *)(void *)fw;
+
+	if (!fw || !IS_ALIGNED_WITH_TYPE(fw, uint32_t)) {
+		EMSG("Invalid fw address %p", fw);
```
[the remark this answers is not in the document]
etienne-lms: I could not fine definitive specification for `PRIxVA`.
I would say `PRIxVA` is rather related to `vaddr_t` (and `PRIxPTR` for `uintptr_t` despite those 2 PRIxXXX are the same).
Recent toolchain seems to accept `%p` for any pointer (`sometype *`) . I remember old toolchains warned on non-`void *` type references.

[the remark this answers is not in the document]
jforissier: > I could not fine definitive specification for `PRIxVA`. I would say `PRIxVA` is rather related to `vaddr_t`

Right.

> (and `PRIxPTR` for `uintptr_t` despite those 2 PRIxXXX are the same). Recent toolchain seems to accept `%p` for any pointer (`sometype *`) . I remember old toolchains warned on non-`void *` type references.

Then perhaps we should always use `"0x" PRIxPTR` to print pointers?

Also, this is an EMSG(), not a debug message, so I think we should not print an address in clear as it could help an attacker defeat ASLR. Perhaps we should consider doing like in the Linux kernel and have a CFG to hide addresses by default?

arnopo (commented):
Hi @jforissier 

> Hi @arnopo,
> 
> Please see comments below. More comments...
> 
> * For "pta: stm32mp: add new remoteproc pta API", I suggest "libutee: add remoteproc PTA API"
>   Why did you use a stm32mp: prefix? Isn't this API generic? It looks like other platforms could re-use this API, only the implementation will be platform-specific.. If not, the file name should probably contain the platform (remoteproc_stm32mp_pta.h).

Yes  it is generic, The implementation will be platform dependent, but not the API


> * For "libutee: add TEE_U*_FROM/TO_LITTLE_ENDIAN()"
>   `Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`
> * For the following commits I also suggest changing subjects:
>   "ta: add remote processor firmware ELF parser" => "ta: remoteproc: add firmware ELF parser"
>   "ta: add remote processor Trusted Application " => " ta: remoteproc: add remote processor Trusted Application"
>   "ta: add the support of the start and stop command in remoteproc TA" => " ta: remoteproc: add support for start and stop commands"
>   "ta: add the support of the resource table in remoteproc TA" => "ta: remoteproc: add support for resource table"
>   Then:
>   `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`

Thanks for the advice, I updated the commit subjects and add your acked-by on them ( I assumed that the acked-by was for all the mentioned commits).

arnopo:
@jforissier 
I updated patches instead of adding fixes on top of to take your comments into account. The reason is that I changed the commit subjects and created a ta/remoteproc/src subfolder. Sorry for any inconvenient for the review process

ta/remoteproc/remoteproc_core.c
```diff
@@ -0,0 +1,1103 @@
...
+					      TEE_PARAM_TYPE_VALUE_INPUT,
+					      TEE_PARAM_TYPE_VALUE_INPUT,
+					      TEE_PARAM_TYPE_VALUE_INPUT);
+		params[1].value.a = da + size;
+		params[2].value.a = mem_size - size;
+		params[3].value.a = 0;
+
+		res = TEE_InvokeTACommand(pta_session, TEE_TIMEOUT_INFINITE,
+					  PTA_RPROC_SET_MEMORY,
+					  param_types, params, NULL);
```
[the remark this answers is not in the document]
jforissier: You're right, my mistake.

jforissier:
- "ta: remoteproc: add firmware ELF parser":

ta/remoteproc/elf_parser.c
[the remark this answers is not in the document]
etienne-lms: > Perhaps we should consider doing like in the Linux kernel and have a CFG to hide addresses by default?

Sounds a good idea. I guess this would be out of the scope of this P-R.
In the meantime, @arnopo, maybe the simplest thing here is to not print the virtual address, if you agree.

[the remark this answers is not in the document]
arnopo: I change the print to `EMSG("Invalid firmware address");`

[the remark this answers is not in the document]
jenswi-linaro: > Then perhaps we should always use "0x" PRIxPTR to print pointers?

I'd rather keep it apart. `PRIxPA` for `paddr_t`, `PRIxVA` for `vaddr_t`, `PRIxPTR` for `uintptr_t`, and `%p` for `void *`.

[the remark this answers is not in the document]
jforissier: > > Then perhaps we should always use "0x" PRIxPTR to print pointers?
> 
> I'd rather keep it apart. `PRIxPA` for `paddr_t`, `PRIxVA` for `vaddr_t`, `PRIxPTR` for `uintptr_t`, and `%p` for `void *`.

Thanks for clarifying. Indeed the `I` in `PRI` stands for integer so it totally makes sense.

jforissier:
@arnopo do you plan on introducing the platform-specific remoteproc PTA soon? I am asking because technically this PR is "reviewed enough", but I'd rather merge it at the same time when the PTA is ready. What do you think?

arnopo:
> @arnopo do you plan on introducing the platform-specific remoteproc PTA soon? I am asking because technically this PR is "reviewed enough", but I'd rather merge it at the same time when the PTA is ready. What do you think?

That's seems reasonable, I will prepare a new PR with the PTA and the associated driver ( probably end of next week)
One question for this PR: Should I add your `acked-by` or `reviewed-by` on https://github.com/OP-TEE/optee_os/pull/6300/commits/787a832cc6a74ecaa832a9baf83b40a645f5d8e9? or are you waiting to review the PR with the PTA first?

jforissier:
> > @arnopo do you plan on introducing the platform-specific remoteproc PTA soon? I am asking because technically this PR is "reviewed enough", but I'd rather merge it at the same time when the PTA is ready. What do you think?
> 
> That's seems reasonable, I will prepare a new PR with the PTA and the associated driver ( probably end of next week) 

Good :+1: 

> One question for this PR: Should I add your `acked-by` or `reviewed-by` on [787a832](https://github.com/OP-TEE/optee_os/commit/787a832cc6a74ecaa832a9baf83b40a645f5d8e9)? or are you waiting to review the PR with the PTA first?

You may add `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>` to that commit already.

arnopo:
> > > @arnopo do you plan on introducing the platform-specific remoteproc PTA soon? I am asking because technically this PR is "reviewed enough", but I'd rather merge it at the same time when the PTA is ready. What do you think?
> > 
> > 
> > That's seems reasonable, I will prepare a new PR with the PTA and the associated driver ( probably end of next week)
> 
> Good 👍
> 
> > One question for this PR: Should I add your `acked-by` or `reviewed-by` on [787a832](https://github.com/OP-TEE/optee_os/commit/787a832cc6a74ecaa832a9baf83b40a645f5d8e9)? or are you waiting to review the PR with the PTA first?
> 
> You may add `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>` to that commit already.

Done, 
Thanks!

arnopo:
Update to suppress unexpected Change-Id added in some commit messages.

arnopo:
thi

> This pull request has been marked as a stale pull request because it has been open (more than) 30 days with no activity. Remove the stale label or add a comment, otherwise this pull request will automatically be closed in 5 days. Note, that you can always re-open a closed issue at any time.

This PR is waiting https://github.com/OP-TEE/optee_os/pull/6470 approval.

etienne-lms (approved):
The series looks ok to me to be merged.
P-R https://github.com/OP-TEE/optee_os/pull/6470 brings a PTA implementation based on the API introduced here.

@arnopo, maybe rebase to get an CI IBART test result.

arnopo:
I rebased on master done and added a fixup https://github.com/OP-TEE/optee_os/pull/6300/commits/c1d24e75998c59ce092cada37be3e60edb1bd040 for review.

arnopo:
> I rebased on master done and added a fixup [c1d24e7](https://github.com/OP-TEE/optee_os/commit/c1d24e75998c59ce092cada37be3e60edb1bd040) for review.

My apologies, I built with an old version of OP-TEE. The fixup is invalid since `TEE_DigestDoFinal` has been updated in  185bf58cd6 ("GP131: Update TEE_{Digest,Cipher,MAC}* functions")
So no update but only a rebase on top of the main branch.

jforissier:
@jenswi-linaro any more comments on this? Thanks.

arnopo:
Rebase on top of #6470 merge, 
I removed commit dab8b5c58d ("libutee: add new remoteproc PTA API") from this PR as it as already been merged in #6470

Issue #6304, opened 2023-09-20
Explore a way to examine TEE secure time


weizhaojiang:
We are working on video streaming product which includes DRM features. In DRM TA, we depends on TEE secure time to identify the license is expired or not. Therefore, the accuracy of secure time is very critical to us. Meanwhile, we are working on different platforms which have different secure time implementation. The secure time might go fast or slow, which cause problems.

Now, we are looking for a way to examine secure time in a run-time. 

One of my thinking is add one cmd in ``stats PTA``: 
```
#define STATS_CMD_TA_SECURE_TIME
```
We pass two parameters: 1) REE system time and 2) delta allowed, when invoking that cmd. The cmd then return a ``bool`` value to tell us if the TEE time is sync with REE time.

This way is using REE system time as base to examine TEE secure time to see if it goes fast or slow.

Please help to review above and let me know if it breaks something in security. If you have other proposal, please kindly share to me. Thanks.

weizhaojiang:
Correct, it's for testing purpose. If test result shows the secure time is not accurate enough, we will go to refine its implementation.
Will have PR on this change. Thanks.

weizhaojiang:
@jenswi-linaro May I have question?
I looking into TEE secure time and call below two functions in PTA to get ``ree_time`` and ``tee_time``.
```
	tee_time_get_ree_time(&ree_time);
	tee_time_get_sys_time(&tee_time);
```
When run on Qume, I found ``ree_time`` is an epoch unix timestamp while ``tee_time`` is timestamp from 0 when system boot up. I believe different platform will have different tee_time source. So, looks like I can't use ``ree_time`` absolute value to examine ``tee_time`` as they have different start point.
To check if a tee time is good or not, I need to make sure
1. the time is monotonically increasing, not rollback across power cycle;
2. The TEE time increment should be consistent with the REE time.

To simplify the testcase design, can I have the CDM ``STATS_CMD_TA_SECURE_TIME`` return the TEE time to REE world? Any security concern here?

jenswi-linaro:
The platform probably needs to maintain a secure time offset using a secure RTC or such.
Another stats cmd shouldn't be a problem.

weizhaojiang:
The patches have been merged, closing the ticket.

PR #6311, opened 2023-09-21
core: implement a method to get secure time

weizhaojiang:
Add new command (STATS_CMD_SECTIME_STATS) in the stats PTA. The command returns the tee and ree time for debug purposes. https://github.com/OP-TEE/optee_os/issues/6304

### DIFF
--- a/core/pta/stats.c
+++ b/core/pta/stats.c
@@ -3,14 +3,15 @@
  * Copyright (c) 2015, Linaro Limited
  */
 #include <compiler.h>
-#include <stdio.h>
-#include <trace.h>
 #include <kernel/pseudo_ta.h>
-#include <mm/tee_pager.h>
+#include <kernel/tee_time.h>
+#include <malloc.h>
 #include <mm/tee_mm.h>
+#include <mm/tee_pager.h>
+#include <stdio.h>
 #include <string.h>
 #include <string_ext.h>
-#include <malloc.h>
+#include <trace.h>
 
 #define TA_NAME		"stats.ta"
 
@@ -37,6 +38,16 @@
  * uint32_t    Biggest byte size which allocation succeeded
  */
 #define STATS_CMD_TA_STATS		3
+
+/*
+ * STATS_CMD_GET_TIME - Get both REE time and TEE time
+ *
+ * [out]    value[0].a        REE time as seen by OP-TEE in seconds
+ * [out]    value[0].b        REE time as seen by OP-TEE, milliseconds part
+ * [out]    value[1].a        TEE system time in seconds
+ * [out]    value[1].b        TEE system time, milliseconds part
+ */
+#define STATS_CMD_GET_TIME		4
 
 #define STATS_NB_POOLS			4
 
@@ -175,6 +186,29 @@ static TEE_Result get_user_ta_stats(uint32_t type,
 	return res;
 }
 
+static TEE_Result get_system_time(uint32_t type,
+				  TEE_Param p[TEE_NUM_PARAMS])
+{
+	TEE_Time ree_time = { };
+	TEE_Time tee_time = { };
+
+	if (TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_OUTPUT,
+			    TEE_PARAM_TYPE_VALUE_OUTPUT,
+			    TEE_PARAM_TYPE_NONE,
+			    TEE_PARAM_TYPE_NONE) != type)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	tee_time_get_sys_time(&tee_time);
+	tee_time_get_ree_time(&ree_time);
+
+	p[0].value.a = ree_time.seconds;
+	p[0].value.b = ree_time.millis;
+	p[1].value.a = tee_time.seconds;
+	p[1].value.b = tee_time.millis;
+
+	return TEE_SUCCESS;
+}
+
 /*
  * Trusted Application Entry Points
  */
@@ -192,6 +226,8 @@ static TEE_Result invoke_command(void *psess __unused,
 		return get_memleak_stats(ptypes, params);
 	case STATS_CMD_TA_STATS:
 		return get_user_ta_stats(ptypes, params);
+	case STATS_CMD_GET_TIME:
+		return get_system_time(ptypes, params);
 	default:
 		break;
 	}


core/pta/stats.c
```diff
@@ -175,6 +177,29 @@ static TEE_Result get_user_ta_stats(uint32_t type,
 	return res;
 }
 
+static TEE_Result get_securetime_stats(uint32_t type,
+				       TEE_Param p[TEE_NUM_PARAMS])
+{
+	TEE_Time tee_time;
+	TEE_Time ree_time;
```
etienne-lms, line 184: `= { }` for both.

core/pta/stats.c
```diff
@@ -37,6 +38,7 @@
  * uint32_t    Biggest byte size which allocation succeeded
  */
 #define STATS_CMD_TA_STATS		3
+#define STATS_CMD_SECTIME_STATS		4
```
etienne-lms, line 41: I suggest to rename the command to `STATS_CMD_SYSTEM_TIME`
Please add an inline description of the ABI, e.g.:
```c
/*
 * STATS_CMD_SYSTEM_TIME - Get system times
 * 
 * [out]    value[0].a        REE time as seen by OP-TEE in seconds
 * [out]    value[0].b        REE time as seen by OP-TEE, milliseconds part
 * [out]    value[1].a        TEE system time in seconds
 * [out]    value[1].b        TEE system time, milliseconds part
 */
#define STATS_CMD_SECTIME_STATS		4
```

core/pta/stats.c
```diff
@@ -11,6 +11,7 @@
 #include <string.h>
 #include <string_ext.h>
 #include <malloc.h>
+#include <kernel/tee_time.h>
```
etienne-lms, line 14: Shoudl follow the alphabetical order.
Bt the way, could you also the other `#include` directives?
```c
#include <compiler.h>
#include <kernel/pseudo_ta.h>
#include <kernel/tee_time.h>
#include <malloc.h>
#include <mm/tee_mm.h>
#include <mm/tee_pager.h>
#include <stdio.h>
#include <string.h>
#include <string_ext.h>
#include <trace.h>
```

etienne-lms:
Would you mind creating a P-R in optee_test also to add an `xtest --stats --time` command?

core/pta/stats.c
weizhaojiang: good points, thanks.

core/pta/stats.c
weizhaojiang: will do.

jforissier:
Why not use a normal TA for this (`TEE_GetSystemTime()` and `TEE_GetREETime()`)?

weizhaojiang:
> Why not use a normal TA for this (`TEE_GetSystemTime()` and `TEE_GetREETime()`)?

Thanks for your comment. We would like to add this cmd in ``stats`` PTA rather than a user TA because:
1. The cmd is to query secure time which is implemented in OS kernel. ``TEE_GetSystemTime()`` and ``TEE_GetREETime()`` essentially call into ``tee_time_get_sys_time()`` and ``tee_time_get_ree_time``.
2. We hope to test TEE secure without loading a user TA. This can simplifying our test procedure design.
3. If we add the cmd into a specific user TA, that means only that TA's client can get TEE time for testing. In our product, the TA is on-demand built into system, so it might not be available in that product. The ``stats`` PTA looks a common place to add and it will make the cmd available for all products.

core/pta/stats.c
```diff
@@ -175,6 +185,29 @@ static TEE_Result get_user_ta_stats(uint32_t type,
 	return res;
 }
 
+static TEE_Result get_system_time(uint32_t type,
+				  TEE_Param p[TEE_NUM_PARAMS])
+{
+	TEE_Time ree_time = {0, 0};
+	TEE_Time tee_time = {0, 0};
```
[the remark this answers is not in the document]
weizhaojiang: sure, will do. Thanks for suggestion.

core/pta/stats.c
```diff
@@ -37,6 +38,15 @@
  * uint32_t    Biggest byte size which allocation succeeded
  */
 #define STATS_CMD_TA_STATS		3
+/*
```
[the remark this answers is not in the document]
weizhaojiang: will do.

core/pta/stats.c
```diff
@@ -36,7 +37,17 @@
...
+ * STATS_CMD_SYSTEM_TIME - Get system times
+ *
+ * [out]    value[0].a        REE time as seen by OP-TEE in seconds
+ * [out]    value[0].b        REE time as seen by OP-TEE, milliseconds part
+ * [out]    value[1].a        TEE system time in seconds
+ * [out]    value[1].b        TEE system time, milliseconds part
+ */
+#define STATS_CMD_SYSTEM_TIME		4
```
jforissier, line 50: Since it's not only "system" time but also "REE" time, I suppose `STATS_CMD_GET_TIME` would be a better name

jforissier (commented):
> > Why not use a normal TA for this (`TEE_GetSystemTime()` and `TEE_GetREETime()`)?
> 
> Thanks for your comment. We would like to add this cmd in `stats` PTA rather than a user TA because:
> 
>   1. The cmd is to query secure time which is implemented in OS kernel. `TEE_GetSystemTime()` and `TEE_GetREETime()` essentially call into `tee_time_get_sys_time()` and `tee_time_get_ree_time`.
>   2. We hope to test TEE secure without loading a user TA. This can simplifying our test procedure design.
>   3. If we add the cmd into a specific user TA, that means only that TA's client can get TEE time for testing. In our product, the TA is on-demand built into system, so it might not be available in that product. The `stats` PTA looks a common place to add and it will make the cmd available for all products.

OK. In any case it is a very small change, so if that makes your life easier I'm good with it.

I have one minor comment below.

core/pta/stats.c
weizhaojiang: good point, thanks.

jforissier:

core/pta/stats.c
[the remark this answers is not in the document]
etienne-lms: please swap line 40 and 41:
```c
 * (...)
 * uint32_t    Number of failed allocation requests
 * uint32_t    Biggest byte size which allocation failed
 * uint32_t    Biggest byte size which allocation succeeded
 */
#define STATS_CMD_TA_STATS		3

/*
 * STATS_CMD_GET_TIME - Get both REE time and TEE time
 *
 * [out]    value[0].a        REE time as seen by OP-TEE in seconds
 * [out]    value[0].b        REE time as seen by OP-TEE, milliseconds part
 * (...)
```

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with remaining comment addressed.

jforissier:
`core: pta: stats: implement...` would be a better prefix

weizhaojiang:
> `core: pta: stats: implement...` would be a better prefix

sounds good and done.

PR #6312, opened 2023-09-22
Core: support fault mitigations in non-threaded code

sichunqin:
Fix the issue of fault mitigation that won't work in interrupt handler function.
assertion 'ct >= 0 && ct < CFG_NUM_THREADS' failed at core/arch/arm/kernel/thread.c:799 <thread_get_id>

The problem is in  __ftmn_get_tsd_func_arg_pp whhich call thread_get_tsd which thread_get_id. The reason is that interrupt handler is not associated with any thread, so the ct (current_thread_id) value is -1 that would cause assert problem.

The fix is to add ftmn_arg to thread_core_local and the new variable would be used when current thread is < 0.

### DIFF
--- a/core/arch/arm/include/kernel/thread_arch.h
+++ b/core/arch/arm/include/kernel/thread_arch.h
@@ -64,6 +64,9 @@ struct thread_core_local {
 #ifdef CFG_CORE_DEBUG_CHECK_STACKS
 	bool stackcheck_recursion;
 #endif
+#ifdef CFG_FAULT_MITIGATION
+	struct ftmn_func_arg *ftmn_arg;
+#endif
 } THREAD_CORE_LOCAL_ALIGNED;
 
 struct thread_vector_table {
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
Please update the commit subject to something like:
`core: support fault mitigations in non-threaded code`
The commit is also missing a `Signed-off-by:` tag.

sichunqin:
Update commit message as suggested. @jenswi-linaro, please take a look. Thanks.

jenswi-linaro:
Please fixup the commit message as:
```
core: support fault mitigations in non-threaded code

Fault mitigation won't work in non-threaded code due to the following 
error:
assertion 'ct >= 0 && ct < CFG_NUM_THREADS' failed at core/arch/arm/kernel/threa
d.c:799 <thread_get_id>

The problem is in  __ftmn_get_tsd_func_arg_pp which calls thread_get_tsd
which thread_get_id. The reason is that the interrupt handler is not 
associated with any thread, so the ct (current_thread_id) value is -1 which
would cause an assert problem.

The fix is to add ftmn_arg to thread_core_local and the new variable would
be used when the current thread is < 0.

Signed-off-by: Sichun Qin <sichun.qin@amlogic.com>
```
Note the lowercase leading letter in "core", an empty line between the commit subject and the rest of the message. Only the cited assertion line is allowed to wrap.

sichunqin:
Updated as suggested. Thanks.

jenswi-linaro (commented):

PR #6317, opened 2023-09-25
[v4] Rem core mbedtls

jenswi-linaro:
This is a part of https://github.com/jenswi-linaro/optee_os/pull/11.

As a first step, we remove support for mbedtls in OP-TEE core, except we still use the bignum implementation with LTC.

Another pull request will follow to update mbedtls to version 3.4.0.


etienne-lms (commented):

larper2axis:
This is unfortunate. We run with mbedtls in the core for its superior RSA performance.

jenswi-linaro:
The RSA implementation in MBedTLS 3.x is unfortunately incompatible with GlobalPlatform TEE Internal Core API, see https://github.com/jenswi-linaro/optee_os/pull/9.

Speaking of performance, I've been toying with the idea of using OpenSSL now that it has a permissive license. But that's not in scope here and will likely not happen either unless someone starts to push for it.

jenswi-linaro:
Tag applied.

larper2axis:
Improved blinding is very much appreciated and necessary to counter the ever improving techniques for extracting keys over side channels.  The GP API is the way it is, but if the user provides a key with to few components, we can always hit them with the runtine penalty of calling mbedtls_rsa_complete(). All in the name of improved security.

jforissier:
> In the meantime, please don't merge this PR.

I intend to keep the [v4] PRs open for a bit more time than usual if that's OK with you. I think we need to leave enough time for people to give feedback.

jenswi-linaro:
> I intend to keep the [v4] PRs open for a bit more time than usual if that's OK with you. I think we need to leave enough time for people to give feedback.

Makes sense, we only need to merge them before tagging the rc1.

jenswi-linaro:
@larper2axis Here's take 3 for updating to MbedTLS 3.4.0. https://github.com/jenswi-linaro/optee_os/pull/12

If it makes sense, please review the few commits without review tags.

I'll push the branch to `import/mbedtls-3.4.0` once we're happy with the pull request and issue another squashed pull request and the master branch.

PR #6327, opened 2023-09-28
ta: pkcs11: prevent user ID verification when user PIN is not set

etienne-lms:
Fixes User PIN verification in ACL mode (protected authentication) so that it always returns PKCS11_CKR_USER_PIN_NOT_INITIALIZED when User PIN has not been initialized yet by the Security Officer. Before this change, this was tested only in the standard PIN path, not for the authenticated TEE identity mode (CFG_PKCS11_TA_AUTH_TEE_IDENTITY=y).

### DIFF
--- a/ta/pkcs11/src/pkcs11_token.c
+++ b/ta/pkcs11/src/pkcs11_token.c
@@ -1192,12 +1192,14 @@ static enum pkcs11_rc check_user_pin(struct pkcs11_session *session,
 	struct token_persistent_main *db = token->db_main;
 	enum pkcs11_rc rc = PKCS11_CKR_OK;
 
+	if (!(db->flags & PKCS11_CKFT_USER_PIN_INITIALIZED))
+		return PKCS11_CKR_USER_PIN_NOT_INITIALIZED;
+
 	if (IS_ENABLED(CFG_PKCS11_TA_AUTH_TEE_IDENTITY) &&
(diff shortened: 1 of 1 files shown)


etienne-lms:
@vesajaaskelainen, @varder, maybe interested by this change?

etienne-lms:
@jforissier, could you remove the Stale label? I think this change is of interest.

jforissier:
> @jforissier, could you remove the Stale label? I think this change is of interest.

Done.

etienne-lms:
thanks

vesajaaskelainen:
> @vesajaaskelainen, @varder, maybe interested by this change?

yes... let's have a look at it.

vesajaaskelainen:
With quick look seems good change. I would still like to execute my "manual test suite" for the change.

vesajaaskelainen:
Sorry it took a bit to get back testing this out.

You need to rebase the PR as there seems to have been some changes in the code in between.

Adapted the code to latest master and tried it out.

Set pin path had the check already there so that cannot be used for testing.

Login path did not have the check -- but there is a bit of problem here for testing it out ;).

One would need to posses NIL UUID which is a bit of a challenge and did not really want to go for that path ;).

But the code is good in sound and manual tests that I performed all passed so -- with the rebased and adaptations done:
```
Reviewed-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>
```

etienne-lms:
Thanks Vesa, I've rebased and apply your R-b tag.

PR #6301, opened 2023-09-19
[RFC] driver: add a voltage regulator framework

etienne-lms:
Proposal for a voltage regulator driver framework.

PR #6328, opened 2023-10-02
drivers: regulator: fix previous commit and implement always-on regulator

etienne-lms:
1 commit to fix a bug in the previously merged commit 1a3d3273040b (my apologies, bad squashing),
followed by the implementation of always-on regulators.


core/drivers/regulator/regulator.c
```diff
@@ -258,7 +252,7 @@ void regulator_print_state(const char *message __maybe_unused)
 	DMSG("name     use\ten\tuV\tmin\tmax\tsupply");
 
 	SLIST_FOREACH(regulator, &regulator_device_list, link)
-		DMSG("%8s %u\t%d\t%d\t%d\t%d\t%#x\t%s\n",
+		DMSG("%8s %u\t%d\t%d\t%d\t%d\t%s\n",
```
jforissier: This should have been caught by CI (mismatch in number of arguments). Why not?

etienne-lms: No board currently enable `CFG_DRIVERS_REGULATOR`. I did enable the switch only i the [last commit](https://github.com/OP-TEE/optee_os/pull/6301/commits/12455cde3bd8522a0e23b27f5d5270dd4fc9b58b#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54aR540-R544) of my series in https://github.com/OP-TEE/optee_os/pull/6301. 
I'll append a commit to enable the switch on plat-stm32mp1 so that we at least build a platform with the framework.

jforissier: Yes, it is a good idea to add some CI (at least building) whenever new code is added.

core/include/drivers/regulator.h
```diff
@@ -12,6 +12,18 @@
 #include <tee_api_types.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <util.h>
+
+/*
+ * Regulator property flags: related to device tree binding properties
+ *
+ * REGUL_ALWAYS_ON
```
jforissier, line 20: No need to repeat the flag name (here it's wrong ;))
```
/* Regulator property flags: related to device tree binding properties */

/* Regulator should never be disabled. DT property: regulator-always-on */
#define REGULATOR_ALWAYS_ON	BIT(0)
```

jforissier (commented):
One comment below.

core/include/drivers/regulator.h
etienne-lms: Ok, i'll fix. Thanks.

etienne-lms:
Review comment addressed and tag applied.

etienne-lms:
"CI / make check (QEMUv8/Xen)" failed with error message "No space left on device".

GseoC:
For the whole patchset:

etienne-lms:
Tags applied.

PR #6331, opened 2023-10-02
[WIP] Investigate CI failure on Xen job (no space left on device)

jforissier:
Draft pull request to try and fix the CI failures (Xen job mainly, but the Rust one is problematic too).


jforissier:
Fixed in https://github.com/OP-TEE/build/pull/689

PR #6333, opened 2023-10-03
plat-mediatek: add support for MT8188 SoC

ZheTing815:
Add OP-TEE support for the MT8188 SoC.


jforissier (commented):

I am not sure at all that two Signed-off-by are needed or even valid given that they represent the same person but with a different email. Would you mind choosing one? (it has to be the same than in the Author field, so if you don't change Author in the git settings, you should keep the google.com one)

Another thing: plat-mediatek has no more official maintainer in MAINTAINERS, would you like to add yourself?

ZheTing815:
> 
> I am not sure at all that two Signed-off-by are needed or even valid given that they represent the same person but with a different email. Would you mind choosing one? (it has to be the same than in the Author field, so if you don't change Author in the git settings, you should keep the google.com one)
Ok, I just modified the author.

> Another thing: plat-mediatek has no more official maintainer in MAINTAINERS, would you like to add yourself?
Hmm, I consider myself not a suitable person for it.

jforissier:
> > `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`
> > I am not sure at all that two Signed-off-by are needed or even valid given that they represent the same person but with a different email. Would you mind choosing one? (it has to be the same than in the Author field, so if you don't change Author in the git settings, you should keep the google.com one)
> 
> Ok, I just modified the author.

Thanks!

> 
> > Another thing: plat-mediatek has no more official maintainer in MAINTAINERS, would you like to add yourself?
> 
> Hmm, I consider myself not a suitable person for it.

That's fine.

PR #6334, opened 2023-10-03
core: add support mtk wait timeout API

ZheTing815:
We introudce a timout mechanismof condvar implementation for some secure driver.
Add the timeout parameter passing in wait_queue implementation for condvar_wait_timeout. Since some secure driver need timeout mechanism for exception handler.

### DIFF
--- a/core/include/kernel/mutex.h
+++ b/core/include/kernel/mutex.h
@@ -95,10 +95,15 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
 void condvar_wait_debug(struct condvar *cv, struct mutex *m,
 			const char *fname, int lineno);
 #define condvar_wait(cv, m) condvar_wait_debug((cv), (m), __FILE__, __LINE__)
+uint32_t condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m, uint32_t tmo,
+			const char *fname, int lineno);
+#define condvar_wait_timeout(cv, m, tmo) \
+			condvar_wait_timeout_debug((cv), (m), tmo, __FILE__, __LINE__)
 #else
 void condvar_signal(struct condvar *cv);
 void condvar_broadcast(struct condvar *cv);
 void condvar_wait(struct condvar *cv, struct mutex *m);
+uint32_t condvar_wait_timeout(struct condvar *cv, struct mutex *m, uint32_t tmo);
 #endif
 
 #endif /*KERNEL_MUTEX_H*/
--- a/core/include/kernel/notif.h
+++ b/core/include/kernel/notif.h
@@ -107,6 +107,11 @@ void notif_free_async_value(uint32_t value);
 TEE_Result notif_wait(uint32_t value);
 
 /*
+ * Wait timeout in normal world for a value to be sent by notif_send()
+ */
+TEE_Result notif_wait_timeout(uint32_t value, uint32_t tmo);
+
+/*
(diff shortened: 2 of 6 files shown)


core/kernel/mutex.c
```diff
@@ -434,23 +435,36 @@ static void __condvar_wait(struct condvar *cv, struct mutex *m,
...
+uint32_t condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m, uint32_t tmo,
+				const char *fname, int lineno)
+{
+	return __condvar_wait_timeout(cv, m, tmo, fname, lineno);
 }
 #else
 void condvar_wait(struct condvar *cv, struct mutex *m)
 {
-	__condvar_wait(cv, m, NULL, -1);
+	__condvar_wait_timeout(cv, m, 0, NULL, -1);
```
jforissier, line 463: `tmo == 0` means infinite timeout? That does not seem right.

core/include/kernel/mutex.h
```diff
@@ -95,10 +95,15 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
...
 #define condvar_wait(cv, m) condvar_wait_debug((cv), (m), __FILE__, __LINE__)
+uint32_t condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m, uint32_t tmo,
+			const char *fname, int lineno);
+#define condvar_wait_timeout(cv, m, tmo) \
+			condvar_wait_timeout_debug((cv), (m), tmo, __FILE__, __LINE__)
 #else
 void condvar_signal(struct condvar *cv);
 void condvar_broadcast(struct condvar *cv);
 void condvar_wait(struct condvar *cv, struct mutex *m);
+uint32_t condvar_wait_timeout(struct condvar *cv, struct mutex *m, uint32_t tmo);
```
jforissier, line 106: What unit is `tmo`? Can an infinite timeout be specified, thus making `condvar_wait_timeout()` behave like `condvar_wait()`?

core/include/kernel/notif.h
```diff
@@ -106,6 +106,11 @@ void notif_free_async_value(uint32_t value);
  */
 TEE_Result notif_wait(uint32_t value);
 
+/*
+ * Wait timeout in normal world for a value to be sent by notif_send()
+ */
+TEE_Result notif_wait_timeout(uint32_t value, uint32_t tmo);
```
jforissier, line 112: It may be better to split in two patches: extending condvars then modifying notifications.

core/kernel/notif.c
```diff
@@ -203,19 +203,30 @@ void notif_deliver_event(enum notif_event ev)
...
 
 TEE_Result notif_send_sync(uint32_t value)
 {
-	return notif_rpc(OPTEE_RPC_NOTIFICATION_SEND, value);
+	return notif_rpc(OPTEE_RPC_NOTIFICATION_SEND, value, 0);
+}
+
+TEE_Result notif_wait_timeout(uint32_t value, uint32_t tmo)
+{
+        return notif_rpc(OPTEE_RPC_NOTIFICATION_WAIT, value, tmo);
```
jenswi-linaro, line 226: Is this an ABI change? What happens if the kernel isn't recent enough?

core/include/kernel/notif.h
ZheTing815: ok, I will spilt it.

PR #6338, opened 2023-10-04
core: riscv: Fix condition of is_from_user()

gagachang:
RISC-V defines that xPP(previous privilege mode) field of CSR status indicates the previous privilege level prior to the trap. Since the encoding of user mode is 0, we should compare the xPP field with 0 here to know that the trap is from user mode.


gagachang:
Hello, @maroueneboubakri 
Please review this minor fix, thanks!

maroueneboubakri:
Hi @gagachang thanks for the patch:

PR #6339, opened 2023-10-04
drivers: regulator: list supported levels

etienne-lms:
Adds regulator API function regulator_supported_voltages() to get the list of the voltage levels supported by the regulator.

Voltage level array is either an array of increasing ordered levels, in microvolt, or is a triplet [min, max, step] for linear step incremental levels.

This P-R is extracted from RFC https://github.com/OP-TEE/optee_os/pull/6301.

### DIFF
--- a/core/include/drivers/regulator.h
+++ b/core/include/drivers/regulator.h
@@ -40,6 +40,37 @@ struct regu_dt_desc {
 };
 
 /*
+ * Defines the format of struct voltages::entries
+ *
+ * If regulator_voltages::type is VOLTAGE_TYPE_FULL_LIST, then
+ * regulator_voltages@entries stores regulator_voltages::num_levels cells,
+ * listing supported voltage levels in uV from lowest to highest value.
+ *
+ * If regulator_voltages::type is VOLTAGE_TYPE_INCREMENT, then
+ * regulator_voltages::entries stores 3 cells: min level, max level and
+ * level increment step, all in uV. When so, regulator_voltages::num_levels
+ * is meaningless.
+ */
+enum voltage_type {
+	VOLTAGE_TYPE_INVALID = 0,
+	VOLTAGE_TYPE_FULL_LIST, /* extensive list in uV */
+	VOLTAGE_TYPE_INCREMENT  /* min, max, increment (in uV) */
+};
+
+/*
+ * struct regulator_voltages - Voltage levels description
+ * @type: Type of level description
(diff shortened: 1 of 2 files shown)


core/include/drivers/regulator.h
```diff
@@ -76,13 +80,19 @@ struct regulator {
  * @get_state: Get regulator effective state
  * @set_voltage: Set voltage level in microvolt (uV)
  * @get_voltage: Get current voltage in microvolt (uV)
+ * @supported_voltages: Get an array of supported levels in microvolt (uV)
+ *      When successful, @count is the number of cells in @levels_us
+ *      or @count is 0 when @levels_us is a 3 cells array [min, max, step],
+ *      in microvolt, for linear incremental level lists.
```
jforissier, line 86: I am not very comfortable with `count == 0` for a 3-cell array, and the fact that the array is not constant format nor self-describing. Wouldn't it be a bit easier to define a struct?
```C
/* Defines the format of struct voltages::entries */
enum voltage_type {
        VOLTAGE_TYPE_FULL_LIST, /* extensive list in uV */
        VOLTAGE_TYPE_INCREMENT  /* min, max, increment (in uV) */
};
struct voltages {
        enum voltage_type type;
        size_t num_levels;
        /* Dynamic part: num_levels entries */
        /* int entries[] */
};
```

etienne-lms: ok, makes sense. I'll update.
Thanks for the proposal.

core/include/drivers/regulator.h
```diff
@@ -39,6 +39,37 @@ struct regu_dt_desc {
 	struct regulator *regulator;
 };
 
+/*
+ * Defines the format of struct voltages::entries
+ *
+ * If regulator_voltages::type is VOLTAGE_TYPE_FULL_LIST, then
+ * regulator_voltages@entries stores regulator_voltages::num_levels cells,
+ * listing supported voltage levels in uV from lowestto highest value.
```
jforissier, line 47: Missing space `lowest to`

jforissier (commented):

etienne-lms:
I addressed the _lowestto_ typo, squashed the fixup commits and applied your review tag.
Thanks.

make check-rust failed with `No space left on device (os error 28)`.

jforissier:
> I addressed the _lowestto_ typo, squashed the fixup commits and applied your review tag. Thanks.

Thanks!

> make check-rust failed with `No space left on device (os error 28)`.

I'll see if I can find a way to save some space when running the Rust job.

PR #6342, opened 2023-10-04
drivers: regulator: DT bindings properties always-on & pull-down

etienne-lms:
Consider regulator DT binding properties `regulator-always-on` and `regulator-pull-down`.

### DIFF
--- a/core/drivers/regulator/regulator_dt.c
+++ b/core/drivers/regulator/regulator_dt.c
@@ -17,6 +17,23 @@
 #include <util.h>
 
 /*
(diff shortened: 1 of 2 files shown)


jenswi-linaro (commented):

etienne-lms:
tag applied.
make check-rust failed with `/usr/bin/ld: final link failed: No space left on device`

jforissier:
> tag applied. make check-rust failed with `/usr/bin/ld: final link failed: No space left on device`

Let's ignore that for now.
Thanks.

PR #6345, opened 2023-10-04
drivers: regulator: GPIO controlled regulator

etienne-lms:
Implements a GPIO controlled regulators driver compliant with DT nodes compatible with regulator-gpio. These regulators use GPIO pins to select the voltage level. The implementation supports only dual voltage level selection using a single pin. The DT bindings allows more pins to select between more voltages but no known platform currently requires that so we preferred the simplified case.

This P-R is extracted from RFC https://github.com/OP-TEE/optee_os/pull/6301.


core/drivers/regulator/regulator_gpio.c
```diff
@@ -0,0 +1,261 @@
...
+					 struct regulator_gpio *regu)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	const fdt32_t *cuint = NULL;
+	struct gpio *gpio = NULL;
+	void *gpio_ref = &gpio;
+	int len = 0;
+
+	/*
+	 * DT bindings allows more than 1 GPIO to controller more than
```
GseoC, line 152: s/to controller/to control

maybe simplify to:
"DT bindings allows one GPIO to control ..."

GseoC (commented):
With comment addressed:

core/drivers/regulator/regulator_gpio.c
etienne-lms: I fixed the typo but did not rephrase as suggested (I think it's less clear why the proposed implementation supports only 1 GPIO)

GseoC: Ok, no problem

etienne-lms:
typo fixed and tag applied

etienne-lms:
Rebased.
Review tag is applied.

etienne-lms:
Anymore review comments on this P-R?
If not, maybe it can be merged.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -99,6 +99,7 @@ $(call force,CFG_DRIVERS_REGULATOR,y)
 $(call force,CFG_GIC,y)
 $(call force,CFG_INIT_CNTVOFF,y)
 $(call force,CFG_PSCI_ARM32,y)
+$(call force,CFG_REGULATOR_GPIO,y)
```
jforissier, line 103: It looks like `force` is being abused here, because I suppose the platform did work before this commit, which implies that `CFG_REGULATOR_GPIO` disabled is a valid config.

jforissier (commented):
For "plat-stm32mp1: conf: fix order for CFG_REGULATOR_FIXED":

core/arch/arm/plat-stm32mp1/conf.mk
etienne-lms: Indeed the platform does not mandate `CFG_REGULATOR_GPIO`. I think it's simpler embeding the driver and let OP-TEE DT tell whether it's used or not.

I can switch to `CFG_REGULATOR_GPIO ?= y`. @GseoC, what do you think?

etienne-lms: comment addressed.

etienne-lms:
@GseoC, I see i forgot to apply your A-b tag on the 2 stm32m1 config related commits. I'll when squashing the fixup commit.

etienne-lms:
Fixup commit squashed.
Tags applied.

@GseoC, I applied your tag despite I made the below change since you posted your R-b tag. Could you confirm your tag still applies?
```patch
-$(call force,CFG_REGULATOR_GPIO,y)
 (...)
+CFG_REGULATOR_GPIO ?= $(CFG_STM32_GPIO)
```

GseoC:
Yes, sure. It's better this way

etienne-lms:
Thanks Gatien. Your review tag is applied.

PR #6349, opened 2023-10-06
core: pta: imx: fix typo DEK blob command

clementfaure:
The PTA command had the manufacturing protection prefix instead of the DEK blob prefix.


jforissier (commented):
Oops!

PR #6355, opened 2023-10-09
core: ffa: use FFA_NORMAL_WORLD_RESUME

jenswi-linaro:
Prior to this FFA_INTERRUPT was always completed using FFA_MSG_WAIT, but at S-EL1 FFA_NORMAL_WORLD_RESUME should be used instead. So fix this by completing a secure interrupt that has preempted the normal world with FFA_NORMAL_WORLD_RESUME if configured with SPMC at S-EL1.

Fixes: 67fec989b586 ("core: ffa: correct response to FFA_INTERRUPT")


balint-dobszay-arm:

jenswi-linaro:
Rebased and tag applied

PR #6358, opened 2023-10-10
Fix Rust CI

jforissier:
Two commits related to the Rust CI job. THe first one fixes the "no space left on device" issue, the other is a cleanup.


jenswi-linaro (commented):

etienne-lms (commented):

PR #6360, opened 2023-10-10
core: sp: fix FFA_MEM_RECLAIM checks

balint-dobszay-arm:
Currently it is assumed that a NS endpoint's ID is always 0. This is not true if multiple VMs are present in the NWd, so the check will fail when reclaiming shared memory from an SP. Fix this by removing the owner ID check if the reclaim call comes from NWd, in this case the necessary checks are done by the hypervisor.


jenswi-linaro:

balint-dobszay-arm:
Can we get this merged please?

balint-dobszay-arm:
Rebased on top of master

jforissier:
> Can we get this merged please?

Sure. Thanks!

PR #6361, opened 2023-10-10
core interrupt: use device tree data

etienne-lms:
Following changes done recently in the interrupt framework, this P-R is based on https://github.com/OP-TEE/optee_os/pull/6097. It provides means to use DT bindings to describe interrupts and their controllers.

2nd commit adds an API function (`dt_register_interrupt_provider()`)for interrupt controller drivers to register as itr_chip providers into the DT_DRIVER framework and few API functions for consumer driver to use DT info to find their interrupt (`interrupt_dt_get()`, `interrupt_dt_get_by_index()`, `interrupt_dt_get_by_name()`).

3rd commit Adds a API function to register a callback handler for an interrupt. 
`interrupt_create_handler()` differs from existing `interrupt_add_handler()` in that the former does not reconfigure the interrupt since already done by `interrupt_dt_get()` or friend. Maybe `interrupt_add_handler2()` would be a better label.

Last commit registers GIC driver as a `DT_DRIVER_INTERRUPT` driver so that interrupt consumer drivers can use `interrupt_dt_get_by_index()` and friends.

### DIFF
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -11,6 +11,7 @@
 #include <drivers/gic.h>
 #include <keep.h>
 #include <kernel/dt.h>
+#include <kernel/dt_driver.h>
 #include <kernel/interrupt.h>
 #include <kernel/panic.h>
 #include <mm/core_memprot.h>
@@ -656,3 +657,64 @@ static void gic_op_set_affinity(struct itr_chip *chip, size_t it,
 
 	gic_it_set_cpu_mask(gd, it, cpu_mask);
 }
+
+#ifdef CFG_DT
+/* Callback for "interrupts" and "interrupts-extended" DT node properties */
+static TEE_Result dt_get_gic_chip_cb(struct dt_pargs *arg, void *priv_data,
+				     struct itr_desc *itr_desc)
+{
+	int itr_num = DT_INFO_INVALID_INTERRUPT;
+	struct itr_chip *chip = priv_data;
+	uint32_t phandle_args[2] = { };
+	uint32_t type = 0;
+	uint32_t prio = 0;
+
+	assert(arg && itr_desc);
+
+	/*
+	 * gic_dt_get_irq() expects phandle arguments passed are still in DT
+	 * format (big-endian) whereas struct dt_pargs carries converted
+	 * formats. Therefore swap again phandle arguments. gic_dt_get_irq()
+	 * consumes only the 2 first arguments.
+	 */
+	if (arg->args_count < 2)
+		return TEE_ERROR_GENERIC;
+	phandle_args[0] = cpu_to_fdt32(arg->args[0]);
+	phandle_args[1] = cpu_to_fdt32(arg->args[1]);
+
+	itr_num = gic_dt_get_irq((const void *)phandle_args, 2, &type, &prio);
+	if (itr_num == DT_INFO_INVALID_INTERRUPT)
+		return TEE_ERROR_GENERIC;
+
+	gic_op_add(chip, itr_num, type, prio);
+
+	itr_desc->chip = chip;
+	itr_desc->itr_num = itr_num;
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result gic_probe(const void *fdt, int offs, const void *cd __unused)
+{
+	if (interrupt_register_provider(fdt, offs, dt_get_gic_chip_cb,
+					&gic_data.chip))
+		panic();
+
+	return TEE_SUCCESS;
+}
+
+static const struct dt_device_match gic_match_table[] = {
+	{ .compatible = "arm,cortex-a15-gic" },
+	{ .compatible = "arm,cortex-a7-gic" },
+	{ .compatible = "arm,cortex-a5-gic" },
+	{ .compatible = "arm,cortex-a9-gic" },
+	{ .compatible = "arm,gic-400" },
+	{ }
+};
+
+DEFINE_DT_DRIVER(gic_dt_driver) = {
+	.name = "gic",
+	.match_table = gic_match_table,
+	.probe = gic_probe,
+};
+#endif /*CFG_DT*/
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -6,6 +6,7 @@
 
 #include <arm.h>
 #include <assert.h>
+#include <dt-bindings/interrupt-controller/arm-gic.h>
 #include <config.h>
 #include <compiler.h>
 #include <drivers/gic.h>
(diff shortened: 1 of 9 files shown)


core/drivers/gic.c
```diff
@@ -604,3 +605,67 @@ static void gic_op_set_affinity(struct itr_chip *chip, size_t it,
...
+	uint32_t prio = 0;
+	int n = 0;
+
+	/*
+	 * gic_dt_get_irq() expects phandle arguments passed are still in DT
+	 * format (big-endian) whereas struct dt_pargs carries converted
+	 * formats. Therefore swap again phandle arguments.
+	 */
+	for (n = 0; n < arg->args_count; n++)
+		arg->args[n] = cpu_to_fdt32(arg->args[n]);
```
etienne-lms, line 627: This is not nice. It modifies `arg` content. If called several times (some late init deferral), them `arg` content will not be reliable. I'll change into using a local copy of the arguments on which to call `cpu_to_fdt32()`.

etienne-lms:
I've squashed the fixup commits.

core/include/kernel/interrupt.h
```diff
@@ -343,4 +358,113 @@ TEE_Result interrupt_alloc_add_handler(struct itr_chip *chip, size_t it_num,
...
+ *
+ * @args	Reference to phandle arguments
+ * @data	Pointer to data given at interrupt_register_provider() call
+ * @out_itr_desc Output pointer to an allocated struct itr_desc upon success
+ * Return TEE_SUCCESS in case of success.
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if controller is not initialized.
+ * Return another TEE_Result code otherwise.
+ *
+ * Upon success, struct itr_desc must point to memory allocated with malloc()
+ * or like as it is freed prior returning to caller function.
```
jforissier, line 392: "freed with free()"?

jforissier (commented):
One small comment for "core: interrupt: registering interrupt providers", with or without it:

jforissier (commented):
"core: dt_driver: don't enforce phandle 1st arg is a phandle":

jforissier (commented):
"core: interrupt: add interrupt_create_handler()":

jforissier (commented):
"drivers: gic: register to dt_driver":

jforissier (commented):
"drivers: gic: use DT bindings":

etienne-lms:
https://github.com/OP-TEE/optee_os/pull/6361#discussion_r1368734996 addressed.
Tags applied.
Thanks for the review.

I've created a P-R in optee_docs for these new API functions: https://github.com/OP-TEE/optee_docs/pull/218.

jforissier:
Please rebase onto master to fix the IBART error, thanks!

etienne-lms:
right, thanks.

etienne-lms:
rebased

etienne-lms:
Arrg! IBART still fails.

etienne-lms:
I did not expect this change would make such platforms to consume more heap. IBART does not use DT for OP-TEE drivers inits.

core/kernel/interrupt.c
```diff
@@ -196,3 +196,129 @@ void interrupt_remove_free_handler(struct itr_handler *hdl)
...
+		res = get_legacy_interrupt_by_index(fdt, node, index, &desc);
+
+	assert((!res && desc) || (res && !desc));
+
+	if (!res) {
+		*chip = desc->chip;
+		*itr_num = desc->itr_num;
+
+		/* Balance malloc() or like from itr_dt_get_func callback */
+		free(desc);
```
jenswi-linaro, line 347: This is odd. Why not allocate this on the stack?

etienne-lms: Makes sense. I think it's feasible.
`get_of_device_func` function type and `dt_driver_device_from_xxx()` currently expect the reference is an output only ref, but I think it's fine if also used as a input argument. I'll propose something.

etienne-lms:
Comment addressed.

core/drivers/gic.c
```diff
@@ -610,11 +610,11 @@ static void gic_op_set_affinity(struct itr_chip *chip, size_t it,
 #ifdef CFG_DT
 /* Callback for "interrupts" and "interrupts-extended" DT node properties */
 static TEE_Result dt_get_gic_chip_cb(struct dt_pargs *arg, void *priv_data,
-				     struct itr_desc **out_itr_desc)
+				     struct itr_desc **itr_desc_p)
```
jenswi-linaro, line 613: Why is this a double-pointer? I can't see that it's needed anywhere and it makes the code more complicated.

etienne-lms: The generic `dt_driver_device_from_xxx()` functions expect a `void **` reference (e.g. [`dt_driver_device_from_node_idx_prop()`](https://github.com/OP-TEE/optee_os/blob/4.0.0/core/include/kernel/dt_driver.h#L156)), hence i used a double pointer reference here also, for consistency. Maybe over killing.

jenswi-linaro: I don't quite see what's consistent here, but it does look strange and error-prone to me.

etienne-lms: Ok, i'll add a fixup patch in the hope I understood your concern.

etienne-lms:
rebased.

jenswi-linaro:
If "core: kernel: dt_driver: reference output device reference as void *" was possible before, why did we ever have pointer-pointers as arguments to those functions?

etienne-lms:
My bad. The several driver frameworks use `struct clk **`, `struct rstctrl **`, `struct regulator **`, etc... hence I chose `void **` generic type in _core/kernel/dt_driver.c_ to represent that (see commit https://github.com/OP-TEE/optee_os/commit/b357d34fe91f4e7f6e0eacea17a7fbe5f6c01e7e) in the generic functions the several device frameworks rely on. Maybe a bad idea, I could have used `void *` at the time.

Note I will place commit "core: kernel: dt_driver: reference output device reference as void *" at the beginning of the series when I will squash the fixup commits.

etienne-lms:
I've squashed the fixup commits and reordered the series.

@jforissier, I removed your A-b tag from commits "core: interrupt: registering interrupt providers"
and "drivers: gic: register to dt_driver" as content sligthly  changed since you reviewed them.

jenswi-linaro:

jforissier:
> I've squashed the fixup commits and reordered the series.
> 
> @jforissier, I removed your A-b tag from commits "core: interrupt: registering interrupt providers" and "drivers: gic: register to dt_driver" as content sligthly changed since you reviewed them.

`Acked-by: Jerome Forissier <jerome.forissier@linaro.org>` for both, thanks.

etienne-lms:
@jbech-linaro, could you help me understand why IBART fails with this series? https://github.com/OP-TEE/optee_os/pull/6361#issuecomment-1787195610 mentions a core heap exhaustion but since the platform used by IBART does not embed any secure DTB, I don't understand why these changes make the platform to more core heap.

etienne-lms:
I forgot to mention that i've applied the review tags.

jbech-linaro:
As discussed in email. I've been able to reproduce the error on two difference machines. So, although I don't now that causes it, it at least seems reproducible.

etienne-lms:
IBART CI test failure should be fixed by https://github.com/OP-TEE/optee_test/pull/711.

etienne-lms:
IBART's happy!

PR #6323, opened 2023-09-26
RISC-V: Implement SBI based protocol

gagachang:
Based on https://github.com/OP-TEE/optee_os/pull/6268
This PR implements SBI based protocol.

In SBI protocol, OP-TEE prepares function arguments in registers a0~a5.
The register a7 encodes SBI TEE extension ID, which is 0x544545(TEE), temporarily.
Then OP-TEE issues `ecall` to trap into higher privileged software, e.g., openSBI.
The higher privileged software will handle this call and execute OS context switch, etc.

PR #6268, opened 2023-08-29
riscv: add TEE ABI

maroueneboubakri:
As per discussion https://github.com/OP-TEE/optee_os/issues/6265, the goal of this PR is to implement TEE ABI for RISC-V allowing OPTEED running on non-secure domain to communicate with OP-TEE OS running in secure domain.
This PR does nothing special except of reanming definitions containing `smc` to `abi` to have generic definitions or maybe comon ones to all architectures.

PR #6362, opened 2023-10-11
[PATCH v2] RISC-V: Implement SBI based protocol 

gagachang:
Based on https://github.com/OP-TEE/optee_os/pull/6268
And apply recommendations from patch v1 https://github.com/OP-TEE/optee_os/pull/6323
This PR implements SBI based protocol.

In SBI protocol, OP-TEE prepares function arguments in registers a0~a5.
The register a7 encodes SBI TEE extension ID, which is 0x544545(TEE), temporarily.
Then OP-TEE issues ecall to trap into higher privileged software, e.g., openSBI.
The higher privileged software will handle this call and execute OS context switch, etc.

OP-TEE also registers its vector table, which has address of entry functions for service, into higher privileged software. The higher privileged software can jump into OP-TEE according to this vector table.

### DIFF
--- a/core/arch/riscv/include/kernel/thread_private_arch.h
+++ b/core/arch/riscv/include/kernel/thread_private_arch.h
@@ -59,6 +59,8 @@ void thread_init_tvec(void);
 void thread_trap_vect(void);
 void thread_trap_vect_end(void);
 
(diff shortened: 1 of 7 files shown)


jenswi-linaro:
Looks good to me.

PR #6365, opened 2023-10-11
core: ffa: fix race in mobj_put() and ffa_inactivate()

jenswi-linaro:
Prior to this patch there was a race condition when mobj_put() is calling ffa_inactivate().
D/TC:0 0 ffa_inactivate:525 cookie 0x100000000000
D/TC:0 1 mobj_ffa_get_by_cookie:401 cookie 0x100000000000 active: refc 1 D/TC:? 1 read_console:114 got 0xd
D/TC:0 1 ffa_inactivate:525 cookie 0x100000000000
D/TC:0 0 ffa_inactivate:525 cookie 0x100000000000
E/TC:0 0 Panic at core/arch/arm/mm/mobj_ffa.c:527 <ffa_inactivate> E/TC:0 0 TEE load address @ 0xe100000
E/TC:0 0 Call stack:
E/TC:0 0  0x0e108c0c print_kernel_stack at ??:?
E/TC:0 0  0x0e115b8c __do_panic at core/kernel/panic.c:24 E/TC:0 0  0x0e10a238 ffa_inactivate at mobj_ffa.c:? E/TC:0 0  0x0e107318 __thread_std_smc_entry at ??:?

As now explained in ffa_inactivate():
/*
 * pop_from_list() can fail to find the mobj if we had just
 * decreased the refcount to 0 in mobj_put() and was going to
 * acquire the shm_lock but another thread found this mobj and
 * reinitialized the refcount to 1. Then before we got cpu time the
 * other thread called mobj_put() and deactivated the mobj again.
 * ... */

If our thread is delayed even further we may even inactivate an unrelated mobj that happened to reuse the same piece of memory.

Fix this by adding another guarding condition so that the mobj is guaranteed to be valid until ffa_inactivate() has returned. By adding a new member in struct mobj_ffa, inactive_refs, we keep track of references even when the mobj have been moved to the inactive list.

Adds a comment describing the non-trivial life cycle of struct mobj_ffa.

### DIFF
--- a/core/arch/arm/mm/mobj_ffa.c
+++ b/core/arch/arm/mm/mobj_ffa.c
@@ -15,16 +15,75 @@
 #include <mm/mobj.h>
 #include <sys/queue.h>
 
+/*
+ * Life cycle of struct mobj_ffa
+ *
+ * SPMC at S-EL1 (CFG_CORE_SEL1_SPMC=y)
+ * ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ * During FFA_MEM_SHARE allocated in mobj_ffa_sel1_spmc_new() and finally
+ * added to the inactive list at the end of add_mem_share() once
+ * successfully filled in.
+ *	registered_by_cookie = false
+ *	mobj.refs.val = 0
+ *	inactive_refs = 0
+ *
+ * During FFA_MEM_RECLAIM reclaimed/freed using
+ * mobj_ffa_sel1_spmc_reclaim().  This will always succeed if the normal
+ * world is only calling this when all other threads are done with the
+ * shared memory object. However, there are some conditions that must be
+ * met to make sure that this is the case:
+ *	mobj not in the active list, else -> return TEE_ERROR_BUSY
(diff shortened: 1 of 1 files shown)


core/arch/arm/mm/mobj_ffa.c
```diff
@@ -15,16 +15,75 @@
...
+ * - move the mobj into the active list
+ * - if not registered_by_cookie ->
+ *	set registered_by_cookie and increase inactive_count
+ * - set mobj.refc.val to 1
+ * - increase inactive_count
+ *
+ * A previously activated mobj is made ready for reclaim using
+ * mobj_ffa_unregister_by_cookie() which only succeeds if the mobj is in
+ * the inactive list and unregistered_by_cookie isn't set and then:
+ * - sets unregistered_by_cookie
```
etienne-lms, line 50: "the inactive list and **~~un~~**registered_by_cookie is**~~n't~~** set and then:"
"- **~~sets~~ clears** **~~un~~**registered_by_cookie"

jenswi-linaro:
Addressing comments.

etienne-lms (commented):

jenswi-linaro:
Squashed and tag applied.

PR #6367, opened 2023-10-11
drivers: regulator: consider DT property regulator-boot-on

etienne-lms:
Defines regulator flag `REGULATOR_BOOT_ON` for regulators with the `regulator-boot-on` property in their DT node.

### DIFF
--- a/core/include/drivers/regulator.h
+++ b/core/include/drivers/regulator.h
@@ -20,8 +20,15 @@
 #define REGULATOR_ALWAYS_ON	BIT(0)
 /* Enables pull down mode. DT property: regulator-pull-down */
 #define REGULATOR_PULL_DOWN	BIT(1)
+/*
+ * It's expected that this regulator was left on by the bootloader.
+ * The core shouldn't prevent it from being turned off later.
+ * DT property: regulator-boot-on
+ */
+#define REGULATOR_BOOT_ON	BIT(2)
 
-#define REGULATOR_FLAGS_MASK	(REGULATOR_ALWAYS_ON | REGULATOR_PULL_DOWN)
+#define REGULATOR_FLAGS_MASK	(REGULATOR_ALWAYS_ON | REGULATOR_PULL_DOWN | \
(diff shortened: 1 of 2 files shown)


core/include/drivers/regulator.h
```diff
@@ -20,8 +20,15 @@
 #define REGULATOR_ALWAYS_ON	BIT(0)
 /* Enables pull down mode. DT property: regulator-pull-down */
 #define REGULATOR_PULL_DOWN	BIT(1)
+/*
+ * regulator-boot-on:
+ * It's expected that this regulator was left on by the bootloader.
```
GseoC, line 25: by the first stage bootloader?

GseoC (commented):
with or without comment addressed:

core/include/drivers/regulator.h
etienne-lms: Effective number of bootstages depends on the platform.

etienne-lms:
@jbech-linaro, could you tell me where IBART fails?

etienne-lms:
Thanks @GseoC. While applying your review tag, I saw the description was not very consistent with the 2 other flag description. I applied your A-b tag but I also applied the below change:
```patch
 /* Enables pull down mode. DT property: regulator-pull-down */
 #define REGULATOR_PULL_DOWN    BIT(1)
 /*
- * regulator-boot-on:
  * It's expected that this regulator was left on by the bootloader.
  * The core shouldn't prevent it from being turned off later.
+ * DT property: regulator-boot-on
  */
 #define REGULATOR_BOOT_ON      BIT(2)
 
```

jforissier:
> @jbech-linaro, could you tell me where IBART fails?

I suspect it's due to the MBedTLS update. IBART does not test a merged branch contrary to GitHub actions. So you need to rebase manually.

etienne-lms:
Ok, thanks.

etienne-lms:
commit rebased

PR #6369, opened 2023-10-11
core: riscv: Fix logic of thread_{get/set}_exceptions()

gagachang:
In ARM, the bits in DAIF register are used to mask the interrupts. While in RISC-V, the bits in CSR XIE are used to enable(unmask) corresponding interrupt sources.

To not modify the function of thread_get_exceptions(), we invert the bits after reading the value of CSR XIE, as mask.

To not modify the function of thread_set_exceptions(), we invert the bits in given "exceptions" before writing "exceptions" into CSR XIE. Therefore, the intended masked exception bits will be cleared when we write the final value into CSR XIE to mask those interrupts.

### DIFF
--- a/core/arch/riscv/kernel/thread_arch.c
+++ b/core/arch/riscv/kernel/thread_arch.c
@@ -41,9 +41,12 @@ void __noreturn __panic_at_abi_return(void)
 	panic();
 }
 
+/* This function returns current masked exception bits. */
 uint32_t __nostackcheck thread_get_exceptions(void)
 {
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

jenswi-linaro (commented):

jforissier:
@gagachang please rebase to resolve the conflict and add the Acked-by: tags, thanks!

gagachang:
Hi @jforissier 
I rebased and applied tha tags, thanks!

PR #6373, opened 2023-10-12
Small ffa fixes

jenswi-linaro:

### DIFF
--- a/core/arch/arm/kernel/thread_spmc.c
+++ b/core/arch/arm/kernel/thread_spmc.c
@@ -653,7 +653,8 @@ static void handle_blocking_call(struct thread_smc_args *args,
 	default:
 		EMSG("Unhandled blocking service ID %#"PRIx32,
 		     (uint32_t)args->a3);
-		panic();
+		spmc_set_args(args, direct_resp_fid, swap_src_dst(args->a1), 0,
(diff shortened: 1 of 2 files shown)


core/arch/arm/kernel/thread_spmc.c
```diff
@@ -653,7 +653,8 @@ static void handle_blocking_call(struct thread_smc_args *args,
 	default:
 		EMSG("Unhandled blocking service ID %#"PRIx32,
 		     (uint32_t)args->a3);
-		panic();
+		spmc_set_args(args, direct_resp_fid, swap_src_dst(args->a1), 0,
+			      TEE_ERROR_BAD_PARAMETERS, 0, 0);
```
etienne-lms, line 657: I think it's rather an FFA code that is expected: `FFA_INVALID_PARAMETERS`?

jenswi-linaro: That should be the case with `FFA_ERROR`, but this is an `FFA_DIRECT_RESPONSE` that carries an OP-TEE specific value.

etienne-lms: Ok

etienne-lms (commented):

jenswi-linaro:
Tag applied

PR #6368, opened 2023-10-11
drivers: implement HiSilicon Queue Management (QM) module

xiaoxuZeng:
The Hisilicon QM is a Queue Management module.
In order to unify the interface between accelerator and software, a unified queue management module QM is used to interact with software. Each accelerator module integrates a QM. Software issues tasks to the SQ (Submmision Queue),and the QM obtains the address of the SQE (Submmision Queue Element). The BD (Buffer Description, same as SQE) information is sent to the accelerator. After the task processing is complete, the accelerator applies for a write-back address from the QM to write back the SQ.

PR #6377, opened 2023-10-13
[part1] The Hisilicon QM is a Queue Management module.

xiaoxuZeng:
In order to unify the interface between accelerator and software, a unified queue management module QM is used to interact with software. Each accelerator module integrates a QM. Software issues tasks to the SQ (Submmision Queue),and the QM obtains the address of the SQE (Submmision Queue Element). The BD (Buffer Description, same as SQE) information is sent to the accelerator. After the task processing is complete, the accelerator applies for a write-back address from the QM to write back the SQ.


xiaoxuZeng:
The original code is too large, and we plan to split it into smaller patches for easier review.
https://github.com/OP-TEE/optee_os/pull/6368

loubaihui:
Acked-by: loubaihui <loubaihui1@huawei.com>

PR #6378, opened 2023-10-13
core: fix data abort during ftrace

jenswi-linaro:
With commit c10e3fa93d24 ("core: fix race in handling TA panic") the resources of a panicked TAs is released as early as possible, including the user space mapped ftrace buffer. However, the pointer to the ftrace buffer is stored in the ts_session for quick and easy access. The ftrace buffer is always retrieved with get_fbuf() that already have a few other checks to see if the buffer is currently available. So add a check to see that the TA hasn't panicked also.

Fixes: c10e3fa93d24 ("core: fix race in handling TA panic")


jenswi-linaro:
Force push to please checkpatch.

etienne-lms (commented):

jforissier:

jenswi-linaro:
Tags applied. Did a `s/is released/are released/` in the commit message.

b49020:
Does this change disallows ftrace dump for user TAs which have panicked?

jforissier:
> Does this change disallows ftrace dump for user TAs which have panicked?

No. I could obtain a dump for e.g., `xtest regression_1007` which is a panic test.

PR #6379, opened 2023-10-13
Add Widevine related device tree options support

qazwsxedcrfvtg14:
On the new ChromeOS mediatek platform, we will use the device tree to pass hardware unique key and the parameters for widevine TA.

BTW,  should we also put the device tree binding document in the OP-TEE repo?
Because the Linux upstream says the OP-TEE only DT bindings don't belong in the kernel.
https://lore.kernel.org/all/20230918194235.GA1548023-robh@kernel.org/


core/arch/arm/plat-mediatek/cros_huk.c
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2023, The ChromiumOS Authors
```
jbech-linaro, line 3: For legal reasons, we need a real name, see https://optee.readthedocs.io/en/latest/general/contribute.html#developer-certificate-of-origin

core/pta/widevine.c
```diff
@@ -0,0 +1,142 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2023, The ChromiumOS Authors
```
jbech-linaro, line 3: For legal reasons, we need a real name, see https://optee.readthedocs.io/en/latest/general/contribute.html#developer-certificate-of-origin

core/pta/widevine.c
```diff
@@ -0,0 +1,142 @@
...
+
+	result = fdt_getprop(fdt, node, "widevine-root-of-trust", &len);
+	if (!result)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	if (len > WIDEVINE_PRIV_MAX_SIZE)
+		return TEE_ERROR_OVERFLOW;
+
+	memcpy(widevine_priv, result, len);
+	widevine_priv_size = len;
```
jbech-linaro, line 80: I am probably missing something obvious here, but, am I reading this correctly that we're reading the private key from device tree and then copy it into a secure (TEE) memory? If yes, how if this secure? What prevents the private key from being read out by others?

core/pta/widevine.c
```diff
@@ -0,0 +1,142 @@
...
+/* The private key for Widevine signing. */
+static uint8_t widevine_priv[WIDEVINE_PRIV_MAX_SIZE];
+static uint32_t widevine_priv_size;
+
+static TEE_Result init_widevine_dt_data(void)
+{
+	int node = 0;
+	int len = 0;
+	void *fdt;
+	const void *result = NULL;
```
jbech-linaro, line 51: Wouldn't it make more sense to rename `result` to `value`? `fdt_getprop(...)` and friends retrieves the value of a given property.

core/pta/widevine.c
```diff
@@ -0,0 +1,142 @@
...
+	/* Make sure we called from Secure World */
+	if (!ts_get_calling_session())
+		return TEE_ERROR_ACCESS_DENIED;
+
+	if (cmd == GET_TPM_PUBKEY) {
+		data = tpm_auth_pub;
+		data_length = tpm_auth_pub_size;
+	} else if (cmd == GET_WIDEVINE_PRIVKEY) {
+		data = widevine_priv;
+		data_length = widevine_priv_size;
```
jbech-linaro, line 110: This works for a system where we can guarantee that no other TA's are deployed. If other TA's could be deployed, then another TA could simply call this TA and just get the private key. I don't think we have something to support this, but it'd great if we could control and verify who is calling this function. @jenswi-linaro , any ideas?

core/pta/widevine.c
```diff
@@ -0,0 +1,142 @@
...
+	} else {
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	if (params[0].memref.size < data_length)
+		return TEE_ERROR_OVERFLOW;
+
+	params[1].value.a = data_length;
```
jbech-linaro, line 118: What the reason for providing the length in `params[1]`? Couldn't you simply use `params[0].memref.size = data_length;`?

core/pta/widevine.c
```diff
@@ -0,0 +1,142 @@
...
+#define TPM_AUTH_PUB_MAX_SIZE 512
+#define WIDEVINE_PRIV_MAX_SIZE 32
+
+/* The public key to encrypt the communication to TPM. */
+static uint8_t tpm_auth_pub[TPM_AUTH_PUB_MAX_SIZE];
+static uint32_t tpm_auth_pub_size;
+
+/* The private key for Widevine signing. */
+static uint8_t widevine_priv[WIDEVINE_PRIV_MAX_SIZE];
+static uint32_t widevine_priv_size;
```
jbech-linaro, line 44: Do you think you could say a bit more about the keys themselves. What kind of keys? AES, RSA, HMAC? Bits vs bytes? I.e., for example mention something "max key size in bits".

core/arch/arm/plat-mediatek/cros_huk.c
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
jbech-linaro, line 1: What makes this file `plat-mediatek` unique? It looks like this could be a generic function.

mk/config.mk
```diff
@@ -1063,3 +1063,8 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# Disable the widevine device tree by default
+CFG_WIDEVINE_DT ?= n
+CFG_WIDEVINE_HUK ?= $(CFG_WIDEVINE_DT)
+CFG_WIDEVINE_PTA ?= $(CFG_WIDEVINE_DT)
```
jbech-linaro, line 1070: This works, but I wonder if it wound't make sense to the order here slightly. Perhaps something like this:
```
CFG_WIDEVINE_PTA ?= n
CFG_WIDEVINE_DT ?= $(CFG_WIDEVINE_PTA)
CFG_WIDEVINE_HUK ?= $(CFG_WIDEVINE_PTA)
```

jbech-linaro (commented):
As a generic comment I think we'd like to change this to transfer lists, transfer elements in the future. But until that is in place DT works.

core/pta/widevine.c
etienne-lms: There is at least a way to check the calling TA has the expected UUID, something like:
```c
	if (to_ta_session(ts_get_current_session())->clnt_id.login != TEE_LOGIN_TRUSTED_APP ||
	    memcmp(&ts_get_calling_session()->ctx->uuid, (char *)&expected_ta_uuid, sizeof(TEE_UUID))) {
		return TEE_ERROR_ACCESS_DENIED;
	}
```
Using this in conjunction with TA namespaces (see doc on [verifying-with-subkeys](https://optee.readthedocs.io/en/latest/architecture/trusted_applications.html#verifying-with-subkeys)) allows specific PTAs to be reachable only from specific TAs.

jbech-linaro:
I think this is good now, please add my tag, rebase it and squash the patches accordingly so we merge it. Thanks!

qazwsxedcrfvtg14:
Rebased and squashed the patches. 
Thanks!

jforissier:
@qazwsxedcrfvtg14 please address the checkpatch issues (CI / Code style job). You may use the sentence you put in this PR ("On the new ChromeOS mediatek platform, we will use the device tree to pass hardware unique key and the parameters for widevine TA.").
 Thanks.

core/pta/widevine.c
```diff
@@ -0,0 +1,199 @@
...
+static TEE_Result init_widevine_dt_data(void)
+{
+	int node = 0;
+	int len = 0;
+	void *fdt = NULL;
+	const void *value = NULL;
+
+	fdt = get_dt();
+	if (!fdt)
+		return TEE_ERROR_NO_DATA;
```
etienne-lms, line 94: Should this check the DTB is located in secure memory?

core/pta/widevine.c
```diff
@@ -0,0 +1,199 @@
...
+
+	memcpy(widevine_priv, value, len);
+	widevine_priv_size = len;
+
+	return TEE_SUCCESS;
+}
+
+service_init(init_widevine_dt_data);
+
+static TEE_Result verify_caller(void)
```
etienne-lms, line 125: I think this should be done from session opening. REE client should not be allowed to open a session towards this PTA.

core/pta/widevine.c
```diff
@@ -0,0 +1,199 @@
...
+static TEE_Result get_dt_data(uint32_t ptypes, TEE_Param params[TEE_NUM_PARAMS],
+			      uint32_t cmd)
+{
+	uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INOUT,
+					  TEE_PARAM_TYPE_NONE,
+					  TEE_PARAM_TYPE_NONE,
+					  TEE_PARAM_TYPE_NONE);
+
+	uint8_t *data;
+	uint32_t data_length;
```
etienne-lms, line 157: init missing (`= NULL` + `= 0`)

mk/config.mk
```diff
@@ -1063,3 +1063,7 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# Disable the widevine PTA by default
+CFG_WIDEVINE_PTA ?= n
+CFG_WIDEVINE_HUK ?= $(CFG_WIDEVINE_PTA)
```
etienne-lms, line 1069: This is a specific WV HUK from DT. I think the config switch name should reflect that.

etienne-lms (changes requested):
Are these DT bindings (DT node pathes as `/options/` and property names) generic? Are they OP-TEE specific (e.g `optee-hardware-unique-key`) ? If these are to be deprecated once Firmware Handoff Transfer List is merged, is it worth merging this change now?

The commit message lacks a description, see CI Code Style result that also reports another issue).

core/drivers/widevine_huk.c
```diff
@@ -0,0 +1,63 @@
...
+
+service_init(init_widevine_huk_dt_data);
+
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	TEE_Result result;
+
+	result = init_widevine_huk_dt_data();
+	if (result != TEE_SUCCESS)
+		return result;
```
etienne-lms, line 58: replace this with a test on `dt_huk_initialized`

core/pta/widevine.c
```diff
@@ -0,0 +1,199 @@
...
+#include <trace.h>
+#include <util.h>
+
+#define PTA_NAME "widevine.pta"
+
+#define GET_TPM_PUBKEY 0
+#define GET_WIDEVINE_PRIVKEY 1
+
+/* 721f4da9-da05-40d4-a1a3-8377c1e08b0a */
+#define WIDEVINE_PTA_UUID                                              \
```
etienne-lms, line 28: This UUID should be exposed from _lib/libutee/include/pta_widevine.h_ as done for other PTAs.

qazwsxedcrfvtg14:
Resolved the comments.

BTW, I'm not sure about the status of "Firmware Handoff Transfer List"...
@Narflex do we have any plans for that?

Narflex:
> Resolved the comments.
> 
> BTW, I'm not sure about the status of "Firmware Handoff Transfer List"... @Narflex do we have any plans for that?

I'm not sure about our future plans for that, but it's not in the timeline for this change, so please proceed with this change as-is.

etienne-lms:
> so please proceed with this change as-is.

Please check first that such `/options/` DT node paths and new DT property are not a open door to messy DT bindings. 
DT bindings is a spec. It should remain somewhat consistent. Thanks for https://github.com/OP-TEE/optee_os/pull/6418, that helps discussion on a real proposal. Is it clear who is in charge of maintaining these bindings evolution and consistency. It seems the [pointed discussion thread](https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/20442/comment/bfbdaa52_659557e0/) did not lead to a definitive consensus.  Unless I missed something in which case I apologies for being noisy.

Narflex:
> > so please proceed with this change as-is.
> 
> Please check first that such `/options/` DT node paths and new DT property are not a open door to messy DT bindings. DT bindings is a spec. It should remain somewhat consistent. Thanks for #6418, that helps discussion on a real proposal. Is it clear who is in charge of maintaining these bindings evolution and consistency. It seems the [pointed discussion thread](https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/20442/comment/bfbdaa52_659557e0/) did not lead to a definitive consensus. Unless I missed something in which case I apologies for being noisy.

The DT bindings being added here are being documented in OP-TEE as you pointed out, so I would figure OP-TEE maintainers would be the ones maintaining the evolution/consistency of that if anything else gets added there in the future. Originally we tried to get these documented in the linux kernel, but got pushback on that which is why they ended up in OP-TEE....and I think that stance will continue to be taken where DT bindings should get added to kernel documentation unless there is pushback on that end, in which case they'd end up in OP-TEE.

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+#include <malloc.h>
+#include <mm/tee_mm.h>
+#include <mm/tee_pager.h>
+#include <pta_widevine.h>
+#include <stdio.h>
+#include <string.h>
+#include <string_ext.h>
+#include <tee_api.h>
+#include <trace.h>
+#include <util.h>
```
etienne-lms, line 21: can remove _boot.h_, _malloc.h_, _tee_mm.h_, _tee_pager.h_, _stdio.h_, _string_ext.h_ and _trace.h_.
Need to include _stdint.h_, _ts_manager.h_, and _tee_te_manager.h_.

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+}
+
+static TEE_Result get_dt_data(uint32_t ptypes, TEE_Param params[TEE_NUM_PARAMS],
+			      uint32_t cmd)
+{
+	uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INOUT,
+					  TEE_PARAM_TYPE_NONE,
+					  TEE_PARAM_TYPE_NONE,
+					  TEE_PARAM_TYPE_NONE);
+
```
etienne-lms, line 146: remove the empty line

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+			   sizeof(TEE_UUID)) == 0)
+			return TEE_SUCCESS;
+
+	return TEE_ERROR_ACCESS_DENIED;
+}
+
+static TEE_Result get_dt_data(uint32_t ptypes, TEE_Param params[TEE_NUM_PARAMS],
+			      uint32_t cmd)
+{
+	uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INOUT,
```
etienne-lms, line 142: it seems this should be a MEMREF_OUTPUT parameter

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+		data = tpm_auth_pub;
+		data_length = tpm_auth_pub_size;
+	} else if (cmd == PTA_WIDEVINE_GET_WIDEVINE_PRIVKEY) {
+		data = widevine_priv;
+		data_length = widevine_priv_size;
+	} else {
+		return TEE_ERROR_NOT_IMPLEMENTED;
+	}
+
+	params[0].memref.size = data_length;
```
etienne-lms, line 163: Must test `data_length` against `params[0].memref.size`:
```c
	if (data_length < params[0].memref.size) {
		params[0].memref.size = data_length;
		return TEE_ERROR_SHORT_BUFFER;
	}
```

I think you should return an error if there `data_length` is zero. Unless you're fine to return `TEE_SUCCESS` and let client consider zero sized buffer means the requested key was not provisioned.

lib/libutee/include/pta_widevine.h
```diff
@@ -0,0 +1,27 @@
...
+	{                                                              \
+		0x721f4da9, 0xda05, 0x40d4,                            \
+		{                                                      \
+			0xa1, 0xa3, 0x83, 0x77, 0xc1, 0xe0, 0x8b, 0x0a \
+		}                                                      \
+	}
+
+/*
+ * Widevine PTA supported commands
+ */
```
etienne-lms, line 23: please add a description of the commands. Something like:
```c
/*
 * PTA_WIDEVINE_GET_TPM_PUBKEY - Get Widevine TPM public key
 * PTA_WIDEVINE_GET_WIDEVINE_PRIVKEY - Get Widevine private key
 *
 * [out]     memref[0]        Retrieved key data
 *
 * Return codes:
 * TEE_SUCCESS - Invoke command success
 * TEE_ERROR_BAD_PARAMETERS - Incorrect parameters
 * TEE_ERROR_SHORT_BUFFER - Output buffer size is too small
 * TEE_ERROR_something - Requested data not available
 */
```
Note that describing the return code is not mandatory as long as it complies with the definition provided by the GP TEE specification.

I think the API should also describe the type/format of the keys, like what's done in _core/pta/widevine.c_. Not a strong opinion if you think it is implicitly state by keywords _"Widevine TPM public key"_ and _"Widevine private key"_.

I think you need a error code for when the requested data is not available (for when `init_widevine_dt_data()` failed).

mk/config.mk
```diff
@@ -1063,3 +1063,13 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# Disable the widevine PTA by default
```
etienne-lms, line 1067: this should rather describe what `CFG_WIDEVINE_PTA` relates to, not what is its default state.

Same comment for `CFG_WIDEVINE_HUK`. IIUC, `CFG_WIDEVINE_HUK` when enabled, uses the widevine HUK provided by secure DTB as OP-TEE HUK. Maybe using `CFG_WIDEVINE_HUK` puts some constraints how the HUK is derived (`CFG_CORE_HUK_SUBKEY_COMPAT` config switch value).

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+static TEE_Result open_session(uint32_t param_types __unused,
+			       TEE_Param params[TEE_NUM_PARAMS] __unused,
+			       void **sess_ctx __unused)
+{
+	size_t i;
+	struct ts_session *session = ts_get_calling_session();
+
+	/* Make sure we called from Secure World */
+	if (!session)
+		return TEE_ERROR_ACCESS_DENIED;
```
etienne-lms, line 123: This test is redundant with the next one.

core/drivers/widevine_huk.c
```diff
@@ -0,0 +1,61 @@
...
+	dt_huk_initialized = true;
+
+	return TEE_SUCCESS;
+}
+
+service_init(init_widevine_huk_dt_data);
+
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	init_widevine_huk_dt_data();
```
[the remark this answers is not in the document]
qazwsxedcrfvtg14: hmmm, because the DTB may be reachable only during initialization.
I think we should reserve `service_init(init_widevine_huk_dt_data);` to make sure we always init the data in the initialization stage. And calling `init_widevine_huk_dt_data` in `tee_otp_get_hw_unique_key` to make sure the data is initialized during initialization.

mk/config.mk
qazwsxedcrfvtg14: Fixed the comments.
BTW, IIUC, we don't have constraints between `CFG_WIDEVINE_HUK` & `CFG_CORE_HUK_SUBKEY_COMPAT`.

core/drivers/widevine_huk.c
```diff
@@ -0,0 +1,61 @@
...
+	return TEE_SUCCESS;
+}
+
+service_init(init_widevine_huk_dt_data);
+
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	TEE_Result result = init_widevine_huk_dt_data();
+
+	if (result != TEE_SUCCESS)
```
etienne-lms, line 52: nitpicking suggestion to make the implementation easier to read:
```c
	TEE_Result result = TEE_SUCCESS;

	/* Ensure we get data fomr the DT, in case called before service_init() handler */
	result = init_widevine_huk_dt_data();
	if (result != TEE_SUCCESS)
		return result;

	(...)

File: core/drivers/widevine_huk.c [hunk 25]
```diff
@@ -0,0 +1,61 @@
...
+		return result;
+
+	if (!dt_huk_initialized)
```
etienne-lms, line 55: I think this is useless since wthe above call to `init_widevine_huk_dt_data()`.

lib/libutee/include/pta_widevine.h
```diff
@@ -0,0 +1,37 @@
...
+ * PTA_WIDEVINE_GET_WIDEVINE_PRIVKEY - Get Widevine private key
+ *
+ * [out]     memref[0]        Retrieved key data
+ *
+ * Return codes:
+ * TEE_SUCCESS - Invoke command success
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect parameters
+ * TEE_ERROR_SHORT_BUFFER - Output buffer size is too small
+ * TEE_ERROR_NO_DATA - Requested data not available
+ * TEE_ERROR_ACCESS_DENIED - Caller is not a TA in the allow list
```
etienne-lms, line 32: `TEE_ERROR_ACCESS_DENIED` will be returned from open-session request, not from the 2 described commands.

mk/config.mk
```diff
@@ -1063,3 +1063,13 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# When enabled, CFG_WIDEVINE_PTA embeds a PTA that expose the keys under
```
etienne-lms, line 1067: s/expose/exposes/

core/pta/widevine.c
```diff
@@ -0,0 +1,187 @@
...
+#include <libfdt.h>
+#include <pta_widevine.h>
+#include <stdint.h>
+#include <string.h>
+#include <tee_api.h>
+#include <util.h>
+
+#define PTA_NAME "widevine.pta"
+
+#define TPM_AUTH_PUB_MAX_SIZE 512
```
etienne-lms, line 21: From the dt-bindings specification, it seems this should be 1024.

mk/config.mk
etienne-lms: OK. I don't really understand why using a Widevine TA mandates OP-TEE to get its HUK from the Widevine integration. And when so, that there is not strict scheme for how this HUK is to be used. But I guess there are considerations I don't catch.

qazwsxedcrfvtg14:
Rebased and resolved the conflicts with master branch.

core/pta/widevine.c
```diff
@@ -0,0 +1,187 @@
...
+ * The format of data should be TPM2B_PUBLIC.
+ * For more information, please reference the 12.2.5 section:
+ * https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part2_Structures_pub.pdf
+ */
+static uint8_t tpm_auth_pub[TPM_AUTH_PUB_MAX_SIZE];
+static uint32_t tpm_auth_pub_size;
+
+/*
+ * The Widevine root of trust secret. Used to sign the widevine
+ * request in OP-TEE. The value is an ECC NIST P-256 scalar.
```
etienne-lms, line 63: s/request/requests/ ?

core/pta/widevine.c
```diff
@@ -0,0 +1,187 @@
...
+service_init(init_widevine_dt_data);
+
+static TEE_Result open_session(uint32_t param_types __unused,
+			       TEE_Param params[TEE_NUM_PARAMS] __unused,
+			       void **sess_ctx __unused)
+{
+	size_t i = 0;
+	struct ts_session *session = ts_get_calling_session();
+
+	/* Make sure we called from a TA */
```
etienne-lms, line 117: "we **are** called"

ditto below

mk/config.mk
```diff
@@ -1077,3 +1077,13 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# When enabled, CFG_WIDEVINE_PTA embeds a PTA that exposes the keys under
+# DT node  "/options/widevine" to some specific TAs.
```
etienne-lms, line 1082: 2 space chars
update the node path

core/drivers/widevine_huk.c
```diff
@@ -0,0 +1,63 @@
...
+
+	fdt = get_secure_dt();
+	if (!fdt)
+		return TEE_ERROR_NO_DATA;
+
+	node = fdt_path_offset(fdt, "/options/op-tee/widevine");
+	if (node < 0)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	value = fdt_getprop(fdt, node, "google,optee-hardware-unique-key", &len);
```
etienne-lms, line 32: From the DT bindings [posted to dt-schemas.org](https://lore.kernel.org/all/20231110043241.2920553-1-yich@chromium.org/), "google,optee-hardware-unique-key" is a required property. Does that means `CFG_WIDEVINE_HUK=y` is also required when `CFG_WIDEVINE_PTA` is enabled?

qazwsxedcrfvtg14: Added CFG_WIDEVINE_HUK to the dependency of CFG_WIDEVINE_PTA.

qazwsxedcrfvtg14:
Resolved the comments.

etienne-lms (commented):
I would rather wait the DT bindings are acked before merging this change.

I still find curious that Widevine mandates the whole OP-TEE to rely on a HUK provided by DT "/options/op-tee/widevine" node without other strict constraints on how this HUK is derived. I also think it puts some restrictions on integration of other TAs when  the Widevine TA is to be supported, which maybe makes sense.

jbech-linaro:
I'm adding the "enhancement" tag, so this won't be closed automatically and since I want this to be merged in one or another way at some point. I do believe this will take some time to sort out. We've just this week had a first call between stakeholders interested in this and we concluded that we have to start with working out the Linux kernel side of things. The secure side, will be a topic in the future when the Linux kernel side of story seems to fall in place. For more information (and a recording of the meeting), please see the topics discussed [here](https://docs.google.com/document/d/1rWNDr4OGyvyQ1-phUsnNryYHohd-IltMZHnUZAh9WYg/edit#heading=h.dsxnk7m2ir23).

core/drivers/widevine_huk.c
```diff
@@ -0,0 +1,63 @@
...
+	if (!value)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	if (len >= HW_UNIQUE_KEY_LENGTH)
+		len = HW_UNIQUE_KEY_LENGTH;
+	else
+		return TEE_ERROR_BAD_FORMAT;
+
+	dt_huk_initialized = true;
```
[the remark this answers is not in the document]
qazwsxedcrfvtg14: Yeah, I missed that part...
Fixed.

Narflex:
The DT bindings documentation change has been merged here: https://github.com/OP-TEE/optee_docs/pull/230 Can we get this PR merged now?

qazwsxedcrfvtg14:
The device tree binding is merged in the optee_docs:
https://github.com/OP-TEE/optee_docs/pull/230

And rebased the PR.

jbech-linaro:
@jforissier @jenswi-linaro  anything else you'd like to add?

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+
+static TEE_Result open_session(uint32_t param_types __unused,
+			       TEE_Param params[TEE_NUM_PARAMS] __unused,
+			       void **sess_ctx __unused)
+{
+	size_t i = 0;
+	struct ts_session *session = ts_get_calling_session();
+
+	/* Make sure we are called from a TA */
+	if (to_ta_session(ts_get_current_session())->clnt_id.login !=
```
jenswi-linaro, line 119: You could check `is_user_ta_ctx(s->ctx)` instead. But if you rather rely on `clnt_id.login` here, then you should use `clnt_id.uuid` below.

core/pta/widevine.c
```diff
@@ -0,0 +1,188 @@
...
+/*
+ * Trusted Application Entry Points
+ */
+static TEE_Result invoke_command(void *psess __unused, uint32_t cmd,
+				 uint32_t ptypes,
+				 TEE_Param params[TEE_NUM_PARAMS])
+{
+	switch (cmd) {
+	case PTA_WIDEVINE_GET_TPM_PUBKEY:
+	case PTA_WIDEVINE_GET_WIDEVINE_PRIVKEY:
```
jenswi-linaro, line 178: Do you expect to add another `cmd`? You have an `if else` equivalent above in `get_dt_data()`

core/pta/widevine.c
qazwsxedcrfvtg14: Done

core/pta/widevine.c
qazwsxedcrfvtg14: Removed the switch case.

jenswi-linaro:

jforissier (commented):

Narflex:
Can we get this PR merged now?

jforissier:
@Narflex yes, sorry for the delay.

PR #6380, opened 2023-10-13
drivers: crypto: stm32: lower verbosity on SAES key derivation

etienne-lms:
Changes SAES context allocation/release trace message from debug level to flow level otherwise each access to the secure storage emits debug messages.


jforissier (commented):

etienne-lms:
tag applied

PR #6383, opened 2023-10-16
tree wide: fixes header guards + a missing notice

etienne-lms:
A series of fixes for header files guards which macro name should reflect the header file name and its base directory relative path, e.g. _kernel/interrupt.h_ is protected with `__KERNEL_INTERRUPT_H`.
By the way, I saw 2 header file lacking a notice (license/copyright): see the 1st commit.


jenswi-linaro:
I would prefer if we were consistent with using `__` as a prefix for these guards. I believe all .h files are used either internally by OP-TEE core or provided as part of "system libraries" to TAs. In especially the latter case we're in a situation where the `__` prefix makes perfect sense. In the former case, it's not that important, but it's still nice to have when integrating third-party libraries.

Those are my technical arguments for prefixing the guards with `__`.

What we do in optee_client is a different story, there we must not prefix anything with `__` as that's reserved for the compiler and/or system libraries vendor.

etienne-lms:
Ok, i'll update with `__` prefixes

jenswi-linaro:
> Ok, i'll update with `__` prefixes

Thanks for doing this cleanup

jenswi-linaro (commented):
With my small comment fixed:

etienne-lms:
_scif.h_ fixed.
Tag applied.
Thanks for the review.

PR #6390, opened 2023-10-20
core: ffa: fix spmc_retrieve_req() buffer

jenswi-linaro:
Prior to this patch when retrieving a shared memory block the memory transaction descriptor (struct ffa_mem_transaction_1_1 or struct ffa_mem_transaction_1_0) was read from the TX buffer instead of the RX buffer where the actual memory transaction is supplied. Fix this by changing to read the memory transaction descriptor from the RX buffer.

This used to work somewhat by chance since OP-TEE before the call to FFA_MEM_RETRIEVE_REQ_32 filled in a memory transaction descriptor of what we expect to retrieve in the TX buffer.

Fixes: f49f23f781e3 ("core: ffa: rename nw_rxtx to my_rxtx")
Fixes: a1c53023cc80 ("core: spmc: support FF-A 1.1")
Reported-by: Olivier Deprez <olivier.deprez@arm.com>


jenswi-linaro:
I'm sorry for this last-minute fix. It affects only FF-A configurations and the change is quite contained.

jforissier:
> I'm sorry for this last-minute fix. It affects only FF-A configurations and the change is quite contained.

Indeed.

jenswi-linaro:
Tag applied

balint-dobszay-arm:

jenswi-linaro:
Tag applied.

PR #6392, opened 2023-10-23
RISC-V: Small modifications for PLIC driver

gagachang:
1. Move `interrupt_main_handler()` from platform specific source file to PLIC driver.
2. Ignore interrupt source ID 0 since it means "no interrupt".

### DIFF
--- a/core/drivers/plic.c
+++ b/core/drivers/plic.c
@@ -275,7 +275,7 @@ void plic_it_handle(void)
 	struct plic_data *pd = &plic_data;
 	uint32_t id = plic_claim_interrupt(pd);
 
-	if (id <= pd->max_it)
+	if (id > 0 && id <= pd->max_it)
 		interrupt_call_handlers(&pd->chip, id);
 	else
 		DMSG("ignoring interrupt %" PRIu32, id);


core/drivers/plic.c
```diff
@@ -282,3 +282,8 @@ void plic_it_handle(void)
 
 	plic_complete_interrupt(pd, id);
 }
+
+void interrupt_main_handler(void)
+{
+	plic_it_handle();
+}
```
jforissier, line 289: What if a specific platform that uses the PLIC driver needs to do something else?

jforissier (commented):
- For "riscv: plic: Implement interrupt_main_handler() in PLIC driver":

  > Move interrupt_main_handler() from platform specific source file to
  > common PLIC driver.

  Why?

  >
  > Signed-off-by: Alvin Chang <alvinga@andestech.com>

- For "drivers: plic: Ignore interrupt source ID 0":

core/drivers/plic.c
gagachang: I referenced this commit https://github.com/OP-TEE/optee_os/commit/67e55c51c9149ea549664b3981ad9032dcf4ce7f.
But now I think you are right, we should not restrict platform's interrupt handling in this early stage.
I'll remove that commit.

gagachang:
Tag applied

PR #6400, opened 2023-10-24
core: ffa: remove AArch32 support

jenswi-linaro:
The FF-A and AArch32 configuration was prior to this patch not compile tested and not used upstream. So remove the AArch32 support for FF-A configurations so save maintenance effort.


jenswi-linaro:
As discussed during the LOC meeting today (https://www.trustedfirmware.org/meetings/).

etienne-lms:

jenswi-linaro:
Tag applied.

PR #6405, opened 2023-10-25
core: tee_svc.c: allow to pass non-NULL memref of size 0

vincent-mailhol:
When calling `TEE_InvokeTACommand()` with a non-NULL memref of size zero, function `tee_svc_copy_param()` will return
`TEE_ERROR_BAD_PARAMETERS` resulting in a panic later on.

In regard to the `TEE_PARAM_TYPE_MEMREF_{INPUT,OUTPUT,INOUT}` types, the Global Platform specification says [1]:

> params[i].memref.size SHALL be initialized with a memory buffer that is accessible with the access rights described in the type. The buffer can be NULL, in which case size SHALL be set to 0 .

Note that the specification only specifies that `NULL` implies a size of zero. The reciprocal: _"a size of zero implies `NULL`"_ is undefined.

It is even more confusing considering that non-`NULL` buffer with a size of zero are explicitly allowed in the client API [2]:

> This design allows a non-NULL buffer with a size of 0 bytes to allow trivial integration with any implementations of the C library malloc, in which is valid to allocate a zero byte buffer and receive a non-NULL pointer which may not be de-referenced in return.

This could become a pitfall if a TA forwards a memref received from the REE to another TA.

Allow the TAs to pass non-NULL memref of size zero to other TAs through `TEE_InvokeTACommand()` by changing the non-NULL pointer into a `NULL` one in such a case.

[1] TEE Internal Core API Specification – Public Release v1.3.1,
    §4.9.4 "Operation Parameters in the Internal Client API"
    Table 4-15: "Interpretation of params[i] on Entry to Internal Client API"

[2] TEE Client API Specification v1.0,
    §4.5.4 TEEC_RegisterSharedMemory,
    paragraph "Implementers' Notes"

### DIFF
--- a/core/tee/tee_svc.c
+++ b/core/tee/tee_svc.c
@@ -700,11 +700,13 @@ static TEE_Result tee_svc_copy_param(struct ts_session *sess,
 		case TEE_PARAM_TYPE_MEMREF_INOUT:
 			va = (void *)param->u[n].mem.offs;
 			s = param->u[n].mem.size;
-			if (!va) {
-				if (s)
-					return TEE_ERROR_BAD_PARAMETERS;
+			if (!s) {
+				param->u[n].mem.mobj = NULL;
 				break;
 			}
+			if (!va)
+				return TEE_ERROR_BAD_PARAMETERS;
+
 			/* uTA cannot expose its private memory */
 			if (vm_buf_is_inside_um_private(&utc->uctx, va, s))
 				return TEE_ERROR_BAD_PARAMETERS;


jenswi-linaro:
I do not often have to say this, but the commit message is a bit long with all the citations from the specs. It takes a while to see what you're trying to achieve.

Just describing the what and why should be enough in this case. It's OK with a few references, but save those after the what and why.

Please add a test case for this. A new case 1041 if you can't find a relevant one to extend.

vincent-mailhol:
> I do not often have to say this, but the commit message is a bit long with all the citations from the specs. It takes a while to see what you're trying to achieve.
> 
> Just describing the what and why should be enough in this case. It's OK with a few references, but save those after the what and why.

I amended the [commit description](https://github.com/OP-TEE/optee_os/pull/6405/commits/eb01587d8fbd29962959497942cb7d30bfce2cf6) (I left the PR text untouched). Please let me know if this addresses your concerns.

> Please add a test case for this. A new case 1041 if you can't find a relevant one to extend.

It took me a bit of time to figure out that you were referring to another code base. I never used `optee_test` before. Now I am trying to figure out how to compile it (c.f. https://github.com/OP-TEE/optee_test/pull/702). I can write the test, but please be patient.

jforissier:
> It took me a bit of time to figure out that you were referring to another code base. I never used `optee_test` before. Now I am trying to figure out how to compile it (c.f. [OP-TEE/optee_test#702](https://github.com/OP-TEE/optee_test/pull/702)). I can write the test, but please be patient.

You can easily create a full development environment as explained here: https://optee.readthedocs.io/en/latest/building/devices/qemu.html#qemu-v8

etienne-lms (commented):
I agree with this change.

As Jens said, a test in optee_test would be nice (for maintenance purpose) and commit message could be simplified. Suggestion:
```
core: tee_svc.c: allow to pass non-NULL memref of size 0

Allows TAs to pass non-NULL memref of size zero to other TAs by
changing the non-NULL pointer into a NULL one in such a case.
GP TEE Internal Core API does not forbid such memref parameter
whereas the previous implementation generated a
TEE_ERROR_BAD_PARAMETERS error code when converting such memref
buffer pointer into a physical memory address.

This change is specifically needed to allow a TA to forward a REE
client memref for which GP TEE Client API explicitly allows such
non-NULL address zero sized memref. It also makes the TA implementation
more flexible when dealing with its own memref.
```

Note: this change impacts both `TEE_OpenTASession()` and
`TEE_InvokeTACommand()`.

vincent-mailhol:
> You can easily create a full development environment as explained here: https://optee.readthedocs.io/en/latest/building/devices/qemu.html#qemu-v8

@jforissier Thanks for the suggestion. I am aware of this. I could use it but I am not familiar with repo which makes it hard for me to modify (and I have no plan to invest time into studying repo). I am using the different modules my way. It takes a bit more time to onboard a new piece to the project, but I enjoy the flexibility. I solved my last issues, so now I am fine.

> As Jens said, a test in optee_test would be nice

@etienne-lms Yes, as I commented, I will do it, but be patient. I did not plan for this extra step when submitting my patch, and, as always, time being a scarce resource, I prefer not to commit on when I will do it.

Thanks for the rephrasing. I like it. I did small changes:

- Use imperative tone: `s/Allows/Allow/g`. This is part of the [Linux kernel guidelines on how to write patches](https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes).
- Add the footnote reference, but this time without the full quotation.

jforissier:
> > You can easily create a full development environment as explained here: https://optee.readthedocs.io/en/latest/building/devices/qemu.html#qemu-v8
> 
> @jforissier Thanks for the suggestion. I am aware of this. I could use it but I am not familiar with repo which makes it hard for me to modify (and I have no plan to invest time into studying repo). I am using the different modules my way. It takes a bit more time to onboard a new piece to the project, but I enjoy the flexibility. I solved my last issues, so now I am fine.

Good to know :+1: Actually you don't have to know anything about `repo`. Just use it to clone everything, but then you can use `git` in each project (and `make` in `build/` to build & run of course) and not care about any repo workflow anymore. That's what I do.

jbech-linaro:
> > > You can easily create a full development environment as explained here: https://optee.readthedocs.io/en/latest/building/devices/qemu.html#qemu-v8
> > 
> > 
> > @jforissier Thanks for the suggestion. I am aware of this. I could use it but I am not familiar with repo which makes it hard for me to modify (and I have no plan to invest time into studying repo). I am using the different modules my way. It takes a bit more time to onboard a new piece to the project, but I enjoy the flexibility. I solved my last issues, so now I am fine.
> 
> Good to know 👍 Actually you don't have to know anything about `repo`. Just use it to clone everything, but then you can use `git` in each project (and `make` in `build/` to build & run of course) and not care about any repo workflow anymore. That's what I do.

Yes, very few steps, like [this](https://optee.readthedocs.io/en/latest/faq/faq.html#faq-try-optee). Although I'd just use [the QEMU v8 manifest](https://optee.readthedocs.io/en/latest/building/devices/qemu.html#id3) nowadays instead of default v7.

jenswi-linaro:
>  I did not plan for this extra step when submitting my patch, and, as always, time being a scarce resource, I prefer not to commit on when I will do it.

We're usually not merging untested code. It's also easier to review code if there's an example that exercises it and shows that it works.

vincent-mailhol:
> > I did not plan for this extra step when submitting my patch, and, as always, time being a scarce resource, I prefer not to commit on when I will do it.
> 
> We're usually not merging untested code. It's also easier to review code if there's an example that exercises it and shows that it works.

Refer to my sentence just before the one you quoted:

> Yes, as I commented, I will do it, but be patient.

I am doing this on my free time, so please forgive if my progress is slow.

If you think this is taking too long, maybe I can close the pull request and reopen it when ready?

vincent-mailhol:
@jenswi-linaro and @etienne-lms, as promised here is a test: https://github.com/OP-TEE/optee_test/pull/708. Of course, merging this PR is a prerequisite for the test to success.

etienne-lms (commented):
Hi Vincent, thanks for the optee_test counter part. It will help this P-R to land.

`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for this change.

~~@jforissier, could you enable CI test for this P-R and for the related optee_test P-R https://github.com/OP-TEE/optee_test/pull/708?~~

PR #6409, opened 2023-10-26
core: arm: rpc_alloc: remove size limit for kernel payload

jenswi-linaro:
Removes the size limit of 1 page imposed in
thread_rpc_alloc_kernel_payload(). The purpose of this limit was to error out early since the kernel doesn't supply a list of physical pages and the source of the error is not obvious at first glance. This is now about to change so remove the limit since the kernel now may supply the needed list of physical pages.

Link to corresponding kernel patch https://lore.kernel.org/lkml/20231026171650.515275-1-jens.wiklander@linaro.org/

### DIFF
--- a/core/arch/arm/kernel/thread_optee_smc.c
+++ b/core/arch/arm/kernel/thread_optee_smc.c
@@ -681,14 +681,6 @@ struct mobj *thread_rpc_alloc_payload(size_t size)
 
 struct mobj *thread_rpc_alloc_kernel_payload(size_t size)
 {
-	/*
-	 * Error out early since kernel private dynamic shared memory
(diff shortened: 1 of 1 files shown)


jforissier (commented):

jenswi-linaro:
Tag applied. We may want to wait a little with merging this until we see how the kernel patch is received.

jforissier:
> Tag applied. We may want to wait a little with merging this until we see how the kernel patch is received.

Sure.

etienne-lms (commented):

jenswi-linaro:
Tagg applied and rebased on master.

jenswi-linaro:
The corresponding kernel patches have been reviewed. I believe this is ready to be merged now.

PR #6416, opened 2023-10-31
libutee: TEE_AllocateOperation(): maxKeySize of digests may take any value

vincent-mailhol:
Commit https://github.com/OP-TEE/optee_os/commit/cf5c060cec76 introduced a check to enforce the `maxKeySize` parameter of digest operations to always be zero. This is a violation of the Global Platform specification.

Revert commit https://github.com/OP-TEE/optee_os/commit/cf5c060cec76 to allow `maxKeySize` to take any value for digest operations.

### DIFF
--- a/lib/libutee/tee_api_operations.c
+++ b/lib/libutee/tee_api_operations.c
@@ -132,7 +132,7 @@ TEE_Result TEE_AllocateOperation(TEE_OperationHandle *operation,
 		break;
 	}
 
-	/* Check algorithm mode (and maxKeySize for digests) */
+	/* Check algorithm mode */
 	switch (algorithm) {
 	case TEE_ALG_AES_CTS:
 	case TEE_ALG_AES_XTS:
@@ -276,8 +276,6 @@ TEE_Result TEE_AllocateOperation(TEE_OperationHandle *operation,
 	case TEE_ALG_SHAKE256:
 	case TEE_ALG_SM3:
 		if (mode != TEE_MODE_DIGEST)
-			return TEE_ERROR_NOT_SUPPORTED;
-		if (maxKeySize)
 			return TEE_ERROR_NOT_SUPPORTED;
 		/* v1.1: flags always set for digest operations */
 		handle_state |= TEE_HANDLE_FLAG_KEY_SET;


vincent-mailhol:
**Additional notes:** the clarification that `maxKeySize` can take any value for digest operations was added in TEE Internal Core API Specification v1.1.2.

etienne-lms (commented):
Sentence "_the maxKeySize parameter may have any value._" indeed means it is permitted to pass any value in`maxKeySize` (for digest operations). However i think that despite `maxKeySize` can have any value, the fix should force this value to 0 for digests, so that `TEE_GetOperationValue()` complies with the spec that states "_**keySize**: For an operation that makes no use of keys, 0._".

vincent-mailhol:
> despite `maxKeySize` can have any value, the fix should force this value to 0 for digests, so that `TEE_GetOperationValue()` complies with the spec that states _"**keySize**: For an operation that makes no use of keys, 0."_

In the documentation of `TEE_GetOperationValue()`, I read:

> `algorithm`, `mode`, `maxKeySize`: The parameters passed to the function `TEE_AllocateOperation`

So the specification requires that `maxKeySize` stays as-is when retuned by `TEE_GetOperationValue()`. It would be incorrect to force that value to zero.

But you raised an interesting point. I will double check that `TEE_AllocateOperation` correctly returns a `keySize` of zero even if `maxKeySize` is set to some random values (it is too late in my time zone to run more test, please wait tomorrow for the results).

vincent-mailhol:
I confirm the behaviour for `keySize`. In `TEE_AllocateOperation()` the `TEE_OperationHandle` is zero initialized: https://github.com/OP-TEE/optee_os/blob/355bf0fba1dd525d922911b9e91037b536886ddb/lib/libutee/tee_api_operations.c#L313 Which means that the `op->info.keySize` is also zeroed. And that value stays untouched if the operation is a digest.

So indeed, for a digest operation, `TEE_GetOperationValue()` returns:

- the same `maxKeySize` as what was passed by `TEE_AllocateOperation()`
- a `keySize` of zero (regardless of the value of `maxKeySize`)

I do not see any gap between my patch and the specification.

etienne-lms:
Right, `maxKeySize` is not used to fill `keySize`. That said, I think storing a non-zero value in `maxKeySize` is not very consistent. I would still be in favor to allow caller to use any value but set it to zero in TEE core (FWIW).

vincent-mailhol:
> Right, `maxKeySize` is not used to fill `keySize`. That said, I think storing a non-zero value in `maxKeySize` is not very consistent. I would still be in favor to allow caller to use any value but set it to zero in TEE core (FWIW).

@etienne-lms, sorry, I do not understand your point. The specification asks that `TEE_GetOperationValue()` returns the `maxKeySize` which was passed to `TEE_AllocateOperation()`. If we set `maxKeySize` to zero, then this requirement can not be fulfilled anymore, right?

If above requirement did not exist, I would have agreed with you. But the specification tells us to do it one way and I will not argue whether or not the specification is consistent.

etienne-lms:
Right, I missed the spec line stating _"algorithm , mode , maxKeySize : The parameters passed to the function
TEE_AllocateOperation"_.
Fine as is. Sorry for the noise.

etienne-lms:
@jforissier, could you enable CI for this P-R?

vincent-mailhol:
@etienne-lms issue solved then! Thanks for the review, I added you tag.

jforissier:
I tested this change against the GlobalPlatform TEE Initial Configuration Test Suite v2_0_0_2-2017_06_09, and unfortunately it causes 6 tests to fail. For example:

> Test: Invoke_Crypto_AllocateOperation_TEE_ALG_MD5_size_not_appropriate (3b-4d-15)  
>  Description: If the mode is not compatible with the algorithm or key size or if the  algorithm is not one of the listed algorithms or if maxKeySize is not  appropriate for the algorithm.

(the other failures are similar, just with different algorithms).

Since as mentioned [above](https://github.com/OP-TEE/optee_os/pull/6416#issuecomment-1787268163) the spec was clarified to explicitly allow what the GP test suite rejects, these tests should be excluded. I will check how to do that and open a PR in `optee_test`.

jforissier:
> Since as mentioned [above](https://github.com/OP-TEE/optee_os/pull/6416#issuecomment-1787268163) the spec was clarified to explicitly allow what the GP test suite rejects, these tests should be excluded. I will check how to do that and open a PR in `optee_test`.

https://github.com/OP-TEE/optee_test/pull/704

jenswi-linaro:

jforissier:
@vincent-mailhol thanks!

PR #6422, opened 2023-11-02
drivers: crypto: caam: use job ring 3 on i.mx8dxlevk

clementfaure:
Use the job ring #3 on i.mx8dxl to avoid resource conflict with other software stacks.


jforissier:

clementfaure:
Tag applied

PR #6412, opened 2023-10-30
Introduce STM32MP25x platforms minimal support

GseoC:
This P-R aims to bring minimal platform support for STM32MP25x 64 bits platforms in OP-TEE.
You'll find a resume of the features in the first commit on this series.

This P-R cleans some code in GPIO and UART driver as well that blocked the compilation of another platform using STM32 drivers.

A weak platform-specific plat_panic() is added to add platform-specific behavior in case of panic().

@etienne-lms As discussed in private, I put you as maintainer of the platform as well as myself.

@etienne-lms @tbourgoi, can you please review this P-R?

PR #6394, opened 2023-10-23
FF-A Notifications

jenswi-linaro:

PR #6423, opened 2023-11-02
Core/Plat: Allow platform specific panic sequence

GseoC:
When panicking, there may be the need to stop the other core or have a specific sequence.

Define a weak plat_panic() function to do such operations and add an implementation for STM32MP1x platforms.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/main.c
+++ b/core/arch/arm/plat-stm32mp1/main.c
@@ -19,6 +19,7 @@
 #include <io.h>
 #include <kernel/boot.h>
 #include <kernel/dt.h>
+#include <kernel/interrupt.h>
 #include <kernel/misc.h>
 #include <kernel/panic.h>
 #include <kernel/spinlock.h>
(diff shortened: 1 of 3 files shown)


core/arch/arm/plat-stm32mp1/main.c
```diff
@@ -584,3 +585,128 @@ early_init_late(init_debug);
...
+#define ARM_CNTXCTL_IMASK	BIT(1)
+
+static void stm32mp_mask_timer(void)
+{
+	/* Mask timer interrupts */
+	write_cntp_ctl(read_cntp_ctl() | ARM_CNTXCTL_IMASK);
+	write_cntv_ctl(read_cntv_ctl() | ARM_CNTXCTL_IMASK);
+}
+
+/* SGI9 (secure SGI 1) informs targeted CPU it shall reset */
```
GseoC, line 656: Reported in https://github.com/OP-TEE/optee_os/pull/6412#discussion_r1377198941_

GseoC: Done

etienne-lms (commented):
Could you remove the Change-id tag from 1st commit message?

GseoC:
> Could you remove the Change-id tag from 1st commit message?

Yes, my bad

core/kernel/panic.c
```diff
@@ -29,11 +28,19 @@ void __do_panic(const char *file __maybe_unused,
 			 func ? "<" : "", func ? func : "", func ? ">" : "");
 
 	print_kernel_stack();
-	/* abort current execution */
-	while (1)
+	plat_panic();
+
+	EMSG("platform failed to abort execution");
```
jforissier, line 33: Uppercase P[latform]

jforissier (commented):
For "core: panic: platform specific panic sequence":

One small comment, then:

core/arch/arm/plat-stm32mp1/main.c
```diff
@@ -584,3 +585,128 @@ early_init_late(init_debug);
...
+	bool display = false;
+	uint32_t *reg = NULL;
+	size_t i = U(0);
+
+	cpu_spin_lock(&lock);
+
+	if (panicking)
+		display = true;
+
+	if (!display || !sm_nsec_ctx)
```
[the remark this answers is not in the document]
GseoC: Probably not

core/arch/arm/plat-stm32mp1/main.c
```diff
@@ -584,3 +585,128 @@ early_init_late(init_debug);
...
+void __noreturn plat_panic(void)
+{
+	struct itr_chip *itr_chip = interrupt_get_main_chip();
+
+	stm32mp_mask_timer();
+
+	if (CFG_TEE_CORE_NB_CORE > 1) {
+		uint32_t target_mask = 0;
+
+		if (get_core_pos() == 0)
```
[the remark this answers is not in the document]
GseoC: I found it clearer to explicit the core number but ok

GseoC:
Comments adressed for MP1 implementation and @jforissier 's tag applied

jenswi-linaro:
For "core: panic: platform specific panic sequence" please apply:

jenswi-linaro:
I'm having second thoughts about this. I'm not so keen on all these weak functions acting like platform hooks, especially a platform hook like this that solves the problem for one platform, but by default leaves all other platforms in the dark.

I believe this would make sense as a generic option instead. With something like "core: itr_raise_sgi() updates" from https://github.com/OP-TEE/optee_os/pull/6394, but with an `ITR_CPU_MASK_TO_OTHER_CPUS` also it should be straight forward.

etienne-lms:
Would you prefer platforms to statically or dynamically register (or not) a platform panic handler, instead of a weak function? By statically I meant something like how `register_ddr()` is implemented.

Or an even more generic solution based on a config switches? Something like a pair of switch like `CFG_CORE_HALT_CORES_ON_PANIC=y|n` + `CFG_CORE_HALT_CORES_ON_PANIC_SGI=<number>`. Note SGI is an Arm specific naming.

`ITR_CPU_MASK_TO_OTHER_CPUS` would indeed be nice.

jenswi-linaro:
> Would you prefer platforms to statically or dynamically register (or not) a platform panic handler, instead of a weak function? By statically I meant something like how `register_ddr()` is implemented.

No, that's still a platform solution.

> 
> Or an even more generic solution based on a config switches? Something like a pair of switch like `CFG_CORE_HALT_CORES_ON_PANIC=y|n` + `CFG_CORE_HALT_CORES_ON_PANIC_SGI=<number>`. Note SGI is an Arm specific naming.

Yes, something like this is what I had in mind.

> 
> `ITR_CPU_MASK_TO_OTHER_CPUS` would indeed be nice.

OK, I have a patch for this, I'll create a PR.

jenswi-linaro:
Feel free to cherry-pick https://github.com/jenswi-linaro/optee_os/tree/sgi_update and use that. I'm afraid that I've only been able to test `ITR_CPU_MASK_TO_THIS_CPU` in earlier versions of this patch and `ITR_CPU_MASK_TO_OTHER_CPUS` is untested.

GseoC:
> Feel free to cherry-pick https://github.com/jenswi-linaro/optee_os/tree/sgi_update and use that. I'm afraid that I've only been able to test `ITR_CPU_MASK_TO_THIS_CPU` in earlier versions of this patch and `ITR_CPU_MASK_TO_OTHER_CPUS` is untested.

Thank you,

I'll will try something generic and test it on our platforms.
As it will completely rework the way of doing it, I'll create a new P-R when ready.

GseoC:
Follow-up: [https://github.com/OP-TEE/optee_os/pull/6440](url)

GseoC:
Closed, accepted version: [#6440](https://github.com/OP-TEE/optee_os/pull/6440)

PR #6424, opened 2023-11-02
mk/aosp_optee.mk: Allow building OPTEE_BIN with early TAs in one shot

taddk:
This change adds OPTEE_TA_DEV_KIT_MK target to replace OPTEE_BIN target in TA build process. It helps to build OPTEE_BIN with early TAs in one shot since the cyclic dependency between TAs and OPTEE_BIN is removed. It also allows to build TA with a preferred TA_TARGET that is not included in the CFG_USER_TA_TARGETS when building OPTEE_BIN. An AOSP common module for TAs in stripped elf format is added to be referred as early TA files when building OPTEE_BIN. For example, we can setup dependency and use the argument below to build OPTEE_BIN with an early TA.

EARLY_TA_PATHS=$(abspath $(PRODUCT_OUT))/unsigned/<ta_uuid>.ta.unsigned

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -32,6 +32,11 @@ basis (R:).
 
 ----------
 
+AOSP build
+R:	Tadd Kao <tadd.kao@mediatek.com> [@taddk]
+S:	Maintained
+F:	mk/aosp_optee.mk
+
 ARM Foundation FVP
 R:	Jens Wiklander <jens.wiklander@linaro.org> [@jenswi-linaro]
 S:	Maintained
--- a/mk/aosp_optee.mk
+++ b/mk/aosp_optee.mk
@@ -60,7 +60,8 @@ TA_TARGET := $(local_optee_ta_target)
 else
 TA_TARGET := $(OPTEE_TA_TARGETS)
 endif
-TA_DEV_KIT_DIR := $(ABS_OPTEE_OS_OUT_DIR)/export-$(TA_TARGET)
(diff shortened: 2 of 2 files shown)


jforissier (commented):
Hi @taddk,

This looks OK to me but TBH I am not very familiar with the AOSP/Android build. As mentioned in the [documentation](https://optee.readthedocs.io/en/latest/building/aosp/aosp.html), we (Linaro) are not doing AOSP builds anymore, and we don't have a maintainer for these makefiles. So let me take this opportunity to ask: would you like to add yourself as a maintainer for the AOSP build? It would not be a big effort on your side, it is essentially about being identified as a knowledgeable person on this topic -- someone who can occasionally run tests and provide feedback on pull requests etc.
If so, please add an entry "AOSP build" to MAINTAINERS.

Anyways:

MAINTAINERS
```diff
@@ -32,6 +32,11 @@ basis (R:).
 
 ----------
 
+AOSP build
+R:	Tadd Kao <tadd.kao@mediatek.com> [@taddk]
+S:	Maintained
+F:	mk/aosp_optee.mk
+
```
jforissier, line 39: That's fine but I would prefer a separate commit for this (you can add my 
`Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`)

taddk: OK.

taddk:
Hi @jforissier,

I can help to maintain the mk/aosp_optee.mk. The MAINTAINERS file is updated in a separated commit. If there is any suggestion, just let me know.

jforissier:
Thanks @taddk, much appreciated!

PR #6425, opened 2023-11-03
plat-stm32mp1: enable async notif on stm32mp13

etienne-lms:
Enables async notif using GIC PPI 15 as non-secure interrupt notifier for STM32MP13 variants.

This change depends on https://github.com/linaro-swg/linux/pull/116 and https://github.com/linaro-swg/linux/pull/115. It should not be merged before these 2 are merged unless what STM32MP13 kernel from linaro-swg/linux will fail to boot. This change is fine when using mainline kernel that already merges the changes from these 2 commits.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -106,6 +106,8 @@ $(call force,CFG_REGULATOR_FIXED,y)
 
 ifeq ($(CFG_STM32MP13),y)
 $(call force,CFG_BOOT_SECONDARY_REQUEST,n)
+$(call force,CFG_CORE_ASYNC_NOTIF,y)
+$(call force,CFG_CORE_ASYNC_NOTIF_GIC_INTID,31)
 $(call force,CFG_CORE_RESERVED_SHM,n)
 $(call force,CFG_DRIVERS_CLK_FIXED,y)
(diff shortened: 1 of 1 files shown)


GseoC:
Optional: maybe specify that GIC PPI15 is GIC ID31 either in commit message or as a comment in the conf.mk file.

jenswi-linaro:

b49020:
Good to see async notifications enabled on a real platform. It will be further nice to see its use-cases (IOW, a real secure interrupt bottom half processing). Do you have any downstream user?

BTW, feel free to add:

etienne-lms:
Tags applied. Thanks.

> Do you have any downstream user?

There is one available on the 3.19.0 based downstream repo related to a wakeup interrupt source. When CPUs are active, a SoC PWR interrupt (GIC SPI secure line) can be used to notify a PMIC event for which an I2C read transfer is needed. You can find the related [PWR driver piece of code](https://github.com/STMicroelectronics/optee_os/blob/3.19.0-stm32mp/core/arch/arm/plat-stm32mp1/drivers/stm32mp1_pwr_irq.c#L137) and the [PMIC consumer handler](https://github.com/STMicroelectronics/optee_os/blob/3.19.0-stm32mp/core/arch/arm/plat-stm32mp1/drivers/stm32mp1_pmic.c#L477-L507). I hope we'll upstream this soon, based on mainline new interrupt controller framework.

PR #6429, opened 2023-11-03
core: kernel: interrupt: fix inline description

etienne-lms:
Fixes the inline description comment for type get_of_device_func and API function dt_driver_device_from_parent().

Fixes: b357d34fe91f ("core: dt_driver: swap TEE_result and retrieved device reference")


jforissier (commented):

PR #6430, opened 2023-11-03
mk: config.mk: describe CFG_WITH_STATS

etienne-lms:
Adds a description for CFG_WITH_STATS and explicitly state that the config switch is default disabled.


mk/config.mk
```diff
@@ -543,6 +543,10 @@ CFG_DRIVERS_CLK_EARLY_PROBE ?= n
 $(call force,CFG_DRIVERS_DT_RECURSIVE_PROBE,n,Mandated by CFG_DT_DRIVER_EMBEDDED_TEST)
 endif
 
+# CFG_WITH_STATS when enabled embeds PTA stats service to allow non-secure
```
jforissier, line 546: Perhaps spell out statistics instead of stats here

jforissier (commented):

mk/config.mk
etienne-lms: Ok, i'll do.

etienne-lms (commented):
Actually, I created this P-R rather to ask a question: should we default init `CFG_WITH_STATS` to `CFG_ENABLE_EMBEDDED_TESTS` rather than `n`? 
That would prevent to have `CFG_WITH_STATS?=y` in platforms _conf.mk_ files that are there only for debug and test purpose IMHO.

jenswi-linaro:
> Actually, I created this P-R rather to ask a question: should we default init `CFG_WITH_STATS` to `CFG_ENABLE_EMBEDDED_TESTS` rather than `n`?
> That would prevent to have `CFG_WITH_STATS?=y` in platforms _conf.mk_ files that are there only for debug and test purpose IMHO.

It seems rather arbitrary to assign the default value from `CFG_ENABLE_EMBEDDED_TESTS`, I'd prefer having it configured independently. One option could be for `CFG_WARN_INSECURE` to refuse `CFG_WITH_STATS` since it could be a security problem.

etienne-lms:
I think there are several config switches that `CFG_WARN_INSECURE=n` should enforce. I'd rather handle that is a specific P-R. For this P-R, i'll stay on `CFG_WITH_STATS ?= n`.

jenswi-linaro (commented):

etienne-lms:
review tags applied

PR #6432, opened 2023-11-06
riscv: virt: Update the configurations

gagachang:
This commit updates the configurations for QEMU RISC-V virtual platform:
1. Enable CFG_RISCV_S_MODE and CFG_RISCV_SBI to run OP-TEE on S-mode and utilize SBI to communicate with M-mode firmware.
2. Do not force CFG_TEE_CORE_NB_CORE and CFG_NUM_THREADS to be 1, since we may run SMP system.
3. Disable CFG_BOOT_SYNC_CPU and CFG_BOOT_SECONDARY_REQUEST, since the boot sequence is controlled by M-mode firmware.
4. Enable CFG_WITH_STAT to build PTA for debug and statistics information.
5. Enable CFG_DT to parse the external DTB passed by M-mode firmware.

### DIFF
--- a/core/arch/riscv/plat-virt/conf.mk
+++ b/core/arch/riscv/plat-virt/conf.mk
@@ -13,8 +13,8 @@ $(call force,CFG_WITH_STACK_CANARIES,n)
 $(call force,CFG_CORE_SANITIZE_KADDRESS,n)
 
 # Hart-related flags
-$(call force,CFG_TEE_CORE_NB_CORE,1)
-$(call force,CFG_NUM_THREADS,1)
+CFG_TEE_CORE_NB_CORE ?= 1
+CFG_NUM_THREADS ?= 1
 $(call force,CFG_BOOT_SYNC_CPU,y)
 
 # RISC-V-specific flags
--- a/core/arch/riscv/plat-virt/conf.mk
+++ b/core/arch/riscv/plat-virt/conf.mk
@@ -4,6 +4,8 @@ $(call force,CFG_CORE_LARGE_PHYS_ADDR,y)
 $(call force,CFG_TEE_CORE_DEBUG,n)
 $(call force,CFG_CORE_DYN_SHM,n)
 
+CFG_DT ?= y
+
 # Crypto flags
 $(call force,CFG_WITH_SOFTWARE_PRNG,y)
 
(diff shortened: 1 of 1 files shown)


gagachang:
Hello @maroueneboubakri 
You may want to take a look at this PR

jforissier:
> This commit updates the configurations for QEMU RISC-V virtual platform:
>
>   1. Enable CFG_RISCV_S_MODE and CFG_RISCV_SBI to run OP-TEE on S-mode and utilize SBI to communicate with M-mode firmware.
>   2. Do not force CFG_TEE_CORE_NB_CORE and CFG_NUM_THREADS to be 1, since we may run SMP system.
>   3. Disable CFG_BOOT_SYNC_CPU and CFG_BOOT_SECONDARY_REQUEST, since the boot sequence is controlled by M-mode firmware.
>   4. Enable CFG_WITH_STAT to build PTA for debug and statistics information.
>   5. Enable CFG_DT to parse the external DTB passed by M-mode firmware.

If all these changes are independent (which I believe they are), it would be better to make one commit for each.

gagachang:
> If all these changes are independent (which I believe they are), it would be better to make one commit for each.
OK I will.

jforissier (commented):
The "if' in the first commit description is somewhat incorrect, I suggest "we enable CFG_RISCV_SBI **so that** OP-TEE utilizes SBI to communicate with other OS". In any case, LGTM. For all commits:

gagachang:
> The "if' in the first commit description is somewhat incorrect, I suggest "we enable CFG_RISCV_SBI **so that** OP-TEE utilizes SBI to communicate with other OS". In any case, LGTM. For all commits:
> 

Apply the commit message and tag. Thanks for suggestion!

maroueneboubakri:
Hi @gagachang,

Thanks for these patches, is there any reason why `CFG_WITH_STATS ` defaults to `y` ? This should be enabled to allow stats/debug info. Also this will increase the core footprint. If I need it I would enable the flag using command line.

Best
Maro

gagachang:
> Hi @gagachang,
> 
> Thanks for these patches, is there any reason why `CFG_WITH_STATS ` defaults to `y` ? This should be enabled to allow stats/debug info. Also this will increase the core footprint. If I need it I would enable the flag using command line.
> 
> Best Maro

OK I can remove that commit. We can enable it by command line.
Thanks! Would you provide the tags for review?

jforissier:
Ping @maroueneboubakri would you like to provide any review tag here? Thanks.

jforissier:
> Hi @jforissier It seems something broken in Hafnium checking

We have just updated Hafnium to v2.10 (https://github.com/OP-TEE/manifest/pull/257) so the error is likely related. @jenswi-linaro is investigating this timeout in another PR (https://github.com/OP-TEE/optee_os/pull/6534#issuecomment-1845604426).

PR #6434, opened 2023-11-06
plat-stm32mp1: default disable CFG_STM32MP1_SCMI_SHM_SYSRAM

etienne-lms:
Default disable CFG_STM32MP1_SCMI_SHM_SYSRAM to sync with mainline Linux kernel that uses OP-TEE shared memory for SCMI communication on STM32MP13 platform instead a piece of SRAM [1].

Link: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f0f0682c384d81bf25e6f8b23971fb8f69122f72


GseoC:
Thanks, maybe we can go a step further its support completely? What are your thoughts on that?
Maybe asking other MP1x based platforms if that's OK? On next TF-A tag, the OP-TEE shared memory support will be deprecated for our platforms (MP15 as well)

etienne-lms:
Let's first default disable this and see later if we fully remove the support.

GseoC:

etienne-lms:
review tag applied

PR #6438, opened 2023-11-07
drivers: crypto: caam: fix job ring interruption number

clementfaure:
The job ring interruption number is 356 for job ring 3.

Fixes: b21f12209 ("drivers: crypto: caam: use job ring 3 on i.mx8dxlevk")


jforissier (commented):

clementfaure:
Tag applied

PR #6440, opened 2023-11-07
Core: Halt other cores on panic in a multi-core system

GseoC:
Following: [https://github.com/OP-TEE/optee_os/pull/6423](url)

This P-R implements a generic sequence to halt all the CPUs (or Processing Elements) of a cluster when one panics using a Software Generated Interrupt (SGI). The interrupt handler is registered when other CPUs are setup init_secondary_helper().

To activate this feature: set CFG_CORE_HALT_CORES_ON_PANIC to 'y' and the desired SGI CFG_CORE_HALT_CORES_ON_PANIC_SGI. 

Default values:
CFG_CORE_HALT_CORES_ON_PANIC ?= n
CFG_CORE_HALT_CORES_ON_PANIC_SGI ?= 9 

Tested on STM32MP157C-DK2 platform that is ARM32 with GICV2.

@jenswi-linaro , I've corrected  gic_it_raise_sgi()'s prototype, cpu_mask was uint8_t and was '&' with BIT(30)/BIT(31)


etienne-lms (commented):
Comments for commit "core: interrupt_raise_sgi() updates":
Implementation looks good to me. Few nitpicking comments.
In commit message: s/this new flag/these new flags/.

core/arch/arm/kernel/boot.c
```diff
@@ -1221,8 +1221,49 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
...
+static void mask_timer(void)
+{
+	/* Mask timer interrupts */
+	write_cntp_ctl(read_cntp_ctl() | ARM_CNTXCTL_IMASK);
+	write_cntv_ctl(read_cntv_ctl() | ARM_CNTXCTL_IMASK);
+}
+#endif /* ARM32 */
+
+static enum itr_return
+	multi_core_panic_it_handler(struct itr_handler *hdl  __unused)
```
etienne-lms, line 1238: remove this tab
I think this function needs a `__noreturn` attribute

core/arch/arm/kernel/boot.c
```diff
@@ -1241,6 +1282,14 @@ static void init_secondary_helper(unsigned long nsec_entry)
 	init_vfp_sec();
 	init_vfp_nsec();
 
+#ifdef CFG_CORE_HALT_CORES_ON_PANIC
+	if (interrupt_add_handler_with_chip(itr_chip,
```
etienne-lms, line 1286: I think you can use `interrupt_get_main_chip()` as first argument in the 2 function calls.
I think its would be better than a local varaible under an `#ifdef` directive.

core/arch/arm/kernel/boot.c
```diff
@@ -1241,6 +1282,14 @@ static void init_secondary_helper(unsigned long nsec_entry)
...
+					    &multi_core_panic_handler))
+		panic();
+
+	interrupt_enable(itr_chip, multi_core_panic_handler.it);
```
etienne-lms, line 1290: Should be called only once in case there are more than 2 cores.

etienne-lms (commented):
Comments for commit "core: interrupt: halt other cores when one is panicking":

core/arch/arm/kernel/boot.c
```diff
@@ -1221,8 +1221,49 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
...
 }
 
+#ifdef CFG_CORE_HALT_CORES_ON_PANIC
+
+#ifdef ARM32
+#define ARM_CNTXCTL_IMASK	BIT(1)
+
+static void mask_timer(void)
+{
+	/* Mask timer interrupts */
```
etienne-lms, line 1231: can remove this inline comment. Redundant with the function name IMO.

core/include/kernel/interrupt.h
```diff
@@ -266,10 +271,12 @@ static inline void interrupt_raise_pi(struct itr_chip *chip, size_t itr_num)
  * interrupt_raise_sgi() - Raise a software generiated interrupt of a controller
  * @chip	Interrupt controller
  * @itr_num	Interrupt number to raise
- * @cpu_mask	Mask of the CPUs targeted by the interrupt
+ * @cpu_mask:	A bitfield of CPUs to forward the interrupt too, unless
```
etienne-lms, line 274: s/too/to/

mk/config.mk
```diff
@@ -1063,3 +1063,11 @@ CFG_HMAC_64_1024_RANGE ?= n
...
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# Make other cores halt on panic() in a multi core setup. An SGI will be
+# triggered to notify all other CPUs in the same cluster that a panic() is
+# occurring. The SGI used to notify other cores is specified in:
+# CFG_CORE_HALT_CORES_ON_PANIC_SGI
+# Note: Other clusters of CPU won't be notified
+CFG_CORE_HALT_CORES_ON_PANIC ?= n
+CFG_CORE_HALT_CORES_ON_PANIC_SGI ?= 9
```
etienne-lms, line 1073: I suggest to rename to `CFG_HALT_CORES_ON_PANIC` and `CFG_HALT_CORES_ON_PANIC_SGI`.

jenswi-linaro: When should we use CFG_CORE_XXX instead of CFG_XXX?

core/drivers/gic.c
```diff
@@ -390,33 +402,59 @@ static void gic_it_set_pending(struct gic_data *gd, size_t it)
...
-	uint64_t mask_aff1 = (mpidr & MPIDR_AFF1_MASK) >> MPIDR_AFF1_SHIFT;
-	uint64_t mask_aff2 = (mpidr & MPIDR_AFF2_MASK) >> MPIDR_AFF2_SHIFT;
-	uint64_t mask_aff3 = (mpidr & MPIDR_AFF3_MASK) >> MPIDR_AFF3_SHIFT;
-	uint64_t mask = (mask_cpu |
-			SHIFT_U64(mask_aff1, 16) |
-			SHIFT_U64(mask_id, 24)   |
-			SHIFT_U64(mask_aff2, 32) |
-			SHIFT_U64(mask_aff3, 48));
+	uint64_t mask = SHIFT_U64(mask_id, 24);
+
```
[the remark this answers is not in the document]
jenswi-linaro: That expression is quite hard to parse and understand. How about:
``` C
bool to_this_cpu = cpu_mask & ITR_CPU_MASK_TO_THIS_CPU;
bool to_other_cpus = cpu_mask & ITR_CPU_MASK_TO_OTHER_CPUS;

assert(cpu_mask && (to_this_cpu + to_other_cpus) < 2);
```
The helper variables can then be used below too.

mk/config.mk
jforissier: > When should we use CFG_CORE_XXX instead of CFG_XXX?

I would say: when it is needed to make it clear that the feature relates to the TEE core. For example `CFG_CORE_HEAP_SIZE` or `CFG_CORE_ASLR` (vs. `CFG_TA_ASLR`). Otherwise it's better to have a shorter name. By that criterion, we probably have a few `CFG_CORE_XXX` too many, but not a huge number.

core/arch/arm/kernel/boot.c
jenswi-linaro: No need to indent this line.

jforissier:
>  core: interrupt: halt other cores when one is panicking
> 
> When one core panics, send an SGI (CFG_CORE_HALT_CORES_ON_PANIC_SGI)
> notifying other core of the same cluster that they should halt if
> CFG_CORE_HALT_CORES_ON_PANIC is set.

What about the cores in other clusters? I would assume the reasons why we want to stop the other cores in the same cluster also apply to the cores in other clusters?

core/arch/arm/kernel/boot.c
jenswi-linaro: Wouldn't it make sense to put this in a boot_final() function instead? We could even move the code to `core/kernel/panic.c` if we could only resolve the masking of timer interrupts.

core/drivers/gic.c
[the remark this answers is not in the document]
GseoC: Using "assert(cpu_mask && (to_this_cpu + to_other_cpus) < 2);" ,maybe missing a check if another bit  is set along with one of those? ITR_CPU_MASK_TO_THIS_CPU and ITR_CPU_MASK_TO_OTHER_CPUS should be exclusive with the rest of the bitfield, right?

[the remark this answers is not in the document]
jenswi-linaro: It depends on how far you want to take this. Having a monster of an assert expression will not make anyone happy. It must be readable.

mk/config.mk
etienne-lms: I fully agree with all points stated by Jérôme's above comment.
I remember those old days when I was in favor of using `CFG_TEE_CORE_` prefix for all config switches related to OP-TEE core, and `CFG_TEE_TA_` for OP-TEE TA related configs. And I remember you said that `CFG_` was far enough in most case. I think you were right.

etienne-lms:
> What about the cores in other clusters? I would assume the reasons why we want to stop the other cores in the same cluster also apply to the cores in other clusters?

This limitation is due to GIC driver current implementation. I think it's fine as a first step. Multi-cluster platforms will maybe prefer to not enable `CFG_[CORE_]HALT_CORES_ON_PANIC` if they have concerns with this limitation.

core/arch/arm/kernel/boot.c
etienne-lms: +1. Having the implementation in _panic.c_ fully makes sense to me. `init_secondary_helper()` would call the feature initialization.

core/drivers/gic.c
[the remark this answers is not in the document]
etienne-lms: My 2 cents: when the assert expression is over complex, it can be handy to implement it in a helper function with an explicit name that tell what's the monster aims to.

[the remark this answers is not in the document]
GseoC: Would you be ok with:

`static bool gic_flags_are_exclusive(uint32_t cpu_mask)
{
	if (((cpu_mask &
	      (ITR_CPU_MASK_TO_THIS_CPU | ITR_CPU_MASK_TO_OTHER_CPUS)) &&
	     (!!(cpu_mask & ITR_CPU_MASK_TO_THIS_CPU) ^
	      !!(cpu_mask & ITR_CPU_MASK_TO_OTHER_CPUS)) &&
	     !(cpu_mask &
	       ~(ITR_CPU_MASK_TO_THIS_CPU | ITR_CPU_MASK_TO_OTHER_CPUS))) ||
	    (!(cpu_mask &
	       (ITR_CPU_MASK_TO_THIS_CPU | ITR_CPU_MASK_TO_OTHER_CPUS))))
		return true;

	return false;

...

	/* Assert exclusivity of mask hint bits and effective CPU mask bits */
	assert(cpu_mask && gic_flags_are_exclusive(cpu_mask));
}`

jenswi-linaro:
> > What about the cores in other clusters? I would assume the reasons why we want to stop the other cores in the same cluster also apply to the cores in other clusters?
> 
> This limitation is due to GIC driver current implementation. I think it's fine as a first step. Multi-cluster platforms will maybe prefer to not enable `CFG_[CORE_]HALT_CORES_ON_PANIC` if they have concerns with this limitation.

Really? From the specifications, it looks like it should work outside the clusters too.

core/drivers/gic.c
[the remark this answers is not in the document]
etienne-lms: I prefer Jens' proposal. Extended to check hint bits and cpu list bitfield are exclusive: 
```c
static bool assert_cpu_mask_is_valid(uint32_t cpu_mask)
{
	bool __maybe_unused to_others = cpu_mask & ITR_CPU_MASK_TO_OTHER_CPUS;
	bool __maybe_unused to_current = cpu_mask & ITR_CPU_MASK_TO_THIS_CPU;
	bool __maybe_unused to_list = cpu_mask & 0xff;

	assert(to_others + to_current + to_list == 1);
}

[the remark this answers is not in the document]
Comment (jenswi-linaro) [reply core/drivers/gic.c:410]: Fine with a helper function, but the expression in that function is digested easily.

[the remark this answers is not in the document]
Comment (jenswi-linaro) [reply core/drivers/gic.c:410]: That assert just begs for a _"There can be only one."_ comment :-)

File: mk/config.mk (hunk 6 above)
Comment (jenswi-linaro) [reply mk/config.mk:1073]: Makes sense, thanks.

**Comment by etienne-lms** [comment]
Sorry, I'm almost a newbie on GICv3+. Reading again the spec, indeed when using xSGI1R[IRM]=1, all CPUs from all clusters are notified. It is only when using TargetList (xSGI1R[IRM]=0) that a single write to xSGI1R can notify CPUs from a unique cluster.
@GseoC: sentence `# Note: Other clusters of CPU won't be notified` can be removed from _mk/config.mk_.

File: core/drivers/gic.c (hunk 7 above)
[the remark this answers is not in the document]
Comment (GseoC) [reply core/drivers/gic.c:410]: Ok for the last proposal + the comment then :)

File: mk/config.mk (hunk 6 above)
Comment (GseoC) [reply mk/config.mk:1073]: I will shorten the switch, thanks for clarifying

**Comment by GseoC** [comment]
Looking at the spec, the cluster-only target statement is indeed false. I'll make the change. Thanks for noticing

File: core/arch/arm/kernel/boot.c (hunk 3 above)
Comment (GseoC) [reply core/arch/arm/kernel/boot.c:1290]: > Should be called only once in case there are more than 2 cores.

Would adding a static assert: static_assert(CFG_TEE_CORE_NB_CORE > 1) be preferable?

-----------------------------------
> Wouldn't it make sense to put this in a boot_final() function instead? We could even move the code to `core/kernel/panic.c` if we could only resolve the masking of timer interrupts.

I've done it this way so that it's called when the secondary CPU init is done.

I can move the handler init code to _panic.c_  and call it here in init_secondary_helper() if IS_ENABLED(CFG_CORE_HALT_CORES_ON_PANIC) if that's ok?

Comment (etienne-lms) [reply core/arch/arm/kernel/boot.c:1290]: > > Should be called only once in case there are more than 2 cores.
> 
> Would adding a static assert: static_assert(CFG_TEE_CORE_NB_CORE > 1) be preferable?

No, I mean if there are 4 core, the 3 secondary cores will go through this but while each secondary core needs to enable/unmask its SGI interrupt (`interrupt_enable()`) only 1 core shall to register the interrupt callback handler (`interrupt_add_handler_with_chip()`) unless what the handler will be registered several times which is not what you want and will even fail since `ITRF_SHARED` is not set.

> I can move the handler init code to panic.c and call it here in init_secondary_helper() if IS_ENABLED(CFG_CORE_HALT_CORES_ON_PANIC) if that's ok?

+1

Comment (GseoC) [reply core/arch/arm/kernel/boot.c:1290]: Sorry I misread your comment. I'll add a static variable to control that

**Comment by GseoC** [comment]
Comments adressed + added [core: panic: add non-secure context dump for ARM32 platforms](https://github.com/OP-TEE/optee_os/pull/6440/commits/b7b2955fbbd150f04c111a47195c242f9992c4a3)

-------------------------------------------

Series tested on:
- STM32MP157C-DK2 : ARM32 V7 with 2 cores
- qemu_armv8a: 8 and 16 cores

Note: QEMU_SMP has a limitation of core in order to compile (16 was ok but not 50)

File: core/kernel/panic.c [hunk 8]
```diff
@@ -4,11 +4,138 @@
...
+static enum itr_return __noreturn
+multi_core_panic_it_handler(struct itr_handler *hdl  __unused)
+{
+	IMSG("Halting CPU %zu", get_core_pos());
+
+#ifdef ARM32
+	mask_timer_interrupts();
+#endif /* ARM32 */
+
+#if defined(ARM32) && !defined(CFG_WITH_ARM_TRUSTED_FW)
```
GseoC, line 98: Looks a bit strange, to be synced with the above ifdef

core/kernel/panic.c
```diff
@@ -4,11 +4,138 @@
...
+		cpu_spin_unlock(&lock);
+		return;
+	}
+
+	DMSG("CPU : %zu\n", get_core_pos());
+
+	reg = (uint32_t *)&sm_nsec_ctx->ub_regs.usr_sp;
+
+	for (i = U(0); i < ARRAY_SIZE(dump_table); i++)
+		DMSG("%10s : %#8"PRIx32, dump_table[i], reg[i]);
```
jenswi-linaro, line 69: Why is dumping these registers meaningful? Anyway, it doesn't belong here in generic code.

core/drivers/gic.c
```diff
@@ -389,34 +401,74 @@ static void gic_it_set_pending(struct gic_data *gd, size_t it)
 	io_write32(gd->gicd_base + GICD_ISPENDR(idx), mask);
 }
 
+static void assert_cpu_mask_is_valid(uint32_t cpu_mask)
+{
+	bool __maybe_unused to_others = cpu_mask & ITR_CPU_MASK_TO_OTHER_CPUS;
+	bool __maybe_unused to_current = cpu_mask & ITR_CPU_MASK_TO_THIS_CPU;
+	bool __maybe_unused to_list = cpu_mask & 0xff;
+
+	/* "There can be only one" flag set */
```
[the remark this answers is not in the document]
GseoC: Highlander will be sad

core/kernel/panic.c
```diff
@@ -4,11 +4,138 @@
...
+static void dump_core_registers(void)
+{
+	struct sm_nsec_ctx *sm_nsec_ctx = sm_get_nsec_ctx();
+	static unsigned int lock = SPINLOCK_UNLOCK;
+	uint32_t *reg = NULL;
+	size_t i = U(0);
+
+	cpu_spin_lock(&lock);
+
+	if (!sm_nsec_ctx) {
```
jenswi-linaro, line 59: `sm_get_nsec_ctx()` doesn't return `NULL` on error.

mk/config.mk
```diff
@@ -1063,3 +1063,10 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# Make other cores halt on panic() in a multi core setup. An SGI will be
+# triggered to notify all other CPUs that a panic() is occurring. The SGI used
+# to notify other cores is specified in CFG_HALT_CORES_ON_PANIC_SGI.
+# Note: Other clusters of CPU won't be notified
```
jenswi-linaro, line 1070: This line can be removed.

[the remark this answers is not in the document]
GseoC: Yes sorry, I've removed one above and didn't remove that one...

core/drivers/gic.c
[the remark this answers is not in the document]
GseoC: Actually, to_list could be constituted of multiple bits as it is the target list and not the target list filter.

See: [https://developer.arm.com/documentation/ihi0048/b/Programmers--Model/Distributor-register-descriptions/Software-Generated-Interrupt-Register--GICD-SGIR?lang=en](url)

core/kernel/panic.c
```diff
@@ -4,11 +4,138 @@
...
 
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
 #include <kernel/panic.h>
 #include <kernel/thread.h>
 #include <kernel/unwind.h>
 #include <trace.h>
 
+static_assert(CFG_HALT_CORES_ON_PANIC_SGI >= 8 &&
+	      CFG_HALT_CORES_ON_PANIC_SGI < 16);
```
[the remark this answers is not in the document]
GseoC: Sounds good

core/kernel/panic.c
```diff
@@ -4,11 +4,138 @@
...
+						    &multi_core_panic_handler))
+			panic();
+
+		interrupt_enable(interrupt_get_main_chip(),
+				 multi_core_panic_handler.it);
+
+		already_registered = true;
+	}
+}
+
```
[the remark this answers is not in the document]
GseoC: In order to avoid registering the handler before at least another CPU is running to avoid sending a SGI in the void. But maybe that's OK.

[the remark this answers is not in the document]
GseoC: Adding a check on CFG_TEE_CORE_NB_CORE > 1 in init_multi_core_panic_handler() seems enough to counter why I did that so I'll switch to call_finalcalls()

core/kernel/panic.c
GseoC: It is useful to get information on a non-secure call that caused a panic(). Of course, only one of the core dump will be relevant.

Do you have a suggestion on where to put this code?

core/drivers/gic.c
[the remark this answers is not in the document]
etienne-lms: Gosh, I missed the ref in Jens' comment :)    (i feel bad, that movie is of my generation)
Maybe "There must be only one" :confused:

core/kernel/panic.c
jenswi-linaro: Has it ever been used? I mean, we're normally entering OP-TEE in the same way so you'll likely to see the same thing each time, except perhaps that you'll be able to see if you entered due to a secure interrupt instead of an SMC. However, that should be easier to see via the panic stack trace.
Perhaps I'm missing something, but this seems like something that hasn't been tried in practice.

etienne-lms: This is intended to help finding out where non-secure world is executing on some specific error cases trapped by OP-TEE: mainly when a piece of non-secure code tries to access an unauthorized resource and when a secure watchdog is about to expire. We configure the related firewall driver in OP-TEE to raise a secure interrupt which print some driver useful related info and then panic (e.g. in [_plat-stm32mp1/plat_tzc400.c_](https://github.com/OP-TEE/optee_os/blob/4.0.0/core/arch/arm/plat-stm32mp1/plat_tzc400.c#L26-L30)). The panic traps all CPUs that (with the proposed change here) print their non-secure context.

jenswi-linaro: Back to my question. Has it ever been used? Do you know that the information is useful? It seems that you're missing the non-secure PC for instance.

etienne-lms: indeed `mon_spsr` and `mon_lr` fields are missing!

core/kernel/panic.c
```diff
@@ -4,11 +4,138 @@
...
+	"pad"
+#endif
+};
+
+static void dump_core_registers(void)
+{
+	struct sm_nsec_ctx *sm_nsec_ctx = sm_get_nsec_ctx();
+	static unsigned int lock = SPINLOCK_UNLOCK;
+	uint32_t *reg = NULL;
+	size_t i = U(0);
```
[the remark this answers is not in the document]
GseoC: Obsolete

core/kernel/panic.c
```diff
@@ -4,11 +4,138 @@
...
+}
+#endif /* ARM32 */
+
+static enum itr_return __noreturn
+multi_core_panic_it_handler(struct itr_handler *hdl  __unused)
+{
+	IMSG("Halting CPU %zu", get_core_pos());
+
+#ifdef ARM32
+	mask_timer_interrupts();
```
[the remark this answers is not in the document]
GseoC: Obsolete

core/drivers/gic.c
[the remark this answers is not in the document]
GseoC: I picked your first suggestion, but would have enjoyed that little Easter egg 😄

GseoC:
`Tested-by: Gatien Chevallier <gatien.chevallier@foss.st.com> (QEMUv8)` 8 and 16 cores
`Tested-by: Gatien Chevallier <gatien.chevallier@foss.st.com> (stm32mp1-157C_DK2)`

for "core: interrupt: halt other cores when one is panicking" with review comments

core/kernel/panic.c
```diff
@@ -4,11 +4,76 @@
...
+static void notify_other_cores(void)
+{
+	interrupt_raise_sgi(interrupt_get_main_chip(),
+			    CFG_HALT_CORES_ON_PANIC_SGI,
+			    ITR_CPU_MASK_TO_OTHER_CPUS);
+}
+
+static TEE_Result init_multi_core_panic_handler(void)
+{
+	static bool already_registered;
```
jenswi-linaro, line 57: This isn't needed any longer.

GseoC: Right, final call of primary boot, thanks

mk/config.mk
```diff
@@ -1063,3 +1063,10 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# Make other cores halt on panic() in a multi core setup. An SGI will be
```
etienne-lms, line 1067:
```
# CFG_HALT_CORES_ON_PANIC when enabled makes any call to panic() help other
# core. The feature currently relies on GIC device to notify other cores
# using an SGI interrupt speficied by CFG_HALT_CORES_ON_PANIC_SGI.

**Review by etienne-lms — COMMENTED** [review]
If @jenswi-linaro is ok, you can squash the fixup commit and update the few commit messages that need to be. It will ease the review IMHO.
E.g. in commit "core: interrupt: halt other cores when one is panicking":
```patch
 When one core panics, send an SGI (CFG_CORE_HALT_CORES_ON_PANIC_SGI)
-notifying other cores of the same cluster that they should halt if
-CFG_CORE_HALT_CORES_ON_PANIC is set.
+to halt the other cores if CFG_CORE_HALT_CORES_ON_PANIC is enabled.
```

GseoC:
> If @jenswi-linaro is ok, you can squash the fixup commit and update the few commit messages that need to be. It will ease the review IMHO. E.g. in commit "core: interrupt: halt other cores when one is panicking":
> 
> ```diff
>  When one core panics, send an SGI (CFG_CORE_HALT_CORES_ON_PANIC_SGI)
> -notifying other cores of the same cluster that they should halt if
> -CFG_CORE_HALT_CORES_ON_PANIC is set.
> +to halt the other cores if CFG_CORE_HALT_CORES_ON_PANIC is enabled.
> ```

I'll address the comment while squashing the fixups if that's ok

jenswi-linaro:
Feel free to squash in the fixups.

GseoC:
Fixups squashed and comment addressed.

core/drivers/gic.c
```diff
@@ -389,34 +401,74 @@ static void gic_it_set_pending(struct gic_data *gd, size_t it)
...
+	uint64_t mask = SHIFT_U64(mask_id, 24);
+
+	assert_cpu_mask_is_valid(cpu_mask);
+
+	if (cpu_mask & ITR_CPU_MASK_TO_OTHER_CPUS) {
+		mask |= BIT64(GICC_SGI_IRM_BIT);
+	} else {
+		uint64_t mpidr = read_mpidr();
+		uint64_t mask_aff1 = (mpidr & MPIDR_AFF1_MASK) >>
+				      MPIDR_AFF1_SHIFT;
```
etienne-lms, line 428: remove 1 space char to fix indentation.

core/kernel/panic.c
```diff
@@ -4,11 +4,57 @@
...
 
+#include <initcall.h>
+#include <kernel/interrupt.h>
+#include <kernel/misc.h>
 #include <kernel/panic.h>
 #include <kernel/thread.h>
 #include <kernel/unwind.h>
 #include <trace.h>
 
+static_assert(!IS_ENABLED(CFG_HALT_CORES_ON_PANIC) ||
```
etienne-lms, line 15: An inline comment is welcome to explain why SGI number should be in that range.
Something like `/* SGI number must be in the secure SGI range */`

core/kernel/panic.c
```diff
@@ -4,11 +4,57 @@
...
+	      (CFG_HALT_CORES_ON_PANIC_SGI >= 8 &&
+	       CFG_HALT_CORES_ON_PANIC_SGI < 16));
+
+static enum itr_return __noreturn
+multi_core_panic_it_handler(struct itr_handler *hdl  __unused)
```
etienne-lms, line 20: 2 space chars

etienne-lms (commented):
"core: interrupt_raise_sgi() updates" (with comment addressed),
"core: interrupt: halt other cores when one is panicking" (with comment addressed),
"plat-stm32mp1: activate CFG_CORE_HALT_CORES_ON_PANIC" and
"plat-stm32mp2: activate CFG_CORE_HALT_CORES_ON_PANIC".

@jenswi-linaro, would you agree do also enable this feature on the several vexpress platforms (_qemu*_, _fvp_ and _juno_) ? (proposal is to use also SGI 9).
If so, @gatien, could you add such change in your series and by the way fix that platform `conf.mk` file: s/`CFG_TEE_CORE_NB_CORE =`/`CFG_TEE_CORE_NB_CORE ?=`/g.

mk/config.mk
```diff
@@ -1077,3 +1077,10 @@ CFG_HMAC_64_1024_RANGE ?= n
 # By default use standard pbkdf2 implementation
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
+
+# CFG_HALT_CORES_ON_PANIC, when enabled, makes any call to panic() halt the
+# other cores. The feature currently relies on GIC device to trap the other
+# cores using an SGI interrupt specified by CFG_HALT_CORES_ON_PANIC_SGI.
+CFG_HALT_CORES_ON_PANIC ?= n
+CFG_HALT_CORES_ON_PANIC_SGI ?= 9
```
jenswi-linaro, line 1085: How about making the default `15` instead? If we're going to donate another secure SGI to the normal world we're likely to want to pick `9` after `8`.

jenswi-linaro:
> @jenswi-linaro, would you agree do also enable this feature on the several vexpress platforms (qemu*, fvp and juno) ? (proposal is to use also SGI 9).

Sure, but perhaps with SGI 15 instead as I mentioned in the `mk/config.mk` comment.

mk/config.mk
GseoC: Sounds good, I'll make the change

GseoC:
I'll force the SGI number at 15 at platform level for ST platforms. I'll discard your R-B tag @etienne-lms so you can let me know if that's ok for you

GseoC:
Tag applied for 
```
core: interrupt_raise_sgi() updates
core: interrupt: halt other cores when one is panicking
```

Added :
`[plat-vexpress: activate CFG_CORE_HALT_CORES_ON_PANIC](https://github.com/OP-TEE/optee_os/pull/6440/commits/7c7dcfaa3096d0711c145c3de23df2e0a75010a7)` for vexpress platforms

---------------------------

Tested-by: Gatien Chevallier <gatien.chevallier@foss.st.com> (QEMUv8) 8 and 16 cores
Tested-by: Gatien Chevallier <gatien.chevallier@foss.st.com> (QEMUv7 (qemu_virt)) 8  cores
Tested-by: Gatien Chevallier <gatien.chevallier@foss.st.com> (stm32mp1-157C_DK2)

jforissier:
For "plat-vexpress: activate CFG_CORE_HALT_CORES_ON_PANIC":

etienne-lms (commented):
"plat-stm32mp1: activate CFG_CORE_HALT_CORES_ON_PANIC" and
"plat-stm32mp2: activate CFG_CORE_HALT_CORES_ON_PANIC".

"plat-vexpress: activate CFG_CORE_HALT_CORES_ON_PANIC".

GseoC:
Tags applied, thanks for the review

PR #6443, opened 2023-11-08
core: fix calls to tee_pobj_get() for object opening

etienne-lms:
Fixes calls to tee_pobj_get() that use boolean false are argument where a enum tee_pobj_usage argument is expected.

Function tee_pobj_get() used to take a boolean @temporary argument between old 2.4.0 and 3.11.0 release tags. Since commit [1] merged in release tag 3.11.0, function prototype changed but the commit missed few calls that since still use value false (0), treated as enumerated value TEE_POBJ_USAGE_OPEN. This change replaces these occurrences with use of TEE_POBJ_USAGE_OPEN argument.

Link: 6885abf2f7ef ("core: tee_pobj_get() takes an enum tee_pobj_usage") [1]

### DIFF
--- a/core/pta/attestation.c
+++ b/core/pta/attestation.c
@@ -228,8 +228,8 @@ static TEE_Result sec_storage_obj_read(TEE_UUID *uuid, uint32_t storage_id,
 	if (obj_id_len > TEE_OBJECT_ID_MAX_LEN)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	res = tee_pobj_get(uuid, (void *)obj_id, obj_id_len, flags, false, fops,
-			   &po);
+	res = tee_pobj_get(uuid, (void *)obj_id, obj_id_len, flags,
+			   TEE_POBJ_USAGE_OPEN, fops, &po);
 	if (res)
 		return res;
 
@@ -272,8 +272,8 @@ static TEE_Result sec_storage_obj_write(TEE_UUID *uuid, uint32_t storage_id,
 	if (obj_id_len > TEE_OBJECT_ID_MAX_LEN)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	res = tee_pobj_get(uuid, (void *)obj_id, obj_id_len, flags, false,
-			   fops, &po);
+	res = tee_pobj_get(uuid, (void *)obj_id, obj_id_len, flags,
+			   TEE_POBJ_USAGE_OPEN, fops, &po);
 	if (res)
 		return res;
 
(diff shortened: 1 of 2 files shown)


etienne-lms (commented):
i'll rephrase the commit message...

jenswi-linaro:
Please add a fixes tag for easier referencing.

etienne-lms:
Updated

core/pta/attestation.c
```diff
@@ -272,8 +272,8 @@ static TEE_Result sec_storage_obj_write(TEE_UUID *uuid, uint32_t storage_id,
 	if (obj_id_len > TEE_OBJECT_ID_MAX_LEN)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	res = tee_pobj_get(uuid, (void *)obj_id, obj_id_len, flags, false,
-			   fops, &po);
+	res = tee_pobj_get(uuid, (void *)obj_id, obj_id_len, flags,
+			   TEE_POBJ_USAGE_OPEN, fops, &po);
```
etienne-lms, line 276: Should this be `TEE_POBJ_USAGE_OPEN` or `TEE_POBJ_USAGE_CREATE`?

jenswi-linaro: To maintain the previous behaviour it should be `TEE_POBJ_USAGE_OPEN` and that probably be good enough since this PTA isn't concurrent.

jenswi-linaro:
Interesting with the fixes tags, I thought it was 6885abf2f7ef ("core: tee_pobj_get() takes an enum tee_pobj_usage") that introduced the problem, but it was actually 42471ecf25b7 ("core: load stmm via secure partition") that introduced the problem that later it seems was also copied in 7e05ec25bd68 ("core: pta: add remote attestation PTA").

I'm not sure that the tag `Fixes: f9cd31c ("core: rename secure_partition to stmm_sp")` is needed, but it doesn't harm so feel free to keep it.

core/pta/attestation.c
etienne-lms: Ok. I had the same concern for the StMM (now in the 1st commit of the series).
This P-R obviously aims at replacing 0/`false` with 0/`TEE_POBJ_USAGE_OPEN` but I think it's a good opportunity to challenge use of the later on write oprations.

jenswi-linaro: Using `TEE_POBJ_USAGE_CREATE` correctly might be a bit more trouble than it's worth. The main reason behind `tee_pobj_get()` is to enforce access logic as defined in the GP TEE Internal Core API spec. Here I guess it's mostly to avoid initializing `struct tee_pobj` directly like in `check_update_version()`.

etienne-lms:
Yes, I'd rather keep both refs as the commit can apply on commit [1] but not on commit [2] even if the fixed issue comes from commit [2].

etienne-lms:
thanks for the review

etienne-lms:
tags aplied

PR #6445, opened 2023-11-09
ta: pkcs11: Improve PIN counter handling robustness

loicpoulain:
Make sure PIN check attempt is saved persistently before continuing with the actual PIN verification, improving counter and flags coherency in case of subsequent failure with persistent saving.


etienne-lms:
Looks good to me. Please fix checkpatch issue. 
With that addressed, `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

jforissier:
I suggest fixing the line length messages with a temporary variable.
```C
struct token_persistent_main *db = token->db_main;
```

PR #6447, opened 2023-11-10
core: riscv: Modify S-mode boot flow to compatible with M-mode firmware

gagachang:
In system of S-mode OP-TEE, M-mode firmware transfers control flow to OP-TEE during boot. M-mode runs one hart, which called boot hart, to executes system initialization and jumps into OP-TEE.

However, the ID of boot hart might not be always zero. That is, M-mode firmware may choose a hart, whose ID is non-zero, to jump into OP-TEE "_start" symbol to do OP-TEE initialization. Therefore, it is not feasible to let all harts enter "_start" symbol, and judge whether current hart is primary boot hart, by checking ID is zero or not. Fix it by letting non-boot hart enter "reset_secondary", in S-mode OP-TEE. So that all the harts can do their right things regardless of their hart ID.

Only OP-TEE runs in M-mode will judge primary hart by checking whether the hart ID is zero.

### DIFF
--- a/core/arch/riscv/kernel/entry.S
+++ b/core/arch/riscv/kernel/entry.S
@@ -148,7 +148,12 @@ FUNC _start , :
 #else
 	mv	s1, a1		/* Save device tree address into s1 */
 #endif
-	bnez	a0, reset_secondary
+
+	/* Only first hart who wins lottery runs the primary boot sequence. */
+	la	a3, hart_lottery
+	li	a2, 1
+	amoadd.w a3, a2, (a3)
+	bnez	a3, reset_secondary
 	jal	reset_primary
 	j	.
 END_FUNC _start
@@ -245,6 +250,13 @@ LOCAL_FUNC unhandled_cpu , :
 	wfi
 	j	unhandled_cpu
 END_FUNC unhandled_cpu
+
+	.section .identity_map.data
+	.balign	8
+LOCAL_DATA hart_lottery , :
+	/* The hart who first increments this variable will be primary hart. */
+	.word	0
+END_DATA hart_lottery
 
 #ifdef CFG_BOOT_SYNC_CPU
 LOCAL_DATA sem_cpu_sync_start , :
--- a/core/arch/riscv/kernel/entry.S
+++ b/core/arch/riscv/kernel/entry.S
@@ -243,6 +243,11 @@ UNWIND(	.cantunwind)
 	cpu_is_ready
 
 	jal	boot_init_secondary
+#ifdef CFG_RISCV_WITH_M_MODE_SM
+	/* Return to untrusted domain */
+	li	a0, TEEABI_OPTEED_RETURN_ON_DONE
+	j	thread_return_to_udomain
+#endif
 	j	.
 END_FUNC reset_secondary
 
--- a/core/arch/riscv/plat-virt/conf.mk
+++ b/core/arch/riscv/plat-virt/conf.mk
@@ -30,6 +30,7 @@ $(call force,CFG_16550_UART,y)
 $(call force,CFG_RISCV_TIME_SOURCE_RDTIME,y)
 CFG_RISCV_MTIME_RATE ?= 10000000
 CFG_RISCV_SBI ?= y
+CFG_RISCV_WITH_M_MODE_SM ?= y
 
 # TA-related flags
 supported-ta-targets = ta_rv64
--- a/core/arch/riscv/riscv.mk
+++ b/core/arch/riscv/riscv.mk
@@ -63,6 +63,13 @@ ifeq ($(CFG_RISCV_SBI_CONSOLE),y)
 $(call force,CFG_RISCV_SBI,y)
 endif
 
+# 'y' to let M-mode secure monitor handle the communication between OP-TEE OS
+# and untrusted domain.
+CFG_RISCV_WITH_M_MODE_SM ?= n
+ifeq ($(CFG_RISCV_WITH_M_MODE_SM),y)
+$(call force,CFG_RISCV_SBI,y)
+endif
+
 # Disable unsupported and other arch-specific flags
 $(call force,CFG_CORE_FFA,n)
 $(call force,CFG_SECURE_PARTITION,n)
--- a/core/arch/riscv/kernel/thread_optee_abi_rv.S
+++ b/core/arch/riscv/kernel/thread_optee_abi_rv.S
@@ -29,7 +29,7 @@
  */
 FUNC thread_return_to_udomain , :
 	/* Caller should provide arguments in a0~a5 */
-#if defined(CFG_RISCV_SBI)
+#if defined(CFG_RISCV_WITH_M_MODE_SM)
 	li	a7, SBI_EXT_TEE		/* extension ID */
 	li	a6, 0			/* function ID (unused) */
 	ecall


gagachang:
Hello @maroueneboubakri 
Please also check this PR, thanks!

maroueneboubakri:
Hi @gagachang, I still need more time on this, since it may impacts our intended execution flow (not SBI based).

gagachang:
> Hi @gagachang, I still need more time on this, since it may impacts our intended execution flow (not SBI based).

Hi @maroueneboubakri ,Sure!
May I know your execution flow?
I remembered your system is "dual-core" based, one hart runs linux and another hart runs OP-TEE ?
Does OP-TEE runs as M-mode ?

maroueneboubakri:
Hi @gagachang, Still didn't try it, I prioritize it once I finish ongoing work in my pipe !

gagachang:
Just rebase to master branch

core/arch/riscv/kernel/entry.S
```diff
@@ -231,13 +234,27 @@ END_FUNC reset_primary
...
 	set_sp
 	set_tp
 	set_satp
 	cpu_is_ready
 
 	jal	boot_init_secondary
+#ifdef CFG_RISCV_S_MODE
+	/* Return to untrusted domain */
+	li	a0, TEEABI_OPTEED_RETURN_ON_DONE
+	j	thread_return_to_ree
```
maroueneboubakri, line 256: @gagachang for this I guess we need to add an extra flag, to decide to use a Security Monitor based solution, or not (could be messaging based as we use today without the need for a SM).
I also suggest to use `thread_return_to_udomain` to keep same wording accross the work.

gagachang: > @gagachang for this I guess we need to add an extra flag, to decide to use a Security Monitor based solution, or not (could be messaging based as we use today without the need for a SM).

It is decided in the implementation of `thread_return_to_ree()`.
Please see my previous commit https://github.com/OP-TEE/optee_os/commit/0fb22936229d51e005aaf159e628e516a60faaa5
If Security Monitor based solution is adopted, `CFG_RISCV_SBI` should be defined and `ecall` is used for communication.
The messaging based solution can be also implemented into `thread_return_to_ree()` by defining another configuration instead of `CFG_RISCV_SBI`.

```C
FUNC thread_return_to_ree , :
	/* Caller should provide arguments in a0~a5 */
#if defined(CFG_RISCV_SBI)
	li	a7, SBI_EXT_TEE		/* extension ID */
	li	a6, 0			/* function ID (unused) */
	ecall
#else
	/* Other protocol */
#endif
	/* ABI to REE should not return */
	panic_at_abi_return
END_FUNC thread_return_to_ree
```

> I also suggest to use `thread_return_to_udomain` to keep same wording accross the work.

OK I agree, "domain" is more precise terminology in RISC-V.
I can rename those `thread_return_to_ree()` to `thread_return_to_udomain()` in another PR.

maroueneboubakri: `CFG_RISCV_SBI` is still needed when running OP-TEE OS in S-Mode, regardless how trusted and trusted domain communicate. In your case you are using a specific extension in SBI to context switch between both worlds, but SBI is nedded for other supervisor operations.

gagachang: > `CFG_RISCV_SBI` is still needed when running OP-TEE OS in S-Mode, regardless how trusted and trusted domain communicate. In your case you are using a specific extension in SBI to context switch between both worlds, but SBI is nedded for other supervisor operations.

Yes you are right. SBI has a bunch of extensions, not only for domain communication.
How about `CFG_MSG_BY_SBI` and `CFG_MSG_BY_RPMSG` ? (Assume message based protocol named RPMSG)

maroueneboubakri: For ARM, there is a flag `CFG_WITH_ARM_TRUSTED_FW` when we make use of ATF as SM. I propose, for your case, `CFG_RISCV_WITH_SM` that depends on `CFG_RISCV_SBI` maybe ? the addition for messaging based solution we will not be pushed soon. Feel free to enable your SM based code with that flag.

gagachang: > I propose, for your case, `CFG_RISCV_WITH_SM` that depends on `CFG_RISCV_SBI` maybe ?

The [RISC-V CoVE](https://github.com/riscv-non-isa/riscv-ap-tee/tree/main) specification has already defined two kinds of SM, one runs as M-mode and another runs as HS-mode.
Currently we have not supported OP-TEE with hypervisor extension.
How about `CFG_RISCV_WITH_M_MODE_FW` ? It may be more precise.

gagachang:
Hi @maroueneboubakri 
Please see https://github.com/OP-TEE/optee_os/pull/6602 for `udomain`

core/arch/riscv/kernel/entry.S
maroueneboubakri: `CFG_RISCV_WITH_M_MODE_FW` is similar to `CFG_RISCV_SBI` an M-Mode FW is needed regardless of the the SM extension. Maybe `CFG_RISCV_WITH_M_MODE_SM` is more precise.

gagachang: > `CFG_RISCV_WITH_M_MODE_FW` is similar to `CFG_RISCV_SBI` an M-Mode FW is needed regardless of the the SM extension. Maybe `CFG_RISCV_WITH_M_MODE_SM` is more precise.

OK, now I know "SM" means the communication is handled by M-mode FW.
So your messaging based solution has M-mode "FW" but not M-mode "SM", right ?
I will add `CFG_RISCV_WITH_M_MODE_SM`.

core/arch/riscv/riscv.mk
```diff
@@ -63,6 +63,15 @@ ifeq ($(CFG_RISCV_SBI_CONSOLE),y)
 $(call force,CFG_RISCV_SBI,y)
 endif
 
+# 'y' to let M-mode secure monitor handle the communication between OP-TEE OS
+# and untrusted domain.
+CFG_RISCV_WITH_M_MODE_SM ?= n
+ifeq ($(CFG_RISCV_WITH_M_MODE_SM),y)
+ifneq (y,$(call cfg-all-enabled,CFG_RISCV_S_MODE CFG_RISCV_SBI))
+$(error Both CFG_RISCV_S_MODE or CFG_RISCV_SBI must be 'y')
+endif
```
maroueneboubakri, line 73: `$(call force,CFG_RISCV_SBI,y)`

core/arch/riscv/plat-virt/conf.mk
```diff
@@ -30,6 +30,7 @@ $(call force,CFG_16550_UART,y)
 $(call force,CFG_RISCV_TIME_SOURCE_RDTIME,y)
 CFG_RISCV_MTIME_RATE ?= 10000000
 CFG_RISCV_SBI ?= y
+CFG_RISCV_WITH_M_MODE_SM ?= y
```
maroueneboubakri, line 33: In case of selecting inter-domains messaging-based solution, how we will select it ?
No plan yet to add our implementation, let's keep it for now.

core/arch/riscv/riscv.mk
```diff
@@ -44,6 +44,7 @@ endif
 
 CFG_RISCV_SBI	 ?= n
 CFG_RISCV_M_MODE ?= y
+CFG_RISCV_S_MODE ?= n
```
maroueneboubakri, line 47: Not needed, the checks after sets it according the `CFG_RISCV_M_MODE `.

core/arch/riscv/kernel/entry.S
```diff
@@ -231,13 +234,27 @@ END_FUNC reset_primary
...
 	set_sp
 	set_tp
 	set_satp
 	cpu_is_ready
 
 	jal	boot_init_secondary
+#ifdef CFG_RISCV_WITH_M_MODE_SM
+	/* Return to untrusted domain */
+	li	a0, TEEABI_OPTEED_RETURN_ON_DONE
+	j	thread_return_to_ree
```
maroueneboubakri, line 256: `s/thread_return_to_ree/thread_return_to_udomain/` commit has been merged.

core/arch/riscv/plat-virt/conf.mk
gagachang: Maybe we can select it by device tree node in the future ?
Like linux optee driver does:
https://elixir.bootlin.com/linux/latest/source/drivers/tee/optee/smc_abi.c#L1427
```c
static optee_invoke_fn *get_invoke_func(struct device *dev)
{
	const char *method;

	pr_info("probing for conduit method.\n");

	if (device_property_read_string(dev, "method", &method)) {
		pr_warn("missing \"method\" property\n");
		return ERR_PTR(-ENXIO);
	}

	if (!strcmp("hvc", method))
		return optee_smccc_hvc;
	else if (!strcmp("smc", method))
		return optee_smccc_smc;

	pr_warn("invalid \"method\" property: %s\n", method);
	return ERR_PTR(-EINVAL);
}
```

core/arch/riscv/riscv.mk
gagachang: OK, I'll drop it.

core/arch/riscv/riscv.mk
gagachang: Done

gagachang:
Hi @maroueneboubakri 
Any update on this PR?
If it is not feasible on your side, maybe we should think another general solution.

gagachang:
@maroueneboubakri 
I change the way to solve the issue.
Please check https://github.com/OP-TEE/optee_os/pull/6447/commits/b54d6b79fac52183c65de07556e472cfc6529f71 and https://github.com/OP-TEE/optee_os/pull/6447/commits/631ee90735e2756d4eaf10ff2796b1f5294fe7cd.

The issue here is because openSBI boots OP-TEE with "random" hart.
This commit uses "hart lottery" to solve this issue.

maroueneboubakri:
Hi @gagachang,

I'll try the PR out this week.

Best
Maro

gagachang:
> Hi @gagachang,
> 
> I'll try the PR out this week.
> 
> Best Maro

Thank you! Let me know if there is any issue.

maroueneboubakri:

gagachang:
@maroueneboubakri 
Tags applied, thanks!

PR #6452, opened 2023-11-10
core: print current guest ID in logs

jenswi-linaro:
If CFG_NS_VIRTUALIZATION is enabled include the current guest ID on each log line. A number is added before the core number identifying the currently set guest ID, for example:
D/TC:2 0 0 call_initcalls:40 level 1 teecore_init_pub_ram()

Where the "2" indicates that this is done with guest ID 2 active.

### DIFF
--- a/lib/libutils/ext/trace.c
+++ b/lib/libutils/ext/trace.c
@@ -7,12 +7,13 @@
 #include <platform_config.h>
 #endif
 
+#include <config.h>
 #include <printk.h>
 #include <stdarg.h>
 #include <string.h>
 #include <trace.h>
-#include <util.h>
 #include <types_ext.h>
+#include <util.h>
 
 #if (TRACE_LEVEL < TRACE_MIN) || (TRACE_LEVEL > TRACE_MAX)
 #error "Invalid value of TRACE_LEVEL"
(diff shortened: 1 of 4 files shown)


etienne-lms (commented):
I would rather print the NSec guest ID after core and thread IDs to for consistency between old and new trace formats, but that not a strong opinion.

In commit message: the guest ID is always added to the trace message, so prefer rephrase the sentence that says "If CFG_NS_VIRTUALIZATION is enabled include the current guest ID..."

`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with 2nd commit addressed and 1st comment addressed or not.

jenswi-linaro:
> I would rather print the NSec guest ID after core and thread IDs to for consistency between old and new trace formats, but that not a strong opinion.

I see your point. I've chosen to have the IDs in the order of significance. Both make sense, but we need to pick one order. What do others think?

jforissier:
> > I would rather print the NSec guest ID after core and thread IDs to for consistency between old and new trace formats, but that not a strong opinion.
> 
> I see your point. I've chosen to have the IDs in the order of significance. Both make sense, but we need to pick one order. What do others think?

I have a slight preference for the guest ID first because it will make it a bit easier to filter out the guests visually.

lib/libutils/ext/trace.c
```diff
@@ -133,6 +149,12 @@ void trace_vprintf(const char *function, int line, int level, bool level_ok,
 	boffs += res;
 
 	if (level_ok && (BIT(level) & CFG_MSG_LONG_PREFIX_MASK)) {
+		/* Print the current guest ID or 0 for Nexus */
+		res = print_guest_id(buf + boffs, sizeof(buf) - boffs);
+		if (res < 0)
```
jforissier, line 154: I think `scripts/symbolize.py` needs updating (`STACK_ADDR_RE`).

jenswi-linaro: Added a commit for this, let me know if you'd rather have it in a single commit.

jforissier: Thanks, one commit would be better indeed. With that:

jenswi-linaro:
Squashed and tags applied. @etienne-lms, did I get the commit message right?

etienne-lms:
Sorry, yes the commit log is right. All good to me.

PR #6456, opened 2023-11-13
core: arm: check for NULL mobj before thread_rpc_free()

jenswi-linaro:
In the SMC and FF-A ABIs check that a mobj representing a shared memory object isn't NULL before doing an RPC to free it in the normal world too. For the FF-A ABI it's harmless without this check, but the SMC ABI may cause an NULL pointer dereference in the OP-TEE kernel driver.


etienne-lms (commented):

jenswi-linaro:
Tag applied

PR #6457, opened 2023-11-13
CFG_INSECURE deprecates CFG_WARN_INSECURE (+ stm32mp1 related change)

etienne-lms:
Changes STM32MP1 shared resources to allow insecure RCC protection with peripherals assigned to secure world when CFG_WARN_INSECURE is enabled. This means for example that some SoC resources can be assigned to OP-TEE without their clock and reset controllers being effectively protected from non-secure accesses. Such configuration can be useful for development and test purposes.

This change does not affect devices provisioned with secret that are in so-called SEC_CLOSED state (BSEC fuses). Indeed this device state currently requires RCC protection to be enabled as already implemented in function check_rcc_secure_configuration().

### DIFF
--- a/core/arch/arm/plat-stm32mp1/shared_resources.c
+++ b/core/arch/arm/plat-stm32mp1/shared_resources.c
@@ -692,8 +692,12 @@ static void check_rcc_secure_configuration(void)
 		}
 	}
 
-	if (have_error)
-		panic();
+	if (have_error) {
+		if (IS_ENABLED(CFG_INSECURE))
+			EMSG("WARNING: CFG_INSECURE allows insecure RCC configuration");
+		else
+			panic();
+	}
 }
 
 static void set_gpio_secure_configuration(void)
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -909,15 +909,26 @@ CFG_REGULATOR_FIXED ?= n
 $(eval $(call cfg-enable-all-depends,CFG_REGULATOR_FIXED, \
 	 CFG_DRIVERS_REGULATOR CFG_DT))
 
-# The purpose of this flag is to show a print when booting up the device that
-# indicates whether the board runs a standard developer configuration or not.
-# A developer configuration doesn't necessarily has to be secure. The intention
+# When enabled, CFG_INSECURE permits insecure configuration of OP-TEE core
+# and shows a print (info level) when booting up the device that
+# indicates that the board runs a standard developer configuration.
+#
+# A developer configuration doesn't necessarily have to be secure. The intention
 # is that the one making products based on OP-TEE should override this flag in
 # plat-xxx/conf.mk for the platform they're basing their products on after
 # they've finalized implementing stubbed functionality (see OP-TEE
 # documentation/Porting guidelines) as well as vendor specific security
 # configuration.
-CFG_WARN_INSECURE ?= y
+#
+# CFG_WARN_INSECURE served the same purpose as CFG_INSECURE but is deprecated.
+ifneq (undefined,$(flavor CFG_WARN_INSECURE))
+$(info WARNING: CFG_WARN_INSECURE is deprecated, use CFG_INSECURE instead)
+CFG_INSECURE ?= $(CFG_WARN_INSECURE)
+ifneq ($(CFG_INSECURE),$(CFG_WARN_INSECURE))
+$(error Inconsistent CFG_INSECURE=$(CFG_INSECURE) and CFG_WARN_INSECURE=$(CFG_WARN_INSECURE))
+endif
+endif # CFG_WARN_INSECURE defined
+CFG_INSECURE ?= y
 
 # Enables warnings for declarations mixed with statements
 CFG_WARN_DECL_AFTER_STATEMENT ?= y
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1191,7 +1191,7 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
 	configure_console_from_dt();
 
 	IMSG("OP-TEE version: %s", core_v_str);
-	if (IS_ENABLED(CFG_WARN_INSECURE)) {
+	if (IS_ENABLED(CFG_INSECURE)) {
 		IMSG("WARNING: This OP-TEE configuration might be insecure!");
 		IMSG("WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html");
 	}
--- a/core/arch/arm/plat-stm32mp1/main.c
+++ b/core/arch/arm/plat-stm32mp1/main.c
@@ -553,7 +553,7 @@ static TEE_Result init_debug(void)
 		return res;
 
 	if (state != BSEC_STATE_SEC_CLOSED && conf) {
-		if (IS_ENABLED(CFG_WARN_INSECURE))
+		if (IS_ENABLED(CFG_INSECURE))
 			IMSG("WARNING: All debug accesses are allowed");
 
 		res = stm32_bsec_write_debug_conf(conf | BSEC_DEBUG_ALL);
--- a/core/arch/riscv/kernel/boot.c
+++ b/core/arch/riscv/kernel/boot.c
@@ -138,7 +138,7 @@ void boot_init_primary_late(unsigned long fdt,
 	update_external_dt();
 
 	IMSG("OP-TEE version: %s", core_v_str);
-	if (IS_ENABLED(CFG_WARN_INSECURE)) {
+	if (IS_ENABLED(CFG_INSECURE)) {
 		IMSG("WARNING: This OP-TEE configuration might be insecure!");
 		IMSG("WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html");
 	}
--- a/core/arch/riscv/plat-spike/conf.mk
+++ b/core/arch/riscv/plat-spike/conf.mk
@@ -67,7 +67,7 @@ $(call force,CFG_SECSTOR_TA,n)
 $(call force,CFG_SHOW_CONF_ON_BOOT,n)
 $(call force,CFG_SYSCALL_WRAPPERS_MCOUNT,n)
 $(call force,CFG_BOOT_SYNC_CPU,y)
-$(call force,CFG_WARN_INSECURE,y)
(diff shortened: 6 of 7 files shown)


jenswi-linaro (commented):

etienne-lms:
Tag applied. Thanks. 

Checkpatch warns but I've intentionally left the implementation as-is since it is a trace message size issue:
```
WARNING: line length of 93 exceeds 80 columns
#37: FILE: core/arch/arm/plat-stm32mp1/shared_resources.c:697:
+			EMSG("Warning: CFG_WARN_INSECURE allows insecure RCC configuration");

total: 0 errors, 1 warnings, 0 checks, 14 lines checked
```

core/arch/arm/plat-stm32mp1/shared_resources.c
```diff
@@ -692,8 +692,12 @@ static void check_rcc_secure_configuration(void)
 		}
 	}
 
-	if (have_error)
-		panic();
+	if (have_error) {
+		if (IS_ENABLED(CFG_WARN_INSECURE))
+			EMSG("Warning: CFG_WARN_INSECURE allows insecure RCC configuration");
```
jforissier, line 697: Since `CFG_WARN_INSECURE` can now change the behavior as opposed to just showing a warning, I suggest it should be renamed `CFG_INSECURE` and documented properly.

jenswi-linaro: Good point. Note that we also have changed behavior in `core/tee/tee_ree_fs.c` concerning the non-volatile counter.

etienne-lms: Do you want I create the change?

jforissier: Yes, one commit to rename the CFG first would be good (mentioning the volatile counter change as a reason to rename, and others to come).

etienne-lms: ok, i'll post in this P-R.

etienne-lms:
I've added commit "tree wide: CFG_INSECURE deprecates CFG_WARN_INSECURE".
I'll rename this P-R, for consistency.

etienne-lms:
CI / Code style failures are all related to issue shown in https://github.com/OP-TEE/optee_os/pull/6457#issuecomment-1809779430).

mk/config.mk
```diff
@@ -909,15 +909,28 @@ CFG_REGULATOR_FIXED ?= n
...
 # is that the one making products based on OP-TEE should override this flag in
 # plat-xxx/conf.mk for the platform they're basing their products on after
 # they've finalized implementing stubbed functionality (see OP-TEE
 # documentation/Porting guidelines) as well as vendor specific security
 # configuration.
-CFG_WARN_INSECURE ?= y
+CFG_INSECURE ?= y
+
+# CFG_WARN_INSECURE serves the same purpose as CFG_INSECURE but is deprecated
+# and should not be used. CFG_WARN_INSECURE will be removed in a future OP-TEE
```
jforissier, line 925: may be removed?

mk/config.mk
```diff
@@ -909,15 +909,28 @@ CFG_REGULATOR_FIXED ?= n
...
+# release.
+ifneq (,$(CFG_WARN_INSECURE))
+$(info CFG_WARN_INSECURE is deprecated, please use CFG_INSECURE instead)
+else
+ifneq ($(CFG_WARN_INSECURE),$(CFG_INSECURE))
+$(error Deprecated CFG_WARN_INSECURE must match CFG_INSECURE)
```
jforissier, line 930: This doesn't work due to "" != "n" but having the same meaning (disabled).
I don't think this is a proper thing to do. How about simply setting the default from the previous variable? 
```
CFG_INSECURE ?= $(CFG_WARN_INSECURE)
CFG_INSECURE ?= y
```

jforissier (commented):
"The new switch better represents what CFG_INSECURE has it does not only prints trace messages but also..." => "The new name is better because the switch not only warns but also change the OP-TEE core behavior as, for example..."

mk/config.mk
```diff
@@ -909,15 +909,28 @@ CFG_REGULATOR_FIXED ?= n
...
-CFG_WARN_INSECURE ?= y
+CFG_INSECURE ?= y
+
+# CFG_WARN_INSECURE serves the same purpose as CFG_INSECURE but is deprecated
+# and should not be used. CFG_WARN_INSECURE will be removed in a future OP-TEE
+# release.
+ifneq (,$(CFG_WARN_INSECURE))
+$(info CFG_WARN_INSECURE is deprecated, please use CFG_INSECURE instead)
+else
+ifneq ($(CFG_WARN_INSECURE),$(CFG_INSECURE))
```
jenswi-linaro, line 929: This doesn't seem very useful. If someone is using `CFG_WARN_INSECURE` in a downstream tree set to `n` we'll error out here. If the downstream tree doesn't touch `CFG_WARN_INSECURE` or otherwise leaves it at its default value we'll pass.

If the objective is to catch deprecated usage of `CFG_WARN_INSECURE`, then we should use "error" instead of "info" above.

mk/config.mk
etienne-lms: I pushed a buggy commit here.
Please re-review with the latest fixup commit applied.

mk/config.mk
etienne-lms: Ok, I'll simplify this to the below (I hope I understood your proposal):
```patch
 ifneq (,$(CFG_WARN_INSECURE))
-$(info CFG_WARN_INSECURE is deprecated, please use CFG_INSECURE instead)
+$(error CFG_WARN_INSECURE is deprecated, please use CFG_INSECURE instead)
-ifneq ($(CFG_WARN_INSECURE),$(CFG_INSECURE))
-$(error Deprecated CFG_WARN_INSECURE must match CFG_INSECURE)
-endif
 endif
```

mk/config.mk
```diff
@@ -909,15 +909,22 @@ CFG_REGULATOR_FIXED ?= n
 $(eval $(call cfg-enable-all-depends,CFG_REGULATOR_FIXED, \
 	 CFG_DRIVERS_REGULATOR CFG_DT))
 
-# The purpose of this flag is to show a print when booting up the device that
-# indicates whether the board runs a standard developer configuration or not.
+# When enabled, CFG_INSECURE permits insecure configuration of OP-TEE core
+# and shows a print (info level) when booting up the device that
+# indicates that the board runs a standard developer configuration.
+#
```
etienne-lms, line 915: maybe should I remove this added empty line

mk/config.mk
jforissier: I still believe that anyone setting `CFG_WARN_INSECURE` to disabled ("" or "n") would want `CFG_INSECURE` disabled, too, because they are confident that they platform is secure. Therefore `$(CFG_WARN_INSECURE)` is a good default for `CFG_INSECURE` -- again, assuming it is set to an empty value or "n". If it is not set then "y" should be the default.
Having a warning about deprecation is probably a good idea too. So let's recap:
```
ifneq (undefined,$(flavor CFG_WARN_INSECURE))
$(warn CFG_WARN_INSECURE is deprecated, use CFG_INSECURE instead)
CFG_INSECURE ?= $(CFG_WARN_INSECURE)
else
CFG_INSECURE ?= y
endif
```

etienne-lms: The issue I see with `CFG_INSECURE ?= $(CFG_WARN_INSECURE)` is the case where one mistakenly sets `CFG_INSECURE` and `CFG_WARN_INSECURE` to different values. When so, I think we should simply trigger a build error.

Triggering an build error when `CFG_WARN_INSECURE` is set, whatever the value, is a another (maybe rude) way to ensure consistency of these insecure OP-TEE core config directives.
`

etienne-lms: @jenswi-linaro, @jforissier, new proposal:
```c
ifneq (undefined,$(flavor CFG_WARN_INSECURE))
$(info CFG_WARN_INSECURE is deprecated, use CFG_INSECURE instead)
CFG_INSECURE ?= $(CFG_WARN_INSECURE)
# If one of them is enabled, both must match
ifeq (y,$(filter y,$(CFG_INSECURE) $(CFG_WARN_INSECURE)))
ifneq ($(CFG_INSECURE),$(CFG_WARN_INSECURE)
$(error Inconsistent CFG_INSECURE=$(CFG_INSECURE) and CFG_WARN_INSECURE=$(CFG_WARN_INSECURE))
endif
endif
endif # CFG_WARN_INSECURE defined
CFG_INSECURE ?= y

Comment (jforissier) [reply mk/config.mk]: > @jenswi-linaro, @jforissier, new proposal:

LGTM

Comment (jenswi-linaro) [reply mk/config.mk]: It looks a bit complicated, how about:
```
ifneq (undefined,$(flavor CFG_WARN_INSECURE))
$(info CFG_WARN_INSECURE is deprecated, use CFG_INSECURE instead)
CFG_INSECURE ?= $(CFG_WARN_INSECURE)
ifneq ($(CFG_INSECURE),$(CFG_WARN_INSECURE)
$(error Inconsistent CFG_INSECURE=$(CFG_INSECURE) and CFG_WARN_INSECURE=$(CFG_WARN_INSECURE))
endif
endif # CFG_WARN_INSECURE defined
CFG_INSECURE ?= y
```

File: mk/config.mk (hunk 4 above)
Comment (etienne-lms) [reply mk/config.mk]: close this discussion thread, superseded by discussion thread related to https://github.com/OP-TEE/optee_os/pull/6457#discussion_r1392595513.

File: mk/config.mk (hunk 5 above)
Comment (etienne-lms) [reply mk/config.mk:915]: I'll discarded the comment. I found it easier to read with the 2 paragraph split with this way.

**Comment by etienne-lms** [comment]
Comment addressed.

File: mk/config.mk [hunk 6]
```diff
@@ -909,15 +909,26 @@ CFG_REGULATOR_FIXED ?= n
...
 # A developer configuration doesn't necessarily has to be secure. The intention
```
jenswi-linaro, line 916: While you're at it, please change "has" to "have".

jenswi-linaro (commented):

jforissier:

etienne-lms:
> While you're at it, please change "has" to "have".

Typo fixed.
Review tags applied (@jforissier, I applied your R-b tag only on the generic change).
CI checkpatch failures are related to the same trace message size.
Thanks for your help.

PR #6459, opened 2023-11-14
core: ffa: fix mem reclaim error reporting

jenswi-linaro:
Until now handle_mem_reclaim() was incorrectly returning the error value in a1 instead of a2 as mandated by the specification. Successful returns are not affected by this since they use the FFA_SUCCESS_32 FID. So fix this by supplying the error value in the right register.

Fixes: 1b302ac09816 ("core: enable FF-A with SPM Core at S-EL1")


etienne-lms (commented):
It seems to be the sole occurrence of register swap in error return arguments.

jenswi-linaro:
> It seems to be the sole occurrence of register swap in error return arguments.

Thanks for double-checking. Tag applied.

PR #6462, opened 2023-11-14
core: arm: fix inline comment on async notif interrupt

etienne-lms:
Fixes the inline comment that describes allowed values for CFG_CORE_ASYNC_NOTIF_GIC_INTID that can be a SPI or a secure PPI.

Fixes: 943972855082 ("core: notif: allow GIC_PPI usage for async notif")


jenswi-linaro (commented):

etienne-lms:
tag applied. thanks.

PR #6465, opened 2023-11-15
plat-stm32mp1: scmi_server: fix case when regulator state is unchanged

etienne-lms:
Fixes the return value when an SCMI regulator is already in the expected state. Prior this change the implementation returned SCMI_GENERIC_ERROR instead of SCMI_SUCCESS.

Changes also the SCMI return code in case of failure from SCMI_GENERIC_ERROR to SCMI_INVALID_PARAMETERS when the requested state is not one of the 2 supported SCMi voltage domain states supported (SCMI_VOLTAGE_DOMAIN_CONFIG_ARCH_ON or
SCMI_VOLTAGE_DOMAIN_CONFIG_ARCH_OFF).

Fixes: 23e200628dad ("plat-stm32mp1: scmi_server: use registered regulators")

### DIFF
--- a/core/arch/arm/dts/stm32mp157c-ed1.dts
+++ b/core/arch/arm/dts/stm32mp157c-ed1.dts
@@ -253,6 +253,7 @@
 			vdd_usb: ldo4 {
 				regulator-name = "vdd_usb";
 				interrupts = <IT_CURLIM_LDO4 0>;
+				regulator-always-on;
 			};
 
 			vdd_sd: ldo5 {
--- a/core/arch/arm/dts/stm32mp15xx-dkx.dtsi
+++ b/core/arch/arm/dts/stm32mp15xx-dkx.dtsi
@@ -360,6 +360,7 @@
 			vdd_usb: ldo4 {
 				regulator-name = "vdd_usb";
 				interrupts = <IT_CURLIM_LDO4 0>;
+				regulator-always-on;
 			};
 
 			vdda: ldo5 {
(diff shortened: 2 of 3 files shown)


core/arch/arm/dts/stm32mp157c-ed1.dts
```diff
@@ -253,6 +253,7 @@
 			vdd_usb: ldo4 {
 				regulator-name = "vdd_usb";
 				interrupts = <IT_CURLIM_LDO4 0>;
+				regulator-always-on;
```
jforissier, line 256: The DT changes should probably go in a separate commit.
I did not quite understand the explanation about this regulator in the commit description :-/ perhaps you could reword it?

jforissier (commented):
- For "plat-stm32mp1: scmi_server: fix regulator framework integration": see comments below
- For "plat-stm32mp1: scmi_server: report invalid regulator state request":

core/arch/arm/dts/stm32mp157c-ed1.dts
etienne-lms: Ok, i'll try :) sorry for my bad phrasing.

etienne-lms:
Thanks for the feedback. I have split the commit in 2 and rephrase the commit message. It's far more verbose now. I hope it clear. Feel free to complain if not.

I have not applied your review tag. I'll do once you're ok with the description.

jforissier:
@etienne-lms crystal clear! Thanks for the details. For all commits:

etienne-lms:
Tag applied.

PR #6469, opened 2023-11-17
core: arm: allow CFG_TZSRAM_START being defined when pager is disabled

etienne-lms:
Fixes case when a platform configuration defines CFG_TZSRAM_START but does not use the pager. CFG_TZSRAM_START defines the based address of the memory used for resident memory and page pool when CFG_WITH_PAGER is enabled.

Since below mentioned commit, TZSRAM_BASE being defined makes core_mmu.c to assume there are 2 secure memories for OP-TEE core internal use. This change ensures that when CFG_WITH_PAGER is disabled, TZSRAM is not defined even if the platform configuration sets CFG_TZSRAM_START.

An example of such issues is when testing an STM32MP15 variant of platform stm32mp1 with pager being disabled. Before this change, OP-TEE boot sequence fails with a error trace message like: E/TC:0 0 Panic 'Unexpected TZC configuration on secure region' at core/arch/arm/plat-stm32mp1/plat_tzc400.c:102 <init_stm32mp1_tzc>

Indeed debug trace messages can show that an invalid physical memory area has been registered by core as TEE_RAM_RO, as shown below. Note that for that platform, internal secure SYSRAM range is [0x2ffc000 0x30000000]:
D/TC:0   add_phys_mem:667 ram_start type TEE_RAM_RO 0x2ffc0000 size 0xae040000

Fixes: e09739a8a6a1 (core: core_mmu.c: use secure_only[] where possible")

### DIFF
--- a/core/arch/arm/include/mm/generic_ram_layout.h
+++ b/core/arch/arm/include/mm/generic_ram_layout.h
@@ -133,7 +133,7 @@
 #endif
 #endif
 
-#if defined(CFG_TZSRAM_START)
+#if defined(CFG_WITH_PAGER) && defined(CFG_TZSRAM_START)
 #define TZSRAM_BASE		CFG_TZSRAM_START
 #define TZSRAM_SIZE		CFG_TZSRAM_SIZE
 #endif


jforissier:
With the checkpatch complaint addressed:

(and thanks for the detailed explanation)

etienne-lms:
`Fixes:` tag fixed
Tag applied
Thanks

jenswi-linaro:

etienne-lms:
Review tags applied

PR #6470, opened 2023-11-17
Core: Introduce stm32mp remoteproc PTA and driver

arnopo:
This series introduces the stm32mp remoteproc PTA and the stm32_remoteproc driver , which enable the authentication of signed images for the stm32mp157 platform.

1) The remote proc TA implements platform-specific requirements associated to the remoteproc PTA API introduced in  https://github.com/OP-TEE/optee_os/pull/6300. The API commit is also part of this PR for build purposes, and should be removed if https://github.com/OP-TEE/optee_os/pull/6300 is merged.
 
2) The stm32mp_remoteproc driver is responsible for relying on the device tree, which is very similar to the Linux one. Note that the remoteproc node is a sub-node of the AHB simple bus. The AHB bus defines the dma-range property that specifoes memory translation between the Cortex-M4 mapping and the Cortex-A7 mapping. This is necessary to find the address translation between the address found in the ELF file and the address to load the segment. For instance, the RETRAM address is 0 from the Cortex-M4 perspective and 0x38000000 from the Cortex-A7 perspective.

Related pull request:

- https://github.com/OP-TEE/optee_os/pull/6027

- https://github.com/OP-TEE/optee_os/pull/6300

- Code associated to the Linux Kernel is proposed as RFC for upstream:


etienne-lms (commented):
Commit "drivers: Add stm32mp1 remoteproc driver" should define a default value for `CFG_DRIVERS_REMOTEPROC` and provide an inline description comment.

Ditto for commit "pta: stm32mp: add new remoteproc pta" regarding `CFG_RPROC_PTA`.

My main concern here is detailed in my review comment on `stm32_rproc_get()` function.

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,389 @@
...
+ * @firmware_id:	identifies the remoteproc firmware to load
+ */
+struct stm32_rproc_compat_data {
+	uint32_t firmware_id;
+};
+
+static SLIST_HEAD(, stm32_rproc_instance) rproc_list =
+		SLIST_HEAD_INITIALIZER(rproc_list);
+
+void *stm32_rproc_get(uint32_t firmware_id)
```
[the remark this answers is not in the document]
arnopo: If I understand correctly @etienne-lms, the question is whether the UUID should guarantee the uniqueness of the service (represented by the PTA API) or the implementation of the PTA itself.
If it guarantees the service, then PTA can be platform dependent. 
If it guarantees the implementation, then the PTA should be generic and we may need to introduce another layer to address specific details that may be too high-level for a driver. 

For instance we could have  to support:
-  management of  a proprietary  image format,
-  specific management of a key to authenticate or decrypt,
-  management of metadata associated to a firmware that could be provided by TLV ,
-  core dump on crash

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,356 @@
...
+#define PTA_NAME	"remoteproc.pta"
+
+/*
+ * UUID of the remoteproc Trusted application authorized to communicate with
+ * the remoteproc pseudo TA
+ */
+
+#define TA_REMOTEPROC_UUID \
+	{ 0x80a4c275, 0x0a47, 0x4905, \
+		{ 0x82, 0x85, 0x14, 0x86, 0xa9, 0x77, 0x1a, 0x08} }
```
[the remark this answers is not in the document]
etienne-lms: You're right, TA's API header file are not exported.
Keep as-is.

core/drivers/remoteproc/stm32_remoteproc.c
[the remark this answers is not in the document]
jforissier: I would say the current implementation LGTM. I believe we should consider the remoteproc PTA UUID identifies a service that the remoteproc TA relies on, and that the PTA implementation can be platform specific. I don't see too much code re-use being made possible by introducing another abstraction layer. IOW the PTA interface is probably the right level for separation between generic code (remoteproc TA) and platform-specific code (remoteproc PTA and driver).

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -174,14 +174,14 @@ TEE_Result stm32_rproc_map(uint32_t firmware_id, paddr_t pa, size_t size,
...
 }
 
-TEE_Result stm32_rproc_unmap(uint32_t firmware_id, paddr_t pa, size_t size)
+TEE_Result stm32_rproc_unmap(uint32_t rproc_id, vaddr_t *va, size_t size)
 {
-	struct stm32_rproc_instance *rproc = stm32_rproc_get(firmware_id);
+	struct stm32_rproc_instance *rproc = stm32_rproc_get(rproc_id);
 	struct stm32_rproc_mem *mems = NULL;
-	vaddr_t *va = NULL;
+	paddr_t pa  = virt_to_phys(va);
```
etienne-lms, line 181: could you replace the 2 space chars with a single one?

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -174,14 +174,14 @@ TEE_Result stm32_rproc_map(uint32_t firmware_id, paddr_t pa, size_t size,
 	return TEE_ERROR_ACCESS_DENIED;
 }
 
-TEE_Result stm32_rproc_unmap(uint32_t firmware_id, paddr_t pa, size_t size)
+TEE_Result stm32_rproc_unmap(uint32_t rproc_id, vaddr_t *va, size_t size)
```
etienne-lms, line 177: this passes a va or a pointer to a va?
I think it should be either `void *va` or `vaddr_t va` here.

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -174,14 +174,14 @@ TEE_Result stm32_rproc_map(uint32_t firmware_id, paddr_t pa, size_t size,
...
 {
-	struct stm32_rproc_instance *rproc = stm32_rproc_get(firmware_id);
+	struct stm32_rproc_instance *rproc = stm32_rproc_get(rproc_id);
 	struct stm32_rproc_mem *mems = NULL;
-	vaddr_t *va = NULL;
+	paddr_t pa  = virt_to_phys(va);
 	unsigned int i = 0;
 
-	if (!rproc)
+	if (!rproc || !va)
```
etienne-lms, line 184: `!va` or `!pa`  (note: pa will be 0 if va is 0)

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,389 @@
...
+			goto err;
+		}
+
+		DMSG("register region %#"PRIxPA" size %#zx",
+		     regions[i].addr, regions[i].size);
+
+		rproc->n_regions++;
+	}
+
+	rproc->regions = regions;
```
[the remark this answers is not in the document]
etienne-lms: indeed, `n_regions` does not change. my bad. discard my comment.

core/pta/stm32mp/remoteproc_pta.c
```diff
@@ -0,0 +1,356 @@
...
+static TEE_Result
+	rproc_pta_open_session(uint32_t pt, TEE_Param params[TEE_NUM_PARAMS],
+			       void **sess_ctx __unused)
+{
+	struct ts_session *s = ts_get_calling_session();
+	const uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE);
+	struct ts_ctx *ctx;
```
etienne-lms, line 333: = NULL

etienne-lms (commented):
I have few comments on the fixup commits. I'm fine if you squash and rebase once addressed.

arnopo (commented):
As requested by @etienne-lms, I have updated the PR based on comment, squashed the fixups  and rebased the PR.
On top of that I  have added a new fixup that adds additional  fixes.

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,388 @@
...
+
+	if ((len % (sizeof(uint32_t) * 3)))
+		return TEE_ERROR_GENERIC;
+
+	nranges = len / sizeof(uint32_t);
+
+	for (i = 0; i < nranges; i++) {
+		uint32_t da = fdt32_to_cpu(list[i++]);
+		uint32_t pa = fdt32_to_cpu(list[i++]);
+		uint32_t size = fdt32_to_cpu(list[i]);
```
etienne-lms, line 242: nitpicking: here I would prefer:
```c
	for (i = 0; i < nranges; i += 3) {
		uint32_t da = fdt32_to_cpu(list[i]);
		uint32_t pa = fdt32_to_cpu(list[i + 1]);
		uint32_t size = fdt32_to_cpu(list[i + 2]);
		(...)

File: core/pta/stm32mp/remoteproc_pta.c [hunk 9]
```diff
@@ -0,0 +1,356 @@
...
+		return TEE_ERROR_NOT_IMPLEMENTED;
+	}
+}
+
+/*
+ * Pseudo Trusted Application entry points
+ */
+static TEE_Result
+	rproc_pta_open_session(uint32_t pt, TEE_Param params[TEE_NUM_PARAMS],
+			       void **sess_ctx __unused)
```
etienne-lms, line 326: prefer:
```c
static TEE_Result rproc_pta_open_session(uint32_t pt,
					 TEE_Param params[TEE_NUM_PARAMS],
					 void **sess_ctx __unused)

**Review by etienne-lms — COMMENTED** [review]
"drivers: Add stm32mp1 remoteproc driver" and
"pta: stm32mp: add new remoteproc pta" with a comment in each addressed.

**Comment by arnopo** [comment]
> `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits "drivers: Add stm32mp1 remoteproc driver" and "pta: stm32mp: add new remoteproc pta" with a comment in each addressed.

Done, thanks @etienne-lms for the reviews!

File: mk/config.mk [hunk 10]
```diff
@@ -1132,3 +1132,10 @@ $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
 CFG_HALT_CORES_ON_PANIC ?= n
 CFG_HALT_CORES_ON_PANIC_SGI ?= 15
 $(eval $(call cfg-depends-all,CFG_HALT_CORES_ON_PANIC,CFG_GIC))
+
+# CFG_DRIVERS_REMOTEPROC, when enabled, embeds support for remote processor
+# management including generic DT bindings for the configuration.
+CFG_DRIVERS_REMOTEPROC ?= n
+
+# CFG_RPROC_PTA, when enabled, embeds remote processor management PTA service.
+CFG_RPROC_PTA ?= n
```
jforissier, line 1141: Would you mind `s/CFG_RPROC_PTA/CFG_REMOTEPROC_PTA/` for consistency with `CFG_DRIVERS_REMOTEPROC` (and better "grepability", too)?

core/drivers/remoteproc/stm32_remoteproc.c
```diff
@@ -0,0 +1,388 @@
...
+	 * the hardware.
+	 */
+	res = rstctrl_deassert_to(rproc->hold_boot, TIMEOUT_US_1MS);
+	if (!res)
+		res = rstctrl_assert_to(rproc->hold_boot, TIMEOUT_US_1MS);
+
+	return res;
+}
+
+static TEE_Result __stm32_rproc_stop(struct stm32_rproc_instance *rproc)
```
jforissier, line 96: Leading underscores/double underscores are not really recommended, why not simply `rproc_stop()`? It is a static function the name can remain very short and generic with no risk of collision.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`for commits
"plat-stm32mp1: Add the remoteproc TA in early-TA list" and
"dts: stm32: update m4_rproc to support the remoteproc OPTEE framework".

jforissier (commented):
Minor comments on "drivers: Add stm32mp1 remoteproc driver".

s/coherence/consistency/ in the commit description. Other comments below.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -151,6 +151,7 @@ CFG_CORE_ASLR ?= n
 
 CFG_STM32MP_REMOTEPROC ?= n
 CFG_DRIVERS_REMOTEPROC ?= $(CFG_STM32MP_REMOTEPROC)
+CFG_RPROC_PTA ?= $(CFG_STM32MP_REMOTEPROC)
```
jforissier, line 154: As already mentioned I would prefer `CFG_REMOTEPROC_PTA`

jforissier (commented):
For commit " pta: stm32mp: add new remoteproc pta"

s/pta/PTA/ in the suject (the `pta:` prefix is OK in lowercase).

jforissier (commented):
For "plat-stm32mp1: Add the remoteproc TA in early-TA list":

s/early-TA/early TA/

core/drivers/remoteproc/stm32_remoteproc.c
arnopo: Done

core/arch/arm/plat-stm32mp1/conf.mk
arnopo: Done

mk/config.mk
arnopo: yes much better  I also move the declaration in the patch that introduce the remoteproc PTA

arnopo (commented):
@jforissier : I have take, into account your comments. In addition I have re-dispatched the CFG-REMOTEPROC-PTA between commits. As consequence, I don't add your tag and @etienne-lms one in commits.
Please, could you confirm that the fixups are ok for you then i will squash them add the tags in commits messages.

jforissier:
@arnopo yes the fixups look good, thanks. When squashing you may also add:

... to "drivers: Add stm32mp1 remoteproc driver".

arnopo:
+1 update to fix a typo in commit adding the acked-by

PR #6471, opened 2023-11-17
PKCS11TA: Transient Token support

### LINKED ISSUE #6402: PKCS11 TA: use of persistent objects vs transient objects
Hi all,

I have a use case in mind where the bootloader (responsible for starting OPTEE) passes on secrets (e.g. found in SOTP) to OPTEE (e.g. through (secure?) device tree) and OPTEE (or another (custom) TA) fetching the keys from the device tree and injecting them into the PKCS11 TA. Those keys would then be stored as "Transient Objects", ensuring that they're not stored anywhere persistently and unavailable when the device is off. 

I've been going through the documentation and the code for the PKCS11 Trusted Application. The PKCS11 TA currently has a fixed (by CFG_PKCS11_TA_TOKEN_COUNT) amount of tokens stored as persistent objects in OPTEE. With the amount of tokens being statically defined and those tokens being Persistent Objects by default, I don't see any room for injecting "run-time keys" into the PKCS11 TA.

I was wondering if the use-case I have is one that (1) complies with the specifications, (2) is sound from a security perspective and (3) if you would be open to receiving code contributions that make this use case possible.

Kind regards,
Lars Feyaerts

fifalars:
Add the ability for a Transient Token. More information is the last commit message. The other commits are for preparing the adding of Transient Tokens and shouldn't change the existing functionality.

closes #6402

### DIFF
--- a/ta/pkcs11/src/object.c
+++ b/ta/pkcs11/src/object.c
@@ -17,6 +17,7 @@
 #include "pkcs11_token.h"
 #include "sanitize_object.h"
 #include "serializer.h"
+#include "processing.h"
 
 /*
  * Temporary list used to register allocated struct pkcs11_object instances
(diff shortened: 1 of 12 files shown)


jbech-linaro:
IBART reports quite a few PKCS#11 related errors when running xtest with the patches in this PR.

fifalars:
Yes I will fix those in the coming weeks, I just wanted to share this concept to get feedback for it.

vesajaaskelainen:
I suppose let's start with discussion in issue tracker first.

BeejShah:
@fifalars This is very interesting, particularly around the use of transient objects. I have a similar use-case but the material I need to "load" into a transient object is somewhat more complex due to the nature of the data we have stored. Can the transient object be populated via another TA or pseudo TA? If so, that would make the whole system more flexible, no?

PR #6476, opened 2023-11-20
mk: config: clarify CFG_VIRTUALIZATION is deprecated

etienne-lms:
Ensures that CFG_VIRTUALIZATION and CFG_NS_VIRTUALIZATION configuration settings do not conflict. If Both are set, they shall have the same value.

Clarifies CFG_VIRTUALIZATION is deprecated in mk/config.mk inline comments.


jenswi-linaro (commented):
In the commit message `s/If Both/if both/`
With that fixed, please apply:

etienne-lms:
Commit log fixed & review tag applied.

PR #6478, opened 2023-11-20
core: kernel: tee_ta_manager.c: add uuid in open session error trace

etienne-lms:
Adds the TA UUID in open session error trace to allow to identify witch TA cause the issue when debug trace are not acitvated.

By the way, fix specifier for res argument that is a uint32_t.


jenswi-linaro (commented):

etienne-lms:
Review tag applied.

PR #6480, opened 2023-11-20
core: static mapping of non-secure DDR and of secure memory with read-only rights

etienne-lms:

### DIFF
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -471,6 +471,7 @@ void core_mmu_set_discovered_nsec_ddr(struct core_mmu_phys_mem *start,
 			break;
 		case MEM_AREA_EXT_DT:
 		case MEM_AREA_MANIFEST_DT:
+		case MEM_AREA_RAM_NSEC:
 		case MEM_AREA_RES_VASPACE:
 		case MEM_AREA_SHM_VASPACE:
 		case MEM_AREA_TS_VASPACE:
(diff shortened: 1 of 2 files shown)


core/mm/core_mmu.c
```diff
@@ -628,9 +628,18 @@ static void verify_special_mem_areas(struct tee_mmap_region *mem_map,
...
+	 *
+	 * Exceptions are the memory areas that maps with the same attributes
+	 * as for example MEM_AREA_RAM_NSEC and MEM_AREA_NSEC_SHM
+	 * which may overlap since they are used for the same purpose
+	 * except that MEM_AREA_NSEC_SHM is always mapped and
+	 * MEM_AREA_RAM_NSEC only uses a dynamic mapping.
 	 */
 	for (mem = start; mem < end; mem++) {
 		for (mmap = mem_map; mmap->type != MEM_AREA_END; mmap++) {
+			if (core_mmu_type_to_attr(mem->type) ==
```
jenswi-linaro, line 641: I'd rather be explicit about which memory area we allow to overlap. The purpose here is to check against unintended overlapping, so let's make it easy to keep track of the allowed exceptions.
Currently, it seems it's only with `MEM_AREA_RAM_NSEC` and `MEM_AREA_NSEC_SHM` where it makes sense to allow an overlap.

etienne-lms: fair. i'll update accordingly.

etienne-lms: I've remove that ch

jenswi-linaro:

Nit: Should the SoBs in the last commit be swapped?

etienne-lms:
I've swapped theS-o-b tags in commit "core: mmu: fix nsec ddr discovery regarding MEM_AREA_RAM_NSEC" and applied the review tag to the 3 commits.
Thanks for the review.

PR #6474, opened 2023-11-17
drivers: clk: print clock tree summary

etienne-lms:
Adds clk_print_summary() to print the clock tree current state on core console using the info trace level. Clock framework spinlock is help while clock tree is printed.

PR #6481, opened 2023-11-21
drivers: regulator: print regulator tree summary

etienne-lms:
Replaces existing `regulator_print_state()` with `regulator_print_summary()` on the model of what's been done for clocks (https://github.com/OP-TEE/optee_os/pull/6474).
The function is renamed for consistency.

### DIFF
--- a/core/drivers/regulator/regulator.c
+++ b/core/drivers/regulator/regulator.c
@@ -288,26 +288,6 @@ TEE_Result regulator_register(struct regulator *regulator)
 }
 
 /*
- * Log regulators state
- */
-void regulator_print_state(const char *message __maybe_unused)
-{
-	struct regulator *regulator = NULL;
-
-	DMSG("Regulator state: %s", message);
-	DMSG("name     use\ten\tuV\tmin\tmax\tflags\tsupply");
-
-	SLIST_FOREACH(regulator, &regulator_device_list, link)
-		DMSG("%8s %u\t%d\t%d\t%d\t%d\t%#x\t%s",
-		     regulator->name, regulator->refcount,
-		     regulator_is_enabled(regulator),
-		     regulator_get_voltage(regulator),
-		     regulator->min_uv, regulator->max_uv, regulator->flags,
-		     regulator->supply ? regulator_name(regulator->supply) :
-		     "<none>");
-}
-
-/*
  * Clean-up regulators that are not used.
  */
 static TEE_Result regulator_core_cleanup(void)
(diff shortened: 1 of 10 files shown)


etienne-lms:
Rebased

etienne-lms:
I've appended a commit for adding a command in the statistics PTA to allow non-secure to request print of clock & regulator trees on the secure console: see commit _"pta: stats: add commands to print clock and regulator trees"_.

CI / make (multi-platform) test failure should be fixed with https://github.com/OP-TEE/optee_os/pull/6484.

core/pta/stats.c
```diff
@@ -51,6 +53,16 @@
 
 #define STATS_NB_POOLS			4
 
+/*
+ * STATS_CMD_PRINT_DRIVERS_INFO - Print device drivers information to console
+ *
+ * [in]    value[0].a        Target driver, one of STATS_DRIVER_TYPE_*
+ */
+#define STATS_CMD_PRINT_DRIVERS_INFO	5
```
jforissier, line 61: `_DRIVER_` (singular)? since the command prints for one driver

core/pta/stats.c
```diff
@@ -209,6 +221,28 @@ static TEE_Result get_system_time(uint32_t type,
 	return TEE_SUCCESS;
 }
 
+static TEE_Result get_drivers_info(uint32_t type, TEE_Param p[TEE_NUM_PARAMS])
```
jforissier, line 224: `print_driver_info()`. This does not get anything.
...which leads to the question: should a buffer be returned to the client instead? That would be more consistent with other statistics and also would be usable when `CFG_TEE_CORE_LOG_LEVEL < TRACE_INFO`.

etienne-lms: Right, I'll change the function label.
Note that command `STATS_CMD_MEMLEAK_STATS` already behaves this way.

etienne-lms:
I've addressed the several CI make error and address another potential build error: cases when `CFG_DRIVERS_{CLK|REGUALTOR}` is disabled..
I've rebased on master branch tip to leverage https://github.com/OP-TEE/optee_os/pull/6484 fix.

core/pta/stats.c
etienne-lms: ok, makes sense. I'll address that.

core/pta/stats.c
jforissier: > Note that command `STATS_CMD_MEMLEAK_STATS` already behaves this way.

Fair enough. Printing to the secure console is probably good enough for the main use cases (debugging).

core/drivers/regulator/regulator.c
```diff
@@ -323,9 +303,116 @@ static TEE_Result regulator_core_cleanup(void)
...
+static void print_regulator_subtree(struct regulator *root __maybe_unused,
+				    int indent __maybe_unused)
+{
+#ifdef CFG_DRIVERS_REGULATOR_PRINT_TREE
+	struct regulator *regulator = NULL;
+
+	SLIST_FOREACH(regulator, &regulator_device_list, link) {
+		if (regulator->supply == root) {
+			print_regulator(regulator, indent + 1);
+			print_regulator_subtree(regulator, indent + 1);
```
jenswi-linaro, line 403: Would it be possible to do this without recursion?

jenswi-linaro: Yes, that's why recursion is discouraged. A comment explaining why this will not be a problem here might be sufficient. Asserting that `indent` isn't too large could also be an option.

core/drivers/regulator/regulator.c
```diff
@@ -323,9 +303,116 @@ static TEE_Result regulator_core_cleanup(void)
...
+			      level_max, level_unit[n_max - 1]);
+
+out:
+	if (!msg)
+		snprintf(msg_end - 4, 4, "...");
+
+	IMSG("%s", msg_buf);
+}
+
+static void print_regulator_subtree(struct regulator *root __maybe_unused,
```
jenswi-linaro, line 394: Wouldn't `parent` be a more suitable name than `root`?

etienne-lms: true. I'll change that.

etienne-lms:
Regarding CI make test failure, i've created an Issue entry to get some help.

core/drivers/regulator/regulator.c
etienne-lms: I'm ok with setting an upper limit to `indent`. Implementing this without recursion looks a bit painfull.
I'll append a fixup commit and add the same to `clk_print_tree()`.

etienne-lms:
Comments addressed.

I've created a specific P-r (https://github.com/OP-TEE/optee_os/pull/6481) for the variable sized field issue reported by Clang. I picked the commit in this series to have CI test running (hopefully) without this issue.

Regarding the recursive call to print the regulator tree, I've addressed the same concern in existing `clk_print_tree()` with P-R https://github.com/OP-TEE/optee_os/pull/6489.

core/drivers/regulator/regulator.c
jenswi-linaro: > Implementing this without recursion looks a bit painfull.

Oh, that sounds almost like a challenge. ;-)
You have parent pointers in the nodes so there's no need to save a list of pointers to find the path back to the root.

etienne-lms: It would be a bit of code to create the ascending tree of devices. Do you think it's worth it? This recursion is less than 10 lines, it like it. That said, I really understand your concern.

etienne-lms: I tried but I always end up with some sort of recursive calls to descend the tree to print the clock info.
The sole way out would be to allocate some big reference matrix to track per item / per level state of what has been printed so far. I found that quite ugly.

I would be in favor for the recursive calls, bounded to 16 levels (32 seems to high from the trace found on clock: https://github.com/OP-TEE/optee_os/pull/6489#issuecomment-1824730803).

jenswi-linaro: This should do the trick:
``` C
static struct regulator *find_next_regulator(struct regulator *parent,
					     struct regulator *sibling)
{
	struct regulator *r = NULL;

	if (sibling)
		r = SLIST_NEXT(sibling, link);
	else
		r = SLIST_FIRST(&regulator_device_list);

	while (r && r->supply != parent)
		r = SLIST_NEXT(r, link);

	return r;
}

static void regulator_print_tree(void)
{
	struct regulator *regulator = NULL;
	struct regulator *parent = NULL;
	struct regulator *next = NULL;
	int indent = -1;

	while (true) {
		next = find_next_regulator(parent, regulator);
		if (next) {
			print_regulator(next, indent + 1);
			/* Enter the subtree of the next regulator */
			parent = next;
			indent++;
			regulator = NULL;
		} else {
			/*
			 * We've processed all children at this level.
			 * If parent is NULL we're at the top and are done.
			 */
			if (!parent)
				break;
			/*
			 * Move up one level to resume with the next
			 * regulator of the parent.
			 */
			regulator = parent;
			parent = parent->supply;
			indent--;
		}
	}
}
```

etienne-lms:
I think i've put too many things in this P-R (regulator tree, regulator fixup, clock tree fixup, print trees from PTA stats).
I'll leave this P-R as is for the moment and create dedocated P-Rs per topic:
- regulator tree: see https://github.com/OP-TEE/optee_os/pull/6498
- regulator fixup regarding variable sized field: see https://github.com/OP-TEE/optee_os/pull/6487
- clock tree fixes: see https://github.com/OP-TEE/optee_os/pull/6497
- print trees from a statistics PTA command: TODO

PR #6483, opened 2023-11-22
dts: stm32mp15: harden RCC secure configuration on ST boards

etienne-lms:
Enable STM32MP15 RCC secure hardening configuration on ST boards (DK1, DK2, ED1 and EV1) to assign SoC clocks, reset controllers and PWR regulators to OP-TEE secure world.

This change removes setting of &rcc node status property from stm32mp157a-dk1.dts, stm32mp157c-dk2.dts as the property is set from stm32mp15xx-dkx.dtsi that is included from the 2 former DTS files.

This change depends on https://github.com/OP-TEE/build/pull/701 for the OP-TEE distribution.

### DIFF
--- a/core/arch/arm/dts/stm32mp157c-ed1.dts
+++ b/core/arch/arm/dts/stm32mp157c-ed1.dts
@@ -338,6 +338,7 @@
 };
 
 &rcc {
+	compatible = "st,stm32mp1-rcc-secure";
 	status = "okay";
 };
 
--- a/core/arch/arm/dts/stm32mp157a-dk1.dts
+++ b/core/arch/arm/dts/stm32mp157a-dk1.dts
@@ -36,7 +36,3 @@
 		reg = <0xf0 0x10>;
 	};
 };
-
-&rcc {
-	status = "okay";
-};
--- a/core/arch/arm/dts/stm32mp157c-dk2.dts
+++ b/core/arch/arm/dts/stm32mp157c-dk2.dts
@@ -98,10 +98,6 @@
 	};
 };
 
-&rcc {
-	status = "okay";
-};
-
 &usart2 {
 	pinctrl-names = "default", "sleep", "idle";
 	pinctrl-0 = <&usart2_pins_c>;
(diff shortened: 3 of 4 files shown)


core/arch/arm/dts/stm32mp157c-ed1.dts
```diff
@@ -338,6 +338,7 @@
 };
 
 &rcc {
+	compatible = "st,stm32mp1-rcc-secure";
```
jforissier, line 341: That's weird, why not `compatible = "st,stm32mp1-rcc";` and `status = "disabled"; secure-status = "okay";`?

etienne-lms: Because playing with secure-status / status is not nice in the long term (e.g virtualized optee or other packages e.g. tf-m). Hence we prefer a specific compatible. This compatible has already been upstreamed to Linux kernel DT bindings for U-Boot/Linux to comply with:
https://github.com/torvalds/linux/blob/v6.6/Documentation/devicetree/bindings/clock/st%2Cstm32mp1-rcc.yaml#L59

Note: we have planned to propose a firewall resource framework in for device driver to possibly grant or query SoC firewall config. We've not proposed anything yet in OP-TEE but that on the todo list. There are discusions on this topic in the LKML, see latest [patch v6 series for STM32 firewall framework](https://lore.kernel.org/lkml/20231010125719.784627-1-gatien.chevallier@foss.st.com/) based on [Oleksii' patch for domain-controller](https://lore.kernel.org/lkml/cover.1668070216.git.oleksii_moisieiev@epam.com/).

jforissier: > Because playing with secure-status / status is not nice in the long term (e.g virtualized optee or other packages e.g. tf-m). Hence we prefer a specific compatible.

I do not really see which difference it makes, it is still the same hardware whether or not it is accessed from secure or non-secure world (which `status` and `secure-status` are for). But if the bindings have been accepted already that's fine. Thanks for context.

jforissier:
FWIW:

etienne-lms:
@GseoC, could you look at this P-R?

GseoC:
Sure,

PR #6484, opened 2023-11-22
ci: sync SCP-firmware on v2.13.0 release tag

etienne-lms:
Synchronize SCP-firmware Git repository revision on v2.13.0 release tag to avoid issues with latest merged changes in SCP-firmware that OP-TEE OS may not be compliant with.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           function _make() { make -j$(nproc) -s O=out $*; }
           function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.4.2 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
 
-          function download_scp_firmware() { git clone --single-branch https://github.com/ARM-software/SCP-firmware.git $HOME/scp-firmware || echo Nervermind; }
+          function download_scp_firmware() { git clone --depth 1 https://github.com/ARM-software/SCP-firmware.git -b v2.13.0 $HOME/scp-firmware || echo Nervermind; }
 
           ccache -s -v
           download_plug_and_trust


.github/workflows/ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           function _make() { make -j$(nproc) -s O=out $*; }
           function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.4.2 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
 
-          function download_scp_firmware() { git clone --single-branch https://github.com/ARM-software/SCP-firmware.git $HOME/scp-firmware || echo Nervermind; }
+          function download_scp_firmware() { { git clone --single-branch https://github.com/ARM-software/SCP-firmware.git $HOME/scp-firmware && git -C $HOME/scp-firmware checkout v2.13.0; } || echo Nervermind; }
```
jforissier, line 85: You do not need the extra `{ }`. The syntax can be simplified as `function f() { cmd1 && cmd2 || cmd3; }`.

jforissier (commented):

...with one comment.

.github/workflows/ci.yml
etienne-lms, line 85: I think we could add `-n`/`--no-checkout` option to the `git clone` command and `-q`/`--quiet` to the `git checkout command`.

jforissier: Yes, and yes.

etienne-lms: I prefer the short forms `-n/-q` to reduce the line size. Tell me if you prefer the long arguments for everyone to easier understand the command line.

jforissier: `-n/-q` is fine IMO.

.github/workflows/ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           function _make() { make -j$(nproc) -s O=out $*; }
           function download_plug_and_trust() { mkdir -p $HOME/se050 && git clone --single-branch -b v0.4.2 https://github.com/foundriesio/plug-and-trust $HOME/se050/plug-and-trust || (rm -rf $HOME/se050 ; echo Nervermind); }
 
-          function download_scp_firmware() { git clone --single-branch https://github.com/ARM-software/SCP-firmware.git $HOME/scp-firmware || echo Nervermind; }
+          function download_scp_firmware() { git clone -n --single-branch https://github.com/ARM-software/SCP-firmware.git $HOME/scp-firmware && git -q -C $HOME/scp-firmware checkout v2.13.0 || echo Nervermind; }
```
etienne-lms, line 85: for maintenance, maybe you would prefer:
```
          export SCP_FIRMWARE_REVISION=v2.13.0
          (...)
          function download_scp_firmware() { git clone ... && git -q -C $HOME/scp-firmware checkout $SCP_FIRMWARE_REVISION || echo Nervermind; }

jforissier, line 85: In fact I believe the simplest/fastest/most efficient command would be `git clone --depth 1 https://github.com/ARM-software/SCP-firmware.git -b v2.13.0`

etienne-lms: `-b` works on a tag revision? If so, yes, it's preferable.

jforissier: No, let's keep things more compact.

etienne-lms: I'll update. The cmdline I wrote is bugged, fails on git checkout argument `-q` (not available on `git checkout`).

etienne-lms:
My hacky tests behaved as expected. I'll squash these fixup commits and apply you're review tag so that CI test are back to normal quick. Thanks for your reactivity and help.

jenswi-linaro (commented):

etienne-lms:
review tags applied
thanks

PR #6491, opened 2023-11-24
drivers: regulator: fix getting gpio for level control

etienne-lms:
Swap the 2 calls to find gpio property in "regulator-gpio" compiatble node. Obviously the first call expects to get the GPIO resource while the second check that the node does not define 2 or more GPIOs to control the regulator level as the implementation only supports one, as described in the inline comment.

Fixes: f164f0f83420 ("drivers: regulator: GPIO controlled regulator")

### DIFF
--- a/core/drivers/regulator/regulator_gpio.c
+++ b/core/drivers/regulator/regulator_gpio.c
@@ -148,6 +148,11 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 	void *gpio_ref = &gpio;
 	int len = 0;
 
+	res = dt_driver_device_from_node_idx_prop("gpios", fdt, node, 0,
+						  DT_DRIVER_GPIO, gpio_ref);
+	if (res)
+		return res;
+
 	/*
 	 * DT bindings allows more than 1 GPIO to control more than
 	 * 2 voltage levels. As it's not used so far in known platforms
@@ -160,11 +165,6 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 		EMSG("Multiple GPIOs not supported for level control");
 		return TEE_ERROR_GENERIC;
 	}
-
-	res = dt_driver_device_from_node_idx_prop("gpios", fdt, node, 0,
-						  DT_DRIVER_GPIO, gpio_ref);
-	if (res)
-		return res;
 
 	cuint = fdt_getprop(fdt, node, "states", &len);
 	if (!cuint || len != 4 * sizeof(fdt32_t)) {


etienne-lms:
I missed something during https://github.com/OP-TEE/optee_os/pull/6345 tests. I swapped 2 instructions and it looks like I forgot to test the implementation at some point.
@GseoC, could you have  a look at this change?

core/drivers/regulator/regulator_gpio.c
```diff
@@ -148,6 +148,11 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 	void *gpio_ref = &gpio;
 	int len = 0;
 
+	res = dt_driver_device_from_node_idx_prop("gpios", fdt, node, 0,
```
GseoC, line 151: This change should not solve an issue as it should have no impact.

If there's no second GPIO reference, then we'll get the GPIO reference and go on.
Whether we get it before or after checking the presence of a second GPIO reference should have no impact.

The issue probably comes from the fact that dt_driver_device_from_node_idx_prop() does not return TEE_ERROR_ITEM_NOT_FOUND in your case.

etienne-lms: You right, this change does not fix the issue for the reason I though it would. Thank you for pointing that: the commit message gives false information. However this change really fixes the issue.

Here, the issue was that when looking for a 2nd gpio phandle (the one of index 1), `dt_driver_device_from_node_idx_prop()` (_dt_driver.c_) returns `TEE_ERROR_DEFER_DRIVER_INIT`  because 1nd gpio phandle refers to a provider driver that has not been probed yet. As we expect the return code is ITEM_NOT_FOUND, we error on DEFER_DRIVER_INIT.

You may say that `dt_driver_device_from_node_idx_prop()` does not really need gpio providers are probed to get their `#gpio-cells` value, it could read straight in the DT node referred by the phandle. However, current implementation of `dt_driver_device_from_node_idx_prop()` is made so that is gets this value from registered driver provider ([see here](https://github.com/OP-TEE/optee_os/blob/4.0.0/core/kernel/dt_driver.c#L391)), not reading straight the DT node.

I'd rather keep  `dt_driver_device_from_node_idx_prop()` generic implementation as is.
A correction could be to not error on `TEE_ERROR_DEFER_DRIVER_INIT`:
```patch
 	/*
 	 * DT bindings allows more than 1 GPIO to control more than
 	 * 2 voltage levels. As it's not used so far in known platforms
 	 * this implementation is simplified to support only 2 voltage
 	 * levels controlled with a single GPIO.
 	 */
-	if (dt_driver_device_from_node_idx_prop("gpios", fdt, node, 1,
-						DT_DRIVER_GPIO, gpio_ref) !=
-	    TEE_ERROR_ITEM_NOT_FOUND) {
+	res =dt_driver_device_from_node_idx_prop("gpios", fdt, node, 1,
+						DT_DRIVER_GPIO, gpio_ref);
+	if (res == TEE_ERROR_DEFER_DRIVER_INIT)
+		return res;
+	if (res !=
 	    TEE_ERROR_ITEM_NOT_FOUND) {
 		EMSG("Multiple GPIOs not supported for level control");
 		return TEE_ERROR_GENERIC;
 	}

	res = dt_driver_device_from_node_idx_prop("gpios", fdt, node, 0,
						  DT_DRIVER_GPIO, gpio_ref);
	if (res)
		return res;
```

I think its far simpler to swap the calls so that when looking up for a 2nd gpio phandle, 1st one refers to a driver already registered in the DT_DRIVER provider list.

I'll fix the commit message to explain the right reason why this change fixes the issue. I'll try to be less verbose than I'm here :)

Thanks again for the feedback.

GseoC: Ok, thanks for the detailed explanation

GseoC:

etienne-lms:
@GseoC, I have rephrased the commit message and applied your review tag. Could you tell you is the message looks OK to you?

GseoC:
+1 :)

etienne-lms:
Review tag applied.

jforissier:
Merged manually with `s/gpio/GPIO/` in the commit subject. Thanks.

etienne-lms:
Thanks Jérôme.

Issue #6453, opened 2023-11-10
PKCS11 TA: Invalid attribute CKA_CHECK_VALUE for secure key object


maroueneboubakri:
Hello,

According to the PKCS#11 specifications, CKA_CHECK_VALUE is an attribute of certificate object (CKO_CERTIFICATE) and also secure key object (CKO_SECRET_KEY). It seems that the PKCS11 TA recognizes this only for certificate object, returing invalid attribute type when requested for a secure key object (CKR_ATTRIBUTE_TYPE_INVALID). It is intentional ? or a bug ? 

I'll open a PR.

@etienne-lms FYI.

Thanks
Maro

etienne-lms:
If you have an implementation for `CKA_CHECK_VALUE`, a P-R a warmly welcome. See https://github.com/OP-TEE/optee_os/issues/6431.

etienne-lms:
Imeplemented by https://github.com/OP-TEE/optee_os/pull/6494.

etienne-lms:
Work still on-going.
https://github.com/OP-TEE/optee_os/pull/6494 has been merged but led to 2 series of fixup changes:
- https://github.com/OP-TEE/optee_os/pull/6579 has been merged.
- https://github.com/OP-TEE/optee_os/pull/6550 is under review.

Issue #6431, opened 2023-11-03
PKCS11: Checksum Attribute Is Not Supported


moose-rivieh:
It seems the checksum attribute is not supported, and some common application requires it (example: Amazon Voice Assistance, see [this link](https://github.com/alexa/avs-device-sdk/blob/b32be0a1139be1aa22f1df5f59ba394d3b6a9697/core/Crypto/Pkcs11/src/PKCS11Key.cpp#L116) for reference). Are there plans to support this attribute? Thank you.

etienne-lms:
You mean `CKA_CHECK_VALUE`  attribute? Indeed it's not supported. I think it could be and should not be too hard to implement. I can add that to my todo list but contributions are warmly welcome, as always (I'm sorry to say it won't be in my top priority tasks).

moose-rivieh:
Yes. That is correct. I'd love to commit the change, but this is my first interaction with optee and PKCS11 so not really sure what I am doing. If you point me to some documentation, I am happy to look it over and see if I can add support for `CKA_CHECK_VALUE`

etienne-lms:
We'll be please to help you. It seems you need `CKA_CHECK_VALUE` for symmetric key objects.

As a start point, you can look he PKCS#11 base spec [latest 3.0 is fine](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html) section 4.10, last paragraph for for an overview of the attribute. 
Especially _"Unless otherwise specified for the object definition, the value of this attribute is derived from the key object by taking the first three bytes of an encryption of a single block of null (0x00) bytes, using the default cipher and mode (e.g. ECB) associated with the key type of the secret key object."_
For object of type `CKK_GENERIC_SECRET`: see PKCS#11 current mechanism spec [3.0](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/pkcs11-curr-v3.0.html) sec

moose-rivieh:
Thank you, @etienne-lms, for the detailed info. I got caught up with something urgent and just got time to get back to this. We have been able to address this issue on our end using a shortcut, but I am still going to review the info shared and attempt a pull request.

etienne-lms:
> It seems that CKA_CHECK_VALUE  can be created from template ? should this case be handled ? even if the template value would conflict with the exepcted/correct value ?

I think yes. When the attribute is supported, we should conform to the specification.

[...]

PR #6494, opened 2023-11-24
Add support of key check value CKA_CHECK_VALUE to PKCS11 TA

maroueneboubakri:
An attempt to support CKA_CHECK_VALUE attribute in PKCS 11 TA. For AES, the value of this attribute is derived from the key object by taking the first three bytes of the ECB encryption of a single block of null (0x00) bytes, using the default cipher associated with the key type of the secret key object. This is useful to detect that a key object has not tampered with. 

For now, we try to calculate the value when a symmetric key is randomly generated. Still need to add for (Create, Unwarp, Derive and Copy) in incoming commits.

Related issues: https://github.com/OP-TEE/optee_os/issues/6453, https://github.com/OP-TEE/optee_os/issues/6431

### DIFF
--- a/ta/pkcs11/src/object.c
+++ b/ta/pkcs11/src/object.c
@@ -352,6 +352,11 @@ enum pkcs11_rc entry_create_object(struct pkcs11_client *client,
 	if (rc)
 		goto out;
 
+	/* Set key check value attribute */
+	rc = set_check_value_attr(&head);
+	if (rc)
+		goto out;
+
 	/*
 	 * Check target object attributes match target processing
 	 * Check target object attributes match token state
@@ -1064,6 +1069,11 @@ enum pkcs11_rc entry_set_attribute_value(struct pkcs11_client *client,
 	if (rc)
 		goto out;
 
+	/* Set key check value attribute */
+	rc = set_check_value_attr(&obj->attributes);
+	if (rc)
+		goto out;
+
 	if (get_bool(obj->attributes, PKCS11_CKA_TOKEN)) {
 		rc = update_persistent_object_attributes(obj);
 		if (rc)
@@ -1213,6 +1223,11 @@ enum pkcs11_rc entry_copy_object(struct pkcs11_client *client, uint32_t ptypes,
 	 * given by the callee
 	 */
 	rc = modify_attributes_list(&head_new, head);
+	if (rc)
+		goto out;
+
+	/* Set key check value attribute */
+	rc = set_check_value_attr(&head_new);
 	if (rc)
 		goto out;
 
--- a/ta/pkcs11/src/pkcs11_attributes.c
+++ b/ta/pkcs11/src/pkcs11_attributes.c
@@ -4,6 +4,7 @@
  */
 
 #include <assert.h>
+#include <config.h>
 #include <inttypes.h>
 #include <mbedtls/asn1write.h>
 #include <mbedtls/ecp.h>
@@ -355,8 +356,12 @@ static const uint32_t pkcs11_certificate_boolprops[] = {
 };
 
 static const uint32_t pkcs11_certificate_optional[] = {
-	PKCS11_CKA_CERTIFICATE_CATEGORY, PKCS11_CKA_CHECK_VALUE,
-	PKCS11_CKA_START_DATE, PKCS11_CKA_END_DATE, PKCS11_CKA_PUBLIC_KEY_INFO,
+	PKCS11_CKA_CERTIFICATE_CATEGORY, PKCS11_CKA_START_DATE,
+	PKCS11_CKA_END_DATE, PKCS11_CKA_PUBLIC_KEY_INFO,
+#ifdef CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE
+	/* Consider KCV attribute only when supported */
+	PKCS11_CKA_CHECK_VALUE,
+#endif
 };
 
 /*
@@ -404,6 +409,10 @@ static const uint32_t symm_key_opt_or_null[] = {
 
 static const uint32_t symm_key_optional[] = {
 	PKCS11_CKA_VALUE_LEN,
+#ifdef CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE
+	/* Consider KCV attribute only when supported */
+	PKCS11_CKA_CHECK_VALUE,
+#endif
 };
 
 /* PKCS#11 specification for any asymmetric public key (+any_key_xxx) */
@@ -2173,6 +2182,7 @@ static bool attr_is_modifiable_secret_key(struct pkcs11_attribute_head *attr,
 	case PKCS11_CKA_VERIFY:
 	case PKCS11_CKA_WRAP:
 	case PKCS11_CKA_UNWRAP:
+	case PKCS11_CKA_CHECK_VALUE:
 		return true;
 	/* Can't be modified once set to CK_FALSE - 12 in Table 10 */
 	case PKCS11_CKA_EXTRACTABLE:
@@ -2478,7 +2488,11 @@ static enum pkcs11_rc set_secret_key_data(struct obj_attrs **head, void *data,
 	if (rc != PKCS11_CKR_OK && rc != PKCS11_RV_NOT_FOUND)
 		return PKCS11_CKR_GENERAL_ERROR;
 
-	return add_attribute(head, PKCS11_CKA_VALUE, data, key_length);
+	rc = add_attribute(head, PKCS11_CKA_VALUE, data, key_length);
+	if (rc)
+		return rc;
+
+	return set_check_value_attr(head);
 }
 
 static enum pkcs11_rc set_private_key_data_rsa(struct obj_attrs **head,
@@ -2826,4 +2840,198 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
 		return set_attribute(priv_head, PKCS11_CKA_ID, id1, id1_size);
 	else
 		return set_attribute(pub_head, PKCS11_CKA_ID, id2, id2_size);
+}
+
+/*
+ * The key check value is derived from the object by taking the first
+ * three bytes of the SHA-1 hash of the object's CKA_VALUE attribute.
+ */
+static enum pkcs11_rc compute_check_value_with_sha1(void *key,
+						    uint32_t key_size,
+						    void *kcv)
+{
+	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
+	TEE_Result res = TEE_ERROR_GENERIC;
+	TEE_OperationHandle op = TEE_HANDLE_NULL;
+	size_t buf_size = TEE_MAX_HASH_SIZE;
+	uint8_t *buf = NULL;
+
+	assert(key && kcv);
+
+	res = TEE_AllocateOperation(&op, TEE_ALG_SHA1, TEE_MODE_DIGEST, 0);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	buf = TEE_Malloc(buf_size, TEE_MALLOC_FILL_ZERO);
+	if (!buf) {
+		rc = PKCS11_CKR_DEVICE_MEMORY;
+		goto out;
+	}
+
+	res = TEE_DigestDoFinal(op, key, key_size, buf, &buf_size);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	TEE_MemMove(kcv, buf, PKCS11_CKA_CHECK_VALUE_SIZE);
+
+out:
+	TEE_Free(buf);
+	TEE_FreeOperation(op);
+
+	return rc;
+}
+
+/*
+ * The key check value that is calculated as follows:
+ * 1) Take a buffer of the cipher block size of binary zeros (0x00).
+ * 2) Encrypt this block in ECB mode.
+ * 3) Take the first three bytes of cipher text as the check value.
+ */
+static enum pkcs11_rc compute_check_value_with_ecb(void *key, uint32_t key_size,
+						   void *kcv)
+{
+	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
+	TEE_Result res = TEE_ERROR_GENERIC;
+	TEE_OperationHandle op = TEE_HANDLE_NULL;
+	TEE_ObjectHandle hkey = TEE_HANDLE_NULL;
+	TEE_Attribute attr = { };
+	uint8_t *buf = NULL;
+	size_t buf_size = 0;
+
+	assert(key && kcv);
+
+	res = TEE_AllocateOperation(&op, TEE_ALG_AES_ECB_NOPAD,
+				    TEE_MODE_ENCRYPT, key_size * 8);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	res = TEE_AllocateTransientObject(TEE_TYPE_AES, key_size * 8, &hkey);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	attr.attributeID = TEE_ATTR_SECRET_VALUE;
+	attr.content.ref.buffer = key;
+	attr.content.ref.length = key_size;
+
+	res = TEE_PopulateTransientObject(hkey, &attr, 1);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	res = TEE_SetOperationKey(op, hkey);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	TEE_CipherInit(op, NULL, 0);
+
+	buf_size = key_size;
+	buf = TEE_Malloc(buf_size, TEE_MALLOC_FILL_ZERO);
+	if (!buf) {
+		rc = PKCS11_CKR_DEVICE_MEMORY;
+		goto out;
+	}
+
+	res = TEE_CipherDoFinal(op, buf, buf_size, buf, &buf_size);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	TEE_MemMove(kcv, buf, PKCS11_CKA_CHECK_VALUE_SIZE);
+
+out:
+	TEE_Free(buf);
+	TEE_FreeTransientObject(hkey);
+	TEE_FreeOperation(op);
+
+	return rc;
+}
+
+enum pkcs11_rc set_check_value_attr(struct obj_attrs **head)
+{
+	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
+	uint32_t val_len = 0;
+	uint32_t kcv2_len = 0;
+	void *val = NULL;
+	uint8_t kcv[PKCS11_CKA_CHECK_VALUE_SIZE] = { };
+	void *kcv2 = NULL;
+
+	assert(head && *head);
+
+	if (!IS_ENABLED(CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE))
+		return PKCS11_CKR_OK;
+
+	switch (get_class(*head)) {
+	case PKCS11_CKO_SECRET_KEY:
+	case PKCS11_CKO_CERTIFICATE:
+		break;
+	default:
+		return PKCS11_CKR_ARGUMENTS_BAD;
+	}
+
+	/* Check whether CKA_CHECK_VALUE has been provided in the template */
+	rc = get_attribute_ptr(*head, PKCS11_CKA_CHECK_VALUE, &kcv2, &kcv2_len);
+
+	if (rc != PKCS11_CKR_OK && rc != PKCS11_RV_NOT_FOUND)
+		return PKCS11_CKR_GENERAL_ERROR;
+
+	/*
+	 * The generation of the KCV may be prevented by the application
+	 * supplying the attribute in the template as a no-value (0 length)
+	 * entry.
+	 */
+	if (rc == PKCS11_CKR_OK && !kcv2_len)
+		return PKCS11_CKR_OK;
+
+	if (rc == PKCS11_CKR_OK && kcv2_len != PKCS11_CKA_CHECK_VALUE_SIZE)
+		return PKCS11_CKR_TEMPLATE_INCONSISTENT;
+
+	/* Get key CKA_VALUE */
+	rc = get_attribute_ptr(*head, PKCS11_CKA_VALUE, &val, &val_len);
+	if (rc)
+		return rc;
+
+	if (get_class(*head) == PKCS11_CKO_SECRET_KEY) {
+		switch (get_key_type(*head)) {
+		case PKCS11_CKK_AES:
+			rc = compute_check_value_with_ecb(val, val_len, kcv);
+			break;
+		case PKCS11_CKK_GENERIC_SECRET:
+		case PKCS11_CKK_MD5_HMAC:
+		case PKCS11_CKK_SHA_1_HMAC:
+		case PKCS11_CKK_SHA256_HMAC:
+		case PKCS11_CKK_SHA384_HMAC:
+		case PKCS11_CKK_SHA512_HMAC:
+		case PKCS11_CKK_SHA224_HMAC:
+			rc = compute_check_value_with_sha1(val, val_len, kcv);
+			break;
+		default:
+			rc = PKCS11_CKR_TEMPLATE_INCONSISTENT;
+			break;
+		}
+	} else {
+		rc = compute_check_value_with_sha1(val, val_len, kcv);
+	}
+
+	if (rc)
+		return rc;
+
+	/*
+	 * If the computed KCV does not match the provided one
+	 * then return CKR_ATTRIBUTE_VALUE_INVALID
+	 */
+	if (kcv2_len) {
+		/* Provided KCV value shall match the computed one */
+		if (TEE_MemCompare(kcv2, kcv, PKCS11_CKA_CHECK_VALUE_SIZE))
+			rc = PKCS11_CKR_ATTRIBUTE_VALUE_INVALID;
+	} else {
+		rc = add_attribute(head, PKCS11_CKA_CHECK_VALUE, kcv,
+				   PKCS11_CKA_CHECK_VALUE_SIZE);
+	}
+
+	return rc;
 }
--- a/ta/pkcs11/src/processing.c
+++ b/ta/pkcs11/src/processing.c
@@ -166,10 +166,10 @@ size_t get_object_key_bit_size(struct pkcs11_object *obj)
 static enum pkcs11_rc generate_random_key_value(struct obj_attrs **head)
 {
 	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
-	void *data = NULL;
 	uint32_t data_size = 0;
 	uint32_t value_len = 0;
 	void *value = NULL;
+	void *data = NULL;
 
 	if (!*head)
 		return PKCS11_CKR_TEMPLATE_INCONSISTENT;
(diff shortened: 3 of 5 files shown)


etienne-lms:
By the way, could you add a test case in optee_test _pkcs11_1000.c_? Maybe  start with a basic test to read the attribute, destroy the attribute empty, and set the attribute to a valid and an invalid value.

Also I wonder if there should be a `CFG_PKCS11_TA_CHECK_VALUE=y|n` config switch to enable or not this feature. @vesajaaskelainen, what do you think?

maroueneboubakri:
Addressed comments above.
To add `CFG_PKCS11_TA_CHECK_VALUE=y|n` flag and a test case in optee_test `pkcs11_1000.c`

ta/pkcs11/src/pkcs11_helpers.c
```diff
@@ -506,6 +506,94 @@ bool key_type_is_asymm_key(uint32_t id)
 	}
 }
 
+/* For AES keys the KCV is 3 bytes length */
+enum pkcs11_rc aes_get_check_value_size(void *key, uint32_t key_size __unused,
```
etienne-lms, line 510: remove `__unused`

ta/pkcs11/src/pkcs11_helpers.c
```diff
@@ -506,6 +506,94 @@ bool key_type_is_asymm_key(uint32_t id)
...
+	buf = TEE_Malloc(buf_size, TEE_MALLOC_FILL_ZERO);
+	if (!buf) {
+		rc = PKCS11_CKR_DEVICE_MEMORY;
+		goto out;
+	}
+
+	res = TEE_CipherDoFinal(op, buf, buf_size, buf, &buf_size);
+	rc = tee2pkcs_error(res);
+	if (rc != PKCS11_CKR_OK) {
+		TEE_Free(buf);
```
etienne-lms, line 581: Not needed, done at line 590.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2478,7 +2484,34 @@ static enum pkcs11_rc set_secret_key_data(struct obj_attrs **head, void *data,
 	if (rc != PKCS11_CKR_OK && rc != PKCS11_RV_NOT_FOUND)
 		return PKCS11_CKR_GENERAL_ERROR;
 
-	return add_attribute(head, PKCS11_CKA_VALUE, data, key_length);
+	rc = add_attribute(head, PKCS11_CKA_VALUE, data, key_length);
+	if (rc != PKCS11_CKR_OK)
+		return rc;
+
+	/* Remove the default empty check value attribute if found */
+	rc = remove_empty_attribute(head, PKCS11_CKA_CHECK_VALUE);
```
etienne-lms, line 2492: Should not remove the attribute if present.
If present and of size 0, then `PKCS11_CKA_CHECK_VALUE` should be that one.
If present and of non-0 size, we should compute the value and succeed only if the provided value matches the expected value.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2478,7 +2484,34 @@ static enum pkcs11_rc set_secret_key_data(struct obj_attrs **head, void *data,
...
+		return rc;
+
+	kcv = TEE_Malloc(kcv_len, TEE_MALLOC_FILL_ZERO);
+	if (!kcv)
+		return PKCS11_CKR_DEVICE_MEMORY;
+
+	/* Compute and add the key check sum value*/
+	rc = aes_compute_check_value(data, key_length, kcv, &kcv_len);
+	if (rc != PKCS11_CKR_OK)
+		return  PKCS11_CKR_GENERAL_ERROR;
```
etienne-lms, line 2508: Need to balance `TEE_Malloc()` above in case of error.

Could this return `rc` instead of `PKCS11_CKR_GENERAL_ERROR`, as done in `generate_random_key_value()`?

Suggestion:
```c
	rc = aes_compute_check_value(data, key_length, kcv, &kcv_len);
	if (rc == PKCS11_CKR_OK) 
			rc = add_attribute(head, PKCS11_CKA_CHECK_VALUE, kcv, kcv_len);

	TEE_Free(kcv);

	return rc;
```

ta/pkcs11/src/processing.c
```diff
@@ -195,8 +197,35 @@ static enum pkcs11_rc generate_random_key_value(struct obj_attrs **head)
 	TEE_GenerateRandom(value, value_len);
 
 	rc = add_attribute(head, PKCS11_CKA_VALUE, value, value_len);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
+
+	/* Remove the default empty check value attribute if found */
+	rc = remove_empty_attribute(head, PKCS11_CKA_CHECK_VALUE);
```
etienne-lms, line 204: Ditto about not removing the attribute if present.

etienne-lms (commented):
The implementation lacks C_CopyObject and C_DeriveObject cases (I don't think there are other cases but I should cross-check).

If you prefer to go step by step, I think we definitely need a config switch ,default disabled, so that the attribute is not half supported as the specs says that if supported, it must be fully supported.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -400,6 +400,7 @@ static const uint32_t symm_key_boolprops[] = {
 static const uint32_t symm_key_opt_or_null[] = {
 	PKCS11_CKA_WRAP_TEMPLATE, PKCS11_CKA_UNWRAP_TEMPLATE,
 	PKCS11_CKA_DERIVE_TEMPLATE, PKCS11_CKA_VALUE,
+	PKCS11_CKA_CHECK_VALUE,
```
etienne-lms, line 403: should move to `symm_key_optional` below.
It will prevent you from removing the attribute when present.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2187,6 +2188,9 @@ static bool attr_is_modifiable_secret_key(struct pkcs11_attribute_head *attr,
 	case PKCS11_CKA_NEVER_EXTRACTABLE:
 	case PKCS11_CKA_ALWAYS_SENSITIVE:
 		return false;
+	/* CKA_CHECK_VALUE cannot be changed once it has been set. */
+	case PKCS11_CKA_CHECK_VALUE:
```
etienne-lms, line 2192: To move to the 1st case above (that returns `true`). The specs says the attribute can be modified: either replaced with a zero-sized value (meaning there is no check value support for the object) or with the valid check-value value.

ta/pkcs11/src/pkcs11_helpers.c
etienne-lms: My bad, I confused `key_size` and `kcv_size`. You were right, keep the `__unused` attribute.
Sorry.

ta/pkcs11/src/pkcs11_attributes.c
maroueneboubakri: @etienne-lms could you clarify which "that one" you are referring to you mean the one calculated from the resulting key ?
For the second case, the specs says that if provided the value provided should be taken even if it conflicts with the expected value. 
Could you please precesie ?

etienne-lms: Sure.
Extracts from the PKCS#11 Base Specification, section 4.10 Secret key objects:
* _The attribute is optional, but if supported, regardless of how the key object is created or derived, the value of the attribute is always supplied._
* _If a value is supplied in the application template (allowed but never necessary) then, if supported, it MUST match what the library calculates it to be or the library returns a CKR_ATTRIBUTE_VALUE_INVALID._
* _The generation of the KCV may be prevented by the application supplying the attribute in the template as a no-value (0 length) entry._
* _The application can query the value at any time like any other attribute using C_GetAttributeValue._
* _C_SetAttributeValue may be used to destroy the attribute, by supplying no-value._

ta/pkcs11/src/processing.c
etienne-lms: The template passed to C_GenerateKey can set CKA_CHECK_VALUE attribute. In which case we should not omit it but ensure it complies with a valid value, that is either a zero-sized value or the value the pkcs11 TA also computes. Other values for this attribute should error with return code CKR_ATTRIBUTE_VALUE_INVALID.

ta/pkcs11/src/pkcs11_attributes.c
etienne-lms: Sorry i was wrong in one point: it seems at object creation, the client template cannot prevent generation of CKA_CHECK_VALUE attribute : _".. it MUST match what the library calculates."_. To destroy it, the client must later use C_SetAttributeValue and set a _"no-value"_ (that is a zero-sized value for the attribute.

etienne-lms:
> > The implementation lacks C_CopyObject and C_DeriveObject cases (I don't think there are other cases but I should cross-check).
> 
> @etienne-lms My understatding is that set_key_data() serves fro key unwrapping and derivation ? while copy object nothing to do since all the attribues will be copied ?

You right for C_DeriveKey, it ends in calls `entry_processing_key()` that calls `set_key_data()`. C_UnwrapKey also ends in `entry_processing_key()`.
C_CreateObject is not yet covered in your P-R. The template provided could set CKA_CHECK_VALUE attribute.
C_CopyObject can provide a template that modifies the copied object attributes, so need to handle CKA_CHECK_VALUE.
C_GenerateKey also provides a template that could define CKA_CHECK_VALUE: unlikely to have to right value, I admit :). in such case, the operation should fail with CKR_ATTRIBUTE_VALUE_INVALID.

maroueneboubakri:
Update: Moving the key check value handeling to a single function `check_attr_checksum()`. This function is called to set the attribute on `C_GenerateKey()`,  `C_CreateObject()`,  `C_UnwrapKey()`, `C_DeriveKey()`,  `C_CopyObject()` and `C_SetAttributeValue()`

Commits will be squashed later, for now just to make sure the control flow is okay.

ta/pkcs11/src/pkcs11_helpers.h
```diff
@@ -55,6 +61,13 @@ bool key_type_is_symm_key(uint32_t key_type_id);
 /* Return true if the key type @key_type_id relates to an asymmetric key */
 bool key_type_is_asymm_key(uint32_t key_type_id);
 
+/* Calculate AES key check value size */
+enum pkcs11_rc aes_get_check_value_size(void *key, uint32_t key_size,
+					uint32_t *kcv_size);
```
etienne-lms, line 66: I think it's easier to remove this function and move `PKCS11_CKA_CHECK_VALUE_SIZE` definition to _pkcs11_attributes.h_, the size applies to all objects (certificate and symm keys).

ta/pkcs11/src/pkcs11_helpers.h
```diff
@@ -55,6 +61,13 @@ bool key_type_is_symm_key(uint32_t key_type_id);
...
+/* Calculate AES key check value data */
+enum pkcs11_rc aes_compute_check_value(void *key, uint32_t key_size,
+				       void *kcv, uint32_t *kcv_size);
```
etienne-lms, line 69: This function is only called from `check_attr_checksum()`, please move it in the related source files as a static function.

The function applies to CKK_AES, CKK_DES and CKK_DES3. I suggest to rename the static function `compute_check_value_with_ecb()` and pass the key_type ID. It would currently support only PKCS11_CKK_AES, but ill be easy to later extend to CKK_DES and CKK_DES3 that are computed the same way.

CKA_CHECK_VALUE attribute is always 3 bytes so need to pass kcv_size as argument. This function could get instead a `struct obj_attrs` as argument, preallocated by caller (`sizeof(struct obj_attrs) + PKCS11_CKA_CHECK_VALUE_SIZE)` and fill it with the computed value.

ta/pkcs11/src/pkcs11_attributes.c
etienne-lms, line 403: > should move to symm_key_optional below.
> It will prevent you from removing the attribute when present.

You forgot to address this comment.

ta/pkcs11/src/pkcs11_attributes.c
etienne-lms, line 2192: > To move to the 1st case above (that returns true). The specs says the attribute can be modified: either replaced with a zero-sized value (meaning there is no check value support for the object) or with the valid check-value value.

You forgot to address this comment.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2831,3 +2831,84 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
 	else
 		return set_attribute(pub_head, PKCS11_CKA_ID, id2, id2_size);
 }
+
+enum pkcs11_rc check_attr_checksum(struct obj_attrs *head)
```
etienne-lms, line 2839: Could you rename this function `set_check_value_attr()` since it not only check to value but also sets it?

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2831,3 +2831,84 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+{
+	enum pkcs11_rc rc = PKCS11_CKR_GENERAL_ERROR;
+
```
etienne-lms, line 2842: remove the empty line

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2831,3 +2831,84 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+	uint32_t value_len = 0;
+	uint32_t kcv_len = 0;
+	void *value = NULL;
+	void *kcv = NULL;
+
+	if (!head)
```
etienne-lms, line 2848: `assert(head)` is enough.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2831,3 +2831,84 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+		return PKCS11_CKR_ARGUMENTS_BAD;
+
+	/* CKA_CHECK_VALUE is also an attribute of CKO_CERTIFICATE */
+	if (get_class(head) != PKCS11_CKO_SECRET_KEY)
+		return PKCS11_RV_NOT_IMPLEMENTED;
+
+	if (get_key_type(head) != PKCS11_CKK_AES)
```
etienne-lms, line 2855: We need to support CKA_CHECK_VALUE for all supported CKO_SECRET_KEY objects, as stated in the spec:
"_The attribute is optional, but if supported, regardless of how the key object is created or derived, the value of the attribute is always supplied._"

`PKCS11_CKK_AES` keys would use `compute_check_value_from_ecb()`.
`PKCS11_CKK_GENERIC_SECRET` would use `compute_check_value_from_sha1()` (compute SHA1 and extract the 3 first bytes).
For `PKCS11_CKK_SHA*_HMAC*`, I understand they fall in the CKK_GENERIC_SECRET case => 3 first byte of the SHA1 hash of the key data.

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2831,3 +2831,84 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+	}
+
+	/*
+	 * The generation of the KCV may be prevented by the application
+	 * supplying the attribute in the template as a no-value (0 length)
+	 * entry.
+	 */
+	if (rc == PKCS11_CKR_OK) {
+		if (value_len == 0)
+			goto out;
```
etienne-lms, line 2895: this test could be done before computation of the value: not needed if client does not want the attribute to be set.

ta/pkcs11/src/processing.c
```diff
@@ -195,7 +195,12 @@ static enum pkcs11_rc generate_random_key_value(struct obj_attrs **head)
 	TEE_GenerateRandom(value, value_len);
 
 	rc = add_attribute(head, PKCS11_CKA_VALUE, value, value_len);
+	if (rc != PKCS11_CKR_OK)
+		goto out;
```
etienne-lms, line 199: simpler if the form of:
```c
	if (rc == PKCS11_CKR_OK)
		rc = check_attr_checksum(*head);

	TEE_Free(value);

	return rc;

File: ta/pkcs11/src/object.c [hunk 16]
```diff
@@ -369,6 +369,10 @@ enum pkcs11_rc entry_create_object(struct pkcs11_client *client,
 	if (rc)
 		goto out;
 
+	rc = check_attr_checksum(head);
```
etienne-lms, line 372: I would prefer this to be called before `check_created_attrs_against_processing()`, for consistency:
first set all attributes, then check they comply, then create the object.

ta/pkcs11/src/object.c
```diff
@@ -1060,6 +1060,11 @@ enum pkcs11_rc entry_set_attribute_value(struct pkcs11_client *client,
 	if (rc)
 		goto out;
 
+	/* Set key check value attribute */
+	rc = check_attr_checksum(head);
```
etienne-lms, line 1064: I would prefer this to be called before `check_attrs_against_modification()`, for consistency:
first set all attributes, then check they comply, then create the object.

etienne-lms (commented):
Thanks for the update. The implementation looks quite nice. I have few comments (see below) but I think the main needed pieces are in place.

Since the spec says the attribute is optional, I really think a `CFG_PCSK11_TA_CHECK_VALUE_ATTRIBUTE=y|n` would be nice. Note this can be discussed in a separated P-R.

maroueneboubakri:
Addressed most of the comments. Stills need to implement `compute_check_value_with_sha1()` add flag `CFG_PCSK11_TA_CHECK_VALUE_ATTRIBUTE`. Investigating issue with `struct obj_attrs` allocation prior to adding the KCV attribute. @etienne-lms

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2827,3 +2829,170 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+	void *val = NULL;
+	void *kcv = NULL;
+	void *kcv2 = NULL;
+	struct obj_attrs *nhead = NULL;
+
+	assert(head);
+
+	if (get_class(head) != PKCS11_CKO_SECRET_KEY &&
+	    get_class(head) == PKCS11_CKO_CERTIFICATE)
+		return PKCS11_CKR_ARGUMENTS_BAD;
```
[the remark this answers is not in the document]
maroueneboubakri: PKCS11_RV_NOT_IMPLEMENTED ?

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2827,3 +2829,170 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+		rc = PKCS11_CKR_ATTRIBUTE_VALUE_INVALID;
+		goto out;
+	}
+
+	nhead = TEE_Realloc(head, sizeof(struct obj_attrs) +
+			    PKCS11_CKA_CHECK_VALUE_SIZE);
+	if (!nhead)
+		return PKCS11_CKR_DEVICE_MEMORY;
+	head = nhead;
+	rc = add_attribute(&head, PKCS11_CKA_CHECK_VALUE,
```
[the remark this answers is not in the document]
maroueneboubakri: Indeed !

ta/pkcs11/sub.mk
```diff
@@ -10,6 +10,9 @@ CFG_PKCS11_TA_HEAP_SIZE ?= (32 * 1024)
 # Defines the number of PKCS11 token implemented by the PKCS11 TA
 CFG_PKCS11_TA_TOKEN_COUNT ?= 3
 
+# Disable optional object checksum value computation
```
etienne-lms, line 13: `# When enabled, embed support for object checksum value computation`

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2478,7 +2481,14 @@ static enum pkcs11_rc set_secret_key_data(struct obj_attrs **head, void *data,
 	if (rc != PKCS11_CKR_OK && rc != PKCS11_RV_NOT_FOUND)
 		return PKCS11_CKR_GENERAL_ERROR;
 
-	return add_attribute(head, PKCS11_CKA_VALUE, data, key_length);
+	rc = add_attribute(head, PKCS11_CKA_VALUE, data, key_length);
+	if (rc)
+		return rc;
+
+	/* Set key check value attribute */
+	rc = set_check_value_attr(*head);
```
etienne-lms, line 2489: nitpicking: `return set_check_value_attr(*head);`

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2827,3 +2837,172 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+}
+
+/*
+ * The key check value that is calculated as follows:
+ * 1) Take a buffer of the cipher block size of binary zeros (0x00).
+ * 2) Encrypt this block in ECB mode.
+ * 3) Take the first three bytes of cipher text as the check value.
+ */
+static enum pkcs11_rc
+compute_check_value_with_ecb(void *key, uint32_t key_size, void *kcv)
```
etienne-lms, line 2855:
```c
static enum pkcs11_rc compute_check_value_with_ecb(void *key, uint32_t key_size,
						   void *kcv)
```

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2827,3 +2837,172 @@ enum pkcs11_rc add_missing_attribute_id(struct obj_attrs **pub_head,
...
+
+	if (!IS_ENABLED(CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE))
+		return PKCS11_CKR_OK;
+
+	if (get_class(head) != PKCS11_CKO_SECRET_KEY &&
+	    get_class(head) != PKCS11_CKO_CERTIFICATE)
+		return PKCS11_CKR_ARGUMENTS_BAD;
+
+	if (get_class(head) == PKCS11_CKO_CERTIFICATE)
+		return PKCS11_RV_NOT_IMPLEMENTED;
```
etienne-lms, line 2935: These `if` instruction tests the same thing. Prefer:
```c
	switch (get_class(head)) {
	case PKCS11_CKO_SECRET_KEY:
		break;
	case PKCS11_CKO_CERTIFICATE:
		return PKCS11_RV_NOT_IMPLEMENTED;
	default:
		return PKCS11_CKR_ARGUMENTS_BAD;
	}
```

Note that you already treat PKCS11_CKO_CERTIFICATE at line 2984, so maybe prefer here:
```c
	switch (get_class(head)) {
	case PKCS11_CKO_SECRET_KEY:
	case PKCS11_CKO_CERTIFICATE:
		break;
	default:
		return PKCS11_RV_NOT_IMPLEMENTED;
	}

**Comment by maroueneboubakri** [comment]
Comments addressed

**Review by etienne-lms — COMMENTED** [review]
I think all commits should be squashed in a single one, inlcuding the change in ta/pkcs11/sub.mk that describes `CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE`.

I'll make another review round on the changes but they look good to me.

**Review by etienne-lms — COMMENTED** [review]
Thanks for the update.
The commit message is far too verbose. You don't need to describe everything that is done in the change, readers can look at the modification. The commit message should rather briefly describe what is changed and why.
Suggestion:
```
Add PKCS11_CKA_CHECK_VALUE as an optional attribute of symmetric
key and certificate objects . As per the PKCS#11 specification, key
check value attribute is optional therefore add pkcs11 TA configuration
switch CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE to embed or not the support.

When supported, as per the spec, the attribute can be either the
legitimate value recomputed by the PKCS#11 token or a zero-sized value
called a no-value for when client does not want the attribute to set
in an object.

This change adds the support for the pcks11 TA commands related to
Cryptoki API functions C_GenerateKey(), C_CreateObject(), C_CopyObject(),
C_SetAttributeValue(), C_UnwrapKey() and C_DeriveKey(). TA command
related to C_FindOjects() support the attribute without any change.

maroueneboubakri:
Comments addressed

etienne-lms (commented):
Could you provide few tests in a new test case in xtest _pkcs11_1000.c_ to test this? By using `CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE=y` in Qemu based CI test config, we would run regression test covering the feature.

etienne-lms:
@vesajaaskelainen, maybe are you interested in looking at this change.

ta/pkcs11/src/object.c
```diff
@@ -1051,6 +1056,11 @@ enum pkcs11_rc entry_set_attribute_value(struct pkcs11_client *client,
 	if (rc)
 		goto out;
 
+	/* Set key check value attribute */
+	rc = set_check_value_attr(&head);
```
maroueneboubakri, line 1060: @etienne-lms I have a concern here, as per your recommandation  `set_check_value_attr()` has been moved before  `check_attrs_against_modification()` to santize `CKA_CHECK_VALUE ` against modification, however, at this stage the object is object attributes are required to compute the KCV such as the class and key type. Here, the template and object are both needed to compute/test the KCV. My question, should this call moved elsewhere where all attributes are consolidated/merged together ?

maroueneboubakri:
> Could you provide few tests in a new test case in xtest pkcs11_1000.c to test this? 

PR which adds test cases  https://github.com/OP-TEE/optee_test/pull/719

maroueneboubakri:
Change: For `C_SetAttributeValue()` -> `entry_set_attribute_value()` function `set_check_value_attr()` operates on object attributes not on the provided template. Object attributes such as the class, key type and key value are required to compute the KCV.

etienne-lms (commented):
One last comment otherwise looks good to me. Let's see it tested.
Last minute, I think the certificate part lacks the attribute definition.

maroueneboubakri:
Here is the test result: 

```
# xtest -t pkcs11 1029
Test ID: 1029
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# pkcs11
#
######################################################
 
* pkcs11_1029 PKCS11: Test support for object checksum value computation
o pkcs11_1029.1 Compute KCV on C_GenerateKey()
  pkcs11_1029.1 OK
o pkcs11_1029.2 Set KCV on C_CreateObject()
  pkcs11_1029.2 OK
o pkcs11_1029.3 Compute KCV on C_CopyObject()
  pkcs11_1029.3 OK
o pkcs11_1029.4 Compute KCV on C_DeriveKey()
  pkcs11_1029.4 OK
o pkcs11_1029.5 Compute KCV on C_UnwrapKey()
  pkcs11_1029.5 OK
o pkcs11_1029.6 Destroy KCV using C_SetAttributeValue()
  pkcs11_1029.6 OK
  pkcs11_1029 OK
Result of testsuite pkcs11 filtered by "1029":
pkcs11_1029 OK
25 subtests of which 0 failed
1 test case of which 0 failed
29 test cases were skipped
TEE test application done!
```

maroueneboubakri:

etienne-lms:
> Change: For C_SetAttributeValue() -> entry_set_attribute_value() function set_check_value_attr() operates on object attributes not on the provided template. Object attributes such as the class, key type and key value are required to compute the KCV.

I've fetched your test (https://github.com/OP-TEE/optee_test/pull/719) and added few others tests (see https://github.com/etienne-lms/optee_test/commit/f6f4cb5d2909fc4bed55a754aa365c6427225856 temp link).

I see an issue with C_SetAttribute.
When client provides an invalid KCV value, the object is changed before the error is reported. The object should not change. We must review integration in C_SetAttributeValue.

maroueneboubakri:
> > Change: For C_SetAttributeValue() -> entry_set_attribute_value() function set_check_value_attr() operates on object attributes not on the provided template. Object attributes such as the class, key type and key value are required to compute the KCV.
> 
> I've fetched your test ([OP-TEE/optee_test#719](https://github.com/OP-TEE/optee_test/pull/719)) and added few others tests (see [etienne-lms/optee_test@f6f4cb5](https://github.com/etienne-lms/optee_test/commit/f6f4cb5d2909fc4bed55a754aa365c6427225856) temp link).
> 
> I see an issue with C_SetAttribute. When client provides an invalid KCV value, the object is changed before the error is reported. The object should not change. We must review integration in C_SetAttributeValue.

@etienne-lms this has been asked in my comment above, for C_SetAttributeValue() both the template attributes and object attributes are required to make the computation and the check. My question was somehow if there is any way where these are merged together to proceed with check value computation and it seems that there is ony a way after the checks are done.

etienne-lms:
C_SetAttributeValue is specific among the other cryptoki service is that it modifies an existing object. I'll propose a way, where this sequence preserves the object original data (attributes) if the service fails at any point. The idea is to copy the object attributes in a temporary reference.

etienne-lms:
See https://github.com/OP-TEE/optee_os/pull/6550.

PR #6487, opened 2023-11-23
drivers: regulator: fix variable sized voltages fallback

etienne-lms:
drivers: regulator: fix variable sized voltages fallback

Fix build issue reported by Clang on variable size field `desc` not being located at the end of `struct voltages_fallback`. The error was reported with a trace message like below:
````
core/include/drivers/regulator.h:118:4: warning: field 'voltages_fallback' with variable sized type 'struct voltages_fallback' not at the end of a struct or class is a GNU extension [-Wgnu-variable-sized-type-not-at-end]
        } voltages_fallback;
          ^
core/drivers/regulator/regulator_fixed.c:27:19: warning: field 'regulator' with variable sized type 'struct regulator' not at the end of a struct or class is a GNU extension [-Wgnu-variable-sized-type-not-at-end]
        struct regulator regulator;
                         ^
2 warnings generated.
```

API function `regulator_supported_voltages()` asserts that `struct regulator_voltages` and `struct voltages_fallback` do match.

By the way, remove empty line in `struct regulator_voltages` inline  description comment.

Fixes: 43c155ba111d ("drivers: regulator: list supported levels")

A second change enabled CI build of STM32MP13 platform with Clang.

PR #6498, opened 2023-11-27
drivers: regualtor: print tree

etienne-lms:
Implementation of `regulator_print_tree()` as per @jenswi-linaro https://github.com/OP-TEE/optee_os/pull/6481#discussion_r1404590521 proposal.

Compared to #6481, this series only implements `regulator_print_tree()` and ensures it's built by CI. But not yet by CLang CI since it requires a fix as proposed by https://github.com/OP-TEE/optee_os/pull/6487.

### DIFF
--- a/core/drivers/regulator/regulator.c
+++ b/core/drivers/regulator/regulator.c
@@ -288,26 +288,6 @@ TEE_Result regulator_register(struct regulator *regulator)
 }
 
 /*
- * Log regulators state
- */
-void regulator_print_state(const char *message __maybe_unused)
-{
-	struct regulator *regulator = NULL;
-
-	DMSG("Regulator state: %s", message);
-	DMSG("name     use\ten\tuV\tmin\tmax\tflags\tsupply");
-
-	SLIST_FOREACH(regulator, &regulator_device_list, link)
-		DMSG("%8s %u\t%d\t%d\t%d\t%d\t%#x\t%s",
-		     regulator->name, regulator->refcount,
-		     regulator_is_enabled(regulator),
-		     regulator_get_voltage(regulator),
-		     regulator->min_uv, regulator->max_uv, regulator->flags,
-		     regulator->supply ? regulator_name(regulator->supply) :
-		     "<none>");
-}
-
-/*
  * Clean-up regulators that are not used.
  */
 static TEE_Result regulator_core_cleanup(void)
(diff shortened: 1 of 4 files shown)


core/drivers/regulator/regulator.c
```diff
@@ -323,9 +303,201 @@ static TEE_Result regulator_core_cleanup(void)
...
+
+	if (ret < 0 || ret >= max_len)
+		return NULL;
+
+	return cur + ret;
+}
+
+/* Regulator is the last supplied one by its supply in the registered list */
+static bool regulator_is_supply_last_supplied(struct regulator *node_regulator)
+{
```
jenswi-linaro, line 333: This could be implemented as `return !find_next_regulator(node_regulator->supply, node_regulator);`

core/drivers/regulator/regulator.c
```diff
@@ -328,6 +328,43 @@ static __printf(3, 4) char *add_msg(char *cur, char *end, const char *fmt, ...)
...
+	struct regulator *indent_plus_one_regulator = NULL;
+	struct regulator *indent_regulator = NULL;
+	struct regulator *regulator = NULL;
+	int indent = 0;
+
+	/* Find the supplied and supply of @cur_indent indentation level */
+	indent_regulator = node_regulator;
+	for (indent = 0; indent < node_indent - cur_indent; indent++) {
+		indent_plus_one_regulator = indent_regulator;
+		indent_regulator = indent_regulator->supply;
```
jenswi-linaro, line 357: It looks like you don't need this below, so the loop could be simplified.

core/drivers/regulator/regulator.c
```diff
@@ -323,9 +303,188 @@ static TEE_Result regulator_core_cleanup(void)
...
+static bool regulator_is_supply_last_supplied(struct regulator *regulator)
+{
+	return !find_next_regulator(regulator->supply, regulator);
+}
+
+/* Supply last node may already be printed for indentation level @cur_indent */
+static bool indent_with_empty_string(struct regulator *node_regulator,
+				     int node_indent, int cur_indent)
+{
+	struct regulator *indent_plus_one_regulator = NULL;
```
jenswi-linaro, line 357: There aren't too many local variables here, how about using a shorter name like `r` for instance?

core/drivers/regulator/regulator.c
```diff
@@ -323,9 +303,188 @@ static TEE_Result regulator_core_cleanup(void)
...
+	int indent = 0;
+
+	/* Find the supplied and supply of @cur_indent indentation level */
+	indent_plus_one_regulator = node_regulator;
+	for (indent = 0; indent < node_indent - cur_indent - 1; indent++)
```
jenswi-linaro, line 362: `indent` is not an indentation level, this might be easier to read using the name `n` or such.

etienne-lms:
Comments addressed

core/drivers/regulator/regulator.c
```diff
@@ -323,9 +303,187 @@ static TEE_Result regulator_core_cleanup(void)
 		}
 	}
 
-	regulator_print_state(__func__);
+	if (TRACE_LEVEL >= TRACE_DEBUG)
```
jforissier, line 306: >= TRACE_INFO?

jforissier (commented):
One comment below. Also you have two blank lines before the tags on the commit descriptions, one is enough.
With that:

core/drivers/regulator/regulator.c
etienne-lms: The previous implementation printed the regulator tree with TRACE_DEBUG level so I've replicated the behavior. Makes sense?

etienne-lms: I think the regulator tree should not be part of the very few info level traces printed at OP-TEE boot-time when TRACE_LEVEL is TRACE_INFO.
I could address your concern replacing use of `IMSG()` with `MSG()` in `regulator_print_tree()` and related functions.

jforissier: Then why not use DMSG() if it's a debug message?

etienne-lms: `DMSG` would require TRACE_DEBUG level to be enabled, which is too often too verbose. Using TRACE_INFO level makes OP-TEE far less verbose, still aloowing one to print some info (as regu/clock trees) when needed.
That's not a strong opinion but I think it's more handy.

jforissier: > `DMSG` would require TRACE_DEBUG level to be enabled, which is too often too verbose

But `if (TRACE_LEVEL >= TRACE_DEBUG)` also requires DEBUG to be enabled! (or FLOW which is even more verbose).
The code as it is proposed does not print anything when TRACE_LEVEL is TRACE_ERROR or TRACE_INFO, and it prints something (with severity INFO) when TRACE_LEVEL is TRACE_DEBUG or TRACE_FLOW in which case all DEBUG messages are always there too. So I don't understand.

This is debugging information. IMO the right thing is just to use `DMSG()` (or `DMSG_RAW()` if you'd rather not have the file/line number which makes sense) and no `if`. It's not like we need to optimize anything here.

etienne-lms: Ok, let's make it simple. I'll switch to `DMSG()`.

etienne-lms: ... and update the commit message, _mk/config.mk_ and _regulator.h_ inline description comments accordingly.

jforissier (commented):

Thanks!

etienne-lms:
One more fixup commit to save 1kB when TRACE_LEVEL < TRACE_DEBUG.

jforissier:
LGTM

etienne-lms:
Thanks, i'll squash the fixup commits, update the related commit message and applied review tags.

Issue #6495, opened 2023-11-25
Modify obj usage for persistence object


kyoWNC:
optee-4.0.0
Case: Create a pure data object that can not share to CA
1. Use the TEE_CreatePersistentObject API to create a pure data object (ID: obj1)  (default obj usage is 0xffffffff)
2. Modify the object usage to 0xfffffffe using TEE_RestrictObjectUsage1
3. Call TEE_GetObjectInfo1 to check object usage, obtaining the result 0xfffffffe

But when using the TEE_GetObjectInfo1 API to get object usage from another CA, it will return 0xffffffff.
Is this the expected result?

Thanks

jenswi-linaro:
I see how this might fail now. Yeah, this is a bug.

kyoWNC:
Hi jenswi-linaro,
I added the following code into regression_6000.c for the xtest.

```
static void xtest_tee_test_6021_single(ADBG_Case_t *c, uint32_t storage_id) 
{
	TEEC_Result res = TEEC_ERROR_GENERIC;
	TEEC_Session sess = { };
	uint32_t orig = 0;
	uint32_t obj = 0;
	TEE_ObjectInfo obj_info = { };
	
	res = xtest_teec_open_session(&sess, &storage_ta_uuid, NULL, &orig);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
		return;

	res = fs_create(&sess, file_01, sizeof(file_01),
				TEE_DATA_FLAG_ACCESS_READ |
				TEE_DATA_FLAG_ACCESS_WRITE |
				TEE_DATA_FLAG_ACCESS_WRITE_META,
				0,
				data_00, sizeof(data_00),
				&obj,
				storage_id);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
		goto exit;

	res = fs_get_obj_info(&sess, obj, &obj_info);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
		goto exit;
	ADBG_EXPECT_COMPARE_UNSIGNED(c, obj_info.objectUsage, ==, 0xffffffff);

	res = fs_restrict_usage(&sess, obj, 0xfffffffe);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
		goto exit;

	res = fs_get_obj_info(&sess, obj, &obj_info);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
		goto exit;
[...]

PR #6505, opened 2023-11-29
core: fix objectUsage handling for persistent objects

jenswi-linaro:
Before this patch was the object usage field stored in the non-persistent part of an object handle, regardless of whether the storage object was persistent. This prevents updates to this field from being restored the next time the persistent object is opened. Updates to the field are also not replicated to eventual other open handles for the object. Fix this by storing the "usage" bits in a new obj_info_usage field in struct tee_pobj for persistent objects. Updates to the field are also written into secure storage to preserve the content the next time the object is opened.

Fixes: b01047730e77 ("Open-source the TEE Core")
Closes: https://github.com/OP-TEE/optee_os/issues/6495


jenswi-linaro:
Fixing checkpatch complaint.

jforissier:
`Fixes: b01047730e77 ("Open-source the TEE Core")`

:open_mouth: quite a long-standing bug!

jenswi-linaro:
> 😮 quite a long-standing bug!

Yeah :-)

Tag applied.

etienne-lms (commented):

jenswi-linaro:
Tag applied.

PR #6509, opened 2023-11-29
core: dt_driver: do not defer probe on skip phandler arguments

etienne-lms:
Prevent `dt_driver_device_from_node_idx_prop()` to request driver probe deferral when needing to skip phandle arguments cells related to providers not yet registered. When skipping these phandle argument cells we don't really need the skipped provider is registered, we can read the `#xxx-cells` property in the provider DT node straight.

For example, consider a driver which DT node defines:
```
  clocks = <&foo_clock 1 2 3>, <&bar_clock 2>;
  clock-names = "foo", "bar";
```
If driver calls `clk_get_by_name(fdt, node, "bar")`, it does not need to wait `&foo_clock` related driver to be already probed, it does not even need the driver to be ever probed.

Reported-by: Gatien Chevallier <gatien.chevallier@foss.st.com>
Link: https://github.com/OP-TEE/optee_os/pull/6491#discussion_r1404181813
Fixes: a22e85b25772 ("core: dt_driver: factorize clk_dt_get_from_provider")

### DIFF
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -383,20 +383,44 @@ TEE_Result dt_driver_device_from_node_idx_prop(const char *prop_name,
 				return TEE_ERROR_GENERIC;
 
 			prv = dt_driver_get_provider_by_node(nodeoffset, type);
-			if (!prv)
-				return TEE_ERROR_DEFER_DRIVER_INIT;
 		} else {
 			prv = dt_driver_get_provider_by_phandle(phandle, type);
-			if (!prv)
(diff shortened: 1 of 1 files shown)


etienne-lms:
Sorry I forgot to test that with debug trace level.

GseoC:
2 minors comments:

GseoC:
> 2 minors comments:
> 
> `Acked-by: Gatien Chevallier <gatien.chevallier@foss.st.com>`

Actually, forget the comments, the format is coherent with what's above

etienne-lms:
Review tags applied and rebased to fix IBART test failure.

PR #6510, opened 2023-11-29
Update transfer list to align to the spec

raymo200915:
1. Add `flags` into the transfer list header and related bit `TL_FLAGS_HAS_CHECKSUM`.
2. Bypass verifying/updating checksum when `TL_FLAGS_HAS_CHECKSUM` is not set.
3. Release the unused bytes from the transfer list when calling function `transfer_list_set_data_size()` instead of just marking it with a void tag.


core/kernel/transfer_list.c
```diff
@@ -389,13 +389,16 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
...
+		mov_dis = ROUNDDOWN(old_ev - new_ev,
+				    TL_ALIGNMENT_FROM_ORDER(tl->alignment));
+		r_new_ev = old_ev - mov_dis;
+		tl->size -= mov_dis;
 	}
+	gap = r_new_ev - new_ev;
+	memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
 
 	if (gap >= sizeof(*dummy_te)) {
 		/* Create a dummy transfer entry to fill up the gap */
```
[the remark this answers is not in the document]
raymo200915: I think we still need it, if the alignment is big enough that the gap will have a chance to bigger than a TE header.

core/include/kernel/transfer_list.h
```diff
@@ -6,7 +6,7 @@
 #ifndef __KERNEL_TRANSFER_LIST_H
 #define __KERNEL_TRANSFER_LIST_H
 
-#define TRANSFER_LIST_SIGNATURE		U(0x006ed0ff)
+#define TRANSFER_LIST_SIGNATURE		U(0x4a0fb10b)
```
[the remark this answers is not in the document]
raymo200915: I think the backward compatibility is not yet considered at this moment since current v0.9 is still not a stable version (The rev doesn't increase with changes).

Below is the related PR for the spec:
https://github.com/FirmwareHandoff/firmware_handoff/pull/24

core/include/kernel/transfer_list.h
```diff
@@ -6,7 +6,7 @@
...
 #define TRANSFER_LIST_VERSION		U(0x0001)
```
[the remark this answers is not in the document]
raymo200915: Yes. The version is still 1 in the spec.

core/include/kernel/transfer_list.h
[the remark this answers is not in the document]
jenswi-linaro: Ouch! Good thing that this is by default off in all platforms since we usually don't break ABIs. I guess that means that it should remain off until we're using a stable version. It might be worth mentioning that in `mk/config.mk`.

core/kernel/transfer_list.c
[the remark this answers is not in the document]
jenswi-linaro: What about this passage in the commit message?
> Releasing the unused bytes from transfer list when downsizing the data
of a transfer entry is required, since currently allocating new transfer
entries from spaces marked with 'TL_TAG_EMPTY' is not supported.

core/include/kernel/transfer_list.h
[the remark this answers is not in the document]
raymo200915: Comment is added.

raymo200915:
Added new commit for comments on CFG_TRANSFER_LIST.
Amended commit message of [core: remove unused bytes from transfer list](https://github.com/OP-TEE/optee_os/pull/6510/commits/78067dd5fb0cc2ee5ba31a1ed1eba161118ef1c3).

raymo200915:
> For "core: update transfer list header and signature" please add a note that this commit causes Transfer List ABI breakage.

Done.

mk/config.mk
```diff
@@ -499,6 +499,8 @@ endif
 
 # This option enables OP-TEE to support boot arguments handover via Transfer
 # List defined in Firmware Handoff specification.
+# Please note that it is an experimental feature until Firmware Handoff
+# specification has a stable release.
```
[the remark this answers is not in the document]
raymo200915: Done.

core/kernel/transfer_list.c
```diff
@@ -389,13 +394,16 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
...
-		gap = old_ev - new_ev;
+		mov_dis = ROUNDDOWN(old_ev - new_ev,
+				    TL_ALIGNMENT_FROM_ORDER(tl->alignment));
+		r_new_ev = old_ev - mov_dis;
+		tl->size -= mov_dis;
 	}
+	gap = r_new_ev - new_ev;
+	memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
 
 	if (gap >= sizeof(*dummy_te)) {
```
jenswi-linaro, line 408: Please add a comment explaining why we have these gaps.

core/kernel/transfer_list.c
```diff
@@ -389,13 +394,16 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
...
-		gap = ru_new_ev - new_ev;
 	} else {
-		gap = old_ev - new_ev;
+		mov_dis = ROUNDDOWN(old_ev - new_ev,
+				    TL_ALIGNMENT_FROM_ORDER(tl->alignment));
+		r_new_ev = old_ev - mov_dis;
+		tl->size -= mov_dis;
 	}
+	gap = r_new_ev - new_ev;
+	memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
```
jenswi-linaro, line 406: Please add a comment explaining that we're moving eventual entries following this entry to either make room or to close the gap.

core/kernel/transfer_list.c
raymo200915: Added few lines of comments in this function.

core/kernel/transfer_list.c
raymo200915: Done.

raymo200915:

core/kernel/transfer_list.c
```diff
@@ -389,14 +394,26 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
...
 		}
-		ru_new_ev = old_ev + mov_dis;
-		memmove((void *)ru_new_ev, (void *)old_ev, tl_old_ev - old_ev);
+		r_new_ev = old_ev + mov_dis;
 		tl->size += mov_dis;
-		gap = ru_new_ev - new_ev;
 	} else {
-		gap = old_ev - new_ev;
+		/*
+		 * Move distance should be round down to meet the requirement of
```
jenswi-linaro, line 401: How about:
`/* Move distance should be rounded down to match the data alignment */`

core/kernel/transfer_list.c
```diff
@@ -389,14 +394,26 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
...
+				    TL_ALIGNMENT_FROM_ORDER(tl->alignment));
+		r_new_ev = old_ev - mov_dis;
+		tl->size -= mov_dis;
 	}
+	/* Move all following entries to fit in the expanded or shrunk space */
+	memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
 
+	/*
+	 * Fill the gap due to round up/down with a valid entry if the size of
+	 * the gap is more than an entry header.
```
jenswi-linaro, line 414: What if the next entry is a `TL_TAG_EMPTY`? Would it make sense to merge the two entries?

raymo200915: I guess, idealy we need another API for fragment management.
But I think we can skip it in this PR, since currently when an entry is marked with void' it will be never re-used.
`transfer_list_add()` will just append new entries in the tail (The spec doesn't mandatorily require reuse of 'void' entries, it is still a question whether we need to support the reuse).

core/kernel/transfer_list.c
raymo200915: Fixed. I updated the comment for the condition of `rounded up` as well.

core/kernel/transfer_list.c
jenswi-linaro: What the spec says about this isn't so important, but I agree that we can save it for another day.

jenswi-linaro:

raymo200915:
Rebased to top of master, added review tags.

PR #6511, opened 2023-11-30
crypto: fix qm cache start and done define

yuzexiyzx:
Address offset of qm_cache_wb_start and qm_cache_wb_done is wrong. 
Fixes: https://github.com/OP-TEE/optee_os/pull/6368


etienne-lms (commented):
I suggest you remove the sentence "So, we try to fix it." from the commit message.
Please add a `Fixes:` tag referring to the commit that is fixed.

PR #6512, opened 2023-11-30
core: arm: fix NS entrypoint for secondary cores

ChiaweiW:
The NS entrypoint was originally backup from LR to R5. 
The commit f332e77c4b7c revises the register preservation to R8.
The way to retrieve the NS entrypoint for secondary cores should be updated as well.

Fixes: f332e77c4b7c ("core: arm: refactor boot argument handling")


jenswi-linaro:
Good catch!

ChiaweiW:
Tag collected with minor commit message rephrasing.

PR #6513, opened 2023-11-30
plat-aspeed: ast2700: use gic_init_v3()

ChiaweiW:
Use gic_init_v3() with the GICR base address.


ChiaweiW:
Tag collected. Thanks!

PR #6514, opened 2023-11-30
plat-stm32mp1: increase default heap size when pager enabled

etienne-lms:
When pager is enabled on STM32MP15 variant, the default heap size was set 48kB (instead of default 64kB set from mk/config.mk) in order to relax pressure on pager page pool. Increase this platform default value to 52kB to pass some xtest regression test that may fail in recent OP-TEE release when RPMB_FS is enabled.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -315,7 +315,7 @@ CFG_LOCKDEP ?= n
 CFG_TA_BGET_TEST ?= n
 # Default disable early TA compression to support a smaller HEAP size
 CFG_EARLY_TA_COMPRESS ?= n
-CFG_CORE_HEAP_SIZE ?= 49152
+CFG_CORE_HEAP_SIZE ?= 53248
 endif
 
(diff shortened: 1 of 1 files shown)


GseoC:

For my sake, is that caused by heap fragmentation?

etienne-lms:
> For my sake, is that caused by heap fragmentation?

No. It is due to my poor sub-optimized implementation of the regulator supported voltage list implementation, recently merged. PMIC voltage list support allocates too many useless bytes in the heap leading to xtest regression 6018 (large object) to  fail when tested with RPMB support. I'll post a change on top of https://github.com/OP-TEE/optee_os/pull/6487 once it's reviewed and merged to address the issue. Afterward, the current 48kB heap is sufficient to pass xtest. I think I will discard to current heap enlarge change.

Thanks for the review anyway. I won't apply the review tags until I made my mind.

etienne-lms:
(@Jérôme, please don't laugh at me :sad:)

etienne-lms:
I'm closing this P-R. No need to increase MP1 default heap size up to now.

PR #6515, opened 2023-11-30
caam: provide plat_rng_init if CFG_WITH_SOFTWARE_PRNG=y 

a3f:
With CFG_NXP_CAAM_RNG_DRV enabled, OP-TEE will use the CAAM
to generate random numbers. Normal world access to the RNG is still
possible as the CAAM is TrustZone aware and provides multiple separate
job rings.

For complete isolation, however, access to CAAM reset and clocks need to
be managed as well. This could be done in theory by restricting access
to the reset and clock controller peripherals to the secure world and
exporting limited access to some resources via SCMI. There is no such
support yet for the i.MX and thus some setups may prefer to avoid using
the CAAM in OP-TEE to stay safe from normal world inducing glitches.

These setups may still need random numbers in OP-TEE. Therefore, access
so have them
access the CAAM only once at startup to initialize OP-TEE's PRNG and
defer subsequent use of the CAAM to the normal world, whenever
CFG_WITH_SOFTWARE_PRNG=y.

To improve RNG quality, the preparatory patches enable prediction resistance if
possible: This instructs the CAAM to wait if low on entropy, which will also benefit
users with CFG_WITH_SOFTWARE_PRNG=n.

This PR can be considered a continuation of both PR #3721 and PR #5145.

---

Cc: @clementfaure
Cc: @emantor

### DIFF
--- a/core/drivers/crypto/caam/caam_rng.c
+++ b/core/drivers/crypto/caam/caam_rng.c
@@ -567,6 +567,27 @@ enum caam_status caam_rng_init(vaddr_t ctrl_addr)
 }
 
 #ifdef CFG_NXP_CAAM_RNG_DRV
+#ifdef CFG_WITH_SOFTWARE_PRNG
+void plat_rng_init(void)
+{
+	TEE_Result res = TEE_SUCCESS;
+	uint8_t buf[64] = { };
+
+	res = do_rng_read(buf, sizeof(buf));
+	if (res) {
+		EMSG("Failed to read RNG: %#" PRIx32, res);
+		panic();
+	}
+
+	res = crypto_rng_init(buf, sizeof(buf));
+	if (res) {
+		EMSG("Failed to initialize RNG: %#" PRIx32, res);
+		panic();
+	}
+
+	RNG_TRACE("PRNG seeded from CAAM");
+}
+#else /* !CFG_WITH_SOFTWARE_PRNG */
 TEE_Result hw_get_random_bytes(void *buf, size_t blen)
 {
 	if (!buf)
@@ -578,4 +599,5 @@ TEE_Result hw_get_random_bytes(void *buf, size_t blen)
 void plat_rng_init(void)
 {
 }
-#endif
+#endif /* CFG_WITH_SOFTWARE_PRNG */
+#endif /* CFG_NXP_CAAM_RNG_DRV */
--- a/core/drivers/crypto/caam/caam_rng.c
+++ b/core/drivers/crypto/caam/caam_rng.c
@@ -60,6 +60,7 @@ struct rngdata {
 struct rng_privdata {
 	vaddr_t baseaddr;                       /* RNG base address */
 	bool instantiated;                      /* RNG instantiated */
+	bool pr_enabled;			/* RNG prediction resistance */
 	struct rngdata databuf[RNG_DATABUF_NB]; /* RNG Data generation */
 	uint8_t dataidx;                        /* Current RNG Data buffer */
 };
@@ -158,6 +159,10 @@ static enum caam_status prepare_gen_desc(struct rngdata *rng)
 {
 	paddr_t paddr = 0;
 	uint32_t *desc = NULL;
+	uint32_t op = RNG_GEN_DATA;
+
+	if (rng_privdata->pr_enabled)
+		op |= ALGO_RNG_PR;
 
 	/* Convert the buffer virtual address to physical address */
 	paddr = virt_to_phys(rng->data);
@@ -168,7 +173,7 @@ static enum caam_status prepare_gen_desc(struct rngdata *rng)
 
 	caam_desc_init(desc);
 	caam_desc_add_word(desc, DESC_HEADER(0));
-	caam_desc_add_word(desc, RNG_GEN_DATA);
+	caam_desc_add_word(desc, op);
 	caam_desc_add_word(desc, FIFO_ST(RNG_TO_MEM, rng->size));
 	caam_desc_add_ptr(desc, paddr);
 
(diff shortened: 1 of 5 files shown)


jforissier:
> Did I cause the IBART failure? I don't see how my change would break QEMU.

Nope, you just need to rebase onto the master branch to get some newly added code which is tested by `xtest 6021`.

core/drivers/crypto/caam/caam_rng.c
```diff
@@ -158,6 +159,11 @@ static enum caam_status prepare_gen_desc(struct rngdata *rng)
 {
 	paddr_t paddr = 0;
 	uint32_t *desc = NULL;
+	uint32_t op;
```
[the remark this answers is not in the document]
clementfaure: or `op = RNG_GEN_DATA;`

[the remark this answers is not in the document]
a3f: Turned it into an initialization.

core/drivers/crypto/caam/caam_rng.c
```diff
@@ -556,10 +554,8 @@ enum caam_status caam_rng_init(vaddr_t ctrl_addr)
 		retstatus = caam_rng_instantiation();
 	}
 
-#ifdef CFG_NXP_CAAM_RNG_DRV
```
clementfaure, line 547: `CFG_NXP_CAAM_RNG_DRV` enables or not the request of CAAM RNG data at boot or runtime. I cannot accept that modification.

core/drivers/crypto/caam/hal/imx_8q/hal_rng.c
```diff
@@ -18,3 +18,12 @@ enum caam_status caam_hal_rng_instantiated(vaddr_t baseaddr __unused)
...
 		return CAAM_NO_ERROR;
 }
+
+bool caam_hal_rng_pr_enabled(vaddr_t baseaddr __unused)
+{
+	/*
+	 * We have no direct access to the CAAM and can't assume prediction
+	 * resistance was enabled by the system controller firmware.
+	 */
+	return false;
```
clementfaure, line 28: No need to reach for the firmware. You can read `RNG_STA_PRx` instead

core/drivers/crypto/caam/hal/common/hal_rng.c
```diff
@@ -62,6 +62,12 @@ bool caam_hal_rng_key_loaded(vaddr_t baseaddr)
 	return io_caam_read32(baseaddr + RNG_STA) & RNG_STA_SKVN;
 }
 
+bool __weak caam_hal_rng_pr_enabled(vaddr_t baseaddr)
+{
+	return (io_caam_read32(baseaddr + RNG_STA) &
+		(RNG_STA_PR0 | RNG_STA_PR1)) == (RNG_STA_PR0 | RNG_STA_PR1);
```
clementfaure, line 68: You may not always have 2 state handles, depending on the platform.

clementfaure (commented):
What platform did you use for your testing?

a3f:
> What platform did you use for your testing?

This was tested on an i.MX7 with initialization done prior to OP-TEE by U-Boot.

core/drivers/crypto/caam/hal/imx_8q/hal_rng.c
a3f: Can I use the implementation `caam_hal_rng_pr_enabled` implementation in `hal_rng.c` unchanged?

core/drivers/crypto/caam/caam_rng.c
a3f: I can use CFG_WITH_SOFTWARE_PRNG instead. Would that be ok?

core/drivers/crypto/caam/hal/common/hal_rng.c
a3f: As far as OP-TEE is concerned, don't we always have two handles, because `RNG_DATABUF_NB == 2`?

core/drivers/crypto/caam/caam_rng.c
clementfaure: That would be better

core/drivers/crypto/caam/hal/imx_8q/hal_rng.c
clementfaure: Yes

core/drivers/crypto/caam/caam_rng.c
a3f: Done just that.

core/drivers/crypto/caam/caam_rng.c
```diff
@@ -556,6 +567,27 @@ enum caam_status caam_rng_init(vaddr_t ctrl_addr)
 }
 
 #ifdef CFG_NXP_CAAM_RNG_DRV
+#ifdef CFG_WITH_SOFTWARE_PRNG
+void plat_rng_init(void)
+{
+	TEE_Result res = TEE_SUCCESS;
```
clementfaure, line 573: We usually initialize `res` with an error like `TEE_ERROR_GENERIC`

clementfaure:
With my last comment addressed `Reviewed-by: Clement Faure <clement.faure@nxp.com>`

core/drivers/crypto/caam/caam_rng.c
jenswi-linaro: I usually use `TEE_SUCCESS`. The point is to get a predictable value.

clementfaure: Ok

a3f:
I've added the R-b, thanks for the review!

core/drivers/crypto/caam/caam_rng.c
jforissier: > My preference would be _not_ to initialize variables with unused values, so any modern compiler may warn if future code changes start using the value uninitialized.

That would be our choice if compilers weren't so poor at diagnosing "used but not initialized" variables. We have observed cases where such variables were not reported (by GCC at least).

a3f:
@clementfaure Are there any blockers still for this to get merged?

clementfaure:
@a3f Nope, all good for me. I don't have permissions to merge PRs, but @jforissier does!

jforissier:
@a3f @clementfaure I am merging this, thanks!

PR #6516, opened 2023-11-30
[RFC] core: gic: check affinity routing before GICR base

jenswi-linaro:
The physical address of the redistributor was, before this patch, always used, regardless of whether affinity routing is enabled or not. In the case where affinity routing isn't enabled there is no need to record the redistributor address, so check for affinity routing before recording the redistributor address.

Fixes: 05089e5f9a56 ("core: gic: use redistributor CPU interface")


glneo:
Nope, same panic:

```
Starting ATF on ARM64 core...

NOTICE:  BL31: v2.10.0	(debug):v2.10.0-42-g9d4819a09
NOTICE:  BL31: Built : 07:42:44, Nov 30 2023
INFO:    GICv3 without legacy support detected.
INFO:    ARM GICv3 driver initialized in EL3
INFO:    Maximum SPI INTID supported: 287
INFO:    SYSFW ABI: 3.1 (firmware rev 0x0009 '9.1.5--v09.01.05 (Kool Koala)')
INFO:    BL31: Initializing runtime services
INFO:    BL31: cortex_a53: CPU workaround for erratum 855873 was applied
INFO:    BL31: cortex_a53: CPU workaround for erratum 1530924 was applied
INFO:    BL31: Initializing BL32
I/TC: 
I/TC: OP-TEE version: 4.0.0-130-g7e629cc75 (gcc version 11.2.1 20220111 (GNU Toolchain for the Arm Architecture 11.2-2022.02 (arm-11.14))) #1 Thu Nov 30 13:42:48 UTC 2023 aarch64
E/TC:0 0 Panic 'GICR_BASE missing for affinity routing' at core/drivers/gic.c:500 <gic_init_base_addr>
E/TC:0 0 TEE load address @ 0x9e800000
E/TC:0 0 Call stack:
E/TC:0 0  0x9e8080f8
E/TC:0 0  0x9e81edc8
E/TC:0 0  0x9e81c694
E/TC:0 0  0x9e807ee0
```

Maybe we could define the redistributor address, but this would only be a fix for K3, other platforms would need the same fix. Another fix would be to use some automatic detection for the GICR address when it is not defined by a platform like I did in TF-A:

https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/tree/plat/ti/k3/common/k3_gicv3.c#n47

Interestingly there is a comment just below that which states

```
The GICv3 driver is initialized in EL3 and does not need
to be initialized again in SEL1. This is because the S-EL1
can use GIC system registers to manage interrupts and does
not need GIC interface base addresses to be configured.
```

It has been some time since I had to do anything with GICv3 so I'm not sure the validity of this statement. Is this possible and what do we lose if we skip this GIC setup in OP-TEE?

jenswi-linaro:
We have recently added support for donating a secure SGI to the normal world. We can't support this with affinity routing enabled without access to the redistributor. This isn't a big loss since the platform anyway needs to be updated to use the feature.

Thanks for testing, I'll rework the patch.

jenswi-linaro:
Update to make redistributor address optional until it's needed by gic_init_donate_sgi_to_ns().

Adding "core: gic: allow gic version 1".

etienne-lms (commented):
"core: gic: allow gic version 1".

Commit "core: gic: make GCIR base optional" looks good to me:

glneo:
This seems to fix the panic(), thanks

`Tested-by: Andrew Davis <afd@ti.com>`

jenswi-linaro:
Tags applied

jforissier:
s/allow gic/allow GIC/ in commit subject and also please rebase to make sure IBART is happy.
This still has an RFC tag, @jenswi-linaro do you need further acks before this gets merged?

sahilnxp:
s/make GCIR base/make GICR base/ in commit  https://github.com/OP-TEE/optee_os/pull/6516/commits/660e8f012a313065c3f321671ef7395bb95b44eb
Also replace the GCIR in commit description also.

jenswi-linaro:
Rebased and commit messages updated

sahilnxp:
image
Please correct this one also.

With this addressed, `Acked-by: Sahil Malhotra <sahil.malhotra@nxp.com>`

PR #6518, opened 2023-12-01
core: ffa: fix FFA_NOTIFICATION_GET vm_id

jenswi-linaro:
handle_notification_get() has until now read the receiver endpoint ID from the upper 16 bits of w1, but the receiver endpoint ID is passed in the lower 16 bits of w1 passed to FFA_NOTIFICATION_GET. So fix the function to read the lower 16 bits instead.

Fixes: 2e02a7374b86 ("core: ffa: add notifications with SPMC at S-EL1")

### DIFF
--- a/core/arch/arm/kernel/thread_spmc.c
+++ b/core/arch/arm/kernel/thread_spmc.c
@@ -1517,7 +1517,7 @@ static void handle_notification_get(struct thread_smc_args *args)
 		w2 = 0;
 		goto out;
 	}
-	vm_id = FFA_SRC(args->a1);
+	vm_id = FFA_DST(args->a1);
 
 	old_itr_status = cpu_spin_lock_xsave(&spmc_notif_lock);
 


core/arch/arm/kernel/thread_spmc.c
```diff
@@ -1517,7 +1517,7 @@ static void handle_notification_get(struct thread_smc_args *args)
 		w2 = 0;
 		goto out;
 	}
-	vm_id = FFA_SRC(args->a1);
+	vm_id = FFA_DST(args->a1);
```
etienne-lms, line 1520: Use of macro `FFA_DST()` looks a bit strange as this command ABI does not use SRC/DST in respectively upper/lower 16bit part of the value passed . An explicit 16bit LSB extract instruction would look more appropriate.

jenswi-linaro: This follows the same logic as in for instance `handle_notification_unbind()` and `handle_notification_bind()` above.
Would it help if I added the comment `/* The destination of the notification */`?

etienne-lms: Ok, makes sense. Discard my comment.

PR #6524, opened 2023-12-04
plat-mediatek: disable console when CFG_TEE_CORE_LOG_LEVEL is 0

massonju:
Following our usage, it can be interesting to disable the console, typically for a final product.

### DIFF
--- a/core/arch/arm/plat-mediatek/main.c
+++ b/core/arch/arm/plat-mediatek/main.c
@@ -12,8 +12,10 @@
 #include <platform_config.h>
 #include <stdint.h>
 
+#if (CFG_TEE_CORE_LOG_LEVEL != 0)
 register_phys_mem_pgdir(MEM_AREA_IO_NSEC,
 			CONSOLE_UART_BASE, SERIAL8250_UART_REG_SIZE);
+#endif
 
 static struct serial8250_uart_data console_data;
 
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-mediatek/main.c
```diff
@@ -12,8 +12,10 @@
 #include <platform_config.h>
 #include <stdint.h>
 
+#if defined(CFG_CONSOLE_ENABLE)
```
jforissier, line 15: `#if (CFG_TEE_CORE_LOG_LEVEL != 0)`

core/arch/arm/plat-mediatek/main.c
```diff
@@ -33,7 +35,9 @@ void boot_primary_init_intc(void)
 
 void console_init(void)
 {
+#if defined(CFG_CONSOLE_ENABLE)
```
jforissier, line 38:
```
if (CFG_TEE_CORE_LOG_LEVEL != 0) {
    ...
}
```

core/arch/arm/plat-mediatek/main.c
massonju: Yes much better thanks for the suggestion

core/arch/arm/plat-mediatek/main.c
massonju: what about using `#if (CFG_TEE_CORE_LOG_LEVEL != 0)` here ?

I guess we make the assumption that the compiler will detect/remove the code around bracket ?

jforissier: > what about using `#if (CFG_TEE_CORE_LOG_LEVEL != 0)` here ?
> 
> I guess we make the assumption that the compiler will detect/remove the code around bracket ?

Exactly. It is a general rule to to improve compile test coverage, see 776d21586. The Linux kernel has a similar policy.

jenswi-linaro:

PR #6525, opened 2023-12-04
scripts: checkpatch: fix word splitting in the command

clementfaure:
The following would fail with Zsh instead of Bash : $ source ./scripts/checkpatch_inc.sh
$ checkpatch HEAD
Unknown option: typedefsfile typedefs.checkpatch
Usage: .../scripts/checkpatch.pl [OPTION]... [FILE]

By setting xtrace in the _checkpatch() function, we can see the built command is interpreted differently depending on the shell:
	$CHECKPATCH $CHECKPATCH_OPT $typedefs_opt -
In Zsh:
	/scripts/checkpatch.pl '--typedefsfile typedefs.checkpatch' -
In Bash:
	/scripts/checkpatch.pl --typedefsfile typedefs.checkpatch -

Bash differs from Zsh when it comes to word splitting for unquoted parameters expansions.
One solution is to use `eval` to execute the built command.

### DIFF
--- a/scripts/checkpatch_inc.sh
+++ b/scripts/checkpatch_inc.sh
@@ -23,7 +23,7 @@ function _checkpatch() {
 				typedefs_opt="";
 		# Ignore NOT_UNIFIED_DIFF in case patch has no diff
 		# (e.g., all paths filtered out)
-		$CHECKPATCH $CHECKPATCH_OPT $typedefs_opt -
+		eval "$CHECKPATCH $CHECKPATCH_OPT $typedefs_opt -"
 }
 
 function checkpatch() {


jforissier:
How is this an issue since `scripts/checkpatch_inc.sh` is used only from `scripts/checkpatch.sh` which has `#!/bin/bash`? Do you intend to use it in a different environment?
I don't mind this change, but since we don't use zsh in this repo we won't be able to ensure that zsh support won't break which is a bit annoying.

jenswi-linaro:
I don't see why we need `scripts/checkpatch_inc.sh` at all. Why not copy the content into `scripts/checkpatch.sh`?

I'm a Zsh user myself, but if a script uses `#!/bin/bash` to interpret I'd let Bash do that.

jforissier:
> I don't see why we need `scripts/checkpatch_inc.sh` at all. Why not copy the content into `scripts/checkpatch.sh`?

I was mistaken. Actually `scripts/checkpatch_inc.sh` is used by CI (`.github/workflows/ci.yml`) and that is the reason why the functions were split, see 1c93c2b5e. But I suppose we could just invoke the `checkpatch.sh` script with the proper syntax and get the same result.

clementfaure:
> How is this an issue since `scripts/checkpatch_inc.sh` is used only from `scripts/checkpatch.sh` which has `#!/bin/bash`? Do you intend to use it in a different environment? I don't mind this change, but since we don't use zsh in this repo we won't be able to ensure that zsh support won't break which is a bit annoying.

I agree, it does not make sense to support Zsh when the shebang says it's Bash. See that modification as a small fix where it's a better practice to use `eval` for building command from string concatenation, that also happens to solve an issue that script had with Zsh.

I get the error when I `source scripts/checkpatch_inc.sh` and do `checkpatch|checkdiff... HEAD`. In this case, the function is executed by Zsh and not Bash

jforissier:

PR #6531, opened 2023-12-06
core: ffa: return fpi size from FFA_PARTITION_INFO_GET

jenswi-linaro:
Until now has FFA_PARTITION_INFO_GET always returned zero in w3, but FF-A v1.1 requires FFA_PARTITION_INFO_GET to return the size of each partition information descriptor returned if
FFA_PARTITION_INFO_GET_COUNT_FLAG isn't set. So fix this by returning the size of a FF-A v1.1 partition information descriptor in w3.

Fixes: a1c53023cc80 ("core: spmc: support FF-A 1.1")


etienne-lms (commented):

jenswi-linaro:
Tag applied

Issue #4783, opened 2021-08-06
PL310 cache opeartions can't take effect on non-secure cacheline if it's specific to the PA


tiger20081015:
Hi, we recently met an issue in our platform(CA9 with PL310 enabled): 
We have a hardware crypto IP in secureos. When the IP DMA's output buffer is a dynamic shared memory allocated from REE with non-secure attribution,  the cpu always gets the output buffer's old data even if we have already done the cache invalidate operation of the ouput buffer's physical address before.  If PL310 is  disabled or the memory attribution is changed from non-secure to secure, the issue disappears. 

We found the explnation in the PL310's spec(DDI0246H_l2c310_r3p3_trm.pdf) as below:
![image](https://user-images.githubusercontent.com/35290009/128447317-c25dd6f5-dc37-4592-abaf-daea8f64d00e.png)

From the spec, it seems in secureos the PL310 cache maintanance can't take effect to non-secure memory if it's specfic to PA. In secureos there are several types of non-secure memory that are impacted by this feature , like MEM_AREA_RAM_NSEC, MEM_AREA_SHM_VASPACE.  Do you have any mind to solve this issue? The set/way solution is likely to have performance impact and PL310 seems has no configuration to change its behavior.  Thanks!

tiger20081015:
hi, @etienne-lms , I noticed you submitted the PL310 related patches. Could you please give some advice about solving the problem? Thanks in advance!

Emantor:
The only way I see to get around this problem is to use a secure world buffer for DMA operations and copying the data after the crypto IP is finished. This way you don't DMA into a buffer which you can't operate with cache maintenance ops on. Obviously this will have a performance impact.

tiger20081015:
@Emantor, thanks for your reply. For the issue I met, as you metioned, using a secure buffer to copy data is donable. But since non-secure types of memory like MEM_AREA_RAM_NSEC, MEM_AREA_SHM_VASPACE all have the same problem, it seems quite difficult to find all the places using these memory and apply the copy buffer solution?

etienne-lms:
Hi @tiger20081015,
I think @Emantor suggested a reliable scheme: in secure world, the operation using DMAs would performed using a temporary secure buffer and once DMA is done, secure world copies (after cache sync) the output secure data to the target output shared memory. This should be needed only for the very specific operations (in OP-TEE) where you use DMA to fill non-secure memory, it would not need changes at all places where OP-TEE uses non-secure memory.

[...]

PR #6535, opened 2023-12-08
core: arm: allow cache_op_outer() to operate on non-secure buffers

clementfaure:
Hello!
I'm digging out an old issue with the PL310 cache controller : https://github.com/OP-TEE/optee_os/issues/4783
I'm not sure if I proceeded well for platform specific implementation.
Let me know what you think!

### DIFF
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -14,6 +14,7 @@
 #include <kernel/tee_misc.h>
 #include <kernel/tlb_helpers.h>
 #include <kernel/tz_ssvce_pl310.h>
+#include <mm/core_memprot.h>
 #include <mm/core_mmu.h>
 #include <platform_config.h>
 #include <trace.h>
@@ -99,6 +100,26 @@ TEE_Result cache_op_outer(enum cache_op op, paddr_t pa, size_t len)
 {
 	TEE_Result ret = TEE_SUCCESS;
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_FOREIGN_INTR);
+	vaddr_t pl310_base_pa_op = 0;
+
+	/*
+	 * According the ARM PL310 documentation, if the operation is specific
+	 * to the PA, the behavior is presented in the following manner:
+	 * - Secure access: The data in the cache is only affected by the
+	 * operation if it is secure.
+	 * - Non-secure access: The data in the cache is only affected by the
+	 * operation if it is non-secure.
+	 *
+	 * https://developer.arm.com/documentation/ddi0246/a/programmer-s-model/register-descriptions/register-7--cache-maintenance-operations
+	 *
+	 * Depending on the buffer location, use the secure or non-secure PL310
+	 * base address to do physical address based cache operation on the
+	 * buffer.
+	 */
+	if (tee_pbuf_is_sec(pa, len))
+		pl310_base_pa_op = pl310_base();
+	else
+		pl310_base_pa_op = pl310_nsbase();
 
 	tee_l2cc_mutex_lock();
 	switch (op) {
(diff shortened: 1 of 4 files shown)


core/arch/arm/mm/core_mmu.c
```diff
@@ -107,21 +128,22 @@ TEE_Result cache_op_outer(enum cache_op op, paddr_t pa, size_t len)
 		break;
 	case DCACHE_AREA_INVALIDATE:
 		if (len)
-			arm_cl2_invbypa(pl310_base(), pa, pa + len - 1);
+			arm_cl2_invbypa(pl310_base_pa_op, pa, pa + len - 1);
```
jenswi-linaro, line 131: Doing `DCACHE_AREA_INVALIDATE` on non-secure memory seems a bit risky. How can we be sure not to break anything in normal world?

clementfaure: An invalidate operation is always risky if used carelessly. How to prevent the user from a misuse ? I'm not sure that we can prevent it.

jenswi-linaro: Why must we do this instead of using a secure bounce buffer? I assume it's for performance, but is it needed that bad? Do we have any use cases where the performance would be unacceptable?

clementfaure: Some crypto operations failure happen when the CAAM is enabled on platforms with PL310 controllers because of the missing cache management for output buffers. 
The CAAM driver implements SGT management to improve performance and avoid data copy. A secure bounce buffer would impact performance and defeat the purpose of the SGT.
If a developer is not comfortable using cache invalidation, it's up to the driver to implement a secure bounce buffer. However, I think developers should have the choice between the two approaches.

jenswi-linaro: Perhaps it's possible to do both? If buffers are cache-line aligned then we can work directly with them, otherwise a secure bounce buffer is needed.

clementfaure: The CAAM already implements that feature
https://github.com/OP-TEE/optee_os/blob/dc6563d7d6a4a1ab83e506aa101819a88efe418c/core/drivers/crypto/caam/utils/utils_dmaobj.c#L336-L346

core/arch/arm/plat-zynq7k/main.c
```diff
@@ -107,6 +107,16 @@ vaddr_t pl310_base(void)
 	return PL310_BASE;
 }
 
+vaddr_t pl310_nsbase(void)
```
jenswi-linaro, line 110: How about providing this as a common __weak function instead? Since all the other related functions are in `core/arch/arm/kernel/tz_ssvce_pl310_a32.S` I suppose it would be more convenient to implement in assembly in that file.

clementfaure: Yes, that would be better

clementfaure:
Force-push : address comments + rebase

jenswi-linaro:
For the commit "core: add pl310_nsbase() function prototype", please mention the added weak function and the purpose of it.

clementfaure:
> For the commit "core: add pl310_nsbase() function prototype", please mention the added weak function and the purpose of it.

In the header file or commit message?

jenswi-linaro:
> > For the commit "core: add pl310_nsbase() function prototype", please mention the added weak function and the purpose of it.
> 
> In the header file or commit message?

In the commit message, to clarify that it's more than just a prototype.

clementfaure:
> > > For the commit "core: add pl310_nsbase() function prototype", please mention the added weak function and the purpose of it.
> > 
> > 
> > In the header file or commit message?
> 
> In the commit message, to clarify that it's more than just a prototype.

Right! Done

jenswi-linaro (commented):

clementfaure:
Any blocker? @jforissier

jforissier:
@clementfaure nope, thanks for the reminder :wink:

PR #6536, opened 2023-12-08
PKCS11: add the possibility of giving rights to a TA when we use ACL to manage PKCS11 acces

sah-cdo:
We need to push RSA/ECC keys from a custom TA to PKCS11 TA
The custom TA will decrypt the secret based on chipset secret unique key and push it to PKCS11 TA

When we use ACL to manage access to PKCS11 slot, we must be able to give rigths
to the custom TA for  pushing the keys in the slot.
This is needed because it's not possible to pass from a pincode  to acl mode

### DIFF
--- a/ta/pkcs11/include/pkcs11_ta.h
+++ b/ta/pkcs11/include/pkcs11_ta.h
@@ -937,7 +937,7 @@ enum pkcs11_user_type {
  * - PIN value: group:<client UUID string>
  *   - TEE group login with client UUID matching group credentials
  * - PIN value: ta:<client UUID string>
- *   - TEE trusted app login with client UUID matching TA UUID
+ *   - TEE trusted application login with client UUID matching TA UUID
  */
 
 /* Keywords for protected authenticated path PIN parser */
--- a/ta/pkcs11/include/pkcs11_ta.h
+++ b/ta/pkcs11/include/pkcs11_ta.h
@@ -936,12 +936,15 @@ enum pkcs11_user_type {
  *   - TEE user login with client UUID matching user credentials
  * - PIN value: group:<client UUID string>
  *   - TEE group login with client UUID matching group credentials
+ * - PIN value: ta:<client UUID string>
+ *   - TEE trusted app login with client UUID matching TA UUID
  */
 
 /* Keywords for protected authenticated path PIN parser */
 #define PKCS11_AUTH_TEE_IDENTITY_PUBLIC	"public"
 #define PKCS11_AUTH_TEE_IDENTITY_USER	"user:"
 #define PKCS11_AUTH_TEE_IDENTITY_GROUP	"group:"
+#define PKCS11_AUTH_TEE_IDENTITY_TA 	"ta:"
 
 /*
  * Values for 32bit session flags argument to PKCS11_CMD_OPEN_SESSION
--- a/ta/pkcs11/src/persistent_token.c
+++ b/ta/pkcs11/src/persistent_token.c
@@ -193,6 +193,9 @@ enum pkcs11_rc setup_identity_auth_from_pin(struct ck_token *token,
 		} else if (strstr(acl_string, PKCS11_AUTH_TEE_IDENTITY_GROUP) ==
 			   acl_string) {
 			identity.login = TEE_LOGIN_GROUP;
+		} else if (strstr(acl_string, PKCS11_AUTH_TEE_IDENTITY_TA) ==
+			   acl_string) {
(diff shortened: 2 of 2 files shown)


ta/pkcs11/include/pkcs11_ta.h
```diff
@@ -936,12 +936,15 @@ enum pkcs11_user_type {
  *   - TEE user login with client UUID matching user credentials
  * - PIN value: group:<client UUID string>
  *   - TEE group login with client UUID matching group credentials
+ * - PIN value: ta:<client UUID string>
+ *   - TEE trusted app login with client UUID matching TA UUID
```
etienne-lms, line 940: s/app/application/

sah-cdo:
@sah-cdo, do you intend to propose a TA library to implement the cryptoki API for TAs to use the pkcs11 TA services?

For the moment, I have no crypto APIs for TA to use PKCS11 TA services.
I use directly `TEE_OpenTASession`, `TEE_InvokeTACommand` and `TEE_CloseTASession` with CMD declared in ta/pkcs11/include/pkcs11_ta.h to import the secret in pkcs11 TA in the custom TA 
@etienne-lms do you think another sets of APIs are needed ?

etienne-lms:
> @etienne-lms do you think another sets of APIs are needed ?

No, I think it's fair a TA can access a pkcs11 TA slot/token using the exposed TA commands API.
I thought having the cryptoki API ported as an OP-TEE TA library could interest people, maybe some day.

vesajaaskelainen:
> The change looks ok to me. Commit message to refine:
> 
> ```
> ta: pkcs11: support TA login in authentication login mode
> 
> Add the possibility of giving rights to a TA when we use ACL to manage
> PKCS11 access.
> 
> S-o-b: ...
> ```
> 
> @vesajaaskelainen, you you agree with such a change?
> 
> @sah-cdo, do you intend to propose a TA library to implement the cryptoki API for TAs to use the pkcs11 TA services?

This use case:
```
We need to push RSA/ECC keys from a custom TA to PKCS11 TA
The custom TA will decrypt the secret based on chipset secret unique key and push it to PKCS11 TA
```
...troubles me bit. Why would you want to do that? Why not leave the secret management to OP-TEE/PKCS#11 TA and let the device itself generate the keys using the RNG. Then you just make sure that on factory reset you can restore back to this state. In that way there is no risk that private key leaks out.

Let's assume TA authentication would be supported.

Now you have some additional problems -- eg. you need to have initialized the PKCS#11 token with correct authorization in order to allow the the keypair import functionality to function. Then after the key material import then in order to make the key material accessible to REE side or so you would need to alter the authentication so that it can work.

Another problem here is that we do not really have a ABI guarantee for internal communication between libckteec.so and pkcs11 TA. We only guarantee that those both taken from same release works together.

Now if you really want to go for this trek -- then I suppose the 'ta:<ta id>' is good choice for the access control.

Should we then include also test TA for the tests that performs authentication during xtest?

sah-cdo:
> Why not leave the secret management to OP-TEE/PKCS#11 TA and let the device itself generate the keys using the RNG
On the device, the secret is provisioned in flash during manufacturing process and encrypted by a key not accessible by the REE.

The choice of a custom TA was done to limit impact on OPTEE/PKCS11 TA sources.

> you need to have initialized the PKCS#11 token with correct authorization in order to allow the the keypair import functionality to function
It's the case, we initiate the slot with the rigth for our TA.

After import we use SO rigth to change USER rigth to appropriate group.

> Another problem here is that we do not really have a ABI guarantee for internal communication between libckteec.so and pkcs11 TA. We only guarantee that those both taken from same release works together.

It's make sense and it's not a problem for us.

> Should we then include also test TA for the tests that performs authentication during xtest?

It's not mandatory from my point of view.
If you concider the new test as mandatory I can help to add this test.

sah-cdo:
@vesajaaskelainen @etienne-lms  any feedback on my last comment?

etienne-lms:
I'm fine with such a change. I think a test in optee_test would be nice to show how a TA can get ownership of the token and maybe how it passes back this control back to a non-secure client, if applicable.

vesajaaskelainen:
> I'm fine with such a change. I think a test in optee_test would be nice to show how a TA can get ownership of the token and maybe how it passes back this control back to a non-secure client, if applicable.

It would be a good idea that there would be test TA that verifies that functionality works and does not get broken as otherwise we do not have anything integrated that would demonstrate that ACL works.

Probably simple dedicated TA in `optee_test/ta` folder for the purpose would be best.

PR #6538, opened 2023-12-08
Fix some warnings on SAM platforms

tprrt:
Fully disable the support of external device tree on SAM platforms because it isn't used, to fix the following warning:
`I/TC: Non-secure external DT found`

<s>And disable by default insecure configurations, to remove the following warning:
`I/TC: WARNING: This OP-TEE configuration might be insecure!`</s>

### DIFF
--- a/core/arch/arm/plat-sam/conf.mk
+++ b/core/arch/arm/plat-sam/conf.mk
@@ -39,6 +39,9 @@ $(call force,CFG_CORE_HAS_GENERIC_TIMER,n)
 $(call force,CFG_TZDRAM_START,0x20000000)
 $(call force,CFG_TZDRAM_SIZE,0x800000)
 
+# This value is forced because this feature isn't used by SAM platforms.
+$(call force,CFG_EXTERNAL_DT,n)
+
 CFG_MMAP_REGIONS ?= 24
 
 CFG_SHMEM_START  ?= 0x21000000


jenswi-linaro:
In the commit "plat-sam: disable CFG_INSECURE" please explain why this platform doesn't need `CFG_INSECURE=y` any longer. @nodeax do you agree?

nodeax:
> In the commit "plat-sam: disable CFG_INSECURE" please explain why this platform doesn't need `CFG_INSECURE=y` any longer. @nodeax do you agree?

Yes, I agree. The secure configuration is typically enabled via at91bootstrap + secure-samba flashing tools. So for OP-TEE to reliably determine if the end device is secure, op-tee needs to read the SECURE_MODE bit of Boot Configuration Word. This would be a runtime check. So I think to truly get rid of the warning, the check in [boot.c](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/boot.c#L1204) needs to be a platform specific callback function...

etienne-lms:
maybe
```c
if (!sam_is_in_secure_mode() && !IS_ENABLED(CFG_INSURE))
	 panic();

nodeax:
> maybe
> 
> ```c
> if (!sam_is_in_secure_mode() && !IS_ENABLED(CFG_INSURE))
> 	 panic();
> ```

Agree, that is another option and the stm32mp1 platform has implemented something similar:
https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-stm32mp1/shared_resources.c#L696

jforissier:
> By the way. If one builds with other config directives for the platform (gprof, hw rng off, test keys, mainline TA authent pub key, ...), should ` CFG_INSURE ?= y` rather be `CFG_INSURE ?= n`?

We don't want to make that too complicated. `CFG_INSECURE` is just there as a reminder that declaring a platform "secure" is a LOT of work and the default configuration certainly isn't suitable in most cases. In fact I don't think any platform should have it disabled upstream.

jenswi-linaro:
> > By the way. If one builds with other config directives for the platform (gprof, hw rng off, test keys, mainline TA authent pub key, ...), should ` CFG_INSURE ?= y` rather be `CFG_INSURE ?= n`?
> 
> We don't want to make that too complicated. `CFG_INSECURE` is just there as a reminder that declaring a platform "secure" is a LOT of work and the default configuration certainly isn't suitable in most cases. In fact I don't think any platform should have it disabled upstream.

+1, but I think it's useful with checks against `CFG_INSECURE` here and there, not just as a warning.

jforissier:
> it's useful with checks against `CFG_INSECURE` here and there, not just as a warning.

Agreed.

tprrt:
Hello,

So, if I understand well, I should drop the patch that disables insecure from this pull-request and open a new pull-request adding functions to check the secure boot status on the SAM platform.
Is that right?

Kind regards,
Thomas Perrot

jenswi-linaro (commented):

PR #6541, opened 2023-12-12
dts: add stm32mp15*-scmi.dts files for when RCC is secure

etienne-lms:
For legacy reason and compatibility with existing platforms embedding OP-TEE with RCC secure hardening being disabled, introduce _-scmi.dts_ for the 4 ST boards STM32MP15x: DK1, DK2, ED1 and EV1 where we enable RCC security require non-secure world to use SCMI resources. Add platform flavors `157x_XXX_SCMI` to ease DTS selection.

_stm32mp15*-<board>.dts_ applies an insecure RCC configuration. _stm32mp15*-<board>-scmi.dts_ applies the secure RCC configuration. This better reflects the configurations supported in the Linux kernel and U-Boot source trees.

### DIFF
--- a/core/arch/arm/plat-stm32mp1/conf.mk
+++ b/core/arch/arm/plat-stm32mp1/conf.mk
@@ -5,17 +5,25 @@ flavor_dts_file-157C_DHCOM_PDK2 = stm32mp157c-dhcom-pdk2.dts
 flavor_dts_file-157C_DK2 = stm32mp157c-dk2.dts
 flavor_dts_file-157C_ED1 = stm32mp157c-ed1.dts
 flavor_dts_file-157C_EV1 = stm32mp157c-ev1.dts
+flavor_dts_file-157A_DK1_SCMI = stm32mp157a-dk1-scmi.dts
+flavor_dts_file-157C_DK2_SCMI = stm32mp157c-dk2-scmi.dts
+flavor_dts_file-157C_ED1_SCMI = stm32mp157c-ed1-scmi.dts
+flavor_dts_file-157F_EV1_SCMI = stm32mp157c-ev1-scmi.dts
 
 flavor_dts_file-135F_DK = stm32mp135f-dk.dts
 
 flavorlist-cryp-512M = $(flavor_dts_file-157C_DK2) \
+		       $(flavor_dts_file-157C_DK2_SCMI) \
 		       $(flavor_dts_file-135F_DK)
 
-flavorlist-no_cryp-512M = $(flavor_dts_file-157A_DK1)
+flavorlist-no_cryp-512M = $(flavor_dts_file-157A_DK1) \
+			  $(flavor_dts_file-157A_DK1_SCMI)
 
 flavorlist-cryp-1G = $(flavor_dts_file-157C_DHCOM_PDK2) \
 		     $(flavor_dts_file-157C_ED1) \
-		     $(flavor_dts_file-157C_EV1)
+		     $(flavor_dts_file-157C_EV1) \
+		     $(flavor_dts_file-157C_ED1_SCMI) \
+		     $(flavor_dts_file-157C_EV1_SCMI)
 
 flavorlist-no_cryp-1G = $(flavor_dts_file-157A_DHCOR_AVENGER96)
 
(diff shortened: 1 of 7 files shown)


core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -5,17 +5,25 @@ flavor_dts_file-157C_DHCOM_PDK2 = stm32mp157c-dhcom-pdk2.dts
 flavor_dts_file-157C_DK2 = stm32mp157c-dk2.dts
 flavor_dts_file-157C_ED1 = stm32mp157c-ed1.dts
 flavor_dts_file-157C_EV1 = stm32mp157c-ev1.dts
+flavor_dts_file-157A_DK1_SCMI = stm32mp157a-dk1-scmi.dts
+flavor_dts_file-157C_DK2_SCMI = stm32mp157c-dk2-scmi.dts
+flavor_dts_file-157C_ED1_SCMI = stm32mp157c-ed1-scmi.dts
+flavor_dts_file-157F_EV1_SCMI = stm32mp157c-ev1-scmi.dts
```
jforissier, line 11: The change could be more local if you used an additional variable (i.e., no change to the platform flavors). For example `ifeq (y,$(CFG_STM32MP1_RCC_SECURE))` ... append the `-scmi` part to the file name ... `endif`. The test could even use some existing `CFG_SCMI_*` variable perhaps?

GseoC:
Hi @etienne-lms, Copyrights and dates can be updated to newer version. Also, "authors" field can be either removed or updated with your name.

Can't we have a dependency between CFG_STM32MP1_RCC_SECURE and these boards?

core/arch/arm/plat-stm32mp1/conf.mk
GseoC: The downside of this would be to specify `CFG_STM32MP1_RCC_SECURE=y` in the command line but I guess that's ok.

etienne-lms: We expect the DT sets the platform configuration. So I would prefer have specific DTSes for specific configuration, instead of a config switch that would override the DT content.

GseoC: I guess Jérome's comment still stand as we can keep the flavor of the non-scmi DT and add the `-scmi` for the `CFG_EMBED_DTB_SOURCE_FILE`. Anyway, not a strong opinion.

etienne-lms: I could remove the `157?_*_SCMI` flavors but I still need to enumerate in some flavorlist to get `CFG_STM32MP15` and other switch set (incl. DDR size). In the end, it is simpler to define `PLATFORM_FLAVOR` values.

jforissier: > I could remove the `157?_*_SCMI` flavors but I still need to enumerate in some flavorlist to get `CFG_STM32MP15` and other switch set (incl. DDR size). In the end, it is simpler to define `PLATFORM_FLAVOR` values.

I don't understand this, because you are only adding the new flavors to lists where the non-SCMI flavor is already present. Can you give an example?

etienne-lms: I rather have mp15xx-xxx.dts files in opte_os related to mp15xx-xxx.dts in Linux/U-Boot,
and mp15xx-xxx-scmi.dts in optee_os for the bards mp15xx-xxx-scmi.dts defined in Linux/U-Boot.

Being listed in a flavour list prefers one from providing fewother needed config switches:
https://github.com/OP-TEE/optee_os/blob/4.0.0/core/arch/arm/plat-stm32mp1/conf.mk#L53-L74
https://github.com/OP-TEE/optee_os/blob/4.0.0/core/arch/arm/plat-stm32mp1/conf.mk#L153-L156

jforissier: > I rather have mp15xx-xxx.dts files in opte_os related to mp15xx-xxx.dts in Linux/U-Boot,
and mp15xx-xxx-scmi.dts in optee_os for the bards mp15xx-xxx-scmi.dts defined in Linux/U-Boot.

 I am fine with having separate DTS files for the "RCC is secure" case.

> Being listed in a flavour list prefers one from providing fewother needed config switches: https://github.com/OP-TEE/optee_os/blob/4.0.0/core/arch/arm/plat-stm32mp1/conf.mk#L53-L74 https://github.com/OP-TEE/optee_os/blob/4.0.0/core/arch/arm/plat-stm32mp1/conf.mk#L153-L156

Ah, here's the problem I think. IMO `$(flavorlist-xxx)` should be a flavor list (!), not a list of DTS files. See for example: https://github.com/OP-TEE/optee_os/blob/4.0.0/core/arch/arm/plat-imx/conf.mk#L1-L235.
Remember that the platform flavor is visible as a C macro (`PLATFORM_FLAVOR_<flavorname>`) so changing it can  impact on the code.

This is just about maintenance, so if you think your proposal is more convenient, I'm OK with that.

etienne-lms:
@jforissier, @GseoC, is the current change state ok to you of do you have some comments?

GseoC:
No strong opinion on the comments above so:

jforissier:
> Hi @etienne-lms, Copyrights and dates can be updated to newer version. 

No. The original copyright date must be preserved. 2023 may be added if changes are made in 2023, if it is not the case there is no need to add it just because the file is committed to this repository on this year.
Note also the "All rights reserved" mention should preferably be avoided.

> Also, "authors" field can be either removed or updated with your name.

OK.

> Can't we have a dependency between CFG_STM32MP1_RCC_SECURE and these boards?

Yeah, here is what I had in mind: #6555.

jforissier:

etienne-lms:
> (preferably with my comments on the Copyrights addressed)

The date and owner name in the DTS file are wrong in my 1st commit, hence change in the fixup commit.
I'll remove the "All right reserved" sentence.

etienne-lms:
Commit fixed and tags applied.

PR #6542, opened 2023-12-13
core: arm: collect all .nex_nozi sections

jenswi-linaro:
.nex_nozi.xxx was prior to this patch not gathered together with the other no-zero-initialize sections used by the nexus in a ns-virtualisation configuration. Fix this by gathering .nex_nozi .nex_nozi.* in .nex_nozi.


jenswi-linaro:
Nevermind, I'll need to look further into this.

PR #6547, opened 2023-12-14
STM32 GPIO: add secure pin configuration via device tree

GseoC:
This P-R adds support for configuring certain pins of a GPIO bank vio the "st,protreg" property.
Pins of peripherals used by the secure-world should be secured.
Other pins stay non-secure.


core/include/dt-bindings/gpio/stm32mp_gpio.h
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (C) STMicroelectronics 2023 - All Rights Reserved
+ */
+
+#ifndef _DT_BINDINGS_GPIO_STM32MP_GPIO_H
+#define _DT_BINDINGS_GPIO_STM32MP_GPIO_H
+
+#include <dt-bindings/gpio/gpio.h>
+
+/* Macro to define the security for GPIO */
+#define TZPROT(id) (1 << (id))
```
jenswi-linaro, line 12: This can give unexpected results when `id` is 31 since `1` is signed, compare this with the BIT() macro.

core/drivers/stm32_gpio.c
```diff
@@ -900,16 +920,44 @@ static TEE_Result stm32_pinctrl_dt_get(struct dt_pargs *pargs,
 }
 #endif /*CFG_DRIVERS_PINCTRL*/
 
+static void __unused stm32_gpio_get_conf_sec(struct stm32_gpio_bank *bank)
+{
+	if (bank->sec_support) {
```
GseoC, line 925: looking back at these APIs, I'll change the patch so that I use the existing ones that configure each pin instead of applying a mask on the bank.

core/include/dt-bindings/gpio/stm32mp_gpio.h
GseoC: Hi Jens,
since this will be used in the device-tree I cannot use types or U().
Is there really a notion of signs in the device-tree files?

jenswi-linaro: I don't know, perhaps it's like assembly instead. Anyway, this macro is available to C code and as you can see in the BIT() macro implementation we deal with assembly too.

GseoC: I'm not an expert on device tree pre-processing but I guess there's a Cpp process to handle things such as includes. There's a usecase of I2C_TEN_BIT_ADDRESS that is (1 << 31) in `include/dt-bindings/i2c/i2c.h` that is then used by I2C OF core but is quite old.

FYI, STM32 GPIOs aren't expected to exceed 16 GPIOs per bank so id will stay < 16.

jenswi-linaro: Perhaps C compatibility isn't a thing for .h files consumed by device-tree files. But I find it odd to have then have them available in the include path for C files.

jenswi-linaro: I can ignore this if others think this is OK.

GseoC:
Updated with use of existing APIs and comment clearing.
I didn't make a change on the macro @jenswi-linaro as I think it's ok in our usecase. Thank you for your input, let's see if others agree.

core/drivers/stm32_gpio.c
```diff
@@ -904,12 +924,33 @@ static TEE_Result stm32_pinctrl_probe(const void *fdt, int node,
...
+			assert(bank->ngpios <= GPIO_PIN_MAX + 1);
+
+			for (i = 0; i < bank->ngpios; i++) {
+				if (!(bank->seccfgr & BIT(i)))
+					continue;
+
+				stm32_gpio_set_secure_cfg(bank->bank_id, i,
+							  true);
+			}
+
```
GseoC, line 949: Maybe I can remove those empty lines

core/include/dt-bindings/gpio/stm32mp_gpio.h
jforissier: `.dts` files are preprocessed with `$(CROSS_COMPILE) -x assembler-with-cpp ...` so they have `__ASSEMBLER__` defined. For what we want here this is enough, because it selects the proper form of the `BIT()` macro that `dtc` will understand. If some day we need more clever things we could add `-D__DTS__` to the preprocessor command (like Linux does) so we could differentiate even more between C, assembler and DT.

jenswi-linaro: Given that it isn't hard to do the right thing, please fix this.

core/drivers/stm32_gpio.c
```diff
@@ -649,6 +654,19 @@ static TEE_Result dt_stm32_gpio_bank(const void *fdt, int node,
 	bank->bank_id = dt_get_bank_id(fdt, node);
 	bank->clock = clk;
 	bank->gpio_chip.ops = &stm32_gpio_ops;
+	bank->sec_support = compat->secure_control;
+	if (bank->sec_support) {
+		/* Secure configuration */
+		bank->base = io_pa_or_va_secure(&pa_va, blen);
+		cuint = fdt_getprop(fdt, node, "st,protreg", NULL);
+		if (cuint)
+			bank->seccfgr = fdt32_to_cpu(*cuint);
```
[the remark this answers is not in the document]
GseoC: I'll remove seccfgr field for now until pm upstream.

Let me propose a rework of this loop

GseoC:
Updated with comments + added  [41baf99](https://github.com/OP-TEE/optee_os/pull/6547/commits/41baf999a7db4dae444898cf382a0c0561b21562)

etienne-lms (commented):
Looks good to me. Can you squash the fixup commits?

GseoC:
Squashed review commits

core/drivers/stm32_gpio.c
```diff
@@ -725,8 +734,30 @@ static TEE_Result dt_stm32_gpio_pinctrl(const void *fdt, int node,
 
 			STAILQ_INSERT_TAIL(&bank_list, bank, link);
 
-			if (IS_ENABLED(CFG_STM32MP13))
-				set_bank_gpio_non_secure(bank);
+			DMSG("Registered GPIO bank %c (%d pins) @%lx",
```
etienne-lms, line 737:
```patch
-			DMSG("Registered GPIO bank %c (%d pins) @%lx",
+			DMSG("Registered GPIO bank %c (%u pins) @%#"PRIxVA,

etienne-lms (commented):
"drivers: stm32_gpio: fix coding style issues", 
"drivers: stm32_gpio: add secure configuration for GPIOs" (with comment addressed),
"dt-bindings: add TZPROT macro",
"dts: stm32: secure pins for peripherals used in the secure world" and
"drivers: stm32_gpio: fix iteration in set_bank_gpio_non_secure()" (prefer in first position in the series).

core/drivers/stm32_gpio.c
GseoC: I'll reorder `dt-bindings: add TZPROT` macro and `drivers: stm32_gpio: add secure configuration for GPIOs` to ensure every patch compiles as well

GseoC:
Tag applied, thanks for the review

PR #6548, opened 2023-12-15
core: fixup of transfer list header size

raymo200915:
Add 4 reserved bytes at the tail of the transfer list header. This fixes a non-8-bytes aligned header when "flags" was introduced into the header.

Fixes: 508e2476b232 ("core: update transfer list header and signature")


raymo200915:
Added acked-by tag.

PR #6550, opened 2023-12-15
ta: pkcs11: fixes for check value attribute

etienne-lms:
Return `PKCS11_CKR_ATTRIBUTE_VALUE_INVALID` instead of a template inconsistency when the key check value attribute is wrong due to its size.

Preserve original object attributes when C_SetAttributeValue service fails instead of possibly changing object attributes before the whole new attribute set is validated.

### DIFF
--- a/ta/pkcs11/src/object.c
+++ b/ta/pkcs11/src/object.c
@@ -988,6 +988,8 @@ enum pkcs11_rc entry_set_attribute_value(struct pkcs11_client *client,
 	size_t template_size = 0;
 	struct pkcs11_object *obj = NULL;
 	struct obj_attrs *head = NULL;
+	struct obj_attrs *head_new = NULL;
+	struct obj_attrs *head_old = NULL;
 	uint32_t object_handle = 0;
 	enum processing_func function = PKCS11_FUNCTION_MODIFY;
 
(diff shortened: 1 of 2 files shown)


maroueneboubakri:
``` Tested-by: Marouene Boubakri <marouene.boubakri@nxp.com> ```

File: core/arch/arm/plat-vexpress/conf.mk [hunk 1]
```diff
@@ -162,3 +162,4 @@ endif
 endif
 
 CFG_PKCS11_TA ?= y
+CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE ?= y
```
jenswi-linaro, line 165: Why isn't this the global default value?

etienne-lms: Previous implementation of the pkcs11 TA did provide that support. The added attribute provides a very little bit of hint on the secret value. For generic backward compatibility, we've set it default disabled in _mk/config.mk_.
I would rather ask @vesajaaskelainen if it's fine to default enable that feature.

vesajaaskelainen: Let's split this comment a bit.

As for whether it should be enabled by default or by "some random machine" then I agree that it should be enabled whenever feature is available so there is no need to make a guess whether feature works in this particular hardware or not.

Now about the CKA_CHECK_VALUE -- personally I feel that the whole feature is a bit dubious as you really cannot use it for anything practical but as it is specified in specification so no objections there. The feature itself is optional.

Now as specified in specification it cannot be used to verify uniqueness of secret key or certificate in anyway -- it is a cue that data might be wrapped/unwrapped properly when transferring the key or so. 

Assuming that we have correct implementation I do not see it as a problem to enable the feature for all.

Now about possibility of leaking key material -- I suppose it is OK to have it as option to make it possible to disable in particular implementation if someone wants to make sure there is no extra path of information leakage. One round of AES-ECB for key itself and taking part of the result in theory gives an advice if someone would be brute forcing the key but the problem itself would be still significant -- same goes for SHA1 for other secret keys and only taking part of the checksum. For certificate we do not care as that information is public.

So from my point of view feel free to enable it by default for all platforms.

maroueneboubakri:
@jforissier could you kindly merge this PR. I guess making the option disabled by default is accepted also.

jforissier:
@maroueneboubakri in order to merge I need at least one Acked-by: or Reviewed-by:, and CI to pass.

maroueneboubakri:
> @maroueneboubakri in order to merge I need at least one Acked-by: or Reviewed-by:, and CI to pass.

Sorry thought it had tags applied.

etienne-lms:
Hi all,
Before going for holidays, i saw qemu-check currently fails when `CFG_PKCS11_TA_CHECK_VALUE_ATTRIBUTE=y`. I didn't notify you by then. I've not looked yet, i'll do when i'll come back. This P-R must wait unitl it's fixed.

core/arch/arm/plat-vexpress/conf.mk
etienne-lms: I've created P-R https://github.com/OP-TEE/optee_os/pull/6579 to fix some issues and default enable CHECK_VALUE support.

etienne-lms:
I've created https://github.com/OP-TEE/optee_os/pull/6579 to fix existing implementation issues in pkcs11 that makes some `xtest -t pkcs11` tests to fail.

maroueneboubakri:
@jforissier, @jenswi-linaro FI.

jforissier (commented):

etienne-lms:
@maroueneboubakri, maybe you want to give a review tag to these changes?

maroueneboubakri:
@etienne-lms Sure !

maroueneboubakri:
@jforissier all is good !

etienne-lms:
Indeed. I've applied the review tags.
@jforissier, would you mind restarting the CI tests for https://github.com/OP-TEE/optee_test/pull/719 once this P-R is merged?

jforissier:
> Indeed. I've applied the review tags. @jforissier, would you mind restarting the CI tests for [OP-TEE/optee_test#719](https://github.com/OP-TEE/optee_test/pull/719) once this P-R is merged?

Sure.

PR #6553, opened 2023-12-18
driver: crypto: hisilicon: Add the mailbox operation lock

yuzexiyzx:
1. mailbox operation need lock to ensure atomicity.
2. Optimize the mailbox read and write operations.

Fixes: c7f9abcee87f ("drivers: implement HiSilicon Queue Management (QM) module")

### DIFF
--- a/core/arch/arm/include/arm64.h
+++ b/core/arch/arm/include/arm64.h
@@ -373,6 +373,19 @@ static inline void write_64bit_pair(uint64_t dst, uint64_t hi, uint64_t lo)
 		      "r" (hi), "r" (lo), "r" (dst) : "memory");
 }
 
+static inline void read_64bit_pair(uint64_t src, uint64_t *hi, uint64_t *lo)
+{
+	uint64_t tmp0 = 0;
+	uint64_t tmp1 = 0;
+
+	/* 128bits should be read from hardware at one time */
+	asm volatile ("ldp %0, %1, [%2]\n" : "=&r"(tmp0), "=&r"(tmp1) :
+		      "r"(src) : "memory");
+
+	*lo = tmp0;
+	*hi = tmp1;
+}
+
 /*
  * Templates for register read/write functions based on mrs/msr
  */
--- a/core/drivers/crypto/hisilicon/hisi_qm.c
+++ b/core/drivers/crypto/hisilicon/hisi_qm.c
@@ -24,6 +24,10 @@
 #define QM_MB_OP_SHIFT		14
 #define QM_MB_OP_WR		0
 #define QM_MB_OP_RD		1
+#define QM_MB_STATUS_MASK	GENMASK_32(12, 9)
+#define QM_MB_WAIT_READY_CNT	10
+#define QM_MB_WAIT_MAX_CNT	21000
+#define QM_MB_WAIT_PERIOD	200
 /* XQC_VFT */
 #define QM_VFT_CFG_OP_ENABLE	0x100054
 #define QM_VFT_CFG_OP_WR	0x100058
@@ -143,49 +147,109 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
 	io_write64(qm->io_base + QM_DOORBELL_SQ_CQ_BASE, doorbell);
 }
 
-static enum hisi_drv_status qm_wait_mb_ready(struct hisi_qm *qm)
-{
-	uint32_t val = 0;
-
-	/* Return 0 mailbox ready, HISI_QM_DRVCRYPT_ETMOUT hardware timeout */
-	if (IO_READ32_POLL_TIMEOUT(qm->io_base + QM_MAILBOX_BASE, val,
-				   !(val & QM_MB_BUSY_BIT), POLL_PERIOD,
-				   POLL_TIMEOUT)) {
-		return HISI_QM_DRVCRYPT_ETMOUT;
-	}
-
-	return HISI_QM_DRVCRYPT_NO_ERR;
-}
-
 static void qm_mb_write(struct hisi_qm *qm, struct qm_mailbox *mb)
 {
 	vaddr_t dst = qm->io_base + QM_MAILBOX_BASE;
 
-	write_64bit_pair(dst, mb->x[0], mb->x[1]);
+	write_64bit_pair(dst, mb->x[1], mb->x[0]);
+	dsb_osh();
 }
 
-static enum hisi_drv_status qm_mb(struct hisi_qm *qm, uint8_t cmd,
-				  vaddr_t dma_addr, uint16_t qn, uint8_t op)
+static void qm_mb_read(struct hisi_qm *qm, struct qm_mailbox *mb)
+{
+	vaddr_t mb_base = qm->io_base + QM_MAILBOX_BASE;
+
+	read_64bit_pair(mb_base, mb->x + 1, mb->x);
+	dsb_osh();
+}
+
+static enum hisi_drv_status qm_wait_mb_ready(struct hisi_qm *qm)
 {
 	struct qm_mailbox mb = { };
+	uint32_t timeout = 0;
 
-	mb.w0 = cmd | SHIFT_U32(op, QM_MB_OP_SHIFT) |
-		BIT32(QM_MB_BUSY_SHIFT);
-	mb.queue = qn;
-	reg_pair_from_64(dma_addr, &mb.base_h, &mb.base_l);
-	mb.token = 0;
-
-	if (qm_wait_mb_ready(qm)) {
-		EMSG("QM mailbox is busy");
-		return HISI_QM_DRVCRYPT_EBUSY;
+	timeout = timeout_init_us(QM_MB_WAIT_PERIOD * QM_MB_WAIT_READY_CNT);
+	while (!timeout_elapsed(timeout)) {
+		/* 128 bits should be read from hardware at one time*/
+		qm_mb_read(qm, &mb);
+		if (!(mb.w0 & QM_MB_BUSY_BIT))
+			return HISI_QM_DRVCRYPT_NO_ERR;
 	}
 
-	qm_mb_write(qm, &mb);
+	EMSG("QM mailbox is busy to start!");
 
-	if (qm_wait_mb_ready(qm)) {
-		EMSG("QM mailbox operation timeout");
-		return HISI_QM_DRVCRYPT_EBUSY;
+	return HISI_QM_DRVCRYPT_EBUSY;
+}
+
+static enum hisi_drv_status qm_wait_mb_finish(struct hisi_qm *qm,
+					      struct qm_mailbox *mb)
+{
+	uint32_t timeout = 0;
+
+	timeout = timeout_init_us(QM_MB_WAIT_PERIOD * QM_MB_WAIT_MAX_CNT);
+	while (!timeout_elapsed(timeout)) {
+		qm_mb_read(qm, mb);
+		if (!(mb->w0 & QM_MB_BUSY_BIT)) {
+			if (mb->w0 & QM_MB_STATUS_MASK) {
+				EMSG("QM mailbox operation failed!");
+				return HISI_QM_DRVCRYPT_EIO;
+			} else {
+				return HISI_QM_DRVCRYPT_NO_ERR;
+			}
+		}
 	}
+
+	return HISI_QM_DRVCRYPT_ETMOUT;
+}
+
+static void qm_mb_init(struct qm_mailbox *mb, uint8_t cmd, uint64_t base,
+		       uint16_t qnum, uint8_t op)
+{
+	mb->w0 = cmd | SHIFT_U32(op, QM_MB_OP_SHIFT) |  QM_MB_BUSY_BIT;
+	mb->queue = qnum;
+	reg_pair_from_64(base, &mb->base_h, &mb->base_l);
+	mb->token = 0;
+}
+
+static enum hisi_drv_status qm_mb_nolock(struct hisi_qm *qm,
+					 struct qm_mailbox *mb)
+{
+	if (qm_wait_mb_ready(qm))
+		return HISI_QM_DRVCRYPT_EBUSY;
+
+	qm_mb_write(qm, mb);
+
+	return qm_wait_mb_finish(qm, mb);
+}
+
+static enum hisi_drv_status hisi_qm_mb_write(struct hisi_qm *qm, uint8_t cmd,
+					     uintptr_t dma_addr, uint16_t qnum)
+{
+	enum hisi_drv_status ret = HISI_QM_DRVCRYPT_NO_ERR;
+	struct qm_mailbox mb = { };
+
+	qm_mb_init(&mb, cmd, dma_addr, qnum, QM_MB_OP_WR);
+	mutex_lock(&qm->mailbox_lock);
+	ret = qm_mb_nolock(qm, &mb);
+	mutex_unlock(&qm->mailbox_lock);
+
+	return ret;
+}
+
+static enum hisi_drv_status hisi_qm_mb_read(struct hisi_qm *qm, uint64_t *base,
+					    uint8_t cmd, uint16_t qnum)
+{
+	enum hisi_drv_status ret = HISI_QM_DRVCRYPT_NO_ERR;
+	struct qm_mailbox mb = { };
+
+	qm_mb_init(&mb, cmd, 0, qnum, QM_MB_OP_RD);
+	mutex_lock(&qm->mailbox_lock);
+	ret = qm_mb_nolock(qm, &mb);
+	mutex_unlock(&qm->mailbox_lock);
+	if (ret)
+		return ret;
+
+	reg_pair_from_64(*base, &mb.base_h, &mb.base_l);
 
 	return HISI_QM_DRVCRYPT_NO_ERR;
 }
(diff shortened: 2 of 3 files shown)


jforissier:
> 1. mailbox operation need lock to ensure atomicity.
> 2. Optimize the mailbox read and write operations.

Please split into two commits.

yuzexiyzx:
I think there are some issues with my expression. I re-implemented the mailbox operation function to achieved these two results.I will rewrite the description of the commit.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +147,127 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
 	io_write64(qm->io_base + QM_DOORBELL_SQ_CQ_BASE, doorbell);
 }
 
+static void qm_mb_write(struct hisi_qm *qm, void *src)
+{
+	uintptr_t dst = qm->io_base + QM_MAILBOX_BASE;
+	unsigned long tmp0 = 0, tmp1 = 0;
+
+	/* 128bits should be written to hardware at one time */
+	asm volatile ("ldp %0, %1, %3\n"
```
jenswi-linaro, line 156: The `ldp` operation can be done in C code and the `stp` operation is already available as `write_64bit_pair()`.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +147,127 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+		      "stp %0, %1, %2\n"
+		      "dmb oshst\n"
```
jenswi-linaro, line 158: It's not obvious to me why this particular instruction is needed. Please add a short comment explaining why `dmb` is needed instead of `dsb` and why `oshst`. If it's indeed `dmb oshst` that is needed, please add a `dmb_oshst()` to `arm64.h`.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -24,6 +24,10 @@
 #define QM_MB_OP_SHIFT		14
 #define QM_MB_OP_WR		0
 #define QM_MB_OP_RD		1
+#define QM_MB_STATUS_MASK GENMASK_32(12, 9)
+#define QM_MB_WAIT_READY_CNT 10
+#define QM_MB_WAIT_MAX_CNT 21000
+#define QM_MB_WAIT_PERIOD 200
```
etienne-lms, line 30: I suggest you add some tabulation indent to values for consistency with the macro definitions below and above.

core/drivers/crypto/hisilicon/hisi_qm.c
yuzexiyzx: Sorry, not needed.
use write_64bit_pair in qm_mb_write and in qm_mb_read now.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,113 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+
+	return HISI_QM_DRVCRYPT_EBUSY;
+}
+
+static enum hisi_drv_status qm_wait_mb_finish(struct hisi_qm *qm,
+					      struct qm_mailbox *mb)
+{
+	uint32_t i = 0;
+
+	while (++i) {
```
jenswi-linaro, line 188: `++i` is always going to be true, with the check below. I'd prefer `for (i = 0; ; i++)` or `while (true)` and increasing `i` somewhere else in the loop, preferably not in the `if` below.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,113 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+		qm_mb_read(qm, mb);
+		if (!(mb->w0 & QM_MB_BUSY_BIT))
+			break;
+
+		if (i > QM_MB_WAIT_MAX_CNT) {
```
jenswi-linaro, line 193: Would it be possible to use `timeout_init_us()` and `timeout_elapsed()` for this instead? Or is it not safe to assume `CFG_CORE_HAS_GENERIC_TIMER=y`?

core/drivers/crypto/hisilicon/hisi_qm.c
yuzexiyzx: solved

core/drivers/crypto/hisilicon/hisi_qm.c
yuzexiyzx: I think cnt maybe clearer than time, since we delay QM_MB_WAIT_PERIOD every time.

jenswi-linaro: You poll for a timeout instead of `mb->w0 & QM_MB_BUSY_BIT`, so you'll likely wait longer than necessary. If the thread is suspended by a non-secure interrupt that time will not be accounted for in the timeout. So in reality you can't say how long it will take to timeout.

yuzexiyzx: solved

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,108 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+				return HISI_QM_DRVCRYPT_NO_ERR;
+			}
+		}
 	}
 
-	qm_mb_write(qm, &mb);
+	return HISI_QM_DRVCRYPT_ETMOUT;
+}
+
+static void qm_mb_pre_init(struct qm_mailbox *mb, uint8_t cmd, uint64_t base,
```
jenswi-linaro, line 204: qm_mb_init(), this is the initialization stage.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,108 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
 }
 
-static void qm_mb_write(struct hisi_qm *qm, struct qm_mailbox *mb)
+static void qm_mb_read(struct hisi_qm *qm, struct qm_mailbox *mb)
 {
-	vaddr_t dst = qm->io_base + QM_MAILBOX_BASE;
+	vaddr_t mb_base = qm->io_base + QM_MAILBOX_BASE;
 
-	write_64bit_pair(dst, mb->x[0], mb->x[1]);
+	write_64bit_pair((vaddr_t)mb, mb_base, mb_base + U64_OFFSET);
```
jenswi-linaro, line 162: Aren't you supposed to read 16 bytes from `qm->io_base + QM_MAILBOX_BASE`?

Is this tested?

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,108 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
 	}
 
-	qm_mb_write(qm, &mb);
+	return HISI_QM_DRVCRYPT_ETMOUT;
+}
+
+static void qm_mb_pre_init(struct qm_mailbox *mb, uint8_t cmd, uint64_t base,
+			   uint16_t qnum, uint8_t op)
+{
+	mb->w0 = (cmd | (op ? 0x1 << QM_MB_OP_SHIFT : 0) |
```
jenswi-linaro, line 207: Please avoid the ternary operation. Wouldn't `SHIFT_U32(op, QM_MB_OP_SHIFT)` be closer to what you're trying to achieve anyway?
The outer `()` aren't needed.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,108 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+	int32_t ret;
+
+	qm_mb_pre_init(&mb, cmd, 0, qnum, QM_MB_OP_RD);
+	mutex_lock(&qm->mailbox_lock);
+	ret = qm_mb_nolock(qm, &mb);
+	mutex_unlock(&qm->mailbox_lock);
+	if (ret)
+		return ret;
+
+	*base = ADDR_U64(mb.base_h, mb.base_l);
```
jenswi-linaro, line 252: Please use `reg_pair_to_64()` instead

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +148,108 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+
+	qm_mb_write(qm, mb);
+
+	return qm_wait_mb_finish(qm, mb);
+}
+
+static enum hisi_drv_status hisi_qm_mb_write(struct hisi_qm *qm, uint8_t cmd,
+					     uintptr_t dma_addr, uint16_t qnum)
+{
+	struct qm_mailbox mb;
```
jenswi-linaro, line 228: Please initialize all local variables, this is required by the coding guidelines.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -780,6 +845,7 @@ enum hisi_drv_status hisi_qp_recv_sync(struct hisi_qp *qp, void *msg)
 	while (true) {
 		ret = hisi_qp_recv(qp, msg);
 		if (ret == HISI_QM_DRVCRYPT_NO_ERR) {
+			udelay(QM_SINGLE_WAIT_TIME);
```
jenswi-linaro, line 848: How about using `timeout_init_us()` and `timeout_elapsed()` as in `qm_wait_mb_ready()`?

core/drivers/crypto/hisilicon/hisi_qm.c
yuzexiyzx: image
The original code was implemented this way. Referring to qm_mb_read, I have re implemented qm_mb_write. Due to significant differences between code on our library and code in community, it is difficulet to test the modified code, but the original code has been tested.I am sorry about that.I will open up the remaining patches as soon as possible and then turn to the community's code.

yuzexiyzx:
@jenswi-linaro All comments have been solved.

core/drivers/crypto/hisilicon/hisi_qm.c
jenswi-linaro: This is confusing. Why are you using `write_64bit_pair()` to do an atomic 128-bit write into the mailbox which isn't shared with the device? It seems a plain `memcpy()` or such would do the job just as well.

jenswi-linaro: The last two arguments to `write_64bit_pair()` are addresses, while the actual data is expected. I don't think this function works as intended.

yuzexiyzx: The hardware requires that 128bits should be read from hardware at one time.It is implemented in kernel in that way using stp as picture above.My modification was wrong.Solved now.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +147,108 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+	vaddr_t dst = qm->io_base + QM_MAILBOX_BASE;
+
+	write_64bit_pair(dst, mb->x[0], mb->x[1]);
+}
+
+static void qm_mb_read(struct hisi_qm *qm, struct qm_mailbox *mb)
+{
+	vaddr_t mb_base = qm->io_base + QM_MAILBOX_BASE;
+
+	write_64bit_pair((vaddr_t)mb, *(uint64_t *)mb_base,
```
jenswi-linaro, line 161: > The hardware requires that 128bits should be read from hardware at one time.It is implemented in kernel in that way using stp as picture above.My modification was wrong.Solved now.

This doesn't make sense. If you need an atomic 128-bit read, then you should do an `ldp` or `read_64bit_pair()` not a `stp` or `write_64bit_pair()`. The compiler might implement the two memory dereferences `*(uint64_t *)mb_base` and `*((uint64_t *)mb_base + 1)` with an `ldp` instruction, but that's not guaranteed. `mb` only resides on the stack when this function is called from `qm_wait_mb_ready()` and `qm_wait_mb_finish()`, so that struct should be accessible in a normal way.

By the way, I don't think it makes sense to submit untested code upstream. Not tested == not working, we've seen that before.

yuzexiyzx: The code has been tested.This function actually writes from mb_base to mb.When tested, I found a problem in qm_mb_write and solved it.

jenswi-linaro: Good that you tested it and could fix problems because of that, but you still haven't addressed my other comment:

> This doesn't make sense. If you need an atomic 128-bit read, then you should do an ldp or read_64bit_pair() not a stp or write_64bit_pair(). The compiler might implement the two memory dereferences *(uint64_t *)mb_base and *((uint64_t *)mb_base + 1) with an ldp instruction, but that's not guaranteed. mb only resides on the stack when this function is called from qm_wait_mb_ready() and qm_wait_mb_finish(), so that struct should be accessible in a normal way.

yuzexiyzx: mb_base and mb are both memory.I need use stp to write info to memory.I use  *(uint64_t *)mb_base and *((uint64_t *)mb_base + 1 to implement ldp，but stp is needed in final.

yuzexiyzx: The original code is implemented in this way. stp is needed. I only change the way to implement ldp.

yuzexiyzx: image

jenswi-linaro: I can see how the `stp` could be needed when writing to the shared mailbox, but when reading from the shared mailbox I don't see why you need an `stp` instruction when saving the read value on the stack. I'd expect `ldp` to be used when reading from the shared mailbox if you intend to use 128-bit atomic memory accesses to the shared mailbox.

yuzexiyzx: solved

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +147,114 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+	dsb();
+}
+
+static void qm_mb_read(struct hisi_qm *qm, struct qm_mailbox *mb)
+{
+	vaddr_t mb_base = qm->io_base + QM_MAILBOX_BASE;
+	uint64_t tmp0 = 0;
+	uint64_t tmp1 = 0;
+
+	asm volatile ("ldp %0, %1, [%2]\n" : "=&r"(tmp0), "=&r"(tmp1) :
```
jenswi-linaro, line 164: Please add a `static inline void read_64bit_pair(uint64_t src, uint64_t *hi, uint64_t *lo)` to arm64.h in a separate commit and use it here instead of using inline assembly here.

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +147,114 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
...
+		      "r"(mb_base) : "memory");
+	dsb();
```
jenswi-linaro, line 166: Why is this done after reading the shared mailbox? Is this `dsb()` needed at all?

core/drivers/crypto/hisilicon/hisi_qm.c
```diff
@@ -143,49 +147,114 @@ static void qm_db(struct hisi_qm *qm, uint16_t qn, uint8_t cmd, uint16_t index,
 	io_write64(qm->io_base + QM_DOORBELL_SQ_CQ_BASE, doorbell);
 }
 
+static void qm_mb_write(struct hisi_qm *qm, struct qm_mailbox *mb)
+{
+	vaddr_t dst = qm->io_base + QM_MAILBOX_BASE;
+
+	write_64bit_pair(dst, mb->x[1], mb->x[0]);
+	dsb();
```
jenswi-linaro, line 155: I don't suppose it matters much, but wouldn't `dsb_osh()` be a better match? It's fine to keep it as it is.

core/drivers/crypto/hisilicon/hisi_qm.c
yuzexiyzx: dsb() is needed to make sure that all previous data read and write operations are completed before the instruction is executed.dsb() has been changed to dsb_osh().

jenswi-linaro: Got it, thanks.

core/arch/arm/include/arm64.h
```diff
@@ -373,6 +373,19 @@ static inline void write_64bit_pair(uint64_t dst, uint64_t hi, uint64_t lo)
 		      "r" (hi), "r" (lo), "r" (dst) : "memory");
 }
 
+static inline void read_64bit_pair(uint64_t dst, uint64_t *hi, uint64_t *lo)
```
jenswi-linaro, line 376: Please add a separate commit for the change in this file.

`s/dst/src/`

core/arch/arm/include/arm64.h
```diff
@@ -373,6 +373,19 @@ static inline void write_64bit_pair(uint64_t dst, uint64_t hi, uint64_t lo)
...
+{
+	uint64_t tmp0 = 0;
+	uint64_t tmp1 = 0;
+
+	/* 128bits should be written to hardware at one time */
```
jenswi-linaro, line 381: This comment is somewhat confusing since we're reading.

jenswi-linaro:

yuzexiyzx:
@jenswi-linaro Thanks.
@jforissier All comments have been solved.

yuzexiyzx:
Thanks!

PR #6554, opened 2023-12-18
Add missing clk.h include in clk_dt.h

GseoC:
This header file does not include clk.h. Just declare `struct clk;` so that the dependency is handled.


GseoC:
Why did I not just include clk.h? I don't know...

I'll update the P-R

GseoC:
Renamed the P-R and update it to include the clk.h header file

jforissier (commented):

jenswi-linaro:

GseoC:
Updated with both tags

PR #6555, opened 2023-12-18
dts: add stm32mp15*-scmi.dts files for when RCC is secure [my proposal]

jforissier:
Alternative implementation of #6541.

### DIFF
--- a/core/arch/arm/dts/stm32mp157a-dk1-scmi.dts
+++ b/core/arch/arm/dts/stm32mp157a-dk1-scmi.dts
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: (GPL-2.0+ OR BSD-3-Clause)
+/*
+ * Copyright (C) STMicroelectronics 2019, 2023
+ */
+
+/dts-v1/;
+
+#include "stm32mp157a-dk1.dts"
+
+/ {
+	model = "STMicroelectronics STM32MP157A-DK1 SCMI Discovery Board";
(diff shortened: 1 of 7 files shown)


etienne-lms (commented):
This change mandates one to use a `PLATFORM_FLAVOR` config switch but one can configure its stm32mp1 platform providing `CFG_STM32MP{13|15}=y` and `CFG_EMBED_DTB_SOURCE_FILE=...`, possibly a custom DTS file.

This is why in current implementation, PLATFORM_FLAVOR is rather a alias for a target configuration including the DTS file.

jforissier:
> This change mandates one to use a `PLATFORM_FLAVOR` config switch but one can configure its stm32mp1 platform providing `CFG_STM32MP{13|15}=y` and `CFG_EMBED_DTB_SOURCE_FILE=...`, possibly a custom DTS file.
> 
> This is why in current implementation, PLATFORM_FLAVOR is rather a alias for a target configuration including the DTS file.

I don't get it, you don't want to be able to provide a custom DTS file?

etienne-lms:
> I don't get it, you don't want to be able to provide a custom DTS file?

Yes I do. When so, there is no pre-selected flavor. I mean I would like to set `CFG_EMBED_DTB_SOURCE_FILE=xxx.dts` without any flavor, or alternatively use `PLATFORM_FLAVOR` to pre-select a DTS and a config.

jforissier:
OK. Configuring based on the DT file name seems a bit fragile to me (I believe should be able to override the DT with any temporary name for convenience [and even give a full path but that's another story]), but that works.

PR #6561, opened 2023-12-21
core: fix #if CFG_xxx into #ifdef CFG_xxx

etienne-lms:
* "core: arm: kernel_spmc: correct CFG_SECURE_PARTITION test"
Replace occurrences of `#if CFG_SECURE_PARTITION` to a `#ifdef` test as the
configuration switch is either defined or not defined.
* "plat-stm32mp2: fix test on CFG_STM32_UART"
Fix test directive on `CFG_STM32_UART` that is either defined or not defined.

### DIFF
--- a/core/arch/arm/kernel/thread_spmc_a64.S
+++ b/core/arch/arm/kernel/thread_spmc_a64.S
@@ -14,7 +14,7 @@
 #include <kernel/thread.h>
 #include <optee_ffa.h>
 
-#if CFG_SECURE_PARTITION
+#ifdef CFG_SECURE_PARTITION
 LOCAL_FUNC thread_ffa_interrupt , :
 	mov_imm	x0, FFA_INTERRUPT		/* FID */
 	/* X1: Endpoint/vCPU IDs is set by caller */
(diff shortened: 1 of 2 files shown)


jforissier:
`#ifdef` and `#if defined(...)` are indeed the proper way to test for "boolean" `CFG_` in C code, so:

Minor: the commit descriptions are not very helpful because "the configuration switch is either defined or not defined" is always true :wink:

etienne-lms:
> Minor: the commit descriptions are not very helpful because "the configuration switch is either defined or not defined" is always true 😉

I'll rephrase to "Boolean CFG_ configuration switches do not have a value."

jforissier:
> > Minor: the commit descriptions are not very helpful because "the configuration switch is either defined or not defined" is always true 😉
> 
> I'll rephrase to "Boolean CFG_ configuration switches do not have a value."

Well, technically they do: the C macros are normally set to `1` when the Makefile variable is `y` (see the generated `$(out-dir)/conf.mk`). But the value is not meaningful, we prefer to rely on whether they are defined or not. So "do not have a meaningful value" would be more correct.

jenswi-linaro:

etienne-lms:
> (...) So "do not have a meaningful value" would be more correct

Thanks, i've used your phrasing.
Review tags applied.

PR #6565, opened 2023-12-26
plat-k3: sa2ul_rng: check if rng is enabled before to do a read

thom24:
Check if rng is enabled in hw_get_random_bytes(), if not the initialization sequence is run.
After a suspend to ram, the rng is in reset state, and it has to be re-initialized.

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul_rng.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul_rng.c
@@ -64,9 +64,35 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, RNG_BASE, RNG_REG_SIZE);
 static unsigned int rng_lock = SPINLOCK_UNLOCK;
 static vaddr_t rng;
 
+static bool sa2ul_rng_is_enabled(void)
+{
+	return io_read32(rng + RNG_CONTROL) & ENABLE_TRNG;
+}
+
+static void sa2ul_rng_init_seq(void)
+{
+	uint32_t val = 0;
+
+	/* Ensure initial latency */
+	val |= RNG_CONFIG_MIN_REFIL_CYCLES << RNG_CONFIG_MIN_REFIL_CYCLES_SHIFT;
+	val |= RNG_CONFIG_MAX_REFIL_CYCLES << RNG_CONFIG_MAX_REFIL_CYCLES_SHIFT;
+	io_write32(rng + RNG_CONFIG, val);
+
+	/* Configure the desired FROs */
+	io_write32(rng + RNG_FRODETUNE, 0x0);
+
+	/* Enable all FROs */
+	io_write32(rng + RNG_FROENABLE, 0xffffff);
+
+	io_write32(rng + RNG_CONTROL, ENABLE_TRNG);
+}
+
 static void sa2ul_rng_read128(uint32_t *word0, uint32_t *word1,
 			      uint32_t *word2, uint32_t *word3)
 {
+	if (!sa2ul_rng_is_enabled())
+		sa2ul_rng_init_seq();
+
 	/* Is the result ready (available)? */
 	while (!(io_read32(rng + RNG_STATUS) & RNG_READY)) {
 		/* Is the shutdown threshold reached? */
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -141,6 +144,13 @@ TEE_Result sa2ul_rng_init(void)
 	io_write32(rng + RNG_FROENABLE, 0xffffff);
 
 	io_write32(rng + RNG_CONTROL, ENABLE_TRNG);
+}
+
+TEE_Result sa2ul_rng_init(void)
+{
+	rng = (vaddr_t)phys_to_virt(RNG_BASE, MEM_AREA_IO_SEC, RNG_REG_SIZE);
+
```
uditkumarti, line 152: Better to have check for RnG init here as well, instead of forced init 
Case of some fw already did initialization of RNG 
 @manorit2001 ?

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
uditkumarti, line 1: in commit message 
"After a suspend to ram, the rng is in reset state, and it has to be
re-initialized." 

to 

After a suspend to ram, the rng may be in reset state, and it has to be
re-initialized if in reset state

uditkumarti (commented):
commit message

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
thom24: ack

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
manorit2001: Wouldn't hurt ig, better to keep that check only incase TIFS does some weird things

thom24: Done

glneo:
After suspend is OP-TEE not informed of the resume in some way? That would be the better spot to re-initialize this driver instead of checking every time we need a random byte.

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -125,22 +151,11 @@ TEE_Result hw_get_random_bytes(void *buf, size_t len)
...
-	io_write32(rng + RNG_CONFIG, val);
-
-	/* Configure the desired FROs */
-	io_write32(rng + RNG_FRODETUNE, 0x0);
-
-	/* Enable all FROs */
-	io_write32(rng + RNG_FROENABLE, 0xffffff);
-
-	io_write32(rng + RNG_CONTROL, ENABLE_TRNG);
+	/* Check if rng has already been initialized by a firmware */
```
glneo, line 156: It might not be firmware, do we really trust the TRNG is setup correctly by just checking that it is enabled?

Previously we setup the firewall to protect our access to the TRNG and then unconditionally run setup on it. Now if someone (could be anyone, before we put up the firewall anyone in the system can mess with it (like R5 SPL)) set it up incorrectly (to return all 0s for instance) then we just trust it now..

glneo (commented):
BTW, I'm planning on moving our TRNG driver up into TF-A then exposing it using the SMCCC interface. So we would move to using the smccc_trng.c driver here (just didn't want you spending too much time on this as this driver might be going away).

thom24:
OP-TEE is not informed of the resume.
The correct way should be to implement suspend/resume at platform level (example: https://elixir.bootlin.com/op-tee/latest/source/core/arch/arm/plat-sam/pm/psci.c#L15). 
At this time op-tee should be informed of the suspend/resume and run suspend/resume callbacks of drivers.
But it's lots of work only for rng. This patch is just an easier solution.

I made some tests using rngtest, performances are quite similar.

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
thom24: You're right.
It's more secure if we setup it unconditionally once the firewall is up.

thom24: I updated the PR.
Now the TRNG is setup unconditionally during the driver init.

core/arch/arm/plat-k3/drivers/sa2ul_rng.c
```diff
@@ -106,6 +129,9 @@ TEE_Result hw_get_random_bytes(void *buf, size_t len)
 	uint8_t *buffer = buf;
 	size_t buffer_pos = 0;
 
+	if (!sa2ul_rng_is_enabled())
+		sa2ul_rng_init_seq();
```
glneo, line 133: hw_get_random_bytes() may be called while there is already enough random bits in our pool, we only need to check if the TRNG is enabled when we are about to actually do a read. This check should go in sa2ul_rng_read128().

thom24: Oh yes.
No need to check if the TRNG is enabled if the pool is not empty.
If the check is moved in sa2ul_rng_read128, it avoids lots of useless accesses.

glneo:
`Reviewed-by: Andrew Davis <afd@ti.com>`

etienne-lms (commented):

PR #6576, opened 2024-01-04
RISC-V: Small fixing

gagachang:
Small fixing from previous PR review and rebase from latest master branch.

### DIFF
--- a/core/arch/riscv/include/riscv.h
+++ b/core/arch/riscv/include/riscv.h
@@ -117,7 +117,7 @@ static inline __noprof void mb(void)
 
 static inline __noprof unsigned long read_tp(void)
 {
-	unsigned long tp;
+	unsigned long tp = 0;
 
 	asm volatile("mv %0, tp" : "=&r"(tp));
 	return tp;
--- a/core/arch/riscv/kernel/thread_arch.c
+++ b/core/arch/riscv/kernel/thread_arch.c
@@ -14,6 +14,7 @@
 #include <keep.h>
 #include <kernel/asan.h>
 #include <kernel/boot.h>
+#include <kernel/interrupt.h>
 #include <kernel/linker.h>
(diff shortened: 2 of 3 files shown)


jenswi-linaro:

gagachang:
Tag applied!

jforissier:
Wouldn't it be time to add a RISC-V build to CI? Let me know which cross-compiler you need and I will add it to the container image that runs the build (https://github.com/jforissier/docker_optee_os_ci/blob/master/Dockerfile).

gagachang:
> Wouldn't it be time to add a RISC-V build to CI? Let me know which cross-compiler you need and I will add it to the container image that runs the build (https://github.com/jforissier/docker_optee_os_ci/blob/master/Dockerfile).

Hi @jforissier 
Sure, that's a good idea!
But until now RISC-V still lacks the macro for speculation load mitigation.
(https://github.com/OP-TEE/optee_os/blob/master/lib/libutils/ext/include/speculation_barrier.h#L505)
Currently we skip it for easily porting.
Now I will check it, and also find the proper RISC-V toolchain which we can add into the Dockerfile.

jforissier:
@gagachang thanks. It doesn't matter if the code is not really usable, I would like to have at least a compile check in the "make (multi-platform)" job. Of course adding runtime tests in QEMU or whatever simulator would be good as a next step but it can wait.

gagachang:
Hi @maroueneboubakri 

Do you have any idea about RISC-V's speculative load mitigation?
It seems RISC-V toolchain hasn't had `__builtin_load_no_speculate()` intrinsic yet.

I've found some materials:
- https://riscv.or.jp/wp-content/uploads/rvdv_07_RISC-V-Day-Vietnam2022-TienLA-.pdf
- https://llvm.org/docs/SpeculativeLoadHardening.html

PR #6582, opened 2024-01-09
Coverity fixes

clementfaure:
Hello!
Here are some Coverity fixes I have for the whole optee-os codebase. Most of them are Low/Medium issues, nothing really critical. Let me know what you think.

### DIFF
--- a/core/arch/arm/tee/entry_fast.c
+++ b/core/arch/arm/tee/entry_fast.c
@@ -30,8 +30,8 @@ static void tee_entry_get_shm_config(struct thread_smc_args *args)
 
 static void tee_entry_fastcall_l2cc_mutex(struct thread_smc_args *args)
 {
-	TEE_Result ret;
 #ifdef ARM32
+	TEE_Result ret = TEE_ERROR_NOT_SUPPORTED;
 	paddr_t pa = 0;
 
 	switch (args->a1) {
@@ -53,15 +53,16 @@ static void tee_entry_fastcall_l2cc_mutex(struct thread_smc_args *args)
 		args->a0 = OPTEE_SMC_RETURN_EBADCMD;
 		return;
 	}
-#else
-	ret = TEE_ERROR_NOT_SUPPORTED;
-#endif
+
 	if (ret == TEE_ERROR_NOT_SUPPORTED)
 		args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
 	else if (ret)
 		args->a0 = OPTEE_SMC_RETURN_EBADADDR;
 	else
 		args->a0 = OPTEE_SMC_RETURN_OK;
+#else
+	args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
+#endif
 }
 
 static void tee_entry_exchange_capabilities(struct thread_smc_args *args)
--- a/core/kernel/dt_driver.c
+++ b/core/kernel/dt_driver.c
@@ -266,8 +266,10 @@ static TEE_Result device_from_provider_prop(struct dt_driver_provider *prv,
 	pargs->fdt = fdt;
 	pargs->phandle_node = phandle_node;
 	pargs->args_count = prv->provider_cells;
-	for (n = 0; n < prv->provider_cells; n++)
+	for (n = 0; n < prv->provider_cells; n++) {
+		assert(prop);
 		pargs->args[n] = fdt32_to_cpu(prop[n]);
+	}
 
 	res = prv->get_of_device(pargs, prv->priv_data, device_ref);
 
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -968,6 +968,8 @@ static void collect_device_mem_ranges(struct tee_mmap_region *memory_map,
 	int subnode = 0;
 	int node = 0;
 
+	assert(fdt);
+
 	node = fdt_node_offset_by_compatible(fdt, 0, compatible);
 	if (node < 0)
 		return;
--- a/ldelf/ta_elf_rel.c
+++ b/ldelf/ta_elf_rel.c
@@ -282,7 +282,7 @@ static void resolve_sym(const char *name, vaddr_t *val, struct ta_elf **mod,
 	if (res) {
 		if (err_if_not_found)
 			err(res, "Symbol %s not found", name);
-		else
+		else if (val)
 			*val = 0;
 	}
 }
--- a/ta/pkcs11/src/pkcs11_attributes.c
+++ b/ta/pkcs11/src/pkcs11_attributes.c
@@ -2539,6 +2539,11 @@ static enum pkcs11_rc set_private_key_data_rsa(struct obj_attrs **head,
 	}
 
 	rsa = mbedtls_pk_rsa(pk);
+	if (!rsa) {
+		rc = PKCS11_CKR_GENERAL_ERROR;
+		goto out;
+	}
+
 	mbedtls_rc = mbedtls_rsa_export(rsa, &n, &p, &q, &d, &e);
 	if (mbedtls_rc) {
 		rc = PKCS11_CKR_ARGUMENTS_BAD;
@@ -2730,6 +2735,11 @@ encode_rsa_private_key_der(struct obj_attrs *head, void **data, uint32_t *sz)
 		goto out;
 
 	rsa = mbedtls_pk_rsa(pk);
+	if (!rsa) {
+		rc = PKCS11_CKR_GENERAL_ERROR;
+		goto out;
+	}
+
 	mbedtls_rc = mbedtls_rsa_import_raw(rsa, n, n_len, p, p_len,
 					    q, q_len, d, d_len, e, e_len);
 	if (mbedtls_rc) {
--- a/core/include/kernel/msg_param.h
+++ b/core/include/kernel/msg_param.h
@@ -30,7 +30,6 @@
 #define __KERNEL_MSG_PARAM_H
 
 #include <compiler.h>
-#include <kernel/msg_param.h>
 #include <mm/mobj.h>
 #include <optee_msg.h>
 #include <stdio.h>
--- a/core/kernel/dt.c
+++ b/core/kernel/dt.c
@@ -495,7 +495,10 @@ static int add_dt_overlay_fragment(struct dt_descriptor *dt, int ioffs)
 	int offs = 0;
 	int ret = 0;
 
-	snprintf(frag, sizeof(frag), "fragment@%d", dt->frag_id);
+	ret = snprintf(frag, sizeof(frag), "fragment@%d", dt->frag_id);
+	if (ret < 0 || (size_t)ret >= sizeof(frag))
+		return -1;
+
 	offs = fdt_add_subnode(dt->blob, ioffs, frag);
 	if (offs < 0)
 		return offs;
--- a/core/tee/fs_dirfile.c
+++ b/core/tee/fs_dirfile.c
@@ -369,7 +369,7 @@ TEE_Result tee_fs_dirfile_get_next(struct tee_fs_dirfile_dirh *dirh,
 {
 	TEE_Result res;
 	int i = *idx + 1;
-	struct dirfile_entry dent;
+	struct dirfile_entry dent = { };
 
 	if (i < 0)
 		i = 0;
--- a/core/tee/tadb.c
+++ b/core/tee/tadb.c
@@ -335,7 +335,7 @@ static TEE_Result populate_files(struct tee_tadb_dir *db)
 	 * unexpected side effects.
 	 */
 	for (idx = 0;; idx++) {
-		struct tadb_entry entry;
+		struct tadb_entry entry = { };
 
 		res = read_ent(db, idx, &entry);
 		if (res) {
--- a/core/tee/tee_svc_cryp.c
+++ b/core/tee/tee_svc_cryp.c
@@ -2133,7 +2133,10 @@ static TEE_Result tee_svc_obj_generate_key_rsa(
 		if (res)
 			return res;
 	} else {
-		crypto_bignum_bin2bn((const uint8_t *)&e, sizeof(e), key->e);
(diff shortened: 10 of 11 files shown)


core/kernel/dt_driver.c
```diff
@@ -256,6 +256,9 @@ static TEE_Result device_from_provider_prop(struct dt_driver_provider *prv,
 	struct dt_pargs *pargs = NULL;
 	unsigned int n = 0;
 
+	if (!prop)
+		return TEE_ERROR_BAD_PARAMETERS;
```
jforissier, line 260: This can't be correct since `device_from_provider_prop()` is called with `prop == NULL` by `dt_driver_device_from_parent()`. So `NULL` has to be a valid case I suppose.

core/mm/core_mmu.c
```diff
@@ -968,6 +968,9 @@ static void collect_device_mem_ranges(struct tee_mmap_region *memory_map,
 	int subnode = 0;
 	int node = 0;
 
+	if (!fdt)
```
jforissier, line 971: I don't think this is supposed to happen. Wouldn't `assert(fdt);` make more sense?
 @LeiSen62

jforissier (commented):
Hi @clementfaure, thanks for the patches. 
- Please see two comments below.
- For "core: arm: fix dead code when ARM32=n":
  Replace "ARM32=n" with "ARM32 is not defined". Then:
- For " core: kernel: dt: check return values from snprintf()"
  "core: check if binary to bignum conversion succeed"
  "core: check if string to uuid conversion succeed"
  "ta: pkcs11: check returned value of mbedtls_pk_rsa()"
  "core: tee: initialize dirfile|tadb_entry objects"
  "ldelf: check val for NULL dereference"
- For "core: remove unnecessaray header files"
  "ta: remove unnecessary header files"
  Please change "unnecessaray header files" to "unnecessary includes", then:

core/arch/arm/tee/entry_fast.c
```diff
@@ -30,8 +28,8 @@ static void tee_entry_get_shm_config(struct thread_smc_args *args)
 
 static void tee_entry_fastcall_l2cc_mutex(struct thread_smc_args *args)
 {
-	TEE_Result ret;
 #ifdef ARM32
+	TEE_Result ret;
```
vincent-mailhol, line 32: The variable `ret` shall be initialized (c.f. [optee coding standards](https://optee.readthedocs.io/en/latest/general/coding_standards.html#coding-standards)). So what about:

```c
static void tee_entry_fastcall_l2cc_mutex(struct thread_smc_args *args)
{
	TEE_Result ret = TEE_ERROR_NOT_SUPPORTED;
#ifdef ARM32
	paddr_t pa = 0;

	switch (args->a1) {
	case OPTEE_SMC_L2CC_MUTEX_GET_ADDR:
		ret = tee_get_l2cc_mutex(&pa);
		reg_pair_from_64(pa, &args->a2, &args->a3);
		break;
	case OPTEE_SMC_L2CC_MUTEX_SET_ADDR:
		pa = reg_pair_to_64(args->a2, args->a3);
		ret = tee_set_l2cc_mutex(&pa);
		break;
	case OPTEE_SMC_L2CC_MUTEX_ENABLE:
		ret = tee_enable_l2cc_mutex();
		break;
	case OPTEE_SMC_L2CC_MUTEX_DISABLE:
		ret = tee_disable_l2cc_mutex();
		break;
	default:
		args->a0 = OPTEE_SMC_RETURN_EBADCMD;
		return;
	}
#endif
	if (ret == TEE_ERROR_NOT_SUPPORTED)
	else if (ret)
		args->a0 = OPTEE_SMC_RETURN_EBADADDR;
	else
		args->a0 = OPTEE_SMC_RETURN_OK;
}
```

vincent-mailhol (commented):
A single comment on top of what @jforissier wrote.

core/kernel/dt_driver.c
clementfaure: Coverity is unhappy about the potential use of `prop` being in NULL in the  `fdt32_to_cpu()` function below.

etienne-lms: `assert(!prv->provider_cells || prop)`?

core/arch/arm/tee/entry_fast.c
clementfaure: Coverity will still whine about the piece of code being dead. If ARM32=n, we got
```
        TEE_Result ret = TEE_ERROR_NOT_SUPPORTED;

	if (ret == TEE_ERROR_NOT_SUPPORTED)
	else if (ret) //DEAD
		args->a0 = OPTEE_SMC_RETURN_EBADADDR; //DEAD
	else //DEAD
		args->a0 = OPTEE_SMC_RETURN_OK; //DEAD
```

core/mm/core_mmu.c
clementfaure: The output of `get_manifest_dt()` is checked in `fip_sp_init_all()`

core/kernel/tee_ta_manager.c
```diff
@@ -21,15 +21,11 @@
 #include <mm/core_mmu.h>
 #include <mm/mobj.h>
 #include <mm/vm.h>
-#include <pta_stats.h>
```
etienne-lms, line 24: _pta_stats.h_ is needed for `struct pta_stats_ta`

core/kernel/dt_driver.c
jforissier: OK but returning early with an error status will probably not work, if it would then it means we could replace `return device_from_provider_prop(prv, fdt, nodeoffset, NULL, device_ref);` by `return TEE_ERROR_BAD_PARAMETERS;`  in `dt_driver_device_from_parent()`.
Perhaps we assume `prop` is non-NULL as soon as `prv->provider_cells` is non-zero? Then how about:
```C
for (n = 0; n < prv->provider_cells; n++) {
		assert(prop);  /* Or "if (prop)"? */
		pargs->args[n] = fdt32_to_cpu(prop[n]);
}
```

clementfaure: I'm fine with the latest.

core/kernel/tee_ta_manager.c
clementfaure: Right ... So it should be 
```
#ifdef CFG_TA_STATS
#include <pta_stats.h>
#endif
````
no ?

Comment (jenswi-linaro) [reply core/kernel/tee_ta_manager.c:24]: Sometimes it might be warranted, but usually it would just make the code harder to read and maintain. We probably need to accept that Coverity sometimes is counterproductive.

File: core/arch/arm/tee/entry_fast.c (hunk 3 above)
Comment (jenswi-linaro) [reply core/arch/arm/tee/entry_fast.c:32]: Yet another situation where Coverity has a different objective.
We don't mind "dead" code, as long as it's used in at least one configuration. This is in fact something we encourage since it gives better compile test coverage.

**Comment by jenswi-linaro** [comment]
Coverity fixes like these are welcome, but it would help the review process if it was done in smaller pieces. Almost all the patches are unrelated to each other.

File: core/kernel/tee_ta_manager.c (hunk 4 above)
Comment (etienne-lms) [reply core/kernel/tee_ta_manager.c:24]: I don't like adding #ifdef for header files inclusion.
Coverity can be configured to not warn on such, no?

Comment (clementfaure) [reply core/kernel/tee_ta_manager.c:24]: I can extend the Coverity code coverage with additional compilation parameters and artifacts. If I add a compilation run with `CFG_TA_STATS=y`, I will get rid of it.

**Comment by clementfaure** [comment]
I will re-shape the PR and remove e7caf4b803098edbb809980f4d088135cd3d5ff6 and 00de76eced8e65e000a79f9e94e04135ab631c44.
I will increase the Coverity code coverage with additionnal compilation runs (I will take the CI builds as a reference) and come back to you with a dedicated PR to ease the review.

File: core/arch/arm/tee/entry_fast.c (hunk 3 above)
Comment (clementfaure) [reply core/arch/arm/tee/entry_fast.c:32]: Well, I do understand Coverity scanning is not in your process and might conflict with some of your coding standards. However, on this particular case, we made sure `paddr_t pa` is within `#ifdef ARM32 ... #endif` why not doing the same thing with `TEE_Result ret` ?

File: ta/pkcs11/src/pkcs11_attributes.c [hunk 5]
```diff
@@ -2539,6 +2539,9 @@ static enum pkcs11_rc set_private_key_data_rsa(struct obj_attrs **head,
 	}
 
 	rsa = mbedtls_pk_rsa(pk);
+	if (!rsa)
```
etienne-lms, line 2542: `rc = PKCS11_CKR_GENERAL_ERROR;`

ta/pkcs11/src/pkcs11_attributes.c
```diff
@@ -2730,6 +2733,9 @@ encode_rsa_private_key_der(struct obj_attrs *head, void **data, uint32_t *sz)
 		goto out;
 
 	rsa = mbedtls_pk_rsa(pk);
+	if (!rsa)
```
etienne-lms, line 2736: `rc = PKCS11_CKR_GENERAL_ERROR;`

etienne-lms (commented):
"core: kernel: dt: check return values from snprintf()"
"core: check if binary to bignum conversion succeed"
"core: check if string to uuid conversion succeed"
"core: tee: initialize dirfile|tadb_entry objects"
"ldelf: check val for NULL dereference"
"core: msg_param: remove recursion in included headers"

ta/pkcs11/src/pkcs11_attributes.c
clementfaure: ok

ta/pkcs11/src/pkcs11_attributes.c
clementfaure: Ok

etienne-lms (commented):
"ta: pkcs11: check returned value of mbedtls_pk_rsa()".

core/arch/arm/tee/entry_fast.c
clementfaure: Ok

clementfaure:
Tags applied

core/kernel/dt_driver.c
jforissier: Not addressed.

clementfaure: Sorry about that ..

clementfaure:
Add missing fix.

jforissier:
@clementfaure no problem. For "core: kernel: check device tree property pointer":

clementfaure:
Tag applied

core/kernel/dt_driver.c
```diff
@@ -264,8 +264,10 @@ static TEE_Result device_from_provider_prop(struct dt_driver_provider *prv,
 	pargs->fdt = fdt;
 	pargs->phandle_node = phandle_node;
 	pargs->args_count = prv->provider_cells;
-	for (n = 0; n < prv->provider_cells; n++)
+	for (n = 0; n < prv->provider_cells; n++) {
+		assert(prop);
```
etienne-lms, line 268: nitpicking
```c
	for (n = 0; n < prv->provider_cells; n++) {
		assert(prop);
		pargs->args[n] = fdt32_to_cpu(prop[n]);
	}
```
or 
```c
	assert(!prv->provider_cells || prop);
	for (n = 0; n < prv->provider_cells; n++)
		pargs->args[n] = fdt32_to_cpu(prop[n]);
```
?

core/mm/core_mmu.c
etienne-lms: I agree

etienne-lms (commented):
"core: kernel: check device tree property pointer" (with comment addressed or not)
and "core: arm: fix dead code when ARM32 is not defined".

core/mm/core_mmu.c
clementfaure: Ok for `assert()` then

etienne-lms: For commit "core: mmu: check pointer to manifest device tree": ok for assert()?
(it's the only commit that has no review tags)

jforissier: @clementfaure please add the `assert(fdt)` to "core: mmu: check pointer to manifest device tree" and check if Coverity is happy about it. Then you may add my `Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>` to that commit.

clementfaure: Coverity is happy, I was waiting for a tag to force-push the PR.

jforissier:
I believe this is safe to merge now although this is post-rc1, @jenswi-linaro?

ldelf/ta_elf_rel.c
```diff
@@ -283,7 +283,8 @@ static void resolve_sym(const char *name, vaddr_t *val, struct ta_elf **mod,
 		if (err_if_not_found)
 			err(res, "Symbol %s not found", name);
 		else
-			*val = 0;
+			if (val)
```
jenswi-linaro, line 286: I believe the indentation is misleading, we should be better off writing this as:
``` C
else if (val)
        *val = 0;
```

jenswi-linaro:
> I believe this is safe to merge now although this is post-rc1, @jenswi-linaro?

Sure, but we are not fixing any real problems. None of the patches are supposed to change the behaviour if I understand it correctly.

jforissier:
> > I believe this is safe to merge now although this is post-rc1, @jenswi-linaro?
> 
> Sure, but we are not fixing any real problems. None of the patches are supposed to change the behaviour if I understand it correctly.

Indeed, that's why I would initially abstain from taking this now, but didn't you mention Coverity fixes in the upcoming blog entry you prepared?

ldelf/ta_elf_rel.c
clementfaure: Alright

jenswi-linaro:
> Indeed, that's why I would initially abstain from taking this now, but didn't you mention Coverity fixes in the upcoming blog entry you prepared?

Yes, but I was probably referring to https://github.com/OP-TEE/optee_os/pull/6591

PR #6584, opened 2024-01-09
Clear FFA_MEM_RETRIEVE_RESP memory descriptor fields

imre-kis-arm:
Clear the memory descriptors in `FFA_MEM_RETRIEVE_RESP` calls in order to set the reserved fields to zero. The caller might check if the reserved fields are zero as it is stated in the FF-A spec. With FF-A v1.1 the memory transaction descriptor's 4 byte field at offset 24 has changed from reserved (MBZ) to Endpoint memory access descriptor size (non-zero). With the reserved field not cleared in the v1.0 descriptor, the caller cannot verify if it got the right version of the memory transaction descriptor.

This issue only affects the `FFA_MEM_RETRIEVE_RESP` call at the S-EL1 <-> S-EL0 interface, in all other cases the descriptors are cleared properly.

### DIFF
--- a/core/arch/arm/kernel/spmc_sp_handler.c
+++ b/core/arch/arm/kernel/spmc_sp_handler.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2021-2023, Arm Limited
+ * Copyright (c) 2021-2024, Arm Limited
  */
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please add a `core: ffa:` prefix to the commit title and copy the description you gave here into the commit message. With that:

imre-kis-arm:
Changed the commit message as requested and added the tag.

PR #6591, opened 2024-01-10
Coverity fixes

clementfaure:
Follow-up of https://github.com/OP-TEE/optee_os/pull/6582

Additional issues have been found.  e9642d1504935bc56970154cb999866892f0c633 and 39ef24b2c76bd7b9ba389d893c1e01d4a7c6f9f3 are the most important one of the pull-request. Let me know what you think.

### DIFF
--- a/core/crypto/cbc-mac.c
+++ b/core/crypto/cbc-mac.c
@@ -83,7 +83,7 @@ static TEE_Result crypto_cbc_mac_update(struct crypto_mac_ctx *ctx,
 	nblocks = MIN(len / mc->block_len,
 		      (size_t)CFG_CRYPTO_CBC_MAC_BUNDLE_BLOCKS);
 	if (nblocks > 1)
-		out_tmp = malloc(nblocks * mc->block_len);
+		out_tmp = calloc(nblocks, mc->block_len);
 
 	while (len >= mc->block_len) {
 		nblocks = MIN(len / mc->block_len,
--- a/core/crypto/sm3.c
+++ b/core/crypto/sm3.c
@@ -50,6 +50,20 @@ void sm3_init(struct sm3_context *ctx)
 	ctx->state[7] = 0xB0FB0E4E;
 }
 
+#define SHL(x, n)	((x) << (n))
+
+static uint32_t rotl(uint32_t val, int shift)
+{
+	shift &= 0x1F;
+
+	if (shift == 0)
+		return val;
+
+	return SHL(val, shift) | (val >> (32 - shift));
+}
+
+#define ROTL(x, n)	rotl((x), (n))
+
 static void __maybe_unused sm3_process(struct sm3_context *ctx,
 				       const uint8_t data[64])
 {
@@ -86,9 +100,6 @@ static void __maybe_unused sm3_process(struct sm3_context *ctx,
 
 #define GG0(x, y, z)	((x) ^ (y) ^ (z))
 #define GG1(x, y, z)	(((x) & (y)) | ((~(x)) & (z)))
-
-#define SHL(x, n)	((x) << (n))
-#define ROTL(x, n)	(SHL((x), (n) & 0x1F) | ((x) >> (32 - ((n) & 0x1F))))
 
 #define P0(x)	((x) ^ ROTL((x), 9) ^ ROTL((x), 17))
 #define P1(x)	((x) ^ ROTL((x), 15) ^ ROTL((x), 23))
--- a/core/drivers/gpio/gpio.c
+++ b/core/drivers/gpio/gpio.c
@@ -32,14 +32,19 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
 
 static char *gpio_get_dt_prop_name(const char *gpio_name)
 {
+	int ret = 0;
 	char *prop_name = NULL;
-	int max_len = strlen(gpio_name) + strlen("-gpios") + 1;
+	size_t max_len = strlen(gpio_name) + strlen("-gpios") + 1;
 
 	prop_name = calloc(1, max_len);
 	if (!prop_name)
 		return NULL;
 
-	snprintf(prop_name, max_len, "%s-gpios", gpio_name);
+	ret = snprintf(prop_name, max_len, "%s-gpios", gpio_name);
+	if (ret < 0 || (size_t)ret >= max_len) {
+		free(prop_name);
+		return NULL;
+	}
 
 	return prop_name;
 }
--- a/core/pta/stats.c
+++ b/core/pta/stats.c
@@ -155,6 +155,7 @@ static TEE_Result get_user_ta_stats(uint32_t type,
 static TEE_Result get_system_time(uint32_t type,
 				  TEE_Param p[TEE_NUM_PARAMS])
 {
+	TEE_Result ret = TEE_ERROR_GENERIC;
 	TEE_Time ree_time = { };
 	TEE_Time tee_time = { };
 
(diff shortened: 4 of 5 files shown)


core/drivers/gpio/gpio.c
```diff
@@ -32,14 +32,17 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
...
+	size_t max_len = strlen(gpio_name) + strlen("-gpios") + 1;
 
 	prop_name = calloc(1, max_len);
 	if (!prop_name)
 		return NULL;
 
-	snprintf(prop_name, max_len, "%s-gpios", gpio_name);
+	ret = snprintf(prop_name, max_len, "%s-gpios", gpio_name);
+	if (ret < 0 || (size_t)ret >= max_len)
+		return NULL;
```
jenswi-linaro, line 45: You need to free `prop_name` before returning.

core/crypto/sm3.c
```diff
@@ -55,7 +55,7 @@ static void __maybe_unused sm3_process(struct sm3_context *ctx,
 {
 	uint32_t SS1, SS2, TT1, TT2, W[68], W1[64];
 	uint32_t A, B, C, D, E, F, G, H;
-	uint32_t T[64];
+	uint64_t T[64];
```
jenswi-linaro, line 58: I'm not sure this will give the desired result. Please fix the `SHL()` macro instead.

core/crypto/cbc-mac.c
```diff
@@ -83,7 +83,7 @@ static TEE_Result crypto_cbc_mac_update(struct crypto_mac_ctx *ctx,
 	nblocks = MIN(len / mc->block_len,
 		      (size_t)CFG_CRYPTO_CBC_MAC_BUNDLE_BLOCKS);
 	if (nblocks > 1)
-		out_tmp = malloc(nblocks * mc->block_len);
+		out_tmp = calloc(1, nblocks * mc->block_len);
```
jenswi-linaro, line 86: Or even better `calloc(nblocks, mc->block_len)`.

jenswi-linaro (commented):
In especially the commit "core: initialize buffer with calloc()", please mention that this fixes a coverity warning since I suspect it's a false positive.

core/drivers/gpio/gpio.c
clementfaure: Oh right

core/crypto/sm3.c
clementfaure: I will proceed as I mentionned earlier then https://github.com/OP-TEE/optee_os/pull/6591#issuecomment-1885208252

clementfaure:
Please let me know what I can do about the CI error `CHECK: Macro argument reuse 'n' - possible side-effects?`

jforissier:
> Please let me know what I can do about the CI error `CHECK: Macro argument reuse 'n' - possible side-effects?`

Ideally, a GCC "statement expression" should be used, see https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html.

jenswi-linaro:
> > Please let me know what I can do about the CI error `CHECK: Macro argument reuse 'n' - possible side-effects?`
> 
> Ideally, a GCC "statement expression" should be used, see https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html. Simple example: `UDIV_ROUND_NEAREST()` in `lib/libutils/ext/include/util.h`.

In this case, it might be easier to just add a static `rotl()` function and keep the `ROTL()` macro as a simple wrapper for that function to keep other changes at a minimum.

core/crypto/sm3.c
```diff
@@ -50,6 +50,16 @@ void sm3_init(struct sm3_context *ctx)
 	ctx->state[7] = 0xB0FB0E4E;
 }
 
+static inline uint32_t rotl(uint32_t val, int shift)
+{
+	if (shift == 0)
+		return val;
+
+#define SHL(x, n)	((x) << (n))
```
jforissier, line 58: This define in the middle of a function looks kinda weird. How about moving the `rotl()` function definition just above the `#define ROTL(x, n) rotl((x), (n))` line?

core/crypto/sm3.c
```diff
@@ -50,6 +50,16 @@ void sm3_init(struct sm3_context *ctx)
 	ctx->state[7] = 0xB0FB0E4E;
 }
 
+static inline uint32_t rotl(uint32_t val, int shift)
```
jforissier, line 53: Please drop then `inline`

jforissier (commented):
Comments on "core: crypto: sm3: fix undefined behavior on right shifting operation".

core/crypto/sm3.c
```diff
@@ -50,6 +50,16 @@ void sm3_init(struct sm3_context *ctx)
...
+{
+	if (shift == 0)
```
jenswi-linaro, line 55: Please add a
``` C
shift &= 0x1F;
```
before this to make sure we're testing the right thing. With this you can also remove the masking of shift further down.

core/crypto/sm3.c
clementfaure: I tried to match the current style with macros defined in the middle of the function.
I need `SHL()` in `rotl()`
```
core/crypto/sm3.c: In function ‘rotl’:
core/crypto/sm3.c:58:16: error: implicit declaration of function ‘SHL’ [-Werror=implicit-function-declaration]
   58 |         return SHL(val, shift & 0x1F) | (val >> (32 - (shift & 0x1F)));
      |                ^~~
core/crypto/sm3.c:58:16: warning: nested extern declaration of ‘SHL’ [-Wnested-externs]
```

clementfaure: In the middle of the `sm3_process()` function?

core/crypto/sm3.c
clementfaure: Ok

core/crypto/sm3.c
```diff
@@ -87,8 +97,7 @@ static void __maybe_unused sm3_process(struct sm3_context *ctx,
 #define GG0(x, y, z)	((x) ^ (y) ^ (z))
 #define GG1(x, y, z)	(((x) & (y)) | ((~(x)) & (z)))
 
-#define SHL(x, n)	((x) << (n))
-#define ROTL(x, n)	(SHL((x), (n) & 0x1F) | ((x) >> (32 - ((n) & 0x1F))))
+#define ROTL(x, n)	rotl((x), (n))
```
jforissier, line 100: What I am proposing is:
```C
#define SHL(x, n)	((x) << (n))
static uint32_t rotl(uint32_t val, int shift)
{
	...
}
#define ROTL(x, n)	rotl((x), (n))
```

jforissier (commented):
> core: crypto: sm3: fix undefined behavior on right shifting operation 
>
> [...]
> Promote SS1 value to 64 bits to avoid the issue.

It is not what the commit is now doing.

core/crypto/sm3.c
```diff
@@ -50,6 +50,19 @@ void sm3_init(struct sm3_context *ctx)
 	ctx->state[7] = 0xB0FB0E4E;
 }
 
+#define SHL(x, n)	((x) << (n))
+static uint32_t rotl(uint32_t val, int shift)
```
etienne-lms, line 54: add an empty line above?
Or move the macro definition into the function body (only used there).

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for the series, with comments addressed for commit 
"core: crypto: sm3: fix undefined behavior on right shifting operation".

jenswi-linaro:
With the previous comments for "core: crypto: sm3: fix undefined behavior on right shifting operation" addressed, please apply for all commits:

core/crypto/sm3.c
clementfaure: I will add an empty line then since https://github.com/OP-TEE/optee_os/pull/6591#discussion_r1448575965

core/pta/stats.c
```diff
@@ -164,8 +165,13 @@ static TEE_Result get_system_time(uint32_t type,
 			    TEE_PARAM_TYPE_NONE) != type)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	tee_time_get_sys_time(&tee_time);
-	tee_time_get_ree_time(&ree_time);
+	ret = tee_time_get_sys_time(&tee_time);
+	if (ret)
+		return ret;
+
+	ret = tee_time_get_ree_time(&ree_time)
```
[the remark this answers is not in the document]
clementfaure: That is already fixed

jforissier:
Once "core: crypto: sm3: fix undefined behavior on right shifting operation" is fixed, all commits are:

clementfaure:
Thanks for the review!

PR #6594, opened 2024-01-11
riscv: mm: Set A/D bits of PTE(page table entry) by default

gagachang:
According to RISC-V privileged ISA manual:
Each leaf PTE contains an accessed (A) and dirty (D) bit. The A bit indicates the virtual page has been read, written, or fetched from since the last time the A bit was cleared. The D bit indicates the virtual page has been written since the last time the D bit was cleared. When a virtual page is accessed and the A bit is clear, or is written and the D bit is clear, a page-fault exception is raised.

And the manual also suggests:
If the supervisor software does not rely on accessed and/or dirty bits, it should always set them to 1 in the PTE to improve performance.

Since OP-TEE does not rely on A/D bits, we by default set them to 1 to avoid unnecessary page-fault exceptions when OP-TEE touches those pages.


maroueneboubakri:
@gagachang thanks for the patch.

PR #6599, opened 2024-01-12
plat-k3: sa2ul_rng: Use mutex instead of spinlock for critical section

glneo:
While spinlock are slightly more lightweight, they currently require that interrupts are disabled during the critical section. If this section is long enough it can have a negative affect on realtime sensitive tasks that require deterministic preemption.

As our RNG gathering can loop while waiting for new random numbers to become available we cannot know how long this section will take, so we should use a mutex. Do that here.

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul_rng.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul_rng.c
@@ -61,7 +61,7 @@
 
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, RNG_BASE, RNG_REG_SIZE);
 
-static unsigned int rng_lock = SPINLOCK_UNLOCK;
+static struct mutex fifo_lock = MUTEX_INITIALIZER;
 static vaddr_t rng;
 
 static bool sa2ul_rng_is_enabled(void)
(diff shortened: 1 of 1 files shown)


jforissier:
@jenswi-linaro is `hw_get_random_bytes()` guaranteed to be called in a thread context? (which is needed I believe for mutexes, right?). I am especially wondering about _start() -> plat_get_random_stack_canaries() -> crypto_rng_read() -> hw_get_random_bytes().

That aside this LGTM:

glneo:
Yeah I was thinking about that earlier, maybe in mutex() we could check if we are in early boot or some other non threading context. As we should always be single threaded at that point then just have the mutex decay into a NOP as there should never be any contention at that point.

As K3 doesn't use RNG during boot currently I just decided to punt the problem for now :)

bryanbrattlof:
Thanks @glneo this solves the large majority of our mysterious 100us latency samples.

`Reviewed-by: Bryan Brattlof <bb@ti.com>`

jenswi-linaro:
> @jenswi-linaro is `hw_get_random_bytes()` guaranteed to be called in a thread context? (which is needed I believe for mutexes, right?). I am especially wondering about _start() -> plat_get_random_stack_canaries() -> crypto_rng_read() -> hw_get_random_bytes().

Yes, that should be OK. During boot we have a fake thread and since (as @glneo pointed out) there is no contention on any of the mutexes they are more or less NOPs.

jenswi-linaro:

PR #6605, opened 2024-01-15
ldelf: aarch32: Accept ELFOSABI_ARM as OS ABI

b49020:
Rust TAs built for no-std mode targeting 32-bit Arm architecture use ELFOSABI_ARM as the OS ABI within ELF header. So allow ldelf to load those Rust TAs built for 32-bit Arm.

### DIFF
--- a/ldelf/ta_elf.c
+++ b/ldelf/ta_elf.c
@@ -104,14 +104,22 @@ static TEE_Result e32_parse_ehdr(struct ta_elf *elf, Elf32_Ehdr *ehdr)
 	if (ehdr->e_ident[EI_VERSION] != EV_CURRENT ||
 	    ehdr->e_ident[EI_CLASS] != ELFCLASS32 ||
 	    ehdr->e_ident[EI_DATA] != ELFDATA2LSB ||
-	    ehdr->e_ident[EI_OSABI] != ELFOSABI_NONE ||
+	    (ehdr->e_ident[EI_OSABI] != ELFOSABI_NONE &&
+	     ehdr->e_ident[EI_OSABI] != ELFOSABI_ARM) ||
 	    ehdr->e_type != ET_DYN || ehdr->e_machine != EM_ARM ||
-	    (ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_VERSION ||
 #ifndef CFG_WITH_VFP
 	    (ehdr->e_flags & EF_ARM_ABI_FLOAT_HARD) ||
 #endif
 	    ehdr->e_phentsize != sizeof(Elf32_Phdr) ||
 	    ehdr->e_shentsize != sizeof(Elf32_Shdr))
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_NONE &&
+	    (ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_V5)
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_ARM &&
+	    (ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_UNKNOWN)
 		return TEE_ERROR_BAD_FORMAT;
 
 	elf->is_32bit = true;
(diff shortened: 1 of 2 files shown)


b49020:
Looks like the CI build failure isn't related to this PR.

ldelf/ta_elf.c
```diff
@@ -104,16 +104,27 @@ static TEE_Result e32_parse_ehdr(struct ta_elf *elf, Elf32_Ehdr *ehdr)
...
-	    (ehdr->e_flags & EF_ARM_ABI_FLOAT_HARD) ||
-#endif
 	    ehdr->e_phentsize != sizeof(Elf32_Phdr) ||
 	    ehdr->e_shentsize != sizeof(Elf32_Shdr))
 		return TEE_ERROR_BAD_FORMAT;
 
+	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_NONE) {
+		if ((ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_V5)
+			return TEE_ERROR_BAD_FORMAT;
+#ifndef CFG_WITH_VFP
```
jenswi-linaro, line 117: Please use ` if (IS_ENABLED(xxx))` instead of `#ifdef` where possible.

ldelf/ta_elf.c
```diff
@@ -104,16 +104,27 @@ static TEE_Result e32_parse_ehdr(struct ta_elf *elf, Elf32_Ehdr *ehdr)
...
+		if (ehdr->e_flags & EF_ARM_ABI_FLOAT_HARD)
+			return TEE_ERROR_BAD_FORMAT;
+#endif
+	}
+
+	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_ARM) {
+		if ((ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_UNKNOWN)
```
jenswi-linaro, line 124: This sucks somewhat. I guess Rust is going to use floating point if it feels like it, but we'll never know.

jforissier: We should perhaps accept both `EF_ARM_ABI_UNKNOWN` and `EF_ARM_ABI_V5` here? We do support EABIv5 so let's be prepared in case Rust sets the flag.

ldelf/ta_elf.c
b49020: Sure I will switch to that.

ldelf/ta_elf.c
b49020: Binutils-gdb reference here [1] indicates that there is 1:1 relation among `ELFOSABI_ARM` and `EF_ARM_ABI_UNKNOWN` in flags field. However, for floating point case I agree corresponding bit `EF_ARM_ABI_FLOAT_HARD` should be checked here as well for Rust case. I will make that common instead as earlier.

[1] https://github.com/bminor/binutils-gdb/blob/master/bfd/elf32-arm.c#L17661

b49020: BTW, Linux arm32 ELF parser does the same here: https://github.com/torvalds/linux/blob/master/arch/arm/kernel/elf.c#L26

jforissier:

b49020:
Thanks, tags applied.

ldelf/ta_elf.c
```diff
@@ -104,16 +104,26 @@ static TEE_Result e32_parse_ehdr(struct ta_elf *elf, Elf32_Ehdr *ehdr)
...
 
+	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_NONE) {
+		if ((ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_V5)
+			return TEE_ERROR_BAD_FORMAT;
+	}
+
+	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_ARM) {
+		if ((ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_UNKNOWN)
+			return TEE_ERROR_BAD_FORMAT;
+	}
```
etienne-lms, line 125: Nitpicking: wrap the `if`'s?
```c
	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_NONE &&
	    (ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_V5)
		return TEE_ERROR_BAD_FORMAT;

	if (ehdr->e_ident[EI_OSABI] == ELFOSABI_ARM &&
	    (ehdr->e_flags & EF_ARM_ABIMASK) != EF_ARM_ABI_UNKNOWN)
		return TEE_ERROR_BAD_FORMAT;

etienne-lms (commented):

ldelf/ta_elf.c
b49020: Looks better, incorporated.

b49020:
Addressed comment and tag applied, thanks.

PR #6610, opened 2024-01-16
core: ffa: Read FF-A version from the SP manifest

imre-kis-arm:
Read the SP's FF-A version from the ffa-version property of the SP manifest. This property is mandatory according to the FF-A specification.
SPs are still able to do runtime version negotiation via the FFA_VERSION interface.

This PR is not a critical fix, so it is intended to be merged after the 4.1.0 release.

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2020-2023, Arm Limited.
+ * Copyright (c) 2020-2024, Arm Limited.
  */
 #include <bench.h>
 #include <crypto/crypto.h>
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Would it make sense to add a "Fixes:" tag?

jenswi-linaro (commented):

imre-kis-arm:
> Would it make sense to add a "Fixes:" tag?

I can't really find a commit which I would mark as the cause of the issue. Unfortunately there are multiple [mandatory manifest properties](https://review.trustedfirmware.org/q/topic:%22spmc-test-doc%22) which are not handled yet. I guess this was not a pressing issue while OP-TEE only had support for FF-A v1.0, and all the other properties (Run-time EL, Execution state) can only have a single value because that's what OP-TEE currently implements by design.
Hopefully I will have some time allocated for increasing the FF-A compliance of the SPMC and then I can add at least a check for all the mandatory properties.

I've added the Acked-by tag.

jenswi-linaro:
I'd say the commit that it is the commit that switched to FF-A 1.1 since this fixes an issue with FF-A 1.1.

By using Fixes tags it's easier to identify which commits are needed in an eventual stable branch. If that FF-A 1.1 enabler commit is in the stable branch, then we want this commit too.

jenswi-linaro:
After some thought, a fixes tag might be a bit much here. Let's keep it as it is.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -1408,6 +1404,27 @@ static TEE_Result read_ns_interrupts_action(const void *fdt,
...
+	uint32_t ffa_version = 0;
+
+	res = sp_dt_get_u32(fdt, 0, "ffa-version", &ffa_version);
+	if (res) {
+		EMSG("Mandatory property is missing: ffa-version");
+		return res;
+	}
+
+	if (ffa_version != FFA_VERSION_1_0 && ffa_version != FFA_VERSION_1_1) {
+		EMSG("Invalid FF-A version value: 0x%08X", ffa_version);
```
etienne-lms, line 1419: `0x%08"PRIx32, ffa_version);`

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with comment addressed.

core/arch/arm/kernel/secure_partition.c
imre-kis-arm: Done

PR #6615, opened 2024-01-16
drivers: stm32_i2c: apply pinctrl config at init

etienne-lms:
Main change: add missing load of stm32_i2c pinctrl state at driver init.
(Fixes: 73ba32eb0f6c ("drivers: stm32_i2c: support CFG_DRIVERS_PINCTRL"))

An preparatory commit simplifies local function `i2c_config_analog_filter()`.

### DIFF
--- a/core/drivers/stm32_i2c.c
+++ b/core/drivers/stm32_i2c.c
@@ -657,17 +657,11 @@ static int i2c_setup_timing(struct i2c_handle_s *hi2c,
  * Configure I2C Analog noise filter.
  * @hi2c: I2C handle structure
  * @analog_filter_on: True if enabling analog filter, false otherwise
- * Return 0 on success or a negative value
  */
-static int i2c_config_analog_filter(struct i2c_handle_s *hi2c,
-				    bool analog_filter_on)
+static void i2c_config_analog_filter(struct i2c_handle_s *hi2c,
+				     bool analog_filter_on)
 {
 	vaddr_t base = get_base(hi2c);
-
-	if (hi2c->i2c_state != I2C_STATE_READY)
-		return -1;
-
-	hi2c->i2c_state = I2C_STATE_BUSY;
 
 	/* Disable the selected I2C peripheral */
 	io_clrbits32(base + I2C_CR1, I2C_CR1_PE);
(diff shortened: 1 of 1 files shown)


core/drivers/stm32_i2c.c
```diff
@@ -824,16 +814,14 @@ int stm32_i2c_init(struct i2c_handle_s *hi2c,
...
-		DMSG("I2C analog filter error %d", rc);
+	i2c_config_analog_filter(hi2c, init_data->analog_filter);
 
 	if (IS_ENABLED(CFG_STM32MP13))
 		stm32_pinctrl_set_secure_cfg(hi2c->pinctrl, true);
 
 	clk_disable(hi2c->clock);
 
-	return rc;
+	return 0;
```
GseoC, line 827: This makes me realize that stm32_i2c_init() returns an int and that the probe function uses the return code as a TEE_Result, can you please make the change?

The exported function are also in the same state but don't require a change for now

core/drivers/stm32_i2c.c
```diff
@@ -821,6 +821,9 @@ int stm32_i2c_init(struct i2c_handle_s *hi2c,
 
 	clk_disable(hi2c->clock);
 
+	if (hi2c->pinctrl && pinctrl_apply_state(hi2c->pinctrl))
+		return -1;
```
GseoC, line 825: Linked to other comment, needs to be updated to use pinctrl_apply_state() error code

etienne-lms: This change simply makes `stm32_i2c_init()` to report an error on pinctrl issues. I think it's simpler, as a fixup commit for coming next release tag. Changing all stm32_i2c API function to use `TEE_Result` code instead of `int` value is a good idea but i'll rather treat that after next tag and on a specific P-R, if you're ok.

core/drivers/stm32_i2c.c
etienne-lms: I'll append a fixup commit for `stm32_i2c_probe()`.

core/drivers/stm32_i2c.c
GseoC: Ok :)

GseoC:

PR #6616, opened 2024-01-16
plat-rcar: disable HW RNG by default (and try to workaround the issue)

lorc:
During repetitive testing I noticed that sometimes OP-TEE panics due to problems with Renesas HW random number generator. While proposed workaround seems to help, I believe it is better to disable this feature by default, at least until I can get some clarification from Renesas side, which can take quite time.

### DIFF
--- a/core/arch/arm/plat-rcar/romapi.c
+++ b/core/arch/arm/plat-rcar/romapi.c
@@ -121,6 +121,7 @@ uint32_t plat_rom_getrndvector(uint8_t rndbuff[PLAT_RND_VECTOR_SZ],
 			       uint8_t *scratch, uint32_t scratch_sz)
 {
 	uint32_t ret = -1;
+	int try = 0;
 	paddr_t func_addr = romapi_getrndvector[get_api_table_index()];
 	paddr_t rndbuff_pa = va2pa(rndbuff);
 	paddr_t scratch_pa = va2pa(scratch);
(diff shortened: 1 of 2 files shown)


core/arch/arm/plat-rcar/romapi.c
```diff
@@ -129,7 +130,13 @@ uint32_t plat_rom_getrndvector(uint8_t rndbuff[PLAT_RND_VECTOR_SZ],
 	assert(rndbuff_pa % RCAR_CACHE_LINE_SZ == 0);
 	assert(scratch_pa % RCAR_CACHE_LINE_SZ == 0);
 
-	ret = plat_call_romapi(func_addr, rndbuff_pa, scratch_pa, scratch_sz);
+	while (try++ < 3) {
+		ret = plat_call_romapi(func_addr, rndbuff_pa, scratch_pa,
+				       scratch_sz);
+		if (ret == 0)
+			break;
+		IMSG("ROM_GetRndVector() returned 0x%X", ret);
```
jenswi-linaro, line 138: `"%#"PRIx32`, gives `0x` prefix automatically and `PRIx32` for a 32-bit type using lower case letters.

jenswi-linaro (commented):

etienne-lms (commented):

lorc:
Thank you. Tags are applied.

PR #6618, opened 2024-01-17
core: scmi-server: sync makefile with changes on SCP-firmware

t-akashi:
On SCP-firmware, extra two files, power_domain_state_checks.c and power_domain_notifications.c, have been added to power-domain module. So sub.mk should be aligned with these changes.

### DIFF
--- a/core/lib/scmi-server/sub.mk
+++ b/core/lib/scmi-server/sub.mk
@@ -209,7 +209,13 @@ $(eval $(call scpfw-embed-optee-module,reset))
 $(eval $(call scpfw-embed-optee-module,smt))
 
 srcs-$(CFG_SCPFW_MOD_CLOCK) += $(scpfw-path)/module/clock/src/clock_tree_management.c
-srcs-$(CFG_SCPFW_MOD_POWER_DOMAIN) += $(scpfw-path)/module/power_domain/src/power_domain_utils.c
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):
The change This change needs to be synced with a SCP-firmware sourcetree version update, for CI (_.github/workflow/ci.yml_) and for the the platforms ([OP-TEE/manifest.git](https://github.com/OP-TEE/manifest.git)) that enable these config switches.

Regarding _ci.yml_, the command need to be changed to sync on a commit SHA1 instead of a branch or tag reference:
>          function download_scp_firmware() { git clone --depth 1 https://github.com/ARM-software/SCP-firmware.git -b v2.13.0 $HOME/scp-firmware || echo Nervermind; }

t-akashi:
The latest master branch of SCP-firmware (after v2.13.0) has not been tagged yet.
Is it too early to merge this commit? I guess that a new tag, v2.14.0?, will come around April.
Or should we specify a specific commit in "revision" if feasible?

jforissier:
> The latest master branch of SCP-firmware (after v2.13.0) has not been tagged yet. Is it too early to merge this commit? I guess that a new tag, v2.14.0?, will come around April. Or should we specify a specific commit in "revision" if feasible?

Specifying a SHA1 in "revision" is possible as you can see here: https://github.com/OP-TEE/manifest/blob/4.0.0/stm32mp1.xml#L26

PR #6619, opened 2024-01-17
core: imx: disable ELE support on imx8ulp, imx93 by default

sahilnxp:
On imx8ulp and imx93, there is only one MU to comunicate with
ELE, which cannot be dedicated on OP-TEE side all the time.
There may be ELE services running on Linux side, which can
cause conflict with OP-TEE. So disablig ELE by default for now.


jforissier (commented):

To be included in the coming release I suppose?

clementfaure:
Ideally yes if possible @jforissier

sahilnxp:
Thanks @jforissier and @clementfaure for quick tags.
Tags updated.

sahilnxp:
Any more comments on this pull request ?

jforissier:
@sahilnxp all good, thanks for the reminder!

PR #6623, opened 2024-01-18
core: riscv: Add .sbss and .sdata sections to linker script

lyctw:
Currently, the unclean .sbss section in RISC-V binary could be
problematic. This is because variables such as puts_lock may have
non-zero initial values, leading to failures in cpu_spin_trylock().
To address this issue, merge .sbss into .bss so that it is properly
cleared on boot. Both OpenSBI [1] and Linux [2] follow this approach.

Also, to benefit from global pointer relaxation, add .sdata after the
__global_pointer$ symbol.

Link: https://github.com/riscv-software-src/opensbi/blob/bb90a9ebf6d9a2fe7726978d594e82cdbaad7799/firmware/fw_base.ldS#L84 [1]
Link: https://github.com/torvalds/linux/blob/296455ade1fdcf5f8f8c033201633b60946c589a/include/asm-generic/vmlinux.lds.h#L1146 [2]
Signed-off-by: Yu Chien Peter Lin <peterlin@andestech.com>

### DIFF
--- a/core/arch/riscv/kernel/kern.ld.S
+++ b/core/arch/riscv/kernel/kern.ld.S
@@ -152,12 +152,15 @@ SECTIONS
 		 * from __global_pointer$.
 		 */
 		PROVIDE(__global_pointer$ = . + 0x800 );
+		*(.sdata .sdata.* .gnu.linkonce.s.*)
 	}
 
 	/* uninitialized data */
 	.bss : {
 		__data_end = .;
 		__bss_start = .;
+		*(.sbss .sbss.*)
+		*(.gnu.linkonce.sb.*)
 		*(.bss .bss.*)
 		*(.gnu.linkonce.b.*)
 		*(COMMON)


jenswi-linaro:
If `CFG_NS_VIRTUALIZATION` isn't enabled, `__nex_bss` and friends are supposed to be empty defines. What does this have to do with `.sbss`?

lyctw:
Hi @jenswi-linaro 

The variables will be placed in the ".sbss" section instead of the ".bss" section. Since we are verifying OPTEE on an FPGA, which is never reset, the memory may contain random values.

RISC-V will only clear the ".bss" section now, as indicated by the range shown in "tee.dmp".

```
 29 .bss          0000ce70  000000000e08f360  000000000e08f360  00090358  2**4
                  ALLOC
 30 .sbss.start_addr 00000008  000000000e09c1d0  000000000e09c1d0  00090358  2**3
                  ALLOC
...
 44 .sbss.puts_lock 00000004  000000000e09c220  000000000e09c220  00090358  2**2
                  ALLOC
```

jenswi-linaro:
I don't doubt that you may need to do something with `.sbss`, but what has that to do with the empty `__xen_bss()` macro?

gagachang:
Hi @jenswi-linaro 

After out internal discussion, the `__nex_bss()` is irrelevant to this issue.
In our environment we don't define `CFG_NS_VIRTUALIZATION`, so `__nex_bss()` is empty.

jenswi-linaro:
Please remove the `__nex_bss()` part of the commit message and this should be good.

jforissier:
Where is `.bss` cleared on RISC-V? Arm has `clear_bss:` in `core/arch/arm/kernel/entry_a{32,64}.S`.
Perhaps `.sbss` should be cleared on its own? Aren't we losing some benefits of having a separate section (whatever they are) if it's merged with `.bss`?

gagachang:
> Where is `.bss` cleared on RISC-V? Arm has `clear_bss:` in `core/arch/arm/kernel/entry_a{32,64}.S`.

RISC-V clears `.bss` in [core/arch/riscv/kernel/entry.S](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/kernel/entry.S#L162).

> Aren't we losing some benefits of having a separate section (whatever they are) if it's merged with `.bss`?

One of RISC-V popular open-sources [openSBI](https://github.com/riscv-software-src/opensbi/blob/master/firmware/fw_base.ldS#L84) merges it with `.bss`.
So does [RISC-V linux kernel](https://github.com/torvalds/linux/blob/master/arch/riscv/kernel/vmlinux.lds.S#L159), the macro is expended [here](https://github.com/torvalds/linux/blob/296455ade1fdcf5f8f8c033201633b60946c589a/include/asm-generic/vmlinux.lds.h#L1146).

AFAIK, `sdata` and `sbss` are used for linker relaxation, such as RISC-V's gp relaxation. You can refer to [this website](https://www.sifive.com/blog/all-aboard-part-3-linker-relaxation-in-riscv-toolchain).
In general, a global pointer is put inside `sdata` and `sbss` sections.
We have already defined it in [core/arch/riscv/kernel/kern.ld.S](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/kernel/kern.ld.S#L154).
Linker can generate more efficient instruction based on this global pointer when CPU accesses variables in `sdata` and `sbss` sections.

jforissier:
> > Where is `.bss` cleared on RISC-V? Arm has `clear_bss:` in `core/arch/arm/kernel/entry_a{32,64}.S`.
> 
> RISC-V clears `.bss` in [core/arch/riscv/kernel/entry.S](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/kernel/entry.S#L162).
> 
> > Aren't we losing some benefits of having a separate section (whatever they are) if it's merged with `.bss`?
> 
> One of RISC-V popular open-sources [openSBI](https://github.com/riscv-software-src/opensbi/blob/master/firmware/fw_base.ldS#L84) merges it with `.bss`. So does [RISC-V linux kernel](https://github.com/torvalds/linux/blob/master/arch/riscv/kernel/vmlinux.lds.S#L159), the macro is expanded [here](https://github.com/torvalds/linux/blob/296455ade1fdcf5f8f8c033201633b60946c589a/include/asm-generic/vmlinux.lds.h#L1146).

Could you add these links to the commit description? Please use the permalink: https://github.com/riscv-software-src/opensbi/blob/bb90a9ebf6d9a2fe7726978d594e82cdbaad7799/firmware/fw_base.ldS#L84 for the first one (has a commit ID instead of "master" in the URL).

```
Merge .sbss into .bss so that it is properly cleared on boot. OpenSBI [1]
and Linux [2] do the same.

Link: https://github.com/riscv-software-src/opensbi/blob/bb90a9ebf6d9a2fe7726978d594e82cdbaad7799/firmware/fw_base.ldS#L84 [1]
Link: https://github.com/torvalds/linux/blob/296455ade1fdcf5f8f8c033201633b60946c589a/include/asm-generic/vmlinux.lds.h#L1146 [2]
Signed-off-by: ...
```

> AFAIK, `sdata` and `sbss` are used for linker relaxation, such as RISC-V's gp relaxation. You can refer to [this website](https://www.sifive.com/blog/all-aboard-part-3-linker-relaxation-in-riscv-toolchain). In general, a global pointer is put near `sdata` and `sbss` sections. We have already defined it in [core/arch/riscv/kernel/kern.ld.S](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/kernel/kern.ld.S#L154). Linker can generate more efficient instruction based on this global pointer when CPU accesses variables in `sdata` and `sbss` sections.

That's quite helpful, thanks. Shouldn't `*(.sdata)` be added at the end of the `.data` section then? This way (and with this PR) the global pointer would be more or less in the middle of [`.sdata`, `.sbss`] which is optimal IIUC.

> 
> > Aren't we losing some benefits of having a separate section (whatever they are) if it's merged with .bss?
> 
> I think it is okay since this PR puts `sbss` near the global pointer.

OK

lyctw:
Hi @jforissier 

Thanks for the suggestion. I have added .sdata with this commit. However, it should be added below the __global_pointer$ symbol, as the address is incremented by 0x800 in the linker script.

jforissier:

gagachang:
Hi @jforissier 
I think this PR is ready to be merged, thanks!

PR #6627, opened 2024-01-20
arm: plat: rcar: gen4: adjust memory map

lorc:
Adjust the OP-TEE memory map for Gen4/S4 SoC to reflect changes of IPL layout made by Renesas. Now BL31 starts at 0x46400000, so we have less memory for OP-TEE.

(Rcar S4 is still not available for general public, so Renesas have time to experiments with IPL layouts. We are expecting S4SK board to be available soon, so I hope this is the last change to memory layout...  )


jenswi-linaro (commented):

PR #6633, opened 2024-01-24
drivers: caam: add caam_hal_rng_pr_enabled() for 8QX, 8DX platforms

sahilnxp:
The SECO firmware enables the RNG prediction resistance by default. There is no need to read the CAAM RNG status registers.

### DIFF
--- a/core/drivers/crypto/caam/hal/common/hal_rng.c
+++ b/core/drivers/crypto/caam/hal/common/hal_rng.c
@@ -62,7 +62,10 @@ bool caam_hal_rng_key_loaded(vaddr_t baseaddr)
 	return io_caam_read32(baseaddr + RNG_STA) & RNG_STA_SKVN;
 }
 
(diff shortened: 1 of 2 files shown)


core/drivers/crypto/caam/hal/common/hal_rng.c
```diff
@@ -62,7 +62,7 @@ bool caam_hal_rng_key_loaded(vaddr_t baseaddr)
 	return io_caam_read32(baseaddr + RNG_STA) & RNG_STA_SKVN;
 }
```
jenswi-linaro, line 64: Please add a comment on when this function is supposed to be overridden.

jenswi-linaro (commented):
I'm not very fond of the way you're using a weak function instead of straight configuration variables. Anyway, with my comment addressed:

sahilnxp:
@jenswi-linaro  Comment addressed and tag updated.

clementfaure:

PR #6638, opened 2024-01-25
drivers: regulator: list voltages controlled by a GPIO

etienne-lms:
Implement .supported_voltages() operation in regulator_gpio.c driver to report voltage levels a level array that is required by scp-firmware voltage domain module when a regulator-gpio is exposed through an SCMI voltage domain. This change requires filed voltage_levels_uv[] array of struct regulator_gpio to be sorted in increasing level value.

### DIFF
--- a/core/drivers/regulator/regulator_gpio.c
+++ b/core/drivers/regulator/regulator_gpio.c
@@ -20,14 +20,16 @@ static_assert(GPIO_LEVEL_HIGH == 1 && GPIO_LEVEL_LOW == 0);
  * @enable_gpio: GPIO for the enable state of the regulator or NULL if always on
  * @enable_delay: Time (in microsecond) for the regulator to get enabled
  * @voltage_gpio: GPIO for the voltage level selection
+ * @levels_desc: Supported voltage levels description
  * @voltage_levels_uv: 2 cells array supported voltage levels, increasing order
- * @voltage_level_high: True if higher voltage level relates to active high GPIO
+ * @voltage_level_high: True if higher voltage level relates to GPIO state 1
  */
 struct regulator_gpio {
 	struct regulator regulator;
 	struct gpio *enable_gpio;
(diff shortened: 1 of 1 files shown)


GseoC (commented):

core/drivers/regulator/regulator_gpio.c
```diff
@@ -175,13 +202,25 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
...
+	regu->levels_desc.type = VOLTAGE_TYPE_FULL_LIST;
+	regu->levels_desc.num_levels = 2;
+	if (level0 < level1) {
+		regu->voltage_levels_uv[0] = level0;
+		regu->voltage_levels_uv[1] = level1;
 		regu->voltage_level_high = true;
+	} else {
+		regu->voltage_levels_uv[0] = level1;
+		regu->voltage_levels_uv[1] = level0;
+		regu->voltage_level_high = false;
```
GseoC, line 223: actually sorry, with is this:

`regu->voltage_level_high = false;`

instead of
`regu->voltage_level_high = true;`

It is sorted, isn't it?

etienne-lms: `level0` relates to the voltage level selected by GPIO state 0 and `level1` to the one selected by GPIO state 1.

`regu->voltage_levels_uv[0]` relates the to lower supported voltage level while `regu->voltage_levels_uv[1]` relates the to higher supported voltage level. This it because supported voltage level list must be sorted in increasing level value, see [`enum voltage_type`](https://github.com/OP-TEE/optee_os/blob/4.1.0/core/include/drivers/regulator.h#L54-L56).

Hence `regu->voltage_level_high` stores whether GPIO state 1 relate to the higher level or to the lower level value.

etienne-lms:
Review tag applied. Thanks.

PR #6641, opened 2024-01-25
add support timeout for notify and condvar

ZheTing815:
We have added a timeout parameter to the rpc notify for greater flexibility. The tee driver checks this parameter to decide which wait_for_completion will use. If the timeout is set to 0, it means waiting MAX_SCHEDULE_TIMEOUT.
The corresponding patch is here. [optee: support wq_sleep_timeout
](https://patchwork.kernel.org/project/linux-mediatek/patch/20240125052744.18866-1-gavin.liu@mediatek.com/)

### DIFF
--- a/core/include/kernel/mutex.h
+++ b/core/include/kernel/mutex.h
@@ -95,10 +95,26 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
 void condvar_wait_debug(struct condvar *cv, struct mutex *m,
 			const char *fname, int lineno);
 #define condvar_wait(cv, m) condvar_wait_debug((cv), (m), __FILE__, __LINE__)
+
+/*
+ * Return TEE_ERROR_TIMEOUT if the normal world returns before
+ * the condvar has been signaled.
+ */
+TEE_Result condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m,
+				      uint32_t timeout_ms, const char *fname,
+				      int lineno);
+#define condvar_wait_timeout(cv, m, timeout_ms) \
+	condvar_wait_timeout_debug((cv), (m), (timeout_ms), __FILE__, __LINE__)
 #else
 void condvar_signal(struct condvar *cv);
 void condvar_broadcast(struct condvar *cv);
 void condvar_wait(struct condvar *cv, struct mutex *m);
+/*
+ * Return TEE_ERROR_TIMEOUT if the normal world returns before
+ * the condvar has been signaled.
+ */
+TEE_Result condvar_wait_timeout(struct condvar *cv, struct mutex *m,
+				uint32_t timeout_ms);
 #endif
 
 #endif /*__KERNEL_MUTEX_H*/
--- a/core/include/kernel/notif.h
+++ b/core/include/kernel/notif.h
@@ -107,6 +107,11 @@ void notif_free_async_value(uint32_t value);
 TEE_Result notif_wait(uint32_t value);
 
 /*
+ * Wait timeout in normal world for a value to be sent by notif_send()
+ */
+TEE_Result notif_wait_timeout(uint32_t value, uint32_t timeout_ms);
+
+/*
  * Send an asynchronous value, note that it must be <= NOTIF_ASYNC_VALUE_MAX
  */
 #if defined(CFG_CORE_ASYNC_NOTIF)
--- a/core/include/kernel/wait_queue.h
+++ b/core/include/kernel/wait_queue.h
@@ -5,6 +5,7 @@
 #ifndef __KERNEL_WAIT_QUEUE_H
 #define __KERNEL_WAIT_QUEUE_H
 
+#include <tee_api_types.h>
 #include <types_ext.h>
 #include <sys/queue.h>
 
@@ -44,9 +45,10 @@ static inline void wq_wait_init(struct wait_queue *wq,
 	wq_wait_init_condvar(wq, wqe, NULL, wait_read);
 }
 
-/* Waits for the wait queue element to the awakened. */
-void wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
-		   const void *sync_obj, const char *fname, int lineno);
+/* Waits for the wait queue element to be awakened or timed out */
+TEE_Result wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
+			 uint32_t timeout_ms, const void *sync_obj,
+			 const char *fname, int lineno);
 
 /* Wakes up the first wait queue element in the wait queue, if there is one */
 void wq_wake_next(struct wait_queue *wq, const void *sync_obj,
--- a/core/kernel/mutex.c
+++ b/core/kernel/mutex.c
@@ -60,7 +60,7 @@ static void __mutex_lock(struct mutex *m, const char *fname, int lineno)
 			 * Someone else is holding the lock, wait in normal
 			 * world for the lock to become available.
 			 */
-			wq_wait_final(&m->wq, &wqe, m, fname, lineno);
+			wq_wait_final(&m->wq, &wqe, 0, m, fname, lineno);
 		} else
 			return;
 	}
@@ -205,7 +205,7 @@ static void __mutex_read_lock(struct mutex *m, const char *fname, int lineno)
 			 * Someone else is holding the lock, wait in normal
 			 * world for the lock to become available.
 			 */
-			wq_wait_final(&m->wq, &wqe, m, fname, lineno);
+			wq_wait_final(&m->wq, &wqe, 0, m, fname, lineno);
 		} else
 			return;
 	}
@@ -393,13 +393,15 @@ void condvar_broadcast(struct condvar *cv)
 }
 #endif /*CFG_MUTEX_DEBUG*/
 
-static void __condvar_wait(struct condvar *cv, struct mutex *m,
-			const char *fname, int lineno)
+static TEE_Result __condvar_wait_timeout(struct condvar *cv, struct mutex *m,
+					 uint32_t timeout_ms, const char *fname,
+					 int lineno)
 {
-	uint32_t old_itr_status;
-	struct wait_queue_elem wqe;
-	short old_state;
-	short new_state;
+	TEE_Result res = TEE_SUCCESS;
+	uint32_t old_itr_status = 0;
+	struct wait_queue_elem wqe = { };
+	short old_state = 0;
+	short new_state = 0;
 
 	mutex_unlock_check(m);
 
@@ -434,23 +436,38 @@ static void __condvar_wait(struct condvar *cv, struct mutex *m,
 	if (!new_state)
 		wq_wake_next(&m->wq, m, fname, lineno);
 
-	wq_wait_final(&m->wq, &wqe, m, fname, lineno);
+	res = wq_wait_final(&m->wq, &wqe, timeout_ms, m, fname, lineno);
 
 	if (old_state > 0)
 		mutex_read_lock(m);
 	else
 		mutex_lock(m);
+
+	return res;
 }
 
 #ifdef CFG_MUTEX_DEBUG
 void condvar_wait_debug(struct condvar *cv, struct mutex *m,
 			const char *fname, int lineno)
 {
-	__condvar_wait(cv, m, fname, lineno);
+	__condvar_wait_timeout(cv, m, 0, fname, lineno);
+}
+
+TEE_Result condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m,
+				      uint32_t timeout_ms, const char *fname,
+				      int lineno)
+{
+	return __condvar_wait_timeout(cv, m, timeout_ms, fname, lineno);
 }
 #else
 void condvar_wait(struct condvar *cv, struct mutex *m)
 {
-	__condvar_wait(cv, m, NULL, -1);
+	__condvar_wait_timeout(cv, m, 0, NULL, -1);
+}
+
+TEE_Result condvar_wait_timeout(struct condvar *cv, struct mutex *m,
+				uint32_t timeout_ms)
+{
+	return __condvar_wait_timeout(cv, m, timeout_ms, NULL, -1);
 }
 #endif
--- a/core/kernel/notif.c
+++ b/core/kernel/notif.c
@@ -117,19 +117,25 @@ out:
 }
 #endif /*CFG_CORE_ASYNC_NOTIF*/
 
-static TEE_Result notif_rpc(uint32_t func, uint32_t value)
+static TEE_Result notif_rpc(uint32_t func, uint32_t value1, uint32_t value2)
 {
-	struct thread_param params = THREAD_PARAM_VALUE(IN, func, value, 0);
+	struct thread_param params =
+		THREAD_PARAM_VALUE(IN, func, value1, value2);
 
 	return thread_rpc_cmd(OPTEE_RPC_CMD_NOTIFICATION, 1, &params);
 }
 
 TEE_Result notif_wait(uint32_t value)
 {
-	return notif_rpc(OPTEE_RPC_NOTIFICATION_WAIT, value);
+	return notif_rpc(OPTEE_RPC_NOTIFICATION_WAIT, value, 0);
 }
 
 TEE_Result notif_send_sync(uint32_t value)
 {
-	return notif_rpc(OPTEE_RPC_NOTIFICATION_SEND, value);
+	return notif_rpc(OPTEE_RPC_NOTIFICATION_SEND, value, 0);
+}
+
+TEE_Result notif_wait_timeout(uint32_t value, uint32_t timeout_ms)
+{
+	return notif_rpc(OPTEE_RPC_NOTIFICATION_WAIT, value, timeout_ms);
 }
--- a/core/kernel/wait_queue.c
+++ b/core/kernel/wait_queue.c
@@ -20,10 +20,10 @@ void wq_init(struct wait_queue *wq)
 	*wq = (struct wait_queue)WAIT_QUEUE_INITIALIZER;
 }
 
-static void do_notif(TEE_Result (*fn)(uint32_t), int id,
-		     const char *cmd_str __maybe_unused,
-		     const void *sync_obj __maybe_unused,
-		     const char *fname, int lineno __maybe_unused)
+static TEE_Result do_notif(bool wait, int id, uint32_t timeout_ms,
+			   const char *cmd_str __maybe_unused,
+			   const void *sync_obj __maybe_unused,
+			   const char *fname, int lineno __maybe_unused)
 {
 	TEE_Result res = TEE_SUCCESS;
 
@@ -33,9 +33,15 @@ static void do_notif(TEE_Result (*fn)(uint32_t), int id,
 	else
 		DMSG("%s thread %d %p", cmd_str, id, sync_obj);
 
-	res = fn(id + NOTIF_SYNC_VALUE_BASE);
+	if (wait)
+		res = notif_wait_timeout(id + NOTIF_SYNC_VALUE_BASE,
+					 timeout_ms);
+	else
+		res = notif_send_sync(id + NOTIF_SYNC_VALUE_BASE);
 	if (res)
 		DMSG("%s thread %d res %#"PRIx32, cmd_str, id, res);
+
+	return res;
 }
 
 static void slist_add_tail(struct wait_queue *wq, struct wait_queue_elem *wqe)
@@ -69,24 +75,33 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
 	cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);
 }
 
-void wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
-		   const void *sync_obj, const char *fname, int lineno)
+static TEE_Result wq_wait_final_helper(struct wait_queue *wq,
+				       struct wait_queue_elem *wqe,
+				       uint32_t timeout_ms,
+				       const void *sync_obj, const char *fname,
+				       int lineno)
 {
-	uint32_t old_itr_status;
-	unsigned done;
+	uint32_t old_itr_status = 0;
 
-	do {
-		do_notif(notif_wait, wqe->handle,
-			 "sleep", sync_obj, fname, lineno);
+	do_notif(true, wqe->handle, timeout_ms, "sleep", sync_obj, fname,
+		 lineno);
 
-		old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
+	old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
+	SLIST_REMOVE(wq, wqe, wait_queue_elem, link);
+	cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);
 
-		done = wqe->done;
-		if (done)
-			SLIST_REMOVE(wq, wqe, wait_queue_elem, link);
+	if (wqe->done)
+		return TEE_SUCCESS;
+	else
+		return TEE_ERROR_TIMEOUT;
+}
 
-		cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);
-	} while (!done);
+TEE_Result wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
+			 uint32_t timeout_ms, const void *sync_obj,
+			 const char *fname, int lineno)
+{
+	return wq_wait_final_helper(wq, wqe, timeout_ms, sync_obj, fname,
+				    lineno);
 }
 
 void wq_wake_next(struct wait_queue *wq, const void *sync_obj,
@@ -130,7 +145,7 @@ void wq_wake_next(struct wait_queue *wq, const void *sync_obj,
 		cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);
 
 		if (do_wakeup)
-			do_notif(notif_send_sync, handle,
+			do_notif(false, handle, 0,
 				 "wake ", sync_obj, fname, lineno);
 
 		if (!do_wakeup || !wake_read)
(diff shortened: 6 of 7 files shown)


core/include/kernel/notif.h
```diff
@@ -124,6 +129,13 @@ static inline void notif_send_async(uint32_t value __unused)
  */
 TEE_Result notif_send_sync(uint32_t value);
 
+/*
+ * Send a synchronous value with timeout, note that it must be <=
+ * NOTIF_VALUE_MAX. The notification is synchronous even if the value
+ * happens to belong in the asynchronous range.
+ */
+TEE_Result notif_send_sync_timeout(uint32_t value, uint32_t tmo);
```
jenswi-linaro, line 137: Why is this function needed? What does it do that `notif_send_sync()` doesn't?

core/kernel/wait_queue.c
```diff
@@ -69,24 +74,45 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
...
+	uint32_t ret;
 
 	do {
-		do_notif(notif_wait, wqe->handle,
+		ret = do_notif(notif_wait_timeout, wqe->handle, tmo,
 			 "sleep", sync_obj, fname, lineno);
 
 		old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
 
+		if (ret == TEE_ERROR_BUSY) /* timed-out */
```
jenswi-linaro, line 92: Wouldn't `TEE_ERROR_TIMEOUT` be a better match?

core/kernel/notif.c
```diff
@@ -117,19 +117,30 @@ void notif_deliver_event(enum notif_event ev)
...
-static TEE_Result notif_rpc(uint32_t func, uint32_t value)
+static TEE_Result notif_rpc(uint32_t func, uint32_t value, uint32_t tmo)
 {
-	struct thread_param params = THREAD_PARAM_VALUE(IN, func, value, 0);
+	struct thread_param params = THREAD_PARAM_VALUE(IN, func, value, tmo);
 
 	return thread_rpc_cmd(OPTEE_RPC_CMD_NOTIFICATION, 1, &params);
 }
 
+
```
jenswi-linaro, line 127: Extra line not needed.

core/kernel/notif.c
```diff
@@ -117,19 +117,30 @@ void notif_deliver_event(enum notif_event ev)
 }
 #endif /*CFG_CORE_ASYNC_NOTIF*/
 
-static TEE_Result notif_rpc(uint32_t func, uint32_t value)
+static TEE_Result notif_rpc(uint32_t func, uint32_t value, uint32_t tmo)
```
jenswi-linaro, line 120: How about
``` C
static TEE_Result notif_rpc(uint32_t func, uint32_t value1, uint32_t value2)
```
to try to be more generic?

jenswi-linaro (commented):
What is the use case for this timeout? The synchronous notifications are used to implement mutex and condvar synchronization.

core/include/kernel/notif.h
ZheTing815: Perhaps we can simply use `notif_send_sync()` with tmo(timeout).

core/kernel/wait_queue.c
ZheTing815: Ok, I didn't notice this code, it would be more suitable for this scenario.
I will change to use `TEE_ERROR_TIMEOUT` in op-tee and linux kernel

core/kernel/notif.c
ZheTing815: Ok, thanks!

core/kernel/notif.c
ZheTing815: Ok, I will update for this.

ZheTing815:
> What is the use case for this timeout? The synchronous notifications are used to implement mutex and condvar synchronization.

I just updated a condvar with this timeout patch for reviewing.

ZheTing815:
WIP. wait for review

core/include/kernel/notif.h
jenswi-linaro: The only purpose of this function is to match the function pointer argument for `do_notif()` in `wq_wake_next()`. That is better done with an internal wrapper function just above `wq_wake_next()` that discards the timeout argument and calls `notif_send_sync()`.

ZheTing815: Ok, I will try to modify.

ZheTing815: I'm not quite sure what ```discards the timeout argument```  means. We have changed the number of parameters for ```do_notify()```, so the timeout argument is necessary, right?
In my opinion, we can expand the parameters of ```notif_send_sync()``` and ```notif_wait()``` by adding a timeout argument, so we don't need to add ```notif_send_sync_timeout()``` and ```notif_wait_timeout()``` separately.

jenswi-linaro: I tried what I suggested, but it got a bit more complicated with the function pointer than I liked. So I suggest doing something like this instead:
``` diff
diff --git a/core/kernel/wait_queue.c b/core/kernel/wait_queue.c
index 670e7c2ba263..373f7a2bc4b3 100644
--- a/core/kernel/wait_queue.c
+++ b/core/kernel/wait_queue.c
@@ -20,10 +20,10 @@ void wq_init(struct wait_queue *wq)
 	*wq = (struct wait_queue)WAIT_QUEUE_INITIALIZER;
 }
 
-static uint32_t do_notif(TEE_Result (*fn)(uint32_t, uint32_t), int id, uint32_t tmo,
-		     const char *cmd_str __maybe_unused,
-		     const void *sync_obj __maybe_unused,
-		     const char *fname, int lineno __maybe_unused)
+static TEE_Result do_notif(bool sleep, int id, uint32_t tmo,
+			   const char *cmd_str __maybe_unused,
+			   const void *sync_obj __maybe_unused,
+			   const char *fname, int lineno __maybe_unused)
 {
 	TEE_Result res = TEE_SUCCESS;
 
@@ -33,7 +33,10 @@ static uint32_t do_notif(TEE_Result (*fn)(uint32_t, uint32_t), int id, uint32_t
 	else
 		DMSG("%s thread %d %p", cmd_str, id, sync_obj);
 
-	res = fn(id + NOTIF_SYNC_VALUE_BASE, tmo);
+	if (sleep)
+		res = notif_wait_timeout(id + NOTIF_SYNC_VALUE_BASE, tmo);
+	else
+		res = notif_send_sync(id + NOTIF_SYNC_VALUE_BASE);
 	if (res)
 		DMSG("%s thread %d res %#"PRIx32, cmd_str, id, res);
 
@@ -81,8 +84,8 @@ static uint32_t wq_wait_final_helper(struct wait_queue *wq,
 	uint32_t ret;
 
 	do {
-		ret = do_notif(notif_wait_timeout, wqe->handle, tmo,
-			 "sleep", sync_obj, fname, lineno);
+		ret = do_notif(true /*sleep*/, wqe->handle, tmo,
+			       "sleep", sync_obj, fname, lineno);
 
 		old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
 
@@ -153,8 +156,8 @@ void wq_wake_next(struct wait_queue *wq, const void *sync_obj,
 		cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);
 
 		if (do_wakeup)
-			do_notif(notif_send_sync_timeout, handle, 0,
-				 "wake ", sync_obj, fname, lineno);
+			do_notif(false /*!sleep*/, handle, 0, "wake ",
+				 sync_obj, fname, lineno);
 
 		if (!do_wakeup || !wake_read)
 			break;
```
with this, there's no need for `notif_send_sync_timeout()`

ZheTing815: Thank you for your suggestion. I have modified the new patch based on it.

jenswi-linaro:
Please fix the checkpatch issues.

ZheTing815:
> Please fix the checkpatch issues.

Ok, I just fixed the checkpatch issues.

core/kernel/wait_queue.c
```diff
@@ -20,10 +20,10 @@ void wq_init(struct wait_queue *wq)
 	*wq = (struct wait_queue)WAIT_QUEUE_INITIALIZER;
 }
 
-static void do_notif(TEE_Result (*fn)(uint32_t), int id,
-		     const char *cmd_str __maybe_unused,
-		     const void *sync_obj __maybe_unused,
-		     const char *fname, int lineno __maybe_unused)
+static uint32_t do_notif(bool sleep, int id, uint32_t timeout_ms,
```
jenswi-linaro, line 23: Please use `TEE_Result` as return type.

core/kernel/wait_queue.c
```diff
@@ -69,24 +75,47 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
 	cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);
 }
 
-void wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
-		   const void *sync_obj, const char *fname, int lineno)
+static uint32_t wq_wait_final_helper(struct wait_queue *wq,
```
jenswi-linaro, line 78: Please use `TEE_Result` as return type.

core/kernel/wait_queue.c
```diff
@@ -69,24 +75,47 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
...
+				     struct wait_queue_elem *wqe,
+				     uint32_t timeout_ms, const void *sync_obj,
+				     const char *fname, int lineno)
 {
 	uint32_t old_itr_status;
 	unsigned done;
+	uint32_t ret;
```
jenswi-linaro, line 85: `TEE_Result`
Please initialize all the local variables in this function since you're adding a local variable.

core/kernel/wait_queue.c
```diff
@@ -69,24 +75,47 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
...
+	return ret;
+}
+
+void wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
+		   const void *sync_obj, const char *fname, int lineno)
+{
+	wq_wait_final_helper(wq, wqe, 0, sync_obj, fname, lineno);
+}
+
+uint32_t wq_wait_final_timeout(struct wait_queue *wq,
```
jenswi-linaro, line 112: Please use `TEE_Result` as return type.

core/include/kernel/mutex.h
```diff
@@ -95,10 +95,19 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
...
+				    uint32_t timeout_ms, const char *fname,
+				    int lineno);
+#define condvar_wait_timeout(cv, m, timeout_ms) \
+			     condvar_wait_timeout_debug((cv), (m), timeout_ms \
+			     __FILE__, __LINE__)
 #else
 void condvar_signal(struct condvar *cv);
 void condvar_broadcast(struct condvar *cv);
 void condvar_wait(struct condvar *cv, struct mutex *m);
+uint32_t condvar_wait_timeout(struct condvar *cv, struct mutex *m,
```
jenswi-linaro, line 109: Please use `TEE_Result` as return type.

core/kernel/mutex.c
```diff
@@ -393,13 +393,15 @@ void condvar_broadcast(struct condvar *cv)
 }
 #endif /*CFG_MUTEX_DEBUG*/
 
-static void __condvar_wait(struct condvar *cv, struct mutex *m,
-			const char *fname, int lineno)
+static uint32_t __condvar_wait_timeout(struct condvar *cv, struct mutex *m,
```
jenswi-linaro, line 396: Please use `TEE_Result` as return type.

core/kernel/mutex.c
```diff
@@ -393,13 +393,15 @@ void condvar_broadcast(struct condvar *cv)
...
+				       uint32_t timeout_ms, const char *fname,
+				       int lineno)
 {
 	uint32_t old_itr_status;
 	struct wait_queue_elem wqe;
 	short old_state;
 	short new_state;
+	uint32_t ret = 0;
```
jenswi-linaro, line 404: `TEE_Result`
Please initialize all the local variables since you're adding a local variable.

core/kernel/wait_queue.c
ZheTing815: Ok.

core/kernel/wait_queue.c
ZheTing815: Ok.

core/kernel/wait_queue.c
ZheTing815: Ok

core/include/kernel/mutex.h
ZheTing815: Ok

ZheTing815:
Push new patch to initialize all the local variables and change return type.

etienne-lms (commented):
Could you update also _core/include/optee_rpc_cmd.h_ for the new ABI, something like:
```patch
 /*
  * Notification from/to secure world.
  *
  * If secure world needs to wait for something, for instance a mutex, it
  * does a notification wait request instead of spinning in secure world.
  * Conversely a synchronous notification can be sent when a secure
  * world mutex with a thread waiting thread is unlocked.
  *
  * This interface can also be used to wait for a asynchronous notification
  * which instead is sent via a non-secure interrupt.
  *
  * Waiting on notification
  * [in]    value[0].a	    OPTEE_RPC_NOTIFICATION_WAIT
  * [in]    value[0].b	    notification value
+ * [in]    value[0].c	    timeout in millisecond or 0 if no tiemout
  *
  * Sending a synchronous notification
  * [in]    value[0].a	    OPTEE_RPC_NOTIFICATION_SEND
  * [in]    value[0].b	    notification value
+ * [in]    value[0].c	    shall be 0, reserved for future evolution.
  */
```

core/kernel/wait_queue.c
```diff
@@ -69,24 +75,48 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
...
 	do {
-		do_notif(notif_wait, wqe->handle,
-			 "sleep", sync_obj, fname, lineno);
+		res = do_notif(true, wqe->handle, timeout_ms,
+			       "sleep", sync_obj, fname, lineno);
 
 		old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
 
+		if (res == TEE_ERROR_TIMEOUT) /* timed-out */
+			wqe->done = true;
```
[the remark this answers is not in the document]
ZheTing815: Yes, thanks for your kind reminder. Kernel returning either ```TEEC_SUCCESS``` or ```TEEC_ERROR_TIMEOUT``` is the correct. ```TEE_SUCCESS``` should be checked, too.

ZheTing815:
@etienne-lms Thank for the review.
The updated patch fixes the return handle and adds comments in _core/include/optee_rpc_cmd.h_.

core/kernel/wait_queue.c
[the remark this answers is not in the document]
jenswi-linaro: If `wqe->done` is `false` then `TEE_ERROR_TIMEOUT` is expected.
If `wqe->done` is `true` then both `TEE_SUCCESS` and `TEE_ERROR_TIMEOUT` are OK, even if `TEE_ERROR_TIMEOUT` only can occur during a small window.

`wqe->done` was previously used to detect a spurious wakeup in this loop. It's at the same time used in `wq_wake_next()` to see that we only wake up one.

Would the following make sense for better control over the return value:
``` C
if (wqe->done)
    res = TEE_SUCCESS;
else
    res = TEE_ERROR_TIMEOUT;
wqe->done = true;
```
We don't need to care about the return value from the RPC, `wqe->done` is a more accurate indicator of whether it was a timeout.

core/include/kernel/mutex.h
```diff
@@ -95,10 +95,19 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
 void condvar_wait_debug(struct condvar *cv, struct mutex *m,
 			const char *fname, int lineno);
 #define condvar_wait(cv, m) condvar_wait_debug((cv), (m), __FILE__, __LINE__)
+
+TEE_Result condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m,
+				      uint32_t timeout_ms, const char *fname,
+				      int lineno);
+#define condvar_wait_timeout(cv, m, timeout_ms) \
+			     condvar_wait_timeout_debug((cv), (m), timeout_ms \
```
etienne-lms, line 103: Coma missing after `timeout_ms`.
Add parentheses around `timeout_ms`.

core/kernel/wait_queue.c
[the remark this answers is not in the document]
ZheTing815: Thank you for the suggestions.

core/kernel/wait_queue.c
```diff
@@ -69,24 +75,51 @@ void wq_wait_init_condvar(struct wait_queue *wq, struct wait_queue_elem *wqe,
...
+		res = do_notif(true, wqe->handle, timeout_ms,
+			       "sleep", sync_obj, fname, lineno);
 
 		old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
 
+		if (wqe->done)
+			res = TEE_SUCCESS;
+		else
+			res = TEE_ERROR_TIMEOUT;
+		wqe->done = true;
```
etienne-lms, line 98: Do this really work? One waiting for a event may be woken but it's resource not yet ready hence going back for another roudnin this loop. Something looks odd to me.

etienne-lms: Reading back https://github.com/OP-TEE/optee_os/pull/6641#discussion_r1548092078 I think I miss understood `wqe->done` value. 

If the sequence here is fine, then I would suggest to simplify this:
```c
{
	uint32_t old_itr_status = 0;

	do_notif(true, wqe->handle, timeout_ms, "sleep", sync_obj, fname,
		 lineno);

	old_itr_status = cpu_spin_lock_xsave(&wq_spin_lock);
	SLIST_REMOVE(wq, wqe, wait_queue_elem, link);
	cpu_spin_unlock_xrestore(&wq_spin_lock, old_itr_status);

	if (wqe->done)
		return TEE_SUCCESS;
	else
		return TEE_ERROR_TIMEOUT;
}
```
I even think `wq_wait_final_helper()` does not need a return value. Caller could check `wqe->done` to know whether the notif waiting process timed out or not. That said, an explicit return value may be convenient.

jenswi-linaro: I was about to suggest the same. I think we can simplify further. We don't need a separate `wq_wait_final_timeout()`, we could add the `timeout_ms` parameter and return value to `wq_wait_final()`.

ZheTing815: > I even think `wq_wait_final_helper()` does not need a return value. Caller could check `wqe->done` to know whether the notif waiting process timed out or not. That said, an explicit return value may be convenient.

The return value of ```wq_wait_final_helper``` will propagated up to ```condvar_wait_timeout```.
The caller can check return the value from ```condvar_wait_timeout```. However, since the caller cannot get ```wqe```, they cannot check ```wqe->done```. Therefore, the return value of ```wq_wait_final_helper``` is required. Is my understanding correct?

ZheTing815: > I was about to suggest the same. I think we can simplify further. We don't need a separate `wq_wait_final_timeout()`, we could add the `timeout_ms` parameter and return value to `wq_wait_final()`.

Ok, I can extend `wq_wait_final()` with parameter `timeout_ms`.

core/kernel/wait_queue.c
```diff
@@ -20,10 +20,10 @@ void wq_init(struct wait_queue *wq)
 	*wq = (struct wait_queue)WAIT_QUEUE_INITIALIZER;
 }
 
-static void do_notif(TEE_Result (*fn)(uint32_t), int id,
-		     const char *cmd_str __maybe_unused,
-		     const void *sync_obj __maybe_unused,
-		     const char *fname, int lineno __maybe_unused)
+static TEE_Result do_notif(bool sleep, int id, uint32_t timeout_ms,
```
etienne-lms, line 23: I think argument `sleep` should be renamed 'wait` since it is used to differentiate between a "notif-wait" request (with or without a timeout value that is set from `timeout_ms` value) and a "notif-send".

core/include/kernel/mutex.h
```diff
@@ -95,10 +95,19 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
 void condvar_wait_debug(struct condvar *cv, struct mutex *m,
 			const char *fname, int lineno);
 #define condvar_wait(cv, m) condvar_wait_debug((cv), (m), __FILE__, __LINE__)
+
+TEE_Result condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m,
+				      uint32_t timeout_ms, const char *fname,
+				      int lineno);
+#define condvar_wait_timeout(cv, m, timeout_ms) \
+			     condvar_wait_timeout_debug((cv), (m), \
+			     (timeout_ms), __FILE__, __LINE__)
```
etienne-lms, line 104: For indentation consistency, prefer:
```c
#define condvar_wait_timeout(cv, m, timeout_ms) \
	condvar_wait_timeout_debug((cv), (m), (timeout_ms), __FILE__, __LINE__)

File: core/include/kernel/mutex.h [hunk 17]
```diff
@@ -95,10 +95,19 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
...
 #else
 void condvar_signal(struct condvar *cv);
 void condvar_broadcast(struct condvar *cv);
 void condvar_wait(struct condvar *cv, struct mutex *m);
+TEE_Result condvar_wait_timeout(struct condvar *cv, struct mutex *m,
```
etienne-lms, line 109: I suggest to add an inline comment stating something like:
`/* Always return TEE_SUCCESS when @timeout_ms is 0 */`
It would clarify why some callers of to `wq_wait_final()` do drop its return value.

Ditto for `condvar_wait_timeout_debug()`  at line 99.

etienne-lms (commented):
Look good to me with some minor comments. It's nice this feature does not break existing transport ABI.

core/include/kernel/mutex.h
ZheTing815: Ok, thank you for the suggestion.

core/include/kernel/mutex.h
ZheTing815: Ok.

core/kernel/wait_queue.c
ZheTing815: Ok, I changed to use 'wait'.

core/include/kernel/mutex.h
```diff
@@ -95,10 +95,20 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
...
+TEE_Result condvar_wait_timeout_debug(struct condvar *cv, struct mutex *m,
+				      uint32_t timeout_ms, const char *fname,
+				      int lineno);
+#define condvar_wait_timeout(cv, m, timeout_ms) \
+	condvar_wait_timeout_debug((cv), (m), (timeout_ms), __FILE__, __LINE__)
 #else
 void condvar_signal(struct condvar *cv);
 void condvar_broadcast(struct condvar *cv);
 void condvar_wait(struct condvar *cv, struct mutex *m);
+/* Always return TEE_SUCCESS when @timeout_ms is 0 */
```
jenswi-linaro, line 109: I'm sorry, but this isn't true.

It will return `TEE_ERROR_TIMEOUT` If the normal world returns before the condvar has been signaled, regardless of the `timeout_ms` parameter. The normal world is not expected to return early, but we don't trust the normal world so it may very well happen.

This is not as bad as it sounds since callers  of `condvar_wait()` and `condvar_wait_timeout()` should something like the following pattern:
``` C
while (some_condition)
    condvar_wait(...);
```
Caring about the return value is optional since it's the evaluation of `some_condition` that matters. With `condvar_wait_timeout()` you have the option of possibly error out early, but if you care about passed time you need to check the secure time and not trust the normal world.

ZheTing815: > It will return `TEE_ERROR_TIMEOUT` If the normal world returns before the condvar has been signaled, regardless of the `timeout_ms` parameter. The normal world is not expected to return early, but we don't trust the normal world so it may very well happen.

Oh, that's true. I didn't consider the case `The normal world returns early`. I will remove the comment.

etienne-lms: Thanks @jenswi-linaro for the feedback.
I think however that a bit of description of how the function behaves and what it returns would help readers/developers.

jenswi-linaro: Agreed.

ZheTing815: Agreed, too.
Could I add the ` Return TEE_ERROR_TIMEOUT If the normal world returns before the condvar has been signaled` as the inline comment for these two functions?

jenswi-linaro: Yeah, that's a good clarification.

jenswi-linaro:

core/include/kernel/wait_queue.h
```diff
@@ -44,9 +45,10 @@ static inline void wq_wait_init(struct wait_queue *wq,
 	wq_wait_init_condvar(wq, wqe, NULL, wait_read);
 }
 
-/* Waits for the wait queue element to the awakened. */
-void wq_wait_final(struct wait_queue *wq, struct wait_queue_elem *wqe,
-		   const void *sync_obj, const char *fname, int lineno);
+/* Waits for the wait queue element to the awakened or timed out */
```
etienne-lms, line 48: s/the/be/

core/include/kernel/mutex.h
```diff
@@ -95,10 +95,26 @@ void condvar_broadcast_debug(struct condvar *cv, const char *fname, int lineno);
 void condvar_wait_debug(struct condvar *cv, struct mutex *m,
 			const char *fname, int lineno);
 #define condvar_wait(cv, m) condvar_wait_debug((cv), (m), __FILE__, __LINE__)
+
+/*
+ * Return TEE_ERROR_TIMEOUT If the normal world returns before
```
etienne-lms, line 100: **s/If/if/** + add a period `.` at the end of the sentence (next line).

Ditto at lines 113/114.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with typos fixed.

core/include/kernel/wait_queue.h
ZheTing815: Updated.

core/include/kernel/mutex.h
ZheTing815: Updated.

PR #6642, opened 2024-01-25
plat-stm32mp2: update to support larger DDR densities and embed tests

GseoC:
Allow up to 8 Gbytes of DDR and embed tests.


etienne-lms (commented):

GseoC:
Tags applied

PR #6644, opened 2024-01-26
drivers: crypto: stm32: fix SAES key selection

etienne-lms:
Correction selection of key in STM32 SAES driver that missed a left bit shift operation. The bug was not experienced before as current platform tests involve only the software key selection (_SAES_CR_KEYSEL_SOFT) which value is 0 and matches the SoC default key selection register value.

Fixes: 4320f5cf30c5 ("crypto: stm32: SAES cipher support")


etienne-lms:
@tbourgoi, could you have a look at this change?

tbourgoi:
Nice catch
you can add my tags
`Acked-by: Thomas Bourgoin <thomas.bourgoin@foss.st.com>`

etienne-lms:
review tag applied.

Issue #6631, opened 2024-01-23
OP-TEE 4.1.0 fails to boot on zynqmp(ZCU102)


callmefriend:
I am compiling OPTEE, my compilation command is as follows:
```
$ repo init -u https://github.com/OP-TEE/manifest.git -m zynqmp.xml 
$ repo sync
$ cd build
$ make toolchains
$ make PLATFORM=zynqmp-zcu102 image
```
After the compilation is over, get the following 2 files in Zynqmp/Directory:

BOOT.bin zynqmp-zcu102.ub

Then I replaced the corresponding files in 2021.1-zcu102-release.tar.xz with the above two files, then copied it to the SD card, and finally use the SD card to boot ZCU102.

Got the following error:
```
Xilinx Zynq MP First Stage Boot Loader 
Release 2021.1   Jun  6 2021  -  07:07:32
NOTICE:  BL31: Secure code at 0x60000000
NOTICE:  BL31: Non secure code at 0x8000000
NOTICE:  BL31: v2.8(release):xilinx-v2023.2
NOTICE:  BL31: Built : 10:44:01, Jan 23 2024
INFO:    ARM GICv2 driver initialized
INFO:    BL31: Initializing runtime services
INFO:    BL31: PM Service Init Complete: API v1.1
INFO:    BL31: Initializing BL32
D/TC:0   add_phys_mem:666 VCORE_UNPG_RX_PA type TEE_RAM_RX 0x60000000 size 0x00081000
D/TC:0   add_phys_mem:666 VCORE_UNPG_RW_PA type TEE_RAM_RW 0x60081000 size 0x0017f000
D/TC:0   add_phys_mem:666 ta_base type TA_RAM 0x60200000 size 0x0fe00000
D/TC:0   add_phys_mem:666 ROUNDDOWN(GIC_BASE + GICD_OFFSET, CORE_MMU_PGDIR_SIZE) type IO_SEC 0xf9000000 size 0x00200000
D/TC:0   add_phys_mem:666 ROUNDDOWN(GIC_BASE, CORE_MMU_PGDIR_SIZE) type IO_SEC 0xf9000000 size 0x00200000
D/TC:0   add_phys_mem:680 Physical mem map overlaps 0xf9000000
D/TC:0   add_phys_mem:666 ROUNDDOWN(CONSOLE_UART_BASE, CORE_MMU_PGDIR_SIZE) type IO_SEC 0xff000000 size 0x00200000
[...]

PR #6645, opened 2024-01-27
core: arm: runtime check for CE

igoropaniuk:
Add runtime check during boot for Crypto Extensions if CFG_CRYPTO_WITH_CE=y.


core/arch/arm/include/arm64.h
```diff
@@ -366,6 +366,17 @@ static inline __noprof void tlbi_vale1is(uint64_t va)
 	asm volatile ("tlbi	vale1is, %0" : : "r" (va));
 }
 
+static inline __noprof uint32_t read_isar5(void)
+{
+	uint32_t isar_reg;
+#if defined(__clang__)
+	asm volatile("mrs %w0, ID_ISAR5_EL1" : "=r"(isar_reg));
```
igoropaniuk, line 373: Added this hack for clang to suppress the error below. Looks like it's a false positive, as `ID_ISAR5_EL1` is 32bit register:
```
core/arch/arm/include/arm64.h:373:45: error: value size does not match register size specified by the constraint and modifier [-Werror,-Wasm-operand-widths]
        asm volatile("mrs %0, ID_ISAR5_EL1" : "=r"(isar_reg));
                                                   ^
core/arch/arm/include/arm64.h:373:20: note: use constraint modifier "w"
        asm volatile("mrs %0, ID_ISAR5_EL1" : "=r"(isar_reg));
                          ^~
                          %w0
1 error generated.
```

and gcc doesn't support that syntax (which works for clang), reporting:
```
{standard input}: Assembler messages:
{standard input}:1822: Error: operand mismatch -- `mrs w0,ID_ISAR5_EL1'
{standard input}:1822: Info:    did you mean this?
{standard input}:1822: Info:    	mrs x0, id_isar5_el1

```

Let me know if you have a better idea how to address that

jforissier:
"core: arm: add masks for ID_ISAR5 fields":

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,29 @@ static void init_vfp_nsec(void)
...
+	if (!(isar5 | ID_ISAR5_AES))
+		return false;
+
+	if (!(isar5 | ID_ISAR5_SHA1))
+		return false;
+
+	if (!(isar5 | ID_ISAR5_SHA2))
+		return false;
+
+	if (!(isar5 | ID_ISAR5_CRC32))
```
jenswi-linaro, line 208: Do we need to care about this one?

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,29 @@ static void init_vfp_nsec(void)
...
 
+/*
+ * Check for supported Crypto Extensions (ARMv8 aarch32/aarch64)
+ * In case one of instructions is not supported false is returned.
+ */
+static bool check_cpuid_ce(void)
+{
+	uint32_t isar5 = read_isar5();
+
+	if (!(isar5 | ID_ISAR5_AES))
```
jenswi-linaro, line 199: `isar5 & ID_ISAR5_AES`?
How did you test this?

igoropaniuk, line 199: maybe also makes sense to additionally check for algo specific flag like `CFG_CRYPTO_AES_ARM_CE` in each if statement, however I'm not sure if it's the case when one algo from crypto extensions is not available, but others are.

jforissier (commented):
Comments on "core: arm: add helper for reading isar5".

core/arch/arm/kernel/boot.c
igoropaniuk: well, kernel does it also https://github.com/torvalds/linux/blob/master/arch/arm/kernel/setup.c#L480, though CRC32 is not a part of CE

core/arch/arm/kernel/boot.c
igoropaniuk: oops, I messed up. thanks!

I was testing in QEMUv8 with CE support enabled (current default setup), however didn't bother to rebuild with crypto disabled (not even sure if it's possible to do that from configure script, maybe manually disabling `CONFIG_ARM_AES_BUILTIN` etc in source tree)

core/arch/arm/kernel/boot.c
igoropaniuk: @jenswi-linaro so what you prefer: delete it or keep it?

core/arch/arm/include/arm32.h
```diff
@@ -276,6 +276,14 @@ static __always_inline __noprof uint32_t read_r7(void)
 	return val;
 }
 
+static __always_inline __noprof uint32_t read_isar5(void)
```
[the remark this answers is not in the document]
igoropaniuk: fixed

core/arch/arm/include/arm64.h
```diff
@@ -366,6 +366,14 @@ static inline __noprof void tlbi_vale1is(uint64_t va)
 	asm volatile ("tlbi	vale1is, %0" : : "r" (va));
 }
 
+static inline __noprof uint32_t read_isar5(void)
```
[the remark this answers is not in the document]
igoropaniuk: fixed

jforissier (commented):
"lib: libutee: define read function for id_isar5" should be "core: arm64: define read function for id_isar5". With that, this commit is:

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,26 @@ static void init_vfp_nsec(void)
 }
 #endif
 
+/*
+ * Check for supported Crypto Extensions (ARMv8 aarch32/aarch64)
+ * In case one of instructions is not supported false is returned.
+ */
+static bool check_cpuid_ce(void)
+{
+	uint32_t isar5 = read_id_isar5();
```
jenswi-linaro, line 197: Shouldn't we use `ID_AA64ISAR0_EL1` when in AArch64? I guess we could even take the different configuration options
`CFG_CRYPTO_AES_ARM_CE`, `CFG_CRYPTO_SHA1_ARM_CE`, `CFG_CRYPTO_SHA256_ARM_CE`, `CFG_CRYPTO_SHA512_ARM_CE`, `CFG_CRYPTO_SHA3_ARM_CE`, `CFG_CRYPTO_SM3_ARM_CE`, `CFG_CRYPTO_SM4_ARM_CE`, and `CFG_CRYPTO_SM4_ARM_AESE` into consideration when testing in this function.

igoropaniuk: > Shouldn't we use ID_AA64ISAR0_EL1 when in AArch64

on one hand it would be correct, on the other hand I wanted to keep it as simple as possible and address the issue described in https://github.com/OP-TEE/optee_os/issues/6631, without adding any logic for differentiation of versions of ARMv8 CE. Btw, this is neither done in makescripts,  as enabling `CFG_CRYPTO_WITH_CE` implies `CFG_CRYPTO_*_ARM_CE` options (unless disabled explicitly), expecting the code to be running at least on Armv8.2-A core.

> I guess we could even take the different configuration options CFG_CRYPTO_AES_ARM_CE, CFG_CRYPTO_SHA1_ARM_CE, CFG_CRYPTO_SHA256_ARM_CE, CFG_CRYPTO_SHA512_ARM_CE, CFG_CRYPTO_SHA3_ARM_CE, CFG_CRYPTO_SM3_ARM_CE, CFG_CRYPTO_SM4_ARM_CE, and CFG_CRYPTO_SM4_ARM_AESE into consideration when testing in this function.

I agree, and ideally that would make more sense if we included both software and arm-ce implementations of these algos so the proper one can be chosen in runtime.

igoropaniuk: @jenswi-linaro added new implementation

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,61 @@ static void init_vfp_nsec(void)
 }
 #endif
 
+/*
+ * Check for supported Crypto Extensions (ARMv8 aarch32/aarch64)
+ * In case one of instructions is not supported false is returned.
```
etienne-lms, line 193: "... one of **the** instructions ..."

etienne-lms (commented):
Typo in commit message for "core: arm64: add masks for ID_AA64ISAR0_EL1 fields":
```patch
-RDM    -   [31:68]
+RDM    -   [31:28]
```

For commit "core: arm: kernel: add runtime check for CE": maybe explicitly say in the commit message for core will panic if configuration enables an Arm CE feature that the hardware does not support.

Aside the minor comments: 

core/arch/arm/kernel/boot.c
igoropaniuk: fixed, thanks!

igoropaniuk:
@jforissier @jenswi-linaro @etienne-lms addressed all comments, applied Etienne's A-b tag

etienne-lms (commented):
Some typos i've missed.

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,61 @@ static void init_vfp_nsec(void)
...
+	}
+
+	if (!feat_sha256_implemented() &&
+	    IS_ENABLED(CFG_CRYPTO_SHA256_ARM_CE)) {
+		EMSG("SHA256 instructions are not supported");
+		return false;
+	}
+
+	if (IS_ENABLED(CFG_ARM32_core))
+		return true;
```
jforissier, line 216: This test should not be needed, since none of the below `CFG_` flags should ever be set on arm32. I believe it is better to remove it as a double check that everything is OK configuration-wise.

core/arch/arm/kernel/boot.c
```diff
@@ -1148,6 +1203,13 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
 	thread_set_exceptions(THREAD_EXCP_ALL);
 	primary_save_cntfrq();
 	init_vfp_sec();
+
+	if (IS_ENABLED(CFG_CRYPTO_WITH_CE) && !check_cpuid_ce()) {
+		EMSG("OP-TEE is built with CRYPTO_WITH_CE=y");
+		EMSG("But some CE instructions are not supported by CPU");
+		panic();
+	}
```
jforissier, line 1211: How about making `check_cpuid_ce()` return void and `panic()` directly instead?

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,61 @@ static void init_vfp_nsec(void)
 }
 #endif
 
+/*
+ * Check for supported Crypto Extensions (ARMv8 aarch32/aarch64)
+ * In case one of instructions is not supported false is returned.
+ */
+static bool check_cpuid_ce(void)
```
jforissier, line 195: I suggest to call it `check_crypto_extensions()` and remove the comment.

jforissier (commented):
- For "core: arm64: add masks for ID_AA64ISAR0_EL1 fields" with Etienne's comment addressed:
- For "core: arm: add helper functions for checking CE support":

More comments below.

core/arch/arm/include/arm.h
```diff
@@ -181,10 +181,9 @@ static inline unsigned int feat_pan_implemented(void)
 static inline bool feat_crc32_implemented(void)
 {
 #ifdef ARM32
-	return false;
```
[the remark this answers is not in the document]
igoropaniuk: agree, will do

core/arch/arm/kernel/boot.c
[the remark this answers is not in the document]
igoropaniuk: I've dropped that comment as @jforissier suggested

core/arch/arm/include/arm.h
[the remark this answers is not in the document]
igoropaniuk: done

core/arch/arm/kernel/boot.c
```diff
@@ -1148,6 +1203,13 @@ static void init_primary(unsigned long pageable_part, unsigned long nsec_entry)
 	thread_set_exceptions(THREAD_EXCP_ALL);
 	primary_save_cntfrq();
 	init_vfp_sec();
+
+	if (IS_ENABLED(CFG_CRYPTO_WITH_CE) && !check_cpuid_ce()) {
+		EMSG("OP-TEE is built with CRYPTO_WITH_CE=y");
```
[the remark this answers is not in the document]
igoropaniuk: I've dropped that message

core/arch/arm/include/arm64.h
```diff
@@ -294,6 +294,22 @@
...
+#define ID_AA64ISAR0_SHA2	GENMASK_64(15, 12)
+#define ID_AA64ISAR0_CRC32	GENMASK_64(19, 16)
+#define ID_AA64ISAR0_ATOMIC	GENMASK_64(23, 20)
+#define ID_AA64ISAR0_TME	GENMASK_64(27, 24)
+#define ID_AA64ISAR0_RDM	GENMASK_64(31, 28)
+#define ID_AA64ISAR0_SHA3	GENMASK_64(35, 32)
+#define ID_AA64ISAR0_SM3	GENMASK_64(39, 36)
+#define ID_AA64ISAR0_SM4	GENMASK_64(43, 40)
+
+#define ID_AA64ISAR0_SHA2_SHIFT			U(12)
```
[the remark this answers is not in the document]
igoropaniuk: done, thanks

core/arch/arm/kernel/boot.c
```diff
@@ -188,6 +188,59 @@ static void init_vfp_nsec(void)
...
 
+static void check_crypto_extensions(void)
+{
+	bool ce_supported = true;
+
+	if (!feat_aes_implemented() &&
+	    IS_ENABLED(CFG_CRYPTO_AES_ARM_CE)) {
+		EMSG("AES instructions are not supported");
+		ce_supported = false;
+	}
```
jforissier, line 199: How about:
```C
	if (!feat_aes_implemented() &&
	    IS_ENABLED(CFG_CRYPTO_AES_ARM_CE))
		panic("AES instructions are not supported");
```
...and same below. In addition to making the code shorter it would print the exact feature that is missing (well, the first found at least).

jforissier (commented):
- Commit "core: arm: rewrite feat_crc32_implemented":
  Add parentheses after function name: `feat_crc32_implemented()` in subject and commit description.
  With that:
- Commit "core: arm64: remove ID_AA64ISAR0_EL1 macros":
- Commit "core: arm: add check in aarch32 for feat_crc32_implemented":
  Add missing parentheses, then:
- Commit "core: arm: add helper functions for checking CE support":

Plus one comment on the last commit. Thanks!

core/arch/arm/kernel/boot.c
igoropaniuk: That was my first thought, but then I decided why not to print all of them?
Easier to tackle the issue, as you report all unsupported instructions at once

jforissier: Fair enough. Good with me.

jforissier:
For "core: arm: kernel: add runtime check for CE":

igoropaniuk:
rebased onto the latest master

igoropaniuk:
@jenswi-linaro @jforissier @etienne-lms any additional comments/objections?

jforissier:
> @jenswi-linaro @jforissier @etienne-lms any additional comments/objections?

The Xen CI job has failed, I have restarted it but I suspect there might be an issue with the access to the config registers...

jforissier:
@etienne-lms I will give @jenswi-linaro the opportunity to comment further since he has reviewed already.

igoropaniuk:
@jenswi-linaro tag applies, thanks

jforissier:
Thanks @igoropaniuk!

PR #6652, opened 2024-01-29
core: arm: generic_ram_layout.h: add TEE_SCMI_SHMEM_*

tprrt:
To get TEE_SCMI_SHMEM_* defined from generic configuration directives.

### DIFF
--- a/core/arch/arm/include/mm/generic_ram_layout.h
+++ b/core/arch/arm/include/mm/generic_ram_layout.h
@@ -18,6 +18,9 @@
  * CFG_SHMEM_SIZE
  *
  * Optional directives:
+ * CFG_SCMI_SHMEM_START Sets the address of SCMI static shared memory.
+ * CFG_SCMI_SHMEM_SIZE  Sets the byte size of SCMI static shared memory.
+ *
  * CFG_TEE_LOAD_ADDR	If defined sets TEE_LOAD_ADDR. If not, TEE_LOAD_ADDR
  *			is set by the platform or defaults to TEE_RAM_START.
  * CFG_TEE_RAM_VA_SIZE	Some platforms may have specific needs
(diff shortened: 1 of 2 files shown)


etienne-lms (commented):
Hi @tprrt. 
I'm not sure this change is really applicable to the generic layout. There may be several pieces of shared memory used in the scope of SCMI services embedded in OP-TEE: an SCMI server may expose several SCMI channels, each being possibly related to an piece of shared memory.
What benefit do you see in supporting `CFG_SCMI_SHMEM_START/_SIZE` in _generic_ram_layout.h_?

tprrt:
Hello @etienne-lms 
> Hi @tprrt. I'm not sure this change is really applicable to the generic layout. There may be several pieces of shared memory used in the scope of SCMI services embedded in OP-TEE: an SCMI server may expose several SCMI channels, each being possibly related to an piece of shared memory. What benefit do you see in supporting `CFG_SCMI_SHMEM_START/_SIZE` in _generic_ram_layout.h_?

An SCMI server may expose several SCMI channels, but is the shared memory discontinuous?

Kind regards,
Thomas

etienne-lms:
> An SCMI server may expose several SCMI channels, but is the shared memory discontinuous?

They could be. Consider several Linux VMs in the non-secure world. They SCMI shared memory would likely be not contiguous.

By the way, if you use the DDR for SCMI message exchange, maybe consider using OP-TEE native shared memory, not a piece a DDR mapped as Device memory for the sole purpose of SCMI: this can be defined in Linux DT by using `compatible = "linaro,scmi-optee"` and no `shmem` property in the `scmi` node. In such case Linux SCMI driver will allocate a piece of OP-TEE generic shared memory for SCMI communications with the OP-TEE SCMI PTA.

PR #6661, opened 2024-01-31
AES-GCM counter fix

jenswi-linaro:

### DIFF
--- a/core/arch/arm/crypto/ghash-ce-core_a64.S
+++ b/core/arch/arm/crypto/ghash-ce-core_a64.S
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
- * Copyright (c) 2020 Linaro Limited
+ * Copyright (c) 2020, 2024 Linaro Limited
  * Copyright (C) 2014 - 2017 Linaro Ltd. <ard.biesheuvel@linaro.org>
  *
  * Accelerated GHASH implementation with ARMv8 PMULL instructions.
@@ -11,11 +11,10 @@
 #define CPU_LE(x...)	x
 
 /*
- * If the lower half of CTR is initialized with zeroes or a low value we
- * can expect that the upper half will remain unchanged. As an optimization
- * make the code to increase the upper half optional.
+ * Only increase the lowest quarter, that is, 32-bits of the counter. If
+ * it wraps it must not propagate into the upper bits.
  */
-#define INC_HALF_CTR	0
+#define INC_QUART_CTR	1
 
 	SHASH		.req	v0
 	SHASH2		.req	v1
@@ -412,7 +411,7 @@ END_FUNC pmull_ghash_update_p8
 	ld1		{SHASH.2d}, [x4], #16
 	ld1		{HH.2d}, [x4]
 	ld1		{XL.2d}, [x1]
-#if INC_HALF_CTR
+#if INC_QUART_CTR
 	ldr		x8, [x5, #8]			// load lower counter
 #else
 	ldp		x9, x8, [x5]			// load counter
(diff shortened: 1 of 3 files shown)


core/arch/arm/crypto/ghash-ce-core_a64.S
```diff
@@ -11,11 +11,10 @@
...
 
 /*
- * If the lower half of CTR is initialized with zeroes or a low value we
- * can expect that the upper half will remain unchanged. As an optimization
- * make the code to increase the upper half optional.
+ * Only increase the lowest quarter, that is, 32-bits of the counter. If
+ * it wraps it must not propagate into the upper bits.
  */
-#define INC_HALF_CTR	0
+#define INC_QUART_CTR	1
```
jforissier, line 17: IIUC the meaning of the setting was "increment the upper half of the counter", and you're changing it to "increment the lower quarter only". In both cases it is a bit complicated... why keep a configuration flag at all?

jforissier (commented):
>  core: ltc: add missing string_ext.h include
> Add a missing include <string_ext.h> needed to fix:
core/lib/libtomcrypt/gcm.c: In function ‘crypto_aes_gcm_dec_final’:
core/lib/libtomcrypt/gcm.c:198:13: error: implicit declaration of function ‘consttime_memcmp’ [-Werror=implicit-function-declaration]
>  198 |         if (consttime_memcmp(dst_tag, tag, tag_len) != 0)

I did not notice it. Which config flags are you using? In any case for this commit:

Comment below for the other one.

omasse-linaro:
> I did not notice it. Which config flags are you using? In any case for this commit:

LTC build was failing when CFG_CRYPTO_AES_GCM_FROM_CRYPTOLIB is set to y

jforissier:
> > I did not notice it. Which config flags are you using? In any case for this commit:
> 
> LTC build was failing when CFG_CRYPTO_AES_GCM_FROM_CRYPTOLIB is set to y

Thanks. BTW I am adding this flag as a new configuration in my release test suite :wink:

core/arch/arm/crypto/ghash-ce-core_a64.S
jenswi-linaro: No, it was "increment the lower half of the counter". Anyway, it's all a bit complicated. I kept it because it makes it easier to see where the counter is manipulated. That can be achieved with a bunch of comments too, but this minimizes the delta from the imported code.

jforissier: > No, it was "increment the lower half of the counter".

Hmmm... so prior to this change we did NOT increment the lower half? (it was 0). Sounds weird, the lower half are the less significant bits aren't they? The ones that matter?

core/lib/libtomcrypt/gcm.c
```diff
@@ -7,6 +7,7 @@
 #include <crypto/crypto.h>
 #include <crypto/crypto_impl.h>
 #include <stdlib.h>
+#include <string_ext.h>
 #include <string.h>
```
etienne-lms, line 11: nitpicking: swap the 2 lines above for consistency of ordering, since right above we have:
```c
#include <crypto/crypto.h>
#include <crypto/crypto_impl.h>

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with nitpicking comments.

In commit message for "core: crypto: fix internal AES-GCM counter implementation" I wonder if "last 32bit" should rather be "least significant 32 bits":
>  In AES-GCM only the last 32 bits of the counter block are increased

core/lib/libtomcrypt/gcm.c
jenswi-linaro: I've sorted the includes with `sort -n` and that's the order it gave. I guess it's ignoring the `_` and `.` so `i` comes after `h`, but `e` before `h`.

etienne-lms: Ok :). discard my comment.

jenswi-linaro:

core/lib/libtomcrypt/gcm.c
jforissier: When I apply `shift+V<select lines>:!sort -n` in `vim` I get `string.h` before `string_ext.h`... could this be due to the locale? That looks a bit better IMO because we tend to split things visually at the dot `.`, but it's no big deal.

PR #6662, opened 2024-01-31
drivers: regulator: read current level from the device

patrickdelaunay:
Always read current voltage level from the device instead of caching the level in struct regulator. This fixes issues for when the regulator level value depends on the parent regulator (supply). It is up the regulator driver to cache or not this value in their private data if applicable.

### DIFF
--- a/core/include/drivers/regulator.h
+++ b/core/include/drivers/regulator.h
@@ -85,7 +85,6 @@ struct regulator_voltages_desc {
  * @name: Regulator string name for debug purpose or NULL
  * @min_uv: Min possible voltage level in microvolt (uV)
  * @max_uv: Max possible voltage level in microvolt (uV)
- * @cur_uv: Current voltage level in microvolt (uV)
  * @flags: REGULATOR_* property flags
  * @refcount: Regulator enable request reference counter
  * @lock: Mutex for concurrent access protection
@@ -101,7 +100,6 @@ struct regulator {
 	int min_uv;
 	int max_uv;
 	/* Fields internal to regulator framework */
-	int cur_uv;
 	unsigned int flags;
 	unsigned int refcount;
 	struct mutex lock;	/* Concurrent access protection */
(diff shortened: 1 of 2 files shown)


core/include/drivers/regulator.h
```diff
@@ -286,10 +284,7 @@ static inline TEE_Result regulator_set_min_voltage(struct regulator *regulator)
  * regulator_get_voltage() - Get regulator current level in microvolt
```
etienne-lms, line 284: I think we should rephrase this to `Get regulator effective voltage level in microvolt`.
but that out of the scope of your P-R.

etienne-lms (commented):
Makes sense to me.

Nitpicking; reading again the commit message, maybe words "current level" can be confusing; this is about the current _voltage_ level, not the current _current_ level. I would suggest to either replace `current level` with `effective level` or explicit `current voltage level` what do you think? Or maybe:
```
drivers: regulator: do not cache voltage level value

...
```

This this fixes the regulator framework, I think it's worth adding a `Fixes:` tag:
`Fixes: 1a3d3273040b ("drivers: regulator framework")`.

core/include/drivers/regulator.h
patrickdelaunay: done

patrickdelaunay:
> drivers: regulator: do not cache voltage level value

done 

> 
> This this fixes the regulator framework, I think it's worth adding a `Fixes:` tag: `Fixes: 1a3d3273040b ("drivers: regulator framework")`.

done

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with minor fix in commit message:
*"It is up the regulator driver to cache or not this value in **its** private data..."*
or
*"It is up the regulator driver**s** to cache or not this value in their private data..."*

patrickdelaunay:
>"It is up the regulator driver**s** to cache or not this value in their private data..."_

done
thanks for the review.

jforissier:
> @jforissier, is this change ok to be merged?

It is, sorry for not noticing.

PR #6666, opened 2024-02-02
core: arm: fix integer overflow in generic_timer_handler()

jenswi-linaro:
read_cntfrq() can return a pretty large 32-bit number, multiplying that with a dealy of 1000 ms can overflow. Fix that by casting the result from read_cntfrq() to a uint64_t to avoid overflow during the calculation.

Fixes: ba6b29591828 ("core: arm64: Add Secure EL1 physical timer framework")

### DIFF
--- a/core/arch/arm/kernel/timer_a64.c
+++ b/core/arch/arm/kernel/timer_a64.c
@@ -13,7 +13,7 @@ static bool timer_running;
 void generic_timer_start(uint32_t time_ms)
 {
 	uint32_t exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
-	uint32_t timer_ticks;
+	uint32_t timer_ticks = 0;
 
 	cpu_spin_lock(&timer_lock);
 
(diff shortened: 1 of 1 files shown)


core/arch/arm/kernel/timer_a64.c
```diff
@@ -60,7 +60,7 @@ void generic_timer_handler(uint32_t time_ms)
 	write_cntps_ctl(0);
 
 	/* Reconfigure timer to fire time_ms from now */
-	timer_ticks = (read_cntfrq() * time_ms) / 1000;
+	timer_ticks = ((uint64_t)read_cntfrq() * time_ms) / 1000;
```
etienne-lms, line 63: Equivalent fix is needed in `generic_timer_start()`.

By the way, could you also  initialize local `timer_ticks` variables in both function, as a cleanup to satisfy the coding style rules?

etienne-lms (commented):
s/dealy/delay/ in commit message.

core/arch/arm/kernel/timer_a64.c
jenswi-linaro: Good point, I'll fix it.

jenswi-linaro:
Addressing review comments, updated to commit message to reflect the changes.

jforissier (approved):

jenswi-linaro:
Tag applied

etienne-lms (commented):

jenswi-linaro:
Tag applied

PR #6669, opened 2024-02-05
SBI related cleanups and introduce Debug Console

lyctw:
- The first 5 patches are cleanps to SBI calls
- The last one introduces debug console SBI call, which is intended to replace the legacy SBI 1.0 `sbi_console_putchar()`.

### DIFF
--- a/core/arch/riscv/kernel/sbi.c
+++ b/core/arch/riscv/kernel/sbi.c
@@ -34,11 +34,12 @@ void sbi_console_putchar(int ch)
 	sbi_ecall(SBI_EXT_0_1_CONSOLE_PUTCHAR, (unsigned long)ch);
 }
 
-int sbi_boot_hart(uint32_t hart_id, paddr_t start_addr, unsigned long arg)
+int sbi_hsm_hart_start(uint32_t hartid, paddr_t start_addr, unsigned long arg)
 {
-	struct sbiret ret;
+	struct sbiret ret = { };
 
-	ret = sbi_ecall(SBI_EXT_HSM, SBI_EXT_HSM_HART_START, hart_id, start_addr, arg);
+	ret = sbi_ecall(SBI_EXT_HSM, SBI_EXT_HSM_HART_START, hartid, start_addr,
+			arg);
 
 	return ret.error;
 }
--- a/core/arch/riscv/kernel/sbi.c
+++ b/core/arch/riscv/kernel/sbi.c
@@ -29,9 +29,13 @@ struct sbiret {
 
 #define sbi_ecall(...) _sbi_ecall(__VA_ARGS__, 0, 0, 0, 0, 0, 0, 0)
 
+/**
+ * sbi_console_putchar() - Writes given character to the console device.
+ * @ch: The data to be written to the console.
+ */
 void sbi_console_putchar(int ch)
 {
-	sbi_ecall(SBI_EXT_0_1_CONSOLE_PUTCHAR, (unsigned long)ch);
+	sbi_ecall(SBI_EXT_0_1_CONSOLE_PUTCHAR, 0, ch);
 }
 
 int sbi_hsm_hart_start(uint32_t hartid, paddr_t start_addr, unsigned long arg)
--- a/core/arch/riscv/kernel/sbi.c
+++ b/core/arch/riscv/kernel/sbi.c
@@ -30,12 +30,43 @@ struct sbiret {
 #define sbi_ecall(...) _sbi_ecall(__VA_ARGS__, 0, 0, 0, 0, 0, 0, 0)
 
 /**
+ * sbi_probe_extension() - Check if an SBI extension ID is supported or not.
+ * @extid: The extension ID to be probed.
+ *
+ * Return: 1 or an extension specific nonzero value if yes, 0 otherwise.
+ */
+int sbi_probe_extension(int extid)
+{
+	struct sbiret ret = { };
+
+	ret = sbi_ecall(SBI_EXT_BASE, SBI_EXT_BASE_PROBE_EXT, extid);
+	if (!ret.error)
+		return ret.value;
+
+	return 0;
+}
+
+/**
  * sbi_console_putchar() - Writes given character to the console device.
  * @ch: The data to be written to the console.
  */
 void sbi_console_putchar(int ch)
 {
 	sbi_ecall(SBI_EXT_0_1_CONSOLE_PUTCHAR, 0, ch);
+}
+
+/**
+ * sbi_dbcn_write_byte() - Write byte to debug console
+ * @ch:         Byte to be written
+ *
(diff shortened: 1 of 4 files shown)


jforissier:
The CI failure is unrelated to this PR, it is yet again a "No space left on device" issue. I'm on it.

jforissier:
> The CI failure is unrelated to this PR, it is yet again a "No space left on device" issue. I'm on it.

https://github.com/OP-TEE/optee_os/pull/6670

core/arch/riscv/kernel/sbi.c
```diff
@@ -29,6 +29,23 @@ struct sbiret {
...
 
+/**
+ * sbi_probe_extension() - Check if an SBI extension ID is supported or not.
+ * @extid: The extension ID to be probed.
+ *
+ * Return: 1 or an extension specific nonzero value if yes, 0 otherwise.
+ */
+int sbi_probe_extension(int extid)
+{
+	struct sbiret ret;
```
etienne-lms, line 40: `= { };`

core/arch/riscv/kernel/sbi.c
```diff
@@ -29,6 +29,23 @@ struct sbiret {
...
+
+	ret = sbi_ecall(SBI_EXT_BASE, SBI_EXT_BASE_PROBE_EXT, extid);
+	if (!ret.error)
+		return ret.value;
```
etienne-lms, line 44: for consistency with the inline description, I suggest to assert `ret.value` is not 0

core/arch/riscv/kernel/sbi.c
```diff
@@ -40,6 +57,20 @@ void sbi_console_putchar(int ch)
...
 
+/**
+ * sbi_dbcn_write_byte() - write byte to debug console
+ * @ch:         byte to be written
+ *
+ * Return:      SBI error code (SBI_SUCCESS = 0 on success)
+ */
+int sbi_dbcn_write_byte(unsigned char ch)
+{
+	struct sbiret ret;
```
etienne-lms, line 68: `= { }`

core/arch/riscv/kernel/sbi_console.c
```diff
@@ -19,19 +19,27 @@ struct sbi_console_data {
 };
 
 static struct sbi_console_data console_data __nex_bss;
+static struct serial_ops sbi_console_ops __nex_bss;
 
-static void sbi_console_putc(struct serial_chip *chip __unused,
-			     int ch)
+static void sbi_console_putc_legacy(struct serial_chip *chip __unused,
+				    int ch)
```
etienne-lms, line 25: fits in a single line

core/arch/riscv/kernel/sbi_console.c
```diff
@@ -19,19 +19,27 @@ struct sbi_console_data {
...
 {
 	sbi_console_putchar(ch);
 }
 
-static const struct serial_ops sbi_console_ops = {
-	.putc = sbi_console_putc,
-};
+static void sbi_console_putc(struct serial_chip *chip __unused,
+			     int ch)
```
etienne-lms, line 31: ditto

core/arch/riscv/kernel/sbi.c
lyctw: Hi @etienne-lms 

Thanks for the review,
you meant like this?

```c
ret = sbi_ecall(SBI_EXT_BASE, SBI_EXT_BASE_PROBE_EXT, extid);
if (!ret.error) {
        assert(ret.value != 0);
        return ret.value;
}
```

If `SBI_EXT_DBCN` is not supported, shouldn't it return 0 (false) so that `sbi_console_ops.putc` can be assigned to `sbi_console_putc_legacy()`?

etienne-lms: > you meant like this?

yes

> If `SBI_EXT_DBCN` is not supported, shouldn't it return 0 (false) so that `sbi_console_ops.putc` can be assigned to `sbi_console_putc_legacy()`?

I'm not sure to understand the question but I think yes, that's what this ecall is expected to return.

lyctw: debug console extension wasn't supported by OpenSBI until v1.3, and there are other SBI implementation may not support such extension as well, in this case, we may want to fall back to `sbi_console_putc_legacy()`, instread of freeze at the assertion here.

etienne-lms: So, when `ret.error == 0` and `ret.value == 0`, it means the SBI ID is not supported, right. If so then fine.

gagachang: > So, when `ret.error == 0` and `ret.value == 0`, it means the SBI ID is not supported

Yes, that's correct behavior. Thanks!

jforissier (commented):
Please add missing parentheses in " riscv: sbi_console: Prefer SBI v2.0 DBCN ecall over legacy sbi_console_putchar": should be sbi_console_putchar()

Note for future commits: do not hesitate to split into small pieces even for very simple things. For instance "riscv: sbi_console: Cleanup SBI console driver" could be split into five separate commits with a meaningful subject, so that one would now what changed  just by reading the subject (and it makes for easier reviews too). "riscv: sbi_console: remove unneeded #ifdef CFG_RISCV_SBI_CONSOLE", "riscv: sbi_console: remove global spinlock", etc.

For the series:

lyctw:
Hi @jforissier 

Sure, I have split the second commit into more specific changes.
Thanks for the review.

core/arch/riscv/kernel/sbi.c
```diff
@@ -29,9 +29,15 @@ struct sbiret {
 
 #define sbi_ecall(...) _sbi_ecall(__VA_ARGS__, 0, 0, 0, 0, 0, 0, 0)
 
+/**
+ * sbi_console_putchar() - Writes given character to the console device.
+ * @ch: The data to be written to the console.
+ *
+ * Return: None
```
etienne-lms, line 53: Could remove this line. Implicit when return type is `void`.

core/arch/riscv/kernel/sbi.c
```diff
@@ -40,6 +57,20 @@ void sbi_console_putchar(int ch)
 	sbi_ecall(SBI_EXT_0_1_CONSOLE_PUTCHAR, 0, ch);
 }
 
+/**
+ * sbi_dbcn_write_byte() - write byte to debug console
+ * @ch:         byte to be written
```
etienne-lms, line 62: s/byte/Byte/ for consistency with the other function inline description comments.
Remove also the inner indentation, also for consistency in this source file.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for the series preferably with my  review comments addressed.

PR #6673, opened 2024-02-06
core: dts: lx2160a: add memory region

sahilnxp:
With patch 8a6ca14 (core: arm: get DDR range from embedded DTB) now DDR ranges are taken from Embedded DTB if enabled and will ignore DDR ranges defined by register_ddr().
Since Dynamic shared memory and Embedded DTB config is enabled on LX2160A platforms, need to add the DDR ranges to the DTS.

### DIFF
--- a/core/arch/arm/dts/fsl-lx2160a.dtsi
+++ b/core/arch/arm/dts/fsl-lx2160a.dtsi
@@ -416,6 +416,12 @@
 		reg = <0x00000000 0x80000000 0 0x80000000>;
 	};
 
+	memory@2080000000 {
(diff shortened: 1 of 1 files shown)


jforissier:
Fix for the Xen CI at https://github.com/OP-TEE/build/pull/734, sorry for the trouble.

jforissier (commented):

sahilnxp:
> Fix for the Xen CI at [OP-TEE/build#734](https://github.com/OP-TEE/build/pull/734), sorry for the trouble.

No Problem.

Tags updated.

PR #6677, opened 2024-02-09
ci: qemuv8: preventively avoid "no space left on device" error

jforissier:
During my testing of build.git PR 731 ("optee_rust_examples_ext: Fix Rust toolchain conflicts"), I noticed that the "no space left on device" error was triggered yet again (obviously due to more size being taken on the disk by the Rust toolchain and the OP-TEE Rust examples).

Therefore, preventively apply the same fix as for other jobs. This way the CI should pass when 731 is merged.

Link: https://github.com/OP-TEE/build


b49020:

PR #6680, opened 2024-02-12
ta: pkcs11: Clarify context reference in step_symm_operation()

etienne-lms:
Function step_symm_operation() defines a local variable to reference the session processing context but uses both session reference and this local variable which can be confusing when reading the code. Change the implementation to only use the local variable for consistency. No functional changes.


jforissier (approved):

etienne-lms:
Thanks. Tag applied.

PR #6684, opened 2024-02-12
virt: arm: fix lock in virt_add_cookie_to_current_guest()

jenswi-linaro:
Prior to this patch was virt_add_cookie_to_current_guest() only masking interrupts while adding a shared memory cookie to the list of cookies. Proper locking is needed to serialize access to the cookie list, replace the interrupt masking with a cpu_spin_lock_xsave().

Fixes: a65dd3a6b64d ("core: spmc: support virtualization with SPMC at S-EL1")


etienne-lms (commented):
Change commit header line?
```patch
-virt: arm: fix lock in virt_add_cookie_to_current_guest()
+core: arm: fix lock in virt_add_cookie_to_current_guest()
```

jenswi-linaro:
Commit message updated, tag applied.

PR #6686, opened 2024-02-12
Misc nexus

jenswi-linaro:

### DIFF
--- a/core/include/initcall.h
+++ b/core/include/initcall.h
@@ -66,9 +66,9 @@ struct initcall {
  *  | At the end of boot_init_primary_late() just before the print:     |
  *  | "Primary CPU switching to normal world boot"                      |
  *  +-------------------------------+-----------------------------------+
- *  | 1. call_preinitcalls()        | In the nexus                      |
+ *  | 1. call_preinitcalls()        | In the nexus, final calls         |
  *  | 2. call_initcalls()           +-----------------------------------+
- *  | 3. call_finalcalls()          | 1. call_finalcalls()              |
+ *  | 3. call_finalcalls()          | 1. boot_final() / nex_*init*()    |
  *  +-------------------------------+-----------------------------------+
  *  | "Primary CPU switching to normal world boot" is printed           |
  *  +-------------------------------+-----------------------------------+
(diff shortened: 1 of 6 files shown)


core/include/initcall.h
```diff
@@ -101,6 +101,31 @@ struct initcall {
 
 #define boot_final(fn)			__define_initcall(final, 1, fn)
 
+/*
+ * These nex_* init-calls are provided for drivers and services that
+ * resides in the nexus in case of virtualization.  The init-calls are
```
etienne-lms, line 106: s/resides/reside/
Can replace the 2 consecutive space chars with a single one.

core/include/initcall.h
```diff
@@ -101,6 +101,31 @@ struct initcall {
...
+ * performed before exiting to at the end of boot initialization. In case
```
etienne-lms, line 107: s/exiting to/exiting to non-secure world/ ?

etienne-lms (commented):
Swap commits "core: add nex_*init-calls" and "core: move _time_source into __nex_bss" in the series.
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with comments addressed.

core/include/initcall.h
jenswi-linaro: Yes, thanks.

jenswi-linaro:
Comments addressed, tags applied.

core/include/mm/core_memprot.h
```diff
@@ -97,8 +97,31 @@ paddr_t virt_to_phys(void *va);
  */
 vaddr_t core_mmu_get_va(paddr_t pa, enum teecore_memtypes type, size_t len);
 
-/* Return true if @va relates to a unpaged section else false */
-bool is_unpaged(void *va);
+/*
```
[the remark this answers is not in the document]
jenswi-linaro: Yes, but we don't care about doxygen. :-)

core/mm/core_mmu.c
```diff
@@ -2451,16 +2451,20 @@ vaddr_t core_mmu_get_va(paddr_t pa, enum teecore_memtypes type, size_t len)
...
-bool is_unpaged(void *va __unused)
+#endif
+
+#ifdef CFG_NS_VIRTUALIZATION
+bool is_nexus(const void *va)
 {
-	return true;
+	vaddr_t v = (vaddr_t)va;
+
+	return v >= VCORE_START_VA && v < VCORE_NEX_RW_PA + VCORE_NEX_RW_SZ;
```
jforissier, line 2467: This `_PA` in `VCORE_NEX_RW_PA` is a bit disturbing. Should all the `_PA` and `_VA` defines in `linker.h` be renamed `_ADDR`? They reference symbols that may be relocated so depending on whether or not the MMU is turned on, they may be PAs or VAs, and when they are VAs their value may differ from the PAs when ASLR is enabled.

jforissier (commented):
Reviewing "core: add is_nexus() and refactor is_unapged():
- s/is_unapged/is_unpaged()/
- ...is updated **to** perform...
- One more comment below

core/mm/core_mmu.c
jenswi-linaro: Yes, it would be nice with a consistent naming for address defines that are updated with relocation (ASLR and `CFG_CORE_PHYS_RELOCATABLE=y`).
We should do something similar for the address defines that don't change with relocation.

But I'd rather do it in a separate PR.

Issue #6649, opened 2024-01-29
boot_init_memtag is called before mmu is enabled.


sichunqin:
Memory tag is associated with VA (mmu), so I understand that the boot_init_memtag should be called after mmu is enabled. 
If it is called before mmu is enabled, the OPTEE would panic on some platforms. Can someone explains why it is called early? Thanks!

jenswi-linaro:
> Memory tag is associated with VA (mmu), so I understand that the boot_init_memtag should be called after mmu is enabled.

Could you tell me how you came to that conclusion? 

> If it is called before mmu is enabled, the OPTEE would panic on some platforms. Can someone explains why it is called early? Thanks!


Please see the comment in `core/arch/arm/kernel/entry_a64.S` where `boot_init_memtag()` is called:
``` C
        /*
         * If FEAT_MTE2 is available, initializes the memtag callbacks.
         * Tags for OP-TEE core memory are then cleared to make it safe to
         * enable MEMTAG below.
         */
        bl      boot_init_memtag
```
The tags must be cleared before they are used. I can't remember why we do that early except that it's likely to avoid potential coherency problems. You may want to consult the Arm ARM for further details.

sichunqin:
Thanks for the reply!
The panic happens on ARMV9. I am afraid that it might be ARMv9 specific problem. BTW, How can I consult ARM?

sichunqin:
BTW, during reset, memory could be cleared up. I guess it also include tag cleap up. It seems that bl32 tag clearing up behavior seems not be required.

jenswi-linaro:
This is the Arm ARM I was talking about https://developer.arm.com/documentation/ddi0487/ja/?lang=en

sichunqin:
Thanks for the share The panic happens on “”dc gva” instruction which is called by boot_init_memtag on ARMv9 platform. The MTE uses virtual address TBI, it is why I think enabled MMU is necessary.

jenswi-linaro:
So do the other `dc` instructions and we all know we should clear the cache before enabling MMU.
Have you looked at ESR when you get the exception? https://esr.arm64.dev/ is quite handy to interpret the value.

sichunqin:
@jenswi-linaro Thanks for the info!
According to the ARM reference document, dc gva instructions seems not relate to cache maintenance, but allocate tag. The reference doc doesn't explain clearly how it work, and how it to be useed. dc I see the dc gva is used to tag or untag memory in OPTEE. 
[...]

PR #6690, opened 2024-02-13
Memtag fix

jenswi-linaro:
I wasn't able to reproduce the problem reported in https://github.com/OP-TEE/optee_os/issues/6649 with the FVP Foundation model or Base model. In the Foundation model, MTE was disabled, so not much to test. With the Base model, I was able to enable MTE with `-C cluster0.memory_tagging_support_level=2`, but it still didn't abort when clearing tags without MMU enabled.

I'm fixing the issue differently here since we're mapping the RO/RX parts with tags also. I'm not sure if that makes sense, but we don't want to risk having garbage in the memory tags or we could have problems.

@odeprez, @sichunqin does this fix the memtag issue for you?

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -374,12 +374,28 @@ static void init_asan(void)
 /* Called from entry_a64.S only when MEMTAG is configured */
 void boot_init_memtag(void)
 {
-	paddr_t base = 0;
-	paddr_size_t size = 0;
-
 	memtag_init_ops(feat_mte_implemented());
-	core_mmu_get_secure_memory(&base, &size);
-	memtag_set_tags((void *)(vaddr_t)base, size, 0);
+}
+
+/* Called from entry_a64.S only when MEMTAG is configured */
+void boot_clear_memtag(void)
+{
+	enum teecore_memtypes mtypes[] = {
+		MEM_AREA_TEE_RAM, MEM_AREA_TEE_RAM_RW, MEM_AREA_NEX_RAM_RO,
+		MEM_AREA_NEX_RAM_RW, MEM_AREA_TEE_ASAN, MEM_AREA_TA_RAM
+	};
+	vaddr_t s = 0;
+	vaddr_t e = 0;
+	size_t n = 0;
+
+	for (n = 0; n < ARRAY_SIZE(mtypes); n++) {
+		core_mmu_get_mem_by_type(mtypes[n], &s, &e);
+		if (e > s) {
+			DMSG("Clearing tags for VA %#"PRIxVA"..%#"PRIxVA,
+			     s, e - 1);
+			memtag_set_tags((void *)s, e - s, 0);
+		}
+	}
 }
 #endif
 
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -353,6 +353,10 @@ clear_nex_bss:
 	bl	console_init
 #endif
 
+#ifdef CFG_MEMTAG
+	bl	boot_clear_memtag
+#endif
+
 #ifdef CFG_NS_VIRTUALIZATION
 	/*
 	 * Initialize partition tables for each partition to
--- a/core/include/kernel/boot.h
+++ b/core/include/kernel/boot.h
@@ -45,6 +45,7 @@ extern const struct core_mmu_config boot_mmu_config;
 void boot_init_primary_early(void);
 void boot_init_primary_late(unsigned long fdt, unsigned long manifest);
 void boot_init_memtag(void);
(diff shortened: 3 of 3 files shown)


odeprez:
Hi @jenswi-linaro sorry for the delay, yes it boots for me on FVP with those patches.
However I get from Arm arch team: 
"Any attempt to write a MTE tag on an PA with a memory attribute different than Cacheable-WB-tagged[1] will be write-ignored (no abort/fault, but no write happening as well)
[1] And you need a MMU on to have such memory attribute"

So I wonder about attempting to clear tags before MMU enabled, as indicated above that just result in ignored accesses.

jenswi-linaro:
Thanks for looking into this. I'll revert to something closer to your original suggestion.

jenswi-linaro:
Another attempt close to @odeprez's original patch.

odeprez:
Good to me. Just a small typo (from me!) to fix at top of commit message "[Based] on following observations ...".

jenswi-linaro:
Thanks, @odeprez, I've updated the commit message.

Would someone with insight into OP-TEE memory types take a look at `boot_clear_memtag()`, please?

core/arch/arm/kernel/boot.c
```diff
@@ -374,12 +374,28 @@ static void init_asan(void)
...
+	size_t n = 0;
+
+	for (n = 0; n < ARRAY_SIZE(mtypes); n++) {
+		core_mmu_get_mem_by_type(mtypes[n], &s, &e);
+		if (e > s) {
+			DMSG("Clearing tags for VA %#"PRIxVA"..%#"PRIxVA,
+			     s, e - 1);
+			memtag_set_tags((void *)s, e - s, 0);
+		}
+	}
```
jforissier, line 398: `s` and `e` need to be reset on each loop, otherwise we'll set tags twice on the same range if it happens that some type is not found. Or if we're assuming that all types must be found and are correctly returned then drop the `if (e > s)` entirely?

jenswi-linaro: `core_mmu_get_mem_by_type()` clears `s` and `e` if a type isn't found.

jforissier: Correct. How did I miss that? :open_mouth: Anyways:

core/arch/arm/kernel/boot.c
```diff
@@ -374,12 +374,28 @@ static void init_asan(void)
...
-	core_mmu_get_secure_memory(&base, &size);
-	memtag_set_tags((void *)(vaddr_t)base, size, 0);
+}
+
+/* Called from entry_a64.S only when MEMTAG is configured */
+void boot_clear_memtag(void)
+{
+	enum teecore_memtypes mtypes[] = {
+		MEM_AREA_TEE_RAM, MEM_AREA_TEE_RAM_RW, MEM_AREA_NEX_RAM_RW,
+		MEM_AREA_TEE_ASAN, MEM_AREA_TA_RAM
```
etienne-lms, line 385: Maybe a stupid question but yhy are `MEM_AREA_TEE_RAM_RW` & `MEM_AREA_NEX_RAM_RW` considered but not the other tagged memory types:
`MEM_AREA_TEE_RAM_RX`
`MEM_AREA_INIT_RAM_RX`
`MEM_AREA_IDENTITY_MAP_RX`
`MEM_AREA_TEE_RAM_RO`
`MEM_AREA_INIT_RAM_RO`
`MEM_AREA_NEX_RAM_RO`

etienne-lms (commented):
The change looks good to me but I did not spent enough time for a R-b tag.
In the mean time:  `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` if my comment/question is stupid.

core/arch/arm/kernel/boot.c
jenswi-linaro: We can't write to tags in read-only memory.

jenswi-linaro:
Tags applied.

core/arch/arm/kernel/boot.c
etienne-lms: Ok thanks, indeed you mention that in the commit message.
From [core_mmu.c](https://elixir.bootlin.com/op-tee/4.1.0/source/core/mm/core_mmu.c#L748) : `MEM_AREA_NEX_RAM_RO` is writable during the initialization:
```c
case MEM_AREA_NEX_RAM_RO: /* This has to be r/w during init runtime */
```
Should it be considered here?

jenswi-linaro: I guess it doesn't harm to add that too.

jenswi-linaro:
Update

etienne-lms (commented):

jenswi-linaro:
Squashed.

PR #6694, opened 2024-02-16
core: reset cancellation mask on TA entry

jenswi-linaro:
Before this patch, the TA cancellation mask was only reset when the session was created, but the GP spec requires the cancellation mask to be reset each time a TA is entered via one of its entry points. So fix this by resetting the cancellation mask at the same time as when setting the cancellation request timeout.

Link: https://github.com/OP-TEE/optee_test/issues/731
Fixes: b01047730e77 ("Open-source the TEE Core")

Depends on https://github.com/OP-TEE/optee_test/pull/732

@bvalliere @c-auger please provide your email addresses if you want a `Reported-by:` tag for this patch.

### DIFF
--- a/core/kernel/user_ta.c
+++ b/core/kernel/user_ta.c
@@ -223,11 +223,12 @@ out:
 	dec_recursion();
 out_clr_cancel:
 	/*
-	 * Clear the cancel state now that the user TA has returned. The next
+	 * Reset the cancel state now that the user TA has returned. The next
 	 * time the TA will be invoked will be with a new operation and should
 	 * not have an old cancellation pending.
 	 */
 	ta_sess->cancel = false;
+	ta_sess->cancel_mask = true;
 
 	return res;
 }


c-auger:
My email adress is cedric.auger@provenrun.com

c-auger:
By the way, I see that the `cancel` field of the session is reset to `false` in https://github.com/OP-TEE/optee_os/blob/897aaf117e39c14cdbae681c91c863f9ef49ee03/core/kernel/user_ta.c#L230
At first glance, I would tell that the setting of `cancel_mask` and `cancel` should probably be done on the same place.

Also, now that I think of it, although `TA_CloseSessionEntryPoint()` and `TA_DestroyEntryPoint()` are not cancellable, maybe that we should set `cancel_mask` to false also in these cases, in case `TEE_Wait()` is called at some place in these places.

jenswi-linaro:
> At first glance, I would tell that the setting of `cancel_mask` and `cancel` should probably be done around the same location.

Yes, that makes sense.

> Also, now that I think of it, although `TA_CloseSessionEntryPoint()` and `TA_DestroyEntryPoint()` are not cancellable, maybe that we should set `cancel_mask` to false also in these cases, in case `TEE_Wait()` is called at some place in these places.

I assume you mean to set `cancel_mask` to true as in masking cancellations. It seems that the spec intends to have cancellations masked by default so that a TA must actively enable cancellations each time it is entered.

`TA_CloseSessionEntryPoint()` is entered via `user_ta_enter_close_session()`, but `TA_DestroyEntryPoint()` called in the TA right after `TA_CloseSessionEntryPoint()` has returned so OP-TEE core can't reset `cancel_mask` in between. We could let the TA itself mask cancellations again, but that might be a bit too much.

etienne-lms (commented):
Typo in commit message?
```patch
- ... the TA cancellation was mask only reset when ...
+ ... the TA cancellation mask was only reset when ...
```

`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for the implmentation.

jforissier (commented):

c-auger:
I think that indeed we should not reset `cancel` in `reset_entry_cancellation()`. I was wrong to suggest it the first time.

That said, when we were implementing our TEE using your frontend on non secure world, we spotted something that seemed strange to us, and is somehow out of the scope of this issue, so I will try to make it short.
Cancellations are by the specification bound to a `TEEC_Operation` which in turn is bound to a session. Still, in your implementation you set `cancel_id` of operations to 0, so I guess that you base your cancellation on sessions rather than operations. For instance, if you follow the following protocol:

- create an operation `o1`
- open a session with it
- create an operation `o2`
- create a thread
- invoke a command waiting for ever using `o2` in one thread
- cancel the operation `o1` (and not `o2`) in the other thread

Then I suspect but did not test (so I might be completely wrong about that) it that the invoke command will be cancelled, because you do not seem to track the operations themselves. Otherwise stated, I think that you cancel sessions and not operations.

Now to go back to the issue, yes, the sessions state should be reset after processing the operation, so that the next operation will start on a cleaned state. Maybe that it would be easier to follow the logic if you create a `struct current_operation` inside the structure of the session which will track the cancellation state. This structure requiring to be cleaned on operation termination and initiated on operation creation.

jenswi-linaro:
Commit message updated and tag applied.

jenswi-linaro:
I'm sorry, I forgot to squash the earlier commits and add Etienne's A-B, I've done that now.

PR #6695, opened 2024-02-19
core: io: fix IO_READ32_POLL_TIMEOUT() when delay is 0us

etienne-lms:
Fix detection of timeout condition in IO_READ32_POLL_TIMEOUT() that was never triggered when delay argument is 0us. Indeed 0 is not a useful increment value for a timeout counter.

Fixes: 97ea199a2ae8 ("core: io: IO_READ32_POLL_TIMEOUT()")

### DIFF
--- a/core/include/io.h
+++ b/core/include/io.h
@@ -7,6 +7,7 @@
 
 #include <compiler.h>
 #include <kernel/delay.h>
+#include <kernel/delay_arch.h>
 #include <stdint.h>
 #include <types_ext.h>
 #include <utee_defines.h>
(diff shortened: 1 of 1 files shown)


GseoC:

core/include/io.h
```diff
@@ -287,14 +288,13 @@ static inline void io_clrsetbits8(vaddr_t addr, uint8_t clear_mask,
  */
 #define IO_READ32_POLL_TIMEOUT(_addr, _val, _cond, _delay_us, _timeout_us) \
 	({ \
-		uint32_t __timeout = 0; \
+		uint64_t __timeout = timeout_init_us(_timeout_us); \
 		uint32_t __delay = (_delay_us); \
 		\
-		while (__timeout < (_timeout_us)) { \
+		while (!timeout_elapsed(__timeout)) { \
```
jforissier, line 294: This adds a dependency on timer counters (`read_cntpct()`/`read_cntvct()`), is it OK? I believe it is but just to make sure.

etienne-lms: Sole current platforms using the macro are platforms supporting `timeout_xxx()` functions (plat-d06 and plat-stm32mp*) so I guess it's OK but maybe i'm too confident. The alternate easiest way is to force use of a 1us delay in the macro when argument `_delay_us` is 0. What do you think?

jforissier: Or `static_assert(IS_ENABLED(CFG_CORE_HAS_GENERIC_TIMER))`?

etienne-lms: Are you ok we use `timeout_xxx()` functions in this macro and address the issue later if some platform does not meet this requirement?

jforissier: Yes, absolutely.

etienne-lms: Ok thanks for the feedback.

etienne-lms:
Review tags applied.

PR #6699, opened 2024-02-20
core: mmu: add MEM_AREA_ROM_SEC in check_mem_map

tbourgoi:
Handle MEM_AREA_ROM_SEC in check_mem_map switch case.

Fixes: fc7e0cc38b99 ("core: MEM_AREA_ROM_SEC maps secure read only cached memory")


etienne-lms (commented):
Indeed I missed this change in the referred commit. Thanks.

I think you could say in the commit message that this is needed unless what OP-TEE core panics when such memory area is being mapped.

Please append parentheses to check_mem_map label in commit message (2 occurrences): s/`check_mem_map`/`check_mem_map()`/g.

With that addressed: `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`

etienne-lms (approved):

PR #6700, opened 2024-02-21
Introduce semihosting facilities and semihosting console driver

gagachang:
Hello,

This series implement semihosting facilities.
Please review it if this is feasible in  OP-TEE system, thanks!

The new config option `CFG_SEMIHOSTING` is used to compile in-kernel semihosting functions.
Architectures who support semihosting should provide their implementation of `__do_semihosting()`, which consists of special instructions to trigger the semihosting.
In this series, we've implemented RISC-V version.

Currently we only implement semihosting  open/read/write operations.
Based on this in-kernel semihosting, semihosting console driver is also introduced.
There are two paths to output the log:
1. host side console
2. host side file

The path is decided by `file_path` parameter provided by the caller of `semihosting_console_init()`.
The new config option `CFG_SEMIHOSTING_CONSOLE` is used to compile semihosting console driver.

Example code to use semihosting console driver:
```c
#include <drivers/semihosting_console.h>

static struct semihosting_console_data console_data __nex_bss;

void console_init(void)
{
          /* User must only choose one of the following two ways */
          /* 1. Output log to a file on semihosting host side system */
	  semihosting_console_init(&console_data, "semihosting.txt");
          /* 2. Output log to semihosting host side console */
	  semihosting_console_init(&console_data, NULL);

}

```

### DIFF
--- a/core/arch/riscv/kernel/sub.mk
+++ b/core/arch/riscv/kernel/sub.mk
@@ -12,6 +12,7 @@ srcs-y += thread_arch.c
 srcs-y += arch_scall_rv.S
 srcs-y += arch_scall.c
 srcs-$(CFG_UNWIND) += unwind_rv.c
+srcs-$(CFG_SEMIHOSTING) += semihosting_rv.S
 srcs-y += thread_optee_abi.c
 srcs-y += thread_optee_abi_rv.S
 asm-defines-y += asm-defines.c
--- a/core/drivers/semihosting_console.c
+++ b/core/drivers/semihosting_console.c
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+
+#include <compiler.h>
+#include <console.h>
+#include <drivers/semihosting_console.h>
+#include <drivers/serial.h>
+#include <kernel/semihosting.h>
+#include <util.h>
+
+/*
+ * struct semihosting_console_data - Structure for semihosting console driver
+ * @chip - General structure for each serial chip
+ * @fd - Handle of the file at @file_path when semihosting_console_init() is
+ *       called, or -1 if using the semihosting console
+ */
+struct semihosting_console_data {
+	struct serial_chip chip;
+	int fd;
+};
+
+static struct semihosting_console_data sh_console_data __nex_bss;
+
+static void semihosting_console_putc(struct serial_chip *chip __unused, int ch)
+{
+	semihosting_sys_writec(ch);
+}
+
+static int semihosting_console_getchar(struct serial_chip *chip __unused)
+{
+	return semihosting_sys_readc();
+}
+
+static const struct serial_ops semihosting_console_ops = {
+	.putc = semihosting_console_putc,
+	.getchar = semihosting_console_getchar,
+};
+DECLARE_KEEP_PAGER(semihosting_console_ops);
+
+static void semihosting_console_fd_putc(struct serial_chip *chip __unused,
+					int ch)
+{
+	if (sh_console_data.fd >= 0)
+		semihosting_write(sh_console_data.fd, &ch, 1);
+}
+
+static const struct serial_ops semihosting_console_fd_ops = {
+	.putc = semihosting_console_fd_putc,
+};
+DECLARE_KEEP_PAGER(semihosting_console_fd_ops);
+
+void semihosting_console_init(const char *file_path)
+{
+	if (file_path) {
+		/* Output log to given file on the semihosting host system. */
+		sh_console_data.chip.ops = &semihosting_console_fd_ops;
+		sh_console_data.fd =
+			semihosting_open(file_path, O_RDWR | O_CREAT | O_TRUNC);
+	} else {
+		/* Output log to semihosting host debug console. */
+		sh_console_data.chip.ops = &semihosting_console_ops;
+		sh_console_data.fd = -1;
+	}
+
+	register_serial_console(&sh_console_data.chip);
+}
--- a/core/include/drivers/semihosting_console.h
+++ b/core/include/drivers/semihosting_console.h
@@ -1,0 +1,23 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+#ifndef __DRIVERS_SEMIHOSTING_CONSOLE_H
+#define __DRIVERS_SEMIHOSTING_CONSOLE_H
+
+#ifdef CFG_SEMIHOSTING_CONSOLE
+/*
+ * Initialize console which uses architecture-specific semihosting mechanism.
+ * If @file_path is not NULL, OP-TEE OS will try to output log to that file,
+ * which is on the semihosting host system.
+ * Otherwise, if @file_path is NULL, OP-TEE OS will try to output log to the
+ * semihosting host debug console.
+ */
+void semihosting_console_init(const char *file_path);
+#else
+static inline void semihosting_console_init(const char *file_path __unused)
+{
+}
+#endif
+
+#endif /* __DRIVERS_SEMIHOSTING_CONSOLE_H */
--- a/core/include/kernel/semihosting.h
+++ b/core/include/kernel/semihosting.h
@@ -1,0 +1,23 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+#ifndef __KERNEL_SEMIHOSTING_H
+#define __KERNEL_SEMIHOSTING_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include <sys/fcntl.h>
+#include <util.h>
+
+/* Perform architecture-specific semihosting instructions. */
+uintptr_t __do_semihosting(uintptr_t op, uintptr_t arg);
+
+char semihosting_sys_readc(void);
+void semihosting_sys_writec(char c);
+int semihosting_open(const char *fname, int flags);
+size_t semihosting_read(int fd, void *ptr, size_t len);
+size_t semihosting_write(int fd, const void *ptr, size_t len);
+int semihosting_close(int fd);
+
+#endif /* __KERNEL_SEMIHOSTING_H */
--- a/core/kernel/console.c
+++ b/core/kernel/console.c
@@ -6,6 +6,7 @@
 #include <compiler.h>
 #include <console.h>
 #include <drivers/cbmem_console.h>
+#include <drivers/semihosting_console.h>
 #include <drivers/serial.h>
 #include <kernel/dt.h>
 #include <kernel/dt_driver.h>
@@ -24,7 +25,10 @@ __weak void plat_console_init(void)
 
 void console_init(void)
 {
-	plat_console_init();
+	if (IS_ENABLED(CFG_SEMIHOSTING_CONSOLE))
+		semihosting_console_init(CFG_SEMIHOSTING_CONSOLE_FILE);
+	else
+		plat_console_init();
 }
 
 void __weak console_putc(int ch)
--- a/core/kernel/console.c
+++ b/core/kernel/console.c
@@ -17,6 +17,16 @@
 
 static struct serial_chip *serial_console __nex_bss;
 
+/* May be overridden by platform */
+__weak void plat_console_init(void)
+{
+}
+
+void console_init(void)
+{
+	plat_console_init();
+}
+
 void __weak console_putc(int ch)
 {
 	if (!serial_console)
--- a/core/kernel/semihosting.c
+++ b/core/kernel/semihosting.c
@@ -1,0 +1,169 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+
+#include <kernel/semihosting.h>
+#include <string.h>
+
+/*
+ * ARM and RISC-V have defined the standard way to perform
+ * the semihosting operations.
+ * - Operation codes and open modes are identical.
+ * - The implementation of the low-level __do_semihosting() call is
+ *   architecture-specific.
+ * - Arm semihosting interface:
+ *   https://developer.arm.com/documentation/dui0471/g/Semihosting/The-semihosting-interface
+ * - RISC-V semihosting interface:
+ *   https://github.com/riscv-non-isa/riscv-semihosting/blob/main/binary-interface.adoc
+ */
+
+/* An integer that specifies the file open mode */
+enum semihosting_open_mode {
+	SEMIHOSTING_OPEN_R = 0,
+	SEMIHOSTING_OPEN_RB = 1,
+	SEMIHOSTING_OPEN_RX = 2,
+	SEMIHOSTING_OPEN_RXB = 3,
+	SEMIHOSTING_OPEN_W = 4,
+	SEMIHOSTING_OPEN_WB = 5,
+	SEMIHOSTING_OPEN_WX = 6,
+	SEMIHOSTING_OPEN_WXB = 7,
+	SEMIHOSTING_OPEN_A = 8,
+	SEMIHOSTING_OPEN_AB = 9,
+	SEMIHOSTING_OPEN_AX = 10,
+	SEMIHOSTING_OPEN_AXB = 11,
+};
+
+enum semihosting_sys_ops {
+	/* Regular operations */
+	SEMIHOSTING_SYS_OPEN = 0x01,
+	SEMIHOSTING_SYS_CLOSE = 0x02,
+	SEMIHOSTING_SYS_WRITEC = 0x03,
+	SEMIHOSTING_SYS_WRITE = 0x05,
+	SEMIHOSTING_SYS_READ = 0x06,
+	SEMIHOSTING_SYS_READC = 0x07,
+};
+
+struct semihosting_param_t {
+	uintptr_t param0;
+	uintptr_t param1;
+	uintptr_t param2;
+};
+
+/**
+ * @brief Read one character byte from the semihosting host debug terminal
+ *
+ * @retval the character read from the semihosting host
+ */
+char semihosting_sys_readc(void)
+{
+	return __do_semihosting(SEMIHOSTING_SYS_READC, 0);
+}
+
+/**
+ * @brief Write one character byte to the semihosting host debug terminal
+ * @param c: the character to be written
+ */
+void semihosting_sys_writec(char c)
+{
+	__do_semihosting(SEMIHOSTING_SYS_WRITEC, (uintptr_t)&c);
+}
+
+/**
+ * @brief Request the semihosting host to open a file on the host system
+ * @param fname: the path or name of the file
+ * @param flags: sys/fcntl.h standard flags to open the file with
+ *
+ * @retval nonzero if OK, or -1 if fails
+ */
+int semihosting_open(const char *fname, int flags)
+{
+	int semi_open_flags = 0;
+	const int flags_mask = O_RDONLY | O_WRONLY | O_RDWR |
+			       O_CREAT | O_TRUNC | O_APPEND;
+	struct semihosting_param_t arg = { };
+
+	/* Convert the flags to semihosting open. */
+	switch (flags & flags_mask) {
+	case O_RDONLY:				/* 'r' */
+		semi_open_flags = SEMIHOSTING_OPEN_R;
+		break;
+	case O_WRONLY | O_CREAT | O_TRUNC:	/* 'w' */
+		semi_open_flags = SEMIHOSTING_OPEN_W;
+		break;
+	case O_WRONLY | O_CREAT | O_APPEND:	/* 'a' */
+		semi_open_flags = SEMIHOSTING_OPEN_A;
+		break;
+	case O_RDWR:				/* 'r+' */
+		semi_open_flags = SEMIHOSTING_OPEN_RX;
+		break;
+	case O_RDWR | O_CREAT | O_TRUNC:	/* 'w+' */
+		semi_open_flags = SEMIHOSTING_OPEN_WX;
+		break;
+	case O_RDWR | O_CREAT | O_APPEND:	/* 'a+' */
+		semi_open_flags = SEMIHOSTING_OPEN_AX;
+		break;
+	default:
+		return -1;
+	}
+
+	arg.param0 = (uintptr_t)fname;
+	arg.param1 = semi_open_flags;
+	arg.param2 = strlen(fname);
+
+	return (int)__do_semihosting(SEMIHOSTING_SYS_OPEN, (uintptr_t)&arg);
+}
+
+/**
+ * @brief Read data from a file on the semihosting host system
+ * @param fd: a handle for a file previously opened
+ * @param ptr: pointer to a buffer
+ * @param len: the number of bytes to read to the buffer from the file
+ *
+ * @retval zero if OK, the same value as @len if fails, smaller value than @len
+ * for partial success
+ */
+size_t semihosting_read(int fd, void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = fd,
+		.param1 = (uintptr_t)ptr,
+		.param2 = len
+	};
+
+	return __do_semihosting(SEMIHOSTING_SYS_READ, (uintptr_t)&arg);
+}
+
+/**
+ * @brief Write data into a file on the semihosting host system
+ * @param fd: a handle for a file previously opened
+ * @param ptr: pointer to a buffer
+ * @param len: the number of bytes to be written from the buffer to the file
+ *
+ * @retval zero if OK, otherwise the number of bytes that are not written
+ */
+size_t semihosting_write(int fd, const void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = fd,
+		.param1 = (uintptr_t)ptr,
+		.param2 = len
+	};
+
+	return __do_semihosting(SEMIHOSTING_SYS_WRITE, (uintptr_t)&arg);
+}
+
+/**
+ * @brief Close a file on the semihosting host system
+ * @param fd: a handle for a file previously opened
+ *
+ * @retval zero if OK, -1 if fails
+ */
+int semihosting_close(int fd)
+{
+	struct semihosting_param_t arg = {
+		.param0 = fd,
+	};
+
+	return (int)__do_semihosting(SEMIHOSTING_SYS_CLOSE, (uintptr_t)&arg);
+}
--- a/core/kernel/sub.mk
+++ b/core/kernel/sub.mk
@@ -54,6 +54,8 @@ srcs-y += pseudo_ta.c
 
 srcs-$(CFG_TRANSFER_LIST) += transfer_list.c
 
+srcs-$(CFG_SEMIHOSTING) += semihosting.c
+
 ifeq ($(CFG_SYSCALL_FTRACE),y)
 # We would not like to profile spin_lock_debug.c file as it provides
 # common APIs that are needed for ftrace framework to trace syscalls.
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -1158,3 +1158,8 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_HUK,CFG_DT))
 # DT node "/options/op-tee/widevine" to some specific TAs.
 CFG_WIDEVINE_PTA ?= n
 $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
+
+# Semihosting is a debugging mechanism that enables code running on an embedded
+# system (also called the target) to communicate with and use the I/O of the
+# host computer.
+CFG_SEMIHOSTING ?= n
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -1159,6 +1159,20 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_HUK,CFG_DT))
 CFG_WIDEVINE_PTA ?= n
 $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
 
+# CFG_SEMIHOSTING_CONSOLE, when enabled, embeds a semihosting console driver.
+# When CFG_SEMIHOSTING_CONSOLE_FILE=NULL, OP-TEE console reads/writes
+# trace messages from/to the debug terminal of the semihosting host computer.
+# When CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}", OP-TEE console
+# outputs trace messages to that file. Output to "optee.log" by default.
+CFG_SEMIHOSTING_CONSOLE ?= n
+ifeq ($(CFG_SEMIHOSTING_CONSOLE),y)
+$(call force,CFG_SEMIHOSTING,y)
+endif
+CFG_SEMIHOSTING_CONSOLE_FILE ?= "optee.log"
+ifeq ($(CFG_SEMIHOSTING_CONSOLE_FILE),)
+$(error CFG_SEMIHOSTING_CONSOLE_FILE cannot be empty)
+endif
+
 # Semihosting is a debugging mechanism that enables code running on an embedded
 # system (also called the target) to communicate with and use the I/O of the
 # host computer.
--- a/core/arch/arm/plat-amlogic/main.c
+++ b/core/arch/arm/plat-amlogic/main.c
@@ -13,7 +13,7 @@ static struct amlogic_uart_data console_data;
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE,
 			CORE_MMU_PGDIR_SIZE);
 
-void console_init(void)
+void plat_console_init(void)
 {
 	amlogic_uart_init(&console_data, CONSOLE_UART_BASE);
 	register_serial_console(&console_data.chip);
--- a/core/arch/arm/plat-aspeed/platform_ast2600.c
+++ b/core/arch/arm/plat-aspeed/platform_ast2600.c
@@ -70,7 +70,7 @@ void boot_secondary_init_intc(void)
 	gic_init_per_cpu();
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 	serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
 			     CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-aspeed/platform_ast2700.c
+++ b/core/arch/arm/plat-aspeed/platform_ast2700.c
@@ -29,7 +29,7 @@ void boot_secondary_init_intc(void)
 	gic_init_per_cpu();
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 	serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
 			     CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-bcm/main.c
+++ b/core/arch/arm/plat-bcm/main.c
@@ -62,7 +62,7 @@ void plat_trace_ext_puts(const char *str)
 		bcm_elog_putchar(*p);
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 	serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
 			     CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-corstone1000/main.c
+++ b/core/arch/arm/plat-corstone1000/main.c
@@ -31,7 +31,7 @@ void boot_secondary_init_intc(void)
 	gic_init_per_cpu();
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 	pl011_init(&console_data, CONSOLE_UART_BASE, CONSOLE_UART_CLK_IN_HZ,
 		   CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-d02/main.c
+++ b/core/arch/arm/plat-d02/main.c
@@ -16,7 +16,7 @@ static struct hi16xx_uart_data console_data;
 register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE,
 			HI16XX_UART_REG_SIZE);
 
-void console_init(void)
+void plat_console_init(void)
 {
 	hi16xx_uart_init(&console_data, CONSOLE_UART_BASE,
 			 CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-d06/main.c
+++ b/core/arch/arm/plat-d06/main.c
@@ -11,7 +11,7 @@ static struct lpc_uart_data console_data __nex_bss;
 
 register_phys_mem_pgdir(MEM_AREA_IO_NSEC, LPC_BASE, LPC_SIZE);
 
-void console_init(void)
+void plat_console_init(void)
 {
 	lpc_uart_init(&console_data, LPC_BASE, CONSOLE_UART_CLK_IN_HZ,
 		      CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-hikey/main.c
+++ b/core/arch/arm/plat-hikey/main.c
@@ -41,7 +41,7 @@ register_ddr(DRAM1_BASE, DRAM1_SIZE_NSEC);
 register_ddr(DRAM2_BASE, DRAM2_SIZE_NSEC);
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 	pl011_init(&console_data, CONSOLE_UART_BASE, CONSOLE_UART_CLK_IN_HZ,
 		   CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-hisilicon/main.c
+++ b/core/arch/arm/plat-hisilicon/main.c
@@ -24,7 +24,7 @@ register_phys_mem(MEM_AREA_IO_SEC, CPU_CRG_BASE, CPU_CRG_SIZE);
 register_phys_mem(MEM_AREA_IO_SEC, SYS_CTRL_BASE, SYS_CTRL_SIZE);
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 	pl011_init(&console_data, CONSOLE_UART_BASE,
 		CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-imx/main.c
+++ b/core/arch/arm/plat-imx/main.c
@@ -97,7 +97,7 @@ register_ddr(CFG_DRAM_BASE, CFG_DDR_SIZE);
 register_ddr(CFG_NSEC_DDR_1_BASE, CFG_NSEC_DDR_1_SIZE);
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 #ifdef CONSOLE_UART_BASE
 	imx_uart_init(&console_data, CONSOLE_UART_BASE);
--- a/core/arch/arm/plat-k3/main.c
+++ b/core/arch/arm/plat-k3/main.c
@@ -42,7 +42,7 @@ void boot_secondary_init_intc(void)
 	gic_init_per_cpu();
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 	serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
 			     CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-ls/main.c
+++ b/core/arch/arm/plat-ls/main.c
@@ -112,7 +112,7 @@ void plat_primary_init_early(void)
 }
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 #ifdef CFG_PL011
 	/*
--- a/core/arch/arm/plat-marvell/main.c
+++ b/core/arch/arm/plat-marvell/main.c
@@ -85,7 +85,7 @@ void boot_primary_init_intc(void)
 }
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 #if defined(PLATFORM_FLAVOR_armada7k8k)
 	serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
--- a/core/arch/arm/plat-mediatek/main.c
+++ b/core/arch/arm/plat-mediatek/main.c
@@ -33,7 +33,7 @@ void boot_primary_init_intc(void)
 }
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 	if (CFG_TEE_CORE_LOG_LEVEL != 0) {
 		serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
--- a/core/arch/arm/plat-nuvoton/main.c
+++ b/core/arch/arm/plat-nuvoton/main.c
@@ -60,7 +60,7 @@ void boot_primary_init_intc(void)
 	gic_init(GICC_BASE, GICD_BASE);
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 	ns16550_init(&console_data, CONSOLE_UART_BASE, IO_WIDTH_U32, 2);
 	register_serial_console(&console_data.chip);
--- a/core/arch/arm/plat-poplar/main.c
+++ b/core/arch/arm/plat-poplar/main.c
@@ -20,7 +20,7 @@ register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE, PL011_REG_SIZE);
 /* for dynamic shared memory */
 register_dynamic_shm(DRAM0_BASE_NSEC, DRAM0_SIZE_NSEC);
 
-void console_init(void)
+void plat_console_init(void)
 {
 	pl011_init(&console_data, CONSOLE_UART_BASE,
 		CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-rcar/main.c
+++ b/core/arch/arm/plat-rcar/main.c
@@ -65,7 +65,7 @@ static struct scif_uart_data console_data __nex_bss;
 uint32_t rcar_prr_value __nex_bss;
 #endif
 
-void console_init(void)
+void plat_console_init(void)
 {
 	scif_uart_init(&console_data, CONSOLE_UART_BASE);
 	register_serial_console(&console_data.chip);
--- a/core/arch/arm/plat-rockchip/main.c
+++ b/core/arch/arm/plat-rockchip/main.c
@@ -32,7 +32,7 @@ void boot_secondary_init_intc(void)
 	gic_init_per_cpu();
 }
 
-void console_init(void)
+void plat_console_init(void)
 {
 #if defined(CFG_EARLY_CONSOLE)
 	/*
--- a/core/arch/arm/plat-rpi3/main.c
+++ b/core/arch/arm/plat-rpi3/main.c
@@ -39,7 +39,7 @@ register_phys_mem_pgdir(MEM_AREA_IO_NSEC,
 
 static struct serial8250_uart_data console_data;
 
-void console_init(void)
+void plat_console_init(void)
 {
 	serial8250_uart_init(&console_data, CONSOLE_UART_BASE,
 			     CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
--- a/core/arch/arm/plat-rzg/main.c
+++ b/core/arch/arm/plat-rzg/main.c
@@ -28,7 +28,7 @@ register_dynamic_shm(NSEC_DDR_3_BASE, NSEC_DDR_3_SIZE);
 
 static struct scif_uart_data console_data __nex_bss;
 
-void console_init(void)
+void plat_console_init(void)
 {
 	scif_uart_init(&console_data, CONSOLE_UART_BASE);
 	register_serial_console(&console_data.chip);
--- a/core/arch/arm/plat-rzn1/main.c
+++ b/core/arch/arm/plat-rzn1/main.c
@@ -34,7 +34,7 @@ register_phys_mem(MEM_AREA_IO_SEC, GIC_BASE, CORE_MMU_PGDIR_SIZE);
 register_phys_mem(MEM_AREA_IO_SEC, PERIPH_REG_BASE, CORE_MMU_PGDIR_SIZE);
 register_ddr(DRAM_BASE, DRAM_SIZE);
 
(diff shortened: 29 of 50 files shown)


core/kernel/semihosting.c
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+
+#include <kernel/semihosting.h>
+#include <string.h>
+
```
jforissier, line 8: Please move the comment from `config.mk` to here:
```
/* 
 * ARM and RISC-V have defined the standard way to perform
 * the semihosting operations.
 * - Operation codes and open modes are identical.
 * - The implementation of the low-level __do_semihosting() call is architecture-specific.
 * - Arm semihosting interface:
 *   https://developer.arm.com/documentation/dui0471/g/Semihosting/The-semihosting-interface
 * - RISC-V semihosting interface:
 *   https://github.com/riscv-non-isa/riscv-semihosting/blob/main/binary-interface.adoc
 */
```

mk/config.mk
```diff
@@ -1159,3 +1159,13 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_HUK,CFG_DT))
...
 $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
+
+# Enable the semihosting functions.
+# The semihosting functions depend on architecture-supported semihosting
+# mechanism. Currently, ARM and RISC-V have defined the standard way to perform
+# the semihosting operations.
+# - Link for ARM's semihosting interface:
+#     https://developer.arm.com/documentation/dui0471/g/Semihosting/The-semihosting-interface
+# - Link for RISC-V's semihosting interface:
+#     https://github.com/riscv-non-isa/riscv-semihosting/blob/main/binary-interface.adoc
```
jforissier, line 1170: Replace with:
```
# Semihosting is a debugging mechanism that enables code running on an embedded
# system (also called the target) to communicate with and use the I/O of the
# host computer.
```

jforissier (commented):
- For "libutils: Move _ansi.h to include folder":
- For " libutils: Import part of sys/fcntl.h":
  Please mention you are importing this from newlib in the commit description. One more comment below.
- For "core: riscv: Add semihosting.S for semihosting instructions":
- For "core: kernel: Add semihosting functions": see comments below
- For "drivers: Implement semihosting based console driver for log":

It would be nice to be able to test this. Since the whole RISC-V QEMU environment is not ready yet, we should consider implementing `__do_semihosting()` for Arm. I may give it a shot ;)

mk/config.mk
gagachang: Done

core/kernel/semihosting.c
gagachang: Done

gagachang:
> It would be nice to be able to test this.

Sure, do you have any suggestion to test this feature ?

> Since the whole RISC-V QEMU environment is not ready yet, we should consider implementing `__do_semihosting()` for Arm. I may give it a shot ;)

Yes please! You may want to try this PR on ARM QEMUv8.
I would like to add your implementation for ARM into this series.

core/arch/riscv/kernel/sub.mk
```diff
@@ -12,6 +12,7 @@ srcs-y += thread_arch.c
 srcs-y += arch_scall_rv.S
 srcs-y += arch_scall.c
 srcs-$(CFG_UNWIND) += unwind_rv.c
+srcs-y += semihosting_rv.S
```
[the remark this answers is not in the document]
gagachang: Thanks, I've added it.

lib/libutils/isoc/include/sys/_default_fcntl.h
```diff
@@ -0,0 +1,72 @@
+#ifndef	_SYS__DEFAULT_FCNTL_H_
```
etienne-lms, line 1: this header and the other newlib imported file need a copyright/licence notice + an SPDX-License-Identifer tag. It's seems to me this is under BSD 2 or 3 clause license, copyrighted by RedHat and others: https://github.com/eblot/newlib/blob/master/COPYING.NEWLIB.

core/kernel/semihosting.c
```diff
@@ -0,0 +1,141 @@
...
+	SEMIHOSTING_OPEN_AX = 10,
+	SEMIHOSTING_OPEN_AXB = 11,
+};
+
+#if defined(O_BINARY)
+#define SEMIHOSTING_O_BINARY_MODE_OFFSET 1 /* binary mode */
+#endif
+
+enum semihosting_sys_ops {
+	// Regular operations
```
etienne-lms, line 42: replace `//` with `/* ... */`

core/kernel/semihosting.c
```diff
@@ -0,0 +1,141 @@
...
+{
+	__do_semihosting(SEMIHOSTING_SYS_WRITEC, (uintptr_t)&c);
+}
+
+int semihosting_open(const char *fname, int flags)
+{
+	int semi_open_flags = 0;
+	const int flags_mask = O_RDONLY | O_WRONLY | O_RDWR |
+			       O_CREAT | O_TRUNC | O_APPEND;
+	struct semihosting_param_t arg = {0};
```
etienne-lms, line 72: `= { }`

core/kernel/semihosting.c
```diff
@@ -0,0 +1,141 @@
...
+#if defined(O_BINARY)
+	if (flags & O_BINARY)
+		semi_open_flags += SEMIHOSTING_O_BINARY_MODE_OFFSET; /* 'b' */
+#endif
+
+	arg.param0 = (uintptr_t)fname;
+	arg.param1 = (uintptr_t)semi_open_flags;
+	arg.param2 = (uintptr_t)strlen(fname);
+
+	return __do_semihosting(SEMIHOSTING_SYS_OPEN, (uintptr_t)&arg);
```
etienne-lms, line 107: Explicitly cast return value from `uintptr_t` to `int`. I would also assert the retrun value is `< INT_MAX`.

core/kernel/semihosting.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+ssize_t semihosting_read(int fd, void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = (uintptr_t)fd,
+		.param1 = (uintptr_t)ptr,
+		.param2 = (uintptr_t)len
+	};
+
```
etienne-lms, line 117: explicit cast to signed `ssize_t`?

core/drivers/semihosting_console.c
```diff
@@ -0,0 +1,56 @@
...
+	return semihosting_readc();
+}
+
+static const struct serial_ops semihosting_console_ops = {
+	.putc = semihosting_console_putc,
+	.getchar = semihosting_console_getchar,
+};
+DECLARE_KEEP_PAGER(semihosting_console_ops);
+
+static void semihosting_console_fd_putc(struct serial_chip *chip __unused,
```
etienne-lms, line 27: remove `__unused`

core/include/drivers/semihosting_console.h
```diff
@@ -0,0 +1,25 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+#ifndef __DRIVERS_SEMIHOSTING_CONSOLE_H
+#define __DRIVERS_SEMIHOSTING_CONSOLE_H
+
+#include <drivers/serial.h>
+
+struct semihosting_console_data {
```
etienne-lms, line 10: an inline description comment is welcome

core/include/drivers/semihosting_console.h
```diff
@@ -0,0 +1,25 @@
...
+	struct serial_chip chip;
+	const char *file_path;
+	int fd;
+};
+
+/*
+ * Initialize console which uses architecture-specific semihosting mechanism.
+ * If "file_path" is not NULL, OP-TEE OS will try to output log to that file.
```
etienne-lms, line 18: s/"file_path"/@file_path/  (for consistency in API function inline description comments in optee_os)

core/kernel/semihosting.c
gagachang: I will also check other functions' return values refer to
https://developer.arm.com/documentation/dui0471/g/Semihosting/SYS-OPEN--0x01-?lang=en

core/kernel/semihosting.c
gagachang: but `len` should be greater or equal to 0?

core/include/drivers/semihosting_console.h
gagachang: OK will add it

core/include/drivers/semihosting_console.h
gagachang: Addressed in https://github.com/OP-TEE/optee_os/pull/6700/commits/c036e76844763d61d6abe980e9b7ac880993cd5d

core/include/drivers/semihosting_console.h
gagachang: Addressed in https://github.com/OP-TEE/optee_os/pull/6700/commits/c036e76844763d61d6abe980e9b7ac880993cd5d

core/drivers/semihosting_console.c
gagachang: Addressed in https://github.com/OP-TEE/optee_os/pull/6700/commits/c036e76844763d61d6abe980e9b7ac880993cd5d

core/kernel/semihosting.c
gagachang: https://github.com/OP-TEE/optee_os/pull/6700/commits/1723ab84f37982c7d2b60a6a4f3e24d267ba8438

core/kernel/semihosting.c
gagachang: Addressed in https://github.com/OP-TEE/optee_os/pull/6700/commits/1723ab84f37982c7d2b60a6a4f3e24d267ba8438

core/include/drivers/semihosting_console.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+#ifndef __DRIVERS_SEMIHOSTING_CONSOLE_H
+#define __DRIVERS_SEMIHOSTING_CONSOLE_H
+
+#include <drivers/serial.h>
+
+/*
+ * struct semihosting_console_data - Structure for semihosting console driver
+ * @chip - General structure for each serial chip
+ * @file_path - Path of the file on the semihosting host system
```
etienne-lms, line 13: `... or NULL if using semihosting console`

core/include/drivers/semihosting_console.h
```diff
@@ -0,0 +1,31 @@
...
+ * @fd - Handle of the file at @file_path
```
etienne-lms, line 14: `... or -1 if using the semihosting console`

core/include/kernel/semihosting.h
```diff
@@ -0,0 +1,37 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+#ifndef __KERNEL_SEMIHOSTING_H
+#define __KERNEL_SEMIHOSTING_H
+
+#include <util.h>
+
```
etienne-lms, line 9: can remove this empty line

gagachang: Addressed in https://github.com/OP-TEE/optee_os/pull/6700/commits/e29f81b15fac69c9111d5b172c7b85a1f1444ed7

core/include/drivers/semihosting_console.h
gagachang: Added into https://github.com/OP-TEE/optee_os/pull/6700/commits/f6bb8ab6267f21b773cd9cc8e48fa377f8288e48

core/include/drivers/semihosting_console.h
gagachang: Added into https://github.com/OP-TEE/optee_os/pull/6700/commits/f6bb8ab6267f21b773cd9cc8e48fa377f8288e48

lib/libutils/isoc/include/sys/_default_fcntl.h
gagachang: Add SPDX-License-Identifer tag in https://github.com/OP-TEE/optee_os/pull/6700/commits/c4e2b5c57f827f7ad3c66e9c05fd57c3d555a250

core/kernel/semihosting.c
```diff
@@ -0,0 +1,182 @@
...
+	uintptr_t param2;
+};
+
+/**
+ * @brief Read one character byte from the semihosting host
+ *
+ * @retval the character read from the semihosting host
+ */
+char semihosting_readc(void)
+{
```
etienne-lms, line 63: I think the function name is confusing: this is about semihosting debug interface if I'm right.
I suggest to rename these 2 functions to `semihosting_sys_readc()` and `semihosting_sys_writec()`,
or `semihosting_debug_readc()` and `semihosting_debug_writec()`,

And change here "semihosting host" to "semihosting host debug terminal" (if what I've just said makes sense).

core/kernel/semihosting.c
```diff
@@ -0,0 +1,182 @@
...
+ */
+void semihosting_writec(char c)
+{
+	__do_semihosting(SEMIHOSTING_SYS_WRITEC, (uintptr_t)&c);
+}
+
+/**
+ * @brief Request the semihosting host to open a file on the host system
+ * @param fname: the path or name of the file
+ * @param flags: the flags to open the file
```
etienne-lms, line 79: ` * @param flags: sys/fcntl.h standard flags to open the file with`

core/kernel/semihosting.c
```diff
@@ -0,0 +1,182 @@
...
+ * @param fd: a handle for a file previously opened
+ *
+ * @retval zero if OK, -1 if fails
+ */
+intptr_t semihosting_close(int fd)
+{
+	struct semihosting_param_t arg = {
+		.param0 = (uintptr_t)fd,
+		.param1 = 0,
+		.param2 = 0
```
etienne-lms, line 177: can remove these 2 lines.

lib/libutils/isoc/include/sys/_default_fcntl.h
```diff
@@ -0,0 +1,74 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
etienne-lms, line 1: We need a copyright here also. From newlib repository, it seems to me the one that applies here is:
`Copyright (c) 1994-2009  Red Hat, Inc. All rights reserved.`

core/include/kernel/semihosting.h
```diff
@@ -0,0 +1,36 @@
...
+/* Perform architecture-specific semihosting instructions. */
+uintptr_t __do_semihosting(uintptr_t op, uintptr_t arg);
+
+/* Read one character from the semihosting host. */
+char semihosting_readc(void);
+
+/* Write one character to the semihosting host. */
+void semihosting_writec(char c);
+
+/* Request the semihosting host to open a file on the host system. */
```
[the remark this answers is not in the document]
gagachang: OK I should remove comments in header.

core/kernel/semihosting.c
gagachang: > And change here "semihosting host" to "semihosting host debug terminal" (if what I've just said makes sense).

Yes, it makes sense.

core/include/kernel/semihosting.h
```diff
@@ -0,0 +1,36 @@
...
+
+/* Request the semihosting host to read data from a file on the host system. */
+size_t semihosting_read(int fd, void *ptr, size_t len);
+
+/* Request the semihosting host to write data into a file on the host system. */
+size_t semihosting_write(int fd, const void *ptr, size_t len);
+
+/*
+ * Request the semihosting host to close a file, which has been opened by
+ * semihosting_open(), on the host system.
```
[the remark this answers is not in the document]
gagachang: Hi @etienne-lms 

We don't do file operations for semihosting console case in semihosting console driver.

lib/libutils/isoc/include/sys/_default_fcntl.h
gagachang: Added in https://github.com/OP-TEE/optee_os/pull/6700/commits/69a53ff771d8ff4badb85a558f91212f1e85d8fc

core/kernel/semihosting.c
gagachang: Fixed

core/kernel/semihosting.c
gagachang: I've rename to `semihosting_sys_readc()` and `semihosting_sys_writec()`.

core/kernel/semihosting.c
gagachang: Removed

core/kernel/semihosting.c
```diff
@@ -0,0 +1,182 @@
...
+	SEMIHOSTING_OPEN_WX = 6,
+	SEMIHOSTING_OPEN_WXB = 7,
+	SEMIHOSTING_OPEN_A = 8,
+	SEMIHOSTING_OPEN_AB = 9,
+	SEMIHOSTING_OPEN_AX = 10,
+	SEMIHOSTING_OPEN_AXB = 11,
+};
+
+#if defined(O_BINARY)
+#define SEMIHOSTING_O_BINARY_MODE_OFFSET 1 /* binary mode */
```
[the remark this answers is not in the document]
gagachang: Done

gagachang:
Hi @jforissier @etienne-lms 

I have one question:
The semihosting console driver is compiled when `CFG_SEMIHOSTING_CONSOLE` is `y`.
Should I also put  `CFG_SEMIHOSTING_CONSOLE` into `mk/config.mk` ?
Besides, it has dependency on `CFG_SEMIHOSTING`

core/include/kernel/semihosting.h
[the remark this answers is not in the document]
etienne-lms: Right, OK, there is the word "file" in this description. 
Fine as-is. Thanks.

etienne-lms:
> Should I also put CFG_SEMIHOSTING_CONSOLE into mk/config.mk ?
> Besides, it has dependency on CFG_SEMIHOSTING

Yes, this is a good idea. This is a generic driver so the relate config switch deserve to be defined and documented in _mk/config.mk_.

gagachang:
Add dedicated https://github.com/OP-TEE/optee_os/pull/6700/commits/4b718117707e3dc7c825eb3cf47d4e42e06242f9 for new configurations: `CFG_SEMIHOSTING` and `CFG_SEMIHOSTING_CONSOLE`.

jforissier:
> > Should I also put CFG_SEMIHOSTING_CONSOLE into mk/config.mk ?
> > Besides, it has dependency on CFG_SEMIHOSTING
> 
> Yes, this is a good idea. This is a generic driver so the relate config switch deserve to be defined and documented in _mk/config.mk_.

I agree, but while we're at it, defining `CFG_SEMIHOSTING_CONSOLE=y` should be enough. Currently some platform-specific code is needed (for example QEMUv8 below). We should be able to avoid that.

```diff
commit e66d6b8fa8b52a0eaa87de755fc6b0f434e07306 (HEAD -> wip/semihosting, jf/wip/semihosting)
Author: Jerome Forissier <jerome.forissier@linaro.org>
Date:   Thu Feb 22 16:28:25 2024 +0100

    qemuv8: use semihosting cosole driver when CFG_SEMIHOSTING_CONSOLE=y

    Implement semihosting console for Aarch64 QEMU.

    Signed-off-by: Jerome Forissier <jerome.forissier@linaro.org>

diff --git a/core/arch/arm/plat-vexpress/main.c b/core/arch/arm/plat-vexpress/main.c
index 42116c2f2..a02017bc6 100644
--- a/core/arch/arm/plat-vexpress/main.c
+++ b/core/arch/arm/plat-vexpress/main.c
@@ -10,6 +10,7 @@
 #include <drivers/gic.h>
 #include <drivers/hfic.h>
 #include <drivers/pl011.h>
+#include <drivers/semihosting_console.h>
 #include <drivers/tzc400.h>
 #include <initcall.h>
 #include <keep.h>
@@ -27,6 +28,7 @@
 #include <trace.h>

 static struct pl011_data console_data __nex_bss;
+static struct semihosting_console_data sh_console_data __nex_bss;

 register_phys_mem_pgdir(MEM_AREA_IO_SEC, CONSOLE_UART_BASE, PL011_REG_SIZE);
 #if defined(PLATFORM_FLAVOR_fvp)
@@ -82,9 +84,14 @@ void boot_primary_init_intc(void)

 void console_init(void)
 {
-       pl011_init(&console_data, CONSOLE_UART_BASE, CONSOLE_UART_CLK_IN_HZ,
-                  CONSOLE_BAUDRATE);
-       register_serial_console(&console_data.chip);
+       if (IS_ENABLED(CFG_SEMIHOSTING_CONSOLE)) {
+               semihosting_console_init(&sh_console_data, "optee.log");
+               register_serial_console(&sh_console_data.chip);
+       } else {
+               pl011_init(&console_data, CONSOLE_UART_BASE,
+                          CONSOLE_UART_CLK_IN_HZ, CONSOLE_BAUDRATE);
+               register_serial_console(&console_data.chip);
+       }
 }

 #if (defined(CFG_GIC) || defined(CFG_CORE_HAFNIUM_INTC)) && \
```

jforissier:
@gagachang when adding fixup commits you should rather keep the subject of the original commit because it will help when it is time to rebase and squash the commits later at the end of the review. For example "drivers: Address comments for semihosting_console" could be "[Review]  drivers: Implement semihosting based console driver for log".

core/include/drivers/semihosting_console.h
```diff
@@ -0,0 +1,33 @@
...
+/*
+ * struct semihosting_console_data - Structure for semihosting console driver
+ * @chip - General structure for each serial chip
+ * @file_path - Path of the file on the semihosting host system or NULL if using
+ *              semihosting console
+ * @fd - Handle of the file at @file_path or -1 if using the semihosting console
+ */
+struct semihosting_console_data {
+	struct serial_chip chip;
+	const char *file_path;
```
jforissier, line 19: `file_path` is not needed it seems (set but never read)

gagachang: Yes, `fd` is enough. I can remove `file_path`.

gagachang:
> Currently some platform-specific code is needed (for example QEMUv8 below). We should be able to avoid that.

Maybe we can change the code in `entry_a64.S` ? 

```c
#ifdef CFG_SEMIHOSTING_CONSOLE
        bl	semihosting_console_init_helper    // Helper prepares console data and file_path
#else
	bl	console_init
#endif
```

jforissier:
> > Currently some platform-specific code is needed (for example QEMUv8 below). We should be able to avoid that.
> 
> Maybe we can change the code in `entry_a64.S` ?
> 
> ```c
> #ifdef CFG_SEMIHOSTING_CONSOLE
>         bl	semihosting_console_init_helper    // Helper prepares console data and file_path
> #else
> 	bl	console_init
> #endif
> ```

Or we could decide `console_init()` is common code in `core/kernel/console.c`. It would deal with semihosting (and possibly other things like `CFG_CBMEM_CONSOLE`), then the platform-specific `console_init()` functions would be renamed `plat_console_init()`.

gagachang:
> I successfully enabled the semihosting console for Aarch64 🎉

That’s great news!!
Hope this PR can help people who want to record OP-TEE log as file.

gagachang:
> Or we could decide `console_init()` is common code in `core/kernel/console.c`. It would deal with semihosting (and possibly other things like `CFG_CBMEM_CONSOLE`), then the platform-specific `console_init()` functions would be renamed `plat_console_init()`.

Hi @jforissier 

Should we add this changes into this series?

jforissier:
> > Or we could decide `console_init()` is common code in `core/kernel/console.c`. It would deal with semihosting (and possibly other things like `CFG_CBMEM_CONSOLE`), then the platform-specific `console_init()` functions would be renamed `plat_console_init()`.
> 
> Hi @jforissier
> 
> Should we add this changes into this series?

Yes, please propose something that works well for you and I will check it with QEMU/Armv8. Also, you can squash the fixup commits. Thanks!

gagachang:
> Yes, please propose something that works well for you and I will check it with QEMU/Armv8.

Hi @jforissier 

Please check the following three commits:
https://github.com/OP-TEE/optee_os/pull/6700/commits/07bc659bb4a789258a293d7ef3400b2bf99075a1
https://github.com/OP-TEE/optee_os/pull/6700/commits/db19c790fca39e3b21dd06e64fa59291efa758d4
https://github.com/OP-TEE/optee_os/pull/6700/commits/9db646fd0e55d7e3421cb38bb82cba4f7a530364

I didn't consider `CFG_CBMEM_CONSOLE`, because the initialization of CBMEM console needs device tree.
The device tree is initialized in `boot_init_primary_late()` -> `init_external_dt()`, which is later than `console_init()`.

mk/config.mk
```diff
@@ -1166,7 +1166,12 @@ CFG_SEMIHOSTING ?= n
...
 # from/to the file or debug terminal on the semihosting host computer.
+#   - If CFG_SEMIHOSTING_CONSOLE_FILE is not empty, output the characters to
+#     the specific file. Output to "optee.log" by default.
+#   - If CFG_SEMIHOSTING_CONSOLE_FILE is empty, output the characters to the
+#     semihosting host terminal.
 CFG_SEMIHOSTING_CONSOLE ?= n
 ifeq ($(CFG_SEMIHOSTING_CONSOLE),y)
 $(call force,CFG_SEMIHOSTING,y)
 endif
+CFG_SEMIHOSTING_CONSOLE_FILE ?= optee.log
```
jforissier, line 1177: It should be enough to do `CFG_SEMIHOSTING_CONSOLE_FILE ?= "optee.log"` (note the double quotes) and `semihosting_console_init(CFG_SEMIHOSTING_CONSOLE_FILE);` in this commit. For output to the host terminal then that would be `make CFG_SEMIHOSTING_CONSOLE_FILE=NULL`.

jforissier (commented):
@gagachang thanks for the update. 
- "core: drivers: Embed console data in semihosting console driver" looks good to me, it should ultimately be folded into "drivers: Implement semihosting based console driver for log".
- One comment on "core: kernel: Let user configure semihosting console output file". The commit should be folded into "drivers: Implement semihosting based console driver for log", too.
- "core: Refactor console_init() and introduce plat_console_init()" looks OK but it should come before "drivers: Implement semihosting based console driver for log".

lib/libutils/isoc/include/sys/_default_fcntl.h
```diff
@@ -0,0 +1,77 @@
...
```
etienne-lms, line 1: BSD-3-Clause seems more appropriate, as regards to the other **newlib** materials already present in optee_os source tree.

Ditto for _fcntl.h_.

core/kernel/semihosting.c
```diff
@@ -0,0 +1,181 @@
...
+	case O_RDWR | O_CREAT | O_APPEND:	/* 'a+' */
+		semi_open_flags = SEMIHOSTING_OPEN_AX;
+		break;
+	default:
+		return -1;
+	}
+
+#if defined(O_BINARY)
+	if (flags & O_BINARY)
+		semi_open_flags += SEMIHOSTING_O_BINARY_MODE_OFFSET; /* 'b' */
```
etienne-lms, line 117: s/`+=`/`|=`/

core/kernel/semihosting.c
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * @brief Request the semihosting host to open a file on the host system
+ * @param fname: the path or name of the file
+ * @param flags: sys/fcntl.h standard flags to open the file with
+ *
+ * @retval nonzero if OK, or -1 if fails
+ */
+intptr_t semihosting_open(const char *fname, int flags)
```
etienne-lms, line 84: `int` would better suit 0/-1 return code. (as expected by `fd` field in next commmit)
I think a `TEE_Result` code is more appropriate in the scope of OP-TEE core, but that's not a strong requirement.

Ditto for `semihosting_close()`.

mk/config.mk
gagachang: Wow it works! I will apply this

lib/libutils/isoc/include/sys/_default_fcntl.h
gagachang: OK

gagachang: Apply into https://github.com/OP-TEE/optee_os/pull/6700/commits/d2b49afed89a11fe3d257e08dbb41e879592d443

core/kernel/semihosting.c
gagachang: Fix in https://github.com/OP-TEE/optee_os/pull/6700/commits/5bce398196244e2e120e86fe8a0cc4dfb4345865

gagachang:
> * "core: drivers: Embed console data in semihosting console driver" looks good to me, it should ultimately be folded into "drivers: Implement semihosting based console driver for log".
> * One comment on "core: kernel: Let user configure semihosting console output file". The commit should be folded into "drivers: Implement semihosting based console driver for log", too.
> * "core: Refactor console_init() and introduce plat_console_init()" looks OK but it should come before "drivers: Implement semihosting based console driver for log".

I've folded them together and reordered the commits.

core/kernel/console.c
```diff
@@ -20,6 +21,11 @@ static struct serial_chip *serial_console __nex_bss;
 
 void console_init(void)
 {
+	if (IS_ENABLED(CFG_SEMIHOSTING_CONSOLE)) {
+		semihosting_console_init(CFG_SEMIHOSTING_CONSOLE_FILE);
+		return;
```
etienne-lms, line 26: nitpicking: prefer:

	if (IS_ENABLED(CFG_SEMIHOSTING_CONSOLE))
		semihosting_console_init(CFG_SEMIHOSTING_CONSOLE_FILE);
	else
		plat_console_init();

core/drivers/semihosting_console.c
```diff
@@ -0,0 +1,69 @@
...
+	.putc = semihosting_console_putc,
+	.getchar = semihosting_console_getchar,
+};
+DECLARE_KEEP_PAGER(semihosting_console_ops);
+
+static void semihosting_console_fd_putc(struct serial_chip *chip, int ch)
+{
+	struct semihosting_console_data *pd =
+		container_of(chip, struct semihosting_console_data, chip);
+
```
etienne-lms, line 46: suggestion:
```c
static void semihosting_console_fd_putc(struct serial_chip *chip __unused, int ch)
{
	if (sh_console_data.fd >= 0)
		semihosting_write(sh_console_data.fd, &ch, 1);
}
```
The test will prevent printing is the semihosting debug console if the open operation failed in `semihosting_console_init()`.

core/kernel/sub.mk
```diff
@@ -54,6 +54,8 @@ srcs-y += pseudo_ta.c
 
 srcs-$(CFG_TRANSFER_LIST) += transfer_list.c
 
+srcs-$(CFG_SEMIHOSTING) += semihosting.c
```
jforissier, line 57: Please add `CFG_SEMIHOSTING ?=n` from the last commit (as well as the associated comments) in "core: kernel: Add semihosting functions".

core/kernel/console.c
```diff
@@ -20,6 +21,11 @@ static struct serial_chip *serial_console __nex_bss;
 
 void console_init(void)
 {
+	if (IS_ENABLED(CFG_SEMIHOSTING_CONSOLE)) {
+		semihosting_console_init(CFG_SEMIHOSTING_CONSOLE_FILE);
```
jforissier, line 25: Please add these two CFG_ to `mk/config.mk` (i.e.e, move what you have in the last commit into "drivers: Implement semihosting based console driver for log").

jforissier (commented):
Looks good. Please move what you do in the last commit into the two previous ones as recommended below. There is no reason to configure things in a separate step.

etienne-lms (commented):
"libutils: Move _ansi.h to include folder",
"libutils: Move _ansi.h to include folder",
"core: kernel: Add semihosting functions",
"mk: config.mk: Add semihosting related configurations" (with comment on commit message).


Comments on commit message for "mk: config.mk: Add semihosting related configurations":
Can remove the enumeration `1.`, `2.`, `3.`.
Prfer replace the word 'kernel' with 'core'.

"core: riscv: Add semihosting.S for semihosting instructions".

core/drivers/semihosting_console.c
gagachang: Fix in https://github.com/OP-TEE/optee_os/pull/6700/commits/f542f6c9e8bb12571ee922bbb715b698c084ee4c

core/kernel/console.c
gagachang: Apply in https://github.com/OP-TEE/optee_os/pull/6700/commits/f542f6c9e8bb12571ee922bbb715b698c084ee4c#diff-49b28b57c451219e3e5f3f7b51994f0d424338fdb7b0134f2217bd44dfdd1081

core/arch/arm/kernel/boot.c
```diff
@@ -106,6 +106,11 @@ __weak void boot_secondary_init_intc(void)
 {
 }
 
+/* May be overridden in plat-$(PLATFORM)/main.c */
+__weak void plat_console_init(void)
+{
+}
```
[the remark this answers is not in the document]
gagachang: Change in https://github.com/OP-TEE/optee_os/pull/6700/commits/a424ceb0a02821653cf37384cd9c137ca16a8145

gagachang:
> Comments on commit message for "mk: config.mk: Add semihosting related configurations": Can remove the enumeration `1.`, `2.`, `3.`. Prfer replace the word 'kernel' with 'core'.

This commit is split and merge into previous commits respectively, suggested by @jforissier .

core/kernel/sub.mk
gagachang: Done https://github.com/OP-TEE/optee_os/pull/6700/commits/20b8d1656665c1ab1b41dc001421ec880ec7ec9b#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54a

core/kernel/console.c
gagachang: Done https://github.com/OP-TEE/optee_os/pull/6700/commits/f542f6c9e8bb12571ee922bbb715b698c084ee4c#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54a

core/kernel/semihosting.c
```diff
@@ -0,0 +1,179 @@
...
+		semi_open_flags = SEMIHOSTING_OPEN_WX;
+		break;
+	case O_RDWR | O_CREAT | O_APPEND:	/* 'a+' */
+		semi_open_flags = SEMIHOSTING_OPEN_AX;
+		break;
+	default:
+		return -1;
+	}
+
+#if defined(O_BINARY)
```
jforissier, line 115: This `#if` is not needed

jforissier (commented):
- Minor: I suggest applying "core: riscv: Add semihosting.S for semihosting instructions" after "core: kernel: Add semihosting functions".
- For "libutils: Import part of sys/fcntl.h":
- For "core: kernel: Add semihosting functions": one comment, with that:
- For "core: Refactor console_init() and introduce plat_console_init()":
- For "drivers: Implement semihosting based console driver for log":
You may remove the paragraph about semihosting and keep only "This commit implements a simple console driver...". Add it to commit description for "core: kernel: Add semihosting functions".

core/kernel/semihosting.c
gagachang: Removed in https://github.com/OP-TEE/optee_os/pull/6700/commits/2a17344798333873b0efca649c28d5c6f4be5ce8

gagachang:
> * Minor: I suggest applying "core: riscv: Add semihosting.S for semihosting instructions" after "core: kernel: Add semihosting functions".

Yes it makes sense. 

>   You may remove the paragraph about semihosting and keep only "This commit implements a simple console driver...". Add it to commit description for "core: kernel: Add semihosting functions".

Done, the paragraph is moved to "core: kernel: Add semihosting functions".

jforissier:
Tested on Aarch64 QEMU. I just had to add the `__do_semihosting()` implementation and a minor fix to the `plat-vexpress` console initialization code, see https://github.com/OP-TEE/optee_os/compare/1f6295e22c76b2caca3c2a4c05d55dfb1efb50bc...jforissier:optee_os:semihosting-aarch64-2
Good work @gagachang :wink:

core/kernel/console.c
```diff
@@ -24,7 +25,10 @@ __weak void plat_console_init(void)
 
 void console_init(void)
 {
-	plat_console_init();
+	if (IS_ENABLED(CFG_SEMIHOSTING_CONSOLE))
+		semihosting_console_init(CFG_SEMIHOSTING_CONSOLE_FILE);
```
jenswi-linaro, line 29: This may cause link errors if compiling without optimizations and `CFG_SEMIHOSTING_CONSOLE` isn't enabled.

core/kernel/semihosting.c
```diff
@@ -0,0 +1,169 @@
...
+ */
+size_t semihosting_read(int fd, void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = (uintptr_t)fd,
+		.param1 = (uintptr_t)ptr,
+		.param2 = (uintptr_t)len
+	};
+
+	return (size_t)__do_semihosting(SEMIHOSTING_SYS_READ, (uintptr_t)&arg);
```
jenswi-linaro, line 134: Why is the `(size_t)` cast needed?

core/kernel/semihosting.c
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @retval zero if OK, the same value as @len if fails, smaller value than @len
+ * for partial success
+ */
+size_t semihosting_read(int fd, void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = (uintptr_t)fd,
+		.param1 = (uintptr_t)ptr,
+		.param2 = (uintptr_t)len
```
jenswi-linaro, line 131: Why is the `(uintptr_t)` cast needed?

core/kernel/semihosting.c
```diff
@@ -0,0 +1,169 @@
...
+ * @param ptr: pointer to a buffer
+ * @param len: the number of bytes to read to the buffer from the file
+ *
+ * @retval zero if OK, the same value as @len if fails, smaller value than @len
+ * for partial success
+ */
+size_t semihosting_read(int fd, void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = (uintptr_t)fd,
```
jenswi-linaro, line 129: Why is the `(uintptr_t)` cast needed?

lib/libutils/isoc/include/sys/_default_fcntl.h
```diff
@@ -0,0 +1,77 @@
...
+/*
+ * Copyright (c) 1994-2009  Red Hat, Inc. All rights reserved.
+ */
+
+#ifndef	_SYS__DEFAULT_FCNTL_H_
+#ifdef __cplusplus
+extern "C" {
+#endif
+#define	_SYS__DEFAULT_FCNTL_H_
+#include <_ansi.h>
```
jenswi-linaro, line 11: Why is this file needed here? It defines a lot of macros that we don't use in OP-TEE. We should try to minimize the amount of legacy stuff introduced, considering that we're starting from scratch here.

jenswi-linaro (commented):
I'm not sure we want to do "libutils: Move _ansi.h to include folder"

core/kernel/console.c
etienne-lms: `CFG_SEMIHOSTING_CONSOLE_FILE` has a default value so it should work.
Maybe _mk/config.mk_ should check the config switch is not empty.

core/kernel/semihosting.c
etienne-lms: argument `fd` is signed and `param0` is not. There must be a cast here IMO.
Maybe not needed for `param1` and `param2`.

core/kernel/semihosting.c
etienne-lms: Explicit cast to `semihosting_read()` return value type. This was a suggestion of mine. No needed?

core/kernel/console.c
jenswi-linaro: I'm concerned with the lack of a `semihosting_console_init()` in that case.

core/kernel/semihosting.c
jenswi-linaro: When changing from unsigned to signed it might make sense, but between different unsigned types, it's only confusing.
Generally, I expect a cast only to make the compiler shut up or when it's needed for sign extension and such.

core/kernel/semihosting.c
jenswi-linaro: What happens if the cast is left out?

gagachang: Since valid `fd` always `>=0` here, maybe the cast can be removed ?

core/kernel/console.c
gagachang: Hi @jenswi-linaro 
Do you mean build optee_os with `CFG_CC_OPT_LEVEL=0` ?
I can see some items in CI build optee_os with `CFG_CC_OPT_LEVEL=0`
https://github.com/OP-TEE/optee_os/blob/7f124eb8587bd4fc11192ef970f4f34a26f19c21/.github/workflows/ci.yml#L114

jenswi-linaro: That's interesting. If it works then I don't object.

lib/libutils/isoc/include/sys/_default_fcntl.h
gagachang: I removed `_default_fcntl.h` and unnecessary definitions.
Please see https://github.com/OP-TEE/optee_os/pull/6700/commits/062625aa6b4881b71f4d075ad6e7d80d0ba308fc

core/kernel/semihosting.c
gagachang: Remove cast in https://github.com/OP-TEE/optee_os/pull/6700/commits/4e31cb8a2902ec97f6231debcc30249758d644e0

core/kernel/semihosting.c
gagachang: Remove cast in https://github.com/OP-TEE/optee_os/pull/6700/commits/4e31cb8a2902ec97f6231debcc30249758d644e0

core/kernel/semihosting.c
gagachang: Remove cast in https://github.com/OP-TEE/optee_os/pull/6700/commits/4e31cb8a2902ec97f6231debcc30249758d644e0

gagachang:
> I'm not sure we want to do "libutils: Move _ansi.h to include folder"

I remove this commit since we only copy necessary definitions into `sys/fcntl.h`

core/kernel/semihosting.c
```diff
@@ -0,0 +1,169 @@
...
+		break;
+	case O_RDWR | O_CREAT | O_APPEND:	/* 'a+' */
+		semi_open_flags = SEMIHOSTING_OPEN_AX;
+		break;
+	default:
+		return -1;
+	}
+
+	arg.param0 = (uintptr_t)fname;
+	arg.param1 = (uintptr_t)semi_open_flags;
```
jenswi-linaro, line 111: Casting isn't needed here or on the line below.

core/include/kernel/semihosting.h
```diff
@@ -0,0 +1,22 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+#ifndef __KERNEL_SEMIHOSTING_H
+#define __KERNEL_SEMIHOSTING_H
+
+#include <sys/fcntl.h>
+#include <sys/types.h>
```
etienne-lms, line 9: _sys/types.h_ is not strictly needed, only _stdint.h_ for `uintptr_t` and _stddef.h_ for `size_t`.

gagachang: Changed in https://github.com/OP-TEE/optee_os/pull/6700/commits/37c6df05cd56ba3e16b9c88e011d0517bc89d49c

core/kernel/semihosting.c
gagachang: Removed in https://github.com/OP-TEE/optee_os/pull/6700/commits/37c6df05cd56ba3e16b9c88e011d0517bc89d49c

core/include/drivers/semihosting_console.h
```diff
@@ -0,0 +1,17 @@
...
+#define __DRIVERS_SEMIHOSTING_CONSOLE_H
+
+/*
+ * Initialize console which uses architecture-specific semihosting mechanism.
+ * If @file_path is not NULL, OP-TEE OS will try to output log to that file,
+ * which is on the semihosting host system.
+ * Otherwise, if @file_path is NULL, OP-TEE OS will try to output log to the
+ * semihosting host debug console.
+ */
+void semihosting_console_init(const char *file_path);
```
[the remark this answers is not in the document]
gagachang: Added in https://github.com/OP-TEE/optee_os/pull/6700/commits/24d064547f12f1dfff9e23f037cc475e093f7a1c

core/kernel/console.c
```diff
@@ -17,6 +18,19 @@
 
 static struct serial_chip *serial_console __nex_bss;
 
+/* May be overridden in plat-$(PLATFORM)/main.c */
```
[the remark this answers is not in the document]
gagachang: I see. Changed in https://github.com/OP-TEE/optee_os/pull/6700/commits/4b51f340c75f55e3076a22f578027ec7c1cb7fdc#diff-49b28b57c451219e3e5f3f7b51994f0d424338fdb7b0134f2217bd44dfdd1081

jforissier:
My review/ack tags still apply.

core/kernel/semihosting.c
```diff
@@ -0,0 +1,169 @@
...
+ * @param ptr: pointer to a buffer
+ * @param len: the number of bytes to read to the buffer from the file
+ *
+ * @retval zero if OK, the same value as @len if fails, smaller value than @len
+ * for partial success
+ */
+size_t semihosting_read(int fd, void *ptr, size_t len)
+{
+	struct semihosting_param_t arg = {
+		.param0 = fd,
```
etienne-lms, line 129: Assigning a signed `int` to an unsigned `uintptr_t` do not deserve an explicit cast?

Same question for lines 148 and 165.

core/kernel/semihosting.c
etienne-lms: comment already  resolved

mk/config.mk
```diff
@@ -1158,3 +1158,20 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_HUK,CFG_DT))
...
+# from/to the file or debug terminal on the semihosting host computer.
+#   - Specify CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}" to output the
+#     characters to that file. Output to "optee.log" by default.
+#   - Specify CFG_SEMIHOSTING_CONSOLE_FILE=NULL to output the characters to the
+#     semihosting host terminal.
+CFG_SEMIHOSTING_CONSOLE ?= n
+ifeq ($(CFG_SEMIHOSTING_CONSOLE),y)
+$(call force,CFG_SEMIHOSTING,y)
+endif
+CFG_SEMIHOSTING_CONSOLE_FILE ?= "optee.log"
```
etienne-lms, line 1177: to prevent a build error if one sets CFG_SEMIHOSTING_CONSOLE_FILE to an empty value, I suggest to add 
```c
ifeq ($(CFG_SEMIHOSTING_CONSOLE_FILE),)
$(error CFG_SEMIHOSTING_CONSOLE_FILE cannot by empty)
endif
```
It will fault on no-value case as well as on empty string case.

mk/config.mk
```diff
@@ -1158,3 +1158,20 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_HUK,CFG_DT))
...
 $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
+
+# Semihosting is a debugging mechanism that enables code running on an embedded
+# system (also called the target) to communicate with and use the I/O of the
+# host computer.
+CFG_SEMIHOSTING ?= n
+
+# Enable the semihosting console driver which inputs/outputs the characters
+# from/to the file or debug terminal on the semihosting host computer.
+#   - Specify CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}" to output the
```
etienne-lms, line 1169: nitpicking: I think it would better without the indentation:
```
# Specify CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}" to output the
# characters to that file. Output to "optee.log" by default.
# Specify CFG_SEMIHOSTING_CONSOLE_FILE=NULL to output the characters to the
# semihosting host terminal.

File: core/kernel/semihosting.c (hunk 42 above)
Comment (jenswi-linaro) [reply core/kernel/semihosting.c:129]: Casting it to `uintptr_t` will not change anything. It depends on what you want to achieve.
If `fd` is `-1` then `.param0` will become `0xffffffffffffffff` on a 64-bit platform. If you want to use only the lower 32 bits then you would need to cast it as `(unsigned int)`, but I don't suppose that's the case here.

File: mk/config.mk (hunk 44 above)
Comment (gagachang) [reply mk/config.mk]: Remove two spaces in https://github.com/OP-TEE/optee_os/pull/6700/commits/35cb6b05f9ef5b50df37c0a3c543028292f82c43

**Comment by gagachang** [comment]
Fix commit message of `libutils: Import part of sys/fcntl.h` and apply review tag.

File: mk/config.mk [hunk 45]
```diff
@@ -1163,3 +1163,18 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
...
+
+# Enable the semihosting console driver which inputs/outputs the characters
+# from/to the file or debug terminal on the semihosting host computer.
+# - Specify CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}" to output the
+#   characters to that file. Output to "optee.log" by default.
+# - Specify CFG_SEMIHOSTING_CONSOLE_FILE=NULL to output the characters to the
+#   semihosting host terminal.
+CFG_SEMIHOSTING_CONSOLE ?= n
+ifeq ($(CFG_SEMIHOSTING_CONSOLE),y)
+$(call force,CFG_SEMIHOSTING,y)
```
etienne-lms, line 1175: This does not work. `make CFG_SEMIHOSTING_CONSOLE=y` fails:
```
$ make -j8 CFG_SEMIHOSTING_CONSOLE=y
mk/config.mk:1175: *** CFG_SEMIHOSTING is set to 'n' (from file) but its value must be 'y'.  Stop.
$
```
You need to default set `CFG_SEMIHOSTING ?= n` after this line:
```
# Enable the semihosting console driver which inputs/outputs the characters
# from/to the file or debug terminal on the semihosting host computer.
# - Specify CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}" to output the
#   characters to that file. Output to "optee.log" by default.
# - Specify CFG_SEMIHOSTING_CONSOLE_FILE=NULL to output the characters to the
#   semihosting host terminal.
CFG_SEMIHOSTING_CONSOLE ?= n
ifeq ($(CFG_SEMIHOSTING_CONSOLE),y)
$(call force,CFG_SEMIHOSTING,y)
endif
(...)

# Semihosting is a debugging mechanism that enables code running on an embedded
# system (also called the target) to communicate with and use the I/O of the
# host computer.
CFG_SEMIHOSTING ?= n

File: mk/config.mk [hunk 46]
```diff
@@ -1163,3 +1163,18 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
 # system (also called the target) to communicate with and use the I/O of the
 # host computer.
 CFG_SEMIHOSTING ?= n
+
+# Enable the semihosting console driver which inputs/outputs the characters
+# from/to the file or debug terminal on the semihosting host computer.
+# - Specify CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}" to output the
+#   characters to that file. Output to "optee.log" by default.
```
etienne-lms, line 1170: s/characters/trace messages/g    (lines 1167, 1170 & 1171)

etienne-lms (commented):
"libutils: Import part of sys/fcntl.h".

"core: Refactor console_init() and introduce plat_console_init()".

Comments for commit "drivers: Implement semihosting based console driver for log": we prefer have the commit message at the imperative mode:
```patch
-This commit implements a simple console driver which uses semihosting
+Implement a simple console driver which uses...
```

A equivalent (and late) comment for commit "core: riscv: Add semihosting.S for semihosting instructions":
```patch
 RISC-V architecture has defined the semihosting binary interface, which
 consists of a special trap instruction sequence, in:
 https://github.com/riscv-non-isa/riscv-semihosting
 
-This commit adds semihosting.S into RISC-V architecture folder and
-implements the trap instruction sequence.
+Add semihosting.S into RISC-V kernel folder to implement the trap
+instruction sequence.

File: mk/config.mk (hunk 46 above)
Comment (gagachang) [reply mk/config.mk]: Changed in https://github.com/OP-TEE/optee_os/pull/6700/commits/4779839d926e158c5384ff3c9bb55b87df9bede5#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54a

File: mk/config.mk (hunk 45 above)
Comment (gagachang) [reply mk/config.mk]: Fixed in https://github.com/OP-TEE/optee_os/pull/6700/commits/4779839d926e158c5384ff3c9bb55b87df9bede5#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54a
Thanks!

File: mk/config.mk [hunk 47]
```diff
@@ -1159,6 +1159,21 @@ $(eval $(call cfg-depends-all,CFG_WIDEVINE_HUK,CFG_DT))
 CFG_WIDEVINE_PTA ?= n
 $(eval $(call cfg-depends-all,CFG_WIDEVINE_PTA,CFG_DT CFG_WIDEVINE_HUK))
 
+# Enable the semihosting console driver which inputs/outputs the trace messages
+# from/to the file or debug terminal on the semihosting host computer.
```
etienne-lms, line 1163: Sorry nitpicking:  input console is supported only when using the semihosting debug console.
Proposal:
```
# CFG_SEMIHOSTING_CONSOLE, when enabled, embeds a semihosting console driver.
# When CFG_SEMIHOSTING_CONSOLE_FILE=NULL, OP-TEE console reads/writes
# characters from/to the debug terminal of the semihosting host computer.
# When CFG_SEMIHOSTING_CONSOLE_FILE="{your_log_file}", OP-TEE console
# outputs trace messages to that file. Output to "optee.log" by default.

etienne-lms (commented):
Comment for commit 

The commit message could be update accordingly:
```patch
 - If the caller of semihosting_console_init() provides the path of the
   file, the driver will try to open that file, and output the log to
   that host side file.
 - If the caller of semihosting_console_init() does not provide the path
+  of the file, the driver will try to output the log to host side
-  of the file, the driver will connect the console to the host debug
   console directly.

mk/config.mk
gagachang: I changed `characters` to `trace messages`.
Please see https://github.com/OP-TEE/optee_os/pull/6700/commits/bd19a9ba6609366f28a70c8ec16a90cd5267ab81

gagachang: Hi @etienne-lms 
Would you like to review this again?

jforissier:
@gagachang thanks!

PR #6707, opened 2024-02-22
plat-stm32mp2: add pm support on stm32mp25 

patrickdelaunay:
This series add a preliminary support of low power mode on STM32MP25 SoC, including power off,
and perform the needed OP-TEE bookkeeping before PSCI stack in TF-A BL31 executes a power management sequence.

The pm function for STM32MP25 drivers  will use the generic HINTs to save the IP configuration 
(PM_HINT_CONTEXT_STATE) or stop their clock (PM_HINT_CLOCK_STATE).

Only the STPMIC25 driver should use the power level information, saved in the PLATFORM_STATE part of HINT 
(bit 16:31), to configure in the PMIC the requested supply configuration for the select low power mode.

So this series provide the needed macros, PM_HINT_IS_STATE()  and defines PM_.*_LEVEL
to implement the pm functions in STM32MP25 driver and in PMIC (STPMIC25) driver in next patches.

Regards

Patrick.

### DIFF
--- a/core/include/kernel/pm.h
+++ b/core/include/kernel/pm.h
@@ -30,6 +30,12 @@
 #define PM_HINT_PLATFORM_STATE_MASK	GENMASK_32(31, 16)
 #define PM_HINT_PLATFORM_STATE_SHIFT	U(16)
 
+#define PM_HINT_STATE(_x)		((_x) & ~PM_HINT_PLATFORM_STATE_MASK)
+#define PM_HINT_PLATFORM_STATE(_x) \
+	(((_x) & PM_HINT_PLATFORM_STATE_MASK) >> PM_HINT_PLATFORM_STATE_SHIFT)
+
+#define PM_HINT_IS_STATE(_x, _name) ((_x) & PM_HINT_ ## _name ## _STATE)
+
 /*
  * PM_OP_SUSPEND: platform is suspending to a target low power state
  * PM_OP_RESUME: platform is resuming from low power state
--- a/core/arch/arm/plat-stm32mp2/stm32mp_pm.c
+++ b/core/arch/arm/plat-stm32mp2/stm32mp_pm.c
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2023-2024, STMicroelectronics
+ */
+
+#include <kernel/misc.h>
+#include <kernel/pm.h>
+#include <kernel/thread.h>
+#include <sm/psci.h>
+#include <stm32mp_pm.h>
+
+/**
+ * @brief   Handler for system off
+ *
+ * @param[in] a0   Unused
+ * @param[in] a1   Unused
+ *
+ * @retval 0       if OK, other value else and TF-A will panic
+ */
+unsigned long thread_system_off_handler(unsigned long a0 __unused,
+					unsigned long a1 __unused)
+{
+	/*
+	 * configure targeted mode in PMIC for system OFF,
+	 * no need to save context
+	 */
+	uint32_t pm_hint = PM_HINT_CLOCK_STATE |
+		((PM_MAX_LEVEL << PM_HINT_PLATFORM_STATE_SHIFT) &
+		  PM_HINT_PLATFORM_STATE_MASK);
+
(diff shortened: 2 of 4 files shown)


core/include/kernel/pm.h
```diff
@@ -30,6 +30,12 @@
 #define PM_HINT_PLATFORM_STATE_MASK	GENMASK_32(31, 16)
 #define PM_HINT_PLATFORM_STATE_SHIFT	U(16)
 
+#define PM_HINT_STATE(x)		((x) & ~PM_HINT_PLATFORM_STATE_MASK)
+#define PM_HINT_PLATFORM_STATE(x) \
+	(((x) & PM_HINT_PLATFORM_STATE_MASK) >> PM_HINT_PLATFORM_STATE_SHIFT)
+
+#define PM_HINT_IS_STATE(x, name) (!!((x) & PM_HINT_ ## name ## _STATE))
```
etienne-lms, line 37: Can remove `!!` operation, `((x) & PM_HINT_ ## name ## _STATE)` is fine.

Prefer use `_x` instead of `x` and `_name` instead of `name` as macro arguments names. This prevent possible collisions with C variables labeled `x` or `name` visible in the scope where these macro are used.

etienne-lms (commented):
Proposal rephrasing of the in commit message, (there are several macros defined in this change):
_"Add helper macros to read and test the power state hints provided by the platform during power management state transitions."_

core/include/kernel/pm.h
patrickdelaunay: !! => removed (bad habit for pre C99 compiler recommendation)

for macro parameter: 'x' / 'name' => '_x' / '_name'
done even if I am surprised by this recommendation for a simple macro parameter
and it is not used in lot of existing code.....

for me, the "x" parameter can't have collision with local variable, 
because the local variable are parsed after pre-processing

I understood this need only for local variables defined when MACRO generate code

#define  INIT_ARRAY(tab) for (int i=0, _i <ARRAY_SIZE(tab); i++) { tab[i] = 0}

 int my_array[3];
 bool i = false;
 long tab[10];

 INIT_ARRAY(my_array) ;

=> this code cause conflict with variable in code generated with macro

in this case it is recommended to use "_i" instead of "i" for local variable (but no issue for parameter)

patrickdelaunay:
> Add helper macros to read and test the power state hints provided by the platform during power management state transitions.

done

core/include/kernel/pm.h
etienne-lms: My bad. You're right. I messed-up with some labels constraints on macro implementation.
My apologies. Thanks for the details.

core/include/kernel/pm.h
```diff
@@ -30,6 +30,12 @@
 #define PM_HINT_PLATFORM_STATE_MASK	GENMASK_32(31, 16)
 #define PM_HINT_PLATFORM_STATE_SHIFT	U(16)
 
+#define PM_HINT_STATE(_x)		((_x) & ~PM_HINT_PLATFORM_STATE_MASK)
+#define PM_HINT_PLATFORM_STATE(_x) \
+	(((_x) & PM_HINT_PLATFORM_STATE_MASK) >> PM_HINT_PLATFORM_STATE_SHIFT)
+
+#define PM_HINT_IS_STATE(_x, _name) (((_x) & PM_HINT_ ## _name ## _STATE))
```
etienne-lms, line 37: can drop a pair of outer parentheses.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with a minor comment.

jenswi-linaro:
Please add a `core: ` prefix to `pm: add macro for PM_HINT_STATE access`

patrickdelaunay:
> Please add a `core: ` prefix to `pm: add macro for PM_HINT_STATE access`

done

jforissier:
@patrickdelaunay would you mind also adding Etienne's `Reviewed-by:` tag, so that I can merge this with no manual work :wink: thanks!

patrickdelaunay:
> Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>

Added one the 2 commits of the serie
Sorry

PR #6709, opened 2024-02-23
RISC-V: Fix creation of PTE when OP-TEE frees the pages

gagachang:
In current implementation, we always set `PTE_V` bit when we create PTE by `core_mmu_pte_create()`.
But when OP-TEE wants to free the pages, we should set zero value into PTE.

This PR fixes this behavior.
The desired PTE bits are already translated from the memory attributes.
Therefore, it is unnecessary to explicitly set `PTE_V` bit in `core_mmu_pte_create()`.


gagachang:
Hi @maroueneboubakri 

Please review this PR~ thanks!

maroueneboubakri:
@gagachang thanks !

gagachang:
Tag applied, thank!

PR #6714, opened 2024-02-26
.gitignore: Ignore all dot files and folder except the existing ones

Upsylonbare:
Hello,
While coding, some IDE creates folder to stores tags or project settings.
Some developers even have development tools that can be configured by project once they find a specific dot file in the project source directory.
This patch aims to leave developer do what they want to do with their tools and project configuration by ignoring all dot files and dot folders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,14 @@
-/.cproject
 /compile_commands.json
 /cscope.*
 /tags
 /TAGS
 /out
 /.project
-*.swp
+# Ignore all dot files and folder
+.*
+# But not those
+!.clang-format
+!.checkpatch.conf
+!.gitattributes
+!.gitignore
+!.github
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
-.cproject
-compile_commands.json
-cscope.*
-tags
-TAGS
-out
-.project
+/.cproject
+/compile_commands.json
+/cscope.*
+/tags
+/TAGS
+/out
+/.project
 *.swp


jforissier (commented):
LGTM, but there is something slightly wrong with all the entries in `.gitconfig` (the older ones and the newer ones as well): they should all start with a forward slash so that only files from the current directory are considered. Would you want to add a commit to fix that first?

Upsylonbare:
Do you mean `.gitignore` ?

jforissier:
> Do you mean `.gitignore` ?

Yes, sorry.

Upsylonbare:
Sorry I forget to remove the Change-Id line in the commit message

Upsylonbare:
Should be better

jforissier:

Thanks!

.gitignore
```diff
@@ -1,8 +1,14 @@
...
-/.project
 /*.swp
+# Ignore all dot files and folder
+.*
+# But not those
+!.clang-format
+!.checkpatch.conf
+!.gitattributes
+!.gitignore
+!.github
```
jenswi-linaro, line 14: Please preserve the end-of-line

Upsylonbare: Hello, yes will do !

.gitignore
```diff
@@ -1,8 +1,8 @@
...
-.project
-*.swp
+/.cproject
+/compile_commands.json
+/cscope.*
+/tags
+/TAGS
+/out
+/.project
+/*.swp
```
etienne-lms, line 6: \*.swp\ should apply to all subdirectories.

etienne-lms (commented):
In commit message for ".gitignore: Ignore all dot files and folder except the existing ones":
In header line:
s/folder/folders/
suggestion: s/existing ones/standard ones/

In message body:

s/as IDE/as IDEs/ or s/as IDE/as an IDE/.
s/file or folder/files or folders/
s/bother/bothered/

.gitignore
jforissier: I agree

Upsylonbare: Indeed, but I have considered that ignoring all dot files in the second commit rewrite this ignore rule.
I propose to remove the forward slash in the first commit and then delete the *.swp line in the second one.

etienne-lms: I would not remove the `*.swp` entry. It was already there and is quite convivient to **vi**-like tools users.

Upsylonbare: Yes of course but the purpose of this PR is to ignore all dot files whatever your preferred text editor. 
To me, keeping the entry `*.swp` alongside the entry `.*` doesn't have so much sense.

etienne-lms: _*.swp_ files are not dot file. So drop legacy? I would like it to be preserved, but if our dear maintainers thinks it's not worth it, then fine :cry:

etienne-lms: my bad, there are. Fine :smiley:

etienne-lms: Note: for first commit to be consistent, it could straight remove the `*.swp` line. `/*.swp` is not very useful to anybody.

jforissier: First step is to add `/` where it makes sense, so NOT on the `*.swp` line. Then the second step adds more `.*` files so it can drop `*.swp` (which should actually have been `.*.swp` because vi does prepend a dot to the file name).

jforissier (commented):
"except for **vi** swap files" perhaps? Otherwise looks fine to me.

Upsylonbare:
I wanted to be generic as my *not prefered text editor nano also create swp file.

jforissier:
> I wanted to be generic as my *not prefered text editor nano also create swp file.

Ah! Fine. Maybe "editor swap files" then?

jenswi-linaro:

etienne-lms (commented):

PR #6718, opened 2024-02-28
plat-hikey: make DRAM0_SIZE and DRAM0_SIZE_NSEC configurable

zwb-233:
This commit aims to enable the dynamic adjustment of DRAM0_SIZE and DRAM0_SIZE_NSEC for Hikey960 based on the changes in CFG_TZDRAM_START.


core/arch/arm/plat-hikey/platform_config.h
```diff
@@ -105,7 +105,7 @@
  */
 
 #define DRAM0_BASE		0x00000000
-#define DRAM0_SIZE		0x3F000000
+#define DRAM0_SIZE		CFG_TZDRAM_START
```
jforissier, line 108: size = start? This doesn't seem to make much sense does it?

core/arch/arm/plat-hikey/platform_config.h
```diff
@@ -130,7 +130,7 @@
 #if (CFG_DRAM_SIZE_GB == 3)
 #define DRAM1_SIZE_NSEC		0x80000000
 #elif (CFG_DRAM_SIZE_GB == 4)
-#define DRAM1_SIZE_NSEC		0xA0000000
+#define DRAM1_SIZE_NSEC		0xE0000000 - DRAM1_BASE
```
etienne-lms, line 133: parentheses needed

core/arch/arm/plat-hikey/platform_config.h
jforissier: You're right, sorry I did not see that. Would you mind adding a comment `/* DRAM0 exists below TZDRAM */`?

etienne-lms: For consistency and for future new readers, prefer: `#define DRAM0_SIZE (CFG_TZDRAM_START - DRAM0_BASE)` even if the result is the same in the generated image.

PR #6719, opened 2024-02-28
Notification based test watchdog

jenswi-linaro:
This introduces a test watchdog based on notifications. The primary purpose is to later be able to exercise asynchronous notifications with ns-virtualization, but to I'm trying a smaller scope to help the review.

The test watchdog is based on the introduced callout service which handles calling a callback at a certain timeout or period. The callout service is currently driven by the secure generic timer, but can easily be driven by another timer if needed.


jenswi-linaro:
It took a while, but CI is happy with the PR now.

etienne-lms (commented):
Comments for commit "core: add callout service".
Out of curiosity, why did you choose _callout_ naming for this feature. It is about delayed calls, right?

core/arch/arm/kernel/generic_timer.c
```diff
@@ -0,0 +1,65 @@
...
+	write_cntps_ctl(1);
+}
+
+static uint64_t
+timer_ms_to_ticks(const struct callout_timer_desc *desc __unused,
+		  uint32_t timeout_ms)
+{
+	uint64_t freq = read_cntfrq();
+
+	return (freq * timeout_ms) / 1000;
```
etienne-lms, line 28: could use `arm_cnt_us2cnt()` from _core/arch/arm/kernel/delay_arm.h_

core/arch/arm/plat-vexpress/main.c
```diff
@@ -255,3 +256,13 @@ int psci_cpu_on(uint32_t core_id, uint32_t entry, uint32_t context_id)
...
 #endif /*PLATFORM_FLAVOR_qemu_virt*/
+
+#if defined(IT_SEC_PHY_TIMER) && !defined(CFG_CORE_SEL2_SPMC)
+static TEE_Result init_callout_service(void)
+{
+	timer_init_callout_service(interrupt_get_main_chip(), IT_SEC_PHY_TIMER);
+
+	return TEE_SUCCESS;
+}
+nex_early_init(init_callout_service);
```
etienne-lms, line 267: add an empty line above.

etienne-lms (commented):
For commit "core: tests: add a notification test watchdog": the commit message is not very verbose about the test, what's expected etc.. Could you add few details?

jenswi-linaro:
> Comments for commit "core: add callout service". Out of curiosity, why did you choose _callout_ naming for this feature. It is about delayed calls, right?

It's something I hacked many years ago, initially inspired by https://github.com/freebsd/freebsd-src/blob/main/sys/sys/callout.h but rewritten so much that only the name remained. If you can think of a better **and** shorter name I don't mind renaming it.

core/include/kernel/callout.h
```diff
@@ -0,0 +1,112 @@
...
+ * struct callout - callout reference
+ * @callback:	function to be called when a callout expires, internal use only
+ * @ctrval:	callout expiry time counter value, internal use only
+ * @ms:		milliseconds to next timeout
+ * @link:	linked list element, internal use only
+ *
+ * @callback is called from an interrupt handler so thread resources must
+ * not be used. The main callout service lock is held while @callback is
+ * called so callout_rem() and callout_add() can't be used, but it is safe
+ * to modify @ms if the call period should be changed. @callback returns
```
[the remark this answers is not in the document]
jenswi-linaro: Makes sense, that's a nice simplification.

core/include/kernel/callout.h
```diff
@@ -0,0 +1,112 @@
...
+};
+
+/*
+ * callout_add() - Add a callout
+ * @co:		callout reference
+ * @callback:	callback function accociated with the callout
+ * @ms:		time to next callout in milliseconds
+ *
+ * Adds a callout to the callout service with an associated callback
+ * function @callout should be assigned and will be called in @ms
```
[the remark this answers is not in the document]
jenswi-linaro: I'll update, thanks.

core/include/kernel/callout.h
```diff
@@ -0,0 +1,112 @@
...
+ * milliseconds.
+ *
+ * If callout_add() is called before callout_service_init() has been called
+ * then it will be called @ms milliseconds after callout_service_init() has
+ * been called.
+ *
+ * The callout structure can reside in global data or on the heap. It's
+ * safe to embed it inside another struct, but it must not be freed until
+ * removed again with callout_rem() or equivalent.
```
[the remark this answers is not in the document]
jenswi-linaro: Yes

core/include/kernel/callout.h
```diff
@@ -0,0 +1,112 @@
...
+	uint64_t (*get_now)(const struct callout_timer_desc *desc);
+};
+
+/*
+ * callout_service_init() - Initialize the callout service
+ * @desc:	Pointer to the timer interface
+ *
+ * The callout service is initialized with the supplied timer interface
+ */
+void callout_service_init(const struct callout_timer_desc *desc);
```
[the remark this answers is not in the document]
jenswi-linaro: Sure. Would you mind if that's something we add at a later stage?

core/kernel/callout.c
```diff
@@ -0,0 +1,168 @@
...
+		if (co->ctrval < co2->ctrval) {
+			TAILQ_INSERT_BEFORE(co2, co, link);
+			return;
+		}
+	}
+
+	TAILQ_INSERT_TAIL(&callout_head, co, link);
+}
+
+static void schedule_next_timeout(const struct callout_timer_desc *desc)
```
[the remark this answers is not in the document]
jenswi-linaro: OK

core/kernel/callout.c
```diff
@@ -0,0 +1,168 @@
...
+	const struct callout_timer_desc *desc = callout_desc;
+	struct callout *co = NULL;
+	bool do_callout = false;
+	uint64_t now = 0;
+
+	/*
+	 * schedule_next_timeout() saves the core it was last called on.
+	 * If if there's a mismatch here it means that another core has be
+	 * scheduled for the next callout, so there's no work to be done
+	 * for this core.
```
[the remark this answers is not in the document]
jenswi-linaro: Good point, I'll update.

core/include/kernel/callout.h
```diff
@@ -0,0 +1,112 @@
...
+#define __KERNEL_CALLOUT_H
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <sys/queue.h>
+
+/*
+ * struct callout - callout reference
+ * @callback:	function to be called when a callout expires, internal use only
+ * @ctrval:	callout expiry time counter value, internal use only
```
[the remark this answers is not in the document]
jenswi-linaro: Using milliseconds has the drawback that we may need to divide the time counter value with the frequency quite often. I suppose it would mean to implement `timer_now()` as
`return barrier_read_counter_timer() * 1000 / read_cntfrq()`.

Will `expiry_value` work better?

core/include/kernel/callout.h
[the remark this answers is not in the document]
etienne-lms: I'm fine it's considered in a later change.

etienne-lms:
shorter than 6 chars, you're quite challenging :)
`delayed_call`, `periodic_call`.
Well, I guess `callout` is ok since `struct callout` inline description comment well describes the scope and purpose of the feature.

jenswi-linaro:
Force push to fix the checkpatch issue.

jforissier (commented):
- For "core: arm64: add {read,write}_cntps_cval()":
- For "core: add callout service":
  s/will be converted to in/will be converted to ticks in/
  s/ If if/If/
  s/has be scheduled/has been scheduled/
  I am not sure about the name of `is_per_cpu`. If the underlying timer ensures that the interrupt is always received by the same core, then `is_per_cpu` doesn't matter, does it? (and it should preferably set to `false` to avoid some overhead). If the timer interrupt picks one (and only one) random core, then `is_per_cpu` needs to be `true`. This sounds backwards to me.
- For "core: define generic callout service initializer":
  s/shall provide and/shall provide an/ (in commit description)
- For "plat-vexpress: initialize callout service":
- For "plat-vexpress: qemu_armv8: define IT_SEC_PHY_TIMER":
- For "core: tests: add a notification test watchdog":
  s/uart/UART/ in the commit description
  The trace "D/TC:0   notif_send_async:93 0x0" every second doesn't seem to stop, that's a bit annoying.

core/arch/arm/kernel/sub.mk
```diff
@@ -2,6 +2,7 @@ srcs-y += rpc_io_i2c.c
 srcs-y += idle.c
 
 srcs-$(CFG_SECURE_TIME_SOURCE_CNTPCT) += tee_time_arm_cntpct.c
+srcs-$(CFG_ARM64_core) += generic_timer.c
```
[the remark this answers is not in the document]
jenswi-linaro: With minor tweaking, it should be usable in AArch32 also. But I can rename it until that is done if you like.

jenswi-linaro:
Force push to apply tags and fix up "plat-vexpress: initialize callout service"

jenswi-linaro:
> I am not sure about the name of `is_per_cpu`. If the underlying timer ensures that the interrupt is always received by the same core, then `is_per_cpu` doesn't matter, does it? (and it should preferably set to `false` to avoid some overhead). If the timer interrupt picks one (and only one) random core, then `is_per_cpu` needs to be `true`. This sounds backwards to me.

`is_per_cpu` is a description of the implementation. Timers can be global or per CPU. If it's per CPU it will only be manipulated by the current CPU. That means that if a new time is set on a different CPU we'll have an updated timer running on another CPU. Instead of updating that timer, we're just ignoring the interrupt from it.

jenswi-linaro:
Added "core: notif_send_async(): remove debug print" to remove the annoying debug print.

jforissier:
> > I am not sure about the name of `is_per_cpu`. If the underlying timer ensures that the interrupt is always received by the same core, then `is_per_cpu` doesn't matter, does it? (and it should preferably set to `false` to avoid some overhead). If the timer interrupt picks one (and only one) random core, then `is_per_cpu` needs to be `true`. This sounds backwards to me.
> 
> `is_per_cpu` is a description of the implementation. Timers can be global or per CPU. If it's per CPU it will only be manipulated by the current CPU. That means that if a new time is set on a different CPU we'll have an updated timer running on another CPU. Instead of updating that timer, we're just ignoring the interrupt from it.

I fail to imagine the use case. Can you add a test where `is_per_cpu` needs to be `true`? Otherwise I'd rather introduce it later when there is a real use case.

core/tests/notif_test_wd.c
```diff
@@ -0,0 +1,126 @@
...
+	if (wd->pending)
+		wd->timeout_count++;
+	wd->call_count++;
+	if (wd->call_count < 10 || !(wd->call_count % 60) || wd->pending)
+		DMSG("WD call_count %u, timeout_count %u",
+		     wd->call_count, wd->timeout_count);
+	wd->pending = true;
+	notif_send_async(NOTIF_VALUE_DO_BOTTOM_HALF);
+
+	return true;
```
jforissier, line 44: Perhaps the test should stop re-scheduling itself at some point, when one can see that everything works as expected?

jenswi-linaro:
> I fail to imagine the use case. Can you add a test where `is_per_cpu` needs to be `true`? Otherwise I'd rather introduce it later when there is a real use case.

I'm sorry, I forgot to update "core: arm64: implement timer_init_callout_service()" when I added this `is_per_cpu` during the review. Perhaps it's more clear now.

core/tests/notif_test_wd.c
jenswi-linaro: It's interesting to see that xtest can be run without any timeouts. I have coming patches where this works with FF-A and virtualization where two guests run this in parallel. I assume a debug print once a minute (per guest) isn't too much logging after the initial 10.

jforissier:
- For "core: tests: add a notification test watchdog":
- For "core: notif_send_async(): remove debug print":

jenswi-linaro:
Squashed and rebased on master.
These two commits have not been fully reviewed yet:
- core: add callout service
- core: arm64: implement timer_init_callout_service()

core/include/kernel/callout.h
```diff
@@ -0,0 +1,127 @@
...
+
+/*
+ * struct callout_timer_desc - callout timer descriptor
+ * @disable_timeout:	disables the timer from triggering an interrupt
+ * @set_next_timeout:	sets the next timeout and enables the timer
+ * @ms_to_ticks:	converts milliseconds to ticks, the counter value
+ *			unit
+ * @get_now:		get the current counter value
+ * @is_per_cpu:		flag to indicate if this timer is per cpu (true) or
+ *			global (false).
```
jforissier, line 96: I wouldn't mind more details here. How about: "When true, callout_add() saves the current core number. When an interrupt is triggered, the callout service will check the core number again and only invoke the callback if the core is the same."?

jforissier (commented):
For "core: add callout service": one comment below. In any case:

core/arch/arm/plat-vexpress/main.c
```diff
@@ -256,3 +257,14 @@ int psci_cpu_on(uint32_t core_id, uint32_t entry, uint32_t context_id)
 	return PSCI_RET_SUCCESS;
 }
 #endif /*PLATFORM_FLAVOR_qemu_virt*/
+
+#if defined(IT_SEC_PHY_TIMER) && !defined(CFG_CORE_SEL2_SPMC)
+static TEE_Result init_callout_service(void)
+{
+	timer_init_callout_service(interrupt_get_main_chip(), IT_SEC_PHY_TIMER);
```
jforissier, line 264: Still wondering about `is_per_cpu` here :wink: what happens when `IT_SEC_PHY_TIMER` is triggered? Is it delivered to only one core (and in this case, always the same or not?), to all cores, to a subset of all cores...?

jforissier (commented):
- For "core: arm64: implement timer_init_callout_service()":

core/arch/arm/plat-vexpress/main.c
jenswi-linaro: There is one timer per CPU, and each of them only triggers a PPI on that CPU. The PPI is triggered only on the CPU where the generic timer expires.

jenswi-linaro:
I've added more description to `struct callout_timer_desc`, I hope this helps.

jforissier:
> I've added more description to `struct callout_timer_desc`, I hope this helps.

Fine with me. Perhaps with `@` in front of `set_next_timeout()` since it is a member of the structure.

jenswi-linaro:
Squashed and tags applied. Should be ready to merge now, unless @etienne-lms or someone else has something to add.

core/tests/notif_test_wd.c
```diff
@@ -0,0 +1,126 @@
...
+struct periodic_data {
+	unsigned int count;
+	struct callout callout;
+};
+
+static bool periodic_callback(struct callout *co)
+{
+	struct periodic_data *d = container_of(co, struct periodic_data,
+					       callout);
+	TEE_Time t = { };
```
etienne-lms, line 97: `__maybe_unused`

core/tests/notif_test_wd.c
```diff
@@ -0,0 +1,126 @@
...
+	return &default_wd_data;
+}
+
+static bool test_wd_callback(struct callout *co)
+{
+	struct wd_data *wd = container_of(co, struct wd_data, callout);
+
+	if (wd->pending)
+		wd->timeout_count++;
+	wd->call_count++;
```
etienne-lms, line 37: Here we expect `wd->timeout_count == wd->call_count`, right?
Maybe an test PTA could report to a new xtest test ID whether this is true or not, so that regression test can check the notif and callout frameworks behave as expected.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with few minor comments to address.

core/tests/notif_test_wd.c
jenswi-linaro: This is always used by `tee_time_get_sys_time()`, even if we may throw away the result if debug prints are disabled. So I'm not sure that `__maybe_unused` is accurate.

core/tests/notif_test_wd.c
jenswi-linaro: No, we expect `wd->timeout_count` to be `0`, but the PTA thing sounds like a nice extension.

jenswi-linaro:
Addressed comments, squashed in the small fixes, and applied the tag.

jforissier:
This PR breaks `make check CFG_WITH_PAGER=y` on QEMUv8. See https://github.com/OP-TEE/optee_os/pull/6747.

PR #6725, opened 2024-03-02
core: kernel: add timeout_elapsed_us()

etienne-lms:
This P-R adds API function `timeout_elapsed_us()` to measure the time since or until an initialized timeout reference is elapsing. This function relies on other `timeout_*()` API functions and therefore depends on `CFG_CORE_HAS_GENERIC_TIMER` being enabled.

While at it, I factorize the `timeout_*()` function between `arm` and `riscv` architectures and slightly touched `mdelay()`.
I also changed _arch/risc/risc.mk_ to force `CFG_CORE_HAS_GENERIC_TIMER` instead of relying on _mk/config.mk_ default enabling that switch.

### DIFF
--- a/core/arch/arm/include/kernel/delay_arch.h
+++ b/core/arch/arm/include/kernel/delay_arch.h
@@ -30,23 +30,18 @@
 #ifndef __KERNEL_DELAY_ARCH_H
 #define __KERNEL_DELAY_ARCH_H
 
+#ifdef CFG_CORE_HAS_GENERIC_TIMER
 #include <arm.h>
-#include <stdbool.h>
 #include <stdint.h>
 
-static inline uint64_t arm_cnt_us2cnt(uint32_t us)
+static inline unsigned int delay_cnt_freq(void)
 {
-	return ((uint64_t)us * (uint64_t)read_cntfrq()) / 1000000ULL;
+	return read_cntfrq();
 }
 
-static inline uint64_t timeout_init_us(uint32_t us)
+static inline uint64_t delay_cnt_read(void)
 {
-	return barrier_read_counter_timer() + arm_cnt_us2cnt(us);
+	return barrier_read_counter_timer();
 }
-
-static inline bool timeout_elapsed(uint64_t expire)
-{
-	return barrier_read_counter_timer() > expire;
-}
-
+#endif /* CFG_CORE_HAS_GENERIC_TIMER */
 #endif
--- a/core/include/kernel/delay.h
+++ b/core/include/kernel/delay.h
@@ -54,6 +54,13 @@ static inline bool timeout_elapsed(uint64_t expire)
 {
 	return delay_cnt_read() > expire;
 }
+
+/*
+ * Return the time in microseconds since/until timeout tick counter @expired,
+ * that was initialized with timeout_init_us() or like, has/will expire.
+ * A positive value means the timeout has expired and a negative one it has not.
+ */
+int timeout_elapsed_us(uint64_t expire);
 #endif /*CFG_CORE_HAS_GENERIC_TIMER*/
 
 /* Wait @us microseconds actively polling on architecture timer */
--- a/core/include/kernel/delay.h
+++ b/core/include/kernel/delay.h
@@ -34,7 +34,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+/* Wait @us microseconds actively polling on architecture timer */
 void udelay(uint32_t us);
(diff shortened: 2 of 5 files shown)


core/arch/arm/include/kernel/delay_arch.h
```diff
@@ -30,23 +30,18 @@
 #ifndef __KERNEL_DELAY_ARCH_H
 #define __KERNEL_DELAY_ARCH_H
 
+#ifdef CFG_CORE_HAS_GENERIC_TIMER
 #include <arm.h>
-#include <stdbool.h>
 #include <stdint.h>
 
-static inline uint64_t arm_cnt_us2cnt(uint32_t us)
+static unsigned int delay_cnt_freq(void)
```
jenswi-linaro, line 37: Why not "static inline"?
Same question below.

core/include/kernel/delay.h
```diff
@@ -33,6 +33,28 @@
...
+/* Convert microsecond base delay @us into architecture time tick counts */
+static inline uint64_t delay_us2cnt(uint32_t us)
+{
+	return ((uint64_t)us * (uint64_t)delay_cnt_freq()) / ULL(1000000);
+}
+
+/* Return delay tick counter for a timeout expiration in @us microseconds */
+static inline uint64_t timeout_init_us(uint32_t us)
+{
+	return barrier_read_counter_timer() + delay_us2cnt(us);
```
jenswi-linaro, line 48: `delay_cnt_read()` instead of `barrier_read_counter_timer()`?

core/kernel/delay.c
```diff
@@ -32,6 +32,20 @@
 #include <kernel/misc.h>
 
 #ifdef CFG_CORE_HAS_GENERIC_TIMER
+int timeout_elapsed_us(uint64_t expire)
+{
+	int64_t diff = delay_cnt_read() - expire;
+
+	if (MUL_OVERFLOW(diff, ULL(1000000), &diff)) {
```
jenswi-linaro, line 39: Is `ULL()` useful here?

core/kernel/delay.c
```diff
@@ -32,6 +32,20 @@
...
+		if (timeout_elapsed(expire))
+			return INT_MAX;
+		else
+			return INT_MIN;
+	}
+
+	return (int)(diff / (uint64_t)delay_cnt_freq());
```
jenswi-linaro, line 46: Why is any casting needed? Casting the result from `delay_cnt_freq()` to the same type as `diff` might make sense, but I'm not sure that there's a point in even that.

core/arch/arm/include/kernel/delay_arch.h
etienne-lms: my mistake. thanks to point that out.

core/include/kernel/delay.h
etienne-lms: Rigth, my mistake. I'll update.

core/kernel/delay.c
etienne-lms: I don't think it is. I used it to mimic this old implementation.
I'll drop it :).

core/kernel/delay.c
etienne-lms: I guess casting `delay_cnt_freq()` is not needed. As for casting the result to an `int`, maybe I should trap them in the above test.

jforissier (commented):
- "core: kernel: describe udelay()/mdelay()":
- "core: riscv: force enable of CFG_CORE_HAS_GENERIC_TIMER":
  In the commit description: s/riscv/RISC-V/ and s/has current the implementation/since the current implementation/?
- "core: kernel: factorize delay and timeout implementation":
  s/riscv/RISC-V/, s/arm/Arm/

core/kernel/delay.c
```diff
@@ -32,6 +32,21 @@
 #include <kernel/misc.h>
 
 #ifdef CFG_CORE_HAS_GENERIC_TIMER
+int timeout_elapsed_us(uint64_t expire)
+{
+	int64_t diff = delay_cnt_read() - expire;
+
+	if (MUL_OVERFLOW(diff, ULL(1000000), &diff) ||
```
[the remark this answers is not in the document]
etienne-lms: sorry, i'm not very focus these days.

etienne-lms:
I've squashed the fixup commits, addressed @jenswi-linaro's above comment and applied the review tags.
Thanks for the help.

PR #6726, opened 2024-03-02
lib: bget_malloc: wipe content outside critical section

etienne-lms:
Changes **bget** integration in OP-TEE to wipe data before core heap spinlock is locked for safe heap management. This change prevents interrupts from being masked for a long time when the wiped buffer is somewhat large (few kilobytes).

`brel()` from _bget.c_ gets an extra argument boolean `wipe_only` argument, set to true together with argument `wipe` when we only request to wipe data content.

### DIFF
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -457,7 +457,17 @@ void raw_free(void *ptr, struct malloc_ctx *ctx, bool wipe)
 	raw_malloc_validate_pools(ctx);
 
 	if (ptr)
-		brel(maybe_untag_buf(ptr), &ctx->poolset, wipe);
+		brel(maybe_untag_buf(ptr), &ctx->poolset, wipe,
+		     false /*!wipe-only*/);
+}
+
+void raw_wipe(void *ptr, struct malloc_ctx *ctx)
+{
+	raw_malloc_validate_pools(ctx);
+
+	if (ptr)
+		brel(maybe_untag_buf(ptr), &ctx->poolset, true /*wipe*/,
+		     true /*wipe-only*/);
 }
 
 void *raw_calloc(size_t hdr_size, size_t ftr_size, size_t pl_nmemb,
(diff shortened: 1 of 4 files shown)


lib/libutils/isoc/bget_malloc.c
```diff
@@ -795,9 +821,13 @@ void *malloc(size_t size)
 
 static void free_helper(void *ptr, bool wipe)
 {
-	uint32_t exceptions = malloc_lock(&malloc_ctx);
+	uint32_t exceptions = 0;
+
+	if (wipe)
+		raw_wipe(ptr, &malloc_ctx);
```
jenswi-linaro, line 827: You're accessing the internal data structure of the heap without any lock, this is bound to cause problems.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -457,7 +457,17 @@ void raw_free(void *ptr, struct malloc_ctx *ctx, bool wipe)
...
+		brel(maybe_untag_buf(ptr), &ctx->poolset, wipe,
+		     false /*!wipe-only*/);
+}
+
+void raw_wipe(void *ptr, struct malloc_ctx *ctx)
+{
+	raw_malloc_validate_pools(ctx);
+
+	if (ptr)
+		brel(maybe_untag_buf(ptr), &ctx->poolset, true /*wipe*/,
```
jenswi-linaro, line 469: Since the wiping always is going to be done in a separate step there's no need to let the `brel()` function do the wiping.
We do something close to wiping in `maybe_untag_buf()` already. How about extending that function to wipe the buffer too?

etienne-lms: We don't knon the size of the space, this is where brel gives a size big hint. We wipe it, indeed using internal states. Needs another though. We must find back from bet the initially required size we can safely wipe without writing to brel internal states.

jenswi-linaro: What's wrong with `bget_buf_size()`?

etienne-lms: it was hidden behind some tree. Here i can see it now :)

PR #6727, opened 2024-03-02
core: kernel: allow not locking output console

etienne-lms:
Add configuration switch `CFG_TEE_CONSOLE_UNLOCKED` to evacuate trace messages without a=hold the contention lock.

This configuration can be handy to still benefit from OP-TEE output console support while to affecting native and foreign interrupts handling latency due to OP-TEE output console traces for example when playing with a real time sensitive system. Enabling the configuration permits collision of trace messages but is worth it instead of fully disabling OP-TEE console for example with `CFG_TEE_CORE_LOG_LEVEL=0`.

Enabling the switch does not guarantee trace messages are printed in a thread context with interrupts not being masked. At least OP-TEE generic trace support will not add constraints on the interrupts and scheduling latency.

By the way, add an initial value to local variable in `trace_ext_puts()`.

### DIFF
--- a/core/kernel/trace_ext.c
+++ b/core/kernel/trace_ext.c
@@ -5,6 +5,7 @@
 #include <stdbool.h>
 #include <trace.h>
 #include <console.h>
+#include <config.h>
 #include <kernel/misc.h>
 #include <kernel/spinlock.h>
 #include <kernel/thread.h>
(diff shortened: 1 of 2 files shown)


core/kernel/trace_ext.c
```diff
@@ -21,14 +22,18 @@ void __weak plat_trace_ext_puts(const char *str __unused)
...
-	const char *p;
+	const char *p = NULL;
 
-	if (mmu_enabled && !cpu_spin_trylock(&puts_lock)) {
-		was_contended = true;
-		cpu_spin_lock_no_dldetect(&puts_lock);
+	if (!IS_ENABLED(CFG_TEE_CONSOLE_UNLOCKED)) {
+		itr_status = thread_mask_exceptions(THREAD_EXCP_ALL);
+
+		if (mmu_enabled && !cpu_spin_trylock(&puts_lock)) {
```
jenswi-linaro, line 33: You can still loop around `cpu_spin_trylock()` and unmask and mask exceptions as needed to take the spinlock, this should avoid mixed logs while still not keeping interrupts masked for a longer time.

PR #6728, opened 2024-03-02
[RFC] core: arm: mm: relax interrupt on dynamic shm mapping

etienne-lms:
Add `CFG_DYN_SHM_INTERRUPTIBLE_LOCK` configuration switch to lower the time during which interrupts are masked for dynamic shared memory registration and mapping operations.

The configuration is intended to be used when OP-TEE OS is executing with as non-secure OS that is sensitive to interrupt and scheduling latency. Indeed playing with a Cortex-A7 clocked at 1GHz, mapping and especially unmapping a shared memory buffer of several MBytes makes OP-TEE to mask interrupts during several hundreds of milliseconds.

When CFG_DYN_SHM_INTERRUPTIBLE_LOCK is enabled, the lock used to protect dynamic shared memory reference is still used but without masking interrupts hence at a cost of potentially high CPU load picks. The memory is mapped page per page because to ensure `core_mmu_map_pages()` and `core_mmu_unmap_pages()` mask interrupts only for the duration of processing a single page.

### DIFF
--- a/core/arch/arm/mm/mobj_dyn_shm.c
+++ b/core/arch/arm/mm/mobj_dyn_shm.c
@@ -69,6 +69,46 @@ static unsigned int reg_shm_map_lock = SPINLOCK_UNLOCK;
 
 static struct mobj_reg_shm *to_mobj_reg_shm(struct mobj *mobj);
 
+static uint32_t lock_reg_shm_slist(void)
+{
+	uint32_t exceptions = 0;
+
+	if (IS_ENABLED(CFG_DYN_SHM_INTERRUPTIBLE_LOCK))
+		thread_spin_lock(&reg_shm_slist_lock);
+	else
+		exceptions = cpu_spin_lock_xsave(&reg_shm_slist_lock);
+
+	return exceptions;
+}
+
+static void unlock_reg_shm_slist(uint32_t exceptions)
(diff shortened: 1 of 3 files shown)


core/arch/arm/mm/mobj_dyn_shm.c
```diff
@@ -120,21 +160,38 @@ static void *mobj_reg_shm_get_va(struct mobj *mobj, size_t offst, size_t len)
...
+		 * Unmap pages one by one to prevent masking interrupts
+		 * over a potentially long period of time.
+		 */
+		burst_pages = 1;
+		burst_count = r->mm->size;
+	}
+
+	va = tee_mm_get_smem(r->mm);
+	for (i = 0; i < burst_count; i++)
+		core_mmu_unmap_pages(va + i * SMALL_PAGE_SIZE, burst_pages);
```
jenswi-linaro, line 182: How about releasing the lock in `core_mmu_unmap_pages()` just after the lookup with `find_map_by_va()` instead? That way we don't need to care about `burst_pages`.

jenswi-linaro (commented):
I'd rather try to reduce the size of the critical sections with interrupts masked. This workaround adds a bit of technical debt.

PR #6729, opened 2024-03-02
Upgrade old use of thread_set_exceptions()

etienne-lms:
Updates from historical source files that use `thread_set_exceptions()` where `thread_unmask_exceptions()` is more appropriate and even deserve to be replaced with `cpu_spin_lock_xsave()`/`cpu_spin_unlock_xrestore()` helper functions.

No functional changes.

PR #6730, opened 2024-03-02
plat-synquacer: add initialization value to local variables

etienne-lms:
Add missing initialization value to local variables in synquacer platform RNG driver to better comply with OP-TEE OS coding style and prevent developers from being confused when using this example as a example.

No functional change.

I propose this change because, when cleaning the patches related to https://github.com/OP-TEE/optee_os/pull/6729 I saw that driver infringes few OP-TEE OS coding style rules and deserve to be updates for the consistency of the source tree.

@b49020, please tell me if you're ok with such modifications.


b49020:

etienne-lms:
Review tag applied
(note: will conflict if https://github.com/OP-TEE/optee_os/pull/6729 is merged first. If so, i'll rebase)

PR #6732, opened 2024-03-04
driver: crypto: hisilicon: fix timeout error

yuzexiyzx:
In last commit, I changed the code from using cnt to using timeout to determine when to break the circulate.However, this do not take account of time cost in function(such as qm_mb_read()), only time cost in delay. This will cause timeout when running the code.So, I rolled back this change.

### DIFF
--- a/core/arch/arm/plat-sam/conf.mk
+++ b/core/arch/arm/plat-sam/conf.mk
@@ -4,6 +4,7 @@ flavor_dts_file-sama5d2xult = at91-sama5d2_xplained.dts
 flavor_dts_file-sama5d2_xplained = at91-sama5d2_xplained.dts
 flavor_dts_file-sama5d27_som1_ek = at91-sama5d27_som1_ek.dts
 flavor_dts_file-sama5d27_wlsom1_ek = at91-sama5d27_wlsom1_ek.dts
+flavor_dts_file-sama7g54_ek = at91-sama7g54_ek.dts
 
 ifeq ($(PLATFORM_FLAVOR),sama5d2xult)
 $(warning "sama5d2xult is deprecated, please use sama5d2_xplained")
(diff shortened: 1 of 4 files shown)


jenswi-linaro (commented):
I don't understand this. `qm_mb_read()` takes some time, but you don't want to account for that while determining if there is a timeout?

yuzexiyzx:
To use cnt, it will delay every time get in the circulate.So, the max time is  QM_MB_WAIT_READY_CNT * (QM_MB_WAIT_PERIOD + qm_mb_read()).But to timeout, we define timeout = timeout_init_us(QM_MB_WAIT_PERIOD * QM_MB_WAIT_READY_CNT);

yuzexiyzx:
No. I want to account for that while determining if there is a timeout.But with timeout = timeout_init_us(QM_MB_WAIT_PERIOD * QM_MB_WAIT_READY_CNT), we do not account for it.

jenswi-linaro:
I still don't get it. Currently, you will get a timeout after `QM_MB_WAIT_PERIOD * QM_MB_WAIT_READY_CNT` milliseconds. But that's not enough, you need to wait a bit longer. How much longer is that? In the patch, you seem to just want to ignore the time spent in `qm_mb_read()`. That's why I can't make sense of it.

yuzexiyzx:
Yes.QM_MB_WAIT_PERIOD * QM_MB_WAIT_READY_CNT is too short.I need to wait a bit longer, QM_MB_WAIT_READY_CNT * (QM_MB_WAIT_PERIOD + time cost in qm_mb_read()).I want to make timeout longer, but I don't know how much time cost in qm_mb_read().To use cnt, I don't care how much time cost in qm_mb_read(), only need to know how many times I get in the circulate.

jenswi-linaro:
That is a very odd timeout.  I don't see why it matters how many loops you do. It sounds like you're missing something essential or you're trying to work around a hardware limitation. Doing strange things without an explanation makes it hard to maintain the code.

Are you doing this as an emergency fix until you have a proper fix, or is this the best you can do for now? If it's the latter I would like to see comments in the code explaining why the delay must be implemented in this way.

Please add a fixes tag. Does
```
Fixes: 6374dbce66f4 ("driver: crypto: hisilicon: Add the mailbox operation lock")
```
make sense?

yuzexiyzx:
@jenswi-linaro I want to account for time cost in qm_mb_read(), not just delay period, but I don‘t know how much time cost in qm_mb_read().I can not find a good way to use timeout to implement this.

yuzexiyzx:
@jenswi-linaro @etienne-lms Do you have any good way to know the time cost in the whole circulate not just the delay period?I don't know what value I should init timeout.
image

yuzexiyzx:
@jenswi-linaro If time cost in qm_mb_read() is 50us and delay is 100us,  I want to init timeous as 150 * QM_MB_WAIT_MAX_CNT, not 100 * QM_MB_WAIT_MAX_CNT.But I donot know don‘t know how much time cost in qm_mb_read().So, I choose to use cnt instead of time.If there is a good way to know that, I agree that using timeout would be better.

yuzexiyzx:
@jenswi-linaro This may be the best I can do for now as above, but this is not related to hardware limitation.I will add a fixes tag.

jenswi-linaro:
I assume the purpose of this timeout is to avoid being stuck here forever and that we'll normally get a result fairly quickly. I'd recommend picking a large timeout, perhaps a second. You can fine-tune this later on if needed. Reverting to having a delay in the poll loop is not the best option since you will not wait for the hardware to be ready instead you will be waiting for time to pass.

yuzexiyzx:
Ok.I will try to do this in that way.Thanks!

yuzexiyzx:
@jenswi-linaro Another question. May I change the type of return value from enum hisi_drv_status to int? I can still use these values,just change the type.

jenswi-linaro:
It depends on the situation. Propose the change and I'll review it.

yuzexiyzx:
@etienne-lms Thanks!

PR #6740, opened 2024-03-06
core: add __must_check attribute to cpu_spin_lock_xsave()

jenswi-linaro:
cpu_spin_lock_xsave() masks exceptions, takes the spinlock and returns previous exception state to be restored by cpu_spin_unlock_xrestore(). The previously returned exception state must always be supplied so add the __must_check attribute to cpu_spin_lock_xsave() and its debug variants.


jforissier (commented):

PR #6743, opened 2024-03-08
ta: remoteproc: allow remoteproc_load_fw re-entrance

arnopo:
The way that the Linux kernel remoteproc framework loads and starts the firmware does not fully match with the OP-TEE remoteproc.
The Linux remoteproc can request the resource table before loading the image, while the OP-TEE remoteproc TA can provide the resource table only after authenticating (and potentially decrypting) the firmware images. the authentication is done during the load step.

This PR is the result of a discussion  with @mathieupoirier on  Linux remoteproc mailing list [1][2]. The aim is to simplify the Linux implementation allowing to call the TA_RPROC_CMD_LOAD_FW twice times.
As we just ignore the second request I don't identify issue of returning TEE_SUCCESS instead of an error.

[1] Related discussion: https://www.spinics.net/lists/kernel/msg5124824.html
[2] Linux kernel series: https://www.spinics.net/lists/kernel/msg5103449.html


jforissier (commented):
That looks reasonable indeed.

arnopo:
> That looks reasonable indeed.
> 

Added in commit message.
thanks!

etienne-lms:
Would you mind adding a reference to Mathieu's request in the commit message?
```
Reported-by: Mathieu Poirier <...>
Link: https://www.spinics.net/lists/kernel/msg5124824.html
```
Since this fixes an issue experienced witht eh Linux kernel, I think it also deserves a `Fixes` tags:
```
Fixes: fcf382e2440c ("ta: remoteproc: add remote processor Trusted Application")
```

arnopo:
> `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`

Added 

> Would you mind adding a reference to Mathieu's request in the commit message?
> 
> ```
> Reported-by: Mathieu Poirier <...>
That a good idea! I put suggested-by instead as it is not a bug but a behavior

> Link: https://www.spinics.net/lists/kernel/msg5124824.html

Done using kernel.org link instead.

> ```
> 
> Since this fixes an issue experienced witht eh Linux kernel, I think it also deserves a `Fixes` tags:
> 
> ```
> Fixes: fcf382e2440c ("ta: remoteproc: add remote processor Trusted Application")
> ```

Added
thanks!

PR #6748, opened 2024-03-12
drivers: mutex for clock and regulator valid during PM suspend/resume sequences

etienne-lms:
New proposal for the change posted in https://github.com/OP-TEE/optee_os/pull/6724.

This P-R propose a (somewhat) generic implementation for the "PM aware" mutex and apply it the the clock framework (to replace current spinning lock) and in the regulator framework (which already implements the sequence that was seen to weak during https://github.com/OP-TEE/optee_os/pull/6724 review.

### DIFF
--- a/core/drivers/clk/clk.c
+++ b/core/drivers/clk/clk.c
@@ -7,18 +7,29 @@
 #include <config.h>
 #include <drivers/clk.h>
 #include <kernel/boot.h>
+#include <kernel/mutex_pm_aware.h>
 #include <kernel/panic.h>
-#include <kernel/spinlock.h>
+#include <kernel/thread.h>
 #include <malloc.h>
 #include <stddef.h>
 #include <stdio.h>
 
-/* Global clock tree lock */
-static unsigned int clk_lock = SPINLOCK_UNLOCK;
+/* Global clock tree access protection complying the power state transitions */
+static struct mutex_pm_aware mu = MUTEX_PM_AWARE_INITIALIZER;
 
 #ifdef CFG_DRIVERS_CLK_PRINT_TREE
 static SLIST_HEAD(, clk) clock_list = SLIST_HEAD_INITIALIZER(clock_list);
 #endif
+
+static void lock_clk(void)
+{
+	mutex_pm_aware_lock(&mu);
+}
+
+static void unlock_clk(void)
+{
+	mutex_pm_aware_unlock(&mu);
+}
 
 struct clk *clk_alloc(const char *name, const struct clk_ops *ops,
 		      struct clk **parent_clks, size_t parent_count)
@@ -173,23 +184,20 @@ static TEE_Result clk_enable_no_lock(struct clk *clk)
 
 TEE_Result clk_enable(struct clk *clk)
 {
-	uint32_t exceptions = 0;
 	TEE_Result res = TEE_ERROR_GENERIC;
 
-	exceptions = cpu_spin_lock_xsave(&clk_lock);
+	lock_clk();
 	res = clk_enable_no_lock(clk);
-	cpu_spin_unlock_xrestore(&clk_lock, exceptions);
+	unlock_clk();
 
 	return res;
 }
 
 void clk_disable(struct clk *clk)
 {
-	uint32_t exceptions = 0;
-
-	exceptions = cpu_spin_lock_xsave(&clk_lock);
+	lock_clk();
 	clk_disable_no_lock(clk);
-	cpu_spin_unlock_xrestore(&clk_lock, exceptions);
+	unlock_clk();
 }
 
 unsigned long clk_get_rate(struct clk *clk)
@@ -236,17 +244,16 @@ static TEE_Result clk_set_rate_no_lock(struct clk *clk, unsigned long rate)
 
 TEE_Result clk_set_rate(struct clk *clk, unsigned long rate)
 {
-	uint32_t exceptions = 0;
 	TEE_Result res = TEE_ERROR_GENERIC;
 
-	exceptions =  cpu_spin_lock_xsave(&clk_lock);
+	lock_clk();
 
 	if (clk->flags & CLK_SET_RATE_GATE && clk_is_enabled_no_lock(clk))
 		res = TEE_ERROR_BAD_STATE;
 	else
 		res = clk_set_rate_no_lock(clk, rate);
 
-	cpu_spin_unlock_xrestore(&clk_lock, exceptions);
+	unlock_clk();
 
 	return res;
 }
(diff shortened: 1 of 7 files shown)


core/include/kernel/mutex.h
```diff
@@ -76,6 +76,27 @@ void mutex_unlock_recursive(struct recursive_mutex *m);
...
+ * non-secure world is not operational, for example in fastcall SMC entries
+ * of the PSCI services. In such case we cannot take a mutex and we expect
+ * the mutex is unlocked. Additionally a spinning lock is attempted to be
+ * locked to check the resource access consistency.
+ */
+struct mutex_pm_aware {
+	struct mutex mutex;	/* access protection in thread context */
+	unsigned int lock;	/* access consistency in PM context */
+};
+
```
jenswi-linaro, line 94: Wouldn't a `MUTEX_PM_AWARE_INITIALIZER` make sense too?

core/include/kernel/mutex.h
```diff
@@ -76,6 +76,27 @@ void mutex_unlock_recursive(struct recursive_mutex *m);
...
+/*
+ * struct mutex_pm_aware - Mutex usable in PM atomic sequence
+ *
+ * Some resources need a mutex protection for runtime operations but are
+ * also accessed during specific system power transition (PM power off,
+ * suspend and resume) that operate in atomic execution environment where
+ * non-secure world is not operational, for example in fastcall SMC entries
+ * of the PSCI services. In such case we cannot take a mutex and we expect
+ * the mutex is unlocked. Additionally a spinning lock is attempted to be
+ * locked to check the resource access consistency.
```
jenswi-linaro, line 88: Please mention that `mutex_pm_aware_lock()` is expected to panic if it detects contention with an atomic context, regardless if called from a threaded or atomic context, but that contention between two threaded contexts is OK.

core/include/drivers/regulator.h
```diff
@@ -87,7 +87,7 @@ struct regulator_voltages_desc {
  * @max_uv: Max possible voltage level in microvolt (uV)
  * @flags: REGULATOR_* property flags
  * @refcount: Regulator enable request reference counter
- * @lock: Mutex for concurrent access protection
+ * @mu: Mutex for concurrent access protection with PM context support
```
jenswi-linaro, line 90: s/mu/mutex/

core/include/kernel/mutex.h
etienne-lms: Sorry, it's not for that reason. But the conflict is there. I'll try to find a way to define it and will tell you.

etienne-lms:
Comments addressed.

core/include/kernel/mutex_pm_aware.h
```diff
@@ -0,0 +1,40 @@
...
+ * struct mutex_pm_aware - Mutex usable in PM atomic sequence
+ *
+ * Some resources need a mutex protection for runtime operations but are
+ * also accessed during specific system power transition (PM power off,
+ * suspend and resume) that operate in atomic execution environment where
+ * non-secure world is not operational, for example in fastcall SMC entries
+ * of the PSCI services. In such case we cannot take a mutex and we expect
+ * the mutex is unlocked. Additionally a spinning lock is attempted to be
+ * locked to check the resource access consistency. In case of contention
+ * between thread and non-thread access to the resource, core will
```
jenswi-linaro, line 21: Contention of non-threaded access always results in a panic. Only contention between threaded accesses will be handled gracefully. This description needs to be rephrased to make that clear.

Please split the comment into two paragraphs, one with the background and one short that describes the intended behavior.

core/kernel/mutex.c
```diff
@@ -342,6 +343,40 @@ unsigned int mutex_get_recursive_lock_depth(struct recursive_mutex *m)
 	return refcount_val(&m->lock_depth);
 }
 
+void mutex_pm_aware_init(struct mutex_pm_aware *m)
+{
+	mutex_init(&m->mutex);
```
jenswi-linaro, line 348: Please use `MUTEX_PM_AWARE_INITIALIZER`.

core/include/kernel/mutex_pm_aware.h
etienne-lms: Ok, I understand my phrasing is confusing. I'll do.

core/kernel/mutex.c
etienne-lms: Right.

etienne-lms:
comment addressed

core/include/kernel/mutex_pm_aware.h
```diff
@@ -17,9 +17,13 @@
...
- * locked to check the resource access consistency. In case of contention
- * between thread and non-thread access to the resource, core will
- * intentionally panic.
+ * locked to check the resource access consistency.
+ *
+ * In case a thread request the mutex while a non-thread access still holds
+ * it, core will intentionally panic since it is not expected.
+ *
+ * In case a non-thread access request the mutex while it is held by a thread,
+ * core will also panic since it is not an expected situation.
```
jenswi-linaro, line 26: Please add the third case with non-threaded access while another core holds the spinlock.

etienne-lms: Ok. Comments addressed.

jenswi-linaro (commented):

core/drivers/clk/clk.c
```diff
@@ -7,19 +7,38 @@
...
 static SLIST_HEAD(, clk) clock_list = SLIST_HEAD_INITIALIZER(clock_list);
 #endif
 
+/*
+ * Clock operation may occur at runtime and during specific
+ * system power transition (PM power off, suspend and resume)
+ * that operate for example in fastcall SMC entries of the PSCI
+ * services where non-secure world is not operational. In these
+ * cases we cannot take a mutex and will expect the mutex is
+ * unlocked.
```
etienne-lms, line 30: Maybe should I simplify this description? (and ditto for the equiv inline comment in _regulator.c_)?
```
/*
 * Clock operation may occur at runtime and during specific
 * system power transition (PM power off, suspend and resume)
 * hence use a PM aware mutex.
 */

etienne-lms (commented):

etienne-lms:
I've removed the [RFC] tag from the P-R header line.

core/drivers/clk/clk.c
jenswi-linaro: Sure, or even better refer to the description of the `struct mutex_pm_aware`.

etienne-lms: I've simplified the description.
clk.c: `/* Global clock tree access protection complying the power state transitions */`
regualtor.c: `/* Access protection mutex complying the power state transitions context */`

etienne-lms:
Comments addressed.

etienne-lms:
Thanks

PR #6757, opened 2024-03-18
drivers: atmel_wdt: upgrade to support sama7g5 watchdog

TonyHan11:
In sama7g5 there's a DWDT (Dual Watchdog Timer) and the registers are not the same as the wdt for sama5d2. Here the DWD is handled as 2 watchdogs.

### DIFF
--- a/core/drivers/atmel_wdt.c
+++ b/core/drivers/atmel_wdt.c
@@ -27,7 +27,8 @@
 #define WDT_MR_WDV_SET(val)	((val) & WDT_MR_WDV)
 #define WDT_MR_WDFIEN		BIT(12)
 #define WDT_MR_WDRSTEN		BIT(13)
-#define WDT_MR_WDDIS		BIT(15)
+#define WDT_MR_WDDIS		BIT(15) /* Watchdog Disable of WDT on bit 15 */
+#define WDT_MR_WDDIS_DWDT	BIT(12) /* Watchdog Disable of DWDT on bit 12 */
 #define WDT_MR_WDD_SHIFT	16
 #define WDT_MR_WDD_MASK		GENMASK_32(11, 0)
 #define WDT_MR_WDD		SHIFT_U32(WDT_MR_WDD_MASK, WDT_MR_WDD_SHIFT)
@@ -39,6 +40,38 @@
 #define WDT_SR			0x8
 #define WDT_SR_DUNF		BIT(0)
 #define WDT_SR_DERR		BIT(1)
+
+/* DWDT: Watchdog Timer Mode Register */
+#define WDT_MR_PERIODRST	BIT(4)
+#define WDT_MR_RPTHRST		BIT(5)
+
+/* DWDT: Watchdog Timer Value Register (Read-only) */
+#define WDT_VR			0x8
+#define WDT_VR_COUNTER_SHIFT	0
+#define WDT_VR_COUNTER_MASK	GENMASK_32(11, 0)
+
+/* DWDT: Watchdog Timer Window Level Register */
+#define WDT_WL			0xc
+#define WDT_WL_RPTH_SHIFT	16
+#define WDT_WL_RPTH_MASK	GENMASK_32(27, 16)
+#define WDT_WL_PERIOD_SHIFT	0
+#define WDT_WL_PERIOD_MASK	GENMASK_32(11, 0)
+
+/* DWDT: Watchdog Timer Interrupt Level Register */
+#define WDT_IL			0x10
+#define WDT_IL_LVLTH_SHIFT	0
+#define WDT_IL_LVLTH_MASK	GENMASK_32(11, 0)
+
+/* DWDT: Watchdog Timer Interrupt Enable/Disable/Status/Mask Register */
+#define WDT_IER			0x14
+#define WDT_IDR			0x18
+#define WDT_ISR			0x1c
+#define WDT_IMR			0x20
+#define WDT_NSRPTHINT		BIT(4)
+#define WDT_NSPERINT		BIT(3)
+#define WDT_LVLINT		BIT(2)
+#define WDT_RPTHINT		BIT(1)
+#define WDT_PERINT		BIT(0)
 
 /*
  * The watchdog is clocked by a 32768Hz clock/128 and the counter is on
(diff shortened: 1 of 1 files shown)


core/drivers/atmel_wdt.c
```diff
@@ -283,8 +365,27 @@ static TEE_Result wdt_node_probe(const void *fdt, int node,
...
+static const struct wdt_compat sama5d2_compat = {
+	.wdt_ps = true,
+	.type = WDT_TYPE_WDT,
+	.dis_mask = WDT_MR_WDDIS,
+};
+
+static const struct wdt_compat sama7g5_compat = {
+	.wdt_ps = false,
+	.type = WDT_TYPE_DWDT,
+	.dis_mask = BIT(12),
```
jforissier, line 377: It would be nice to have a define for this bit

core/drivers/atmel_wdt.c
```diff
@@ -11,12 +11,10 @@
 #include <kernel/delay.h>
 #include <kernel/dt.h>
 #include <kernel/dt_driver.h>
-#include <kernel/interrupt.h>
 #include <kernel/pm.h>
 #include <matrix.h>
 #include <mm/core_mmu.h>
 #include <platform_config.h>
-#include <tee_api_types.h>
```
jforissier, line 19: Removing these #include should be a separate commit

core/drivers/atmel_wdt.c
```diff
@@ -59,12 +89,24 @@
...
+	enum wdt_type type; /* Type of Watchdog Timer */
+	uint32_t dis_mask; /* Mask of Watchdog Disable in Mode Register */
+};
 
 struct atmel_wdt {
 	struct wdt_chip chip;
+	enum wdt_type type;
+	uint32_t dis_mask;
 	vaddr_t base;
-	unsigned long rate;
```
jforissier, line 67: This too should be a separate commit

jforissier (commented):
Hello @TonyHan11,
Please see a couple of comments below. With that addressed:

TonyHan11:
Updated with adding a definition for the `disable_mask`, adding 2 separate commits for removing the unused header file and variable, review tags added. Thank you very much @jforissier

core/drivers/atmel_wdt.c
jforissier: `#include <kernel/interrupt.h>` should be removed in this commit also

TonyHan11: Do you mean the `kernel/interrupt.h` in [core/include/drivers/wdt.h](https://github.com/OP-TEE/optee_os/blob/63bfec5e2/core/include/drivers/wdt.h#L10)? As interrupt is used in [this file](https://github.com/OP-TEE/optee_os/blob/63bfec5e264b66c00bfc1b6ba807ec0f75495891/core/drivers/atmel_wdt.c#L252) the header can not be removed.
Updated with removing some other header files, thanks.

core/drivers/atmel_wdt.c
```diff
@@ -224,8 +295,9 @@ static void atmel_wdt_register_pm(struct atmel_wdt *wdt __unused)
 #endif
 
 static TEE_Result wdt_node_probe(const void *fdt, int node,
-				 const void *compat_data __unused)
+				 const void *compat_data)
 {
+	struct wdt_compat *compat = (struct wdt_compat *)compat_data;
```
etienne-lms, line 300: could use `const struct wdt_compat *compat` and remove the cast.

TonyHan11:
Removed the  `!` characters and update the code for `*compat`, thanks.

core/drivers/atmel_wdt.c
etienne-lms: `#include <tee_api_types.h>` is needed for `TEE_Result` definition.
`#include <kernel/delay.h>` is needed for `udelay()` declaration.
`#include <kernel/dt_driver.h>` is needed for `DEFINE_DT_DRIVER()` definition.

etienne-lms (commented):
Comments for commit "drivers: atmel_wdt: remove the header files not used from "#include"".
By the way, I don't think you need to list the removed header files in the commit message, but that's not a strong opinion.

core/drivers/atmel_wdt.c
etienne-lms: I forget to save many other comments when posting my review outcome. Here they are.

Many of these header files are needed IMO:
`#include <kernel/delay.h>` is needed for `udelay()` declaration.
`#include <kernel/dt.h>` is needed for `dt_map_dev()` and `fdt_get_status()` declaration and some other macros.
`#include <kernel/dt_driver.h>` is needed for `DEFINE_DT_DRIVER()` and few other definitions/declarations.
`#include <mm/core_mmu.h>` is needed for `core_mmu_remove_mapping()` declaration.
`#include <tee_api_types.h>` could be omitted but _tee_api_defines.h_ is needed for definition of `TEE_SUCCESS` and `TEE_ERROR_*`.

`#include <kernel/interrupt.h>` is needed for many function declarations and types definitions.

There are some header files that are missing: _util.h_, _stdbool.h_, _stdint.h_, _kernel/panic.h_, _trace.h_, _malloc.h_.

etienne-lms (commented):
Comments for commit "drivers: atmel_wdt: remove the header files not used from "#include"".
By the way, I don't think you need to list the removed header files in the commit message, but that's not a strong opinion.

etienne-lms (commented):
"drivers: atmel_wdt: upgrade to support sama7g5 watchdog" and
"drivers: atmel_wdt: remove the unused variable from "struct atmel_wdt"".

core/drivers/atmel_wdt.c
TonyHan11: Update `#include list of the header files by:
 - Remove the unused header files from "#include".
 - `#include` the header files needed explicitly even if they are included indirectly.

TonyHan11:
Not remove `#include <kernel/interrupt.h>` from commit `drivers: atmel_wdt: upgrade to support sama7g5 watchdog`.

TonyHan11:
Reabse the commits onto the latest 'master' branch.

etienne-lms (commented):

TonyHan11:
`Acked-by` added.

PR #6759, opened 2024-03-19
core: arm64: increase STACK_ABT_SIZE from 1024 to 3072 when log level is 0

jforissier:
When adding "make check CFG_WITH_PAGER=y CFG_TEE_CORE_LOG_LEVEL=0" to the QEMUv8 CI job, I noticed that OP-TEE fails to boot and hangs with no message printed on the console. The root cause is memory corruption of the translation tables triggered by a stack overflow. Fix this by removing the particular case for log level 0.

Suggested-by: Jens Wikander <jens.wiklander@linaro.org>

### DIFF
--- a/core/arch/arm/include/kernel/thread_private_arch.h
+++ b/core/arch/arm/include/kernel/thread_private_arch.h
@@ -49,11 +49,7 @@
 #define STACK_THREAD_SIZE	(8192 + CFG_STACK_THREAD_EXTRA)
 #endif
 
-#if TRACE_LEVEL > 0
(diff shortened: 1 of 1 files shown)


jbech-linaro:
`Acked-by: Joakim Bech <joakim.bech@linaro.org>`

jenswi-linaro (commented):

jforissier:
I have added more debugging info to the commit description. @jenswi-linaro thanks for pointing me in the right direction :wink:

PR #6762, opened 2024-03-20
stm32 drivers: remove use of undefined CFG_PM

etienne-lms:
Remove use of `CFG_PM` config switch that is not defined in OP-TEE OS.


etienne-lms:
@GseoC, could you have a look at these changes?

GseoC:

etienne-lms:
Comments addressed. Tag applied.

PR #6764, opened 2024-03-20
drivers: stm32_i2c: protect bus access with a mutex

etienne-lms:
Protect concurrent accesses to an STM32 I2C bus with a PM aware mutex.

### DIFF
--- a/core/drivers/stm32_i2c.c
+++ b/core/drivers/stm32_i2c.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: (GPL-2.0 OR BSD-3-Clause)
 /*
- * Copyright (c) 2017-2019, STMicroelectronics
+ * Copyright (c) 2017-2024, STMicroelectronics
  *
  * The driver API is defined in header file stm32_i2c.h.
  *
@@ -20,6 +20,7 @@
 #include <kernel/delay.h>
 #include <kernel/dt.h>
 #include <kernel/dt_driver.h>
+#include <kernel/mutex_pm_aware.h>
 #include <kernel/panic.h>
 #include <libfdt.h>
 #include <stdbool.h>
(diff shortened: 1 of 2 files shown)


etienne-lms:
> You should justify why using a pm_mutex instead of a regular one somewhere

I've added a word about that in the inline description comment. Is that okay to you?

core/drivers/stm32_i2c.c
```diff
@@ -753,6 +754,8 @@ int stm32_i2c_init(struct i2c_handle_s *hi2c,
 	vaddr_t base = 0;
 	uint32_t val = 0;
 
+	mutex_pm_aware_init(&hi2c->mu);
```
GseoC, line 757: Actually, shouldn't this be done at probe instead of this exported function?

GseoC:
> > You should justify why using a pm_mutex instead of a regular one somewhere
> 
> I've added a word about that in the inline description comment. Is that okay to you?

Should be OK as the type name is explicit

core/drivers/stm32_i2c.c
etienne-lms: We need to initialize the mutex from `stm32_i2c_init()` because this is the init function used by an Stm32 I2C device that do not rely on DT_DRIVER driver probing support, see _[core/drivers/crypto/se050/glue/i2c_stm32.c](https://github.com/OP-TEE/optee_os/blob/4.1.0/core/drivers/crypto/se050/glue/i2c_stm32.c#L89)_. Note that `stm32_i2c_init()` is called from the probe function, as you expect.

etienne-lms:
I think I've addressed all review comment. May I squash the fixup commits?

core/drivers/stm32_i2c.c
GseoC: Then stm32_i2c_init() should be called once? Can we protect against several use of this function per i2c handler?

etienne-lms:
squashed and tag applied.
thanks

PR #6772, opened 2024-03-26
vexpress-qemu_armv8a: increase CFG_CORE_HEAP_SIZE to 131072

jforissier:
Set the default core heap size for QEMUv8 to 128K because 64K is not enough to complete the "make check" test with CFG_RPMB_FS=y CFG_RPMB_WRITE_KEY=y.


jenswi-linaro:

PR #6775, opened 2024-03-29
drivers: Add RISC-V Zkr hardware random number generator support

lyctw:
The RISC-V Zkr entropy source extension provides NIST SP 800-90B or BSI AIS-31 compliant physical entropy source via seed CSR.

To allow OP-TEE OS (S-mode) to read the CSR, mseccfg.SSEED bit is expected to be set by M-mode. As we cannot read mseccfg here, use a test trap handler to verify the accessibility of seed.

### DIFF
--- a/core/arch/riscv/include/riscv.h
+++ b/core/arch/riscv/include/riscv.h
@@ -8,6 +8,7 @@
 
 #include <compiler.h>
 #include <encoding.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <sys/cdefs.h>
 #include <util.h>
@@ -394,6 +395,8 @@ static inline __noprof uint32_t read_cntfrq(void)
 {
 	return CFG_RISCV_MTIME_RATE;
 }
+
+__noprof bool riscv_detect_csr_seed(void);
 
 #endif /*__ASSEMBLER__*/
 
--- a/core/drivers/riscv_zkr_rng.c
+++ b/core/drivers/riscv_zkr_rng.c
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024 Andes Technology Corporation
+ */
+
+#include <crypto/crypto.h>
+#include <kernel/panic.h>
+#include <riscv.h>
+#include <rng_support.h>
+#include <tee/tee_cryp_utl.h>
+
+#define OPST_BIST 0b00
+#define OPST_WAIT 0b01
+#define OPST_ES16 0b10
+#define OPST_DEAD 0b11
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	uint8_t *ptr = buf;
+	uint32_t val = 0;
+
+	while (len > 0) {
+		/*
+		 * The seed register must be accessed using CSR read-write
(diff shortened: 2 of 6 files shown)


gagachang:

core/drivers/riscv_zkr_rng.c
```diff
@@ -0,0 +1,93 @@
...
+static unsigned long read_seed(void)
+{
+	unsigned long ret = 0;
+
+	/*
+	 * The seed register must be accessed using CSR read-write
+	 * instructions. The write operation is ignored and serves
+	 * to indicate polling and flushing.
+	 */
+	asm volatile("csrrw %0, seed, x0" : "=r" (ret) : : "memory");
```
jenswi-linaro, line 26: In the OP-TEE project, we avoid mixing inline assembly with C code. Implementing a single instruction like this as a static inline function in `core/arch/riscv/include/riscv.h` is preferred.

core/drivers/riscv_zkr_rng.c
```diff
@@ -0,0 +1,93 @@
...
+	bool seed_readable = true;
+	unsigned long tmp = 0;
+
+	/*
+	 * If seed CSR read permission in S-mode is not
+	 * granted by M-mode firmware, OP-TEE will raise
+	 * illegal instruction and jumps to a test trap
+	 * handler below to unset seed_readable.
+	 */
+	asm volatile ("  .align 4\n"
```
jenswi-linaro, line 72: This is too much as inline assembly, please move it into a `.S` file as a helper function.

lyctw: Moved to `readable_csr()` in core/arch/riscv/include/riscv.h

core/arch/riscv/include/riscv.h
```diff
@@ -77,6 +77,30 @@
 		__tmp;							\
 	})
 
+#define readable_csr(csr)						\
+	({								\
+		bool __readable = true;					\
+		unsigned long __tmp = 0;				\
+		asm volatile ("	j 1f\n"					\
```
jenswi-linaro, line 84: Pardon my bluntness, but this is just as horrible in a macro. Why don't you want to put it in a .S file implemented as a proper function?
Wouldn't the name `csr_seed_is_readable()` make more sense?
Does this code assume that interrupts and other exceptions are masked?

core/arch/riscv/include/riscv.h
```diff
@@ -77,6 +77,30 @@
...
+			"	.align 4\n"				\
+			"2:\n"						\
+			"	csrr %0, sepc\n"			\
+			"	addi %0, %0, 4\n"			\
+			"	csrw sepc, %0\n"			\
+			"	add %0, x0, x0\n"			\
+			"	sret\n"					\
```
jenswi-linaro, line 91: I guess this is a return instruction, that's bound to surprise other users of this macro.

lyctw: Agreed, let me fix it.

jenswi-linaro:

gagachang:
Hello @jforissier 
This PR is ready to be merged. Thanks!

jforissier:
@gagachang indeed, thanks for reminding me.

PR #6782, opened 2024-04-10
drivers: crypto: remove assertions on device handlers

etienne-lms:
Remove assertions added by the commit referred below. They are useless since the handlers are registered only if the related device (stm32_cryp or stm32_saes) has its driver successfully probed. These assertion also prevent enabling both CFG_STM32_SAES and CFG_STM32_CRYP for a platform which is a valid configuration for when we rely on the DT to state which of both is enabled.

Fixes: 03de2c7bb316 ("drivers: crypto: stm32_saes: fallback to software on 192bit AES keys")


etienne-lms:
@tbourgoi, could you have a look at this change?

tbourgoi:
Hello,
Fix looks OK to me, thanks.
I think there is some other condition that can be removed, but they are not related to this patch so I'll do them in a another PR.
You can add my tag
`Reviewed-by: Thomas Bourgoin <thomas.bourgoin@foss.st.com>`

PR #6784, opened 2024-04-11
drivers: regulator: fix error case in regulator_supported_voltages() + split asserts

etienne-lms:
2 changes:
* Fix case when `.supported_voltages()` handler is supported to assert voltage description when the handler succeeds and to use pre-filled fallback description when the handler returns `TEE_ERROR_NOT_SUPPORTED` instead of asserting @*desc content while that pointer was likely not initialized with a valid address.
Fixes: af5b9881111c ("drivers: regulator: supported voltage consider levels bounds")
* Split assertions in `regulator_supported_voltages()` to ease debugging, providing a better indication of which condition is not fulfilled.

### DIFF
--- a/core/drivers/regulator/regulator.c
+++ b/core/drivers/regulator/regulator.c
@@ -216,18 +216,18 @@ TEE_Result regulator_supported_voltages(struct regulator *regulator,
 					struct regulator_voltages_desc **desc,
 					const int **levels)
 {
+	TEE_Result res = TEE_ERROR_NOT_SUPPORTED;
+
 	assert(regulator && desc && levels);
 
-	if (regulator->ops->supported_voltages) {
-		TEE_Result res = TEE_ERROR_GENERIC;
-
+	if (regulator->ops->supported_voltages)
 		res = regulator->ops->supported_voltages(regulator, desc,
 							 levels);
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):
@gatien or @patrickdelaunay, may I ask you to have a look at these 2 changes?

GseoC:
The commit description of `drivers: regulator: fix error case in regulator_supported_voltages()` commit is a bit unclear to me, would you mind rephrasing it? maybe something like:

```
Fix case when .supported_voltages() handler exists but reports TEE_ERROR_NOT_SUPPORTED.
In this case, the regulator should use the fallback values.
```

I let you mix my suggestion with whatever you esteem right.

With this changed:

patrickdelaunay:
For the series:

Acked-by: Patrick Delaunay <patrick.delaunay@foss.st.com>

etienne-lms:
> The commit description of drivers: regulator: fix error case in regulator_supported_voltages() commit is a bit unclear to me, would you mind rephrasing it? maybe something like:

Thanks Gatien, indeed the sentence is far to long and not readable.

etienne-lms:
I've rephrased in 1st commit message and applied the review patches to both.

GseoC (approved):

GseoC:
Keep alive

PR #6785, opened 2024-04-11
core: make sure tee_entry_get_os_revision() uses a proper TEE_IMPL_GIT_SHA1

jforissier:
tee_entry_get_os_revision() stores TEE_IMPL_GIT_SHA1 into a 32 or 64-bit register, depending on the platform. Unfortunately the command that creates TEE_IMPL_GIT_SHA1 does not provide any guarantee that the value will fit. For instance it can happen that 8 characters are not enough to disambiguate two commits in the repository, in which case git rev-parse --short=8 will happily return 9 or more characters. In this case a 32-bit build would display a warning and TEE_IMPL_GIT_SHA1 would be truncated in a way we don't want (discarding the most significant bits).

Therefore, make sure TEE_IMPL_GIT_SHA1 is exactly 8 or 16 hexadecimal characters (plus the leading 0x).

### DIFF
--- a/core/arch/arm/kernel/thread_spmc.c
+++ b/core/arch/arm/kernel/thread_spmc.c
@@ -748,7 +748,8 @@ static void handle_blocking_call(struct thread_smc_args *args,
 	case OPTEE_FFA_GET_OS_VERSION:
 		spmc_set_args(args, direct_resp_fid, swap_src_dst(args->a1), 0,
 			      CFG_OPTEE_REVISION_MAJOR,
-			      CFG_OPTEE_REVISION_MINOR, TEE_IMPL_GIT_SHA1);
+			      CFG_OPTEE_REVISION_MINOR,
+			      TEE_IMPL_GIT_SHA1 >> 32);
 		break;
 	case OPTEE_FFA_EXCHANGE_CAPABILITIES:
 		sec_caps = OPTEE_FFA_SEC_CAP_ARG_OFFSET;
--- a/core/core.mk
+++ b/core/core.mk
@@ -11,6 +11,17 @@ include mk/config.mk
 # $(ARCH).mk also sets the compiler for the core module
 include core/arch/$(ARCH)/$(ARCH).mk
 
+ifeq ($(CFG_OS_REV_REPORTS_GIT_SHA1),y)
+ifeq ($(arch-bits-core),64)
+git-sha1-len := 16
+else
+git-sha1-len := 8
+endif
+TEE_IMPL_GIT_SHA1 := 0x$(shell git rev-parse --short=$(git-sha1-len) HEAD 2>/dev/null || echo 0 | cut -c -$(git-sha1-len))
+else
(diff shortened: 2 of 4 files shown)


core/core.mk
```diff
@@ -11,6 +11,17 @@ include mk/config.mk
 # $(ARCH).mk also sets the compiler for the core module
 include core/arch/$(ARCH)/$(ARCH).mk
 
+ifeq ($(CFG_OS_REV_REPORTS_GIT_SHA1),y)
+ifeq ($(arch-bits-core),64)
+git-sha1-len := 16
+else
+git-sha1-len := 8
+endif
+TEE_IMPL_GIT_SHA1 := 0x$(shell git rev-parse --short=$(git-sha1-len) HEAD 2>/dev/null | cut -c -$(git-sha1-len) || echo 0)
```
jenswi-linaro, line 20: Since we pipe through `cut -c -$(git-sha1-len)`, do we need the `--short=$(git-sha1-len)` parameter?

jforissier: No we don't. Updated.

GseoC: --short=$(git-sha1-len) will also change the number of character printed for 64bits platforms.
If nothing is specified, the number of characters will depend on each person's git configuration. Looking at the man page, it will default to at least 4. 
If the 4 first characters aren't ambiguous, it will remain 4 AFAICT. Else, I guess the command will add one more until it's not.

jforissier: > --short=$(git-sha1-len) will also change the number of character printed for 64bits platforms.

Yes and I believe that's fine.

> If nothing is specified, the number of characters will depend on each person's git configuration. [...]

Right. So `--short=$(git-sha1-len)` is needed after all. Updated, thanks!

jforissier:
Sorry for the multiple pushes, it helped me debug the CI errors. It should be good now.

core/arch/arm/kernel/thread_spmc.c
```diff
@@ -748,7 +748,8 @@ static void handle_blocking_call(struct thread_smc_args *args,
 	case OPTEE_FFA_GET_OS_VERSION:
 		spmc_set_args(args, direct_resp_fid, swap_src_dst(args->a1), 0,
 			      CFG_OPTEE_REVISION_MAJOR,
-			      CFG_OPTEE_REVISION_MINOR, TEE_IMPL_GIT_SHA1);
+			      CFG_OPTEE_REVISION_MINOR,
+			      SHIFT_U32(TEE_IMPL_GIT_SHA1, 16));
```
jenswi-linaro, line 752: What is this supposed to do?

jforissier: Oops. I want to keep the upmost 32 bits. Clearly not what this is doing :cry: `TEE_IMPL_GIT_SHA1 >> 16` maybe? :wink:

jforissier:
Update

jenswi-linaro (commented):

GseoC:
Considering this applies for git part:

PR #6791, opened 2024-04-16
Make libutee buffering AOSP compatible

samitolvanen:
In tee_buffer_update, libutee currently delays processing an input block until more space is needed in the buffer, which is perfectly valid behavior, but doesn't match AOSP compatibility requirements.

Specifically, both CTS ([testKatEncryptOneByteAtATime](https://android.googlesource.com/platform/cts/+/refs/heads/main/tests/tests/keystore/src/android/keystore/cts/BlockCipherTestBase.java#779)) and VTS ([EncryptionOperationsTest.*OneByteAtATime](https://android.googlesource.com/platform/hardware/interfaces/+/refs/heads/main/security/keymint/aidl/vts/functional/KeyMintAidlTestBase.cpp#827)) expect block cipher implementations to produce an output block as soon as a full block of input has been received. Change libutee behavior to be AOSP compatible.

### DIFF
--- a/lib/libutee/tee_api_operations.c
+++ b/lib/libutee/tee_api_operations.c
@@ -1118,6 +1118,12 @@ static TEE_Result tee_buffer_update(
 		l = ROUNDUP(op->buffer_offs + slen - buffer_size,
 				op->block_size);
 		l = MIN(op->buffer_offs, l);
+		/*
+		 * If we're buffering only a single block, process it
+		 * immediately.
+		 */
+		if (!op->buffer_two_blocks)
+			l = op->block_size;
 		tmp_dlen = dlen;
 		res = update_func(op->state, op->buffer, l, dst, &tmp_dlen);
 		if (res != TEE_SUCCESS)


lib/libutee/tee_api_operations.c
```diff
@@ -1118,6 +1118,9 @@ static TEE_Result tee_buffer_update(
 		l = ROUNDUP(op->buffer_offs + slen - buffer_size,
 				op->block_size);
 		l = MIN(op->buffer_offs, l);
+		/* If we have a full buffer, process it immediately */
```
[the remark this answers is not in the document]
samitolvanen: After staring at the code for a bit longer, it looks like if we ever end up in a situation where `op->buffer_two_blocks` is true, and `op->buffer_offs == buffer_size`, we'd end up processing an extra block with this patch when `1 <= slen <= op->block_size`.

Considering the subtlety of this code, I think gating this behind `!op->buffer_two_blocks` is the safest way forward despite the existing tests passing, and explicitly using `op->block_size` should make the intent more obvious:
```
if (!op->buffer_two_blocks && op->buffer_offs == op->block_size)
        l = op->block_size;
```
I confirmed that this is sufficient to fix the `EncryptionOperationsTest.*OneByteAtATime` VTS tests, and `xtest -l` is still happy.

lib/libutee/tee_api_operations.c
```diff
@@ -1118,6 +1118,12 @@ static TEE_Result tee_buffer_update(
 		l = ROUNDUP(op->buffer_offs + slen - buffer_size,
 				op->block_size);
 		l = MIN(op->buffer_offs, l);
+		/*
+		 * If we're buffering only a single block and have a full
+		 * buffer, process it immediately.
+		 */
+		if (!op->buffer_two_blocks && op->buffer_offs == op->block_size)
```
jenswi-linaro, line 1125: op->buffer_offs is guaranteed either op->block_size or 2 * op->block_size at this point, so it should be enough to check op->buffer_two_blocks.

jenswi-linaro (commented):
With my comment addressed:

`tee_buffer_update()` is a bit hard to understand and maintain. Once this is merged, I'll try to write some more tests targeting this function and hopefully be able to simplify it.

lib/libutee/tee_api_operations.c
samitolvanen: Sure, changed.

PR #6793, opened 2024-04-18
ldelf: Fix dumping physical address of ELF segment

gagachang:
Fix the strange values of "pa" when TA ELF mappings is dumped.
The function argument of print_seg() should be explicit physical address of mapping rather than offset of the segment.


jenswi-linaro (commented):

jforissier:

PR #6797, opened 2024-04-18
Import mbedtls 3.6.0

tomvaneyck:
This pull request imports MbedTLS 3.6.0, and additionally includes the PSA files necessary for TLS 1.3.


lib/libmbedtls/mbedtls/library/common.h
```diff
@@ -20,7 +20,16 @@
 #include <stddef.h>
 
 #if defined(__ARM_NEON)
+/* Undefine and restore __section and __data from compiler.h to prevent
```
jenswi-linaro, line 23:
``` C
/*
 * Please use this
 * style.
 */

**Review by jenswi-linaro — COMMENTED** [review]
Comments for the commit "libmbedtls allow inclusion of arm_neon.h"

Please add a `:` after `libmbedtls`.

File: lib/libmbedtls/core/rsa.c [hunk 2]
```diff
@@ -171,6 +177,16 @@ static void mbd_rsa_free(mbedtls_rsa_context *rsa, struct rsa_keypair *key)
 	mbedtls_rsa_free(rsa);
 }
 
+static void mbd_pk_free(mbedtls_pk_context *ctx, struct rsa_keypair *key)
+{
+	mbedtls_rsa_context *rsa = (mbedtls_rsa_context *)ctx->pk_ctx;
+
+	// Executing mbedtls_rsa_free twice is fine, as it does nothing if its
```
jenswi-linaro, line 184:
``` C
/*
 * Please comment
 * using this style.
 */

File: lib/libmbedtls/core/rsa.c [hunk 3]
```diff
@@ -454,11 +470,24 @@ TEE_Result sw_crypto_acipher_rsaes_decrypt(uint32_t algo,
 	size_t blen = 0;
 	size_t mod_size = 0;
 	void *buf = NULL;
-	mbedtls_rsa_context rsa = { };
+	mbedtls_pk_context ctx;
```
jenswi-linaro, line 473: Please initialize with `= { };`

Same for the rest of this file.

lib/libmbedtls/core/rsa.c
```diff
@@ -454,11 +470,24 @@ TEE_Result sw_crypto_acipher_rsaes_decrypt(uint32_t algo,
...
+		return TEE_ERROR_NOT_SUPPORTED;
+	}
+
+	mbedtls_pk_init(&ctx);
+	res = mbedtls_pk_setup(&ctx, pk_info);
+	if (res != 0) {
+		goto out;
+	}
+
+	rsa = (mbedtls_rsa_context *)ctx.pk_ctx;
```
jenswi-linaro, line 489: There's no need to cast a void pointer.

Same for the rest of this file.

jenswi-linaro (commented):
Comments for "libmbedtls: adjust use of rsa pk_wrap API"

jenswi-linaro (commented):
For "libmbedtls: reset minimum rsa key size"
Please apply:

jenswi-linaro:
Comments on the commit message of "Import mbedtls-3.6.0"

Please update the version in "cp -R path/to/mbedtls-3.4.0/mbedtls"

For "rm cmake/MbedTLSConfig.cmake.in", how about adding "cmake" to the list of directories we remove instead?

Please add .gitmodules and .readthedocs.yaml to the list of files to remove

jenswi-linaro:
I've replicated the source importing and it adds up.

tomvaneyck:
@jenswi-linaro, thank you for the review!

> For "rm cmake/MbedTLSConfig.cmake.in", how about adding "cmake" to the list of directories we remove instead?

I don't see a reason we couldn't, so I've added it to the requested changes.

jenswi-linaro:
For "libmbedtls: allow inclusion of arm_neon.h" with the amend, please apply:

jenswi-linaro:
For "libmbedtls: adjust use of rsa pk_wrap API" and "Import mbedtls-3.6.0" with the fixup/ammend, apply:

Please squash the amended commits.

jforissier:
AFAICT the CI error is totally unrelated the MbedTLS update. `xtest regression_4011` is painfully slow with QEMUv7, in fact the QEMUv7 `make check` is much much slower than QEMUv8 and therefore I believe it can fail randomly. 4011 runs 10 iterations which take ~1min 7s each on my build machine so a bit over 11 minutes for the whole test. The "make check" timeout is 15 minutes and the CI VM is probably slower than my 5GHz CPU.

jforissier:
> AFAICT the CI error is totally unrelated the MbedTLS update. 

Wrong.

> `xtest regression_4011` is painfully slow [...]
> 4011 runs 10 iterations which take ~1min 7s each on my build machine [...]

With mainline MBedTLS one iteration takes only 6 seconds so there is clearly something wrong.

jforissier:
> > AFAICT the CI error is totally unrelated the MbedTLS update.
> 
> Wrong.
> 
> > `xtest regression_4011` is painfully slow [...]
> > 4011 runs 10 iterations which take ~1min 7s each on my build machine [...]
> 
> With MBedTLS from mainline optee_os one iteration takes only 6 seconds so there is clearly something wrong.

jforissier:
The slowdown is related to the bignum functions in Mbed TLS (which are used even when LibTomcrypt is selected as the core crypto library). I measured that `mbedtls_mpi_core_exp_mod()` takes about 5.4s because `mbedtls_mpi_core_montmul()` takes about 1ms and is called 5470 times during `ta_entry_asymmetric_verify_digest()`. I also see that many constant time functions are used (`mbedtls_ct_*`).

jenswi-linaro:
Are we missing some configuration options? It seems like `mbedtls_mpi_core_montmul()` has had a nice optimization. But such an increase in the number of times it's called doesn't seem right.

tomvaneyck:
I think the increase in calls of `mbedtls_mpi_core_montmul` is because they changed the algorithm in this commit: https://github.com/Mbed-TLS/mbedtls/commit/1ba40585f9446372f69f19693c20f593283930af. This change introduces a call to `mbedtls_mpi_core_exp_mod` which was not there before (as far as I can tell). I don't know why this would make it that much slower though.

jforissier:
@tomvaneyck thanks for the hint! Indeed the commit you mention is the culprit. Here is a fix: https://github.com/jforissier/optee_os/commit/f147fc17b69efc9f22d122466d7b6b9678a6ad61. Could you please cherry-pick that patch and add it on top of your series?
IMO this issue should also be reported upstream.

tomvaneyck:
Done! It should now pass the CI.

jenswi-linaro:
For "libmbedtls: bignum: restore mbedtls_mpi_exp_mod() from v3.5.2", please apply:

jforissier:
So, regardless of the QEMUv7 performance regression which may take quite a long time to investigate, and which we have worked around here, I believe this is ready for merging. Then a squash commit can be created to be merged in master.

Thanks @tomvaneyck for the good work :+1:

jforissier:
> Then a squash commit can be created to be merged in master.

... @tomvaneyck would you please make that squash commit and create a new PR against master? Or I can do it if you prefer.

tomvaneyck:
Sure, I can create the squash commit! I'll create the pull request tomorrow.

PR #6806, opened 2024-04-26
core: riscv: Fix store 32-bit thread_core_local flags

gagachang:
The thread_core_local flags is a 32-bit variable.
Thus, we must explicitly use "lw" and "sw" instructions, which mean load and store 32-bit value from specific memory address, to operate the thread_core_local flags.


gagachang:
Hello @jforissier , Would you like to give this small fix a tag?

jforissier (commented):

Minor: you should not mention the lw instruction in the commit description since the commit is only using sw.

gagachang:
@jforissier Thanks! I've fixed the commit message and applied the tags.

PR #6811, opened 2024-04-30
core: mmu: fix collect_device_mem_ranges() to pass size

ysbnim:
The 6th parameter of add_phys_mem() has to be the size of memory region and not the end address of the region.

### DIFF
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -997,7 +997,7 @@ static void collect_device_mem_ranges(struct tee_mmap_region *memory_map,
 		}
 
 		add_phys_mem(memory_map, num_elems, name, MEM_AREA_IO_SEC,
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):
Could you mention in the header line it's FF-A/SPMC-EL2 related. Or maybe:
```
core: mmu: fix memory regions found from ff-a manifest

Fix the 6th parameter of add_phys_mem() in collect_device_mem_ranges()
that has to be the size of memory region and not the end address of the
region.

S-o-b: ...
```

Could you add a fix tag?
`Fixes: 72a6827a6353 ("core: arm: SPMC obtain device memory info from DTB")`
with that tag added: 

ysbnim:
Thank you for your review. I changed the commit message and added fix and reviewed tag.

jenswi-linaro (commented):

PR #6812, opened 2024-05-02
core: callout: disable obsolete timeouts

jenswi-linaro:
In callout_service_cb() when a timeout interrupt is received there's a check to see if this is the last scheduled CPU. If not the interrupt is ignored, but not disabled causing it to trigger again and again. So fix this by disabling the timeout too.

Fixes: cf707bd0d695 ("core: add callout service")


jforissier (commented):

jenswi-linaro:
Tag applied

PR #6815, opened 2024-05-03
ta: pkcs11: correct RSA keys extended attributes sanitation

etienne-lms:
Fix RSA key attributes function load_tee_rsa_key_attrs() that badly checks that the 5 extended RSA attributes are found in the key object.

Link: https://github.com/OP-TEE/optee_test/issues/721#issuecomment-2068055537
Link: https://github.com/OP-TEE/optee_test/issues/721#issuecomment-2072064963
Fixes: 0442c956edfb ("ta: pkcs11: Add support for RSA signing & verification")

### DIFF
--- a/ta/pkcs11/src/processing_rsa.c
+++ b/ta/pkcs11/src/processing_rsa.c
@@ -569,6 +569,8 @@ enum pkcs11_rc load_tee_rsa_key_attrs(TEE_Attribute **tee_attrs,
 			break;
 		}
 
+		rc = PKCS11_CKR_GENERAL_ERROR;
+
 		if (pkcs2tee_load_attr(&attrs[count], TEE_ATTR_RSA_PRIME1, obj,
 				       PKCS11_CKA_PRIME_1))
 			count++;


etienne-lms:
@anvisriv, do you have any comments on this change?
@vesajaaskelainen, maybe you want to have a look at this change.

vesajaaskelainen:
> @anvisriv, do you have any comments on this change? @vesajaaskelainen, maybe you want to have a look at this change.

Sure.

If I understood the problem correctly it is about incorrect templates given for RSA pre-computed values -- when you operate with happy templates then it is OK but if some RSA argument is missing the it gives incorrect return value.

There needs to be at least `PKCS11_CKA_PRIME_1` in order to trigger the processing of the pre-computed values -- if that is missing and other fields are there then all the rest of pre-computed values will be silently discarded. Unless there is compliance problem with that I would just keep that functionality as is. I suppose the operation will just be a bit slower.

So, your change looks good.

```
Reviewed-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>
```

etienne-lms:
> If I understood the problem correctly it is about incorrect templates given for RSA pre-computed values 

Yes, It is about a GPD TEE Internal API constraint, spec v1.3, table 5-10, row TEE_TYPE_RSA_KEYPAIR:
*"The CRT parameters are optional. If any of these attributes is provided, then all of them SHALL be provided:"*
The current implementation did not validate that "all of them SHALL be provided". I prefer the TA to conform with the GPD TEE TA API spec.
Maybe I could add this spec reference in the commit message.

> There needs to be at least PKCS11_CKA_PRIME_1 in order to trigger the processing of the pre-computed values -- if that is missing and other fields are there then all the rest of pre-computed values will be silently discarded.

Yes, this behavior is preserved.

anvisriv:
Hi @etienne-lms 
As discussed in https://github.com/OP-TEE/optee_test/issues/721#issuecomment-2109820370
The count value will keep increasing till 8 even if the client only passes 5 attributes. As [get_attribute_ptrs](https://github.com/OP-TEE/optee_os/blob/master/ta/pkcs11/src/attributes.c#L103), doesn't differentiate between NULL and Non-NULL params.

vesajaaskelainen:
Did you do test case for xtest that exercises this to demonstrate that the fix is working?

etienne-lms:
Hi @vesajaaskelainen. No, i've not writen any extra test.

From @anvisriv last comment and looking again at the implementation, I wonder if we really need to check the attribute size (no NULL pointer/no size==0). The GPD TEE say that the attributes are needed by group. But nothing about the size of the attribute. 

I think the current change proposal is fine. The attribute IDs are loaded by group or operation fails. Don't check  if the attribute size is zero or not in `pkcs2tee_load_attr()`.

@anvisriv what do you think?

anvisriv:
Hi @etienne-lms 

Please correct me in understanding. Are we saying that if client sends 5 parameters then TA will populate the missing parameters with size 0 and NULL?
If yes, then is this change really required? The TA is already doing this.

etienne-lms:
Sorry for the late reply.
Yes, I think this change should be merged. In the current mainline implementation, if client provides a (pkcs#11) object template with `PKCS11_CKA_PRIME_1` attribute but not its companion attributes `PKCS11_CKA_PRIME_2`, `PKCS11_CKA_EXPONENT_1/2` and `PKCS11_CKA_COEFFICIENT`, then the object will be successfully created but client will likely get an error (possibly a pkcs11 TA panic) when the object is used as some GP TEE attributes are missing, not conforming to the GP TEE Internal Core API specification.

etienne-lms:
I confirm my belief that the change is needed.
(code extracts below relate to the state of the implementation in OP-TEE release tag 4.2.0)

`pkcs2tee_load_attr()` does not load the TEE attribute when the related PKCS11 attribute is not found in the object template. The function initialize the attribute only if `get_attribute_ptrs()` returns true: (4.20. tag implementation below)
https://github.com/OP-TEE/optee_os/blob/12d7c4ee4642d2d761e39fbcf21a06fb77141dea/ta/pkcs11/src/pkcs11_helpers.c#L650-L655

When the PKCS11 attribute is not found in the object (template), `get_attribute_ptr()` returns `PKCS11_RV_NOT_FOUND`:
https://github.com/OP-TEE/optee_os/blob/12d7c4ee4642d2d761e39fbcf21a06fb77141dea/ta/pkcs11/src/attributes.c#L152-L160

Therefore, if some PKCS11 attributes are missing, the TEE object will lack some the the GP TEE attributes the spec expects.

For consistency of the implementation, the pkcs11 TA should report an error as proposed in this P-R.

jforissier:
@etienne-lms it seems this PR is ready to be merged, can you please apply @vesajaaskelainen's R-b? Thanks.

etienne-lms:
R-b tag applied. Thanks for the reminder.

PR #6819, opened 2024-05-06
core: riscv: Improve the trap handler

gagachang:
In current RISC-V trap handler, we always save the context into `struct thread_trap_regs` first, and copy the saved context into another specific structure to handle that trap. For example, `thread_user_ecall_handler()` copies the `struct thread_trap_regs` to be `struct thread_scall_regs` before handling this system call.
In fact, such copies may drop the performance and they should be eliminated.

In this commit, the specific data structure used to save the context is determined early in the trap handler by checking the type of the trap from CSR XCAUSE. Thus, the copies between context structures are eliminated.

### DIFF
--- a/core/arch/riscv/include/kernel/thread_arch.h
+++ b/core/arch/riscv/include/kernel/thread_arch.h
@@ -83,14 +83,14 @@ struct thread_abort_regs {
 	unsigned long t4;
 	unsigned long t5;
 	unsigned long t6;
-	unsigned long status;
-	unsigned long cause;
 	unsigned long epc;
+	unsigned long status;
+	unsigned long ie;
+	unsigned long cause;
 	unsigned long tval;
-	unsigned long satp;
-};
+} __aligned(16);
 
-struct thread_trap_regs {
+struct thread_scall_regs {
 	unsigned long ra;
 	unsigned long sp;
 	unsigned long gp;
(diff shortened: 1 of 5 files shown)


gagachang:
Hi @maroueneboubakri , Would you like to review this PR?

lyctw:
Reviewed-by: Yu Chien Peter Lin <peterlin@andestech.com>

maroueneboubakri:
Hi @gagachang,

Thanks for this optimization, kindly let me try it.

Best
Maro

jforissier:
Re. the CI error with QEMUv7, it is unrelated and fixed in https://github.com/OP-TEE/manifest/pull/280.

gagachang:
> Hi @gagachang,
> 
> Thanks for this optimization, kindly let me try it.
> 
> Best Maro

Sure! Fell free to ask the questions.

gagachang:
Hello @maroueneboubakri , Any progress or issue on this ?

gagachang:
Just rebase to master, nothing changed.

gagachang:
@maroueneboubakri Thank you!!
BTW, we read your paper and learned a lot from it.
https://ieeexplore.ieee.org/abstract/document/10551366
Do you have plan to upstream your code ? Especially the RVK and RVB extensions.

PR #6821, opened 2024-05-07
core: spmc: fix direct request handler

balint-dobszay-arm:
The FF-A direct request handling has an error: if the destination ID is the SPMC ID, the handler is trying to forward the message to an SP with this ID, which is obviously non-existent so this gives an error.


jenswi-linaro (commented):

balint-dobszay-arm:
Thanks, tag applied.

PR #6823, opened 2024-05-08
drivers:crypto:hisilicon:Add HASH and HMAC algorithm

LeiSen62:
Add HASH and HMAC algorithm by SEC, and support SHA1, SHA224, SHA256, SHA384, SHA512, MD5, SM3, and HMAC algorithms based on these algorithms.


jenswi-linaro:
I'm concerned with the header files in `core/drivers/crypto/hisilicon/include`. They are in the global include path so they can potentially be included by some .h file and made available to any .c file. It seems that they are not intended to be used in that way, but that's not entirely clear.

I would prefer if all .h files here were kept in `core/drivers/crypto/hisilicon` and included as `#include "..."`. Then it's clear that they are internal .h files that aren't supposed to be included in source files outside of this directory. What do others think?

LeiSen62:
The header files in core/drivers/crypto/hisilicon/includes are only used by the source files in core/drivers/crypto/hisilicon, so I also think it would be better to put all .h files in core/drivers/crypto/hisilicon/.

jenswi-linaro:
Thanks, then we're on the same page. Please add a patch on top of this that moves the .h files as we agreed.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+
+static void sec_digest_fill_long_bd2(struct hashctx *ctx,
+				     struct hisi_sec_sqe *sqe)
+{
+	uint64_t total_bits = 0;
+
+	if (ctx->has_next && !ctx->iv_len) {
+		/* LONG BD FIRST */
+		sqe->ai_apd_cs |= AI_GEN_INNER;
+		sqe->ai_apd_cs |= (AUTHPAD_NOPAD << SEC_APAD_OFFSET);
```
jenswi-linaro, line 37: `SHIFT_U32()` or `SHIFT_U64()` are generally preferred since they are safer.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+	hash_ctx->has_next = false;
+	hash_ctx->long_data_len = 0;
+	hash_ctx->scene = SCENE_NOTHING;
+
+	/*
+	 * In reset ctx scenarios, sec_hash_initialize will be called.
+	 * To ensure in data is NULL, reset ctx need to free in data
+	 * which is not NULL.
+	 */
+	if (hash_ctx->in) {
```
jenswi-linaro, line 281: There's no need to check this since `free(NULL)` does nothing.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+{
+	struct crypto_hash *hash;
+	struct hashctx *hash_ctx;
+
+	if (!ctx) {
+		EMSG("Input ctx is NULL");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	hash = to_hash_ctx(ctx);
```
jenswi-linaro, line 304: `to_hash_ctx()` doesn't return `NULL`.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+		hash_ctx->key_len = len;
+		memcpy(hash_ctx->key, key, len);
+	}
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result sec_hash_initialize(struct crypto_hash_ctx *ctx)
+{
+	struct crypto_hash *hash;
```
jenswi-linaro, line 296: Please initialize all local variables. Same for the rest of this file.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+	if (!hash_ctx->in) {
+		if (len <= SMALL_BUF_SIZE)
+			hash_ctx->buf_len = SMALL_BUF_SIZE;
+		else if (len <= MAX_AUTH_LENGTH)
+			hash_ctx->buf_len = ALIGN(len, HISI_QM_ALIGN128);
+		else
+			hash_ctx->buf_len = MAX_AUTH_LENGTH;
+
+		hash_ctx->in_len = 0;
+		hash_ctx->in = (uint8_t *)malloc(hash_ctx->buf_len);
```
jenswi-linaro, line 332: No need to cast a void pointer.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+			memcpy(hash_ctx->in + hash_ctx->in_len, data,
+			       left_size);
+			hash_ctx->in_len = hash_ctx->buf_len;
+			hash_ctx->scene = SCENE_STREAM;
+			hash_ctx->has_next = true;
+			data += left_size;
+			len -= left_size;
+			ret = sec_digest_do_task(hash_ctx->qp, hash_ctx);
+			if (ret) {
+				EMSG("Fail to do digest task! ret = 0x%x",
```
jenswi-linaro, line 362: `s/0x%x"/%#"PRIx32/`

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	hash_ctx->has_next = false;
+	ret = sec_digest_do_task(hash_ctx->qp, hash_ctx);
+	if (ret) {
+		EMSG("Fail to do digest task! ret = 0x%x.", ret);
+		return ret;
+	}
+
+	if (hash_ctx->mac_len > len)
```
jenswi-linaro, line 417: How about
``` C
memcpy(digest, hash_ctx->out, MIN(len, hash_ctx->mac_len));
```

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+{
+	hisi_qm_release_qp(hash_ctx->qp);
+
+	if (hash_ctx->in) {
+		free(hash_ctx->in);
+		hash_ctx->in = NULL;
+	}
+
+	memzero_explicit(hash_ctx->key,
+			 sizeof(uint8_t) * SEC_DIGEST_MAX_KEY_SIZE);
```
jenswi-linaro, line 444: I think we can agree that `sizeof(uint8_t)` is 1.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+	struct hashctx *hash_ctx = hash->ctx;
+
+	return sec_digest_do_final(hash_ctx, digest, len);
+}
+
+void sec_digest_ctx_free(struct hashctx *hash_ctx)
+{
+	hisi_qm_release_qp(hash_ctx->qp);
+
+	if (hash_ctx->in) {
```
jenswi-linaro, line 438: No need to check `hash_ctx->in`.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+
+static void sec_hash_ctx_free(struct crypto_hash_ctx *ctx)
+{
+	struct crypto_hash *hash;
+	struct hashctx *hash_ctx;
+
+	if (!ctx)
+		return;
+
+	hash = to_hash_ctx(ctx);
```
jenswi-linaro, line 457: This can't be `NULL`

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+			   struct hashctx *in_hash_ctx)
+{
+	out_hash_ctx->iv_len = in_hash_ctx->iv_len;
+	out_hash_ctx->buf_len = in_hash_ctx->buf_len;
+	out_hash_ctx->key_len = in_hash_ctx->key_len;
+	out_hash_ctx->has_next = in_hash_ctx->has_next;
+	out_hash_ctx->long_data_len = in_hash_ctx->long_data_len;
+
+	if (in_hash_ctx->in) {
+		out_hash_ctx->in = (uint8_t *)malloc(out_hash_ctx->buf_len);
```
jenswi-linaro, line 481: No need to cast void pointers.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+
+static struct crypto_hash_ops hash_ops = {
+	.init = sec_hash_initialize,
+	.update = sec_hash_do_update,
+	.final = sec_hash_do_final,
+	.free_ctx = sec_hash_ctx_free,
+	.copy_state = sec_hash_copy_state,
+};
+
+struct crypto_hash *to_hash_ctx(struct crypto_hash_ctx *ctx)
```
jenswi-linaro, line 532: This function should be static

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,674 @@
...
+	if (done != SEC_HW_TASK_DONE || sqe->error_type) {
+		EMSG("SEC BD3 fail! done=0x%x, etype=0x%x",
+		     done, sqe->error_type);
+		return HISI_QM_DRVCRYPT_IN_EPARA;
+	}
+
+	return HISI_QM_DRVCRYPT_NO_ERR;
+}
+
+TEE_Result sec_digest_ctx_init(struct hashctx *hash_ctx,
```
jenswi-linaro, line 262: Global functions should use a prefix. Wouldn't `hisi_sec_` make sense in this file? Same for all the other globals in this file.

core/drivers/crypto/hisilicon/sec_hmac.c
```diff
@@ -0,0 +1,177 @@
...
+	struct crypto_hmac *hash = NULL;
+	struct hashctx *hash_ctx = NULL;
+
+	if (!ctx || !key) {
+		EMSG("Input ctx is NULL");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	hash = to_hmac_ctx(ctx);
+	if (!hash) {
```
jenswi-linaro, line 25: `to_hmac_ctx()` doesn't return NULL.

core/drivers/crypto/hisilicon/sec_hmac.c
```diff
@@ -0,0 +1,177 @@
...
+
+static struct crypto_mac_ops hash_ops = {
+	.init = sec_hmac_initialize,
+	.update = sec_hmac_do_update,
+	.final = sec_hmac_do_final,
+	.free_ctx = sec_hmac_ctx_free,
+	.copy_state = sec_hmac_copy_state,
+};
+
+struct crypto_hmac *to_hmac_ctx(struct crypto_mac_ctx *ctx)
```
jenswi-linaro, line 120: This function should be static.

core/drivers/crypto/hisilicon/sec_hash.c
LeiSen62: ai_apd_cs is u8

jenswi-linaro: Sure in this particular case there no issue since `SEC_APAD_OFFSET` is `2`, but you you need to crosscheck to see that it isn't 31 in case `AUTHPAD_NOPAD` is a signed 32-bit integer. So it's better to use these shift macros where possible.

core/drivers/crypto/hisilicon/sec_hash.c
LeiSen62: That's good.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,663 @@
...
+		ctx->iv_len = 0;
+	} else {
+		/* SHORT BD */
+		ctx->iv_len = 0;
+	}
+}
+
+static enum hisi_drv_status sec_digest_fill_bd3_sqe(void *bd, void *msg)
+{
+	struct hisi_sec_bd3_sqe *sqe = (struct hisi_sec_bd3_sqe *)bd;
```
jenswi-linaro, line 192: No need to cast a void pointer, same below.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,663 @@
...
+	enum hisi_drv_status ret = HISI_QM_DRVCRYPT_NO_ERR;
+	struct hashctx *ctx = (struct hashctx *)msg;
+	uint32_t alg_type = 0;
+
+	sqe->bd_param = (BD_TYPE3 | SHIFT_U32(ctx->scene, SEC_SCENE_OFFSET_V3));
```
jenswi-linaro, line 197: The outer `()` are not needed.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,663 @@
...
+	sqe->data_src_addr = ctx->in_dma;
+	sqe->mac_addr = ctx->out_dma;
+
+	if (ctx->mode == WCRYPTO_DIGEST_HMAC) {
+		ret = sec_digest_set_hmac_bd3_key(ctx, sqe);
+		if (ret)
+			return HISI_QM_DRVCRYPT_IN_EPARA;
+	}
+
+	sqe->auth_mac_key |= SHIFT_U64((ctx->mac_len / SEC_ENCODE_BYTES),
```
jenswi-linaro, line 209: The inner `()` are not needed.

core/drivers/crypto/hisilicon/sec_hash.c
```diff
@@ -0,0 +1,663 @@
...
+	struct crypto_hash *hash = NULL;
+	struct hashctx *hash_ctx = NULL;
+	TEE_Result ret = TEE_SUCCESS;
+
+	if (!ctx) {
+		EMSG("Ctx is NULL");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	hash = (struct crypto_hash *)calloc(1, sizeof(*hash));
```
jenswi-linaro, line 623: No need to cast a void pointer, please fix all the places in this patch.

jenswi-linaro (commented):
In the title of the commit messages, please add a single space after each colon.

jenswi-linaro:

LeiSen62:
Hi @jforissier ,all review comments have been modified. Can you help me merge the patch?

jforissier:
A few space characters added to the second commit subject, and pushed to master. Thanks!

PR #6831, opened 2024-05-13
drivers: sam: update sama7g5 configurations for wakeup from low-power

TonyHan11:


core/drivers/pm/sam/at91_pm.c
```diff
@@ -405,6 +405,14 @@ static TEE_Result at91_pm_sram_init(const void *fdt)
 
 	at91_pm_copy_suspend_to_sram();
 
+	/*
+	 * In order to run code for low-power out of SRAM without abort,
+	 * configure regions with write permission with not forced to
+	 * XN (Execute-never) attribute.
+	 */
+	if (read_id_pfr1() & GENMASK_32(15, 12))
```
jenswi-linaro, line 433: Please add and use proper defines for both the mask and the bit below.

jenswi-linaro: Where do you restore the XN bit?

core/drivers/pm/sam/at91_pm.c
```diff
@@ -395,7 +395,7 @@ static TEE_Result at91_pm_sram_init(const void *fdt)
 	at91_suspend_sram_pbase = virt_to_phys((void *)at91_suspend_sram_base);
 
 	/* Map the secure ram suspend code to be executable */
```
jenswi-linaro, line 417: Please update this comment to match what you're doing.

core/drivers/pm/sam/at91_pm.c
```diff
@@ -236,6 +236,7 @@ static int at91_pm_config_ws_ulp1(bool set)
  */
 static bool at91_pm_verify_clocks(void)
 {
+	int pck_cnt = 4;
```
jenswi-linaro, line 239: Please add and use an `AT91_PMC_PKC_COUNT` define depending on `CFG_SAMA7G5` instead of using magic numbers here.

TonyHan11:
Thank you @jenswi-linaro . Code & commit message updated according to the comments.

core/drivers/pm/sam/at91_pm.c
```diff
@@ -298,6 +311,7 @@ static TEE_Result at91_enter_backup(void)
 TEE_Result atmel_pm_suspend(uintptr_t entry, struct sm_nsec_ctx *nsec)
 {
 	TEE_Result res = TEE_ERROR_GENERIC;
+	int sctlr = 0;
```
jenswi-linaro, line 314: This should be of an unsigned type, either `uint32_t` or perhaps `unsigned long`.

core/drivers/pm/sam/at91_pm.c
```diff
@@ -311,13 +325,27 @@ TEE_Result atmel_pm_suspend(uintptr_t entry, struct sm_nsec_ctx *nsec)
 
 	sm_save_unbanked_regs(&nsec->ub_regs);
 
+	/*
+	 * In order to run code for low-power out of SRAM without abort,
+	 * configure regions with write permission with not forced to
+	 * XN (Execute-never) attribute.
+	 */
+	if (read_id_pfr1() & IDPFR1_VIRT_MASK) {
```
jenswi-linaro, line 333: Wouldn't it make sense to read `SCTLR` and if `SCTLR_WXN` is set clear it, regardless of `ID_PFR1`?
The bit is reserved as `0` if not supported.

core/drivers/pm/sam/at91_pm.c
```diff
@@ -311,13 +325,27 @@ TEE_Result atmel_pm_suspend(uintptr_t entry, struct sm_nsec_ctx *nsec)
...
+		sctlr = read_sctlr();
+		write_sctlr(sctlr & ~SCTLR_WXN);
```
jenswi-linaro, line 335: From the Arm ARM it looks like you need to invalidate the TLB also.

core/drivers/pm/sam/at91_pm.c
```diff
@@ -311,13 +325,27 @@ TEE_Result atmel_pm_suspend(uintptr_t entry, struct sm_nsec_ctx *nsec)
...
+
 	if (soc_pm.mode == AT91_PM_BACKUP) {
 		res = at91_enter_backup();
 	} else {
 		at91_suspend_sram_fn(&soc_pm);
 		res = TEE_SUCCESS;
 	}
 
+	/* Restore the XN attribute */
+	if (read_id_pfr1() & IDPFR1_VIRT_MASK)
```
jenswi-linaro, line 346: How about checking if `SCTLR_WXN` is set in the `sctlr` variable and only write it then?
I believe the TLB should be invalidated if `SCTLR_WXN` is changed in `SCTLR` here too.

core/drivers/pm/sam/at91_pm.c
jenswi-linaro: This comment is not addressed.

TonyHan11:
Updated according to the comments, thank you.

core/drivers/pm/sam/at91_pm.c
```diff
@@ -311,13 +326,34 @@ TEE_Result atmel_pm_suspend(uintptr_t entry, struct sm_nsec_ctx *nsec)
...
+	/*
+	 * In order to run code for low-power out of SRAM without abort,
+	 * configure regions with write permission with not forced to
+	 * XN (Execute-never) attribute.
+	 */
+	if (IS_ENABLED(CFG_HWSUPP_MEM_PERM_WXN)) {
+		sctlr = read_sctlr();
+		if (sctlr & SCTLR_WXN) {
+			write_sctlr(sctlr & ~SCTLR_WXN);
+			tlbi_all();
```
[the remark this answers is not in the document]
TonyHan11: Rebased with the review tags added and thank you for you suggestion. Currently we do not consider about the `ASID` and just use `tlbi_all()` as the simplest way to make it work.

Issue #6801, opened 2024-04-22
SingleInstance, MultiSession TA is loaded multiple times


KhOlegDc:
Hello, 

I have two CAs, connecting to the same TA. I want that only one TA instance will be in memory, serving requests from both CAs. I understand that requests will be served sequentially, that's OK and even desired. The idea is to share TA's RAM (heap objects) when serving requests from different CAs.

To achieve this, I configured TA:
```
#define TA_FLAGS      (TA_FLAG_SINGLE_INSTANCE | TA_FLAG_MULTI_SESSION)
```
and initialized each of CAs:
```
    TEEC_Result res = TEEC_InitializeContext(NULL, &teec_context);
    if (res != TEEC_SUCCESS) {
        return ERROR;
    }

    uint32_t resOrigin = 0;
    res = TEEC_OpenSession(&teec_context, &teec_session, &uuid, TEEC_LOGIN_PUBLIC, NULL, NULL,
                           &resOrigin);
    if (res != TEEC_SUCCESS) {
        TEEC_FinalizeContext(&teec_context);
        return ERROR;
    }
    return OK;
```

I run both CAs, making sure that sessions from both overlap in time. However, in the log I see that a separate TA instance is loaded for each CA session:
```
D/LD:  ldelf:176 ELF (fbaecc2c-xxxx-xxxx-xxxx-xxxxxxxxxxxx) at 0x4005a000  <=== first instance
E/TA:  TA_CreateEntryPoint:40 TA_CreateEntryPoint called
E/TA:  TA_CreateEntryPoint:48 Property SingleInstance is set to true
E/TA:  TA_CreateEntryPoint:57 Property MultiSession is set to true
E/TA:  TA_OpenSessionEntryPoint:84 TA_OpenSessionEntryPoint called
E/TA:  TA_OpenSessionEntryPoint:95 Session opened, TA version: 1.1.0
D/LD:  ldelf:176 ELF (fbaecc2c-xxxx-xxxx-xxxx-xxxxxxxxxxxx) at 0x4003d000   <=== second one
E/TA:  TA_CreateEntryPoint:40 TA_CreateEntryPoint called
E/TA:  TA_CreateEntryPoint:48 Property SingleInstance is set to true
E/TA:  TA_CreateEntryPoint:57 Property MultiSession is set to true
E/TA:  TA_OpenSessionEntryPoint:84 TA_OpenSessionEntryPoint called
[...]

PR #6836, opened 2024-05-14
core: fix race condition on single instance TA loading

etienne-lms:
Fix race condition on single instance TA creation where several instances of a single instance TA could be created if invoked close enough that they are both created after tee_ta_init_session() calls tee_ta_init_session_with_context().

Closes: https://github.com/OP-TEE/optee_os/issues/6801

### DIFF
--- a/core/kernel/pseudo_ta.c
+++ b/core/kernel/pseudo_ta.c
@@ -294,6 +294,12 @@ TEE_Result tee_ta_init_pseudo_ta_session(const TEE_UUID *uuid,
 	struct tee_ta_ctx *ctx;
 	const struct pseudo_ta_head *ta;
 
+	/*
+	 * Caller is expected to hold tee_ta_mutex for safe changes
+	 * in @s and registering of the context in tee_ctxes list.
+	 */
+	assert(mutex_is_locked(&tee_ta_mutex));
+
 	DMSG("Lookup pseudo TA %pUl", (void *)uuid);
 
 	ta = SCATTERED_ARRAY_BEGIN(pseudo_tas, struct pseudo_ta_head);
@@ -306,7 +312,6 @@ TEE_Result tee_ta_init_pseudo_ta_session(const TEE_UUID *uuid,
 		ta++;
 	}
 
-	/* Load a new TA and create a session */
 	DMSG("Open %s", ta->name);
 	stc = calloc(1, sizeof(struct pseudo_ta_ctx));
 	if (!stc)
@@ -319,10 +324,8 @@ TEE_Result tee_ta_init_pseudo_ta_session(const TEE_UUID *uuid,
 	ctx->ts_ctx.uuid = ta->uuid;
 	ctx->ts_ctx.ops = &pseudo_ta_ops;
 
-	mutex_lock(&tee_ta_mutex);
 	s->ts_sess.ctx = &ctx->ts_ctx;
 	TAILQ_INSERT_TAIL(&tee_ctxes, ctx, link);
-	mutex_unlock(&tee_ta_mutex);
 
 	DMSG("%s : %pUl", stc->pseudo_ta->name, (void *)&ctx->ts_ctx.uuid);
 
--- a/core/kernel/tee_ta_manager.c
+++ b/core/kernel/tee_ta_manager.c
@@ -617,22 +617,33 @@ static TEE_Result tee_ta_init_session(TEE_ErrorOrigin *err,
 
 	/* Look for already loaded TA */
 	res = tee_ta_init_session_with_context(s, uuid);
-	mutex_unlock(&tee_ta_mutex);
-	if (res == TEE_SUCCESS || res != TEE_ERROR_ITEM_NOT_FOUND)
+	if (res == TEE_SUCCESS || res != TEE_ERROR_ITEM_NOT_FOUND) {
+		mutex_unlock(&tee_ta_mutex);
 		goto out;
+	}
 
 	/* Look for secure partition */
 	res = stmm_init_session(uuid, s);
-	if (res == TEE_SUCCESS || res != TEE_ERROR_ITEM_NOT_FOUND)
+	if (res == TEE_SUCCESS || res != TEE_ERROR_ITEM_NOT_FOUND) {
+		mutex_unlock(&tee_ta_mutex);
+		if (res == TEE_SUCCESS)
+			res = stmm_complete_session(s);
+
 		goto out;
+	}
 
 	/* Look for pseudo TA */
 	res = tee_ta_init_pseudo_ta_session(uuid, s);
-	if (res == TEE_SUCCESS || res != TEE_ERROR_ITEM_NOT_FOUND)
+	if (res == TEE_SUCCESS || res != TEE_ERROR_ITEM_NOT_FOUND) {
+		mutex_unlock(&tee_ta_mutex);
 		goto out;
+	}
 
 	/* Look for user TA */
 	res = tee_ta_init_user_ta_session(uuid, s);
+	mutex_unlock(&tee_ta_mutex);
+	if (res == TEE_SUCCESS)
+		res = tee_ta_complete_user_ta_session(s);
 
 out:
 	if (!res) {
--- a/core/kernel/tee_ta_manager.c
+++ b/core/kernel/tee_ta_manager.c
@@ -527,8 +527,7 @@ static TEE_Result tee_ta_init_session_with_context(struct tee_ta_session *s,
 		if (!ctx)
 			return TEE_ERROR_ITEM_NOT_FOUND;
 
-		if (!is_user_ta_ctx(&ctx->ts_ctx) ||
-		    !to_user_ta_ctx(&ctx->ts_ctx)->uctx.is_initializing)
+		if (!ctx->is_initializing)
 			break;
 		/*
 		 * Context is still initializing, wait here until it's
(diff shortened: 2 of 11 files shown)


core/kernel/user_ta.c
```diff
@@ -486,6 +486,22 @@ TEE_Result tee_ta_init_user_ta_session(const TEE_UUID *uuid,
...
+
+	/*
+	 * Before updating the context list check again if a context
+	 * already exists in the list for a single instance TA, maybe
+	 * created recently enough so that it was not found at entry
+	 * in tee_ta_init_session().
+	 */
+	res = tee_ta_init_session_with_context(s, uuid);
+	if (res != TEE_ERROR_ITEM_NOT_FOUND) {
+		vm_info_final(&utc->uctx);
```
jenswi-linaro, line 498: I'd prefer checking if the context is available before allocating the `utc`. That means holding the mutex during the call to calloc(), but that should be OK.

core/kernel/user_ta.c
```diff
@@ -486,6 +486,22 @@ TEE_Result tee_ta_init_user_ta_session(const TEE_UUID *uuid,
 #endif
 
 	mutex_lock(&tee_ta_mutex);
+
+	/*
+	 * Before updating the context list check again if a context
```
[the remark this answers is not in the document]
etienne-lms: right, i missed that.

etienne-lms:
I've update the P-R to covert PTAs and StMM contexts creation.
Tested on vexpress-qemu_armv8a against `xtest -l 15`.

etienne-lms:
FYI the StMM part now builds but fails on target.

core/kernel/tee_ta_manager.c
```diff
@@ -539,23 +538,25 @@ static TEE_Result tee_ta_init_session_with_context(struct tee_ta_session *s,
...
 
-	/*
-	 * If TA isn't single instance it should be loaded as new
-	 * instance instead of doing anything with this instance.
-	 * So tell the caller that we didn't find the TA it the
-	 * caller will load a new instance.
-	 */
-	if ((ctx->flags & TA_FLAG_SINGLE_INSTANCE) == 0)
-		return TEE_ERROR_ITEM_NOT_FOUND;
+	if (is_user_ta_ctx(&ctx->ts_ctx)) {
```
jenswi-linaro, line 541: The flags `TA_FLAG_SINGLE_INSTANCE` and `TA_FLAG_MULTI_SESSION` apply to all TAs.
See for instance `PTA_MANDATORY_FLAGS`

core/kernel/user_ta.c
```diff
@@ -458,9 +458,23 @@ TEE_Result tee_ta_init_user_ta_session(const TEE_UUID *uuid,
 	TEE_Result res = TEE_SUCCESS;
 	struct user_ta_ctx *utc = NULL;
 
+	mutex_lock(&tee_ta_mutex);
+
+	/*
+	 * Before updating the context list check again if a context
+	 * already exists in the list for a single instance TA.
+	 */
+	res = tee_ta_init_session_with_context(s, uuid);
```
jenswi-linaro, line 467: I don't like calling `tee_ta_init_session_with_context()` twice. How about splitting `tee_ta_init_user_ta_session()` (and `stmm_init_session()`) into a two-step operation? The first function, for instance `tee_ta_init_user_ta_session()` does everything up to the first `mutex_unlock()`, but with the mutex held by the calling function. The second function, `tee_ta_complete_user_ta_session()` is called after `tee_ta_init_user_ta_session()` has returned OK without the mutex held, instead it takes the mutex internally as needed.

etienne-lms (commented):
Sorry I left this for a while. I'll be back on this change.

core/kernel/tee_ta_manager.c
jenswi-linaro: What is the point of checking if it's a user_ta first? Even if those two checks are always false for non-user_ta's, it is not harmful and those TAs must maintain those flags properly anyway.

etienne-lms: Right, indeed we could only test `ctx->flags`.

Regarding the debug trace message `"Re-open TA <uuid>"`, it was printed only for (user) TAs. I would be in favor of preserving this, to avaid confusion:
```c
if (is_user_ta_ctx(&ctx->ts_ctx))
	DMSG("Re-open TA %pUl", (void *)&ctx->ts_ctx.uuid);
```
or maybe change that trace message to make it more generic (TA + PTA + StMM). Maybe something like:
```c
DMSG("Re-open trusted service %pUl", (void *)&ctx->ts_ctx.uuid);
```

core/kernel/user_ta.c
etienne-lms: Note calling `tee_ta_init_session_with_context()` twice means that each time a session is opened on an already loaded single instance TA, we start fetching in the TA binary (early-ta or secstore-ta or nsec-fs-ta), then once done, we see that its already loaded, hence releasing the resources. It is worth it?

core/kernel/tee_ta_manager.c
jenswi-linaro: The latter is fine.

core/kernel/user_ta.c
jenswi-linaro: I was thinking more along https://github.com/jenswi-linaro/optee_os/tree/race_fix_example
It's not even compile-tested and I think the mutex lock/unlock sequences have room for improvement.

etienne-lms: Thanks for the proposal. I'll update this series straight replacing some of my commits with a commit based on your proposal.

core/kernel/user_ta.c
```diff
@@ -477,15 +477,18 @@ TEE_Result tee_ta_init_user_ta_session(const TEE_UUID *uuid,
...
+		return res;
+	}
+
 	utc->ta_ctx.is_initializing = true;
 
 #ifdef CFG_TA_PAUTH
 	crypto_rng_read(&utc->uctx.keys, sizeof(utc->uctx.keys));
 #endif
 
-	mutex_lock(&tee_ta_mutex);
```
jenswi-linaro, line 488: Would adding an `assert(mutex_is_locked(&tee_ta_mutex));` make sense here or at the beginning of this function?
We don't have a `mutex_is_locked()` function, but that's trivial to implement.

jforissier: `mutex_try_lock()` would be better because `mutex_is_locked()` cannot be reliable when it returns false.

jenswi-linaro: `mutex_is_locked()` would never be reliable, but it would make the intent clear and catch enough errors that we'd understand something is wrong.
`mutex_try_lock()` here would in the best case be deceiving since we don't know which thread is holding the lock.

jforissier: With `mutex_is_locked()` you would not know which thread is holding the lock either. Perhaps we need `mutex_is_locked_by_current_thread()`.

jenswi-linaro: That requires further bookkeeping. The point with the assert isn't a 100% catch rate. If it works while there's no further contention on the mutex that's good enough. Another option is to only leave a comment here explaining that we expect the mutex to be held, but that will quite likely have an even lower catch rate. ;-)

etienne-lms: If you think it's safer in the long term to keep the mutex lock/unlock calls near where it's needed (from inside each pta/uta/stmm context/session setup functions), we can discard this session init functions split and come back to the previous patch series proposal.

At the time,  your concern was that `tee_ta_init_session_with_context()` was called twice 
(https://github.com/OP-TEE/optee_os/pull/6836#discussion_r1615578427). I think we can remove the 1st call, from _tee_ta_manager.c_. _pseudo_ta.c_ would call it once, only if UUID matches a PTA UUID. _stmm_sp.c_ would call it only if  the UUID matches StMM interface one. _user_ta.c_ would call it once.

jenswi-linaro: No, I'm not concerned. A comment is fine as a reminder if there's a problem with an assert.

etienne-lms: I've appended a fixup commit to address this discussion thread. Please tell me if it matches your expectations.

core/kernel/pseudo_ta.c
```diff
@@ -294,6 +294,12 @@ TEE_Result tee_ta_init_pseudo_ta_session(const TEE_UUID *uuid,
 	struct tee_ta_ctx *ctx;
 	const struct pseudo_ta_head *ta;
 
+	/*
+	 * Caller is expected to hold tee_ta_mutex for safe changes
+	 * in @s and registering of the context in tee_ctxes list.
+	 */
+	assert(!mutex_trylock(&tee_ta_mutex));
```
jenswi-linaro, line 301: This is not very nice even if it gives a somewhat usable behaviour. I don't want to risk getting this pattern copied.
Please remove this assert if `mutex_is_locked()` isn't acceptable.

etienne-lms: Do you mean you'd rather have this line replaced by `assert(mutex_is_locked())`?

As `mutex_is_locked()` is likely not to be very bullet-proof but for weak debug purpose, i wonder if a debug helper `assert_mutex_is_locked()` would be more suitable.

jenswi-linaro: > Do you mean you'd rather have this line replaced by `assert(mutex_is_locked())`?

Yes, either that or removed.

> 
> As `mutex_is_locked()` is likely not to be very bullet-proof but for weak debug purpose, i wonder if a debug helper `assert_mutex_is_locked()` would be more suitable.

Nor is `mutex_trylock()`. If the mutex is already held by another thread it will do no better than `mutex_is_locked(). If the mutex is unlocked it will have a side effect before the assert triggers. It may also lead to confusion about whether there's a missing `mutex_unlock()` or if it's intended that `mutex_trylock()` only should be called when debug is enabled.

I only want a reminder that the mutex must be held here. I thought `assert(mutex_is_locked())` would be a nice and easy complement to that but it seems it was neither.

etienne-lms: series updated (and rebased for CI support)

jenswi-linaro (commented):

etienne-lms:
Thinking twice, should we rather add a `mutex_is_write_locked()` helper function instead of `mutex_is_locked()` in this specific case?

jenswi-linaro:
Look good to me.

etienne-lms:
Thanks. Fixup commit squashed.

PR #6848, opened 2024-05-21
core: fix possible overflow in shdr_alloc_and_copy()

jenswi-linaro:
Prior to this patch, if SHDR_GET_SIZE() overflows it will return 0 and further down in the function lead to an out-of-bounds access. So fix this with an explicit test before using shdr_size in shdr_alloc_and_copy().

Fixes: 064663e8bd27 ("core: crypto: add struct shdr helper functions")


jforissier (commented):

PR #6852, opened 2024-05-22
Rpmb probe v7 wip

jenswi-linaro:
This pull request is a work in progress, preparing for the next version of the corresponding kernel patches.
The purpose for now is to help coordinate the testing of these OP-TEE patches.
We should not merge this until we know the corresponding kernel patches will be accepted.


jforissier:
> We should not merge this until we know the corresponding kernel patches will be accepted.

Then let's make them "draft" maybe? ("Convert to draft" on the right hand side of the page).

lib/libutee/include/user_ta_header.h
```diff
@@ -24,16 +24,17 @@
...
+	 * world, first before the tee-supplicant has started and then once
+	 * more when the tee-supplicant is started. The flags below control
+	 * if the TA should be reported in the first or second or case.
 	 */
 #define TA_FLAG_DEVICE_ENUM		BIT32(9)  /* without tee-supplicant */
 #define TA_FLAG_DEVICE_ENUM_SUPP	BIT32(10) /* with tee-supplicant */
 	/* See also "gpd.ta.doesNotCloseHandleOnCorruptObject" */
 #define TA_FLAG_DONT_CLOSE_HANDLE_ON_CORRUPT_OBJECT \
 					BIT32(11)
+#define TA_FLAG_DEVICE_ENUM_RPMB	BIT32(12) /* with rpmb */
```
jforissier, line 37: How can a TA know this? Unless `TEE_STORAGE_PRIVATE_RPMB` is used (which is an OP-TEE extension), the TA does not know if the secure storage depends on the REE FS, RPMB, or both.
So in my opinion the flag should indicate the use of secure storage. For example `TA_FLAG_USE_SECURE_STORAGE`.  Then the TEE core can infer the RPMB dependency if that flag is set **and** RPMB is enabled.

jenswi-linaro: Makes sense, I'll update.

etienne-lms (commented):
Commit "core: rpmb: set device ID in tee_rpmb_init()" LGTM with a minor comment.

Commits "core: add THREAD_SHM_CACHE_USER_RPMB",
"core: rpmb: use thread_rpc_shm_cache_alloc()", 
"core: add flag to indicate dependency on RPMB" and
"ci: build with CFG_RPMB_WRITE_KEY=y" look good to me.

Few minor comments for commit "core: rpmb: probe for kernel RPMB driver".

FYI I've tested this series on by stm32mp1 (32bit) platform. Xtest reported no errors but I faced issues with StMM accessing the RPMB from U-Boot. Failure occurs when I apply commit "core: rpmb: probe for kernel RPMB driver". I'll try to investigate a bit further.

jenswi-linaro:
This is still work in progress.

core/pta/device.c
```diff
@@ -82,13 +83,21 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 				 uint32_t nCommandID, uint32_t nParamTypes,
 				 TEE_Param pParams[TEE_NUM_PARAMS])
 {
+	TEE_Result res = TEE_SUCCESS;
+
```
apalos, line 87: before that, a check needs to be added in add_ta().
That function already treats TA_FLAG_DEVICE_ENUM and TA_FLAG_DEVICE_ENUM_SUPP as mutually exclusive.
The new RPMB flag has to be checked as well

core/include/optee_rpc_cmd.h
```diff
@@ -180,6 +180,38 @@
...
+ *
+ * [out]    value[0].a	    OPTEE_RPC_SHM_TYPE_*, the parameter for
+ *			    OPTEE_RPC_CMD_SHM_ALLOC
+ */
+#define OPTEE_RPC_CMD_RPMB_PROBE_RESET	U(22)
+
+/*
+ * Probe next RPMB device
+ *
+ * [out]    value[0].a	    Must be OPTEE_RPC_RPMB_EMMC
```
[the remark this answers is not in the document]
jenswi-linaro: Good point. I'll update with something.

core/tee/tee_rpmb_fs.c
```diff
@@ -1105,18 +1101,16 @@ static TEE_Result tee_rpmb_init(uint16_t dev_id)
...
 
-	rpmb_ctx->dev_id = dev_id;
 
 	if (!rpmb_ctx->dev_info_synced) {
 		DMSG("RPMB: Syncing device information");
 
 		dev_info.rpmb_size_mult = 0;
 		dev_info.rel_wr_sec_c = 0;
-		res = tee_rpmb_get_dev_info(dev_id, &dev_info);
+		res = tee_rpmb_get_dev_info(rpmb_ctx->dev_id, &dev_info);
```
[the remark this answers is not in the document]
jenswi-linaro: OK

core/tee/tee_rpmb_fs.c
```diff
@@ -1046,57 +1105,56 @@ static TEE_Result tee_rpmb_write_and_verify_key(void)
...
+	rpmb_ctx->legacy_operation = true;
+	rpmb_ctx->dev_id = CFG_RPMB_FS_DEV_ID;
+	rpmb_ctx->shm_type = THREAD_SHM_TYPE_APPLICATION;
 
+	if (!rpmb_ctx->dev_info_synced) {
 		dev_info.rpmb_size_mult = 0;
 		dev_info.rel_wr_sec_c = 0;
 		res = tee_rpmb_get_dev_info(rpmb_ctx->dev_id, &dev_info);
 		if (res != TEE_SUCCESS)
 			goto func_exit;
```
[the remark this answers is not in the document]
jenswi-linaro: OK

jenswi-linaro:
Update. OK to squash and rebase on master?

jenswi-linaro:
Squashed and rebased on master.

jenswi-linaro:
There's a checkpatch warning for `IS_ENABLED(RPMB_DRIVER_MULTIPLE_WRITE_FIXED)`.
Should I revert to `#ifdef RPMB_DRIVER_MULTIPLE_WRITE_FIXED` to fix it?

jforissier:
For "core: rpmb: set device ID in tee_rpmb_init()":

jforissier:
For "core: add THREAD_SHM_CACHE_USER_RPMB":

core/tee/tee_rpmb_fs.c
```diff
@@ -267,6 +268,7 @@ struct tee_rpmb_ctx {
 	bool key_derived;
 	bool key_verified;
 	bool dev_info_synced;
+	enum thread_shm_type shm_type;
```
jforissier, line 275: This field is only ever set to `THREAD_SHM_TYPE_APPLICATION` in this commit. Do you expect other values can be used later?

jenswi-linaro: Yes, it's updated in "core: rpmb: probe for kernel RPMB driver" based on the result of OPTEE_RPC_CMD_RPMB_PROBE_RESET.

core/pta/device.c
```diff
@@ -23,8 +24,11 @@
 static void add_ta(uint32_t flags, const TEE_UUID *uuid, uint8_t *buf,
 		   uint32_t blen, uint32_t *pos, uint32_t rflags)
 {
-	if ((flags & TA_FLAG_DEVICE_ENUM) &&
-	    (flags & TA_FLAG_DEVICE_ENUM_SUPP)) {
+	flags &= (TA_FLAG_DEVICE_ENUM | TA_FLAG_DEVICE_ENUM_SUPP |
+		  TA_FLAG_DEVICE_ENUM_SEC_STORAGE);
+	if (flags && flags != TA_FLAG_DEVICE_ENUM &&
+	    flags != TA_FLAG_DEVICE_ENUM_SUPP &&
+	    flags != TA_FLAG_DEVICE_ENUM_SEC_STORAGE) {
```
jforissier, line 31: I think we should handle the "enumerable" bit and the dependencies separately. Then `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` would be best renamed to `TA_FLAG_DEPENDS_ON_SEC_STORAGE`.
In addition, "secure storage" is ambiguous since it can be implemented either with RPMB or the REE FS, which might end up requiring tee-supplicant or not. But the TA does not necessarily know about it (it knows in case it uses `_PRIVATE_RPMB` or `_PRIVATE_REE_FS` but not if it use `_PRIVATE`, in this case only the TEE core knows).

jforissier:
For "ci: build with CFG_RPMB_WRITE_KEY=y":

core/pta/device.c
jenswi-linaro: `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` is an abstraction that selects the required stage to be enumerated. It does not cover `TEE_STORAGE_PRIVATE_REE` or `TEE_STORAGE_PRIVATE_RPMB`, that would require yet another flag.

So the flag isn't about what the TA depends on, it's the stage where it should be enumerated.

jenswi-linaro: I'll update the commit message for "core: add flag to indicate dependency on RPMB" to align better with this.

jenswi-linaro:
Tags applied, update commit message for "core: add flag to enumerate TAs when secure storage is ready", replacing "core: add flag to indicate dependency on RPMB".

jforissier:
For "core: rpmb: use thread_rpc_shm_cache_alloc()":

core/tee/tee_rpmb_fs.c
```diff
@@ -1045,57 +1104,56 @@ static TEE_Result tee_rpmb_write_and_verify_key(void)
...
-		dev_info.rel_wr_sec_c = 0;
-		res = tee_rpmb_get_dev_info(&dev_info);
-		if (res != TEE_SUCCESS)
-			goto func_exit;
+	memcpy(rpmb_ctx->cid, dev_info->cid, RPMB_EMMC_CID_SIZE);
 
-		DMSG("RPMB: RPMB size is %d*128 KB", dev_info.rpmb_size_mult);
-		DMSG("RPMB: Reliable Write Sector Count is %d",
-		     dev_info.rel_wr_sec_c);
+	if (IS_ENABLED(RPMB_DRIVER_MULTIPLE_WRITE_FIXED))
```
jforissier, line 1129: Perhaps this conditional could be removed? I can see it dates back to commit b01047730e77 ("Open-source the TEE Core") so it may very well be irrelevant now.

jforissier (commented):
For commit "core: rpmb: probe for kernel RPMB driver": one question below, in any case:

core/pta/device.c
```diff
@@ -82,18 +86,31 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
...
-				   TA_FLAG_DEVICE_ENUM);
+		rflags = TA_FLAG_DEVICE_ENUM;
+		break;
 	case PTA_CMD_GET_DEVICES_SUPP:
-		return get_devices(nParamTypes, pParams,
-				   TA_FLAG_DEVICE_ENUM_SUPP);
-	default:
+		rflags = TA_FLAG_DEVICE_ENUM_SUPP;
+		if (IS_ENABLED(CFG_REE_FS))
+			rflags |= TA_FLAG_DEVICE_ENUM_SEC_STORAGE;
```
jforissier, line 99: - So here we have two flags set but `add_ta()` doesn't allow it and will return no TA
- What if `CFG_REE_FS` is disabled but `CFG_RPMB_FS` is enabled?

core/pta/device.c
```diff
@@ -82,18 +86,31 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
...
 		break;
+	case PTA_CMD_GET_DEVICES_RPMB:
+		if (!IS_ENABLED(CFG_REE_FS)) {
```
jforissier, line 102: `IS_ENABLED(CFG_RPMB_FS)`, no?

lib/libutee/include/user_ta_header.h
```diff
@@ -24,16 +24,39 @@
...
+	 *
+	 * The flags below control at which stage a TA will be enumerated:
+	 * TA_FLAG_DEVICE_ENUM - at stage 1
+	 * TA_FLAG_DEVICE_ENUM_SEC_STORAGE -
+	 *      when secure storage is available, at stage 2 or 3 depending
+	 *      on whether TEE_STORAGE_PRIVATE is using RPMB FS
+	 *      (CFG_REE_FS=n CFG_RPMB_FS=y) or REE FS (CFG_REE_FS=y). The
+	 *      former utilizes in kernel RPMB routing, and the latter
+	 *      depends on tee-supplicant to access secure storage.
+	 *      TA_FLAG_DEVICE_ENUM_SUPP - at stage 3
```
jforissier, line 41: Indentation

lib/libutee/include/user_ta_header.h
```diff
@@ -24,16 +24,39 @@
...
+	 *
+	 * TA_FLAG_DEVICE_ENUM_SEC_STORAGE enumerates at stage 2 or 3
+	 * depending on whether TEE_STORAGE_PRIVATE is using RPMB FS
+	 * (CFG_REE_FS=n CFG_RPMB_FS=y) or REE FS (CFG_REE_FS=y). The
+	 * former utilizes in kernel RPMB routing, and the latter depends
+	 * on tee-supplicant to access secure storage.
```
jforissier, line 47: Duplicate text

lib/libutee/include/user_ta_header.h
```diff
@@ -24,16 +24,39 @@
...
+	 *
+	 * The TA is enumerated at stage 2 if
+	 * TA_FLAG_DEVICE_ENUM_SEC_STORAGE is set and TEE_STORAGE_PRIVATE
+	 * is using RPMB FS, or if it's using REE FS it will be
+	 * enumerated at stage 3.
```
jforissier, line 52: Consider the case where both `CFG_REE_FS` and `CFG_RPMB_FS` are `y`. A TA with `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` which uses only `TEE_STORAGE_PRIVATE_RPMB` could in theory be enumerated at stage 2 but with the current code it will be enumerated at stage 3.
I have a feeling it can be a common case: enabling both FS to benefit from larger storage capacity, yet with specific TAs (such as fTPM) that should be enumerated before user space is available or need DoS protection and therefore would use pure RPMB storage via `TEE_STORAGE_PRIVATE_RPMB`. The TA flag needs to be more specific IMO.

jforissier (commented):
Comments on "core: add flag to enumerate TAs when secure storage is ready". I don't understand how it is supposed to be used.

core/tee/tee_rpmb_fs.c
jenswi-linaro: I'd rather remove that in a separate commit, if it should be removed at all. I don't see any harm in keeping it though, it explains what `rel_wr_sec_c` is supposed to be used for even if we don't trust it at the moment.

core/pta/device.c
jenswi-linaro: > * So here we have two flags set but `add_ta()` doesn't allow it and will return no TA

`rflags` what's compared with `flags`, where the latter is from the TA and may only have one of the `TA_FLAG_DEVICE_ENUM`, `TA_FLAG_DEVICE_ENUM_SUPP`, `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` flags set. `rflags` is permitted to have any bits set.

> * What if `CFG_REE_FS` is disabled but `CFG_RPMB_FS` is enabled?

Then will the `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` bit be set during `PTA_CMD_GET_DEVICES_RPMB` instead.

core/pta/device.c
jenswi-linaro: No, this is supposed to be the complement of `IS_ENABLED(CFG_REE_FS)` above.
RPMB only needs to be initialized if the `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` should be used here.

lib/libutee/include/user_ta_header.h
jenswi-linaro: `TA_FLAG_DEVICE_ENUM_SEC_STORAGE` only concerns `TEE_STORAGE_PRIVATE`, also the storage used by fTPM. We'd need another flag for `TEE_STORAGE_PRIVATE_RPMB`.

Would `TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE` be better?
I guess we could add a `TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE_RPMB` too, but that would be without any user at the moment.

core/pta/device.c
jforissier: OK

core/tee/tee_rpmb_fs.c
jforissier: ok

lib/libutee/include/user_ta_header.h
jforissier: > Would `TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE` be better?

Yes, I think so.

jenswi-linaro:
Update

jforissier:
For "core: add flag to enumerate TAs when secure storage is ready":

jenswi-linaro:
Tags applied

jenswi-linaro:
Dropped "Add TA_FLAG_DEVICE_ENUM_SEC_STORAGE for compat"

jforissier:
For "user_ta_header.h: update TA_FLAGS_MASK for RPMB support" (sorry I missed it):

jenswi-linaro:
Tag applied.

jenswi-linaro:
Added the commit "core: add CFG_RPMB_ANNOUNCE_PROBE_CAP"

jforissier:
For "core: add CFG_RPMB_ANNOUNCE_PROBE_CAP":

Thanks!

jenswi-linaro:
Tag applied.

jforissier:
It looks like this series is ready. @etienne-lms would you like to give ack/review tags?

etienne-lms (commented):
Should commit "user_ta_header.h: update TA_FLAGS_MASK for RPMB support" bit squashed in commit "core: add flag to enumerate TAs when secure storage is ready", for consistency?

That said,  `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commit
"core: rpmb: probe for kernel RPMB driver"
"core: rpmb: set device ID in tee_rpmb_init()",
"core: add THREAD_SHM_CACHE_USER_RPMB",
"core: rpmb: use thread_rpc_shm_cache_alloc()",
"core: add flag to enumerate TAs when secure storage is ready",
"ci: build with CFG_RPMB_WRITE_KEY=y",
"user_ta_header.h: update TA_FLAGS_MASK for RPMB support" and
"core: add CFG_RPMB_ANNOUNCE_PROBE_CAP".

jenswi-linaro:
Tags applied. I'd rather keep "user_ta_header.h: update TA_FLAGS_MASK for RPMB support" separate since that was contributed by Mikko.

jforissier:
Nice and quite helpful feature, thanks!

PR #6853, opened 2024-05-22
core: pager: fix arguments passed to calloc in alloc_merged_pgt_array()

GseoC:
It seems that the parameters are inverted. More of these are present in libtomcrypt.

Note: I haven't been able to test this patch yet as I don't have a board with me this week. Therefore, I didn't add the `Fixes:` yet.


jenswi-linaro:
This is mostly a style issue, the code is still working as intended but is worth fixing.

GseoC:
Updated with tag applied. I added the `Fixes` tag, let me know if that's okay.

jenswi-linaro:
Looks good.

PR #6861, opened 2024-05-28
core: compiler friendly 32-bit FTMN_FUNC_HASH() version

jenswi-linaro:
Add a more compiler friendly 32-bit implementation of FTMN_FUNC_HASH(). This gives the same output as the previous 64-bit based implementation, but base the computation on 32-bit values.


jenswi-linaro:
Tag applied

jenswi-linaro:
Commit message updated.

jenswi-linaro:
Fixing checkpatch issue.

PR #6863, opened 2024-05-29
ftrace: ftrace_format.py: display seconds

jforissier:
When the time spent in a function is 1 second or more, display it as seconds not milliseconds in order to keep the output nicely aligned.


jbech-linaro (commented):

jenswi-linaro:

PR #6865, opened 2024-05-30
drivers: sam: update PLLA, MCK and LPM for sama7g5 low-power support

TonyHan11:

### DIFF
--- a/core/drivers/pm/sam/pm_suspend.S
+++ b/core/drivers/pm/sam/pm_suspend.S
@@ -515,7 +515,95 @@ SUSPEND_END_FUNC(at91_backup_mode)
 #endif
 .endm
 
+#ifdef CFG_SAMA7G5
+/*
+ * at91_mckx_ps_enable:	save MCK1..4 settings and switch it to main clock
+ *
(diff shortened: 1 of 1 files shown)


core/drivers/pm/sam/pm_suspend.S
```diff
@@ -25,10 +25,20 @@ tmp3	.req	r6
...
+ * @r_mckid:	register holding master clock identifier
+ *
+ * Side effects: overwrites tmp1, tmp3
  */
-.macro wait_mckrdy
+.macro wait_mckrdy r_mckid
+#ifdef CFG_SAMA7G5
+	cmp	\r_mckid, #0
+	beq	2f
+	mov	tmp3, #AT91_PMC_MCKXRDY
```
jenswi-linaro, line 36: How about using a define for the clock identifier instead of runtime tests that always will result in the same?

TonyHan11: Updated, thank you.

jenswi-linaro:

TonyHan11:
Rebased on the latest `master` branch with review tags added.

PR #6867, opened 2024-05-30
core: ltc: Fix building with mbedtls

jkroon81:
Fix building OP-TEE with:

make PLATFORM=vexpress \
     PLATFORM_FLAVOR=juno \
     CFG_CRYPTOLIB_NAME=mbedtls \
     CFG_CRYPTOLIB_DIR=lib/libmbedtls
...
core/lib/libtomcrypt/aes_accel.c: In function ‘aes_ctr_encrypt_nblocks’: core/lib/libtomcrypt/aes_accel.c:182:21: error: ‘CTR_COUNTER_LITTLE_ENDIAN’ undeclared (first use in this function)
  182 |         if (mode == CTR_COUNTER_LITTLE_ENDIAN) {

### DIFF
--- a/core/lib/libtomcrypt/aes_accel.c
+++ b/core/lib/libtomcrypt/aes_accel.c
@@ -170,6 +170,7 @@ static int aes_cbc_decrypt_nblocks(const unsigned char *ct, unsigned char *pt,
 	return CRYPT_OK;
 }
 
+#ifdef LTC_CTR_MODE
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

We should probably do the same for the other modes, but that's not in the scope of this PR.

jforissier:
> We should probably do the same for the other modes, but that's not in the scope of this PR.

+1

PR #6868, opened 2024-05-30
ci: xen: test FF-A mediator with SPMC_AT_EL=1

jenswi-linaro:
Test Xen FF-A mediator with SPMC_AT_EL=1. This test replaces the previous test of the Xen OP-TEE mediator.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -384,6 +384,41 @@ jobs:
 
           make -j$(nproc) check XEN_BOOT=y
 
+  QEMUv8_Xen_ffa_check:
+    name: make check (QEMUv8, Xen FF-A)
+    runs-on: ubuntu-latest
+    container: jforissier/optee_os_ci:qemu_check
+    steps:
+      - name: Remove /__t/*
+        run: rm -rf /__t/*
+      - name: Restore build cache
+        uses: actions/cache@v3
+        with:
+          path: /github/home/.cache/ccache
+          key: qemuv8_xen_ffa_check-cache-${{ github.sha }}
+          restore-keys: |
+            qemuv8_xen_ffa_check-cache-
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Update Git config
+        run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
+      - shell: bash
+        run: |
+          # make check task
+          set -e -v
+          export LC_ALL=C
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -382,7 +382,7 @@ jobs:
           ln -s ${OPTEE_OS_TO_TEST} ${TOP}/optee_os
           cd ${TOP}/build
 
-          make -j$(nproc) check XEN_BOOT=y
+          make -j$(nproc) check XEN_BOOT=y SPMC_AT_EL=1
```
jforissier, line 385: This makes FF-A mandatory, right? Don't we want to still support (and test) the "legacy" mediator?

jenswi-linaro: Yes, it's mandatory for this test, just as it is for the Hafnium test.
I don't know how relevant the legacy mediator is.

jforissier: @lorc ?

lorc: Well, we are still using legacy mediator in (semi-)production. So for us, at least, it is still relevant.

Frankly, I didn't followed OP-TEE development closely. Are you planning to ditch old SMC API and switch to FF-A only?

jenswi-linaro: The old SMC API is still in use and supported, but FF-A is getting more interesting as virtualization goes. We don't plan to remove anything. We could continue supporting the legacy mediator in the CI loop if we can afford it.

lorc: > We could continue supporting the legacy mediator in the CI loop if we can afford it.

IMO, it would be ideal. But if it really stresses out CI, just test the FF-A case. I believe it is being developed more actively than the legacy API, so it is more important to test it.

jforissier: We don't reach any CI limit at the moment so we could make that a separate job and we would not notice any slowdown (assuming both take roughly the same amount of time).

jenswi-linaro: OK, let's keep the old test.

jenswi-linaro:
Are we happy with this?

lorc:
Well, no objections from my side.

`Acked-by: Volodymyr Babchuk <volodymyr_babchuk@epam.com>`

jforissier (approved):

jenswi-linaro:
Thanks, tags applied.

PR #6879, opened 2024-06-07
core: kernel: change DT getters to use manifest DT

ysbnim:
This change get_dt() and get_secure_dt() to use manifest DT as an alternative if there is no embedded DT or external DT.

I was wondering if this feature has to be provided by a CONFIG or as default.
Please give me any opinions about this.

Thank you.


jbech-linaro:
Looks OK to me.

core/kernel/dt.c
```diff
@@ -449,6 +450,11 @@ void *get_dt(void)
 	if (!fdt)
 		fdt = get_external_dt();
 
+#if defined(CFG_CORE_FFA)
```
jforissier, line 453: 1. Drop the `#if`, is it not needed.
2. Shouldn't `get_manifest_dt()` be declared in `include/kernel/dt.h` and implemented here? TBD in a separate commit prior to this one.

ysbnim: 1. Dropped `#if`
2. Added another commit to move manifest DT functions to kernel/dt.c

jforissier (commented):
Thanks for the update. LGTM:

core/include/kernel/dt.h
```diff
@@ -237,7 +237,9 @@ int fdt_get_reg_props_by_index(const void *fdt, int node, int index,
 int fdt_get_reg_props_by_name(const void *fdt, int node, const char *name,
 			      paddr_t *base, size_t *size);
 
-/* Returns embedded DTB if present, then external DTB if found, then NULL */
+/* Returns embedded DTB if present, then external DTB if found,
```
etienne-lms, line 240:
```patch
-/* Returns embedded DTB if present, then external DTB if found,
+/*
+ * Returns embedded DTB if present, then external DTB if found,

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with inline comment format fixed.

core/include/kernel/dt.h
ysbnim: Done

PR #6880, opened 2024-06-07
checkconf.mk: do not use full path to generate guard symbol in conf.h

Villemoes:
The combination of building with -g3 (which emits definitions of all defined preprocessor macros to the debug info) and using a full path to define the name of this preprocessor guard means that the output is not binary reproducible across different build hosts. For example, in my Yocto build, the string

  __home_ravi_yocto_tmp_glibc_work_stm32mp135fdk_oe_linux_gnueabi_optee_os_stm32mp_3_19_0_stm32mp_r1_1_build_stm32mp135f_dk_include_generated_conf_h_

appears in several build artifacts. Another developer or buildbot would not build in some /home/ravi/... directory.

In order to increase binary reproducibility, only use the path sans the $(out-dir)/ prefix of the conf.h file.

### DIFF
--- a/mk/checkconf.mk
+++ b/mk/checkconf.mk
@@ -17,7 +17,8 @@ define check-conf-h
 	cnf='$(strip $(foreach var,				\
 		$(call cfg-vars-by-prefix,$1),			\
 		$(call cfg-make-define,$(var))))';		\
-	guard="_`echo $@ | tr -- -/.+ _`_";			\
+	guardpath="$(patsubst $(out-dir)/%,%,$@)"		\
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

jforissier:
@Villemoes the CI error can be ignored, as you said it makes sense to keep the long line. Please apply Jens' Reviewed-by and I will merge this. Thank you!

jforissier:
It seems that trustedfirmware.org is unreliable (CI shows Git clone errors). I'll restart the jobs again and consider using the GitHub mirror instead.

PR #6885, opened 2024-06-11
plat-sam: add OPP support, assign clock rate for sama7g5 qspi0

TonyHan11:


core/arch/arm/plat-sam/freq.c
```diff
@@ -12,8 +12,102 @@
...
+		panic("Invalid opp-table");
+
+	fdt_for_each_subnode(opp_table, fdt, offset) {
+		if (dt_get_opp_hz(fdt, opp_table, &rate) != TEE_SUCCESS)
+			panic("Get opp-hz failed");
+
+		opp_rates->rates[opp_rates->rate_num++] = rate;
+	}
+	opp_rates->clk = clk;
+	((struct clk_ops *)clk->ops)->get_rates_array = get_rates_array;
```
jenswi-linaro, line 106: This doesn't look right. Why isn't `clk->ops` assigned correctly from the start?

TonyHan11: Yes, it's better that `clk->ops` be assigned in `static const struct clk_ops sam9x60_div_pll_ops_chg` in [clk-sam9x60-pll.c](https://github.com/OP-TEE/optee_os/blob/master/core/drivers/clk/sam/clk-sam9x60-pll.c#L389) in this case. But the CPU clock in use is defined in [the device tree](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/dts/sama7g5.dtsi#L33), it might be another clock so the assignment needs to be changed accordingly.

Due to the frequency of CPU is handled in this file I added the code of getting rates for OPP in this file too instead of into a file for clock. Do you have any suggestions? Thank you.

PS: As `ops` is defined as [`const struct clk_ops *`](https://github.com/OP-TEE/optee_os/blob/master/core/include/drivers/clk.h#L39), `(struct clk_ops *)` is added here for converting the type to avoid the compile error.

jenswi-linaro: @etienne-lms can you help us, please?

core/arch/arm/plat-sam/freq.c
```diff
@@ -12,8 +12,102 @@
...
+	if (!prop)
+		return TEE_SUCCESS;
+
+	opp_rates = calloc(1, sizeof(*opp_rates));
+	if (!opp_rates) {
+		EMSG("Fail to alloc opp_rates");
+
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+	memset(opp_rates, 0, sizeof(*opp_rates));
```
GseoC, line 93: You can remove this, you use calloc that already zeroes the memory.

core/arch/arm/plat-sam/freq.c
```diff
@@ -12,8 +12,102 @@
...
+static TEE_Result opp_rates_setup(const void *fdt, int node, struct clk *clk)
+{
+	const char *compatible = "operating-points-v2";
+	const struct fdt_property *prop = NULL;
+	unsigned long rate = 0;
+	int opp_table = 0;
+	int offset = 0;
+	int len = 0;
+
+	prop = fdt_get_property(fdt, node, compatible, &len);
```
GseoC, line 83: you don't use len, you can remove it

core/arch/arm/plat-sam/freq.c
```diff
@@ -12,8 +12,102 @@
...
+	opp_rates = calloc(1, sizeof(*opp_rates));
+	if (!opp_rates) {
+		EMSG("Fail to alloc opp_rates");
+
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+	memset(opp_rates, 0, sizeof(*opp_rates));
+
+	offset = (int)prop - (int)fdt;
+	if (fdt_node_check_compatible(fdt, offset, compatible))
```
GseoC, line 96: I don't really get the above operation, isn't the node you're searching in simply "node"?
Why is `if (fdt_node_check_compatible(fdt, node, compatible))` not enough?

core/arch/arm/plat-sam/freq.c
GseoC: Then maybe it should be a dedicated clock with your pll as its parent?
What do you think @etienne-lms

core/arch/arm/plat-sam/freq.c
TonyHan11: Here the node in search is [opp-table](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/dts/sama7g5.dtsi#L40) and the compatible value is `operating-points-v2`. The `node` parameter of the function is for the [`cpu node`](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/dts/sama7g5.dtsi#L29) in which `cpu_opp_table` is one of its items.

core/arch/arm/plat-sam/freq.c
TonyHan11: Removed the definition for `len` and replaced `&len` with `NULL`.

core/arch/arm/plat-sam/freq.c
TonyHan11: `memset(opp_rates, 0, sizeof(*opp_rates));` removed, thanks.

core/arch/arm/plat-sam/freq.c
etienne-lms: prefer
```patch
- 	offset = (int)prop - (int)fdt;
+	offset = fdt_node_offset_by_phandle(fdt, fdt32_to_cpu(*cuint));
+	if (offset < 0) {
+		// error case
+	}
+
 	if (fdt_node_check_compatible(fdt, offset, compatible)) {
 		panic("Invalid opp-table");

File: core/arch/arm/plat-sam/freq.c (hunk 1 above)
Comment (etienne-lms) [reply core/arch/arm/plat-sam/freq.c:104]: I agree with @GseoC, it would be better this opp-v2 node is a clock provider that registers a clock with parent clock being `clk` and with a `.set_rate` handler that ensures the requested clock rate matches the supported ones. In that scope, your SCMI server should expose this cpu_opp clock.

But... the clk framework will not allow that since it holds a mutex preventing a clock handle to call clk API functions as `clk_set_rate()`.
If this piece of code can ensure this added "cpu_opp" clock has no parents, its `.set_rate` handler could call the `parent->ops->set_rate` handler straight once it has checked the request clock is valid.
Something like:
```c
static TEE_Result cpu_opp_clk_set_rate(struct clk *clk, unsigned long rate, unsigned long parent_rate __unused)
{
	// check @rate is valid

	assert(clk->parent && !clk->parent->parent);
	return clk->parent->ops->set_rate(clk->parent, rate, 0);
}

static struct clk_ops cpu_opp_clk_ops = {
	.set_rate = cpu_opp_clk_set_rate,
	.get_rates_array = get_rates_array,
};

(...)

static TEE_Result get_freq_from_dt(void)
{
	struct clk *cpu_opp_clk = NULL;
	(...)

	cpu_opp_clk = clk_alloc("cpu-opp", &cpu_opp_clk_ops, &clk, 1);
	// check error case

	if (opp_rates_setup(fdt, node, clk)) {
		// erro case
	}

	res = clk_register(cpu_opp_clk);
	(...)

	/* CPU clock is likely always enabled so set its refcount */
	if (clk_enable(cpu_opp_clk))
		panic();
}

File: core/arch/arm/plat-sam/freq.c [hunk 5]
```diff
@@ -12,8 +12,100 @@
...
+		panic("Invalid opp-table");
+
+	fdt_for_each_subnode(opp_table, fdt, offset) {
+		if (dt_get_opp_hz(fdt, opp_table, &rate) != TEE_SUCCESS)
+			panic("Get opp-hz failed");
+
+		opp_rates->rates[opp_rates->rate_num++] = rate;
```
etienne-lms, line 101: prefer in 2 step an let compile optimize the whole (and check that index does not overflows).
Suggestion:
```c
	size_t rate_num = 0;

	(...)

	fdt_for_each_subnode(opp_table, fdt, offset) {
		if (rate_num >= OPP_RATES_MAX)
			panic("CPU OPPs overflow");

		if (dt_get_opp_hz(fdt, opp_table, opp_rates->rates + rate_num))
			panic("Get opp-hz failed");

		rate_num++;
	}
	opp_rates->rate_num = rate_num;

File: core/arch/arm/plat-sam/freq.c (hunk 4 above)
Comment (TonyHan11) [reply core/arch/arm/plat-sam/freq.c]: updated with the code at [`core/drivers/clk/sam/at91_cpu_opp.c:96`](https://github.com/OP-TEE/optee_os/pull/6885/files#diff-501b0efc0f38b2e470cc29c13962cd636701f80226260de47265cf185e6b8be3R96), thanks.

File: core/arch/arm/plat-sam/freq.c (hunk 5 above)
Comment (TonyHan11) [reply core/arch/arm/plat-sam/freq.c]: Updated, thanks.

File: core/arch/arm/plat-sam/freq.c (hunk 1 above)
Comment (TonyHan11) [reply core/arch/arm/plat-sam/freq.c]: @etienne-lms Thank you for your suggestions.
Now a new clock for CPU OPP is created if OPP is available and the clock is registered as `AT91_SCMI_CLK_CPU_OPP` in SCMI clock list. Most of the code is moved to a new file `core/drivers/clk/sam/at91_cpu_opp.c`.

File: core/drivers/clk/sam/at91_cpu_opp.c [hunk 6]
```diff
@@ -18,6 +18,44 @@ struct clk_rates {
...
+	       *nb_elts * sizeof(*rates));
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result cpu_opp_clk_set_rate(struct clk *clk, unsigned long rate,
+				       unsigned long parent_rate)
+{
+	assert(clk->parent);
+
```
etienne-lms, line 50: I would recommend here to check target `parent_rate` is lists in `opp_rates->rates[]`.

core/drivers/clk/sam/at91_cpu_opp.c
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ *  Copyright (C) 2024 Microchip Technology Inc.
+ */
+
+#include <at91_clk.h>
+#include <kernel/misc.h>
+#include <kernel/panic.h>
+#include <libfdt.h>
```
etienne-lms, line 9:
```c
#include <malloc.h>
#include <trace.h>

**Review by etienne-lms — COMMENTED** [review]
The overall looks ok. Some comments.

File: core/drivers/clk/sam/at91_cpu_opp.c (hunk 6 above)
Comment (TonyHan11) [reply core/drivers/clk/sam/at91_cpu_opp.c:51]: Updated according to the comments except this one, do you mean to check `parent_rate` with the following patch? Thanks.
```
--- a/core/drivers/clk/sam/at91_cpu_opp.c
+++ b/core/drivers/clk/sam/at91_cpu_opp.c
@@ -14,6 +14,7 @@

 /* the clock rates for CPU OPP */
 struct clk_rates {
+       unsigned long parent_rate;
        size_t rate_num; /* the number of valid clock rates in @rates */
        unsigned long rates[OPP_RATES_MAX];
 };
@@ -48,6 +49,7 @@ static TEE_Result cpu_opp_clk_set_rate(struct clk *clk, unsigned long rate,
                                       unsigned long parent_rate)
 {
        assert(clk->parent);
+       assert(opp_rates->parent_rate == parent_rate);

        return clk->parent->ops->set_rate(clk->parent, rate, parent_rate);
 }
@@ -132,6 +134,7 @@ TEE_Result at91_clk_register_cpu_opp(const void *fdt, int node, struct clk *clk)
        if (res)
                return res;

+       opp_rates->parent_rate = clk_get_rate(clk);
        cpu_opp_clk = clk_alloc("cpu-opp", &cpu_opp_clk_ops, &clk, 1);
        if (!cpu_opp_clk)
                panic("CPU OPP clock alloc failed");
```

etienne-lms: No, I was rather thinking of something like:
```c
for (n = 0; n < opp_rates->rate_num; n++)
    if (rate == opp_rates->rates[n])
        break;
if (n == opp_rates->rate_num)

TonyHan11: The `rate` check added, thank you.

etienne-lms (commented):
For commit message of "drivers: clk: sam: add the implement of CPU OPP clock": s/ti's/its/.
With this comment addressed: 

TonyHan11:
Updated with commit message and `Acked-by`

PR #6889, opened 2024-06-13
ci: set FORCE_UNSAFE_CONFIGURE=1

jforissier:
The newer version of Buildroot recently selected in the OP-TEE manifest [1] does not allow building as root by default (apparently this is not a Buildroot policy but rather defined in some packages included in Buildroot). While this makes sense in general, we could not care less in a CI container. Therefore set the appropriate environment variable to bypass this check. We could as well change the Docker image to build as a non-root user but let's just pick the easiest way for now.

Link: https://github.com/OP-TEE/manifest/pull/281 [1]


jenswi-linaro (commented):

PR #6897, opened 2024-06-18
scripts: fix invalid escape sequence

### LINKED ISSUE #6896: SyntaxWarning from gen_compile_commands.py
I see there's `noqa: W605` to suppress the complaint "Invalid escape sequence" from linter, but this is becoming a SyntaxWarning from Python 3.12 [(ref)](https://github.com/python/cpython/issues/98401).

https://github.com/OP-TEE/optee_os/blob/956c2d50d60109a6053e14da6ff97d6243ea5d65/scripts/gen_compile_commands.py#L176-L180

I'd propose we properly escape the back slash by changing it to 

``` python
command_prefix.replace(r'\#', '#')
```

LuminousXLB:
A backslash-character pair that is not a valid escape sequence is generating a SyntaxWarning in Python 3.12 and could generate a SyntaxError in a future version.

Use a raw string to avoid the escape.

fix #6896


jbech-linaro (commented):

LuminousXLB:
Tag appended.

PR #6898, opened 2024-06-18
core: REE_FS: fix dirfile handle refcount

jenswi-linaro:
The counter ree_fs_dirh_refcount is used to determine when ree_fs_dirh should be free, not as a guarantee that ree_fs_dirh is still valid. This wasn't the assumption in ree_fs_readdir_rpc(), ree_fs_closedir_rpc(), and ree_fs_opendir_rpc(). So fix that by using get_dirh() in ree_fs_readdir_rpc as needed.

Fixes: ace6039fd434 ("core: REE_FS: refcount dirfile handle")

### DIFF
--- a/core/tee/tee_ree_fs.c
+++ b/core/tee/tee_ree_fs.c
@@ -510,6 +510,14 @@ static const struct tee_fs_dirfile_operations ree_dirf_ops = {
 	.commit_writes = ree_dirf_commit_writes,
 };
 
+/*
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Added a comment for `ree_fs_dirh` and `ree_fs_dirh_refcount`.

jenswi-linaro:
Adding @ZheTing815's Reported-by.

jforissier (commented):

jenswi-linaro:
Comment addressed and tag applied.

PR #6899, opened 2024-06-18
plat-mediatek: add support to extend MAX_XLAT_TABLE

ZheTing815:
We will create the memory pool from MEM_AREA_RES_VASPACE.
When using the reserved virtual memory space, it is necessary to increase MAX_XLAT_TABLE based on its size.

### DIFF
--- a/core/arch/arm/plat-mediatek/conf.mk
+++ b/core/arch/arm/plat-mediatek/conf.mk
@@ -14,6 +14,10 @@ CFG_DRAM_BASE ?= 0x40000000
 # default DRAM size 1 GiB
 CFG_DRAM_SIZE ?= 0x40000000
 
+# When need to create a virtual memory pool for mapping other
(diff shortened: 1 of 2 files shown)


core/arch/arm/plat-mediatek/conf.mk
```diff
@@ -14,6 +14,8 @@ CFG_DRAM_BASE ?= 0x40000000
 # default DRAM size 1 GiB
 CFG_DRAM_SIZE ?= 0x40000000
 
+CFG_MTK_RESERVED_VA ?= n
```
jenswi-linaro, line 17: Please add a comment describing this new option.

ZheTing815:
add some descriptions for the config

jenswi-linaro (commented):

ZheTing815:
tag applied

PR #6907, opened 2024-06-21
drivers: gic: Dynamically assign interrupts to non-secure world

ChenRunyang:
Add gic_spi_release_to_ns() API function in GIC driver to release an interrupt to Non secure settings. This functionality is essential for scenarios where a specific interrupt needs to be dynamically set to either Group 1 Secure (G1S) or Group 1 Non-Secure (G1NS) at different times.

### DIFF
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -17,6 +17,7 @@
 #include <kernel/dt_driver.h>
 #include <kernel/interrupt.h>
 #include <kernel/misc.h>
+#include <kernel/mutex.h>
 #include <kernel/panic.h>
 #include <libfdt.h>
 #include <mm/core_memprot.h>
@@ -138,6 +139,7 @@ struct gic_data {
 
 static bool gic_primary_done __nex_bss;
 static struct gic_data gic_data __nex_bss;
+static struct mutex gic_mutex = MUTEX_INITIALIZER;
 
 static void gic_op_add(struct itr_chip *chip, size_t it, uint32_t type,
 		       uint32_t prio);
@@ -856,6 +858,36 @@ void gic_dump_state(void)
 			     gic_it_get_group(gd, i), gic_it_get_target(gd, i));
 		}
 	}
+}
+
+TEE_Result gic_spi_release_to_ns(size_t it)
+{
+	struct gic_data *gd = &gic_data;
+	size_t idx = it / NUM_INTS_PER_REG;
+	uint32_t mask = BIT32(it % NUM_INTS_PER_REG);
+
+	if (it >= gd->max_it || it < GIC_SPI_BASE)
+		return TEE_ERROR_BAD_PARAMETERS;
+	/* Make sure it's already disabled */
+	if (!gic_it_is_enabled(gd, it))
+		return TEE_ERROR_BAD_STATE;
+	/* Assert it's secure to start with */
+	if (!gic_it_get_group(gd, it))
+		return TEE_ERROR_BAD_STATE;
+
+	mutex_lock(&gic_mutex);
+	gic_it_set_cpu_mask(gd, it, 0);
+	gic_it_set_prio(gd, it, GIC_SPI_PRI_NS_EL1);
+
+	/* Clear pending status */
+	io_write32(gd->gicd_base + GICD_ICPENDR(idx), mask);
+	/* Assign it to NS Group1 */
+	io_setbits32(gd->gicd_base + GICD_IGROUPR(idx), mask);
+#if defined(CFG_ARM_GICV3)
+	io_clrbits32(gd->gicd_base + GICD_IGROUPMODR(idx), mask);
+#endif
+	mutex_unlock(&gic_mutex);
+	return TEE_SUCCESS;
 }
 
 static void __maybe_unused gic_native_itr_handler(void)
--- a/core/include/drivers/gic.h
+++ b/core/include/drivers/gic.h
@@ -31,6 +31,8 @@
 #define GIC_SGI_SEC_BASE	8
 /* Max ID for secure SGIs */
 #define GIC_SGI_SEC_MAX		15
+/* Default IRQ priority for SPIs in Non-Sec EL1 */
+#define GIC_SPI_PRI_NS_EL1	0x50
 
 /*
  * The two gic_init() and gic_init_v3() functions initializes the struct
@@ -56,4 +58,11 @@ void gic_init_per_cpu(void);
 
 /* Print GIC state to console */
 void gic_dump_state(void);
+
+/*
+ * Reassign one of the SPIs to normal world and set its priority to
+ * GIC_SPI_PRI_NS_EL1. Ensure that the interrupt is currently
+ * assigned to secure world and disabled when this function is called.
+ */
+TEE_Result gic_spi_release_to_ns(size_t it);
 #endif /*__DRIVERS_GIC_H*/


core/include/kernel/interrupt.h
```diff
@@ -299,6 +300,16 @@ static inline void interrupt_raise_sgi(struct itr_chip *chip, size_t itr_num,
 	chip->ops->raise_sgi(chip, itr_num, cpu_mask);
 }
 
+/*
+ * interrupt_reset() - Reset the property for an interrupt
+ * @chip	Interrupt controller
+ * @itr_num	Interrupt number to reset
+ */
+static inline void interrupt_reset(struct itr_chip *chip, size_t itr_num)
+{
```
etienne-lms, line 309: `assert(chip->ops->reset)`

core/drivers/gic.c
```diff
@@ -996,6 +1024,26 @@ static void gic_op_raise_sgi(struct itr_chip *chip, size_t it,
 	gic_it_raise_sgi(gd, it, cpu_mask, ns);
 }
 
+static void gic_it_reset(struct gic_data *gd, size_t it)
+{
+	size_t idx = it / NUM_INTS_PER_REG;
+	uint32_t mask = 1 << (it % NUM_INTS_PER_REG);
+
```
etienne-lms, line 1031: must check that `it >= GIC_SPI_BASE`

For consistency, it think this function should also ensures the the interrupt was previously disabled. I think this means `git_it_reset()` needs a return code.

core/drivers/gic.c
```diff
@@ -996,6 +1024,26 @@ static void gic_op_raise_sgi(struct itr_chip *chip, size_t it,
 	gic_it_raise_sgi(gd, it, cpu_mask, ns);
 }
 
+static void gic_it_reset(struct gic_data *gd, size_t it)
```
etienne-lms, line 1027: this whole function `gic_it_reset()` seems supported only when `CFG_ARM_GICV3` is enabled.

core/include/kernel/interrupt.h
```diff
@@ -299,6 +300,16 @@ static inline void interrupt_raise_sgi(struct itr_chip *chip, size_t itr_num,
 	chip->ops->raise_sgi(chip, itr_num, cpu_mask);
 }
 
+/*
+ * interrupt_reset() - Reset the property for an interrupt
+ * @chip	Interrupt controller
+ * @itr_num	Interrupt number to reset
```
etienne-lms, line 306: This is supported only for SPIs, could you add this information in this inline description comment?

etienne-lms (changes requested):
I've not fully dug into the implication of this feature. That said, here are some comments. Also, I wonder it `interrupt_release()` would be a better label, rather than `interrupt_reset()`.

core/include/kernel/interrupt.h
ChenRunyang: Done

core/drivers/gic.c
ChenRunyang: Checked, if it < GIC_SPI_BASE, it will panic

ChenRunyang: The API only reset the properties of interrupt, It is not necessary that interrupts have been disabled before. Functionally, the two are not coupled.

ChenRunyang:
> I've not fully dug into the implication of this feature. That said, here are some comments. Also, I wonder it `interrupt_release()` would be a better label, rather than `interrupt_reset()`.

Replaced the label from interrupt_reset to interrupt_release.

core/include/kernel/interrupt.h
ChenRunyang: Done

core/drivers/gic.c
ChenRunyang: It also works when CFG_ARM_GICV3 is not enabled, The GIC version affects whether IGROUPMODR is configured.

core/drivers/gic.c
```diff
@@ -530,13 +535,27 @@ static void gic_init_base_addr(paddr_t gicc_base_pa, paddr_t gicd_base_pa,
 		gd->chip.dt_get_irq = gic_dt_get_irq;
 }
 
+static void gic_init_pre_prop(void)
+{
+	struct gic_data *gd = &gic_data;
+
+	gd->pre_prio = calloc(gd->max_it + 1, sizeof(*gd->pre_prio));
```
jenswi-linaro, line 542: This will only be needed in very specific configurations, but like this, it will impact all configurations. How about keeping this information outside the GIC driver and letting `interrupt_release()` take the needed parameters to restore the state?

core/drivers/gic.c
```diff
@@ -786,6 +805,26 @@ static void gic_it_raise_sgi(struct gic_data *gd __maybe_unused, size_t it,
 #endif
 }
 
+static void gic_it_release(struct gic_data *gd, size_t it)
+{
+	size_t idx = it / NUM_INTS_PER_REG;
+	uint32_t mask = 1 << (it % NUM_INTS_PER_REG);
+
+	/* Assigned to group0 */
```
jenswi-linaro, line 813: What about group1 interrupts?

jenswi-linaro (commented):
I wonder if this doesn't boil to changing the group assignment of the interrupt id. As long as the interrupt is disabled we could leave it to the previous owner to restore the other properties.

core/drivers/gic.c
ChenRunyang: The API is used for change group0 interrupts to group1 interrupts, so if group1 interrupts call this API, it will be asserted

core/drivers/gic.c
ChenRunyang: Add the gic_spi_get_prop API to get the interrupt properties, and the interrupt_release need to specify the interrupt attributes as parameters.

ChenRunyang:
> I wonder if this doesn't boil to changing the group assignment of the interrupt id. As long as the interrupt is disabled we could leave it to the previous owner to restore the other properties.

Yes, it will change the group assignment of the interrupt id. 
Updated the gic_op_release API to accept additional properties as parameters. These settings will be restored to their previous state by the original owner.

core/drivers/gic.c
jenswi-linaro: I'm not sure it's that simple (I'm happy to be proven wrong though). What if the interrupt is active?

core/include/kernel/interrupt.h
```diff
@@ -72,6 +72,8 @@ struct itr_ops {
 	void (*raise_pi)(struct itr_chip *chip, size_t it);
 	void (*raise_sgi)(struct itr_chip *chip, size_t it,
 			  uint32_t cpu_mask);
+	void (*release)(struct itr_chip *chip, size_t it,
```
anilreddy3, line 75: release() is doing other way of add(). so, would it be much clear if they are next to each other?

core/drivers/gic.c
```diff
@@ -786,6 +789,27 @@ static void gic_it_raise_sgi(struct gic_data *gd __maybe_unused, size_t it,
...
+static void gic_it_release(struct gic_data *gd, size_t it, uint8_t prio,
+			   uint8_t cpu_mask)
+{
+	size_t idx = it / NUM_INTS_PER_REG;
+	uint32_t mask = 1 << (it % NUM_INTS_PER_REG);
+
+	/* Assigned to group0 */
+	assert(!(io_read32(gd->gicd_base + GICD_IGROUPR(idx)) & mask));
+
+	gic_it_set_cpu_mask(gd, it, cpu_mask);
```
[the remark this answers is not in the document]
ChenRunyang: It's OK to clear the cpu_mask, but is it possible to set the priority to the Linux kernel's default priority setting of 0xa0?And the current behavior of Kernel setting prio is undefined.

core/drivers/gic.c
```diff
@@ -858,6 +888,20 @@ void gic_dump_state(void)
 	}
 }
 
+void gic_spi_get_prop(size_t it, uint8_t *cpu_mask, uint8_t *prio)
```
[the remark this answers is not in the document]
ChenRunyang: I want to get property before requesting the IRQ so that it can be used later in the irq_release flow. I will remove this function if the cpu_mask is cleared and the priority set to default value.

core/drivers/gic.c
ChenRunyang: OK, I will disable the irq in release flow.

core/include/kernel/interrupt.h
```diff
@@ -299,6 +301,20 @@ static inline void interrupt_raise_sgi(struct itr_chip *chip, size_t itr_num,
...
 }
 
+/*
+ * interrupt_release() - Reset the property for a shared peripheral interrupt
+ * @chip	Interrupt controller
+ * @itr_num	Interrupt number to release
+ * @prio	Interrupt priority
+ * @cpu_mask	Mask of the CPUs targeted by the interrupt
+ */
+static inline void interrupt_release(struct itr_chip *chip, size_t itr_num,
```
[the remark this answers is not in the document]
ChenRunyang: OK, I will use interrupt_release_to_ns instead of interrupt_release.

core/include/kernel/interrupt.h
ChenRunyang: Thanks for your comment, It's sorted them in descending order by the first letter

core/drivers/gic.c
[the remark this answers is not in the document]
jenswi-linaro: Sure, that makes sense.

jenswi-linaro:
By the way, rebasing the patches on the latest should take care of the failing test case.

jenswi-linaro:
Please use rebase instead of merge, we don't accept merges in pull requests.

core/drivers/gic.c
```diff
@@ -858,6 +858,30 @@ void gic_dump_state(void)
 	}
 }
 
+void gic_spi_release_to_ns(size_t it)
+{
+	struct gic_data *gd = &gic_data;
+	size_t idx = it / NUM_INTS_PER_REG;
+	uint32_t mask = 1 << (it % NUM_INTS_PER_REG);
```
jenswi-linaro, line 865: Please use the `BIT32()` macro instead. I know we're doing a bit of both in this file, but using `BIT32()` is the preferred way.

jenswi-linaro (commented):

core/include/drivers/gic.h
```diff
@@ -56,4 +58,7 @@ void gic_init_per_cpu(void);
 
 /* Print GIC state to console */
 void gic_dump_state(void);
+
+/* Release the property for a shared peripheral interrupt */
```
etienne-lms, line 62: Explicitly say that the interrupt is re-assigned to the non-secure interrupt group and its priority reset to `GIC_SPI_PRI_NS_EL1`.
State also that the interrupt shall be disabled when this function is called.

core/drivers/gic.c
```diff
@@ -858,6 +858,30 @@ void gic_dump_state(void)
...
+
+	assert(it < gd->max_it && it >= GIC_SPI_BASE);
+	/* Assert it's already disabled */
+	assert(!gic_it_is_enabled(gd, it));
```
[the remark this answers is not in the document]
jenswi-linaro: Most errors here either cause an assert or a panic. I don't see how this error is a special case.
What should a caller do on this error that he shouldn't have done already?
This is a static programming error, not a dynamic error like malloc() failing.

core/drivers/gic.c
```diff
@@ -858,6 +858,30 @@ void gic_dump_state(void)
...
+	/* Assert it's secure to start with */
+	assert(!gic_it_get_group(gd, it));
+
+	gic_it_set_cpu_mask(gd, it, 0);
+	gic_it_set_prio(gd, it, GIC_SPI_PRI_NS_EL1);
+
+	/* Clear pending status */
+	io_write32(gd->gicd_base + GICD_ICPENDR(idx), mask);
+	/* Assign it to NS Group1 */
+	io_setbits32(gd->gicd_base + GICD_IGROUPR(idx), mask);
```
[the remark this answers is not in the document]
jenswi-linaro: I agree, since we're writing to a value register, not a set or clear bits register.
This function is in contrast to `void gic_init_donate_sgi_to_ns()` used post-boot when more than one core might be active in the secure world.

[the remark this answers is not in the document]
ChenRunyang: done

core/drivers/gic.c
ChenRunyang: done

core/drivers/gic.c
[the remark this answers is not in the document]
ChenRunyang: done

core/include/drivers/gic.h
ChenRunyang: done

etienne-lms:
@ChenRunyang, do you plan to update this P-R?

ChenRunyang:
> @ChenRunyang, do you plan to update this P-R?

Sorry for late, I've been on vacation recently,  I will modify the patch according to your suggestions.

jforissier:
@ChenRunyang it looks like this PR is almost ready, would you like to make the changes requested by @etienne-lms? Thanks.

core/drivers/gic.c
```diff
@@ -858,6 +860,36 @@ void gic_dump_state(void)
 	}
 }
 
+int gic_spi_release_to_ns(size_t it)
```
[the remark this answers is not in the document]
ChenRunyang: done

jforissier:
We don't want any merge commits in pull requests so please update (rebase) you branch to be master plus one commit. One way to do it is to `git reset --hard` your branch to the OP-TEE OS master branch, then `git cherry-pick 3d0e061`. Thanks.

core/drivers/gic.c
```diff
@@ -858,6 +860,36 @@ void gic_dump_state(void)
 	}
 }
 
+TEE_Result gic_spi_release_to_ns(size_t it)
+{
+	struct gic_data *gd = &gic_data;
+	size_t idx = it / NUM_INTS_PER_REG;
+	uint32_t mask = BIT32(it % NUM_INTS_PER_REG);
+
+	if (it < gd->max_it && it >= GIC_SPI_BASE)
```
etienne-lms, line 869: `if (it >= gd->max_it || it < GIC_SPI_BASE)`?

core/drivers/gic.c
```diff
@@ -858,6 +860,36 @@ void gic_dump_state(void)
...
+		return TEE_ERROR_BAD_PARAMETERS;
+	/* Make sure it's already disabled */
+	if (!gic_it_is_enabled(gd, it))
+		return TEE_ERROR_BAD_STATE;
+	/* Assert it's secure to start with */
+	if (!gic_it_get_group(gd, it))
+		return TEE_ERROR_SECURITY;
```
etienne-lms, line 876: `TEE_ERROR_BAD_STATE`?

core/drivers/gic.c
```diff
@@ -858,6 +860,36 @@ void gic_dump_state(void)
...
+
+	gic_it_set_cpu_mask(gd, it, 0);
+	gic_it_set_prio(gd, it, GIC_SPI_PRI_NS_EL1);
+
+	mutex_lock(&gic_mutex);
```
etienne-lms, line 881: It is not expected this function is called concurrently for the same interrupt, but for consistency I think it would be better the mutex is locked right before line 872, at least before line 878.

core/include/drivers/gic.h
```diff
@@ -56,4 +58,11 @@ void gic_init_per_cpu(void);
 
 /* Print GIC state to console */
 void gic_dump_state(void);
+
+/*
+ * Reassign one of the SPIs to normal world and set its priority to
+ * GIC_SPI_PRI_NS_EL1. Ensure that the interrupt is disabled when this
```
etienne-lms, line 64: "Ensure that the interrupt is **currently assigned to secure world and** disabled when this ..."

etienne-lms (commented):
Commit message header line is a bit to long and prefix is not right. Shorten to something like:
`drivers: gic: runtime assign interrupts to non-secure world`
?

etienne-lms (commented):
I missed a minor comment on commit message:
```patch
-This patch introduces gic_spi_release_to_ns API to release an interrupt
+Add gic_spi_release_to_ns() API function in GIC driver to release an interrupt

core/include/drivers/gic.h
ChenRunyang: Done

core/drivers/gic.c
ChenRunyang: Done

etienne-lms (commented):

jenswi-linaro:

jforissier:
@ChenRunyang would you mind adding the review tags from `Reviewed-by:` and `Acked-by:` tags given by @etienne-lms and @jenswi-linaro to the commit description? Thanks.

PR #6908, opened 2024-06-21
core: mm: add CFG_PGT_CACHE_ENTRIES

etienne-lms:
Add CFG_PGT_CACHE_ENTRIES to allow platforms to customize the page table cache size. This is needed for example when a platform is to support very large TAs of several dozen of Mbytes of private memory (code/data).

By the way, fix pgt_cache.h layout to have header files includes first followed by macro definitions.

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -1097,6 +1097,26 @@ ifeq (y-y,$(CFG_CORE_PREALLOC_EL0_TBLS)-$(CFG_WITH_PAGER))
 $(error "CFG_WITH_PAGER can't support CFG_CORE_PREALLOC_EL0_TBLS")
 endif
 
+# CFG_PGT_CACHE_ENTRIES defines the number of entries on the memory
+# mapping page table cache used for Trusted Application mapping.
+# CFG_PGT_CACHE_ENTRIES is ignored when CFG_CORE_PREALLOC_EL0_TBLS
+# is enabled.
+#
+# A proper value for CFG_PGT_CACHE_ENTRIES depends on many factors:
+# CFG_WITH_LPAE, CFG_TA_ASLR, size of TAs, size of memrefs passed
+# to TA, CFG_ULIBS_SHARED and possibly others. The default value
+# is based on the number of threads as an indicator on how large
+# the system might be.
+ifeq ($(CFG_NUM_THREADS),1)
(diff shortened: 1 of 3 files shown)


mk/config.mk
```diff
@@ -1097,6 +1097,21 @@ ifeq (y-y,$(CFG_CORE_PREALLOC_EL0_TBLS)-$(CFG_WITH_PAGER))
 $(error "CFG_WITH_PAGER can't support CFG_CORE_PREALLOC_EL0_TBLS")
 endif
 
+# CFG_PGT_CACHE_ENTRIES defines the number of entries on the memory
```
jenswi-linaro, line 1100: Please mention that `CFG_PGT_CACHE_ENTRIES` is ignored if `CFG_CORE_PREALLOC_EL0_TBLS=y`.

mk/config.mk
```diff
@@ -1097,6 +1097,24 @@ ifeq (y-y,$(CFG_CORE_PREALLOC_EL0_TBLS)-$(CFG_WITH_PAGER))
...
+#
+# A proper value for CFG_PGT_CACHE_ENTRIES depends on many factors:
+# CFG_WITH_LPAE, CFG_TA_ASLR, size of TAs, size of memrefs passed
+# to TA, CFG_ULIBS_SHARED and possibly others. The default value
+# is based on the number of threads as an indicator on how large
+# the system might be.
+ifeq ($(CFG_NUM_THREADS),1)
+CFG_PGT_CACHE_ENTRIES ?= 4
+endif
+ifeq ($(CFG_NUM_THREADS)-$(CFG_WITH_LPAE),2-n)
```
jenswi-linaro, line 1113: We can only check `y/n` variables against `y`.

etienne-lms: Oh... that's right, thanks.
The below is a bit heavy, but I guess there's no alternate way.
```make
ifeq ($(CFG_NUM_THREADS),2)
ifneq ($(CFG_WITH_LPAE),y)
CFG_PGT_CACHE_ENTRIES ?= 8
endif
endif
```

jenswi-linaro: Yeah, but it's readable. :-)

etienne-lms: Thanks. Addressed.

jenswi-linaro (commented):

etienne-lms:

PR #6912, opened 2024-06-25
fix use of calloc() against GCC 14.1 directive -Wcalloc-transposed-args

etienne-lms:
This change complements https://github.com/OP-TEE/optee_os/pull/6853.
Changes in libtomcrypt are under review in https://github.com/libtom/libtomcrypt/pull/647.
No functional changes.

### DIFF
--- a/core/arch/arm/kernel/thread_spmc.c
+++ b/core/arch/arm/kernel/thread_spmc.c
@@ -1109,7 +1109,7 @@ static int add_mem_share(struct ffa_mem_transaction_x *mem_trans,
 		return FFA_NO_MEMORY;
 
 	if (flen != blen) {
-		struct mem_frag_state *s = calloc(sizeof(*s), 1);
+		struct mem_frag_state *s = calloc(1, sizeof(*s));
 
 		if (!s) {
 			rc = FFA_NO_MEMORY;
--- a/core/arch/arm/mm/sp_mem.c
+++ b/core/arch/arm/mm/sp_mem.c
@@ -224,7 +224,7 @@ struct sp_mem *sp_mem_new(void)
 	uint32_t exceptions = 0;
 	int i = 0;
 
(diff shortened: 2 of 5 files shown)


GseoC:

GseoC:
By the way, when issues are found in an external component, what are the steps to follow?
Should we fix it in OP-TEE and in the component independently or should we wait for the merge in the component tree and backport it in OP-TEE?

etienne-lms:
I think the issue should be fixed in both: in the original component and in OP-TEE. Fixing the original component does not means OP-TEE should upgrade to the latest (fixed) release of that component which may have other implication. Picking the fix can be enough.

For the very case of this change, as it is quite simple, https://github.com/libtom/libtomcrypt/pull/647 has been quite quickly merged. I think it ok we only pick the fix, as proposed in this P-R.

Maybe should I refer to https://github.com/libtom/libtomcrypt/pull/647 (or to https://github.com/libtom/libtomcrypt/commit/2302a3a89752b317d59e9cdb67d2d4eb9b53be8e) in the commit message.

For info: the checkpatch warning report is due to the build error trace message that exceeds 75 char/line in the commit message.

jenswi-linaro:
> Maybe should I refer to libtom/libtomcrypt#647 (or to libtom/libtomcrypt@2302a3a) in the commit message.

Yes, please.

jforissier:
One minor thing: for the second commit, I'd rather use the `libtomcrypt:` prefix than `core: ltc` since it is really a libtomcrypt fix (as opposed to a fix in the integration code with the core).
For both commits:

etienne-lms:
I've updated the commit messages and applied posted review tags.

jenswi-linaro:

jforissier:
FYI I have cherry-picked the `libtomcrypt` patch on top of `import/libtomcrypt-1.18.2-develop-20240412`.

PR #6922, opened 2024-06-28
plat-stm32mp1: conf: restore generic default heap size

etienne-lms:
Remove reduced default heap size configuration of 48kB when pager is enabled on stm32mp1 platforms. 48kB of core heap may not always be enough to pass OP-TEE Test regression test 4011 related to Bleichenbacher attack since it consumes 4.5kB more memory on in OP-TEE core since we upgraded to Mbed TLS library 3.6.0. The platform now default uses the generic 64kB default heap size set from mk/config.mk.


etienne-lms:
Qemu failure on pkcs11_1007 test (CI make check (QEMUv7)) seems unrelated to this P-R. Do we have an older regression here? I'll try to have a look.

jenswi-linaro:
That one rings a bell, I've seen occasional failures in that case too.

GseoC:

etienne-lms:
Review tag applied.

Issue #6870, opened 2024-05-30
QEMU: syscall ftrace does not work - RCU


jforissier:
Platform: QEMUv7
```
$ make -j$(nproc) CFG_FTRACE_SUPPORT=y CFG_SYSCALL_FTRACE=y CFG_FTRACE_BUF_SIZE=1000000 CFLAGS_ta_arm32=-pg QEMU_VIRTFS_AUTOMOUNT=y run
```
NW console:
```
optee: probing for conduit method.
optee: revision 4.2 (af3fb624)
optee: Asynchronous notifications enabled
optee: dynamic shared memory is enabled
optee: initialized driver
NET: Registered PF_PACKET protocol family
9pnet: Installing 9P2000 support
Registering SWP/SWPB emulation handler
rcu: INFO: rcu_sched detected stalls on CPUs/tasks:
rcu:    0-...0: (6 ticks this GP) idle=0174/1/0x40000000 softirq=97/101 fqs=995
rcu:    (detected by 1, t=2103 jiffies, g=-1107, q=13 ncpus=2)
```
SW console:
```
D/LD:  ldelf:176 ELF (f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c) at 0x123000
```

So it looks like it hangs as soon as it enters the trusted keys TA.

Commenting out `cflags$(sm)     += -pg` in `core/core.mk` solves the problem. Adding a few `-pg` here and there works (for example in Mbed TLS). So I guess we need to find out which core function should not be instrumented.

jforissier:
Still reproducible.

jenswi-linaro:
Should be fixed by https://github.com/OP-TEE/optee_os/pull/6930

PR #6930, opened 2024-07-05
arm32: ftrace: skip profiling of __aeabi functions

jenswi-linaro:
When compiling the __aeabi functions, skip profiling if CFG_FTRACE_SUPPORT is enabled to avoid recursive calls.

Closes: https://github.com/OP-TEE/optee_os/issues/6870

### DIFF
--- a/lib/libutils/isoc/arch/arm/sub.mk
+++ b/lib/libutils/isoc/arch/arm/sub.mk
@@ -7,7 +7,6 @@ srcs-$(CFG_ARM32_$(sm)) += arm32_aeabi_ldivmod_a32.S
 srcs-$(CFG_ARM32_$(sm)) += arm32_aeabi_ldivmod.c
 srcs-$(CFG_ARM32_$(sm)) += arm32_aeabi_shift.c
 
-ifeq ($(CFG_ULIBS_MCOUNT),y)
 # We would not like to profile __aeabi functions as these provide
 # internal implementations for "/ %" operations. Also, "/ %" operations
 # could be used inside profiling code which could create an incorrect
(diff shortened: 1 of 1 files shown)


lib/libutils/isoc/arch/arm/sub.mk
```diff
@@ -7,7 +7,7 @@ srcs-$(CFG_ARM32_$(sm)) += arm32_aeabi_ldivmod_a32.S
 srcs-$(CFG_ARM32_$(sm)) += arm32_aeabi_ldivmod.c
 srcs-$(CFG_ARM32_$(sm)) += arm32_aeabi_shift.c
 
-ifeq ($(CFG_ULIBS_MCOUNT),y)
+ifeq ($(call cfg-one-enabled, CFG_ULIBS_MCOUNT CFG_FTRACE_SUPPORT),y)
```
jforissier, line 10: To be precise that's `CFG_SYSCALL_FTRACE` that should be used here instead of `CFG_FTRACE_SUPPORT`, because what this block needs to do is revert what is done by:
- [lib.mk](https://github.com/OP-TEE/optee_os/blob/6f48ab5d3e350bca562beced9e3a261f24f02396/mk/lib.mk#L22-L26): adding `-pg` to the user space build of libutils when `CFG_ULIBS_MCOUNT=y`
- [core.mk](https://github.com/OP-TEE/optee_os/blob/6f48ab5d3e350bca562beced9e3a261f24f02396/core/core.mk#L72-L74): adding `-pg` to the kernel space build of libutils when `CFG_SYSCALL_FTRACE=y`

That being said it would be much simpler to just remove `-pg` unconditionally since in fact we *never* want it. Especially since the test is also missing the kernel vs. user mode aspect. So I suggest simply dropping the `ifeq / endif` here.

jforissier (commented):
Good catch! Please see below.

jenswi-linaro:
Updated as requested.

jforissier (approved):
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (vexpress-qemu_virt)`

jenswi-linaro:
Tags applied.

PR #6935, opened 2024-07-08
core: fix undefined behavior aborts during xtest

Abhishek-612:
1. Replace i64 with u64 in `lib/libtomcrypt/tweetnacl.c` to avoid `shift_out_of_bounds` 
error by UBSan.

2. Add null pointer checks in `kernel/user_access.c` before memcpy to avoid 
`nonnull_args` error by UBSan.

Signed-off-by: Abhishek Revadekar [abhishek.rvdkr@yahoo.com](mailto:abhishek.rvdkr@yahoo.com)

### DIFF
--- a/core/kernel/user_access.c
+++ b/core/kernel/user_access.c
@@ -51,7 +51,7 @@ TEE_Result copy_from_user(void *kaddr, const void *uaddr, size_t len)
 
 	uaddr = memtag_strip_tag_const(uaddr);
 	res = check_user_access(flags, uaddr, len);
-	if (!res) {
+	if (!res && kaddr && uaddr) {
 		enter_user_access();
 		memcpy(kaddr, uaddr, len);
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
It looks like you're using markup in the commit messages. Commit messages are supposed to be plain text so please drop all the back-ticks. With that fixed, please apply:

Abhishek-612:
> It looks like you're using markup in the commit messages. Commit messages are supposed to be plain text so please drop all the back-ticks. With that fixed, please apply: `Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>`

Commit messages have been fixed along with the Review tags.

jforissier:
The second commit should also be proposed upstream (https://github.com/libtom/libtomcrypt) in order minimize the local changes.

PR #6937, opened 2024-07-09
TEE core will crash when CFG_TEE_CORE_DEBUG=y and CFG_MEMTAG=y

charles1129:
When CFG_TEE_CORE_DEBUG=y and CFG_MEMTAG=y, core will crash, if there are some modules call virt_to_phys() or core_vbuf_is() and so on. Because the MM can not find map by the tagged VA, so strip the tag of VA when check_pa_matches_va().

Acked-by: Jens Wiklander <jens.wiklander@linaro.org>

### DIFF
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -18,6 +18,7 @@
 #include <kernel/user_mode_ctx.h>
 #include <kernel/virtualization.h>
 #include <libfdt.h>
+#include <memtag.h>
 #include <mm/core_memprot.h>
 #include <mm/core_mmu.h>
 #include <mm/mobj.h>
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please remove the period, `.`, at the end of the commit title. I think the following tags would make more sense:
```
Suggested-by: Jens Wiklander <jens.wiklander@linaro.org>
Signed-off-by: wentao.sun <wentao.sun@amlogic.com>
```
and when the above is addressed apply:
```
```
It would be nice if someone else could review the patch as well since I'm biased.

charles1129:
ok, update the commit message.

jforissier (commented):

charles1129:
Update

PR #6939, opened 2024-07-10
Add NVMEM improvements and fixes

tprrt:
- Fix the length of the HUK and die-id NVMEM cell.
- Allow the use of a larger NVMEM cell for the HUK.
- Redefine HUK and die id cells in sama5d2.dtsi.
- Fix the regression test 4013.

### DIFF
--- a/core/drivers/nvmem/nvmem_huk.c
+++ b/core/drivers/nvmem/nvmem_huk.c
@@ -8,45 +8,55 @@
 #include <kernel/dt.h>
 #include <kernel/huk_subkey.h>
 #include <kernel/tee_common_otp.h>
+#include <libfdt.h>
 #include <malloc.h>
 #include <tee_api_defines.h>
 #include <tee_api_types.h>
+#include <trace.h>
 #include <types_ext.h>
 
-static struct nvmem_cell *huk_cell;
+static uint8_t *huk;
 
 TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 {
-	TEE_Result res = TEE_ERROR_GENERIC;
-	uint8_t *huk = NULL;
-	size_t len = 0;
-
-	res = nvmem_cell_malloc_and_read(huk_cell, &huk);
-	if (res)
-		goto out_free_cell;
-
-	if (len != HW_UNIQUE_KEY_LENGTH) {
-		res = TEE_ERROR_GENERIC;
-		goto out_free_cell;
+	if (!huk) {
+		EMSG("no HUK");
+		return TEE_ERROR_GENERIC;
 	}
 
 	memcpy(hwkey->data, huk, HW_UNIQUE_KEY_LENGTH);
 
-out_free_cell:
-	nvmem_put_cell(huk_cell);
-
-	return res;
-}
-
-static TEE_Result nvmem_huk_get_cell(const void *fdt, int node)
-{
-	return nvmem_get_cell_by_name(fdt, node, "hw_unique_key", &huk_cell);
+	return TEE_SUCCESS;
 }
 
 static TEE_Result nvmem_huk_probe(const void *fdt, int node,
 				  const void *compat_data __unused)
 {
-	return nvmem_huk_get_cell(fdt, node);
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct nvmem_cell *cell = NULL;
+	uint8_t *data = NULL;
+
+	res = nvmem_get_cell_by_name(fdt, node, "hw_unique_key", &cell);
+	if (res)
+		return res;
+
+	if (cell->len < HW_UNIQUE_KEY_LENGTH) {
+		EMSG("cell %s is too small", fdt_get_name(fdt, node, NULL));
+		nvmem_put_cell(cell);
+		return TEE_ERROR_GENERIC;
+	}
+
+	if (cell->len > HW_UNIQUE_KEY_LENGTH)
+		IMSG("nvmem_huk: HUK truncated from %zu to %u bytes",
+		     cell->len, HW_UNIQUE_KEY_LENGTH);
+
+	res = nvmem_cell_malloc_and_read(cell, &data);
+	if (!res)
+		huk = data;
+
+	nvmem_put_cell(cell);
+
+	return res;
 }
 
 static const struct dt_device_match nvmem_huk_match_table[] = {
--- a/core/arch/arm/dts/sama5d2.dtsi
+++ b/core/arch/arm/dts/sama5d2.dtsi
@@ -746,25 +746,24 @@
 				#address-cells = <1>;
 				#size-cells = <1>;
 
-				sfc_kr: cell@0 {
-					reg = <0x0 0x4>;
-					bits = <0 7>;
+				sfc_dr0: cell@20 {
+					reg = <0x20 0x20>;
 				};
 
-				sfc_dr0: cell@20 {
-					reg = <0x20 0x4>;
+				sfc_dr1: cell@24 {
+					reg = <0x24 0x20>;
 				};
 			};
 
 			die_id: die_id {
 				compatible = "optee,nvmem-die-id";
-				nvmem-cells = <&sfc_kr>;
+				nvmem-cells = <&sfc_dr0>;
 				nvmem-cell-names = "die_id";
 			};
 
 			huk: huk {
 				compatible = "optee,nvmem-huk";
-				nvmem-cells = <&sfc_dr0>;
+				nvmem-cells = <&sfc_dr1>;
 				nvmem-cell-names = "hw_unique_key";
 			};
 
--- a/core/drivers/nvmem/atmel_sfc.c
+++ b/core/drivers/nvmem/atmel_sfc.c
@@ -32,6 +32,9 @@ static TEE_Result atmel_sfc_read_cell(struct nvmem_cell *cell, uint8_t *data)
 {
 	struct atmel_sfc *atmel_sfc = cell->drv_data;
 
+	if (cell->offset + cell->len > ATMEL_SFC_CELLS_8)
+		return TEE_ERROR_GENERIC;
+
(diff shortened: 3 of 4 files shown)


core/drivers/nvmem/nvmem_huk.c
```diff
@@ -19,19 +19,20 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
+	if (huk_cell->len < HW_UNIQUE_KEY_LENGTH) {
 		res = TEE_ERROR_GENERIC;
 		goto out_free_cell;
 	}
 
+	res = nvmem_cell_malloc_and_read(huk_cell, &huk);
+	if (res)
+		goto out_free_cell;
+
 	memcpy(hwkey->data, huk, HW_UNIQUE_KEY_LENGTH);
```
TonyHan11, line 32: As the data copied is counted with the unit byte, the `bits` need to be replaced with `bytes` in the commit messages of 31059bc & 086e75a

tprrt: Hello @TonyHan11,

> As the data copied is counted with the unit byte, the `bits` need to be replaced with `bytes` in the commit messages of [31059bc](https://github.com/OP-TEE/optee_os/commit/31059bc92c18a8ca0e73cac4ac139faa0bb68c9e) & [086e75a](https://github.com/OP-TEE/optee_os/commit/086e75a460946d0577d7e93fbc5cfbdc284e4163)

Thank you, I corrected commit messages.

Kind regards, 
Thomas

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -31,6 +31,8 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 
 	memcpy(hwkey->data, huk, HW_UNIQUE_KEY_LENGTH);
 
+	return res;
+
 out_free_cell:
 	nvmem_put_cell(huk_cell);
```
etienne-lms, line 37: With the change you propose, I think you should not "put" the NVMEM cell even if line 23 fails.
The test on the NVMEM cell regarding `HW_UNIQUE_KEY_LENGTH` should rather be done from `nvmem_huk_probe()` (or `nvmem_huk_get_cell()` if you prefer) and `tee_otp_get_hw_unique_key()` to return an error code if `huk_cell` is NULL.

As `tee_otp_get_hw_unique_key()` is likely to be called during OP-TEE initialization, I think you could even get the cell data straight from `nvmem_huk_probe()` so that if probe succeeds, `tee_otp_get_hw_unique_key()` will always succeed also.

etienne-lms (commented):
For commit "drivers: nvmem: atmel_sfc: check the length before read cell"
Perfer have a commit message body that fully describes the change without one needing to read the header line. In the case, something like:
```patch
 drivers: nvmem: atmel_sfc: check the length before read cell
 
+Check requested NVMEM cells length in Atmel SFC driver
 to prevent issues in case the length is greater than the fuse size.
```

Ditto in the next commit:
```patch
 driver: nvmem: nvmem-die-id: fix the length
 
-Obviously, the cell length wasn't set during the probe then the
+Obviously, the cell length wasn't set during nvmem-die-id driver probe and the
 subkey length was always used.
```

For commit "drivers: nvmem: nvmem-huk: fix the regression test 4013":
In commit message: s/huk/HUK/. This commit deserves a `Fixes:` tag IMHO.

For commit "dts: sama5d2: update NVMEM cells", this looks like a fix. Could you add a `Fixes:` tag.

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -29,6 +29,8 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 
 	memcpy(hwkey->data, huk, HW_UNIQUE_KEY_LENGTH);
 
+	return res;
```
etienne-lms, line 32: I see an issue here. Each call to `tee_otp_get_hw_unique_key()` will allocate a new instance of the NVMEM cell data. It would be much more simple to fully prepare a reliable `huk_cell` reference from `nvmem_huk_probe()`.

etienne-lms (commented):
For commit "drivers: nvmem: nvmem-huk: fix the regression test 4013":
remove the empty line between the `Fixes:` tag and the S-o-b tag.

For commit "dts: sama5d2: update NVMEM cells": could you add a `Fixes:` tag?

"drivers: nvmem: atmel_sfc: check the length before read cell",
"driver: nvmem: nvmem-die-id: fix the length" and
"drivers: nvmem: nvmem-huk: allow to use a larger cell".

etienne-lms:
By the way: fix typo in commit headere line:
```patch
-driver: nvmem: nvmem-die-id: fix the length
+drivers: nvmem: nvmem-die-id: fix cell length
```

tprrt:
Hi @etienne-lms,
I took all your comments into account.
Kind regards,
Thomas

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -13,39 +13,39 @@
...
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct nvmem_cell *cell = NULL;
+	uint8_t *data = NULL;
+
+	res = nvmem_get_cell_by_name(fdt, node, "hw_unique_key", &cell);
+	if (res)
+		return res;
+
+	if (cell->len < HW_UNIQUE_KEY_LENGTH)
+		return TEE_ERROR_GENERIC;
```
etienne-lms, line 40: `nvmem_put_cell()`
I think some trace messages could be useful. Suggestion:
```c
	if (huk_cell->len < HW_UNIQUE_KEY_LENGTH) {
		EMSG("cell %s is too small",
		nvmem_put_cell(huk_cell);	
		huk_cell = NULL;
		return TEE_ERROR_GENERIC;
	}

	if (huk_cell->len > HW_UNIQUE_KEY_LENGTH)
		IMSG("nvmem_huk: HUK truncated from %zu to %u bytes",
		     huk_cell->len, HW_UNIQUE_KEY_LENGTH);
```

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -13,39 +13,39 @@
 #include <tee_api_types.h>
 #include <types_ext.h>
 
-static struct nvmem_cell *huk_cell;
+static	uint8_t *huk;
```
etienne-lms, line 16: replace tab with a single space char

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -13,39 +13,39 @@
...
 
 TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 {
-	TEE_Result res = TEE_ERROR_GENERIC;
-	uint8_t *huk = NULL;
-
-	if (huk_cell->len < HW_UNIQUE_KEY_LENGTH)
+	if (!huk)
```
etienne-lms, line 20: `EMSG("no HUK")` would be welcome I think.

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -13,39 +13,39 @@
...
+	if (cell->len < HW_UNIQUE_KEY_LENGTH)
+		return TEE_ERROR_GENERIC;
+
+	res = nvmem_cell_malloc_and_read(cell, &data);
+	if (!res)
+		huk = data;
+
+	nvmem_put_cell(cell);
+
+	return TEE_SUCCESS;
```
etienne-lms, line 48: `return res`

etienne-lms (commented):
Commit "drivers: nvmem: nvmem-huk: allow to use a larger cell" fixes an issue but leave the driver with a buggy behavior (the API function can be safely called only once).
Commit "drivers: nvmem: nvmem-huk: fix the regression test 4013" introduces another issue.
Commit "drivers: nvmem: nvmem-huk: allocate NVMEM cell data in the probe" aims the fix the whole driver. 
=> I think you should squash those 3 commits into a single one that would really fix everything.

For commit "dts: sama5d2: update NVMEM cells": could you add a Fixes: tag?

Fix typo in commit header line:
```patch
-driver: nvmem: nvmem-die-id: fix the length
+drivers: nvmem: nvmem-die-id: fix cell length

File: core/drivers/nvmem/nvmem_huk.c [hunk 8]
```diff
@@ -8,45 +8,57 @@
...
+	struct nvmem_cell *cell = NULL;
+	uint8_t *data = NULL;
+
+	res = nvmem_get_cell_by_name(fdt, node, "hw_unique_key", &cell);
+	if (res)
+		return res;
+
+	if (cell->len < HW_UNIQUE_KEY_LENGTH) {
+		EMSG("cell %s is too small",
+		     fdt_get_name(fdt, node, NULL));
```
etienne-lms, line 45: my bad, this fits in a single line.

core/drivers/nvmem/nvmem_huk.c
```diff
@@ -8,45 +8,57 @@
...
+		nvmem_put_cell(cell);
+		huk = NULL;
```
etienne-lms, line 47: sorry, my suggestion was not clean. No need to clear `huk` here.

etienne-lms (commented):
For commit "drivers: nvmem: nvmem-huk: fix the regression test 4013":
The change does not aim at fix a regression test, rather at fixing the driver hence prefer another commit header line, e.g.:
`drivers: nvmem: nvmem-huk: fix cell size and memory allocation issues`

etienne-lms (commented):
"drivers: nvmem: nvmem-huk: fix cell size and memory allocation issues".

For commit "dts: sama5d2: update NVMEM cells":
please also add `Fixes: 6c6c4d9eb45d ("dts: sama5d2: add NVMEM die_id node")` as the change also fixes that commit. With this comment addressed:
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

etienne-lms (approved):

PR #6944, opened 2024-07-12
libutils: qsort helper for standard types

etienne-lms:
Add `qsort_int()`, `qsort_uint()`, `qsort_u32()` and friends as `qsort()` helper functions to sort standard typed arrays.

I needed to use `qsort()` in few separated source files and needed to redefine locally the `cmp_xxx()` callback so I think it's worth having some generic helper functions.

### DIFF
--- a/lib/libutils/ext/qsort_helpers.c
+++ b/lib/libutils/ext/qsort_helpers.c
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024, STMicroelectronics
+ */
+
+#include <stdlib.h>
+#include <util.h>
+
+#define QSORT_HELPER(name, type)				\
+static int cmp_ ## name(const void *a, const void *b)		\
+{								\
+	const type *ia = a;					\
+	const type *ib = b;					\
+								\
+	return CMP_TRILEAN(*ia, *ib);				\
+}								\
+								\
+void qsort_ ## name(type *aa, size_t n)				\
+{								\
+	qsort(aa, n, sizeof(*aa), cmp_ ## name);		\
+}
+
+QSORT_HELPER(int, int);
+QSORT_HELPER(uint, unsigned int);
+QSORT_HELPER(long, long int);
+QSORT_HELPER(ul, unsigned long int);
+QSORT_HELPER(ll, long long int);
+QSORT_HELPER(ull, unsigned long long int);
+QSORT_HELPER(s8, int8_t);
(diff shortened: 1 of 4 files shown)


lib/libutils/ext/qsort_helpers.c
```diff
@@ -0,0 +1,189 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) 2024, STMicroelectronics
+ */
+
+#include <stdlib.h>
+#include <util.h>
+
+static int qsort_cmp_int(const void *a, const void *b)
```
jenswi-linaro, line 9: There's a lot of repeated code here. How about something like:
``` C
#define QSORT_HELPLER(name, type)				\
static int cmp_ ## name(const void *a, const void *b)		\
{								\
	const type *ia = a;					\
	const type *ib = b;					\
								\
	return CMP_TRILEAN(*ia, *ib);				\
}								\
								\
void qsort_ ## name(type *aa, size_t n)				\
{								\
	qsort(aa, n, sizeof(*aa), cmp_ ## name);			\
}

QSORT_HELPLER(int, int);
QSORT_HELPLER(uint, unsigned int);
```

etienne-lms: Indeed, it would be more concise. I'll update. Thanks.

etienne-lms:
Comment addressed, thanks gain for the macro. I've straight fix the commit instead of appending a fixup commit.

lib/libutils/ext/qsort_helpers.c
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) 2024, STMicroelectronics
+ */
+
+#include <stdlib.h>
+#include <util.h>
+
+#define QSORT_HELPLER(name, type)				\
```
jenswi-linaro, line 9: I'm sorry, it seems that you've copied my typo. It should be `QSORT_HELPER`

jenswi-linaro:
With my comment addressed:

lib/libutils/ext/qsort_helpers.c
etienne-lms: Indeed!!!

etienne-lms:
Macro name fixed. I also changed the license terms: it was BSD-3-Clause in my initial post, but I think BSD-2-Clause better applies since it is OP-TEE OS native license terms.
With these 2 changes, I've applied your R-b tag that I think is still accurate.

PR #6945, opened 2024-07-12
Required changes for PACBTI enablement in Trusted Services

gabor-toth-arm:
- Enable BTI for binary SP-s via manifest (and provide info to the SP whether BTI/PAC are enabled by the OS)
- Enable PAC for SP-s (similar way as it is implemented for TA-s)
- Enable PAC before SP initialization and se atomic pauth keys at init, so the used keys will not change between the first SEL1->SEL0->SEL1 transition resulting in error

### DIFF
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -1400,6 +1400,22 @@ static TEE_Result handle_hw_features(void *fdt)
 			return res;
 	}
 
+	/* Modify the property only if it's already present */
+	if (!sp_dt_get_u32(fdt, node, "bti", &val)) {
+		res = fdt_setprop_u32(fdt, node, "bti",
+				      feat_bti_is_implemented());
+		if (res)
+			return res;
+	}
+
+	/* Modify the property only if it's already present */
+	if (!sp_dt_get_u32(fdt, node, "pauth", &val)) {
+		res = fdt_setprop_u32(fdt, node, "pauth",
+				      feat_pauth_is_implemented());
+		if (res)
+			return res;
+	}
+
 	return TEE_SUCCESS;
 }
 
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -215,6 +215,10 @@ static TEE_Result sp_create_ctx(const TEE_UUID *bin_uuid, struct sp_session *s)
 
 	set_sp_ctx_ops(&spc->ts_ctx);
 
+#ifdef CFG_TA_PAUTH
+	crypto_rng_read(&spc->uctx.keys, sizeof(spc->uctx.keys));
+#endif
+
 	return TEE_SUCCESS;
 
 err:
@@ -1583,6 +1587,10 @@ TEE_Result sp_enter(struct thread_smc_args *args, struct sp_session *sp)
 	ctx->sp_regs.x[5] = args->a5;
 	ctx->sp_regs.x[6] = args->a6;
 	ctx->sp_regs.x[7] = args->a7;
+#ifdef CFG_TA_PAUTH
+	ctx->sp_regs.apiakey_hi = ctx->uctx.keys.apia_hi;
+	ctx->sp_regs.apiakey_lo = ctx->uctx.keys.apia_lo;
+#endif
 
 	res = sp->ts_sess.ctx->ops->enter_invoke_cmd(&sp->ts_sess, 0);
 
--- a/core/arch/arm/kernel/secure_partition.c
+++ b/core/arch/arm/kernel/secure_partition.c
@@ -37,6 +37,7 @@
 #define SP_MANIFEST_ATTR_WRITE		BIT(1)
 #define SP_MANIFEST_ATTR_EXEC		BIT(2)
 #define SP_MANIFEST_ATTR_NSEC		BIT(3)
+#define SP_MANIFEST_ATTR_GP		BIT(4)
 
 #define SP_MANIFEST_ATTR_RO		(SP_MANIFEST_ATTR_READ)
 #define SP_MANIFEST_ATTR_RW		(SP_MANIFEST_ATTR_READ | \
@@ -880,6 +881,15 @@ static TEE_Result handle_fdt_load_relative_mem_regions(struct sp_ctx *ctx,
 			return TEE_ERROR_BAD_FORMAT;
 		}
 
+		if (IS_ENABLED(CFG_TA_BTI) &&
+		    attributes & SP_MANIFEST_ATTR_GP) {
+			if (!(attributes & SP_MANIFEST_ATTR_RX)) {
+				EMSG("Guard only executable region");
+				return TEE_ERROR_BAD_FORMAT;
+			}
+			perm |= TEE_MATTR_GUARDED;
+		}
+
 		res = sp_dt_get_u32(fdt, subnode, "load-flags", &flags);
 		if (res != TEE_SUCCESS && res != TEE_ERROR_ITEM_NOT_FOUND) {
 			EMSG("Optional field with invalid value: flags");
@@ -1207,6 +1217,15 @@ static TEE_Result handle_fdt_mem_regions(struct sp_ctx *ctx, void *fdt)
 		default:
 			EMSG("Invalid memory access permissions");
 			return TEE_ERROR_BAD_FORMAT;
+		}
+
+		if (IS_ENABLED(CFG_TA_BTI) &&
+		    attributes & SP_MANIFEST_ATTR_GP) {
+			if (!(attributes & SP_MANIFEST_ATTR_RX)) {
+				EMSG("Guard only executable region");
+				return TEE_ERROR_BAD_FORMAT;
+			}
+			perm |= TEE_MATTR_GUARDED;
 		}
 
 		/*
--- a/core/arch/arm/kernel/thread.c
+++ b/core/arch/arm/kernel/thread.c
@@ -1008,10 +1008,18 @@ static struct thread_pauth_keys *thread_get_pauth_keys(void)
 {
 #if defined(CFG_TA_PAUTH)
 	struct ts_session *s = ts_get_current_session();
-	/* Only user TA's support the PAUTH keys */
-	struct user_ta_ctx *utc = to_user_ta_ctx(s->ctx);
 
-	return &utc->uctx.keys;
+	if  (is_user_ta_ctx(s->ctx)) {
+		struct user_ta_ctx *utc = to_user_ta_ctx(s->ctx);
+
+		return &utc->uctx.keys;
+	} else if (is_sp_ctx(s->ctx)) {
+		struct sp_ctx *spc = to_sp_ctx(s->ctx);
+
+		return &spc->uctx.keys;
+	}
+
+	panic("[abort] Only user TAs and SPs support PAUTH keys");
 #else
 	return NULL;
 #endif


core/arch/arm/kernel/secure_partition.c
```diff
@@ -1591,6 +1595,10 @@ TEE_Result sp_enter(struct thread_smc_args *args, struct sp_session *sp)
 	ctx->sp_regs.x[5] = args->a5;
 	ctx->sp_regs.x[6] = args->a6;
 	ctx->sp_regs.x[7] = args->a7;
+#if defined(ARM64) && (defined(CFG_TA_PAUTH) || defined(CFG_CORE_PAUTH))
```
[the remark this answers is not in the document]
gabor-toth-arm: It is not enough. apiakey_hi and apiakey_lo depends on all the three macros:

https://github.com/OP-TEE/optee_os/blob/932059bfed12a8024a995894a1c1c7ef86e9c1c1/core/arch/arm/include/kernel/thread_arch.h#L143

https://github.com/OP-TEE/optee_os/blob/932059bfed12a8024a995894a1c1c7ef86e9c1c1/core/arch/arm/include/kernel/thread_arch.h#L179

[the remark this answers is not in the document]
jenswi-linaro: `CFG_TA_PAUTH` can't be defined without `ARM64`, and the other condition is or:ed.

core/arch/arm/kernel/secure_partition.c
```diff
@@ -37,6 +37,9 @@
 #define SP_MANIFEST_ATTR_WRITE		BIT(1)
 #define SP_MANIFEST_ATTR_EXEC		BIT(2)
 #define SP_MANIFEST_ATTR_NSEC		BIT(3)
+#ifdef CFG_TA_BTI
```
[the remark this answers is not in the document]
gabor-toth-arm: Fixed

core/arch/arm/kernel/secure_partition.c
```diff
@@ -880,6 +883,18 @@ static TEE_Result handle_fdt_load_relative_mem_regions(struct sp_ctx *ctx,
 			return TEE_ERROR_BAD_FORMAT;
 		}
 
+#ifdef CFG_TA_BTI
```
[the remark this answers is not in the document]
gabor-toth-arm: Fixed

core/arch/arm/kernel/secure_partition.c
```diff
@@ -37,6 +37,9 @@
 #define SP_MANIFEST_ATTR_WRITE		BIT(1)
 #define SP_MANIFEST_ATTR_EXEC		BIT(2)
 #define SP_MANIFEST_ATTR_NSEC		BIT(3)
+#ifdef CFG_TA_BTI
+#define SP_MANIFEST_ATTR_GP BIT(4)
```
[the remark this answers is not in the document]
gabor-toth-arm: It is not included in the FFA spec yet. We have discussed it on slack and will be included in future version of the specification.

core/arch/arm/kernel/thread.c
```diff
@@ -1008,10 +1008,18 @@ static struct thread_pauth_keys *thread_get_pauth_keys(void)
...
-	/* Only user TA's support the PAUTH keys */
-	struct user_ta_ctx *utc = to_user_ta_ctx(s->ctx);
 
-	return &utc->uctx.keys;
+	if  (is_user_ta_ctx(s->ctx)) {
+		struct user_ta_ctx *utc = to_user_ta_ctx(s->ctx);
+
+		return &utc->uctx.keys;
+	} else if (is_sp_ctx(s->ctx)) {
+		struct sp_ctx *utc = to_sp_ctx(s->ctx);
```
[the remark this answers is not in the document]
gabor-toth-arm: Fixed

core/arch/arm/kernel/secure_partition.c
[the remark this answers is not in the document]
gabor-toth-arm: You are right. Fixed it.

core/arch/arm/kernel/entry_a64.S
```diff
@@ -137,9 +137,8 @@
 	.endm
 
 	.macro init_pauth_per_cpu
-		msr	spsel, #1
-		ldp	x0, x1, [sp, #THREAD_CORE_LOCAL_KEYS]
-		msr	spsel, #0
+		adr_l	x3, threads
```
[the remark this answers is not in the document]
gabor-toth-arm: In this case, I reworked this commit and I have a different approach of the problem. Instead of trying to change keys/context I tell the compiler with attribute to NOT compile with pointer authentication the functions which are affected the changing of the keys. This way the feature is already enabled when the SP is loaded, but some functions in optee core will not contain the paciasp/autiasp instructions avoiding the errors.
What is your opinion of this solution?

core/arch/arm/include/arm64.h
```diff
@@ -305,6 +305,20 @@
...
+ */
+#if (defined(CFG_CORE_PAUTH) || defined(CFG_TA_PAUTH))
+	#if (defined(CFG_CORE_BTI) || defined(CFG_TA_BTI))
+		#define DISABLE_PAUTH __attribute__((target("branch-protection=bti")))
+	#else
+		#define DISABLE_PAUTH __attribute__((target("branch-protection=none")))
+	#endif
+#else
+	#define DISABLE_PAUTH
+#endif
```
etienne-lms, line 320: prefer without the indentation.
can remove the outer parentheses at lines 312 and 313.

core/arch/arm/kernel/boot.c
```diff
@@ -1061,6 +1061,17 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
...
 		init_tee_runtime();
 	}
+}
+
+/*
+ * Note: this function is weak just to make it possible to exclude it from
+ * the unpaged area.
+ */
+#ifdef ARM64
+DISABLE_PAUTH
```
etienne-lms, line 1071: Could this macro be renamed `__pauth` and defined also in _arm32.h_ as `#define __pauth`? It would prevent need for `#ifdef ARM64`/`#endif`, e.g. here:
```c
void __weak __pauth boot_init_primary_finalcalls(void)

File: core/arch/arm/kernel/secure_partition.c [hunk 9]
```diff
@@ -37,6 +37,7 @@
 #define SP_MANIFEST_ATTR_WRITE		BIT(1)
 #define SP_MANIFEST_ATTR_EXEC		BIT(2)
 #define SP_MANIFEST_ATTR_NSEC		BIT(3)
+#define SP_MANIFEST_ATTR_GP			BIT(4)
```
etienne-lms, line 40: fix indentation

core/arch/arm/kernel/secure_partition.c
```diff
@@ -880,6 +891,18 @@ static TEE_Result handle_fdt_load_relative_mem_regions(struct sp_ctx *ctx,
...
 
+		if (IS_ENABLED(CFG_TA_BTI)) {
+			/* GP bit shall only be set for executable regions */
+			if (attributes & SP_MANIFEST_ATTR_GP) {
+				if (attributes & SP_MANIFEST_ATTR_RX) {
+					perm |= TEE_MATTR_GUARDED;
+				} else {
+					EMSG("Guard only executable region");
+					return TEE_ERROR_BAD_FORMAT;
+				}
```
etienne-lms, line 902: nitpicking:
```c
		if (IS_ENABLED(CFG_TA_BTI) &&
		    attributes & SP_MANIFEST_ATTR_GP) {
			if (!attributes & SP_MANIFEST_ATTR_RX)) {
				EMSG("Guard only executable region");
				return TEE_ERROR_BAD_FORMAT;
			}
			perm |= TEE_MATTR_GUARDED;
		}
```
Ditto at line 1235.

core/kernel/initcall.c
```diff
@@ -7,6 +7,9 @@
 #include <initcall.h>
 #include <trace.h>
 #include <kernel/linker.h>
+#ifdef ARM64
+#include <arm64.h>
```
etienne-lms, line 11:
```patch
+#include <arm.h>
 #include <initcall.h>
 #include <trace.h>
 #include <kernel/linker.h>
-#ifdef ARM64
-#include <arm64.h>
-#endif
```

By the way,  _kernel/linker.h_ should be reordered but that out of the scope of this P-R:
```c
#include <arm.h>
#include <initcall.h>
#include <kernel/linker.h>
#include <trace.h>
```

core/arch/arm/include/arm64.h
```diff
@@ -305,6 +305,20 @@
 #define ID_AA64ISAR0_SHA2_FEAT_SHA256	U(1)
 #define ID_AA64ISAR0_SHA2_FEAT_SHA512	U(2)
 
+/* Pauth keys change from atomic ones to thread ones during SP initialization.
```
etienne-lms, line 308:
```patch
-/* Pauth keys change from atomic ones to thread ones during SP initialization.
+/*
+ * Pauth keys change from atomic ones to thread ones during SP initialization.

**Review by etienne-lms — COMMENTED** [review]
Some coding style comments

File: core/arch/arm/include/arm64.h (hunk 7 above)
Comment (gabor-toth-arm) [reply core/arch/arm/include/arm64.h:321]: Fixed

File: core/arch/arm/kernel/boot.c (hunk 8 above)
Comment (gabor-toth-arm) [reply core/arch/arm/kernel/boot.c]: Fixed

File: core/kernel/initcall.c (hunk 11 above)
Comment (gabor-toth-arm) [reply core/kernel/initcall.c]: Fixed

File: core/arch/arm/include/arm64.h (hunk 12 above)
Comment (gabor-toth-arm) [reply core/arch/arm/include/arm64.h]: Fixed

File: core/arch/arm/kernel/secure_partition.c (hunk 10 above)
Comment (gabor-toth-arm) [reply core/arch/arm/kernel/secure_partition.c]: Fixed

File: core/arch/arm/kernel/secure_partition.c (hunk 9 above)
Comment (gabor-toth-arm) [reply core/arch/arm/kernel/secure_partition.c]: Fixed

File: core/kernel/initcall.c [hunk 13]
```diff
@@ -4,10 +4,18 @@
...
+#include <arm.h>
+#endif
 #include <initcall.h>
 #include <trace.h>
 #include <kernel/linker.h>
 
+/* Define this manually on non-arm architectures */
+#ifndef __pauth
+#define __pauth
+#endif
```
etienne-lms, line 17: I wonder if `__pauth` macro should rather be defined in _lib/libutils/ext/include/compiler.h_. @jforissier, what do you think?

core/arch/arm/kernel/entry_a32.S
```diff
@@ -253,7 +253,7 @@ DECLARE_KEEP_INIT _start
 		 * equivalent of:
 		 * extern const u32 stack_tmp_stride;
 		 * u32 stack_tmp_stride_rel = (u32)&stack_tmp_stride -
-		 *			      (u32)&stack_tmp_stride_rel
```
etienne-lms, line 256: unrelated change. Prefer in a dedicated commit.

core/arch/arm/include/arm64.h
```diff
@@ -305,6 +305,21 @@
 #define ID_AA64ISAR0_SHA2_FEAT_SHA256	U(1)
 #define ID_AA64ISAR0_SHA2_FEAT_SHA512	U(2)
 
+/*
+ * Pauth keys change from atomic ones to thread ones during SP initialization.
+ * Pointer authentication must be disabled during this time or the functions
+ * will not be able to return.
```
etienne-lms, line 311: I'm a bit confused with this description. This means only the function tagged with attribute `__pauth` will enable BTI for PAUTH support?

etienne-lms (commented):
For commit "core:spmc: Enable BTI for binary SPs" message header line:
```patch
-core:spmc: Enable BTI for binary SPs
+core: spmc: Enable BTI for binary SPs
```

For commit "core: spmc: handle bti/pauth info in SP manifest" message body:
s/bti/BTI/, s/pauth/PAUTH/, s/optee/OP-TEE/

With these addressed, 
"core:spmc: Enable BTI for binary SPs",
"core: Enable pointer authentication for SPs" and
"core: spmc: handle bti/pauth info in SP manifest".

core/arch/arm/include/arm64.h
jforissier: My understanding is that the macro is meant to disable PAuth only but leave BTI if it is enabled. In fact a better name would be `__nopauth`. It is a bit complex to do because the `branch-protection` attribute combines PAuth settings (`pac-ret+leaf` is what we enable in OP-TEE when `CFG_*_BTI` is enabled) as well as the BTI setting (`bti`).

The below definition isn't OK because it does not handle the core vs. TA case properly. It should not only test `CFG_CORE_BTI` and `CFG_TA_BTI` but also if the file being compiled matches those configs (`__KERNEL__`, `__LDELF__`...).

It looks like the usage is limited to kernel mode in `boot.c` and `secure_partition.c`. Therefore I suggest defining the macro in `kernel/boot.h` where it can be assumed it if for the core only:
```C
#if defined(CFG_CORE_PAUTH)
#if defined(CFG_CORE_BTI)
#define __kernel_nopauth __attribute__((target("branch-protection=bti")))
#else
#define __kernel_nopauth __attribute__((target("branch-protection=none")))
#endif
#else
#define __kernel_nopauth
#endif
```

This is still a bit ugly and I would prefer we could do differently. How is the SP case fundamentally different from the TA case?

core/arch/arm/kernel/secure_partition.c
```diff
@@ -581,6 +586,7 @@ static TEE_Result load_binary_sp(struct ts_session *s,
 	return res;
 }
 
+__pauth
 static TEE_Result sp_open_session(struct sp_session **sess,
```
jforissier, line 590: Attributes should come after the return type
```suggestion
static TEE_Result __pauth
sp_open_session(struct sp_session **sess,
```

core/arch/arm/kernel/entry_a64.S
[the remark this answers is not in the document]
jenswi-linaro: I'm not so keen on this approach. It's messy and confusing with the `__pauth` directives. I'd rather update the runtime environment to be ready for pointer authentication.

core/arch/arm/include/arm64.h
gabor-toth-arm: Yes, you are right about the intent of this macro. __nopauth is really a better name so I will rename it.
I will also remove CFG_TA_PAUTH and CFG_TA_BTI from the conditions of this macro, because you are right that they are not needed since the enable the protection for EL0.

And about your question.
In case of SP-s the problem is, that at the beginning of their initialization, the keys of the atomic context are used for PAUTH, and after the SP returns to, the kernel context keys are used. After this, when FFA messages arrive to the SP the kernel keys are used all the time both before entering EL0 and after leaving EL0. So the problem is that at init there is a switching of context (keys) which results in the PAUTH failure. 
As I see TA-s are started a bit later, when the SP-s are already initialized and kernel context is used all along, so there are no key changing.
Early TA-s are started way before enabling PAUTH or the initialization of keys.
To be honest I did not find any "rules" of the usage of atomic/kernel contexts, so I am not sure if the aforementioned behavior could be changed to avoid this issue.

core/arch/arm/kernel/secure_partition.c
gabor-toth-arm: Fixed

core/arch/arm/kernel/entry_a32.S
gabor-toth-arm: Was commited by mistake. Removed the change

core/kernel/initcall.c
gabor-toth-arm: Renamed to __kernel_nopauth  according Jerome's advice.

core/arch/arm/kernel/entry_a64.S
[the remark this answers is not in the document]
gabor-toth-arm: By "ready for pointer authentication" do you mean initializing the kernel thread and switching to it right before the call of call_finalcalls instead of using the atomic thread ?

jenswi-linaro:
https://github.com/OP-TEE/optee_os/pull/6988 should help with thread-specific pauth keys.

core/arch/arm/kernel/thread.c
```diff
@@ -1008,10 +1008,18 @@ static struct thread_pauth_keys *thread_get_pauth_keys(void)
...
+		struct user_ta_ctx *utc = to_user_ta_ctx(s->ctx);
+
+		return &utc->uctx.keys;
+	} else if (is_sp_ctx(s->ctx)) {
+		struct sp_ctx *spc = to_sp_ctx(s->ctx);
+
+		return &spc->uctx.keys;
+	}
+
+	panic("[abort] Only user TA's and SP-s support PAUTH keys");
```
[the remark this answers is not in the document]
gabor-toth-arm: Fixed

gabor-toth-arm:
> #6988 should help with thread-specific pauth keys.

I have just rebased my commits on your PR to test it and it works as expected. I removed my top two commits, because they are not needed after your changes.
Thank you for the help!

jforissier:
> > #6988 should help with thread-specific pauth keys.
> 
> I have just rebased my commits on your PR to test it and it works as expected.

Nice. Therefore I will merge this after #6988 is merged.

gabor-toth-arm:
> > > #6988 should help with thread-specific pauth keys.
> > 
> > 
> > I have just rebased my commits on your PR to test it and it works as expected.
> 
> Nice. Therefore I will merge this after #6988 is merged.

Hi Jerome!

https://github.com/OP-TEE/optee_os/pull/6988 has been merged last week. Could you also merge this PR and the related build PR https://github.com/OP-TEE/build/pull/771?

jforissier:
> > > > #6988 should help with thread-specific pauth keys.
> > > 
> > > 
> > > I have just rebased my commits on your PR to test it and it works as expected.
> > 
> > 
> > Nice. Therefore I will merge this after #6988 is merged.
> 
> Hi Jerome!
> 
> #6988 has been merged last week. Could you also merge this PR and the related build PR [OP-TEE/build#771](https://github.com/OP-TEE/build/pull/771)?

Merging this one but please see my comment on OP-TEE/build#771. Thanks!

PR #6946, opened 2024-07-16
core: firewall: add dedicated firewall configuration APIs

GseoC:
Add dedicated firewall device APIs for handling firewall configurations from the device tree. Firewall configurations are inseparable sets of one or more firewall queries.
These APIs are based on the ".*(?<=)-access-conf$" property. This property will be proposed in the Linux kernel.

While there, remove the check for foreign entities API as it seems unlikely it'll be used.

Edit: Link to kernel patch: https://lkml.org/lkml/2024/7/16/849

### DIFF
--- a/core/drivers/firewall/firewall.c
+++ b/core/drivers/firewall/firewall.c
@@ -10,38 +10,30 @@
 #include <kernel/panic.h>
 #include <libfdt.h>
 #include <malloc.h>
+#include <stdio.h>
 #include <trace.h>
 
 /* The firewall framework requires device tree support */
 static_assert(IS_ENABLED(CFG_DT));
 
 static TEE_Result firewall_get(struct dt_pargs *parg, void *data,
-			       struct firewall_query **out_fw)
+			       struct firewall_query *out_fw)
 {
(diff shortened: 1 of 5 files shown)


core/drivers/firewall/firewall.c
```diff
@@ -90,6 +176,57 @@ TEE_Result firewall_set_configuration(struct firewall_query *fw)
...
+	TEE_Result res = TEE_ERROR_GENERIC;
+	size_t i = 0;
+
+	assert(conf && conf->nb_queries);
+
+	for (i = 0; i < conf->nb_queries; i++) {
+		/*
+		 * In case of error, report it to the caller. Note that
+		 * the firewall configurations may be partially loaded. In such
+		 * case, it is the caller responsibility to decide what to do.
```
etienne-lms, line 190: This description should be also found in `firewall_set_configuration()` inline description comment.

etienne-lms (commented):
some nitpicking comments. The overall looks ok to me.

GseoC:
Keep alive (waiting for Linux feedback on the binding)

GseoC:
keep alive

PR #6951, opened 2024-07-18
drivers: scmi-msg: correct voltage domain protocol version & clock rates enumeration

etienne-lms:
Fix the version ID of the implemented SCMI voltage domain protocol that is v2.0 (ID 0x20000), not v3.0 (ID 0x30000).
(Fixes: 006d89b8f49f ("drivers: scmi-msg: add SCMI Voltage Domain protocol"))

Fix value of remaining clocks to describe on SCMI clock protocol message CLOCK_DESCRIBE_RATES that does not take into account the number of returned clock in the response.
(Fixes: 90252e2a52c7 ("drivers: scmi-msg: clock adapts to output buffer size"))

### DIFF
--- a/core/drivers/scmi-msg/clock.c
+++ b/core/drivers/scmi-msg/clock.c
@@ -321,7 +321,7 @@ static void scmi_clock_describe_rates(struct scmi_msg *msg)
 		}
 
 		out_count = rate_index - in_args->rate_index;
-		remaining = nb_rates - in_args->rate_index;
+		remaining = nb_rates - rate_index;
 		p2a.num_rates_flags = SCMI_RATES_BY_ARRAY(out_count, remaining);
 	} else if (status == SCMI_NOT_SUPPORTED) {
 		unsigned long triplet[3] = { 0, 0, 0 };
--- a/core/drivers/scmi-msg/voltage_domain.h
+++ b/core/drivers/scmi-msg/voltage_domain.h
@@ -12,7 +12,7 @@
 
 #include "common.h"
 
-#define SCMI_PROTOCOL_VERSION_VOLTAGE_DOMAIN	0x30000
+#define SCMI_PROTOCOL_VERSION_VOLTAGE_DOMAIN	0x20000
 
 /*
  * Identifiers of the SCMI Clock Management Protocol commands


core/drivers/scmi-msg/clock.c
```diff
@@ -321,7 +321,7 @@ static void scmi_clock_describe_rates(struct scmi_msg *msg)
 		}
 
 		out_count = rate_index - in_args->rate_index;
-		remaining = nb_rates - in_args->rate_index;
+		remaining = nb_rates - in_args->rate_index - out_count;
```
jforissier, line 324: `nb_rates - rate_index` would be better, no?

jforissier (commented):
For "drivers: scmi-msg: correct voltage domain protocol version":

core/drivers/scmi-msg/clock.c
GseoC: I would have said `nb_rates - out_count` as out_count is the number of entries filled in the buffer. Maybe I'm missing something.

etienne-lms: > nb_rates - rate_index would be better, no?

Indeed simpler.

> I would have said `nb_rates - out_count` as out_count is the number of entries filled in the buffer. Maybe I'm missing something.

`nb_rates` gives the overall number of supported rates. `out_count` gives the number of rates filled in this very response message but does not count the rates skipped, from index 0 to index `in_args->rate_index - 1`.

etienne-lms (commented):
I'll update the change.

jforissier:

GseoC (approved):

etienne-lms:
Fixup commit squashed and review tag applied.
@GseoC, do you want to provide a review tag?

pkcs11_1007 test sporadic failure faced during Ci test for this P-R but it's unrelated to this change (I still don't understand exactly what happens there, still tracking through https://github.com/OP-TEE/optee_os/issues/6952).

GseoC:
Sure, `Acked-by: Gatien Chevallier <gatien.chevallier@foss.st.com>`

etienne-lms:
Review tags applied. Thanks.

PR #6953, opened 2024-07-19
drivers: clk: clk_get_rates_array() returns ordered rates

etienne-lms:
Explicitly state in clk_get_rates_array() inline description comment that the output rates arrays is ordered by increasing frequency values. This change allows to better fit the sole consumer of this API function that is the SCMI server implementation. SCMI specification states that discrete clock rates list shall follow this order.

Update at91_cpu_opp clock driver to ensure it satisfy this constraint. The SAM platforms that embed this driver (sama7g5) already satisfy this constraints but only at its DTS level. This change ensures the driver will always.

### DIFF
--- a/core/drivers/clk/sam/at91_cpu_opp.c
+++ b/core/drivers/clk/sam/at91_cpu_opp.c
@@ -118,6 +118,10 @@ static TEE_Result opp_rates_setup(const void *fdt, int node)
 
 		rate_num++;
 	}
+
+	/* Ensure rates are in ascending order */
+	qsort_ul(opp_rates->rates, rate_num);
+
 	opp_rates->rate_num = rate_num;
 
 	return TEE_SUCCESS;
--- a/core/include/drivers/clk.h
+++ b/core/include/drivers/clk.h
@@ -68,7 +68,7 @@ struct clk_duty_cycle {
  * @get_parent: Get the current parent index of the clock
  * @set_rate: Set the clock rate
  * @get_rate: Get the clock rate
- * @get_rates_array: Get the supported clock rates as array
+ * @get_rates_array: Get the supported clock rates as array in ascending order
  * @get_rates_steps: Get support clock rates by min/max/step representation
  * @get_duty_cycle: Get duty cytcle of the clock
  *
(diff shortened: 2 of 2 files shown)


core/drivers/clk/sam/at91_cpu_opp.c
```diff
@@ -118,6 +118,10 @@ static TEE_Result opp_rates_setup(const void *fdt, int node)
 
 		rate_num++;
 	}
+
+	/* Ensure rates are increasing ordered */
```
jforissier, line 122: "in ascending order" is the commonly used term

core/include/drivers/clk.h
```diff
@@ -212,7 +212,7 @@ struct clk *clk_get_parent_by_index(struct clk *clk, size_t pidx);
 TEE_Result clk_set_parent(struct clk *clk, struct clk *parent);
 
 /**
- * clk_get_rates_array() - Get supported rates as an array
+ * clk_get_rates_array() - Get supported rates as an increasing frequency array
```
jforissier, line 215: Get supported rates in ascending order

core/include/drivers/clk.h
```diff
@@ -68,7 +68,7 @@ struct clk_duty_cycle {
  * @get_parent: Get the current parent index of the clock
  * @set_rate: Set the clock rate
  * @get_rate: Get the clock rate
- * @get_rates_array: Get the supported clock rates as array
+ * @get_rates_array: Get the supported clock rates as array in increasing order
```
jforissier, line 71: ascending

jforissier (commented):
Minor comments.

core/drivers/clk/sam/at91_cpu_opp.c
etienne-lms: ok, thanks.

core/include/drivers/clk.h
etienne-lms: +1

core/include/drivers/clk.h
etienne-lms: +1

etienne-lms:
Inline comments fixed and review tag applied.

PR #6954, opened 2024-07-20
RK3566 port

DaniilKl:
This port is a part of integration of safety features into [Zarhus project](https://github.com/zarhus), the main effort can be found in the following PR: https://github.com/zarhus/meta-zarhus-bsp-rockchip/pull/5, so the precise hardware and software configuration being used can be checked there. Though the project uses Yocto Project as a build environment, all the test binaries are built manually using several tools, so the configuration and changes in the PR may not be up to date.

---

Brief description of test environment.

Hardware:
* Radxa CM3 SoM based on RK3566 SoC;
* Radxa CM3 IO extention board.

Software:

I think this image explains software stack in the best way:

Rockchip_bootflow20181122

> Source: [`opensource.rock-chips.com`](https://opensource.rock-chips.com/wiki_Boot_option#Boot_introduce).

**The red** boot flow is being used for tests.

This is my first interaction with OPTEE and Trust Zone stuff, so I will be grateful for any help here!

### DIFF
--- a/core/arch/arm/plat-rockchip/conf.mk
+++ b/core/arch/arm/plat-rockchip/conf.mk
@@ -74,6 +74,29 @@ CFG_EARLY_CONSOLE_BAUDRATE ?= 1500000
 CFG_EARLY_CONSOLE_CLK_IN_HZ ?= 24000000
 endif
 
+ifeq ($(PLATFORM_FLAVOR),rk3566)
+include core/arch/arm/cpu/cortex-armv8-0.mk
+arm32-platform-cpuarch  := cortex-a55
+
+$(call force,CFG_TEE_CORE_NB_CORE,4)
+# cortex-a55 sopports GICv3 or GICv4. There is no support for GICv4 in OP-TEE,
+# so use GICv3:
+$(call force,CFG_ARM_GICV3,y)
+
+CFG_TZDRAM_START ?= 0x08400000
+CFG_TZDRAM_SIZE  ?= 0x02000000
+CFG_SHMEM_START  ?= 0x0A400000
+CFG_SHMEM_SIZE   ?= 0x00400000
+
+CFG_EARLY_CONSOLE ?= y
+CFG_EARLY_CONSOLE_BASE ?= UART2_BASE
+CFG_EARLY_CONSOLE_SIZE ?= UART2_SIZE
+CFG_EARLY_CONSOLE_BAUDRATE ?= 1500000
+CFG_EARLY_CONSOLE_CLK_IN_HZ ?= 24000000
+
+CFG_WITH_ARM_TRUSTED_FW ?= y
+endif
+
 ifeq ($(platform-flavor-armv8),1)
 $(call force,CFG_ARM64_core,y)
 $(call force,CFG_WITH_ARM_TRUSTED_FW,y)
--- a/core/arch/arm/plat-rockchip/main.c
+++ b/core/arch/arm/plat-rockchip/main.c
@@ -31,7 +31,11 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, GIC_BASE, GIC_SIZE);
 
 void boot_primary_init_intc(void)
 {
+	#ifdef PLATFORM_FLAVOR_rk3566
+	gic_init_v3(0, GICD_BASE, GICR_BASE);
+	#else
 	gic_init(GICC_BASE, GICD_BASE);
+	#endif
 }
 
 void boot_secondary_init_intc(void)
--- a/core/arch/arm/plat-rockchip/platform_config.h
+++ b/core/arch/arm/plat-rockchip/platform_config.h
@@ -67,6 +67,34 @@
 #define SGRF_BASE		(MMIO_BASE + 0x07330000)
 #define SGRF_SIZE		SIZE_K(64)
 
+#elif defined(PLATFORM_FLAVOR_rk3566)
+
+/* This information is from Technical Reference Manual for RK3568, which is said
+ * to be very close in configuration to RK3566. RTM for RK3566 has net been
+ * found. */
+
+#define MMIO_BASE		0xF0000000
+
+#define UART0_BASE		(MMIO_BASE + 0x0DD50000)
+#define UART0_SIZE		SIZE_K(64)
+
+#define UART1_BASE		(MMIO_BASE + 0x0E650000)
+#define UART1_SIZE		SIZE_K(64)
+
+#define UART2_BASE		(MMIO_BASE + 0x0E660000)
+#define UART2_SIZE		SIZE_K(64)
+
+#define UART3_BASE		(MMIO_BASE + 0x0E670000)
+#define UART3_SIZE		SIZE_K(64)
+
+#define SGRF_BASE		(MMIO_BASE + 0x0DC60000)
+#define SGRF_SIZE		SIZE_K(64)
+
+#define GIC_BASE		(MMIO_BASE + 0x0d400000)
+#define GIC_SIZE		SIZE_M(4)
+#define GICD_BASE		(GIC_BASE)
+#define GICR_BASE		(GIC_BASE + 0x60000)
+
 #elif defined(PLATFORM_FLAVOR_px30)
 
 #define GIC_BASE		0xff130000
--- a/core/arch/arm/plat-rockchip/platform_rk3566.c
+++ b/core/arch/arm/plat-rockchip/platform_rk3566.c
@@ -1,0 +1,44 @@
+#include <common.h>
+#include <io.h>
+#include <kernel/panic.h>
+#include <mm/core_memprot.h>
+#include <platform.h>
+#include <platform_config.h>
+
+#define SGRF_DDRRGN_CON0_16(n)		((n) * 4)
+#define SGRF_DDR_RGN_0_16_WMSK		GENMASK_32(11, 0)
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, SGRF_BASE, SGRF_SIZE);
+
+int platform_secure_ddr_region(int rgn, paddr_t st, size_t sz)
+{
+	vaddr_t sgrf_base = (vaddr_t)phys_to_virt_io(SGRF_BASE, SGRF_SIZE);
+	paddr_t ed = st + sz;
+	uint32_t st_mb = st / SIZE_M(1);
+	uint32_t ed_mb = ed / SIZE_M(1);
+
+	if (!sgrf_base)
+		panic();
+
+	assert(rgn <= 7);
+	assert(st < ed);
+
+	/* Check aligned 1MB */
+	assert(st % SIZE_M(1) == 0);
+	assert(ed % SIZE_M(1) == 0);
+
+	DMSG("protecting region %d: 0x%lx-0x%lx", rgn, st, ed);
+
+	/* Set ddr region addr start */
+	io_write32(sgrf_base + SGRF_DDRRGN_CON0_16(rgn),
+		   BITS_WITH_WMASK(st_mb, SGRF_DDR_RGN_0_16_WMSK, 0));
+
+	/* Set ddr region addr end */
+	io_write32(sgrf_base + SGRF_DDRRGN_CON0_16(rgn + 8),
+		   BITS_WITH_WMASK((ed_mb - 1), SGRF_DDR_RGN_0_16_WMSK, 0));
+
+	io_write32(sgrf_base + SGRF_DDRRGN_CON0_16(16),
+		   BIT_WITH_WMSK(rgn));
+
+	return 0;
+}
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -2542,10 +2542,19 @@ void *phys_to_virt(paddr_t pa, enum teecore_memtypes m, size_t len)
 		va = NULL;
 		break;
 	default:
+		if ( m == MEM_AREA_IO_SEC )
+			IMSG("Getting virtual address for physical address %lx", pa);
 		va = map_pa2va(find_map_by_type_and_pa(m, pa, len), pa, len);
+		if ( m ==  MEM_AREA_IO_SEC)
+			IMSG("The virtual address is %lx", pa);
 	}
-	if (m != MEM_AREA_SEC_RAM_OVERALL)
+	if (m != MEM_AREA_SEC_RAM_OVERALL){
+		if ( m == MEM_AREA_IO_SEC )
+			IMSG("Checking virtual address...");
 		check_va_matches_pa(pa, va);
+	}
+	if ( m == MEM_AREA_IO_SEC )
+		IMSG("Returning virtual address...");
 	return va;
 }
 
@@ -2566,9 +2575,12 @@ void *phys_to_virt_io(paddr_t pa, size_t len)
 
 vaddr_t core_mmu_get_va(paddr_t pa, enum teecore_memtypes type, size_t len)
 {
-	if (cpu_mmu_enabled())
+	if (cpu_mmu_enabled()){
+		IMSG("MMU is enabled, returning virtal address...");
 		return (vaddr_t)phys_to_virt(pa, type, len);
+	}
 
+	IMSG("MMU is not enbaled, returning real address");
 	return (vaddr_t)pa;
 }
 
--- a/core/arch/arm/plat-rockchip/sub.mk
+++ b/core/arch/arm/plat-rockchip/sub.mk
@@ -5,6 +5,7 @@ srcs-$(PLATFORM_FLAVOR_px30) += platform_px30.c
 srcs-$(PLATFORM_FLAVOR_rk322x) += platform_rk322x.c
 srcs-$(PLATFORM_FLAVOR_rk3399) += platform_rk3399.c
 srcs-$(PLATFORM_FLAVOR_rk3588) += platform_rk3588.c
+srcs-$(PLATFORM_FLAVOR_rk3566) += platform_rk3566.c
 
 ifeq ($(PLATFORM_FLAVOR),rk322x)
 srcs-y += plat_init.S


DaniilKl:
For now, I was able to boot the describett boot flow up to Linux userspace without any errors/warning from OPTEE side:
```shell
I/TC: OP-TEE version: 4.1.0-dev (gcc version 13.2.0 (GCC)) #1 Fri Jan 19 17:14:14 UTC 2024 aarch64
I/TC: Primary CPU switching to normal world boot
INFO:    BL31: BL32 initialization succeded!
INFO:    BL31: Preparing for EL3 exit to normal world
INFO:    Entry point address = 0xa00000
INFO:    SPSR = 0x3c9
```

### Further steps and problems to solve

#### Linking problem

After build I get an ELF with segments that seem to be ok for my configuration:

```shell
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000010000 0x0000000008400000 0x0000000008400000
                 0x000000000006ee98 0x00000000000ac400  RWE    0x10000
  LOAD           0x0000000000080000 0x0000000008600000 0x0000000008600000
                 0x0000000000005f98 0x0000000000005f98  RW     0x10000
  DYNAMIC        0x0000000000080000 0x0000000008600000 0x0000000008600000
                 0x0000000000000110 0x0000000000000110  RW     0x8
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x10
```

The ELF get linked with U-Boot and ATF into one image and then being booted. But during boot ATF cannot boot OPTEE (it gets blocked for several second on communicate `INFO:    BL31: Initializing BL32`, drops it, and continues to boot U-Boot). After some debugging I found out that SPL images gives to ATF `0x8600000` instead of `0x8400000` as an TEE entry point. After some digging I was able to relink the image manually so the entry was `0x840000` and boot it (the lopgs from the beginning). Still searching of what I have done wrong here.

#### Linux Kernel oops

After integrating self built OPTEE and ATF I got following oops during Linux Kernel boot:

```shell
[    0.493196] scmi_protocol scmi_dev.1: Enabled polling mode TX channel - prot_id:16
[    0.494329] arm-scmi firmware:scmi: SCMI Notifications - Core Enabled.
[    0.505441] mmc_host mmc1: Bus speed (slot 0) = 375000Hz (slot req 400000Hz, actual 375000HZ div = 0)
[    0.554984] ------------[ cut here ]------------
[    0.555414] WARNING: CPU: 2 PID: 1 at /drivers/firmware/arm_scmi/shmem.c:60 shmem_tx_prepare+0xe4/0x150
[    0.556277] Modules linked in:
[    0.556569] CPU: 2 PID: 1 Comm: swapper/0 Not tainted 6.6.23-yocto-standard-00118-g2d01bc1d4eea #1
[    0.557374] Hardware name: Radxa Compute Module 3(CM3) IO Board (DT)
[    0.557946] pstate: 20400009 (nzCv daif +PAN -UAO -TCO -DIT -SSBS BTYPE=--)
[    0.558577] pc : shmem_tx_prepare+0xe4/0x150
[    0.558971] lr : shmem_tx_prepare+0x124/0x150
[    0.559372] sp : ffff80008169b910
[    0.559676] x29: ffff80008169b910 x28: 0000000000000000 x27: ffff0000013ae7a8
[    0.560333] x26: ffff000001294080 x25: ffff8000815bc3d8 x24: 0000000000000000
[    0.560986] x23: 0000000000000000 x22: ffff800081875004 x21: ffff0000013a8a40
[    0.561638] x20: ffff800081875000 x19: ffff00000133d880 x18: 0000000000000006
[    0.562288] x17: 746f7270202d206c x16: 656e6e6168632058 x15: 076e076f07690774
[    0.562941] x14: 0761076307690766 x13: ffff8000814b0438 x12: 00000000000002dc
[    0.563592] x11: 0000000000000040 x10: ffff000001296138 x9 : ffff000001296130
[    0.564244] x8 : ffff0000009d7dc0 x7 : 0000000000000000 x6 : 0000000007d7939b
[    0.564896] x5 : 00ffffffffffffff x4 : 00206cc800324b00 x3 : 0000000000000018
[    0.565547] x2 : 0000000000000112 x1 : 0000000000000000 x0 : 0000000000000000
[    0.566198] Call trace:
[    0.566428]  shmem_tx_prepare+0xe4/0x150
[    0.566794]  smc_send_message+0x44/0xe4
[    0.567152]  do_xfer+0xb0/0x1b4
[    0.567451]  version_get+0x90/0xbc
[    0.567770]  scmi_base_protocol_init+0x5c/0x458
[    0.568190]  scmi_get_protocol_instance+0x160/0x28c
[    0.568639]  scmi_probe+0x2b0/0x720
[    0.568967]  platform_probe+0x68/0xc8
[    0.569315]  really_probe+0x148/0x2b0
[    0.569657]  __driver_probe_device+0x78/0x12c
[    0.570060]  driver_probe_device+0x3c/0x15c
[    0.570446]  __driver_attach+0x90/0x19c
[    0.570802]  bus_for_each_dev+0x7c/0xdc
[    0.571159]  driver_attach+0x24/0x30
[    0.571493]  bus_add_driver+0xe8/0x1f8
[    0.571841]  driver_register+0x5c/0x124
[    0.572198]  __platform_driver_register+0x28/0x34
[    0.572635]  scmi_driver_init+0x80/0xbc
[    0.572998]  do_one_initcall+0x80/0x1c8
[    0.573355]  kernel_init_freeable+0x1bc/0x284
[    0.573760]  kernel_init+0x24/0x1dc
[    0.574087]  ret_from_fork+0x10/0x20
[    0.574421] ---[ end trace 0000000000000000 ]---
[    0.574874] scmi_protocol scmi_dev.1: Timeout waiting for a free TX channel !
[    0.575544] arm-scmi firmware:scmi: unable to communicate with SCMI
[    0.576325] arm-scmi: probe of firmware:scmi failed with error -95
[    0.577366] SMCCC: SOC_ID: ARCH_SOC_ID not implemented, skipping ....
[    0.578660] usbcore: registered new interface driver usbhid
[    0.579173] usbhid: USB HID core driver
[    0.582295] hw perfevents: enabled with armv8_cortex_a55 PMU driver, 7 counters available
[    0.587587] NET: Registered PF_PACKET protocol family
[    0.588129] 9pnet: Installing 9P2000 support
[    0.588465] mmc_host mmc1: Bus speed (slot 0) = 50000000Hz (slot req 50000000Hz, actual 50000000HZ div = 0)
[    0.589551] mmc1: new high speed SDHC card at address aaaa
[    0.591110] mmcblk1: mmc1:aaaa SC16G 14.8 GiB
```

For now I do not know exactly what may be the cause, but, I suppose, it is OPTEE and memory initialization related stuff. The code in `core/arch/arm/plat-rockchip/platform_rk3566.c` was copied from `core/arch/arm/plat-rockchip/platform_rk3566.c` without changes, with assumption that both SoC's have the same `Armv8` architecture and therefore memory initialization will not need a lot of changes. This means that I do not fully understand the code there.

But, actually, the problem might as well has come from ATF, Because I am using latest `master` of `trusted-firmware-a` to build an ATF binary because only [has support for RK3566](https://github.com/ARM-software/arm-trusted-firmware/commit/9fd9f1d024872b440e3906eded28037330b6f422), and `master` builds may be unstable.

core/arch/arm/plat-rockchip/platform_rk3566.c
```diff
@@ -0,0 +1,44 @@
+#include <common.h>
+#include <io.h>
+#include <kernel/panic.h>
+#include <mm/core_memprot.h>
+#include <platform.h>
+#include <platform_config.h>
+
+#define SGRF_DDRRGN_CON0_16(n)		((n) * 4)
```
DaniilKl, line 8: And this one I do not understand. This seems to be some RK3399 registers writing, but I have not been able do understand what exactly is going on here so far.

DaniilKl: This seems to be a helper macro for DDR region registers as stated [here](https://gist.github.com/CrystalGamma/a68333fa4c9fda7eb6c09d30ad4937fe). And [there is](https://github.com/rockchip-linux/u-boot/blob/63c55618fbdc36333db4cf12f7d6a28f0a178017/arch/arm/include/asm/arch-rockchip/grf_rk3399.h#L274) a precise address of the SGRF registers for rk3399, but I have not been able to find these for rk3566 or rk3568 so far :cry: .

core/mm/core_mmu.c
```diff
@@ -2418,10 +2418,19 @@ void *phys_to_virt(paddr_t pa, enum teecore_memtypes m, size_t len)
 		va = NULL;
 		break;
 	default:
+		if ( m == MEM_AREA_IO_SEC )
```
jenswi-linaro, line 2421: Please remove the changes in this file, it looks like it's only debug code while you're testing things.

core/arch/arm/plat-rockchip/platform_rk3566.c
```diff
@@ -0,0 +1,44 @@
+#include <common.h>
```
jenswi-linaro, line 1: This file looks like a copy of `platform_rk3399.c`. I'd rather reuse the same code than use diverging copies that are poorly understood.

core/arch/arm/plat-rockchip/platform_config.h
```diff
@@ -66,6 +66,34 @@
 #define SGRF_BASE		(MMIO_BASE + 0x07330000)
 #define SGRF_SIZE		SIZE_K(64)
 
+#elif defined(PLATFORM_FLAVOR_rk3566)
+
+/* This information is from Technical Reference Manual for RK3568, which is said
```
jenswi-linaro, line 71:
``` C
/*
 * Please format
 * multiline comments
 * like this.
 */
```

core/arch/arm/plat-rockchip/main.c
```diff
@@ -24,7 +24,11 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, GIC_BASE, GIC_SIZE);
 
 void boot_primary_init_intc(void)
 {
+	#ifdef PLATFORM_FLAVOR_rk3566
```
jenswi-linaro, line 27: Please check for `GICR_BASE` or if `CFG_ARM_GICV3` is enabled.
You don't need to indent preprocessor directives.

core/arch/arm/plat-rockchip/conf.mk
```diff
@@ -56,6 +56,29 @@ CFG_SHMEM_SIZE   ?= 0x00400000
 CFG_EARLY_CONSOLE ?= n
 endif
 
+ifeq ($(PLATFORM_FLAVOR),rk3566)
+include core/arch/arm/cpu/cortex-armv8-0.mk
+arm32-platform-cpuarch  := cortex-a55
+
+$(call force,CFG_TEE_CORE_NB_CORE,4)
+# cortex-a55 sopports GICv3 or GICv4. There is no support for GICv4 in OP-TEE,
```
jenswi-linaro, line 64: What does the soc have?

jenswikl (commented):
This port is a bit shaky, obviously since it crashes in the kernel but also that it's based on unconfirmed assumptions.

core/mm/core_mmu.c
DaniilKl: I will do it when submitting this PR to the final review. Until then it will be in `draft state` and I will keep here some WIP or debug commits.

core/arch/arm/plat-rockchip/conf.mk
DaniilKl: According to [rk3568 TRF](https://opensource.rock-chips.com/images/2/26/Rockchip_RK3568_TRM_Part1_V1.3-20220930P.PDF) the SoC has GIC-600, which [implements GICv3](https://developer.arm.com/Processors/CoreLink%20GIC-600).

core/arch/arm/plat-rockchip/platform_rk3566.c
DaniilKl: Sure, but firstly I need to get `SGRF` info from Rockchip.

DaniilKl:
Currently this port is not ready for review (so, currently in `graft` status), the cause is: missing `SGRF` info for RK3566. I am currently waiting for a response from Rockchip, so, I hope I will get needed information soon.

DaniilKl:
Some rebasing on new `master`.

DaniilKl:
The communication with Rockchip is still ongoing.

core/arch/arm/plat-rockchip/platform_rk3566.c
DualTachyon: Unfortunately, RockChip will not provide any SGRF to anyone who is not a $$$ customer. I have a somewhat decent relationship with them, I even reported to them a critical security vulnerability with one of their SoCs and they still won't help much, even signing NDAs.

klogg:
Hey @DaniilKl, any news from Rockchip?

DaniilKl:
Hey @klogg, unfortunately no. And I am a bit pessimistic about it.

DaniilKl:
Some rebasing on latest `master`.

DaniilKl:
I was not able to get any inf. from Rockchip. if anyone else will get these registers - fell free to take over my changes.

jenswi-linaro:
> I was not able to get any inf. from Rockchip. if anyone else will get these registers - fell free to take over my changes.

That's too bad, thanks for trying.

core/arch/arm/plat-rockchip/platform_rk3566.c
ALTracer: > but I have not been able to find these for rk3566 or rk3568 so far

Have you looked in the same U-Boot repo and why not just try `grf_rk3568.h`? IIUC, RK3568/RK3566 are very similar,  and this file was published by Joseph Chen in 2020 https://github.com/rockchip-linux/u-boot/commit/7e26af3867078dd5ef0709f27ed7a74afa7aa6a2
I may not know about Aarch64 and secure firewalls much, but OP-TEE only needs to restrict access via DDR Regions? Does TF-A treat these two SoC differently? I am only interested in RK3568, is there another PR for it?

DaniilKl: > I may not know about Aarch64 and secure firewalls much, but OP-TEE only needs to restrict access via DDR Regions?

Yep, it uses MMU to do so, and there is no MMU registers in the file you linked. AFAIK, the needed registers are hidden in SGRF region, which is not available publicly.

> Does TF-A treat these two SoC differently? I am only interested in RK3568, is there another PR for it?

I suppose it does not. Because, if you take a look into Rockchips rkbin repository that contains binaries of TEE stack you will notice that there is no binaries for rk3566, only binaries for rk3568 are available. I have been using these binaries on rk4566 and have not met any problems so far.

DualTachyon: What you want is probably here: https://review.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/refs/heads/master/plat/rockchip/rk3568/drivers/soc/soc.c#55

it's a lot less info than other RK chips, you'd have to trial and error. 0xfe200080 would contain the enable/disable region bits (CON_REG), and 0xfe200000..40 (a guess) would have the region start+end.

DaniilKl: Thanks, will try to find some time for this.

DualTachyon: You could also dump all registers at 0xFE200000 up to 0xFE2001FC. This will show how many registers there really are, since some mirroring will be shown, as well as likely meanings of each registers (e.g. a bunch of same values in sequence = address  ranges, a single value by itself = control register).

klogg:
@jenswi-linaro @DaniilKl I have reverse engineered Rockhip's OP-TEE for 3566 and can provide some info if someone wants to collaborate taking this forward. I am still working on OTP and keys storage, but at least for DDR protection here are some registers:

```c
platform_secure_ddr_region() // called from the patched init_primary()
{
  // DDR Firewall mapping is not stored so not used anywhere else
  v7 = phys_to_virt_io(0xFE200000LL, 0x8400000, 0xE00000, v5, v6);// FIREWALL_DDR, 64kB
  *(_DWORD *)(v7 + 4) = 0x48F0420;              // FIREWALL_DDR_FW_DDR_RGN = rgn * 0x4 ---> rgn = 1
                                                // RG_MAP_SECURE(top, base) ((((top) - 1) << 16) | (base))
                                                // top = 490; base = 420
  *(_DWORD *)(v7 + 0x80) |= 2u;                 // FIREWALL_DDR_FW_DDR_CON_REG = 0x8
                                                // BIT(rgn) = 1 << rgn = 2
}
```
NOTE: This is a snippet, not the whole function just to demonstrate that registers for DDR region securing is extremely similar to some other RK SoC ;)

mmind:
just stumbled over that new comment here

@klogg oh nice ... I think the basic DDR protection for the optee area was the only really missing component for this.
So this is really great. Moving optee forward for the rk356x would be really great, so I'd be on board to help :-)

mmind:
@klogg I've played around with those registers a bit. Interestingly contrary to the other socs, they don't seem to work on MB sizes but instead on 128K steps, so setting 
- FW_DDR_RGN(0) to 0x00000, protects the area 0x0 - 0x020000-1
- FW_DDR_RGN(0) to 0x10000, protects the area 0x0 - 0x040000-1
- FW_DDR_RGN(0) to 0x10001, protects the area 0x020000 - 0x040000-1
which I guess makes sense, as the rk3568 supports a maximum of 8GB of memory anyway - which the 0xffff max value then fills up.

klogg:
@mmind hope you can make some use of it!

I had a break working on this project (business travels, vacation, firefighting at work) but plan to return to it soon. Most of the changes RK introduced to their OP-TEE relate to OTP management (keys, ...) plus SCMI related stuff that is already published somewhere. I am now mostly interested in OTP because I want to enable proper configurable secure boot without sticking to RK tools

mmind:
@klogg of definitly, I've started with implementing the missing memory protection for TF-A :-) .

Both the rk3588 and rk3576 submissions did contain the needed code, but the rk3568 TF-A never protected its own memory so far. Will try the OPTEE part after that :-) . The goal being trying that fTPM thing on my QNAP NAS :-) .

klogg:
@mmind as soon as you have OP-TEE - I am happy to review and test

klogg:
@mmind testing your A-TF patches with https://github.com/klogg/optee_os/commit/4bf09abb65476e2f64270113bf450d669ada09f5 on RK3566 Orange Pi CM4 - seems working well

Weird that Arm CE instructions, while declared as supported, do not work on RK356x...

DualTachyon:
> @mmind testing your A-TF patches with [klogg@4bf09ab](https://github.com/klogg/optee_os/commit/4bf09abb65476e2f64270113bf450d669ada09f5) on RK3566 Orange Pi CM4 - seems working well
> 
> Weird that Arm CE instructions, while declared as supported, do not work on RK356x...

RockChip hides certain ARM features behind hidden (unknown) registers in SGRF. RK3588 had the same problem. With trial and error I even found how to enable the GIC SFR interface. Example: https://github.com/DualTachyon/rk3588-svd/blob/main/def/security/grf/dsu_s.def#L91

Most likely the 3566 has a similar register. The bit needs toggling and the cores soft reset to enable.

klogg:
@DualTachyon thanks for the insight and pointer.
Do you have this code somewhere published?

DualTachyon:
No code that exists today. I looked at the official changes made by RK on the official TF-A Gerrit website, to get the clues, then I just ran a bruteforce bit toggle+reset in a loop to find what else was gated. Printing the ARM ID/features registers was enough because when disabled, the stuff like CE and GIC SFR appear non existent on said registers.

klogg:
I have completely parsed all SYS_SGRF registers, played with all bits there - no luck. Will try to understand what component enables the Arm CE and disasm it.
[SYS_SGRF.txt](https://github.com/user-attachments/files/22642029/SYS_SGRF.txt)

DualTachyon:
> I have completely parsed all SYS_SGRF registers, played with all bits there - no luck. Will try to understand what component enables the Arm CE and disasm it. [SYS_SGRF.txt](https://github.com/user-attachments/files/22642029/SYS_SGRF.txt)

You need to check DSU_SGRF if there is one. DSU is the ARM DynamiQ Unit, and on the 3588 it is where you can configure the cores .

klogg:
Unfortunately on RK356x there is no DSU GRF, its settings are scattered all around the place - PMU, CRU, ...

OOHehir:
Hello @DaniilKl @klogg @mmind — I realise this may be a little late but just in case..

I've been doing some work on the OP-TEE port for the RK3506B, not exactly your target but close. I've managed, after a fair bit of work, to do a complete bring-up and validated it. xtest §7.1 = 5895/5895, 0 failed, across 3 reboot-separated runs, including HW secure-RAM isolation via FW_DDR. I see Rockchip haven't been forthcoming so a few findings might help. I had an eye on up-streaming throughout, so I worked on this without reverse-engineering the vendor blob — which should make the findings safe for you to reuse:

1. The FW_DDR/SGRF register map is in the public GPL U-Boot, not just the blob.

`mach-rockchip/rk3506/rk3506.c` (+ `cru_rk3506.h`) gives the bases directly — on rk3506 e.g. `FIREWALL_DDR_BASE = 0xff5f0000`, `FW_DDR_MST1 = 0x24`, `SGRF_BASE = 0xff210000`. The region/control encodings (`RGN(i) = i*4`, `CON = 0x40`, the per-master `ACC` regs, and the region field encoding) match the BSD-2 OP-TEE `platform_px30.c` / `platform_rk3588.c` already in-tree. Addresses are rk3506-specific, but the equivalent public GPL U-Boot board file exists for rk356x — that may get you the map you were waiting on Rockchip for.

2. I can confirm @mmind's 128 K observation regarding the FW_DDR.

On the rk3506 I measured a 128 KB region granule (vs rk3588's 1 MB) on a LuckFox Lyra Ultra — matching the rk3568 "128 K steps" finding. The slot value packs a 128 KB-granule base+size; the vendor's `RGN0 = 0x20000` -> protects `[0, 384 KB)`. One caveat specific to my setup: the protected region has to sit in low DRAM — a region at the old TEE-high base (`0x18000000`) fell outside the firewall's effective coverage, so I relocated `TEE_RAM` low. The base itself is programmable in 128 KB units (consistent with @mmind's non-zero-base result above); I just anchor near 0 — at `0x1000` rather than literal `0x0`, since `0x0` collides with the FIT load address and won't boot. With that in place I managed to allocate 16 MB, verified by NS `devmem` external-aborting across the range (4K/2M/8M/12M/15M) while 17M+ reads fine.

3. SGRF: you might not need the vendor per-slave table at all.

The one part of my port that still looks vendor-shaped is the SGRF slave-security table — knowing which peripherals to mark secure vs non-secure. But I found you can largely sidestep that: the `SLAVE_ALL_NS` idiom from the existing BSD-2 `rk322x` port (a `0xffff0000` masked write that marks all slaves non-secure) works as a clean substitute, and it only needs two public facts — the SGRF base (GPL U-Boot) and the masked-write semantics (already in the in-tree `rk322x` platform). No Rockchip-supplied table required, which is what got me past the SGRF wait. To open NS access for the console UART, that's all it took.

Please note: all-NS is deliberately permissive — no peripheral is hardware-protected from the normal world, so it's not fully secure. The boundary that actually matters for OP-TEE is still the FW_DDR firewall on the TEE RAM (points 1–2) plus the CPU secure/NS state, so it boots and runs cleanly. For a hardened product you'd come back and re-secure the sensitive slaves (crypto, OTP/eFuse, secure timer). For getting OP-TEE up, though, all-NS is enough. In my own port I only kept two slaves secure to match the vendor — they're the only vendor-config values left, and they could just as easily be dropped for pure `SLAVE_ALL_NS`.

I've released Yocto images for two of the LuckFox boards & am working on cleaning up the OP_TEE for upstream & prepping a BitBake recipe. Hopefully ready soon.

Hope this helps!

(corrected base address comment)

klogg:
Hey @OOHehir, really great progress!
On 356x I have latest OP-TEE 4.10 working with all sdram protections, TRNG, Crypto and OTP management + MCU for RT stuff. Too busy to upstream :( now researching RK closed binaries for more info on keyladder/keyreader and trying to understand how to do verified/measured boot with fully encrypted boot chain. Any info welcome!!!

EDIT: Only thing I am still using from RK repos and not upstream is the TPL with DDR init - got sdram control regs dump but it's not on a critical path.

OOHehir:
@klogg Great to hear 4.10 is working on 356x — that matches where we've landed on RK3506B (triple Cortex-A7), and the TPL/DDR-init situation is identical for us: it's the only RK blob left in our pre-OP-TEE chain. I'm working towards CRA on the RK306 but I think this can be justified.

On verified boot, here's what I mapped using public GPL U-Boot + the in-tree DT, in case it saves you time on 356x (it should be similar):

- Fuse-state oracle, no SWD/OTP poke needed: the SPL prints `## Verified-boot: N` every boot. `arch/arm/mach-rockchip/fit_misc.c` (`fit_board_verify_required_sigs`) reads the secure-boot enable byte from OTP and prints `(byte == 0xff)`. `0` means the enable fuse is not blown (virgin part); you don't need to read OTP directly to know the secure-boot state.

- OTP layout (`include/configs/rk3506_common.h`, public GPL): enable byte @ `0x20`, root RSA pubkey SHA-256 @ `0x180` (32 B), U-Boot anti-rollback @ `0x350`. Trust-root reads go through the secure-OTP device `secure-otp@ff520000` (`rk3506-u-boot.dtsi`: `reg=<0xff520000 0x8000>`, `mask_addr=<0xff528000>`, `secure_conf=<0xff210100>`) — not the NS NVMEM `otp@ff4f0000`.

- Enforcement gate: a FIT signature is only enforced when the SPL is built with `CONFIG_SPL_FIT_SIGNATURE=y`. Without it the FIT is checked for sha256 integrity only (origin not verified). The crypto deps (`CONFIG_RSA`/`SPL_RSA`/`FIT_HW_CRYPTO`/`SPL_ROCKCHIP_SECURE_OTP`) are already `=y` in `rk3506_defconfig`.

- Brick interlock (worth hard-coding in any provisioning script): `common/spl/spl_fit.c` calls `hang()` ("Verified-boot requires CONFIG_SPL_FIT_SIGNATURE enabled") if the OTP enable bit is set while the SPL lacks `SPL_FIT_SIGNATURE`. So `SPL_FIT_SIGNATURE=y` is an absolute prerequisite to burning the enable fuse — fuse-before-config bricks the unit.

- Burn logic (for when you do fuse): `lib/rsa/rsa-verify.c` (`rsa_burn_key_hash()`) writes the 32-B pubkey hash to OTP `0x180` and the enable byte to `0x20`. Rockchip's `rk_sign_tool sl/sb` signs the loader/idblock and `scripts/fit-sign.sh` signs the FIT — a closed DDR blob inside a signed idblock is still authenticated as a unit.

I've now validated the un-fused half of this on the Lyra board: SPL rebuilt with `CONFIG_SPL_FIT_SIGNATURE=y`, FIT signed (RSA-2048, dev key) with the pubkey embedded in the SPL control DTB, idbloader repacked, and the signed image boots to userspace. Cross-checked that a wrong-key-signed FIT fails verification against the embedded key, so the enforcement path is real. We're deliberately keeping the irreversible OTP burns for production units only.

On keyladder / keyreader / encrypted boot — I haven't looked at them I'm afraid; they live in RK closed binaries and our project's whole point is CRA-auditable clean-room provenance, so we don't RE them. One thing that might help your RE though: https://github.com/OOHehir/ghidra-bringup . It has `MmioXref` (find every function that touches a given MMIO window — handy for locating keyladder/crypto register sequences) and `SysregSmc` (CP15 + SMC call sites). It's tuned for ARMv7-LE (RK3506 is Cortex-A7); for rk356x (ARMv8) the `MmioXref` pass and the overall flow port directly, but the CP15/SMC detector would need an aarch64 variant. If you find a public-source path to the keyladder I'd be very interested.

Re the earlier thread point: we independently confirmed the 128 KB FW_DDR granule on 3506B silicon (corroborates mmind's rk356x 128K-step finding), and that the FW_DDR/SGRF register map is recoverable from public GPL U-Boot (`mach-rockchip/rk3506.c`, the `secure_conf` DT node) rather than only from the blob.

Hope this is of some use!

PR #6956, opened 2024-07-22
plat-stm32mp1: fix use after free in PMIC driver

etienne-lms:
Fix PMIC regulator levels arrays handling that missed a pointer reset after the buffer is freed. At runtime, pmic_list_voltages() handler function uses that reference and is expected to allocate back the buffer in case non-secure world requests voltage enumeration for the related regulator.

Fixes: a7990eb02b82 ("plat-stm32mp1: set voltage list at pmic driver init")


jforissier (commented):

PR #6957, opened 2024-07-23
core: riscv: Translate to PA when allocating PGT with MMU enabled

gagachang:
If MMU is enabled, `core_mmu_pgt_alloc()` returns virtual address of PGT instead of physical address.
Thus, it leads to some errors when we invoke `pa_to_ppn()` with returned PGT which is actually the virtual address of that PGT.

Fix it by checking whether MMU is enabled or not.
If MMU is enabled, we translate the returned PGT to its physical address.


gagachang:
Hi @maroueneboubakri, would you like to review this small patch?

maroueneboubakri:
Thanks for this patch @gagachang 

PR #6963, opened 2024-07-25
core: drivers: riscv_zkr_rng: limit seed reading time

lyctw:
```
If the attempts to read SEED exceed 1 second for 16-bit
randomness, we consider it a failure.

Also, move seed CSR field encodings to the header file.

Signed-off-by: Yu Chien Peter Lin <peterlin@andestech.com>
Tested-by: Alvin Chang <alvinga@andestech.com>
```


core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,65 @@
...
-			return TEE_ERROR_BAD_STATE;
+			return false;
+		case SEED_OPST_BIST:
+		case SEED_OPST_WAIT:
 		default:
-			break; /* can't happen */
+			retry_count++;
+			continue;
+		}
+	} while (retry_count <= CFG_SEED_RETRY_COUNT_MAX);
```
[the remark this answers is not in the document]
lyctw: Hi @GseoC,

Thank you for your insights. While a timeout might still fall short in generating sufficient entropy, in that case, increasing RNG_TIMEOUT_US is still needed, right? I would like to keep using retry counter and allow platforms to decide `CFG_SEED_RETRY_COUNT_MAX` based on their use case and quality of entropy source. 

I agreed that the patch still has a bandwidth issue and should add a `cpu_relax()` before each retry iteration, , similar to the Linux kernel's approach:  https://github.com/torvalds/linux/blob/master/arch/riscv/include/asm/archrandom.h#L44

core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,65 @@
...
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	uint8_t *ptr = buf;
+	uint16_t seed = 0;
+
+	while (len > 0) {
+		if (!seed_get_random_u16(&seed))
+			return TEE_ERROR_ACCESS_DENIED;
+		*ptr++ = seed & 0xff;
```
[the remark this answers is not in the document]
lyctw: The scalar crypto spec does not explicitly rule out the case AFAIK, so it may be valid.

core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,66 @@
...
-		 * The seed register must be accessed using CSR read-write
-		 * instructions. The write operation is ignored and serves
-		 * to indicate polling and flushing.
+		 * The seed register must be accessed using CSR
+		 * read-write instructions. The write operation
+		 * is ignored and serves to indicate polling and
+		 * flushing.
 		 */
-		val = swap_csr(CSR_SEED, val);
+		seed = swap_csr(CSR_SEED, 0);
```
jenswi-linaro, line 26: Is this a per CPU/Hart register or does it need a lock to serialize accesses?

core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,66 @@
...
+		case SEED_OPST_DEAD:
 			/* Unrecoverable self-test error */
-			return TEE_ERROR_BAD_STATE;
+			return false;
+		case SEED_OPST_BIST:
+		case SEED_OPST_WAIT:
 		default:
-			break; /* can't happen */
+			retry_count++;
+			cpu_relax();
```
jenswi-linaro, line 40: What do you expect `cpu_relax()` to do?

core/drivers/riscv_zkr_rng.c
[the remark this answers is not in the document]
jenswi-linaro: Using a timeout value makes more sense to me than a retry count since the process of gathering entropy most likely is timebound. I guess the purpose of the timeout is to catch the unlikely case where we can't get any random number and need a way to abort the operation. Depending on where this happens it will most likely mean panicking a TA. So the number should be large enough that it can't happen during normal operations. Trying with different values until it doesn't occur seems like a fragile approach.

core/arch/riscv/include/riscv.h
```diff
@@ -155,6 +155,16 @@ static inline __noprof void wfi(void)
...
 }
 
+static inline __noprof void cpu_relax(void)
+{
+#ifdef __riscv_muldiv
+	unsigned long dummy = 0;
+
+	asm volatile ("div %0, %0, zero" : "=r" (dummy));
+#endif
+	asm volatile ("" : : : "memory");
```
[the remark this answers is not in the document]
jenswi-linaro: What is the expected outcome?

core/drivers/riscv_zkr_rng.c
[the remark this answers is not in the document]
lyctw: Sure, I will switch to the timeout method. Since there is not actual hardware I can test with, I will assume a 1-second fixed timeout. We may still need something like `CFG_SEED_TIMEOUT` for platforms with varying entropy bandwidth.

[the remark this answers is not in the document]
lyctw: > Could be yes, but the hardware block may have a duration (dependent of its clock) above which the entropy should have been generated in the worst case scenario. Is it the case? If you use a timeout, I suggest IO_READ32_POLL_TIMEOUT().

Thank you for the suggestion! However, in this case, the seed is a register, so it can't be accessed with `IO_READ32_POLL_TIMEOUT()`

core/drivers/riscv_zkr_rng.c
lyctw: it is per CPU register, thus no need to protect with locks.

core/drivers/riscv_zkr_rng.c
lyctw: Wait for multiple CPU cycles; we may also use the `pause` instruction [1] to reduce power consumption.

[1] https://github.com/riscv/riscv-isa-manual/blob/98a92230a0d4fa7b5a6e297c32177f9177afd674/src/zihintpause.adoc#L4

core/arch/riscv/include/riscv.h
```diff
@@ -155,6 +155,25 @@ static inline __noprof void wfi(void)
...
+	 * for multiple CPU cycles.
+	 */
+	asm volatile ("div %0, %0, zero" : "=r" (dummy));
+
+	/*
+	 * Use the encoding of the 'pause' instruction,
+	 * thus no need to verify toolchain support for
+	 * zihintpause.
+	 */
+	asm volatile (".4byte 0x100000f"); /* pause */
```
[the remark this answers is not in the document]
lyctw: > Should this not be supported, I guess it defaults to a NOP code?

Yes, it is [NOP](https://github.com/riscv/riscv-isa-manual/blob/04179addb9725a0b253689ce59235041161ea832/src/mm-eplan.adoc?plain=1#L656-L657) if hardware doesn't support zihintpause.

> BTW, I could not find what "f" stands for in the MISC-MEM(opcode) field. I found the match here
https://github.com/riscv/riscv-test-env/blob/master/encoding.h

"f" (0b0001111) can be found on the opcode map in "[Volume I: RISC-V Unprivileged ISA V20191214-draft](https://github.com/riscv/riscv-isa-manual/blob/04179addb9725a0b253689ce59235041161ea832/src/rv-32-64g.adoc#L8)":
image

@GseoC Thanks for the review, would you like to give a tag?

core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,64 @@
...
 #include <tee/tee_cryp_utl.h>
 
-#define OPST_BIST 0b00
-#define OPST_WAIT 0b01
-#define OPST_ES16 0b10
-#define OPST_DEAD 0b11
+#define RNG_TIMEOUT_US	1000000
 
-TEE_Result hw_get_random_bytes(void *buf, size_t len)
+static inline bool __must_check seed_get_random_u16(uint16_t *val)
```
etienne-lms, line 16: Prefer no `inline` attribute in source file, only in header files.

core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,64 @@
...
-			return TEE_ERROR_BAD_STATE;
+			return false;
+		case SEED_OPST_BIST:
+		case SEED_OPST_WAIT:
 		default:
-			break; /* can't happen */
+			cpu_relax();
+			continue;
+		}
+	} while (!timeout_elapsed(timeout));
```
etienne-lms, line 45: In a thread context, this timeout can be triggered due to thread rescheduling.

GseoC: He's right, maybe the best thing would be to put after the loop:
`
seed = swap_csr(CSR_SEED, 0);
opst = seed & SEED_OPST;
if (opst != SEED_OPST_ES16) {
        EMSG("Failed to produce a sufficient amount of entropy");
        return false;
}

*val = seed & SEED_ENTROPY

return true;
`

what do you think?

gagachang: > In a thread context, this timeout can be triggered due to thread rescheduling.

Do you mean the foreign interrupts ?
More detail please.

core/drivers/riscv_zkr_rng.c
```diff
@@ -4,46 +4,64 @@
...
+
+TEE_Result hw_get_random_bytes(void *buf, size_t len)
+{
+	uint8_t *ptr = buf;
+	uint16_t seed = 0;
+
+	while (len > 0) {
+		if (!seed_get_random_u16(&seed))
+			return TEE_ERROR_ACCESS_DENIED;
+		*ptr++ = seed & 0xff;
```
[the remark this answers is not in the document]
gagachang: > Actually, don't you skip the first 8 bytes of the buffer here? As ++ happens before the assignation.

After checking generated assembly code, I can confirm that the assignment (store instruction) is executed on correct address instead of (address + 1).
Do you concern that different compilers might generate different program order ?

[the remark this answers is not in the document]
GseoC: First, I swapped "bytes" for "bits", then I missed that the pointer is incremented after the assignation so please, discard the first part of my comment :)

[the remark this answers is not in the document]
gagachang: Got it! no problem :)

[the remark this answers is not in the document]
lyctw: > can you proceed by adding commits [Review] <commit title> that can be squashed after the tag is given? This ensure that no other modifications are made to the original commit and simplifies the review :)

OK, will do.

core/drivers/riscv_zkr_rng.c
lyctw: Hi All,

Here is a problem, we are using mtime as generic timer which is shared between REE and TEE. Given 1-second timeout, If foreign interrupts occur, the time available for the while loop will definitely be less than 1 second, right? Even worse, if it returns to TEE after more than 1 second, it will directly timeout and no change to retry the seed.

Using a retry counter still seems necessary? Any thoughts?

etienne-lms: Yes, this is related to foreign interrupts. A TEE thread can be interrupted by a foreign interrupt. When so, the TEE thread is suspended and CPU goes back to normal world (let's say to the Linux kernel). The TEE thread is resumed when the Linux kernel schedules back the thread related to that TEE thread.
When `timeout_elapsed()` function is called from an interruptible TEE thread context and it returns `true`, caller of that function should be aware that the condition he/she expects may be already reached and that the timeout out was triggered only because the Linux kernel scheduler scheduled some other thread(s) before resuming the thread related to that TEE thread.
For example this a why helper macro `IO_READ32_POOL_TIMEOUT()` was implemented.

I was to give some example based on the current implementation of this P-R, but I see that actually the `while()` instruction at line 45 is never reached. Indeed, the block either exits from `return true` (line 35) / `return false` (line 38) or spins back from `continue` (line 43)` even before we reach line 45. Therefore my suggestion (feel free to implement something else):
```patch
	do {
		/*
		 * The seed register must be accessed using CSR
		 * read-write instructions. The write operation
		 * is ignored and serves to indicate polling and
		 * flushing.
		 */
		seed = swap_csr(CSR_SEED, 0);

		opst = seed & SEED_OPST;

		switch (opst) {
		case SEED_OPST_ES16:
			*val = seed & SEED_ENTROPY;
			return true;
		case SEED_OPST_DEAD:
			/* Unrecoverable self-test error */
			return false;
		case SEED_OPST_BIST:
		case SEED_OPST_WAIT:
		default:
			cpu_relax();
-			continue;
		}
	} while (!timeout_elapsed(timeout))
+
+	/* Consider timeout case due to normal world scheduler */
+	seed = swap_csr(CSR_SEED, 0);
+	if (seed & SEED_OPST == SEED_OPST_ES16) {
+		*val = seed & SEED_ENTROPY;
+		return true;
+	}

	return false;
 }
```

Note: `cpu_relax()` is not implemented in OP-TEE. If there a cpu-pause instruction for RiscV, i'd rather suggest to name the helper fonction something like `riscv_cpu_pause()` to avoid confusion.

GseoC: Yes @lyctw, this is why we suggested to the direct timeout case. It is also the case for STM32 platforms and we proceed the same way.

@etienne-lms 's implementation is looking good to me. Maybe the comment can be rephrased. You can also keep your error message to explicit what failed.

lyctw: Hi @etienne-lms 

> I see that actually the while() instruction at line 45 is never reached

Nice catch! Will fix.

lyctw:
```
	} while (!timeout_elapsed(timeout))
+
+	/* Consider timeout case due to normal world scheduler */
+	seed = swap_csr(CSR_SEED, 0);
+	if (seed & SEED_OPST == SEED_OPST_ES16) {
+		*val = seed & SEED_ENTROPY;
+		return true;
+	}
```

this change is equal to:

```
	} while (!timeout_elapsed(timeout))
	
+       retry_count = 1;
+       do {
+           // read seed
+       } while(retry_count--);
```

If the last retry fails, it returns false right away, regardless of the OPST state, which may still allow the next retry.
I am leaning towards a retry counter as I originally implemented, as the seed and timer are shared. We simply cannot use timeout and `cpu_relax()` should've addressed bandwidth issue. 

I hope that makes sense.

GseoC: I still think the timeout is more relevant as long as gathering the entropy is based on time :) If the timeout value expires, you read the seed state and that there's still no entropy available, you should expect to fail.

Based on the description, the RISCV pause instruction you suggested seemed appropriate for a cpu_relax()-like function. It can even be implemented in RISCV arch helpers IMO.

GseoC:

lyctw:
Tag applied, thanks @GseoC !

core/arch/riscv/include/riscv.h
```diff
@@ -155,6 +155,27 @@ static inline __noprof void wfi(void)
...
 
+static inline __noprof void riscv_cpu_pause(void)
+{
+	unsigned long dummy = 0;
+
+	/*
+	 * Use a divide instruction to force wait
+	 * for multiple CPU cycles.
+	 */
+	asm volatile ("div %0, %0, zero" : "=r" (dummy));
```
jforissier, line 166: I was somewhat surprised by this line until I read that "RISC-V doesn’t raise an exception on divide by zero". Perhaps worth adding as a comment here?

PR #6965, opened 2024-07-26
Mem map cleanup

jenswi-linaro:

### DIFF
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -609,26 +609,27 @@ static bool core_mmu_entry_copy(struct core_mmu_table_info *tbl_info,
 }
 
 static void core_init_mmu_prtn_tee(struct mmu_partition *prtn,
-				   struct tee_mmap_region *mm)
+				   struct memory_map *mem_map)
 {
-	size_t n;
+	size_t n = 0;
 
-	assert(prtn && mm);
+	assert(prtn && mem_map);
 
-	for (n = 0; !core_mmap_is_end_of_table(mm + n); n++) {
+	for (n = 0; n < mem_map->count; n++) {
+		struct tee_mmap_region *mm = mem_map->map + n;
 		debug_print(" %010" PRIxVA " %010" PRIxPA " %10zx %x",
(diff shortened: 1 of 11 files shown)


core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -609,26 +609,27 @@ static bool core_mmu_entry_copy(struct core_mmu_table_info *tbl_info,
...
-				   struct tee_mmap_region *mm)
+				   struct memory_map *mem_map)
 {
-	size_t n;
+	size_t n = 0;
 
-	assert(prtn && mm);
+	assert(prtn && mem_map);
 
-	for (n = 0; !core_mmap_is_end_of_table(mm + n); n++) {
```
jforissier, line 618: `core_mmap_is_end_of_table()` is now unused and should be removed from `core/include/mm/core_mmu.h`, and with that `MEM_AREA_END` should be removed from `enum teecore_memtypes` as well.

jforissier (commented):
- For "libutils: add array helpers": two minor comments, in any case:
- For "core: mm: introduce struct memory_map":
   With my comment addressed: 
- For " core: add_phys_mem(): fix mergeable physical memory":

lib/libutils/ext/include/string_ext.h
```diff
@@ -52,4 +52,66 @@ char *nex_strdup(const char *s);
...
+ * @base:        start of the array
+ * @elem_count:  the number of elements in the array
+ * @elem_size:   the size of each element in the array
+ * @pos:         element position counted in units of @elem_size
+ *
+ * Removes the element at @pos by advancing the element after this position
+ * to fill the space. The now unused element at the end of the array is
+ * zero-initialized.
+ */
+void rem_array_elem_pad_zero(void *base, size_t elem_count, size_t elem_size,
```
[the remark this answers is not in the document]
jenswi-linaro: OK, I'll update.

core/arch/arm/mm/core_mmu_lpae.c
jenswi-linaro: Good point, I'll update.

jenswi-linaro:
Updated and tags applied.

PR #6966, opened 2024-07-28
Add imx95 support

sahilnxp:
This pull request adds support for i.MX95evk platform


jforissier (commented):
One comment below. Plus, I believe i.MX95 should be written i.MX95 not imx95 when it appears in "normal" text (commit description, commit subject, comments). Reserve imx95 for subject prefix (`core: imx95: do x and y...`). So for instance:
```
core: imx: add i.MX95 SoC ID

Add i.MX95 SoC ID

Signed-off-by: ...
```
... instead of:
```
core: imx: add imx95 SoC ID

add imx95 SoC ID

Signed-off-by: ...
```

With that, for all commits:

sahilnxp:
Thanks @jforissier for review.
Comments addressed and tags updated.

PR #6974, opened 2024-08-01
Remove build time paths from version.o, use CFLAGS from build environment

mikkorapeli-linaro:
Default debug flag -g3 was already getting removed from version-o-cflags but on corstone1000 platform-cflags-debug-info is set to -gdwarf-2 and thus debug info was included in the generated version.o. This then included full absolute build time path which breaks "buildpaths" yocto QA check since the path is also in final tee.elf binary. Remove all debug flags from version-o-cflags as a fix/workaround.

Fixes yocto build reproducibility issues on corstone1000:

https://gitlab.com/jonmason00/meta-arm/-/jobs/7472950196

Cc: Jon Mason <jon.mason@arm.com>

### DIFF
--- a/core/arch/arm/kernel/link.mk
+++ b/core/arch/arm/kernel/link.mk
@@ -151,7 +151,7 @@ define update-buildcount
 endef
 
 # filter-out to workaround objdump warning
-version-o-cflags = $(filter-out -g3,$(core-platform-cflags) \
+version-o-cflags = $(filter-out -g3,$(CFLAGS) $(core-platform-cflags) \
 			$(platform-cflags) $(cflagscore))
 # SOURCE_DATE_EPOCH defined for reproducible builds
 ifneq ($(SOURCE_DATE_EPOCH),)
--- a/core/arch/arm/kernel/link.mk
+++ b/core/arch/arm/kernel/link.mk
@@ -163,14 +163,15 @@ CORE_CC_VERSION = `$(CCcore) -v 2>&1 | grep "version " | sed 's/ *$$//'`
 define gen-version-o
 	$(call update-buildcount,$(link-out-dir)/.buildcount)
 	@$(cmd-echo-silent) '  GEN     $(link-out-dir)/version.o'
-	$(q)echo -e "const char core_v_str[] =" \
+	$(q)cd $(link-out-dir) && \
+		echo -e "const char core_v_str[] =" \
 		"\"$(TEE_IMPL_VERSION) \"" \
 		"\"($(CORE_CC_VERSION)) \"" \
 		"\"#$(BUILD_COUNT_STR) \"" \
 		"\"$(DATE_STR) \"" \
 		"\"$(CFG_KERN_LINKER_ARCH)\";\n" \
 		| $(CCcore) $(version-o-cflags) \
-			-xc - -c -o $(link-out-dir)/version.o
+			-xc - -c -o version.o
 endef
 $(link-out-dir)/version.o:
 	$(call gen-version-o)
--- a/core/arch/riscv/kernel/link.mk
+++ b/core/arch/riscv/kernel/link.mk
@@ -74,14 +74,15 @@ CORE_CC_VERSION = `$(CCcore) -v 2>&1 | grep "version " | sed 's/ *$$//'`
 define gen-version-o
 	$(call update-buildcount,$(link-out-dir)/.buildcount)
 	@$(cmd-echo-silent) '  GEN     $(link-out-dir)/version.o'
-	$(q)echo -e "const char core_v_str[] =" \
+	$(q)cd $(link-out-dir) && \
+		echo -e "const char core_v_str[] =" \
 		"\"$(TEE_IMPL_VERSION) \"" \
 		"\"($(CORE_CC_VERSION)) \"" \
 		"\"#$(BUILD_COUNT_STR) \"" \
 		"\"$(DATE_STR) \"" \
 		"\"$(CFG_KERN_LINKER_ARCH)\";\n" \
 		| $(CCcore) $(version-o-cflags) \
-			-xc - -c -o $(link-out-dir)/version.o
+			-xc - -c -o version.o
 endef
 
 $(link-out-dir)/version.o:
--- a/core/arch/riscv/kernel/link.mk
+++ b/core/arch/riscv/kernel/link.mk
@@ -62,7 +62,7 @@ define update-buildcount
 endef
 
 # filter-out to workaround objdump warning
-version-o-cflags = $(filter-out -g3,$(core-platform-cflags) \
+version-o-cflags = $(filter-out -g3,$(CFLAGS) $(core-platform-cflags) \
 			$(platform-cflags) $(cflagscore))
 # SOURCE_DATE_EPOCH defined for reproducible builds
 ifneq ($(SOURCE_DATE_EPOCH),)
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -80,7 +80,8 @@ comp-compiler-$2 := $$(CC$(sm))
 comp-flags-$2 = $$(filter-out $$(CFLAGS_REMOVE) $$(cflags-remove) \
 			      $$(cflags-remove-$$(comp-sm-$2)) \
 			      $$(cflags-remove-$2), \
-		   $$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) $$(CFLAGS_WARNS) \
+		   $$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) $$(CFLAGS) \
+		   $$(CFLAGS_WARNS) \
 		   $$(comp-cflags$$(comp-sm-$2)) $$(cflags$$(comp-sm-$2)) \
 		   $$(cflags-lib$$(comp-lib-$2)) $$(cflags-$2))
 ifeq ($C,1)


core/arch/arm/kernel/link.mk
```diff
@@ -150,8 +150,10 @@ define update-buildcount
 	fi
 endef
 
-# filter-out to workaround objdump warning
-version-o-cflags = $(filter-out -g3,$(core-platform-cflags) \
+# filter-out to workaround objdump warning and to
+# remove build time paths
+version-o-cflags = $(filter-out \
+	$(platform-cflags-debug-info),$(core-platform-cflags) \
```
jforissier, line 156: IIUC the build reproducibility is desired for non-debug builds only, right? Because, enabling debug info is likely to insert non-reproducible information such as as build time or full paths or whatever. If that is true then I believe the proper fix would be:
```diff
diff --git a/core/arch/arm/plat-corstone1000/conf.mk b/core/arch/arm/plat-corstone1000/conf.mk
index 98347b143..3c0bd5756 100644
--- a/core/arch/arm/plat-corstone1000/conf.mk
+++ b/core/arch/arm/plat-corstone1000/conf.mk
@@ -1,8 +1,10 @@
 PLATFORM_FLAVOR ?= mps3

 # ARM debugger needs this
+ifeq ($(CFG_DEBUG_INFO),y)
 platform-cflags-debug-info = -gdwarf-2
 platform-aflags-debug-info = -gdwarf-2
+endif

 $(call force,CFG_HWSUPP_MEM_PERM_WXN,y)
 $(call force,CFG_HWSUPP_MEM_PERM_PXN,y)
```
What do you think? Can you try that please?

mikkorapeli-linaro: Without this patch, version.o on corstone1000 has:

```
strings ../build/core/version.o 
4.2.0-1-ge6d7dc0aa-dev (gcc version 14.1.0 (GCC)) #1 Fri Apr 12 09:51:21 UTC 2024 aarch64
<stdin>
/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/git
<stdin>
core_v_str
long unsigned int
GNU C17 14.1.0 -mgeneral-regs-only -mcpu=cortex-a35 -mstrict-align -mno-outline-atomics -mgeneral-regs-only -mlittle-endian -mabi=lp64 -gdwarf-2 -gdwarf-2 -Os -Os -ffunction-sections -fdata-sections -fpie -fstack-protector-strong
char
GCC: (GNU) 14.1.0
<stdin>
core_v_str
.symtab
.strtab
.shstrtab
.text
.data
.bss
.rodata.core_v_str
.rela.debug_info
.debug_abbrev
.rela.debug_aranges
.debug_line
.debug_str
.comment
.note.GNU-stack
```

Thus only the absolute path to source directory is causing issues. Rest of the details are reproducible.

With this patch:

```
# strings ../build/core/version.o 
4.2.0-1-ge6d7dc0aa (gcc version 14.1.0 (GCC)) #2 Fri Apr 12 09:51:21 UTC 2024 aarch64
GCC: (GNU) 14.1.0
<stdin>
core_v_str
.symtab
.strtab
.shstrtab
.text
.data
.bss
.rodata.core_v_str
.comment
.note.GNU-stack
```

With your proposed patch:

```
--- a/core/arch/arm/plat-corstone1000/conf.mk
+++ b/core/arch/arm/plat-corstone1000/conf.mk
@@ -1,8 +1,10 @@
 PLATFORM_FLAVOR ?= mps3
 
 # ARM debugger needs this
+ifeq ($(CFG_DEBUG_INFO),y)
 platform-cflags-debug-info = -gdwarf-2
 platform-aflags-debug-info = -gdwarf-2
+endif
 
 $(call force,CFG_HWSUPP_MEM_PERM_WXN,y)
 $(call force,CFG_HWSUPP_MEM_PERM_PXN,y)
```

results is also good:

```
# strings ../build/core/version.o 
4.2.0-2-gbc91fb52b-dev (gcc version 14.1.0 (GCC)) #3 Fri Apr 12 09:51:21 UTC 2024 aarch64
GCC: (GNU) 14.1.0
<stdin>
core_v_str
.symtab
.strtab
.shstrtab
.text
.data
.bss
.rodata.core_v_str
.comment
.note.GNU-stack
```

I can switch to this but it does switch the whole corstone1000 to non-debug flag builds by default. Is that ok?

mikkorapeli-linaro: Is CFG_DEBUG_INFO by default set to "y" yes?

It looks like it is, but it's not getting applied everywhere in the yocto/meta-arm build.

mikkorapeli-linaro: Reason I'm wondering is that yocto builds with "-g" by default in CFLAGS everywhere. Then binaries are stripped so that debug symbols are capture to dbg binary packages.

Now optee has bugs in this area and meta-arm recipe for it has disabled this strip completely. Applying the patch from @jforissier causes, I guess, the "-g" earlier in CFLAGS to be effective instead of "-gdwarf-2" and thus  binary size of the non-stripped tee.elf increases from 4.8 Mb to  5.1 Mb. My simpler fix impacted only version.o so binary sizes was reduced by a few hundred bytes.

I'm not sure but I've understood that corstone1000 is quite resource limited and thus this change may not be ok, and ARM debugger compatibility issues too.

jforissier: Yes `CFG_DEBUG_INFO` is set to `y` by default in `mk/config.mk`. My proposal doesn't work because `core/arch/arm/plat-corstone1000/conf.mk` is included by `core/core.mk` *before* `mk/config.mk`. Let me think a bit more about this.

jforissier: The following patch should make sure no flags are passed when `CFG_DEBUG_INFO != y` while still allowing the platform's `conf.mk` to set specific flags when needed (`-gdwarf-2` instead of `-g3`):
```diff
diff --git a/core/arch/arm/arm.mk b/core/arch/arm/arm.mk
index 2da4f97ff..5bb1362ae 100644
--- a/core/arch/arm/arm.mk
+++ b/core/arch/arm/arm.mk
@@ -243,8 +243,13 @@ endif
 platform-cflags-optimization ?= -O$(CFG_CC_OPT_LEVEL)
 
 ifeq ($(CFG_DEBUG_INFO),y)
+# Use ?= to not override any value possibly set in $(platform-dir)/conf.mk
 platform-cflags-debug-info ?= -g3
 platform-aflags-debug-info ?= -g
+else
+# Make sure no flag remains even if set in $(platform-dir)/conf.mk
+platform-cflags-debug-info :=
+platform-aflags-debug-info :=
 endif
 
 core-platform-cflags += $(platform-cflags-optimization)
```

So with this patch if you build with `make ... CFG_DEBUG_INFO=y` (or just `make ...` since `y` is the default for `CFG_DEBUG_INFO`), you should have the absolute paths in all the object files. But if you build with `make ... CFG_DEBUG_INFO=n` there should be no such absolute path, including in `version.o`.

Can you please try that?

mikkorapeli-linaro: I think the paths should be relative with or without CFG_DEBUG_INFO. Yocto defaults to building with -g but it also adds the needed flags to strip out the prefixes and paths are reproducible and relative. Problem here is that these fdebug-prefix-map flags from CFLAGS are not used by this single version.o object file which gets generated. I worked around these by removing the debug flags since that was done before, but maybe the correct fix is to apply the full yocto CFLAGS all the time and everywhere. It's ok to amend to those flags but not ok to remove anything from there.

For example, from mbedtls version.o which is correctly handled:

```
aarch64-oe-linux-musl-gcc -O2 -g ... \
-fmacro-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/git=/usr/src/debug/optee-os/4.2.0 \
-fdebug-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/git=/usr/src/debug/optee-os/4.2.0 \
...
-c lib/libmbedtls/mbedtls/library/version.c -o /home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/build/ta_arm64-lib/libmbedtls/mbedtls/library/version.o
```

The optee-os version.o generation and compile command is missing all the debug prefix maps from yocto standard CFLAGS:

```
echo -e "const char core_v_str[] =" "\"4.2.0-dev \"" "\"(`aarch64-oe-linux-musl-gcc -v 2>&1 | grep "version " | sed 's/ *$//'`) \"" "\"#`cat /home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/build/core/.buildcount` \"" "\"`LC_ALL=C date -u -d @1712915481` \"" "\"aarch64\";\n" | aarch64-oe-linux-musl-gcc -Os -ffunction-sections -fdata-sections -pipe -fpie -mcpu=cortex-a35 -mstrict-align -mno-outline-atomics -mgeneral-regs-only -Os -ffunction-sections -fdata-sections -pipe -fpie -mcpu=cortex-a35 -mstrict-align -mno-outline-atomics -mgeneral-regs-only -fstack-protector-strong -xc - -c -o /home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/build/core/version.o
```

The yocto standard flags are set in the compile environment:

```
export CFLAGS=" -O2 -g -fcanon-prefix-map  -fmacro-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/git=/usr/src/debug/optee-os/4.2.0  -fdebug-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/git=/usr/src/debug/optee-os/4.2.0  -fmacro-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/build=/usr/src/debug/optee-os/4.2.0  -fdebug-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/build=/usr/src/debug/optee-os/4.2.0  -fdebug-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/recipe-sysroot=  -fmacro-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp
-oe-linux-musl/optee-os/4.2.0/recipe-sysroot=  -fdebug-prefix-map=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/recipe-sysroot-native=  -pipe --sysroot=/home/mikko.rapeli/src/base/meta-arm/build/tmp-musl_corstone1000-fvp/work/corstone1000_fvp-oe-linux-musl/optee-os/4.2.0/recipe-sysroot"
```

So how to set the CFLAGS/AFLAGS for all optee-os compile commands? It's ok for optee-os to amend to those, but it's not ok to override them. Only exceptions could be HW specific workarounds.

jforissier: You make a very valid point. I agree there should be a way for OP-TEE to pick up those flags. However blindly using `CFLAGS` may be problematic because builds often include cross compiling so there may be an ambiguity whether it's for the host or the target. How about this:
```diff
diff --git a/mk/compile.mk b/mk/compile.mk
index 338535bf3..d6ce6c228 100644
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -80,7 +80,8 @@ comp-compiler-$2 := $$(CC$(sm))
 comp-flags-$2 = $$(filter-out $$(CFLAGS_REMOVE) $$(cflags-remove) \
 			      $$(cflags-remove-$$(comp-sm-$2)) \
 			      $$(cflags-remove-$2), \
-		   $$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) $$(CFLAGS_WARNS) \
+		   $$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) $$(CFLAGS_$(sm)) \
+		   $$(CFLAGS_WARNS) \
 		   $$(comp-cflags$$(comp-sm-$2)) $$(cflags$$(comp-sm-$2)) \
 		   $$(cflags-lib$$(comp-lib-$2)) $$(cflags-$2))
 ifeq ($C,1)
```

You will need to update the yocto build to do `make ... CFLAGS_core=$(CFLAGS)` or something like that. I am proposing this solution because OP-TEE already has `CFLAGS_ta_arm32`, `CFLAGS_ta_arm64` and the 'ta_armX` par is the "submodule". "core" is the submodule name for the TEE binary. So this is all quite consistent.

mikkorapeli-linaro: I will try this out, thanks!

A related fix to the asm/.S files was this:

https://gitlab.com/mikko.rapeli/meta-arm/-/commit/a7556683c1336eb80791d41dd49c1f867f93f19d

so it seemed that AFLAGS was correctly picked up based on on yocto side CFLAGS to optee side compilations.

mikkorapeli-linaro: @jforissier Your patch still leaves version.o out because that is not using the cupplied flags.

jforissier: Indeed `AFLAGS` are picked up [1]... which TBH I don't really like. I'd rather it would be `AFLAGS_core`. I am afraid using `CFLAGS` could break things quite easily, for instance I recall of some Linux distro (Gentoo I believe) setting CFLAGS by default in the environment and that's for the host of course.

Feel free to change `$$(AFLAGS) to `$(AFLAGS_$(sm))` for consistency, and `$(CXXFLAGS)` too :wink:.

The previous patch (https://github.com/OP-TEE/optee_os/pull/6974#discussion_r1700015127) is still good to have as it fixes a different issue (Corstone1000 building with debug flags although `CFG_DEBUG_INFO` is not `y`).

[1] https://github.com/OP-TEE/optee_os/blob/5fe5dd14dbd813f9937f19c8d0e1409c64e99590/mk/compile.mk#L98

mikkorapeli-linaro: I think simplest would be to just use the provided CC, LD, CFLAGS, CPPFLAGS, etc.

If someone sets those to some values, then it's their problem. Sane build environments set those to working, sane, distro wide defaults also when cross compiling.

That's what yocto does, and Debian, Ubuntu, RedHat. The integrator needs do work if these variables are not used and custom ones used instead. All these complicate the integration and debugging work.

There is still the issue that generated version.o is not using the same default flags which are used elsewhere, even if the debug info build is disabled.

jforissier: OK so how about:
```diff
diff --git a/core/arch/arm/kernel/link.mk b/core/arch/arm/kernel/link.mk
index 49e9f4fa1..377a82b65 100644
--- a/core/arch/arm/kernel/link.mk
+++ b/core/arch/arm/kernel/link.mk
@@ -151,7 +151,7 @@ define update-buildcount
 endef
 
 # filter-out to workaround objdump warning
-version-o-cflags = $(filter-out -g3,$(core-platform-cflags) \
+version-o-cflags = $(filter-out -g3,$(CFLAGS) $(core-platform-cflags) \
 			$(platform-cflags) $(cflagscore))
 # SOURCE_DATE_EPOCH defined for reproducible builds
 ifneq ($(SOURCE_DATE_EPOCH),)
diff --git a/mk/compile.mk b/mk/compile.mk
index 338535bf3..091640a7a 100644
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -80,7 +80,8 @@ comp-compiler-$2 := $$(CC$(sm))
 comp-flags-$2 = $$(filter-out $$(CFLAGS_REMOVE) $$(cflags-remove) \
 			      $$(cflags-remove-$$(comp-sm-$2)) \
 			      $$(cflags-remove-$2), \
-		   $$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) $$(CFLAGS_WARNS) \
+		   $$(CFLAGS$$(arch-bits-$$(comp-sm-$2))) $$(CFLAGS) \
+		   $$(CFLAGS_WARNS) \
 		   $$(comp-cflags$$(comp-sm-$2)) $$(cflags$$(comp-sm-$2)) \
 		   $$(cflags-lib$$(comp-lib-$2)) $$(cflags-$2))
 ifeq ($C,1)
```
The first chunk fixes `version.o` not using $(CFLAGS) and the second one fixes all other `.c `files.

core/arch/arm/kernel/link.mk
```diff
@@ -163,14 +163,15 @@ CORE_CC_VERSION = `$(CCcore) -v 2>&1 | grep "version " | sed 's/ *$$//'`
 define gen-version-o
 	$(call update-buildcount,$(link-out-dir)/.buildcount)
 	@$(cmd-echo-silent) '  GEN     $(link-out-dir)/version.o'
-	$(q)echo -e "const char core_v_str[] =" \
+	$(q)cd $(link-out-dir) && \
+		$(q)echo -e "const char core_v_str[] =" \
```
jforissier, line 167: You need to remove the `$(q)` here (that's what makes CI fail)

core/arch/riscv/kernel/link.mk
```diff
@@ -74,14 +74,15 @@ CORE_CC_VERSION = `$(CCcore) -v 2>&1 | grep "version " | sed 's/ *$$//'`
 define gen-version-o
 	$(call update-buildcount,$(link-out-dir)/.buildcount)
 	@$(cmd-echo-silent) '  GEN     $(link-out-dir)/version.o'
-	$(q)echo -e "const char core_v_str[] =" \
+	$(q)cd $(link-out-dir) && \
+		$(q)echo -e "const char core_v_str[] =" \
```
jforissier, line 78: And here too of course

core/arch/arm/kernel/link.mk
mikkorapeli-linaro: Yep, done now.

core/arch/arm/kernel/link.mk
mikkorapeli-linaro: This is good, CFLAGS goes to all compilations now.

jforissier (commented):
For all commits:

Thanks!

PR #6975, opened 2024-08-01
Add i.MX91 support

sahilnxp:
This pull request adds support for i.MX91 EVK platform


jforissier (commented):

About the `drivers:` commits, I'm thinking it would be so much easier to add new boards if these drivers were taking info from the DT rather than hardcode them...

ricardosalveti:
> About the `drivers:` commits, I'm thinking it would be so much easier to add new boards if these drivers were taking info from the DT rather than hardcode them...

Yeah, it would be nice to see DT used more on the imx8/9 targets.

sahilnxp:
Thanks @jforissier @ricardosalveti for comments.
Regarding the drivers relying on DT, we will try to work towards that.

Tags updated.

PR #6983, opened 2024-08-09
RISC-V: Implement foreign interrupt handler

gagachang:
Foreign interrupts are interrupts that assert during OP-TEE execution, but are not owned and handled by OP-TEE.
For example, in SMP system, if an Linux side interrupt assert during TA execution, OP-TEE can suspend current TA and yield the execution to Linux to handle that interrupt.
This PR implements such foreign interrupt handler for RISC-V architecture.

To determine whether the incoming interrupt is foreign interrupt or not, we leverage `THREAD_EXCP_FOREIGN_INTR`, which aliases bits of `mip` or `sip`.
The type of incoming taken interrupt can be determined by value of `XCAUSE` CSR.
First, we translate `XCAUSE` to corresponding `mip` or `sip` bit.
Then we compare it with `THREAD_EXCP_FOREIGN_INTR` to determine whether the incoming interrupt is foreign interrupt or not.
If yes, we call `thread_foreign_interrupt_handler()` to suspend current thread and return to udomain.

Besides, we let each platform choose their own type of foreign interrupts.
By default, only external interrupt are regarded as foreign interrupt.
For RISC-V virt machine, all external/timer/software interrupts are regarded as foreign interrupts.

### DIFF
--- a/core/arch/riscv/plat-virt/platform_config.h
+++ b/core/arch/riscv/plat-virt/platform_config.h
@@ -9,6 +9,7 @@
 #define PLATFORM_CONFIG_H
 
 #include <mm/generic_ram_layout.h>
+#include <riscv.h>
 
 /* The stack pointer is always kept 16-byte aligned */
 #define STACK_ALIGNMENT		16
@@ -91,5 +92,9 @@
 #else
 #define RISCV_MTIME_RATE 1000000
 #endif
+
+#define PLAT_THREAD_EXCP_FOREIGN_INTR	\
+	(CSR_XIE_EIE | CSR_XIE_TIE | CSR_XIE_SIE)
+#define PLAT_THREAD_EXCP_NATIVE_INTR	(0)
 
 #endif /*PLATFORM_CONFIG_H*/
--- a/core/arch/riscv/include/kernel/thread_arch.h
+++ b/core/arch/riscv/include/kernel/thread_arch.h
@@ -8,9 +8,21 @@
 
 #ifndef __ASSEMBLER__
 #include <compiler.h>
-#include <riscv.h>
 #include <types_ext.h>
 #endif
+
+#include <platform_config.h>
+#include <riscv.h>
+
+/*
+ * Each RISC-V platform must define their own values.
+ * See core/arch/riscv/plat-virt/platform_config.h for example.
+ */
+#define THREAD_EXCP_FOREIGN_INTR	PLAT_THREAD_EXCP_FOREIGN_INTR
+#define THREAD_EXCP_NATIVE_INTR		PLAT_THREAD_EXCP_NATIVE_INTR
+
+#define THREAD_EXCP_ALL			(THREAD_EXCP_FOREIGN_INTR |\
+					 THREAD_EXCP_NATIVE_INTR)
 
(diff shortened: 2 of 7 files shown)


gagachang:
Hello @maroueneboubakri , would you like to review this PR?
Any problems or recommendations are welcome.

core/arch/riscv/plat-virt/platform_config.h
```diff
@@ -92,4 +93,8 @@
 #define RISCV_MTIME_RATE 1000000
 #endif
 
+#define PLAT_THREAD_EXCP_FOREIGN_INTR	\
```
jenswi-linaro, line 96: We have only one riscv platform, but that one can't use the default values. Does that make any sense?

gagachang: Actually, we expect that each riscv platform (even new platforms with new ISA extensions in the future) may define their own value.
So I proposed this solution.
Do you have any recommendation ?

jenswi-linaro: In that case, there's no need for a default value.

gagachang: > In that case, there's no need for a default value.

You mean we should define `THREAD_EXCP_FOREIGN_INTR` and `THREAD_EXCP_NATIVE_INTR` in each `platform_config.h`.
And remove the default value ?

jenswi-linaro: I think it's good to have a `PLAT_` prefix, but there's no need for the
``` C
#else
#define THREAD_EXCP_NATIVE_INTR		(CSR_XIE_SIE | CSR_XIE_TIE)
#endif
````
So in `core/arch/riscv/include/kernel/thread_arch.h` it would become:
``` C
#define THREAD_EXCP_FOREIGN_INTR	PLAT_THREAD_EXCP_FOREIGN_INTR
#define THREAD_EXCP_NATIVE_INTR		PLAT_THREAD_EXCP_NATIVE_INTR
```

gagachang: I've applied what @jenswi-linaro suggested.
@maroueneboubakri Do you have any comments ?

gagachang:
Rebase to master.

Also I changed to keep original definitions for foreign interrupts and native interrupts
1. Set external interrupts as foreign interrupts
2. Set timer and software interrupts as native interrupts

Spike platform also needs these two `PLAT_THREAD_EXCP_XXX` definitions so add them for Spike.

jenswi-linaro:

gagachang:
> Also I changed to keep original definitions for foreign interrupts and native interrupts
> 
> 1. Set external interrupts as foreign interrupts
> 2. Set timer and software interrupts as native interrupts

I decided to keep all kinds of interrupts as foreign interrupts, on RISC-V virt platform.
Because current RISE PoC adopts this scenario.
We would modify it once PoC supports OP-TEE native interrupts.

gagachang:
Hi @jforissier , seems there are no more reviewers' comments.
We would like to merge this PR!

jforissier:
@gagachang sure, thank you!

PR #6985, opened 2024-08-12
Secure interrupt handling in S-EL0 SPs

imre-kis-arm:
The FF-A spec allows secure partitions to receive interrupts. This PR enables the SPMC to configure interrupts based on the SP manifest and forward them to the SPs.

When an interrupt is triggered which targets the secure world, the SPMD will catch it in EL3 and then it sends an FFA_INTERRUPT message to the SPMC. In OP-TEE we continue running in `ffa_msg_loop` disable native interrupts in the [first instruction](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/thread_spmc_a64.S#L57). However the active secure interrupt will trigger an exception just before this instruction. OP-TEE handles this exception and clears the interrupt flag and returns from the exception. After this, point the FFA_INTERRUPT message is handled normally, in `thread_spmc_msg_recv`.

The difficult part of this feature is that the SP has to handle the interrupt and clear the peripheral interrupt flag, before we clear the interrupt flag in the interrupt controller (i.e. in the GIC) from OP-TEE. After approaching the problem from several different directions I ended up with this solution, where the native interrupt handler is diverted before exiting from the interrupt context to the point where the SPs can handle the interrupt.

Related patches:

- [optee_test](https://github.com/OP-TEE/optee_test/pull/753)
- [Trusted Service](https://review.trustedfirmware.org/c/TS/trusted-services/+/26669)

### DIFF
--- a/core/arch/arm/kernel/thread_a64.S
+++ b/core/arch/arm/kernel/thread_a64.S
@@ -22,13 +22,6 @@
 		madd	x\res, x\tmp0, x\tmp1, x\res
 	.endm
 
-	.macro return_from_exception
-		eret
-		/* Guard against speculation past ERET */
-		dsb nsh
-		isb
-	.endm
-
 	.macro b_if_spsr_is_el0 reg, label
 		tbnz	\reg, #(SPSR_MODE_RW_32 << SPSR_MODE_RW_SHIFT), \label
 		tst	\reg, #(SPSR_64_MODE_EL_MASK << SPSR_64_MODE_EL_SHIFT)
@@ -1312,8 +1305,12 @@ END_FUNC el0_sync_abort
 	/* Restore x0..x1 */
 	load_xregs sp, THREAD_CORE_LOCAL_X0, 0, 1
 
+#if defined(CFG_SECURE_PARTITION)
+	b thread_return_to_sp_interrupt_handler
+#else /* CFG_SECURE_PARTITION */
 	/* Return from exception */
 	return_from_exception
+#endif /* CFG_SECURE_PARTITION */
 1:	b	eret_to_el0
 .endm
 
--- a/core/arch/arm/kernel/thread_spmc_a64.S
+++ b/core/arch/arm/kernel/thread_spmc_a64.S
@@ -120,6 +120,110 @@ FUNC spmc_sp_thread_entry , :
 	mov     w7, w27
 	b .ffa_msg_loop
 END_FUNC spmc_sp_thread_entry
+
+FUNC thread_return_to_sp_interrupt_handler , :
+	/* Save interrupt ELR as LR */
+	mrs	lr, elr_el1
+
+	/* Push X0 */
+	str	x0, [sp, #8]
+
+	/* Set ELR_EL1 to point to thread_native_intr_sp_handler */
+	adr_l	x0, thread_sp_interrupt_handler_starter
+	msr	elr_el1, x0
+
+	/* Pop X0 */
+	ldr	x0, [sp, #8]
+
+	return_from_exception
+
+thread_sp_interrupt_handler_starter:
+	/* Disable native interrupts */
+	msr	daifset, #DAIFBIT_NATIVE_INTR
+
+	/* Store original FF-A call arguments */
+	sub	sp, sp, #THREAD_SMC_ARGS_SIZE
+	store_xregs sp, THREAD_SMC_ARGS_X0, 0, 7
+
+	/* Store original ELR_EL1 and SP to thread stack */
+	add x0, sp, #THREAD_SMC_ARGS_SIZE
+	sub	sp, sp, #16
+	stp lr, x0, [sp]
+
+	/* Set function arguments */
+	add	x0, sp, #16
+	mov x1, sp
+
+	b thread_sp_alloc_and_run_interrupt_handler
+END_FUNC thread_return_to_sp_interrupt_handler
+
+/**
+ * void spmc_sp_interrupt_handler_entry(struct thread_smc_args *args,
+ * 					uint64_t *lr_sp)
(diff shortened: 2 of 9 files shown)


core/arch/arm/kernel/thread_a64.S
```diff
@@ -1312,8 +1305,12 @@ END_FUNC el0_sync_abort
 	/* Restore x0..x1 */
 	load_xregs sp, THREAD_CORE_LOCAL_X0, 0, 1
 
+#if defined(CFG_SECURE_PARTITION)
+	b thread_return_to_sp_interrupt_handler
```
jenswi-linaro, line 1309: We don't guarantee that native interrupts are always masked at S-EL0, they are always unmasked for TAs.

core/arch/arm/kernel/thread_spmc_a64.S
```diff
@@ -120,6 +120,110 @@ FUNC spmc_sp_thread_entry , :
...
+	/* Store original ELR_EL1 and SP to thread stack */
+	add x0, sp, #THREAD_SMC_ARGS_SIZE
+	sub	sp, sp, #16
+	stp lr, x0, [sp]
+
+	/* Set function arguments */
+	add	x0, sp, #16
+	mov x1, sp
+
+	b thread_sp_alloc_and_run_interrupt_handler
```
jenswi-linaro, line 157: What if we fail to allocate a thread?

core/arch/arm/kernel/thread_spmc_a64.S
```diff
@@ -120,6 +120,110 @@ FUNC spmc_sp_thread_entry , :
...
+	msr	elr_el1, x0
+
+	/* Pop X0 */
+	ldr	x0, [sp, #8]
+
+	return_from_exception
+
+thread_sp_interrupt_handler_starter:
+	/* Disable native interrupts */
+	msr	daifset, #DAIFBIT_NATIVE_INTR
```
jenswi-linaro, line 142: This is too late if another native interrupt is pending already.

jenswi-linaro (commented):
Each S-EL0 SP has its own preallocated thread, I think that FFA_INTERRUPT should be delivered on that thread too. It would avoid the problem with running out of threads.

How about letting the `sp_interrupt_callback()` record in thread_core_local that there's an interrupt pending for an SP that should be handled before returning from the exception (but after the stack as been restored), provided that the SP thread is idle? If the SP thread isn't idle you can record the pending interrupt in the SP session to be delivered as soon as the SP returns.

imre-kis-arm:
Thank you for the comments. TBH it is a bit hard to follow the thread handling in OP-TEE. Is there any documentation about it?

> How about letting the sp_interrupt_callback() record in thread_core_local that there's an interrupt pending for an SP that should be handled before returning from the exception (but after the stack as been restored), provided that the SP thread is idle? If the SP thread isn't idle you can record the pending interrupt in the SP session to be delivered as soon as the SP returns.

If I understand this part correctly, you are suggesting the call sequence of the diagram below.
sec_int

- Is this correct?
- How can we return from the SP thread (marked with red ???) to the interrupt handler? Or should we return directly to the interrupted context?

jenswi-linaro:
> Each S-EL0 SP has its own preallocated thread, I think that FFA_INTERRUPT should be delivered on that thread too. It would avoid the problem with running out of threads.

I'm sorry, I was wrong, S-EL0 SPs don't have preallocated threads.

We must figure out what to do when an interrupt is triggered but there isn't any thread available to deliver it to the target SP.

>  TBH it is a bit hard to follow the thread handling in OP-TEE. Is there any documentation about it?

https://optee.readthedocs.io/en/latest/architecture/core.html is all we have.

imre-kis-arm:
I was thinking of implementing a slightly a different approach which might solve some of the issues.

1. Stores and mask the pending interrupt `sp_interrupt_callback`.
2. Return from the interrupt to the interrupted context without any change.
3. Check for stored interrupt in `ffa_msg_loop` before the SMC call. If there are stored interrupts, then propagate the interrupt to the SP though the normal call path.
4. Unmask the interrupt after the SP returned with FFA_MSG_WAIT.

Secure interrupts can happen at a few places:

- Non-Secure side/EL3
- While running a TA
- After the SMC instruction, just after entering into S-EL1
- Before the SMC instruction, before exiting S-EL1

The first three case should work fine with this approach. However in the last case, we still have to divert the return address of the interrupt so OP-TEE gets a chance to do step 3 and 4 before the SMC call.

Would this be a better method of handling secure interrupts in SPs?

imre-kis-arm:
> I think that we could remove some complexity if all SP handling were done using the temporary stack. 

Wouldn't this cause an issue when the SP is interrupted by a normal world interrupt and then we return to OP-TEE via `FFA_RUN`? Currently this is done by `thread_state_suspend` and `thread_resume_from_rpc`.

> That obviously rules out access to OP-TEE services, but I don't think SPs will ever need that anyway.

I think this is fine, I also don't expect using any OP-TEE service from SPs.

jenswi-linaro:
> > I think that we could remove some complexity if all SP handling were done using the temporary stack.
> 
> Wouldn't this cause an issue when the SP is interrupted by a normal world interrupt and then we return to OP-TEE via `FFA_RUN`? Currently this is done by `thread_state_suspend` and `thread_resume_from_rpc`.

Yes, that's one of the obstacles to overcome. The alternative is to add error handling when `thread_sp_alloc_and_run()` fails.

gyuri-szing:
This change is still relevant, pls reopen.

jforissier:
@gyuri-szing reopened. I understand some changes were requested by @jenswi-linaro and were not addressed.

PR #6993, opened 2024-08-15
core: riscv: Simplify SP setup in setup_unwind_user_mode()

gagachang:
The parameter "regs" is the stack pointer which is allocated to store system call registers when calling thread_scall_handler().
Thus, we can simply get the original stack pointer by "regs + 1" equation, and use when exiting user mode.

The code is referenced from ARM's [setup_unwind_user_mode()](https://github.com/OP-TEE/optee_os/blob/9cb4152f1b0c297b964bdb7bd84b49615afd73fb/core/arch/arm/kernel/thread.c#L1099).

### DIFF
--- a/core/arch/riscv/kernel/thread_arch.c
+++ b/core/arch/riscv/kernel/thread_arch.c
@@ -100,7 +100,13 @@ static void setup_unwind_user_mode(struct thread_scall_regs *regs)
 {
 	regs->ra = (uintptr_t)thread_unwind_user_mode;
 	regs->status = xstatus_for_xret(true, PRV_S);
-	regs->sp = thread_get_saved_thread_sp();
+	/*
+	 * We are going to exit user mode. The stack pointer must be set as the
+	 * original value it had before allocating space of scall "regs" and
+	 * calling thread_scall_handler(). Thus, we can simply set stack pointer
+	 * as (regs + 1) value.
+	 */
+	regs->sp = (uintptr_t)(regs + 1);
 }
 
 static void thread_unhandled_trap(unsigned long cause __unused,


gagachang:
This PR depends on https://github.com/OP-TEE/optee_os/pull/6819, which improves the stack usage of trap handlers.

gagachang:
Hi @maroueneboubakri , would you like to review this small patch ?

gagachang:
Hi @jforissier , seems the CI failure is related to fetching or hafnium.

```
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
error: 60833 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
fatal: Fetched in submodule path 'prebuilts', but it did not contain 5cd8406945e7f8f4bc9814167113886ae2a206fc. Direct fetching of that commit failed.
fatal: 
make: *** [Makefile:398: .hafnium_checkout] Error 128
make: *** Waiting for unfinished jobs....
```

jforissier:
@gagachang yes, some temporary error with the Hafnium repository. I tried `make .hafnium_checkout` on my machine and it worked so I have restarted the CI job and hopefully it should complete this time.
Anyways, it is not the first time that `git.trustedfirmware.org` has caused transient problems so I am leaning towards replacing all the TFO links with their GitHub mirrors. GitHub has always been rock solid.

gagachang:
Hi @maroueneboubakri , do you have time to review this PR ?

gagachang:
Hi @jforissier , if there is no reviewer's feedback, we want to merge this PR.
I think the modification is simple.

jforissier:
> Hi @jforissier , if there is no reviewer's feedback, we want to merge this PR. I think the modification is simple.

Sure, and the commit has a Reviewed-by from your colleague so it's fine.

PR #6997, opened 2024-08-16
core: arm64: increase size of abort stack to 4096

jenswi-linaro:
To avoid stack overruns with CFG_WITH_PAGER=y and
CFG_CORE_DEBUG_CHECK_STACKS=y increase the abort stack from 3072 to 4096.


jforissier (commented):

jenswi-linaro:
Tag applied.

PR #6999, opened 2024-08-19
Phys mem allocation functions

jenswi-linaro:

### DIFF
--- a/core/mm/fobj.c
+++ b/core/mm/fobj.c
@@ -13,6 +13,7 @@
 #include <mm/core_memprot.h>
 #include <mm/core_mmu.h>
 #include <mm/fobj.h>
+#include <mm/phys_mem.h>
 #include <mm/tee_mm.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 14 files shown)


jenswi-linaro:
Rebased on master

core/mm/fobj.c
```diff
@@ -759,7 +762,7 @@ struct fobj *fobj_sec_mem_alloc(unsigned int num_pages)
 	if (MUL_OVERFLOW(num_pages, SMALL_PAGE_SIZE, &size))
 		goto err;
 
-	f->mm = tee_mm_alloc(&tee_mm_sec_ddr, size);
+	f->mm = phys_mem_ta_alloc(size);
```
etienne-lms, line 765: This is embedded only when pager is enabled, as for other changes in this source file. Wouldn't it be more consistent is `nex_phys_mem_ta_alloc()` is used here?

etienne-lms (commented):
"core: add phys_mem allocation functions",
"core: rename to core_mmu_init_phys_mem()" and.
"checkpatch: add ignore=STRLCPY". 

"core: replace tee_mm_sec_ddr with phys_mem functions". There changes in the _virtualization.c_ part i'm not very familiar with, the rest looks all good to me.

core/mm/phys_mem.c
```diff
@@ -0,0 +1,213 @@
...
+
+	tee_mm_init(pool, b, sz, CORE_MMU_USER_CODE_SHIFT, flags);
+	return pool;
+}
+
+void nex_phys_mem_init(paddr_t core_base, paddr_size_t core_size,
+		       paddr_t ta_base, paddr_size_t ta_size)
+{
+	uint32_t flags = TEE_MM_POOL_NEX_MALLOC;
+
```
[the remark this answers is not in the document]
jenswi-linaro: Sure, this function is supposed to be called once only so `assert(!nex_core_pool && !nex_ta_pool)` should be enough. I'll add the same for `phys_mem_init()` below.

core/mm/fobj.c
jenswi-linaro: Yes and no. The pager isn't supported with virtualization so it doesn't change anything, but we have quite a few places around here where non-nex functions are used. Either we'd have to fix all those too or it would only make things even more confusing. I'd also like to keep the scope to what this commit is about, getting rid of "tee_mm_sec_ddr".

I wouldn't mind a cleanup where the pager uses nex resources as needed, but I don't think that will be enough to get the pager to work with virtualization.

jenswi-linaro:

jenswi-linaro:
Ping?

PR #7002, opened 2024-08-19
ci: add build with Mbed TLS and crypto extensions

jforissier:
Build for the Juno platform with Mbed TLS as the core crypto library. This would have caught the issue fixed in commit 021fee0affe5 ("core: mbedtls: Fix build").


jenswi-linaro:
Why Juno? Wouldn't it work with QEMU?

jforissier:
For some reason the build doesn't fail on QEMUv8, I need to check why.

jforissier:
> For some reason the build doesn't fail on QEMUv8, I need to check why.

Ah, the reason is `CFG_CRYPTO_WITH_CE`. PLATFORM=vexpress-qemu_armv8a` will fail in the same way when `CFG_CRYPTO_WITH_CE=y`. @jenswi-linaro would you prefer I add a QEMU build instead of Juno?

jenswi-linaro:
> > For some reason the build doesn't fail on QEMUv8, I need to check why.
> 
> Ah, the reason is `CFG_CRYPTO_WITH_CE`. `PLATFORM=vexpress-qemu_armv8a` will fail in the same way when `CFG_CRYPTO_WITH_CE=y`. @jenswi-linaro would you prefer I add a QEMU build instead of Juno?

Yes, please. :-)

jforissier:
Update

jenswi-linaro (commented):

PR #7004, opened 2024-08-21
plat-sam: implement PL310 SMC protocol

TonyHan11:
When Linux runs in normal world, it expects the PL310 to be initially disabled, and then invokes SMCs to enable it.
Let CFG_PL310_SIP_PROTOCOL=y, and the L2 cache will be left untouched until the OS enables it.

### DIFF
--- a/core/arch/arm/plat-sam/nsec-service/sm_platform_handler.c
+++ b/core/arch/arm/plat-sam/nsec-service/sm_platform_handler.c
@@ -10,13 +10,34 @@
 #include <kernel/tee_misc.h>
 #include <mm/core_memprot.h>
 #include <sam_sfr.h>
+#include <sam_pl310.h>
 #include <sm/optee_smc.h>
 #include <sm/sm.h>
 #include <smc_ids.h>
 
 static enum sm_handler_ret sam_sip_handler(struct thread_smc_args *args)
 {
+	/*
+	 * As all sama5 SoCs are single-core ones, check the code compiled for a
+	 * single core. No serializations done to protect against concurrency.
+	 */
+	static_assert(CFG_TEE_CORE_NB_CORE == 1);
+
 	switch (OPTEE_SMC_FUNC_NUM(args->a0)) {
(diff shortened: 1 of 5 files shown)


core/arch/arm/plat-sam/nsec-service/sm_platform_handler.c
```diff
@@ -10,13 +10,28 @@
...
 #include <sam_sfr.h>
+#include <sam_pl310.h>
 #include <sm/optee_smc.h>
 #include <sm/sm.h>
 #include <smc_ids.h>
 
 static enum sm_handler_ret sam_sip_handler(struct thread_smc_args *args)
 {
 	switch (OPTEE_SMC_FUNC_NUM(args->a0)) {
+#ifdef CFG_PL310_SIP_PROTOCOL
```
jenswi-linaro, line 21: Do the accesses to the PL310 need to be serialized to protect against concurrency?

TonyHan11: Actually, in the Linux kernel, only PL310_ENABLED is called for only once. And it's assumed that the OS will keep operations serialized.

TonyHan11: Would it be OK with adding the following at the beginning of this function? Thanks.
```
/*
 * As all sama5 SoCs are single-core ones, check the code compiled for a
 * single core. No serializations done to protect against concurrency.
 */
COMPILE_TIME_ASSERT(CFG_TEE_CORE_NB_CORE == 1);
```

jenswi-linaro: Sounds good, please replace `COMPILE_TIME_ASSERT` with `static_assert`.

TonyHan11: Updated, thank you.

jenswi-linaro (commented):

TonyHan11:
Review tags added.

PR #7008, opened 2024-08-26
RISC-V: Support CFG_CORE_DEBUG_CHECK_STACKS

gagachang:
To support CFG_CORE_DEBUG_CHECK_STACKS=y for RISC-V, we increase the following stacks:
- Set STACK_THREAD_SIZE as 10240 bytes
- Set STACK_ABT_SIZE as 4096 bytes
- Set STACK_CHECK_EXTRA as 1536 bytes, like what ARM does


jenswi-linaro:

PR #7009, opened 2024-08-26
RISC-V: Fix initial value of a0 in "detect_csr" assembly macro

gagachang:
We want to set value of the register a0 to 1 here.
To set initial value of the register a0 to 1, the assembly code should be "li a0, 1" instead of "addi a0, a0, 1".
Otherwise, a0 will be any value greater than 1.


jenswi-linaro (commented):

PR #7010, opened 2024-08-27
riscv: core_mmu_arch: fix address translation helpers

lyctw:
Fix misuse of `virt_to_phys()` and `phys_to_virt()` where is not aware MMU is disabled.


jenswi-linaro (commented):

PR #7012, opened 2024-08-28
plat-k3: drivers: Change SA2UL_init service to service_init_crypto

manorit2001:
Since commit https://github.com/OP-TEE/optee_os/commit/11d8578d93f0aee793b28221b06187f62bb7b24b ("core: arm: call call_driver_initcalls()
late"), driver_init is deferred and thread_update_canaries tries to get
random_stack_canaries which requires the TRNG driver to be setup. Since
it was being setup as part of driver_init, it lead to crash on K3
platforms.

Change driver_init to service_init_crypto which is meant to be used for
initialization of crypto operations. Also, for the TISCI services to be
available before service_init_crypto, change init_ti_sci invocation to
early_init_late.

Some other driver vendors using driver_init for RNG initialization might 
also be affected.

Signed-off-by: Kamlesh Gurudasani <kamlesh@ti.com>

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul.c
@@ -150,4 +150,4 @@ static TEE_Result sa2ul_init(void)
 
 	return TEE_SUCCESS;
 }
-driver_init(sa2ul_init);
+service_init_crypto(sa2ul_init);
--- a/core/arch/arm/plat-k3/main.c
+++ b/core/arch/arm/plat-k3/main.c
@@ -64,7 +64,13 @@ static TEE_Result init_ti_sci(void)
 	return TEE_SUCCESS;
 }
 
-service_init(init_ti_sci);
+/*
+ * TISCI services are required for initialization of TRNG service that gets
+ * initialized during service_init_crypto.
+ *
+ * Initialize TISCI service just before service_init_crypto.
+ */
+early_init_late(init_ti_sci);
 
 static TEE_Result secure_boot_information(void)
 {


jenswi-linaro:

glneo:
> vendors using driver_init for RNG initialization

This is my worry, I see a couple examples of that already, I'm assuming they are broken but just haven't tested the latest master branch yet. A generic fix would be better.

Moving our TI-SCI service to not use service_init() but instead early_init_late() seems fragile, and it may not be obvious without a comment in-code why we do it at that stage.

manorit2001:
> > vendors using driver_init for RNG initialization
> 
> This is my worry, I see a couple examples of that already, I'm assuming they are broken but just haven't tested the latest master branch yet. A generic fix would be better.

I understand that but I don't see a way for generic fix tbh, it's wrong implementation only if we were using driver_init instead of service_init_crypto, even if I go and change all the vendor rng init to be using service_init_crypto, there platforms could still be broken as we did have to change our tisci init as well to early init late to make it work and at that point it's just vendor specific implementation that I don't see a generic fix for...

> Moving our TI-SCI service to not use service_init() but instead early_init_late() seems fragile, and it may not be obvious without a comment in-code why we do it at that stage.

The priority of service_init_crypto itself is less so the only point where we can instantiate pur tisci driver seems to be this only.... Thought yeah, ig adding a common can help , would do the needful

GseoC:
> > vendors using driver_init for RNG initialization
> 
> This is my worry, I see a couple examples of that already, I'm assuming they are broken but just haven't tested the latest master branch yet. A generic fix would be better.

I didn't see that it could break the boot on our platforms because we didn't  have a probe deferral on the stm32_rng driver. Faking a driver probe defer, I do get the `Panic 'Failed to generate random stack canary'` ... @etienne-lms , we may have to look into it :)

Thanks for raising the issue

manorit2001:
Hi @jforissier , would it be possible to get this patch merged? It's breaking all the K3 platforms right now and hampering boot.

jforissier:
@manorit2001 sure

manorit2001:
> @manorit2001 sure

thanks @jforissier !

PR #7016, opened 2024-08-29
mk/compile.mk: fix pre-processed device tree dependencies file target

GseoC:
Fixes the target in the pre-processed device tree dependency file


jforissier (commented):

GseoC:
Tag applied

PR #7017, opened 2024-08-30
riscv: plat-virt: Set CFG_BOOT_SYNC_CPU=n

gagachang:
On RISC-V QEMU virt platform, OP-TEE OS runs as S-mode.
There is a secure monitor runs as M-mode and controls the hart state of the secondary CPUs in SMP system (e.g., by SBI HSM extension) during OP-TEE OS secondary CPUs booting.

Thus, RISC-V virt platform does not need CFG_BOOT_SYNC_CPU.


jenswi-linaro (commented):

PR #7021, opened 2024-09-02
Removal of firewall_dt_probe_bus()  and stm32mp25 firewall bus populate + RNG updates

GseoC:
Remove firewall_dt_probe_bus() from the firewall framework as it seems unlikely that we can have a consensual implementation of this feature for every vendor.

This P-R adds the probing of peripherals under the RIFSC firewall bus. This means that from now on, access on peripherals when probing the RIFSC will be filtered to catch any firewall inconsistencies. In case the peripheral is shared between different context, the RIF semaphore is acquired, if available.

Also update the RNG to better handle the different platforms, pm and RCC resources.


etienne-lms (commented):
"drivers: firewall: remove firewall_dt_probe_bus()",
"dts: stm32: declare RIFSC as an access-controller on stm32mp2 platforms",
"drivers: stm32_rifsc: restrain access on non secure peripherals for OP-TEE",
"plat-stm32mp2: conf: default enable CFG_DRIVERS_FIREWALL"

Few minor comments for  commits
"drivers: stm32_rng: update clock and power management" and
"drivers: stm32_rifsc: update RIFSC as a firewall controller".
It shouold be ok for me once addressed.

GseoC:
Comments addressed

etienne-lms (commented):
"drivers: stm32_rifsc: update RIFSC as a firewall controller" and
"drivers: stm32_rng: update clock and power management".

GseoC:
Tags applied, thanks

PR #7022, opened 2024-09-03
core: add CFG_CORE_STATIC_ASLR_SEED

jenswi-linaro:
Add CFG_CORE_STATIC_ASLR_SEED to override the used seed if CFG_CORE_ASLR=y. CFG_CORE_STATIC_ASLR_SEED is intended to help debugging ASLR related issues by using the same address layout each time.

CFG_CORE_STATIC_ASLR_SEED requires CFG_INSECURE=y.

### DIFF
--- a/core/arch/arm/kernel/entry_a32.S
+++ b/core/arch/arm/kernel/entry_a32.S
@@ -473,6 +473,9 @@ shadow_stack_access_ok:
 
 #ifdef CFG_CORE_ASLR
 	bl	get_aslr_seed
+#ifdef CFG_CORE_ASLR_SEED
+	mov_imm	r0, CFG_CORE_ASLR_SEED
+#endif
 #else
 	mov	r0, #0
 #endif
--- a/core/arch/arm/kernel/entry_a64.S
+++ b/core/arch/arm/kernel/entry_a64.S
@@ -314,6 +314,9 @@ clear_nex_bss:
 
 #ifdef CFG_CORE_ASLR
 	bl	get_aslr_seed
+#ifdef CFG_CORE_ASLR_SEED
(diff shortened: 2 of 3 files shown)


core/arch/arm/kernel/entry_a32.S
```diff
@@ -473,6 +473,9 @@ shadow_stack_access_ok:
 
 #ifdef CFG_CORE_ASLR
 	bl	get_aslr_seed
+#ifdef CFG_CORE_STATIC_ASLR_SEED
+	mov_imm	r0, CFG_CORE_STATIC_ASLR_SEED
+#endif
```
jforissier, line 478: Why not skip the call to `get_aslr_seed()` when `CFG_CORE_STATIC_ASLR_SEED` is defined?
How about dropping `_STATIC`? i.e., `CFG_CORE_ASLR_SEED`?

jenswi-linaro: It's supposed to change as little as possible. Skipping the call to `get_aslr_seed()` would get that code and it's dependencies GC'ed so it might have an impact. Sure, I'll rename it to `CFG_CORE_ASLR_SEED`.

jenswi-linaro:
Update

core/arch/arm/kernel/entry_a32.S
jforissier: > It's supposed to change as little as possible.

Fair enough.

> Sure, I'll rename it to `CFG_CORE_ASLR_SEED`.

Thanks!

jforissier:

jenswi-linaro:
Tag applied.

PR #7025, opened 2024-09-03
core: virt: initialize heap from virt_guest_created()

jenswi-linaro:
Replace the preinit_early() guest heap initialization with function call in virt_guest_created().

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -912,17 +912,6 @@ static void update_external_dt(void)
 }
 #endif /*!CFG_DT*/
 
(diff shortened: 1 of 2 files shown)


jforissier:

With this, `preinit_early()` is not used anymore. and `preinit_late()` was already unused it seems, so the `preinit_*` could be removed altogether?

jenswi-linaro:
Tag applied.

jenswi-linaro:
> With this, `preinit_early()` is not used anymore. and `preinit_late()` was already unused it seems, so the `preinit_*` could be removed altogether?

I'm not in a hurry to remove them. But if you feel like it, please go ahead.

PR #7026, opened 2024-09-03
core: open-code thread_init_stack()

jenswi-linaro:
The implementations of thread_init_stack() are identical and trivial for both arm and riscv. So simplify code further and open-code it where it's called from in core/kernel/thread.c.


gagachang:
Thanks!

```
Tested-by: Alvin Chang <alvinga@andestech.com>
```

jenswi-linaro:
Tags applied.

Issue #7023, opened 2024-09-03
Potential memory leak?


manoj23:
Hello,

The commit 6959626e1621baa41a2b83be367fdfcfa0f8f78a introduces 2 pointers `head_new` and `head_old`.

`head_new` is being set to NULL before being freed at line  1092.
Shouldn't we free `TEE_Free(obj->attributes);` before `obj->attributes = head_old;` at line 1097?

Thank you,

Georges

jforissier:
Hello @manoj23,

> Shouldn't we free `TEE_Free(obj->attributes);` before `obj->attributes = head_old;` at line 1097?

Yes. Could you please create a PR for this? Thanks!

manoj23:
Hello @jforissier, I created the PR.

PR #7027, opened 2024-09-04
ta: pkcs11: fix memory leak

manoj23:
On update_persistent_object_attributes failure, head_new (stored in
obj->attributes) is not freed, causing a memory leak.

Closes: #7023
Fixes: 6959626 ("ta: pkcs11: preserve object when set attribute fails")
Reported-by: Christian Zoia <czoia@amazon.es>


jforissier (approved):
With the checkpatch errors fixed:

etienne-lms (commented):

PR #7028, opened 2024-09-05
[RISC-V] Allow enabling CFG_TEE_CORE_DEBUG

lyctw:
This series enables CFG_TEE_CORE_DEBUG, a crucial feature for early error detection on RISC-V platforms, such as assertions.

### DIFF
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -450,7 +450,7 @@ void asid_free(unsigned int asid)
 	if (asid) {
 		int i = asid - 1;
 
-		assert(i < RISCV_MMU_ASID_WIDTH && bit_test(&g_asid, i));
+		assert(i < RISCV_MMU_ASID_WIDTH && bit_test(g_asid, i));
 		bit_clear(g_asid, i);
 	}
 
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -289,7 +289,7 @@ static void core_init_mmu_prtn_ta(struct mmu_partition *prtn)
 {
 	unsigned int core = 0;
 
-	assert(user_va_idx != -1);
+	assert(core_mmu_user_va_range_is_defined());
 
 	memset(prtn->user_pgts, 0, CFG_NUM_THREADS * RISCV_MMU_PGT_SIZE);
 	for (core = 0; core < CFG_TEE_CORE_NB_CORE; core++)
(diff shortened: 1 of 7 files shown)


core/arch/riscv/mm/core_mmu_arch.c
```diff
@@ -466,6 +466,7 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
 	int level = 0;
 	unsigned int idx = 0;
 	struct mmu_partition *prtn = core_mmu_get_prtn();
+	vaddr_t offset_mask = 0; /* page offset mask */
```
etienne-lms, line 469: nitpicking: I don't think the inline comment is really useful.

core/arch/riscv/mm/core_mmu_arch.c
```diff
@@ -628,16 +628,16 @@ static void set_user_va_idx(struct mmu_partition *prtn)
 			return;
 		}
 	}
-	if (user_va_idx < 0)
-		panic();
+
+	assert(core_mmu_user_va_range_is_defined());
```
etienne-lms, line 632: This replaces a always embedded test with a debug config assertion. Looks fine but I think it should be stated in the commit message.

core/mm/core_mmu.c
```diff
@@ -2381,7 +2381,14 @@ paddr_t virt_to_phys(void *va)
...
 
-#if defined(CFG_TEE_CORE_DEBUG)
+/*
+ * Don't use check_va_matches_pa() for RISC-V, as its callee
+ * arch_va2pa_helper() will call it eventually, this creates
+ * indirect recursion and can lead to a stack overflow.
+ * Moreover, if arch_va2pa_helper() returns true, it implies
+ * the va2pa mapping is matched, no need to check again.
+ */
+#if defined(CFG_TEE_CORE_DEBUG) && !defined(RV64)
```
etienne-lms, line 2391: Only for `RV64`, not for `RV32`? it seems the implementation of the same for these 2 configurations.

etienne-lms (commented):
For commit "core: riscv: core_mmu_arch: fix arch_va2pa_helper() on superpage translation": commit message header line is a bit too long: could you shorten, e.g.:
"core: riscv: mm: fix arch_va2pa_helper() on superpage translation"

core/arch/riscv/mm/core_mmu_arch.c
lyctw: Upon reflection, I should keep the test unchanged.

etienne-lms (commented):

etienne-lms:
**CI Code style** error report seems due to a false positive.

lyctw:
Tag applied, thanks!

PR #7041, opened 2024-09-16
drivers: se050: Avoid calling i2c functions

hoihochan:
Create a new option called `CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY` indicating
there is no native I2C driver inside TEE and all I2C access are done through
the I2C trampoline.

As a result some of the native I2C calls are either stubbed or skipped.

### DIFF
--- a/core/drivers/crypto/se050/adaptors/apis/sss.c
+++ b/core/drivers/crypto/se050/adaptors/apis/sss.c
@@ -162,7 +162,7 @@ sss_status_t se050_enable_scp03(sss_se05x_session_t *session)
 		if (status != kStatus_SSS_Success)
 			continue;
 
-		if (session->subsystem)
+		if (session && session->subsystem)
 			sss_se05x_session_close(session);
 
 		if (!se050_core_early_init(&keys)) {
--- a/core/drivers/crypto/se050/core/apdu.c
+++ b/core/drivers/crypto/se050/core/apdu.c
@@ -16,6 +16,11 @@ TEE_Result crypto_se_do_apdu(enum crypto_apdu_type type,
 {
 	sss_status_t status = kStatus_SSS_Fail;
 
+	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY) && !se050_session) {
+		/* Defer until REE is ready */
+		return TEE_ERROR_COMMUNICATION;
+	}
+
 	status = sss_se05x_do_apdu(&se050_session->s_ctx, type,
 				   hdr, hdr_len, src_data, src_len,
 				   dst_data, dst_len);
--- a/core/drivers/crypto/se050/glue/i2c.c
+++ b/core/drivers/crypto/se050/glue/i2c.c
@@ -47,7 +47,8 @@ int glue_i2c_write(uint8_t *buffer, int len)
 
 int glue_i2c_init(void)
 {
-	if (transfer == rpc_io_i2c_transfer)
+	if (transfer == rpc_io_i2c_transfer ||
+	    IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY))
 		return 0;
 
 	transfer = native_i2c_transfer;
--- a/core/drivers/crypto/se050/glue/include/i2c_native.h
+++ b/core/drivers/crypto/se050/glue/include/i2c_native.h
@@ -9,8 +9,24 @@
 
 #include <kernel/rpc_io_i2c.h>
 
+#ifdef CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY
+static inline TEE_Result
+native_i2c_transfer(struct rpc_i2c_request *req __unused,
+		    size_t *bytes __unused)
+{
+	/* Stub only */
+	return TEE_ERROR_NOT_SUPPORTED;
+}
+
+static inline int native_i2c_init(void)
+{
+	/* Stub only */
+	return 0;
+}
+#else
 TEE_Result native_i2c_transfer(struct rpc_i2c_request *req,
 			       size_t *bytes);
 int native_i2c_init(void);
+#endif
 
 #endif
--- a/core/drivers/crypto/se050/session.c
+++ b/core/drivers/crypto/se050/session.c
@@ -120,6 +120,11 @@ static TEE_Result se050_early_init_scp03(void)
 
 static TEE_Result se050_session_init(void)
 {
+	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY)) {
+		/* Nothing to do */
+		return TEE_SUCCESS;
+	}
+
 	if (IS_ENABLED(CFG_CORE_SCP03_ONLY))
 		return se050_early_init_scp03();
 
--- a/core/drivers/crypto/se050/crypto.mk
+++ b/core/drivers/crypto/se050/crypto.mk
@@ -37,6 +37,8 @@ CFG_CORE_SE05X_BAUDRATE ?= 3400000
 CFG_CORE_SE05X_I2C_BUS ?= 2
 # I2C access via REE after TEE boot
 CFG_CORE_SE05X_I2C_TRAMPOLINE ?= y
+# TEE has no native I2C driver - all I2C access done via REE
+CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY ?= n
 
 # Extra stacks required to support the Plug and Trust external library
 ifeq ($(shell test $(CFG_STACK_THREAD_EXTRA) -lt 8192; echo $$?), 0)


core/drivers/crypto/se050/adaptors/apis/sss.c
```diff
@@ -162,7 +162,7 @@ sss_status_t se050_enable_scp03(sss_se05x_session_t *session)
 		if (status != kStatus_SSS_Success)
 			continue;
 
-		if (session->subsystem)
+		if (session && session->subsystem)
```
jforissier, line 165: Unrelated change. Please move to a separate commit.

core/drivers/crypto/se050/glue/i2c.c
```diff
@@ -50,9 +50,14 @@ int glue_i2c_init(void)
...
 		return 0;
 
+#if !CFG_CORE_SE05X_I2C_TRAMPOLINE
 	transfer = native_i2c_transfer;
 
 	return native_i2c_init();
+#else
+	/* not reached */
+	return 0;
+#endif
```
jforissier, line 60:
```suggestion
	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE)
		return 0;

	transfer = native_i2c_transfer;

	return native_i2c_init();
```

core/drivers/crypto/se050/session.c
```diff
@@ -123,7 +123,10 @@ static TEE_Result se050_session_init(void)
 	if (IS_ENABLED(CFG_CORE_SCP03_ONLY))
 		return se050_early_init_scp03();
 
-	return se050_early_init_default();
+	if (!IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE))
+		return se050_early_init_default();
+
+	return TEE_SUCCESS;
```
jforissier, line 129:
```suggestion
	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE)) {
		/* Nothing to do */
		return TEE_SUCCESS;
	}

	return se050_early_init_default();
```

jforissier (commented):
Hi @hoihochan,

Please see comments & suggestions below. Thanks.

core/drivers/crypto/se050/glue/i2c.c
hoihochan: Hmm. I think this will generate linker errors though, since only imx/stm32 has a native i2c implementation and has `native_i2c_init()` defined.

jforissier: Then perhaps `native_i2c_transfer()` and `native_i2c_init()` should to be stubbed in `core/drivers/crypto/se050/glue/include/i2c_native.h` when `CFG_CORE_SE05X_I2C_TRAMPOLINE` is enabled? @ldts

hoihochan:
all comments addressed.

core/drivers/crypto/se050/glue/include/i2c_native.h
```diff
@@ -13,4 +13,17 @@ TEE_Result native_i2c_transfer(struct rpc_i2c_request *req,
...
+{
+	/* Stub only */
+	return TEE_SUCCESS;
+}
+
+int native_i2c_init(void)
+{
+	/* Stub only */
+	return 0;
+}
```
jforissier, line 28: That's not how stubs are done, they need to be static inline, see other headers.
```
#if ...
void foo(void);
#else
static inline void foo(void)
{}
#endif
```

hoihochan: done

hoihochan:
all comments address and CI builds passed.

core/drivers/crypto/se050/glue/include/i2c_native.h
```diff
@@ -9,8 +9,25 @@
 
 #include <kernel/rpc_io_i2c.h>
 
+#if defined(CFG_CORE_SE05X_I2C_TRAMPOLINE) && CFG_CORE_SE05X_I2C_TRAMPOLINE
```
etienne-lms, line 12: `#ifdef CFG_CORE_SE05X_I2C_TRAMPOLINE` only

core/drivers/crypto/se050/glue/include/i2c_native.h
```diff
@@ -9,8 +9,25 @@
...
+static inline TEE_Result native_i2c_transfer(struct rpc_i2c_request *req,
+			       size_t *bytes)
+{
+	/* Stub only */
+	(void)req;
+	(void)bytes;
+	return TEE_SUCCESS;
```
etienne-lms, line 19: prefer:
```c
static inline TEE_Result
native_i2c_transfer(struct rpc_i2c_request *req __unused,
		    size_t *bytes __unused)
{
	return TEE_SUCCESS;
}
```

`return TEE_ERROR_NBOT_SUPPORTED` would be even better, unless some side effect on caller functions falling into the `__no_return` case.

core/arch/arm/plat-stm32mp1/conf.mk
```diff
@@ -411,3 +411,8 @@ CFG_STM32_DEBUG_ACCESS ?= $(CFG_TEE_CORE_DEBUG)
 ifeq ($(call cfg-all-enabled,CFG_STM32MP15 CFG_STM32MP13),y)
 $(error CFG_STM32MP13_CLK and CFG_STM32MP15_CLK are exclusive)
 endif
+
+# Disable I2C trampoline if native I2C is used
+ifeq ($(CFG_STM32_I2C),y)
+$(call force,CFG_CORE_SE05X_I2C_TRAMPOLINE,n)
```
etienne-lms, line 417: Not strictly need. The switch is not default enbled in `mk/config.mk`, right?

etienne-lms (commented):
Could you squash the commits the relate to the same goal?

core/arch/arm/plat-stm32mp1/conf.mk
hoihochan: It is default to `y` in `core/drivers/crypto/se050/crypto.mk`, which gets included via `core/core.mk`.

core/drivers/crypto/se050/glue/include/i2c_native.h
hoihochan: Can we undef a config option? Since`core/drivers/crypto/se050/crypto.mk` sets `CFG_CORE_SE05X_I2C_TRAMPOLINE` to `y`, it will show up in the generated `conf.h`.

core/arch/arm/plat-stm32mp1/conf.mk
etienne-lms: In _core/drivers/crypto/se050/crypto.mk_, `CFG_CORE_SE05X_I2C_TRAMPOLINE` defaults to `y` only when `CFG_NXP_SE05X` is enabled. So platforms not enabling `CFG_NXP_SE05X` should not bother with SE050 config directives.

core/drivers/crypto/se050/glue/include/i2c_native.h
etienne-lms: Boolean `CFG_xxx` config switches seen by source files are expected to be defined when the switch is enabled (`y`), and not defined when disabled (`n`).

If your platform needs to have `CFG_CORE_SE05X_I2C_TRAMPOLINE` disabled, it can simply use
`CFG_CORE_SE05X_I2C_TRAMPOLINE ?= n` or `$(call force,CFG_CORE_SE05X_I2C_TRAMPOLINE,n)`
in its _plat-xxx/conf.mk_ file
Or alternatively set the config from your build command line:
`make PLATFORM=... CFG_CORE_SE05X_I2C_TRAMPOLINE=n`.

core/drivers/crypto/se050/glue/include/i2c_native.h
```diff
@@ -9,8 +9,24 @@
 
 #include <kernel/rpc_io_i2c.h>
 
+#if defined(CFG_CORE_SE05X_I2C_TRAMPOLINE) && CFG_CORE_SE05X_I2C_TRAMPOLINE
+static inline TEE_Result native_i2c_transfer(
+				struct rpc_i2c_request *req __unused,
+				size_t *bytes __unused)
```
etienne-lms, line 15: for consistency with other header and source files, prefer line escape as below:
```c
static inline TEE_Result
native_i2c_transfer(struct rpc_i2c_request *req __unused,
		    size_t *bytes __unused)

File: core/drivers/crypto/se050/glue/include/i2c_native.h (hunk 5 above)
Comment (hoihochan) [reply core/drivers/crypto/se050/glue/include/i2c_native.h]: done

File: core/drivers/crypto/se050/glue/include/i2c_native.h (hunk 8 above)
Comment (hoihochan) [reply core/drivers/crypto/se050/glue/include/i2c_native.h]: done

**Review by ldts — COMMENTED** [review]
could you explain under which condition this could trigger?  https://github.com/OP-TEE/optee_os/pull/7041/commits/c137bf3c63ff24ec40e1a3159734390fbfaaad6f

**Comment by hoihochan** [comment]
> c137bf3

I have a platform that doesn't have native i2c driver inside OP-TEE, but we use the i2c trampoline and enables SCP once Linux fully boots up.

**Comment by ldts** [comment]
> > [c137bf3](https://github.com/OP-TEE/optee_os/commit/c137bf3c63ff24ec40e1a3159734390fbfaaad6f)
> 
> I have a platform that doesn't have native i2c driver inside OP-TEE, but we use the i2c trampoline and enables SCP once Linux fully boots up.

Sure but the fix is only required due to the modification of the time at which the session is being open.

Yes that is a configuration we validated but I never pushed those changes as there are other things to consider since the driver routes TEE requests to I2C and might affect the default configurations (ie, RNG)

In your case, is the REE (u-boot or linux) available before the TEE is loaded (I believe Chromium loads OP-TEE after Linux)   or does it come after OP-TEE ?

**Comment by ldts** [comment]
Another thing to consider is that some types of the SE, must have SCP03 enabled on session startup, so this new configuration - if enabled - would be breaking the support for those devices ( ie for instance the SE050F FIPS 140-2 certified device)

**Comment by hoihochan** [comment]
> > > [c137bf3](https://github.com/OP-TEE/optee_os/commit/c137bf3c63ff24ec40e1a3159734390fbfaaad6f)
> > 
> > 
> > I have a platform that doesn't have native i2c driver inside OP-TEE, but we use the i2c trampoline and enables SCP once Linux fully boots up.
> 
> Sure but the fix is only required due to the modification of the time at which the session is being open (ie your first commit). So one depends on the other I believe - just thinking out-loud :)
> 
> Yes that is a configuration we validated but I never pushed those changes as there are other things to consider since the driver routes TEE requests to I2C and might affect the default configurations (ie, RNG)
> 
> In your case, is the REE (u-boot or linux) available before the TEE is loaded (I believe Chromium loads OP-TEE after Linux) or does it come after OP-TEE ?

In my case, TrustedFirmware-A loads the OP-TEE prior to running Linux. So REE runs after OP-TEE.

**Comment by hoihochan** [comment]
> Another thing to consider is that some types of the SE, must have SCP03 enabled on session startup, so this new configuration - if enabled - would be breaking the support for those devices ( ie for instance the SE050F FIPS 140-2 certified device) - check the commit that introduced CFG_CORE_SCP03_ONLY

The SE05x part I have specifically require SCP03 enabled for most operations - however we don't have a native i2c driver inside OP-TEE, so we have to defer all i2c operations till Linux/REE runs.

File: core/drivers/crypto/se050/adaptors/apis/sss.c (hunk 1 above)
Comment (hoihochan) [reply core/drivers/crypto/se050/adaptors/apis/sss.c:165]: I have retested that since `se050_session_init` now calls the i2c stubs whenever `CFG_CORE_SE05X_I2C_TRAMPOLINE` is enabled, so this is no longer necessary but it is still a good defense-in-depth as `se050_enable_scp03` is an exported function.

Comment (hoihochan) [reply core/drivers/crypto/se050/adaptors/apis/sss.c:165]: There exists a path for REE to enable SCP:

```
PTA_CMD_ENABLE_SCP03
   -> crypto_se_enable_scp03
     -> se050_enable_scp03
```

Now, on systems without native I2C driver in OP-TEE that solely relies on I2C trampolines, `session` is going to be `NULL` initially until SCP is enabled and working. So attempts to 'close' the previous session will just result in null-pointer exception.

**Comment by ldts** [comment]
> > Another thing to consider is that some types of the SE, must have SCP03 enabled on session startup, so this new configuration - if enabled - would be breaking the support for those devices ( ie for instance the SE050F FIPS 140-2 certified device) - check the commit that introduced CFG_CORE_SCP03_ONLY
> 
> The SE05x part I have specifically require SCP03 enabled for most operations - however we don't have a native i2c driver inside OP-TEE, so we have to defer all i2c operations till Linux/REE runs.

yes, that is exactly how I brought up initially STM32. I think your PR is mostly right but I think we should have some more documentation and a slight change.

File: core/drivers/crypto/se050/glue/i2c.c [hunk 9]
```diff
@@ -47,7 +47,7 @@ int glue_i2c_write(uint8_t *buffer, int len)
 
 int glue_i2c_init(void)
 {
-	if (transfer == rpc_io_i2c_transfer)
+	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE))
```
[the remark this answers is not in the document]
ldts: sorry I dont understand your point. You are breaking the behaviour. 
With your change, if glue_i2c_init is called after we have exited the TEE and CFG_CORE_SE05X_I2C_TRAMPOLINE is not enabled then we will likely crash.

core/drivers/crypto/se050/adaptors/apis/sss.c
ldts: > Unrelated change. Please move to a separate commit.

this is not correct. The commits are related - it is not a bug but part of his initial implementation

core/drivers/crypto/se050/session.c
```diff
@@ -123,6 +123,11 @@ static TEE_Result se050_session_init(void)
 	if (IS_ENABLED(CFG_CORE_SCP03_ONLY))
 		return se050_early_init_scp03();
 
+	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE)) {
```
ldts, line 126: move this above the previous condition.

core/drivers/crypto/se050/glue/i2c.c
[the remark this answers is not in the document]
ldts: please keep the original condition and just OR it with the new one

ldts:
CFG_CORE_SE05X_I2C_TRAMPOLINE does not mean what you say in your commit. So please do not reuse it - for reference, CFG_CORE_SE05X_I2C_TRAMPOLINE means that the user is allowed to use the trampoline. If not set, it will continue to use the native driver.

You need a new configuration for the feature you are implementing.

core/drivers/crypto/se050/glue/i2c.c
ldts: sorry I am seeing all of the comments now. The issue is that usage of this configuration is being misinterpreted.  The configuration does what is supposed to do -and the implementation is correct.

What @hoihochan is after is a new configuration flag to postpone the creation of the session until the REE is ready to handle the I2C transfers. At that point, the session can be created using the PTA to configure the secure channel.

hoihochan:
Have refactored the code to use a new config option `CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY` to indicate the platform does not have any native I2C drivers and rely solely on REE I2C trampolines for accessing the SE05X.

That being said, `se050_session` will continue to be `NULL` (i.e. no valid SE05X session) prior to invoking `PTA_CMD_ENABLE_SCP03` on the SCP PTA from REE, so the second commit 4fca423 will still be necessary. Re-pasting the call flow below:

```
PTA_CMD_ENABLE_SCP03
   -> crypto_se_enable_scp03(bool rotate_keys)
     -> se050_enable_scp03(sss_se05x_session_t *session) <-- will crash if session is NULL
```

Feedbacks welcome.

ldts:
> Have refactored the code to use a new config option `CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY` to indicate the platform does not have any native I2C drivers and rely solely on REE I2C trampolines for accessing the SE05X.
> 
> That being said, `se050_session` will continue to be `NULL` (i.e. no valid SE05X session) prior to invoking `PTA_CMD_ENABLE_SCP03` on the SCP PTA from REE, so the second commit [4fca423](https://github.com/OP-TEE/optee_os/commit/4fca4234cb392b3ee26de54af8390685a7982566) will still be necessary. Re-pasting the call flow below:
> 
> ```
> PTA_CMD_ENABLE_SCP03
>    -> crypto_se_enable_scp03(bool rotate_keys)
>      -> se050_enable_scp03(sss_se05x_session_t *session) <-- will crash if session is NULL
> ```
> 
> Feedbacks welcome.

thanks for doing this @hoihochan , I'll have a look.

yes the second commit will be necessary but just as part of the first one (they are not independent, one causes the need for the other so the second one would be a fix for the first one... and since introducing a commit causing a segmentation fault doesn't make sense, it should be folded IMO)

ldts:
@jforissier both commits should be folded please

ldts:
@hoihochan could you reword the commit message please? 

drivers: se050: new CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY

Also maybe and something explicitly warning in the commit  that that when enabling this option OP-TEE must not use any of the resources provided by the secure element until the REE is up. This is because the secure element driver can present services that OP-TEE could try to use early.

The option CFG_CORE_SCP03_ONLY must still be supported so it requires to be added in the places where the session now could be open:

1) in crypto_se_enable_scp03 ()
2) in crypto_se_do_apdu()

Notice that the way to enable SCP03 is slightly different when the session didnt exist: you need to do as it is done in se050_early_init_scp03() so we can store the OEFID.

core/drivers/crypto/se050/adaptors/apis/sss.c
jforissier: > > Unrelated change. Please move to a separate commit.
> 
> this is not correct. The commits are related - it is not a bug but part of his initial implementation

My mistake. Let's have a single commit.

hoihochan:
> The option CFG_CORE_SCP03_ONLY must still be supported

By the way, how are we going to support `CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY` and `CFG_CORE_SCP03_ONLY` at the same time? 

OP-TEE wouldn't know REE is ready unless someone tells it (e.g. via a invoke to the SCP03 PTA to enable SCP).

ldts:
> Another thing I'd like to ask is why we fill the msg with `0x55`s in `se050_scp03_subkey_derive` when `CFG_CORE_SCP03_ONLY` is enabled?
> 
> ```
>         if (IS_ENABLED(CFG_CORE_SCP03_ONLY)) {
>                 memset(msg, 0x55, sizeof(msg));
>         } else {
>                 /* add some randomness */
>                 if (tee_otp_get_die_id(msg + 3, SE050_SCP03_KEY_SZ))
>                         return kStatus_SSS_Fail;
>         }
> ```
> 
> Don't we always want device-unique entropy?

the reason for that bit of code is because tee_otp_get_die_id could be calling the SE - and with SCP03_ONLY that would be a catch22 since the session has not been established at that point.

ldts:
> > The option CFG_CORE_SCP03_ONLY must still be supported
> 
> By the way, how are we going to support `CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY` and `CFG_CORE_SCP03_ONLY` at the same time?
> 
> OP-TEE wouldn't know REE is ready unless someone tells it (e.g. via an invoke to the SCP03 PTA to enable SCP).

CFG_CORE_SCP03_ONLY means that the SE can **only** be accessed from a secure session...so we can't establish the session and then enable SCP03 as we use in the default case.;  for CFG_CORE_SCP03_ONLY it doesn't matter whether we use the trampoline or the native i2c interface really

core/drivers/crypto/se050/core/apdu.c
```diff
@@ -16,6 +16,11 @@ TEE_Result crypto_se_do_apdu(enum crypto_apdu_type type,
 {
 	sss_status_t status = kStatus_SSS_Fail;
 
+	if (IS_ENABLED(CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY) && !se050_session) {
+		/* Defer until REE is ready */
```
ldts, line 20: we should open the session here instead of deferring

ldts: we should also open in **crypto_se_enable_scp03** (handle the open in that function as well); otherwise se050_enable_scp03 will break looking for the OEFID unless CFG_CORE_SE05X_OEFID is set which is not necessarily true.

ldts (changes requested):

core/drivers/crypto/se050/core/apdu.c
ldts: an to avoid racing while doing it - since we might be coming from a multi-threaded environment - please  use some form of exclusion mechanism

ldts: try calling se050_session_init() from both places - so no longer a static method - I think that should cover all cases. then when that CFG_CORE_SE05X_I2C_TRAMPOLINE_ONLY is enabled, dont let driver_init call the function

ldts:
@hoihochan just checking, do you plan to continue with this feature? it is not a lot of extra work on top of what you have done already

hoihochan:
> @hoihochan just checking, do you plan to continue with this feature? it is not a lot of extra work on top of what you have done already

Yes, sorry I was out on an extended vacation. Will get back to it next week.

ldts:
> Yes, sorry I was out on an extended vacation. Will get back to it next week.

awesome. I think there is value here

PR #7043, opened 2024-09-18
Miscellaneous fixes and improvements for RISC-V

gagachang:
- Commit "core: riscv: Set exception return PC into XEPC for entering user mode"
removes redundant usage of "ra" register.
- Commit "core: riscv: Fix misconfiguration of XSCRATCH when XRET to kernel mode"
fixes the value of XSCRATCH when XRET to kernel mode.
- Commit "core: riscv: Use sp as base register of load instructions"
uses "sp" as base register for load instructions to reduce the code size.

### DIFF
--- a/core/arch/riscv/kernel/thread_arch.c
+++ b/core/arch/riscv/kernel/thread_arch.c
@@ -502,7 +502,7 @@ static void set_ctx_regs(struct thread_ctx_regs *regs, unsigned long a0,
 		.a3 = a3,
 		.s0 = 0,
 		.sp = user_sp,
-		.ra = entry_func,
+		.epc = entry_func,
 		.status = status,
 		.ie = ie,
 	};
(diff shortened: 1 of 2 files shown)


gagachang:
Hello @jforissier , is there any feedback or acknowledgment for this PR and https://github.com/OP-TEE/optee_os/pull/7046 ?
We would appreciate it if there is any progress, thanks!

jenswi-linaro:
@jforissier, that makes sense. I was about to ack these just to let them through, but I think your approach is better.

gagachang:
> There are obviously less contributors to OP-TEE who run RISC-V than Arm and that is a bit problematic for PR reviews.

Yes.. We've also been aware of this problem. RISC-V needs more time to enlarge the ecosystem.

> We have a rule that any commit must have at least one Acked-by or Reviewed-by tag, preferably from someone who wasn't directly involved in the development of the patch. 

Agreed, the more feedback we got, the less bugs we will encounter in the future.

> So I suppose after a week or so has passed without any feedback from the community I can merge your patches.

LGTM! at least we can move forward to more patches.
Thank you! @jforissier @jenswi-linaro

etienne-lms:
@jforissier, I'm fine with your approach.

PR #7044, opened 2024-09-18
core: delay: fix timeout expiration threshold

etienne-lms:
Fix timeout threshold that should return once the time expires, not 1 tick after it has expired. This change makes udelay(0) to immediately return and not wait a generic timer counter increment.

This change fixes an issue where a driver stops the counter feeding Arm generic timer counter and indirectly calls IO_READ32_POLL_TIMEOUT() with a delay of 0us. It that case, since counter never increments, the udelay(0) call in IO_READ32_POLL_TIMEOUT() never returns while we expect the macro to endlessly poll (as timeout would never be detected) until poll condition is met or system watchdog is triggered.

### DIFF
--- a/core/include/kernel/delay.h
+++ b/core/include/kernel/delay.h
@@ -52,7 +52,7 @@ static inline uint64_t timeout_init_us(uint32_t us)
 
 static inline bool timeout_elapsed(uint64_t expire)
 {
-	return delay_cnt_read() > expire;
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

jforissier:

etienne-lms:
Commit message header line fixed and review tags applied.
Thanks.

PR #7048, opened 2024-09-20
Allow to seed software RNG without STM32 RNG peripheral

tbourgoi:
In case the node RNG is disabled in the device tree and we have compiled the driver stm32_rng.c, the function plat_rng_init() will try to read a random value from the RNG peripheral to seed the PRNG of OP-TEE but the peripheral is not initialized.
As a fallback, we could use the default implementation of plat_rng_init() to seed PRNG.

This pull request aims to rename the default implementation of plat_rng_init() to __plat_rng_init().
The second patch update the stm32_rng driver to fallback to __plat_rng_init() if the RNG peripheral is disabled in the DT.


jenswi-linaro (commented):

jforissier:

tbourgoi:
both tag applied on the 2 commits

etienne-lms (approved):

tbourgoi:
tag applied

PR #7052, opened 2024-09-23
plat-k3: set TRNG as enabled unconditionally

glneo:
All K3 devices already have this enabled, remove the check and set this unconditionally.


jforissier (approved):

etienne-lms (commented):
The commit message can be misleading IMHO. The RRNG is now default disabled for all k3 based platforms, not set unconditionally as external build config directives can still override the default config and enable PRNG.
That said,  `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` (I may be too picky :)

glneo:
Ping, seems this got closed due to inactivity..

etienne-lms (approved):
Looks like ready to be merged.

PR #7053, opened 2024-09-24
core: plat-corstone1000: Increase TZDRAM size

alioza0:
Increased TZDRAM size using space.

NS_SHARED_RAM region is not used by Corstone1000 platform. It is removed to create more space in secure RAM for BL32 image. Thus, there is more space in the secure ram that can be used by OP-TEE.


core/arch/arm/plat-corstone1000/conf.mk
```diff
@@ -34,7 +34,7 @@ CFG_TEE_CORE_NB_CORE ?= 1
 CFG_TZDRAM_START ?= 0x02002000
 
 # TEE_RAM (OPTEE kernel + DATA) + TA_RAM = 3MB
```
jforissier, line 36: Please fix the comment. I would remove the "= 3MB". And while you're at it: `s/OPTEE kernel/OP-TEE kernel/`.

jforissier (commented):
One comment.

core/arch/arm/plat-corstone1000/conf.mk
alioza0: Thank you.
I removed "= 3MB".
I fixed s/secure ram/secure RAM/ and s/OPTEE kernel/OP-TEE kernel/.

PR #7054, opened 2024-09-25
Ftpm

jenswi-linaro:
Imports  OP-TEE fTPM TA into optee_os/ta/ftpm folder.
Based on https://github.com/microsoft/ms-tpm-20-ref and 
https://github.com/zeschg/ms-tpm-20-ref.git branch feat_add_tee_crypto

Depends on https://github.com/OP-TEE/build/pull/780 and https://github.com/OP-TEE/manifest/pull/293


ta/ftpm/include/TEE/TpmToTEEHash.h
```diff
@@ -0,0 +1,194 @@
...
+//          Internal                    External
+//          Designation                 Designation
+// external name of the
+// initialization method
+#  define tpmHashStart_SHA1         TEE_InitSha1
+#  define tpmHashData_SHA1          TEE_DigestUpdateWrapper
+#  define tpmHashEnd_SHA1           TEE_FinalizeHash
+#  define tpmHashStateCopy_SHA1     memcpy
+#  define tpmHashStateExport_SHA1   memcpy
+#  define tpmHashStateImport_SHA1   memcpy
```
jenswi-linaro, line 162: This will not work as intended since only a TEE_OperationHandle is stored in the hash state. We'll be copying a pointer.
The GP TEE Internal Core API doesn't have a function to import or export the hash state. That's a problem since the TPM implementation seems to depend on this.

We will most likely need to use a different implementation for hashes.

jenswi-linaro (commented):
Comment for the commit "ftpm: import TEE crypto API wrappers"

etienne-lms:
Comments on the preliminary patches.

"mk/subdir.mk: refactor process-subdir-{srcs-y,gensrcs-helper}",
"ta_dev_kit.mk: use spec-srcs and spec-out-dir",
"mk: introduce global-incdirs_ext-y".

Some questions around commit "mk/subdir.mk: introduce srcs_ext-y and srcs_ext_base-y".

mk/subdir.mk
```diff
@@ -90,36 +90,7 @@ $2: $$(depends-$1)
...
-cxxflags-lib-y			:=
-cppflags-$(1)-y			:=
-cppflags-remove-$(1)-y		:=
-cppflags-lib-y			:=
-aflags-$(1)-y 			:=
-aflags-remove-$(1)-y		:=
-incdirs-$(1)-y			:=
-fname				:=
-oname				:=
-
```
jforissier, line 122: "mk/subdir.mk: refactor process-subdir-{srcs-y,gensrcs-helper}":
That is probably OK but would be better as a separate commit

jforissier (commented):
For "ta_dev_kit.mk: use spec-srcs and spec-out-dir":

jforissier (commented):
For "mk: introduce global-incdirs_ext-y":

jforissier (commented):
For "mk/subdir.mk: introduce srcs_ext-y and srcs_ext_base-y":

mk/subdir.mk
jenswi-linaro: What do you mean?

jforissier: Please ignore this comment. It was a mistake.

jforissier (commented):
"mk/subdir.mk: refactor process-subdir-{srcs-y,gensrcs-helper}":

mk/subdir.mk
```diff
@@ -49,23 +49,44 @@ cppflags-lib-y			:=
...
+oname		:= $(out-dir)/$(base-prefix)$(basename $$(sub-dir)/$1).o
 endif
 endif
-$$(eval $$(call process-file-vars,$1,$$(oname)))
+$$(eval $$(call process-file-vars,$1,$$(oname),$$(fname)))
+endef #process-subdir-srcs-y
+
+define process-subdir-srcs_ext-y
+ifneq ($(filter /%,$(1)),)
+$$(error Absolute path not supported for srcs_ext-y: $(1)
```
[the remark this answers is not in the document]
jenswi-linaro: Thanks, I'll fix.

mk/subdir.mk
```diff
@@ -123,9 +144,13 @@ endif
 # Process files in current directory
 $$(foreach g, $$(gensrcs-y), $$(eval $$(call process-subdir-gensrcs-y,$$(g))))
 $$(foreach s, $$(srcs-y), $$(eval $$(call process-subdir-srcs-y,$$(s))))
+$$(foreach s, $$(srcs_ext-y), $$(eval $$(call \
+	process-subdir-srcs_ext-y,$$(s),$$(firstword $$(srcs_ext_base-y)))))
```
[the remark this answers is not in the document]
jenswi-linaro: What is the best action if `srcs_ext_base-y` has more than one word? Error out or ignore it?

mk/subdir.mk
```diff
@@ -49,23 +49,44 @@ cppflags-lib-y			:=
...
 aflags-remove-$(1)-y		:=
 incdirs-$(1)-y			:=
-endef #process-file-vars
+endef #process-file-vars-helper
+
+define process-file-vars
+# $1 is local source file name
+# $2 is output file name
+# $3 is tree source file name
+oname-$(sm)-$3	:= $$(if $$(oname-$1-y),$(out-dir)/$(base-prefix)/$$(oname-$1-y),$2)
```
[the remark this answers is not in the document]
jenswi-linaro: It's supposed to be set in the `sub.mk` file if used.

ta/ftpm/include/ftpm_ta.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef FTPM_TA_H
+#define FTPM_TA_H
+
+/* This UUID is generated with uuidgen */
+#define TA_FTPM_UUID { 0xBC50D971, 0xD4C9, 0x42C4, \
+	{0x82, 0xCB, 0x34, 0x3F, 0xB7, 0xF3, 0x78, 0x96}}
+
+/* The TAFs ID implemented in this TA */
+#define TA_FTPM_SUBMIT_COMMAND  (0)
+#define TA_FTPM_EMULATE_PPI     (1)
```
b49020, line 49: AFAICS, `TA_FTPM_EMULATE_PPI` isn't being used from Linux kernel driver point of view. Also, it seems that Physical Presence Interface (PPI) specification currently implemented for ACPI based systems. So we should rather drop corresponding source code import until we see a real use-case upstream for OP-TEE based DT systems.

b49020 (commented):
Just few overall code import comments.

ta/ftpm/Makefile
```diff
@@ -0,0 +1,13 @@
+BINARY=bc50d971-d4c9-42c4-82cb-343fb7f37896
+
+O ?= ../out/fTPM
+WOLF_ROOT := ../../../../external/wolfssl/
+TPM_ROOT := ../../../../
```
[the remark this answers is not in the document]
jenswi-linaro: Thanks, I'll fix.

ta/ftpm/include/ftpm_ta.h
jenswi-linaro: The primary target is OP-TEE based DT systems and the only thing we can realistically test for. However, I'm not so keen on removing features from fTPM since this is supposed to be the new upstream for the OP-TEE TA replacing the old deprecated location.

b49020: Agree and the git history of unmodified fTPM source import should be easily retrievable. The features which can't be tested and an interface is provided to normal world looks like a security concern to me. To be able to deploy fTPM for production use-cases, we really need to keep required bits enabled. At the very least you can disable this unused fTPM invoke command.

jenswi-linaro: That makes sense, I'll add a configuration variable.

jenswi-linaro:
We need to find a way of testing that this code works as intended. Both for the current use case, but also that the TPM does what it should with other currently unused interfaces or features.

b49020:
> We need to find a way of testing that this code works as intended. Both for the current use case, but also that the TPM does what it should with other currently unused interfaces or features.

Agree, I think measured boot is something we should at-least enable for our Qemu setup in the CI based on fTPM. Apart from that we can write basic user-space tests using tpm2-tools to perform signing/verification. There is also tpm2-selftest (https://manpages.debian.org/testing/tpm2-tools/tpm2_selftest.1.en.html) which can be really handy for testing fTPM.

We can also create tests for tpm2 provider for openssl (https://github.com/tpm2-software/tpm2-openssl) which will directly relate to real world use-cases.

jenswi-linaro:
Thanks, I was looking for something like tpm2_selftest.

jenswi-linaro:
tpm2_selftest is passing and with logging enabled, it excercises the bignum implementation which is my greatest concern. I think we have what's needed to prepare for a proper review. My plan is as follows:
1. apply tags already provided
2. drop the hash part from "ftpm: import TEE crypto API wrappers"
3. refactor "ftpm: import TEE crypto API wrappers" to better follow our coding style
4. fix TODOs etc in "ftpm: import TEE crypto API wrappers"

This should make "ftpm: import TEE crypto API wrappers" easier to review.

ta/ftpm/include/TEE/TpmToTEEMath.h
```diff
@@ -1,70 +1,72 @@
...
  * Portions Copyright Microsoft Corporation, see below for details:
  *
- *  The copyright in this software is being made available under the BSD License,
- *  included below. This software may be subject to other third party and
- *  contributor rights, including patent rights, and no such rights are granted
- *  under this license.
+ * The copyright in this software is being made available under the BSD
+ * License, included below. This software may be subject to other third
+ * party and contributor rights, including patent rights, and no such
+ * rights are granted under this license.
```
b49020, line 15: Can we just drop the license text which is covered by `SPDX-*` here and other places? The copyrights should be retained though.

b49020:
> Should I squash "ftpm: fixup TEE crypto API wrappers" into "ftpm: import TEE crypto API wrappers" or keep it as is?

Yeah that makes sense to me.

> The commit "ta: ftpm: import initial source" is quite large and it's not reasonable to review it in detail. This code has been used for fTPM until now so we should be able to rely on previous reviews in that repository.

That's reasonable, it's going to take time to make this fTPM import production ready with more stress testing and catching any corner cases. But we should go ahead with the initial merge after fixing obvious things.

b49020:
BTW, do you plan to add CI tests to OP-TEE CI for fTPM? `tpm2_selftest` can be a good candidate to start with.

ta/ftpm/include/TEE/TpmToTEEMath.h
jenswi-linaro: I don't know if that's possible. The last time I checked, it deviated slightly from the license by `SPDX-License-Identifier: BSD-2-Clause`.

mk/subdir.mk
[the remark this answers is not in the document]
etienne-lms: I wonder. It would be a programming error, indeed, and expended sources would not be build likely leading to some linkage error.
I guess ignoring it here is enough.

mk/subdir.mk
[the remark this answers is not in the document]
etienne-lms: Not easy to guess. If there was an example in the inline description above, it could maybe help.

etienne-lms (commented):
Regarding preparatory commits
* `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits
"tee_api_types.h: default alignment for TEE_BigIntFMMContext" and
"libmbedtls: mbedtls_config_uta.h: enable SHA-384 and SHA-512 support".

* Still few comments for commit "mk/subdir.mk: introduce srcs_ext-y and srcs_ext_base-y".

ta/ftpm/include/TEE/TpmToTEEMath.h
b49020: IMO, we should have clarity regarding any deviation from standard BSD-2-Clause license with the authors. Otherwise it will be hard for users to understand licensing requirements.

mk/subdir.mk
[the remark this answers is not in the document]
etienne-lms: Discard my last comment. I missed to look at the commit message that defines how `oname-<filename>-y` should be used.

etienne-lms (commented):
"mk/subdir.mk: introduce srcs_ext-y and srcs_ext_base-y".

ta/ftpm/include/TEE/TpmToTEEMath.h
jenswi-linaro: IANAL but for reasonable and practical purposes this is a BSD-2-Clause. I don't know what
```
 * The copyright in this software is being made available under the BSD
 * License, included below. This software may be subject to other third
 * party and contributor rights, including patent rights, and no such
 * rights are granted under this license.
```
is supposed to add. Perhaps it's only a lawyer trying to motivate its existence by causing troubles like this.
However, removing the license below affects the meaning of this text.

I propose that we keep this for now and if there's interest later we can try to remove it. fTPM has already been used with this license so we're not changing anything.

b49020: Interesting and I see a corresponding patent already filed by Microsoft here: https://patents.google.com/patent/WO2013019369A1/en. IANAL too but this license text infers that there ins't any royalty free grant provided which is usually provided by a `BSD-2-Clause Plus Patent License` (https://spdx.org/licenses/BSD-2-Clause-Patent.html).

So I am not sure if bundling fTPM source within optee-os repo enforces that patent restriction on optee-os source/binary distribution as well. IMHO, we really need to seek lawyer's advice or consult with Microsoft. However, if we create a separate fTPM TA repo within OP-TEE github project then optee-os source may remain independent of any such restrictions.

jenswi-linaro: > Interesting and I see a corresponding patent already filed by Microsoft here: https://patents.google.com/patent/WO2013019369A1/en. IANAL too but this license text infers that there ins't any royalty free grant provided which is usually provided by a `BSD-2-Clause Plus Patent License` (https://spdx.org/licenses/BSD-2-Clause-Patent.html).

My interpretation of the quoted text is the opposite. They emphasize that they don't grant patent rights. But that's already implied by the `BSD-2-Clause` license, so the text is generally redundant.

> So I am not sure if bundling fTPM source within optee-os repo enforces that patent restriction on optee-os source/binary distribution as well. IMHO, we really need to seek lawyer's advice or consult with Microsoft. 

I'd say that it's not our problem.

> However, if we create a separate fTPM TA repo within OP-TEE github project then optee-os source may remain independent of any such restrictions.

It does not make a difference from a license point of view if files are in the same repository.

jenswi-linaro:
> BTW, do you plan to add CI tests to OP-TEE CI for fTPM? `tpm2_selftest` can be a good candidate to start with.

We should do something to interact with fTPM. It seems that it does some self-testing when being probed so just checking for the device might be enough. On the other hand, we may want to have the tpm2-tools in the rootfs regardless so we might as well use `tpm2_selftest`.

ta/ftpm/include/TEE/TpmToTEEMath.h
b49020: > My interpretation of the quoted text is the opposite.

Lets leave that interpretation for lawyers or better consult with Microsoft for clarification. @jan-kiszka might have some clarity on this since they may be enabling fTPM in their products.

> > So I am not sure if bundling fTPM source within optee-os repo enforces that patent restriction on optee-os source/binary distribution as well. IMHO, we really need to seek lawyer's advice or consult with Microsoft.
> 
> I'd say that it's not our problem.

It will become a problem for downstream (like OE distro etc.) consumers of `optee-os` where just a version uprev will add a possible license restriction even if they aren't enabling fTPM. We at least need to update `optee-os` license accordingly such that it becomes known to consumers.

> > However, if we create a separate fTPM TA repo within OP-TEE github project then optee-os source may remain independent of any such restrictions.
> 
> It does not make a difference from a license point of view if files are in the same repository.

It will make a difference for consumers who choose to not enable fTPM in their products due to license constraints.

jenswi-linaro: >It will become a problem for downstream (like OE distro etc.) consumers of optee-os where just a version uprev will add a possible license restriction even if they aren't enabling fTPM. We at least need to update optee-os license accordingly such that it becomes known to consumers.

The main OP-TEE license is BSD-2-Clause, but there are exceptions. We don't have a uniform license that can be updated. We would in particular not want to add this since it only causes trouble when trying to classify the license we are just experiencing.

> > > However, if we create a separate fTPM TA repo within OP-TEE github project then optee-os source may remain independent of any such restrictions.
> > 
> > 
> > It does not make a difference from a license point of view if files are in the same repository.
> 
> It will make a difference for consumers who choose to not enable fTPM in their products due to license constraints.

Why? If the files aren't used why does it matter if they exist?

b49020: I found an interesting read about "Patents in Open Source" here [1], please have a look at it. I hope it helps to clarify licensing concerns I am trying to raise here.

[1] https://google.github.io/opencasebook/patents/#express-but-non-specific-licenses

jenswi-linaro:
We have decided to keep fTPM in a separate git instead. https://github.com/OP-TEE/optee_os/pull/7090 and https://github.com/OP-TEE/optee_ftpm/pull/1 replace this PR.

PR #7056, opened 2024-09-26
add rk3588 support

n1ptune:
support to rockchip rock5b which use rk3588

### DIFF
--- a/core/arch/arm/plat-rockchip/platform_config.h
+++ b/core/arch/arm/plat-rockchip/platform_config.h
@@ -85,6 +85,25 @@
 #define FIREWALL_DDR_BASE	0xff534000
 #define FIREWALL_DDR_SIZE	SIZE_K(16)
 
+#elif defined(PLATFORM_FLAVOR_rk3588)
+
+#define GIC_BASE		0xFE600000
+#define GIC_SIZE		SIZE_K(64)
+#define GICD_BASE		GIC_BASE
+#define GICC_BASE		0
+
+#define UART1_BASE		0xfeb40000
+#define UART1_SIZE		SIZE_K(64)
+
+#define UART2_BASE		0xfeb50000
+#define UART2_SIZE		SIZE_K(64)
+
+#define UART5_BASE		0xfeb80000
+#define UART5_SIZE		SIZE_K(64)
+
+#define FIREWALL_DDR_BASE	0xfe030000
+#define FIREWALL_DDR_SIZE	SIZE_K(32)
+
 #else
 #error "Unknown platform flavor"
 #endif
--- a/core/arch/arm/plat-rockchip/platform_rk3588.c
+++ b/core/arch/arm/plat-rockchip/platform_rk3588.c
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2019, Theobroma Systems Design und Consulting GmbH
+ */
+
+#include <common.h>
+#include <io.h>
+#include <kernel/panic.h>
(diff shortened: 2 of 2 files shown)


core/arch/arm/plat-rockchip/platform_config.h
```diff
@@ -85,6 +85,25 @@
 #define FIREWALL_DDR_BASE	0xff534000
 #define FIREWALL_DDR_SIZE	SIZE_K(16)
 
+#elif defined(PLATFORM_FLAVOR_rk3588)
```
jforissier, line 88: Having a SoC as the name of the flavor is not a good idea IMO. How would you name different boards using the same SoC? Although yes, ideally if we use DT and runtime detection etc. it may be possible but it doesn't seem to apply here.

jforissier (commented):
Please add some description to the commit and also your `Signed-off-by:`. And see my comment below. Thanks!

jforissier:
This is very similar to https://github.com/OP-TEE/optee_os/pull/7059. @n1ptune can we close this one? But, back to my remark on the platform name, how would one build for Rock 5 or NanoPC-T6 if the OP-TEE `PLATFORM=` is the same (i.e., `rockchip-rk3588`)?

edtubbs:
> This is very similar to #7059. @n1ptune can we close this one? But, back to my remark on the platform name, how would one build for Rock 5 or NanoPC-T6 if the OP-TEE `PLATFORM=` is the same (i.e., `rockchip-rk3588`)? @edtubbs

It seems like the `build` system handles this with the makefiles for specific boards, but we could extend the platform (e.g., `PLATFORM_FLAVOR_rk3588_rock5`) if board-level configuration is needed.

core/arch/arm/plat-rockchip/platform_config.h
```diff
@@ -85,6 +85,25 @@
...
+#define UART1_BASE		0xfeb40000
+#define UART1_SIZE		SIZE_K(64)
+
+#define UART2_BASE		0xfeb50000
+#define UART2_SIZE		SIZE_K(64)
+
+#define UART5_BASE		0xfeb80000
+#define UART5_SIZE		SIZE_K(64)
+
+#define FIREWALL_DDR_BASE	0xfe030000
```
DaniilKl, line 104: I see rk3588 has two firewalls: `FIREWALL_DDR` with base address `0xFE030000` and `FIREWALL_SYSMEM` with base address `0xFE038000` have you been able to find out the difference? I guess the `FIREWALL_DDR` hold MMU registers, and `FIREWALL_SYSMEM` holds SMMU registers, but I cannot confirm this.

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -0,0 +1,52 @@
...
+#define FIREWALL_DDR_FW_DDR_MST(i)	(0x20 + (i) * 0x4)
+#define FIREWALL_DDR_FW_DDR_CON_REG	0x40
+#define FIREWALL_DDR_FW_DDR_RGN_NUM	8
+#define FIREWALL_DDR_FW_DDR_MST_NUM	6
+
+#define RG_MAP_SECURE(top, base)	((((top) - 1) << 16) | (base))
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, FIREWALL_DDR_BASE, FIREWALL_DDR_SIZE);
+
+int platform_secure_ddr_region(int rgn, paddr_t st, size_t sz)
```
DaniilKl, line 23: How have you been able to find out how to set MMU configuration for this SoC? I cannot find any exact MMU register mappings in SoC's TRM.

PR #7057, opened 2024-09-27
More improvements for RISC-V

gagachang:
- Commit "core: riscv: Ensure XSTATUS is restored before XIE"
avoids accidental interrupt during register restoring.
- Commit "core: riscv: Improve thread user mode record"
improves register save/restore during entering/existing user mode.


gagachang:
Hi @jforissier , It's been a week since we submitted this PR.
I think the PR can be merged if there are no any comments.

jforissier:
> Hi @jforissier , It's been a week since we submitted this PR. I think the PR can be merged if there are no any comments.

I agree. Thanks!

PR #7060, opened 2024-09-30
notif: fix build when NS_VIRTIALIZATION is enabled

lorc:
Right now OP-TEE build fails if CFG_NS_VIRTUALIZATION=y and CFG_CORE_ASYNC_NOTIF=n with the following error:

core/kernel/notif.c: In function 'nex_init_notif': core/kernel/notif.c:185:42: error: 'notif_data_id' undeclared (first use in this function); did you mean 'notif_wait'?
  185 |         return virt_add_guest_spec_data(&notif_data_id,
      |                                          ^~~~~~~~~~~~~
      |                                          notif_wait
core/kernel/notif.c:185:42: note: each undeclared identifier is reported only once for each function it appears in
core/kernel/notif.c:186:48: error: invalid application of 'sizeof' to incomplete type 'struct notif_data'
  186 |                                         sizeof(struct notif_data), NULL);
      |                                                ^~~~~~
core/kernel/notif.c:187:1: warning: control reaches end of non-void function [-Wreturn-type]
  187 | }
      | ^

Move `#ifdef CFG_NS_VIRTUALIZATION` section under
`#ifdef CFG_CORE_ASYNC_NOTIF` to fix this.

### DIFF
--- a/core/kernel/notif.c
+++ b/core/kernel/notif.c
@@ -154,6 +154,16 @@ out:
 	mutex_unlock(&notif_mutex);
 	virt_put_guest(prtn);
 }
+
+#ifdef CFG_NS_VIRTUALIZATION
+static TEE_Result nex_init_notif(void)
+{
+	return virt_add_guest_spec_data(&notif_data_id,
+					sizeof(struct notif_data), NULL);
+}
+nex_early_init(nex_init_notif);
+#endif
+
 #endif /*CFG_CORE_ASYNC_NOTIF*/
 
 static TEE_Result notif_rpc(uint32_t func, uint32_t value1, uint32_t value2)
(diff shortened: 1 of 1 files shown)


jforissier (commented):
With `s/NS_VIRTIALIZATION/NS_VIRTUALIZATION/` in the commit subject:

jenswi-linaro:

lorc:
Thanks. I have updated the commit message subject and added tags.

Looks like checkpatch is unhappy because included error message is too long.

jforissier:
> Thanks. I have updated the commit message subject and added tags.
> 
> Looks like checkpatch is unhappy because included error message is too long.

Yes, but let's ignore it. I believe checkpatch needs to be updated to better deal with these kind of messages with line numbers and pipe characters.

PR #7064, opened 2024-10-02
Add TZC400 device tree bindings + configure STM32 TZC400 using the device tree on stm32mp1 platforms

GseoC:
This P-R adds a device tree binding file to be able to describe the TZC400 memory regions with their firewall attributes configuration in the device tree.

Refactor the stm32 platform TZC400 driver to support this feature and add the DT configuration of a memory region on the stm32mp135f-dk platform.

Note: I tried to make a generic TZC400 binding file so that it can be used by other platforms, let me know what you think of it. These "bindings" are more like macros to be used in the DT.


gagachang:
Hi @GseoC , may I know why the permissions are embedded in `optee_framebuffer` node with `st,protreg` property ?

Another way is to embed them  into `tzc400` node, like:

```diff
&tzc400 {
-	memory-region = <&optee_framebuffer>;
+	memory-region = <&optee_framebuffer DT_TZC_REGION_S_RDWR 0>;
};
```
Any special consideration for this ? Thanks!

GseoC:
> Hi @GseoC , may I know why the permissions are embedded in `optee_framebuffer` node with `st,protreg` property ?
> 
> Another way is to embed them into `tzc400` node, like:
> 
> ```diff
> &tzc400 {
> -	memory-region = <&optee_framebuffer>;
> +	memory-region = <&optee_framebuffer DT_TZC_REGION_S_RDWR 0>;
> };
> ```
> 
> Any special consideration for this ? Thanks!

Hi @gagachang,

I cannot do this because the memory-region dt-bindings file states that the content of the property is a phandle array : [https://github.com/devicetree-org/dt-schema/blob/main/dtschema/schemas/reserved-memory/memory-region.yaml](url). Each phandle is a /reserved-memory child node assigned to the device.

For now I have to use a proprietary property- but I hope that I'll be able to soon use a generic one that is a variant of `access-controllers` to be able to describe firewall configurations in the device tree. I've proposed a patch at Linux upstream: [https://lkml.org/lkml/2024/7/16/849](url) but since I'm alone (for now) to push for this, maintainers aren't keen on replying fast...

gagachang:
Hi @GseoC , Thanks for information!

> I cannot do this because the memory-region dt-bindings file states that the content of the property is a phandle array : [memory region YAML](https://github.com/devicetree-org/dt-schema/blob/main/dtschema/schemas/reserved-memory/memory-region.yaml). Each phandle is a /reserved-memory child node assigned to the device.

I am trying to understand the `access-controllers` property.
IIUC, the `memory-region` property is only used when the phandle is a /reserved-memory child node.
If the phandle is not a /reserved-memory child node (e.g., uart5 in your [Patch](https://lkml.org/lkml/2024/7/16/849)), we should use the `access-controllers` property instead. Am I correct?

> For now I have to use a proprietary property- but I hope that I'll be able to soon use a generic one that is a variant of `access-controllers` to be able to describe firewall configurations in the device tree. I've proposed a patch at Linux upstream: [Patch](https://lkml.org/lkml/2024/7/16/849) but since I'm alone (for now) to push for this, maintainers aren't keen on replying fast...

I saw your https://github.com/OP-TEE/optee_os/pull/7066.  Is `st,decprot` the proprietary property you mentioned ?

```diff
+ &etzpc {
+	st,decprot =
+		<&etzpc DECPROT(STM32MP1_ETZPC_USART1_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_SPI6_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_I2C4_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_I2C6_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_RNG1_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_HASH1_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_CRYP1_ID, DECPROT_NS_RW, DECPROT_UNLOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_DDRCTRL_ID, DECPROT_NS_R_S_W, DECPROT_LOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_DDRPHYC_ID, DECPROT_NS_R_S_W, DECPROT_LOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_STGENC_ID, DECPROT_S_RW, DECPROT_LOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_BKPSRAM_ID, DECPROT_S_RW, DECPROT_LOCK)>,
+		<&etzpc DECPROT(STM32MP1_ETZPC_IWDG1_ID, DECPROT_S_RW, DECPROT_LOCK)>;
+ };
```

GseoC:
> I am trying to understand the `access-controllers` property.
> IIUC, the `memory-region` property is only used when the phandle is a /reserved-memory child node.
> If the phandle is not a /reserved-memory child node (e.g., uart5 in your [Patch](https://lkml.org/lkml/2024/7/16/849)), we should use the `access-controllers` property instead. Am I correct?

The `access-controllers` property is basically used to link an access controller(that can be a firewall controller) to its consumers. Its consumers can either be subnodes or completely separate nodes in the device tree. It should be used when a device needs a service from an access controller(In OP-TEE case, a firewall controller) on a resource that is under its responsibility.

As for the firewall configuration, for now, there's no way to describe a firewall controller configuration with a generic property. Hence my effort on the above linux patch.

> I saw your https://github.com/OP-TEE/optee_os/pull/7066. Is st,decprot the proprietary property you mentioned ?

The `st,decprot` property is used to describe ETZPC firewall boot configuration. It is a proprietary property and only makes sense for STMicroelectronics' ETZPC driver implementation. If I could use a generic property: something like `default-access-config` I would :) . And i would try to implement ops in the OP-TEE firewall framework to be able to set a firewall configuration. [See this P-R](https://github.com/OP-TEE/optee_os/pull/6946)

GseoC:
> > I am trying to understand the `access-controllers` property.
> > IIUC, the `memory-region` property is only used when the phandle is a /reserved-memory child node.
> > If the phandle is not a /reserved-memory child node (e.g., uart5 in your [Patch](https://lkml.org/lkml/2024/7/16/849)), we should use the `access-controllers` property instead. Am I correct?
> 
> The `access-controllers` property is basically used to link an access controller(that can be a firewall controller) to its consumers. Its consumers can either be subnodes or completely separate nodes in the device tree. It should be used when a device needs a service from an access controller(In OP-TEE case, a firewall controller) on a resource that is under its responsibility.
> 
> As for the firewall configuration, for now, there's no way to describe a firewall controller configuration with a generic property. Hence my effort on the above linux patch.
> 
> > I saw your #7066. Is st,decprot the proprietary property you mentioned ?
> 
> The `st,decprot` property is used to describe ETZPC firewall boot configuration. It is a proprietary property and only makes sense for STMicroelectronics' ETZPC driver implementation. If I could use a generic property: something like `default-access-config` I would :) . And i would try to implement ops in the OP-TEE firewall framework to be able to set a firewall configuration. [See this P-R](https://github.com/OP-TEE/optee_os/pull/6946)

(STM32 use-case) For our platforms, we may need to perform some firewall operations before using a peripheral to be able to access it. See in `core/drivers/firewall/stm32_rifsc.c`, the function `stm32_rifsc_dt_probe_bus()`. We do lighter checks for the ETZPC (Only check that the peripheral is secure before using it in OP-TEE)

gagachang:
Thank you @GseoC ! 
I am working on RISC-V access protection controller.
We may need similar way to configure the registers of access protection controller using the device tree.
We will refer to your effort. Best wishes on your patch.

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+	unsigned int i = 0;
+	int len = 0;
+
+	conf_list = fdt_getprop(fdt, node, "memory-region", &len);
+	if (!conf_list)
+		return TEE_SUCCESS;
+
+	nregions = len / sizeof(uint32_t);
+	if (nregions > tzc_dev->ddata->nb_regions)
+		return TEE_ERROR_BAD_PARAMETERS;
```
etienne-lms, line 317: I suggest here a debug or error message: 
`EMSG("Too many regions defined in %s", fdt_get_name(fdt, node, NULL))`

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+		if (!IS_PAGE_ALIGNED(region_base) ||
+		    !IS_PAGE_ALIGNED(region_size))
+			return TEE_ERROR_BAD_PARAMETERS;
+
+		region_cfg.base = region_base;
+		region_cfg.top = region_base + region_size - 1;
+		region_cfg.filters = filter_mask(tzc_dev->ddata->nb_filters);
+
+		prop = fdt_getprop(fdt, pnode, "st,protreg", &len);
+		if (!prop || len != (2 * sizeof(uint32_t)))
```
etienne-lms, line 344: needs a cast IMHO (compare signed and unsigned value)

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+	if (base.pa == DT_INFO_INVALID_REG)
+		return TEE_ERROR_BAD_PARAMETERS;
 
-	tzc_init((vaddr_t)base);
-	tzc_dump_state();
+	reg_size = fdt_reg_size(fdt, node);
+	if (reg_size == DT_INFO_INVALID_REG_SIZE)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	tzc_dev->pdata.name = fdt_get_name(fdt, node, NULL);
```
etienne-lms, line 420: Please use `strdup()` to not reference the DT at runtime.

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+	tzc_dev->pdata.base = io_pa_or_va_secure(&base, reg_size);
+
+	res = interrupt_dt_get(fdt, node, &tzc_dev->pdata.itr_chip,
```
etienne-lms, line 423: FYI: I can be more efficient to get al deferrable resources first before digging into the DT for other information.
Here: first get the clock and interrupt resources, then get IOMEM base/size, name ...

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+			       &tzc_dev->pdata.itr_num);
+	if (res)
+		return res;
+
+	res = clk_dt_get_by_index(fdt, node, 0, tzc_dev->pdata.clk);
+	if (res)
+		return res;
+
+	res = clk_dt_get_by_index(fdt, node, 1, tzc_dev->pdata.clk + 1);
+	if (res || !tzc_dev->pdata.clk[1])
```
etienne-lms, line 433: Would be more strict with:
```c
	res = clk_dt_get_by_index(fdt, node, 1, tzc_dev->pdata.clk + 1);
	if (res == TEE_ERROR_ITEM_NOT_FOUND)
		DMSG("No secondary clock for %s",
	else if (res)
		return res;

File: core/arch/arm/plat-stm32mp1/plat_tzc400.c [hunk 6]
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+		DMSG("No secondary clock for %s",
+		     fdt_get_name(fdt, node, NULL));
+
+	/* Use memory node instead of that new one */
```
etienne-lms, line 437: I find this inline comment a bit confusing. I suggest to remove it.

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+			      sizeof(*tzc_dev->reg));
+	if (!tzc_dev->reg)
+		panic();
+
+	if (clk_enable(tzc_dev->pdata.clk[0]))
+		panic();
+	if (tzc_dev->pdata.clk[1] && clk_enable(tzc_dev->pdata.clk[1]))
+		panic();
+
+	tzc_init((vaddr_t)tzc_dev->pdata.base);
```
etienne-lms, line 501: once `base` isof `vaddr_t`, remove this cast.

GseoC:
> Thank you @GseoC ! I am working on RISC-V access protection controller. We may need similar way to configure the registers of access protection controller using the device tree. We will refer to your effort. Best wishes on your patch.

Good to hear that we share that need, it makes more sense to find a proper generic solution.

@gagachang, I'll send another version of the other linux patch soon. Do you want to be put as To/CC of the patch so you can give your opinion or state that you may be interested in it?

gagachang:
> @gagachang, I'll send another version of the other linux patch soon. Do you want to be put as To/CC of the patch so you can give your opinion or state that you may be interested in it?

Sure! Please CC that patch to my email: `alvinga@andestech.com`. Thanks!

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+					      reg_exclude->top -
+					      reg_exclude->base + 1,
+					      reg->region.base,
+					      reg->region.top -
+					      reg->region.base + 1);
+		if (found)
+			break;
+	}
+
+	if (!found || !reg)
```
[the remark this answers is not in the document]
GseoC: I think it explicit to keep the found variable. I'll remove reg as found will be false if no hit while ending the loop.

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+	} else if (reg_exclude->base == reg->region.base) {
+		reg->region.base = reg_exclude->top + 1;
+	} else if (reg_exclude->top == reg->region.top) {
+		reg->region.top = reg_exclude->base - 1;
+	} else {
+		struct tzc_region_non_sec *new_nsec =
+			calloc(1, sizeof(*new_nsec));
+
+		if (!new_nsec || !reg)
+			panic();
```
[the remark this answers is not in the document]
GseoC: Fair

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +86,468 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+	struct tzc_device *tzc_dev =
+		(struct tzc_device *)PM_CALLBACK_GET_HANDLE(hdl);
+
+	if (op == PM_OP_RESUME) {
+		stm32mp_tzc_region0(true);
+
+		stm32mp_tzc_reset_region(tzc_dev);
+
+		for (i = 0; i < tzc_dev->nb_reg_used; i++)
+			tzc_configure_region(i + 1, &tzc_dev->reg[i]);
```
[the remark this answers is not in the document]
GseoC: I've implemented you comment but I wonder if we should really default to NSEC for the region 0 or define the NSEC regions in the DT. Probably fine as is.

core/arch/arm/plat-stm32mp1/plat_tzc400.c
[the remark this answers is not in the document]
GseoC: Probably sufficient as SLIST_FOREACH(var, head, field) seems to exits when !var:
```
#define	SLIST_FOREACH(var, head, field)					\
	for((var) = (head)->slh_first; (var); (var) = (var)->field.sle_next)
```

GseoC:
Comments addressed.

GseoC:
Comments addressed

GseoC:
@etienne-lms, any more comments?

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +114,440 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+			break;
+	}
+
+	if (!reg)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	if (reg_exclude->base == reg->region.base &&
+	    reg_exclude->top == reg->region.top) {
+		/* Remove this entry */
+		SLIST_REMOVE(&nsec_region_list, reg, tzc_region_non_sec, link);
```
etienne-lms, line 221: `free(reg)`?

etienne-lms (commented):
2 last comments. LGTM otherwise.
Can you squash the fixup commits?

core/arch/arm/plat-stm32mp1/plat_tzc400.c
GseoC: I'm changing to SLIST_FOREACH_SAFE above then.

core/arch/arm/plat-stm32mp1/plat_tzc400.c
```diff
@@ -33,91 +114,440 @@ static enum itr_return tzc_it_handler(struct itr_handler *handler __unused)
...
+	for (i = 0; i < tzc_dev->ddata.nb_regions; i++)
+		tzc_configure_region(i + 1, &cfg);
+}
+
+static TEE_Result append_region(struct tzc_device *tzc_dev,
+				const struct tzc_region_config *region_cfg)
+{
+	TEE_Result res = TEE_SUCCESS;
+	unsigned int index = 0;
+	index = tzc_dev->nb_reg_used;
```
etienne-lms, line 178: Reported by checkpatch:
```patch
	unsigned int index = 0;
+
	index = tzc_dev->nb_reg_used;
```
or
```c
	unsigned int index = tzc_dev->nb_reg_used;
```

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with below comment addressed. Other checkpatch reports are false positive related to trace message instruction exceeding 80char/line.

The other CI test errors are due to https://xenbits.xen.org/git-http/xen.git not responding. Likely a temporary issue in Xen repositories server.

PR #7067, opened 2024-10-02
core: cache some embedded DT info + generic bisect helper function

etienne-lms:
Parse the embedded DTB prior using it and cache information that can take time for libdft functions to find. Cached info is stored in sorted arrays so that we can bisect to find the target info. Cached information are:
- parent node of a node (see [`fdt_parent_offset()`](https://github.com/OP-TEE/optee_os/blob/master/core/lib/libfdt/include/libfdt.h#L924-L925)).
- possibly heavily used `#address-cells` and `#size-cells` properties of a parent node.
- phandle of a node (see [`fdt_node_offset_by_phandle()`](https://github.com/OP-TEE/optee_os/blob/master/core/lib/libfdt/fdt_ro.c#L666-L671)).

The feature saves from few dozen to few hundreds of milliseconds of boot time depending on the DTB size (more specifically depending on the number of nodes in the DTB), for example, about 400ms boot time saved on STM32MP15/pager, and about 600ms saved on our STM32MP25 downstream config.

The patch series adds a generic bisect helper function and a test case in PTA test selftests. The test is proceed by CI on qemuv7 and qemuv8 paltform. These both platforms also embed a very small DTB in OP-TEE core so the DT cached info are also slightly tested (their embedded DTB only contains 7 nodes and 1 phandle).

### DIFF
--- a/core/include/kernel/dt.h
+++ b/core/include/kernel/dt.h
@@ -511,6 +511,15 @@ int fdt_find_cached_parent_node(const void *fdt, int node_offset,
  */
 int fdt_find_cached_parent_reg_cells(const void *fdt, int node_offset,
 				     int *address_cells, int *size_cells);
+/*
+ * Find the node offset from its phandle in the phandle cache
+ * @fdt: FDT to work on
+ * @phandle: Node phandle
+ * @node_offset: Pointer to output node offset upon success
+ * @return 0 on success and -FDT_ERR_NOTFOUND on failure
+ */
+int fdt_find_cached_node_phandle(const void *fdt, uint32_t phandle,
+				 int *node_offset);
 #else
 static inline int fdt_find_cached_parent_node(const void *fdt __unused,
 					      int node_offset __unused,
@@ -523,6 +532,13 @@ static inline int fdt_find_cached_parent_reg_cells(const void *fdt __unused,
 						   int node_offset __unused,
 						   int *address_cells __unused,
 						   int *size_cells __unused)
+{
+	return -1;
+}
+
+static inline int fdt_find_cached_node_phandle(const void *fdt __unused,
+					       uint32_t phandle __unused,
+					       int *node_offset __unused)
 {
 	return -1;
 }
--- a/core/include/kernel/dt.h
+++ b/core/include/kernel/dt.h
@@ -489,4 +489,42 @@ static inline void *get_manifest_dt(void)
 }
 
 #endif /* !CFG_DT */
+
+#ifdef CFG_DT_CACHED_NODE_INFO
+/*
+ * Find the offset of a parent node in the parent node cache
+ * @fdt: FDT to work on
+ * @node_offset: Offset of the node we look for its parent
+ * @parent_offset: Output parent node offset upon success
+ * @return 0 on success and -1 on failure
+ */
+int fdt_find_cached_parent_node(const void *fdt, int node_offset,
+				int *parent_offset);
+
+/*
+ * Find the address/size cells value of a parent node in the parent node cache
+ * @fdt: FDT to work on
+ * @node_offset: Offset of the node we look for its parent
+ * @address_cells: Pointer to output #address-cells value upon success or NULL
+ * @size_cells: Pointer to output #size-cells value upon success or NULL
+ * @return 0 on success and -FDT_ERR_NOTFOUND on failure
+ */
+int fdt_find_cached_parent_reg_cells(const void *fdt, int node_offset,
+				     int *address_cells, int *size_cells);
+#else
+static inline int fdt_find_cached_parent_node(const void *fdt __unused,
+					      int node_offset __unused,
+					      int *parent_offset __unused)
+{
+	return -1;
+}
+
+static inline int fdt_find_cached_parent_reg_cells(const void *fdt __unused,
+						   int node_offset __unused,
+						   int *address_cells __unused,
+						   int *size_cells __unused)
+{
+	return -1;
+}
+#endif /* CFG_DT_CACHED_NODE_INFO */
 #endif /* __KERNEL_DT_H */
--- a/core/kernel/dt.c
+++ b/core/kernel/dt.c
@@ -4,6 +4,7 @@
  */
 
 #include <assert.h>
+#include <bisect.h>
 #include <config.h>
 #include <initcall.h>
 #include <kernel/dt.h>
@@ -178,22 +179,24 @@ bad:
 
 paddr_t fdt_reg_base_address(const void *fdt, int offs)
 {
-	const void *reg;
-	int ncells;
-	int len;
-	int parent;
-
-	parent = fdt_parent_offset(fdt, offs);
-	if (parent < 0)
-		return DT_INFO_INVALID_REG;
+	const void *reg = NULL;
+	int ncells = 0;
+	int len = 0;
+	int parent = 0;
 
 	reg = fdt_getprop(fdt, offs, "reg", &len);
 	if (!reg)
 		return DT_INFO_INVALID_REG;
 
-	ncells = fdt_address_cells(fdt, parent);
-	if (ncells < 0)
-		return DT_INFO_INVALID_REG;
+	if (fdt_find_cached_parent_reg_cells(fdt, offs, &ncells, NULL)) {
+		parent = fdt_parent_offset(fdt, offs);
+		if (parent < 0)
+			return DT_INFO_INVALID_REG;
+
+		ncells = fdt_address_cells(fdt, parent);
+		if (ncells < 0)
+			return DT_INFO_INVALID_REG;
+	}
 
 	return fdt_read_paddr(reg, ncells);
 }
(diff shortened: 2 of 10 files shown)


etienne-lms:
CI / Code style tet failure is due to false positive warnings on trace messages exceeding 80 char/line:
```
8deb8ab9d core: pta: test bisect implementation
WARNING: line length of 105 exceeds 80 columns
#98: FILE: core/pta/tests/misc.c:620:
+						LOG("-> Found unexpected target %d (size %zu, hide %zu)",

WARNING: line length of 102 exceeds 80 columns
#105: FILE: core/pta/tests/misc.c:627:
+						LOG("- Failed to find target %d (size %zu, hide %zu)",

WARNING: line length of 108 exceeds 80 columns
#110: FILE: core/pta/tests/misc.c:632:
+						LOG("- Wrongly found %d for target %d (size %zu, hide %zu)",

total: 0 errors, 3 warnings, 0 checks, 111 lines checked

File: core/include/kernel/dt.h [hunk 1]
```diff
@@ -324,6 +324,41 @@ void reinit_manifest_dt(void);
...
 void *get_manifest_dt(void);
 
+/*
+ * Find the offset of a parent node in the parent node cache
+ * @fdt: FDT to work on
+ * @node_offset: Offset of the node we look for its parent
+ * @parent_offset: Output parent node offset upon success
+ * @return 0 on success and -1 on failure
+ *
+ * This function is supported when CFG_DT_CACHED_NODE_INFO is enabled.
```
jforissier, line 334: I believe the cache feature should be more transparent. Ideally hook the cache into the FDT lib code. Or if we don't want to touch libfdt, make a wrapper. But the caller should not need to know if the cache is enabled or not.

lib/libutils/isoc/include/bisect.h
```diff
@@ -0,0 +1,36 @@
...
+/*
+ * Bisect into array to find the cell that matches an entry
+ * @array: Sorted array (according the @cmp) to bisect in
+ * @n: Number of cells in @array
+ * @cell_size: Byte size of a single cell of @array
+ * @target: Entry for which we find an equal entry (with @cmp) in sorted @array
+ * @cmp: Trilean comparision helper function applicable to @array
+ * @return: Pointer to the cell in @array matching @target, NULL if none found
+ */
+void *bisect_equal(const void *array, size_t n, size_t cell_size,
```
jforissier, line 33: Strange name, how about `bisect_find()`?

jforissier (commented):
> Parse the embedded DTB prior using it and cache information that can take time for libdft functions to find. Cached info is stored in sorted arrays so that we can bisect to find the target info. 

Have you considered using a hash table? I have a feeling it would be even faster.

More comments below.

core/include/kernel/dt.h
etienne-lms: I'm fine patch _libfdt_ but I'm pretty confident such change would never hit libfdt mainstream.
The way `fdt_reg_base_address()` is implemented in our _kernel/dt.h_ make that fully leverage adresse cell (to prevent a node property parsing) makes that this very function needs to be patched. 
Same for `fdt_reg_size()` (to prevent 2 parsing of node properties) and `fdt_fill_device_info()` (2 parent node lookup + 2 node properties parsing).

That said, I can change this inline description to:
```patch
- * Find the offset of a parent node in the parent node cache
+ * Find the offset of a parent node in the parent node cache if any
 * (...)
- *
- * This function is supported when CFG_DT_CACHED_NODE_INFO is enabled.

File: lib/libutils/isoc/include/bisect.h (hunk 2 above)
Comment (etienne-lms) [reply lib/libutils/isoc/include/bisect.h:33]: I agree the name may sound strange.
I choose `bisect_equal()` since the function find a cell that matches the target in an equal case regarding `cmp()` callback.
I think there could be `bisect_roundup()`, `bisect_rounddow()` and `bisect_nearest()` for returning the closest matching cell.
That said, maybe `bisect()` straight could be applicable for the exact matching case.

**Review by etienne-lms — COMMENTED** [review]
> > Parse the embedded DTB prior using it and cache information that can take time for libdft functions to find. Cached info is stored in sorted arrays so that we can bisect to find the target info.
> 
> Have you considered using a hash table? I have a feeling it would be even faster.

Why do you think a hash table would speed up the processed?  We would still need to parse the hash table to find our target.

File: core/include/kernel/dt.h (hunk 1 above)
Comment (etienne-lms) [reply core/include/kernel/dt.h]: I've removed the fixup commit that held your comment. The commit was noisy and buggy so I it replaced with another fixup commit that should work better.
Please feel free to use this comment thread if needed. I won't tag it 'Resolved' :)

**Comment by etienne-lms** [comment]
> [Hashing] Parse the DT, for each value, apply a hash function, copy the hash and the value into an array of linked lists (buckets), then to find a value apply the hash function and read from the appropriate linked list.

Parsing a list to find a hash value seems less efficient (to me) than an array bisect operation. I must have missed something.

**Comment by jforissier** [comment]
> > [Hashing] Parse the DT, for each value, apply a hash function, copy the hash and the value into an array of linked lists (buckets), then to find a value apply the hash function and read from the appropriate linked list.
> 
> Parsing a list to find a hash value seems less efficient (to me) than an array bisect operation. I must have missed something.

Hopefully there is no collision and in this case you reach the value you are looking for immediately (that is, the linked lists have a single element). A linked list is one possible implementation to address collisions: https://en.wikipedia.org/wiki/Hash_table#Separate_chaining

**Review by jenswi-linaro — COMMENTED** [review]
Comment for "libutils: add bisect helper functions", this looks very much like the standard function bsearch(). Can we import `bsearch()` from newlib instead?

**Comment by etienne-lms** [comment]
Thanks for the pointer. I was looking for a bisect open source implementation but failed to find a compliant one. I missed to look into newlib source tree. I'll remove my implementation in favor to this far more mature one.

File: core/kernel/dt.c [hunk 3]
```diff
@@ -472,6 +496,203 @@ void *get_secure_dt(void)
...
+
+/*
+ * struct parent_node_cache - Cache of FDT parent node references
+ *
+ * @node_offset: Offset of a node in @cached_node_info_fdt
+ * @parent_offset: Offset of @node_offset parent node or -1 if root node
+ * @address_cells: Parent node #address-cells property value or 0
+ * @size_cells: Parent node #size-cells property value or 0
+ */
+struct parent_node_cache {
```
jenswi-linaro, line 515: I find the `parent` part in the name confusing since the struct describes a node. How about `struct cached_node`?

jenswi-linaro (commented):
Comments for "core: dt: cache embedded DTB node parent information"

jenswi-linaro:
Please drop the `Change-Id:` tags.

etienne-lms:
I've tested use of a hash table, using a simplistic hash algo (some XOR on offset/phandle value bytes). It gives the same boot perfs on the platforms i've tested. Only adds some extra heap consumption.

By the way I've also tested few other schemes and I found that on my platforms with at most several hundreds of nodes in the DT (tested worse case with 1 thousand), scanning a list or an array of the preloaded structures, without any bisect or hash indices optimization, give the overall same results. All in one, I wonder if bisect or hash indices are really useful here, seen the relatively small size of the database (hundreds of nodes in the DTB).

I think I should simplify this to the minimum even is less elegant than using hashed or sorted arrays.

jforissier:
@etienne-lms that's interesting. If there is no clear benefit in performance I would also think the simplest algorithm is the best.

core/kernel/dt.c
etienne-lms: The struct intended to relate to parent node information of a node, hence this name.
I'll fix that to use a common struct for both parent node info and node phandle info. I'll use `cached_node`.

core/kernel/dt.c
```diff
@@ -472,6 +496,203 @@ void *get_secure_dt(void)
...
+	int parent_offset;
+	size_t address_cells;
+	size_t size_cells;
+};
+
+/* Reference to node's parent information */
+static struct {
+	struct parent_node_cache *array;
+	size_t count;
+	size_t alloced_count;
```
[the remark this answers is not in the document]
etienne-lms: I allocated struct by chunk of 64 instances, for some efficiency. Therefore I count the allocated and filled cells separately.

core/kernel/dt.c
```diff
@@ -472,6 +496,203 @@ void *get_secure_dt(void)
...
+
+static void release_parent_node_cache(void)
+{
+	free(parent_node_cache.array);
+	parent_node_cache.array = NULL;
+	parent_node_cache.count = 0;
+	parent_node_cache.alloced_count = 0;
+}
+
+static TEE_Result enlarge_parent_node_cache(void)
```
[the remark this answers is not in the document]
etienne-lms: ok

core/kernel/dt.c
```diff
@@ -472,6 +496,203 @@ void *get_secure_dt(void)
...
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	const fdt32_t *cuint = NULL;
+	int subnode_offset = 0;
+	size_t addr_cells = 0;
+	size_t size_cells = 0;
+
+	cuint = fdt_getprop(fdt, node_offset, "#address-cells", NULL);
+	if (cuint)
+		addr_cells = (int)fdt32_to_cpu(*cuint);
```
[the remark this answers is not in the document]
etienne-lms: I see that libfdt does not go 1 parent-level up to get these information. It would make sense but I don't think this is how DT support is implemented.

jenswi-linaro:
So you mean that not caching is as fast as caching?

etienne-lms:
No, I meant parsing an array or list of cached info is as fast as finding the cached info using bisect or or hash table.

jenswi-linaro:
OK, because I suspect there's some overhead in collecting the list.

etienne-lms:
Actually the overhead for bisect or hashing is very small. I'v removed these optims because they don't add much value (until a platform comes with a DTB with several thousands of node) and cost some more memory. We'll be able to integrate such optims if needed in the futur.

Since the changes needed in updating this series are quite big and revert commits, I preferred to create a new P-R instead of appending fixup commits. Please have  a look at https://github.com/OP-TEE/optee_os/pull/7073. Consequently, I close this P-R.

PR #7069, opened 2024-10-03
Default enable CFG_NOTIF_TEST_WD only when dependencies are enabled

jforissier:
Set CFG_NOTIF_TEST_WD ?= y only when the features it needs are enabled. Fixes the following warning on platforms that enable CFG_ENABLE_EMBEDDED_TESTS but not CFG_CORE_ASYNC_NOTIF or CFG_CALLOUT:

 mk/config.mk:1039: Warning: Disabling CFG_NOTIF_TEST_WD [requires CFG_CALLOUT CFG_CORE_ASYNC_NOTIF]

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -1035,7 +1035,8 @@ endif
 CFG_CALLOUT ?= $(CFG_CORE_ASYNC_NOTIF)
 
 # Enable notification based test watchdog
-CFG_NOTIF_TEST_WD ?= $(CFG_ENABLE_EMBEDDED_TESTS)
+CFG_NOTIF_TEST_WD ?= $(call cfg-all-enabled,CFG_ENABLE_EMBEDDED_TESTS \
+		       CFG_CALLOUT CFG_CORE_ASYNC_NOTIF)
 $(eval $(call cfg-depends-all,CFG_NOTIF_TEST_WD,CFG_CALLOUT \
 	 CFG_CORE_ASYNC_NOTIF))
 


mk/config.mk
```diff
@@ -1035,7 +1035,9 @@ endif
 CFG_CALLOUT ?= $(CFG_CORE_ASYNC_NOTIF)
 
 # Enable notification based test watchdog
-CFG_NOTIF_TEST_WD ?= $(CFG_ENABLE_EMBEDDED_TESTS)
+ifeq (yyy,$(CFG_ENABLE_EMBEDDED_TESTS)$(CFG_CALLOUT)$(CFG_CORE_ASYNC_NOTIF))
```
etienne-lms, line 1038: as a sanity test on the config values, prefer
`ifeq (y-y-y,$(CFG_ENABLE_EMBEDDED_TESTS)-$(CFG_CALLOUT)-$(CFG_CORE_ASYNC_NOTIF))`

That said, I'd rather suggest:
```makefile
CFG_NOTIF_TEST_WD ?= $(call cfg-all-enabled,CFG_ENABLE_EMBEDDED_TESTS CFG_CALLOUT CFG_CORE_ASYNC_NOTIF)

etienne-lms (commented):
Nice change indeed.
Could you prefix the commit header file with `mk: config:`?
E.g. `mk: config: default enable CFG_NOTIF_TEST_WD upon dependencies`

jforissier:
Update

etienne-lms (commented):
The commit message header line is a bit too large.
Aside that, `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

jforissier:
Subject shortened, tag applied. Thanks!

PR #7073, opened 2024-10-09
core: cache some embedded DT info (take 2)

etienne-lms:
This change supersedes https://github.com/OP-TEE/optee_os/pull/7067. Main difference is that we no more bisect into the cached node structure (since it does not speeds up boot time) and this change proposes a new helper function: `fdt_reg_info()` to look-up parent node once when retrieving the `reg` property base address and size.

Parse the embedded DTB prior using it and cache information that can take time for libdft functions to find. Cached info is stored in an array so that we can find the target info without parsing through the DTB. Cached information are:
- parent node of a node (see [`fdt_parent_offset()`](https://github.com/OP-TEE/optee_os/blob/master/core/lib/libfdt/include/libfdt.h#L924-L925)).
- possibly heavily used `#address-cells` and `#size-cells` properties of a parent node.
- phandle of a node (see [`fdt_node_offset_by_phandle()`](https://github.com/OP-TEE/optee_os/blob/master/core/lib/libfdt/fdt_ro.c#L666-L671)).

The feature saves from a few dozen to several hundreds of milliseconds of boot time depending on the DTB size (more specifically depending on the number of nodes in the DTB), for example, about 400ms boot time saved on STM32MP15/pager, and more than 1sec saved on our STM32MP25 downstream config.

These changes are processed by CI on qemuv7 and qemuv8 platforms since both embed a very small DTB in OP-TEE core (their embedded DTBs only contains 7 nodes and 1 phandle).

### DIFF
--- a/core/kernel/dt.c
+++ b/core/kernel/dt.c
@@ -110,11 +110,7 @@ int dt_map_dev(const void *fdt, int offs, vaddr_t *base, size_t *size,
 	if (st == DT_STATUS_DISABLED)
 		return -1;
 
-	pbase = fdt_reg_base_address(fdt, offs);
-	if (pbase == DT_INFO_INVALID_REG)
-		return -1;
-	sz = fdt_reg_size(fdt, offs);
-	if (sz == DT_INFO_INVALID_REG_SIZE)
+	if (fdt_reg_info(fdt, offs, &pbase, &sz))
 		return -1;
 
 	switch (mapping) {
@@ -176,30 +172,6 @@ bad:
 
 }
 
-paddr_t fdt_reg_base_address(const void *fdt, int offs)
-{
-	const void *reg = NULL;
-	int ncells = 0;
-	int len = 0;
-	int parent = 0;
-
-	reg = fdt_getprop(fdt, offs, "reg", &len);
-	if (!reg)
-		return DT_INFO_INVALID_REG;
-
-	if (fdt_find_cached_parent_reg_cells(fdt, offs, &ncells, NULL)) {
-		parent = fdt_parent_offset(fdt, offs);
-		if (parent < 0)
-			return DT_INFO_INVALID_REG;
-
-		ncells = fdt_address_cells(fdt, parent);
-		if (ncells < 0)
-			return DT_INFO_INVALID_REG;
-	}
-
-	return fdt_read_paddr(reg, ncells);
-}
-
 static size_t fdt_read_size(const uint32_t *cell, int n)
 {
 	uint32_t sz = 0;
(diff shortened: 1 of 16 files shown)


core/kernel/dt.c
```diff
@@ -473,6 +496,223 @@ void *get_secure_dt(void)
...
+
+	*node_offset = cell->node_offset;
+
+	return 0;
+}
+
+static TEE_Result grow_cached_node_array(void)
+{
+	if (dt_node_cache.count + 1 > dt_node_cache.alloced_count) {
+		/* Allocate by chunk of 64 cells for efficiency */
```
jenswi-linaro, line 606: The normal approach for scaling up is to double the allocations each time.

core/kernel/dt.c
```diff
@@ -473,6 +496,223 @@ void *get_secure_dt(void)
...
+
+	if (!cell)
+		return -FDT_ERR_NOTFOUND;
+
+	*node_offset = cell->node_offset;
+
+	return 0;
+}
+
+static TEE_Result grow_cached_node_array(void)
```
jenswi-linaro, line 603: The name of this function would be accurate without the `if` check below. Now the function rather guarantees room for at least one more element.

core/kernel/dt.c
```diff
@@ -473,6 +496,223 @@ void *get_secure_dt(void)
...
+};
+
+/*
+ * struct dt_node_cache - Reference to cached information of DT nodes
+ * @array: Array of the cached node
+ * @count: Number of initialized cells in @array
+ * @alloced_array
+ */
+struct dt_node_cache {
+	struct cached_node *array;
```
jenswi-linaro, line 518: Wouldn't it make more sense to keep `cached_node_info_fdt` in this struct?

core/kernel/dt.c
```diff
@@ -473,6 +496,223 @@ void *get_secure_dt(void)
...
+	size_t count;
+	size_t alloced_count;
+};
+
+static struct dt_node_cache dt_node_cache;
```
jenswi-linaro, line 523: How about allocating this and only keep a pointer to it so save memory once we're done with dt?

jenswi-linaro (commented):
Comment for "core: dt: cache embedded DTB node information"

core/kernel/dt.c
```diff
@@ -284,10 +292,25 @@ void fdt_fill_device_info(const void *fdt, struct dt_node_info *info, int offs)
...
+	int size_cells = 0;
+
+	if (fdt_find_cached_parent_reg_cells(fdt, offs, &addr_cells,
+					     &size_cells) == 0) {
+		int len = 0;
 
-	dinfo.reg = fdt_reg_base_address(fdt, offs);
-	dinfo.reg_size = fdt_reg_size(fdt, offs);
+		cuint = fdt_getprop(fdt, offs, "reg", &len);
+		if (cuint &&
```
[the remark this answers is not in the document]
etienne-lms: If the node does not have a valid `reg` property, `dinfo` contains the invalid IDs and caller will handle that, as prior this change.

core/kernel/dt.c
etienne-lms: Indeed, i'll add it.

core/kernel/dt.c
etienne-lms: Fair

core/kernel/dt.c
etienne-lms: We talking about the DT, so I think there should be a limit where doubling the allocated size would be not optimal. 64 cells consume 64x16=1kByte. I though it would be a fair compromise.

core/kernel/dt.c
etienne-lms: `realloc_cached_node_array()` would maybe be better.

core/kernel/dt.c
jenswi-linaro: What is the problem with doubling? In the worst case, you have half the allocation as overhead.
The problem with your approach is that it doesn't scale, even if it might be sufficient for now.

etienne-lms: Fair, i'll update. Indeed, it would better scale to generic cases.

core/lib/libfdt/fdt_ro.c
```diff
@@ -661,6 +674,10 @@ int fdt_node_offset_by_phandle(const void *fdt, uint32_t phandle)
 	if ((phandle == 0) || (phandle == -1))
 		return -FDT_ERR_BADPHANDLE;
 
+	/* OP-TEE helper function to leveral cached information from the DT */
```
jenswi-linaro, line 677: s/leveral/leverage/, but perhaps better to only say "use".

core/lib/libfdt/fdt_ro.c
```diff
@@ -8,6 +8,9 @@
 #include <fdt.h>
 #include <libfdt.h>
 
+/* OP-TEE helper function to leveral cached information from the DT */
```
jenswi-linaro, line 11: I think we have quite a few copies of this comment already. Can we remove some of them?

core/lib/libfdt/fdt_ro.c
```diff
@@ -489,6 +492,7 @@ uint32_t fdt_get_phandle(const void *fdt, int nodeoffset)
 	const fdt32_t *php;
 	int len;
 
+
```
jenswi-linaro, line 495: Please remove this line.

core/lib/libfdt/fdt_ro.c
etienne-lms: Another way to less hack into **libfdt** would be to rename the modified libfdt functions and implement them in `core/kernel/dt.c`. For example, regarding `fdt_node_offset_by_phandle()`:
In `core/lib/libfdt/fdt_ro.c` and `core/lib/libfdt/include/libfdt.h`:
```patch
-int fdt_node_offset_by_phandle(const void *fdt, uint32_t phandle)
+int _fdt_node_offset_by_phandle(const void *fdt, uint32_t phandle)
```
and in `core/kernel/dt.c`:
```c
int fdt_node_offset_by_phandle(const void *fdt, uint32_t phandle)
{
	int offset = 0;

	if (fdt_find_cached_node_phandle(fdt, phandle, &offset) == 0)
		return offset;

	return _fdt_node_offset_by_phandle(fdt, phandle)
}
```

Would it be preferable?

etienne-lms:
Comments addressed.

jenswi-linaro:

etienne-lms:
Review tag applied.

PR #7075, opened 2024-10-10
core: imx: disable ELE support on i.MX91 by default

sahilnxp:
On i.MX91, there is only one MU to communicate with ELE, which cannot be dedicated on OP-TEE side all the time. There may be ELE services running on Linux side, which can cause conflict with OP-TEE.
So disabling ELE by default for now.

### DIFF
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -262,7 +262,7 @@ $(call force,CFG_IMX_OCOTP,n)
 $(call force,CFG_TZC380,n)
 $(call force,CFG_NXP_CAAM,n)
 CFG_IMX_MU ?= y
-CFG_IMX_ELE ?= y
+CFG_IMX_ELE ?= n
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Is linux upstream using any of the ELE services? Because if not then I don't see why changing the default behavior here as you can already configure that downstream.

sahilnxp:
@ricardosalveti Yes, it is in process of getting upstreamed.

Also, we have kept the ELE driver by default disabled for i.M8ULP and i.MX93, so on same lines it needs to be disabled too.

FYI, on i.MX95 we have dedicated MU for Secure World, so we will be keeping ELE support enabled on that one by default.

sahilnxp:
@jforissier @jenswi-linaro can you please have a look this  ?

jforissier:
No objection from me... FWIW:

sahilnxp:
Thanks @jforissier
Tags updated.

PR #7078, opened 2024-10-14
RISC-V small fix: Disable traps by clearing XIE CSR

gagachang:
Ensure we disable traps by clearing XIE CSR instead of clearing XSTATUS.IE which is global interrupt enable bit.


maroueneboubakri:

gagachang:
@maroueneboubakri , It’s been so long! Thanks!
Tag applied.

PR #7081, opened 2024-10-15
drivers: stm32_rng: MP15 RNG is non-secure when PRNG is enable

etienne-lms:
Register stm32_rng device as non-secure when software PRNG is enabled instead of testing the firewall configuration that is applied from stm32mp1_init_final_shres() at driver_init_late initcall level, far after RNG initialization.

Fixes: d773ec0baf4c ("drivers: stm32_rng: update clock and power management")

### DIFF
--- a/core/drivers/stm32_rng.c
+++ b/core/drivers/stm32_rng.c
@@ -672,7 +672,7 @@ static TEE_Result stm32_rng_probe(const void *fdt, int offs,
 
 #if defined(CFG_STM32MP15)
 	/* Only STM32MP15 requires a software registering of RNG secure state */
-	if (etzpc_get_decprot(STM32MP1_ETZPC_RNG1_ID) == ETZPC_DECPROT_NS_RW)
+	if (IS_ENABLED(CFG_WITH_SOFTWARE_PRNG))
 		stm32mp_register_non_secure_periph_iomem(stm32_rng->base.pa);
 	else
 		stm32mp_register_secure_periph_iomem(stm32_rng->base.pa);


GseoC:
To be functional before your fix, we indeed need #7066. So:

I'll take care on rebasing my series afterwards as I suppose it'll be merged after the final tag.

etienne-lms:
I mean once #7066 is merged, this fix will no more be needed since the ETZPC config for STM32 RNG will then be loaded before stm32_rng driver init. Since #7066 is not yet merged, we need this fix to ensure STM32 RNG is assigned to non-secure world and U-Boot can safely use it (as required with U-Boot tag v2024.10).

etienne-lms:
Review tag applied.

@jforissier, would you be ok to consider this platform fix in the  4.4.0 release tag? 
It fixes an issue seen with latest U-Boot tag v2024.10 on stm32mp15 boards regarding STM32 RNG subsystem (accessed by U-Boot since commit [1] makes U-Boot to generate an ASLR seed for the dear kernel).
Stm32mp1/OP-TEE release build is currently based on U-Boot v2024.01-rc1 [2] and does not strictly needs this fix but I would rather have the coming 4.4.0 tag to comply with the latest released U-Boot tag for these platforms.

[1] https://source.denx.de/u-boot/u-boot/-/commit/ea955eea4f662b7e37d74228fed0c9147e6dba88
[2] https://github.com/OP-TEE/manifest/blob/refs/tags/4.3.0/stm32mp1.xml#L18

jforissier:
@etienne-lms thanks for the detailed explanation. Yes this fix makes perfect sense to get into 4.4.0.

etienne-lms:
Thanks.

GseoC:
Thank you :)

PR #7084, opened 2024-10-16
ci: qemuv8: add configurations with CFG_ULIBS_SHARED=y

jforissier:
Add configurations to the QEMUv8 job to test shared library support with GCC as well as Clang. Shared libraries are somewhat already tested by xtest 1022 which performs dlopen()/dlsym() on a custom library, but CFG_ULIBS_SHARED=y will thoroughly test the loading and symbol resolution at TA load time.


jenswi-linaro:

PR #7085, opened 2024-10-18
mk: add global-cppflags-y

jenswi-linaro:
Add global-cppflags-y as a new sub.mk variable. The content of the variable is added to the C preprocessor flags passed to all files compiled in the current submodule, $(sm), similarly to how the content of global-incdirs-y is passed on.

This flag is intended for libraries that in addition to exporting an include path also need certain defines set.

### DIFF
--- a/mk/subdir.mk
+++ b/mk/subdir.mk
@@ -140,6 +140,7 @@ endif
 
 include $1/sub.mk
 sub-subdirs := $$(addprefix $1/,$$(subdirs-y)) $$(subdirs_ext-y)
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

jenswi-linaro:
Tag applied.

jforissier (commented):

jenswi-linaro:
Tag applied.

jenswi-linaro:
Ready do merge?

PR #7089, opened 2024-10-22
plat-k3: Add locks for TI-SCI protocol

manorit2001:
Currently TI-SCI layer doesn't have any locks that can lead to race conditions, add locks in appropriate places to avoid race conditions.

### DIFF
--- a/core/arch/arm/plat-k3/drivers/ti_sci.c
+++ b/core/arch/arm/plat-k3/drivers/ti_sci.c
@@ -8,6 +8,7 @@
  */
 
 #include <assert.h>
+#include <kernel/mutex.h>
 #include <malloc.h>
 #include <platform_config.h>
 #include <string.h>
@@ -20,6 +21,7 @@
 #include "ti_sci_protocol.h"
 
 static uint8_t message_sequence;
+static struct mutex ti_sci_mutex_lock = MUTEX_INITIALIZER;
 
 /**
  * struct ti_sci_xfer - Structure representing a message flow
@@ -136,11 +138,13 @@ static inline int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 	struct k3_sec_proxy_msg *msg = &xfer->tx_message;
 	int ret = 0;
 
+	mutex_lock(&ti_sci_mutex_lock);
+
 	/* Send the message */
 	ret = k3_sec_proxy_send(msg);
 	if (ret) {
 		EMSG("Message sending failed (%d)", ret);
-		return ret;
+		goto unlock;
 	}
 
 	/* Get the response */
@@ -148,10 +152,12 @@ static inline int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 	if (ret) {
 		if ((TEE_Result)ret != TEE_ERROR_ACCESS_DENIED)
 			EMSG("Failed to get response (%d)", ret);
-		return ret;
+		goto unlock;
 	}
 
-	return 0;
+unlock:
+	mutex_unlock(&ti_sci_mutex_lock);
+	return ret;
 }
 
 int ti_sci_get_revision(struct ti_sci_msg_resp_version *rev_info)
--- a/core/arch/arm/plat-k3/drivers/ti_sci.c
+++ b/core/arch/arm/plat-k3/drivers/ti_sci.c
@@ -109,6 +109,9 @@ static int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 		goto unlock;
 	}
 
+	FMSG("Sending %"PRIx16" with seq %"PRIu8" host %"PRIu8,
+	     txhdr->type, txhdr->seq, txhdr->host);
+
 	/* Get the response */
 	for (; retry > 0; retry--) {
 		/* Receive the response */
@@ -134,7 +137,11 @@ static int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 	if (!(rxhdr->flags & TI_SCI_FLAG_RESP_GENERIC_ACK)) {
 		DMSG("Message not acknowledged");
 		ret = TEE_ERROR_ACCESS_DENIED;
+		goto unlock;
 	}
+
+	FMSG("Receive %"PRIx16" with seq %"PRIu8" host %"PRIu8,
+	     rxhdr->type, rxhdr->seq, rxhdr->host);
 
 unlock:
 	mutex_unlock(&ti_sci_mutex_lock);
--- a/core/arch/arm/plat-k3/drivers/ti_sci.c
+++ b/core/arch/arm/plat-k3/drivers/ti_sci.c
@@ -20,8 +20,6 @@
 #include "ti_sci.h"
 #include "ti_sci_protocol.h"
 
-static uint8_t message_sequence;
-static struct mutex ti_sci_mutex_lock = MUTEX_INITIALIZER;
 
 /**
  * struct ti_sci_xfer - Structure representing a message flow
@@ -82,80 +80,60 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
 }
 
 /**
- * ti_sci_get_response() - Receive response from mailbox channel
- *
- * @xfer:	Transfer to initiate and wait for response
- *
- * Return: 0 if all goes well, else appropriate error message
- */
-static inline int ti_sci_get_response(struct ti_sci_xfer *xfer)
-{
-	struct k3_sec_proxy_msg *msg = &xfer->rx_message;
-	struct ti_sci_msg_hdr *hdr = NULL;
-	unsigned int retry = 5;
-	int ret = 0;
-
-	for (; retry > 0; retry--) {
-		/* Receive the response */
-		ret = k3_sec_proxy_recv(msg);
-		if (ret) {
-			EMSG("Message receive failed (%d)", ret);
-			return ret;
-		}
-
-		/* msg is updated by Secure Proxy driver */
-		hdr = (struct ti_sci_msg_hdr *)msg->buf;
-
-		/* Sanity check for message response */
-		if (hdr->seq == message_sequence)
-			break;
-
-		IMSG("Message with sequence ID %u is not expected", hdr->seq);
-	}
-	if (!retry) {
-		EMSG("Timed out waiting for message");
-		return TEE_ERROR_BUSY;
-	}
-
-	if (!(hdr->flags & TI_SCI_FLAG_RESP_GENERIC_ACK)) {
-		DMSG("Message not acknowledged");
-		return TEE_ERROR_ACCESS_DENIED;
-	}
-
-	return 0;
-}
-
-/**
  * ti_sci_do_xfer() - Do one transfer
  *
  * @xfer: Transfer to initiate and wait for response
  *
  * Return: 0 if all goes well, else appropriate error message
  */
-static inline int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
+static int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 {
-	struct k3_sec_proxy_msg *msg = &xfer->tx_message;
-	struct ti_sci_msg_hdr *hdr = NULL;
+	struct k3_sec_proxy_msg *txmsg = &xfer->tx_message;
+	struct k3_sec_proxy_msg *rxmsg = &xfer->rx_message;
+	struct ti_sci_msg_hdr *txhdr = (struct ti_sci_msg_hdr *)txmsg->buf;
+	struct ti_sci_msg_hdr *rxhdr = (struct ti_sci_msg_hdr *)rxmsg->buf;
+	static uint8_t message_sequence;
+	static struct mutex ti_sci_mutex_lock = MUTEX_INITIALIZER;
+	unsigned int retry = 5;
 	int ret = 0;
 
 	mutex_lock(&ti_sci_mutex_lock);
 
-	hdr = (struct ti_sci_msg_hdr *)msg->buf;
-	hdr->seq = ++message_sequence;
+	message_sequence++;
+	txhdr->seq = message_sequence;
 
 	/* Send the message */
-	ret = k3_sec_proxy_send(msg);
+	ret = k3_sec_proxy_send(txmsg);
 	if (ret) {
 		EMSG("Message sending failed (%d)", ret);
 		goto unlock;
 	}
 
 	/* Get the response */
-	ret = ti_sci_get_response(xfer);
-	if (ret) {
-		if ((TEE_Result)ret != TEE_ERROR_ACCESS_DENIED)
-			EMSG("Failed to get response (%d)", ret);
+	for (; retry > 0; retry--) {
+		/* Receive the response */
+		ret = k3_sec_proxy_recv(rxmsg);
+		if (ret) {
+			EMSG("Message receive failed (%d)", ret);
+			goto unlock;
+		}
+
+		/* Sanity check for message response */
+		if (rxhdr->seq == message_sequence)
+			break;
+
+		IMSG("Message with sequence ID %"PRIu8" is not expected",
+		     rxhdr->seq);
+	}
+	if (!retry) {
+		EMSG("Timed out waiting for message");
+		ret = TEE_ERROR_BUSY;
 		goto unlock;
+	}
+
+	if (!(rxhdr->flags & TI_SCI_FLAG_RESP_GENERIC_ACK)) {
+		DMSG("Message not acknowledged");
+		ret = TEE_ERROR_ACCESS_DENIED;
 	}
 
 unlock:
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -68,11 +71,16 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
...
 	hdr = (struct ti_sci_msg_hdr *)tx_buf;
+
+	exceptions = cpu_spin_lock_xsave(&ti_sci_spin_lock);
+
 	hdr->seq = ++message_sequence;
 	hdr->type = msg_type;
 	hdr->host = OPTEE_HOST_ID;
 	hdr->flags = msg_flags | TI_SCI_FLAG_REQ_ACK_ON_PROCESSED;
 
+	cpu_spin_unlock_xrestore(&ti_sci_spin_lock, exceptions);
```
etienne-lms, line 83: I think only `message_sequence` incrementing needs to be protected.

manorit2001: Done.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -134,24 +144,32 @@ static inline int ti_sci_get_response(struct ti_sci_xfer *xfer)
 static inline int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 {
 	struct k3_sec_proxy_msg *msg = &xfer->tx_message;
+	struct ti_sci_msg_hdr *hdr = NULL;
 	int ret = 0;
 
+	mutex_lock(&ti_sci_mutex_lock);
+
+	hdr = (struct ti_sci_msg_hdr *)msg->buf;
+	current_sequence = hdr->seq;
```
glneo, line 153: We could set the `hdr->seq` here instead of up in `ti_sci_setup_xfer()`, that would put it inside this mutex section and remove the need for the extra spinlock. Then as both updating and checking against message_sequence happen in this same mutex section, you'll also not need the extra temp `current_sequence`, just use `message_sequence` in `ti_sci_get_response()` as before.

glneo (changes requested):

core/arch/arm/plat-k3/drivers/ti_sci.c
manorit2001: Hey @glneo , that seems like a good idea but in general wondering something.. I see that `message_sequence` is a uint8_t datatype which means that the maximum limit of messages should be around 255 before it rolls over and comes back to 0. 

Am guessing that we do more message communications than this and we must've been rolling over and coming to 0? If yes, do we even need a counter if we are keeping this code under mutex? ( Every call will be sequential so we shouldn't require it am guessing and we can keep the hdr->seq = 0 always? )

glneo: Right, `message_sequence` rolls over every now and then, it is just a check for us that the message we get back is probably in response to the message we sent out if the number matches. The actual value is unimportant, just that it is different from the messages around it. It could be a random number generated each time too.

In theory, each send and receive call pair should be inside a mutex together so this shouldn't matter. In practice I have been saved with this check in some odd cases where a message was stuck in the SProxy queue and even though I just sent a message, I was getting a response for a message sent before OP-TEE even started.

manorit2001: Oh okay, then am wondering that the check should be at host_id level rather than message_sequence right?

glneo: Are you asking if we only need to check `host_id`? That would prevent messages from other hosts in our queue like my above example, yes, but still not prevent any issues in OP-TEE from being detected. I'd like to point out the issue that lead to this PR was discovered because we had this check in place. Could you imaging how hard it would have been to debug if we didn't have the check and we simply processed the returned message as if it was the correct order..

If we are confident the mutex is enough to prevent any future bugs then I'd be okay with dropping the sequence check.

manorit2001: Yeah don't mind keeping the check for sanity purpose, just wondered if can add host_id check as well then as that might make future issues more clear incase they arise, would update the PR. Thanks!

manorit2001: @glneo  updated

manorit2001:
@glneo I haven't put the host id check that I had mentioned as started facing some issues while testing. Will debug that separately and open a new PR if required, have refactored it the way you wanted, please review.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -136,22 +138,26 @@ static inline int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
...
 		if ((TEE_Result)ret != TEE_ERROR_ACCESS_DENIED)
 			EMSG("Failed to get response (%d)", ret);
-		return ret;
+		goto unlock;
 	}
 
-	return 0;
+unlock:
+	mutex_unlock(&ti_sci_mutex_lock);
+	return ret;
```
etienne-lms, line 180: By the way and out of the scope of this change: I see `k3_sec_proxy_xxx()` return `TEE_Result` values while caller use `int`. Likely something to be fixed at some point of time.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -8,6 +8,7 @@
  */
 
 #include <assert.h>
+#include <kernel/thread.h>
```
etienne-lms, line 11: prefer `kernel/mutex.h` for mutex support.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -136,10 +135,14 @@ static inline int ti_sci_get_response(struct ti_sci_xfer *xfer)
 static inline int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 {
 	struct k3_sec_proxy_msg *msg = &xfer->tx_message;
+	struct ti_sci_msg_hdr *hdr = NULL;
 	int ret = 0;
 
 	mutex_lock(&ti_sci_mutex_lock);
 
+	hdr = (struct ti_sci_msg_hdr *)msg->buf;
+	hdr->seq = ++message_sequence;
```
etienne-lms: For consistency in OP-TEE OS implementation, prefer a 2 step implementation and let compiler optimize the thing:
```c
message_sequence++;
hdr->seq = message_sequence;

File: core/arch/arm/plat-k3/drivers/ti_sci.c [hunk 6]
```diff
@@ -81,6 +81,25 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
...
 }
 
+/**
+ * ti_sci_send_request() - Send request to mailbox channel
+ *
+ * @xfer:	Transfer to initiate and wait for response
+ *
+ * Return: 0 if all goes well, else appropriate error message
+ */
+static inline int ti_sci_send_request(struct ti_sci_xfer *xfer)
```
etienne-lms, line 91: Could drop the `inline` attribute. Not strictly needed. But since there are such `inline` directive for many of the functions in this source file, maybe my comment is out of the scope of this P-R.

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -96,6 +96,8 @@ static inline int ti_sci_send_request(struct ti_sci_xfer *xfer)
 	hdr = (struct ti_sci_msg_hdr *)msg->buf;
 	hdr->seq = ++message_sequence;
 
+	FMSG("Sending %x with seq %u host %u", msg->type, msg->seq, msg->host);
```
etienne-lms, line 99: `type` is `uint16_t`, `seq` and `host` are `uint8_t`, hence use `PRIu16` and `PRIu8`. Ditto at line 146.
```c
	FMSG("Sending %#"PRIx16" with seq %"PRIu8" host %"PRIu8,
	     msg->type, msg->seq, msg->host);

Comment (glneo) [reply core/arch/arm/plat-k3/drivers/ti_sci.c:99]: The original line doesn't even compile.. Building with `CFG_TEE_CORE_LOG_LEVEL=4` shows that. `type`, `seq`, and `host` are members of `hdr`, not `msg`.

Comment (manorit2001) [reply core/arch/arm/plat-k3/drivers/ti_sci.c]: Ah shit, had missed it. Thanks!

File: core/arch/arm/plat-k3/drivers/ti_sci.c (hunk 6 above)
Comment (manorit2001) [reply core/arch/arm/plat-k3/drivers/ti_sci.c:91]: Yeah, I think the code mostly is kept similar to how it is in tf-a and for driver fixes and implementations to come easily from optee -> tfa and vice versa. These things are kept simillar.

File: core/arch/arm/plat-k3/drivers/ti_sci.c [hunk 8]
```diff
@@ -80,6 +81,28 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
...
+ *
+ * Return: 0 if all goes well, else appropriate error message
+ */
+static inline int ti_sci_send_request(struct ti_sci_xfer *xfer)
+{
+	struct k3_sec_proxy_msg *msg = &xfer->tx_message;
+	struct ti_sci_msg_hdr *hdr = NULL;
+
+	hdr = (struct ti_sci_msg_hdr *)msg->buf;
+	hdr->seq = ++message_sequence;
```
manorit2001, line 97: I haven't changed the following @etienne-lms , if you could let me know the reasoning as to why we need to split it into two and let the compiler do the optimisations..

glneo: Easier to read. Prefix increment is a bit rare, so having it here makes one stop for just an instant to parse out "oh yeah, that means update *then* assign", vs just updating then assigning written out in the code. Code is for humans, the compiler will produce the same assembly in both cases, so why be fancy when you can be explicit :)

manorit2001:
> So I see you factored out `ti_sci_send_request()` from `ti_sci_do_xfer()`, and this makes sense to match the receive side function for the same `ti_sci_get_response()`. But why not go the other way and merge `ti_sci_get_response()` into `ti_sci_do_xfer()`? These send/get functions are only called once from `ti_sci_do_xfer()`, and that function isn't all that long anyway.

I prefer it this way tbh, don't see a reason to put it all in ti_sci_do_xfer. Currently both of the functions ti_sci_get_response and ti_sci_send_request might be small but if together combined in one function of ti_sci_do_xfer will make that function big so IMO it makes sense to keep it this way.

core/arch/arm/plat-k3/drivers/ti_sci.c
manorit2001: Ah okay, makes sense. Would do the need

glneo:
Combining removes a lot of repeated definitions, the combined function ends up rather small, in fact here is the whole thing, just 50 lines, half the lines of the 3 split functions:

```
static int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
{
	struct k3_sec_proxy_msg *txmsg = &xfer->tx_message;
	struct k3_sec_proxy_msg *rxmsg = &xfer->rx_message;
	struct ti_sci_msg_hdr *txhdr = (struct ti_sci_msg_hdr *)txmsg->buf;
	struct ti_sci_msg_hdr *rxhdr = (struct ti_sci_msg_hdr *)rxmsg->buf;

	unsigned int retry = 5;
	int ret = 0;

	mutex_lock(&ti_sci_mutex_lock);

	message_sequence++;
	txhdr->seq = message_sequence;

	/* Send the message */
	ret = k3_sec_proxy_send(txmsg);
	if (ret) {
		EMSG("Message sending failed (%d)", ret);
		goto unlock;
	}

	/* Get the response */
	for (; retry > 0; retry--) {
		/* Receive the response */
		ret = k3_sec_proxy_recv(rxmsg);
		if (ret) {
			EMSG("Message receive failed (%d)", ret);
			goto unlock;
		}

		/* Sanity check for message response */
		if (rxhdr->seq == message_sequence)
			break;

		IMSG("Message with sequence ID %u is not expected", rxhdr->seq);
	}
	if (!retry) {
		EMSG("Timed out waiting for message");
		ret = TEE_ERROR_BUSY;
		goto unlock;
	}

	if (!(rxhdr->flags & TI_SCI_FLAG_RESP_GENERIC_ACK)) {
		DMSG("Message not acknowledged");
		ret = TEE_ERROR_ACCESS_DENIED;
	}

unlock:
	mutex_unlock(&ti_sci_mutex_lock);
	return ret;
}
```

manorit2001:
Hey @glneo , I still feel that it's better if we keep it split only. I don't think we should be merging ti_sci_get_response and even though small but I feel that in a separate function it's more contained and modular. I know the new function that I made might not be necessary but I just created it to keep consistency with ti_sci_get_response and I don't feel that we should be going the other way round for it IMO. Will split the prefix increment though like you and @etienne-lms mentioned.

glneo (approved):
Now that `message_sequence` and `ti_sci_mutex_lock` are only used in the one function, we could probably make them static local vars. Maybe a cleanup for another time. Everything else LGTM

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -81,77 +82,70 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
...
-static inline int ti_sci_get_response(struct ti_sci_xfer *xfer)
+static int ti_sci_do_xfer(struct ti_sci_xfer *xfer)
 {
-	struct k3_sec_proxy_msg *msg = &xfer->rx_message;
-	struct ti_sci_msg_hdr *hdr = NULL;
+	struct k3_sec_proxy_msg *txmsg = &xfer->tx_message;
+	struct k3_sec_proxy_msg *rxmsg = &xfer->rx_message;
+	struct ti_sci_msg_hdr *txhdr = (struct ti_sci_msg_hdr *)txmsg->buf;
+	struct ti_sci_msg_hdr *rxhdr = (struct ti_sci_msg_hdr *)rxmsg->buf;
+
```
etienne-lms, line 97: prefer without this extra empty line (not a strong opinion)

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -81,77 +82,70 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
...
-		/* msg is updated by Secure Proxy driver */
-		hdr = (struct ti_sci_msg_hdr *)msg->buf;
-
 		/* Sanity check for message response */
-		if (hdr->seq == message_sequence)
+		if (rxhdr->seq == message_sequence)
 			break;
 
-		IMSG("Message with sequence ID %u is not expected", hdr->seq);
+		IMSG("Message with sequence ID %u is not expected", rxhdr->seq);
```
etienne-lms, line 129: s/`%u`/`%"PRIu8"`/

etienne-lms (commented):
some minor comments

core/arch/arm/plat-k3/drivers/ti_sci.c
manorit2001: Updated

manorit2001:
+ @jforissier

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -82,80 +80,60 @@ static int ti_sci_setup_xfer(uint16_t msg_type, uint32_t msg_flags,
...
-	struct k3_sec_proxy_msg *msg = &xfer->rx_message;
-	struct ti_sci_msg_hdr *hdr = NULL;
+	struct k3_sec_proxy_msg *txmsg = &xfer->tx_message;
+	struct k3_sec_proxy_msg *rxmsg = &xfer->rx_message;
+	struct ti_sci_msg_hdr *txhdr = (struct ti_sci_msg_hdr *)txmsg->buf;
+	struct ti_sci_msg_hdr *rxhdr = (struct ti_sci_msg_hdr *)rxmsg->buf;
+	static uint8_t message_sequence;
+	static struct mutex ti_sci_mutex_lock = MUTEX_INITIALIZER;
 	unsigned int retry = 5;
 	int ret = 0;
```
etienne-lms, line 98: There are still few inconsistencies in the return value type.
Here, `ret` should be of type `TEE_Result` (see `k3_sec_proxy_send()` and  `k3_sec_proxy_recv()`), but `ti_sci_do_xfer()` still returns an `int` value so there should be some conversion somewhere, of callers of `ti_sci_do_xfer()` should be changed to expect a `TEE_Result` value.

manorit2001: I think this will have to be taken up in a separate PR @etienne-lms , this requires a cleanup across the full TISCI and secproxy layer to fix this..

etienne-lms (commented):
In case you really prefer to address the return value issues later:

core/arch/arm/plat-k3/drivers/ti_sci.c
manorit2001: Yeah, I think this would require a proper cleanup so would do it together only in one shot.

manorit2001:
Rebased and updated the acks.

PR #7094, opened 2024-10-24
drivers: caam: skip JR init of CFG_JR_HAB_INDEX

sahilnxp:
On iMX8M SoC, the HAB requires the JR0 to be set to secure world to decrypt the kernel image when loading the image in U-boot.

Before reaching u-boot, OPTEE and ATF set the JR0 to the non-secure domain that leads to a HAB failure when trying to decrypt the kernel.

To fix the issue, this commit introduces CFG_JR_HAB_INDEX that specifies which JR the HAB uses. OPTEE will skip the initialization of CFG_JR_HAB_INDEX and leave it as secure.

It will also disable its usage in the device tree to inform the kernel.

### DIFF
--- a/core/drivers/crypto/caam/hal/common/hal_cfg.c
+++ b/core/drivers/crypto/caam/hal/common/hal_cfg.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright 2017-2019, 2021 NXP
+ * Copyright 2017-2019, 2021, 2024 NXP
  *
  * Brief   CAAM Configuration.
  */
@@ -66,6 +66,12 @@ enum caam_status caam_hal_cfg_get_conf(struct caam_jrcfg *jrcfg)
 
 	retstatus = CAAM_NO_ERROR;
 
+#ifdef CFG_NXP_CAAM_RUNTIME_JR
+	caam_hal_jr_prepare_backup(jrcfg->base, jrcfg->offset);
+#endif
+
+	caam_hal_cfg_hab_jr_mgmt(jrcfg);
+
 exit_get_conf:
 	HAL_TRACE("HAL CFG Get CAAM config ret (0x%x)\n", retstatus);
 	return retstatus;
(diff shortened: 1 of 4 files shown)


core/drivers/crypto/caam/hal/common/hal_cfg.c
```diff
@@ -66,6 +66,12 @@ enum caam_status caam_hal_cfg_get_conf(struct caam_jrcfg *jrcfg)
 
 	retstatus = CAAM_NO_ERROR;
 
+#ifdef CFG_NXP_CAAM_RUNTIME_JR
```
jenswi-linaro, line 69: Please use `if (IS_ENABLED(CFG_NXP_CAAM_RUNTIME_JR))` instead of `#ifdef`.

core/drivers/crypto/caam/hal/common/hal_cfg.c
```diff
@@ -98,3 +110,12 @@ void __weak caam_hal_cfg_setup_nsjobring(struct caam_jrcfg *jrcfg)
 			caam_hal_jr_prepare_backup(jrcfg->base, jr_offset);
 	}
 }
+
+__weak void caam_hal_cfg_hab_jr_mgmt(struct caam_jrcfg *jrcfg __unused)
```
jenswi-linaro, line 114: Why are you using a weak function instead of configuration variables for this configuration?

jenswi-linaro (commented):
Please update the commit message to use the official spelling for OP-TEE, U-Boot, and TF-A.

core/drivers/crypto/caam/hal/common/hal_cfg.c
sahilnxp: Sorry @jenswi-linaro, I don't understand your point.
Can you please give me a reference for this ?

jenswi-linaro: We normally do configuration using CFG_ variables unless it becomes too complicated.
In this case, I can't see what the problem would be, so I'm wondering if it was done for no particular reason.

sahilnxp:
@jenswi-linaro took care of your comments.
Please have a look and let me know your comments.

jenswi-linaro:

sahilnxp:
Thanks @jenswi-linaro,  Tags updated.

PR #7095, opened 2024-10-29
core: fix race in mobj_reg_shm_get_by_cookie()

jenswi-linaro:
Until this patch in mobj_reg_shm_get_by_cookie() there's a small window after cpu_spin_unlock_xrestore() before the reference counter is increased with mobj_get(). Fix that by calling mobj_get() before unlocking reg_shm_slist_lock.

Fixes: b96514926b8e ("core: reference count struct mobj")


etienne-lms (commented):

jenswi-linaro:
Tag applied.

PR #7097, opened 2024-10-29
drivers: imx_csu: add settings for i.MX6

Emantor:
Add the CSU SA settings for i.MX6(Q/D). This setting ensures that no non-TrustZone aware master is able to read secure memory. Information on the CSU SA register values were taken from i.MX6 Security Reference Manual rev 0.

### DIFF
--- a/core/drivers/imx_csu.c
+++ b/core/drivers/imx_csu.c
@@ -72,11 +72,12 @@ const struct csu_setting csu_setting_imx7ds[] = {
 	{(-1), 0},
 };
 
-/* Set all masters to non-secure except the Cortex-A7 */
+/* Set all masters to non-secure except the Cortex-A cores */
+const struct csu_sa_setting csu_sa_imx6 = { 0x15554554, 0x2aaa8aaa };
 const struct csu_sa_setting csu_sa_imx6ul = { 0x10554550, 0x20aa8aa2 };
 const struct csu_sa_setting csu_sa_imx7ds = { 0x15554554, 0x2aaa8aaa };
 
-const struct csu_config csu_imx6 = { NULL, csu_setting_imx6 };
+const struct csu_config csu_imx6 = { &csu_sa_imx6, csu_setting_imx6 };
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Ping @mrvan, @sdininno, @sahilnxp?

sahilnxp:
Hi @jenswi-linaro
Thanks for ping, currently I am on leave due to Diwali in India, I will look at it next week.

Sorry for the delay.

core/drivers/imx_csu.c
```diff
@@ -72,11 +72,12 @@ const struct csu_setting csu_setting_imx7ds[] = {
 	{(-1), 0},
 };
 
-/* Set all masters to non-secure except the Cortex-A7 */
+/* Set all masters to non-secure except the Cortex-A cores */
+const struct csu_sa_setting csu_sa_imx6 = { 0x15554554, 0x2aaa8aa8 };
```
sahilnxp, line 76: Here don't we need to lock the settings also ?
Locking in case of imx7ds and imx6ul

Emantor: I didn't understand the question, but while writing this answer realized what was meant.
Keeping the explanation here. The first block:
```
0x15554554                                                                                                                                                                                    
 28  24  20  16  12   8   4   0
0b10101010101010100010101010100
```
Sets all non-TrustZone aware masters to non-secure except the Cortex-A cores, which is similar to i.MX6ull & imx7ds.
The second block:
```
0x2aaa8aa8
  28  24  20  16  12   8   4   0
0b101010101010101000101010101000
```
than sets the lock bits for every master.

After writing this I guess you meant to ask whether we want to lock the settings for Bit 0, the cortex-A cores?
I'll change that.

sahilnxp: Yes, you got it right.

sahilnxp:
`Reviewed-by: Sahil Malhotra <sahil.malhotra@nxp.com>`

Emantor:
Tag applied, thanks :+1:

PR #7105, opened 2024-11-05
plat-stm32mp1: move sanity of RCC secure state against BSEC state

etienne-lms:
Move implementation that verifies STM32MP1 device Secure Closed state (read from BSEC OTP fuses) against RCC secure hardening configuration. It is moved from shared_resource.c platform driver to platform main.c. This change prepares the removal of shared_resource.c driver that is no longer needed since integration of the firewall framework.


core/arch/arm/plat-stm32mp1/main.c
```diff
@@ -455,6 +460,15 @@ TEE_Result stm32_get_iwdg_otp_config(paddr_t pbase,
 	otp_data->disable_on_standby = otp_value &
 				       BIT(idx + HW2_OTP_IWDG_FZ_STANDBY_SHIFT);
 
+#ifdef CFG_STM32MP15
+	/* Device in Secure Closed state require RCC secure hardening */
+	if (stm32_bsec_get_state(&state))
+		panic();
```
GseoC, line 466: why but it in a iwdg related function?

Either put it in `init_stm32mp1_drivers()` or create a new service?

etienne-lms: Absolutely right, i'll move it here (I have messed something here)

GseoC:
BSEC include could be enclosed the same way as the RCC but I really don't mind leaving it this way. `Reviewed-by: Gatien Chevallier <gatien.chevallier@foss.st.com>`

etienne-lms:
Review tag applied

PR #7109, opened 2024-11-06
crypto: se05x: improve object deletion logs

ldts:
For SE05x, only private keys are stored in the secure element: the OP-TEE secure storage REE/RPMB retains the full public key but just a handle to the private key.

If the secure element's persistent storage is erased, but OP-TEE's secure storage remains, the public key can still be accessed while the private key is inaccessible. However, in such cases, the 'key' will still appear as present in the PKCS#11 database.

When CFG_CORE_SE05X_BLOCK_OBJ_DEL_ON_ERROR is enabled (not by default) and the key pointed to by the handle is not present in the secure element, OP-TEE PKCS#11 clients will encounter an error when attempting to delete the private key information held in the OP-TEE secure storage.

If the setting is disabled, the PKCS#11 storage clears the private key handle without errors.

This commit removes some ambiguity, so users do not see error messages when operations complete successfully.

It also fails on sss_se05x_key_object_init errors unconditionally since a failure on this function can only signify some form of stack corruption.

### DIFF
--- a/core/drivers/crypto/se050/core/storage.c
+++ b/core/drivers/crypto/se050/core/storage.c
@@ -71,21 +71,21 @@ TEE_Result crypto_storage_obj_del(struct tee_obj *o)
 		return TEE_SUCCESS;
 
 	status = sss_se05x_key_object_init(&k_object, se050_kstore);
-	if (status != kStatus_SSS_Success) {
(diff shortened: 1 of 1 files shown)


ldts:
@jenswi-linaro @jforissier tag added

PR #7113, opened 2024-11-06
drivers: caam: check only format bit in operation

sahilnxp:
operation variable has been updated after entering the
function so we cannot check against full RSA operations.
Updated to check only format bit in operation


core/drivers/crypto/caam/include/caam_desc_defines.h
```diff
@@ -317,6 +317,7 @@
 #define PROT_RSA_FMT(format)	SHIFT_U32((PROT_RSA_FMT_##format) & 0x1, 12)
 #define PROT_RSA_FMT_NO		0
 #define PROT_RSA_FMT_PKCS_V1_5	1
+#define PROT_RSA_FMT_MASK SHIFT_U32(0x1, 12)
```
jenswi-linaro, line 320: Please align the defined value, `SHIFT_U32`...

sahilnxp: Took care of comment.

jenswi-linaro:

PR #7121, opened 2024-11-12
drivers: caam: fix CFG_CORE_BIGNUM_MAX_BITS

sahilnxp:
CFG_CORE_BIGNUM_MAX_BITS should be 4576
4096 (RSA Max key size) +
8 * 60 (Header serialization and Black blob overhead in bytes)


jenswi-linaro (commented):

sahilnxp:
Thanks @jenswi-linaro, Tags updated.

PR #7128, opened 2024-11-14
Rework RNG driver design to get RNG from CAAM and introduce CONFIG flag for Prediction Resistance

sahilnxp:
Previous design of keeping RNG data in a buffer and giving random number
to user from that buffer is vulnerable to attacks and also not NIST/FIPS compliant.
So, to make it more secure and NIST/FIPS compliant, will get random
number from CAAM on each user request.

With prediction resistance enabled, on every random number request
CAAM is forced to do reseeding of DRBG, which is time taking process
which leads to lower Random number generation performance.
So to give use the flexibility to enable/disable this feature a flag
CFG_CAAM_RNG_RUNTIME_PR is introduced.
By default it will be disabled and user can enable it as per its
requirement.

### DIFF
--- a/core/drivers/crypto/caam/caam_rng.c
+++ b/core/drivers/crypto/caam/caam_rng.c
@@ -107,7 +107,7 @@ static TEE_Result do_rng_read(uint8_t *buf, size_t len)
 		goto exit;
 	}
 
-	if (rng_privdata->pr_enabled)
+	if (IS_ENABLED(CFG_CAAM_RNG_RUNTIME_PR) && rng_privdata->pr_enabled)
 		op |= ALGO_RNG_PR;
 
 	caam_desc_init(desc);
@@ -310,7 +310,8 @@ end_inst:
 	if (retstatus == CAAM_NO_ERROR) {
 		rng_privdata->instantiated = true;
 		rng_privdata->pr_enabled =
-			caam_hal_rng_pr_enabled(rng_privdata->baseaddr);
+			caam_hal_rng_pr_enabled(rng_privdata->baseaddr) &
+			IS_ENABLED(CFG_CAAM_RNG_RUNTIME_PR);
 
 		RNG_TRACE("RNG prediction resistance is %sabled",
 			  rng_privdata->pr_enabled ? "en" : "dis");
--- a/core/drivers/crypto/caam/caam_rng.c
+++ b/core/drivers/crypto/caam/caam_rng.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /**
- * Copyright 2017-2021 NXP
+ * Copyright 2017-2021, 2024 NXP
  *
  * Brief   CAAM Random Number Generator manager.
  *         Implementation of RNG functions.
@@ -11,6 +11,7 @@
 #include <caam_jr.h>
 #include <caam_rng.h>
 #include <caam_utils_mem.h>
+#include <caam_utils_status.h>
 #include <crypto/crypto.h>
 #include <kernel/panic.h>
 #include <mm/core_memprot.h>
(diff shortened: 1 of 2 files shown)


sahilnxp:
@jenswi-linaro @jforissier Can you please have a look at this one ?

core/drivers/crypto/caam/crypto.mk
```diff
@@ -153,6 +153,14 @@ CFG_CAAM_JR_DISABLE_NODE ?= y
 # 48 * 8 (Black blob overhead in bytes) = 4576 bits
 CFG_CORE_BIGNUM_MAX_BITS ?= 4576
 
+# Define if Prediction Resistance is enabled for every Random number
+# request to CAAM.
+# Performance of getting Random number from CAAM drastically decreases with
+# PR enabled.
+# Users who want CAAM RNG to get reseeded on every Random number request
+# can set this flag to y.
+CFG_CAAM_RNG_RUNTIME_PR ?= n
```
jforissier, line 162: Any particular reason for capitalizing 'Random'? How about:
```
# CAAM RNG Prediction Resistance
# When this flag is y, the CAAM RNG is reseeded on every random number request.
# In this case the performance is drastically reduced, but this makes the process
# compliant with NIST FIPS xxx.
```
(of course replace xxx with the proper reference to the FIPS requirement as requested by Jens)

jforissier (commented):
Hi @sahilnxp,
One comment below, then for both commits:

core/drivers/crypto/caam/crypto.mk
sahilnxp: @jforissier Thanks for the comments.
Getting CAAM every time from CAAM will make the process compliant with NIST/FIPS compliant.

sahilnxp:
> Comments for the commit "drivers: caam: rework RNG driver to get RNG from CAAM on every user request."
> 
> Please remove the period from end of the title, the title is also a bit long. How about: "driver: caam: disable RNG buffering"
> 
> It would be nice with a reference to the relevant part in the NIST or FIPS standard. By the way, what kind of attacks do you anticipate?


Reference is Section 2.7 of  [NIST SP 800-90C ](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90C.4pd.pdf)

Attack can be If we have a pool from where we get the Random number, same pool will be used for Symmetric keys.
Even if we save that key in secure storage, same data is still there in Random number pool which can be exploited.

sahilnxp:
Comments taken care and tags updated.

jenswi-linaro:
For the commit message "driver: caam: disable RNG buffering", please add the section also.

Out of curiosity, how big is the performance impact when removing the buffering? Is it noticeable?

sahilnxp:
> For the commit message "driver: caam: disable RNG buffering", please add the section also.

Done.

> Out of curiosity, how big is the performance impact when removing the buffering? Is it noticeable?

Yes, it is noticeable.
In term of time taken by `xtest` to complete
**Buffered approach** completes in around **3 mins**
**Non-buffered approach** it completed in **14 mins**
Most of the time is increased because of regression_1006 (Multiple Random number generation calls are there in this test)

Test was done on i.MX8MM.

jenswi-linaro:
Buffering or no buffering is a tradeoff between security and speed. It would be interesting to see if we could add a generic layer to handle this, keeping in mind the NIST recommendation for the different classes.

Anyway, please apply:

sahilnxp:
> Buffering or no buffering is a tradeoff between security and speed. It would be interesting to see if we could add a generic layer to handle this, keeping in mind the NIST recommendation for the different classes.
> 
Yeah, you are right..

Thanks @jenswi-linaro Tags added.

core/drivers/crypto/caam/caam_rng.c
```diff
@@ -107,7 +107,7 @@ static TEE_Result do_rng_read(uint8_t *buf, size_t len)
 		goto exit;
 	}
 
-	if (rng_privdata->pr_enabled)
+	if (IS_ENABLED(CFG_CAAM_RNG_RUNTIME_PR) && rng_privdata->pr_enabled)
```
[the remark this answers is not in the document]
sahilnxp: Thanks @etienne-lms for catching this.
Updated the pull request.

PR #7132, opened 2024-11-15
sign_encrypt.py: fix typo in stitch usage message

etienne-lms:
Fix typo in stitch usage message.


etienne-lms:
Review tag applied.

Issue #7133, opened 2024-11-19
MbedTLS upgrade to 3.6.2


ysbnim:
Hi,

I'd like to upgrade MbedTLS upgrade to the latest version, 3.6.2
Because we found MbedTLS 3.6.0 has several CVEs as the follows :
- CVE-2024-45159
- CVE-2024-45158
- CVE-2024-45157
- CVE-2024-49195

Actually, To upgrade to 3.6.2, I already made the patch : https://github.com/ysbnim/optee_os/tree/import/mbedtls-3.6.2
by referring the PR https://github.com/OP-TEE/optee_os/pull/6797.

But there is no import/mbedtls-3.6.2 I can make a pull request of this.

Could you please help me create import/mbedtls-3.6.2 branch?
Or please give me any opinions about this.

Thanks!

jenswi-linaro:
Thanks for taking care of this. I've created the branch `import/mbedtls-3.6.2`.

PR #7135, opened 2024-11-19
Import/mbedtls 3.6.2

ysbnim:
This pull request imports MbedTLS 3.6.2 to fix CVEs (Related issue : https://github.com/OP-TEE/optee_os/issues/7133)
This also rebases several commits according to PR https://github.com/OP-TEE/optee_os/pull/6797.


etienne-lms:
The [changelog of mbedtls 3.6.1 mentions the issue you reported](https://github.com/Mbed-TLS/mbedtls/blob/v3.6.2/ChangeLog#L152-L153):
> * Fix unintended performance regression when using short RSA public keys.
     Fixes #9232.

So I guess it should be fixed with 3.6.2.

etienne-lms:
Tested on my 32bit platform: `xtest regression_4011` is very slow without commit 
"libmbedtls: bignum: restore mbedtls_mpi_exp_mod() from v3.5.2" applied.

ysbnim:
Tested `time xtest 4011` on my machine. But without commit, QEMUv7 was extremely slow.

  |   | QEMUv8 | QEMUv7
-- | -- | -- | --
3.6.0 | w/ commit | 1.87s | 25.22s
3.6.0 | w/o commit | 38.25s | 14m 52.65s
3.6.2 | w/ commit | 1.87s | 25.58s
3.6.2 | w/o commit | 41.54s | 14m 30.97s

jforissier:
Commit "[libmbedtls: bignum: restore mbedtls_mpi_exp_mod() from v3.5.2](https://github.com/OP-TEE/optee_os/pull/7135/commits/832aca4de4f396b65bbf65435720feae086ab2cc)" should be replaced by this one: "[libmbedtls: use 'unsafe' algorithm for mbedtls_mpi_exp_mod()](https://github.com/OP-TEE/optee_os/commit/5f3424c31ebc178265f4d383f9438d406af31fe6).

ysbnim:
Checked that it takes 26.12s on my machine. Thanks @jforissier !

etienne-lms:
Test OK on my 32bit platforms. I see the 'unsafe' implementation seems <10% slower compared to the patch that restore the mbedTLS v3.5.2 implementation, but that is not an issue at all.

I wonder whether there should be a config switch to define which of the safe/unsafe modular exponentiation scheme to embed in mbedTLS? It should somewhat clear that a timing-attack defense of mbedTLS is withdrawn when the 'unsafe' function is used.
If so, should there be one config switch for the core and one for mbedTLS as a TA lib? (e.g. `CFG_CORE_SAFE_MODEXP` / `CFG_TA_MBEDTLS_SAFE_MODEXP`).

ysbnim:
How about intoducing new `CFG_CORE_UNSAFE_MODEXP` and then adding it only to ci check for QEMUv7?

I think QEMUv7 is the only environment that needs unsafe mod exp. 64bit QEMU also has some performance regression but it would be acceptible.

jforissier:
@ysbnim sounds good to me. Would you please update this PR to do just that? Thanks!

etienne-lms:
> How about intoducing new CFG_CORE_UNSAFE_MODEXP and then adding it only to ci check for QEMUv7?

I agree, but 32bit platform maintainers should also be warned. I think i'll also default enable this switch in `plat-stm32mp1/conf.mk` to run xtest in a decent amout of time. On these platforms, using the 'safe' implementation also affects `regression_4006` (1min=>6min) and `pkcs11_1021/1022/1023/1026` (`xtest -t pkcs11` lasts 20min instead of 3min). I've test the GP tests: the difference is acceptable: 9min instead of 5min, the extra 4min being spread over many `gp_50xxx` tests.

mk/config.mk
```diff
@@ -641,7 +641,7 @@ endif
 #   is a totally standard shared object, and should be used to link against.
 #   The '.ta' file is a signed version of the '.so' and should be installed
 #   in the same way as TAs so that they can be found at runtime.
-CFG_ULIBS_SHARED ?= n
+CFG_ULIBS_SHARED ?= y
```
etienne-lms, line 644: unrelated change. Prefer discard it.

mk/config.mk
```diff
@@ -1236,3 +1236,9 @@ CFG_SEMIHOSTING ?= n
 # writes trace messages via FFA interface to the SPM (Secure Partition Manager)
 # like hafnium.
 CFG_FFA_CONSOLE ?= n
+
+# CFG_CORE_UNSAFE_MODEXP, when enabled, makes modular exponentiation uses
+# 'unsafe' algorithm having better performance. To resist against timing
+# attacks, 'safe' one is designed to take constant-time that is generally much
+# slower
```
etienne-lms, line 1243: Add a terminal dot.

etienne-lms (commented):
`CFG_CORE_UNSAFE_MODEXP`, as per its name, relates to OP-TEE core but the switch is also applied when libmbedtls is built as a user land trusted library.
It would be nice the 2 configs are distinguished. Or the swith may be renamed `CFG_MBEDTLS_UNSAFE_MODEXP` but IMHO 2 config switches would be better. @jforissier, what do you think?

mk/config.mk
ysbnim: Sorry. Got confused while doing something else.

mk/config.mk
ysbnim: Done

lib/libmbedtls/mbedtls/library/bignum.c
```diff
@@ -1843,7 +1843,11 @@ int mbedtls_mpi_exp_mod(mbedtls_mpi *X, const mbedtls_mpi *A,
                         const mbedtls_mpi *E, const mbedtls_mpi *N,
                         mbedtls_mpi *prec_RR)
 {
+#ifdef CFG_CORE_UNSAFE_MODEXP
+    return mbedtls_mpi_exp_mod_unsafe(X, A, E, N, prec_RR);
+#else
     return mbedtls_mpi_exp_mod_optionally_safe(X, A, E, MBEDTLS_MPI_IS_SECRET, N, prec_RR);
+#endif
```
etienne-lms, line 1850: Regarding my comment https://github.com/OP-TEE/optee_os/pull/7135#pullrequestreview-2451844869, what about:
```c
#if (defined(__KERNEL__) && defined(CFG_CORE_UNSAFE_MODEXP)) || \
    (!defined(__KERNEL__) && defined(CFG_TA_MEBDTLS_UNSAFE_MODEXP))
    return mbedtls_mpi_exp_mod_unsafe(X, A, E, N, prec_RR);
#else
    return mbedtls_mpi_exp_mod_optionally_safe(X, A, E, MBEDTLS_MPI_IS_SECRET, N, prec_RR);
#endif

Comment (ysbnim) [reply lib/libmbedtls/mbedtls/library/bignum.c:1851]: Changed to your suggestion

File: mk/config.mk [hunk 4]
```diff
@@ -1236,3 +1236,14 @@ CFG_SEMIHOSTING ?= n
...
+
+# CFG_CORE_UNSAFE_MODEXP, when enabled, makes modular exponentiation on TEE
+# core use 'unsafe' algorithm having better performance. To resist against
+# timing attacks, 'safe' one is designed to take constant-time that is
+# generally much slower.
+CFG_CORE_UNSAFE_MODEXP ?= n
+
+# CFG_TA_MBEDTLS_UNSAFE_MODEXP, similar to CFG_CORE_UNSAFE_MODEXP,
+# when enabled, makes MBedTLS library for TAs use 'unsafe' modular
+# exponentiation algorthim.
```
jforissier, line 1248: algorithm

jforissier (commented):
No need for a blank line after the `Link:` tags.
With my comment below addressed, please add:
```
Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`
```
...to commit "libmbedtls: add CFG_CORE_UNSAFE_MODEXP and CFG_TA_MEBDTLS_UNSAFE_MODEXP".

Thanks!

mk/config.mk
ysbnim: FIxed

ysbnim:
Removed blank line and add review tag

etienne-lms (commented):
"libmbedtls: add CFG_CORE_UNSAFE_MODEXP and CFG_TA_MEBDTLS_UNSAFE_MODEXP".

jforissier:
@ysbnim thank you for your contribution. I will now merge this PR into the import branch, and I will later take care of creating the squashed commit and pull request to merge into master.

jforissier:
https://github.com/OP-TEE/optee_os/pull/7149

PR #7150, opened 2024-11-25
drivers: caam: sm: fix SM partition permission in SMAG registers

sahilnxp:
SM partition SMAG1 permissions were wrongly set for generating DEK blob which results in error while decapsulating DEK blob during HAB encrypted boot. Setting the permissions correctly fix this issue.


jenswi-linaro (commented):
Please add a `Fixes:` for the commit that introduced the error, with that:

sahilnxp:
@jenswi-linaro Any more comments on this one ?

jforissier:
@sahilnxp thanks!

PR #7163, opened 2024-12-03
ci: split QEMUv8_check job

jenswi-linaro:
Split the QEMUv8_check job into two smaller jobs to allow them to run in parallel.


jenswi-linaro:
I'm investigating the Xen FF-A failure. It's obviously not caused by this PR.

jforissier:

It would be nice to factor out what is common in several jobs, but I don't know how to do it in YAML.

jenswi-linaro:
Tag applied

PR #7167, opened 2024-12-05
libutils: util.h: add usage description to {get,set}_field_u{32,64}()

meremST:
Hello everyone,

I propose to add a clearer description of get and set field function at there misunderstanding can be error prone.

Cheers,

Maxime

commit message:

Added comments to get_field_u{32,64}() and set_field_u{32,64}() to improve understanding and reduce errors.

### DIFF
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -188,7 +188,40 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/*
+ * Functions to get and set bit fields in a 32/64-bit bitfield.
+ *
+ * These helper functions allow setting and extracting specific bits in
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the bitfield should be updated or extracted.
+ * These functions exist in both 32-bit and 64-bit versions.
+ *
+ * set_field_u32()
+ * set_field_u64() - Modifies specific bits in a bitfield by clearing
+ *		     the bits specified by the mask and then setting
+ *		     these bits to the new value @val.
+ * @reg:  The original 32-bit or 64-bit bitfield value.
+ * @mask: A bitmask indicating which bits in the bitfield should be
+ *	  updated.
+ * @val:  The new value to be loaded into the bitfield, left shifted
+ * 	  according to @mask rightmost non-zero bit position.
+ * Returns the updated bitfield value with the specified bits set to
+ * the new value.
+ *
+ * E.g. set_bitfield_u32(0x123456, 0xf0ff00, 0xabcd) returns 0xa2cd56.
+ *
+ * get_field_u32()
+ * get_field_u64() - Extracts the value of specific bits in a bitfield
+ *		     by isolating the bits specified by the mask and
+ *		     then shifting them to the rightmost position.
+ * @reg:  The original 32-bit or 64-bit bitfield value.
+ * @mask: A bitmask indicating which bits in the bitfield should be
+ *	  extracted.
+ * Returns the value of the bits specified by the mask, shifted to the
+ * @mask rightmost non-zero bit position.
+ *
+ * E.g. get_bitfield_u32(0x123456, 0xf0ff00) returns 0x1034.
+ */
 static inline uint32_t get_field_u32(uint32_t reg, uint32_t mask)
 {
 	return (reg & mask) / (mask & ~(mask - 1));


lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/**
+ * @brief Functions to get and set bit fields in a 32/64-bit register.
```
jenswi-linaro, line 192: Please drop the doxygen tags. These comments are read in place and should be optimized for human reading.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * These inline functions allow setting and extracting specific bits in
+ * a register (`reg`) according to a given mask (`mask`). The mask
```
jenswi-linaro, line 195: s/\`reg\`/@reg/
s/\`mask\`/@mask/

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ * specifies which bits in the register should be updated or extracted.
+ *
+ * - `set_field`: Modifies specific bits in a register by clearing the
```
jenswi-linaro, line 198: s/- \`set_field\`/set_field*()/
Please indent the following two lines so they align with the start of the word "Modifies".

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *   bits specified by the mask and then setting these bits to the new
+ *   value (`val`).
+ * - `get_field`: Extracts the value of specific bits in a register by
+ *   isolating the bits specified by the mask and then shifting them to
+ *   the rightmost position.
+ *
+ * @param reg The original 32-bit or 64-bit register value.
```
jenswi-linaro, line 205: s/@​param /@/
If the description doesn't fit on a single line, please indent the following lines to align with the first word of the description.

jenswi-linaro (commented):
Thanks for taking the time to document a few functions. I've added a few suggestions on how to make it easier to read.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/**
+ * @brief Functions to get and set bit fields in a 32/64-bit register.
+ *
+ * These inline functions allow setting and extracting specific bits in
```
etienne-lms, line 194: Maybe drop `inline`. Could replace with `helper`: `These helper functions allow ...`

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,33 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/**
+ * Functions to get and set bit fields in a 32/64-bit register.
```
etienne-lms, line 192: I would replace `register` with `value` (several occurrences) . These helper functions are not specifically designed for values read from/written to hardware registers, they apply to any 32bit/64bit values.

lib/libutils/ext/include/util.h
meremST: Hello Jens,

Thank you for your remarks. I have taken them into account and I've improved the rest of the text to make it coherent with your recommendations.

cheers

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,33 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * These inline functions allow setting and extracting specific bits in
+ * a register @reg according to a given mask @mask. The mask
+ * specifies which bits in the register should be updated or extracted.
+ *
+ * - set_field*(): Modifies specific bits in a register by clearing the
```
etienne-lms, line 198: Maybe `set_field_u*()`.

No need to tell bits are cleared before being set, they are just set.
By the way, the implementation could be slightly improved:
```patch
-	return (reg & ~mask) | (val * (mask & ~(mask - 1)));
+	return reg | (val * (mask & ~(mask - 1)));
```

I think you should also tell that bits to be set are shifted to \@mask rightmost position.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,33 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * @reg The original 32-bit or 64-bit register value.
+ * @mask A bitmask indicating which bits in the register should be
+ * updated or extracted.
+ * @val The new value to be inserted into the register at the
+ * positions specified by the mask (only for set_field*() functions).
+ * These functions returns the updated register value with the specified
+ * bits set to the new value (for set_field*() functions), or the value
+ * of the bits specified by the mask, shifted to the rightmost position
+ * (for get_field*() functions).
```
etienne-lms, line 213: To make the description clearer, I would suggest to explicitly split the *get_field_uXX()* et *set_field_uXX()* descriptions:
```c
/*
 * These inline functions allow ...
 *
 * set_field_u32() / set_field_u64() - Modifies specific bits in ...
 * @reg: ...
 * @mask: ...
 * @val: ...
 * @return: ...
 *
 * get_field_u32() / get_field_u64() - Extracts the value of specific bits in 
 * @reg: ...
 * @mask: ...
 * @return: ...
 */
```
This is only a suggestion. In the end, what's important is that these information are as readable and clear as possible.

lib/libutils/ext/include/util.h
meremST: I agree with you these functions could be use to manipulate 32/64 bit value but they are intended for use on registers that why the first argument is named "reg". So I think it is better keep the register name to avoid confusion.

lib/libutils/ext/include/util.h
meremST: I don' t think your improvement might works. If you are removing the "& ~mask" the function will no longer "erease" the bits already present where the mask is placed.

example set 1010 on 1000011110000000 at position BIT(8) :

with the original :   1000010100000000
with your version :  1000011110000000

etienne-lms: Right, my mistake. Discard this part of my comment.

lib/libutils/ext/include/util.h
etienne-lms: > ... they are intended for use on registers ...

I don't agree. One can use them for some other purpose, e.g. extracting bits from a DT config argument or any memory cell.

meremST: OK I propose to simply replace register by bitfield

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/**
```
jenswi-linaro, line 191: Please drop the extra `*`.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * These helper functions allow setting and extracting specific bits in
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the register should be updated or extracted.
+ * These functions exist in both 32-bit and 64-bit versions
+ *
+ * set_field_u32() / set_field_u64() - Modifies specific bits in a
+ * bitfield by clearing the bits specified by the mask and then setting
+ * these bits to the new value @val.
+ * @reg: The original 32-bit or 64-bit bitfield value.
```
jenswi-linaro, line 202: How about:
```
 * @reg:  The original 32-bit or 64-bit bitfield value.
 * @mask: A bitmask indicating which bits in the bitfield should be
 *        updated or extracted.
 * @val:  The new value to be inserted into the bitfield at the
 *        positions specified by the mask.
 * Returns the updated bitfield value with the specified bits set to
 * the new value.
```

meremST: Agreed, much clearer

jenswi-linaro:
Looks good.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the register should be updated or extracted.
+ * These functions exist in both 32-bit and 64-bit versions
+ *
+ * set_field_u32() / set_field_u64() - Modifies specific bits in a
+ * bitfield by clearing the bits specified by the mask and then setting
+ * these bits to the new value @val.
+ * @reg:  The original 32-bit or 64-bit bitfield value.
+ * @mask: A bitmask indicating which bits in the bitfield should be
+ *	  updated or extracted.
```
etienne-lms, line 204: Remove `extracted`.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ * @val:  The new value to be inserted into the bitfield at the
+ *	  positions specified by the mask.
```
etienne-lms, line 206: `... at the positions specified by @mask rightmost non-zero bit.`

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ * Returns the updated bitfield value with the specified bits set to
+ * the new value.
```
etienne-lms, line 208: Maybe an exemple could help: `E.g. set_bitfield_u32(0x123456, 0xf0ff00, 0xabcd) returns 0xa2cd56.`

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * get_field_u32() / get_field_u64() - Extracts the value of specific
+ * bits in a bitfield by isolating the bits specified by the mask and
+ * then shifting them to the rightmost position.
+ * @reg:  The original 32-bit or 64-bit bitfield value.
+ * @mask: A bitmask indicating which bits in the bitfield should be
+ *	  updated or extracted.
+ * Returns the value of the bits specified by the mask, shifted to the
+ * rightmost position.
```
etienne-lms, line 217: Maybe an exemple could help: `E.g. get_bitfield_u32(0x123456, 0xf0ff00) returns 0x1034.`

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ * Functions to get and set bit fields in a 32/64-bit register.
+ *
+ * These helper functions allow setting and extracting specific bits in
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the register should be updated or extracted.
+ * These functions exist in both 32-bit and 64-bit versions
+ *
+ * set_field_u32() / set_field_u64() - Modifies specific bits in a
+ * bitfield by clearing the bits specified by the mask and then setting
+ * these bits to the new value @val.
```
etienne-lms, line 201: What about;
```
 * set_bitfield_u32()
 * set_bitfield_u64() - Modify bit fields defined by @mask to @val shifted value
 * @reg: ...
 * @...
 * 
 * set_bitfield_u32()
 * get_bitfield_u64() - Extract/right shift value of bit fields defined by @mask
 * @reg: ...
 * @...
 */

File: lib/libutils/ext/include/util.h [hunk 16]
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *	  positions specified by the mask.
+ * Returns the updated bitfield value with the specified bits set to
+ * the new value.
+ *
+ * get_field_u32() / get_field_u64() - Extracts the value of specific
+ * bits in a bitfield by isolating the bits specified by the mask and
+ * then shifting them to the rightmost position.
+ * @reg:  The original 32-bit or 64-bit bitfield value.
+ * @mask: A bitmask indicating which bits in the bitfield should be
+ *	  updated or extracted.
```
etienne-lms, line 215: remove `updated`

lib/libutils/ext/include/util.h
etienne-lms: s/register/value/?

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/*
+ * Functions to get and set bit fields in a 32/64-bit register.
+ *
+ * These helper functions allow setting and extracting specific bits in
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the register should be updated or extracted.
```
etienne-lms, line 196: s/register/input 32bit or 64bit data/

etienne-lms (commented):
In commit message: 

s/_field_u/_bitfield_u/g
s/Added comments/Add inline description comments/



I think it would be worth using explicit full function labels:

> Add inline description comments to get_bitfield_u32(), get_bitfield_u64(),
set_bitfield_u32() and set_bitfield_u64() to improve understanding and
reduce errors.

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,34 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
 	*reg1 = low32_from_64(val);
 }
 
-/* Get and set bit fields  */
+/*
+ * Functions to get and set bit fields in a 32/64-bit register.
```
etienne-lms, line 192: s/register/data/?

jforissier:
> In commit message:
> 
> s/_field_u/_bitfield_u/g s/Added comments/Add inline description comments/
> 
> I think it would be worth using explicit full function labels:
> 
> > Add inline description comments to get_bitfield_u32(), get_bitfield_u64(),
> > set_bitfield_u32() and set_bitfield_u64() to improve understanding and
> > reduce errors.

I agree. The subject is fine with the abbreviations (it should be kept short), but expanding the function names in the description makes it easier to find the functions, for example with `git log --grep=get_bitfield` etc.

meremST:
> > In commit message:
> > s/_field_u/_bitfield_u/g s/Added comments/Add inline description comments/
> > I think it would be worth using explicit full function labels:
> > > Add inline description comments to get_bitfield_u32(), get_bitfield_u64(),
> > > set_bitfield_u32() and set_bitfield_u64() to improve understanding and
> > > reduce errors.
> 
> I agree. The subject is fine with the abbreviations (it should be kept short), but expanding the function names in the description makes it easier to find the functions, for example with `git log --grep=get_bitfield` etc.

OK I've updated the commit message

lib/libutils/ext/include/util.h
meremST: done

lib/libutils/ext/include/util.h
meremST: OK I've updated the text to remove register

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,40 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * These helper functions allow setting and extracting specific bits in
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the bitfield should be updated or extracted.
+ * These functions exist in both 32-bit and 64-bit versions
+ *
+ * set_field_u32()
+ * set_field_u64() - Modifies specific bits in a bitfield by clearing
+ * the bits specified by the mask and then setting these bits to the new
+ * value @val.
```
etienne-lms, line 202: Indent so that it is readable. I think a single line description is simpler :(
```c
 * set_field_u32()
 * set_field_u64() - Modifies specific bits in a bitfield by clearing the bits
 *                   specified by the mask and then setting these bits to the
 *                   new value @val.

File: lib/libutils/ext/include/util.h [hunk 20]
```diff
@@ -188,7 +188,40 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ * 	  according to @mask rightmost non-zero bit position.
+ * Returns the updated bitfield value with the specified bits set to
+ * the new value.
+ *
+ * E.g. set_bitfield_u32(0x123456, 0xf0ff00, 0xabcd) returns 0xa2cd56.
+ *
+ * get_field_u32()
+ * get_field_u64() - Extracts the value of specific bits in a bitfield
+ * by isolating the bits specified by the mask and then shifting them to
+ * the rightmost position.
```
etienne-lms, line 216: Ditto:
```c
 * get_field_u32()
 * get_field_u64() - Extracts the value of specific bits in a bitfield by
 *                   isolating the bits specified by the mask and then shifting
 *                   them to the rightmost position.

File: lib/libutils/ext/include/util.h [hunk 21]
```diff
@@ -188,7 +188,40 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
 }
 
-/* Get and set bit fields  */
+/*
+ * Functions to get and set bit fields in a 32/64-bit bitfield.
+ *
+ * These helper functions allow setting and extracting specific bits in
+ * a bitfield @reg according to a given mask @mask. The mask
+ * specifies which bits in the bitfield should be updated or extracted.
+ * These functions exist in both 32-bit and 64-bit versions
```
etienne-lms, line 197: Add a period.

lib/libutils/ext/include/util.h
meremST: OK for indentation. The single line description would make the text harder to read in my view.

lib/libutils/ext/include/util.h
meremST: ack

lib/libutils/ext/include/util.h
```diff
@@ -188,7 +188,40 @@ static inline void reg_pair_from_64(uint64_t val, uint32_t *reg0,
...
+ *
+ * get_field_u32()
+ * get_field_u64() - Extracts the value of specific bits in a bitfield
+ *		     by isolating the bits specified by the mask and
+ *		     then shifting them to the rightmost position.
+ * @reg:  The original 32-bit or 64-bit bitfield value.
+ * @mask: A bitmask indicating which bits in the bitfield should be
+ *	  extracted.
+ * Returns the value of the bits specified by the mask, shifted to the
+ * rightmost position.
```
etienne-lms, line 221: For consistency:
```patch
- * rightmost position.
- * @mask rightmost non-zero bit position.

etienne-lms (commented):
With last comment addressed or not:

lib/libutils/ext/include/util.h
meremST: done

etienne-lms:
@meremST, could you [add in your commit message](https://optee.readthedocs.io/en/latest/general/contribute.html#finalizing-your-contribution) the review tags @jenswi-linaro (https://github.com/OP-TEE/optee_os/pull/7167#issuecomment-2523327809, I think it still applies) and I (review comment right above) posted? Once done, our dear maintainers will be able to merge you changes.

etienne-lms:
@jforissier, I think this P-R can be merged.

PR #7170, opened 2024-12-09
dts: stm32: disable stm32mp15 SD switch regulator node

etienne-lms:
SD switch regulator is not used by OP-TEE on STM32MP15 based boards hence disable this node in the OP-TEE secure device tree for boards DHCOR Avenger96 (stm32mp15xx-dhcor-avenger96.dtsi) ST ED1/EV1 (stm32mp157c-ed1.dts).

This change fixes a issue related to the integration of stm32_gpio driver as a firewall controller, which is highlighted by ab error trace message like:

E/TC:0 0 stm32_gpio_get_dt:837 node regulator-sd_switch requests secure GPIO F14 that cannot be secured E/TC:0 0 Panic

Fixes: 4675225ed84f ("drivers: stm32_gpio: check secure state of consumed GPIOs")

### DIFF
--- a/core/arch/arm/dts/stm32mp157c-ed1.dts
+++ b/core/arch/arm/dts/stm32mp157c-ed1.dts
@@ -94,6 +94,7 @@
 		gpios-states = <0>;
 		states = <1800000 0x1>,
 			 <2900000 0x0>;
+		status = "disabled";
 	};
 
 	vin: vin {
--- a/core/arch/arm/dts/stm32mp15xx-dhcor-avenger96.dtsi
+++ b/core/arch/arm/dts/stm32mp15xx-dhcor-avenger96.dtsi
@@ -84,6 +84,7 @@
 		gpios-states = <0>;
 		states = <1800000 0x1>,
 			 <2900000 0x0>;
+		status = "disabled";
 	};
 
 	sound {


core/arch/arm/dts/stm32mp157c-ed1.dts
```diff
@@ -90,10 +90,14 @@
 		regulator-type = "voltage";
 		regulator-always-on;
 
+//		gpios = <&gpiof 14 (GPIO_STM32_NON_SECURE |GPIO_ACTIVE_HIGH)>;
```
jforissier, line 93: Why add this line?

etienne-lms: My mistake! A leftover from few local tests. I'll remote it. Thanks.

core/arch/arm/dts/stm32mp157c-ed1.dts
```diff
@@ -90,10 +90,14 @@
...
 		gpios = <&gpiof 14 GPIO_ACTIVE_HIGH>;
+
 		gpios-states = <0>;
 		states = <1800000 0x1>,
 			 <2900000 0x0>;
+		/* SD switch is not used by OP-TEE */
```
etienne-lms, line 99: I think I can also remove this inline comment. The Git log should be enough.

jforissier: Agreed.

jforissier:

etienne-lms:
@jneuhauser, maybe you want to have a look at this change.

etienne-lms:
There seems to be test issues on QEMUv8 with recent P-Rs, as for this one. Looking at the logs, it's sometime a pkcs11_1021.x (not always the same: 2_2, Xen FF-A, Clang here), sometimes regression_4006.7 (Xen here). 
I don't reproduce these issues running QEMUv8 test locally. Could it be some spurious processing time based issue?

jforissier:
@etienne-lms yes there is something wrong. I have run my release test suite for QEMUv8. 14 configurations passed and 6 failed and not in the same way (regression_4006.7 TEE_ERROR_TARGET_DEAD, pkcs11_1023.21 CKR_GENERAL_ERROR, gp_50049 TEE_ERROR_TARGET_DEAD, pkcs11_1022.16 CKR_GENERAL_ERROR, xtest pkcs11_1021.10 CKR_GENERAL_ERROR...). It seems `make -j$(nproc) check XTEST_ARGS="-l 14"` will fail most of the time. I will investigate more.

jforissier:
It looks like `MEASURED_BOOT_FTPM=y` is the culprit. Or at least it triggers something bad.

etienne-lms:
Review tag applied.

PR #7171, opened 2024-12-10
drivers: scmi-msg: fix clock min/max/step triplet description support

TonyHan11:
The return value of [plat_scmi_clock_rates_array()](https://github.com/OP-TEE/optee_os/blob/eaa924efbcf5f567a0c528f29cdd386f49b7ee82/core/drivers/scmi-msg/clock_generic.c#L79) is one of "SCMI_DENIED", "SCMI_GENERIC_ERROR" and "SCMI_SUCCESS".

[The code in scmi_clock_describe_rates() for clock min/max/step triplet description support](https://github.com/OP-TEE/optee_os/blob/eaa924efbcf5f567a0c528f29cdd386f49b7ee82/core/drivers/scmi-msg/clock.c#L326) would never be executed due to the return value of plat_scmi_clock_rates_array() could never be "SCMI_NOT_SUPPORTED".

Fix by modified the return value of plat_scmi_clock_rates_array().


etienne-lms (commented):

TonyHan11:
The review tag added.

PR #7175, opened 2024-12-11
[DO NOT MERGE] ci: QEMUv8: set MEASURED_BOOT_FTPM=n

jforissier:
Disable fTPM to check if this un-breaks CI.


jforissier:
https://github.com/OP-TEE/build/pull/795

PR #7176, opened 2024-12-11
ta: pkcs11: default disable raw RSA mechanism, default enable in plat-vexpress

etienne-lms:
Raw RSA signature (PKCS#11 mechanism CKM_RSA_X_509) is not a recommended scheme because unsafe when used with unsafe padding schemes. It's been added in the pkcs11 TA because needed for some TLS v1.2  scenario. I propose to disable it in the pkcs11 TA default config to prevent confusion but enable it in **plat-vexpress** default config so that it is tested (build + xtest).

### DIFF
--- a/core/arch/arm/plat-vexpress/conf.mk
+++ b/core/arch/arm/plat-vexpress/conf.mk
@@ -172,3 +172,4 @@ endif
 endif
 
 CFG_PKCS11_TA ?= y
+CFG_PKCS11_TA_RSA_X_509 ?= y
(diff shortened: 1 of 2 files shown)


jforissier:

Preferably with `s/pkcs11 TA/PKCS#11 TA` in the second commit description.

etienne-lms:
Commit message fixed and review tag applied.

etienne-lms:
I've triggered a CI test round since recent merges in build.git should now lead to an all-success status.

PR #7179, opened 2024-12-12
ta: pkcs11: fix CKM_RSA_X_509 signature size verification

etienne-lms:
When using raw RSA decryption to generate a raw RSA signature, the decrypted message (that is our generated signature) may be shorter than required since RSA may have stripped out leading zero bytes. Fix the PKCS#11 TA implementation to insert them back to reach the target signature size.

Fixes: e02f17f374b6 ("ta: pkcs11: add CKM_RSA_X_509 authentication")

### DIFF
--- a/ta/pkcs11/src/processing_asymm.c
+++ b/ta/pkcs11/src/processing_asymm.c
@@ -899,14 +899,16 @@ enum pkcs11_rc step_asymm_operation(struct pkcs11_session *session,
 						    tee_attrs, tee_attrs_count,
 						    in_buf, in_size,
 						    temp_buf, &temp_size);
-			if (!res && temp_size != sz) {
-				EMSG("CMK_RSA_X509: signature size %zu != %zu",
-				     temp_size, sz);
-				rc = PKCS11_CKR_DATA_INVALID;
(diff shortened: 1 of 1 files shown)


etienne-lms:
@Emantor, @vesajaaskelainen, may I ask you to have a look at this P-R?

Emantor (approved):

etienne-lms:
Thanks @Emantor, I've applied your `Acked-by` tag to the commit.

etienne-lms:
Hi @vesajaaskelainen, this support was needed in OP-TEE for system embedding some specific versions of OpenSSL 3.x.y. We saw this issue on a release embedding OpenSSL 3.1.4.

The issue seems related to https://github.com/openssl/openssl/issues/23397 that was fixed with https://github.com/openssl/openssl/pull/23063 which reverts a patch from Oct'2023. The issue is that some releases of OpenSSL have the faulty patch applied and not yet reverted: 3.0.{12|13}, 3.1.{4|5}, 3.2.{0|1|2}.

etienne-lms:
Please keep alive.

PR #7183, opened 2024-12-16
libutils: util.h: corrections on some ROUND*() macros

etienne-lms:
Correct inline description comments for ROUND*() macros that request an operand to be a power of 2, not a multiple of 2.
Assert this condition macros where assertion are applicable.
Propose a implementation for relax this constraint for `ROUNDUP()` and `ROUNDDOWN()` macros where we can't use assertions.

### DIFF
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -146,7 +146,7 @@
 	}))
 
 /*
- * Round up the result of x / y to the nearest upper integer if result is not 
+ * Round up the result of x / y to the nearest upper integer if result is not
  * already an integer.
  */
 #define DIV_ROUND_UP(x, y) (((x) + (y) - 1) / (y))
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -156,11 +156,11 @@
 	(__extension__ ({ __typeof__(x) _x = (x); \
 	  __typeof__(y) _y = (y); \
 	  (_x + (_y / 2)) / _y; }))
-#else
+#else /* __ASSEMBLER__ */
 #define ROUNDUP(x, y)			((((x) + (y) - 1) / (y)) * (y))
-#define ROUNDDOWN(x, y)		(((x) / (y)) * (y))
+#define ROUNDDOWN(x, y)			(((x) / (y)) * (y))
 #define UDIV_ROUND_NEAREST(x, y)	(((x) + ((y) / 2)) / (y))
-#endif
+#endif /* __ASSEMBLER__ */
 
 /* x has to be of an unsigned type */
 #define IS_POWER_OF_TWO(x) (((x) != 0) && (((x) & (~(x) + 1)) == (x)))
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -135,8 +135,15 @@
  */
 #define ROUNDUP_DIV(x, y) (ROUNDUP((x), (y)) / (__typeof__(x))(y))
 
+/* Round down the even multiple of size */
+#define ROUNDDOWN(x, y)		(((x) / (__typeof__(x))(y)) * (__typeof__(x))(y))
+
 /* Round down the even multiple of size, size has to be a power of 2 */
-#define ROUNDDOWN(v, size) ((v) & ~((__typeof__(v))(size) - 1))
+#define ROUNDDOWN2(v, size) \
+	(__extension__({ \
+		assert(IS_POWER_OF_TWO(size)); \
+		((v) & ~((__typeof__(v))(size) - 1)); \
+	}))
 
 /*
  * Round up the result of x / y to the nearest upper integer if result is not 
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -9,6 +9,7 @@
 #include <inttypes.h>
 
 #ifndef __ASSEMBLER__
+#include <assert.h>
 #include <stddef.h>
 #endif
 
@@ -66,19 +67,31 @@
 }))
 
 /*
+ * ROUNDUP2_DIV(x, y)
+ *
  * Rounds up to the nearest multiple of y and then divides by y. Safe
  * against overflow, y has to be a power of 2.
  *
  * This macro is intended to be used to convert from "number of bytes" to
  * "number of pages" or similar units. Example:
- * num_pages = ROUNDUP_DIV(num_bytes, SMALL_PAGE_SIZE);
+ * num_pages = ROUNDUP2_DIV(num_bytes, SMALL_PAGE_SIZE);
  */
-#define ROUNDUP_DIV(x, y) (__extension__({ \
-	typeof(x) __roundup_x = (x); \
-	typeof(y) __roundup_mask = (typeof(x))(y) - 1; \
-	\
-	(__roundup_x / (y)) + (__roundup_x & __roundup_mask ? 1 : 0); \
-}))
+#define ROUNDUP2_DIV(x, y) \
+	(__extension__({ \
+		typeof(x) __roundup_x = (x); \
+		typeof(y) __roundup_mask = (typeof(x))(y) - 1; \
+		\
+		assert(IS_POWER_OF_TWO(y)); \
+		(__roundup_x / (y)) + (__roundup_x & __roundup_mask ? 1 : 0); \
+	}))
+
+/*
+ * ROUNDUP_DIV(x, y)
+ *
+ * Rounds up to the nearest multiple of y and then divides by y. Safe
+ * against overflow.
+ */
+#define ROUNDUP_DIV(x, y) (ROUNDUP((x), (y)) / (__typeof__(x))(y))
 
 /* Round down the even multiple of size, size has to be a power of 2 */
 #define ROUNDDOWN(v, size) ((v) & ~((__typeof__(v))(size) - 1))
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -55,16 +55,49 @@
 			  ~((__typeof__(v))(size) - 1))
 
 /*
- * Round up the even multiple of size and return if result overflow
- * output value range. Size has to be a power of 2.
+ * ROUNDUP_OVERFLOW(v, size, res)
+ *
+ * @v: Input value to round
+ * @size: Rounding operand
+ * @res: Pointer where boolean overflow status (0/false or 1/true) is stored
+ * @return: boolean overflow status of the resulting rounded value
+ *
+ * Round up value @v to the even multiple of @size and return if result
+ * overflows the output value range pointed by @res. The rounded value is
+ * stored in the memory address pointed by @res.
  */
-#define ROUNDUP_OVERFLOW(v, size, res) (__extension__({ \
-	typeof(*(res)) __roundup_tmp = 0; \
-	typeof(v) __roundup_mask = (typeof(v))(size) - 1; \
-	\
-	ADD_OVERFLOW((v), __roundup_mask, &__roundup_tmp) ? 1 : \
-		((void)(*(res) = __roundup_tmp & ~__roundup_mask), 0); \
-}))
+#define ROUNDUP_OVERFLOW(v, size, res) \
+	(__extension__({ \
+		typeof(v) __roundup_mod = 0; \
+		typeof(v) __roundup_add = 0; \
+		\
+		__roundup_mod = (v) % (typeof(v))(size); \
+		if (__roundup_mod) \
+			__roundup_add = (typeof(v))(size) - __roundup_mod; \
+		ADD_OVERFLOW((v), __roundup_add, (res)); \
+	}))
+
+/*
+ * ROUNDUP2_OVERFLOW(v, size, res)
+ *
+ * @v: Input value to round
+ * @size: Rounding operand, must be a power of 2
+ * @res: Pointer where boolean overflow status (0/false or 1/true) is stored
+ * @return: boolean overflow status of the resulting rounded value
+ *
+ * Round up value @v to the even multiple of @size and return if result
+ * overflows the output value range pointed by @res. The rounded value is
+ * stored in the memory address pointed by @res.
+ */
+#define ROUNDUP2_OVERFLOW(v, size, res) \
+	(__extension__({ \
+		typeof(*(res)) __roundup_tmp = 0; \
+		typeof(v) __roundup_mask = (typeof(v))(size) - 1; \
+		\
+		assert(IS_POWER_OF_TWO(size)); \
+		ADD_OVERFLOW((v), __roundup_mask, &__roundup_tmp) ? 1 : \
+			((void)(*(res) = __roundup_tmp & ~__roundup_mask), 0); \
+	}))
 
 /*
  * ROUNDUP2_DIV(x, y)
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -50,9 +50,18 @@
 #endif
 
 #ifndef __ASSEMBLER__
+/* Round up the even multiple of size */
+#define ROUNDUP(x, y) \
+	((((x) + (__typeof__(x))(y) - 1) / (__typeof__(x))(y)) * \
+	 (__typeof__(x))(y))
+
 /* Round up the even multiple of size, size has to be a power of 2 */
-#define ROUNDUP(v, size) (((v) + ((__typeof__(v))(size) - 1)) & \
-			  ~((__typeof__(v))(size) - 1))
+#define ROUNDUP2(v, size) \
+	(__extension__({ \
+		assert(IS_POWER_OF_TWO(size)); \
+		(((v) + ((__typeof__(v))(size) - 1)) & \
+		 ~((__typeof__(v))(size) - 1)); \
+	}))
 
 /*
  * ROUNDUP_OVERFLOW(v, size, res)
--- a/lib/libutils/ext/include/util.h
+++ b/lib/libutils/ext/include/util.h
@@ -49,10 +49,14 @@
 #endif
 
 #ifndef __ASSEMBLER__
-/* Round up the even multiple of size, size has to be a multiple of 2 */
+/* Round up the even multiple of size, size has to be a power of 2 */
 #define ROUNDUP(v, size) (((v) + ((__typeof__(v))(size) - 1)) & \
 			  ~((__typeof__(v))(size) - 1))
 
+/*
+ * Round up the even multiple of size and return if result overflow
+ * output value range. Size has to be a power of 2.
+ */
 #define ROUNDUP_OVERFLOW(v, size, res) (__extension__({ \
 	typeof(*(res)) __roundup_tmp = 0; \
 	typeof(v) __roundup_mask = (typeof(v))(size) - 1; \
@@ -63,7 +67,7 @@
 
 /*
  * Rounds up to the nearest multiple of y and then divides by y. Safe
- * against overflow, y has to be a multiple of 2.
+ * against overflow, y has to be a power of 2.
  *
  * This macro is intended to be used to convert from "number of bytes" to
  * "number of pages" or similar units. Example:
@@ -76,7 +80,7 @@
 	(__roundup_x / (y)) + (__roundup_x & __roundup_mask ? 1 : 0); \
 }))
 
-/* Round down the even multiple of size, size has to be a multiple of 2 */
+/* Round down the even multiple of size, size has to be a power of 2 */
 #define ROUNDDOWN(v, size) ((v) & ~((__typeof__(v))(size) - 1))
 
 /*
--- a/core/drivers/clk/clk-stm32-core.c
+++ b/core/drivers/clk/clk-stm32-core.c
@@ -329,7 +329,7 @@ static unsigned long stm32_div_get_rate(int div_id, unsigned long prate)
 	if (!div)
 		return prate;
 
-	return ROUNDUP_DIV((uint64_t)prate, div);
+	return ROUNDUP2_DIV((uint64_t)prate, div);
 }
 
 TEE_Result stm32_div_set_rate(int div_id, unsigned long rate,
--- a/core/drivers/crypto/caam/utils/utils_dmaobj.c
+++ b/core/drivers/crypto/caam/utils/utils_dmaobj.c
@@ -386,7 +386,7 @@ static TEE_Result check_buffer_alignment(struct priv_dmaobj *priv,
 		}
 
 		va_start = (vaddr_t)entry->origbuf.data;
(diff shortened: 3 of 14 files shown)


lib/libutils/ext/include/util.h
```diff
@@ -50,9 +50,12 @@
 #endif
 
 #ifndef __ASSEMBLER__
-/* Round up the even multiple of size, size has to be a power of 2 */
-#define ROUNDUP(v, size) (((v) + ((__typeof__(v))(size) - 1)) & \
-			  ~((__typeof__(v))(size) - 1))
+/* Round up the even multiple of size */
+#define ROUNDUP(v, size) \
+	(IS_POWER_OF_TWO(size) ? \
```
jenswi-linaro, line 55: How about using two helper macros, one for the power of 2 case and one for the other?
While you're at it, it would be nice to evaluate the arguments only once, like in `UDIV_ROUND_NEAREST()` below.
Same below for `ROUNDDOWN()`.
Would it be possible to fix `ROUNDUP_OVERFLOW()` to allow a size that isn't a power of 2? It would align better with the new versions of `ROUNDUP()` and `ROUNDDOWN()`

etienne-lms: > How about using two helper macros, one for the power of 2 case and one for the other?

Okay, i'll go for `ROUNDUP_POW2()` and `ROUNDUP()` if that sounds ok to you (@jforissier ?)

That said, how to ensure one does not use a non power of 2 on `ROUNDUP_POW2()` for example. We cannot use `assert()` because the macro is used in `uint8_t buf[ROUNDUP(foo, bar)]` like array definitions?
One way: use non-pow2 `ROUNDUP()` macro in these cases (we expect the result is evaluated at build time).
Another way: have `ROUNDUP_POW2()` and `ROUNDDOWN_POW2()` to get the bit shift value instead of the power-of-2 size value.

> While you're at it, it would be nice to evaluate the arguments only once, like in UDIV_ROUND_NEAREST() below.
Same below for ROUNDDOWN().

Ok, i'll see that.

> Would it be possible to fix ROUNDUP_OVERFLOW() to allow a size that isn't a power of 2? It would align better with the new versions of ROUNDUP() and ROUNDDOWN()

I agree. I'll do that.

etienne-lms: Ok, i'll update this P-R

etienne-lms:
I've force pushed a new series (instead of appendix fixup commits).

I've preserved the PTA tests on `ROUNDUP()`/`ROUNDDOWN()`. Tests are quite basic and cover only there 2 macros. Maybe useless.

Last commit *"[RFC] tree-wide: use ROUND{UP|DOWN}_FAST() where applicable"* proposes few places where `ROUNDxxx_FAST()` could be used but I'm not sure it has a noticeable performance effect. I've tagged it *"[RFC]"*, please let me know if any of the changes are worth it, maybe only the MMU mapping related ones?

lib/libutils/ext/include/util.h
```diff
@@ -70,15 +70,42 @@
...
+		typeof(*(res)) __roundup_tmp = 0; \
+		typeof(v) __roundup_mask = (typeof(v))(size) - 1; \
+		\
+		ADD_OVERFLOW((v), __roundup_mask, &__roundup_tmp) ? 1 : \
+			((void)(*(res) = __roundup_tmp & ~__roundup_mask), 0); \
+	}))
+
+#define ROUNDUP_OVERFLOW_FAST(v, size, res) \
+	(__extension__({ \
+		assert(IS_POWER_OF_TWO(size); \
```
etienne-lms, line 86: My mistake: closing parenthesis missing.

lib/libutils/ext/include/util.h
```diff
@@ -70,15 +70,42 @@
...
+	(__extension__({ \
+		typeof(*(res)) __roundup_tmp = 0; \
+		typeof(v) __roundup_mod = 0; \
+		typeof(v) __roundup_add = 0; \
+		\
+		__roundup_mod = (v) % (typeof(v))(size); \
+		if (__roundup_mod) \
+			__roundup_add = (typeof(v))(size) - __roundup_mod; \
+		ADD_OVERFLOW((v), __roundup_add, &__roundup_tmp) ? 1 : \
+			((void)(*(res) = __roundup_tmp), 0); \
```
etienne-lms, line 103: This implementation is suboptimal. I've tested it successfully through xtest, forcing `ROUNDUP_OVERFLOW()` to use this "slow" path.

jenswi-linaro:
> Last commit _"[RFC] tree-wide: use ROUND{UP|DOWN}_FAST() where applicable"_ proposes few places where `ROUNDxxx_FAST()` could be used but I'm not sure it has a noticeable performance effect. I've tagged it _"[RFC]"_, please let me know if any of the changes are worth it, maybe only the MMU mapping related ones?

I've changed my view on `ROUNDUP()` a bit.

Before this PR, all the calls to `ROUNDUP()` must use a size that is a power of 2, so usage and implementation are optimal. We've always had this constraint of the size parameter to `ROUNDUP()`; changing it would be confusing. We shouldn't add anything to `ROUNDUP()` that might make it slower, or we'd be tempted to add a `ROUNDUP_FAST()` that then would need to be used instead of `ROUNDUP()` in most cases.

It would be nice to evaluate the `ROUNDUP()` arguments only once. We must be careful when adding an assert for the size parameter so the macro doesn't become less useful. Perhaps `__builtin_choose_expr()` together with `__builtin_constant_p()` can be used to select between a regular assert and a static_assert.

Calls that don't use a size of power of 2 should use a special macro since that's the exception. I can't come up with a good name at the moment, both `ROUNDUP_SAFE()` and `ROUNDUP_SLOW()` miss the point. We don't need this macro at the moment so we can wait with it until at least until we have a good name for it.

etienne-lms:
> It would be nice to evaluate the ROUNDUP() arguments only once.

Sorry, I missed that.

> (...) Perhaps __builtin_choose_expr() together with __builtin_constant_p() can be used to select between a regular assert and a static_assert.

`ROUNDUP()` is used in places where we the result need to be evaluated straight: `uint8_t foo[ROUNDUP(SIZE, ALIGN)]`. There is no place for `assert()` or `static_assert()` in such cases.
Yet, I think we need a way a ensure `ROUNDUP()` is not misused, passing a non-power-of-2 size argument.

> Calls that don't use a size of power of 2 should use a special macro since that's the exception. I can't come up with a good name at the moment, both ROUNDUP_SAFE() and ROUNDUP_SLOW() miss the point. We don't need this macro at the moment so we can wait with it until at least until we have a good name for it.

`ROUNDUP2()` could be good... unless the `2` suffix could look strange for something that accepts non-power-of-**2** size argument.

jenswi-linaro:
How about
``` C
#define ROUNDUP_CONST(v, sz) \
        (DIV_ROUND_UP(__builtin_choose_expr(__builtin_constant_p(v), \
                                            v, (void)0), \
                      __builtin_choose_expr(__builtin_constant_p(sz) && \
                                            __builtin_popcountl(sz) == 1, \
                                            sz, (void)0)) * (sz))


uint8_t foo[ROUNDUP_CONST(SIZE, ALIGN)];
```
We may need some wrapping for the `__builtin_*()` functions.

etienne-lms:
Hi @jenswi-linaro, sorry to bother you but I fail to use these builtin helpers. They don't behave as I would expect they do.
I've test this small code:
```c
/* Trigger a build error if constant value is not a power of 2. */
#define CONST_IS_POWER_OF_2(v) \
        (__builtin_choose_expr(__builtin_popcountl(v) == 1, 1, (void)0))

/* Return whether variable value is power of 2, enforce constant value is. */
#define IS_POWER_OF_2(v) \
        (__builtin_choose_expr(__builtin_constant_p(v), \
			       CONST_IS_POWER_OF_2(v), \
			       (((v) != 0) && (((v) & (~(v) + 1)) == (v)))))

int main(int argc, char **argv)
{
	if (IS_POWER_OF_2(16) ||
	    IS_POWER_OF_2(argc) ||
	    IS_POWER_OF_2(3))
		return 0;

	return 1;
}
```

I would expect build produce an error at line 15 (`IS_POWER_OF_2(3)`), but it also generates an error at line 14 because `argc` is not a constant:
```
$ gcc a.c
a.c: In function ‘main’:
a.c:3:10: error: first argument to ‘__builtin_choose_expr’ not a constant
    3 |         (__builtin_choose_expr(__builtin_popcountl(v) == 1, 1, (void)0))
      |          ^~~~~~~~~~~~~~~~~~~~~
a.c:8:11: note: in expansion of macro ‘CONST_IS_POWER_OF_2’
    8 |           CONST_IS_POWER_OF_2(v), \
      |           ^~~~~~~~~~~~~~~~~~~
a.c:14:6: note: in expansion of macro ‘IS_POWER_OF_2’
   14 |      IS_POWER_OF_2(argc) ||
      |      ^~~~~~~~~~~~~
a.c:3:64: error: void value not ignored as it ought to be
    3 |         (__builtin_choose_expr(__builtin_popcountl(v) == 1, 1, (void)0))
      |                                                                ^
a.c:8:11: note: in expansion of macro ‘CONST_IS_POWER_OF_2’
    8 |           CONST_IS_POWER_OF_2(v), \
      |           ^~~~~~~~~~~~~~~~~~~
a.c:15:6: note: in expansion of macro ‘IS_POWER_OF_2’
   15 |      IS_POWER_OF_2(3))
      |      ^~~~~~~~~~~~~
```
It looks like the compile evaluates `CONST_IS_POWER_OF_2()` for `argc` despite à`IS_POWER_OF_2()` should not have considered this expression at line 14.

In the scope of this P-R, it means I failed to define a unique macro `ROUNDUP()` to address cases where the size is a constant value (to trigger build error on non-power-of-2 cases) and cases where the size is a variable value (to assert it's a power of 2 at runtime). The only way I found is the have 2 macros, one for constant values, and one to use when the value is variable.

Any idea?

jenswi-linaro:
My idea with `ROUNDUP_CONST()` was that it should be used when a constant result is needed, like when defining the thread stacks. A constant result requires constant arguments, I added the extra tests for that. I think that `__builtin_popcountl()` might only work for constants, but I think that our normal `IS_POWER_OF_TWO()` macro should give a constant result with a constant input.

`__builtin_choose_expr()` is tricky, it seems it does a bit more than one might expect with the unused expression even if it will not cause any side effects. I think that the only differences compared to the `? :` operator is that the type of the two different expressions doesn't affect each other and that they can be used as lvalues.

It might be possible to work around it with macro trickery a la `IS_ENABLED()`.

etienne-lms:
> I think that __builtin_popcountl() might only work for constants, but I think that our normal IS_POWER_OF_TWO() macro should give a constant result with a constant input.

Agree. `IS_POWER_OF_TWO()` does the job.

> It might be possible to work around it with macro trickery a la IS_ENABLED().

The issue is to have the compiler not evaluating the part that don't apply (e.g. don't evaluate the `(void)0` case when not needed in your example in https://github.com/OP-TEE/optee_os/pull/7183#issuecomment-2548615114). With a `IS_ENABLED()` like macro, compiler still evaluates to 2 possible paths (for syntax conformance) even if it optimizes the code to embedded only the useful part.

My proposal: preserve currently massive use of `ROUNDUP()` where size arg is a constant (and instrument the macro to ensure at build time that the size is a power of 2) and add a `ROUNDUP_VAR()` for the few cases () where the size is not a constant value know form the compiler. See [this commit](https://github.com/etienne-lms/optee_os/pull/32/commits/903297da6052a8784f94dd91a464656f13956a0a) in https://github.com/etienne-lms/optee_os/pull/32 (volatile reference).

etienne-lms:
Sorry, I can't find my way out of this, considering case known constant, variables and that `IS_ENABLED()` `__ftmn_step_count()` are based on build time tricks on macro values aggregated with `##`.

Coming back to the initial goal of this P-R, I see that building core using either the "slow" variants of `ROUNDUP/DOWN()` (add, sub, div and mult) or with the fast variants (bit mask based) ends in the _**very same instructions sequence**_ for instructions like `val = ROUNDUP(data, CONSTANT_VALUE)` (thanks to the dear compiler smart optims, e.g. see below), or of course for instructions like `int foo[ROUNDUP(...)];`. Tested on 32bit and 64bit Arm archs. I guess RiscV will behaves the same way, but I've not checked.

Of course, when macros are used with a variable size argument,the resulting instructions differ a bit:
On Qemu armv7: core image is increased by 24 byte.
On Qemu armv8: core image is increased by 120 bytes.

Therefore I think the simplest way would be to use the "slow" variants for `ROUNDUP/DOWN()` (that are safer regarding the size argument).

For the few functions when really want to optimize rounding with a variable size argument (I assume mult/div instructions can take few more CPU cycles than add/mask instructions), we can add new `ROUNDUP_POW2()`/`RONDDOWN_POW2()` macros, based on bitmask. These could assert (in debug build mode) that the size is a power of 2. Only 11 source files (30 LOC) at most would need to be changed, most of them not being very performance sensitive.

The below C-source produce the very same instructions with the 2 variants of `ROUNDUP()`/`ROUNDDOWN()`:
```c
134: static void rem_um_region(struct user_mode_ctx *uctx, struct vm_region *r)
135: {
136: 	vaddr_t begin = ROUNDDOWN(r->va, CORE_MMU_PGDIR_SIZE);
137: 	vaddr_t last = ROUNDUP(r->va + r->size, CORE_MMU_PGDIR_SIZE);
138: 	struct vm_region *r2 = NULL;
139: 
	...
------------
000000000e11dcd0 <rem_um_region>:
rem_um_region():
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:135
 e11dcd0:	a9bd7bfd 	stp	x29, x30, [sp, #-48]!
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:137
 e11dcd4:	b24053e3 	mov	x3, #0x1fffff              	// #2097151
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:135
 e11dcd8:	910003fd 	mov	x29, sp
 e11dcdc:	a90153f3 	stp	x19, x20, [sp, #16]
 e11dce0:	aa0103f3 	mov	x19, x1
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:136
 e11dce4:	f9400821 	ldr	x1, [x1, #16]
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:135
 e11dce8:	a9025bf5 	stp	x21, x22, [sp, #32]
 e11dcec:	aa0003f6 	mov	x22, x0
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:137
 e11dcf0:	f9400e62 	ldr	x2, [x19, #24]
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:136
 e11dcf4:	926ba835 	and	x21, x1, #0xffffffffffe00000
.../optee-4.4.0-qemuv8/optee_os/core/mm/vm.c:137
 e11dcf8:	8b020022 	add	x2, x1, x2
 e11dcfc:	8b030054 	add	x20, x2, x3
```

@jenswi-linaro, I know you said you would preferred to preserve the power-of-2 constraint on `ROUNDUP-()`/`ROUNDDOWN()`. Is there a reason I missed?

Regarding ROUNDUP_OVERFLOW(), we can keep the existing constraint and add an assertion on the size being a power of 2.

jenswi-linaro:
I'm OK with your approach in general, with some small changes to make them easier to use.

I've checked what other projects (Linux, OpenBSD, NetBSD, Zephyr) do about this and it's more or less the same as we have upstream. I've seen the optimization we have with a bitmask. NetBSD has an interesting optimization in their `roundup2()` macro. https://github.com/NetBSD/src/blob/8059f17e777dbe181638d60ac854f737858871f4/sys/sys/param.h#L451

I propose adding new macros with the power-of-two constraint. Adding `ROUNDUP2()` and `ROUNDDOWN2()` should be the easiest. For consistency, `ROUNDUP_OVERFLOW()` needs a `ROUNDUP2_OVERFLOW()` counterpart.

etienne-lms:
OK, sounds good to me, I'll update this P-R accordingly. Thanks for the feedback.

etienne-lms:
To summarize and be sure I'm synced with you:
1- Add `ROUNDUP2()`, `ROUNDDOWN2()`, `ROUNDUP2_OVERFLOW()` with power-of-2 constraint (asserted).
2- Relax `ROUNDUP()`, `ROUNDDOWN()`, `ROUNDUP_OVERFLOW()` from the power-of-2 constraint.
3- Keep the power-of-2 constraint in `ROUNDUP_DIV()` and add an assertion for when rounding operand is not a constant
  (maybe and a `ROUNDUP2_DIV()` alias for consistency)
4- Change the callers using `ROUND{UP|UP_OVERFLOW|DOWN())` with a variable rounding operand to use the `ROUNDxxx2()` variants.

lib/libutils/ext/include/util.h
```diff
@@ -55,16 +55,29 @@
 			  ~((__typeof__(v))(size) - 1))
 
 /*
- * Round up the even multiple of size and return if result overflow
- * output value range. Size has to be a power of 2.
+ * ROUNDUP_OVERFLOW(v, size, res)
+ *
+ * @v: Input value to round
+ * @size: Rouding operand
```
jenswi-linaro, line 70: s/Rouding/Rounding/

lib/libutils/ext/include/util.h
```diff
@@ -66,19 +67,27 @@
...
-	typeof(y) __roundup_mask = (typeof(x))(y) - 1; \
-	\
-	(__roundup_x / (y)) + (__roundup_x & __roundup_mask ? 1 : 0); \
-}))
+#define ROUNDUP_DIV(x, y) \
+	(__extension__({ \
+		typeof(x) __roundup_x = (x); \
+		typeof(y) __roundup_mask = (typeof(x))(y) - 1; \
+		\
+		assert(IS_POWER_OF_TWO(y)); \
```
jenswi-linaro, line 130: This is inconsistent with the other `ROUNDUP*()` macros, those without a `2` should work with any alignment or granule.

lib/libutils/ext/include/util.h
```diff
@@ -66,19 +67,27 @@
...
+ * ROUNDUP_DIV(x, y)
+ * ROUNDUP2_DIV(x, y)
+ *
  * Rounds up to the nearest multiple of y and then divides by y. Safe
  * against overflow, y has to be a power of 2.
  *
  * This macro is intended to be used to convert from "number of bytes" to
  * "number of pages" or similar units. Example:
  * num_pages = ROUNDUP_DIV(num_bytes, SMALL_PAGE_SIZE);
+ *
```
jenswi-linaro, line 123: Please remove this line.

jenswi-linaro (commented):
Thanks for your patience and effort. This looks good. I only have a few comments.

etienne-lms:
I did not port the NetBSD roundup implementation because of license terms of our *util.h* file vs *param.h*. I think it's doable but rather with a dedicated *util2.h* header, maybe included from *util.h*.

lib/libutils/ext/include/util.h
etienne-lms: Agree. I'll make a dedicated relaxed `ROUNDUP_DIV()` macro.

lib/libutils/ext/include/util.h
etienne-lms: oops, sorry.

etienne-lms (commented):
Thanks for your patience too :-) and your (always precious) help.

jenswi-linaro:
> I did not port the NetBSD roundup implementation because of license terms of our _util.h_ file vs _param.h_. I think it's doable but rather with a dedicated _util2.h_ header, maybe included from _util.h_.

No worries, we can do that in another PR.

etienne-lms:
I've tested 32b and 64bit with the NetBSD macro: it tends to add 1 instruction but not always, depending on the rounding operand and the surrounded instructions. In the end, heavy using the optimized `ROUNDUP()` macro in OP-TEE core makes the image 80 instruction bytes (resp. 136 instruction bytes) bigger for Qemu armv7a (resp. for Qemu armv8a) when using the NetBSD macro implementation. It seems not worth it.

etienne-lms:
Comments addressed.

jenswi-linaro:
Please squash in the updates.

etienne-lms:
I've squashed the fixup commits, fixed few remaining typos.
I've also updated commit "tree-wide: use power-of-2 rounding macros where applicable" to consider also `ROUDUP2_DIV()` and `ROUNDUP2_OVERFLOW()` (very few occurrences).

lib/libutils/ext/include/util.h
```diff
@@ -55,16 +55,51 @@
...
+	(__extension__({ \
+		typeof(*(res)) __roundup_tmp = 0; \
+		typeof(v) __roundup_mod = 0; \
+		typeof(v) __roundup_add = 0; \
+		\
+		__roundup_mod = (v) % (typeof(v))(size); \
+		if (__roundup_mod) \
+			__roundup_add = (typeof(v))(size) - __roundup_mod; \
+		ADD_OVERFLOW((v), __roundup_add, &__roundup_tmp) ? 1 : \
+			((void)(*(res) = __roundup_tmp), 0); \
```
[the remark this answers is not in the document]
jenswi-linaro: Agreed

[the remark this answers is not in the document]
etienne-lms: Addressed with:
```patch
@@ -68,15 +68,13 @@
  */
 #define ROUNDUP_OVERFLOW(v, size, res) \
        (__extension__({ \
-               typeof(*(res)) __roundup_tmp = 0; \
                typeof(v) __roundup_mod = 0; \
                typeof(v) __roundup_add = 0; \
                \
                __roundup_mod = (v) % (typeof(v))(size); \
                if (__roundup_mod) \
                        __roundup_add = (typeof(v))(size) - __roundup_mod; \
-               ADD_OVERFLOW((v), __roundup_add, &__roundup_tmp) ? 1 : \
-                       ((void)(*(res) = __roundup_tmp), 0); \
+               ADD_OVERFLOW((v), __roundup_add, (res)); \
        }))
 
 /*
```

etienne-lms:
Commit message fixed and `ROUNDUP_OVERFLOW()` optimized according to https://github.com/OP-TEE/optee_os/pull/7183#discussion_r1892724608.
I've applied your review tag. Thanks again.
CI tests should be all successful.

jforissier:
>  CI tests should be all successful.

It seems we haven't fixed all the issues (QEMUv8 with pager, regression_1006 timeout). Last time I checked, on my local build machine that test took ~5 minutes. The timeout in qemu-check.exp is 900 seconds (15 minutes), perhaps it's not enough on the CI machines. I will experiment a bit. In the meantime this PR can be merged.

etienne-lms:
Thanks @jforissier. The qemu_armv8a test I ran on my side on this P-R were ok. bur I did not test all the 4 configs 'CI check QEMUv8 1 / 2' runs.

PR #7185, opened 2024-12-16
dt-bindings: stm32: fix CLKSRC for RTC in stm32mp13 clock bindings

VCASTM:
Bad copy/paste, use MUX ID to configure the clock source of RTC and not the clock ID.

Fixes: 19a4632e0f179 ("dt-bindings: stm32: add stm32mp13 clock and reset bindings")


VCASTM:
This patch is already in another PR:
https://github.com/OP-TEE/optee_os/pull/7184

PR #7188, opened 2024-12-17
core: arm: add guest id check at VM creation

yuvraj1803:
These 3 commits fix OPTEE going into panic under certain SMCs from NW when CONFIG_NS_VIRTUALIZATION is enabled.

### DIFF
--- a/core/arch/arm/tee/entry_fast.c
+++ b/core/arch/arm/tee/entry_fast.c
@@ -296,11 +296,11 @@ void __tee_entry_fast(struct thread_smc_args *args)
 
 	case OPTEE_SMC_ENABLE_ASYNC_NOTIF:
 		if (IS_ENABLED(CFG_CORE_ASYNC_NOTIF)) {
-#if defined(CFG_NS_VIRTUALIZATION)
-			notif_deliver_atomic_event(NOTIF_EVENT_STARTED, args->a7);
-#else
-			notif_deliver_atomic_event(NOTIF_EVENT_STARTED, 0);
-#endif
+			if(IS_ENABLED(CFG_NS_VIRTUALIZATION)){
+				notif_deliver_atomic_event(NOTIF_EVENT_STARTED, args->a7);
+			}else{
+				notif_deliver_atomic_event(NOTIF_EVENT_STARTED, 0);
+			}
 			args->a0 = OPTEE_SMC_RETURN_OK;
 		} else {
 			args->a0 = OPTEE_SMC_RETURN_UNKNOWN_FUNCTION;
--- a/core/arch/arm/tee/entry_fast.c
+++ b/core/arch/arm/tee/entry_fast.c
@@ -171,12 +171,7 @@ static void tee_entry_get_thread_count(struct thread_smc_args *args)
 static void tee_entry_vm_created(struct thread_smc_args *args)
 {
 	uint16_t guest_id = args->a1;
-
-	if(guest_id == HYP_CLNT_ID){
-		args->a0 = OPTEE_SMC_RETURN_ENOTAVAIL;
-		return;
-	}
-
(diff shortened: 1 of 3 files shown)


core/kernel/notif_default.c
```diff
@@ -87,7 +87,7 @@ void notif_send_async(uint32_t value, uint16_t guest_id __maybe_unused)
 	uint32_t old_itr_status = 0;
 	struct itr_chip *itr_chip = interrupt_get_main_chip();
 
-	assert(value <= NOTIF_ASYNC_VALUE_MAX && !guest_id);
+	assert(value <= NOTIF_ASYNC_VALUE_MAX);
```
jenswi-linaro, line 90: This will not work as intended for `guest_id != 0`, the `notif_values` bitstring must be per guest.
For the FF-A configuration, we have a separate implementation (in `core/arch/arm/kernel/thread_spmc.c`) due to how it must interact with the FF-A framework.

core/arch/arm/tee/entry_fast.c
```diff
@@ -296,7 +296,11 @@ void __tee_entry_fast(struct thread_smc_args *args)
 
 	case OPTEE_SMC_ENABLE_ASYNC_NOTIF:
 		if (IS_ENABLED(CFG_CORE_ASYNC_NOTIF)) {
+#if defined(CFG_NS_VIRTUALIZATION)
```
jenswi-linaro, line 299: prefer `if (IS_ENABLED(CFG_NS_VIRTUALIZATION))`

core/arch/arm/tee/entry_fast.c
```diff
@@ -172,6 +172,11 @@ static void tee_entry_vm_created(struct thread_smc_args *args)
 {
 	uint16_t guest_id = args->a1;
 
+	if(guest_id == HYP_CLNT_ID){
```
jenswi-linaro, line 175: This is better to check in `virt_guest_created()`.

yuvraj1803: Noted.

core/arch/arm/tee/entry_fast.c
yuvraj1803: Noted.

core/kernel/notif_default.c
yuvraj1803: Is there a way around this? Because, test_wd_callback() does pass guest_id>0 here.

jenswi-linaro: I don't think Xen supports notifications via the SMC ABI. That's why you're running into this problem.
Xen does support notifications via the FF-A ABI, but I assume you're using the SMC ABI for a reason.
I can try to write some code to support this with the SMC ABI, but I don't have a way of testing it so I need your help.

yuvraj1803: Sure

yuvraj1803: @jenswi-linaro please check the new PR, where I have made the suggested changes for the other two commits.

Sorry for the duplicate, made some unresolvable changes on local tree.

yuvraj1803: > I don't think Xen supports notifications via the SMC ABI. That's why you're running into this problem. Xen does support notifications via the FF-A ABI, but I assume you're using the SMC ABI for a reason. I can try to write some code to support this with the SMC ABI, but I don't have a way of testing it so I need your help.

Is is possible to init notif_values (or any other entity personal to a guest) into the TEE part? So its replicated for each guest?

jenswi-linaro: Yes, I'll post a link in the new PR.

yuvraj1803: Sure

PR #7190, opened 2024-12-18
ci: QEMUv8: disable fTPM with pager

jforissier:
Disable fTPM when testing with pager enabled, because execution is too slow otherwise and sometimes fails with a timeout error. For example I noticed that the execution time of "xtest regression 1006" may be increased from 5 to 10 minutes.


etienne-lms (commented):

jenswi-linaro:

PR #7193, opened 2024-12-20
vexpress-qemu_armv8a: increase CFG_CORE_TZSRAM_EMUL_SIZE

jforissier:
Increase the emulated TrustZone SRAM size from 512 to 640 KB for the vexpress-qemu_armv8a. This drastically improves the execution speed when pager is enabled. For example, without this change the command "time xtest regression_1006" takes around 3 minutes on my build machine, but it takes only 9 seconds with the increase TZSRAM. Similarly, the same test on the GitHub CI runners needs 10 minutes before the change and only 15 seconds after.

This is expected to solve the occasional timeouts that we see occurring with the QEMUv8_check1 CI job.


jenswi-linaro:

jforissier:
> You may want to mention commit [46fdfee](https://github.com/OP-TEE/optee_os/commit/46fdfeea761f66af7f460f829faadfabb8aae687) ("vexpress-qemu_armv8a: increase CFG_CORE_HEAP_SIZE to 131072") since that effectively took away 64k from the pager. @jbech-linaro, this is probably the increased delay you've noticed.

Good catch. This is also related to b4ed37a8c754 ("plat-vexpress: increase QEMU heap size"). So the pager actually  lost 64K + 64K. The first commit caused the execution time of 1006 to rise from 9 to 14 seconds, and the second one increased it to 3 minutes ;) I mentioned both commits in the description.

PR #7198, opened 2024-12-29
Fix linking problems/warnings in ubsan

alexhoppus:
The following changes are done in ubsan subsystem:

1) `__ubsan_handle_invalid_builtin` callback was added. This ubsan handler
will be invoked when undefined behavior appears in `__builtin_*`
functions. That could happen if invalid arguments were provided.
E.g. passing 0 as the argument to `__builtin_ctz_` or `__builtin_clz_` invokes
undefined behavior and could be potentially caught by ubsan.
I was not able to build latest optee for **vexpress-qemu_armv8a** 
(CFG_CORE_SANITIZE_UNDEFINED=y)
without this change, because this callback was generated in
[libmbedtls code](https://github.com/OP-TEE/optee_os/blame/master/lib/libmbedtls/mbedtls/library/bignum_core.c#L36C23-L36C23)
2) It seems that signatures of ubsan functions are not compatible
with latest (or more or less fresh) gcc ubsan interfaces. That creates
warnings during compilation. These warnings were fixed.
3) It seems `__ubsan_handle_unreachable` should be `__ubsan_handle_builtin_unreachable`.
At least that corresponds to latest gcc and to what was there since version 4.9.0
4) Small refactoring to reduce code size

### DIFF
--- a/core/kernel/ubsan.c
+++ b/core/kernel/ubsan.c
@@ -63,6 +63,11 @@ struct nonnull_arg_data {
 	struct source_location loc;
 };
 
+struct invalid_builtin_data {
+	struct source_location loc;
+	unsigned char kind;
+};
+
 /*
  * When compiling with -fsanitize=undefined the compiler expects functions
  * with the following signatures. The functions are never called directly,
@@ -99,6 +104,7 @@ void __ubsan_handle_nonnull_arg(struct nonnull_arg_data *data
 				, size_t arg_no
 #endif
 			       );
+void __ubsan_handle_invalid_builtin(struct invalid_builtin_data *data);
 
 static void print_loc(const char *func, struct source_location *loc)
 {
@@ -235,6 +241,13 @@ void __ubsan_handle_nonnull_arg(struct nonnull_arg_data *data
 				, size_t arg_no __unused
 #endif
 			       )
+{
+	print_loc(__func__, &data->loc);
+	if (ubsan_panic)
+		panic();
+}
+
+void __ubsan_handle_invalid_builtin(struct invalid_builtin_data *data)
 {
 	print_loc(__func__, &data->loc);
 	if (ubsan_panic)
--- a/core/kernel/ubsan.c
+++ b/core/kernel/ubsan.c
@@ -75,36 +75,25 @@ struct invalid_builtin_data {
  */
 void __ubsan_handle_type_mismatch(struct type_mismatch_data *data,
 				  unsigned long ptr);
-void __ubsan_handle_type_mismatch_v1(struct type_mismatch_data *data,
-				     unsigned long ptr);
-void __ubsan_handle_add_overflow(struct overflow_data *data,
-				  unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_sub_overflow(struct overflow_data *data,
-				  unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_mul_overflow(struct overflow_data *data,
-				  unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_negate_overflow(struct overflow_data *data,
-				    unsigned long old_val);
-void __ubsan_handle_divrem_overflow(struct overflow_data *data,
-				    unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_pointer_overflow(struct overflow_data *data,
-				     unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_shift_out_of_bounds(struct shift_out_of_bounds_data *data,
-					unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_out_of_bounds(struct out_of_bounds_data *data,
-				  unsigned long idx);
+void __ubsan_handle_type_mismatch_v1(void *data_, void *ptr);
+void __ubsan_handle_add_overflow(void *data_, void *lhs, void *rhs);
+void __ubsan_handle_sub_overflow(void *data_, void *lhs, void *rhs);
+void __ubsan_handle_mul_overflow(void *data_, void *lhs, void *rhs);
+void __ubsan_handle_negate_overflow(void *data_, void *old_val);
+void __ubsan_handle_divrem_overflow(void *data_, void *lhs, void *rhs);
+void __ubsan_handle_pointer_overflow(void *data_, void *lhs, void *rhs);
+void __ubsan_handle_shift_out_of_bounds(void *data_, void *lhs, void *rhs);
+void __ubsan_handle_out_of_bounds(void *data_, void *idx);
 void __ubsan_handle_unreachable(struct unreachable_data *data);
-void __ubsan_handle_missing_return(struct unreachable_data *data);
-void __ubsan_handle_vla_bound_not_positive(struct vla_bound_data *data,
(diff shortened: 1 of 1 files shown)


core/kernel/ubsan.c
```diff
@@ -75,36 +75,25 @@ struct invalid_builtin_data {
...
-				    unsigned long old_val);
-void __ubsan_handle_divrem_overflow(struct overflow_data *data,
-				    unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_pointer_overflow(struct overflow_data *data,
-				     unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_shift_out_of_bounds(struct shift_out_of_bounds_data *data,
-					unsigned long lhs, unsigned long rhs);
-void __ubsan_handle_out_of_bounds(struct out_of_bounds_data *data,
-				  unsigned long idx);
+void __ubsan_handle_type_mismatch_v1(void *data_, void *ptr);
```
jenswi-linaro, line 78: How did you find the proper declaration of this and the other `__ubsan_handle_` functions?
Do you have a link?

core/kernel/ubsan.c
```diff
@@ -196,11 +196,11 @@ void __ubsan_handle_out_of_bounds(void *data_, void *idx __unused)
...
 }
 
-void __ubsan_handle_unreachable(struct unreachable_data *data)
+void __ubsan_handle_builtin_unreachable(void *data_)
 {
+	struct unreachable_data *data = data_;
 	print_loc(__func__, &data->loc);
-	if (ubsan_panic)
-		panic();
+	panic();
```
jenswi-linaro, line 185: Doesn't this result in a warning?

alexhoppus: Hi,

I think the opposite results in a warning, since this function marked with attribute noreturn. So if it doesn't call explicitly any other noreturn function or doesn't end with infinite loop I get the warning.

```
core/kernel/ubsan.c: In function ‘__ubsan_handle_builtin_unreachable’:
core/kernel/ubsan.c:188:1: warning: ‘noreturn’ function does return
  188 | }
```
That is the reason why I treated noreturn functions in a little bit special way here.

core/kernel/ubsan.c
alexhoppus: Hi,

There are at least 3 ways

1) gcc gives warning during compilation and shows correct prototype at the same time. So one can rely on this gcc warning messages to figure out correct function signature. The initial idea behind this was to fix warnings.
For example:
```
core/kernel/ubsan.c:78:6: warning: conflicting types for built-in function ‘__ubsan_handle_type_mismatch_v1’; expected ‘void(void *, void *)’ [-Wbuiltin-declaration-mismatch]
   78 | void __ubsan_handle_type_mismatch_v1(void *data_, unsigned long ptr);
```
2) one can refer to [linux kernel implementation](https://github.com/torvalds/linux/blob/master/lib/ubsan.c#L402)
3) [look](https://github.com/gcc-mirror/gcc/blob/master/gcc/sanitizer.def#L515) into gcc . I hope I gave a link to correct place

core/kernel/ubsan.c
jenswi-linaro: That means that this function has an implicit `noreturn` attribute. How about adding it explicitly to be clear?

alexhoppus: > That means that this function has an implicit noreturn attribute. How about adding it explicitly to be clear?

That was exactly my thought and I tried to do this. But for some weird reason I get a warning message in this case

```
core/kernel/ubsan.c:182:1: warning: ignoring attribute ‘noreturn’ in declaration of a built-in function ‘__ubsan_handle_builtin_unreachable’ because it conflicts with attribute ‘const’ [-Wattributes]
  182 | {
      | ^
```
After some research I found [this](https://patches.linaro.org/project/lkml/patch/20180202154813.1625742-1-arnd@arndb.de/).  So I can mark it as noreturn, but we will see this warning. If you are okay, then I will do that.

jenswi-linaro: Fair enough, it seems like a forever bug in GCC, let's keep it as it is.

core/kernel/ubsan.c
```diff
@@ -63,44 +63,42 @@ struct nonnull_arg_data {
...
 				, size_t arg_no
 #endif
 			       );
+void __ubsan_handle_invalid_builtin(void *data_);
 
-static void print_loc(const char *func, struct source_location *loc)
+static volatile bool ubsan_panic = true;
+
+static void ubsan_handle_error(const char *func, struct source_location *loc,
+		bool should_panic)
```
jenswi-linaro, line 101: Please align with the opening `(` on the line above.

alexhoppus: Done

jenswi-linaro (commented):

jforissier:
Regarding the CI Code style errors:
- I believe the "'externs" warnings are false positives, checkpatch seems to be confused. Let's ignore them.
- Please fix the "unwrapped commit description" warning: https://github.com/OP-TEE/optee_os/actions/runs/12632861093/job/35262346722?pr=7198#step:5:109
- See my comment about the "volatile"

alexhoppus:
Thanks for the reply.

> Regarding the CI Code style errors:
> 
> * I believe the "'externs" warnings are false positives, checkpatch seems to be confused. Let's ignore them.
> * Please fix the "unwrapped commit description" warning: https://github.com/OP-TEE/optee_os/actions/runs/12632861093/job/35262346722?pr=7198#step:5:109

Fixed in this place and all other places, should be ok now.

> * See my comment about the "volatile"

I'm not sure which comment do you mean. The volatile variable, checkpatch complained to, was not introduced by me, but anyway I removed it to make checkpatch happy. I squashed this change directly to last commit, hope that is ok. Thanks.

jforissier:
Regarding the CI Code style errors:
- I believe the "'externs" warnings are false positives, checkpatch seems to be confused. Let's ignore them.
- Please fix the "unwrapped commit description" warning: https://github.com/OP-TEE/optee_os/actions/runs/12632861093/job/35262346722?pr=7198#step:5:109
- See my comment about the "volatile"

> Thanks for the reply.
> 
> > Regarding the CI Code style errors:
> > 
> > * I believe the "'externs" warnings are false positives, checkpatch seems to be confused. Let's ignore them.
> > * Please fix the "unwrapped commit description" warning: https://github.com/OP-TEE/optee_os/actions/runs/12632861093/job/35262346722?pr=7198#step:5:109
> 
> Fixed in this place and all other places, should be ok now.
> 
> > * See my comment about the "volatile"
> 
> I'm not sure which comment do you mean. The volatile variable, checkpatch complained to, was not introduced by me, but anyway I removed volatile modifier to make checkpatch happy. Per my understanding removing this volatile modifier shouldn't affect anything. I squashed this change directly to last commit, hope that is ok. Thanks.

Sorry I forgot to submit the review comment... but that's what I meant. Thanks.

jenswi-linaro:
`ubsan_panic` was `volatile` to prevent the compiler from warning that a function didn't return. It looks like the extra indirection with a parameter to `ubsan_handle_error()` also does the trick.

PR #7203, opened 2025-01-06
ta: pkcs11: set TA version ID to 1.0.0

etienne-lms:
PKCS#11 TA version ID is 0.1.0 since its early integration. Let's set it 1.0.0 since it now mature enough to deserve a non-zero major version number.


jforissier:
Good idea. And then we should increment the minor/patch numbers as new features & fixes are added...

jenswi-linaro:

PR #7205, opened 2025-01-07
core: mm: add memory area type of transfer list

Akshay-Belsare:
Add memory area type of transfer list in switch case. During early boot in OP-TEE OS, the transfer list is mapped into secure memory by the transfer_list_map function. However, when mapping non-secure memory from the DTB memory node, OP-TEE panics if the transfer list is located within non-secure memory ranges.
To address this issue, the OP-TEE memory type for the transfer list has been added to the core_mmu_set_discovered_nsec_ddr function.

### DIFF
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -524,6 +524,7 @@ void core_mmu_set_discovered_nsec_ddr(struct core_mmu_phys_mem *start,
 		case MEM_AREA_SHM_VASPACE:
 		case MEM_AREA_TS_VASPACE:
 		case MEM_AREA_PAGER_VASPACE:
+		case MEM_AREA_TRANSFER_LIST:
 			break;
 		default:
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
The memory ranges passed to `core_mmu_set_discovered_nsec_ddr()` should not overlap with secure memory. It seems like waiting for a security problem to allow OP-TEE to treat some secure memory as non-secure memory.

Akshay-Belsare:
@raymo200915 could you please comment on this, Thanks!

raymo200915:
> @raymo200915 could you please comment on this, Thanks!

The TL is keeping stay in secure memory until right before existing EL3.
So in BL32 please directly add nodes into the TL instead of mapping a non-secure region to it.

Akshay-Belsare:
After syncing with @jenswi-linaro and @raymo200915, exploring another method to resolve this. Withdrawing this PR!

PR #7208, opened 2025-01-08
sign_encrypt.py: fix bugs in 'verify' and 'display' with encrypted TA

meeneemaru:
Fix the following bugs:
- The display command displays an incorrect tag value.
- The verify command requires --enc_key for encrypted TA, but an error occurs when --enc_key option is used.


jforissier (commented):
Hi @meeneemaru, thanks for the fixes!

Since this is really fixing two different bugs, I'd prefer to have two commits if you don't mind. With that:

meeneemaru:
@jforissier 
Thank you for your review.
I split the commit.

jforissier:
I think this can safely go into 4.5.0 so I am merging this PR.

PR #7210, opened 2025-01-08
(FIX)clk: clk-stm32-core: fix use of ROUNDUP2_DIV in stm32_div_get_rate()

GseoC:


etienne-lms (commented):
In commit message:
s/ROUNDUP2_DIV/ROUNDUP2_DIV()/
s/ROUNDUP_DIV/ROUNDUP_DIV()/
I would replace "In this case, the platform panics." with something like "This issue makes STM32MP13-DK platform to panic in debug mode. In non-debug mode, the computed clock rate can differ by few Hertz/.

etienne-lms:
I missed to say:
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with these comments addressed.

Gabriel-Fernandz:
Acked-by: Gabriel Fernandez <gabriel.fernandez@foss.st.com>

GseoC:
All tags applied, thanks

Issue #7217, opened 2025-01-13
Missing nullptr checks in TEE_BigIntComputeExtendedGcd?


snst:
Hello,

I think two nullptr checks are missing in function TEE_BigIntComputeExtendedGcd for u and v at below position.
According TEE Internal Core API Specification parameter u, v, or both can be NULL.

https://github.com/OP-TEE/optee_os/blob/96e8f740b7601089fcabb8806a06bbd704d33f3f/lib/libutee/tee_api_arith_mpi.c#L867C1-L868C44

`

		MPI_CHECK(copy_mpi_to_bigint(&mpi_u, u));
		MPI_CHECK(copy_mpi_to_bigint(&mpi_v, v));
`

Thanks
stefan

jenswi-linaro:
I agree, it's a bug.

etienne-lms:
The case is addressed few lines above:
https://github.com/OP-TEE/optee_os/blob/96e8f740b7601089fcabb8806a06bbd704d33f3f/lib/libutee/tee_api_arith_mpi.c#L838-L839

etienne-lms:
Hmmm: I may be wrong; the case where only one is NULL is not addressed. To be checked.

snst:
yes, I am referring to the case where only one pointer is null.

etienne-lms:
Fixing with something like the below?
```patch
  		mpi_u.s *= s1;
 		mpi_v.s *= s2;
 
-		MPI_CHECK(copy_mpi_to_bigint(&mpi_u, u));
-		MPI_CHECK(copy_mpi_to_bigint(&mpi_v, v));
+		if (u)
+			MPI_CHECK(copy_mpi_to_bigint(&mpi_u, u));
+		if (v)
+			MPI_CHECK(copy_mpi_to_bigint(&mpi_v, v));
 		mbedtls_mpi_free(&mpi_u);
 		mbedtls_mpi_free(&mpi_v);

snst:
For me this fix would be fine.

etienne-lms:
@snst, I've created https://github.com/OP-TEE/optee_os/pull/7218.

PR #7218, opened 2025-01-13
libtuee: allow NULL output coefficients in TEE_BigIntComputeExtendedGcd()

etienne-lms:
Fix TEE_BigIntComputeExtendedGcd() for when only one of u and v output coefficients reference if NULL as allowed by the GP TEE Internal Core API specification.

Reported-by: Stefan Schmidt <snst@meek.de>
Closes: https://github.com/OP-TEE/optee_os/issues/7217


jforissier:
s/if NULL/is NULL/ in the commit description.

etienne-lms:
Thanks.
Commit message fixed and review tag applied.

jforissier:
Taking this one too in the release since it is a bugfix. Thanks!

Issue #7215, opened 2025-01-10
QEMUv8: fTPM sanity test fails when CFG_REE_FS=n


jforissier:
Platform: QEMUv8.
Build command:
```
make -j$(nproc) CFG_REE_FS=n CFG_RPMB_FS=y CFG_RPMB_WRITE_KEY=y CFG_RPMB_ANNOUNCE_PROBE_CAP=n run
```
Test command:
```
buildroot login: test
$ xtest 1041
Test ID: 1041
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# regression+pkcs11
#
######################################################

* regression_1041 Test fTPM sanity
/tmp/optee/out-br/build/optee_test_ext-1.0/host/xtest/regression_1000.c:3384: Expression "stat(fname, &sb) == 0" (-1 == 0) is false
stat("/dev/tpm0"): No such file or directory
Perhaps fTPM hasn't finished probing
  regression_1041 FAILED
+-----------------------------------------------------
Result of testsuite regression+pkcs11 filtered by "1041":
regression_1041 FAILED first error at /tmp/optee/out-br/build/optee_test_ext-1.0/host/xtest/regression_1000.c:3384
+-----------------------------------------------------
2 subtests of which 1 failed
1 test case of which 1 failed
137 test cases were skipped
TEE test application done!
```

Configurations with REE FS enabled do work (with or without RPMB):
```
make -j$(nproc) CFG_REE_FS=y CFG_RPMB_FS=y CFG_RPMB_WRITE_KEY=y CFG_RPMB_ANNOUNCE_PROBE_CAP=n run
make -j$(nproc) CFG_REE_FS=y CFG_RPMB_FS=n run
```

jforissier:
One thing I noticed: when REE FS is enabled, the fTPM TA is invoked during boot (before the login prompt). It is not the case when REE FS is disabled.
```
$ grep bc50d971 /tmp/boot_* | grep -E "(Early|Lookup)"
/tmp/boot_REE_n_RPMB_y.txt:D/TC:0 0 early_ta_init:56 Early TA bc50d971-d4c9-42c4-82cb-343fb7f37896 size 176724 (compressed, uncompressed 380096)
/tmp/boot_REE_y_RPMB_n.txt:D/TC:0 0 early_ta_init:56 Early TA bc50d971-d4c9-42c4-82cb-343fb7f37896 size 176724 (compressed, uncompressed 380096)
[...]

PR #7220, opened 2025-01-15
core: pta: device: fix enumeration for PTA_CMD_GET_DEVICES_SUPP

jforissier:
TAs which depend on TEE_STORAGE_PRIVATE do need the TEE supplicant if REE FS is disabled (in which case secure storage can only be RPMB) and RPMB is not routed via the kernel.

Fixes: a96033ca7bee ("core: add flag to enumerate TAs when secure storage is ready")

### DIFF
--- a/core/pta/device.c
+++ b/core/pta/device.c
@@ -88,6 +88,11 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 {
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t rflags = 0;
+	/*
+	 * This should also be true if CFG_RPMB_ANNOUNCE_PROBE_CAP is
+	 * enabled when the kernel does not support OP-TEE RPMB operations.
+	 */
+	bool rpmb_needs_supp = !IS_ENABLED(CFG_RPMB_ANNOUNCE_PROBE_CAP);
 
 	switch (nCommandID) {
 	case PTA_CMD_GET_DEVICES:
@@ -95,7 +100,8 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 		break;
 	case PTA_CMD_GET_DEVICES_SUPP:
 		rflags = TA_FLAG_DEVICE_ENUM_SUPP;
-		if (IS_ENABLED(CFG_REE_FS))
+		if (IS_ENABLED(CFG_REE_FS) ||
+		    (IS_ENABLED(CFG_RPMB_FS) && rpmb_needs_supp))
 			rflags |= TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE;
 		break;
 	case PTA_CMD_GET_DEVICES_RPMB:


core/pta/device.c
```diff
@@ -88,14 +88,21 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 {
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t rflags = 0;
+	/*
+	 * FIXME: this should also be true if CFG_RPMB_ANNOUNCE_PROBE_CAP is
+	 * enabled but the kernel does not support OP-TEE RPMB operations
+	 * (probe failed)
```
etienne-lms, line 94: Should we assume `PTA_CMD_GET_DEVICES` is always called before `PTA_CMD_GET_DEVICES_SUPP` and that `case PTA_CMD_GET_DEVICES_RPMB` may be called in between (which means kernel supports it new RPMB subsystem): using so, we could force `rflags |= TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE` when either `PTA_CMD_GET_DEVICES_RPMB` was not called or it failed on `tee_rpmb_reinit()`?

core/pta/device.c
```diff
@@ -88,14 +88,21 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
...
 	switch (nCommandID) {
 	case PTA_CMD_GET_DEVICES:
 		rflags = TA_FLAG_DEVICE_ENUM;
 		break;
 	case PTA_CMD_GET_DEVICES_SUPP:
 		rflags = TA_FLAG_DEVICE_ENUM_SUPP;
-		if (IS_ENABLED(CFG_REE_FS))
+		if (IS_ENABLED(CFG_REE_FS) ||
+		    (IS_ENABLED(CFG_RPMB_FS) && rpmb_needs_supp))
 			rflags |= TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE;
```
etienne-lms, line 106: I wonder if `TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE` is sufficient.
If a TA uses the RPMB_FS and expects to be enumerated, it should set a `TA_FLAG_DEVICE_ENUM_TEE_STORAGE_RPMB`-like flag, regardless of whether the REE_FS is enabled or not.
That comment may be a bit outside the scope of this P-R.

core/pta/device.c
jenswi-linaro: I'm not sure we can assume that `PTA_CMD_GET_DEVICES_RPMB` always will be done before `PTA_CMD_GET_DEVICES_SUPP`, it's two independent actions in the kernel.

jforissier: I agree. And we also don't want to call `tee_rpmb_reinit()` each time. We need a dedicated function `tee_rpmb_needs_supplicant()` I think.

core/pta/device.c
```diff
@@ -88,14 +88,21 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 {
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t rflags = 0;
+	/*
+	 * FIXME: this should also be true if CFG_RPMB_ANNOUNCE_PROBE_CAP is
```
jenswi-linaro, line 92: I don't think this can be fixed, at least not without adding more exchanges between the kernel driver and OP-TEE. 
How big a problem is this?

etienne-lms: Could Linux set an `OPTEE_{SMC|FFA}_NSEC_CAP_RPMB_RPC` capability flag in `OPTEE_*_EXCHANGE_CAPABILITIES` command, when RPMB driver is supported?

This could ease addressing cases where Linux supports (and needs) RPMB subsystem while U-Boot does not (and rely on its tee-supplicant to be used.

core/pta/device.c
jforissier: That's probably right, but the fix as it is at least allows the test mentioned in #7215 to pass. So maybe we should keep it like so for the moment.

core/pta/device.c
jenswi-linaro: > Could Linux set an `OPTEE_{SMC|FFA}_NSEC_CAP_RPMB_RPC` capability flag in `OPTEE_*_EXCHANGE_CAPABILITIES` command, when RPMB driver is supported?
> 

Unfortunately not, due to:
``` C
        if (args->a1 & ~OPTEE_SMC_NSEC_CAP_UNIPROCESSOR) {
                /* Unknown capability. */
                args->a0 = OPTEE_SMC_RETURN_ENOTAVAIL;
                return;
        }
```

> This could ease addressing cases where Linux supports (and needs) RPMB subsystem while U-Boot does not (and rely on its tee-supplicant to be used.

tee-supplicant is always ready in U-Boot so it could do `optee_enumerate_devices(PTA_CMD_GET_DEVICES_RPMB)`, followed by `optee_enumerate_devices(PTA_CMD_GET_DEVICES_SUPP)` without lying.

etienne-lms: > > This could ease addressing cases where Linux supports (and needs) RPMB subsystem while U-Boot does not (and rely on its tee-supplicant to be used.
> 
> tee-supplicant is always ready in U-Boot so it could do optee_enumerate_devices(PTA_CMD_GET_DEVICES_RPMB), followed by optee_enumerate_devices(PTA_CMD_GET_DEVICES_SUPP) without lying.

It does not work well since eMMC device driver may not be initialized when optee driver is initialized. `PTA_CMD_GET_DEVICES_RPMB` makes OP-TEE core to init the RPMB driver that ends in device not being found (when eMMC is not inited), so in the end U-Boot `PTA_CMD_GET_DEVICES_RPMB` fails. This enumeration needs to wait both eMMC and optee device drivers are initialized.

etienne-lms: Regarding:
```
 if (args->a1 & ~OPTEE_SMC_NSEC_CAP_UNIPROCESSOR) {
                /* Unknown capability. */
                args->a0 = OPTEE_SMC_RETURN_ENOTAVAIL;
                return;
        }
```
there could some evolution here to allow NSEC to pass other capability bits. New kernel that set other bits and get a `OPTEE_SMC_RETURN_ENOTAVAIL` respond would know passing extra NSEC caps feature is not supported by OP-TEE and would need to re-issue a new EXCHANGE_CAPS command without these bits set.
Of course, there could also be new dedicated EXCHANGE_NSEC_CAPS smc/ffa command, or some new commands in PTA_DEVICE to exchange dedicated info between both sides.

jenswi-linaro: > It does not work well since eMMC device driver may not be initialized when optee driver is initialized. `PTA_CMD_GET_DEVICES_RPMB` makes OP-TEE core to init the RPMB driver that ends in device not being found (when eMMC is not inited), so in the end U-Boot `PTA_CMD_GET_DEVICES_RPMB` fails. This enumeration needs to wait both eMMC and optee device drivers are initialized.

Will it? I thought U-Boot initialized devices on demand.

jenswi-linaro: > Regarding:
> 
> ```
>  if (args->a1 & ~OPTEE_SMC_NSEC_CAP_UNIPROCESSOR) {
>                 /* Unknown capability. */
>                 args->a0 = OPTEE_SMC_RETURN_ENOTAVAIL;
>                 return;
>         }
> ```
> 
> there could some evolution here to allow NSEC to pass other capability bits. New kernel that set other bits and get a `OPTEE_SMC_RETURN_ENOTAVAIL` respond would know passing extra NSEC caps feature is not supported by OP-TEE and would need to re-issue a new EXCHANGE_CAPS command without these bits set. Of course, there could also be new dedicated EXCHANGE_NSEC_CAPS smc/ffa command, or some new commands in PTA_DEVICE to exchange dedicated info between both sides.

The usual way to do it is like for `OPTEE_SMC_FUNCID_ENABLE_ASYNC_NOTIF`.

etienne-lms: > > This enumeration needs to wait both eMMC and optee device drivers are initialized.
>
> Will it? I thought U-Boot initialized devices on demand.

Testing few configurations I see this constraint actually comes from system services we embedded in OP-TEE. Having SCMI services in OP-TEE make U-Boot to probe the optee driver very early, possibly before mmc driver is initialized. In our stm32mp configurations where the eMMC clocks/regulators depend on OP-TEE SCMI service, the dependency is even stronger.

etienne-lms:
I'm not fully convinced of the consistency of this change but I see no regression it could induce and indeed it fixes an service enumeration issue when core is built with `CFG_TEE_FS=n CFG_RPMB_FS=y CFG_RPMB_ANNOUNCE_PROBE_CAP=n`.
Therefore `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

Maybe remove the `FIXME` word, the comment rephrasing is quite explicit IMO. IIRC we said some day that `FIXME` and `TODO` inline comments were to be avoided in the implementation.

jforissier:
Would it be OK to call `tee_rpmb_reinit()` when needed to know if the kernel supports RPMB?
```C
case PTA_CMD_GET_DEVICES_SUPP:
	rflags = TA_FLAG_DEVICE_ENUM_SUPP;
	if (IS_ENABLED(CFG_REE_FS) ||
	    (IS_ENABLED_(CFG_RPMB_FS) &&
	     (!IS_ENABLED(CFG_RPMB_ANNOUNCE_PROBE_CAP) ||
	      tee_rpmb_reinit()))) {
	 	rflags |= TA_FLAG_DEVICE_ENUM_TEE_STORAGE_PRIVATE;
	}
```

jenswi-linaro:

PR #7224, opened 2025-01-16
core: mm: zero initialize phys_mem pool instances

etienne-lms:
Ensure phys_mem pool instance structure is zero initialized when created. This change fixes an issue where phys_mem pool max_allocated field may contain a fuzzy value.

Fixes: c596d8359eb3 ("core: add phys_mem allocation functions")

### DIFF
--- a/core/mm/tee_mm.c
+++ b/core/mm/tee_mm.c
@@ -51,12 +51,14 @@ bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, paddr_size_t size,
 
 	assert(((uint64_t)size >> shift) < (uint64_t)UINT32_MAX);
 
-	pool->lo = lo;
-	pool->size = size;
-	pool->shift = shift;
-	pool->flags = flags;
-	pool->entry = pcalloc(pool, 1, sizeof(tee_mm_entry_t));
+	*pool = (tee_mm_pool_t){
+		.lo = lo,
+		.size = size,
+		.shift = shift,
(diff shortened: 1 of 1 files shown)


core/mm/phys_mem.c
```diff
@@ -26,9 +26,9 @@ static tee_mm_pool_t *init_pool(paddr_t b, paddr_size_t sz, uint32_t flags)
 		panic("invalid phys mem");
 
 	if (flags & TEE_MM_POOL_NEX_MALLOC)
-		pool = nex_malloc(sizeof(*pool));
+		pool = nex_calloc(1, sizeof(*pool));
 	else
-		pool = malloc(sizeof(*pool));
+		pool = calloc(1, sizeof(*pool));
```
jenswi-linaro, line 31: While `calloc()` is a safe choice, I'd expect `tee_mm_init()` to be able to initialize the pool correctly regardless.
How about fixing `tee_mm_init()`?

jforissier: I was about to suggest the same thing ;)

etienne-lms: Ok, that would be more consistent. I'll update. Thanks.

etienne-lms: Updated. I've preserved the `Fixes` tag in the commit message. Tell me if you think it's not really applicable.

jenswi-linaro:

jforissier:

PR #7226, opened 2025-01-17
ta: pkcs11: Update PKCS11 ECDSA flags to align with GP Specification

anvisriv:
This commit updates the PKCS11 ECDSA-related flags to ensure they map correctly according to the GlobalPlatform (GP) specifications. This alignment helps maintain consistency and compatibility with the GP standards.

Addresses [Comment](https://github.com/OP-TEE/optee_test/issues/721#issuecomment-1970596624)

### DIFF
--- a/ta/pkcs11/src/processing_ec.c
+++ b/ta/pkcs11/src/processing_ec.c
@@ -438,41 +438,48 @@ enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
 				   struct pkcs11_object *obj)
 {
 	switch (proc_params->id) {
+	/* In case of ECDSA signing without hashing. */
 	case PKCS11_CKM_ECDSA:
-	case PKCS11_CKM_ECDSA_SHA1:
-	case PKCS11_CKM_ECDSA_SHA224:
-	case PKCS11_CKM_ECDSA_SHA256:
-	case PKCS11_CKM_ECDSA_SHA384:
-	case PKCS11_CKM_ECDSA_SHA512:
+		switch (get_object_key_bit_size(obj)) {
+		case 192:
+			*tee_id = TEE_ALG_ECDSA_P192;
+			break;
+		case 224:
+			*tee_id = TEE_ALG_ECDSA_P224;
+			break;
+		case 256:
+			*tee_id = TEE_ALG_ECDSA_P256;
+			break;
+		case 384:
+			*tee_id = TEE_ALG_ECDSA_P384;
+			break;
+		case 521:
+			*tee_id = TEE_ALG_ECDSA_P521;
+			break;
+		default:
+			TEE_Panic(0);
+			break;
+		}
 		break;
-	default:
+    /* In CASE OF ECDSA signing with Hash */
+    case PKCS11_CKM_ECDSA_SHA1:
+		*tee_id = TEE_ALG_ECDSA_SHA1;
+		break;
+    case PKCS11_CKM_ECDSA_SHA224:
+		*tee_id = TEE_ALG_ECDSA_SHA224;
+		break;
+    case PKCS11_CKM_ECDSA_SHA256:
+		*tee_id = TEE_ALG_ECDSA_SHA256;
+		break;
+    case PKCS11_CKM_ECDSA_SHA384:
+		*tee_id = TEE_ALG_ECDSA_SHA384;
+		break;
+    case PKCS11_CKM_ECDSA_SHA512:
+		*tee_id = TEE_ALG_ECDSA_SHA512;
+		break;
+    default:
 		return PKCS11_CKR_GENERAL_ERROR;
-	}
-
-	/*
-	 * TODO: Fixing this in a way to support also other EC curves would
-	 * require OP-TEE to be updated for newer version of GlobalPlatform API
-	 */
-	switch (get_object_key_bit_size(obj)) {
-	case 192:
-		*tee_id = TEE_ALG_ECDSA_P192;
-		break;
-	case 224:
-		*tee_id = TEE_ALG_ECDSA_P224;
-		break;
-	case 256:
-		*tee_id = TEE_ALG_ECDSA_P256;
-		break;
-	case 384:
-		*tee_id = TEE_ALG_ECDSA_P384;
-		break;
-	case 521:
-		*tee_id = TEE_ALG_ECDSA_P521;
-		break;
-	default:
-		TEE_Panic(0);
-		break;
-	}
+    }
 
 	return PKCS11_CKR_OK;
 }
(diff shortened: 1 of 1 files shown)


etienne-lms:
Hello @anvisriv, thanks for this P-R. I'll take a deeper look at it, considering also https://github.com/OP-TEE/optee_os/pull/6230. In the mean time these changes make raleted `xtest -t pkcs11` test cases to fail. Could you create a P-R on optee_test to make the necessary updates.

ta/pkcs11/src/processing_ec.c
```diff
@@ -438,41 +438,48 @@ enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
...
+	/* In case of ECDSA signing without hashing. */
 	case PKCS11_CKM_ECDSA:
-	case PKCS11_CKM_ECDSA_SHA1:
-	case PKCS11_CKM_ECDSA_SHA224:
-	case PKCS11_CKM_ECDSA_SHA256:
-	case PKCS11_CKM_ECDSA_SHA384:
-	case PKCS11_CKM_ECDSA_SHA512:
+		switch (get_object_key_bit_size(obj)) {
+		case 192:
+			*tee_id = TEE_ALG_ECDSA_P192;
```
etienne-lms, line 445: These `TEE_ALG_ECDSA_P*` IDs are deprecated. Prefer use `TEE_ALG_ECDSA_SHA*` ones instead.

ta/pkcs11/src/processing_ec.c
```diff
@@ -438,41 +438,48 @@ enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
...
-	case 224:
-		*tee_id = TEE_ALG_ECDSA_P224;
+    /* In CASE OF ECDSA signing with Hash */
+    case PKCS11_CKM_ECDSA_SHA1:
+		*tee_id = TEE_ALG_ECDSA_SHA1;
 		break;
-	case 256:
-		*tee_id = TEE_ALG_ECDSA_P256;
+    case PKCS11_CKM_ECDSA_SHA224:
+		*tee_id = TEE_ALG_ECDSA_SHA224;
```
etienne-lms, line 469: OP-TEE core implementation currently expects that `TEE_ALG_ECDSA_SHA224` is used only with 224bit private keys. And same for the other ECDSA flavors.
See https://github.com/OP-TEE/optee_os/blob/4.4.0/core/lib/libtomcrypt/ecc.c#L58-L64
and https://github.com/OP-TEE/optee_os/blob/4.4.0/lib/libutee/tee_api_operations.c#L2662-L2692.

For the pkcs11 TA to use any of `CMK_ECDSA_SHA*` with any EC key, we need to change these tests, assuming allowing these is fully safe and agreed by our dear OP-TEE maintainers.

By the way, could you fix the indentation here (lines 464 to 482).

ta/pkcs11/src/processing_ec.c
```diff
@@ -438,41 +438,48 @@ enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
 				   struct pkcs11_object *obj)
 {
 	switch (proc_params->id) {
+	/* In case of ECDSA signing without hashing. */
```
etienne-lms, line 441: Can remove this inline comment and the one at line 464.

ta/pkcs11/src/processing_ec.c
```diff
@@ -847,16 +854,20 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
 
 	TEE_GetOperationInfo(op, &info);
 
-	switch (info.algorithm) {
-	case TEE_ALG_ECDSA_P192:
+	/*
+	 *  Determining curve size in bytes with the help of
+	 *  maxkeysize attribute instead of algorithm
```
etienne-lms, line 859: ìnstead of algorithm`: this comment should rather be in the commit message, not in the source file.

etienne-lms (commented):
These changes look legitimate to me. The issue faced with these changes IMO is that  OP-TEE Test `pkcs11_1000.c` current implementation tests all of [CKM_ECDSA_SHA{1|224|256|384|512} mechanisms](https://github.com/OP-TEE/optee_test/blob/4.4.0/host/xtest/pkcs11_1000.c#L5573-L5579) for each of the [tested NIST curves](https://github.com/OP-TEE/optee_test/blob/4.4.0/host/xtest/pkcs11_1000.c#L5872-L5883) but OP-TEE OS currently does not allow (see my review comment), hence the xtest failures in all CI test reports.

ta/pkcs11/src/processing_ec.c
anvisriv: We can modify the tests as suggested, but wouldn’t it be better to fix the GP implementation in OPTEE?

ta/pkcs11/src/processing_ec.c
anvisriv: According to GP specifications, ‘ECDSA algorithm identifiers should be tied to the size of the digest, not the key.’ So, in the case of PKCS11_CKM_ECDSA mode, the mapping should be from key to digest? Like mapping of P192 to SHA1.

ta/pkcs11/src/processing_ec.c
etienne-lms: Right, I've wrote 2 comments that contradict themselves. So, all in 2:
Option 1: restrict ECDSA_SHA1 to 192bit keys, ECDSA_SHA224 to 224bit keys, etc... which is the current OP-TEE implementation (and keep OP-TEE tests as they are).
Option 2: change OP-TEE OS to allow ECDSA_SHA* on all supported EC curves (key sizes).

I guess option 2 would be better but it deserves a deeper changes in OP-TEE core.

etienne-lms (commented):
Hi @anvisriv,
For the change you want, I think we need deeper changes in OP-TEE core asymm crypto implementation to clarify implementation of _TEE_ALG_ECDSA_SHA*_ algorithms. I'll try to propose something but if you have ideas, do no hesitate to create a P-R with some proposal we can discuss.

ta/pkcs11/src/processing_ec.c
anvisriv: For CKM_ECDSA, is it acceptable to map to TEE_ALG_ECDSA_SHA1 or TEE_ALG_ECDSA_SHA224 until the OPTEE core resolves the ECDSA GP implementation issue?

ta/pkcs11/src/processing_ec.c
anvisriv: For the time being, we can proceed with option 1 until the OP-TEE core issue is resolved.
I can raise a patch in test as well. If that works for PKCS11 maintainers?

anvisriv:
Hi @etienne-lms 
Could you review my responses? If they work for you, I’ll proceed with raising new patches.

ta/pkcs11/src/processing_ec.c
etienne-lms: That would break an existing OP-TEE PKCS#11 support and few tests, because OP-TEE core falsely assumes TEE_ALG_ECDSA_SHA1 relates to 192bit keys, ..._SHA224 to 224bit keys etc...

I'd rather address the issue in OP-TEE core. I'll create a dedicate Issue entry to discuss possible changes in core.

ta/pkcs11/src/processing_ec.c
etienne-lms: Such change would break test but also cases already supported by OP-TEE. @vesajaaskelainen, you may want to give you point of view.
As said above, I've rather have OP-TEE core fixed against this inconsistency and update the pkcs11 TA afterward.

For info, I am maintaining the pkcs11 TA :)

ta/pkcs11/src/processing_ec.c
anvisriv: Hi @etienne-lms 
Can we introduce these changes as part of a GP version check? This way, once the GP issue is resolved, we can enable the changes, and in the meantime, it will continue to function as it currently does.

etienne-lms: Since such changes could break TAs, I've created a dedicated Issue entry to track this issue (https://github.com/OP-TEE/optee_os/issues/7235). I propose we move the discussion there and find a way to address that in a smooth way.

etienne-lms:
For the record: see also issue https://github.com/OP-TEE/optee_os/issues/7235.

anvisriv:
Keeping alive

PR #7227, opened 2025-01-20
Malloc flags and cleanup

jenswi-linaro:

### DIFF
--- a/lib/libutils/isoc/bget_malloc.c
+++ b/lib/libutils/isoc/bget_malloc.c
@@ -502,16 +502,14 @@ void *raw_realloc(void *ptr, size_t hdr_size, size_t ftr_size,
 			     pl_size, ctx);
 }
 
-#ifdef ENABLE_MDBG
-
 struct mdbg_hdr {
 	const char *fname;
 	uint16_t line;
-	uint32_t pl_size;
-	uint32_t magic;
-#if defined(ARM64)
+#ifdef __LP64__
 	uint64_t pad;
 #endif
+	uint32_t pl_size;
+	uint32_t magic;
 };
 
 #define MDBG_HEADER_MAGIC	0xadadadad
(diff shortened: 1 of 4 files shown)


lib/libutils/isoc/bget_malloc.c
```diff
@@ -782,38 +791,21 @@ static void free_helper(void *ptr, bool wipe)
...
 }
 
-static void *realloc_unlocked(struct malloc_ctx *ctx, void *ptr,
-			      size_t size)
-{
-	return raw_realloc(ptr, 0, 0, size, ctx);
-}
+/* For use in raw_malloc_add_pool() below */
+#define realloc_unlocked(ctx, ptr, size) \
+	raw_mem_alloc(MAF_NULL, ptr, 0, 0, 1, 1, size, ctx)
```
jforissier: Keeping a static function is preferable IMO

jforissier (approved):
One minor comment, otherwise:

lib/libutils/isoc/bget_malloc.c
jenswi-linaro: This is an intermediate change that in the end turns into:
``` C
#define realloc_unlocked(ctx, ptr, size)                                      \
	mem_alloc_unlocked(MAF_NULL, (ptr), 1, 1, (size), __FILE__, __LINE__, \
			   (ctx))
```
This one should preferably be a macro for usable file and line info. I can rework the intermediate change if it's worth it.

jenswi-linaro:
Tag applied

lib/libutils/isoc/include/malloc.h
```diff
@@ -91,33 +83,27 @@ void malloc_reset_stats(void);
 
 #ifdef CFG_NS_VIRTUALIZATION
 
+void *nex_malloc(size_t size);
+void *nex_calloc(size_t nmemb, size_t size);
+void *nex_realloc(void *ptr, size_t size);
+void *nex_memalign(size_t alignment, size_t size);
```
etienne-lms, line 95: These function declaration are now defined even when `ENABLE_MDBG` is disabled. Won't they conflict with the macros defined right below when `ENABLE_MDBG` is enabled?

etienne-lms (commented):
The overall look good to me but I have 2 questions?

lib/libutils/isoc/include/malloc.h
jenswi-linaro: They don't conflict, but they are unnecessary and confusing. I'll change to only define them without `ENABLE_MDBG`.

lib/libutils/isoc/include/malloc_flags.h
```diff
@@ -0,0 +1,30 @@
...
+ * Memory allocation flags to control how buffers are allocated. Flags may
+ * be ignored depending on configuration or if they don't apply. These
+ * flags are primarily used by malloc() and friends in malloc.h, but can be
+ * extended with flags that only mean something for other functions to
+ * avoid needless translation of one class of flags to another class of
+ * flags.
+ */
+#define MAF_NULL	0x0	/* Passed if no flags are needed */
+#define MAF_ZERO_INIT	0x1	/* Zero initialize the allocated buffer */
+#define MAF_NEX		0x2	/* Allocate from nex heap */
```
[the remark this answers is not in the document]
jenswi-linaro: You're right, I'll update.

jenswi-linaro:
Update

lib/libutils/isoc/include/malloc.h
```diff
@@ -86,37 +86,29 @@ void malloc_get_stats(struct pta_stats_alloc *stats);
...
 
+#define nex_mdbg_check(x)        do { } while (0)
+
 void *nex_malloc(size_t size);
 void *nex_calloc(size_t nmemb, size_t size);
 void *nex_realloc(void *ptr, size_t size);
 void *nex_memalign(size_t alignment, size_t size);
-
-#define nex_mdbg_check(x)        do { } while (0)
+void nex_free(void *ptr);
```
etienne-lms, line 111: `nex_free()` needs to be declared whatever  `ENABLE_MDBG` value.

lib/libutils/isoc/include/malloc.h
etienne-lms: Maybe do the same for `malloc()` and friends.

lib/libutils/isoc/include/malloc.h
jenswi-linaro: Yes, you're right. I'll fix.

lib/libutils/isoc/include/malloc.h
jenswi-linaro: No, those should be kept due to commit eacabbbceb43 ("libutils: ta: provide malloc(), calloc() and realloc() when debug is on")

jenswi-linaro:
Update

etienne-lms (commented):
For commit "libutils: refactor the MDBG malloc functions":
could you replace `mdbg_malloc()` with `__mdbg_malloc()` in `ta/ta.mk` inline comment?

ta/ta.mk
```diff
@@ -22,8 +22,8 @@ $(sm)-platform-cppflags += -DMBEDTLS_SELF_TEST
 endif
 
 ifeq ($(CFG_TEE_TA_MALLOC_DEBUG),y)
-# Build malloc debug code into libutils: (mdbg_malloc(), mdbg_free(),
-# mdbg_check(), etc.).
+# Build malloc debug code into libutils: (__mdbg_alloc(), mdbg_free(),
+# and mdbg_check().).
```
jforissier, line 26: It looks kind of weird to have `__mdbg_alloc()` and no `mdbg_malloc()` although there is `mdbg_free()`. I think it would be better to not mention these implementation details. TAs are not supposed to depend on them. Better mention what it does (more checks, stats?)

jenswi-linaro: OK, I'll update.

jenswi-linaro:
Update

jforissier:

jenswi-linaro:
Squashed

etienne-lms (approved):

PR #7229, opened 2025-01-21
core: fix potential panic when setting transfer entry size

raymo200915:
Fix a potential panic when the roundup end-of-entry exceeds the boundary of the old end-of-transfer-list when setting the size of a transfer entry.

### DIFF
--- a/core/kernel/transfer_list.c
+++ b/core/kernel/transfer_list.c
@@ -407,7 +407,8 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
 		tl->size -= mov_dis;
 	}
 	/* Move all following entries to fit in the expanded or shrunk space */
-	memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
+	if (tl_old_ev > old_ev)
+		memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
 
 	/*
 	 * Fill the gap due to round up/down with a void entry if the size of


core/kernel/transfer_list.c
```diff
@@ -407,7 +407,8 @@ bool transfer_list_set_data_size(struct transfer_list_header *tl,
 		tl->size -= mov_dis;
 	}
 	/* Move all following entries to fit in the expanded or shrunk space */
-	memmove((void *)r_new_ev, (void *)old_ev, tl_old_ev - old_ev);
+	if (tl_old_ev - old_ev)
```
jenswi-linaro, line 410: I don't see how this check can fix any panic.
By the way, it seems we're dealing with a panic caused by an exception due to memory that we have just corrupted. Please update the commit message to better describe the error we're fixing.

raymo200915: > I don't see how this check can fix any panic. By the way, it seems we're dealing with a panic caused by an exception due to memory that we have just corrupted. Please update the commit message to better describe the error we're fixing.

The change is amended with updated commit message.

jforissier:
Do we have any CI tests for transfer lists?

raymo200915:
> Do we have any CI tests for transfer lists?

No in OP-TEE at the moment, but we do have in TF-A and U-Boot(bloblist).

jenswi-linaro:
> > Do we have any CI tests for transfer lists?
> 
> No in OP-TEE at the moment, but we do have in TF-A and U-Boot(bloblist).

We need it in OP-TEE CI too. Would you mind adding it?

jenswi-linaro:
Please replace "potential panic" with "potential crash" in the commit message.

It's clear that we're dealing with an integer underflow, but it's a bit hard to understand how it happens. Why is `old_ev` ever larger than `tl_old_ev`?

raymo200915:
> > > Do we have any CI tests for transfer lists?
> > 
> > 
> > No in OP-TEE at the moment, but we do have in TF-A and U-Boot(bloblist).
> 
> We need it in OP-TEE CI too. Would you mind adding it?

I will add this when TPM eventlog handoff feature is done.

raymo200915:
> Please replace "potential panic" with "potential crash" in the commit message.

Updated with replacing "panic" with "crash".

> 
> It's clear that we're dealing with an integer underflow, but it's a bit hard to understand how it happens. Why is `old_ev` ever larger than `tl_old_ev`?

The `tl_old_ev` points to the actual end of the TL which is not rounded up. When user increase the data size of the last entry, potentially `old_ev` would be rounded up.

jenswi-linaro:
This sentence in the commit message needs to be reworked to make sense:
> Logically only when 'tl_old_ev > old_ev' which indicates there are following entries that need to be moved.

raymo200915:
> This sentence in the commit message needs to be reworked to make sense:
> 
> > Logically only when 'tl_old_ev > old_ev' which indicates there are following entries that need to be moved.

Fixed that grammar issue.

jenswi-linaro:

raymo200915:
Added the `Acked-by` tag.
All review comments are addressed.

PR #7230, opened 2025-01-21
core: expand the fdt transfer entry right before it is being used

raymo200915:
Move the expanding of fdt transfer entry from boot_init_primary_early to boot_init_primary_late.


jenswi-linaro:
Please add the motivation for this change to the commit message.

jenswi-linaro:
> > Please add the motivation for this change to the commit message.
> 
> More information is added in the commit message.

I assume you're doing this to avoid an experienced or anticipated error. Please add that to the description.

raymo200915:
> > > Please add the motivation for this change to the commit message.
> > 
> > 
> > More information is added in the commit message.
> 
> I assume you're doing this to avoid an experienced or anticipated error. Please add that to the description.

Actually no errors with the current code, but since all of fdt operations happen in boot_init_primary_late, and the expanding of fdt entry should be there as well. It is an improvement for making clear the logic other than a fix.

PR #7231, opened 2025-01-22
drivers: dra7_rng: Change dra7_rng_init to service_init_crypto

glneo:
Since commit 11d8578d93f0 ("core: arm: call call_driver_initcalls() late"), driver_init is deferred and thread_update_canaries tries to get random_stack_canaries which requires the TRNG driver to be setup. Since it was being setup as part of driver_init, it lead to crash on DRA7 platforms.

Change driver_init to service_init_crypto which is meant to be used for initialization of crypto operations.

While here, add an assert that checks for DRA7 TRNG being used before initialization is complete.

### DIFF
--- a/core/drivers/dra7_rng.c
+++ b/core/drivers/dra7_rng.c
@@ -112,6 +112,8 @@ TEE_Result hw_get_random_bytes(void *buf, size_t len)
 	uint8_t *buffer = buf;
 	size_t buffer_pos = 0;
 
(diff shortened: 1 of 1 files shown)


jforissier (commented):

jenswi-linaro:

jforissier:
@glneo would you mind adding the review tags? Thanks!

PR #7234, opened 2025-01-23
core: mmu: fix memory regions found from ff-a manifest

ysbnim:
Fix the 5th parameter of add_phys_mem() in collect_device_mem_ranges() that has to be the size of memory region and not the end address of the region.

Fixes: b8ef8d0 ("core: mm: introduce struct memory_map")


ysbnim:
This was solved by https://github.com/OP-TEE/optee_os/pull/6811 once.
But there is a regression in this commit https://github.com/OP-TEE/optee_os/commit/b8ef8d0b6ff45bcf0c1e092fbe61cb5a32f83e8e .

jenswi-linaro:

PR #7236, opened 2025-01-24
[RFC] riscv: drivers: Implement APLIC and IMSIC drivers

HuangBorong:
Implement the APLIC and IMSIC drivers which introduced by the "RISC-V Advanced Interrupt Architecture" specification: [https://github.com/riscv/riscv-aia](url)

The drivers support two modes:

1. APLIC only:
Some systems without IMSIC only have APLIC, which replaces the original PLIC as the Advanced PLIC. The role of APLIC is to serve as interrupt controllers.
2. APLIC + IMSIC:
In these systems, both APLIC and IMSIC are present. The harts receive external interrupts only in the form of MSIs. Therefore, the role of APLIC is to convert wired interrupts into MSIs for harts, while the interrupt controllers in these systems are IMSICs.

There are some limitations in the drivers:

1. It is assumed that the OP-TEE OS runs in S-mode. Consequently, the APLIC driver only interacts with the supervisor-level interrupt domain, while IMSIC only interacts with supervisor-level interrupt files.
2. It relies on OpenSBI to perform some initialization to APLIC and IMSIC in M-mode.
3. It does not support systems with multiple APLICs (NUMA).
4. It does not support multiple interrupt domains in APLIC; in other words, it only supports a single S-level interrupt domain.


HuangBorong:
By the way, I used some of the code related to APLIC and IMSIC from Linux and OpenSBI. Should I inform the original authors or take any specific actions?

jenswi-linaro:
> By the way, I used some of the code related to APLIC and IMSIC from Linux and OpenSBI. Should I inform the original authors or take any specific actions?

If you copy code, you must do so under the license provided by the copyright holder and preserve the copyright. However, we don't accept GPL licenses (copyleft) in OP-TEE. The copyright holder may agree to change the license, but it can get tricky if there are many copyright holders.

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		if (!aplic->targets_mmode)
+			return TEE_SUCCESS;
+
+		node =
+		    fdt_node_offset_by_compatible(fdt, node, APLIC_COMPATIBLE);
+	}
+
+	if (aplic->targets_mmode)
```
etienne-lms, line 101: The loops seems to parse several node but take into account only a single one. Maybe a bit of inline comment could help understanding what is expected by this function.

core/include/drivers/aplic_priv.h
```diff
@@ -0,0 +1,107 @@
...
+ * Copyright (c) 2025 Beijing Institute of Open Source Chip (BOSC)
+ */
+
+#ifndef __DRIVERS_APLIC_PRIV_H
+#define __DRIVERS_APLIC_PRIV_H
+
+#include <io.h>
+#include <util.h>
+#include <types_ext.h>
+#include <kernel/interrupt.h>
```
etienne-lms, line 12: sort in alphabetical order.

etienne-lms (commented):
Comments to *core/drivers/aplic_direct.c* may apply to other source files.
Please fix all indention issue. Can use `scripts/checkpatch.sh` to fund them.

HuangBorong:
> If you copy code, you must do so under the license provided by the copyright holder and preserve the copyright. However, we don't accept GPL licenses (copyleft) in OP-TEE. The copyright holder may agree to change the license, but it can get tricky if there are many copyright holders.

Thank you for your insights. I need to consider removing the part of code copied from Linux and rewrite the relevant parts. Additionally, since OpenSBI [https://github.com/riscv-software-src/opensbi](url) under the BSD-2-Clause license, I can use their code in my PR by preserve the copyright, right?

jenswi-linaro:
> Additionally, since OpenSBI [https://github.com/riscv-software-src/opensbi](url) under the BSD-2-Clause license, I can use their code in my PR by preserve the copyright, right?

Correct.

core/drivers/aplic_msi.c
```diff
@@ -0,0 +1,174 @@
...
+
+#define APLIC_DEFAULT_EIID		2
+
+static struct aplic_data aplic_data __nex_bss;
+
+static void aplic_set_target(struct aplic_data *aplic, uint32_t source,
+			     uint32_t hart_idx, uint32_t guest_idx,
+			     uint32_t eiid)
+{
+	vaddr_t target;
```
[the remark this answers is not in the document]
HuangBorong: Thanks! Fixed

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,191 @@
...
+	io_write32(idc_base + APLIC_IDC_ITHRESHOLD, APLIC_ENABLE_ITHRESHOLD);
+}
+
+void aplic_it_handle(void)
+{
+	struct aplic_data *aplic = &aplic_data;
+	uint32_t id = 0;
+
+	id = io_read32(aplic->aplic_base + APLIC_IDC_CLAIMI);
+	id >>= APLIC_IDC_TOPI_ID_SHIFT;
```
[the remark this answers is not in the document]
HuangBorong: I think shift id is ok because the interrupt identity in bits 25:16 is written by hardwire and will not overflow.

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,191 @@
...
+		panic();
+	aplic_set_target(aplic, it, hartid, prio);
+}
+
+static void aplic_op_enable(struct itr_chip *chip, size_t it)
+{
+	struct aplic_data *aplic = container_of(chip, struct aplic_data, chip);
+
+	if (!it || it > aplic->num_source)
+		panic();
```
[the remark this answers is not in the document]
HuangBorong: Thanks! I add a helper function called `aplic_is_bad_it()` to replace this.

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,138 @@
...
+	if (rc < 0 || !reg_addr || !reg_size)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+	aplic->aplic_base =
+	    core_mmu_get_va(reg_addr, MEM_AREA_IO_SEC, reg_size);
+	if (!aplic->aplic_base)
+		return TEE_ERROR_GENERIC;
+	aplic->size = reg_size;
+
+	val = fdt_getprop(fdt, nodeoff, "riscv,num-sources", &len);
+	if (len > 0)
```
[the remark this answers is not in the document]
HuangBorong: I copied this code from OpenSBI. The APLIC `num-sources` property looks like this: `riscv,num-sources = <63>;` Only the first cell is used.

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,138 @@
...
+#include <kernel/dt_driver.h>
+#include <libfdt.h>
+#include <string.h>
+
+static TEE_Result fdt_parse_aplic_node(const void *fdt, int nodeoff,
+				       struct aplic_data *aplic)
+{
+	const fdt32_t *val;
+	struct imsic_data imsic = { 0 };
+	int i, len, noff, rc;
```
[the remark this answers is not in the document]
HuangBorong: Fix variable initialization problems on 2b172bf

core/drivers/aplic_priv.c
HuangBorong: Thanks! I added some comments to explain the reason for that in commit 2b172bf

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,191 @@
...
+
+static const struct itr_ops aplic_ops = {
+	.add = aplic_op_add,
+	.enable = aplic_op_enable,
+	.disable = aplic_op_disable,
+	.mask = aplic_op_disable,
+	.unmask = aplic_op_enable,
+	.raise_pi = aplic_op_raise_pi,
+	.raise_sgi = NULL,
+	.set_affinity = NULL
```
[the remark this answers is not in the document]
HuangBorong: Removed in commit 2b172bf

core/include/drivers/aplic_priv.h
HuangBorong: Reordered the header files in alphabetical order in commit 2b172bf

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,191 @@
...
+#include <kernel/panic.h>
+#include <kernel/interrupt.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <drivers/aplic.h>
+#include <drivers/aplic_priv.h>
+
+#define APLIC_MAX_IDC			           BIT(14)	// 16384
+#define APLIC_IDC_BASE			            0x4000
+#define APLIC_IDC_SIZE			                32
```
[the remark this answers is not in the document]
HuangBorong: Got it! I removed redundant comments in commit 2b172bf

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,191 @@
...
+/* Interrupt Delivery Control (IDC) structure*/
+#define APLIC_IDC_IDELIVERY                   0x00	// idelivery
+#define APLIC_IDC_IFORCE		              0x04	// iforce
+#define APLIC_IDC_ITHRESHOLD	              0x08	// ithreshold
+
+#define APLIC_IDC_TOPI			              0x18	// topi
+#define APLIC_IDC_TOPI_ID_SHIFT		            16
+#define APLIC_IDC_TOPI_ID_MASK               0x3FF
+#define APLIC_IDC_TOPI_ID		    (APLIC_IDC_TOPI_ID_MASK << \
+				APLIC_IDC_TOPI_ID_SHIFT)	// bits 25:16
```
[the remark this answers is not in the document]
HuangBorong: Thanks! I believe the macros `SHIFT_U32()` and `GENMASK_32` make the code clearer, so I used them in commit 2b172bf

core/drivers/aplic_msi.c
```diff
@@ -0,0 +1,174 @@
...
+	    (hart_idx & APLIC_TARGET_HART_IDX_MASK) <<
+	    APLIC_TARGET_HART_IDX_SHIFT;
+	val |=
+	    (guest_idx & APLIC_TARGET_GUEST_IDX_MASK) <<
+	    APLIC_TARGET_GUEST_IDX_SHIFT;
+	val |= (eiid & APLIC_TARGET_EIID_MASK) << APLIC_TARGET_EIID_SHIFT;
+
+	target =
+	    aplic->aplic_base + APLIC_TARGET_BASE + (source -
+						     1) * sizeof(uint32_t);
```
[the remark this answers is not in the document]
HuangBorong: Fixed the indentation issue in commit 2b172bf

gagachang:
@maroueneboubakri Do you have interest to review this PR?

core/drivers/imsic.c
```diff
@@ -0,0 +1,376 @@
...
+
+static void
+imsic_local_eix_update(uint32_t base_id, uint32_t num_id, bool pend, bool val)
+{
+	uint32_t i, isel, ireg;
+	uint32_t id = base_id, last_id = base_id + num_id;
+
+	while (id < last_id) {
+		isel = id / RISCV_XLEN_BITS;
+		isel *= RISCV_XLEN_BITS / IMSIC_EIPx_BITS;
```
etienne-lms, line 125: `isel = ROUNDDOWN(id, RISCV_XLEN_BITS) / IMSIC_EIPx_BITS` maybe be more explicit.

core/drivers/imsic.c
```diff
@@ -0,0 +1,376 @@
...
+	return TEE_ERROR_ITEM_NOT_FOUND;
+}
+
+void imsic_it_handle(void)
+{
+	struct imsic_data *imsic = &imsic_data;
+	uint32_t id = imsic_claim_interrupt();
+
+	if (id == IMSIC_IPI_ID)
+		IMSG("Interprocessor interrupt");
```
etienne-lms, line 357: Maybe `DMSG()`?

core/drivers/imsic.c
```diff
@@ -0,0 +1,376 @@
...
+	while (node != -FDT_ERR_NOTFOUND) {
+		res = fdt_parse_imsic_node(fdt, node, imsic);
+		if (res) {
+			EMSG("Parse IMSIC node failed");
+			return res;
+		}
+		if (!imsic->targets_mmode)
+			return TEE_SUCCESS;
+		node =
+		    fdt_node_offset_by_compatible(fdt, node, IMSIC_COMPATIBLE);
```
etienne-lms, line 345: prefer:
```c
		node = fdt_node_offset_by_compatible(fdt, node,
						     IMSIC_COMPATIBLE);

File: core/drivers/imsic.c [hunk 15]
```diff
@@ -0,0 +1,392 @@
...
+#define IMSIC_DISABLE_EITHRESHOLD 1
+#define IMSIC_ENABLE_EITHRESHOLD  0
+
+static struct imsic_data imsic_data __nex_bss;
+
+/*
+ * The IMSIC CSRs need to be indirectly accessed through
+ * the *iselect(miselect/siselect) and *ireg(mireg/sireg) CSRs.
+ */
+#define imsic_csr_write(__c, __v)	\
```
etienne-lms, line 61: Prefer upper case for macro names. Also indent the macro value:
```c
#define IMISC_CSR_WRITE(__c, __v)		\
	do {					\
		write_csr(CSR_XISELECT, __c);	\
		write_csr(CSR_XIREG, __v);	\
	} while (0)
```

Maybe consider using function, unless there are strong performance constrinats.

core/drivers/imsic.c
```diff
@@ -0,0 +1,392 @@
...
+do { \
+	write_csr(CSR_XISELECT, __c); \
+	write_csr(CSR_XIREG, __v); \
+} while (0)
+
+#define imsic_csr_read(__c)	\
+({ \
+	unsigned long __v; \
```
etienne-lms, line 69: as per codiing style: need a init value + add an empty line below:
```c
	unsigned long __v = 0; \
	\
	write_csr(CSR_XISELECT, __c); \

File: core/drivers/aplic_priv.c [hunk 17]
```diff
@@ -0,0 +1,151 @@
...
+		for (i = 0; i < len; i += 2) {
+			if (fdt32_to_cpu(val[i + 1]) == IRQ_M_EXT) {
+				aplic->targets_mmode = true;
+				break;
+			}
+		}
+		aplic->num_idc = len / 2;
+	} else {
+		val = fdt_getprop(fdt, nodeoff, "msi-parent", &len);
+		if (val && (size_t)len >= sizeof(fdt32_t)) {
```
etienne-lms, line 66: If there is no `interrupt-extended` nor `msi-parent` property, shouldn't this function return a error?
If `interrupt-extended` as only 1 cell, is it intentional to discard its value and consider only `msi-parent` property?

etienne-lms (commented):
Some coding style and other minor comments.

gagachang:
Hi @HuangBorong , An year ago I implemented a simple UART RX interrupt handler to test secure interrupt.
You can add the following code into `plat-virt/main.c` to test it.
Just boot to OP-TEE, enable interrupts, and enter some keys in OP-TEE console to see if something is printed.
The code is old, you may need to modify the code to be compatible with current OP-TEE.


```c
static void read_console(void)
{
	struct serial_chip *cons = &console_data.chip;

	if (!cons->ops->getchar || !cons->ops->have_rx_data)
		return;

	while (cons->ops->have_rx_data(cons)) {
		int ch __maybe_unused = cons->ops->getchar(cons);

		DMSG("got 0x%x", ch);
	}
}

static enum itr_return console_itr_cb(struct itr_handler *h __maybe_unused)
{
	read_console();
	return ITRR_HANDLED;
}

static struct itr_handler console_itr = {
	.it = UART1_IRQ,
	.flags = ITRF_TRIGGER_LEVEL,
	.handler = console_itr_cb,
};
DECLARE_KEEP_PAGER(console_itr);

static TEE_Result init_console_itr(void)
{

	console_itr.chip = interrupt_get_main_chip();
	res = interrupt_add_configure_handler(&console_itr, IRQ_TYPE_LEVEL_HIGH,
					      1);
	if (res)
		return res;

	interrupt_enable(console_itr.chip, console_itr.it);

	return TEE_SUCCESS;
}
driver_init(init_console_itr);
```

gagachang:
@HuangBorong , About the Supervisor top interrupt CSR (stopi), yes we definitely need it.
A simple solution is to implement AIA version of `thread_native_interrupt_handler()`.
For example:

```c
#if defined(CFG_RISCV_APLIC_MSI) || defined(CFG_RISCV_IMSIC)
void thread_native_interrupt_handler(struct thread_ctx_regs *regs,
				     unsigned long cause)
{
	unsigned long topi;

	while ((topi = csr_read(CSR_TOPI))) {
		topi = topi >> TOPI_IID_SHIFT;

		switch (topi) {
		case IRQ_XTIMER:
			clear_csr(CSR_XIE, CSR_XIE_TIE);
			break;
		case IRQ_XSOFT:
			thread_unhandled_trap(regs, cause);
			break;
		case IRQ_XEXT:
			thread_irq_handler();
			break;
		default:
			thread_unhandled_trap(regs, cause);
		}
	}
}
#endif
```

core/drivers/aplic_priv.c
HuangBorong: > If there is no `interrupt-extended` nor `msi-parent` property, shouldn't this function return a error?

I beleive that if the node has neither the `interrupt-extended` property nor the `msi-parent` property, it's an illegal node, and this function should return an error.

> If `interrupt-extended` as only 1 cell, is it intentional to discard its value and consider only msi-parent property?

If `interrupt-extended` has only 1 cell, it's also an illegal node, and this function should return an error.

Thank you for mentioning that. I will fix this in the next commit.

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,151 @@
...
+	int node = FDT_ERR_NOTFOUND;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	fdt = get_dt();
+	if (!fdt) {
+		EMSG("Unable to get DTB, APLIC init failed");
+		return TEE_ERROR_ITEM_NOT_FOUND;
+	}
+
+	/* Currently, only the S-level interrupt domain is considered.
```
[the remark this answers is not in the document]
HuangBorong: Fixed, thanks!

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,191 @@
...
+	aplic->aplic_base = aplic_base;
+	aplic->num_source = APLIC_NUM_SOURCE;
+	aplic->num_idc = APLIC_NUM_IDC;
+}
+
+static void aplic_op_add(struct itr_chip *chip, size_t it, uint32_t type,
+			 uint32_t prio)
+{
+	struct aplic_data *aplic = container_of(chip, struct aplic_data, chip);
+	size_t hartid = get_core_pos();
```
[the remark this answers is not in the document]
gagachang: > `struct itr_ops::add` may be called without foreign interrupts being masked. Should this be handled here?

How does GIC handle this case ?
`get_gicr_base()` calls `get_core_pos()` too.
`get_gicr_base()` is also called in `gic_op_add()`.

core/drivers/imsic.c
HuangBorong: Thanks! I refactored these IMSIC CSRs access macros in commit 1939b8f：

1. Using functions instead
2. Removed redundant `inline`
3. Added the `__unused` macro to those unused functions

core/drivers/imsic.c
HuangBorong: According to AIA spec Chapter 7, Interprocessor Interrupts (IPIs), IMSIC can handle IPIs; however, I have not implemented this functionality at this time. So I just print a message here.

core/drivers/imsic.c
HuangBorong: Thanks! I used it in commit 1939b8f.

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,181 @@
...
+	io_write32(idc_base + APLIC_IDC_IDELIVERY, APLIC_ENABLE_IDELIVERY);
+	io_write32(idc_base + APLIC_IDC_ITHRESHOLD, APLIC_ENABLE_ITHRESHOLD);
+}
+
+void aplic_it_handle(void)
+{
+	struct aplic_data *aplic = &aplic_data;
+	uint32_t id = 0;
+
+	id = io_read32(aplic->aplic_base + APLIC_IDC_CLAIMI);
```
gagachang, line 174: This is wrong. This should be 
```c
vaddr_t idc_base = aplic_get_idc_base();

id = io_read32(idc_base + APLIC_IDC_CLAIMI);
```

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,181 @@
...
+}
+
+static void aplic_set_target(struct aplic_data *aplic, uint32_t source,
+			     uint32_t hart_idx, uint32_t iprio)
+{
+	vaddr_t target = 0;
+	uint32_t val = 0;
+
+	val = SHIFT_U32(hart_idx & APLIC_TARGET_HART_IDX_MASK,
+			APLIC_TARGET_HART_IDX_SHIFT);
```
gagachang, line 59: Based on `APLIC_TARGET_HART_IDX_MASK`, the mask operation should be done after shift.
```c
	val = SHIFT_U32(hart_idx, APLIC_TARGET_HART_IDX_SHIFT) & APLIC_TARGET_HART_IDX_MASK;
```

core/drivers/aplic_msi.c
```diff
@@ -0,0 +1,154 @@
...
+
+static void aplic_set_target(struct aplic_data *aplic, uint32_t source,
+			     uint32_t hart_idx, uint32_t guest_idx,
+			     uint32_t eiid)
+{
+	vaddr_t target = 0;
+	uint32_t val = 0;
+
+	val = SHIFT_U32(hart_idx & APLIC_TARGET_HART_IDX_MASK,
+			APLIC_TARGET_HART_IDX_SHIFT);
```
gagachang, line 36: According to `APLIC_TARGET_HART_IDX_MASK`, the mask operation should be done after shift.

```c
	val = SHIFT_U32(hart_idx, APLIC_TARGET_HART_IDX_SHIFT) & APLIC_TARGET_HART_IDX_MASK;
```

core/drivers/aplic_msi.c
```diff
@@ -0,0 +1,154 @@
...
+	val |= SHIFT_U32(guest_idx & APLIC_TARGET_GUEST_IDX_MASK,
+			APLIC_TARGET_GUEST_IDX_SHIFT);
```
gagachang: According to `APLIC_TARGET_GUEST_IDX_MASK`, the mask operation should be done after shift.

```c
	val |= SHIFT_U32(guest_idx, APLIC_TARGET_GUEST_IDX_SHIFT) & APLIC_TARGET_GUEST_IDX_MASK;
```

core/drivers/aplic_msi.c
```diff
@@ -0,0 +1,160 @@
...
+			     uint32_t hart_idx, uint32_t guest_idx,
+			     uint32_t eiid)
+{
+	vaddr_t target = 0;
+	uint32_t val = 0;
+
+	val = SHIFT_U32(hart_idx & APLIC_TARGET_HART_IDX_MASK,
+			APLIC_TARGET_HART_IDX_SHIFT);
+	val |= SHIFT_U32(guest_idx & APLIC_TARGET_GUEST_IDX_MASK,
+			 APLIC_TARGET_GUEST_IDX_SHIFT);
```
gagachang, line 38: According to `APLIC_TARGET_GUEST_IDX_MASK`, the mask operation should be done after shift.

```c
	val |= SHIFT_U32(guest_idx, APLIC_TARGET_GUEST_IDX_SHIFT) & APLIC_TARGET_GUEST_IDX_MASK;
```

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,155 @@
...
+	int rc = -1;
+	paddr_t reg_addr = 0;
+	size_t reg_size = 0;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (nodeoff < 0 || !aplic || !fdt)
+		return TEE_ERROR_BAD_PARAMETERS;
+	memset(aplic, 0, sizeof(*aplic));
+
+	rc = fdt_get_reg_props_by_index(fdt, nodeoff, 0, &reg_addr, &reg_size);
```
gagachang, line 41: Prefer using `fdt_reg_info()`

core/drivers/imsic.c
```diff
@@ -0,0 +1,383 @@
...
+	paddr_t reg_addr = 0;
+	size_t reg_size = 0;
+	int i = 0;
+	int rc = -1;
+	int len = 0;
+
+	if (nodeoff < 0 || !imsic || !fdt)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	rc = fdt_get_reg_props_by_index(fdt, nodeoff, 0, &reg_addr, &reg_size);
```
gagachang, line 252: Prefer using `fdt_reg_info()`

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,155 @@
...
+	if (rc < 0 || !reg_addr || !reg_size)
```
gagachang, line 42: `!reg_addr` implies that base address can not be 0.
I am not sure if it is correct or not.

HuangBorong: Well, I think if `reg_addr` can be 0, it implies a corner case where the APLIC MMIO address starts at 0x0000_0000. However, the address 0x0000_0000 is sometimes reserved or designated for debug address space, rather than being allocated to peripherals.

gagachang: Agreed, thanks.

core/drivers/aplic_direct.c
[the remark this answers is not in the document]
etienne-lms: Good catch. I think the instruction block in `gic_op_add()` that acts on GICR should mask interrupt for executed sequence.

HuangBorong:
@gagachang Thank you for your review! I have fixed the issues you mentioned above in commit 5d901b7.

gagachang:
> @gagachang Thank you for your review! I have fixed the issues you mentioned above in commit [5d901b7](https://github.com/OP-TEE/optee_os/commit/5d901b7e7b9bf50127f935eeb03a6cbdf6d1b7f5).

You're welcome, did you test secure interrupt ?

core/drivers/imsic.c
```diff
@@ -0,0 +1,383 @@
...
+	if (val && len > 0)
+		imsic->guest_index_bits = fdt32_to_cpu(*val);
+	else
+		imsic->guest_index_bits = 0;
+
+	val = fdt_getprop(fdt, nodeoff, "riscv,hart-index-bits", &len);
+	if (val && len > 0)
+		imsic->hart_index_bits = fdt32_to_cpu(*val);
+	else
+		imsic->hart_index_bits = IMSIC_HART_INDEX_BITS;
```
HuangBorong, line 285: @gagachang  If the IMSIC node is missing the `hart_index_bits` property, what about calculating it from `CFG_TEE_CORE_NB_CORE` ? Are you aware of any macros or helper functions in OP-TEE that can calculate the number of bits a variable holds?

HuangBorong:
> You're welcome, did you test secure interrupt ?

Hi @gagachang,

So far, I’ve tested secure interrupts in APLIC direct mode. You can find my code in my repository [https://github.com/HuangBorong/riscv-optee-dev-env/tree/test-aplic-direct](url) under the branch `test-aplic-direct`, which is dedicated to testing APLIC direct mode in OP-TEE.

To test the secure interrupt, I’ve done the following:
1. Added a simple timer in QEMU virt: It can be set to a specific duration and trigger an external interrupt.
2. Enabled M-mode interrupts after OpenSBI completes initialization and before returning to OP-TEE.
3. Since there is no Linux, S-mode interrupts are enabled when OP-TEE completes its initialization, and OP-TEE enters an infinite loop after initialization.
4. Added an `itr_handler` for the timer.
5. Added the UART RX you provided earlier.

However, I’ve encountered an issue:

While OP-TEE successfully receives interrupts from the timer, it fails to receive interrupts from UART RX. However, the console is able to capture the characters I enter. I’m unsure whether this is due to a misconfiguration or related to other underlying problems.

core/drivers/imsic.c
gagachang: > what about calculating it from `CFG_TEE_CORE_NB_CORE` ?

I think it is OK to calculate it from `CFG_TEE_CORE_NB_CORE`.

> Are you aware of any macros or helper functions in OP-TEE that can calculate the number of bits a variable holds?

Maybe you can use `32 - __builtin_clz(CFG_TEE_CORE_NB_CORE - 1)`

HuangBorong: Thanks!

gagachang:
> While OP-TEE successfully receives interrupts from the timer, it fails to receive interrupts from UART RX. However, the console is able to capture the characters I enter. I’m unsure whether this is due to a misconfiguration or related to other underlying problems.

Sorry I forgot to give you some configurations. You need to enable `CFG_16550_UART`.
And the UART IP interrupt must be enabled. You need to add code to configure `UART_IER` register bit 0 in `ns16550_init()`.

HuangBorong:
> Sorry I forgot to give you some configurations. You need to enable `CFG_16550_UART`. 
And the UART IP interrupt must be enabled. You need to add code to configure `UART_IER` register bit 0 in `ns16550_init()`.

Thanks! I can receive UART interrupt now.

gagachang:
> Thanks! I can receive UART interrupt now.

Great! It seems the APLIC direct mode works well.

core/arch/riscv/plat-virt/conf.mk
```diff
@@ -28,7 +28,10 @@ $(call force,CFG_BOOT_SYNC_CPU,n)
 
 $(call force,CFG_RISCV_M_MODE,n)
 $(call force,CFG_RISCV_S_MODE,y)
-$(call force,CFG_RISCV_PLIC,y)
+$(call force,CFG_RISCV_PLIC,n)
+$(call force,CFG_RISCV_APLIC,n)
+$(call force,CFG_RISCV_APLIC_MSI,y)
+$(call force,CFG_RISCV_IMSIC,y)
```
gagachang, line 34: These configurations should be build time configurable.
I have proposed https://github.com/OP-TEE/optee_os/pull/7278 for `CFG_RISCV_PLIC`.
Please consider the following code:
```diff
CFG_RISCV_PLIC ?= y
+ CFG_RISCV_APLIC ?= n
+ CFG_RISCV_APLIC_MSI ?= n
+ CFG_RISCV_IMSIC ?= n
```

Thus, the developers can select different interrupt controllers when building OP-TEE OS.

core/arch/riscv/plat-virt/main.c
```diff
@@ -16,6 +18,10 @@ register_ddr(DRAM_BASE, DRAM_SIZE);
 
 register_phys_mem_pgdir(MEM_AREA_IO_NSEC, UART0_BASE,
 			CORE_MMU_PGDIR_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, APLIC_BASE,
+			APLIC_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, IMSIC_BASE,
+			IMSIC_SIZE);
```
gagachang, line 24: Can apply configuration macros upon these two registrations.
```diff
+ #if defined(CFG_RISCV_APLIC) || defined(CFG_RISCV_APLIC_MSI)
register_phys_mem_pgdir(MEM_AREA_IO_SEC, APLIC_BASE,
			APLIC_SIZE);
+ #endif
+ #if defined(CFG_RISCV_APLIC_MSI) || defined(CFG_RISCV_IMSIC)
register_phys_mem_pgdir(MEM_AREA_IO_SEC, IMSIC_BASE,
			IMSIC_SIZE);
+#endif
```

core/arch/riscv/plat-virt/platform_config.h
```diff
@@ -32,6 +32,25 @@
...
+
+/* IMSIC */
+#ifndef IMSIC_BASE
+#define IMSIC_BASE		0x28000000
+#define IMSIC_SIZE		0x00004000
+#define IMSIC_NUM_IDS	0xFF
+#define IMSIC_GUEST_INDEX_BITS   0
+#define IMSIC_HART_INDEX_BITS 0x02
+#define IMSIC_GROUP_INDEX_BITS 0x00
+#define IMSIC_GROUP_INDEX_SHIFT 0x18
```
gagachang, line 51: Please format/align these macros.

core/arch/riscv/plat-virt/conf.mk
gagachang: The PR is just merged today. Please rebase.

gagachang:
@HuangBorong You can hide the resolved reviews so that the reviewers can know which reviews were resolved.
BTW, do you have test environment or idea for IMSIC/MSI?
PCIe devices are most common devices inject MSI, but we don't have PCIe drivers in OP-TEE.
Maybe the easiest way is to triggering IPI via MSI ? E.g., OP-TEE hart0 writes OP-TEE hart1 IMSIC.

HuangBorong:
> BTW, do you have test environment or idea for IMSIC/MSI?

I haven't tested the IMSIC/MSI yet.

> PCIe devices are most common devices inject MSI, but we don't have PCIe drivers in OP-TEE.
Maybe the easiest way is to triggering IPI via MSI ? E.g., OP-TEE hart0 writes OP-TEE hart1 IMSIC.

Could APLIC MSI delivery mode test the IMSIC? The APLIC can convert the wire interrupt into an MSI write and send it to IMSIC.
In this version of the code, I have not implemented the IPI handler for the IMSIC.

core/arch/riscv/plat-virt/platform_config.h
HuangBorong: Thank you for your advice! The related issues have been fixed in commit ae8b1829.

gagachang:
> Could APLIC MSI delivery mode test the IMSIC? The APLIC can convert the wire interrupt into an MSI write and send it to IMSIC. In this version of the code, I have not implemented the IPI handler for the IMSIC.

Oh! You are right. We can configure secure timer or UART to use APLIC MSI delivery mode.
IPI is unnecessary, please ignore my comment.

core/arch/riscv/plat-virt/conf.mk
```diff
@@ -28,6 +28,9 @@ $(call force,CFG_BOOT_SYNC_CPU,n)
 
 # Interrupt controller
 CFG_RISCV_PLIC ?= y
+CFG_RISCV_APLIC ?= n
+CFG_RISCV_APLIC_MSI ?= n
+CFG_RISCV_IMSIC ?= n
```
gagachang, line 33: `CFG_RISCV_PLIC` is incompatible with the other three configurations.
I think they must not be enabled simultaneously.
Recommend to add check on this case:

```
ifeq ($(CFG_RISCV_PLIC)-$(call cfg-one-enabled,CFG_RISCV_APLIC CFG_RISCV_APLIC_MSI CFG_RISCV_IMSIC),y-y)
$(error PLIC is incompatible with AIA)
endif
```

gagachang: The check can be added into `core/arch/riscv/riscv.mk`.

core/drivers/imsic.c
```diff
@@ -0,0 +1,384 @@
...
+
+void imsic_it_handle(void)
+{
+	struct imsic_data *imsic = &imsic_data;
+	uint32_t id = imsic_claim_interrupt();
+
+	if (id == IMSIC_IPI_ID)
+		DMSG("Interprocessor interrupt");
+
+	if (id > 1 && id <= imsic->num_ids)
```
gagachang, line 351: Prefer using macro instead of hardcode.
```c
if (id > IMSIC_IPI_ID && id <= imsic->num_ids)
```

core/drivers/imsic.c
```diff
@@ -0,0 +1,384 @@
...
+		imsic->guest_index_bits = fdt32_to_cpu(*val);
+	else
+		imsic->guest_index_bits = 0;
+
+	val = fdt_getprop(fdt, nodeoff, "riscv,hart-index-bits", &len);
+	if (val && len > 0)
+		imsic->hart_index_bits = fdt32_to_cpu(*val);
+	else
+		imsic->hart_index_bits = 32 -
+			__builtin_clz(CFG_TEE_CORE_NB_CORE - 1);
```
gagachang, line 286: Suggest to put calculation in one line:
```c
		imsic->hart_index_bits =
			32 - __builtin_clz(CFG_TEE_CORE_NB_CORE - 1);
```

HuangBorong: Thank you for your review! I have fixed the related issues in commit 11231c0.

HuangBorong:
Hi @gagachang,

I have tested the secure interrupt in APLIC MSI mode. You can find my code in my repository [https://github.com/HuangBorong/riscv-optee-dev-env/tree/test-aplic-msi](url) under `test-aplic-msi` branch.

I refactored the IMSIC interrupt handler in commit 99e2277. Here are the changes I made:

1. In the previous version, All external interrupts from APLIC shared share a single EIID, which was defined in the macro `APLIC_DEFAULT_EIID`. In this version, I have made the APLIC interrupt sources number be the EIID directly.
2. Added a pointer in `imsic_data` which points to `aplic_data`
3. When calling `interrupt_configure()`, the `imsic_op_add()` function now also calls `aplic_op_add()`.

gagachang:
> Hi @gagachang,
> 
> I have tested the secure interrupt in APLIC MSI mode. You can find my code in my repository [https://github.com/HuangBorong/riscv-optee-dev-env/tree/test-aplic-msi](url) under `test-aplic-msi` branch.
> 
> I refactored the IMSIC interrupt handler in commit [99e2277](https://github.com/OP-TEE/optee_os/commit/99e22770e4b7ac90507a1b4c1cdfc9baba321bed). Here are the changes I made:
> 
> 1. In the previous version, All external interrupts from APLIC shared share a single EIID, which was defined in the macro `APLIC_DEFAULT_EIID`. In this version, I have made the APLIC interrupt sources number be the EIID directly.
> 2. Added a pointer in `imsic_data` which points to `aplic_data`
> 3. When calling `interrupt_configure()`, the `imsic_op_add()` function now also calls `aplic_op_add()`.

Amazing work! Thanks.
One quick question is that the code assumes IMSIC and APLIC are always tied together.
A platform can adopt IMSIC independently. There is no APLIC behind the IMSIC in this case.

Besides, I think the relationship between APLIC and IMSIC in your code should be reversed.
If IMSIC and APLIC are tied together, IMSIC is parent of APLIC.
You can check APLIC device node `msi-parent`, and code in [linux APLIC-MSI driver](https://github.com/torvalds/linux/blob/master/drivers/irqchip/irq-riscv-aplic-msi.c#L26).


IMO, the `itr_main_chip` should be APLIC not IMSIC.
We can introduce `parent` member into `struct itr_chip` which points to parent `struct itr_chip`.
Call `aplic_op_add()` will check if it has `parent` (IMSIC in our case) and call `imsic_op_add()` .
What do you think?

HuangBorong:
> One quick problem is that the code assumes IMSIC and APLIC are always tied together.
> A platform can adopt IMSIC independently. There is no APLIC behind the IMSIC in this case.

I have considered this scenario before. However, the QEMU virt only supports two AIA types: APLIC only or APLIC + IMSIC. Currently, if a platform only has the IMSIC, we can not emulate it on QEMU virt. Furthermore, the IMSIC only receives MSI writes as interrupt signals. If the platform can adopt IMSIC independently, all devices on this platform should be PCI devices or support MSI writes. However, some simple devices, like UART, only support wired interrupts (using a wire to connect to the interrupt controller and possibly utilizing a positive edge as the interrupt signal).

HuangBorong:
Hi @gagachang 
> Besides, I think the relationship between APLIC and IMSIC in your code should be reversed.
> If IMSIC and APLIC are tied together, IMSIC is parent of APLIC.
> You can check APLIC device node msi-parent, and code in [linux APLIC-MSI driver](https://github.com/torvalds/linux/blob/master/drivers/irqchip/irq-riscv-aplic-msi.c#L26).

I believe you are correct. The IMSIC should be the parent of the APLIC. 
In the Linux APLIC-MSI driver, when enabling(unmasking) an IRQ, it first enables it on the IMSIC and then on the APLIC. Conversely, when disabling(masking) an IRQ, it performs the reverse: it first disables the IRQ on the APLIC and then on the IMSIC.

> IMO, the `itr_main_chip` should be APLIC not IMSIC.

I believe the `itr_main_chip` should be IMSIC instead, as it should point to the parent (root) interrupt controller, especially if we plan to add multiple APLIC interrupt domains in the future.

> We can introduce `parent` member into `struct itr_chip` which points to parent `struct itr_chip`.

I believe the current `struct itr_chip` is acceptable. The `itr_main_chip` uses a pointer points to `imsic_data` and `imsic_data` uses a pointer points to `aplic_data`. Currently, it only supports a single S-level APLIC interrupt domain.

gagachang:
> I believe you are correct. The IMSIC should be the parent of the APLIC. In the Linux APLIC-MSI driver, when enabling(unmasking) an IRQ, it first enables it on the IMSIC and then on the APLIC. Conversely, when disabling(masking) an IRQ, it performs the reverse: it first disables the IRQ on the APLIC and then on the IMSIC.

That's correct.

> I believe the `itr_main_chip` should be IMSIC instead, as it should point to the parent (root) interrupt controller, especially if we plan to add multiple APLIC interrupt domains in the future.

Can you provide use case to add multiple APLIC interrupt domains into OP-TEE ?

> I believe the current `struct itr_chip` is acceptable. The `itr_main_chip` uses a pointer points to `imsic_data` and `imsic_data` uses a pointer points to `aplic_data`. Currently, it only supports a single S-level APLIC interrupt domain.

It's OK to me. At least please add some sanity checks:
```diff
-	interrupt_configure(&imsic->aplic->chip, it, type, prio);
+	if (imsic->aplic)
+		interrupt_configure(&imsic->aplic->chip, it, type, prio);
	imsic_it_disable(it);
	imsic_it_clear_pending(it);
```
So that the driver also works for IMSIC systems without APLIC.

HuangBorong:
> Can you provide use case to add multiple APLIC interrupt domains into OP-TEE ?

To be honest, I don't know. The AIA specification states that the APLIC can have multiple interrupt domains, so I guess it may be used in the future. However, if we want to support multiple APLICs (NUMA), there will be multiple `aplic_data` entries, one for each APLIC. This will need to be managed by a data structure such as a linked list.

gagachang:
> To be honest, I don't know. The AIA specification states that the APLIC can have multiple interrupt domains, so I guess it may be used in the future. However, if we want to support multiple APLICs (NUMA), there will be multiple `aplic_data` entries, one for each APLIC. This will need to be managed by a data structure such as a linked list.

OK. Let's put this use case aside and merge infrastructure with single APLIC domain first.

jforissier:
@HuangBorong please fix the checkpatch errors too.

HuangBorong:
> @HuangBorong please fix the checkpatch errors too.

The checkpatch errors are mainly due to a missing 'Signed-off-by' and some misspellings. I will squash the commit and fix these issues after the code review.

gagachang:
> Hi @gagachang , I fixed the issues we discussed earlier in commit [3530ddc](https://github.com/OP-TEE/optee_os/commit/3530ddc91fa33a1cda4fb253c621090848d690cd).

LGTM. Please squash the commits and fix the `CI/Code style` errors. Otherwise the PR can't be accepted.

gagachang:
> @HuangBorong , About the Supervisor top interrupt CSR (stopi), yes we definitely need it. A simple solution is to implement AIA version of `thread_native_interrupt_handler()`. For example:

Hi @HuangBorong , what's the status of Supervisor top interrupt CSR (stopi)?
Do you plan to add AIA version of `thread_native_interrupt_handler()` ?

HuangBorong:
> Hi @HuangBorong , what's the status of Supervisor top interrupt CSR (stopi)? 
Do you plan to add AIA version of `thread_native_interrupt_handler()` ?

When I tested the APLIC and IMSIC drivers, I find it's OK to use the old `thread_native_interrupt_handler()`. These PR I focus on the APLIC and IMSIC drivers. I will add AIA version of `thread_native_interrupt_handler()` and other parts in AIA spec in the future.

gagachang:
> When I tested the APLIC and IMSIC drivers, I find it's OK to use the old `thread_native_interrupt_handler()`. These PR I focus on the APLIC and IMSIC drivers. I will add AIA version of `thread_native_interrupt_handler()` and other parts in AIA spec in the future.

Sure. Let's discuss them in future PR. This PR should focus on drivers. Thanks!

core/drivers/aplic_priv.c
```diff
@@ -0,0 +1,150 @@
...
+
+			aplic->targets_mmode = imsic.targets_mmode;
+
+			return TEE_SUCCESS;
+		}
+	}
+	return TEE_ERROR_ITEM_NOT_FOUND;
+}
+
+TEE_Result aplic_init_from_device_tree(struct aplic_data *aplic)
```
gagachang, line 84: @HuangBorong, Another topic is how to differentiate secure/non-secure APLIC S-level interrupt domains in one device tree.
In general, OP-TEE is assigned a dedicated APLIC S-level interrupt domain, while Linux is assigned another APLIC S-level interrupt domain.
If OP-TEE and Linux share one DTB, they must probe its own interrupt domain only.

ARM has "secure-status", see https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/secure.txt.
RISC-V needs similar way. Do you have any idea?

HuangBorong: > @HuangBorong, Another topic is how to differentiate secure/non-secure APLIC S-level interrupt domains in one device tree. In general, OP-TEE is assigned a dedicated APLIC S-level interrupt domain, while Linux is assigned another APLIC S-level interrupt domain. If OP-TEE and Linux share one DTB, they must probe its own interrupt domain only.
> 
> ARM has "secure-status", see https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/secure.txt. RISC-V needs similar way. Do you have any idea?

I think the simplest solution is to use a dedicated DTB for OP-TEE and another DTB for Linux and OpenSBI. However, if OP-TEE and Linux share a single DTB, the current APLIC DTS documentation (_Linux/Documentation/devicetree/bindings/interrupt-controller/riscv,aplic.yaml_) does not provide a way to distinguish between secure and non-secure APLIC S-level interrupt domains. Perhaps we could propose adding a property in the APLIC DTS, similar to ARM, to indicate whether the APLIC interrupt domains are secure or non-secure.

gagachang: > I think the simplest solution is to use a dedicated DTB for OP-TEE and another DTB for Linux and OpenSBI.

Yeah, I have the same thought as you. OP-TEE has `CFG_EMBED_DTB` and  `CFG_EMBED_DTB_SOURCE_FILE` to embed a dedicated DTB into OP-TEE image. This might be simplest solution.

> Perhaps we could propose adding a property in the APLIC DTS, similar to ARM, to indicate whether the APLIC interrupt domains are secure or non-secure.

Since RISC-V's security model is "multiple supervisor domains" not just two secure/non-secure worlds, I don't think we want to use "secure-status". We need a more general property.

core/arch/riscv/plat-virt/conf.mk
HuangBorong: Hi @gagachang ,


> `CFG_RISCV_PLIC` is incompatible with the other three configurations. I think they must not be enabled simultaneously. Recommend to add check on this case:
> 
> ```
> ifeq ($(CFG_RISCV_PLIC)-$(call cfg-one-enabled,CFG_RISCV_APLIC CFG_RISCV_APLIC_MSI CFG_RISCV_IMSIC),y-y)
> $(error PLIC is incompatible with AIA)
> endif
> ```

I didn't include this when I squashed the commits. Maybe you could add a commit or submit a PR later? I think this code was written by you and, logically speaking, it does not belong to any of my three commits.

core/drivers/aplic_priv.c
HuangBorong: > Since RISC-V's security model is "multiple supervisor domains" not just two secure/non-secure worlds, I don't think we want to use "secure-status". We need a more general property.

Maybe we can propose adding a property in the APLIC DTS called `supervisor-domain-id`. Linux or OP-TEE can read this property and compare it with its own domain ID. If they match, then the APLIC interrupt domain can be used; otherwise, it will be skipped. 
Since this requires modifying the current APLIC DTS, I think it will need further discussion.

core/drivers/aplic_direct.c
```diff
@@ -0,0 +1,193 @@
...
+		if (res)
+			panic();
+	} else {
+		aplic_init_base_addr(aplic, aplic_base_pa);
+	}
+
+	aplic->chip.ops = &aplic_ops;
+
+	io_write32(aplic->aplic_base + APLIC_DOMAINCFG, APLIC_DOMAINCFG_IE);
+	aplic_init_per_hart();
```
gagachang, line 167: Suggest to call `aplic_init_per_hart()` first before enabling `APLIC_DOMAINCFG_IE`.
Per-hart IDC should be initialized before enabling APLIC global IE.

core/include/drivers/aplic.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#ifndef __DRIVERS_APLIC_H
+#define __DRIVERS_APLIC_H
+
+#include <types_ext.h>
+
+/*
+ * The aplic_init() function initializes the struct aplic_data which
+ * is then used by other functions. These function also initializes
```
gagachang, line 13: "These function" -> "This function".

core/include/drivers/imsic.h
```diff
@@ -0,0 +1,72 @@
...
+	uint32_t hart_index_bits;
+	uint32_t group_index_bits;
+	uint32_t group_index_shift;
+	struct itr_chip *aplic_chip;
+	struct itr_chip chip;
+};
+
+/*
+ * The imsic_init() function initializes the struct imsic_data which
+ * is then used by other functions. These function also initializes
```
gagachang, line 34: "These function" -> "This function".

core/drivers/aplic_msi.c
```diff
@@ -0,0 +1,160 @@
...
+	} else {
+		aplic_init_base_addr(aplic, aplic_base_pa);
+	}
+
+	aplic->chip.ops = &aplic_ops;
+
+	io_write32(aplic->aplic_base + APLIC_DOMAINCFG,
+		   APLIC_DOMAINCFG_IE | APLIC_DOMAINCFG_DM);
+
+	interrupt_main_init(&aplic_data.chip);
```
gagachang, line 147: Since in this scenario APLIC is not main interrupt controller,
can we just call `itr_chip_init()` here to avoid confusion ?

```diff
- interrupt_main_init(&aplic_data.chip);
+ if (itr_chip_init(&aplic_data.chip))
+     panic();
```

gagachang:
@HuangBorong With the above comments are addressed, please add my tag:
```
```

core/drivers/aplic_msi.c
HuangBorong: > Since in this scenario APLIC is not main interrupt controller,
> can we just call `itr_chip_init()` here to avoid confusion ?

I think `interrupt_main_init(&aplic_data.chip);` is OK for the following reasons:

In the previous version of `imsic_data` structure, I used a pointer `struct aplic_data *aplic` to point to the global variable `static struct aplic_data aplic_data __nex_bss;`. This variable is defined in `aplic_msi.c`, and I added a function `get_aplic_data()` to get its address. However, when I squashed commits, I noticed that this approach increased the coupling between the IMSIC and APLIC driver code. Furthermore, in IMSIC, we only need `aplic_data->chip` to call APLIC functions and don't care about other data members in `aplic_data`.

Therefore, in this version of the code, I directly use `struct itr_chip *aplic_chip` in the `imsic_data` structure. When the `boot_primary_init_intc()` function initializes the interrupt controller, `aplic_init()` will first call `interrupt_main_init(&aplic_data.chip);` to register its `struct itr_chip` address to `itr_main_chip`. Then, during the `imsic_init()` initialization, it will retrieve that address and save it to `struct aplic_data *aplic` and call `interrupt_main_init(&imsic->chip);` again to update `itr_main_chip`.

gagachang: Oh! Now I understand your approach.
It's OK to me.
Thanks for clarification.

core/include/drivers/imsic.h
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+#ifndef __DRIVERS_IMSIC_H
+#define __DRIVERS_IMSIC_H
+
+#include <kernel/interrupt.h>
+#include <tee_api_defines.h>
+#include <types_ext.h>
+
+struct imsic_data {
```
etienne-lms, line 19: For future readers, could you add an inline description comment like
```c
/*
 * struct imsic_data - IMISC interrupt controller
 * @imisc_base: IOmem base address
 * @size: IOmem address range size
 * ...
```
It would be nice `struct aplic_data` gets also described.
Can be addressed later. Not blocking for the series.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with below comments addressed or not.

Maybe shorten these 2 commit message header lines:
```patch
-drivers: implement Advanced Platform-Level Interrupt Controller (APLIC) driver
+drivers: add RiscV APLIC interrupt driver
```
```patch
-drivers: implement Incoming MSI Controller (IMSIC) driver
+drivers: add RiscV IMSIC interrupt driver
```

It may  be worth adding 2 CI build make commands:
```
_make PLATFORM=virt CFG_RISCV_PLIC=n CFG_RISCV_APLIC=y
_make PLATFORM=virt CFG_RISCV_PLIC=n CFG_RISCV_APLIC_MSI=y CFG_RISCV_IMSIC=y

HuangBorong:
Fix related issues and add CI build in commit af2d7b8.

PR #7240, opened 2025-01-24
Interrupt

etienne-lms:
* drivers: gic: allow detection type configuration
Allow consumers to configure SPI interrupts when the provided type is not `IRQ_TYPE_NONE`. The allowed value are then `IRQ_TYPE_EDGE_RISING` and `IRQ_TYPE_LEVEL_HIGH`, as per Arm GIC specification.

* core: interrupt: fix property count in dt_get_irq_type_prio()    
Fix the number of properties passed by `dt_get_irq_type_prio()` to `dt_get_irq` handler of the interrupt main chip.

* core: interrupt: enable/disable fallback to mask/unmask
Fallback to unmask (resp. mask) handlers when an interrupt controller does not implement its enable (resp. disable)
handler.

* ... and few other minor changes in *interrupt.c* and *interrupt.h*.


core/include/kernel/interrupt.h
```diff
@@ -365,6 +366,24 @@ TEE_Result interrupt_create_handler(struct itr_chip *itr_chip, size_t itr_num,
 				    itr_handler_t callback, void *priv,
 				    uint32_t flags,
 				    struct itr_handler **out_hdl);
```
GseoC, line 373: The comment states the use of `@itr_desc` that is not a parameter of the function.
The part where it is stated that the call configured the interrupt is a bit unclear. (dt_get_interrupt_by_index does not exist, maybe `interrupt_dt_get`_xxx?); 
Eventually I don't really get why this needs to be stubbed, can you explain what made you do this please?

core/include/kernel/interrupt.h
```diff
@@ -233,7 +232,10 @@ static inline void interrupt_unmask(struct itr_chip *chip, size_t itr_num)
  */
 static inline void interrupt_enable(struct itr_chip *chip, size_t itr_num)
 {
-	chip->ops->enable(chip, itr_num);
+	if (chip->ops->enable)
+		chip->ops->enable(chip, itr_num);
+	else
+		interrupt_mask(chip, itr_num);
```
GseoC, line 238: unmask then?

core/drivers/gic.c
```diff
@@ -679,9 +680,9 @@ static void gic_it_set_cpu_mask(struct gic_data *gd, size_t it,
 	target_shift = (it % NUM_TARGETS_PER_REG) * ITARGETSR_FIELD_BITS;
 	target &= ~(ITARGETSR_FIELD_MASK << target_shift);
 	target |= cpu_mask << target_shift;
-	DMSG("cpu_mask: writing 0x%x to 0x%" PRIxVA, target, itargetsr);
+	DMSG("cpu_mask: writing %#"PRIx32" to 0x%" PRIxVA, target, itargetsr);
```
GseoC, line 683: Nitpick, might as well use # instead of 0x for adresses, no?

GseoC:
For commits:
core: interrupt: fix property count in dt_get_irq_type_prio()
core: interrupt: enhance registered controller chips validity

Is "core: interrupt: explicit dt_get_irq handler expects big endian data" necessary as it is inherent to the format of the DTB file?

core/include/kernel/interrupt.h
etienne-lms: drivers: gic: allow detection type configuration
> The comment states the use of @itr_desc that is not a parameter of the function.
> The part where it is stated that the call configured the interrupt is a bit unclear. (dt_get_interrupt_by_index does not exist, maybe interrupt_dt_get_xxx?);

Thanks. I'll fix the existing comment and the added one.

> Eventually I don't really get why this needs to be stubbed, can you explain what made you do this please?

Maybe this function does not need to be stubbed. I made the stub version in the header file to highlight in which context it is expected to be used. Maybe overkilling, indeed one could use `interrupt_create_handler()` and then `interrupt_configure()` to set the detection type (and maybe the prio).

If commit "core: interrupt: interrupt_create_handler() depends ion DT support" does not make sense, I can discard it and add instead a commit to fix the existing inline description comment.

core/include/kernel/interrupt.h
etienne-lms: :-) indeed!

core/drivers/gic.c
etienne-lms: ok, i'll do.

core/include/kernel/interrupt.h
GseoC: Yes, I think the proposition to fix the inline description fits better

etienne-lms (commented):
I've appended another fixup commit.

For info, commit "drivers: gic: allow detection type configuration" deserves tag
`Fixes: 14885eb1688b ("drivers: gic: register to dt_driver")` since commit 14885eb1688b implicitly claimed interrupt detection type was configured from the DT whereas it was not.

May I squash the fixup commits (and update the commit messages)?

jenswi-linaro:
Please go ahead.

etienne-lms:
I've squashed the fixup commits, updated the commit message and reorder some commits in the series for incremental consistency reasons.

core/drivers/gic.c
```diff
@@ -124,6 +125,13 @@
 #define GICD_SGIR_NSATT_SHIFT			15
 #define GICD_SGIR_CPU_TARGET_LIST_SHIFT		16
 
+/* GICD ICFGR bit fields */
+#define ICFGR_TYPE_EDGE		2
```
jenswi-linaro, line 129: Please add a  `GICD_` prefix since the name of the register(s) starts with `GICD_ICFGR`

core/drivers/gic.c
```diff
@@ -1062,10 +1108,14 @@ static TEE_Result dt_get_gic_chip_cb(struct dt_pargs *arg, void *priv_data,
 	 */
 	if (arg->args_count < 2)
 		return TEE_ERROR_GENERIC;
+
 	phandle_args[0] = cpu_to_fdt32(arg->args[0]);
 	phandle_args[1] = cpu_to_fdt32(arg->args[1]);
+	if (arg->args_count == 3)
```
jenswi-linaro, line 1115: What if `arg->args_count` is larger than `3`?

core/include/kernel/interrupt.h
```diff
@@ -233,7 +232,10 @@ static inline void interrupt_unmask(struct itr_chip *chip, size_t itr_num)
  */
 static inline void interrupt_enable(struct itr_chip *chip, size_t itr_num)
 {
-	chip->ops->enable(chip, itr_num);
+	if (chip->ops->enable)
+		chip->ops->enable(chip, itr_num);
+	else
+		interrupt_unmask(chip, itr_num);
```
jenswi-linaro, line 238: Wouldn't it make sense for the driver to supply the unmask function instead of the enable function if it doesn't have an enable function?

etienne-lms: Okay, i'll drop commit "core: interrupt: enable/disable fallback to mask/unmask".

jenswi-linaro:

etienne-lms:
I've removed commit "core: interrupt: enable/disable fallback to mask/unmask" and applied the review tag.

PR #7242, opened 2025-01-24
drivers: stm32_stgen: fix type in debug trace

GseoC:
Cast the result of an operation in stm32_stgen_pm_resume() to fix a compilation warning.

### DIFF
--- a/core/drivers/counter/stm32_stgen.c
+++ b/core/drivers/counter/stm32_stgen.c
@@ -154,7 +154,7 @@ static void stm32_stgen_pm_resume(void)
 
 	io_setbits32(stgen_d.base + STGENC_CNTCR, STGENC_CNTCR_EN);
 
-	DMSG("Time spent in low-power: %"PRIu64"ms",
+	DMSG("Time spent in low-power: %lld ms",
 	     (nb_pm_count_ticks * 1000) / clock_src_rate);
 }
 


jforissier:
> Perhaps `rtc_diff_calendar_tick()` should return a `uint64_t` to begin with? Anyways this looks good:
> 

...or what about using `%lld` since the argument is a signed long long int?

core/drivers/counter/stm32_stgen.c
```diff
@@ -155,7 +155,7 @@ static void stm32_stgen_pm_resume(void)
 	io_setbits32(stgen_d.base + STGENC_CNTCR, STGENC_CNTCR_EN);
 
 	DMSG("Time spent in low-power: %"PRIu64"ms",
-	     (nb_pm_count_ticks * 1000) / clock_src_rate);
+	     (uint64_t)(nb_pm_count_ticks * 1000) / clock_src_rate);
```
etienne-lms, line 158: `nb_pm_count_ticks` is of type `signed long long` so the `DMSG()` format specifier should be `lld` instead of `"PRIu64"`.

etienne-lms: Oops, I didn't see Jérôme's comment :-)

GseoC:
Ok, fair, comment addressed (tag not applied)

etienne-lms:

GseoC:
Tag applied, thanks

PR #7243, opened 2025-01-24
ltc: ctr: fix counter increment when LTC_FAST is defined

jforissier:
In CTR mode the counter needs to be incremented upon completion of each block. Unfortunately, if there is no crypto acceleration (accel_ctr_encrypt == NULL) and if LTC_FAST is defined, this does not happen. Add the missing call to fix the issue.

Reported-by: Jork Loeser <jork.loeser@microsoft.com>
Closes: https://lists.trustedfirmware.org/archives/list/op-tee@lists.trustedfirmware.org/thread/J4MMZPCM2MNKC2KWAXZUMTVEJP56U6OI/


jenswi-linaro:

PR #7247, opened 2025-01-29
ci: QEMUv8: check SPMC at S-EL1 and SPs

jenswi-linaro:
Add a check entry for SPMC at S-EL1 and secure partitions.


jforissier (commented):

jenswi-linaro:
Tag applied

PR #7249, opened 2025-01-30
Allow NIST curves on all ECDSA algorithms in libtomcrypt and libutee

etienne-lms:
According to GP TEE specification *"ECDSA algorithm identifiers should be tied to the size of the digest, not the key."*. Therefore allow use of any `TEE_ALG_ECDSA_SHA*` algorithm for any of the supported NIST curve.

The constraint removed by these changes are present only in libutee and libtomcrypt.

### DIFF
--- a/lib/libutee/tee_api_operations.c
+++ b/lib/libutee/tee_api_operations.c
@@ -75,21 +75,18 @@ TEE_Result TEE_AllocateOperation(TEE_OperationHandle *operation,
 			return TEE_ERROR_NOT_SUPPORTED;
 		break;
 
-	case TEE_ALG_ECDSA_SHA1:
 	case __OPTEE_ALG_ECDSA_P192:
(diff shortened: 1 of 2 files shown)


jforissier (commented):
For "ltc: allow NIST curves on all ECDSA algorithms":

lib/libutee/tee_api_operations.c
```diff
@@ -2713,6 +2730,7 @@ TEE_Result TEE_IsAlgorithmSupported(uint32_t alg, uint32_t element)
 	}
 
 	return TEE_ERROR_NOT_SUPPORTED;
+
```
jforissier, line 2733: Extra line not needed

jforissier (commented):
For "libutee: allow NIST curves on all ECDSA algorithms": one minor comment.

jenswi-linaro (commented):

PR #7250, opened 2025-01-30
libutee: add TEE_ALG_ECDSA_RAW

etienne-lms:
Add GP TEE Internal Core API extension algorithm ID `TEE_ALG_ECDSA_RAW` for ECDSA signature computation and verification operations when the digest algorithm is not specified. This algorithm ID is supported upon build configuration switch `CFG_CRYPTO_ECDSA_NOHASH`.

### DIFF
--- a/lib/libutee/include/utee_defines.h
+++ b/lib/libutee/include/utee_defines.h
@@ -78,6 +78,8 @@ static inline uint32_t __tee_alg_get_class(uint32_t algo)
 		return TEE_OPERATION_ASYMMETRIC_SIGNATURE;
 	if (algo == TEE_ALG_RSAES_PKCS1_OAEP_MGF1_MD5)
 		return TEE_OPERATION_ASYMMETRIC_CIPHER;
+	if (algo == TEE_ALG_ECDSA_RAW)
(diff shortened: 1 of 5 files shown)


jenswi-linaro (commented):
This feature is summarized as either "RAW" or "NOHASH", can we pick only one?

lib/libutee/include/utee_defines.h
```diff
@@ -78,6 +79,9 @@ static inline uint32_t __tee_alg_get_class(uint32_t algo)
 	if (algo == TEE_ALG_RSAES_PKCS1_OAEP_MGF1_MD5)
 		return TEE_OPERATION_ASYMMETRIC_CIPHER;
 
+	/* Implementation defined algorithm IDs must be specifically handled */
+	assert(((algo >> 28) & 0xF) != 0xF);
```
jenswi-linaro, line 85: Can we return `__TEE_OPERATION_UNDEFINED`  (with the value 0) or such instead of asserting?

etienne-lms:
Please keep alive. I'll work back on this topic soon...

PR #7251, opened 2025-01-30
libutee: TEE_ALG_[ECDSA|ECDH]_P* relate to API v1.1

etienne-lms:
Remove definition of TEE_ALG_ECDSA_P192/224/256/384/521 and TEE_ALG_ECDH_P192/224/256/384/521 when libutee is built without GP TEE Internal Core API v1.1 compatibility support.

These macros are now related to OP-TEE internal __OPTEE_ALG_ECDSA_P* and __OPTEE_ALG_ECDH_P* algorithm IDs that are more accurate.

This change will allow to later remove the relation between the ECDSA/ECDH algorithm and the EC key with when using API v1.3.1 while still supporting that relation when building for API v1.1.

Update pkcs11 TA and remoteproc TA accordingly.


lib/libutee/include/tee_api_defines.h
```diff
@@ -331,17 +331,23 @@
...
-#define TEE_ALG_ECDSA_P256			TEE_ALG_ECDSA_SHA256
-#define TEE_ALG_ECDSA_P384			TEE_ALG_ECDSA_SHA384
-#define TEE_ALG_ECDSA_P521			TEE_ALG_ECDSA_SHA512
-#define TEE_ALG_ECDH_P192		TEE_ALG_ECDH_DERIVE_SHARED_SECRET
-#define TEE_ALG_ECDH_P224		TEE_ALG_ECDH_DERIVE_SHARED_SECRET
-#define TEE_ALG_ECDH_P256		TEE_ALG_ECDH_DERIVE_SHARED_SECRET
-#define TEE_ALG_ECDH_P384		TEE_ALG_ECDH_DERIVE_SHARED_SECRET
-#define TEE_ALG_ECDH_P521		TEE_ALG_ECDH_DERIVE_SHARED_SECRET
+/*
+ * TEE_ALG_ECDSA_P* and TEE_ALG_ECDH_P* are deprecated on API v1.2 or later
```
jenswi-linaro, line 335: Why is it important to get rid of these?
 What is the harm of keeping TEE_ALG_ECDSA_P* and TEE_ALG_ECDH_P* defines regardless of __OPTEE_CORE_API_COMPAT_1_1?

See also the comment above at lines 26-30.

etienne-lms:
I've appended 5 commits to update `__tee_alg_get_digest_hash()` to handle `TEE_ALG_ECDSA_SHA*` IDs (and also ensure `__OPTEE_ALG_ECDSA_P*` IDs do not mis-return as hash ID whereas they should not.

lib/libutee/include/tee_api_defines.h
etienne-lms: I propose this to prevent using using the deprecated macros since they change of value (now here `__OPTEE_ALG_ECxx_Py`) hence of meaning. But I agree it's maybe overkillling.
I'm fine removing the condition for the IDs definition.

By this way, I think I should restore all `TEE_ALG_ECDH_P*` being aliases for the unique `TEE_ALG_ECDH_DERIVE_SHARED_SECRET` as it would still make sense.

core/tee/tee_svc_cryp.c
```diff
@@ -3353,6 +3353,7 @@ static TEE_Result get_hkdf_params(uint32_t algo, const TEE_Attribute *params,
 		*hash_id = TEE_ALG_SHA256;
 	} else {
 		*hash_id = TEE_ALG_GET_DIGEST_HASH(algo);
+		assert(*hash_id != TEE_MAIN_ALGO_UNDEFINED);
```
jenswi-linaro, line 3356: Can this ever happen due to invalid TA input?

etienne-lms: Fair, an assertion here is not the right instruction.

That said in one hand, in current implementation, only `TEE_ALG_HKDF` algo ends in calling `get_hkdf_params()` so I think instruction block starting at line 3355 looks like dead code. In the other hand, if the extracted hash ID is not valid (any unexpected value including 0), it will be rejected later when used (from `tee_alg_get_digest_size()` or like, or `crypto_hash_alloc_ctx()`). All in one, all assertions I added in commit "libutee: add __OPTEE_ALG_ECDSA_P* in __tee_alg_get_digest_hash()" seem useless. I'll crosscheck that and remove them if confirmed.

By the way, I was wondering whether `__tee_alg_get_digest_hash()` should be more strict and check the hash ID is a valid ID but for the reason detailed above, I don't think it's needed.

lib/libutee/include/tee_api_defines.h
```diff
@@ -331,12 +331,25 @@
...
-#define TEE_ALG_ECDSA_P521			TEE_ALG_ECDSA_SHA512
+/*
+ * TEE_ALG_ECDSA_P* are deprecated on API v1.2 or later and are replaced
+ * with TEE_ALG_ECDSA_SHA* and that have the same value but a different
+ * meaning (defined the related hash algorithm). Therefore assign ID values
+ * of __OPTEE_ALG_ECDSA_P* to the deprecated macros to avoid confusion
+ * while still supporting the algorithm for portability and backward
+ * compatibility purposes.
+ */
+#define TEE_ALG_ECDSA_P192			__OPTEE_ALG_ECDSA_P192
```
jenswi-linaro, line 342: The previous code tried to phase out the old values defined by __OPTEE_ALG_ECDSA_P* so if a TA is recompiled it will use the new values. The day when we finally say good riddance to __OPTEE_ALG_ECDSA_P* in OP-TEE core only the old TAs that never were recompiled will stop working. With this change, things become less future-compatible.

ta/pkcs11/src/processing_ec.c
```diff
@@ -455,19 +455,19 @@ enum pkcs11_rc pkcs2tee_algo_ecdsa(uint32_t *tee_id,
 	 */
 	switch (get_object_key_bit_size(obj)) {
 	case 192:
-		*tee_id = TEE_ALG_ECDSA_P192;
+		*tee_id = __OPTEE_ALG_ECDSA_P192;
```
jenswi-linaro, line 458: Why not `TEE_ALG_ECDSA_SHA1`?

lib/libutee/include/tee_api_defines.h
etienne-lms: I guess then I need to change management of all `TEE_ALG_ECDSA_Px` to allow any key size and change management of  `TEE_ALG_ECDSA_SHA*` to allow any hash size.

ta/pkcs11/src/processing_ec.c
etienne-lms: IIUC a 192bit key is not expected to enforce use of SHA1.

lib/libutee/include/tee_api_defines.h
jenswi-linaro: All the `__OPTEE_ALG_ECDSA_P*` are already translated into `TEE_ALG_ECDSA_SHA*` in OP-TEE core before it's used. Why do you need to change `TEE_ALG_ECDSA_SHA*` to allow any hash size?

ta/pkcs11/src/processing_ec.c
jenswi-linaro: I can't find anywhere that `TEE_ALG_ECDSA_SHA1` must enforce the use of SHA-1.
It might make sense for `TEE_ALG_ECDSA_SHA*` to check that the supplied digest length doesn't get truncated due to the limitations of the curve. I can't find that in the specification either. Does the PKCS#11 TA need to supply digests larger than the curve can handle without implicit truncation?

lib/libutee/include/tee_api_defines.h
etienne-lms: If `TEE_ALG_ECDSA_P*` is not expected to enforce the hash size, having `TEE_ALG_ECDSA_SHAx=TEE_ALG_ECDSA_Py` will not allow extracting the hash algo from `TEE_ALG_ECDSA_SHA*`, nor enforcing `TEE_ALG_ECDSA_Py` targets a y-bit key.
Or maybe we can explicitly mention that each `TEE_ALG_ECDSA_P*` enforces a specific a specific hash algo and does not put constraint on the effective key size.

ta/pkcs11/src/processing_ec.c
etienne-lms: In the GP TEE internal core API spec description of `TEE_AsymmetricSignDigest()` (ditto in description of `TEE_AsymmetricVerifyDigest()`):
> Where a hash algorithm is specified in the algorithm, digestLen SHALL be equal to the digest length of this
hash algorithm. For TEE_ALG_ED25519 and TEE_ALG_ED448, if the TEE_ATTR_EDDSA_PREHASH attribute
has a=1 , b=0, then the implementation SHALL accept a digestLen of 64, and MAY accept other values.

and later:
>• If digestLen is not equal to the hash size of the algorithm in non-XOF functions



ECDSA RFC6979 doc (https://www.ietf.org/rfc/rfc6979.txt) gives test vectors for various NIST P-xxx key sizes and each with various SHA* hashing schemes.

[NIST FIPS 186-4](https://csrc.nist.rip/external/nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf) referred by the GP TEE spec for `TEE_ALG_ECDSA_SHA*. (NIST 186-5, from 2023 states the same, adding XOR digests support):
https://csrc.nist.rip/external/nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf
>"It is recommended that the security strength associated with the bit length of *n* and the security
strength of the hash function be the same unless an agreement has been made between
participating entities to use a stronger hash function. When the length of the output of the hash
function is greater than the bit length of *n*, then the leftmost *n* bits of the hash function output
block **shall** be used in any calculation using the hash function output during the generation or
verification of a digital signature. A hash function that provides a lower security strength than
the security strength associated with the bit length of *n* ordinarily **should not** be used, since this
would reduce the security strength of the digital signature process to a level no greater than that
provided by the hash function.

Last, from the literature it is not that clear whether, for example, ECDSA P-256 implicitly relate to a SHA256 hash even if mentionned here and there.

From what I understand:
* Stronger (likely wider) hash algos are permitted and shall be truncated.
* Weaker (incl. smaller) hash algos are permitted even if strongly not recommended. We can maybe enforce that.

etienne-lms: The specs mentions *"any"* input size *"no longer than 1024 bits"* telling *"Internally, this will be truncated to the appropriate number of bits."* ([spec v3.0 section 2.3.6 ECDSA without hashing](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061189)).

I think it is right that a PKCS#11 token implementation puts extra constraints on operation like requiring ECDSA to operate on pre-hashed values that are at least the size of the key. I would propose to follow this scheme.
@vesajaaskelainen, may I ask your thoughts on this topic?

etienne-lms:
FYI, i'll update this series to ensure the key size is not weaker than the selected hash.
I also saw that drvcypt layer needs to be updated the same way: it puts the same constraints on key size/hash size, see [`core/drivers/crypto/crypto_api/acipher/ecc.c: algo_is_valid()`](https://github.com/OP-TEE/optee_os/blob/4.5.0/core/drivers/crypto/crypto_api/acipher/ecc.c#L83). 
I wonder if these same should rather be factorized in core crypto API (*core/crypto/crypto.c*) rather than from each of ltc/libmbedtls/drvcrypt.

jenswi-linaro:
> I wonder if these same should rather be factorized in core crypto API (_core/crypto/crypto.c_) rather than from each of ltc/libmbedtls/drvcrypt.

Yes, that makes sense.

PR #7253, opened 2025-01-30
Core: add platform abort handler + stm32 SERC usage

GseoC:
Implement a platform abort handler to handle use cases such as data aborts generated by peripherals on the bus.

These kinds of abort could be caused by platform-specific features, hence the weak function.

### DIFF
--- a/core/arch/arm/kernel/abort.c
+++ b/core/arch/arm/kernel/abort.c
@@ -22,6 +22,7 @@ enum fault_type {
 	FAULT_TYPE_USER_MODE_VFP,
 	FAULT_TYPE_PAGEABLE,
 	FAULT_TYPE_IGNORE,
+	FAULT_TYPE_EXTERNAL_ABORT,
 };
 
 #ifdef CFG_UNWIND
@@ -114,6 +115,10 @@ static __maybe_unused const char *fault_to_str(uint32_t abort_type,
 		return " (write permission fault)";
 	case CORE_MMU_FAULT_TAG_CHECK:
 		return " (tag check fault)";
+	case CORE_MMU_FAULT_SYNC_EXTERNAL:
+		return " (Synchronous external abort)";
+	case CORE_MMU_FAULT_ASYNC_EXTERNAL:
+		return " (Asynchronous external abort)";
 	default:
 		return "";
 	}
@@ -525,10 +530,12 @@ static enum fault_type get_fault_type(struct abort_info *ai)
 		return FAULT_TYPE_PAGEABLE;
 
 	case CORE_MMU_FAULT_ASYNC_EXTERNAL:
+	case CORE_MMU_FAULT_SYNC_EXTERNAL:
 		if (!abort_is_user_exception(ai))
 			abort_print(ai);
-		DMSG("[abort] Ignoring async external abort!");
-		return FAULT_TYPE_IGNORE;
+		DMSG("[abort]%s", fault_to_str(ai->abort_type,
+					       ai->fault_descr));
+		return FAULT_TYPE_EXTERNAL_ABORT;
 
 	case CORE_MMU_FAULT_TAG_CHECK:
 		if (abort_is_user_exception(ai))
@@ -561,6 +568,12 @@ void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs)
 		save_abort_info_in_tsd(&ai);
 		vfp_disable();
 		handle_user_mode_panic(&ai);
+		break;
+	case FAULT_TYPE_EXTERNAL_ABORT:
+#ifdef CFG_EXTERNAL_ABORT_PLAT_HANDLER
+		/* Allow platform-specific handling */
+		plat_external_abort_handler(&ai);
+#endif
 		break;
 #ifdef CFG_WITH_VFP
 	case FAULT_TYPE_USER_MODE_VFP:
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -1452,7 +1452,7 @@ enum core_mmu_fault core_mmu_get_fault_type(uint32_t fault_descr)
 		return CORE_MMU_FAULT_ASYNC_EXTERNAL;
 	case 0x21: /* b100001 Alignment fault */
 		return CORE_MMU_FAULT_ALIGNMENT;
-	case 0x12: /* b100010 Debug event */
+	case 0x22: /* b100010 Debug event */
 		return CORE_MMU_FAULT_DEBUG_EVENT;
 	default:
 		break;
--- a/core/arch/arm/mm/core_mmu_lpae.c
+++ b/core/arch/arm/mm/core_mmu_lpae.c
@@ -1443,10 +1443,15 @@ enum core_mmu_fault core_mmu_get_fault_type(uint32_t fault_descr)
 	assert(fault_descr & FSR_LPAE);
 
 	switch (fault_descr & FSR_STATUS_MASK) {
-	case 0x21: /* b100001 Alignment fault */
-		return CORE_MMU_FAULT_ALIGNMENT;
+	case 0x10: /* b010000 Synchronous extern abort, not on table walk */
+	case 0x15: /* b010101 Synchronous extern abort, on table walk L1 */
+	case 0x16: /* b010110 Synchronous extern abort, on table walk L2 */
+	case 0x17: /* b010111 Synchronous extern abort, on table walk L3 */
+		return CORE_MMU_FAULT_SYNC_EXTERNAL;
 	case 0x11: /* b010001 Asynchronous extern abort (DFSR only) */
 		return CORE_MMU_FAULT_ASYNC_EXTERNAL;
+	case 0x21: /* b100001 Alignment fault */
+		return CORE_MMU_FAULT_ALIGNMENT;
 	case 0x12: /* b100010 Debug event */
 		return CORE_MMU_FAULT_DEBUG_EVENT;
 	default:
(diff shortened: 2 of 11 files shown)


core/include/kernel/abort.h
```diff
@@ -27,11 +27,15 @@ struct abort_info {
 
 /* Print abort info to the console */
 void abort_print(struct abort_info *ai);
+
```
jenswi-linaro, line 30: Spurious change?

core/arch/arm/kernel/abort.c
```diff
@@ -555,6 +560,8 @@ void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs)
 
 	switch (get_fault_type(&ai)) {
 	case FAULT_TYPE_IGNORE:
+		/* Allow platform-specific handling */
+		plat_abort_handler(regs);
```
jenswi-linaro, line 564: I'm not so keen on adding hooks that the compiler can't optimize away if unused.
How about using a config variable so the function call can be optimized away if not needed?

core/include/kernel/abort.h
```diff
@@ -27,11 +27,15 @@ struct abort_info {
...
 /* Print abort info + stack dump to the console */
 void abort_print_error(struct abort_info *ai);
 
 void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs);
 
+/* Platform overload, should be implemented in platform code */
+void plat_abort_handler(struct thread_abort_regs *regs);
```
jenswi-linaro, line 37: Isn't it only external aborts we care about here? If so, how about `plat_external_abort_handler()`?
This function should probably take a `struct abort_info *ai` argument instead.

core/arch/arm/kernel/abort.c
```diff
@@ -555,6 +560,8 @@ void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs)
 
 	switch (get_fault_type(&ai)) {
 	case FAULT_TYPE_IGNORE:
```
jenswi-linaro, line 562: We should add a `FAULT_TYPE_EXTERNAL_ABORT` if we need special treatment for external aborts.

core/include/kernel/abort.h
GseoC: Yes, it is about external aborts. Maybe we can even restrain the call to platform code on external aborts that are not table walks or table updates?
I'm fine with your suggestions

core/arch/arm/kernel/abort.c
GseoC: Ok for using a config switch, how about `CFG_EXTERNAL_ABORT_PLAT_HANDLER`,

core/arch/arm/kernel/abort.c
GseoC: Ok so that type of fault should be returned when encountering `CORE_MMU_FAULT_ASYNC_EXTERNAL` then?

I think we also miss the synchronous external abort cases, looking at the documentation for AARCH64: "Synchronous External abort, not on translation tablewalk or hardware update of translation table." is 0x10 in FSR.

So 0x10 and 0x12->0x17 are all external aborts. We miss these descriptions in `core/arch/arm/include/arm64.h`

core/include/kernel/abort.h
jenswi-linaro: OK to keep it.

GseoC:
Updated with comments addressed.

I'm having thoughts on the sharing of bindings between AARCH64 ESR_FSC_x and AARCH32 DFSR because it's supposed to to be architecturally mapped. Please let me know what you think of it as it is now.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -1444,6 +1444,8 @@ enum core_mmu_fault core_mmu_get_fault_type(uint32_t fault_descr)
 	assert(fault_descr & FSR_LPAE);
 
 	switch (fault_descr & FSR_STATUS_MASK) {
+	case 0x8: /* b01000 Asynchronous extern abort */
+		return CORE_MMU_FAULT_SYNC_EXTERNAL;
```
GseoC, line 1448: Typo here, should be CORE_MMU_FAULT_ASYNC_EXTERNAL

core/arch/arm/kernel/abort.c
```diff
@@ -525,10 +530,12 @@ static enum fault_type get_fault_type(struct abort_info *ai)
 		return FAULT_TYPE_PAGEABLE;
 
 	case CORE_MMU_FAULT_ASYNC_EXTERNAL:
+	case CORE_MMU_FAULT_SYNC_EXTERNAL:
 		if (!abort_is_user_exception(ai))
 			abort_print(ai);
-		DMSG("[abort] Ignoring async external abort!");
-		return FAULT_TYPE_IGNORE;
+		DMSG("[abort] %s!", fault_to_str(ai->abort_type,
```
etienne-lms, line 536: can remove the bang char `!`.
Can remove the leading space since since `fault_to_str()` prefixes a space char.

core/arch/arm/kernel/abort.c
```diff
@@ -546,6 +553,11 @@ static enum fault_type get_fault_type(struct abort_info *ai)
 	}
 }
 
+void __weak plat_abort_handler(struct abort_info *ai __unused)
+{
+	/* Do nothing */
+}
```
etienne-lms, line 559: Is this weak function needed?

core/include/kernel/abort.h
```diff
@@ -27,11 +27,15 @@ struct abort_info {
 
 /* Print abort info to the console */
 void abort_print(struct abort_info *ai);
+
 /* Print abort info + stack dump to the console */
 void abort_print_error(struct abort_info *ai);
 
 void abort_handler(uint32_t abort_type, struct thread_abort_regs *regs);
 
+/* Platform overload, should be implemented in platform code */
```
etienne-lms, line 36: Could you explicitly mention `CFG_EXTERNAL_ABORT_PLAT_HANDLER` must be enabled to have the platform handler be called.  I think the function label should explicitly mention `_external_abort_` if called only for external aborts.

core/arch/arm/kernel/abort.c
GseoC: For whatever reason, if the switch is enabled and the function is not defined? Given that the switch is intended for this purpose, I think  there is indeed no need for this weak function.

GseoC:
Removed "RFC" from the P-R title, comments addressed

core/arch/arm/kernel/abort.c
etienne-lms: If the switch is enabled, the platform should implement the handler. Having a build error is a nice way to trigger that it's missing IMO.

core/arch/arm/arm.mk
```diff
@@ -207,6 +207,10 @@ endif
 # this in platform config if different.
 CFG_MAX_CACHE_LINE_SHIFT ?= 6
 
+# CFG_EXTERNAL_ABORT_PLAT_HANDLER is used to implement platform-specific
+# handling of external abort using the plat_external_abort_handler() function.
```
etienne-lms, line 211: I suggest to replace "using" with "implementing".

core/arch/arm/plat-stm32mp2/main.c
```diff
@@ -172,3 +174,9 @@ bool stm32mp_allow_probe_shared_device(const void *fdt, int node)
 
 	return false;
 }
+
+void plat_external_abort_handler(struct abort_info *ai __unused)
+{
+	/* If fault is ignored, it could be due to a SERC event */
```
etienne-lms, line 180: Prefer `/* External abort may be due to SERC events */`.

core/arch/arm/mm/core_mmu_lpae.c
```diff
@@ -1444,6 +1444,8 @@ enum core_mmu_fault core_mmu_get_fault_type(uint32_t fault_descr)
 	assert(fault_descr & FSR_LPAE);
 
 	switch (fault_descr & FSR_STATUS_MASK) {
+	case 0x8: /* b01000 Asynchronous extern abort */
+		return CORE_MMU_FAULT_ASYNC_EXTERNAL;
```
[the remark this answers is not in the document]
GseoC: There is indeed an error on the value here, it is 0x10 for synchronous external abort.
Extract from ARM A32 documentation when TTCBR.EAE == 1:

0b000000 Address size fault in translation table base register.
0b000001 Address size fault, level 1.
0b000010 Address size fault, level 2.
0b000011 Address size fault, level 3.
0b000101 Translation fault, level 1
 0b000110 Translation fault, level 2.
0b000111 Translation fault, level 3.
0b001001 Access flag fault, level 1.
0b001010 Access flag fault, level 2.
0b001011 Access flag fault, level 3.
0b001101 Permission fault, level 1.
0b001110 Permission fault, level 2.
0b001111 Permission fault, level 3.
0b010000 Synchronous External abort, not on translation table walk.
0b010001 Asynchronous SError exception.
0b010101 Synchronous External abort on translation table walk,level 1.
0b010110 Synchronous External abort on translation table walk,level 2.
0b010111 Synchronous External abort on translation table walk,level 3.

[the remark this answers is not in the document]
GseoC: I will also add a patch to fix the debug event part

GseoC:
Comments addressed and added [core: fix debug event fault value for ARM32 with LPAE](https://github.com/OP-TEE/optee_os/pull/7253/commits/8ab635b066af37a944e53f96cedf67a077fb2589).

I can squash the commits if necessary

GseoC:
Fixups squashed and tag applied. 2 remaining commits

core/arch/arm/arm.mk
```diff
@@ -207,6 +207,11 @@ endif
 # this in platform config if different.
 CFG_MAX_CACHE_LINE_SHIFT ?= 6
 
+# CFG_EXTERNAL_ABORT_PLAT_HANDLER is used to implement platform-specific
+# handling of external abort implementing the plat_external_abort_handler()
+# function.
+CFG_EXTERNAL_ABORT_PLAT_HANDLER ?= n
```
jenswi-linaro, line 213: The should probably go into `mk/config.mk`.

GseoC: Sure, I didn't think of other architecture since I've only modified ARM abort.c.
I'll update the P-R

jenswi-linaro:
For "core: add platform-specific abort handler", please apply:
For the rest:

GseoC:
Tags applied, thank you!

etienne-lms (commented):

PR #7254, opened 2025-01-30
plat-rockchip: rk3588: Allow Passing of Devicetree

macromorgan:
In order to pass the devicetree from the first stage bootloader through Arm Trusted Firmware and into Optee, we need to increase the max allowable size of the FDT. 128KiB is enough today for my specific use case (an Indiedroid Nova), however if we add symbols and allow for future growth using a larger value would be prudent. Arm Trusted Firmware today has a hard limit with its current configuration of slightly above 512KiB, so set it to 384KiB to allow for future growth of also Arm Trusted Firmware code.

https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/34997

Additionally, U-Boot today uses an address for compressed kernels by default that is very close to the address used by Optee. As a result, if we use the default values from both projects the system will fail to boot when Optee is used. Update the TZDRAM_START and SHMEM_START to values used by RK3399 and PX30, which uses addresses not in use today by U-Boot.


edtubbs:
These changes hang the NanoPC-T6, including the TF-A `FDT_BUFFER_SIZE 0x60000` patch. Will need time to debug.

```
U-Boot SPL board init
U-Boot SPL 2017.09-ge4e124926e-230922 #lxh (Sep 25 2023 - 10:58:38), fwver: v1.13
Trying to boot from MMC2
No misc partition
Trying fit image at 0x4000 sector
## Verified-boot: 0
## Checking atf-1 0x00040000 ... sha256(c85fdb5713...) + OK
## Checking uboot 0x00200000 ... sha256(0b958ce0ab...) + OK
## Checking fdt 0x003385c0 ... sha256(9ff849197b...) + OK
## Checking atf-2 0xff100000 ... sha256(20d7ba5854...) + OK
## Checking optee 0x08400000 ... sha256(0e23999ce3...) + OK
Jumping to U-Boot(0x00200000) via ARM Trusted Firmware(0x00040000)
Total: 666.765/825.461 ms

NOTICE:  BL31: v2.12.0(release):v2.12.0
NOTICE:  BL31: Built : 19:46:40, Jan 23 2025
```

edtubbs:
Updating the overlay [rk3588-nanopi6-optee.dtso](https://github.com/OP-TEE/build/blob/master/nanopc-t6/rk3588-nanopi6-optee.dtso) to `0x30000000` works. I’ll submit a PR in `build` repo. NanoPC-T6 U-Boot [nanopi6-v2017.09](https://github.com/friendlyarm/uboot-rockchip/tree/nanopi6-v2017.09) lacks `CONFIG_OPTEE_LIB` so I can't use that feature.

macromorgan:
That seems to be a very old version of U-Boot. Any reason you're still using that? That's not the Rockchip BSP version is it?

edtubbs:
`TEE_OFFSET=0x08400000` is the default per [fit_args.sh#L42](https://github.com/friendlyarm/uboot-rockchip/blob/nanopi5-v2017.09/arch/arm/mach-rockchip/fit_args.sh#L42), consistent with OP-TEE builds for RK3588 from Rockchip. That said, I'm patching this script to `0x30000000` to align with your changes. It works, so I'll submit a PR to the `build` repo.

edtubbs (approved):

jforissier (commented):
For " plat-rockchip: rk3588: Increase FDT Max Size to 384KiB":
s/Optee/OP-TEE/ in the commit description

Then for both commits:

jforissier:
> ```
> Fixes https://github.com/OP-TEE/optee_os/commit/14754b93b1b78d8cc6715ad83aa709d1e21b1fad ("plat-rockchip: add support for Rockchip rk3588")
> 
> Signed-off-by: Chris Morgan <macromorgan@hotmail.com>
> ```

Please use:
```
Fixes: 14754b93b1b7 ("plat-rockchip: add support for Rockchip rk3588")
Signed-off-by: Chris Morgan <macromorgan@hotmail.com>
```

PR #7255, opened 2025-01-31
print warnings on insecure configurations

saschahauer:
The weak functions plat_rng_init() and tee_otp_get_hw_unique_key() must be overwritten in a secure product, but it's easy
to forget this either by misconfiguration or lack of knowledge. Add some warnings in these functions to avoid this pitfall

### DIFF
--- a/core/kernel/otp_stubs.c
+++ b/core/kernel/otp_stubs.c
@@ -14,14 +14,17 @@
  * Override these in your platform code to really fetch device-unique
  * bits from e-fuses or whatever.
  *
- * The default implementation just sets it to a constant.
+ * The default implementation just sets it to a constant and cannot be
+ * used in a secure environment.
  */
 
+#ifdef CFG_INSECURE
 __weak TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 {
 	memset(&hwkey->data[0], 0, sizeof(hwkey->data));
 	return TEE_SUCCESS;
 }
+#endif
 
 __weak int tee_otp_get_die_id(uint8_t *buffer, size_t len)
 {
--- a/core/tee/tee_cryp_utl.c
+++ b/core/tee/tee_cryp_utl.c
@@ -200,7 +200,14 @@ void __plat_rng_init(void)
 	}
 }
 
+/*
+ * Override this in your platform code. This default implementation only seeds
+ * the random number generator from an easily predictable timestamp value or a
+ * constant value. It is not suitable for a secure environment.
+ */
+#ifdef CFG_INSECURE
 void plat_rng_init(void) __weak __alias("__plat_rng_init");
+#endif
 
 static TEE_Result tee_cryp_init(void)
 {


core/kernel/otp_stubs.c
```diff
@@ -19,6 +19,8 @@
 
 __weak TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 {
+	EMSG("Using zero HUK");
```
jbech-linaro, line 22: I wonder if not IMSG would be more appropriate here (and for RNG), since strictly speaking this isn't an error as such.

Related, something I believe we touched upon in the past is whether we could make a "product" configuration. I.e., some kind of config that would give true (config) errors if things hasn't been take care of. Also, this type of config should disable various debug features as well by default. But, this is just a related side note.

jenswi-linaro: How about not defining this function even as __weak if `CFG_INSECURE` isn't defined?
That should get attention if needed.

saschahauer: > I wonder if not IMSG would be more appropriate here (and for RNG), since strictly speaking this isn't an error as such.

There already is a EMSG("Warning: seeding RNG with zeroes"), so I followed suit, but sure, I can use IMSG instead if that's preferred.

etienne-lms: I think using an `IMSG()` here is preferred: `IMSG("Warning: Using weak all zero HUK")`
I seems the error trace in `core/tee/tee_cryp_utl.c`should be updated to use `IMSG()` also, for consistency.
And likely seeding with PRNG with zeros should panic if `CFG_INSECURE=n`.

saschahauer: > How about not defining this function even as __weak if `CFG_INSECURE` isn't defined? That should get attention if needed.

Yes, that should attract enough attention ;)
Only for tee_otp_get_hw_unique_key() or would that that be suitable for plat_rng_init() as well?

jenswi-linaro: I'd say all functions that are critical to be replaced.

core/kernel/otp_stubs.c
```diff
@@ -17,11 +17,15 @@
  * The default implementation just sets it to a constant.
  */
 
+#ifdef CFG_INSECURE
 __weak TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 {
+	IMSG("Using weak all zero HUK");
```
jenswi-linaro, line 23: I don't think we need this. There is already a security warning at boot.

saschahauer: > I don't think we need this. There is already a security warning at boot.

Ok, dropped it.

jforissier (commented):

jenswi-linaro:

jforissier:
@saschahauer please fix the checkpatch warning and add the Reviewed-by tags, thanks!

core/kernel/otp_stubs.c
```diff
@@ -17,11 +17,13 @@
  * The default implementation just sets it to a constant.
  */
 
+#ifdef CFG_INSECURE
```
etienne-lms, line 20: Add an explicit statement like `This default implementation cannot be used when CFG_INSECURE is disabled.` in the inline comment above?
Ditto for `plat_rng_init()`.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`  with a suggestion.

etienne-lms:
@jforissier, could you trigger CI tests for this P-R?

jforissier:
> @jforissier, could you trigger CI tests for this P-R?

Sure, they're running now.

PR #7256, opened 2025-01-31
Update i.MX ELE support

saschahauer:
Implement plat_rng_init() for i.MX SoCs with ELE unit and enable ELE support for SoCs which have it.

### DIFF
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -228,7 +228,7 @@ CFG_TEE_CORE_NB_CORE ?= 2
 $(call force,CFG_NXP_SNVS,n)
 $(call force,CFG_IMX_OCOTP,n)
 CFG_IMX_MU ?= y
-CFG_IMX_ELE ?= n
+CFG_IMX_ELE ?= y
 else ifneq (,$(filter $(PLATFORM_FLAVOR),$(mx93-flavorlist)))
 $(call force,CFG_MX93,y)
 $(call force,CFG_ARM64_core,y)
@@ -241,7 +241,7 @@ $(call force,CFG_TZC380,n)
 $(call force,CFG_CRYPTO_DRIVER,n)
 $(call force,CFG_NXP_CAAM,n)
 CFG_IMX_MU ?= y
-CFG_IMX_ELE ?= n
+CFG_IMX_ELE ?= y
 else ifneq (,$(filter $(PLATFORM_FLAVOR),$(mx95-flavorlist)))
 $(call force,CFG_MX95,y)
 $(call force,CFG_ARM64_core,y)
@@ -263,7 +263,7 @@ $(call force,CFG_IMX_OCOTP,n)
 $(call force,CFG_TZC380,n)
 $(call force,CFG_NXP_CAAM,n)
 CFG_IMX_MU ?= y
-CFG_IMX_ELE ?= n
+CFG_IMX_ELE ?= y
 else
 $(error Unsupported PLATFORM_FLAVOR "$(PLATFORM_FLAVOR)")
 endif


sahilnxp:
Hi @saschahauer 

Did you test this pull request on all platforms (i.MX8ULP, i.MX91, i.MX93) impacted by this?

core/drivers/imx_ele.c
```diff
@@ -462,6 +464,7 @@ static TEE_Result imx_ele_rng_get_random(void *buffer, size_t size)
 	reg_pair_from_64((uint64_t)pa, &cmd.addr_msb, &cmd.addr_lsb);
 
 	cmd.size = (uint32_t)size;
+	update_crc(&msg);
```
saschahauer, line 467: This hunk shouldn't be here. Will remove

saschahauer:
> Hi @saschahauer
> 
> Did you test this pull request on all platforms (i.MX8ULP, i.MX91, i.MX93) impacted by this?

No, only on i.MX93. I don't have access to the other platforms

core/drivers/imx_ele.c
```diff
@@ -413,22 +416,70 @@ static TEE_Result imx_ele_rng_get_trng_state(void)
...
+ *
+ * @buffer: data output
+ * @size: RNG data size
+ */
+static TEE_Result imx_ele_rng_get_random(void *buffer, size_t size)
 {
-	TEE_Result res = TEE_ERROR_GENERIC;
-	uint64_t timeout = timeout_init_us(10 * 1000);
+	TEE_Result res;
+	paddr_t pa;
```
etienne-lms, line 434: `TEE_Result = TEE_<something>;`
`paddr_t pa = 0;`

core/drivers/imx_ele.c
```diff
@@ -413,22 +416,70 @@ static TEE_Result imx_ele_rng_get_trng_state(void)
...
+		pa = virt_to_phys(buf);
+	} else {
+		pa = (paddr_t)buffer;
+	}
+
+	reg_pair_from_64((uint64_t)pa, &cmd.addr_msb, &cmd.addr_lsb);
+
+	cmd.size = (uint32_t)size;
+
+	memcpy(msg.data.u8, &cmd, sizeof(cmd));
```
etienne-lms, line 468: `update_crc(&msg)` really not needed?
If that is the case, could you explicitly mention that in the commit message.

core/drivers/imx_ele.c
```diff
@@ -504,6 +506,26 @@ unsigned long plat_get_aslr_seed(void)
 	return aslr;
 }
 
+void plat_rng_init(void)
+{
+	TEE_Result res;
```
etienne-lms, line 511: `= TEE_<something>`

core/drivers/imx_ele.c
```diff
@@ -504,6 +506,26 @@ unsigned long plat_get_aslr_seed(void)
 	return aslr;
 }
 
+void plat_rng_init(void)
```
etienne-lms, line 509: Should consider whether or not `CFG_WITH_SOFTWARE_PRNG` is enabled?

core/drivers/imx_ele.c
saschahauer: This patch is based on [1] which removes this as well, so indeed it doesn't seem to be needed.
However, it also doesn't hurt, so I am considering just to keep it in.

[1] https://github.com/nxp-imx/imx-optee-os/commit/e7dadb871dd696d5cdbb404b07e07ced3411268b

sahilnxp:
@saschahauer Regarding enabling the ELE by default on every platform, Sorry, we cannot do that as of now because we have faced issues in past in which both Linux and OP-TEE tries to access the ELE and create issues which is mentioned in the commit message too.
```
There is only one MU to communicate with ELE, which cannot be
dedicated on OP-TEE side all the time. There may be ELE services
running on Linux side, which can cause conflict with OP-TEE.

```

ELE APIs are changed a lot in this time (just like CRC thing is not needed anymore, session is not needed while getting tee_otp_get_die_id() and more) also we are using TRUST MU dedicated for OP-TEE side which will remove the issue of MU sharing between Linux and OP-TEE.

We are in process of taking care of these changes and will upstream in coming weeks.

saschahauer:
> @saschahauer Regarding enabling the ELE by default on every platform, Sorry, we cannot do that as of now because we have faced issues in past in which both Linux and OP-TEE tries to access the ELE and create issues which is mentioned in the commit message too.
> 
> ```
> There is only one MU to communicate with ELE, which cannot be
> dedicated on OP-TEE side all the time. There may be ELE services
> running on Linux side, which can cause conflict with OP-TEE.

I am aware of this and mentioned it in the commit message:

> However, the ELE is currently only used for implementation of
> plat_rng_init() and tee_otp_get_hw_unique_key(). The former is used only
> during initialisation and the latter might be used during runtime, but
> the ELE is actually only accessed once and the HUK is then cached for
> later calls.

To put it short: The ELE is currently used during OP-TEE initialisation when no Linux is running.
Linux can have the ELE all for itself when running.

PR #7258, opened 2025-02-01
plat-k3: Initial support for AM62Lx SoC

r-vignesh:
AM62Lx newest among on the K3 class of SoCs designed to be low footprint system where DDR can be as small as 128M. Hence, move the DDR location to the beginning of DDR right after TF-A.

Disable TI SCI, secure boot info and HW unique ID support for now, they will be incrementally at later point in time as the underlying communication layer is different than AM62x.


core/arch/arm/plat-k3/conf.mk
```diff
@@ -2,8 +2,13 @@ CFG_WITH_STATS ?= y
 CFG_CRYPTO_WITH_CE ?= y
 CFG_CONSOLE_UART ?= 0
 
+ifeq (,$(filter ${PLATFORM_FLAVOR},am62lx))
 CFG_TZDRAM_START ?= 0x9e800000
 CFG_TZDRAM_SIZE ?= 0x01400000 # 20MB
+else
+CFG_TZDRAM_START ?= 0x80200000
+CFG_TZDRAM_SIZE ?= 0x00400000 # 20MB
+endif
```
jforissier, line 11: `filter` is not needed, and better flip the test around. How about:
```
ifeq ($(PLATFORM_FLAVOR),am62lx)
...
else
...
endif
```

jforissier (commented):
One suggestion below. In any case:

core/arch/arm/plat-k3/main.c
```diff
@@ -105,3 +106,4 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 
 	return TEE_SUCCESS;
 }
+#endif
```
DhruvaG2000, line 109: Nit: Can we pleaser end the #endif with ``//PLATFORM_FLAVOR_am62lx`` ?
Also, while at it please update all files copyright dates.

Thanks,
Reviewed-by: Dhruva Gole <d-gole@ti.com>

DhruvaG2000 (changes requested):

core/arch/arm/plat-k3/main.c
jforissier: Nit in the nit ;-) prefer C-style comments `/* PLATFORM_FLAVOR_am62lx */`

r-vignesh:
@jforissier  Thanks for the reviews! I see CI check fails for make (mulit-platform). But the same build passes locally as well as on CI on my fork (Although QEMU tests fail there) :  https://github.com/r-vignesh/optee_os/actions/runs/13097607768 . 

Any idea what I am I missing?

jforissier:
> @jforissier Thanks for the reviews! I see CI check fails for make (mulit-platform). But the same build passes locally as well as on CI on my fork (Although QEMU tests fail there) : https://github.com/r-vignesh/optee_os/actions/runs/13097607768 .
> 
> Any idea what I am I missing?

I see nothing obvious so I have restarted the failed job in case it would have been run on stale code.

jforissier:
> > @jforissier Thanks for the reviews! I see CI check fails for make (mulit-platform). But the same build passes locally as well as on CI on my fork (Although QEMU tests fail there) : https://github.com/r-vignesh/optee_os/actions/runs/13097607768 .
> > Any idea what I am I missing?
> 
> I see nothing obvious so I have restarted the failed job in case it would have been run on stale code.

No I see it. Your working branch is way, way behind master (424 commits to be precise). Please rebase onto master and you should be able to reproduce the build error. Thanks!

.github/workflows/ci.yml
```diff
@@ -212,6 +212,8 @@ jobs:
           _make PLATFORM=k3-am64x CFG_ARM64_core=y
           _make PLATFORM=k3-am62x
           _make PLATFORM=k3-am62x CFG_ARM64_core=y
+          _make PLATFORM=k3-am62lx CFG_WITH_SOFTWARE_PRNG=y
+          _make PLATFORM=k3-am62lx CFG_ARM64_core=y CFG_WITH_SOFTWARE_PRNG=y
```
jforissier, line 216: If the build fails when software PRNG is disabled then perhaps `am62lx` should `$(call force,CFG_WITH_SOFTWARE_PRNG=y)`?

r-vignesh: Agreed. This platform will gain TRNG support in future. But perhaps till then its better to force SOFTWARE_PRNG. will push an update accordingly.

DhruvaG2000 (approved):

PR #7260, opened 2025-02-03
ci: add arm64 job

jforissier:
Add a CI job that builds and runs QEMUv8 tests (make check) on an arm64 host.


jforissier:
Note, we now have quite a lot of duplicated code in `ci.yml`, but unfortunately GitHub Actions do not support YAML anchors and aliases. We could probably use `matrix` to factor code used by all the "make check" jobs. That is left for another PR ;)

jenswi-linaro (commented):

PR #7262, opened 2025-02-05
core: arm: enable native interrupts before init calls

etienne-lms:
Enable native interrupts on Arm architecture before initcalls handlers are called. This change allows drivers to leverage interrupts during OP-TEE core initialization, as for example detecting a firewall access violation when it happens. Before this change, OP-TEE needed to complete its initialization and return the secure monitor for secure interrupts to be handled.

The feature is embedded upon config switch CFG_UNMASK_NATIVE_INTR_AT_BOOT and applies only when CFG_NS_VIRTUALIZATION is disabled. When CFG_NS_VIRTUALIZATION is enabled, early/service/driver initcalls already execute with native interrupts unmasked, from a standard call entry when the non-secure guest is registered.

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1069,8 +1069,14 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
 
 	boot_primary_init_intc();
 	init_vfp_nsec();
-	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION))
+	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION)) {
+		/* Unmask native interrupts during driver initcalls */
+		thread_set_exceptions(thread_get_exceptions() &
+				      ~THREAD_EXCP_NATIVE_INTR);
 		init_tee_runtime();
+		thread_set_exceptions(thread_get_exceptions() |
+				      THREAD_EXCP_NATIVE_INTR);
+	}
 }
 
 /*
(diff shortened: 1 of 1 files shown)


core/arch/arm/kernel/boot.c
```diff
@@ -1069,8 +1069,15 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
 
 	boot_primary_init_intc();
 	init_vfp_nsec();
-	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION))
+	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION)) {
+		if (IS_ENABLED(CFG_UNMASK_NATIVE_INTR_AT_BOOT)) {
```
jenswi-linaro, line 1073: Why not always enable this?

core/arch/arm/kernel/boot.c
```diff
@@ -1069,8 +1069,15 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
...
+			/* Unmask native interrupts during driver initcalls */
+			thread_set_exceptions(thread_get_exceptions() &
+					      ~THREAD_EXCP_NATIVE_INTR);
+		}
+
 		init_tee_runtime();
+	}
```
jenswi-linaro, line 1080: We should probably disable the interrupts again since `entry_a64.S` doesn't expect interrupts to be enabled when `boot_init_primary_late()` returns.

core/arch/arm/kernel/boot.c
etienne-lms: Fair.

core/arch/arm/kernel/boot.c
etienne-lms: Right. Then i'll unmask them back in `boot_init_primary_final()` for `call_driver_initcalls()` and `call_finalcalls()` execution.

core/arch/arm/kernel/boot.c
```diff
@@ -1082,10 +1088,21 @@ void __weak boot_init_primary_final(void)
...
-	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION))
+	if (!IS_ENABLED(CFG_NS_VIRTUALIZATION)) {
+		/* Unmask native interrupts during init/finalcalls */
+		thread_set_exceptions(thread_get_exceptions() &
+				      ~THREAD_EXCP_NATIVE_INTR);
+
 		call_driver_initcalls();
+	}
+
 	call_finalcalls();
```
jenswi-linaro, line 1099: `CFG_NS_VIRTUALIZATION` may also benefit from unmasked native interrupts during finalcalls.

etienne-lms: Right, I missed that. Thanks, i'll update.

jenswi-linaro (commented):

etienne-lms:
Many thanks @jenswi-linaro for the feedback on this change.

etienne-lms:
I forgot to say that I've changed the commit message body. Please cross-check that it still matches change description expectations.

jenswi-linaro:
Looks good. :-)

PR #7263, opened 2025-02-05
drivers: firewall: stm32_etzpc: remove header file

etienne-lms:
Remove stm32_etzpc.h header file that is not required since the declared and defined resources are used internally in stm32_etzpc.c

By the way, also remove inclusion of stm32mp15-etzpc.h DT bindings header file from stm32_rng.c where it is not needed.


GseoC:

etienne-lms:
Thanks. Tag applied.

PR #7265, opened 2025-02-06
plat-versal2: disable ARM CE v8.2

Akshay-Belsare:
Disable ARM CE configs for SHA-512 and SHA3 as those are not available on the SoC.
Ensure that the ARM CE configuration can be overridden at runtime while the default value remains enabled.

### DIFF
--- a/core/arch/arm/plat-versal2/conf.mk
+++ b/core/arch/arm/plat-versal2/conf.mk
@@ -34,8 +34,7 @@ CFG_WITH_STATS		?= y
 CFG_ARM64_core		?= y
 
 # Enable ARM Crypto Extensions(CE)
-$(call force,CFG_CRYPTO_WITH_CE,y)
-$(call force,CFG_CRYPTO_WITH_CE82,y)
+CFG_CRYPTO_WITH_CE ?= y
 
 # Define the number of cores per cluster used in calculating core position.
 # The cluster number is shifted by this value and added to the core ID,


core/arch/arm/plat-versal2/conf.mk
```diff
@@ -34,7 +38,7 @@ CFG_WITH_STATS		?= y
 CFG_ARM64_core		?= y
 
 # Enable ARM Crypto Extensions(CE)
-$(call force,CFG_CRYPTO_WITH_CE,y)
+CFG_CRYPTO_WITH_CE ?= y
```
jenswi-linaro, line 41: This will not work as intended. If `CFG_CRYPTO_WITH_CE82` is `y`, then `CFG_CRYPTO_WITH_CE` must also be `y`.

jenswi-linaro:
> Ensure that the ARM CE configuration can be overridden at runtime while the default value remains enabled.

What do you mean by that? The CE configuration is a compile-time option.

So of the 8.2 crypto extensions you support SM3 and SM4 but have them implicitly disabled via
```
$(call force,CFG_CRYPTO_SM3,n)
$(call force,CFG_CRYPTO_SM4,n)
```
Am I missing something?

jenswi-linaro:
Please squash the commits and update the commit message.

michalsimek:
Works for me
Acked-by: Michal Simek <michal.simek@amd.com>

jenswi-linaro:

Akshay-Belsare:
@jforissier the change is ready to be merged.

PR #7267, opened 2025-02-07
core: interrupt: rename .add handler to .configure + minor inline comment changes

etienne-lms:
Rename field `add` of `struct itr_ops` to `configure` for consistency since that handler is used the configure the interrupt. Update existing interrupt drivers accordingly.

That `configure` handler is not required for interrupt providers which consumer only use the DT to get and configure their interrupts (with `interrupt_dt_get_by_*()` and `interrupt_create_handler()`). Therefore change `itr_chip_is_valid()` to not enforce its support but add back that constraint for the interrupt main controller.

Plus few minor changes in inline comments in interrupt.h and interrupt.c.

### DIFF
--- a/core/include/kernel/interrupt.h
+++ b/core/include/kernel/interrupt.h
@@ -305,7 +305,7 @@ static inline void interrupt_raise_sgi(struct itr_chip *chip, size_t itr_num,
 /*
  * interrupt_set_affinity() - Set CPU affinity for a controller interrupt
  * @chip	Interrupt controller
- * @itr_num	Interrupt number to raise
+ * @itr_num	Interrupt number
  * @cpu_mask	Mask of the CPUs targeted by the interrupt
  */
 static inline void interrupt_set_affinity(struct itr_chip *chip, size_t itr_num,
--- a/core/include/kernel/interrupt.h
+++ b/core/include/kernel/interrupt.h
@@ -59,9 +59,15 @@ struct itr_chip {
  * @raise_sgi	Raise a SGI or NULL if not applicable to that controller
  * @set_affinity Set interrupt/cpu affinity or NULL if not applicable
  *
- * Handlers @enable, @disable, @mask, @unmask and @configure are mandated.
- * Handlers @mask and @unmask have unpaged memory constraints.
- * See itr_chip_is_valid().
+ * Handlers @enable, @disable, @mask and @unmask are mandated. Handlers
+ * @mask and @unmask have unpaged memory constraints. These requirements are
+ * verified by itr_chip_init() and itr_chip_dt_only_init().
+ *
+ * Handler @configure is needed for interrupt providers which do not rely on
+ * DT for consumers interrupt configuration, that is interrupt consumers using
+ * interrupt_configure() or friends (interrupt_add_handler(),
+ * interrupt_add_configure_handler(), interrupt_add_handler_with_chip(),
+ * etc...).
  */
 struct itr_ops {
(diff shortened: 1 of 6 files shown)


core/include/kernel/interrupt.h
```diff
@@ -132,8 +138,7 @@ static inline bool itr_chip_is_valid(struct itr_chip *chip)
 	       is_unpaged((void *)chip->ops) &&
 	       chip->ops->mask && is_unpaged(chip->ops->mask) &&
 	       chip->ops->unmask && is_unpaged(chip->ops->unmask) &&
-	       chip->ops->enable && chip->ops->disable &&
-	       chip->ops->add;
```
jenswi-linaro, line 136: I'd rather have a separate version of `itr_chip_is_valid()` for the DT use case, for clarity and consistent checking.

etienne-lms:
I've rebased the series to fix the merge conflict.
I've appended a commit to address https://github.com/OP-TEE/optee_os/pull/7267#discussion_r1946099787.

core/kernel/interrupt.c
```diff
@@ -105,7 +133,12 @@ void interrupt_call_handlers(struct itr_chip *chip, size_t itr_num)
 TEE_Result interrupt_configure(struct itr_chip *chip, size_t itr_num,
 			       uint32_t type, uint32_t prio)
 {
-	chip->ops->add(chip, itr_num, type, prio);
+	if (!chip->ops->configure) {
+		EMSG("No configure handler in itr_chip %s", chip->name);
+		return TEE_ERROR_NOT_IMPLEMENTED;
```
jenswi-linaro, line 138: `add_configure_handler()` below should handle this error.

jenswi-linaro:

etienne-lms:
I've slightly modified the content of commits *"core: interrupt: rename .add handler to .configure"* and
*"core: interrupt: itr_chip may not require configure handler"* regarding the previously reviewed series to have each commit do only what's it's expecting to (renaming then removing a constraint.). Therefore I've not yet applied your review tag on those 2 commits.

Aside there content changed, the resulting diff, once both applied, is inline comments update (and spelling fixes):
```patch
  * Handlers @enable, @disable, @mask and @unmask are mandated. Handlers
- * @mask and @unmask have unpaged memory contrainsts. See itr_chip_is_valid().
+ * @mask and @unmask have unpaged memory constraints. These requirements are
+ * verified by itr_chip_init() and itr_chip_dt_only_init().
  *
- * Handler @configure is needed for interrupt provider which do not rely on
- * DT for consumer interrupt configuration, that is interrupt consumer using
+ * Handler @configure is needed for interrupt providers which do not rely on
+ * DT for consumers interrupt configuration, that is interrupt consumers using
  * interrupt_configure() or friends (interrupt_add_handler(),
  * interrupt_add_configure_handler(), interrupt_add_handler_with_chip(),
  * etc...).
```

jenswi-linaro:
Looks good if you remove
```
# Conflicts:
#	core/include/kernel/interrupt.h
```
from the commit message in "core: interrupt: itr_chip may not require configure handler"

etienne-lms:
Right, sorry, i'll fix.

etienne-lms:
Commit log fixed and review tags applied.

Failure of CI Qemu build/run on arm64 host seems unrelated to this change but it's strange it fails on this P-R (systematic over 3 recent series udpate) while I saw a recently updated P-R that successfully passed all CI tests.

PR #7268, opened 2025-02-08
Updates to ELE driver on i.MX platforms

sahilnxp:
This PR is doing multiple changes in ELE driver wrt

- Moving of ELE driver in a separate folder under driver/crypto/ele
- Introduction of Memory management functions to be used in driver
- Multiple command formats changes
- Enablement of TRUST_MU on i.MX91 and i.MX93
- Get random number from ELE

I have put all these changes in a single PR.
Please let me know if need to be break it into multiple PRs

CC: @saschahauer 
https://github.com/OP-TEE/optee_os/pull/7256


core/drivers/crypto/ele/utils_mem.c
```diff
@@ -0,0 +1,108 @@
...
+	ptr = memalign(cacheline_size, alloc_size);
+	if (!ptr) {
+		EMSG("alloc Error - NULL");
+		return NULL;
+	}
+
+	memset(ptr, 0, alloc_size);
+
+	return ptr;
+}
```
saschahauer, line 34: This function does the same as alloc_cache_aligned(). Consider using that instead

core/drivers/crypto/ele/utils_mem.c
```diff
@@ -0,0 +1,108 @@
...
+/*
+ * Free allocated area
+ *
+ * @ptr  area to free
+ */
+static void imx_ele_free(void *ptr)
+{
+	if (ptr)
+		free(ptr);
+}
```
saschahauer, line 45: free() already accepts NULL pointers, no need to wrap this into a separate function

core/drivers/crypto/ele/utils_mem.c
```diff
@@ -0,0 +1,108 @@
...
+
+void imx_ele_buf_cache_op(enum utee_cache_operation op,
+			  struct imx_ele_buf *ele_buf)
+{
+	if (ele_buf && ele_buf->data)
+		cache_operation(op, ele_buf->data, ele_buf->size);
+}
```
saschahauer, line 52: This function is never used outside this file, no need to export it.

core/drivers/crypto/ele/crypto.mk
```diff
@@ -1,6 +1,9 @@
 ifeq ($(CFG_IMX_ELE),y)
-ifneq (,$(filter $(PLATFORM_FLAVOR),$(mx93-flavorlist) $(mx91-flavorlist)))
-# Disable software RNG when ELE driver is enabled
-$(call force, CFG_WITH_SOFTWARE_PRNG,n,Mandated by CFG_IMX_ELE)
-endif
```
saschahauer: This was just introduced with the last commit. Could you merge this patch with the previous patch? I think this would make it easier to follow

core/drivers/crypto/ele/ele.c
```diff
@@ -520,9 +572,6 @@ unsigned long plat_get_aslr_seed(void)
 	 */
 	assert(!cpu_mmu_enabled());
```
saschahauer, line 573: This patch introduces the virtual/physical address translation and cache maintenance, so the comment is no longer true and the assert() no longer needed.

core/arch/arm/plat-imx/registers/imx93.h
```diff
@@ -9,7 +9,13 @@
 #define GICR_BASE 0x48040000
 
 #define UART1_BASE 0x44380000
-#define MU_BASE	   0x47520000
+/*
+ * For Normal MU - Use MU_BASE as 0x47520000
+ * For Trust MU - Use MU_BASE as 0x47530000
+ */
+#define MU_BASE 0x47530000
```
saschahauer, line 16: This patch changes the MU base used for the ELE from the "Normal MU" to the "Trust MU". Could you add a few words to the commit message what the difference between both actually is and why we prefer the "Trust MU"?
This would allow people like me who have only limited access to documentation help understanding what's going on.

I am especially confused about an earlier commit message stating:

> There is only one MU to communicate with ELE, which cannot be
> dedicated on OP-TEE side all the time. There may be ELE services
> running on Linux side, which can cause conflict with OP-TEE.

Now this commit seems to imply there are actually two MUs (Normal and Trust), so why can't one be used for OP-TEE and the other one for Linux?

core/drivers/crypto/ele/crypto.mk
sahilnxp: I kept these in 2 commits, to give more context
Like we enabled the RNG first and then we found some issue with that, and we disabled that.
If squash both commits, need to explain both the introduction of RNG and disabling the RNG in a single commit which may make it more difficult to understand.

Maybe we can ask others opinion on this
@jforissier @jenswi-linaro @etienne-lms your views on this?

core/drivers/crypto/ele/ele.c
sahilnxp: I am sorry, I didn't get this statement. Where **we are introducing the virtual/physical address translation and cache maintenance** in this patch?

saschahauer: This patch introduces imx_ele_rng_get_random() which handles an enabled MMU. plat_get_aslr_seed() now uses this function, so it should work with MMU enabled

core/drivers/crypto/ele/crypto.mk
saschahauer: I don't think the history of how you enabled it and found problems afterwards is interesting. The interesting
part is that the ELE can either be used in OP-TEE or in the Kernel, but not both. Enabling CFG_WITH_SOFTWARE_PRNG=y makes sure that OP-TEE uses the ELE only during initialization,
but not during runtime of Linux.

sahilnxp: Ok, Will do.

core/arch/arm/plat-imx/registers/imx93.h
sahilnxp: Ok, Will update in the commit message.

KarlK90: Communcation via the TRUST_MU fails on boot with the following log:

```
D/TC:0   ele_trace_print_msg:15 Header version 0x6 size 0x1 tag 0x17 command 0xa4
D/TC:0   ele_trace_print_msg:32 Request:
D/TC:0   ele_trace_print_msg:36         [0] 0000000000
E/TC:0   imx_ele_call:203 Failed to transmit message: 0xffff000b
D/TC:0   ele_trace_print_msg:15 Header version 0x7 size 0x4 tag 0x17 command 0xcd
D/TC:0   ele_trace_print_msg:32 Request:
D/TC:0   ele_trace_print_msg:36         [0] 0000000000
D/TC:0   ele_trace_print_msg:36         [1] 0xbe0fb8c0
D/TC:0   ele_trace_print_msg:36         [2] 0x00000008
D/TC:0   ele_trace_print_msg:36         [3] 0000000000
E/TC:0   imx_ele_call:203 Failed to transmit message: 0xffff000d
E/TC:0   Panic 'Cannot retrieve random data from ELE' at /usr/src/debug/optee-os/4.5.0/core/drivers/crypto/ele/ele.c:632 <plat_get_aslr_seed>
```

I have tested this on an i.MX93 using this PR as well as downstream NXP op-tee and tf-a versions. What are the prerequisites to successfully communicate with the trust MU?

KarlK90, line 16: Communication via the TRUST_MU fails on boot with the following log:

```
D/TC:0   ele_trace_print_msg:15 Header version 0x6 size 0x1 tag 0x17 command 0xa4
D/TC:0   ele_trace_print_msg:32 Request:
D/TC:0   ele_trace_print_msg:36         [0] 0000000000
E/TC:0   imx_ele_call:203 Failed to transmit message: 0xffff000b
D/TC:0   ele_trace_print_msg:15 Header version 0x7 size 0x4 tag 0x17 command 0xcd
D/TC:0   ele_trace_print_msg:32 Request:
D/TC:0   ele_trace_print_msg:36         [0] 0000000000
D/TC:0   ele_trace_print_msg:36         [1] 0xbe0fb8c0
D/TC:0   ele_trace_print_msg:36         [2] 0x00000008
D/TC:0   ele_trace_print_msg:36         [3] 0000000000
E/TC:0   imx_ele_call:203 Failed to transmit message: 0xffff000d
E/TC:0   Panic 'Cannot retrieve random data from ELE' at /usr/src/debug/optee-os/4.5.0/core/drivers/crypto/ele/ele.c:632 <plat_get_aslr_seed>
```

I have tested this on an i.MX93 using this PR as well as downstream NXP op-tee and tf-a versions. What are the prerequisites to successfully communicate with the trust MU?

sahilnxp: Which BSP version are you using for this testing ?

sahilnxp:
PR updated with comments addressed.

core/arch/arm/plat-imx/registers/imx93.h
KarlK90: That was kind of a false alarm. I was using a i.MX93 SoC revision **A0**, which seems to be broken. With a **A1** this works as expected.

sahilnxp: Thanks @KarlK90  for the update.

core/drivers/crypto/ele/crypto.mk
```diff
@@ -0,0 +1,9 @@
+ifeq ($(CFG_IMX_ELE),y)
+
+# Issues in the ELE FW prevent OPTEE and Kernel from using
+# the RNG concurrently at runtime. To prevent any issue,
+# use the software RNG instead in OPTEE.
+# But with Kernel ELE driver disabled, Runtime ELE RNG
+# generation can be done.
+CFG_WITH_SOFTWARE_PRNG ?= y
```
saschahauer, line 8: This comment says that the Kernel and OP-TEE cannot both use the RNG at runtime whereas the commit message talks about problems when both use the ELE at runtime.
Could you clarify this a bit? Are the problems limited to the RNG or are we generally forbidden to use the ELE in both the Kernel and OP-TEE?
Are any of the problems solved by using two MUs instead of one as done earlier in this series?

sahilnxp: Problem comes when doing simultaneous request to ELE from Kernel and OP-TEE not specific to RNG, and we are working on the solution, which will be available soon.

sahilnxp: Done.

sahilnxp:
Any more comments on this one ?

jforissier:
Reviewing this makes me realize we might not have made the best choices regarding the file paths (sources and includes) for the crypto drivers. I think there is no need for a new include directory for each crypto IP, and having qualified paths in the `#include <...>` would be more clear. I mean `#include <drivers/crypto/imx/ele.h` rather than `#include <imx_ele.h>`. Keeping `imx` in the path because it is not obvious that `ele` is `imx` and also to better organize the code (`imx/ele`, `imx/caam`...).
So I would suggest the following paths under `core/` (all sources under `drivers`, all public headers under `include/drivers`):
- C files: `drivers/crypto/imx/ele/ele.c`, `drivers/crypto/imx/ele/*.c` etc.
- Header(s): `include/drivers/crypto/imx/ele.h`,  used as  `#include <drivers/crypto/imx/ele.h>`
...instead of what is proposed here (all source under `drivers`, some public headers under `include/drivers`, some others under `drivers/something/include`):
- C files: `drivers/crypto/ele/ele.c`, `drivers/crypto/ele/*.c`
- Headers(s): `drivers/crypto/ele/include/ele.h` 

Applying the same to CAAM: `./drivers/crypto/caam/include/caam_cipher.h` could become `./include/drivers/crypto/caam/cipher.h` thus avoiding the double `caam` in the path and the extra include folder. It would be used as `#include <drivers/crypto/caam/cipher.h>` instead of `#include <caam_cipher.h>`. Many C files could also drop the `caam_` prefix (`./drivers/crypto/caam/acipher/caam_rsa.c` -> `./drivers/crypto/caam/acipher/rsa.c` etc.).

I know this is mostly cosmetic but as new drivers are introduced the older ones serve as templates so if we're not comfortable with something we should change it, otherwise it will get worse with time.

What do others think? We could also decide to rework this later. This is all very much internal to the OP-TEE tree anyways,

One more related thing: why export utils_mem.h ? It seems it could be private to the ELE driver in drivers/crypto/imx/ele/ and used with a local `#include "..."` by `ele.c` etc. And `memutils.h` sounds better IMO.

jenswi-linaro:
The proposed paths look good to me. It might be best if you rename the files after this PR is merged.

sahilnxp:
Thanks @jenswi-linaro @jforissier 
Overall, I agree with your suggestion.

As @jenswi-linaro said, I will take up this thing after this PR is merged.

BTW, apart from this any other comment on this PR?

core/drivers/crypto/ele/ele.c
```diff
@@ -359,33 +313,65 @@ static TEE_Result imx_ele_session_close(uint32_t session_handle)
...
+	memcpy(uid, rsp.uid, MIN(sizeof(rsp.uid), len));
+
+	is_fetched = true;
 out:
-	/*
-	 * In the device info array return by the ELE, the words 2, 3, 4 and 5
-	 * are the device UID.
-	 */
-	memcpy(buffer, rsp.chip_uid, MIN(sizeof(rsp.chip_uid), len));
+	memcpy(buffer, uid, MIN(sizeof(uid), len));
```
jenswi-linaro, line 366: Am I right to assume that the "die id" will remain unchanged, it's only a new way of obtaining it?

core/drivers/crypto/ele/ele.c
```diff
@@ -395,45 +391,65 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 		uint16_t key_size;
 		uint16_t ctx_size;
 		uint32_t crc;
-	} __packed cmd = { };
+	} __packed cmd = {};
```
jenswi-linaro, line 403: The preferred form is `{ }`, note the space between the curly braces. Please fix this for all changes in this PR.

core/drivers/crypto/ele/ele.c
```diff
@@ -395,45 +391,65 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+{
+	static const char pattern[] = "TEE_for_HUK_ELE";
+	static uint8_t key[HW_UNIQUE_KEY_LENGTH];
+	static bool is_fetched;
+
+	if (is_fetched)
+		goto out;
+
+	if (imx_ele_derive_key((const uint8_t *)pattern, sizeof(pattern), key,
```
jenswi-linaro, line 458: Am I right to assume that the HUK will remain the same, it's only the way you're obtaining it that has changed?

core/drivers/crypto/ele/ele.c
```diff
@@ -495,33 +497,73 @@ static TEE_Result imx_ele_rng_get_trng_state(void)
...
+
+		if (!IS_ALIGNED_WITH_TYPE(pa, uint32_t))
+			return TEE_ERROR_BAD_PARAMETERS;
+
+		reg_pair_from_64((uint64_t)pa, &cmd.addr_msb, &cmd.addr_lsb);
+	}
 
-	reg_pair_from_64((uint64_t)&aslr, &cmd.addr_msb, &cmd.addr_lsb);
-	cmd.size = sizeof(aslr);
+	cmd.size = (uint32_t)size;
```
jenswi-linaro, line 546: The cast isn't needed.

core/drivers/crypto/ele/ele.c
```diff
@@ -495,33 +497,73 @@ static TEE_Result imx_ele_rng_get_trng_state(void)
...
+out:
+	imx_ele_buf_free(&rng);
+
+	return res;
+}
+
+unsigned long plat_get_aslr_seed(void)
+{
+	uint64_t timeout = timeout_init_us(10 * 1000);
+
```
jenswi-linaro, line 565: Please remove this empty line.

core/drivers/crypto/ele/ele.c
```diff
@@ -495,33 +497,73 @@ static TEE_Result imx_ele_rng_get_trng_state(void)
...
+ *
+ * @buffer: data output
+ * @size: RNG data size
+ */
+static TEE_Result imx_ele_rng_get_random(uint8_t *buffer, size_t size)
 {
 	TEE_Result res = TEE_ERROR_GENERIC;
-	uint64_t timeout = timeout_init_us(10 * 1000);
+	struct imx_ele_buf rng = {};
+
```
jenswi-linaro, line 514: Please remove the empty line.

core/drivers/crypto/ele/ele.c
sahilnxp: Yes, you are right.

core/drivers/crypto/ele/ele.c
sahilnxp: Yes, you are right.

core/drivers/crypto/ele/ele.c
sahilnxp: Done.

core/drivers/crypto/ele/ele.c
sahilnxp: Done.

core/drivers/crypto/ele/ele.c
sahilnxp: Done.

core/drivers/crypto/ele/ele.c
sahilnxp: Done.

sahilnxp: CI is reporting an error without new line: https://github.com/OP-TEE/optee_os/actions/runs/13407450824/job/37449890100?pr=7268#step:5:61

sahilnxp: @jenswi-linaro please help me with this.

jenswi-linaro: The checkpatch warning looks like a false positive.
I usually run checkpatch as `./scripts/checkpatch.sh github/master..`, if `github/master` is the base of my patches.

sahilnxp: @jenswi-linaro IIUC this error is coming from `checkdiff` script.
`checkpatch` is passing correctly.

jenswi-linaro: I guess you can source `scripts/checkpatch_inc.sh` in bash and call the function from the prompt.

sahilnxp: Thanks @jenswi-linaro. I am able to run the `checkdiff` on my local machine.
It is also reporting the same issue:
```
:~/optee-os$ checkdiff acdc32afe18f8b71915184a01e7279ffcbe0ec20 47d18800fe372fdb4fad575daba61262e814d171
WARNING: Missing a blank line after declarations
#669: FILE: core/drivers/crypto/ele/ele.c:564:
+       uint64_t timeout = timeout_init_us(10 * 1000);
+       unsigned long aslr __aligned(CACHELINE_SIZE) = 0;

total: 0 errors, 1 warnings, 0 checks, 932 lines checked
:~/optee-os$
```

So, we should ignore it, right ?

jenswi-linaro: Yes

sahilnxp: Ok. Thanks @jenswi-linaro

jenswi-linaro:
Looks good to me.

core/drivers/crypto/ele/ele.c
```diff
@@ -359,33 +313,65 @@ static TEE_Result imx_ele_session_close(uint32_t session_handle)
...
+	memcpy(uid, rsp.uid, MIN(sizeof(rsp.uid), len));
+
+	is_fetched = true;
 out:
-	/*
-	 * In the device info array return by the ELE, the words 2, 3, 4 and 5
-	 * are the device UID.
-	 */
-	memcpy(buffer, rsp.chip_uid, MIN(sizeof(rsp.chip_uid), len));
+	memcpy(buffer, uid, MIN(sizeof(uid), len));
```
etienne-lms, line 366: nitpicking suggestion:
```c
int tee_otp_get_die_id(uint8_t *buffer, size_t len)
{
	static uint32_t uid[4];
	static bool is_fetched;
	struct get_info_rsp rsp = { };

	assert(buffer && len);

	if (!is_fetched) {
		if (imx_ele_get_device_info(&rsp))
			panic("Fail to get the device UID");

		memcpy(uid, rsp.uid, MIN(sizeof(rsp.uid), len));
		is_fetched = true;
	}

	memcpy(buffer, uid, MIN(sizeof(uid), len));

	return 0;
}
```

core/drivers/crypto/ele/ele.c
```diff
@@ -34,6 +34,7 @@
 #define ELE_CMD_TRNG_STATE	    0xA4
 #define ELE_CMD_GET_INFO	    0xDA
 #define ELE_CMD_DERIVE_KEY	    0xA9
+#define ELE_CMD_SAB_INIT 0x17
```
etienne-lms, line 39: Indent like the comand IDs above?

core/drivers/crypto/ele/ele.c
```diff
@@ -423,7 +409,7 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 	/*
 	 * Intermediate msb and lsb values are needed. Directly using
 	 * key_addr_msb and key_addr_lsb might be unaligned because of the
-	 * __packed attribute of key_derive_cmd {}
```
etienne-lms, line 527: change needed? maybe unrelated.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for "drivers: ele: move ELE to a dedicated directory".

Comment to address (if I'm right) in commit "drivers: ele: add memory management functions".

`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for the other commits in the series, 2 with minor comments addressed or not.

core/drivers/crypto/ele/include/utils_mem.h
```diff
@@ -0,0 +1,65 @@
...
+};
+
+/*
+ * Cache operation on IMX ELE buffer
+ *
+ * @op: Cache operation
+ * @ele_buf: Buffer on which cache operation to be performed
+ */
+void imx_ele_buf_cache_op(enum utee_cache_operation op,
+			  struct imx_ele_buf *ele_buf);
```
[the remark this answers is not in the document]
sahilnxp: Changed to` static function` in fixup commit.

core/drivers/crypto/ele/utils_mem.c
```diff
@@ -0,0 +1,108 @@
...
+TEE_Result imx_ele_buf_copy(struct imx_ele_buf *ele_buf, uint8_t *buf,
+			    size_t size)
+{
+	if (!ele_buf || !buf || !size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (size < ele_buf->size)
+		return TEE_ERROR_SHORT_BUFFER;
+
+	imx_ele_buf_cache_op(TEE_CACHEINVALIDATE, ele_buf);
```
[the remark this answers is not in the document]
sahilnxp: I don't think that is required, we are invalidating the `ele_buf->data` in `imx_ele_buf_cache_op()`

core/drivers/crypto/ele/ele.c
sahilnxp: Not needed, removed.

sahilnxp:
@etienne-lms comments addressed, renamed `utils_mem.x` to `memutils.x`
Will move to required directory as part of directory restructuring pull request.
Also commits squashed and tags updated.

core/drivers/crypto/ele/ele.c
```diff
@@ -491,33 +493,71 @@ static TEE_Result imx_ele_rng_get_trng_state(void)
...
+out:
+	imx_ele_buf_free(&rng);
+
+	return res;
+}
+
+unsigned long plat_get_aslr_seed(void)
+{
+	uint64_t timeout = timeout_init_us(10 * 1000);
+	unsigned long aslr __aligned(CACHELINE_SIZE) = 0;
```
etienne-lms, line 560: Checkpatch seems happier with attributes before the variable label, not between the label and `= 0` affectation.
```patch
 unsigned long plat_get_aslr_seed(void)
 {
 	uint64_t timeout = timeout_init_us(10 * 1000);
-	unsigned long aslr __aligned(CACHELINE_SIZE) = 0;
+	unsigned long __aligned(CACHELINE_SIZE) aslr = 0;
 
 	/*
 	 * Check the current TRNG state of the ELE. The TRNG must be
```
It seems gcc does not care.

sahilnxp: oh, that was the issue. Thanks @etienne-lms
Updated pull request.

core/drivers/crypto/ele/crypto.mk
bith3ad: @sahilnxp gentle ping.

PR #7269, opened 2025-02-10
core: mm: initialize max_allocated size in pool init function

hangyuli0:
The TA max_allocated size obtained from stat PTA is incorrect. Pool max_allocated size is not initialized.
Initialize max_allocated size in pool init function.

Change-Id: Ia26c2a9dcc9692d6b4201b9aaea59c24de4a70f4

### DIFF
--- a/core/mm/tee_mm.c
+++ b/core/mm/tee_mm.c
@@ -68,6 +68,10 @@ bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, paddr_size_t size,
 	pool->entry->pool = pool;
 	pool->lock = SPINLOCK_UNLOCK;
 
+#ifdef CFG_WITH_STATS
+	pool->max_allocated = 0;
+#endif
+
 	return true;
 }
 


core/mm/tee_mm.c
```diff
@@ -68,6 +68,10 @@ bool tee_mm_init(tee_mm_pool_t *pool, paddr_t lo, paddr_size_t size,
 	pool->entry->pool = pool;
 	pool->lock = SPINLOCK_UNLOCK;
 
+#ifdef CFG_WITH_STATS
+	pool->max_allocated = 0;
+#endif
+
```
jforissier, line 74: This should not be needed due to the way `*pool` is initialized above. All fields except `.lo`, `.size`, `.shift` and `.flags` (which are set explicitly) are expected to be zero-initialized.
Which compiler are you using?

etienne-lms:
@hangyuli0, maybe your issue was fixed with https://github.com/OP-TEE/optee_os/pull/7224

hangyuli0:
> @hangyuli0, maybe your issue was fixed with #7224

@etienne-lms Yes, my issue could be fixed with #7224. My local code is v4.4.0, do not include this fix patch. Thank you.

core/mm/tee_mm.c
hangyuli0: My compiler is gcc 13.3. This issue could be fixed with https://github.com/OP-TEE/optee_os/pull/7224. Thanks for your reply.

jforissier:
> > @hangyuli0, maybe your issue was fixed with #7224
> 
> @etienne-lms Yes, my issue could be fixed with #7224. My local code is v4.4.0, do not include this fix patch. Thank you.

OK, therefore I am closing this PR. Thanks for reporting though.

PR #7270, opened 2025-02-10
AMD mailbox driver support

Akshay-Belsare:
Add support for the AMD Mailbox Service for use on AMD platforms.
The Mailbox Service requires the Local IPI ID as a mandatory input,
provided via CFG_MAILBOX_LOCAL_ID.
Mailbox initialization for CFG_MAILBOX_LOCAL_ID occurs as a late init
service and will be utilized by various drivers to communicate with other
firmware.
The remote ID will be supplied by different drivers using the mailbox
service.

### DIFF
--- a/core/drivers/amd/mailbox_driver.c
+++ b/core/drivers/amd/mailbox_driver.c
@@ -9,6 +9,7 @@
 #include <initcall.h>
 #include <kernel/delay.h>
 #include <kernel/panic.h>
+#include <kernel/thread_arch.h>
 #include <mailbox_private.h>
 #include <malloc.h>
 #include <mm/core_mmu.h>
@@ -57,7 +58,7 @@ TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
 	if (!ipi.req) {
 		EMSG("Failed to map request buffer");
 		ipi.remote = 0;
-		return res;
+		return TEE_ERROR_GENERIC;
 	}
 
 	ipi.rsp = core_mmu_add_mapping(MEM_AREA_RAM_SEC,
@@ -65,7 +66,9 @@ TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
 				       payload_size);
 	if (!ipi.rsp) {
 		EMSG("Failed to map response buffer");
-		/* Remove mapping for request buffer */
+		/* Remove mapping for request buffer.
+		 * Not checking return value res, since already in error case.
+		 */
 		res = core_mmu_remove_mapping(MEM_AREA_RAM_SEC,
 					      ipi.req,
 					      payload_size);
@@ -79,9 +82,9 @@ TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
 
 	res = mailbox_call(IPI_MAILBOX_OPEN, IPI_BLOCK);
 	if (res != TEE_SUCCESS) {
-		EMSG("Failed to open Mailbox for remote id %"PRIu32, remote_id);
-		/* Reset struct ipi_info.
-		 * Not checking return, already in error case
+		EMSG("Failed to open Mailbox for remote ID %"PRIu32, remote_id);
+		/* Reset IPI information structure.
+		 * Not checking return value res, already in error case
 		 */
 		res = core_mmu_remove_mapping(MEM_AREA_RAM_SEC,
 					      ipi.req,
@@ -105,7 +108,7 @@ TEE_Result mailbox_release(uint32_t remote_id, size_t payload_size)
 	TEE_Result res = TEE_ERROR_GENERIC;
 
 	if (ipi.remote != remote_id) {
-		EMSG("Mailbox not available for remote id %"PRIu32, remote_id);
+		EMSG("Mailbox not available for remote ID %"PRIu32, remote_id);
 		return res;
 	}
 
@@ -113,7 +116,7 @@ TEE_Result mailbox_release(uint32_t remote_id, size_t payload_size)
 
 	res = mailbox_call(IPI_MAILBOX_RELEASE, IPI_BLOCK);
 	if (res != TEE_SUCCESS) {
-		EMSG("Mailbox release failed for remote id %"PRIu32, remote_id);
+		EMSG("Mailbox release failed for remote ID %"PRIu32, remote_id);
 		goto out;
 	}
 
@@ -121,7 +124,7 @@ TEE_Result mailbox_release(uint32_t remote_id, size_t payload_size)
 				      ipi.req,
 				      payload_size);
 	if (res != TEE_SUCCESS) {
-		EMSG("Failed to remove response mapping for remote id %"PRIu32,
+		EMSG("Failed to remove response mapping for remote ID %"PRIu32,
 		     remote_id);
 		goto out;
 	}
@@ -130,7 +133,7 @@ TEE_Result mailbox_release(uint32_t remote_id, size_t payload_size)
 				      ipi.rsp,
 				      payload_size);
 	if (res != TEE_SUCCESS) {
-		EMSG("Failed to remove request mapping for remote id %"PRIu32,
+		EMSG("Failed to remove request mapping for remote ID %"PRIu32,
 		     remote_id);
 	}
 
@@ -149,7 +152,7 @@ static TEE_Result mailbox_write(uint32_t remote_id,
 	TEE_Result res = TEE_ERROR_GENERIC;
 
 	if (ipi.remote != remote_id) {
-		EMSG("Mailbox not available for remote id %"PRIu32, remote_id);
+		EMSG("Mailbox not available for remote ID %"PRIu32, remote_id);
 		return res;
 	}
 
@@ -175,20 +178,16 @@ static TEE_Result mailbox_read(uint32_t remote_id,
 			       size_t payload_size,
 			       uint32_t *status)
 {
-	TEE_Result res = TEE_ERROR_GENERIC;
-
 	cache_operation(TEE_CACHEINVALIDATE, ipi.rsp, payload_size);
 
 	*status = *(uint32_t *)ipi.rsp;
 
-	if (*status) {
-		res = TEE_ERROR_GENERIC;
-		goto err;
-	}
+	if (*status)
+		return TEE_ERROR_GENERIC;
 
 	memcpy(ipi.rsp, payload, payload_size);
-err:
-	return res;
+
+	return TEE_SUCCESS;
 }
 
 TEE_Result mailbox_notify(uint32_t remote_id,
@@ -200,7 +199,7 @@ TEE_Result mailbox_notify(uint32_t remote_id,
 	uint32_t remote_status = 0;
 
 	if (ipi.remote != remote_id) {
-		EMSG("Mailbox not available for remote id %"PRIu32, remote_id);
+		EMSG("Mailbox not available for remote ID %"PRIu32, remote_id);
 		return res;
 	}
 
@@ -239,8 +238,8 @@ out:
 static TEE_Result mailbox_init(void)
 {
 	ipi.local = CFG_MAILBOX_LOCAL_ID;
-	ipi.buf = (PLAT_IPI_BASE_ADDR +
-			(CFG_MAILBOX_LOCAL_ID * IPI_BASE_MULTIPLIER));
+	ipi.buf = PLAT_IPI_BASE_ADDR +
+		CFG_MAILBOX_LOCAL_ID * IPI_BASE_MULTIPLIER;
 
 	mutex_init(&ipi.lock);
 
--- a/core/drivers/amd/mailbox_driver.c
+++ b/core/drivers/amd/mailbox_driver.c
@@ -1,0 +1,211 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+#include <initcall.h>
+#include <kernel/delay.h>
+#include <kernel/panic.h>
+#include <malloc.h>
+#include <mm/core_mmu.h>
+#include <string.h>
+#include <tee/cache.h>
+
+#include <mailbox_private.h>
+
+#include <drivers/amd/mailbox_driver.h>
+
+static struct ipi_info_t ipi_info;
+
+static TEE_Result mailbox_call(enum mailbox_api ipi_api, uint32_t blocking)
+{
+	struct thread_smc_args args = {
+		.a0 = PLAT_SMC_SIP_IPI | ipi_api,
+		.a1 = reg_pair_to_64(0, ipi_info.local),
+		.a2 = reg_pair_to_64(0, ipi_info.remote),
+		.a3 = reg_pair_to_64(0, blocking),
+	};
+
+	thread_smccc(&args);
+
+	if (args.a0) {
+		EMSG("Mailbox Call returned with error 0x%lx", args.a0);
+		return TEE_ERROR_GENERIC;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	ipi_info.remote = remote_id;
+
+	paddr_t req_addr = (ipi_info.buf + (remote_id * IPI_OFFSET_MULTIPLIER) +
+			IPI_REQ_OFFSET);
+
+	paddr_t rsp_addr = (ipi_info.buf + (remote_id * IPI_OFFSET_MULTIPLIER) +
+			IPI_RESP_OFFSET);
+
+	ipi_info.req = core_mmu_add_mapping(MEM_AREA_RAM_SEC,
+					    req_addr,
+					    payload_size);
+
+	ipi_info.rsp = core_mmu_add_mapping(MEM_AREA_RAM_SEC,
+					    rsp_addr,
+					    payload_size);
+	if (!ipi_info.req || !ipi_info.rsp)
+		panic();
+
+	mutex_lock(&ipi_info.lock);
+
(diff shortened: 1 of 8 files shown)


core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,212 @@
...
+	res = mailbox_call(IPI_MAILBOX_RELEASE, IPI_BLOCK);
+	if (res != TEE_SUCCESS) {
+		EMSG("Mailbox release failed for remote id %d", remote_id);
+		goto err;
+	}
+
+	res = core_mmu_remove_mapping(MEM_AREA_RAM_SEC,
+				      ipi_info.req,
+				      payload_size);
+	res |= core_mmu_remove_mapping(MEM_AREA_RAM_SEC,
```
jenswi-linaro, line 94: This is a micro-optimization, please report the errors separately.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,212 @@
...
+				       ipi_info.rsp,
+				       payload_size);
+	if (res != TEE_SUCCESS) {
+		EMSG("Failed to remove memory mappings for remote id %d",
```
jenswi-linaro, line 98: `remote_id` is a `uint32_t`, so please use `PRId32` instead of `d`

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,212 @@
...
+	res = mailbox_call(IPI_MAILBOX_OPEN, IPI_BLOCK);
+	if (res != TEE_SUCCESS)
+		EMSG("Failed to open Mailbox for remote id %d", remote_id);
+
+	mutex_unlock(&ipi_info.lock);
+
+	return res;
+}
+
+TEE_Result mailbox_release(uint32_t remote_id, size_t payload_size)
```
jenswi-linaro, line 74: Is this function expected to be called more than once?

jenswi-linaro (commented):
Please fix the checkpatch issues.
This code is unused, do you plan to upstream code that uses the mailbox driver any time soon?

core/drivers/amd/mailbox_driver.c
Akshay-Belsare: nope, only once.

jenswi-linaro: Good, because `core_mmu_remove_mapping()` isn't always able to return the used virtual memory. I'm working on upstreaming better support for dynamic mapping of secure memory, but until then we're stuck with `core_mmu_remove_mapping()`.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,212 @@
...
+				      payload_size);
+	res |= core_mmu_remove_mapping(MEM_AREA_RAM_SEC,
+				       ipi_info.rsp,
+				       payload_size);
+	if (res != TEE_SUCCESS) {
+		EMSG("Failed to remove memory mappings for remote id %d",
+		     remote_id);
+	}
+
+err:
```
etienne-lms, line 102: Could you rename `err` to `out` since this path is not dedicated to error only cases.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,212 @@
...
+	if (res) {
+		EMSG("IPI error");
+		goto out;
+	}
+
+	res = mailbox_read(ipi_info.remote, payload, payload_size, &remote_status);
+	if (res)
+		EMSG("Can't read the remote response");
+
+	if (remote_status)
```
etienne-lms, line 191: Nitpicking: `res` is already set to `TEE_ERROR_GENERIC` when `remote_status` is not 0.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,212 @@
...
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	cache_operation(TEE_CACHEINVALIDATE, ipi_info.rsp, payload_size);
+
+	*status = *(uint32_t *)ipi_info.rsp;
+
+	if (*status)
+		goto err;
+
+	if (ipi_info.rsp) {
```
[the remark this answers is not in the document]
Akshay-Belsare: @etienne-lms, ipi_info.rsp contans address and at line 146 we are getting the value at the address pointed by ipi_info.rsp
if the value is >0 we return error.

core/include/drivers/amd/mailbox_driver.h
```diff
@@ -0,0 +1,45 @@
...
+	uint32_t remote;
+
+	paddr_t buf;
+
+	/* Exclusive access to the IPI shared buffer */
+	struct mutex lock;
+
+	void *rsp;
+	void *req;
+} ipi_info_t;
```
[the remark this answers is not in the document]
Akshay-Belsare: will use struct ipi_info_t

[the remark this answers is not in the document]
jenswi-linaro: How about making it `struct ipi_info`?

core/drivers/amd/mailbox_driver.c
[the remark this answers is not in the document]
etienne-lms: I meant that if `ipi_info.rsp == NULL`, we can't reach this line.
I guess you don't expect one calls `mailbox_notify()` is `mailbox_open()` failed. So this address should never be NULL if I'm right.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,216 @@
...
+}
+
+static TEE_Result mailbox_read(uint32_t remote_id,
+			       void *payload,
+			       size_t payload_size,
+			       uint32_t *status)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	cache_operation(TEE_CACHEINVALIDATE, ipi_info.rsp, payload_size);
```
etienne-lms, line 146: By the way, `cache_operation()` may return an error code. If never awaited, maybe asserting the return value is enough. That said, I see that almost all callers of that function do not check the return value :(
There are all in `core/drivers/...`. Note that if `ipi_info.rsp` is NULL, the function would have returned an error.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	res = core_mmu_remove_mapping(MEM_AREA_RAM_SEC,
+				      ipi_info.req,
+				      payload_size);
+	if (res != TEE_SUCCESS) {
+		EMSG("Failed to remove response mapping for remote id %"PRIu32,
+		     remote_id);
+		goto out;
+	}
```
etienne-lms, line 96: Here and below, `ipi_info.req` and `ipi_info.rsp` should be reset to NULL to avoid pointing to an unmapped address or more dramatically another physical location.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,216 @@
...
+	TEE_Result res = TEE_ERROR_GENERIC;
+	paddr_t req_addr = 0, rsp_addr = 0;
+
+	ipi_info.remote = remote_id;
+
+	req_addr = (ipi_info.buf + (remote_id * IPI_OFFSET_MULTIPLIER) +
+		    IPI_REQ_OFFSET);
+
+	rsp_addr = (ipi_info.buf + (remote_id * IPI_OFFSET_MULTIPLIER) +
+		    IPI_RESP_OFFSET);
```
etienne-lms, line 49: Nitpicking: could remove inner and outer parentheses in both instructions above.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,216 @@
...
+		return TEE_ERROR_GENERIC;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	paddr_t req_addr = 0, rsp_addr = 0;
```
etienne-lms, line 41: Prefer 1 variable definition per line.

core/drivers/amd/mailbox_driver.c
```diff
@@ -0,0 +1,216 @@
...
+		.a0 = PLAT_SMC_SIP_IPI | ipi_api,
+		.a1 = reg_pair_to_64(0, ipi_info.local),
+		.a2 = reg_pair_to_64(0, ipi_info.remote),
+		.a3 = reg_pair_to_64(0, blocking),
+	};
+
+	thread_smccc(&args);
+
+	if (args.a0) {
+		EMSG("Mailbox Call returned with error 0x%lx", args.a0);
```
etienne-lms, line 31: `args.a0` is always `uint64_t` in the case of this driver, if I'm right, so prefer `PRIx64`.
`		EMSG("Mailbox Call returned with error 0x%"PRIx64, args.a0);`

That said, if 32bit core is to be considered (I see that plat-versal2/conf.mk allows that: `CFG_ARM64_core ?= y`) then maybe prefer using a cast to ensure both cases are supported:
`		EMSG("Mailbox Call returned with error 0x%lx", (unsigned long)args.a0);`

core/include/drivers/amd/mailbox_driver.h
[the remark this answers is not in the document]
Akshay-Belsare: sure, making it struct ipi_info

core/drivers/amd/mailbox_driver.c
Akshay-Belsare: Let me revisit and modify accordingly.

ldts:
um, isnt this driver just the same than versal_mbox.c currently upstream? it seems that the only difference is the base address

ldts:
Also, if this is an `init_late` it means you wont be able to use some of the chip cryptographic features like the PUF or SHA (or fuses for that matter) during boot. Not sure if it applies to your platform but just a heads up.
This is the reason why on versal we chose to open the mailbox early.

ldts:
Also notice that given the nature of the firmware receiving the requests,  mbox requests/responses usually consist of  several buffers and the payload and each buffer had alignment requirements. as @jenswi-linaro mentioned, if would be nice to have an example of usage.

Akshay-Belsare:
@ldts This driver is meant for a newer AMD Versal Gen 2 family. This change is having the base infra, more will be added to it later. The use case is different from the the versal platform.

ldts:
> @ldts This driver is meant for a newer AMD Versal Gen 2 family and later versions. This change is having the base infra, more will be added to it later. The use case is different from the the versal platform.

What do you mean by "more will be added to it later"? Are you referring to additional changes to this mailbox driver under review? This driver is a simplified version of versal_mbox.c, and my concern is that you may eventually need to make it more complex, resulting in something similar to what already exists.

ldts:
one other piece of information, @jcorbier is doing some extensions to the versal_mbox driver (buffer memory management, mailbox timeouts etc https://github.com/OP-TEE/optee_os/pull/6738/commits/5418d26ca9fb873c577744314565469224c0a212 ). So maybe that is something that you could also use (ie, timeouts)

Akshay-Belsare:
@ldts 
> What do you mean by "more will be added to it later"? Are you referring to additional changes to this mailbox driver under review? This driver is a simplified version of versal_mbox.c, and my concern is that you may eventually need to make it more complex, resulting in something similar to what already exists.

As I mentioned, AMD Versal Gen 2 is new family and the underlying workings are not exactly same as Versal and Versal NET.

> one other piece of information, @jcorbier is doing some extensions to the versal_mbox driver (buffer memory management, mailbox timeouts etc https://github.com/OP-TEE/optee_os/commit/5418d26ca9fb873c577744314565469224c0a212 ). So maybe that is something that you could also use (ie, timeouts)
Appreciate, will have a look in to those changes as well.

ldts:
thanks. but please, do take the time to ask my questions to avoid going in circles. This driver is a simplified version of what we currently have.

Akshay-Belsare:
@etienne-lms @jenswi-linaro 
Added new patch addressing the review comments.

ldts:
um you made me curious about the IPI Gen2 now, where can I find the PLM code? should it still be in the [embeddedsw](https://github.com/Xilinx/embeddedsw.git) repo or is it now hosted somwhere else?

nathan-menhorn:
> it has less features - no tracing support, no validation of multiple buffers - but other than that I fail to see the difference.
> 
> could you comment please?

Hi @ldts all PLM code is in the embedded software directory. This will be using the versal_aiegen2 specific code along with the common code.

Akshay-Belsare:
@etienne-lms Addressed comments from you.

ldts:
> > it has less features - no tracing support, no validation of multiple buffers - but other than that I fail to see the difference.
> > could you comment please?
> 
> Hi @ldts all PLM code is in the embedded software directory. This will be using the versal_aiegen2 specific code along with the common code.

thanks Nathan. Um so I fail to see how this is architecturally (software wise) different than Versal or Versal net: they share the same TF-A mailbox service and similar PLM services (multiple buffers on the payload). It seems to me it will ultimately need the versal_mbox features but will see.

core/drivers/amd/mailbox_driver.c
```diff
@@ -79,9 +82,9 @@ TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
 
 	res = mailbox_call(IPI_MAILBOX_OPEN, IPI_BLOCK);
 	if (res != TEE_SUCCESS) {
-		EMSG("Failed to open Mailbox for remote id %"PRIu32, remote_id);
-		/* Reset struct ipi_info.
-		 * Not checking return, already in error case
+		EMSG("Failed to open Mailbox for remote ID %"PRIu32, remote_id);
+		/* Reset IPI information structure.
```
etienne-lms, line 86:
```patch
-		/* Reset IPI information structure.
+		/*
+		 * Reset IPI information structure.
 		 * Not checking return value res, already in error case
 		 */

File: core/drivers/amd/mailbox_driver.c [hunk 14]
```diff
@@ -57,15 +58,17 @@ TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size)
...
 
 	ipi.rsp = core_mmu_add_mapping(MEM_AREA_RAM_SEC,
 				       rsp_addr,
 				       payload_size);
 	if (!ipi.rsp) {
 		EMSG("Failed to map response buffer");
-		/* Remove mapping for request buffer */
+		/* Remove mapping for request buffer.
+		 * Not checking return value res, since already in error case.
+		 */
```
etienne-lms, line 71:
```patch
-		/* Remove mapping for request buffer.
+		/*
+		 * Remove mapping for request buffer.
 		 * Not checking return value res, since already in error case.
 		 */

**Review by etienne-lms — COMMENTED** [review]
LGTM with few comments.
Could you squash the fixup commits?

**Comment by ldts** [comment]
@etienne-lms could you comment my concerns before providing any form of ack?

File: core/drivers/amd/mailbox_driver.c (hunk 13 above)
Comment (etienne-lms) [reply core/drivers/amd/mailbox_driver.c:86]: Could you address these coding convention issues?

**Review by etienne-lms — COMMENTED** [review]
Few remaining comments.
Fell free to squash current and coming fixup commit in your next update.

"plat-versal2: enable AMD mailbox driver".

File: core/include/drivers/amd/mailbox_driver.h [hunk 15]
```diff
@@ -0,0 +1,22 @@
...
+#define __AMD_MAILBOX_DRIVER_H__
+
+#include <tee_api_types.h>
+#include <types_ext.h>
+#include <util.h>
+
+#define IPI_NON_BLOCK	U(0x0)
+#define IPI_BLOCK	U(0x1)
+
+TEE_Result mailbox_open(uint32_t remote_id, size_t payload_size);
```
jenswi-linaro, line 17: Please add an `amd_` prefix for these functions.

jenswi-linaro:
@Akshay-Belsare, please address @ldts's concerns. He has been a long-term contributor to this area. OP-TEE is a community-driven project, so the community's view matters.

Akshay-Belsare:
@ldts
We sincerely appreciate your concern. However, the current implementation of the Versal Mailbox driver is highly specific to the Versal family and, in its present form, cannot be seamlessly extended to support the next generation of SoCs.
For example, the Versal Mailbox driver is designed to facilitate communication with a single component, whereas next SoC families will require a more scalable solution that supports multiple components.
Our approach focuses on developing  drivers that can cater to a broader range of SoCs.  Once these generic drivers are in place, the Versal family can gradually transition to them over time. 
Additionally, we want to ensure that our approach preserves the existing functionality of the Versal family, allowing continued development without disruption to its current capabilities.

etienne-lms:
Hi @ldts, sorry, I see I missed you comment https://github.com/OP-TEE/optee_os/pull/7270#issuecomment-2687733839. I would hardly tell whether this driver is worth being factorized with the Versal drivers as I don't known the hardware and the firmwares ABI. I would rather rely on your view.

ldts:
Hi @etienne-lms, no worries—it happens to me all the time.

I think the issue is best understood through a simple example that illustrates my point. Take a look at the PUF registration: [PUF registration code](https://github.com/OP-TEE/optee_os/blob/master/core/drivers/versal_puf.c#L97).

As it can be observed, this specific mailbox request to the PUF firmware service consists of five separate buffers, requiring alignment and cache management for both read and write operations: [Mailbox request handling](https://github.com/OP-TEE/optee_os/blob/master/core/drivers/versal_mbox.c#L293).

If the AMD driver aims to one day support Versal, it will need to implement something similar to what we already have so both drivers will end up being the same on. Just partitioned a bit differently perhaps.

In any case I do not oppose to the creation of new drivers as long as the license is respected (copying and splitting doesn't void the terms of the license WRT to copyright: https://opensource.org/license/bsd-2-clause )

nathan-menhorn:
Hi @Akshay-Belsare  what's the status of this PR?

PR #7272, opened 2025-02-10
mk/compile.mk: fix dependency in .d file when removing a header file

tbourgoi:
Hello,

During tests in our CI we found that the dependency for header file is not correctly managed.
When we remove a header file and it's include from a source file without deleting the build directory the build fails.

You can reproduce the use case by building stm32mp135f-dk on `539836f97 core: arm: virt-aware FF-A thread_foreign_intr_exit()`
And then build `7d9d593df drivers: firewall: stm32_etzpc: remove header file` which remove `stm32_etzpc.h`

With @etienne-lms we found the option `-MP` in gcc and it fix the issue (you need to have the patch for the first build)

### DIFF
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -123,7 +123,7 @@ comp-cppflags-$2 = $$(filter-out $$(CPPFLAGS_REMOVE) $$(cppflags-remove) \
 		      $$(cppflags-lib$$(comp-lib-$2)) $$(cppflags-$2)) \
 		      -D__FILE_ID__=$$(subst -,_,$$(subst /,_,$$(subst .,_,$$(patsubst $$(out-dir)/%,%,$1))))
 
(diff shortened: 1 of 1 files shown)


jforissier:
Nice fix! I have been annoyed by this a few times in the past :wink:

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`. For info, I've tested this change with GCC, not with Clang.

jenswi-linaro:
That's great, that avoids postprocessing the .d like the kernel does (or used to do?).

jforissier:
> `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`. For info, I've tested this change with GCC, not with Clang.

It does work with Clang too (I've tested 12.0.0 and 18.1.7).

tbourgoi:
The 3 review tags have been applied

Issue #7275, opened 2025-02-12
The 1GB VA constraint


markz-zhang:
We hit an issue recently. Here is the configuration info:
- CFG_LPAE_ADDR_SPACE_BITS = 38
- CFG_WITH_PAGER=n
- CFG_CORE_ASLR=n
- CFG_CORE_FFA=y
- CFG_CORE_SEL2_SPMC=y

The issue is when optee boots up and assigns the VA addresses for different memories, the VA range may span two GBs.
Check this debug log as an example:
```
VM 8001: D/TC:00    dump_mmap_table:925 type SHM_VASPACE  va 0x1fff000000..0x2000ffffff pa 0x00000000..0x01ffffff size 0x02000000 (pgdir)
VM 8001: D/TC:00    dump_mmap_table:925 type RES_VASPACE  va 0x2001200000..0x20031fffff pa 0x00000000..0x01ffffff size 0x02000000 (pgdir)
VM 8001: D/TC:00    dump_mmap_table:925 type IO_SEC       va 0x20035f4000..0x2003623fff pa 0x8189850000..0x818987ffff size 0x00030000 (smallpg)
VM 8001: D/TC:00    dump_mmap_table:925 type TA_RAM       va 0x2003624000..0x200521ffff pa 0x2005824000..0x200741ffff size 0x01bfc000 (smallpg)
VM 8001: D/TC:00    dump_mmap_table:925 type IO_SEC       va 0x200538f000..0x200540efff pa 0x80100000..0x8017ffff size 0x00080000 (smallpg)
VM 8001: D/TC:00    dump_mmap_table:925 type IO_SEC       va 0x200540f000..0x200541efff pa 0x08830000..0x0883ffff size 0x00010000 (smallpg)
VM 8001: D/TC:00    dump_mmap_table:925 type IO_SEC       va 0x200541f000..0x200541ffff pa 0x08800000..0x08800fff size 0x00001000 (smallpg)
VM 8001: D/TC:00    dump_mmap_table:925 type TEE_RAM_RO   va 0x2005420000..0x2005423fff pa 0x2005420000..0x2005423fff size 0x00004000 (smallpg)
VM 8001: D/TC:00    dump_mmap_table:925 type TEE_RAM_RX   va 0x2005424000..0x20054a3fff pa 0x2005424000..0x20054a3fff size 0x00080000 (smallpg)
[...]

PR #7276, opened 2025-02-12
core: mmu: Add dynamic VA regions' mapping to page table

markz-zhang:
When optee boots, the initial mapping for MEM_AREA_RES_VASPACE and MEM_AREA_SHM_VASPACE should be added into page tables and replicated to all CPU cores too. This fixes an issue when the VA of MEM_AREA_RES_VASPACE or MEM_AREA_SHM_VASPACE is not in a same 1GB region with other memory regions.

Link: https://github.com/OP-TEE/optee_os/issues/7275

### DIFF
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -1976,7 +1976,14 @@ void core_mmu_map_region(struct mmu_partition *prtn, struct tee_mmap_region *mm)
 				panic("Page is already mapped");
 
 			core_mmu_set_entry(&tbl_info, idx, paddr, mm->attr);
-			paddr += block_size;
+			/*
+			 * Dynamic vaspace regions don't have a physical
+			 * address initially but we need to allocate and
+			 * initialize the translation tables now for later
+			 * updates to work properly.
+			 */
+			if (paddr)
+				paddr += block_size;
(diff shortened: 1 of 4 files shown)


core/mm/core_mmu.c
```diff
@@ -1976,7 +1976,9 @@ void core_mmu_map_region(struct mmu_partition *prtn, struct tee_mmap_region *mm)
 				panic("Page is already mapped");
 
 			core_mmu_set_entry(&tbl_info, idx, paddr, mm->attr);
-			paddr += block_size;
+			/* MEM_AREA_SHM_VASPACE and MEM_AREA_RES_VASPACE don't have PAs initially */
```
jenswi-linaro, line 1979: This line is longer than 80 columns. How about something like:
``` C
/*
 * Dynamic vaspace regions don't have a physical address initially
 * but we need to allocate and initialize the translation tables now
 * for later updates to work properly.
 */

jenswi-linaro:

markz-zhang:
Is there anything else I need to do to get this patch merged?

jforissier:
> Is there anything else I need to do to get this patch merged?

No, I was just waiting for CI to complete. Thanks for contributing!

PR #7279, opened 2025-02-22
riscv: plat-virt: Let console devices be build time configurable

gagachang:
Currently RISC-V virtual platform enforces 16550 UART to be console device. However, there are other console devices which can be chose by developer. Thus, we allow the configurations for console device to be overridden at build time while keeping the default value enabled.

Besides, fix CFG_SBI_CONSOLE to be CFG_RISCV_SBI_CONSOLE.


peterlin-sifive:

PR #7281, opened 2025-02-24
core: do not acknowledge end of interrupt for special GIC interrupt IDs

GseoC:
According to the ARM documentation for GICV2/3/4, there is no need to write to the end of interrupt register for some special IDs. Apply this recommendation to avoid writing to io memory in this time sensitive sequence.

### DIFF
--- a/core/drivers/gic.c
+++ b/core/drivers/gic.c
@@ -948,10 +948,35 @@ static void __maybe_unused gic_native_itr_handler(void)
 	iar = gic_read_iar(gd);
 	id = iar & GICC_IAR_IT_ID_MASK;
 
+	if (id >= 1020 && id <= 1023) {
+		/*
+		 * Special INTIDs
+		 * 1020: Interrupt expected to be handled at SEL1 or SEL2.
+		 *       PE (Processing Element) is either executing at EL3
+		 *       in AArch64 state or in monitor mode in AArch32 state.
+		 *       Reserved on GIC V1 and GIC V2.
+		 * 1021: Interrupt expected to be handled at NSEL1 or NSEL2
+		 *       PE (Processing Element) is either executing at EL3
+		 *       in AArch64 state or in monitor mode in AArch32 state.
+		 *       Reserved on GIC V1 and GIC V2.
+		 * 1022: -(GICv3.3): Interrupt is an NMI
+		 *       -(Legacy): Group 1 interrupt to be signaled to the
+		 *        PE and acknowledged using alias registers. Reserved if
+		 *        interrupt grouping is not supported.
+		 * 1023: No pending interrupt with sufficient priority
+		 *       (spurious) or the highest priority pending interrupt is
+		 *       not appropriate for the current security state or
+		 *       interrupt group.
+		 */
+		DMSG("Special interrupt %"PRIu32, id);
+
+		return;
+	}
+
 	if (id <= gd->max_it)
 		interrupt_call_handlers(&gd->chip, id);
 	else
-		DMSG("ignoring interrupt %" PRIu32, id);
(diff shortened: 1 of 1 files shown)


core/drivers/gic.c
```diff
@@ -950,8 +952,13 @@ static void __maybe_unused gic_native_itr_handler(void)
 
 	if (id <= gd->max_it)
 		interrupt_call_handlers(&gd->chip, id);
+	else if (IS_ENABLED(CFG_ARM_GICV3) && id >= 1020 && id <= 1023)
```
jenswi-linaro, line 955: Please add a comment about these IDs.
I think it would be useful to have a DMSG() even for these IDs since we normally don't receive them and if it happens a lot it might be a sign that something is wrong in the system.

etienne-lms (commented):
In commit message: s/io memory/IO memory/. 
Could you mention the trace message  level change in the commit message? LGTM but others may have concerns on this log level change.

core/drivers/gic.c
```diff
@@ -948,11 +950,34 @@ static void __maybe_unused gic_native_itr_handler(void)
...
+		 * 1023: No pending interrupt with sufficient priority
+		 *       (spurious) or the highest priority pending interrupt is
+		 *       not appropriate for the current security state or
+		 *       interrupt group.
+		 */
+		DMSG("Special interrupt %"PRIu32, id);
+
+		return;
+	} else if (!IS_ENABLED(CFG_ARM_GICV3) &&
+		   id == GICV2_SPURIOUS_INTERRUPT_ID) {
```
jenswi-linaro, line 974: This is a "Special interrupt" too.

core/drivers/gic.c
```diff
@@ -948,11 +950,34 @@ static void __maybe_unused gic_native_itr_handler(void)
 	iar = gic_read_iar(gd);
 	id = iar & GICC_IAR_IT_ID_MASK;
 
-	if (id <= gd->max_it)
+	if (id <= gd->max_it) {
 		interrupt_call_handlers(&gd->chip, id);
-	else
-		DMSG("ignoring interrupt %" PRIu32, id);
+		goto write_eoir;
+	} else if (IS_ENABLED(CFG_ARM_GICV3) && id >= 1020 && id <= 1023) {
```
jenswi-linaro, line 956: It's getting hard to follow all possible flows. The `else` isn't needed. However, I think it would be even better to handle the special interrupts before "if (id <= gd->max_it)" since we're going to return early from the function for those.

core/drivers/gic.c
```diff
@@ -948,11 +950,34 @@ static void __maybe_unused gic_native_itr_handler(void)
...
+		/*
+		 * Special INTIDs
+		 * 1020: Interrupt expected to be handled at SEL1 or SEL2
```
jenswi-linaro, line 959: It might be worth mentioning that it's only received at `EL3`, if I understood correctly.

core/drivers/gic.c
GseoC: Looking back at the documentation, even for GICV1/2, 1020-1023 are reserved IDs even if some are not used. Therefore, I'll stop discriminating GIC version and  try to group everything together.

core/drivers/gic.c
GseoC: Yes indeed, I'll come up with something clearer while applying your suggestion

GseoC:
I grouped the "special interrupts" together, before testing against `max_it`, regardless of the GIC version and detailed what is reserved or not in the description.

jenswi-linaro:

GseoC:
Tag applied, thank you for the review

PR #7283, opened 2025-02-25
drivers: imx: rngb: early initialization

ldts:
The RNGB module must be ready during init_tee_runtime to provide a random stack canary value during bootup.


jforissier (commented):

ldts:
@jforissier thanks, tag added.

PR #7297, opened 2025-03-06
sign_encrypt.py: Fix typo from 'Unkown' to 'Unknown'

AristoChen:
The value of enc_algo_name and flags_name was set to 'Unkown', should be 'Unknown' instead.


jenswi-linaro:

PR #7298, opened 2025-03-06
drivers: amd: Add PS GPIO Support

Akshay-Belsare:
Add PS GPIO Driver support for AMD Platforms.

The PS GPIO Controller is managed through the PS subsystem and can operate in either the Secure World or the Non-Secure World. The driver utilizes the Device Tree Blob (DTB) to determine whether the PS GPIO Controller should be supported in the Secure World.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,6 +274,8 @@ jobs:
           _make PLATFORM=d06
           _make PLATFORM=d06 CFG_HISILICON_ACC_V3=y
           _make PLATFORM=telechips-tcc805x
+          _make PLATFORM=versal2
+          _make PLATFORM=versal2 CFG_AMD_PS_GPIO=y
 
           export ARCH=riscv
           unset CROSS_COMPILE32
--- a/core/drivers/amd/gpio_common.c
+++ b/core/drivers/amd/gpio_common.c
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2002-2021 Xilinx, Inc.  All rights reserved.
+ * Copyright (c) 2022 Foundries.io Ltd. (jorge@foundries.io)
+ * Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#include <assert.h>
+#include <kernel/panic.h>
+#include <malloc.h>
+#include <trace.h>
+
+#include "gpio_private.h"
+
+void amd_gpio_get_bank_and_pin(struct amd_gbank_data *bdata, uint32_t gpio,
+			       uint32_t *bank, uint32_t *pin)
+{
+	uint32_t i = 0;
+
+	assert(gpio < bdata->ngpio);
+
+	for (i = 0; i < bdata->max_bank; i++) {
+		if (gpio >= bdata->bank_min[i] &&
+		    gpio <= bdata->bank_max[i]) {
+			*bank = i;
+			*pin = gpio - bdata->bank_min[i];
+			return;
+		}
+	}
+
+	/* Ideally, should never reach over here */
+	EMSG("Invalid GPIO pin number: %"PRIu32, gpio);
+	panic();
+}
+
+TEE_Result amd_gpio_get_dt(struct dt_pargs *pargs, void *data,
+			   struct gpio **out_gpio)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct amd_gpio_info *amd = (struct amd_gpio_info *)data;
+	struct gpio *gpio = NULL;
+
+	res = gpio_dt_alloc_pin(pargs, &gpio);
+	if (res)
+		return res;
+
+	if (gpio->pin >= amd->bdata->ngpio) {
+		DMSG("GPIO is outside of GPIO Range");
+		free(gpio);
+		return TEE_ERROR_GENERIC;
+	}
+
+	gpio->chip = &amd->chip;
+	*out_gpio = gpio;
+
+	return TEE_SUCCESS;
+}
--- a/core/drivers/amd/ps_gpio_driver.c
+++ b/core/drivers/amd/ps_gpio_driver.c
@@ -1,0 +1,250 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2002-2021 Xilinx, Inc.  All rights reserved.
+ * Copyright (c) 2022 Foundries.io Ltd. (jorge@foundries.io)
+ * Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#include <assert.h>
+#include <drivers/gpio.h>
+#include <io.h>
+#include <kernel/dt.h>
+#include <kernel/panic.h>
+#include <libfdt.h>
+#include <malloc.h>
+#include <mm/core_mmu.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <trace.h>
+#include <util.h>
+
+#include "gpio_private.h"
+
+static struct amd_gbank_data ps_bank = {
+	.label = "ps_gpio",
+	.ngpio = 58,
+	.max_bank = PS_BANK_MAX,
+	.bank_min[0] = 0,
+	.bank_max[0] = 25,
+	.bank_min[3] = 26,
+	.bank_max[3] = 57,
+};
+
+/* Standard GPIO Operations */
+static enum gpio_level ps_gpio_get_value(struct gpio_chip *chip,
+					 unsigned int gpio_pin)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if (io_read32(ps->vbase + DATA_RO_OFFSET(bank)) & BIT(pin))
+		return GPIO_LEVEL_HIGH;
+
+	return GPIO_LEVEL_LOW;
+}
+
+static void ps_gpio_set_value(struct gpio_chip *chip,
+			      unsigned int gpio_pin,
+			      enum gpio_level level)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	uint32_t offset = 0;
+	uint32_t lvl = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if (pin < GPIO_NUM_MAX) {
+		offset = DATA_LSW_OFFSET(bank);
+	} else {
+		pin -= GPIO_NUM_MAX;
+		offset = DATA_MSW_OFFSET(bank);
+	}
+
+	/* Explicitly compare the enum value */
+	if (level == GPIO_LEVEL_HIGH)
+		lvl = 1;
+	else
+		lvl = 0;
+
+	lvl = ~BIT32(pin + GPIO_NUM_MAX) &
+		(SHIFT_U32(lvl, pin) | GPIO_UPPER_MASK);
+
+	io_write32(ps->vbase + offset, lvl);
+}
+
+static enum gpio_dir ps_gpio_get_dir(struct gpio_chip *chip,
+				     unsigned int gpio_pin)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if (io_read32(ps->vbase + DIRM_OFFSET(bank)) & BIT(pin))
+		return GPIO_DIR_OUT;
+
+	return GPIO_DIR_IN;
+}
+
+static void ps_gpio_set_dir(struct gpio_chip *chip,
+			    unsigned int gpio_pin,
+			    enum gpio_dir direction)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if (direction == GPIO_DIR_OUT) {
+		/* set the GPIO pin as output */
+		io_setbits32(ps->vbase + DIRM_OFFSET(bank), BIT(pin));
+
+		/* configure the output enable reg for the pin */
+		io_setbits32(ps->vbase + OUTEN_OFFSET(bank), BIT(pin));
+
+		/* set the state of the pin */
+		ps_gpio_set_value(chip, gpio_pin, GPIO_LEVEL_LOW);
+	} else {
+		/* Set the GPIO pin as input */
+		io_clrbits32(ps->vbase + DIRM_OFFSET(bank), BIT(pin));
+	}
+}
+
+static enum gpio_interrupt ps_gpio_get_intr(struct gpio_chip *chip,
+					    unsigned int gpio_pin)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if ((io_read32(ps->vbase + INTMASK_OFFSET(bank)) & BIT(pin)))
+		return GPIO_INTERRUPT_DISABLE;
+
+	return GPIO_INTERRUPT_ENABLE;
+}
+
+static void ps_gpio_set_intr(struct gpio_chip *chip,
+			     unsigned int gpio_pin,
+			     enum gpio_interrupt interrupt)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	uint32_t offset = 0;
+	uint32_t mask = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	mask = io_read32(ps->vbase + INTMASK_OFFSET(bank)) & BIT(pin);
+
+	/*
+	 * mask = 1 --> Interrupt is masked/disabled.
+	 * mask = 0 --> Interrupt is un-masked/enabled.
+	 */
+	if (mask && interrupt == GPIO_INTERRUPT_ENABLE) {
+		offset = INTEN_OFFSET(bank);
+	} else if (!mask && interrupt == GPIO_INTERRUPT_DISABLE) {
+		offset = INTDIS_OFFSET(bank);
+	} else {
+		DMSG("No change, interrupt already %s",
+		     interrupt ? "Enabled" : "Disabled");
+		return;
+	}
+
+	io_setbits32(ps->vbase + offset, BIT(pin));
+}
+
+static const struct gpio_ops ps_gpio_ops = {
+	.get_direction = ps_gpio_get_dir,
+	.set_direction = ps_gpio_set_dir,
+	.get_value = ps_gpio_get_value,
+	.set_value = ps_gpio_set_value,
+	.get_interrupt = ps_gpio_get_intr,
+	.set_interrupt = ps_gpio_set_intr,
+};
+
+static TEE_Result amd_ps_gpio_probe(const void *fdt, int node,
+				    const void *compat_data __unused)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	int status = DT_STATUS_DISABLED;
+	struct amd_gpio_info *ps_gpio = NULL;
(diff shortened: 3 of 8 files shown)


ldts:
please keep the original copyright from the files you are taking the code from (ie, versal_gpio.c)

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+	status = fdt_get_status(fdt, node);
+
+	/* PS GPIO Controller to be in Non Secure World */
+	if (status & DT_STATUS_OK_NSEC) {
+		DMSG("PS GPIO controller configured for NS world");
+		return TEE_SUCCESS;
+	}
+
+	/* PS GPIO Controller is disabled for Secure World as well*/
+	if (!(status & DT_STATUS_OK_SEC)) {
```
etienne-lms, line 205: This is not strictly needed. If `status & DT_STATUS_OK_SEC == 0`, then we're sure `status & DT_STATUS_OK_SEC != 0` unless what status would be disabled and this driver would not have been probed for that device (see [**dt_driver.c**](https://elixir.bootlin.com/op-tee/4.5.0/source/core/kernel/dt_driver.c#L775-L776))

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+	assert(base);
+	len = ps_gpio->nodeinfo.reg_size;
+	assert(len);
+
+	ps_gpio->vbase = (vaddr_t)core_mmu_add_mapping(MEM_AREA_IO_SEC, base,
+						       len);
+	if (!ps_gpio->vbase) {
+		EMSG("AMD PS GPIO initialization Failed");
+		free(ps_gpio);
+		ps_gpio = NULL;
```
etienne-lms, line 234: You can remove this line.

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+		offset = INT_DIS_OFFSET(bank);
+	} else {
+		IMSG("No change, interrupt already %s",
+		     interrupt ? "Enabled" : "Disabled");
+		return;
+	}
+
+	regval = io_read32(ps->vbase + offset);
+	regval &= BIT(pin);
+	io_write32(ps->vbase + offset, regval);
```
etienne-lms, line 174: It is intended to clear all bits but  `BIT(pin)` that is left to its current value? 
Shouldn't lin 173 rather be `|=` (in which case `io_setbits32(ps->vbase + offset, BIT(pin))` would do the job).

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+	uint32_t mask = 0;
+	uint32_t regval = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	mask = io_read32(ps->vbase + INT_MASK_OFFSET(bank)) & BIT(pin);
+
+	/* mask = 1 --> Interrupt is masked/disabled.
```
etienne-lms, line 159:
```patch
-/* mask = 1 --> Interrupt is masked/disabled.
+/*
+ * mask = 1 --> Interrupt is masked/disabled.

File: core/drivers/amd/gpio_common.c [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+	assert(gpio >= bdata->ngpio);
+
+	for (i = 0; i < bdata->max_bank; i++) {
+		if (gpio >= bdata->bank_min[i] &&
+		    gpio <= bdata->bank_max[i]) {
+			*bank = i;
+			*pin = gpio - bdata->bank_min[i];
+			break;
+		}
+	}
```
etienne-lms, line 28: If the GPIO is not found, the function returns with unmodified `*bank` and `*pin` values. I think panicking or returning an error code would be better.

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if (direction == GPIO_DIR_OUT) {
+		/* set the GPIO pin as output */
+		regval = io_read32(ps->vbase + DIRM_OFFSET(bank));
+		regval |= BIT(pin);
+		io_write32(ps->vbase + DIRM_OFFSET(bank), regval);
```
etienne-lms, line 111: `io_setbits32(ps->vbase + DIRM_OFFSET(bank), BIT(pin)`
Ditto for lines 114/116.

Ditto for line 121/123 (use `io_clrbits32()`)

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+#include <gpio_private.h>
+#include <io.h>
+#include <kernel/dt.h>
+#include <kernel/boot.h>
+#include <kernel/interrupt.h>
+#include <kernel/panic.h>
+#include <libfdt.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <sys/queue.h>
```
etienne-lms, line 18: *kernel/boot.h*, *kernel/interrupt.h* and *sys/queue.h* are not needed.

core/drivers/amd/include/gpio_private.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ */
+
+#ifndef __GPIO_PRIVATE_H__
+#define __GPIO_PRIVATE_H__
+
+#include <drivers/gpio.h>
+#include <kernel/dt.h>
+#include <stdlib.h>
+#include <util.h>
```
etienne-lms, line 13: add `#include <tee_api_types.h>`

core/drivers/amd/gpio_common.c
```diff
@@ -0,0 +1,52 @@
...
+#ifndef CFG_DRIVERS_GPIO
+#error AMD GPIO drivers expects CFG_DRIVERS_GPIO
+#endif
+
+void get_bank_and_pin(struct amd_gbank_data *bdata, uint32_t gpio,
+		      uint32_t *bank, uint32_t *pin)
+{
+	uint32_t i;
+
+	assert(gpio >= bdata->ngpio);
```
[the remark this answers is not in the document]
Akshay-Belsare: yes, let me retest, added this at last.

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,246 @@
...
+	ps_gpio->vbase = (vaddr_t)core_mmu_add_mapping(MEM_AREA_IO_SEC, base,
+						       len);
+	if (!ps_gpio->vbase) {
+		EMSG("AMD PS GPIO initialization Failed");
+		free(ps_gpio);
+		ps_gpio = NULL;
+		return TEE_ERROR_GENERIC;
+	}
+
+	res = gpio_register_provider(fdt, node, amd_gpio_get_dt, ps_gpio);
```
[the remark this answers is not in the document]
etienne-lms: for consistency, yes : release unused allocated memory.

core/drivers/amd/ps_gpio_driver.c
Akshay-Belsare: We require this driver to be explicitly set to a 'disabled' state for regular operation, while maintaining an 'okay' state in secure environments.

Akshay-Belsare:
> please keep the original copyright from the files you are taking the code from (ie, versal_gpio.c)

Apologies, missed that. Now updated.

core/drivers/amd/gpio_common.c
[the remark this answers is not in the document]
Akshay-Belsare: Updated all the required checks. some how only checked negative cases with higher value so did not catch that up.

ldts:
please keep the copyright in the .c file

Akshay-Belsare:
> please keep the copyright in both .c files as well.

Updated, Identifier to be BSD-2-Clause.

I have a query regarding the copyright notices in the C files. Given that the underlying code structures and logic are derived from the original AMD-Xilinx work [gpio-zynq.c](https://github.com/Xilinx/linux-xlnx/blob/master/drivers/gpio/gpio-zynq.c) , what is the specific rationale for including copyright declarations beyond the reused macro definitions?

Akshay-Belsare:
@etienne-lms @jenswi-linaro 
I have addressed the review comments. Could you please take a look?

core/drivers/amd/sub.mk
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+#
+#
+
+incdirs-y += include
+
+ifeq ($(CFG_AMD_PS_GPIO),y)
+srcs-y += gpio_common.c
```
jenswi-linaro, line 10: Please use `srcs-$(CFG_AMD_PS_GPIO)` and drop the `ifeq`.

ldts:
@Akshay-Belsare,

Given the effort involved in porting and validating across different OS environments and a separate project with a distinct use case, Foundries.io retains shared copyright on the original file that served as the basis for this work. Since the modifications are derived from kernel/U-Boot files, we have preserved the original copyright notice.

However, you should not add your own copyright, as you are merely restructuring an existing file within the same project rather than creating a substantially new work. Your current approach to these changes is incorrect.

Instead, you should reorganize the files within your new structure while maintaining the original copyright. Once this restructuring is complete. send your pull request removing the original files (versal_mbox.c and versal_gpio.c) and, if desired, add yourself as a co-maintainer.

Your contributions will be acknowledged through commit authorship, but they do not warrant a copyright change unless they introduce substantial modifications—which is not the case, as your changes are/will be only incremental.

core/drivers/amd/gpio_common.c
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2002-2021 Xilinx, Inc.  All rights reserved.
+ * Copyright (c) 2022 Foundries.io Ltd. (jorge@foundries.io)
+ * Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#include <assert.h>
+#include <gpio_private.h>
```
jenswi-linaro, line 10: It looks like this is only shared between `gpio_common.c` and `ps_gpio_driver.c`.
How about keeping the three files in the same directory and including it as "gpio_private.h"?

core/drivers/amd/gpio_common.c
```diff
@@ -0,0 +1,59 @@
...
+#include <kernel/panic.h>
+
+#ifndef CFG_DRIVERS_GPIO
+#error AMD GPIO drivers expects CFG_DRIVERS_GPIO
```
jenswi-linaro, line 14: This is supposed to be handled in .mk files. If `CFG_DRIVERS_GPIO` isn't configured, what is the worst that can happen? A link error?

jenswi-linaro (commented):
Please add a line in `.github/workflows/ci.yml` to build this in CI.

core/drivers/amd/gpio_common.c
Akshay-Belsare: This is supposed to be handled in .mk files.
Ok, I referenced the existing code to handle this dependency.

 If CFG_DRIVERS_GPIO isn't configured, what is the worst that can happen? A link error?
No error during build but TEE_ERROR_NOT_SUPPORTED at runtime [gpio.h](https://github.com/OP-TEE/optee_os/blob/master/core/include/drivers/gpio.h#L136:L184)

core/drivers/amd/gpio_common.c
Akshay-Belsare: The file is actually at core/drivers/amd/include/gpio_private.h and getting added to include path through "incdirs-y += include" in makefile. Is this not the recommended way?

jenswi-linaro: We do a bit of both, but we try to expose the `*_private.h` files as little as possible to minimize dependency on internal stuff.
``` shell
%ls **/*_private.h
core/arch/arm/kernel/unwind_private.h
core/arch/arm/kernel/vfp_private.h
core/arch/arm/sm/sm_private.h
core/include/kernel/thread_private.h
core/lib/libtomcrypt/src/headers/tomcrypt_private.h
core/lib/qcbor/inc/qcbor/qcbor_private.h
lib/libutee/tee_api_private.h
lib/libutee/tee_socket_private.h
```

Akshay-Belsare: Got it. Thanks!

core/drivers/amd/gpio_common.c
jenswi-linaro: > > This is supposed to be handled in .mk files.
> 
> Ok, I referenced the existing code to handle this dependency.

After some searching, I found the one place you're referring to in `core/drivers/stm32_gpio.c`.

I'm not keen on this kind of dependency checking, it isn't structured and mixes code with configuration.
If it has to be done in code I'd rather do it based on something provided by <drivers/gpio.h>. However, I'd prefer something in a .mk file. @etienne-lms, what do you think?

> 
> > If CFG_DRIVERS_GPIO isn't configured, what is the worst that can happen? A link error?
> 
> No error during build but TEE_ERROR_NOT_SUPPORTED at runtime [gpio.h](https://github.com/OP-TEE/optee_os/blob/master/core/include/drivers/gpio.h#L136:L184)

I see why you want to catch this at build time.

Akshay-Belsare: @jenswi-linaro

This is expected to be 
OPTEE_OS_COMMON_FLAGS += CFG_AMD_PS_GPIO=y  (similar will be for our yocto flow).

While the platform conf has
```
ifeq ($(CFG_AMD_PS_GPIO),y)
$(call force,CFG_MAP_EXT_DT_SECURE,y)
$(call force,CFG_DRIVERS_GPIO,y)
endif
```

If I keep the check in the .mk file, the error occurs immediately during the build. That's when I checked the repository for its handling and referenced the core/drivers/stm32_gpio.c.

Akshay-Belsare: Thanks @jenswi-linaro @etienne-lms! 
Removing the similar check from gpio_common.c as well.

.github/workflows/ci.yml
```diff
@@ -274,6 +274,7 @@ jobs:
           _make PLATFORM=d06
           _make PLATFORM=d06 CFG_HISILICON_ACC_V3=y
           _make PLATFORM=telechips-tcc805x
+          _make PLATFORM=versal2 CFG_AMD_PS_GPIO=y
```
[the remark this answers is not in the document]
Akshay-Belsare: Done!

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,249 @@
...
+	/* Status Check */
+	status = fdt_get_status(fdt, node);
+
+	/* PS GPIO Controller to be in Non Secure World */
+	if (status & DT_STATUS_OK_NSEC) {
+		DMSG("PS GPIO controller configured for NS world");
+		return TEE_SUCCESS;
+	}
+
+	/* PS GPIO Controller is disabled for Secure World as well*/
```
etienne-lms, line 199: Add a space char before `*/`.

etienne-lms (commented):
Few minor comments. I think you can squash all fixup commits.

jenswi-linaro:

nathan-menhorn:
Copyrights are currently under legal review.

michalsimek:
Acked-by: Michal Simek <michal.simek@amd.com>

Akshay-Belsare:
@jenswi-linaro updated the commit message with acked-by tags.

core/drivers/amd/gpio_common.c
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (C) 2002-2021 Xilinx, Inc.  All rights reserved.
+ * Copyright (c) 2022 Foundries.io Ltd. (jorge@foundries.io)
```
nathan-menhorn, line 4: After discussing this with legal there is no copy paste code from the original Versal OP-TEE port. Additionally, the Versal OP-TEE port used drivers from the original Xilinx Linux driver but wasn't exactly copy paste either and did not need a Xilinx copyright either. According to AMD legal department this is only required when code is reused one-for-one. Looking at the other code and using similar ideas doesn't warrant a copyright.

nathan-menhorn (commented):
Comments around copyright.

cc: @ldts @jenswi-linaro @michalsimek @Akshay-Belsare

core/drivers/amd/gpio_common.c
nathan-menhorn: @Akshay-Belsare  That is my understanding unless there is one-for-one code reuse.

nathan-menhorn:
@Akshay-Belsare @michalsimek any updates?

core/drivers/amd/gpio_common.c
ldts: > After discussing this with legal there is no copy paste code from the original Versal OP-TEE port. Additionally, the Versal OP-TEE port used drivers from the original Xilinx Linux driver but wasn't exactly copy paste either and did not need a Xilinx copyright either. According to AMD legal department this is only required when code is reused one-for-one. Looking at the other code and using similar ideas doesn't warrant a copyright.

hi @nathan-menhorn let me check this with Qualcomm and Foundries.io Legal and will come back to you on the next steps.

ldts: Qualcomm/Foundries legal still reviewing, will let you know as soon as they come back to me.

However, from my side and at a personal level (as the author and under EU legislation) I am going to ask that my name is retained on all files _derived_ from the original ones. Let me know how you want to proceed.

ldts: so from my end please keep the copyright + author as that is my understanding, **nak without these changes**

core/drivers/amd/gpio_common.c
```diff
@@ -0,0 +1,56 @@
...
+ * Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#include <assert.h>
+#include <kernel/panic.h>
```
etienne-lms, line 10: add also `trace.h` and `malloc.h`.

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,249 @@
...
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if (direction == GPIO_DIR_OUT) {
+		/* set the GPIO pin as output */
+		io_setbits32(ps->vbase + DIRM_OFFSET(bank), BIT(pin));
```
etienne-lms, line 111: Seen the few  `io_setbits32()`/`io_clrbits32()` calls in this function, I think it needs some kind of concurrent access protection.

Ditto for `ps_gpio_set_intr()`.

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,249 @@
...
+ *
+ */
+
+#include <assert.h>
+#include <drivers/gpio.h>
+#include <io.h>
+#include <kernel/dt.h>
+#include <kernel/panic.h>
+#include <libfdt.h>
+#include <mm/core_mmu.h>
```
etienne-lms, line 15: `include <malloc.h>`

core/drivers/amd/ps_gpio_driver.c
```diff
@@ -0,0 +1,249 @@
...
+					 unsigned int gpio_pin)
+{
+	uint32_t bank = 0;
+	uint32_t pin = 0;
+	struct amd_gpio_info *ps = container_of(chip, struct amd_gpio_info,
+						chip);
+
+	amd_gpio_get_bank_and_pin(ps->bdata, gpio_pin, &bank, &pin);
+
+	if ((io_read32(ps->vbase + DATA_RO_OFFSET(bank)) >> pin) & BIT(0))
```
etienne-lms, line 44: Nitpicking: `io_read32(ps->vbase + DATA_RO_OFFSET(bank)) & BIT(pin)`? 
more consistent with other functions implementation.

Akshay-Belsare:
Please note that the copyright headers for this patch will be preserved in their current state.

@ldts @nathan-menhorn @jenswi-linaro

core/drivers/amd/ps_gpio_driver.c
Akshay-Belsare: @etienne-lms,
Given that the GPIO number is provided as input to operations, and the bank and pin are subsequently identified, it is expected that GPIO-specific access protection to be implemented and managed by the GPIO consumer.

Akshay-Belsare:
@etienne-lms
Review comments are addressed.

etienne-lms (commented):

Akshay-Belsare:
@jenswi-linaro @jforissier 
All tags are applied and change is ready to be merged.

ldts:
Not sure if you need this but here it goes. Thanks for all the follow ups.

`Acked-by: Jorge Ramirez-Ortiz <jorge@foundries.io>`

Akshay-Belsare:
thanks @ldts, will add this tag as well to the commit.
CC : @jforissier

Akshay-Belsare:
> thanks @ldts, will add this tag as well to the commit. CC : @jforissier

Pushed the change with the tag in commit message.
CC: @jforissier

Issue #7296, opened 2025-03-05
STM32MP157a Avenger96 board panic when probing i2c


daniel-lang:
When using the `157A_DHCOR_AVENGER96` flavor of the `stm32mp1` platform, the following error and subsequent panic is shown:

``` 
I/TC: Early console on UART#4
I/TC: 
I/TC: 
I/TC: Pager is enabled. Hashes: 2752 bytes
I/TC: Pager pool size: 92kB
I/TC: Non-secure external DT found
I/TC: Embedded DTB found
I/TC: OP-TEE version: Unknown_4.5 (gcc version 14.2.0) #3 Wed Mar  5 07:30:59 UTC 2025 arm
I/TC: WARNING: This OP-TEE configuration might be insecure!
I/TC: WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html
I/TC: Primary CPU initializing
I/TC: RCC is non-secure
E/TC:0 0 probe_driver_node:522 Failed to probe stm32_i2c on node i2c@40013000: 0xffff0006
E/TC:0 0 probe_driver_node:522 Failed to probe stm32_i2c on node i2c@40012000: 0xffff0006
I/TC: Platform stm32mp1: flavor 157A_DHCOR_AVENGER96 - DT stm32mp157a-dhcor-avenger96.dts
I/TC: DTB enables console
E/TC:0 0 probe_dt_drivers:859 Probe sequence result: 0
E/TC:0 0 print_probe_list:475 |- Driver stm32_i2c on node i2c@40012000 failed
E/TC:0 0 print_probe_list:475 |- Driver stm32_i2c on node i2c@40013000 failed
E/TC:0 0 Panic
``` 
As far as I was enable to analyse the issue, this is due to the fact, that i2c1 and i2c2 are set to `status = "okay"` in `stm32mp15xx-dhcor-avenger96.dtsi` but ` clk-stm32mp15.c` only defines the clocks for i2c 4, 5 and 6.

As I'm new to OP-TEE, I'm not sure what the best way forward it.
Option 1: Disable i2c1 and i2c2 in the device tree.
Option 2: Implement the clock tree for i2c1 and i2c2.

etienne-lms:
Hello @daniel-lang,
[...]

PR #7299, opened 2025-03-06
dts: stm32: disable unused i2c1 and i2c2

daniel-lang:
As discussed in #7296, i2c1 and i2c2 on the STM32MP157a Avenger96 board are enabled because the device tree was copied from the linux kernel.
Both i2c interfaces are not consumed in OP-TEE and the clock tree is missing resulting in a panic on boot when the driver is probed.


GseoC:

jforissier:
@daniel-lang would you please add @GseoC's Acked-by and force push? Thanks!

daniel-lang:
@jforissier, done, thanks for the reminder!

etienne-lms (commented):
My 2 cents, I missed that: could you mention "on Avenger board" in the commit message?
Otherwise `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

daniel-lang:
Good point, I have added it and your Acked-by

PR #7305, opened 2025-03-10
Deprecate libgcc

jenswi-linaro:
Linking with libgcc makes it harder than necessary to test BTI, so make libgcc optional.

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -1284,3 +1284,8 @@ endif
 # handling of external abort implementing the plat_external_abort_handler()
 # function.
 CFG_EXTERNAL_ABORT_PLAT_HANDLER ?= n
+
+# CFG_TA_LIBGCC, when enabled, links user mode TAs with libgcc. Linking
+# TAs with libgcc is deprecated, but keep this flag while sorting out the
+# out remaining issues with supporting C++.
+CFG_TA_LIBGCC ?= y
--- a/core/arch/arm/kernel/link.mk
+++ b/core/arch/arm/kernel/link.mk
@@ -52,7 +52,7 @@ link-objs-init := $(filter-out \
 		    $(out-dir)/$(arch-dir)/kernel/link_dummies_init.o, \
 		    $(objs))
 ldargs-tee.elf := $(link-ldflags) $(link-objs) $(link-out-dir)/version.o \
-		  $(link-ldadd) $(libgcccore)
+		  $(link-ldadd)
 
 link-script-cppflags := \
 	$(filter-out $(CPPFLAGS_REMOVE) $(cppflags-remove), \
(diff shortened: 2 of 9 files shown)


mk/config.mk
```diff
@@ -1284,3 +1284,14 @@ endif
...
+
+# CFG_CORE_LIBGCC, when enabled, links OP-TEE core and ldelf with libgcc.
+# Linking OP-TEE core and ldelf with libgcc is deprecated, but keep this
+# flag while sorting out remaining issues with removing the libgcc
+# dependency.
+CFG_CORE_LIBGCC ?= n
+
+# CFG_TA_LIBGCC, when enabled, links user mode TAs with libgcc. Linking
+# TAs with libgcc is deprecated, but keep this flag while sorting out the
+# out remaining issues with supporting C++.
```
jbech-linaro, line 1296: Yes! We're deprecating the libgcc support. This essentially means that we deprecate the C++ support as well if we're not able to come up with another solution for that. Besides concluding the challenges with C++, I don't think we have a great plan for it right now.

Perhaps we should ask ourselves how many OP-TEE users we believe are using C++. C is the primary language and for additional language support, it might be better to invest time in Rust rather than trying to "retrofit" C++. Thoughts?

mk/config.mk
```diff
@@ -1284,3 +1284,9 @@ endif
 # handling of external abort implementing the plat_external_abort_handler()
 # function.
 CFG_EXTERNAL_ABORT_PLAT_HANDLER ?= n
+
+# CFG_CORE_LIBGCC, when enabled, links OP-TEE core and ldelf with libgcc.
+# Linking OP-TEE core and ldelf with libgcc is deprecated, but keep this
+# flag while sorting out remaining issues with removing the libgcc
+# dependency.
+CFG_CORE_LIBGCC ?= n
```
jforissier, line 1292: What exactly needs `CFG_CORE_LIBGCC=y`? Perhaps we could avoid this CFG altogether?

jforissier (commented):
For "libmbedtls: config: disable long long divisions":
(hoping that there is no significant performance hit?)

jforissier (commented):
For "libutils: asm.S: fix BTI() macro":

jforissier (commented):
For "Deprecate libgcc for TAs":

mk/config.mk
jenswi-linaro: Nothing I'm aware of, I'm only trying to be nice to downstream for a while.

jforissier: That's good but tbh without the introduction of this CFG the patch changes only two lines so it's easy enough to revert.

jenswi-linaro: OK, I can remove it if that's preferred.

jenswi-linaro:
Squashed and tags applied.
@jbech-linaro, I took the liberty to apply your R-B on all the commits after the fixup.

PR #7310, opened 2025-03-13
build: support building on ARM64 host

calebccff:
On ARM64 hosts we don't want to set CROSS_COMPILE when building for arm64. Don't set a fallback value in this case.


jforissier:
Is this really needed? We do have an arm64 build in CI already. Can you give more details on what goes wrong for you without this?

calebccff:
If I just try to run make without this patch on my ARM64 laptop I get

```shell-session
; make PLATFORM=qcom-sc7280
mk/config.mk:909: Warning: Disabling CFG_DRIVERS_CLK_DT [requires CFG_DRIVERS_CLK CFG_DT]
  CHK     out/arm-plat-qcom/conf.mk
  UPD     out/arm-plat-qcom/conf.mk
  CHK     out/arm-plat-qcom/include/generated/conf.h
  UPD     out/arm-plat-qcom/include/generated/conf.h
  CC      out/arm-plat-qcom/core/include/generated/.asm-defines.s
bash: line 1: aarch64-linux-gnu-gcc: command not found
```

jforissier:
If you're running Ubuntu or Debian you just need to install `gcc-aarch64-linux-gnu` (and possibly `gcc-arm-linux-gnueabihf` too for 32-bit arm support). It makes things easier if we can assume we always have a cross compiler by default even if it's not actually cross-compiling. You can always override `CROSS_COMPILE`/`CROSS_COMPILE32`/`CROSS_COMPILE64` depending on what your Linux distribution provides.

calebccff:
My distro (Alpine Linux/postmarketOS) doesn't provide an aarch64 cross compiler for aarch64 hosts (since you'd expect to just use `gcc` and not set CROSS_COMPILE). It isn't possible to override CROSS_COMPILE to be unset.

> It makes things easier if we can assume we always have a cross compiler by default even if it's not actually cross-compiling.

It isn't typical to disallow unsetting CROSS_COMPILE, I can appreciate setting some default value (which is done here) but it doesn't make any sense to do this if we're on an ARM64 device and building FOR ARM64, since we aren't cross compiling...

mk/config.mk
```diff
@@ -18,9 +18,12 @@
 
 # Cross-compiler prefix and suffix
 ifeq ($(ARCH),arm)
+# Don't cross-compile if building on aarch64 natively
+ifneq ($(shell arch),aarch64)
 CROSS_COMPILE ?= arm-linux-gnueabihf-
```
calebccff, line 23: Hmm, I didn't try building for an ARM target on ARM64, I guess it really ought to be like this

```suggestion
CROSS_COMPILE ?= arm-linux-gnueabihf-
ifneq ($(shell arch),aarch64)
```

jforissier:
> My distro (Alpine Linux/postmarketOS) doesn't provide an aarch64 cross compiler for aarch64 hosts (since you'd expect to just use `gcc` and not set CROSS_COMPILE). It isn't possible to override CROSS_COMPILE to be unset.

You can unset `CROSS_COMPILE64` and/or `CROSS_COMPILE32`, it is the two variable that matter in the end. `CROSS_COMPILE` is an alias for `CROSS_COMPILE32`: `make CROSS_COMPILE64=`.

> > It makes things easier if we can assume we always have a cross compiler by default even if it's not actually cross-compiling.
> 
> It isn't typical to disallow unsetting CROSS_COMPILE, I can appreciate setting some default value (which is done here) but it doesn't make any sense to do this if we're on an ARM64 device and building FOR ARM64, since we aren't cross compiling...

Fair enough.

mk/config.mk
jforissier: Some platforms such as `vexpress-qemu_armv8a` need both a 64-bit and 32-bit compiler because 32-bit TAs are supported as well as 64-bit TAs when the OP-TEE OS is 64-bit. And that's where things get messy, because the `gcc` compiler on an arm64 host certainly cannot produce 32-bit arm. IOW using `gcc` instead of `aarch64-linux-gnu-gcc` is only possible if you're not building the 32-bit TA devkit (SDK).
Still, I am OK with your proposed change. It is small enough and makes your life easier so that's fine :wink:

calebccff: Then yeah it should look like what i propose in the suggestion, since we need an ARM cross compiler on ARM64, but not an ARM64 cross compiler.

Thanks

calebccff:
> You can unset CROSS_COMPILE64 and/or CROSS_COMPILE32 on the command line: make CROSS_COMPILE64=.

Oops, I was unsetting `CROSS_COMPILE` :facepalm:. Good to know that's an option. But yeah it would be nice to have this in regardless.

jforissier:

mk/config.mk
```diff
@@ -19,8 +19,11 @@
 # Cross-compiler prefix and suffix
 ifeq ($(ARCH),arm)
 CROSS_COMPILE ?= arm-linux-gnueabihf-
+# Don't cross-compile if building on aarch64 natively
+ifneq ($(shell uname -m),aarch64)
 CROSS_COMPILE64 ?= aarch64-linux-gnu-
 endif
+endif
```
jforissier, line 26: Allowing CROSS_COMPILE64 not to be set breaks the Clang build:
```
$ make PLATFORM=vexpress-qemu_armv8a COMPILER=clang
  GEN     out/arm-plat-vexpress/core/include/generated/arm32_sysreg.h
  GEN     out/arm-plat-vexpress/core/include/generated/arm32_sysreg.S
  CHK     out/arm-plat-vexpress/conf.mk
  UPD     out/arm-plat-vexpress/conf.mk
  CHK     out/arm-plat-vexpress/include/generated/conf.h
  UPD     out/arm-plat-vexpress/include/generated/conf.h
  CC      out/arm-plat-vexpress/core/include/generated/.asm-defines.s
clang: error: unsupported option '-mno-unaligned-access' for target ''
clang: error: unsupported option '-mgeneral-regs-only' for target ''
```
The follwowing patch is needed:
```diff
diff --git a/mk/clang.mk b/mk/clang.mk
index 87f42e88a..ba18c0e1f 100644
--- a/mk/clang.mk
+++ b/mk/clang.mk
@@ -7,13 +7,16 @@ clang-target	:= $(patsubst %-,%,$(notdir $(lastword $(CROSS_COMPILE_$(sm)))))
 ifeq ($(clang-target),aarch64-linux)
 clang-target	:= aarch64-linux-gnu
 endif
+ifneq ($(clang-target),)
+clang-target-opt := --target=$(clang-target)
+endif
 ccache-cmd	:= $(if $(findstring ccache,$(CROSS_COMPILE_$(sm))),$(firstword $(CROSS_COMPILE_$(sm))) ,)
 
-CC$(sm)		:= $(ccache-cmd)$(OPTEE_CLANG_COMPILER_PATH)clang --target=$(clang-target)
+CC$(sm)		:= $(ccache-cmd)$(OPTEE_CLANG_COMPILER_PATH)clang $(clang-target-opt)
 CXX$(sm)	:= false # Untested yet
 # Due to the absence of clang-cpp in AOSP's prebuilt version of clang,
 # use the equivalent command of 'clang -E'
-CPP$(sm)	:= $(ccache-cmd)$(OPTEE_CLANG_COMPILER_PATH)clang --target=$(clang-target) -E
+CPP$(sm)	:= $(ccache-cmd)$(OPTEE_CLANG_COMPILER_PATH)clang $(clang-target-opt) -E
 LD$(sm)		:= $(ccache-cmd)$(OPTEE_CLANG_COMPILER_PATH)ld.lld
 
 AR$(sm)		:= $(ccache-cmd)$(OPTEE_CLANG_COMPILER_PATH)llvm-ar
```

jforissier (changes requested):

jforissier (approved):

PR #7311, opened 2025-03-13
Initial Qualcomm platform support

kcxt:
Introduce initial support for running OP-TEE on Qualcomm platforms, starting with the RB3 Gen 2 which is based on SC7280.

Included is basic support as well as initial support for starting up one of the DSPs using the same SMC API as Qualcomms proprietary secure world.

The TF-A patches for RB3 Gen 2 and OP-TEE support can be found here: https://review.trustedfirmware.org/q/topic:%22caleb/bringup%22

### DIFF
--- a/core/arch/arm/plat-qcom/conf.mk
+++ b/core/arch/arm/plat-qcom/conf.mk
@@ -1,0 +1,35 @@
+# Qualcomm platform support
+
+PLATFORM_FLAVOR ?= sc7280
+
+$(call force,CFG_GIC,y)
+$(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
+$(call force,CFG_ARM64_core,y)
+$(call force,CFG_WITH_ARM_TRUSTED_FW,y)
+$(call force,CFG_CORE_ARM64_PA_BITS,40)
+$(call force,CFG_DRIVERS_CLK,y)
+$(call force,CFG_DRIVERS_CLK_DT,y)
+$(call force,CFG_DRIVERS_QCOM_CLK,y)
+$(call force,CFG_CORE_LARGE_PHYS_ADDR,y)
+ta-targets = ta_arm64
+supported-ta-targets ?= ta_arm64
+
+CFG_DT ?= n
+CFG_DTB_MAX_SIZE ?= 0x40000
+CFG_NUM_THREADS ?= 8
+
+ifeq ($(PLATFORM_FLAVOR),sc7280)
+include core/arch/arm/cpu/cortex-armv8-0.mk
+$(call force,CFG_TEE_CORE_NB_CORE,8)
+$(call force,CFG_ARM_GICV3,y)
+$(call force,CFG_GENI_UART,y)
+
+CFG_TZDRAM_START ?= 0x1c120000
+CFG_TZDRAM_SIZE  ?= 0x00ee0000
+CFG_SHMEM_START  ?= 0x1d000000
+CFG_SHMEM_SIZE   ?= 0x00400000
+
+CFG_EARLY_CONSOLE ?= y
+
+CFG_TEE_CORE_LOG_LEVEL ?= 2
+endif
--- a/core/arch/arm/plat-qcom/main.c
+++ b/core/arch/arm/plat-qcom/main.c
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2025, Linaro Limited
+ */
+
+#include <console.h>
+#include <drivers/geni_uart.h>
+#include <drivers/clk_qcom.h>
+#include <kernel/boot.h>
+#include <kernel/panic.h>
+#include <io.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <stdint.h>
+#include <tee/entry_fast.h>
+#include <tee/entry_std.h>
+#include <drivers/gic.h>
+#include <sm/optee_smc.h>
+#include <kernel/delay.h>
+
+#include "q6dsp.h"
+
+/*
+ * Register the physical memory area for peripherals etc. Here we are
+ * registering the UART console.
+ */
+register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE,
+			GENI_UART_REG_SIZE);
+register_phys_mem(MEM_AREA_IO_NSEC, GCC_BASE, 0x100000);
+register_phys_mem(MEM_AREA_IO_NSEC, WPSS_BASE, 0x1000);
+register_phys_mem(MEM_AREA_IO_NSEC, ADSP_BASE, 0x1000);
+#ifdef ADSP_LPASS_EFUSE
+/* We only need the one 32-bit register... */
+register_phys_mem(MEM_AREA_IO_NSEC, ADSP_LPASS_EFUSE, 0x1000);
+#endif
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICR_BASE, 0x100000);
+
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+register_ddr(DRAM1_BASE, DRAM1_SIZE);
+register_ddr(DRAM2_BASE, DRAM2_SIZE);
+register_ddr(DRAM3_BASE, DRAM3_SIZE);
+
+static struct geni_uart_data console_data;
+
+void plat_console_init(void)
+{
+	geni_uart_init(&console_data, CONSOLE_UART_BASE);
+	register_serial_console(&console_data.chip);
+}
+
+void boot_primary_init_intc(void)
+{
+	gic_init_v3(0, GICD_BASE, GICR_BASE);
+}
+
+void boot_secondary_init_intc(void)
+{
+	gic_init_per_cpu();
+}
+
+#define SMC_SVC_PIL 0x02
+#define QTI_SIP_SVC_AVAILABLE_ID	U(0xc2000601)
+#define QCOM_SCM_PIL_PAS_INIT_IMAGE	U(0xc2000201)
+#define QCOM_SCM_PIL_PAS_MEM_SETUP	U(0xc2000202)
+#define QCOM_SCM_PIL_PAS_AUTH_AND_RESET	U(0xc2000205)
+#define QCOM_SCM_PIL_PAS_SHUTDOWN	U(0xc2000206)
+#define QCOM_SCM_PIL_PAS_IS_SUPPORTED	U(0xc2000207)
+
+static bool qti_check_syscall_availability(uint64_t smc_fid)
+{
+	/*
+	 * Ignore the top two bytes when checking if an SMC call
+	 * is supported since those aren't set.
+	 */
+	smc_fid = (smc_fid & 0x3fffffff) | 0xc0000000;
+	switch (smc_fid) {
+	case QTI_SIP_SVC_AVAILABLE_ID:
+	case QCOM_SCM_PIL_PAS_INIT_IMAGE:
+	case QCOM_SCM_PIL_PAS_MEM_SETUP:
+	case QCOM_SCM_PIL_PAS_AUTH_AND_RESET:
+	case QCOM_SCM_PIL_PAS_SHUTDOWN:
+	case QCOM_SCM_PIL_PAS_IS_SUPPORTED:
+		return true;
+	default:
+		return false;
+	}
+}
+
+/* Add handlers for DSP init */
+void tee_entry_fast(struct thread_smc_args *args)
+{
+	/* Mask off the function ID */
+	uint32_t smc = args->a0 & 0xFFFFFF00;
+
+	if (args->a0 == QTI_SIP_SVC_AVAILABLE_ID) {
+		args->a0 = OPTEE_SMC_RETURN_OK;
+		args->a1 = qti_check_syscall_availability(args->a2) ? 1 : 0;
+		return;
+	}
+
+	/* SIP call for PIL svc */
+	if (smc == 0xc2000200)
+		qcom_handle_pil_smc(args);
+	else
+		__tee_entry_fast(args);
+}
--- a/core/arch/arm/plat-qcom/q6dsp.c
+++ b/core/arch/arm/plat-qcom/q6dsp.c
@@ -1,0 +1,221 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2025, Linaro Limited
+ */
+
+#include <console.h>
+#include <drivers/geni_uart.h>
+#include <drivers/clk_qcom.h>
+#include <kernel/boot.h>
+#include <kernel/panic.h>
+#include <io.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <stdint.h>
+#include <tee/entry_fast.h>
+#include <drivers/gic.h>
+#include <sm/optee_smc.h>
+#include <kernel/delay.h>
+
+#include "q6dsp.h"
+
+#define WPSS_MEM_PHYS 0x84800000
+
+/*QDSP6SS register offsets*/
+#define RST_EVB_REG			0x10
+#define CORE_START_REG			0x400
+#define BOOT_CMD_REG			0x404
+#define BOOT_STATUS_REG			0x408
+#define RET_CFG_REG			0x1C
+
+#define QDSP6SS_XO_CBCR		0x38
+#define QDSP6SS_CORE_CBCR	0x20
+#define QDSP6SS_SLEEP_CBCR	0x3c
+#define LPASS_BOOT_CORE_START	BIT(0)
+#define LPASS_BOOT_CMD_START	BIT(0)
+#define LPASS_EFUSE_Q6SS_EVB_SEL 0x0
+
+#define BOOT_FSM_TIMEOUT		10000
+
+#if defined(PLATFORM_FLAVOR_sc7280)
+#define PAS_ID_WPSS 0x06
+static struct qcom_q6dsp_data wpss_dsp_data = {
+	.pas_id = PAS_ID_WPSS,
+	.base.pa = WPSS_BASE,
+	.clk_group = QCOM_CLKS_WPSS,
+};
+#endif
+
+static void sc7280_dsp_start(struct thread_smc_args *args,
+			     struct qcom_q6dsp_data *selected)
+{
+	uint32_t val;
+	uint64_t timer;
+	vaddr_t base = io_pa_or_va(&selected->base, 0x500);
+	vaddr_t lpass_efuse = 0;
+
+	if (selected->lpass_efuse_base.pa)
+		lpass_efuse = io_pa_or_va(&selected->lpass_efuse_base, 0x10);
+
+	/* Enable the XO clock */
+	io_write32(base + QDSP6SS_XO_CBCR, 1);
+
+	/* Enable the QDSP6SS sleep clock */
+	io_write32(base + QDSP6SS_SLEEP_CBCR, 1);
+
+	/* Enable the QDSP6 core clock */
+	io_write32(base + QDSP6SS_CORE_CBCR, 1);
+
+	/* Program boot address */
+	io_write32(base + RST_EVB_REG, selected->firmware_base >> 4);
+
+	if (lpass_efuse)
+		io_write32(lpass_efuse, LPASS_EFUSE_Q6SS_EVB_SEL);
+
+	/* Flush configuration */
+	dsb();
+
+	/* De-assert QDSP6 stop core. QDSP6 will execute after out of reset */
+	io_write32(base + CORE_START_REG, LPASS_BOOT_CORE_START);
+
+	/* Trigger boot FSM to start QDSP6 */
+	io_write32(base + BOOT_CMD_REG, LPASS_BOOT_CMD_START);
+
+	/* Wait for core to come out of reset */
+	timer = timeout_init_us(BOOT_FSM_TIMEOUT);
+	do {
+		val = io_read32(base + BOOT_STATUS_REG);
+		if (val & BIT(0))
+			break;
+		if (timeout_elapsed(timer))
+			break;
+		udelay(10);
+	} while (1);
+
+	if ((val & BIT(0)) == 0) {
+		EMSG("Timed out waiting for DSP to boot :(");
+		args->a0 = QCOM_SCM_ERROR;
+		return;
+	}
+
+	args->a0 = OPTEE_SMC_RETURN_OK;
+	args->a1 = 0;
+	args->a2 = 0;
+	args->a3 = 0;
+	args->a4 = 0;
+	args->a5 = 0;
+}
+
+#define QCOM_SCM_PIL_PAS_INIT_IMAGE	0x0201
+#define QCOM_SCM_PIL_PAS_MEM_SETUP	0x0202
+#define QCOM_SCM_PIL_PAS_AUTH_AND_RESET	0x0205
+#define QCOM_SCM_PIL_PAS_SHUTDOWN	0x0206
+#define QCOM_SCM_PIL_PAS_IS_SUPPORTED	0x0207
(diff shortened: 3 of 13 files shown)


core/drivers/geni_uart.c
```diff
@@ -0,0 +1,250 @@
+// SPDX-License-Identifier: GPL-2.0+
```
ricardosalveti, line 1: This will be an issue as it will force the final build to be GPLv2 based, other GPL files are dual licensed with a more permissive license, such as BSD-2 (main one used by optee).

core/drivers/clk/qcom/clock-qcom.c
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: GPL-2.0
```
ricardosalveti, line 1: Same.

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,162 @@
...
+	switch (pas_id) {
+	case PAS_ID_WPSS: return "wpss";
+	default: return "???";
+	}
+}
+
+void qcom_handle_pil_smc(struct thread_smc_args *args)
+{
+	uint32_t sv_cmd = args->a0 & 0xffff;
+	DMSG("Handling %s cmd for %s\n", smc_decode_pil_cmd(sv_cmd), pas_id_name(args->a2));
```
jenswi-linaro, line 127: There's no need for a `\n` at the end for the `*MSG()` macros. Please remove all the redundant `\n` in this PR.

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,69 @@
...
+}
+
+#define SMC_SVC_PIL 0x02
+
+/* Add handlers for DSP init */
+void tee_entry_fast(struct thread_smc_args *args)
+{
+	uint32_t svc = (args->a0 >> 8) & 0xff;
+
+	if (svc == SMC_SVC_PIL) {
```
jenswi-linaro, line 64: You also need to check the upper 16 bits.
The extra `{}` are redundant, please remove them.

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,69 @@
...
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+register_ddr(DRAM1_BASE, DRAM1_SIZE);
+register_ddr(DRAM2_BASE, DRAM2_SIZE);
+register_ddr(DRAM3_BASE, DRAM3_SIZE);
+
+static struct geni_uart_data console_data;
+
+void plat_console_init(void)
+{
+	geni_uart_init(&console_data, CONSOLE_UART_BASE);
```
jenswi-linaro, line 43: This function is introduced in a later patch. Each patch should compile as they are added.

jenswi-linaro:
Please fix the checkpatch issues. You can check them offline using `scripts/checkpatch.sh github/master..`, assuming that your patches are based on `github/master`.

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,69 @@
...
+#include <kernel/panic.h>
+#include <io.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <stdint.h>
+#include <tee/entry_fast.h>
+#include <tee/entry_std.h>
+#include <drivers/gic.h>
+#include <sm/optee_smc.h>
+#include <kernel/delay.h>
```
etienne-lms, line 19: Please sort in alphabetical order.

This comment applies to a few other source files.

core/arch/arm/plat-qcom/main.c
calebccff: Is it necessary to check bit 31 or the MBZ bits?

jenswi-linaro: Yes

core/drivers/geni_uart.c
calebccff: The code here and for the clock are both based on U-Boot drivers which are licensed this way, and I don't have permission from the original authors to relicense them.

jenswi-linaro: It must have a permissive license or we can't take it.

core/arch/arm/plat-qcom/conf.mk
```diff
@@ -0,0 +1,42 @@
...
+$(call force,CFG_DRIVERS_CLK,y)
+$(call force,CFG_DRIVERS_CLK_DT,y)
+$(call force,CFG_DRIVERS_QCOM_CLK,y)
+ta-targets = ta_arm64
+supported-ta-targets ?= ta_arm64
+
+CFG_CORE_LARGE_PHYS_ADDR := y
+
+CFG_DT ?= n
+CFG_DTB_MAX_SIZE := 0x40000
```
jforissier, line 20: Either `?=` or `$(call force...)`. See [here](https://optee.readthedocs.io/en/latest/building/gits/optee_os.html#there-are-only-two-ways).

core/arch/arm/plat-qcom/conf.mk
```diff
@@ -0,0 +1,42 @@
...
+CFG_FTRACE_SUPPORT := y
+CFG_SYSCALL_FTRACE := y
+CFG_TA_ASLR := 0
```
jforissier, line 23: `n` not `0`. What's the problem with ASLR?

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,165 @@
...
+	do {
+		val = io_read32(base + BOOT_STATUS_REG);
+		if ((val & BIT(0)) != 0)
+			break;
+		if (timeout_elapsed(timer))
+			break;
+		udelay(10);
+	} while (0);
+
+	if ((val & BIT(0)) != 0) {
```
jforissier, line 84: `if (val & BIT(0))` is preferred

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,165 @@
...
+	io_write32(base + CORE_START_REG, LPASS_BOOT_CORE_START);
+
+	/* Trigger boot FSM to start QDSP6 */
+	io_write32(base + BOOT_CMD_REG, LPASS_BOOT_CMD_START);
+
+	/* Wait for core to come out of reset */
+	timer = timeout_init_us(BOOT_FSM_TIMEOUT);
+	do {
+		val = io_read32(base + BOOT_STATUS_REG);
+		if ((val & BIT(0)) != 0)
```
jforissier, line 77: `if (val & BIT(0))`

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,165 @@
...
+			break;
+		if (timeout_elapsed(timer))
+			break;
+		udelay(10);
+	} while (0);
```
jforissier, line 82: You mean `while (1);` I suppose. I have a slight preference for `for (;;)`

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,165 @@
...
+static inline const char *smc_decode_pil_cmd(uint32_t cmd)
+{
+	switch (cmd) {
+	case QCOM_SCM_PIL_PAS_INIT_IMAGE: return "init_image";
+	case QCOM_SCM_PIL_PAS_MEM_SETUP: return "mem_setup";
+	case QCOM_SCM_PIL_PAS_AUTH_AND_RESET: return "auth_and_reset";
+	case QCOM_SCM_PIL_PAS_SHUTDOWN: return "shutdown";
+	case QCOM_SCM_PIL_PAS_IS_SUPPORTED: return "is_supported";
+	case QCOM_SCM_PIL_PAS_MSS_RESET: return "mss_reset";
+	default:  return "???";
```
jforissier, line 113: Style: prefer with a line break after the case statement.

jforissier (commented):
Comments on "Qualcomm platforms support".
Please cite the boards in the subject ("Qualcomm SC7280/QCM6490 support") and see below. Thanks!

jforissier (commented):
For "build: support building on ARM64 host":

Edit: "build: do not set CROSS_COMPILE* when building on ARM64 host" would be a better subject IMO

calebccff:
Hi @jforissier thanks for the review! I'll add your ack in https://github.com/OP-TEE/optee_os/pull/7310

core/arch/arm/plat-qcom/conf.mk
```diff
@@ -0,0 +1,42 @@
+# Qualcomm platform support
+
+PLATFORM_FLAVOR ?= sc7280
+
+$(call force,CFG_GIC,y)
+$(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
+$(call force,CFG_8250_UART,y)
```
stephan-gh, line 7: Is this needed? Pretty sure QC doesn't have 8250-based UART. :-)

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,165 @@
...
+		wpss_dsp_data.firmware_base = args->a3;
+		break;
+	case QCOM_SCM_PIL_PAS_AUTH_AND_RESET:
+		qcom_clock_enable(QCOM_CLKS_WPSS);
+		kodiak_wpss_start(args);
+		break;
+	case QCOM_SCM_PIL_PAS_SHUTDOWN:
+	case QCOM_SCM_PIL_PAS_IS_SUPPORTED:
+	case QCOM_SCM_PIL_PAS_MSS_RESET:
+		break;
```
stephan-gh, line 153: Please make sure you don't handle these with an OK return status if they're not implemented. We have enough broken firmware versions already. :sweat_smile: 

- SHUTDOWN should return some "not implemented" error code.
- You can probably just remove the MSS_RESET one, I think this was pretty much just used on good old msm8916.

Keeping IS_SUPPORTED here is okay I guess (should just return OK), but maybe add a comment that it's only okay because the if statement further above already checks for `if (args->a2 != PAS_ID_WPSS)`.

b49020:
Glad to see support for Qcom devices coming up, you should also add some documentation here: https://github.com/OP-TEE/optee_docs/tree/master/building/devices as to how one can bring up OP-TEE on Qcom devices.

ldts:
> Glad to see support for Qcom devices coming up, you should also add some documentation here: https://github.com/OP-TEE/optee_docs/tree/master/building/devices as to how one can bring up OP-TEE on Qcom devices.

here as well would be really useful for regression testing https://github.com/OP-TEE/build

core/arch/arm/plat-qcom/q6dsp.c
calebccff: * SHUTDOWN now returns -ENOTAVAIL
* Dropped MSS_RESET and IS_SUPPORTED since they're specific to the old WCNSS driver anyways

And yeah cleaned up the logic so we're not misbehaving

stephan-gh: > Dropped MSS_RESET and IS_SUPPORTED since they're specific to the old WCNSS driver anyways

I don't think IS_SUPPORTED is specific to WCNSS, even though it's only used by that driver in upstream Linux at the moment. If I remember correctly, QC firmware still implements it nowadays.

This call would be useful to have if someone wanted to add runtime selection of PAS vs direct booting in Linux. Can you try implementing it? It should be straightforward, just add a case for it and set `a1 = true`. (Ideally test it by adding `qcom_scm_pas_supported(PAS_ID_WPSS)` somewhere in Linux to be sure...)

kcxt: OK I re-added the IS_SUPPORTED call and validated it from the PAS driver, even though it's unused it should work as expected if that changes.

core/drivers/clk/qcom/clock-qcom.c
```diff
@@ -0,0 +1,50 @@
...
+/* CBCR register fields */
+#define CBCR_BRANCH_ENABLE_BIT  BIT(0)
+#define CBCR_BRANCH_OFF_BIT     BIT(31)
+
+/* Enable clock controlled by CBC soft macro */
+static void clk_enable_cbc(paddr_t cbcr)
+{
+	io_write32(cbcr, CBCR_BRANCH_ENABLE_BIT);
+
+	while (io_read32(cbcr) & CBCR_BRANCH_OFF_BIT)
```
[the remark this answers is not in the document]
kcxt: Thanks, added that so we don't risk getting totally trapped in OP-TEE heh

core/drivers/geni_uart.c
```diff
@@ -0,0 +1,249 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Qualcomm GENI serial engine UART driver
+ *
+ * Copyright (c) 2021 Dzmitry Sankouski <dsankouski@gmail.com>
+ * Copyright (c) 2024, Linaro Limited
+ *
+ * Based on U-Boot driver
```
jenswi-linaro, line 8: Looking at https://github.com/u-boot/u-boot/blob/master/drivers/serial/serial_msm_geni.c, it says "Based on Linux driver". In other words, it's derived from. When comparing the U-Boot driver with this driver it's obvious that it has been copied and hacked up to fit here. But it's now as clear to me how the U-Boot driver relates to the kernel driver. Is it also copied and hacked up to fit into U-Boot, or did it only serve as inspiration and might not be considered derived work? 

If we change the license, we should preferably show that we have permission from the copyright holder. It would be nice with an eventual link to a public mail in the commit message. But it should at least be mentioned in the commit message that you asked the author and got an OK.

We should have enough information in the git history to answer how we can use this license after an eventual future Black Duck scan or similar.

jenswi-linaro:
Please add a comment here when you push changes, since github often doesn't send out updates for that.

core/drivers/geni_uart.c
kcxt: The UART driver has been re-written from scratch now based on the TF-A driver which is also under a permissive license.

b49020 (changes requested):
Need to add a proper OP-TEE based platform specific pseudo TA service instead of supporting random SMC calls.

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,108 @@
...
+
+	if (args->a0 == QTI_SIP_SVC_AVAILABLE_ID) {
+		args->a0 = OPTEE_SMC_RETURN_OK;
+		args->a1 = qti_check_syscall_availability(args->a2) ? 1 : 0;
+		return;
+	}
+
+	/* SIP call for PIL svc */
+	if (smc == 0xc2000200)
+		qcom_handle_pil_smc(args);
```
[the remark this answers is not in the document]
jenswi-linaro: I don't see how this violates the SMC calling convention. We already support platform-specific SMCs, and I don't see what difference it makes whether they are in the SIP range or another range.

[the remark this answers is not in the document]
b49020: Yeah we do have platform specific SMCs but those were only for ARMv7 platforms. I could find an example here: `core/arch/arm/plat-zynq7k/main.c +204`. I suppose that was only because the secure monitor was bundled into OP-TEE OS binary. Do we have any existing example with TF-A?

Essentially with TF-A the secure monitor is a separate entity and that was only done to decouple responsibilities of Trusted OS and secure monitor. Now if we are going again on a similar path to intermix SMC calls handling among TF-A and OP-TEE then we are defeating that purpose.

Following is the screenshot from https://developer.arm.com/documentation/den0028/latest/, showing SMC owning entity number for Trusted OS and Trusted Applications.

image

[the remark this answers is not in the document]
b49020: We already have proper support for OP-TEE platform services exposed over TEE bus in the Linux kernel. The firmware authentication and loading service being implemented here using special SIP SMCs which aren't meant for Trusted OS (OP-TEE) to handle can rather be implemented as a proper pseudo TA service. `stm32mp` provides a good reference for such a service: https://github.com/OP-TEE/optee_os/blob/master/core/pta/stm32mp/remoteproc_pta.c

core/drivers/geni_uart.c
```diff
@@ -0,0 +1,60 @@
...
+	struct geni_uart_data *pd =
+		container_of(chip, struct geni_uart_data, chip);
+
+	return io_pa_or_va(&pd->base, GENI_UART_REG_SIZE);
+}
+
+static void wait_tx_done(vaddr_t base)
+{
+	uint64_t timer = timeout_init_us(1000 * 1000);
+	while (io_read32(base + GENI_STATUS_REG) & GENI_STATUS_REG_CMD_ACTIVE) {
```
etienne-lms, line 34: Add an empty line after local variable definition.

core/include/drivers/geni_uart.h
```diff
@@ -0,0 +1,21 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024, Linaro Limited
```
etienne-lms, line 3: 2025?

core/drivers/clk/qcom/clock-qcom.c
```diff
@@ -0,0 +1,50 @@
...
+
+#include <drivers/clk.h>
+#include <drivers/clk_qcom.h>
+#include <libfdt.h>
+#include <malloc.h>
+#include <stdint.h>
+#include <io.h>
+#include <mm/core_memprot.h>
+#include <trace.h>
+#include <platform_config.h>
```
etienne-lms, line 14: Please sort in alphabetical order.
Some header files could be removed from this list: *drivers/clk.h*, *libfdt.h* and *malloc.h*.

core/include/drivers/clk_qcom.h
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2025, Linaro Limited
+ */
+
+#ifndef _CLK_QCOM_H_
+#define _CLK_QCOM_H_
```
etienne-lms, line 7: `_DRIVERS_CLK_QCOM_H_`

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,221 @@
...
+	.base.pa = WPSS_BASE,
+	.clk_group = QCOM_CLKS_WPSS,
+};
+#endif
+
+static void sc7280_dsp_start(struct thread_smc_args *args,
+			     struct qcom_q6dsp_data *selected)
+{
+	uint32_t val;
+	uint64_t timer;
```
etienne-lms, line 53: `= 0` for each of the 2 lines above.

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,221 @@
...
+}
+
+#define QCOM_SCM_PIL_PAS_INIT_IMAGE	0x0201
+#define QCOM_SCM_PIL_PAS_MEM_SETUP	0x0202
+#define QCOM_SCM_PIL_PAS_AUTH_AND_RESET	0x0205
+#define QCOM_SCM_PIL_PAS_SHUTDOWN	0x0206
+#define QCOM_SCM_PIL_PAS_IS_SUPPORTED	0x0207
+#define QCOM_SCM_PIL_PAS_MSS_RESET	0x020a
+
+static inline const char *smc_decode_pil_cmd(uint32_t cmd)
```
etienne-lms, line 116: Prefer without the `inline` keyword.
Ditto at line 134.

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,221 @@
...
+	default:
+		return "???";
+	}
+}
+
+void qcom_handle_pil_smc(struct thread_smc_args *args)
+{
+	uint32_t sv_cmd = args->a0 & 0xffff;
+	struct qcom_q6dsp_data *selected;
+	TEE_Result res;
```
etienne-lms, line 148:
```c
	struct qcom_q6dsp_data *selected = NULL;

File: core/arch/arm/plat-qcom/q6dsp.c [hunk 25]
```diff
@@ -0,0 +1,221 @@
...
+		 */
+		args->a0 = OPTEE_SMC_RETURN_OK;
+		args->a1 = 1;
+		return;
+	case QCOM_SCM_PIL_PAS_INIT_IMAGE:
+		break;
+	case QCOM_SCM_PIL_PAS_MEM_SETUP:
+		/* Save base address of firmware */
+		selected->firmware_base = args->a3;
+		IMSG("Got firmware base: 0x%lx\n", selected->firmware_base);
```
etienne-lms, line 186: Nitpicking: prefer without `\n` termination character.
Ditto at line 204.

core/arch/arm/plat-qcom/q6dsp.c
```diff
@@ -0,0 +1,221 @@
...
+		res = qcom_clock_enable(selected->clk_group);
+		if (res != TEE_SUCCESS) {
+			EMSG("Failed to enable clocks: %d", res);
+			args->a0 = QCOM_SCM_EINVAL_ADDR;
+			return;
+		}
+		sc7280_dsp_start(args, selected);
+		DMSG("DSP start done!\n");
+		selected = NULL;
+	} break;
```
etienne-lms, line 206:
```patch
-	case QCOM_SCM_PIL_PAS_AUTH_AND_RESET: {
+	case QCOM_SCM_PIL_PAS_AUTH_AND_RESET:
		/*
		 * Can't boot the DSP if we didn't get the firmware
		 * address yet!
		 */
		(...)
-	} break;
+		break;

File: core/arch/arm/plat-qcom/q6dsp.c [hunk 27]
```diff
@@ -0,0 +1,221 @@
...
+		 * Can't boot the DSP if we didn't get the firmware
+		 * address yet!
+		 */
+		if (!selected->firmware_base) {
+			args->a0 = QCOM_SCM_EINVAL_ADDR;
+			return;
+		}
+		res = qcom_clock_enable(selected->clk_group);
+		if (res != TEE_SUCCESS) {
+			EMSG("Failed to enable clocks: %d", res);
```
etienne-lms, line 199:
```suggestion
			EMSG("Failed to enable clocks: %#"PRIx32, res);

File: core/drivers/clk/qcom/clock-qcom.c [hunk 28]
```diff
@@ -18,12 +18,20 @@
 #define CBCR_BRANCH_OFF_BIT     BIT(31)
 
 /* Enable clock controlled by CBC soft macro */
-static void clk_enable_cbc(paddr_t cbcr)
+static TEE_Result clk_enable_cbc(paddr_t cbcr)
 {
+	uint64_t timer;
```
etienne-lms, line 23: `= 0`

core/drivers/clk/qcom/clock-qcom.c
```diff
@@ -18,12 +18,20 @@
...
+
 	io_write32(cbcr, CBCR_BRANCH_ENABLE_BIT);
 
-	while (io_read32(cbcr) & CBCR_BRANCH_OFF_BIT)
-		;
+	timer = timeout_init_us(10000 * 1000);
```
etienne-lms, line 27: Using a well named macro would be nice, e.g. `TIMEOUT_10S_US`.

core/drivers/clk/qcom/clock-qcom.c
```diff
@@ -18,12 +18,20 @@
...
+	do {
+		if (!(io_read32(cbcr) & CBCR_BRANCH_OFF_BIT))
+			return TEE_SUCCESS;
+		if (timeout_elapsed(timer))
+			return TEE_ERROR_TIMEOUT;
+		udelay(10);
```
etienne-lms, line 33: Add `#include <kernel/delay.h>` at source file top.

core/drivers/clk/qcom/clock-qcom.c
```diff
@@ -35,12 +43,19 @@ TEE_Result qcom_clock_enable(enum qcom_clk_group group)
 {
 	struct io_pa_va base = { .pa = GCC_BASE };
 	vaddr_t gcc_base = io_pa_or_va(&base, 0x100000);
+	TEE_Result res;
```
etienne-lms, line 46: `= TEE_ERROR_GENERIC`

PR #7312, opened 2025-03-13
Extending UBSan logic to trusted applications and ldelf

alexhoppus:
Hi,

**What is this about:**

with this change internal **trusted applications** (located in optee os repo) and **ldelf** are instrumented with undefined sanitizer checks (-fsanitize=undefined) if **CFG_CORE_SANITIZE_UNDEFINED** is enabled. Linked libraries are instrumented as well. The change also adds logic which suppresses redundant ubsan reports for the same source location if panic mode is disabled.

**Some testing:**

Running `xtest -l 15` gives several new ubsan warnings, from the first glance all of them are minor or false positives and couldn't lead to real problems. However it makes sense to look at them closer. 

```
/* Dst or src for memcpy/memove is NULL the size is also 0. Surprisingly big amount
   of these occurs in ftpm context. Probably it makes sense to check why. */
E/TA:  Undefined behavior nonnull_arg at lib/libutee/tee_api_arith_mpi.c:114 col 3
E/TA:  Undefined behavior nonnull_arg at lib/libutee/tee_api_arith_mpi.c:74 col 2
E/TA:  Undefined behavior nonnull_arg at lib/libutee/tee_api.c:612 col 2
/* Probably alignment issues. Briefly looked under gdb and it seems related
    structure had misaligned address, idk if that is ok.*/
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/attributes.c:222 col 8
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/attributes.c:285 col 20
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/attributes.c:286 col 17
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/attributes.c:289 col 10
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/attributes.c:292 col 29
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/object.c:866 col 44
E/TA:  Undefined behavior type_mismatch_v1 at ta/pkcs11/src/object.c:891 col 15
/* On stack array with zero length which is passed as pointer to kernel but never referenced */
E/TA:  Undefined behavior vla_bound_not_positive at lib/libutee/tee_api_objects.c:550 col 24
E/TA:  Undefined behavior vla_bound_not_positive at lib/libutee/tee_api_operations.c:2021 col 24
E/TA:  Undefined behavior vla_bound_not_positive at lib/libutee/tee_api_operations.c:2097 col 24
E/TA:  Undefined behavior vla_bound_not_positive at lib/libutee/tee_api_operations.c:2173 col 24
E/TA:  Undefined behavior vla_bound_not_positive at lib/libutee/tee_api_operations.c:2247 col 24
```
You can get full trace if you will run with `ubsan_panic = true`

Please let me know if you need this change or not.


lib/libutils/ext/ubsan.c
```diff
@@ -19,12 +20,30 @@
 # define _ubsan_panic() _utee_panic(TEE_ERROR_GENERIC)
 #endif
 
+#define UBSAN_LOC_REPORTED (UINT32_C(1) << 31)
```
jenswi-linaro, line 23: Please use the `BIT32()` macro.

lib/libutils/ext/ubsan.c
```diff
@@ -19,12 +20,30 @@
...
+
 struct source_location {
 	const char *file_name;
 	uint32_t line;
-	uint32_t column;
+	union {
```
jenswi-linaro, line 28: I prefer skipping the union.

lib/libutils/ext/ubsan.c
```diff
@@ -19,12 +20,30 @@
...
 };
 
+static bool was_already_reported(struct source_location *loc)
+{
+	uint32_t reported;
+
+	reported = loc->reported;
+	if (reported & UBSAN_LOC_REPORTED) {
+		return true;
+	} else {
```
jenswi-linaro, line 41: The `else` isn't needed since you return above.

lib/libutils/ext/ubsan.c
```diff
@@ -19,12 +20,30 @@
...
+		uint32_t reported;
+	};
 };
 
+static bool was_already_reported(struct source_location *loc)
+{
+	uint32_t reported;
+
+	reported = loc->reported;
+	if (reported & UBSAN_LOC_REPORTED) {
```
jenswi-linaro, line 39: Is this an optimization to avoid a potentially expensive barrier in `atomic_cas_u32()`?

lib/libutils/ext/ubsan.c
```diff
@@ -113,11 +132,14 @@ static void ubsan_handle_error(const char *func, struct source_location *loc,
...
 
+	if (was_already_reported(loc))
+	    return;
+
 	if (!memcmp(f, func_prefix, sizeof(func_prefix) - 1))
 		f += sizeof(func_prefix);
 
 	EMSG_RAW("Undefined behavior %s at %s:%" PRIu32 " col %" PRIu32,
-		 f, loc->file_name, loc->line, loc->column);
+		 f, loc->file_name, loc->line, (loc->column & (~UBSAN_LOC_REPORTED)));
```
jenswi-linaro, line 142: You can skip the extra `()`.

lib/libutils/ext/ubsan.c
```diff
@@ -19,12 +20,30 @@
...
+static bool was_already_reported(struct source_location *loc)
+{
+	uint32_t reported;
+
+	reported = loc->reported;
+	if (reported & UBSAN_LOC_REPORTED) {
+		return true;
+	} else {
+		return !atomic_cas_u32(&loc->reported, &reported,
+			reported | UBSAN_LOC_REPORTED);
```
jenswi-linaro, line 43: Please indent properly.

lib/libutils/ext/ubsan.c
```diff
@@ -4,11 +4,20 @@
...
 
 #include <compiler.h>
-#include <kernel/panic.h>
 #include <string.h>
 #include <trace.h>
 #include <types_ext.h>
 
+#if defined(__KERNEL__)
+# include <kernel/panic.h>
+# define _ubsan_panic() panic()
```
jenswi-linaro, line 14: `ubsan_panic()`

lib/libutils/ext/sub.mk
```diff
@@ -12,7 +12,10 @@ srcs-y += fault_mitigation.c
 srcs-y += qsort_helpers.c
 srcs-y += array.c
 srcs-y += base64.c
-
+ifeq ($(CFG_CORE_SANITIZE_UNDEFINED),y)
+srcs-y += ubsan.c
+cflags-remove-ubsan.c-y += -fsanitize=undefined
```
jenswi-linaro, line 17: It makes sense not to sanitize `ubsan.c`, but this change should be mentioned in the commit message.

lib/libutils/ext/sub.mk
```diff
@@ -12,7 +12,10 @@ srcs-y += fault_mitigation.c
 srcs-y += qsort_helpers.c
 srcs-y += array.c
 srcs-y += base64.c
-
+ifeq ($(CFG_CORE_SANITIZE_UNDEFINED),y)
+srcs-y += ubsan.c
```
jenswi-linaro, line 16: `srcs-$(CFG_CORE_SANITIZE_UNDEFINED)`

jenswi-linaro (commented):
For the commit messages, please write full sentences and start each with a capital letter.

jenswi-linaro:
We need to choose between adding a `CFG_TA_SANITIZE_UNDEFINED` or renaming `CFG_CORE_SANITIZE_UNDEFINED` to `CFG_SANITIZE_UNDEFINED`. I prefer the former since we shouldn't have any undefined in core so it makes sense to be able to skip it.

jforissier:
> We need to choose between adding a `CFG_TA_SANITIZE_UNDEFINED` or renaming `CFG_CORE_SANITIZE_UNDEFINED` to `CFG_SANITIZE_UNDEFINED`. I prefer the former since we shouldn't have any undefined in core so it makes sense to be able to skip it.

+1

lib/libutils/ext/ubsan.c
alexhoppus: Not really. This is just a part of the logic to get correct result. We need "old value" here for CAS, where "reported" flag is cleared. So we should either check that this flag is cleared or clear it directly.

lib/libutils/ext/ubsan.c
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: Done. Shouldn't that be caught by checkpatch? It seems I don't see any warnings, but maybe I do something wrong.

lib/libutils/ext/ubsan.c
alexhoppus: Done

lib/libutils/ext/sub.mk
alexhoppus: Done

lib/libutils/ext/sub.mk
alexhoppus: Done

alexhoppus:
Hi @jenswi-linaro all comments are addressed.

ldelf/ldelf.mk
```diff
@@ -29,6 +29,10 @@ cppflags$(sm)	+= -include $(conf-file)
 cppflags$(sm)	+= -DTRACE_LEVEL=$(CFG_TEE_CORE_LOG_LEVEL)
 cppflags$(sm)	+= -D__LDELF__
 
+ifeq ($(CFG_TA_SANITIZE_UNDEFINED),y)
```
jenswi-linaro, line 32: Please use `CFG_CORE_SANITIZE_UNDEFINED` for LDELF.

lib/libutils/ext/sub.mk
```diff
@@ -12,7 +12,11 @@ srcs-y += fault_mitigation.c
 srcs-y += qsort_helpers.c
 srcs-y += array.c
 srcs-y += base64.c
-
+srcs-$(call cfg-one-enabled,CFG_TA_SANITIZE_UNDEFINED \
+			    CFG_CORE_SANITIZE_UNDEFINED) += ubsan.c
+cflags-remove-ubsan.c-$(call cfg-one-enabled,CFG_TA_SANITIZE_UNDEFINED \
```
jenswi-linaro, line 17: Please use `cflags-remove-ubsan.c-y` to keep it simple.

lib/libutils/ext/ubsan.c
```diff
@@ -103,14 +129,17 @@ static void ubsan_handle_error(const char *func, struct source_location *loc,
...
 
+	if (was_already_reported(loc))
+	    return;
+
 	if (!memcmp(f, func_prefix, sizeof(func_prefix) - 1))
 		f += sizeof(func_prefix);
 
 	EMSG_RAW("Undefined behavior %s at %s:%" PRIu32 " col %" PRIu32,
-		 f, loc->file_name, loc->line, loc->column);
+		 f, loc->file_name, loc->line, loc->column & (~UBSAN_LOC_REPORTED));
```
jenswi-linaro, line 139: Please drop the redundant `()`.

ldelf/sub.mk
```diff
@@ -13,3 +13,4 @@ srcs-y += ta_elf.c
 srcs-y += ta_elf_rel.c
 srcs-$(CFG_FTRACE_SUPPORT) += ftrace.c
 srcs-$(CFG_TA_PAUTH) += pauth.c
+cflags-lib-$(CFG_TA_SANITIZE_UNDEFINED) += -fsanitize=undefined
```
jenswi-linaro, line 16: Is this needed? You already assign `cflags$(sm)` in `ldelf.mk`

ta/ta.mk
```diff
@@ -57,6 +57,9 @@ aflags$(sm)	:= $(platform-aflags) $($(sm)-platform-aflags)
 # compiled, these flags are not propagated to the TA
 cppflags$(sm)	+= -include $(conf-file)
 cppflags$(sm) += -DTRACE_LEVEL=$(CFG_TEE_TA_LOG_LEVEL)
+ifeq ($(CFG_TA_SANITIZE_UNDEFINED),y)
+cppflags$(sm) += -fsanitize=undefined
```
jenswi-linaro, line 61: `cflags$(sm)`

lib/libutils/ext/ubsan.c
```diff
@@ -188,20 +217,25 @@ void __ubsan_handle_out_of_bounds(void *data_, void *idx __unused)
 	ubsan_handle_error(__func__, &data->loc, ubsan_panic);
 }
 
-void __ubsan_handle_builtin_unreachable(void *data_)
+/*
+ * __ubsan_handle_builtin_unreachable and __ubsan_handle_missing_return
+ * will give "‘noreturn’ function does return" warning, because
```
jenswi-linaro, line 222: Warnings aren't accepted, you need to define `ubsan_panic()` so that it doesn't cause warnings.

mk/config.mk
```diff
@@ -449,6 +449,7 @@ CFG_CORE_BGET_BESTFIT ?= $(call cfg-one-enabled, CFG_WITH_PAGER CFG_LOCKDEP)
 # Enable support for detected undefined behavior in C
 # Uses a lot of memory, can't be enabled by default
 CFG_CORE_SANITIZE_UNDEFINED ?= n
+CFG_TA_SANITIZE_UNDEFINED ?= n
```
jenswi-linaro, line 452: Please add a 
`_make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SANITIZE_UNDEFINED=y CFG_TA_SANITIZE_UNDEFINED=y`
line in `.github/workflows/ci.yml`.

ldelf/ldelf.mk
alexhoppus: Done

lib/libutils/ext/sub.mk
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: Done

ldelf/sub.mk
alexhoppus: Removed. Done

lib/libutils/ext/ubsan.c
alexhoppus: Done

mk/config.mk
alexhoppus: Done. I added also CFG_TEE_RAM_VA_SIZE=0x00400000 , otherwise it doesn't compile.

ta/ta.mk
```diff
@@ -57,6 +57,9 @@ aflags$(sm)	:= $(platform-aflags) $($(sm)-platform-aflags)
 # compiled, these flags are not propagated to the TA
 cppflags$(sm)	+= -include $(conf-file)
 cppflags$(sm) += -DTRACE_LEVEL=$(CFG_TEE_TA_LOG_LEVEL)
+ifeq ($(CFG_TA_SANITIZE_UNDEFINED),y)
```
jenswi-linaro, line 60: Please add `CFG_TA_SANITIZE_UNDEFINED` to `ta-mk-file-export-vars-$(sm)` above.

ta/ta.mk
```diff
@@ -57,6 +57,9 @@ aflags$(sm)	:= $(platform-aflags) $($(sm)-platform-aflags)
...
+cflags$(sm) += -fsanitize=undefined
```
jenswi-linaro, line 61: I expect that `ta/mk/build-user-ta.mk` and `ta/mk/ta_dev_kit.mk` also need to be updated for the TAs to get this flag.

mk/config.mk
jenswi-linaro: Yes, option 2 makes the most sense to me too.

jenswi-linaro:
Looks good, please squash the fixups and I'll make a final review.

ta/ta.mk
alexhoppus: Done

ta/ta.mk
alexhoppus: > I expect that ta/mk/build-user-ta.mk and ta/mk/ta_dev_kit.mk also need to be updated for the TAs to get this flag.

I added fsanitize flags in `ta/mk/ta_dev_kit.mk` .  I see that this is enough to cover instrumentation for optee examples.
```
$ pwd
/optee/out-br/build/optee_examples_ext-1.0
optee_examples_ext-1.0 $ for i in `find . -name *.cmd`; do echo -n $i; if [ -z "$(cat $i | grep undefined)" ];  then echo ":  no ubsan"; else echo ":  ubsan"; fi; done
./random/ta/out/.random_example_ta.o.cmd:  ubsan
./random/ta/out/.user_ta_header.o.cmd:  ubsan
./hello_world/ta/out/.user_ta_header.o.cmd:  ubsan
./hello_world/ta/out/.hello_world_ta.o.cmd:  ubsan
./acipher/ta/out/.acipher_ta.o.cmd:  ubsan
./acipher/ta/out/.user_ta_header.o.cmd:  ubsan
./plugins/ta/out/.plugin_ta.o.cmd:  ubsan
./plugins/ta/out/.user_ta_header.o.cmd:  ubsan
./aes/ta/out/.user_ta_header.o.cmd:  ubsan
./aes/ta/out/.aes_ta.o.cmd:  ubsan
./hotp/ta/out/.hotp_ta.o.cmd:  ubsan
./hotp/ta/out/.user_ta_header.o.cmd:  ubsan
./secure_storage/ta/out/.secure_storage_ta.o.cmd:  ubsan
./secure_storage/ta/out/.user_ta_header.o.cmd:  ubsan
```
however I'm not sure why they also should be added in `ta/mk/build-user-ta.mk` . Please explain, if you think this is necessary. Thanks.

mk/config.mk
alexhoppus: Done

alexhoppus:
> Looks good, please squash the fixups and I'll make a final review.

Done, all comments are addressed.

jenswi-linaro (commented):
Comments for the commit "core: move ubsan implementation to libutils"

Feel free to rewrite commits and force-push as needed.

lib/libutils/ext/sub.mk
```diff
@@ -12,7 +12,10 @@ srcs-y += fault_mitigation.c
 srcs-y += qsort_helpers.c
 srcs-y += array.c
 srcs-y += base64.c
-
+ifeq ($(CFG_CORE_SANITIZE_UNDEFINED),y)
```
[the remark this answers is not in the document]
jforissier: No big deal but I'd rather avoid building a file needlessly. The conditional is not that complicated:
```make
ifeq $($(sm)-$(CFG_CORE_SANITIZE_UNDEFINED),core-y)
build-ubsan := y
endif
ifeq $($(sm)-$(CFG_CORE_SANITIZE_UNDEFINED),ldelf-y)
build-ubsan := y
endif
src-$(build-ubsan) += ubsan.c
```

jforissier (commented):
For "ldelf: enable ubsan support":

lib/libutils/ext/sub.mk
```diff
@@ -12,10 +12,9 @@ srcs-y += fault_mitigation.c
 srcs-y += qsort_helpers.c
 srcs-y += array.c
 srcs-y += base64.c
-ifeq ($(CFG_CORE_SANITIZE_UNDEFINED),y)
-srcs-y += ubsan.c
+srcs-$(call cfg-one-enabled,CFG_TA_SANITIZE_UNDEFINED \
+			    CFG_CORE_SANITIZE_UNDEFINED) += ubsan.c
```
jforissier, line 16: As I said for a previous commit, I'd rather build this file only when needed.
```
ifneq (,$(filter $(sm)-$(CFG_CORE_SANITIZE_UNDEFINED),core-y ldelf-y))
build-ubsan := y
endif
ifeq(,$(filter $(sm),core ldelf) # TA
build-ubsan := $(CFG_TA_SANITIZE_UNDEFINED)
endif
src-$(build-ubsan) += ubsan.c
```

jforissier (commented):
For commit "ta: enable ubsan support for TAs" with my comment addressed:

lib/libutils/ext/ubsan.c
```diff
@@ -39,6 +43,18 @@ static void __noreturn ubsan_panic(void)
 	while (1);
 }
 
+static bool was_already_reported(struct source_location *loc)
+{
+	uint32_t column;
+
+	column = loc->column;
```
jforissier, line 50: Why consider only the column? The file path and the line number are important as well.

lib/libutils/ext/ubsan.c
```diff
@@ -39,6 +43,18 @@ static void __noreturn ubsan_panic(void)
...
+	if (column & UBSAN_LOC_REPORTED)
```
jforissier, line 51: Unneeded micro-optimization IMO.

jforissier (commented):
Reviewing "core: suppress duplicate ubsan reports".

jforissier (commented):
For "build: always link in-tree TAs with the -static flag":

jforissier (commented):
For "ci: add ubsan build job":

.github/workflows/ci.yml
```diff
@@ -138,6 +138,7 @@ jobs:
           _make PLATFORM=vexpress-qemu_armv8a CFG_{ATTESTATION,DEVICE_ENUM,RTC,SCMI,SECSTOR_TA_MGT,VERAISON_ATTESTATION}_PTA=y CFG_WITH_STATS=y CFG_TA_STATS=y
           _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SEL1_SPMC=y CFG_NS_VIRTUALIZATION=y
           _make PLATFORM=vexpress-qemu_armv8a CFG_CRYPTO_WITH_CE=y CFG_CRYPTOLIB_NAME=mbedtls CFG_CRYPTOLIB_DIR=lib/libmbedtls
+          _make PLATFORM=vexpress-qemu_armv8a CFG_CORE_SANITIZE_UNDEFINED=y CFG_TA_SANITIZE_UNDEFINED=y CFG_TEE_RAM_VA_SIZE=0x00400000
```
jforissier, line 141: I could be nice to set CFG_TEE_RAM_VA_SIZE to a sufficient value by default in plat-vexpress when ubsan is enabled. But it's also OK like that.

lib/libutils/ext/ubsan.c
alexhoppus: > Why consider only the column? The file path and the line number are important as well.

The unique source location have 1:1 mapping with `struct source_location *loc `, the last bit of column field only used to store flag, that this source location was already reported. Previously I used union, but it was [requested](https://github.com/OP-TEE/optee_os/pull/7312#discussion_r1995079647) to remove it.

lib/libutils/ext/ubsan.c
alexhoppus: This is [not](https://github.com/OP-TEE/optee_os/pull/7312#discussion_r1996870017) a micro-optimization. Could you please provide a code snippet demonstrating the changes you'd like to apply to this code?

jforissier:
```C
static bool was_already_reported(struct source_location *loc)
{
	uint32_t column = loc->column;

	return !atomic_cas_u32(&loc->column, &column,
			       column | UBSAN_LOC_REPORTED);
}
```

lib/libutils/ext/ubsan.c
jforissier: Got it. For simplicity I would personally have added a `bool reported;` to `struct source_location` and dropped the atomic swap (who cares if the same location might be reported twice once in a blue moon?) but the code you propose is clearly correct and better, if a bit more complex. So that's fine with me.

alexhoppus: > Got it. For simplicity I would personally have added a bool reported; to struct source_location and dropped the atomic swap (who cares if the same location might be reported twice once in a blue moon?) but the code you propose is clearly correct and better, if a bit more complex. So that's fine with me.

That is not possible, this structures are created/passed to callbacks by gcc, we can't change their layout. Linux [do](https://github.com/torvalds/linux/blob/master/lib/ubsan.c#L112) the same

lib/libutils/ext/ubsan.c
alexhoppus: Hi, thanks.

I don't think that would work. Let's consider an example. Consider before call to our `was_already_reported` `UBSAN_LOC_REPORTED` was set already some time ago. So we should return `true` what means "the result was already reported".


```
/* static inline bool atomic_cas_u32(uint32_t *p, uint32_t *oval, uint32_t nval) */
uint32_t column = loc->column;
return !atomic_cas_u32(&loc->column, &column,
			       column | UBSAN_LOC_REPORTED);
```

The `oval` (old value) passed is `column | UBSAN_LOC_REPORTED` is equal to `&loc->column` content. So `atomic_cas_u32` we will return `true` and `was_already_reported` will return `false`, what is not correct. So in theory we can do:

```
/* static inline bool atomic_cas_u32(uint32_t *p, uint32_t *oval, uint32_t nval) */
uint32_t column = loc->column;

column = (column & ~UBSAN_LOC_REPORTED);
return !atomic_cas_u32(&loc->column, &column,
			       column | UBSAN_LOC_REPORTED);
```

but I prefer my initial version, why we need to clear the flag if we can check it.

lib/libutils/ext/ubsan.c
```diff
@@ -4,17 +4,57 @@
...
+	 * _ldelf_panic and _utee_panic are not marked as noreturn,
+	 * however they should be. To prevent "‘noreturn’ function
+	 * does return" warning the while loop is used.
+	 */
+	while (1);
+}
+
+static bool was_already_reported(struct source_location *loc)
+{
+	uint32_t column;
```
etienne-lms, line 48: Coding style: set an initialization value, either `= 0` or `= loc->column` straight

etienne-lms (commented):
Some nitpicking comments.
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` once addressed.

lib/libutils/ext/sub.mk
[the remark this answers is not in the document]
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: Done

lib/libutils/ext/ubsan.c
alexhoppus: > You're right, sorry for the wrong advice.

No problem.

Done

lib/libutils/ext/ubsan.c
```diff
@@ -8,40 +8,37 @@
 #include <trace.h>
 #include <types_ext.h>
 #include <atomic.h>
```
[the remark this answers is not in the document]
alexhoppus: Done

lib/libutils/ext/ubsan.c
```diff
@@ -29,6 +26,23 @@ struct source_location {
...
+	_ldelf_panic(2);
+#else
+	_utee_panic(TEE_ERROR_GENERIC);
+#endif
+	/*
+	 * _ldelf_panic and _utee_panic are not marked as noreturn,
+	 * however they should be. To prevent "‘noreturn’ function
+	 * does return" warning the while loop is used.
+	 */
+	while (1);
```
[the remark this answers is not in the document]
alexhoppus: Done

lib/libutils/ext/ubsan.c
```diff
@@ -95,127 +135,130 @@ void __ubsan_handle_nonnull_arg(void *data_
...
 
 static void ubsan_handle_error(const char *func, struct source_location *loc,
-			       bool should_panic)
+			       bool panic_flag)
 {
 	const char *f = func;
 	const char func_prefix[] = "__ubsan_handle";
 
+	if (was_already_reported(loc))
+	    return;
```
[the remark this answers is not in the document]
alexhoppus: Done

jforissier (commented):

jenswi-linaro:

alexhoppus:
Hi @jenswi-linaro @jforissier all comments were addressed, reviewed-by tags were placed, also rebased to master branch. Let me know if anything else should be done here.

PR #7315, opened 2025-03-14
core: arm: boot: fix calling page_alloc_init()

jenswi-linaro:
The functions page_alloc_init() and nex_page_alloc_init() depends on MEM_AREA_TEE_DYN_VASPACE and MEM_AREA_NEX_DYN_VASPACE, but the memory areas are only available with CFG_DYN_CONFIG so check that before calling the functions.

Fixes: 0e12fb0c2d75 ("core: arm: boot: call page_alloc_init()")

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -956,7 +956,9 @@ static void init_primary(unsigned long pageable_part)
 		 */
 		assert(va && va <= boot_cached_mem_end);
 		boot_cached_mem_end = va;
+	}
 
+	if (IS_ENABLED(CFG_DYN_CONFIG)) {
 		/*
 		 * This is needed to enable virt_page_alloc() now that
 		 * boot_mem_alloc() can't be used any longer.


jforissier (commented):

jenswi-linaro:
Tag applied

core/arch/arm/kernel/boot.c
```diff
@@ -956,7 +956,9 @@ static void init_primary(unsigned long pageable_part)
 		 */
 		assert(va && va <= boot_cached_mem_end);
 		boot_cached_mem_end = va;
+	}
 
+	if (IS_ENABLED(CFG_DYN_CONFIG)) {
```
etienne-lms, line 961: Will `virt_page_alloc()` be used when pager is enabled?

jenswi-linaro: The corresponding function with pager is `tee_pager_alloc()`. The pager manages all available physical memory that doesn't have to be mapped permanently. So it's a bit different. But as I transform the code to be more dynamic it might make sense to hide that behind an abstract interface.

jenswi-linaro:
Ready to merge?

PR #7328, opened 2025-03-27
core: pta: fix virt_to_phys() self tests

jenswi-linaro:
Only test on VCORE_FREE_* addresses if CFG_DYN_CONFIG=y since those addresses otherwise are unmapped or even unassigned.

Fixes: a05b3a459a3b ("core: pta: add self tests for virt_to_phys()")


jforissier (commented):

PR #7334, opened 2025-03-28
scripts: use /usr/bin/env bash

Emantor:
Some distributions don't install bash inside of bash since the only program thats required to be available in /bin according to POSIX is sh. Consequently running scripts with #!/bin/bash fails on such systems. Use /usr/bin/env (which is required to be available according to POSIX) to run the correct bash binary.

### DIFF
--- a/scripts/checkpatch.sh
+++ b/scripts/checkpatch.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 DIR="${BASH_SOURCE%/*}"
 
--- a/scripts/checkpatch_inc.sh
+++ b/scripts/checkpatch_inc.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 CHECKPATCH="${CHECKPATCH:-checkpatch.pl}"
 CHECKPATCH_OPT="${CHECKPATCH_OPT:-}"
--- a/ta/pkcs11/scripts/dump_ec_curve_params.sh
+++ b/ta/pkcs11/scripts/dump_ec_curve_params.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # SPDX-License-Identifier: BSD-2-Clause
 
 EC_CURVES="prime192v1 secp224r1 prime256v1 secp384r1 secp521r1"
--- a/ta/pkcs11/scripts/verify-helpers.sh
+++ b/ta/pkcs11/scripts/verify-helpers.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # SPDX-License-Identifier: BSD-2-Clause
 
 SHOW_DETAILS=1


jforissier:
> Some distributions don't install bash inside of bash

You probably mean in /bin not bash

> since the only program thats required to be available in /bin according to POSIX is sh. 

I don't think this is true, sh might be somewhere else, but that's irrelevant since we want bash

> Consequently running scripts with #!/bin/bash fails on such systems. Use /usr/bin/env (which is required to be available according to POSIX) 

Reference? I don't think POSIX requires that either. I still agree with the proposed change since I think it is more portable.

> to run the correct bash binary.

jforissier:
With an updated commit description:

jenswi-linaro:

Emantor:
> > Some distributions don't install bash inside of bash
> 
> You probably mean in /bin not bash

Yes, doh on my part.

> > since the only program thats required to be available in /bin according to POSIX is sh.
> 
> I don't think this is true, sh might be somewhere else, but that's irrelevant since we want bash

Yes, the spec even mentions that `/bin/sh/`should *not* be assumed to be the correct PATH for `sh. TIL, thanks for the pointers.
> 
> > Consequently running scripts with #!/bin/bash fails on such systems. Use /usr/bin/env (which is required to be available according to POSIX)
> 
> Reference? I don't think POSIX requires that either. I still agree with the proposed change since I think it is more portable.

You are also right, POSIX requires env to be available, but the location is not specified either. My guess is that NixOS decided to support `/usr/bin/env` to not break even more scripts.

> > to run the correct bash binary.

I'll change the commit message, thanks again :+1:

Emantor:
> > > Some distributions don't install bash inside of bash
> > 
> > 
> > You probably mean in /bin not bash
> 
> Yes, doh on my part.
> 
> > > since the only program thats required to be available in /bin according to POSIX is sh.
> > 
> > 
> > I don't think this is true, sh might be somewhere else, but that's irrelevant since we want bash
> 
> Yes, the spec even mentions that `/bin/sh/`should _not_ be assumed to be the correct PATH for `sh`. TIL, thanks for the pointers.
> 
> > > Consequently running scripts with #!/bin/bash fails on such systems. Use /usr/bin/env (which is required to be available according to POSIX)
> > 
> > 
> > Reference? I don't think POSIX requires that either. I still agree with the proposed change since I think it is more portable.
> 
> You are also right, POSIX requires env to be available, but the location is not specified either. My guess is that NixOS decided to support `/usr/bin/env` to not break even more scripts.
> 
> > > to run the correct bash binary.
> 
> I'll change the commit message, thanks again 👍

Emantor:
Pushed and applied tags, please check the commit message again.

jforissier:
> Pushed and applied tags, please check the commit message again.

Looks good, thank you!

Emantor:
Force pushed to apply tag from Etienne.

etienne-lms (commented):
Nitpicking: since the patch changes file in **scripts/** and from **ta/...**, the commit message header line could be change to something like:
`tree wide: use /usr/bin/env bash in bash scripts`

Emantor:
Sounds good, applied.

PR #7335, opened 2025-03-28
drivers: imx_csu: add SA settings for i.MX6ULL

Emantor:
Add the CSU SA settings for i.MX6ULL. The i.MX6ULL implements the same non-Trustzone aware masters as the i.MX6UL, so the same settings can be used. This setting ensures that no non-TrustZone aware master is able to read secure memory. Information on the CSU SA register values were taken from i.MX6ULL Security Reference Manual rev 0.


Emantor:
Tag applied, thanks :+1:

Issue #7331, opened 2025-03-27
fs-htree aad length issue


anilreddy3:
Hi,
I'm implementing hardware offload AES-GCM authenc driver. I see my driver receiving crypto operation calls from fs_htree.c->authenc_init(), noticed an issue, where crypto_authenc_init calling aad_len **68bytes** but crypto_authenc_update_aad called multiple times to update aad data of total **52bytes** only. due which hardware crypto engine throws an error.

anilreddy3:
Looks like first crypto_authenc_update_aad() calls add_len TEE_FS_HTREE_FEK_SIZE instead of TEE_FS_HTREE_HASH_SIZE.

jenswi-linaro:
Yes, you're right. Could you make a pull request to fix it?

anilreddy3:
Yes,  Sure.

minminlittleshrimp:
I see the patch merged, is there any further discussions on this bug?
If not can the maintainers kindly close this one for not confusing in future reading.
Thanks a lot.

Issue #6532, opened 2023-12-07
crypto_authenc_update_aad on decrypt path in hash tree 


Amit-Radur:
I have few queries regarding crypto function used under hash tree.

1) Do we really need to call crypto_authenc_update_aad when TEE_OperationMode  is decrypt, shouldn't the Additional 
    authenticated data is updated only when TEE_OperationMode is encrypt ?

   https://github.com/OP-TEE/optee_os/blob/master/core/tee/fs_htree.c#L476C9-L476C34

2) Is this a BUG, using TEE_FS_HTREE_FEK_SIZE instead of TEE_FS_HTREE_HASH_SIZE ?

``
                 
 
            if (!ni) {
                   res = crypto_authenc_update_aad(ctx, mode, ht->root.node.hash,
     -                                               TEE_FS_HTREE_FEK_SIZE);
     +                                               TEE_FS_HTREE_HASH_SIZE);

3) How do one fully reset secure storage, do we need to create the /data/tee directory every time we wanted to reset 
    the secure storage ?

jenswi-linaro:
1. Yes, it's needed to calculate the tag correctly.
2. Yes, fixing it will unfortunately break already existing secure storage objects. This means that we're currently using truncated SHA-256 hashes. I would very much like to fix this, but we should do it in a compatible way.
3. That depends on what you use for rollback protection. If you have none, it's enough to remove the content of `/data/tee`.

jenswi-linaro:
https://optee.readthedocs.io/en/latest/debug/abort_dumps.html

jenswi-linaro:
I'm sorry, I can't help debugging your code.

Amit-Radur:
Sure, No issues. I am closing it as your answers helped me with few things.

PR #7340, opened 2025-03-31
core: fs_htee: Fix wrong aad length in authenc_init()

anilreddy3:
In authenc_init(), aad length field passed to
crypto_authenc_init() does not match with the total aad data passed via crypto_authenc_update_aad() for lower layer crypto computation. This fixes the issue.

Link: https://github.com/OP-TEE/optee_os/issues/7331

### DIFF
--- a/core/tee/fs_htree.c
+++ b/core/tee/fs_htree.c
@@ -4,6 +4,7 @@
  */
 
 #include <assert.h>
+#include <config.h>
 #include <crypto/crypto.h>
 #include <initcall.h>
 #include <kernel/tee_common_otp.h>
@@ -473,8 +474,13 @@ static TEE_Result authenc_init(void **ctx_ret, TEE_OperationMode mode,
 		goto err_free;
 
 	if (!ni) {
+		size_t hash_size = TEE_FS_HTREE_HASH_SIZE;
+
+		if (IS_ENABLED(CFG_REE_FS_HTREE_HASH_SIZE_COMPAT))
+			hash_size = TEE_FS_HTREE_FEK_SIZE;
+
 		res = crypto_authenc_update_aad(ctx, mode, ht->root.node.hash,
-						TEE_FS_HTREE_FEK_SIZE);
+						hash_size);
 		if (res != TEE_SUCCESS)
 			goto err;
 
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -149,6 +149,12 @@ CFG_TEE_FW_MANUFACTURER ?= FW_MAN_UNDEF
 # TEE_STORAGE_PRIVATE is passed to the trusted storage API)
 CFG_REE_FS ?= y
 
+# CFG_REE_FS_HTREE_HASH_SIZE_COMPAT, when enabled, supports the legacy
+# REE FS hash tree tagging implementation that uses a truncated hash.
+# Be warned that disabling this config could break accesses to existing
+# REE FS content.
+CFG_REE_FS_HTREE_HASH_SIZE_COMPAT ?= y
+
 # RPMB file system support
 CFG_RPMB_FS ?= n
 


jenswi-linaro:
This looked familiar, and now I remember https://github.com/OP-TEE/optee_os/issues/6532
Nothing has changed since then. I'm not sure we can fix it right now, but we should keep track of the issue and fix it for a future 5.0 OP-TEE release.

anilreddy3:
Okay, what would be the probable time frame for 5.0 release.

etienne-lms:
I think it would be worth introducing the correction now upon a config switch (default disabled) for those that do not want to introduce that legacy in their platform and will also face compatibility issues with such a 5.x release stream.

jenswi-linaro:
Yes, that's a good idea.

etienne-lms:
May I suggest something like:
```patch
	if (!ni) {
+		size_t hash_size = TEE_FS_HTREE_HASH_SIZE;
+
+		if (IS_ENABLED(CFG_REE_FS_HTREE_HASH_SIZE_COMPAT))
+			 hash_size = TEE_FS_HTREE_FEK_SIZE;
+
		res = crypto_authenc_update_aad(ctx, mode, ht->root.node.hash,
-						TEE_FS_HTREE_FEK_SIZE);
+						hash_size);
		if (res != TEE_SUCCESS)
			goto err;
```
and default initialize `CFG_REE_FS_HTREE_HASH_SIZE_COMPAT ?= y` in **mk/config.mk**?

anilreddy3:
That's a nice idea to get this fix for now.

jforissier:
@etienne-lms done.

core/tee/fs_htree.c
```diff
@@ -473,8 +474,13 @@ static TEE_Result authenc_init(void **ctx_ret, TEE_OperationMode mode,
 		goto err_free;
 
 	if (!ni) {
+		size_t hash_size = TEE_FS_HTREE_HASH_SIZE;
+
+		if (IS_ENABLED(CFG_REE_FS_HTREE_HASH_SIZE_COMPAT))
+			hash_size = TEE_FS_HTREE_FEK_SIZE;
```
jenswi-linaro, line 480: Please fix `aad_len` too.

etienne-lms: `aad_len` already considers `TEE_FS_HTREE_HASH_SIZE` at line 457, even if we used to pass a truncated hash size. I think we should not change (fix) the AAD size in the legacy config (`CFG_REE_FS_HTREE_HASH_SIZE_COMPAT=y`).

jenswi-linaro: AES-GCM doesn't code the supplied `aad_len` into the tag or anything. It only accumulates the AAD, but as @anilreddy3 discovered, it might be checked in the AES-GCM implementation. That's why I think that `aad_len` should match the length of the supplied AAD.
If someone (for some reason) replaced AES-GCM with AES-CCM, then `aad_len` must match the length of the supplied AAD.

anilreddy3: Yes, correct.

mk/config.mk
```diff
@@ -149,6 +149,12 @@ CFG_TEE_FW_MANUFACTURER ?= FW_MAN_UNDEF
 # TEE_STORAGE_PRIVATE is passed to the trusted storage API)
 CFG_REE_FS ?= y
 
+# Current REE file system Hash Tree using truncated hash size, fixing it
```
etienne-lms, line 152: Could you replace "Current" with "Legacy".
s/using/uses/
s/break/breaks/
s/introduced/introduce/

Maybe a bit of rephrasing:
```
CFG_REE_FS_HTREE_HASH_SIZE_COMPAT, when enabled, supports the legacy
REE FS hash tree tagging imeplementation that uses a truncated hash.
Be warned that disabling this config could break accesses to existing
REE FS content.

core/tee/fs_htree.c
etienne-lms: Ok, thanks @jenswi-linaro for the explanation. @anilreddy3, could you add a dedicated commit for this change, not mix within the commit dedicated to `CFG_REE_FS_HTREE_HASH_SIZE_COMPAT`.

anilreddy3: I think, if someone replaces AES-GCM with AES-CCM it anyway break the existing file system, In that case they should move to fix version by disabling CFG_REE_FS_HTREE_HASH_SIZE_COMPAT config, with that aad_len match with the supplied AAD data.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` fix a few fixes in the commit message:
* s/aad/AAD/g (several occurrences)
* s/that not to break/without breaking/
* s/introduced/introduce/
* Could remove sentence "*Override the default value by (...) in platform specific
configuration*".

Maybe @jenswi-linaro comment https://github.com/OP-TEE/optee_os/pull/7340/commits/717d48b693c14419169f6ca4dfa8561f309503a2#r2024696778 can be addressed in a specific P-R, based on `CFG_REE_FS_HTREE_HASH_SIZE_COMPAT` once this P-R is merged.

jenswi-linaro:
With Etienne's comments addressed, please apply:

anilreddy3:
Thanks @etienne-lms, updated PR with comments addressed and tags applied.

PR #7341, opened 2025-03-31
libutils: fix realloc() for user-space

jenswi-linaro:
User-space expects realloc() to zero any newly allocated memory, but with the introduction of the mem_alloc() helper it was accidentally removed when realloc() is passed a NULL pointer. Fix this by adding MAF_ZERO_INIT for realloc() from user-space.

Fixes: 07c190fbde10 ("libutils: refactor the default malloc functions")


jforissier:
> User-space expects realloc() to zero any newly allocated memory

Really? The man page for realloc() on my Linux system (Ubuntu 24.10) says "_If the new size is larger than the old size, the added memory will __not__ be initialized._"

jenswi-linaro:
It simplifies TEE_Realloc(). But it looks like this code depends on it:
https://github.com/OP-TEE/optee_os/blob/228cf40e2ba89ab3d8afac3eba57e4563fd6c630/lib/libutee/tcb.c#L113-L122

Our user-space isn't the same as linux user-space.

jforissier:
> Our user-space isn't the same as linux user-space.

Sure, but still, `realloc()` is a standard function and the spec mentions that the content of the memory is undefined (https://pubs.opengroup.org/onlinepubs/9699919799/functions/realloc.html). I don't think we should rely on non-standard behavior. Better fix the callers. Or introduce a wrapper.

jenswi-linaro:
Update

jforissier:
There is a double space in the commit message for "libutils: avoid unneeded zero init in realloc()". Other than that, looks good! Thanks.

etienne-lms (commented):
For commit "libutee: TEE_Realloc(): avoid realloc()", in commit message:
"TEE_Realloc() and realloc() **~~has~~ have** ..."

For commit "libutee: fix __utee_tcb_init()", I wonder if the `Fixes:` tag applies since there are quite many dependencies to consider for this commit to be applied on the referred one.

That said, looks all good to me.

etienne-lms:
> We've had fixes tags referring to commit https://github.com/OP-TEE/optee_os/commit/b01047730e77127c23a36591643eeb8bb0487d68 ("Open-source the TEE Core").

That's right, I remember :laughing: Good point.

PR #7342, opened 2025-03-31
libutils: merge malloc pools if possible

jenswi-linaro:
When adding pools with for instance malloc_add_pool(), check if the pool can be merged with other pools. This decreases fragmentation by removing unneeded allocated end of pool sentinels between merged pools.

The end of pool sentinel is changed from a struct bhead to a struct bfhead, increasing it by two pointers, to allow it to be freed using brel() once it has been turned into a regular allocated block.

### DIFF
--- a/lib/libutils/isoc/bget.c
+++ b/lib/libutils/isoc/bget.c
@@ -1125,7 +1125,7 @@ void bpool(buf, len, poolset)
        it  had	better	not  be  (much) larger than the largest buffer
        whose size we can store in bhead.bsize. */
 
-    assert(len - sizeof(struct bhead) <= -((bufsize) ESent + 1));
+    assert(len - sizeof(struct bfhead) <= -((bufsize) ESent + 1));
 
     /* Clear  the  backpointer at  the start of the block to indicate that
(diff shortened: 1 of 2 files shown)


jforissier:
The caller of `malloc_add_pool()` should be able to determine if the pools can be merged, no? (and call it only once rather than multiple times). In fact I see nowhere in the code where pools would be merged, I mean in `core/arch/arm/kernel/boot.c` there is:
```
malloc_add_pool(__heap2_start, __heap2_end - __heap2_start);
malloc_add_pool(__heap1_start, __heap1_end - __heap1_start);
```
...but heap1 and heap2 are not contiguous. As for user space TAs, they add only one pool. So where would this be used?

jenswi-linaro:
It will be useful when we start to allocate memory malloc pools dynamically, either on demand or based on external configuration. I can save this patch for the first PR that benefits from it if you prefer that.

jforissier:
> It will be useful when we start to allocate memory malloc pools dynamically, either on demand or based on external configuration. 

Got it.

> I can save this patch for the first PR that benefits from it if you prefer that.

Yes, I think it would make more sense.

jenswi-linaro:
OK

PR #7343, opened 2025-04-01
drivers: stm32_rifsc: remove semaphore acquisition when applying the RIF conf

GseoC:
Remove the semaphore acquisition when applying the configuration so that they are taken, if necessary, during the firewall bus probe. This avoid semaphores being taken by OP-TEE when the peripheral isn't used.


etienne-lms (commented):

in commit message: 
\- "This avoid semaphores being taken ..."
\+"This avoid**s** semaphores **from** being taken ..."

GseoC:
Comment and tag applied, thanks

PR #7344, opened 2025-04-01
core: imx: disable ELE on i.MX8ULP by default

sahilnxp:
Sorry for back and forth about the ELE support enablement.
On i.MX8ULP, there is only one MU to communicate with ELE,
which cannot be dedicated on OP-TEE side all the time.
There may be ELE services running on Linux side, which can
cause conflict with OP-TEE, So disabling ELE by default.
Moreover i.MX8ULP also has CAAM, so HUK and Random
number are coming from CAAM.


jenswi-linaro:

sahilnxp:
Thanks @jenswi-linaro 
Tags updated.

PR #7345, opened 2025-04-02
core: arm: ffa: enable FF-A version 1.2 for virtualization

jenswi-linaro:
With Xen version 4.20 released we can announce version 1.2 for OP-TEE when negotiating the version to use. So remove the special check for CFG_NS_VIRTUALIZATION=y when exchanging versions.


jforissier (commented):

jenswi-linaro:
Tag applied

PR #7346, opened 2025-04-02
Core + stm32: Add external device tree repository support

GseoC:
This P-R adds the support for specifying an external repository for the device tree sources.


jenswi-linaro:
Why is this needed?

GseoC:
Saying this is strictly needed would be quite the overstatement but this feature would be nice to:

1. Allow grouping device tree source files for the different components of a software ecosystem and its different flavors. Maybe even reference a system device tree someday
2. Avoid modifying OP-TEE source code to make device tree changes (even though it's compiled with along with it)

jenswi-linaro:
It might be useful downstream in some cases but not at all upstream. 

How about adding support to provide an external already built DTB to be embedded instead? What do others think?

PR #7349, opened 2025-04-03
mk/compile.mk: fix header dependency for device tree and assembly files

tbourgoi:
When compiling a device tree I found that the error of include depencies can occurs too.
I also check the rest of the file and saw that the assembly files are subject to this compilation error too.

So here is a patch to add -MP parameter when compiling both device trees and assembly files similarly to what's done in :
acdc32afe18 ("mk/compile.mk: fix header dependency in .d file")


etienne-lms:
Agree. Thanks @jforissier for checking this. @tbourgoi, I think you can make is single "tree wide: fix header dependency in generated .d files" commit to address all these.

etienne-lms:
By the way, thanks @tbourgoi for the finding.

tbourgoi:
> This raises the question: are there any other file using `-MD` which should also use `-MP`? Shouldn't `-MP` be used always when `-MD` is used, regardless of what is being compiled?
> 
> ```
> $ git grep -n -- " -MD" | grep -v -- -MP
> core/arch/arm/kernel/link.mk:140:       $(q)$(CPPcore) -P -MT $@ -MD -MF $(link-script-dep) \
> core/arch/riscv/kernel/link.mk:51:      $(q)$(CPPcore) -P -MT $@ -MD -MF $(link-script-dep) \
> ldelf/link.mk:47:       $(q)$(CPP$(sm)) -P -MT $$@ -MD -MF $(link-script-dep$(sm)) \
> mk/compile.mk:209:comp-flags-$3 += -MD -MF $$(comp-dep-$3) -MT $$@
> mk/compile.mk:274:                 -MD -MF $$(dtb-predep-$2) -MT $$(dtb-predts-$2)
> ta/link.mk:91:  $(q)$(CPP$(sm)) -P -MT $$@ -MD -MF $(link-script-dep$(sm)) \
> ```
> 
> I think all these occurrences should be fixed.

I reproduced the error with  `core/arch/arm/kernel/link.mk` : 
1.  I include a dummy.h header in core/arch/arm/kernel/kern.ld.S
2. Compile and link from scratch
3. Delete dummy.h and remove include from kern.ld.S
4. Compile (without cleaning the output directory)

etienne-lms (commented):
LGTM. Comments on commit messages.

For commit "mk/compile.mk: fix header dependency for device tree and assembly files":
s/device trees/device tree source files/
s/header file/header files/
Use 12 digits SHA1 refs for commit.

Overall rephrasing proposal:
```
mk/compile.mk: fix header files dependencies for DTS and assembly files

Add pre-processor option -MP to fix build error generated when
header files included from assembly files or devices tree source
files are removed. This change is equivalent to the one provided by
commit acdacdc32afe ("mk/compile.mk: fix header dependency in .d file")
for C source files.
```

For commit "link.mk: fix header dependency for device tree and assembly files"
Same comments + fix the header line, proposal:
`tree wide: fix header files dependencies in linker files`

jforissier:
With the changes requested by Etienne:

@tbourgoi thanks for taking the time to clean this up. Such things really help the user experience.

tbourgoi:
I update the commit messages to use 12 digits SHA1 and remove typos from commit modifying link.mk files.
tags from @jforissier applied on both commits

jenswi-linaro:

tbourgoi:
tag of @jenswi-linaro applied on both commits

PR #7350, opened 2025-04-03
core: ffa: reserve physical memory for manifest

jenswi-linaro:
With CFG_CORE_SEL2_SPMC=y (Hafnium as SPMC at S-EL2), the FF-A manifest passed to OP-TEE resides in the memory reserved for OP-TEE just before the load address. The physical memory pool is initialized with the entire range of secure memory, with holes carved out for already used memory.

Temporarily allocate the physical memory used by the manifest until it's not needed any longer and released by release_manifest_dt().

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -1346,7 +1346,8 @@ static void *get_fdt_from_boot_info(struct ffa_boot_info_header_1_1 *hdr)
 	return fdt;
 }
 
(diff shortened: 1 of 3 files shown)


jenswi-linaro:
Yes, and also to avoid allocating the DT memory for other things before it's released. It's currently only a minor problem, or possibly none at all.

jenswi-linaro:
Tag applied

PR #7352, opened 2025-04-04
Add pta self test for firmware handoff

raymo200915:
Fix a pointer casting bug in transfer_list_add().

Add self tests for transfer list.
Adapt CFG_TRANSFER_LIST with its dependencies and add CFG_TRANSFER_LIST_TEST.

### DIFF
--- a/core/kernel/transfer_list.c
+++ b/core/kernel/transfer_list.c
@@ -497,7 +497,7 @@ struct transfer_list_entry *transfer_list_add(struct transfer_list_header *tl,
 	tl->size += ev - tl_ev;
 
 	if (data)
-		memmove(tl_e + tl_e->hdr_size, data, data_size);
+		memmove((void *)tl_e + tl_e->hdr_size, data, data_size);
 
 	transfer_list_update_checksum(tl);
 
--- a/core/pta/tests/invoke.c
+++ b/core/pta/tests/invoke.c
@@ -433,6 +433,10 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 		return core_aes_perf_tests(nParamTypes, pParams);
 	case PTA_INVOKE_TESTS_CMD_DT_DRIVER_TESTS:
 		return core_dt_driver_tests(nParamTypes, pParams);
+#if defined(CFG_TRANSFER_LIST_TEST)
+	case PTA_INVOKE_TESTS_CMD_TRANSFER_LIST_TESTS:
+		return core_transfer_list_tests(nParamTypes, pParams);
+#endif
 	default:
 		break;
 	}
--- a/core/pta/tests/sub.mk
+++ b/core/pta/tests/sub.mk
@@ -6,3 +6,4 @@ cflags-misc.c-y += -fno-builtin
 srcs-y += mutex.c
 srcs-y += aes_perf.c
 srcs-$(CFG_DT_DRIVER_EMBEDDED_TEST) += dt_driver_test.c
+srcs-$(CFG_TRANSFER_LIST_TEST) += tl.c
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -513,11 +513,10 @@ endif
 # has a stable release.
 # This feature requires the support of Device Tree.
 CFG_TRANSFER_LIST ?= n
-ifeq ($(CFG_TRANSFER_LIST),y)
-$(call force,CFG_DT,y)
-$(call force,CFG_EXTERNAL_DT,y)
-$(call force,CFG_MAP_EXT_DT_SECURE,y)
-endif
+$(eval $(call cfg-enable-all-depends,CFG_TRANSFER_LIST, \
+	 CFG_DT CFG_EXTERNAL_DT CFG_MAP_EXT_DT_SECURE))
+
+CFG_TRANSFER_LIST_TEST ?= $(call cfg-all-enabled,CFG_TRANSFER_LIST)
 
 # Maximum size of the Device Tree Blob, has to be large enough to allow
 # editing of the supplied DTB.
--- a/core/pta/tests/misc.h
+++ b/core/pta/tests/misc.h
@@ -37,4 +37,7 @@ TEE_Result core_aes_perf_tests(uint32_t param_types,
 TEE_Result core_dt_driver_tests(uint32_t param_types,
 				TEE_Param params[TEE_NUM_PARAMS]);
 
+TEE_Result core_transfer_list_tests(uint32_t nParamTypes,
+				    TEE_Param pParams[TEE_NUM_PARAMS]);
+
 #endif /*CORE_PTA_TESTS_MISC_H*/
--- a/core/pta/tests/tl.c
+++ b/core/pta/tests/tl.c
@@ -1,0 +1,247 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2025, Linaro Limited
+ */
+
+#include <kernel/transfer_list.h>
+#include <mm/core_memprot.h>
+#include <pta_invoke_tests.h>
+#include <stdlib.h>
+#include <string.h>
+#include <trace.h>
+#include <util.h>
+
+#include "misc.h"
+
+#define TEST_TL_MAX_SIZE	256
+#define TEST_TE2_ALIGN_POWER	4
+#define TEST_TE1_ID		0xf000
+#define TEST_TE2_ID		0xf0f0
+#define TEST_TE3_ID		0xff00
+
+static const char test_str1[] = "first added entry";
+static const char test_str2[] = "second added entry";
+static const char test_str3[] = "last added entry";
+
+static TEE_Result test_add_te(struct transfer_list_header *tl,
+			      uint16_t tag_id, uint32_t data_size,
+			      const void *data, uint8_t align,
+			      struct transfer_list_entry **tle)
+{
+	uint8_t *te_dat;
+	bool new_max_align = false;
+	vaddr_t old_tl_ev = (vaddr_t)tl + tl->size;
+	struct transfer_list_entry *tl_e;
+
+	if (align > tl->alignment)
+		new_max_align = true;
+
+	if (!align)
+		tl_e = transfer_list_add(tl, tag_id, data_size, data);
+	else
+		tl_e = transfer_list_add_with_align(tl, tag_id, data_size, data,
+						    align);
+
+	if (!tl_e)
+		return TEE_ERROR_GENERIC;
+
+	/* tl->alignment keeps the max entry data alignment of the TL */
+	if (new_max_align && tl->alignment != align)
+		return TEE_ERROR_CORRUPT_OBJECT;
+
+	if (tl_e->tag_id != tag_id || tl_e->hdr_size != sizeof(*tl_e) ||
+	    tl_e->data_size != data_size)
+		return TEE_ERROR_CORRUPT_OBJECT;
+
+	te_dat = transfer_list_entry_data(tl_e);
+
+	if (!te_dat || te_dat != (uint8_t *)tl_e + sizeof(*tl_e))
+		return TEE_ERROR_CORRUPT_OBJECT;
+
+	/*
+	 * If an align arg is passed in:
+	 * 1. Entry data must start from an aligned address, and;
+	 * 2. To align the entry data, a minimum gap should be inserted before
+	 *    the new entry.
+	 */
+	if (align) {
+		vaddr_t mask = TL_ALIGNMENT_FROM_ORDER(align) - 1;
+		size_t gap_min = (~(old_tl_ev + sizeof(*tl_e)) + 1) & mask;
+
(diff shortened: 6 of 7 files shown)


core/kernel/transfer_list.c
```diff
@@ -497,7 +497,7 @@ struct transfer_list_entry *transfer_list_add(struct transfer_list_header *tl,
 	tl->size += ev - tl_ev;
 
 	if (data)
-		memmove(tl_e + tl_e->hdr_size, data, data_size);
+		memmove((void *)tl_e + tl_e->hdr_size, data, data_size);
```
jenswi-linaro, line 500: Please replace `void` with `uint8_t` since we don't allow pointer arithmetic with void pointers.

raymo200915: Fixed by using `transfer_list_entry_data()` instead.

mk/config.mk
```diff
@@ -513,11 +513,10 @@ endif
...
 CFG_TRANSFER_LIST ?= n
-ifeq ($(CFG_TRANSFER_LIST),y)
-$(call force,CFG_DT,y)
-$(call force,CFG_EXTERNAL_DT,y)
-$(call force,CFG_MAP_EXT_DT_SECURE,y)
-endif
+$(eval $(call cfg-enable-all-depends,CFG_TRANSFER_LIST, \
+	 CFG_DT CFG_EXTERNAL_DT CFG_MAP_EXT_DT_SECURE))
+
+CFG_TRANSFER_LIST_TEST ?= $(call cfg-all-enabled,CFG_TRANSFER_LIST)
```
jenswi-linaro, line 519: `CFG_TRANSFER_LIST_TEST ?= $(CFG_TRANSFER_LIST)` is enough.

core/pta/tests/sub.mk
```diff
@@ -6,3 +6,4 @@ cflags-misc.c-y += -fno-builtin
 srcs-y += mutex.c
 srcs-y += aes_perf.c
 srcs-$(CFG_DT_DRIVER_EMBEDDED_TEST) += dt_driver_test.c
+srcs-$(CFG_TRANSFER_LIST_TEST) += tl.c
```
jenswi-linaro, line 9: `tl.c` is a bit short. Please make it `transfer_list.c`.

mk/config.mk
raymo200915: Fixed.

core/pta/tests/sub.mk
raymo200915: Fixed.

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,247 @@
...
+
+static TEE_Result transfer_list_tests(void)
+{
+	uint8_t *va_tl;
+	paddr_t pa_tl;
+	struct transfer_list_header *tl;
+	struct transfer_list_entry *te1, *te2, *te3;
+	TEE_Result ret;
+
+	va_tl = memalign(TL_ALIGNMENT_FROM_ORDER(TRANSFER_LIST_INIT_MAX_ALIGN),
```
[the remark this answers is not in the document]
raymo200915: But `nex_phys_mem_core_alloc()` has not an arg for a user specified alignment, is there any other alternatives I can take?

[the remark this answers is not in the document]
jenswi-linaro: You get full pages, so the alignment shouldn't be an issue. Please use `phys_mem_core_alloc ()` instead of `nex_phys_mem_core_alloc ()`. I made a typo above.

core/pta/tests/invoke.c
```diff
@@ -433,6 +433,10 @@ static TEE_Result invoke_command(void *pSessionContext __unused,
 		return core_aes_perf_tests(nParamTypes, pParams);
 	case PTA_INVOKE_TESTS_CMD_DT_DRIVER_TESTS:
 		return core_dt_driver_tests(nParamTypes, pParams);
+#if defined(CFG_TRANSFER_LIST_TEST)
+	case PTA_INVOKE_TESTS_CMD_TRANSFER_LIST_TESTS:
+		return core_transfer_list_tests(nParamTypes, pParams);
+#endif
```
jforissier, line 439: There is a problem when `CFG_TRANSFER_LIST_TEST` is disabled: `xtest` (as it is currently in https://github.com/OP-TEE/optee_test/pull/788) will report a failure. One option is to return success (see `core_dt_driver_tests()`). But
TBH I'd rather we return `TEE_ERROR_NOT_SUPPORTED` and then have `xtest` print the proper warning, I think it makes more sense.

raymo200915: In case `CFG_TRANSFER_LIST_TEST` is disabled, how about to print warnings in `core_dt_driver_tests()` but still return success? This can avoid having a new `Do_ADBG_#` function in optee_test.

raymo200915: Please see commit https://github.com/OP-TEE/optee_test/pull/788/commits/388bd391ad7d63efd0a1ed8cc5c15e7b0f28eec0
I have added macros and helper function to avoid returning error for not supported tests.

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,247 @@
...
+			      uint16_t tag_id, uint32_t data_size,
+			      const void *data, uint8_t align,
+			      struct transfer_list_entry **tle)
+{
+	uint8_t *te_dat;
+	bool new_max_align = false;
+	vaddr_t old_tl_ev = (vaddr_t)tl + tl->size;
+	struct transfer_list_entry *tl_e;
+
+	if (align > tl->alignment)
```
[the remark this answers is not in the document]
raymo200915: Good idea. Done.

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,247 @@
...
+static const char test_str1[] = "first added entry";
+static const char test_str2[] = "second added entry";
+static const char test_str3[] = "last added entry";
+
+static TEE_Result test_add_te(struct transfer_list_header *tl,
+			      uint16_t tag_id, uint32_t data_size,
+			      const void *data, uint8_t align,
+			      struct transfer_list_entry **tle)
+{
+	uint8_t *te_dat;
```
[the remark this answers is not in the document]
raymo200915: Done.

core/pta/tests/transfer_list.c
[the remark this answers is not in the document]
raymo200915: Done.

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,242 @@
...
+	if (!IS_ALIGNED(mov_dis, TL_ALIGNMENT_FROM_ORDER(tl->alignment)))
+		return TEE_ERROR_CORRUPT_OBJECT;
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result transfer_list_tests(void)
+{
+	struct transfer_list_header *tl = NULL;
+	struct transfer_list_entry *te1 = NULL, *te2 = NULL, *te3 = NULL;
```
jenswi-linaro, line 147: Please declare one local variable per line, like:
``` C
struct transfer_list_entry *te1 = NULL;
struct transfer_list_entry *te2 = NULL;
...
```

core/pta/tests/invoke.c
jforissier: Good, thank you.

jforissier:
For "core: kernel: fix bug in transfer_list_add" with the fixup:
Please add parentheses in the commit subject: `transfer_list_add()`, then:

jforissier:
For commit "core: pta: add self tests for transfer list" with the fixups and Jens' comment addressed:

mk/config.mk
```diff
@@ -513,11 +513,10 @@ endif
...
 CFG_TRANSFER_LIST ?= n
-ifeq ($(CFG_TRANSFER_LIST),y)
-$(call force,CFG_DT,y)
-$(call force,CFG_EXTERNAL_DT,y)
-$(call force,CFG_MAP_EXT_DT_SECURE,y)
-endif
+$(eval $(call cfg-enable-all-depends,CFG_TRANSFER_LIST, \
+	 CFG_DT CFG_EXTERNAL_DT CFG_MAP_EXT_DT_SECURE))
+
+CFG_TRANSFER_LIST_TEST ?= $(CFG_TRANSFER_LIST)
```
etienne-lms, line 519: A short description would be nice. Maybe mention that `CFG_TRANSFER_LIST_TEST` is not effective when `CFG_ENABLE_EMBEDDED_TESTS` is disabled.

etienne-lms, line 519: A short description would be nice. Maybe tell that `CFG_ENABLE_EMBEDDED_TESTS` needs to be enabled to have the test embedded.

etienne-lms: > I guess you mean CFG_TEE_CORE_EMBED_INTERNAL_TESTS?

Yes, sorry.

> I can replace it with:
> (...)

Sound good to me, still with a short description for `CFG_TRANSFER_LIST_TEST`, like:
`# Embed transfer list support self test when enabled`

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,242 @@
...
+	if (transfer_list_check_header(tl) == TL_OPS_NONE) {
+		ret = TEE_ERROR_CORRUPT_OBJECT;
+		goto unmap_tl;
+	}
+
+	if (tl->hdr_size != sizeof(*tl) ||
+	    tl->alignment != TRANSFER_LIST_INIT_MAX_ALIGN ||
+	    tl->size != sizeof(*tl) ||
+	    tl->max_size != TEST_TL_MAX_SIZE ||
+	    tl->flags != TL_FLAGS_HAS_CHECKSUM) {
```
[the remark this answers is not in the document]
raymo200915: Added.

core/pta/tests/transfer_list.c
raymo200915: Done.

mk/config.mk
raymo200915: Done.

raymo200915:
All comments are addressed.
All commits are squashed and rebased.
Added tag `Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>` to commit "core: kernel: fix bug in transfer_list_add()" and `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>` to commit "core: pta: add self tests for transfer list"

jenswi-linaro (commented):

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,245 @@
...
+	vaddr_t old_tl_ev = (vaddr_t)tl + tl->size;
+	struct transfer_list_entry *tl_e = NULL;
+
+	old_tl_align = tl->alignment;
+
+	if (!align)
+		tl_e = transfer_list_add(tl, tag_id, data_size, data);
+	else
+		tl_e = transfer_list_add_with_align(tl, tag_id, data_size, data,
+						    align);
```
etienne-lms, line 43: Nitpicking: calling only `transfer_list_add_with_align(..., align)` should be enough.

core/pta/tests/transfer_list.c
```diff
@@ -0,0 +1,245 @@
...
+	if (!transfer_list_set_data_size(tl, tl_e, new_data_size))
+		return TEE_ERROR_GENERIC;
+
+	if (!transfer_list_verify_checksum(tl))
+		return TEE_ERROR_CORRUPT_OBJECT;
+
+	new_te_next = transfer_list_next(tl, tl_e);
+
+	/* skip the inserted void entry if it exists */
+	if (new_te_next->tag_id == TL_TAG_EMPTY)
```
etienne-lms, line 126: Suggestion in case there are several empty slot + in case there is no next entry:
```patch
-	if (new_te_next->tag_id == TL_TAG_EMPTY)
+	while (new_te_next && new_te_next->tag_id == TL_TAG_EMPTY)
```

Bythe way, how come `new_te_next` is not NULL when this function is called by `transfer_list_tests()` at line 215 to change `TEST_TE3_ID` data size.

etienne-lms (commented):
"core: kernel: fix bug in transfer_list_add()".

Nitpicking comment/question for commit "core: pta: add self tests for transfer list". Otherwise LGTM.

core/pta/tests/transfer_list.c
raymo200915: Here is just to test different API entries.

core/pta/tests/transfer_list.c
raymo200915: Good question. When expending the data region of an entry, it tries to ending up with the address that fits the potential 'next' entry with the data alignment required, even when it is the last entry - That is the case you see in line 215, a void entry was added at the tail.
It is not bad but I think it is fair enough to check if it is the last one and avoid to filling the gap with the void entry.
I will add a small fix.

jforissier (commented):
For commit "core: kernel: remove the last appended void transfer entry":

raymo200915:
All tags are added.

PR #7353, opened 2025-04-06
ta: pkcs11: Increase destination key size for TEE_AsymmetricDecrypt

anvisriv:
This update increases the destination key size for the TEE_AsymmetricDecrypt function within the unwrap_rsa_aes_key function. According to the GP specification, the destination size must be greater than or equal to the source length.

### DIFF
--- a/ta/pkcs11/src/processing_asymm.c
+++ b/ta/pkcs11/src/processing_asymm.c
@@ -1179,7 +1179,7 @@ static enum pkcs11_rc unwrap_rsa_aes_key(struct active_processing *proc,
 	struct rsa_aes_key_wrap_processing_ctx *ctx = proc->extra_ctx;
 	mbedtls_nist_kw_context kw_ctx = { };
 	uint8_t aes_key_value[32] = { };
-	size_t aes_key_size = ctx->aes_key_bits / 8;
+	size_t aes_key_size = 0;
 	uint32_t wrapped_key_size = 0;
 	uint32_t rsa_key_size = 0;
 	size_t target_key_size = 0;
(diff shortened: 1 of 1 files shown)


anvisriv:
Hi @etienne-lms
Can you please review the change?

anvisriv:
Hi @etienne-lms 
Any update?

ta/pkcs11/src/processing_asymm.c
```diff
@@ -1193,6 +1193,7 @@ static enum pkcs11_rc unwrap_rsa_aes_key(struct active_processing *proc,
 	rsa_key_size = info.keySize / 8;
 	wrapped_key_size = data_sz - rsa_key_size;
 	target_key_size = wrapped_key_size - 8;
+	aes_key_size = rsa_key_size;
```
etienne-lms, line 1196: Looking at your change, whether it's applied or not, I think we are missing a test after line 1206, to at least check that the decrypted temporary key has the expected size:
```patch
 	res = TEE_AsymmetricDecrypt(proc->tee_op_handle,
 				    NULL, 0,
 				    data, rsa_key_size,
 				    aes_key_value, &aes_key_size);
 	if (res) {
 		rc = tee2pkcs_error(res);
 		goto out;
 	}
+	if (aes_key_size != ctx->aes_key_bits / 8) {
+		DMSG("Unexpected temporary key size %zu, expecting %"PRIu32,
+		     aes_key_size, ctx->aes_key_bits / 8);
+		return PKCS11_CKR_WRAPPED_KEY_INVALID;
+	}

etienne-lms (commented):
Sorry for this later feedback.

In the commit message, you say that "According to the GP specification, the destination size must be greater than or equal to the source length.".  Where did you see this constraint? I can't find it.

anvisriv:
Commenting it so that issue is not marked stale

anvisriv:
Keep it alive

jenswikl:
@anvisriv, do you plan to complete this PR?

anvisriv:
Hi @jenswi-linaro 
I haven’t been able to look into this and other PRs due to some internal work and other pressing priorities over the past few weeks. That said, I’ll be aiming to revisit and work toward closing this PR in the upcoming weeks.

PR #7354, opened 2025-04-06
ta: pkcs11: Update PKCS11 ECDSA flags to align with GP Specification

anvisriv:
This commit updates the PKCS11 ECDSA-related flags to ensure they map correctly according to the GlobalPlatform (GP) specifications. This alignment helps maintain consistency and compatibility with the GP standards.

Addresses https://github.com/OP-TEE/optee_test/issues/721#issuecomment-1970596624


anvisriv:
Same PR as https://github.com/OP-TEE/optee_os/pull/7226

anvisriv:
Hi @etienne-lms 
Any update?

ta/pkcs11/src/processing_ec.c
```diff
@@ -847,16 +854,20 @@ size_t ecdsa_get_input_max_byte_size(TEE_OperationHandle op)
 
 	TEE_GetOperationInfo(op, &info);
 
-	switch (info.algorithm) {
-	case TEE_ALG_ECDSA_P192:
+	/*
+	 *  Determining curve size in bytes with the help of
+	 *  maxkeysize attribute instead of algorithm
```
etienne-lms, line 859: The wording *"instead of algorithm"* relates to the patch rather than to the function implementation.
Could you update the inline comment. Maybe `/* EDCSA input data max size is the curve size */`.

Also, can you update the error trace message at line 873:
```patch
-		DMSG("Unexpected ECDSA algorithm %#"PRIx32, info.algorithm);
+		DMSG("Unexpected ECDSA key size %#"PRIx32, info.maxKeySize);

etienne-lms (commented):
Sorry for this late feedback.
I'm fine with the proposed change. It does not address the GP requirement on `TEE_AsymmetricSign()` and `TEE_AsymmetricVerifyDigest()` that both expect to error *"If digestLen is not equal to the hash size of the algorithm in non-XOF functions"*. That said, this can be addressed seperately even if there are compatibility issues with how legacy IDs  `TEE_ALG_ECDSA_P*` are handled.

Could you please rebase commit "ta: pkcs11: Update PKCS11 ECDSA flags to align with GP Specification" instead of using a merge commit?

See also my review comments.

ta/pkcs11/src/processing_ec.c
anvisriv: Hi @etienne-lms 
Thanks for your response. I will update the patch.

anvisriv:
Commenting it so that issue is not marked stale

anvisriv:
Keep it alive

jenswikl:
@anvisriv, you could also address the review comments so it can be merged.

PR #7357, opened 2025-04-09
core: fix discovered ns-mem check

jenswi-linaro:
When discovering or assigning available non-secure physical memory it's checked against overlaps with other memory types. Memory types reserving virtual memory space should be excluded including the two recently added types MEM_AREA_NEX_DYN_VASPACE and MEM_AREA_TEE_DYN_VASPACE. This was missed when the memory types where added so add the check to exclude them now.

This fixes an error like:
E/TC:0   check_phys_mem_is_outside:455 Non-sec mem (0:0x60000000) overlaps map (type 10 0:0x100000) E/TC:0   Panic at core/mm/core_mmu.c:459 <check_phys_mem_is_outside>

Fixes: 96f43358c593 ("core: add nex_dyn_vaspace and tee_dyn_vaspace areas")


etienne-lms (commented):

PR #7364, opened 2025-04-14
drivers: sam: fix the failure of getting proper intrerupts from dt

TonyHan11:
The issue is found on sama5d2 platform.
Get interrupt fails due to the change of function parameter (from count by bytes to count by words), fixed by making corresponding changes to the function called later.

Fixes: 63873401c ("core: interrupt: fix property count in dt_get_irq_type_prio()")


TonyHan11:
Update the commit message with longer commit ID following "Fixes:".

jforissier (commented):
Misspelling in the commit subject. I suggest: "drivers: sam: fix getting interrupts from DT".

TonyHan11:
Updated, thank you very much.

Issue #7363, opened 2025-04-11
QEMUv8: boot hangs with CFG_CORE_DEBUG_CHECK_STACKS=y CFG_STACK_THREAD_EXTRA=256


jforissier:
Platform manifest: qemu_v8.xml

```
$ make CFG_CORE_DEBUG_CHECK_STACKS=y CFG_STACK_THREAD_EXTRA=256 check
[...]
Starting QEMU...
```
`out/bin/serial0.log` shows that the boot hangs after:
```
NOTICE:  BL31: Built : 15:04:15, Dec  9 2024
```
4.5.0 boots OK. I will bisect.

jforissier:
Bisected:
```
commit ea991d7459f6f9e268769e8d7c7f30c3b1ca0169
Author: Jens Wiklander <jens.wiklander@linaro.org>
Date:   Fri Mar 21 11:26:25 2025 +0100

    core: arm: remove THREAD_CORE_LOCAL_STACKCHECK_RECURSION

    THREAD_CORE_LOCAL_STACKCHECK_RECURSION was introduced in the commit
    b5ec8152f3e5 ("core: arm: refactor boot"). However, clearing the
    stackcheck_recursion flag from assembly during boot isn't needed since
    the stack pointer is set up in synch with the recorded information in
    thread_core_local. So remove the unnecessary clearing and remove
    THREAD_CORE_LOCAL_STACKCHECK_RECURSION.

    Reported-by: Alvin Chang <alvinga@andestech.com>
    Closes: https://github.com/OP-TEE/optee_os/commit/b5ec8152f3e5ad8cc111952f0483f5cf903aac7c#r154088026
    Signed-off-by: Jens Wiklander <jens.wiklander@linaro.org>
    Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>

 core/arch/arm/kernel/asm-defines.c |  4 ----
 core/arch/arm/kernel/entry_a32.S   | 13 -------------
 core/arch/arm/kernel/entry_a64.S   |  7 -------
 3 files changed, 24 deletions(-)
```

jenswi-linaro:
This should be fixed by https://github.com/OP-TEE/optee_os/pull/7365

PR #7365, opened 2025-04-14
core: arm: update recorded SP first after MMU is enabled

jenswi-linaro:
With CFG_CORE_ASLR=y, stored addresses must be updated after MMU has been enabled to match the map offset. In particular the recorded stack pointers in thread_core_local[] must be updated to match the new offset before any calls can be done into C code or check_stack_limits() with CFG_CORE_DEBUG_CHECK_STACKS=y might catch an inconsistent stack pointer.

Currently, boot_mem_relocate() is called before the recorded stack pointers have been updated and causes a crash with CFG_CORE_ASLR=y and CFG_CORE_DEBUG_CHECK_STACKS=y. So fix this by calling delaying the call to boot_mem_relocate() to after the stack pointers in thread_core_local[] has been updated.

Closes: https://github.com/OP-TEE/optee_os/issues/7363
Fixes: ea991d7459f6 ("core: arm: remove THREAD_CORE_LOCAL_STACKCHECK_RECURSION")

### DIFF
--- a/core/arch/arm/kernel/entry_a32.S
+++ b/core/arch/arm/kernel/entry_a32.S
@@ -568,12 +568,12 @@ shadow_stack_access_ok:
 	bl	__get_core_pos
 	bl	enable_mmu
 #ifdef CFG_CORE_ASLR
-	ldr	r0, =boot_mmu_config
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
Tags applied

Issue #7216, opened 2025-01-10
QEMUv8: core data abort with ftrace


jforissier:
Platform: QEMUv8
Build command: `make CFG_FTRACE_SUPPORT=y CFG_SYSCALL_FTRACE=y CFLAGS_ta_arm32=-pg CFLAGS_ta_arm64=-pg check`
The test times out due to a core crash. Symbolized stack from `out/bin/serial0.log` is mostly useless due to the ftrace instrumentation.
```
E/TC:0 0 Core data-abort at address 0x10 (translation fault)
E/TC:0 0  esr 0x96000006  ttbr0 0x400000e213000   ttbr1 0x00000000   cidr 0x0
E/TC:0 0  cpu #0          cpsr 0x80000004
E/TC:0 0  x0  2f736c6974756269 x1  0000000000000460
E/TC:0 0  x2  0000000000000bfa x3  00000000000002b0
E/TC:0 0  x4  00000000db0816f0 x5  00000000dafd500c
E/TC:0 0  x6  00000000db041168 x7  00000000d8001930
E/TC:0 0  x8  00000000db0412c8 x9  00000000db037810
E/TC:0 0  x10 00000000db0412a8 x11 00000000db036ee0
E/TC:0 0  x12 0000000000000001 x13 00000000daf6dbb0
E/TC:0 0  x14 00000000db081400 x15 0000000000000000
E/TC:0 0  x16 0000000000000000 x17 0000000000000000
E/TC:0 0  x18 0000000000000000 x19 00000000db027000
E/TC:0 0  x20 00000000db041168 x21 00000000db0412a8
E/TC:0 0  x22 000000000000000a x23 00000000db0412c8
E/TC:0 0  x24 00000000000000f0 x25 00000000db0813e8
E/TC:0 0  x26 00000000db0412a8 x27 00000000db041298
E/TC:0 0  x28 0000000000000010 x29 00000000db081390
E/TC:0 0  x30 0000000000000000 elr 00000000dafd5288
E/TC:0 0  sp_el0 0000000000000000
E/TC:0 0 TEE load address @ 0xdaf69000
E/TC:0 0 Call stack:
E/TC:0 0  0xdafd5288 ftrace_return at optee_os/lib/libutils/ext/ftrace/ftrace.c:136
E/TC:0 0  0xdafd5008 _mcount at optee_os/lib/libutils/ext/arch/arm/mcount_a64.S:55
[...]

PR #7368, opened 2025-04-16
core: fix crash during syscall ftrace

jenswi-linaro:
Syscall ftrace collects data during a syscall. get_fbuf() checks if thread_get_id_may_fail() != -1 to see if a function is called under normal thread execution. This can lead to an inconsistent state if a native interrupt occur while ftrace_enter() or ftrace_return() is recording data in the ftrace buffer. So fix this by using thread_is_in_normal_mode() to exclude ftrace during interrupt processing.

Closes: https://github.com/OP-TEE/optee_os/issues/7216
Fixes: 099918f6744c ("ftrace: Add support for syscall function tracer")

### DIFF
--- a/core/arch/arm/kernel/thread.c
+++ b/core/arch/arm/kernel/thread.c
@@ -439,7 +439,7 @@ vaddr_t thread_get_saved_thread_sp(void)
 #endif /*ARM64*/
 
 #ifdef ARM32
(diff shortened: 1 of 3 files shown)


jforissier:
Nice fix!

jenswi-linaro:
Tags applied.

PR #7369, opened 2025-04-16
Crypto legacy algs

jbech-linaro:
Some updates to stay aligned with the [crypto recommendations from GlobalPlatform](https://globalplatform.org/specs-library/globalplatform-technology-cryptographic-algorithm-recommendations/). Three patches:
* add SHA-224 to the list of deprecated algorithms
* updates the key size to 4096bits for our default key.
* add a function to check for deprecated and legacy key sizes.

There will be a dependency between this PR and https://github.com/OP-TEE/optee_test/pull/792 in optee_test, since I had to re-generate the subkeys.

I think we should wait merge this until after the ongoing release.


core/crypto/signed_hdr.c
```diff
@@ -52,7 +52,63 @@ struct shdr *shdr_alloc_and_copy(size_t offs, const void *img, size_t img_size)
 static bool is_weak_hash_algo(uint32_t algo)
 {
 	return algo == TEE_ALG_MD5 || algo == TEE_ALG_SHA1 ||
-	       algo == TEE_ALG_MD5SHA1;
+	       algo == TEE_ALG_MD5SHA1 || algo == TEE_ALG_DSA_SHA224 ||
+	       algo == TEE_ALG_HMAC_SHA224 ||
+	       algo == TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA224 ||
+	       algo == TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA224 ||
+	       algo == TEE_ALG_RSASSA_PKCS1_V1_5_SHA224 ||
```
jbech-linaro, line 59: These are not `hash_algo`, they are `shdr->algo`. So, this compiles, but isn't correct, since I believe that this call in `shdr_verify_signature`
```c
	hash_algo = TEE_DIGEST_HASH_TO_ALGO(shdr->algo);
```
will never return these three. I added it still to get comments from the rest of you. If you have any ideas how to approach this.

core/crypto/signed_hdr.c
```diff
@@ -52,7 +52,63 @@ struct shdr *shdr_alloc_and_copy(size_t offs, const void *img, size_t img_size)
...
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA224:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA256:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA384:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA512:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_224:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_256:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_384:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_512:
+	case TEE_ALG_RSA_NOPAD:
+		if (key_size_bits <= 2048)
```
jbech-linaro, line 103: Is hardcoded value OK here (and below)? Or should I define them? If yes, where?

keys/default.pem
```diff
@@ -1,27 +1,52 @@
...
-xKizXPR7oqYpm4WucYDB/zRHPjSh41CYQEQ6sOAArUYcl4pmqmN5oSKOGFX37ASx
-oiEaKV9vMuIGIp+1A4NGvyc734FTKmo04Kv3d/mCR38bZINeYscFVou11EzTgXZg
-fZjwDFnooH8XcUCAHZuu3zECgYEAt6pltSMcgE7Z9JdIas/TI5vHLQEtgzhdH+bm
-SHUjRaWnekvy3VWW7deVGtMMlmi0V+US0iVgl5rXposv+gJfh0HNUrTSg2qZiWlm
-MQ+/Zb2BoeZv/A8Wdfnpue+svP+Hq8gwC6mlcJI7RFk1uw8LiXxwfWuFRv9LqMPH
-wBBnar0CgYAJF3ukzm7rnut19lzS4r1uEIuuPwk/DjEkkil7U9wE5fr+ds5rCtXj
-9591DLHrtE/CBOWym2+tIDr7OnmfU+934524mtLWVZdf39/DFM9Fc1VaSualrj/1
-BOLhBb7MEZqN7hHcQP7+49YW72GgX3pm0Ov5ke33FzN3ubIwAS6T7A==
------END RSA PRIVATE KEY-----
+-----BEGIN PRIVATE KEY-----
```
jbech-linaro, line 1: Maybe I should say something in the commit message that this is a "test" private key not to be used other than for testing purposes. Various automatic tools tends to find private keys and report them as security issues etc.

core/crypto/signed_hdr.c
jenswi-linaro: It should be enough to add `algo == TEE_ALG_SHA224`, all the combined algorithms should be detected automatically.

keys/default.pem
jenswi-linaro: Yeah, that makes sense.

core/crypto/signed_hdr.c
```diff
@@ -60,6 +60,57 @@ static bool is_weak_hash_algo(uint32_t algo)
...
 
+/*
+ * Checks for key sizes in bits, that are either depracted or put into legacy
+ * by national standards bodies. For deprecated key sizes return true. For
+ * legacy key sizes, return false, but print an informative message if
+ * CFG_WARN_INSECURE was set when compiling the TEE.
+ */
+static bool is_weak_key_size(uint32_t algo, size_t key_size_bits)
+{
+	switch (algo) {
```
jenswi-linaro, line 71: Do you mean `if (TEE_ALG_GET_MAIN_ALG(algo) == TEE_MAIN_ALGO_RSA))`?

core/crypto/signed_hdr.c
```diff
@@ -60,6 +60,57 @@ static bool is_weak_hash_algo(uint32_t algo)
...
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_224:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_256:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_384:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_512:
+	case TEE_ALG_RSA_NOPAD:
+		if (key_size_bits <= 2048)
+			return true;
+
+		if (IS_ENABLED(CFG_WARN_INSECURE) && key_size_bits < 3000)
+			IMSG("WARNING: Using legacy key size for RSA");
```
jenswi-linaro, line 107: This warning doesn't make sense. The problem is that we accept weak keys, not whether this particular key is weak. If we need to warn about this, it should be done `boot_init_primary_runtime()` when we complain about `CFG_INSECURE`.

core/crypto/signed_hdr.c
```diff
@@ -60,6 +60,57 @@ static bool is_weak_hash_algo(uint32_t algo)
...
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA512:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_224:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_256:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_384:
+	case TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA3_512:
+	case TEE_ALG_RSA_NOPAD:
+		if (key_size_bits <= 2048)
+			return true;
+
+		if (IS_ENABLED(CFG_WARN_INSECURE) && key_size_bits < 3000)
```
jenswi-linaro, line 106: The name is `CFG_INSECURE` now.

core/crypto/signed_hdr.c
jbech-linaro: If that as we works as a "catch-all" for all of the ones below, then absolutely yes! Can you confirm that this is the case?

jenswi-linaro: Yes, and this is also a macro that we maintain and extend as needed, so we should normally not need to revisit this function if another RSA algorithm is ever added.

jenswi-linaro:
The diff looks good. Please fix up the commit messages, etc.

core/crypto/signed_hdr.c
jbech-linaro: I removed it.

jbech-linaro:
> The diff looks good. Please fix up the commit messages, etc.

Please have a look at the updated commit messages.

core/crypto/signed_hdr.c
```diff
@@ -75,6 +87,9 @@ TEE_Result shdr_verify_signature(const struct shdr *shdr)
 	if (is_weak_hash_algo(hash_algo))
 		goto err;
 
+	if (is_weak_key_size(shdr->algo, ta_pub_key_modulus_size * 8))
```
jenswi-linaro, line 90: Shouldn't we also have this check in `shdr_verify_signature2()`?

jenswi-linaro (commented):
For commit "shdr: add check for weak key sizes"

Please mention that this only affects verifying TA and subkey signatures.

core/crypto/signed_hdr.c
jbech-linaro: > Shouldn't we also have this check in `shdr_verify_signature2()`?

Yes, I'll add the same check there as well.

jbech-linaro: Done!

jbech-linaro:
Added check to the `shdr_verify_signature2` as well and added the suggested note about TA's and subkeys to the commit message (force pushed to show the commit message as intended).

jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready to be merged together with https://github.com/OP-TEE/optee_test/pull/792 after the ongoing release! Thanks!

jbech-linaro:
@jforissier I've just rebased and pushed this and the related test PR. The CI check will fail, since there is a strong dependency between this the patches in `optee_test`. So, they should be merged together. It should be ready to merge.

PR #7376, opened 2025-04-23
core: fs_htree: Fix AAD length when CFG_REE_FS_HTREE_HASH_SIZE_COMPAT=y

etienne-lms:
Correct the hash size declared in AAD length declared in REE FS hash tree authentication sequence when CFG_REE_FS_HTREE_HASH_SIZE_COMPAT is enabled in which case the hash is truncated to the size of the FEK key (TEE_FS_HTREE_FEK_SIZE).

The issue has currently no impact since REE FS hash tree authentication is based on AES-GCM but it would be of importance if, for example, one moves to an AES-CCM scheme while still enabling
CFG_REE_FS_HTREE_HASH_SIZE_COMPAT (even if unlikely to happen). To prevent such issue in the future, let's declare the effectively used hash size.

Suggested-by: Jens Wiklander <jens.wiklander@linaro.org>
Link: https://github.com/OP-TEE/optee_os/pull/7340/commits/087325faec7c057a638cca80f0549e9abe49f190#r2024716984

### DIFF
--- a/core/tee/fs_htree.c
+++ b/core/tee/fs_htree.c
@@ -454,7 +454,17 @@ static TEE_Result authenc_init(void **ctx_ret, TEE_OperationMode mode,
 		iv = ni->iv;
 	} else {
 		iv = ht->head.iv;
-		aad_len += TEE_FS_HTREE_HASH_SIZE + sizeof(ht->head.counter);
+		aad_len += sizeof(ht->head.counter);
+
+		/*
+		 * With CFG_REE_FS_HTREE_HASH_SIZE_COMPAT, hash data passed
+		 * for AAD is truncated to TEE_FS_HTREE_FEK_SIZE bytes so
+		 * use the correct size aad_len computation.
+		 */
+		if (IS_ENABLED(CFG_REE_FS_HTREE_HASH_SIZE_COMPAT))
+			aad_len += TEE_FS_HTREE_FEK_SIZE;
+		else
+			aad_len += TEE_FS_HTREE_HASH_SIZE;
 	}
 
 	if (mode == TEE_MODE_ENCRYPT) {


etienne-lms:
Minor fix, no functional change. 
Related to discussion thread in https://github.com/OP-TEE/optee_os/pull/7340.

core/tee/fs_htree.c
```diff
@@ -454,7 +454,12 @@ static TEE_Result authenc_init(void **ctx_ret, TEE_OperationMode mode,
 		iv = ni->iv;
 	} else {
 		iv = ht->head.iv;
-		aad_len += TEE_FS_HTREE_HASH_SIZE + sizeof(ht->head.counter);
+		aad_len += sizeof(ht->head.counter);
+
+		if (IS_ENABLED(CFG_REE_FS_HTREE_HASH_SIZE_COMPAT))
```
jenswi-linaro, line 459: Please add a comment explaining that this matches the update of `hash_size` to initialize with the correct AAD length.

core/tee/fs_htree.c
```diff
@@ -456,6 +456,11 @@ static TEE_Result authenc_init(void **ctx_ret, TEE_OperationMode mode,
 		iv = ht->head.iv;
 		aad_len += sizeof(ht->head.counter);
 
+		/*
+		 * With CFG_REE_FS_HTREE_HASH_SIZE_COMPAT, hash data passed
+		 * for AAD is truncated to TEE_FS_HTREE_FEK_SIZE bytes so
+		 * use the relevant size aad_len computation.
```
jenswi-linaro, line 462: s/relevant/correct/

jenswi-linaro (commented):

etienne-lms:
Inline comment fixed, fixup commit squashed and review tag applied.

PR #7377, opened 2025-04-23
plat-k3: drivers: add platform flavors for 62A and 62P devices

bryanbrattlof:
Even though the SA2UL integration on the AM62Ax and AM62Px platforms are functionally identical to the AM62x platforms many, when building OP-TEE manually, are using the platform name they are building for and not 'am62x' which leaves SA2UL_BASE undefined and to failed builds:

    In file included from core/include/mm/core_memprot.h:9,
                     from core/include/kernel/interrupt.h:10,
                     from core/arch/arm/plat-k3/drivers/sa2ul_rng.c:12:
    core/arch/arm/plat-k3/./platform_config.h:91:34: error: ‘SA2UL_BASE’ undeclared here (not in a function); did you mean ‘SCU_BASE’?
       91 | #define RNG_BASE                (SA2UL_BASE + 0x10000)
          |                                  ^~~~~~~~~~

For now let's just define the AM62Ax and AM62Px platform flavors identical to how AM62x is defined and include an #else statement to catch when a undefined platform flavor tries to build the SA2UL driver

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul.c
@@ -121,7 +121,10 @@ static TEE_Result sa2ul_init(void)
 	start_address = RNG_BASE;
 	end_address = RNG_BASE + RNG_REG_SIZE - 1;
 	permissions[num_perm++] = (FW_BIG_ARM_PRIVID << 16) | FW_SECURE_ONLY;
-#if defined(PLATFORM_FLAVOR_am62x)
+#if defined(PLATFORM_FLAVOR_am62x) || \
+	defined(PLATFORM_FLAVOR_am62ax) || \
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
Wouldn't it also make sense to catch unknown platform flavors?

bryanbrattlof:
Ah I didn't really point out that logic too much in the commit but yes. The real fix is adding that `#error` statement at the end of the SA2UL definitions block to catch what is most likely incorrect build arguments coming from the command line. I added the `!defined(CFG_WITH_SOFTWARE_PRNG)` there mainly so platform flavors that still use software based rng, like the `am62lx` wouldn't need to be modified

jenswi-linaro:

bryanbrattlof:
Picked up your trailers @jenswi-linaro and @glneo. Thanks for the reviews

Issue #7374, opened 2025-04-22
Possible performance regression


jenswi-linaro:
The commit d3ec4328a070 ("libmbedtls: config: disable long long divisions") causes drastically increased running times for `xtest -l15` on some platforms.

I've tested this on Hikey620 first with the commit 45c754cea36a ("core: fix crash during syscall ftrace")
``` shell
time xtest -l15
real    16m 52.78s
user    0m 6.15s
sys     14m 25.60s
```
and then with the two commit 45fecab08117 ("Deprecate libgcc for OP-TEE core and ldelf") and d3ec4328a070 ("libmbedtls: config: disable long long divisions") reverted on top
``` shell
time xtest -l15
real    12m 59.59s
user    0m 5.99s
sys     13m 55.95s
```

jforissier:
The regression is noticeable on QEMUv8 too. For example, running `xtest 4006` takes 45.8 seconds on master versus 35.8 with the two commits above reverted.

jenswi-linaro:
It is surprising since `__udivti3()` is a pure software implementation. It suggests that the bignum divisions in Mbedtls have room for optimization. Fixing that might be harder than finding a good `__udivti3()` implementation, even if the potential for increased performance might be greater in a bignum implementation.

Reverting the problematic commits fixes this problem, but it may cause other concerning BTI.

jforissier:
I agree. Anyways, https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/umodti3.c may be a good starting point.

jenswi-linaro:
Replying to @sahilnxp https://github.com/OP-TEE/optee_os/pull/7359#issuecomment-2821716011
> Hi @jenswi-linaro
> 
> Did you notice the change in `xtest` running time on `imx-mx8mqevk` ? In my case it increased from `16mins` to `22mins`. I am running command `xtest l -l15 -x 8001`

I didn't time it. Let me check again.

> 
> Also In the commit [d3ec432](https://github.com/OP-TEE/optee_os/commit/d3ec4328a0704e16b30fde84d7458ef8e83eb79d) It is mentioned that it will impact only Aarch64 platforms, but i.MX6DL is Arm32 platform.
> 
> Any idea why it is impacting Arm32 platforms also ?

[...]

PR #7380, opened 2025-04-28
Crt builtins for long long division

jenswi-linaro:
This is trying to address https://github.com/OP-TEE/optee_os/issues/7374

### DIFF
--- a/lib/libutils/sub.mk
+++ b/lib/libutils/sub.mk
@@ -1,2 +1,3 @@
 subdirs-$(CFG_LIBUTILS_WITH_ISOC) += isoc
 subdirs-y += ext
+subdirs-$(CFG_LIBUTILS_WITH_COMPILER_RT) += compiler-rt
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -251,6 +251,15 @@ TA_PUBLIC_KEY ?= $(TA_SIGN_KEY)
 # may not because they obtain the isoc functions from elsewhere
 CFG_LIBUTILS_WITH_ISOC ?= y
 
+# Include lib/libutils/compiler-rt in the build. Most platforms need this.
+# Provides some functions called "compiler builtins", which the compiler
+# may invoke to perform low-level operations such as long long division
+# etc. Such functions typically come with compiler runtime libraires (GCC
+# has libgcc, Clang has compiler-rt). OP-TEE often can't use them because
+# they may be Linux-specific or bring unwanted dependencies. Therefore,
+# this imports and builds only what's needed.
+CFG_LIBUTILS_WITH_COMPILER_RT ?= y
+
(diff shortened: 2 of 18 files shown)


lib/libutils/sub.mk
```diff
@@ -1,2 +1,3 @@
 subdirs-$(CFG_LIBUTILS_WITH_ISOC) += isoc
 subdirs-y += ext
+subdirs-y += compiler-rt
```
jenswi-linaro, line 3: Should we add a `$(CFG_LIBUTILS_WITH_COMPILER_RT)`?

jforissier: Preferably, yes. Although I doubt it will often be disabled, at least this is consistent with `CFG_LIBUTILS_WITH_ISOC` and it is also an opportunity to document in `mk/config.mk` what purpose compiler-rt serves.

jenswi-linaro:
Update

mk/config.mk
```diff
@@ -251,6 +251,11 @@ TA_PUBLIC_KEY ?= $(TA_SIGN_KEY)
 # may not because they obtain the isoc functions from elsewhere
 CFG_LIBUTILS_WITH_ISOC ?= y
 
+# Include lib/libutils/compiler-rt in the build. Most platforms need this,
+# but if built with a special toolchains that doesn't use a compiler
+# runtime compatible with libgcc or LLVM compiler-rt this can be disabled.
```
jforissier, line 256: After "Most platforms need this", I'd rather explain what this really does and why. How about:

> Provides some functions called "compiler builtins", which the compiler may invoke to perform low-level operations such as long long division etc.
> Such functions typically come with compiler runtime libraires (GCC has libgcc, Clang has compiler-rt). OP-TEE often can't use them because they may be Linux-specific or bring unwanted dependencies. Therefore, this imports and builds only what's needed.

Related to Clang: `libgcc$(sm) := ... -rtlib=compiler-rt -print-libgcc-file-name ...` in `mk/clang.mk` should not be needed anymore. Which is good news because this compiler-rt library was the reason why we had to build and package a Clang toolchain specifically for OP-TEE (see [here](https://github.com/jforissier/docker_optee_os_ci/blob/master/get_clang.sh#L1-L10)). Could you please add a commit to get rid of the useless line?

jforissier:
For "libmbedtls: config: re-enable long long divisions":

Performance is back to normal.

mk/config.mk
jforissier: > Could you please add a commit to get rid of the useless line?

Actually I did it in https://github.com/OP-TEE/optee_os/pull/7382

jforissier:

jenswi-linaro:
Squashed and tag applied.

PR #7381, opened 2025-04-28
core: kernel: dt_driver: variable may be unused

jforissier:
Fix the following warning with Clang 18.6.1:

 core/kernel/dt_driver.c:456:15: error: variable 'count' set but not used [-Werror,-Wunused-but-set-variable]
   456 |         unsigned int count = 0;
       |                      ^

Indeed, when CFG_TEE_CORE_LOG_LEVEL <= 2, the value of count is never read.


jenswi-linaro (commented):

GseoC:

PR #7383, opened 2025-04-28
plat-rockchip: rk3399: remove GIC configuration

jenswi-linaro:
From commit 773c05f417fa ("irqchip/gic-v3: Work around insecure GIC integrations") in the Linux kernel it appears that the hardware integration of the GIC500 isn't correct. For v6.13 kernels which includes that commit this has the effect of OP-TEE printing and endless stream of:
D/TC:0 0 gic_native_itr_handler:971 Special interrupt 1023

Fix this by removing GIC configuration for RK3399 so the device can be used with v6.13 kernels and later.


jenswi-linaro:
By the way, I have prepared a new branch for the v6.13 kernel at https://github.com/jenswi-linaro/linux-1/tree/optee_on_v6.13. It's just a simple rebase of the old optee branch.

jforissier (commented):

jforissier:
> By the way, I have prepared a new branch for the v6.13 kernel at https://github.com/jenswi-linaro/linux-1/tree/optee_on_v6.13. It's just a simple rebase of the old optee branch.

Thanks. Since 6.14 is out, we may as well rebase on it I suppose? I can take care of that and pushing to linaro-swg/linux optee.

jenswi-linaro:
Tag applied. Yes, please go ahead with 6.14.

jforissier:
Branch https://github.com/linaro-swg/linux/tree/optee rebased onto v6.14.

PR #7386, opened 2025-04-30
plat-k3: drivers: Protect OP-TEE memory from normal world

saschahauer:
This protects the OP-TEE memory from accesses from the normal world. Without this patch the whole OP-TEE memory is accessible both read and write from the normal world. This patch has been tested on AM62x. The other K3 platforms likely suffer from the same problem, but I don't have a way to test any of them, so these are untouched for now.

I've put the code into core/arch/arm/plat-k3/drivers/sa2ul.c which is likely the wrong file. I used this as it contains some defines I need. Might need some refactoring before this can finally be merged, but before I do this I'd like to ask if the approach is generally fine and if there are eventually patches floating which cover more K3 SoCs

### DIFF
--- a/core/arch/arm/plat-k3/drivers/sa2ul.c
+++ b/core/arch/arm/plat-k3/drivers/sa2ul.c
@@ -33,7 +33,67 @@
 #define FW_SECURE_ONLY          GENMASK_32(7, 0)
 #define FW_NON_SECURE           GENMASK_32(15, 0)
 
+#define DDR16SS0_TI_SCI_FW_ID	1
+#define TI_SCI_FW_RGN_ID_DDR16SS0_OPTEE	0
+
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, SA2UL_BASE, SA2UL_REG_SIZE);
+
+static int protect_optee_memory(void)
+{
+	uint8_t owner_privid = 0;
+	uint8_t owner_index = OPTEE_HOST_ID;
+	uint16_t owner_permission_bits = 0;
+	uint16_t fwl_id = DDR16SS0_TI_SCI_FW_ID;
+	unsigned int region = TI_SCI_FW_RGN_ID_DDR16SS0_OPTEE;
+	uint32_t control = 0;
+	uint32_t permissions[FWL_MAX_PRIVID_SLOTS] = { };
+	uint64_t start_address = 0;
+	uint64_t end_address = 0;
+	int ret = 0;
+
+	ret = ti_sci_change_fwl_owner(fwl_id, region, owner_index,
+				      &owner_privid, &owner_permission_bits);
+	if (ret) {
+		EMSG("Cannot set owner %u/%u\n", fwl_id, region);
+		return ret;
+	}
+
+	ret = ti_sci_get_fwl_region(fwl_id, region, 1,
+				    &control, permissions,
+				    &start_address, &end_address);
+	if (ret) {
+		EMSG("Could not get firewall region information");
+		return ret;
+	}
+
+	/* Modify firewall to allow all others access */
+	control = FW_ENABLE_REGION | FW_BACKGROUND_REGION;
+	start_address = DRAM0_BASE;
+	end_address = (uint64_t)DRAM0_BASE + DRAM0_SIZE - 1;
+	permissions[0] = (FW_WILDCARD_PRIVID << 16) | FW_NON_SECURE;
+	ret = ti_sci_set_fwl_region(fwl_id, region, 1,
+				    control, permissions,
+				    start_address, end_address);
+	if (ret) {
+		EMSG("Could not set firewall region information");
+		return ret;
+	}
+
+	/* Modify firewall to block all others access */
+	control = FW_ENABLE_REGION;
+	start_address = CFG_TZDRAM_START;
+	end_address = (uint64_t)CFG_TZDRAM_START + CFG_TZDRAM_SIZE - 1;
+	permissions[0] = (FW_WILDCARD_PRIVID << 16) | FW_SECURE_ONLY;
+	ret = ti_sci_set_fwl_region(fwl_id, 1, 1,
+				    control, permissions,
+				    start_address, end_address);
+	if (ret) {
+		EMSG("Could not set firewall region information");
+		return ret;
+	}
+
+	return 0;
+}
 
 static TEE_Result sa2ul_init(void)
 {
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -33,8 +33,67 @@
 #define FW_SECURE_ONLY          GENMASK_32(7, 0)
 #define FW_NON_SECURE           GENMASK_32(15, 0)
 
+#define DDR16SS0_TI_SCI_FW_ID	1
+
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, SA2UL_BASE, SA2UL_REG_SIZE);
 
+static __maybe_unused int protect_opee_memory(void)
```
GseoC, line 40: typo: s/opee/optee

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -33,8 +33,67 @@
...
+	uint16_t owner_permission_bits = 0;
+	uint16_t fwl_id = DDR16SS0_TI_SCI_FW_ID;
+	unsigned int region = 0;
+	uint32_t control;
+	uint32_t permissions[FWL_MAX_PRIVID_SLOTS] = { };
+	uint64_t start_address;
+	uint64_t end_address;
+	int ret = 0;
+
+	ret = ti_sci_change_fwl_owner(fwl_id, region, owner_index,
```
GseoC, line 53: What is region 0 for PLATFORM_FLAVOR_am62x? I can't find the information

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -33,8 +33,67 @@
...
+		EMSG("Could not set firewall region information");
+		return ret;
+	}
+
+	/* Modify firewall to block all others access */
+	control = FW_ENABLE_REGION;
+	start_address = CFG_TZDRAM_START;
+	end_address = (uint64_t)CFG_TZDRAM_START + CFG_TZDRAM_SIZE - 1;
+	permissions[0] = (FW_WILDCARD_PRIVID << 16) | FW_SECURE_ONLY;
+	ret = ti_sci_set_fwl_region(fwl_id, 1, 1,
```
GseoC, line 86: Why not use SA2UL_TI_SCI_FW_RGN_ID here?

GseoC: Ok this is not the same firewall but explicit defines would help in this case (regions IDs)

core/arch/arm/plat-k3/drivers/sa2ul.c
```diff
@@ -65,6 +124,11 @@ static TEE_Result sa2ul_init(void)
 
 	IMSG("Activated SA2UL device");
 
+#if defined(PLATFORM_FLAVOR_am62x)
```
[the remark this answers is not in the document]
Emantor: I strongly prefer `IS_ENABLED()` but I am the one who initially added that from zephyr, so I am biased.
The real advantage is that the code is compiled when any `plat-k3` flavor is build instead of only building it if am62x is selected.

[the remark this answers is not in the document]
jforissier: @GseoC `IS_ENABLED()` is preferred for the reason @Emantor mentioned.

core/arch/arm/plat-k3/drivers/sa2ul.c
saschahauer: From my information it's just the first region. As all regions in this firewall id are unused I just picked the first one. I can add a define for it if that's preferred

core/arch/arm/plat-k3/drivers/sa2ul.c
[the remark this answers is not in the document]
GseoC: Ok, thanks for confirming!

core/arch/arm/plat-k3/drivers/sa2ul.c
GseoC: Sure, DDR16SS0_TI_SCI_FW_BASE_RGN_ID could do it.

I guess the region0 is covering all the DDR and you pick a chunk of it for OP-TEE?
DDR16SS0_TI_SCI_FW_OPTEE_RGN_ID can do it as well for region id 1 then.

glneo:
Hello @saschahauer,

For K3 SoCs there are actually two ways to setup firewalls: the API you are using here, and also a certificate method.

The issue with the API method as used here is that there is a window before you this function is run that OP-TEE is exposed to other running masters in the system. To help with this, we also have a certificate based method. All binaries (TF-A/OP-TEE/bootloader) to be run on these devices are wrapped in a signed certificate. This certificate has many extensions including one for setting firewalls[0]. What this allows for is a firewall to be setup first, and then OP-TEE loaded and authenticated behind this firewall. This means no window in time for which the content of OP-TEE is exposed.

This certificate is applied to the boot images while they are packaged into the `tispl.bin` binary by U-Boot build. The certificate is generated by the same using `binman`. You can find a binman template for the OP-TEE firewall here[1], and it expanded and used on J721e SoCs here[2].

So yes it isn't implemented by default yet for AM62 devices in U-Boot, but that would be the best method.

What you can do here in OP-TEE, which I think would be really useful, is still check for the presence of these pre-setup firewalls and warn if they are not already in place (or then add them manually if not detected).

[0] https://software-dl.ti.com/tisci/esd/latest/2_tisci_msgs/security/sec_cert_format.html#sysfw-fwl-ext
[1] https://source.denx.de/u-boot/u-boot/-/blob/master/arch/arm/dts/k3-binman.dtsi?ref_type=heads#L482
[2] https://source.denx.de/u-boot/u-boot/-/blob/master/arch/arm/dts/k3-j721e-binman.dtsi?ref_type=heads#L410

saschahauer:
> Hello @saschahauer,
> 
> For K3 SoCs there are actually two ways to setup firewalls: the API you are using here, and also a certificate method.

Thanks, I didn't know about the certificate method and that's the missing piece of information for me. I'll try it out.

> What you can do here in OP-TEE, which I think would be really useful, is still check for the presence of these pre-setup firewalls and warn if they are not already in place (or then add them manually if not detected).

I'll do so once I have the certificate method up.

glneo:
> Thanks, I didn't know about the certificate method and that's the missing piece of information for me.

We added it only a year or so ago, so it probably isn't well advertised yet. Anyway, if you have any issues getting it working just let me know :+1:

PR #7387, opened 2025-04-30
core: ffa: reserve physical nexus memory for manifest

jenswi-linaro:
In reinit_manifest_dt() when reserving physical memory for the manifest, allocate from physical nexus memory instead to support ns-virtualization.

Fixes: 414123ae8ca5 ("core: ffa: reserve physical memory for manifest")


etienne-lms (commented):

jenswi-linaro:
Tag applied.

PR #7390, opened 2025-05-05
Add K3 TI-SCI call functions for writing Software and Key Revision Counters

glneo:
These are added for completeness with the full set of TI-SCI OTP revision calls[0]. These can be used from a TA should modifying these counters need be done in OP-TEE.

[0] https://software-dl.ti.com/tisci/esd/latest/2_tisci_msgs/security/otp_revision.html

### DIFF
--- a/core/arch/arm/plat-k3/drivers/ti_sci.c
+++ b/core/arch/arm/plat-k3/drivers/ti_sci.c
@@ -425,6 +425,31 @@ int ti_sci_lock_otp_row(uint8_t row_idx, uint8_t hw_write_lock,
 	return 0;
 }
 
+int ti_sci_set_swrev(uint8_t identifier, uint32_t swrev)
+{
+	struct ti_sci_msq_req_set_swrev req = { };
+	struct ti_sci_msq_resp_set_swrev resp = { };
+	struct ti_sci_xfer xfer = { };
+	int ret = 0;
+
+	ret = ti_sci_setup_xfer(TI_SCI_MSG_WRITE_SWREV, 0,
+				&req, sizeof(req),
+				&resp, sizeof(resp),
+				&xfer);
+	if (ret)
+		return ret;
+
+	req.identifier = identifier;
+	req.swrev = swrev;
+
+	ret = ti_sci_do_xfer(&xfer);
+	if (ret)
+		return ret;
+
+	memzero_explicit(&req, sizeof(req));
+	return 0;
+}
+
 int ti_sci_get_swrev(uint32_t *swrev)
 {
 	struct ti_sci_msq_req_get_swrev req = { };
--- a/core/arch/arm/plat-k3/drivers/ti_sci.c
+++ b/core/arch/arm/plat-k3/drivers/ti_sci.c
@@ -495,6 +495,34 @@ int ti_sci_get_keycnt_keyrev(uint32_t *key_cnt, uint32_t *key_rev)
 	return 0;
 }
 
+int ti_sci_set_keyrev(uint32_t keyrev,
+		      uint32_t cert_addr_lo,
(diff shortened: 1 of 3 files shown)


jenswi-linaro:

core/arch/arm/plat-k3/drivers/ti_sci.c
```diff
@@ -425,6 +425,31 @@ int ti_sci_lock_otp_row(uint8_t row_idx, uint8_t hw_write_lock,
...
+	struct ti_sci_msq_resp_set_swrev resp = { };
+	struct ti_sci_xfer xfer = { };
+	int ret = 0;
+
+	ret = ti_sci_setup_xfer(TI_SCI_MSG_WRITE_SWREV, 0,
+				&req, sizeof(req),
+				&resp, sizeof(resp),
+				&xfer);
+	if (ret)
+		return ret;
```
etienne-lms, line 440: The change looks good to me in that it is consistent with the rest if **ti_sci.c** implementation. That said, I see that all the functions in this source file use `int` return values whereas they return `TEE_ERROR_xxx` or `0` values. I think it would be nice someone changes this to use `TEE_Result` explicit return values and `TEE_SUCCESS` instead of `0` where applicable.

core/arch/arm/plat-k3/drivers/ti_sci.h
```diff
@@ -161,6 +161,17 @@ int ti_sci_write_otp_row(uint8_t row_idx, uint32_t row_val, uint32_t row_mask);
...
+/**
+ * ti_sci_set_swrev - Write Software Revision
+ * @identifier: One of the entries from enum tisci_otp_revision_identifier
+ * @swrev:	Software Revision
+ *
+ * Writes the software revision.
+ *
+ * Return: 0 if all goes well, else appropriate error message
+ */
+int ti_sci_set_swrev(uint8_t identifier, uint32_t swrev);
```
etienne-lms, line 173: Would it make sense to have `enum tisci_otp_revision_identifier` as type for `identifier`?
Maybe not if caller gets the value from a raw `int`, `uint8_t` or like.

core/arch/arm/plat-k3/drivers/ti_sci.c
glneo: Agree, this has been an issue for a while, will be good to fix this all these functions to return `TEE_Result` after we finish the churn of this series.

core/arch/arm/plat-k3/drivers/ti_sci.h
glneo: There will need to be a narrowing from `enum` to `uint8_t` at some point, I'd like to either make it the job of the caller or maybe have a couple versions of this function, one for each type of revision identifier, so we can hide away the enum from the API callers. This would also help to prevent calling with bad identifier enum values. Will think on that.

core/arch/arm/plat-k3/drivers/ti_sci.h
```diff
@@ -161,6 +161,17 @@ int ti_sci_write_otp_row(uint8_t row_idx, uint32_t row_val, uint32_t row_mask);
 int ti_sci_lock_otp_row(uint8_t row_idx, uint8_t hw_write_lock,
 			uint8_t hw_read_lock, uint8_t row_soft_lock);
 
+/**
+ * ti_sci_set_swrev - Write Software Revision
+ * @identifier: One of the entries from enum tisci_otp_revision_identifier
+ * @swrev:	Software Revision
+ *
+ * Writes the software revision.
```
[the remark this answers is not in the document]
glneo: Updated.

etienne-lms (commented):

PR #7391, opened 2025-05-08
Refactor ASLR for RISC-V architecture support

lyctw:
This patch series is in preparation of adding ASLR support for RISC-V.

### DIFF
--- a/core/arch/arm/mm/core_mmu.c
+++ b/core/arch/arm/mm/core_mmu.c
@@ -225,6 +225,19 @@ out:
 	return ret;
 }
 
+vaddr_t arch_aslr_base_addr(vaddr_t start_addr, uint64_t seed,
+			    unsigned int iteration_count)
+{
+	vaddr_t base_addr = start_addr + seed;
+	const unsigned int va_width = core_mmu_get_va_width();
+	const vaddr_t va_mask = GENMASK_64(va_width - 1, SMALL_PAGE_SHIFT);
+
+	if (iteration_count)
+		base_addr ^= BIT64(va_width - iteration_count);
+
+	return base_addr & va_mask;
+}
+
 bool cpu_mmu_enabled(void)
 {
 	uint32_t sctlr;
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -261,6 +261,24 @@ static unsigned int core_mmu_pgt_idx(vaddr_t va, unsigned int level)
 	return idx & RISCV_MMU_VPN_MASK;
 }
 
+/* Get the pointed VA base of specific PTE in page table */
+static inline vaddr_t core_mmu_pgt_get_va_base(unsigned int level,
+					       unsigned int idx)
+{
+#ifdef RV32
+	return SHIFT_U32(idx, CORE_MMU_SHIFT_OF_LEVEL(level));
+#else
+	vaddr_t va_base = SHIFT_U64(idx, CORE_MMU_SHIFT_OF_LEVEL(level));
+	vaddr_t va_width_msb = BIT64(RISCV_MMU_VA_WIDTH - 1);
+	vaddr_t va_extended_mask = GENMASK_64(63, RISCV_MMU_VA_WIDTH);
+
+	if (va_base & va_width_msb)
+		return va_extended_mask | va_base;
+
+	return va_base;
+#endif
+}
+
 static struct mmu_partition *core_mmu_get_prtn(void)
 {
 	return &default_partition;
@@ -745,7 +763,7 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
 		pgt = core_mmu_xlat_table_entry_pa2va(pte, pgt);
 		if (!pgt)
 			goto out;
-		va_base += SHIFT_U64(idx, CORE_MMU_SHIFT_OF_LEVEL(level));
+		va_base += core_mmu_pgt_get_va_base(level, idx);
 		level--;
 	}
 out:
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -699,6 +699,13 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
 	return false;
 }
 
+vaddr_t arch_aslr_base_addr(vaddr_t start_addr, uint64_t seed __unused,
+			    unsigned int iteration_count __unused)
+{
+	/* Placeholder */
+	return start_addr;
+}
+
 bool cpu_mmu_enabled(void)
 {
 	return read_satp();
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -1015,7 +1015,6 @@ void core_init_mmu_prtn(struct mmu_partition *prtn, struct memory_map *mem_map)
 void core_init_mmu(struct memory_map *mem_map)
 {
 	struct mmu_partition *prtn = &default_partition;
-	uint64_t max_va = 0;
 	size_t n = 0;
 
 	if (IS_ENABLED(CFG_DYN_CONFIG)) {
@@ -1041,17 +1040,15 @@ void core_init_mmu(struct memory_map *mem_map)
 	core_init_mmu_prtn_tee(prtn, mem_map);
 
 	for (n = 0; n < mem_map->count; n++) {
-		vaddr_t va_end = mem_map->map[n].va + mem_map->map[n].size - 1;
-
-		if (va_end > max_va)
-			max_va = va_end;
+		if (!core_mmu_va_is_valid(mem_map->map[n].va) ||
+		    !core_mmu_va_is_valid(mem_map->map[n].va +
+					  mem_map->map[n].size - 1))
+			panic("Invalid VA range in memory map");
 	}
 
 	set_user_va_idx(prtn);
 
 	core_init_mmu_prtn_ta(prtn);
-
-	assert(max_va < BIT64(RISCV_MMU_VA_WIDTH));
 }
 
 void core_init_mmu_regs(struct core_mmu_config *cfg)
--- a/core/include/mm/core_mmu.h
+++ b/core/include/mm/core_mmu.h
@@ -682,6 +682,9 @@ void core_mmu_map_region(struct mmu_partition *prtn,
 
 bool arch_va2pa_helper(void *va, paddr_t *pa);
 
+vaddr_t arch_aslr_base_addr(vaddr_t start_addr, uint64_t seed,
+			    unsigned int iteration_count);
+
 static inline bool core_mmu_check_end_pa(paddr_t pa, size_t len)
 {
 	paddr_t end_pa = 0;
--- a/core/include/mm/core_mmu.h
+++ b/core/include/mm/core_mmu.h
@@ -477,6 +477,10 @@ void core_mmu_get_entry(struct core_mmu_table_info *tbl_info, unsigned idx,
 static inline unsigned core_mmu_va2idx(struct core_mmu_table_info *tbl_info,
 			vaddr_t va)
 {
+#ifdef RV64
+	if (tbl_info->level == CORE_MMU_BASE_TABLE_LEVEL)
+		va &= ~GENMASK_64(63, RISCV_MMU_VA_WIDTH);
+#endif
 	return (va - tbl_info->va_base) >> tbl_info->shift;
 }
 
--- a/core/mm/core_mmu.c
+++ b/core/mm/core_mmu.c
@@ -1507,17 +1507,11 @@ static struct memory_map *init_mem_map(struct memory_map *mem_map,
 		add_pager_vaspace(mem_map);
 
 	if (IS_ENABLED(CFG_CORE_ASLR) && seed) {
-		vaddr_t base_addr = start_addr + seed;
-		const unsigned int va_width = core_mmu_get_va_width();
-		const vaddr_t va_mask = GENMASK_64(va_width - 1,
-						   SMALL_PAGE_SHIFT);
-		vaddr_t ba = base_addr;
+		vaddr_t ba = 0;
 		size_t n = 0;
 
 		for (n = 0; n < 3; n++) {
-			if (n)
-				ba = base_addr ^ BIT64(va_width - n);
-			ba &= va_mask;
+			ba = arch_aslr_base_addr(start_addr, seed, n);
 			if (assign_mem_va(ba, mem_map) &&
 			    mem_map_add_id_map(mem_map, id_map_start,
 					       id_map_end)) {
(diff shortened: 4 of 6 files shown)


gagachang:
This line can also be replaced with `if (!arch_va_is_valid(va))`.
https://github.com/OP-TEE/optee_os/blob/9330bb144c711701bf841f660ae2740852b33b6c/core/mm/core_mmu.c#L1367

core/arch/arm/mm/core_mmu.c
```diff
@@ -225,6 +225,36 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
...
+			   vaddr_t start_addr, vaddr_t id_map_start,
+			   vaddr_t id_map_end, unsigned long *offset)
+{
+	const unsigned int va_width = core_mmu_get_va_width();
+	const vaddr_t va_mask = GENMASK_64(va_width - 1, SMALL_PAGE_SHIFT);
+	vaddr_t base_addr = start_addr + seed;
+	size_t n = 0;
+
+	for (n = 0; n < 3; n++) {
+		vaddr_t ba = (n == 0) ? base_addr :
```
jenswi-linaro, line 238: Please restore the if statements instead of the ternary operator.

core/include/mm/core_mmu.h
```diff
@@ -471,7 +471,13 @@ void core_mmu_get_entry(struct core_mmu_table_info *tbl_info, unsigned idx,
 static inline unsigned core_mmu_va2idx(struct core_mmu_table_info *tbl_info,
 			vaddr_t va)
 {
-	return (va - tbl_info->va_base) >> tbl_info->shift;
+#ifdef RV64
+	uint16_t idx_mask = tbl_info->num_entries - 1;
+
+	return ((va - tbl_info->va_base) >> tbl_info->shift) & idx_mask;
```
jenswi-linaro, line 477: Don't the top bits of `tbl_info->va_base` match the top bits of `va`?

core/arch/arm/mm/core_mmu.c
```diff
@@ -225,6 +225,41 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
 	return ret;
 }
 
+int arch_core_aslr_mapping(struct memory_map *mem_map, unsigned long seed,
```
gagachang, line 228: Maybe change `int` to `TEE_Result` ?

core/arch/arm/mm/core_mmu.c
```diff
@@ -225,6 +225,41 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
...
+		vaddr_t ba = (n == 0) ? base_addr :
+					base_addr ^ BIT64(va_width - n);
+		ba &= va_mask;
+
+		if (core_assign_mem_va(ba, mem_map) &&
+		    core_mem_map_add_id_map(mem_map, id_map_start, id_map_end)) {
+			DMSG("Mapping core at %#"PRIxVA" offs %#lx",
+			     ba, ba - start_addr);
+			*offset = ba - start_addr;
+			return 0;
```
gagachang, line 247: `return TEE_SUCCESS;`

core/mm/core_mmu.c
```diff
@@ -1472,6 +1472,7 @@ static struct memory_map *init_mem_map(struct memory_map *mem_map,
 	vaddr_t id_map_end = (vaddr_t)__identity_map_init_end;
 	vaddr_t start_addr = secure_only[0].paddr;
 	unsigned long offs = 0;
+	int rc = 0;
```
gagachang, line 1475: Recommend to use `TEE_Result res = TEE_SUCCESS;`

core/include/mm/core_mmu.h
gagachang: @lyctw  Seems we need to check if  bits[63:39] of `tbl_info->va_base` is 1 if bit 38 of `tbl_info->va_base` is 1.

gagachang:
For commit message of `core: mm: fix va to index calculation for RV64`

```diff
- e.g. For Sv47, bits 63-48 must be set when bit 47 is set to form a
+ e.g. For Sv48, bits 63-48 must be set when bit 47 is set to form a
```

core/arch/riscv/mm/core_mmu_arch.c
```diff
@@ -699,6 +699,17 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
...
 }
 
+int arch_core_aslr_mapping(struct memory_map *mem_map __unused,
+			   unsigned long seed __unused,
+			   vaddr_t start_addr __unused,
+			   vaddr_t id_map_start __unused,
+			   vaddr_t id_map_end __unused,
+			   unsigned long *offset __unused)
+{
+	/* Placeholder */
```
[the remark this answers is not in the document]
lyctw: Hi @jenswi-linaro,

Here is our implementation: https://github.com/lyctw/optee_os/commit/210039b254ea19e2752fc8729ff8834487c9a006#diff-c8d9c38f631f84e96d4e79094e9e8c133a2e5a72f65cb0b9993efa4fb6c0eafaR726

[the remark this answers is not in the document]
jenswi-linaro: You only need a separate function to extend the top va bit into the rest of the top bits when constructing the new base address based on the seed. How about adding this helper function instead:
``` C
static inline vaddr_t arch_aslr_base_addr(vaddr_t start_addr, uint64_t seed, unsigned int iteration_count)
```

core/include/mm/core_mmu.h
gagachang: > Don't the top bits of `tbl_info->va_base` match the top bits of `va`?

In ARM this would be correct.
But in RISC-V, it has a special rule:
- For Sv39 (39-bit VA), bits 63–39 all equal to bit 38 of VA
- For Sv48 (48-bit VA), bits 63–48 all equal to bit 47 of VA
- For Sv57 (57-bit VA), bits 63–57 all equal to bit 56 of VA

Precisely, in Sv48, `VA=0x7FFFFFFFFFFF+1` would be `0xFFF800000000000` (because bit 47 of VA in this case is 1).
Now giving an example: Sv48 `VA=0xFFF800000000000` with `tbl_info->va_base=0`.
The top bits of `tbl_info->va_base` does not match the top bits of VA.
This is why we need `idx_mask`.

gagachang: @lyctw Please consider [this commit](https://github.com/gagachang/optee_os/commit/f0e4a60a671eb0d593f6519cc97431d2ebd76123) into this series which fixes `tbl_info->va_base` when MSB of va_width is 1.

core/arch/riscv/mm/core_mmu_arch.c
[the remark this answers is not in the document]
gagachang: @lyctw I think @jenswi-linaro's suggestion is simpler and workable.

core/include/mm/core_mmu.h
jenswi-linaro: Isn't `VA=0x7FFFFFFFFFFF` always an invalid address? If not, when can it be used?

gagachang: > always an invalid address? If not, when can it be used?

Yeah, `0x7FFFFFFFFFFF` is invalid address.
I just wanted to give an example that RISC-V needs to extend the top VA bit into the rest of the top bits.

jenswi-linaro: Assuming we only use valid VAs, the top bits should always match.

gagachang: > Now giving an example: Sv48 `VA=0xFFF800000000000` with `tbl_info->va_base=0`.
> The top bits of `tbl_info->va_base` does not match the top bits of VA.

But the following case would not match:
Sv48 VA=0xFFF800000000000~0xFFFFFFFFFFFFFFFF with `tbl_info->va_base=0`.

jenswi-linaro: Aha, we have a special case when `tbl_info->va_base=0`. We have asserts in various places that the `idx` isn't out of range, so masking it here would make those asserts useless for risc-v. How about something like this:
``` C
#ifdef RV64
        if (tbl_info->va_base)
                assert(tbl_info->level == CORE_MMU_BASE_TABLE_LEVEL);
        else
                va &= ~GENMASK_64(63, RISCV_MMU_VA_WIDTH);
#endif

        return (va - tbl_info->va_base) >> tbl_info->shift;
```
I'm not sure if the `assert()` makes sense.

gagachang: > Aha, we have a special case when `tbl_info->va_base=0`. We have asserts in various places that the `idx` isn't out of range, so masking it here would make those asserts useless for risc-v. How about something like this:
> 
> ```c
> #ifdef RV64
>         if (tbl_info->va_base)
>                 assert(tbl_info->level == CORE_MMU_BASE_TABLE_LEVEL);
>         else
>                 va &= ~GENMASK_64(63, RISCV_MMU_VA_WIDTH);
> #endif
> 
>         return (va - tbl_info->va_base) >> tbl_info->shift;
> ```
> 
> I'm not sure if the `assert()` makes sense.

Not only base level table has `tbl_info->va_base=0`.
I would like to give an example of VA ranges in Sv48 (48-bit VA):
```c
Root table index 000 PTE ==> 0x000000000000000 ~ 0x000007FFFFFFFFF
    Next-level table index 000 PTE ==> 0x000000000000000 ~ 0x00000003FFFFFFF
Root table index 001 PTE ==> 0x000008000000000 ~ 0x00001FFFFFFFFFF
Root table index 002 PTE ==> 0x000010000000000 ~ 0x000017FFFFFFFFF
......
Root table index 255 PTE ==> 0x0007F8000000000 ~ 0x0007FFFFFFFFFFF
Root table index 256 PTE ==> 0xFFF800000000000 ~ 0xFFF807FFFFFFFFF  (bits 63–48 all equal to bit 47 which is 1)
Root table index 257 PTE ==> 0xFFF808000000000 ~ 0xFFF81FFFFFFFFFF  (bits 63–48 all equal to bit 47 which is 1)
......
Root table index 511 PTE ==> 0xFFFFF8000000000 ~ 0xFFFFFFFFFFFFFFF  (bits 63–48 all equal to bit 47 which is 1)
```
In this case the "Root table" has `tbl_info->va_base=0`.
The "Next-level table" also has `tbl_info->va_base=0`.
The indexes 256~511 are special cases that this PR tries to solve, which causing the top bits of `tbl_info->va_base` unmatch the top bits of VA.

I am not sure if we can simply do this because it may hide out of range VA:
```c
#ifdef RV64
        va &= ~GENMASK_64(63, RISCV_MMU_VA_WIDTH);
#endif

        return (va - tbl_info->va_base) >> tbl_info->shift;
```

jenswi-linaro: When calculating the index at the top level translation table, we only use the lower 48 bits, bit 47 to bit 0. At all the other levels I expect that the top bits will equal the ones in `tbl_info->va_base`. Does this work:
``` C
#ifdef RV64
        if (tbl_info->level == CORE_MMU_BASE_TABLE_LEVEL)
                va &= ~GENMASK_64(63, RISCV_MMU_VA_WIDTH);
#endif 
```

gagachang: > Does this work:
> 
> ```c
> #ifdef RV64
>         if (tbl_info->level == CORE_MMU_BASE_TABLE_LEVEL)
>                 va &= ~GENMASK_64(63, RISCV_MMU_VA_WIDTH);
> #endif 
> ```

This works for me! Yes, we only need the lower `va_width` bits at the top level translation table.
@lyctw Can you also try this version?

core/arch/arm/mm/core_mmu.c
```diff
@@ -225,6 +225,37 @@ bool arch_va2pa_helper(void *va, paddr_t *pa)
...
+
+	for (n = 0; n < 3; n++) {
+		if (n)
+			ba = base_addr ^ BIT64(va_width - n);
+		ba &= va_mask;
+		if (core_assign_mem_va(ba, mem_map) &&
+		    core_mem_map_add_id_map(mem_map, id_map_start,
+					    id_map_end)) {
+			DMSG("Mapping core at %#"PRIxVA" offs %#lx",
+			     ba, ba - start_addr);
```
etienne-lms, line 246: `ba` and `start_addr` are of `vaddr_ŧ` so prefer use `PRIxVA` for the trace message.

peterlin-sifive:
Rebased onto master due to CI failure.

core/arch/arm/mm/core_mmu.c
peterlin-sifive: Hi @etienne-lms,

Thanks for the review and there are no changes to this line in the updated version, so I'll leave it as is.

core/include/mm/core_mmu.h
peterlin-sifive: Patch updated, thanks @jenswi-linaro and @gagachang for suggestion.

gagachang:
Thanks @peterlin-sifive !
It seems that the CI failure is false positive.

core/arch/riscv/mm/core_mmu_arch.c
```diff
@@ -745,7 +763,7 @@ bool core_mmu_find_table(struct mmu_partition *prtn, vaddr_t va,
 		pgt = core_mmu_xlat_table_entry_pa2va(pte, pgt);
 		if (!pgt)
 			goto out;
-		va_base += SHIFT_U64(idx, CORE_MMU_SHIFT_OF_LEVEL(level));
+		va_base += core_mmu_pgt_get_va_base(level, idx);
```
jenswi-linaro, line 766: Does this work as intended?
The previous code adds the va index offset in the translation table, but this new code does something different for non-RV32 code, if I understand it correctly.

gagachang: Actually, this fixes the cases that index >= 256:

```shell
Root table index 000 ==> va_base = 0x000000000000000
Root table index 001 ==> va_base = 0x000008000000000
......
Root table index 255 ==> va_base = 0x0007F8000000000
Root table index 256 ==> va_base = 0xFFF800000000000 (bit[63:48] all equal to bit 47 which is 1)
Root table index 257 ==> va_base = 0xFFF808000000000 (bit[63:48] all equal to bit 47 which is 1)
......
Root table index 511 ==> va_base = 0xFFFFF8000000000 (bit[63:48] all equal to bit 47 which is 1)
```

Adds the va index offset in the translation table is not enough.
As you can see, when the index >= 256, the upper bit[63:48] become 1.
Because bit[47] needs to be extended to bit[63:48] in RISC-V Sv48.
This is what `core_mmu_pgt_get_va_base()` does.

jenswi-linaro: I see that, but what happens on the next iteration when the top bits of `va_base` are already `1`?

jenswi-linaro: Got it, thanks.

jenswi-linaro:
Looks good.

For:
- core: mm: refactor ASLR mapping for architecture support
- core: mm: factor out virtual address range validation to arch code
- core: mm: fix va to index calculation for RV64
 
please apply

and for
riscv: mm: fix map_offset data type
core: riscv: mm: add per-mapping VA range validation
riscv: mm: Fix VA base for canonical addresses

PR #7393, opened 2025-05-12
core: riscv: replace get_core_pos() with thread_get_current_hartindex

peterlin-sifive:
Replace get_core_pos() with more descriptive
thread_get_current_{hartid,hartindex}() on RISC-V, 
and handle range checks in one place
(i.e. assert(hartidx < CFG_TEE_CORE_NB_CORE)),
also fix CLINT_MTIMECMP accesses where should
use hartid instead of hartindex.

Drop thread_get_hartid_by_hartindex() as we
only need to access local hart information.

### DIFF
--- a/core/arch/riscv/include/kernel/clint.h
+++ b/core/arch/riscv/include/kernel/clint.h
@@ -35,12 +35,12 @@ static inline void clint_ipi_clear(unsigned long hart)
 static inline void clint_set_mtimecmp(uint64_t timecmp)
 {
 	/* Each hart has a separate source of timer interrupts */
-	io_write64(CLINT_MTIMECMP(get_core_pos()), timecmp);
+	io_write64(CLINT_MTIMECMP(thread_get_current_hartid()), timecmp);
 }
 
 static inline uint64_t clint_get_mtimecmp(void)
 {
-	return io_read64(CLINT_MTIMECMP(get_core_pos()));
+	return io_read64(CLINT_MTIMECMP(thread_get_current_hartid()));
 }
 
 static inline uint64_t clint_get_mtime(void)
--- a/core/arch/riscv/include/kernel/thread_arch.h
+++ b/core/arch/riscv/include/kernel/thread_arch.h
@@ -198,7 +198,8 @@ static inline void thread_user_clear_vfp(struct user_mode_ctx *uctx __unused)
 #endif
 
 vaddr_t thread_get_saved_thread_sp(void);
-uint32_t thread_get_hartid_by_hartindex(uint32_t hartidx);
+uint32_t thread_get_current_hartindex(void);
+uint32_t thread_get_current_hartid(void);
 
 static inline void thread_get_user_kcode(struct mobj **mobj, size_t *offset,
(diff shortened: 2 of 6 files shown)


peterlin-sifive:
CC. @gagachang

peterlin-sifive:
Hi @jforissier,

Indeed, we still have `get_core_pos()` in common code, the intention here was just providing a readable API for hartid or hart index, we can revert this patch and simply add comment to clarify the use of `get_core_pos()`:

```diff
--- a/core/arch/riscv/kernel/thread_rv.S
+++ b/core/arch/riscv/kernel/thread_rv.S
@@ -38,6 +38,12 @@

 /* size_t __get_core_pos(void); */
 FUNC __get_core_pos , : , .identity_map
+       /*
+        * Return the logical hart index which is ranging from
+        * 0 to (CFG_TEE_CORE_NB_CORE - 1). If you need to get
+        * hart ID of this core, use:
+        * thread_get_hartid_by_hartindex(get_core_pos())
+        */
        lw      a0, THREAD_CORE_LOCAL_HART_INDEX(tp)
        ret
 ```
 
 @gagachang , any thoughts ?

**Comment by gagachang** [comment]
I agree what @jforissier said:
```
Also, the thread_ prefix is not justified for thread_get_current_hartindex() since the function can be used outside a thread context
```

@peterlin-sifive  How about reserving `get_core_pos()` for hart index, and introducing `get_core_id()` for hart ID?
However, I am not sure whether ARM needs `get_core_id()` or not.


The reason why RISC-V needs two functions is because that RISC-V specification says:
```
Hart IDs might not necessarily be numbered contiguously in a multiprocessor system,
but at least one hart must have a hart ID of zero.
Hart IDs must be unique within the execution environment.
```

Thus, RISC-V can not use hart(core) ID directly as array indexes because it might not be contiguous number.

peterlin-sifive:
> Thus, RISC-V can not use hart(core) ID directly as array indexes because it might not be contiguous number.

yes, for example, [SiFive FU740-based](https://github.com/torvalds/linux/blob/38fec10eb60d687e30c8c6b5420d86e8149f7557/arch/riscv/boot/dts/sifive/fu740-c000.dtsi#L25-L151) SoC has 5 CPUs, 
- 1x MCU class CPU: hart0
- 4x Application class CPU: hart1-4 (running Linux and OP-TEE)

jenswi-linaro:
The situation is similar for Arm, see `get_core_pos_mpidr()`:
https://github.com/OP-TEE/optee_os/blob/d52be6948b6eefd364d5b2f802593773ec0fb596/core/arch/arm/kernel/misc_a64.S#L19-L58

jenswi-linaro:
I'm considering something that doesn't require a function to calculate the core position based on MPIDR or the corresponding thing in risc-v. But I'd like to get all my dynamic configuration patches merged before I start on that to avoid tricky rebasing. For Aarch64, SP_EL1 has the value of `thread_core_local[get_core_pos ()]`. For Aarch32, it's `abt_sp` and `und_sp`.

PR #7395, opened 2025-05-14
ci: do not override QEMU version

jforissier:
The "make check (QEMUv8, BTI+MTE+PAC)" CI job currently checks out an old version of QEMU. Not only is this workaround not necessary anymore, but worse it makes the check fail with the following output (from out/bin/serial0.log):

 Starting kernel ...

 **
 ERROR:../target/arm/internals.h:742:regime_is_user: code should not be reached
 Bail out! ERROR:../target/arm/internals.h:742:regime_is_user: code should not be reached
 Aborted (core dumped)
 send: spawn id exp9 not open
     while executing
 "send -- "root\r\r""
     (file "/root/optee/build/../build/qemu-check.exp" line 129

Therefore, let the QEMU version from the manifest be used.


jenswi-linaro (commented):

Issue #7402, opened 2025-05-19
Crash in `pl011_init` during `configure_console_from_dt` with `CFG_TRANSFER_LIST=y` on FVP


harrisonmutai-arm:
When enabling firmware handoff support on FVP by setting `CFG_TRANSFER_LIST=y`, OP-TEE crashes during late kernel initialization, specifically in the `configure_console_from_dt()` function, which attempts dynamic initialization of the console driver.

This crash occurs in `pl011_init`, resulting in a synchronous external data abort, likely due to dereferencing an unmapped console register. 

```
I/TC: pl011: device parameters ignored (115200n8)
E/TC:0   
E/TC:0   Core data-abort at address 0x6490030 (Synchronous external abort)
E/TC:0    esr 0x96000210  ttbr0 0x060a1000   ttbr1 0x00000000   cidr 0x0
E/TC:0    cpu #0          cpsr 0x600003c4
E/TC:0    x0  0000000006490000 x1  0000000000001000
E/TC:0    x2  0000000000000000 x3  0000000000000000
E/TC:0    x4  000000000000000d x5  000000000000000e
E/TC:0    x6  0000000000000000 x7  0000000000000fff
E/TC:0    ...
E/TC:0    elr 000000000600f9d8
E/TC:0    sp_el0 00000000061fdf00
D/TC:0   get_fault_type:536 [abort] (Synchronous external abort)
```

FVP typically does not rely on Device Tree for console configuration under normal operation. The introduction of `CFG_TRANSFER_LIST=y` forces DT-based console init, even though `CFG_DT` might not be explicitly enabled or needed on the platform otherwise.

Should `CFG_TRANSFER_LIST` be relaxed to support platforms with `CFG_DT=n`? Alternatively, should OP-TEE fall back to static console configuration if DT parsing fails or is unavailable?

jenswi-linaro:
> Should `CFG_TRANSFER_LIST` be relaxed to support platforms with `CFG_DT=n`?

Yes, I think that's the best option.

PR #7405, opened 2025-05-20
mk: core: relax rules around CFG_TRANSFER_LIST

harrisonmutai-arm:
`CFG_TRANSFER_LIST` forces DT-based console init, even though `CFG_DT` might not be explicitly enabled or needed on the platform otherwise. Relax the rules around `CFG_TRANSFER_LIST` to allow for a platform to receive boot arguments via Firmware Handoff without enabling `CFG_DT`.

Fixes: #7402

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -529,8 +529,6 @@ endif
 # has a stable release.
 # This feature requires the support of Device Tree.
 CFG_TRANSFER_LIST ?= n
-$(eval $(call cfg-enable-all-depends,CFG_TRANSFER_LIST, \
-	 CFG_DT CFG_EXTERNAL_DT CFG_MAP_EXT_DT_SECURE))
 
 # Maximum size of the Device Tree Blob, has to be large enough to allow
 # editing of the supplied DTB.


jenswi-linaro:
I was a bit surprised it was as easy as that, but since it's passing CI, please apply:

harrisonmutai-arm:
@jenswi-linaro sorry I misunderstood your comment, applied. 
Fyi @raymo200915 in case you have some thoughts on this.

mk/config.mk
```diff
@@ -529,8 +529,6 @@ endif
 # has a stable release.
 # This feature requires the support of Device Tree.
 CFG_TRANSFER_LIST ?= n
-$(eval $(call cfg-enable-all-depends,CFG_TRANSFER_LIST, \
-	 CFG_DT CFG_EXTERNAL_DT CFG_MAP_EXT_DT_SECURE))
```
raymo200915, line 533: I think it should be separated into two cases:
A) CFG_EXTERNAL_DT and CFG_MAP_EXT_DT_SECURE are required if CFG_DT is selected.
B) Nothing is needed when if CFG_DT is not selected.
Your changes covered B).

raymo200915:
> @jenswi-linaro sorry I misunderstood your comment, applied. Fyi @raymo200915 in case you have some thoughts on this.

I think it should be separated into two cases:
a) DT in use: 

> @jenswi-linaro sorry I misunderstood your comment, applied. Fyi @raymo200915 in case you have some thoughts on this.

I just left one comment.

mk/config.mk
harrisonmutai-arm: I believe part of this is already covered [here (L521)](https://github.com/OP-TEE/optee_os/pull/7405/files/9275c0399587a3d4aa9fece79712910a2a555d52#diff-6301a2703b709daa4d84ead1c30bf67cdaf088abed68753a8f9c740f1e58b54aL521). CFG_MAP_EXT_DT_SECURE depends on CFG_DT, and CFG_EXTERNAL_DT inherits the value of CFG_DT unless explicitly overridden. Are you suggesting we harden the relationship between all 3? Let me know if you see a clean way to handle that distinction.

etienne-lms: Transfer list expects the target memory is mapped secure ([transfer_list.c#L53-L57](
https://github.com/OP-TEE/optee_os/blob/4.6.0/core/kernel/transfer_list.c#L53-L57) and [core_mmu.c#L858-L859](
https://github.com/OP-TEE/optee_os/blob/4.6.0/core/mm/core_mmu.c#L858-L859)) unless what core panics ([boot.c#L1122-L1124](https://github.com/OP-TEE/optee_os/blob/4.6.0/core/arch/arm/kernel/boot.c#L1122-L1124)) hence I agree removing these 2 lines is enough.

harrisonmutai-arm:
@jenswi-linaro are there any additional actions required to progress this on my side?

jforissier:
@harrisonmutai-arm this is ready to merge, thanks.

PR #7407, opened 2025-05-21
Improve get_core_pos() and introduce get_core_pos_hartid() for RISC-V

peterlin-sifive:
This is patch v2 for https://github.com/OP-TEE/optee_os/pull/7393.

### DIFF
--- a/core/arch/riscv/include/kernel/clint.h
+++ b/core/arch/riscv/include/kernel/clint.h
@@ -35,12 +35,12 @@ static inline void clint_ipi_clear(unsigned long hart)
 static inline void clint_set_mtimecmp(uint64_t timecmp)
 {
 	/* Each hart has a separate source of timer interrupts */
-	io_write64(CLINT_MTIMECMP(get_core_pos()), timecmp);
+	io_write64(CLINT_MTIMECMP(get_core_pos_hartid()), timecmp);
 }
 
 static inline uint64_t clint_get_mtimecmp(void)
 {
-	return io_read64(CLINT_MTIMECMP(get_core_pos()));
+	return io_read64(CLINT_MTIMECMP(get_core_pos_hartid()));
 }
 
 static inline uint64_t clint_get_mtime(void)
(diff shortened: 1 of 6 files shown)


peterlin-sifive:
CC. @gagachang

gagachang:
Don't we need similar check for foreign interrupt like `get_core_pos()` ?

```c
size_t __get_core_pos_hartid(void);

static inline size_t __noprof get_core_pos_hartid(void)
{
	/*
	 * Foreign interrupts must be disabled before playing with current
	 * core since we otherwise may be rescheduled to a different core.
	 */
	assert(thread_get_exceptions() & THREAD_EXCP_FOREIGN_INTR);
	return __get_core_pos_hartid();
}
```

peterlin-sifive:
> Don't we need similar check for foreign interrupt like `get_core_pos()` ?
> 
> ```c
> size_t __get_core_pos_hartid(void);
> 
> static inline size_t __noprof get_core_pos_hartid(void)
> {
> 	/*
> 	 * Foreign interrupts must be disabled before playing with current
> 	 * core since we otherwise may be rescheduled to a different core.
> 	 */
> 	assert(thread_get_exceptions() & THREAD_EXCP_FOREIGN_INTR);
> 	return __get_core_pos_hartid();
> }
> ```

I think yes, I'll add a wrapper function. Thanks.

core/arch/riscv/include/kernel/clint.h
```diff
@@ -35,12 +35,12 @@ static inline void clint_ipi_clear(unsigned long hart)
 static inline void clint_set_mtimecmp(uint64_t timecmp)
 {
 	/* Each hart has a separate source of timer interrupts */
-	io_write64(CLINT_MTIMECMP(get_core_pos()), timecmp);
+	io_write64(CLINT_MTIMECMP(get_core_pos_hartid()), timecmp);
```
gagachang, line 38: I don't think we want to use hart ID here.
The RISC-V Advanced Core Local Interruptor Specification uses "HART index".

jenswi-linaro:
Since you read the hartid or "hart index" from `struct thread_core_local`, how about calling the function `thread_get_hartid()`, `thread_get_hartidx()`, or such (still only riscv specific). The `get_core_pos_` prefix should be avoided since that relates to logical core or CPU index, and "Hart index" seems to be something different since it's treated separately.

PR #7409, opened 2025-05-21
dt-bindings: update LTDC layer numbers on stm32mp2x platforms

GseoC:
Small update for LTDC layer numbers to be aligned with the reference manual.

No fixes tag as I have to fix DT and bindings at the same time and there's no functional change.


GseoC:
Tag applied, thanks

PR #7412, opened 2025-05-23
Fix Kasan

jenswi-linaro:

### DIFF
--- a/core/arch/arm/kernel/entry_a32.S
+++ b/core/arch/arm/kernel/entry_a32.S
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
 /*
- * Copyright (c) 2014, Linaro Limited
+ * Copyright (c) 2014-2025, Linaro Limited
  * Copyright (c) 2021-2023, Arm Limited
  */
 
@@ -433,6 +433,7 @@ shadow_no_access:
 	cmp	r0, r1
 	bls	shadow_no_access
 
+#if !defined(CFG_DYN_STACK_CONFIG)
 	/* Mark the entire stack area as OK */
 	ldr	r2, =CFG_ASAN_SHADOW_OFFSET
 	ldr	r0, =__nozi_stack_start	/* start */
@@ -446,6 +447,7 @@ shadow_stack_access_ok:
 	strb	r2, [r0], #1
 	cmp	r0, r1
 	bls	shadow_stack_access_ok
+#endif
 #endif
 
 #if defined(CFG_DYN_STACK_CONFIG)
--- a/core/core.mk
+++ b/core/core.mk
@@ -62,8 +62,11 @@ $(error error: CFG_CORE_SANITIZE_KADDRESS not supported with Clang)
 endif
 cflags_kasan	+= -fsanitize=kernel-address \
 		   -fasan-shadow-offset=$(CFG_ASAN_SHADOW_OFFSET)\
-		   --param asan-stack=1 --param asan-globals=1 \
+		   --param asan-globals=1 \
 		   --param asan-instrumentation-with-call-threshold=0
+ifneq ($(CFG_DYN_STACK_CONFIG),y)
+cflags_kasan	+= --param asan-stack=1
+endif
 cflags$(sm)	+= $(cflags_kasan)
 endif
 ifeq ($(CFG_CORE_DEBUG_CHECK_STACKS),y)
--- a/core/mm/boot_mem.c
+++ b/core/mm/boot_mem.c
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2024, Linaro Limited
+ * Copyright (c) 2024-2025, Linaro Limited
  */
 
 #include <assert.h>
+#include <kernel/asan.h>
 #include <kernel/boot.h>
 #include <mm/core_memprot.h>
 #include <mm/core_mmu.h>
@@ -69,6 +70,9 @@ static void *mem_alloc_tmp(struct boot_mem_desc *desc, size_t len, size_t align)
 	if (va < desc->mem_start)
 		panic();
 	desc->mem_end = va;
+
+	asan_tag_access((void *)va, (void *)(va + len));
+
 	return (void *)va;
 }
 
@@ -102,6 +106,9 @@ static void *mem_alloc(struct boot_mem_desc *desc, size_t len, size_t align)
 		panic();
 	add_padding(desc, va);
 	desc->mem_start = ve;
+
+	asan_tag_access((void *)va, (void *)(va + len));
+
 	return (void *)va;
 }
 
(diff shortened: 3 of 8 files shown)


jforissier:
Perhaps we should enable `CFG_CORE_SANITIZE_KADDRESS` in the QEMUv8 CI?

jforissier:
The CI error for QEMUv8 is:
```
root@340e5f3ebc11:~/optee/build# make -j$(nproc) check CFG_CORE_SANITIZE_KADDRESS=y CFG_CORE_ASLR=n
root@340e5f3ebc11:~/optee# export PATH=/root/optee/toolchains/aarch64/bin:$PATH
root@340e5f3ebc11:~/optee# tail -22 out/bin/serial0.log | ./optee_os/scripts/symbolize.py -d optee_os/out/arm/core
o regression_1037.2 TEE attestation
D/TC:? 0 write_key:325 Wrote 1033 bytes
D/TC:? 0 tee_ta_close_session:460 csess 0xe251560 id 10
D/TC:? 0 tee_ta_close_session:479 Destroy session
D/TC:? 0 tee_ta_init_session_with_context:557 Re-open trusted service 39800861-182a-4720-9b67-2bcd622bc0b5
E/TC:1 0 Panic at core/kernel/asan.c:190 <check_access>
E/TC:1 0 TEE load address @ 0xe100000
E/TC:1 0 Call stack:
E/TC:1 0  0x0e108ad8 print_kernel_stack at optee_os/core/arch/arm/kernel/unwind_arm64.c:89
E/TC:1 0  0x0e11b124 __do_panic at optee_os/core/kernel/panic.c:73
E/TC:1 0  0x0e114aa8 check_access at optee_os/core/kernel/asan.c:190
E/TC:1 0  0x0e103a2c memcpy at optee_os/lib/libutils/isoc/newlib/memcpy.c:131
E/TC:1 0  0x0e154ddc s_sha256_compress at optee_os/core/lib/libtomcrypt/src/hashes/sha2/sha256.c:79 (discriminator 3)
E/TC:1 0  0x0e155224 sha256_process at optee_os/core/lib/libtomcrypt/src/hashes/sha2/sha256.c:220
E/TC:1 0  0x0e152b30 ltc_hash_update at optee_os/core/lib/libtomcrypt/hash.c:51
E/TC:1 0  0x0e13156c cmd_hash_tee_memory at optee_os/core/pta/attestation.c:716
E/TC:1 0  0x0e1239f8 pseudo_ta_enter_invoke_cmd at optee_os/core/kernel/pseudo_ta.c:209
E/TC:1 0  0x0e11cda4 tee_ta_invoke_command at optee_os/core/kernel/tee_ta_manager.c:775
E/TC:1 0  0x0e136c3c entry_invoke_command at optee_os/core/tee/entry_std.c:437
E/TC:1 0  0x0e106a38 call_entry_std at optee_os/core/arch/arm/kernel/thread_optee_smc.c:165
E/TC:1 0  0x0e106ebc std_entry_with_regd_arg at optee_os/core/arch/arm/kernel/thread_optee_smc.c:259
I/TC: Halting CPU 0
```

Not sure why CI reported a "Timeout" instead of the exact panic message which should be detected here: https://github.com/OP-TEE/build/blob/master/qemu-check.exp#L96, but that's another story ;)

jforissier:
> > The CI error for QEMUv8 is:
> > [...]
> 
> Thanks
> 
> > Not sure why CI reported a "Timeout" instead of the exact panic message which should be detected here: https://github.com/OP-TEE/build/blob/master/qemu-check.exp#L96, but that's another story ;)
> 
> All secure prints are disabled.

Ah, yes. Perhaps we should leave CFG_TEE_CORE_LOG_LEVEL at 2 (error), that would be convenient.

> The problem is that the PTA tries to access memory that isn't variables, which ASAN is supposed to protect against. So it works! :-) We need to disable that PTA with ASAN, because I don't think it's worth it to try to come up with a scheme to bypass ASAN only for this.

Makes sense. Adding `CFG_ATTESTATION_PTA=n` fixes the problem indeed but the `check-only` takes ~25 min (on my machine) and that's with `CFG_CORE_UNSAFE_MODEXP=y` too. Perhaps running a subset of xtest would be good enough.

core/mm/boot_mem.c
```diff
@@ -228,6 +259,8 @@ void boot_mem_foreach_padding(bool (*func)(vaddr_t va, size_t len, void *ptr),
 			DMSG("keeping %p %#"PRIxVA" len %#zx",
 			     pad, start, len);
 			prev = &pad->next;
+			asan_tag_no_access((void *)start,
+					   (void *)(start + len));
 		}
```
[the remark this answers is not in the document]
jenswi-linaro: No, that should be covered by this range. However, I think I should change the padding areas to start on ASAN-aligned addresses to keep things easy and accurate with regards to ASAN.

jforissier (commented):
For commit "ci: set CFG_TEE_CORE_LOG_LEVEL=2":
That's OK but the commit description is not entirely accurate. Setting the log level to 2 at this points makes `qemu-check.exp` able to detect TEE core asserts and panics and report them immediately rather than just time out. Getting more debug info in the CI log would require setting `DUMP_LOGS_ON_ERROR` to a non-empty string, which we should probably do as well.

jenswi-linaro:
Tag applied, updated commit message for "ci: set CFG_TEE_CORE_LOG_LEVEL=2"

jenswikl:
Ping?

core/arch/arm/kernel/entry_a32.S
```diff
@@ -433,6 +433,7 @@ shadow_no_access:
 	cmp	r0, r1
 	bls	shadow_no_access
 
+#if !defined(CFG_DYN_STACK_CONFIG)
```
etienne-lms: Note: to update to `CFG_DYN_CONFIG` (https://github.com/OP-TEE/optee_os/pull/7421 has been merged)

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` (with the few `CFG_DYN_CONFIG` updates).

jenswikl:
Rebased on master, updated to `CFG_DYN_CONFIG`, and tags applied.

core/core.mk
```diff
@@ -62,8 +62,11 @@ $(error error: CFG_CORE_SANITIZE_KADDRESS not supported with Clang)
 endif
 cflags_kasan	+= -fsanitize=kernel-address \
 		   -fasan-shadow-offset=$(CFG_ASAN_SHADOW_OFFSET)\
-		   --param asan-stack=1 --param asan-globals=1 \
+		   --param asan-globals=1 \
 		   --param asan-instrumentation-with-call-threshold=0
+ifneq ($(CFG_DYN_STACK_CONFIG),y)
```
gagachang: Should be CFG_DYN_CONFIG

jenswikl: Yes, you're right.

jenswikl:
Update with one missing `CFG_DYN_CONFIG`

jenswikl:
It seems `CFG_CORE_SANITIZE_KADDRESS=y ` slows down the rust tests too much, so let's disable them.

jenswikl:
The tests have passed now.

jforissier:
> The tests have passed now.

Thanks. 1.5 hours for QEMUv8 check 1/2 and same for 2/2 is a bit too long IMO, especially since the longest of the other jobs (Xen) is 50 minutes. I will propose a PR to further split these two jobs.

PR #7413, opened 2025-05-26
libmbedtls: fix compilation warning with GCC14 (2)

jforissier:
Cherry-picking commit 7505c3588f44 ("libmbedtls: fix compilation warning with GCC14") which was lost in commit c3deb3d6f3b1 ("Squashed commit upgrading to mbedtls-3.6.3"). It should have been pushed onto the import/mbedtls-3.6.2 branch when it was accepted in master but that didn't happen and therefore it was unfortunately left aside when upgrading. This time it has been applied to import/mbedtls-3.6.3 [1] so it will hopefully not be forgotten in the next upgrade.

Link: https://github.com/OP-TEE/optee_os/commit/b526c146f87 [1]


etienne-lms:

PR #7424, opened 2025-06-04
riscv: mm: fix relocation of pointers in user_vpn2_table_va[]

gagachang:
The user_vpn2_table_va[] stores several pointers pointing to the level 2 page tables used by user TA. When CFG_CORE_ASLR=y, these pointers must be relocated to ASLR VA as well. To fix this issue, we call boot_mem_add_reloc() onto each element of user_vpn2_table_va[], so that the pointers have a chance to be relocated by boot_mem_relocate() later.

Discussed in https://github.com/OP-TEE/optee_os/pull/7371#discussion_r2107604866

### DIFF
--- a/core/arch/riscv/mm/core_mmu_arch.c
+++ b/core/arch/riscv/mm/core_mmu_arch.c
@@ -1054,6 +1054,11 @@ void core_init_mmu(struct memory_map *mem_map)
 #endif
 	}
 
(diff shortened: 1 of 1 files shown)


peterlin-sifive:
Thanks for the fix!

jenswikl:

gagachang:
Tags applied and passed CI!

PR #7426, opened 2025-06-04
Dynamic config for pager

jenswikl:

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -913,11 +913,11 @@ static void init_primary(unsigned long pageable_part)
 	 * every virtual partition separately. Core code uses nex_malloc
 	 * instead.
 	 */
+#ifndef CFG_DYN_CONFIG
 #ifdef CFG_WITH_PAGER
 	/* Add heap2 first as heap1 may be too small as initial bget pool */
 	malloc_add_pool(__heap2_start, __heap2_end - __heap2_start);
 #endif
-#ifndef CFG_DYN_CONFIG
 #ifdef CFG_NS_VIRTUALIZATION
 	nex_malloc_add_pool(__nex_heap_start, __nex_heap_end -
 					      __nex_heap_start);
@@ -983,6 +983,12 @@ static void init_primary(unsigned long pageable_part)
 		 * init_runtime()).
 		 */
 		thread_get_core_local()->curr_thread = 0;
+		if (IS_ENABLED(CFG_DYN_CONFIG)) {
+			threads = calloc(1, sizeof(*threads));
+			if (!threads)
+				panic();
+			thread_count = 1;
+		}
 		init_pager_runtime(pageable_part);
 	}
 
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -26,6 +26,7 @@
 #include <kernel/panic.h>
 #include <kernel/tee_misc.h>
 #include <kernel/thread.h>
+#include <kernel/thread_private.h>
 #include <kernel/tpm.h>
 #include <kernel/transfer_list.h>
 #include <libfdt.h>
@@ -334,7 +335,9 @@ static void init_asan(void)
 #ifdef CFG_WITH_PAGER
 	asan_tag_access(__pageable_start, __pageable_end);
 #endif /*CFG_WITH_PAGER*/
+#ifndef CFG_DYN_CONFIG
 	asan_tag_access(__nozi_start, __nozi_end);
+#endif
 #ifdef ARM32
 	asan_tag_access(__exidx_start, __exidx_end);
 	asan_tag_access(__extab_start, __extab_end);
@@ -880,6 +883,15 @@ static bool add_padding_to_pool(vaddr_t va, size_t len, void *ptr __unused)
 	return true;
 }
 
+static bool calc_padding_size(vaddr_t va __unused, size_t len, void *ptr)
+{
+	size_t *tot_size = ptr;
+
+	if (len >= MALLOC_INITIAL_POOL_MIN_SIZE)
+		(*tot_size) += len;
+	return false; /* don't consume */
+}
+
 static void init_primary(unsigned long pageable_part)
 {
 	vaddr_t va = 0;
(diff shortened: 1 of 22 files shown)


jenswikl:
CI is passing. There are a couple of false positives from checkpatch.

core/mm/mobj.c
```diff
@@ -196,6 +197,36 @@ struct mobj *mobj_phys_alloc(paddr_t pa, size_t size, uint32_t mem_type,
...
+
+	m->flags = f;
+	m->battr = battr;
+	m->mem_type = (core_mmu_type_to_attr(memtype) >>
+		       TEE_MATTR_MEM_TYPE_SHIFT) & TEE_MATTR_MEM_TYPE_MASK;
+	m->mobj.size = size;
+	m->mobj.ops = &mobj_phys_ops;
+	refcount_set(&m->mobj.refc, 1);
+	m->pa = pa;
+	m->va = va;
```
etienne-lms, line 229: Allow one to provide a NULL `va` value?
We have this test in `mobj_phys_alloc()` whre only SDP buffer may not have a VA.

core/mm/mobj.c
```diff
@@ -196,6 +197,36 @@ struct mobj *mobj_phys_alloc(paddr_t pa, size_t size, uint32_t mem_type,
 	return mobj_phys_init(pa, size, mem_type, battr, area_type);
 }
 
+struct mobj *mobj_phys_alloc_flags(vaddr_t va, paddr_t pa, size_t size,
+				   enum teecore_memtypes memtype,
+				   enum buf_is_attr battr, uint32_t flags)
+{
+	uint32_t f = (flags & MAF_NEX) | MAF_ZERO_INIT;
```
etienne-lms, line 204: Should this return an error if flag other than `MAF_NEX` and `MAF_ZERO_INIT` is requested?

core/arch/arm/kernel/thread.c
```diff
@@ -561,6 +570,9 @@ set_core_local_kcode_offset(struct thread_core_local *cls, long offset)
 
 static void init_user_kcode(void)
 {
+	__maybe_unused struct mobj *m;
```
etienne-lms: `= NULL`

core/arch/arm/kernel/thread.c
```diff
@@ -1056,10 +1082,15 @@ void thread_get_user_kdata(struct mobj **mobj, size_t *offset,
...
+#ifdef CFG_DYN_CONFIG
+	*mobj = thread_user_kdata_page_mobj;
+	*offset = 0;
+#else
 	*mobj = mobj_tee_ram_rw;
-	*sz = sizeof(thread_user_kdata_page);
 	*offset = (vaddr_t)thread_user_kdata_page -
 		  (vaddr_t)mobj_get_va(*mobj, 0, *sz);
+#endif
+	*sz = thread_user_kdata_page_size;
```
etienne-lms, line 1130: size if always `ROUNDUP(sizeof(struct thread_core_local) * CFG_TEE_CORE_NB_CORE, SMALL_PAGE_SIZE)`. Is it worth using a memory cell to store it?

core/arch/arm/kernel/boot.c
```diff
@@ -332,7 +333,9 @@ static void init_asan(void)
 #ifdef CFG_WITH_PAGER
 	asan_tag_access(__pageable_start, __pageable_end);
 #endif /*CFG_WITH_PAGER*/
+#ifndef CFG_DYN_CONFIG
 	asan_tag_access(__nozi_start, __nozi_end);
```
etienne-lms, line 339: MMU tables may still be allocated from the nozi section. Shouldn't it still be tagged for ASAN access in these cases?

etienne-lms (commented):
Some feedback. I've not cover the whole series yet and I'll need to re-read some commits. In the mean time:

`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits:
"core: boot_mem: remove multiple of align restriction" and
"core: simplify core_init_mmu_map()".

"core: mm: update is_nexus()",
"core: mm: add mobj_page_alloc()",
"core: mm: pgt_init(): use virt_page_alloc() if available" and
"core: set uref_base for uref calculations".

Commits that LGTM with questions or minor comments:
"core: mm: add mobj_phys_alloc_flags()", and
"core: arm: init_user_kcode(): use mobj_page_alloc() if available".

core/mm/mobj.c
jenswikl: No, the idea is to ignore unknown or "don't care" MAF-flags to allow flags to propagate between layers without all the functions the chain knowing all the flags.

core/mm/mobj.c
jenswikl: Good point, I'll add a check.

core/arch/arm/kernel/thread.c
jenswikl: In a not too distant future it will become `ROUNDUP(sizeof(struct thread_core_local) * thread_core_count, SMALL_PAGE_SIZE)`. I can wrap it in a helper function instead if you prefer that.

core/arch/arm/kernel/boot.c
jenswikl: `__nozi_start` and `__nozi_end` doesn't exist with `CFG_DYN_CONFIG`. This is instead covered by the `boot_mem_init_asan()` call above.

core/arch/arm/kernel/thread.c
jenswikl: Yes, I'll fix.

jenswikl:
Fixed up "core: mm: add mobj_phys_alloc_flags()" and "core: arm: init_user_kcode(): use mobj_page_alloc() if available" as
requested.

Tags applied.

jenswikl:
Rebased on master

etienne-lms:
Hi @jenswi-linaro, sorry i've left this P-R for quite a while now. The last time I tested it on my HW, it seemed the pagestore hashes where corrupted during OP-TEE initialization. I'll check back on the rebased update  and tell you.

jenswikl:
Rebased with fixes commit for the problems I found when testing on Hikey with pager enabled.

jenswikl:
Rebased on master to get the CI updates.

jenswikl:
Rebased on master to get the CI updates.

etienne-lms:
Sorry for this very late feedback.
I've tested over stm32mp1 (32bit) with pager. I face issues when enabling `CFG_TEE_CORE_DEBUG`: some console trace messages are corrupted and boot hangs after pager initialization. Strangely, no issue when ``CFG_TEE_CORE_DEBUG=n` and the TZSRAM is drastically shortened. I could not find find why. I need to investigate further.

jenswikl:
Thanks for testing.

In generic code, CFG_TEE_CORE_DEBUG enabled asserts, detailed panics, double-check virt_to_phys() and vice versa, and deadlock detection for spinlocks. The latter two seem more suspect. Perhaps you can try modifying the code to disable one of them independently?

etienne-lms:
Hi. I've faced issues with these changes on stm32mp15 paged target. It consumes to many internal RAM leaving few pages to the pager pool. Resulting core is very very slow. I tried to find a way to reused wasted pages but failed... and missed to notify you at the time.
Sorry for this very very late feedback.

jenswikl:
So it works then? Even if it's not usable on your platform. How about trying to optimize memory usage in a later PR? Do you know how many physical pages became unavailable for paging?

Another option is to forget about dynamic configuration for a pager setup and instead focus on isolating the pager setup from the dynamic configuration.

jenswikl:
@etienne-lms, as agreed during the OP-TEE Contributions Forum.
Rebased and squashed in the updates. There are a few commits that are missing review tags.

With this, `CFG_DYN_CONFIG` will be enabled by default for all arm configurations. However, it can still be configured with `CFG_DYN_CONFIG=n` if needed. We will not make `CFG_DYN_CONFIG=y` mandatory until all relevant platforms work with it.

etienne-lms:
Keep alive.

PR #7427, opened 2025-06-04
mk/config.mk: disable CFG_CALLOUT for CFG_CORE_SEL2_SPMC=y

jenswi-linaro:
With Hafnium at S-EL1 (CFG_CORE_SEL2_SPMC=y) the callout service isn't initialized. To avoid unexpected aborts and errors, set CFG_CALLOUT=n.


jforissier (commented):

jenswikl:
Tag applied

PR #7431, opened 2025-06-11
KASAN: several minor fixes / selftests

alexhoppus:
Hi,

this MR proposes several improvements for **KASAN**.

**Adds runtime KASAN tests**
Added a dedicated KASAN self-test suite covering:
- Stack overflow
- Global buffer overflow
- Heap overflow
- Use-after-free
- Invalid accesses using memcpy / memset

These tests could help to track potential regression in KASAN functionality.

**Fixes check_access()**

It seems implementation didn't cover some corner cases, please see commit msg for details.
I borrowed some [freebsd code](https://github.com/freebsd/freebsd-src/blob/main/sys/kern/subr_asan.c) to rewrite [check_access](https://github.com/OP-TEE/optee_os/blob/master/core/kernel/asan.c#L169) in correct/optimized way.

 **Some changes in heap tagging**

It seems in some cases it tag more memory as available than it should.

**Adds support for unpoisoning stack on longjmp**

KASAN tests use **setjmp/longjmp** , to utilize it alongside with **KASAN**
some additional support is needed.

Please let me know if these changes would be useful for you.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -327,6 +327,7 @@ jobs:
 
           make -j$(nproc) check CFG_LOCKDEP=y CFG_LOCKDEP_RECORD_STACK=n CFG_IN_TREE_EARLY_TAS=pkcs11/fd02c9da-306c-48c7-a49c-bbd827ae86ee CFG_PKCS11_TA=y CFG_CORE_UNSAFE_MODEXP=y XTEST_ARGS="-x pkcs11_1007"
           make -j$(nproc) check CFG_CORE_SANITIZE_KADDRESS=y CFG_CORE_ASLR=n CFG_ATTESTATION_PTA=n XTEST_ARGS="n_1001 n_1003 n_1004"
+          make -j$(nproc) check CFG_CORE_SANITIZE_KADDRESS=y CFG_CORE_ASLR=n CFG_ATTESTATION_PTA=n CFG_DYN_CONFIG=n XTEST_ARGS="n_1001 n_1003 n_1004"
 
   QEMUv8_check1:
     name: make check (QEMUv8) 1 / 4
--- a/core/kernel/asan.c
+++ b/core/kernel/asan.c
@@ -9,6 +9,7 @@
 #include <kernel/asan.h>
 #include <kernel/panic.h>
 #include <printk.h>
+#include <setjmp.h>
 #include <string.h>
 #include <trace.h>
 #include <types_ext.h>
@@ -326,4 +327,13 @@ void __asan_unregister_globals(struct asan_global *globals, size_t size);
 void __asan_unregister_globals(struct asan_global *globals __unused,
 			       size_t size __unused)
 {
+}
+
+void asan_handle_longjmp(void *old_sp)
+{
+	void *top = old_sp;
+	void *bottom = (void *)ROUNDDOWN((vaddr_t)&top,
+					 ASAN_BLOCK_SIZE);
+
+	asan_tag_access(bottom, top);
 }
--- a/core/kernel/asan.c
+++ b/core/kernel/asan.c
@@ -8,6 +8,7 @@
 #include <keep.h>
 #include <kernel/asan.h>
 #include <kernel/panic.h>
+#include <printk.h>
 #include <string.h>
 #include <trace.h>
 #include <types_ext.h>
@@ -41,6 +42,7 @@ struct asan_global {
 static vaddr_t asan_va_base;
 static size_t asan_va_size;
 static bool asan_active;
+static asan_panic_cb_t asan_panic_cb = asan_panic;
 
 static int8_t *va_to_shadow(const void *va)
 {
@@ -166,6 +168,48 @@ void asan_start(void)
 	asan_active = true;
 }
 
+void __noreturn asan_panic(void)
+{
+	panic();
+}
+
+void asan_set_panic_cb(asan_panic_cb_t panic_cb)
+{
+	asan_panic_cb = panic_cb;
+}
+
+static void asan_report(vaddr_t addr, size_t size)
+{
+#ifdef KASAN_DUMP_SHADOW
+	char buf[128] = {0};
+	int r = 0, rc = 0;
+	vaddr_t b = 0, e = 0, saddr = 0;
+
+	b = ROUNDDOWN(addr, ASAN_BLOCK_SIZE) - ASAN_BLOCK_SIZE;
+	e = ROUNDDOWN(addr, ASAN_BLOCK_SIZE) + ASAN_BLOCK_SIZE;
+
+	/* Print shadow map nearby */
+	if (va_range_inside_shadow((void *)b, (void *)e)) {
+		rc = snprintk(buf + r, sizeof(buf) - r, "%lx: ", b);
+		assert(rc > 0);
+		r += rc;
+		for (saddr = b; saddr <= e; saddr += ASAN_BLOCK_SIZE) {
+			int8_t *sbyte = va_to_shadow((void *)saddr);
+
+			rc = snprintk(buf + r, sizeof(buf) - r,
+				      "0x%02x ", (uint8_t)*sbyte);
+			assert(rc > 0);
+			r += rc;
+		}
+		EMSG("%s", buf);
+	}
+#endif
+	EMSG("[ASAN]: access violation, addr: %lx size: %zu\n",
+	     addr, size);
+
+	asan_panic_cb();
+}
+
 static void check_access(vaddr_t addr, size_t size)
 {
 	void *begin = (void *)addr;
@@ -187,11 +231,11 @@ static void check_access(vaddr_t addr, size_t size)
 	e = va_to_shadow((void *)(addr + size - 1));
 	for (a = va_to_shadow(begin); a <= e; a++)
 		if (*a < 0)
-			panic();
+			asan_report(addr, size);
 
 	if (!va_is_well_aligned(end) &&
 	    va_misalignment(end) > (size_t)(*e - ASAN_BLOCK_SIZE))
-		panic();
+		asan_report(addr, size);
 }
 
 static void check_load(vaddr_t addr, size_t size)
--- a/core/kernel/asan.c
+++ b/core/kernel/asan.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
  * Copyright (c) 2016, Linaro Limited
+ * Copyright (c) 2018-2020 Maxime Villard, m00nbsd.net
  */
 
 #include <assert.h>
@@ -44,6 +45,12 @@ static vaddr_t asan_va_base;
 static size_t asan_va_size;
 static bool asan_active;
 static asan_panic_cb_t asan_panic_cb = asan_panic;
+
+static bool addr_crosses_scale_boundary(vaddr_t addr, size_t size)
+{
+	return (addr >> ASAN_BLOCK_SHIFT) !=
+	       ((addr + size - 1) >> ASAN_BLOCK_SHIFT);
+}
 
 static int8_t *va_to_shadow(const void *va)
 {
@@ -122,7 +129,7 @@ void asan_tag_access(const void *begin, const void *end)
 	asan_memset_unchecked(va_to_shadow(begin), 0,
 			      va_range_to_shadow_size(begin, end));
 	if (!va_is_well_aligned(end))
-		*va_to_shadow(end) = ASAN_BLOCK_SIZE - va_misalignment(end);
+		*va_to_shadow(end) = va_misalignment(end);
 }
 
 void asan_tag_heap_free(const void *begin, const void *end)
@@ -211,14 +218,87 @@ static void asan_report(vaddr_t addr, size_t size)
 	asan_panic_cb();
 }
 
-static void check_access(vaddr_t addr, size_t size)
+static __always_inline bool asan_shadow_1byte_isvalid(vaddr_t addr)
 {
+	int8_t last = (addr & ASAN_BLOCK_MASK) + 1;
+	int8_t *byte = va_to_shadow((void *)addr);
+
+	if (*byte == 0 || last <= *byte)
+		return true;
+
+	return false;
+}
+
+static __always_inline bool asan_shadow_2byte_isvalid(vaddr_t addr)
+{
+	if (addr_crosses_scale_boundary(addr, 2)) {
+		return (asan_shadow_1byte_isvalid(addr) &&
+			asan_shadow_1byte_isvalid(addr + 1));
+	} else {
+		int8_t last = ((addr + 1) & ASAN_BLOCK_MASK) + 1;
+		int8_t *byte = va_to_shadow((void *)addr);
+
+		if (*byte == 0 || last <= *byte)
+			return true;
+
+		return false;
+	}
+}
+
+static __always_inline bool asan_shadow_4byte_isvalid(vaddr_t addr)
+{
+	if (addr_crosses_scale_boundary(addr, 4)) {
+		return (asan_shadow_2byte_isvalid(addr) &&
+			asan_shadow_2byte_isvalid(addr + 2));
+	} else {
+		int8_t last = ((addr + 3) & ASAN_BLOCK_MASK) + 1;
+		int8_t *byte = va_to_shadow((void *)addr);
+
+		if (*byte == 0 || last <= *byte)
+			return true;
+
+		return false;
+	}
+}
+
+static __always_inline bool asan_shadow_8byte_isvalid(vaddr_t addr)
+{
+	if (addr_crosses_scale_boundary(addr, 8)) {
+		return (asan_shadow_4byte_isvalid(addr) &&
+			asan_shadow_4byte_isvalid(addr + 4));
+	} else {
+		int8_t last = ((addr + 7) & ASAN_BLOCK_MASK) + 1;
+		int8_t *byte = va_to_shadow((void *)addr);
+
+		if (*byte == 0 || last <= *byte)
+			return true;
+
+		return false;
+	}
+}
+
+static __always_inline bool asan_shadow_Nbyte_isvalid(vaddr_t addr,
+						      size_t size)
+{
+	size_t i = 0;
+
+	for (; i < size; i++) {
+		if (!asan_shadow_1byte_isvalid(addr + i))
+			return false;
+	}
+
+	return true;
+}
+
+static __always_inline void check_access(vaddr_t addr, size_t size)
+{
+	bool valid = false;
 	void *begin = (void *)addr;
 	void *end = (void *)(addr + size);
-	int8_t *a;
-	int8_t *e;
 
-	if (!asan_active || !size)
+	if (!asan_active)
+		return;
+	if (size == 0)
 		return;
 	if (va_range_outside_shadow(begin, end))
 		return;
@@ -229,22 +309,38 @@ static void check_access(vaddr_t addr, size_t size)
 	if (!va_range_inside_shadow(begin, end))
 		panic();
 
-	e = va_to_shadow((void *)(addr + size - 1));
-	for (a = va_to_shadow(begin); a <= e; a++)
-		if (*a < 0)
-			asan_report(addr, size);
+	if (__builtin_constant_p(size)) {
+		switch (size) {
+		case 1:
+			valid = asan_shadow_1byte_isvalid(addr);
+			break;
+		case 2:
+			valid = asan_shadow_2byte_isvalid(addr);
+			break;
+		case 4:
+			valid = asan_shadow_4byte_isvalid(addr);
+			break;
+		case 8:
+			valid = asan_shadow_8byte_isvalid(addr);
+			break;
+		default:
+			valid = asan_shadow_Nbyte_isvalid(addr, size);
+			break;
+		}
+	} else {
+		valid = asan_shadow_Nbyte_isvalid(addr, size);
+	}
 
-	if (!va_is_well_aligned(end) &&
-	    va_misalignment(end) > (size_t)(*e - ASAN_BLOCK_SIZE))
+	if (!valid)
 		asan_report(addr, size);
 }
 
-static void check_load(vaddr_t addr, size_t size)
+static __always_inline void check_load(vaddr_t addr, size_t size)
 {
 	check_access(addr, size);
 }
 
-static void check_store(vaddr_t addr, size_t size)
+static __always_inline void check_store(vaddr_t addr, size_t size)
 {
 	check_access(addr, size);
 }
--- a/core/pta/tests/misc.c
+++ b/core/pta/tests/misc.c
@@ -5,11 +5,13 @@
  */
 #include <assert.h>
 #include <config.h>
+#include <kernel/asan.h>
 #include <kernel/dt_driver.h>
 #include <kernel/linker.h>
 #include <kernel/panic.h>
 #include <malloc.h>
 #include <mm/core_memprot.h>
+#include <setjmp.h>
 #include <stdbool.h>
 #include <trace.h>
 #include <util.h>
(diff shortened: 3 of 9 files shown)


core/pta/tests/misc.c
```diff
@@ -658,14 +660,195 @@ static int self_test_va2pa(void)
...
+static int asan_call_test(struct asan_test_ctx *ctx,
+			  void (*test)(struct asan_test_ctx *ctx),
+			  const char __maybe_unused *desc)
+{
+	int ret;
+
+	if (setjmp(asan_test_jmp) == ASAN_TEST_SUCCESS) {
+		ret = 0;
+	} else {
+		test(ctx);
```
jenswikl, line 724: This should only be done if `setjmp()` returns `0`.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,195 @@ static int self_test_va2pa(void)
...
+		ret = -1;
+	}
+
+	LOG("  => [asan] test %s: %s", desc, !ret ? "ok" : "FAILED");
+	return ret;
+}
+
+#pragma GCC diagnostic push
```
jenswikl: I'd prefer that we didn't use these pragmas. Can the code be written with clever helper functions to avoid the warning and possibly also avoid the volatile attribute? Perhaps by adding a write function pointer to `struct asan_test_ctx`?

core/pta/tests/misc.c
```diff
@@ -658,14 +660,195 @@ static int self_test_va2pa(void)
...
+	(void)p;
+}
+
+static void asan_use_after_free(struct asan_test_ctx *ctx)
+{
+	char *a = malloc(ASAN_TEST_BUF_SIZE);
+
+	if (a) {
+		free(a);
+		a[0] = ctx->write_value;
```
jenswikl: How about a case for reading after free also?

core/pta/tests/misc.c
```diff
@@ -658,14 +660,195 @@ static int self_test_va2pa(void)
 	return ret;
 }
 
+#ifdef CFG_CORE_SANITIZE_KADDRESS
+
+#define ASAN_TEST_SUCCESS 1
+#define ASAN_TEST_BUF_SIZE 15
+
+static volatile char sgbuf[ASAN_TEST_BUF_SIZE] = {0};
```
jenswikl: Globals are always zero-initialized. Please try to avoid the `volatile` attribute using a helper function if needed.

lib/libutils/isoc/arch/arm/setjmp_a64.S
```diff
@@ -82,6 +82,13 @@ BTI(	bti	c)
 	bl	ftrace_longjmp
 	ldp	x29, x30, [sp], #16
 	ldp	x0, x1, [sp], #16
+#endif
+#if defined(__KERNEL__) && defined(CFG_CORE_SANITIZE_KADDRESS) && \
```
jenswikl, line 86: Please fix Arm32 and Risc-v too.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -230,8 +231,7 @@ static void *maybe_tag_buf(uint8_t *buf, size_t hdr_size, size_t requested_size)
 	}
 
 #if defined(__KERNEL__)
-	if (IS_ENABLED(CFG_CORE_SANITIZE_KADDRESS))
```
jenswikl, line 233: This change should go into a separate commit, same below with the other `IS_ENABLED(CFG_CORE_SANITIZE_KADDRESS)`.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -218,8 +218,9 @@ static void *maybe_tag_buf(uint8_t *buf, size_t hdr_size, size_t requested_size)
 	COMPILE_TIME_ASSERT(MEMTAG_GRANULE_SIZE <= SizeQuant);
 
 	if (MEMTAG_IS_ENABLED) {
-		size_t sz = ROUNDUP(requested_size, MEMTAG_GRANULE_SIZE);
+		size_t sz = MAX(SizeQuant, requested_size);
```
jenswikl: How come this is needed for `MEMTAG_IS_ENABLED` when it's not needed for `CFG_CORE_SANITIZE_KADDRESS`?

core/kernel/asan.c
```diff
@@ -213,14 +220,93 @@ static void asan_report(vaddr_t addr, size_t size)
 		asan_panic_cb();
 }
 
-static void check_access(vaddr_t addr, size_t size)
+static __always_inline bool asan_shadow_1byte_isvalid(unsigned long addr)
```
jenswikl: Why are all the `__always_inline` needed?

core/kernel/asan.c
```diff
@@ -45,6 +46,12 @@ static size_t asan_va_size;
 static bool asan_active;
 static asan_panic_cb_t asan_panic_cb = asan_panic;
 
+static inline bool addr_crosses_scale_boundary(uintptr_t addr, size_t size)
```
jenswikl: Please drop the `inline`.
I'd prefer to use `vaddr_t` type when manipulating virtual addresses represented as integers, unless there's a good reason to use another type.

core/pta/tests/misc.c
alexhoppus: Sure. We can also do all tests in both (r,w) variants, may I ask, why you want exactly uaf read test? In the end all accesses are checked identically in `check_access` now, so I was in doubt whether we need to test both read and write.

jenswikl: You're right, it's not needed.

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

lib/libutils/isoc/arch/arm/setjmp_a64.S
alexhoppus: > Please fix Arm32

Done

> and Risc-v too.

I'm not sure that I can do that. As far as I see asan support is disabled for risc v for [virt](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/plat-virt/conf.mk#L21) , [spike](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/plat-spike/conf.mk#L34)  and [sifive](https://github.com/OP-TEE/optee_os/blob/master/core/arch/riscv/plat-sifive/conf.mk#L18), which are all available platforms. That means if I understand correctly, that I can't test this code. I would prefer to add compilation assert instead. Let me know if that's okay, or if I'm missing something.

lib/libutils/isoc/bget_malloc.c
alexhoppus: Done

lib/libutils/isoc/bget_malloc.c
alexhoppus: I removed `MAX(SizeQuant, requested_size)` but now CI fails, I need more time to understand better what is happening there.

core/kernel/asan.c
alexhoppus: So the idea is straightforward, since the asan hooks are used widely in the code (every load or store) inline everything for performance reasons. Actually that is [from](https://cgit.freebsd.org/src/tree/sys/kern/subr_asan.c?id=ea3fbe0707f9a02a29875966668b6f15284f335a#n291) free-bsd kernel, I didn't change that.

core/kernel/asan.c
alexhoppus: > Please drop the inline.

Ok, please read my [comment above](https://github.com/OP-TEE/optee_os/pull/7431#discussion_r2144433782) if the inlines are not needed I will remove them.

> I'd prefer to use vaddr_t type when manipulating virtual addresses represented as integers, unless there's a good reason to use another type.

Done.

core/kernel/asan.c
jenswikl: Did you notice any difference in performance? Usually, it's better to allow the compiler to optimize freely unless it's known to do the wrong thing in a particular case.

core/kernel/asan.c
jenswikl: This `inline` means nothing since it's a static function in the same .c file.

core/kernel/asan.c
```diff
@@ -283,3 +432,13 @@ void __asan_unregister_globals(struct asan_global *globals __unused,
 			       size_t size __unused)
 {
 }
+
+void asan_handle_longjmp(void *old_sp)
+{
+	int local = 0;
+	void *top = old_sp;
+	void *bottom = (void *)ROUNDDOWN((uintptr_t)&local,
```
jenswikl: s/uintptr_t/vaddr_t/
Can we take the address of `top` or `bottom` instead?

core/pta/tests/misc.c
```diff
@@ -665,17 +665,30 @@ static int self_test_va2pa(void)
...
+	char *pmalloc2[3];
 	char write_value;
+	void (*write_func)(char *buf, size_t pos, char value);
+	char (*read_func)(char *buf, size_t pos);
 };
 
+static void asan_out_of_bounds_write(char *buf, size_t pos,
+				     char value)
+{
+	*(buf + pos) = value;
```
jenswikl: This looks like a complicated way of doing `buf[pos] = value`

core/pta/tests/misc.c
```diff
@@ -665,17 +665,30 @@ static int self_test_va2pa(void)
...
+}
+
+static char asan_out_of_bounds_read(char *buf, size_t pos)
+{
+	return *(buf + pos);
```
jenswikl: `buf[pos]`?

core/pta/tests/misc.c
```diff
@@ -768,31 +775,31 @@ static void asan_use_after_free(struct asan_test_ctx *ctx)
...
 }
 
 static void asan_memcpy_dst(struct asan_test_ctx *ctx __maybe_unused)
 {
-	static char b[ASAN_TEST_BUF_SIZE + 1] = {0};
-	static char a[ASAN_TEST_BUF_SIZE] = {0};
+	static char b[ASAN_TEST_BUF_SIZE + 1];
+	static char a[ASAN_TEST_BUF_SIZE];
 
 	memcpy(a, b, sizeof(b));
```
jenswikl: It looks like the compiler might optimize away this call.

lib/libutils/isoc/arch/arm/setjmp_a64.S
jenswikl: Sure, a static assert works too.

gagachang: @alexhoppus RISC-V hasn't supported KASAN. A static assert is enough, thanks!

alexhoppus: Done

lib/libutils/isoc/bget_malloc.c
alexhoppus: > How come this is needed for MEMTAG_IS_ENABLED when it's not needed for CFG_CORE_SANITIZE_KADDRESS?

So the asan and memtag are **not** really symmetric after my change. asan tags amount of bytes actually allocated, from the perspective of malloc user, if we do `malloc(5)`  5 bytes will be tagged in shadow map. Memtag tags actual space allocated from the perspective of `bget`, and the minimal quant here is 16 what is in fact afaik a granularity limit for memtag. In this sense asan can do more precise tagging.

This `MAX(SizeQuant, requested_size)` is actually needed for one corner case, when `requested_size == 0`, in this case we need to pass `requested_size == SizeQuant` further to `memtag` because that is what actually will be allocated by `bget`. The asan in this case will mark nothing in shadow map.

core/kernel/asan.c
alexhoppus: > Did you notice any difference in performance?

I see median `sys time` difference `time xtest -t regression 1001 1002 1003 1004` (10 measurements) 3.2s (all inlines removed) vs 2.6s (everything as in the MR) . It also could be seen from disassembly, that compiler doesn't inline at least `check_access` and some of `asan_shadow_*byte_is_valid` functions. 

Anyway, that is a quick test I came up with, it's not completely clear to me how representative these results actually are.  The another way would be to abandon these "inlines" and introduce them in a separate MR with the proper measurements.

core/kernel/asan.c
alexhoppus: Done

core/kernel/asan.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Several possibilities here:

**Variant a)**

I can reintroduce volatile despite checkpatch warnings. I consider this a false positive, since the use of volatile here is intentional - to prevent optimization. The kernel document referenced by the checkpatch explicitly states:

> The key point to understand with regard to volatile is that its purpose is
to suppress optimization, which is almost never what one really wants to
do.

**Variant b)**
`__attribute__((optimize("O0")))` through compiler agnostic macros

jenswikl: Perhaps it's better to register `memcpy()` and `memset()` as callbacks to avoid the compiler being too clever.

alexhoppus: > Perhaps it's better to register memcpy() and memset() as callbacks to avoid the compiler being too clever.

Done

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+{
+	char *a = malloc(ASAN_TEST_BUF_SIZE);
+
+	if (a) {
+		free(a);
+		ctx->write_func(a, 0, ctx->write_value);
+	}
+}
+
+static void asan_memcpy_dst(struct asan_test_ctx *ctx __maybe_unused)
```
jenswikl: Please remove `__maybe_unused`. Same below.

alexhoppus: Done

alexhoppus:
Squashed

core/kernel/asan.c
```diff
@@ -166,6 +168,50 @@ void asan_start(void)
...
+			r += rc;
+		}
+		EMSG("%s", buf);
+	}
+#endif
+	EMSG("[ASAN]: access violation, addr: %lx size: %zu\n",
+	     addr, size);
+
+	assert(asan_panic_cb);
+	if (asan_panic_cb)
```
jenswikl: This check shouldn't be needed. If it needs to be checked, it should be done by `asan_set_panic_cb()`, but I doubt it's worth it.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
 	return ret;
 }
 
+#ifdef CFG_CORE_SANITIZE_KADDRESS
+
+#define ASAN_TEST_SUCCESS 1
+#define ASAN_TEST_BUF_SIZE 15
+
+static char sgbuf[ASAN_TEST_BUF_SIZE];
+char gbuf[ASAN_TEST_BUF_SIZE];
```
jenswikl: Please add a suitable prefix if this can't be a static.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+
+static jmp_buf asan_test_jmp;
+
+struct asan_test_ctx {
+	char *pmalloc1;
+	char *pmalloc2[3];
+	char write_value;
+	void (*write_func)(char *buf, size_t pos, char value);
+	char (*read_func)(char *buf, size_t pos);
```
jenswikl: It seems this is unused.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+{
+	longjmp(asan_test_jmp, ASAN_TEST_SUCCESS);
+}
+
+static void asan_test_cleanup(struct asan_test_ctx *ctx)
+{
+	unsigned int i;
+
+	if (ctx->pmalloc1)
+		free(ctx->pmalloc1);
```
jenswikl: No need to check for NULL before calling `free()`. Same below.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+
+	for (i = 0; i < sizeof(ctx->pmalloc2) / sizeof(uint8_t *); i++) {
```
jenswikl: Please use the `ARRAY_SIZE()` macro from `<util.h.>`

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+	ctx->pmalloc1 = malloc(ASAN_TEST_BUF_SIZE);
+
+	if (ctx->pmalloc1)
+		ctx->write_func(ctx->pmalloc1, ASAN_TEST_BUF_SIZE,
+				ctx->write_value);
+}
+
+static void asan_malloc2(struct asan_test_ctx *ctx)
+{
+	unsigned int i;
```
jenswikl: All local variables should be initialized.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+static void asan_memset(struct asan_test_ctx *ctx)
+{
+	static char b[ASAN_TEST_BUF_SIZE];
+
+	ctx->memset_func(b, ctx->write_value, ASAN_TEST_BUF_SIZE + 1);
+}
+
+static int self_test_asan(void)
+{
+	uint32_t vfp_state;
```
jenswikl: All local variables should be initialized.

core/pta/tests/misc.c
```diff
@@ -658,14 +660,218 @@ static int self_test_va2pa(void)
...
+	asan_set_panic_cb(asan_panic_test);
+	/*
+	 * We need enable access to floating-point registers, in other
+	 * way sync exception during setjmp/longjmp will occur.
+	 */
+	vfp_state = thread_kernel_enable_vfp();
+
+	if (asan_call_test(&ctx, asan_global_stat, "(s) glob overflow") ||
+	    asan_call_test(&ctx, asan_global, "glob overflow") ||
+#ifndef CFG_DYN_CONFIG
```
jenswikl, line 833: How about always calling `asan_stack()` and instead relying on `if (IS_ENABLED(CFG_DYN_CONFIG))` inside that function to decide if anything should be done. That should eliminate the need for so many ifdefs.

lib/libutils/isoc/arch/arm/setjmp_a32.S
```diff
@@ -252,6 +252,18 @@ SYM (\name):
 	bl		ftrace_longjmp
 	ldmia		sp!, { a1, a2, lr }
 #endif
+#if defined(__KERNEL__) && defined(CFG_CORE_SANITIZE_KADDRESS) && \
+!defined(CFG_DYN_CONFIG)
+	stmdb		sp!, { a1, a2, a3, lr }
+
+#ifdef __thumb2__
+	ldr		a1, [a1, #32]
```
jenswikl, line 260: How can this work? `a1` is a passed integer value.
By the way, `asan_handle_longjmp()` only uses `a0` aka `r0`.

lib/libutils/isoc/arch/arm/setjmp_a64.S
```diff
@@ -82,6 +82,13 @@ BTI(	bti	c)
...
+!defined(CFG_DYN_CONFIG)
+	stp	x0, x1, [sp, #-16]!
+	REG_ONE(x0, 96)
```
jenswikl: Please use `ldr` directly instead of the `REG_ONE()` macro.

lib/libutils/isoc/arch/arm/setjmp_a64.S
```diff
@@ -82,6 +82,13 @@ BTI(	bti	c)
 	bl	ftrace_longjmp
 	ldp	x29, x30, [sp], #16
 	ldp	x0, x1, [sp], #16
+#endif
+#if defined(__KERNEL__) && defined(CFG_CORE_SANITIZE_KADDRESS) && \
+!defined(CFG_DYN_CONFIG)
```
jenswikl: Please indent

lib/libutils/isoc/arch/arm/setjmp_a32.S
```diff
@@ -252,6 +252,18 @@ SYM (\name):
 	bl		ftrace_longjmp
 	ldmia		sp!, { a1, a2, lr }
 #endif
+#if defined(__KERNEL__) && defined(CFG_CORE_SANITIZE_KADDRESS) && \
+!defined(CFG_DYN_CONFIG)
```
jenswikl: Please indent.

lib/libutils/isoc/arch/riscv/setjmp_rv.S
```diff
@@ -62,6 +62,10 @@ FUNC longjmp , :
 	LDR	s0, REGOFF(2)(sp)
 	LDR	ra, REGOFF(3)(sp)
 	addi	sp, sp, REGOFF(4)
+#endif
+#if defined(__KERNEL__) && defined(CFG_CORE_SANITIZE_KADDRESS) && \
+!defined(CFG_DYN_CONFIG)
```
jenswikl: Please indent.

lib/libutils/isoc/bget_malloc.c
```diff
@@ -218,7 +218,12 @@ static void *maybe_tag_buf(uint8_t *buf, size_t hdr_size, size_t requested_size)
 	COMPILE_TIME_ASSERT(MEMTAG_GRANULE_SIZE <= SizeQuant);
 
 	if (MEMTAG_IS_ENABLED) {
-		size_t sz = ROUNDUP(requested_size, MEMTAG_GRANULE_SIZE);
+		size_t sz;
```
jenswikl: All locals shall be initialized. See https://optee.readthedocs.io/en/latest/general/coding_standards.html point three.

lib/libutils/isoc/bget_malloc.c
jenswikl: Please add a summary of this explanation as a comment in the code.

lib/libutils/isoc/arch/arm/setjmp_a32.S
alexhoppus: > How can this work? a1 is a passed integer value.

There is no `a0`, please take a [look](https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#611core-registers)  at aapcs synonyms. `a1` is the first passed argument i.e. `jump_buf`

> By the way, asan_handle_longjmp() only uses a0 aka r0

`a1, a2` are both caller saved, so they need to be saved. `a3` is for stack alignment [see](https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#6212stack-constraints-at-a-public-interface)

jenswikl: OK, thanks.

core/kernel/asan.c
alexhoppus: Check removed. Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Done

core/pta/tests/misc.c
alexhoppus: Well, that will make whole logic a little bit more complicated, since we will need third test state -> test skipped. (we have 2 now: asan caught an error or asan didn't catch an error). So imo that will introduce more complexity. I'll change it if you still think it's needed.

lib/libutils/isoc/arch/arm/setjmp_a32.S
alexhoppus: Done

lib/libutils/isoc/arch/arm/setjmp_a64.S
alexhoppus: Done

lib/libutils/isoc/arch/arm/setjmp_a64.S
alexhoppus: Done

lib/libutils/isoc/arch/arm/setjmp_a32.S
alexhoppus: Done

lib/libutils/isoc/arch/riscv/setjmp_rv.S
alexhoppus: Done

lib/libutils/isoc/bget_malloc.c
alexhoppus: Done

alexhoppus:
hi @jenswi-linaro comments are addressed here. Should we also probably add **CFG_DYN_CONFIG=n** [here](https://github.com/OP-TEE/optee_os/blob/master/.github/workflows/ci.yml#L329)? That will enable asan stack checks.

jenswikl:
> hi @jenswi-linaro comments are addressed here. Should we also probably add **CFG_DYN_CONFIG=n** [here](https://github.com/OP-TEE/optee_os/blob/master/.github/workflows/ci.yml#L329)? That will enable asan stack checks.

Sure sounds good. Please squash in the updates.

core/kernel/asan.c
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
  * Copyright (c) 2016, Linaro Limited
+ * Copyright (c) 2018-2020 Maxime Villard, m00nbsd.net
```
alexhoppus, line 4: @jenswi-linaro could you please also help with the licensing stuff, I just copied copyright string from the [original](https://github.com/freebsd/freebsd-src/blob/main/sys/kern/subr_asan.c#L4) implementation. Is that enough/correct? Thanks.

jenswikl: Yes, it's OK.

alexhoppus: Done

alexhoppus:
@jenswi-linaro it seems ci fails due to certificates issue

> stdout: fatal: unable to access 'https://git.trustedfirmware.org/TS/trusted-services.git/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none

.github/workflows/ci.yml
```diff
@@ -326,7 +326,7 @@ jobs:
           cd ${TOP}/build
 
           make -j$(nproc) check CFG_LOCKDEP=y CFG_LOCKDEP_RECORD_STACK=n CFG_IN_TREE_EARLY_TAS=pkcs11/fd02c9da-306c-48c7-a49c-bbd827ae86ee CFG_PKCS11_TA=y CFG_CORE_UNSAFE_MODEXP=y XTEST_ARGS="-x pkcs11_1007"
-          make -j$(nproc) check CFG_CORE_SANITIZE_KADDRESS=y CFG_CORE_ASLR=n CFG_ATTESTATION_PTA=n XTEST_ARGS="n_1001 n_1003 n_1004"
+          make -j$(nproc) check CFG_CORE_SANITIZE_KADDRESS=y CFG_CORE_ASLR=n CFG_ATTESTATION_PTA=n CFG_DYN_CONFIG=n XTEST_ARGS="n_1001 n_1003 n_1004"
```
jenswikl, line 330: Testing with `CFG_DYN_CONFIG` enabled is more important since that's the default. We can perhaps test both.

alexhoppus: Done

jenswikl:
For "libutils: tag only actual allocated size in ASan heap tagging" and "libutils: drop redundant IS_ENABLED checks"
please apply:

and for the rest:

alexhoppus:
Tags have been applied.

alexhoppus:
@jenswi-linaro hi, what are further steps here? Thanks.

jforissier:
@alexhoppus we are about to release OP-TEE 4.7.0 this Friday. This will go into the next release.

PR #7432, opened 2025-06-13
Introduce RCC driver STM32MP21 platform

Gabriel-Fernandz:
This patch-set introduces a clock and reset driver for STM32MP21, based on the Arm Cortex-A35.

PR #7437, opened 2025-06-17
drivers: stm32_tamp: reset the platform in case of tamper event detection

GseoC:
This patchset implements a platform API to force a system reset.
This API is called in case of tamper event detection if the behavior on tamper event detection is to reset the platform

### DIFF
--- a/core/drivers/stm32_tamp.c
+++ b/core/drivers/stm32_tamp.c
@@ -18,6 +18,7 @@
 #include <libfdt.h>
 #include <mm/core_memprot.h>
 #include <stdbool.h>
+#include <stm32_util.h>
 
 /* STM32 Registers */
 #define _TAMP_CR1			0x00U
@@ -1067,9 +1068,8 @@ static enum itr_return stm32_tamp_it_handler(struct itr_handler *h __unused)
 				io_setbits32(base + _TAMP_SCR,
 					     _TAMP_SCR_ITAMP(id));
 
-			/* TODO: reset the platform */
 			if (ret & TAMP_CB_RESET)
-				MSG("System will reset");
+				do_reset("Internal tamper event detected");
 		}
 		i++;
 	}
@@ -1093,9 +1093,8 @@ static enum itr_return stm32_tamp_it_handler(struct itr_handler *h __unused)
 				io_setbits32(base + _TAMP_SCR,
 					     _TAMP_SCR_ETAMP(id));
 
-			/* TODO: reset the platform */
 			if (ret & TAMP_CB_RESET)
-				MSG("System will reset");
+				do_reset("External tamper event detected");
 		}
 		i++;
 	}


GseoC:
Hello @etienne-lms, following your comment on the previous TAMP P-R

GseoC:
This P-R has a dependency on #7432 for the STM32MP21 reset bindings. Content should remain the same.

GseoC:
@jenswi-linaro, @jforissier ,
Do you think it's acceptable to rely on the `multi_core_panic_it_handler()`'s SGI to trap the other CPUs in another context than a `panic()`?

jforissier:
> @jenswi-linaro, @jforissier , Do you think it's acceptable to rely on the `multi_core_panic_it_handler()`'s SGI to trap the other CPUs in another context than a `panic()`?

With proper documentation and perhaps some renaming (like s/multi_core_panic_it_handler/multi_core_halt_it_handler/ and s/CFG_HALT_CORES_ON_PANIC_SGI/CFG_HALT_CORES_SGI/), I'd say yes.

GseoC:
Should it be moved to some other file like `core/kernel/tee_misc.c` or to a new file named "core/kernel/core_management.c" (I'm open to naming suggestions of course :))

jforissier:
> Should it be moved to some other file like `core/kernel/tee_misc.c` or to a new file named "core/kernel/core_management.c" (I'm open to naming suggestions of course :))

Well, the main function is `core/kernel/panic.c` is `__do_panic()`and the interrupt handler which halts the other CPUs is a secondary function, so I think the file name is good. It's just about being clear that the handler is not only called "on panic" but "to halt the CPU core".

GseoC:
Ok, I will come up with a patch then, thank you

GseoC:
I tried so split the use cases, adding a new switch. Please tell me if that seems reasonable to you. I forced a panic from a kernel request on stm32mp157c-dk2 board only to make sure I still hit the 2 cores.

core/kernel/panic.c
```diff
@@ -13,52 +13,51 @@
...
-	.it = CFG_HALT_CORES_ON_PANIC_SGI,
-	.handler = multi_core_panic_it_handler,
+static struct itr_handler multi_core_halt_handler __nex_data = {
+	.it = CFG_HALT_CORES_SGI,
+	.handler = multi_core_halt_it_handler,
 };
-DECLARE_KEEP_PAGER(multi_core_panic_handler);
+DECLARE_KEEP_PAGER(multi_core_halt_handler);
 
 static void notify_other_cores(void)
```
jforissier: `halt_other_cores()` would be a better name

mk/config.mk
```diff
@@ -1214,12 +1214,20 @@ CFG_RSA_PUB_EXPONENT_3 ?= n
...
+$(eval $(call cfg-depends-all,CFG_MULTI_CORE_HALTING,CFG_GIC))
+
 # CFG_HALT_CORES_ON_PANIC, when enabled, makes any call to panic() halt the
-# other cores. The feature currently relies on GIC device to trap the other
-# cores using an SGI interrupt specified by CFG_HALT_CORES_ON_PANIC_SGI.
-CFG_HALT_CORES_ON_PANIC ?= n
-CFG_HALT_CORES_ON_PANIC_SGI ?= 15
-$(eval $(call cfg-depends-all,CFG_HALT_CORES_ON_PANIC,CFG_GIC))
+# other cores. This feature relies on the SGI specified by CFG_HALT_CORES_SGI
+# to trap the other cores.
```
jforissier: s/trap/halt/ we need to choose one word, either halt or trap

GseoC: I'm replacing all occurrences in commit message and code.

GseoC:
Comments addressed and tag applied for commit "core: panic: allow core halting on SGI in other cases than panic()", thanks.

jforissier:
The "make (multi-platform)" CI job has a genuine error.

GseoC:
> The "make (multi-platform)" CI job has a genuine error.

Hello, yes, I have a dependency on @Gabriel-Fernandz 's P-R for stm32mp21x build

mk/config.mk
```diff
@@ -1214,12 +1214,20 @@ CFG_RSA_PUB_EXPONENT_3 ?= n
 CFG_CRYPTO_HW_PBKDF2 ?= n
 $(eval $(call cfg-depends-all,CFG_CRYPTO_HW_PBKDF2,CFG_CRYPTO_PBKDF2))
 
+# CFG_MULTI_CORE_HALTING, when enabled, registers an interrupt handler for the
+# Software Generated Interrupt(SGI) specified by: CFG_HALT_CORES_SGI. Upon
```
etienne-lms, line 1218:
```suggestion
# Software Generated Interrupt (SGI) specified by: CFG_HALT_CORES_SGI. Upon

File: core/arch/arm/plat-stm32mp2/main.c [hunk 4]
```diff
@@ -180,3 +182,27 @@ void plat_external_abort_handler(struct abort_info *ai __unused)
 	/* External abort may be due to SERC events */
 	stm32_serc_handle_ilac();
 }
+
+void __noreturn do_reset(const char *str __maybe_unused)
+{
+	struct rstctrl *rstctrl = NULL;
+
+	if (CFG_TEE_CORE_NB_CORE > 1) {
+		/* Rely on multi_core_halt_it_handler() to trap CPUs */
```
etienne-lms: I think you should not use the function label here, in case it changes in the future.
`/* Halt execution of other CPUs */` should be enough.

core/drivers/stm32_tamp.c
```diff
@@ -1067,9 +1068,8 @@ static enum itr_return stm32_tamp_it_handler(struct itr_handler *h __unused)
 				io_setbits32(base + _TAMP_SCR,
 					     _TAMP_SCR_ITAMP(id));
 
-			/* TODO: reset the platform */
 			if (ret & TAMP_CB_RESET)
-				MSG("System will reset");
+				do_reset("Tamper event detected");
```
etienne-lms: `Internal tamper event` here
and `External tamper event` at line 1097?

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` with comments addressed or not.

mk/config.mk
GseoC: I feel like it's clearer as is (closer to hand writing way).

GseoC:
Comments addressed and tags applied, thanks.
Now waiting for the other P-R to be merged to solve the build issue

GseoC:
Rebased to satisfy build dependency.

PR #7442, opened 2025-06-22
core: riscv: fix hartid at secondary hart entry point

peterlin-sifive:
The a0 register is corrupted during enable_mmu, so get secondary hartid from s0 instead.

Fixes: 29661368f51d ("core: riscv: preserve hartid in s0 register at entry point")


peterlin-sifive:
C.c. @gagachang

gagachang:
Thanks!

PR #7443, opened 2025-06-23
[RFC] ta: pkcs11: Panic on init failure

johankor:
How should I deal with failures in `TA_CreateEntryPoint()` for `keep-alive` TAs? If my understanding is correct, that the initialization is only run once (regardless of failure), then it seems to me we need to tear down the TA (and session+ctx?).

What I'm observing is that the first client fails as expected (the session won't open) if initialization fails, but for the next client the initialization is always skipped, and in turn it ends up causing an abort as it tries to access uninitialized memory.

The commit in this PR is just to highlight the issue - I realize it's not a solution (if there's even a problem). If it turns out there is a problem though, I'd be be more than happy to look into a proper fix. 

I'm running OP-TEE 4.2.0 with customization.

### DIFF
--- a/ta/pkcs11/src/entry.c
+++ b/ta/pkcs11/src/entry.c
@@ -17,7 +17,13 @@
 
 TEE_Result TA_CreateEntryPoint(void)
 {
-	return pkcs11_init();
+	TEE_Result ret;
+
+	ret = pkcs11_init();
+	if (ret)
+		TEE_Panic(0);
+
+	return TEE_SUCCESS;
 }
 
 void TA_DestroyEntryPoint(void)


jenswikl:
> What I'm observing is that the first client fails as expected (the session won't open) if initialization fails,

That's what I expect, and I can follow in the code.

> but for the next client the initialization is always skipped, and in turn it ends up causing an abort as it tries to access uninitialized memory.

I can't quite see that behaviour from inspecting the code. From what I gather, `tee_ta_open_session()` calls `release_ta_ctx()` on panic. Perhaps I'm missing something.

johankor:
> > What I'm observing is that the first client fails as expected (the session won't open) if initialization fails,
> 
> That's what I expect, and I can follow in the code.
> 
> > but for the next client the initialization is always skipped, and in turn it ends up causing an abort as it tries to access uninitialized memory.
> 
> I can't quite see that behaviour from inspecting the code. From what I gather, `tee_ta_open_session()` calls `release_ta_ctx()` on panic. Perhaps I'm missing something.

The link from `core/kernel/tee_ta_manager` and `lib/libutee/user_ta_entry` is not completely clear to me (looking into it), so it might be that I'm missing something there.

In case my description of the problem was unclear: if the TA panics during `TA_CreateEntryPoint`, both my understanding and observation are consistent with what you write. The problem arises once the TA fails without panicking, but instead returns an error code.

jenswikl:
> In case my description of the problem was unclear: if the TA panics during `TA_CreateEntryPoint`, both my understanding and observation are consistent with what you write. The problem arises once the TA fails without panicking, but instead returns an error code.

Thanks, now I understand. I believe that the current behavior is incorrect. The spec states that if `TA_CreateEntryPoint()` returns any value other than `TEE_SUCCESS`:
_If any other code is returned, then the instance is not created, and no other entry
points of this instance will be called. The Framework SHALL reclaim all resources and dereference all
objects related to the creation of the instance._

This should also be valid for a keep-alive TA, since it hasn't been successfully created yet.

etienne-lms:
A fix for this issue? in **lib/libutee/user_ta_entry.c**:
```patch
 static TEE_Result ta_header_add_session(uint32_t session_id)
 {
 	struct ta_session *itr = ta_header_get_session(session_id);
 	TEE_Result res;
 
 	if (itr)
 		return TEE_SUCCESS;
 
 	if (!init_done) {
-		init_done = true;
 		res = init_instance();
 		if (res)
 			return res;
+		init_done = true;
 	}
 
 	itr = TEE_Malloc(sizeof(struct ta_session),
 			TEE_USER_MEM_HINT_NO_FILL_ZERO);
 	if (!itr)
 		return TEE_ERROR_OUT_OF_MEMORY;
 	itr->session_id = session_id;
 	itr->session_ctx = 0;
 	TAILQ_INSERT_TAIL(&ta_sessions, itr, link);
 
 	return TEE_SUCCESS;
 }

**Comment by johankor** [comment]
@etienne-lms : Yes, I think I've actually done that in one of my local branches 🙂 I can give it another test on my system here. If you think that's an OK solution, I can update this PR accordingly.

**Comment by etienne-lms** [comment]
Yes, please, feel free to create a P-R.
Since it's a fix, it would be nice it's merged before next release tag, if maintainers agree to, and the fix is fully safe.
Whatever, the best is often, when you have a change proposal, to create a P-R to share it and get feedback from review comments. If the fix is not the right one, reviewer will likely help you to each a valid implementation.

**Comment by jenswikl** [comment]
That fix may take care of the problem with your TA @johankor, but I'm not sure it's 100% aligned with the spec I cited above.

**Comment by etienne-lms** [comment]
The fact is core currently does not check if the call is a first entry in the TA (to call `TA_CreateEntryPoint()`). It is delegated to the userspace instance. For non-keepalive TAs, if the 1st session to TA fails, the session, instance etc.. are released. At least should be.
For keepalive TAs, since instances are not destroyed we have the instance created but with `TA_CreateEntryPoint()` not successfully executed. Do we want to allow new creation attempt if one fails? The seems ok with that, it spec says:
> The exact moment when a keep-alive single instance is
created is implementation-defined but it SHALL be no later
than the first session opening.

We *only* have to ensure `TA_CreateEntryPoint()` succeeds before any other entry points is reached. It seems to me the proposed change ensures that. I may miss something...

It looks like we need some extra basic regression tests in xtest for these cases.

**Comment by jenswikl** [comment]
OK, let's try this then.

**Comment by johankor** [comment]
I'll need to have a look at this tomorrow. I tested with the patch above (moving `init_done = true;`). While it no longer aborted, the error below combined with another instance of an indefinite hang didn't inspire confidence.
```
E/TA:  assertion 'b->bh.bsize < 0' failed at <snip>lib/libutils/isoc/bget.c:965 in brel()
```

**Comment by etienne-lms** [comment]
> `E/TA:  assertion 'b->bh.bsize < 0' failed at <snip>lib/libutils/isoc/bget.c:965 in brel()`

That's TA heap corruption. Looks weird. 
Which TA showed this behavior?
I'll have to test that change...

**Comment by johankor** [comment]
> That's TA heap corruption. Looks weird. Which TA showed this behavior? I'll have to test that change...

This is the pkcs11 TA (with some customization), running on OP-TEE 3.21(!). I'll see if I can pin down exactly what happens today. I could try to move to OP-TEE 4.6 if that makes more sense, before continuing the debugging.

**Comment by etienne-lms** [comment]
A nice thing would be to share some code or patch to reproduce the issue on a qemu platform.

**Comment by johankor** [comment]
@etienne-lms : I understand, I'll see if I can prepare that. Meanwhile, maybe the following is also of interest:

The symbolized stack trace:
```
./scripts/symbolize.py -d ../build/ta/pkcs11/fd02c9da-306c-48c7-a49c-bbd827ae86ee.elf
E/TC:? 0 TA panicked with code 0xffff0000 (TEE_ERROR_GENERIC)
E/LD:  Status of TA fd02c9da-306c-48c7-a49c-bbd827ae86ee
E/LD:   arch: aarch64
E/LD:  region  0: va 0xc0004000 pa 0x96233000 size 0x002000 flags rw-s (ldelf)
E/LD:  region  1: va 0xc0006000 pa 0x96235000 size 0x008000 flags r-xs (ldelf)
E/LD:  region  2: va 0xc000e000 pa 0x9623d000 size 0x001000 flags rw-s (ldelf)
E/LD:  region  3: va 0xc000f000 pa 0x9623e000 size 0x004000 flags rw-s (ldelf)
E/LD:  region  4: va 0xc0013000 pa 0x96242000 size 0x001000 flags r--s
E/LD:  region  5: va 0xc0014000 pa 0x962aa000 size 0x002000 flags rw-s (stack)
E/LD:  region  6: va 0xc0055000 pa 0x00001000 size 0x055000 flags r-xs [0] .ta_head .text .eh_frame .rodata .gnu.hash .dynsym .dynstr .hash .rela.dyn
E/LD:  region  7: va 0xc00aa000 pa 0x00056000 size 0x012000 flags rw-s [0] .dynamic .got .rela.data .rela.got .data .bss
E/LD:   [0] fd02c9da-306c-48c7-a49c-bbd827ae86ee @ 0xc0055000 (/.../optee-os/3.21.0.imx/build/ta/pkcs11/fd02c9da-306c-48c7-a49c-bbd827ae86ee.elf)
E/LD:  Call stack:
E/LD:   0xc0067fd4 _assert_break at /usr/src/debug/optee-os/3.21.0.imx/lib/libutee/assert.c:24
E/LD:   0xc008eee8 _assert_trap at /usr/src/debug/optee-os/3.21.0.imx/lib/libutils/isoc/include/assert.h:19
E/LD:   0xc008f298 brel at /usr/src/debug/optee-os/3.21.0.imx/lib/libutils/isoc/bget.c:932 (discriminator 1)
E/LD:   0xc008fab4 raw_realloc at /usr/src/debug/optee-os/3.21.0.imx/lib/libutils/isoc/bget_malloc.c:519
E/LD:   0xc008fc24 realloc_unlocked at /usr/src/debug/optee-os/3.21.0.imx/lib/libutils/isoc/bget_malloc.c:816
E/LD:   0xc0070e34 init_instance at /usr/src/debug/optee-os/3.21.0.imx/lib/libutee/user_ta_entry.c:172
E/LD:   0xc0066b20 __ta_entry at /usr/src/debug/optee-os/3.21.0.imx/export-ta_arm64/src/user_ta_header.c:70
```

So, in addition to moving the `init_done = true` as discussed, with the patch below it now works for me. There's another `raw_malloc_add_pool` for `__ta_no_share_heap_size`, not sure if that also needs to be dealt with.
```diff
diff --git i/lib/libutee/user_ta_entry.c w/lib/libutee/user_ta_entry.c
index dca489142..c89d84285 100644
--- i/lib/libutee/user_ta_entry.c
+++ w/lib/libutee/user_ta_entry.c
@@ -167,9 +167,16 @@ static unsigned int get_memtag_implementation(void)

 static TEE_Result init_instance(void)
 {
+       static bool malloc_inited;
+
        trace_set_level(tahead_get_trace_level());
        __utee_gprof_init();
-       malloc_add_pool(ta_heap, ta_heap_size);
+
+       if (!malloc_inited) {
+               malloc_add_pool(ta_heap, ta_heap_size);
+               malloc_inited = true;
+       }
+
        if (__ta_no_share_heap_size) {
                __ta_no_share_malloc_ctx = malloc(raw_malloc_get_ctx_size());
                if (__ta_no_share_malloc_ctx) {

jenswikl:
Everything up to the call to `TA_CreateEntryPoint()` will either succeed or panic the TA, so that entire block should be guarded to only execute once.

johankor:
Closed in favor of #7498

PR #7444, opened 2025-06-23
core: ree_fs: initialize ta_ver.db in one operation

gavin-liu106:
Combined the object creation and data writing operations into a single step to enhance reliability. This change addresses the situation where, if object creation occurs but the data writing fails, an empty object would be left behind, leading to potential issues during the next boot.

Fixes: OP-TEE#7438

### DIFF
--- a/core/kernel/ree_fs_ta.c
+++ b/core/kernel/ree_fs_ta.c
@@ -111,12 +111,8 @@ static TEE_Result check_update_version(const char *db_name,
 		goto out;
 
 	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
-		res = ops->create(&pobj, false, NULL, 0, NULL, 0, NULL, NULL,
-				   0, &fh);
-		if (res != TEE_SUCCESS)
-			goto out;
-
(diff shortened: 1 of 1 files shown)


jenswikl:
If possible, please add a Fixes tag for the commit introducing the error.

gavin-liu106:
Hi,
Do you mean adding a "Fixes: #xxxx" tag? If so, I have already included it in the commit message.
Please let me know if you would like me to reference a specific commit instead. Thank you!

jenswikl:
I meant a reference to the commit that introduced the problem. In this case it seems to be
`Fixes: 183398139c9c ("core: enable rollback protection for REE-FS TAs")`

jenswikl:
> I think it's still worth to add a link to the Github issue being fixed. Maybe with a `Link:` tag:

Absolutely. Or another Fixes tag. We can have as many as we like.

With that fixed, please apply:

gavin-liu106:
Thank you for the review.
I updated the commit message.

etienne-lms (approved):
All good to me.

PR #7449, opened 2025-06-24
plat-versal2: conf: Add maximum size of the DTB

Akshay-Belsare:
The DTB size for the AMD platform is larger and does not fit into the default size, leading to failure or panic at boot time due to size issues.

Thus setting an explicit maximum size for the Device Tree Blob to allow safe modifications. This ensures there is enough space when appending or editing nodes/properties in the DTB.


jenswikl:

Akshay-Belsare:
Thanks @jenswi-linaro!
Added my acked-by also in the commit message.
@jforissier tags are updated in the commit message.

PR #7451, opened 2025-06-25
Allow testing with Hafnium timer

jenswi-linaro:


GseoC:
For commits "core: arm64.h: add {read,write}_cntp_{ct,tva,cva}l()" and "core: arm: support EL1 physical timer":

jenswikl:
Added Hafnium output and applied tags.

PR #7452, opened 2025-06-26
core: imx: add generic boards

saschahauer:
For i.MX we have to specify the exact board we want to build. Most of the time a board support merely consists of setting CFG_UART_BASE and CFG_DDR_SIZE for that board. These variables could be specified from the outside by the build system as well.

Instead of forcing the user to patch OP-TEE for each new board allow to build optee generically for a i.MX SoC by adding the SoC name to the flavorlists. For example an OP-TEE binary for i.MX6UL can then be built with

make PLATFORM=imx-mx6ul CFG_UART_BASE=UART1_BASE CFG_DDR_SIZE=0x10000000

### DIFF
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -6,10 +6,12 @@ mx6ul-flavorlist = \
 	mx6ul9x9evk \
 	mx6ulccimx6ulsbcpro \
 	mx6ulccbv2 \
+	mx6ul
 
 mx6ull-flavorlist = \
 	mx6ullevk \
 	mx6ulzevk \
+	mx6ull
 
 mx6q-flavorlist = \
 	mx6qsabrelite \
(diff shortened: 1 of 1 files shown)


jforissier:
Looks like a good idea. I'll leave it to the imx maintainers but FWIW:

sahilnxp:
@saschahauer Thanks for the pull request.
Overall, it looks like a good idea to add generic boards, but it will increase the chances of error if all the CONFIG flags are not correctly passed.

Also, there are multiple versions of i.MX SoCs like we take example for i.MX6UL only as shown in the image below:
image

We don't support OP-TEE on any i.MX6UL platform apart from ones supported in [core/arch/arm/plat-imx/conf.mk]
(https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-imx/conf.mk#L4)

This pull request will drastically increase the scope of i.MX platform on which OP-TEE is supported and not tested.

saschahauer:
> @saschahauer Thanks for the pull request. Overall, it looks like a good idea to add generic boards,

Indeed ;)

> but it will increase the chances of error if all the CONFIG flags are not correctly passed.

You can use the board specific flavors then.

> 
> Also, there are multiple versions of i.MX SoCs like we take example for i.MX6UL only as shown in the image below: image
> 
> We don't support OP-TEE on any i.MX6UL platform apart from ones supported in [core/arch/arm/plat-imx/conf.mk] (https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-imx/conf.mk#L4)

The currently supported boards are mostly eval boards only. These are not the boards people outside actually use except for evaluation purposes and early prototyping. So people do not get support for their boards anyway (except maybe they pay NXP for it). Adding SoC specific flavors changes nothing in that situation, but it makes it easier for people to adopt OP-TEE for their board. Right now people either carry an unnecessary board patch or they just pick a board from the compatible list that seems similar enough, both possibilities do not seem very fortunate.

> 
> This pull request will drastically increase the scope of i.MX platform on which OP-TEE is supported and not tested.

You have this situation already, you just don't see it because people do not care to send patches for their custom boards.

sahilnxp:
@saschahauer If we go this way, we need to put in documentation somewhere about config flags which we need to enable for one board like if we compile using make PLATFORM=imx-mx6ul, It fails with following compilation error:

> In file included from core/arch/arm/include/mm/core_mmu_arch.h:20,
>                  from core/include/mm/core_mmu.h:19,
>                  from core/include/kernel/thread_private.h:16,
>                  from core/arch/arm/kernel/asm-defines.c:9:
> core/arch/arm/plat-imx/./platform_config.h:39:2: error: #error "CFG_DDR_SIZE not defined"
>    39 | #error "CFG_DDR_SIZE not defined"
>       |  ^~~~~

Another example like for i.MX8DXL

`make PLATFORM=imx-mx8dxl` results in error below:

> core/arch/arm/plat-imx/./platform_config.h:39:2: error: #error "CFG_DDR_SIZE not defined"
>    39 | #error "CFG_DDR_SIZE not defined"
>       |  ^~~~~

`make PLATFORM=imx-mx8dxl CFG_DDR_SIZE=0x40000000` gives following error

> In file included from core/include/mm/core_memprot.h:9,
>                  from core/include/kernel/interrupt.h:10,
>                  from core/include/drivers/gic.h:10,
>                  from core/arch/arm/plat-imx/main.c:33:
> core/arch/arm/plat-imx/./platform_config.h:43:42: error: ‘CFG_UART_BASE’ undeclared here (not in a function); did you mean ‘CFG_DRAM_BASE’?
>    43 | #define CONSOLE_UART_BASE               (CFG_UART_BASE)
>       |                                          ^~~~~~~~~~~~~
> core/include/mm/core_mmu.h:194:62: note: in definition of macro ‘__register_memory’
>   194 |                 { .name = (_name), .type = (_type), .addr = (_addr), \
>       |                                                              ^~~~~
> core/include/mm/core_mmu.h:217:40: note: in expansion of macro ‘ROUNDDOWN’
>   217 |         __register_memory(#addr, type, ROUNDDOWN(addr, CORE_MMU_PGDIR_SIZE), \
>       |                                        ^~~~~~~~~
> core/arch/arm/plat-imx/main.c:45:1: note: in expansion of macro ‘register_phys_mem_pgdir’
>    45 | register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE,
>       | ^~~~~~~~~~~~~~~~~~~~~~~
> core/arch/arm/plat-imx/main.c:45:43: note: in expansion of macro ‘CONSOLE_UART_BASE’
>    45 | register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE,
>       |                                           ^~~~~~~~~~~~~~~~~
> cc1: note: unrecognized command-line option ‘-Wno-c2x-extensions’ may have been intended to silence earlier diagnostics
> make: *** [mk/compile.mk:166: out/arm-plat-imx/core/arch/arm/plat-imx/main.o] Error 1

When user executes `make PLATFORM=<platform>`, we should print a message on screen telling user, that `<platform>` require `CFG_DDR_SIZE`, `CFG_UART_BASE` etc config flags (which are mandatory) to be given instead of waiting for compilation error.

Also, we need to mention that this support is just for making the compilation easier and after this, need not to patch OP-TEE for each variant of SoC. In addition to have this ease, there is potential disclaimer that needs to be published in documentation that:
Every board variant of an SoC is not tested with OP-TEE. Only boards specific flavor which are configured in conf.mk are supported.

PR #7453, opened 2025-06-26
i.MX6/8M OCOTP write support

bith3ad:
This adds the OCOTP write support for i.MX8M and i.MX6 devices.

### DIFF
--- a/core/drivers/imx_ocotp.c
+++ b/core/drivers/imx_ocotp.c
@@ -11,14 +11,20 @@
 #include <drivers/imx_ocotp.h>
 #include <kernel/tee_common_otp.h>
 
-#define OCOTP_CTRL		    0x0
-#define OCOTP_CTRL_ERROR	    BIT32(9)
-#define OCOTP_CTRL_BUSY		    BIT32(8)
+#define OCOTP_CTRL			0x0
+
+#if defined(CFG_MX8MP)
+#define OCOTP_CTRL_ERROR		BIT32(10)
+#define OCOTP_CTRL_BUSY			BIT32(9)
+#else
+#define OCOTP_CTRL_ERROR		BIT32(9)
+#define OCOTP_CTRL_BUSY			BIT32(8)
+#endif
 
 #if defined(CFG_MX6) || defined(CFG_MX7ULP)
-#define OCOTP_SHADOW_OFFSET(_b, _w) ((_b) * (0x80) + (_w) * (0x10) + 0x400)
+#define OCOTP_SHADOW_OFFSET(_b, _w)	((_b) * (0x80) + (_w) * (0x10) + 0x400)
 #else
-#define OCOTP_SHADOW_OFFSET(_b, _w) ((_b) * (0x40) + (_w) * (0x10) + 0x400)
+#define OCOTP_SHADOW_OFFSET(_b, _w)	((_b) * (0x40) + (_w) * (0x10) + 0x400)
 #endif
 
 struct ocotp_instance {
--- a/core/drivers/imx_ocotp.c
+++ b/core/drivers/imx_ocotp.c
@@ -12,6 +12,7 @@
 #include <kernel/tee_common_otp.h>
 
 #define OCOTP_CTRL			0x0
+#define OCOTP_CTRL_CLR			0x8
 
 #if defined(CFG_MX8MP)
 #define OCOTP_CTRL_ERROR		BIT32(10)
@@ -103,7 +104,7 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
 	ocotp_clock_enable();
 
 	/* Clear error bit */
-	io_clrbits32(g_base_addr + OCOTP_CTRL, OCOTP_CTRL_ERROR);
+	io_write32(g_base_addr + OCOTP_CTRL_CLR, OCOTP_CTRL_ERROR);
 
 	/* Wait for busy flag to be cleared */
 	ret = ocotp_ctrl_wait_for(OCOTP_CTRL_BUSY);
--- a/core/drivers/imx_ocotp.c
+++ b/core/drivers/imx_ocotp.c
@@ -9,18 +9,35 @@
 #include <imx.h>
 #include <io.h>
 #include <drivers/imx_ocotp.h>
+#include <kernel/delay.h>
 #include <kernel/delay_arch.h>
 #include <kernel/tee_common_otp.h>
 
 #define OCOTP_CTRL			0x0
+#define OCOTP_CTRL_SET			0x4
 #define OCOTP_CTRL_CLR			0x8
+#define OCOTP_TIMING			0x10
+#define OCOTP_DATA			0x20
+
+#define OCOTP_CTRL_WR_UNLOCK_KEY	0x3E77
+
+#define OCOTP_TIMING_WAIT		GENMASK_32(27, 22)
+#define OCOTP_TIMING_STROBE_READ	GENMASK_32(21, 16)
+#define OCOTP_TIMING_RELAX		GENMASK_32(15, 12)
+#define OCOTP_TIMING_STROBE_PROG	GENMASK_32(11, 0)
 
 #if defined(CFG_MX8MP)
+#define OCOTP_CTRL_WR_UNLOCK		GENMASK_32(31, 16)
+#define OCOTP_CTRL_RELOAD_SHADOWS	BIT(11)
 #define OCOTP_CTRL_ERROR		BIT32(10)
 #define OCOTP_CTRL_BUSY			BIT32(9)
+#define OCOTP_CTRL_ADDR			GENMASK_32(8, 0)
 #else
+#define OCOTP_CTRL_WR_UNLOCK		GENMASK_32(31, 16)
+#define OCOTP_CTRL_RELOAD_SHADOWS	BIT(10)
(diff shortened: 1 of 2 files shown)


core/drivers/imx_ocotp.c
```diff
@@ -68,6 +69,21 @@ static inline void ocotp_clock_enable(void) { }
 #error "Platform not supported"
 #endif
 
+#if defined(CFG_MX8M)
```
sahilnxp, line 108: This check needed only for i.MX8M platforms? why not for other platforms ?

bith3ad: What platform do you have in mind?

i.MX8M (ARMv8-A) are the only ones which support the ARM generic timers. As written in the commit message of:  https://github.com/OP-TEE/optee_os/pull/7453/commits/d2664c6ba21be39214fb89bbb2d945748f5e358a this is an improvement to avoid assumptions. IIRC, The i.MX9* series uses another driver since the OCOTP is behind the ELE. All other i.MX8 SoCs can be added on demand by someone who can test it, I don't have them on my desk.

sahilnxp: I compiled i.MX6UL, i.MX7DSABRESD and i.MX8MP platform.
Each platform has the `CFG_CORE_HAS_GENERIC_TIMER=y`

`timeout_init_us()` is internaly calling `delay_cnt_read()` -> `barrier_read_counter_timer()` -> `read_cntpct()`
`read_cntpct()` will be defined in `arm32_sysreg.S` file generated during compilation.

So, this means these platforms also support ARM generic timer.

Please correct me if I am wrong.

core/drivers/imx_ocotp.c
```diff
@@ -76,7 +108,8 @@ static TEE_Result ocotp_ctrl_wait_for(uint32_t mask)
 
 	assert(g_base_addr);
 
-	timeout = timeout_init_us(20);
+	/* Shadow reload needs more time */
+	timeout = timeout_init_us(1000);
```
sahilnxp, line 116: Same thing needs to be done for platforms other than i.MX8M ?

core/drivers/imx_ocotp.c
```diff
@@ -140,6 +173,116 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
...
+	if (!val)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (bank > g_ocotp->nb_banks || word > g_ocotp->nb_words)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!g_ocotp->write_fuse)
+		return TEE_ERROR_NOT_IMPLEMENTED;
+
+	assert(g_base_addr && g_ocotp);
```
sahilnxp, line 193: This should be checked before accessing `g_ocotp` fileds.

core/drivers/imx_ocotp.c
```diff
@@ -140,6 +173,116 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
...
+	timing |= field_prep(OCOTP_TIMING_STROBE_PROG, strobe_prog);
+
+	io_write32(g_base_addr + OCOTP_TIMING, timing);
+
+	return ocotp_ctrl_wait_for(OCOTP_CTRL_BUSY);
+}
+
+static TEE_Result ocotp_mx8m_write_fuse(unsigned int bank, unsigned int word, uint32_t val)
+{
+	TEE_Result ret;
```
sahilnxp, line 255: Initialize `ret `with default value.

core/drivers/imx_ocotp.c
```diff
@@ -9,18 +9,35 @@
...
+
+#define OCOTP_CTRL_WR_UNLOCK_KEY	0x3E77
+
+#define OCOTP_TIMING_WAIT		GENMASK_32(27, 22)
+#define OCOTP_TIMING_STROBE_READ	GENMASK_32(21, 16)
+#define OCOTP_TIMING_RELAX		GENMASK_32(15, 12)
+#define OCOTP_TIMING_STROBE_PROG	GENMASK_32(11, 0)
 
 #if defined(CFG_MX8MP)
+#define OCOTP_CTRL_WR_UNLOCK		GENMASK_32(31, 16)
```
sahilnxp, line 30: Can keep this macro outside of `#if defined`

core/drivers/imx_ocotp.c
```diff
@@ -140,6 +173,116 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
...
+	io_write32(g_base_addr + OCOTP_CTRL_CLR, OCOTP_CTRL_ERROR);
+
+	/* Wait for busy flag to be cleared */
+	ret = ocotp_ctrl_wait_for(OCOTP_CTRL_BUSY);
+	if (ret) {
+		EMSG("OCOTP is busy");
+		goto out;
+	}
+
+	ret = g_ocotp->write_fuse(bank, word, val);
```
sahilnxp, line 209: Please add error check on `ret`.

core/drivers/imx_ocotp.c
Emantor: I can confirm that only the Cortex-A9 devices are missing the generic timers, every other supported i.MX SoC should have them.

sahilnxp (changes requested):

core/drivers/imx_ocotp.c
sahilnxp: Thanks @Emantor, so this check needs to include other platforms also.

core/drivers/imx_ocotp.c
```diff
@@ -140,6 +173,116 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
...
+	DMSG("OCOTP Bank %d Word %d Fuse 0x%" PRIx32, bank, word, val);
+out:
+	mutex_unlock(&fuse_read);
+
+	return ret;
+}
+
+static TEE_Result ocotp_mx8m_set_timing(void)
+{
+	uint32_t relax, strobe_read, strobe_prog;
```
sahilnxp, line 229: Initialize these variables.

core/drivers/imx_ocotp.c
```diff
@@ -140,6 +173,116 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
...
+	reg |= field_prep(OCOTP_CTRL_ADDR, OCOTP_ADDR(bank, word));
+	reg |= field_prep(OCOTP_CTRL_WR_UNLOCK, OCOTP_CTRL_WR_UNLOCK_KEY);
+	io_write32(g_base_addr + OCOTP_CTRL, reg);
+
+	/* Clear error bit */
+	io_write32(g_base_addr + OCOTP_CTRL_CLR, OCOTP_CTRL_ERROR);
+
+	io_write32(g_base_addr + OCOTP_DATA, val);
+	ret = ocotp_ctrl_wait_for(OCOTP_CTRL_BUSY);
+	if (ret)
```
sahilnxp, line 276: Please put a debug print here.

core/drivers/imx_ocotp.c
```diff
@@ -140,6 +173,116 @@ TEE_Result imx_ocotp_read(unsigned int bank, unsigned int word, uint32_t *val)
...
+		return ret;
+
+	/* Write postamble */
+	udelay(2000);
```
sahilnxp, line 280: image
According to above image, We will be writing maximum of 32bits in one go and In that case, I think 400us is enough here.

sahilnxp (changes requested):
Please let me know on which platform you have tested this and which fuse you tried writing ?

core/drivers/imx_ocotp.c
bith3ad: Sure, albeit one could argue that the whole `OCOTP_CTRL_` definitions are separated by the `#ifdef` which is valid too. However, I can change it if you want.

core/drivers/imx_ocotp.c
bith3ad: I see, my assumption was that archticted-timers are an ARMv8 feature. However, I will include the Cortex-A7 based SoCs as well.

core/drivers/imx_ocotp.c
bith3ad: Same timeout should be added yes. Should be the same IP, albeit we already tested the fusing on i.MX6 (Cortex-A9) SoCs successfully.
However, we should set it to 1000us to be safe.

core/drivers/imx_ocotp.c
bith3ad: Sure

core/drivers/imx_ocotp.c
bith3ad: May I ask why, they are set later on anyways? Also why just these 3 variables and not `clk_rate` and `timing`?

core/drivers/imx_ocotp.c
bith3ad: Will be assigned 3 lines below. Why only `ret` and not `reg`, is there a guideline which vars have to be initialized?

core/drivers/imx_ocotp.c
bith3ad: You're right, waiting 2ms is far to long. You do quote the time we should wait for the busy flag to clear, this udelay targets the postamble which is 2us according the TRM.

PR #7455, opened 2025-06-26
i.MX TZASC improvements

bith3ad:
This PR is to hardening the TZASC configuration on i.MX SoCs.

First patch fixes the missing TZASC setup for i.MX6QP platforms.

Patch2 adds the support to dump the very important region0 which covers the complete SoC AXI space incl. memory alias addresses.

Patches 3-4 add the support to verify that the prev. running firmware doesn't re-configure region0 to secure + non-secure RW access. For more information see: https://lore.kernel.org/barebox/20250626144527.416697-1-m.felsch@pengutronix.de/T/#m9efb0988a0b594ba23fc8622671f2828c07a1e0f

Patch 5 ensures that the TZASC is enabled.

### DIFF
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -525,6 +525,12 @@ ifneq (,$(filter y, $(CFG_MX6) $(CFG_MX7)))
 CFG_IMX_CSU ?= y
 endif
 
+ifneq (,$(filter y, $(CFG_MX8M)))
+ifneq ($(CFG_INSECURE),y)
+$(call force,CFG_TZASC_REGION0_SECURE,y)
+endif
+endif
+
 ifeq ($(filter y, $(CFG_PSCI_ARM32)), y)
 CFG_HWSUPP_MEM_PERM_WXN = n
 CFG_IMX_WDOG ?= y
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -531,6 +531,15 @@ $(call force,CFG_TZASC_REGION0_SECURE,y)
 endif
 endif
 
+# We can't make it default since due to missing i.MX7 and i.MX9 support.
+# Once all plattforms are supported we can remove the CFG_TZASC_CHECK_ENABLED
+# and perform the check always.
+ifneq (,$(filter y, $(CFG_MX6) $(CFG_MX8M)))
+ifneq ($(CFG_INSECURE),y)
+$(call force,CFG_TZASC_CHECK_ENABLED,y)
+endif
+endif
+
 ifeq ($(filter y, $(CFG_PSCI_ARM32)), y)
 CFG_HWSUPP_MEM_PERM_WXN = n
 CFG_IMX_WDOG ?= y
--- a/core/arch/arm/plat-imx/tzc380.c
+++ b/core/arch/arm/plat-imx/tzc380.c
@@ -67,6 +67,20 @@ static TEE_Result imx_configure_tzasc(void)
 
 		tzc_init(addr[i]);
 
+		/*
+		 * TZC380 is not memory alias aware so an attacker could read
+		 * the OP-TEE core memory if the system does support memory
+		 * aliasing.
+		 *
+		 * To fix this region0 needs to be configured as secure access
+		 * only (0xc). This is the default if not changed by the
+		 * previous running firmware. Region0 covers the complete
+		 * platform AXI address space.
+		 */
+		if (IS_ENABLED(CFG_TZASC_REGION0_SECURE) &&
+		    tzc_verify_region0_secure() != TEE_SUCCESS)
+			panic("region0 is not secure configured, non-secure memory alias access possible!");
+
 		region = imx_tzc_auto_configure(CFG_DRAM_BASE, CFG_DDR_SIZE,
 						TZC_ATTR_SP_NS_RW, region);
 		region = imx_tzc_auto_configure(CFG_TZDRAM_START,
--- a/core/arch/arm/plat-imx/tzc380.c
+++ b/core/arch/arm/plat-imx/tzc380.c
@@ -11,6 +11,7 @@
 #include <drivers/tzc380.h>
 #include <imx-regs.h>
 #include <initcall.h>
+#include <io.h>
 #include <kernel/panic.h>
 #include <kernel/pm.h>
 #include <mm/core_memprot.h>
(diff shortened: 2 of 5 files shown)


core/drivers/tzc380.c
```diff
@@ -338,7 +338,9 @@ void tzc_dump_state(void)
 	     io_read32(tzc.base + SECURITY_INV_EN_OFF));
 	for (n = 0; n <= REGION_MAX; n++) {
 		temp_32reg = tzc_read_region_attributes(tzc.base, n);
-		if (!(temp_32reg & TZC_ATTR_REGION_EN_MASK))
+
+		/* region0 is always enabled */
+		if (n && !(temp_32reg & TZC_ATTR_REGION_EN_MASK))
```
sahilnxp, line 366: If region0 is always enabled, that region will be dumped.
I don't think we need a special check here.
But keeping the comment `/* region0 is always enabled */` is fine.

core/drivers/tzc380.c
```diff
@@ -348,6 +350,9 @@ void tzc_dump_state(void)
 		DMSG("region_base: 0x%08x%08x", temp_32reg_h, temp_32reg);
 		temp_32reg = tzc_read_region_attributes(tzc.base, n);
 		DMSG("region sp: %x", temp_32reg >> TZC_ATTR_SP_SHIFT);
+		/* Skip printing the size for region0 */
```
sahilnxp, line 376: Why need to skip printing the size?

core/drivers/tzc380.c
```diff
@@ -315,6 +315,29 @@ TEE_Result tzc_regions_lockdown(void)
 	return TEE_SUCCESS;
 }
 
+/*
+ * `tzc_verify_region0_secure` verifies that default region0 only permitts
```
sahilnxp, line 319: `permits `in place of `permitts`

core/drivers/tzc380.c
```diff
@@ -315,6 +315,29 @@ TEE_Result tzc_regions_lockdown(void)
...
+ * secure-world access to overcome memory alias access breaches.
+ *
+ * Return
+ *  - TEE_SUCCESS if region0 is secure world only access
+ *  - TEE_ERROR_SECURITY if region0 can be accessed be non-secure world
```
sahilnxp, line 324: if region0 can be accessed `by` non-secure world

core/arch/arm/plat-imx/tzc380.c
```diff
@@ -67,6 +67,20 @@ static TEE_Result imx_configure_tzasc(void)
...
 		tzc_init(addr[i]);
 
+		/*
+		 * TZC380 is not memory alias aware so an attacker could read
+		 * the OP-TEE core memory if the system does support memory
+		 * aliasing.
+		 *
+		 * To fix this region0 needs to be configured as secure access
+		 * only (0xc). This is the default if not changed by the
+		 * previous running firmware. Region0 covers the complete
```
sahilnxp, line 174: Minor: `region0 `in place of `Region0 `to maintain consistency

core/arch/arm/plat-imx/conf.mk
```diff
@@ -525,6 +525,12 @@ ifneq (,$(filter y, $(CFG_MX6) $(CFG_MX7)))
 CFG_IMX_CSU ?= y
 endif
 
+ifneq (,$(filter y, $(CFG_MX8M)))
```
sahilnxp, line 528: Since there are many platforms which are using TZC380, don't we need the same thing for other platforms ?

core/arch/arm/plat-imx/conf.mk
```diff
@@ -531,6 +531,15 @@ $(call force,CFG_TZASC_REGION0_SECURE,y)
 endif
 endif
 
+# We can't make it default since due to missing i.MX7 and i.MX9 support.
+# Once all plattforms are supported we can remove the CFG_TZASC_CHECK_ENABLED
```
sahilnxp, line 535: Minor: Typo `platforms` in place of `plattforms`

core/arch/arm/plat-imx/tzc380.c
```diff
@@ -29,6 +30,99 @@ register_phys_mem(MEM_AREA_IO_SEC, TZASC2_BASE, TZASC_SIZE);
...
+
+	mask = IMX8M_LOCK_GPR_TZASC_EN | IMX8M_TZASC_ID_SWAP_BYPASS |
+	       IMX8M_TZASC_EN;
+
+	return (io_read32(addr + IOMUXC_GPR_GPR10_OFFSET) & mask) == mask;
+}
+
+static bool imx_tzasc_is_enabled(void)
+{
+
```
sahilnxp, line 111: Please remove blank space.

core/arch/arm/plat-imx/tzc380.c
```diff
@@ -29,6 +30,99 @@ register_phys_mem(MEM_AREA_IO_SEC, TZASC2_BASE, TZASC_SIZE);
...
+#ifndef IOMUXC_GPR_GPR10_OFFSET
+#define IOMUXC_GPR_GPR10_OFFSET		0
+#endif
+
+#define IMX6_TZASC2_BYP			BIT(1)
+#define IMX6_TZASC1_BYP			BIT(0)
+
+#define IMX8M_LOCK_GPR_TZASC_EN		BIT(16)
+#define IMX8M_TZASC_ID_SWAP_BYPASS	BIT(1)
+#define IMX8M_TZASC_EN			BIT(1)
```
sahilnxp, line 60: This should be BIT32(0)

core/arch/arm/plat-imx/tzc380.c
```diff
@@ -29,6 +30,99 @@ register_phys_mem(MEM_AREA_IO_SEC, TZASC2_BASE, TZASC_SIZE);
...
+/* Not all platforms support the GPR offsets yet */
+#ifndef IOMUXC_GPR9_OFFSET
+#define IOMUXC_GPR9_OFFSET		0
+#endif
+
+#ifndef IOMUXC_GPR_GPR10_OFFSET
+#define IOMUXC_GPR_GPR10_OFFSET		0
+#endif
+
+#define IMX6_TZASC2_BYP			BIT(1)
```
sahilnxp, line 55: Since these registers are 32 bits, it is better to define these as `BIT32()` instead of `BIT()`

sahilnxp, line 55: Better to define platform specific Macros in registers/imx6.h or registers/imx8m.h

sahilnxp (changes requested):
Some comments for minor typos in commit description of following commits.

For commit: `drivers: imx: tzc380: add CFG_MX6QP TZASC2 configuration`: `plattform` to `platform`
For commit: `drivers: tzc380: add support to dump region0`: `interessting` to `interesting` 
For commit: `drivers: imx: tzc380: add support to check TZASC enable state`: `plattform` to `platform`

core/arch/arm/plat-imx/tzc380.c
```diff
@@ -67,6 +67,20 @@ static TEE_Result imx_configure_tzasc(void)
...
+		 * aliasing.
+		 *
+		 * To fix this region0 needs to be configured as secure access
+		 * only (0xc). This is the default if not changed by the
+		 * previous running firmware. Region0 covers the complete
+		 * platform AXI address space.
+		 */
+		if (IS_ENABLED(CFG_TZASC_REGION0_SECURE) &&
+		    tzc_verify_region0_secure() != TEE_SUCCESS)
+			panic("region0 is not secure configured, non-secure memory alias access possible!");
```
alexhoppus, line 179: Hi,

thanks for this change. I just want to confirm that I get the core problem correctly,

here is what IMX8MPRM says:
> Memory "aliasing" implications on TZASC settings - in systems which does not utilize
the maximal supported DDR space the controller is designed for, the whole DDR
memory map becomes "aliased" (replicated) by the size of the physical memory used.

is that the issue which you want to solve? Thanks.

core/arch/arm/plat-imx/conf.mk
bith3ad: We know that the i.MX8M SoCs are not memory alias safe. i.MX6/7 SoCs are memory alias aware if the DCD configures the MMDC correctly but could also be affected if one DCD table is used to cover multiple RAM configurations.

We can drop this filter and make the check as default=y if INSECURE=n. I just wanted to not cause any regression since we covered that many BL2 firmware set the region0 to non-secure access.

core/arch/arm/plat-imx/tzc380.c
bith3ad: Why? These defines are only useful for this driver.

core/arch/arm/plat-imx/tzc380.c
bith3ad: The main problem:
- OP-TEE configures the OP-TEE OS memory as secure only (only the real memory address, not the aliased address)
- BL2 (pre-optee-os) firmware may have reconfigured region0 to non-secure+secure access.
- This is problematic because now an attacker could perform memory readout attacks via aliased memory addresses, because OP-TEE didn't configured all possible aliased addresses.

The main problem is that the pre-opteee-os firmware has changed the sane default reset value which is secure-only for region0, so that the memory-alias attack can't be performed because each alias address is secure-only.

core/drivers/tzc380.c
bith3ad: Please see the commit message, I want to dump `region0` which is currently not the case because this bitfield is marked as `reserved` for region0.

https://developer.arm.com/documentation/ddi0431/c/programmers-model/register-descriptions/region-attributes--n--register?lang=en

core/drivers/tzc380.c
bith3ad: region0 is always from `0x0 - AXI-BUS-WIDTH` (see: https://developer.arm.com/documentation/ddi0431/c/programmers-model/register-descriptions/region-attributes--n--register?lang=en)

bith3ad:
@sahilnxp thanks for the review and sorry for the late response, I will open a new PR with your feedback included.

PR #7456, opened 2025-06-27
Fix naming of SYSCONFIG for STM32MP2

tbourgoi:
Hello,

During a code review, I've discovered that defines for SYSCONFIG registers for STM32MP2x were not following the name of the registers in the reference manual.
Here is a small PR to align the registers names of SYSCONFIG.
First patch reorder the old defines 
The second patch rename the registers and update clk drivers for STM32MP25 and STM32MP21

Thomas


jforissier:
Please add @etienne-lms' R-b to the second commit. Should this go into 4.7.0?

tbourgoi:
Hello, 
It's not critical for 4.7.0, clock drivers works, this is to prepare upstream of future drivers so we have the correct naming convention.
This PR can wait for 4.8.0.

GseoC:
This P-R is ready to merge, right @tbourgoi ?

tbourgoi:
> This P-R is ready to merge, right @tbourgoi ?

Hello, Yes the P-R can be merged

PR #7459, opened 2025-06-28
core: riscv: add initial support for CVA6 APU on Genesys2

maroueneboubakri:
Add initial support for the CVA6 APU on the Genesys2 board.

This includes two build flavors:

  - `cv64a6_genesys_2`: 64-bit variant, RV64IMAFDC ISA with Sv39 MMU.
  - `cv32a6_genesys_2`: 32-bit variant, RV32IMAC ISA with Sv32 MMU.

Main platform specification:

  - CPU: Single-core OpenHW Group CVA6 CPU
  - ISA: RV64IMAFDC (64-bit) or RV32IMAC (32-bit)
  - MMU: Sv39 (64-bit), Sv32 (32-bit)
  - Timer: APB timer mapped at 0x18000000 with 4 interrupts
  - UART: ns16550 compatible, base 0x10000000
  - CLINT: at 0x2000000 for timer and software interrupts
  - PLIC: at 0xc000000 with 30 devices
  - Memory: 1 GB DDR3 DRAM at 0x80000000

The secure memory region reserved for OP-TEE starts at address 0xBE000000 and is 16MB in size, placed near the top of the 1GB DRAM.

### DIFF
--- a/core/arch/riscv/plat-cva6apu/conf.mk
+++ b/core/arch/riscv/plat-cva6apu/conf.mk
@@ -1,0 +1,53 @@
+PLATFORM_FLAVOR ?= cv64a6_genesys_2
+
+ifeq ($(PLATFORM_FLAVOR),cv64a6_genesys_2)
+$(call force,CFG_RV64_core,y)
+$(call force,CFG_RISCV_FPU,y)
+$(call force,CFG_RISCV_MMU_MODE,39)
+supported-ta-targets = ta_rv64
+endif
+
+ifeq ($(PLATFORM_FLAVOR),cv32a6_genesys_2)
+$(call force,CFG_RV32_core,y)
+$(call force,CFG_RISCV_FPU,n)
+$(call force,CFG_RISCV_MMU_MODE,32)
+supported-ta-targets = ta_rv32
+endif
+
+$(call force,CFG_RISCV_ISA_C,y)
+
+$(call force,CFG_CORE_LARGE_PHYS_ADDR,y)
+$(call force,CFG_CORE_RESERVED_SHM,n)
+$(call force,CFG_CORE_DYN_SHM,y)
+
+$(call force,CFG_DYN_CONFIG,n)
+
+CFG_DT ?= y
+
+$(call force,CFG_WITH_SOFTWARE_PRNG,y)
+
+$(call force,CFG_CORE_ASLR,n)
+$(call force,CFG_CORE_SANITIZE_KADDRESS,n)
+
+$(call force,CFG_TEE_CORE_NB_CORE, 1)
+$(call force,CFG_NUM_THREADS, 1)
+$(call force,CFG_BOOT_SYNC_CPU,n)
+
+CFG_RISCV_PLIC ?= y
+$(call force,CFG_RISCV_APLIC,n)
+$(call force,CFG_RISCV_APLIC_MSI,n)
+$(call force,CFG_RISCV_IMSIC,n)
+
+CFG_RISCV_SBI_CONSOLE ?= n
+CFG_16550_UART ?= y
+
+$(call force,CFG_RISCV_M_MODE,n)
+$(call force,CFG_RISCV_S_MODE,y)
+$(call force,CFG_RISCV_TIME_SOURCE_RDTIME,y)
+CFG_RISCV_MTIME_RATE ?= 10000000
+CFG_RISCV_SBI ?= y
+CFG_RISCV_WITH_M_MODE_SM ?= y
+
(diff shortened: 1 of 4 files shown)


gagachang:
Hi @maroueneboubakri, Just curious about this variant:
`cv32a6_genesys_2: 32-bit variant, RV32IMAC ISA with Sv32 MMU.`
We (RISE) haven't ever tried to run RV32 OP-TEE.
Does current upstream OP-TEE support RV32 and Sv32 MMU ?

maroueneboubakri:
> Hi @maroueneboubakri, Just curious about this variant: `cv32a6_genesys_2: 32-bit variant, RV32IMAC ISA with Sv32 MMU.` We (RISE) haven't ever tried to run RV32 OP-TEE. Does current upstream OP-TEE support RV32 and Sv32 MMU ?

@gagachang when compiling for RV32 had to fix several issues, I'll open a separate PR to fix those.

gagachang:
> @gagachang when compiling for RV32 had to fix several issues, I'll open a separate PR to fix those.

Great, Thanks!

core/arch/riscv/plat-cva6apu/conf.mk
```diff
@@ -0,0 +1,53 @@
...
+supported-ta-targets = ta_rv32
+endif
+
+$(call force,CFG_RISCV_ISA_C,y)
+
+$(call force,CFG_CORE_LARGE_PHYS_ADDR,y)
+$(call force,CFG_CORE_RESERVED_SHM,n)
+$(call force,CFG_CORE_DYN_SHM,y)
+
+$(call force,CFG_DYN_CONFIG,n)
```
gagachang, line 23: Do you have reason why this is disabled?

gagachang: This configuration is planed to be "always enabled", so does ARM.
Which means the configuration may be removed, and the related source code will always be executed. 
If any platform cannot, or doesn't want to, support this configuration, we want to know the reason and see if we can solve the issue here.

maroueneboubakri: There has been an issue with `CFG_DYN_CONFIG` enabled I thought you fixed it in f165144. Will check if that fixes the issue.

gagachang: OK. Please try it!

core/arch/riscv/plat-cva6apu/conf.mk
```diff
@@ -0,0 +1,53 @@
...
+
+CFG_DT ?= y
+
+$(call force,CFG_WITH_SOFTWARE_PRNG,y)
+
+$(call force,CFG_CORE_ASLR,n)
```
peterlin-sifive, line 29: Hi @maroueneboubakri 

We recently added CFG_CORE_ASLR which is enabled by default. When you have a moment, could you please try enabling it and let us know if you encounter any issues?

peterlin-sifive: BTW, the arch-specific `arch_aslr_base_addr()` was added specifically for RV64, but RV32 should be similar to ARM's implementation.

peterlin-sifive:
Do you want to add CI compilation check for your platform ? 
https://github.com/OP-TEE/optee_os/blob/614b28146e96edda3e92c316eb78b592e00072c2/.github/workflows/ci.yml#L289-L292

core/arch/riscv/plat-cva6apu/conf.mk
gagachang: Maybe CVA6 does not support Zkr ?

peterlin-sifive: Okay, that is indeed a blocker for ASLR support if there isn’t an alternative RNG available.

gagachang:
@maroueneboubakri Any followup on this PR?

maroueneboubakri:
Added entry in the CI pipeline to build for the platform.

core/arch/riscv/plat-cva6apu/conf.mk
gagachang: > There has been an issue with `CFG_DYN_CONFIG` enabled I thought you fixed it in [f165144](https://github.com/OP-TEE/optee_os/commit/f1651448a92150c573f763f95293fccaed3af843). Will check if that fixes the issue.

@maroueneboubakri Did you check this issue?

gagachang:
Hi @maroueneboubakri Any progress on this PR?

PR #7460, opened 2025-06-29
RISC-V: Add support of Message Proxy Extension (MPXY)

maroueneboubakri:
The MPXY extension lets supervisor software exchange messages with the SBI implementation using a generic interface that supports different messaging protocols. It requires per-hart shared memory setup before use. Our goal is to add MPXY support in OP-TEE so it can communicate with the Linux kernel (running in the untrusted REE).


gagachang:
CC @peterlin-sifive 

Hi @maroueneboubakri Thanks for bringing this PR in.
In fact, in [RISE OP-TEE project](https://lf-rise.atlassian.net/wiki/spaces/HOME/pages/8587868/OPTEE_00_01+-+OP-TEE+support) we also evaluated the communication based on MPXY and RPMI.
There was a discussion in RVI: https://lists.riscv.org/g/tech-rpmi/topic/109004733
In this discussion we considered using new RPMI/MM and RPMI/ReqFwd service groups to build the communication.
The implementation based on MPXY with RPMI/MM and RPMI/ReqFwd can be found here:
https://gitlab.com/riseproject/riscv-optee/optee_os/-/commit/cac3ca0573786f11affeefbb82597c4ffc56e448

Linux calls MPXY with RPMI/MM service group to send the message to OP-TEE.
OP-TEE calls MPXY with RPMI/ReqFwd service group to retrieve, handle, and return the result to Linux.

Would you like to review this approach and give us some feedback ?

maroueneboubakri:
Hi @gagachang,

Thanks ! The usage of ReqFwd + MM makes much sense.
Are you planning to open a PR soon to add support ?

Best
Maro

gagachang:
Sure, I can create subsequent PR once your PR got merged.
Just want to confirm, does this PR follow latest RPMI spec?

maroueneboubakri:
> Sure, I can create subsequent PR once your PR got merged. Just want to confirm, does this PR follow latest RPMI spec?

This should be separate form RPMI.

gagachang:
> > Sure, I can create subsequent PR once your PR got merged. Just want to confirm, does this PR follow latest RPMI spec?
> 
> This should be separate form RPMI.

Oh sorry, I meant SBI MPXY.
Now I see that the commit message says "Introduced definitions for the ratified MPXY extension as specified in SBI specification version 3.0.". I assume this PR follows latest MPXY spec, thanks.

core/arch/riscv/kernel/sbi.c
```diff
@@ -107,3 +110,417 @@ int sbi_hsm_hart_get_status(uint32_t hartid, enum sbi_hsm_hart_state *status)
...
+ * and copies them from shared memory into the provided buffer.
+ *
+ * Return: SBI_SUCCESS on success, negative SBI error code on failure.
+ */
+int
+sbi_mpxy_get_notification_events(uint32_t channel_id,
+				 struct sbi_mpxy_notification_data *notif_data,
+				 unsigned long *events_data_len)
+{
+	struct mpxy_core_local *mpxy = mpxy_get_core_local();
```
gagachang: This line should be called after `exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);`
Otherwise, CPU may be changed due to any foreign interrupts.
Leading wrong `struct mpxy_core_local`

The same concern happens in other similar functions.

core/arch/riscv/include/sbi.h
```diff
@@ -164,6 +188,30 @@ int sbi_dbcn_write_byte(unsigned char ch);
 int sbi_hsm_hart_start(uint32_t hartid, paddr_t start_addr, unsigned long arg);
 int sbi_hsm_hart_get_status(uint32_t hartid, enum sbi_hsm_hart_state *status);
 
+/*  SBI MPXY */
```
peterlin-sifive: nit

```diff
-/*  SBI MPXY */
+/* SBI MPXY */
```

core/arch/riscv/kernel/sbi.c
```diff
@@ -1,10 +1,13 @@
...
 /*
- * Copyright 2022 NXP
+ * Copyright 2022, 2025 NXP
  */
 
 #include <riscv.h>
 #include <sbi.h>
+#include <kernel/misc.h>
+#include <mm/core_memprot.h>
+#include <string.h>
```
peterlin-sifive: nit: need to put in alphabetical order?

core/arch/riscv/kernel/sbi.c
```diff
@@ -107,3 +110,417 @@ int sbi_hsm_hart_get_status(uint32_t hartid, enum sbi_hsm_hart_state *status)
...
+	ret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_SEND_MSG_WITHOUT_RESP,
+			channel_id, message_id, message_len, 0, 0, 0);
+
+	if (ret.error)
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", ret.error,
+		     ret.value);
+
+	thread_unmask_exceptions(exceptions);
+	return ret.error;
+}
```
peterlin-sifive: `sbi_mpxy_send_message_{with|without}_response()` functions have quite a bit of duplicate code. Can we factor out the common parts?

core/arch/riscv/kernel/sbi.c
maroueneboubakri: Fixed

core/arch/riscv/kernel/sbi.c
maroueneboubakri: I would keep as-is for better readability; the functions are short enough to remain clear.

core/arch/riscv/kernel/sbi.c
maroueneboubakri: Fixed

core/arch/riscv/include/sbi.h
maroueneboubakri: Fixed

core/arch/riscv/plat-virt/conf.mk
```diff
@@ -34,6 +34,9 @@ CFG_RISCV_IMSIC ?= n
 CFG_RISCV_SBI_CONSOLE ?= n
 CFG_16550_UART ?= y
 
+# SBI extensions
+CFG_RISCV_SBI_MPXY ?= y
```
gagachang, line 38: @peterlin-sifive Does SiFive platform need this flag ?

gagachang:
For
- riscv: sbi.h: extend SBI error codes
- riscv: sbi.h: add SBI_EXT_MPXY definition
- riscv: include: misc_arch.h: include stdint.h for fixed-width types
- riscv: sbi.h: expose sbi_ecall helper in public header

Please apply

core/arch/riscv/include/sbi_mpxy.h
```diff
@@ -0,0 +1,169 @@
...
+	uint32_t reserved;
+	uint8_t events_data[];
+};
+
+#include <compiler.h>
+#include <encoding.h>
+#include <stdint.h>
+#include <sys/cdefs.h>
+#include <types_ext.h>
+#include <util.h>
```
gagachang: Move these lines after `#ifndef __ASSEMBLER__`.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+ *
+ * Return: 0 on success, negative SBI error code on failure.
+ */
+int sbi_mpxy_get_shmem_size(unsigned long *shmem_size)
+{
+	struct sbiret sbiret = {};
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_GET_SHMEM_SIZE, 0, 0, 0,
+			   0, 0, 0);
+	if (sbiret.error) {
```
gagachang, line 60: The SBI v3.0 spec says:
```
This function always returns SBI_SUCCESS in sbiret.error and it will return the shared memory size in
sbiret.uvalue.
```

No need to handle the error here.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+	uint32_t exceptions = 0;
+	int ret = SBI_ERR_FAILURE;
+
+	exceptions = thread_mask_exceptions(THREAD_EXCP_ALL);
+
+	mpxy = mpxy_get_core_local();
+	if (mpxy->shmem_active)
+		goto out;
+
+	shmem = memalign(SMALL_PAGE_SIZE, SMALL_PAGE_SIZE);
```
gagachang, line 95: The spec says:
```
The shmem_phys_lo parameter MUST be 4096 bytes aligned
and the shared memory size is assumed to be same as returned by the Get shared memory size function
(FID #0).
```

I think the allocated size here must the same as `shmem_size` from `sbi_mpxy_get_shmem_size()`

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+		goto out;
+	}
+
+	data = mpxy->shmem;
+
+	do {
+		sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_GET_CHANNEL_IDS,
+				   start_index, 0, 0, 0, 0, 0);
+		if (sbiret.error) {
+			EMSG("MPXY SBI call failed: error=%ld value=%ld",
```
gagachang: No need to print `sbiret.value` here.
Because the spec says:
```
The sbiret.uvalue is always set to zero ...
```

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+		memcpy(&channel_ids[start_index], data->channel_array,
+		       count * sizeof(uint32_t));
+		start_index += count;
+	} while (remaining && start_index < channel_count);
+
+	ret = sbiret.error;
+
+out:
+	thread_unmask_exceptions(exceptions);
+	return ret;
```
gagachang: You can return `sbiret.error` directly.
And drop line 163 and 178.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+			goto out;
+		}
+
+		memcpy(response, mpxy->shmem, response_bytes);
+		if (response_len)
+			*response_len = response_bytes;
+	}
+
+	if (sbiret.error)
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here since the spec does not define it.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	if (message_len)
+		memcpy(mpxy->shmem, message, message_len);
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_SEND_MSG_WITHOUT_RESP,
+			   channel_id, message_id, message_len, 0, 0, 0);
+
+	if (sbiret.error)
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here since the spec does not define it.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+/**
+ * sbi_mpxy_get_channel_count - Get the total number of MPXY channels
+ * @channel_count: Pointer to store the total number of channels
+ *
+ * Makes an SBI call to retrieve the number of channels by reading
+ * the remaining and returned fields from the shared memory structure.
+ *
+ * Return: SBI_SUCCESS on success, negative SBI error code on failure.
+ */
+int sbi_mpxy_get_channel_count(uint32_t *channel_count)
```
gagachang, line 398: We already have `sbi_mpxy_get_channel_ids()`.
Do we really need this API?
What's its purpose?

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+
+	if (!mpxy->shmem_active) {
+		ret = SBI_ERR_NO_SHMEM;
+		goto out;
+	}
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_GET_NOTIFICATION_EVENTS,
+			   channel_id, 0, 0, 0, 0, 0);
+	if (sbiret.error) {
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here since the spec does not define it.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+		goto out;
+	}
+
+	memcpy(mpxy->shmem, attributes_buf, attribute_count * sizeof(uint32_t));
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_WRITE_ATTRS, channel_id,
+			   base_attribute_id, attribute_count, 0, 0, 0);
+
+	if (sbiret.error)
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here since the spec does not define it.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+		goto out;
+	}
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_READ_ATTRS, channel_id,
+			   base_attribute_id, attribute_count, 0, 0, 0);
+	if (!sbiret.error)
+		memcpy(attribute_buf, (void *)mpxy->shmem,
+		       attribute_count * sizeof(uint32_t));
+	else
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here since the spec does not define it.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+	if (!shmem)
+		goto out;
+
+	mpxy->shmem = shmem;
+	mpxy->shmem_pa = virt_to_phys(shmem);
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_SET_SHMEM, mpxy->shmem_pa,
+			   0, 0);
+	if (sbiret.error) {
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here since the spec does not define it.

core/arch/riscv/kernel/sbi_mpxy.c
maroueneboubakri: Anticipate buffer size to be dynamically allocated for `sbi_mpxy_get_channel_ids()`

gagachang: Oh I see!

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+		ret = SBI_ERR_NO_SHMEM;
+		goto out;
+	}
+
+	data = mpxy->shmem;
+
+	sbiret = sbi_ecall(SBI_EXT_MPXY, SBI_EXT_MPXY_GET_CHANNEL_IDS, 0, 0, 0,
+			   0, 0, 0);
+	if (sbiret.error) {
+		EMSG("MPXY SBI call failed: error=%ld value=%ld", sbiret.error,
```
gagachang: No need to print `sbiret.value` here because it is always zero.

core/arch/riscv/kernel/sbi_mpxy.c
maroueneboubakri: Isn't it actually the other way around ?

gagachang: The needed size of shared memory is determined by SBI implementation.
```
1. The shared memory size MUST be same for all HARTs
2. The shared memory size MUST be at least 4096 bytes
3. The shared memory size MUST be multiple of 4096 bytes
4. The shared memory size MUST not be less than the biggest MSG_DATA_MAX_LEN attribute value
across all MPXY channels
```

core/arch/riscv/plat-virt/conf.mk
peterlin-sifive: Yes, we follow the same approach with plat-virt.

gagachang: @maroueneboubakri Please also add this line into `core/arch/riscv/plat-sifive/conf.mk`.

core/arch/riscv/kernel/sbi_mpxy.c
```diff
@@ -0,0 +1,502 @@
...
+	struct mpxy_core_local *mpxy = NULL;
+	size_t pos = 0;
+	uint32_t hart_id = 0;
+
+	assert((thread_get_exceptions() & THREAD_EXCP_ALL) == THREAD_EXCP_ALL);
+
+	pos = get_core_pos();
+	hart_id = thread_get_hartid_by_hartindex(pos);
+
+	mpxy = &mpxy_core_local_array[hart_id];
```
peterlin-sifive, line 40: Just a heads-up: `mpxy_core_local_array[]` cannot be indexed by `hart_id`, since some platforms have CPUs with non-contiguous hart IDs, we need some way to associate hart_id/channel_id/shmem in future patches. 

C.c. @gagachang.

peterlin-sifive:
If convenient, please add `CFG_RISCV_SBI_MPXY ?= y` to `plat-sifive/conf.mk`.

For the series,

core/arch/riscv/kernel/sbi_mpxy.c
gagachang: Sure.

gagachang:
For 
- riscv: introduce CFG_RISCV_SBI_MPXY flag for MPXY SBI extension support

Please apply `Reviewed-by: Alvin Chang <alvinga@andestech.com>`

jenswikl:
Please add a `core: ` prefix to the commits that don't already have one.

jforissier:
This looks ready for merging. @gagachang any comment?

gagachang:
> This looks ready for merging. @gagachang any comment?

There are still two unresolved comments in the PR.
I will comment them later.

core/arch/riscv/kernel/sbi_mpxy.c
gagachang: Hi @maroueneboubakri 
Any reason why we need to handle `sbiret.error` here?
As the spec says, this SBI call always returns `SBI_SUCCESS`.
The error handling here is seemed as though redundant.

Furthermore, this function can be declared as
```c
void sbi_mpxy_get_shmem_size(unsigned long *shmem_size)
```

core/arch/riscv/kernel/sbi_mpxy.c
gagachang: @maroueneboubakri 
According to the spec and the referenced implementation in OpenSBI.
It's possible that SBI implementation returns `shmem_size` larger than `SMALL_PAGE_SIZE` in `SBI_EXT_MPXY_GET_SHMEM_SIZE`.

In my opinion, the code here should be 
```c
sbi_mpxy_get_shmem_size(&shmem_size);
shmem = memalign(SMALL_PAGE_SIZE, shmem_size);
```

maroueneboubakri: @gagachang still unclear, according to the spec, `sbi_mpxy_set_shmem()` should be the very first function to be called before calling any other function of the extension. Quick look at the SBI implemenation in Opensbi, looks like it is upading the shared memory size through this function, so if we set it to 4Kb (current OP-TEE implementation) then later `sbi_mpxy_get_shmem_size()` should return 4Kb. Do you agree ? So it does not make much sense to request shared memory size and update it with same size (somehow).

gagachang: > @gagachang still unclear, according to the spec, `sbi_mpxy_set_shmem()` should be the very first function to be called before calling any other function of the extension.

Yes, I also notice the note under `20.6. Function: Set shared memory (FID #1)`
```
The supervisor software MUST call this function to setup the shared memory first before
calling any other function in this extension.
```

However, I believe`sbi_mpxy_get_shmem_size()` must be prior to `sbi_mpxy_set_shmem()`.
Please see my analysis below.

> Quick look at the SBI implemenation in Opensbi, looks like it is upading the shared memory size through this function, so if we set it to 4Kb (current OP-TEE implementation) then later `sbi_mpxy_get_shmem_size()` should return 4Kb. Do you agree ? So it does not make much sense to request shared memory size and update it with same size (somehow).

I don't think so.
Look at OpenSBI, [sbi_mpxy_get_shmem_size()](https://github.com/riscv-software-src/opensbi/blob/master/lib/sbi/sbi_mpxy.c#L328) returns `mpxy_shmem_size` global variable.
However, [`sbi_mpxy_set_shmem()`](https://github.com/riscv-software-src/opensbi/blob/master/lib/sbi/sbi_mpxy.c#L333) does not update `mpxy_shmem_size`.

The `mpxy_shmem_size` is updated in [`sbi_mpxy_register_channel()`](https://github.com/riscv-software-src/opensbi/blob/master/lib/sbi/sbi_mpxy.c#L234) based on `channel->attrs.msg_data_maxlen`.
For example, if there is any registered channel's message data length greater than 4KB, `mpxy_shmem_size` will be updated to 8KB.

Then, according to this requirement in the spec `20.5. Function: Get shared memory size (FID #0)`:
```
4. The shared memory size MUST not be less than the biggest MSG_DATA_MAX_LEN attribute value
across all MPXY channels
```

And the sentence in the spec `20.6. Function: Set shared memory (FID #1)`
```
The shmem_phys_lo parameter MUST be 4096 bytes aligned
and the shared memory size is assumed to be same as returned by the Get shared memory size function
(FID #0).
```

What do you think?

gagachang: `sbi_mpxy_register_channel()` is executed during OpenSBI boot, prior to any supervisor domain.
If `mpxy_shmem_size` is updated to 8KB during `sbi_mpxy_register_channel()`, it implies supervisor domain needs to allocate MPXY shared-memory at least 8KB.
How supervisor domain knows this size constraint and allocates 8KB memory before `sbi_mpxy_set_shmem()`?
Obviously, supervisor domain must call `sbi_mpxy_get_shmem_size()`.

This is why I think `sbi_mpxy_get_shmem_size()` is prior to `sbi_mpxy_set_shmem()`.

gagachang: @maroueneboubakri 
I found the original proposal of `sbi_mpxy_get_shmem_size()` in https://lists.riscv.org/g/tech-prs/message/1286.
The `sbi_mpxy_get_shmem_size()` is used to let supervisor software discover "the biggest
MSG_DATA_MAX_LEN attribute value across all MPXY channels".

maroueneboubakri: Which version of OpenSBI are you using ? on 1.6 it updates the shared memory size https://github.com/riscv-software-src/opensbi/blob/v1.6/lib/sbi/sbi_mpxy.c#L297 
Since RPMI PR depends on this and stalling, let's proceed with the merge and then feel free to open a PR to align.
Does this work for you @gagachang ?

gagachang: OpenSBI v1.6 does not support SBI v3.0 version.
You need to integrate OpenSBI v1.7 version.

I just raise this MPXY shared memory question on PRS TG mailing list.

maroueneboubakri: Let's wait for their reply.

core/arch/riscv/kernel/sbi_mpxy.c
gagachang: @maroueneboubakri Please consider this change.
This change is compatible with latest SBI v3.0 and OpenSBI v1.7 implementation.

If you concern about older SBI implementation which doesn't support `SBI_EXT_MPXY_GET_SHMEM_SIZE`, you should consider using `SBI_EXT_BASE_GET_SPEC_VERSION` to check whether SBI implementation supports MPXY or not.
The SBI MPXY extension is added from SBI v3.0 specification.

core/arch/riscv/kernel/sbi_mpxy.c
gagachang: > Let's wait for their reply.

Got a reply: https://lists.riscv.org/g/tech-prs/message/1480
If it is just a recommendation, we can merge this PR first and improve it in future PR.

maroueneboubakri: Ok then let's merge and open a new PR to handle the recommandations/suggestions.

gagachang:
@maroueneboubakri 

For
- core: riscv: implement MPXY SBI extension
- core: riscv: introduce CFG_RISCV_SBI_MPXY flag for MPXY SBI support

Please apply `Reviewed-by: Alvin Chang <alvinga@andestech.com>`

maroueneboubakri:
@jforissier kinldy merge this PR.

gagachang (approved):

PR #7465, opened 2025-07-02
Improve support of IWDG in STM32MPU

Upsylonbare:


core/drivers/stm32_iwdg.c
```diff
@@ -135,25 +158,68 @@ static TEE_Result iwdg_wait_sync(struct stm32_iwdg_device *iwdg)
...
+	 */
+	io_write32(iwdg_base + IWDG_KR_OFFSET, IWDG_KR_WPROT_KEY);
+
+	/* Disable early interrupt */
+	if (iwdg->hw_version >= IWDG_ICR_MIN_VER)
+		io_setbits32(iwdg_base + IWDG_ICR_OFFSET, IWDG_ICR_EWIC);
+	else
+		io_setbits32(iwdg_base + IWDG_EWCR_OFFSET, IWDG_EWCR_EWIC);
+
+	panic("Watchdog");
```
GseoC: Maybe we can specify: `Early watchdog interrupt` here?

core/drivers/stm32_iwdg.c
```diff
@@ -158,6 +162,16 @@ static TEE_Result iwdg_wait_sync(struct stm32_iwdg_device *iwdg)
 	return TEE_SUCCESS;
 }
 
+static void stm32_iwdg_it_ack(struct stm32_iwdg_device *iwdg)
+{
+	vaddr_t iwdg_base = get_base(iwdg);
+
+	if (iwdg->hw_version >= IWDG_ICR_MIN_VER)
+		io_setbits32(iwdg_base + IWDG_ICR_OFFSET, IWDG_ICR_EWIC);
+	else
```
GseoC, line 179: I'd rename the function to : `stm32_iwdg_disable_ewi()`

core/drivers/stm32_iwdg.c
```diff
@@ -275,9 +286,27 @@ static void iwdg_wdt_start(struct wdt_chip *chip)
...
+	if (iwdg->reset)
+		if (rstctrl_assert(iwdg->reset))
+			panic();
+}
+
+static void iwdg_wdt_stop(struct wdt_chip *chip)
+{
+	struct stm32_iwdg_device *iwdg = wdt_chip_to_iwdg(chip);
+
+	if (iwdg->reset) {
```
GseoC, line 318: Is there a bit somewhere in the IWDG/RCC to know if this is feasible? I guess this sequence will simply have no effect for IWDG1?
BTW, in the commit message: s/STM32MP/STM32MP2

core/drivers/stm32_iwdg.c
```diff
@@ -275,9 +286,27 @@ static void iwdg_wdt_start(struct wdt_chip *chip)
 	struct stm32_iwdg_device *iwdg = wdt_chip_to_iwdg(chip);
 
 	iwdg_start(iwdg);
+	if (iwdg->reset && iwdg->itr_handler)
+		stm32_iwdg_it_ack(iwdg);
```
GseoC, line 304: I don't understand why we do this at start and not before calling` interrupt_disable()` in `iwdg_wdt_stop()?`

core/drivers/stm32_iwdg.c
```diff
@@ -132,6 +139,17 @@ static bool iwdg_wdt_is_enabled(struct stm32_iwdg_device *iwdg)
 	return iwdg->flags & IWDG_FLAGS_ENABLED;
 }
 
+static void iwdg_refresh(struct stm32_iwdg_device *iwdg)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	res = tee_time_get_sys_time(&iwdg->last_refresh);
```
GseoC: Why is this function moved?

core/drivers/stm32_iwdg.c
```diff
@@ -209,11 +209,13 @@ static enum itr_return stm32_iwdg_it_handler(struct itr_handler *h)
 	 */
 	io_write32(iwdg_base + IWDG_KR_OFFSET, IWDG_KR_WPROT_KEY);
 
-	/* Disable early interrupt */
+	/* Ack early interrupt */
 	stm32_iwdg_it_ack(iwdg);
 
 	if (iwdg->nb_int > 1) {
-		iwdg->nb_int--;
+		/* Decrease interrupt counter when watchdog is not stopped*/
```
GseoC, line 206: maybe we can use a boolean representing the status of the watchdog instead, no?

But anyway, I guess we do this to support IWDG that don't support their reset capability being removed? If so, this should be explicitly mentionned

core/drivers/stm32_iwdg.c
```diff
@@ -48,9 +48,9 @@
 #define IWDG_KR_RELOAD_KEY	U(0xAAAA)
 #define IWDG_KR_START_KEY	U(0xCCCC)
 
-/* Use a fixed prescaler divider of 256 */
-#define IWDG_PRESCALER_256	U(256)
-#define IWDG_PR_DIV_256		U(0x06)
+/* Use a fixed prescaler divider of 1024 */
+#define IWDG_PRESCALER_1024	U(1024)
```
GseoC, line 52: This should be 0xf instead IMO. The bitfield is 4 bits wide

GseoC (commented):
Hello Clément, first round.

Can you move device tree commits before the platform makefile commit please?

There is a  leftover change IDs in "plat-stm32mp2: enable watchdog SMC service" and "plat-stm32mp2: conf: default enable CFG_STM32_IWDG" commit messages. These commits are also missing your SoB.

core/drivers/stm32_iwdg.c
clegoffic: In the reference manual it is plainly written 1024. And I find it easier to read / use for calculations.

core/drivers/stm32_iwdg.c
clegoffic: I'll undo the move, indeed no need.

core/drivers/stm32_iwdg.c
clegoffic: No it is an acknowledgment but I can use `stm32_iwdg_early_wakeup_it_clear` if you prefer
We don't disable anything here, just clear the IT bit

core/drivers/stm32_iwdg.c
clegoffic: I think you miss the point here.
`nb_int` is used to support big timeout like 10 minutes so we divide this big timeout by our maximum timeout.
You can check `stm32_iwdg_handle_big_timeout`  to see nb_int assingation

core/drivers/stm32_iwdg.c
clegoffic: We do a ack when we start the watchdog to be sure we don't have any pending interrupt.

core/drivers/stm32_iwdg.c
clegoffic: Yes we have `RCC_IWDGC1CFGCLRR` `RCC_IWDGC1CFGSETR` but nothing is interesting when the CPU1 is the trusted CPU.

clegoffic: I'll update the commit message

core/drivers/stm32_iwdg.c
clegoffic: The purpose of the early watchdog is to inform of the reset of the platform. So I think it is ok like that no ?

clegoffic:
Added my SoB, remove the change-id, reorder patches. And undo the function move.

core/drivers/stm32_iwdg.c
GseoC: The reset reason (if any) is not displayed by OP-TEE but anyway, it seems okay as is. Just discard this

core/drivers/stm32_iwdg.c
GseoC: It's not what I meant but I mixed the value in the register and the prescaler value so yes, it's correct.

core/drivers/stm32_iwdg.c
GseoC: My bad, I mixed `IWDG_EWCR_EWIC` and `IWDG_EWCR_EWIE`. Meanings and bit position changed from one revision to another so it's a bit of a mess.
Keep as is, it's correct

core/drivers/stm32_iwdg.c
```diff
@@ -313,19 +325,49 @@ static void iwdg_wdt_refresh(struct wdt_chip *chip)
...
+	long w = 0;
+
+	rate = clk_get_rate(iwdg->clk_lsi);
+	iwdg->max_hw_timeout = (IWDG_CNT_MASK + 1) * IWDG_PRESCALER_256 / rate;
+	interval = iwdg->max_hw_timeout - IWDG_ETIMEOUT_SEC;
+
+	if (timeout_sec > IWDG_ETIMEOUT_SEC)
+		n = (timeout_sec - IWDG_ETIMEOUT_SEC) / interval;
+
+	if ((timeout_sec - IWDG_ETIMEOUT_SEC) % interval != 0) {
```
GseoC: if `timeout_sec< IWDG_ETIMEOUT_SEC `we'll face an underflow here. Maybe when parsing the DT and in `iwdg_wdt_set_timeout()`, we should forbid values less than `IWDG_ETIMEOUT_SEC`.

This will also allow to remove the if condition lign 344

GseoC: in `iwdg_wdt_set_timeout()`:

```
if (timeout < IWDG_ETIMEOUT_SEC)
```

and simple assert when parsing the DT.

Moreover, as the early watchdog delay is fixed. I suspect trying to set a value barely above it will not work as well (no time to kick the watchdog), the EWI will fire practically instantly, no?

clegoffic: Indeed need to test this

clegoffic: So for the underflow, as we remove and then add IWDG_TIMEOUT_SEC I think we don't have it.
Then for the ETIMEOUT what can be considered as a low value for you ?
Like 2 times it ?
Maybe I can add a WARN message as it is dangerous but it should be allowed

GseoC: For: `if ((timeout_sec - IWDG_ETIMEOUT_SEC) % interval != 0) {`

`(timeout_sec - IWDG_ETIMEOUT_SEC)` will underflow, then be divided by interval to get the rest, so...

clegoffic: Will be fix

clegoffic:
Hello, new version with various fixes after some tests.
The move of the setup() function in the probe increase readability and is less error prone.
I also fixed two error path in the parse_fdt and in the pm function add in the probe.
The function handle_timeouts is now more readable.

core/drivers/stm32_iwdg.c
```diff
@@ -135,25 +158,72 @@ static TEE_Result iwdg_wait_sync(struct stm32_iwdg_device *iwdg)
...
 	assert(iwdg_wdt_is_enabled(iwdg));
 
 	rlr_value = iwdg_timeout_cnt(iwdg, iwdg->timeout);
 	if (!rlr_value)
 		return TEE_ERROR_GENERIC;
 
+	if (iwdg->itr_handler) {
+		if (iwdg->timeout >= 2 * IWDG_ETIMEOUT_SEC)
+			early_timeout = IWDG_ETIMEOUT_SEC;
+		else
```
GseoC: This approach is safer but there's a range of value that's not really reachable.

Don't  you think the approach that was taken in the kernel to have a proportion of the timeout is preferable?

TBH, I'd rather even have an "st,early-timeout" property in the device tree that have a default value at SoC level and assert its value against the timeout value. What do you think of it?

core/drivers/stm32_iwdg.c
```diff
@@ -383,6 +413,10 @@ static TEE_Result stm32_iwdg_parse_fdt(struct stm32_iwdg_device *iwdg,
 			return res;
 	}
 
+	res = rstctrl_dt_get_by_index(fdt, node, 0, &iwdg->reset);
+	if (res != TEE_SUCCESS && res != TEE_ERROR_ITEM_NOT_FOUND)
```
GseoC: Nit: could be `if (res && res != TEE_ERROR_ITEM_NOT_FOUND)` for consistency with the rest

core/drivers/stm32_iwdg.c
```diff
@@ -317,22 +325,51 @@ static void iwdg_wdt_refresh(struct wdt_chip *chip)
...
+	unsigned long interval = 0;
+	unsigned long rate = 0;
+	unsigned long n = 0;
+	long w = 0;
+
+	rate = clk_get_rate(iwdg->clk_lsi);
+	iwdg->max_hw_timeout = (IWDG_CNT_MASK + 1) * IWDG_PRESCALER_256 / rate;
+
+	if (timeout_sec > iwdg->max_hw_timeout) {
+		IMSG("Timeout exceeds hardware capability, approximate it\n");
```
GseoC: `\n` not needed

core/drivers/stm32_iwdg.c
```diff
@@ -358,7 +395,8 @@ static TEE_Result iwdg_wdt_get_timeleft(struct wdt_chip *chip, bool *is_started,
 	if (res)
 		panic();
 
-	time.seconds = iwdg->timeout;
+	time.seconds = (iwdg->timeout - IWDG_ETIMEOUT_SEC) * iwdg->saved_nb_int
```
GseoC: I think `iwdg->early_timeout` should be used in place of `IWDG_ETIMEOUT_SEC` here.

core/drivers/stm32_iwdg.c
```diff
@@ -317,22 +325,51 @@ static void iwdg_wdt_refresh(struct wdt_chip *chip)
...
 
-	if (!iwdg_timeout_cnt(iwdg, timeout))
-		return TEE_ERROR_BAD_PARAMETERS;
-
-	iwdg->timeout = timeout;
-
 	if (iwdg_wdt_is_enabled(iwdg)) {
 		TEE_Result res = TEE_ERROR_GENERIC;
 
+		stm32_iwdg_handle_timeouts(iwdg, timeout);
```
GseoC, line 380: Why do we do this only if the watchdog is already enabled? If the watchdog wasn't started before, then the `stm32_iwdg_setup()` won't do anything to the timeout value and we cannot set the timeout value with the .set_timeout() ops before starting the watchdog with .start() ops, right?

core/drivers/stm32_iwdg.c
clegoffic: Yes I admit it it is a choice.
The kernel approach is fine TBH having an early timeout that is 3/4 of the timeout is a bit overkill to me.
The devicetree property would be the best approach but for the moment I don't consider it as there is nothing in the kernel bindings.

core/drivers/stm32_iwdg.c
clegoffic: Ack

core/drivers/stm32_iwdg.c
clegoffic: Ack

core/drivers/stm32_iwdg.c
clegoffic: It has been a choice, if the watchdog is not started by a prior software component (in our case TFA) we don't do anything. 
We can think to allow a set_timeout but is this really needed ? IDK

core/drivers/stm32_iwdg.c
clegoffic: Right good catch

core/drivers/stm32_iwdg.c
```diff
@@ -302,26 +447,45 @@ static TEE_Result stm32_iwdg_parse_fdt(struct stm32_iwdg_device *iwdg,
...
-		EMSG("Timeout %lu not applicable", iwdg->timeout);
-		return TEE_ERROR_BAD_PARAMETERS;
+	if (!iwdg->timeout) {
+		res = TEE_ERROR_BAD_PARAMETERS;
+		goto out;
 	}
 
 	return TEE_SUCCESS;
+
+out:
```
GseoC: Please rename `out` label to `err` as it is used for error cases

core/drivers/stm32_iwdg.c
GseoC: To me, it seems strange to enable a peripheral's node and configure timeout values in the device tree without starting and configuring it. If we're doing this, I expect at least a trace at probe to inform about that.

In a second time (other P-R) we can decide if we should start the watchdog after discussing it with other contributors.

GseoC (commented):
With comments addressed, please squash and apply:

core/drivers/stm32_iwdg.c
```diff
@@ -209,31 +300,82 @@ static void iwdg_wdt_start(struct wdt_chip *chip)
...
+		w = ((timeout_sec - IWDG_ETIMEOUT_SEC) / (n + 1)) +
+		IWDG_ETIMEOUT_SEC;
+		iwdg->timeout = w;
+		iwdg->early_timeout = IWDG_ETIMEOUT_SEC;
+	} else {
+		iwdg->timeout = timeout_sec;
+		if (iwdg->timeout >= 2 * IWDG_ETIMEOUT_SEC)
+			iwdg->early_timeout = IWDG_ETIMEOUT_SEC;
+		else
+			iwdg->early_timeout = iwdg->timeout / 4;
```
GseoC, line 362: if `iwdg->timeout < 4`, then `iwdg->early_timeout == 0`. To avoid such cases, either change the computation so that's it's more of a percentage of the timeout or assert `iwdg->early_timeout` isn't null

core/drivers/stm32_iwdg.c
clegoffic: Ack

core/drivers/stm32_iwdg.c
clegoffic: Maybe we should rely on DT property such as UBOOT has "uboot,autostart" to force the start it if it has not been started.

core/drivers/stm32_iwdg.c
clegoffic: if (!iwdg->early_timeout)
		iwdg->early_timeout = 1;

core/drivers/stm32_iwdg.c
GseoC: Ok let's see in a second time

GseoC (approved):

GseoC:
Hello @jforissier, can you merge this P-R if that's OK for you as well please?

jforissier:
@GseoC sure. Thanks for the reminder.

GseoC:
Thank you :)

PR #7467, opened 2025-07-03
[RFC] Revert "plat-rockchip: rk3399: remove GIC configuration"

jenswi-linaro:
With commit 4cb77793842a ("irqchip/gic-v3: Fix rk3399 workaround when secure interrupts are enabled") in the Linux kernel OP-TEE panics after the kernel has booted with:
E/TC:3 0 Panic 'Secure interrupt handler not defined' at core/kernel/interrupt.c:105 <interrupt_main_handler>

So for kernels after v6.14 we need another workaround. The easiest is to revert commit 447c5f6bc49ff5408c0543ceaaabf0cb8f23804d. The GIC is still broken, but the device is still usable in other aspects.


jforissier:
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (rockchip-rk3399) (Rockpi4B)`

jenswikl:
Tag applied

jenswikl:
I missed Etienne's tag, it's applied now.

PR #7469, opened 2025-07-03
Fix boot STM32MP157c-ev1 probe stm32_tamp always defer

tbourgoi:
Hello,

The board stm32mp157c-ev1 fails to boot because of the driver stm32_tamp which always return TEE_ERROR_DEFER_DRIVER_INIT.
TAMP drivers mandate to have the RTC enabled.
TAMP node is enabled by default in SoC device tree but the RTC node is enabled in the board (ex stm32mp135f-dk.dts)

On stm32mp25, TAMP and RTC node are enabled by default in the SoC device tree.

To be coherent with stm32mp25, this pull request enable RTC by default in SoC device tree (stm32mp131.dtsi and stm32mp151.dtsi) and remove RTC node enable in board device tree (stm32mp135f-dk, stm32mp157c-ed1 and stm32mp15xx-dkx.dtsi)

Thomas

### DIFF
--- a/core/arch/arm/dts/stm32mp157c-ed1.dts
+++ b/core/arch/arm/dts/stm32mp157c-ed1.dts
@@ -370,10 +370,6 @@
 	status = "okay";
 };
 
-&rtc {
-	status = "disabled";
-};
-
 &sdmmc1 {
 	pinctrl-names = "default", "opendrain", "sleep";
 	pinctrl-0 = <&sdmmc1_b4_pins_a &sdmmc1_dir_pins_a>;
--- a/core/arch/arm/dts/stm32mp135f-dk.dts
+++ b/core/arch/arm/dts/stm32mp135f-dk.dts
@@ -516,10 +516,6 @@
 	clock-error-detect;
 };
 
-&rtc {
-	status = "okay";
-};
-
 &saes {
 	status = "okay";
 };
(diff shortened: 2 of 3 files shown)


core/arch/arm/dts/stm32mp157c-ed1.dts
```diff
@@ -370,10 +370,6 @@
 	status = "okay";
 };
 
-&rtc {
-	status = "disabled";
```
GseoC, line 374: This belongs in the second commit

GseoC (commented):
One comment, else:

This patchset is needed for 4.7.0

etienne-lms (commented):
Enabling of RTC for mp15/mp13 should be done in the board DTS files, not in the SoC DTSI files IMHO.

core/arch/arm/dts/stm32mp157c-ed1.dts
tbourgoi: Done

tbourgoi:
> Enabling of RTC for mp15/mp13 should be done in the board DTS files, not in the SoC DTSI files IMHO.

In this case we cannot enable by default TAMP in SoC DTSI file

GseoC:
TAMP is mandatory ATM for the backup register management

etienne-lms:
Ok then, at least this is consistent with MP25.

tbourgoi:
Comment from GseoC taken into account and both tags applied.

PR #7470, opened 2025-07-03
plat-corstone1000: increase CFG_TZDRAM_SIZE

frazer-arm:
TZDRAM is a 4MB SRAM in Corstone-1000. Its start address is `0x0200_0000` but the first 0x2000 bytes are reserved for future use. `CFG_TZDRAM_SIZE` can be increased to `0x360000` so OP-TEE has more RAM.


jenswikl:

frazer-arm:
ack added to commit msg; ready to merge

PR #7472, opened 2025-07-04
plat-vexpress: qemu-v8: fix CFG_CORE_CLUSTER_SHIFT

jenswi-linaro:
QEMU virt puts 8 or 16 CPUs in each cluster, depending on the GIC configuration [1]. So set CFG_CORE_CLUSTER_SHIFT to match the GIC configuration to calculate core position correctly for more than 16 CPUs.

Link: https://gitlab.com/qemu-project/qemu/-/blob/v10.0.0/hw/arm/virt.c?ref_type=tags#L1785-1789 [1]

This isn't urgent, so it can wait until after the release if we've already tested QEMU-v8.


core/arch/arm/plat-vexpress/conf.mk
```diff
@@ -136,6 +136,11 @@ CFG_CORE_HEAP_SIZE ?= 196608
 CFG_HALT_CORES_ON_PANIC ?= y
 CFG_TEE_CORE_NB_CORE ?= 4
 CFG_AUTO_MAX_PA_BITS ?= y
+ifeq ($(CFG_ARM_GICV3),y)
+CFG_CORE_CLUSTER_SHIFT ?= 4
```
jforissier, line 140: This line does fix `make check CFG_TEE_CORE_NB_CORE=24 CFG_NUM_THREADS=24 QEMU_SMP=24 GICV3=y`...

core/arch/arm/plat-vexpress/conf.mk
```diff
@@ -136,6 +136,11 @@ CFG_CORE_HEAP_SIZE ?= 196608
...
+else
+CFG_CORE_CLUSTER_SHIFT ?= 3
```
jforissier, line 142: ...but I can't see which difference this one makes. With `GICV3=n` QEMU cannot be run with more than 8 cores (`QEMU_SMP=8`) which was already working before this change (`make -j$(nproc) check CFG_TEE_CORE_NB_CORE=8 CFG_NUM_THREADS=8 QEMU_SMP=8 GICV3=n`).

jenswikl: That works because the cluster ID is always `0`, so the core ID can overflow into the cluster part of the core position without getting garbled. This is more correct and hopefully less confusing.

jforissier: Got it, thanks

jforissier:

jenswikl:
Tags applied

PR #7475, opened 2025-07-08
Add support for marvell CN20K SoCs

anilreddy3:
Add support for Octeon20(CN20K) SoCs from Marvell.

Only tested 64-bit mode with default configurations:

1. Build command 
             make PLATFORM=marvell-cn20ka 
             make PLATFORM=marvell-cnf20ka
3. Passed xtest


core/arch/arm/plat-marvell/cn20k
jenswikl, line 1: How about using cn10k/core_pos.S directly instead?
I'm also sceptical of symlinks in the repository. I don't think we have any others yet.

anilreddy3: Okay.
I will make a change use cn10k/core_pos.S directly.

anilreddy3:
Thanks @jenswi-linaro, applied the tag.

jforissier:
Although not a bug fix this looks totally safe to be merged for the 4.7.0 release so I'm taking it now. Thanks!

PR #7477, opened 2025-07-09
ta/link.mk: update the default TA encryption key

jforissier:
When the TA signing key for OP-TEE was changed [1], the TA encryption key (which is a derived key) was not updated. As a result, CFG_ENCRYPT_TA=y is broken. Fix that by updating TA_ENC_KEY to reflect the output of tee_otp_get_ta_enc_key().

Fixes: 5d5d7d0b1c03 ("keys: increase default RSA key size to 4096 bits") [1]


jenswikl:
Please add instructions for how the key was derived.

jforissier:
Update

jenswikl:

PR #7478, opened 2025-07-09
Fix GPIO framework to prevent glitch when requesting an output GPIO

tbourgoi:
Hello,

During internal test I discovered that a voltage glitch appeared on boards with discrete regulators.
After further investigation I find out that when a GPIO is retrieved from the device tree it's configured as input and then the function gpio_set_direction() is called to configure the GPIO as output.
This intermediate configuration creates a glitch on the voltage line, we should only configure once a GPIO. Especially since the configuration might already be set by the previous bootloader.

When discussing the issue with @patrickdelaunay we found another bug in the `regulator_gpio` and `regulator_fixed` drivers.
The properties `enable-active-high`, `gpio-open-drain` from the device tree are not taken into account. The device tree is parsed and variable `gpio->dt_flags` is updated but the configuration of the GPIO with the device tree flags is done in function `gpio_dt_get_by_index()` which is called before parsing the custom options.

So this PR does the following :
- Update `gpio_dt_get_by_index()` to search for properties with and without suffixes
- Add `.configure()` ops in GPIO framework configure a GPIO after the parsing of custom properties.
- Add function `gpio_dt_cfg_by_index()` to request a GPIO from the DT and configure it.

Thomas

### DIFF
--- a/core/drivers/gpio/gpio.c
+++ b/core/drivers/gpio/gpio.c
@@ -10,6 +10,9 @@
 #include <tee_api_types.h>
 #include <util.h>
 
+/* gpio suffixes used for device tree lookup */
+static const char * const gpio_suffixes[] = { "gpios", "gpio" };
+
 TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
 {
 	struct gpio *gpio = NULL;
@@ -29,41 +32,33 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
 	return TEE_SUCCESS;
 }
 
-static char *gpio_get_dt_prop_name(const char *gpio_name)
-{
-	int ret = 0;
-	char *prop_name = NULL;
-	size_t max_len = strlen(gpio_name) + strlen("-gpios") + 1;
-
-	prop_name = calloc(1, max_len);
-	if (!prop_name)
-		return NULL;
-
-	ret = snprintf(prop_name, max_len, "%s-gpios", gpio_name);
-	if (ret < 0 || (size_t)ret >= max_len) {
-		free(prop_name);
-		return NULL;
-	}
-
-	return prop_name;
-}
-
 TEE_Result gpio_dt_get_by_index(const void *fdt, int nodeoffset,
 				unsigned int index, const char *gpio_name,
 				struct gpio **gpio)
 {
 	TEE_Result res = TEE_ERROR_GENERIC;
-	char *prop_name = NULL;
+	char prop_name[32]; /* 32 is max size of property name in DT */
 	void *out_gpio = NULL;
+	unsigned int i = 0;
 
-	prop_name = gpio_get_dt_prop_name(gpio_name);
-	if (!prop_name)
-		return TEE_ERROR_OUT_OF_MEMORY;
+	/* Try GPIO properties "foo-gpios" and "foo-gpio" */
+	for (i = 0; i < ARRAY_SIZE(gpio_suffixes); i++) {
+		if (gpio_name)
+			snprintf(prop_name, sizeof(prop_name), "%s-%s",
+				 gpio_name, gpio_suffixes[i]);
+		else
+			snprintf(prop_name, sizeof(prop_name), "%s",
+				 gpio_suffixes[i]);
 
-	res = dt_driver_device_from_node_idx_prop(prop_name, fdt, nodeoffset,
-						  index, DT_DRIVER_GPIO,
-						  &out_gpio);
-	free(prop_name);
+		res = dt_driver_device_from_node_idx_prop(prop_name, fdt,
+							  nodeoffset,
+							  index, DT_DRIVER_GPIO,
+							  &out_gpio);
+
+		if (res != TEE_ERROR_ITEM_NOT_FOUND)
+			break;
+	}
+
 	if (!res)
 		*gpio = out_gpio;
 
--- a/core/drivers/gpio/gpio.c
+++ b/core/drivers/gpio/gpio.c
@@ -69,6 +69,12 @@ TEE_Result gpio_configure(struct gpio *gpio, enum gpio_flags flags)
 {
 	enum gpio_level value = GPIO_LEVEL_LOW;
 
+	assert(!gpio || (gpio->chip && gpio->chip->ops));
+
+	/* Configure GPIO with DT flags */
+	if (gpio && gpio->chip->ops->configure)
+		gpio->chip->ops->configure(gpio->chip, gpio);
+
 	/* Process requester flags */
 	if (flags & GPIO_FLAGS_BIT_DIR_SET) {
 		if (flags & GPIO_FLAGS_BIT_DIR_OUT) {
--- a/core/drivers/gpio/gpio.c
+++ b/core/drivers/gpio/gpio.c
@@ -64,3 +64,34 @@ TEE_Result gpio_dt_get_by_index(const void *fdt, int nodeoffset,
 
 	return res;
 }
+
+TEE_Result gpio_configure(struct gpio *gpio, enum gpio_flags flags)
+{
+	enum gpio_level value = GPIO_LEVEL_LOW;
+
+	/* Process requester flags */
+	if (flags & GPIO_FLAGS_BIT_DIR_SET) {
+		if (flags & GPIO_FLAGS_BIT_DIR_OUT) {
+			if (flags & GPIO_FLAGS_BIT_DIR_VAL)
+				value = GPIO_LEVEL_HIGH;
+			gpio_set_value(gpio, value);
+			gpio_set_direction(gpio, GPIO_DIR_OUT);
+		} else {
+			gpio_set_direction(gpio, GPIO_DIR_IN);
+		}
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result gpio_dt_cfg_by_index(const void *fdt, int nodeoffset,
+				unsigned int index, const char *gpio_name,
+				enum gpio_flags flags, struct gpio **gpio)
+{
+	TEE_Result res = gpio_dt_get_by_index(fdt, nodeoffset, index, gpio_name,
+					      gpio);
+	if (!res)
+		gpio_configure(*gpio, flags);
+
+	return res;
+}
--- a/core/drivers/regulator/regulator_fixed.c
+++ b/core/drivers/regulator/regulator_fixed.c
@@ -78,14 +78,8 @@ static TEE_Result get_enable_gpio(const void *fdt, int node,
 	TEE_Result res = TEE_ERROR_GENERIC;
 	const fdt32_t *cuint = NULL;
 	struct gpio *gpio = NULL;
-	void *gpio_ref = &gpio;
 
-	res = dt_driver_device_from_node_idx_prop("gpios", fdt, node, 0,
-						  DT_DRIVER_GPIO, gpio_ref);
-	if (res == TEE_ERROR_ITEM_NOT_FOUND)
-		res = dt_driver_device_from_node_idx_prop("gpio", fdt, node, 0,
-							  DT_DRIVER_GPIO,
-							  gpio_ref);
+	res = gpio_dt_get_by_index(fdt, node, 0, NULL, &gpio);
 	if (res == TEE_ERROR_ITEM_NOT_FOUND) {
 		regu->enable_gpio = NULL;
 
--- a/core/drivers/regulator/regulator_fixed.c
+++ b/core/drivers/regulator/regulator_fixed.c
@@ -108,6 +108,10 @@ static TEE_Result get_enable_gpio(const void *fdt, int node,
 	if (cuint)
 		regu->off_on_delay = fdt32_to_cpu(*cuint);
 
+	/* Low level initialisation with updated dt_flags */
+	gpio_configure(gpio, GPIO_ASIS);
+
+	/* Configure output, don't change level */
 	gpio_set_direction(gpio, GPIO_DIR_OUT);
 
 	regu->enable_gpio = gpio;
--- a/core/drivers/regulator/regulator_gpio.c
+++ b/core/drivers/regulator/regulator_gpio.c
@@ -170,13 +170,11 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 	TEE_Result res = TEE_ERROR_GENERIC;
 	const fdt32_t *cuint = NULL;
 	struct gpio *gpio = NULL;
-	void *gpio_ref = &gpio;
 	int level0 = 0;
 	int level1 = 0;
 	int len = 0;
 
-	res = dt_driver_device_from_node_idx_prop("gpios", fdt, node, 0,
-						  DT_DRIVER_GPIO, gpio_ref);
+	res = gpio_dt_get_by_index(fdt, node, 0, NULL, &gpio);
 	if (res)
 		return res;
 
@@ -186,8 +184,7 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 	 * this implementation is simplified to support only 2 voltage
 	 * levels controlled with a single GPIO.
 	 */
-	if (dt_driver_device_from_node_idx_prop("gpios", fdt, node, 1,
-						DT_DRIVER_GPIO, gpio_ref) !=
+	if (gpio_dt_get_by_index(fdt, node, 1, NULL, &gpio) !=
 	    TEE_ERROR_ITEM_NOT_FOUND) {
 		EMSG("Multiple GPIOs not supported for level control");
 		return TEE_ERROR_GENERIC;
--- a/core/drivers/regulator/regulator_gpio.c
+++ b/core/drivers/regulator/regulator_gpio.c
@@ -157,6 +157,10 @@ static TEE_Result get_enable_gpio(const void *fdt, int node,
 	if (cuint)
 		regu->enable_delay = fdt32_to_cpu(*cuint);
 
+	/* Low level initialisation with updated dt_flags */
+	gpio_configure(gpio, GPIO_ASIS);
+
+	/* Configure output, don't change level */
 	gpio_set_direction(gpio, GPIO_DIR_OUT);
 
 	regu->enable_gpio = gpio;
@@ -174,7 +178,8 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 	int level1 = 0;
 	int len = 0;
 
-	res = gpio_dt_get_by_index(fdt, node, 0, NULL, &gpio);
+	/* GPIO to high per specification or "gpios-states" (not supported) */
+	res = gpio_dt_cfg_by_index(fdt, node, 0, NULL, GPIO_OUT_HIGH, &gpio);
 	if (res)
 		return res;
 
@@ -184,7 +189,7 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 	 * this implementation is simplified to support only 2 voltage
 	 * levels controlled with a single GPIO.
 	 */
-	if (gpio_dt_get_by_index(fdt, node, 1, NULL, &gpio) !=
+	if (gpio_dt_cfg_by_index(fdt, node, 1, NULL, GPIO_OUT_HIGH, &gpio) !=
 	    TEE_ERROR_ITEM_NOT_FOUND) {
 		EMSG("Multiple GPIOs not supported for level control");
 		return TEE_ERROR_GENERIC;
@@ -219,8 +224,6 @@ static TEE_Result get_voltage_level_gpio(const void *fdt, int node,
 		regu->voltage_levels_uv[1] = level0;
 		regu->voltage_level_high = false;
 	}
-
-	gpio_set_direction(gpio, GPIO_DIR_OUT);
 
 	regu->voltage_gpio = gpio;
 
--- a/core/drivers/stm32_gpio.c
+++ b/core/drivers/stm32_gpio.c
@@ -283,9 +283,6 @@ static void stm32_gpio_set_level(struct gpio_chip *chip, unsigned int gpio_pin,
 	if (clk_enable(bank->clock))
 		panic();
 
-	assert(((io_read32(bank->base + GPIO_MODER_OFFSET) >>
-		 (gpio_pin << 1)) & GPIO_MODE_MASK) == GPIO_MODE_OUTPUT);
-
 	if (level == GPIO_LEVEL_HIGH)
 		io_write32(bank->base + GPIO_BSRR_OFFSET, BIT(gpio_pin));
 	else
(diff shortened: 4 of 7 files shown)


core/drivers/gpio/gpio.c
```diff
@@ -29,41 +32,33 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
...
+	unsigned int i = 0;
 
-	prop_name = gpio_get_dt_prop_name(gpio_name);
-	if (!prop_name)
-		return TEE_ERROR_OUT_OF_MEMORY;
+	/* Try GPIO properties "foo-gpios" and "foo-gpio" */
+	for (i = 0; i < ARRAY_SIZE(gpio_suffixes); i++) {
+		if (gpio_name)
+			snprintf(prop_name, sizeof(prop_name), "%s-%s",
+				 gpio_name, gpio_suffixes[i]);
```
etienne-lms, line 48: `snprintf()` return value to check?  I could not find where in the DT spec property names shall be at most 32bytes (including terminal \0 char?)

core/include/drivers/gpio.h
```diff
@@ -166,16 +199,39 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
 					      int nodeoffset __unused,
 					      unsigned int index  __unused,
 					      const char *gpio_name  __unused,
+					      enum gpio_flags mode __unused,
 					      struct gpio **gpio __unused)
 {
 	return TEE_ERROR_NOT_SUPPORTED;
 }
 
+static inline gpio_configure(struct gpio *gpio, enum gpio_flags mode);
```
etienne-lms:
```suggestion
static inline TEE_Result gpio_configure(struct gpio *gpio, enum gpio_flags mode)
{
	return TEE_ERROR_NOT_SUPPORTED;
}
```
CI build test complained on that.

core/include/drivers/gpio.h
```diff
@@ -166,16 +199,39 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if GPIO controller is not initialized
+ * Return a TEE_Result compliant code in case of error
+ */
+TEE_Result gpio_dt_cfg_by_index(const void *fdt, int nodeoffset,
+				unsigned int index, const char *gpio_name,
+				enum gpio_flags mode,
+				struct gpio **gpio);
 #else
-static inline TEE_Result gpio_dt_get_by_index(const void *fdt __unused,
+static inline TEE_Result gpio_dt_cfg_by_index(const void *fdt __unused,
```
etienne-lms, line 238: Could you restore `gpio_dt_get_by_index()` inline stub definition?

core/include/drivers/gpio.h
```diff
@@ -166,16 +199,39 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
+ * gpio_dt_cfg_by_index() - Get a GPIO controller at a specific index in
+ * 'gpios' property and configure with provided flags
+ *
+ * @fdt: Device tree to work on
+ * @nodeoffset: Node offset of the subnode containing a 'gpios' property
+ * @index: GPIO pin index in '*-gpios' property
+ * @gpio_name: Prefix of the GPIO properties in device tree, can be NULL
+ * @flags: requester flags of GPIO
+ * @gpio: Output GPIO pin reference upon success
+ *
```
etienne-lms, line 220: Add a note about the property allowed to be "gpio", "gpios", "<prefix>-gpio" or "<prefix>-gpios".

core/include/drivers/gpio.h
```diff
@@ -156,7 +156,7 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
  * @fdt: Device tree to work on
  * @nodeoffset: Node offset of the subnode containing a 'gpios' property
  * @index: GPIO pin index in '*-gpios' property
- * @gpio_name: Name of the GPIO pin
+ * @gpio_name: Prefix of the GPIO properties in device tree, can be NULL
  * @gpio: Output GPIO pin reference upon success
  *
```
etienne-lms, line 198: Could you add a sentence to say that the DT property looked for can be
"gpio", "gpios", "<prefix>-gpio" or "<prefix>-gpios", as per generic DT bindings?

core/drivers/gpio/gpio.c
```diff
@@ -69,6 +69,12 @@ TEE_Result gpio_configure(struct gpio *gpio, enum gpio_flags flags)
 {
 	enum gpio_level value = GPIO_LEVEL_LOW;
 
+	assert(!gpio || (gpio->chip && gpio->chip->ops));
```
etienne-lms: It's strange to allow `gpio` to be NULL.

etienne-lms (commented):
Some preliminary comments...

core/drivers/gpio/gpio.c
tbourgoi: The limit of 32 bytes depends on the devicetree specifications : 
https://devicetree-specification.readthedocs.io/en/v0.3/devicetree-basics.html#properties
[2.2.4.1](http://2.2.4.1/). 
In linux the function of_find_gpio() has the following comment : 
`char propname[32]; /* 32 is max size of property name */`

core/include/drivers/gpio.h
tbourgoi: Yes the function should be removed as it is already defined independently of CFG_DT flags

core/drivers/gpio/gpio.c
etienne-lms: Ok, clear. Thanks.

For consistency and help maintenance, could you introduce a `DT_PROP_NAME_SIZE_MAX` macro in **core/kernel/dt.h** through a dedicated commit (which commit message could refer to URL you pointed above)?

I still think `snprintf()` return value should be checked, at least asserted.

core/drivers/gpio/gpio.c
```diff
@@ -29,43 +32,72 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
...
+}
+
+TEE_Result gpio_dt_cfg_by_index(const void *fdt, int nodeoffset,
+				unsigned int index, const char *gpio_name,
+				enum gpio_flags flags, struct gpio **gpio)
+{
+	TEE_Result res = gpio_dt_get_by_index(fdt, nodeoffset, index, gpio_name,
+					      gpio);
+	if (!res)
+		gpio_configure(*gpio, flags);
```
tbourgoi: gpio_configure() return code should be forwarded to caller.

core/drivers/gpio/gpio.c
tbourgoi: Yep, it was from gpio_put but it does not apply here

tbourgoi:
Hello, 

I've update the series with the most of the comments. Mainly to fix the CI compiling error ASAP.

core/drivers/gpio/gpio.c
tbourgoi: I'll search for a commit in Linux, 

I cannot see why it's mandatory to check the error of `snprintf()`, The function cannot overflow and if the property name is too long the search in the DT will fail. It seems to be overprotecting the code.

etienne-lms: You're right. Checking the return value is not needed.
Discard my comment. Sorry for the noise.

core/include/drivers/gpio.h
```diff
@@ -133,6 +135,39 @@ static inline void gpio_put(struct gpio *gpio)
...
+	GPIO_OUT_LOW	= GPIO_FLAGS_BIT_DIR_SET | GPIO_FLAGS_BIT_DIR_OUT,
+	GPIO_OUT_HIGH	= GPIO_FLAGS_BIT_DIR_SET | GPIO_FLAGS_BIT_DIR_OUT |
+			  GPIO_FLAGS_BIT_DIR_VAL,
+};
+
+/**
+ * gpio_configure() - Configure a GPIO controller
+ *
+ * @gpio: GPIO pin reference upon success
+ * @flags: requester flags of GPIO
```
etienne-lms, line 164: s/flags/mode/
Or maybe the a more consistent change would be to rename the argument from `mode` to `flags`.

Ditto in `gpio_dt_cfg_by_index()` inline description comment.

core/drivers/gpio/gpio.c
```diff
@@ -29,43 +32,72 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
...
+			if (flags & GPIO_FLAGS_BIT_DIR_VAL)
+				value = GPIO_LEVEL_HIGH;
+			gpio_set_value(gpio, value);
+			gpio_set_direction(gpio, GPIO_DIR_OUT);
+		} else {
+			gpio_set_direction(gpio, GPIO_DIR_IN);
+		}
+	}
+
+	return TEE_SUCCESS;
```
etienne-lms, line 94: If always successful, return value is not needed.
That said, I think it would be better to allow `configure` ops handler to return a status.

core/drivers/regulator/regulator_fixed.c
```diff
@@ -114,6 +108,10 @@ static TEE_Result get_enable_gpio(const void *fdt, int node,
 	if (cuint)
 		regu->off_on_delay = fdt32_to_cpu(*cuint);
 
+	/* Low level initialisation with updated dt_flags */
+	gpio_configure(gpio, GPIO_ASIS);
```
etienne-lms: If `gpio_configure()` returns a status, it should be checked.

etienne-lms (commented):
Thanks for the updates. I'm fine you squash any fixup commits.

core/drivers/gpio/gpio.c
```diff
@@ -29,43 +32,72 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **out_gpio)
...
 }
+
+TEE_Result gpio_configure(struct gpio *gpio, enum gpio_flags flags)
+{
+	enum gpio_level value = GPIO_LEVEL_LOW;
+
+	assert(gpio && gpio->chip && gpio->chip->ops);
+
+	/* Configure GPIO with DT flags */
+	if (gpio && gpio->chip->ops->configure)
```
clegoffic: gpio already checked in the assert above

core/drivers/stm32_gpio.c
```diff
@@ -353,6 +350,45 @@ static TEE_Result consumed_gpios_pm(enum pm_op op, unsigned int pm_hint,
...
 					      unsigned int pin);
 
+static void stm32_gpio_configure(struct gpio_chip *chip, struct gpio *gpio)
+{
+	struct stm32_gpio_bank *bank = gpio_chip_to_bank(chip);
+	uint32_t exceptions = 0;
+	uint32_t otype = 0;
+	uint32_t pupd = 0;
+	unsigned int shift_1b = gpio->pin;
+	unsigned int shift_2b = SHIFT_U32(gpio->pin, 1);
```
clegoffic, line 361: is reverse Xmas tree initialisation prefered in OP-TEE ?

core/include/drivers/gpio.h
```diff
@@ -133,6 +135,39 @@ static inline void gpio_put(struct gpio *gpio)
 		gpio->chip->ops->put(gpio->chip, gpio);
 }
 
+#define GPIO_FLAGS_BIT_DIR_SET		BIT(0)
+#define GPIO_FLAGS_BIT_DIR_OUT		BIT(1)
+#define GPIO_FLAGS_BIT_DIR_VAL		BIT(2)
+
+/**
+ * enum gpio_flags - Optional flags that used to configure direction and output
```
clegoffic: s/that used/that are used/

core/include/drivers/gpio.h
```diff
@@ -133,6 +135,39 @@ static inline void gpio_put(struct gpio *gpio)
...
+ *                   value. These values cannot be OR'd.
+ *
+ * @GPIOD_ASIS:		Don't change anything
+ * @GPIO_IN:		Set lines to input mode
+ * @GPIO_OUT_LOW:	Set lines to output and drive them low
+ * @GPIO_OUT_HIGH:	Set lines to output and drive them high
+
```
clegoffic: It seems the configuration in gpio_configure is done per line so maybe remove the extra 's' and s/them/it/

core/include/drivers/gpio.h
```diff
@@ -133,6 +135,39 @@ static inline void gpio_put(struct gpio *gpio)
...
+	GPIO_IN		= GPIO_FLAGS_BIT_DIR_SET,
+	GPIO_OUT_LOW	= GPIO_FLAGS_BIT_DIR_SET | GPIO_FLAGS_BIT_DIR_OUT,
+	GPIO_OUT_HIGH	= GPIO_FLAGS_BIT_DIR_SET | GPIO_FLAGS_BIT_DIR_OUT |
+			  GPIO_FLAGS_BIT_DIR_VAL,
+};
+
+/**
+ * gpio_configure() - Configure a GPIO controller
+ *
+ * @gpio: GPIO pin reference upon success
```
clegoffic: It seems @ gpio always refer to a GPIO pin reference why is it documented with "upon success" ?

core/include/drivers/gpio.h
```diff
@@ -151,12 +186,14 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
  * gpio_dt_get_by_index() - Get a GPIO controller at a specific index in
- * 'gpios' property
+ * 'gpios' or 'gpio' properties with or without suffixes.
  *
  * @fdt: Device tree to work on
- * @nodeoffset: Node offset of the subnode containing a 'gpios' property
- * @index: GPIO pin index in '*-gpios' property
- * @gpio_name: Name of the GPIO pin
+ * @nodeoffset: Node offset of the subnode containing a 'gpios' or 'gpio'
+ *		property.
```
clegoffic, line 193: s/property/prefixed property/ if gpios and gpio can have suffixes

core/include/drivers/gpio.h
```diff
@@ -166,11 +203,44 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
+ * @index: GPIO pin index in 'gpios' property find in device tree.
+ * @gpio_name: Prefix of a '-gpios' or '-gpio' properties in device tree,
+ *	       can be NULL to search for 'gpios' or 'gpio' properties
+ * @flags: requester flags of GPIO
+ * @gpio: Output GPIO pin reference upon success
+ *
+ * Return TEE_SUCCESS in case of success
+ * Return TEE_ERROR_DEFER_DRIVER_INIT if GPIO controller is not initialized
+ * Return a TEE_Result compliant code in case of error
```
clegoffic, line 223: It seems TEE_ERROR_ITEM_NOT_FOUND can also be returned, I imagine it enters in the case of the "Return a TEE_Result compliant code in case of error" but why TEE_ERROR_DEFER_DRIVER_INIT error code has a specific line ?
Either refine or delete the special case for TEE_ERROR_DEFER_DRIVER_INIT no ?

Upsylonbare (commented):
Hi Thomas, here are some comments, hope they will be interesting for you.
You can close the conversation if some are irrelevant

core/drivers/gpio/gpio.c
tbourgoi: Yes but assert() are disabled when OP-TEE is compiled in release mode. 
One of the good practice in C is to check a pointer before using it so I want to keep the `if(gpio)`.
IMO it's especially important in a framework.
The assert is here to check during development that the GPIO structure is correctly generated.
So I want to keep both lines as is

core/drivers/stm32_gpio.c
tbourgoi: Not to my knowledge and I cannot find any reference to this in https://optee.readthedocs.io/en/latest/general/coding_standards.html
gpio.c was not following this rule before so I don't see any reason to follow it now. (cf function `gpio_get_dt_prop_name()`)

core/include/drivers/gpio.h
tbourgoi: This modification looks too restrictive too me, the function is also searching for `gpio` (and `gpios`) non prefixed property as the sumarry of the function explain it so I don't want to add `prefixed` here

core/include/drivers/gpio.h
tbourgoi: I'm sure many other TEE_Result might be returned since we are using the device tree lib... This will make explaining all return codes impossible and irrelevant.
Since the function search for gpio pin, if it returns TEE_ERROR_ITEM_NOT_FOUND it's not a big assumption to suppose that we failed to find the gpio pin we were looking for. 
Other function are doing the same (ex : `regulator_dt_register()` and in gpio.h)

core/drivers/gpio/gpio.c
tbourgoi: I have changed to forward the return value of configure ops.

tbourgoi:
Hello @jenswi-linaro @jforissier 

If you have any comments on the PR, I'm open to have a feedback.
I can squash fixup as @etienne-lms suggested if it's easier for you to read the PR.

Thanks in advance,

Thomas

jenswikl:
Please squash the fixups. I can take a quick look, but I think that we need @etienne-lms' feedback before merging this.

core/include/drivers/gpio.h
```diff
@@ -151,12 +151,14 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
 
 /**
  * gpio_dt_get_by_index() - Get a GPIO controller at a specific index in
- * 'gpios' property
+ * 'gpios' or 'gpio' properties with or without suffixes.
```
etienne-lms: s/suffixes/prefixes/

core/drivers/gpio/gpio.c
```diff
@@ -64,3 +64,34 @@ TEE_Result gpio_dt_get_by_index(const void *fdt, int nodeoffset,
...
+	return TEE_SUCCESS;
+}
+
+TEE_Result gpio_dt_cfg_by_index(const void *fdt, int nodeoffset,
+				unsigned int index, const char *gpio_name,
+				enum gpio_flags flags, struct gpio **gpio)
+{
+	TEE_Result res = gpio_dt_get_by_index(fdt, nodeoffset, index, gpio_name,
+					      gpio);
+	if (!res)
```
etienne-lms, line 104: Add an empty line between local variable definition and in 1st instruction.

Maybe the below looks better, but that's only my feeling.
```c

	res = gpio_dt_get_by_index(fdt, nodeoffset, index, gpio_name, gpio);
	if (res)
		return res;

	return gpio_configure(*gpio, flags);

File: core/drivers/regulator/regulator_gpio.c [hunk 20]
```diff
@@ -157,6 +157,10 @@ static TEE_Result get_enable_gpio(const void *fdt, int node,
 	if (cuint)
 		regu->enable_delay = fdt32_to_cpu(*cuint);
 
+	/* Low level initialisation with updated dt_flags */
+	gpio_configure(gpio, GPIO_ASIS);
```
etienne-lms: Must test the return value.

core/include/drivers/gpio.h
```diff
@@ -133,6 +133,39 @@ static inline void gpio_put(struct gpio *gpio)
...
+ * @GPIO_OUT_HIGH:	Set line to output and drive it high
+
+ */
+enum gpio_flags {
+	GPIO_ASIS	= 0,
+	GPIO_IN		= GPIO_FLAGS_BIT_DIR_SET,
+	GPIO_OUT_LOW	= GPIO_FLAGS_BIT_DIR_SET | GPIO_FLAGS_BIT_DIR_OUT,
+	GPIO_OUT_HIGH	= GPIO_FLAGS_BIT_DIR_SET | GPIO_FLAGS_BIT_DIR_OUT |
+			  GPIO_FLAGS_BIT_DIR_VAL,
+};
```
etienne-lms: Prefer have all macros and enums at header file top: first macros, then enums and structs definitions, and last functions declarations and inline functions definitions. That's said it not a strong opinion.

core/include/drivers/gpio.h
```diff
@@ -168,11 +201,43 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
+
+/**
+ * gpio_dt_cfg_by_index() - Get a GPIO controller at a specific index in
+ * 'gpios' or 'gpio' properties with or without suffixes configure with provided
+ * flags.
+ *
+ * @fdt: Device tree to work on
+ * @nodeoffset: Node offset of the subnode containing a 'gpios' or 'gpio'
+ *		property.
+ * @index: GPIO pin index in 'gpios' property find in device tree.
```
etienne-lms: s/find/found/

core/include/drivers/gpio.h
```diff
@@ -168,11 +201,43 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
 TEE_Result gpio_dt_get_by_index(const void *fdt, int nodeoffset,
 				unsigned int index, const char *gpio_name,
 				struct gpio **gpio);
+
+/**
+ * gpio_dt_cfg_by_index() - Get a GPIO controller at a specific index in
+ * 'gpios' or 'gpio' properties with or without suffixes configure with provided
```
etienne-lms:
```suggestion
 * 'gpios' or 'gpio' properties (possibly prefixed) and configure it with
 * provided @flags.
```

core/include/drivers/gpio.h
```diff
@@ -168,11 +201,43 @@ TEE_Result gpio_dt_alloc_pin(struct dt_pargs *pargs, struct gpio **gpio);
...
+ * flags.
+ *
+ * @fdt: Device tree to work on
+ * @nodeoffset: Node offset of the subnode containing a 'gpios' or 'gpio'
+ *		property.
+ * @index: GPIO pin index in 'gpios' property find in device tree.
+ * @gpio_name: Prefix of a '-gpios' or '-gpio' properties in device tree,
+ *	       can be NULL to search for 'gpios' or 'gpio' properties
+ * @flags: requester flags of GPIO
```
etienne-lms: `Configuration flags for the GPIO` would better match the description above.

core/drivers/gpio/gpio.c
etienne-lms: Calling `gpio_configure()` with a NULL gGPIO reference is a programming error so I think it's prefectly find if OP-TEE crashes (segfault) at this point.

core/drivers/gpio/gpio.c
```diff
@@ -68,6 +68,13 @@ TEE_Result gpio_dt_get_by_index(const void *fdt, int nodeoffset,
 TEE_Result gpio_configure(struct gpio *gpio, enum gpio_flags flags)
 {
 	enum gpio_level value = GPIO_LEVEL_LOW;
+	TEE_Result res = TEE_SUCCESS;
+
+	assert(gpio && gpio->chip && gpio->chip->ops);
+
+	/* Configure GPIO with DT flags */
+	if (gpio && gpio->chip->ops->configure)
+		res = gpio->chip->ops->configure(gpio->chip, gpio);
```
etienne-lms, line 77: Return `res` upon failure without apply direction/value configuration.

core/include/drivers/gpio.h
tbourgoi: IMO If I read `found` I would interpret that `index` is a return value of the function (thanks to a pointer)
I propose s/find/to get/

core/include/drivers/gpio.h
tbourgoi: Done

tbourgoi:
All comment taken into account, I'll apply review tag on all commit at once when squashing fixup

etienne-lms:
LGTM. `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits
"drivers: gpio: add configure functions" and
"drivers: gpio: add new ops configure to initialize gpio".

*CI / make check (QEMUv8, Hafnium)* is likely unrelated:
```
2025-08-14T10:09:20.9280926Z fetch-pack: unexpected disconnect while reading sideband packet
2025-08-14T10:09:20.9281578Z fatal: early EOF
2025-08-14T10:09:20.9299628Z fatal: fetch-pack: invalid index-pack output
2025-08-14T10:09:20.9303621Z fatal: Fetched in submodule path 'prebuilts', but it did not contain 03e6ef3b3c0aa74e87ff68264248777d367a3773. Direct fetching of that commit failed.
2025-08-14T10:09:20.9310149Z make: *** [Makefile:458: .hafnium_checkout] Error 128
```

jforissier:
@tbourgoi it looks like this is almost ready for merging, can you please squash the fixups and apply the review tags? Thanks.

tbourgoi:
> @tbourgoi it looks like this is almost ready for merging, can you please squash the fixups and apply the review tags? Thanks.

Hello, yes no problem, I'll update the PR next week with the review tags

tbourgoi:
Hello, 
I've back ported the series on our dowstream to run a broader test suit. 
During tests, we discovered that some SD card are failing because the regulator GPIO are initialised with the high value by default.
The value to use during initialisation of the regulator depends on the board.
To handle this, the Linux kernel is using the property `gpios-states`.
So I added a fixup to the series to implement the property `gpios-states` to support all SDcard.

I'm sorry for the late discovery, 

Thomas

jforissier:
@tbourgoi no worries. Catching issues early with more testing is always the preferred option :+1:

etienne-lms (commented):
For commit "fixup! drivers: gpio: add configure functions":
~~LGTM aside the commit message: could you refine the header line and body and remove all tags but the S-o-b one.~~

(edited): my mistake, indeed this is a fixup for the 1st commit of the series. All good once squashed. My R-b tag still applies that that commit once squashed.

tbourgoi:
I squash the fixup

PR #7481, opened 2025-07-16
Several ASan fixes

alexhoppus:
Hi,

This PR introduces two improvements related to AddressSanitizer (ASan) support:

- Redzones after global variables
Adds redzones after global variables, particularly benefiting `.rodata` variables. The shadow memory related to `.rodata` section is zeroed out in `init_asan()` (what is not true for `.data` or `.bss`), which can hide certain read overflows for constant variables whose size is aligned to `ASAN_BLOCK`. (see test-case)

- Disables certain optimized string functions in newlib under ASan
Disables the use of some optimized string manipulation functions from newlib when ASan is enabled. These versions use the `DETECTNULL` macro, which performs out-of-bounds reads. While this is generally safe due to word-aligned buffer starts, it causes ASan to panic. Notably, the issue was reproducible only with `-O0` builds (no optimization).

### DIFF
--- a/core/kernel/asan.c
+++ b/core/kernel/asan.c
@@ -411,11 +411,17 @@ void __asan_handle_no_return(void)
 void __asan_register_globals(struct asan_global *globals, size_t size);
 void __asan_register_globals(struct asan_global *globals, size_t size)
 {
-	size_t n;
+	size_t n = 0;
 
-	for (n = 0; n < size; n++)
-		asan_tag_access((void *)globals[n].beg,
-				(void *)(globals[n].beg + globals[n].size));
+	for (n = 0; n < size; n++) {
+		vaddr_t begin = globals[n].beg;
+		vaddr_t end = begin + globals[n].size;
+		vaddr_t end_align = ROUNDUP(end, ASAN_BLOCK_SIZE);
+		vaddr_t end_rz = begin + globals[n].size_with_redzone;
+
+		asan_tag_access((void *)begin, (void *)end);
+		asan_tag_no_access((void *)end_align, (void *)end_rz);
+	}
 }
 DECLARE_KEEP_INIT(__asan_register_globals);
 
--- a/core/pta/tests/misc.c
+++ b/core/pta/tests/misc.c
@@ -667,6 +667,7 @@ static int self_test_va2pa(void)
 
 static char asan_test_sgbuf[ASAN_TEST_BUF_SIZE];
 char asan_test_gbuf[ASAN_TEST_BUF_SIZE];
+static const char asan_test_sgbuf_ro[ASAN_TEST_BUF_SIZE + 1];
 
 static jmp_buf asan_test_jmp;
 
@@ -675,6 +676,7 @@ struct asan_test_ctx {
 	char *pmalloc2[3];
 	char write_value;
 	void (*write_func)(char *buf, size_t pos, char value);
+	char (*read_func)(char *buf, size_t pos);
 	void *(*memcpy_func)(void *__restrict dst,
 			     const void *__restrict src, size_t size);
 	void *(*memset_func)(void *buf, int val, size_t size);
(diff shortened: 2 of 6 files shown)


core/kernel/asan.c
```diff
@@ -413,9 +413,16 @@ void __asan_register_globals(struct asan_global *globals, size_t size)
 {
 	size_t n;
```
jenswikl: Since you're changing this function, please zero-initialize `n`.

core/kernel/asan.c
```diff
@@ -413,9 +413,16 @@ void __asan_register_globals(struct asan_global *globals, size_t size)
...
 
-	for (n = 0; n < size; n++)
-		asan_tag_access((void *)globals[n].beg,
-				(void *)(globals[n].beg + globals[n].size));
+	for (n = 0; n < size; n++) {
+		vaddr_t end = globals[n].beg + globals[n].size,
```
jenswikl: Please declare each local variable in separate statements, that is, end this line with `;` and do the same for the other local variables.

alexhoppus: Done

core/kernel/asan.c
alexhoppus: Done

core/kernel/asan.c
```diff
@@ -411,11 +411,18 @@ void __asan_handle_no_return(void)
 void __asan_register_globals(struct asan_global *globals, size_t size);
 void __asan_register_globals(struct asan_global *globals, size_t size)
 {
-	size_t n;
+	size_t n = 0;
```
jenswikl, line 414: The purpose of this initialization is to make sure that whatever mistake we make, at least it has a well-defined value.

I prefer `for (n = 0;...)` below since that's what we had before.

core/kernel/asan.c
```diff
@@ -411,11 +411,18 @@ void __asan_handle_no_return(void)
...
+
+	for (; n < size; n++) {
+		vaddr_t end = globals[n].beg + globals[n].size;
+		vaddr_t end_align = ROUNDUP(end, ASAN_BLOCK_SIZE);
+		vaddr_t end_rz = globals[n].beg
+			+ globals[n].size_with_redzone;
```
jenswikl: Please end lines with the operator.

core/kernel/asan.c
```diff
@@ -411,11 +411,18 @@ void __asan_handle_no_return(void)
...
+		vaddr_t begin = globals[n].beg;
```
jenswikl, line 417: Since you've added this helper variable, why don't you move it up a bit so it can be used in the other expressions?

core/kernel/asan.c
alexhoppus: Done

core/kernel/asan.c
alexhoppus: Done

core/kernel/asan.c
alexhoppus: Done

alexhoppus:
Hi, @jenswi-linaro 

- the CI fails probably out of space issue

> 89755 2025-08-11T17:51:56.0987582Z Copying files into the device: Image Name:   Root file system
89756 2025-08-11T17:51:56.0988301Z Created:      Mon Aug 11 17:51:55 2025        
89757 2025-08-11T17:51:56.0988853Z Image Type:   AArch64 Linux RAMDisk Image (gzip compressed)
89758 2025-08-11T17:51:56.0989457Z Data Size:    48217962 Bytes = 47087.85 KiB = 45.98 MiB
89759 2025-08-11T17:51:56.0989928Z Load Address: 45000000                        
89760 2025-08-11T17:51:56.0990282Z Entry Point:  45000000                        
89761 2025-08-11T17:51:56.3041527Z __populate_fs: Could not allocate block in ext2 filesystem while writing file "rootfs.cpio.gz"
89762 2025-08-11T17:51:56.3042456Z mke2fs: Could not allocate block in ext2 filesystem while populating file system
89763 2025-08-11T17:51:56.3044811Z make: *** [Makefile:545: xen-create-image] Error 1
89764 2025-08-11T17:51:56.3103801Z ##[error]Process completed with exit code 2.  

- I squash everything since changes are minor, let me know if I need keep changes in separate commits

jenswikl:

alexhoppus:
Tags applied

jforissier:
Thanks @alexhoppus

PR #7483, opened 2025-07-17
core: kernel: align the address of transfer entry

raymo200915:
When adding a new transfer entry, align the address of the header with TRANSFER_LIST_GRANULE.
This solves the behaviour gap between how a transfer entry is added in transfer_list_add() and how it is enumerated in transfer_list_next().

### DIFF
--- a/core/kernel/transfer_list.c
+++ b/core/kernel/transfer_list.c
@@ -474,20 +474,19 @@ struct transfer_list_entry *transfer_list_add(struct transfer_list_header *tl,
 
 	max_tl_ev = (vaddr_t)tl + tl->max_size;
 	tl_ev = (vaddr_t)tl + tl->size;
-	ev = tl_ev;
+	if (ROUNDUP_OVERFLOW(tl_ev, TRANSFER_LIST_GRANULE, &ev))
+		return NULL;
+
+	tl_e = (struct transfer_list_entry *)ev;
 
 	/*
 	 * Skip the step 1 (optional step).
 	 * New transfer entry will be added into the tail
 	 */
 	if (ADD_OVERFLOW(sizeof(*tl_e), data_size, &sz) ||
-	    ADD_OVERFLOW(ev, sz, &ev) ||
-	    ROUNDUP_OVERFLOW(ev, TRANSFER_LIST_GRANULE, &ev) ||
-	    ev > max_tl_ev) {
+	    ADD_OVERFLOW(ev, sz, &ev) || ev > max_tl_ev)
 		return NULL;
-	}
 
-	tl_e = (struct transfer_list_entry *)tl_ev;
 	*tl_e = (struct transfer_list_entry){
 		.tag_id = tag_id,
 		.hdr_size = sizeof(*tl_e),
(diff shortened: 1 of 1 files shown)


core/kernel/transfer_list.c
```diff
@@ -474,7 +474,10 @@ struct transfer_list_entry *transfer_list_add(struct transfer_list_header *tl,
 
 	max_tl_ev = (vaddr_t)tl + tl->max_size;
 	tl_ev = (vaddr_t)tl + tl->size;
-	ev = tl_ev;
+	if (ROUNDUP_OVERFLOW(tl_ev, TRANSFER_LIST_GRANULE, &ev))
+		return NULL;
+
+	tl_e = (struct transfer_list_entry *)ev;
```
jenswikl, line 480: Spurious change?

raymo200915: The alignment step is intentional. `transfer_list_next()` aligns each entry before reading, so we need to ensure that each new entry added by `transfer_list_add()` is also aligned. Before patching, `tl_e` was set directly to `tl_ev`(`tl + tl->size`), which may be unaligned. This created a mismatch between how entries are written and read.

The first `ROUNDUP_OVERFLOW()` ensures `tl_e` starts at an aligned address before we compute its size and update `tl->size`. The later `ROUNDUP_OVERFLOW()` is checking the total size after adding does not exceed the max boundary, so both are needed.

raymo200915:
Added same fix to `transfer_list_add_with_align()` when computing the TE start address.
Removed the potential roundup padding (the later `ROUNDUP_OVERFLOW()`) at the TL tail, tl->size should reflect the current actual TL size without roundup padding at the tail.

core/kernel/transfer_list.c
raymo200915: [rev] The later `ROUNDUP_OVERFLOW()` is not needed. The alignment requirement is only for the start address of a TE, not the end of it.

jenswikl: Sorry for being unclear. Did you move the `tl_e = (struct transfer_list_entry *)tl_ev` assignment from line 490 to here on purpose?

raymo200915: They are not exactly the same:
`tl_e = (struct transfer_list_entry *)ev;` v.s
`tl_e = (struct transfer_list_entry *)tl_ev;` - `ev` is the rounded up `tl_ev`.
I have to move it ahead since `ev` is being reused to calculate the new end of TL from line 486 to 489.

raymo200915:
Aquashed the fixes.

core/kernel/transfer_list.c
jenswikl: Got it, thanks.

raymo200915:
> The diff looks good, but I'd like to have a clear commit message. This part is too vague:
> 
> > This solves the behaviour gap between how a transfer entry is added
> > via transfer_list_add() or transfer_list_add_with_align()  and how
> > it is enumerated in transfer_list_next().
> 
> > Moreover, tl->size should reflect the actual TL size without padding
> > at the tail.
> 
> There are two problems fixed in `transfer_list_add()`:
> 
> 1. avoid adding entries at an unaligned address
> 2. don't include eventual implicit padding at the end of an entry in the size of the entry.
> 
> Please add a fixes tag identifying the commit that introduced the problem.

Done.

raymo200915:
Commit message updated.

jenswikl:
Since there is more than one version of the spec, please mention the version you're using.

raymo200915:
> Since there is more than one version of the spec, please mention the version you're using.

Done.

raymo200915:
Updated the commit message to refer to Firmware Handoff spec v1.0

jenswikl:

raymo200915:
Review tag is added.

PR #7484, opened 2025-07-18
plat-versal2: use auto PA bit discovery

Akshay-Belsare:
Removes hardcoded configuration for large physical address and ARM64 PA bits, enabling automatic discovery of the maximal PA supported by the hardware.


Akshay-Belsare:
@jforissier  tag applied!

Issue #7482, opened 2025-07-17
The optee.ta.instanceKeepCrashed property doesn't work as expected.


joseph-lo-nvtw:
The new introduced optee.ta.instanceKeepCrashed property in OP-TEE v4.7.0 doesn't work as expected.

With the two new patches below:
https://github.com/OP-TEE/optee_os/commit/941a58d78c99c4754fbd4ec3079ec9e1d596af8f
https://github.com/OP-TEE/optee_ftpm/commit/ce33372ab772e879826361a1ca91126260bd9be1

The fTPM TA still can be restarted after it panicked.

The issue of that is the "release_ta_ctx" function doesn't take care of these properties. It removes the panicked TA from the tee_ctxes list.
https://github.com/OP-TEE/optee_os/blob/master/core/kernel/tee_ta_manager.c#L666

So, when we re-open the TA, since it's not in the list. The "tee_ta_init_session_with_context" couldn't find it and the "tee_ta_open_session" couldn't mantain the paniced ctx in TEE_ERROR_TARGET_DEAD state.
https://github.com/OP-TEE/optee_os/blob/master/core/kernel/tee_ta_manager.c#L720

The TA still can be reloaded with a new clean state. So the issue remains the same.
https://github.com/OP-TEE/optee_os/security/advisories/GHSA-f35r-hm2m-p6c3

I did a experience of the code below which make the paniced fTPM TA won't be respawned.

```
static void release_ta_ctx(struct tee_ta_ctx *ctx)
{
	bool was_releasing = false;
	bool keep_crashed = false;
	bool keep_alive = false;

	if (ctx->flags & TA_FLAG_SINGLE_INSTANCE)
		keep_alive = ctx->flags & TA_FLAG_INSTANCE_KEEP_ALIVE;
	if (keep_alive)
		keep_crashed = ctx->flags & TA_FLAG_INSTANCE_KEEP_CRASHED;

	if ((ctx->panicked && !keep_crashed) || !keep_alive) {
		mutex_lock(&tee_ta_mutex);
		was_releasing = ctx->is_releasing;
		ctx->is_releasing = true;
		if (!was_releasing) {
			DMSG("Releasing panicked TA ctx");
			TAILQ_REMOVE(&tee_ctxes, ctx, link);
		}
		mutex_unlock(&tee_ta_mutex);

		if (!was_releasing) {
			ctx->ts_ctx.ops->release_state(&ctx->ts_ctx);
			destroy_context(ctx);
		}
[...]

PR #7485, opened 2025-07-21
core: maintain the panicked TAs with instanceKeepCrashed property

joseph-lo-nvtw:
The panicked TAs with TA_FLAG_SINGLE_INSTANCE,
TA_FLAG_INSTANCE_KEEP_ALIVE, and TA_FLAG_INSTANCE_KEEP_CRASHED flags
should be kept in the TA context list to maintain their paincked status
and prevent respawning.

### DIFF
--- a/core/kernel/tee_ta_manager.c
+++ b/core/kernel/tee_ta_manager.c
@@ -2,6 +2,7 @@
 /*
  * Copyright (c) 2014, STMicroelectronics International N.V.
  * Copyright (c) 2020, Arm Limited
+ * Copyright (c) 2025, NVIDIA Corporation & AFFILIATES.
  */
 
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


joseph-lo-nvtw:
PR for fixing https://github.com/OP-TEE/optee_os/issues/7482.

core/kernel/tee_ta_manager.c
```diff
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: BSD-2-Clause
 /*
- * Copyright (c) 2014, STMicroelectronics International N.V.
- * Copyright (c) 2020, Arm Limited
+ * SPDX-FileCopyrightText: Copyright (c) 2014, STMicroelectronics International N.V.
+ * SPDX-FileCopyrightText: Copyright (c) 2020, Arm Limited
+ * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA Corporation & AFFILIATES.
```
jforissier: Please drop the SPDX tags.

core/kernel/tee_ta_manager.c
```diff
@@ -666,18 +667,32 @@ static TEE_Result tee_ta_init_session(TEE_ErrorOrigin *err,
...
+		keep_alive = ctx->flags & TA_FLAG_INSTANCE_KEEP_ALIVE;
+	if (keep_alive)
+		keep_crashed = ctx->flags & TA_FLAG_INSTANCE_KEEP_CRASHED;
+
+	/*
+	 * Keep panicked TAs with SINGLE_INSTANCE, KEEP_ALIVE, and KEEP_CRASHED
+	 * flags in the context list to maintain their panicked status and
+	 * prevent respawning.
+	 */
+	if ((ctx->panicked && !keep_crashed) || !keep_alive) {
```
jforissier: - This function is only ever called when `ctx->panicked` is true, so this check is not needed. It should be replaced with an `assert()`.
- Why test `keep_alive` here?
- Since the context is not always released, I suggest renaming the function `maybe_release_ta_ctx()`

core/kernel/tee_ta_manager.c
joseph-lo-nvtw: Removed, thanks for review.

core/kernel/tee_ta_manager.c
joseph-lo-nvtw: > Why test keep_alive here?

Yes, with line 673 to 676, we should be able to check keep_crashed only.

Thanks for review.

jenswikl:
Please add a fixes tag `Fixes: 941a58d78c99 ("Add optee.ta.instanceKeepCrashed property")` to the commit message. 
With that:

jforissier:

joseph-lo-nvtw:
Patch updated, thanks for review.

PR #7486, opened 2025-07-21
Enable ELE support on i.MX95

sahilnxp:
This pull request does following:

- Update imx_mu_8ulp.c to imx_8ulp_9x.c to give more meaningful name to this file.
- Enable ELE support for i.MX95.
- Update get_info_rsp{} struct for i.MX95.
- Due to introduction of dual firmware, remove sab_init() initialization.


core/drivers/crypto/ele/ele.c
```diff
@@ -375,7 +375,7 @@ int tee_otp_get_die_id(uint8_t *buffer, size_t len)
 /*
  * Initialize EdgeLock Enclave services
  */
-static TEE_Result imx_ele_sab_init(void)
+static TEE_Result __maybe_unused imx_ele_sab_init(void)
```
jenswikl, line 378: This function is never called. Do we need to keep it?

sahilnxp: Will remove this function completely.

sahilnxp:
Can we change the status of this pull request to open please ?

jenswikl:
The reopen button is greyed out for me.

jforissier:
> The reopen button is greyed out for me.

Same here :-/ there is a tooltip saying "The github_master_imx95_ele_branch was force-pushed or recreated.". Not sure it is related to the fact we can't reopen. In any case @sahilnxp I think the only option is to open a new PR.

sahilnxp:
Thanks @jenswi-linaro  and @jforissier .
Will raise a new PR then.

sahilnxp:
New PR: https://github.com/OP-TEE/optee_os/pull/7539

PR #7497, opened 2025-08-04
Fix potential panic in trace / print handler

jsetje:
Hi, I have not contributed to this work in the past, so this is a relatively simple patch to get used to any project
specific style or procedure requirements. I did run checkpatch from a recent linux kernel tree and I am familiar with
DCO. The code builds and works for me.

While porting to a new platform I ended up with some broken mappings during dom0less NS init and serial_console->ops ended up unmapped. This meant the DMSG() in do_init_calls() would invariably panic at that time. This adds a check to avoid trying to call the unmapped ops vector.

Without this check the result is a recursive panic in the console output code with no output, which can be tedious to debug on actual hardware.

If everything is located in the right place and there is sufficient VA available, this should never happen. So this is really just a quality of life improvement for someone getting optee to work in a new situation.

### DIFF
--- a/core/kernel/console.c
+++ b/core/kernel/console.c
@@ -36,7 +36,8 @@ void console_init(void)
 
 void __weak console_putc(int ch)
 {
-	if (!serial_console)
+	if (!serial_console || !serial_console->ops ||
+	    !serial_console->ops->putc)
 		return;
 
 	if (ch == '\n')
@@ -46,7 +47,8 @@ void __weak console_putc(int ch)
 
 void __weak console_flush(void)
 {
-	if (!serial_console || !serial_console->ops->flush)
+	if (!serial_console || !serial_console->ops ||
+	    !serial_console->ops->flush)
 		return;
 
 	serial_console->ops->flush(serial_console);


jsetje:
Thank you for any and all comments, even if it's just to get the port right on the first try and close this PR.

core/kernel/console.c
```diff
@@ -36,7 +36,8 @@ void console_init(void)
 
 void __weak console_putc(int ch)
 {
-	if (!serial_console)
+	if (!serial_console || !serial_console->ops ||
+	    !serial_console->ops->putc)
```
ivila, line 40: Given that `putc` is mandatory in the design, perhaps we can omit the check of `putc`.

https://github.com/OP-TEE/optee_os/blob/61150e5450956dec327375ec120e31210774c991/core/include/drivers/serial.h#L18-L27

ivila:
Consider adding comments or conditional compilation flags to clarify that those checks is intended for debugging purposes; otherwise, people may wonder why the code checks whether the only const struct field (`ops`) in the static struct (`serial_console`) is `NULL`. Normally, we would simply set the static struct itself to `NULL`, but certain edge cases can occur during porting.

core/kernel/console.c
jsetje: Hah, thank you. I was wondering why that didn't match flush, that explains it. I'll test once I reproduce the failure again.

jsetje:
I can try to reproduce the failure now that I have the port working so I can articulate exactly what led to it. That should help understand how interesting this is to have on by default. (I have a full day of meetings today, I may or may not get to that today.)

jsetje:
Of course now I can't reproduce this failure mode.  - Sigh. Of course lots of things have moved around a bit since I hit this.

I am able to build with a CFG_TEE_RAM_VA_SIZE that's just slightly larger than the build time assertion checks for and get that to break with a perfectly reasonable panic. Which seems fine, that's a boneheaded thing to do anyway unless you're trying to force a failure.

I'm starting to think that I just got spectacularly unlucky. I will try to reproduce this a bit more. I'm closing the PR for now.

I'll re-open if needed, but there's a good chance that whatever state I got this into should be covered by a build time assertion instead.

PR #7498, opened 2025-08-05
libutee: Fix init handling

johankor:
Previous PR: https://github.com/OP-TEE/optee_os/pull/7443


lib/libutee/user_ta_entry.c
```diff
@@ -224,10 +233,10 @@ static TEE_Result ta_header_add_session(uint32_t session_id)
 		return TEE_SUCCESS;
 
 	if (!init_done) {
-		init_done = true;
 		res = init_instance();
 		if (res)
 			return res;
+		init_done = true;
```
jforissier: Perhaps we could remove the `init_done` test here and always call:
```C
	res = init_instance();
	if (res)
		return res;

	itr = TEE_Malloc(sizeof(struct ta_session), ...
```

Then both booleans could be local to `init_instance()` and the flow would be a bit simpler.

lib/libutee/user_ta_entry.c
```diff
@@ -167,8 +167,14 @@ static unsigned int get_memtag_implementation(void)
 
 static TEE_Result init_instance(void)
 {
+	static bool init_elf_done;
+
 	trace_set_level(tahead_get_trace_level());
 	__utee_gprof_init();
+
+	if (init_elf_done)
```
jforissier: Why not cover the `trace_set_level()` and `__utee_gprof_init()` calls?

lib/libutee/user_ta_entry.c
```diff
@@ -167,8 +167,14 @@ static unsigned int get_memtag_implementation(void)
 
 static TEE_Result init_instance(void)
 {
+	static bool init_elf_done;
```
jforissier: Assuming the below suggestion, how about:
```C
static boot init_done;
static bool create_entrypoint_done;
```

jforissier (commented):
Hi @johankor, thanks for the contribution! Please see below.

lib/libutee/user_ta_entry.c
```diff
@@ -198,6 +198,7 @@ static TEE_Result init_instance(void)
 static void uninit_instance(void)
 {
 	__utee_gprof_fini();
+	init_done = false;
```
jforissier: This one I don't understand. After uninit_instance() the TA is supposed to be unloaded. The next session open should cause a new load and init from scratch.

johankor: Apologies @jforissier, in that case this does indeed not make any sense.

Could you give me a pointer on how I can read this from the code? What I mean is that it's not obvious to me how keep-alive TAs persist, while non keep-alive are unloaded. The only difference I can see is the `uninit_instance()` call for non keep-alive TAs, but how exactly does this cause an unload (and hence reset of `init_done`)?

jforissier: This is handled outside of libutee, in the TEE core. See https://github.com/OP-TEE/optee_os/blob/61150e5450956dec327375ec120e31210774c991/core/kernel/tee_ta_manager.c#L452. The `keep_alive` boolean and `destroy_context()`.

lib/libutee/user_ta_entry.c
johankor: Fixed.

lib/libutee/user_ta_entry.c
johankor: Fixed.

lib/libutee/user_ta_entry.c
johankor: Agreed! Fixed.

lib/libutee/user_ta_entry.c
johankor: Thanks! I've dropped this commit now.

lib/libutee/user_ta_entry.c
```diff
@@ -167,6 +165,16 @@ static unsigned int get_memtag_implementation(void)
 
 static TEE_Result init_instance(void)
 {
+	static bool create_entrypoint_done;
```
jforissier: Let's call this `internal_init_done`

lib/libutee/user_ta_entry.c
```diff
@@ -167,6 +165,16 @@ static unsigned int get_memtag_implementation(void)
...
+	static bool init_done;
+	TEE_Result res = TEE_SUCCESS;
+
+	if (init_done)
+		goto init_done;
+
+	if (create_entrypoint_done)
+		goto create_entrypoint_done;
```
jforissier:
```suggestion
	if (internal_init_done)
		goto create_entrypoint;
```

lib/libutee/user_ta_entry.c
```diff
@@ -183,7 +191,16 @@ static TEE_Result init_instance(void)
 	_TEE_MathAPI_Init();
 	__utee_tcb_init();
 	__utee_call_elf_init_fn();
-	return TA_CreateEntryPoint();
+
+create_entrypoint_done:
+	create_entrypoint_done = true;
```
jforissier:
```suggestion
	internal_init_done = true;

create_entrypoint:
```

lib/libutee/user_ta_entry.c
```diff
@@ -183,7 +191,16 @@ static TEE_Result init_instance(void)
...
+
+	res = TA_CreateEntryPoint();
+	if (res == TEE_SUCCESS)
```
jforissier:
```suggestion
	if (!res)
```

lib/libutee/user_ta_entry.c
```diff
@@ -167,6 +165,16 @@ static unsigned int get_memtag_implementation(void)
 
 static TEE_Result init_instance(void)
 {
+	static bool create_entrypoint_done;
+	static bool init_done;
+	TEE_Result res = TEE_SUCCESS;
+
+	if (init_done)
+		goto init_done;
```
jforissier:
```suggestion
		return TEE_SUCCESS;
```

lib/libutee/user_ta_entry.c
```diff
@@ -183,7 +191,16 @@ static TEE_Result init_instance(void)
...
-	return TA_CreateEntryPoint();
+
+create_entrypoint_done:
+	create_entrypoint_done = true;
+
+	res = TA_CreateEntryPoint();
+	if (res == TEE_SUCCESS)
+		init_done = true;
+
+init_done:
```
jforissier:
```suggestion
```

johankor:
@jforissier : Thanks! I've made the changes.

jforissier:
LGTM. Please squash the fixups, change the subject to "libutee: Fix handling of failed TA_CreateEntryPoint() in keep-alive TAs" and add a link to the issue where the problem and its solution was discussed more in depth:
```
Link: https://github.com/OP-TEE/optee_os/pull/7443
```
Then also add: `Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`

Also, do you think you could add a non-regression test for this fix in `optee_test`? That would be helpful.

johankor:
@jforissier : Thank you for the review! Sure, I'll look into a non-regression test.

cc: @jenswi-linaro , @etienne-lms

jforissier:
It looks like the recent [Buildroot update](https://github.com/OP-TEE/manifest/commit/aab39982ed1219b33ddae75c85953036de085eb2) has made us run out of disk space in the CI VMs for the Xen jobs :-/
I will investigate how to save space.

Edit: https://github.com/OP-TEE/build/pull/843

jenswikl:
Ready to merge?

jforissier:
> Ready to merge?

Yes!

PR #7501, opened 2025-08-11
core: imx: Fix TZDRAM_START address for imx93evk

fronius-eliott:
According to the imx-mkimage tool, the TZDRAM_START address should be set to 0x96000000 for the imx93evk
(see: https://github.com/nxp-imx/imx-mkimage/blob/lf-6.12.20_2.0.0/iMX93/soc.mak#L44C1-L44C28).

This is also true for the mainline mkimage U-Boot (see: https://github.com/u-boot/u-boot/blob/master/arch/arm/mach-imx/imx9/container.cfg#L16)

Currently, the computed TZDRAM address is wrong and is equal to 0xfe000000 (see:
https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-imx/conf.mk#L548)

### DIFF
--- a/core/arch/arm/plat-imx/conf.mk
+++ b/core/arch/arm/plat-imx/conf.mk
@@ -232,6 +232,7 @@ CFG_IMX_ELE ?= n
 else ifneq (,$(filter $(PLATFORM_FLAVOR),$(mx93-flavorlist)))
 $(call force,CFG_MX93,y)
 $(call force,CFG_ARM64_core,y)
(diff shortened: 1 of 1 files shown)


sahilnxp:
@fronius-eliott This address is according to the NXP i.MX BSP. We don't want to impose same thing on others.
They can have different address. If you are using with NXP i.MX BSP, you can use optee from there also.
or you can compile by passing `CFG_TZDRAM_START = 0x96000000` on command line during compilation.

PR #7502, opened 2025-08-12
Fix TPM log area mapping, zeroing, and address parsing for improved stability and consistency.

joseph-lo-nvtw:
This pull request introduces several improvements to the handling of the TPM event log within OP-TEE.

- __Device Tree Mapping__: We’ve simplified the device tree mapping process by replacing conditional compilation with a unified get_dt() function. This enhances code readability and maintainability while ensuring the correct device tree source is selected.
- __TPM Log Area Zeroing__: A fix has been implemented to prevent a write permission fault during TPM log area zeroing when using the SPMC manifest device tree. The zeroing operation is now conditional on the device tree source.
- __TPM Log Address Parsing__: We’ve addressed a potential issue with ARM32 systems by ensuring the full 64-bit TPM event log address is parsed consistently across all ARM architectures. This resolves potential address truncation problems.

These changes contribute to the stability, reliability, and consistency of the TPM log interaction within OP-TEE.

### DIFF
--- a/core/arch/arm/kernel/boot.c
+++ b/core/arch/arm/kernel/boot.c
@@ -2,6 +2,7 @@
 /*
  * Copyright (c) 2015-2023, Linaro Limited
  * Copyright (c) 2023, Arm Limited
+ * Copyright (c) 2025, NVIDIA Corporation & AFFILIATES.
  */
 
 #include <arm.h>
@@ -1014,7 +1015,7 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
 
 	init_external_dt(boot_arg_fdt, fdt_size);
 	reinit_manifest_dt();
-#ifdef CFG_CORE_SEL1_SPMC
+#ifdef CFG_CORE_FFA
 	tpm_map_log_area(get_manifest_dt());
 #else
 	tpm_map_log_area(get_external_dt());
--- a/core/kernel/tpm.c
+++ b/core/kernel/tpm.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 /*
  * Copyright (c) 2020-2023, ARM Limited. All rights reserved.
+ * Copyright (c) 2025, NVIDIA Corporation & AFFILIATES.
  */
 
 #include <compiler.h>
@@ -47,7 +48,7 @@ static int read_dt_tpm_log_info(void *fdt, int node, paddr_t *buf,
 
 	log_addr = fdt32_to_cpu(property[1]);
 
-	if (!IS_ENABLED(CFG_CORE_SEL1_SPMC)) {
+	if (!IS_ENABLED(CFG_CORE_FFA)) {
 		err = fdt_setprop(fdt, node, dt_tpm_event_log_addr, &zero_addr,
 				  sizeof(uint32_t) * 2);
 		if (err < 0) {
--- a/core/kernel/tpm.c
+++ b/core/kernel/tpm.c
@@ -46,7 +46,8 @@ static int read_dt_tpm_log_info(void *fdt, int node, paddr_t *buf,
 	if (!property  || len_prop != sizeof(uint32_t) * 2)
 		return -1;
 
-	log_addr = fdt32_to_cpu(property[1]);
+	log_addr = reg_pair_to_64(fdt32_to_cpu(property[0]),
+				  fdt32_to_cpu(property[1]));
 
 	if (!IS_ENABLED(CFG_CORE_FFA)) {
 		err = fdt_setprop(fdt, node, dt_tpm_event_log_addr, &zero_addr,


core/arch/arm/kernel/boot.c
```diff
@@ -1014,11 +1015,7 @@ void __weak boot_init_primary_late(unsigned long fdt __unused,
 
 	init_external_dt(boot_arg_fdt, fdt_size);
 	reinit_manifest_dt();
-#ifdef CFG_CORE_SEL1_SPMC
-	tpm_map_log_area(get_manifest_dt());
-#else
-	tpm_map_log_area(get_external_dt());
-#endif
+	tpm_map_log_area(get_dt());
```
jenswikl: Unfortunately, there isn't one common order of priority for DTs. The order depends on what we're looking for. `get_dt()` often gives the right DT, but not here.

core/kernel/tpm.c
```diff
@@ -46,7 +46,8 @@ static int read_dt_tpm_log_info(void *fdt, int node, paddr_t *buf,
 	if (!property  || len_prop != sizeof(uint32_t) * 2)
 		return -1;
 
-	log_addr = fdt32_to_cpu(property[1]);
+	log_addr = fdt32_to_cpu(property[0]);
+	log_addr = (log_addr << 32) | fdt32_to_cpu(property[1]);
```
jenswikl: Prefer
``` C
log_addr = reg_pair_to_64(fdt32_to_cpu(property[0]), fdt32_to_cpu(property[1]));
```

jenswikl:
> core: tpm: fix TPM log area zeroing condition to avoid manifest DT write fault
> Change the condition for zeroing TPM event log address in device tree
from CFG_CORE_SEL1_SPMC to CFG_CORE_FFA. When CFG_CORE_FFA is enabled,
the device tree comes from the SPMC manifest which is mapped as

It's only an SPMC manifest for CFG_CORE_SEL1_SPMC=y, for the other FF-A cases it's an SP manifest.

> read-only, not from the external device tree. Attempting to zero the
tpm_event_log_addr property in the manifest DT would cause a write
permission fault exception.

> The fix prevents this fault by ensuring the zeroing operation only occurs
when using the external device tree (CFG_CORE_FFA=n), where write access
is permitted.

joseph-lo-nvtw:
> > core: tpm: fix TPM log area zeroing condition to avoid manifest DT write fault
> > Change the condition for zeroing TPM event log address in device tree
> > from CFG_CORE_SEL1_SPMC to CFG_CORE_FFA. When CFG_CORE_FFA is enabled,
> > the device tree comes from the SPMC manifest which is mapped as
> 
> It's only an SPMC manifest for CFG_CORE_SEL1_SPMC=y, for the other FF-A cases it's an SP manifest.
> 
> > read-only, not from the external device tree. Attempting to zero the
> > tpm_event_log_addr property in the manifest DT would cause a write
> > permission fault exception.
> 
> > The fix prevents this fault by ensuring the zeroing operation only occurs
> > when using the external device tree (CFG_CORE_FFA=n), where write access
> > is permitted.
> > core: tpm: fix TPM log area zeroing condition to avoid manifest DT write fault
> > Change the condition for zeroing TPM event log address in device tree
> > from CFG_CORE_SEL1_SPMC to CFG_CORE_FFA. When CFG_CORE_FFA is enabled,
> > the device tree comes from the SPMC manifest which is mapped as
> 
> It's only an SPMC manifest for CFG_CORE_SEL1_SPMC=y, for the other FF-A cases it's an SP manifest.

Oops, it's typo. It should be SP manifest for a generic term.

> 
> > read-only, not from the external device tree. Attempting to zero the
> > tpm_event_log_addr property in the manifest DT would cause a write
> > permission fault exception.
> 
> > The fix prevents this fault by ensuring the zeroing operation only occurs
> > when using the external device tree (CFG_CORE_FFA=n), where write access
> > is permitted.

core/kernel/tpm.c
joseph-lo-nvtw: Okay, will fix.

jenswikl:
> Oops, it's typo. It should be SP manifest for a generic term.

SPMC and SP manifest are two different things. Let's call it a manifest.

core/arch/arm/kernel/boot.c
joseph-lo-nvtw: Hi Jens,

Thanks for your review.

The `get_dt()` internally invokes all the cases to query DT. As the OP-TEE (depends on different platforms or configurations) image supports one of them during boot time, it should be only one of the get dt APIs that works. May I know your concern as to why it isn't  working here?

Should we still use compiler flag here and change `CFG_CORE_SEL1_SPMC` to `CFG_CORE_FFA` to fix the issue?

jenswikl: On QEMU virt with `CFG_CORE_SEL1_SPMC=y`, we have both an external DT and an SPMC manifest. It's messy, but I don't know what would be the best solution.

joseph-lo-nvtw: Okay, thanks for sharing the test case.

jenswikl:

joseph-lo-nvtw:
Update the patch set with reviewed tag and thank you for your review.

PR #7503, opened 2025-08-15
Add PLATFORM=vexpress-qemu_sbsa

jenswi-linaro:


jforissier:
- For "core: ffa: discover non-secure memory from manifest":

core/kernel/boot.c
```diff
@@ -99,20 +101,21 @@ static int __maybe_unused get_nsec_memory_helper(void *fdt, struct core_mmu_phys
 }
 
 #ifdef CFG_DT
-static struct core_mmu_phys_mem *get_nsec_memory(void *fdt, size_t *nelems)
+static struct core_mmu_phys_mem *get_nsec_memory(void *fdt, size_t *nelems,
+						 const char *dev_type)
```
jforissier, line 105: The prototype in the `#else` branch needs updating too.

jenswikl: Thanks, I'll update.

jenswikl:
Addressing the comment and adding the tag.

core/kernel/boot.c
```diff
@@ -132,12 +136,26 @@ void discover_nsec_memory(void)
 	struct core_mmu_phys_mem *mem;
```
etienne-lms: By the way, could you also update this? `= NULL`

core/arch/arm/plat-vexpress/conf.mk
```diff
@@ -170,6 +174,21 @@ CFG_CORE_ASYNC_NOTIF_GIC_INTID ?= 219
...
+CFG_AUTO_MAX_PA_BITS ?= y
+CFG_CORE_RESERVED_SHM ?= n
+ifeq ($(CFG_CORE_SEL2_SPMC),y)
+else
+$(call force,CFG_ARM_GICV3,y)
+$(call force,CFG_DT_ADDR,0)
+$(call force,CFG_DT,y)
+CFG_DTB_MAX_SIZE ?= 0x100000
+endif
+endif
```
etienne-lms: `endif #PLATFORM==qemu_sbsa`?

jenswikl: I'll update

core/kernel/boot.c
jenswikl: OK

jenswikl:

PR #7507, opened 2025-08-18
plat-rdaspen: introduce RD-Aspen platform support

ahmedtiba:
Add initial support for the RD-Aspen platform. This includes:

1. GIC and console initialization functions.
2. Platform-specific memory layout definitions.
3. Makefile support for platform integration.
4. Assembly function `get_core_pos_mpidr`, used to compute the linear core position based on MPIDR_EL1.


core/arch/arm/plat-rdaspen/rdaspen_core_pos.S
```diff
@@ -0,0 +1,27 @@
...
+/*
+ * Copyright (c) 2025, Arm Limited
+ */
+
+#include <asm.S>
+#include <arm.h>
+#include "platform_config.h"
+
+FUNC get_core_pos_mpidr , :
+	mrs	x0, mpidr_el1
```
jenswikl: This function is supposed to take `mpidr` as an argument in `x0`.
By the way, why isn't the default `get_core_pos_mpidr()` used?

jenswikl:
How is this platform different from core/arch/arm/plat-rd1ae?

ahmedtiba:
> How is this platform different from core/arch/arm/plat-rd1ae?

This platform differs from plat-rd1ae mainly in system topology and peripheral base addresses:

Topology:

	- RD-1AE supports up to 16 clusters × 1 CPU per cluster.
	- RD-Aspen has a single cluster with 4 CPUs per cluster.

GIC:

	- RD-1AE uses GICv2.
	- RD-Aspen uses GICv3.

Memory map:

	- RD-1AE defines two DRAM regions.
	- RD-Aspen defines a single DRAM region.
	
UART:

	- RD-1AE console at UART1_BASE = 0x2A41_0000.
	- RD-Aspen console at UART1_BASE = 0x1A41_0000.

core/arch/arm/plat-rdaspen/rdaspen_core_pos.S
ahmedtiba: Done.
The implementation now uses the MPIDR passed in x0.
A custom implementation is needed to match RD-Aspen’s topology.

core/arch/arm/plat-rdaspen/rdaspen_core_pos.S
```diff
@@ -0,0 +1,26 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2025, Arm Limited
+ */
+
+#include <asm.S>
+#include <arm.h>
+#include "platform_config.h"
+
+FUNC get_core_pos_mpidr , :
```
jenswikl: Is this function different from `get_core_pos_mpidr()` with `CFG_CORE_CLUSTER_SHIFT=RDASPEN_MAX_CPUS_PER_CLUSTER` and `CFG_CORE_THREAD_SHIFT=RDASPEN_MAX_PE_PER_CPU`?

jenswikl:
> This platform differs from plat-rd1ae mainly in system topology and peripheral base addresses

What if we use the same platform and different variants to share common code? The approach of one variant per platform doesn't scale well in terms of maintenance.

ahmedtiba:
> > This platform differs from plat-rd1ae mainly in system topology and peripheral base addresses
> 
> What if we use the same platform and different variants to share common code? The approach of one variant per platform doesn't scale well in terms of maintenance.

Great point, I agree one-variant-per-platform doesn’t scale.
 Here’s what I propose :
 Create a shared RD "family" layer (e.g. core/arch/arm/plat-automotive-rd/common/) that holds:
	•	Common console (PL011) setup, generic DDR registration helpers.
	•	GIC init that switches on CFG_GICV3 (gic_init_v3 vs gic_init/gic_init_per_cpu).
	•	No platform override of get_core_pos_mpidr(), rely on the default by setting per-variant shifts.
	•	platform_config.h with only address map (UART/GIC/DRAM).
	•	Makefile flags set CFG_CORE_CLUSTER_SHIFT and CFG_CORE_THREAD_SHIFT (log2 of counts). With that, variants don't need their own get_core_pos_mpidr().

Structure:
	•	plat-automotive-rd/common/rd_common.{c,S}, rd_common.h
	•	plat-automotive-rd/<variant>/platform_config.h (+ minimal main.c if needed)
	•	core/arch/arm/plat-automotive-rd/sub.mk.

Outcome: shared code lives once; variants carry only address maps + config. For RD-Aspen specifically, I can drop the custom get_core_pos_mpidr() by setting:
	•	CFG_CORE_CLUSTER_SHIFT = 2 (4 CPUs/cluster)
	•	CFG_CORE_THREAD_SHIFT = 0 (1 PE/CPU)

If that direction sounds good, I’ll rework this patch to: (1) move the common bits under plat-automotive-rd/common, (2) keep RD-Aspen as a thin variant with just its map and Makefile, and (3) delete the local get_core_pos_mpidr() override.

jenswikl:
Which platforms would sort under plat-automotive-rd? Is RD-1 AE included?

ahmedtiba:
> Which platforms would sort under plat-automotive-rd? Is RD-1 AE included?

RD-Aspen and RD-1 AE. Yes RD-1 AE is included.

jenswikl:
Good. Please remove `common` from `core/arch/arm/plat-automotive-rd/common/`. A flat structure should be enough for now.

ahmedtiba:
> Good. Please remove `common` from `core/arch/arm/plat-automotive-rd/common/`. A flat structure should be enough for now.

Thanks for the guidance.

ahmedtiba:
plat-automotive-rd gets rewritten internally to: 

PLATFORM = automotive
PLATFORM_FLAVOR = rd

Is there a supported way to bypass or disable this automatic hyphen-splitting, so the build system will use PLATFORM=automotive-rd literally and look for core/arch/arm/plat-automotive-rd/conf.mk?

jforissier:
@ahmedtiba no, I suggest using a underscore instead: `PLATFORM=automotive_rd`.

ahmedtiba:
Would it be okay to use a separate main.c file for each variant, or would it be better to maintain a single shared main.c across all variants?

jenswikl:
Let's try a single shared main.c first.

davidhuziji:
Hi @jenswi-linaro , thanks for the suggestion to maintain both platforms under `plat-automotive-rd` folder. It makes sense to me as both platforms are Arm automotive solutions.

Meanwhile, these 2 platforms are very independent platforms. They are built based on different CPU series, with different topology, different GICs and other different features. And RD-Aspen may bring in further changes, more different from the RD1AE platform.

I wondering if it makes sense to you to keep RD1AE and RD-Aspen under `plat-automotive-rd` folder, but maintain their own main.c, Makefile and config files. Otherwise, it will become a maintenance overhead for us to update RD-Aspen, while keeping RD1AE implementations in the same file.

jenswikl:
I think we should start with everything shared. That's how it's done in the other platforms. It can hardly get worse than plat-vexpress, which is still manageable.

core/arch/arm/plat-automotive_rd/conf.mk
```diff
@@ -0,0 +1,15 @@
+ifeq ($(CFG_RD1AE)$(CFG_RDASPEN),yy)
```
jenswikl: Please use `PLATFORM_FLAVOR`.

core/arch/arm/plat-automotive_rd/conf.mk
```diff
@@ -0,0 +1,15 @@
...
+$(error Both CFG_RD1AE and CFG_RDASPEN are set. Select only one.)
+endif
+
+ifeq ($(CFG_RD1AE)$(CFG_RDASPEN),)
+$(error No variant selected for PLATFORM=automotive_rd. Set exactly one of: \
+CFG_RD1AE=y or CFG_RDASPEN=y)
+endif
+
+ifeq ($(CFG_RD1AE),y)
+include core/arch/arm/plat-automotive_rd/rd1ae/conf.mk
```
jenswikl: I'd rather see the content of each of those mk-files in this mk-file.

core/arch/arm/plat-automotive_rd/rd1ae/variant_config.h
jenswikl, line 1: Please use a single shared `platform_config.h` file for this platform. There's no need for a per-variant `variant_config.h`.

core/arch/arm/plat-automotive_rd/sub.mk
```diff
@@ -0,0 +1,6 @@
+global-incdirs-y += .
+srcs-y += main.c
+
+# Variant selection
+subdirs-$(CFG_RDASPEN) += rdaspen
```
jenswikl: We don't need subdirectories for variants.

ahmedtiba: Removed subdirectories for variants.

core/arch/arm/plat-automotive_rd/rd1ae/variant_config.h
ahmedtiba: Done

core/arch/arm/plat-automotive_rd/conf.mk
ahmedtiba: Done

core/arch/arm/plat-automotive_rd/conf.mk
ahmedtiba: Done

core/arch/arm/plat-automotive_rd/platform_config.h
```diff
@@ -0,0 +1,71 @@
...
+#ifndef PLATFORM_CONFIG_H
+#define PLATFORM_CONFIG_H
+
+#include <mm/generic_ram_layout.h>
+
+/* Make stacks aligned to data cache line length */
+#define STACK_ALIGNMENT			64
+
+/* Common size helper macros */
+#define SZ_2G				UL(0x80000000)
```
jenswikl: Please use `SIZE_2G` from `<util.h>` instead.

ahmedtiba: Done

jenswikl:

ahmedtiba:

Thanks for the Acked-by, Jens.

Just to confirm, is the next step to wait for merge, or should I take any further action ?
Let me know if there’s anything needed from my side to move this forward.

jforissier:
@ahmedtiba please add the Acked-by: to the commit description and force push. Then I will merge.

ahmedtiba:
> @ahmedtiba please add the Acked-by: to the commit description and force push. Then I will merge.

Done, Thanks for clarification.

PR #7509, opened 2025-08-19
core: console: Reduce logging verbosity during runtime

glneo:
The default console can be disabled at runtime for power management. Any prints after boot from OP-TEE could cause a bus data abort if the UART has been disabled. Add an option to lower the log level after boot has completed to only print error level messages. This matches what is also done in TF-A.

Note: This is a continuation of https://github.com/OP-TEE/optee_os/pull/6136


core/kernel/console.c
```diff
@@ -57,6 +58,17 @@ void register_serial_console(struct serial_chip *chip)
 	serial_console = chip;
 }
 
+#ifdef CFG_CONSOLE_RUNTIME
+static TEE_Result console_runtime(void)
+{
+	/* Console is only used for errors after this point */
+	trace_set_level(TRACE_ERROR);
```
jforissier: How is this supposed to work? Say console has been disabled, `console_runtime()` has been called. Then an error occurs. Crash?

glneo: Right, it might crash here. I could simply disable all output after boot, even error output, but given an error messages should be fatal anyway I figured it was worth the risk. To be fully correct I'll disable those also.

The main thing I'm trying to avoid here is all the silly info level messages the get printed at runtime that crash our systems. Error messages are rare enough this hasn't been a problem, but I can turn those off at runtime also.

jforissier: Much better would be to fix the cause of the data abort. Does OP-TEE have a way to detect the console is disabled? Could the firmware that disables the UART also notify OP-TEE? That would be proper runtime handling of the situation.

glneo: Yes OP-TEE could ask the system control firmware if the IP is enabled, but we can't know when Linux will decide to do runtime power management on that IP so we would race to print out our string before it was disabled.

The slightly more correct option would then be to tell the system control firmware that we will be using it so it shouldn't be powered down. The issue here is since we don't know when we will print, OP-TEE would either have to request the IP at startup and keep it for the whole runtime (which would be very power inefficient), or we request it before each time we are going to print and release it after.

For that second option, there doesn't seem to be a unified API we could tap into to do the enabling outside of the `console_putc` and `console_getchar`, which would mean doing this for every character printed..

core/kernel/console.c
```diff
@@ -57,6 +58,17 @@ void register_serial_console(struct serial_chip *chip)
 	serial_console = chip;
 }
 
+#ifdef CFG_CONSOLE_RUNTIME
```
jenswikl: The name should reflect that the console is disabled at some point.

core/arch/arm/plat-k3/conf.mk
```diff
@@ -1,6 +1,7 @@
 CFG_WITH_STATS ?= y
 CFG_CRYPTO_WITH_CE ?= y
 CFG_CONSOLE_UART ?= 0
+CFG_CONSOLE_RUNTIME ?= y
```
jenswikl: Please document this in `mk/config`, take `CFG_FFA_CONSOLE` and an example.

jenswikl:

etienne-lms (commented):

jforissier:
@glneo would you please add the R-b/A-b tags? Thanks.

core/kernel/console.c
```diff
@@ -57,6 +58,17 @@ void register_serial_console(struct serial_chip *chip)
...
 
+#ifdef CFG_CONSOLE_RUNTIME_DISABLE
+static TEE_Result console_runtime_disable(void)
+{
+	/* Do not print trace messages after this point */
+	trace_set_level(TRACE_MIN);
+
+	return TEE_SUCCESS;
+}
+boot_final(console_runtime_disable);
```
etienne-lms: CI checkpatch complains on a missing empty line above. Could you add it?
CI failure on QEMUv8/Hafnium seems unrelated (network issues?)

jforissier: > CI failure on QEMUv8/Hafnium seems unrelated (network issues?)

Yes, it looks like the CI host often has problems cloning from trustedfirmware.org. I could not reproduce the issue on my PC. It is unfortunate that there is no other mirror (that I know of) for the project and its submodules. GitHub has a copy but the submodules are still on trustedfirmware.org so that won't help.

core/kernel/console.c
```diff
@@ -57,6 +58,17 @@ void register_serial_console(struct serial_chip *chip)
 	serial_console = chip;
 }
 
+#ifdef CFG_CONSOLE_RUNTIME_DISABLE
+static TEE_Result console_runtime_disable(void)
+{
+	/* Do not print trace messages after this point */
```
etienne-lms: Maybe a last info trace message could be nice here, e.g.:
```c
	IMSG("Disabling output console");
```

core/kernel/console.c
```diff
@@ -57,6 +58,17 @@ void register_serial_console(struct serial_chip *chip)
...
+	trace_set_level(TRACE_MIN);
```
etienne-lms: By the way, that makes me think we could add a nice command to `xtest --stats` to change the core log level at runtime. Would it be worth it? I think so, allow to build with debug log level but reducing to error or info level when 

Also I wonder it would be worth to add a generic `CFG_TEE_CORE_RUNTIME_LOG_LEVEL=0/1/2/3/4` to allow selecting a lower log level straight after OP-TEE core inits. (Meaning `CFG_TEE_CORE_RUNTIME_LOG_LEVEL=0` is equivalent to `CFG_CONSOLE_RUNTIME_DISABLE=y`).

etienne-lms (commented):
A few comments, but my A-b tag still applies.

core/kernel/console.c
glneo: Changing the log level at runtime might be useful, especially when running something like `xtest` when we might want more verbose output. But would probably be out of scope for this PR.

For setting the level to some value, I'd need to keep a different define to hold if the level should be changed at all, otherwise we would always switch to some default value for the RUNTIME_LOG_LEVEL define. If that is fine I've prototypes the suggestion and updated this PR with that.

PR #7515, opened 2025-08-27
plat-rockchip: rk3588: cache HUK in memory

tretter:
I observed timeout errors when OP-TEE reads the HUK from the OTP area
while running the optee-xtests (tests 1006 and 4013) or using the
pkcs#11 TA.

This issue is circumvented by reading the HUK once and caching it in
memory for later use. As a side effect, this reduces the accesses/reads
from the OTP area.

Unfortunately, I don't know the root cause for the timeout while reading
the fuses. I guess that there is a disabled clock which prevents the
read, but I didn't look further, since caching works fine.

While the documentation recommends to never process the HUK in software,
it is read and processed anyway if it can be read from the fuses. Thus,
I don't think that caching has an effect on the security of the HUK.
The caching is inspired by the HUK handling implemented in the nvmem
driver.

### DIFF
--- a/core/arch/arm/plat-rockchip/platform_rk3588.c
+++ b/core/arch/arm/plat-rockchip/platform_rk3588.c
@@ -174,11 +174,21 @@ static TEE_Result generate_huk(struct tee_hw_unique_key *hwkey)
 {
 	TEE_Result res = TEE_SUCCESS;
 	uint8_t buffer[HW_UNIQUE_KEY_LENGTH] = { };
+	size_t i = 0;
+	bool key_is_zero = true;
 
 	/* Generate random 128-bit key from TRNG */
 	res = hw_get_random_bytes(buffer, sizeof(buffer));
 	if (res)
 		return res;
+
+	/* All zero HUK cannot be written to OTP and indicates TRNG failure */
+	for (i = 0; i < ARRAY_SIZE(buffer); i++) {
+		if (buffer[i] != 0)
+			key_is_zero = false;
+	}
+	if (key_is_zero)
+		return TEE_ERROR_NO_DATA;
 
 	memcpy(hwkey->data, buffer, HW_UNIQUE_KEY_LENGTH);
 
--- a/core/arch/arm/plat-rockchip/platform_rk3588.c
+++ b/core/arch/arm/plat-rockchip/platform_rk3588.c
@@ -14,6 +14,7 @@
 #include <platform.h>
 #include <platform_config.h>
 #include <rng_support.h>
+#include <stdlib_ext.h>
 #include <string.h>
 #include <string_ext.h>
 #include <utee_defines.h>
@@ -56,6 +57,9 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, TRNG_S_BASE, TRNG_S_SIZE);
 
 static struct mutex trng_mutex = MUTEX_INITIALIZER;
 static struct mutex huk_mutex = MUTEX_INITIALIZER;
+
+/* Cache the HUK in memory */
+static struct tee_hw_unique_key *huk;
 
 int platform_secure_ddr_region(int rgn, paddr_t st, size_t sz)
 {
@@ -239,20 +243,34 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 
 	mutex_lock(&huk_mutex);
 
+	/* Try to use the cached HUK from memory */
+	if (huk)
+		goto out;
+
+	huk = malloc(sizeof(*huk));
+	if (!huk) {
+		res = TEE_ERROR_OUT_OF_MEMORY;
+		goto out;
+	}
+
 	/* Try to read and cache the HUK persisted in the OTP */
-	res = read_huk(hwkey);
+	res = read_huk(huk);
 	if (res != TEE_ERROR_NO_DATA)
 		goto out;
 
 	/* Try to generate and use a new HUK and persist it in the OTP */
-	res = generate_huk(hwkey);
+	res = generate_huk(huk);
 	if (res != TEE_SUCCESS)
 		goto out;
-	res = persist_huk(hwkey);
+	res = persist_huk(huk);
 
 out:
-	if (res != TEE_SUCCESS)
-		memzero_explicit(hwkey->data, HW_UNIQUE_KEY_LENGTH);
+	if (res == TEE_SUCCESS) {
+		memcpy(hwkey->data, huk->data, HW_UNIQUE_KEY_LENGTH);
+	} else if (huk) {
+		free_wipe(huk);
+		huk = NULL;
+	}
 
 	mutex_unlock(&huk_mutex);
 
--- a/core/arch/arm/plat-rockchip/platform_rk3588.c
+++ b/core/arch/arm/plat-rockchip/platform_rk3588.c
@@ -166,17 +166,44 @@ TEE_Result hw_get_random_bytes(void *buf, size_t blen)
 	return TEE_SUCCESS;
 }
 
-TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+static TEE_Result generate_huk(struct tee_hw_unique_key *hwkey)
+{
+	TEE_Result res = TEE_SUCCESS;
+	uint8_t buffer[HW_UNIQUE_KEY_LENGTH] = { };
+
+	/* Generate random 128-bit key from TRNG */
+	res = hw_get_random_bytes(buffer, sizeof(buffer));
+	if (res)
+		return res;
+
+	memcpy(hwkey->data, buffer, HW_UNIQUE_KEY_LENGTH);
+
+	return res;
+}
+
+static TEE_Result persist_huk(struct tee_hw_unique_key *hwkey)
+{
+	TEE_Result res = TEE_SUCCESS;
+	uint32_t buffer[ROCKCHIP_OTP_HUK_SIZE] = { };
+
+	memcpy(buffer, hwkey->data, HW_UNIQUE_KEY_LENGTH);
+
+	/* Write the new HUK into OTP at ROCKCHIP_OTP_HUK_INDEX */
+	res = rockchip_otp_write_secure(buffer, ROCKCHIP_OTP_HUK_INDEX,
+					ROCKCHIP_OTP_HUK_SIZE);
+
+	/* Clear buffer memory */
+	memzero_explicit(buffer, sizeof(buffer));
+
+	return res;
+}
+
+static TEE_Result read_huk(struct tee_hw_unique_key *hwkey)
 {
 	TEE_Result res = TEE_SUCCESS;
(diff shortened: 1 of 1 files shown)


core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -342,17 +342,44 @@ static TEE_Result tee_otp_write_secure(const uint32_t *value, uint32_t index,
 	return TEE_SUCCESS;
 }
 
-TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
+static TEE_Result tee_rockchip_generate_huk(struct tee_hw_unique_key *hwkey)
```
jenswikl: This is a local function, so there's no need to add a prefix. I prefer `generate_huk()`. Same for the other static functions with a `tee_rockchip_` prefix.

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -368,18 +395,8 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
-			goto out;
-
-		/* Write the new HUK into OTP at HW_UNIQUE_KEY_INDEX */
-		res = tee_otp_write_secure(buffer,
-					   HW_UNIQUE_KEY_INDEX,
-					   HW_UNIQUE_KEY_LENGTH /
-					   sizeof(uint32_t));
-		if (res)
-			goto out;
+		res = TEE_ERROR_NO_DATA;
```
jenswikl: `buffer` is already zeroed, so we can return `TEE_ERROR_NO_DATA` directly.

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -389,6 +406,30 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	mutex_lock(&huk_mutex);
+
+	res = tee_rockchip_read_huk(hwkey);
+	if (res == TEE_ERROR_NO_DATA) {
+		res = tee_rockchip_generate_huk(hwkey);
+		if (res != TEE_SUCCESS)
+			goto out;
+		res = tee_rockchip_persist_huk(hwkey);
```
etienne-lms:
```suggestion
		if (res == TEE_SUCCESS)
			res = tee_rockchip_persist_huk(hwkey);
```
And remove the `out` branch label. Would look simpler;

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -368,18 +395,8 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
-
-		/* Write the new HUK into OTP at HW_UNIQUE_KEY_INDEX */
-		res = tee_otp_write_secure(buffer,
-					   HW_UNIQUE_KEY_INDEX,
-					   HW_UNIQUE_KEY_LENGTH /
-					   sizeof(uint32_t));
-		if (res)
-			goto out;
+		res = TEE_ERROR_NO_DATA;
+		goto out;
```
etienne-lms: Nitpicking: `return TEE_ERROR_NOT_DATA` would be enough.

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -342,17 +342,44 @@ static TEE_Result tee_otp_write_secure(const uint32_t *value, uint32_t index,
...
 {
 	TEE_Result res = TEE_SUCCESS;
 	uint32_t buffer[HW_UNIQUE_KEY_LENGTH / sizeof(uint32_t)] = { };
-	bool key_is_empty = true;
-	size_t i = 0;
 
-	if (!hwkey)
-		return TEE_ERROR_BAD_PARAMETERS;
+	/* Generate random 128-bit key from TRNG */
+	res = hw_get_random_bytes(buffer, sizeof(buffer));
```
etienne-lms, line 181: What if `hw_get_random_bytes()` generates 128 bits of 0s? Unlikely or impossible?
`tee_otp_get_hw_unique_key()` would return `TEE_SUCCESS` with an all-0 key (and try to burn it in the OTPs) but a next call to `tee_otp_get_hw_unique_key()` would try to generate a new key.

Actually, I see the issue already exists in the original code, so this may be a bit out of the scope of this P-R. But worth to address here, no?

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -426,6 +432,16 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
 
+	huk = malloc(sizeof(*huk));
+	if (!huk) {
+		res = TEE_ERROR_OUT_OF_MEMORY;
+		goto out;
+	}
+	memcpy(huk->data, hwkey->data, HW_UNIQUE_KEY_LENGTH);
+
+cached:
+	/* Copy HUK into hwkey->data */
```
etienne-lms: Nitpicking: Is the inline comment really needed (instruction below seems quite explicit)?

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -418,6 +421,9 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
 
 	mutex_lock(&huk_mutex);
 
+	if (huk)
+		goto cached;
+
 	res = tee_rockchip_read_huk(hwkey);
```
etienne-lms: It would look simpler to allocate `huk` here, then generate the HUK in `huk` and copy it to `hwkey`  at the end, upon success.

tretter: If the `huk` is allocated first, it's more difficult to handle errors when reading or generating the HUK fails. A second call to `tee_otp_get_hw_unique_key` would see that `huk` is set and use the cached, but invalid value. Thus, the error handling would have to ensure that `huk` is explicitly freed and set to `NULL` in case of an error. On the other hand, explicit error handling would make this more clear. I'll give it a try.

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: Yes, but this would rely on the check for zero done before to know that we can skip the cleanup. Using the same cleanup path for errors seems more robust to me.

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: I'll drop the comment. There are several other comments in the code that just describe, what the next line is doing, without adding further information. Should I remove such comments?

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: Are there some means to check the randomness of a generated key (or random data) in OP-TEE?

While 128 bits of 0s are bad, because the HUK would change on a second call of `tee_otp_get_hw_unique_key` (or on a second boot in case of the cached `huk`), maybe we could just run a randomness test on the generated HUK to prevent the generation of a HUK with a faulty PRNG before writing it to the OTP?

tretter:
> Unfortunately, I don't know the root cause for the timeout while reading
> the fuses. I guess that there is a disabled clock which prevents the
> read, but I didn't look further, since caching works fine.

I debugged this further. Disabling the `clk_otpc_arb` via the CRU from the normal world causes read timeouts for OTP in the secure world, which includes the HUK in the OTP. Unfortunately, I cannot find much documentation of how the OTP work or how these clocks are driven.

tretter:

When squashing the commits, 2198aebd3faadeb09c07586096027052994df036 needs a bit of special care, because it would be useful to squash some changes into the refactoring commit and some into the commit that adds the caching. Splitting the change into separate fixups on top doesn't improve the conflict situation, and adding a fixup between the commits conflicts with the contributing guidelines.

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: I reworked the control flow in this function and added comments to explain the different methods that are tried to get the HUK. I hope this is actually simpler and easier to follow, now.

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: I added a test, if the generated key consists of zeros. `generate_huk` now returns an error in this case, the HUK will not be persisted and `tee_otp_get_hw_unique_key` will return an error, too. It's now the caller's responsibility to handle this error.

jenswi-linaro (commented):

core/arch/arm/plat-rockchip/platform_rk3588.c
```diff
@@ -389,6 +416,49 @@ TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
...
+		goto out;
+	res = persist_huk(huk);
+
+out:
+	if (res == TEE_SUCCESS) {
+		memcpy(hwkey->data, huk->data, HW_UNIQUE_KEY_LENGTH);
+	} else if (huk) {
+		/* Get rid of cached HUK in case of an error */
+		memzero_explicit(huk, sizeof(*huk));
+		free(huk);
```
etienne-lms: Suggestion (fell free to address or not):
```suggestion
	} else {
		free_wipe(huk);

File: core/arch/arm/plat-rockchip/platform_rk3588.c [hunk 9]
```diff
@@ -84,6 +84,9 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, OTP_S_BASE, OTP_S_SIZE);
 static struct mutex trng_mutex = MUTEX_INITIALIZER;
 static struct mutex huk_mutex = MUTEX_INITIALIZER;
 
+/* Cache the HUK in memory */
+struct tee_hw_unique_key *huk;
```
etienne-lms: `static`

etienne-lms (commented):
Fine with me with the `static` keyword added to `huk`.
Could you squash the fixup commits?

tretter:
Fixed up commits with tag and ready to be merged

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: Fixed.

core/arch/arm/plat-rockchip/platform_rk3588.c
tretter: Nice, I was looking for something like this, but failed to find it myself. Thanks!

etienne-lms:

tretter:
Solved merge conflicts with master, fixed up commits with tag and ready to be merged.

PR #7517, opened 2025-09-03
AMD ASU Support

Akshay-Belsare:
Add support for the AMD Application Security Unit (ASU), the on-chip
    Hardware Security Module (HSM) for Versal Gen 2.
    The ASU manages all device-level security services for user
    applications, extending beyond accelerator-centric tasks.
    Its firmware also exposes several software-based cryptographic
    primitives, including:
    - Key transfer
    - RSA authentication (multiple padding schemes)
    - HMAC
    - Key Derivation Function (KDF)
    - Key wrap / unwrap

### DIFF
--- a/core/drivers/amd/asu/asu_doorbell.h
+++ b/core/drivers/amd/asu/asu_doorbell.h
@@ -1,0 +1,66 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024-2026, Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ */
+
+#ifndef __ASU_DOORBELL_H__
+#define __ASU_DOORBELL_H__
+
+#include <stdint.h>
+
+#define PAR_IPIPSU_NUM_INSTANCES	1U
+
+/* Parameter definitions for peripheral ASU IPI */
+#define PAR_IPIPSU_0_DEVICE_ID		0U
+#define PAR_IPIPSU_0_BASEADDR		0xEB330000U
+#define PAR_IPIPSU_0_BIT_MASK		0x00000001U
+#define PAR_IPIPSU_0_INT_ID		0x59U
+#define PAR_IPIPSU_NUM_TARGETS		16U
+#define PAR_PSV_IPI_PMC_BIT_MASK	0x2U
+#define PAR_PSV_IPI_0_BIT_MASK		0x4U
+#define PAR_PSV_IPI_1_BIT_MASK		0x8U
+#define PAR_PSV_IPI_2_BIT_MASK		0x10U
+#define PAR_PSV_IPI_3_BIT_MASK		0x20U
+#define PAR_PSV_IPI_4_BIT_MASK		0x40U
+#define PAR_PSV_IPI_5_BIT_MASK		0x80U
+#define PAR_PSV_IPI_PMC_NOBUF_BIT_MASK	0x100U
+#define PAR_PSV_IPI_6_BIT_MASK		0x200U
+#define IPIPSU_ALL_MASK			0xFFFFU
+
+#define IPIPSU_TRIG_OFFSET	0x00U
+#define IPIPSU_OBS_OFFSET	0x04U
+#define IPIPSU_ISR_OFFSET	0x10U
+#define IPIPSU_IMR_OFFSET	0x14U
+#define IPIPSU_IER_OFFSET	0x18U
+
+struct asu_doorbell_config {
+	uint32_t deviceid;
+	uintptr_t baseaddr;
+	uint32_t bitmask;
+	uint32_t intrid;
+	uintptr_t intrparent;
+	uint32_t target_count;
+	uint32_t target_mask[PAR_IPIPSU_NUM_TARGETS];
+};
+
+struct asu_doorbell_config asu_configtable = {
+	.deviceid = PAR_IPIPSU_0_DEVICE_ID,
+	.baseaddr = PAR_IPIPSU_0_BASEADDR,
+	.bitmask = PAR_IPIPSU_0_BIT_MASK,
+	.intrid = PAR_IPIPSU_0_INT_ID,
+	.target_count = PAR_IPIPSU_NUM_TARGETS,
+	.target_mask = {
+		PAR_PSV_IPI_PMC_BIT_MASK,
+		PAR_PSV_IPI_0_BIT_MASK,
+		PAR_PSV_IPI_1_BIT_MASK,
+		PAR_PSV_IPI_2_BIT_MASK,
+		PAR_PSV_IPI_3_BIT_MASK,
+		PAR_PSV_IPI_4_BIT_MASK,
+		PAR_PSV_IPI_5_BIT_MASK,
+		PAR_PSV_IPI_PMC_NOBUF_BIT_MASK,
+		PAR_PSV_IPI_6_BIT_MASK,
+	}
+};
+
+#endif /* __ASU_DOORBELL_H__ */
--- a/core/drivers/amd/asu/asu_main.c
+++ b/core/drivers/amd/asu/asu_main.c
@@ -1,0 +1,491 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2025-2026, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#include <assert.h>
+#include <drivers/amd/asu_client.h>
+#include <initcall.h>
+#include <io.h>
+#include <kernel/delay.h>
+#include <kernel/dt.h>
+#include <kernel/interrupt.h>
+#include <kernel/panic.h>
+#include <kernel/spinlock.h>
+#include <libfdt.h>
+#include <mm/core_mmu.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+#include <trace.h>
+#include <util.h>
+
+#include "asu_doorbell.h"
+
+#define ASU_QUEUE_BUFFER_FULL		0xFFU
+#define ASU_CLIENT_READY		0xFFFFFFFFU
+#define ASU_TARGET_IPI_INT_MASK		1U
+
+#define ASU_BASEADDR			0xEBF80000U
+#define ASU_GLOBAL_CNTRL		(ASU_BASEADDR + 0x00000000U)
+
+#define ASU_BASEADDR_SIZE		0x10000U
+#define ASU_GLOBAL_ADDR_LIMIT		0x1000U
+
+#define ASU_GLOBAL_CNTRL_FW_IS_PRESENT_MASK	0x10U
+#define ASU_ASUFW_BIT_CHECK_TIMEOUT_VALUE	0xFFFFFU
+
+#define ASU_CHNL_IPI_BITMASK		GENMASK_32(31, 16)
+
+struct asu_client {
+	struct asu_channel_memory *chnl_memptr;
+	uint32_t is_ready;
+	unsigned int slock;          /* chnl_memptr spin lock */
+	void *global_ctrl;
+	void *doorbell;
+};
+
+struct asu_ids {
+	uint8_t ids[ASU_UNIQUE_ID_MAX];
+	unsigned int slock;          /* id array spin lock */
+};
+
+static struct asu_ids asuid;
+static struct asu_client *asu;
+
+/*
+ * asu_fwcheck() - Check if ASU firmware is present and ready
+ * Polls the ASU global control register to verify if the HSM firmware
+ * is present and ready for interaction. Uses a timeout of approximately
+ * 1 second for the check.
+ *
+ * Return: TEE_SUCCESS if firmware is ready, TEE_ERROR_BAD_STATE if not present
+ */
+static TEE_Result asu_fwcheck(void)
+{
+	uint64_t timeout = 0;
+
+	/*
+	 * Timeout is set to ~1sec.
+	 * This is the worst case time within which ASUFW ready for interaction
+	 * with components requests.
+	 */
+	timeout = timeout_init_us(ASU_ASUFW_BIT_CHECK_TIMEOUT_VALUE);
+
+	do {
+		if (io_read32((vaddr_t)asu->global_ctrl) &
+		    ASU_GLOBAL_CNTRL_FW_IS_PRESENT_MASK) {
+			DMSG("ASU FW is ready!");
+			return TEE_SUCCESS;
+		}
+	} while (!timeout_elapsed(timeout));
+
+	EMSG("ASU FW is not present!");
+
+	return TEE_ERROR_BAD_STATE;
+}
+
+/*
+ * asu_get_channelID() - Determine the ASU channel ID for APU communication
+ *
+ * Maps the Runtime Configuration Area (RTCA) to find which ASU channel
+ * should be used by the APU for communication. Searches through available
+ * channels to find one matching the APU's local IPI ID.
+ *
+ * Return: Channel ID (0 to ASU_MAX_IPI_CHANNELS-1) or ASU_MAX_IPI_CHANNELS on
+ *	   failure
+ */
+static uint32_t asu_get_channelID(void)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	void *comm_chnl_info = NULL;
+	uint32_t channel_id = ASU_MAX_IPI_CHANNELS;
+	vaddr_t membase = 0;
+	uint32_t id = 0;
+
+	/*
+	 * RTCA is mapped only to find the ASU Channel ID
+	 * to be used by APU.
+	 * After reading the information RTCA region is unmapped.
+	 */
+	comm_chnl_info = core_mmu_add_mapping(MEM_AREA_IO_SEC,
+					      ASU_RTCA_BASEADDR,
+					      ASU_GLOBAL_ADDR_LIMIT);
+	if (!comm_chnl_info) {
+		EMSG("Failed to map runtime config area");
+		return channel_id;
+	}
+
+	for (id = 0; id < ASU_MAX_IPI_CHANNELS; id++) {
+		membase = (vaddr_t)comm_chnl_info +
+			  ASU_RTCA_CHANNEL_BASE_OFFSET +
+			  ASU_RTCA_CHANNEL_INFO_LEN * id;
+		if ((io_read32(membase) & ASU_CHNL_IPI_BITMASK) ==
+		    (CFG_AMD_APU_LCL_IPI_ID << 16)) {
+			channel_id = id;
+			DMSG("Use ASU channel ID %"PRIu32, channel_id);
+			break;
+		}
+	}
+
+	if (channel_id == ASU_MAX_IPI_CHANNELS)
+		EMSG("Failed to identify ASU channel ID for APU");
+
+	ret = core_mmu_remove_mapping(MEM_AREA_IO_SEC,
+				      comm_chnl_info, ASU_GLOBAL_ADDR_LIMIT);
+	if (ret)
+		EMSG("Failed to unmap RTCA");
+
+	return channel_id;
+}
+
+/*
+ * asu_alloc_unique_id() - Generate a unique identifier for ASU operations
+ *
+ * Creates a unique ID by cycling through available IDs in the callback
+ * reference array. Ensures no ID collision by checking if the slot is
+ * already in use.
+ *
+ * Return: Unique ID (1 to ASU_UNIQUE_ID_MAX-1) or ASU_UNIQUE_ID_MAX if none
+ *	   available
+ */
+uint8_t asu_alloc_unique_id(void)
+{
+	uint8_t unqid = 0;
+	uint32_t state = 0;
+
+	state = cpu_spin_lock_xsave(&asuid.slock);
+	while (unqid < ASU_UNIQUE_ID_MAX) {
+		if (asuid.ids[unqid] == ASU_UNIQUE_ID_MAX) {
+			asuid.ids[unqid] = unqid;
+			DMSG("Got unique ID %"PRIu8, unqid);
+			break;
+		}
+		unqid++;
+	};
+	cpu_spin_unlock_xrestore(&asuid.slock, state);
+
+	return unqid;
+}
+
+/**
+ * asu_free_unique_id() - Release a previously allocated unique ID
+ * @uniqueid: The unique ID to be freed
+ *
+ * Marks the specified unique ID as available for reuse.
+ * The released ID is set to ASU_UNIQUE_ID_MAX
+ * to indicate its availability. Intended to be used in environments where
+ * concurrent access to the unique ID pool occurs.
+ */
+void asu_free_unique_id(uint8_t uniqueid)
+{
+	uint32_t state = 0;
+
+	state = cpu_spin_lock_xsave(&asuid.slock);
+	asuid.ids[uniqueid] = ASU_UNIQUE_ID_MAX;
+	cpu_spin_unlock_xrestore(&asuid.slock, state);
+}
+
+/*
+ * get_free_index() - Find a free buffer index in the specified priority queue
+ * @priority: Priority level (ASU_PRIORITY_HIGH or ASU_PRIORITY_LOW)
+ *
+ * Searches for an available buffer slot in either the high priority (P0) or
+ * low priority (P1) channel queue. Updates the next free index pointer.
+ *
+ * Return: Buffer index (0 to ASU_MAX_BUFFERS-1) or ASU_MAX_BUFFERS if queue
+ *	   is full
+ */
+static uint8_t get_free_index(uint8_t priority)
+{
+	struct asu_channel_queue *qptr = NULL;
+	uint8_t index = 0;
+	uint32_t state = 0;
+
+	state = cpu_spin_lock_xsave(&asu->slock);
+
+	if (priority == ASU_PRIORITY_HIGH)
+		qptr = &asu->chnl_memptr->p0_chnl_q;
+	else
+		qptr = &asu->chnl_memptr->p1_chnl_q;
+
+	while (index < ASU_MAX_BUFFERS) {
+		if (qptr->queue_bufs[index].reqbufstatus == 0U ||
+		    qptr->queue_bufs[index].respbufstatus == 0U)
+			break;
+
+		index++;
+	}
+
+	if (index < ASU_MAX_BUFFERS)
+		qptr->queue_bufs[index].reqbufstatus = ASU_COMMAND_IS_PRESENT;
+
+	cpu_spin_unlock_xrestore(&asu->slock, state);
+
+	return index;
+}
+
+static void put_free_index(struct asu_channel_queue_buf *bufptr)
+{
+	uint32_t state = 0;
+
+	state = cpu_spin_lock_xsave(&asu->slock);
+	bufptr->reqbufstatus = 0;
+	bufptr->respbufstatus = 0;
+	cpu_spin_unlock_xrestore(&asu->slock, state);
+}
+
+/*
+ * send_doorbell() - Send IPI doorbell interrupt to ASU
+ *
+ * Triggers an Inter-Processor Interrupt (IPI) to notify the ASU
+ * that a new command is available in the shared memory queue.
+ *
+ * Return: TEE_SUCCESS on successful doorbell trigger
+ */
+static TEE_Result send_doorbell(void)
+{
+	io_write32((vaddr_t)asu->doorbell + IPIPSU_TRIG_OFFSET,
+		   ASU_TARGET_IPI_INT_MASK);
+
+	return TEE_SUCCESS;
+}
+
+/*
+ * asu_update_queue_buffer_n_send_ipi() - Queue command and send IPI
+ * @param: Client parameters including priority
+ * @req_buffer: Request buffer containing command data
+ * @size: Size of request data
+ * @header: Command header information
+ * @status: FW return status
+ *
+ * Places a command in the appropriate priority queue buffer, updates
+ * queue status, and sends an IPI to notify ASU of the pending command.
+ *
+ * Return: TEE_SUCCESS on success, appropriate error code on failure
+ */
+TEE_Result asu_update_queue_buffer_n_send_ipi(struct asu_client_params *param,
+					      void *req_buffer,
+					      uint32_t size,
+					      uint32_t header,
+					      uint32_t *status)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	uint8_t freeindex = 0;
+	struct asu_channel_queue_buf *bufptr = NULL;
+	struct asu_channel_queue *qptr = NULL;
+
+	if (!param) {
+		EMSG("Invalid parameters provided");
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	if (asu->is_ready != ASU_CLIENT_READY) {
+		EMSG("ASU client is not ready");
+		return TEE_ERROR_BAD_STATE;
+	}
+
+	freeindex = get_free_index(param->priority);
+	if (freeindex == ASU_MAX_BUFFERS) {
+		EMSG("ASU buffers full");
+		return TEE_ERROR_SHORT_BUFFER;
+	}
+
+	if (param->priority == ASU_PRIORITY_HIGH) {
+		bufptr = &asu->chnl_memptr->p0_chnl_q.queue_bufs[freeindex];
+		qptr = &asu->chnl_memptr->p0_chnl_q;
+	} else {
+		bufptr = &asu->chnl_memptr->p1_chnl_q.queue_bufs[freeindex];
+		qptr = &asu->chnl_memptr->p1_chnl_q;
+	}
+
+	bufptr->req.header = header;
+	if (req_buffer && size != 0U)
+		memcpy(bufptr->req.arg, req_buffer, size);
+
+	bufptr->respbufstatus = 0;
+
+	qptr->cmd_is_present = true;
+	qptr->req_sent++;
+	ret = send_doorbell();
+	if (ret != TEE_SUCCESS) {
+		EMSG("Failed to communicate to ASU");
+		return TEE_ERROR_COMMUNICATION;
+	}
+	while (io_read8((vaddr_t)&bufptr->respbufstatus) !=
+	       ASU_RESPONSE_IS_PRESENT) {
+		/*
+		 * WFE will return on SEV generated by the
+		 * interrupt handler or by a spin_unlock
+		 */
+		wfe();
+	}
+
+	*status = bufptr->resp.arg[ASU_RESPONSE_STATUS_INDEX];
+	if (param->cbhandler && !*status)
+		ret = param->cbhandler(param->cbptr, &bufptr->resp);
+	put_free_index(bufptr);
+
+	return ret;
+}
+
+static void asu_clear_intr(void)
+{
+	uint32_t status = 0;
+
+	status = io_read32((vaddr_t)asu->doorbell + IPIPSU_ISR_OFFSET);
+	io_write32((vaddr_t)asu->doorbell + IPIPSU_ISR_OFFSET,
+		   status & IPIPSU_ALL_MASK);
+}
+
+/*
+ * asu_resp_handler() - Interrupt handler for ASU responses
+ * @handler: Interrupt handler structure (unused)
+ *
+ * Interrupt service routine that processes responses from both high
+ * priority (P0) and low priority (P1) queues when ASU completes
+ * command processing.
+ *
+ * Return: ITRR_HANDLED indicating interrupt was handled
+ */
+static enum itr_return asu_resp_handler(struct itr_handler *handler __unused)
+{
+	sev();
+	asu_clear_intr();
+
+	return ITRR_HANDLED;
+}
+
+static struct itr_handler doorbell_handler = {
+	.it = PAR_IPIPSU_0_INT_ID,
+	.handler = asu_resp_handler,
+};
+
+/*
+ * setup_doorbell() - Initialize doorbell interrupt handling
+ *
+ * Maps the doorbell register region, configures interrupt settings,
+ * registers the interrupt handler, and enables the interrupt for
+ * receiving ASU response notifications.
+ *
+ * Return: Pointer to mapped doorbell region or NULL on failure
+ */
+static void *setup_doorbell(void)
+{
+	void *dbell = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	dbell = core_mmu_add_mapping(MEM_AREA_IO_SEC,
+				     asu_configtable.baseaddr,
+				     ASU_BASEADDR_SIZE);
+	if (!dbell) {
+		EMSG("Failed to map doorbell register");
+		return dbell;
+	}
+
+	io_write32((vaddr_t)dbell + IPIPSU_IER_OFFSET, IPIPSU_ALL_MASK);
+	io_write32((vaddr_t)dbell + IPIPSU_ISR_OFFSET, IPIPSU_ALL_MASK);
+
+	doorbell_handler.chip = interrupt_get_main_chip();
+
+	res = interrupt_add_configure_handler(&doorbell_handler,
+					      IRQ_TYPE_LEVEL_HIGH, 7);
+	if (res)
+		panic();
+
+	interrupt_enable(doorbell_handler.chip, doorbell_handler.it);
+
+	return dbell;
+}
+
+static void asu_init_unique_id(void)
+{
+	uint32_t idx = 0;
+
+	asuid.slock = SPINLOCK_UNLOCK;
+	for (idx = 0; idx < ARRAY_SIZE(asuid.ids); idx++)
+		asuid.ids[idx] = ASU_UNIQUE_ID_MAX;
+}
+
+/*
+ * asu_init() - Initialize the ASU driver and communication channel
+ *
+ * Performs complete ASU driver initialization including memory allocation,
+ * firmware readiness check, channel ID discovery, shared memory mapping,
+ * doorbell setup, and marking the client as ready for operation.
+ *
+ * Return: TEE_SUCCESS on successful initialization, appropriate error code on
+ *	   failure
+ */
+static TEE_Result asu_init(void)
+{
+	uint32_t channel_id = 0;
+	void *asu_shmem = NULL;
+	uint64_t membase = 0;
+
+	asu = calloc(1, sizeof(struct asu_client));
+	if (!asu) {
+		EMSG("Failed to allocate memory for ASU");
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+
+	asu->global_ctrl = core_mmu_add_mapping(MEM_AREA_IO_SEC,
+						ASU_BASEADDR,
+						ASU_BASEADDR_SIZE);
+	if (!asu->global_ctrl) {
+		EMSG("Failed to initialize ASU");
+		goto free;
+	}
+
+	if (asu_fwcheck() != TEE_SUCCESS) {
+		EMSG("ASU FW check failed");
+		goto global_unmap;
+	}
+
+	channel_id = asu_get_channelID();
+
+	if (channel_id == ASU_MAX_IPI_CHANNELS) {
+		EMSG("ASU channel for APU not configured");
+		goto global_unmap;
+	}
+
+	membase = ASU_CHANNEL_MEMORY_BASEADDR +
+			ASU_GLOBAL_ADDR_LIMIT * channel_id;
+	asu_shmem = core_mmu_add_mapping(MEM_AREA_IO_SEC,
+					 membase,
+					 ASU_GLOBAL_ADDR_LIMIT);
+	if (!asu_shmem) {
+		EMSG("Failed to map ASU SHM");
+		goto global_unmap;
+	}
+	asu_init_unique_id();
+	asu->doorbell = setup_doorbell();
+	if (!asu->doorbell) {
+		EMSG("Failed to set up ASU doorbell");
+		goto sh_unmap;
+	}
+
+	asu->chnl_memptr = asu_shmem;
+	asu->is_ready = ASU_CLIENT_READY;
+	asu->slock = SPINLOCK_UNLOCK;
+
+	IMSG("ASU initialization complete");
+
+	return TEE_SUCCESS;
+
+sh_unmap:
+	core_mmu_remove_mapping(MEM_AREA_IO_SEC, asu_shmem,
+				ASU_GLOBAL_ADDR_LIMIT);
+global_unmap:
+	core_mmu_remove_mapping(MEM_AREA_IO_SEC, asu->global_ctrl,
+				ASU_BASEADDR_SIZE);
+free:
+	free(asu);
+
+	EMSG("Failed to initialize ASU");
+
+	return TEE_ERROR_GENERIC;
+}
+
+service_init(asu_init);
--- a/core/drivers/crypto/asu_driver/asu_hash.c
+++ b/core/drivers/crypto/asu_driver/asu_hash.c
@@ -1,0 +1,448 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2025-2026, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#include <assert.h>
+#include <drivers/amd/asu_client.h>
+#include <drvcrypt_hash.h>
+#include <initcall.h>
+#include <io.h>
+#include <kernel/mutex.h>
+#include <kernel/panic.h>
+#include <kernel/unwind.h>
+#include <mm/core_memprot.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <tee/cache.h>
+#include <trace.h>
+#include <util.h>
+
+#define ASU_SHA_OPERATION_CMD_ID		0U
+/* SHA modes */
+#define ASU_SHA_MODE_SHA256			0U
+#define ASU_SHA_MODE_SHA384			1U
+#define ASU_SHA_MODE_SHA512			2U
+#define ASU_SHA_MODE_SHAKE256			4U
+
+/* SHA operation mode */
+#define ASU_SHA_START				0x1U
+#define ASU_SHA_UPDATE				0x2U
+#define ASU_SHA_FINISH				0x4U
+
+/* SHA hash lengths */
+#define ASU_SHA_256_HASH_LEN			32U
+#define ASU_SHA_384_HASH_LEN			48U
+#define ASU_SHA_512_HASH_LEN			64U
+#define ASU_SHAKE_256_HASH_LEN			32U
+#define ASU_SHAKE_256_MAX_HASH_LEN		136U
+#define ASU_DATA_CHUNK_LEN			4096U
+
+struct asu_shadev {
+	bool sha2_available;
+	bool sha3_available;
+	/* Control access to engine*/
+	struct mutex engine_lock;
+};
+
+struct asu_sha_op_cmd {
+	uint64_t dataaddr;
+	uint64_t hashaddr;
+	uint32_t datasize;
+	uint32_t hashbufsize;
+	uint8_t shamode;
+	uint8_t islast;
+	uint8_t opflags;
+	uint8_t shakereserved;
+};
+
+struct asu_hash_ctx {
+	struct crypto_hash_ctx hash_ctx; /* Crypto Hash API context */
+	struct asu_client_params cparam;
+	uint32_t shamode;
+	uint32_t shastart;
+	uint8_t uniqueid;
+	uint8_t module;
+};
+
+struct asu_hash_cbctx {
+	uint8_t *digest;
+	size_t len;
+};
+
+static const struct crypto_hash_ops asu_hash_ops;
+static struct asu_shadev *asu_shadev;
+static struct asu_hash_ctx *to_hash_ctx(struct crypto_hash_ctx *ctx);
+
+/**
+ * asu_hash_get_alg() - Get fw engine module ID and Hash mode.
+ * @algo:	TEE algo type.
+ * @module:	Engine module ID
+ * @mode:	Hash operation mode
+ * Map TEE algo type to fw module ID amd mode.
+ *
+ * Return: TEE_SUCCESS or TEE_ERROR_NOT_IMPLEMENTED
+ */
+
+static TEE_Result asu_hash_get_alg(uint32_t algo,
+				   uint32_t *module,
+				   uint32_t *mode)
+{
+	TEE_Result ret = TEE_SUCCESS;
+
+	switch (algo) {
+	case TEE_ALG_SHA256:
+		*module = ASU_MODULE_SHA2_ID;
+		*mode = ASU_SHA_MODE_SHA256;
+		break;
+	case TEE_ALG_SHA384:
+		*module = ASU_MODULE_SHA2_ID;
+		*mode = ASU_SHA_MODE_SHA384;
+		break;
+	case TEE_ALG_SHA512:
+		*module = ASU_MODULE_SHA2_ID;
+		*mode = ASU_SHA_MODE_SHA512;
+		break;
+	case TEE_ALG_SHA3_256:
+		*module = ASU_MODULE_SHA3_ID;
+		*mode = ASU_SHA_MODE_SHA256;
+		break;
+	case TEE_ALG_SHA3_384:
+		*module = ASU_MODULE_SHA3_ID;
+		*mode = ASU_SHA_MODE_SHA384;
+		break;
+	case TEE_ALG_SHA3_512:
+		*module = ASU_MODULE_SHA3_ID;
+		*mode = ASU_SHA_MODE_SHA512;
+		break;
+	default:
+		ret = TEE_ERROR_NOT_IMPLEMENTED;
+		break;
+	}
+
+	return ret;
+}
+
+/**
+ * asu_hash_initialize() - Initialize private asu_hash_ctx for hash operation.
+ * @ctx: crypto context used by the crypto_hash_*() functions
+ * Initialize hash operation request
+ *
+ * Return: TEE_SUCCESS or TEE_ERROR_BAD_PARAMETERS
+ */
+
+static TEE_Result asu_hash_initialize(struct crypto_hash_ctx *ctx)
+{
+	to_hash_ctx(ctx)->shastart = ASU_SHA_START;
+
+	return TEE_SUCCESS;
+}
+
+/**
+ * asu_sha_op() - Perform hash operation.
+ * @asu_hashctx:Request private hash context
+ * @op:		asu_sha_op_cmd parameters for fw engine
+ * @module:	Engine module ID
+ * @data:	Output digest received from engine
+ * Create request header, send and wait for result
+ * from engine.
+ *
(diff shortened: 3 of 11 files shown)


core/include/drivers/amd/asu_client.h
jenswikl, line 1: I'd prefer providing this file and the other non-private include files in `core/include/drivers/amd/`

core/include/drivers/amd/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+
+TEE_Result asu_validate_client_parameters(struct asu_client_params *param_ptr);
+TEE_Result asu_update_queue_buffer_n_send_ipi(struct asu_client_params *param,
+					      void *req_buffer,
+					      uint32_t size,
+					      uint32_t header,
+					      void *hashaddr,
+					      int *status,
+					      uint32_t bufsize);
+uint8_t asu_reg_callback_n_get_unique_id(struct asu_client_params *param,
```
jenswikl, line 51: Non-existing function.

core/include/drivers/amd/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+					 uint8_t *resp_buffer_ptr,
+					 uint32_t size);
+void asu_update_callback_details(uint8_t unique_id,
+				 uint8_t *resp_buffer_ptr,
+				 uint32_t size);
+uint8_t alloc_unique_id(void);
+void free_unique_id(uint8_t uniqueid);
+void *asu_update_n_get_ctx(uint8_t unique_id);
```
jenswikl: Non-existing function.

core/include/drivers/amd/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+TEE_Result asu_verify_n_get_unique_id_ctx(const void *context,
```
jenswikl: Non-existing function.

core/include/drivers/amd/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+#include "asu_sharedmem.h"
+
+#define ASU_PRIORITY_LOW		1
+#define ASU_PRIORITY_HIGH		0
+#define ASU_TRUE				1U
+#define ASU_FALSE				0U
+#define ASU_MODULE_SHA2_ID		1U
+#define ASU_MODULE_SHA3_ID		2U
+
+struct asu_client_params {
```
jenswikl, line 19: What is an ASU Client? Is it a piece of hardware?

core/drivers/amd/asu/include/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+#define ASU_PRIORITY_LOW		1
+#define ASU_PRIORITY_HIGH		0
+#define ASU_TRUE				1U
+#define ASU_FALSE				0U
+#define ASU_MODULE_SHA2_ID		1U
+#define ASU_MODULE_SHA3_ID		2U
+
+struct asu_client_params {
+	void (*cbhandler)(void *cbrefptr, uint32_t status);
+	void *cbrefptr;
```
jenswikl: What's `cbrefptr`? Callback reference pointer? Would it work with just `ptr`?

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+static void asu_init_unique_id(void)
+{
+	uint32_t idx = 0;
+
+	ids.slock = SPINLOCK_UNLOCK;
+	for (idx = 0; idx < ASU_UNIQUE_ID_MAX; idx++)
+		ids.id_array[idx] = ASU_UNIQUE_ID_MAX;
+}
+
+/**
```
jenswikl: Please remove the extra `*`. Same for the other comments with an extra `*`.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+	 */
+	comm_chnl_info = core_mmu_add_mapping(MEM_AREA_IO_SEC,
+					      ASU_RTCA_BASEADDR,
+					      ASU_GLOBAL_ADDR_LIMIT);
+	if (!comm_chnl_info) {
+		EMSG("Failed to Map Run Time Config Area");
+		return channel_id;
+	}
+
+	for (uint32_t id = 0; id < ASU_MAX_IPI_CHANNELS; id++) {
```
jenswikl: Please declare `id` together with the other locals above.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+		membase = (uintptr_t)comm_chnl_info +
```
jenswikl: Please use `vaddr_t` for this.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+			ASU_RTCA_CHANNEL_BASE_OFFSET +
```
jenswikl: Please align these with the first non-whitespace character after `=` on the line above.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+ *	   failure
+ */
+static uint32_t asu_get_channelID(void)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	void *comm_chnl_info = NULL;
+	uint32_t channel_id = ASU_MAX_IPI_CHANNELS;
+	uintptr_t membase = 0;
+
+	/* RTCA is mapped only to identify the ASU Channel ID
```
jenswikl: Wouldn't `find` be clearer? When we search for something, we usually find it rather than identify it.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+			break;
+		}
+	}
+
+	if (channel_id == ASU_MAX_IPI_CHANNELS)
+		EMSG("Failed to Identify ASU Channel ID for APU");
+
+	ret = core_mmu_remove_mapping(MEM_AREA_IO_SEC,
+				      comm_chnl_info, ASU_GLOBAL_ADDR_LIMIT);
+	if (ret != TEE_SUCCESS)
```
jenswikl: `if (ret)`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+ * 1 second for the check.
+ *
+ * Return: TEE_SUCCESS if firmware is ready, TEE_ERROR_BAD_STATE if not present
+ */
+static TEE_Result asu_fwcheck(void)
+{
+	uint64_t timeout = 0;
+	bool fw_state = false;
+
+	/* Timeout is set to ~1sec.
```
jenswikl:
``` C
/*
 * Please format
 * multi-line comments
 * like this.
 */

File: core/drivers/amd/asu/asu_main.c [hunk 13]
```diff
@@ -0,0 +1,535 @@
...
+#include <util.h>
+
+#include <asu_client.h>
+#include <asu_doorbell.h>
+
+#define ASU_QUEUE_BUFFER_FULL		0xFFU
+#define ASU_CLIENT_READY		0xFFFFFFFFU
+#define ASU_TARGET_IPI_INT_MASK		1U
+
+#define ASU_GLOBAL_BASEADDR		0xEBF80000U
```
jenswikl: What's "global" in `ASU_GLOBAL_BASEADDR`? Is it a piece of hardware?

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+	 * with components requests.
+	 */
+	timeout = timeout_init_us(ASU_ASUFW_BIT_CHECK_TIMEOUT_VALUE);
+
+	do {
+		if (io_read32((vaddr_t)asu->global_ctrl) &
+			      ASU_GLOBAL_GLOBAL_CNTRL_FW_IS_PRESENT_MASK) {
+			fw_state = true;
+			DMSG("Got the HSM FW");
+			break;
```
jenswikl: Why not return success here instead to keep things simple?

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+ *
+ * Return: Channel ID (0 to ASU_MAX_IPI_CHANNELS-1) or ASU_MAX_IPI_CHANNELS on
+ *	   failure
+ */
+static uint32_t asu_get_channelID(void)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	void *comm_chnl_info = NULL;
+	uint32_t channel_id = ASU_MAX_IPI_CHANNELS;
+	uintptr_t membase = 0;
```
jenswikl: `vaddr_t`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+};
+
+struct asu_ref_to_callback {
+	struct asu_client_params *cparam;
+	uint8_t *resp_buffer_ptr;
+	uint32_t size;
+};
+
+struct asu_ids {
+	uint8_t id_array[ASU_UNIQUE_ID_MAX];
```
jenswikl: How about `ids` instead?

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+	unsigned int slock;          /* id array spin lock */
+};
+
+static struct asu_ids ids;
```
jenswikl: I think the code would be easier to read if we used `static struct asu_ids asu_ids;`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+
+struct asu_client {
+	struct asu_channel_memory *chnl_memptr;
+	uint32_t is_ready;
+	unsigned int slock;          /* chnl_memptr spin lock */
+	void *global_ctrl;
+	void *doorbell;
+};
+
+struct asu_ref_to_callback {
```
jenswikl: This struct is unused.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+
+	return dbell;
+}
+
+static void asu_init_unique_id(void)
+{
+	uint32_t idx = 0;
+
+	ids.slock = SPINLOCK_UNLOCK;
+	for (idx = 0; idx < ASU_UNIQUE_ID_MAX; idx++)
```
jenswikl: `ARRAY_SIZE(ids.id_array)`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+		ids.id_array[idx] = ASU_UNIQUE_ID_MAX;
```
jenswikl: It's confusing to use the size of the array as the value for the unused entry.

core/drivers/amd/asu/include/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#ifndef _ASU_CLIENT_H_
+#define _ASU_CLIENT_H_
+
+#include "asu_sharedmem.h"
+
+#define ASU_PRIORITY_LOW		1
+#define ASU_PRIORITY_HIGH		0
+#define ASU_TRUE				1U
```
jenswikl: Do we really need special defines for true and false in the context of ASU?

core/drivers/amd/asu/include/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+
+static inline uint32_t asu_create_header(uint8_t cmd_id,
+					 uint8_t unique_id,
+					 uint8_t module_id,
+					 uint8_t command_len)
+{
+	uint32_t header = 0;
+
+	header = (cmd_id & ASU_COMMAND_ID_MASK) |
+		(unique_id << ASU_UNIQUE_REQ_ID_SHIFT) |
```
jenswikl: Please align with the `(` on the line above.
Please us the `SHIFT_U32()` macro from `<util.h>`
Same for the two lines below.

core/drivers/amd/asu/include/asu_client.h
```diff
@@ -0,0 +1,65 @@
...
+					      void *hashaddr,
+					      int *status,
+					      uint32_t bufsize);
+uint8_t asu_reg_callback_n_get_unique_id(struct asu_client_params *param,
+					 uint8_t *resp_buffer_ptr,
+					 uint32_t size);
+void asu_update_callback_details(uint8_t unique_id,
+				 uint8_t *resp_buffer_ptr,
+				 uint32_t size);
+uint8_t alloc_unique_id(void);
```
jenswikl: Please use proper prefixes.

core/drivers/amd/asu/include/asu_doorbell.h
```diff
@@ -0,0 +1,66 @@
...
+	uint32_t deviceid;
+	uintptr_t baseaddr;
+	uint32_t bitmask;
+	uint32_t intrid;
+	uintptr_t intrparent;
+	uint32_t target_count;
+	uint32_t target_mask[PAR_IPIPSU_NUM_TARGETS];
+};
+
+struct doorbell_config configtable = {
```
jenswikl: The initialization should be done in a .c file.

core/drivers/amd/asu/include/asu_sharedmem.h
```diff
@@ -0,0 +1,71 @@
...
+#define ASU_MAX_BUFFERS			8U
+#define ASU_CHANNEL_RESERVED_MEM	1188U
+#define ASU_COMMAND_IS_PRESENT		0x1U
+#define ASU_RESPONSE_IS_PRESENT		0x1U
+#define ASU_RESPONSE_STATUS_INDEX	0U
+#define ASU_RESPONSE_BUFF_ADDR_INDEX	1U
+#define ASU_COMMAND_ID_MASK		0x0000003FU
+#define ASU_UNIQUE_REQ_ID_MASK		0x00000FC0U
+#define ASU_UNIQUE_REQ_ID_SHIFT		6U
+#define ASU_UNIQUE_ID_MAX		(ASU_MAX_BUFFERS << 1U)
```
jenswikl: Please use `SHIFT_U32()`.

core/drivers/amd/asu/include/asu_sharedmem.h
```diff
@@ -0,0 +1,71 @@
...
+struct asu_channel_queue_buf {
+	uint8_t reqbufstatus;
+	uint8_t respbufstatus;
+	uint16_t reserved;
+	struct asu_req_buf req_buf;
+	struct asu_resp_buf resp_buf;
+};
+
+struct asu_channel_queue {
+	uint32_t is_cmd_present;
```
jenswikl: Is this the name used in the reference manual? If not, I'd prefer `cmd_is_present` since it reads better.

core/drivers/amd/asu/include/asu_sharedmem.h
```diff
@@ -0,0 +1,71 @@
...
+	uint32_t arg[ASU_COMMAND_RESP_ARGS];
+	uint32_t additionalstatus;
+	uint32_t reserved;
+};
+
+struct asu_channel_queue_buf {
+	uint8_t reqbufstatus;
+	uint8_t respbufstatus;
+	uint16_t reserved;
+	struct asu_req_buf req_buf;
```
jenswikl: Is the `_buf` suffix needed?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+#define ASU_SHA_256_HASH_LEN			(32U)
+#define ASU_SHA_384_HASH_LEN			(48U)
+#define ASU_SHA_512_HASH_LEN			(64U)
+#define ASU_SHAKE_256_HASH_LEN			(32U)
+#define ASU_SHAKE_256_MAX_HASH_LEN		(136U)
+#define ASU_DATA_CHUNK_LEN			(4096U)
+
+static struct crypto_hash_ops asu_hash_ops;
+struct asu_shadev {
+	uint32_t sha2_available;
```
jenswikl: Can we use `bool` instead?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+
+/* SHA hash lengths */
+#define ASU_SHA_256_HASH_LEN			(32U)
+#define ASU_SHA_384_HASH_LEN			(48U)
+#define ASU_SHA_512_HASH_LEN			(64U)
+#define ASU_SHAKE_256_HASH_LEN			(32U)
+#define ASU_SHAKE_256_MAX_HASH_LEN		(136U)
+#define ASU_DATA_CHUNK_LEN			(4096U)
+
+static struct crypto_hash_ops asu_hash_ops;
```
jenswikl: When possible, try to declare structs before global variables.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+struct asu_sha_op_cmd {
+	uint64_t dataaddr;
+	uint64_t hashaddr;
+	uint32_t datasize;
+	uint32_t hashbufsize;
+	uint8_t shamode;
+	uint8_t islast;
+	uint8_t opflags;
+	uint8_t shakereserved;
+} asu_ShaOpCmd;
```
jenswikl: What's `asu_ShaOpCmd` used for?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+	cache_operation(TEE_CACHEFLUSH, data, len);
+	op.hashaddr = 0;
+	op.hashbufsize = 0;
+	op.shamode = asu_hashctx->shamode;
+	op.islast = 0;
+
+	processed = 0;
+	remaining = len;
+	while (remaining) {
+		op.datasize = remaining > ASU_DATA_CHUNK_LEN ?
```
jenswikl: MAX(remaining, ASU_DATA_CHUNK_LEN)?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+				     ASU_DATA_CHUNK_LEN : remaining;
+		op.opflags = ASU_SHA_UPDATE |
+					   asu_hashctx->shastart;
+		op.dataaddr = (uint64_t)virt_to_phys(data + processed);
```
jenswikl: Is cast needed?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+	op.shamode = asu_hashctx->shamode;
+	op.islast = 0;
+
+	processed = 0;
+	remaining = len;
+	while (remaining) {
+		op.datasize = remaining > ASU_DATA_CHUNK_LEN ?
+				     ASU_DATA_CHUNK_LEN : remaining;
+		op.opflags = ASU_SHA_UPDATE |
+					   asu_hashctx->shastart;
```
jenswikl: Please fold up on the line above.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+				     const uint8_t *data, size_t len)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+
+	if (!len) {
+		IMSG("This is 0 len task, skip");
+		return TEE_SUCCESS;
+	}
+
+	if (!ctx || (!data && len)) {
```
jenswikl: Can `ctx` ever be NULL?
If we reach this point, we know that len is non-zero, so no need to check it again.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+	return ret;
+}
+
+static TEE_Result asu_hash_do_update(struct crypto_hash_ctx *ctx,
+				     const uint8_t *data, size_t len)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+
+	if (!len) {
+		IMSG("This is 0 len task, skip");
```
jenswikl: `DMSG`?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+	if (!digest || len == 0)
+		return TEE_ERROR_BAD_PARAMETERS;
+	cparam = &asu_hashctx->cparam;
+	cparam->priority = ASU_PRIORITY_HIGH;
+	cparam->cbrefptr = asu_hashctx;
+
+	/* Inputs of SHA request */
+	op.dataaddr = 0;
+	op.datasize = 0;
+	op.hashaddr = (uint64_t)virt_to_phys((void *)digest);
```
jenswikl: Is any of the casts needed?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+ * Send final request to engine and populate digest result
+ *
+ * Return: TEE_SUCCESS, TEE_ERROR_BAD_PARAMETERS or TEE_ERROR_GENERIC
+ */
+
+static TEE_Result asu_hash_final(struct asu_hash_ctx *asu_hashctx,
+				 uint8_t *digest, size_t len)
+{
+	TEE_Result ret = TEE_SUCCESS;
+	struct asu_sha_op_cmd op;
```
jenswikl: Please initialize the locals.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+	free_unique_id(asu_hashctx->uniqueid);
+	asu_hashctx->uniqueid = ASU_UNIQUE_ID_MAX;
+
+	return ret;
+}
+
+static TEE_Result asu_hash_do_final(struct crypto_hash_ctx *ctx,
+				    uint8_t *digest, size_t len)
+{
+	struct asu_hash_ctx *asu_hashctx;
```
jenswikl: Please initialize.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+
+	if (!ctx)
```
jenswikl: Can `ctx` be NULL?

core/include/drivers/amd/asu_client.h
Akshay-Belsare: ASU Client is the code interacting with the ASU Firmware.

core/drivers/amd/asu/include/asu_client.h
Akshay-Belsare: Callback reference pointer.
Given that the code handles a significant number of pointers, it was used for differentiation. Let me revisit the variables once.

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: sure!

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: This is the ASU register base address from the hardware.

jenswikl: I don't understand. Aren't all physical addresses global in a sense?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+		op.opflags = ASU_SHA_UPDATE |
+					   asu_hashctx->shastart;
+		op.dataaddr = (uint64_t)virt_to_phys(data + processed);
+		remaining -= op.datasize;
+		processed += op.datasize;
+		ret = asu_sha_op(asu_hashctx, &op,
+				 asu_hashctx->module, NULL);
+		if (ret)
+			break;
+		asu_hashctx->shastart = 0;
```
ainwood87, line 216: Is this a valid value for shastart?
I was wondering if we could avoid the break by checking ret in the loop-condition
`while (remaining && !ret)`
Can setting shastart = 0 be done unconditionally at the beginning of the loop?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+ * Return: TEE_SUCCESS or TEE_ERROR_GENERIC
+ */
+
+static TEE_Result asu_hash_update(struct asu_hash_ctx *asu_hashctx,
+				  uint8_t *data, size_t len)
+{
+	TEE_Result ret = TEE_SUCCESS;
+	struct asu_sha_op_cmd op = {};
+	struct asu_client_params *cparam;
+	uint32_t processed;
```
ainwood87: suggestion:
You could remove this variable.
This variable is only used to update op.dataaddr, but this could be done using op.datasize
``` 
data += op.datasize;
op.dataaddr = (uint64_t)virt_to_phys(data)
```
up to you.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,422 @@
...
+
+static void asu_hash_ctx_free(struct crypto_hash_ctx *ctx)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+
+	if (!ctx)
+		return;
+	asu_hashctx  = to_hash_ctx(ctx);
+	mutex_lock(&asu_shadev->engine_lock);
+	if (asu_hashctx->module == ASU_MODULE_SHA2_ID &&
```
ainwood87: Are sha2_available and sha3_available mutually exclusive?
Can't we just set them like (if using boolean)
```
asu_shadev->sha2_available = (asu_hashctx->module == ASU_MODULE_SHA2_ID );
asu_shadev->sha3_available = (asu_hashctx->module == ASU_MODULE_SHA3_ID );
```

core/drivers/amd/asu/include/asu_doorbell.h
Akshay-Belsare: The information in this file is generated through tool and the plan is to get this info through some other mechanism.

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: Addressed!

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Yes

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: It unused. Can be removed.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Will change to MIN(remaining, ASU_DATA_CHUNK_LEN)

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: It can be removed.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: sure

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: This is entry point for Driver. not NULL is preventive check of input data.
len == 0 is valid case and driver returning SUCCESS, but
(len && !data) ==> means length is non zero but buffer pointer *data* is NULL, which is BAD parameter.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Can be removed

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, will do in next patch

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: This is entry point for Driver. not NULL is preventive check of input data.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: >Is this a valid value for shastart?
Yes, It's like flag, Setting it to 1 tell Hardware that it is 1st update request. Subsequent update requests should not set this flag.
> Can setting shastart = 0 be done unconditionally at the beginning of the loop?
No, We are initialising it to "ASU_SHA_START" in asu_hash_initialize.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Yes, it can be done. Thanks, will do

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: > Are sha2_available and sha3_available mutually exclusive?
No, H/W has separate instance of sha2 and sha3 engine. Both type of requests can be processed in parallel.
> Can't we just set them like
No, Being shared variable, It will update both sha2_available and sha3_available.

core/include/drivers/amd/asu_client.h
jenswikl: So, ASU Client is the driver? Doesn't that make the client part redundant?

Edit: Keep this as is, if it makes more sense to you.

Akshay-Belsare: @jenswi-linaro 
The ASU Client is shared code that interfaces with the firmware and exposes a general API set for crypto drivers to use.

Akshay-Belsare:
@jenswi-linaro 
I would appreciate it if you could take another look at this review when you have the opportunity.

core/drivers/amd/asu/asu_main.c
jenswikl: It's still unused.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,535 @@
...
+
+	/* Timeout is set to ~1sec.
+	 * This is the worst case time within which ASUFW ready for interaction
+	 * with components requests.
+	 */
+	timeout = timeout_init_us(ASU_ASUFW_BIT_CHECK_TIMEOUT_VALUE);
+
+	do {
+		if (io_read32((vaddr_t)asu->global_ctrl) &
+			      ASU_GLOBAL_GLOBAL_CNTRL_FW_IS_PRESENT_MASK) {
```
jenswikl: Please align with `io_read32`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,530 @@
...
+
+	qptr->cmd_is_present = true;
+	qptr->req_sent++;
+	ret = send_doorbell();
+	if (ret != TEE_SUCCESS) {
+		EMSG("Failed to communicate to ASU");
+		return TEE_ERROR_COMMUNICATION;
+	}
+	while (io_read8((vaddr_t)(&bufptr->respbufstatus)) !=
+	      ASU_RESPONSE_IS_PRESENT) {
```
jenswikl: Please align with `io_read8`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,530 @@
...
+	bufptr->respbufstatus = 0U;
+
+	qptr->cmd_is_present = true;
+	qptr->req_sent++;
+	ret = send_doorbell();
+	if (ret != TEE_SUCCESS) {
+		EMSG("Failed to communicate to ASU");
+		return TEE_ERROR_COMMUNICATION;
+	}
+	while (io_read8((vaddr_t)(&bufptr->respbufstatus)) !=
```
jenswikl: Please remove the `()` around `&bufptr->respbufstatus`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,530 @@
...
+	      ASU_RESPONSE_IS_PRESENT) {
+		/*
+		 * WFE will return on SEV generated by the
+		 * interrupt handler or by a spin_unlock
+		 */
+		wfe();
+	}
+
+	*status = bufptr->resp.arg[ASU_RESPONSE_STATUS_INDEX];
+	if (param->cbhandler && !(*status))
```
jenswikl: Please remove `()` around `*status`.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,530 @@
...
+
+	io_write32((vaddr_t)dbell + IPIPSU_IER_OFFSET, IPIPSU_ALL_MASK);
+	io_write32((vaddr_t)dbell + IPIPSU_ISR_OFFSET, IPIPSU_ALL_MASK);
+
+	doorbell_handler.chip = interrupt_get_main_chip();
+
+	res = interrupt_add_configure_handler(&doorbell_handler,
+					      IRQ_TYPE_LEVEL_HIGH, 7);
+	if (res) {
+		core_mmu_remove_mapping(MEM_AREA_IO_SEC, dbell,
```
jenswikl: Why do this if you're going to panic?

core/drivers/amd/asu/include/asu_doorbell.h
```diff
@@ -0,0 +1,66 @@
...
+#define PAR_PSV_IPI_6_BIT_MASK		0x200U
+#define IPIPSU_ALL_MASK			0xFFFFU
+
+#define IPIPSU_TRIG_OFFSET	0x00U
+#define IPIPSU_OBS_OFFSET	0x04U
+#define IPIPSU_ISR_OFFSET	0x10U
+#define IPIPSU_IMR_OFFSET	0x14U
+#define IPIPSU_IER_OFFSET	0x18U
+
+struct doorbell_config {
```
jenswikl: Please add an `asu_` prefix.

core/drivers/amd/asu/include/asu_doorbell.h
```diff
@@ -0,0 +1,66 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 - 2025 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ */
+
+#ifndef _ASU_DOORBELL_H_
```
jenswikl: Can we use `__` instead of `_` as a prefix?

core/drivers/amd/asu/asu_doorbell.h
jenswikl, line 1: Does this file need to be a global include file, or can it be placed in `core/drivers/amd/asu` instead?

core/drivers/amd/asu/include/asu_doorbell.h
jenswikl: We should ideally use a proper prefix for global variables.

core/drivers/amd/asu/include/asu_sharedmem.h
```diff
@@ -0,0 +1,72 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ */
+
+#ifndef _ASU_SHAREDMEM_H_
```
jenswikl: Prefer a `__` prefix over `_`

core/include/drivers/amd/asu_sharedmem.h
jenswikl, line 1: Please move this into `core/include/drivers/amd`, where `asu_client.h` is kept.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+ * Initialize hash operation request
+ *
+ * Return: TEE_SUCCESS or TEE_ERROR_BAD_PARAMETERS
+ */
+
+static TEE_Result asu_hash_initialize(struct crypto_hash_ctx *ctx)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+
+	if (!ctx) {
```
jenswikl: Apart from `sec_hash_initialize()`, this is the only `hash_op::init()` function to check this. Callers should not supply a NULL pointer, and `to_hash_ctx()` catches that with an assert.

This function could be simplified as:
``` C
to_hash_ctx(ctx)->shastart = ASU_SHA_START;

return TEE_SUCCESS;
```

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+ *
+ * Return: TEE_SUCCESS or TEE_ERROR_GENERIC
+ */
+
+static TEE_Result asu_sha_op(struct asu_hash_ctx *asu_hashctx,
+			     struct asu_sha_op_cmd *op,
+			     uint8_t module)
+{
+	TEE_Result ret = TEE_SUCCESS;
+	uint32_t header;
```
jenswikl: Please initialize all locals as described in https://optee.readthedocs.io/en/latest/general/coding_standards.html point 3. Applies to the entire PR.

core/drivers/crypto/asu_driver/asu_hash.c
jenswikl: I don't think that `ctx` needs to be checked. The other drivers don't check it.
We know that len is non-zero, so a NULL `data` is always wrong at this point.

core/drivers/crypto/asu_driver/asu_hash.c
jenswikl: I don't think that `ctx` needs to be checked. The other drivers don't check it.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+ * Release crypto engine and free private context memory.
+ *
+ * Return: void
+ */
+
+static void asu_hash_ctx_free(struct crypto_hash_ctx *ctx)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+
+	if (!ctx)
```
jenswikl: This is already checked in `crypto_hash_free_ctx()`.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+	asu_hashctx->uniqueid = ASU_UNIQUE_ID_MAX;
+	mutex_lock(&asu_shadev->engine_lock);
+	if (asu_hashctx->module == ASU_MODULE_SHA2_ID &&
+	    !asu_shadev->sha2_available)
+		asu_shadev->sha2_available = 1;
+	else if (asu_hashctx->module == ASU_MODULE_SHA3_ID &&
+		 !asu_shadev->sha3_available)
+		asu_shadev->sha3_available = 1;
+	mutex_unlock(&asu_shadev->engine_lock);
+	asu_hashctx = to_hash_ctx(ctx);
```
jenswikl: This is already done above.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+	free(asu_hashctx);
+}
+
+static struct crypto_hash_ops asu_hash_ops = {
```
jenswikl: Please make it `const` also.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+	.init = asu_hash_initialize,
+	.update = asu_hash_do_update,
+	.final = asu_hash_do_final,
+	.free_ctx = asu_hash_ctx_free,
+	.copy_state = NULL,
```
jenswikl, line 351: I don't think this can be NULL.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+
+static TEE_Result asu_hash_ctx_allocate(struct crypto_hash_ctx **ctx,
+					uint32_t algo)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+	uint32_t module;
+	uint32_t shamode;
+	TEE_Result ret = TEE_SUCCESS;
+
+	if (!ctx) {
```
jenswikl: This must not be NULL, so no need to check it. The other drivers don't check it.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+	return ret;
+}
+
+static TEE_Result asu_hash_init(void)
+{
+	TEE_Result ret = TEE_SUCCESS;
+
+	asu_shadev = calloc(1, sizeof(*asu_shadev));
+	mutex_init(&asu_shadev->engine_lock);
+	asu_shadev->sha2_available = 1;
```
jenswikl: `sha2_available` and `sha3_available` are `bool` and should be assigned `true` or `false`. Please fix on all occasions.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,462 @@
...
+	if (ret)
+		return ret;
+	mutex_lock(&asu_shadev->engine_lock);
+	if (module == ASU_MODULE_SHA2_ID && asu_shadev->sha2_available) {
+		asu_shadev->sha2_available = 0;
+	} else if (module == ASU_MODULE_SHA3_ID && asu_shadev->sha3_available) {
+		asu_shadev->sha3_available = 0;
+	} else {
+		mutex_unlock(&asu_shadev->engine_lock);
+		return TEE_ERROR_NOT_IMPLEMENTED;
```
jenswikl, line 393: If a TA opens a second hash state, it will face `TEE_ERROR_NOT_IMPLEMENTED`. How is this supposed to work?

core/include/drivers/amd/asu_client.h
```diff
@@ -0,0 +1,63 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 - 2025 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ */
+
+#ifndef _ASU_CLIENT_H_
```
jenswikl: Prefer `__` prefix.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, will remove ctx !NULL  check.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, will remove ctx !NULL check.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, Will update

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, Will remove

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Will remove

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Engine does not support state copy operation. This is limitation of Hardware.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, removed

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: It will fallback to software.

core/drivers/amd/asu/asu_main.c
harshjain32: I missed it, Will remove in next version

core/drivers/amd/asu/asu_main.c
harshjain32: Sure

core/drivers/amd/asu/asu_main.c
harshjain32: Sure

core/drivers/amd/asu/asu_main.c
harshjain32: Sure

core/drivers/amd/asu/asu_main.c
harshjain32: Sure

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: removed!

core/drivers/amd/asu/asu_doorbell.h
Akshay-Belsare: It's used by the ASU Infra code and will not be utilized by any ASU drivers.

Akshay-Belsare:
@jenswi-linaro Please review at your earliest convenience.

core/include/drivers/amd/asu_client.h
jenswikl: This function is still neither implemented nor called.

core/include/drivers/amd/asu_client.h
jenswikl: This function is still neither implemented nor called.

core/drivers/amd/asu/asu_doorbell.h
```diff
@@ -0,0 +1,66 @@
...
+
+#define PAR_IPIPSU_NUM_INSTANCES	1U
+
+/* Parameter definitions for peripheral ASU IPI */
+#define PAR_IPIPSU_0_DEVICE_ID		0U
+#define PAR_IPIPSU_0_BASEADDR		0xEB330000U
+#define PAR_IPIPSU_0_BIT_MASK		0x00000001U
+#define PAR_IPIPSU_0_INT_ID		0x59U
+#define PAR_IPIPSU_NUM_TARGETS		16U
+#define PAR_PSV_IPI_PMC_BIT_MASK	(0x2U)
```
jenswikl: The `()` around the values are not needed. Please fix all occurrences in this PR.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+ * Return: TEE_SUCCESS on success, appropriate error code on failure
+ */
+TEE_Result asu_update_queue_buffer_n_send_ipi(struct asu_client_params *param,
+					      void *req_buffer,
+					      uint32_t size,
+					      uint32_t header,
+					      int *status)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	uint8_t freeindex = 0U;
```
jenswikl: `0` should be enough.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+	if (param->cbhandler && !*status)
+		ret = param->cbhandler(param->cbptr, &bufptr->resp);
+	put_free_index(bufptr);
+
+	return ret;
+}
+
+static void asu_clear_intr(void)
+{
+	uint32_t status;
```
jenswikl: Please zero-initialize.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+ * firmware readiness check, channel ID discovery, shared memory mapping,
+ * doorbell setup, and marking the client as ready for operation.
+ *
+ * Return: TEE_SUCCESS on successful initialization, appropriate error code on
+ *	   failure
+ */
+static TEE_Result asu_init(void)
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	uint32_t channel_id = 0U;
```
jenswikl: `0`

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+	void *asu_shmem = NULL;
+	uint64_t membase = UINT64_C(0);
```
jenswikl: Initializing integers of any size, signed or unsigned, with `0` should be enough. Please fix all occurrences in this PR.

core/drivers/crypto/asu_driver/asu_hash.c
jenswikl: Then we'll crash if `crypto_hash_copy_state()` is called. Is this code tested with xtest?

harshjain32: We skipped copy related xtest test cases in testing.
By default we are disabling the driver in conf, till we have workaround in place.

Akshay-Belsare:
@jenswi-linaro When convenient, please take a look at the code changes.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,444 @@
...
+static void asu_hash_ctx_free(struct crypto_hash_ctx *ctx)
+{
+	struct asu_hash_ctx *asu_hashctx = NULL;
+
+	asu_hashctx = to_hash_ctx(ctx);
+	asu_free_unique_id(asu_hashctx->uniqueid);
+	asu_hashctx->uniqueid = ASU_UNIQUE_ID_MAX;
+	mutex_lock(&asu_shadev->engine_lock);
+	if (asu_hashctx->module == ASU_MODULE_SHA2_ID &&
+	    !asu_shadev->sha2_available)
```
jenswikl: If `asu_hashctx->module == ASU_MODULE_SHA2_ID` then `asu_shadev->sha2_available` must be false or there's an inconsistency. There's no need to check `asu_shadev->sha2_available`. If we must, then an `assert()` before setting it to true makes more sense.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,444 @@
...
+ * Return: TEE_SUCCESS or TEE_ERROR_GENERIC
+ */
+
+static TEE_Result asu_sha_op(struct asu_hash_ctx *asu_hashctx,
+			     struct asu_sha_op_cmd *op,
+			     uint8_t module)
+{
+	TEE_Result ret = TEE_SUCCESS;
+	uint32_t header = 0;
+	int status = TEE_ERROR_GENERIC;
```
jenswikl: If you must use a non-zero value, please use `-1`, but I think `0` should be enough.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+	while (io_read8((vaddr_t)&bufptr->respbufstatus) !=
+	       ASU_RESPONSE_IS_PRESENT) {
+		/*
+		 * WFE will return on SEV generated by the
+		 * interrupt handler or by a spin_unlock
+		 */
+		wfe();
+	}
+
+	*status = bufptr->resp.arg[ASU_RESPONSE_STATUS_INDEX];
```
jenswikl, line 325: bufptr->resp.arg[] is an array of `uint32_t`, why not make `status` a `uint32_t` pointer?

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+
+/*
+ * send_doorbell() - Send IPI doorbell interrupt to ASU
+ *
+ * Triggers an Inter-Processor Interrupt (IPI) to notify the ASU
+ * that a new command is available in the shared memory queue.
+ *
+ * Return: TEE_SUCCESS on successful doorbell trigger
+ */
+static TEE_Result send_doorbell(void)
```
jenswikl, line 247: Do you plan for a more complex version of this function that eventually can fail?

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+/*
+ * asu_validate_client_parameters() - Validate client parameter structure
+ * @param_ptr: Pointer to client parameters to validate
+ *
+ * Performs validation checks on client parameters including null pointer
+ * checks, callback function availability, and priority value validation.
+ *
+ * Return: TEE_SUCCESS if valid, appropriate error code otherwise
+ */
+TEE_Result asu_validate_client_parameters(struct asu_client_params *param_ptr)
```
jenswikl: This function is unused.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,522 @@
...
+	/*
+	 * RTCA is mapped only to find the ASU Channel ID
+	 * to be used by APU.
+	 * After reading the information RTCA region is unmapped.
+	 */
+	comm_chnl_info = core_mmu_add_mapping(MEM_AREA_IO_SEC,
+					      ASU_RTCA_BASEADDR,
+					      ASU_GLOBAL_ADDR_LIMIT);
+	if (!comm_chnl_info) {
+		EMSG("Failed to Map Run Time Config Area");
```
jenswikl: Please use normal capitalization. Please fix all the obvious ones in this PR.

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: Currently, our focus is on developing additional drivers based on this platform. While we have not yet implemented a more complex version of this function that could potentially fail, we are open to considering such enhancements in the future as requirements evolve. Do you have any specific scenarios or suggestions in mind that we should take into account?

jenswikl: The function only returns `TEE_SUCCESS`, so I was wondering if it must return anything.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Actually yes, We should assert, updated

Thanks

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: done

core/drivers/amd/asu/asu_main.c
Akshay-Belsare: Noted.

core/drivers/amd/asu/asu_main.c
```diff
@@ -0,0 +1,492 @@
...
+{
+	TEE_Result ret = TEE_ERROR_GENERIC;
+	uint32_t channel_id = 0;
+	void *asu_shmem = NULL;
+	uint64_t membase = 0;
+
+	asu = calloc(1, sizeof(struct asu_client));
+	if (!asu) {
+		EMSG("Failed to allocate memory for ASU");
+		return ret;
```
jenswikl: `return TEE_ERROR_OUT_OF_MEMORY;`?

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -0,0 +1,445 @@
...
+	cparam->cbptr = &cbctx;
+	cparam->cbhandler = asu_hash_cb;
+
+	/* Inputs of SHA request */
+	op.dataaddr = 0;
+	op.datasize = 0;
+	op.hashaddr = virt_to_phys((void *)digest);
+	op.hashbufsize = len;
+	if (asu_hashctx->shamode == ASU_SHA_MODE_SHA256)
+		op.hashbufsize = ASU_SHA_256_HASH_LEN;
```
jenswikl, line 291: What if `len` is smaller than `ASU_SHA_256_HASH_LEN`?

harshjain32: Hi,

Actually FW don't copy hash to digest buffer, but it expects "op.hashaddr" as non zero. It populates hash in response buffer and driver copy it in asu_hash_cb() based on len only.

It will be safe if len is less that hash size.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Yes, Done, set to 0.

Akshay-Belsare:
@jenswi-linaro The reason for the CI failure is unclear.

jforissier:
@Akshay-Belsare see https://github.com/OP-TEE/manifest/pull/342. I'm restarting the jobs.

Akshay-Belsare:
Thanks @jforissier !
@jenswi-linaro Please review at your earliest convenience.

core/drivers/crypto/asu_driver/asu_hash.c
jenswikl: If any non-zero value works for `op.hashaddr`, why not use `UINT64_MAX` or some other safe value?

jenswikl:
Please squash the fixups into their parent commits.

core/drivers/crypto/asu_driver/asu_hash.c
jenswikl: I think it needs to be documented that this code is work-in-progress since you'd normally expect things to work unless there's a warning.

core/drivers/crypto/asu_driver/asu_hash.c
jenswikl: ?

jenswikl:
Please check for questions or comments I've made that haven't been answered. There are at least two above.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, Will add comment before copy_state.
Current engine does not support partial state copy operation.

core/drivers/crypto/asu_driver/asu_hash.c
harshjain32: Sure, Will update to UINT64_MAX.

core/drivers/crypto/asu_driver/asu_hash.c
```diff
@@ -345,6 +345,9 @@ static const struct crypto_hash_ops asu_hash_ops = {
 	.update = asu_hash_do_update,
 	.final = asu_hash_do_final,
 	.free_ctx = asu_hash_ctx_free,
+	/*
+	 * Current engine does not support partial state copy operation.
```
jenswikl, line 349: I was having a stronger and more visible message in mind to avoid needless security reports from people finding a new way of crashing OP-TEE.

harshjain32: How about changing it to.

Current engine does not support partial state copy operation.  Any operation from REE involving state copy can crash the OS when driver is enabled,

harshjain32: @jenswi-linaro : let me know if you any more details needs to be added. I will update the comment accordingly.

jenswikl: Please add a note in `core/arch/arm/plat-versal2/conf.mk`, where the config flags enabling this are documented.

jenswikl:

Akshay-Belsare:
thanks @jenswi-linaro!
Tag applied to the patches.
CC: @harshj-xlnx

PR #7522, opened 2025-09-10
drivers: stm32_rtc: only register interrupt if necessary

tbourgoi:
When EXTI IRQ controller registers an interrupt line in OP-TEE the line is set as secure.
If a line is set as secure, the Linux drivers will refuse to register it after.
The RTC interrupt line is only used if the RTC is set as a wakeup source or if the RTC is secured by the RIF (only for STM32MP2x) On STM32MP1, Linux is using the wakeup capabilities of the RTC and requires the interrupt, so OP-TEE should not register the interrupt line and securise it.
To prevent OP-TEE from setting the interrupt line as secure when Linux will need it, we only register the interrupt line when the property "wakeup-source" is present in the DT or if the RIF configuration the RTC as secure.

This patch fix the probe error of the RTC driver of the Linux kernel on stm32mp157c-dk2

### DIFF
--- a/core/drivers/stm32_rtc.c
+++ b/core/drivers/stm32_rtc.c
@@ -332,7 +332,6 @@ static void apply_rif_config(bool is_tdcid)
 	/* Check if all resources must be secured */
 	if (seccfgr == RTC_RIF_FULL_SECURED) {
 		io_setbits32(base + RTC_SECCFGR, RTC_SECCFGR_FULL_SEC);
-		rtc_dev.is_secured = true;
 
 		if (!(io_read32(base + RTC_SECCFGR) & RTC_SECCFGR_FULL_SEC))
 			panic("Bad RTC seccfgr configuration");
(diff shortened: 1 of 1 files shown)


core/drivers/stm32_rtc.c
```diff
@@ -475,6 +462,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
 
+static TEE_Result stm32_rtc_register_dt_interrupt(const void *fdt, int node)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (fdt_get_property(fdt, node, "wakeup-source", NULL))
+		rtc_dev.rtc->is_wakeup_source = true;
+
+	if (!rtc_dev.rtc->is_wakeup_source &&
+	    !(IS_ENABLED(CFG_RTC_PTA) && IS_ENABLED(CFG_CORE_ASYNC_NOTIF) &&
```
GseoC: What is the purpose of all of these conditions?
isn't `if (!rtc_dev.rtc->is_wakeup_source || !rtc_dev.is_secured)` enough?
If not, can you elaborate a bit more in a comment above please?

core/drivers/stm32_rtc.c
```diff
@@ -475,6 +462,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
 	return TEE_SUCCESS;
 }
 
+static TEE_Result stm32_rtc_register_dt_interrupt(const void *fdt, int node)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (fdt_get_property(fdt, node, "wakeup-source", NULL))
+		rtc_dev.rtc->is_wakeup_source = true;
```
clegoffic: As it parses the DT it should be in parse_dt() as it was no ?

core/drivers/stm32_rtc.c
```diff
@@ -475,6 +462,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
+
+	if (!rtc_dev.rtc->is_wakeup_source &&
+	    !(IS_ENABLED(CFG_RTC_PTA) && IS_ENABLED(CFG_CORE_ASYNC_NOTIF) &&
+	      rtc_dev.is_secured))
+		return TEE_SUCCESS;
+
+	res = interrupt_dt_get(fdt, node, &rtc_dev.itr_chip, &rtc_dev.itr_num);
+	if (res)
+		return res;
```
clegoffic: dito, maybe do not get the interrupt when wakeup source is not set ?

core/drivers/stm32_rtc.c
```diff
@@ -475,6 +462,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
+
+		interrupt_enable(rtc_dev.itr_chip, rtc_dev.itr_num);
+
+		return TEE_SUCCESS;
+
+out_rtc_secured_and_itr_chip:
+		interrupt_remove_handler(rtc_dev.itr_handler);
+		notif_free_async_value(rtc_dev.notif_id);
+		return res;
+	}
```
clegoffic: No need to move all this code.
I think just add the is_wakeup_source in the condition is enough

core/drivers/stm32_rtc.c
clegoffic: When the RTC is secured I added the PTA to be able to read the RTC value from the kernel, and the async notif is mandatory if the kernel set a "secured" alarm

core/drivers/stm32_rtc.c
tbourgoi: Yes and it will be easier to handle the return error "defer_init" that can create a memory leak with the pointer  caused by the allocation used to store the RIF configuration from the device tree

tbourgoi:
I pushed a V2 that take into account the comments from @Upsylonbare 

The registering of the IRQ is done in the parse_dt() function after the parsing of the RIF configuration.
This allows to update the variable `is_secured` 
Since the registration of the IRQ line can return a defer init if the IRQ controler is not probed. I need to release allocated memory prior to return the error code.

core/drivers/stm32_rtc.c
```diff
@@ -430,47 +429,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
-	if (!cuint) {
-		DMSG("No RIF configuration available");
-		return TEE_SUCCESS;
+	res = interrupt_dt_get(fdt, node, &rtc_dev.itr_chip, &rtc_dev.itr_num);
+	if (res) {
+		free(rtc_dev.conf_data->cid_confs);
+		free(rtc_dev.conf_data->sec_conf);
+		free(rtc_dev.conf_data->priv_conf);
+		free(rtc_dev.conf_data->access_mask);
+		free(rtc_dev.conf_data);
```
etienne-lms: `rtc_dev.conf_data` may be NULL.
```c
	if (res) {
		if (rtc_dev.conf_data) {
			free(rtc_dev.conf_data->cid_confs);
			free(rtc_dev.conf_data->sec_conf);
			free(rtc_dev.conf_data->priv_conf);
			free(rtc_dev.conf_data->access_mask);
			free(rtc_dev.conf_data);
		}
		return res;
	}

File: core/drivers/stm32_rtc.c [hunk 6]
```diff
@@ -430,47 +429,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
+					    RTC_NB_RIF_RESOURCES);
+
+		if (rtc_dev.conf_data->sec_conf[0] == RTC_RIF_FULL_SECURED)
+			rtc_dev.is_secured = true;
 	}
 
-	if (!rtc_dev.compat->has_rif_support)
+	cuint = fdt_getprop(fdt, node, "wakeup-source", NULL);
+	if (!cuint && !(IS_ENABLED(CFG_RTC_PTA) &&
+			IS_ENABLED(CFG_CORE_ASYNC_NOTIF) && rtc_dev.is_secured))
```
etienne-lms, line 472: Could you cross check this condition?
In which condition do you expect to handle the RTC interrupt?

core/drivers/stm32_rtc.c
```diff
@@ -430,47 +429,60 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
-	rtc_dev.conf_data->priv_conf = calloc(1, sizeof(uint32_t));
-	rtc_dev.conf_data->access_mask = calloc(1, sizeof(uint32_t));
-	if (!rtc_dev.conf_data->cid_confs || !rtc_dev.conf_data->sec_conf ||
-	    !rtc_dev.conf_data->priv_conf || !rtc_dev.conf_data->access_mask)
-		panic("Not enough memory capacity for RTC RIF config");
-
-	for (i = 0; i < rtc_dev.nb_res; i++)
-		stm32_rif_parse_cfg(fdt32_to_cpu(cuint[i]), rtc_dev.conf_data,
-				    RTC_NB_RIF_RESOURCES);
+	if (cuint) {
```
clegoffic, line 479: The variable name cuint for wakeup_source and st,protreg isn't very clear to me.
What does it represents ?
Maybe use two different variables with clear names.

core/drivers/stm32_rtc.c
clegoffic: Indeed as it raise a lot of comments maybe a comment inside the code or in the commit message can help to understand.

tbourgoi: The condition is OK, I'll add a comment to explain it

core/drivers/stm32_rtc.c
tbourgoi: `c` is for constant and it's a pointer to the value of the property
There are a lot of driver using this name for the return code of fdt_getprop(). I don't want to create 2 variables if that's not necessary since were handling the property one by one. I'll keep it like this

core/drivers/stm32_rtc.c
tbourgoi: Ok Done

core/drivers/stm32_rtc.c
clegoffic: Thank you for the info

core/drivers/stm32_rtc.c
```diff
@@ -461,17 +461,26 @@ static TEE_Result parse_dt(const void *fdt, int node)
 	}
 
 	cuint = fdt_getprop(fdt, node, "wakeup-source", NULL);
+	/*
+	 * if the wakeup-source property is not present in the DT
+	 *    AND
+	 *    the RTC_PTA is disable or ASYNC_NOTIF are disable
+	 *    or the rtc in not secured
```
clegoffic: Typo s/in/is/

tbourgoi: done

tbourgoi:
@etienne-lms do you have any other remark ? are you okay if I squash both fixups ?

etienne-lms (commented):
LGTM. Can you squash the fixup commits. 
Could you refine the commit message? Especially the 2 first sentences. Maybe rephrase that as something like *"Prevent registering of RTC interrupt when it's not used by OP-TEE as it prevents non-secure world from handling that interrupt"*.

tbourgoi:
> Prevent registering of RTC interrupt when it's not used by OP-TEE as it prevents non-secure world from handling that interrupt

@etienne-lms I've update the commit title and simplified the first 2 sentences of the commit message

jforissier:
The CI errors are unrelated to this PR ("no space left on device"). I will investigate.

core/drivers/stm32_rtc.c
```diff
@@ -430,47 +429,69 @@ static TEE_Result parse_dt(const void *fdt, int node)
...
-	    !rtc_dev.conf_data->priv_conf || !rtc_dev.conf_data->access_mask)
-		panic("Not enough memory capacity for RTC RIF config");
-
-	for (i = 0; i < rtc_dev.nb_res; i++)
-		stm32_rif_parse_cfg(fdt32_to_cpu(cuint[i]), rtc_dev.conf_data,
-				    RTC_NB_RIF_RESOURCES);
+	if (cuint) {
+		if (!interrupt_can_set_wake(rtc_dev.itr_chip)) {
+			EMSG("%s does not have wakeup capabilities",
+			     rtc_dev.itr_chip->name);
```
etienne-lms, line 482: For consistency, this should also free `rtc_dev.conf_data` resources. Maybe a `err` branch label to factorize that?

etienne-lms (commented):
Reword the commit message?
```patch
 The RTC interrupt line is only used if the RTC is set as a wakeup source
 or if the RTC is secured by the RIF (only for STM32MP2x)
 
-The RTC interrupt line is only used if the RTC is set as a wakeup source
-or if the RTC is secured by the RIF (only for STM32MP2x)
-On STM32MP1, Linux is using the wakeup capabilities of the RTC and
-requires the interrupt, so OP-TEE should not register the interrupt line
-and securise it.
-To prevent OP-TEE from setting the interrupt line as secure when Linux
-will need it, we only register the interrupt line when the property
-"wakeup-source" is present in the DT or if the RIF configuration the RTC
-as secure.
+The RTC interrupt line is only used if the RTC is set as a wakeup source
+or RTC is secured by the RIF (only for STM32MP2x) and RTC PTA and
+async notif are enabled.

core/drivers/stm32_rtc.c
tbourgoi: Ok done, it's better for consistency even if the core of OP-TEE will crash due to the error returned

tbourgoi:
Commit message reword

tbourgoi:
Hello,
@jforissier do you see anything missing preventing from merging this PR ?

jforissier:
> Hello, @jforissier do you see anything missing preventing from merging this PR ?

No, it can be merged. Thanks.

